PR #1897, opened 2018-07-16
yt-4.0 normalization fix

AshKelly:
## PR Summary

Fixed normalization of some pixelization routines with the addition of two utility functions to normalize

The plots with the fix:
with_fix

The plots without the fix:
without_fix

NOTE: These fixes may break the tests, I'm not 100% sure.

## PR Checklist

- **NA** New features are documented, with docstrings and narrative docs
- **NA** Adds a test for any bugs fixed. Adds tests for new features.

### DIFF
--- a/doc/source/yt4differences.rst
+++ b/doc/source/yt4differences.rst
@@ -68,30 +68,27 @@ lifting in these functions is undertaken by cython functions.
 
 It is now possible to generate slice plots, projection plots and arbitrary grids
 of smoothed quanitities using these operations. The following code demonstrates
-how this could be achieved.
+how this could be achieved:
 
-```python
+.. code-block:: python
 
-import yt
+    import yt
 
-ds = yt.load('snapshot_033/snap_033.0.hdf5')
+    ds = yt.load('snapshot_033/snap_033.0.hdf5')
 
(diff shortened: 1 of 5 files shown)


ngoldbaum:
So if I understand correctly, we're now always doing normalization, is that correct? Is `use_normalization` still useful to keep around even though we've hard-coded it to `False`? Maybe it would be less confusing to remove it?

AshKelly:
Yes, all of them are getting normalized after the entire deposition.

Hm, well I guess we don't **need** it now - but I'd have to go through and check everywhere we have used it where we have done the deposition in a single iteration that I use the utility function to normalize, that is probably more maintainable.

If a user wanted to view a plot without a normalization, we don't currently have an API to support that. Do you think that should also be a dataset property?

ngoldbaum:
That would probably be the most natural way to expose it, yeah, something like `ds.use_sph_normalization = True`.

JBorrow:
Just to chime in -- this looks a lot better now! I was concerned yesterday about the circles that spuriously appeared in the old version. Good job @AshKelly!

AshKelly:
not sure why this failed on xcode, but I think this ready to merge

ngoldbaum:
The Mac failure is a timeout, not your fault, happens sometimes.

PR #1899, opened 2018-07-17
Fixing removed files from manifest

cphyc:
## PR Summary

Commit a271399f12da6a887d61aa9e5780209f0005f28d removed the static files required for the mapserver (to use online ones). This PR removes leftover paths listed in MANIFEST.in that point to non-existing file in the trunk.


ngoldbaum:
Good catch!

PR #1905, opened 2018-07-19
Fix implementation of AnswerTestingTest.prefix

git-abhishek:
## PR Summary

For an answer testing test, if we specify `prefix` then it should be compared against the data file in `prefix` directory of  `answer-store`.  However, it is not the case currently.

With this fix, all the answer tests could be executed with just one `nosetests` command.

For example, with the following command all the tests having the decorator `@attr("answer_test")` could be executed (all these tests are of type answer test).
```
nosetests --attr "answer_test" --with-answer-testing --nologcapture -d -v --local --local-dir=answer-store yt
```

 Without this fix, we have to pass one more parameter 
`--answer-name="name_of_the_answer_file_in_answer_store"` and execute each of the answer tests one by one.

## PR Checklist

- [ ] **TODO:** Update the doc [Answer Testing](http://yt-project.org/doc/developing/testing.html#answer-testing) with `prefix` usage 

## Adding Reviewers

### DIFF
--- a/yt/utilities/answer_testing/framework.py
+++ b/yt/utilities/answer_testing/framework.py
@@ -138,11 +138,11 @@ class AnswerTesting(Plugin):
                 print('Please supply an output directory with the --local-dir option')
                 sys.exit(1)
             storage_class = AnswerTestLocalStorage
+            output_dir = os.path.realpath(options.output_dir)
             # Fix up filename for local storage
             if self.compare_name is not None:
-                self.compare_name = "%s/%s/%s" % \
-                    (os.path.realpath(options.output_dir), self.compare_name,
-                     self.compare_name)
+                self.compare_name = os.path.join(output_dir, self.compare_name,
+                                                 self.compare_name)
 
             # Create a local directory only when `options.answer_name` is
             # provided. If it is not provided then creating local directory
@@ -150,13 +150,10 @@ class AnswerTesting(Plugin):
             # test, this case is handled in AnswerTestingTest.
             if self.store_name is not None and options.store_results \
                     and options.answer_name is not None:
-                name_dir_path = "%s/%s" % \
-                    (os.path.realpath(options.output_dir),
-                    self.store_name)
+                name_dir_path = os.path.join(output_dir, self.store_name)
                 if not os.path.isdir(name_dir_path):
                     os.makedirs(name_dir_path)
-                self.store_name= "%s/%s" % \
-                        (name_dir_path, self.store_name)
+                self.store_name= os.path.join(name_dir_path, self.store_name)
         else:
             storage_class = AnswerTestCloudStorage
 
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Some of these failures are due to no old answer being available (which could be fixed by incrementing the appropriate answer number I think) but it looks like many are dying in shelve with tracebacks like this:

```
  File "/usr/lib64/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/usr/lib64/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/fido/workspace/yt_py2_git/yt/utilities/answer_testing/framework.py", line 358, in __call__
    dd = self.reference_storage.get(self.storage_name)
  File "/home/fido/workspace/yt_py2_git/yt/utilities/answer_testing/framework.py", line 256, in get
    ds = shelve.open(self.reference_name, protocol=-1)
  File "/usr/lib64/python2.7/shelve.py", line 243, in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
  File "/usr/lib64/python2.7/shelve.py", line 227, in __init__
    Shelf.__init__(self, anydbm.open(filename, flag), protocol, writeback)
  File "/usr/lib64/python2.7/anydbm.py", line 85, in open
    return mod.open(file, flag, mode)
error: (2, 'No such file or directory')
```

Which I think is a case where no old answer was available but the answer testing framework isn't handling it correctly (just speculation, I haven't looked into this deeply).

git-abhishek:
### Listing various usages for anwer testing using nosetests command line arguments:

Assume `test_plot.py` has 4 tests, 2 are answer tests (`answer_test_1`, `answer_test_2`) and 2 unit tests (`unit_test_1`, `unit_test_2`) and directory structure is `yt/plots/tests/test_plot.py`
```python
#test_plot.py
#------------
...

def test_unit_test_1():
    pass

def test_unit_test_2():
    pass

def test_answer_test_1():
    # prefix value is set to be answer_test_1
    pass

def test_answer_test_2():
    # prefix value is set to be answer_test_2
    pass

...

```

1. `nosetests --with-answer-testing --local --local-dir=answer-store yt/plots/tests/test_plot.py`

    Will run all the 4 tests. For the answer tests to pass, answers should be present at the directory location `answer-store`. This was not possible earlier.

2. `nosetests --with-answer-testing --local --local-dir=answer-store yt/plots/tests/test_plot.py:test_answer_test_1`

    Will execute `test_answer_test_1`. Earlier we had to put one more argument `--answer-name=answer_test_1`.

3. `nosetests --with-answer-testing --local --local-dir=answer-store --answer-store yt/plots/tests/test_plot.py:test_answer_test_1`

    Will store golden answers of `test_answer_test_1`. Earlier we had to put one more argument `--answer-name=answer_test_1`.



4. `nosetests --with-answer-testing --local --local-dir=answer-store --answer-store yt/plots/tests/test_plot.py`

    This will generate and store all the answers in  **one** answer-store file. Hence, we have to execute the above statement individually for each answer test. **Do not generate answers in batch mode.**

ngoldbaum:
Awesome, the new features look great!

It would be nice to update the documentation on answer testing to reflect the improvements you're making here and in the other PR:

http://yt-project.org/doc/developing/testing.html#answer-testing

There are probably a number of ways that section of the docs could be improved. If you feel like updating those docs should be saved for a subsequent pull request that's cool, just let me know.

@Xarthisius as the person who is unfortunately most familiar with the answer testing framework, could you take a look at this?

@brittonsmith @jwise77 fair warning there is a small but nonzero chance that this might break the enzo tests. Worth testing?

ngoldbaum:
I spoke with Britton and he is ok with merging this since it won't directly impact the enzo tests until after yt 3.5 is released. Going to merge now. Thanks for your patience here abhishek!

PR #1912, opened 2018-07-20
[bugfix] python3 fix for selection in dm-only Enzo data

brittonsmith:
This fixes a corner case for data selection in dark matter only Enzo datasets in Python 3.

In an Enzo dm-only dataset, grids with no particles do not have an associated filename.  In Python 2, sorting will happily and meaninglessly sort objects of different type, but Python 3 doesn't like that.  This fix removes grids with no filename from the selection list entirely. This is ok because we don't want them anyway.

I would add a test, but the dataset in which I can reproduce this is 2 GB and it may be difficult to reproduce. I don't know how often you end up with dark matter only datasets with grids that have no particles.

Who thought there were any py2->py3 bugs left?

### DIFF
--- a/yt/geometry/grid_geometry_handler.py
+++ b/yt/geometry/grid_geometry_handler.py
@@ -283,8 +283,6 @@ class GridIndex(Index):
     def _identify_base_chunk(self, dobj):
         fast_index = None
         def _gsort(g):
-            if g.filename is None:
-                return g.id
             return g.filename
         if dobj._type_name == "grid":
             dobj._chunk_info = np.empty(1, dtype='object')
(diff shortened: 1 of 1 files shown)


brittonsmith:
@MatthewTurk, excellent point.  I have checked for the test case that broke, and the total cell volume is the same whether grids with `filename` of None are included or not. That said, I'm not convinced this couldn't be a problem in another circumstance.  I've issued PR #1919 with an alternative fix more along the lines of what you describe. Let me know which you think is better.

brittonsmith:
This solution definitely won't work since things like `fake_amr_ds` are in memory only and so `grid.filename` is always `None`. I'm going to close this in favor of PR #1919.

PR #1914, opened 2018-07-22
Improve GIZMO fields

qobilidop:
## PR Summary

This PR aims to improve GIZMO frontend's field system. The following are done:

- The field list is cleaned up according to the GIZMO User Manual.
- New derived fields are added:
  - Star creation time: `("PartType4", "creation_time")`
  - Star age: `("PartType4", "creation_time")`

While I'm at it, if you have any suggestions about GIZMO fields, I'm willing to add them to this PR.

## PR Checklist

### DIFF
--- a/yt/frontends/gizmo/fields.py
+++ b/yt/frontends/gizmo/fields.py
@@ -62,7 +62,9 @@ class GizmoFieldInfo(GadgetFieldInfo):
          ("code_magnetic", ["particle_magnetic_field"], None)),
         ("DivergenceOfMagneticField",
          ("code_magnetic / code_length", [], None)),
-        ("StellarFormationTime", ("code_time", ["creation_time"], None)),
+        ("StellarFormationTime", ("", [], None)),
+        # "StellarFormationTime" has different meanings in (non-)cosmological
+        # runs, so units are left blank here.
         ("BH_Mass", ("code_mass", [], None)),
         ("BH_Mdot", ("code_mass / code_time", [], None)),
         ("BH_Mass_AlphaDisk", ("code_mass", [], None)),
@@ -79,6 +81,8 @@ class GizmoFieldInfo(GadgetFieldInfo):
         if ptype in ("PartType0",):
             self.setup_gas_particle_fields(ptype)
             setup_species_fields(self, ptype)
+        if ptype in ("PartType4",):
+            self.setup_star_particle_fields(ptype)
 
     def setup_gas_particle_fields(self, ptype):
         super(GizmoFieldInfo, self).setup_gas_particle_fields(ptype)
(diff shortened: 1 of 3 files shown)


qobilidop:
@ngoldbaum I'm not sure if my choices of the derived fields' names are appropriate. Are there standard names for those fields?

yt/frontends/gizmo/fields.py
```diff
@@ -32,25 +32,20 @@
...
         ("Density", ("code_mass / code_length**3", ["density"], None)),
-        ("MaximumTemperature", ("K", [], None)),
-        ("Temperature", ("K", ["temperature"], None)),
-        ("Epsilon", ("code_length", [], None)),
-        ("Metals", ("code_metallicity", ["metallicity"], None)),
-        ("Metallicity", ("code_metallicity", ["metallicity"], None)),
-        ("Phi", ("code_length", [], None)),
+        ("SmoothingLength", ("code_length", ["smoothing_length"], None)),
+        ("ElectronAbundance", ("", [], None)),
+        ("NeutralHydrogenAbundance", ("", [], None)),
```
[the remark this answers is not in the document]
ngoldbaum: I'd give the on-disk field units and then update `setup_gas_particle_fields` to adapt to that. There's no reason to leave on-disk fields as dimensionless when we know their units.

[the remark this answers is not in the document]
qobilidop: But `"NeutralHydrogenAbundance"` is just a fraction, so it should be dimensionless, right?

qobilidop:
@ngoldbaum I have checked the first commit of GIZMO, and the velocity and mass fields are named `"Velocities"` and `"Masses"` from then. So I think there won't be any issue of breaking the ability to read old GIZMO files. @chummels probably knows more about this.

yt/frontends/gizmo/fields.py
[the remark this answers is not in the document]
ngoldbaum: Oh, duh. Please ignore me :)

yt/frontends/gizmo/tests/test_outputs.py
```diff
@@ -75,3 +73,52 @@ def test_gizmo_mhd():
         f = ad[ptype, fmag + '_' + axis]
         assert str(f.units) == 'code_magnetic'
         assert f.shape == (409013,)
+
+
+@requires_ds(gmhd, big_data=True)
```
qobilidop: I'm not sure if it's necessary to set `big_data=True` here.

ngoldbaum: It doesn’t make any difference

ngoldbaum: Actually both of the new tests you added aren't answer tests and shouldn't use `required_ds` at all. Instead use `requires_file`. That way they will run as part of the unit tests if anyone has the test dataset you're using in their `test_data_dir` path.

qobilidop: Thanks! I didn't know this.

ngoldbaum:
@yt-fido test this please

yt/frontends/gizmo/fields.py
```diff
@@ -32,25 +32,20 @@
                   "Mg", "Si", "S", "Ca", "Fe"]
 
 class GizmoFieldInfo(GadgetFieldInfo):
+    # The known fields list is according to the GIZMO User Guide. See
+    # http://www.tapir.caltech.edu/~phopkins/Site/GIZMO_files/gizmo_documentation.html#snaps-reading
     known_particle_fields = (
-        ("Mass", ("code_mass", ["particle_mass"], None)),
```
chummels: So we're removing these previously-defined fields because they're not in the base-level gizmo file?

qobilidop: I'm doing this mostly to make the list cleaner and more maintainable because the list now maps from the GIZMO User Manual directly. I have checked the first commit of the GIZMO code and believe there are no historical GIZMO snapshots with these variant field names. It would be great if I could get further confirmation from you.

ngoldbaum:
@yt-fido test this please

PR #1918, opened 2018-07-24
add masked I/O to unstructured mesh frontends

### LINKED ISSUE #1916: Unstructured mesh slices return data with inconsistent shapes
### Bug report

**Bug summary**

This was initially reported on the mailing list as a failure in the annotate_contours plot callback.

The issue is illustrated with the following script:

```
import yt
ds = yt.load('source.e', step=-1)

slc = ds.slice(2, 1.0)

# fcoords expects four elements in the slice
print(slc.fcoords)

# but we seem to have 8 elements returned
print(slc['all', 'group_0'])
```

Right now this prints:

```
[[0.5 0.5 1.5]
 [1.5 0.5 1.5]
 [0.5 1.5 1.5]
 [1.5 1.5 1.5]] code_length
[0.02927492 0.00967331 0.00967331 0.00469298 0.         0.
 0.         0.        ] dimensionless
```

**Expected outcome**

```
[[0.5 0.5 1.5]
 [1.5 0.5 1.5]
 [0.5 1.5 1.5]
 [1.5 1.5 1.5]] code_length
[0.02927492 0.00967331 0.00967331 0.00469298] dimensionless
```

**Version Information**

  * Operating System: mac
  * Python Version: 3.7.0
  * yt version: current master branch

ngoldbaum:
closes #1916 

This makes it so that data object accesses on unstructured mesh data return data with the correct shape (i.e. not equal to the shape of the total number of mesh elements in the dataset).

As far as I can tell, this was not a problem until now because the unstructured mesh pixelizer uses `ds.all_data()` to generate the data it needs to feed to the pixelizer. I think with this change in principle we could use a slice data object there.

Do I need to touch other places? I'm not sure. Maybe the PYNE/MOAB frontend? @munkm this might impact your Denuvo frontend.

### DIFF
--- a/yt/frontends/exodus_ii/io.py
+++ b/yt/frontends/exodus_ii/io.py
@@ -86,6 +86,7 @@ class IOHandlerExodusII(BaseIOHandler):
                                                        (field_ind + 1, mesh_id)][:]
                         data = fdata[self.ds.step, :]
                         ind += g.select(selector, data, rv[field], ind)  # caches
+                rv[field] = rv[field][:ind]
             return rv
 
     def _read_chunk_data(self, chunk, fields):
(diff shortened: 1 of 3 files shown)


munkm:
Thanks for the heads up @ngoldbaum!

PR #1919, opened 2018-07-24
[bugfix] Alternate fix for PR #1912

brittonsmith:
This is an alternative implementation of the bugfix PR #1912.

After @MatthewTurk's comment, I've implemented a fix that does not remove grids where the `filename` is None and instead allows them to be sorted and potentially operated on. This may be safer than what I did in PR #1912. Only one of these two should be accepted.

### DIFF
--- a/yt/geometry/grid_geometry_handler.py
+++ b/yt/geometry/grid_geometry_handler.py
@@ -282,10 +282,6 @@ class GridIndex(Index):
 
     def _identify_base_chunk(self, dobj):
         fast_index = None
-        def _gsort(g):
-            if g.filename is None:
-                return str(g.id)
-            return g.filename
         if dobj._type_name == "grid":
             dobj._chunk_info = np.empty(1, dtype='object')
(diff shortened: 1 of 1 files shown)


ngoldbaum:
I have no idea why this impacts `test_axial_pixelization`, although I could see why it happens since that test use `fake_amr_ds` under the hood.

The travis failures are unrelated, I suspect we have a memory leak somewhere.

brittonsmith:
Ok thanks for clarifying that. I'll look into `test_axial_pixelization` and see if there's something to it.

brittonsmith:
The failures in `test_axial_pixelization` were due to the change from sorting by `grid.id` to `str(grid.id)`, which changed the grid order set in `_identify_base_chunk`. Not sure why the order should matter for those tests, but it definitely did.

ngoldbaum:
The python3 test failures can be ignored.

Nice job Britton! Thanks for pushing on this.

Issue #1921, opened 2018-07-25
Additions to StreamlineCallback


KiranEiden:
I recently added some additional options to yt's StreamlineCallback that I used for plotting some 2D axisymmetric datasets, and I would like to incorporate the changes into the main repository. There is likely a better way to implement the changes, however, especially the second set, so I would prefer to get some feedback here if possible before making a pull request.

The first plot shows the output with both additional features enabled (display threshold and streamline colorbar), and the second without using a display threshold.

![plt40811_with_threshold](https://user-images.githubusercontent.com/32306975/43214659-e1350904-9007-11e8-8960-48d31aff0b0b.png)
![plt40811_temp_slice](https://user-images.githubusercontent.com/32306975/43214950-d91b5f6a-9008-11e8-8bbe-c19eea6e0647.png)

### Display Threshold

The first addition allows for a display threshold to be set; any sections of streamline where the color field is less than the threshold will be removed from the plot (linewidth will be set to 0 - matplotlib doesn't currently support modulating alpha). I added the following code under ``if self.field_color:`` in the callback:

```python
if self.display_threshold:

    mask = (field_colors > self.display_threshold)
    lwdefault = matplotlib.rcParams['lines.linewidth']

    if 'linewidth' in self.plot_args:
        linewidth = self.plot_args['linewidth']
    else:
        linewidth = lwdefault

    try:
        linewidth *= mask
        self.plot_args['linewidth'] = linewidth
    except ValueError:
        err_msg = "Error applying display threshold: linewidth" + \
                "must have shape ({}, {}) or be scalar"
        err_msg = err_msg.format(ny, nx)
        raise ValueError(err_msg)
```

### Streamline Colorbar

I was uncertain how to add a second colorbar to a single plot, and have all of the font settings and the unit formatting be automatically applied, without refactoring yt's internals. I also couldn't find where the aspect for a regular colorbar is set, so I hardcoded an approximate formula here. I would ideally like to make this more maintainable, without having explicit, independent settings, before submitting a pull request. Here's the code in its present state, where the ``strm`` variable was returned by matplotlib's streamplot function:

```python
if self.add_colorbar and self.field_color:

    import matplotlib.colorbar as colorbar
    from yt.units.unit_object import Unit
    from matplotlib.mathtext import MathTextParser

    aspect = 25 * (yy1 - yy0) / (xx1 - xx0)
    cax, kw = colorbar.make_axes(plot._axes, location='right',
             aspect=aspect)
[...]

PR #1922, opened 2018-07-26
Adds display_threshold option to StreamlineCallback

KiranEiden:
## PR Summary

Adds a feature to the StreamlineCallback that suppresses streamline segments where the color field fails to meet a given threshold. Also added a test for the callback.

**Note:** It is doubtful that the test ran properly - I may be having issues with my nose installation. If someone could make certain that the test executes and passes that would be appreciated.

The feature was discussed in issue #1921.

## PR Checklist

### DIFF
--- a/doc/source/visualizing/callbacks.rst
+++ b/doc/source/visualizing/callbacks.rst
@@ -498,7 +498,8 @@ Overplot Streamlines
 ~~~~~~~~~~~~~~~~~~~~
 
 .. function:: annotate_streamlines(self, field_x, field_y, factor=16, \
-                                   density = 1, plot_args=None)
+                                   density=1, display_threshold=None, \
+                                   plot_args=None)
 
    (This is a proxy for
    :class:`~yt.visualization.plot_modifications.StreamlineCallback`.)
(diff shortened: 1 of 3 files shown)


AshKelly:
It looks like the python3.6 AppVeyor build fail is unrelated to your work. The rest of your tests appear to have passed other than a python3.6 timeout which I've restarted.

ngoldbaum:
Also if you want to describe a little bit more what issues you were having with nose and running the new tests you added I might be able to provide suggestions. In particular if you can make a gist with the full error message and traceback (if any) that you saw.

For what it's worth it looks like the test ran on jenkins and passed.

KiranEiden:
Great, thank you! There were no error messages actually - the test just passed for everything, even in cases where I explicitly raised an error. The only thing that would cause it to fail was invalid python syntax. I can take a look at my nose setup; I'm not very familiar with the framework, so it may be a relatively straightforward issue to fix if I read the docs.

ngoldbaum:
Could it have been that you didn't have the dataset specified by `requires_file` in the `test_data_dir`? You'd needed to have set `test_data_dir` in your yt config file and the dataset would need to be extracted there for the test to actually run locally on your machine.

ngoldbaum:
http://yt-project.org/doc/reference/configuration.html#the-configuration-file

KiranEiden:
That may be it, I'll try adding that to my config file.

ngoldbaum:
Maybe @MatthewTurk, @chummels, or @jisuoqing  could take a look at this since you all are familiar with the plot callback code. If anyone else wants to do a review pass please feel free. I think we just need one more reviewer and this should be good to merge.

ngoldbaum:
Thank you for your contribution! :fireworks: :tada: :shipit:

KiranEiden:
You're welcome, thanks for your input!

PR #1925, opened 2018-07-27
set verbosity to 2 in nose config

ngoldbaum:
This lets us more easily see what tests are running on travis. We've recently been seeing random test failures (probably due to using too much memory somewhere or a new memory leak) and I want to know what test it's dying on.

### DIFF
--- a/nose.cfg
+++ b/nose.cfg
@@ -6,6 +6,6 @@ cover-package=yt
 detailed-errors=1
 exclude=answer_testing
 nocapture=1
-verbosity=1
+verbosity=2
 where=yt
 with-timer=1


git-abhishek:
Probably we could also add `df -h` before and after the tests to see memory usage as suggested by [I’m running out of disk space in my build](https://docs.travis-ci.com/user/common-build-problems/#im-running-out-of-disk-space-in-my-build). Or [Running in Debug mode](https://docs.travis-ci.com/user/running-build-in-debug-mode/) could be an option too.

ngoldbaum:
Self-merging as this will help debug failures elsewhere.

ngoldbaum:
Ah, I didn't know travis had a debug mode, looks useful!

PR #1926, opened 2018-07-27
In boxlib/data_structures split using os.sep

git-abhishek:
## PR Summary

Fixes `IndexError: list index out of range` while executing following script in windows:

```python
import yt
from yt.utilities.answer_testing.framework import data_dir_load
raw_fields = "Laser/plt00015"
ds = data_dir_load(raw_fields)
```

[Full error log](http://paste.yt-project.org/show/43/)

## PR Checklist


MatthewTurk:
As a note, we might also be able to make use of https://docs.python.org/2/library/os.path.html#os.path.split (2) and https://docs.python.org/3.7/library/os.path.html#os.path.split (3)

PR #1929, opened 2018-07-29
Derived quantity particles

RicardaBeckmann:
## PR Summary

This pull request adds an option argument called "p_type" to a variety of derived quantities, such as center_of_mass and similar, which allows that derived quantity to be calculated for a given particle type only. 

This is a new feature that will aid in data analysis, allowing for example the spin paramter of stars or the centre of mass of DM to be calculated easily and quickly.

The default behaviour of all derived quantities has not been changed, as p_type defaults to "all". 

## PR Checklist

PS: Sorry, I also removed a bunch of trailing whitespaces which makes the pull request look messier than it actually is.

### DIFF
--- a/yt/data_objects/derived_quantities.py
+++ b/yt/data_objects/derived_quantities.py
@@ -228,6 +228,10 @@ class CenterOfMass(DerivedQuantity):
         Flag to include particles in the calculation.  Particles are ignored
         if not present.
         Default: False
+    p_type : str
+        Particle type to be used for Center of mass calculation when use_particle
+        = True.
+        Default: all
 
     Examples
     --------
@@ -237,22 +241,22 @@ class CenterOfMass(DerivedQuantity):
     >>> print ad.quantities.center_of_mass()
 
     """
-    def count_values(self, use_gas = True, use_particles = False):
+    def count_values(self, use_gas = True, use_particles = False, p_type="all"):
         use_gas &= \
           (("gas", "cell_mass") in self.data_source.ds.field_info)
         use_particles &= \
-          (("all", "particle_mass") in self.data_source.ds.field_info)
+          ((p_type, "particle_mass") in self.data_source.ds.field_info)
         self.num_vals = 0
         if use_gas:
             self.num_vals += 4
         if use_particles:
             self.num_vals += 4
 
-    def process_chunk(self, data, use_gas = True, use_particles = False):
+    def process_chunk(self, data, use_gas = True, use_particles = False, p_type="all"):
         use_gas &= \
           (("gas", "cell_mass") in self.data_source.ds.field_info)
         use_particles &= \
-          (("all", "particle_mass") in self.data_source.ds.field_info)
+          ((p_type, "particle_mass") in self.data_source.ds.field_info)
         vals = []
         if use_gas:
             vals += [(data["gas", ax] *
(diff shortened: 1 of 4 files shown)


RicardaBeckmann:
@chummels Ah sorry, I didn't see that. Looks like you have already put a lot of effort into the angular momentum vector so we should definitely stick with that.

As for the kwarg: Sure. Is simply renaming p_type to particle_type enough, or do I need to do anything else with it? Sorry for the naive question.

ngoldbaum:
I agree with Cameron that the new arguement should be spelled `particle_type`. It should be sufficient to just replace all occurrences in this PR. I haven’t looked through the code yet so I may have other suggestions.

ngoldbaum:
The python3 jenkins tear failures are unrelated to this PR and can be ignored.

RicardaBeckmann:
I changed the particle type name as discussed.

I noticed that the error message raised when a garbage string is entered for particle_type isn't very useful so I'll need to do something about that.

ngoldbaum:
The content of `ds.particle_types` or `ds.particle_types_raw` should be useful for that 😬

yt/data_objects/derived_quantities.py
```diff
@@ -237,33 +243,35 @@ class CenterOfMass(DerivedQuantity):
     >>> print ad.quantities.center_of_mass()
 
     """
-    def count_values(self, use_gas = True, use_particles = False):
+    def count_values(self, use_gas = True, use_particles = False, particle_type="all"):
+        if not particle_type in self.data_source.ds.particle_types:
```
ngoldbaum: `flake8` would prefer this to be phrased `if particle_type not in self.data_source.ds.particle_types`. This also happens a few more times below.

RicardaBeckmann:
For the tests, if I use the fake_particle_ds, is it okay to hardcode numbers for comparison or is that considered bad practice?

ngoldbaum:
`fake_particle_ds` uses randomly generated particle positions, so those will only agree if the seed is fixed (which I think it is by default)

RicardaBeckmann:
So... is that a yes or a no? 

I tried recreating it a few times in different environments and it to produce consistent values but i guess if anything about it changes then the tests would be in trouble.

ngoldbaum:
Yeah, yes you could do it, but I think we'd generally prefer to use a dataset where you know the answer ahead of time. Like a grid of particles maybe? That makes the test easier to understand and reason about.

That said we're using the numpy RNG which is guaranteed to have a stable stream stable across numpy versions, python versions and even CPU architectures so I think in *principle* it's OK.

RicardaBeckmann:
Makes sense. I'll see what I can come up with.

RicardaBeckmann:
I tried creating my own particle ds using yt.load_particles but about 50 % of the time I'd get a segfault  so I ended up hard-coding in a couple of numbers for the fake_particle_ds.

RicardaBeckmann:
Is there a way to preview the changes to the documentation in a browser-like setting? I have edited the .rst file but I've not been able to figure out how to see the final product that it produces.

ngoldbaum:
The easiest way to do that is to do a "quick" build of the docs as described here:

http://yt-project.org/doc/developing/building_the_docs.html#building-the-docs-quick

Then navigate to the build/html folder and launch a webserver (python has one built-in):

```
$ python -m http.server
```

(on python2 it's `python -m SimpleHTTPServer`) and open http://0.0.0.0:8000 in your browser.

The docs will also build on jenkins (the "docs" build listed above) and you can navigate the built docs there once the build is finished. Look for the link to "artifacts" in the the jenkins interface, it's a little buried in the menus.

PR #1931, opened 2018-07-30
Per field config

### LINKED ISSUE #1888: Feature request: per-field configuration
### Feature request

As of today, there is no way for users to configure the defaults of the plots (colormap, unit, background color, ...) _per field_. It would be nice to extend the yt configuration file so that you can pass it some defaults for fields.

### Suggested format

```
[gas]
density_colormap = dusk
density_unit = mp/cm**3
temperature_colormap = seismic
[star]
particle_mass_colormap = viridis
```

### What should and should not be in ?

If we implement this one, we should discuss what can and cannot be part of the default configuration. For example, using the above configuration, what should be the default unit for the following line
```python
p = yt.ProjectionPlot(ds, 'x', 'density')
```
The configuration seem to tell that it should be using `mp/cm**3` but the result is homogeneous to `mp/cm**2`.

The question essentially boils down to whether we want
- a per field configuration
- a per field per plot type configuration

The second one, while more cumbersome for the user seem to be the easiest to implement. However, it doesn't help distinguish
```python
p1 = yt.ProjectionPlot(ds, 'x', 'density')
p2 = yt.ProjectionPlot(ds, 'x', 'density', weight_field='cell_mass')
```

### Suggested list of tweakable defaults

* colormap
* unit ?
* boundaries (`zlim`)
* background color

We should also make clear what strategy is applied if there are different configuration in, say the main yt config file and a local `ytrc` file (I guess the latter overrides the former?).

cphyc:
**UPDATE**: most of the text below has been invalidated since #2981 has been merged.

## PR Summary

This PR adds the possibility to configure the plots field-wise instead of globally.

This would fix #1888. 

The PR also forces the migration from pure `ini` files to the toml format. Note that this requires installing a new package (`toml`). Unfortunately, this format is not 100% compatible with the previous yt version, in particular the strings have to be surrounded by `"` or `'` and the boolean are in lower case.

Here is an example toml file with a resulting image 
```toml
[yt]
loglevel = 20
coloredlogs = true

[config.gas.temperature]
cmap = 'plasma'
units = '1e3*K'
```
```python
import yt

p = yt.SlicePlot(ds, 0, 'temperature')
p.save('/tmp/')
```
info_00080_slice_x_temperature

## Improvement to this PR

In theory, we could extend this PR to:
* selecting log/linear plot
* the unit and default extent but then we'd have to differentiate projection plots from slices in a non-trivial way

We could also keen the ini format and use keys like `config_gas_temperature` instead of `config.gas.temperature`.

## PR Checklist

- [x] New features are documented
  - [x] with docstrings ...
  - [x] ... and narrative docs
- [x] Adds tests for the change in `config.py`. 
- [x] Adds tests for the new feature.


cphyc:
Here is an updated version. 

The outcome of the following script is shown below
```python
import yt

p = yt.SlicePlot(ds, 'x', [('gas', 'density'), ('gas', 'temperature')])
p.save('1')
p.set_cmap('density', 'dusk')
p.save('2')
```
The configuration file (`yt.cfg`) is
1_slice_x_density
2_slice_x_density

```
[yt]
loglevel = 20
coloredlogs = true

[config.gas.density]
cmap = Reds

# This is ignored
[config.foo.bar]
cmap = none
```

cphyc:
For now, one can only configure whether the plot is a log plot or not and the colormap.

yt/config.py
```diff
@@ -16,8 +16,14 @@
...
 
 import os
+import re
 import warnings
+from collections import defaultdict
 from yt.extern.six.moves import configparser
+from yt.utilities.exceptions import \
+    YTFieldNotFound
+
+SECTION_RE = re.compile('^config\.(?P<ftype>\w+)\.(?P<fname>\w+)$')
```
ngoldbaum: @Xarthisius would you mind giving this regex a sanity check? I don't trust myself to :)

cphyc: FYI, the following strings
```
'config.gas.density'            # match
'config.gas_tracer.temperature' # match
'coonfig.doesnot.match'         # don't match
'config.does.not.match'         # don't match
' config.doesnot.match'         # don't match
'config.doesnot.match '         # don't match
```
But I'm no regexp guru so an extra eye won't hurt.

cphyc:
I've moved over to `toml` for the configuration. You can try setting `yt.toml` (in the local folder) to
```toml
[yt]
loglevel = 20
coloredlogs = true

[config]

[config.gas]
path_length_unit = 'mile'

[config.gas.density]
path_length_unit = 'km'
cmap = 'Reds'
units = 'mp/cm**3'
log = false

[config.gas.temperature]
units = '1e4*K'

[config.foo.bar]
cmap = 'none'
```
Here is the output. Note that the log-behavior, the units and the colormaps have been configured per field. One nice feature is that the `path_length_unit` can be set per field type or per field.
1_projection_x_density
2_projection_x_density
1_projection_x_temperature
2_projection_x_temperature

cphyc:
TODO list:
- [ ] write migration script from old config format to new `toml` format
- [ ] write documentation

cphyc:
@yt-fido test this please

neutrinoceros:
Hi @cphyc 
This change looks awesome, any chance of bringing it back to life ? or has it proved unfeasible in the end ?

cphyc:
As far as I remember, there was some thoughts to have on which format should be used (toml, yaml, ...). Apart from this it would still make a nice addition to yt 4

17 Feb 2020 11:05:23 Clément Robert <notifications@github.com>:

neutrinoceros:
I agree with @ngoldbaum that toml is probably better. I think it's more broadly used in the Python community _and_ is very close to the builtin configfile format we're currently using.

matthewturk:
I think before we decide on toml/yaml we need to be more specific about the
file organization -- the hierarchical structure of yaml is much nicer than
toml in some cases, in my opinion.  I made some attempts at field
configuration stuff that was file-based a couple months ago and think that
once we've gotten the yt4/master merge done we could re-explore this a bit.

cphyc:
Should we revive this PR for yt-4.0?

I have labeled it accordingly, but feel free to remove the label if you think this shouldn't get included.

neutrinoceros:
For what it's worth I agree that a major release feels like the right time to have such a change.
Maybe it would be worth breaking this into several steps though.
For instance, we could switch to toml (or yaml ?) config files by 4.0 but only deprecate INI files later (4.1 ?), and have a migration script/recipe to help with the conversion of existing config files in the mean time ?
note that whichever format we collectively agree to land on, we should also discuss dependency options (e.g. toml is supported by `toml`, `pytoml`). Personnally I find `toml` sometimes feels a bit unprofessional to use as it sometimes logs errors with literal "WTF??".

cphyc:
This PR shall be revived once #2981 is merged.

yt/visualization/plot_container.py
```diff
@@ -208,6 +215,11 @@ class PlotContainer:
     _plot_type = None
     _plot_valid = False
 
+    # Plot defaults
+    _colormap_config: DictWithFactory = None
+    _log_config: DictWithFactory = None
+    _units_config: DictWithFactory = None
```
neutrinoceros, line 221: you'll need to add `from typing import Optional`, this is needed to satisfy type checkers when a type is declared but the default value is `None`
```suggestion
    _colormap_config: Optional[DictWithFactory] = None
    _log_config: Optional[DictWithFactory] = None
    _units_config: Optional[DictWithFactory] = None
```

yt/visualization/plot_window.py
```diff
@@ -290,6 +296,37 @@ def _recreate_frb(self):
         # At this point the frb has the valid bounds, size, aliasing, etc.
         if old_fields is None:
             self._frb._get_data_source_fields()
+
+            # New frb, apply default units (if any)
+            for field, field_unit in self._units_config.items():
+                field_unit = Unit(field_unit, registry=self.ds.unit_registry)
+                if field_unit is not None:
```
neutrinoceros, line 303: you could save an indentation level here by writing instead
```python
if field_unit is None:
    unit = field_unit
    continue
# implicit else...
```

yt/visualization/plot_container.py
```diff
@@ -226,6 +238,24 @@ def __init__(self, data_source, figure_size, fontsize):
         self._minorticks = {}
         self._field_transform = {}
 
+        self.setup_defaults()
+
+    def setup_defaults(self):
+        def default_from_config(key, default):
```
neutrinoceros, line 244: you may want to use `functools.wraps` here

yt/funcs.py
```diff
@@ -1310,3 +1310,13 @@ def sglob(pattern):
     Return the results of a glob through the sorted() function.
     """
     return sorted(glob.glob(pattern))
+
+
+class DictWithFactory(dict):
```
neutrinoceros, line 1315: I'm not convinced a _class_ belongs in `funcs.py` (even though there are already a handful of these living here)

neutrinoceros (commented):
general comments for improvements, I'll review more thoroughly soon !

cphyc:
Status:
- [x] Support setting colormap
- [x] Support setting log scale
- [ ] Setting defaults units

cphyc:
Note: I have rebased on #3088 as I had some issues with the units in tests, but the former should be merged before this one goes in.

yt/visualization/plot_container.py
cphyc: What would be the benefit? `default_from_config` is not a decorator but is merely used to create a nice closure to the `getter` function.

neutrinoceros: sorry I misread this

yt/funcs.py
```diff
@@ -1310,3 +1311,17 @@ def sglob(pattern):
     Return the results of a glob through the sorted() function.
     """
     return sorted(glob.glob(pattern))
+
+
+def dictWithFactory(factory: Callable[[Any], Any]) -> dict:
```
neutrinoceros, line 1316: you'll need to include `from typing import Type`
```suggestion
def dictWithFactory(factory: Callable[[Any], Any]) -> Type:
```

Also I'm not very happy about using `Any` here, especially given that the function is only used with very limited set of types (at least for now). I think `Callable` might suffice as an annotation here.
Also note that we'd need to define a custom type such as
`FieldType = Union[str, Tuple[str, str]]`, which would belong in `yt._typing` (#3064)

yt/visualization/plot_container.py
```diff
@@ -226,6 +239,35 @@ def __init__(self, data_source, figure_size, fontsize):
...
+            def getter(field):
+                ftype, fname = self.data_source._determine_fields(field)[0]
+                ret = [
+                    ytcfg.get_most_specific(ftype, fname, key, fallback=default)
+                    for key, default in zip(_keys, _defaults)
+                ]
+                if len(ret) == 1:
+                    return ret[0]
+                else:
+                    return ret
```
neutrinoceros, line 258:
```suggestion
                return ret
```

yt/visualization/plot_container.py
```diff
@@ -208,6 +216,11 @@ class PlotContainer:
     _plot_type = None
     _plot_valid = False
 
+    # Plot defaults
+    _colormap_config: dict
+    _log_config: dict
+    _units_config: dict
```
neutrinoceros, line 222: this is more of note to myself for #3064 but 
```suggestion
    _colormap_config: Dict[FieldType: CMapType]
    _log_config: Dict[FieldType: bool]
    _units_config: Dict[FieldType: UnitType]
```

neutrinoceros (commented):
I have to run but here's a partial review (mostly about typing)

yt/funcs.py
cphyc: Actually `Callable[[Any], Any]` is stricter than `Callable`, since it allows only *one* argument to be passed to the factory.

cphyc:
For reference: here is the latest build of the doc: https://tests.yt-project.org/job/yt_docs_new/1025/artifact/doc/build/html/reference/configuration.html#available-per-field-configuration-options

neutrinoceros:
Apart from that one small conflict, this looks like it's ready ? I'm ok to review this again @cphyc

cphyc:
> Apart from that one small conflict, this looks like it's ready ? I'm ok to review this again @cphyc

Conflict fixed (apparently), this should be ready be reviewed

yt/visualization/tests/test_callbacks.py
```diff
@@ -876,7 +876,8 @@ def test_line_integral_convolution_callback():
 
 def test_accepts_all_fields_decorator():
     fields = ["density", "velocity_x", "pressure", "temperature"]
```
neutrinoceros, line 878: #3101 ;)

neutrinoceros (approved):
Excellent, I see no real issue with this.
I would like to suggest that the per-field options may be attached to a "plot" or "plotting" section so that they can in principle be written as

```toml
[yt]
...

[plot]
gas.temperature.cmap = "thermal"
part0.temperature.cmap = "hot"
```
If you think it's a good idea.

cphyc:
> Excellent, I see no real issue with this.
> I would like to suggest that the per-field options may be attached to a "plot" or "plotting" section so that they can in principle be written as
> 
> ```toml
> [yt]
> ...
> 
> [plot]
> gas.temperature.cmap = "thermal"
> part0.temperature.cmap = "hot"
> ```
> 
> If you think it's a good idea.

I'd be happy to have some inputs from others and eventually write a small YTEP to settle this once and for all? @matthewturk @munkm what do you think?

The questions is how to keep our settings tidy while not getting in the way of users.

neutrinoceros:
Yeah I mean I realized when reading this again that I was missing an opportunity to have parameters outside the main `[yt]` section in YTEP-0039.
I'm not sure we need a dedicated YTEP for this, but we should clearly align our decisions here and in YTEP-0039. Once we have an existing convention to respect it should be clearer what to do in the future.

brittonsmith (approved):

neutrinoceros:
@cphyc to follow up on last week's YTEP, I think we agreed to settle for a `[plot]` section. I push this PR through this last step for you if you want, let me know !

munkm:
Yes! In the triage meeting we said the [plot] part but then shortly thereafter @matthewturk replied on the mailing list that he would prefer it as field specific I think? Maybe we need to make a vote and try to figure out what people want.

matthewturk:
@munkm I am happy to formally set aside my loosely-held reservations.

munkm:
Ok! I'll review this now.

cphyc:
Updated!

doc/source/reference/configuration.rst
```diff
@@ -140,6 +142,43 @@ used internally.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is possible to customize the default behaviour of plots using per-field configuration.
+The default options for plotting a given field can be specified in the configuration file
+in ``[field_type.field_name]`` blocks. The available keys are
+
+* ``cmap`` (default: ``yt.default_colormap``, see :ref:`global-config`): the colormap to
+  use for the field.
+* ``log`` (default: ``True``): use a log scale (or symlog if linthresh is also set).
+* ``linthresh`` (default: None): the linear threshold for symlog plots.
```
munkm, line 158: To be extra explicit we should say that if this is set the plot _will_ be logged.

yt/visualization/plot_window.py
```diff
@@ -228,6 +229,14 @@ def __init__(
                 self._field_transform[field] = log_transform
             else:
                 self._field_transform[field] = linear_transform
+
+            log, linthresh = self._log_config[field]
+            if (log is not None and log is not finfo.take_log) or linthresh:
```
munkm, line 234: Maybe I need more ☕ but I'm a little confused by the part after the `and` here. Is it necessary? Should `finfo.take_log` matter if the user is overriding it?

munkm (commented):
A few minor things but this is awesome! Thank you so much!!!

yt/visualization/tests/test_plotwindow.py
```diff
@@ -406,6 +406,73 @@ def test_creation_with_width(self):
...
+        newConfig = {
+            ("yt", "default_colormap"): "viridis",
+            ("plot", "gas", "log"): False,
+            ("plot", "gas", "density", "units"): "lb/yard**3",
+            ("plot", "gas", "density", "path_length_units"): "mile",
+            ("plot", "gas", "density", "cmap"): "plasma",
+            ("plot", "gas", "temperature", "log"): True,
+            ("plot", "gas", "temperature", "linthresh"): 100,
+            ("plot", "gas", "temperature", "cmap"): "hot",
+            ("plot", "gas", "pressure", "log"): True,
```
munkm, line 425: Do we want to also add a test here where `linthresh` is set but log isn't?

doc/source/reference/configuration.rst
munkm: Oh wait, it will be symlogged if both?

neutrinoceros (approved):

doc/source/reference/configuration.rst
```diff
@@ -140,6 +142,43 @@ used internally.
...
+* ``units`` (defaults to the units of the field): the units to use to represent the field.
+* ``path_length_units`` (default: ``cm``): the unit of the integration length when doing
```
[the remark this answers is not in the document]
cphyc: Yup it should. Am making it explicit

doc/source/reference/configuration.rst
```diff
@@ -140,6 +142,43 @@ used internally.
...
+
+It is possible to customize the default behaviour of plots using per-field configuration.
+The default options for plotting a given field can be specified in the configuration file
+in ``[field_type.field_name]`` blocks. The available keys are
+
+* ``cmap`` (default: ``yt.default_colormap``, see :ref:`global-config`): the colormap to
+  use for the field.
+* ``log`` (default: ``True``): use a log scale (or symlog if linthresh is also set).
+* ``linthresh`` (default: None): the linear threshold for symlog plots.
+* ``units`` (defaults to the units of the field): the units to use to represent the field.
```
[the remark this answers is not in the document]
cphyc: At the moment you get a `UnitConversionError` whenever you make a plot. I can add some logic to catch it earlier and only raise a warning?

doc/source/reference/configuration.rst
cphyc: I have updated it. Tell me if that is clearer :)

doc/source/reference/configuration.rst
```diff
@@ -140,6 +142,45 @@ used internally.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is possible to customize the default behaviour of plots using per-field configuration.
+The default options for plotting a given field can be specified in the configuration file
+in ``[plot.field_type.field_name]`` blocks. The available keys are
+
+* ``cmap`` (default: ``yt.default_colormap``, see :ref:`global-config`): the colormap to
+  use for the field.
+* ``log`` (default: ``True``): use a log scale (or symlog if ``linthresh`` is also set).
+* ``linthresh`` (default: None): if set to a float different than ``None`` and ``log`` is
```
neutrinoceros, line 158:
```suggestion
* ``linthresh`` (default: ``None``): if set to a float different than ``None`` and ``log`` is
```

doc/source/reference/configuration.rst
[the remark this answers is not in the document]
cphyc: This is now the behaviour, what do you think?

I am catching unit conversion errors, show an error with the logger and fallback to the original units.

image

yt/visualization/plot_window.py
cphyc: You're right, the logic is broken here. I have fixed it.

yt/visualization/tests/test_plotwindow.py
cphyc: Yes we do, since doing it uncovered a bug related to your comment (https://github.com/yt-project/yt/pull/1931#discussion_r602553624)!

doc/source/reference/configuration.rst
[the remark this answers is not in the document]
munkm: I really like this update! Perhaps saying "retaining units of _____" right after the "tried converting" sentence just to make it extra clear what default units it's falling back to.

doc/source/reference/configuration.rst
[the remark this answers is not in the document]
munkm: Looks great!

yt/visualization/tests/test_plotwindow.py
munkm: !!!! Ah I'm glad my last review helped!

doc/source/reference/configuration.rst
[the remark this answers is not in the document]
munkm: That is extremely minor and I don't think is required to merge this though!

munkm (approved):
I think this is looking great! I love the way this is set up and I think you've done a nice job designing this feature. Thank you so much @cphyc

yt/frontends/ramses/fields.py
```diff
@@ -90,7 +91,7 @@ class RAMSESFieldInfo(FieldInfoContainer):
         ("x-acceleration", (ra_units, ["acceleration_x"], None)),
         ("y-acceleration", (ra_units, ["acceleration_y"], None)),
         ("z-acceleration", (ra_units, ["acceleration_z"], None)),
-        ("Potential", (ener_units, ["potential"], None)),
+        ("Potential", (spec_ener_units, ["potential"], None)),
```
munkm, line 94: Is this maybe supposed to be in yt-project/yt#3160?

cphyc: Oh, my bad! It should indeed be there.

munkm:
🎉 🎉 🎉   Thank you so much @cphyc for getting this through!

PR #1902, opened 2018-07-19
Adds ability to specify particle_type in angular momentum derived quantity

chummels:
## PR Summary

This adds the ability to specify a particle_type for the derived quantity for determining the angular momentum vector for a data object.  Currently yt will only calculate the angular momentum vector for all particles, not allow a user to specify *just* the gas, or *just* the DM particles.  This adds that functionality.  Examples in docstrings.

PR #1934, opened 2018-07-31
 use inverse matrix to do rotations for SPH off-axis projections 

ngoldbaum:
This fixes the issues @chummels described in #1902. The angular momentum vector was correct, but the rotation performed by the off-axis projection machinery was calculating the rotation matrix necessary to rotate the z-axis to the specified normal vector. However, what we wanted was the rotation matrix that does the inverse operation: rotating the the specified normal vector to the z-axis. The fix was just to add a call to `np.lingalg.inv` on the return statement for `get_rotation_matrix`. There are a couple other minor formatting changes, I also updated the tests which had encoded the incorrect answer before.


ngoldbaum:
ping @bqia0

bqia0:
I see.  Thanks a lot for finding and fixing this.

chummels:
Whoa--this is amazing!  I'll test this out and get back to you ASAP!

PR #1935, opened 2018-08-01
Use patch groups in GAMER

hyschive:
## PR Summary

This PR uses patch groups (i.e. 8 nearby patches that share the same parent patch) instead of individual patches to map to the yt grid objects in the GAMER frontend. It significantly reduces the memory overhead by reducing the total number of grid objects by a factor of 8. This improvement can be crucial since each grid object consumes ~10 kB of memory and some GAMER snapshots have as many as ~2e7 patches. So with this PR the memory overhead associated with grid objects reduces from ~200 GB to ~25 GB. The revision is also measured to be faster.

This PR slightly changes the results of analysis using the `io_cic` field, probably because the grid *boundaries* are different. So the answer tests version number is also updated. 

## PR Checklist

- [v] Code passes flake8 checker

### DIFF
--- a/yt/frontends/gamer/data_structures.py
+++ b/yt/frontends/gamer/data_structures.py
@@ -64,6 +64,8 @@ class GAMERHierarchy(GridIndex):
         self._group_particle  = ds._group_particle
         self.float_type       = 'float64' # fixed even when FLOAT8 is off
         self._particle_handle = ds._particle_handle
+        self.refine_by        = ds.refine_by
+        self.pgroup           = self.refine_by**3 # number of patches in a patch group
         GridIndex.__init__(self, ds, dataset_type)
 
     def _detect_output_fields(self):
@@ -77,21 +79,21 @@ class GAMERHierarchy(GridIndex):
 
     def _count_grids(self):
         # count the total number of patches at all levels
-        self.num_grids = self.dataset.parameters['NPatch'].sum()
+        self.num_grids = self.dataset.parameters['NPatch'].sum()/self.pgroup
 
     def _parse_index(self):
         parameters       = self.dataset.parameters
         gid0             = 0
-        grid_corner      = self._handle['Tree/Corner'].value
+        grid_corner      = self._handle['Tree/Corner'].value[::self.pgroup]
         convert2physical = self._handle['Tree/Corner'].attrs['Cvt2Phy']
 
-        self.grid_dimensions[:] = parameters['PatchSize']
+        self.grid_dimensions[:] = parameters['PatchSize']*2
 
         for lv in range(0, parameters['NLevel']):
-            num_grids_level = parameters['NPatch'][lv]
+            num_grids_level = parameters['NPatch'][lv]/self.pgroup
             if num_grids_level == 0: break
 
-            patch_scale = parameters['PatchSize']*parameters['CellScale'][lv]
+            patch_scale = parameters['PatchSize']*parameters['CellScale'][lv]*2
 
             # set the level and edge of each grid
             # (left/right_edge are YT arrays in code units)
@@ -102,7 +104,7 @@ class GAMERHierarchy(GridIndex):
                 = (grid_corner[ gid0:gid0 + num_grids_level ] + patch_scale)*convert2physical
 
             gid0 += num_grids_level
-        self.grid_left_edge += self.dataset.domain_left_edge
+        self.grid_left_edge  += self.dataset.domain_left_edge
         self.grid_right_edge += self.dataset.domain_left_edge
 
         # allocate all grid objects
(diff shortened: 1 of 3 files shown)


jzuhone:
@hyschive I will test this out hopefully in the next couple of days on some data I have. Thanks!

yt/frontends/gamer/data_structures.py
```diff
@@ -77,21 +78,21 @@ def _detect_output_fields(self):
 
     def _count_grids(self):
         # count the total number of patches at all levels
-        self.num_grids = self.dataset.parameters['NPatch'].sum()
+        self.num_grids = self.dataset.parameters['NPatch'].sum()/self.pgroup
```
jzuhone: For Python 3 this should be integer division

yt/frontends/gamer/data_structures.py
```diff
@@ -77,21 +78,21 @@ def _detect_output_fields(self):
...
-        grid_corner      = self._handle['Tree/Corner'].value
+        grid_corner      = self._handle['Tree/Corner'].value[::self.pgroup]
         convert2physical = self._handle['Tree/Corner'].attrs['Cvt2Phy']
 
-        self.grid_dimensions[:] = parameters['PatchSize']
+        self.grid_dimensions[:] = parameters['PatchSize']*self.refine_by
 
         for lv in range(0, parameters['NLevel']):
-            num_grids_level = parameters['NPatch'][lv]
+            num_grids_level = parameters['NPatch'][lv]/self.pgroup
```
jzuhone: Same here--need to use `//`

yt/frontends/gamer/data_structures.py
```diff
@@ -130,11 +132,10 @@ def _populate_grid_objects(self):
 
         for gid in range(self.num_grids):
             grid     = self.grids[gid]
-            son_gid0 = son_list[gid]
+            son_gid0 = son_list[gid*self.pgroup:(gid+1)*self.pgroup]/self.pgroup
```
jzuhone: again here--use `//`

yt/frontends/gamer/data_structures.py
```diff
@@ -174,15 +175,20 @@ def _validate_parent_children_relationship(self):
 
             # parent index is consistent with the loaded dataset
             if grid.Level > 0:
-                father_gid = father_list[grid.id]
+                father_gid = father_list[grid.id*self.pgroup]/self.pgroup
```
jzuhone: Same as above

yt/frontends/gamer/io.py
```diff
@@ -115,14 +117,34 @@ def _read_fluid_selection(self, chunks, selector, fields, size):
         mylog.debug( "Reading %s cells of %s fields in %s grids",
                      size, [f2 for f1, f2 in fields], ng )
 
+        # shortcuts
+        ps2 = self.patch_size
+        ps1 = ps2/2
```
jzuhone: Integer division here also

yt/frontends/gamer/io.py
```diff
@@ -150,13 +172,32 @@ def _read_chunk_data(self, chunk, fields):
         # fluid
         if len(fluid_fields) == 0: return rv
 
+        ps2 = self.patch_size
+        ps1 = ps2/2
```
jzuhone: Integer division here also

jzuhone:
I tried it out on a script which takes a slice and then computes the mass in a large sphere. Saw a factor of 5 speedup. Nice!

hyschive:
@jzuhone Thanks for catching this Python3 issue and confirming the results! Is there anything else you would like to test further?

jzuhone:
@xarthisius can we remove the latest GAMER answers on Jenkins and try the test again?

jzuhone:
@hyschive my student is making plots now using this, after she shows them to me I’ll report back. I also made an off-axis projection I haven’t looked at yet.

ngoldbaum:
Rather than removing the answers, the easiest way to generate new answers is just to bump the number in `tests.yaml`.

hyschive:
I thought I have updated the answer version number in `tests.ytml`. See [here](https://github.com/yt-project/yt/pull/1935/files#diff-4683557162ae75f11cc1389791432c96R29).

jzuhone:
@ngoldbaum I know that, I just don't like doing that when the answers aren't actually changing but we had some other failure.

ngoldbaum:
Sure, it's just it requires manual intervention by me or kacper. @hyschive you did but now the answers from before the python3 fix are saved in the answer test file. John wants one us to delete that file on the jenkins test box. I'm saying that requires manual intervention on the test box's filesystem and I think it's generally safer to just bump the number.

jzuhone:
@ngoldbaum @hyschive I'm fine with just bumping the number.

hyschive:
@ngoldbaum Thanks for the clarification! I just incremented the version number.

jzuhone:
@hyschive I made a change at the top, `from __future__ import division`, which should be all that we need. 

I've done some more tests and everything appears to check out wonderfully.

once the tests pass one last time I am satisfied with this. this is great!

ngoldbaum:
probably want that future import in io.py as well.

jzuhone:
Thanks for catching that, @ngoldbaum

hyschive:
@jzuhone Thanks for validating this PR!

Issue #1933, opened 2018-07-31
Creating a geometric object from a cut_region object does not respect the cut_region boundaries for particles.


RicardaBeckmann:
### Bug report

**Bug summary**
When using a cut_region object as a data_source for a (larger) geometric object, the number of particles in the final object is larger than the number of particles in the original cut_region. 

For example, if I define a cut_region that only contains 50 stars, and entirely encompass it in a ds.box object, the new box will suddenly contain 138 stars. If I make a sphere that only contains 50 stars, and encompass it in a larger box, the larger box will still only contain 50 stars.

The example code below demonstrates this in an extreme case. Using my full simulation, the number of particles I get for the "cut region in box" situation isn't necessarily equal to the total number of particles, as in the example here. Instead, it is always larger but not by a consistent factor and not by an integer either. I've seen 27.2 times the original number of particles, and 98.33 times. The new number of particles is consistent for a given set of input parameters. 

**Code for reproduction**
```
mport yt

ds=yt.load('ramses_new_format/output_00002/info_00002.txt')
ad=ds.all_data()

print("Number of particles in the whole box",len(ad['io','particle_mass']))

#Geometric region in geometric region                                                                                                
print("-- Geometric region in geometric region")
sphere=ds.sphere([0.5,0.5,0.5],ds.quan(0.2,'code_length'))
print("Number of particles in a small sphere",len(sphere['io','particle_mass']))
sphere_in_box=ds.box([0.0,0.0,0.0],[1.0,1.0,1.0],data_source=sphere)
print("Number of particles in a box based on the sphere",len(sphere_in_box['io','particle_mass']))

#Cut region in geometric region                                                                                                      
print("-- Cut region in geometric region")
cut_region=ad.cut_region("obj['density'].in_units('amu/cm**3')<1")
print("Number of particles in the cut region",len(cut_region['io','particle_mass']))
cut_region_in_box=ds.box([0.0,0.0,0.0],[1.0,1.0,1.0],data_source=cut_region)
print("Number of particles in a box based on the cut_region",len(cut_region_in_box['io','particle_mass']))
```
**Output**
```
yt : [INFO     ] 2018-07-31 18:08:20,648 Parameters: current_time              = 0.0201172518911859
yt : [INFO     ] 2018-07-31 18:08:20,648 Parameters: domain_dimensions         = [32 32 32]
yt : [INFO     ] 2018-07-31 18:08:20,649 Parameters: domain_left_edge          = [ 0.  0.  0.]
[...]

PR #1937, opened 2018-08-03
Add a data_source for the annotate_particle callback for plots

RicardaBeckmann:
## PR Summary
This pull-request allows a data_source to be passed to the annotate_particle callback, instead of producing a standardgeometric source from a given width. This allows e.g particles in a cut_region or a subsection of the plot only to be annotated. 

The width argument remains in use for all data_sources, except for cut regions. This is due to Issue #1933 

This is mostly  new feature, making the annotate_particle callback more flexible. The default behaviour of the call_back has not changed.

I have also removed the "minimum_mass" keyword argument, as I feel mass cuts ( as well as all other sub-division) should ideally be handled via the (existing) ptype argument. 

Adding extra particles to cover simulation domains outside the original box for periodic simulations was broken, and is now fixed and works correctly. 

### Before
projection_density_00020_z_kpc0010 00000_sink

### Now
projection_density_00020_z_kpc0010 00000_sink_2018_07_31

## PR Checklist


yt/visualization/plot_modifications.py
```diff
@@ -1643,7 +1645,11 @@ def __call__(self, plot):
         y0, y1 = plot.ylim
         xx0, xx1 = plot._axes.get_xlim()
         yy0, yy1 = plot._axes.get_ylim()
-        reg = self._get_region((x0,x1), (y0,y1), plot.data.axis, data)
+        if type(self.data_source)==YTCutRegion:
+            mylog.warn("Parameter 'width' is ignored in annotate_particles if the data_source is a cut_region.")
+            self.region=self.data_source
+        else:
+            self._get_region((x0,x1), (y0,y1), plot.data.axis, data)
```
RicardaBeckmann: The width is incompatible with a cut_region data_source due to issue #1933 .

Using width + a cut_region leads to plots like this one, instead of the one shown in the description of the pull request.

projection_density_00020_z_kpc0010 00000_sink_2018_07_31_noif

yt/visualization/plot_modifications.py
```diff
@@ -1643,7 +1645,11 @@ def __call__(self, plot):
         y0, y1 = plot.ylim
         xx0, xx1 = plot._axes.get_xlim()
         yy0, yy1 = plot._axes.get_ylim()
-        reg = self._get_region((x0,x1), (y0,y1), plot.data.axis, data)
+        if type(self.data_source)==YTCutRegion:
+            mylog.warn("Parameter 'width' is ignored in annotate_particles if the data_source is a cut_region.")
```
ngoldbaum: I will try to look closer at this, unfortunately I'm totally buried under other yt issues :/

Maybe add a link to the github issue tracking this problem in the error text?

yt/visualization/plot_modifications.py
```diff
@@ -1632,6 +1633,7 @@ def __init__(self, width, p_size=1.0, col='k', marker='o', stride=1,
         self.ptype = ptype
         self.minimum_mass = minimum_mass
```
ngoldbaum: So I guess we're still taking `minimum_mass` but you'd like to deprecate it in favor of using a particle filter? that seems ok to me although we need to bear in mind that we want old scripts to keep on working and producing the same result, at least until we make a major version number change to signal that we are breaking backward compatibility in places.

That's ok, we can still deprecate `minimum_mass` but to do that it will still need to keep on working the same way until the functionality is completely removed. So, if someone does pass in a `minimum_mass` not set to `None` here you'd issue a deprecation warning using `issue_deprecation_warning` in `yt.funcs` with a suggestion to use a particle filter.

yt/visualization/plot_modifications.py
```diff
@@ -1659,15 +1665,12 @@ def __call__(self, plot):
...
-                                                        reg[pt, field_y],
+        particle_x, particle_y = self._enforce_periodic(self.region[pt, field_x],
+                                                        self.region[pt, field_y],
                                                         x0, x1, period_x,
                                                         y0, y1, period_y)
         gg = ( ( particle_x >= x0 ) & ( particle_x <= x1 )
            &   ( particle_y >= y0 ) & ( particle_y <= y1 ) )
-        if self.minimum_mass is not None:
-            gg &= (reg[pt, "particle_mass"] >= self.minimum_mass)
-            if gg.sum() == 0: return
```
ngoldbaum: Since I think we only want to deprecate `minimum_mass`, not render it non-functional, this code needs to be restored.

yt/visualization/plot_modifications.py
```diff
@@ -1682,16 +1685,16 @@ def _enforce_periodic(self,
...
             particle_y = uhstack((particle_y, particle_y))
-        if self.periodic_x and x1 < self.RE[0]:
+        if self.periodic_x and x1 < self.LE[0]:
             particle_x = uhstack((particle_x, particle_x - period_x))
             particle_y = uhstack((particle_y, particle_y))
-        if self.periodic_y and y0 > self.LE[1]:
+        if self.periodic_y and y0 > self.RE[1]:
             particle_y = uhstack((particle_y, particle_y + period_y))
             particle_x = uhstack((particle_x, particle_x))
-        if self.periodic_y and y1 < self.RE[1]:
```
ngoldbaum: Why did these change? Was it wrong before?

yt/visualization/plot_modifications.py
RicardaBeckmann: Not necessarily, as self.region is set within _get_region. I've added it anyway for consistency.

yt/visualization/plot_modifications.py
RicardaBeckmann: Yes. With the current set of logical statements, particles were duplicated when the domain to be plotted was entirely contained within the simulation box. With the fix, particles are only duplicated if the domain crosses the boundary of a periodic box, as intended. They were simply the wrong way around.

RicardaBeckmann:
@ngoldbaum Done!

PR #1942, opened 2018-08-06
Remove the test of making slices through raw fields 

git-abhishek:
## PR Summary

In my opinion, something similar has been tested in the PR #1941 but with fake dataset.
I do not think it is testing some different corner case or is a test of possible bug fix.
Furthermore, I feel there is no need to test this with real dataset.

*Please feel free to reject this PR if I have misunderstood this due to limited knowledge!*

## PR Checklist

## Adding Reviewers


atmyers:
I believe that this tests a different code path than the mesh slices - nodal fields in the context of a structured mesh. Also, I kind of like having the real dataset in this case, since it also verifies that the WarpX frontend is working correctly.

git-abhishek:
Since, this test requires the real dataset, closing this issue. Please see #1967 which is a feature to download small test dataset on Travis which was conceived due to this PR.

Issue #1906, opened 2018-07-20
OffAxisProjection only projects in code units


chummels:
### Bug report

**Bug summary**

The addition of the `OffAxisProjectionPlot` for use with particle-based datasets is great and a wonderful contribution to the code, but I've discovered a few small issues.  One of which is while `ProjectionPlot` gives results in CGS, `OffAxisProjectionPlot` seems to only give results in code units for particle-based datasets.  

**Code for reproduction**

```python
import yt
ds = yt.load('FIRE/FIRE_M12i_ref12/snapshot_600.hdf5')
yt.OffAxisProjectionPlot(ds, [1,0,0], ('gas', 'density')).save()
```

**Actual outcome**
![snapshot_600_offaxisprojection_density](https://user-images.githubusercontent.com/8250999/42977375-a5bb01ca-8b7b-11e8-974e-4bb2e5675589.png)

**Expected outcome**

Something like a `ProjectionPlot` like this one where the z field is in CGS:

![snapshot_600_projection_x_density](https://user-images.githubusercontent.com/8250999/42977409-e1ad609c-8b7b-11e8-83d3-d1dba1b90d4e.png)

chummels:
Fixed in #1961

PR #1945, opened 2018-08-06
[bugfix] Making north vector work with SPH off axis projections; fixes #1907

### LINKED ISSUE #1907: North vector does not work in OffAxisProjections
### Bug report

**Bug summary**

The addition of the OffAxisProjectionPlot for use with particle-based datasets is great and a wonderful contribution to the code, but I've discovered a few small issues. One of which is that the `north_vector` doesn't ever get used in the code, so one just gets an arbitrary north_vector applied no matter what.

**Code for reproduction**

```python
import yt
import numpy as np
ds = yt.load('FIRE/FIRE_M12i_ref11/snapshot_600.hdf5')
x = np.array([1,0,0])
y = np.array([0,1,0])
z = np.array([0,0,1])
p = yt.OffAxisProjectionPlot(ds, x, ('gas', 'density'))
p.save('OAP.png')
p = yt.OffAxisProjectionPlot(ds, x, ('gas', 'density'), north_vector=y)
p.save('OAP_north_y.png')
p = yt.OffAxisProjectionPlot(ds, x, ('gas', 'density'), north_vector=z)
p.save('OAP_north_z.png')
```

**Actual outcome**
All three images are the same, and there is no rotation seen in the image, despite the change in `north_vector`:

oap

chummels:
## PR Summary

This implements the `north_vector` keyword for `off_axis_projection` and `OffAxisProjectionPlot` within the demeshening.  Had to apply a second rotation to the data to rotate the north-vector to the +y direction but now it works.  This closes issue #1907 .

As a demonstration, this script gives the following results showing that on-axis and off-axis projections are identical when using the appropriate axes:

```
import yt
import numpy as np
fn = 'GadgetDiskGalaxy/snapshot_200.hdf5'
ds = yt.load(fn)
_, c = ds.find_max('density')

x = ds.arr([1,0,0], 'unitary')
y = ds.arr([0,1,0], 'unitary')
z = ds.arr([0,0,1], 'unitary')

yt.ProjectionPlot(ds, 'x', ('gas', 'density'), center=c, width=(200, 'kpc')).save('OnAxis_x.png')
yt.ProjectionPlot(ds, 'y', ('gas', 'density'), center=c, width=(200, 'kpc')).save('OnAxis_y.png')
yt.ProjectionPlot(ds, 'z', ('gas', 'density'), center=c, width=(200, 'kpc')).save('OnAxis_z.png')
yt.OffAxisProjectionPlot(ds, x, ('gas', 'density'), north_vector=z, center=c, width=(200, 'kpc'))save('OffAxis_x.png')
yt.OffAxisProjectionPlot(ds, y, ('gas', 'density'), north_vector=x, center=c, width=(200, 'kpc'))save('OffAxis_y.png')
yt.OffAxisProjectionPlot(ds, z, ('gas', 'density'), north_vector=y, center=c, width=(200, 'kpc'))save('OffAxis_z.png')
```

OnAxis_x.png:
onaxis_x

OffAxis_x.png:
offaxis_x

OnAxis_y.png:
onaxis_y

OffAxis_y.png:
offaxis_y

OnAxis_z.png:
onaxis_z

OffAxis_z.png:
offaxis_z

The units are still wrong, but that's issue #1906, and will be addressed in another PR.

### DIFF
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -1473,12 +1473,17 @@ def off_axis_projection_SPH(np.float64_t[:] px,
     if np.allclose(normal_vector, np.array([0., 0., 0.]), rtol=1e-09):
         return
 
+    # We want to do two rotations, one to first rotate our coordinates to have
+    # the normal vector be the z-axis (i.e., the viewer's perspective), and then
+    # another rotation to make the north-vector be the y-axis (i.e., north).
+    # Fortunately, total_rotation_matrix = rotation_matrix_1 x rotation_matrix_2
     cdef int num_particles = np.size(px)
     cdef np.float64_t[:] z_axis = np.array([0., 0., 1.], dtype='float_')
     cdef np.float64_t[:] y_axis = np.array([0., 1., 0.], dtype='float_')
-    cdef np.float64_t[:, :] rotation_normal_matrix = get_rotation_matrix(normal_vector, z_axis)
-    cdef np.float64_t[:, :] rotation_north_matrix = get_rotation_matrix(north_vector, y_axis)
-    cdef np.float64_t[:, :] rotation_matrix = np.matmul(rotation_normal_matrix, rotation_north_matrix)
+    cdef np.float64_t[:, :] normal_rotation_matrix = get_rotation_matrix(normal_vector, z_axis)
+    cdef np.float64_t[:] transformed_north_vector = np.matmul(normal_rotation_matrix, north_vector)
+    cdef np.float64_t[:, :] north_rotation_matrix = get_rotation_matrix(transformed_north_vector, y_axis)
+    cdef np.float64_t[:, :] rotation_matrix = np.matmul(north_rotation_matrix, normal_rotation_matrix)
     cdef np.float64_t[:] px_rotated = np.empty(num_particles, dtype='float_')
     cdef np.float64_t[:] py_rotated = np.empty(num_particles, dtype='float_')
 
@@ -1547,9 +1552,10 @@ cpdef np.float64_t[:, :] get_rotation_matrix(normal_vector, final_vector):
     # identity matrix
     if np.isclose(c, 1, rtol=1e-09):
         return np.identity(3, dtype='float_')
-    # if the normal vector is the negative z-axis, return:
-    #if np.isclose(s, 0, rtol=1e-09):
-    #    return np.array([[0, -1, 0],[-1, 0, 0],[0, 0, -1]], dtype='float_')
+    # if the normal vector is the negative final vector, return the negative
+    # identity matrix:
+    if np.isclose(s, 0, rtol=1e-09):
+        return -np.identity(3, dtype='float_')
 
     cdef np.float64_t[:, :] cross_product_matrix = np.array([[0, -1 * v[2], v[1]],
                                                       [v[2], 0, -1 * v[0]],
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -1530,7 +1530,7 @@ cdef np.float64_t[:] rotation_matmul(np.float64_t[:, :] rotation_matrix,
 @cython.wraparound(False)
 cpdef np.float64_t[:, :] get_rotation_matrix(normal_vector, final_vector):
     """ Returns a numpy rotation matrix corresponding to the
-    rotation of the given normal vector to the z-axis ([0, 0, 1]).
+    rotation of the given normal vector to the specified final_vector.
     See https://math.stackexchange.com/a/476311 although note we return the
     inverse of what's specified there.
     """
@@ -1539,16 +1539,17 @@ cpdef np.float64_t[:, :] get_rotation_matrix(normal_vector, final_vector):
     cdef np.float64_t[:] final_vector_np = np.array([final_vector[0], final_vector[1], final_vector[2]], 
                                                      dtype='float_')
     cdef np.float64_t[:] normal_unit_vector = normal_vector_np / np.linalg.norm(normal_vector_np)
-    cdef np.float64_t[:] v = np.cross(final_vector, normal_unit_vector)
+    cdef np.float64_t[:] final_unit_vector = final_vector_np / np.linalg.norm(final_vector_np)
+    cdef np.float64_t[:] v = np.cross(final_unit_vector, normal_unit_vector)
     cdef np.float64_t s = np.linalg.norm(v)
-    cdef np.float64_t c = np.dot(final_vector, normal_unit_vector)
-    # if the normal vector is identical to the z-axis, just return the
+    cdef np.float64_t c = np.dot(final_unit_vector, normal_unit_vector)
+    # if the normal vector is identical to the final vector, just return the
     # identity matrix
     if np.isclose(c, 1, rtol=1e-09):
         return np.identity(3, dtype='float_')
     # if the normal vector is the negative z-axis, return:
-    if np.isclose(s, 0, rtol=1e-09):
-        return np.array([[0, -1, 0],[-1, 0, 0],[0, 0, -1]], dtype='float_')
+    #if np.isclose(s, 0, rtol=1e-09):
+    #    return np.array([[0, -1, 0],[-1, 0, 0],[0, 0, -1]], dtype='float_')
 
     cdef np.float64_t[:, :] cross_product_matrix = np.array([[0, -1 * v[2], v[1]],
                                                       [v[2], 0, -1 * v[0]],
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -1480,19 +1480,23 @@ def off_axis_projection_SPH(np.float64_t[:] px,
     cdef int num_particles = np.size(px)
     cdef np.float64_t[:] z_axis = np.array([0., 0., 1.], dtype='float_')
     cdef np.float64_t[:] y_axis = np.array([0., 1., 0.], dtype='float_')
-    cdef np.float64_t[:, :] normal_rotation_matrix = get_rotation_matrix(normal_vector, z_axis)
-    cdef np.float64_t[:] transformed_north_vector = np.matmul(normal_rotation_matrix, north_vector)
-    cdef np.float64_t[:, :] north_rotation_matrix = get_rotation_matrix(transformed_north_vector, y_axis)
-    cdef np.float64_t[:, :] rotation_matrix = np.matmul(north_rotation_matrix, normal_rotation_matrix)
+    cdef np.float64_t[:, :] normal_rotation_matrix
+    cdef np.float64_t[:] transformed_north_vector
+    cdef np.float64_t[:, :] north_rotation_matrix
+    cdef np.float64_t[:, :] rotation_matrix
+
+    normal_rotation_matrix = get_rotation_matrix(normal_vector, z_axis)
+    transformed_north_vector = np.matmul(normal_rotation_matrix, north_vector)
+    north_rotation_matrix = get_rotation_matrix(transformed_north_vector, y_axis)
+    rotation_matrix = np.matmul(north_rotation_matrix, normal_rotation_matrix)
+
     cdef np.float64_t[:] px_rotated = np.empty(num_particles, dtype='float_')
     cdef np.float64_t[:] py_rotated = np.empty(num_particles, dtype='float_')
-
     cdef np.float64_t[:] coordinate_matrix = np.empty(3, dtype='float_')
     cdef np.float64_t[:] rotated_coordinates
     cdef np.float64_t[:] rotated_center
     rotated_center = rotation_matmul(
-        rotation_matrix,
-        np.array([center[0], center[1], center[2]]))
+        rotation_matrix, np.array([center[0], center[1], center[2]]))
 
     # set up the rotated bounds
     cdef np.float64_t rot_bounds_x0 = rotated_center[0] - width[0] / 2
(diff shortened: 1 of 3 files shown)


yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1467,21 +1467,32 @@ def off_axis_projection_SPH(np.float64_t[:] px,
...
+    # another rotation to make the north-vector be the y-axis (i.e., north).
+    # Fortunately, total_rotation_matrix = rotation_matrix_1 x rotation_matrix_2
     cdef int num_particles = np.size(px)
-    cdef np.float64_t[:, :] rotation_matrix = get_rotation_matrix(normal_vector)
+    cdef np.float64_t[:] z_axis = np.array([0., 0., 1.], dtype='float_')
+    cdef np.float64_t[:] y_axis = np.array([0., 1., 0.], dtype='float_')
+    cdef np.float64_t[:, :] normal_rotation_matrix = get_rotation_matrix(normal_vector, z_axis)
+    cdef np.float64_t[:] transformed_north_vector = np.matmul(normal_rotation_matrix, north_vector)
+    cdef np.float64_t[:, :] north_rotation_matrix = get_rotation_matrix(transformed_north_vector, y_axis)
+    cdef np.float64_t[:, :] rotation_matrix = np.matmul(north_rotation_matrix, normal_rotation_matrix)
```
AshKelly: Could we move these definitions to look more like this?
```python
cdef np.float64_t[:, :] normal_rotation_matrix
cdef np.float64_t[:] transformed_north_vector
cdef np.float64_t[:, :] north_rotation_matrix
cdef np.float64_t[:, :] rotation_matrix

normal_rotation_matrix = get_rotation_matrix(normal_vector, z_axis)
transformed_north_vector = np.matmul(normal_rotation_matrix, north_vector)
north_rotation_matrix = get_rotation_matrix(transformed_north_vector, y_axis)
rotation_matrix = np.matmul(north_rotation_matrix, normal_rotation_matrix)
```
I think it makes these lines a little easier on the eye

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1520,29 +1531,31 @@ cdef np.float64_t[:] rotation_matmul(np.float64_t[:, :] rotation_matrix,
...
-    rotation of the given normal vector to the z-axis ([0, 0, 1]).
+    rotation of the given normal vector to the specified final_vector.
     See https://math.stackexchange.com/a/476311 although note we return the
     inverse of what's specified there.
     """
     cdef np.float64_t[:] normal_vector_np = np.array([normal_vector[0], normal_vector[1], normal_vector[2]], 
                                                      dtype='float_')
-    cdef np.float64_t[:] z_axis = np.array([0., 0., 1.], dtype='float_')
+    cdef np.float64_t[:] final_vector_np = np.array([final_vector[0], final_vector[1], final_vector[2]], 
+                                                     dtype='float_')
```
AshKelly: I don't think we need these intermediate `_np`'s, I think we can do this at the input level:
```python
cpdef np.float64_t[:, :] get_rotation_matrix(np.float64_t[:] normal_vector, np.float64_t[:] final_vector):
```

yt/visualization/volume_rendering/off_axis_projection.py
```diff
@@ -233,7 +235,8 @@ def off_axis_projection(data_source, center, normal_vector,
                     width.to('code_length').d,
                     chunk[weight].to(wounits),
                     weight_buff,
-                    normal_vector)
+                    normal_vector,
+                    north)
 
             buf /= weight_buff
```
AshKelly: I know this isn't your line, but I've just realised that this line could return NaN's, particularly with the scatter approach. I actually added some utility functions to the pixelization routines for doing this.

You should be able to do 
```python
from yt.utilities.lib.pixelization_routines import normalization_2d_utility

...

normalization_2d_utility(buf, weight_buff)
```
These utility functions just take two 2D memoryview's and check if the denominator is not zero and then do the division.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1520,29 +1531,31 @@ cdef np.float64_t[:] rotation_matmul(np.float64_t[:, :] rotation_matrix,
...
+    # if the normal vector is identical to the final vector, just return the
     # identity matrix
     if np.isclose(c, 1, rtol=1e-09):
         return np.identity(3, dtype='float_')
-    # if the normal vector is the negative z-axis, return:
+    # if the normal vector is the negative final vector, return the negative
+    # identity matrix:
     if np.isclose(s, 0, rtol=1e-09):
-        return np.array([[0, -1, 0],[-1, 0, 0],[0, 0, -1]], dtype='float_')
+        return -np.identity(3, dtype='float_')
```
ngoldbaum: Why change the answer here?

ngoldbaum:
Thanks for digging into this! Odd that jenkins triggered on this, it shouldn't have. Also not sure why travis didn't trigger. I think just pushing another commit will re-trigger all the webhooks and you should be good to go.

AshKelly:
I think this happens when you set the base branch as master and switch to yt-4.0 afterwards. I've had this happen.

yt/utilities/lib/pixelization_routines.pyx
chummels: This method was originally defined to *only* calculate the rotation matrix for rotating to the z-axis.  I generalized it to work with any arbitrary vector (see the comment two lines up).  Thus, it seemed wrong to leave this manually calculated rotation matrix which was tied to a hard-coded z-axis.

yt/utilities/lib/pixelization_routines.pyx
chummels: I was just following the code style that you originally wrote (line 1476), but sure, I can change it.

AshKelly: I see! None of this code is mine actually, I haven't really worked on this code. This is another summer student that Nathan had who implemented all of the cool projection stuff.

yt/utilities/lib/pixelization_routines.pyx
chummels: I'm confused.  You don't think I should be doing all of the intermediate normalization of these vectors in the few lines below this?

yt/visualization/volume_rendering/off_axis_projection.py
chummels: Which line are you talking about?  Line 239/242?  And you want me to include the added code just prior to that line?

yt/utilities/lib/pixelization_routines.pyx
ngoldbaum: But the original answer was chosen because that's the limiting case when `s` goes to zero. Unless I'm misunderstanding I don't think you changed the math in a way that would cause the limiting case to go to the negative identity matrix instead of `[[0, -1, 0], [-1, 0, 0], [0, 0, -1]]`. Am I misunderstanding?

yt/utilities/lib/pixelization_routines.pyx
AshKelly: Oh, sorry! No, it's the conversion to a memoryview,

```python
cdef np.float64_t[:] normal_vector_np = np.array([normal_vector[0], normal_vector[1], normal_vector[2]], dtype='float_')
```
If you change the function declaration to 
```python
cpdef np.float64_t[:, :] get_rotation_matrix(np.float64_t[:] normal_vector, np.float64_t[:] final_vector)
```
I don't _think_ you need to do that intermediate conversion step because you are passing in a  numpy array. I think with the above function declaration you can just use the `normal_vector` in place of `normal_vector_np`.

yt/visualization/volume_rendering/off_axis_projection.py
AshKelly: so the current line `buf /= weight_buff` (L241) which I know isn't yours can lead to NaN's. But if you include `from yt.utilities.lib.pixelization_routines import normalization_2d_utility` at the top of the file then that line can be replaced with `normalization_2d_utility(buf, weight_buff)` which is much more NaN safe

yt/utilities/lib/pixelization_routines.pyx
chummels: In the old version of the code that was making a rotation matrix to rotate to the hard-coded z-axis, the conditional `np.isclose(s, 0, rtol=1e-09)` means something different.  If you just look at the old code, there are two hard-coded conditionals.  The first (line1542/1554) is looking at whether or not the input `normal_vector` is the positive z-axis.  If so, it sets the rotation matrix to be the identity matrix.  Since I'm removing the hard-coded z-axis from this function and generalizing it to rotate to any `final_vector`, I have changed this first check, to check to see if the `normal_vector` is the `final_vector` and then return the identity matrix.

In the second check, the one that your comment is regarding (line1544/1557) it was originally testing whether the `normal_vector` was the negative z-axis.  Again, I'm removing the hard-coded z-axis constraint and changing it to work with the `final_vector`. I'm not sure why the arbitrary rotation matrix was given `[[0, -1, 0], [-1, 0, 0], [0, 0, -1]]`.  But it seemed appropriate that for when the two vectors were anti-aligned to give the negative identity matrix, since we're giving the identity matrix when the two vectors are aligned and the same.  But maybe my linear algebra is wrong here.  

Do you know why the arbitrary rotation matrix was provided and why it should remain?  I honestly do not.

ngoldbaum: Take a look at this example:

```
In [4]: np.array(get_rotation_matrix([0, 0, -1], [0, 0, 1]))
Out[4]:
array([[-1., -0., -0.],
       [-0., -1., -0.],
       [-0., -0., -1.]])

In [5]: np.array(get_rotation_matrix([1e-6, 1e-6, -1], [0, 0, 1]))
Out[5]:
array([[ 1.33126297e-04, -9.99866874e-01, -9.99467566e-07],
       [-9.99866874e-01,  1.33126297e-04, -9.99467566e-07],
       [ 9.99467566e-07,  9.99467566e-07, -9.99733747e-01]])
```

In the second case I'm choosing a `normal_vector` that is very close to `[0, 0, -1]` but not exactly equal. Note how your change is causing there to be a discontinuity in the return value of this function as we get close to `[0, 0, -1]`.

chummels: OK, that's fair.  I didn't check close-by cases like this to figure out what the continuous behavior was.  I still don't know *why* things tend toward this arbitrary matrix (which I thought was tied to the z-axis hardcoding), but that's OK.  You've demonstrated that I should leave it as is.  This was the only part of my PR I wasn't 100% on, so thanks for sussing it out.  Code review for the win!

yt/utilities/lib/pixelization_routines.pyx
chummels: The code definitely looks better when I make your suggestion, as it is more succinct, but unfortunately, cython gives a compile error, so I don't think this is a feasible change.

AshKelly: oh fair enough. I did quickly test the following locally,

```python
@cython.boundscheck(False)
@cython.wraparound(False)
cpdef np.float64_t[:, :] get_rotation_matrix(np.float64_t[:] normal_vector,
                                             np.float64_t[:] final_vector):
    """ Returns a numpy rotation matrix corresponding to the
    rotation of the given normal vector to the specified final_vector.
    See https://math.stackexchange.com/a/476311 although note we return the
    inverse of what's specified there.
    """
    cdef np.float64_t[:] normal_unit_vector = normal_vector / np.linalg.norm(normal_vector)
    cdef np.float64_t[:] final_unit_vector = final_vector / np.linalg.norm(final_vector)
```

and didn't get any errors.

chummels: Weird, it works this time.  I don't know what I did wrong last time.  OK, thanks.

AshKelly:
It looks like the failed tests are related to these changes. I think it's because you've changed the input to the function `off_axis_projection` and the tests needs to reflect that.

The tests are in `yt/visualization/volume_rendering/tests/test_off_axis_SPH.py`

chummels:
I made the changes discussed in slack, allowing for `'gas'` fields to be used in addition to explicit particle fields with OAP (for working with trident-generated ion fields).  I also added a test that includes the `north_vector` functionality implemented.

Unfortunately, I'm getting two remaining testing errors with some of the rotation transformations that were originally implemented with off_axis_projections.  As far as I can tell, my rotations are sound and have held up to all of my internal tests, but something is awry.  Will dig into this more.

yt/visualization/volume_rendering/tests/test_off_axis_SPH.py
```diff
@@ -18,7 +18,7 @@ def test_no_rotation():
     image buffer if processed directly through 
     pixelize_sph_kernel_projection
     """
-    normal_vector = [0., 0., 1.]
+    normal_vector = np.array([0., 0., 1.])
```
ngoldbaum: This should be handled inside of the `yt.off_axis_projection`, I think. It just needs a line at the top that does `normal_vector = np.array(normal_vector, dtype='float64')` and then users will still be able to pass in lists.

chummels: Oh, good call.  That makes it cleaner.

chummels:
The test failures were due to the assumption that specifying a start_vector and final_vector fully defined a rotation, which it does not.  There is still a rotation about the `normal_vector` that is allowed by such a transformation.  By fully specifying the rotation solution with a `north_vector`, the failures went away.  I think this PR is good to go, but let's way to make sure the tests pass.

yt/visualization/volume_rendering/off_axis_projection.py
```diff
@@ -132,11 +138,13 @@ def off_axis_projection(data_source, center, normal_vector,
 
         raise_error = False
 
+        # Assure that the field we're trying to off-axis project is a
+        # particle field or a 'gas' field.
```
ngoldbaum: super minor nit: we're not generically checking for particle fields, we're checking if the field type is the SPH particle type or if the field is an alias to an SPH field.

ngoldbaum:
Thanks so mucb for your efforts on this Cameron.

Issue #1948, opened 2018-08-08
boxlib frontend does not work for pure hydro runs 


wolfram-schmidt:
### Bug report

I spotted a problem with the boxlib/amrex frontend in the case of pure hydro runs. The problem is that the frontend by default expects particles. A while ago Mike Zingale fixed by putting if ("particles" in self.ds.parameters) before the read particle call, see https://bitbucket.org/yt_analysis/yt/pull-requests/2571/add-support-for-nyx-datasets-without/diff. Now, after compiling yt from the development branch and doing a test run with the latest Nyx/AMRex, it appears that this does not work any longer, as the error trace reports:

**Code for reproduction**

```python
import yt
PATH = "/gfs1/work/hzfbhsws/DrivenTurbulence/test-omp4/"
ds = yt.load(PATH + "plt00461")
ds.field_list
```

Can provide data set or Nyx parameter files if required.

**Actual outcome**

```
--------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-7-3fabb4ffc7bc> in <module>()
----> 1 ds.field_list

/home/b/hzfbhsws/yt-src/yt/data_objects/static_output.py in field_list(self)
    546     @property
    547     def field_list(self):
--> 548         return self.index.field_list
    549 
    550     def create_field_info(self):

/home/b/hzfbhsws/yt-src/yt/data_objects/static_output.py in index(self)
    504                 raise RuntimeError("You should not instantiate Dataset.")
    505             self._instantiated_index = self._index_class(
--> 506                 self, dataset_type=self.dataset_type)
    507             # Now we do things that we need an instantiated index for
    508             # ...first off, we create our field_info now.

/home/b/hzfbhsws/yt-src/yt/frontends/boxlib/data_structures.py in __init__(self, ds, dataset_type)
   1228 
   1229             self._read_particles("DM", is_checkpoint, 
-> 1230                                  nyx_extra_real_fields[0:self.ds.dimensionality+1])
   1231 
   1232 

/home/b/hzfbhsws/yt-src/yt/frontends/boxlib/data_structures.py in _read_particles(self, directory_name, is_checkpoint, extra_field_names)
    563 
    564     def _read_particles(self, directory_name, is_checkpoint, extra_field_names=None):
--> 565         pheader = self._determine_particle_output_type(directory_name)
    566         self.particle_headers[directory_name] = pheader(self.ds,
    567                                                         directory_name,

[...]

PR #1950, opened 2018-08-08
Fix Issue #1948

atmyers:
The Nyx frontend was using the presence of a `DM` subdirectory to determine whether particles exist or not. Apparently, this always gets created in Nyx, whether or not there are any particles present. However, the file `DM/Header` should only exist if particles are actually in the run.


ngoldbaum:
ping @wolfram-schmidt

PR #1957, opened 2018-08-09
Use set() instead of list, since same ans test could have iterations

git-abhishek:
## PR Summary

Use set() instead of list for `missing_answers` and `failed_answers`, since same anwer test could have iterations.
eg:
```python
    for lens_type in ['plane-parallel', 'perspective']:
        frame = 0

        cam = sc.add_camera(ds, lens_type=lens_type)
        cam.resolution = (1000, 1000)
        cam.position = ds.arr(np.array([-4., 0., 0.]), 'code_length')
        cam.switch_orientation(normal_vector=[1., 0., 0.],
                               north_vector=[0., 0., 1.])
        cam.set_width(ds.domain_width*2.)
        test1_prefix = '%s_%04d' % (lens_type, frame)
        test1 = VRImageComparisonTest(sc, ds, test1_prefix, decimals)
        test1.prefix = test1_prefix
        test1.answer_name = test_name
        yield test1
```

In this if no answer is present in the answer store, we will get 2 error messages but we only want it once.
## PR Checklist

## Adding Reviewers

### DIFF
--- a/tests/report_failed_answers.py
+++ b/tests/report_failed_answers.py
@@ -314,25 +314,25 @@ def parse_nose_xml(nose_xml):
         eg. ['yt.visualization.tests.test_line_plots:test_multi_line_plot']
 
     """
-    missing_answers = []
-    failed_answers = []
+    missing_answers = set()
+    failed_answers = set()
     missing_errors = ["No such file or directory",
                       "There is no old answer available"]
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Makes sense!

PR #1959, opened 2018-08-09
[bugfix] Make sure that units are respected properly for particle annotations

jzuhone:
I had a `SlicePlot` where I set the `center` and `width` arguments by hand with a specific unit, and then tried to set the `width` parameter in `annotate_particles` using `code_length`:

```python
slc = yt.SlicePlot(ds, "x", ['density'], width=(1.5, "Mpc"),
                             center=ds.arr([0.0, 0.0, 0.2], "Mpc"), origin='native')
slc.annotate_particles(3.0856e24, col='k', stride=100)
```

Two things happened here which prevented me from seeing the particles:

1. `annotate_particles` now assumes that the thing you pass in has units, either as a `YTQuantity` or as a `(value, unit)` tuple, so this fixes it so if you pass in a float it assumes `code_length`. 
2. If one mixes different types of length units in the current code, you end up somewhere down the line with combining `YTQuantity` instances in a list and then wrapping that in `YTArray` (in the `ds.region` code), and that happily clobbers the differences between the units. The solution is to make sure everything is converted to `code_length` beforehand.

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -48,7 +48,6 @@ from yt.geometry.unstructured_mesh_handler import UnstructuredIndex
 from yt.utilities.lib.mesh_triangulation import triangulate_indices
 from yt.utilities.exceptions import \
     YTDataTypeUnsupported
-
 callback_registry = {}
 
 def _verify_geometry(func):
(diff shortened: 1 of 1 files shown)


jzuhone:
By the way, I recognize that this code is somewhat ugly--happy to have suggestions as to make it better.

ngoldbaum:
No worries, not ugly, I've had to make similar fixes elsewhere in the plot callbacks. I agree with Corentin that a test would be nice but merging now because it's a clear improvement over the status quo.

PR #1960, opened 2018-08-09
Art Frontend Fix

sflarkin:
Two changes to fix problems with ART frontend. First changes an int check, while the other removes an unnecsscary check for files that causes load issues.

### DIFF
--- a/yt/frontends/art/data_structures.py
+++ b/yt/frontends/art/data_structures.py
@@ -70,7 +70,7 @@ class ARTIndex(OctreeIndex):
         """
         # Overloaded
         ds = self.dataset
-        return (1.0/ds.domain_dimensions.astype('f8') /
+        return (ds.domain_width /ds.domain_dimensions /
                 (2**self.max_level)).min()
 
     def _initialize_oct_handler(self):
--- a/yt/frontends/art/data_structures.py
+++ b/yt/frontends/art/data_structures.py
@@ -203,9 +203,6 @@ class ARTDataset(Dataset):
                 if possible.endswith(numericstr+suffix):
                     if os.path.basename(possible).startswith(prefix):
                         match = possible
-                elif possible.endswith(suffix):
-                    if os.path.basename(possible).startswith(prefix):
-                        match = possible
             if match is not None:
                 mylog.info('discovered %s:%s', filetype, match)
                 setattr(self, "_file_"+filetype, match)
--- a/yt/frontends/art/io.py
+++ b/yt/frontends/art/io.py
@@ -448,7 +448,7 @@ def read_particles(file, Nrow, idxa, idxb, fields):
     words = 6  # words (reals) per particle: x,y,z,vx,vy,vz
     real_size = 4  # for file_particle_data; not always true?
     np_per_page = Nrow**2  # defined in ART a_setup.h, # of particles/page
-    num_pages = os.path.getsize(file)/(real_size*words*np_per_page)
+    num_pages = os.path.getsize(file)//(real_size*words*np_per_page)
     fh = open(file, 'r')
     skip, count = idxa, idxb - idxa
     kwargs = dict(words=words, real_size=real_size,


ngoldbaum:
I haven’t heard back from Kenza so I think I’m just going to merge this. These are pretty clearly improvements for real-world users.

ngoldbaum:
Thanks for the contribution Sean! Looking forward to hearing more, please keep sending in improvements for the art frontend, rockstar, and yt in general. :tada: :shipit: :100: :fireworks:

sflarkin:
No prob, I have another fix for the get_smallest_dx() having no units, but my test runs keep core dumping, will let you know when I have a test complete.

ngoldbaum:
That change was included in the PR because you pushed it to your `master` branch on github. In the future you probably want to use feature branches so work doesn't get combined like that. See http://yt-project.org/doc/developing/developing.html#making-and-sharing-changes for a very detailed example workflow.

ngoldbaum:
Also your tests are core dumping? That doesn't sound good! Would love to get more info about that if you think it might be a yt issue.

sflarkin:
Oh whoopse, still newish to github, will do that in the future.

The core dump happens about 3 min after I submit a rockstar job. I can get you a copy of they system log in a bit if that might be helpful. Based on the timing, it seems to be occuring right when rockstar is set to run after I apply a particle filter.

sflarkin:
Sorry this is taking so long, I tried to run again with faulthandler, and now my total_particles finder isnt working. Not sure why.

ngoldbaum:
No worries and I'm not waiting on you or anything. Just trying to help :)

sflarkin:
Yeah, Im having this error, which I cannot figure out. Seeing as the only thing that changed recently in my repository is the adding units to get_smallest, that change might want to be reverted, as this probably impacts more than art.

    total_particles = ad.quantities.total_quantity((particle_type, "particle_ones"))
  File "/home1/sflarkin/build/yt/yt/data_objects/derived_quantities.py", line 176, in __call__
    rv = super(TotalQuantity, self).__call__(fields)
  File "/home1/sflarkin/build/yt/yt/data_objects/derived_quantities.py", line 66, in __call__
    for sto, ds in parallel_objects(chunks, -1, storage = storage):
  File "/home1/sflarkin/build/yt/yt/utilities/parallel_tools/parallel_analysis_interface.py", line 521, in parallel_objects
    storage.update(new_storage)
P019 yt : [ERROR    ] 2018-08-15 12:46:12,357 TypeError: 'int' object is not iterable
P019 yt : [ERROR    ] 2018-08-15 12:46:12,357 Error occured on rank 19.
MPT ERROR: Rank 17(g:17) is aborting with error code 1.
        Process ID: 47003, Host: r635i7n12, Program: /home1/sflarkin/build/miniconda3/envs/yt/bin/python3.6

ngoldbaum:
Ok, you shoukd see if reverting that fixes this issue. I could also help more if you share a script that triggers this issue, preferably the script would use one of the piblic test datasets on yt-project.org/data so I can reproduce it locally and devug. You’re also welcome to try to devug it yourself, I’d start by trying to figure out why we’re treating and int as an interavle where it’s failing (presumably an int is getting passed somewere where instead we should be passing something else)

sflarkin:
Just tried reverting and the error still persists. Now I have no idea what broke this.

ngoldbaum:
I'm not able to trigger the crash you saw, everything works fine when I run `runrockstar_noMPI.py` on three processors in parallel, although I *did* hit the seg fault you reported later on. I did have to make a couple modifications to `runrockstar_noMPI.py` for it to run. It also only runs in parallel as far as I can see. Were you really able to reproduce the crash on one core? I'm not able to get a python traceback from the seg fault because faulthandler prints a mysterious error about missing object code.

However I was able to get a C-level traceback from a core dump, it looks like the crash is happening inside of rockstar:

```
#0  strlen () at ../sysdeps/x86_64/strlen.S:106
#1  0x00007f383e6c389e in init_clients ()
   from /home/goldbaum/miniconda3/lib/librockstar.so
#2  0x00007f383e6c6df4 in server ()
   from /home/goldbaum/miniconda3/lib/librockstar.so
#3  0x00007f383eb4d40e in __pyx_pf_2yt_16analysis_modules_12halo_finding_8rockstar_18rockstar_interface_17RockstarInterface_6start_server
    (__pyx_v_self=<optimized out>)
    at yt/analysis_modules/halo_finding/rockstar/rockstar_interface.c:5419
#4  __pyx_pw_2yt_16analysis_modules_12halo_finding_8rockstar_18rockstar_interface_17RockstarInterface_7start_server (
    __pyx_v_self=<optimized out>, unused=<optimized out>)
    at yt/analysis_modules/halo_finding/rockstar/rockstar_interface.c:5385
```

specifically in a call to `strlen` in `init_clients`.

Sorry to not have more concrete advice, this is well outside my personal wheelhouse (I don't do cosmological simulations or generally work with halo finders at all, let alone rockstar).

Didn't you mention that you guys have an updated version of rockstar that yt could use instead of our fork? Does this work with that version of rockstar?

sflarkin:
Rockstar has to run on at least 3 cores as it is written, so I dont think running on one core would be possible. Strange you didnt see the error, Ill reinstall yt and see if that fixes it.

As far as the updated version, I do have an updated version, but it has not been modified to work in yt, as I know is required. I will look what is required for that.

ngoldbaum:
specifically I was running on commit 956f8d59139

ngoldbaum:
Ah and regarding the updated rockstar interface, I see. I might have misunderstood our prior communication about it. It's my understanding that Matt had to do some substantial changes to rockstar itself to make it possible to wrap and that subsequent versions changed the internals enough that it was hard for matt to simply rebase his work. It may be nontrivial to get yt to wrap rockstar-galaxies.

We may also want to rethink our strategy of wrapping it directly at a binary level and instead wrap it using e.g. subprocess calls.

PR #1962, opened 2018-08-10
[wip] yt-4.0 dedicated swift frontend

AshKelly:
## PR Summary

Currently within yt [swift](http://icc.dur.ac.uk/swift/) is partially supported via the Gadget frontend. However, due to the substantial differences - unit system, chemical abundances, etc - it makes sense to reduce unnecessary complexity to build the swift frontend directly onto the SPH frontend. 

This PR adds a basic interface that can correctly load swift files and enable the correct unit system.

## PR Checklist


AshKelly:
So, this is nearly there now. The tests work - as long as the datasets are present - and we can successfully load a swift dataset and get the correct coordinates.

I haven't added support for a multi-file snapshots yet - but, I can do that as soon as I have confirmed the format.

Please feel free to have a look over @ngoldbaum and maybe, @JBorrow could try this locally with some swift datasets?

ngoldbaum:
Can you make a pull request to the website repo that adds a section to `data/datafiles.json` for data from SWIFT and adds entries for the datasets you're using here? Once you've done that you'll also need to upload the datasets somewhere. They should be uploaded as a gzipped tarball that unzips as a folder with a descriptive name containing the dataset. You can use `yt upload /path/to/file.tar.gz` to do the upload, just share the link that prints out when you create the pull request to the website repository but you can also use dropbox, google drive, or space you control on a publicly facing server you have access to. Once you've done that I can stage the dataset on the server that hosts yt-project.org as well as on the server at NCSA that runs the tests.

yt/frontends/swift/data_structures.py
```diff
@@ -0,0 +1,209 @@
...
+from yt.frontends.sph.data_structures import \
+    SPHDataset, \
+    SPHParticleIndex
+
+from .fields import SwiftFieldInfo
+
+from yt.data_objects.static_output import \
+    ParticleFile
+
+from yt.funcs import only_on_root
```
ngoldbaum: super minor nit: can you group the imports together in logical sections with no line breaks? First standard library and external library imports, then internal yt imports. Like this:

```
import numpy as np
from uuid import uuid4

from yt.frontends.sph.data_structures import \
    SPHDataset, \
    SPHParticleIndex
from .fields import SwiftFieldInfo
from yt.data_objects.static_output import \
    ParticleFile
from yt.funcs import only_on_root
```

ngoldbaum: ultimately I want to start using [black](https://github.com/ambv/black) to automate discussions like this but until then...

yt/frontends/swift/data_structures.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Currently sets length, mass, time, and temperature.
+
+        Swift outputs in comoving coordinates and no h factors required.
+        """
+        units = self._get_info_attributes("Units")
+
+        if self.cosmological_simulation == 1:
+            only_on_root(
+                    mylog.info, "Assuming length units are in cm (comoving)")
```
ngoldbaum: I'd say `"assuming length units are in comoving centimeters"` if only because `cm` can be understood to mean either comoving or centimeter depending on the context (especially in yt where we overload that in the unit system).

yt/frontends/swift/data_structures.py
```diff
@@ -0,0 +1,209 @@
...
+
+        # SWIFT is either all periodic, or not periodic at all
+        periodic = int(runtime_parameters["PeriodicBoundariesOn"])
+
+        if periodic:
+            self.periodicity = [True] * self.dimensionality
+        else:
+            self.periodicity = [False] * self.dimensionality
+
+        self.current_time = float(header["Time"])
```
ngoldbaum: does this end up with units attached after the dataset is fully loaded? if not it probably should have units explicitly attached here

yt/frontends/swift/data_structures.py
```diff
@@ -0,0 +1,209 @@
...
+                self.omega_lambda = float(parameters["Cosmology:Omega_lambda"])
+                self.omega_matter = float(parameters["Cosmology:Omega_m"])
+                # This is "little h"
+                self.hubble_constant = float(parameters["Cosmology:h"])
+            except KeyError:
+                mylog.info(
+                    "Could not find cosmology information in Parameters."
+                )
+                mylog.info(
+                    "Setting up as a non-cosmological run. Check this!"
```
ngoldbaum: It would also be worth mentioning in this message that `policy["cosmological integration"]` is a non-null value so swift is saying it was configured to run with comoving coordinates but no cosmology parameters are present, which would be an odd inconsistent state. I might also change these to `mylog.warn` or even a user-visible runtime warning to more loudly notify them that their dataset is odd or possibly broken. Unless of course this can happen a lot.

yt/frontends/swift/data_structures.py
```diff
@@ -0,0 +1,209 @@
...
+        Checks to see if the file is a valid output from SWIFT.
+        This requires the file to have the Code attribute set in the
+        Header dataset to "SWIFT".
+        """
+        filename = args[0]
+        valid = True
+        # Attempt to open the file, if it's not a hdf5 then this will fail:
+        try:
+            handle = h5py.File(filename, "r")
+        except:
```
ngoldbaum: never use bare except: https://stackoverflow.com/questions/14797375/should-i-always-specify-an-exception-type-in-except-statements

I know there are places in other frontends where we do that. We should fix those frontends.

yt/frontends/swift/fields.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
+from yt.frontends.sph.fields import \
+    SPHFieldInfo
+
+class SwiftFieldInfo(SPHFieldInfo):
```
ngoldbaum: I kinda think we should separate out SPHFieldInfo into a `GadgetFieldInfo` and a `TipsyFieldInfo`, if only to keep things tidy. If you'd be interested in doing that cleanup it would be very much appreciated.

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
+"""
+Swift data-file handling function
+
+
+
+
+"""
+from __future__ import print_function
```
ngoldbaum: the future imports aren't necessary since you're not using print anywhere

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+        for data_file in sorted(data_files, key=lambda x: x.filename):
+            si, ei = data_file.start, data_file.end
+            f = h5py.File(data_file.filename, "r")
+            # This double-reads
+            for ptype, field_list in sorted(ptf.items()):
+                if data_file.total_particles[ptype] == 0:
+                    continue
+                x = f["/%s/Coordinates" % ptype][si:ei, 0].astype("float64")
+                y = f["/%s/Coordinates" % ptype][si:ei, 1].astype("float64")
+                z = f["/%s/Coordinates" % ptype][si:ei, 2].astype("float64")
```
ngoldbaum: Can't you do this by broadcasting and only do one I/O for the full coordinates array? I don't know offhand if that's actually any faster or more memory conservative, this may be fine.

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+        pcount = f["/Header"].attrs["NumPart_ThisFile"][:].astype("int")
+        np.clip(pcount - si, 0, ei - si, out=pcount)
+        pcount = pcount.sum()
+        for key in f.keys():
+            if not key.startswith("PartType"):
+                continue
+            if "Coordinates" not in f[key]:
+                continue
+            if needed_ptype and key != needed_ptype:
+                continue
```
ngoldbaum: It may be neater to combine all these into one if statement that you chain with `or`.

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+            ds = f[key]["Coordinates"][si:ei,...]
+            dt = ds.dtype.newbyteorder("N") # Native
+            pos = np.empty(ds.shape, dtype=dt)
+            pos[:] = ds
```
ngoldbaum: Can't you just say `pos = np.array(ds, dtype=dt)`? If so there are a few other places below where you could do the same thing.

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+                continue
+
+            g = f[key]
+            if cname not in g:
+                continue
+
+            ptype = str(key)
+            for k in g.keys():
+                    kk = k
+                    # a little hack to deal with swift masses
```
ngoldbaum: i don't see what the "hack" is here. Can you provide a little more detail? If it's not worth over-explaining I'd just delete the comment.

yt/frontends/swift/tests/test_outputs.py
```diff
@@ -0,0 +1,112 @@
...
+
+import numpy as np
+
+from yt.testing import requires_file
+from yt.frontends.swift.api import SwiftDataset
+from yt.testing import assert_almost_equal
+from yt.utilities.on_demand_imports import _h5py as h5py
+
+keplerian_ring = "KeplerianRing/keplerian_ring_0020.hdf5"
+EAGLE_6 = "EAGLE_6/eagle_0005.hdf5"
```
ngoldbaum: so these tests won't actually run on appveyor or travis and I don't think we're planning on setting up jenkins for the yt-4.0 branch anytime soon. It might be a good idea to think about modifying our test setup on travis to download a minimal (small on disk, I think we only have 2 GB of storage on travis) subset of the test datasets, (in particular SPH test datasets) so the unit tests that use real data will fire on travis. I'm pretty sure there are some tests that use real datasets that are currently failing at the moment.

yt/frontends/swift/data_structures.py
AshKelly: yeah! When do you propose to start using black? Josh always talks about this - it looks really cool

yt/frontends/swift/data_structures.py
AshKelly: That's a good point - I should check that. I should add a test for time/velocity units

yt/frontends/swift/data_structures.py
AshKelly: Yep - I'm fairly sure if this happens something really dodgy is going on!

yt/frontends/swift/fields.py
AshKelly: i.e get rid of SPHFieldInfo and just have Tipsy and Gadget?
I'm happy to do this, just unsure exactly what is being asked

yt/frontends/swift/io.py
AshKelly: this is a copy and paste, thanks for the catch :)

yt/frontends/swift/fields.py
ngoldbaum: So `SPHFieldInfo` contains both Tipsy and Gadget field names, mostly for historical reasons. It's probably not the best idea to combine them like that. We should separate them. It might ultimately make sense to keep some parts of `SPHFieldInfo` (e.g. some of the methods might be generically useful) but the field names that are particular to each frontend should show up in that frontend, not in a base class that's shared by multiple frontends.

yt/frontends/swift/io.py
ngoldbaum: Or even `np.asarray` to avoid the copy.

yt/frontends/swift/fields.py
AshKelly: I see yep - I can have a look at that!

AshKelly:
I can make a PR to the website repo - I'm just traveling at the moment but will do so once I arrive.

I'll also work through the rest of the comments on here! Thanks for the review :smile:

yt/frontends/gadget/data_structures.py
```diff
@@ -546,6 +546,7 @@ def _is_valid(self, *args, **kwargs):
             fh = h5py.File(args[0], mode='r')
             valid = all(ng in fh["/"] for ng in need_groups) and \
                 not any(vg in fh["/"] for vg in veto_groups)
+            valid = fh["Header"].attrs["Code"] != b"SWIFT"
```
JBorrow: I would prefer it if you would convert the stuff coming out of hdf5 to utf-8; the specific formatting might change in the future.

yt/frontends/swift/api.py
```diff
@@ -0,0 +1,26 @@
+"""
+API for SWIFT frontend
+
```
JBorrow: Please reference the SWIFT code here and give a bit more information. Say that you worked on the file maybe so people know who to go and find when things don't work?

yt/frontends/swift/data_structures.py
```diff
@@ -0,0 +1,209 @@
...
+        super().__init__(filename, dataset_type, units_override=units_override)
+        self.storage_filename = storage_filename
+
+    def _set_code_unit_attributes(self):
+        """
+        Sets the units from the SWIFT internal unit system.
+
+        Currently sets length, mass, time, and temperature.
+
+        Swift outputs in comoving coordinates and no h factors required.
```
JBorrow: Swift -> SWIFT. Also, "SWIFT uses comoving co-ordinates without the usual h-factors."

yt/frontends/swift/data_structures.py
```diff
@@ -0,0 +1,209 @@
...
+        units = self._get_info_attributes("Units")
+
+        if self.cosmological_simulation == 1:
+            only_on_root(
+                    mylog.info, "Assuming length units are in cm (comoving)")
+            self.length_unit = self.quan(
+                float(units["Unit length in cgs (U_L)"]), "cmcm")
+        else:
+            only_on_root(
+                    mylog.info, "Assuming length units are in cm (physical)")
```
JBorrow: Follow-up on Nathan's comment:

"Assuming length units are in physical centimetres"

yt/frontends/swift/data_structures.py
JBorrow: Yes this time should be in the same units as the rest of the file (based on the I/O units).

yt/frontends/swift/data_structures.py
JBorrow: I am a little confused by what you mean Nathan. In SWIFT, cosmological integration is what we call an "engine policy"; you do not have to compile with(out) cosmological integration. To run with cosmology, you just pass an extra flag to the binary (`swift -c ...`). If you attempt to run with `-c` and you do not have the cosmological parameters in your parameter file, we don't let you run.

yt/frontends/swift/data_structures.py
JBorrow: You are looking for `except (FileNotFoundError, OSError) as e` as far as I am aware.

yt/frontends/swift/io.py
JBorrow: I would advocate for `x, y, z = f["/%s/Coordinates" % ptype][si:ei].T`. I am unsure if the `astype` conversion should be necessary as we store them in the HDF5 as doubles anyway.

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+            # This double-reads
+            for ptype, field_list in sorted(ptf.items()):
+                if data_file.total_particles[ptype] == 0:
+                    continue
+                x = f["/%s/Coordinates" % ptype][si:ei, 0].astype("float64")
+                y = f["/%s/Coordinates" % ptype][si:ei, 1].astype("float64")
+                z = f["/%s/Coordinates" % ptype][si:ei, 2].astype("float64")
+                if ptype == self.ds._sph_ptype:
+                    hsml = f[
+                        "/%s/SmoothingLength" % ptype][si:ei].astype("float64")
```
JBorrow: Smoothing length's should not need to be doubles (`float32` should be fine) -- I don't think the conversion is necessary (unless the internals of `yt` require this).

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def _read_particle_coords(self, chunks, ptf):
+        # This will read chunks and yield the results.
+        chunks = list(chunks)
+        data_files = set([])
+        for chunk in chunks:
+            for obj in chunk.objs:
+                data_files.update(obj.data_files)
+        for data_file in sorted(data_files, key=lambda x: x.filename):
+            si, ei = data_file.start, data_file.end
```
JBorrow: Why do you do this? Is this to do with chunking?

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+                        "/%s/SmoothingLength" % ptype][si:ei].astype("float64")
+                else:
+                    hsml = 0.0
+                yield ptype, (x, y, z), hsml
+            f.close()
+
+    def _yield_coordinates(self, data_file, needed_ptype=None):
+        si, ei = data_file.start, data_file.end
+        f = h5py.File(data_file.filename)
+        pcount = f["/Header"].attrs["NumPart_ThisFile"][:].astype("int")
```
JBorrow: This is not necessary for SWIFT; we never write multiple output files like GADGET.

yt/frontends/swift/io.py
JBorrow: Also: why not just call the `_get_smoothing_length` function here? Am I missing something?

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+        ptype = self.ds._sph_ptype
+        ind = int(ptype[-1])
+        si, ei = data_file.start, data_file.end
+        with h5py.File(data_file.filename) as f:
+            pcount = f["/Header"].attrs["NumPart_ThisFile"][ind].astype("int")
+            pcount = np.clip(pcount - si, 0, ei - si)
+            ds = f[ptype]["SmoothingLength"][si:ei,...]
+            dt = ds.dtype.newbyteorder("N") # Native
+            if position_dtype is not None and dt < position_dtype:
+                # Sometimes positions are stored in double precision
```
JBorrow: Positions are _always_ doubles in SWIFT. Smoothing lengths are _always_ singles.

yt/frontends/swift/io.py
```diff
@@ -0,0 +1,168 @@
...
+                data_files.update(obj.data_files)
+
+        for data_file in sorted(data_files, key=lambda x: x.filename):
+            si, ei = data_file.start, data_file.end
+            f = h5py.File(data_file.filename, "r")
+            for ptype, field_list in sorted(ptf.items()):
+                if data_file.total_particles[ptype] == 0:
+                    continue
+                g = f["/%s" % ptype]
+                coords = g["Coordinates"][si:ei].astype("float64")
```
JBorrow: Does this explicit type conversion cost anything? Or is this just a proxy for "read this all please"

yt/frontends/swift/data_structures.py
AshKelly: the issue here is the file _thought_ it was a cosmo run as `self.cosmological_simulation` is true, but then when we tried to load the parameters we failed. Surely, that is a strange case?

yt/frontends/swift/io.py
AshKelly: We keep smoothing lengths are float64 for better compatibility with a bunch of cython functions. But yeah, we should call `_get_smoothing_length`

yt/frontends/swift/data_structures.py
JBorrow: Ah yeah, that is a very strange case; hence the logging information. Perhaps it should be expanded to say "Could not find cosmology information in Parameters, despite having ran with -c signifying a cosmological run."

AshKelly:
I'm getting a segault when testing plots with EAGLE_6 dataset, but not with the Kep ring. I'm going to look into the cause of this.

AshKelly:
In hindsight, I think the bug I just mentioned may be related to the kdtree particle projections based on what I saw when it crashed.

JBorrow:
For the segfault, what happens when you read the data with the GADGET frontend?

yt/frontends/gadget/data_structures.py
AshKelly: so what do you mean here? h5py loads this as a numpy._bytes which is in the utf-8 format. What would you prefer here?

JBorrow: `valid = fh["Header"].attrs["Code"].decode("utf-8") != "SWIFT"`

AshKelly:
I've added some basic swift doc, but the quite a bit of load doc needs updating for the yt-4.0 branch. There seems to be quite a lot of mention to the in-memory octree.

MatthewTurk:
@AshKelly the octree gets talked about a fair bit because we figured that would be a highly-visible albatross...  😄

JBorrow:
FWIW, it's probably worth just merging this now (to prevent it from diverging too much), and we can check the other stuff later :)

ngoldbaum:
That's probably a good point. Let me re-trigger CI on this since it's been a while since this has seen updates, then I'll merge later today. I'd still really like to add testing using real datasets but we're not doing that for the other SPH frontends so this isn't really any worse than the status quo.

JBorrow:
Great, thanks very much to you both for your work on this!

PR #1965, opened 2018-08-10
[WIP] xy-grid extrusion along z

MatthewTurk:
This implements the first steps toward a data index that is defined by an xy grid of points with constant z.  For instance, the particular use case is ARM data that is taken at multiple sweeps along the azimuth, at roughly onstant elevation, and *always* constant "range" (distance).  This could be defined as a hexahedral mesh, but because we have a reduced dimensionality in xy, it can reduce memory considerably to define it with `xy_grid` and `z_grid`.

Also!  I may have messed up my git branching!  So there's the likely possibility that I'm including irrelevant changesets!  But, when Nathan's PR goes in and I rebase, maybe that will go away?

There are numerous todo's, including:

- [ ] finish implementation of extrusion base data object
- [ ] finish fwidth
- [ ] integrate into pixelizers the new extruded mesh pixelizer
- [ ] implement `cf_sweep` frontend

### DIFF
--- a/yt/data_objects/point_grid_extruded_mesh.py
+++ b/yt/data_objects/point_grid_extruded_mesh.py
@@ -1,0 +1,127 @@
+"""
+Unstructured mesh base container.
+
+
+
+"""
+
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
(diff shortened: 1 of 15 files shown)


matthewturk:
I'm converting this to draft until after 4.0 goes out so that it won't show up in my filters.  This is a worthwhile change, but I need to update it and move forward with an updated version.

matthewturk:
I do want to revisit this, albeit perhaps with a rebase, at some point in the future.

PR #1967, opened 2018-08-11
[WIP] Download test data on Travis for testing

git-abhishek:
## PR Summary

This feature helps in testing with real dataset, if the dataset is not present in local repository. The aim is to download test dataset that are smaller in size thus, enabling them with the exisiting `@require_ds()`
decorator to be run on Travis.

If the test dataset is already present, this function will not do anything. If test dataset is not present, it will check the test data cache for corresponding tarfile. If tarfile is present, it will extract it to the test data directory location, otherwise it will download and extract the tar file.

This would help in testing cases like #1942.

## Possible Concerns
- I think it will be good to wrap the `download_file()` with time out decorator/function. The timeout could be set to 1-2 min depending upon the size of file we want to download.

- I tried couple of locking mechanisms that would lock access to `travis_cache_dir` so that only one process would download the data file. However, all my tries were futile and I think locking does not work with different python process (should work for forked subprocess). One possible hack for this is to only enable download for only one python environment eg, python 3.6 Answer Tests of Travis. 
Lastly, this issue will only surface when different enviroments are executing the same test at the same time where the test has `@require_ds` decorator with `download=True`. 

## PR Checklist

- [ ] **New features are documented, with docstrings and narrative docs**

## Adding Reviewers


git-abhishek:
@yt-fido test this please.

git-abhishek:
So should I have a new yaml file that will enlist all the real dataset (small in size) that needs to be downloaded on Travis? Then, a python script will read this yaml, cache and extract it.

ngoldbaum:
that sounds like a good approach to me

matthewturk:
I'm going to close this PR, as there are new developments in both data storage and testing we are trying out.  Thank you!

PR #1968, opened 2018-08-11
merging with master

AshKelly:
## PR Summary

Keeping the yt-4.0 branch and master branch similar. It will also apply the `sudo` patch to the yt-4.0 which may help with sporadically failing tests.

I will patch if any tests are broken.


AshKelly:
Hey @ngoldbaum it looks like this merge was fairly straight forward. I only had to resolve a couple of conflicts!

PR #1979, opened 2018-08-20
[Bugfix] WarpX datasets with no particles

atmyers:
This change is needed to properly read WarpX datasets that don't have particles.


ngoldbaum:
Feel free to self-merge assuming the tests pass, this is a simple enough change.

ngoldbaum:
It might also be worthwhile to add a test for this, along with a dataset with no particles, but that's not necessary to merge this.

PR #1981, opened 2018-08-21
create individual particle fields using SNIa and SNII metallicities 

claytonstrawn:
TRIDENT requires fields named `O_nuclei_mass_density, C_nuclei_mass_density, Fe_nuclei_mass_density`, etc in order to use specific atoms in it's density calculations. If they are missing, then it uses an overall "metallicity" field and assumes solar abundance. 

Since ART tracks type Ia and type II SNe separately, and they have very different abundances, it would be more accurate to insert atom fields at this stage using that information, so that TRIDENT doesn't have to assume this.

This PR adds those fields to the setup_fluid_fields method.


ngoldbaum:
@chummels can you take a look at this? I know you don't know about ART internals but you do know about the species fields and trident :)

PR #1982, opened 2018-08-22
yt-4.0: disable testing on python2.7

ngoldbaum:
This makes the tests only run on python3. I've kept the python2.7 linting stage to make backporting to the 3.x branch easier but now we don't run the unit tests. This should allow people to use python3-specific functionality as long as you don't use python3-specific syntax, which I think is a good compromise for now until we no longer support a maintenance branch that supports python2.7.


ngoldbaum:
I should also say that this also adds testing on python3.7 using the `xenial` Travis image.

PR #1984, opened 2018-08-23
 avoid float precision issues leading to overflow in bitmap index 

ngoldbaum:
This fixes the issues @AshKelly ran into with the test SWIFT dataset in #1962.

Ultimately the issue was due to going between float coordinates and int coordinates. The fix is just to avoid those conversions and calculate the int coordinates as we recurse.

In addition we converted some pointer arrays to memoryviews, that makes it easier to debug this code when it breaks.

### DIFF
--- a/yt/geometry/particle_oct_container.pyx
+++ b/yt/geometry/particle_oct_container.pyx
@@ -1830,6 +1830,15 @@ cdef class ParticleBitmapSelector:
                             ncur_ind[0], ncur_ind[1], ncur_ind[2])
                         if sbbox == 2: # an edge cell
                             if self.is_refined(mi1) == 1:
+                                # note we pass zeros here in the last argument
+                                # this is because we now need to generate
+                                # *refined* indices above order1 so we need to
+                                # start a new running count of refined indices.
+                                #
+                                # note that recursive_morton_mask does not
+                                # mutate the last argument (a new index is
+                                # calculated in each stack frame) so this is
+                                # safe
                                 self.recursive_morton_mask(
                                     nlevel, npos, ndds, mi1, zeros)
                         self.add_coarse(mi1, sbbox)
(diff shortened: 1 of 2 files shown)


AshKelly:
This works perfect on my machine to generate the projection plots which were previously segfaulting. 

The results also agree with the `OffAxisProjection`which was previously, and still is, working.

yt/geometry/particle_oct_container.pyx
```diff
@@ -18,7 +18,7 @@ from oct_container cimport OctreeContainer, Oct, OctInfo, ORDER_MAX, \
     SparseOctreeContainer, OctKey, OctAllocationContainer
 cimport oct_visitors
 from oct_visitors cimport cind, OctVisitor
-from libc.stdlib cimport malloc, free, qsort
+from libc.stdlib cimport malloc, realloc, free, qsort
```
MatthewTurk: I *think* we don't use `realloc`

yt/geometry/particle_oct_container.pyx
```diff
@@ -1846,38 +1806,44 @@ cdef class ParticleBitmapSelector:
...
+                                nlevel, npos, ndds, mi1, ncur_ind)
                     elif nlevel == self.order1:
-                        mi1 = bounded_morton_dds(npos[0], npos[1], npos[2], self.DLE, ndds)
+                        mi1 = encode_morton_64bit(
+                            ncur_ind[0], ncur_ind[1], ncur_ind[2])
                         if sbbox == 2: # an edge cell
                             if self.is_refined(mi1) == 1:
-                                self.recursive_morton_mask(nlevel, npos, ndds, mi1)
+                                self.recursive_morton_mask(
+                                    nlevel, npos, ndds, mi1, zeros)
```
MatthewTurk: I think this is right, but it took me a second to work out that this was not mutated on the other end.

yt/geometry/particle_oct_container.pyx
AshKelly: oops, I added this before sending the memoryview refactor to Nathan. I was using it to experiment

yt/geometry/particle_oct_container.pyx
ngoldbaum: yup! you did this in a nice clever way without mutating anything :)

ngoldbaum: i guess this is worth a comment though

PR #1989, opened 2018-08-24
[bugfix] Make extra check before writing bitmasks to avoid crash

chummels:
I've been having problems on Caltech's cluster where yt fails when trying to access some particle-based datasets, because it tries to write the bitmask (.ewah) file when it thinks permissions are writable but they are not.  I don't know why os.access mis-reports this information, but it does.  This just avoids the failure if it cannot actually write the .ewah file.


ngoldbaum:
Seems fine to me. Only comment is that the new comment should be at the same level of identation as the try/except

chummels:
Good call.

PR #1990, opened 2018-08-27
ENH: Adding cfradial frontend.

zssherman:
## PR Summary

This pull request contains the commits from @ngoldbaum , @MatthewTurk  and I for the cfradial frontend. The new front ends allows for a radar file to be loaded using yt. Unit tests have been written for the frontend. I had to do a revert because I forgot I had the plot callback in this branch, so I reverted the commit, and have the plot callback code in a separate branch for a future pull request.

## PR Checklist

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -183,7 +183,7 @@ answer_tests:
   local_nc4_cm1_001: # PR  2176
     - yt/frontends/nc4_cm1/tests/test_outputs.py:test_cm1_mesh_fields
 
-  local_cf_radial_001: # PR  1990
+  local_cf_radial_002: # PR  1990
     - yt/frontends/cf_radial/tests/test_outputs.py:test_cfradial_grid_field_values
 
 other_tests:


neutrinoceros:
Hi @zssherman. I'm willing to give you a hand in getting this merged.
The first issue I'm seeing with the current state of the code base is that you're modifying a file that is not part of yt anymore, and now belongs in `unyt`. @jzuhone should be able to guide you through migrating this to the other project.

Also it's been a while since the tests were run, and the infrastucture has changed quite a bit.
You might want to rebase your branch against the `main` branch (⚠️ , it used to be called `master`). Be aware you'll have to resolve conflicts, but they look pretty easy to me (except for the file that moved to a child project).

Note that we now use pre-commit to lint and auto-fix many nitpicks in code style, you might want to read this to learn more
https://yt-project.org/docs/dev/developing/developing.html#automatically-checking-and-fixing-code-style

Another quick remark is that we removed copyright headers a little while ago.
```python
#-----------------------------------------------------------------------------
# Copyright (c) 2013, yt Development Team.
# ...
```
can you please remove the ones still in this PR ?

I will review this more in depth when you're ready. Don't hesitate to reach out here or on our Slack space if you need help !

zssherman:
Hi @neutrinoceros , thanks for the information and taking a look at this! I will apply the changes you suggested and if I run into roadblocks let you know. I should be able to take a look at this again this weekend or early next week.

chrishavlin:
This is great! It'll be great to get it in! 

One quick comment: I'm not sure about the top level `xarray` import. We probably should use `yt.utilities.on_demand_imports` to avoid making `xarray` a hard dependency for yt. We actually ran into this with the `nc4_cm1` frontend -- I initially added `xarray` to `on_demand_imports` (in this [commit](https://github.com/yt-project/yt/pull/2176/commits/42b6abf8c53892edd8990042f953ae0b224c9e8d)). Since we were just using `xarray` for reading fields it was actually simpler to use the existing netcdf handler `yt.utilities.file_handler.NetCDF4FileHandler` and I ended up removing that code before the frontend got merged. Since this frontend actually writes a processed `.nc` file it may warrant using `xarray`, but we should still avoid the top level import here.

In no way should that affect this PR, but there is actually a lot of overlap with the `nc4_cm1` frontend. It make me think that a more general abstract netcdf frontend class would be helpful (I started one some months ago when starting on a frontend for seismic tomography... need to brush that off...).

zssherman:
@neutrinoceros I believe I got most of the conflicts sorted out, and removed copywrite info. I also submitted a [PR](https://github.com/yt-project/unyt/pull/184) here for the unit fix. I wasn't entirely sure what to fix next.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,237 @@
...
+        self,
+        filename,
+        dataset_type="cf_radial",
+        storage_filename=None,
+        units_override=None,
+    ):
+        self.fluid_types += ("cf_radial",)
+        self._handle = xarray.open_dataset(filename)
+        self.refine_by = 2
+        if "x" not in self._handle.coords:
```
chrishavlin, line 126: Could add a check for the existence of the gridded file here in case we've already gridded and saved the new file in a previous load (unless you see a reason to re-grid every time).

chrishavlin (commented):
Nice work on sorting out the conflicts! Some initial comments, suggestions and questions below.

zssherman:
Thank you @chrishavlin for taking a look! I've read your suggestions and will implement them soon. Yeah I believe when we originally wrote this I believe we were getting it to work on simpler or more general cases. I agree that changing the fields and grid parameter is needed as hardcoded the parameters is not ideal. I'll do those changes soon.

yt/frontends/cf_radial/data_structures.py
zssherman: For this, I believe when we wrote this, the if x was the check if the file was a grid file, if so we skip the pyart gridding to cartesian step. I can also add to the if: and "_grid.nc" not in filename, but wasn't too sure and state our original intentions might have covered this one.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,237 @@
...
+    def _is_valid(self, *args, **kwargs):
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+        try:
+            import xarray
+        except ImportError:
+            return False
+
+        try:
+            ds = xarray.open_dataset(args[0])
```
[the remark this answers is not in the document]
zssherman: Has been updated

zssherman:
@chrishavlin I believed I completed all your suggestions but the one I commented on. I added the pyart in on demand imports as well. I believe I added both xarray to pyart to on_demand_imports correctly. Thanks again for the comments and suggestions! I can do more changes if needed.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,237 @@
...
+    def __init__(self, filename, dataset_type='cf_radial',
+                 grid_shape=(40, 200, 200),
+                 grid_limits=((0., 2000.), (-1e5, 1e5), (-1e5, 1e5)),
+                 storage_filename=None,
+                 units_override=None):
+        self.fluid_types += ('cf_radial',)
+        self._handle = xr.open_dataset(filename)
+        self.refine_by = 2
+        if 'x' not in self._handle.coords:
+            from yt.utilities.on_demand_imports import pyart
```
chrishavlin, line 126:
```suggestion
            from yt.utilities.on_demand_imports import _pyart as pyart
```

yt/frontends/cf_radial/data_structures.py
chrishavlin: ok, that makes some sense. My point is just that if a user calls `yt.load("cfradial_notgridded.nc")`, it would correctly trigger the pyart gridding, after which "cfradial_notgridded_gridded.nc" would exist. So in subsequent calls to `yt.load("cfradial_notgridded.nc")` it'd be nice to simply reload "cfradial_notgridded_gridded.nc" since it already exists. I was thinking something like (in pseudo-code):

```python
if "x" not in self._handle.coords:

    if gridded_file does not exist:
         do the pyart gridding, create gridded file

    load up the gridded file into self._handle    
```

This really only matters if the gridding step is computationally difficult -- if it's quick then doing it every time is OK, but if the gridding step takes some time, why not re-use the gridded files?

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,237 @@
...
+            radar = pyart.io.read_cfradial(filename)
+            self.grid_shape = grid_shape
+            self.grid_limits = grid_limits
+            grid = pyart.map.grid_from_radars(
+                (radar, ), grid_shape=self.grid_shape,
+                grid_limits=grid_limits=self.grid_limits)
```
chrishavlin, line 132:
```suggestion
                grid_limits=self.grid_limits)
```

yt/utilities/on_demand_imports.py
```diff
@@ -251,6 +251,44 @@ def pooch(self):
...
+
+    @property
+    def xarray(self):
+        if self._xarray is None:
+            try:
+                import xarray as xr
+            except ImportError:
+                xarray = NotAModule(self._name)
+            self._xarray = xarray
+        return self._xarray
```
neutrinoceros, line 286: sorry I'm late to review this.
Actually there is a slightly cleaner way define such classes that I advise you follow (it will make deduplicating that code a little easier if we ever do it).
Specifically I'm thinking of how it's done for  e.g. `f90nml_imports` `requests_imports`

neutrinoceros (commented):
can't review fully now but here's a suggestion ! keep up the good work

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,237 @@
...
+        filename,
+        dataset_type="cf_radial",
+        storage_filename=None,
+        units_override=None,
+    ):
+        self.fluid_types += ("cf_radial",)
+        self._handle = xarray.open_dataset(filename)
+        self.refine_by = 2
+        if "x" not in self._handle.coords:
+            import pyart
```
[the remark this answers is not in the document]
neutrinoceros: I do ! see bellow :)

chrishavlin:
Great! Thanks, @zssherman! one question: any idea if the test data, `"CfRadialGrid/grid1.nc"`, has been uploaded to our testing server? We'll know once the tests start making it past the initial import but I suspect it's not been uploaded... Assuming it's not there, could you share a link to the file and we can get it where it needs to go? you can use yt at the command line to upload an archive of the `CfRadialGrid` directory with `yt upload CfRadialGrid.tar.gz` and it will upload the file to the yt hub and give you a link you can share here ([docs on yt upload](https://yt-project.org/doc/reference/command-line.html#upload)).

neutrinoceros:
@chrishavlin `CfRadialGrid` is registered on the server already :)
(see https://raw.githubusercontent.com/yt-project/website/master/data/datafiles.json)

yt/frontends/cf_radial/data_structures.py
zssherman: It depends on the grid dimensions, so it could be slow, so I agree. I added a check, I believe I did it as you mentioned, if not I can fix it if need be.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,239 @@
...
+
+"""
+
+import os
+import stat
+import weakref
+
+import numpy as np
+
+from yt.data_objects.grid_patch import AMRGridPatch
```
chrishavlin, line 14: Looks like this module changed locations since the initial PR: 
```suggestion
from yt.data_objects.index_subobjects.grid_patch import AMRGridPatch
```

chrishavlin:
Hey @zssherman , is it OK if I work through a rebase of this branch to correct the bulk of the failing tests (the current changeset is missing the `answer-store`, easiest fix is a rebase)? Do you have any changes not pushed? You would have to reset to the rebased branch after I push (and you would lose any changes you haven't pushed).

zssherman:
@chrishavlin No worries, any changes I've done are all pushed to this PR, nothing local on my end so I can reset to the rebase

chrishavlin:
So this is almost there! The failing answer tests are only related to the new tests in this PR (not sure why the other build crashed... will investigate...). 

Two things I'm wondering: 

1. maybe it would make sense to change the "dBz" units to "dimensionless" for now so we're not blocked by the fix to unyt? We could open an issue to keep track of this.

2. I think it'd be good to add a test that actually does the pyart gridding (the new tests for this only use an already gridded file). @zssherman do you have a good test file we could use?

zssherman:
@chrishavlin I do have a file that we can use prior to gridding for testing, doesn't look like I can attach it here, but here is a link to it:
[radar_file](https://engineering.arm.gov/~jhelmus/pyart_example_data/XSW110520113537.RAW7HHL)
And for the dbz change, under the field_dict in this PR?

chrishavlin:
Thanks for the file! Ya -- the `dBz` change would be in the field dict defined in this PR in `cf_radial.fields.CFRadialFieldInfo` and then also in the new tests in `cf_radial.tests.test_outputs`. And I think an info/warning message stating something to the effect of "Loading in reflectivity as nondimensional" would be useful, but let me think about how/where to do that best...

chrishavlin:
So looks like that file requires `pyart.read_rsl()` rather than `pyart.read_cfradial()` (which requires building `pyart` with the TRMM RSL library, which is probably a few steps beyond what we want for this PR...). In reading through the pyart docs and examples more carefully, it looks like `swx_20120520_0641.nc`, `sex_20120520_0641.nc` and `sgpcsaprsurcmacI7.c0.20110520.095101.nc`  from the [pyart example data](https://engineering.arm.gov/~jhelmus/pyart_example_data/) are all cfradial files. OK if we use one of those? I'd lean towards `swx` as it's the smallest of the 3.

And so I don't forget, when we add `pyart` as an optional dependency the proper package to install is `arm_pyart` (not just `pyart`).

zssherman:
@chrishavlin Yeah, sorry I chose the wrong file, either of those would work as well. And yeah the conda forge and pip install of pyart is arm_pyart, however the import in code is just pyart

chrishavlin:
ok, great! I can write a new test and add a few more infrastructure fixes soon -- I'm happy to add that `dBz` change while I'm at it.

chrishavlin:
Just pushed some infrastructure changes and another rebase that should get all the tests passing. I'd say this is ready for wider review!  @zssherman if any comments come up related to the changes I added on at the end here, I can help sort those out. I'm also happy to approve this PR at this point, but since I added some changes I'll hold off until others have a chance to review.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+                        units = ""
+                        warn_units += 1
+
+                    for findstr, repstr in fic.unit_subs:
+                        units = units.replace(findstr, repstr)
+
+                    flds_to_add += ((key, (units, [], self.ds._handle[key].long_name)),)
+
+        if flds_to_add:
+            self.ds._field_info_class.known_other_fields += flds_to_add
```
neutrinoceros, line 92:
```suggestion
        self.ds._field_info_class.known_other_fields += tuple(flds_to_add)
```
Actually this feels very much like a hack and I'd like to understand your motivation to update an external class here. I'm convinced this isn't necessary or that at least it shouldn't be.

yt/frontends/cf_radial/io.py
```diff
@@ -0,0 +1,67 @@
...
+        # data.  Note that if you're reading grid data, you might need to
+        # special-case a grid selector object.
+        # Also note that "chunks" is a generator for multiple chunks, each of
+        # which contains a list of grids. The returned numpy arrays should be
+        # in 64-bit float and contiguous along the z direction. Therefore, for
+        # a C-like input array with the dimension [x][y][z] or a
+        # Fortran-like input array with the dimension (z,y,x), a matrix
+        # transpose is required (e.g., using np_array.transpose() or
+        # np_array.swapaxes(0,2)).
+        chunks = list(chunks)
```
neutrinoceros, line 45: I don't think this is necessary, and in fact it may have a negative impact on perfs

yt/frontends/cf_radial/io.py
```diff
@@ -0,0 +1,67 @@
...
+
+        rv = {}
+        for field in fields:
+            rv[field] = np.empty(size, dtype="float64")
```
neutrinoceros, line 49:
```suggestion
        rv = {field: np.empty(size, dtype="float64") for field in fields}
```

yt/frontends/cf_radial/io.py
```diff
@@ -0,0 +1,67 @@
...
+
+        offset = 0
+
+        for field in fields:
+            for chunk in chunks:
+                for grid in chunk.objs:
+                    ds = self.ds._handle
+                    variable = ds.variables[field[1]]
+                    data = variable.values[0, ...].T
```
neutrinoceros, line 58: the transposition may have undesirable perfs impact and could probably be avoided by writing the data in the appropriate order from the start.

yt/frontends/cf_radial/io.py
```diff
@@ -0,0 +1,67 @@
...
+        # ptypes and the values are lists of fields.
+        pass
+
+    def _read_particle_fields(self, chunks, ptf, selector):
+        # This gets called after the arrays have been allocated.  It needs to
+        # yield ((ptype, field), data) where data is the masked results of
+        # reading ptype, field and applying the selector to the data read in.
+        # Selector objects have a .select_points(x,y,z) that returns a mask, so
+        # you need to do your masking here.
+        pass
```
neutrinoceros, line 29: I think you can just remove those altogether

yt/frontends/cf_radial/fields.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+
+"""
+
+from yt.fields.field_info_container import FieldInfoContainer
+
+# We need to specify which fields we might have in our dataset.  The field info
+# container subclass here will define which fields it knows about.  There are
+# optionally methods on it that get called which can be subclassed.
```
neutrinoceros, line 12: this comment can be removed

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+        #   self.omega_matter               <= float
+        #   self.hubble_constant            <= float
+        self.cosmological_simulation = 0.0
+        self.current_redshift = 0.0
+        self.omega_lambda = 0.0
+        self.omega_matter = 0.0
+        self.hubble_constant = 0.0
+
+    @classmethod
+    def _is_valid(self, filename, *args, **kwargs):
```
neutrinoceros, line 245: This is a _class_ method (not your fault ! this error was likely copy-pasted from skeleton before I corrected it there)
```suggestion
    @classmethod
    def _is_valid(cls, filename, *args, **kwargs):
```

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+            conventions = ds.attrs.get("Conventions", "")
+            conventions += ds.attrs.get("conventions", "")
+        except AttributeError:
+            # either attrs does not exist or its not a dictionary, so not valid
+            return False
+
+        if "CF/Radial" in conventions:
+            return True
+
+        return False
```
neutrinoceros, line 269: I like compact code 😎 
```suggestion
        return "CF/Radial" in conventions
```

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+        except OSError:
+            return False
+
+        try:
+            # handle the different upper/lower case cases
+            conventions = ds.attrs.get("Conventions", "")
+            conventions += ds.attrs.get("conventions", "")
+        except AttributeError:
+            # either attrs does not exist or its not a dictionary, so not valid
+            return False
```
neutrinoceros, line 264: try blocks are more expensive than if statements when conditions are typically not met (as is the case in _is_valid methods in yt). Not all of them can be refactored but this one can !
```suggestion
        if not (ds.hasattr("Conventions") or ds.hasattr("convention"):
            # either attrs does not exist or its not a dictionary, so not valid
            return False
```

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+        #
+        # We also set up cosmological information.  Set these to zero if
+        # non-cosmological.
+        #
+        #   self.cosmological_simulation    <= int, 0 or 1
+        #   self.current_redshift           <= float
+        #   self.omega_lambda               <= float
+        #   self.omega_matter               <= float
+        #   self.hubble_constant            <= float
+        self.cosmological_simulation = 0.0
```
neutrinoceros, line 238:
```suggestion
        self.cosmological_simulation = 0
```

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+
+        #
+        #   self.unique_identifier      <= unique identifier for the dataset
+        #                                  being read (e.g., UUID or ST_CTIME)
+        self.unique_identifier = int(os.stat(self.parameter_filename)[stat.ST_CTIME])
+        #   self.parameters             <= full of code-specific items of use
+        self.parameters = {}
+
+        coords = self._handle.coords
+        x, y, z = [coords[d] for d in "xyz"]
```
neutrinoceros, line 209:
```suggestion
        x, y, z = [self._handle.coords[d] for d in "xyz"]
```

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+    def _parse_parameter_file(self):
+        # This needs to set up the following items.  Note that these are all
+        # assumed to be in code units; domain_left_edge and domain_right_edge
+        # will be converted to YTArray automatically at a later time.
+        # This includes the cosmological parameters.
+
+        #
+        #   self.unique_identifier      <= unique identifier for the dataset
+        #                                  being read (e.g., UUID or ST_CTIME)
+        self.unique_identifier = int(os.stat(self.parameter_filename)[stat.ST_CTIME])
```
neutrinoceros, line 204: this was copy-pasted in a lot of frontends, and it's now redundant with the `unique_identifier` property defined in the base class 
```suggestion
```

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+                    filename,
+                    new_filename,
+                )
+                grid.write(new_filename)
+            self._handle = xr.open_dataset(new_filename)
+            filename = new_filename
+        super().__init__(filename, dataset_type, units_override=units_override)
+        self.storage_filename = storage_filename
+        # refinement factor between a grid and its subgrid
+        # self.refine_by = 2
```
neutrinoceros, line 177: Even if unused, this should be set to avoid breaking other methods (this attribute doesn't have a default value)
```suggestion
        self.refine_by = 2
```

neutrinoceros (commented):
Good job @zssherman and @chrishavlin !
Here's an avalanche of small comments to help improve the code quality.
I believe most if not all or them are very easy to implement. Keep it up !

yt/frontends/cf_radial/data_structures.py
chrishavlin: ya, it was a bit hacky (this change was mine) -- the basic issue was just that netcdf's can have a lot of fields and writing out all possible fields in `known_other_fields` isn't really possible. Without that modification, fields added to `self.field_list` that don't exist in `known_other_fields` load OK but don't have any units attached. This was my attempt to dynamically pull out the netcdf attribute data in a way that the units and descriptions would be used by yt -- maybe something exists already to do this more cleanly?

yt/frontends/cf_radial/tests/test_outputs.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+@requires_ds(cf_nongridded)
+def test_gridding():
+    # loads up a radial dataset, which triggers the gridding
+    ds = data_dir_load(cf_nongridded)
+
+    # check that the gridded file exists now
+    full_fname = os.path.join(ds.fullpath, "swx_20120520_0641_grid.nc")
+    assert os.path.exists(full_fname)
```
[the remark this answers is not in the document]
chrishavlin: Ya, I wasn't sure about the best way to handle the file that gets written in this test, figured it'd come in review  :) 

So when this frontend loads a cf radial file that is not gridded it does some gridding and writes out a gridded file (all this happens in `CFRadialDataset.__init__`). Right now it just writes it alongside the base file  -- so maybe  what is needed is an argument to better control where this new gridded file gets written? is this what the `storage_filename` argument is supposed to be for? Then this test could write to a temporary directory.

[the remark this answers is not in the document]
neutrinoceros: I've never needed the `storage_filename` argument, but it sounds like a job for it indeed. @matthewturk , can you confirm ?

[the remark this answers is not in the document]
matthewturk: Yes, but I think it's not used much.

yt/frontends/cf_radial/data_structures.py
chrishavlin: Thought a bit more about this and looked through `FieldInfoContainer` again. Maybe what I'm doing here actually belongs in `CFRadialFieldInfo.setup_fluid_fields()` with calls to `CFRadialFieldInfo.add_output_field()`.

neutrinoceros: I remember doing something hacky with dust-related fields in the AMRVAC frontend (since there can be an arbitrary number of them). I think that the field recognition also isn't flexible enough for the real world but obviously changing it is way beyond the scope of this PR. In any case I could be inclined to approve of a hack if only it happened closer to where the problem lies (the FieldInfo subclass). Do you think it's possible ?

chrishavlin: yes! just worked out how to do this in `setup_fluid_fields` in a cleaner way similar to other frontends... Will push it up soon!

yt/frontends/cf_radial/tests/test_outputs.py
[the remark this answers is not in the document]
chrishavlin: Latest push should resolve this.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+        # fluid type is usually the dataset_type and the on-disk particle type
+        # (for a single population of particles) is "io".
+        self.field_list = []
+
+        # pull out some info for dynamically setting fields and units
+        fic = self.ds._field_info_class
+        known_fieldnames = [fld[0] for fld in fic.known_other_fields]
+
+        flds_to_add = tuple()  # entries to add to known_other_fields
+        warn_units = 0  # flag if we encounter units we cannot handle
```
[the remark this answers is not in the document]
chrishavlin: cleaned this up, should be resolved

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+        for key in self.ds._handle.variables.keys():
+            if (
+                all(x in self.ds._handle[key].dims for x in ["time", "z", "y", "x"])
+                is True
+            ):
+                field_tup = ("cf_radial", key)
+                warn_units += key in fic.dBz_fields
```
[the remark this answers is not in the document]
chrishavlin: cleaned this up, should be resolved

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+                self.field_list.append(field_tup)
+                if key not in known_fieldnames:
+                    # this field is not in known fields, add it now after some
+                    # sanitization of the units
+                    units = self.ds._handle[key].units
+
+                    if units in fic.units_not_handled:
+                        units = ""
+                        warn_units += 1
```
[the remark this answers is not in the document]
chrishavlin: cleaned this up, should be resolved

yt/frontends/cf_radial/data_structures.py
chrishavlin: Ok, this might be resolved now. it's at least a lot cleaner and follows what other frontends do.

yt/frontends/cf_radial/io.py
chrishavlin: ya, I'm not sure. These netcdf files will only have a single grid in a single chunk, so probably no performance issue in this case. We convert our chunk iterators to lists in many places throughout the code base, and it's not always clear to me why...

yt/frontends/cf_radial/io.py
chrishavlin: I don't follow this -- these netcdf files are what they are, we won't have control over the ordering they're written in.

yt/frontends/cf_radial/data_structures.py
chrishavlin: point taken! I just pushed a version with `if` statements for the Conventions check, but note that in cf-netcdf world, `ds.attrs` is a dictionary, so the check has to be a bit different from your suggestion.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+    def _populate_grid_objects(self):
+        # For each grid, this must call:
+        #   grid._prepare_grid()
+        #   grid._setup_dx()
+        # This must also set:
+        #   grid.Children <= list of child grids
+        #   grid.Parent   <= parent grid
+        # This is handled by the frontend because often the children must be
+        # identified.
+        pass
```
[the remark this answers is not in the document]
chrishavlin: cleaned this up so that `_populate_grid_objects` is now used (moved the `_prepare_grid` and `_setup_dx` from where it was in `_parse_index` to here.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+        #   self.max_level = self.grid_levels.max()
+        self.grid_left_edge[0][:] = self.ds.domain_left_edge[:]
+        self.grid_right_edge[0][:] = self.ds.domain_right_edge[:]
+        self.grid_dimensions[0][:] = self.ds.domain_dimensions[:]
+        self.grids = np.empty(self.num_grids, dtype="object")
+        for i in range(self.num_grids):
+            g = self.grid(i, self, self.grid_levels.flat[i], self.grid_dimensions[i])
+            g._prepare_grid()
+            g._setup_dx()
+            self.grids[i] = g
```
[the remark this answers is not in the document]
chrishavlin: removed the loop (and moved this down to `_populate_grid_objects` where I think it belongs)

yt/frontends/cf_radial/io.py
chrishavlin: removed

yt/frontends/cf_radial/fields.py
chrishavlin: removed

yt/frontends/cf_radial/data_structures.py
chrishavlin: this was set higher up... deleted that line, uncommented this one.

yt/frontends/cf_radial/data_structures.py
neutrinoceros: oh right, my mistake, I think I misread this.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,178 @@
...
+                ds = xr.open_dataset(filename)
+            except (OSError, AttributeError, TypeError):
+                # catch all these to avoid errors when xarray cant handle a file
+                return False
+
+            if hasattr(ds, "attrs") and isinstance(ds.attrs, dict):
+                conventions += ds.attrs.get("Conventions", "")
+                conventions += ds.attrs.get("conventions", "")
+
+        return "CF/Radial" in conventions
```
neutrinoceros, line 178: I tried to make this flatter, not crucial
```suggestion
        if not xr._module:
            return False
        try:
            ds = xr.open_dataset(filename)
        except (OSError, AttributeError, TypeError):
            # catch all these to avoid errors when xarray cant handle a file
            return False

        if hasattr(ds, "attrs") and isinstance(ds.attrs, dict):
            return "CF/Radial" in ds.attrs.get("Conventions", "") + ds.attrs.get("conventions", "")
        return False
```

yt/frontends/cf_radial/io.py
neutrinoceros: in this particular case `chunks` is only used as an itertable, so I think the conversion isn't necessary.

yt/frontends/cf_radial/io.py
neutrinoceros: Honestly I'll admit I might be very confused about this. I _think_ transpositions are very rarely necessary because oftentimes they are used to compensate for arrays that are read in the wrong order ("F" VS "C"). This might be either wrong or irrelevant here, I'm sorry, never managed to properly get my head around this problem.

yt/frontends/cf_radial/data_structures.py
chrishavlin: nice! will add it in after I see how the tests do...

chrishavlin: well I ended up adjusting `is_valid` a couple times -- didn't realize that `xr._module` would raise an `ImportError` if there is no `xarray`. The code now is fairly compact (and works!).

yt/frontends/cf_radial/io.py
chrishavlin: no worries! I find it very confusing as well... I always end up with a lot of trial and error with netcdf index and coordinate ordering... I do think it's needed in this case though.

yt/frontends/cf_radial/io.py
chrishavlin: Ok, will test it out. I agree it's probably not necessary here...

chrishavlin: (ya, not necessary, took out the conversion)

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,269 @@
...
+        return "CFRadialGrid_%04i (%s)" % (self.id, self.ActiveDimensions)
+
+
+class CFRadialHierarchy(GridIndex):
+    grid = CFRadialGrid
+
+    def __init__(self, ds, dataset_type="cf_radial"):
+        self.dataset_type = dataset_type
+        self.dataset = weakref.proxy(ds)
+        # for now, the index file is the dataset!
```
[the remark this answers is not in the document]
chrishavlin: this was likely a long term decision: I suppose if the data were split up across netcdf files on their own grids this might change? But that seems unlikely and would require refactoring large portions of this frontend... So I adjusted this comment.

munkm (commented):
A few initial questions but overall this is looking great!

yt/frontends/cf_radial/definitions.py
```diff
@@ -0,0 +1 @@
+# This file is often empty.  It can hold definitions related to a frontend.
```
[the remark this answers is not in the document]
munkm: I guess most of the frontends keep it so nothing needs to happen in this PR to this file.

yt/frontends/cf_radial/tests/test_outputs.py
```diff
@@ -0,0 +1,154 @@
...
+    domain_left_array = [-50000.0, -50000.0, 0.0]
+    assert_equal(ds.domain_left_edge, domain_left_array)
+
+
+@requires_file(cf)
+@requires_module("xarray")
+def test_domain_right_edge():
+    ds = data_dir_load(cf)
+    domain_right_array = [50000.0, 50000.0, 15000.0]
+    assert_equal(ds.domain_right_edge, domain_right_array)
```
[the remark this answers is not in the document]
chrishavlin: good idea! I can put this change in.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,176 @@
...
+class CFRadialDataset(Dataset):
+    _index_class = CFRadialHierarchy
+    _field_info_class = CFRadialFieldInfo
+
+    def __init__(
+        self,
+        filename,
+        dataset_type="cf_radial",
+        grid_shape=(40, 200, 200),
+        grid_limits=((0.0, 2000.0), (-1e5, 1e5), (-1e5, 1e5)),
```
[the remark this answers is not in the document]
chrishavlin: Could use some input from Zach here -- I wonder if the grid_limits at least could be pulled (or calculated) from the ungridded netcdf file. Seems reasonable to me that the `grid_shape` would be user configurable, but it'd be nice to be able to not have to set the `grid_limits` if possible.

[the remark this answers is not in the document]
zssherman: @chrishavlin @munkm So, this is kinda tricky, most files at least the many that I've used don't really have a set grid limits or anything to pull from, this usually has always been up to the users because different radars have different settings of range, users don't want to include certain returns, etc. The height is really low, can't remember why we set it to those numbers specifically, might have been for testing purposes when we all were originally working on this code. One way is maybe that when reading in the ungridded file with pyart we do for a default:
max_range = radar.range['data'].max()
max_height = radar.altitude['data'].max()
grid_limits = ((0.0, max_height), (-max_range, max_range), (-max_range, max_range))

Another note: In Py-ART we don't have defaults typically, the user has to provide their own grid shape and limits unique to their case, what they want to accomplish etc otherwise error for missing parameter. So we can maybe add an error if they don't have something specified or go with a default based on the max distances of the radar.

yt/frontends/cf_radial/tests/test_outputs.py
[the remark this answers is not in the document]
chrishavlin: combined now!

yt/utilities/on_demand_imports.py
```diff
@@ -251,6 +251,43 @@ def pooch(self):
...
+    def __init__(self):
+        try:
+            import xarray as myself
+
+            self._module = myself
+        except ImportError:
+            self._module = NotAModule(self._name)
+
+    def __getattr__(self, attr):
+        return getattr(self._module, attr)
```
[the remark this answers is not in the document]
chrishavlin: now just importing `open_dataset`. I also updated the `pyart` import to only import the `io` and `map` modules that we're using.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,176 @@
...
+
+        self.domain_left_edge = np.array([x.min(), y.min(), z.min()])
+        self.domain_right_edge = np.array([x.max(), y.max(), z.max()])
+        self.dimensionality = 3
+        dims = [self._handle.dims[d] for d in "xyz"]
+        self.domain_dimensions = np.array(dims, dtype="int64")
+        self._periodicity = (False, False, False)
+        self.current_time = float(self._handle.time.values)
+
+        # Cosmological information set to zero (not in space).
```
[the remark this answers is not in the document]
chrishavlin: i guess.... like... earth is in space?

yt/frontends/cf_radial/data_structures.py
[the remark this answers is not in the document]
chrishavlin: ok, thanks for input, Zach! Maybe what we should do is set the default values to None here, and then if the file is not gridded, we raise an error if the `grid_shape` and `grid_limit` variables are not set (with an error message along the lines of "This file is not yet gridded, please provide the grid_shape and grid_limit arguments")?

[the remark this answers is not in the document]
chrishavlin: It also might make sense to have a dataset method in our `CFRadialDataset` class that wraps the `grid_from_radars` call (essentially move what we have in the `__init__` to it's own method). That would allow someone to easily re-grid without re-loading a dataset. So you could do something like:

```python
import yt
ds = yt.load("ungridded_cf_radial.nc", grid_limits = (blah), grid_shape = (more blah))

<< do some things, decide to regrid>>

ds.grid_from_radars(grid_shape, grid_limit) 

<< do more things with new grid >>
```

zssherman:
@chrishavlin @neutrinoceros Was going through some old emails, and forgot about this, is there anything else that needs to be done for merging this? I see a new conflict.

chrishavlin:
Hi @zssherman ! Thanks for the ping -- yes, there was a remaining issue with how to handle the regridding on initial load. I had some ideas, but let me go back to my notes and report back to make sure I'm remembering properly (I'll follow up in the next few days).

neutrinoceros:
Whoops, I'm so sorry I didn't follow this up... I'm going to fix the one conflict that we have now and refresh CI.

neutrinoceros:
pre-commit.ci autofix

chrishavlin:
> Whoops, I'm so sorry I didn't follow this up... I'm going to fix the one conflict that we have now and refresh CI.

oof, me too! I let this fall off my radar. This would be great to get through! I'll look back at the bit I was uncertain on (had to do with how the initial re-gridding works, but I need to refresh my memory).

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def __init__(self, id, index, level, dimensions):
+        super().__init__(id, filename=index.index_filename, index=index)
+        self.Parent = None
+        self.Children = []
+        self.Level = level
+        self.ActiveDimensions = dimensions
+
+    def __repr__(self):
+        return "CFRadialGrid_%04i (%s)" % (self.id, self.ActiveDimensions)
```
neutrinoceros, line 33: quick note: this becomes unneeded with #3936

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,173 @@
...
+    def _detect_output_fields(self):
+        # This sets self.field_list, containing all the available on-disk fields and
+        # records the units for each field.
+        self.field_list = []
+        units = {}
+        for key in self.ds._handle.variables.keys():
+            if (
+                all(x in self.ds._handle[key].dims for x in ["time", "z", "y", "x"])
+                is True
+            ):
```
neutrinoceros, line 61:
```suggestion
            if all(x in self.ds._handle[key].dims for x in ["time", "z", "y", "x"]):
```

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,173 @@
...
+        # our index file is the dataset itself:
+        self.index_filename = self.dataset.parameter_filename
+        self.directory = os.path.dirname(self.index_filename)
+        # float type for the simulation edges and must be float64 now
+        self.float_type = np.float64
+        super().__init__(ds, dataset_type)
+
+    def _initialize_state_variables(self):
+        super()._initialize_state_variables()
+
```
neutrinoceros, line 51: what's the rationale for having this method ? in this state it's equivalent to not overriding the parent class' method
```suggestion
```

yt/frontends/cf_radial/fields.py
```diff
@@ -0,0 +1,37 @@
...
+    def setup_fluid_fields(self):
+        # Here we dynamically add fields available in our netcdf file for to the
+        # FieldInfoContainer with sanitized units.
+
+        for field in self.field_list:  # field here is ('fluid_type', 'field) tuple
+            units = self.ds.field_units.get(
+                field, ""
+            )  # where does sanitization happen??
+
+            # sanitization of the units
```
neutrinoceros, line 29: I'm confused. Looks like the second comment immediately answers the first. Is it the case ?

yt/frontends/cf_radial/io.py
neutrinoceros: I don't think this should be a blocker, I was overcorrecting for my lack of understanding of what numpy does with transposition. Let's just resolve this thread

yt/utilities/on_demand_imports.py
```diff
@@ -249,6 +249,55 @@ def __getattr__(self, attr):
 _pooch = pooch_imports()
 
 
+class pyart_imports:
```
neutrinoceros, line 252: Note: I worked out a way to simplify this part in https://github.com/yt-project/yt/pull/3935
If my PR gets merged first (or ever), I'm happy to help rewriting this (mostly) boilerplate code

neutrinoceros (dismissed):
Mostly superficial questions and remarks, I think this is very close to a mergeable state AFAICT.

yt/utilities/on_demand_imports.py
neutrinoceros: done

chrishavlin:
Hey @zssherman ! I've been working to get this through (at last! apologies for the **long** delay on this...). 

When we last left off, the remaining issue was just the question of initial re-gridding parameters when necessary. In our earlier discussion you had suggested

> One way is maybe that when reading in the ungridded file with pyart we do for a default:
max_range = radar.range['data'].max()
max_height = radar.altitude['data'].max()
grid_limits = ((0.0, max_height), (-max_range, max_range), (-max_range, max_range))

I went ahead and tried to code that up -- didn't **quite** work out as simple as that, but I think I arrived at a good solution. Ok if I push up those changes to your branch?

zssherman:
@chrishavlin No worries at all! And sounds good you can push to the branch, I appreciate your work on thus! I would try to assist more, but on vacation at the moment.

chrishavlin:
Great! I'll push up some changes in the next few days and point out some places in particular that could use your input. So whenever you're back and available we should be able to polish it off pretty quickly! (I'll also get tests passing and address the latest comments from @neutrinoceros if I can).

yt/frontends/cf_radial/data_structures.py
chrishavlin: Probably just leftovers from initially building out the frontend. I think it's ok to delete!

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,309 @@
...
+    ) -> Tuple[Tuple, Tuple]:
+        # this function sets all the re-gridding parameters if they are not set
+
+        # radar is a pyart radar handle, from pyart.io.read_cfradial()
+        max_range_meters = radar.range["data"].max()  # along-ray distance
+        elevation = radar.elevation["data"]  # angle above horizontal
+        if radar.elevation["units"] == "degrees":
+            elevation = elevation * np.pi / 180.0
+        max_height = max_range_meters * np.sin(elevation.max())
+        max_distance = max_range_meters * np.cos(elevation.max())
```
chrishavlin, line 218: @zssherman this bit is what I'm not sure of...

The test dataset we have only has a single value for altitude:

```python
import pyart
ds_pyart = pyart.io.read_cfradial("swx_20120520_0641.nc")
print(ds_pyart.altitude)

{'units': 'meters',
 'standard_name': 'Altitude',
 'data': masked_array(data=[214.],
              mask=False,
        fill_value=1e+20)}
```
I thought maybe this was the instrument altitude, not the measurement coordinate?

And based on my reading of the coordinate info in `ds_pyart.info()`, poking around the pyart docs and skimming the CFRadial file format documentation, I understood elevation, azimuth and range to be a spherical coordinate system.. and so I ignored the azimuth and just used the extents of range and elevation to calculate the z and x/y ranges. Could **totally** be wrong about all of this though :)

yt/frontends/cf_radial/fields.py
chrishavlin: I think so, I removed this in the latest batch

chrishavlin:
Oops, forgot to run mypy locally... I'll fix those errors soon.

zssherman:
@chrishavlin Looks like I couldn't comment on the code comment itself, i'll post here. I realized earlier when I suggested using altitude, that it was a mistake on my part. Altitude as you mentioned is just the radar instrument altitude. What I should have suggested was radar.gate_altitiude['data']. Radar gate_altitude is the actually location of the radar gate, point of data in meters, so the max radar.gate_altitude['data'] should give the highest height of were data is at along a ray. After talking with @rcjackson, for the - and + max range, be ideal to round to the next 10km so if -147 147, the min and max would be -150, 150 utilizing radar.range['data']. Then py-arts gridding converts all of this to the Cartesian grid coordinates.

chrishavlin:
> @chrishavlin Looks like I couldn't comment on the code comment itself, i'll post here. I realized earlier when I suggested using altitude, that it was a mistake on my part, good catch! Altitude as you mentioned is just the radar instrument altitude. What I should have suggested was radar.gate_altitiude['data']. Radar gate_altitude is the actually location of the radar gate, point of data in meters, so the max radar.gate_altitude['data'] should give the highest height of were data is at along a ray. After talking with @rcjackson, for the - and + max range, be ideal to round to the next 10km so if -147 147, the min and max would be -150, 150 utilizing radar.range['data']. Then py-arts gridding converts all of this to the Cartesian grid coordinates.

Great! Many thanks for the clarification! I'll update my changes. I'm also working on getting the tests passing... almost there!

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,329 @@
...
+class CFRadialDataset(Dataset):
+    _index_class = CFRadialHierarchy
+    _field_info_class = CFRadialFieldInfo
+
+    def __init__(
+        self,
+        filename,
+        dataset_type="cf_radial",
+        storage_filename: Optional[str] = None,
+        storage_overwrite: Optional[bool] = False,
```
neutrinoceros, line 90: Just noticed that. Is there a special meaning to `None` here ? I'd assume only booleans are actually valid
```suggestion
        storage_overwrite: bool = False,
```

chrishavlin: good catch! no, no special meaning

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,329 @@
...
+            yield xrds
+
+    def _validate_grid(
+        self,
+        radar,
+        grid_shape: Optional[Tuple[int, int, int]] = None,
+        grid_limit_x: Optional[Tuple[float, float]] = None,
+        grid_limit_y: Optional[Tuple[float, float]] = None,
+        grid_limit_z: Optional[Tuple[float, float]] = None,
+    ) -> Tuple[Tuple, Tuple]:
```
neutrinoceros, line 210: Note that this is implicitly equivalent to `Tuple[Tuple[Any, ...], Tuple[Any, ...]]` and I think that at least one of the errors reported by mypy is a direct consequence of this type being too loose

chrishavlin: Ok! Ya, this is causing issues... but I finally got mypy working again locally after some issues with my environment, so next push should hopefully resolve the type checking. Will wait to see if all the builds pass though... I **think** I fixed the remaining issues that were causing those to fail.

chrishavlin:
OK, I fixed the typing issues locally but haven't pushed yet. The failing answer test is cause I changed the tests for this frontend and forgot to update the answer test config -- I think I'll fix this by creating a new answer test with an actual image comparison. 

The other failure is more complicated -- it's from an internal warning deep in xarray. it's a version-dependent issue and I'm not sure the best way to deal with it. It only comes when yt calls this frontend's `_is_valid()` with data that xarray can't handle and I believe our tests ultimately fail because pytest is set to fail on warnings. I'll investigate further and get back with a fix soon hopefully. 

**almost** there!

neutrinoceros:
> The other failure is more complicated -- it's from an internal warning deep in xarray. it's a version-dependent issue

Whose version does it depend on and how ?

>It only comes when yt calls this frontend's _is_valid() with data that xarray can't handle and I believe our tests ultimately fail because pytest is set to fail on warnings.

Correct. If there's no better way we could filter exactly this warning _locally_ using a context manager `with warnings.filerwarnings(...):`
If it only happens with old versions of xarray we can make it conditional as we do for Matplotlib. If it happens with recent versions, it could be worth opening an issue upstream.

chrishavlin:
Turns out it was an easy fix -- using `xr.open_dataset(..., engine="netcdf4")` raises an error without the warnings. I also realized that I could have added `RuntimeWarning` to the long list of exceptions we catch here, but ultimately decided on using the `engine` parameter since this frontend is loading netcdf files and if someone does not have netcdf4 installed a ValueError gets raised (which we're already catching). 

The latest batch of commits also includes:
* updated auto-gridding params using the proper `gate_altitude`, including the suggested rounding to the nearest +/-10km
* reorganization of the auto-gridding methods -- I was having a hard time with type checking, it got easier when I split up the one big `_validate_grid` function to a couple smaller ones. 
* a new answer test -- went with a grid/field values answer test instead of an image test. Realizing just now I didn't bump the answer test number but I'll let all the other tests run before pushing anything.

neutrinoceros:
The new warning look like a self-triggered DeprecationWarning in xarray. The best way to handle them is to 
- report upstream
- add a filter in conftest.py with a comment linking to the issue

chrishavlin:
[looks like it was already reported and fixed upstream](https://github.com/pydata/xarray/issues/6514), so xarray's next release should fix it. But on yt's side, adjusting conftest.py would get tests passing, but since this is triggered by calls to `_is_valid()`, wouldn't the warning messages still display? I've actually been trying to reproduce this locally to see but haven't succeeded yet (doesn't seem to be as simple as python >= 3.10 and xarray <= 2022.03.0).

yt/frontends/cf_radial/data_structures.py
chrishavlin: Latest commits split up this function and it made the typing much easier to get passing mypy. I'll go ahead and resolve the conversation since this section of code is outdated now, but feel free to suggest more typing improvements if you have ideas.

neutrinoceros:
@chrishavlin sounds good, thank you for documenting a removal condition for the filter too !
Seems like a conflict is preventing tests to run again, however

chrishavlin:
I also had messed up the conftest... will fix the conflict now too.

yt/utilities/on_demand_imports.py
```diff
@@ -204,6 +204,36 @@ def create(self):
 
 _pooch = pooch_imports()
 
+_pooch = pooch_imports()
```
chrishavlin, line 207: oops, looks like a merge or copy/paste mistake here. will let tests finish before fixing this...  
```suggestion
_pooch = pooch_imports()
```

yt/frontends/cf_radial/fields.py
```diff
@@ -0,0 +1,36 @@
...
+    field_units_ignored: List[str] = []  # fields for which units have been ignored
+
+    # (find, replace) pairs for sanitizing:
+    unit_subs = (("degrees", "degree"), ("meters", "m"), ("_per_", "/"))
+
+    def setup_fluid_fields(self):
+        # Here we dynamically add fields available in our netcdf file for to the
+        # FieldInfoContainer with sanitized units.
+
+        for field in self.field_list:  # field here is ('fluid_type', 'field) tuple
```
neutrinoceros, line 25:
```suggestion
        for field in self.field_list:
        # field here is ('fluid_type', 'field') tuple
```

yt/frontends/cf_radial/fields.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+from typing import List
+
+from yt.fields.field_info_container import FieldInfoContainer
+
+
+class CFRadialFieldInfo(FieldInfoContainer):
+    known_other_fields = ()  # fields are set dynamically
+    known_particle_fields = ()
+    units_to_ignore = ("dBz", "dBZ", "ratio")  # set as nondimensional if found
```
neutrinoceros, line 15: why ?

neutrinoceros (approved):
Overall LGTM. Only found one typo and one question. I think this is probably ready to go now.

yt/frontends/cf_radial/fields.py
chrishavlin: If I remember right, it's just that these units are complex (e.g., https://en.wikipedia.org/wiki/DBZ_(meteorology) ) and it wasn't obvious how to handle them... but I don't quite remember why the `units_to_ignore` attribute was added as opposed to simply adding them as dimensionless fields in `known_other_fields`. Might have been just to have an explicit record of how the fields were not being handled well.  

I'm pretty sure they are in fact all dimensionless units so I think we could define them in the `CFRadialFieldInfo.__init__()` method (e.g., `self.ds.define_unit("dBZ", 1*unyt.dimensionless)`). But I also remember this being an issue with the nc4_cm1 frontend (and it may have actually been handled differently), so I think it'd be better to improve this together with nc4_cm1 in a followup (I can open an issue to track it).

neutrinoceros: Sounds good. Thank you !

chrishavlin: opened! https://github.com/yt-project/yt/issues/3972 I'll let you resolve the thread

neutrinoceros:
Ah ! I'm sorry I should have pointed this out earlier but we also need to update the following docs:

doc/source/reference/code_support.rst
doc/source/reference/api/api.rst
doc/source/examining/loading_data.rst

Beyond that I think we're done !

doc/source/examining/loading_data.rst
```diff
@@ -3082,3 +3082,33 @@ non-default cosmological parameters, you may pass an empty dictionary.
    import yt
 
    ds = yt.load(filename, cosmology_parameters={})
+
+.. _loading-cfradial-data:
+
+CfRadial Data
+-------------
+
+yt provides support for loading cartesian-gridded CfRadial netcdf-4 files. When
```
chrishavlin, line 3091: @zssherman want to add a sentence or adjust the wording here to describe what a CfRadial dataset is?

chrishavlin: oops, meant to do a multi-line select there... you'll wanna open up the full file to read the full description.

chrishavlin:
Ok, updated the docs. Let's give @zssherman a chance to look over my description of CfRadial data (which was intentionally lacking details...).

chrishavlin:
oh interesting. the answer tests are missing their answers? I'm not sure why that's the case... but I think they should pass on re-running. But we can wait on any changes to the docs (or confirmation that they're good as is) before triggering.

zssherman:
@chrishavlin No worries! So more details on cfradial here:
[Cf/Radial](http://www.ral.ucar.edu/projects/titan/docs/radial_formats/cfradial.html) is a CF compliant netCDF convention for radial data from radar and lidar platforms that supports both airborne and ground-based sensors. Because of its CF-compliance, CfRadial will allow researchers familiar with CF to read the data into a wide variety of analysis tools, models etc. Essential the data is in polar coordinates. Utilizes range, azimuth and elevation in lat lon alt, and is then converted to x,y,z Cartesian. Py-ART's gridding uses interpolation to plot that data to a new grid. I can add this to the PR, if need be, would it be under that rst file you linked?

chrishavlin:
Great! Thanks, @zssherman 

Ya, if you could edit docs/source/examining/loading_data.rst so that it has more of that context/background that would be great! You should be able to just pull while on your cf-radial branch and get all the changes I've pushed in the past ~week. I'm also happy to take a stab at integrating your overview into what I wrote initially, but since you're the expert on the data here I'd rather defer to you. Just let me know! 

For context, the current version of that page is rendered at https://yt-project.org/doc/examining/loading_data.html and it's where we mention all the various formats (the "frontends") that yt handles. For most of the yt frontends, it just gets straight into how to load data (I think it's assumed that folks trying to load their data know the background of their code/data source). But I'd prefer to follow the example of the Exodus II description on that page, where it starts with a sentence describing what the data format is.

neutrinoceros:
> the answer tests are missing their answers? I'm not sure why that's the case... 

this is fixed by bumping the answers in `tests/tests.yaml`

> I think it's assumed that folks trying to load their data know the background of their code/data source

correct. This page is more oriented toward specificities of loading data _with respect to the common rules in yt_. Most frontends require only a path to load, but there are cases where that's not sufficient.

zssherman:
@chrishavlin I'll try to add in the doc stuff tomorrow, sadly got swamped.

chrishavlin:
Great, thanks @zssherman ! I'll fix the final issues with the testing after you get those docs changes in.

zssherman:
@chrishavlin I apologize for the lack of knowledge on this, but how would I update my local to this PR? Was going to update the docs but having conflict issues.

chrishavlin:
@zssherman I messaged you on the yt slack -- if you run into more issues feel free to reach out there. Totally ok to keep discussion here as well, but slack is sometimes easier for quick back and forth.

doc/source/examining/loading_data.rst
```diff
@@ -3082,3 +3082,43 @@ non-default cosmological parameters, you may pass an empty dictionary.
...
+.. _cfradial_gridding:
+
+Gridding Behavior
+^^^^^^^^^^^^^^^^^
+
+When you load a CfRadial dataset in polar coordinates (elevation, azimuth and
+range), yt will first build a sample by mapping the data onto a cartesian grid
+using the Python-ARM Radar Toolkit (pyart). Grid points are found by interpolation of all data points within a specified radius of influence. This data, now
+in x, y, z coordiante domain is then saved as a new dataset and subsequent
+loads will load the original native CfRadial dataset will use the gridded file.
```
chrishavlin, line 3117: typo (let's wait until the current tests finish before committing this)
```suggestion
loads of the original native CfRadial dataset will use the gridded file.
```

doc/source/examining/loading_data.rst
```diff
@@ -3082,3 +3082,43 @@ non-default cosmological parameters, you may pass an empty dictionary.
...
+   ds = yt.load("CfRadialGrid/grid1.nc")
+
+.. _cfradial_gridding:
+
+Gridding Behavior
+^^^^^^^^^^^^^^^^^
+
+When you load a CfRadial dataset in polar coordinates (elevation, azimuth and
+range), yt will first build a sample by mapping the data onto a cartesian grid
+using the Python-ARM Radar Toolkit (pyart). Grid points are found by interpolation of all data points within a specified radius of influence. This data, now
```
matthewturk, line 3115:
```suggestion
using the Python-ARM Radar Toolkit (`pyart<https://github.com/ARM-DOE/pyart>`_). Grid points are found by interpolation of all data points within a specified radius of influence. This data, now
```

doc/source/reference/code_support.rst
```diff
@@ -30,6 +30,8 @@ each supported output format using yt.
 +-----------------------+------------+-----------+------------+-------+----------+----------+------------+-------------+
 | Castro                |     Y      |  Y [#f3]_ |   Partial  |   Y   |    Y     |    Y     |     N      |   Full      |
 +-----------------------+------------+-----------+------------+-------+----------+----------+------------+-------------+
+| CfRadial              |    N/A     |    N/A    |     N/A    |   Y   |    Y     |    Y     |     Y      |   [#f5]_    |
```
matthewturk, line 33: So this is strange, but I think that the way it's written these are correct, but for the intent of the table the first should probably be "Yes", N/A, and I *think* "Yes"?  Do we provide any metadata?

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,315 @@
...
+    def _count_grids(self):
+        self.num_grids = 1
+
+    def _parse_index(self):
+        self.grid_left_edge[0][:] = self.ds.domain_left_edge[:]
+        self.grid_right_edge[0][:] = self.ds.domain_right_edge[:]
+        self.grid_dimensions[0][:] = self.ds.domain_dimensions[:]
+        self.grid_particle_count[0][0] = 0
+        self.grid_levels[0][0] = 1
+        self.max_level = 1
```
matthewturk, line 71: There seem to be some varying opinions on this among the frontends, but I think this should be 0.

```suggestion
        self.max_level = 0
```

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def _count_grids(self):
+        self.num_grids = 1
+
+    def _parse_index(self):
+        self.grid_left_edge[0][:] = self.ds.domain_left_edge[:]
+        self.grid_right_edge[0][:] = self.ds.domain_right_edge[:]
+        self.grid_dimensions[0][:] = self.ds.domain_dimensions[:]
+        self.grid_particle_count[0][0] = 0
+        self.grid_levels[0][0] = 1
```
matthewturk, line 70:
```suggestion
        self.grid_levels[0][0] = 0
```

I think this should be zero as well (for same reason as below).

I think this is likely only going to be relevant in cases where we use the `domain_dimensions` and the level to compute the grid size.  I believe in most places it will be done via `dims / (refine_factor ** level)` which here would give the wrong result.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,315 @@
...
+                raise RuntimeError("Dataset has no filename yet.")
+
+        with xr.open_dataset(filename) as xrds:
+            yield xrds
+
+    def _validate_grid_dim(
+        self, radar, dim: str, grid_limit: Optional[Tuple[float, float]] = None
+    ) -> Tuple[float, float]:
+        if grid_limit is None:
+            if dim == "z":
```
matthewturk, line 209: will this always be lowercase?  I suspect so, but want to check.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -0,0 +1,315 @@
...
+        self.omega_matter = 0.0
+        self.hubble_constant = 0.0
+
+    @classmethod
+    def _is_valid(cls, filename, *args, **kwargs):
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+
+        try:
+            ds = xr.open_dataset(filename, engine="netcdf4")
```
matthewturk, line 297: Should we explicitly close this?

yt/frontends/cf_radial/io.py
```diff
@@ -0,0 +1,35 @@
...
+import numpy as np
+
+from yt.utilities.io_handler import BaseIOHandler
+
+
+class CFRadialIOHandler(BaseIOHandler):
+    _particle_reader = False
+    _dataset_type = "cf_radial"
+
+    def _read_fluid_selection(self, chunks, selector, fields, size):
```
matthewturk, line 17: I am genuinely pleased at how short this file is, given how complicated they often can be!

matthewturk (approved):
I have made a few suggestions, and asked a couple questions.  However, I am totally fine with accepting this as-is, although I do think I want to think about the level values at some point.  Nice work, everyone.  Thank you.

yt/frontends/cf_radial/data_structures.py
zssherman: I believe, if the files follow the cfradial conventions, they should always be lowercase.

matthewturk: sounds good!

doc/source/reference/code_support.rst
chrishavlin: > Do we provide any metadata?

Is that what would normally end up in `ds.parameters`? This frontend does implement a `_parse_parameter_file` and initializes an empty dict for `ds.parameters` ([link to src](https://github.com/zssherman/yt/blob/793a273c33f0ad85b5aa7aaae7144e3c8a21617f/yt/frontends/cf_radial/data_structures.py#L267)), but that method just sets all the domain-related attributes and doesn't actually put anything into `parameters`. So does that mean no metadata?

matthewturk: Yeah, let's leave this as-is.

yt/frontends/cf_radial/data_structures.py
zssherman: Are you thinking something along the lines of such?:

```
if hasattr(ds, "attrs") and isinstance(ds.attrs, dict):
    con = "Conventions"
    conv_check = "CF/Radial" in ds.attrs.get(con, "") + ds.attrs.get(con.lower(), "")
    ds.close()
    return conv_check
ds.close()
return False
```

chrishavlin: yes, I think we should... or change it to use `xr.open_dataset` as a context manager. I'll push up a change shortly.

doc/source/reference/code_support.rst
chrishavlin: changes I just pushed have Y, N/A, Y now. I do think the metadata handling could be improved here, but not in this PR. could copy over metadata into `parameters` or could expose the underlying xarray handle for exploration?

matthewturk: Sounds good.  Future!

yt/frontends/cf_radial/data_structures.py
chrishavlin: just pushed up a change (didn't see your comment til too late, @zssherman!). I ended up switching to `with xr.open_dataset(filename, engine="netcdf4") as ds:`  so that it'd be closed and shuffling things around to make that work.

zssherman: No worries! 'with' is better to use, don't know why I didn't think of that first.

chrishavlin:
looks like the updates to the grid_level values have changed the new answer tests. Not yet sure if the change is good or bad...

matthewturk:
Can we verify them?  It should be straightforward to do something like compute the volume, which might be related, or the path length.

chrishavlin:
Ok, ran some comparisons and I can confirm that the current version is correct. Easiest to see this with projection plots. With the current version of the code, the following

```python
import yt
ds = yt.load("CfRadialGrid/grid1.nc")
p = yt.ProjectionPlot(ds, "x", ("cf_radial", "reflectivity"), method="max")
p.set_log(("cf_radial", "reflectivity"), False)
p.set_cmap(("cf_radial", "reflectivity"), 'dusk')
```
results in this (very pretty) projection plot:

image

with the old version, the domain gets incorrectly scaled: 

image


I'll bump the answer tests now! 

One extra note: the projection plot above uses `method="max"` because the default (method='integrate') does not handle NaNs well and results in an all-NaN projection. See https://github.com/yt-project/yt/issues/3977

tests/tests.yaml
```diff
@@ -183,6 +183,9 @@ answer_tests:
   local_nc4_cm1_001: # PR  2176
     - yt/frontends/nc4_cm1/tests/test_outputs.py:test_cm1_mesh_fields
 
+  local_cf_radial_001: # PR  1990
```
chrishavlin, line 186:
```suggestion
  local_cf_radial_002: # PR  1990
```

neutrinoceros (approved):
Alright, I think this is now complete. I will let you guys deal with your remaining threads if you want to add anything still, but AFAIC it's good to go.
Great job everyone !

chrishavlin:
Just resolved the remaining threads -- @matthewturk wanna refresh your review?

zssherman:
Thank you all for all your help with this!

matthewturk:
Thank *you* for your patience and hard work!

PR #1991, opened 2018-08-28
fix to_glue method, add test

ngoldbaum:
The glue internal API changed, this adds a fix that avoids the `ImportError` that currently gets generated with the latest version of glue.

Since glue is a QT gui app, it's not straightforward to directly test this. Instead we're using the dummy base glue application class within testing to mock the full GUI app. This isn't perfect and we may want to rethink this if/when integration gets more extensive but this should allow us to detect breakage like changed internal imports in glue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ env:
     NOSETIMER=nose-timer
     REQUESTS=requests
     GLUE=glueviz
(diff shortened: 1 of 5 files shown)


ngoldbaum:
github and travis are having issues at the moment :/

PR #1998, opened 2018-08-30
[bugfix] min/max location must use different signs for "huge" values

jzuhone:
To find the location of the max of a field in the `SampleAtMaxFieldValues` class, we fill an array with `-HUGE` values and then fill this array with real data from chunks. If a chunk happens to have no values of the particular field, then the `-HUGE` values stay, which is fine because we are trying to find the maximum.

The problem is that `SampleAtMinFieldValues` inherits from this class but only changes the function from `argmax` to `argmin` without recognizing that the sign of the value that serves as the limit also needs to change for the operation to work correctly in case a chunk has no data.

In yt-4.0 this was manifesting itself in a weird way--trying to find the minimum potential of star particles from a Gadget simulation was giving the wrong answer because some chunks didn't have star particles, and therefore the min value was `-HUGE`. We don't really have chunking for particle data in 3.x, so it would be hard for this bug to manifest there. However, the origin of the bug is still in the 3.x series, and so it should be fixed here.

### DIFF
--- a/yt/data_objects/derived_quantities.py
+++ b/yt/data_objects/derived_quantities.py
@@ -531,6 +531,7 @@ class Extrema(DerivedQuantity):
                 for mis, mas in zip(values[::2], values[1::2])]
 
 class SampleAtMaxFieldValues(DerivedQuantity):
+    _sign = -1
     r"""
     Calculates the maximum value and returns whichever fields are asked to be
     sampled.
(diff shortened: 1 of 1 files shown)


ngoldbaum:
nice catch!

jzuhone:
@ngoldbaum this also failed for the same reason outside of this PR:

https://travis-ci.org/yt-project/yt/builds/422363360

ngoldbaum:
OK, we probably need to debug that then. Can you open an issue about that so I can refer to it in the future. It's a little obnoxious that this seems to only trigger on the travis mac builds.

PR #2004, opened 2018-09-05
[WIP] [bugfix] yt 4.0 deposit bug fix

### LINKED ISSUE #2003: error with demeshing yt in accessing deposited fields in arbitrary grids
### Bug report

**Bug summary**

When creating an arbitrary grid with demeshed yt, one cannot access the deposited gas mass from an SPH simulation. 

**Code for reproduction**
```
from __future__ import print_function
import yt

fname = '/users/desika.narayanan/Dropbox/yt_datasets/GadgetDiskGalaxy/snapshot_200.hdf5'

 
ds = yt.load(fname)

#from the notebook
center =  [ 31995.63476562,31473.6640625,28969.88671875]
arb_center = ds.arr(center,'code_length')
left_edge = arb_center-ds.quan(100,'kpc')
right_edge = arb_center+ds.quan(100,'kpc')

arb = ds.arbitrary_grid(left_edge, right_edge,dims=[128, 128, 128],field_parameters={'center': center})
mass = arb[('deposit', 'PartType0_mass')].in_units('Msun')

```

**Actual outcome**

```
In [8]: run arb
yt : [INFO     ] 2018-09-04 06:30:37,956 Calculating time from 3.448e-01 to be 1.108e+17 seconds
yt : [INFO     ] 2018-09-04 06:30:37,956 Assuming length units are in kpc/h (comoving)
yt : [INFO     ] 2018-09-04 06:30:37,968 Parameters: current_time              = 1.1075810732534829e+17 s
yt : [INFO     ] 2018-09-04 06:30:37,968 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2018-09-04 06:30:37,968 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2018-09-04 06:30:37,969 Parameters: domain_right_edge         = [64000. 64000. 64000.]
yt : [INFO     ] 2018-09-04 06:30:37,969 Parameters: cosmological_simulation   = 1
yt : [INFO     ] 2018-09-04 06:30:37,969 Parameters: current_redshift          = 1.8999965286929705
yt : [INFO     ] 2018-09-04 06:30:37,969 Parameters: omega_lambda              = 0.721
yt : [INFO     ] 2018-09-04 06:30:37,969 Parameters: omega_matter              = 0.279
yt : [INFO     ] 2018-09-04 06:30:37,969 Parameters: hubble_constant           = 0.7
yt : [INFO     ] 2018-09-04 06:30:38,150 Allocating for 1.191e+07 particles
Loading particle index: 100%|██████████████████| 19/19 [00:00<00:00, 216.84it/s]
Interpolating SPH field ('deposit', 'PartType0_mass'): 0it [00:00, ?it/s]---------------------------------------------------------------------------
NeedsGridType                             Traceback (most recent call last)
~/yt/yt/data_objects/data_containers.py in _generate_fluid_field(self, field)
    309         try:
--> 310             finfo.check_available(gen_obj)
    311         except NeedsGridType as ngt_exception:

~/yt/yt/fields/derived_field.py in check_available(self, data)
    205         for validator in self.validators:
--> 206             validator(data)
    207         # If we don't get an exception, we're good to go

~/yt/yt/fields/derived_field.py in __call__(self, data)
    451         if not getattr(data, '_spatial', False):
--> 452             raise NeedsGridType(self.ghost_zones,self.fields)
    453         if self.ghost_zones <= data._num_ghost_zones:

NeedsGridType: (0, None)

YTNonIndexedDataContainer                 Traceback (most recent call last)
~/Desktop/scratch/arb.py in <module>()
     14
     15 arb = ds.arbitrary_grid(left_edge, right_edge,dims=[128, 128, 128],field_parameters={'center': center})
---> 16 mass = arb[('deposit', 'PartType0_mass')].in_units('Msun')

~/yt/yt/data_objects/data_containers.py in __getitem__(self, key)
    255                 return self.field_data[f]
    256             else:
--> 257                 self.get_data(f)
    258         # fi.units is the unit expression string. We depend on the registry
    259         # hanging off the dataset to define this unit object.

~/yt/yt/data_objects/construction_data_containers.py in get_data(self, fields)
    684         if len(part) > 0 and len(alias) == 0:
    685             if(is_sph_field):
--> 686                 self._fill_sph_particles(fields)
    687                 for field in fields:
    688                     if field in gen:

~/yt/yt/data_objects/construction_data_containers.py in _fill_sph_particles(self, fields)
    756
    757                 pbar = tqdm(desc="Interpolating SPH field {}".format(field))
--> 758                 for chunk in self._data_source.chunks([field],"io"):
    759                     px = chunk[(ptype,'particle_position_x')].in_base("code").d
    760                     py = chunk[(ptype,'particle_position_y')].in_base("code").d

~/yt/yt/data_objects/data_containers.py in chunks(self, fields, chunking_style, **kwargs)
   1297                 continue
   1298             with self._chunked_read(chunk):
-> 1299                 self.get_data(fields)
   1300                 # NOTE: we yield before releasing the context
   1301                 yield self

~/yt/yt/data_objects/data_containers.py in get_data(self, fields)
   1398
   1399         fields_to_generate += gen_fluids + gen_particles
-> 1400         self._generate_fields(fields_to_generate)
   1401         for field in list(self.field_data.keys()):
   1402             if field not in ofields:

~/yt/yt/data_objects/data_containers.py in _generate_fields(self, fields_to_generate)
   1418                 fi = self.ds._get_field_info(*field)
   1419                 try:
-> 1420                     fd = self._generate_field(field)
   1421                     if fd is None:
   1422                         raise RuntimeError

~/yt/yt/data_objects/data_containers.py in _generate_field(self, field)
    292                 tr = self._generate_particle_field(field)
    293             else:
--> 294                 tr = self._generate_fluid_field(field)
    295             if tr is None:
    296                 raise YTCouldNotGenerateField(field, self.ds)

~/yt/yt/data_objects/data_containers.py in _generate_fluid_field(self, field)
    310             finfo.check_available(gen_obj)
    311         except NeedsGridType as ngt_exception:
--> 312             rv = self._generate_spatial_fluid(field, ngt_exception.ghost_zones)
    313         else:
    314             rv = finfo(gen_obj)

~/yt/yt/data_objects/data_containers.py in _generate_spatial_fluid(self, field, ngz)
    337                     o = self._current_chunk.objs[0]
    338                     if accumulate:
--> 339                         rv = self.ds.arr(np.empty(o.ires.size, dtype="float64"),
    340                                          units)
    341                         outputs.append(rv)

~/yt/yt/data_objects/data_containers.py in ires(self)
   1575         if self._current_chunk is None:
   1576             self.index._identify_base_chunk(self)
-> 1577         return self._current_chunk.ires
   1578
   1579     @property

~/yt/yt/geometry/geometry_handler.py in cached_func(self)
    267             return getattr(self, n)
    268         if self.data_size is None:
--> 269             tr = self._accumulate_values(n[1:])
    270         else:
    271             tr = func(self)

~/yt/yt/geometry/geometry_handler.py in _accumulate_values(self, method)
    295         for obj in self._fast_index or self.objs:
    296             f = getattr(obj, mname)
--> 297             arrs.append(f(self.dobj))
    298         if method == "dtcoords":
    299             arrs = [arr[0] for arr in arrs]

~/yt/yt/data_objects/particle_container.py in _func_non_indexed(self, *args, **kwargs)
     28 def _non_indexed(name):
     29     def _func_non_indexed(self, *args, **kwargs):
---> 30         raise YTNonIndexedDataContainer(self)
     31     return _func_non_indexed
     32

YTNonIndexedDataContainer: The data container (<class 'yt.data_objects.particle_container.ParticleContainer'>) is an unindexed type.  Operations such as ires, icoords, fcoords and fwidth will not work on it.
```

**Expected outcome**

**Version Information**

  * Operating System: OSX and/or Linux
  * Python Version: ```In [2]: sys.version
Out[2]: '3.6.5 |Anaconda, Inc.| (default, Apr 26 2018, 08:42:37) \n[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]'```
  * yt version: 4.0

installed from source

AshKelly:
## PR Summary

I've added some support such that `covering_grid` and `arbitrary_grid` in the demeshed branch can accept,
```python
arb[('deposit', 'PartType0_mass')]
```
as well as,
```python
arb[('gas', 'mass')] 
```

fixes #2003 

**This isn't currently working as expected**

## PR Checklist

- [ ] **NA** New features are documented, with docstrings and narrative docs
- [ ] **NA** Adds a test for any bugs fixed. Adds tests for new features.


ngoldbaum:
Looks like some of the `arbitrary_grid` tests are failing.

Can you also add tests for this use case? I'm pretty sure the changes you've made will work for `covering_grid` and `aribtrary_grid` out of the box, but best to explicitly test that.

Thanks for sending in the patch and working on this!

AshKelly:
hm, so this doesn't work. We get to a point where if you pass in `('PartType0', 'mass')` or `('deposit', 'parttype0_mass')` that we can't tell if we should read from the file or SPH interpolate. 

The split fields decomposes saying we need,
`[('PartType0', 'Coordinates'), ('PartType0', 'Masses')]`
and then we don't know whether to read the `Coordinates` and `Masses` or try to interpolate as `field[0]` is the SPH particle type. Obviously we should ever interpolate coordinates, but in some instances people may want the `Masses` to be interpolated.

ngoldbaum:
When you get passed `('deposit', 'PartType0_mass')` you could short-circuit directly to the deposition field machinery, and get the underlying particle field data from the data object the covering grid uses for data access (`grid._data_source`).

AshKelly:
It works (I think) for all other particles except `PartType0` but I don't think I currently understand the flow of this section of code to figure this out after all. I'll have to come back to it later and try to understand more what is happening with the stars so I can replicate it for the gas

ngoldbaum:
OK. If you need a hand just let me know, I'm pretty sure I originally wrote the relevant code.

AshKelly:
yeah, so, what seems to happen is the `('deposit', 'PartType0_mass')` comes in and then this goes into `generate_fields` which then sends  `('PartType0', 'mass')` and  `('PartType0', 'Coordinates')` through. I can't seem to get a handle on how to distinguish against that and when somebody directly asks for `('PartType0', 'mass')`

AshKelly:
so this is my test script 
```python
import yt

fname = 'GadgetDiskGalaxy/snapshot_200.hdf5'
ds = yt.load(fname)

center = [31995.63476562, 31473.6640625, 28969.88671875]
arb_center = ds.arr(center, 'code_length')
left_edge = arb_center-ds.quan(100, 'kpc')
right_edge = arb_center+ds.quan(100, 'kpc')

arb = ds.arbitrary_grid(left_edge, right_edge, dims=[128, 128, 128],
                        field_parameters={'center': center})
ds.index

print("#### Star mass deposit")
arb[('deposit', 'PartType4_mass')].in_units('Msun')

print("#### Gas mass deposit")
arb[('deposit', 'PartType0_mass')].in_units('Msun')
```

Essentially the call from `fields/particle_fields.py` which I have added a print for makes a call which to me is indistinguishable from someone asking for `['gas', 'mass']`

matthewturk:
@AshKelly is this still something we should address?

AshKelly:
yeah, so I'm still stuck with this. In summary, my issue is that we want to intercept
`("gas", "XXX")` and as its an SPH field, but then it was suggested that `("PartType0", "XXX")` should also get SPH smoothed.
The issue is when `('deposit', 'PartType0_mass')` is passed in, it gets decomposed into `('PartType0', 'mass')` and `('PartType0', 'Coordinates')` and which then revisit the initial function. So then I try to SPH smooth the `('PartType0', 'mass')` which is wrong.
An easy-ish fix is to just SPH smooth `("gas", "XXX")` I guess or to make SPH smoothing a thing e.g,
`('smooth_deposit', 'PartType0_mass')`

matthewturk:
This might be a place where we need to break the similarity between mesh and particle codes, and make it clear that we are explicitly smoothing something.  If we agree to that -- and it might need to come up on yt-dev -- then I think the final option you propose would be good.

munkm:
Hey @AshKelly do we want to add the 4.0 label for this PR?

AshKelly:
Yeah, this a big thing we should fix. This is related to the whole gas field name API stuff.

AshKelly:
**TLDR**

We wired up the `covering_grid` and `arbitrary_grid` so that if they detect the particle type `PartType0` in the `get_data`
function then we short-circuit and go into some SPH deposition.

[see, `L686` of `yt/data_objects/construction_data_containers.py` in this branch]
```
        if(hasattr(self.ds, '_sph_ptype') and len(part)>0):
            is_sph_field = self.ds._sph_ptype in part[0]
```

**but,** sometimes people want to use the older machinery of depositing gas and pretending it is not an SPH field, e.g. if we care about total mass, `arb[('deposit', 'PartType0_mass')]`. Unfortunately, the method `get_data` recursively identifies that it needs the fields `('PartType0', 'Coordinates')` and `('PartType0', 'mass')` and calls itself recursively with those fields.
 
Then we go **"ah! this is an SPH field, let's SPH smooth!"** because we have forgotten the original function call. This means we end up trying to SPH smooth, but the user didn't ask for that.

So, I'm happy to work on this, but does anyone have any clever suggestions how I may fix this? If not, I'll just go along on my own and see if I can dig something out which works.

Pinging @munkm @neutrinoceros @matthewturk (for both suggestions and to make sure we keep this alive and fix before release)

stonnes:
Hi all,
I am making my way through our PRs, and noticed that things were moving along here but seem to have hit a snag on SPH smoothing of particles.  @AshKelly were you ever able to figure that out?   @munkm @neutrinoceros @matthewturk you were all pinged to see if you had ideas on the problem detailed above, this is just an extra nudge to make sure it doesn't fall off your radar!  Thanks!

AshKelly:
I think this is certainly something we should fix, however it's stagnant because I'm lost on how to fix it. I'm happy to zoom with someone and sort of discuss, brainstorm and hopefully fix, but I don't understand the field system well enough to fix this alone.

munkm:
@AshKelly I definitely think a zoom bugfix session is a good idea. Maybe we can poll people on slack and the mailing list to see who wants to join. I am happy to join but maybe another person (@jzuhone @Xarthisius or @matthewturk maybe?) would be interested too?

matthewturk:
@AshKelly I've thought through this a bit, and I have an opinion, but I'm not sure how strongly I hold it.

I think that if we're asking for `density`, specifically, it should put the density down -- and that means to smooth it.  I think that asking for a particle field name, for instance `Density`, would be fine to add as a non-smoothed field.  But I don't completely know the right way to implement this, unless we *also* did something like either check if it's an alias field (thus smoothed) or to add, say, a Particle Union that maps the `PartType0` to some other field name.  That would mean that if you accessed, say, `("PartType0", "density")` you'd get smoothed, but if you made a new alias to `"PartType0_unsmooth"` (or something) then when you accessed `("PartType0_unsmooth", "density")` it would just jam the values in.

I think that `density` is maybe the bad field example (that I chose!) because stuff like "Temperature" might be the ones you wouldn't want smoothed (plus "Density" is also calculable by doing mass/volume, and volume here will be a free paramter.)

Anyway, kinda what I'm thinking is that we probably don't want to support raw deposition of particles as opposed to smoothed, because doing so can be ill-defined, and requires more attention to the intensive/extensive properties of the field.  In the cases where you do want to do that, it might be better suited to a raw binning via something like `Profile3D`, although we don't have any docs that say that.

AshKelly:
@matthewturk I'm happy with both suggestions. Dropping the support, but also allowing some kind of override name `("PartType0_unsmooth", "density")`. 

I was actually to implement the latter but due to the recursive nature of the `get_data` function the initial call sort of gets washed out. If someone asks for `("PartType0_unsmooth", "density")` then `get_data` says "Ok, I need positions and densitys for `PartType0`". It then recursively calls itself with one of the fields as `("PartType0", "density")`. I can't distinguish this between this call from a user asking for SPH smoothing of `("PartType0", "density")`.

PR #2005, opened 2018-09-05
[Gadget frontend] Calculate header size from header spec to allow for non-standard headers

### LINKED ISSUE #1846: Extraheader with Gadget
### Bug report

**Bug summary**

When reading a gadget file with an extra header (e.g. https://obswww.unige.ch/~lhausamm/swift/IC/AgoraDisk/Gear/snapshot_0000), you do not take it into account here https://github.com/yt-project/yt/blob/master/yt/frontends/gadget/data_structures.py#L396

**Code for reproduction**

I am using the AGORA analysis script (https://bitbucket.org/mornkr/agora-analysis-script/src/8f12b112e19614d66274ece5a5bf21af23a76778/script-Paper4-disk-new.py?at=default&fileviewer=file-view-default) where they define the extra header here https://bitbucket.org/mornkr/agora-analysis-script/src/8f12b112e19614d66274ece5a5bf21af23a76778/script-Paper4-disk-new.py?at=default&fileviewer=file-view-default#script-Paper4-disk-new.py-150 and the code fails when adding a field here https://bitbucket.org/mornkr/agora-analysis-script/src/8f12b112e19614d66274ece5a5bf21af23a76778/script-Paper4-disk-new.py?at=default&fileviewer=file-view-default#script-Paper4-disk-new.py-501

**Actual outcome**

```
  File "./plot_solution.py", line 519, in <module>
    pf.add_field(("gas", "density_squared"), function=_density_squared, units="g**2/cm**6")
  File "/usr/local/lib/python2.7/dist-packages/yt/data_objects/static_output.py", line 1191, in add_field
    self.index
  File "/usr/local/lib/python2.7/dist-packages/yt/data_objects/static_output.py", line 504, in index
    self, dataset_type=self.dataset_type)
  File "/usr/local/lib/python2.7/dist-packages/yt/geometry/particle_geometry_handler.py", line 39, in __init__
    super(ParticleIndex, self).__init__(ds, dataset_type)
  File "/usr/local/lib/python2.7/dist-packages/yt/geometry/geometry_handler.py", line 50, in __init__
    self._setup_geometry()
  File "/usr/local/lib/python2.7/dist-packages/yt/geometry/particle_geometry_handler.py", line 50, in _setup_geometry
    self._initialize_particle_handler()
  File "/usr/local/lib/python2.7/dist-packages/yt/geometry/particle_geometry_handler.py", line 104, in _initialize_particle_handler
    self._initialize_indices()
  File "/usr/local/lib/python2.7/dist-packages/yt/geometry/particle_geometry_handler.py", line 134, in _initialize_indices
    self.io._initialize_index(data_file, self.regions)
  File "/usr/local/lib/python2.7/dist-packages/yt/frontends/gadget/io.py", line 351, in _initialize_index
    data_file, count, 0, regions, DLE, DRE)
  File "/usr/local/lib/python2.7/dist-packages/yt/frontends/gadget/io.py", line 334, in _get_morton_from_position
    pp = np.fromfile(f, dtype=self._endian + self._float_type,
TypeError: cannot concatenate 'str' and 'int' objects

```

**Expected outcome**

**Version Information**

  * Operating System: Linux Ubuntu 16.04
  * Python Version: 2.7.12
  * yt version: 3.4.1 (installed through pip)

qobilidop:
## PR Summary

Currently, Gadget frontend assumes a fixed header size of 256, which is the standard value. But since header spec is already configurable, it would be more self-consistent to calculate header size from header spec instead, thus allowing for non-standard headers.

Resolves #1846.

## PR Checklist

### DIFF
--- a/yt/frontends/gadget/data_structures.py
+++ b/yt/frontends/gadget/data_structures.py
@@ -58,7 +58,7 @@ def _compute_header_size(header_spec):
     return sum(field[1] * np.dtype(field[2]).itemsize for field in header_spec)
 
 
-def _get_gadget_format(filename, header_size=256):
+def _get_gadget_format(filename, header_size):
     # check and return gadget binary format with file endianness
     ff = open(filename, 'rb')
     (rhead,) = struct.unpack('<I', ff.read(4))
--- a/yt/frontends/gadget/data_structures.py
+++ b/yt/frontends/gadget/data_structures.py
@@ -54,7 +54,7 @@ def _byte_swap_32(x):
     return struct.unpack('>I', struct.pack('<I', x))[0]
 
 
-def _get_gadget_format(filename):
+def _get_gadget_format(filename, header_size=256):
     # check and return gadget binary format with file endianness
     ff = open(filename, 'rb')
     (rhead,) = struct.unpack('<I', ff.read(4))
@@ -63,9 +63,9 @@ def _get_gadget_format(filename):
         return 2, '>'
     elif (rhead == 8):
         return 2, '<'
-    elif (rhead == _byte_swap_32(256)):
+    elif (rhead == _byte_swap_32(header_size)):
         return 1, '>'
-    elif (rhead == 256):
+    elif (rhead == header_size):
         return 1, '<'
     else:
         raise RuntimeError("Incorrect Gadget format %s!" % str(rhead))
@@ -73,7 +73,7 @@ def _get_gadget_format(filename):
 
 class GadgetBinaryFile(ParticleFile):
     def __init__(self, ds, io, filename, file_id):
-        gformat = _get_gadget_format(filename)
+        gformat = _get_gadget_format(filename, ds._header_size)
         with open(filename, "rb") as f:
             if gformat[0] == 2:
                 f.seek(f.tell() + SNAP_FORMAT_2_OFFSET)
@@ -120,6 +120,8 @@ class GadgetDataset(SPHDataset):
             return
         self._header_spec = self._setup_binary_spec(
             header_spec, gadget_header_specs)
+        self._header_size = sum(field[1] * np.dtype(field[2]).itemsize
+                                for field in self._header_spec)
         self._field_spec = self._setup_binary_spec(
             field_spec, gadget_field_specs)
         self._ptype_spec = self._setup_binary_spec(
@@ -175,7 +177,7 @@ class GadgetDataset(SPHDataset):
     def _get_hvals(self):
         # The entries in this header are capitalized and named to match Table 4
         # in the GADGET-2 user guide.
-        gformat = _get_gadget_format(self.parameter_filename)
+        gformat = _get_gadget_format(self.parameter_filename, self._header_size)
         f = open(self.parameter_filename, 'rb')
         if gformat[0] == 2:
             f.seek(f.tell() + SNAP_FORMAT_2_OFFSET)
@@ -354,7 +356,7 @@ class GadgetDataset(SPHDataset):
         self.specific_energy_unit = self.quan(*specific_energy_unit)
 
     @staticmethod
-    def _validate_header(filename):
+    def _validate_header(filename, header_size=256):
         '''
         This method automatically detects whether the Gadget file is big/little endian
         and is not corrupt/invalid using the first 4 bytes in the file.  It returns a
@@ -380,9 +382,9 @@ class GadgetDataset(SPHDataset):
             f.close()
             return False, 1
         # Use value to check endianness
-        if rhead == 256:
+        if rhead == header_size:
             endianswap = '<'
-        elif rhead == _byte_swap_32(256):
+        elif rhead == _byte_swap_32(header_size):
             endianswap = '>'
         elif rhead in (8, _byte_swap_32(8)):
             # This is only true for snapshot format 2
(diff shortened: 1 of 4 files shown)


yt/frontends/gadget/data_structures.py
```diff
@@ -350,7 +356,7 @@ def _set_code_unit_attributes(self):
         self.specific_energy_unit = self.quan(*specific_energy_unit)
 
     @staticmethod
-    def _validate_header(filename):
+    def _validate_header(filename, header_size=256):
```
qobilidop: I have a question here. Currently I'm setting default header size to be 256, which would then be used in `GadgetDataset._is_valid`. This means datasets with non-standard headers would not be able to be loaded using `yt.load`. Instead, they have to be loaded using `GadgetDataset` directly.

I'm thinking about another method to solve this. In the endianness check below, we could do something like:
```python
def endian_check(f):
    for endianswap in '<>':
        f.seek(0)
        header_size = struct.unpack(endianswap + 'I', f.read(4))
        f.seek(4 + header_size)
        header_size_check = struct.unpack(endianswap + 'I', f.read(4))
        if header_size == header_size_check:
            return endianswap
    return False
```
This is only a consistency check, which shall work for any header size. But the drawback is that there could be false positives (although these should be rare cases).

Any suggestions?

ngoldbaum: One option would be to detect this case and emit a loud warning if it happens saying the user is doing something nonstandard and yt had to guess a bit

yt/frontends/gadget/data_structures.py
```diff
@@ -116,6 +124,11 @@ def __init__(self, filename, dataset_type="gadget_binary",
             return
         self._header_spec = self._setup_binary_spec(
             header_spec, gadget_header_specs)
+        self._header_size = _compute_header_size(self._header_spec)
+        if self._header_size != 256:
+            only_on_root(
+                mylog.warn, "Non-standard header size! (%s instead of 256)",
```
ngoldbaum: Can you add a sentence or two with a little bit of extra information about why this might have happened, e.g. they're using a non-standard version of the gadget output format, they should verify that their data are loaded correctly, and they might also want to customize their field configuration, and maybe even add a link to a relevant place in the docs.

For long multiline strings like that I usually format it like this:

```
only_on_root(mylog.warn,
             "this is a very long message that I'm spreading out over "
             "multiple lines. Note that I ended the previous line with "
             "a space. Also note how there aren't any characters "
             "between the open and end quotes besides newlines.")
```

qobilidop: Thanks for the suggestion. I'll make the change.

qobilidop:
Unfortunately, I don't have a real dataset to test upon. The dataset I talked about in #1846 turns out to suffer from a different issue (the snapshot file is corrupted). @loikki might still have the snapshot to trigger the issue. But from what I see on his website, he has switched to use HDF5.

I'm currently experimenting generating fake gadget snapshots programmatically. It seems realtively easy to do. I'll come back later about how it goes.

yt/frontends/gadget/data_structures.py
qobilidop: Before I commit the change, does this message look good?
```python
only_on_root(
    mylog.warn,
    "Non-standard header size is detected! "
    "Gadget-2 standard header is 256 bytes, but yours is %s. "
    "Make sure a non-standard header is actually expected. "
    "Otherwise something is wrong, "
    "and you might want to check how the dataset is loaded. "
    "Futher information about header specification can be found in "
    "https://yt-project.org/docs/dev/examining/loading_data.html#header-specification.",
    self._header_size
)
```

ngoldbaum: It looks like you didn't actually use the `%` operator in the string to get the formatting to process, but other than that LGTM.

ngoldbaum:
If a test dataset isn't readily available I won't object to merging this.

yt/frontends/gadget/data_structures.py
qobilidop: Actually it is used. The format is copied from other occurrences of `only_on_root(mylog.xxx, ...)`. But know I wonder how. It must be some magic in how `mylog` works.

qobilidop:
@ngoldbaum I'm testing it locally using a mock dataset, and there are some issues. Don't merge it yet.

ngoldbaum:
wasn’t going to :)

btw if your fake dataset generator isn’t too insanely complicated please feel free to include it in the test suite. it might be useful to use that to get rid of our dependence on real gadget data in the test suite too :)

yt/frontends/gadget/data_structures.py
```diff
@@ -50,26 +50,34 @@ def _fix_unit_ordering(unit):
     return unit
 
 
-def _get_gadget_format(filename):
+def _byte_swap_32(x):
```
MatthewTurk: This should always result in a big endian, right?  Is it really swapping, or just converting to big-endian?

qobilidop: Yes, in the usage below, it's always converting from little-endian to big-endian. But since the conversion either way is the same operation, I just use this neutral name for the process.

yt/frontends/gadget/io.py
```diff
@@ -242,7 +242,7 @@ def __init__(self, ds, *args, **kwargs):
         self._fields = ds._field_spec
         self._ptypes = ds._ptype_spec
         self.data_files = set([])
-        gformat = _get_gadget_format(ds.parameter_filename)
+        gformat = _get_gadget_format(ds.parameter_filename, ds._header_size)
```
qobilidop: @MatthewTurk This is the bug I referred to in the commit message. It's an error made by myself that I forgot to pass the header size through.

qobilidop:
I'm still having trouble figuring out what went wrong in the CI tests. I'll take another look during the sprint.

yt/frontends/gadget/data_structures.py
MatthewTurk: Oh, sheesh, you're right.  I got confused in how I was thinking about it.  You're right.  Thank you.

PR #2008, opened 2018-09-08
Adding standard deviation to profile.save_as_dataset.

brittonsmith:
## PR Summary

Previously, the standard deviation data for a profile was not saved with the `profile.save_as_dataset` function, making it in accessible on reload. This adds that data and allows it to be properly restored.
As an example, the last line was not possible until this PR.
```
ds = yt.load("DD0046/DD0046")
sp = ds.sphere(ds.domain_center, 0.2)
pr = yt.create_profile(sp, 'density', ['temperature'], weight_field='cell_mass')
# this is the standard deviation
print (pr.standard_deviation)
# fn = pr.save_as_dataset()
pr_ds = yt.load(fn)
# previously not possible
print (pr_ds.profile.standard_deviation)
```

Additionally, a profile dataset's `profile` object was being recreated each time it was accessed. This fixes that as well.

## PR Checklist


yt/frontends/ytdata/data_structures.py
```diff
@@ -732,19 +732,24 @@ def _is_valid(self, *args, **kwargs):
...
     @property
     def profile(self):
-        if self.dimensionality == 1:
-            return Profile1DFromDataset(self)
-        if self.dimensionality == 2:
-            return Profile2DFromDataset(self)
-        if self.dimensionality == 3:
-            return Profile3DFromDataset(self)
-        return None
+        if self._profile is None:
```
MatthewTurk: Any chance you could switch these, so that if `self._profile is not None` it returns early, and then this section of the conditional lives one level of indentation higher?

brittonsmith: Sure, I can do that. I did it this way as having the single return underneath the creation step seemed like standard procedure for properties.

brittonsmith:
Looks better this way.

PR #2009, opened 2018-09-08
Use Cython C++ idioms for EWAH

MatthewTurk:
This modifies some of the EWAH code to use more cython-like idioms and to reduce some of the lines of code.  It should produce *no* changes, and if it does, that is a bug.  I have not identified changes in my testing.

There are a few other places that we could use Cython idioms for iteration and the like, but I have not yet applied these changes there, as I'm not 100% certain it would improve overall readability (and I want to avoid copy-on-assignment).

### DIFF
--- a/yt/utilities/lib/ewah_bool_wrap.pyx
+++ b/yt/utilities/lib/ewah_bool_wrap.pyx
@@ -183,29 +183,25 @@ cdef class FileBitmasks:
         coll_coll = <map[np.uint64_t, ewah_bool_array]*> coll.ewah_coll
         for ifile in range(self.nfiles):
             map_bitmask = (<map[np.uint64_t, ewah_bool_array]**> self.ewah_coll)[ifile]
-            it_mi1 = map_bitmask[0].begin()
-            while it_mi1 != map_bitmask[0].end():
-                mi1 = dereference(it_mi1).first
-                iarr = dereference(it_mi1).second
+            for it_mi1 in map_bitmask[0]:
+                mi1 = it_mi1.first
+                iarr = it_mi1.second
                 map_keys[mi1].logicaland(iarr, arr_two)
                 map_keys[mi1].logicalor(iarr, arr_swap)
                 map_keys[mi1].swap(arr_swap)
                 map_refn[mi1].logicalor(arr_two, arr_swap)
                 map_refn[mi1].swap(arr_swap)
-                preincrement(it_mi1)
         coll_coll[0] = map_refn
         # Count
         cdef int nc, nm
         nc = 0
         nm = 0
-        it_mi1 = map_refn.begin()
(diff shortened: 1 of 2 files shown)


yt/utilities/lib/ewah_bool_wrap.pyx
```diff
@@ -1747,32 +1740,20 @@ cdef class SparseUnorderedBitmaskSet:
     cdef void _fill(self, np.uint8_t[:] mask):
         cdef np.uint64_t ind
         cdef cset[np.uint64_t] *entries = <cset[np.uint64_t]*> self.entries
-        cdef cset[np.uint64_t].iterator it
-        it = entries[0].begin()
-        while it != entries[0].end():
-            ind = dereference(it)
-            mask[ind] = 1
-            preincrement(it)
+        for it in entries[0]:
```
langmm: Here, and in a few other places, the iterator (it) is not declared. It also looks like there could be some clean up of unused variables (e.g. ind) that were only used when dereferencing pointers.

MatthewTurk: This is true, it's not declared!  I checked into it, and it looks like in this particular case, it infers the type.  The generated code here looks like:

```
  __pyx_t_2 = &(__pyx_v_entries[0]);
  __pyx_t_1 = __pyx_t_2->begin();
  for (;;) {
    if (!(__pyx_t_1 != __pyx_t_2->end())) break;
    __pyx_t_3 = *__pyx_t_1;
    ++__pyx_t_1;
    __pyx_v_it = __pyx_t_3;
```

which is, I think, correct.  I believe it *has* to infer the type, because if it can't, it would not be able to cast the iter to a python object and would bail.

I'm working on the cleanup of unused vars!

MatthewTurk:
In the current version, I'm run using

```
cython --cplus -Wextra -Werror --fast-fail yt/utilities/lib/ewah_bool_wrap.pyx
```

and got it to compile cleanly with the exception of the function `_intersects`, where it looks like some variables are relied on existing after the end of a closure in a loop.  This is probably safe, but I did not want to modify them without fully understanding what I was doing, and after trying, I couldn't convince myself I did understand it.

I think this is ready for another look.

yt/utilities/lib/ewah_bool_wrap.pyx
langmm: Looks good!

ngoldbaum:
I'm currently working on unbreaking the tests and will get to this once that's fixed. Sorry for the trouble.

MatthewTurk:
I *think*, but am not sure, that the tests issues here are related to the ones just fixed on `master`, but I am not certain because of the timeout -- it's possible that the timeout resulted from some error or lack-of-optimization I put in, so if anyone has guidance on that it'd be helpful.

Issue #2012, opened 2018-09-11
radial velocity changes with normal vector


brittonsmith:
### Bug report

Velocity magnitudes in spherical and cylindrical coordinates seem to be dependent on the normal vector, at least if the normal vector is not fully aligned with a cartesian axis. In addition, the spherical radial velocity itself seems dependent on direction of the normal vector under the same conditions.

This can be reproduced by adding the following line to `yt/fields/tests/test_vector_fields.py`:
```
ad.set_field_parameter('normal', np.ones(3) / np.sqrt(3))
```

This will trigger a failure in the test comparing velocity magnitude in all three coordinate systems. As I mentioned above, setting the normal to be entirely in the x, y, or z axis does not result in a test failure. To be clear, this is not roundoff error; the answers are not at all close.

brittonsmith:
Ah, sure, right! This is loading a single dataset and calculating, say, the radial velocity for a sphere, then changing the normal and calculating it again. Here's an example below:
```
import numpy as np
import yt

ds = yt.load("DD0046/DD0046")

sp = ds.sphere(ds.domain_center, 0.2)
print (sp['velocity_spherical_radius'])

sp.set_field_parameter('normal', np.ones(3) / np.sqrt(3))
sp.clear_data()
print (sp['velocity_spherical_radius'])
```
This gives:
```
[-14111561.13148242 -15307975.99893196 -12899747.19029205 ...
   2581133.91582618   4770170.78048369   6544593.17199936] cm/s
[ -8491218.10477266 -12443362.16458249  -6273167.96971074 ...
  -1141804.32237623   1094212.46564677   2928625.15379348] cm/s
```

MatthewTurk:
I believe @ngoldbaum tracked this down to `get_ortho_basis`.  Is it possible to use `ortho_find` instead?

ngoldbaum:
I fixed `get_ortho_basis`:

```
diff --git a/yt/utilities/math_utils.py b/yt/utilities/math_utils.py
index 19ae65c4c..0067297f7 100644
--- a/yt/utilities/math_utils.py
+++ b/yt/utilities/math_utils.py
@@ -1189,10 +1189,15 @@ def rotation_matrix_to_quaternion(rot_matrix):
     return np.array([w, x, y, z])

 def get_ortho_basis(normal):
-    xprime = np.cross([0.0,1.0,0.0],normal)
-    if np.sum(xprime) == 0: xprime = np.array([0.0, 0.0, 1.0])
[...]

PR #2013, opened 2018-09-12
Fix bug in get_ortho_base

brittonsmith:
## PR Summary

This fixes Issue #2012, in which the radial velocity was dependent on the normal vector of the data container. The heart of the issue was a bug in computing the basis vectors in `yt.utilities.math_utils.get_ortho_base`. After much effort on the part of @ngoldbaum and myself, @MatthewTurk made me aware of `yt.utilities.math_utils.ortho_find` (written by @chummels), which solves exactly this problem.

I've also made `test_vector_component_conversions ` a little more robust by testing some random velocity and normal vectors. As @ngoldbaum found a solution that fixed those tests but still didn't work for real simulation data, I will add a test that uses real data shortly, but first I must go home and put my child to bed.

A huge thanks to everyone mentioned in this PR for the effort they put in, both past and present, deliberate and not.

## PR Checklist

### DIFF
--- a/yt/utilities/math_utils.py
+++ b/yt/utilities/math_utils.py
@@ -1189,10 +1189,7 @@ def rotation_matrix_to_quaternion(rot_matrix):
     return np.array([w, x, y, z])
 
 def get_ortho_basis(normal):
-    xprime = np.cross([0.0,1.0,0.0],normal)
-    if np.sum(xprime) == 0: xprime = np.array([0.0, 0.0, 1.0])
-    yprime = np.cross(normal,xprime)
-    zprime = normal
+    zprime, xprime, yprime = ortho_find(normal)
     return (xprime, yprime, zprime)
 
 def get_sph_r(coords):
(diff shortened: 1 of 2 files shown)


yt/utilities/math_utils.py
```diff
@@ -1189,10 +1189,7 @@ def rotation_matrix_to_quaternion(rot_matrix):
     return np.array([w, x, y, z])
 
 def get_ortho_basis(normal):
-    xprime = np.cross([0.0,1.0,0.0],normal)
-    if np.sum(xprime) == 0: xprime = np.array([0.0, 0.0, 1.0])
-    yprime = np.cross(normal,xprime)
-    zprime = normal
+    zprime, xprime, yprime = ortho_find(normal)
     return (xprime, yprime, zprime)
```
ngoldbaum: Let's just delete this function and replace it with `ortho_find`.

ngoldbaum:
It looks like the travis and appveyor test failures are real: https://gist.github.com/ngoldbaum/ee2cd41bd18496832cacacfeee56ae40

ngoldbaum:
@MatthewTurk would you mind double-checking britton? I think he's right but since the old test answers were wrong I want to be extra special sure.

PR #2020, opened 2018-09-13
avoid using cached field data in chunked reads. fixes #1874

### LINKED ISSUE #1874: [BUG] (RAMSES ?) OffAxisProjection
### Bug report

**Bug summary**

When doing OffAxisProjections with a weight field for multiple fields at the same time, the resulting images are maps of the first field only, but with the wrong unit (see images).

**Code for reproduction**
```python
import yt
import numpy

ds = yt.load('ramses_rt_00088/output_00088/info_00088.txt')
p = yt.OffAxisProjectionPlot(ds, [1, 0, 0], ['density', 'temperature'], weight_field='cell_volume')
p.save('/tmp/')
```

**Actual outcome**

info_00088_offaxisprojection_density_cell_volume
info_00088_offaxisprojection_temperature_cell_volume

**Expected outcome**
Here is the expected outcome for the regular projection plot on the same axis.

info_00088_projection_x_temperature_cell_volume
info_00088_projection_x_density_cell_volume

**Version Information**

  * Operating System: Linux
  * Python Version: 3.6.5
  * yt version: 3.5_dev, from git (master branch, ef4be5d5497b1771af453fe47a5795f510cb860c).

EDIT:

This script runs a lot more quickly:

```
import yt

ds = yt.testing.fake_octree_ds()
p = yt.OffAxisProjectionPlot(ds, [1, 0, 0], ['density', 'velocity_x'],
                             weight_field='cell_volume')
p.save()
```

ngoldbaum:
Octree data has a little bit more indirection in how it stores data than the grid frontends. To catch this extra bit of cached data I needed to add a more logic to the `_chunked_read` context manager so that old cached I/O results don't pollute future field accesses.

### DIFF
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -1514,9 +1514,10 @@ class YTSelectionContainer(YTDataContainer, ParallelAnalysisInterface):
         # There are several items that need to be swapped out
         # field_data, size, shape
         obj_field_data = []
-        for obj in chunk.objs:
-            obj_field_data.append(obj.field_data)
-            obj.field_data = YTFieldData()
+        if hasattr(chunk, 'objs'):
+            for obj in chunk.objs:
+                obj_field_data.append(obj.field_data)
+                obj.field_data = YTFieldData()
         old_field_data, self.field_data = self.field_data, YTFieldData()
         old_chunk, self._current_chunk = self._current_chunk, chunk
         old_locked, self._locked = self._locked, False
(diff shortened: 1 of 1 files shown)


yt/data_objects/data_containers.py
```diff
@@ -1513,13 +1513,19 @@ def _ds_hold(self, new_ds):
     def _chunked_read(self, chunk):
         # There are several items that need to be swapped out
         # field_data, size, shape
+        obj_field_data = []
```
MatthewTurk: I'm not 100% convinced this is right.  Can you remind me which object composes `chunk.objs` in the issue we're dealing with?

This also may result in caching all field data from all grid objects in a grid frontend.

ngoldbaum: In the issue they're instances of `OctreeSubset`.

ngoldbaum:
@MatthewTurk is there any chance you can suggest a test I might do to assuage your caching fears :)

Xarthisius:
@yt-fido test this please (testing if getting rid of ceph speeds things up)

ngoldbaum:
@yt-fido test this please

Issue #1609, opened 2017-11-03
save_as_dataset should complain when trying to save data with incompatible shapes


ngoldbaum:
### Bug report

**Bug summary**

save_as_dataset is happy to save unstructured field data with different shapes. This leads to errors when trying to actually use the data later.

**Code for reproduction**

```python
import yt
import numpy as np

density = np.random.random(128)
temperature = np.random.random(128)
cell_mass = np.random.random((128, 128))

my_data = {
    "density": density,
    "temperature": temperature,
    "cell_mass": cell_mass}
fake_ds_med = {"current_time": yt.YTQuantity(10, "Myr")}
yt.save_as_dataset(fake_ds_med, "mydata.h5", my_data)

ds = yt.load('mydata.h5')

yt.PhasePlot(ds.data, 'temperature', 'density', 'cell_mass')
```

**Actual outcome**

```
Traceback (most recent call last):
  File "testt.py", line 17, in <module>
    yt.PhasePlot(ds.data, 'temperature', 'density', 'cell_mass')
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/visualization/profile_plotter.py", line 761, in __init__
    fractional=fractional)
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/data_objects/profiles.py", line 1095, in create_profile
    obj.add_fields([field for field in fields])
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/data_objects/profiles.py", line 127, in add_fields
    self._bin_chunk(chunk, fields, temp_storage)
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/data_objects/profiles.py", line 567, in _bin_chunk
    rv = self._get_data(chunk, fields)
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/data_objects/profiles.py", line 255, in _get_data
    arr[:,i] = chunk[field][filter].in_units(units)
ValueError: could not broadcast input array from shape (128,128) into shape (128)
```

**Expected outcome**

I'd expect yt to raise an error during the save_as_dataset call.

**Version Information**

  * Operating System:
  * Python Version: 3.6
  * yt version: 3.5dev
  * Other Libraries (if applicable):

ngoldbaum:
This was motivated by this mailing list thread:

http://lists.spacepope.org/pipermail/yt-users-spacepope.org/2017-November/008948.html

@brittonsmith do you agree that we should be raising an error in these cases?

[...]

PR #2025, opened 2018-09-13
Raise exception for profile fields of unequal shape.

brittonsmith:
This fixes Issue #1609. We check that bin fields and profile fields are all of the correct shape. I also changed all instances of `filter` to `pfilter` since `filter` is a reserved word. Tests added as well.

### DIFF
--- a/yt/data_objects/tests/test_profiles.py
+++ b/yt/data_objects/tests/test_profiles.py
@@ -1,6 +1,12 @@
-import yt
 import numpy as np
+import os
+import shutil
+import tempfile
+import unittest
+import yt
 
+from yt.utilities.exceptions import \
+    YTProfileDataShape
 from yt.data_objects.particle_filters import add_particle_filter
 from yt.data_objects.profiles import Profile1D, Profile2D, Profile3D,\
     create_profile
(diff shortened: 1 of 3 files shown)


yt/data_objects/tests/test_profiles.py
```diff
@@ -319,3 +325,62 @@ def test_profile_override_limits():
...
+        self.tmpdir = tempfile.mkdtemp()
+        self.curdir = os.getcwd()
+        os.chdir(self.tmpdir)
+
+    def tearDown(self):
+        os.chdir(self.curdir)
+        # clean up
+        shutil.rmtree(self.tmpdir)
+
+    def test_unequal_data_shape_profile(self):
```
ngoldbaum: These tests need to be decorated with `@requires_module('h5py')` I *think* just decorating the test methods is sufficient but you can double-check locally by uninstalling h5py and running just these tests.

yt/data_objects/tests/test_profiles.py
```diff
@@ -319,3 +325,62 @@ def test_profile_override_limits():
...
+            "density": density,
+            "temperature": temperature,
+            "cell_mass": cell_mass}
+        fake_ds_med = {"current_time": yt.YTQuantity(10, "Myr")}
+        yt.save_as_dataset(fake_ds_med, "mydata.h5", my_data)
+
+        ds = yt.load('mydata.h5')
+
+        # We want this to fail with a YTProfileDataShape exception.
+        try:
```
ngoldbaum: you can replace everything below the `try` with an `assert_raises` like this:

```
assert_raises(YTProfileDataShape, yt.PhasePlot, ds.data, 'temperature', 'density', 'cell_mass')
```

yt/data_objects/tests/test_profiles.py
```diff
@@ -319,3 +325,62 @@ def test_profile_override_limits():
...
```
ngoldbaum: same here with `assert_raises`

yt/utilities/exceptions.py
```diff
@@ -679,6 +679,17 @@ def __str__(self):
...
 
+class YTProfileDataShape(YTException):
+    def __init__(self, field1, shape1, field2, shape2):
+        self.field1 = field1
+        self.shape1 = shape1
+        self.field2 = field2
+        self.shape2 = shape2
+
+    def __str__(self):
+        return "Profile fields must have same shape: %s is %s and %s is %s." % \
```
ngoldbaum: Instead of "is" I'd say "has shape"

PR #2027, opened 2018-09-14
fix attribute error in clump finding tests

ngoldbaum:
This is code from two years ago and I have no idea why we're only noticing this now, but, currently this test fails with the following error:

```
ERROR: yt.data_objects.level_sets.tests.test_clump_finding.test_clump_tree_save
  File "/usr/local/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
diff --git a/yt/data_objects/level_sets/tests/test_clump_finding.py b/yt/data_objects/level_sets/tests/test_clump_finding.py
index e674ead02..640fe3e72 100644
    self.test(*self.arg)
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/data_objects/level_sets/tests/test_clump_finding.py", line 134, in test_clump_tree_save
    ic1 = np.argsort(mc1).d.astype(int)
AttributeError: 'numpy.ndarray' object has no attribute 'd'
```

The fix is just to not use `.d` at all.


ngoldbaum:
Ah good point! Merging so the tests don't break for #2028

PR #2029, opened 2018-09-14
Fix a number of docs errors and warnings.

brittonsmith:
This fixes a number of errors and warnings that show up in the docs build log.

In addition to this, some errors were caused by the need for `pyx` and `bottle` to be installed. We can either make these on-demand imports or just install them on the build machine.


ngoldbaum:
@Xarthisius can we get those dependencies installed on the doc builder?

Xarthisius:
`Successfully installed bottle-0.12.13 pyx-0.14.1`

PR #2034, opened 2018-09-17
[bugfix] catch particles on domain edges for Gadget_FOF data

brittonsmith:
## PR Summary

In this frontend, and likely others, particles on domain boundaries were being lost because the code to bump them in from the edges only existed in `_initialize_index` and none of the other functions in the io handler that read particles. This fixes that by implementing a class method on the data file for reading particle positions and doing the correction there rather than have three copies of the same code.

In addition, I implemented this with a cache since particle positions need to be read in for every field read. For the rather large dataset I was testing, this reduced the load time for a field from ~3.5s to ~0.5s (after the positions had been read in once). Of course, this will increase the steady-state memory use, but I think it's probably worth it (although I'm open to other opinions).

Like many bugs of this nature, it took a pretty large dataset (73 GB) to find it, so including it for a test is probably difficult.

I suspect that this affects all similar frontends (all the halo catalogs at least). I'm open to porting this fix around if that's desirable. I could also implement some sort of `correct_positions` function at a higher class level.

## PR Checklist

### DIFF
--- a/yt/frontends/halo_catalog/data_structures.py
+++ b/yt/frontends/halo_catalog/data_structures.py
@@ -64,14 +64,13 @@ class HaloCatalogFile(ParticleFile):
                 self._coords[ptype] = None
             return None
 
-        # These are 32 bit numbers, so we give a little lee-way.
-        # Otherwise, for big sets of particles, we often will bump into the
-        # domain edges.  This helps alleviate that.
+        # Correct for periodicity.
         dle = self.ds.domain_left_edge.to('code_length').v
-        dre = self.ds.domain_right_edge.to('code_length').v
-        dx = 2. * np.finfo(np.float32).eps
+        dw = self.ds.domain_width.to('code_length').v
         pos = self._read_particle_positions(ptype, f=f)
-        np.clip(pos, dle + dx, dre - dx, pos)
+        np.subtract(pos, dle, out=pos)
+        np.mod(pos, dw, out=pos)
+        np.add(pos, dle, out=pos)
 
         if cache_pos:
             self._coords[ptype] = pos
--- a/yt/frontends/halo_catalog/data_structures.py
+++ b/yt/frontends/halo_catalog/data_structures.py
@@ -44,14 +44,47 @@ class HaloCatalogParticleIndex(ParticleIndex):
               [cls(self.dataset, self.io,
                    self.dataset.parameter_filename, 0)]
 
-class HaloCatalogHDF5File(ParticleFile):
+class HaloCatalogFile(ParticleFile):
+    def __init__(self, ds, io, filename, file_id):
+        self._coords = {}
+        super(HaloCatalogFile, self).__init__(
+            ds, io, filename, file_id)
+
+    def _read_particle_positions(self, ptype, f=None):
+        raise NotImplementedError
+
+    def _get_particle_positions(self, ptype, f=None):
+        if ptype in self._coords:
+            return self._coords[ptype]
+
+        pcount = self.total_particles[ptype]
+        cache_pos = getattr(self.ds, "cache_positions", False)
+        if pcount == 0:
+            if cache_pos:
+                self._coords[ptype] = None
+            return None
+
+        # These are 32 bit numbers, so we give a little lee-way.
+        # Otherwise, for big sets of particles, we often will bump into the
+        # domain edges.  This helps alleviate that.
+        dle = self.ds.domain_left_edge.to('code_length').v
+        dre = self.ds.domain_right_edge.to('code_length').v
+        dx = 2. * np.finfo(np.float32).eps
+        pos = self._read_particle_positions(ptype, f=f)
+        np.clip(pos, dle + dx, dre - dx, pos)
+
+        if cache_pos:
+            self._coords[ptype] = pos
+        return pos
+
+class HaloCatalogHDF5File(HaloCatalogFile):
     def __init__(self, ds, io, filename, file_id):
         with h5py.File(filename, "r") as f:
             self.header = dict((field, parse_h5_attr(f, field)) \
                                for field in f.attrs.keys())
+        super(HaloCatalogHDF5File, self).__init__(
+            ds, io, filename, file_id)
 
-        super(HaloCatalogHDF5File, self).__init__(ds, io, filename, file_id)
-    
 class HaloCatalogDataset(SavedDataset):
     _index_class = ParticleIndex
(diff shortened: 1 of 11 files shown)


ngoldbaum:
For the really big dataset you found this on what’s the memory cost of the cache?

brittonsmith:
This dataset has ~14 million halos, which comes out to ~300 MB for the positions. That's not even a very big dataset by particle standards, so I could see not wanting to cache for things much large than this. That could either be implemented on a case by case basis or configurable somehow, or I don't know, I could ditch the caching.

For what it's worth, the bulk of the 73 GB of this dataset is storing member particles and fields for each halo.

ngoldbaum:
Looks like the test failure is real.

I thought about this a bit last night and I think it's probably OK to keep the cache around if the speedup is really so significant. @MatthewTurk might want to weigh in as well. It might be nice to make that configurable somehow so there's a knob to turn it off (e.g. in a keyword argument passed to `load` or in a yt config option). We'd probably want to explicitly test both cases though.

I don't think you should be responsible for fixing this elsewhere. However, if you could identify where it's a problem elsewhere and open an issue describing how you'd fix it that would help a ton.

MatthewTurk:
I'm not super into this method of caching, but it seems to be a workable solution.

One thing that I'd be a bit curious about is just how the particles on the domain boundaries get lost -- where do they stop getting selected?  I'm wondering if we should modify the cython routines to ensure that the right edge is considered inclusively for particles or not.

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -44,14 +44,68 @@ def _setup_filenames(self):
...
+            return None
+
+        # These are 32 bit numbers, so we give a little lee-way.
+        # Otherwise, for big sets of particles, we often will bump into the
+        # domain edges.  This helps alleviate that.
+        dle = self.ds.domain_left_edge.to('code_length').v
+        dre = self.ds.domain_right_edge.to('code_length').v
+        dx = 2. * np.finfo(np.float32).eps
+        pos = self._read_particle_positions(ptype, f=f)
+        np.clip(pos, dle + dx, dre - dx, pos)
```
MatthewTurk: I admit I'm not totally following this, but are we respecting the "only read in these bounds" rules that we allow in the constructor to the dataset object?

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -44,14 +44,68 @@ def _setup_filenames(self):
...
+
+        if cache_pos:
+            self._coords[ptype] = pos
```
MatthewTurk: One thing to keep an eye on here is that we're caching it, but it's entirely possible that in-place operations will happen on the arrays later on.  It might be a good idea when accessing things from the cache to ensure that they're always in the right units.

qobilidop:
Just one suggestion, if a real dataset is too large, could there be an artificial dataset made to test these edge cases? Otherwise, it looks good to me.

yt/frontends/halo_catalog/data_structures.py
brittonsmith: I don't understand this comment. What are these rules you're referring to?

This clipping strategy was code that appeared in `_initialize_index` for each of the frontends here. It seemed like a good idea to apply the same logic to the actual positions, but maybe it doesn't matter.

yt/frontends/halo_catalog/data_structures.py
brittonsmith: I didn't think of that. There may `YTArrays` made with these changing all the units around and it would be difficult to know. Maybe this just won't work.

yt/frontends/halo_catalog/data_structures.py
ngoldbaum: I think he's referring to supplying a `bounding_box` keyword argument to `yt.load()`:

```
In [1]: ds.domain_left_edge
Out[1]: YTArray([ -547.51629822,  -557.30091064, -3675.19602173]) code_length

In [2]: ds.domain_right_edge
Out[2]: YTArray([ 579.153078  ,  472.81018799, 1591.69712036]) code_length

In [3]: ad = ds.all_data()

In [4]: ad['gas', 'density'].shape
Out[4]: (169240,)

In [6]: ds = yt.load('TipsyGalaxy/galaxy.00300', bounding_box=[ds.domain_left_edge/2, ds.domain_right_edge/2])

In [7]: ds.domain_left_edge
Out[7]: YTArray([ -273.75814911,  -278.65045532, -1837.59801086]) code_length

In [8]: ds.domain_right_edge
Out[8]: YTArray([289.576539  , 236.40509399, 795.84856018]) code_length

In [9]: ad = ds.all_data()

In [10]: ad['gas', 'density'].shape
Out[10]: (167728,)
```

brittonsmith:
If memory serves, they were getting lost in the `selector.select_points(x, y, z, 0.0)` call because they were exactly on the right boundary. I think it's more proper continue with a periodic boundary that is inclusive on the left and exclusive on the right since it avoids the possibility of having two numerical values for a single physical location. I can see that becoming a problem somewhere. In addition, nudging particles inward as a general practice could result in a significant displacement from a particle's actual position. Rather, if the boundary is periodic then we should be shifting the particles to their true position within the domain. I am going to change this PR to do that. No one seems to like the cache, so I'm going to get rid of that, too.

brittonsmith:
@qobilidop, I've added a couple tests using some fake halo catalogs, including a test where I put particles on the domain boundaries. I've confirmed that this fails prior to this PR.

yt/frontends/halo_catalog/tests/test_outputs.py
```diff
@@ -0,0 +1,99 @@
...
+          'current_time': YTQuantity(1, 'yr'),
+          'domain_left_edge': YTArray(np.zeros(3), 'cm'),
+          'domain_right_edge': YTArray(np.ones(3), 'cm')}
+    save_as_dataset(
+        ds, filename, data,
+        field_types=ftypes, extra_attrs=extra_attrs)
+    return filename
+
+class HaloCatalogTest(TempDirTest):
+    def test_halo_catalog(self):
```
ngoldbaum: These tests need to be decorated with `@requires_module('h5py')`

ngoldbaum:
@yt-fido test this please

yt/frontends/halo_catalog/tests/test_outputs.py
brittonsmith: Right, I always forget this. Thanks!

Issue #2032, opened 2018-09-17
User defined units not found with `magnetic_pressure`


maxbeegee:
### Bug report

**Bug summary**

yt does not seem to be able to use user defined units with the field `magnetic_pressure`

**Code for reproduction**

```python
ds = yt.load(some_file)
ds.define_unit("ptest", (1.0, "dyn/cm**2"))
ad = ds.all_data()
ad['pressure'].min().to('ptest') # note that this works
ad['magnetic_pressure'].min().to('ptest') # this fails
```

**Actual outcome**

```
Traceback (most recent call last):
  File "yt_but_magnetic_pressure.py", line 6, in <module>
    print(ad['magnetic_pressure'].min().to('ptest'))
  File "/home1/05311/maxbg/libs/python/my_yt/yt/yt/units/yt_array.py", line 639, in to
    return self.in_units(units, equivalence=equivalence, **kwargs)
  File "/home1/05311/maxbg/libs/python/my_yt/yt/yt/units/yt_array.py", line 621, in in_units
    new_units = _unit_repr_check_same(self.units, units)
  File "/home1/05311/maxbg/libs/python/my_yt/yt/yt/units/yt_array.py", line 269, in _unit_repr_check_same
    other_units = Unit(other_units, registry=my_units.registry)
  File "/home1/05311/maxbg/libs/python/my_yt/yt/yt/units/unit_object.py", line 268, in __new__
    unit_data = _get_unit_data_from_expr(unit_expr, registry.lut)
  File "/home1/05311/maxbg/libs/python/my_yt/yt/yt/units/unit_object.py", line 577, in _get_unit_data_from_expr
    return _lookup_unit_symbol(str(unit_expr), unit_symbol_lut)
  File "/home1/05311/maxbg/libs/python/my_yt/yt/yt/units/unit_object.py", line 668, in _lookup_unit_symbol
    "symbols." % symbol_str)
yt.units.unit_registry.UnitParseError: Could not find unit symbol 'ptest' in the provided symbols.
```

**Expected outcome**

The snippet should run and the two lines give the same result.

**Version Information**

  * Operating System: unix
  * Python Version: Python 2.7.14 :: Anaconda custom (64-bit)
  * yt version: 3.5.dev0
  * Other Libraries (if applicable): 

yt installed via from github.

jzuhone:
I'll have a look at this one. @maxbeegee, do you have a particular test dataset this happens on?

maxbeegee:
[...]

PR #2035, opened 2018-09-17
[bugfix] A Dataset's UnitSystem should have its UnitRegistry

jzuhone:
During the creation of a `Dataset` instance it is assigned a `UnitSystem`, which the end-user can specify. Upon importing `yt` a number of `UnitSystem` instances are automatically created, such as "cgs", "mks", "galactic", etc. One of these is specified to `yt.load` (or the default) and it is assigned as the `Dataset`'s unit system. 

If done in this way, the `UnitSystem` does not have the correct `UnitRegistry` object. This PR fixes that problem by enforcing that a new `UnitSystem` instance is created which is associated with the `Dataset`'s `UnitRegistry`.  

I'm not thrilled about this solution, but the other solutions I came up with were messier than this, and I thought doing a `deepcopy` or something similar of the original `UnitSystem` instance and simply fixing the registry was a bad idea.

Closes Issue #2032.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -67,8 +67,9 @@ from yt.utilities.minimal_representation import \
 from yt.units.yt_array import \
     YTArray, \
     YTQuantity
-from yt.units.unit_systems import create_code_unit_system, \
-    UnitSystem
+from yt.units.unit_systems import \
+    create_code_unit_system, \
+    _make_unit_system_copy
 from yt.data_objects.region_expression import \
     RegionExpression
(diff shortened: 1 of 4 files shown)


jzuhone:
@yt-fido test this please

ngoldbaum:
@yt-fido test this please

jzuhone:
@yt-fido test this please

ngoldbaum:
The test failures are unrelated. This looks good to me!

ngoldbaum:
@yt-fido test this please

PR #2037, opened 2018-09-20
add testing on python3.7

ngoldbaum:
This will hopefully also allow us to clear the answer test failures that @qobilidop pointed out on slack.


ngoldbaum:
I fixed this in another way later

PR #2038, opened 2018-09-20
[bugfix] Don't apply units to metallicity field.

brittonsmith:
## PR Summary

With the way the metallicity field is currently defined, changing the value of the solar metallicity changes the metallicity field in absolute (i.e., dimensionless units) rather that in units of `Zsun`. for example, the following script:
```
import yt
ds = yt.load("DD0046/DD0046")
print (ds.r['metallicity'].to(""))
del ds

ds = yt.load("DD0046/DD0046")
ds.unit_registry.modify('Zsun', 0.0204)
print (ds.r['metallicity'].to(""))
```
gives:
```
[9.83077184e-11 9.99523614e-11 1.00078951e-10 ... 7.38725398e-11
 8.22170876e-11 8.92656217e-11] dimensionless
[1.54863124e-10 1.57453913e-10 1.57653328e-10 ... 1.16370642e-10
 1.29515721e-10 1.40619203e-10] dimensionless
```
This is incorrect, as the dimensionless value is simply `metal_density / density` and should not change when the units change.

After this fix, the dimensionless values of the metallicity field remain constant and the values in units of `Zsun` change, as is appropriate.

## PR Checklist

### DIFF
--- a/yt/fields/fluid_fields.py
+++ b/yt/fields/fluid_fields.py
@@ -33,9 +33,6 @@ from yt.utilities.physical_constants import \
     mh, \
     kboltz
 
-from yt.utilities.physical_ratios import \
-    metallicity_sun
-
 from yt.utilities.lib.misc_utilities import \
     obtain_relative_velocity_vector
 
@@ -149,9 +146,7 @@ def setup_fluid_fields(registry, ftype = "gas", slice_info = None):
                        function=_entropy)
 
     def _metallicity(field, data):
-        tr = data[ftype, "metal_density"] / data[ftype, "density"]
-        tr /= metallicity_sun
-        return data.apply_units(tr, "Zsun")
+        return data[ftype, "metal_density"] / data[ftype, "density"]
     registry.add_field((ftype, "metallicity"),
                        sampling_type="cell",
                        function=_metallicity,


ngoldbaum:
Seems fine.

Out of curiosity why were you changing the value of `Zsun` in yt? Is that something we should expose at a higher level?

jzuhone:
The measurements for `Zsun` by different groups are all over the place, and different sub-specialties in astronomy use different values even, so we probably should expose this at a higher level. But we could just wrap around the call to `ds.unit_registry.modify` that @brittonsmith is doing above.

brittonsmith:
@ngoldbaum: yt, Enzo, and Grackle changed their default solar metallicity values a few years ago from 0.0204 to 0.01295 to be consistent with codes like Cloudy. I was working with some Enzo data that predated this change and so I wanted the metallicities to be consistent. I could see others wanting to do something similar for reasons like what @jzuhone laid out.

I'm not sure how much higher level than `ds.unit_registry.modify` we need to go. It seems pretty straightforward to me.

ngoldbaum:
We could have `ds.set_solar_metallicity()` maybe? This could be an astronomy-specific function that gets enabled only in the astronomy domain context, along with e.g. the `cosmology` attribute. Another option would be just to document the solution Britton used somewhere. I'm not sure where the best place to document it would be though.

ngoldbaum:
(I'm working on the answer test failures, for now you can ignore them)

chummels:
Wait, so now all metallicity field values will go from being in units of `Zsun` to being in absolute values?  If this is true, is there a clean way to get old functionality back, at least in a modified field definition so it works in Projections and such?  As far as I can tell, one would have to make a totally new derived field like `metallicity_zsun` or something and then just divide out `Zsun` in that, but is there something cleaner?  Or a place it could be documented (although I have no idea where).

ngoldbaum:
No, it’s still in units of solar metallicity. See the `units` jeywoed argument to `add_field` below.

brittonsmith:
@chummels, this does not change the default behavior of the code at all. There was a bug in  how changing the value of the solar metallicity changed the metallicity field. This just fixes that bug.

chummels:
Oh I see.  I guess I didn't realize that you could specify the units in that way for unitless fields.  I thought it had to be done in the field definition itself.  Cool!  Thanks for the clarification!

As a side note, the metallicity field for the Gizmo frontend *is* defined in absolute terms, not in terms of `Zsun`, so my question still remains for dealing with those.  Is there a way for me to redefine an existing metallicity field that is already in absolute terms (e.g. unitless) to be in units of `Zsun`?  Or must I create a totally new field (metallicity2 or something) and do it there?

brittonsmith:
`Zsun` is just a dimensionless unit (i.e., it has equivalent dimensions to something with units of ""). You can do:
```
import yt
t = yt.YTQuantity(1, "")
print (t.to("Zsun"))
```
If you want individual species metallicities defined in solar units, you would need to add units for the solar value of each element, i.e.,
```
# solar carbon abundance
Csun = 2.45e-04
ds.unit_registry.add('Csun', Csun, "")
print (carbon_field_array.to('Csun'))
```

chummels:
This is super helpful!  thank you!

PR #2040, opened 2018-09-21
Run xray fields value test only once for each object/field

Xarthisius:
Unless I'm missing something, `FieldValue` test is independent of `axis`


Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

Issue #1702, opened 2018-02-22
Plot answer tests are failing randomly on python3


ngoldbaum:
For some reason a subset of the plot answer tests are failing every few test runs on python3. We should try to figure out why that's happening.

The solution might be to simply update matplotlib on the test machine as we're currently running matplotlib 1.5.1.  That will require coordination with an accompanying bump in the plot answers.

ngoldbaum:
Here's what it looks like in Jenkins when this happens:

<img width="902" alt="image" src="https://user-images.githubusercontent.com/3126246/38770936-e15304fe-3fdf-11e8-8750-8ecda09bfb44.png">

ngoldbaum:
We updated the doc builder to use python 3.6 and matplotlib 2.2 but it looks like this just hapenned again:

https://tests.yt-project.org/job/yt_py3_git/272/

That means this is likely due to a real bug in yt that is causing some small amount of random jitter in the position of contours drawn by matplotlib on yt plots.

ngoldbaum:
remilestoned for 4.0 because we don't have a way to reliably trigger this behavior.

PR #2043, opened 2018-09-22
Reverse angular momentum orientation

### LINKED ISSUE #2042: Angular momentum orientation
I'm confused by the current definition of specific angular momentum as [`-ucross(r_vec, v_vec)`](https://github.com/yt-project/yt/blob/4ca693aed0ce6c3ef791be44ec8a6fe6743c82c4/yt/fields/particle_fields.py#L306). Why isn't it `ucross(r_vec, v_vec)` instead, which should be the standard definition?

I digged into the history and found the definition introduced in this [PR](https://bitbucket.org/yt_analysis/yt/pull-requests/1585). The situation was a little bit complicated, and there was actually a [conversation](https://bitbucket.org/yt_analysis/yt/pull-requests/1585/bugfix-improvement-specific-angular/activity#LC7248981T310) specifically about the angular momentum orientation.

The situation was that the definition of the angular momentum vector and those of the components were opposite. The vector was defined as [`np.cross(r_vec, v_vec)`](https://bitbucket.org/yt_analysis/yt/pull-requests/1585/bugfix-improvement-specific-angular/diff#Lyt/fields/particle_fields.pyF285), while the x component was defined as [`yv*z - zv*y`](https://bitbucket.org/yt_analysis/yt/pull-requests/1585/bugfix-improvement-specific-angular/diff#Lyt/fields/particle_fields.pyF302) (similarly for other components), corresponding to `-np.cross(r_vec, v_vec)`. Eventually the decision was made to use the latter convention.

But I'm not convinced this is a proper convention. Could we reconsider this? @cosmosquark @MatthewTurk @ngoldbaum was in the previous conversation and might have more to say.

qobilidop:
## PR Summary

Resolves #2042.

## PR Checklist


ngoldbaum:
I think this would be too much of a disruptive change for 3.5, so let's hold off on merging this until after 3.5 gets released.

cphyc:
@yt-fido test this please

cphyc:
To be fully consistent, one would need to also flip the axes for the gas angular momentum (as defined in `yt/fields/angular_momentum.py`).

qobilidop:
@cphyc Thanks for spotting this! I've made a patch.

ngoldbaum:
@qobilidop looks like a conflict has creeped in. Can you merge or rebase to clear it?

Also it's awesome that the tests pass!

qobilidop:
@ngoldbaum It's now resolved! Sorry for not noticing this earlier.

ngoldbaum:
Heh, this broke the test that @cphyc just added that tests the angular momentum value of a real dataset. Needs a minus sign ;)

qobilidop:
That's expected, right? It's great now this change is caught in tests. Thanks, @cphyc! I guess there needs to be further discussion back in #2042 then.

ngoldbaum:
That's expected given the change in this PR. I'd just add the minus sign to the broken test. That test was added in between the time when this PR was originally opened and now.

qobilidop:
I see. Thanks for the clarification.

ngoldbaum:
OK, I think this is ready to merge. I'm going to send an e-mail to yt-dev about it first though because I feel like this one might be more disruptive than usual.

qobilidop:
I totally agree. A wider awareness and discussion is necessary.

ngoldbaum:
Just because the tests haven't run on this in a while I'm going to close and reopen it to retrigger them. Assuming the tests pass I agree that we should merge this.

ngoldbaum:
The test failures on python3 are due to #1702 and can be ignored. Merging.

PR #2045, opened 2018-10-02
minor clump refactor

brittonsmith:
## PR Summary

This fixes the issue reported by @stonnes on the mailing list where a clump finding operation would sometimes end with:
```
TypeError: 'NoneType' object is not iterable
```

In doing this, I found the practice of initializing `Clump.children` to None to result in overly cumbersome if statements, so I changed the default to be an empty list. I have also deprecated the `get_lowest_clumps` function in favor of a `Clump.leaves` property. Finally, I removed two functions that have been deprecated for more than few stable releases now.

## PR Checklist

### DIFF
--- a/yt/data_objects/level_sets/api.py
+++ b/yt/data_objects/level_sets/api.py
@@ -19,9 +19,7 @@ from .contour_finder import \
 from .clump_handling import \
     Clump, \
     find_clumps, \
-    get_lowest_clumps, \
-    write_clump_index, \
-    write_clumps
+    get_lowest_clumps
 
 from .clump_info_items import \
     add_clump_info
--- a/yt/data_objects/level_sets/clump_handling.py
+++ b/yt/data_objects/level_sets/clump_handling.py
@@ -55,7 +55,6 @@ def add_contour_field(ds, contour_key):
                  units='')
 
 class Clump(TreeContainer):
-    children = None
     def __init__(self, data, field, parent=None,
                  clump_info=None, validators=None,
                  base=None, contour_key=None,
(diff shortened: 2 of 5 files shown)


yt/data_objects/level_sets/clump_handling.py
```diff
@@ -93,14 +91,25 @@ def __init__(self, data, field, parent=None,
         # Return value of validity function.
         self.valid = None
 
+    _leaves = None
+    @property
+    def leaves(self):
```
ngoldbaum: nice!

yt/data_objects/level_sets/clump_handling.py
```diff
@@ -67,6 +64,7 @@ def __init__(self, data, field, parent=None,
         self.min_val = self.data[field].min()
         self.max_val = self.data[field].max()
         self.info = {}
+        self.children = []
```
ngoldbaum: Yes, it's much clearer to add an empty one in the initializer instead of assuming it'll get created later

yt/data_objects/level_sets/api.py
```diff
@@ -19,9 +19,7 @@
 from .clump_handling import \
     Clump, \
     find_clumps, \
-    get_lowest_clumps, \
-    write_clump_index, \
-    write_clumps
```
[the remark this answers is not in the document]
brittonsmith: Good question. Firstly, `write_clumps` and `write_clump_index` are two variations if saving clump information to a text file. The difference is that `write_clumps` saves only the leaf nodes, whereas `write_clump_index` saves the full hierarchy denoting levels with indentation. 

Both of these functions have been deprecated since yt 3.3 and were removed from the narrative docs and cookbook recipes sometime before yt 3.4.

PR #2047, opened 2018-10-03
Fix travis

ngoldbaum:
I've identified an issue with the script that reports errors which is fixed here. I still need to do some other stuff to fix the tests though so this can't be merged yet.

EDIT:

I think I've fully got this working now. The issue was that the script that reported failed tests didn't actually have any logic for handling failed tests, only missing answers. There are also a couple other miscellaneous fixes for other tests that were failing, I also simplified the vr orientation test so there isn't travis-specific logic.


ngoldbaum:
@git-abhishek i'd still like you to take a look at this when you have a chance but I'm merging now to unbreak the tests.

PR #2052, opened 2018-10-08
ensure plot._background_color.data source gets updated

ngoldbaum:
This should hopefully clear the errors that are currently breaking the doc build.


ngoldbaum:
@yt-fido build docs please

Issue #1608, opened 2017-11-03
Improve documentation for ParticleProjectionPlot weight_field


osterwood:
### Bug report

**Bug summary**

The default of `weight_none` is not appropriate for many kinds of fields that ParticleProjectionPlot can plot.  This affects the color mapping and is not currently well documented.

The addition of `weight_field='particle_ones'` to the code creates the expected output.  The documentation for [ParticleProjectionPlot](http://yt-project.org/doc/reference/api/yt.visualization.particle_plots.html?highlight=particleprojectionplot#yt.visualization.particle_plots.ParticleProjectionPlot) mentions the weight field, but does not provide any examples of its use or list valid options for it. 

**Code for reproduction**

```python
#!/usr/bin/env python

import os, sys, inspect, time
import math, string

import numpy as np
import yt

CMAP_NAME = 'discrete'
COUNT     = 100000
OFFSET    = 8

yt.make_colormap([('blue', 20), ('green', 20), ('red', 20)],
                 name=CMAP_NAME, interpolate=False)

zeros = np.array([np.zeros(COUNT)])

# Create three clusters of X,Y,Z,Temp data
# Each cluster has a different XYZ position and a different (fixed) temperature
A = np.concatenate((np.random.randn(COUNT,3), zeros.T), axis=1) + [0,0,OFFSET, 1]
B = np.concatenate((np.random.randn(COUNT,3), zeros.T), axis=1) + [0,OFFSET,0, 2]
C = np.concatenate((np.random.randn(COUNT,3), zeros.T), axis=1) + [OFFSET,0,0, 3]

data = np.concatenate((A, B, C), axis=0)

print np.shape(data)

Xs = data[:,0]
Ys = data[:,1]
Zs = data[:,2]
Is = data[:,3]

print "" 
print "Compute Bounding Box" 

PADD = 1.0

min_x = np.min(Xs) - PADD
min_y = np.min(Ys) - PADD
min_z = np.min(Zs) - PADD

max_x = np.max(Xs) + PADD
max_y = np.max(Ys) + PADD
max_z = np.max(Zs) + PADD

params = {
    'cmap'    : CMAP_NAME,
    'field'   : 'temperature'
}

data = {'particle_position_x': Xs,
        'particle_position_y': Ys,
        'particle_position_z': Zs,
        params['field']: Is}

print "Rendering"

bbox = np.array([[min_x, max_x], [min_y, max_y], [min_z, max_z]])

[...]

PR #2058, opened 2018-10-10
Expand documentation of fields and weight_fields in particle plots

brittonsmith:
This closes Issue #1608. I've added examples of using fields and weight fields so show the difference between the two and expanded the docstrings a bit.


ngoldbaum:
Awesome, huge improvement. Thanks Britton!

PR #2063, opened 2018-10-13
[bugfix] yt-4.0 random order particle loading fix

### LINKED ISSUE #2059: yt.load doesn't maintain particle order for multifile datasets
### Bug report

**Bug summary**

It seems that `yt.load` on multifile snapshots (I haven't extensively tested single file datasets) fails to load the particles in the same order. This was first reported by Ben Kimock on the slack help channel.

**Code for reproduction**

```python
import yt
import numpy as np

ds = yt.load("snap_033.0.hdf5")
ad = ds.all_data()

np.save("smoothing_lengths.npy", ad[ds.fields.PartType0.smoothing_length])
```
reloading python,

```python
import yt
import numpy as np

ds = yt.load("snap_033.0.hdf5")
ad = ds.all_data()

hsml = np.load("smoothing_lengths.npy")
assert(np.array_equal(ad[ds.fields.PartType0.smoothing_length], hsml))
```

**Version Information**
  * Operating System: Ubuntu 18.04
  * Python Version: 
  * yt version: yt-4.0

AshKelly:
## PR Summary

Added an extra keyword to the datafiles sort, so they don't just use the filename, but also use the start particle as well - this should fix subfiles loading in a random order.

fixes #2059 

## PR Checklist

- **NA** New features are documented, with docstrings and narrative docs
- **NA** Adds a test for any bugs fixed. Adds tests for new features.

### DIFF
--- a/yt/frontends/flash/io.py
+++ b/yt/frontends/flash/io.py
@@ -186,7 +186,7 @@ class IOHandlerFLASHParticle(BaseIOHandler):
         px, py, pz = self._position_fields
         p_fields = self._handle["/tracer particles"]
         assert(len(data_files) == 1)
-        for data_file in sorted(data_files):
+        for data_file in sorted(data_files, key=lambda x: (x.filename, x.start)):
             pcount = self._count_particles(data_file)["io"]
             for ptype, field_list in sorted(ptf.items()):
                 total = 0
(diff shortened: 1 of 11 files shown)


AshKelly:
Hey @ngoldbaum - I've identified the problem with the gadget frontend, and whilst I was at it I tried to ensure that the consistency is maintained across other. I do worry, that I may have added this to datafiles that do not have those attributes. 

I guess the tests probably wouldn't pick this up?

ngoldbaum:
There are tests for the I/O routines for all of the frontends, but you'll need to run them locally since right now travis and appveyor aren't running any tests with real datasets. 

In particular, I'd ensure that all the tests in `yt/data_objects/tests/test_sph_data_objects.py` pass. You'll need to download the data files used in those tests and stage them in your `test_data_dir` as specified in your yt config file.

Thanks for taking this on and finding the fix!

AshKelly:
After this commit, I think the frontends are all working except `gadget_fof`, but that doesn't work on the yt-4.0 branch anyway, for some reason

ngoldbaum:
Thanks for diving into this one!

PR #2064, opened 2018-10-14
Doc dev

neutrinoceros:
Minor revisions and corrections to the documentation for *create a new frontend*.

* Follow PEP8 recommendation regarding spaces around `=` in function arguments (avoid them)
* Following a comment in `frontends/gamer/data_structures.py`, do not initialize `AMRGridPatch.Parent` as `[]` because
    - it's not meant to be a list (I think)
    - some existing functionalities are made to catch a lacking init in this attribute by testing if it's `None`
* workaround a possibly confusing variable naming where `grid` is being used as a local variable in a class method, when the class itself has an attribute `grid`.


ngoldbaum:
Thanks for the fixes, this is a clear improvement!

PR #2073, opened 2018-10-22
avoid use of deprecated field names internally

ngoldbaum:
Currently fields like "particle_angular_momentum" end up going through one of these code paths and triggering deprecation warnings. Let's use the new names for these fields internally. The deprecated field names are still available for use with this change, we just don't use them as dependencies for other fields in the yt field system anymore.


ngoldbaum:
The test failure is due to a timeout on MacOS, presumably because the mac builders on travis are under heavy disk load at the moment. Merging.

PR #1844, opened 2018-06-20
Adding a display_ytarray() function

MatthewTurk:
This is a draft, possibly-to-be-rejected, function that adds a `display_ytarray()` function.  At present, it is separately defined from the usual `YTArray` objects, as it only works with reasonably sized objects.  It's possible we could put it into the `YTArray` itself (and just bail on ipython display if the size is too big), but I'm not sure if we want to do that.

It looks like this by default:

screenshot 2018-06-20 at 4 00 02 pm

The boxes are read-only.  Changing the displayed unit does *not* change the underlying array, but only the display.

screenshot 2018-06-20 at 4 01 03 pm

PR #2077, opened 2018-10-25
Making unicode ellipsis into three periods

MatthewTurk:
This may or may not be the cause of occasional doc failures, but it did
show up in the log for #1844.


Xarthisius:
@yt-fido test this please

Xarthisius:
Travis failure was unrelated (exactly the same tests fail on other PRs right now)

PR #2078, opened 2018-10-25
Fixing three flake8 errors

MatthewTurk:
These three flake8 errors were showing up as failures on a few recent
builds, and were all straightforward fixes.


ngoldbaum:
I guess a new version of flake8 came out? There's still some warnings it looks like, which also look new. Unless you ant to fix all of them I'd say it's ok to add the new warnings to the flake8 rule ignore list.

brittonsmith:
Yeah, it seems to be a new flake8. W504 is particularly annoying.

MatthewTurk:
I added them in a quick second commit!

PR #2084, opened 2018-11-01
ensure all failed tests get entries in the failure report

ngoldbaum:
Right now if more than one test fails for a single test specification (for example, if there is a `unittest.TestCase` subclass that generates multiple independent tests), only one entry from that test gets included in the report. This ensures that all the entries get included. Once i've verified that this does indeed fix the issue I will update the test answers in this PR to fix the broken travis tests.


ngoldbaum:
the jenkins failures are unrelated, i'm merging this now to unbreak travis

PR #1966, opened 2018-08-10
Add support for cartopy projections with geographic data

munkm:
## PR Summary

This PR adds in support for projecting geographic data sources on various map types. As it stands, the code in this PR assumes that geographic data is read in as `PlateCarree` type and can be projected onto the maps listed here: https://scitools.org.uk/cartopy/docs/v0.15/crs/projections.html.  

## PR Checklist

- [x] Code passes flake8 checker
- [x] New features are documented, with docstrings and narrative docs
- [x] Adds a test for any bugs fixed. Adds tests for new features.

PR #2091, opened 2018-11-08
sort image paths before reading them in image comparison tests

ngoldbaum:
This should (hopefully) fix occasional answer test failures for image comparison tests that save more than one image to disk. The sorting of the list returned by `glob.glob` is arbitrary, so the ordering of images we get back is also arbitrary.

I wouldn't be surprised if some of the existing answers were sensitive to this ordering.


ngoldbaum:
OK cool, looks like this didn't matter for any of the stored answers. Now let's merge and see if this fixes the failures on #1966.

PR #2092, opened 2018-11-09
[RAMSES] Fix RT and groups

cphyc:
## PR Summary

This PR fixes a small bug in Python3 that was preventing from using groups with yt. It also enables the possibility to have variable speed of light for RT runs.

## PR Checklist

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -401,11 +401,8 @@ class RAMSESDataset(Dataset):
             # note: we exclude the unlikely event that one of the group is actually a file
             # instad of a folder
             self.num_groups = len(
-                list(
-                    filter(lambda e: os.path.isdir(e),
-                           glob(os.path.join(root_folder, 'group_?????')))
-                    )
-                )
+                [_ for _ in glob(os.path.join(root_folder, 'group_?????'))
+                 if os.path.isdir(_)])
             self.root_folder = root_folder
         else:
             self.root_folder = os.path.split(filename)[0]
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -401,8 +401,11 @@ class RAMSESDataset(Dataset):
             # note: we exclude the unlikely event that one of the group is actually a file
             # instad of a folder
             self.num_groups = len(
-                filter(lambda e: os.path.isdir(e),
-                       glob(os.path.join(root_folder, 'group_?????'))))
+                list(
+                    filter(lambda e: os.path.isdir(e),
+                           glob(os.path.join(root_folder, 'group_?????')))
+                    )
+                )
             self.root_folder = root_folder
         else:
             self.root_folder = os.path.split(filename)[0]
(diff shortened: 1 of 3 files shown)


MatthewTurk:
Variable speed of light?

> Farnsworth: These are the dark matter engines I invented. They allow my starship to 
> travel between galaxies in mere hours.
> 
> Cubert: That's impossible. You can't go faster than the speed of light.
> 
> Farnsworth: Of course not! That's why scientists increased the speed of light in 2208.

yt/frontends/ramses/data_structures.py
```diff
@@ -401,8 +401,11 @@ def __init__(self, filename, dataset_type='ramses',
             # note: we exclude the unlikely event that one of the group is actually a file
             # instad of a folder
             self.num_groups = len(
-                filter(lambda e: os.path.isdir(e),
-                       glob(os.path.join(root_folder, 'group_?????'))))
+                list(
+                    filter(lambda e: os.path.isdir(e),
```
MatthewTurk: I know this isn't new, but any chance we could rework this to a list comprehension?

`[_ for _ in glob(os.path.join(root_folder, 'group_?????')) if os.path.isdir(_)]`

yt/frontends/ramses/field_handlers.py
```diff
@@ -444,12 +444,17 @@ def read_rhs(cast):
...
         with open(fname, 'r') as f:
+            # Read nRTvar, nions, ngroups, iions
             for i in range(4): read_rhs(int)
             f.readline()
+            # Read X and Y fractions
             for i in range(2): read_rhs(float)
             f.readline()
-            for i in range(3): read_rhs(float)
+            # Reat unit_np, unit_pfd, rt_c_frac
+            for i in range(2): read_rhs(float)
```
MatthewTurk: Can we get an updated comment here?  Others might catch it, but it took me a second to realize that the special case was for `rt_c_frac` being multiple items on a line.

cphyc: Sure.

cphyc:
Hi,
I don't have access to a small dataset with variable speed of light (as I don't use VSL myself). I can try to trigger the other bug in the existing group dataset that's online.

For the variable speed of light, this is an approximation that's made in RAMSES when dealing with cosmological radiative transfer simulations. In regions with fast varying luminosities (e.g. galaxies), the speed of light is close to its real value while in empty regions, the speed of light is artificially decreased (so that the timestep can be increased). The `rt_c_frac` encodes the effective speed of light used by RAMSES compared to the physical value of `c`. In the case of variable speed of light, this is not a constant but an array with one value per AMR level. I hope this makes sense!

ngoldbaum:
If it's a huge pain to add new tests don't worry too much.

yt/frontends/ramses/field_handlers.py
ngoldbaum: also minor nit, can you put the calls to `read_rhs` in their own blocks, like this:

```
for i in range(2): 
    read_rhs(float)
```

I know it was already like that, formatting with explicit carriage returns and tabs makes it clearer there's a new block there.

ngoldbaum:
Test failure is unrelated, I'm merging.

cphyc:
@mtrebitsch

PR #2094, opened 2018-11-15
remove noise from line integral convolution in zero-value regions

jisuoqing:
## PR Summary

In the regions with zero vector field values, the line integral convolution (LIC) just creates noise. This PR removes these regions from LIC, which is particular useful when the non-cartesian simulation region does not fill the entire plot canvas.

## PR Checklist

### DIFF
--- a/yt/utilities/lib/line_integral_convolution.pyx
+++ b/yt/utilities/lib/line_integral_convolution.pyx
@@ -81,6 +81,8 @@ def line_integral_convolution_2d(
 
     for i in range(w):
         for j in range(h):
+            if vectors[i,j,0]==0 and vectors[i,j,1]==0:
+                continue
             x = i
             y = j
             fx = 0.5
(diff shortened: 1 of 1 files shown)


yt/utilities/lib/line_integral_convolution.pyx
```diff
@@ -104,5 +104,8 @@ def line_integral_convolution_2d(
                         &x, &y, &fx, &fy, w, h)
                 l-=1
                 result[i,j] += kernel[l]*texture[x,y]
-                    
+
+            if vectors[i,j,0]==0 and vectors[i,j,1]==0:
+                result[i,j] = 0.
```
ngoldbaum: I think this early return condition should come right after the double for loops, there's no need to do the double while loop for these pixels if we're just going to overwrite the result. That should make it clearer that this is an early return condition as well.

PR #2101, opened 2018-11-26
adapt to new colorbar behavior in mpl 3.0. fixes #2099

### LINKED ISSUE #2099: Messed up colorbars for phase plots in matplotlib 3
The following script:

```
import yt
plot = yt.PhasePlot(ds.all_data(), 'density', 'temperature', 'cell_mass', weight_field=None)
plot.save()
```

Produces the following result in matplotlib 3.0.2 (as well as in 3.0.1 and 3.0.0):

galaxy0030_2d-profile_density_temperature_cell_mass

Whereas with matplotlib 2.2.3, I get:

galaxy0030_2d-profile_density_temperature_cell_mass

Will try to make an example outside of yt that reproduces this behavior and report upstream to matplotlib...

ngoldbaum:
Matplotlib 3.0 includes a number of improvements to colorbars, one of which makes a hack in yt's handling of colorbars unnecessary, leading to the weird behavior in #2099. The fix is to guard the hack behind a matplotlib version check.

### DIFF
--- a/yt/visualization/profile_plotter.py
+++ b/yt/visualization/profile_plotter.py
@@ -49,6 +49,8 @@ from yt.funcs import \
     matplotlib_style_context, \
     iterable
 
+MPL_VERSION = LooseVersion(matplotlib.__version__)
+
 def get_canvas(name):
     from . import _mpl_imports as mpl
     suffix = get_image_suffix(name)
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Ugh wish I'd looked closer at the test failures when I updated the tests to use matplotlib 3, would have caught this then. Oh well.... Will push new test answers shortly.

PR #2102, opened 2018-11-26
[bugfix] Fix incorrect weight function for cluster temperatures

jzuhone:
yt has a derived field known as `"mazzotta_weighting"` whose sole function is to provide a weight for projections which approximates temperature maps of galaxy clusters taken by Chandra or XMM-Newton. Reference is Equation 15 of http://adsabs.harvard.edu/abs/2004MNRAS.354...10M.

I was looking at this field again today and noticed that it has the wrong exponent! I introduced it into yt way-back-when, and I've used it quite a bit, but I don't know who else has. In any case, here is the fix.


jzuhone:
Thanks @ngoldbaum

PR #2106, opened 2018-11-28
 remove caching for particle indices in cut region. fixes #2104 

### LINKED ISSUE #2104: cut_region can't be used to derive angular momentum quantities
### Bug report

I'm trying to calculate the angular momentum vector of a sphere, and in particular, for the region inside the sphere that have density higher than certain level. But result from the cut_region cannot be used to calculate the angular momentum quantities. So here is the code I used for one of the test datasets on yt. 

Short version of the bug: 
IndexError: boolean index did not match indexed array along dimension 0; dimension is 44461 but corresponding boolean dimension is 3218

```python
# Paste your code here
#
#
```
`import yt

sp = ds.sphere((0.5, 0.5, 0.5), (5, "kpc"))
dense_sp = sp.cut_region(['obj["H_p0_number_density"]>= 1e-2'])
dense_sp.quantities.angular_momentum_vector()`

**Version Information**

  * Operating System: Mac Os High Sierra 10.13.6
  * Python Version: 3.6.6
  * yt version: 3.5.dev0 (installed from source)

ngoldbaum:
In a chunked read (e.g. what happens when finding the value of a derived quantity), the size of a data object can change depending on which chunk we're looking at. This breaks the caching in the `_part_ind` property of the cut region data object. My quick and dirty fix is just to get rid of the caching. I *could* cache based on the underlying data chunk, which will be unique for each chunk, but then I'd blow up the memory requirements for doing any read on a cut_region, since that will require storing a mask for each chunk.

Suggestions from e.g. @brittonsmith would be great, particularly if the caching is a big performance win for the clump finder. Perhaps we could do the caching over there instead of on the cut region data object itself?

### DIFF
--- a/yt/data_objects/selection_data_containers.py
+++ b/yt/data_objects/selection_data_containers.py
@@ -876,7 +876,6 @@ class YTCutRegion(YTSelectionContainer3D):
         self.conditionals = ensure_list(conditionals)
         self.base_object = data_source
         self._selector = None
-        self._particle_mask = {}
         # Need to interpose for __getitem__, fwidth, fcoords, icoords, iwidth,
         # ires and get_data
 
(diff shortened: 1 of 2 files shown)


brittonsmith:
@ngoldbaum, I have a reasonably large test case I can run. I'll do that and report back.

brittonsmith:
I've done a reasonably large clump-finding run and here are some timing results:

Before this PR:
- clump-finding: 8265 seconds
- write-out: 5 seconds

After this PR:
- 8426 seconds
- write-out: 20 seconds

The time difference in the clump finding is ~2% and probably close to the run-to-run variation. The write-out time is simply the time to run `master_clump.save_as_dataset()` with particle fields. The factor of 4 increase is probably real, but it's a negligible fraction of the total time either way, so I can live with it. The test I ran was a pretty large example, so I think most use cases wouldn't see a big impact.

PR #2107, opened 2018-11-29
Adding support for active particles

fearmayo:
adding support for active particles

## PR Checklist


ngoldbaum:
Hi,

we already have support for enzo-3.0 active particles. What issues did you run into this these changes solve? Do the existing test datasets we have with active particles still work correctly with these changes?

ngoldbaum:
It looks like the answer to my second question is no, since the tests for loading active particle data are failing.

fearmayo:
Hi Nathan,

The problem I found was that while there is support for enzo3.0 active particles there is no support for enzo-dev active particles. 

OK. Looks like I need to follow up and see what now breaks with the ap tests.

ngoldbaum:
enzo-dev actuve particles? like for my WIP pull request to enzo? Have you continues that work? Because it’s a work-in-progress pull request we did not make the necessary changes to yt to add support for it yet.

If you have taken that work farther than I took it, great, however I think we need to wait on adding support for it here until the Enzo pull request is merged at least. Also we need to continue supporting enzo-3.0 active particles since people are doing production work with that, so you’ll need to make sure all the existing tests pass. In addition we’ll need a new test dataset and new tests for the new functionality
you want to add.

Perhaps it would be best to start a discussion about the way forward for active particles on the enzo-dev mailing list. I’m very keen to get that Enzo pull request merged and wasn’t aware that people were using the WIP code!

fearmayo:
Yeah I pulled that WIP into my branch and it's working (there are probably a few updates required in your PR and docs of course but not too much). I have a couple of more things I need to check but in principle it works fine. I'm taking the Renaissance datasets and zooming into them with active particles and since Enzo-3.0 is being effectively discontinued I reverted to enzo-dev.

ngoldbaum:
OK, well hopefully the path forward I outlined in my previous pull request will allow you to make progress, just to summarize here's what needs to happen on this PR before it can be merged:

- [ ] All existing tests pass
- [ ] Add new tests, including adding a new example dataset to the data archive at yt-project.org/data and new tests that use that new dataset.
- [ ] Add a small update to the docs on loading enzo data
- [ ] Enzo pull request merged, or at least have a discussion on enzo-dev about the active particle storage format so ensure it gets stabilized in a fashion that people agree on.

matthewturk:
@fearmayo Is this PR still on your radar?

matthewturk:
Would anyone be able to take this PR under their wing?

fearmayo:
> Would anyone be able to take this PR under their wing?

Yes. I can restart this. The APs are in Enzo now. Give me a few weeks though as teaching prep is weighing on me right now.

brittonsmith:
@yt-fido test this please

brittonsmith:
Does anyone understand the Build and Test failures?

yt/frontends/enzo/io.py
```diff
@@ -75,15 +78,21 @@ def _read_particle_fields(self, chunks, ptf, selector):
...
-                    else:
                         pds = ds
+                    else:
+                        if g.NumberOfActiveParticles[ptype] == 0:
+                            continue
+                        pds = ds.get("Active Particles/%s" % ptype)
                     pn = _particle_position_names.get(ptype, r"particle_position_%s")
                     x, y, z = (
-                        np.asarray(pds.get(pn % ax)[()], dtype="=f8") for ax in "xyz"
+                        np.asarray(pds.get(pn % ax).value, dtype="=f8") for ax in "xyz"
```
brittonsmith, line 95: The use of `.value` has been removed in newer versions of h5py. This could be what's causing the problem. If you switch it back to how it was, this should get things working.
```suggestion
                        np.asarray(pds.get(pn % ax)[()], dtype="=f8") for ax in "xyz"
```

yt/frontends/enzo/io.py
```diff
@@ -75,16 +78,24 @@ def _read_particle_fields(self, chunks, ptf, selector):
                     continue
                 ds = f.get("/Grid%08i" % g.id)
                 for ptype, field_list in sorted(ptf.items()):
-                    if ptype != "io":
+                    if ptype == "io":
+                        if g.NumberOfParticles == 0:
```
chrishavlin, line 82: I think should this be `g.NumberOfParticles[ptype]`?

chrishavlin: oh... hmm. Looks like I was wrong about this? doesn't look like any tests failed due to this.

chrishavlin: oh, ya, I skimmed through too quickly. `NumberOfParticles` is not the same as `NumberOfActiveParticles`... Ignore above comments :)

yt/frontends/enzo/data_structures.py
```diff
@@ -370,20 +370,21 @@ def _detect_active_particle_fields(self):
...
-                    sh = node[ptype][field].shape
-                    if len(sh) > 1:
-                        self.io._array_fields[field] = (sh[1],)
-                fields += [(ptype, field) for field in _fields.pop(ptype)]
-            handle.close()
+            if ptype != "DarkMatter":
+                gs = self.grids[select_grids > 0]
+                g = gs[0]
+                handle = h5py.File(g.filename, "r")
+                node = handle["/Grid%08i/Active Particles/" % g.id]
```
brittonsmith, line 377: The failing test is because our sample data has a different name for the active particles in the file. The code below made it work for me.
```suggestion
                grid_group = handle[f"/Grid{g.id:08d}"]
                found = False
                for pname in ["Active Particles", "Particles"]:
                    if pname in grid_group:
                        found = True
                        break
                if not found:
                    raise RuntimeError(
                        "Could not find active particle group in data.")
                node = grid_group[pname]
```

yt/frontends/enzo/io.py
```diff
@@ -75,12 +78,25 @@ def _read_particle_fields(self, chunks, ptf, selector):
...
                             continue
-                        pds = ds.get(f"Particles/{ptype}")
-                    else:
                         pds = ds
+                    else:
+                        if g.NumberOfActiveParticles[ptype] == 0:
+                            continue
+                        for pname in ["Active Particles", "Particles"]:
+                            pds = ds.get(f"{pname}/{ptype}")
+                            if pds != None:
```
brittonsmith, line 94:
```suggestion
                            if pds is not None:
```

yt/frontends/enzo/io.py
```diff
@@ -75,12 +78,25 @@ def _read_particle_fields(self, chunks, ptf, selector):
...
+                                break
+                        if pds == None:
```
brittonsmith, line 96:
```suggestion
                        if pds is None:
```

yt/frontends/enzo/data_structures.py
```diff
@@ -370,20 +370,29 @@ def _detect_active_particle_fields(self):
...
+                g = gs[0]
+                handle = h5py.File(g.filename, "r")
+                grid_group = handle[f"/Grid{g.id:08d}"]
+                found = False
+                for pname in ["Active Particles", "Particles"]:
+                    if pname in grid_group:
+                        found = True
+                        break
+                if not found:
+                    raise RuntimeError("Could not find active particle group in data.")
```
neutrinoceros, line 384: I know the "for-else" construct is somewhat "obscure", but this is exactly what it's supposed to be used for :)
```suggestion
                for pname in ["Active Particles", "Particles"]:
                    if pname in grid_group:
                        break
                else:
                    raise RuntimeError("Could not find active particle group in data.")
```

neutrinoceros, line 384: I know the "for-else" construct is somewhat "obscure", but this is exactly what it's supposed to be used for :)
```suggestion
                for pname in ["Active Particles", "Particles"]:
                    if pname in grid_group:
                        break
                else:
                    raise RuntimeError("Could not find active particle group in data.")
```

yt/frontends/enzo/data_structures.py
```diff
@@ -370,20 +370,29 @@ def _detect_active_particle_fields(self):
...
+                node = grid_group[pname]
+                for ptype in (str(p) for p in node):
+                    if ptype not in _fields:
+                        continue
+                    for field in (str(f) for f in node[ptype]):
+                        _fields[ptype].append(field)
+                        sh = node[ptype][field].shape
+                        if len(sh) > 1:
+                            self.io._array_fields[field] = (sh[1],)
```
neutrinoceros, line 393: Trying to use better numpy idioms here
```suggestion
                        if node[ptype][field].ndim > 1:
                            self.io._array_fields[field] = (node[ptype][field][1],)
```

yt/frontends/enzo/data_structures.py
```diff
@@ -370,20 +370,29 @@ def _detect_active_particle_fields(self):
...
-                    _fields[ptype].append(field)
-                    sh = node[ptype][field].shape
-                    if len(sh) > 1:
-                        self.io._array_fields[field] = (sh[1],)
-                fields += [(ptype, field) for field in _fields.pop(ptype)]
-            handle.close()
+            if ptype != "DarkMatter":
+                gs = self.grids[select_grids > 0]
+                g = gs[0]
+                handle = h5py.File(g.filename, "r")
```
neutrinoceros, line 376: I can't make this a suggestion, but `h5py.File` now implements the context manager protocol. In other words:
```python
handle = h5py.File(...)
...
handle.close()
```
can (and should) be refactored into
```python
with h5py.File(...) as handle:
    ...
```

yt/frontends/enzo/io.py
```diff
@@ -75,12 +78,25 @@ def _read_particle_fields(self, chunks, ptf, selector):
...
+                        pds = ds
+                    elif ptype == "DarkMatter":
                         if g.NumberOfActiveParticles[ptype] == 0:
                             continue
-                        pds = ds.get(f"Particles/{ptype}")
-                    else:
                         pds = ds
+                    else:
+                        if g.NumberOfActiveParticles[ptype] == 0:
+                            continue
```
neutrinoceros, line 91:
```suggestion
                    elif not g.NumberOfActiveParticles[ptype]:
                        continue
                    else:
```

yt/frontends/enzo/io.py
```diff
@@ -75,12 +78,25 @@ def _read_particle_fields(self, chunks, ptf, selector):
...
+                        for pname in ["Active Particles", "Particles"]:
+                            pds = ds.get(f"{pname}/{ptype}")
+                            if pds is not None:
+                                break
+                        if pds is None:
```
neutrinoceros, line 96:
```suggestion
                        else:
```

fearmayo: It's not obvious to me that `else` is the right approach here. The 'if' statement is checking if 'pds' is 'None'

yt/frontends/enzo/data_structures.py
fearmayo: Interesting. So everything between the 'with' and the 'close is indented and in that scope? Is that how that works?

neutrinoceros:
I've reviewed this purely with code-quality in mind. I don't know enough about what's actually going on to do an _actual_ review here. FWIW I think that actual test coverage for this is required if we're going to merge it before the yt 4.0 release

fearmayo:
> I've reviewed this purely with code-quality in mind. I don't know enough about what's actually going on to do an _actual_ review here. FWIW I think that actual test coverage for this is required if we're going to merge it before the yt 4.0 release

Thanks very merge. Test coverage also on the way.

yt/frontends/enzo/io.py
neutrinoceros: but that should only ever happen in case the `break` statement above isn't hit, correct ?
The `else` in `for` loops can be read as `if no break`. I've thought about this for a a minute and I really can't see any case where it's not equivalent, because the for loop is _always_ going to run (given the iterable is hard-coded as non-empty), so I can't see how `pds` could end up as `None` in any other scenario than "the break wasn't hit"

yt/frontends/enzo/data_structures.py
neutrinoceros: that's it. Basically it works as a the builtin `open` function. At the cost of adding one indentation level to your code, you gain confidence in that the file will always be closed correctly, even if you forget to manually close it, or if an exception is raised before you do.

```python
with open(...) as handle:
    ...
```
is essentially the same as
```python
handle = open(...)
try:
    ...
finally:
    handle.close()
```

munkm:
> FWIW I think that actual test coverage for this is required if we're going to merge it before the yt 4.0 release

As discussed in the development slack, the test coverage will come as a follow up PR from @brittonsmith with data included.

brittonsmith (approved):

neutrinoceros:
Depending on how you guys want to approach merging this and the companion PR, consider you have my approval to move forward. I would still prefer to see some test right away but I don't want to be a blocker

yt/frontends/enzo/data_structures.py
```diff
@@ -370,20 +370,26 @@ def _detect_active_particle_fields(self):
...
+                else:
+                    raise RuntimeError("Could not find active particle group in data.")
+                node = grid_group[pname]
+                for ptype in (str(p) for p in node):
+                    if ptype not in _fields:
+                        continue
+                    for field in (str(f) for f in node[ptype]):
+                        _fields[ptype].append(field)
+                        if node[ptype][field].ndim > 1:
+                            self.io._array_fields[field] = (node[ptype][field][1],)
```
brittonsmith, line 390: This needs to get the shape instead of the value.
```suggestion
                            self.io._array_fields[field] = (node[ptype][field].shape[1:],)
```

munkm:
Woooo @fearmayo! Really great job getting this PR over the line. We really appreciate your patience and the time you put into this. Thank you!

PR #2108, opened 2018-11-30
This breaks for dim > 255, so upcast.

MatthewTurk:
We can't use uint8 for access along a dimension with more than 255
cells.


ngoldbaum:
Fair enough, good catch!

PR #2109, opened 2018-12-02
fix error message for low dimensionality data in CIC deposit

ngoldbaum:
The error was generating a string formatting exception. The fix is to make sure we use a formatted string.


Xarthisius:
@yt-fido test this please

Issue #1494, opened 2017-07-15
Cell volume for demeshening data


astrosmith:
### Bug report

**Bug summary**

When a particle dataset is loaded (e.g. unstructured mesh Gizmo data) some of the the fields in derived_field_list such as cell volume may need to be redefined or return a better error message.

**Code for reproduction**

```python
import yt
ds = yt.load(file)
ad = ds.all_data()
ad['cell_volume']
```

**Actual outcome**

```
---------------------------------------------------------------------------
YTNonIndexedDataContainer                 Traceback (most recent call last)
<ipython-input-25-78e4331205e0> in <module>()
----> 1 ad['cell_volume']

~/yt/yt/data_objects/data_containers.py in __getitem__(self, key)
    281                 return self.field_data[f]
    282             else:
--> 283                 self.get_data(f)
    284         # fi.units is the unit expression string. We depend on the registry
    285         # hanging off the dataset to define this unit object.

~/yt/yt/data_objects/data_containers.py in get_data(self, fields)
   1369 
   1370         fields_to_generate += gen_fluids + gen_particles
-> 1371         self._generate_fields(fields_to_generate)
   1372         for field in list(self.field_data.keys()):
   1373             if field not in ofields:

~/yt/yt/data_objects/data_containers.py in _generate_fields(self, fields_to_generate)
   1389                 fi = self.ds._get_field_info(*field)
   1390                 try:
-> 1391                     fd = self._generate_field(field)
   1392                     if fd is None:
   1393                         raise RuntimeError

~/yt/yt/data_objects/data_containers.py in _generate_field(self, field)
    318                 tr = self._generate_particle_field(field)
    319             else:
--> 320                 tr = self._generate_fluid_field(field)
    321             if tr is None:
    322                 raise YTCouldNotGenerateField(field, self.ds)

~/yt/yt/data_objects/data_containers.py in _generate_fluid_field(self, field)
    338             rv = self._generate_spatial_fluid(field, ngt_exception.ghost_zones)
    339         else:
--> 340             rv = finfo(gen_obj)
    341         return rv
    342 

~/yt/yt/fields/derived_field.py in __call__(self, data)
    240                 "for %s" % (self.name,))
    241         with self.unit_registry(data):
--> 242             dd = self._function(self, data)
    243         for field_name in data.keys():
    244             if field_name not in original_fields:

~/yt/yt/geometry/coordinates/coordinate_handler.py in _dds(field, data)
[...]

PR #2110, opened 2018-12-03
[WIP] Add a volume field for non-grid datasets

jzuhone:
We lack a volume field for non-grid datasets. We have `"cell_volume"` for the latter, which is aliased to `"volume"`. There is also the notion of `"cell_mass"`, which is aliased to `"mass"`, but that field also does work for non-grid datasets. 

This PR adds a definition for a `"volume"` field for non-grid datasets. This closes Issue #1494. 

I'm leaving it as WIP since I'm sure others may have ideas on how to best do this. Once we have settled on an agreed-upon way, I'd like to run through the code and replace instances of `"cell_mass"` with `"mass"` and `"cell_volume"` with `"volume"` in field definitions and other places.

### DIFF
--- a/yt/fields/fluid_fields.py
+++ b/yt/fields/fluid_fields.py
@@ -65,6 +65,14 @@ def setup_fluid_fields(registry, ftype = "gas", slice_info = None):
                        units=unit_system["mass"])
     registry.alias((ftype, "mass"), (ftype, "cell_mass"))
 
+    if (ftype, "volume") not in registry:
+        def _volume(field, data):
+            return data[ftype, "mass"]/data[ftype, "density"]
+        registry.add_field((ftype, "volume"),
+                           sampling_type="local",
+                           function=_volume,
+                           units=unit_system["volume"])
+
     def _sound_speed(field, data):
         tr = data.ds.gamma * data[ftype, "pressure"] / data[ftype, "density"]
         return np.sqrt(tr)


ngoldbaum:
This makes sense to me although I’d like an SPH expert to weigh in @ashkelly.

It would be good to figure out which fields depend on volume and see if there are alternative definitions for SPH data that make more sense without the volume.

AshKelly:
Yeah, I think people do occasionally use the volume field. I've recently seem some volume weighted SPH plots. I guess this is fairly easy to set up - something of the order of magnitude of the smoothing length cubed would be reasonable?

AshKelly:
@JBorrow probably has very strong opinions here

jzuhone:
@AshKelly is there a reason why we wouldn't want to use volume = mass / density?

jzuhone:
@AshKelly that's what the code change here does, for clarification, is volume = mass / density. the context is that there are a lot of places where we use the volume field and we'd want it to be something that's not just the right order of magnitude but exact, like say we have a per volume emissivity in erg/cm**3/s but we want to get a luminosity in erg/s.

jzuhone:
And then of course we have non-AMR data that doesn't have a smoothing length either.

ngoldbaum:
Why can’t we use the local smoothing length? non-SPH data usually has a well/defined volume element so we shouldn’t worry about that here I think.

jzuhone:
@ngoldbaum I think that we can use it (and it sounds like we must). Sorry, my ignorance of SPH is showing here, but I appreciate the education I'm getting from @AshKelly.

jzuhone:
@JBorrow:

> It's unclear to me how you would make the jump from the per volume emissivity to luminosity in this way, dividing out the volume; what you would really want to do here is integrate out that volume dependence.

So right now (for example) in `yt/fields/xray_emissivity_fields` we compute X-ray emissivity and the luminosity in the following (standard) way:

```
e_X = nH*nH*lambda(T,Z)
L_X = e_X*V = nH*nH*V*lambda(T,Z)
```
where `nH` is a hydrogen number density, `lambda` is the emissivity function, and  

So in theory you could do this I suppose:

```
nH*V = rho*V/(muH*nH) = M/(muH*mp)
L_X = nH*M/(muH*mp)*lambda(T,Z)
```
where `M` is the mass of the cell or particle in yt, IOW the `('gas', 'mass')` field, `muH` is the mean molecular weight of hydrogen and `mp` is the proton mass. This would compute the field in terms of things that we have already. Not pretty or obvious, but accurate. 

Not sure we could do this for all fields that use `"cell_volume"`, but it's worth a shot.

ngoldbaum:
Thanks for the detailed explanation!

jzuhone:
I had to fix the math in there just now but the idea is the same. Thanks, all. I think I will close this and work this problem in the way @jborrow suggested.

PR #2114, opened 2018-12-06
Correct computation of angular momentum

cphyc:
## PR Summary

Until now, there was a mistake in the computation of angular momentum which was leading to the angular momentum units being messed up.

## PR Checklist

### DIFF
--- a/yt/fields/tests/test_angular_momentum.py
+++ b/yt/fields/tests/test_angular_momentum.py
@@ -8,13 +8,14 @@
 
 import numpy as np
 
+import yt
 from yt.testing import \
     assert_allclose_units, \
-    fake_random_ds, \
     requires_file
 
+@requires_file('output_00080/info_00080.txt')
 def test_AM_value():
-    ds = fake_random_ds(16)
+    ds = yt.load('output_00080/info_00080.txt')
 
     sp = ds.sphere([.5]*3, (0.1, 'code_length'))
 
@@ -30,5 +31,3 @@ def test_AM_value():
     sAM = ds.arr([sp['specific_angular_momentum_'+k] for k in 'xyz']).T
 
     assert_allclose_units(sAM_manual, sAM)
-
-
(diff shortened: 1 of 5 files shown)


yt/fields/tests/test_angular_momentum.py
```diff
@@ -0,0 +1,33 @@
...
+#-----------------------------------------------------------------------------
+
+import numpy as np
+
+import yt
+from yt.testing import \
+    assert_allclose_units, \
+    requires_file
+
+@requires_file('output_00080/info_00080.txt')
```
ngoldbaum: Any chance we can run this test using e.g. `fake_amr_ds` or some other dataset that doesn't need to download. I don't think anything about this test is specific to this RAMSES output?

ngoldbaum:
The test failures on python3 are unrelated to this PR.

yt/fields/tests/test_angular_momentum.py
cphyc: I'm not sure why but the test passes even without the fix when using `fake_amr_ds`. Any guess why?

cphyc:
Both were wrong. This is essentially due to the centre being expressed in the wrong units (hence the relative position was wrong, etc.).

cphyc:
Both were wrong. This is essentially due to the centre being expressed in the wrong units (hence the relative position was wrong, etc.).

chummels:
That's odd.  I've definitely used this in the past and it has worked to get the angular momentum vector for a galactic disk.  I never used the magnitude of the vector for anything, but the direction seemed to be spot on.  Also, I would've thought the flurry of PRs regarding angular momentum in yt-4 a few months back would've caught this issue: PR #1902 PR #1934 .

cphyc:
The bug is triggered in the case where units mismatch, so I guess that if your datasets were using the "right" units, you didn't run into the bug.

chummels:
Oh I see, that makes sense...  Well, either way--great job!

PR #1944, opened 2018-08-06
[WIP] Adding Cholla Frontend

chummels:
## PR Summary

This adds a frontend to yt for working with datasets from the CHOLLA code (https://github.com/cholla-hydro/cholla).  It's currently a work-in-progress, so do not merge it yet until I've removed that header and this text.

Currently, it can read in the dataset and give the field list ok, but it fails with an "endian" error when you actually try to look at some of the data.  This is a strange error, since all of the datasets are HDF5 files, which shouldn't yield issues with endianness.

For simplicity, I've also made it so nprocs=1, so it reads a single grid for the whole dataset, since it's unigrid.  After I get this working, I'll make yt chunk it into smaller-sized chunks for for efficiency.

PR #2120, opened 2018-12-11
ensure gadget binary field data are in native endianness

ngoldbaum:
Some low-level routines (in the test case in the bug report it was the selection machinery) expect raw field data from the low-level I/O routines to be in native byte order. However, here we're reading bigendian data and pass it directly to those low-level routines. This breaks on little endian architectures.

The fix is to ensure the data are in native byte order after we read it in.

Ping @weiguangcui and maybe @MatthewTurk to look this over?


chummels:
@ngoldbaum Whoa, this is great!  Similar code may also be able to address the endian issues I described in PR #1944 ?  If so, you just saved me a ton of time...

PR #2121, opened 2018-12-13
[bug fix] allow prj.annotate_marker to add multiple markers at once

sheagk:
`sanitize_coord_system` takes in a "set of x,y (and z) coordinates"  and converts them to be ready for plotting.  However, passing in len(N) arrays for the x,y,z coordinates would only return the sanitized version of the first entry in that array.  This allows you to do, e.g., `prj.annotate_marker((xlist, ylist, zlist))` to add multiple markers at once.

## PR Checklist

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -220,7 +220,7 @@ class PlotCallback(object):
             coord = self.convert_to_plot(plot, coord)
             # Convert coordinates from a tuple of ndarray to a tuple of floats
             # since not all callbacks are OK with ndarrays as coords (eg arrow)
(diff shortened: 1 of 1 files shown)


ngoldbaum:
It looks like a couple of the tests have failed. Let me know if you need help parsing the output from the test runners. I didn't look closely at all the test platforms, but the two tests that failed on jenkins were `yt.visualization.tests.test_callbacks.test_arrow_callback` and `annotations.py` in the cookbook.

chummels:
I can help with these.

ngoldbaum:
It would also be good to get modifications to the tests for the marker callback (if there are any?) to test that the new feature you've added works correctly.

Issue #1839, opened 2018-06-19
Input validation for plot callbacks


ngoldbaum:
### Bug report

**Summary**

Many plot callbacks do not do proper input validation. This leads to confusing error messages or silently incorrect results. We should make use of  the input validation helper functions added in #1833 to the plot callbacks as well as the data objects.

Below is an example from a user bug report. It took me about 10 minutes of staring at the code to figure out where the error was coming from.

**Code for reproduction**

```python
import yt

ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")

val, loc = ds.find_max('density')
plt = yt.SlicePlot(ds, 'z', 'density)
plt.annotate_particles(ds, 10000*pc)
plt.save("test.png")
```

**Actual outcome**

```python traceback
Traceback (most recent call last):
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/visualization/plot_window.py", line 1001, in run_callbacks
    callback(cbw)
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/visualization/plot_modifications.py", line 64, in _check_geometry
    return func(self, plot)
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/visualization/plot_modifications.py", line 1619, in __call__
    reg = self._get_region((x0,x1), (y0,y1), plot.data.axis, data)
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/visualization/plot_modifications.py", line 1680, in _get_region
    LE[zax] = data.center[zax] - self.width*0.5
TypeError: unsupported operand type(s) for *: 'EnzoDataset' and 'float'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "test.py", line 14, in <module>
    plt.save("test.png")
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/visualization/plot_container.py", line 95, in newfunc
    args[0]._setup_plots()
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/visualization/plot_window.py", line 950, in _setup_plots
    self.run_callbacks()
  File "/Users/goldbaum/Documents/yt-git-fixes/yt/visualization/plot_window.py", line 1007, in run_callbacks
    sys.exc_info()[2])
  File "/usr/local/lib/python3.6/site-packages/six.py", line 692, in reraise
    raise value.with_traceback(tb)
[...]

PR #2122, opened 2018-12-14
Make some annotations work with multiple points

chummels:
## PR Summary

Enables annotate_marker() and annotate_arrow() to work with multiple points.

Code can now be specified as:

```
import yt
p = yt.OffAxisProjectionPlot(ds, [1,0,0], 'temperature', weight_field='density', north_vector=[0,1,0])
p.annotate_arrow([0.6, 0.6, 0.6], coord_system='data')
p.annotate_arrow([[0.6, 0.7], [0.6, 0.7], [0.6, 0.7]], coord_system='data')
p.annotate_arrow([[0.6, 0.7, 0.8], [0.6, 0.7, 0.8], [0.6, 0.7, 0.8]], coord_system='data')
p.annotate_arrow([0.5, 0.5], coord_system='plot')
p.annotate_arrow([[0.5, 0.4], [0.5, 0.4]], coord_system='plot')
p.annotate_arrow([[0.5, 0.4, 0.3], [0.5, 0.4, 0.3]], coord_system='plot')
p.save()
```

## PR Checklist

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -105,7 +105,7 @@ class PlotCallback(object):
         code_length units.  Projected data units are 2D versions of the
         simulation data units relative to the axes of the final plot.
         """
-        if len(coord) == 3:
+        if coord.shape[0] == 3:
             if not isinstance(coord, YTArray):
                 coord = plot.data.ds.arr(coord, 'code_length')
             coord.convert_to_units('code_length')
@@ -121,7 +121,9 @@ class PlotCallback(object):
             # we have to calculate where the data coords fall in the projected
             # plane
             elif ax == 4:
-                coord_vectors = coord - plot.data.center
+                # transpose is just to get [[x1,x2,...],[y1,y2,...],[z1,z2,...]]
+                # in the same order as plot.data.center for array arithmetic
+                coord_vectors = coord.transpose() - plot.data.center
                 x = np.dot(coord_vectors, plot.data.orienter.unit_vectors[1])
                 y = np.dot(coord_vectors, plot.data.orienter.unit_vectors[0])
                 # Transpose into image coords. Due to VR being not a
@@ -181,8 +183,9 @@ class PlotCallback(object):
 
     def sanitize_coord_system(self, plot, coord, coord_system):
         """
-        Given a set of x,y (and z) coordinates and a coordinate system,
-        convert the coordinates (and transformation) ready for final plotting.
+        Given a set of one or more x,y (and z) coordinates and a coordinate 
+        system, convert the coordinates (and transformation) ready for final 
+        plotting.
 
         Parameters
         ----------
@@ -191,7 +194,9 @@ class PlotCallback(object):
            The plot that we are converting coordinates for
 
         coord: array-like
-           Coordinates in some coordinate system.
+           Coordinates in some coordinate system: [x,y,z].
+           Alternatively, can specify multiple coordinates as:
+           [[x1,x2,...,xn], [y1, y2,...,yn], [z1,z2,...,zn]]
 
         coord_system: string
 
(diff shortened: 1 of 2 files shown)


ngoldbaum:
Seems sensible to me! Thanks for updating the docs and tests.

chummels:
@sheagk do you want to test this to assure it works for your use case?

yt/visualization/plot_modifications.py
```diff
@@ -105,7 +105,7 @@ def project_coords(self, plot, coord):
         code_length units.  Projected data units are 2D versions of the
         simulation data units relative to the axes of the final plot.
         """
-        if len(coord) == 3:
+        if coord.shape[0] == 3:
```
MatthewTurk: Is `coord` always going to be an array?  Is it necessary to do something weird like: `if getattr(coord, 'shape', [len(coord)])[0]` to make it work?

yt/visualization/plot_modifications.py
```diff
@@ -105,7 +105,7 @@ def project_coords(self, plot, coord):
...
             if not isinstance(coord, YTArray):
```
MatthewTurk: We could maybe put this outside the conditional, which would fix my question above...

yt/visualization/plot_modifications.py
cphyc: Isn't the version with `len` a bit more generic than the one with shape, no? Both should return the same value for arrays, but the `len` will work with lists, tuples, ...

ngoldbaum: Just looking at a few places this function is used, it looks like it is exposed to user-supplied data (for example, in the `LinePlotCallback`, that was just the first place I looked, there may be others).

We should probably be making sure all user-supplied data are converted to numpy arrays, in which case the change you've made here would be fine, but that's a much bigger job than just this PR (see https://github.com/yt-project/yt/issues/1839 for more detail about this). If `len` is sufficient, I'd just leave this as-is. If for some reason this PR needs that to change then Matt's suggestion would probably also work.

chummels: The only place in all of yt that `project_coords()` is called is in `sanitize_coord_system()` and I explicitly changed the first line in that function to assure that coord is a numpy array (see below), so there is never any danger that project_coords gets a non-numpy array object.  

To answer @ngoldbaum 's suggestions, `sanitize_coord_system()` is the first function called in all of the callbacks, and its first line is to convert the coordinates to a numpy array.  So it basically covers all of the problem cases.

When I originally wrote this function, `project_coords()`, I should've made it private, but I didn't, and it seemed reckless to make it private now, but maybe that's appropriate?

yt/visualization/plot_modifications.py
```diff
@@ -105,7 +105,7 @@ def project_coords(self, plot, coord):
...
                 coord = plot.data.ds.arr(coord, 'code_length')
```
ngoldbaum: Given the change to pass coord through `np.array` is this block of the if statement ever executed?

yt/visualization/plot_modifications.py
chummels: OK, I went and made those internal methods private.

yt/visualization/plot_modifications.py
chummels: Good catch.  This actually helped me catch a corner case because I incorrectly assumed that when you wrap a YTArray with a numpy array, that it retains its YTArray-ness.  It does not.  So now I make a check at the beginning of `sanitize_coord_system()` to assure things are either a YTArray or numpy array, but not to just clobber the YTArray if it already exists.  Thanks for the note!

yt/visualization/plot_modifications.py
```diff
@@ -211,16 +216,15 @@ def sanitize_coord_system(self, plot, coord, coord_system):
                 2D coordinates within figure object from (0,0) in lower left
                 to (1,1) in upper right.  Same as matplotlib figure coords.
         """
+        # Assure coords are either a YTArray or numpy array
```
chummels: Thanks to @ngoldbaum for the suggestion on this numpy function of which I was unaware, which preserves YTArrays or wraps as a numpy array.

yt/visualization/plot_modifications.py
chummels: The concern here was that some arrays *shouldn't* be `YTArray` objects for dealing with plot, figure, or axis coord systems.  But this should now be addressed in the code.  The switch to `coord.shape[0]` from `len(coord)` was residual from a previous iteration at trying to fix this issue.  Either should be fine here because I'm sanitizing the inputs, but I swapped it back to `len(coord)` just to minimize code changes and make things simpler.

PR #2138, opened 2019-01-24
Fixing minor issues with some binary-format Gadget

MatthewTurk:
Using some data from @jcoughlin11 , I found that there were some things for binary data that had not been updated in yt-4.0.  This is just a set of two minor fixes for that that bring it in line -- the first is calling the right header validation and the second is to only remove `_sph_ptype` if it's actually used.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -574,7 +574,7 @@ class Dataset(object):
         if "nbody" not in self.particle_types:
             mylog.debug("Creating Particle Union 'nbody'")
             ptypes = list(self.particle_types_raw)
-            if hasattr(self, '_sph_ptype'):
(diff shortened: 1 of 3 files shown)


ngoldbaum:
Looks like there’s a flake8 error that needs to be fixed.

MatthewTurk:
I've identified a correction to a previous fix!  And no, we don't need it in master because of how we do things slightly differently wrt data files there.  I think this change is explicitly for yt-4.0 and the `GadgetBinaryFile` object.

This now loads -- and in fact removes a bunch of (correct) errors that the IO handler spit out, without disabling those errors.

ngoldbaum:
Awesome, thanks Matt! Will merge when the tests pass.

PR #2139, opened 2019-01-24
Replace deprecated h5py syntax

brittonsmith:
As of h5py 2.9.0, reading a dataset via `dataset.value` has been deprecated in favor of `dataset[()]`. This updates all `.value` calls to get rid of the deprecation warning. This syntax is supported in h5py 2.8.0, which I believe is what we have listed for the minimum h5py version.

### DIFF
--- a/doc/source/cookbook/extract_fixed_resolution_data.py
+++ b/doc/source/cookbook/extract_fixed_resolution_data.py
@@ -30,4 +30,4 @@ f.close()
 
 # If we want to then access this datacube in the h5 file, we can now...
 f = h5py.File("my_data.h5", "r")
-print(f["density"].value)
+print(f["density"][()])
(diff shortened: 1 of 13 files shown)


ngoldbaum:
Looks like we require 2.4.0 or newer.

Can you check when that syntax was added?

Could bump the minimim h5py version but 2.4.0 is old enough that it might catch some people by surprise so might want to make sure the error message you get is good.

ngoldbaum:
https://github.com/yt-project/yt/blob/master/yt/utilities/on_demand_imports.py#L293

brittonsmith:
Oh, thanks for pointing that out. I was going off of the version I found for h5py in `tests/test_requirements.txt`. I was able to test h5py 2.4.0 with Python 2.7 and the `[()]` syntax is supported. Interestingly, I get import errors with h5py 2.4.0 on Python 3.5 and later, so that version may not even be supported by Python 3 at all. From my testing, h5py 2.5.0 was the earliest major release that I could get to work with Python 3.

ngoldbaum:
The versions in `test_requirements.txt` are just the versions that get installed on the test machine, not the minimim supported version.

If that syntax is supported in sufficiently old versions of `h5py` then great, thanks very much for checking.

PR #2145, opened 2019-02-07
Gadget read

jcoughlin11:
## PR Summary

Currently, for a multi-file Gadget snapshot, yt requires that the first file be passed, e.g. snapshot_000.0, to yt.load(). This is a straightforward fix for that. It allows multi-file snapshots to be passed by the base name of the snapshot. The only thing I'm not sure of is how other data formats handle multi-file outputs, so it might be better to delegate this to a separate function that can have a different case for each code, rather than simply putting it directly in load().

## PR Checklist

- [ x] Code passes flake8 checker

### DIFF
--- a/yt/convenience.py
+++ b/yt/convenience.py
@@ -50,11 +50,6 @@ def load(*args ,**kwargs):
                 if os.path.exists(os.path.join(ytcfg.get("yt", "test_data_dir"), arg)):
                     valid_file.append(True)
                     args[argno] = os.path.join(ytcfg.get("yt", "test_data_dir"), arg)
-                # Check for multi-file gadget snapshot that's been passed without the
-                # '.0' extension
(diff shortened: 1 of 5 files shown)


ngoldbaum:
This logic should probably be in `GadgetBinaryDataset.__init__`, since it’s particular to the Gadget binary data format. You might also need to modify the `_is_valid` implementation as well in that frontend. I’m happy to talk through this in person if you need it.

ngoldbaum:
@qobilidop do you see any possible issues with this?

@jcoughlin11 it would be good to get tests for this in the test suite. It would also be good to update the documentation on loading gadget data (e.g. this note: https://github.com/yt-project/yt/blob/yt-4.0/doc/source/examining/loading_data.rst#L1057).

qobilidop:
LGTM

PR #2148, opened 2019-02-11
address github vulnerability warning

ngoldbaum:
see https://github.com/yt-project/yt/network/alert/tests/test_requirements.txt/pyyaml/open

I don't think Britton could have known about this ahead of time...


brittonsmith:
Ooh, good catch! Thanks, Nathan!

PR #2149, opened 2019-02-14
Adding dynamic parallelism to DatasetSeries piter

jwise77:
## PR Summary

The DatasetSeries piter() did not have the dynamic option. I've added it in this PR.

## PR Checklist

### DIFF
--- a/yt/data_objects/time_series.py
+++ b/yt/data_objects/time_series.py
@@ -43,7 +43,7 @@ from yt.utilities.exceptions import \
     YTException, \
     YTOutputNotIdentified
 from yt.utilities.parallel_tools.parallel_analysis_interface \
-    import parallel_objects, parallel_root_only
+    import parallel_objects, parallel_root_only, communication_system
 from yt.utilities.parameter_file_storage import \
     simulation_time_series_registry
      
@@ -209,7 +209,7 @@ class DatasetSeries(object):
     def outputs(self):
         return self._pre_outputs
 
-    def piter(self, storage = None):
+    def piter(self, storage = None, dynamic = False):
         r"""Iterate over time series components in parallel.
 
         This allows you to iterate over a time series while dispatching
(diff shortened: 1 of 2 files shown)


yt/data_objects/time_series.py
```diff
@@ -274,14 +274,22 @@ def piter(self, storage = None):
...
                 njobs = self.parallel
+        else:
+            my_communicator = communication_system.communicators[-1]
+            nsize = my_communicator.size
+            if nsize == 1:
+                self.parallel = False
+                dynamic = False
+                njobs = 1
+            else:
+                njobs = nsize - 1
```
brittonsmith: @jwise77, is it necessary to set `njobs` this way here? This also happens inside `dynamic_parallel_objects`, which is the function called by `parallel_objects` when `dynamic=True`. This might be removing one too many jobs.

brittonsmith:
@chummels, this would parallelize iterating over a dataset series using a task queue. I agree though, it would be good to add this to the `piter` docstring.

ngoldbaum:
Agreed that this needs a mention int he docstring at least. If you have time it might also be worth mentioning in the narrative docs on parallelism: http://yt-project.org/doc/analyzing/parallel_computation.html

ngoldbaum:
ping @jwise77 any chance we can get a docstring update?

jwise77:
It'll be early next week when I can get to it. I was at a conference this week.

yt/data_objects/time_series.py
jwise77: I looked at this, and `dynamic_parallel_objects` only removes one process if `njobs` isn't set. But since we're passing in `njobs`, we have to subtract 1. I blindly tried to remove the "-1" and I got this error message,
`P001 yt : [ERROR    ] 2019-02-26 09:38:05,303 You have asked for 4 jobs, but only 3 processors are available.`
So I'm keeping this -1 in there.

jwise77:
> Agreed that this needs a mention in the docstring at least. If you have time it might also be worth mentioning in the narrative docs on parallelism: http://yt-project.org/doc/analyzing/parallel_computation.html

This is a good idea, and I'll add to the narrative docs since I didn't find any description of dynamic load balancing anywhere else.

jwise77:
Okay, I've added the `dynamic` docstring and narrative docs. Let me know if there's anything else that needs addressing.

PR #2153, opened 2019-02-22
Backports for yt 3.5.1

ngoldbaum:
This backports a number of pull requests to the stable branch so we can do a 3.5.1 release.


ngoldbaum:
The test failures on travis are unrelated to the backports and can be ignored. Fixing them would require a bunch of additional work with no marginal benefit since all the issues are already fixed on master.

PR #2159, opened 2019-02-26
add support for yt_astro_analysis in the halo plot modification

ngoldbaum:
Dunno if there are other places in yt where we need to do the same sort of wrapping around yt_astro_analysis which may or may not be installed.


ngoldbaum:
Merge conflicts are fine.

PR #2160, opened 2019-02-26
add clarification to release docs on multibuild, tagging, scp

munkm:
This adds some clarifications in the release docs that I thought would be helpful. 
* explicitly lists the remote ssh address to copy the tarball, rather than assuming a ssh config is set up
* switches the ordering of updating the conda-forge and building wheels sections so that the wheels section comes right before uploading to PyPI. 
* while tagging is specific to a commit hash, it's probably a good practice tag from the stable branch so a clarification has been added about that. 
* updates the notes about being an owner or maintainer of yt on PyPI

### DIFF
--- a/doc/source/developing/releasing.rst
+++ b/doc/source/developing/releasing.rst
@@ -216,10 +216,10 @@ You will be prompted for your PyPI credentials and then the package should
 upload. Note that for this to complete successfully, you will need an account on
 PyPI and that account will need to be registered as an "owner" or "maintainer" 
 of the yt package. 
-Right now there are five owners: Matt Turk, Britton Smith, Nathan
-Goldbaum, John ZuHone, and Kacper Kowalik. In addition, you should attempt to
-upload the yt package along with compiled binary wheel packages for various
-platforms that we support.
+Right now the following people have access to upload packages: Matt Turk, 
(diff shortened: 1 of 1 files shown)


doc/source/developing/releasing.rst
```diff
@@ -201,12 +209,14 @@ issue the following command:
...
+the ``dist`` folder before doing this. Directions on generating binary wheels 
+are described in the section immediately preceding this one. 
 
 You will be prompted for your PyPI credentials and then the package should
 upload. Note that for this to complete successfully, you will need an account on
-PyPI and that account will need to be registered as an "owner" of the yt
-package. Right now there are five owners: Matt Turk, Britton Smith, Nathan
+PyPI and that account will need to be registered as an "owner" or "maintainer" 
+of the yt package. 
+Right now there are five owners: Matt Turk, Britton Smith, Nathan
```
ngoldbaum: you can add yourself too now

munkm: I got a notice from PyPI that I was a maintainer (not an owner) so I'm not sure I belong in this list.

ngoldbaum: Maintainers can upload releases too and not everyone on this list is actually an owner. I'd just say "right now the following people have access to upload packages:"

munkm: Sounds good! I've fixed that now.

PR #2161, opened 2019-02-27
Revert "Gadget read"

MatthewTurk:
Reverts yt-project/yt#2145

This was merged before it was ready.


ngoldbaum:
We don't need to revert the PR since it's an improvement, I think we just need to remember to document it and add tests. @jcoughlin11 would you mind opening an issue to track that if you can't take care of it immediately? I can walk you through how to do that if you need a hand.

PR #2163, opened 2019-02-27
Remove most py2 compat code

matthewturk:
This removes most of the python2 compatibility code.  There are a
handful of things that we no longer need to do, specifically lots of
`class Something(object):` invocations.

Unit tests pass; I anticipate there may be issues in answer testing
related to mistakes in importing or decode/encode of strings, and I look
forward to fixing those.


ngoldbaum:
One thing that I'm a little nervous about on the yt-4.0 branch, especially for a big invasive change like this, is that we don't have great test coverage for the frontends because travis and appveyor don't run tests with real data.

It would be great if we could get Jenkins running against the yt-4.0 branch next week.

MatthewTurk:
OK, so this is a non-starter until then?

ngoldbaum:
Make sure `python_requires` in `setup.py` gets updated, it can now just be `>=3.5` I think.

ngoldbaum:
Getting this working would definitely be worthwhile since you'll have to do this annoying mechanical work anyway, there just might be more brokenness in the frontends or in code that only has tests that uses real data that you won't notice until jenkins gets turned on so this probably can't be merged until then.

MatthewTurk:
To close the loop, this is dependent on #2208 and #2172 , and then (pending those tests passing) I believe it is ready to go.

munkm:
I think this is now also dependent on yt-project/yt#2260

jzuhone:
LGTM!

setup.py
```diff
@@ -349,10 +349,10 @@ def run(self):
                  "Operating System :: POSIX :: Linux",
                  "Programming Language :: C",
                  "Programming Language :: Python :: 2",
```
chummels: Don't we want to get rid of this line too? No Py2 coverage?

matthewturk:
OK!  I have merged, and I think this is ready-for-review again.

PR #2167, opened 2019-03-04
[yt-4.0 FIXME] Update docs and code comments on loading SPH data in stream frontend.

brittonsmith:
This clears one of the yt-4.0 FIXMEs. I've added additional code comments and docstrings relating to loading and identifying SPH data in the stream frontend.

### DIFF
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -1087,10 +1087,9 @@ class StreamParticlesDataset(StreamDataset):
             stream_handler, storage_filename=storage_filename,
             geometry=geometry, unit_system=unit_system)
         fields = list(stream_handler.fields['stream_file'].keys())
+        # This is the current method of detecting SPH data.
+        # This should be made more flexible in the future.
         if ('io', 'density') in fields and ('io', 'smoothing_length') in fields:
-            # FIXME FIXME FIXME don't merge PR with this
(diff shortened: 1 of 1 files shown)


yt/frontends/stream/data_structures.py
```diff
@@ -1101,15 +1100,18 @@ def load_particles(data, length_unit = None, bbox=None,
...
-    * There must be sufficient space in memory to contain both the particle
-      data and the octree used to index the particles.
+    * There must be sufficient space in memory to contain all the particle
+      data.
     * Parallelism will be disappointing or non-existent in most cases.
+    * Fluid fields are not supported.
 
-    This will initialize an Octree of data.  Note that fluid fields will not
-    work yet, or possibly ever.
+    Note: in order for the dataset to take advantage of SPH functionality,
```
chummels: I'm confused.  The sample dataset FIRE_M12I_ref11 does not contain these two fields, and yet it appears to work with SPH functionality.  Am I missing something here?  Must they be in `io` field type?  Because that dataset has both `('gas', 'density')` and `('gas', 'smoothing_length')`.

ngoldbaum: This is specifically for data loaded via `yt.load_particles`

chummels: Oh! Now I see this is contained in the docstring for that method.  My bad.  LGTM.

PR #2168, opened 2019-03-04
vendor cykdtree

ngoldbaum:
This adds the serial kdtree from cykdtree to yt proper. This sidesteps a number of packaging headaches associated with the parallel kdtree in cykdtree as well as the headaches of yt depending on a new external package that would need to have wheels distributed for it.


ngoldbaum:
@chummels any chance you can try applying this pull request locally and let me know if it gives you trouble? Once this PR is merged there will be no need to separately install cykdtree. 

One issue is that I think you might need to completely rebuild yt with a `git clean -fxd` once this PR gets merged. It would be great if you could verify whether that is indeed the case.

chummels:
I will test it out today!  Excited!  Thanks!

ngoldbaum:
I just tested by switching to the `yt-4.0` branch, compiling, then switching back to the branch for this PR, and I didn't have to rebuild yt from scratch so I think this is fine actually.

chummels:
Testing now.  Will let you know in ~15 min.

chummels:
Yeah, it works here.  I removed cykdtree from my system, pulled this new branch, compiled and ran some tests (including some with the sph-viz branch of trident), and it all works.  Awesome!  Thanks for getting this going!

ngoldbaum:
It turns out that we needed to bump our minimum required cython version to get cython to recognize the `.hpp` headers in cykdtree.

PR #2172, opened 2019-03-05
[WIP] merge yt-4 with master

ngoldbaum:
This tracks merging with yt-4 branch with master. This is not going to be useful for review but I'm opening now so I can work on getting the tests passing.

We will have a discussion on the mailing list about merging this.


ngoldbaum:
The py2 build failures can be ignored because master won't support python2 anymore once this PR is merged.

MatthewTurk:
I've updated the merge branch with most recent changes on both master and yt-4.0.  The vector operations were a bit tricky to merge, but I think I got it.

@ngoldbaum I'm not sure if we can edit the pull request, but if the messages get too annoying we can switch it to being from my fork/branch.

ngoldbaum:
No worries, let me know if you need a hand with anything.

MatthewTurk:
Sneaky `proj` and `quad_proj` change now fixed for the ytdata tests!

MatthewTurk:
@qobilidop Not sure if you're the right person to ask, but you might have an idea.  The answer tests for tipsy are failing, and I'm not sure why they were ever *passing* before, but you might know why!

Right now the Tipsy tests run:

```
for test in sph_answer(ds, 'galaxy.00300', 315372, tg_fields):
    ...
    yield test
```

We're testing these fields:

```
tg_fields = OrderedDict(                                                                                                                      
    [                                                                                                                                         
        (('gas', 'density'), None),                                                                                                           
        (('gas', 'temperature'), None),                                                                                                       
        (('gas', 'temperature'), ('gas', 'density')),                                                                                         
        (('gas', 'velocity_magnitude'), None),                                                                                                
        (('gas', 'Fe_fraction'), None),                                                                                                       
        (('Stars', 'Metals'), None),                                                                                                          
    ]                                                                                                                                         
)  
```

The thing is, it seems to me that this should not actually *ever* pass -- the only fields that are of that particular size are the ones that are the `all` particle union.

So I guess I'm not sure why this ever worked?  But more to the point, I wonder if you have any insight into why the unions and whatnot might give different results now.

MatthewTurk:
Never mind, I figured it out!  We now have an "nbody" type.  I'll address this.

jzuhone:
@MatthewTurk keep in mind that PR #2208 needs to be merged before a lot of these tests even work?

MatthewTurk:
@jzuhone well crumbs, now I feel extra silly.

matthewturk:
I'm trying to keep this up to date.  The place where I'm not entirely sure I made the right decisions in merging is in the `vector_operations.py` file, which was last touched by @jisuoqing .  I believe I correctly merged the results from #2136 .

matthewturk:
The remaining errors I see when I run locally are all related to `over_refine_factor`.  @AshKelly do you think this is something worth preserving, in light of your (open) PR to add back in some more stuff to the octree (#2284 )?  I am tempted to disable the tests.  (cc @brittonsmith )

matthewturk:
I pasted the errors here, btw: https://gist.github.com/118bf0bffe8929e6ae3b7d85fcf1b438

AshKelly:
Sorry, I'm out at the moment so on mobile. These errors don't look like octree errors... They look like data loading errors? Maybe I'm misunderstanding

ngoldbaum:
Instead of disabling those tests I think it would be better to remove the tests and add error messages that users will see if they pass in the keyword argument telling them why that keyword no longer makes sense.

matthewturk:
@AshKelly Oh, they're kind of both!  They are for a parameter that I think no longer shows up, which is for over refinement.  I wasn't sure if that was in your plans (or if I've missed it) for octree generation.

ngoldbaum:
So since there’s no longer an in-memory octree for indexing particle data I don’t think it makes sense for `over_refine_factor` to be a keyword argument for `load_particles`. Since it used to be available in the API, we need to be careful about explaining to users why it’s no longer there lest they run into the error in Matt’s gist.

matthewturk:
@ngoldbaum I agree with that, which is why I was proposing disabling the tests unless the keyword was (or was going to be) in the octree construction, in which case moving them to test that would work.

brittonsmith:
If the octree is coming back eventually, I'd rather see tests related to that disabled (instead of being totally removed) so they are easier to bring back when the time comes.

ngoldbaum:
It doesn’t make sense to use `over_refine_factor` with `load_particles` anymore. That’s why I was suggesting removing the tests usage of that keyword argument.  I don’t think it makes sense to disable the tests and leave in them in the codebase because it won’t make sense to pass the keyword to `load_particles` unless we brought back the octree indexing system.

The octree data object Ash created during his GSOC project does support `over_refine_factor`:

https://github.com/yt-project/yt/blob/yt-4.0/yt/data_objects/construction_data_containers.py#L2139

if you add a user visible error or warning if someone passes `over_refine_factor` to `load_particles` it message could mention `ds.octree` after saying that an in-memory octree is no longer created every time you load a particle dataset

matthewturk:
If the tests have already been replicated in the GSOC code, then we can just remove them -- but if they need to be replicated (i.e., to test for that functionality) then I'll move them over.

@ngoldbaum I do also agree that the over refine factor should emit an error of some type; I did not mean to suggest it should not, but I didn't think that was the full solution to what to do with the tests.

matthewturk:
I'll investigate the testing cover of oref in the GSOC code in the next two days.  (cc @AshKelly )

matthewturk:
I ended up removing that test because it doesn't have meaning anymore with particle datasets being "non-indexed" now.

matthewturk:
Looks like at least a firefly test error.

matthewturk:
My reading is that all of the tests that are now failing are answer tests, not unit tests.  Can someone else verify this?  @brittonsmith can you double check me, in the output from travis/fido?

matthewturk:
My plan for addressing answer testing differences, which I will start by doing manually:

 * All grid and octree data should have either the *same*, or *justifiably different* answers.  In my view, "justifiably different" includes unit changes, and also includes changes as a result of particles being selected on edges of grid boundaries.  These will be documented.  (See #2195 )
 * Results for all data selections in SPH and particle datasets should be identical in count and values, although the order will likely be different.  To address this, we will have a "sorting order" field that is the morton index of the particles, and the values will be sorted and compared.  This should help to identify situations where different numbers of particles are being selected (typically they should not be, except for situations related to smoothing length.)
 * Manual inspection of visualizations that avoid meshing in 4.0 and that utilize meshing in 3.x.

I believe this should cover most of the cases, and will take to the next step of verification.  To conduct these tests, I'm going to work on a script that outputs the appropriate values into an HDF5 file, and then compare the results for both.  This will be somewhat distinct from the answer testing and designed for ease of exploration.

brittonsmith:
@matthewturk, I concur that all of the failing tests are answer tests. I think you will also find some failures due to switching to `unyt`, which has slightly different physical constants than `yt.units`. This was my experience with Trident.

jzuhone:
@matthewturk do you need help with addressing the three test areas above you mentioned? should we start parceling some of this stuff out? I think we should make getting this PR through a big priority.

matthewturk:
I would *love* help with the tests.

matthewturk:
Current status:

 * The commit before merging from master and yt-4.0 passes all unit tests on my machine
 * I merged from master
 * I merged from yt-4.0

Right now I think this is ready for folks to take a look at.  I may poke at it a bit if I see things having issues.

matthewturk:
I'd note that from what I can see, the tests that don't pass are all expected *not* to pass -- answer tests, py2, etc.

matthewturk:
So #2208 has been merged, and I am going to poke more at this.

matthewturk:
I'm closing and re-issuing from my repository!

PR #2176, opened 2019-03-05
netCDF Front End for Cloud Model Simulations

keltonhalbert:
## PR Summary
Added a front-end for reading CF compliant netCDF files from atmospheric cloud model simulations from the NCAR CM1 model. It uses xarray for disk io, but considering switching to the netCDF4 API due to extra/unnecessary dependencies required through xarray. This is in part to a  dependency in order to import xarray within a try/except block. 

Right now this will only work for netCDF files generated from CM1, as it depends on following the specific axis naming conventions. Ideally, this should be generalized such that any CF compliant netCDF can be read. 

So far it seems like it handles the reading properly and plots - however, I am unsure of what tests should be conducted to make sure there aren't any issues hiding.

yt-image


Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+                    ncoords = len(vcoords)  # number of coordinates in variable
+                    coordspassed = 0  # number of coordinates that pass for a variable
+                    for vc in vcoords:  # iterate over the coordinates for the variable
+                        if vc in coords:
+                            # variable coordinate and global coordinate are same
+                            coordspassed += 1
+                    if coordspassed != ncoords:
+                        failed_vars.append(var)
+
+                if len(failed_vars) > 0:
```
cphyc, line 209: It is a matter of taste, but I would rather show a warning than an error here. I'm fine if you think it should be an error.

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+            # loop over the variable names in the netCDF file, record only those on the
+            # "zh","yh","xh" grid.
+            varnames = []
+            for key in ds.variables.keys():
+                vardims = ds.variables[key].dimensions
+                if all(x in vardims for x in ["time", "zh", "yh", "xh"]) is True:
+                    varnames.append(key)
+            self.parameters["variable_names"] = varnames
+            self.parameters["lofs_version"] = ds.cm1_lofs_version
+            self.parameters["is_uniform"] = ds.uniform_mesh
```
neutrinoceros, line 160: This feels like it's moving backwards. `self.parameters` should contain all the meta data that directly from the parameter/data file, and be used to populate the dataset attributes, not the other way around.
Besides, I don't understand when `ds.cm1_lofs_version` is defined, but it should probably not be a class attribute.

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+        # will be converted to YTArray automatically at a later time.
+        # This includes the cosmological parameters.
+        #
+        #   self.unique_identifier      <= unique identifier for the dataset
+        #                                  being read (e.g., UUID or ST_CTIME)
+        self.unique_identifier = int(os.stat(self.parameter_filename)[stat.ST_CTIME])
+        self.parameters = {}  # code-specific items
+        with self._handle.open_ds() as ds:
+            dims = [ds.dimensions[i].size for i in ["xh", "yh", "zh"]]
+            coords = {i: ds.variables[i][:] for i in ["xh", "yh", "zh"]}
```
neutrinoceros, line 138: looks like this isn't used before line 169. I suggest merging those

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+            for dim, meta in ds.dimensions.items():
+                dim_info[dim] = meta.size
+            self.parameters["dimensions"] = dim_info
+
+        self.parameters["coords"] = coords
+
+        self.dimensionality = 3
+        self.domain_dimensions = np.array(dims, dtype="int64")
+        self.periodicity = (False, False, False)
+        self.parameters["time"] = self.current_time
```
neutrinoceros, line 174: same as above

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # Set cosmological information to zero for non-cosmological.
+        self.cosmological_simulation = 0.0
```
neutrinoceros, line 177: This one should really be a boolean, but for now it's an int !
```suggestion
        self.cosmological_simulation = 0
```

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+    @classmethod
+    def _is_valid(self, *args, **kwargs):
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+
+        warn_netcdf(args[0])
+        try:
+            from netCDF4 import Dataset
+
+            with Dataset(args[0], "r", keepweakref=True) as ds:
```
neutrinoceros, line 192: I suggest getting some inspiration in #2739 an use a named argument "filename" instead of `args[0]`. Otherwise this is the current standard and can be merged as is.

yt/frontends/nc4_cm1/io.py
```diff
@@ -0,0 +1,62 @@
...
+        # ptypes and the values are lists of fields.
+        pass
+
+    def _read_particle_fields(self, chunks, ptf, selector):
+        # This gets called after the arrays have been allocated.  It needs to
+        # yield ((ptype, field), data) where data is the masked results of
+        # reading ptype, field and applying the selector to the data read in.
+        # Selector objects have a .select_points(x,y,z) that returns a mask, so
+        # you need to do your masking here.
+        pass
```
cphyc, line 28: To other reviewers: should these functions be:

- like they currently are
- raise a NotImplementedError
- be removed

Also, this goes well beyond the scope of this PR, but is `_read_particle_coords` still a thing? It seems to be only used for SPH-specific datasets.

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # Set cosmological information to zero for non-cosmological.
+        self.cosmological_simulation = 0.0
+        self.current_redshift = 0.0
+        self.omega_lambda = 0.0
+        self.omega_matter = 0.0
+        self.hubble_constant = 0.0
+
+    @classmethod
+    def _is_valid(self, *args, **kwargs):
```
neutrinoceros, line 184: This is a class method
```suggestion
    def _is_valid(cls, *args, **kwargs):
```

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+
+        warn_netcdf(args[0])
+        try:
+            from netCDF4 import Dataset
+
+            with Dataset(args[0], "r", keepweakref=True) as ds:
+                is_cm1_lofs = hasattr(ds, "cm1_lofs_version")
+                is_cm1 = hasattr(ds, "cm1 version")
```
neutrinoceros, line 194: I'm guessing we're missing an underscore here ?
```suggestion
                is_cm1 = hasattr(ds, "cm1_version")
```

cphyc (approved):
Overall this looks good to me! Congrats ;)

I left some minor comments that should not delay much further the merge (of course provided the tests pass).

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+                coords = ds.dimensions  # get the dataset wide coordinates
+                failed_vars = []  # list of failed variables
+                for var in ds.variables:  # iterate over the variables
+                    vcoords = ds[var].dimensions  # get the dimensions for the variable
+                    ncoords = len(vcoords)  # number of coordinates in variable
+                    coordspassed = 0  # number of coordinates that pass for a variable
+                    for vc in vcoords:  # iterate over the coordinates for the variable
+                        if vc in coords:
+                            # variable coordinate and global coordinate are same
+                            coordspassed += 1
```
neutrinoceros, line 205:
```suggestion
                    # number of coordinates that pass for a variable
                    coordspassed = sum(vc in coords for vc in vcoords)
```

yt/frontends/nc4_cm1/data_structures.py
neutrinoceros, line 209: python for the win !
```suggestion
                if failed_vars:
```

yt/frontends/nc4_cm1/tests/test_outputs.py
```diff
@@ -0,0 +1,69 @@
...
+    can_run_ds,
+    data_dir_load,
+    requires_ds,
+    small_patch_amr,
+)
+
+_fields = ("thrhopert", "zvort")
+cm1sim = "cm1_tornado_lofs/nc4_cm1_lofs_tornado_test.nc"
+
+
```
cphyc, line 15: Should an answer test be added here?

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+                    mylog.error(
+                        (
+                            "It looks like you are trying to load a cm1 netcdf file, "
+                            "but at present yt only supports cm1_lofs output. Until"
+                            " support is added, you can likely use"
+                            " yt.load_uniform_grid() to load your cm1 file manually."
+                        )
+                    )
+                return False
+        except Exception:
```
neutrinoceros, line 230: Exception catching should be as specific as possible, can you please catch only the relevant exceptions here ?

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+                    mylog.error(
+                        (
+                            "Trying to load a cm1_lofs netcdf file but the coordinates "
+                            "of the following fields do not match the coordinates of "
+                            f"the dataset: {failed_vars}"
+                        )
+                    )
+                    return False
+
+            if is_cm1_lofs is False:
```
neutrinoceros, line 219:
```suggestion
            if not is_cm1_lofs:
```

yt/utilities/on_demand_imports.py
```diff
@@ -56,6 +56,15 @@ class netCDF4_imports:
     _name = "netCDF4"
     _Dataset = None
 
+    def __init__(self):
+        try:
+            import netCDF4
+
+            netCDF4.__version__
```
neutrinoceros, line 63:
```suggestion
```

yt/utilities/on_demand_imports.py
```diff
@@ -67,7 +76,7 @@ def Dataset(self):
         return self._Dataset
 
 
-_netCDF4 = netCDF4_imports()
+_netCDF4 = netCDF4_imports()  # Always do that before initializing h5py_imports() !!!
```
neutrinoceros, line 79: Could the comment also include an explanation ?

yt/utilities/file_handler.py
```diff
@@ -111,6 +111,6 @@ def __init__(self, filename):
     def open_ds(self):
         from yt.utilities.on_demand_imports import _netCDF4 as netCDF4
 
-        ds = netCDF4.Dataset(self.filename)
+        ds = netCDF4.Dataset(self.filename, "r")
```
neutrinoceros, line 114: wild guessing here. If I'm right, please commit this. If I'm wrong, it shows that positional arguments are not clear enough, please add the argument name explicitly
```suggestion
        ds = netCDF4.Dataset(self.filename, mode="r")
```

yt/frontends/nc4_cm1/io.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def _read_fluid_selection(self, chunks, selector, fields, size):
```
neutrinoceros, line 30: Important note. As a frontend user, I did not understand what this method was for before it was too late. We now have a bug in the AMRVAC frontend (unfortunately not solved yet) here https://github.com/yt-project/yt/issues/2720

I suggest we test this method with a similar script as found in the bug report.

yt/frontends/nc4_cm1/io.py
```diff
@@ -0,0 +1,62 @@
...
+    def __init__(self, ds):
+        self.filename = ds.filename
+        self._handle = NetCDF4FileHandler(self.filename)
+        super(CM1IOHandler, self).__init__(ds)
+
+    def _read_particle_coords(self, chunks, ptf):
+        # This needs to *yield* a series of tuples of (ptype, (x, y, z)).
+        # chunks is a list of chunks, and ptf is a dict where the keys are
+        # ptypes and the values are lists of fields.
+        pass
```
neutrinoceros, line 20:
```suggestion
        raise NotImplementedError
```

yt/frontends/nc4_cm1/io.py
neutrinoceros, line 28:
```suggestion
        raise NotImplementedError
```

yt/frontends/nc4_cm1/fields.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    known_particle_fields = (
+        # Identical form to above
+        # ( "name", ("units", ["fields", "to", "alias"], # "display_name")),
+    )
+
+    def __init__(self, ds, field_list):
+        super(CM1FieldInfo, self).__init__(ds, field_list)
+        # If you want, you can check self.field_list
```
neutrinoceros, line 57: So this does nothing more than the method it's subclassing ? Could it be removed then ?

neutrinoceros (changes requested):
Overall this looks nice, well done !
I have a few minor questions and remarks
- this is committing an empty `misc.py` file that should either be removed or populated
- I'm confused that the `CM1Dataset` class has a `variables` attribute that seems very close to what `parameters` is supposed to be for. Is there a reason to have both ?
- I'm not reviewing the tests for now. We would need to coordinate this with the testing framework change.

I'll happily approve this once my questions and suggestions are answered :)

yt/frontends/nc4_cm1/io.py
neutrinoceros: For now I advise `raise NotImplementedError`. If CM1 dataset never have particles however they can be removed.

> Also, this goes well beyond the scope of this PR, but is _read_particle_coords still a thing? It seems to be only used for SPH-specific datasets.

Yeah that's one of the reasons why we should have a cleaner class separation for dataset types.

yt/frontends/nc4_cm1/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: beat me to it !

yt/frontends/nc4_cm1/tests/test_outputs.py
chrishavlin: `test_cm1_mesh_fields` is an answer test, ya?

yt/utilities/on_demand_imports.py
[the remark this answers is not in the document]
chrishavlin: ya, so this is confusing... but the `import` is there to ensure due to an hdf5/netCDF4 import order issue. Without this, if hdf5 has been imported previously then you can get weird file lock issues when trying to read using netCDF4. This was causing the travis tests to fail.... 

The `netCDF4.__version__` is only there to avoid flake8 catching an unused import :)

yt/utilities/on_demand_imports.py
chrishavlin: yes haha -- explained above in another comment, but this comment should be clearer here...

yt/utilities/on_demand_imports.py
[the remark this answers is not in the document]
neutrinoceros: In that case just add `# noqa F401` to the import line

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+        super(CM1Dataset, self).__init__(
+            filename, dataset_type, units_override=units_override, unit_system="mks"
+        )
+        self.storage_filename = storage_filename
+        self.filename = filename
+
+    def _setup_coordinate_handler(self):
+        super(CM1Dataset, self)._setup_coordinate_handler()
+        self.coordinates._x_pairs = (("x", "y"), ("y", "x"), ("z", "x"))
+        self.coordinates._y_pairs = (("x", "z"), ("y", "z"), ("z", "y"))
```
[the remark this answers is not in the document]
chrishavlin: I believe this sets the axis ordering on slice plots so that when slicing `z` is always on the vertical axis -- @matthewturk or @keltonhalbert should confirm...

matthewturk:
Yes, it does.

yt/frontends/nc4_cm1/data_structures.py
chrishavlin: `ds` here is the netcdf file handle, where `ds.cm1_lofs_version` is a global attribute. Maybe we should use a different variable name from `ds` to avoid confusion with a `yt` dataset?

matthewturk:
often we use "_handle"

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+        self.hubble_constant = 0.0
+
+    @classmethod
+    def _is_valid(self, *args, **kwargs):
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+
+        warn_netcdf(args[0])
+        try:
+            from netCDF4 import Dataset
```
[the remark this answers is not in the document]
chrishavlin: I think it could be... this follows the `exodus_ii` `is_valid()` but I'm not sure if there was a reason for importing netcdf directly there either.

yt/frontends/nc4_cm1/data_structures.py
chrishavlin: weirdly enough, it's not actually a typo... The standard netcdf output has this global attribute `"cm1 version"` with the space.

yt/frontends/nc4_cm1/data_structures.py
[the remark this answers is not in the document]
munkm: You're correct @chrishavlin. This correctly orders the axes so plots aren't rotated.

yt/frontends/nc4_cm1/data_structures.py
neutrinoceros: oh yes, definitely. Something like `handle` would match existing similar objects in other frontends

yt/frontends/nc4_cm1/data_structures.py
neutrinoceros: wow I didn't expect that. I think this warrants an inline comment.

yt/frontends/nc4_cm1/data_structures.py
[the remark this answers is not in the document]
chrishavlin: ok, cool. we should add a comment to the code here I think.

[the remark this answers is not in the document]
neutrinoceros: Ok thank you both for clarifying this. I suggest to add an inline comment.

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+            # TO DO: eneralize this to be coordiante variable name agnostic in order to
+            # make useful for WRF or climate data. For now, we're hard coding for CM1
+            # specifically and have named the classes appropriately
+            xh, yh, zh = [ds.variables[i][:] for i in ["xh", "yh", "zh"]]
+            self.domain_left_edge = np.array(
+                [xh.min(), yh.min(), zh.min()], dtype="float64"
+            )
+            self.domain_right_edge = np.array(
+                [xh.max(), yh.max(), zh.max()], dtype="float64"
+            )
```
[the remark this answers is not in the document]
chrishavlin: we'll need @keltonhalbert to weigh in on `xh`, `yh` and `zh` (I think they are actually cell centers but I'm not positive?). I think adding a comment to the code would be good here -- maybe including a link to the `cm1_lofs` docs website? 

But I think some confusion is coming from the fact that `ds` here is our netcdf dataset file handle, not a `yt` dataset. So `ds.variables` contains the netcdf objects with the actual data that need to be parsed and read while `ds.dimensions` is a netcdf attribute with dimensionality metadata. I think this will be clearer by switching `ds` to a different name (`_handle` as Matt suggested).

yt/frontends/nc4_cm1/data_structures.py
chrishavlin: ya, it's odd. @keltonhalbert , can you confirm this is right? I'm going off of the standard CM1 data file that Leigh shared with us.

keltonhalbert: @chrishavlin it is odd, yes, but unfortunately that is the way the original, untouched-by-us dataset is for the wider modeling community that uses this tool. 
Screen Shot 2020-09-30 at 1 49 50 PM

yt/frontends/nc4_cm1/io.py
keltonhalbert: We do have particles, but they are contained in separate NetCDF files. I'd lean towards the NotImplementedError for that reason, as we may try to add that in a future pull request

yt/frontends/nc4_cm1/data_structures.py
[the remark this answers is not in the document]
chrishavlin: oh, sorry, was conflating two issues. You just meant the import. Yes, should definitely be able to use the `_netCDF4` import. But I've been confused about why the `is_valid` (here and in exodus_ii) did not use the `NetCDF4FileHandler` that is already imported. After more digging, the reason for **that** seems to be the need for the `keepweakref` argument, since `NetCDF4FileHandler.open_ds` uses hardcoded arguments for the call to the netcdf dataset. So I'm inclined to add `**kwargs` to `NetCDF4FileHandler.open_ds` so that we can simply pass along the `keepweakref` using the same `NetCDF4FileHandler` used in the rest of the front end.

chrishavlin:
just pushed some changes that incorporate most of the suggestions (but not all). will go through now and note the ones that still need addressing.

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+        # for now, the index file is the dataset!
+        self.index_filename = self.dataset.parameter_filename
+        self.directory = os.path.dirname(self.index_filename)
+        # float type for the simulation edges and must be float64 now
+        self.float_type = np.float64
+        super(CM1Hierarchy, self).__init__(ds, dataset_type)
+
+    def _initialize_state_variables(self):
+        super(CM1Hierarchy, self)._initialize_state_variables()
+        self.num_grids = 1
```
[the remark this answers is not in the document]
chrishavlin: yes, I think we can remove this.

yt/frontends/nc4_cm1/data_structures.py
chrishavlin: I ended up removing `coords` entirely -- I'm not sure why we were copying the coordinate data over to parameters as it's not used elsewhere and we can extract it from the yt dataset after loading.

yt/utilities/on_demand_imports.py
```diff
@@ -56,6 +56,17 @@ class netCDF4_imports:
...
+    def __init__(self):
+        # this ensures the import ordering between netcdf4 and h5py. If hy5py is
+        # imported first, can get file lock errors on some systems (including travis-ci)
+        # so we need to do this before initializing h5py_imports()!
+        # similar to this issue https://github.com/pydata/xarray/issues/2560
+        try:
+            import netCDF4  # noqa F401
+        except ImportError:
+            pass
+        super(netCDF4_imports, self).__init__()
```
neutrinoceros, line 68: I don't get this. This class isn't inheriting from anywhere, so what's the effect of super here ?

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,221 @@
...
+            if not is_cm1_lofs:
+                if is_cm1:
+                    mylog.warning(
+                        (
+                            "It looks like you are trying to load a cm1 netcdf file, "
+                            "but at present yt only supports cm1_lofs output. Until"
+                            " support is added, you can likely use"
+                            " yt.load_uniform_grid() to load your cm1 file manually."
+                        )
+                    )
```
neutrinoceros, line 216:
```suggestion
            if not is_cm1_lofs and is_cm1:
                mylog.warning(
                        "It looks like you are trying to load a cm1 netcdf file, "
                        "but at present yt only supports cm1_lofs output. Until"
                        " support is added, you can likely use"
                        " yt.load_uniform_grid() to load your cm1 file manually."
                    )
```

yt/utilities/on_demand_imports.py
Xarthisius: That might be my fault. I just blindly copied `h5py_imports.__init__` without reading it.

chrishavlin: a good question. The issue we were seeing was due to the import order of `h5py` and `netCDF4` (if `h5py` is imported first, using `netCDF4` to read files fails), we were not able to figure out why exactly. @Xarthisius figured out this workaround, he might have some insight.

Xarthisius: I think @neutrinoceros only meant that `super()` is not necessary here, cause there's no parent `__init__`

chrishavlin: We did try the environment variable solution mentioned in the comment [link](https://github.com/pydata/xarray/issues/2560), that didn't work. Didn't seem to be version related and couldn't reproduce it locally, only in the test environment.

neutrinoceros: > I think @neutrinoceros only meant that super() is not necessary here, cause there's no parent __init__

exactly

chrishavlin: Ok, got it. We'll take out the super then

yt/frontends/nc4_cm1/data_structures.py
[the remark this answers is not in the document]
chrishavlin: added to the comments around this section noting the xh, yh and zh coordinates.

yt/frontends/nc4_cm1/data_structures.py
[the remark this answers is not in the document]
chrishavlin: latest commit fixes this: uses the NetCDF4FileHandler and no longer imports from netCDF4 directly.

yt/frontends/nc4_cm1/data_structures.py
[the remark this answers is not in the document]
chrishavlin: removed in latest commit

yt/frontends/nc4_cm1/data_structures.py
chrishavlin: latest commit uses "filename"

yt/frontends/nc4_cm1/data_structures.py
chrishavlin: added a short inline comment noting that the space should be there.

yt/utilities/file_handler.py
chrishavlin: latest commit uses the `mode` keyword.

yt/frontends/nc4_cm1/fields.py
chrishavlin: removed!

yt/frontends/nc4_cm1/io.py
chrishavlin: I just tested this out and it does look like it is an issue here too -- when making a slice plot with a cut region as the data source I get similar behavior as in #2720

chrishavlin: I'll look into it some more and see what I find... but I don't think this has to block this PR. I can open an issue if this PR gets merged before I figure it out?

yt/utilities/on_demand_imports.py
chrishavlin: removed the super call in the latest commit

yt/frontends/nc4_cm1/data_structures.py
chrishavlin: updated in latest commits

yt/frontends/nc4_cm1/io.py
neutrinoceros: Yeah it would be unfair to hold this PR just fo this. A new issue wouldn't be necessary if the bug is similar enough to the one in AMRVAC, but you get the opportunity to add a warning or a NotImplementedError somewhere before this is merged :)

neutrinoceros: @matthewturk and @n-claes, I think this thread might interest you

yt/utilities/on_demand_imports.py
```diff
@@ -56,6 +56,16 @@ class netCDF4_imports:
     _name = "netCDF4"
     _Dataset = None
 
+    def __init__(self):
+        # this ensures the import ordering between netcdf4 and h5py. If hy5py is
```
neutrinoceros, line 60:
```suggestion
        # this ensures the import ordering between netcdf4 and h5py. If h5py is
```

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,216 @@
...
+        self.directory = os.path.dirname(self.index_filename)
+        # float type for the simulation edges and must be float64 now
+        self.float_type = np.float64
+        super(CM1Hierarchy, self).__init__(ds, dataset_type)
+
+    def _detect_output_fields(self):
+        # build list of on-disk fields for dataset_type 'cm1'
+        self.field_list = []
+        for vname in self.dataset.parameters["variable_names"]:
+            self.field_list.append(("cm1", vname))
```
neutrinoceros, line 48:
```suggestion
        vnames = self.dataset.parameters["variable_names"]
        self.field_list = [("cm1", vname) for vname in vnames]
```

neutrinoceros (approved):
Great job @chrishavlin ! I like where this is going. Even though there are still some open questions that need addressing, I'll approve this to indicate the changes I requested were dealt with :-)

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -0,0 +1,233 @@
...
+        #   g.Children <= list of child grids
+        #   g.Parent   <= parent grid
+        # This is handled by the frontend because often the children must be
+        # identified.
+        self.grids = np.empty(self.num_grids, dtype="object")
+        for i in range(self.num_grids):
+            g = self.grid(i, self, self.grid_levels.flat[i], self.grid_dimensions[i])
+            g._prepare_grid()
+            g._setup_dx()
+            self.grids[i] = g
```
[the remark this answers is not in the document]
neutrinoceros: @matthewturk can you weigh in here please ?

chrishavlin:
added a note about the cut region bug in the existing issue #2720

yt/frontends/nc4_cm1/data_structures.py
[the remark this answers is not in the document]
matthewturk: So it depends.  The reason we ever do `self.grids[i]` is if we are creating an object array in advance.  I think it might also no longer be necessary -- there was a segfault that we ran into many years ago as a result of numpy object arrays, which is perhaps no longer operative.

But the idea was, create an empty numpy object array, then fill it.  It's possible that in AMRVAC it creates the grid objects elsewhere and they are already populated.

munkm:
Ok, awesome! I think this looks good to go, then, so I'll merge it and we can resolve the issue a bit down the road. 🙂

munkm:
Congratulations on getting this frontend over the finish line @keltonhalbert and @chrishavlin !!!

PR #2186, opened 2019-03-05
Facilitate SPH visualization of non-SPH particles

### LINKED ISSUE #2072: No way to do ProjectionPlot on non-SPH demeshed particle data
### Bug report

**Bug summary**

Asking for a projection plot on non-SPH particle data in yt-4.0 (i.e. demeshed yt) suggests using a ParticlePlot (which does not provide the same visual or scientific appeal), or fails if you try to push it too hard.

**Code for reproduction**

Once loading particle data as `ds`, here accomplished with `yt.load_particles`:

```python

prj = yt.ProjectionPlot(ds, 0, ('star', 'particle_mass'), center=center)
   ...:
YTInvalidFieldType                        Traceback (most recent call last)
<ipython-input-8-f0cebe42ef5b> in <module>
----> 1 prj = yt.ProjectionPlot(ds, 0, ('star', 'particle_mass'), center=center)

~/yt4/src/yt/yt/visualization/plot_window.py in __init__(self, ds, axis, fields, center, width, axes_unit, weight_field, max_level, origin, right_handed, fontsize, field_parameters, data_source, method, proj_style, window_size, aspect)
   1462         # plotting classes to avoid an exception
   1463         test_data_source = ds.all_data()
-> 1464         validate_mesh_fields(test_data_source, fields)
   1465
   1466         if isinstance(ds, YTSpatialPlotDataset):

~/yt4/src/yt/yt/visualization/plot_window.py in validate_mesh_fields(data_source, fields)
    147
    148     if len(invalid_fields) > 0:
--> 149         raise YTInvalidFieldType(invalid_fields)
    150
    151

YTInvalidFieldType:
SlicePlot, ProjectionPlot, and OffAxisProjectionPlot can only plot fields that
are defined on a mesh or for SPH particles, but received the following N-body
particle fields:

    [('star', 'particle_mass')]

Did you mean to use ParticlePlot or plot a deposited particle field instead?
```

or, alternatively (and probably what _should_ work in the future):

```python

prj = yt.ProjectionPlot(ds, 0, ('deposit', 'star_density'), center=center)
   ...:
YTNonIndexedDataContainer                 Traceback (most recent call last)
<ipython-input-9-81448988a4b4> in <module>
----> 1 prj = yt.ProjectionPlot(ds, 0, ('deposit', 'star_density'), center=center)

~/yt4/src/yt/yt/visualization/plot_window.py in __init__(self, ds, axis, fields, center, width, axes_unit, weight_field, max_level, origin, right_handed, fontsize, field_parameters, data_source, method, proj_style, window_size, aspect)
   1478                            center=center, data_source=data_source,
   1479                            field_parameters=field_parameters, method=method,
-> 1480                            max_level=max_level)
   1481         PWViewerMPL.__init__(self, proj, bounds, fields=fields, origin=origin,
   1482                              right_handed=right_handed, fontsize=fontsize, window_size=window_size,

~/yt4/src/yt/yt/data_objects/construction_data_containers.py in __init__(self, field, axis, weight_field, center, ds, data_source, style, method, field_parameters, max_level)
    446
    447         if not self.deserialize(field):
--> 448             self.get_data(field)
    449             self.serialize()
    450

~/yt4/src/yt/yt/data_objects/construction_data_containers.py in get_data(self, fields)
    248                     self._initialize_projected_units(fields, chunk)
    249                     _units_initialized = True
--> 250                 self._handle_chunk(chunk, fields, tree)
    251         # if there's less than nprocs chunks, units won't be initialized
    252         # on all processors, so sync with _projected_units on rank 0

~/yt4/src/yt/yt/data_objects/construction_data_containers.py in _handle_chunk(self, chunk, fields, tree)
    503     def _handle_chunk(self, chunk, fields, tree):
    504         mylog.debug("Adding chunk (%s) to tree (%0.3e GB RAM)",
--> 505                     chunk.ires.size, get_memory_usage()/1024.)
    506         if self.method == "mip" or self._sum_only:
    507             dl = self.ds.quan(1.0, "")

~/yt4/src/yt/yt/data_objects/data_containers.py in ires(self)
   1575         if self._current_chunk is None:
   1576             self.index._identify_base_chunk(self)
-> 1577         return self._current_chunk.ires
   1578
   1579     @property

~/yt4/src/yt/yt/geometry/geometry_handler.py in cached_func(self)
    267             return getattr(self, n)
    268         if self.data_size is None:
--> 269             tr = self._accumulate_values(n[1:])
    270         else:
    271             tr = func(self)

~/yt4/src/yt/yt/geometry/geometry_handler.py in _accumulate_values(self, method)
    295         for obj in self._fast_index or self.objs:
    296             f = getattr(obj, mname)
--> 297             arrs.append(f(self.dobj))
    298         if method == "dtcoords":
    299             arrs = [arr[0] for arr in arrs]

~/yt4/src/yt/yt/data_objects/particle_container.py in _func_non_indexed(self, *args, **kwargs)
     28 def _non_indexed(name):
     29     def _func_non_indexed(self, *args, **kwargs):
---> 30         raise YTNonIndexedDataContainer(self)
     31     return _func_non_indexed
     32

YTNonIndexedDataContainer: The data container (<class 'yt.data_objects.particle_container.ParticleContainer'>) is an unindexed type.  Operations such as ires, icoords, fcoords and fwidth will not work on it.

```

**Expected outcome**

Should be able to visualize at least the density fields of non-SPH particle data.

**Version Information**

  * Operating System:
  * Python Version:
  * yt version:

Using yt-4.0dev0, merge-rev 8154cbffd 751b2239e

qobilidop:
## PR Summary

This PR provides a way to resolve #2072. For example, the following workflow is enabled to make SPH projection of DM particles (without existing smoothing length field):

```python
import yt

# Load dataset and center on the dense region
_, center = ds.find_max(('PartType0', 'density'))

# Reload DM particles into a stream dataset
ad = ds.all_data()
pt = 'PartType1'
fields = ['particle_mass'] + [f'particle_position_{ax}' for ax in 'xyz']
data = {field: ad[pt, field] for field in fields}
ds_dm = yt.load_particles(data, data_source=ad)

# Generate the missing SPH fields
ds_dm.add_sph_fields()

# Make the SPH projection plot
p1 = yt.ProjectionPlot(ds_dm, 'z', ('io', 'density'),
                       center=center, width=(800, 'kpc'))
p1.set_unit('density', 'Msun/kpc**2')
p1.set_zlim('density', 1e5, 3e9)
p1.show()
p1.save()

# Compare with the poor particle plot
fx, fy = [('io', f'particle_position_{ax}') for ax in 'xy']
p2 = yt.ParticlePlot(ds_dm, fx, fy, ('io', 'particle_mass'),
                     center=center, width=(800, 'kpc'))
p2.set_unit('particle_mass', 'Msun')
p2.set_zlim('particle_mass', 1e5, 3e9)
p2.show()
p2.save()
```

Output:

ParticleData_Projection_z_density
ParticleData_Particle_z_particle_mass

Main API changes:

- An argument `data_source` is added to `load_particles` to facilitate loading from existing datasets.
- A method `add_sph_fields` is added to `StreamParticlesDataset` to enable generating `smoothing_length` and `density` fields on demand.

## PR Checklist

- [x] Add docstrings
- [x] Add narrative docs
- [x] Add example
- [x] Add tests


qobilidop:
Wrong branch.

yt/frontends/stream/data_structures.py
```diff
@@ -1216,6 +1219,118 @@ def load_particles(data, length_unit = None, bbox=None,
...
+
+    >>> ds_sph = load_as_sph(ds, 'PartType1')
+
+    """
+
+    l_unit = 'code_length'
+    m_unit = 'code_mass'
+    d_unit = 'code_mass / code_length**3'
+
+    ad = ds.all_data()
```
chummels: It might be worthwhile to make it so a data object can be specified as a kwarg, and default to `all_data`.  I can see a use case for people wanting to create an sph dataset from a subset of the overall dataset.

yt/frontends/stream/data_structures.py
```diff
@@ -26,6 +26,7 @@
 
 from numbers import Number as numeric_type
 
+from cykdtree import PyKDTree
```
chummels: You'll want to swap this import to the new vendored cykdtree:
from yt.utilities.lib.cykdtree import PyKDTree

yt/frontends/stream/data_structures.py
qobilidop: Thanks for the suggestion. I've changed the API.

qobilidop:
@chummels I've redesigned the API to abstract out a `StreamParticlesDataset.add_sph_fields` method, which might help reduce the code repetition in your #2187.

yt/frontends/stream/data_structures.py
```diff
@@ -1069,6 +1072,49 @@ def _setup_data_io(self):
         else:
             self.io = io_registry[self.dataset_type](self.ds)
```
qobilidop: @jzuhone May I get your review of the following method? It's modified from `StreamHierarchy.update_data` which I believe is written by you. There are several method calls I feel are specific to grid datasets. I've tried my best to check, but I'm not entirely sure. It would be very helpful if someone more familiar with the codebase could take a look.

yt/frontends/stream/data_structures.py
```diff
@@ -1069,6 +1072,49 @@ def _setup_data_io(self):
...
 
+    def update_data(self, data):
```
qobilidop: @jzuhone May I get your review of this method? It's modified from StreamHierarchy.update_data which I believe is written by you. There are several method calls I feel are specific to grid datasets. I've tried my best to check, but I'm not entirely sure. It would be very helpful if someone more familiar with the codebase could take a look.

jzuhone: I will try to have a look at this before the end of the week.

qobilidop: Thanks a lot!

qobilidop:
I'm thinking about adding an example like the above one. Any suggestion where to add it?

ngoldbaum:
A cookbook recipe?

qobilidop:
Good idea, probably in the "Calculating Dataset Information" section?

ngoldbaum:
I'm not sure. It's more an example of how to use the stream frontend to load arbitrary data though, which right now lives in the loading data docs.

This could probably live in an entirely new section on how to work with SPH data specifically but that doesn't exist yet.

Use your best judgement and keep in mind that the docs still probably need a lot of work before we release yt 4.0.

yt/frontends/stream/data_structures.py
```diff
@@ -1092,11 +1138,81 @@ def __init__(self, stream_handler, storage_filename=None,
...
+                right_edge=right_edge,
+                periodic=self.periodicity,
+                leafsize=2*int(n_neighbors),
+            )
+            pos_kd = pos[kdtree.idx]
+            hsml = generate_smoothing_length(pos_kd, kdtree, n_neighbors)
+            hsml = hsml[np.argsort(kdtree.idx)]
+            data[(sph_ptype, fname)] = (hsml, l_unit)
+
+        # Compute density field
```
qobilidop: I need some help with the computation of density given mass and smoothing length. There are issues at two levels:

1. The most accurate SPH density estimation requires a non-local operation, which is not readily available in yt yet. Do we want to go in that direction and add the tool?

2. It's probably okay we don't do the most accurate computation here. Because the user could provide the density field themselves if this is important. We just need to make it clear in the docs what we are doing here. `dens = mass / vol` is correct in dimension, but might be wrong in normalization, which is dependant on the smoothing kernel. My proposal in this direction is to add a `normalization_factor` argument to this method, which defaults to the case of the cubic spline.

I couldn't make a decision on myself. So I'm looking forward to having more discussions with people.

yt/frontends/stream/data_structures.py
```diff
@@ -1092,11 +1138,73 @@ def __init__(self, stream_handler, storage_filename=None,
...
+        m_unit = "code_mass"
+        d_unit = "code_mass / code_length**3"
+
+        # Read basic fields
+        ad = self.all_data()
+        pos = ad[sph_ptype, "particle_position"].to(l_unit).d
+        mass = ad[sph_ptype, "particle_mass"].to(m_unit).d
+
+        # Construct k-d tree
+        left_edge = self.domain_left_edge.to(l_unit).v
```
jzuhone: Super minor nit here, but `YTArray` has a `to_value` method which basically does the same as `to(l_unit).v` does here.

yt/frontends/stream/data_structures.py
jzuhone: I think this looks ok to me.

doc/source/examining/loading_data.rst
```diff
@@ -1661,6 +1661,41 @@ The ``load_particles`` function also accepts the following keyword parameters:
...
+
+    # Load dataset
+    ds = yt.load('FIRE_M12i_ref11/snapshot_600.hdf5')
+
+    # Reload star particles into a stream dataset
+    ad = ds.all_data()
+    pt = 'PartType4'
+    fields = ['particle_mass'] + [f'particle_position_{ax}' for ax in 'xyz']
+    data = {('io', field): ad[pt, field] for field in fields}
+    ds_star = yt.load_particles(data, data_source=ad)
```
jzuhone: So what is exactly happening here? Do the "star" fields become "gas" fields in `ds_star`? If so, that seems a bit confusing to me.

jzuhone: It looks like they're just `"io"`, I guess.

qobilidop: Yes, the particle type becomes `'io'`, which is specified in the data dictionary. But this is just an arbitrary choice following the convention that `'io'` is the SPH particle type. Later in `ds_star.add_sph_fields()`, `sph_ptype='io'` is assumed as the default. But it could be set to other value, `ds_star.add_sph_fields(sph_ptype='star')` for example.

yt/frontends/stream/data_structures.py
qobilidop: Thanks! I don't know that before.

doc/source/examining/loading_data.rst
jzuhone: Ok, this is fine then. Thanks for the clarification!

jzuhone: So, I can't help wondering if there is a way to encapsulate these steps into a single function:

```python
data = {('io', field): ad[pt, field] for field in fields}
ds_star = yt.load_particles(data, data_source=ad)
```
Something where you'd pass in `pt` and `fields` and then this would happen under the hood.

qobilidop: Thanks for bringing this up. I have thought about that in a [previous design](https://github.com/yt-project/yt/pull/2186/commits/d00a96efe022cd66e36345fc43ee4976b7c84635#diff-fc0105fc04a164b7cdc4e4fc59b91064L1336), but later turned away. The main reason is I haven't figured out how to do it without adding a new function. And I'm not convinced this use case is common enough to justify a special function.

There is probably an elegant way to do it. I'd like to have more discussions about this. @MatthewTurk @brittonsmith @ngoldbaum might also have some perspectives.

qobilidop:
This PR is mostly done, except for an API design discussion (see the thread above). Reviews are welcome.

doc/source/examining/loading_data.rst
jzuhone: The thing is that the `add_sph_fields` method is new, but I can only see it realistically being used for this use case (feel free to object, however, as I could be missing something). So I wonder if `add_sph_fields` should really be `_add_sph_fields` and there is a method that hangs off the original dataset which creates the new stream dataset and calls `_add_sph_fields` at the end.

ngoldbaum: I think it's good that it's public. It should be possible to do this with an arbitrary N-body particle dataset, it doesn't necessarily need to be done via `load_particles` I think.

I also think making this project too magical might be brittle or hard to extend, it's good that it's sort of manual right now as it forces the user to think about what they're doing.

jzuhone: That's fair.

yt/frontends/stream/data_structures.py
```diff
@@ -1092,11 +1138,71 @@ def __init__(self, stream_handler, storage_filename=None,
...
+    def add_sph_fields(self, n_neighbors=32, kernel="cubic", sph_ptype="io"):
+        """Add SPH fields for the specified particle type.
+
+        For a particle type with "particle_position" and "particle_mass" already
+        defined, this method adds the "smoothing_length" and "density" fields.
+        "smoothing_length" is computed as the distance to the nth nearest
+        neighbor. "density" is computed as the SPH (gather) smoothed mass.
+
+        Caution: this method will overwrite existing "smoothing_length" and
+        "density" fields.
```
qobilidop: This is probably my last design question. Currently, the `add_sph_fields` method overwrites existing fields, which is the easiest to implement, less ambiguous to interpret, but not as flexible to use. My argument for this is that "smoothing_length" and "density" are usually computed together, so it doesn't make much sense to use this method to only generate one but not the other field. But I might have overlooked some use cases. Should I change this behavior?

ngoldbaum: I think that's fine. Maybe it should noisily print a user-visible warning when a field actually gets clobbered?

We could also add a mode where we use user-supplied smoothing lengths or densities or both, and just be ok with the possibility that they aren't self-consistent.

qobilidop: Okay, it's probably better to change this behavior then. I'll take a look. Thanks!

doc/source/examining/loading_data.rst
chummels: This functionality will be included in my current PR #2187 when this one gets merged.

yt/frontends/stream/data_structures.py
qobilidop: In the new design, smoothing length and density are computed separately. And the `add_sph_fields` method only adds the SPH fields when they don't already exist.

qobilidop:
I have added some Cython code to do ball search with the k-d tree, which is not in the best shape it could be. There's a lot of repeated code I just copied and tweaked. For reviewers who see immediate ways to improve, I'm more than happy to take your suggestions. But for more in-depth refactoring, I might not have the energy to do so any time soon. Given that this functionality is needed in some other places, I prefer to merge it as soon as the public API looks good.

qobilidop:
@ngoldbaum I've done the manual tests with bounds checking on using the example above and there's no issue 🍻

doc/source/examining/loading_data.rst
qobilidop: I think this is settled for now. I'll stay with the current design.

AshKelly:
Hey, this is awesome,

I had some issues testing as I couldn't get yt to read in the Auriga, Apostle or EAGLE snapshots that I had to hand (unrelated to this PR it seems). But, I eventually tested it with FIRE snap and it looks great!

Thanks @qobilidop!

ngoldbaum:
That doesn’t sound great... If you don’t have time to fix those issues a bug report would be appreciated. Thank you for testing :)

PR #2187, opened 2019-03-06
[WIP] Adds function for creating particle-based dataset from grid-based dataset

chummels:
## PR Summary

This adds the `monte_carlo_sample()` function for converting grid-based datasets to particle-based datasets.  This can be used in a number of ways including providing a means of direct comparisons between grid-based and particle-based codes using the same starting point.  However, this will help yt to be used with the FIREFly code for real-time visualization in the web-browser for both particle-based as well as grid-based datasets (FIREFly was built for the particle-based FIRE simualtions).

This is a WIP so I'm open to making significant changes to this code, like its location and method.  Couldn't find an obvious place to put it which is why it is here.

This function is reasonably fast at the 1e5-1e6 sampling level, operating in just a few seconds when I use `('gas', 'dx')` as an analog for the `('gas', 'smoothing_length')`.  However, when I use the kdtree to determine the smoothing length field, it introduces some problems.  One, it is a bit slower.  Two, the smoothing lengths appear to be really really large, causing the image generation step to take a lot longer and the image to get all washed out.  I guess I could just arbitrarily shrink the `smoothing_length` value from the one calculated by the kdtree, but I figure either I'm doing something wrong or there is a bug in the `generate_smoothing_length()` code.  Images below.  Also, almost all of the code for the kdtree stuff was taken directly from @qobilidop 's open PR #2186 (thanks, @qobilidop !), so it may make more sense to modularize some of these functions and use them in both pieces of functionality.  

Script to test and generate images:

```
import yt
from yt.data_objects.data_containers import \
    monte_carlo_sample
fn = '~/src/yt-data/IsolatedGalaxy/galaxy0030/galaxy0030'
ds = yt.load(fn)
ds_particle = monte_carlo_sample(ds)

p = yt.ProjectionPlot(ds, 'z', ('gas', 'density'), width=(0.04, 'Mpc'))
p.set_zlim(('gas', 'density'), 1e-4, 1e-2)
p.save('grid.png')

p = yt.ProjectionPlot(ds_particle, 'z', ('gas', 'density'), width=(0.04, 'Mpc'))
p.set_zlim(('gas', 'density'), 1e-4, 1e-2)
p.save('particles.png')
```

original grid dataset -- grid.png:

grid

generated particle dataset -- particle.png (incorrectly using `dx` as the `smoothing_length`):

particle_fixed

generated particle dataset -- particle.png (using the kdtree to generate the `smoothing_length`):

particles

## PR Checklist

- [ ] Adds tests.

### DIFF
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -71,6 +71,8 @@ from yt.units.yt_array import uconcatenate
 from yt.data_objects.field_data import YTFieldData
 from yt.data_objects.profiles import create_profile
 from yt.utilities.lib.cykdtree import PyKDTree
+from yt.utilities.lib.particle_kdtree_tools import \
+    generate_smoothing_length
 
 data_object_registry = {}
 
@@ -2181,16 +2183,46 @@ def _reconstruct_object(*args, **kwargs):
     obj.field_parameters.update(field_parameters)
     return ReconstructedObject((ds, obj))
 
-def monte_carlo_sample(ds, n_samples=100000, fields=None, n_neighbors=2):
+def monte_carlo_sample(ds, n_samples=100000, fields=None, n_neighbors=8):
     """
+    Uses Monte Carlo sampling to create a particle-based dataset out of a
+    grid-based dataset.  For each sample, the code probabilistically selects
+    a cell weighted according to its mass, and a particle is created possessing
+    the gas fields of the cell at some random location within the cell's volume.
+
+    By default the fields that are created are mass, density, smoothing_length,
+    and the position and velocity fields.
+
+    Parameters
+    ----------
+    ds : Dataset
+        Grid-based dataset used as source
+
+    n_samples : int
+        Number of times to sample the dataset
+
+    fields : list
+        List of additional fields you want to include in returned dataset
+
+    n_neighbors : int
+        Number of neighbors for each node in kdtree to determine
+        smoothing_length
+
+    Returns
+    -------
+    StreamParticlesDataset
+
+    Examples
+    --------
+    >>> ds_grid = yt.load('IsolatedGalaxy/galaxy0030/galaxy0030')
+    >>> ds_sph = monte_carlo_sample(ds_grid)
     """
-    from yt.utilities.lib.particle_kdtree_tools import \
-        generate_smoothing_length
     from yt.frontends.stream.api import \
         load_particles
     l_unit = 'code_length'
     m_unit = 'code_mass'
     d_unit = 'code_mass/code_length**3'
+    v_unit = 'code_length/code_time'
     rands = np.random.rand(n_samples)
     m_tot = ds.r[('gas', 'cell_mass')].sum()
     bins = np.cumsum(ds.r[('gas', 'cell_mass')] / m_tot)
@@ -2221,9 +2253,13 @@ def monte_carlo_sample(ds, n_samples=100000, fields=None, n_neighbors=2):
     data = {'particle_position_x' : (x_pos.to(l_unit), l_unit),
             'particle_position_y' : (y_pos.to(l_unit), l_unit),
             'particle_position_z' : (z_pos.to(l_unit), l_unit),
+            'particle_velocity_x' : (ds.r[('gas', 'velocity_x')][indices].to(v_unit), v_unit),
+            'particle_velocity_y' : (ds.r[('gas', 'velocity_y')][indices].to(v_unit), v_unit),
+            'particle_velocity_z' : (ds.r[('gas', 'velocity_z')][indices].to(v_unit), v_unit),
             'particle_mass' : (masses.to(m_unit), m_unit),
             'density' : (ds.r[('gas', 'density')][indices].to(d_unit), d_unit),
             'smoothing_length' : (hsml, l_unit)}
+            #'smoothing_length' : (ds.r[('gas', 'dx')][indices].to(l_unit), l_unit)}
 
     # Collect dataset code units
     code_units = {}
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -2183,7 +2183,7 @@ def _reconstruct_object(*args, **kwargs):
     obj.field_parameters.update(field_parameters)
     return ReconstructedObject((ds, obj))
 
-def monte_carlo_sample(ds, n_samples=100000, fields=None, n_neighbors=8):
+def monte_carlo_sample(ds, n_samples=100000, fields=None, n_neighbors=1):
     """
     Uses Monte Carlo sampling to create a particle-based dataset out of a
     grid-based dataset.  For each sample, the code probabilistically selects
@@ -2259,7 +2259,6 @@ def monte_carlo_sample(ds, n_samples=100000, fields=None, n_neighbors=8):
             'particle_mass' : (masses.to(m_unit), m_unit),
             'density' : (ds.r[('gas', 'density')][indices].to(d_unit), d_unit),
             'smoothing_length' : (hsml, l_unit)}
-            #'smoothing_length' : (ds.r[('gas', 'dx')][indices].to(l_unit), l_unit)}
 
     # Collect dataset code units
     code_units = {}
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -70,6 +70,7 @@ from yt.extern.six import add_metaclass, string_types
 from yt.units.yt_array import uconcatenate
 from yt.data_objects.field_data import YTFieldData
 from yt.data_objects.profiles import create_profile
+from yt.utilities.lib.cykdtree import PyKDTree
 
 data_object_registry = {}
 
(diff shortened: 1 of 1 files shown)


MatthewTurk:
@chummels It's reporting far more PEP8 errors than our flake8 usually catches, but I'm not sure why.  In the latest commit on master it excludes these.  So for now, I would suggest ignoring its reply...

MatthewTurk:
(The PEP-8 bot should be fixed by #2188 .)

chummels:
That's pretty sweet it gives all this feedback!  Much better than the PEP8 tests failing and then you have to look through the test outputs to figure out what went wrong.

chummels:
dx_vs_smoothing_len

No, not linearly increasing.  In this simulation, the grids are for the most part nested around the galaxy with discrete transitions, whereas the smoothing length not so much.

chummels:
But it does illustrate that the smoothing lengths are ~10x larger than the dx values, hence the really washed out structure in the final image.

ngoldbaum:
If you use more particles to generate the sampled particle dataset does it converge better and look less washed out?

chummels:
Yes, a bit, but it becomes increasingly long processing time.  For references, the number of grid elements in the original dataset is 3.6e6.  Here are the results of various different values of `n_samples` (i.e., particles) approaching the total number of grids, along with the processing time.

1e3 particles, 9 seconds on a Macbook Pro:
particles_1e3

1e4 particles, 14 seconds on a Macbook Pro:
particles_1e4

1e5 particles, 44 seconds on a Macbook Pro:
particles_1e5

1e6 particles, 483 seconds on a Macbook Pro:
particles_1e6

Things sharpen a bit as I drop the `n_neighbors` value, which is why my default is down at 8 instead of 64.  This is 1e6 particles with `n_neighbors` of 2, 1206 seconds on a Macbook Pro:
particles

In contrast to 1e6 particles using the `dx` field for the smoothing length, 15 seconds on a Macbook Pro:
particles

chummels:
For further reference, using `n_neighbors = 1` and `n_samples = 3e6` (the same number of particles now as the number of cells in the original dataset, we still don't entirely look like the original dataset.  I don't know of a way to make it "sharper", and the processing time on this took over an hour, for what is nominally not that large of an original dataset.  3884 seconds.

particles

chummels:
It seems to me that because we're producing the particle-based dataset from the original grid-based dataset, there is additional information present that can be used for the `smoothing_length` instead of just calculating it from scratch, which is nominally what we're doing by invoking the cykdtree calculation step.  But it seems that just using the `dx` value is inappropriate and arguably too small, which is leading to the noise seen when we use that method.  It also leads to grid artifacts across discrete grid boundaries in the original dataset.  But perhaps some variant on that?  

I've tried smoothing the `dx` field with a gaussian kernel before using it as the `smoothing_length`, which works OK, but ends up smearing out structure between the galactic disk and its halo, unsurprisingly.  But it does get rid of the grid artifacts.  And it's fast.  Definitely needs some more thought.

particles

yt/data_objects/data_containers.py
```diff
@@ -2179,3 +2182,93 @@ def _reconstruct_object(*args, **kwargs):
     obj = cls(*new_args)
     obj.field_parameters.update(field_parameters)
     return ReconstructedObject((ds, obj))
+
+def monte_carlo_sample(ds, n_samples=100000, fields=None, n_neighbors=1):
```
MatthewTurk: I think if this is a top-level function, it should probably go into a different file, perhaps into a file all by itself, but it seems to me that making it a function on a data object might be a bit easier and could keep it in this file.

It would be nice to have it possible to specify the `field_type` to sample, too.

yt/data_objects/data_containers.py
```diff
@@ -2179,3 +2182,93 @@ def _reconstruct_object(*args, **kwargs):
...
+    >>> ds_sph = monte_carlo_sample(ds_grid)
+    """
+    from yt.frontends.stream.api import \
+        load_particles
+    l_unit = 'code_length'
+    m_unit = 'code_mass'
+    d_unit = 'code_mass/code_length**3'
+    v_unit = 'code_length/code_time'
+    rands = np.random.rand(n_samples)
+    m_tot = ds.r[('gas', 'cell_mass')].sum()
```
MatthewTurk: If this gets turned into something that samples a data object, this won't be a relevant comment, but I would suggest instead of using `ds.r[tuple]`, you make a new data object by doing `ds.all_data()` -- when queried in that way, `ds.r[]` will re-use the same data object, so the fields will stick around.

yt/data_objects/data_containers.py
```diff
@@ -2179,3 +2182,93 @@ def _reconstruct_object(*args, **kwargs):
...
+    bins = np.cumsum(ds.r[('gas', 'cell_mass')] / m_tot)
```
MatthewTurk: should it be possible to choose mass or volume weighting?

yt/data_objects/data_containers.py
```diff
@@ -2179,3 +2182,93 @@ def _reconstruct_object(*args, **kwargs):
...
+    indices = np.digitize(rands, bins)
+    x_off = np.random.uniform(-0.5, 0.5, n_samples)
```
MatthewTurk: do you anticipate anyone ever wanting to change the random function?

yt/data_objects/data_containers.py
```diff
@@ -2179,3 +2182,93 @@ def _reconstruct_object(*args, **kwargs):
...
+            'particle_velocity_x' : (ds.r[('gas', 'velocity_x')][indices].to(v_unit), v_unit),
+            'particle_velocity_y' : (ds.r[('gas', 'velocity_y')][indices].to(v_unit), v_unit),
+            'particle_velocity_z' : (ds.r[('gas', 'velocity_z')][indices].to(v_unit), v_unit),
+            'particle_mass' : (masses.to(m_unit), m_unit),
+            'density' : (ds.r[('gas', 'density')][indices].to(d_unit), d_unit),
+            'smoothing_length' : (hsml, l_unit)}
+
+    # Collect dataset code units
+    code_units = {}
+    for key in ['length', 'mass', 'time', 'velocity', 'magnetic']:
```
MatthewTurk: Could this be gotten from the unit registry, instead of via `getattr` and `hasattr`?

yt/data_objects/data_containers.py
chummels: Yeah, I'm dithering on whether or not it should be a `Dataset` method.  Honestly, I don't think it has enough widespread use to necessitate being attached to the `Dataset` object.  Plus, I'm not entirely sure how to only have it be on Grid-based datasets short of going through and adding it to each grid-based frontend.  So it may be best to just have it as a stand-alone function sitting somewhere, but not sure where.

yt/data_objects/data_containers.py
chummels: That's fair.  I wasn't sure if that would unnecessarily allocate all of `ad`s fields and use up memory.  But I think you're right that this is the way to go.

ngoldbaum: If it does end up as a standalone function it'll need to be added to the API docs manually.

yt/data_objects/data_containers.py
chummels: Sure, I can make it an option.  I *think* that `cell_mass` is the only one that really makes sense for reproducing a grid-dataset, but I guess someone could use whatever field they wanted as the weight field.  I'll leave it as an option and default to cell mass.

yt/data_objects/data_containers.py
chummels: I didn't really, no.  I guess i could make this a kwarg, but given the nature of Eulerian hydrodynamics, it seems appropriate to uniformly sample the entirety of a cell's volume as opposed to weighting it toward the center in any way.  But I could make this more flexible.

ngoldbaum:
It occurs to me that it should be possible to convert an AMR dataset into an SPH dataset than back into an AMR dataset using @AshKelly's cyoctree :)

yt/data_objects/data_containers.py
chummels: Yeah, probably.  I stole this from @qobilidop 's PR, since he did such a great job marking this up for his PR #2186 :). :)

chummels:
Yeah, a bit like google translating things back and forth to <language of choice>.  I think it would introduce some noise, but should in principle work.

yt/data_objects/data_containers.py
MatthewTurk: Ah!  Makes sense.  I think you've answered my question and it doesn't need any changing!

yt/data_objects/data_containers.py
MatthewTurk: I wasn't entirely sure if `cell_volume` would ever be appropriate but (as noted below) I have not necessarily thought about this carefully enough.  I was able to sort of convince myself a case *might* exist for volume, but not the specifics.

yt/data_objects/data_containers.py
MatthewTurk: I think you're safer with `ad = ds.all_data()` since the fields will not be persistent outside the function.

It does make me wonder if it would be worthwhile to add a derived quantity that does a random sampling.

yt/data_objects/data_containers.py
MatthewTurk: Oh, I meant data object -- so that you could just resample, say, a sphere.

yt/data_objects/data_containers.py
qobilidop: I wasn't clear where to get the code units of a dataset when I wrote this. I'll look into the unit registry. Thanks for the suggestion!

agurvich:
What happens if you crank up the number of particles to say, 1e7? It might take forever but does it get the right answer? What about self-consistent particle merging afterwards to follow the density gradient?

MatthewTurk:
@chummels do you know what the rate determining step is? There might also be straightforward optimizations that could be applied...

yt/data_objects/data_containers.py
jzuhone: There is the little-used and not-documented particle generators stuff:

https://github.com/yt-project/yt/blob/master/yt/utilities/particle_generator.py

jzuhone: Basically what you are doing is a better version of this class:

https://github.com/yt-project/yt/blob/master/yt/utilities/particle_generator.py#L328

chummels:
@MatthewTurk The bottleneck in this process is never the monte carlo step.  In the case where the number of particles (`n_samples`) is large (>=1e6), then the bottleneck is the creation of the kdtree.  The other slowdown can occur when the smoothing length is relatively large, in which case the actual `ProjectionPlot` rendering step takes forever because each pixel, which is intrinsically a line integral, has to include more particles that overlap its trajectory.

chummels:
@agurvich I can try with 1e7 particles, but it's going to take like 8-12 hours to run, and I don't think this is a viable solution for most datasets, given the fact that this is a reasonably small, simple dataset.

agurvich:
yes, I agree. Just curious if it converges. Claude-André and I were discussing it at group meeting Friday morning.

yt/data_objects/data_containers.py
qobilidop: I've switched to use `ds.quan(1, 'code_length').in_cgs()` and the like in my PR. I haven't figured out how to interact with unit registry effectively. Is this usage okay?

yt/data_objects/data_containers.py
chummels: Hmm.  So your suggestion is to make it attached to the derived quantity functionality?  But this isn't really a "quantity" as opposed to an additional data_object, I think.  But I do like the idea of making it work on individual data objects as opposed to datasets as a whole.

yt/data_objects/data_containers.py
chummels: I'm not an expert with the unit_registry, so @ngoldbaum would be the person to answer that question.

chummels:
I addressed a bunch of the feedback from the first iteration.  It now works with a specified data object; it can either calculate the smoothing length from the kdtree or just use the `dx` values of the original dataset for speed or when the smoothing length isn't necessary to be 100% accurate (i.e., firefly) and allows a non-mass weighting.  I'm still open to moving this to another location or changing it if there are other suggestions.  When PR #2186 merges, I will use the functions @qobilidop included to avoid too much code repetition.

matthewturk:
@chummels Do you still want this to go in, and if so, do you need help with the merge conflicts?

chummels:
Yeah, I think this would be useful to go in, since all the basic functionality is there.  I guess we just never settled on what to use as the smoothing length, whether to grab a smoothed version of dx or to let the kd_tree generate it, which is either really slow but reasonably accurate (for lots of particles) or fast and really washed out (for few particles).  I guess we could have both options available for the end user and describe the caveats of both methods?  

I haven't looked at all the conflicts, but I'd be up for giving this the final push to get in the code.  I think it could be useful for code comparison projects or at the very least for cramming grid-based datasets into tools that only work with particle-based datasets (like firefly or potentially cosmovis).

agurvich:
I'd love to play around with some of this in Firefly, so I'd definitely like to see it merged. The only thing I can add is that I've been using [this code](https://github.com/agurvich/FIRE_studio/tree/master/firestudio/utils/stellar_utils/c_libraries/StellarHsml) to compute smoothing lengths for my own purposes (I think either Phil or Volker wrote it originally). 

I imagine it's pretty similar to what y'all are using for the KD tree ¯\_(ツ)_/¯

neutrinoceros:
> we just never settled on what to use as the smoothing length, whether to grab a smoothed version of dx or to let the kd_tree generate it, which is either really slow but reasonably accurate (for lots of particles) or fast and really washed out (for few particles). I guess we could have both options available for the end user and describe the caveats of both methods?

I'm very late to the party but this sounds reasonable to me; if there is no sane default, it should be document why and the decision left to the user.

@chummels if you're still up for it, I think we can push this forward. For now I will refresh CI.

neutrinoceros:
Ah the close/reopen trick doesn't re-trigger most jobs because the branch is older than our migration to Github Actions.
Instead I will merge from main by resolving the only conflict here.

Since the conflict is only in import statements, I can resolve it without breaking anything by just accepting every line. Pre-commit + isort will remove duplicates and flag any unused import.

neutrinoceros:
pre-commit.ci autofix

PR #2189, opened 2019-03-06
[WIP] Added a render sampler for external lighting

keltonhalbert:
## PR Summary
So, it turns out that the Cython render code has a sample option for volume rendering a field under external lighting. I copied the VolumeRenderSampler python method and made a sampler for the lighting. I've marked this as WIP because I am unsure about how you all would want this implemented into the code base, but in this pull request, you can turn on the lighting render through the following code:
```python
sc = yt.create_scene(ds, 'qc')
source = sc[0]
...
cam = sc.camera
source.sampler_type = 'light-render'
source.set_sampler(cam)
```
Whether or not this is the correct way to expose this option to the user, I would like to leave up to the maintainers to discuss. Additionally, right now there isn't really user-exposed control about which direction the lighting is coming from. This could likely be an attribute added to the camera class and passed through to the sampler.

Below are two examples of the rendering, one with the lighting sampler and one with the default volume sampler. I have yet to fully dig into what's going on, and if the fidelity of the result can be improved.
yt-lightrender
yt-nolightrender

## PR Checklist


ngoldbaum:
Wow!

Ping @samskillman

MatthewTurk:
This seems like a reasonable way to select this rendering method.

ngoldbaum:
maybe call it `light_source_volume_render`?

This'll need docs and some tests before it can be merged. Let me know if you need a hand getting the tests passing.

keltonhalbert:
@ngoldbaum I'll sheepishly admit I don't have much experience when it comes to tests, but I'd be more than happy to do the work if you can point me in the right direction regarding what needs to be done. I'll fill in what documentation I can within the new methods created.

yt/visualization/volume_rendering/utils.py
```diff
@@ -74,6 +74,36 @@ def new_volume_render_sampler(camera, render_source):
...
+    if "camera_data" in params:
+        kwargs['camera_data'] = params['camera_data']
+    if render_source.zbuffer is not None:
+        kwargs['zbuffer'] = render_source.zbuffer.z
+        args[4][:] = np.reshape(render_source.zbuffer.rgba[:], \
+            (camera.resolution[0], camera.resolution[1], 4))
+    else:
+        kwargs['zbuffer'] = np.ones(params['image'].shape[:2], "float64")
+
+    sampler = LightSourceRenderSampler(*args, **kwargs, light_dir=[-1, 1, 1])
```
ngoldbaum: Right now the tests are failing because this isn't allowed on Python2. Instead you'll need to insert `light_dir` into `kwargs`.

Also I think it should be possible for users to set `light_dir` so we should probably think about how to expose that to users.

ngoldbaum:
For adding new tests, take a look at the existing volume rendering tests in `yt/visualization/volume_rendering/tests/`. There are also instructions on how to write tests and run the test suite in the documentation: http://yt-project.org/doc/developing/testing.html. 

Basically you'll be writing a python function that makes a volume rendering using your new sampler and checks to make sure that whatever you're doing doesn't crash the code and generates a correct answer. For things like volume rendering where the answer is an image it's sometimes not straightforward to write a test that checks if the answer is correct without directly comparing images, but if you can think of a way to do that that would be best. It's also possible to save reference images and compare with those using the answer testing infrastructure, some of the volume rendering tests do that. It's also ok (but not great) to just write smoke tests that make sure the code runs without crashing and not do any further verification to make sure the answer is correct or doesn't change.

If possible try to write your test using fake in-memory data (e.g. maybe take a look at `yt.testing.fake_vr_orientation_ds` as well as the tests that use it).

yt/visualization/volume_rendering/utils.py
ngoldbaum: It occurs to me that another way to do this would be to add a `light_direction` property to the `VolumeSource` that defaults to `None` and can be set to a 3-element array-like. If it's set you call the new sampler, if it's not set you call the regular volume rendering sampler.

If we did this I guess we might also need to add a light source to the mesh renderer and I don't know how straightforward that would be or if it even makes any sense. I guess you could also raise an error if it's set and you're rendering an unstructured mesh.

samskillman:
This is awesome. Glad that it was re-discovered and amazed it works :).

yt/visualization/volume_rendering/utils.py
keltonhalbert: I've made a local change I'll commit here in a bit, but what I ended up doing was setting the light_dir to the normal vector of the camera. This actually works out really well for most situations, so when I make the suggested changes and make it more user exposed, this will be what it defaults to instead of an arbitrary number. I should have some time over next weekend to work on this.

stonnes:
Hi @keltonhalbert I see that this has not been updated in a while!  Would you like some help getting this finished and merged?

neutrinoceros:
pre-commit.ci autofix

neutrinoceros:
I did some house cleaning (merge conflict resolution + lint), and CI seems to be green again, but we would still require new tests (and docs) in order to merge this.

@keltonhalbert, I am sorry this has lingered for so long. Do you wish to ever come back to it or should we look for someone else to take over ?

keltonhalbert:
Hi @neutrinoceros - sorry for the slow response! Unfortunately, I'm currently in the process of starting up new employment while also finishing up my thesis. This is not likely something I have the time to add to my plate right now, so please feel free to bring on somebody else to finish this up!

neutrinoceros:
Thank you for taking the time to respond, there is no urgency on our end. Good luck with your defense !

PR #2190, opened 2019-03-06
add support for exporting dataset to Firefly 

agurvich:
via reader = yt.visualization.firefly.create_firefly_object() and reader.dumpToJSON. See documentation for firefly_api for reader.options and reader.particleGroups functionality

Use `reader = yt.visualization.firefly.create_firefly_object()` to initialize a `firefly_api.reader.Reader` object containing data stored in a YTRegion and a subsequent `reader.dumpToJSON` to output the contents of that reader to JSON. 

See documentation for Firefly (www.github.com/ageller/Firefly/wiki) for `reader.options` and `reader.particleGroups` functionality between initializing the `reader` object and dumping to JSON (note this documentation will be duplicated with `firefly_api` as well soon!). 

## PR Summary

Adds support to export data to Firefly (www.github.com/ageller/Firefly) format for visualizing particle based data within a browser. Requires firefly_api (www.github.com/agurvich/firefly_api) as an extension, data must be output to a valid Firefly/data subdirectory (and so users will also have to clone their own copy of the full Firefly repository. 

## PR Checklist


agurvich:
firefly_api is available on [pypi](https://pypi.org/project/firefly-api/).

ngoldbaum:
Before this can be merged it’ll need tests and some coverage in the yt documenation. For the tests you’ll need to set up firefly in e.g. our Travis CI setup and then add a test to our test suite that runs if firefly is installed.

For the docs you’ll need to expose the new finction you’ve added in the yt API docs and probably also add some narrative documenation.

It might be more natural in the yt API to expose this functionality via a method on the base yt data object class. You could look at the `to_glue` method for an example where we export data to an external optional dependency.

I’d also guard the firefly imports in a try/except ImportError block to generate a nicer error message about how to install firefly or where to find installation instructions.

agurvich:
For some reason the example script I put in the docs wasn't showing up in the html I was building (but neither were any of the other python-scripts and it said `/Users/agurvich/research/repos/yt/doc/source/visualizing/volume_rendering.rst:859: WARNING: Unknown directive type "python-script"`. 

Though just having the text indented made it look right when I didn't have the python-script business there so I _think_ it's alright?

ngoldbaum:
So `python-script` is one of the sphinx extensions that executes code when the docs get built. The reason it's an issue for you is that you're building the docs with `READTHEDOCS=True` set in your environment.

That said `python-script` is only really useful for displaying plots and other images inline. Since your example doesn't generate an image there's no need to actually execute the script in the doc build. In that case I'd replace `.. python-script::` with `.. code-block:: python`, which will include a syntax highlighted version of the script in the docs.

yt/data_objects/tests/test_firefly.py
```diff
@@ -0,0 +1,20 @@
+import nose
+
+from yt.config import ytcfg
+from yt.testing import fake_particle_ds, requires_module
+
+import tempfile
+
+@requires_module('firefly_api')
```
ngoldbaum: You're going to need to add `firefly_api` to `tests/test_requirements.txt` so that it gets installed on the test machine.

agurvich: you mean avoiding the test completely doesn't count as passing it?? 

I've added firefly_api/0.0.2 to pypi (with my most recent changes) and put the module in test_requirements-- should be better now.

MatthewTurk: I dunno, I always thought there was some merit to passing tests that way ...

tape_fell_off

agurvich:
Does this look okay? Do I need to touch up the documentation at all?

yt/data_objects/data_containers.py
```diff
@@ -662,6 +663,182 @@ def to_glue(self, fields, label="yt", data_collection=None):
...
+            from firefly_api.errors import warnings as firefly_warnings
+        except ImportError:
+            raise ImportError("Can't find firefly_api, ensure it"+
+                "is in your python path or install it with"+
+                "'$ pip install firefly_api'. It is also available"+
+                "on github at github.com/agurvich/firefly_api")
+        try:
+            assert len(fields_units) == len(fields_to_include)
+        except AssertionError:
+            raise FireflyError("Each requested field must have units.")
```
ngoldbaum: I would phrase this try/except like this:

```
if len(field_units) != len(fields_to_include):
    raise RuntimeError("Each requested field must have units.")
```

Rather than catching the assertion you can just use an if statement and raise whatever error you want to raise if the if statement is not true. It's also  a little weird to raise an error from an outside package from inside yt, so rather than importing errors and warnings from firefly, just use either regular python errors and warnings, or if you really need custom errors, then use the ones from yt.

yt/data_objects/data_containers.py
```diff
@@ -662,6 +663,182 @@ def to_glue(self, fields, label="yt", data_collection=None):
...
+                ## read the available extra fields from yt
+                this_ptype_fields = self.ds.particle_fields_by_type[ptype]
+
+                ## load the extra fields
+                for field in this_ptype_fields:
+                    if 'position' in field or 'velocity' in field:
+                        continue
+                    if field not in fields_to_include:
+                        firefly_warnings.warn(FireflyWarning(
+                            'detected (but did not request) {} {}'.format(ptype,field)))
```
ngoldbaum: Here I'd use e.g. `mylog.warn` so the warnings get printed in yt's log output like other yt warning messages.

yt/data_objects/data_containers.py
```diff
@@ -662,6 +663,182 @@ def to_glue(self, fields, label="yt", data_collection=None):
                 pass
         else:
             data_collection.append(gdata)
+    
+    def create_firefly_object(
+        self,
+        path_to_firefly,
+        fields_to_include= [],
+        fields_units = [],
```
ngoldbaum: It's almost never correct to use a mutable as the default value of a keyword argument. Use `None` instead and initialize these to an empty list in the function definition. See e.g. [this blog post](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) which explains exactly why this is a bad idea.

ngoldbaum:
@agurvich I asked @munkm to look this over. If you can help her out with any installation or other issues that she runs into that would be great.

ngoldbaum:
One comment re: the docs, it might be nice if the firefly docs went into their own page separate from the rest of the volume rendering docs. Right now the firefly docs appear sort of in the middle of the volume rendering docs and I find that flow kind of confusing.

agurvich:
Thanks for the great feedback, I'll adjust accordingly. Happy to help @munkm ! Hopefully installation should be easy (just clone the git repository) and pip install firefly_api

MatthewTurk:
@ngoldbaum @agurvich Having it in the middle of the volren docs also downplays its distinction from the rest of yt, which does a disservice to the firefly developers.

chummels:
I fully agree with Nathan and Matt's statements.  Hopefully when I finish my PR on converting grid-data to particle-data, that will provide more content for these docs on exporting to firefly for all frontends.

yt/data_objects/data_containers.py
agurvich: Sure thing.

yt/data_objects/data_containers.py
agurvich: I see, I didn't realize that's how that was done.

yt/data_objects/data_containers.py
agurvich: i am bad about this, thanks for pointing it out :]

agurvich:
I went ahead and split the docs into their own page-- volume rendering was definitely the wrong place for them but perhaps an argument could be made for nesting them under Interactive Data Visualization. However, I'm happy to have it live separately on its own, as long as y'all are! 

I definitely agree, @chummels, that these should be revisited once your routine for casting grid datasets is finished. They are a bit sparse now-- I didn't want to repeat stuff that is in the Firefly documentation already but perhaps I can reproduce some of the basic stuff here at a later date and refer the user to our docs for more "advanced" features.

chummels:
Sounds like a good plan.  Just put links directly to the firefly docs in here for how to get things started on the firefly end maybe?  I'm happy to add to these docs when my PR is in a more mergible state, which will likely be after your PR is merged.

MatthewTurk:
Hi @agurvich!  This is looking really good -- anything I can do to help get it over the finish line?

agurvich:
@MatthewTurk I'm not sure-- I think @munkm was going to try it out and see if there were any problems but I haven't heard from her.

munkm:
Yep! That's right! Thanks for your patience @agurvich. 🙂

PR #2194, opened 2019-03-06
Deprecate smoothed fields

qobilidop:
## PR Summary

Since yt 4.0, the global octree for particle datasets no longer exists. So the notion of a smoothed field (onto a global octree) is deprecated. This PR will update the codebase and documentation to reflect this.

## PR Checklist

- [x] Update codebase
- [x] Update documentation


ngoldbaum:
Can we also add a helpful error message for when someone calls `ds.add_smoothed_particle_field`?

qobilidop:
@ngoldbaum Sure, that's on my todo list.

yt/fields/particle_fields.py
```diff
@@ -806,51 +806,10 @@ def _pfunc_avg(field, data):
...
-        rv /= hsml.uq**3 / hsml.uq.in_base(unit_system.name).uq**3
-        rv = data.apply_units(rv, field_units)
-        return rv
-    registry.add_field(field_name,
-                       sampling_type="cell",
-                       function = _vol_weight,
-                       validators = [ValidateSpatial(0)],
-                       units = field_units)
-    registry.find_dependencies((field_name,))
-    return [field_name]
```
qobilidop: I'm not sure if it's appropriate to remove the codes for a deprecated function. This is based on my observation of other deprecated functions. Let me know if that' not the case.

MatthewTurk: Since we are deprecating something that no longer works, rather than signaling that something will cease to work, I think it's okay.

yt/fields/particle_fields.py
```diff
@@ -806,51 +806,10 @@ def _pfunc_avg(field, data):
...
+    issue_deprecation_warning(
+        "This function is deprecated since yt-4.0, because the global octree"
+        " for particle datasets no longer exists."
```
ngoldbaum: I think this error message needs to do more to tell a user what they need to do instead. We know the name of the field they're trying to smooth because they passed it in, maybe we can suggest they just access that field instead?

qobilidop: I agree the warning here is not sufficient. I was thinking about leaving a concise description here, and link to the central yt3 to yt4 transition doc for a further explanation. But I'm not sure if it's a good idea to have a link in the warning. Probably it's not because links may break and this will become a maintenance burden. Or probably there's a tech solution to make the link robust. What do you think?

ngoldbaum: I think the link to that document should be pretty stable (assuming you use the https://yt-project.org/doc link, not a link for a specific version). There are a few other places where the docs are referenced in error messages and docstrings so it's not completely unprecedented.

yt/fields/particle_fields.py
```diff
@@ -803,54 +803,21 @@ def _pfunc_avg(field, data):
...
     return fn
 
+DEP_MSG_SMOOTH_FIELD = (
+    "The notion of a smoothed field no longer exists since yt-4.0, because the "
+    "global octree is removed. Now the SPH particle fields directly accessible "
+    "from disk are treated properly in the visualization machinaries. So there "
+    "is no need to add a separate smoothed field. "
+    "See this page to learn more:\n"
+    "https://yt-project.org/doc/yt4differences.html"
+)
```
qobilidop: @ngoldbaum Does the warning message look good to you now?

ngoldbaum: I think it's true to say that yt doesn't have smoothed fields at all, since SPH smoothing onto images, octrees, and grids is still a thing. Maybe just say that it's no longer necessary to add a field specifically for smoothing, as a particle field will be smoothed onto a data object as appropriate and field values will be returned at particle locations otherwise.

qobilidop: Thanks for the feedback. How about this:

Since yt-4.0, it's no longer necessary to add a field specifically for smoothing, because the global octree is removed. The old behavior of interpolating onto a grid structure can be recovered through data objects like `ds.arbitrary_grid`, `ds.covering_grid`, and most closely `ds.octree`. The visualization machinery now treats SPH fields properly by smoothing onto pixel locations. See this page to learn more: https://yt-project.org/doc/yt4differences.html

ngoldbaum: That's perfect :)

qobilidop:
This PR is ready for final review.

doc/source/analyzing/fields.rst
```diff
@@ -455,6 +455,8 @@ General Particle Fields
 Every particle will contain both a ``particle_position`` and ``particle_velocity``
 that tracks the position and velocity (respectively) in code units.
 
+.. FIXME: Update the following sections to reflect differences in yt-4.0.
```
ngoldbaum: Rather than leaving this comment here can you file an issue? That will be less likely to be missed.

qobilidop: Sure, will do.

qobilidop: It's tracked in #2218 now.

PR #2200, opened 2019-03-07
fix crash in check for ramses non-hilbert ordering

ngoldbaum:
This fixes a crash reported by @suniverse on yt's slack. I guess we could also make sure `bbox` is initialized to a default value before this line if that makes more sense but I'll defer to @cphyc on this.

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -530,7 +530,7 @@ class RAMSESDataset(Dataset):
                     dom, mi, ma = f.readline().split()
                     self.hilbert_indices[int(dom)] = (float(mi), float(ma))
 
-        if rheader['ordering type'] != 'hilbert' and self.bbox:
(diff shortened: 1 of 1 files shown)


yt/frontends/ramses/data_structures.py
```diff
@@ -530,7 +530,7 @@ def read_rhs(f, cast):
                     dom, mi, ma = f.readline().split()
                     self.hilbert_indices[int(dom)] = (float(mi), float(ma))
 
-        if rheader['ordering type'] != 'hilbert' and self.bbox:
+        if rheader['ordering type'] != 'hilbert' and hasattr(self, 'bbox'):
```
cphyc: The attribute is named `_bbox` and is always defined for a RAMSES dataset. I'd suggest the following
```suggestion
        if rheader['ordering type'] != 'hilbert' and self._bbox is not None:
```

ngoldbaum:
The failure on python2 is unrelated to this PR. Merging.

Issue #1021, opened 2015-05-11
Set buff size based on aspect ratio, dpi, and window size


yt-fido:
Originally reported by: **Matt Turk (Bitbucket: [MatthewTurk](https://bitbucket.org/MatthewTurk), GitHub: [MatthewTurk](https://github.com/MatthewTurk))**

----------------------------------------

Right now, it seems like our `buff_size` is always set identically to `(800,800)` in `SlicePlot` and the like.

This can be overcome by doing a `set_buff_size` call, but this call is also somewhat counterintuitive, as it is in reverse order (height, width).  I think, but am not certain, that we can get desired behavior by setting `buff_size` to `None` in the constructor, and then inside `_setup_plots` (where we'll have the aspect ratio info as well as the figure size, although we should probably assume 100 or 300 DPI) we can guess the buffer size.

(Thanks to @maxhutch for the pointer on this.)

----------------------------------------
- Bitbucket: https://bitbucket.org/yt_analysis/yt/issue/1021

ngoldbaum:
This is worth doing still. We should more intelligently choose the image size.

sheagk:
OK, it looks like I should edit `_recreate_frb` in `plot_window.py` to manage the `buff_size` intelligently based on aspect ratio and window size, since I believe both of those are accessible to the class.  Handling the dpi is trickier though, since that won't be specified until `save` (on a related note, can someone point me to the right `save` function?)  I could add an optional argument to `recreate_frb` to allow it to take in either the dpi or (more generally) mpl_kwargs.  Then, when calling `save`, it could check whether the dpi is specified.  If it is, then recreate the buffer with the right size (if necessary); if it isn't, then set the dpi in `save` based on the window size and `buff_size`.

However, all of this is made significantly more complicated by the fact that the buffer sometimes fills the image (i.e. if the colorbar and axes are turned off) and sometimes doesn't.  The former case is easy, but when the buffer doesn't fill the image, then you need to provide extra dots to show all the resolution elements, so you need a higher dpi for a given `buff_size`.  Exactly how much higher may not be trivial -- do you have access to quantities like axes.top and axes.left without first running _setup_plots?

ngoldbaum:
No, we don't create the matplotlib until we actually need to save something to disk or display it. The scale of the axes is known ahead of time (although I think right now we don't actually compute it until the plot is about to be created), see `ImagePlotMPL._get_best_layout`, which calculates the axes extents for the main plot axis and the colorbar axis.

PR #2203, opened 2019-03-12
expose buff_size argument to wrappers around PWViewerMPL in plot_window.py

sheagk:
Added `buff_size` argument (and added to docstring) for `ProjectionPlot`, `OffAxisSlicePlot`, `OffAxisProjectionPlot`, and `AxisAlignedSlicePlot`, which gets passed along to `PWViewerMPL`, which then gets passed on to `FixedResolutionBuffer`.

## PR Summary

Wasn't a way to create a higher resolution image with the `ProjectionPlot` command (at least, not that I could find).  Your refinement was effectively limited by the `buff_size`.

## PR Checklist

### DIFF
--- a/doc/source/cookbook/complex_plots.rst
+++ b/doc/source/cookbook/complex_plots.rst
@@ -48,17 +48,18 @@ of the four preceding classes.
 2. `figure_size`, which can be altered with either 
 :meth:`~yt.visualization.plot_container.PlotContainer.set_figure_size`
 or with :meth:`~yt.visualization.plot_container.PlotWindow.set_window_size`
-(which simply calls `set_figure_size`), or can be set during image
-creation with the `window_size` argument.  This sets the size of 
-the final image (including the visualization and, if applicable, the
-axes and colorbar as well) in inches.
+(the latter simply calls 
+:meth:`~yt.visualization.plot_container.PlotContainer.set_figure_size`), 
+or can be set during image creation with the `window_size` argument.  
+This sets the size of the final image (including the visualization and, 
+if applicable, the axes and colorbar as well) in inches.
 
 3. `dpi`, i.e. the dots-per-inch in your final file, which can also 
 be thought of as the actual resolution of your image.  This can
 only be set on save via the `mpl_kwargs` parameter to 
 :meth:`~yt.visualization.plot_container.PlotContainer.save`.  The
 `dpi` and `figure_size` together set the true resolution of your 
-image (final image will be :math:`dpi * figure_size` pixels on a
+image (final image will be `dpi` :math:`*` `figure_size` pixels on a
 side), so if these are set too low, then your `buff_size` will not
 matter.  On the other hand, increasing these without increasing
 `buff_size` accordingly will simply blow up your resolution 
@@ -72,11 +73,11 @@ underlying resolution of your simulation.  Regardless, for either
 grid data or deposited particle data, your image will never be 
 higher resolution than your simulation data.  In other words,
 if you are visualizing a region 50 kpc across that includes 
-data that reaches a resolution (or interparticle spacing) of
-100 pc, then there's no reason to set a `buff_size` (or a 
-:math:`dpi * figure_size`) above :math:`50 kpc/ 100 pc = 500`.
+data that reaches a resolution of 100 pc, then there's no reason 
+to set a `buff_size` (or a `dpi` :math:`*` `figure_size`) above 
+50 kpc/ 100 pc = 500.
 
-The included script demonstrates how each of these can be varied.
+The below script demonstrates how each of these can be varied.
 
 .. yt_cookbook:: image_resolution.py
 
--- a/doc/source/cookbook/complex_plots.rst
+++ b/doc/source/cookbook/complex_plots.rst
@@ -15,6 +15,72 @@ See :ref:`slice-plots` for more information.
 
 .. yt_cookbook:: multi_width_image.py
 
+.. _image-resolution-primer:
+
+Varying the resolution of an image
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This illustrates the various parameters that control the resolution 
+of an image, including the (deprecated) refinement level, the size of
+the :class:`~yt.visualization.fixed_resolution.FixedResolutionBuffer`,
+and the number of pixels in the output image.
+
+In brief, there are three parameters that control the final resolution, 
+with a fourth entering for particle data that is deposited onto a mesh
+(i.e. pre-4.0).  Those are:
+
+1. `buff_size`, which can be altered with 
+:meth:`~yt.visualization.plot_window.PlotWindow.set_buff_size`, which
+is inherited by 
+:class:`~yt.visualization.plot_window.AxisAlignedSlicePlot`,
+:class:`~yt.visualization.plot_window.OffAxisSlicePlot`,
+:class:`~yt.visualization.plot_window.ProjectionPlot`, and
+:class:`~yt.visualization.plot_window.OffAxisProjectionPlot`.  This
+controls the number of resolution elements in the 
+:class:`~yt.visualization.fixed_resolution.FixedResolutionBuffer`,
+which can be thought of as the number of individually colored 
+squares (on a side) in a 2D image.  `buff_size` can be set 
+after creating the image with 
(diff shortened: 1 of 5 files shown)


ngoldbaum:
For what it’s worth, we didn’t expose all of the options for the plots in the initializers on purpose. Would it be so bad to call `set_buff_size`? There’s also the `to_pw` method on the `FixedResolutionBuffer` class.

I’m -0 on this PR because exposing more options in the initializers makes the API more confusing.

ngoldbaum:
Also totally separate from my concerns, this needs tests and probably also narrative docs before it can be merged.

MatthewTurk:
I think if we want to reduce the number of arguments, the most elegant (in my mind) solution would be to use a `traitlets`-ified parameter that is "plot parameters" or something, and have the validation step convert a dictionary to that.  This would be similar to `plot_args` in the callbacks, but would also benefit from having strict validation of inputs.  We don't use `traitlets` anywhere else, but if we were just to spitball, that'd be my idea.

sheagk:
@ngoldbaum  I actually only happened upon the `set_buff_size` function after making this change, while trying to figure out what the appropriate tests to add are.  I'd personally argue that the resolution of the image is a fundamental enough quantity that it makes sense to be able to specify it in the main initialization, but I can understand the hesitation to over-expose the API.  I guess the preferred workflow is 

```
prj = yt.ProjectionPlot(...)
prj.set_buff_size(...)
prj.save(...)
```

Is that right?

As for the tests -- I'm happy to add whatever, just not sure where the appropriate spot would be.  Would that be `yt/visualization/tests/test_plotwindow.py`?  Should I create a new function that calls each of the 4 classes I changed with different buff_size arguments?

For the narrative docs, I'm not sure exactly what you mean...

ngoldbaum:
Also regarding the comment in the PR description about it not being possible to control the `buff_size` for `ProjectionPlot`, does that not include `ProjectionPlot.set_buff_size()`? Or were you not aware of that function? Or are you saying that there's no way to control that from the initializer, in which case I agree with you (but as I said above we set things up that way on purpose).

sheagk:
Yes, I wasn't aware of that function until after I'd already put the initial PR in.  I did find it once I searched for it directly, but missed it when reading through the tutorial docs.

ngoldbaum:
> Is that right?

Yes

> As for the tests -- I'm happy to add whatever, just not sure where the appropriate spot would be. Would that be yt/visualization/tests/test_plotwindow.py? Should I create a new function that calls each of the 4 classes I changed with different buff_size arguments?

Yes but ideally you would also be checking to make sure the image resolution is correct. One way to do this would be to make sure that the resulting images you get back from the `frb` attribute have the correct resolution:

```
plot = yt.ProjectionPlot(..., buff_size=(1600, 1600))
image = plot.frb['gas', 'density']
assert image.shape == (1600, 1600)
```

There's more documentation about running the tests and writing new ones in the developer guide: http://yt-project.org/doc/developing/testing.html

For the narrative docs I mean specifically the docs on this page: http://yt-project.org/doc/visualizing/plots.html

The source code for that page is here: https://github.com/yt-project/yt/blob/master/doc/source/visualizing/plots.rst

There's more information about exiting and building the docs in the developer guide: http://yt-project.org/doc/developing/building_the_docs.html

ngoldbaum:
Regarding Matt's comment yeah, I think exposing a way to control these attributes in a more general way would be nicer than exposing each option individually in the signature of the initializer. Not saying that you need to do what Matt is suggesting @sheagk since that might be a bigger task than you want to take on.

ngoldbaum:
Also if you can think of ways to make the existing behavior (e.g. needing to call `set_buff_size`) easier to understand in either the tutorials, the documentation, the cookbook, or anywhere else in the docs, that would be great. I've found the beginners usually have a much better idea of how the docs are deficient than more experienced users or developers like me.

sheagk:
@chummels I'm a little hesitant to use the word "pixels" because it's not really accurate.  I'll explain here why I was so confused about the resolution of the images I was getting out, and maybe that'll help elucidate a better way to explain all this in the docs:

Basically, there are 4 different parameters _that I know of_ (some of which are exposed and some of which aren't) that control the resolution of the final image in some way or another (at least for particle data in 3.5):

1. `n_ref`, which controls the max refinement in the octree (I see this is gone in `load_particles` in 4.0, but maybe there's a new equivalent parameter?  I don't know, but it would be useful to).  This essentially controls the blockiness of the figure, particularly when you're trying to visualize large volumes that might have very low resolution (e.g. the stars in a MW-size halo).  I assume this doesn't exist for grid data.
2. `buff_size`, which was the last one I found.  You could say that this is the number of "pixels" in the image, but it's not really.  I think a better term is the number of resolution elements in the image.
3 & 4. `window_size` (which is exposed in, e.g., `ProjectionPlot.__init__`, though it seems less important to me than `buff_size`) and the dpi you pass to matplotlib (or have set in your rcParams).  These control the actual number of pixels in the image, so if you set them too low, it won't matter how large your `buff_size` is or how small your `n_ref` is, your final image will be low resolution.  Set them too high, and you'll just stretch the resolution elements in the image to cover more pixels.

I understand the mentality of not overloading the user with too many options, but I hope you guys see why this can be a bit confusing when approaching it for the first time, particularly when `window_size` is exposed in the docstring of `ProjectionPlot`, `n_ref` is exposed in the docstring of `load_particles`, and `mpl_kwargs` is exposed in the docstring of `ProjectionPlot.save`, while `set_buff_size` is only mentioned once, about halfway down, in this huge writeup of _all_ things related plotting:  http://yt-project.org/doc/visualizing/plots.html.  I'd also argue again that the resolution of the image is one of the more important options, and is worth exposing directly.  Maybe there needs to be a separate writeup somewhere of how the four options (and maybe others?) above work together to set the resolution of your image? 

I also came across this old issue while googling for `set_buff_size` to see where it was mentioned in the docs:  https://github.com/yt-project/yt/issues/1021.  Maybe a better solution to the pull request I put in is to just implement that behavior?  If someone can point me to where `buff_size` would be initialized if it's passed as `None`, I can take a shot at implementing that.  Presumably the tests for that would be changing the window size, dpi, etc. and making sure you get the right `buff_size`?

ngoldbaum:
To make it even more confusing, matplotlib's DPI setting also comes into play. For what it's worth, the resolution of the plot saved to disk is the product of the matplotlib figure size (also called `window_size` in yt) and DPI. This is completely separate from the `buff_size`, which is the resolution of the image returned by the `FixedResolutionBuffer` associated with the plot.

I definitely agree that controlling the resolution of the final image is confusing and that the docs need to be improved to describe how it works. I would also say that it was a mistake to expose `window_size` in the initializer at all, I believe that was a carryover from `PlotCollection` (the old yt plotting interface that `PlotWindow` replaced).

I'm still -0 on exposing more options in the initializers because it makes the API more confusing (you need to go through docstrings that are already too long, this makes them even longer) and because it makes it harder to cover all possible use cases in the tests. That means I wouldn't do it myself but I'm not objecting to this being merged, particularly if others think this is a substantial improvement over the status quo.

chummels:
While I totally understand @ngoldbaum 's fear of having too many constructor arguments, I think this is a good exception because resolution is so important.  I would argue that the buff size (and the final image resolution) are fundamental and should be exposed in the constructor as this PR enables.  

I agree that there needs to potentially be more documentation beyond a simple docstring.  Perhaps a cookbook recipe showing its use (and the other knobs that can control the final image resolution) would make buff_size more discoverable and elucidate the complexity of this task.  I know every time I try to change the final res of my image, I have to do a bunch of experimenting with these knobs.

sheagk:
Just a heads up that I pushed a commit that includes a new cookbook recipe that discusses & illustrates the 4 parameters (now 3) that control the resolution of an image, along with docs for the `buff_size` parameter and tests that it's working correctly.

I will look to tackle the more programmatic way of setting `buff_size` at some point in the future, but I can't promise when/if that'll happen.

qobilidop:
About whether the `buff_size` parameter should be exposed, I could think of one important factor for it, which is efficiency. Changing `buff_size` could be time-consuming for large dataset. This is very different from other annotation and style type of parameters. So I support the idea of exposing `buff_size`. Or alternatively, an API with lazy-evaluation functionality could also work, but this might require much more work.

PR #2206, opened 2019-03-14
Sort .so input file list

bmwiedemann:
Sort .so input file list
so that EnzoHop.cpython-37m-x86_64-linux-gnu.so builds in a reproducible way
in spite of indeterministic filesystem readdir order
and http://bugs.python.org/issue30461

See https://reproducible-builds.org/ for why this is good.

Without this patch, building the openSUSE python-yt package always had variations in /usr/lib64/python3.7/site-packages/yt/analysis_modules/halo_finding/hop/EnzoHop.cpython-37m-x86_64-linux-gnu.so machine code.

This PR was done while working on reproducible builds for openSUSE.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -212,10 +212,10 @@ extensions = [
                "yt/analysis_modules/halo_finding/fof/kd.c"],
               libraries=std_libs),
     Extension("yt.analysis_modules.halo_finding.hop.EnzoHop",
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Fair enough. Thanks for the contribution and thank you for helping to package yt for downstream users.

PR #2208, opened 2019-03-19
Fix SPH answer tests in yt-4.0

jzuhone:
The `sph_answer` tests need updating for yt-4.0. This PR is a WIP to update them. Things that have to be done:

- [x] Remove deposit fields from testing in SPH frontends (except tipsy, which doesn't use `sph_answer`)
- [x] Use raw particle types to compute particle sizes
- [x] Fix projections for non-region objects

### DIFF
--- a/yt/frontends/gadget/tests/test_outputs.py
+++ b/yt/frontends/gadget/tests/test_outputs.py
@@ -50,10 +50,6 @@ iso_fields = OrderedDict(
         (("gas", "temperature"), None),
         (("gas", "temperature"), ('gas', 'density')),
         (('gas', 'velocity_magnitude'), None),
-        (("deposit", "all_density"), None),
-        (("deposit", "all_count"), None),
-        (("deposit", "all_cic"), None),
-        (("deposit", "PartType0_density"), None),
     ]
 )
 iso_kwargs = dict(bounding_box=[[-3, 3], [-3, 3], [-3, 3]])
--- a/yt/frontends/gizmo/tests/test_outputs.py
+++ b/yt/frontends/gizmo/tests/test_outputs.py
@@ -34,9 +34,6 @@ fields = OrderedDict(
         (("gas", "metallicity"), ('gas', 'density')),
         (("gas", "O_metallicity"), ('gas', 'density')),
         (('gas', 'velocity_magnitude'), None),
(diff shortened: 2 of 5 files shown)


jzuhone:
So these modifications to `sph_answer` are now working for every dataset which uses it except two: the FLASH particle dataset (which only tests deposit fields) and `TipsyGalaxy`, which is giving some strange errors like this:

```
RuntimeError: Error: yt attempted to read outside the boundaries of a non-periodic domain along dimension 1.
Region left edge = -2675.6919323730467 code_length, Region right edge = 2591.2012097167967 code_length
Dataset left edge = -557.3009106445312 code_length, Dataset right edge = 472.81018798828126 code_length
```

I suspect that this this has to do with the way the Tipsy datasets have to calculate their boundaries? I don't know much about this data at all, however.

ngoldbaum:
It's because that dataset doesn't have periodic boundary conditions and something in the test is causing a read outside the boundary. Should probably be looked into more.

jzuhone:
@ngoldbaum for this PR or separate?

ngoldbaum:
It doesn't really matter. If you want we can merge this now and come back to it in the context of the PR merging with master.

jzuhone:
Selfishly, I'd prefer the latter so we can also get the Arepo PR merged in.

ngoldbaum:
I don't think this needs to be merged before the arepo one.

jzuhone:
Well, the only way to make sure the Arepo tests pass is to merge this one first. No hurry of course.

ngoldbaum:
That would be true if we were running the SPH answer tests, which we aren't right now. This one only needs to be merged before we get rid of the yt-4.0 branch.

MatthewTurk:
@jzuhone I'm not entirely sure that I understand how to fix the issue of growing the domain, which I *think* is what's happening here.  Perhaps the simplest thing to do would be to -- for cases that are non-periodic and particle-based -- just say it's okay if we read outside the domain for particle fields.

MatthewTurk:
@jzuhone I think a check for the region to see if the `Index` type is a particle type and if the domain is non-periodic would be fine; just swallow the error in that case.

jzuhone:
This is ready to go, we can handle the Tipsy issue separately.

PR #2211, opened 2019-03-19
Changes default weight_field in PhasePlot to 'mass' Fixes #2210

### LINKED ISSUE #2210: ProfilePlot fails on particle-based data when no weight_field set
### Bug report

**Bug summary**

The default weight_field for a ProfilePlot is `cell_mass`, which obviously fails for particle-based datasets.

**Code for reproduction**

```python
import yt
p = yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), ('gas', 'mass'))
```

**Actual outcome**

```
  File "test.py", line 3, in <module>
    p = yt.PhasePlot(ds, ('gas', 'temperature'), ('gas', 'density'), ('gas', 'mass'))
  File "/Users/chummels/src/yt-conda/src/yt-git/yt/visualization/profile_plotter.py", line 933, in __init__
    fractional=fractional)
  File "/Users/chummels/src/yt-conda/src/yt-git/yt/data_objects/profiles.py", line 1047, in create_profile
    wf = data_source.ds._get_field_info(weight_field)
  File "/Users/chummels/src/yt-conda/src/yt-git/yt/data_objects/static_output.py", line 821, in _get_field_info
yt.utilities.exceptions.YTFieldNotFound: Could not find field '('nbody', 'cell_mass')' in snapshot_600.
```

**Expected outcome**

A ProfilePlot

chummels:
## PR Summary

Changes defaults in PhasePlot, Profiler, and ProfilePlot to use ('gas', 'mass') field instead of 'cell_mass' field.  ('gas', 'mass') field is aliased to 'cell_mass' for grid-based frontends, but also works with particle-based frontends.  This addresses Issue #2210.

## PR Checklist


yt/visualization/profile_plotter.py
```diff
@@ -1455,7 +1455,7 @@ def set_xlim(self, xmin=None, xmax=None):
 
         >>> import yt
         >>> ds = yt.load('IsolatedGalaxy/galaxy0030/galaxy0030')
-        >>> pp = yt.PhasePlot(ds.all_data(), 'density', 'temperature', 'cell_mass')
+        >>> pp = yt.PhasePlot(ds.all_data(), 'density', 'temperature', ('gas', 'mass'))
```
ngoldbaum: Here and below can you make all three of these tuples?

chummels: Sure, I can.  I was just trying to make this PR directly address a single issue.  But I can start slowly modifying things to be full-tuple field designation.

ngoldbaum: No need to do it everywhere if you don't want to but since you're already touching these lines I think it makes sense to update them in this PR.

ngoldbaum:
Looks like some code isn't handling field tuples correctly still (sigh...) if you need a hand fixing the unit tests just ping me.

chummels:
I updated all the files I touched to have the appropriate tuple field designation in the docstrings, but it makes this PR substantially longer.

chummels:
Unfortunately, I don't have any more time to work on this today.  I was trying to fix these bugs so I could actually make a simple PhasePlot for a research project and I figured I'd push it back to source.  But I actually need to get that done.  If you or someone else could try to address the failure of the unit tests, I would be most thankful.

chummels:
The primary issue here is that `save_as_dataset()` outputs fields without their ftype.  So when you save a dataset that has a field `('gas', 'density')` to disk, it writes it as `data/density` in the HDF5 file.  Then these tests are failing because they're looking for the `('gas', 'density')` field in the HDF5 file itself.  I can hardwire this to just search for the `'density'` field in the HDF5 file, but this seems like a larger problem.  If we really want to require full-tuples for fields, we should have some way to save them in the HDF5 files output by `save_as_dataset()`, and I'm not sure how to do this.  I think `h5py` (and the HDF5 standard) requires you to use strings explicitly and not tuples.

ngoldbaum:
ping @brittonsmith

chummels:
I changed the tests to explicitly set the weight field, which fixes the issue.  But I'm going to rescind my previous statement about needing `save_as_dataset()` to store fields to tuples in the HDF5 file.  Those previously-failing tests are not really a valid means of specifying a dataset:

```
        my_data = {
            'density': density,
            'temperature': temperature,
            'mass': mass}
        fake_ds_med = {"current_time": yt.YTQuantity(10, "Myr")}
        yt.save_as_dataset(fake_ds_med, "mydata.h5", my_data)
```

While it doesn't fail to create a dataset this way, I would argue that all we're doing is setting the fields according to a dictionary, which is pretty non-standard, as it doesn't use any of the standard functionality for inputting data (e.g., `load()`, `load_particles()`), which would avoid this error.  But I'm open to discussion on this.  It just seems the scope of this PR is getting extremely large when it was only meant to be a quick fix to a small problem.

ngoldbaum:
Sometimes making a seemingly small change can lead to bigger unintented changes elsewhere. Such is yt development.

I believe Britton explicitly wanted to make it possible to save dicts with `save_as_dataset` but I'll wait for him to chime in about this.

If this PR no longer exhibits the weird behavior you were seeing and if you still think it's a bug it would be awesome if you could open a separate issue to track that bug.

brittonsmith:
Hi, yes, saving arbitrary data via `yt.save_as_dataset` is a core feature of this functionality. Maybe I'm the only one, but I use it all the time. The default field type that data gets saved to is `data`, but this can be changed with the `field_types` keyword argument. @chummels, in the example you posted, you should be able to do:
```
field_types = {f:'gas' for f in my_data}
yt.save_as_dataset(fake_ds_med, "mydata.h5", my_data, field_types=field_types)
```

chummels:
Thanks for the comment.  That should fix it up.

chummels:
The tests fail because I have to match both ('gas', 'mass') as well as ('all', 'mass') between the saved dataset and the dictionary.  So I just changed the default behavior to use `'mass'` as the default weight_field.  This sidesteps that bug and makes it all inclusive for when the profiler operates on non-gas datasets.

ngoldbaum:
Thanks for getting the tests passing :)

yt/data_objects/tests/test_profiles.py
```diff
@@ -356,39 +356,39 @@ def tearDown(self):
...
         density = np.random.random(128)
         temperature = np.random.random(128)
-        cell_mass = np.random.random((128, 128))
+        mass = np.random.random((128, 128))
 
         my_data = {
-            "density": density,
-            "temperature": temperature,
-            "cell_mass": cell_mass}
+            ('gas', 'density'): density,
```
brittonsmith: This isn't going to result in the field type being saved as `gas`. It will still be `data` unless you pass the `field_types` keyword the way I mentioned in the earlier comment.

chummels: I know.  My last comment explained that I switched the behavior partially back on this.  The problem is that for some reason the existing tests need to match both `('all', 'mass')` as well as `('gas', 'mass')` in the HDF5 output.  It also made me realize that if I set the default weighting to `('gas', 'mass')`, `PhasePlot` doesn't really make sense for any non-gas field.  So I just changed it to `mass`, which not only fixes the tests but it also means that the weight field is ambiguous enough to go with the available `mass` field.

brittonsmith: Ok, but this code here doesn't do what it suggests so would be misleading to anyone reading it. The tuple as keys in `my_data` should be removed.

matthewturk:
Is this still a priority to discuss/explore?  If not, shall we close the PR?

jzuhone:
@chummels I'm going to work with you to get this through.

yt/data_objects/derived_quantities.py
```diff
@@ -135,7 +135,7 @@ class WeightedAverageQuantity(DerivedQuantity):
     >>> ad = ds.all_data()
     >>> print ad.quantities.weighted_average_quantity([("gas", "density"),
     ...                                                ("gas", "temperature")],
-    ...                                               ("gas", "cell_mass"))
+    ...                                               ("gas", "mass"))
```
cphyc, line 138:
```suggestion
    ...                                                ("gas", "mass"))
```

yt/data_objects/derived_quantities.py
```diff
@@ -391,7 +391,7 @@ class WeightedVariance(DerivedQuantity):
     >>> ad = ds.all_data()
     >>> print ad.quantities.weighted_variance([("gas", "density"),
     ...                                        ("gas", "temperature")],
-    ...                                       ("gas", "cell_mass"))
+    ...                                       ("gas", "mass"))
```
cphyc, line 394:
```suggestion
    ...                                        ("gas", "mass"))
```

yt/data_objects/profiles.py
```diff
@@ -966,7 +966,7 @@ def sanitize_field_tuple_keys(input_dict, data_source):
 
 def create_profile(data_source, bin_fields, fields, n_bins=64,
                    extrema=None, logs=None, units=None,
-                   weight_field="cell_mass",
+                   weight_field='mass',
```
cphyc, line 969: With respect to https://github.com/yt-project/yt/pull/3194 and https://github.com/yt-project/yt/pull/3101
```suggestion
                   weight_field=("gas", "mass"),
```

yt/data_objects/profiles.py
```diff
@@ -1001,7 +1001,7 @@ def create_profile(data_source, bin_fields, fields, n_bins=64,
     weight_field : str or tuple field identifier
         The weight field for computing weighted average for the profile
         values.  If None, the profile values are sums of the data in
-        each bin. Defaults to "cell_mass".
+        each bin. Defaults to 'mass'.
```
cphyc, line 1004:
```suggestion
        each bin. Defaults to ("gas", "mass").
```

yt/data_objects/tests/test_profiles.py
```diff
@@ -356,39 +356,39 @@ def tearDown(self):
...
+            ('gas', 'mass'): mass}
         fake_ds_med = {"current_time": yt.YTQuantity(10, "Myr")}
         yt.save_as_dataset(fake_ds_med, "mydata.h5", my_data)
 
         ds = yt.load('mydata.h5')
 
         assert_raises(
             YTProfileDataShape,
-            yt.PhasePlot, ds.data, 'temperature', 'density', 'cell_mass')
+            yt.PhasePlot, ds.data, 'temperature', 'density', 'mass')
```
cphyc, line 372: This will conflict with #3101 and #3194
```suggestion
            yt.PhasePlot, ds.data, ("gas", "temperature"), ("gas", "density"), ("gas", "mass"))
```

yt/data_objects/tests/test_profiles.py
```diff
@@ -356,39 +356,39 @@ def tearDown(self):
...
```
cphyc, line 391:
```suggestion
            yt.PhasePlot, ds.data, ("gas", "temperature"), ("gas", "density"), ("gas", "mass"))
```

yt/visualization/profile_plotter.py
```diff
@@ -136,48 +136,48 @@ def data_object_or_all_data(data_source):
...
         The binning field for the profile.
     y_fields : str or list
         The field or fields to be profiled.
     weight_field : str
-        The weight field for calculating weighted averages.  If None, 
+        The weight field for calculating weighted averages.  If None,
         the profile values are the sum of the field values within the bin.
         Otherwise, the values are a weighted average.
-        Default : "cell_mass".
+        Default : 'mass'
```
cphyc, line 164:
```suggestion
        Default : ("gas", "mass")
```

yt/visualization/profile_plotter.py
```diff
@@ -136,48 +136,48 @@ def data_object_or_all_data(data_source):
...
+        The data object to be profiled, such as all_data, region, or
         sphere. If a dataset is passed in instead, an all_data data object
         is generated internally from the dataset.
     x_field : str
         The binning field for the profile.
     y_fields : str or list
         The field or fields to be profiled.
     weight_field : str
-        The weight field for calculating weighted averages.  If None, 
+        The weight field for calculating weighted averages.  If None,
```
cphyc, line 161:
```suggestion
        The weight field for calculating weighted averages. If None,
```

yt/visualization/profile_plotter.py
```diff
@@ -234,7 +235,7 @@ class ProfilePlot(object):
     _plot_valid = False
 
     def __init__(self, data_source, x_field, y_fields,
-                 weight_field="cell_mass", n_bins=64,
+                 weight_field='mass', n_bins=64,
```
cphyc, line 238:
```suggestion
                 weight_field=("gas", "mass"), n_bins=64,
```

yt/visualization/profile_plotter.py
```diff
@@ -861,25 +870,25 @@ class PhasePlot(ImagePlotContainer):
     z_fields : str or list
         The field or fields to be profiled.
     weight_field : str
-        The weight field for calculating weighted averages.  If None, 
+        The weight field for calculating weighted averages.  If None,
         the profile values are the sum of the field values within the bin.
         Otherwise, the values are a weighted average.
-        Default : "cell_mass".
+        Default : 'mass'
```
cphyc, line 876:
```suggestion
        Default : ("gas", "mass")
```

yt/visualization/profile_plotter.py
```diff
@@ -1164,11 +1173,13 @@ def from_profile(cls, profile, fontsize=18, figure_size=8.0):
         >>> extrema = {
         ... 'density': (1e-31, 1e-24),
         ... 'temperature': (1e1, 1e8),
-        ... 'cell_mass': (1e-6, 1e-1),
+        ... ('gas', 'mass'): (1e-6, 1e-1),
         ... }
```
cphyc, line 1177: Should it be instead
```python
        >>> extrema = {
        ...     ("gas", "density"): (1e-31, 1e-24),
        ...     ("gas", "temperature"): (1e1, 1e8),
        ...     ("gas", "mass"): (1e-6, 1e-1),
        ... }
```

cphyc (commented):
I left a few hints to make sure the PR is compatible with #3101 and #3194.

yt/data_objects/derived_quantities.py
jzuhone: I resolved this one when I fixed confilcts with `main`.

yt/data_objects/profiles.py
jzuhone: I'm going to have to come back to this because @chummels felt the need to not use a tuple here because of conflicts with `yt.save_as_dataset`, see the conversation above.

yt/data_objects/profiles.py
jzuhone: Same as above.

yt/data_objects/tests/test_profiles.py
jzuhone: Still think this is going to cause the yt data tests to fail.

yt/visualization/profile_plotter.py
jzuhone: yep

jzuhone:
@chummels ok if I take this over? I think I can resolve it quickly.

chummels:
Sure @jzuhone , thanks for the offer!  If you want to patch this up, then feel free to do so.  Can I just review it all before it gets merged?  I realize a lot has changed under the hood in the last couple of years since these issues were originally encountered.  It's a bit challenging to keep all the threads together to ensure the end result makes sense.

jzuhone:
@chummels have a look, tests pass now

yt/data_objects/tests/test_profiles.py
jzuhone: Is this conversation resolved?

yt/data_objects/profiles.py
jzuhone: I made this change.

yt/visualization/profile_plotter.py
jzuhone: I made this change

yt/visualization/profile_plotter.py
jzuhone: I made this change

yt/visualization/profile_plotter.py
jzuhone: I made this change

yt/data_objects/tests/test_profiles.py
jzuhone: This has been changed.

yt/data_objects/profiles.py
jzuhone: This was done.

yt/data_objects/tests/test_profiles.py
jzuhone: This was done.

jzuhone:
This needs review again since it's been so long, but it's pretty straightforward and so it shouldn't be too bad.

brittonsmith (approved):

munkm:
I'll review it this morning!

munkm (commented):
The changes here are looking good! 

I have a few questions about other files to update though. Should we also update
- `yt/visualization/tests/test_profile_plots.py`
- narrative docs in the `docs` dir of PhasePlot, profile plot, etc. 
- and maybe a note in the yt4 differences so grid based users might see a different default but their scripts shouldn't be affected by this?

jzuhone:
@munkm changing `yt/visualization/tests/test_profile_plots.py` screws up the answer tests due to the old answers being indexed by `"cell_mass"` instead of mass. Thoughts on how to deal with that?

jzuhone:
Ok, so I've implemented @munkm's suggestions, but now we have to sort out what to do about the answer tests failing due to the key change.

jzuhone:
@Xarthisius the answer tests now fail because I changed the field from `mass` to `cell_mass` and they were expecting a `cell_mass` key. It's obviously easy just to bump the answer number, but do we have a way to check to make sure the values are the same just in case?

Xarthisius:
> @Xarthisius the answer tests now fail because I changed the field from `mass` to `cell_mass` and they were expecting a `cell_mass` key. It's obviously easy just to bump the answer number, but do we have a way to check to make sure the values are the same just in case?

For nose we don't. We can only manually inspect the shelve files before and after. For pytest it's gonna be a lot easier, but those tests are unfortunately not ported yet.

jzuhone:
> For nose we don't. We can only manually inspect the shelve files before and after. 

Do we have docs on how to do this?

Xarthisius:
> > For nose we don't. We can only manually inspect the shelve files before and after.
> 
> Do we have docs on how to do this?

I'm not aware of any docs on that matter. Here's roughly how can you access it:

```
>>> import shelve
>>> fp = shelve.open("answer-store/py37_profile_plots/py37_profile_plots")
>>> list(fp.keys())
["odict_keys([('gas', 'velocity_x')])_3_UniformGridData", "odict_keys([('gas', 'cell_mass')])_21_UniformGridData", "odict_keys([('gas', 'velocity_x')])_18_UniformGridData", "odict_keys([('gas', 'temperature')])_17_UniformGridData", "odict_keys([('index', 'radius')])_24_UniformGridData", "odict_keys([('gas', 'velocity_x')])_19_UniformGridData", "odict_keys([('gas', 'temperature')])_0_UniformGridData", "odict_keys([('gas', 'temperature')])_1_UniformGridData", "odict_keys([('gas', 'temperature')])_2_UniformGridData", "odict_keys([('index', 'radius')])_25_UniformGridData", "odict_keys([('gas', 'cell_mass')])_28_UniformGridData", "odict_keys([('gas', 'temperature')])_15_UniformGridData", "odict_keys([('index', 'radius')])_26_UniformGridData", "odict_keys([('gas', 'velocity_x')])_20_UniformGridData", "odict_keys([('gas', 'cell_mass')])_6_UniformGridData", "odict_keys([('gas', 'cell_mass')])_13_UniformGridData", "odict_keys([('gas', 'cell_mass')])_7_UniformGridData", "odict_keys([('gas', 'cell_mass')])_12_UniformGridData", "odict_keys([('index', 'radius')])_10_UniformGridData", "odict_keys([('gas', 'cell_mass')])_23_UniformGridData", "odict_keys([('gas', 'velocity_x')])_4_UniformGridData", "odict_keys([('gas', 'cell_mass')])_8_UniformGridData", "odict_keys([('index', 'radius')])_11_UniformGridData", "odict_keys([('gas', 'temperature')])_16_UniformGridData", "odict_keys([('gas', 'cell_mass')])_29_UniformGridData", "odict_keys([('index', 'radius')])_9_UniformGridData", "odict_keys([('gas', 'cell_mass')])_27_UniformGridData", "odict_keys([('gas', 'temperature'), ('gas', 'velocity_x')])_30_UniformGridData", "odict_keys([('gas', 'cell_mass')])_14_UniformGridData", "odict_keys([('gas', 'velocity_x')])_5_UniformGridData", "odict_keys([('gas', 'cell_mass')])_22_UniformGridData"]
```

jzuhone:
@Xarthisius I ran this:

```
 nosetests --with-answer-testing --local --local-dir . --answer-name=main --answer-store --attr=answer_test test_profile_plots.py
```
to generate the answers but `shelve` doesn't know what to do with the `main.db` file it generated in the `main` directory. I'm assuming I am doing something wrong.

chummels (commented):
Thanks for finishing this one off, @jzuhone !  Great work!

munkm (approved):

PR #2213, opened 2019-03-19
Enabling PhasePlots to have log axis ranges by default Fixes #2212

### LINKED ISSUE #2212: Default PhasePlot Behavior Different for Particle and Grid Datasets
### Bug report

**Bug summary**

When one creates a PhasePlot with a grid-based dataset, it defaults to using log-based axis ranges.  When one does the same script with a particle-based dataset, it uses linear axis ranges, which is usually not desired.

**Code for reproduction**

```python
import yt
yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), ('gas', 'mass'), weight_field=('gas', 'mass')).save()
yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), ('gas', 'mass'), weight_field=('gas', 'mass')).save()
```

**Actual outcome**

galaxy0030_2d-Profile_density_temperature_mass

snapshot_600_2d-Profile_density_temperature_mass

**Expected outcome**

For both PhasePlots to be in log-space.

**Version Information**
Tip of yt-4.0: f5975b0248c9c1410992e944d97b869d90ea37c9

chummels:
## PR Summary

Enables PhasePlots to have log-based axis ranges by default, just like grid-based datasets.  Fixes #2212.

Now when you run the script from that issue:

```
import yt
yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), ('gas', 'mass'), weight_field=('gas', 'mass')).save()
yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), ('gas', 'mass'), weight_field=('gas', 'mass')).save()
```

galaxy0030_2d-Profile_density_temperature_mass

snapshot_600_2d-Profile_density_temperature_mass

## PR Checklist


chummels:
I *think* this may have been put in place to make it so ParticlePhasePlots were in linear space by default?  I'm not sure.  But I know that has been the default behavior, but I think it is misguided.  I think most people using the PhasePlot will want it to operate in log space by default.

http://yt-project.org/docs/dev/visualizing/plots.html#particle-plots

MatthewTurk:
If this goes in, I think we should bookmark it to come back to when fields can specify if they should be logged or not, which right now is unused.

chummels:
That's fine by me.  I'm just trying to create similar behavior for particle-based frontends as grid-based frontends for this PR.

chummels:
Notably, this changes the default behavior for which fields get scaled as "log" and which fields get scaled as "linear" in Profiles and PhasePlots.  Up until now, the distinction has been effectively that grid-based fields were scaled as log, and particle-based fields were scaled as linear.  This made sense back when yt was grid-centric, and particled-based fields were primarily for N-body particles where you were plotting things like position and velocity in ParticlePlots: 

http://yt-project.org/docs/dev/visualizing/plots.html#particle-plots

But now that we're treating particle-based datasets natively, the distinction seems problematic, as particle- and grid-based datasets are getting treated differently in things like PhasePlots as noted in Issue #2212 .  This PR changes the behavior so that *all* coordinate and velocity fields are scaled as linear, and everything else is scaled as log, regardless of grid or particle origin.

As an example, here is a script much like the one in the link above looking at the positions and velocities of both grid-based and particle-based fields associated with the IsolatedGalaxy dataset:

```
import yt
p = yt.PhasePlot(ds, ('all', 'particle_position_x'), ('all', 'particle_velocity_z'), ('all', 'particle_mass'), weight_field=None)
p.set_unit(('all', 'particle_position_x'), 'Mpc')
p.set_unit(('all', 'particle_velocity_z'), 'km/s')
p.set_unit(('all', 'particle_mass'), 'Msun')
p.save()
p = yt.PhasePlot(ds, ('gas', 'x'), ('gas', 'velocity_z'), ('gas', 'mass'), weight_field=None)
p.set_unit(('gas', 'x'), 'Mpc')
p.set_unit(('gas', 'velocity_z'), 'km/s')
p.set_unit(('gas', 'mass'), 'Msun')
p.save()
```

galaxy0030_2d-Profile_particle_position_x_particle_velocity_z_particle_mass

galaxy0030_2d-Profile_x_velocity_z_mass

I think this is the preferred behavior, and the second plot was not linear prior to this PR. I'm happy to discuss the merits of this method.

chummels:
Unsurprisingly, the answer tests failed because of this change in behavior for Profiles moving to linear scaling for coordinate and velocity fields.

yt/fields/field_info_container.py
```diff
@@ -397,4 +397,19 @@ def check_derived_fields(self, fields_to_check = None):
             mylog.debug("Succeeded with %s (needs %s)", field, fd.requested)
         dfl = set(self.ds.derived_field_list).union(deps.keys())
         self.ds.derived_field_list = list(sorted(dfl, key=tupleize))
+        self.set_linear_fields()
         return deps, unavailable
+
+    def set_linear_fields(self):
```
ngoldbaum: Can you make this private by renaming this function to be `_set_linear_fields`?

yt/fields/field_info_container.py
```diff
@@ -397,4 +397,19 @@ def check_derived_fields(self, fields_to_check = None):
             mylog.debug("Succeeded with %s (needs %s)", field, fd.requested)
         dfl = set(self.ds.derived_field_list).union(deps.keys())
         self.ds.derived_field_list = list(sorted(dfl, key=tupleize))
+        self.set_linear_fields()
```
chummels: I wasn't sure if this was the best place to call this function.  But it seemed to work.  I just want to assure it gets called when a new dataset is loaded and after all of the fields are defined.

MatthewTurk:
Once the tests are passing this looks good to me!

chummels:
I have held up this PR because I'm trying to find an additional solution to the current problem.  Right now, if you try to generate a `PhasePlot` for a field that has any negative values in it, yt barfs because it tries to put it in log-space.  Even if you generate it, and then attempt to change it to linear space with one of the callbacks, it still barfs because it defaults to generating it in log space during the `create_profile()` call embedded in the first steps of `PhasePlot`.

My solution in the current PR is to set a bunch of fields that commonly have negative values (e.g., the x,y,z, coordinates and velocities) to be put in linear space by default.  But it doesn't *really* address the main problem, which is that any field with a negative value will cause PhasePlot to die.  I've been searching for a way to make `PhasePlot` accept a kwarg for the linear/log behavior of a field in its initial call so as to avoid this problem, but I've been getting bogged down in all of the conditional behavior in the PhasePlot code.  But perhaps someone else knows of a way to make this work that I just haven't stumbled upon?

That's the main reason for the hold up here.  I figured that such code would complete this PR and then I'd regenerate the answer tests for final acceptance.  If anyone has another solution, though, I'm all ears.

MatthewTurk:
@chummels It might be possible to use the symlog normalizer to generate a bunch of bins as input to the histogram, then use symlog on the opposite end.  There's still the free parameter, though, of where to do log/lin cutoff.

chummels:
OK, that's interesting.  I can try to check out the symlog normalizer as the safeguard to prevent it from dying.  In my opinion, coordinates and velocities should be linear and everything else should be log by default.  But the coordinates/velocities I check in this PR are just the cartesian ones, as opposed to cylindrical and spherical.  I could potentially add those too if people think this is a good idea.

MatthewTurk:
@chummels Sorry, what I meant is, if you're looking for a way to generate bins that are both lin and log, that might be a way to do it.  Feel free to ignore.

neutrinoceros:
Adding the 4.0 label since it's present on the issue this is solving.

chummels:
@neutrinoceros I'm confused.  I don't think this is a backwards incompatible change, nor is it on Issue #2212 .

matthewturk:
Is this still a priority, and if not, shall we close the PR?

jzuhone:
@chummels I'm going to work with you to get this through.

yt/fields/field_info_container.py
```diff
@@ -397,4 +397,19 @@ def check_derived_fields(self, fields_to_check = None):
             mylog.debug("Succeeded with %s (needs %s)", field, fd.requested)
         dfl = set(self.ds.derived_field_list).union(deps.keys())
         self.ds.derived_field_list = list(sorted(dfl, key=tupleize))
+        self._set_linear_fields()
         return deps, unavailable
+
+    def _set_linear_fields(self):
```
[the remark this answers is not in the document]
munkm: Maybe we could have an issue to remind ourselves int he future?? I think this is a good idea!

jzuhone:
@chummels want me to pick this one up as well?

chummels:
I figured there would be failing answer tests, because I'm changing the default behavior of particle-based PhasePlots.  But in looking at the failing tests, I am having a hard time deciphering what is going on.  When I look at the failing: `Build and Test / build (ubuntu-18.04, 3.7, full, answer) (pull_request) Failing after 10m — build (ubuntu-18.04, 3.7, full, answer)
`, it just shows a ton of missing output files for 92 tests (`_gdbm.error: [Errno 2] No such file or directory`).  I don't understand why they're all missing.

munkm:
Cameron it looks like you deleted the answer store submodule somehow. Can you try to point to the most recent commit in the answer store submodule and see if the tests pass?

chummels:
Oh shoot.  I guess when I did the merge, somehow the answer_store submodule didn't get included.  I'm not exactly sure how to bring it back.

neutrinoceros:
@chummels I checked and only your most recent commit (3d45d1ea89db034c33bad4d233e315865ece1481) is problematic.
I think this should do the trick:
```
git checkout 3d45d1ea89db034c33bad4d233e315865ece1481
git reset HEAD~
git add yt/data_objects/profiles.py
git commit -m "Ensuring particle phaseplots default to log axis scaling"
git push --force
```

chummels:
Many thanks, @neutrinoceros !  I think what happened is that I didn't have answer_store present locally, and somehow inadvertently committed the lack of that directory as a deletion in the repo.  My mistake.  Thanks, for all of the help to you and to @munkm !  Now we'll see which tests fail instead of all of them. :)

neutrinoceros:
Now I suppose we'll need to update the answers again, right ?

neutrinoceros:
Also note that the pre-commit bot doesn't auto fix anymore.
You can still run it locally though

Right now you'd need this
```
pre-commit run black --all-files
```
but it will take a few minutes... it's much shorter to have the hooks run on every commit (`pre-commit install`).
You can also run it here with the magic comment (see https://pre-commit.ci)

chummels:
Thanks for displaying one of the failed test results.  I would argue that the changed behavior is desired, but I guess I'm open to others' opinion on this.  I made it so position and velocity data is plotted in linear space instead of log space, and this plot is consistent with that shift.  I can regress if others think this is a bad idea, though.

neutrinoceros:
For the record I don't have an opinion on this. I'm fine with calling this a bug or a feature (as long as it's documented)

jzuhone:
@chummels this is fine by me--let's note it as a feature.

matthewturk:
I think this is fine.  I've looked at the difference in the images, and I think we should update the answer tests -- assuming the one John linked is indeed representative.

(This will affect profile plots because they follow the same code path)

yt/fields/field_info_container.py
```diff
@@ -667,4 +667,19 @@ def check_derived_fields(self, fields_to_check=None):
...
+        PhasePlots. Default for all fields is set to log, so this sets which
+        are linear.  For now, set linear to geometric fields: position and
+        velocity coordinates.
+        """
+        log_prefixes = ['', 'velocity_', 'particle_position_', 'particle_velocity_']
+        coords = ['x', 'y', 'z']
+        log_fields = [prefix+coord for prefix in log_prefixes for coord in coords]
+        for field in self.ds.derived_field_list:
+            if field[1] in log_fields:
+                self[field].take_log = False
```
cphyc, line 685: The variable got me confused. Shouldn't it be instead

```suggestion
        non_log_prefixes = ('', 'velocity_', 'particle_position_', 'particle_velocity_')
        coords = ('x', 'y', 'z')
        non_log_fields = [prefix+coord for prefix in log_prefixes for coord in coords]
        for field in self.ds.derived_field_list:
            if field[1] in non_log_fields:
                self[field].take_log = False
```

cphyc (approved):

neutrinoceros:
Beware that there is currently a diff in two answer tests that appeared 2 days ago when matplotlib 3.4.2 was released, so until it is fixed we should avoid updating answers.

jzuhone (approved):

jzuhone:
@neutrinoceros @chummels so what do we need to do to get this across the finish line?

neutrinoceros:
@jzuhone

1) [ ] fix the 2 failing tests that were broken by the latest Matplotlib version (or pin mpl < 3.4.2)
2) [ ] rerun the answer workflow here
3) [ ] submit a PR to the answer store repo with the result
4) [ ] merge there
5) [ ] bump the target commit of the submodule here

chummels:
I guess I also need to document the change in behavior in the yt4 change docs.  But I'm waiting for #3232 to merge.

jzuhone:
@chummels I approved and merged PR #3232.

neutrinoceros:
Also there are still real failures on Jenkins (NameError), and it's also reported by the pre-commit.ci bot (flake8)

chummels:
As far as I can tell, the only tests failing now are the answer tests of images not matching up.  But I'm not sure how to go through each of those and actually look at the image differences to ensure that they are all desired changes.

neutrinoceros:
@chummels go to the answer test job log, at the end of it (second to last stage of the workflow) you'll find 2 links, one of which will download an html report for you to inspect !

chummels:
OK, thank you for the help.  Yes, I have looked over all of the answer tests, and they all seem to be an improvement over the previous values.  The example that @neutrinoceros posted earlier is representative of their overall behavior.

neutrinoceros:
The other link is a patch to commit to the answer store (target the 4.0 branch there)

neutrinoceros:
update: got bored of the GitHub action outage. Submitted the required PR to the answer store https://github.com/yt-project/answer-store/pull/22
Hope you don't mind @chummels :)

chummels:
You are a champion--thank you!

neutrinoceros:
One last step to check @chummels ! We're getting there

chummels:
Thanks for the help and patience, everyone!

neutrinoceros:
Congrats ! Nice work, team 👏

Issue #2209, opened 2019-03-19
Cannot make projections of SPH data using non-region sources in yt-4.0


jzuhone:
### Bug report

**Bug summary**

The `sph_answer` test executes `PixelizedProjectionValuesTest` tests, which use a sphere as a selection object. This currently does not work for SPH data in yt-4.0. One gets error messages like this:

```
======================================================================
ERROR: PixelizedProjectionValues_snap_505_sphere_('c', (0_1, 'unitary'))_('gas', 'temperature')_2_('gas', 'density')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jzuhone/miniconda3/lib/python3.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/jzuhone/Source/yt/yt/utilities/answer_testing/framework.py", line 354, in __call__
    nv = self.run()
  File "/Users/jzuhone/Source/yt/yt/utilities/answer_testing/framework.py", line 580, in run
    frb[self.field]
  File "/Users/jzuhone/Source/yt/yt/visualization/fixed_resolution.py", line 135, in __getitem__
    int(self.antialias))
  File "/Users/jzuhone/Source/yt/yt/geometry/coordinates/cartesian_coordinates.py", line 205, in pixelize
    antialias, dimension, periodic)
  File "/Users/jzuhone/Source/yt/yt/geometry/coordinates/cartesian_coordinates.py", line 300, in _ortho_pixelize
    le = data_source.data_source.left_edge.in_units('code_length')
AttributeError: 'YTSphere' object has no attribute 'left_edge'

======================================================================
ERROR: FieldValues_snap_505_sphere_('c', (0_1, 'unitary'))_('gas', 'temperature')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jzuhone/miniconda3/lib/python3.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/jzuhone/Source/yt/yt/utilities/answer_testing/framework.py", line 354, in __call__
    nv = self.run()
  File "/Users/jzuhone/Source/yt/yt/utilities/answer_testing/framework.py", line 471, in run
    field, weight=weight_field)
  File "/Users/jzuhone/Source/yt/yt/data_objects/derived_quantities.py", line 147, in __call__
    rv = super(WeightedAverageQuantity, self).__call__(fields, weight)
[...]

PR #2215, opened 2019-03-20
Code to find object bounding boxes for SPH projections.

jzuhone:
This PR provides a function to compute a minimum bounding box for generic data containers. Most objects which would be relevant for creating projections of SPH (and other) data are included. 

Closes Issue #2209. 

A test has been added.


jzuhone:
Example here: https://gist.github.com/d46f42f6ba0dff88fffe8083256763ee

yt/geometry/coordinates/cartesian_coordinates.py
```diff
@@ -36,6 +36,71 @@
 from yt.utilities.nodal_data_utils import get_nodal_data
 from yt.units.yt_array import uconcatenate
 
+
+def _find_object_bbox(data_source):
```
ngoldbaum: I think it would be cleaner to make this a method of `YTSelectionContainer3D` that subclasses implement. That makes it clearer for implementors of future data objects that they'd need to touch this code.

jzuhone: I agree!

jzuhone:
For some reason this doesn't work with ellipsoids--you just get the whole projection back.

yt/geometry/coordinates/cartesian_coordinates.py
```diff
@@ -36,6 +36,71 @@
...
+    """
+    This function determines the minimum bounding box
+    of a data container.
+    """
+    if hasattr(data_source, "base_object"):
+        # This a cut region so we'll figure out
```
chummels: minor nit: "This is a"

yt/geometry/coordinates/cartesian_coordinates.py
```diff
@@ -72,28 +137,28 @@ def _sample_ray(ray, npoints, field):
...
+                                           field].in_base("code"))
 
-        for chunk in data.all_data().chunks([], "io"):
-            for field in fields:
-                field_data[field].append(chunk[ptype,
-                                               field].in_base("code"))
+    for field in fields:
+        field_data[field] = uconcatenate(field_data[field])
 
+    if kdtree is True:
```
chummels: Why does this drop out of the chunk loop on line 145

jzuhone: Unfortunately the diff is not doing this justice. What happened in this `all_data` function is that it was indented 8 spaces to the right instead of 4 at the first level. I just moved it back. I'd check [this](https://github.com/yt-project/yt/blob/d281976893517adab229fd5c5856f256abf16b3e/yt/geometry/coordinates/cartesian_coordinates.py#L139-L161) instead of the diff to see what happened.

ngoldbaum: Yeah, looks like this was missed in code review from one of Ash's GSOC pull requests.

yt/geometry/coordinates/cartesian_coordinates.py
jzuhone: One other thing though--these rules only really work for cartesian datasets. I guess there's nothing we can do about that now?

ngoldbaum: Mayve check the geometry of the dataset and raise a NotImplementedError in the base class implementation? Right now we’re only using this for SPH data and non-cartesian SPH data is not something we’ve thought about at all so in practice that error won’t be hit unless someone tries to repurpose this for ither things in the future, at which point we can come back to this.

MatthewTurk:
I genuinely don't understand why the sphere selection (for instance) doesn't work as-is. Why is it that we can't take the data source and access the particles that belong to it? Is it an issue of not getting the neighboring particles? I guess what I mean is, why is a new get bbox necessary as opposed to using the existing selector? I see that there's a left edge query in the traceback but is that really necessary?

jzuhone:
@MatthewTurk this is above my pay grade.

ngoldbaum:
@matthewturk does my explanation in this comment make sense:

https://github.com/yt-project/yt/issues/2209#issuecomment-474468817

This avoids doing I/O unnecesarily when we are zoomed very far into a slice or porjection. Uptimately this is needed because generating images for SPH slices and priojections aren’t cheap like grid slices and projections.

MatthewTurk:
Ah, I see.

This is probably also why specifying buff size in the constructor is necessary, especially for SPH - the viz and the display are linked in a way they aren't for mesh datasets.

jzuhone:
Assuming the tests pass, this is ready to go now.

ngoldbaum:
I think `YTBooleanContainer` also needs an implementation, which I guess also might be complicated to implement. I guess a conservative way to do it would be to calculate the bounding boxes for `dobj1` and `dobj2` and then calculate the bounding box for the union of those two boxes.

PR #2219, opened 2019-03-26
Unyt

ngoldbaum:
Make yt use unyt.

WIP for now until I get the tests to pass.


ngoldbaum:
All the tests are passing but it looks like there's been a big performance drop. I'm investigating....

doc/source/yt4differences.rst
```diff
@@ -166,15 +166,112 @@ It should be noted that the
...
+``ds.units``
+~~~~~~~~~~~~
+
+Each dataset now has a set of unit symbols and physical constants associated
+with it, allowing easier customization and smoother interaction, especially in
+workflows that need to use code units or cosmological units. The ``ds.units``
+object has a large number of attributes corresponding to the names of units and
+physical constants. All units known to the dataset will be available, including
+custom units. In situations where you might have used ``ds.arr`` or ``ds.quan``
+before, you can now safely use ``ds.units``:
```
MatthewTurk: 👍

yt/data_objects/construction_data_containers.py
```diff
@@ -2183,8 +2183,8 @@ def __init__(self, left_edge=None, right_edge=None, n_ref=32, over_refine_factor
         YTSelectionContainer3D.__init__(self,
             center, ds, field_parameters)
 
-        self.left_edge = self._sanitize_edge(left_edge)
-        self.right_edge = self._sanitize_edge(right_edge)
+        self.left_edge = self._sanitize_edge(left_edge, ds.domain_left_edge)
```
MatthewTurk: This is a behavior change, right?  Now you can not specify left and right edges to this object?  This is an improvement -- I am just asking a clarifying question.

yt/data_objects/data_containers.py
```diff
@@ -643,7 +653,7 @@ def save_as_dataset(self, filename=None, fields=None):
 
         extra_attrs = dict([(arg, getattr(self, arg, None))
                             for arg in self._con_args + self._tds_attrs])
-        extra_attrs["con_args"] = self._con_args
+        extra_attrs["con_args"] = repr(self._con_args)
```
MatthewTurk: I'm not sure I understand why we need to do this -- `_con_args` is a tuple of strings, so it should serialize correctly in h5py without this.  Now we're turning the tuple into a string that represents the tuple, yes?

yt/data_objects/octree_subset.py
```diff
@@ -186,8 +186,8 @@ def deposit(self, positions, fields = None, method = None,
         op.initialize()
         mylog.debug("Depositing %s (%s^3) particles into %s Octs",
             positions.shape[0], positions.shape[0]**0.3333333, nvals[-1])
-        pos = np.asarray(positions.convert_to_units("code_length"),
-                         dtype="float64")
+        positions.convert_to_units("code_length")
```
MatthewTurk: I think as long as `positions` is an internally-generated field, it will be safe to drop the `float64`.  But I did have to stop and think about that.

yt/data_objects/static_output.py
```diff
@@ -216,6 +217,70 @@ def __get__(self, instance, owner):
     def __set__(self, instance, value):
         self.data[instance] = value
 
+class ConstantContainer(object):
```
MatthewTurk: Can these go into a separate file?  And I think for my own understanding, just a comment about what each is supposed to do (doesn't have to be docstrings or anything extensive) would be helpful.

yt/data_objects/tests/test_dataset_access.py
```diff
@@ -24,9 +29,9 @@ def test_box_creation():
     region = ds.region(center, left_edge, right_edge)
 
     for b in boxes:
-        assert_equal(b.left_edge, region.left_edge)
-        assert_equal(b.right_edge, region.right_edge)
-        assert_equal(b.center, region.center)
+        assert_almost_equal(b.left_edge, region.left_edge)
```
MatthewTurk: Why did are these now `almost`?

yt/fields/magnetic_field.py
```diff
@@ -62,7 +64,7 @@ def _magnetic_field_strength(field, data):
 
     def _magnetic_energy(field, data):
         B = data[ftype,"magnetic_field_strength"]
-        return 0.5*B*B/mag_factors[B.units.dimensions]
+        return 0.5*B*B/mag_factors(B.units.dimensions)
```
MatthewTurk: I sometimes get confused by closures -- since `mag_factors` only has `ds` from the outer namespace, I think it might be safer to make it accept `ds` as an argument and put it at the outermost function level.

yt/data_objects/static_output.py
```diff
@@ -1068,7 +1147,7 @@ def set_units(self):
                               omega_lambda=self.omega_lambda,
                               omega_radiation=self.omega_radiation,
                               use_dark_factor = use_dark_factor,
-                              w_0 = w_0, w_a = w_a)
+                              w_0 = w_0, w_a = w_a, unit_registry=self.unit_registry)
```
brittonsmith: This will likely have unintended consequences. Most of the cosmology calculations return values in the comoving frame but over multiple redshifts. The only way to make this possible was to equate the comoving and proper reference frames in the cosmology calculator's unit registry. I think the dataset and its cosmology calculator need to have distinct unit registries.

yt/frontends/ytdata/data_structures.py
```diff
@@ -86,7 +86,7 @@ def _parse_parameter_file(self):
             for key in f.attrs.keys():
                 v = parse_h5_attr(f, key)
                 if key == "con_args":
-                    v = v.astype("str")
+                    v = eval(v)
```
MatthewTurk: This is the pair to the `repr()` call above, I think.  I'm not totally sure this is safe.  What's the use case for this particular set of changes?

brittonsmith: This is a backward incompatible change. Old datasets won't be loadable. Is there a reason for this?

yt/data_objects/static_output.py
ngoldbaum: OK - can you give me a test case where this would matter?

yt/frontends/ytdata/data_structures.py
ngoldbaum: I did this to fix a test. The ytdata frontend is using the json representation of the unit registry, which I had to change in unyt so I don't think there's a clear way to load old registries written out by yt.units and load them with unyt again.

Are we really supporting loading ytdata datasets accross yt versions? There's no tests for that, for example...

yt/data_objects/construction_data_containers.py
ngoldbaum: Not sure what you're talking about here. You can still specify that.

yt/data_objects/data_containers.py
ngoldbaum: h5py serializes a tuple of strings as a string array, this was leading to a case where I wasn't able to round-trip `con_args` anymore. Using `repr` and `eval` seemed like the most straightforward way to take care of that.

yt/data_objects/octree_subset.py
ngoldbaum: OK, I can add an `astype('float64')`

yt/data_objects/tests/test_dataset_access.py
ngoldbaum: I think there's an extra floating point operation happening somewhere that introduced some roundoff error.

yt/frontends/ytdata/data_structures.py
ngoldbaum: I explained above in the question about `repr`. Not sure what you mean by "safe" - are you talking about a security concern or something? This is all data that's internal to yt...

yt/utilities/cosmology.py
```diff
@@ -97,13 +99,15 @@ def __init__(self, hubble_constant = 0.71,
         self.omega_lambda = float(omega_lambda)
         self.omega_curvature = float(omega_curvature)
         if unit_registry is None:
-            unit_registry = UnitRegistry()
-            unit_registry.modify("h", hubble_constant)
+            unit_registry = UnitRegistry(unit_system=unit_system)
```
ngoldbaum: Also @brittonsmith note how there are still distinct unit registries, I'm just using the datasets's unit system.

yt/data_objects/octree_subset.py
MatthewTurk: No no, I think `astype` will make a copy, and I think we're safe without it.

yt/data_objects/static_output.py
brittonsmith: It looks like you've already changed it, but PR #2230 adds a test for this.

MatthewTurk:
@ngoldbaum -- looks like the only remaining conflict is in `appveyor.yml`, and all existing tests pass.  Does the current state work with older `YTData` outputs?  If so, what else remains before we can merge this in?

ngoldbaum:
There would need to either be support in yt or unyt to load old unit registries. I was hoping to finish that off before I left NCSA but did not get it done unfortunately.

MatthewTurk:
@ngoldbaum Do you have a suggestion for how we can move forward?  Is this something that could reasonably be implemented by someone else?

ngoldbaum:
I think so, yes. Help would be greatly appreciated. Support could either live in yt or unyt, whichever is easier.

brittonsmith:
We now have tests in `master` for loading old `ytdata` datasets, so we either have a block here or in merging `yt-4.0` into `master`. I'd be happy to look into this, but I don't really know where to look. @ngoldbaum, would it be possible to lay out what needs to be done to make sure old `ytdata` is still loadable?

ngoldbaum:
The json unit registry loader in unyt would either need to be patched to be able to load old unit registries or there would need to be a shim in yt that can parse the old json unit registries and emit a new unyt UnitRegistry object.

brittonsmith:
Can you say a bit about how the registries are different?

ngoldbaum:
The relevant change is this one: https://github.com/yt-project/unyt/commit/6283d20e69219735cc1c734f16895517dd72fb25. Also take a look at the explanation in the pull request: https://github.com/yt-project/unyt/pull/81

The change in the json output format is because we no longer use `srepr` to represent the symbol objects and instead just save the name of the symbol object. That means we go from saving e.g. `Symbol('mass', positive=True)` in the output to saving just `mass`.

With the old way of loading unit registries, sympy creates a distinct symbol object that satisfies `==` equality but not `is` equality. This is problematic because we use `is` equality a bunch of places inside `unyt` for speed. My fix was to make the dimensions in `unyt.dimensions` be singletons and use them as the namespace for restoring the unit registry json.

I think you'd need to do a similar trick with the old json objects, parsing e.g. `Symbol('energy', positive=True)` to just `'energy'`, and then either creating the unit registry manually or using the `unyt` loader.

I'm very sorry that I didn't do this myself and finish this before leaving. I'm pretty much completely burnt out on yt development and every time I tried to start on this I'd distract myself and feel guilty about it. I should have asked for help earlier and I'm sorry that I didn't.

ngoldbaum:
https://github.com/yt-project/unyt/pull/93 should hopefully unblock this

matthewturk:
Looks like appveyor dies with:

```
FAIL: yt.frontends.stream.tests.test_stream_particles.test_load_particles_with_data_source
  File "C:\Miniconda36-x64\envs\test\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\projects\yt\yt\frontends\stream\tests\test_stream_particles.py", line 338, in test_load_particles_with_data_source
    assert in_cgs(get_cu(ds1, dim)) == in_cgs(get_cu(ds2, dim))
AssertionError
```

I'm not entirely sure but maybe this is just a test that doesn't need to be run anymore?

ngoldbaum:
No the test probably needs to be fixed, it’s been a while since this PR was updated. I didn’t actually try running the test locally yet.

ngoldbaum:
I've opened https://github.com/yt-project/unyt/pull/96 to fix the failing test. Unfortunately we're going to need another unyt version bump to get the fix here in yt.

matthewturk:
AWESOME!

PR #2226, opened 2019-04-04
Adding conditional string decode.  Fixes #2223.

### LINKED ISSUE #2223: get_yt_version not working in Python 3
### Bug report

Running `yt.get_yt_version` shows the following behavior.

```
>>> import yt
>>> yt.get_yt_version()
  File "/Users/britton/Documents/work/yt/yt/yt/funcs.py", line 652, in get_yt_version
    return version[:12].strip().decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```

This is on the yt dev tip with Python 3.7.1. I can take care of this, but not at the moment. I'm making this issue so I don't forget.

MatthewTurk:
I've wrapped this decode in a conditional because I believe that
gitpython versions may have changed this behavior, and I don't want to
mess it up for any previous versions.

### DIFF
--- a/yt/funcs.py
+++ b/yt/funcs.py
@@ -649,7 +649,10 @@ def get_yt_version():
     if version is None:
         return version
     else:
-        return version[:12].strip().decode('utf-8')
+        v_str = version[:12].strip()
+        if hasattr(v_str, 'decode'):
+            v_str = v_str.decode('utf-8')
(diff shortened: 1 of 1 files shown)


ngoldbaum:
@brittonsmith that's because of #1702, you can safely ignore it.

Issue #2233, opened 2019-04-10
Chaining cut regions gives inconsistent results


stonnes:
### Bug report

**Bug summary**

There are two ways to "chain" cut regions:
1) `a = ds.cut_region(selection)`
    `b = a.cut_region(next selection)`

or 

2) `b = ds.cut_region(selection & next selection)`

They do not give the same results, and I think 1) is giving the incorrect results.  This will be, I hope, a more clear update of issue #1093.

**Code for reproduction**

```python
import sys
import numpy as np
import yt 

ds = yt.load("/mnt/ceph/users/stonnesen/runs/IsolatedGalaxy/galaxy0030/galaxy0030")

r0 = ds.all_data()
r1 = ds.disk([0.5,0.5,0.5],[0,0,1],(4,"kpc"),(10,"kpc"))
r1c = r1.cut_region(["(obj['cylindrical_r'].in_units('kpc') > 2.0)"])
r2c = r1.cut_region(["(obj['density'] > 1e-26)"])
r3c = r1c.cut_region(["(obj['density'] > 1e-26)"])
r4c = r1.cut_region(["(obj['cylindrical_r'].in_units('kpc') > 2.0) & (obj['density'] > 1e-26)"])

print ('Only density cut:')
print ('total cell volume:',r2c.quantities.total_quantity([("index","cell_volume")]))
print ('total cell mass:',r2c.quantities.total_quantity([("gas","cell_mass")]))
print ('extrema z:',r2c.quantities.extrema([("index","z")]))
print ('extrema y:',r2c.quantities.extrema([("index","y")]))
print ('extrema x:',r2c.quantities.extrema([("index","x")]))
print ('extrema cyl_r:',r2c.quantities.extrema([("index","cylindrical_r")]))
 
print ('Only spatial cut:')
print ('total cell volume:',r1c.quantities.total_quantity([("index","cell_volume")]))
print ('total cell mass:',r1c.quantities.total_quantity([("gas","cell_mass")]))
print ('extrema z:',r1c.quantities.extrema([("index","z")]))
print ('extrema y:',r1c.quantities.extrema([("index","y")]))
print ('extrema x:',r1c.quantities.extrema([("index","x")]))
print ('extrema cyl_r:',r1c.quantities.extrema([("index","cylindrical_r")]))
    
print ('Added density cut')
print ('total cell volume:',r3c.quantities.total_quantity([("index","cell_volume")]))
[...]

PR #2234, opened 2019-04-10
[bugfix] fix io chunking for chained cut regions

brittonsmith:
## PR Summary

This closes Issue #2233.

When chained cut regions do io chunking, only the conditionals of the last cut region are applied. Conditionals from parent cut regions are ignored. This only shows up when calculating derived quantities.

The solution implemented here changes how cut regions derived from other cut regions are initialized. Instead of creating a chain of objects that rely on each's parent, we add the conditionals from the parent cut region and create the new one from the parent's base object. Anecdotally, this also appears to speed up field access as we're only applying conditionals once.

## PR Checklist

### DIFF
--- a/yt/data_objects/selection_data_containers.py
+++ b/yt/data_objects/selection_data_containers.py
@@ -871,16 +871,18 @@ class YTCutRegion(YTSelectionContainer3D):
                 raise RuntimeError(
                     "Cannot use both base_object and data_source")
             data_source=base_object
+
+        self.conditionals = ensure_list(conditionals)
         if isinstance(data_source, YTCutRegion):
-            self.__init__(
-                data_source.base_object,
-                data_source.conditionals + conditionals,
-                ds=ds, field_parameters=field_parameters,
-                base_object=base_object)
-            return
+            # If the source is also a cut region, add its conditionals
+            # and set the source to be its source.
+            # Preserve order of conditionals.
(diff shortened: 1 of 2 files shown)


yt/data_objects/selection_data_containers.py
```diff
@@ -871,6 +871,13 @@ def __init__(self, data_source, conditionals, ds=None,
                 raise RuntimeError(
                     "Cannot use both base_object and data_source")
             data_source=base_object
+        if isinstance(data_source, YTCutRegion):
+            self.__init__(
+                data_source.base_object,
+                data_source.conditionals + conditionals,
```
MatthewTurk: I think `ensure_list(conditionals)` needs to happen before this

ngoldbaum: Calling `__init__` inside of `__init__` is a little weird to me. Is it possible to set `data_source` and `conditionals` in this if block and then set up the instance using the existing logic below?

Also I realize this is how it is already, but a straightforward cleanup would be to deprecate (with a visible warning) passing `base_object` in and to call the attribute `data_source` instead of `base_object`.

brittonsmith: Totally right!

brittonsmith: Yeah, this can probably be simplified like that. I'll give it a go.

PR #2236, opened 2019-04-12
[bugfix] Arepo smoothing length is supposed to be float64

jzuhone:
This got missed in the final cleanup steps in the recent Arepo frontend merge.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,9 +29,9 @@ install:
     - "python --version"
 
     # Install specified version of numpy and dependencies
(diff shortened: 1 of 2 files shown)


ngoldbaum:
The appveyor failure is fixed on master and can be ignored. If you want to separately port the fix in this PR that would be nice since I can’t do that easily at the moment.

jzuhone:
@ngoldbaum sorry, what do you mean exactly?

jzuhone:
oh wait, never mind

Issue #2242, opened 2019-04-16
A bug in arepo test


vikramkhaire:
### Bug report
I was trying some test examples as suggested in the yt/yt/frontends/arepo/tests. I downloaded the ArepoBullet data. And tried following

**Code for reproduction**
```
import yt
#file= path and file where the snapshot_150.hdf5 stored
ds=yt.load(file)
ds.print_stats()
```

**I get the following error message**

```
#
yt : [INFO     ] 2019-04-16 05:07:48,529 Allocating for 2.653e+07 particles
Initializing coarse index :   0%|                                                                                                                     | 0/51 [00:00<?, ?it/s]---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
~/yt/yt/geometry/particle_geometry_handler.py in _initialize_index(self)
    142         try:
--> 143             rflag = self.regions.load_bitmasks(fname)
    144             rflag = self.regions.check_bitmasks()

~/yt/yt/geometry/particle_oct_container.pyx in yt.geometry.particle_oct_container.ParticleBitmap.load_bitmasks()

OSError: 

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-14-d2b97580ae55> in <module>()
----> 1 ds.print_stats()

~/yt/yt/utilities/parallel_tools/parallel_analysis_interface.py in root_only(*args, **kwargs)
    317     def root_only(*args, **kwargs):
    318         if not parallel_capable:
--> 319             return func(*args, **kwargs)
    320         comm = _get_comm(args)
    321         rv = None

~/yt/yt/data_objects/static_output.py in print_stats(self)
    552     @parallel_root_only
    553     def print_stats(self):
--> 554         self.index.print_stats()
    555 
    556     @property

~/yt/yt/data_objects/static_output.py in index(self)
    514                 raise RuntimeError("You should not instantiate Dataset.")
    515             self._instantiated_index = self._index_class(
--> 516                 self, dataset_type=self.dataset_type)
    517             # Now we do things that we need an instantiated index for
    518             # ...first off, we create our field_info now.

~/yt/yt/geometry/particle_geometry_handler.py in __init__(self, ds, dataset_type)
     41         self.float_type = np.float64
     42         super(ParticleIndex, self).__init__(ds, dataset_type)
---> 43         self._initialize_index()
     44 
     45     def _setup_geometry(self):

~/yt/yt/frontends/sph/data_structures.py in _initialize_index(self)
[...]

PR #2244, opened 2019-04-17
[bugfix] Make sure Arepo "smoothing length" has correct dtype

jzuhone:
In the Gadget frontend we apply a fix if there is a mismatch in precision between the positions and the smoothing length fields stored on-disk. Apparently this fix is also necessary for the Arepo frontend if we construct "smoothing lengths" from the density and mass fields.

Fixes issue #2242.


ngoldbaum:
LGTM!

PR #2245, opened 2019-04-17
Some updates and additions to astrophysics-related fields

jzuhone:
## PR Summary

This PR updates a number of the astronomy/astrophysics-specific fields in yt. In detail:

* Wherever possible, change `"cell_mass"` to `"mass"` so that fields are not only defined for grid-based datasets. Similarly, if possible, do not use `"cell_volume"` for fields which are meant to be general. 
* Add a computation for the number density fields for free electrons, protons, and helium nuclei for when there are no species fields, assuming primordial abundances and full ionization. 
* Redefine `"number_density"` and `"mean_molecular_weight"` fields so that they can be calculated even if no species are defined, by assuming a constant mean molecular weight and primordial abundances. 
* Add an `"optical_depth"` field which depends on the `"El_number_density"` field. Redefine the two SZ fields in terms of this field (otherwise they just assume full ionization, which may not be true).
* Add a `"velocity_los"` field for projections, which computes the velocity component along a line of sight, on-axis or off-axis. Use it in the `"sz_kinetic"` field. 
* Add a `"magnetic_field_los"` field for projections, which computes the magnetic field strength component along a line of sight, on-axis or off-axis. This is used for the Faraday rotation measure field, which is also added here. 
* Add an alias for the `"particle_gpot"` field in the FLASH frontend.
* Move the `"entropy"` field to `astro_fields`, since its field definition is very astro-specific (and in fact it's very high-energy-astro-specific). Make it depend on the `"El_number_density"` field (otherwise it just assumes full ionization, which may not be true). 
* Remove `chandra_emissivity` field. It's not well-documented and it is not clear at all what it represents (i.e. what is the energy band? what is the instrument assumed?), so it's misleading. 
* Remove extra whitespace and parentheses in various places
* Set up a function which computes a default mean molecular weight for frontends that may use one for defining things like temperature and number density but don't have one. This default `mu` is now used in the `athena`, `athena_pp`, and `gadget` frontends unless the user passes in a different value when loading the dataset. NOTE: This default `mu` assumes _full_ ionization.
* Made sure that the `flash`, `enzo`, and `gamer` frontends get to define fields like `"number_density"` the way they want. 
* Fixed a bug which prevented magnetic fields from being defined properly in SPH frontends.
* The `"cell_mass"` and `"courant_time_step"` fields should have `sampling_type="cell"`.
* Add species fields to the Arepo frontend. 
* Fixed some bugs created in off-axis projections 

## PR Checklist

### DIFF
--- a/doc/source/analyzing/fields.rst
+++ b/doc/source/analyzing/fields.rst
@@ -426,6 +426,8 @@ Within a field function, these can then be retrieved and used in the same way.
 
 For a practical application of this, see :ref:`cookbook-radial-velocity`.
 
+.. _gradient_fields:
+
 Gradient Fields
 ---------------
 
@@ -448,6 +450,54 @@ of how to create and use these fields, see :ref:`cookbook-complicated-derived-fi
 .. note::
 
     ``add_gradient_fields`` currently only supports Cartesian geometries!
+
+.. _los_fields:
+
+Line of Sight Fields
+--------------------
+
+In astrophysics applications, one often wants to know the component of a vector 
+field along a given line of sight. If you are doing a projection of a vector 
+field along an axis, or just want to obtain the values of a vector field 
+component along an axis, you can use a line-of-sight field. For projections, 
+this will be handled automatically:
+
+.. code-block:: python
+
+    prj = yt.ProjectionPlot(ds, "z", ("gas","velocity_los"), 
+                            weight_field=("gas","density"))
+
+Which, because the axis is ``"z"``, will give you the same result if you had
+projected the `"velocity_z"`` field. This also works for off-axis projections:
+
+.. code-block:: python
+
+    prj = yt.OffAxisProjectionPlot(ds, [0.1, -0.2, 0.3], ("gas","velocity_los"), 
+                                   weight_field=("gas","density"))
+
+
+This shows that the projection axis can be along a principle axis of the domain 
+or an arbitrary off-axis 3-vector (which will be automatically normalized). If 
+you want to examine a line-of-sight vector within a 3-D data object, set the 
+``"axis"`` field parameter:
+
+.. code-block:: python
+
+    dd = ds.all_data()
+    # Set to one of [0, 1, 2] for ["x", "y", "z"] axes
+    dd.set_field_parameter("axis", 1)
+    print(dd["gas","magnetic_field_los"])
+    # Set to a three-vector for an off-axis component
+    dd.set_field_parameter("axis", [0.3, 0.4, -0.7])
+    print(dd["gas","velocity_los"])
+
+At this time, this functionality is enabled for the velocity and magnetic vector
+fields, ``("gas","velocity_los")`` and ``("gas","magnetic_field_los")``. The 
+following fields built into yt make use of these line-of-sight fields:
+
+* ``("gas","sz_kinetic")`` uses ``("gas","velocity_los")``
+* ``("gas","rotation_measure")`` uses ``("gas","magnetic_field_los")``
+
 
 General Particle Fields
 -----------------------
--- a/doc/source/analyzing/fields.rst
+++ b/doc/source/analyzing/fields.rst
@@ -548,6 +548,12 @@ you want to examine a line-of-sight vector within a 3-D data object, set the
     dd.set_field_parameter("axis", [0.3, 0.4, -0.7])
     print(dd["gas","velocity_los"])
 
+.. warning::
+
+    If you need to change the axis of the line of sight on the *same* data container
+    (sphere, box, cylinder, or whatever), you will need to delete the field using
+    ``del dd["velocity_los"]`` and re-generate it. 
+
 At this time, this functionality is enabled for the velocity and magnetic vector
 fields, ``("gas","velocity_los")`` and ``("gas","magnetic_field_los")``. The 
 following fields built into yt make use of these line-of-sight fields:
--- a/doc/source/analyzing/fields.rst
+++ b/doc/source/analyzing/fields.rst
@@ -424,7 +424,9 @@ the following fields are defined:
 The ``mean_molecular_weight`` field will be constructed from the abundances of the elements
 in the dataset. If no element or molecule fields are defined, the above fields for the ionized
 primordial H/He plasma are defined, and the ``mean_molecular_weight`` field is correspondingly set
-to :math:`\mu \approx 0.6`.
+to :math:`\mu \approx 0.6`. Some frontends do not directly store the gas temperature in their
+datasets, in which case it must be computed from the pressure and/or thermal energy as well
+as the mean molecular weight, so check this carefully!
 
 Particle Fields
 ---------------
--- a/doc/source/analyzing/fields.rst
+++ b/doc/source/analyzing/fields.rst
@@ -385,9 +385,12 @@ within the dataset, as well as their abundances and ionization states. Examples 
 
 The naming scheme for the fields starts with prefixes in the form ``MM[_[mp][NN]]``. ``MM``
 is the molecule, defined as a concatenation of atomic symbols and numbers, with no spaces or
-underscores. The second sequence is only required if the ionization state is not neutral,
-and is of the form ``p`` and ``m`` to indicate "plus" or "minus" respectively, followed by
-the number. For the examples above, the prefixes would be:
+underscores. The second sequence is only required if ionization states are present in the
+dataset, and is of the form ``p`` and ``m`` to indicate "plus" or "minus" respectively, 
+followed by the number. If a given species has no ionization states given, the prefix is
+simply ``MM``. 
+
+For the examples above, the prefixes would be:
 
 * ``CO``
 * ``Co``
@@ -408,8 +411,7 @@ defined:
 * ``MM[_[mp][NN]]_mass``
 
 To refer to the number density of the entirety of a single atom or molecule (regardless
-of its ionization state), please use the ``MM_nuclei_density`` fields, as opposed to
-``MM_number_density`` fields.
+of its ionization state), please use the ``MM_nuclei_density`` fields.
 
 Finally, if the abundances of hydrogen and helium are not defined, it is assumed that
 assumed that these elements are fully ionized with primordial abundances In this case,
--- a/doc/source/analyzing/fields.rst
+++ b/doc/source/analyzing/fields.rst
@@ -371,6 +371,61 @@ different magnetic field units in the different :ref:`unit systems <unit_systems
 determine how to set up special magnetic field handling when designing a new frontend, check out
 :ref:`bfields-frontend`.
 
+Species Fields
+--------------
+
+For many types of data, yt is able to detect different chemical elements and molecules
+within the dataset, as well as their abundances and ionization states. Examples include:
+
+* CO (Carbon monoxide)
+* Co (Cobalt)
+* OVI (Oxygen ionized five times)
+* H:math:`^{2+}` (Molecular Hydrogen ionized once)
+* H:math:`^{-}` (Hydrogen atom with an additional electron)
+
+The naming scheme for the fields starts with prefixes in the form ``MM[_[mp][NN]]``. ``MM``
+is the molecule, defined as a concatenation of atomic symbols and numbers, with no spaces or
+underscores. The second sequence is only required if the ionization state is not neutral,
+and is of the form ``p`` and ``m`` to indicate "plus" or "minus" respectively, followed by
+the number. For the examples above, the prefixes would be:
+
+* ``CO``
+* ``Co``
+* ``O_p5``
+* ``H2_p1``
+* ``H_m1``
+
+The name ``El`` is used for electron fields, as it is unambiguous and will not be
+utilized elsewhere. Neutral ionic species (e.g. H I, O I) are represented as ``MM_p0``.
+Additionally, the isotope of :math:`^2`H will be included as ``D``.
+
+Finally, in those frontends which are single-fluid, these fields for each species are
+defined:
+
+* ``MM[_[mp][NN]]_fraction``
+* ``MM[_[mp][NN]]_number_density``
+* ``MM[_[mp][NN]]_density``
+* ``MM[_[mp][NN]]_mass``
+
+To refer to the number density of the entirety of a single atom or molecule (regardless
+of its ionization state), please use the ``MM_nuclei_density`` fields, as opposed to
+``MM_number_density`` fields.
+
+Finally, if the abundances of hydrogen and helium are not defined, it is assumed that
+assumed that these elements are fully ionized with primordial abundances In this case,
+the following fields are defined:
+
+* ``H_p1_number_density``
+* ``H_nuclei_density``
+* ``He_p2_number_density``
+* ``He_nuclei_density``
+* ``El_number_density``
+
+The ``mean_molecular_weight`` field will be constructed from the abundances of the elements
+in the dataset. If no element or molecule fields are defined, the above fields for the ionized
+primordial H/He plasma are defined, and the ``mean_molecular_weight`` field is correspondingly set
+to :math:`\mu \approx 0.6`.
+
 Particle Fields
 ---------------
 
--- a/doc/source/examining/loading_data.rst
+++ b/doc/source/examining/loading_data.rst
@@ -1066,9 +1066,12 @@ present, Then the mean molecular weight is the inverse of ``"sumy"``, and the fi
 defined using the following mathematical definitions:
 
 * ``"El_number_density"`` :math:`n_e = N_AY_e\rho`
-* ``"ion_number_density"`` :math:`n_i = N_A\Sigma{Y}\rho`
+* ``"ion_number_density"`` :math:`n_i = N_A\rho/\bar{A}`
 * ``"number_density"`` :math:`n = n_e + n_i`
 
+where :math:`n_e` and :math:`n_i` are the electron and ion number densities, 
+:math:`\rho` is the mass density, :math:`Y_e` is the electron number per baryon,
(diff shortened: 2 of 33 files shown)


jzuhone:
After rebasing this PR I'd like to come back to it. The major unresolved issue is what I said above (see the full text above for context):

> There is one subtle inconsistency in this setup at the moment. The field definitions I wrote for the "number_density' and "mean_molecular_weight" fields assume primordial abundances and zero ionization. This is what we do in a number of the SPH frontends when we need to compute the temperature but there are no species fields to compute the molecular weight. But defining an free electron number density field really only makes sense if at least the gas is partially ionized, so in that case I've assumed full ionization. But there's obviously an inconsistency there. It would be easy to just define the number density fields here and the temperature fields in the SPH frontends to assume full ionization instead if there are no species defined, but this is potentially a significant backwards-incompatible change for the latter.

The possible resolutions I can see are:

1. Have the default for both species fields and the mean molecular weight be zero ionization
2. Have the default for both species fields and the mean molecular weight be full ionization
3. Have this set up in a frontend-dependent way, perhaps by using parameters

jzuhone:
@brittonsmith I like that idea regarding vector operations. 

Anyone who works with SPH data have any concerns about changing the zero ionization assumption? @qobilidop or @AshKelly? Note that this only affects data which doesn't have any abundances defined.

Relevant code is here:

https://github.com/yt-project/yt/blob/33dcf445e951d847c4553d32f58d55c2fa557a6b/yt/frontends/gadget/fields.py#L70-L92

jzuhone:
Would also love any comment from @chummels

brittonsmith:
@jzuhone, I think Cameron is busily preparing to leave on a long trip at the end of the week. It's likely he's not going to have time to comment here.

AshKelly:
yeah, I think zero ionization is appropriate if there are no other fields, e.g electron fraction. Hopefully, if people don't explicitly store electron fraction or temperature then they aren't too worried about it

jzuhone:
@ashkelly the proposal here currently is to assume _full_ ionization unless specified. 

Unfortunately it’s not true that if people don’t store those things that they don’t care (although in theory it should be). For most applications in X-ray astrophysics, full ionization is assumed and so you need a mean molecular weight which reflects that to get electron number density and temperature, both of which are needed for emission and other calculations.

So we need to decide if we want to assume full ionization by default (a decidedly higher-temperature astrophysics bias) or zero ionization (a lower temperature general physics bias). Or if we want to have some kind of switch that does either.

matthewturk:
@jzuhone This is a really interesting question.  I'm not sure I have the *right* answer, but I would propose that since we're talking about a *default*, we make room for different codes and frontends to set it, and we also identify how to set it -- probably in something less fragile than `field_parameters`.

I think that if we can't *deduce* it, then it's safe for *this field* to assume full ionization.  Does that make sense?

jzuhone:
Ok, I made a flurry of changes which I think settles the question regarding how to set mean molecular weights. Please have a look at the latest changes starting with https://github.com/yt-project/yt/pull/2245/commits/3e412fd15d00f150dd33b9cb1d6183be66404beb and leave some feedback.

yt/fields/vector_operations.py
```diff
@@ -89,6 +90,37 @@ def _relative_vector(field, data):
...
+                     ftype='gas', slice_info=None,
+                     validators=None):
+    axis_order = registry.ds.coordinates.axis_order
+
+    field_comps = [(ftype, "%s_%s" % (basename, ax)) for ax in axis_order]
+    field_comps_rel = [(ftype, "relative_%s_%s" % (basename, ax)) 
+                       for ax in axis_order]
+
+    def _los_field(field, data):
+        axis = data.get_field_parameter("axis")
```
brittonsmith: Quick question, does `axis` need to be checked to see if it's a unit vector?

yt/frontends/athena_pp/fields.py
```diff
@@ -81,11 +81,7 @@ def _thermal_energy(field, data):
                            units=unit_system["specific_energy"])
         # Add temperature field
         def _temperature(field, data):
-            if data.has_field_parameter("mu"):
-                mu = data.get_field_parameter("mu")
-            else:
-                mu = 0.6
-            return (data["gas","pressure"]/data["gas","density"])*mu*mh/kboltz
+            return (data["gas","pressure"]/data["gas","density"])*data.ds.mu*mh/kboltz
```
brittonsmith: Looks like above you're setting `self.parameters["mu"]`. Do you need to use that here?

yt/frontends/enzo/fields.py
```diff
@@ -291,8 +291,18 @@ def _tot_minus_kin(field, data):
...
                 units=unit_system["specific_energy"])
         if multi_species == 0 and 'Mu' in params:
+            def _mean_molecular_weight(field, data):
+                return params["Mu"]*data['index', 'ones']
+
+            self.add_field(
+                ("gas", "mean_molecular_weight"),
+                sampling_type="cell",
+                function=_mean_molecular_weight,
+                units=unit_system["number_density"])
```
brittonsmith: Are these the correct units?

yt/frontends/flash/fields.py
```diff
@@ -171,21 +179,43 @@ def _nion(field, data):
...
+            def _number_density(field, data):
+                Na_code = data.ds.quan(Na, '1/code_mass')
+                return data["flash", "dens"]*Na_code/data["flash", "abar"]
+
+            self.add_field(("gas", "number_density"),
+                           sampling_type="cell",
+                           function=_number_density,
+                           units=unit_system["number_density"])
+
+        if "ye" in self.field_list:
```
brittonsmith: I'm not that familiar with the FLASH format. Can you get both this condition and the one above (`if ("flash", "abar") in self`)? If so, you could have one number_density field overwriting another. Should this be an `elif`?

yt/utilities/chemical_formulas.py
```diff
@@ -43,3 +45,14 @@ def __init__(self, formula_string):
...
         return self.formula_string
+
+
+def compute_mu():
+    # Assume full ionization and cosmic abundances
+    mu = ChemicalFormula("H").weight / \
+         (2.0 * _primordial_mass_fraction["H"])
+    mu += ChemicalFormula("He").weight / \
+          (3.0 * _primordial_mass_fraction["He"])
+
```
brittonsmith: This never returns mu.

yt/utilities/chemical_formulas.py
```diff
@@ -43,3 +45,14 @@ def __init__(self, formula_string):
 
     def __repr__(self):
         return self.formula_string
+
+
+def compute_mu():
+    # Assume full ionization and cosmic abundances
```
brittonsmith: Something's not quite right here. This function results in a value for `default_mu` of 6.22.

yt/fields/species_fields.py
```diff
@@ -157,22 +155,34 @@ def add_nuclei_density_fields(registry, ftype):
...
 
 def _default_nuclei_density(field, data):
     ftype = field.name[0]
     element = field.name[1][:field.name[1].find("_")]
-    return data[ftype, "density"] * _primordial_mass_fraction[element] / \
-      ChemicalFormula(element).weight / data.ds.units.physical_constants.amu_cgs
-        
+    amu_cgs = data.ds.units.physical_constants.amu_cgs
+    if element == "El":
+        # This assumes full ionization!
```
brittonsmith: Could you use `default_mu` here?

yt/frontends/enzo/fields.py
jzuhone: No, you are right, I'll fix that. Thanks for the catch.

yt/fields/species_fields.py
jzuhone: You can't because this is for the electron number density only and not the average number density (over all species).

yt/frontends/athena_pp/fields.py
jzuhone: Ah, I didn't do this right. Thanks for the find again.

yt/fields/vector_operations.py
jzuhone: It could be either a unit vector or a single integer for an axis-aligned l.o.s. Do you have a suggestion for a check?

yt/frontends/flash/fields.py
```diff
@@ -148,44 +150,67 @@ def _eint(field, data):
...
+                           sampling_type="cell",
+                           function=_number_density,
+                           units=unit_system["number_density"])
+
+        else:
+            def _number_density(field, data):
+                Na_code = data.ds.quan(Na, '1/code_mass')
+                return data["flash", "dens"]*Na_code/data["gas", "mean_molecular_weight"]
+
+            self.add_field(("gas", "number_density"),
```
brittonsmith: Minor, but you could indent this `self.add_field` call one to the left and remove the identical call inside the if statement above.

jzuhone:
Hi @brittonsmith, 

Thanks. I think this needs some more testing and docs before I can have this  merged in, so I'm leaving it as WIP for now.

brittonsmith:
Ok, that's fair. I can take another look when you're ready to get this merged.

jzuhone:
@brittonsmith I note that if we have species in the dataset, then it appears we implicitly assume that they are all neutral unless told otherwise:

https://github.com/yt-project/yt/blob/cab1f48f982c28fbd1f8c1075ce3f1a0db5e686c/yt/fields/species_fields.py#L231-L236

which appears inconsistent with the full ionization assumption employed in this PR, since if (say) a dataset defines C, N, and O, and they are assumed to be fully neutral, but then the H and He will be ionized by default.

Any thoughts on the best way to go here?

jzuhone:
Actually, the YTEP (https://ytep.readthedocs.io/en/master/YTEPs/YTEP-0003.html) says this:

> Neutral ionic species (e.g. H I, O I) are represented as MM_p0. For backwards compatibility, neutral species will be mirrored to MM_ fields, but this practice is deprecated as it is somewhat ambiguous if we are referring to just the neutral species or all atoms of that type. As an example, the neutral hydrogen density field will now be called H_p0_density, but it will be mirrored to the H_density field for backwards compatibility.

So I think we should just remove this aliasing from `MM_` to `MM_p0_`.

jzuhone:
Ok, after a lot of tinkering this is now ready. Please look in particular at the new docs I've added, and _especially_ at `species_fields.py`.

jzuhone:
@brittonsmith @matthewturk would you like to take another look at this? The Appveyor test failure is unrelated.

jzuhone:
Since this touches a lot of stuff, we should probably enforce three approvals here. 
@qobilidop, @AshKelly, @munkm, anyone?

jzuhone:
Anyone want to take a peek at this?

doc/source/analyzing/fields.rst
```diff
@@ -449,6 +508,60 @@ of how to create and use these fields, see :ref:`cookbook-complicated-derived-fi
...
+
+.. code-block:: python
+
+    dd = ds.all_data()
+    # Set to one of [0, 1, 2] for ["x", "y", "z"] axes
+    dd.set_field_parameter("axis", 1)
+    print(dd["gas","magnetic_field_los"])
+    # Set to a three-vector for an off-axis component
+    dd.set_field_parameter("axis", [0.3, 0.4, -0.7])
+    print(dd["gas","velocity_los"])
```
matthewturk: Somehow, I did not realize this was part of this PR.  This is amazing.

yt/data_objects/construction_data_containers.py
```diff
@@ -180,6 +180,7 @@ def __init__(self, field, axis, weight_field=None, center=None, ds=None,
         self._set_center(center)
         self._projected_units = {}
         if data_source is None: data_source = self.ds.all_data()
+        data_source.set_field_parameter("axis", self.axis)
```
matthewturk: I *think* we want this to be set in the `self` object, and assume it will get passed down during the contextmanager during the accessing of `chunks` on the `data_source`.

yt/fields/astro_fields.py
```diff
@@ -36,11 +35,11 @@ def _dynamical_time(field, data):
                        units=unit_system["time"])
 
     def _jeans_mass(field, data):
-        MJ_constant = (((5.0 * pc.kboltz) / (pc.G * pc.mh)) ** (1.5)) * \
-          (3.0 / (4.0 * np.pi)) ** (0.5)
+        MJ_constant = (((5.0 * pc.kboltz) / (pc.G * pc.mh)) ** 1.5) * \
```
matthewturk: Should we define this somewhere that it could be gotten at the top level?

jzuhone: The usage is a bit obscure, not sure if it's worth it

yt/data_objects/construction_data_containers.py
jzuhone: Sorry, could you elaborate?

jzuhone: I recall this not working if I didn't do this.

jzuhone: Ok, for some reason projections of particle datasets don't work if I don't do this. Maybe the same logic for grids needs to be implemented in the particle chunking vis-a-vis field parameters?

jzuhone: Never mind, fixed this

matthewturk:
I see no reasons to hold off on merging this.

jzuhone:
@matthewturk fine by me! Thanks!

doc/source/examining/loading_data.rst
```diff
@@ -1336,7 +1372,9 @@ value can be changed when loading an Arepo dataset by setting the
    import yt
    ds = yt.load("snapshot_100.hdf5", smoothing_factor=1.5)
 
-Currently, only Arepo HDF5 snapshots are supported.
+Currently, only Arepo HDF5 snapshots are supported. If the "GFM" metal fields are
+present in your dataset, they will be loaded in and aliased to the appropriate
```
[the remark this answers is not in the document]
jzuhone: Yup, sounds good.

doc/source/examining/loading_data.rst
```diff
@@ -1032,7 +1047,28 @@ grid structure and are at the same simulation time, the particle data may be loa
...
+this is used as the mean molecular weight. In either case, the number density field
+is calculated using this weight. 
+
+If you are running a FLASH simulation where the fields ``"sumy"`` and ``"ye"`` are 
+present, Then the mean molecular weight is the inverse of ``"sumy"``, and the fields 
+``"El_number_density"``, ``"ion_number_density"``, and ``"number_density"`` are 
+defined using the following mathematical definitions:
+
+* ``"El_number_density"`` :math:`n_e = N_AY_e\rho`
+* ``"ion_number_density"`` :math:`n_i = N_A\Sigma{Y}\rho`
```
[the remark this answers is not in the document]
jzuhone: Sum over the Y's for different ions. I can add some clarification.

doc/source/analyzing/fields.rst
```diff
@@ -449,6 +508,60 @@ of how to create and use these fields, see :ref:`cookbook-complicated-derived-fi
...
+
+.. warning::
+
+    If you need to change the axis of the line of sight on the *same* data container
+    (sphere, box, cylinder, or whatever), you will need to delete the field using
+    ``del dd["velocity_los"]`` and re-generate it. 
+
+At this time, this functionality is enabled for the velocity and magnetic vector
+fields, ``("gas","velocity_los")`` and ``("gas","magnetic_field_los")``. The 
+following fields built into yt make use of these line-of-sight fields:
```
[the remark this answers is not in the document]
jzuhone: Not yet, but in fact, none of the vector field operations (like magnitude, etc) are available for any other fields than these two. We should probably expose this in a separate PR somehow.

yt/fields/xray_emission_fields.py
```diff
@@ -203,6 +203,8 @@ def add_xray_emissivity_field(ds, e_min, e_max, redshift=0.0,
             raise RuntimeError("Your dataset does not have a {} field! ".format(metallicity) +
                                "Perhaps you should specify a constant metallicity instead?")
 
+    H_field = "H_nuclei_density" if table_type == "cloudy" else "H_p1_number_density"
```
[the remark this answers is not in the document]
jzuhone: I can fix this.

doc/source/analyzing/fields.rst
```diff
@@ -371,6 +371,63 @@ different magnetic field units in the different :ref:`unit systems <unit_systems
...
+Finally, in those frontends which are single-fluid, these fields for each species are
+defined:
+
+* ``MM[_[mp][NN]]_fraction``
+* ``MM[_[mp][NN]]_number_density``
+* ``MM[_[mp][NN]]_density``
+* ``MM[_[mp][NN]]_mass``
+
+To refer to the number density of the entirety of a single atom or molecule (regardless
+of its ionization state), please use the ``MM_nuclei_density`` fields, as opposed to
```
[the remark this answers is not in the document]
jzuhone: I should clarify that as a result of this PR that `MM_number_density` refers to elements when no ionization states are present. This should be noted in the docs. If ionization states are present, `MM_number_density` does not exist.

jzuhone:
@munkm I think I've addressed everything, but let me know if there is anything else. Thanks for the comments!

doc/source/analyzing/fields.rst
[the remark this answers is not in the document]
munkm: I hadn't realized that! Good clarification.

doc/source/examining/loading_data.rst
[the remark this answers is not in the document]
munkm: I really like the clarifications!

PR #2248, opened 2019-04-18
[bugfix] Fix aliases and units for magnetic fields in SPH frontends

jzuhone:
## PR Summary

This PR fixes some issues with the Arepo and Gizmo frontends in terms of magnetic fields.

* The aliases to "local" magnetic fields (e.g. `("gas","magnetic_field_x")`) were not being set. This happened because the magnetic fields were being handled after the other aliases had been set up.
* The correct magnetic units were not being set in these frontends. I know that what I have done below is correct for Arepo, someone more familiar with Gizmo (perhaps @chummels?) should check what I have done here for that.
* I added a new test for Arepo with a new dataset, and we also now test the Gizmo MHD dataset for the magnetic field aliases. 
* I fixed an unrelated bug in Gadget's I/O handler (which both of these frontends use) which occurred when the selector was `all_data` and one tried to construct the mass field for a particle type without a variable particle mass.

### DIFF
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -24,8 +24,6 @@ from yt.frontends.gadget.data_structures import \
 from .fields import \
     GizmoFieldInfo
 
-import numpy as np
-
 
 class GizmoDataset(GadgetHDF5Dataset):
     _field_info_class = GizmoFieldInfo
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -24,6 +24,9 @@ from yt.frontends.gadget.data_structures import \
 from .fields import \
     GizmoFieldInfo
 
+import numpy as np
+
+
 class GizmoDataset(GadgetHDF5Dataset):
     _field_info_class = GizmoFieldInfo
 
@@ -58,3 +61,12 @@ class GizmoDataset(GadgetHDF5Dataset):
             valid = False
             pass
         return valid
+
+    def _set_code_unit_attributes(self):
+        super(GizmoDataset, self)._set_code_unit_attributes()
+        munit = np.sqrt(self.mass_unit /
+                        (self.time_unit**2 * self.length_unit)).to("gauss")
+        if self.cosmological_simulation:
+            self.magnetic_unit = self.quan(munit.value, "%s/a**2" % munit.units)
+        else:
+            self.magnetic_unit = munit
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -64,9 +64,4 @@ class GizmoDataset(GadgetHDF5Dataset):
 
     def _set_code_unit_attributes(self):
         super(GizmoDataset, self)._set_code_unit_attributes()
-        munit = np.sqrt(self.mass_unit /
-                        (self.time_unit**2 * self.length_unit)).to("gauss")
-        if self.cosmological_simulation:
-            self.magnetic_unit = self.quan(munit.value, "%s/a**2" % munit.units)
-        else:
-            self.magnetic_unit = munit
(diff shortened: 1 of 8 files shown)


ngoldbaum:
@yt-fido test this please

ngoldbaum:
The new answer tests you've added won't actually be run until either we get jenkins working on the yt-4.0 branch or merge yt-4.0 with master. You will also need to edit `tests.yaml` to include the arepo frontend before any of the answer tests will run.

jzuhone:
@ngoldbaum the Arepo frontend is already in `tests.yaml`, but since we're not running the answer tests yet here I think I don't have to increment the number?

ngoldbaum:
Ah, sorry, I was checking on `master`.

yt/frontends/gizmo/data_structures.py
```diff
@@ -58,3 +61,12 @@ def _is_valid(self, *args, **kwargs):
             valid = False
             pass
         return valid
+
+    def _set_code_unit_attributes(self):
+        super(GizmoDataset, self)._set_code_unit_attributes()
+        munit = np.sqrt(self.mass_unit /
+                        (self.time_unit**2 * self.length_unit)).to("gauss")
+        if self.cosmological_simulation:
+            self.magnetic_unit = self.quan(munit.value, "%s/a**2" % munit.units)
```
qobilidop: @tkc004 might want to have a look if this definition of code magnetic unit is correct.

tkc004: Magnetic fields in the output should be in unit of ``gauss'' (physical). We do not need to convert it from mass_unit, time_unit, etc. Even in cosmological runs, the magnetic field is in ``gauss'' (physical) and no a is needed.

tkc004: scripts/gizmo_documentation.html:MAGNETIC_FIELD_code = UnitMagneticField_in_gauss = (params.txt default = 1 Gauss)  
scripts/gizmo_documentation.html:MAGNETIC_FIELD_physical = MAGNETIC_FIELD_code  
scripts/gizmo_documentation.html:MAGNETIC_FIELD_physical = MAGNETIC_FIELD_code (note that *in* the code, co-moving units are used B_code=a_scale^2 B_phys, but in outputs these are converted to physical)

jzuhone: @tkc004 so will this always be true or is the unit user-configurable? And if it is configurable can we read it from the output?

qobilidop: I have some thoughts about this. All the code units in GIZMO (and other Gadget derivatives as well I believe) are [configurable in the parameter file](http://www.tapir.caltech.edu/~phopkins/Site/GIZMO_files/gizmo_documentation.html#params-generic-units) using constants like `UnitMagneticField_in_gauss`. But this information is not available in the output snapshots. Up until now, the code units in these frontends are set according to the defaults from the official documentation. That is fine because the code units could also be set manually if needed. The most important thing might be making sure the users are aware of this. Personally, I think it could be convenient to have an interface to set up code unit systems from the same set of constants the code requires like `ds.set_unit_system(UnitLength_in_cm, UnitMass_in_g, ...)`. @ngoldbaum's unyt PR might also be related to this discussion.

jzuhone:
@yt-fido test this please

ngoldbaum:
Fido doesn't run on the yt-4.0 branch.

jzuhone:
This is ready to go now that I've addressed the Gizmo issue and all tests are passing. Let me know if there's anything else.

jzuhone:
I agree. @tkc004 or @chummels?

tkc004:
Great, @jzuhone . I am not familiar to other aspects, but the magnetic field unit in GIZMO is correct.

jzuhone:
Thanks @tkc004. Mind hitting the "approve" button?

PR #2250, opened 2019-04-23
Enable SPH visualization of filtered particle types

jzuhone:
This is a WIP PR to enable SPH visualization of filtered particle types which have an original type which is SPH. 

A side effect of the work here will be to (at least in theory) make multiple SPH particle types possible, since the core of it is to change the private attribute `Dataset._sph_ptype` to `Dataset._sph_ptypes`, where the latter is a tuple.

Currently I'm able to make slices and projections correctly but otherwise at the moment I'm sure some things are broken. 

- [x] figure out how to handle smoothing lengths for filtered particles
- [ ] add tests
- [x] figure out what to do about gather vs. scatter smoothing
- [ ] add a note to docs

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -745,14 +745,17 @@ class YTCoveringGrid(YTSelectionContainer3D):
         fields = [f for f in fields if f not in self.field_data]
         if len(fields) == 0: return
 
-        ptype = self.ds._sph_ptype
         smoothing_style = getattr(self.ds, 'sph_smoothing_style', 'scatter')
         normalize = getattr(self.ds, 'use_sph_normalization', True)
 
         bounds, size = self._get_grid_bounds_size()
 
-        if(smoothing_style == "scatter"):
+        if smoothing_style == "scatter":
             for field in fields:
+                fi = self.ds._get_field_info(field)
+                ptype = fi.name[0]
+                if ptype not in self.ds._sph_ptypes:
+                    raise KeyError("%s is not a SPH particle type!" % ptype)
                 buff = np.zeros(size, dtype="float64")
                 if normalize:
                     buff_den = np.zeros(size, dtype="float64")
@@ -778,7 +781,6 @@ class YTCoveringGrid(YTSelectionContainer3D):
                 if normalize:
                     normalization_3d_utility(buff, buff_den)
 
-                fi = self.ds._get_field_info(field)
                 self[field] = self.ds.arr(buff, fi.units)
                 pbar.close()
 
@@ -2328,8 +2330,8 @@ class YTOctree(YTSelectionContainer3D):
         elif isinstance(fields, list):
             fields = fields[0]
 
-        sph_ptype = getattr(self.ds, '_sph_ptype', 'None')
-        if fields[0] == sph_ptype:
+        sph_ptypes = getattr(self.ds, '_sph_ptypes', 'None')
+        if fields[0] in sph_ptypes:
             smoothing_style = getattr(self.ds, 'sph_smoothing_style', 'scatter')
             normalize = getattr(self.ds, 'use_sph_normalization', True)
 
--- a/yt/data_objects/selection_data_containers.py
+++ b/yt/data_objects/selection_data_containers.py
@@ -283,16 +283,16 @@ class YTRay(YTSelectionContainer1D):
             raise KeyError(field)
 
         length = unorm(self.vec)
-        pos = self[self.ds._sph_ptype, "particle_position"]
+        pos = self[self.ds._sph_ptypes[0], "particle_position"]
         r = pos - self.start_point
         l = udot(r, self.vec/length)
 
         if field == "t":
             return l / length
 
-        hsml = self[self.ds._sph_ptype, "smoothing_length"]
-        mass = self[self.ds._sph_ptype, "particle_mass"]
-        dens = self[self.ds._sph_ptype, "density"]
+        hsml = self[self.ds._sph_ptypes[0], "smoothing_length"]
+        mass = self[self.ds._sph_ptypes[0], "particle_mass"]
+        dens = self[self.ds._sph_ptypes[0], "density"]
         # impact parameter from particle to ray
         b = np.sqrt(np.sum(r**2, axis=1) - l**2)
 
--- a/yt/frontends/gadget/io.py
+++ b/yt/frontends/gadget/io.py
@@ -67,7 +67,7 @@ class IOHandlerGadgetHDF5(IOHandlerSPH):
                 x = f["/%s/Coordinates" % ptype][si:ei, 0].astype("float64")
                 y = f["/%s/Coordinates" % ptype][si:ei, 1].astype("float64")
                 z = f["/%s/Coordinates" % ptype][si:ei, 2].astype("float64")
-                if ptype == self.ds._sph_ptype:
+                if ptype in self.ds._sph_ptypes:
                     pdtype = f["/%s/Coordinates" % ptype].dtype
                     pshape = f["/%s/Coordinates" % ptype].shape
                     hsml = self._get_smoothing_length(data_file, pdtype, pshape)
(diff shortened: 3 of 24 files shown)


jzuhone:
@qobilidop With the latest changes I pushed, this is now working, though it only works for "scatter" smoothing. 

Getting this to work for "gather" smoothing would require re-calculating the smoothing length for the filtered type (if I remember correctly), which I suspect is not a trivial addition. We could either try to implement this or take a pass on it and don't let filtered types get smoothed using "gather". What do you think?

qobilidop:
@jzuhone Thanks! I think "scatter" would already be very useful. Personally, I have only used "scatter" for visualization. So I'm okay with dealing with "gather" in a separate PR. @AshKelly knows more about how "gather" is implemented, and might have more to say.

I'll come back to test this soon.

AshKelly:
I'm going to look at this first thing Monday morning!

yt/data_objects/selection_data_containers.py
```diff
@@ -283,16 +283,16 @@ def _generate_container_field_sph(self, field):
             raise KeyError(field)
 
         length = unorm(self.vec)
-        pos = self[self.ds._sph_ptype, "particle_position"]
```
AshKelly: How we we know this will always be "_sph_ptypes[0]", why can't it be another index?

AshKelly:
I think in principle this isn't too hard to add to `gather' but I would prefer that to be a in a separate PR. I'm really looking forward to use this :) other than my confusion with the `_sph_ptypes[0]' this looks good to me!

jzuhone:
@AshKelly, good question. For now, the way the code is set up, when one loads an SPH dataset the first SPH ptype will always be the one loaded from disk. This assumes of course that the disk only has one SPH ptype, but that was already the case in yt as this was just a single string and now in this PR it is a tuple.

I think that having to support more than one SPH ptype on disk will have to come down the road.

yt/data_objects/construction_data_containers.py
```diff
@@ -2328,8 +2330,8 @@ def get_data(self, fields = None):
         elif isinstance(fields, list):
             fields = fields[0]
 
-        sph_ptype = getattr(self.ds, '_sph_ptype', 'None')
-        if fields[0] == sph_ptype:
+        sph_ptypes = getattr(self.ds, '_sph_ptypes', 'None')
```
qobilidop: Would `getattr(self.ds, '_sph_ptypes', [])` (or something like that) be better? I don't think it's likely to run into an edge case that `fields[0] in 'None'` would be `True` unexpectedly. So just a minor suggestion.

yt/frontends/gadget/io.py
```diff
@@ -67,7 +67,7 @@ def _read_particle_coords(self, chunks, ptf):
                 x = f["/%s/Coordinates" % ptype][si:ei, 0].astype("float64")
                 y = f["/%s/Coordinates" % ptype][si:ei, 1].astype("float64")
                 z = f["/%s/Coordinates" % ptype][si:ei, 2].astype("float64")
-                if ptype == self.ds._sph_ptype:
+                if ptype == self.ds._sph_ptypes[0]:
```
qobilidop: I have to look further into the context. But for now, I'm wondering if the condition should be `ptype in self.ds._sph_ptypes` instead. I have the same question for other `ptype == self.ds._sph_ptypes[0]` occurrences.

yt/geometry/coordinates/cartesian_coordinates.py
```diff
@@ -369,20 +369,20 @@ def _ortho_pixelize(self, data_source, field, bounds, size, antialias,
                     for chunk in data_source.chunks([], 'io'):
                         pixelize_sph_kernel_slice(
                             buff,
-                            chunk[ptype, px_name],
-                            chunk[ptype, py_name],
-                            chunk[ptype, 'smoothing_length'],
-                            chunk[ptype, 'particle_mass'],
+                            chunk[ptype, px_name].to('code_length'),
```
qobilidop: Would it be better to use `to_value` in these cases? Or probably it doesn't matter.

yt/geometry/coordinates/cartesian_coordinates.py
```diff
@@ -318,7 +318,7 @@ def _ortho_pixelize(self, data_source, field, bounds, size, antialias,
                             chunk[ptype, px_name].in_units('cm'),
                             chunk[ptype, py_name].in_units('cm'),
                             chunk[ptype, 'smoothing_length'].in_units('cm'),
-                            chunk[ptype, 'particle_mass'].in_units('g'),
+                            chunk[ptype, 'mass'].in_units('g'),
```
qobilidop: I'm a little bit worried about this. Projection in yt-4.0 has been working for me. Although when looking at this I started to wonder why. I'll need to do some experiments to figure that out.

yt/data_objects/construction_data_containers.py
jzuhone: Here I'm not modifying the original logic of the code, so we'd have to look carefully into the  original reason for this line.

yt/frontends/gadget/io.py
jzuhone: By the time you get down in here you will only see a `ptype` which is on-disk, so I think this logic is correct.

yt/geometry/coordinates/cartesian_coordinates.py
jzuhone: I don't think it does matter, though I note that the other arrays have units.

yt/geometry/coordinates/cartesian_coordinates.py
jzuhone: `mass` is just an alias to `particle_mass`, so there shouldn't be any issue here, correct?

qobilidop: I noticed that for SPH particle type like `'gas'`, there is only `'mass'` but no `'particle_mass'` field. Similarly, all the `'particle_xxx'` fields are renamed without the `'particle_'` prefix for the SPH particle type. This must be a decision made earlier in the yt-4.0 refactoring.

yt/geometry/coordinates/cartesian_coordinates.py
qobilidop: I guess it works because the implicit type conversion from `unyt` to `numpy` array won't change the array value.

yt/frontends/gadget/io.py
qobilidop: That answered my question. Thanks!

yt/data_objects/construction_data_containers.py
qobilidop: Yeah, I think both the original and current code should work 99.99% of the time. The original logic would only be flawed if the queried field happens to be named like `('None', 'XXX')`, and then it could pass the test `fields[0] == sph_ptype` unexpectedly. But that should be an extremely rare case if not impossible. In the modified code, the same could happen. That's why I made the suggestion.

qobilidop:
I tried the new feature locally with this example:
```python
import yt

sph_pt = 'gas'

@yt.particle_filter(requires=['temperature'], filtered_type=sph_pt)
def cool_gas(pfilter, data):
    return data[pfilter.filtered_type, 'temperature'].to_value('K') < 1e5

@yt.particle_filter(requires=['temperature'], filtered_type=sph_pt)
def hot_gas(pfilter, data):
    return data[pfilter.filtered_type, 'temperature'].to_value('K') > 1e5

ds.add_particle_filter('cool_gas')
ds.add_particle_filter('hot_gas')

_, center = ds.find_max((sph_pt, 'density'))
for pt in ['cool_gas', 'hot_gas']:
    field = (pt, 'H_nuclei_density')
    p = yt.ProjectionPlot(
        ds, 'z', field,
        center=center, width=(600, 'kpc')
    )
    p.set_unit(field, 'cm**-2')
    p.set_zlim(field, 1e18, 1e22)
    p.show()
```
And I get:
- cool gas
image
- hot gas
image

Works like a charm!

PR #2251, opened 2019-04-23
[bugfix] Don't try to sum a mask if it is None

jzuhone:
I introduced a bug in Gadget I/O in a previous PR of mine. If `mask` is `None` after selecting points from the selector, we shouldn't try to sum it. In this case, we won't need to compute the sum anyway because the loop will be continued a couple of lines down.


jzuhone:
This should probably get merged in quickly, as it breaks everything that uses Gadget's HDF5 I/O in the `yt-4.0` branch.

PR #2252, opened 2019-04-24
[bugfix] Generate smoothing length if it's needed for frontends which don't have it

jzuhone:
Another nit I discovered by poking around. Some Gadget-based frontends (e.g. Arepo) have to generate their own smoothing lengths. Normally when this field is asked for it's already been generated for data selection purposes, but there is the case of selecting all data where it hasn't. This PR ensures that the smoothing lengths are always generated if they are needed.

### DIFF
--- a/yt/frontends/gadget/io.py
+++ b/yt/frontends/gadget/io.py
@@ -128,10 +128,10 @@ class IOHandlerGadgetHDF5(IOHandlerSPH):
                 if data_file.total_particles[ptype] == 0:
                     continue
                 g = f["/%s" % ptype]
-                hsmls = None
                 if getattr(selector, 'is_all_data', False):
                     mask = slice(None, None, None)
                     mask_sum = ei-si
+                    hsmls = None
                 else:
                     coords = g["Coordinates"][si:ei].astype("float64")
                     if ptype == 'PartType0':
--- a/yt/frontends/gadget/io.py
+++ b/yt/frontends/gadget/io.py
@@ -128,6 +128,7 @@ class IOHandlerGadgetHDF5(IOHandlerSPH):
                 if data_file.total_particles[ptype] == 0:
                     continue
                 g = f["/%s" % ptype]
+                hsmls = None
                 if getattr(selector, 'is_all_data', False):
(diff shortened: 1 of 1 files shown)


yt/frontends/gadget/io.py
```diff
@@ -128,6 +128,7 @@ def _read_particle_fields(self, chunks, ptf, selector):
                 if data_file.total_particles[ptype] == 0:
                     continue
                 g = f["/%s" % ptype]
+                hsmls = None
                 if getattr(selector, 'is_all_data', False):
                     mask = slice(None, None, None)
                     mask_sum = ei-si
```
ngoldbaum: I think it might be slightly clearer to set `hsmls` to `None` in here or to call `_get_smoothing_length` here rather than below if `field == 'smoothing_length`.

jzuhone: I agree that it might be clearer, but the issue is that suppose that you have a frontend which doesn't ever get to the `if field == 'smoothing_length'` logic below (Gadget, for example, shouldn't ever need this). I think it's better to not always generate them if we don't need them when the selector is all data.

ngoldbaum: I agree that you shouldn't generate them if you don't need them, which is why I suggested only binding `hsmls` after checking if `field == 'smoothing_length'. To make that concrete, I'm suggesting making this look like:

```

                if getattr(selector, 'is_all_data', False):
                    mask = slice(None, None, None)
                    mask_sum = ei-si
                    if field == "smoothing_length":
                        hsmls = self._get_smoothing_length(...)
```

And then delete the other place where you're reading the smoothing lengths below.

Alternatively, I also think it would be clearer just to bind `hsmls` to `None` inside the if statement, as that makes it clearer that setting it to `None` is the all-data case and not just the default.

ngoldbaum: Oh wait, I'm sorry, you don't know `field` yet. Please disregard.

jzuhone: but I agree with moving `hsmls = None` inside the `if` statement, so I did it.

PR #2266, opened 2019-05-29
Fix reading wrong atom at evaluation time

claytonstrawn:
The old version of this would call `_specific_metal_density` _after_ cycling through all the atoms in the list because that was when it was actually evaluated.

Unfortunately, when it did that, `atom` refers to the last one used, which was always `Zn` instead of `O` or whatever. Now it generates a separate function for each one, making sure that when this finally gets evaluated, `atom` refers to the atom you actually want. Also, changed the units of the table at bottom, this is a much more natural way to read the tables from Iwamoto et al. 1999.


yt/frontends/art/fields.py
```diff
@@ -141,53 +141,42 @@ def _metallicity(field, data):
...
+            self.add_field(('gas','%s_nuclei_mass_density'%atom),
+                            sampling_type="cell",
+                            function=_specific_metal_density_function(atom),
                             units=unit_system["density"])
 
 
-
 # based on Iwamoto et al 1999
-# number of atoms per gram of SNIa metal
+# number of grams atom per gram of SNIa metal
```
chummels: I'm not sure what a grams atom per gram is.

yt/frontends/art/fields.py
```diff
@@ -141,53 +141,42 @@ def _metallicity(field, data):
         atoms = ['C', 'N', 'O', 'F', 'Ne', 'Na', 'Mg', \
         'Al', 'Si', 'P', 'S', 'Cl', 'Ar', 'K', 'Ca', 'Sc', \
         'Ti', 'V', 'Cr', 'Mn', 'Fe', 'Co', 'Ni', 'Cu', 'Zn']
-        for atom in atoms:
+        def _specific_metal_density_function(atom):
```
chummels: This seems problematic to have two nested functions.  Is this a mistake?

yt/frontends/art/fields.py
```diff
@@ -141,53 +141,42 @@ def _metallicity(field, data):
...
-    'Ne' : 2.70E+21, 'Na' : 6.67E+19, 'Mg' : 1.19E+21, 
-    'Al' : 1.29E+20, 'Si' : 1.02E+21, 'P'  : 9.20E+18, 
-    'S'  : 3.02E+20, 'Cl' : 7.95E+17, 'Ar' : 4.71E+19, 
-    'K'  : 4.06E+17, 'Ca' : 3.45E+19, 'Sc' : 1.20E+15, 
-    'Ti' : 6.47E+17, 'V'  : 4.62E+16, 'Cr' : 5.96E+18, 
-    'Mn' : 1.65E+18, 'Fe' : 3.82E+20, 'Co' : 2.90E+17, 
-    'Ni' : 2.40E+19, 'Cu' : 4.61E+15, 'Zn' : 6.81E+16}
-
-# taken from TRIDENT
-atomic_mass = {
```
chummels: `atomic_mass` is no longer defined?  Doesn't this get used in the code?

yt/frontends/art/fields.py
```diff
@@ -141,53 +141,42 @@ def _metallicity(field, data):
...
+    'N'  : 8.47E-07, 'O'  : 1.04E-01, 'F'  : 4.14E-10, 
+    'Ne' : 3.30E-03, 'Na' : 4.61E-05, 'Mg' : 6.25E-03, 
+    'Al' : 7.19E-04, 'Si' : 1.14E-01, 'P'  : 2.60E-04, 
+    'S'  : 6.35E-02, 'Cl' : 1.27E-04, 'Ar' : 1.14E-02, 
+    'K'  : 5.72E-05, 'Ca' : 8.71E-03, 'Sc' : 1.61E-07, 
+    'Ti' : 2.50E-04, 'V'  : 5.46E-05, 'Cr' : 6.19E-03, 
+    'Mn' : 6.47E-03, 'Fe' : 5.46E-01, 'Co' : 7.59E-04, 
+    'Ni' : 9.17E-02, 'Cu' : 2.19E-06, 'Zn' : 2.06E-05}
+
+# number of grams atom per gram of SNII metal
```
claytonstrawn:
```suggestion
# mass fraction from SNII in each atom 
```

yt/frontends/art/fields.py
claytonstrawn: It's just a mass fraction. The comment is badly worded

```suggestion
# mass fraction from SNII in each atom 
```

yt/frontends/art/fields.py
claytonstrawn: No this is actually the major fix. Before, the _specific_metal_density function would use the local variable `atom`. But since it was not evaluating until later, when it _did_ get evaluated, `atom` meant `'Zn'`, no matter which field you were querying.

Now the function `_specific_metal_density_function` creates a separate field definition function for each atom, and that will locally know the right atom to use.

yt/frontends/art/fields.py
claytonstrawn: This is needed later on in TRIDENT, but that's not what it was doing here in the `art` frontend. Changing the abundance dictionaries from "atoms per gram" to "mass fraction" as mentioned in my other reply made this unnecessary here.

claytonstrawn:
@chummels the background of this is that I added these `_specific_metal_density` fields to ART a while ago in order to take advantage in TRIDENT of the fact that ART tracks SNIa metal and SNII metal separately. However, the old version had a problem which meant it was not reading the correct "atom" when it actually evaluated `_specific_metal_density`. 

This defines a different `_specific_metal_density` function for each atom density field (that's what the outer nested function is doing, creating those functions), so that it accesses it correctly when it gets there.

matthewturk:
@claytonstrawn Did this get merged elsewhere?

claytonstrawn:
@matthewturk Yeah, I had mangled my fork's merge history somehow so I deleted it and resubmitted the change in #2267 instead.

PR #2267, opened 2019-05-30
fix art specific_metal_field definition

claytonstrawn:
edit: resubmitting PR because it was too dirty last time and I couldn't get to the bottom of it. This is just my commits with all the changes now. @chummels 

Several months ago, I added these _specific_metal_density fields to ART in order to take advantage in TRIDENT of the fact that ART tracks SNIa metal and SNII metal separately. However, the old version had a problem which meant it was not reading the correct "atom" when it actually evaluated _specific_metal_density.

This defines a different _specific_metal_density function for each atom density field (that's what the outer nested function is doing, creating those functions), so that it accesses it correctly when it gets there. The old version of this would call `_specific_metal_density` _after_ cycling through all the atoms in the list because that was when it was actually evaluated.

Unfortunately, when it did that, `atom` refers to the last one used, which was always `Zn` instead of `O` or whatever. Now it generates a separate function for each one, making sure that when this finally gets evaluated, `atom` refers to the atom you actually want. Also, changed the units of the table at bottom, this is a much more natural way to read the tables from Iwamoto et al. 1999.

### DIFF
--- a/yt/frontends/art/fields.py
+++ b/yt/frontends/art/fields.py
@@ -15,7 +15,6 @@ ART-specific fields
 
 from yt.fields.field_info_container import \
     FieldInfoContainer
-from yt.utilities.physical_constants import mh
 
 b_units = "code_magnetic"
 ra_units = "code_length / code_time**2"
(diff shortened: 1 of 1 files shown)


chummels:
Thank you for cleaning this up, @claytonstrawn .  I don't have time at this moment to review this PR, but hopefully we can get to it in the next few days.

brittonsmith:
@claytonstrawn, good fix and I like how you've done it. It looks like the tests are only failing because `yt.utilities.physical_constants.mh` is no longer needed. Once you remove that, I think this should be good to go.

PR #2270, opened 2019-06-04
[BUGFIX] [OpenPMD frontend] Allow openpmd files that don't have any particles

atmyers:
As of version 1.1.0, both mesh and particle data are optional in the OpenPMD format. This allows yt to read these datasets.


Xarthisius:
@yt-fido test this please

PR #2271, opened 2019-06-11
Cache the result of _identify_fields to improve dataset initialization speed

saethlin:
This is a follow-on to #2146; I'm trying to cut down the overhead of yt to increase its usability in similar-looking applications (datasets >10 GB, but few operations on the dataset).

This PR halves the runtime of scripts such as mentioned in #2146 on our system, for 4-file MassiveFIRE snapshots. On such snapshots yt used to redundantly scan the backing HDF5 files 243 times for what Datasets they contain.


matthewturk:
Clever idea!  Thanks for submitting this.  It looks good to me!

dnarayanan:
nice work @saethlin !

Issue #2273, opened 2019-06-12
Annotating mesh lines on semi-structured data fails


clairekope:
### Bug report

**Bug summary**

Cannot annotate mesh lines onto slices of semi-structured data. Creating issue for use with pull request which includes a fix.

**Code for reproduction**

```python
coords, connect = yt.hexahedral_connectivity(xbounds, ybounds, np.array([0,1]))
ds = yt.load_hexahedral_mesh(data, connect, coords, bbox=bbox,
                             sim_time=time, length_unit=m, mass_unit=kg,
                             time_unit=s)
slc = yt.SlicePlot(ds,'z','total_density')
slc.annotate_mesh_lines()
slc.save()
```

**Actual outcome**

```
YTPlotCallbackError: annotate_mesh_lines callback failed with the following error: ndarray is not C-contiguous
```

**Expected outcome**

Mesh lines should be annotated without issue.

**Version Information**

  * Python Version: 3.7.3
  * yt version: 3.6.dev0

Issue #2272, opened 2019-06-12
Exodus II data from ALEGRA not loading properly


clairekope:
### Bug report

**Bug summary**

Data from ALEGRA, in Exodus II format, does not load correctly. One example is below, and the other was a mesh issue. Both were found and fixed by Kris Beckwith; I'm creating this issue for use in a pull request to contribute them back.

**Code for reproduction**

```python
ds = yt.load('data_output.exo')
ds.r['all','DENSITY']
```

**Actual outcome**

```
ValueError: Buffer dtype mismatch, expected 'float64_t' but got 'float'
```

**Expected outcome**

Display the values of the `('all','DENSITY')` field

**Version Information**
  * Python Version: 3.7.3
  * yt version: 3.6.dev0
  * Other Libraries (if applicable):

PR #2274, opened 2019-06-12
Bugfixes for Exodus II and Semi-Structured Frontends

clairekope:
## PR Summary

Certain Exodus II data output would not load properly; fixes to this were supplied by Kris Beckwith. Additionally, annotating mesh lines onto slices of semi-structured data (2D forced to 3D) failed because of how the connectivity matrix was returned.

Issues #2272 and #2273 

## PR Checklist

### DIFF
--- a/yt/frontends/exodus_ii/data_structures.py
+++ b/yt/frontends/exodus_ii/data_structures.py
@@ -318,7 +318,7 @@ class ExodusIIDataset(Dataset):
             else:
                 coords = np.array([coord for coord in
                                    ds.variables["coord"][:]]).transpose().copy()
-            return coords
+            return coords.astype(np.float64)
 
     def _apply_displacement(self, coords, mesh_id):
 
--- a/yt/frontends/exodus_ii/data_structures.py
+++ b/yt/frontends/exodus_ii/data_structures.py
@@ -315,11 +315,11 @@ class ExodusIIDataset(Dataset):
             if "coord" not in ds.variables:
                 coords = np.array([ds.variables["coord%s" % ax][:]
                                    for ax in coord_axes]
-                                  ).transpose().astype(np.float64)
+                                  ).transpose().astype("f8")
             else:
                 coords = np.array([coord for coord in
                                    ds.variables["coord"][:]]
-                                  ).transpose().astype(np.float64)
+                                  ).transpose().astype("f8")
             return coords
 
     def _apply_displacement(self, coords, mesh_id):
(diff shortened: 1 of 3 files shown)


yt/frontends/exodus_ii/data_structures.py
```diff
@@ -318,7 +318,7 @@ def _read_coordinates(self):
             else:
                 coords = np.array([coord for coord in
                                    ds.variables["coord"][:]]).transpose().copy()
-            return coords
+            return coords.astype(np.float64)
```
cphyc: Since `astype` copies the data by default, we could avoid an extraneous copy here by replacing the calls to `.copy()` on line 317 and 320 by `.astype(np.float64)`.
If you want to make sure the copy is done, you can use `astype(np.float64, copy=True)` which is the same as the default but makes the intention clear.

clairekope: That's a good point! I'll go ahead and make that change.

yt/frontends/exodus_ii/data_structures.py
```diff
@@ -314,10 +314,12 @@ def _read_coordinates(self):
         with self._handle.open_ds() as ds:
             if "coord" not in ds.variables:
                 coords = np.array([ds.variables["coord%s" % ax][:]
-                                   for ax in coord_axes]).transpose().copy()
+                                   for ax in coord_axes]
+                                  ).transpose().astype(np.float64)
```
matthewturk: Great catch.  One thing I would suggest is using the string "f8" to be consistent with other places in yt, but this is not a necessary change.

yt/frontends/exodus_ii/io.py
```diff
@@ -69,7 +69,7 @@ def _read_fluid_selection(self, chunks, selector, fields, size):
                     mesh_ids = [mesh.mesh_id + 1 for mesh in self.ds.index.mesh_union]
                     objs = [mesh for mesh in self.ds.index.mesh_union]
                 else:
-                    mesh_ids = [int(ftype[-1])]
+                    mesh_ids = [int(ftype.replace("connect",""))]
```
matthewturk: I'm not sure I understand this change

matthewturk:
@clairekope I think unless you want to make the minor nits I commented, this is ready to go in, right?

clairekope:
Whoops, didn't see your earlier comments. I don't actually understand the replace call either, as that was just part of the fix that was handed to me. I'll make the "f8" change and then it's good to go in!

matthewturk:
Thank you, @clairekope !

matthewturk:
@Xarthisius has let me know that this test failure is not real, so I'm going to merge.  Thank you, @clairekope!

PR #2278, opened 2019-06-14
Simplify and correct tipsy sub-chunking.

### LINKED ISSUE #2276: Chunksize does not seem to work in some cases with tipsy
Debugging an issue with @claytonstrawn led to us determining that there is some strange interaction between `CHUNKSIZE` and tipsy that results in particle offsets that are off the end of the file, thus leading to zero-size arrays being read.  It seems that there's potentially an interaction with the smoothing length calculation here as well.

In his case, setting `CHUNKSIZE` to `256**3`, rather than `64**3`, sidestepped the problem -- so my guess is that something with the way that chunksize is splitting the file up is not being carried through correctly in the different chunking operations.

matthewturk:
Fixes #2276.

The tipsy chunking was not correctly subchunking, and this seemed to only show up when the particular ordering and sizes of the particle types was set in a specific way.  This would always have a "hard fail", so there have to my knowledge this has not been the source of any quiet failures.

What this does is simplify the particle offset calculation.  Instead of our previous way of doing things, which I could not quite parse out (and which was probably dating from before we did any subchunking) the new way is to compute a global offset for *each particle type* and then offset into that based on `data_file.start`.  I also terminate earlier when trying to read particles that aren't in a sub-chunk.

This uses data from @claytonstrawn .

My verification script, which I ran with changing `CHUNKSIZE` in `yt/geometry/particle_geometry_handler.py` to a big number and leaving it at the small number:

```python
import yt
import os

if os.path.isfile("testnihaoinput/g8.26e11/g8.26e11.00320.ewah"):
    print("removing ewah")
    os.unlink("testnihaoinput/g8.26e11/g8.26e11.00320.ewah")
if os.path.isfile("testnihaoinput/g8.26e11/g8.26e11.00320-hsml"):
    print("removing hsml")
    os.unlink("testnihaoinput/g8.26e11/g8.26e11.00320-hsml")

#ds = yt.load("testnihaoinput/g2.63e10/g2.63e10.00320")
ds = yt.load("testnihaoinput/g8.26e11/g8.26e11.00320")

dd = ds.all_data()

import h5py

f = h5py.File("output_big.h5", "w")

for ft, fn in ds.field_list:
    if ft not in f: f.create_group(ft)
    print("Creating ", ft, fn, dd[ft, fn].size)
    f[ft].create_dataset(fn, data=dd[ft, fn])
f.close()
```

The results are identical between the two.  I should note that for some reason reading "smoothing_length" doesn't work, which is a bit confusing for me.

### DIFF
--- a/yt/frontends/tipsy/io.py
+++ b/yt/frontends/tipsy/io.py
@@ -43,7 +43,7 @@ class IOHandlerTipsyBinary(IOHandlerSPH):
     _ptypes = ("Gas",
                "DarkMatter",
                "Stars")
-    _chunksize = 64 * 64 * 64
+    _chunksize = CHUNKSIZE
 
     _aux_fields = None
     _fields = (("Gas", "Mass"),
@@ -118,7 +118,9 @@ class IOHandlerTipsyBinary(IOHandlerSPH):
             tp = data_file.total_particles
             f = open(data_file.filename, "rb")
             for ptype, field_list in sorted(ptf.items(),
-                                            key=lambda a: poff[a[0]]):
+                                            key=lambda a: poff.get(a[0], -1)):
+                if data_file.total_particles[ptype] == 0:
+                    continue
                 f.seek(poff[ptype])
                 total = 0
                 while total < tp[ptype]:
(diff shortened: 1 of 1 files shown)


yt/frontends/tipsy/io.py
```diff
@@ -423,22 +425,37 @@ def _identify_fields(self, data_file):
...
             if ptype not in self._pdtypes:
+                # This means we don't have any, I think, and so we shouldn't
+                # stick it in the offsets.
                 continue
+            # Note that much of this will be computed redundantly; future
+            # refactorings could fix this.
+            global_offsets[ptype] = pos
             size = self._pdtypes[ptype].itemsize
-            field_offsets[ptype] = pos
             params = self.ds.parameters
```
brittonsmith: This line is triggering the `flake8` error since `params` is no longer used.

matthewturk: Should be fixed now!

Issue #2279, opened 2019-06-17
Accessing octree cell volumes in yt-4.0 exceeds maximum recursion depth


mccbc:
### Bug report

**Bug summary**
Attempting to access the 'volume' octree field to get octree cell volumes in yt-4.0 exceeds Python's maximum recursion depth.

**Code for reproduction**

```python
import numpy as np
import yt

ds = yt.load('snapshot_200.hdf5')
left = np.array([0, 0, 0], dtype='float64')
right = np.array([64000, 64000, 64000], dtype='float64')

# generate an octree
octree = ds.octree(left, right, n_ref=64)

# get the volume of each cell
volume = octree[('index', 'volume')]
```

**Actual outcome**

```
...
/home/cmcclellan1010/yt/yt/data_objects/construction_data_containers.pyc in get_data(self, fields)
   2340                 self.gather_smooth(fields, units, normalize)                              
   2341         elif fields[0] == 'index':                                                        
-> 2342             return self[fields]                                                           
   2343         else:                                                                             
   2344             raise NotImplementedError                                                     
                                                                                                  
/home/cmcclellan1010/yt/yt/data_objects/data_containers.pyc in __getitem__(self, key)             
    255                 return self.field_data[f]                                                 
    256             else:                                                                         
--> 257                 self.get_data(f)                                                          
    258         # fi.units is the unit expression string. We depend on the registry               
    259         # hanging off the dataset to define this unit object.                             
                                                                                              
/home/cmcclellan1010/yt/yt/data_objects/construction_data_containers.pyc in get_data(self, fields)
   2340                 self.gather_smooth(fields, units, normalize)                              
   2341         elif fields[0] == 'index':                                                        
-> 2342             return self[fields]                                                           
   2343         else:                                                                             
   2344             raise NotImplementedError                                                     
                                                                                                  
/home/cmcclellan1010/yt/yt/data_objects/data_containers.pyc in __getitem__(self, key)             
    255                 return self.field_data[f]                                                 
    256             else:                                                                         
--> 257                 self.get_data(f)                                                          
    258         # fi.units is the unit expression string. We depend on the registry               
    259         # hanging off the dataset to define this unit object.                             
                                                                                     
/home/cmcclellan1010/yt/yt/data_objects/construction_data_containers.pyc in get_data(self, fields)
   2340                 self.gather_smooth(fields, units, normalize)                          
   2341         elif fields[0] == 'index':                                                    
-> 2342             return self[fields]                                                       
[...]

PR #2284, opened 2019-06-28
[WIP] yt-4.0 added refined status and cell sizes to cyoctree 

AshKelly:
## PR Summary

Added the refined field, and the size of cells in various dimensions to the octree
e.g

```
octree = ...
```
in the case of 1 oct, refined into 8 leaves,
```
octree[('index', 'refined')]
```
returns
```
[True, False, False, False, False, False, False, False, False]
```

**this needs tests and docs**

## PR Checklist

### DIFF
--- a/yt/utilities/lib/cyoctree.pyx
+++ b/yt/utilities/lib/cyoctree.pyx
@@ -93,6 +93,7 @@ cdef struct Node:
     np.int64_t leaf_id              # This is used in depositions (maybe)
     unsigned char depth
 
+
 @cython.boundscheck(False)
 @cython.wraparound(False)
 @cython.cdivision(True)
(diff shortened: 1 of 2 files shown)


yt/utilities/lib/cyoctree.pyx
```diff
@@ -360,6 +360,66 @@ cdef class CyOctree:
...
+        cdef np.int8_t[:] refined = np.zeros(
+                (num_leaves*self._num_cells + self.nodes.size()),
+                dtype='int8'
+        )
+
+        z = 0
+        for i in range(self.nodes.size()):
+            if self.nodes[i].leaf == 0:
+                refined[z] = 1
+                z += 1
```
matthewturk: Only comment would be to remove the `else` and add a `continue` here so we drop the indentation.  But not a big deal.

matthewturk:
@AshKelly This is great -- what else needs to happen for this to go in?

AshKelly:
I need to add some tests and then I'll add the `volume` field which will fix #2279

matthewturk:
@AshKelly Anything we can do to help get this over the finish line?

dnarayanan:
@AshKelly i just checked and is it the case that #2279 still returns the "max recursion depth exceeded?"?

i'm on the `desika_octree` branch of your fork

Issue #2282, opened 2019-06-21
TypeError: 'NoneType object is not subscriptable for all_data object


jcoughlin11:
### Bug report

**Bug summary**

When trying to access fields for an all_data object created from a gadget dataset, a TypeError: 'NoneType' object is not subscriptable exception is raised. The data used is snapshot_010 from yt-project.org/data

**Code for reproduction**
```python
import yt
ds = yt.load('snapshot_010')
ad = ds.all_data()
ad[('all', 'particle_position')]
```

**Actual outcome**
http://paste.yt-project.org/show/146/

**Expected outcome**
A YTArray containing the particle position data.

**Version Information**
  * Operating System: Ubuntu 18.04
  * Python Version: 3.6.8
  * yt version: 4.0.dev0
  * Other Libraries (if applicable): 

The python I'm using is the default system python that shipped with the OS. yt was built from source by cloning the git repo.

jcoughlin11:
@matthewturk has also seen this bug.

matthewturk:
@jcoughlin11 Can you try with the fix from #2288 , which has been merged?

matthewturk:
This is fixed!  Thanks, @brittonsmith .

PR #2288, opened 2019-07-08
[bugfix] Don't try to read fields if no particles present.

brittonsmith:
This addresses Issue #2282.


matthewturk:
So it looks to me like the travis check is all done, but not reported.  I'm merging!

PR #2292, opened 2019-07-11
Use proper units for pixelization of quiver fields. Fixes #2291

### LINKED ISSUE #2291: Quiver callback doesn't work for OffAxisProjection
Following code:

```
import yt
s = yt.OffAxisSlicePlot(ds, [1,1,0], ["density"], center="c")
s.annotate_cquiver('cutting_plane_velocity_x', 'cutting_plane_velocity_y',
                   factor=10, plot_args={'color':'orange'})
s.zoom(1.5)
s.save()
```

yields:

bad

but it should produce:

good

Xarthisius:
For some reason, data used to compute vector field was passed in cgs...


matthewturk:
It looks like the failures are all from really minor shifts in contours.  I don't see how that could have happened.

matthewturk:
@Xarthisius has informed me this is a known bug in the tests, and not real!  So I'm merging.

PR #2298, opened 2019-07-18
Proposed: Remove copyright and license headers from individual source files

matthewturk:
We don't need to include copyright and licensing information in the individual source files.  This pull request removes them from all of the individual files in the yt-4.0 branch.  It only touches `.py` files and skips anything in `yt/extern`.

See my email to yt-dev for more information.


chummels:
This PR seems reasonable.  Having the copyright in every file seems to beg getting outdated.

brittonsmith:
I have no problem with this, but I am rather ignorant on these matters. Could someone else weigh in on this?

yt/config.py
```diff
@@ -1,20 +1,3 @@
-"""
-This module is very simple.  It imports the configuration
-we have written for yt.
-Everything will be returned in a global config dictionary ``ytcfg``
```
munkm: Since you're removing these strings in addition to the copyright info, do we want to put anything in our developer guide that we don't recommend having header strings in new code to preserve style?

munkm:
> I have no problem with this, but I am rather ignorant on these matters. Could someone else weigh in on this?

I approved this PR and I can't think of a reason that removing the copyright and license headers would be bad, but I'm also in the same boat as @brittonsmith and am fairly ignorant on these matters too.

yt/config.py
matthewturk: Good point

matthewturk:
@munkm Ah, I think those `from future` imports were leftover, because our "py3-only" already went in, but I must have missed removing those.

matthewturk:
OK, I think I've updated the docs in the only place we mention copyright / blurb / headers.

munkm:
Matt and I talked about this and the rest of the `print_function` stuff will be removed in a different PR, so I think this is good to go.

PR #1347, opened 2017-04-20
Adding a nearest-neighbor value field.

matthewturk:
Migrated from https://bitbucket.org/yt_analysis/yt/pull-requests/2444/adding-a-nearest-neighbor-value-field/diff

This exposes an existing set of functionality, but does not add any new smoothing methods.  Here's a notebook showing this (the results aren't amazing) on some tesselation-based data.

http://nbviewer.jupyter.org/url/use.yt/upload/548d391f

PR #2301, opened 2019-07-19
Neighbor value

matthewturk:
I'm reopening this from #1347.  It needs at least:

- [x] testing
- [x] a note in the docs

### DIFF
--- a/yt/geometry/tests/test_neighbor_search.py
+++ b/yt/geometry/tests/test_neighbor_search.py
@@ -16,7 +16,8 @@ Tests for neighbor finding
 import numpy as np
 
 from yt.fields.particle_fields import \
-    add_nearest_neighbor_field
+    add_nearest_neighbor_field, \
+    add_nearest_neighbor_value_field
 from yt.testing import \
     fake_particle_ds, \
     assert_equal, \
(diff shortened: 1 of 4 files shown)


yt/geometry/tests/test_neighbor_search.py
```diff
@@ -56,3 +57,27 @@ def test_neighbor_search():
...
+    values_in = dd["particle_mass"]
+    values_out = dd[fn]
+    for i in range(index_pos.shape[0]):
+        r2 = particle_pos[:,0]*0
+        for j in range(3):
+            DR = (index_pos[i,j] - particle_pos[:,j])
+            DRo = DR.copy()
+            DR[DRo >  ds.domain_width[j]/2.0] -= ds.domain_width[j]
+            DR[DRo < -ds.domain_width[j]/2.0] += ds.domain_width[j]
+            r2 += DR*DR
```
brittonsmith: Looks like your test is failing here because `r2` only has dimension of length.

matthewturk:
Weirdly, the test passed here.  So, time to dig into why.

matthewturk:
I think this failure was related to adding to 0 in unyt, which works on python 3 but not python 2.  Anyway, to be on the safe side, this should fix it.

PR #2304, opened 2019-07-23
[WIP] EWAH Bitmask API refactor

matthewturk:
This absolutely cannot be merged until answer testing is turned back on.  Preliminary tests suggest it work, but as we've seen before in (for instance in https://github.com/yt-project/yt/commit/3c1e1b3ce1132e6eba680a6ffe9fec421a3e8d63 ) there are ways this can be funky.  (And I might have accidentally undone that patch.)

Part of the motivation here is that one of the original design goals was to build out a library that only utilized C in the class definitions, to avoid C++.  That ended up not being the result, and so now, we can back off from it and use C++ more and rely less on the way we did the pointer shuffling.

Long-term, this should enable using the `BoolArrayCollection` in more places, and even the `FileBitmasks` stuff as well.


matthewturk:
I'm converting this to draft until after 4.0 goes out so that it won't show up in my filters.  I do want to revisit it.

PR #2314, opened 2019-08-06
Updates ARTIO frontend to support compressed files.

ngnedin:
This PR updates the artio reader to support new compressed artio files. 

In original artio files the variables ...refined[i] took only values 1 and 0. In new, compressed files a value of -1 is added. Hence, the artio reader is updated to replace all checks "if ( ...lrefined[i] )" with "if ( ...lrefined[i] > 0)".

This PR does not affect existing artio files in any way.


ngnedin:
The failed continuous-integration test has nothing to do with the changes I made. There is a bug in your testing suite.

jzuhone:
Hi @ngnedin,

Yes, unfortunately there is an issue currently with NumPy 1.17 which is used by the Appveyor tests and causes them to fail. As you said, you can safely ignore it.

PR #2321, opened 2019-09-02
Add support for AMRVAC native data files

neutrinoceros:
# PR Summary

We propose a new frontend for yt to support MPI-AMRVAC (http://amrvac.org/index.html)
Most functionalities are hopefully ready for production, however a few issues are identified and should be adressed in this PR or later as separate issues. (see last section)

KH2D

## Testing
data samples can be fetched here
https://www.dropbox.com/s/45sbkq57neacbu8/datasets_amrvac.tar.gz?dl=0

and should be registered to https://github.com/yt-project/website/blob/master/data/datafiles.json

tests are located in `yt/frontends/amrvac/tests/test_unit.py`
and currently refer to a local (not bundled in) data directory (`tmpdir`). This obvious needs some cleanup before final validation.

## PR Checklist

**note : I do not know if the code is documented enough. Given that it is subclassing existing class from yt, it should be clear. If not, let us know.**

## Known issues

- `plot.annotate_velocities()` breaks (crash)
- ~default units displayed for figure axes should be "code units" instead of cgs.~ SOLVED ✅ 

- ~`ProjectionPlot` breaks : all patches are drawn in a corner of the figure frame while the grids seem correctly parsed~ SOLVED ✅ 
```
# example with a cartesian dataset
ds = yt.load("bw_3d0000.dat")
projplot = yt.ProjectionPlot(ds, "z", "rho")
```
projplot


cphyc:
Hi !
Thanks for the PR, that looks great! 

I haven't reviewed the PR (prepping for my defense!) but it certainly looks neat (and tested, yay!).

If you want to get the dataset uploaded on the yt website, you can submit a PR on https://github.com/yt-project/website/ (see for example https://github.com/yt-project/website/pull/59 for a PR adding a new dataset for RAMSES). The upload of the dataset will be uploaded by a member of the team.
I don't know whether it should come as separate datasets or a large one though.

Anyway, thanks for the work!

neutrinoceros:
So continuous integration fails because data files out are not properly registered

neutrinoceros:
Alright so the failure in continuous integration indeed comes from lack of proper registration for our datafiles. Can anyone please clarify the format I should submit them in ? Is one .tar.gz archive right ?

matthewturk:
@neutrinoceros Thanks for submitting!  I'll check on how the data files are registered -- I think it's one tar.gz for everything, that extracts into a subdirectory.

My guess is that the projection issue is related to the use of `global_index` somewhere.  I'm going to check out this PR and take a look.

Again, thank you!  This is awesome!  And, as we discussed in the slack channel, I think it's entirely reasonable to have quick iteration on this frontend once it's in a basically ready state.

matthewturk:
As a quick update, my ideas as to why the projection is not working were not correct!  The global index values seem right, and so do the values for the px, py, pdx, pdy.  The widths of slices and projections are the same, and the px, py values seem about right.  So I'm not 100% sure yet, but I am going to keep checking.

neutrinoceros:
I just submitted a PR to the website for our datasets.
https://github.com/yt-project/website/pull/62

I am unsure I did it correctly though. :/

matthewturk:
Hi @neutrinoceros! I'm digging in, and what I'm finding -- after a fair bit of checking! -- is that for some reason, most of the `projplot.data_source['amrvac','rho']` field is zero.  So it's plotting the full dataset, *but*, it thinks most of it's zero.

I'm working on why just now, and I hope to have a solution soon.

matthewturk:
OK, I figured it out!  I believe it has to do with the way that you're compensating for the difference in 0 and 1 indexing of the root grid level.

Just to make sure I am on the right track: `bw_3d0000.dat` should have 32 zones on its root level (i.e., your level 1, yt's level 0), right?

matthewturk:
What I've determined is that yt is getting confused by the disconnected grids that don't have parents -- which is all of them!  I looked at the source code for `amrvac` and it looks like the information for the grids never gets written out at all, which makes sense since you can reconstruct it all from the information that is in the dataset (and it's not strictly necessary anyway).

What this has led me to do is to make a few minor changes, which I have pushed to the branch `amrvac_support` in my fork, `matthewturk/yt` .  I'm happy to push them to your fork, but felt I should ask first.

What I've done is 1) overridden the `get_global_startindex()` method to know that for amrvac, not all grids without `Parent` attributes are root level grids, and I've also changed `domain_dimensions` from being the dimensions on the max level to being the dimensions on level 0 (in yt notation).

I believe that the projection i sthe only place that these types of errors will show up, but regardless, this should fix it globally.

Let me know if that looks alright to you!

matthewturk:
I pushed that change to your fork (as per your suggestion!) and I've also pushed a flake8 `noqa` change, to make travis happy(ier).

neutrinoceros:
Nice. I just tried it out and it indeed fixed the projection issue, magic 🎩 🐰 !

n-claes:
Hi @matthewturk, thanks for looking into this! I can also confirm that the projection plot works like a charm now, I just tried it out on one of my own datasets.

However, today I encountered another issue when trying to create a volume rendering, and I was hoping the issue was solved with the above fix but sadly it still persists. I loaded in the 3D dataset, called `sc = yt.create_scene(ds, field=('gas', 'density'))` and finally `sc.render()`.
Calling `render()` literally cluttered my console with hundreds of messages `Failed to split grids`. 

This actually finishes without crashing (except for the warnings), however, `sc.show()` _does_ crash, I've added a screenshot of the errors that it throws below. Maybe this is somehow related to the above issue concerning the grids? The crash can be reproduced using the `bw_0003.dat` dataset, this does not give the `failure to split grids` warning though for some reason.
render_crash1
render_crash2

matthewturk:
So I *think* what is happening is that it's failing to split grids because it expects there to be parents.  But, on the other hand, there not being parents means we don't need to even assemble them in the first place as done before.  So I think there is a solution that we could do, which would be to just add them all as "level 0" grids to the kdtree.

I don't have time just now to do this but I will add it to my list!

n-claes:
Great, thanks! I don't know if this is relevant for the above issue, but one of the things that can happen in an AMRVAC dataset is that the simulation is set up with for example 5 levels of AMR, but everything is on the two finest levels (so yt levels 3 and 4) due to some refinement criteria during runtime. Levels 0, 1 and 2 will have no blocks in this case.

Children/parent information is indeed never stored in the `dat` file, it's true that the code reconstructs this information when it reads in the dataset.

matthewturk:
Thanks for that info -- I think with these changes, that case should work
as well.

Do the fields for parent nodes get stored in the dat file, or does that
also get reconstructed at runtime?

n-claes:
> Thanks for that info -- I think with these changes, that case should work as well. Do the fields for parent nodes get stored in the dat file, or does that also get reconstructed at runtime?
> […](#)

Nope, everything gets reconstructed at initialisation, only the blocks that effectively contain data are saved in the dat file. Everything higher up in the blocktree is just empty.

yt/frontends/amrvac/tests/test_unit.py
```diff
@@ -0,0 +1,135 @@
+import numpy as np
+
+import yt
+from yt.utilities.parameter_file_storage import output_type_registry
+
+# todo : remove tmpdir
+tmpdir = "data/"
```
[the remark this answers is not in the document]
neutrinoceros: I don't know if I am doing this right but I tried applying the decorator on two tests for starters, and put them in tests/test_output.py
Am I on the right track ?

yt/frontends/amrvac/tests/test_output.py
```diff
@@ -0,0 +1,39 @@
...
+from yt.utilities.answer_testing.framework import requires_ds, data_dir_load
+
+blastwave_cartesian_3D = "amrvac/bw_3d0000.dat"
+blastwave_polar_2D = "amrvac/bw_polar_2D0000.dat"
+blastwave_spherical_2D = "amrvac/bw_2d0000.dat"
+blastwave_cylindrical_3D = "amrvac/bw_cylindrical_3D0000.dat"
+khi_cartesian_2D = "amrvac/kh_2D0000.dat"
+khi_cartesian_3D = "amrvac/kh_3D0000.dat"
+jet_cylindrical_25D = "amrvac/Jet0003.dat"
+riemann_cartesian_175D = "amrvac/R_1d0005.dat"
```
munkm: Since your tests only require the `blastwave_cartesian_3D` dataset, I think you don't need these other definitions in the file. If you add tests for these other datasets, you can add the definitions as needed.

yt/frontends/amrvac/tests/test_output.py
```diff
@@ -0,0 +1,39 @@
+import numpy as np
+
+import yt
+from yt.utilities.parameter_file_storage import output_type_registry
+from yt.utilities.answer_testing.framework import requires_ds, data_dir_load
+
+blastwave_cartesian_3D = "amrvac/bw_3d0000.dat"
```
munkm: The path to your datafile looks good!

yt/frontends/amrvac/tests/test_output.py
neutrinoceros: sure. I'll either clean this up or figure out how to move our tests there. I don't think I ever used generators before so there's definitely going to be a little challenge for me here but I'm on it.

yt/frontends/amrvac/tests/test_output.py
```diff
@@ -0,0 +1,100 @@
...
+        assert isinstance(g.RightEdge, yt.units.yt_array.YTArray)
+        assert isinstance(g.ActiveDimensions, np.ndarray)
+        assert isinstance(g.Level, (np.int32, np.int64, int))
+
+@requires_ds(blastwave_polar_2D)
+def test_bw_polar_2d():
+    ds = data_dir_load(blastwave_polar_2D)
+    for test in small_patch_amr(ds, ds.field_list):
+        test_bw_polar_2d.__name__ = test.description
+        yield test
```
munkm: This is really nitpicky, but would you mind adding an extra line here between the tests?

yt/frontends/amrvac/tests/test_output.py
```diff
@@ -0,0 +1,100 @@
...
+    """Check for correct box size, see bw_3d.par"""
+    ds = data_dir_load(blastwave_cartesian_3D)
+    for lb in ds.domain_left_edge:
+        assert int(lb) == 0
+    for rb in ds.domain_right_edge:
+        assert int(rb) == 2
+    for w in ds.domain_width:
+        assert int(w) == 2
+
+@requires_ds(blastwave_cartesian_3D)
```
munkm: Because this test (and the `test_domain_size` test) only do assertions on the data (and don't yield a test), the decorator should be updated on both to `@requires_file()` rather than `@requires_ds()`. Sorry I missed this the other day.

yt/frontends/amrvac/tests/test_output.py
neutrinoceros: sure :)

neutrinoceros:
> Hi @neutrinoceros thank you so much for your PR! I haven't quite gotten to do a full review yet, but at this point you'll need to modify some of your tests. We have your data in the testing framework now so the paths to the data will need to be modified. Since your tests require this data, you'll also need to decorate your tests with the `@requires_ds()` decorator. We have a bit of documentation about writing answer tests for a frontend, which you can find here: https://yt-project.org/doc/developing/testing.html#how-to-write-answer-tests-for-a-frontend. I'd suggest restructuring your tests to follow the format documented there. I hope that'll help, but let me know if you have any questions.

@munkm I think this is now completed. Let me know if there's anything left to be done on this.

neutrinoceros:
wow I didn't catch that we actually have 582 falling tests. I *guess* that task is not over then. @munkm, any idea what's wrong with how things are done now ?

matthewturk:
My personal guess is that the majority of these are *not* your fault.  I
will investigate and get back to you!

neutrinoceros:
I'm noting that which parts of the CI specifically fail vary from commit to commit and don't seem to be caused by my changes (see for instance 11c70db, which only contain a typo-correction and suddenly fails the Python2.7 unit tests even though the previous version passed.

neutrinoceros:
also the codecov tools yields weird results... is it comparing the the branch to the latest master's build ? Should I rebase on master or merge master to the PR branch ?

munkm:
@neutrinoceros the Python 2.7 failures on Travis you described should be fixed on master now. If you merge or rebase with master, those failures shouldn't show up randomly.

neutrinoceros:
alright, then it is merged !

neutrinoceros:
important update: we rewrote the unit-handling sections of the frontend. It is now conform to other frontends in terms of behaviour.

I imagine that recent changes may have altered the results of answer-tests, so I expect that stored answer would need to be regenerated. Who has control over this ? (is it me ?)

neutrinoceros:
and apparently I broke projection plots along the line. Specifically, it broke in ee2e1a06d (thank git bisect)

neutrinoceros:
another important note: the issue volume rendering seems fixed since at least b2814eaa8.
(maybe a bit earlier but hard to tell since other bugs with unit parsing cause crashes)

neutrinoceros:
Reverting ee2e1a0 indeed fixes projection plots (without breaking volume rendering) , but breaks `ds.covering_grid()` for datasets with less than 3D dimensionality (I discussed this on slack). I'll take a closer look and hopefully find a way to make everything work at the same time

neutrinoceros:
I now believe that ee2e1a0 was indeed a necessary bugfix. The error I'm now getting when trying to proceed projection plots looks like this 
Screenshot 2019-10-26 at 18 17 24
Any clue ?

This is the last coding issue (that we currently know of). After that it's just a matter of cleaning the answer tests so we finally pass the builds and CI.

neutrinoceros:
Fixed ! I now believe the PR is mature for a proper review, as we solved every blocking issue we encountered so far. 🎉
I also added the required documentation for the website.

tests/tests.yaml
```diff
@@ -1,4 +1,7 @@
 answer_tests:
+
+  local_amrvac_001:
+    - yt/frontends/amrvac/tests/test_outputs.py
```
munkm:
```suggestion
    - yt/frontends/amrvac/tests/test_outputs.py
    
```

Xarthisius:
Docs failure is caused by memory leak. When run on this branch:

```
python doc/source/cookbook/multiplot_2x2.py
```

for some reason eats up whole available RAM on my laptop, which doesn't happen on `master`

yt/frontends/amrvac/data_structures.py
```diff
@@ -0,0 +1,255 @@
...
+        self.refine_by = 2
+
+    @classmethod
+    def _is_valid(self, *args, **kwargs):
+        """At load time, check whether data is recognized as AMRVAC formatted."""
+        # required class method
+        validation = False
+        try:
+            with open(args[0], "rb") as istream:
+                get_header(istream)
```
Xarthisius: What you're doing here is very unsafe. You're not actually checking if that's AMRVAC format, but instead you're trying to parse the file as if it *were* AMRVAC file. The only check you have there is:

```
if struct.unpack(fmt, istream.read(struct.calcsize(fmt)) < 3:
    raise IOError("Unsupported AMRVAC .dat file version: %d", h['datfile_version'])
```

Apparently reading a random bit from a file results in a number less than 3 more often than not. It causes allocation of <insert random big int> few lines later.

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -0,0 +1,103 @@
...
+blastwave_polar_2D = "amrvac/bw_polar_2D0000.dat"
+blastwave_spherical_2D = "amrvac/bw_2d0000.dat"
+blastwave_cylindrical_3D = "amrvac/bw_cylindrical_3D0000.dat"
+khi_cartesian_2D = "amrvac/kh_2D0000.dat"
+khi_cartesian_3D = "amrvac/kh_3D0000.dat"
+jet_cylindrical_25D = "amrvac/Jet0003.dat"
+riemann_cartesian_175D = "amrvac/R_1d0005.dat"
+
+@requires_ds(blastwave_cartesian_3D)
+def test_domain_size(self):
```
munkm:
```suggestion
def test_domain_size():
```

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -0,0 +1,103 @@
...
+    ds = data_dir_load(blastwave_cartesian_3D)
+    for lb in ds.domain_left_edge:
+        assert int(lb) == 0
+    for rb in ds.domain_right_edge:
+        assert int(rb) == 2
+    for w in ds.domain_width:
+        assert int(w) == 2
+
+@requires_file(blastwave_cartesian_3D)
+def test_grid_attributes(self):
```
munkm:
```suggestion
def test_grid_attributes():
```

yt/frontends/amrvac/data_structures.py
neutrinoceros: wow, I can't believe I let that as it is, thank you so much for pointing this out !

yt/frontends/amrvac/tests/test_outputs.py
neutrinoceros: actually no I never knew it was even *possible* to run the answer tests on my laptop. Will the `@requires_file` decorator download data from the yt-data server ? Is it as simple as running `nosetest` locally ?

neutrinoceros: apparently not... To clarify, I was using pytest to run tests locally before submitting this PR. Since the tests were updated with the `@requires_ds` decorator I've been committing in the blind :-(

yt/frontends/amrvac/data_structures.py
neutrinoceros: 7e3a432 adresses this problem but is probably a bit too optimistic.

yt/frontends/amrvac/data_structures.py
```diff
@@ -0,0 +1,249 @@
...
+                                            units_override=units_override, unit_system=unit_system)
+        self.fluid_types += ('amrvac',)
+        # refinement factor between a grid and its subgrid
+        self.refine_by = 2
+
+    @classmethod
+    def _is_valid(self, *args, **kwargs):
+        """At load time, check whether data is recognized as AMRVAC formatted."""
+        # required class method
+        return args[0].endswith(".dat")
```
matthewturk: Whoops, I left this comment on the commit, but will replicate here.

There might be a way to do this that's slightly less optimistic, but also more deterministic. What about implementing a little bit of the get_header routine here:

 1. Return False if it doesn't end with .dat
 2. Return False if the version is not "good"
 3. Check if the values of what would be offset_tree and offset_blocks are both less than the total file size

That might be enough to drop false positives to zero.

neutrinoceros: excellent ! I just implemented exactly this, but in a form that only has one `return` statement (matter of personal taste) :)

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -0,0 +1,104 @@
...
+#blastwave_cartesian_3D = "amrvac/bw_3d0000.dat"
+#blastwave_polar_2D = "amrvac/bw_polar_2D0000.dat"
+#blastwave_cylindrical_3D = "amrvac/bw_cylindrical_3D0000.dat"
+
+@requires_file(khi_cartesian_2D)
+def test_AMRVACDataset():
+    assert isinstance(data_dir_load(khi_cartesian_2D), AMRVACDataset)
+
+
+#@requires_ds(blastwave_cartesian_3D)
```
Xarthisius: For the future reference: Both `requires_ds` and `requires_file` are meant to disable a given test if the required file is missing, so there's no need to comment them out. Now that the files were uploaded to testing server, it'd have been sufficient to just re-trigger tests, if this part hadn't been commented out.

neutrinoceros: nice, good to know !
btw, there is still one commented test that I need to get my head around because it fails to load the jet data set locally (just noticed that yesterday)

matthewturk:
@yt-fido please test this

matthewturk:
Looks like there were some bad matplotlib cache elements.

neutrinoceros:
Possibly my bad for rebasing the branch

Xarthisius:
@yt-fido test this please.

matthewturk:
@munkm Are your concerns satisfied with PR as-is?  I think they might be?

munkm:
Yep! This looks good to me now.

matthewturk:
So, any reason we can't push the button???  @Xarthisius ?

munkm:
🎉 🎆 😎 Congratulations @neutrinoceros and @n-claes! You've put a lot of hard work into this PR and we really appreciate it!

PR #2323, opened 2019-09-03
[bugfix] Fix code magnetic units, etc

jzuhone:
This PR fixes incorrect conversions between code units for magnetic fields and other magnetic field units. It ensures that no matter what unit system we are using and no matter what unit the magnetic field has, the conversions happen correctly. I added a test for this.

I threw in two other bug fixes for good measure:

* `OWLSFieldInfo` did not define ionized hydrogen fields.
* Deriving the mass field for Gadget datasets (and their derivatives) was not happening correctly when 1) the mass value is in the mass array and 2) the `is_all_data` selector is being used.


jzuhone:
Can we get one more look at this?

PR #2329, opened 2019-09-16
Backporting unyt PR 102 for clip

matthewturk:
This backports https://github.com/yt-project/unyt/pull/102/files to fix
some issues with `np.clip`.


matthewturk:
OK, I've commented `mul` rather than removing, as we may eventually need it if we do more substantial backports.

ngoldbaum:
The appveyor failure looks like some unrelated issue in conda, if the tests are passing that means clip is working so I’d say that this is good to merge.

matthewturk:
There are issues I'm working to fix.

PR #2331, opened 2019-09-19
Return p0 alias fields.

brittonsmith:
This undoes a change that removed the aliasing of `<element>_p0_<field>` to `<element>_<field>` (e.g., H_p0_density to H_density). This has caused a breakage in Trident, where the `p0` field names are used. My understanding was that the decision made in YTEP-0003 to leave out the `p0` in neutral ions was now largely considered to be a mistake. Either way, unless this breaks something in yt, I'd like to bring this back.

@jzuhone made this change. John, is there a reason we can't bring this back? Let me know if I've missed something.

### DIFF
--- a/yt/fields/species_fields.py
+++ b/yt/fields/species_fields.py
@@ -243,4 +243,10 @@ def setup_species_fields(registry, ftype = "gas", slice_info = None):
             # Skip it
             continue
         func(registry, ftype, species)
+        # Adds aliases for all neutral species from their raw "MM_"
+        # species to "MM_p0_" species to be explicit.
+        # See YTEP-0003 for more details.
+        if ChemicalFormula(species).charge == 0:
+            alias_species = "%s_p0" % species.split('_')[0]
+            add_species_aliases(registry, "gas", alias_species, species)
     add_nuclei_density_fields(registry, ftype)


matthewturk:
I missed that change, but am generally in favor of the more explicit inclusion of `p0`

jzuhone:
IIRC this would break something in the changes I made, but I won’t be able to know for sure until maybe later tonight or tomorrow when I can look at this in detail. What I remember is that allowing the p0 fields to alias like this was inconsistent with the new assumptions about full ionization.

jzuhone:
What exactly is the “more explicit inclusion” of p0? Is that in favor of the aliases or against?

matthewturk:
I think we *should* have `_p0` included to refer to neutral species.  That way one could do something like the sum of all `_p*` and `_m*` fields to get the total.

jzuhone:
Right, but what does that imply for these aliases?

matthewturk:
My reading of the PR is that I am in favor of this change.

jzuhone:
So what field then would be the sum over all (say) C ionization states? Currently it would be C density, but this would alias C_p0_density to that.

matthewturk:
@jzuhone Oh, then my reading of this is *wrong* -- I was under the assumption here the `C_p0` would *only* refer to the neutral species.

jzuhone:
@matthewturk that's right, it does. The change here would alias (say) `C_p0_density` to `C_density`. But some frontends explicitly set `C_p0_density`, right? I just made a change to `OWLSFieldInfo` which does exactly this. 

Thinking about it more, I'm actually not sure this is such a big problem, but it does pose the problem that if simulations define `C_density` and we alias `C_p0_density` to it by default (when the density of neutrals isn't defined), it conflicts with the new-ish default assumption of full ionization. 

And I answer my own question from above, it's `C_nuclei_density` that is the sum of all the ionization states.

brittonsmith:
Eventually, I'd like to see `X_p0_density` become the primary derived field/alias for neutral X density and for `X_density` to be left out entirely.

The specific problem I am facing right now is that Enzo outputs a field called `HI_Density`, which is aliased to `H_density`, and which used to be aliased to `H_p0_density`. I'd like to revisit YTEP-0003 one day and make `HI_density` alias `H_p0_density` directly, but for now, returning this to the way it was will suffice.

Are there any objections to pulling this in?

jzuhone:
> Eventually, I'd like to see X_p0_density become the primary derived field/alias for neutral X density and for X_density to be left out entirely.
> I'd like to revisit YTEP-0003 one day and make HI_density alias H_p0_density directly, but for now, returning this to the way it was will suffice.

We're at a major version release, so if we want to make big changes like that, we should do it now, in my view at least. Everything would be much more clear if we did this now. I'm happy to help with it.

brittonsmith:
Yes, you're right. Let's do it. Ok, I will commit to doing this as soon as I can. If possible, I'd like to get the Trident workflow fixed first. Are you opposed to merging this now?

jzuhone:
I need to figure out what merging this will break first, which I will try to do before I leave work this afternoon.

brittonsmith:
PR #2336 is a lasting solution to this problem. Closing this.

PR #2333, opened 2019-09-20
[BUGFIX] Fix build failure on Windows

matthewturk:
At present, this adds a diagnostic test for the `np.clip` function that
should help us to identify why builds are now failing on appveyor for
python 3.6 with recursion counts at `yt/frontends/ytdata/io.py` line
351.

### DIFF
--- a/yt/units/tests/test_ytarray.py
+++ b/yt/units/tests/test_ytarray.py
@@ -1417,3 +1417,14 @@ def test_clip():
 
     assert_array_equal(data, answer)
     assert data.units == answer.units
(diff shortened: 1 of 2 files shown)


matthewturk:
To be clear, this does not yet fix the failure, but *hopefully* will at least create an isolated test to cause a failure.  I cannot replicate the failure on my machine.

matthewturk:
After a few false starts, this is now ready to go.  Thanks to @ngoldbaum for figuring it out in `unyt`.

PR #2334, opened 2019-09-20
Use mirror copy of SDF data

Xarthisius:
Trying to speed up SDF test by using NCSA mirror.

### DIFF
--- a/yt/frontends/sdf/tests/test_outputs.py
+++ b/yt/frontends/sdf/tests/test_outputs.py
@@ -21,8 +21,9 @@ from yt.extern.six.moves import urllib
 
 
 _fields = (('deposit', 'all_cic'))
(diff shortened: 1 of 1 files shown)


matthewturk:
LGTM

yt/frontends/sdf/tests/test_outputs.py
```diff
@@ -40,7 +41,7 @@ def internet_on():
 def test_scivis():
     if not internet_on():
```
munkm: the internet_on function has a call to the `scivis_data`. If this PR switches to the mirror should that function be updated to the mirror?

Xarthisius: good catch! Thanks!

munkm:
It looks like the line with `assert_equal(str(ds), "ds14_scivis_0128_e4_dt04_1.0000")` is causing a failure because the string of the new ds is associated with the yt.upload name `744abba3`.

matthewturk:
Since we're not having a lot of success with getting this to work with the mirror, I am going to push to this PR a change that disables the test.

PR #2336, opened 2019-09-21
reverse neutral species alias fields

brittonsmith:
This is a replacement for PR #2331 .

As discussed in the comment thread there, this gives neutral species fields names of `X_p0_<field>` instead of `X_<field>`. Previously, `X_p0_<field>` was created as an alias to `X_<field>`, largely as a nod to mistakes made in YTEP-0003.

This PR makes the `p0` field the primary naming convention and adds `X_<field>` as an alias with a deprecation warning.

Looking through other frontends, I think it's clear we need to do this. Arepo and OWLS already appear to define `X_density` as we have defined `X_nuclei_density`. It's best we get away from `X_<field>` signifying the neutral species as soon as possible.


jzuhone:
@brittonsmith I'll have a look at this tonight, sorry for the delay on responding to this overall issue.

brittonsmith:
@jzuhone, no problem, just had some time on my hands.

I forgot f strings aren't supported in python 3.5. One might argue that if we're going to take advantage of later python features, yt-4.0 is the place to do it. @matthewturk, thoughts?

jzuhone:
Hi @brittonsmith,

I think that this is good. My only question is, are we getting rid of the `X_density` (and associated) fields altogether? For some frontends there might not be any notion whatsoever of an ionization state. Might it be a good idea to let the frontend define them anyway? And in that case we would just have to rip the bandaid off completely now and not allow any aliases between those and the p0 fields. 

What do you think?

jzuhone:
So for number densities we'd do nuclei density for frontends with no ionization information, to be consistent with the others. But it would still be nice to be able to define `X_fraction` and `X_density` for those. If you like I can issue a separate PR to do this after this one (though I would simply remove the aliases you're allowing for here in the process).

I know we want to get these `p0` fields back in very quickly for Trident, so I'd like to find a way to merge this today.

brittonsmith:
@jzuhone, I think it's important that `X_density` have as consistent a definition as possible. If you'd like to add that and `X_fraction` in a following PR, that should be fine. It's true some frontends are using it. Thanks for working with me on this!

PR #2338, opened 2019-09-25
[bugfix] Fixing issues with particles

jzuhone:
This is one of those PRs which basically led one down a rabbit hole, finding several issues along the way. 

1. The `DiskSelector` was computing selections incorrectly for SPH particles. 
2. The `relative_particle_position` and `relative_particle_velocity` fields were using `modify_reference_frame` to compute the relative fields. This takes a normal vector and then returns the relative coordinates in a rotated coordinate system, in which the new normal vector is the z-axis. However, the fields for particle positions in spherical and cylindrical coordinates (such as radius, etc.) were assuming the `normal` field parameter was the original one instead and using the relative fields, which gave incorrect results. I'm changing here the way the relative fields are computed, consistent with the way we handle relative fields from grid datasets, by only subtracting the center off. As far as I can tell, this only occurs in `particle_fields.py`.  
3. For some reason merges with master did not get the fixes from PR #2043 into the `yt-4.0` branch, this adds them back in.

Open to suggestions about new tests.

## PR Checklist

- [ ] Adds a test for any bugs fixed.

### DIFF
--- a/yt/geometry/selection_routines.pyx
+++ b/yt/geometry/selection_routines.pyx
@@ -1230,7 +1230,7 @@ cdef class DiskSelector(SelectorObject):
         h = d = 0
         for i in range(3):
             temp = self.periodic_difference(pos[i], self.center[i], i)
-            h += pos[i] * self.norm_vec[i]
+            h += temp * self.norm_vec[i]
             d += temp*temp
         r2 = (d - h*h)
         d = self.radius+radius
--- a/yt/utilities/lib/misc_utilities.pyx
+++ b/yt/utilities/lib/misc_utilities.pyx
@@ -568,7 +568,8 @@ def get_box_grids_below_level(
 @cython.cdivision(True)
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def obtain_position_vector(data):
+def obtain_position_vector(
+    data, field_names = ('x', 'y', 'z')):
     # This is just to let the pointers exist and whatnot.  We can't cdef them
     # inside conditionals.
     cdef np.ndarray[np.float64_t, ndim=1] xf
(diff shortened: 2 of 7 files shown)


matthewturk:
Thank you for this.  Tomorrow on the triage call let's go over this, but it
seems quite important.

yt/geometry/selection_routines.pyx
```diff
@@ -1230,7 +1230,7 @@ cdef class DiskSelector(SelectorObject):
         h = d = 0
         for i in range(3):
             temp = self.periodic_difference(pos[i], self.center[i], i)
-            h += pos[i] * self.norm_vec[i]
+            h += temp * self.norm_vec[i]
```
matthewturk: ouch, great catch.

yt/utilities/lib/misc_utilities.pyx
```diff
@@ -568,7 +568,8 @@ def get_box_grids_below_level(
 @cython.cdivision(True)
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def obtain_position_vector(data):
+def obtain_position_vector(
+    data, field_names = ('x', 'y', 'z')):
```
matthewturk: should maybe add a note that this may result in an ambiguous access

matthewturk:
Some of the angular momentum stuff is included in #2172 but this is a set of great catches.  Thanks.

yt/utilities/lib/misc_utilities.pyx
munkm: I agree with this.

munkm:
@jzuhone is going to add a couple tests to this and then it'll be ready to go.

jzuhone:
After the tests pass, this is ready to go. I think disk objects need to be added to the AMR and SPH answer tests, which would have to come in another PR.

PR #2339, opened 2019-10-01
Add turbo colormap.

brittonsmith:
This adds the turbo colormap, written by Anton Mikhailov. Turbo is an improved version of jet that is colorblind safe and lacks the bright banding that gives the perception of false structure. An example is below.
DD0046_Projection_x_density


ngoldbaum:
What happens if turbo gets added to matplotlib? Would that be an issue? I think it might eventually: https://github.com/matplotlib/matplotlib/issues/15091

matthewturk:
We probably just yank it if it does, but I think ours would be registered
"over top" of the other.

Xarthisius:
@yt-fido test this please

PR #2347, opened 2019-10-18
Ramses read namelist

cphyc:
## PR Summary

This adds the automatic loading of the `namelist.txt` file which contains the parameter file RAMSES used to produce the output.
This relies on the package `f90nml` being installed (for the tests). If the package is absent, `ds.parameters['namelist']` contains an error message.

## PR Checklist

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -595,6 +595,9 @@ class RAMSESDataset(Dataset):
                     nml = f90nml.read(f)
             except ImportError as e:
                 nml = "An error occurred when reading the namelist: %s" % str(e)
+            except ValueError as e:
+                mylog.warn("Could not parse `namelist.txt` file as it was malformed.")
+                return
 
             self.parameters['namelist'] = nml
 
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -37,6 +37,7 @@ from yt.data_objects.octree_subset import \
 from yt.data_objects.particle_filters import add_particle_filter
 
 from yt.utilities.physical_constants import mp, kb
+from yt.utilities.on_demand_imports import _f90nml as f90nml
 from .definitions import ramses_header, field_aliases, particle_families
 from .fields import \
     RAMSESFieldInfo, _X
@@ -581,6 +582,21 @@ class RAMSESDataset(Dataset):
 
         if self.num_groups > 0:
             self.group_size = rheader['ncpu'] // self.num_groups
+
+        # Read namelist.txt file (if any)
+        self.read_namelist()
+
+    def read_namelist(self):
+        """Read the namelist.txt file in the output folder, if present"""
+        namelist_file = os.path.join(self.root_folder, 'namelist.txt')
+        if os.path.exists(namelist_file):
+            try:
+                with open(namelist_file, 'r') as f:
+                    nml = f90nml.read(f)
+            except ImportError as e:
+                nml = "An error occurred when reading the namelist: %s" % str(e)
+
+            self.parameters['namelist'] = nml
 
     @classmethod
     def _is_valid(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


yt/frontends/ramses/data_structures.py
```diff
@@ -582,6 +583,21 @@ def read_rhs(f, cast):
...
+
+    def read_namelist(self):
+        """Read the namelist.txt file in the output folder, if present"""
+        namelist_file = os.path.join(self.root_folder, 'namelist.txt')
+        if os.path.exists(namelist_file):
+            try:
+                with open(namelist_file, 'r') as f:
+                    nml = f90nml.read(f)
+            except ImportError as e:
+                nml = "An error occurred when reading the namelist: %s" % str(e)
```
matthewturk: If we make this an exception of some type, we might be more easily able to determine if it's the success or not, with an `isinstance`.

cphyc: I am not sure I understand your comment.

However, the reason I settled for this is that reading `namelist.nml` is not required to properly load the simulation, so I was trying to avoid throwing errors at the user.

matthewturk: NBD, just thought if we had an object we could check if it read correctly without doing a string comparison.  I agree about not throwing errors!

cphyc: I am still confused ; where do you see a string comparison? Here I am storing the content of the namelist in `ds.parameters['namelist']`. If we're missing the `f90nml` package, we store the error message instead.

That is probably not the best solution though.

matthewturk: Sorry, I meant -- let's pretend I wanted to know if the namelist had been parsed.  I could do that with an `isinstance` or I could check the contents of the string, right?  Or is it only going to be an instance of `str` if it succeeds?

cphyc: There are three cases:

1. the file does not exist and `ds.parameters['namelist']` does not exist,
2. the file exists and `f90nml` is missing: `ds.parameters['namelist']` then contains the error message as a string,
3. the file exists and `f90nml` is installed: `ds.parameters['namelist']` is a dict-like structure with the content of the namelist.

Any idea to make it better is welcome!

matthewturk:
Nah, this looks great to me!

yt/frontends/ramses/data_structures.py
cphyc: So right now if I wanted to see if the namelist was parsed, I wouldn't have a proper way of doing it except checking `isinstance(ds['parameters'], str)`.

munkm:
@Xarthisius what do we need to do to get `f90nml` on Jenkins?

Xarthisius:
> The tests fail because the new test assumes `f90nml` to be installed (which is not the case). Would it be possible to install it on jenkins?

Since it's a python package, all you need to do is add it to `yt/tests/test_requirements.txt`

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -437,4 +437,16 @@ def test_ramses_empty_record():
     ds.index
 
     # Access some field
-    ds.r[('gas', 'density')]
\ No newline at end of file
+    ds.r[('gas', 'density')]
+
+output_00080 = "output_00080/info_00080.txt"
+@requires_ds(output_00080)
```
munkm: Because this test uses the f90nml module I think you need the `@requires_module()` decorator.

munkm: Ha nice mind-meld with @Xarthisius! 😎

cphyc:
@matthewturk I think we should rather catch the error and display it as a warning to the user. There used to be a bug in RAMSES that would write the configuration file in binary format. I'm working on it.

cphyc:
@yt-fido test this please.

cphyc:
The tests seem to fail for unrelated reasons, don't they?

munkm:
I'm working on a fix for the 3.6 answer tests on Travis.

munkm:
@cphyc if you merge or rebase with master you should get the fix I pushed for the Travis tests.

Xarthisius:
@yt-fido test this please

PR #2348, opened 2019-10-22
[bugfix] WeightedVariance is broken in a couple of ways in yt-4.0

jzuhone:
In the `yt-4.0` branch, the `WeightedVariance` derived quantity (and thus the `std` method of data containers which is downstream from it) has a couple of bugs. 

1. If one runs into a case where some chunks have a weight which sums to zero, the code returns floating-point zeros to be used in the final variance calculation. However, the chunks with data are not dimensionless in general, and so when one attempts to add everything together you get a `IterableUnitCoercionError`. This PR fixes that by dropping the units during the calculating and assigning them at the end. 

2. I also noticed after fixing 1. that the answers given were not the same as if one had simply computed the weighted variance by hand using NumPy on the data object's field. I tracked this second error down to commit https://github.com/yt-project/yt/commit/191056c8eec45debd777643ff5cec2696ad70dc0.

Which was just a mix-up in the reformatting of the line. 

I'll see what I can do about improving test coverage for this.

### DIFF
--- a/yt/data_objects/derived_quantities.py
+++ b/yt/data_objects/derived_quantities.py
@@ -381,18 +381,21 @@ class WeightedVariance(DerivedQuantity):
 
     def __call__(self, fields, weight):
         fields = ensure_list(fields)
+        units = [self.data_source.ds._get_field_info(field).units 
+                 for field in fields]
         rv = super(WeightedVariance, self).__call__(fields, weight)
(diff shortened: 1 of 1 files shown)


munkm:
Ok @jzuhone the fix for the Travis tests should be on the 4.0 branch now. If you merge or rebase those tests should pass. 🙂

jzuhone:
I'm satisfied with the test coverage here, actually. This is ok to merge whenever.

PR #2351, opened 2019-10-28
fix incorrect usage of is comparison with == comparison

ngoldbaum:
The usage of "is" comparison here was breaking the generation of the `_cartesian_x` field for polar geometries. The fix is to use == comparison.

Ping @neutrinoceros


matthewturk:
LGTM, once tests pass should be merged.

Issue #2353, opened 2019-10-30
Plot axes labels should be in code units, not CGS, when the dataset uses a code unit system


ngoldbaum:
## Bug report

I'm submitting this on behalf of @neutrinoceros who for some reason is unable to submit an issue via the GitHub web UI.

**Bug summary**
`yt.load(<datafile>, unit_system="code")` produces inconsistent unit labelling between the field and the plot axes, e.g. a density field is labelled in "code mass/code length^3" but the axes are labelled in "cm" instead of "code length"

**Code for reproduction**
I'm using the public dataset `KelvinHelmholtz` from Pluto. I selected it because it's lightweight, but the behaviour I'm reporting is general to all frontends I've tested (including Enzo with dataset EnzoKelvinHelmholtz).

```python
dp = yt.load("data.0004.hdf5", unit_system="code")
# flipping the axes for readability since this dataset's domain is rectangular
ds.coordinates.x_axis[2] = 1
ds.coordinates.x_axis['z'] = 1
ds.coordinates.y_axis[2] = 0
ds.coordinates.y_axis['z'] = 0

p1 = yt.plot_2d(ds, "density")
p1.show()
```

**Actual outcome**
![data 0004 hdf5_Slice_z_density raw](https://user-images.githubusercontent.com/14075922/67873928-8503ca00-fb34-11e9-8947-479bda8b4b81.png)

**related**
setting the plot -square- width  without specifying a unit will make the axes labels in "code units" (regardless the unit system used), so there must be everything we need already.
```python
p1.set_width(1, )
p1.show()
```
![data 0004 hdf5_Slice_z_density width_1](https://user-images.githubusercontent.com/14075922/67873931-8634f700-fb34-11e9-8fee-83fc569235ee.png)

munkm:
I double checked with `EnzoKelvinHelmholtz` and this is now resolved in master from the fix in yt-project/yt#2354

neutrinoceros:
thanks @munkm !

munkm:
Thank you @neutrinoceros! Your contributions have been really great!

PR #2354, opened 2019-10-31
Correctly set plot axes units to "code length" for datasets loaded with `unit_system="code"`

neutrinoceros:
## PR Summary

propose a fix for https://github.com/yt-project/yt/issues/2353
the fix is very simple but I'm not 100% confident it will not break anything else.

### DIFF
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -864,8 +864,10 @@ class PWViewerMPL(PlotWindow):
             axis_index = self.data_source.axis
 
             xc, yc = self._setup_origin()
-
-            if self._axes_unit_names is None:
+            if self.ds.unit_system.name.startswith("us"):
+                # this should happen only if the dataset was initialized with argument unit_system="code"
+                (unit_x, unit_y) = ('code_length', 'code_length')
+            elif self._axes_unit_names is None:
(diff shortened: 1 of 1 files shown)


yt/visualization/plot_window.py
```diff
@@ -864,8 +864,10 @@ def _setup_plots(self):
             axis_index = self.data_source.axis
 
             xc, yc = self._setup_origin()
-
-            if self._axes_unit_names is None:
+            if self.ds.unit_system.name.startswith("us"):
```
matthewturk: Why `"us"`?  Is that because it's short for "unit system"?

neutrinoceros: yes. That's the workaround I found. From the example I experimented with, `self.ds.unit_system.name` seems to start either with `"cgs_us"`, `"mks_us"` or just `"us"` , depending on which value is used to set the unit system at load time `yt.load(..., unit_system=<"cgs"|"mks"|"code">)`. If there are more possible cases I should try them out before we merge.

matthewturk: I don't think we need to do that!  I think this PR is probably good to go.

neutrinoceros:
@yt-fido test this please.

PR #2355, opened 2019-10-31
Included instructions on how to install yt on the Intel Distribution …

sacielo:
…for Pyhton, and referenced the relative performance gain.

Desirable features, beyond our scope or abilities:
- introduce a switch in the all-in-one install script to use intel python;
- we would like to reference this also in the home page; the default links
  provided there don't offer this possibility to new users.

## PR Summary

As from prior communication on yt-users, we measured a very significant performance and scalability boost when using Intel python (via conda) on Intel architectures for parallel computation. This possibility was not presented in the documentation, so many users may miss it.
We changed the documentation only.

## PR Checklist


matthewturk:
Thank you!

PR #2357, opened 2019-11-13
[AMRVAC] ensure consistent renormalisation of amrvac datasets

n-claes:
## PR Summary
The AMRVAC code (amrvac.org) uses specific normalisations based on a helium abundance.
When specifying normalisations, amrvac has only three degrees of freedom: a unit length and unit number density should always be specified, together with either a unit temperature or unit length. These normalisations are however _never_ saved in the dataset and should hence by supplied by the user through `ds = yt.load(amrvacdataset, units_override={})`.

This PR extends the `_set_code_unit_attributes()` method in the newly accepted AMRVAC frontend (#2321) to ensure consistency and correct re-normalisation of code units between AMRVAC and yt. It also provides support for additional yt override possibilities (like time, mass, etc) while maintaining consistency.

## PR Checklist

### DIFF
--- a/doc/source/examining/loading_data.rst
+++ b/doc/source/examining/loading_data.rst
@@ -63,6 +63,19 @@ The user has two ways to control displayed units, through
   units_override = dict(length_unit=(100., 'au'), mass_unit=yt.units.mass_sun)
   ds = yt.load("output0010.dat", units_override=units_override, unit_system="mks")
 
+To ensure consistency with normalisations as used in AMRVAC we only allow
+overriding a maximum of three units. Allowed unit combinations at the moment are
+
+.. code-block:: none
+
+  {numberdensity_unit, temperature_unit, length_unit}
+  {mass_unit, temperature_unit, length_unit}
+  {mass_unit, time_unit, length_unit}
+  {numberdensity_unit, velocity_unit, length_unit}
+  {mass_unit, velocity_unit, length_unit}
+
+Appropriate errors are thrown for other combinations.
+
 
 .. rubric:: Partially supported and unsupported features
 
@@ -2385,7 +2398,7 @@ stars might look like this:
 
 For a cosmological simulation, this filter will distinguish between stars and
 dark matter particles.
-        
+
 .. _loading-sph-data:
 
 SPH Particle Data
--- a/yt/frontends/amrvac/data_structures.py
+++ b/yt/frontends/amrvac/data_structures.py
@@ -30,7 +30,7 @@ from yt.funcs import \
 from yt.data_objects.static_output import \
    Dataset
 from yt.utilities.physical_constants import \
-    boltzmann_constant_cgs
+    boltzmann_constant_cgs as kb_cgs
 
 from .fields import AMRVACFieldInfo
 from .datfile_utils import get_header, get_tree_info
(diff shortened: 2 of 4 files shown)


neutrinoceros:
We've been working on this since before the merge off the main chunk of the frontend. It's not clear to me how standard we need to keep the frontend for consistency with the rest of yt, plus @n-claes has a much better understanding of the renomarlisations intrinsic to amrvac. But I'm keeping an eye on this :)

n-claes:
Alright, so I added a couple of tests as well to check the renormalisations when providing overrides and they all pass locally. The py2 and py3 automated tests now fail though, the Travis log shows an assertion error between the fields which is likely due to the new normalisations. Can I force re-generation of the stored answers to those tests somehow?

neutrinoceros:
So if I'm reading those correctly, you're allowing a bunch of scenarios and testing exactly those. I'd suggest adding tests to check that the behaviour is the one you except in forbidden cases:  should we log a warning, or raise an error ? (I'd go with the second)

What's the minimal correct way to tests for compatibility in provided overrides ?
Is it enough to check that the arguments' dimensions form a linearly independent set ?

n-claes:
> So if I'm reading those correctly, you're allowing a bunch of scenarios and testing exactly those. I'd suggest adding tests to check that the behaviour is the one you except in forbidden cases: should we log a warning, or raise an error ? (I'd go with the second)

Just added a couple of tests to check the forbidden cases, I removed the length constraint though because it was not needed. If it's not consistent it raises an error (which it should, imo).

> What's the minimal correct way to tests for compatibility in provided overrides ?
> Is it enough to check that the arguments' dimensions form a linearly independent set ?

Well no, the thing is that there are actually more combinations allowed than currently implemented.
We _can_ add more, but right now 5 different cases are covered already which I think is more than sufficient. The only 'constraint' is that the normalisations are consistently calculated using three different arguments (with either temperature or velocity specified, not both).

neutrinoceros:
Thanks for the tests !
I am just worried we create a huge code base for this thing that will turn into a maintenance nightmare. It would greatly improve the contribution if the consistency check was much simpler, and I believe it could be.

n-claes:
The issue I'm having is that we have to supply re-dimensionalisation through `units_override` (which has a lot of possibilities), such that we _need_ those consistency checks. Especially since a lot of people use the normalisations in amrvac's configuration file during problem setup, so we need them to be consistent.

It would be _much_, _much_ simpler if they were included in the data file itself, but that's not gonna happen any time soon (if at all).

neutrinoceros:
I'm not arguing against having checks, I'm just wondering out loud if there could be a much denser way to express the same logic (and more ?).

n-claes:
Yeah so that's what I meant, I would prefer having this much simpler but as of right now I don't really see an obvious way to condense this into a clean, short block of code, exactly because of the possibilities in `units_override`. Completely agree with you that it's quite an elaborate block now...
The thing is that there are 8 unit normalisations, and depending on the case considered they have to be calculated differently so it gets quite long rather fast.
I'll change this into WIP until I figure out if I can condense it a bit

munkm:
@n-claes if you need to regenerate answers for Jenkins, bump the folder number in the `yt/tests/tests.yaml` file for the amrvac frontend. Then new reference answers will be generated for the frontend. There's some documentation about that here: https://yt-project.org/doc/developing/testing.html#updating-answers

yt/frontends/amrvac/data_structures.py
```diff
@@ -250,15 +251,110 @@ def _set_code_unit_attributes(self):
...
+        setdefaultattr(self, "velocity_unit", amrvac_units_dict['velocity_unit'])
+        setdefaultattr(self, "density_unit", amrvac_units_dict['density_unit'])
+        setdefaultattr(self, "numberdensity_unit", amrvac_units_dict['numberdensity_unit'])
+
+        setdefaultattr(self, "temperature_unit", amrvac_units_dict['temperature_unit'])
+        setdefaultattr(self, "pressure_unit", amrvac_units_dict['pressure_unit'])
+        setdefaultattr(self, "magnetic_unit", amrvac_units_dict['magnetic_unit'])
+
+
+    def _get_amrvac_normalisations(self):
```
neutrinoceros: after carefully reading this whole function, I don't see a good reason not to merge it with `_set_code_unit_attributes`.

yt/frontends/amrvac/data_structures.py
```diff
@@ -250,15 +251,110 @@ def _set_code_unit_attributes(self):
...
+        # 2. calculations for velocity
+        if self.units_override.get('time_unit') is not None:
+            # in this case time was supplied
+            velocity_unit = length_unit / self.time_unit
+        else:
+            # other case: velocity was supplied. Fall back to zero (default) if no overrides supplied
+            velocity_unit = getattr(self, 'velocity_unit', self.quan(0, 'cm*s**-1'))
+
+        # 3. calculations for pressure and temperature
+        if velocity_unit.value != 0:
```
neutrinoceros: I really don't see the point of using `self.quan(0, 'cm*s**-1')` as a default here. `None` would do the job equally well and be much easier to understand :)
(for reference, I attempted 5 times to provide a constructive proposal for these lines in my confusion, now I think I got it)

n-claes: Yeah you're right, I was thinking too much like it was implemented in amrvac. Changed this to None, much easier to read now

yt/frontends/amrvac/data_structures.py
```diff
@@ -250,15 +251,110 @@ def _set_code_unit_attributes(self):
...
+        if not self.units_override:
+            return
+
+        # there are only three degrees of freedom, so explicitly check for this
+        if len(self.units_override) > 3:
+            raise ValueError('Normalisations in amrvac only allow for three degrees of freedom, '
+                             'but {} were specified in units_override'.format(len(self.units_override)))
+
+        # temperature and velocity cannot both be specified
+        if hasattr(self, 'temperature_unit') and hasattr(self, 'velocity_unit'):
```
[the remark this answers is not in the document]
n-claes: Good idea, placed `_check_override_consistency` before the 
`super(AMRVACDataset, self).__init__` call, this now takes `units_override` as an argument. The checks now happen on the provided dictionary itself and before the dataset is loaded. Much cleaner this way

n-claes:
Squashed some commits and rebased to master to resolve previous merge conflict

yt/frontends/amrvac/data_structures.py
```diff
@@ -250,16 +251,101 @@ def _set_code_unit_attributes(self):
...
+        else:
+            # other case: velocity was supplied. Fall back to zero (default) if no overrides supplied
+            velocity_unit = getattr(self, 'velocity_unit', None)
+
+        # 3. calculations for pressure and temperature
+        if velocity_unit is None:
+            # velocity and time not given, see if temperature is given. Fall back to one (default) if not
+            temperature_unit = getattr(self, 'temperature_unit', self.quan(1, 'K'))
+            pressure_unit = ((2.0 + 3.0 * He_abundance) * numberdensity_unit * kb_cgs * temperature_unit).to('dyn*cm**-2')
+            velocity_unit = (np.sqrt(pressure_unit / density_unit)).to('cm*s**-1')
```
neutrinoceros: You might want use the `YTQuantity` method `in_cgs()` instead to `to()` here.

Although I'm actually curious, _why_ are you force-converting to cgs ?

yt/frontends/amrvac/data_structures.py
```diff
@@ -250,16 +251,101 @@ def _set_code_unit_attributes(self):
...
+        else:
+            # velocity is not zero if either time was given OR velocity was given
+            pressure_unit = (density_unit * velocity_unit ** 2).to('dyn*cm**-2')
+            temperature_unit = (pressure_unit / ((2.0 + 3.0 * He_abundance) * numberdensity_unit * kb_cgs)).to('K')
```
neutrinoceros: same as above

yt/frontends/amrvac/data_structures.py
```diff
@@ -250,16 +251,101 @@ def _set_code_unit_attributes(self):
...
+
+        # 4. calculations for magnetic unit and time
+        time_unit = getattr(self, 'time_unit', length_unit / velocity_unit)  # if time given use it, else calculate
+        magnetic_unit = (np.sqrt(4 * np.pi * pressure_unit)).to('gauss')
```
neutrinoceros: same as above.
I'm also wondering how that specific line would work out... since when is "gauss" equivalent to "dyne**1/2" ?

yt/frontends/amrvac/data_structures.py
n-claes: Because apparently yt does not simplify this by default, if I don't specifically ask for the conversion they remain in terms like `length**1/2` or something like that. This ensures that they play nicely with each other :) And I think I used `to_cgs()` before actually, it probably didn't simplify the units as well so that's why I dropped it

yt/frontends/amrvac/data_structures.py
n-claes: Because the magnetic constant  `mu_0` is defined as 4pi in cgs?

neutrinoceros: Wait no, this doesn't make sense. mu0 is not dimensionless, especially if you define it "_in  cgs_", which would be a meaningless statement if you were talking about a truly dimensionless quantity.

yt/frontends/amrvac/data_structures.py
n-claes: Using `in_cgs()` actually worked out nicely, using that now instead

neutrinoceros:
Pretty sure the py2 faillures have nothing to do with the PR
@yt-fido test this please

n-claes:
Should be fixed now, @yt-fido test this please

ngoldbaum:
Does the existing discussion about units in the AMRVAC "loading data" docs need any expansion? https://yt-project.org/docs/dev/examining/loading_data.html#amrvac-data

yt/frontends/amrvac/data_structures.py
```diff
@@ -242,24 +243,106 @@ def _parse_parameter_file(self):
...
+
+        # there are only three degrees of freedom, so explicitly check for this
+        if len(overrides) > 3:
+            raise ValueError('More than 3 degrees of freedom were specified '
+                             'in units_override ({} given)'.format(len(overrides)))
+        # temperature and velocity cannot both be specified
+        if 'temperature_unit' in overrides and 'velocity_unit' in overrides:
+            raise ValueError('Either temperature or velocity is allowed in units_override, not both.')
+
+        allowed_combinations = [{'numberdensity_unit', 'temperature_unit', 'length_unit'},
```
ngoldbaum: nit: I would define at module level and give it an all-uppercase name like `ALLOWED_UNIT_COMBINATIONS` to make it clearer this is constant data.

yt/frontends/amrvac/data_structures.py
```diff
@@ -242,24 +243,106 @@ def _parse_parameter_file(self):
...
+            pressure_unit = ((2.0 + 3.0 * He_abundance) * numberdensity_unit * kb_cgs * temperature_unit).in_cgs()
+            velocity_unit = (np.sqrt(pressure_unit / density_unit)).in_cgs()
+        else:
+            # velocity is not zero if either time was given OR velocity was given
+            pressure_unit = (density_unit * velocity_unit ** 2).in_cgs()
+            temperature_unit = (pressure_unit / ((2.0 + 3.0 * He_abundance) * numberdensity_unit * kb_cgs)).in_cgs()
+
+        # 4. calculations for magnetic unit and time
+        time_unit = getattr(self, 'time_unit', length_unit / velocity_unit)  # if time given use it, else calculate
+        magnetic_unit = (np.sqrt(4 * np.pi * pressure_unit)).to('gauss')
```
ngoldbaum: I haven't thoroughly checked any of the math here, leaving that to @neutrinoceros :stuck_out_tongue:

neutrinoceros: I reviewed this a million times already, and I'm not any kind of authority when it comes to magnetic stuff (in amrvac or anywhere). :)
On this particular line, it's very hard to check for it, but I trust @n-claes that he discussed it with AMRVAC's guru, who happens to be his phs advisor

neutrinoceros: But we *did* have a discussion about the unit conversion here and I'm positive that it's correct.

n-claes: Yep this is fine, `velocity_unit` is defined as `sqrt(pressure_unit/density_unit)` so eventually it results in the same normalisation as used by eg. the enzo frontend

yt/frontends/amrvac/data_structures.py
```diff
@@ -269,6 +274,8 @@ def _set_code_unit_attributes(self):
         else:
             # other case: numberdensity is supplied. Fall back to one (default) if no overrides supplied
             numberdensity_override = self.units_override.get('numberdensity_unit', (1, 'cm**-3'))
+            if 'numberdensity_unit' in self.units_override: # print similar warning as yt when overriding numberdensity
```
neutrinoceros: If you're intentionally copying warning template here, this may make it harder to trace where the warning was emmited in case there's a bug with it or the code that it comes with

n-claes:
Note that I added a small `mylog.info` message as well when overriding numberdensity so it's similar to what yt does when overriding other units.

yt/frontends/amrvac/data_structures.py
n-claes: Ah wow you reviewed before my comment :)
yeah so yt prints info to the console confirming the unit override, except for numberdensity.
I added this manually so it's consistent, otherwise you get an info message for eg. both length and temperature and not for numberdensity, it's a bit confusing

neutrinoceros: So the rest of yt won't ever warn about overriding THIS quantity because it doesn't override it. Fair enough. But my point is that one may grep for `Overriding` or `mylog.info("Overriding`, and end up looking up the code in yt's core instead of here, so you might actually prefer to _break_ this warning's pattern to make this situation even less likely.

neutrinoceros: So the rest of yt won't ever warn about overriding THIS quantity because it doesn't override it. Fair enough. But my point is that one may grep for Overriding or mylog.info("Overriding, and end up looking up the code in yt's core instead of here, so you might actually prefer to break this warning's pattern to make this situation even less likely.

n-claes: That's true, but personally I would prefer consistency over a small chance that grep gives an additional result. I was a bit confused myself as I gave `numberdensity_unit` together with length and temperature as an override, but yt printed that only the latter two were overridden with no info whatsoever on numberdensity. Yt's info messages are of this form 
`mylog.info("Overriding %s_unit: %g %s." ...`
while I explicitly put `numberdensity_unit` in the info message. Numberdensity is specific only to our frontend, yt does not support overriding that by default (but it's needed for amrvac). So actually, if people grep for it they should come out in the amrvac frontend, which is what we want, no?

ngoldbaum:
Hey just responding because I got pinged to re-review this pull request.

However, It looks like my question about expanding the docs never got answered. Does this change warrant expanding the discussion about units in the documentation on loading AMRVAC data? 

It also looks like the other comment I left with a style nit was never responded to.

I'd appreciate it if you could respond to my code review comments before pinging me to review things.

Regarding number density, it may make more sense to add the ability to override that unit in the yt core, which you should feel free to change if it makes sense to change it there, rather than changing things only in this frontend. I don't really strongly care though and don't think it should block merging this PR.

yt/frontends/amrvac/data_structures.py
neutrinoceros: Ok, sold

n-claes:
Sorry, my bad, overlooked the comment regarding the docs. You're right, I just pushed a docs update which includes the new normalisation system for the frontend

yt/frontends/amrvac/data_structures.py
n-claes: Done as suggested!

doc/source/examining/loading_data.rst
```diff
@@ -63,6 +63,19 @@ The user has two ways to control displayed units, through
   units_override = dict(length_unit=(100., 'au'), mass_unit=yt.units.mass_sun)
   ds = yt.load("output0010.dat", units_override=units_override, unit_system="mks")
 
+  To ensure consistency with normalisations as used in AMRVAC we only allow
+  overriding a maximum of three units. Allowed unit combinations at the moment are
```
ngoldbaum: This is at the same indent level as the code above and will be interpreted as a continuation of the above code block. You can see for yourself in the rich view of this file on github.

n-claes: Indeed... Fixed, amended the commit

n-claes:
The text in the docs is fine now but the code block was _still_ not properly indented. Had to amend again, now it's fine (finally)

PR #2358, opened 2019-11-18
Add an instruction to the doc about continuous mesh colormap

shoheiogawa:
## PR Summary

The commit included in this PR adds an explanation to the doc about
why the rendered color is not continuous when rendering two-part
unstructured mesh and how to explicitly set each colormap bound
for a continuous color plot.

## PR Checklist

### DIFF
--- a/doc/source/visualizing/unstructured_mesh_rendering.rst
+++ b/doc/source/visualizing/unstructured_mesh_rendering.rst
@@ -365,6 +365,46 @@ with two meshes on it:
     im = sc.render()
     sc.save()
 
+However, in the rendered image above, we note that the color is discontinuous on
+in the middle and upper part of the cylinder's side. In the original data,
+there are two parts but the value of `diffused` is continuous at the interface.
+This discontinuous color is due to an independent colormap setting for the two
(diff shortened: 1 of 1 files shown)


matthewturk:
Awesome!  Thank you!  I think once the build finishes and it's green this should be merged.

ngoldbaum:
LGTM as well, may just want to check in the doc build that the rendered version has an image inline, since this is using `python-script` it will be run as part of the doc build and the image it generates will be appended inline. Thank you so much for contributing!

shoheiogawa:
Is there any way for me to look at the build data at the CI environments? I was wondering if I can make sure in advance that the result is what I intended for in a non-local environment.

ngoldbaum:
Yup, it's a build artifact from the "docs" jenkins job that ran on your PR. The page you modified is here: https://tests.yt-project.org/job/yt_docs_new/367/artifact/doc/build/html/visualizing/unstructured_mesh_rendering.html

If you think it looks correct I'm happy to hit the merge button.

shoheiogawa:
Thank you. The image is what I was intended.

PR #2362, opened 2019-11-21
[AMRVAC] fix derived unit parsing

neutrinoceros:
Solves issue 2361.
Marked as WIP because I realised that I don't see a good reason to use `setdefaultattr` for the three affected dataset attributes, which are derived units, merely aliasing more fondamental units normalisation factors.

I'll wait for the tests to run once before I rewrite this without `setdefaultattr`.

### DIFF
--- a/yt/frontends/amrvac/data_structures.py
+++ b/yt/frontends/amrvac/data_structures.py
@@ -254,9 +254,10 @@ class AMRVACDataset(Dataset):
         setdefaultattr(self, "mass_unit", self.quan(1, "g"))
         setdefaultattr(self, "time_unit", self.quan(1, "s"))
 
-        setdefaultattr(self, "velocity_unit", self.quan(self.length_unit/self.time_unit, "cm*s**-1"))
-        setdefaultattr(self, "density_unit", self.quan(self.mass_unit/self.length_unit**3, "g*cm**-3"))
-        setdefaultattr(self, "numberdensity_unit", self.quan(self.length_unit**-3, "cm**-3"))
+        # derived units
+        setdefaultattr(self, "velocity_unit", self.quan(self.length_unit/self.time_unit))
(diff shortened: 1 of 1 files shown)


ngoldbaum:
I think you need to use `setdefaultattr`, we use that because the `units_override` handling might be setting these attributes before we get to the place in the frontends where the attributes are set normally without `units_override`.

ngoldbaum:
Although looking at the `units_override` code I guess it will never actually set anything besides length, matt, time, velocity, magnetic, and temperature, so I think you're right that `setdefaultattr` isn't necessary here (although it won't hurt). It may make more sense to make this loop happen over all the entries of `units_override` that are passed in by the user:

https://github.com/yt-project/yt/blob/master/yt/data_objects/static_output.py#L1033-L1043

neutrinoceros:
I'm not sure this is a good idea. I think it's best to keep a limited number of primitive and independent quantities exposed to override. Although it's actually not the case already, because currently we expose velocity as well as length and time...

neutrinoceros:
So if we keep it consistent with the existing loop, you're right, it should run on all user-provided quantities. It would actually be more flexible and simpler code !

ngoldbaum:
It may be the case that `units_override` predates the checking for `density_unit` in the code that sets up the code unit registry, so it was just never added to that for loop.

ngoldbaum:
I don't think the py2 test failure is related to this PR, may be worth re-running the tests though to see if it's repeatable. You can do that by commenting "@yt-fido test this please" in a PR

ngoldbaum:
oh lol I triggered the build with that last comment

neutrinoceros:
So what we discussed may be suited for a completely separate issue, I think. But in the mean time the branch can probably be merge as is. Thoughts, @ngoldbaum ?

PR #2286, opened 2019-07-02
[WIP] Pytest

jcoughlin11:
## PR Summary

This Work in Progress PR is meant to address answer testing in yt. Currently, yt uses nose for its testing. Additionally, the answer tests are quite time and memory intensive, at the moment. This PR is meant to do two things: switch from nose to pytest to simplify the answer testing environment, and to implement hashes in the answer tests to hopefully streamline them.

Also, I know this PR is being issued off of master, and I'm not supposed to do that, but I couldn't get the answer tests to work on yt-4.0 (hopefully this PR addresses this, as well), so my fork's branch was made off of master. I apologize if this is an issue.

## PR Checklist

- [ ] Code passes flake8 checker
- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #2363, opened 2019-11-22
Use sorted glob in athena

matthewturk:
This problem showed up in the pytest PR (#2286).

We were able to verify that using unsorted glob did *not* result in any differences in *derived* products, only in the actual index order of the grids post-reconstruction.

(Also the view sure was nice up on my [high horse](https://github.com/yt-project/yt/pull/2286), but ... :grimacing:)

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -6,7 +6,7 @@ answer_tests:
   local_artio_001:
     - yt/frontends/artio/tests/test_outputs.py
 
-  local_athena_004:
+  local_athena_005:
     - yt/frontends/athena
 
   local_athena_pp_001:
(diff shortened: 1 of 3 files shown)


ngoldbaum:
Ah yes, operating systems not returning file lists in a sorted order. Reminds me of https://arstechnica.com/information-technology/2019/10/chemists-discover-cross-platform-python-scripts-not-so-cross-platform/

matthewturk:
In my original PR description, the markdown link for "high horse" was
supposed to be that very same article, but I pasted the wrong copy/paste
buffer so it ended up just being a link to the pytest PR.

munkm:
I think you'll need to bump the folder number in `tests.yaml` to get the Jenkins stuff to pass.

Issue #2365, opened 2019-11-24
Loading kwargs are ignored when accessing a timeseries's slice


neutrinoceros:
> This may not be specific to the AMRVAC frontend but I did not manage to load a Enzo time series with yt.load + wildcard.
> it would help if anyone could reproduce this with another frontend (or report otherwise) 

**bug summary**
Individual datasets from a timeseries are not loaded properly (kwargs are ignored) specifically when the timeseries is accessed through a slice.

**code for reproduction**
As there is no amrvac time series currently stored on the website, here's one (5 snapshots)

http://use.yt/upload/22d10c10

Now, assuming this archive was untared to a "amrvac_ts" dir

```python
ts = yt.load("amrvac_ts/bw_polar_2D*.dat", units_override={"time_unit": (1, "yr")})
for ds in ts:
    print(ds.time_unit, end=" "*2)
for ds in ts[:]:
    print(ds.time_unit, end=" "*2)
```

**expected output**
```
1.0 yr  1.0 yr  1.0 yr  1.0 yr  1.0 yr  1.0 yr
1.0 yr  1.0 yr  1.0 yr  1.0 yr  1.0 yr  1.0 yr
```

**actual output**
```
1.0 yr  1.0 yr  1.0 yr  1.0 yr  1.0 yr  1.0 yr
1.0 s  1.0 s  1.0 s  1.0 s  1.0 s  1.0 s
```

Note that this is not specific to the `units_override` kwarg, as I actually discovered this bug with `geometry_override`; however, the former is more broadly used across frontends and this doesn't look like something frontend specific (needs confirmation).

neutrinoceros:
note : this is where the timeseries's `__getitem__` method is defined
https://github.com/yt-project/yt/blob/e76f71c23bb274e7fa53bc7a6c40ac8e751bd284/yt/data_objects/time_series.py#L193

neutrinoceros:
bug found. Turns out it had nothing to do with AMRVAC's frontend. See PR https://github.com/yt-project/yt/pull/2366

PR #2366, opened 2019-11-24
fix issue 2365

neutrinoceros:
fixes https://github.com/yt-project/yt/issues/2365
`DatasetSeries.kwargs` were not properly passed down to child instance created by slicing. This commit fixes it.

Ideally this behaviour should be tested. I never added tests for yt's core yet so guidance might be useful.

### DIFF
--- a/yt/data_objects/time_series.py
+++ b/yt/data_objects/time_series.py
@@ -195,7 +195,9 @@ class DatasetSeries(object):
             if isinstance(key.start, float):
                 return self.get_range(key.start, key.stop)
             # This will return a sliced up object!
-            return DatasetSeries(self._pre_outputs[key], self.parallel)
+            return DatasetSeries(self._pre_outputs[key],
+                                 parallel=self.parallel,
+                                 **self.kwargs)
         o = self._pre_outputs[key]
         if isinstance(o, string_types):
             o = self._load(o, **self.kwargs)


ngoldbaum:
It turns out `DatasetSeries` doesn't have explicit tests at the moment. If you'd like you could create a new file in `yt/data_objects/tests` called e.g. `test_time_series.py` and then add tests there. Ideally the test would use fake in-memory data (see the particle trajectories tests for examples with particle datasets, you could do something similar with e.g. `fake_random_ds` and you can grep the codebase for examples of tests that use `fake_random_ds`. You can also use real datasets from yt-project.org/data as long as you decorate the test with `requires_file`, which makes sure the tests only runs if the file is available in `test_data_dir`. Tests that use real data are less useful because they only ever get run on jenkins or on developer machines that happen to have the required data downloaded.

There's also `requires_ds` which is used by the answer testing system but you shouldn't use for these tests because you wouldn't be writing answer tests.

More detail in the docs: https://yt-project.org/docs/dev/developing/testing.html.

yt/data_objects/time_series.py
```diff
@@ -195,7 +195,9 @@ def __getitem__(self, key):
             if isinstance(key.start, float):
                 return self.get_range(key.start, key.stop)
             # This will return a sliced up object!
-            return DatasetSeries(self._pre_outputs[key], self.parallel)
+            return DatasetSeries(self._pre_outputs[key],
+                                 parallel=self.parallel,
+                                 **self.kwargs)
```
ngoldbaum: do we need to do the same thing with `setup_function`?

ngoldbaum:
I put in https://github.com/yt-project/yt/pull/2367 for the appveyor failure

neutrinoceros:
I see. Yet it's unclear to me how to create a fake time series without using `yt.load()`

yt/data_objects/time_series.py
neutrinoceros: I don't know what it's supposed to do, but I don't think so. Apparently everywhere a `self._setup_function` is called upon a given `ds`, it happens right after said `ds` is initialised and passed the kwargs.

ngoldbaum: You can create a time series from a list of datasets, so you can do something like:

```
ts = DatasetSeries([fake_random_ds(16) for i in range(n_steps)])
```

Like this test: https://github.com/yt-project/yt/blob/master/yt/data_objects/tests/test_particle_trajectories.py#L54

neutrinoceros: I don't think it will work here, because the behaviour I want to test is for datasets that are  instantiated by the `DatasetSeries` container, which only happens if `DatasetSeries._pre_outputs` contains strings-likes, rather that existing datasets (see https://github.com/yt-project/yt/blob/e76f71c23bb274e7fa53bc7a6c40ac8e751bd284/yt/data_objects/time_series.py#L200)

This is turning into a separate issue.

ngoldbaum: Ah that makes sense. In that case just write a test with a real time series dataset from yt-project.org/data.

Or don’t 🤷

Tests aren’t necessary for bugfix PRs.

neutrinoceros: I'd rather do it another time, when I feel more confortable with timeseries as a user :)

Issue #2368, opened 2019-11-25
[AMRVAC] Optionally supply parfiles to populate AMRVACDataset.parameters


neutrinoceros:
## feature proposal

Populate `ds.parameters` with a `"namelist"` (working name) key, which value would be a `f90nml.Namelist` object, defaulting to `None`, through the optional load kwarg `parfiles`.

This optional argument should also be stored in `ds.parameters` to keep track of where information came from in a dataset.

**example user code**

```python
# minimal example
ds = yt.load("bw_2d0000.dat", parfiles="amrvac.par")

# a more complex example
parfiles = ["amrvac.par", "add_dust.par", "boundary_conditions.nml"]
ds = yt.load("bw_2d0000.dat", parfiles=parfiles)
```

**benefits**
- allow building of (user *or* default) derived fields using variables that are not found in the datfile header, such as pressure and sound-speed (and Mach number, while we're at it)
- more debugging information available to the user

**Requirements**
- on-demand-import for f90nml (this was already done here https://github.com/yt-project/yt/pull/2347)
- an additional routine to "patch" configurations on top of each other replicating AMRVAC's behaviour (I wrote exactly this in a personal codebase already)
- this needs to stay optional, although automatically discovering existing "amrvac.par" could help in many user cases (open to discussion)

**possible follow-up**
A subsequent feature proposal would be to expose the amrvac-namelist-parsing function so it can be used in a user-script to easily merge parfiles in one line

```python
from yt.frontends.amrvac import read_amrvac_namelist

parfiles = ["amrvac.par", "add_dust.par", "boundary_conditions.nml"]
unified_namelist = read_amrvac_namelist(parfiles)
unified_namelist.write("unique_namelist.par")
```
~I'll need help in decided where to put that function.~ (solved with help from @ngoldbaum)

ping @n-claes

neutrinoceros:
companion PR https://github.com/yt-project/yt/pull/2369

neutrinoceros:
fixed in #2369

PR #2369, opened 2019-11-25
[AMRVAC] Parse AMRVAC's parfiles if user-provided

neutrinoceros:
## PR Summary:

implement proposal https://github.com/yt-project/yt/issues/2368

**checklist**
- [x] implement core function
- [x] add kwarg to `AMRVACDataset`
- [x] store namelist and parfiles in `AMRVACDataset.parameters`

edit: parfiles are actually stored as a hidden parameter for debugging; namely `ds._parfiles`, and namelist is stored as its own separate attribute to avoid dict nesting and make it more clearly exposed.

 _to be discussed_ (hence the "help wanted" label):
- [x] ~maybe add autodiscovery default parfile name `amrvac.par`~ WONTFIX (not worth the potential bugs)
- [x] expose the core function -> **where ?** SOLVED: exposed it at the top level of the frontend (thanks @ngoldbaum)


neutrinoceros:
this py2 failures look a lot like the ones that happened in https://github.com/yt-project/yt/pull/2357

ngoldbaum:
Looks like https://github.com/yt-project/yt/issues/1702 (although that's talking about py3 failures). Regardless it's unrelated to this PR. Sorry about the flaky tests...

neutrinoceros:
Now that the independent py2 matplotlib failures are (hopefully) fixed...

@yt-fido test this please

ngoldbaum:
Can we get a test for this function?

neutrinoceros:
sure, but I'll need to update some example parfiles first.

neutrinoceros:
@ngoldbaum here you go !
I'm not too confident this is the correct way to store the sample parfiles used for the test though.

neutrinoceros:
so apparently every tests are breaking because f90nml is required _and_ not found. I'm going to have a look at how @cphyc did it for RAMSES in https://github.com/yt-project/yt/pull/2347.

neutrinoceros:
So _this_  failure should be fixed now. But I think the tests will still break because of how the input files are placed. I should probably upload them to the website and use a `@requires_file` decorator. Only thing is that I don't know of an example for this: uploading input files for tests that are _not_ datasets. Since the main parfile I'm using for testing is literally the one that's used to generate the "amrvac_blastwave_cartesian_3D" tests, I think I can just add them there.

neutrinoceros:
should be good now, but I'll write a website-PR to actually store the parfiles there.

ngoldbaum:
I don’t think you need to upload the data files you added or use `requires_file`, they’re already available in version control. However you do need to refer to the absolute paths of the data files, possibly constructing that path using `__file__` (see e.g. https://stackoverflow.com/questions/3430372/how-do-i-get-the-full-path-of-the-current-files-directory). You might also need to ensure the data files get included in the yt source distribution by ensuring they get included in it by editing `MANIFEST.in`. You might also need to edit `setup.py` to ensure the data get installed.

Another way to do it that won’t require as much custom handling would be to encode the data files as strings inside of a python file (e.g. the test files).

ngoldbaum:
You can take a look at how we deal with this in our handling of unstructured mesh data:

https://github.com/yt-project/yt/blob/cab1f48f982c28fbd1f8c1075ce3f1a0db5e686c/yt/testing.py#L308

Here the data are imported from a file where the data files are all encoded as strings in a python module.

neutrinoceros:
Yes I thought of those options but I'd rather go with adding the parfiles to the website. It would have been a good idea to attach them to the datasets initially actually, had I though of implementing this at the time.

I will not however do it for every other datasets on the website, so hopefully this is a one-of.
Unless you have strong opinions _against_ the current proposed state of this tests, I'd say I'm happy with it.

neutrinoceros:
also: about their availability in version control, I also intend to remove them by rebasing.

ngoldbaum:
OK fair enough. The reason I don’t like tests that use external data outside version control is they only ever run on jenkins and on developer machines where the data happens to be downloaded and things are configured correctly.

It also requires intervention by someone at NCSA to ensure the data files are downloaded on the jenkins test box.

neutrinoceros:
This is a reasonable point. Actually, the reason I didn't originally got with the `__file__` option is that I thought it was bad practice (but I use it anyway in smaller projects...).
One can also argue against the current state because the "modifier" parfile was written solely for this test and it might be confusing to bundle it with an output file that has virtually nothing to do with it.
bottom line : I'll change it.

neutrinoceros:
Done. So if I got this right, `MANIFEST.in` and `setup.py` may need updating, but is it required for the test to be run ? Unless it is, I'll wait for the test to run smoothly before I finalise this :)

ngoldbaum:
No that's only needed for the tests to pass on an installed copy of yt. You can check for yourself by doing "python setup.py install" in the yt source distribution (after uninstalling and yt installs you already have) and then checking in the installed version to see if the data files are there. You can find the path of the installed location by doing `import yt; print(yt.__file__)`.

neutrinoceros:
I’d rather do that in a separate conda environment but the principle is the same. Thanks for the indications !

neutrinoceros:
(A while ago, I setup a ˋwhych` command on my laptop to do exactly this)

neutrinoceros:
I tried my best to register them properly. It should be noted that I actually did not manage to test wether that last commit was required or not because when I install yt from a fresh clone, the files are present but I don't see how they couldn't be, since they are bundled with the rest of the repo.

ngoldbaum:
> when I install yt from a fresh clone, the files are present

Yes, they're present in the repo, but they may not necessarily be present in the installed version of yt. However if you tested locally and it works you should be good to go.

setup.py
```diff
@@ -220,6 +220,9 @@
                             "yt/geometry/",
                             "yt/utilities/lib/"],
               depends=glob.glob("yt/frontends/artio/artio_headers/*.c")),
+    Extension("yt.frontends.amrvac.tests.parfiles",
+              "yt/frontends/amrvac/test/sample_parfiles/*.par",
+    )
```
ngoldbaum: This isn't correct, you're not building a C extension. Let's chat about this in a lower-latency context.

neutrinoceros: force-pushed so there no trace of this blatant mistake 🙈

ngoldbaum:
Should this be mentioned in the documentation section on loading AMRVAC data? https://yt-project.org/docs/dev/examining/loading_data.html#amrvac-data

neutrinoceros:
Right now it doesn't need documentation because it's doing nothing, except for storing parameters in the dataset. Once this is merge I'll open a new PR _using_ this to automatically derive some additional fields. At that point, I'll also add documentation for the feature, along the lines of "In order to get some derived fields such as pressure and sound speed, you can optionally pass the parfiles to yt.load such and such..."

Issue #2370, opened 2019-11-25
Bug: plot_2d breaks if provided with a spherical dataset


neutrinoceros:
### Bug report

**Bug summary**
No proper handling of the (admittedly rare) case where `plot_2d`'s argument dataset has spherical geometry.

**Code for reproduction**
Using amrvac_blastwave_spherical_2D

```python
ds = yt.load("amrvac/bw_2d0000.dat")
yt.plot_2d(ds, "density")
```

**Actual outcome**
```
E       UnboundLocalError: local variable 'axis' referenced before assignment
```
Indeed `plot_2D` doesn't currently handle this geometry, but it looks like a mistake rather than a limitation, because there's no explicit blocking.

**Expected outcome**
A 2d "Slice" plot with `axis == phi`.

neutrinoceros:
fixed https://github.com/yt-project/yt/pull/2371

PR #2371, opened 2019-11-25
Add support for spherical geometries in plot_2d

neutrinoceros:
Fix issue #2370 

Simply add an if statement to cover the spherical 2D case, and it works like a charm.
See for example :

2d_sphere p jpg_Slice_phi_density

### DIFF
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -2066,6 +2066,8 @@ def plot_2d(ds, fields, center='c', width=None, axes_unit=None,
         axis = "z"
     elif ds.geometry == "cylindrical":
         axis = "theta"
+    elif ds.geometry == "spherical":
+        axis = "phi"
     # Part of the convenience of plot_2d is to eliminate the use of the
     # superfluous coordinate, so we do that also with the center argument
     if not isinstance(center, string_types) and obj_length(center) == 2:
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -2068,6 +2068,8 @@ def plot_2d(ds, fields, center='c', width=None, axes_unit=None,
         axis = "theta"
     elif ds.geometry == "spherical":
         axis = "phi"
+    else:
+        raise NotImplementedError("plot_2d does not yet support datasets with {} geometries".format(ds.geometry))
     # Part of the convenience of plot_2d is to eliminate the use of the
     # superfluous coordinate, so we do that also with the center argument
     if not isinstance(center, string_types) and obj_length(center) == 2:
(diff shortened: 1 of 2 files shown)


yt/visualization/plot_window.py
```diff
@@ -2066,6 +2066,8 @@ def plot_2d(ds, fields, center='c', width=None, axes_unit=None,
         axis = "z"
     elif ds.geometry == "cylindrical":
         axis = "theta"
+    elif ds.geometry == "spherical":
+        axis = "phi"
```
ngoldbaum: While we're at it can you add an `else` block that raises a `NotImplementedError` so future users unsupported geometries get a better error message:

```
else:
    raise NotImplementedError("plot_2d does not yet support datasets with {} geometries".format(ds.geometry)")
```

neutrinoceros: excellent suggestion

neutrinoceros:
I was surprised (and glad) to find out how easy it was ! Sounds easy enough, I'm adding the test

yt/visualization/tests/test_plotwindow.py
```diff
@@ -510,6 +510,7 @@ def test_set_unit():
     assert str(slc.frb['gas', 'temperature'].units) == 'keV'
 
 WD = "WDMerger_hdf5_chk_1000/WDMerger_hdf5_chk_1000.hdf5"
+blast_wave = "amrvac/bw_2d0000.dat"
 
 @requires_ds(WD)
```
ngoldbaum: You need to decorate this with @requires_file(blast_wave). I would also change this other `requires_ds` above to `requires_file`, it's incorrect that it's using `requires_ds` right now.

This decorator makes it so the test only runs if the file is available.

neutrinoceros: right, my bad

neutrinoceros: what's the correct way to stack those decorators ?

ngoldbaum: Just put one after the other:

```
@requires_file(WD)
@requires_file(blast_wave)
def test_plot_2d():
```

ngoldbaum: You'll of course also need to import `requires_file` from `yt.testing`.

neutrinoceros: done ! (rebased + force pushed)

neutrinoceros:
I don't think the test was run actually. In any case I can't find a trace for in in the logs.
see https://tests.yt-project.org/job/yt_py3_git/1213/testReport/yt.visualization.tests.test_plotwindow/

ngoldbaum:
Looks like it ran: https://tests.yt-project.org/job/yt_py3_git/1213/testReport/yt.visualization.tests/test_plotwindow/test_plot_2d/

neutrinoceros:
oh, good. Then can I push the button ? 👀

Issue #2375, opened 2019-11-26
[AMRVAC] bug: "e" field is assigned wrong name and unit


neutrinoceros:
# Bug report
In AMRVAC, the on-disk field `"e"` is currently assigned the name `"energy"` and corresponding unit `mass_unit * length_unit**2 / time_unit_**2`, which is incorrect.

This field is in fact defined as the "energy density" in AMRVAC's documentation, though it remains ambiguous because the expression could be loosely used to mean "specific energy", which is not the same thing.

The fix is easy, but it needs to be discussed with AMRVAC's experts first.

neutrinoceros:
related although independent https://github.com/yt-project/yt/issues/1301. I'm surprised this issue is still open almost 3yrs later, has this seen any evolution ? (ping @jzuhone)

n-claes:
The on-disk field `e` in amrvac datasets denotes the _total energy density_ defined as
`rho * v**2`, and hence has units `g / (cm * s**2)` in cgs (since `rho` has units `g / cm**3`)

PR #2376, opened 2019-11-26
[AMRVAC] fix energy field name and unit

neutrinoceros:
will fix https://github.com/yt-project/yt/issues/2375 (missing a clear definition for the unit atm)

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -1,6 +1,6 @@
 answer_tests:
 
-  local_amrvac_002:
+  local_amrvac_003:
     - yt/frontends/amrvac/tests/test_outputs.py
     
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Turns out, changing the name of the "e" field breaks the answer tests, as it should, actually.
So we'll need to regenerate answers. Problem: @n-claes just did it for this -pending- PR #2357.

So I'm guessing this branch should be rebased onto master once #2357 is merge, and the answers regenerated after that. Thoughts, @munkm ?

neutrinoceros:
rebased the branch now that #2357 is merged. I'll wait for the results to complete and probably update the tests answers.

PR #2377, opened 2019-11-26
Update magnetic field handling for RAMSES

RicardaBeckmann:
## PR Summary
This pull request ensures that magnetic fields in RAMSES follow both yt naming convention, and are loaded with the correct units. In this way, magnetic fields from RAMSES outputs can be handled consistently by yt, and derived magnetic quantities such as the plasma beta and Alfven speed become available to users. 

Before this fix, magnetic fields in RAMSES were loaded into yt as dimensionless extra fluid fields which severely limited their use for analysis, and made derived magnetic quantities unavailable. 

This pull request also adds the magnetic divergence per cell, divB, as an extra magnetic fields, which is a commonly used diagnostic to monitor that magnetic simulations are running as expected. 

## PR Checklist

### DIFF
--- a/yt/frontends/ramses/fields.py
+++ b/yt/frontends/ramses/fields.py
@@ -217,15 +217,15 @@ class RAMSESFieldInfo(FieldInfoContainer):
 
         def _divB(field,data):
             '''Calculate magnetic field divergence'''
-            out=data['zeros']
+            out = np.zeros_like(data['magnetic_field_x_right'])
             for ax in data.ds.coordinates.axis_order:
-                out=out+data['magnetic_field_%s_right' % ax]-data['magnetic_field_%s_left' % ax]
-            return out
+                out += data['magnetic_field_%s_right' % ax] - data['magnetic_field_%s_left' % ax] 
+            return out/data['dx']
 
         self.add_field(('gas','divB'),
                        sampling_type='cell',
                        function=_divB,
-                       units=self.ds.unit_system['magnetic_field_cgs'])
+                       units=self.ds.unit_system['magnetic_field_cgs']/ self.ds.unit_system['length'])
 
     def create_rt_fields(self):
         self.ds.fluid_types += ('rt', )
--- a/yt/frontends/ramses/fields.py
+++ b/yt/frontends/ramses/fields.py
@@ -203,6 +203,7 @@ class RAMSESFieldInfo(FieldInfoContainer):
             self.create_magnetic_fields()
 
     def create_magnetic_fields(self):
+        #Calculate cell-centred magnetic fields from face-centred
         def mag_field(ax):
             def _mag_field(field, data):
                 return (data['magnetic_field_%s_left' % ax]+data['magnetic_field_%s_right' % ax])/2
--- a/yt/frontends/ramses/fields.py
+++ b/yt/frontends/ramses/fields.py
@@ -38,6 +38,9 @@ from yt.utilities.physical_constants import \
     mp
 from yt.funcs import mylog
 
+from yt.fields.magnetic_field import \
+    setup_magnetic_field_aliases
+
 b_units = "code_magnetic"
 ra_units = "code_length / code_time**2"
 rho_units = "code_density"
@@ -105,7 +108,13 @@ class RAMSESFieldInfo(FieldInfoContainer):
         ("x-acceleration", (ra_units, ['acceleration_x'], None)),
         ("y-acceleration", (ra_units, ['acceleration_y'], None)),
         ("z-acceleration", (ra_units, ['acceleration_z'], None)),
-        ("Potential", (ener_units, ['potential'], None))
+        ("Potential", (ener_units, ['potential'], None)),
+        ("B_x_left", (b_units, ['magnetic_field_x_left'],None)),
+        ("B_x_right", (b_units, ['magnetic_field_x_right'],None)),
+        ("B_y_left", (b_units, ['magnetic_field_y_left'],None)),
+        ("B_y_right", (b_units, ['magnetic_field_y_right'],None)),
+        ("B_z_left", (b_units, ['magnetic_field_z_left'],None)),
+        ("B_z_right", (b_units, ['magnetic_field_z_right'],None)),
     )
     known_particle_fields = (
         ("particle_position_x", ("code_length", [], None)),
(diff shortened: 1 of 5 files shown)


yt/frontends/ramses/fields.py
```diff
@@ -192,6 +198,35 @@ def _temperature(field, data):
...
+            self.add_field(('gas',"magnetic_field_%s" % ax),
+                           sampling_type='cell',
+                           function=mag_field(ax),
+                           units=self.ds.unit_system['magnetic_field_cgs'])
+
+        def _divB(field,data):
+            '''Calculate magnetic field divergence'''
+            out=data['zeros']
+            for ax in data.ds.coordinates.axis_order:
+                out=out+data['magnetic_field_%s_right' % ax]-data['magnetic_field_%s_left' % ax]
```
cphyc: You can use inplace operation to save memory here:
```suggestion
                out += data['magnetic_field_%s_right' % ax] - data['magnetic_field_%s_left' % ax]
```
Also, by reading your code, it feels like it is missing a `1/dx` term, isn't it?

yt/frontends/ramses/fields.py
```diff
@@ -192,6 +198,35 @@ def _temperature(field, data):
...
+            return out
+
+        self.add_field(('gas','divB'),
+                       sampling_type='cell',
+                       function=_divB,
+                       units=self.ds.unit_system['magnetic_field_cgs'])
```
cphyc: If you update the above code to divide by `1/dx`, this should be changed to be 
```suggestion
                       units=self.ds.unit_system['magnetic_field_cgs'] / <unit_length>)
```

yt/frontends/ramses/io.py
```diff
@@ -278,6 +278,10 @@ def _read_fluid_file_descriptor(fname):
         ('pressure', 'Pressure'),
         ('metallicity', 'Metallicity'),
     ]
+
+    #Magnetic field file descriptors
+    magnetic=np.array([['B_{0}_{1}'.format(dim,side) for side in ['left','right']] for dim in ['x','y','z']]).ravel()
```
cphyc: Since you are only using `magnetic` below to find varnames, it may be better to use a simple python list instead. You can build it using
```suggestion
    magnetic = ['B_{0}_{1}'.format(dim,side) for side in ['left','right'] for dim in ['x','y','z']]
```

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -462,3 +462,16 @@ def test_namelist_reading_should_not_fail():
...
+@requires_file(ramses_mhd_128)
+def test_magnetic_field_aliasing():
+    # Test if RAMSES magnetic fields are correctly aliased to yt magnetic fields and if 
+    # derived magnetic quantities are calculated
+    ds = data_dir_load(ramses_mhd_128)
+    
+    assert ('gas','magnetic_field_x') in ds.derived_field_list
+    assert ('gas','magnetic_field_magnitude') in ds.derived_field_list
+    assert ('gas','alfven_speed') in ds.derived_field_list
+    assert ('gas','divB') in ds.derived_field_list
```
cphyc: While you are at it, why don't you try to access the fields to ensure they do not throw an error? Something like this
```python
    for field in <list of fields>:
        assert field in ds.derived_field_list
        ad[field]  # this should not throw an error
```

yt/frontends/ramses/io.py
cphyc: Also if we want to write a code that would be consistent with the `mapping` above, you could rewrite it as follows:
```python
magnetic_mapping = [(key, key) for key in ('B_{0}_{1}'.format(dim,side) for side in ['left','right'] for dim in ['x','y','z'])]
mapping += magnetic_mapping
```
and remove bellow the lines reading
```python
elif varname in magnetic:
    varname = varname
```

yt/frontends/ramses/fields.py
```diff
@@ -192,6 +198,35 @@ def _temperature(field, data):
...
+
+        for ax in self.ds.coordinates.axis_order:
+            self.add_field(('gas',"magnetic_field_%s" % ax),
+                           sampling_type='cell',
+                           function=mag_field(ax),
+                           units=self.ds.unit_system['magnetic_field_cgs'])
+
+        def _divB(field,data):
+            '''Calculate magnetic field divergence'''
+            out=data['zeros']
```
cphyc: With my comments below, I am afraid this may result in `data['zeros']` being updated in place. Maybe 
```suggestion
            out = np.zeros_like(data['magnetic_field_x_right'])
``` 
would be better?

yt/frontends/ramses/io.py
```diff
@@ -278,6 +278,10 @@ def _read_fluid_file_descriptor(fname):
...
+
     # Convert in dictionary
```
cphyc:
```suggestion
     # Convert to dictionary
```
This one's on me!

yt/frontends/ramses/fields.py
```diff
@@ -192,6 +198,35 @@ def _temperature(field, data):
...
+                           units=self.ds.unit_system['magnetic_field_cgs'])
+
+        def _divB(field,data):
+            '''Calculate magnetic field divergence'''
+            out = np.zeros_like(data['magnetic_field_x_right'])
+            for ax in data.ds.coordinates.axis_order:
+                out += data['magnetic_field_%s_right' % ax] - data['magnetic_field_%s_left' % ax] 
+            return out/data['dx']
+
+        self.add_field(('gas','divB'),
```
cphyc: Is this consistent with yt naming convention for magnetic fields? For velocity, it is for example `('gas', 'velocity_divergence')` so I would have expected to see `('gas', 'B_divergence')` here?

RicardaBeckmann: There is already a field called magnetic_field_magnitude, which is auto-generated once all the fields are correctly loaded, so I don't think we need B^2.

I'm agnostic about calling it magnetic_field_divergence or divB. I've always heard it refer to as divB in conversation, but obviously magnetic_field_divergence would be more consistent.

RicardaBeckmann: I opted for consistency and renamed it.

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -462,3 +462,19 @@ def test_namelist_reading_should_not_fail():
...
+def test_magnetic_field_aliasing():
+    # Test if RAMSES magnetic fields are correctly aliased to yt magnetic fields and if 
+    # derived magnetic quantities are calculated
+    ds = data_dir_load(ramses_mhd_128)
+    ad=ds.all_data()
+
+    for field in ['magnetic_field_x',
+                  'magnetic_field_magnitude',
+                  'alfven_speed',
+                  'divB']:
```
cphyc: Following the changes of #ad6d45f, this is likely to fail
 ```suggestion
                  'magnetic_field_divergence']:
```

RicardaBeckmann: Well spotted. I've fixed it.

PR #2378, opened 2019-11-26
Address some cartopy issues.  Fixes #2373.

### LINKED ISSUE #2373: Geographic coordinate handler unit issues
In plots that use the geographic coordinate handler, we should be specifying the units for `extent` as `code_length` (i.e., lat / lon) but it's converting them to `cm`.  This is done for completely understandable reasons (and is the confluence of pushing lat/lon into a mono-unit array) but results in cartopy returning a `MULTISTRING EMPTY` projection for the bounds.

This can likely be handled at the level of the coordinate handler.

This bug isn't actionable yet, but I will update it as I continue digging.

matthewturk:
## PR Summary

This fixes #2373 by not (incorrectly/falsely) converting lat/lon to cm
and then calling `set_extent` on the cartopy axes subclass.

There is still a lingering issue with the zeroing and setting of
extents, but now a plot made with `origin = "native"` and the correct
lat/lon attributes will work.


matthewturk:
@munkm and I have been looking at this, and will try to make another pass at it soon.

matthewturk:
As an update, I can't figure out the weirdness and I could use some help.

matthewturk:
@munkm and I are going to take another look at this next week.

matthewturk:
@yt-fido test this please

matthewturk:
As a quick note, this is one way to pull up a portion of the failing tests:

```
nosetests yt.visualization.tests.test_geo_projections.TestGeoProjections
```

matthewturk:
So I *think* with this set of changes, it is ready for review and to go.

I'm still not sure about how we're doing the `extent` calculations and where/when we set them, but I believe that overriding our origin setting fixes it.

yt/visualization/plot_window.py
```diff
@@ -868,8 +868,12 @@ def _setup_plots(self):
...
             xc, yc = self._setup_origin()
-            if self.ds.unit_system.name.startswith("us"):
-                # this should happen only if the dataset was initialized with argument unit_system="code"
+            if self.ds.unit_system.name.startswith("us") or self.ds.no_cgs_equiv_length:
+                # this should happen only if the dataset was initialized with
+                # argument unit_system="code" or if it's set to have no CGS
+                # equivalent.  This only needs to happen here in the specific
+                # case that we're doing a computationally intense operation
+                # like using cartopy, but it prevents crashes in that case.
                 (unit_x, unit_y) = ('code_length', 'code_length')
```
neutrinoceros, line 877: I think I wrote the original version of this `if` to be able to display "code length" as a unit in x and y labels for 2D plots. Unless I'm missing something here, the change reads to me like this it's now always go to be what's in those axis labels when using certain geometries, including `"spherical"`, and I feel like it's out of scope here.

(In truth I'm almost certain I'm missing something)

yt/visualization/tests/test_geo_projections.py
```diff
@@ -29,7 +29,7 @@ def setup():
 def compare(ds, field, idir, test_prefix, test_name, projection,
             decimals=12, annotate=False):
     def slice_image(filename_prefix):
-        sl = yt.SlicePlot(ds, idir, field)
+        sl = yt.SlicePlot(ds, idir, field, origin = "native")
```
neutrinoceros, line 32: so, am I right in guessing that this change and the likes are meant to force backwards compatibility in these tests results ? (nothing against it, I just want to know if I'm reading this right)

yt/visualization/plot_window.py
matthewturk: Well, so, hm.

It turns out that `unit_x` and `unit_y` get used for setting the values.  The ultimate result of this is that at some point, instead of using the lat/lon, they get converted to some other intermediate unit, and this messes stuff up.

*But*, you should be able to manually change them, at a later time, I believe.

neutrinoceros: As discussed on Slack, I just checked that my suspicions were unfounded. You leave me no choice but to approve the change :-)

munkm:
I have no idea why this is still open? I'll close and we can reopen if there are issues.

PR #2379, opened 2019-11-27
[AMRVAC] add missing tests/__init__.py file

neutrinoceros:
Add an empty `__init__.py` file to amrvac's test folder so it gets properly sorted by Jenkins in the future.


neutrinoceros:
So the tests are now correctly discovered and registered by Jenkins, which was the whole purpose of this.
Screenshot 2019-11-28 at 09 35 20

PR #2381, opened 2019-12-06
[AMRVAC] add derived fields for hydro runs

neutrinoceros:
**note:**
~For now, this branch is based on #2376, I'll rebase it onto master once the former is merged.~ ✅ 

## PR Summary
Add kinetic energy density, thermal pressure, sound speed, and mach number to derived fields.
This is made possible since there's an api for passing configuration "parfiles" at load time, so we can now reconstruct the equation of state that was used in the run.

## PR Checklist
- [x] internal documentation (docstrings, comments)
- [x] finish implementing various related "todos" in the frontend (there should be 2)
- [x] website documentation (add entry to `doc/examining/loading_data.rst`)

I suppose it would suffice to update the answer tests to cover this. However, I am not certain that derived fields are thoroughly tested for. Any clue on this @munkm ?

~This will need immediate extension to the mhd case, but I'd suggest to keep that for a following PR.~
**edit:** actually, the mhd case is the only missing bit to make this PR complete and it does not look that much more work, so I might as well include it here once I fully understand how it works in AMRVAC.

### DIFF
--- a/doc/source/examining/loading_data.rst
+++ b/doc/source/examining/loading_data.rst
@@ -45,6 +45,30 @@ A ``ndim``-long ``periodic`` boolean array was also added to improve
 comptatibility with yt. See http://amrvac.org/md_doc_fileformat.html
 for details.
 
+.. rubric:: Auto-setup for derived fields
+
+Yt will attempt to mimic the way AMRVAC internally defines kinetic energy,
+pressure, and sound speed. To see a complete list of fields that are defined after
+loading, one can simply type
+
+.. code-block:: python
+    print(ds.derived_field_list)
+
+Note that for adiabatic (magneto-)hydrodynamics, i.e. `(m)hd_energy = False` in
+AMRVAC, additional input data is required in order to setup some of these fields.
+This is done by passing the corresponding parfile(s) at load time
+
+.. code-block:: python
+    # example using a single parfile
+    ds = yt.load("output0010.dat", parfiles="amrvac.par")
+
+    # ... or using multiple parfiles
+    ds = yt.load("output0010.dat", parfiles=["amrvac.par", "modifier.par"])
+
+In case more than one parfile is passed, yt will create a single namelist by
+replicating AMRVAC's rules (see "Using multiple par files"
+http://amrvac.org/md_doc_commandline.html).
+
 
 .. rubric:: Unit System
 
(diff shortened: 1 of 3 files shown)


neutrinoceros:
note for later : the if statement in `_thermal_pressure` doesn't need to check for the namelist actually, it suffices to check whether the `e` (energy density) native field is present.

doc/source/examining/loading_data.rst
```diff
@@ -45,6 +45,30 @@ A ``ndim``-long ``periodic`` boolean array was also added to improve
 comptatibility with yt. See http://amrvac.org/md_doc_fileformat.html
 for details.
 
+.. rubric:: Auto-setup for derived fields
+
+Yt will attempt to mimic the way AMRVAC internally defines kinetic energy,
```
matthewturk:
```suggestion
`yt` will attempt to mimic the way AMRVAC internally defines kinetic energy,
```

yt/frontends/amrvac/data_structures.py
```diff
@@ -171,16 +171,33 @@ def __init__(self, filename, dataset_type='amrvac',
...
-                    mylog.error("Inconsistent values in gamma: datfile {}, parfiles {}".format(self.gamma, namelist_gamma))
 
-        else:
-            # todo: add a warning here when having this parameter becomes useful
-            self.namelist = None
+        namelist = None
+        namelist_gamma = None
+        c_adiab = None
+        e_is_internal = None
+        if parfiles is not None:
```
matthewturk: Do you think there is ever a chance that `parfiles` will be an iterable that is not a list/tuple?  i.e., if it's a string, will this cause failure?  If so, you could wrap in `yt.funcs.ensure_list` to be on the safe side.

neutrinoceros: It's actually _meant_ to be a string in 99% of relevant cases, because few people in the AMRVAC community know and use the multiple parfile approach to set up their simulations. In any case, the one time that the `parfile` argument is used is the line right following this extract, namely `namelist = read_amrvac_namelist(parfiles)`, and this is tested against a string path in `tests/test_read_amrvac_namelist.py`.

neutrinoceros:
Thanks @matthewturk for reviewing this. However we're still discussing it internally with @n-claes because results are wrong in the MHD case and we are not yet sure why.

n-claes:
> Thanks @matthewturk for reviewing this. However we're still discussing it internally with @n-claes because results are wrong in the MHD case and we are not yet sure why.

Finally figured this one out, this is due to the fact that amrvac defines the magnetic field in units where `mu0=1` internally and hence calculates magnetic energy density as such. Accounting for the missing mu0 in eae3459 fixed it, tested it for `cgs`, `mks` and `code` unit systems and everything looks good now!

PR #2384, opened 2019-12-11
[AMRVAC] applied numpy docstring style to most of AMRVAC's frontend

neutrinoceros:
Following @munkm's advice, propagate numpy's docstring styling to AMRVAC's frontend.

Add one liner docstrings to internal methods that are subclassed from `Dataset` and were missing documentation.

While I was at it, I changed one or two internal variable names, hopefully making them clearer. I also changed `AMRVACDataset.parse_geometry()` to `AMRVACDataset._parse_geometry()` to make it clear it is not intented to be exposed.


neutrinoceros:
right, I forgot about this file while I was doing this, I can do a follow up commit in this PR

PR #2386, opened 2019-12-12
Fix python 3.8 warnings

cphyc:
In Python 3.8, the usage of `is` in some context raises a
warning. This is due to the fact that `is` checks for equality of
object, while `==` checks for equality of their value (as is expected
in the code).


ngoldbaum:
Thank you, all of these clearly should be `==`. Now that h5py wheels are up it should be pretty easy to add CI for 3.8 if anyone is up for that.

PR #2387, opened 2019-12-12
[AMRVAC] Add auto-discovery of dust fields and set some corresponding derived fields

neutrinoceros:
## PR Summary

Started by refactoring how velocity aliases are setup for gas field and added dust fields
- `dust<idust>_density`
- `dust<idust>_moment<idir>`
- `dust<idust>_velocity<idir>` as well as alias to `dust<idust>_velocity<axname>`
where `<idust>` are dust indexes, `<idir>` are AMRVAC direction indexes and `<axname>` are corresponding direction names according geometry used.

Additionally, added `dust_total_density` and `dust_to_gas_ratio`.

## PR Checklist

- [x] Test added

## to be discussed (or left for future improvement)
I could not figure out yet how to add an infinite amount of known_fields (AMRVAC allows for arbitrary amounts of "dust" fluids) so I used an arbitrary limit at 100, which should be more than enough for most applications. An error is logged in case this limit is ever exceeded.


neutrinoceros:
Seems like the tuple unpacking magic I'm using is a SyntaxError in python 2...
Will https://ytep.readthedocs.io/en/master/YTEPs/YTEP-0033.html be enforced soon ? :-p

munkm:
ytep 0033 will be enforced as soon as we move to yt-4.0! So yes, I'd say it'll be soon! 😎

neutrinoceros:
I was thinking I could conditionally add `total_dust_density` and `dust_to_gas_ratio` but actually it's not causing any issue to _always_ add them, even when there is no dust at all : they simply return null arrays, which makes sense, physically.

neutrinoceros:
@munkm great ! Is there a common way to mark up stuff that should be rewritten in "pure" python 3 once python 2 is dropped ?

neutrinoceros:
adding the WIP flag waiting for test data to be added to the websit (https://github.com/yt-project/website/pull/76)

neutrinoceros:
note: I'm seeing some "true_divide" warnings in my test (currently only run locally), when retrieving dust velocities. The reason is that the density for dust fluid can (and will) go to exactly zero, so when the velocities are computed (v = m/rho), there's a zero division. This apparently does not affect the result visually but we get some NaNs in the resulting velocity data. I don't know whether this behaviour is satisfactory or not. If it's not. If it isn't, then I need to find a way to avoid the division where the density goes to zero.

yt/frontends/amrvac/fields.py
```diff
@@ -29,14 +29,43 @@
...
+    idust : int, optional
+        the dust species index (AMRVAC). idust >= 1
+    """
+    if idust is None:
+        dust_label = ""
+    else:
+        dust_label = "dust%d_" % idust
+    def velocity_idir(field, data):
+        return data["gas", "%smoment_%d" % (dust_label, idir)] / data["gas", "%sdensity" % dust_label]
+    return velocity_idir
```
n-claes: I'm having a hard time actually 'reading' this (never used dust before in amrvac), can you put some notes here what it should be when dust is present? I'm reading the fields now as `('gas', 'dust1_moment_1')`, where it's possible to have this up to `dust100_moment_1`, correct?

yt/frontends/amrvac/fields.py
```diff
@@ -29,14 +29,43 @@
...
+code_pressure = "code_mass / code_length / code_time**2"
+
+# for now, define a finite family of dust fields (up to 100 species, should be enough)
+MAXN_DUST_SPECIES = 100
+known_dust_fields = [("rhod%d" % idust, (code_density, ["dust%d_density" % idust], None))
+                     for idust in range(1, MAXN_DUST_SPECIES+1)]
+for idir in range(1, 4):
+    known_dust_fields += [("m%dd%d" % (idir, idust), (code_moment, ["dust%d_moment_%d" % (idust, idir)], None))
+                          for idust in range(1, MAXN_DUST_SPECIES+1)]
+
```
n-claes: This will be always executed regardless if dust is present in the simulation or not, isn't it better to do this _only_ when there is dust?

yt/frontends/amrvac/fields.py
```diff
@@ -45,33 +74,65 @@ class AMRVACFieldInfo(FieldInfoContainer):
...
-                            dimensions=dimensions.velocity,
-                            sampling_type="cell")
-            self.alias(("gas", "velocity_%s" % idir), ("gas", "velocity_%s" % alias),
+            self.add_field(("gas", "%svelocity_%s" % (dust_label, alias)), function=_velocity(idir, idust),
+                            units=us['velocity'], dimensions=dimensions.velocity, sampling_type="cell")
+            self.alias(("gas", "%svelocity_%d" % (dust_label, idir)), ("gas", "%svelocity_%s" % (dust_label, alias)),
                         units=us["velocity"])
-            self.alias(("gas", "moment_%s" % alias), ("gas", "moment_%s" % idir),
+            self.alias(("gas", "%smoment_%s" % (dust_label, alias)), ("gas", "%smoment_%d" % (dust_label, idir)),
                         units=us["density"]*us["velocity"])
```
n-claes: At this point we're really starting to intertwine dust fields and regular fields, making it very hard to distinguish between the two. I see why you're doing it this way (it's nice and short), but I'm wondering if isn't more convenient to calculate dust fields in a completely separate method (`_setup_dust_fluid_fields()` or something like that), and call it _only_ when there is dust present? It will make it easier for maintaining and/or extending the fields specifically for each case (dust or no dust, or others in the future), no?

yt/frontends/amrvac/fields.py
```diff
@@ -45,33 +74,65 @@ class AMRVACFieldInfo(FieldInfoContainer):
...
+        self.add_field(("gas", "total_dust_density"), function=_total_dust_density,
+                       dimensions=dimensions.density,
+                       units=us["density"], sampling_type="cell")
+
+        def dust_to_gas_ratio(field, data):
+            return data["total_dust_density"] / data["density"]
+
+        self.add_field(("gas", "dust_to_gas_ratio"), function=dust_to_gas_ratio,
+                        dimensions=dimensions.dimensionless,
+                        sampling_type="cell")
```
n-claes: Maybe put a check here so these are only added when there is dust present?

yt/frontends/amrvac/fields.py
neutrinoceros: No, it's fine. `known_fields` is only a whitelist of fields we expect to find on disk with some info regarding the units it should be read in and how to display it and such...
`know_dust_fields` is just a intermediary variable I use to populate the `known_fields` list, which already contains fields that are not in every datfiles (m3, b1...)

yt/frontends/amrvac/fields.py
neutrinoceros: Actually I realised that in the process but decided against making it a seperate case because I don't think it hurts to have  "total_dust_density" or "dust_to_gas_ratio" recipes even is cases where both are 0 (no dust). Even though, this would create a de facto inconsistency with how we treat e.g. magnetic energy... I'll think I'll change it then.

neutrinoceros: Actually I realised that in the process but decided against making it a seperate case because I don't think it hurts to have "total_dust_density" or "dust_to_gas_ratio" recipes even is cases where both are 0 (no dust). Even though, this would create a de facto inconsistency with how we treat e.g. magnetic energy... I'll think I'll change it then.

neutrinoceros: In any case I'm against duplicated code when I know how to avoid it (most of the time), so I don't think it's a good strategy to seperate them more than they are.

neutrinoceros:
Thanks @n-claes 
There's one thing still bugging me with this : we get some "invalid divid" warnings from numpy as some derived field happen to divide by 0 (where the dust density is zero I think), so I'll try to fix this before we can ship it.

neutrinoceros:
Should be solved now. This is ready to go !

yt/frontends/amrvac/fields.py
```diff
@@ -29,14 +30,45 @@
...
+    """Velocity = linear momentum / density"""
+    # This is meant to be used with functools.partial to produce
+    # functions with only 2 arguments (field, data)
+    # idir : int
+    #    the direction index (1, 2 or 3)
+    # prefix : str
+    #    used to generalize to dust fields
+    if prefix is None:
+        prefix = ""
+    moment = data["gas", "%smoment_%d" % (prefix, idir)]
```
matthewturk: I am genuinely pleased to see index-by-one for moment fields, although I don't know why.  Index-by-zero for axes always bugs me.

yt/frontends/amrvac/fields.py
```diff
@@ -29,14 +30,45 @@
...
+    rho = data["gas", "%sdensity" % prefix]
+
+    idx1 = np.where(rho == 0)[0]
```
matthewturk: I think that here we could use the standard, rather than fancy, indexing.  Something like:

`idx1 = (rho==0)`

and then similarly for `idx2`.  You can get total with `.sum()` on the bool arrays, and you can then avoid doing the `in1d` by doing `np.all(idx1 == idx2)` by itself to check the True/False status.

yt/frontends/amrvac/fields.py
```diff
@@ -29,14 +30,45 @@
...
+
+code_density = "code_mass / code_length**3"
+code_moment = "code_mass / code_length**2 / code_time"
+code_pressure = "code_mass / code_length / code_time**2"
+
+# for now, define a finite family of dust fields (up to 100 species, should be enough)
+MAXN_DUST_SPECIES = 100
+known_dust_fields = [("rhod%d" % idust, (code_density, ["dust%d_density" % idust], None))
+                     for idust in range(1, MAXN_DUST_SPECIES+1)]
+for idir in range(1, 4):
```
matthewturk: For this particular case, you could do `for idir in (1, 2, 3)` just for clarity, but not a big deal.

neutrinoceros: I agree, this would be clearer.

yt/frontends/amrvac/fields.py
neutrinoceros: haha, it seems I can never find a legitimate use case for `np.where`. Every time I try to use it, it turns out there's a better way. Thanks !
Otherwise, I'm confused; why should I need the total of those boolean arrays ? then it's not as straight forward as the equality check you're suggesting because there can be 0s in `moment` where `density` is strictly positive (that would be a perfectly still patch of "fluid").

matthewturk: Ah -- I was referring to the check of `idx1.size > 0`.  That also could be changed to `idx1.any()`.

neutrinoceros: I like it, except I don't see how I can avoid using np.in1d + np.where in the following, though it's not a critical check that was mostly useful for me to get a sense of what I was doing. Now I guess I could simply remove it, would you agree ?

matthewturk: What I *think* you can do is something like:

```
if idx1.any():
    mylog.info(...)
    idx2 = (moment==0)
    if not np.all(idx1 == idx2):
        raise RuntimeError
```

Unless I'm misunderstanding, and there's a possibility that `idx2` will have more 0 values than `idx1`.  If that's the case, you could do something like:

```
((idx1 & idx2) == idx1).all()
```

I think.

neutrinoceros: Took me a few minutes to get my head around this but you are absolutely correct, this is exactly what's needed here. So now I should have covered all your suggestions :)

matthewturk:
I think that once the tests pass we should merge this.

PR #2388, opened 2019-12-12
chombo frontend :: solved bug related to slice the ghost cells out

cjoana:
## PR Summary

In the CHOMBO frontend, I encounter a bug when reading datasets containing ghost cells.
The issue is when doing the slice, a minus sign should be a plus sign, in order to filter out only the "right-hand side" ghost cells.  

I tried some of my datasets (from GRChombo) and now the code works well. 

Please check that this (small) change make sense. 

## PR Checklist

### DIFF
--- a/yt/frontends/chombo/io.py
+++ b/yt/frontends/chombo/io.py
@@ -111,7 +111,7 @@ class IOHandlerChomboHDF5(BaseIOHandler):
         data = lev[self._data_string][start:stop]
         data_no_ghost = data.reshape(shape, order='F')
         ghost_slice = tuple(
-            [slice(g, d-g, None) for g, d in zip(self.ghost, dims)])
+            [slice(g, d+g, None) for g, d in zip(self.ghost, dims)])
         ghost_slice = ghost_slice[0:self.dim]
         return data_no_ghost[ghost_slice]
 


ngoldbaum:
If anyone is curious about the code archeology, this comes back to: https://github.com/yt-project/yt/commit/227bc4dc779dc9e6d6a027d27c260f98b45cb3c9

Ping @atmyers.

I think this is pretty clearly correct.

It would be really nice to get a chombo dataset with ghost zones for yt-project.org/data. If you have one that’s ok to share publicly and let us host I can let you know how to add a dataset to that page. Not necessary to merge this PR.

ngoldbaum:
Wait actually I'm confused. Isn't this code supposed to select only active zones, ignoring ghost zones? With this change you're now selecting half the ghost zones. That seems strange to me. Can you go into a little more detail about why you want to only select some of the ghost zones here?

cjoana:
The code, as you said, it's supposed to only get the ghost cells. That's why slices using  `slice(init, end, step)`  which initiates after the (left-hand-side) ghosts and ends after the ghost-cells + the active dimension (ignoring the rhs ghosts-cells).   

About the dataset, I am contacting with my collaborators to try to get one that is simple enough and not too heavy from our GRChombo simulations.

ngoldbaum:
That makes a lot of sense, thank you for the explanation! I think given that explanation and that it fixes a bug for you I'm going to go ahead and merge this, happy to chat on slack about getting the dataset uploaded if/when you get permission. Thank you so much for the contribution!

Issue #2383, opened 2019-12-08
Slow data retrieval for some YTRay objects


zhafen:
### Bug report

**Bug summary**

yt appears to be very slow at reading data from h5py files for use in `YTRay` objects, especially for particle-based simulations with snapshot size >~ 7 GB.

**Code for reproduction**

[A notebook for reproducing the issue is available here, including ewah file, results of cProfile, and data download links.](https://github.com/yt-project/yt/files/3936681/ray_profiling.zip)

**Actual outcome**
Ray generation is ~30 s per field, for a total of ~10 min when enough fields are retrieved.

**Expected outcome**
Using h5py to open the full data directly and perform simple calculations takes <5 seconds, so at minimum a factor of ~6 faster. If caching is possible then the speed-up could be much greater.

**Notes**
* Most of the time is spent in `yt/utilities/io_handler.py:158(_read_particle_selection)`.
* The simulation capable of creating this problem has high dynamic range: Particle mass resolution ~7000 Msun, for a total gas mass of a few times 1e12 Msun (i.e. out to a few times the virial radius of a L* galaxy).
* The raw data is a Lagrangian cosmological zoom-in simulation with data separated into four files. Total file size ~8GB.
* Intersecting the box edge doesn't significantly change computation time. (Note that this is true for the YT ray, but it will take twice as long for the Trident ray because Trident will break up rays that go beyond the box into smaller rays, each of which is equally expensive.)

**Version Information**
  * Operating System: Linux
  * Python Version: 3.7.1
  * yt version: 4.0.dev0 changeset 947ebdc22c48
  * Other Libraries (if applicable): h5py 2.8.0, Trident 1.3.dev1

**Installation Method**
```bash
$ # install yt
$ git clone https://github.com/yt-project/yt --branch=yt-4.0
$ cd yt
$ pip install -e .
$ cd ..
$ # install trident
$ git clone https://github.com/trident-project/trident
$ cd trident
$ pip install -e .
```

zhafen:
@matthewturk , got a chance to open the issue, continuing from our conversation on Slack. Thanks for your help so far!

matthewturk:
[...]

PR #2391, opened 2019-12-13
Switch logic for using all-data in particle IO

matthewturk:
After examining #2383, I came to believe that this logic was inverted.
In short, I believe that if `is_all_data` is `True`, we should be using
all of the data without validating.  If it is `False`, we should check
for collisions.

### DIFF
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -218,12 +218,12 @@ class ParticleIndex(Index):
             else:
                 # TODO: only return files
                 if getattr(dobj.selector, 'is_all_data', False):
+                    nfiles = self.regions.nfiles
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Oops, my fault. Thanks for the fix!

It looks like this pointing at master instead of `yt-4.0`.

matthewturk:
Whoops, you're right!  I just  changed the base and I think it's correctly pointed now.

matthewturk:
@yt-fido please test this

ngoldbaum:
I don’t think those can be cleared without udjusting the jenkins webhook since jenkins isn’t configured to run on this branch. Regardless this should be good to merge.

brittonsmith:
@matthewturk assures me this is ok to merge, so I'm doing it.

PR #2392, opened 2019-12-14
[WIP] Load from path objects

neutrinoceros:
## PR Summary

Allow yt to load from path objects. The fix was extremely simple for python 3.7, but I expect I'll need to change the syntax and maybe the import statement for compatibility with python 2.

**Important note :**
From Python 3.6, `os.path.expandusers()` accepts paths as input and still returns strings, so `yt.load()` sanitizes the input really well as is.
https://docs.python.org/3.7/library/os.path.html#os.path.expanduser

I don't know what's the current oldest python 3 supported version, but I expect that it will be compatible with all Python versions once I make it work for 2.7

In particular, I checked that it still works with timeseries from wildcards, e.g.
```python
from pathlib import Path
ts1 = yt.load("output*.dat")
ts2 = yt.load(Path("output*.dat"))
```

### DIFF
--- a/yt/convenience.py
+++ b/yt/convenience.py
@@ -34,7 +34,7 @@ def _sanitize_load_args(*args):
     if PY3:
         from pathlib import PosixPath
         acceptable_input_types.append(PosixPath)
-    args = [os.path.expanduser(arg) if isinstance(arg, tuple(acceptable_input_types))
+    args = [os.path.expanduser(str(arg)) if isinstance(arg, tuple(acceptable_input_types))
             else arg for arg in args]
     return args
 
--- a/yt/convenience.py
+++ b/yt/convenience.py
@@ -28,6 +28,16 @@ from yt.utilities.exceptions import \
     YTSimulationNotIdentified
 from yt.utilities.hierarchy_inspection import find_lowest_subclasses
 
+def _sanitize_load_args(args):
+    """Filter out non-string-like arguments and replace "~" with user name"""
+    acceptable_input_types = list(string_types)
+    if PY3:
+        from pathlib import PosixPath
+        acceptable_input_types.append(PosixPath)
+    args = [os.path.expanduser(arg) if isinstance(arg, tuple(acceptable_input_types))
+            else arg for arg in args]
+    return args
+
 def load(*args ,**kwargs):
     """
     This function attempts to determine the base data type of a filename or
@@ -38,12 +48,7 @@ def load(*args ,**kwargs):
     """
     candidates = []
 
-    acceptable_input_types = list(string_types)
-    if PY3:
-        from pathlib import PosixPath
-        acceptable_input_types.append(PosixPath)
-    args = [os.path.expanduser(arg) if isinstance(arg, tuple(acceptable_input_types))
-            else arg for arg in args]
+    args = _sanitize_load_args(args)
     valid_file = []
     for argno, arg in enumerate(args):
         if isinstance(arg, string_types):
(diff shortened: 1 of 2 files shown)


neutrinoceros:
note: pathlib is stdlib in python3 while pathlib2 is a third party back porting for python2...
It's not worth adding it as a new dependency (provided it's not already). In this case, I could make this feature python 3 exclusive.

neutrinoceros:
Alright, in light of the latest failure, better make this python3 only

neutrinoceros:
The failing CI test makes little sense to me. I can't reproduce the error locally in python 3 or python 2, and in any case I seriously doubt this PR affects the units module in any way.

matthewturk:
I agree, I don't see any way this could cause that error.

neutrinoceros:
Wait. Did you just re-triggered it ? is this error reproduced ??

matthewturk:
I did indeed retrigger it...

neutrinoceros:
Sure. An important case to cover imo is the use of a wildcard to instantiate a timeseries. I tried it locally personnal data but I suppose we have at least one data folder that can be used for this, would anyone know of one ?

neutrinoceros:
actually, I'm not sure about how this can be elegantly tested either, as I don't believe the `==` operator to be implemented for datasets. Would it suffice to check that the same data is available ? i.e.
```python
from pathlib import Path

data_file = "output.dat"
ds1 = yt.load(data_file)
ds2 = yt.load(Path(data_file))
s1 = set(ds1.field_list)
s2 = set(ds2.field_list)
# check that the first set is not empty
assert s1
# check that the symmetric difference is empty
assert s1 ^ s2 == set()
```

also, is there something like a `@requires_python3` decorator ?

munkm:
I'm not sure that we have a requires python3 decorator. Though once we switch to 4.0 it won't be necessary, so I don't recommend adding one now. 

You make a good point about the timeseries data. Let me check and get back to you. 

I don't think you need to check with `==` on the datasets. I personally would write this test (or test(s)) to try to load the dataset with minor variations on the path and then check that something occurs in the load function that you'd expect if it's loaded correctly or incorrectly. What you've done checking the field data is a good option (so feel free to use it if you're happy with it), but I might simplify it even further and check that once it's loaded it's is an instance of whatever `Dataset` you're loading. You could check if an invalid Path is passed that you get the error you'd expect too.

neutrinoceros:
actually, I am pretty sure that all the testing we need is this (plus variations with different input string formats)

```python
from pathlib import path
from os.path import expand_users
p1 = "not/a/real/datafile.hdf5"
p2 = Path(p1)

s1 = expand_users(p1)
s2 = expand_users(p2)
assert s1 == s2
```
but that essentially has nothing to do with yt... that's why I did not bother writing the tests from the start. I don't think there's any reason to expect the result of this script to change over time (maybe when python 4.0 comes out, who knows ?)

munkm:
This version of the test wouldn't be verifying that yt.load() is behaving as expected though, right?

neutrinoceros:
It would test the for the one line where things _might_ differ from when one loads from a string.

neutrinoceros:
Come to think of it I think the way to go would be to isolate the first few lines of `yt.load` in a separate method, say `yt._sanitize_load_args`, and test _this_ method extensively.

neutrinoceros:
just added some tests, but there's no mechanism to pass them in python 2 (yet).
Also, I learned that the minimal python version for os.path.expandusers to accept path likes objects as input in actually 3.6 so... do we want to support 3.5 ? (if so I just need to add a str() conversion)

neutrinoceros:
damnit the py3 tests are using 3.5
Maybe this PR would be best suited for yt 4.0

neutrinoceros:
The "new" PR targeting 4.0 is this one https://github.com/yt-project/yt/pull/2405
I'll close the present one for now.

PR #2395, opened 2019-12-16
install sympy 1.4 on appveyor to work around regression in 1.5

ngoldbaum:
This should fix the failures that started cropping up on appveyor recently. I'm going to report this upstream as a regression.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ install:
     - "python --version"
 
     # Install specified version of numpy and dependencies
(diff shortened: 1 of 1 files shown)


ngoldbaum:
I reported https://github.com/sympy/sympy/issues/18054 upstream.

ngoldbaum:
Yes, but I have no idea how long it will take for sympy to do a bugfix release.

munkm:
Oh yeah, that makes sense. I was just curious about what our general process was in situations like these. 🙂

PR #2400, opened 2019-12-18
[yt-4.0] new depth-first octree

AshKelly:
## PR Summary

Added a depth-first octree

Currently, to deposit an SPH field currently you just do

`("PartType0", "any_field_currently_defined_in_the_ds")`

but this needs to change to something similar to the old tree.

## PR Checklist

## Example

```
import numpy as np
import yt


def project_octree_field2d(octree, field, pixels_x=1600, pixels_y=1600):
    refined = octree[("index", "refined")]
    dx = octree[("index", "dx")].d[~refined]/2
    dy = octree[("index", "dy")].d[~refined]/2
    dz = octree[("index", "dz")].d[~refined]/2

    sx = (octree.right_edge[0].d - octree.left_edge[0].d)
    sy = (octree.right_edge[1].d - octree.left_edge[1].d)

    x = octree[("index", "x")].d[~refined] - octree.left_edge[0].d
    y = octree[("index", "y")].d[~refined] - octree.left_edge[1].d

    x /= sx
    dx /= sx

    y /= sy
    dy /= sy

    grid = np.zeros((pixels_x, pixels_y))
    for xi, yi, dxi, dyi, dzi, fieldi in zip(x, y, dx, dy, dz, field):
        x_ind = int(pixels_x * xi)
        y_ind = int(pixels_y * yi)
        x_size = int(pixels_x * dxi) + 1
        y_size = int(pixels_y * dyi) + 1

        x_lower = max(0, x_ind-x_size)
        x_upper = min(x_ind+x_size, pixels_x+1)
        y_lower = max(0, y_ind-y_size)
        y_upper = min(y_ind+y_size, pixels_y+1)
        grid[x_lower:x_upper, y_lower:y_upper] += (fieldi.d*dzi)

    return grid


ds = yt.load("GadgetDiskGalaxy/snapshot_200.hdf5")

n_ref = 32
octree = ds.octree(n_ref=n_ref)

# octree scatter
dens = octree[("PartType0", "density")].to("code_mass/code_length**3")
# octree gather
octree.sph_smoothing_style = 'gather'
dens_g = octree[("PartType0", "density")].to("code_mass/code_length**3")

grid = project_octree_field2d(octree, dens)
grid_g = project_octree_field2d(octree, dens_g)

fig, axes = plt.subplots(1, 3, figsize=(15, 5), dpi=200)

ax = axes[0]
ax.imshow(np.log10(grid+1e-12), origin="lower", extent=[0, 64000, 0, 64000],
          vmax=0.3, vmin=-7.0)

ax = axes[1]
ax.imshow(np.log10(grid_g+1e-12), origin="lower", extent=[0, 64000, 0, 64000],
          vmax=0.3, vmin=-7.0)

ax = axes[2]
proj = yt.ProjectionPlot(ds, 'z', ("PartType0", "density"))
img = proj.frb.data[("PartType0", "density")].to("code_mass/code_length**2").d
ax.imshow(np.log10(img+1e-12).T, origin="lower", extent=[0, 64000, 0, 64000],
          vmax=0.3, vmin=-7.0)

for ax in axes.flatten():
    ax.set_xlim(20000, 64000-20000)
    ax.set_ylim(20000, 64000-20000)
    ax.set_xticks([])
    ax.set_yticks([])

fig.tight_layout()
fig.savefig("test.png")
```
which outputs:
test


ngoldbaum:
Cool! Thank you for getting this in a PR :)

Can you elaborate a little bit more how this changes the behavior of the octree? What does this buy us? Besides simplifying the code a lot it looks like. Since we haven’t done a release of this code yet I think it’s ok to change its behavior, especially if that makes it easier for users to reproduce results they got with old versions of yt. I just want to understand what exactly we’re changing.

Do you understand the test failures? Do the tests just need to be updated to reflect the new ordering?

`mpl_project2d` looks useful for debugging but if you want to add it upsteam it’ll need to be tested and documented. Maybe instead of something that directly interacts with matplotlib you could add a thing that returns an image buffer at some resolution containing a visualization of the projected octree. I think that might be more generally useful and we can leave it to users to actually do the plot with matplotlib.

AshKelly:
Ah, sorry. Yeah I'm planning to add tests and documentation. I should have labelled this as a WIP as there are several other things that need fixing, adding and removing. Unfortunately I'm on mobile as I'm in Amsterdam on holiday for the next few days.

I just wanted to open a PR with the main features Desika needed so he could start testing and using it for his stuff.

ngoldbaum:
Great, feel free to ping me when you’re ready to have someone review it in preparation to merge it.

dnarayanan:
does the octree build every time?  (i.e. it looks like the `force_build` arg was removed)

AshKelly:
Yes, the old save code is useless. However adding a new save feature is much easier because the whole tree is just a few arrays so lends itself nicely to hdf5

AshKelly:
Also please leave feedback on how you want the fields to be named e.g. smoothed, sums and weighted sums

dnarayanan:
happy to thanks! playing with the integration into `powderday` today and tomorrow so will learn more about the behavior/interface as i progress

dnarayanan:
so i can confirm now that this branch works well within powderday (my dust RT package), and i'm able to get identical results as i did with yt3.x's octree!

matthewturk:
@AshKelly can you merge with master?

munkm:
@matthewturk do you mean with `yt-4.0`?

matthewturk:
Oops, yes.

AshKelly:
There's probably a few more things the octree could do (basically anything @dnarayanan wants), but it's probably best to just get this merged. 

I also added the ability to change between `gather' and `scatter' using 
```
octree.sph_smoothing_style = 'gather'
'''
or
``` 
octree.sph_smoothing_style = 'scatter'
'''
before a field call to the octree

doc/source/yt4differences.rst
```diff
@@ -159,24 +157,20 @@ It should be noted that the
...
 
+    ds.sph_smoothing_style = "scatter"
+
     # the density will be calculated using SPH scatter
     density = octree[('PartType0', 'density')]
 
     # this will return the x positions of the octs
     x = octree[('index', 'x')]
 
 The above code can be modified to use the scatter approach by using
```
munkm, line 168:
```suggestion
The above code can be modified to use the gather approach by using
```

doc/source/yt4differences.rst
```diff
@@ -159,24 +157,20 @@ It should be noted that the
     # generate an octree
     octree = ds.octree(left, right, n_ref=64)
 
+    ds.sph_smoothing_style = "scatter"
```
munkm, line 160: It might be worth mentioning here that scatter is the default behavior (as an inline comment or something)

doc/source/yt4differences.rst
```diff
@@ -159,24 +157,20 @@ It should be noted that the
...
-The ``density_factor`` keyword allows the construction of dense octrees
-trees. In a traditional octree, if a leaf has more particles that a critical
-value `n_ref`, then it divides into 8 new children (hence the name oct). The
-value of `density_factor` allows the node to divide into 2^(3*density_factor)
-zones instead. This creates an octree structure similar that used by AMR codes
-like FLASH that make use of an octree of grid patches.
+``ds.sph_smoothing_style = 'gather'`` before any field access.
+
+The octree implementation is very simple. It uses a recursive algorithm to build
+a ``depth-first`` which is consistent with the results from yt-3.
```
munkm, line 172: We don't have documentation anywhere describing a depth-first octree, right? This isn't my area, but it might be worth adding a sentence explaining what this is.

doc/source/yt4differences.rst
```diff
@@ -159,24 +157,20 @@ It should be noted that the
...
-parameter in yt-3.0 that could be passed to ``yt.load``, and determines how many
-particles are in each leaf.
-
-The ``density_factor`` keyword allows the construction of dense octrees
-trees. In a traditional octree, if a leaf has more particles that a critical
-value `n_ref`, then it divides into 8 new children (hence the name oct). The
-value of `density_factor` allows the node to divide into 2^(3*density_factor)
-zones instead. This creates an octree structure similar that used by AMR codes
-like FLASH that make use of an octree of grid patches.
+``ds.sph_smoothing_style = 'gather'`` before any field access.
```
munkm, line 169: so sph_smoothing_style is accessible from both the ds and the octree objects? Do we want to document that here too? e.g what happens if the sph_smoothing_style is set on the ds object but not the octree? What happens when one is changed?

yt/data_objects/tests/test_octree.py
```diff
@@ -1,44 +1,16 @@
...
 def test_building_tree():
     '''
     Test function to build an octree and make sure correct number of particles
     '''
     ds = fake_sph_grid_ds()
-    octree = ds.octree(n_ref=1)
-    assert(type(octree) == YTOctree)
-    assert(octree[('index', 'x')].shape[0] == 456)
-
-def test_saving_loading():
```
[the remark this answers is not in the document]
AshKelly: I haven't added loads and saves yet. I need to though. The loading and saving is much simple now, I just need to output a few arrays. I'm thinking about hdf5, do you have strong opinions?

yt/data_objects/construction_data_containers.py
```diff
@@ -2139,7 +2139,7 @@ def _upload_to_sketchfab(self, data, files):
 class YTOctree(YTSelectionContainer3D):
     """A 3D region with all the data filled into an octree. This container
     will mean deposit particle fields onto octs using a kernel and SPH
-    smoothing.
+    smoothing. The octree is built in a depth-first fashion.
 
     Parameters
```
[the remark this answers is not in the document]
AshKelly: Thanks! Removed.

doc/source/yt4differences.rst
AshKelly: The explanation is a little abstract, but hopefully that is more informative than before

doc/source/yt4differences.rst
AshKelly: I didn't even catch this. I think that is a bad precedent to set. I have instead reverted to just defining those properties on the `ds` and then I the octree grabs them each time it smooths a field so they can be updated/viewed throughout a script

yt/data_objects/construction_data_containers.py
```diff
@@ -2303,59 +2314,56 @@ def _sanitize_edge(self, edge, default):
             edge_units = edge.units
         else:
             edge_units = 'code_length'
-        return self.ds.arr(edge, edge_units)
+        return 1.0*self.ds.arr(edge, edge_units, dtype="float64")
```
AshKelly, line 2317: so this 1.0 is a complete fudge. However, if I don't include it then 
`self.ds.arr([0, 0, 0], dtype="float64")` seems to revert to `int64`.
Any ideas what causes this?

matthewturk: I *assume* but am not certain that this is because `unyt` tries to allow things to be in any type, and maybe (?) there's a logical check that doesn't catch the `dtype`?

dnarayanan:
Hi,

just checking in to see if it's possible to get eyes on this PR.  powderday currently requires this PR in order to work in yt4.x for particle-based codes, so would be awesome to have this merged into main eventually

AshKelly:
I've attempted a rebase for the first time in my life... I have no idea how it's gone.

munkm:
I'll come back for round 2 of review today!

yt/data_objects/tests/test_octree.py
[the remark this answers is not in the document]
munkm: hdf5 sounds great to me. I don't have strong opinions for something else. But maybe somebody else does!

stonnes:
@munkm this looks ready to my untrained eye--is it going to be merged soon?

munkm:
It's ready for review! I just keep not setting aside time to fully review it! I will move it up my list. We also need one more reviewer.

stonnes:
@cphyc how would you feel about reviewing this PR?  It needs another person before it can be merged and you would be great for it!

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
 
 """
 
+#-----------------------------------------------------------------------------
+# Copyright (c) 2019, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
```
cphyc, line 15: We have been removing the header comment in all our files.
```suggestion
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+
+    Returns:
+        0
+    """
+    cdef np.int64_t splits[9]
+    cdef np.int64_t i, j, k, n, start, end
+    cdef np.float64_t lx, ly, lz, sz
+
+    if ((tree.pend[node_idx] - tree.pstart[node_idx] > tree.n_ref) and
+            (tree.depth[node_idx] < tree.max_depth)):
```
cphyc, line 91:
```suggestion
    if (
        (tree.pend[node_idx] - tree.pstart[node_idx] > tree.n_ref) and
        (tree.depth[node_idx] < tree.max_depth)
    ):
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+    The `refine` creates 8 sub octs, within our oct. We indenfity the particles
+    in each of the subocts. We then recursively call this function on each of
+    the subocts.
+
+    Args:
+        tree: A pointer to the octree
+        node_idx: The index of the current node we are processing
+
+    Returns:
+        0
```
cphyc, line 84: We are using numpy-like docstrings in our code base.

```suggestion
    Parameters
    ----------
    tree : A pointer to the octree
    node_idx : The index of the current node we are processing

    Returns
    -------
    0
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+        tree.refined[node_idx] = 1
+
+        # As we have decided to refine, we need to know which of the particles
+        # in this oct will go into each of the 8 child octs
+        split_helper(tree, node_idx, splits)
+
+        # Figure out the size of the current oct
+        sx = tree.size[0] / (2**tree.depth[node_idx])
+        sy = tree.size[1] / (2**tree.depth[node_idx])
+        sz = tree.size[2] / (2**tree.depth[node_idx])
```
cphyc, line 106: For performance reason, shouldn't this be stored in a temporary variable? Note that the following suggestion will also require defining `one_over_twotodepth`, and I am wondering whether we shouldn't use `1. / 2**tree.depth[node_idx]` to be on the safe side and prevent this from being interpreted as integer division?
```suggestion
        one_over_twotodepth = 1 / 2**tree.depth[node_idx]
        sx = tree.size[0] * one_over_twotodepth
        sy = tree.size[1] * one_over_twotodepth
        sz = tree.size[2] * one_over_twotodepth
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+                    tree.pstart[child] = start
+                    tree.pend[child] = end
+                    tree.num_nodes += 1
+
+                    # Recursively refine child
+                    octree_build_node(tree, child)
+
+                    n += 1
+
+    return 0
```
cphyc, line 139: Since this function only returns `0`, why isn't it a void function?

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+    """
+    This is the main allocation function in the octree. We allocate all of the
+    arrays we require to store information about every single oct in the tree
+
+    Args:
+        octree: A pointer to the octree
+        num_nodes: The maximum number of nodes to allocate for
+
+    Returns:
+        0
```
cphyc, line 154:
```suggestion
    Parameters
    ----------
    octree : A pointer to the octree
    num_nodes : The maximum number of nodes to allocate for

    Returns
    -------
    0
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+    octree.size = <np.float64_t *> malloc(3 *sizeof(np.float64_t))
+    octree.children = <np.int64_t *> malloc(8 * num_nodes * sizeof(np.int64_t))
+
+    octree.pstart = <np.int64_t *> malloc(num_nodes * sizeof(np.int64_t))
+    octree.pend = <np.int64_t *> malloc(num_nodes * sizeof(np.int64_t))
+
+    octree.refined = <np.uint8_t *> malloc(num_nodes * sizeof(np.int8_t))
+    octree.depth = <np.uint8_t *> malloc(num_nodes * sizeof(np.int8_t))
+
+    return 0
```
cphyc, line 167: Same comment here about the function being `void` rather than `int`.

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+    Case 2: We have finished building the octree and we have used less nodes
+    than we originally allocated. We are now shrinking the octree and giving
+    the spare memory back.
+
+    Args:
+        octree: A pointer to the octree
+        num_nodes: The maximum number of nodes to (re)allocate for
+
+    Returns:
+        0
```
cphyc, line 191:
```suggestion
    See Notes for when we want to re-allocate.
    
    Parameters
    ----------
    octree : A pointer to the octree
    num_nodes : The maximum number of nodes to (re)allocate for

    Returns
    ------
    0

    Notes
    -----
    Why do we want to re-allocate?

    Well 2 cases:

    Case 1: The octree is still building and we have ran out of space, so we
    have asked for an increased number of nodes and are reallocating each array

    Case 2: We have finished building the octree and we have used less nodes
    than we originally allocated. We are now shrinking the octree and giving
    the spare memory back.
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef int octree_deallocate(Octree * octree):
+    """
+    Just free-ing every array we allocated to ensure we don't leak.
+
+    Args:
+        octree: A pointer to the octree
+    Returns:
+        0
```
cphyc, line 248:
```suggestion
    Just free-ing every array we allocated to ensure we don't leak.

    Parameter
    ---------
    octree : pointer to the octree
    
    Returns
    -------
    0
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+
+    free(octree.pstart)
+    free(octree.pend)
+
+    free(octree.refined)
+    free(octree.depth)
+
+    free(octree.pidx)
+
+    return 0
```
cphyc, line 262: Should be void?

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+
+    old_arr_uint = octree.depth
+    octree.depth = <np.uint8_t *> malloc(num_nodes * sizeof(np.int8_t))
+    for i in range(octree.num_nodes):
+        octree.depth[i] = old_arr_uint[i]
+    free(old_arr_uint)
+
+    octree.max_num_nodes = num_nodes
+
+    return 0
```
cphyc, line 236: Should be void?

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
-                    z+=1
-                    split_id+=1
+        We copy the inputted particle positions and make the root node. We then
+        refine the octree until every leaf has either less particles than n_ref
+        or is at the maximum depth.
 
-    @property
-    def size_bytes(self):
-        return sizeof(Node) * self.nodes.size()
+        Finally, we re-alloacte all of the memory required by tree to ensure we
```
cphyc, line 292:
```suggestion
        Finally, we re-allocate all of the memory required by tree to ensure we
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+            input_pos: A memory view of 2D dimensional array which stores the
+                particles positions in the format (num_particles, 3)
+            <>_edge: A 3 element array with the xyz of the lower left
+                (upper right) corner of the octree.
+            n_ref: The maximum number of particles per leaf, if more, the oct
+                will refine
+            max_depth: The maximum depth the octree will refine to. If we set
+                this too high then we may hit a stack overflow due to the
+                recursive nature of the build
+        """
```
cphyc, line 305:
```suggestion
        Parameters
        ----------
        input_pos : 2D memory view
            Particles positions in the format (num_particles, 3)
        {left,right}_edge : array of length 3
            xyz coordinates of the lower left (upper right) corner of the octree.
        n_ref : 
            The maximum number of particles per leaf, if more, the oct
            will refine
        max_depth :
            The maximum depth the octree will refine to. If we set
            this too high then we may hit a stack overflow due to the
            recursive nature of the build
        """
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
-    def n_ref(self):
-        return self._n_ref
+    def node_positions(self):
+        """
+        The centre of every node within the octree
+        """
+        cdef np.npy_intp shape[2]
+        shape[0] = <np.npy_intp> self.c_octree.num_nodes
+        shape[1] = 3
+        arr = np.PyArray_SimpleNewFromData(2, &shape[0], np.NPY_FLOAT64, <void *>self.c_octree.node_positions)
```
cphyc, line 352: What happens if `self.c_octree.node_positions` is freed?

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+
+        In the case that the left_edge/right_edge is not defined then we select
+        a tree that is sufficiently big to contain every particle.
+
+        However, if they are defined then we need to loop through and find out
+        which particles are actually in the tree.
+
+        Args:
+            <>_edge: A 3 element array with the xyz of the lower left
+                (upper right) corner of the octree or NONE
```
cphyc, line 409:
```suggestion
        Parameters
        ----------
        {left,right}_edge : array of size 3 or None
            xyz coordinates of the lower left (upper right) corner of the octree.
            If None, the tree will be made large enough to encompass all particles.
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+                properties used in the deposition
+            num_node: The current node we are checking to see if refined or not
+            use_normalization: Do we want a normalized sph field? If so, fill
+                the buff_den.
+        """
+
+        cdef Octree * tree = self.c_octree
+        cdef double q_ij, diff_x, diff_y, diff_z, diff, sx, sy, sz, size
+        cdef int i
+        cdef long int child_node
```
cphyc, line 540: Should `buff` be `0`-filled before this process?

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+
+                sx = tree.size[0] / (2**tree.depth[child_node])
+                sy = tree.size[1] / (2**tree.depth[child_node])
+                sz = tree.size[2] / (2**tree.depth[child_node])
+                size = math.sqrt(sx*sx + sy*sy + sz*sz)
+
+                if diff - size - hsml < 0:
+                    self.smooth_onto_cells(buff, buff_den, posx, posy, posz,
+                                    hsml, prefactor, prefactor_norm,
+                                    child_node, use_normalization=use_normalization)
```
cphyc, line 572: Note: you could spare some CPU cycles by computing 
```python
local_dx = tree.size[0] / 2.**tree.depth[root_node]
```
when hitting the root node, and then pass `local_dx/2` as an argument to `smooth_onto_cells`.
This could look like
```suggestion
                # THIS CODE IS NOT COMPLETE, `local_d{xyz}` should have been computed at the root node.
                diff_x = tree.node_positions[3*child_node] - posx
                diff_y = tree.node_positions[3*child_node+1] - posy
                diff_z = tree.node_positions[3*child_node+2] - posz
                diff = math.sqrt(diff_x*diff_x + diff_y*diff_y + diff_z*diff_z)

                sx = local_dx
                sy = local_dy
                sz = local_dz
                size = math.sqrt(sx*sx + sy*sy + sz*sz)

                if diff - size - hsml < 0:
                    self.smooth_onto_cells(buff, buff_den, posx, posy, posz,
                                    hsml, prefactor, prefactor_norm,
                                    child_node, use_normalization=use_normalization,
                                    local_dx=local_dx/2, local_dy=local_dy/2, local_dz=local_dz/2)
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+    cdef np.int64_t split = start
 
-    return 0
+    cdef np.float64_t * pos = octree.pposx
+    cdef np.int64_t * pidx = octree.pidx
+
+    for index in range(start, end):
+        if pos[pidx[index]] < value:
+            pidx[split], pidx[index] = pidx[index], pidx[split]
+            split+=1
```
cphyc, line 639:
```suggestion
            split += 1
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+    cdef np.int64_t index
+    cdef np.int64_t split = start
 
-    return 0
+    cdef np.float64_t * pos = octree.pposx
+    cdef np.int64_t * pidx = octree.pidx
+
+    for index in range(start, end):
+        if pos[pidx[index]] < value:
+            pidx[split], pidx[index] = pidx[index], pidx[split]
```
cphyc, line 638: I dig this one-liner to swap the values!

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+            split+=1
+
+    return split
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+cdef int seperate_y(Octree * octree, double value, np.int64_t start, np.int64_t end) nogil except -1:
```
cphyc, line 647: This function looks exactly like `separate_x` and `separate_z`. Shouldn't they be merged into only one?

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
-    '''
-    cdef double lower = left[0]
-    cdef double upper = right[0]
-
-    split_arr[0] = start
-    split_arr[num_children] = end
-
-    split(0, num_children, max_splits, 0, 0, split_arr, idx, pos, left, right,
-          lower, upper)
+cdef int seperate_x(Octree * octree, double value, np.int64_t start, np.int64_t end) nogil except -1:
```
cphyc, line 613: Shouldn't this be `separate_x`?
In addition the function is very similar to separate_{yz}. Maybe you could refactor as
```cython
cdef inline int _separate_helper(np.float64_t * pos, np.int64_t * pidx, double value, const np.int64_t start, np.int64_t end) nogil except -1:
    cdef int index
    for index in range(start, end):
        if pos[pidx[index]] < value:
            pidx[split], pidx[index] = pidx[index], pidx[split]
            split += 1

    return split


cdef int separate_x(Octree * octree, double value, np.int64_t start, np.int64_t end) nogil except -1:
    """docstring"""
    return _separate_helper(octree.pposx, octree.pidx, value, start, end)


cdef int separate_y(Octree * octree, double value, np.int64_t start, np.int64_t end) nogil except -1:
    """docstring"""
    return _separate_helper(octree.pposy, octree.pidx, value, start, end)


cdef int separate_z(Octree * octree, double value, np.int64_t start, np.int64_t end) nogil except -1:
    """docstring"""
    return _separate_helper(octree.pposz, octree.pidx, value, start, end)
```

yt/utilities/lib/cyoctree.pyx
```diff
@@ -1,679 +1,756 @@
...
+    cdef np.float64_t * posz = octree.pposz
+    cdef np.int64_t * pidx = octree.pidx
+
+    for index in range(start, end):
+        if posx[pidx[index]] < right_edge[0] and posx[pidx[index]] > left_edge[0]:
+            if posy[pidx[index]] < right_edge[1] and posy[pidx[index]] > left_edge[1]:
+                if posz[pidx[index]] < right_edge[2] and posz[pidx[index]] > left_edge[2]:
+                    if split < index:
+                        pidx[split], pidx[index] = pidx[index], pidx[split]
+                    split+=1
```
cphyc, line 754:
```suggestion
    cdef np.int64_t * pidx = octree.pidx
    cdef np.int64_t this_idx
    cdef np.float64_t px, py, pz
    for index in range(start, end):
        this_idx = pidx[index]
        px = posx[this_idx]
        if left_edge[0] < px < right_edge[0]:
            py = posy[this_idx]
            if left_edge[1] < py < right_edge[1]:
                pz = posz[this_idx]
                if left_edge[2] < pz < right_edge[2]:
                    if split < index:
                        pidx[split], pidx[index] = pidx[index], pidx[split]
                    split += 1
```

cphyc (commented):
Overall this looks good to me, thanks so much for doing this!

- Among my comments, I left numerous code changes suggestions to refactor the docstrings. I guess that you should either ignore them (I'm fine with this or update *all* docstrings (I did not leave a comment for each).
- I have some concern about a numpy array having its data managed in C: what happens if the data is `free`d?
- I suggested some code refactors to prevent code repetition prevent code repetition 

On a tangential note, most of the code here looks similar to the one I wrote for raycasting through octrees, and is similar to the octree structure used for ART/Ramses frontends. This shouldn't be done here, but we should really clean this up at some point™.

yt/utilities/lib/cyoctree.pyx
cphyc: Note: this should be ignore if you use the code refactor above.

AshKelly:
Thanks for the review @cphyc! I'll go through it after lunch and make your changes. I actually really dislike this code, I wrote the bulk of it over 2.5 years ago, and modified it slightly about 15 months ago. I wouldn't write it this way now... but, whatever!

matthewturk:
@AshKelly that is so relatable

AshKelly:
@matthewturk I was hoping that as I got more experienced my code would converge, and I'd start liking stuff I wrote 2 years ago. But so far, I think I'm diverging from that

yt/data_objects/construction_data_containers.py
```diff
@@ -2774,36 +2782,29 @@ def get_data(self, fields=None):
...
             return
 
-        # not sure on the best way to do this
-        if isinstance(fields, list) and len(fields) > 1:
-            for field in fields:
-                self.get_data(field)
-            return
-        elif isinstance(fields, list):
-            fields = fields[0]
+        sph_ptypes = getattr(self.ds, "_sph_ptypes", ["PartType0"])
```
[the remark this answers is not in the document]
AshKelly: I think the reason we don't fall back to `["all"]` is because we want the type being used to have smoothing lengths generated. Maybe that doesn't matter anymore, I'll think about this

AshKelly:
I'm planning to check if the re-allocation has been successful here:
https://github.com/yt-project/yt/blob/a1882ceebd348e68650cf749bd9ac179f68a5549/yt/utilities/lib/cyoctree.pyx#L92

But since this is in a recursive function what do you think the best to do this is? A return value of 1 which propogates back up the recursive call, and ensures we don't try and do any further work? or, raising an exception directly in cython?

cphyc:
> I'm planning to check if the re-allocation has been successful here:
> 
> https://github.com/yt-project/yt/blob/a1882ceebd348e68650cf749bd9ac179f68a5549/yt/utilities/lib/cyoctree.pyx#L92
> 
> But since this is in a recursive function what do you think the best to do this is? A return value of 1 which propogates back up the recursive call, and ensures we don't try and do any further work? or, raising an exception directly in cython?

I think the best would be to propagate a value of `1` up the stack if something wrong happens and eventually raise an Exception on the caller side? I'm concerned about performance with raising exception in cython, but tbh I have no idea if and how costly that would be.

matthewturk:
You can specify that some return values from Cython *mean* exceptions; I
think this is the "except -1" or something you might see in other functions
in yt.

AshKelly:
Thanks both! I didn't actually realise the except notation worked like that.

Also, sorry to be a nuisance, is there a linter for `.pyx` files I can run? I just want to make sure the code, and doc strings are better formatted before I push anything else.

yt/data_objects/tests/test_octree.py
```diff
@@ -85,11 +58,14 @@ def test_sph_interpolation_gather():
     """
     ds = fake_sph_grid_ds(hsml_factor=26.0)
     ds.index
+    ds._sph_ptypes = ("io",)
```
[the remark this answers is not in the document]
AshKelly: This is just for testing purposes.

yt/data_objects/tests/test_octree.py
```diff
@@ -88,3 +88,53 @@ def test_sph_interpolation_gather():
...
+            2,
+            2,
+            2,
+            2,
+            2,
+            2,
+        ],
+        dtype=np.int64,
+    )
+    assert_equal(depth, depth_ans)
```
cphyc, line 123: I'd suggest 
```suggestion
    depth_ans = np.array([0] + [1]*8 + [2]*8, dtype=np.int64)
    assert_equal(depth, depth_ans)
```

AshKelly: Lol! After this I went for that style on the next test with the intention to come back, then my memory failed me.

yt/mods.py
```diff
@@ -3,7 +3,6 @@
 #
 
 import os
-import numpy as np
```
AshKelly: Nice! suspected this would happen

cphyc (approved):

yt/data_objects/construction_data_containers.py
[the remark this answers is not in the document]
jzuhone: This assumes a Gadget or Gadget-like dataset--I think we need to have a fallback that checks the particle types in the dataset and selects the one which is likely to have the smoothing length.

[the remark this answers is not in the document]
AshKelly: yeah, that is true! Is there already a place in the code that performs this check? I assume I'm not the first to come across some logic looking for particles with smoothing lengths.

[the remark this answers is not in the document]
jzuhone: There isn't, actually--we always rely on the datasets themselves to specify what the SPH type is. Let me see if I can come up with something.

Which makes me wonder--is there a reason you wouldn't expect `_sph_ptypes` to be defined here?

jzuhone, line 2784: Something like this:
```suggestion
        if hasattr(self.ds, "_sph_ptypes"):
            sph_ptypes = self.ds._sph_ptypes
        else:
            sph_ptypes = tuple(value for value in self.ds.particle_types_raw if value in ["PartType0","Gas","gas","io"])
            
        if len(sph_ptypes) ==  0: raise RuntimeError 
```

jzuhone: @AshKelly what do you think about the change I suggested?

AshKelly: Thanks @jzuhone, this is great! I was a little hesitant to hard code the `["PartType0","Gas","gas","io"]` but I guess these options shouldn't really change.

jzuhone: I agree tho--we need to find something better long-term. Should be able to, but for now this is fine.

jzuhone (approved):

jzuhone:
@dnarayanan this got merged in

dnarayanan:
hooray!!!!  thanks so much all -- this is wonderful!

PR #2401, opened 2019-12-18
Pytest infrastructure

jcoughlin11:
## PR Summary

This PR is the first in a series of PRs related to breaking up PR https://github.com/yt-project/yt/pull/2286, which deals with converting yt's testing framework from nose to pytest. The purpose of breaking the aforementioned PR up into smaller chunks is to facilitate a smoother review process. Given the complex history of https://github.com/yt-project/yt/pull/2286, this process is hopefully easier and smoother than cherry-picking would be.

This PR brings over three core changes from https://github.com/yt-project/yt/pull/2286. The first is the root conftest.py file. This file handles all of the core pytest configuration and fixtures.

The second change is the addition of yt/utilities/answer_testing/utils.py, which is a new file created in https://github.com/yt-project/yt/pull/2286 that brings together a lot of useful helper functions and decorators that are used throughout the answer tests.

The third and final change of this PR is the addition of the AnswerTest class to yt/utilities/answer_testing/framework.py. In https://github.com/yt-project/yt/pull/2286, this class has supplanted all of the original framework.py code, but here, for ease of transitioning purposes, it has simply been appended.

Originally, every answer test (such as the field values test) was its own class, complete with constructor, run, and compare methods. These tests were also yield tests. Pytest does not allow test classes to have constructors and no longer supports yield tests. As such, in the new paradigm, each of these answer test classes (such as field values test) are now methods of the AnswerTest class. This allows for test classes (such as TestEnzo) to inherit from AnswerTest and gain access to the tests it needs. Additionally, this allows for easier integration with pytest's parametrize feature, which is what is now recommended over yielding.

My thought on moving forward is, since this is such a major change, once this PR is accepted, rather than merging straight into master and mucking it up, the additional PRs in this series would be created off of this branch and then merged into this branch upon acceptance. Once all of those have been merged here, then this PR can be merged into master, hopefully providing a smoother transition from nose to pytest.  I'm open to other suggestions, though!

## PR Checklist


ngoldbaum:
Hi Jared, I haven’t looked at this closely but just a small style nit: it’s not necessary to have comments with the names of all the functions about their definitions, as one can see the name on the function definition. I’d say just delete all the comments that just contain a function name.

conftest.py
```diff
@@ -0,0 +1,193 @@
...
+            item.add_marker(skip_answer)
+        # If it's an answer test that requires big data and the CL
+        # option hasn't been set, skip it
+        if "big_data" in item.keywords and not config.getoption("--answer-big-data"):
+            item.add_marker(skip_big)
+
+
+#============================================
+#                   temp_dir
+#============================================
```
neutrinoceros: Is there a good reason not to use std lib here ?
https://docs.python.org/3.7/library/tempfile.html

yt/utilities/answer_testing/utils.py
```diff
@@ -0,0 +1,491 @@
...
+    return plot
+
+
+#============================================
+#               create_plot2
+#============================================
+def create_plot2(data_source, x_field, y_field, z_field,
+                plot_type, plot_kwargs=None):
+    # plot_type should be a string
+    # plot_kwargs should be a dict
```
neutrinoceros: todo: apply numpy style to this (and make it a docstring)

matthewturk:
Can you add pytest to the requirements, as well?

conftest.py
matthewturk: I think it is, right?  It's using `tempfile.mkdtemp()` if not using specific arguments.

yt/utilities/answer_testing/utils.py
munkm: Here's an example @jcoughlin11 

https://yt-project.org/doc/developing/developing.html#docstrings

conftest.py
neutrinoceros: oops, went much to fast here, sorry.

jcoughlin11:
Thanks for the feedback! I think I've addressed all of the comments so far (removed the function title headers, made the docstrings NumPy style, and added pytest to the list of requirements in setup.py). Let me know if I missed anything!

setup.py
```diff
@@ -387,6 +387,7 @@ def run(self):
         'sympy>=1.0',
         'numpy>=1.10.4',
         'IPython>=1.0',
+        'pytest>=5.2',
```
ngoldbaum: Wait, why is this getting added here? Is pytest now necessary to import yt? We don’t have nose here already, for example.

ngoldbaum: I think Matt may have been referring to `tests/test_requirements.txt` and `tests/test_minimal_requirements.txt`. Appveyor doesn’t use that so you’ll have to update appveyor.yml seperately. I don’t think this PR should add a new runtime dependency to yt, only a test dependency.

ngoldbaum:
Also if you have questions about getting the tests to pass on this PR please feel free to say hi on slack or here and ask questions :)

setup.py
jcoughlin11: Oh, my bad, I'll fix that!

yt/utilities/answer_testing/framework.py
```diff
@@ -63,6 +81,332 @@
...
+            ff = f[1]
+            if not ff.endswith("_Density"):
+                continue
+            start_strings = ["Electron_", "SFR_", "Forming_Stellar_",
+                             "Dark_Matter", "Star_Particle_"]
+            if any([ff.startswith(ss) for ss in start_strings]):
+                continue
+            dens_enzo -= dd[f]
+        delta_enzo = np.abs(dens_enzo / dd["Density"])
+        np.testing.assert_almost_equal(delta_yt, delta_enzo)
```
ngoldbaum: This should probably be moved to the Enzo frontend since it's not saving any answers and isn't really answer-testing specific.

yt/utilities/answer_testing/framework.py
```diff
@@ -63,6 +81,332 @@
...
+        ppv_hd = self.pixelized_projection_values_test(ds, axis, field, weight, ds_obj)
+        hex_digests['pixelized_projection_values'] = ppv_hd 
+        return hex_digests
+
+    def generic_array_test(self, func, args=None, kwargs=None):
+        if args is None:
+            args = []
+        if kwargs is None:
+            kwargs = {}
+        return func(*args, **kwargs)
```
ngoldbaum: Are you sure this is doing the same thing as the old `generic_array_test`? It doesn't appear to be saving or comparing answers, for example.

yt/utilities/answer_testing/framework.py
```diff
@@ -63,6 +81,332 @@
...
+        else:
+            particle_type = False
+        if particle_type is False:
+            ppv_hd = self.pixelized_projection_values_test(ds, axis, field, weight, ds_obj)
+            hex_digests['pixelized_projection_values'] = ppv_hd
+        fv_hd = self.field_values_test(ds, field, ds_obj, particle_type=particle_type)
+        hex_digests['field_values'] = fv_hd
+        return hex_digests
+
+    def yt_field_test(self, ds, field, geometric):
```
ngoldbaum: maybe call this `test_field_size_and_mean`? `yt_field_test` indicates to me that this is doing something with all of the field values in the test. Plus we have plenty of other field tests in here.

yt/utilities/answer_testing/framework.py
```diff
@@ -63,6 +81,332 @@
...
+        plot = utils.create_plot2(data_source, x_field, y_field,
+                                z_field, plot_type, plot_kwargs)
+        attr = getattr(plot, attr_name)
+        attr(*attr_args[0], **attr_args[1])
+        tmpfd, tmpname = tempfile.mkstemp(suffix='.png')
+        os.close(tmpfd)
+        plot.save(name=tmpname)
+        image = mpimg.imread(tmpname)
+        os.remove(tmpname)
+        return image
```
ngoldbaum: For the image tests was the ultimate decision not to save the hashes of the images? That's fine, just want to double-check.

ngoldbaum:
Looks like pytest 5.2.2 never came out on python3. If you don't care about Python2 for this PR then I think you should target this at the yt-4.0 branch. If you don't want to do that then you'll need to make sure everything works correctly on python2 and python3 or start a conversation about dropping python2 support on master.

tests/test_minimal_requirements.txt
```diff
@@ -3,6 +3,7 @@ matplotlib==1.5.3
 sympy==1.0
 nose==1.3.7
 nose-timer==0.7.3
+pytest==5.2.2
```
ngoldbaum: Is this the earliest version we're gonna support? Have you tried using older versions of pytest? If it works with older versions then in this file indicate the oldest version we're supporting, a job will then run on travis to make sure that old version works without error. Having a little flexibility here makes it easier on developers who might get pytest shipped with their OS or in a distribution like anaconda.

munkm:
> My thought on moving forward is, since this is such a major change, once this PR is accepted, rather than merging straight into master and mucking it up, the additional PRs in this series would be created off of this branch and then merged into this branch upon acceptance. Once all of those have been merged here, then this PR can be merged into master, hopefully providing a smoother transition from nose to pytest. 

Just so it's also announced in this PR -- Jared sent out an e-mail to yt-dev with a modified process of how we'll be transitioning to pytest based on the work that he has done. It differs slightly than what was proposed in the quoted text.

This PR sets up the pytest infrastructure and when it is approved we'll merge it into master. To support this, we will add an additional job for Jenkins to run pytest separately from nose. As the answer test PRs come in, they will run on the new pytest Jenkins jobs, while the unchanged answer tests will remain running on nose. Eventually all of the answers will be using pytest and we can remove the old nose Jenkins jobs.

yt/utilities/answer_testing/framework.py
ngoldbaum: Oh I see, I guess it just needs a usage example to be clearer, but I understand how this works now. You pass it a function and arguments and keyword arguments to that function and it saves the result of that function.

yt/utilities/answer_testing/framework.py
```diff
@@ -63,6 +81,332 @@
 _latest_local = ytcfg.get("yt", "local_standard_filename")
 _url_path = ytcfg.get("yt", "answer_tests_url")
 
+
+# New AnswerTestclass
+class AnswerTest():
```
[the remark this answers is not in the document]
jcoughlin11: That makes sense. I've done as you suggested. I only had them together in a class because I had originally been applying a class-level fixture to the AnswerTest class, but that is no longer the case so there's no longer a need for the AnswerTest class. The changes will show up in the next push.

conftest.py
```diff
@@ -0,0 +1,175 @@
...
+
+import pytest
+import yaml
+
+from yt.config import ytcfg
+from yt.utilities.answer_testing import utils
+
+
+# Global variables can be added to the pytest namespace
+pytest.answer_files = {}
```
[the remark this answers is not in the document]
jcoughlin11: I found it in a stack overflow answer. I can't remember why I even needed to do that in the first place, so I've changed it to a module-level variable.

conftest.py
```diff
@@ -0,0 +1,175 @@
...
+
+# List of answer files
+answer_file_list = 'tests/tests.yaml'
+answer_dir = os.path.join(ytcfg.get('yt', 'test_data_dir'), 'answers')
+
+
+def pytest_addoption(parser):
```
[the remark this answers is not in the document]
jcoughlin11: I just tried it in a minimum working example and it turns out that it can't. It complains about pytest_add_option being an unknown hook. I could probably register it as a hook or some such thing, if you'd like?

conftest.py
```diff
@@ -0,0 +1,175 @@
...
+    # file
+    with open(answer_file_list, 'r') as f:
+        pytest.answer_files = yaml.safe_load(f)
+    # Register custom marks for answer tests and big data
+    config.addinivalue_line('markers', 'answer_test: Run the answer tests.')
+    config.addinivalue_line('markers', 'big_data: Run answer tests that require'
+        ' large data files.')
+
+
+def pytest_collection_modifyitems(config, items):
```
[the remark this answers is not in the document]
jcoughlin11: Renaming also causes an error, same as when changing pytest_addoption -> pytest_add_option.

conftest.py
```diff
@@ -0,0 +1,175 @@
...
+        if "answer_test" in item.keywords and not config.getoption("--with-answer-testing"):
+            item.add_marker(skip_answer)
+        # If it's an answer test that requires big data and the CL
+        # option hasn't been set, skip it
+        if "big_data" in item.keywords and not config.getoption("--answer-big-data"):
+            item.add_marker(skip_big)
+
+
+@pytest.fixture(scope='function')
+def temp_dir():
```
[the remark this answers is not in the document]
jcoughlin11: I didn't use either the built-in tmp_path or tmpdir fixtures because those fixtures only provide temporary directory names, they don't actually take care of changing the working directory to the created temporary directories. I figured it was easier to implement that functionality as a fixture than it was to have to call os.chdir each time. I suppose I could leave the current working directory as is and simply append tmpdir or tmp_path to the file name so that when the temporary data is saved, it gets saved to the temporary directory without having to change directories?

conftest.py
```diff
@@ -0,0 +1,175 @@
...
+    # co_varnames is all of the variable names local to the function
+    # starting with self, then the passed args, then the vars defined
+    # in the function body. This excludes fixture names
+    args = func.__code__.co_varnames[1:func.__code__.co_argcount]
+    # funcargs includes the names and values of all arguments, including
+    # fixtures, so we use args to weed out the fixtures. Need to have
+    # special treatment of the data files loaded in fixtures for the
+    # frontends
+    for key, val in request.node.funcargs.items():
+        if key in args and not key.startswith('ds_'):
```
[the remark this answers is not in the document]
jcoughlin11: @ngoldbaum This one has me stumped. I agree that doesn't seem like the best way to do this, but I can't seem to find another way. The purpose of this function is to generate a dictionary of {test_parameter_name : test_parameter_value} pairs that can then be written to the answer file. This allows for easier debugging when a test fails. The crux of the issue is that pytest considers parametrized function arguments as fixtures, so all of the introspecting methods I can find that pytest provides (such as request.node.funcargs) include the non-function-argument parameters (such as hashing and answer_file). I use test parameterization so that each combination of test parameters for a given test would get it's own hash and entry in the answer file. Any thoughts and/or suggestions on fixing this would be greatly appreciated, whenever you get a chance! Thanks!

yt/utilities/answer_testing/utils.py
```diff
@@ -0,0 +1,459 @@
...
+            Hex digest of the hashed data.
+    """
+    # Sometimes md5 complains that the data is not contiguous, so we
+    # make it so here
+    if isinstance(data, np.ndarray):
+        data = np.ascontiguousarray(data)
+    # Try to hash. Some tests return hashable types (like ndarrays) and
+    # others don't (such as dictionaries)
+    try:
+        hd = hashlib.md5(data).hexdigest()
```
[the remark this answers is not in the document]
jcoughlin11: I haven't done any format benchmarking, but just from running it, it seems like the time to actually run the test is the dominant factor compared to actually performing the hashing. I did do some small testing with fnv_hash, though, and it is 1-1.2x faster on the numpy arrays I tried, so it could definitely still be worth going that route, but I wouldn't say that hashing is the bottleneck at the moment.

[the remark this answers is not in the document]
ngoldbaum: If it’s only 20% faster to use fnv this is fine as-is.

yt/utilities/answer_testing/framework.py
jcoughlin11: The image data hashes get saved.

tests/test_minimal_requirements.txt
jcoughlin11: The last version of pytest that supports python2 is 4.6.x. and starting with 5.0, there's only python3 support. I'll look into 4.6, but from pytest's changelog it seems like there were several bugfixes made to test parameterization. I'm not sure yet if those affect yt.

jcoughlin11: Using pytest 4.6 with python 3 doesn't seem to work, but using it with python 2.7 does. While the tests run, some of the hashes differ between the pytest 4.6 + python 2.7 setup and the pytest 5.2 python 3.7 setup, which is not great.

matthewturk:
Could that be the dictionary key ordering that I think I convinced you wasn't a big deal?  Can we narrow down the actual differences?

In principle it would not be the worst thing to have different values for py2 and py3 if we at least understood very, very well where those differences arose.

jcoughlin11:
Not sure, but that's a good idea! I'll check!

ngoldbaum:
I brought this up before but I think it bears repeating. Another thing that might be worth considering is dropping python2 support on the master branch. Is it really worth supporting python2 for 0 (if the next release is yt 4.0) or 1 yt releases (if we do another 3.x release)?

jcoughlin11:
Here's a gist with the differences for the Enzo frontend: https://gist.github.com/jcoughlin11/e76c21ca2f157abbbf075e44a24e258c

There are two files. The first was generated using python 3.6 and pytest 5.2 and the second was generated with python 2.7 and pytest 4.6. The tests with differing hashes are: grid_hierarchy, grid_values, projection_values, and pixelized_projection_values.

jzuhone:
FWIW, I think dropping Python 2 makes sense for 4.0. I think that when we decided on keeping it for 4.0 it was a while back and we thought that we’d have it out quicker. Now it’s later and we’re deeper into general lack of Python 2 support, so it makes less sense to keep it.

jcoughlin11:
The tests moved into answer_tests.py appear to be causing CI failures (which makes sense, since they're pytest infrastructure tests). What is the best way to add those tests to a nose blacklist so that they aren't currently run?

matthewturk:
Where are we at?  Are we blocked on deciding on dropping py2 for the current 3.x series?

jcoughlin11:
@matthewturk There seems to be an issue when installing pytest. At first it was that I had specified pytest version 5.2, but that's not available for python2, so I added version 4.6 for python 2 (and a python_version >= '3.0' for 5.2) in the test_requirements.txt and test_minimal_requirements.txt. However, I had made a syntax error. I fixed that and pushed changes a little while ago, but, from the looks of things, it just failed again, so I'll have to look at why.

jcoughlin11:
There were now three more errors. I fixed the first one, which was that pyyaml needs to be a test requirement for the pytest framework, so I added it. The second was what I think was a bug in test_requires_backend. That's separate from this, though, so I'll issue a different PR on it. I fixed it, though. The third is an error in test_geo_projections: setUp() takes exactly one argument (0 given). Not sure what to do about that, but I'll look.

matthewturk:
@jcoughlin11 so am I understanding correctly that this would *not* require dropping py2?

jcoughlin11:
@matthewturk Correct. The only thing to keep in mind is that some of the hashes for some of the answer tests are different in python2 than they are in python3. I **think** this is due to the differences in how the two versions handle dictionary ordering, but I haven't had a chance to look into that in depth, yet. Also, I don't really understand the errors in the raw travis output. It seems that the setUp method for test_geo_projections isn't right? But it seems to look the same as the other implementations of setUp that I looked at. The other remaining error is a pytest thing. Nose is collecting a test I don't want it to and complaining when it can't find the fixtures. I'll fix that one.

matthewturk:
@munkm can you take a look at the geo projection issue?

Xarthisius:
@jcoughlin11 geo projection test fails because you're wrapping it with `requires_module` decorator that's meant for functions. Try wrapping individual methods of `TestGeoProjections`

munkm:
@Xarthisius I wonder why that error appeared now? That class has been wrapped with the `requires_module` decorator for a while. 

Not necessary for this PR, but the `LightConeProjectionTest` class is also wrapped with a `requires_module` decorator. I don't think that will be necessary to update, since it's in analysis_modules.

Xarthisius:
> @Xarthisius I wonder why that error appeared now?

Jared made some changes to it in this PR.

matthewturk:
Hey, all the tests are passing!  🎆

matthewturk:
We think that the py2/py3 stuff might be fixed by judicious use of `bytearray` in lieu of `bytes`, as per https://stackoverflow.com/questions/1740696/bytes-vs-bytearray-in-python-2-6-and-3

jcoughlin11:
@ngoldbaum @neutrinoceros Now that the tests all pass, I was wondering if you wouldn't mind giving this a review when you get the chance? I updated the api.rst file, but I'm not sure I did it correctly, so thoughts on that would be greatly appreciated.  Thank you!

conftest.py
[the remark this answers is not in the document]
ngoldbaum: nah it's fine

conftest.py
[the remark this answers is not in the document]
ngoldbaum: It may be worth asking one of the pytest developers on e.g. their gitter chat. I'm still nervous about this but don't have time to dive in and help you with the code.

doc/source/reference/api/api.rst
```diff
@@ -938,3 +938,31 @@ These are yt-provided functions:
...
    ~yt.testing.fake_random_ds
    ~yt.testing.expand_keywords
+
+These are for the pytest infrastructure:
+
+.. autosummary::
+
+    ~conftest.tempdir
+    ~conftest.answer_file
+    ~conftest.hashing
```
ngoldbaum: Do these show up in the built API docs? You can see the built docs as a build artifact of the docs test job on jenkins.

If not you might need to move them into the yt module and then import them in conftest.py.

jcoughlin11: They show up, but the functions from conftest.py can't actually be interacted with (that is, they are not links). I'll look into fixing it.

Xarthisius:
@jcoughlin11 could you please rebase or merge to remove `appveyor.yml` conflict?

jcoughlin11:
@Xarthisius I did a patch from master to update appveyor.yml, but that removed pytest from the list of packages to install, so I added it back in, but apparently that still causes a conflict. What is the right way to resolve this? Sorry for the bother, but git is very strange sometimes.

jcoughlin11:
I think I fixed it.

matthewturk:
So I think this is ready to go, but I think we need sign-off from at least two other folks.  (I sign off, but I don't know that it counts since I have some commits in this somewhere, I think.)

PR #2405, opened 2019-12-21
allow loading from path objects (py3.6 or above)

neutrinoceros:
this is a second try after #2392 
This PR is targeting yt-4.0 as the feature is much more easy to write for python 3.6 and above.

if we want to support python 3.5 too, an additional step is required to convert pathlike-non-str input to str.

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -140,6 +140,6 @@ answer_tests:
 other_tests:
   unittests:
      - '--exclude=test_mesh_slices'  # disable randomly failing test
-     - '--exclude=test_load_from_path'  # PY2
+     - '--exclude=test_load_from_path'  # py2
   cookbook:
      - 'doc/source/cookbook/tests/test_cookbook.py'
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -140,5 +140,6 @@ answer_tests:
 other_tests:
   unittests:
      - '--exclude=test_mesh_slices'  # disable randomly failing test
+     - '--exclude=test_load_from_path'  # PY2
   cookbook:
      - 'doc/source/cookbook/tests/test_cookbook.py'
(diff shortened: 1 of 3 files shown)


matthewturk:
@neutrinoceros are these test failures related to this PR?  I'm not quite sure that the ones about units are.

neutrinoceros:
Hard to tell, I don't remember why I did not go through with this branch. I'll rebase and test it locally before pushing.

neutrinoceros:
welp. Still failing. As I don't understand how to setup yt 4.0 right now (even in a separate env), I won't be able to solve those failures right now.

neutrinoceros:
Since we're not supporting python 2 in the master branch anymore, I'll rebase and change target to master.

neutrinoceros:
py2 build falling is expected. On the other hand, appveyor is also failing the tests, while they _do_ pass when I run locally. I don't understand what causes the failure there.

Xarthisius:
Try adding this:

```diff
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -140,5 +140,6 @@ answer_tests:
 other_tests:
   unittests:
      - '--exclude=test_mesh_slices'  # disable randomly failing test
+     - '--exclude=test_load_from_path'  # PY2
   cookbook:
      - 'doc/source/cookbook/tests/test_cookbook.py'
```

tests/tests.yaml
```diff
@@ -140,5 +140,6 @@ answer_tests:
 other_tests:
   unittests:
      - '--exclude=test_mesh_slices'  # disable randomly failing test
+     - '--exclude=test_load_from_path'  # PY2
```
Xarthisius, line 143: oops, it was disabled for py3 too. It's apparently case sensitive, it should be "py2". Sorry for the wrong example.

neutrinoceros: Oh, nice catch !

neutrinoceros:
Thanks to @Xarthisius , I think this is now ready for review !

PR #2407, opened 2019-12-30
add support for sympy 1.5

ngoldbaum:
This fixes test failures on the latest version of sympy. I will also be issuing a PR with a similar fix for unyt.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ install:
     - "python --version"
 
     # Install specified version of numpy and dependencies
-    - "conda install --yes -c conda-forge numpy scipy nose setuptools ipython Cython sympy=1.4 fastcache h5py matplotlib mock pandas cartopy conda-build pyyaml"
+    - "conda install --yes -c conda-forge numpy scipy nose setuptools ipython Cython sympy fastcache h5py matplotlib mock pandas cartopy conda-build pyyaml"
     - "conda develop -b ."
 
 # Not a .NET project
--- a/tests/test_requirements.txt
+++ b/tests/test_requirements.txt
@@ -15,7 +15,7 @@ nose==1.3.7
 pandas==0.23.4
 requests==2.20.0
 scipy==1.1.0
-sympy==1.3
+sympy==1.5
 pyqt5==5.11.3; python_version >= '3.0'
 thingking==1.0.2; python_version < '3.0'
 pint==0.8.1
--- a/yt/units/unit_object.py
+++ b/yt/units/unit_object.py
@@ -130,7 +130,7 @@ def get_latex_representation(expr, registry):
     prefix = None
     if isinstance(expr, Mul):
         coeffs = expr.as_coeff_Mul()
-        if coeffs[0] == 1 or not isinstance(coeffs[0], Float):
+        if coeffs[0] == 1 or not isinstance(coeffs[0], Number):
             pass
         else:
             expr = coeffs[1]
@@ -146,7 +146,7 @@ def get_latex_representation(expr, registry):
     else:
         return latex_repr
 
-unit_text_transform = (auto_positive_symbol, rationalize, auto_number)
+unit_text_transform = (auto_positive_symbol, auto_number, rationalize)
 
 class Unit(Expr):
     """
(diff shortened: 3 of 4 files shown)


ngoldbaum:
Hmm, I'm not able to reproduce the answer test failures locally by playing around with pickles of YTArray data. Is there any chance someone at NCSA could share one of the answer test shelve files so I can reproduce the failures locally?

munkm:
Ok you should be able to download the py3 enzo answer store at http://use.yt/upload/b6209ef0 . Let me know if that works or if you need something else!

ngoldbaum:
So it turns out this is ultimately caused by yt's `Unit` class being a subclass of sympy's `Expr` class. This isn't a problem in unyt since I made it so that unyt doesn't subclass expr anymore.

I'm not sure how to fix this without breaking support for comparing with old unit data.

ngoldbaum:
Ah actually maybe I have an easy fix...

ngoldbaum:
Hmm, darn, looks like that didn't work.

ngoldbaum:
I've tried running the answer tests locally using this branch and the answer file provided by Madicken and the tests pass. I don't understand why they're still failing.

matthewturk:
This is indeed very strange.  Is there any remote possibility that it's platform-specific somehow for linux?

matthewturk:
@yt-fido please test this

Xarthisius:
@yt-fido test this please

Xarthisius:
FWIW I'm able to reproduce one of the failures locally (ubuntu 18.04).

```bash
$ cd codes/yt-project/yt
$ git checkout master
$ git pull
$ git reset --hard origin/master
$ git checkout -b pr_2407
$ git fetch ngoldbaum sympy-1.5
$ git merge FETCH_HEAD 
$ pip install -U sympy==1.5 --user
$ pip install -e . --user &> /dev/null
$ curl -s -JO http://use.yt/upload/b6209ef0
$ python pr_2407.py 
LE [[ True  True  True]]
LE units False
RE False
RE units False
```

where test case is:

```bash
$ cat pr_2407.py 
import shelve
import yt
from yt.testing import assert_equal

ds = yt.load("ToroShockTube/DD0001/data0001")
nle = ds.index.grid_left_edge
nre = ds.index.grid_right_edge

fp = shelve.open("py35_local_enzo_004", "r")
ole = fp["data0001"]["GridHierarchy_data0001_all"]["grid_left_edges"]
ore = fp["data0001"]["GridHierarchy_data0001_all"]["grid_right_edges"]

print("LE", nle == ole)
print("LE units", nle.units.registry["code_length"] == ole.units.registry["code_length"])
print("RE", nre == ore)
print("RE units", nre.units.registry["code_length"] == ore.units.registry["code_length"])
```

ngoldbaum:
Thanks Kacper, that was really helpful!

Let's see if this latest push fixes things.

Note that I tested unyt and it doesn't seem to be affected by this issue for some reason, I think because the dimensions go through sympify.

Xarthisius:
The remaining Enzo failures are a bit magical:

```
import yt
import shelve

with shelve.open("py35_local_enzo_004", "r") as fp:
    ole = fp["data0001"]["GridHierarchy_data0001_all"]["grid_left_edges"]

ds = yt.load("DD0010/moving7_0010")
ds.index.grids[0]["velocity_divergence"]
```

It only triggers when I read something from the shelve. Notice that I don't do anything with it and that it even doesn't belong to the same dataset.

ngoldbaum:
Yeah it has something to do with how data gets stored in caches. I think I have a workaround though...

Xarthisius:
> Yeah it has something to do with how data gets stored in caches. I think I have a workaround though...

There are only two hard things in Computer Science... ;-)

ngoldbaum:
It looks like making `Unit` no longer a subclass of `Expr` was a bit too much of a change in behavior, so I've reverted that.

Also for some reason in some case we can have `unit1.dimensions == unit2.dimensions` but not have `unit1.dimensions / unit2.dimensions == sympy_one`. That's what was ultimately causing the "magic" failure Kacper alluded to in his previous comment. I don't really understand why and don't really want to dive into sympy to figure out the root cause tonight. However I think it's safe to juse use `==` equality - although let's see what the tests say.

PR #2410, opened 2020-01-06
Fix a few more is_all_data logic switches

matthewturk:
These missed in #2391 .  I have inspected and I believe that they are correct with the True/False for when we are and are not using `is_all_data`.

### DIFF
--- a/yt/frontends/gadget/io.py
+++ b/yt/frontends/gadget/io.py
@@ -110,7 +110,7 @@ class IOHandlerGadgetHDF5(IOHandlerSPH):
                 if data_file.total_particles[ptype] == 0:
                     continue
                 g = f["/%s" % ptype]
-                if getattr(selector, 'is_all_data', False):
+                if getattr(selector, 'is_all_data', True):
                     mask = slice(None, None, None)
                     mask_sum = data_file.total_particles[ptype]
                     hsmls = None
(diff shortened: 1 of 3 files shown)


ngoldbaum:
Are there are any data objects for which `obj.is_all_data` is False? If not then do we even need the other branch in the if statement? If so then won’t this change mean we always read all of the data? Does that cause a performance hit when only selecting a small amount of data?

matthewturk:
Good questions, which have led me to reconsider what is going on and why this would result in such a speedup on my end.

`is_all_data` should not be true for objects that are subsets, like rays.

ngoldbaum:
Yeah, it looks like this attribute is only set for the `RegionSelector`, so the ray selector would get the default value from the `getattr` call that you've changed here.

matthewturk:
Yup, I'll dig in more and then either close this or update this and #2383 .

Just as a note, on my machine with this applied, it takes about 1 second to do the ray from the profiling test over there, and 15 seconds before it has been applied.  These numbers are reasonably stable over multiple tests.  So, seems like the particle IO needs a much closer look.

matthewturk:
Yup, what I saw as a speedup really just shows problems with the selective-IO we usually do!  Closing, as this is the exact wrong fix for that problem.

PR #2411, opened 2020-01-07
Gadget io fewer calls to reading

matthewturk:
As mentioned in #2383, we can speed things up a fair bit by reading fewer times from disk.  I tested this with both this terse form and the longer version of `x = c[:,0]` etc and found them about the same.

What *was* happening was that each time we read from the dataset via `__getitem__`, we read from disk, and because we were striding it along a vector axis, we made it even worse.  This formulation reads it once, then we grab views into that one read (post-casting) and use that.  I get a fairly good speedup.

I've got a few merge commits, so merge & squash or something would probably be best here, if this passes muster.


matthewturk:
I *believe* these errors are unrelated, and that this is likely ready for merging.

jzuhone:
Should probably get one more set of eyeballs on it.

PR #2414, opened 2020-01-07
[bugfix] Various bug fixes which came up in daily use of yt-4.0

jzuhone:
## PR Summary

Sorry for the grab-bag PR. This addresses a number of issues that I came across
in daily use of yt-4.0 over the past couple of months. They are as follows:

* The APEC emissivity models need to be scaled by `n_H*n_e`, where `n_H` is the _total_ hydrogen number density, but I was only scaling it by the free proton density. Similarly, the `emission_measure` field needs to have the same scaling. I have also updated the `apec_emissivity` file, which now has the `n_H*n_e` scaling. 
* When SPH pixelization of any kind is done, one must ensure that all of the arrays being passed to the various Cython routines are in the same base, in this case code units. Otherwise scaling factors might be very off (i.e., because "mass" might be in 'g' and "density" might be in code units unless all of the units are converted to the same base beforehand). 
* I was getting unit registry-related errors when attempting to use the temperature field in the GAMER frontend, due to subtle changes in unit registry handling in `unyt`. Simply swapping the order of multiplication in the field definition fixes this.
* I found I could not use the `yt download` command on certain platforms unless I added `import urllib.request` to the top of the `yt/utilities/command_line.py` because I got an `ImportError`.
* FITSImageData instances don't know about code units, so convert any images with code units to CGS.
* One can create new `FITSImageData` instances from combining others or from reading a FITS file from disk. When I previously set this up, I neglected to check that the units and the current time were being properly set from the previous instance. I have fixed this and added tests.

## PR Checklist

- [x] Adds a test for any bugs fixed.

### DIFF
--- a/yt/fields/xray_emission_fields.py
+++ b/yt/fields/xray_emission_fields.py
@@ -211,7 +211,7 @@ def add_xray_emissivity_field(ds, e_min, e_max, redshift=0.0,
         # APEC wants to scale by nH*ne
         def _norm_field(field, data):
             return data[ftype, "H_nuclei_density"]*data[ftype, "El_number_density"]
-    ds.add_field((ftype, "norm_field"), _norm_field, units="cm**6",
+    ds.add_field((ftype, "norm_field"), _norm_field, units="cm**-6",
                  sampling_type='local')
 
     my_si = XrayEmissivityIntegrator(table_type, data_dir=data_dir, 
--- a/yt/fields/xray_emission_fields.py
+++ b/yt/fields/xray_emission_fields.py
@@ -15,7 +15,7 @@ from yt.utilities.linear_interpolators import \
 from yt.units.yt_array import YTArray, YTQuantity
 from yt.utilities.cosmology import Cosmology
 
-data_version = {"cloudy": 3,
+data_version = {"cloudy": 2,
                 "apec": 3}
 
 data_url = "http://yt-project.org/data"
--- a/yt/fields/xray_emission_fields.py
+++ b/yt/fields/xray_emission_fields.py
@@ -205,12 +205,13 @@ def add_xray_emissivity_field(ds, e_min, e_max, redshift=0.0,
 
     if table_type == "cloudy":
         # Cloudy wants to scale by nH**2
-        def _norm_field(field, data):
-            return data[ftype, "H_nuclei_density"]**2
+        other_n = "H_nuclei_density"
     else:
         # APEC wants to scale by nH*ne
-        def _norm_field(field, data):
-            return data[ftype, "H_nuclei_density"]*data[ftype, "El_number_density"]
+        other_n = "El_number_density"
+
+    def _norm_field(field, data):
+        return data[ftype, "H_nuclei_density"]*data[ftype, other_n]
     ds.add_field((ftype, "norm_field"), _norm_field, units="cm**-6",
                  sampling_type='local')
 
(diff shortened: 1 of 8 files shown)


jzuhone:
Sorry, it's been a bit since I've submitted a PR and I've also been unable to keep up with latest developments in testing. Should I expect py2/py3 Jenkins test to pass here?

yt/fields/xray_emission_fields.py
```diff
@@ -204,11 +204,15 @@ def add_xray_emissivity_field(ds, e_min, e_max, redshift=0.0,
                                "Perhaps you should specify a constant metallicity instead?")
 
     if table_type == "cloudy":
-        # Cloudy wants to know the total number density of hydrogen
-        H_field = "H_nuclei_density"
+        # Cloudy wants to scale by nH**2
```
matthewturk: Could we have the second field be a string argument, and just have one function def?

yt/visualization/fits_image.py
```diff
@@ -251,7 +260,7 @@ def __init__(self, data, fields=None, length_unit=None, width=None,
                         hdu.header[key] = float(value.value)
                         hdu.header.comments[key] = "[%s]" % value.units
                 if self.current_time is not None:
-                    hdu.header["time"] = self.current_time
+                    hdu.header["time"] = float(self.current_time.value)
```
munkm: I might just be confused here, so sorry in advance (and it's a little bit outside of the relevance of this PR). 

In the previous few lines you set the `time` header with `value.value`, and then if `current_time` exists you overwrite `time` with that value. Is that right? Would there be a case where `time` exists but `current_time` does not? Can we always assume that the units of `current time` are the same as `time`?

jzuhone: Sorry, this code isn't as well-documented as it should be. That should be my next job. :)

The value that is set above by `value.value` is `header["TUNIT"]`, whereas `header["TIME"]` is set by the `self.current_time` value.

Actually, I don't think that `self.current_time` will ever be `None` given the way I've set things up. So I should probably not have this conditional here.

yt/visualization/tests/test_fits_image.py
```diff
@@ -43,6 +43,12 @@ def test_fits_image():
 
     assert_equal(fid1["density"].data, new_fid1["density"].data)
     assert_equal(fid1["temperature"].data, new_fid1["temperature"].data)
+    assert fid1.length_unit == new_fid1.length_unit
+    assert fid1.time_unit == new_fid1.time_unit
+    assert fid1.mass_unit == new_fid1.mass_unit
+    assert fid1.velocity_unit == new_fid1.velocity_unit
+    assert fid1.magnetic_unit == new_fid1.magnetic_unit
+    assert fid1.current_time == new_fid1.current_time
```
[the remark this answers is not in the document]
jzuhone: I think you're right, will fix.

yt/visualization/fits_image.py
```diff
@@ -250,8 +259,7 @@ def __init__(self, data, fields=None, length_unit=None, width=None,
                     if value is not None:
                         hdu.header[key] = float(value.value)
                         hdu.header.comments[key] = "[%s]" % value.units
-                if self.current_time is not None:
-                    hdu.header["time"] = self.current_time
+                hdu.header["time"] = float(self.current_time.value)
```
munkm: Is it worth having the units of current time here as header comments like you did above, or is that unnecessary?

jzuhone: Good question--it should always be in the time units of the file stored in `header["TUNIT"]`, so it shouldn't be necessary to include it in the comments.

jzuhone:
Would love for @brittonsmith (X-ray changes) or @qobilidop (SPH changes) to have a look at this also.

Issue #2412, opened 2020-01-07
Evaluate benefits of pre-allocation of arrays


matthewturk:
At present, the frontends do an indexing check to identify the size of arrays to allocate before conducting any IO.  For grid-based frontends, this is not terribly onerous (it does cost floating point operations, but a best effort is made to cache those operations) but for particle frontends, it is quite taxing as it requires an IO pass.

The reason this decision was made was to avoid having to do large-scale concatenation of arrays, which results in a memory doubling at the finalization step.  However, this finalization step is often not even required, as nearly all of the operations are chunked anyway.

It is my suspicion that we could speed up *considerably* the operations in yt that use particles if we dropped the pre-allocation requirement, and moved instead to concatenating arrays (or reallocing them) and provide only *upper* bounds on the size of the arrays we expect, rather than *exact* bounds.  But, this is just intuition -- which is how the decision was made initially to do the preallocation!

This issue is a placeholder for evaluating this.  I believe this could be tested in small-scale by changing how `_count_particles` operates, to have it return `None` and in the case of `None`, to have the IO operations (which are mostly consolidated in `yt/utilities/io_handler.py:BaseIOHandler._read_particle_selection`) to grow a list of values rather than filling-as-they-go with a running index.

matthewturk:
My email to yt-dev ([archive link](https://mail.python.org/archives/list/yt-dev@python.org/thread/AZKDWQXORFGZXXV2X4IYAZEWULD33VSR/)) is reproduced here:

> Hi folks,
> 
> Over the last little while, I've been looking at this issue and having some thoughts:
> 
> https://github.com/yt-project/yt/issues/2383
> 
> Back when yt-3.0 was first set up, it was built-in that the arrays were all pre-allocated before being read.  (An important note about this is that this *pre-dates* the demeshening.)  What this means is that typically the process of reading data from any type of frontend goes something like this:
> 
>  * _identify_base_chunk -> figure out how big the whole thing is
>  * either read, or subdivide into chunks
> 
> When a chunk is read -- which includes the chunk style "all" that reads all the data in a given data object in a single go -- the destination buffer is preallocated.  For grid objects, this can be done without reading any data off of disk.  The process is still expensive, but we cache the most-recently-used grid mask every time we count a given grid.
> 
[...]

PR #2416, opened 2020-01-09
ENH: Stop preallocating particles

matthewturk:
This is a reasonably self-contained pull request that removes all preallocation of particles.

With *just* this change, the example code in #2383 goes from 20s per iteration to 5 seconds per iteration.

See discussion in #2412 .


ngoldbaum:
Is `_count_particles_chunks` still needed or can that be deleted in all the particle frontendd?

This looks great!

matthewturk:
That particular function isn't used anywhere else, no.

matthewturk:
These errors look real, and point to some things that need a'fixin', so I will address them.

matthewturk:
As an update -- I've tracked this down to the `OnceIndirect` object behaving differently when it is handed a field that is "empty," which results when we have empty grid objects.  I'm trying to determine at which point this was short-circuited in the previous iteration, or to at least correct the behavior here.

neutrinoceros:
Hey I'm adding the 4.0 label here since this looks like a very nice improvement that's worth shipping with the major release, but feel free to remove it if you don't believe it can be completed on reasonable time scales anymore.

matthewturk:
I'm definitely going to take another shot at this. It had slipped my mind!

matthewturk:
Looks like I've got some work to do!

matthewturk:
I'm going to look at this again in the near future, in collaboration with @chrishavlin .

brittonsmith:
@matthewturk it looks like this isn't too far away. There are a couple conflicts, but other than that, what's left?

matthewturk:
I was getting differences in the results.  But, let's try to pick it up,
and maybe distance has helped!

matthewturk:
so I don't think this error is actually related to this PR :(

neutrinoceros:
Indeed it's eyebrow raising... did you try compiling this branch locally ?

matthewturk:
compiles fine here!  :)

neutrinoceros:
Indeed, the tests also crashed in #3129, were I'm _pretty_ confident they shouldn't.

jzuhone:
FWIW, the same thing is happening with #3127

neutrinoceros:
@matthewturk Now that John got the fix in, can you rebase this please ? :)

matthewturk:
I tried rebasing, and it didn't work like I expected, so I merged.  We can squash/rebase when this goes in to make the history nicer.

jzuhone:
@yt-fido test this please

cphyc:
@yt-fido test this please

munkm:
@matthewturk do you think this should block the 4.0 release? Or maybe could we bump it to 4.1?

brittonsmith:
Too late! I've punted it to the 4.1 milestone, but @matthewturk please return it to 4.0 if you want.

matthewturk:
I think that's good.

matthewturk:
Aiming to get this for 4.0.

cphyc:
The failures seem to be due to some datasets (incl. Gadget) being stored in float32 precision while the pixelization routines expect float64.

The minimal change to make the tests happy is the patch below. Note that we may want to change the signature of other functions in `pixelization_routines.pyx` as well.
```patch
index d3155c8c0..b893f20a4 100644
@@ -81,16 +81,19 @@ cdef extern from "pixelization_constants.hpp":
     int WEDGE_NF
     np.uint8_t wedge_face_defs[MAX_NUM_FACES][2][2]
 
+ctypedef fused any_float:
+    np.float32_t
+    np.float64_t
 
 @cython.cdivision(True)
 @cython.boundscheck(False)
 @cython.wraparound(False)
 def pixelize_cartesian(np.float64_t[:,:] buff,
-                       np.float64_t[:] px,
-                       np.float64_t[:] py,
-                       np.float64_t[:] pdx,
-                       np.float64_t[:] pdy,
-                       np.float64_t[:] data,
+                       any_float[:] px,
+                       any_float[:] py,
+                       any_float[:] pdx,
+                       any_float[:] pdy,
+                       any_float[:] data,
                        bounds,
                        int antialias = 1,
                        period = None,
@@ -981,12 +984,12 @@ cdef class SPHKernelInterpolationTable:
 @cython.cdivision(True)
 def pixelize_sph_kernel_projection(
         np.float64_t[:, :] buff,
-        np.float64_t[:] posx,
-        np.float64_t[:] posy,
-        np.float64_t[:] hsml,
-        np.float64_t[:] pmass,
-        np.float64_t[:] pdens,
-        np.float64_t[:] quantity_to_smooth,
+        any_float[:] posx,
+        any_float[:] posy,
+        any_float[:] hsml,
+        any_float[:] pmass,
+        any_float[:] pdens,
+        any_float[:] quantity_to_smooth,
         bounds,
         kernel_name="cubic",
         weight_field=None,
```

yt/utilities/lib/fp_utils.pxd
```diff
@@ -58,3 +58,7 @@ cdef inline _ensure_code(arr):
             return arr
         arr.convert_to_units("code_length")
     return arr
+
+ctypedef fused any_float:
```
neutrinoceros, line 62: While we're at it, any reason not to include float16 and float128 ?

matthewturk: Only that I don't think we use them anywhere?  Maybe in a future revision, if we end up having a use case that we can trust with them.

neutrinoceros: Sounds fair. Thanks !

cphyc: Is there a use case for also applying the fuse type to e.g. `pixelize_sph_kernel_arbitrary_grid` et al.?

matthewturk: Yes -- I'll add some of these once the tests pass and I know that I'm on the right track.  :)

neutrinoceros:
```
ValueError: Buffer dtype mismatch, expected 'float32_t' but got 'double'
```

matthewturk:
I'm still working on getting the fused types working, but almost there.

matthewturk:
[preallocate_console_log.txt](https://github.com/yt-project/yt/files/6640751/preallocate_console_log.txt)

I've attached here the console log from the most recent run so that, in case the testing failures cycle off Jenkins, we can take a look at all of the particular failure locations.

matthewturk:
I'm adding a cast to the process grid, and then will work on the remainder.

matthewturk:
current log
[preallocation_20210618.txt](https://github.com/yt-project/yt/files/6678607/preallocation_20210618.txt)

chrishavlin:
attempted a merge using the github interface! let's see if it worked :)

yt/frontends/sph/io.py
```diff
@@ -14,6 +14,8 @@ class IOHandlerSPH(BaseParticleIOHandler):
     This exists to handle particles with smoothing lengths, which require us
     to read in smoothing lengths along with the the particle coordinates to
     determine particle extents.
+
+    At present this is non-functional.
```
chrishavlin, line 18: What'd you mean by "non-functional" here? Lots of frontends inherit from this class.

yt/utilities/lib/fp_utils.pxd
chrishavlin: Is this still an issue?

chrishavlin (approved):
Assuming tests pass, I think this is good to go? There was one outstanding conversation about [fp_utils.pxd](https://github.com/yt-project/yt/pull/2416/files#diff-9fe1ade6c0229c02fae5467f5579f89a0f4bf4654d95baf7a2438c3164d6924f) that might still need addressing but I'm not sure.

chrishavlin:
oh, spoke too soon :)

chrishavlin:
Oh, still kind of a merge issue, I'll see if I can fix it.

chrishavlin:
ok, those mypy errors are new -- I'll leave this for now.

neutrinoceros:
I fixed most of the typing problems. The remaining one may be more delicate as it could affect many frontends at once. I'll avoid touching that while CI is unstable

yt/utilities/io_handler.py
```diff
@@ -254,46 +196,29 @@ def _read_particle_selection(
...
+                data[field_f].append(vals)
+
+        rv: Dict[Tuple[str, str], np.ndarray] = {}  # the return dictionary
+        fields = list(data.keys())
+        for field_f in fields:
+            # We need to ensure the arrays have the right shape if there are no
+            # particles in them.
+            total = sum(_.size for _ in data[field_f])
+            if total > 0:
+                rv[field_f] = np.concatenate(data.pop(field_f), axis=0)
```
chrishavlin, line 214: So the following should fix the failing tests (I only tested the gadget answer tests locally, but I suspect it'll fix the other failures too). The issue was the when we were pre-allocating we were setting the empty array to `dtype="float64"` and now we're not... and I guess the returns from all the various read functions for the individual frontends aren't totally consistent in what they return. So I'm inclined to just put the `astype` here, but are there downsides to that? Maybe there's a better way? 

```suggestion
                rv[field_f] = np.concatenate(data.pop(field_f), axis=0).astype("float64")
```

Oh and with numpy >= 1.20.0, concatenate actually accepts a dtype argument. It has a `casting` argument that could be useful: https://numpy.org/doc/stable/reference/generated/numpy.concatenate.html#numpy.concatenate But I didn't think it was worth adding a version dependent behavior here... but could?

neutrinoceros: + 1 for this patch

> But I didn't think it was worth adding a version dependent behavior here... but could?

FWIW, a _lot_ of functions in numpy acquired the `dtype` argument over the years, but the process is extremely gradual, so it's really hard to keep track of (and proabably not worth the trouble). In the future ™️, I envision that we could go through all usages of `.astype` manually and use `dtype` instead, but AFAIK there is'nt anything equivalent to e.g. pyupgrade that can be used to automate this process

chrishavlin: I'll give @matthewturk a chance to chime in on this before committing the change... I also considered doing the cast within the loop where we append to the `data` list, but that would necessitate dealing with empty chunks explicitly rather relying on concatenate to squash them.

matthewturk: So I think my only concern with this is in the possible case of having an int array.  I do not actually remember if we support that in *general* for particle fields.  If we do not, then an astype should be fine.

chrishavlin: that's a good question... previously we were allocating all the arrays initially with `rv[field] = np.empty(shape, dtype="float64")` and then filling. That would have implicitly converted any int to float64, even if a reader returned an int array, ya? 

I think more broadly we could use a refactor to better handle particle dtypes. Or at least document how we handle them. The various `_read_` implementations for each frontend frequently include casting to float64 and it's never been clear to me when or exactly why it's needed. But I'm fairly certain that casting here is on par with what this function did previously.

neutrinoceros: I think the dtype constraint really comes from yt's core, where many Cython functions expect float64 and nothing else. Though that's just an educated guess, I never tried looking deep into this.

chrishavlin: I'm going to go ahead and commit this suggestion...

neutrinoceros:
pre-commit.ci autofix

yt/utilities/io_handler.py
```diff
@@ -254,46 +196,31 @@ def _read_particle_selection(
...
+                data[field_f].append(vals)
+
+        rv: Dict[Tuple[str, str], np.ndarray] = {}  # the return dictionary
+        fields = list(data.keys())
+        for field_f in fields:
+            # We need to ensure the arrays have the right shape if there are no
+            # particles in them.
+            total = sum(_.size for _ in data[field_f])
+            if total > 0:
+                rv[field_f] = np.concatenate(data.pop(field_f), axis=0).astype(
```
neutrinoceros, line 214: @chrishavlin maybe you want to refactor this so formatting doesn’t ruin readability?

chrishavlin: yup! spent a while trying to understand if applying the type conversion to individual chunks before concatenating had any performance differences... to which I can say probably not but maybe??? and so I'm going to keep it here. So will clean it up to be more readable.

neutrinoceros:
Nice to see tests passing again. What remains to be done before this can go in ?

chrishavlin:
I think it might be ready? @matthewturk , what do you think?

I do want to understand some dtype issues some more, but the current state is not really any different than the pre-allocation approach in that regard, so I'm happy with this as is. Since this turned into somewhat of a joint effort, I'll hold off on official approval so I think it'd be good to get another review (or a re-review from @cphyc ).

neutrinoceros:
I think I can also review it. My only contribution was getting mypy happy, which didn't require a lot of in depth understanding from me :-)
I would also like to know if we can close #2383 by merging this ?

neutrinoceros (approved):
LGTM. We should probably resolve the few hanging threads though

yt/frontends/gadget/io.py
```diff
@@ -14,7 +14,12 @@
 
 class IOHandlerGadgetHDF5(IOHandlerSPH):
     _dataset_type = "gadget_hdf5"
-    _vector_fields = ("Coordinates", "Velocity", "Velocities", "MagneticField")
+    _vector_fields = {
```
matthewturk, line 17: I'm going to make this comment here, on a PR I originally authored, just to note that this is one place that a *proper* DSL, or even using the DSL we have *already defined* for Gadget-like fields would be extremely useful and helpful.

yt/frontends/sph/io.py
matthewturk: I *think* what I meant was that it was a base class that wasn't used?  Or maybe, it was meaning that the smoothing lengths were not globally implemented?

jzuhone (approved):

yt/frontends/sph/io.py
chrishavlin: ah, gotchya. Cause this base class is not actually adding a global smoothing length implementation -- all the classes inheriting from this implement their own. It's still a useful organizational feature as is though! And I've done some work that actually adds stuff to this base class (don't remember exactly what work -- either one of my open PRs or stuff I haven't pushed). But in any case, I'm fine with this comment :)

yt/utilities/lib/fp_utils.pxd
chrishavlin: I think this one is OK to resolve since tests are passing now, ya?

@neutrinoceros ?

yt/frontends/gadget/io.py
chrishavlin: agreed!

yt/utilities/lib/fp_utils.pxd
neutrinoceros: sure. And unless I'm missing something, seems to me that's the last hanging thread, so maybe you can merge now ?

chrishavlin: Agreed! I think we're ready!

PR #2424, opened 2020-01-22
Enzo-p Frontend Bugfix to Load Single Block Datasets

mabruzzo:
## PR Summary
Enzo-p grids are split across one or more "blocks". Previously, the frontend supported loading a dataset with a grid split across multiple blocks. This pull request fixed a bug that had prevented the loading of a grid stored in a single block.

## PR Checklist

I added an additional case to the unit test for the `get_block_info` function. However, I have not added a test demonstrating that the front-end can now load in a single-block dataset because it would involve a new answer test dataset. I'm not sure how necessary that is.

### DIFF
--- a/yt/frontends/enzo_p/data_structures.py
+++ b/yt/frontends/enzo_p/data_structures.py
@@ -207,12 +207,18 @@ class EnzoPHierarchy(GridIndex):
                 line = buff[bnl:nnl]
                 block_name, block_file = line.split()
 
-                # The B__ block is a negative refinement level
-                # that cannot be used.
                 if "0" not in block_name and "1" not in block_name:
-                    level = -1
-                    left = self.ds.domain_left_edge.d
-                    right = self.ds.domain_right_edge.d
+                    if self.num_grids == 1:
+                        # Because the B__ block is the only block in the grid,
+                        # it has a refinement level of 0
+                        level = 0
+                        rbid = 0
+                    else:
+                        # The B__ block is a negative refinement level
+                        # that cannot be used.
+                        level = -1
+                    left = np.zeros(self.ds.dimensionality)
+                    right = np.ones(self.ds.dimensionality)
                 else:
                     level, left, right = get_block_info(block_name)
                     rbindex = get_root_block_id(block_name)
--- a/yt/frontends/enzo_p/data_structures.py
+++ b/yt/frontends/enzo_p/data_structures.py
@@ -207,23 +207,12 @@ class EnzoPHierarchy(GridIndex):
                 line = buff[bnl:nnl]
                 block_name, block_file = line.split()
 
-                if "0" not in block_name and "1" not in block_name:
-                    if self.num_grids == 1:
-                        # Because the B__ block is the only block in the grid,
-                        # it has a refinement level of 0
-                        level = 0
-                        rbid = 0
-                    else:
-                        # The B__ block is a negative refinement level
-                        # that cannot be used.
-                        level = -1
-                    left = np.zeros(self.ds.dimensionality)
-                    right = np.ones(self.ds.dimensionality)
-                else:
-                    level, left, right = get_block_info(block_name)
-                    rbindex = get_root_block_id(block_name)
-                    rbid = rbindex[0] * rbdim[1:].prod() + \
-                      rbindex[1] * rbdim[2:].prod() + rbindex[2]
+                # Handling of the B, B_, and B__ blocks is consistent with
+                # other unrefined blocks
+                level, left, right = get_block_info(block_name)
+                rbindex = get_root_block_id(block_name)
+                rbid = rbindex[0] * rbdim[1:].prod() + \
+                  rbindex[1] * rbdim[2:].prod() + rbindex[2]
 
                 # There are also blocks at lower level than the
                 # real root blocks. These can be ignored.
--- a/yt/frontends/enzo_p/tests/test_misc.py
+++ b/yt/frontends/enzo_p/tests/test_misc.py
@@ -51,6 +51,11 @@ def test_get_block_info():
         assert left == float(n) / max_n
         assert right == float(n+1) / max_n
 
+    level, left, right = get_block_info("B__")
+    assert level == 0
+    assert (left == np.zeros(3)).all()
+    assert (right == np.ones(3)).all()
+
 def test_root_blocks():
     rs = np.random.RandomState(45652)
     for i in range(2, 6):
--- a/yt/frontends/enzo_p/tests/test_misc.py
+++ b/yt/frontends/enzo_p/tests/test_misc.py
@@ -19,16 +19,23 @@ from yt.frontends.enzo_p.misc import \
     get_block_info, \
     get_root_blocks, \
     get_root_block_id, \
+    is_parent, \
     nested_dict_get
 
 def get_random_block_string(max_n=64, random_state=None, level=None):
+    if max_n == 1:
+        assert level is None or level == 0
+        return 0, 0, 'B'
+    elif max_n < 1:
+        raise ValueError('max_n must be a positive integer')
+
     if random_state is None:
         random_state = np.random.RandomState()
 
     max_l = int(np.log2(max_n))
     form = "%0" + str(max_l) + "d"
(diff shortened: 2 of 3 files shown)


mabruzzo:
So I realized that I didn't supply any proof that I actually fixed the problem. I have attached the outputs of a simple single-block simulation to this commit:
[single-block.tar.gz](https://github.com/yt-project/yt/files/4109180/single-block.tar.gz)
This simulation simply initializes a 4x4x4 grid stored on a single block with a density field that is set to 1.0 everywhere. The field undergoes no evolution (but the simulation cycle is incremented by 1)

To replicate the error that this pull-request addresses, run the following in the python REPL after untarring the files:
`>>> import yt; ds = yt.load("single-block/single-block-1/single-block-1.block_list"); ad = ds.all_data(); print(ad["density"].mean()); print(ad["density"].var())`
This command will fail with the current version of yt, but will succeed with the changes from the pull request.

Let me know if I should be adding this as an answer test.

yt/frontends/enzo_p/data_structures.py
```diff
@@ -207,12 +207,18 @@ def _parse_index(self):
...
+                        # Because the B__ block is the only block in the grid,
+                        # it has a refinement level of 0
+                        level = 0
+                        rbid = 0
+                    else:
+                        # The B__ block is a negative refinement level
+                        # that cannot be used.
+                        level = -1
+                    left = np.zeros(self.ds.dimensionality)
+                    right = np.ones(self.ds.dimensionality)
```
munkm: Is there a reason that you're assuming the left edge and right edge arrays here and overriding what you read in from your file? Your fix with the updated logic works if you continue to use `self.ds.domain_left_edge.d` and `self.ds.domain_right_edge.d`.

yt/frontends/enzo_p/data_structures.py
```diff
@@ -207,12 +207,18 @@ def _parse_index(self):
...
                 block_name, block_file = line.split()
 
-                # The B__ block is a negative refinement level
-                # that cannot be used.
                 if "0" not in block_name and "1" not in block_name:
-                    level = -1
-                    left = self.ds.domain_left_edge.d
-                    right = self.ds.domain_right_edge.d
+                    if self.num_grids == 1:
+                        # Because the B__ block is the only block in the grid,
```
munkm: Thanks for the descriptive inline comments here. These are helpful.

yt/frontends/enzo_p/data_structures.py
mabruzzo: Right, this a subtle change that my toy test don't address. The use of `self.ds.domain_left_edge.d` and `self.ds.domain_right_edge.d` leads to incorrect grid edges. I'll provide a modified example showing this in a following comment (it requires a slightly modified simulation).

Within the for-loop, `left` and `right` effectively have units of "fraction of the domain". After the for-loop, the values of `left` and `right` are re-scaled to have units of `"code_length"` (see lines 276 to 281). Since the highlighted code is handling a single block extending over the full simulation domain, `left=[0,0,0]` and `right=[1,1,1]`. If we assigned values to `left` and `right` based on the results of a call to `get_block_info` (this is done all other blocks on line 223), the values of  `left` and `right` would be the same. If you think it would be more clear, I could replace the highlighted lines with: `left, right = get_block_info(block_name)[1:]`

mabruzzo:
A slightly modified simulation to help justify why we need the change on lines 220-221 of yt/frontends/enzo_p/data_structures.py can be found here: [single-block-new.tar.gz](https://github.com/yt-project/yt/files/4123488/single-block-new.tar.gz)

This is identical to the simulation I provided in the earlier comment (it's a 4x4x4 grid stored on a single block with a density field that is set to 1.0 everywhere and the simulation cycle get's incremented by 1). However, the domain has been modified. The lower edge of the domain is `[-3.,-0.5,0.5]` while the upper edge is  `[4.,2.,1.]`.

The following code block will print `False` twice if `self.ds.domain_left_edge.d` and `self.ds.domain_right_edge.d` are used. It will print `True` twice the code included in this pull request is used.
```python
import yt
ds = yt.load('single-block-new/single-block-1/single-block-1.block_list')
ind = ds.index
print((ind.grid_left_edge[0]  == ds.arr([-3.,-0.5,0.5], 'code_length')).all())
print((ind.grid_right_edge[0] == ds.arr([ 4., 2., 1. ], 'code_length')).all())
```

yt/frontends/enzo_p/data_structures.py
matthewturk: I think I follow this, but I am not sure that memorializing the discussion here will be enough to prevent future confusion.  Could you convert this into a short comment?

yt/frontends/enzo_p/tests/test_misc.py
```diff
@@ -51,6 +51,11 @@ def test_get_block_info():
         assert left == float(n) / max_n
         assert right == float(n+1) / max_n
 
+    level, left, right = get_block_info("B__")
+    assert level == 0
+    assert (left == np.zeros(3)).all()
```
matthewturk: You don't have to change this, but isn't this the same as `(left == 0.0).all()` and likewise for `1.0` below?

yt/frontends/enzo_p/data_structures.py
munkm: +1 to adding an inline comment to the code. I think that would be a helpful addition.

mabruzzo:
I'll be happy to add in comments explaining the change. I'm going to hold off on doing that for a few days before doing that. I've been communicating with the lead developer about the logic governing how Enzo-P writes blocks to files to see if it's possible to simplify this logic in the frontend (which might render special handling of the single root block case mute).

yt/frontends/enzo_p/tests/test_misc.py
mabruzzo: Yes, this is the same. I addressed it in the new commit.

yt/frontends/enzo_p/data_structures.py
mabruzzo: In the new commit, this should hopefully be unnecessary. Blocks named `B`, `B_`, and `B__` are no longer handled as a special case

mabruzzo:
Sorry for the delay. This pull request is now ready for final review. Sorry for making extra work for you, but I think these changes provide a much better solution.

I made some additional changes to simplify the logic in the ``_parse_index`` instance method of ``EnzoPHierarchy`` and fixed bugs preventing the loading of datasets which have a single root block and include AMR (see  [single-block-amr.tar.gz](https://github.com/yt-project/yt/files/4188059/single-block-amr.tar.gz) for an example dataset - this wasn't addressed by my earlier commit). I also solved a bug in the ``is_parent`` helper function which was creating unnecessary, additional work for loading AMR datasets and had been an obstacle to loading 1D datasets. 

Additionally, I added a comment to provide more detail about the behavior of ``get_block_info``. I had originally written up a summary describing the naming convention for all Enzo-P blocks, but it seemed a little too long to include here. I still have it, if you think it would be worth including.

Finally, I made some changes to the unit tests.
- I improved testing coverage in ``test_get_block_info`` to test ``get_block_info`` for 1D, 2D, and 3D names of unrefined blocks that extend over the entire root grid (instead of just the 3D case)
- I improved testing coverage in ``test_root_blocks`` to consider names of blocks from root grids containing only 1 or 2 blocks (this required a minor change to ``get_random_block_string`` to improve handling in an edge case)
- I added ``test_is_parent`` (and it's helper function ``flip_random_block_bit``). This is a unit test for ``is_parent``, which explicitly tests against cases that ``is_parent`` returned incorrect results.

Please let me know if I should be making any other changes. Thanks.

munkm:
I'll be able to take a look at this a little bit later! Thank you for being so clear about your changes. 🙂 

We just merged a PR that fixed some issues that were popping up on appveyor due to a new release from numpy. Would you mind merging the current version of master into your development branch just so we can make sure everything passes?

matthewturk:
This looks like it might be good to go -- @munkm ?

munkm:
Thanks so much for your contribution @mabruzzo! It was nice working with you on this PR. 🙂

PR #2166, opened 2019-03-04
[WIP yt 3] Octree ghost zones

cphyc:
## PR Summary

This PR adds support for ghost zones (or a similar interface).

## PR Checklist

- [x] Code passes flake8 checker
- [x] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Face-centred data
- [x] Gradient, vorticity, divergence support

PR #2425, opened 2020-01-23
Octree ghost zones — reloaded

cphyc:
## PR Summary

This PR adds support for ghost zones (or a similar interface). This is another attempt than #2166 based on the OctVisitors machinery.

## The good
```python
import numpy as np
import yt

ad = ds.all_data()
center = ds.arr(ad.argmax('density'))
sp = ds.sphere(center, (100, 'kpc'))
ds.add_gradient_fields(('gas', 'density'))

center = np.array([.1, .5, .5])
w = np.array([.05, 1, 1])
reg = ds.box(center-w/2, center+w/2)
p = yt.ProjectionPlot(ds, 'x', ['density'] + [f'density_gradient_{k}' for k in 'xyz'], center=center, data_source=reg, weight_field='density')
p.set_axes_unit('Mpc')

for k in 'xyz':
    p.set_log(f'density_gradient_{k}', True, linthresh=1e-55)
    p.set_zlim(f'density_gradient_{k}', -1e-53, 1e-53)
    p.set_cmap(f'density_gradient_{k}', 'bwr')
p.save('frames/the_good/')
```

info_00080_Projection_x_density_density
info_00080_Projection_x_density_gradient_x_density
info_00080_Projection_x_density_gradient_y_density
info_00080_Projection_x_density_gradient_z_density

## The bad
```python

ad = ds.all_data()
center = ds.arr(ad.argmax('density'))
sp = ds.sphere(center, (100, 'kpc'))
ds.add_gradient_fields(('gas', 'density'))

p = yt.SlicePlot(ds, 'x', ['density'] + [f'density_gradient_{k}' for k in 'xyz'], width=sp.radius, center=center, data_source=sp)
p.set_axes_unit('kpc')

dens = p.frb['density']


p.annotate_clear()

for k in 'xyz':
    p.set_log(f'density_gradient_{k}', True, linthresh=1e-48)
    p.set_zlim(f'density_gradient_{k}', -1e-45, 1e-45)
    p.set_cmap(f'density_gradient_{k}', 'bwr')

p.annotate_domain_edges()
p.save('frames/the_bad/')
```
info_00080_Slice_x_density
info_00080_Slice_x_density_gradient_x
info_00080_Slice_x_density_gradient_y
info_00080_Slice_x_density_gradient_z


matthewturk:
I dig the fstrings in your example.

cphyc:
Current version:

## The good
```python
import numpy as np
import yt

ad = ds.all_data()
center = ds.arr(ad.argmax('density'))
sp = ds.sphere(center, (100, 'kpc'))
ds.add_gradient_fields(('gas', 'density'))

center = np.array([.1, .5, .5])
w = np.array([.05, 1, 1])
reg = ds.box(center-w/2, center+w/2)
p = yt.ProjectionPlot(ds, 'x', ['density'] + [f'density_gradient_{k}' for k in 'xyz'], center=center, data_source=reg, weight_field='density')
p.set_axes_unit('Mpc')

for k in 'xyz':
    p.set_log(f'density_gradient_{k}', True, linthresh=1e-55)
    p.set_zlim(f'density_gradient_{k}', -1e-53, 1e-53)
    p.set_cmap(f'density_gradient_{k}', 'bwr')
p.save('frames/the_good/')
```
info_00080_Projection_x_density_density
info_00080_Projection_x_density_gradient_y_density
info_00080_Projection_x_density_gradient_x_density
info_00080_Projection_x_density_gradient_z_density


## The bad
```python

ad = ds.all_data()
center = ds.arr(ad.argmax('density'))
sp = ds.sphere(center, (100, 'kpc'))
ds.add_gradient_fields(('gas', 'density'))

p = yt.SlicePlot(ds, 'x', ['density'] + [f'density_gradient_{k}' for k in 'xyz'], width=sp.radius, center=center, data_source=sp)
p.set_axes_unit('kpc')

dens = p.frb['density']


p.annotate_clear()

for k in 'xyz':
    p.set_log(f'density_gradient_{k}', True, linthresh=1e-48)
    p.set_zlim(f'density_gradient_{k}', -1e-45, 1e-45)
    p.set_cmap(f'density_gradient_{k}', 'bwr')

p.annotate_domain_edges()
p.save('frames/the_bad/')
```
info_00080_Slice_x_density_gradient_y
info_00080_Slice_x_density
info_00080_Slice_x_density_gradient_z
info_00080_Slice_x_density_gradient_x

yt/fields/fluid_fields.py
```diff
@@ -195,12 +195,29 @@ def grad_func(axi, ax):
         slice_3dl = slice_3d[:axi] + (sl_left,) + slice_3d[axi+1:]
         slice_3dr = slice_3d[:axi] + (sl_right,) + slice_3d[axi+1:]
         def func(field, data):
-            ds = div_fac * data[ftype, "d%s" % ax]
-            f  = data[grad_field][slice_3dr]/ds[slice_3d]
-            f -= data[grad_field][slice_3dl]/ds[slice_3d]
+            block_reorder = getattr(data, '_block_reorder', 'C')
+            if block_reorder == 'F':
```
cphyc: This causes many tests to fail. It is however required for RAMSES datasets (otherwise the gradients are in the wrong direction).
The issue is that
1. I do not want to add some RAMSES-specific switch here
2. I do not know why only RAMSES requires this reordering among Fortran-ordered datasets.

yt/data_objects/octree_subset.py
```diff
@@ -36,6 +36,7 @@ class OctreeSubset(YTSelectionContainer):
     _domain_offset = 0
     _cell_count = -1
     _block_reorder = None
+    _gradient_swap_axes = False  # Set to True if one should swap the axes when computing gradient
```
matthewturk: does the need for this get explained anywhere?

yt/fields/fluid_fields.py
```diff
@@ -195,8 +195,8 @@ def grad_func(axi, ax):
         slice_3dl = slice_3d[:axi] + (sl_left,) + slice_3d[axi+1:]
         slice_3dr = slice_3d[:axi] + (sl_right,) + slice_3d[axi+1:]
         def func(field, data):
-            block_reorder = getattr(data, '_block_reorder', 'C')
-            if block_reorder == 'F':
+            reorder = getattr(data, '_gradient_swap_axes', False)
```
matthewturk: oh, never mind, I see now -- this is an additional check for what we were already doing.

setup.py
```diff
@@ -66,6 +66,11 @@
                "yt/utilities/lib/tsearch.c"],
               include_dirs=["yt/utilities/lib"],
               libraries=std_libs),
+    Extension("yt.geometry.ramses_oct_container",
```
matthewturk: Maybe we should put this under the ramses frontend?  I don't know.

yt/fields/fluid_fields.py
```diff
@@ -195,12 +195,29 @@ def grad_func(axi, ax):
...
+
+            okl = np.isfinite(vl)
+            okrl = okl & np.isfinite(vr)
+
+            f = np.where(okrl, (vr - vl) / 2,
+                np.where(okl, vc - vl, vr - vc)) / ds[slice_3d]
+
             new_field = np.zeros_like(data[grad_field], dtype=np.float64)
-            new_field = data.ds.arr(new_field, f.units)
+            new_field = data.ds.arr(new_field, vr.units / ds.units)
```
matthewturk: Any chance you'd consider changing `ds` here to something else, that isn't overloaded elsewhere?  I had to check was type `ds` was since we usually use it for `Dataset`.

yt/geometry/oct_container.pyx
```diff
@@ -217,7 +217,7 @@ cdef class OctreeContainer:
             ind32[i] = ind[i]
         self.get_root(ind32, &next)
         # We want to stop recursing when there's nowhere else to go
-        while next != NULL and level <= max_level:
+        while next != NULL and level < max_level:
```
matthewturk: this seems like a substantive change -- is it fixing an error?

yt/fields/fluid_fields.py
cphyc: I am actually still confused as to why this is required to start with... any guess?

setup.py
cphyc: This could actually be included in the octree_container as it is not Ramses specific anymore.

yt/geometry/oct_container.pyx
cphyc: Imagine that the `get` function is called with `max_level = 0`, then we'd expect one and only one iteration to be done (as `level` starts with a value of `-1`). Prior to the change, up to two iterations would be performed, eventually returning an oct at `max_level + 1`.

This never triggered an error as the `max_level` optional argument was not used anywhere.

yt/fields/fluid_fields.py
matthewturk: There are some strange corner cases where we get F/C ordering in these that isn't consistent, because of mixed ordering among the axes.

cphyc: I corrected this in 10cbdec, so this should be good now.

cphyc:
@yt-fido test this please

cphyc:
This should be ready to go!

matthewturk:
I suspect it will be fine to include it now.  I will work on getting answer tests in #2437 working for the RAMSES datasets before merging this into that PR.

brittonsmith:
I think as long as answer tests are run at some point, it's ok with me. If that happens as part of the yt-4.0 PR, that's fine, too.

yt/fields/fluid_fields.py
```diff
@@ -195,12 +195,29 @@ def grad_func(axi, ax):
         slice_3dl = slice_3d[:axi] + (sl_left,) + slice_3d[axi+1:]
         slice_3dr = slice_3d[:axi] + (sl_right,) + slice_3d[axi+1:]
         def func(field, data):
-            ds = div_fac * data[ftype, "d%s" % ax]
```
Xarthisius: What happened to `div_fac`? I don't see it in the code below. Is it no longer necessary?

yt/frontends/ramses/data_structures.py
```diff
@@ -197,13 +212,101 @@ def fill(self, fd, fields, selector, file_handler):
...
+            new_fwidth[:, :, :] = fwidth[:, 0:1, :]
+            fwidth = new_fwidth.reshape(-1, 3)
+        return fwidth
+
+    @property
+    def fcoords(self):
+        fcoords = super(RAMSESDomainSubset, self).fcoords
+        num_ghost_zones = self._num_ghost_zones
+        if num_ghost_zones == 0:
+            return fcoords
```
Xarthisius: `fcoords` AFAICT is only used if `self._num_ghost_zones == 0`. Wouldn't it make sense to move it inside the if clause?

```
if self._num_ghost_zones == 0:
    return super(RAMSESDomainSubset, self).fcoords
```

I'm not sure how costly it is and if it really matters...

cphyc:
Since @brittonsmith approved the PR, I refactored parts of it to reduce code repetition and make it more readable. Sorry for updating a PR post-review!

cphyc:
For information, I just rebased the branch on yt-4.0. @matthewturk, you mentioned #2437 (in https://github.com/yt-project/yt/pull/2425#issuecomment-582912787), is it still a blocker to merge this PR?

munkm:
We're temporarily blocking all merges to the yt-4.0 branch until yt-project/yt#2437 gets merged. Is that what you're asking?

cphyc:
Yes, thanks for the reply @munkm ;)

munkm:
I'll do my best to review it today! I can't think of anything at the moment. 🙂

yt/geometry/oct_visitors.pyx
```diff
@@ -336,3 +336,172 @@ cdef class MortonIndexOcts(OctVisitor):
...
+
+        # Loop over cells in and directly around oct
+        for i in range(-1, 3):
+            ishift[0] = i
+            for j in range(-1, 3):
+                ishift[1] = j
+                for k in range(-1, 3):
+                    ishift[2] = k
+                    self.set_neighbour_info(o, ishift)
+
```
cphyc, line 441: This piece of code is very much optimizable. Here are possible routes to achieve much better performance:
1. Work oct by oct, which would reduce the number of neighbor lookup from 4³=64 to 3³=27,
2. Use faster neighbor lookup method(s). For now, all searches are started from the root mesh down to leaf nodes, but we could instead go up the tree from the central oct then down to find all neighbors (see e.g. https://geidav.wordpress.com/2017/12/02/advanced-octrees-4-finding-neighbor-nodes/). 
3. Pre-compute the face-neighbors of *all* octs.

Note that for the last point, algorithms exist that generate the neighbors of all octs in O(1) time (https://link.springer.com/article/10.1007/s13319-015-0060-9) during the octree construction.
Another possible solution would be to keep a unordered hash map of all the octs indexed by their (3-integers) position. With such structure, finding a neighbor takes O(1) time. This could even come as a replacement of the current pointer-based octree structure.

cphyc:
@yt-fido test this please

cphyc:
So the new tests answer tests are hidden here https://tests.yt-project.org/job/yt_py38_git/423/testReport/yt.frontends.ramses.tests (they pass).

Xarthisius:
> So the new tests answer tests are hidden here https://tests.yt-project.org/job/yt_py38_git/423/testReport/yt.frontends.ramses.tests (they pass).

The test that you've added won't work as intended. There's a few things that need to be fulfilled for a method to work as an answer test:

1. It has to be decorated with `requires_ds` (not `requires_file`)
1. It has to yield test classes based on `AnswerTestingTest` e.g. `FieldValuesTest`
1. It has to be explicitly enumerated in `tests/tests.yaml` as `yt/frontends/ramses/tests/test_outputs.py:test_field_accession`
1. In case of name collision (which would happen here with `test_output_00800` should you yield some additional FieldValuesTest) a unique suffix needs to be added to the test name.

Ultimately it would look roughly like this:

```python
def test_field_accession():
    ds = data_dir_load(output_00080)
    fields = [ 
        ('gas', 'density'),  # basic ones
        ('gas' ,'pressure'),
        ('gas', 'pressure_gradient_magnitude'), # requires ghost zones
    )   
    dso = ( 
        None,
        ("sphere", ([.1]*3, (0.01, 'unitary'))),
        ("sphere", ([.5]*3, (0.05, 'unitary'))),
        ("box", ([.1]*3, [.2]*3)),
    )   
    for dobj_name in dso:
        for field in fields:
            test = FieldValuesTest(ds, field, dobj_name)
            test.suffix = "accession"
            test_field_accession.__name__ = test.description + test.suffix
            yield test
```
Having said all that, I think most of the things from this tests are already done in the `test_output_00800`. It tests density and pressure gradient magnitude, it tests entire domain and one sphere. If you would add `('gas' ,'pressure')` to L20, and additional objects in L28 it would do the same (and more) for you. For the sake of efficiency though I'd evaluate if `test_output_00800` is sufficient to test your work even without any additions.

yt/geometry/oct_container.pyx
```diff
@@ -728,17 +728,196 @@ cdef class OctreeContainer:
...
+                else:
+                    dest[i + offset] = source[file_inds[i], cell_inds[i]]
+
+    def fill_index(self, SelectorObject selector = AlwaysSelector(None)):
+        """Get the on-file index of each cell"""
+        cdef StoreIndex visitor
+
+        cdef np.int64_t[:, :, :, :] cell_inds
+
+        cell_inds = np.full((self.nocts, 2, 2, 2), -1, dtype=np.int64)
```
neutrinoceros, line 748: not used to Cython, so I may be plain wrong, but I read the dtype argument here as completely redundant with the array declaration.

yt/geometry/oct_container.pyx
```diff
@@ -728,17 +728,196 @@ cdef class OctreeContainer:
...
+        num_octs : int, optional
+            The number of octs to read in
+        domain_id : int, optional
+            The domain to perform the selection over
+
+        Returns
+        -------
+        oct_inds : int64 ndarray (nocts*8, )
+            The on-domain index of the octs containing each cell
+        cell_inds : uint8 array (nocts*8, )
```
neutrinoceros, line 774:
```suggestion
        cell_inds : uint8 ndarray (nocts*8, )
```

yt/geometry/oct_container.pyx
```diff
@@ -728,17 +728,196 @@ cdef class OctreeContainer:
...
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    @cython.cdivision(True)
+    def fill_level_with_domain(
+                   self, int level,
+                   np.uint8_t[:] levels,
+                   np.uint8_t[:] cell_inds,
+                   np.int64_t[:] file_inds,
+                   np.int32_t[:] domains,
+                   dict dest_fields, dict source_fields,
```
neutrinoceros, line 809:
```suggestion
                   dict dest_fields,
                   dict source_fields,
```

yt/geometry/oct_container.pyx
```diff
@@ -728,17 +728,196 @@ cdef class OctreeContainer:
...
+        cdef np.ndarray[np.float64_t, ndim=2] source
+        cdef np.ndarray[np.float64_t, ndim=1] dest
+        cdef int i, count
+
+        for key in dest_fields:
+            dest = dest_fields[key]
+            source = source_fields[key]
+            count = 0
+            for i in range(levels.shape[0]):
+                if levels[i] != level or domains[i] != domain: continue
```
neutrinoceros, line 828: I don't know if this is valid in Cython but I'll suggest it anyway
```suggestion
            for i, (lev, dom) in enumerate(zip(levels, domains)):
                if lev != level or dom != domain: continue
```

yt/geometry/oct_container.pyx
```diff
@@ -728,17 +728,196 @@ cdef class OctreeContainer:
...
+        return count
+
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    @cython.cdivision(True)
+    def file_index_octs_with_ghost_zones(
+            self, SelectorObject selector, int domain_id,
+            int num_cells = -1):
+        """Similar as file_index_octs, but return as well the level, cell index,
+        file index and domain of the neighbouring cells.
```
neutrinoceros, line 843:
```suggestion
        """Similar as file_index_octs, but return the level, cell index,
        file index and domain of the neighbouring cells as well.
```

yt/geometry/oct_container.pyx
```diff
@@ -728,17 +728,196 @@ cdef class OctreeContainer:
...
+
+        Arguments
+        ---------
+        selector : SelectorObject
+            The selector object. It is expected to select all cells for a selected oct.
```
neutrinoceros, line 848: Would this make sense ? I'm trying to avoid the repetition of "select"
```suggestion
            The selector object. It is expected to select all cells for a given oct.
```

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -461,3 +461,17 @@ def test_magnetic_field_aliasing():
...
         ad[('gas',field)]
+
+output_00080 = "output_00080/info_00080.txt"
+@requires_file(output_00080)
+def test_field_accession():
+    ds = yt.load(output_00080)
+    fields = [
+        ('gas', 'density'),  # basic ones
+        ('gas' ,'pressure'),
+        ('gas', 'pressure_gradient_magnitude'), # requires ghost zones
```
neutrinoceros, line 472: Wouldn't it be equivalent and more direct to test for `density_gratient_magnitude` instead ?
My understanding is that you need a gradient field to test the ghost zones' effect, but if pressure is a derived field then you're actually testing two things at once (and maybe that's a good thing tbh, I don't have a strong opinion, just curious)

cphyc:
> > So the new tests answer tests are hidden here https://tests.yt-project.org/job/yt_py38_git/423/testReport/yt.frontends.ramses.tests (they pass).
> 
> The test that you've added won't work as intended. There's a few things that need to be fulfilled for a method to work as an answer test:
> 
>     1. It has to be decorated with `requires_ds` (not `requires_file`)
> 
>     2. It has to yield test classes based on `AnswerTestingTest` e.g. `FieldValuesTest`
> 
>     3. It has to be explicitly enumerated in `tests/tests.yaml` as `yt/frontends/ramses/tests/test_outputs.py:test_field_accession`
> 
>     4. In case of name collision (which would happen here with `test_output_00800` should you yield some additional FieldValuesTest) a unique suffix needs to be added to the test name.
> 
> 
> Ultimately it would look roughly like this:
> 
> ```python
> @requires_ds(output_00080)
> def test_field_accession():
>     ds = data_dir_load(output_00080)
>     fields = [ 
>         ('gas', 'density'),  # basic ones
>         ('gas' ,'pressure'),
>         ('gas', 'pressure_gradient_magnitude'), # requires ghost zones
>     )   
>     dso = ( 
>         None,
>         ("sphere", ([.1]*3, (0.01, 'unitary'))),
>         ("sphere", ([.5]*3, (0.05, 'unitary'))),
>         ("box", ([.1]*3, [.2]*3)),
>     )   
>     for dobj_name in dso:
>         for field in fields:
>             test = FieldValuesTest(ds, field, dobj_name)
>             test.suffix = "accession"
>             test_field_accession.__name__ = test.description + test.suffix
>             yield test
> ```
> 
> Having said all that, I think most of the things from this tests are already done in the `test_output_00800`. It tests density and pressure gradient magnitude, it tests entire domain and one sphere. If you would add `('gas' ,'pressure')` to L20, and additional objects in L28 it would do the same (and more) for you. For the sake of efficiency though I'd evaluate if `test_output_00800` is sufficient to test your work even without any additions.

Thanks for the detailed answer. The answer tests should now be included in the already-existing `test_output_00080` answer tests. I also added some other unit tests, tell me if you're ok with them.

yt/fields/fluid_fields.py
```diff
@@ -205,16 +205,25 @@ def grad_func(axi, ax):
...
-            f  = data[grad_field][slice_3dr]/ds[slice_3d]
-            f -= data[grad_field][slice_3dl]/ds[slice_3d]
+            f  = field_data[slice_3dr]/ds[slice_3d]
+            f -= field_data[slice_3dl]/ds[slice_3d]
             new_field = np.zeros_like(data[grad_field], dtype=np.float64)
-            new_field = data.ds.arr(new_field, f.units)
+            new_field = data.ds.arr(new_field, field_data.units / ds.units)
             new_field[slice_3d] = f
+
+            if block_reorder == 'F':
```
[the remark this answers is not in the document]
cphyc: In principle `block_reorder` should either be `F` or `C|None`, so there is a boolean logic behind it. This being said, I would rather keep the `F` and `C` explicit to mimic numpy's internal `order=` argument.

I can probably replace the `block_reorder` argument by `block_order` and set its value to either `C` or `F`.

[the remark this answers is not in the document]
neutrinoceros: > I can probably replace the block_reorder argument by block_order and set its value to either C or F.

That would be awesome, and much better than my original suggestion.

yt/frontends/ramses/tests/test_outputs.py
cphyc: So it probably is slightly better to test `pressure_gradient_magnitude`. The reason for this is that reading "ghost zones" requires reading from file and not from memory, as some cells may be in boundary regions, or refined, etc.
So when testing `pressure_gradient_magnitude`, we're testing the gradient is properly computed _and_ the underlying pressure as well, including in cells that would normally not be read.

Does this make sense?

neutrinoceros: You’re right, it causes less redundancy.

yt/fields/fluid_fields.py
```diff
@@ -205,16 +205,25 @@ def grad_func(axi, ax):
         slice_3dl = slice_3d[:axi] + (sl_left,) + slice_3d[axi+1:]
         slice_3dr = slice_3d[:axi] + (sl_right,) + slice_3d[axi+1:]
         def func(field, data):
+            block_reorder = getattr(data, '_block_reorder', None)
+            if block_reorder == 'F':
+                field_data = data[grad_field].swapaxes(0, 2)
```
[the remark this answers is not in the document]
cphyc: I don't think there is a way around this unfortunately.

yt/geometry/oct_container.pyx
cphyc: Done

yt/geometry/oct_container.pyx
cphyc: Done.

yt/geometry/oct_container.pyx
cphyc: Done.

yt/geometry/oct_container.pyx
cphyc: Done.

yt/fields/geometric_fields.py
```diff
@@ -94,9 +94,10 @@ def _zeros(field, data):
     def _ones(field, data):
         """Returns one for all cells"""
         arr = np.ones(data.ires.shape, dtype="float64")
+        tmp = data.apply_units(arr, field.units)
         if data._spatial:
-            return data._reshape_vals(arr)
-        return data.apply_units(arr, field.units)
+            return data._reshape_vals(tmp)
+        return tmp
```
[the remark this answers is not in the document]
cphyc: Done

yt/fields/fluid_fields.py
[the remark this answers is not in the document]
cphyc: This is now fixed!

yt/fields/fluid_fields.py
```diff
@@ -205,16 +205,25 @@ def grad_func(axi, ax):
...
                 ds *= data[ftype, "r"]
             if ax == "phi":
                 ds *= data[ftype, "r"] * np.sin(data[ftype, "theta"])
-            f  = data[grad_field][slice_3dr]/ds[slice_3d]
-            f -= data[grad_field][slice_3dl]/ds[slice_3d]
+            f  = field_data[slice_3dr]/ds[slice_3d]
+            f -= field_data[slice_3dl]/ds[slice_3d]
             new_field = np.zeros_like(data[grad_field], dtype=np.float64)
-            new_field = data.ds.arr(new_field, f.units)
+            new_field = data.ds.arr(new_field, field_data.units / ds.units)
```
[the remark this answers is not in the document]
cphyc: I changed it from `ds` to `dx`.

yt/data_objects/data_containers.py
```diff
@@ -352,13 +352,11 @@ def _generate_spatial_fluid(self, field, ngz):
...
-                    else:
-                        ind += wogz.select(
-                            self.selector,
-                            gz[field][ngz:-ngz, ngz:-ngz, ngz:-ngz],
-                            rv, ind)
+                    data = gz[field][ngz:-ngz, ngz:-ngz, ngz:-ngz]
+                    ind += wogz.select(
+                        self.selector,
+                        data,
+                        rv, ind)
```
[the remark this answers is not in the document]
cphyc: Done

yt/data_objects/octree_subset.py
```diff
@@ -455,8 +456,8 @@ def __init__(self, ind, block_slice):
     def __getitem__(self, key):
         bs = self.block_slice
         rv = bs.octree_subset[key][:,:,:,self.ind].T
-        if bs.octree_subset._block_reorder:
-            rv = rv.copy(order=bs.octree_subset._block_reorder)
+        if bs.octree_subset._block_order:
+            rv = rv.copy(order=bs.octree_subset._block_order)
```
cphyc, line 460: Would this be better written as
```python
         rv = np.require(
            bs.octree_subset[key][:,:,:,self.ind].T,
            requirements=bs.octree_subset._block_order)
```
which I believe does the same job but avoid unnecessary copies.
            
```

Comment (neutrinoceros) [reply yt/data_objects/octree_subset.py:460]: I agree !
Moreover, I was going to suggest moving the transposition to after the closing parenthesis but I tested it and you version is actually more efficient by a factor 4 for the 100x100 array it tested this on.

**Comment by cphyc** [comment]
@yt-fido test this please.

**Comment by cphyc** [comment]
@yt-fido test this please.

File: yt/data_objects/data_containers.py [hunk 23]
```diff
@@ -364,13 +368,12 @@ def _generate_spatial_fluid(self, field, ngz):
...
                                 dtype="float64"), units)
                         outputs.append(rv)
-                    if gz._type_name == 'octree_subset':
-                        raise NotImplementedError
-                    else:
-                        ind += wogz.select(
-                            self.selector,
-                            gz[field][ngz:-ngz, ngz:-ngz, ngz:-ngz],
-                            rv, ind)
+                    data = gz[field][ngz:-ngz, ngz:-ngz, ngz:-ngz]
```
matthewturk, line 371: I like getting rid of the conditional!  Nice.

yt/frontends/ramses/data_structures.py
```diff
@@ -197,13 +214,85 @@ def fill(self, fd, fields, selector, file_handler):
...
 
+    def _fill_with_ghostzones(self, fd, fields, selector, file_handler, num_ghost_zones):
+        ndim = self.ds.dimensionality
+        ncpu = self.ds.parameters['ncpu']
+        # Here we get a copy of the file, which we skip through and read the
+        # bits we want.
+        oct_handler = self.oct_handler
+        all_fields = [f for ft, f in file_handler.field_list]
+        fields = [f for ft, f in fields]
+        tr = {}
```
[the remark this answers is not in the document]
matthewturk: In fairness to @cphyc we do use `tr` as "to return" in lots of places.  I take the blame for that.

yt/frontends/ramses/data_structures.py
```diff
@@ -197,13 +214,85 @@ def fill(self, fd, fields, selector, file_handler):
...
+
+    @property
+    def fwidth(self):
+        fwidth = super(RAMSESDomainSubset, self).fwidth
+        if self._num_ghost_zones > 0:
+            fwidth = fwidth.reshape(-1, 8, 3)
+            n_oct = fwidth.shape[0]
+            new_fwidth = np.zeros((n_oct, self.nz**3, 3), dtype=fwidth.dtype)
+            new_fwidth[:, :, :] = fwidth[:, 0:1, :]
+            fwidth = new_fwidth.reshape(-1, 3)
```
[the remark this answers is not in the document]
matthewturk: I also did not know this!  And I will use this a lot.

yt/geometry/oct_container.pyx
matthewturk: Unfortunately it is required to be this way -- we're declaring a cython variable for `cell_inds` but it has to receive it from the python function, which needs to know the dtype.

yt/geometry/oct_visitors.pxd
```diff
@@ -136,3 +142,34 @@ cdef inline int cind(int i, int j, int k) nogil:
...
+from oct_container cimport OctreeContainer
+
+cdef class StoreIndex(OctVisitor):
+    cdef np.int64_t[:,:,:,:] cell_inds
+
+# cimport oct_container
+cdef class BaseNeighbourVisitor(OctVisitor):
+    cdef int idim      # 0,1,2 for x,y,z
+    cdef int direction # +1 for +x, -1 for -x
+    cdef np.int8_t[:] neigh_ind
```
matthewturk, line 154: two quick things -- should `neigh_ind` be a 3d view?  and, it's a bit faster if we declare `::1` in them so that it knows they have stride 1.

yt/geometry/oct_visitors.pxd
```diff
@@ -136,3 +142,34 @@ cdef inline int cind(int i, int j, int k) nogil:
...
+    cdef inline np.uint8_t neighbour_rind(self):
+        cdef int d = (1 << self.oref)
+        return (((self.neigh_ind[2]*d)+self.neigh_ind[1])*d+self.neigh_ind[0])
+
+cdef class NeighbourCellIndexVisitor(BaseNeighbourVisitor):
+    cdef np.uint8_t[:] cell_inds
+    cdef np.int64_t[:] domain_inds
+
+cdef class NeighbourCellVisitor(BaseNeighbourVisitor):
+    cdef np.uint8_t[:] levels
```
matthewturk, line 171: same here about stride

yt/geometry/oct_visitors.pyx
```diff
@@ -338,3 +338,172 @@ cdef class MortonIndexOcts(OctVisitor):
...
+        local_oct = True
+
+        # Compute position of neighbouring cell
+        for i in range(3):
+            c = <np.float64_t> (self.pos[i] << self.oref)
+            fcoords[i] = (c + 0.5 + ishift[i]) * dx / self.octree.nn[i]
+            # Assuming periodicity
+            if fcoords[i] < 0:
+                fcoords[i] += 1
+            elif fcoords[i] > 1:
```
matthewturk, line 380: can you make this explicitly a float?  and, shouldn't this be checking if it's greater than (and above, less than) the domain width?

matthewturk: actually i may be wrong on that, but want confirmation.

yt/geometry/oct_visitors.pyx
matthewturk: I think this should be in a comment.  It obviously doesn't need to be implemented right now.

cphyc: Done.

yt/frontends/ramses/data_structures.py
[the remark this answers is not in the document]
cphyc: So I'm torn between changing `tr`  here and break the "convention", or keep it... I'm leaving as is for now, unless someone argues to change it (in which case we should probably grep `tr` and replace it at other places as well).

yt/geometry/oct_container.pxd
```diff
@@ -27,12 +27,6 @@ cdef struct OctKey:
     np.int64_t *indices
     np.int64_t pcount
 
-cdef struct OctInfo:
```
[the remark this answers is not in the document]
cphyc: This is due to a circular dependency issue between `oct_container.p[yx|xd]` and `oct_visitor.p[yx|xd]`, the former being dependent on the latter, so `OctInfo` needs to be in `oct_visitor` instead.

This was not necessary before, because `OctInfo` wasn't used in `oct_visitor` but it is now.

If you know how to solve that more gracefully, I would be glad to implement it!

yt/frontends/ramses/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: I'd be in favour of changing it since I managed to avoid this convention for more than a year of reading/writing for yt, so it really doesn't seem obvious to me, and I'm all about reducing overhead on the dev side :)
In any case I don't want to cause any more hurdle than necessary for this PR, so imo this can stay as is.

yt/geometry/oct_container.pyx
neutrinoceros: fair enough ! I might end up learning Cython just by reviewing this stuff :)

yt/frontends/ramses/data_structures.py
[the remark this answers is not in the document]
cphyc: Done!

yt/geometry/oct_visitors.pyx
cphyc: Done!

PR #2428, opened 2020-01-27
Fixed a bug in test_testing.py

jcoughlin11:
The issue was that requires_backend returned the function objects, but did not actually call the function. As such, when assert_equal was called in test_testing.py, the function object was being compared to the expected correct return result of the function (either None or True). By calling the returned function object, the test now passes. test_testing.py was modified instead of requires_backend (to return the called functions) because other parts of the code use requires_backend and I did not want to break those.

## PR Checklist

### DIFF
--- a/yt/tests/test_testing.py
+++ b/yt/tests/test_testing.py
@@ -25,5 +25,5 @@ def test_requires_backend():
     def plot_b():
         return True
 
-    assert_equal(plot_a(), None)
-    assert_equal(plot_b(), True)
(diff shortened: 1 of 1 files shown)


jcoughlin11:
For some reason the update made to requires_backend on master did not propagate to my pytest_infrastructure branch when I fetched and merged into my master then merge my master into pytest_infrastructure, so I misread things. The removal of the last wrapper layer makes the current version on master correct, rendering this PR obsolete.

Issue #2432, opened 2020-01-29
bug: annotate_timestamp method for ds using "code" unit_system


neutrinoceros:
## Bug report

**Bug summary**
the "code" unit system is correctly being used for x, y, z labels (since #2354) but not for timestamps.

**Code for reproduction**
```python
import yt
ds = yt.load("amrvac/bw_2d0000.dat", unit_system="code")
p = yt.plot_2d(ds, "density")
p.annotate_timestamp()
```

**Actual outcome**
![code_timestamp_Slice_phi_density](https://user-images.githubusercontent.com/14075922/73372404-6fc74480-42b7-11ea-8076-3ac30645a388.png)

**Expected outcome**

the time stamp should use "code time" as a unit.

munkm:
I'm working on a fix for this. Should be ready soon.

munkm:
Resolved via yt-project/yt#2435

PR #2435, opened 2020-01-30
[bugfix] Update annotate_timestamp to work with "code" unit system 

munkm:
## PR Summary

This resolves the bug reported in yt-project/yt#2432 . Now when a dataset loaded using `unit_system=code` and `annotate_timestamp()` is called, the time returned is in `code_time`. 

With the fix applied, the plot looks like so:
timestamp_annotated

Let me know if you'd like a test added for this fix. 

## PR Checklist

### DIFF
--- a/yt/utilities/definitions.py
+++ b/yt/utilities/definitions.py
@@ -37,8 +37,6 @@ formatted_length_unit_names = {'au'      : 'AU',
                                'rsun'    : 'R_\odot',
                                'code_length': 'code\ length'}
 
-formatted_time_unit_names = {'code_time' : 'code\ time'}
-
 # How many seconds are in each thing
 sec_conversion = {'Gyr'   : sec_per_Gyr,
                   'Myr'   : sec_per_Myr,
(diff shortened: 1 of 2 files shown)


jzuhone:
Hi @munkm,

This looks fine to me, but is it necessary to set the `time_unit` attribute to the code unit and then do formatting, or could one not just set `self.time_unit = '(code_time)'` or something similar?

munkm:
Great question @jzuhone. I think that's another option we could take. In doing this I followed sort of what we did in [this part](https://github.com/yt-project/yt/blob/master/yt/visualization/plot_container.py#L638-L640) of `_get_axes_unit_labels` and then looked at how we had formatted our field names in the plotters. I noticed most of those were Unit objects and used the latex function to format them. 

I could instead set up a dictionary like the `formatted_length_unit_names` so it will automatically format code_time properly when it's printed as a label. 

I maybe have an overcomplicated solution here.

matthewturk:
@munkm @jzuhone Do you think this should wait for the rewritten version?  I don't think it's that big of a deal to bring it in now.

munkm:
I'm ok with it now, but happy to change it if we feel that would be better.

matthewturk:
I'm going to go ahead and merge this, and additional things can be added later.

PR #2436, opened 2020-01-30
update imports to use utility function 

munkm:
## PR Summary

I noticed in a few places we have imports of packages that are defined in `on_demand_imports` being imported without that functionality. This PR updates them to use the utility function available for the specified package.

## PR Checklist


matthewturk:
So this looks okay, but do we want to import the underscored versions in the tests, or the un-underscored ones?

munkm:
I think we want to import the undescored versions of each module. That's their name from `on_demand_imports`.

matthewturk:
OK!  Thanks, I got confused.

Issue #2195, opened 2019-03-06
deposited artio fields have slightly different answers under yt-4.0


ngoldbaum:
This is causing answer test failures on #2172.

**Code for reproduction**

```python
import yt

ds = yt.load('sizmbhloz-clref04SNth-rs9_a0.9011/sizmbhloz-clref04SNth-rs9_a0.9011.art')

sph = ds.sphere('max', (0.1, 'unitary'))

count = sph['deposit', 'all_count']

print(count.sum())
```

**Actual outcome**

On master this prints `201313.0 dimensionless` whereas on yt-4.0 it prints `201134.0 dimensionless`.

ngoldbaum:
Here's a slightly modified version of the script in the description:

```
import yt
import numpy as np

ds = yt.load('sizmbhloz-clref04SNth-rs9_a0.9011/sizmbhloz-clref04SNth-rs9_a0.9011.art')

sph = ds.sphere('max', (0.1, 'unitary'))

count = sph['deposit', 'all_count']

counts_master = yt.YTArray.from_hdf5('counts_master.h5')

print(count.sum())

wh = np.where(count != counts_master)

print(count[wh])

print(counts_master[wh])
```

This uses a file I've uploaded here: http://use.yt/upload/17eccce3

On the yt-4 branch this prints:

```
[0. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0.
 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 1. 0. 0. 0. 0. 1. 0. 0.
 0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 2. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0.
 0. 1. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0.
 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.
 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 1. 0. 0. 0. 0. 0. 0. 0. 0. 1. 1. 1.
[...]

PR #2437, opened 2020-01-31
merging yt-4 into master

matthewturk:
This is a reopening of #2172 on my repository.  I'm removing WIP because, although it's still a work in progress, it's getting close enough that I want others to check in with it once in a while.


matthewturk:
What's currently killing me is that my conda environment doesn't have `dbm.gnu` and that's what the shelve'd answer results are in.  Sigh.

matthewturk:
I verified the images are all correct and have committed the correct versions.

We probably need to get the image answer tests somewhere other than the main repo at some point, but I do like the new system for inspecting and verifying them.

tests/tests.yaml
```diff
@@ -1,8 +1,9 @@
-answer_tests:
-
```
Xarthisius: This breaks answer testing (not a valid yaml file anymore...)

Xarthisius:
@yt-fido test this please

matthewturk:
@ngoldbaum so today, @Xarthisius, @munkm and @samwalkow and I were looking at the test results and directly comparing them, and we got some strange results.  I genuinely don't know how to interpret these.

Is it possible that this is simply because the `cm` in yt 3 and the `cm` in yt 4 don't resolve to being the same object?  Or something along the lines of a fundamental unyt / yt3-units incompatibility?

https://gist.github.com/matthewturk/38d36d8cba8127d0faf4916950648f1c

Thanks for any ideas.

matthewturk:
Ah, we dug into it a bit more -- and, it looks like both resolve the unit registry as being from `unyt` when both are imported.  And, the registry has different values for `cm`:
```
>>> r_new['cm']
(0.01, (length), 0.0, '\\rm{cm}', False)
>>> r_old['cm']
(1.0, (length), 0.0, '\\rm{cm}')
```

ngoldbaum:
I’m not sure. Maybe? Can you make a self-contained example where you save data with yt.units and load it again later with unyt that hits this problem? There’s a code path in unyt that deals with data saved by yt.units, maybe we need to do some additional massaging of the data there to get things loaded correctly?

Please feel free to open an issue in unyt about this with details and optimally steps
I can use to reproduce. I’m in Austin on a work trip this week but hopefully I’ll be able to get to it soon if you can’t come up with a fix.

ngoldbaum:
Yes, unyt used an MKS unit system internally, yt.units uses CGS.

Hmm I wonder if the problem is that the unpuckling code in unyt isn’t applying the same transformations that the json loading code does for data saved by yt.units:

https://github.com/yt-project/unyt/blob/master/unyt/unit_registry.py#L265

matthewturk:
@ngoldbaum We believe that it's because of

https://github.com/yt-project/unyt/pull/2

where the default unit system was changed.  If the unit definitions internal to a unit registry are stored with respect to the "default" and the "default" is then changed, this will change the results.

There's an elaborate analogy one could make about changing parallel universes between one where the golden meter stick is used and the golden centimeter stick is used but not *knowing* they use different golden sticks, but it's doesn't make for very exciting prose.

ngoldbaum:
Yeah I think there needs to be some logic here that detects when we’re loading data saved by yt.units:

https://github.com/yt-project/unyt/blob/master/unyt/array.py#L1928

You should be able to detect it in the same way as the json loading code I linked to earlier. Ping me on slack or here if you have any questions :)

ngoldbaum:
https://github.com/yt-project/unyt/pull/93 is also probably relevant

ngoldbaum:
@yt-fido test this please

ngoldbaum:
(hopefully that picks up the new unyt release?)

Xarthisius:
With latest `unyt` yt fails to import with:

```python
>>> import yt
  File "yt/__init__.py", line 82, in <module>
    frontends = _frontend_container()
  File "yt/frontends/api.py", line 44, in __init__
    setattr(self, frontend, importlib.import_module(_mod))
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "yt/frontends/adaptahop/api.py", line 17, in <module>
    from .data_structures import \
  File "yt/frontends/adaptahop/data_structures.py", line 47, in <module>
    class AdaptaHOPDataset(Dataset):
  File "yt/frontends/adaptahop/data_structures.py", line 53, in AdaptaHOPDataset
    _code_length_to_Mpc = Mpc.to('cm').value / 3.08e24
AttributeError: 'Unit' object has no attribute 'to'
```

matthewturk:
I hotfixed this specific issue by multiplying `Mpc` by 1.0.

matthewturk:
My plan is to continue investigating and addressing grid-based dataset differences before moving on to particles.

These in particular:

https://tests.yt-project.org/job/yt_py3_git/1382/testReport/junit/yt.frontends.chombo.tests/test_outputs/GridHierarchy_data_0000_3d_hdf5_all/
https://tests.yt-project.org/job/yt_py3_git/1382/testReport/junit/yt.frontends.chombo.tests/test_outputs/ProjectionValues_data_0005_3d_hdf5_all_velocity_magnitude_1_density/

are the next two up for me.

matthewturk:
I'm going through and addressing the answer test issues.  The first obvious issue is with sizmbholz in ARTIO.  This can be replicated by examining:

```python
import yt

ds = yt.load("sizmbhloz-clref04SNth-rs9_a0.9011/sizmbhloz-clref04SNth-rs9_a0.9011.art")

sp = ds.sphere("max", (0.1, "unitary"))
print(sp["deposit", "all_count"].sum())
```

In the yt 3.x, this results in 201313.  In the merge commit, it's 201134, for a difference of 179.  This was discussed in detail in #2195 , and I will defer additional comment to that issue.

I'm now moving on to the next set of items.

matthewturk:
@jzuhone Moving on to the Athena frontend, we seem to be getting errors about units.

https://tests.yt-project.org/job/yt_py3_git/1472/testReport/junit/yt.frontends.athena.tests/test_outputs/FieldValues_Blast_0100_all_velocity_magnitude/

The error message is:

```
The field function associated with the field '('gas', 'velocity_x')' returned data with units 'code_length/code_time' but was defined with units 'cm/s'.
```

The full traceback:

```
  File "/home/fido/workspace/yt_py3_git/yt/data_objects/data_containers.py", line 1639, in _generate_fields
    fd.convert_to_units(fi.units)
  File "/home/fido/shiningpanda/jobs/305a7a64/virtualenvs/d41d8cd9/lib/python3.5/site-packages/unyt/array.py", line 618, in convert_to_units
    new_units, self.dtype
  File "/home/fido/shiningpanda/jobs/305a7a64/virtualenvs/d41d8cd9/lib/python3.5/site-packages/unyt/unit_object.py", line 682, in get_conversion_factor
    return _get_conversion_factor(self, other_units, dtype)
  File "/home/fido/shiningpanda/jobs/305a7a64/virtualenvs/d41d8cd9/lib/python3.5/site-packages/unyt/unit_object.py", line 928, in _get_conversion_factor
    old_units, old_units.dimensions, new_units, new_units.dimensions
unyt.exceptions.UnitConversionError: Cannot convert between 'code_length/code_time' (dim '(length)/(time)') and 'cm/s' (dim '(length)/(time)').

  File "/usr/lib64/python3.5/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/usr/lib64/python3.5/unittest/case.py", line 600, in run
    testMethod()
  File "/home/fido/shiningpanda/jobs/305a7a64/virtualenvs/d41d8cd9/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/fido/workspace/yt_py3_git/yt/utilities/answer_testing/framework.py", line 344, in __call__
    nv = self.run()
  File "/home/fido/workspace/yt_py3_git/yt/utilities/answer_testing/framework.py", line 464, in run
    field, weight=weight_field)
  File "/home/fido/workspace/yt_py3_git/yt/data_objects/derived_quantities.py", line 128, in __call__
    rv = super(WeightedAverageQuantity, self).__call__(fields, weight)
  File "/home/fido/workspace/yt_py3_git/yt/data_objects/derived_quantities.py", line 56, in __call__
    sto.result = self.process_chunk(ds, *args, **kwargs)
  File "/home/fido/workspace/yt_py3_git/yt/data_objects/derived_quantities.py", line 134, in process_chunk
    for field in fields]
  File "/home/fido/workspace/yt_py3_git/yt/data_objects/derived_quantities.py", line 134, in <listcomp>
    for field in fields]
  File "/home/fido/workspace/yt_py3_git/yt/data_objects/data_containers.py", line 256, in __getitem__
    self.get_data(f)
  File "/home/fido/workspace/yt_py3_git/yt/data_objects/data_containers.py", line 1595, in get_data
    self._generate_fields(fields_to_generate)
  File "/home/fido/workspace/yt_py3_git/yt/data_objects/data_containers.py", line 1648, in _generate_fields
    raise YTFieldUnitError(fi, fd.units)
yt.utilities.exceptions.YTFieldUnitError: The field function associated with the field '('gas', 'velocity_x')' returned data with units 'code_length/code_time' but was defined with units 'cm/s'.
```

Does this ring any bells?

jzuhone:
@matthewturk This error is bizarre. It seems to suggest that you can't convert between two units of the same dimensions.

matthewturk:
I know, right?  I'm wondering if the key element there is that they're in
"code" units, and perhaps the specific meaning of "code" is different --
since we're comparing code yt.units versus unyt ?  Maybe?

jzuhone:
I would imagine. Is this the only Athena test that fails?

matthewturk:
No, but from what I can tell they are almost all variations of this one
theme.

Xarthisius:
@matthewturk have you been able to reproduce it outside of testing framework? Cause it might be as well a red herring and the actual error is related to state/test serialization and unyt. For me, this code:

```python
import yt

# yt.frontends.athena.tests.test_outputs.FieldValues_Blast_0100_all_velocity_magnitude

blast = "MHDBlast/id0/Blast.0100.vtk"
ds = yt.load(blast)
obj = ds.all_data()

for field in ("temperature", "density", "velocity_magnitude"): 
    field = obj._determine_fields(field)[0]
    fd = ds.field_info[field]
    if fd.is_sph_field:
        weight_field = (field[0], "ones")
    else:
        weight_field = ("index", "ones")

    avg = obj.quantities.weighted_average_quantity(field, weight=weight_field)
    mi, ma = obj.quantities.extrema(field)
    print(avg, mi, ma)
```
works fine and this is what that failing test is doing.

matthewturk:
Yeah, I actually was unable to get it to produce a "code_length/code_time"
field when I requested it from the command line.  So ... Not quite sure
what to do here, except possibly set up a special case for the merge where
we compare the numpy arrays rather than the unyt/YTArrays.

Xarthisius:
as a side note I'm able to reproduce the issue with `temperature` without any problem:

```python
import yt
ds = yt.load("MHDBlast/id0/Blast.0100.vtk")
ad = ds.all_data()["temperature"]
```

yields

```
  File "ala.py", line 3, in <module>
    ad = ds.all_data()["temperature"]
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/data_containers.py", line 249, in __getitem__
    f = self._determine_fields([key])[0]
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/data_containers.py", line 1366, in _determine_fields
    finfo = self.ds._get_field_info("unknown", fname)
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/static_output.py", line 758, in _get_field_info
yt.utilities.exceptions.YTFieldNotFound: Could not find field '('io', 'temperature')' in Blast.0100.
```

matthewturk:
For what it's worth, `"gas","temperature"` also fails.

Xarthisius:
This one also fails:

```python
$ cat athena.py 
import shelve
import yt
from yt.testing import assert_equal

old = shelve.open("py35_local_athena_005/py35_local_athena_005")
blast_old = old["Blast.0100"]
old_result = blast_old["GridHierarchy_Blast_0100_all"]

blast = "MHDBlast/id0/Blast.0100.vtk"
ds = yt.load(blast)
result = {}
result["grid_dimensions"] = ds.index.grid_dimensions
result["grid_left_edges"] = ds.index.grid_left_edge
result["grid_right_edges"] = ds.index.grid_right_edge
result["grid_levels"] = ds.index.grid_levels
result["grid_particle_count"] = ds.index.grid_particle_count
new_result = result

for k in new_result:
    assert_equal(new_result[k], old_result[k])
```

I have a hunch that figuring this out, will fix `cm/s != code_length/code_time` too...

Xarthisius:
@yt-fido test this please (I doubled the number of threads to increase test separation)

Xarthisius:
@yt-fido test this please

matthewturk:
Oops, didn't mean to close this.

matthewturk:
@neutrinoceros One thing to note is that I've changed `test_add_gradient_fields_curvilinear` in this merge, as it doesn't work with the current yt-4.0 implementation of `apply_units`, which will not coerce between dimensions.

matthewturk:
One of the more worrisome test failures I ran into is the values for Enzo's galaxy0030 field `velocity_divergence`.  It seems as though these should not change -- the grid machinery is largely untouched.  The values differ by extremely small amounts, on the order of relatively change of 1e-10 at most and 1e-16 at the minimum.  What I circled around as a possible reason is that `unyt` allows for 32 bit arrays, whereas the `YTArray` items are always 64-bit.

I have done the following things to attempt to debug:

1. Examined in 3 and 4 the values for velocity xyz, magnitude, and divergence.  Only divergence shows the issue.
2. Attempted to reduce the moving parts -- don't use relative velocity, only use one field, etc.  This did not change the results.
3. Force the enzo fields to 64 bit in the IO area.

None of these things materially changed the results.  I was still getting non-bitwise identical values.  *However*, I did also export the (32-bit!) values to an hdf5 file which I re-loaded using `load_amr_grids`, but that was identical.

So where I'm at is that I think that there's a floating point order-of-operations and order-of-conversions that is changing the results by a tiny amount between the two.  But, since I can't *actually* figure out where that's occurring, I'm not yet sure that I can correct it.  I'm going to spend a bit more time on this.

matthewturk:
As an addenda, I have now upcast the entirety of the `IsolatedGalaxy` dataset into 64 bits:

```python
import glob
import h5py


for fn in sorted(glob.glob("IsolatedGalaxy/galaxy0030/*.cpu*")):
    print(fn)
    with h5py.File(fn, "r") as f:
        with h5py.File(fn.replace("IsolatedGalaxy/", "IsolatedGalaxy64/"), "w") as f2:
            def saveit(n, ds):
                if not hasattr(ds, 'dtype'): return
                vals = ds[()]
                if ds.dtype == ">f4":
                    vals = vals.astype("f8")
                f2.create_dataset(n, data=vals)
            f.visititems(saveit)
```

and it still produces the incorrect results, so I'm somewhat less certain now.

matthewturk:
So @Xarthisius tracked this difference down to https://github.com/yt-project/unyt/pull/58/ ; the differences are exceptionally small, so the new behavior is likely an improvement, but reduces our bitwise comparison ability.  Right now we're working out how to proceed with testing so that we can verify, but also not have to hackily patch unyt to do that verification.

Xarthisius:
@yt-fido test this please

matthewturk:
@neutrinoceros Does this look familiar to you?

```
  File "/usr/lib64/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib64/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/usr/lib64/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/fido/shiningpanda/jobs/a7d58705/virtualenvs/d41d8cd9/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/fido/workspace/yt_py38_git/yt/utilities/answer_testing/framework.py", line 344, in __call__
    nv = self.run()
  File "/home/fido/workspace/yt_py38_git/yt/utilities/answer_testing/framework.py", line 524, in run
    obj = create_obj(self.ds, self.obj_type)
  File "/home/fido/workspace/yt_py38_git/yt/utilities/answer_testing/framework.py", line 1039, in create_obj
    obj = cls(*obj_type[1])
  File "/home/fido/workspace/yt_py38_git/yt/data_objects/selection_data_containers.py", line 787, in __init__
    super(YTSphere, self).__init__(center, ds,
  File "/home/fido/workspace/yt_py38_git/yt/data_objects/data_containers.py", line 2017, in __init__
    self._set_center(center)
  File "/home/fido/workspace/yt_py38_git/yt/data_objects/data_containers.py", line 193, in _set_center
    self.center = self.ds.find_max(("gas", "density"))[1]
  File "/home/fido/workspace/yt_py38_git/yt/data_objects/static_output.py", line 795, in find_max
    center = self.arr([mx, my, mz], dtype="float64").to('code_length')
  File "/home/fido/shiningpanda/jobs/a7d58705/virtualenvs/d41d8cd9/lib/python3.8/site-packages/unyt/array.py", line 526, in __new__
    return _coerce_iterable_units(input_array, registry)
  File "/home/fido/shiningpanda/jobs/a7d58705/virtualenvs/d41d8cd9/lib/python3.8/site-packages/unyt/array.py", line 221, in _coerce_iterable_units
    raise IterableUnitCoercionError(input_object)
unyt.exceptions.IterableUnitCoercionError: Received a list or tuple of quantities with nonuniform units: [unyt_quantity(0.5703, 'cm'), unyt_quantity(2.7734, '(dimensionless)'), unyt_quantity(0.5, 'cm')]
```

neutrinoceros:
It doesn’t. Is this test running on non-Cartesian data ?

matthewturk:
So this comes from just merging all of the tests that are on master into yt
4.0.  So it may be, if that is tested.

matthewturk:
I should note, it's probably coming up because we now use `unyt` instead of yt's units, and they have diverged a bit.

neutrinoceros:
My money is on the failing dataset to be cylindrical. The fact is that angular coordinates are assigned a length dimension when returned by methods such as ds.find_max, at least on the master branch, so there surely are some implementations that rely on this odd behavior. I’m guessing it was changed on the 4.0 branch (yay), which makes those workarounds broken.

matthewturk:
@neutrinoceros What I would genuinely like to see happen -- which is likely a long ways off, if ever -- is to have a set of index-focused array types that could be 1, 2 or 3 elements (or dimensions) and that would have fixed and unchanging units, which would be different for each dimension.  That's not something I can personally implement, though.

Do you think we should try reverting some of the workarounds?  Can you give me a pointer?

matthewturk:
I verified the AREPO tests manually, so I am bumping their version number, and I am merging to get the updated answer tests elsewhere.

matthewturk:
Note that the travis tests will now likely fail because I'm using the yt4 branch of the `answer-store` submodule; I will then have to verify the results and recommit them to the yt4 branch of that submodule.

matthewturk:
masonry

matthewturk:
I'm not able to see the same errors here for `enzo_p`.

matthewturk:
Looks like a fair number of errors in fido's answer testing, but I think the next update of the answer tests repo hash will fix travis's run.

matthewturk:
At present, I've fixed the RAMSES tests and have asked it to regenerate the Athena++ tests.  I think that the tests are hanging because of the bitmask changes, and I've asked to regenerate the bitmask indexes too.

This next run should go to completion and then I'll have a better idea of the census of tests.

One thing I would note is that once this is merged, I intend to regenerate *all* of the answer tests.  Right now, an enormous amount of time is being spent in the function `_convert_old_unit_registry` in unyt, and I think that regenerating the (passing) answer tests will drop this considerably.

matthewturk:
The xray field tests are all failing because of a missing set of old answers; this is because the old answers used non-tupleized field names, and the new ones use tupleized names.

I corrected this locally and found that there were differences, which I've included in the attached log message.  The upshot is this set of differences:

```
Max absolute difference: 1.36941478e-05
Max relative difference: 0.01010277    
Max absolute difference: 1.43113586e-05
Max relative difference: 0.00990764    
Max absolute difference: 1.21520488e-05
Max relative difference: 0.01014338   
Max absolute difference: 3.84510788e-29
Max relative difference: 0.00964158   
Max absolute difference: 1.31701732e-05
Max relative difference: 0.01031818   
Max absolute difference: 1.37593654e-05
Max relative difference: 0.01010183    
Max absolute difference: 1.16300891e-05
Max relative difference: 0.01031622  
Max absolute difference: 3.84510788e-29
Max relative difference: 0.01020763   
Max absolute difference: 2.40297975e+62
Max relative difference: 0.00950217   
Max absolute difference: 2.48471128e+62
Max relative difference: 0.0093848     
Max absolute difference: 2.17731299e+62
Max relative difference: 0.00973833
Max absolute difference: 5.38638704e+38
Max relative difference: 0.00964158
Max absolute difference: 1.84493264e+62                                                                   
Max relative difference: 0.01031818             
Max absolute difference: 1.92746913e+62                                                                   
Max relative difference: 0.01009939    
Max absolute difference: 1.62919125e+62
Max relative difference: 0.01031622    
Max absolute difference: 5.38638704e+38
Max relative difference: 0.01020763    
Max absolute difference: 3204.88638303
Max relative difference: 0.00715961    
Max absolute difference: 3436.27911469
Max relative difference: 0.00701946    
Max absolute difference: 2815.88910964
Max relative difference: 0.00735913    
Max absolute difference: 9.09026729e-21
Max relative difference: 0.00645986    
Max absolute difference: 3077.3545124
Max relative difference: 0.00712593    
Max absolute difference: 3297.04995432
Max relative difference: 0.00668652    
Max absolute difference: 2686.15565521
Max relative difference: 0.00724716    
Max absolute difference: 9.09026729e-21
Max relative difference: 0.00645986    
```

I am going to inspect the really *big* ones, but there's also the matter of the new atomic database being used, and switching from `apec_emissivity_v2.h5` to `apec_emissivity_v3.h5`.  So I'm going to hazard a guess that I'll just need to regenerate these.

[xray_field.txt](https://github.com/yt-project/yt/files/4748594/xray_field.txt)

matthewturk:
I have manually inspected the Chombo answers, and they are likely a result of changing CGS to MKS.  The degree of difference gets worse (still under a percent) in projections.  So I am bumping those answers.

matthewturk:
As of now, `test_attributes_wt` needs to be bumped, but `test_raw_field_slices` actually requires a fix in either yt or unyt (see https://github.com/yt-project/unyt/issues/145 ) to work around some matplotlib issues.  So I am not bumping `local_pw_028` until that's been addressed.  `test_particle_phase_answers` also needs bumping to correct the no-longer-italicized `M` in the units.

Xarthisius:
@yt-fido test this please

matthewturk:
@yt-fido test this please

matthewturk:
In #2624 I've updated the TNG halo tests, and when that PR is accepted into upstream, I'll merge, and then the tests for Arepo will regenerate.

matthewturk:
Assuming that these PRs get merged:

 * #2627 
 * #2624 
 * #2617 

the following frontends need to be addressed:

 * `enzo_p` -- I don't know why this is failing and cannot reproduce it.
 * `gadget` -- these will need to be validated and then re-generated.
 * `gamer` -- these will need to be validated and then re-generated.
 * `gizmo` -- these will need to be validated and then re-generated.
 * `owls` -- these will need to be validated and then re-generated.
 * `tipsy` -- these will need to be validated and then re-generated.

Additionally, the visualization tests will need a fix applied to unyt before they can be re-generated, but they seem all correct to me when I inspect them other than the unyt issue mentioned above.

I don't know why the xray tests are failing again, after I recently regenerated them.  Kacper thinks it might be a race condition.

This is extremely tractable!  I will return to this on Thursday and hopefully we can get this knocked out ASAP.

jzuhone:
@matthewturk I'm taking GAMER tests

chummels:
In terms of manually inspecting the answer test results from the SPH frontends you list above, which test results should we examine?  ProjectionPlot or SlicePlot answer tests?

matthewturk:
Probably both, if you can.  The more eyes we can get on different things,
the better.

matthewturk:
@chummels It occurs to me that perhaps we could instrument the `sph_tests` function instead, to generate lots of images to examine.  I'm going to try that, and then I'll upload the images somewhere.

chummels:
Sounds great!  Keep me posted!

matthewturk:
OK, so this should be a quick and easy way to run and get images out.

```
diff --git a/yt/utilities/answer_testing/framework.py b/yt/utilities/answer_testing/framework.py
index 9f22ea01c..2e860e6ef 100644
--- a/yt/utilities/answer_testing/framework.py
+++ b/yt/utilities/answer_testing/framework.py
@@ -1013,6 +1013,7 @@ def big_patch_amr(ds_fn, fields, input_center="max", input_weight="density"):
 
 
 def sph_answer(ds, ds_str_repr, ds_nparticles, fields):
+    import yt
     if not can_run_ds(ds):
         return
     assert_equal(str(ds), ds_str_repr)
@@ -1030,9 +1031,19 @@ def sph_answer(ds, ds_str_repr, ds_nparticles, fields):
                 particle_type = False
             for axis in [0, 1, 2]:
                 if not particle_type:
-                    yield PixelizedProjectionValuesTest(
+                    a =  PixelizedProjectionValuesTest(
                         ds, axis, field, weight_field,
                         dobj_name)
+                    yield a
+                    output_fn_base = os.path.join(
+                           os.path.realpath(a.options.output_dir),
+                            a.description)
+                    _dobj = create_obj(ds, dobj_name)
+                    p = yt.ProjectionPlot(ds, axis, field, weight_field = weight_field,
+                                          data_source = _dobj)
+                    print(p.save(output_fn_base + "_zoom0.png"))
+                    p.zoom(5)
+                    print(p.save(output_fn_base + "_zoom1.png"))
             yield FieldValuesTest(ds, field, dobj_name,
                                   particle_type=particle_type)
 
```

This requires that you have your config file set up.  So for instance, I run with this as my `nose_generate_answers.cfg` file:

```
[nosetests]
detailed-errors=1
local-dir=/data2/yt_sample_data/answer-store2/
answer-store=1
answer-name=yt3
local=1
nocapture=1
nologcapture=1
verbosity=2
where=yt
with-answer-testing=1
with-timer=1
with-xunit=1
xunit-file=answer_nosetests.xml
answer-big-data=1
```

and I run with this command to run a specific frontend, in this case Arepo, specifically the bullet cluster dataset:

`nosetests -s -c nose_generate_answers.cfg frontends/arepo/tests/test_outputs.py:test_arepo_bullet`

This will generate the plots in the directory specified in the answer store variable in the config file.

Not all of the datasets will necessarily work, just because some of them don't use `sph_answer` but this should get us part of the way.

I'll check if we can just run this somewhere and post the images, rather than having folks do it manually, and I'll work on a similar thing for yt-3.

matthewturk:
I'm going to instrument the appropriate tests and generate the visualizations, and I'll set up a side-by-side deployment of the images.

That will leave the three outstanding PRs linked above, the unyt / unicode issue, and the xray fields.

matthewturk:
Here is what I have come up with for a gallery generator.  This just does the yt4 images so far.

https://github.com/matthewturk/yt4-gallery

Working on adding yt3, and then I'll build some HTML scaffolding and upload it to gh-pages.

chummels:
Awesome!  Thanks for making this more automated and easier.  I can see this potentially be employed down the line whenever we see answer test differences for image data beyond the yt4 merge.  Great work!

jzuhone:
@matthewturk I've checked GAMER and it's good.

jzuhone:
@matthewturk x-ray tests are good

matthewturk:
I've built out a gallery in this repository:

https://github.com/matthewturk/yt4-gallery

and the build is here: https://matthewturk.github.io/yt4-gallery/

I've also found a handful of places that need a bit of cleanup in the merge and yt 4 codebase, and identified one big fault (`snipshot399_z000p000` can't be loaded, as it doesn't have or fall back on generating a smoothing length) that needs to be addressed.  I'll file them all or issue PRs later tonight or tomorrow.

I am hoping that folks can give the o-k on updating the answers with this; I'm going to probably update the gallery to also include some cutouts as well, and potentially additional fields, but right now it should be sufficient.

(ping @jzuhone @AshKelly @chummels )

AshKelly:
Some comments:
`snapshot_033` - Temperature: the yt-4.0 has some weird cold spots. I actually think I know the origin of this. I think it's to do with inconsistently increasing a smoothing length during deposition - I debugged this with @munkm and @brittonsmith a little in a triage.

`IsothermalCollapse` - All checks out. It looks beautiful in yt-4.0

`gizmo_64` - The same as `snapshot_033`

`snapshot_028_z000p000` - The same as `snapshot_033`. Again, the yt-4.0 images are a thing of beauty, especially the velocity magnitude.

`GadgetDiskGalaxy` - It's very clear this is `scatter` smoothing, but yeah again, everything seems to check out other than those spots.

`TipsyGalaxy` - The velocity magnitude looks wrong. In yt-4 the surrounding region is yellow ~10^9 but in yt-3.5 it's just white. It sort of looks like a fast moving, large smoothing length particle has deposited.

matthewturk:
@AshKelly Thank you!!

So for `snapshot_033` should we debug that before merging into master, or just update the answer tests a second time later?

For `TipsyGalaxy`, I think you're probably right.  That's the one I had the hardest time making work, too, and I had to do some stuff with adjusting the bounding box.  So it's possible that this is something that will need more work.

neutrinoceros:
- `IsothermalCollapse`, it looks to me that structures are lost specifically in the temperature snapshots (all other quantities look fine)

- `TipsyGalaxy` : on top of Ash's comment, I would add that velocity magnitude's gradient's sign is apparently reversed, I don't know if this is the same issue you were discussing

AshKelly:
> So for snapshot_033 should we debug that before merging into master, or just update the answer tests a second time later?

If the answer tests are fairly easy to replace then we should probably just wait. This is already a colossal merge! 

>     * `IsothermalCollapse`, it looks to me that structures are lost specifically in the temperature snapshots (all other quantities look fine)

I sort of feel like the yt-4.0 answer is correct. If it's an isothermal collapse then I believe the temperature is constant and it should be structure-less.

>     * `TipsyGalaxy` : on top of Ash's comment, I would add that velocity magnitude's gradient's sign is apparently reversed, I don't know if this is the same issue you were discussing.
>       More over, the temperature's structures and ranges look inconsistent

True, I think the yt-3.5 is clipping though so hard to tell.

neutrinoceros:
> If it's an isothermal collapse then I believe the temperature is constant and it should be structure-less.
Makes sense. So the structures we do see in yt 3.5 are in fact the artefacts I guess ? Do you guys understand where they came from ? (sorry I feel like I'm only going to ask questions and give no answer here)

AshKelly:
> Makes sense. So the structures we do see in yt 3.5 are in fact the artefacts I guess ? Do you guys understand where they came from ? (sorry I feel like I'm only going to ask questions and give no answer here)

I *think* they are artefacts in yt-3.5, but what causes them? hm, good question, I wish I knew. I'm not sure entirely how yt-3.5 is depositing the fields, but I guess its related to how yt is sampling the field. Questions are good; otherwise things go unspotted!

neutrinoceros:
Another thing I just noticed with the temperature maps for TipsyGalaxy: something's going wrong with the colorbar's placement
Screenshot 2020-06-16 at 19 54 33

matthewturk:
So if I'm reading this right ... y'all might be giving me the sign-off to bump the answers???  :)

matthewturk:
I've attempted to address the unyt change in https://github.com/yt-project/unyt/pull/153 .

matthewturk:
I've bumped the enzo_p tests; these are the diffs.  First column is old, second column is new, third is difference.  Looks like a very slight alignment difference that is on an extremely sharp gradient, so thus the major differences in values.
enzo_p_diff

Xarthisius:
@yt-fido test this please

matthewturk:
Well, in fun news, I don't know why the answers I just bumped are failing now.  Hooray!  That'll be a job for tomorrow.

matthewturk:
So here's the current state.  I believe if we merge #2624 , #2617, #2627 and #2644, the remaining *frontend* answer tests will go away.

And with the change in `unyt`, *many* of the plot answer tests will be fixed, and the remaining can safely be bumped.

So I think if we can review and merge those 4 PRs, I can update this one, and then we should be in a position to have this merged by, say, end of week.

chummels:
@matthewturk thanks for generating those galleries to compare the answer tests in yt 3.5 and yt 4.0.  I've now had a chance to examine all of the images and here are my notes:

The main change here across the board is that in yt4, there is less blockiness caused by the construction of the octree for particle-based datasets.  This results in the absence of rectilinear artifacts in the yt4 projections, much more akin to what the simulations are producing.  It also gives substantially higher spatial resolution, particularly in the centers of galactic halos, where the particle density is higher.  This is most obvious in the column density projections, but shows up in the averaged projections as well.  This increased spatial resolution leads to an increase in the dynamic range of these fields, which changes the colormap range to accommodate it.  All this is to say is, yt4 consistently gives better results, consistent with expectations from the demeshening.  It's like on a TV show when the supervisor asks the computer person to "enhance!"--we're seeing the enhancement with yt4!

A few other notes:

snipshot_399 isn't producing yt4 results in your gallery.  Is this just an error in your rendering software, or is it timing out because it's taking too long, or is it that the tests are not running for snipshot_399 in yt4?  If the latter, this may be a problem.

TipsyGalaxy is having some weird behavior, in that the octree seems to have very fixed boundaries in a well-defined domain.  In yt-4, the smoothing length of the particles appears to take it far beyond that constrained domain, which is somewhat expected.  However, it looks like there are a handful of gas particles with extremely large smoothing lengths, on the order of 5-10 times the size of the domain and offset from the domain by as much.  This looks fishy, like the centers of these particles might be wrong.  See TipsyGalaxy, density, x, for a good example of this.  

I really like the "orange peel" effect of the temperature slice for the Isothermal Collapse simulation.  It looks like the surface of the Sun in H-alpha.  And it also looks the way we would expect for an isothermal sphere--definitely an improvement over yt3.5.

It sounds like you've already bumped the answer tests for these, but I think we should investigate snipshot_399 and the TipsyGalaxy results to ensure we understand what is going on before accepting these changes entirely.

matthewturk:
> @matthewturk thanks for generating those galleries to compare the answer tests in yt 3.5 and yt 4.0. I've now had a chance to examine all of the images and here are my notes:
> 
> The main change here across the board is that in yt4, there is less blockiness caused by the construction of the octree for particle-based datasets. This results in the absence of rectilinear artifacts in the yt4 projections, much more akin to what the simulations are producing. It also gives substantially higher spatial resolution, particularly in the centers of galactic halos, where the particle density is higher. This is most obvious in the column density projections, but shows up in the averaged projections as well. This increased spatial resolution leads to an increase in the dynamic range of these fields, which changes the colormap range to accommodate it. All this is to say is, yt4 consistently gives better results, consistent with expectations from the demeshening. It's like on a TV show when the supervisor asks the computer person to "enhance!"--we're seeing the enhancement with yt4!
> 
> A few other notes:
> 
> snipshot_399 isn't producing yt4 results in your gallery. Is this just an error in your rendering software, or is it timing out because it's taking too long, or is it that the tests are not running for snipshot_399 in yt4? If the latter, this may be a problem.

This is due to an issue fixed in #2645 -- I intend to regenerate.

> 
> TipsyGalaxy is having some weird behavior, in that the octree seems to have very fixed boundaries in a well-defined domain. In yt-4, the smoothing length of the particles appears to take it far beyond that constrained domain, which is somewhat expected. However, it looks like there are a handful of gas particles with extremely large smoothing lengths, on the order of 5-10 times the size of the domain and offset from the domain by as much. This looks fishy, like the centers of these particles might be wrong. See TipsyGalaxy, density, x, for a good example of this.
> 
> I really like the "orange peel" effect of the temperature slice for the Isothermal Collapse simulation. It looks like the surface of the Sun in H-alpha. And it also looks the way we would expect for an isothermal sphere--definitely an improvement over yt3.5.
> 
> It sounds like you've already bumped the answer tests for these, but I think we should investigate snipshot_399 and the TipsyGalaxy results to ensure we understand what is going on before accepting these changes entirely.

snipshot_399 is definitely under way; just above, @AshKelly mentioned some stuff about `TipsyGalaxy`.  I am not sure if it's wrong or not, or if we are just visualizing in a space that isn't well-defined.

chummels:
I dug a little deeper on the TipsyGalaxy stuff.  yt definitely believes that there is a single errant particle 3.5 Mpc away from all of the other gas particles in the simulation domain, at: unyt_array([ -510.06057739,   232.48110962, -3623.55981445], 'code_length').  I have no real way of inspecting the native Tipsy data, as I'm not sure how to read in that binary data in the absence of a code like yt.  

As Ash says, it seems that yt 3.5 just clipped the domain, so it wasn't really an issue before?  Basically in both yt3.5 and yt4, things look unphysical outside of the inner region that's well-populated with particles.  So I guess in this regime, the data aren't really well-defined and shouldn't be trusted anyway.  Anyway, I think it's OK with just leaving that answer test alone moving forward.  It might be worth revisiting if Tipsy users complain though.

cphyc:
> -510.06057739, 232.48110962, -3623.55981445

I loaded the dataset using pynbody and was able to find the same particle at location `SimArray([ -510.06057739,   232.48110962, -3623.55981445], 'kpc')`, so I doubt there's something fishy on yt side.

chummels:
@cphyc OK, awesome!  Thanks for looking at that.  I guess the dataset is just a bit wonky.  So, it seems like yt4 is doing the right thing here.  In my mind, that means all the frontend tests are correct, and the only one that needs to be checked is the `snipshot` test after regenerating it by incorporating the add_sph_fields() functionality.

matthewturk:
The before and after plots with the different `mu` unicode symbol are here:

old_mu
new_mu

As you can see, the change in the unicode number has caused the mu to be italicized, which is likely because now it can be, or something, I am not sure.  Whatever.  I am not going to dip back into that well just yet, and instead, going to push forward with the answer verification.

matthewturk:
Wait, forget that last comment, I read the namesb ackwards.  *Now* it's not italicized.  So that's good.

matthewturk:
This new `degF` thing is unfamiliar to me, but might be a result of running off `unyt` master.  Maybe?

Xarthisius:
@yt-fido test this please (regenerating xray fields tests, due to missing emissivity file)

Xarthisius:
@yt-fido test this please (regenerating ytdata tests since now they actually work, at least half of them...)

matthewturk:
I've pushed a handful of changes that should both fix a couple tests *and* show a couple more tests that don't work.

`owls_subfind` has not been updated to use the new start/stop machinery; because it's already a complicated beast with internal offset calculations, I've opted to fix this by allowing infinite chunk sizes in #2658.

The two tipsy tests `test_pkdgrav` and `test_gasoline_dmonly` that were failing should be partially passing now; the `particle_types_raw` attribute should have been used, as it reimplements a few things that are in `sph_answers` and `nbody_answers` and were not updated.  Note that it should now fail with fields-not-found errors, or something, and we should change to use the new machinery.

One issue I am having with pkdgrav (specifically) right now is that I am unable to load it when I use the arguments given in the testing file.  It needs a value for `h` in the default unit registry, but one does not exist, and so the bootstrapping logic for setting up cosmology and units is out of sorts.  I am working on addressing that in a different pull request.

I filed a handful of issues today while working on this: #2659, #2657, and #2656.

Xarthisius:
@yt-fido test this please

matthewturk:
Thanks to @xarthisius for all the work he put in on this.  Hopefully this most recent push will reduce the number of failing tests to a single one, which is the `ParticleSelectionComparison` for pkdgrav.  I'm not yet sure why, but it's returning empty arrays in some cases.  Working on that...

munkm:
🎉 🥇 🎈 🎊 🎈 🥇 🎉

PR #2438, opened 2020-02-03
Use all available cores for both cythonize and build_ext

Xarthisius:
This patch helps to speed up `yt`'s build time by parallelizing both `cythonize` and disutils' extension build.

Before:

```bash
$ time python setup.py build_ext
real	4m25.461s
user	4m24.232s
sys	0m6.524s
```

After (on my 4core/8HT laptop) :

```bash
$ time python setup.py build_ext
real	1m32.580s
user	10m40.215s
sys	0m12.188s
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,6 @@
 import os
+import multiprocessing
+from concurrent.futures import ThreadPoolExecutor as Pool
 import glob
 import sys
 from sys import platform as _platform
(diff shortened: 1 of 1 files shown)


ngoldbaum:
A little nervous about this possibly breaking on e.g. windows or other non-linux architectures. But I guess we'll get bug reports from users?

setup.py
```diff
@@ -12,9 +14,39 @@
...
 from distutils.version import LooseVersion
+from distutils.ccompiler import CCompiler
 import pkg_resources
 
 
+def _get_cpu_count():
+    try:
+        nthreads = os.cpu_count()
+    except AttributeError:
+        nthreads = multiprocessing.cpu_count()
```
ngoldbaum: Out of curiosity, why can't we just use `os.cpu_count`?

Xarthisius: For some reason I thought `os.cpu_count()` was introduced fairly recently, but now I see it was py3.4

setup.py
```diff
@@ -12,9 +15,36 @@
     read_embree_location, \
     in_conda_env
 from distutils.version import LooseVersion
+from distutils.ccompiler import CCompiler
 import pkg_resources
 
 
+def _get_cpu_count():
+    if platform.system() != "Windows":
+        return os.cpu_count()
```
munkm: I'm curious what happens if somebody does run this with a Windows machine? Do we handle that somewhere else?

Xarthisius: This function will return `None`, which is a default value for `nthreads` in `cythonize` and it will evaluate to `False` in if statement in `build_extensions`

munkm: Great! Thanks for explaining that. 🙂

Xarthisius: I just noticed it's actually 0 that's default but it looks like it works with `None` too. However, I'll fix that

PR #2440, opened 2020-02-05
Update documentation to reflect usage of rcfile (no brackets allowed)

### LINKED ISSUE #2439: bug: can not set rcfile values between brackets (example with default_colormap)
related previous issue and fix PR
https://github.com/yt-project/yt/issues/2389
https://github.com/yt-project/yt/pull/2390

### Bug report

**Bug summary**

setting my `$HOME/.config/yt/ytrc` with either
```
default_colormap = "cmo.thermal"
# or
default_colormap = "thermal"
```
results in the same error when attempting to plot stuff. (`ValueError`)

**Code for reproduction**

```python
ds = yt.testing.fake_amr_ds() 
yt.plot_2d(ds, "Density")
```

**Actual outcome**
http://paste.yt-project.org/show/196/
note that `"thermal"` *is* among the supposedly registered options.

neutrinoceros:
fix #2439 
It's possible that doing this (stripping `"` and `'` from values in the rcfile) is actually not a good idea for reasons I'm not aware of.

### DIFF
--- a/doc/source/reference/configuration.rst
+++ b/doc/source/reference/configuration.rst
@@ -91,7 +91,7 @@ used internally.
 * ``test_data_dir`` (default: ``/does/not/exist``): The default path the
   ``load()`` function searches for datasets when it cannot find a dataset in the
   current directory.
-* ``reconstruct_index`` (default: ``True``): If True, grid edges for patch AMR
+* ``reconstruct_index`` (default: ``True``): If true, grid edges for patch AMR
   datasets will be adjusted such that they fall as close as possible to an
   integer multiple of the local cell width. If you are working with a dataset
   with a large number of grids, setting this to False can speed up loading
--- a/doc/source/reference/configuration.rst
+++ b/doc/source/reference/configuration.rst
@@ -80,15 +80,15 @@ The following external parameters are available.  A number of parameters are
 used internally.
 
 * ``coloredlogs`` (default: ``False``): Should logs be colored?
-* ``default_colormap`` (default: ``"arbre"``): What colormap should be used by
+* ``default_colormap`` (default: ``arbre``): What colormap should be used by
   default for yt-produced images?
 * ``loadfieldplugins`` (default: ``True``): Do we want to load the plugin file?
-* ``pluginfilename``  (default ``"my_plugins.py"``) The name of our plugin file.
+* ``pluginfilename``  (default ``my_plugins.py``) The name of our plugin file.
 * ``logfile`` (default: ``False``): Should we output to a log file in the
   filesystem?
 * ``loglevel`` (default: ``20``): What is the threshold (0 to 50) for
   outputting log files?
-* ``test_data_dir`` (default: ``"/does/not/exist"``): The default path the
+* ``test_data_dir`` (default: ``/does/not/exist``): The default path the
   ``load()`` function searches for datasets when it cannot find a dataset in the
   current directory.
 * ``reconstruct_index`` (default: ``True``): If True, grid edges for patch AMR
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Instead of doing things for all config values, which may break existing config files in unexpected ways, maybe it would be better to do this for the config option you were having trouble with. It would be nice to get a test too, see the tests for the config file, which you may have missed:

https://github.com/yt-project/yt/blob/a209a67f2977571925bc297739003254130ef9b3/yt/utilities/tests/test_config.py

neutrinoceros:
Actually it's the second time I'm running into this, and I seem to always assume that using brackets is okay, for string values. I don't think it would make sense to restrict this to a subset of parameters as it would make the api more confusing instead of more flexible. You're right that having a test would be nice. I really can't think of a scenario where this change would break existing refiles but you are right to be cautious, so I'd need to ponder on this a little more to see if I can come up with relevant tests to cover the "unexpected". If no one here can actually think of a specific way this could be a problem though, I'd argue it's reasonable to assume this is safe, but only if enough effort is dedicated in thinking of a way to break it :)

neutrinoceros:
as a note, the documentation actually uses brackets `'` for every single example value (even non-strings)
https://yt-project.org/docs/dev/reference/configuration.html#available-configuration-options

ngoldbaum:
Hi just want to note that I'm not going to engage on this pull request until my concerns about breaking existing configuration files are sufficiently addressed. A simple way to move forward would be to implement my suggestion about fixing the specific issue you had any not making an invasive change to the configuration system without sufficient testing.

I also want to say that I find the experience of having me, an experienced maintainer with lots of experience with yt and changes to yt breaking users, having my concerns ignored by a relatively new contributor to be extremely frustrating and not conducive to working with you in the future in a productive way.

matthewturk:
Hi folks, I think this is a use case we want to support.  I'm not entirely certain that I see why this would break existing user config files -- in general, I think folks using yt *don't* probably do a lot of complicated config files, and if we look at the documentation for what is allowed, I think we are likely in a fine position to change this particular use case.

@ngoldbaum I understand your concerns about breaking changes.  The documentation is indeed a bit unclear about whether we want the quotes or not.

We're limited in that we are using the python built-in configuration system with a little bit of wrapping for value-setting, which is brittle and inflexible, but having brackets is probably worthwhile, especially since it's breaking an existing use case for @neutrinoceros.

(Looking over the configuration, many of these options are not even necessary anymore.)

But in general, I think we should figure out a way forward, and I'll volunteer taking over reviewing this PR and addressing #2439 and we can spread out some of the reviewing.

neutrinoceros:
So I figure the way to go would be targetting strictly the "str typed" parameters, which should be exactly this list:
- default_colormap
- pluginfilename
- test_data_dir
- notebook_password
- supp_data_dir

And I would update the documentation itself so that `"` and `'` are not used for other parameters.

neutrinoceros:
Alright so the change is now targeted for a subset of parameters. While admittedly safer, it crossed my mind that this approach comes with an additional maintenance cost and could be the source of later "bugs" or inconsistency if someone added a 'str-typed' parameter to the configuration file without updating this condition.
I honestly don't know what's the best option here.

neutrinoceros:
> This does mean that the yt documentation is incorrect and should be changed to remove all single or double quotes, but also the change to `yt/config.py` stripping the quotes should be reverted. Is this acceptable to you?

yes, absolutely. My concern was that the documentation didn't reflect the usage. I'm happy to revert the original commit of this PR !

neutrinoceros:
done !

neutrinoceros:
Thanks everyone involved in the discussion ! I just updated the title to better reflect the end contribution.

brittonsmith:
@neutrinoceros, even the string type parameters should have no quotation marks around them.

neutrinoceros:
You're right, sorry. Should be ok now !

PR #2441, opened 2020-02-05
[setup] return 0 as number of cpus on windows to disable parallel build

Xarthisius:
Minor tweak to PR #2438 , instead of returning `None` in `_get_cpu_count`, return 0. Now it matches default value of `cythonize` and it's more clear in `build_extensions`.


ngoldbaum:
I think you can remove the multiprocessing import too.

PR #2443, opened 2020-02-06
raise a meaningful error on negative and zero zooming factors

neutrinoceros:
I want code that thinks for me. Therefore, I'm proposing to make yt smarter than myself and catch absurd values for zooming factors early rather than downstream.

### DIFF
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -306,7 +306,7 @@ class PlotWindow(ImagePlotContainer):
 
     @invalidate_data
     def zoom(self, factor):
-        r"""This zooms the window by *factor*.
+        r"""This zooms the window by *factor* > 0.
 
         Parameters
         ----------
@@ -314,6 +314,8 @@ class PlotWindow(ImagePlotContainer):
             multiplier for the current width
 
         """
+        if factor <= 0:
+            raise ValueError("Only positive zooming factors are meaningful.")
         Wx, Wy = self.width
         centerx = self.xlim[0] + Wx*0.5
         centery = self.ylim[0] + Wy*0.5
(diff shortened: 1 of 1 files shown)


neutrinoceros:
while I'm at it, I'm also adding the info I was originally looking for to the docstring

yt/visualization/plot_window.py
```diff
@@ -306,14 +306,16 @@ def bounds(self):
...
+        r"""This zooms the window by *factor* > 0.
 
         Parameters
         ----------
         factor : float
             multiplier for the current width
 
         """
+        if factor <= 0:
+            raise ValueError("Only positive zooming factors are meaningful.")
```
munkm: What was your use case for accidentally using negative values in the zoom? I could imagine a user potentially thinking negative values zoom out. This isn't necessary for the PR, but maybe it would be helpful to update our docs on zooming to be explicit that zooming out is defined with values between 0 and 1.

munkm: Ah nevermind I see you fixed it in the update you pushed while I was reviewing things. Thank you!

neutrinoceros: *Blows out smoking gun* ;-)
And to answer your question I did not really have a use case for trying a zero factor. I just got curious and noticed it was not explicitly forbidden. I tried negative values in an attempt to zoom out and my notebook just... crashed, which motivated the PR :)

munkm: 😎 🆒 🥇

munkm: Ah I forgot the 🤠 emoji. That's basically perfect for the documentation-sharpshooting you just did. 😉

munkm:
Ok the appveyor issues should be fixed now! Would you mind merging master into your development branch just so we can check that everything passes? Then I'll be happy to merge this! 🙂 🤠

neutrinoceros:
Sure thing. Actually I'll just rebase all my PR branches.

neutrinoceros:
looks like it's ready to go ! thanks a lot for fixing the numpy regressions !

PR #2444, opened 2020-02-06
deprecate ImagePlotContainer.set_cbar_minorticks 

neutrinoceros:
The idea here is to uniformise the existing Colorbar api by using a single naming convention ('cbar' -> 'colorbar')

Considering the existing methods
- set_colorbar_label
- show_colorbar
- set_cbar_minorticks

I propose to just use "colorbar" everywhere with one change.
I'm keeping the old method for backwards compatibility but I also set a deprecation warning. I don't think this is common practice around here so it's unclear when the "deprecated" method should be removed (I'd suggest yt 4.0 to make it cleaner, but since I'm not sure we'll get a release before this one, it may be a bit of a rush).

### DIFF
--- a/doc/source/reference/api/api.rst
+++ b/doc/source/reference/api/api.rst
@@ -734,6 +734,7 @@ particularly with complicated layouts.
    ~yt.visualization.eps_writer.single_plot
    ~yt.visualization.eps_writer.multiplot
    ~yt.visualization.eps_writer.multiplot_yt
+   ~yt.visualization.eps_writer.return_cmap
    ~yt.visualization.eps_writer.return_colormap
 
 .. _derived-quantities-api:
@@ -804,6 +805,7 @@ See also :ref:`colormaps`.
 
 .. autosummary::
 
+   ~yt.visualization.color_maps.add_cmap
    ~yt.visualization.color_maps.add_colormap
    ~yt.visualization.color_maps.make_colormap
    ~yt.visualization.color_maps.show_colormaps
--- a/doc/source/reference/api/api.rst
+++ b/doc/source/reference/api/api.rst
@@ -734,7 +734,7 @@ particularly with complicated layouts.
    ~yt.visualization.eps_writer.single_plot
    ~yt.visualization.eps_writer.multiplot
    ~yt.visualization.eps_writer.multiplot_yt
-   ~yt.visualization.eps_writer.return_cmap
+   ~yt.visualization.eps_writer.return_colormap
 
 .. _derived-quantities-api:
 
@@ -804,7 +804,7 @@ See also :ref:`colormaps`.
 
 .. autosummary::
 
-   ~yt.visualization.color_maps.add_cmap
+   ~yt.visualization.color_maps.add_colormap
    ~yt.visualization.color_maps.make_colormap
    ~yt.visualization.color_maps.show_colormaps
 
--- a/doc/source/visualizing/plots.rst
+++ b/doc/source/visualizing/plots.rst
@@ -805,7 +805,7 @@ The minorticks may be removed using the
 :meth:`~yt.visualization.plot_window.AxisAlignedSlicePlot.set_minorticks`
 function, which either accepts a specific field name including the 'all' alias
 and the desired state for the plot as 'on' or 'off'. There is also an analogous
-:meth:`~yt.visualization.plot_window.AxisAlignedSlicePlot.set_cbar_minorticks`
+:meth:`~yt.visualization.plot_window.AxisAlignedSlicePlot.set_colorbar_minorticks`
 function for the colorbar axis.
 
 .. python-script::
@@ -814,7 +814,7 @@ function for the colorbar axis.
    ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
    slc = yt.SlicePlot(ds, 'z', 'density', width=(10,'kpc'))
    slc.set_minorticks('all', False)
-   slc.set_cbar_minorticks('all', False)
+   slc.set_colorbar_minorticks('all', False)
    slc.save()
 
 
--- a/doc/source/visualizing/plots.rst
+++ b/doc/source/visualizing/plots.rst
@@ -813,8 +813,8 @@ function for the colorbar axis.
    import yt
    ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
    slc = yt.SlicePlot(ds, 'z', 'density', width=(10,'kpc'))
-   slc.set_minorticks('all', 'off')
-   slc.set_cbar_minorticks('all', 'off')
+   slc.set_minorticks('all', False)
+   slc.set_cbar_minorticks('all', False)
    slc.save()
 
 
--- a/yt/visualization/color_maps.py
+++ b/yt/visualization/color_maps.py
@@ -36,8 +36,8 @@ yt_colormaps = {}
 
 def add_cmap(name, cdict):
     """Deprecated alias, kept for backwards compatibility."""
-    from yt.funcs import mylog
-    mylog.warning("Deprecated alias. Use add_colormap instead.")
+    from yt.funcs import issue_deprecation_warning
+    issue_deprecation_warning("Deprecated alias. Use add_colormap instead.")
     add_colormap(name, cdict)
 
 def add_colormap(name, cdict):
--- a/yt/visualization/color_maps.py
+++ b/yt/visualization/color_maps.py
@@ -34,7 +34,13 @@ def check_color(name):
 
 yt_colormaps = {}
 
-def add_cmap(name, cdict):
+def add_cbar(name, cdict):
+    """Deprecated alias, kept for backwards compatibility."""
+    from yt.funcs import mylog
+    mylog.warning("Deprecated alias. Use add_colormap instead.")
+    add_colormap(name, cdict)
+
+def add_colormap(name, cdict):
     """
     Adds a colormap to the colormaps available in yt for this session
     """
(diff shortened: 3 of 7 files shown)


matthewturk:
THANK YOU for doing this.  I think this is a fine change to make.

neutrinoceros:
Also, I find it weird that it only works with `state == "on"`, which is unusual in python and also unexpected in yt since most comparable methods use the plain boolean `True` to do just this, I'm proposing to accept both for backward compatibility (not sure if the `"on"` value should be deprecated or kept ad vitam)

neutrinoceros:
Wait, I can actually NEVER make it do what it's supposed to do. Maybe it's a subtle mpl backend bug or something but now I'm worried this method may actually be plain broken.

neutrinoceros:
@yt-fido test this please

neutrinoceros:
There's something I'm not getting with the CI results. It seems to me that everything runs smoothly but it's still showing a "pending" status... Apparently this has to do with the build validation. Maybe caused by recent updates in #2448 , ping @munkm you might want to have a look at this :)

munkm:
Hm based on the warnings I'm not sure it's from the changes in 2448, but I'll try to come up with a fix! 

Other PRs have passed with this warning, so travis might just be acting a little weird too. All of the tests are passing, so with enough reviews this should be mergeable even with that pending icon.

munkm:
ok yt-project/yt#2451 fixes the config warnings.

yt/visualization/plot_container.py
```diff
@@ -906,28 +906,43 @@ def set_zlim(self, field, zmin, zmax, dynamic_range=None):
...
+        """turn colorbar minor ticks on or off in the current plot
+
+        Displaying minor ticks reduces performance; turn them off
+        using set_colorbar_minorticks('all', False) if drawing speed is a problem.
+
+        Parameters
+        ----------
+        field : string
+            the field to remove colorbar minorticks
+        state : bool
```
munkm: Is there a reason you changed the API here from a string to a bool? I suspect a new user to yt would probably find it more intuitive to write 'on' or 'off' as an arg to set_colorbar_minorticks().

munkm: I think it would be ok to use the bool here as long as `set_minorticks` is also updated to be consistent.

neutrinoceros: I believe that most yt newcomers are at least familiar with matplotlib, which has a lot of such `set_<stuff>` methods and function arguments using plain boolean to decide whether a specific plot element should be visible or not, so I think it's a good opportunity to change the api to this behaviour.
I don't know which one of us has the most realistic idea of what feels intuitive to a new user, though I agree that the change should be consistently deployed to other methods.

neutrinoceros:
Alright it should be more consistent now, though I think you referred twice the same documentation file by mistake, so there's must be another that need to be adressed.

Also, while browsing documentation I noticed to other methods using "cmap" in there names, so I'm going to deprecate them as well.

neutrinoceros:
The one method I'm not willing to touch is `ImagePlotContainer.set_cmap` because it's very heavily used and would probably feel more like a disturbance to users.

munkm:
Ok I edited the comment to be a little bit clearer. What I meant was that you should update the references with `:meth:~yt.visualization.plot_window.AxisAlignedSlicePlot.set_cbar_minorticks` to the updated function. They link back to the API docs that are autogenerated, so users will be redirected to the correct place. Then also update the examples that use set_cbar_minortics() to also use the new function. The references should be between lines 800 and 820 of that file.

yt/visualization/color_maps.py
```diff
@@ -34,7 +34,13 @@ def check_color(name):
 
 yt_colormaps = {}
 
-def add_cmap(name, cdict):
+def add_cbar(name, cdict):
```
munkm: did you mean to change the name of the deprecated function here?

doc/source/visualizing/plots.rst
```diff
@@ -813,8 +813,8 @@ function for the colorbar axis.
    import yt
    ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
    slc = yt.SlicePlot(ds, 'z', 'density', width=(10,'kpc'))
-   slc.set_minorticks('all', 'off')
-   slc.set_cbar_minorticks('all', 'off')
+   slc.set_minorticks('all', False)
+   slc.set_cbar_minorticks('all', False)
```
munkm:
```suggestion
   slc.set_colorbar_minorticks('all', False)
```

yt/visualization/eps_writer.py
```diff
@@ -1364,16 +1368,16 @@ def return_cmap(cmap=None, label="", range=(0,1), log=False):
         name of the (matplotlib) colormap to use
     label : string
         colorbar label
-    range : tuple of floats
+    crange : tuple of floats
```
munkm: I know this is nitpicky, but I think the `range` of a colormap makes more sense as a kwarg than the `crange` of a colormap. I think these kwargs should be changed back to range.

neutrinoceros:
Thanks for clarifying this. Indeed I missed some references in the .rst doc file. Should be fixed now !

doc/source/reference/api/api.rst
```diff
@@ -734,7 +734,7 @@ particularly with complicated layouts.
    ~yt.visualization.eps_writer.single_plot
    ~yt.visualization.eps_writer.multiplot
    ~yt.visualization.eps_writer.multiplot_yt
-   ~yt.visualization.eps_writer.return_cmap
+   ~yt.visualization.eps_writer.return_colormap
```
munkm: I would keep the old references here too so it's clear they are deprecated functions if somebody uses them and comes across the API docs. Also, the references to the minorticks will automatically include both the deprecated and updated functions so this way we're being consistent.

doc/source/reference/api/api.rst
```diff
@@ -804,7 +804,7 @@ See also :ref:`colormaps`.
 
 .. autosummary::
 
-   ~yt.visualization.color_maps.add_cmap
+   ~yt.visualization.color_maps.add_colormap
```
munkm: ++ my reference to old references here too

yt/visualization/plot_container.py
munkm: > I don't know which one of us has the most realistic idea of what feels intuitive to a new user

I don't really know how to feel about this, but I think both of us are trying to support our users in good faith.

yt/visualization/color_maps.py
neutrinoceros: absolutely not, good catch !

yt/visualization/eps_writer.py
neutrinoceros: Ok. The idea was to avoid using python keywords as variable names but in any case it's out of scope so I'm happy to revert this change.

yt/visualization/plot_container.py
matthewturk: OK, so I have to say, this is a *fascinating* discussion, especially when we look at the history of `PlotWindow` and the like.  In some ways, what we've done is provide a big wrapper around a ton of matplotlib functionality to support a use case that I *think* is reasonably not-extensive, which is the idea that you have multiple fields in a single `PlotWindow` and you apply the same annotations, callbacks, etc, to all of them.  But I think for the most part, people don't do that!

So we're in a weird place where so much of what we've done has been to support a use case that we don't really focus on much in the docs, even, and that includes a fair bit of wrapping of functionality.  This isn't *bad*, and it does provide some nice ways to set defaults, but it's a layer between us and matplotlib.

Anyway, I think we should evaluate this, but I do think that any *major* API changes need to wait for 4.0, which shouldn't actually be too much longer to be `master`.

doc/source/reference/api/api.rst
neutrinoceros: done :)

yt/visualization/plot_container.py
neutrinoceros: To add to this, I personally have mixed feelings about this use case. At the same time, it makes the simple, most repeated task extremely easy to perform, which is precisely what drove me into yt in the first place, but it also makes more refined plots somewhat harder to produce when one means to "talk to matplotlib" trough yt.

Anyway, I'm absolutely fine with delaying these changes to yt-4.0 if you think it's appropriate. However, I'm not sure I'm reading your statement right. Do you consider this a _major_ change ? I guess not, but I'll admit I'm unsure.

neutrinoceros:
Thanks for pointing out this function, I didn't know we had one ! I think I've addressed your concerns now :)

yt/visualization/color_maps.py
```diff
@@ -35,6 +35,12 @@ def check_color(name):
 yt_colormaps = {}
 
 def add_cmap(name, cdict):
+    """Deprecated alias, kept for backwards compatibility."""
+    from yt.funcs import issue_deprecation_warning
```
munkm: I think this import is small enough it can move up to the top of the file with the other import statements.

yt/visualization/eps_writer.py
```diff
@@ -1355,6 +1355,11 @@ def single_plot(plot, field=None, figsize=(12,12), cb_orient="right",
 
 #=============================================================================
 def return_cmap(cmap=None, label="", range=(0,1), log=False):
+    from yt.funcs import issue_deprecation_warning
```
munkm: same with this import

neutrinoceros: done !

Xarthisius:
@yt-fido test this please

yt/visualization/plot_container.py
munkm: I suspect these changes are not major API changes and can be merged before yt-4.0 (though Matt can also weigh in here if I've misinterpreted things). I think this is especially true because you've kept the old functions and have some depreciation warnings when they're called. You've certainly added to the usability of yt and I think this is a helpful contribution. 🙂

matthewturk: I agree!

PR #2447, opened 2020-02-07
add a frame rotating method to datasets (only for polar & cylindrical geometries)

neutrinoceros:
## PR Summary
*experimental, feedback is most welcome*

I'm adding a method to the Dataset class that overrides the on-disk azimuthal coordinates with a given shifting angle. Although it's primarily useful for direct visualisation, I found it also has other applications. For example, as yt automatically adds a `"velocity_cartesian_x"` (and y) derived field to polar datasets, this methods provides a way for the user to redefine the x and y axis as they please.

This function was originally meant for my personal yt plugin, but since it affects parts of the dataset that I believe should not be touched by users (updating the grid edges !), I think it makes sense to contribute it to the shared api.
Another motivation was that I found a lot of *bad* ways to do this, so I think it's best to have one built-in way to do it rather that letting all potential users reinvent it and probably break stuff along the way.

In principle it could be applied to other non-cartesian geometries, but for now I want to know if other devs think it's worth adding to the shared codebase, and if so, I'd have to find relevant ways to test this doesn't break anything.
 

## PR Checklist

   - [x] with docstrings
   - [ ] and narrative docs
- [ ] Adds tests for new features.

## Example application

```python
import numpy as np
import yt
yt.mylog.setLevel(30)

# using radians
assert ds.geometry == "polar"
for i in range(3):
    ds.rotate_frame_by(np.pi/4)
    p = yt.plot_2d(ds, "density")
    p.set_figure_size(3)
    p.save(f"/tmp/rad_{i}.png")

# using degrees
for i in range(3):
    ds.rotate_frame_by(45, deg=True)
    p = yt.plot_2d(ds, "density")
    p.set_figure_size(3)
    p.save(f"/tmp/deg_{i}.png")
```

rad_0
rad_1
rad_2
deg_0
deg_1
deg_2

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1481,10 +1481,11 @@ class Dataset(object):
         """
         rotating_axes = {"polar": 1, "cylindrical": 2}
         if self.geometry not in rotating_axes.keys():
-            raise ValueError("Unsuported geometry %s" % self.geometry)
+            raise ValueError("Unsupported geometry '%s'" % self.geometry)
         raxis = rotating_axes[self.geometry]
+
         if self.dimensionality < raxis - 1:
-            raise ValueError("How did we get here ?")
+            raise ValueError("Unsupported dimensionality '%s' with geometry '%s'" % (self.dimensionality, self.geometry))
         if deg:
             shift_angle = np.deg2rad(shift_angle)
 
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1480,12 +1480,10 @@ class Dataset(object):
         deg : bool, optional
             set to True to use a *shift_angle* in degrees
         """
-        rotating_axes = {"polar": 1, "cylindrical": 2}
-
-        if self.geometry not in rotating_axes.keys():
+        if self.geometry not in ("polar", "cylindrical"):
             raise ValueError("Unsupported geometry '%s'." % self.geometry)
 
-        theta_axis = rotating_axes[self.geometry]
+        theta_axis = self.coordinates.axis_id["theta"]
 
         if self.dimensionality < theta_axis - 1:
             raise ValueError("Unsupported dimensionality '%s' with geometry '%s'." % (self.dimensionality, self.geometry))
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1467,6 +1467,27 @@ class Dataset(object):
         _define_unit(self.unit_registry, symbol, value, tex_repr=tex_repr,
                      offset=offset, prefixable=prefixable)
 
+    def rotate_frame_by(self, shift_angle, deg=False):
+        """
+        (experimental)
+        Shift azimuth coordinates by *shift_angle*, counter-clockwise, and in radians by default.
+
+        Parameters
+        ----------
+        shift_angle : float
+            angle by which coordinates should be shifted
+        deg : bool, optional
+            state that shift_angle is in degrees
+        """
+        if not self.geometry == "polar":
+            raise ValueError("This feature is only available for polar geometries (for now).")
+        if deg:
+            shift_angle = np.deg2rad(shift_angle)
+
+        for edge in [self.index.grid_left_edge[:, 1], self.index.grid_right_edge[:, 1]]:
+            edge += shift_angle * self.length_unit
+            edge %= 2*np.pi * self.length_unit
+
 def _reconstruct_ds(*args, **kwargs):
     datasets = ParameterFileStore()
     ds = datasets.get_ds_hash(*args)
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1491,10 +1491,27 @@ class Dataset(object):
         if deg:
             shift_angle = np.deg2rad(shift_angle)
 
-        theta_edges = [self.index.grid_left_edge[:, theta_axis], self.index.grid_right_edge[:, theta_axis]]
+        # convert to length unit for internal operations compatibility
+        shift_angle *= self.length_unit
+        
+        # only rotate by a multiple of the coarsest angular step
+        # devnote: not sure this is necessary, though it should help making sure
+        # the feature's flow is controlled
+        dx = self.index.select_grids(level=self.min_level)[0].dds
+        dtheta = dx[theta_axis]
+        shift_angle -= shift_angle % dtheta
+
+        theta_edges = [self.index.grid_left_edge, self.index.grid_right_edge]
+
+        #h = self.index.grids[0].index # devnote why is is not the same as ds.index ?
+        #theta_edges += [h.grid_left_edge, h.grid_right_edge]
+
         for edge in theta_edges:
-            edge += shift_angle * self.length_unit
-            edge %= 2*np.pi * self.length_unit
+            edge[:, theta_axis] += shift_angle
+            edge[:, theta_axis] %= 2*np.pi * self.length_unit
+
+        #for g in self.index.grids:
+        #    g._prepare_grid()
 
 def _reconstruct_ds(*args, **kwargs):
     datasets = ParameterFileStore()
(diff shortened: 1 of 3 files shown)


neutrinoceros:
Now works for cylindrical datasets too.
I'm not sure about spherical though, and I don't have a solid dataset to test it on yet so I'd rather leave this to future contributions if need be.

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@matthewturk, any thoughts on this ? Is it worth adding to yt in your opinion ? Is it yt-idiomatic enough ?

neutrinoceros:
I just discovered that this corrupts the underlying data such that `ds.covering_grid` will produce artifacts.

for instance
```python
import numpy as np
import yt

fig, axes = plt.subplots(ncols=3, figsize=(10,10), sharex=True, sharey=True)
increment = np.pi/2/len(axes)
total_rot = 0

for i, ax in enumerate(axes):
    cg = ds.covering_grid(
            level=ds.max_level,
            left_edge=ds.domain_left_edge,
            dims=[128, 128, 1],
            use_pbar=False,
    )
    density_map = cg["density"].squeeze().value
    ax.imshow(density_map, cmap="thermal")
    ax.set_title(f"rot = {total_rot/np.pi:.2f} $\pi$")
    ds.rotate_frame_by(increment)
    total_rot += increment
```
results in
ytrot

where black regions have uniform value "0".

As the definition for covering_grid seems to be obfuscated, I don't know how to inspect this further.

yt/data_objects/static_output.py
```diff
@@ -1456,6 +1456,37 @@ def define_unit(self, symbol, value, tex_repr=None, offset=None, prefixable=Fals
...
+        This method is supported for polar and cylindrical geometries, not for spherical.
+
+        Parameters
+        ----------
+        shift_angle : float
+            angle by which coordinates should be shifted
+        deg : bool, optional
+            set to True to use a *shift_angle* in degrees
+        """
+        rotating_axes = {"polar": 1, "cylindrical": 2}
```
matthewturk: So this is *almost* right, but we sometimes run into issues where we have different axes corresponding to different integer axis indices.

To get the right axis, you'll need to check `ds.coordinates.axis_id['theta']` to get the integer index of theta.

yt/data_objects/static_output.py
```diff
@@ -1456,6 +1456,37 @@ def define_unit(self, symbol, value, tex_repr=None, offset=None, prefixable=Fals
...
+
+        theta_axis = rotating_axes[self.geometry]
+
+        if self.dimensionality < theta_axis - 1:
+            raise ValueError("Unsupported dimensionality '%s' with geometry '%s'." % (self.dimensionality, self.geometry))
+
+        if deg:
+            shift_angle = np.deg2rad(shift_angle)
+
+        theta_edges = [self.index.grid_left_edge[:, theta_axis], self.index.grid_right_edge[:, theta_axis]]
```
matthewturk: Just to check, this is a list, so when you iterate a few lines below, it's just over a list of two items, right?  The left and right edges?

yt/data_objects/static_output.py
```diff
@@ -1456,6 +1456,37 @@ def define_unit(self, symbol, value, tex_repr=None, offset=None, prefixable=Fals
...
+        Parameters
+        ----------
+        shift_angle : float
+            angle by which coordinates should be shifted
+        deg : bool, optional
+            set to True to use a *shift_angle* in degrees
+        """
+        rotating_axes = {"polar": 1, "cylindrical": 2}
+
+        if self.geometry not in rotating_axes.keys():
```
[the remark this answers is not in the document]
neutrinoceros: oh, I forgot dicts behaved this way. Nice !

yt/data_objects/static_output.py
neutrinoceros: Exactly !

yt/data_objects/static_output.py
neutrinoceros: I wasn't aware we had this, or that there would be edge cases. Thanks, this makes this a lot more robust !

neutrinoceros:
> Specifically, what we want -- I _think_ -- is that we want to be able to have the `get_global_startindex()` routines on the grid objects to remain self-consistent.

Very likely so. I'm thinking that overseeing this is what causes the glitches over `ds.covering_grid` I've shown earlier. Yet, I don't understand what the method does or how it does it as I have very little to no experience with the corresponding class (for now :)).

> One other issue I'm slightly concerned about is the case where the grids are rotated by a non-integer multiple of the root grid dx. This is potentially just fine, but may result in some strange issues.

This is a fair concern. From the issues it currently raises, I can see a least one test I could add right now (will do very soon).

chverbeke:
If you want to rotate in spherical coordinates by an angle phi, the following change works:
```
        if self.geometry not in ("polar", "cylindrical", "spherical"):
            raise ValueError("Unsupported geometry '%s'." % self.geometry)

        if self.geometry == "spherical":
            theta_axis = self.coordinates.axis_id["phi"]
        else:
            theta_axis = self.coordinates.axis_id["theta"]
```
I only tested this on my dataset.

neutrinoceros:
Thank you @chverbeke  for this input, I'll use this as soon as I figure out what's wrong with the current code.
For now, I've added a "security" so ensure that the rotation angle is always a multiple of the coarsest corresponding dx. It does not solve the existing issues but maybe it'll help getting there.

neutrinoceros:
I also report that at least some of the underlying problems can be seen with plain fake data

```python
import yt
import numpy as np
ds = yt.testing.fake_amr_ds(fields=["density"], geometry="cylindrical")
ds.rotate_frame_by(np.pi/2)
p.save("/tmp/fake_ds_polar_rot_90")
```
fake_ds_polar_rot_90_Slice_z_density

```python
ds = yt.testing.fake_amr_ds(fields=["density"], geometry="cylindrical")
ds.rotate_frame_by(np.pi)
p.save("/tmp/fake_ds_polar_rot_180")
p.show()
```

fake_ds_polar_rot_180_Slice_z_density

@matthewturk ,  I've been trying to enforce consistency within `get_global_startindex` by updating edges at grid level instead of dataset level (see commented lines). While it does change the results from get_global_startindex, it doesn't change the visualization results presented here. I also tried updating edges at both dataset and grids levels, and it seems redundant: visually, the rotation is doubled in this case.

I'm puzzled.

neutrinoceros:
possibly related, deeper issue (concerning the armvac breakage examples) #2479

matthewturk:
I'm hoping to provide some suggestions, because the more I think about this
the more I think it's a really useful thing to do.

One thing we have explored in the past is having a conversion between "Code
units" for radial angles and "real" radial angles and having this provide
an offset.  (I think that is likely not the way forward but it's something
to think about.)

matthewturk:
Is this still WIP?

neutrinoceros:
definitely. Can't tell when this will be ready to ship, but clearly not in time for 3.6.

neutrinoceros:
I discovered that using `yt.SlicePlot` isn't equivalent to creating a slice object and creating a plot from it with `.to_pw()`. Only in the first case do I get corrupted data from the pixelizer.

I've added two tests to highlight this. The second one is the carbon-copy of the first, except that for each rotation, it produces a SlicePlot, which causes it to fail.

This looks to me like an independent bug in `yt.SlicePlot`.

As an illustration of what I mean, here are the visual results of the end-state of the dataset at the end of the test, as plotted with the two different methods:

**With SlicePlot**
sp_Slice_z_Density


**with slc.to_pw()**
to_pw_Slice_z_Density

neutrinoceros:
Another note is that even the test ending with .to_pw will fail if I remove this line:
```
buff = get_buffer(ds)
```
(and related assertions)

I'm guessing something must be cached the first time `ds.coordinates.pixelize()` is called, though it's not apparent in the pure python part of the code. The more I dig into this, the less I think I understand why it still doesn't work.

neutrinoceros:
I haven't made any progress on this for months now and I don't think I'll find a solution in the near future, so I'm closing for now.

Issue #2445, opened 2020-02-07
numpy-induced bugs caught on apveyor


neutrinoceros:
### Bug report

**Bug summary**

As I reported a few days back on slack, 2 new bugs have emerged with the latest numpy release (1.18.1) and are currently affecting all PRs (e.g. #2440 #2443 #2444 ). I'm reporting them properly here to help us keep track of them. I attempted to fix both and did not succeed thusfar.

**Code for reproduction**

failing tests are the following

- yt.units.tests.test_ytarray.test_ufuncs
Traceback :
```python
Traceback (most recent call last):
3233  File "C:\Miniconda36-x64\envs\test\lib\site-packages\nose\case.py", line 197, in runTest
3234    self.test(*self.arg)
3235  File "C:\projects\yt\yt\units\tests\test_ytarray.py", line 878, in test_ufuncs
3236    unary_ufunc_comparison(ufunc, YTArray([.3, .4, .5], 'cm'))
3237  File "C:\projects\yt\yt\units\tests\test_ytarray.py", line 809, in unary_ufunc_comparison
3238    ret1, ret2 = ufunc(a)
3239  File "C:\projects\yt\yt\units\yt_array.py", line 1378, in __array_ufunc__
3240    out_arr = func(np.asarray(inp), out=out, **kwargs)
3241TypeError: 'out' must be a tuple of arrays
```
I tried to monkey-patch this one but couldn't find an elegant (or any) fix, so I'm letting to someone with more experience with numpy's ufuncs. Note that this error can be reproduced locally with pytest already.

- yt.visualization.tests.test_color_maps.TestColorMaps.test_make_colormap
Traceback :
```python
Traceback (most recent call last):
3247  File "C:\Miniconda36-x64\envs\test\lib\site-packages\numpy\core\function_base.py", line 117, in linspace
3248    num = operator.index(num)
3249TypeError: 'numpy.float64' object cannot be interpreted as an integer
3250
3251During handling of the above exception, another exception occurred:
3252
3253Traceback (most recent call last):
3254  File "C:\projects\yt\yt\visualization\tests\test_color_maps.py", line 50, in test_make_colormap
3255    name='french_flag', interpolate=False)
3256  File "C:\projects\yt\yt\visualization\color_maps.py", line 554, in make_colormap
3257    np.linspace(color[j], next_color[j], interval)
3258  File "<__array_function__ internals>", line 6, in linspace
[...]

PR #2448, opened 2020-02-07
add support for numpy 1.18.1 

munkm:
## PR Summary

This should fix the issues reported in yt-project/yt#2445 caused by updates in our upstream packages. 

For the yt_units errors I followed the same approach taken in yt-project/unyt#115. Thanks @ngoldbaum; your work made this a lot easer for me. 

## PR Checklist

### DIFF
--- a/yt/units/tests/test_ytarray.py
+++ b/yt/units/tests/test_ytarray.py
@@ -828,7 +828,10 @@ def unary_ufunc_comparison(ufunc, a):
 
 
 def binary_ufunc_comparison(ufunc, a, b):
-    out = a.copy()
+    if ufunc in [np.divmod]:
+        out = (a.copy(), a.copy())
+    else:
+        out = a.copy()
     if ufunc in yield_np_ufuncs([
             'add', 'subtract', 'remainder', 'fmod', 'mod', 'arctan2', 'hypot',
             'greater', 'greater_equal', 'less', 'less_equal', 'equal',
@@ -860,7 +863,10 @@ def binary_ufunc_comparison(ufunc, a, b):
         assert_true(not isinstance(ret, YTArray) and
                     isinstance(ret, np.ndarray))
     if isinstance(ret, tuple):
-        assert_array_equal(ret[0], out)
+        assert isinstance(out, tuple)
+        assert len(out) == len(ret)
+        for o, r in zip(out, ret):
+            assert_array_equal(r, o)
     else:
         assert_array_equal(ret, out)
     if (ufunc in (np.divide, np.true_divide, np.arctan2) and
--- a/yt/units/yt_array.py
+++ b/yt/units/yt_array.py
@@ -308,6 +308,8 @@ trigonometric_operators = (
     sin, cos, tan,
 )
 
+multiple_output_operators = {modf: 2, frexp: 2, divmod_: 2}
+
 class YTArray(np.ndarray):
     """
     An ndarray subclass that attaches a symbolic unit object to the array data.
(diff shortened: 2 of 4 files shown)


ngoldbaum:
Thank you for backporting that!

Maybe we should bump the numpy version in `test_requirements.txt` too?

munkm:
You're welcome! 🎉 It was a fun learning experience. 

So we don't set the version for numpy in test_requirements, but I've bumped the version in the travis config file!

yt/units/tests/test_ytarray.py
```diff
@@ -828,7 +828,10 @@ def unary_ufunc_comparison(ufunc, a):
 
 
 def binary_ufunc_comparison(ufunc, a, b):
-    out = a.copy()
+    if ufunc in [np.divmod]:
```
neutrinoceros:
```suggestion
    if ufunc is np.divmod:
```
No strong opinion about this though. I get that your syntax scales better in case we need to add other special cases in the future.

yt/units/tests/test_ytarray.py
```diff
@@ -860,7 +863,10 @@ def binary_ufunc_comparison(ufunc, a, b):
         assert_true(not isinstance(ret, YTArray) and
                     isinstance(ret, np.ndarray))
     if isinstance(ret, tuple):
-        assert_array_equal(ret[0], out)
+        assert isinstance(out, tuple)
+        assert len(out) == len(ret)
+        for o, r in zip(out, ret):
+            assert_array_equal(r, o)
```
neutrinoceros:
```suggestion
            assert_array_equal(o, r)
```
just a matter of consistency in ordering.

yt/units/yt_array.py
```diff
@@ -1370,9 +1372,25 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
                 out_orig = kwargs.pop('out')
-                out = np.asarray(out_orig[0])
+                if ufunc in multiple_output_operators:
+                    outs = []
+                    for arr in out_orig:
+                        outs.append(arr.view(np.ndarray))
+                    out = tuple(outs)
+                else:
+                    out_element = out_orig[0]
+                    if out_element.dtype.kind in ("u", "i"):
```
neutrinoceros:
```suggestion
                    if out_element.dtype.kind in "ui":
```
is equivalent but is arguably harder to read.

ngoldbaum:
Hey Clement,

All of the changes to yt.units are backported from unyt and probably shouldn’t be changed to make future backports (if any) easier.

yt/visualization/color_maps.py
```diff
@@ -548,10 +548,10 @@ def make_colormap(ctuple_list, name=None, interpolate=True):
 
         # Interpolate the R, G, and B channels from one color to the next
         # Use np.round to make sure you're on a discrete index
-        interval = np.round(next_index)-np.round(rolling_index)
+        interval = int(np.round(next_index)-np.round(rolling_index))
```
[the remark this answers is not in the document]
ngoldbaum: I disagree that this is any clearer.

neutrinoceros:
Oh right, my mistake. I did not realise the nature of the backport. Sorry !

munkm:
Ok, I've done the following:
- removed testing on travis for Python 2.7 with newer versions of upstream packages
- retained testing minimal versions for py2.7
- added two new test jobs in travis for minimal versions of both unit tests and answer tests in python 3.6

We do have a test job running python 3.5, which is set to be sundown in September. Do we want to remove this job and add a new one for py 3.8? (note we also don't have tests on travis for py3.7 yet either)

ngoldbaum:
> retained testing minimal versions for py2.7

Why keep these? I don’t think we need to have any python2.7 testing at all.

Dropping 3.5 and adding 3.7 and 3.8 makes sense to me.

That said, I think there only needs to be one minimal dependency version job, with `MINIMAL=1`, that runs on python3.6 with the earliest known supported combination of packages we can easily install on CI.

I don’t think there’s enough marginal benefit to justify the additional complication of doing the minimal tests on multiple python versions, especially now that we’ve dropped python2.7 and don’t have to worry about 2/3 compatibility anymore.

munkm:
> Why keep these? I don’t think we need to have any python2.7 testing at all.

I'm happy to remove them if we decide that's best. I mostly kept them because we could catch backwards incompatibilities with them. But since we're dropping py2 support it makes sense to remove them. If we do, then there will only be two minimal dependency version jobs (for unit tests and answer tests in python3) and the issue with multiple`MINIMAL` definitions will be resolved.

neutrinoceros:
The test_ufuncs test is now failing against python 3.6 (minimal deps run), though you clearly bumped the minimal numpy version to 1.18.1... I'm not getting it.

ngoldbaum:
No, the minimal numpy is 1.12 on 3.6.

@mmunk it might be easiest to only run these multiple-output ufunc tests on sufficiently new numpy. Another option would be to bump the minimum required numpy to match unyt (1.13), those tests seem to pass for unyt on numpy 1.13.

PR #2450, opened 2020-02-10
Add Python 3.7 and Python 3.8 to CI test jobs 

munkm:
## PR Summary

This PR is a progression from the work happening in yt-project/yt#2448. In that PR some of the CI test jobs on travis were changed as a result of updates our code related to numpy compatibility. The highest level of Python that we're testing against on both appveyor and travis right now is 3.6. This PR adds jobs to both appveyor and travis for Python 3.7 and 3.8. It also removes the python 2.7 testing from appveyor. 

(this should wait to be merged until after 2448 is merged.)


munkm:
@yt-fido test this please

munkm:
The environment issue for appveyor on python 3.8 is solved now, so this should be good to merge.

munkm:
Nevermind. The build resolves but the tests aren't running.

neutrinoceros:
I remember you mentioning matplotlib 3.2 failures on appveyor, but couldn't find your statement back. Since the present PR is challenging in itself, I'd suggest we freeze mpl to 3.1 here, for the time being.

munkm:
So the test failures in this particular PR are not stemming from matplotlib 3.2 (yet. I'm sure they will appear if I push an update.). For some reason the Python 3.8 environment isn't running the tests at all and I need to figure out why.

munkm:
@yt-fido test this please

munkm:
Ok I did some digging on the failures on Jenkins. I *think* the issue is that for SciPy v 1.4 there are import errors with Python 3.5 (see here: https://github.com/scipy/scipy/issues/11356) . When SciPy 1.5 is released we can use that (or if we update to Python 3.6). For now I'm bumping the SciPy version back down to 1.3.3.

Xarthisius:
I run into an issue with MPL-3.0.3 and py3.8. TriLinearInterpolation seems to take ages. I recall we had a similar problem when cython/c module responsible for it was compiled in debug mode upstream. I checked that mpl-3.1.3 works just fine. Can we use that version as min requirement for tests?

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

munkm:
Ok, done! I bumped the mpl version from 3.0.3 to 3.1.1 in the`test_requirements` file, not the `test_minimum_requirements` file. Let me know if I should bump another version!

.travis.yml
```diff
@@ -91,6 +91,16 @@ jobs:
       python: 3.6
       script: coverage run $(which nosetests) -c nose_unit.cfg
 
+    - stage: tests
+      name: "Python: 3.7 Unit Tests"
+      python: 3.7
+      script: coverage run $(which nosetests) -c nose_unit.cfg
```
Xarthisius: I think it's sufficient to test only 3.6 and 3.8. Hopefully it's going to be enough and will save us a lot of time.

munkm: ok i've removed py 3.5 and 3.7 from appveyor and travis

matthewturk:
The answer test difference results are here:

http://use.yt/upload/1f30ffb2

Looking these over, they all look quite reasonable, and likely are the result of matplotlib changes.

munkm:
Ok all of the tests pass now, so this should be ready for review!

tests/test_requirements.txt
```diff
@@ -16,12 +16,14 @@ pandas==0.23.4
 pytest~=5.2; python_version >= '3.0'
 pytest~=4.6; python_version < '3.0'
 requests==2.20.0
-scipy==1.1.0
+scipy==1.1.0; python_version < '3.0'
```
cphyc, line 19: Since Python 2 is not supported, is there a reason to keep this?

tests/test_requirements.txt
```diff
@@ -16,12 +16,14 @@ pandas==0.23.4
...
+scipy==1.3.3; python_version >= '3.0'
 sympy==1.5
 pyqt5==5.11.3; python_version >= '3.0'
 thingking==1.0.2; python_version < '3.0'
 pint==0.8.1
-netCDF4==1.4.2
+netCDF4==1.4.2; python_version < '3.0'
+netCDF4==1.5.3; python_version >= '3.0'
```
cphyc, line 26: Same as above, I would advocate to remote the `python_version < '3.0'` line.

.travis.yml
```diff
@@ -48,8 +48,12 @@ install:
     $PIP install --upgrade pip
     $PIP install --upgrade wheel
     $PIP install --upgrade setuptools
+    echo "Travis build stage: $TRAVIS_BUILD_STAGE_NAME"
     # install dependencies yt
-    if [[ $TRAVIS_BUILD_STAGE_NAME != "Lint" ]]; then
+    if [[ ${TRAVIS_BUILD_STAGE_NAME} == "lint" ]]; then
+      export TRAVIS_BUILD_STAGE_NAME="Lint"
+    fi
+    if [[ ${TRAVIS_BUILD_STAGE_NAME} != "Lint" ]]; then
```
cphyc, line 56: Here and below, what is the reason to use `Lint` instead of `lint`?

matthewturk: I suppose it could be either, but basically what seemed to be happening is that while previously Travis had been capitalizing the stage names (i.e., `lint` became `Lint`) for some reason it no longer was in our jobs.  And, because OS X runs bash 3.2, we couldn't use the `${VAR^^}` or `${VAR,,}` construction to change the case as a whole.

So, this is in there in case it switches back.  It could use either one, but we figured that the least invasive one would be to assume the old behavior would return some day.

tests/test_requirements.txt
Xarthisius: We're still running tests for py2.7, just not on Travis. It hasn't been dropped yet.

PR #2452, opened 2020-02-19
Allow PY37 and PY38 in the docs

neutrinoceros:
I noticed that the docs states python 3.5 OR 3.6 was required to install from source. I propose to state that 3.5 is the minimal without further detail instead.

It might be a place where we want to drop mentions of python 2 also.

### DIFF
--- a/doc/source/installing.rst
+++ b/doc/source/installing.rst
@@ -431,7 +431,7 @@ source include:
 
 - ``git``
 - A C compiler such as ``gcc`` or ``clang``
-- ``Python 2.7``, ``Python >= 3.5``
+- ``Python >= 3.5``
 
 In addition, building yt from source requires ``numpy`` and ``cython``
 which can be installed with ``pip``:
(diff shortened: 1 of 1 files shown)


doc/source/installing.rst
```diff
@@ -431,7 +431,7 @@ source include:
 
 - ``git``
 - A C compiler such as ``gcc`` or ``clang``
-- ``Python 2.7``, ``Python 3.5``, or ``Python 3.6``
+- ``Python 2.7``, ``Python >= 3.5``
```
munkm: I think since we're going through and taking out py2 testing, we should just recommend Python >= 3.5 here in the docs.

munkm:
The docs build issue is unrelated to this PR so I'm happy to merge it.

doc/source/installing.rst
neutrinoceros: Sure thing !

neutrinoceros:
there are also conda failures now. 100% unrelated to this PR. Apparently we can't install `conda-build` outside the base env now, which means that, if building the test env involves `conda install` instances, they'll probably need to be replaced with `pip install`. It probably also has other consequences I'm not anticipating at all.

munkm:
I'm working on a fix for the appveyor builds in another PR.

neutrinoceros:
thanks Madicken !

PR #2453, opened 2020-02-19
Improve doc + fix docstring in deposition

cphyc:
## PR Summary

Expands the doc about particle field deposition. Previously, the function `ds.add_deposited_particle_field` was never mentioned in the doc.

It also updates the docstring of the deposition method to match its accepted values.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1254,7 +1254,7 @@ class Dataset(object):
         method : string
            This is the "method name" which will be looked up in the
            `particle_deposit` namespace as `methodname_deposit`.  Current
-           methods include `simple_smooth`, `sum`, `std`, `cic`, `weighted_mean` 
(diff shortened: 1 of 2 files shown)


yt/data_objects/static_output.py
```diff
@@ -1254,8 +1254,8 @@ def add_deposited_particle_field(self, deposit_field, method, kernel_name='cubic
         method : string
            This is the "method name" which will be looked up in the
            `particle_deposit` namespace as `methodname_deposit`.  Current
-           methods include `simple_smooth`, `sum`, `std`, `cic`, `weighted_mean`,
-           `mesh_id`, and `nearest`.
+           methods include `simple_smooth`, `sum`, `std`, `cic`, `weighted_mean`
```
brittonsmith:
```suggestion
           methods include `simple_smooth`, `sum`, `std`, `cic`, `weighted_mean`,
```

PR #2454, opened 2020-02-19
[bugfix] allow code units in two derived fields

brittonsmith:
For the two fields below (`dynamical_time` and `jeans_mass`), conversion to code units was not possible because the constants don't use the unit registry of the dataset and when you do `x = a * b`, `x` has the unit registry of `a`. I fix this issue by moving the constants to after the arrays that use the dataset's unit registry.

### DIFF
--- a/yt/fields/astro_fields.py
+++ b/yt/fields/astro_fields.py
@@ -50,7 +50,7 @@ def setup_astro_fields(registry, ftype = "gas", slice_info = None):
         """
         sqrt(3 pi / (16 G rho))
         """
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Good catch!

ngoldbaum:
If you have time a test that accesses these fields (or maybe all fields? might be harder) and tries to convert to code units would be great.

brittonsmith:
I'm pretty squeezed for time at the moment, but in a few weeks I could look into this. Is there an easy way to check conversion to code units, like a version of `.in_cgs()`?

ngoldbaum:
I think `in_base(“code”)` works iirc

Issue #2446, opened 2020-02-07
Bug in Dataset.find_max() for non-cartesian (polar) datasets


neutrinoceros:
### Bug report

**Bug summary**
using `ds.find_max(field)` returns surprising values for the position array (second return), that do not seem to translate to the actual coordinates of the maximum.

**Code for reproduction**

Here are three different ways to compute an angular position for the maximum density in a polar dataset (amrvac):

```python
from numpy import arctan2
import yt
yt.mylog.setLevel(30)

ds = yt.load("amrvac/bw_polar_2D0000.dat")
assert ds.geometry == "polar"

# first method
ad = ds.all_data()
t1 = ad["theta"][ad["density"].argmax()]
print(t1)

# second method
t2 = ad.argmax("density", axis="theta")
print(t2)

# third method (broken)
x,y,_ = ds.find_max("density")[1]
t3 = arctan2(y, x)
print(t3)
```

**Actual outcome**

```
2.773437264497239 dimensionless
2.773437264497239 dimensionless
1.3679895475969959 dimensionless # <- inconsistent
```

neutrinoceros:
I just discovered that this corrupts the underlying data such that `ds.covering_grid` will produce artifacts.

for instance
```python
import numpy as np
import matplotlib.pyplot as plt
import yt
ds = yt.load("amrvac/bw_polar_2D0000.dat")

fig, axes = plt.subplots(ncols=3, figsize=(10,10), sharex=True, sharey=True)
increment = np.pi/2/len(axes)
total_rot = 0

for i, ax in enumerate(axes):
    cg = ds.covering_grid(
            level=ds.max_level,
            left_edge=ds.domain_left_edge,
            dims=[128, 128, 1],
            use_pbar=False,
    )
    density_map = cg["density"].squeeze().value
    ax.imshow(density_map, cmap="thermal")
    ax.set_title(f"rot = {total_rot/np.pi:.2f} $\pi$")
    ds.rotate_frame_by(increment)
    total_rot += increment
```
results in
![ytrot](https://user-images.githubusercontent.com/14075922/74938352-21a4ed00-53ee-11ea-949e-4b128365b37e.png)

where black regions have uniform value "0".

As the definition for covering_grid seems to be obfuscated, I don't know how to inspect this further.

matthewturk:
[...]

PR #2457, opened 2020-02-20
fix an inconsistency between argmax and argmin methods in YTDataContainer class

neutrinoceros:
## PR Summary

I was working on #2446 and noticed another bug.
Apparently argmin and argmax are heavily code-duplicated from one another, and some convenience functionality of argmax was not ported to argmin, so this PR fixes it.
It would be healthier to deduplicate this code but for now this hotfix should do :)

### DIFF
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -917,6 +917,8 @@ class YTDataContainer(object):
         if axis is None:
             mv, pos0, pos1, pos2 = self.quantities.min_location(field)
             return pos0, pos1, pos2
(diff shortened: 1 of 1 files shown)


matthewturk:
One thing we need to keep an eye on is that in yt 4 we need to switch `string_types` back to `str`.

neutrinoceros:
Good. Will we need to remove six from the repo completely before release ?

matthewturk:
In 4.0 it's removed already -- I think for next in the 3.x series we should
leave it.

PR #2460, opened 2020-02-23
[Bugfix] Invoke _setup_dx in the enzo inline analysis

hyschive:
When using the embedded python analysis in Enzo,  the `grid.ddx` attribute is missing. This PR fixes that.

I'm not sure if there is a built-in answering test in yt for the enzo inline analysis (I guess not since apparently that requires running enzo as well). So I do not add a test here. But I've confirmed that it works at least for a single MPI rank. Using more than one MPI ranks still fails but that's likely a different issue.

## PR Checklist


matthewturk:
This is indeed an oversight -- Thanks for the fix!

PR #2462, opened 2020-02-26
Use numpy.full_like to initialize array in add_contour_fields

blackwer:
## PR Summary

Switches routine that initializes an array by multiplying ones by -1 to using numpy array creation routine that fills the routine on the fly. This is more readable and should typically be marginally faster.

## PR Checklist


matthewturk:
Good catch @neutrinoceros -- I may go through and issue a PR that changes those.  I think it's OK to leave this one just with the one area of concern, though, as I think @blackwer is working on that particular module at the moment.

matthewturk:
And of course I meant to ping @cphyc instead of @neutrinoceros .  It's been a day.

blackwer:
I'd be happy to address other calls to properly fill rather than create and modify, but this actually is unfortunately a bit more subtle than I initially thought, since the `data["index", "ones"]` creates a temporary array on the fly, rather than exposing some existing internal array. My PR is actually slower because it's creating a temporary big array every call and immediately destroying it. I'm currently investigating issues where memory usage (and allocation/deallocation) is a severe bottleneck for us, and this slightly exacerbates that while also increasing CPU usage.

I've duplicated a simpler version of the `_ones` function that eventually gets called here.
```
In [2]: def _ones(data): return np.ones(data.shape, dtype="float64")                                                                                                                                                                           

In [4]: testarr = np.zeros((256, 256, 256), dtype="float64")                                                                                                                                                                                   

In [5]: %timeit _ones(testarr) * -1                                                                                                                                                                                                            
45 ms ± 241 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: %timeit np.full_like(_ones(testarr), -1.0)                                                                                                                                                                                             
67.6 ms ± 232 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [10]: %timeit np.full(_ones(testarr).shape, -1.0)                                                                                                                                                                                           
67.7 ms ± 163 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [11]: %timeit np.full(testarr.shape, -1.0)                                                                                                                                                                                                  
33.8 ms ± 201 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
The proper solution should, probably not surprisingly, just use the shape of the incoming data to generate a new array (much like how `_ones` does). Internally the `_ones` function uses `data.ires.shape` and then modifies the array to have units if _spatial is unset (`_reshape_vals` returns the input array unmodified).
```
    def _ones(field, data):
        if data._spatial:
            return data._reshape_vals(arr)
        return data.apply_units(arr, field.units)
```
I'm not really familiar enough with the project to start fiddling around with the internals of the basic datatypes, but I wonder if it's possible to add an API call to create a new array like the existing internal one with some fixed value. I'd just use the shape directly from ires, but I assume that can generate issues down the road.

matthewturk:
So I've thought more about this, and I *believe* that it works using `ires` because we can't rely on the existence of a "size" for the data object in advance.

So that means the first time it's called, `ires` is generated (and likely cached) and then "ones" is generated and cached.

Let me think about how we might do this otherwise, because there must be a better way.

neutrinoceros:
@matthewturk any thoughts on this ? should we hold it ?

matthewturk:
I think we can/should close it; after conversing about it it seems this is
not the rate-limiting step.

neutrinoceros:
Alright, don't mind if I do.

PR #2463, opened 2020-02-26
try to use pip instead of conda-build for develop install

munkm:
## PR Summary

A number of recent PRs are failing on appveyor because conda-build can't be installed in the conda environment and needs to be installed in the root environment. It looks like we use conda-build only to be able to do a development build of yt with `conda develop -b`. Maybe the appveyor builds can be fixed by using pip instead.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,8 +30,8 @@ install:
     - "python --version"
 
     # Install specified version of numpy and dependencies
-    - "conda install --yes -c conda-forge numpy scipy nose pytest setuptools ipython Cython sympy fastcache h5py matplotlib mock pandas cartopy conda-build pyyaml"
-    - "conda develop -b ."
+    - "conda install --yes -c conda-forge numpy scipy nose pytest setuptools ipython Cython sympy fastcache h5py matplotlib mock pandas cartopy pyyaml"
+    - "pip install -e . "
 
 # Not a .NET project
 build: false


munkm:
The Python 2.7 builds fail, but this is due to the updated numpy support that we added from yt-project/yt#2448. 

yt-project/yt#2450 removes the python 2 testing from appveyor, so I think these failures shouldn't block merging this PR.

neutrinoceros:
- I tried to look up pip's documentation for a replacement of the `-b` flag (build extensions in place) in `conda develop -b .`, but couldn't find one.
- Meanwhile, the build still fails on python 3 with a surprising (to say the least) error
`AttributeError: module 'numpy' has no attribute 'divmod'`

Do you think there is a chance those two points are related ?

neutrinoceros:
update: seems like pip doesn't have an equivalent flag _by design_ and relies on the environment manager to invoke `python setup.py build_ext --inplace` (see https://github.com/pypa/pip/issues/7509)

munkm:
@neutrinoceros the divmod errors are because the numpy version that's being pulled by appveyor is too old (it's 1.11) to be compatible with our changes that added 1.18 support.

munkm:
Ok, it seems that the changes introduced by conda have been remedied and this isn't an issue anymore (thanks @rebeccabilbro for figuring it out!). I retriggered the appveyor tests and we can install yt from source with `pip install` rather than `conda develop`. The failing tests for Python 3.6 are for matplotlib 3.2 (which I'm working on in another PR). 

If anybody feels strongly about using `pip install` instead of `conda develop` in our appveyor scripts I'm happy to reopen this PR. However, I don't think this is a change that needs to be merged.

PR #2466, opened 2020-02-28
extend support of some plot callbacks to polar geometries

neutrinoceros:
## PR Summary

I’m looking to extend support for some of them to non-cartesian geometries.

As an experiment, I’ve just added “polar” to the list of supported geometries in `LinePlotCallback` and `SpherePlotCallback`. It seems to work just fine, though there's one case where the behaviour might not be what the user expects. Namely, with `coord_system="data"`, the start-point and end-point coordinates are passed as cartesian coords, instead of the actual data coordinate system.

For instance
```python
import yt
p = yt.plot_2d(ds, "density")
p.annotate_line([-2, 1, 0], [2, -1, 0], coord_system="data")
```
linepolar_Slice_z_density

This looks fine-ish to me as long as it is documented as such, which seems to me like the way to go, as of now.

## PR Checklist

### DIFF
--- a/doc/source/visualizing/callbacks.rst
+++ b/doc/source/visualizing/callbacks.rst
@@ -85,6 +85,35 @@ of the x-plane (i.e. with axes in the y and z directions):
                     text_args={'color':'black'})
     s.save()
 
+Note that for non-cartesian geometries and ``coord_system="data"``, the coordinates
+are still interpreted in the corresponding cartesian system. For instance using a polar
+dataset from AMRVAC 
+.. python-script::
+
+    import yt
+
+    ds = yt.load("amrvac/bw_polar_2D0000.dat")
+    s = yt.plot2d(ds, 'density')
+    s.set_background_color("density", "black")
+
+    # Plot marker and text in data coords
+    s.annotate_marker((0.2, 0.5, 0.9), coord_system='data')
+    s.annotate_text((0.2, 0.5, 0.9), 'data: (0.2, 0.5, 0.9)', coord_system='data')
+
+    # Plot marker and text in plot coords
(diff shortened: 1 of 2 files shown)


neutrinoceros:
actually I find the behaviour for cylindrical dataset + `coord_system="data"` is quite inconsistent with what I described.

For instance, using the exact same line parameters as above, with a cylindrical dataset, I get a different line.
```python
ds = yt.load("amrvac/bw_cylindrical_3D0000.dat")
p.annotate_line([-2, 1, 0], [2, -1, 0], coord_system="data")

p.save("/tmp/cylindrical_annotate_line")
```

cylindrical_annotate_line_Slice_z_density

So It appears that the coordinates I'm passing are read as "x, z, y", rather counter-intuitively.
I assume this is because `z` appears second in `ds.coordinates.axis_order`.
I wouldn't expect that changing this would break a LOT of user code but eventually, it could, so I'd need to make it clear in the documentation + warnings.

munkm:
We talked about this in the PR triage meeting, and adding a couple of examples to the narrative docs should be good to get this merged. There's no need to modify the underlying cylindrical/polar axes ordering.

neutrinoceros:
Ok, I'm on it !

neutrinoceros:
I just pushed an addition to the documentation. However, I'm confused by the way images are integrated to the final html page, as there's no mention of them in the .rst file. I assumed that they are automatically built and inserted from embedded python script (in which case: neat !).

While I was at it, I also "unlocked" the following callbacks for polar geometries:
- `annotate_text`
- `annotate_marker`

matthewturk:
Once we get the testing in, this will be ready to go, I think.

PR #2471, opened 2020-03-03
[bugfix] use dict.get to pull attributes that may not exist

brittonsmith:
It seems that some old profiles that were saved with `yt.save_as_dataset` did not save the `accumulation` attribute and so loading them fails with a `KeyError`. This fixes that by using the `get` function with an appropriate default. I also took the opportunity to do some small code cleanup in the surrounding lines.

### DIFF
--- a/yt/data_objects/profiles.py
+++ b/yt/data_objects/profiles.py
@@ -415,7 +415,7 @@ class ProfileNDFromDataset(ProfileND):
     An ND profile object loaded from a ytdata dataset.
     """
     def __init__(self, ds):
-        ProfileND.__init__(self, ds.data, ds.parameters["weight_field"])
+        ProfileND.__init__(self, ds.data, ds.parameters.get("weight_field", None))
         self.fractional = ds.parameters.get("fractional", False)
         self.accumulation = ds.parameters.get("accumulation", False)
         exclude_fields = ["used", "weight"]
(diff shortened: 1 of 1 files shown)


yt/data_objects/profiles.py
```diff
@@ -288,7 +288,7 @@ def __getitem__(self, field):
                 fname = self.field_map.get(field[1], None)
                 if fname != field:
                     raise KeyError("Asked for field '{}' but only have data for "
-                                   "field '{}'".format(field, fname))
+                                   "fields '{}'".format(field, list(self.field_data.keys())))
```
brittonsmith: For reviewers, this was a separate bug I found. We were not printing the available fields but just the value of `fname`, which is usually None here.

yt/data_objects/profiles.py
```diff
@@ -416,18 +416,20 @@ class ProfileNDFromDataset(ProfileND):
...
         for ax in "xyz"[:ds.dimensionality]:
             setattr(self, ax, ds.data[ax])
-            setattr(self, "%s_bins" % ax, ds.data["%s_bins" % ax])
-            field_name = tuple(ds.parameters["%s_field" % ax])
-            setattr(self, "%s_field" % ax, field_name)
+            ax_bins = "%s_bins" % ax
+            ax_field = "%s_field" % ax
+            ax_log = "%s_log" % ax
+            setattr(self, ax_bins, ds.data[ax_bins])
+            field_name = tuple(ds.parameters[ax_field])
```
ngoldbaum: Does it make sense to have a default value here too or are `x_field` `y_field` and `z_field` always in `ds.parameters`?

yt/data_objects/profiles.py
```diff
@@ -416,18 +416,20 @@ class ProfileNDFromDataset(ProfileND):
...
+            setattr(self, ax_field, field_name)
             self.field_info[field_name] = ds.field_info[field_name]
-            setattr(self, "%s_log" % ax, ds.parameters["%s_log" % ax])
-            exclude_fields.extend([ax, "%s_bins" % ax,
-                                   ds.parameters["%s_field" % ax][1]])
+            setattr(self, ax_log, ds.parameters.get(ax_log, False))
+            exclude_fields.extend([ax, ax_bins, ds.parameters[ax_field][1]])
```
ngoldbaum: I can verify that the logic is unchanged in the cleanups in this for loop body.

yt/data_objects/profiles.py
ngoldbaum: good catch!

yt/data_objects/profiles.py
brittonsmith: I think this is a more reliable assumption, but I agree it's probably better not to error out with a `KeyError` in the unlikely event. I'll convert that to a get, too.

PR #2474, opened 2020-03-07
find and replace bare excepts

neutrinoceros:
## PR Summary

replace bare excepts to only catch `Exception` subclasses instead of `BaseException`.

`except:` catches `KeyboardInterrupt` while `except Excetption:` doesn't.
See for instance
https://stackoverflow.com/questions/22408098/in-python-why-ever-use-except

I started with the simple `except: pass`, then I extended it to all remaining bare excepts. I did this with automatic find and replace tools (vscode) matching `except:\n\s+pass` then simply `except:`

### DIFF
--- a/doc/helper_scripts/run_recipes.py
+++ b/doc/helper_scripts/run_recipes.py
@@ -45,7 +45,7 @@ def run_recipe(payload):
             cmd = ["python", recipe]
         try:
             subprocess.check_call(cmd)
-        except:
+        except Exception:
             trace = "".join(traceback.format_exception(*sys.exc_info()))
             trace += " in module: %s\n" % module_name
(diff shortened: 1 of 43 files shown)


matthewturk:
The test failures on appveyor are pretty confusing to me.  Are these the same ones that @munkm noticed in another PR and said were not relevant?

neutrinoceros:
Couldn't find her comment back but I think so. They are also showing
on my most recent PRs that are completely uncorrelated with this. If I recall correctly those errors appeared with the latest matplotlib release (3.2).

neutrinoceros:
Indeed, freezing to mpl 3.1 resolves the appveyor issues. See https://github.com/yt-project/yt/pull/2484

PR #2475, opened 2020-03-07
[AMRVAC] add missing min_level

neutrinoceros:
## PR Summary
Very straightforward PR, I noticed that `ds.min_level` was always returning 0 for amrvac datasets, even those where the whole grid is actually refined.


munkm:
The failing tests in this PR are the same that are on master and are unrelated to the changes in this PR. None of the AMRVAC tests appear to be failing.

Issue #2426, opened 2020-01-24
Discussion: make Dataset an abstract class ?


neutrinoceros:
about frontends development in general: the Dataset class, as it stands, is an abstract class as
some of its code relies on unimplemented methods. When we wrote the AMRVAC frontend with @n-claes, especially as we were getting started, it was quite confusing how there wasn't a really explicit edge between what has to be there for the child class `<YourFrontend>Dataset` to work, and what can be left unimplemented (at least a first). Taking inspiration from existing frontends also adds to the confusion in some ways, because most frontends will add code-specific methods, so as a newcomer you have to jump from one to the other in order to grasp which is which.

I was thinking that it could help future contributors to make Dataset an actual abstract class by inheriting from `abc.ABC` . As the exact syntax for doing this seems different in python 2, maybe that could be a change reserved for yt 4.0.

I'm not aware of any good reason why not to do this, though there might very well be. Any input ?

matthewturk:
I strongly support this and would like to participate.

neutrinoceros:
side note: we do have "notes" about this as  _comments_ in the `AMRVACDataset` class, where each method is labelled as required or frontend-specific, which could help.

I do not expect the change itself to be difficult, especially if we give up on python 2.

cphyc:
That would be a very valuable improvement! I had similar issues when working on the RAMSES frontend or the AdaptaHOP one.

neutrinoceros:
So unless I'm mistaken, the most fundamental change required here would be this:

```python
# current
class RegisteredDataset(type):

# proposed
import abc
class RegisteredDataset(abc.ABCMeta):
```

then it's just a matter of defining the abstract methods and adding documentation.
Since the AMRVAC frontend is not yet ported to yt 4.0, I won't try to implement this just yet without a solid continuous integration frame I can use locally

n-claes:
I’m also up for this if needed! It may be useful to note that we’ll have an AMRVAC developers meeting the first week of April which will probably be followed by some frontend modifications (but that will not influence this topic much I guess). As we’re probably going to work on the particle routines I’m planning to extend the frontend to include those features as well, so this discussion may come up again at some point.

Especially since yt 4.0 has a complete overhaul of the particle system (performance improvements etc) if I’m not mistaken? So this may be a good time to first port the frontend over to 4.0 and then work on the abstract class etc as @neutrinoceros mentioned

PR #2478, opened 2020-03-10
A consistent implementation for Dataset.min_level

neutrinoceros:
## PR Summary

this is a follow up to #2475 
~I added a test to the answer test framework to ensure the error corrected in #2475 doesn't show up again in the future. Admittedly the test is in fact necessary but not sufficient, as broken frontends may consistently pass it, unless there's a test dataset that happens to be fully refined.~

edit : the PR's focus was changed to making the Dataset.min_level attribute consistent with the actual value, mimicking the existing Dataset.max_level property

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -263,10 +263,6 @@ class Dataset(object):
         self._set_derived_attrs()
         self._setup_classes()
 
-    @property
-    def min_level(self):
-        return self.index.min_level
-
     def _set_derived_attrs(self):
         if self.domain_left_edge is None or self.domain_right_edge is None:
             self.domain_center = np.zeros(3)
@@ -1428,6 +1424,18 @@ class Dataset(object):
     @max_level.setter
     def max_level(self, value):
         self._max_level = value
+
+    _min_level = None
+    @property
+    def min_level(self):
+        if self._min_level is None:
+            self._min_level = self.index.min_level
+
+        return self._min_level
+
+    @min_level.setter
+    def min_level(self, value):
+        self._min_level = value
 
     def define_unit(self, symbol, value, tex_repr=None, offset=None, prefixable=False):
         """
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Actually I just realised that the previous PR did *not* fix this for the AMRVAC frontend. This is because `ds.min_level` is currently not related to `ds.index.min_level`. The easy fix is to make `Dataset.index` a property, I hope there is no reason against it.

neutrinoceros:
Actually `Dataset.max_level` was *already* a property, so there should not be any issues with making the api more symmetric.

neutrinoceros:
Playing around with https://yt-project.org/docs/dev/cookbook/calculating_information.html#downsampling-an-amr-dataset , I realised that the `Dataset.max_level` setter function can be used to downsample amr datasets, which is a pretty useful feature. However, there doesn't seem to be a way to *up*sample amr, at least currently. Indeed, the setter I added for `Dataset.min_level` is not sufficient to make this work, so I don't think it currently serves any purpose and should probably not be included in this PR.

neutrinoceros:
I removed the test when I realised that AnswerTesting subclasses do not extend as I expected.
The test I was trying to add is neither a "unit test" or an actual "answer test" because it relies on real data but doesn't require comparison with previous answers. Moreover I think that the real issue was in the way the `Dataset.min_level` attribute was initialised, which should now be fixed.

neutrinoceros:
Turns out some tests are failing since I removed the setter, because the attribute is being explicitly set in at least some frontends (including, but not restricted to AMRVAC). I suspect they don't actually need to do it, so I'll try to remove those culprit lines and see if tests pass.

neutrinoceros:
On second thoughts, frontends turn out to be inconsistent with each other, as some will set the attribute in the dataset and then copy the value in the index class, while others do the opposite. 
There may be room for improvement here, in terms of design, but it should be addressed as part of a larger effort (e.g. https://github.com/yt-project/yt/issues/2426)

For now, I'll just restore the setter and hope for the best.

PR #2482, opened 2020-03-15
Extend region expression to 2d datasets

neutrinoceros:
## PR Summary

Currently, `ds.r[...]` only works with 3D datasets (hardcoded limitation). I'm extending it to 2D datasets. As I'm writing this, it stands as a proof of concept because some use cases are not treated yet.

## demo with spherical 2d dataset

```python
ds = yt.load("amrvac/bw_2d0000.dat")

p1 = yt.plot_2d(ds, "density")
p1.set_figure_size(3)
p1.save("/tmp/spherical_alldata")

r = ds.r[:, np.pi/3:2/3*np.pi]
p2 = yt.plot_2d(ds, "density", data_source=r)
p2.set_figure_size(3)
p2.save("/tmp/spherical_azim_selection")

r = ds.r[1.3:2.6, np.pi/3:2/3*np.pi]
p3 = yt.plot_2d(ds, "density", data_source=r)
p3.set_figure_size(3)
p3.save("/tmp/spherical_2d_selection")
```

spherical_alldata_Slice_phi_density

spherical_azim_selection_Slice_phi_density

spherical_2d_selection_Slice_phi_density

## To be clarified

- any idea what sort of test should be added ?
- as of now, I don't understand the purpose of the `nslices` variable in `RegionExpression.__getitem__` so I'm only dealing with the `nslices==2` case.

## to do

### DIFF
--- a/yt/data_objects/region_expression.py
+++ b/yt/data_objects/region_expression.py
@@ -58,9 +58,6 @@ class RegionExpression(object):
             # implicitly.  Note that this happens *after* the implicit expansion
             # of a single slice.
             raise YTDimensionalityError(len(item), self.ds.dimensionality)
-        #if self.ds.dimensionality != 3:
-            # We'll pass on this for the time being.
-        #    raise YTDimensionalityError(self.ds.dimensionality, '3')
 
         # OK, now we need to look at our slices.  How many are a specific
         # coordinate?
--- a/yt/data_objects/region_expression.py
+++ b/yt/data_objects/region_expression.py
@@ -142,8 +142,8 @@ class RegionExpression(object):
         return l, r
 
     def _create_region(self, bounds_tuple):
-        left_edge = self.ds.domain_left_edge
-        right_edge = self.ds.domain_right_edge
+        left_edge = self.ds.domain_left_edge.copy()
+        right_edge = self.ds.domain_right_edge.copy()
         dims = []
         for ax, b in enumerate(bounds_tuple):
             l, r = self._slice_to_edges(ax, b)
--- a/yt/data_objects/region_expression.py
+++ b/yt/data_objects/region_expression.py
@@ -58,9 +58,9 @@ class RegionExpression(object):
             # implicitly.  Note that this happens *after* the implicit expansion
             # of a single slice.
             raise YTDimensionalityError(len(item), self.ds.dimensionality)
-        if self.ds.dimensionality != 3:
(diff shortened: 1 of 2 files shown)


yt/data_objects/region_expression.py
```diff
@@ -136,13 +142,13 @@ def _slice_to_edges(self, ax, val):
         return l, r
 
     def _create_region(self, bounds_tuple):
-        left_edge = []
-        right_edge = []
+        left_edge = self.ds.domain_left_edge
```
matthewturk: So I am pretty sure that we set `domain_left_edge` to be a "read-only" property right now, but it would make me feel a lot better if this were done via copies.

yt/data_objects/region_expression.py
```diff
@@ -136,13 +142,13 @@ def _slice_to_edges(self, ax, val):
...
+        right_edge = self.ds.domain_right_edge
```
matthewturk:
```suggestion
        left_edge = self.ds.domain_left_edge.copy()
        right_edge = self.ds.domain_right_edge.copy()
```

yt/data_objects/region_expression.py
neutrinoceros: you got it !

matthewturk:
@neutrinoceros is this still WIP?

neutrinoceros:
Yes, but maybe I can finish this by tomorrow. I need to make sure nothing weird happens with other geometries.

neutrinoceros:
@matthewturk , so I went back to my notebooks and this actually works fine in cartesian too.
However, in the opening message, I noted two points that I would like answered before I'm confident enough to remove the WIP tag.

neutrinoceros:
Actually I'm removing the WIP flag and adding the yt-3.6 label to make it clear that I'm waiting for reviews rather than actively working on it now.

matthewturk:
Ah, `nslices` I think is just to indicate the number of dimensions to reduce by.  And for tests, I think just testing that it *works* is fine.  You can make a fake 2D dataset and check that it produces same results by slicing the `.r` as it does by making an object.

neutrinoceros:
As we discussed on slack, it turns out that, right now, I don't think there's a working way to create fake 2D datasets. `fake_random_ds` looks pretty easy to extend but it relies on `load_uniform_data`, which seems broken (https://github.com/yt-project/yt/issues/2511).

PR #2483, opened 2020-03-15
Extend Dataset.add_gradient_fields to curvilinear geometries

neutrinoceros:
## PR Summary

Gradient computation is marginally more complex in curvilinear geometries.
Here I give a first sketch of an extension for `Dataset.add_gradient_fields` to `cylindrical` and `polar` coordinates.

Note that, for a `"polar"` (2D) dataset, plotting a meaningful gradient field in the radial direction is currently tricky because of boundary issues. Indeed the two usual workarounds are compromised, respectively because
- it makes little sense to assume/overwrite periodicity in that particular direction
- selecting a radially narrow region with e.g. `ds.r[1:2, :]` is currently not working in 2D, but a fix is in the making here https://github.com/yt-project/yt/pull/2482

I'll add visualisation examples here when this point is solved.

## PR Checklist

- [x] refactoring to minimise duplicated code
- [x] check that the theta gradient makes in practice
- [x] extension to spherical geometries

about the second point: I made a choice in the computation of the theta gradient, where both left and right terms have a common divisor `r` that is the central value. This may not be relevant in practice and I'll check whether or not this is the usual method.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1400,7 +1400,7 @@ class Dataset(object):
         """
         self.index
         if not isinstance(input_field, tuple):
-            raise ValueError
+            raise TypeError
         ftype, input_field = input_field[0], input_field[1]
         units = self.field_info[ftype, input_field].units
         setup_gradient_fields(self.field_info, (ftype, input_field), units)
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1390,13 +1390,17 @@ class Dataset(object):
 
         Examples
         --------
-        with a cartesian dataset ds
+
         >>> grad_fields = ds.add_gradient_fields(("gas","temperature"))
         >>> print(grad_fields)
         [('gas', 'temperature_gradient_x'),
          ('gas', 'temperature_gradient_y'),
          ('gas', 'temperature_gradient_z'),
          ('gas', 'temperature_gradient_magnitude')]
+
+        note that the above example assumes ds.geometry == 'cartesian'
+        In general, the function will create gradients components along the axes of the dataset coordinate system.
+        For instance, with cylindrical data, one gets 'temperature_gradient_<r,theta,z>'
         """
         self.index
         if not isinstance(input_field, tuple):
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -85,6 +85,8 @@ from yt.geometry.coordinates.api import \
     GeographicCoordinateHandler, \
     SpectralCubeCoordinateHandler, \
     InternalGeographicCoordinateHandler
+from yt.geometry.geometry_handler import \
+    is_curvilinear
 
 # We want to support the movie format in the future.
 # When such a thing comes to pass, I'll move all the stuff that is constant up
@@ -1390,6 +1392,7 @@ class Dataset(object):
 
         Examples
         --------
+        with a cartesian dataset ds
         >>> grad_fields = ds.add_gradient_fields(("gas","temperature"))
         >>> print(grad_fields)
         [('gas', 'temperature_gradient_x'),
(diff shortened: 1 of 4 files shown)


neutrinoceros:
Alright I'm now down to a minimalistic implementation that I _think_ should cover all curvilinear cases.

matthewturk:
@neutrinoceros is this still WIP?

neutrinoceros:
This depends on https://github.com/yt-project/yt/pull/2482 !
Because I'm not aware of ways to generate non-cartesian fake datasets, I'm using AMRVAC's example data to test this. Thing is, we currently do not read ghost cells data for AMRVAC's format, so we need to "cheat" and force periodicity in all directions to be able to plot gradients on the whole domain... which is meaningless in the radial direction at least.

This means that we're bound to get numerical artifacts at the radial boundaries, so I would feel more confortable to provide working examples if I could include some 2D slicing (essentially truncating the radial edges).

matthewturk:
Oh!  You can generate fake non-cartesian datasets with:

```
yt.testing.fake_amr_ds( ..., geometry = "spherical")
```

it works for spherical, cartesian (default), polar, cylindrical,
geographic and internal geographic.

neutrinoceros:
Great ! I wasn't aware, this will be of huge help to benchmark basically all of my stuff !

then here's a very simple reproducible script that gives satisfying results (imo)

```python
import yt
for geom in ("cartesian", "cylindrical"):
    ds = yt.testing.fake_amr_ds(fields=["density"], geometry=geom)
    ds.add_gradient_fields(("stream", "density"))

    p = yt.SlicePlot(ds, "z", fields=["density", "density_gradient_magnitude"])
    p.set_cmap("all", "arbre")
    p.set_figure_size(4)
    p.save(f"/tmp/{geom}_grad")
```

## cartesian
cartesian_grad_Slice_z_density
cartesian_grad_Slice_z_density_gradient_magnitude

## cylindrical
cylindrical_grad_Slice_z_density
cylindrical_grad_Slice_z_density_gradient_magnitude


Now this feature is still to be used with caution, for the reasons I already explained, hence the log warning I bundled in already...
Anyway I think this is ready to ship actually :)

yt/fields/fluid_fields.py
```diff
@@ -191,8 +193,10 @@ def _mean_molecular_weight(field, data):
                           weight="cell_mass")
 
 def setup_gradient_fields(registry, grad_field, field_units, slice_info = None):
-    # Current implementation for gradient is not valid for curvilinear geometries
-    if is_curvilinear(registry.ds.geometry): return
+
+    geom = registry.ds.geometry
+    if is_curvilinear(geom):
+        mylog.warning("In %s geometry, gradient fields may contain artifacts near cartesian axes." % geom)
```
munkm, line 199: I'm glad you thought to include this useful logging message here. 👍

yt/data_objects/static_output.py
```diff
@@ -1390,6 +1390,7 @@ def add_gradient_fields(self, input_field):
 
         Examples
         --------
+        with a cartesian dataset ds
```
munkm, line 1393: it might be worth adding a curvilinear example after this one since you're being explicit that this example is cartesian.

neutrinoceros: I agree ! Rather than copy pasting the existing example with a mere 3 characters change, I left a text note.

neutrinoceros:
Done ! I chose however to use a spherical geometry for the test, as it hits a bit more error-prone code (phi-gradients).

yt/fields/tests/test_fields.py
```diff
@@ -233,6 +233,27 @@ def test_add_gradient_fields():
...
+    ds = fake_amr_ds(geometry="spherical")
+    gfields = ds.add_gradient_fields(("stream", "density"))
+    gfields += ds.add_gradient_fields(("index", "ones"))
+    field_list = [('stream', 'density_gradient_r'),
+                  ('stream', 'density_gradient_theta'),
+                  ('stream', 'density_gradient_phi'),
+                  ('stream', 'density_gradient_magnitude'),
+                  ('index', 'ones_gradient_x'),
+                  ('index', 'ones_gradient_y'),
+                  ('index', 'ones_gradient_z'),
```
munkm, line 246: Should these be r, theta, and phi?

munkm:
> I chose however to use a spherical geometry for the test, as it hits a bit more error-prone code (phi-gradients).

That sounds great.

yt/fields/tests/test_fields.py
neutrinoceros: oh absolutely ! thanks !

matthewturk:
Looks great to me!

yt/fields/field_info_container.py
```diff
@@ -398,8 +398,8 @@ def check_derived_fields(self, fields_to_check = None):
         unavailable = []
         fields_to_check = fields_to_check or list(self.keys())
         for field in fields_to_check:
-            mylog.debug("Checking %s", field)
-            if field not in self: raise RuntimeError
+            if field not in self:
+                raise RuntimeError("Could not find field %s" % str(field))
```
brittonsmith, line 402: Since you're fixing this, could you make this a `YTFieldNotFound` exception for consistency?
```suggestion
                raise YTFieldNotFound(str(field))
```

neutrinoceros: Makes a lot of sense, should I thought of this... done ! Thanks !

PR #2484, opened 2020-03-16
[TEST] freezing matplotlib version

neutrinoceros:
## PR Summary

quick attempt to "fix" some failing tests on appveyor by freezing mpl=3.1

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ install:
     - "python --version"
 
     # Install specified version of numpy and dependencies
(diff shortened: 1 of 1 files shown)


neutrinoceros:
ok, appveyor tests are now passing for python3. They break on python2 because mpl 3.1 is not compatible with it, but it doesn't matter since we're also dropping support for it.

neutrinoceros:
I don't understand why the latest attempt failed (conda can't find packages as specified), since I was able to check that the conda install command works locally.

neutrinoceros:
supersede by #2489

PR #2487, opened 2020-03-17
Change order of operations.

matthewturk:
This is necessary so that the primary item in a binary operation in unyt has the necessary entry in the unit registry (specifically code_pressure).


neutrinoceros:
@yt-fido test this please

PR #2490, opened 2020-03-18
[WIP] Accumulators

jcoughlin11:
## PR Summary

This PR adds the ability for yt to integrate any field along an arbitrary path and accumulate the results of this integration along the way.

The tests need to be finished and the code cleaned up to pass flake8, but the core functionality is present.
## PR Checklist

- [ x] New features are documented, with docstrings and narrative docs
- [ x] Adds a test for any bugs fixed. Adds tests for new features.

### DIFF
--- a/yt/visualization/accumulators.py
+++ b/yt/visualization/accumulators.py
@@ -1,0 +1,211 @@
+import numpy as np
+
+from yt.units.yt_array import YTArray
+
+
+def _accumulate_vector_field(path, field_vals):
+    r"""
+    This function integrates the given vector field along the given
+    path p. The integral is done in a piecewise manner
+    (segment-by-segment) so as to be able to store the accumulated
+    values from each of the previous segments.
+
+    For a vector field, the line integral is:
+
+    ..math::
+
+        I = \int_C \vec{a} \cdot d\vec{r}
+
+    where :math:`C` is the path being integrated along, :math:`\vec{a}` is
+    the vector field being integrated, and :math:`d\vec{r}` points along the
+    path :math:`C`. This is equivalent to:
+    
+    ..math::
+
+        \lim_{N \rightarrow \infty} \sum_{p=1}^N \vec{a}(x_p, y_p, ...) \cdot \Delta \vec{r}_p
+        
+    with the understanding that :math:`\Delta \vec{r}_p \rightarrow 0` as
+    :math:`N \rightarrow \infty`.
+        
+    The vector pointing along the segment connecting two adjacent points is:
+    
+    ..math::
+    
+        \Delta\vec{r}_p = (\vec{r}_{p+1} - \vec{r}_p)
+        
+    so the full dot product can be written out as:
+
+    ..math::
+        I = \lim_{N \rightarrow \infty}\sum_{p=1}^N\sum_{i=1}^n a_{p,i}(r_{p+1,i}-r_{p,i})
+        
+    where :math:`n` is the number of dimensions.
+
+    This can be done as a matrix operation. If :math:`\vec{r} \equiv \vec{r}_{p+1} - \vec{r}_p`
+    then the above sum is (leaving out the limit):
+    
+    ..math::
+    
+        I = \vec{a}_1 \dot \vec{r}_1 + \vec{a}_2 \cdot \vec{r}_2 \ldots \vec{a}_N \cdot \vec{r}_N
+        
+    If we write the matrix :math:`A = ([a1x, a1y, ...], [a2x, a2y, ...], ...)`
+    and :math:`R = ([r1x, r1y, ...], [r2x, r2y, ...], ...)^T`, then the dot
+    products in the sum are the diagonal elements of the resulting matrix
+    multiplication :math:`AR`. The accumulation is then obtained by doing a
+    cumsum of this diagonal.
+
+    Parameters
+    ----------
+    p : YTArray 
+        The path to be integrated along
+
+    field_vals : YTArray 
+        An array containing the components of the vector field to be
+        integrated. The values are sampled at the starting point of
+        each path segment as well as the endpoint for the last segment
+
+    Returns
+    -------
+        accum : YTArray 
+            The cumulative value of the field integral at each path
+            segment
+    """
+    accum = np.cumsum(np.diag(np.dot(field_vals[:-1], (p[1:] - p[:-1]).T)))
+    return accum
+
+
+def _accumulate_scalar_field(p, field_vals):
+    r"""
+    This function integrates a scalar field along a path. It uses a
+    similar method to that in _accumulate_vector_field, but the integral
+    is now:
+    
+    ..math::
+    
+        I = \int_C \phi(x1,x2,...,xn)d\vec{r}
+
+    Parameters
+    ----------
+    p : YTArray 
+        The path to be integrated along
+
+    fieldVals : YTArray 
+        An array containing the values of the scalar field to be
+        integrated at the location of the starting point of each
+        path segment as well as the endpoint for the last segment
+
+    Returns
+    -------
+    accum : YTArray 
+        The cumulative value of the field integral at each path
+        segment 
+    """
+    accum = field_vals[:-1] * (p[1:] - p[:-1])
+    # Since this integral results in a vector, I think the accumulation
+    # should add the vectors together. The default for np.cumsum is
+    # axis=None, which performs the cumulative summation along the
+    # flattened array. Axis=0 adds the columns, which is akin to
+    # performing a cumulative summation of each of the components
+    accum = np.cumsum(accum, axis=0)
+    return accum
+
+
+class Accumulators:
+    r"""
+    Container for creating and storing the path integrals of various
+    fields along various paths.
+
+    The class takes in a list of user-defined paths and a dataset. From
+    these, the user can compute the integral of any field in the dataset
+    along the given paths. The results of these path integrals are
+    stored in a cumulative fashion, which means that, once computed, the
+    accumulated value of the field is stored at each point along the
+    path, allowing the user to query this information.
+
+    Attributes
+    ----------
+    pass
+
+    Methods
+    -------
+    pass
+
+    Examples
+    --------
+    pass
+    """
+    def __init__(self, paths, ds):
+        self.paths      = paths
+        self.ds         = ds
+        self.ad         = ds.all_data()
+        self.accum      = []
+
+    def _join_field_components(self, field):
+        r"""
+        This function takes the disparate components of the field,
+        which are stored in the three different arrays fx, fy, and fz,
+        and joins them into one ndarray. The result needs to have a
+        shape of  (nGridCells, nDims)
+
+        Parameters
+        ----------
+        field : iterable
+            Contains the components of the field to be accumulated
+            along the desired paths.
+
+        Returns
+        -------
+        field_grid_vals : np.ndarray
+            An array of shape (n_cells, n_field_components). This form
+            allows for easier matrix manipulation.
+        """
+        # Calling np.stack results in the unit-ful array having
+        # units of 'dimensionless.' It's easier to to just save the
+        # original units, work with the raw values, and then add the
+        # units back in at the end
+        unit = self.ad[field[0]].units
+        # Strip units from coordinates
+        x_field = self.ad[field[0]].d
+        y_field = self.ad[field[1]].d
+        z_field = self.ad[field[2]].d
+        field_grid_vals = np.stack((x_field, y_field, z_field), axis=1)
+        # Add units back
+        field_grid_vals = YTArray(field_grid_vals, unit)
+        return field_grid_vals
+
+    def accumulate(self, field, is_vector=None):
+        r"""
+        This function is the driver function for integrating the desired
+        field along each path in the bundle.
+
+        Parameters
+        ----------
+        field : YTFieldData, tuple
+            The name of the field to integrate along the paths. If this
+            is a vector field, field should be an iterable containing
+            the components of the field (i.e, field[0] is the x-component
+            data, field[1] is the y-component, etc.).
+
+        is_vector : bool
+            If True, field is a vector field. If False, then field is
+            a scalar field.
+
+        Raises
+        ------
+        ValueError
+            If is_vector is not set.
+        """
+        if is_vector is None:
+            raise ValueError("`is_vector` parameter not set.")
+        if is_vector:
+            # Join the components together into a more convenient form
+            field = self._join_field_components(field)
+        # Loop over each path
+        for p in self.paths:
+            # Calculate the values of the field at each point
+            field_vals = ds.find_field_values_at_points(field, p)
+            # Integrate the field along the path based on whether or not
+            # the field is a vector or scalar field
+            if is_vector:
+                self.accum.append(_accumulate_vector_field(p, field_vals))
+            else:
+                self.accum.append(_accumulate_scalar_field(p, field_vals))
--- a/yt/visualization/accumulators.py
+++ b/yt/visualization/accumulators.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 from yt.units.yt_array import YTArray
+from yt.utilities.amr_kdtree.api import AMRKDTree
 
 
 def _accumulate_vector_field(path, field_vals):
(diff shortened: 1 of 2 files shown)


matthewturk:
@jcoughlin11 can you merge with upstream yt-4.0?

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,348 @@
...
+    -------
+        accum : YTArray 
+            The cumulative value of the field integral at each path
+            segment
+    """
+    # https://en.wikipedia.org/wiki/Line_integral
+    # np.dot doesn't combine units correctly and there does not appear to be a
+    # unyt implementation of dot, so units will be handled manually
+    # for the time being
+    accum = np.cumsum(np.diag(np.dot(field_vals[:-1].d, (path[1:].d - path[:-1].d).T)))
```
matthewturk, line 81: `path` is unitless, right?  That's why this works?

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,348 @@
...
+    ..math::
+    
+        I = \int_C \phi(x1,x2,...,xn)d\vec{r}
+
+    Parameters
+    ----------
+    p : YTArray 
+        The path to be integrated along
+
+    fieldVals : YTArray
```
matthewturk, line 101:
```suggestion
    field_vals : YTArray 
```

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,348 @@
...
+    # np.linalg.norm returns a ndarray, so when multiplied by field_vals, which
+    # is a YTArray, this leads to incorrect units. There does not appear to be
+    # a unyt implementation of norm, as far as I'm aware, so units will be
+    # handled manually for the time being
+    accum = np.cumsum(field_vals[:-1].d * np.linalg.norm(p[1:].d - p[:-1].d, axis=1))
+    accum = YTArray(accum, field_vals.units * p.units)
+    return accum
+
+
+def get_row_major_index(ncells, cell_ind):
```
matthewturk, line 122: Is this the same functionality as `np.unravel_index` and `np.ravel_multi_index`?  If so, we might be able to ditch it.

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,348 @@
...
+    fields along various paths.
+
+    The class takes in a list of user-defined paths and a dataset. From
+    these, the user can compute the integral of any field in the dataset
+    along the given paths. The results of these path integrals are
+    stored in a cumulative fashion, which means that, once computed, the
+    accumulated value of the field is stored at each point along the
+    path, allowing the user to query this information.
+
+    Attributes
```
matthewturk, line 179: I think you can get rid of these lines, since you don't fill them out, and generally we don't include stuff in the main docstring for these attributes anyway.

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,348 @@
...
+        self.paths      = []
+        self.ds         = ds
+        self.ad         = ds.all_data()
+        self.accum      = []
+        self.left_edge  = self.ds.domain_left_edge
+        self.right_edge = self.ds.domain_right_edge
+        # Make sure that the path has proper units. If no units are
+        # specified, assume code units
+        for p in paths:
+            if not isinstance(p, YTArray) or p.units == unyt.dimensionless:
```
matthewturk, line 201:
```suggestion
            if not isinstance(p, YTArray) or p.units.is_dimensionless:
```

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,348 @@
...
+        -------
+        path_field_values : np.ndarray
+            The array containing the field values at each point along the
+            path.
+        """
+        node = tree.locate_node(path[idx])
+        # Data is a list, with one element for each field component
+        data = node.data.my_data
+        # Number of cells in the node
+        ncells = data[0].shape
```
matthewturk, line 258: Maybe change the name of `ncells` to something like `dims` to indicate that it's not just number of cells, but cells-by-dimension.

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,348 @@
...
+            left_check = path[idx] < self.left_edge
+            right_check = path[idx] >= self.right_edge
+            if np.sum(np.logical_or(left_check, right_check)):
+                msg = f"Point `{path[idx]}` at index `{idx}` outside domain bounds."
+                msg += f"LE: `{self.left_edge}`, RE: `{self.right_edge}`"
+                raise ValueError(msg)
+            # Figure out which cell in the node the point falls within
+            # Origin of node can be offset from origin of origin of volume,
+            # so we have to subtract it off to get the right cell indices
+            cell_ind = ((path[idx] - node_left_edge) / cell_size).astype(int)
```
matthewturk, line 277:
```suggestion
            cell_ind = ((path[idx] - node_left_edge) / cell_size).astype('i8')
```

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,348 @@
...
+            I = get_row_major_index(ncells, cell_ind)
+            # Get the value of each component for the current cell
+            vals[idx] = data[I] 
+            # See if next point is still within the same node (if the next point
+            # is still in range of the array)
+            idx += 1
+            if idx != npts:
+                left_check = path[idx] < node_left_edge
+                right_check = path[idx] >= node_right_edge
+                if np.sum(np.logical_or(left_check, right_check)):
```
matthewturk, line 290:
```suggestion
                if np.any(left_check | right_check):
```

yt/visualization/accumulators.py
jcoughlin11: Yeah, path is unitless. Looking at it now, this probably isn't a good way of doing it. I assume that the path is in code units. I think I need to handle the units better, as well as make path a unyt array. I'll do that.

yt/visualization/accumulators.py
jcoughlin11: Not sure, as I've never heard of those numpy functions before haha! I'll take a look, and if they are the same then I'll use those! Thanks!

yt/visualization/accumulators.py
jcoughlin11: No, I was wrong. path either has units specified by the user or, if no units are given by the user, then code length units are assumed. The above construction was done because np.dot was removing the units completely and I did not find a unyt implementation of np.dot. As such, I used .d to just access the quantities in the path and field arrays and then manually handled the units. There might be a better way to do that, though.

jcoughlin11: The units are manually added in on line 82

matthewturk:
@jcoughlin11 Can you update answer store, and I'll pursue reviewing this in more detail?

jcoughlin11:
@matthewturk Just did. I also merge with the pytest branch, so everything should be up to date

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,334 @@
...
+    ..math::
+
+        I = \int_C \phi(x1,x2,...,xn)d\vec{r}
+
+    Parameters
+    ----------
+    p : YTArray
+        The path to be integrated along
+
+    fieldVals : YTArray
```
cphyc, line 101:
```suggestion
    field_vals : YTArray
```

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,334 @@
...
+        data = ustack(data, axis=1)
+        # Cell width
+        node_left_edge = self.ds.arr(node.get_left_edge(), "code_length")
+        node_right_edge = self.ds.arr(node.get_right_edge(), "code_length")
+        cell_size = (node_right_edge - node_left_edge) / ncells
+        while idx < npts:
+            # Make sure point is within domain
+            left_check = path[idx] < self.left_edge
+            right_check = path[idx] >= self.right_edge
+            if np.sum(np.logical_or(left_check, right_check)):
```
cphyc, line 256:
```suggestion
            if np.any(left_check | right_check):
```

cphyc (commented):
It would be great to run pre-commit.ci autofix on this one so that the formatting diff disappear.

jcoughlin11:
@cphyc Sorry if this is a dumb question, but how do I run pre-commit.ci on this?

yt/visualization/accumulators.py
jcoughlin11: It's the same as `np.ravel_multi_index` so I'll use that instead.

yt/visualization/accumulators.py
jcoughlin11: Done.

yt/frontends/stream/tests/test_stream_unstructured.py
```diff
@@ -8,31 +8,17 @@ def test_multi_mesh():
...
-        ],
-        dtype=np.int64,
-    )
-    connect2 = np.array(
-        [
-            [1, 2, 3],
-        ],
-        dtype=np.int64,
-    )
+    connect1 = np.array([[0, 1, 3]], dtype=np.int64)
```
matthewturk, line 11: @jcoughlin11 hey, any idea why this got changed?  Did black get run differently?

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+    accum : YTArray
+        The cumulative value of the field integral at each path
+        segment
+    """
+    # https://en.wikipedia.org/wiki/Line_integral
+    # np.linalg.norm returns a ndarray, so when multiplied by field_vals, which
+    # is a YTArray, this leads to incorrect units. There does not appear to be
+    # a unyt implementation of norm, as far as I'm aware, so units will be
+    # handled manually for the time being
+    accum = np.cumsum(field_vals[:-1].d * np.linalg.norm(p[1:].d - p[:-1].d, axis=1))
```
matthewturk, line 115: Why is it `field_vals[:-1].d`?  It kind of seems to me that if this is the case, it should all be offset by the initial value, shouldn't it?

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+    def __init__(self, paths, ds):
+        self.paths = []
+        self.ds = ds
+        self.ad = ds.all_data()
+        self.accum = []
+        self.left_edge = self.ds.domain_left_edge
+        self.right_edge = self.ds.domain_right_edge
+        # Make sure that the path has proper units. If no units are
+        # specified, assume code units
+        for p in paths:
```
matthewturk, line 142: We need to make sure that paths, if it's a list of scalars, is converted to the right dimensionality.  I think the safest thing to do is to ensure (by validating the input) that `paths` is either a `YTArray` or a `np.ndarray`, and if it's a list, it needs to be a list of either scalars (in which case it gets dumped into a `(N,1)` array) or of ndarrays or ytarrays of all the same size.

I believe you can test this by checking if it's an array; if it is, then call `ensure_2d` on it.  If it's not -- if it's a list -- try converting it to a YTArray.

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+            If True, then `field` is a vector and it is assumed that it's
+            an iterable containing the x, y, and z components of the field.
+            Otherwise, the field is assumed to be a scalar field.
+
+        Returns
+        -------
+        tree : AMRKDTree
+            An AMRKDTree for the dataset containing the given field values.
+        """
+        tree = AMRKDTree(self.ds)
```
matthewturk, line 168: Does this recreate the tree every time?  You could cache `tree` as an attribute or property, and just add fields as needed.

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+        use_log = [False for i in range(len(field))]
```
matthewturk, line 169:
```suggestion
        use_log = [False for f in field]
```

But also, doesn't this mean that if `field` is a tuple of field type and field name, it will be two long?  How are you checking that field is a list?

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+        field = [self.ad._determine_fields(f)[0] for f in field]
+        tree.set_fields(field, use_log, False)
+        return tree
+
+    def _get_path_field_values(self, tree, path, idx, vals, npts):
+        r"""
+        Determines the value of the field at each point along the path.
+
+        Parameters
```
matthewturk, line 178: these parameters are out of date -- what are `idx` `vals` `npts`?  I'm not sure I understand this.

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+        data = node.data.my_data
+        # Number of cells in the node
+        dims = data[0].shape
+        # Put the data in a more convenient form: dims x ndims array
+        data = [d.flatten() for d in data]
+        data = ustack(data, axis=1)
+        # Cell width
+        node_left_edge = self.ds.arr(node.get_left_edge(), "code_length")
+        node_right_edge = self.ds.arr(node.get_right_edge(), "code_length")
+        cell_size = (node_right_edge - node_left_edge) / dims
```
matthewturk, line 204: So if it's *vertex* centered data, which I would encourage you to check, this will not be the cell size -- the cell size will be `/ (dims-1)` (that's not valid python code).

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+        while idx < npts:
+            # Make sure point is within domain
+            left_check = path[idx] < self.left_edge
```
matthewturk, line 207: This could be applied outside of the loop.

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+            # Get the value of each component for the current cell
+            vals[idx] = data[I]
+            # See if next point is still within the same node (if the next point
+            # is still in range of the array)
+            idx += 1
+            if idx != npts:
+                left_check = path[idx] < node_left_edge
+                right_check = path[idx] >= node_right_edge
+                if np.any(left_check | right_check):
+                    vals, idx = self._get_path_field_values(tree, path, idx, vals, npts)
```
matthewturk, line 230: It seems to me that this might also be doable outside of the loop.  So do a check if it's inside, and if it's not, call the function.

yt/visualization/accumulators.py
```diff
@@ -0,0 +1,290 @@
...
+        # Number of cells in the node
+        dims = data[0].shape
+        # Put the data in a more convenient form: dims x ndims array
+        data = [d.flatten() for d in data]
+        data = ustack(data, axis=1)
+        # Cell width
+        node_left_edge = self.ds.arr(node.get_left_edge(), "code_length")
+        node_right_edge = self.ds.arr(node.get_right_edge(), "code_length")
+        cell_size = (node_right_edge - node_left_edge) / dims
+        while idx < npts:
```
matthewturk, line 205: I believe this entire loop can be vectorized.

I think what we should do is familiarize ourselves with the behavior -- make sure that we have passing and appropriate tests, that they cover some corner cases, etc, and then turn it into a vectorized loop.  But this definitely can be vectorized, because we're just selecting values here, not computing an integral (until later.)

matthewturk (changes requested):
Looking much better.  I have some comments.

yt/visualization/accumulators.py
jcoughlin11: How do I check that?

yt/visualization/accumulators.py
jcoughlin11: The reason it's not vectorized is because it's recursive, since not every point will be in the same node, and I wasn't sure how to vectorize that.

yt/visualization/accumulators.py
jcoughlin11: I'm not sure I understand how? `path[idx]` is a different point for each value of `idx`

yt/visualization/accumulators.py
jcoughlin11: As above, I'm not sure I understand how, since the checks depend on the value of `idx`

jcoughlin11:
@neutrinoceros Ok, thanks! I'm a bit confused, though, as black runs automatically as a part of precommit whenever I commit locally. If my precommit file is out of sync with that of main, is there no way around this sort of issue?

matthewturk:
@jcoughlin11 you may want to try the rebase or whatever again

neutrinoceros:
⚠️ rebasing with commits like "merge from main" on the branch is a very bad idea if you don't know what you're doing. This is why I advised against merging.

neutrinoceros:
I admit I didn't check for that... @jcoughlin11 , can you give us a post-mortem here so we know what you just did ?

jcoughlin11:
Whoops. ctrl+z!

PR #2493, opened 2020-03-20
[WIP] try fixing appveyor (conda develop -> pip install) ?

neutrinoceros:
## PR Summary
So after @xarthisius successfully fixed the recent issues with appveyor, I’ve rebased most of my branches on master. Two of them (namely #2482 and #2483) are however still failling on appveyor, but the error seems new.  

The failure occurs on py27, where conda develop -b .  fails with 
```bash
WindowsError: [Error 126] The specified module could not be found
597
```

I'm trying `pip install -e .` as a replacement. Let's see what happens.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,9 +8,9 @@ environment:
 
   matrix:
       - PYTHON_VERSION: "2.7"
-        MPL_VERSION: "2.2.4"
+        MPL_VERSION: "2.2.4"  # latest MPL version supporting python 2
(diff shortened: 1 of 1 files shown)


neutrinoceros:
So... apparently appveyor's tests now pass, *but* some py3 projection tests fail.
@munkm , I see you're working on stuff that hopefully is related to this ? Actually I don't understand why the error I'm trying to fix here seemingly didn't show up everywhere yet, so it might not be related at all :/

PR #2494, opened 2020-03-21
[doc] Add existing packages in the documentation

migueldvb:
## PR Summary

yt has been packaged in some platforms. It could be useful to include a list of
packages and versions in those repositories in the installation documentation to
keep track of where the project is packaged.

https://repology.org/project/python:yt/versions

For some reason Gentoo's `sci-visualization/yt` is not in this list but will report it
and hopefully it will get fixed soon.

## PR Checklist

### DIFF
--- a/doc/source/installing.rst
+++ b/doc/source/installing.rst
@@ -318,6 +318,13 @@ Then you can install yt normally, either from the conda-forge channel as above, 
 yt Distribution Packages
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
+.. note::
+
+  Since the third-party packages listed below are not officially supported by
+  yt developers, support should not be sought out on the project mailing lists
+  or Slack channels.  All support requests related to these packages should be
+  directed to their official maintainers.
+
 While we recommended installing ``yt`` with either the ``conda`` or ``pip``
 package managers, a number of third-party packages exist for the distributions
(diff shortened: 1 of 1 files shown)


migueldvb:
Thank you for your comments, I totally understand the point.  Although I cannot
tell for sure for each of them, installing and updating packages with the
package manager on those platforms should work without hiccups because they are
supposed to be functional and the system takes care of all required
dependencies.  So it can be considerably more straightforward to use this
installation method for new people who already use these systems and may not
know that the packages exist.  Regarding the maintenance burden, the packages
generally have a public maintainer that can be contacted in the case there are
any bugs in their build, together with issue trackers and sometimes official
repositories for the packaging efforts.  I don't expect that installation
problems would be reported to ``yt`` upstream maintainers but it could then be
a matter of forwarding to the appropriate place.  I will take the inline
comments into account in the commit and update the PR.

matthewturk:
We did have a couple issues with questions when we were packaged for debian.  I think a straightforward disclaimer would thread the needle and help us move forward on both advertising what's available *and* ensuring that packaging issues were reported where they would be the most useful.

munkm:
@migueldvb if you add the disclaimer that matt and I mentioned I'd be happy to merge this.

migueldvb:
@munkm @matthewturk  I have added a disclaimer under the heading title.

PR #2496, opened 2020-03-22
Fix load_sample function

munkm:
There was an issue with load_sample that a dataset would get loaded, but the ds ended up being a NoneType. The load_sample function wasn't returning the loader, which is fixed in this PR. 

## PR Checklist


zingale:
This fixes the issue I had.  Thanks!

PR #2498, opened 2020-03-24
add an annotate_list method

zingale:
## PR Summary

The `annotate_clear()` method takes an optional index of the annotation to remove, but there
 is not high-level method to see the indices of the currently applied annotations.  This PR adds
an `annotate_list()` method that simply prints the annotations and their indices.

## PR Checklist

### DIFF
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -1119,6 +1119,15 @@ class PWViewerMPL(PlotWindow):
         self.setup_callbacks()
         return self
 
+    def annotate_list(self):
+        """
+        List the current callbacks for the plot, along with their index.  This
+        index can be used with annotate_clear to remove a callback from the
+        current plot.
+        """
+        for i, cb in enumerate(self._callbacks):
+            print(i, cb)
+
     def run_callbacks(self):
         for f in self.fields:
             keys = self.frb.keys()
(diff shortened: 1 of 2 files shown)


zingale:
Note: I also fixed a sphinx error in the annotate_clear title `~` length.

yt/visualization/plot_window.py
```diff
@@ -1119,6 +1119,15 @@ def annotate_clear(self, index=None):
         self.setup_callbacks()
         return self
 
+    def annotate_list(self):
```
munkm, line 1122: Ok, I see that you're matching the `annotate_clear` method in the way that this operates on all annotations. That said, I'm wondering if it's maybe clearer to name it something like `list_annotations`? I feel like that might be more intuitive when looking through different operations one could do on a plot. If I were casually using yt, I'd think that `annotate_list` would be a specific annotation of a list object, just like `annotate_scale` annotates a scale.

zingale: I agree, it would be clearer, and I'm happy to change the name, but I was trying to match the convention already in place.  E.g., `clear_annotations()` would also be more clear than `annotations_clear()`

zingale:
because of some fork problems on my end, I will close and reissue this PR.  Sorry :(

PR #2503, opened 2020-03-24
clarify that the "height" in a disk region is from the midplane

zingale:
## PR Summary

During our yt tutorial today, it came up that the disk region was a little confusing in that the height is measured from the midplane and not top to bottom of the cylinder.  This PR adds a clarification to the region objects overview docs.

## PR Checklist


zingale:
please don't merge yet -- we will use this for the tutorial tomorrow on contributing to yt

PR #2505, opened 2020-03-24
Read new RAMSES info_rt format (compatible with old)

MartinAlvarezSergio:
New RAMSES-RT info files contain information about the RT precision. This can be read in a backwards-compatible manner with a simple try / except.

### DIFF
--- a/yt/frontends/ramses/field_handlers.py
+++ b/yt/frontends/ramses/field_handlers.py
@@ -491,10 +491,10 @@ class RTFieldFileHandler(FieldFileHandler):
                 read_rhs(int)
             # Try to read rtprecision
             try:
-		read_rhs(int)
-		f.readline()
+                read_rhs(int)
+                f.readline()
             except Exception:
-                pass    
+                pass
 
             # Read X and Y fractions
             for _ in range(2):
--- a/yt/frontends/ramses/field_handlers.py
+++ b/yt/frontends/ramses/field_handlers.py
@@ -489,6 +489,11 @@ class RTFieldFileHandler(FieldFileHandler):
             # Read nRTvar, nions, ngroups, iions
             for _ in range(4):
                 read_rhs(int)
+            # Try to read new rtprecision
+            try:
+                read_rhs(int)
+            except Exception:
+                pass                
             f.readline()
 
             # Read X and Y fractions
--- a/yt/frontends/ramses/field_handlers.py
+++ b/yt/frontends/ramses/field_handlers.py
@@ -489,11 +489,15 @@ class RTFieldFileHandler(FieldFileHandler):
             # Read nRTvar, nions, ngroups, iions
             for _ in range(4):
                 read_rhs(int)
-            # Try to read rtprecision
+
+            # Try to read rtprecision.
+            # Either it is present or the line is simply blank, so
+            # we try to parse the line as an int, and if it fails,
+            # we simply ignore it.
             try:
                 read_rhs(int)
                 f.readline()
-            except Exception:
+            except ValueError:
                 pass
 
             # Read X and Y fractions
--- a/yt/frontends/ramses/field_handlers.py
+++ b/yt/frontends/ramses/field_handlers.py
@@ -489,12 +489,12 @@ class RTFieldFileHandler(FieldFileHandler):
             # Read nRTvar, nions, ngroups, iions
             for _ in range(4):
                 read_rhs(int)
-            # Try to read new rtprecision
+            # Try to read rtprecision
             try:
                 read_rhs(int)
+		f.readline()
             except Exception:
-                pass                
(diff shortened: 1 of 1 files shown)


yt/frontends/ramses/field_handlers.py
```diff
@@ -448,7 +448,12 @@ def read_rhs(cast):
             # Read nRTvar, nions, ngroups, iions
             for i in range(4):
                 read_rhs(int)
-            f.readline()
+            # Try to read rtprecision
+            try:
+		read_rhs(int)
+		f.readline()
+            except Exception:
+                pass
```
cphyc: I am a bit concerned about these lines, as the `Exception` is likely to be raised after the line has been read (see line 443). As a consequence, if the code fails to parse the input as `rtprecision`, then you should rewind the line.

Also, why do you read an extra line after calling `read_rhs`? This is only required if the immediate line after `rtprecision` is blank.
I would suggest

```suggestion
            try:
                ipos = f.tell()
                read_rhs(int)
            except Exception:
                f.seek(ipos)
```

MartinAlvarezSergio: Hi Corentin,

Thanks for looking at this. I wanted to make a small push before I start pushing other modifications I've done to my yT that might come useful to other people.

Certainly, I can share a fragment of one of the new files here. The top of a file with the new format looks as follows:
`nRTvar      =         12
nIons       =          3
nGroups     =          3
iIons       =         10
rtprecision =          8
 
X_fraction  =  0.759999990463257E+00
Y_fraction  =  0.239999994635582E+00
 
unit_np     =  0.462706068183299E+04
unit_pf     =  0.693930649464614E+13`

The only difference with the previous format is the addition of an extra line containing an integer for rtprecision. As you mention in the new format, the code will read the new line and then the following one, which is an empty line, so it does work in that instance. Alternatively, in the old version, the exception will be raised when trying to split the empty line, thus there is no need for a second readline. I thought this a simple way to make the update, as it works with both formats, and should be easy to update in the future. I have also tested that it correctly works with both type of files.

cphyc: Oh I see, my bad! I hadn't seen you replaced `f.readline()` by your code snippet.

cphyc: This should fix the tests (replacing tabs with spaces)

```suggestion
            # Try to read rtprecision
            try:
                read_rhs(int)
                f.readline()
            except Exception:
                pass
```

Xarthisius:
@yt-fido test this please

matthewturk:
So I think I talked myself into this being backward compatible, but it did require me thinking things through.  Is this safe, in that we'll always have a new int-parseable line in the format?  And, if so, can we catch the `ValueError` or whatever instead of `Exception` and put a comment in explaining that it'll only get called in the case that the line is blank?

cphyc (dismissed):

munkm:
I'd like to second @matthewturk 's question here.

yt/frontends/ramses/field_handlers.py
```diff
@@ -448,7 +448,12 @@ def read_rhs(cast):
             # Read nRTvar, nions, ngroups, iions
             for i in range(4):
                 read_rhs(int)
-            f.readline()
+            # Try to read rtprecision
+            try:
+                read_rhs(int)
+                f.readline()
+            except Exception:
```
neutrinoceros, line 455:
```suggestion
            except ValueError:
```

cphyc:
> So I think I talked myself into this being backward compatible, but it did require me thinking things through. Is this safe, in that we'll always have a new int-parseable line in the format? And, if so, can we catch the `ValueError` or whatever instead of `Exception` and put a comment in explaining that it'll only get called in the case that the line is blank?

Either the variable `rtprecision` is set (see also https://github.com/yt-project/yt/pull/2505#discussion_r398180381), then the file looks like:
```
nGroups     =          3
iIons       =         10
rtprecision =          8
 
X_fraction  =  0.759999990463257E+00
```
or the variable does not exist and the file reads
```
nGroups     =          3
iIons       =         10
 
X_fraction  =  0.759999990463257E+00
```
In the first case, the line with `rtprecision` is read, split into two, and the r.h.s. is parsed as an int, the code then reads an extra empty line.
In the second case, the code reads the line directly after `ilons=...` (which is empy), and fails at splitting it into two parts raising a ValueError.

Does this answer the question?

neutrinoceros:
I'll close + reopen to refresh tests

matthewturk:
declare_approval

matthewturk (approved):

matthewturk:
@MartinAlvarezSergio thank you for your contribution!

neutrinoceros (approved):

PR #2507, opened 2020-03-27
Bump netcdf4 requirement

matthewturk:
This fixes confusing failures when running the cookbook, as FLASH datasets on py38 seem to cause issues with netCDF4 < 1.5.2.


munkm:
I have a netCDF4 bump to 1.5.3 in github/yt-project/yt#2450.

matthewturk:
So you do!

PR #2508, opened 2020-03-28
[WIP] Initial attempt at refactoring extension building

matthewturk:
This is my first pass at refactoring how we build extensions.

It's not ready for *review*, but it is ready for *feedback* on if it would be helpful or not.  I haven't finished it -- got interrupted about 3/4 the way through, and will revisit it ASAP, but the sketch is along the lines I'm hoping for.


setup.py
```diff
@@ -188,23 +111,38 @@
...
-for ext_name in lib_exts:
-    cython_extensions.append(
-        Extension("yt.utilities.lib.{}".format(ext_name),
-                  ["yt/utilities/lib/{}.pyx".format(ext_name)]))
+ext_info = yaml.load(open("extensions.yml", "r"))
+for lib_collection in ext_info:
+    info = ext_info[lib_collection]
+    libraries = []
+    extra_link_args = []
+    extra_compile_args = []
```
Xarthisius: Can you allow those to be defined as arrays in the yaml too?

matthewturk:
So I have reconsidered, and I am now thinking that we may be able to *just* use `cythonize` as input.  I'm working on implementing a proof of concept of that, but it'll be on yt-4.0 just to minimize disruption.

PR #2513, opened 2020-04-01
Porting #2450 (test job updates) to yt-4.0

matthewturk:
I did my best in porting #2450 to yt-4.0.  Let's see if it succeeds!  There may be additional bits I missed that were not explicitly included in 2450, however.


.travis.yml
```diff
@@ -74,40 +78,29 @@ install:
 
 jobs:
   include:
-    - stage: lint
-      python: 3.7
+    - stage: Lint
+      python: 3.6
```
munkm, line 82: Maybe we should leave the linter at python 3.7 here? It looks like it was bumped up at some point in 4.0's history but not in master.

matthewturk: OK!  I'll let this run and then update.

munkm: On second thought, maybe not since we're testing w/ py 3.6 and 3.8, so why use an odd numbered version?

appveyor.yml
```diff
@@ -9,6 +9,8 @@ environment:
   matrix:
       - PYTHON_VERSION: "3.6"
         NUMPY_VERSION: "1.16"
+      - PYTHON_VERSION: "3.8"
+        NUMPY_VERSION: "1.16"
```
munkm, line 13: should the numpy versions be bumped to 1.18 to match travis?

.travis.yml
munkm: So maybe bump the linter up to 3.8!

matthewturk:
Looks like some errors I need to deal with.  Will do that and push a new set of commits with the updates.

matthewturk:
I removed `NUMPY_VERSION` from `appveyor.yml` as it's not in master, and seems to break appveyor.

matthewturk:
OK, so, it's now failing with 31 actual errors, which I believe are all real. At this point it seems that we're stuck with checking in new answers eventually, so let me check a few things out before proceeding.

matthewturk:
Basically, where this is at: the test differences are real but not significant (pixel-level shifts), but I'm reluctant to check in yet another set of new tests until we figure out how to do so without bloating the repo too much.  I'm hoping this will be resolved through something like `git-lfs`, but until this is discussed on the yt-dev mailing list I'm not sure.

matthewturk:
So the plan now is to see the *current* travis test failures, validate the "answer results," and if they are valid, to commit them to https://github.com/yt-project/answer-store/ in the branch yt-4.0 and re-trigger a build here to make sure they work.

matthewturk:
@Xarthisius I'm not sure I totally get the codecov results, but I *think* maybe it's because we're not testing the same number of versions?  If that's indeed the case, is it OK if we merge this?

Xarthisius:
> @Xarthisius I'm not sure I totally get the codecov results, but I _think_ maybe it's because we're not testing the same number of versions? If that's indeed the case, is it OK if we merge this?

The missing coverage is from not running answer tests on jenkins. Yes, it's ok to merge.

PR #2516, opened 2020-04-02
turn on mathtext for a plot_window colorbar

zingale:
## PR Summary

This is a reissue of PR #2249 in a clean branch (I had done that on my master before and github doesn't let me change the target. 

This change turns on mathtext printing for a plot_window colorbar. This means that any offset text will appear in typeset scientific notation instead of something like 1.e-2. 

See the original PR for a before and after image.

## PR Checklist


zingale:
I think I've updated this all (with @munkm 's expert assistance)

PR #2497, opened 2020-03-23
add first cut at export to matplotlib figure

zingale:
## PR Summary

A common operation is to make a matplotlib figure and grid of axes to display slice plots or projection plots, e.g., as shown in this cookbook recipe:

https://yt-project.org/docs/dev/cookbook/complex_plots.html#multipanel-with-axes-labels

This adds a method to `PWViewerMPL` to automate a lot of that setup.  E.g., you would now do that recipe as:
```

import yt

ds = yt.load_sample("IsolatedGalaxy")

fields = ['density', 'velocity_x', 'velocity_y', 'velocity_magnitude']
p = yt.SlicePlot(ds, 'z', fields)
p.set_log('velocity_x', False)
p.set_log('velocity_y', False)

fig = p.export_to_mpl_figure((2,2))
fig.tight_layout()

fig.savefig("test.png")
```

## PR Checklist

- [ ] Code passes flake8 checker
- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #2517, opened 2020-04-03
hotfix: bug in cylindrical coordinates background color with log norm

### LINKED ISSUE #2480: Background not transparant for plotting in spherical coordinates (no log).
When plotting a slice, the parts of the plot that do not belong to the dataset are usually plotted to be transparant (or you can set the background color yourself).
However, when using spherical coordinates and not have a log_scale, this is not the case and the background is always set to lowest value of the colorbar. This is the case whether you pick the limits of your colorbar yourself or not, and also whether you set the background color yourself or not.

```
import yt
ds_fake = yt.testing.fake_amr_ds(geometry='spherical')
slice_fake_1 = yt.SlicePlot(ds_fake, axis='theta', fields='Density')
slice_fake_1.show()
slice_fake_2 = yt.SlicePlot(ds_fake, axis='theta', fields='Density')
slice_fake_2.set_log('Density', False)
slice_fake_2.show()
```
spherical_1
spherical_2

neutrinoceros:
## PR Summary

Fix #2480 
@matthewturk was **one** character away from this fix in the embedded discussion.
I spent an unreasonable amount of time trying to find _where_ I could write a workaround the matplotlib warning triggered by using `np.nan` as a filler value, but in the end I discovered that, while `-np.inf` indeed breaks everything (as noted by @chverbeke)  `np.inf` will do the job perfectly.
Issue solved, no warning, no problem 👏 

For closure, I tested this using
- matplotlib 3.1.3
- 1.18.1

### DIFF
--- a/doc/source/developing/testing.rst
+++ b/doc/source/developing/testing.rst
@@ -402,7 +402,7 @@ Here is an example test function:
        yield test
 
 .. note:: The inner function ``create_image`` can create any number of images,
-          as long as the corresponding filenames conform to the prefix.
+   as long as the corresponding filenames conform to the prefix.
 
 Another good example of an image comparison test is the
 ``PlotWindowAttributeTest`` defined in the answer testing framework and used in
--- a/doc/source/developing/testing.rst
+++ b/doc/source/developing/testing.rst
@@ -387,17 +387,10 @@ Here is an example test function:
    def test_my_ds():
        ds = data_dir_load(my_ds)
 
-       def create_images(filename_prefix):
-            plot = yt.SlicePlot(ds, "z", "density")
-            plot.save("%s_log" % filename_prefix)
-
-            plot.set_log("density", False)
-            plot.save("%s_lin" % filename_prefix)
-
-            plt.plot([1, 2], [1, 2])
-            plt.savefig("%s_lineplot" % filename_prefix)
-
-       test = GenericImageTest(ds, create_images, 12)
+       def create_image(filename_prefix):
+           plt.plot([1, 2], [1, 2])
+           plt.savefig("%s_lineplot" % filename_prefix)
+       test = GenericImageTest(ds, create_image, 12)
 
        # this ensures the test has a unique key in the
        # answer test storage file
@@ -407,7 +400,8 @@ Here is an example test function:
        test_my_ds.__name__ = test.description
 
        yield test
-.. note:: The inner function ``create_images`` can create any number of images,
+
+.. note:: The inner function ``create_image`` can create any number of images,
           as long as the corresponding filenames conform to the prefix.
 
 Another good example of an image comparison test is the
--- a/doc/source/developing/testing.rst
+++ b/doc/source/developing/testing.rst
@@ -387,10 +387,17 @@ Here is an example test function:
    def test_my_ds():
        ds = data_dir_load(my_ds)
 
-       def create_image(filename_prefix):
-           plt.plot([1, 2], [1, 2])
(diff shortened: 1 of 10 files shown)


neutrinoceros:
Potentially controversial: because some existing regression tests was already "wiping out nans" before comparison (under the assumption that fans were filler values, I assume), I took the liberty to also wipe out infs to make it pass.

neutrinoceros:
The idea is to check that doing this change doesn't break any existing test. In that case I _think_ it should be okay to do.

as a note, I missed an opportunity to use `~np.isfinite` which is the exact equivalent of `np.isnan | np.isinf`

neutrinoceros:
@munkm @matthewturk so I tried to include image comparison tests (I think _this_ time Jenkins will run them), hopefully I didn't get it _too_ wrong.

In order to keep copy-pasta under control, because I have 8 actual plots, I grouped them into 4 + 4, so there are really just two tests, but the code therein could be _much_ shorter if we merge #2497 first :)

neutrinoceros:
So the tests do run, but fail.
I note that [the doc](https://yt-project.org/docs/dev/developing/testing.html#how-to-write-image-comparison-tests) explicitly states

> You must decorate your test function with requires_ds, otherwise the answer testing machinery will not be properly set up.

which is irrelevant here (I think ?) because we're testing against fake data.

There's also the `requires_anser_testing` decorator which seems relevant, but I'm not sure it applies here.

neutrinoceros:
I think this time I got the hang of it: I was able to run the new answer test locally. Fingers crossed.

neutrinoceros:
The test returned a “no old answer” error. The documentation mentions that this is what happens when new tests are added to an existing set of answers, though I don't think it is expected when the set is brand new. I think what happened is that I got Jenkins all mixed-up by first declaring the set of answers and declaring the test's name only a few commits later.
If I'm reading the doc correctly, it should now be ok to "bump" the set's number, as if I was extending an existing set.

doc/source/developing/testing.rst
```diff
@@ -400,6 +407,8 @@ Here is an example test function:
        test_my_ds.__name__ = test.description
 
        yield test
+.. note:: The inner function ``create_images`` can create any number of images,
```
munkm, line 410: I think this is a good note to add to the documentation

doc/source/developing/testing.rst
```diff
@@ -387,10 +387,17 @@ Here is an example test function:
...
-           plt.plot([1, 2], [1, 2])
-           plt.savefig(filename_prefix)
-       test = GenericImageTest(ds, create_image, 12)
+       def create_images(filename_prefix):
+            plot = yt.SlicePlot(ds, "z", "density")
+            plot.save("%s_log" % filename_prefix)
+
+            plot.set_log("density", False)
+            plot.save("%s_lin" % filename_prefix)
+
```
munkm, line 396: I don't think that this change really enhances this part of the documentation. are you trying to show that you can have multiple yt plot objects or something?

neutrinoceros: Yes that was the idea, because I only grasped it from reading the source code. But since you think the note is enough I will revert this.

munkm: I personally think the note is helpful. Thank you for reverting it! I think we should have a community discussion about doc improvements after the 3.6 release; you've had a lot of good feedback about the docs and I think it warrants a bigger discussion!

munkm:
Thanks so much for this hotfix @neutrinoceros. You rock! 🥇 💯

PR #2526, opened 2020-04-10
Add changelog entry for 3.6, merge stable into master

munkm:
This is in preparation for the 3.6 release.

PR #2527, opened 2020-04-10
bump dev version on master

munkm:
This should go in after yt-project/yt#2526 is merged


munkm:
@matthewturk should the number after "dev" be bumped like I've done or is that unnecessary?

matthewturk:
I think it's fine as-is.

munkm:
@Xarthisius I can't see why the docs were failing on this PR? Did I increment the version in a way that broke it?

Xarthisius:
> @Xarthisius I can't see why the docs were failing on this PR? Did I increment the version in a way that broke it?

It was probably a random failure. Ignore it.

setup.py
```diff
@@ -34,7 +34,7 @@
 except pkg_resources.DistributionNotFound:
     pass  # yay!
 
-VERSION = "3.6.dev0"
+VERSION = "3.6.dev1"
```
Xarthisius, line 37: Since we released 3.6 shouldn't it be `3.7.dev0` ?

munkm: wow. Yes. Thank you.

PR #2531, opened 2020-04-11
Adding workflow for enforcing PR labels

matthewturk:
This adds a configuration file for the "Enforce PR Labels" github action:

https://github.com/marketplace/actions/enforce-pr-label

This action is not yet enabled, but this is the necessary config for it.  I have configured it to require one of "bug", "enhancement", "new feature" on all pull requests for them to merge.


munkm:
I think you should add a docs label to the list too. We've had several docs-only PRs and I think it'd be worth separating them from "enhancements".

matthewturk:
Marvelous idea.

matthewturk:
We have a bunch of "docs" labels so I added all of them.  Might be better if we reduced the number of labels, or just added a single one that was just "documentation", but, NBD.

matthewturk:
I'm actually going to see if I can work up a config for https://github.com/mergeability/mergeable instead...

munkm:
I had another thought -- we should probably add another label to the list. I think the "infrastructure" label could be good for the PRs related to version bumps, CI fixes, and things like that. 

Another thing that would be nice with this bot is to enforce approving reviews differently depending on the labels. That is, PRs with the "docs" label will only require one approving review, while "enhancements" and other labels require two. 

> We have a bunch of "docs" labels so I added all of them. Might be better if we reduced the number of labels, or just added a single one that was just "documentation"

Hmmm this is a good point. I think we chose to make these three in the workshop last year so it'd be clear where fixes were going. I like the granularity, but does it actually give us anything?

Another point that will be nice here is if we start enforcing PR labelling we can do some interesting analysis every release on where our features are going, and where we're spending most of our developer time.

matthewturk:
Awesome ideas.  Can we consolidate the existing labels and then delete the old ones, moving forward with just `docs` or `documentation`?  I will then update this PR.  And, I'll add `infrastructure` as well.

matthewturk:
OK, so, it's now failing with 31 actual errors, which I *believe* are all real.  At this point it seems that we're stuck with checking in new answers eventually, so let me check a few things out before proceeding.

matthewturk:
OK, I re-did this on top of `master` instead of `yt-4.0`, as per offline discussion.

matthewturk:
AppVeyor didn't build because I made a silly push, but it does work, I promise.  ;)

brittonsmith:
I'm ok with this being merged.

.github/mergeable.yml
```diff
@@ -0,0 +1,13 @@
+# config file for mergeable: https://github.com/mergeability/mergeable
+version: 2
+mergeable:
+  - when: pull_request.*
+    validate:
+      - do: title
+        must_exclude:
+          regex: ^\[WIP\]
+          message: "WIP pull requests can't be meged."
```
cphyc, line 9:
```suggestion
          message: "WIP pull requests can't be merged."
```

.github/mergeable.yml
```diff
@@ -0,0 +1,13 @@
+# config file for mergeable: https://github.com/mergeability/mergeable
+version: 2
+mergeable:
+  - when: pull_request.*
+    validate:
+      - do: title
+        must_exclude:
+          regex: ^\[WIP\]
```
cphyc, line 8: This is going to miss titles like e.g. `WIP: Something`. Maybe you could instead match
```suggestion
          regex: ^\[?WIP\b
```
It matches the first three but not the last three
```
WIP: stuffing stuffy stuff # Matched
WIP: test                  # Matched
[WIP] Foo bar baz          # Matched

# None of the following match
WIMPs or Dark Matter?
WIPING STUFF OUT
wip: asdf movies
```

PR #2534, opened 2020-04-12
Allow for local plugin file (not system wide)

neutrinoceros:
## PR Summary

As a user, I find the plugin feature of yt too restrictive as it currently allows to have only *one* plugin file per system, which makes it a unsuited for more local definitions. Say I have a project-specific-folder full of notebooks to analyse closely related datasets, then it makes sense to me to have companion a plugin file within the same folder. It would make it easier for me to work with collaborators if we simply share the notebooks' folder, and none of us would have to worry about collisions with our own private, system-wide plugins.

This change makes this use-case possible so that a user can organise a folder such as:
```bash
.
└── notebooks
    ├── yt_pluging.py
    └── notebook.ipynb
```

where `notebook.ipynb` would open as
```python
import yt
yt.enable_plugins("yt_pluging.py")
```

## PR Checklist

### DIFF
--- a/doc/source/reference/configuration.rst
+++ b/doc/source/reference/configuration.rst
@@ -133,7 +133,8 @@ To enable a plugin file, call the function
 System wise plugin file
 ^^^^^^^^^^^^^^^^^^^^^^^
 yt will look for and recognize the file ``$HOME/.config/yt/my_plugins.py`` as a
-plugin file.
+plugin file. It is possible to rename this file to ``$HOME/.config/yt/<pluginfilename>.py``
+by defining ``pluginfilename`` in your ytrc file, as mentioned above.
 
 .. note::
 
--- a/doc/source/reference/configuration.rst
+++ b/doc/source/reference/configuration.rst
@@ -124,8 +124,8 @@ Plugin Files
 ------------
 
 Plugin files are a means of creating custom fields, quantities, data objects,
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Reviewing myself, I realised there was a niche case where this would have crashed. Namely, if the specified plugin file path was rooted in the old config dir (`~/.yt`). I refactored so this wouldn't be an issue. Sorry @matthewturk for pushing after approval !

doc/source/reference/configuration.rst
```diff
@@ -120,30 +120,41 @@ used internally.
...
-objects, colormaps, and other code classes and objects to be used in future
+Plugin files are a means of creating custom fields, quantities, data objects,
+colormaps, and other code executable functions or classes to be used in future 
 yt sessions without modifying the source code directly.
 
-To force the plugin file to be parsed, call the function
+To enable a plugin file, call the function
 :func:`~yt.funcs.enable_plugins` at the top of your script.
 
+System wise plugin file
```
munkm, line 133: I think it's worth changing this section header to "global system plugin file", since the next section you have is about local plugin files and there's narrative continuity.

neutrinoceros: done !

neutrinoceros:
I saw a bunch of logic simplifications and applied them, the logic tree was a bit redundant.

PR #2537, opened 2020-04-13
[WIP] Use tox for managing environments

matthewturk:
After #2513, #2450 and a couple others, I'd like to propose that we explore simplifying our usage of travis.  Instead of having a large `.travis.yml` file, I think simplifying to use `tox.ini` would be prudent.

This is a rough draft of a very simple tox ini file.  Using it with travis would look something like what's outlined in [this blog post](https://medium.com/@alejandrodnm/testing-against-multiple-python-versions-with-tox-9c68799c7880), and would reduce the travis config file in size -- and probably help us with migrating to something like Circle CI or Github Actions in the future.

This PR will be a staging ground for replicating the travis configuration in tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,17 @@
+# tox (https://tox.readthedocs.io/) is a tool for running tests
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox" from this directory.
+
+[tox]
+envlist = py35, py36, py37, py38
+
+[testenv]
+deps =
+    -rtests/test_minimal_requirements.txt
+    cython 
+setenv =
+    MINIMAL = 1
+
+commands =
+    nosetests -c nose_unit.cfg


matthewturk:
(It may also end up being the case that this gets us nothing, in which case, nothing ventured, nothing lost.)

neutrinoceros:
I note that we can avoid the extra file now that we have a pyproject.toml
(https://tox.readthedocs.io/en/latest/example/basic.html)

matthewturk:
I kind of felt ambivalent about this, too. I am not sure it solves anything
for us but maybe I'm wrong.

neutrinoceros:
well if the goal is purely to simplify our usage of Travis I think the solution might just be to migrate to GH workflows for unit tests. And from what I understand tox doesn't add anything there.

matthewturk:
Sounds good to me.  Feel free to close it, unless there's a good reason not
to!

neutrinoceros:
I think it's ok to close for now :)

PR #2540, opened 2020-04-14
Initialize git LFS for answer-store directory in 4.0

matthewturk:
This initializes [git LFS](https://git-lfs.github.com/) for the 4.0 `answer-store` directory.

Note that, as of right now, I *believe* it's not converting any files that are already in there to LFS.  That would happen in a subsequent commit.


matthewturk:
I've updated the docs to include a note about this.

If this is accepted, the roadmap would be to issue a second PR that moves the yt 4.0 answers *into* LFS and simultaneously enables LFS checkouts on Travis.  That should demonstrate the functionality.

matthewturk:
Turns out I might misunderstand git LFS and so let's hold on this!

matthewturk:
Closing this in favor of an alternate, non-LFS approach.

matthewturk:
See #2545 .

PR #2541, opened 2020-04-14
[AMRVAC] bugfix geometry parsing

neutrinoceros:
hotfix a bug in amrvac frontend where 'default' geometry would cause a ValueError to be raised.


neutrinoceros:
So... something like this ?

matthewturk:
I like it!

neutrinoceros:
I don't think it's well known. I just discovered it myself and I _think_ that prior to the yt frontend, almost nobody in the community was using the native data files for analysis anyway... well, I'm adding the note then 😅

PR #2545, opened 2020-04-15
Put answers in a subrepo

matthewturk:
This PR puts our `answer-store` directory in a submodule.  Unless someone clones `--recursive`, by default they won't get the full history of the answer tests.

One open question I have is if I can or should do anything to suggest a depth of the clone; from what I can tell, specifying `--depth` in the `git submodule add` command wouldn't change anything for folks cloning the repo, but I may be misreading.

Travis should clone by default all the submodules, so let's see how it goes...


matthewturk:
@munkm suggests this go on 4.0 -- I will leave this open and issue a new PR too.

PR #2546, opened 2020-04-15
Put answers in submodule in yt-4.0

matthewturk:
This is the same as #2545, but for yt-4.0.

I re-used the git module from my master checkout, so hopefully this will be fine!

What I'm planning to do is see if Travis complains -- it should about *wrong* answers, but not *missing* answers -- and then if it does, update the answers in https://github.com/yt-project/answer-store/ in the yt-4.0 branch, then update here.


munkm:
Ok, so the tests are failing at the linter on Travis because this branch doesn't have the updates from  https://github.com/yt-project/yt/pull/2513. Ugh. @matthewturk are the tests that this is pointing to identical to the ones that already exist in answer-store on yt-4.0? Perhaps we can merge this one now, update 2513, and (if needed) do a separate PR to update the tests in `answer-store` to fix failures. What do you think?

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "answer-store"]
+	path = answer-store
+	url = https://github.com/yt-project/answer-store
+	branch = yt-4.0
```
munkm, line 4: +1 to using a separate branch for these tests. Then we can have updates for tests on master too and after the 4.0 merge we won't have annoying history conflicts. 

Maybe we'll also need a 4.0 merge PR for the answer store then, too? 😉

matthewturk:
I think that's *exactly* what we should do!

.gitmodules
matthewturk: Yes, I think so!

PR #2549, opened 2020-04-18
add explicit support for jpeg exports

neutrinoceros:
With recent enough versions of matplotlib and if PIL (pillow) is installed, matplotlib is able to export to jpeg.
When it's not, it raises perfectly sensible errors such as
```
ValueError: Format "jpg" is not supported.
Supported formats: eps, pdf, pgf, png, ps, raw, rgba, svg, svgz.
```

Thus, I don't think there's a good case for preventing exporting to jpg within yt, which is currently the case because of how `get_image_suffix` is implemented.

### DIFF
--- a/yt/funcs.py
+++ b/yt/funcs.py
@@ -815,7 +815,11 @@ def fix_axis(axis, ds):
 def get_image_suffix(name):
     suffix = os.path.splitext(name)[1]
     supported_suffixes = ['.png', '.eps', '.ps', '.pdf', '.jpg', '.jpeg']
-    return suffix if suffix in supported_suffixes else ''
+    if suffix in supported_suffixes or suffix == '':
+        return suffix
+    else:
+        mylog.warning('Unsupported image suffix requested (%s)' % suffix)
+        return ''
 
 def get_output_filename(name, keyword, suffix):
     r"""Return an appropriate filename for output.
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Ah, that explains why the fix was so easy: most of the work was already done ! 🎉

yt/funcs.py
```diff
@@ -814,7 +814,12 @@ def fix_axis(axis, ds):
 
 def get_image_suffix(name):
     suffix = os.path.splitext(name)[1]
-    return suffix if suffix in ['.png', '.eps', '.ps', '.pdf'] else ''
+    supported_suffixes = ['.png', '.eps', '.ps', '.pdf', '.jpg', '.jpeg']
+    if suffix in supported_suffixes or suffix == '':
+        return suffix
+    else:
+        mylog.warning('Unsupported image suffix requested (%s)' % suffix)
```
matthewturk, line 821: so if I don't supply a suffix am I going to get this warning every time?

munkm: if you don't supply a suffix it should be handled in the `if` part of this conditional I think. I just saved without adding a suffix and it chose 'png' for me without displaying this warning.

matthewturk: Oh, you're right.  Never mind.

neutrinoceros: That’s right, is.path.splittext will return 2 strings no matter what, but the second is empty when there’s no suffix to be isolated

PR #2553, opened 2020-04-21
Abstract data structures

### LINKED ISSUE #2426: Discussion: make Dataset an abstract class ?

neutrinoceros:
## PR Summary

fix #2426 

This is a first effort to make new frontend crafting process easier.
Class methods that require implementation are marked as abstract, and I also made some slight improvements over the existing internal docs both in `Dataset` definition and in the _skeleton frontend template.

Note that one frontend required minor adjustments to this abstraction (GADGET).

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -263,19 +263,55 @@ class Dataset(abc.ABC):
         self._set_derived_attrs()
         self._setup_classes()
 
+    @classmethod
+    @abc.abstractmethod
+    def _is_valid(cls, *args, **kwargs):
+        """An heuristic test to determine if the data format can be interpreted
+           with the present frontend."""
+        return False
+
     @abc.abstractmethod
     def _parse_parameter_file(self):
-        """Init instance attribute from file data, including:
-        - self.current_time (float)
-        - self.dimensionality (str)
-        - self.domain_dimensions (3-element ndarray(dtype=int64))
-        - self.geometry (str)
-        - self.cosmological_simulation (int)
-        - self.current_redshift (float)
-        - self.omega_matter (float)
-        - self.omega_lambda (float)
-        - self.hubble_constant (float)
-        """
+        # This needs to set up the following items.  Note that these are all
+        # assumed to be in code units; domain_left_edge and domain_right_edge
+        # will be converted to YTArray automatically at a later time.
+        # This includes the cosmological parameters.
+        #
+        #   self.unique_identifier      <= unique identifier for the dataset
+        #                                  being read (e.g., UUID or ST_CTIME)
+        #   self.parameters             <= dict full of code-specific items of use
+        #   self.domain_left_edge       <= three-element array of float64
+        #   self.domain_right_edge      <= three-element array of float64
+        #   self.dimensionality         <= int
+        #   self.domain_dimensions      <= three-element array of int64
+        #   self.periodicity            <= three-element tuple of booleans
+        #   self.current_time           <= simulation time in code units (float)
+        #
+        # We also set up cosmological information.  Set these to zero if
+        # non-cosmological.
+        #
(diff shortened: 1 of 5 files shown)


neutrinoceros:
Nothing too bad, but this first draft uncovered two weaknesses in the Gadget frontend. Namely `GadgetFOFHaloDataset` is missing  `_is_valid` and `_set_code_unit_attributes` methods.
The latter is currently circumvented by doing the work in `set_code_units()`, so I'm hoping this can easily be corrected without breaking the frontend. The former simply means that this class is never meant to be instantiated directly by `yt.load()`, so it's very easy to fix too.

neutrinoceros:
Important note: `abc.ABC` was actually introduced in python 3.4, so to allow the py2 tests to pass, I renounced inheriting from it anywhere. In this implementation, it's still allowed to instantiate classes that have abstract methods, so it's more permissive than intented, but we preserve the main result: frontends authors get a very clear error if they forget to implement those methods.

In the future, when py2 tests are dropped, this work can be completed.

matthewturk:
In principle I'm quite supportive of making us more formalized.  I think it might be wise to identify this and other improvements that would all serve to support that, as well, and see if we can pull the band-aid off all at once.

munkm:
@matthewturk where would be a good place to have that conversation? In the linked issue, or here? I'd like to make sure we have the discussions about this in one place so it's easier to track the history and make sure everything gets done. 

This might be a big enough improvement that it merits a [YTEP](https://ytep.readthedocs.io/en/master/).

neutrinoceros:
Weirdly, the py38 job was completed (successfully) hours ago but is still shown as "pending" here.

brittonsmith:
I like this. As the person responsible for the `GadgetFOFHaloDataset`, both of those changes seem fine. You're right, the dataset is not meant to be instantiated with `yt.load`, so the `_is_valid` returning False is good. I think the other change to setting units is also fine, but am slightly less sure.

One other comment, I wonder if it would be better to issue this to the `yt-4.0` branch. Python 2 compatibilities like the `six` module have been removed in that branch, so if this PR goes into `master`, we'll need to do some more of that when `yt-4.0` goes into `master`. On the other hand, `yt-4.0` isn't running answer tests yet, and I'd definitely like to see those be run on this.

neutrinoceros:
Thank you !
Unfortunately I'm stuck on the master's branch because I never managed to install yt-4.0 in my env yet (despite several attempts), so this factor outweighs other criteria... plus, answer tests are definitely needed to validate this. You are right though, this change would look better without six/python2.

jzuhone:
@neutrinoceros maybe we should try helping you get `yt-4.0` set up? I agree with @brittonsmith that it would be better to get this working there.

neutrinoceros:
Nevermind I finally found a workaround to the error that was showing up on my Mac (from https://github.com/huggingface/neuralcoref/issues/97), thanks for offering to help though :)

neutrinoceros:
Alright let's try it with yt4 then (#2556)

neutrinoceros:
I'm closing this PR this the yt-4.0 version of it is moving forward.

PR #2555, opened 2020-04-21
bugfix colormap minorticks

### LINKED ISSUE #2543: bug set_colorbar_minorticks callback has no effect
### Bug report

**Bug summary**
title.

**Code for reproduction**

```python
import yt
ds = yt.testing.fake_amr_ds()
p = yt.SlicePlot(ds, "z", "Density")
p.set_colorbar_minorticks("Density", True)
```

**Actual outcome**
coucou_Slice_z_density

**Expected outcome**

The colorbar minor ticks should be visible.

**Version Information**
  * Operating System: macOS
  * Python Version: 3.8.2
  * yt version: 3.6dev1
  * matplotlib version: 3.1.3

neutrinoceros:
## PR Summary

fix #2543 

I tried respecting the original intended logic tree but I didn't test in the case `MPL_VERSION < 2.0`.
I expect a lot of image comparison tests should fail because, as it turns out, colorbar minor ticks are in fact supposed to be _on_ by default, althewhile being impossible to actually draw in log transforms with MPL > 2.
A work around those failure would be to embrace the de facto default behaviour (no colorbar minor ticks).

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -71,7 +71,7 @@ answer_tests:
     - yt/analysis_modules/photon_simulator/tests/test_spectra.py
     - yt/analysis_modules/photon_simulator/tests/test_sloshing.py
 
-  local_unstructured_010:
+  local_unstructured_011:
     - yt/visualization/volume_rendering/tests/test_mesh_render.py:test_composite_mesh_render
     - yt/visualization/volume_rendering/tests/test_mesh_render.py:test_composite_mesh_render_pyembree
     - yt/visualization/volume_rendering/tests/test_mesh_render.py:test_hex20_render
@@ -128,7 +128,7 @@ answer_tests:
   local_axialpix_006:
     - yt/geometry/coordinates/tests/test_axial_pixelization.py:test_axial_pixelization
 
-  local_cylindrical_background_001:
+  local_cylindrical_background_002:
     - yt/geometry/coordinates/tests/test_cylindrical_coordinates.py:test_noise_plots
 
   local_particle_trajectory_001:
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -51,7 +51,7 @@ answer_tests:
   local_owls_002:
     - yt/frontends/owls/tests/test_outputs.py
 
-  local_pw_027:
+  local_pw_028:
     - yt/visualization/tests/test_plotwindow.py:test_attributes
     - yt/visualization/tests/test_plotwindow.py:test_attributes_wt
     - yt/visualization/tests/test_particle_plot.py:test_particle_projection_answers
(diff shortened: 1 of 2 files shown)


neutrinoceros:
As a note, here's a sum up of the current image-comp failures

# Python 3:

- colorbars get minor ticks by default, while they don't is the stored results (this one I anticipated and may be desirable)
- x minor ticks also show up ! unsurprisingly, they don't make any sense. I haven't reproduced this locally so it may depend on mpl's version...

# Python 2:

- even in the log_transform case, some colorbars seem to get _additional_ minor ticks that look linear-spaced.

In any case I'll need to replicate the test envs to test this more easily.

neutrinoceros:
related PRs
- #2101
- #2249 (maybe)

neutrinoceros:
I note that in py2 (mpl 2.2.3), on master, minor ticks can not be set _off_ with log_transform. This is another form of the bug that I'm fixing here, but I don't think it's worth fixing it since compatibility with python 2 is to be dropped, so I choose a conservative solution that should "fix" the answer tests in that env.

neutrinoceros:
Remaining errors (both in py2 and py3) are all plots that _gained_ minor ticks, which I believe can be considered a fix. I will however not update answer stores without external validation.
Another thing to condider is the fact that, in the log_transform + very narrow dynamical range case, new answers images are slightly truncated (colobar label), and I don't know how to dress this.

neutrinoceros:
Here are some examples of the current diffs:
## py3
**old**
old
**new**
new

## py2
**old**
old2
**new**
new2

I also note that some tests in py2 are reported are failures even though their diffs look like this
py2diffnull

yt/visualization/plot_window.py
```diff
@@ -1052,20 +1052,32 @@ def _setup_plots(self):
...
+                elif self._field_transform[f] == log_transform:
+                    if MPL_VERSION >= LooseVersion('3.0.0'):
+                        self.plots[f].cax.minorticks_on()
+                        self.plots[f].cax.xaxis.set_visible(False)  # experimental
+                    else:
+                        mticks = self.plots[f].image.norm(get_log_minorticks(vmin, vmax))
+                        self.plots[f].cax.yaxis.set_ticks(mticks, minor=True)
+
+                else:
+                    mylog.error("Unable to draw cbar minorticks.")
```
munkm, line 1077: It might be worth printing in the error whatever transform the user is passing to the plot object. That way a user knows why the minorticks can't be drawn and might try a different transform that's supported.

neutrinoceros: Good point, done.

tests/tests.yaml
```diff
@@ -71,7 +71,7 @@ answer_tests:
     - yt/analysis_modules/photon_simulator/tests/test_spectra.py
     - yt/analysis_modules/photon_simulator/tests/test_sloshing.py
 
-  local_unstructured_010:
+  local_unstructured_011:
```
munkm, line 74: you'll also want to change l54 from local_pw_027 to local_pw_028

neutrinoceros:
Thanks a lot @munkm for your help with updating the test answers !

PR #2557, opened 2020-04-22
Fixing FLASHParticleFile IO

matthewturk:
This implements `_yield_coordinates` for the `FLASHParticleDataset` type, which is a blocker for running the tests.

@jzuhone this look reasonable to you?

### DIFF
--- a/yt/frontends/flash/io.py
+++ b/yt/frontends/flash/io.py
@@ -239,17 +239,6 @@ class IOHandlerFLASHParticle(BaseIOHandler):
             ind += self._chunksize
         return morton
 
-    def _yield_coordinates(self, sub_file, needed_ptype=None):
-        si, ei = sub_file.start, sub_file.end
-        p_fields = self._handle["/tracer particles"]
-        px, py, pz = self._position_fields
-        npart = ei-si
-        pos = np.empty((npart, 3), dtype="=f8")
-        pos[:, 0] = p_fields[si:ei, px]
-        pos[:, 1] = p_fields[si:ei, py]
-        pos[:, 2] = p_fields[si:ei, pz]
-        yield "io", pos
-
     _pcount = None
     def _count_particles(self, data_file):
         if self._pcount is None:
--- a/yt/frontends/flash/io.py
+++ b/yt/frontends/flash/io.py
@@ -30,6 +30,7 @@ def determine_particle_fields(handle):
         _particle_fields = {}
     return _particle_fields
 
+
 class IOHandlerFLASH(BaseIOHandler):
     _particle_reader = False
     _dataset_type = "flash_hdf5"
@@ -145,6 +146,7 @@ class IOHandlerFLASH(BaseIOHandler):
                     rv[g.id][field] = np.asarray(data[...,i], "=f8")
         return rv
 
+
 class IOHandlerFLASHParticle(BaseIOHandler):
     _particle_reader = True
     _dataset_type = "flash_particle_hdf5"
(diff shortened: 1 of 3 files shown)


jzuhone:
@matthewturk I actually worked on this a bit last week:

https://github.com/jzuhone/yt/commit/143807491734a9406f62193b6dee057a82d43935

Your code is cleaner, but I still had some issues which I wasn’t able to figure out.

matthewturk:
I don't mind pulling your changeset in instead, if you'd rather!

jzuhone:
You can--you should check my work regarding `pcount`. Also, I was getting failures in the `assert(len(data_files) == 1)` lines.

matthewturk:
@jzuhone `len(data_files)==1` will only be true if we do what gets done in the `_read_particle_fields` call, which is to make `data_files` be a set of file names.  That applies to sub-chunking, which we (explicitly) don't need to do in `_yield_coordinates`.

jzuhone:
@matthewturk ok--so how should we handle things?

yt/frontends/flash/io.py
```diff
@@ -183,6 +183,12 @@ def _read_particle_coords(self, chunks, ptf):
                     total += count
                     yield ptype, (x, y, z)
 
+    def _yield_coordinates(self, data_file, needed_ptype=None):
+        px, py, pz = self._position_fields
+        p_fields = self._handle["/tracer particles"]
+        pxyz = np.asarray(p_fields[data_file.start:data_file.end], dtype="=f8")
+        yield ("io", pxyz)
```
jzuhone, line 190: I don't think this does what you want--I think it grabs all of the fields instead of only the position ones.

matthewturk: HA!  You're right.  Before my correction for flake8, it was `p_fields[data_file.start:data_file.end,(px,py,pz)]`.

matthewturk:
@jzuhone Do you have any changes to other stuff in that branch of development?  If so, I'll go ahead and merge them both together, but if the `_yield_coordinates` function I wrote is OK I'll keep it.

jzuhone:
@matthewturk with the fix we just discussed, your `_yield_coordinates` is probably preferable. It would probably be nice to have a function that creates smoothing lengths similar to what is done for tipsy, but I think that's probably out of the scope of this PR because if we're going to use it for multiple frontends we should probably do something more broad.

matthewturk:
@jzuhone ok -- I looked it over and what I'm going to do is merge from your branch and address the changes (including whitespace and `_count_particles`) then push to this PR.

matthewturk:
Need to fix one more thing, but am away for a few.

yt/frontends/flash/io.py
```diff
@@ -170,18 +172,22 @@ def _read_particle_coords(self, chunks, ptf):
...
-                total = 0
-                while total < pcount:
-                    count = min(self._chunksize, pcount - total)
-                    x = np.asarray(p_fields[total:total+count, px], dtype="=f8")
-                    y = np.asarray(p_fields[total:total+count, py], dtype="=f8")
-                    z = np.asarray(p_fields[total:total+count, pz], dtype="=f8")
-                    total += count
-                    yield ptype, (x, y, z)
+            pxyz = np.asarray(p_fields[data_file.start:data_file.end, (px,py,pz)], dtype="=f8")
+            yield "io", pxyz
```
matthewturk, line 184:
```suggestion
            yield "io", pxyz.T
```

jzuhone: I suspect that should get you the right shape?

matthewturk: yup, am fixing `_read_particle_fields` now

matthewturk:
@jzuhone I am merging this in to my `yt4-merge` work, but I need your advice -- now, a projection of `("deposit", "all_density")` throws an error.  What should I replace that with?

jzuhone:
@matthewturk I don't think we can replace it with anything, actually? these particles don't have smoothing lengths, so there's nothing to deposit or project. The assumption with these files is that there is no grid to speak of.

matthewturk:
Alright, that sounds good to me.  I'll yank those tests in the other PR.

yt/frontends/flash/io.py
```diff
@@ -170,18 +172,15 @@ def _read_particle_coords(self, chunks, ptf):
...
-                    total += count
-                    yield ptype, (x, y, z)
+            pxyz = np.asarray(p_fields[data_file.start:data_file.end, (px,py,pz)], dtype="=f8")
+            yield "io", pxyz.T
+
+    def _yield_coordinates(self, data_file, needed_ptype=None):
+        px, py, pz = self._position_fields
+        p_fields = self._handle["/tracer particles"]
+        pxyz = np.asarray(p_fields[data_file.start:data_file.end, (px,py,pz)], dtype="=f8")
+        yield ("io", pxyz)
```
munkm, line 183: is there a reason this is `yield ("io", pxyz)` and not `yield "io", pxyz`?

munkm:
So these commits somehow got incorporated into the 4.0 merge PR and already exist on master now.

PR #2556, opened 2020-04-22
Abstract data structures

neutrinoceros:
## PR Summary

At @brittonsmith and @jzuhone's joined request, here's #2553, reforged for yt-4.0 (and python 3 only 🎉).

Haven't been able to test this locally yet, so let's see what happens (inheritance + metaclasses conflicts seem very likely).

Fix https://github.com/yt-project/yt/issues/2426

PR #2558, opened 2020-04-23
[yt-4.0] bugfix to ensure sph normalization, gather and neighbors are applied

AshKelly:
## PR Summary

bugfix to ensure sph normalization, gather and neighbors are applied

## PR Checklist

- *NA* New features are documented, with docstrings and narrative docs

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -728,8 +728,10 @@ class YTCoveringGrid(YTSelectionContainer3D):
         fields = [f for f in fields if f not in self.field_data]
         if len(fields) == 0: return
 
-        smoothing_style = getattr(self.ds, '_sph_smoothing_style', 'scatter')
-        normalize = getattr(self.ds, '_use_sph_normalization', False)
+        smoothing_style = getattr(self.ds, 'sph_smoothing_style', 'scatter')
+        normalize = getattr(self.ds, 'use_sph_normalization', False)
+
+        print("smoothing style: ", smoothing_style)
 
         bounds, size = self._get_grid_bounds_size()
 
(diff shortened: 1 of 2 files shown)


matthewturk:
@AshKelly It looks to me like the results to

 * `yt.data_objects.tests.test_sph_data_objects.test_arbitrary_grid`
 * `yt.data_objects.tests.test_sph_data_objects.test_compare_arbitrary_grid_slice`

changed.  Are you able to run `nose -c nose_answer.cfg` locally and verify that your new answers are improvements?

AshKelly:
Yes - we discovered this bug goes deeper.

In frontends that inherit from the SPH frontend the correct answer is with '_' but for datasets that don't inherit from there, the answer should no have '_'.

I think the solution is to remove '_' in the SPH frontend and then things should work properly

matthewturk:
@AshKelly It strikes me that this is one very good reason to move forward with things like #2556 !  :)

AshKelly:
It was the introduction of the getters/setters which into the `SPHFrontend` and subsequent change of name for the attributes to include an `_` at the front.

For now - I think the easiest solution is what I pushed. Remove the setters/getters and then every dataset/container can just use `<>.use_use_sph_normalization` etc. But, I guess #2556 could solve this better later on!

AshKelly:
I should add that Robin Kooistra found this bug!

AshKelly:
I need to check `yt.data_objects.tests.test_sph_data_objects.test_arbitrary_grid` tests. I *think* this is because the change in `_sph_types` means they are no longer being smoothed.

matthewturk:
@AshKelly Can we help get this over the finish line?  Do you want a hand with fixing the conflicts?

munkm:
@AshKelly would you mind updating this with the current version of `main`, please? I think this would be good to get in for 4.0 too!

PR #2563, opened 2020-04-27
[bugfix] add units to profile weight field

brittonsmith:
## PR Summary

This adds units to profile weight fields, so the following works.

```
ds = yt.load(...)
p = yt.create_profile(..., weight_field=<not None>)
print (p.weight.units)
```

Currently `p.weight` is just a numpy array without units.

## PR Checklist

### DIFF
--- a/yt/data_objects/profiles.py
+++ b/yt/data_objects/profiles.py
@@ -224,6 +224,8 @@ been deprecated, use profile.standard_deviation instead."""
                   array_like_field(self.data_source,
                                    all_std[...,i], field)
                 self.standard_deviation[field][blank] = 0.0
+                self.weight = array_like_field(
+                    self.data_source, self.weight, self.weight_field)
             self.field_data[field][blank] = 0.0
             self.field_units[field] = self.field_data[field].units
             if isinstance(field, tuple):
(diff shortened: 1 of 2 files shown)


jzuhone:
@munkm `create_profile` does allow one to create 2D, 3D profiles if `bin_fields` is a list of fields.

jzuhone:
@munkm sorry, I think I misunderstood your question. But it looks like @brittonsmith did edit the `Profile` object?

munkm:
Whoops, yes you're right. Sorry Britton! 

But for some reason when I do the following:

```
import yt
ds = yt.load("/Users/madicken/yt_data/IsolatedGalaxy/galaxy0030/galaxy0030") 
dd = ds.all_data()
p1 = yt.create_profile(dd, 'density', 'cell_mass')
rho_min, rho_max = dd.quantities.extrema("density")
p2 = yt.Profile1D(dd, 'density', 32, rho_min, rho_max, True, weight_field='cell_mass')
``` 

p1 has a `weight` and `weight.units`, but p2 does not.  Am I missing something?

munkm:
Ah, I see. p2 needs to have `add_fields()` called in order to access the `weight` attribute. Sorry for the confusion!

PR #2567, opened 2020-04-27
first cut at showing multiple transfer functions on VR

zingale:
## PR Summary

Currently `save_annotated()` doesn't work correctly if we have multiple volume sources in the volume render.  This is an attempt to fix that by looping over sources and showing all the transfer functions.  With multiple transfer functions, we don't use the colorbar axis, but instead define a rectangle (`tf_rect`) in which to place the transfer functions.  This is subdivided vertically to account for the number of render sources and the transfer functions are shown there.  This gives some control over the location and sizing of the legends.

## PR Checklist


zingale:
Here's an example output:

test_cb

zingale:
updated output, with tf_rect manually changed to `[0.8, 0.2, 0.08, 0.6]`

test_cb

zingale:
I've added a cookbook receipe demonstrating this, using the standard IsolatedGalaxy dataset.  The render appears as:

test_cb

zingale:
note also, in producing that example, I fixed a bug in the latex rendering -- the exponent was not places in `{}`

zingale:
I just updated this with a fix for the case where not all the render sources have a transfer function (e.g., a volume source + a point source)

zingale:
I've synced this up with `master` and verified that it still renders the two fields for the isolated galaxy correctly.

zingale:
can I get some guidance on what people would like to see for this PR in order to get it merged?

munkm:
Hi @zingale, I'm so sorry I dropped the ball on reviewing this. Would you mind merging or rebasing your PR with the current master branch so we can see how the tests look (and the conflicts are resolved)? I'm super excited about this feature and I think it will be a great addition to the project!

neutrinoceros:
Wow it's been a while.
That said, the only 2 conflicts here are caused by auto-formatting. It should be trivial to fix.
@zingale do you still wish to get this merge ? As Madicken suggested, you need'd to update this branch somehow (rebase or merge), and then hopefully tests will stay green.

zingale:
I'll sync this up today.  Thanks for the poke

neutrinoceros:
A couple more fixes are needed. Most of them should be automatable with pre-commit.
Typing issues (mypy), not so much, but I can help there if needed @zingale :)

zingale:
okay, I merged and fixed some issues due to volume rendering API changes, and the example works again:

render_two_fields_tf

zingale:
I think maybe some fonts changed, but the rendering looks right here.

neutrinoceros:
I bet the font change is caused by matplotlib >= 3.4, and could use a fix similar to #3520. So, not your fault, I'm guessing

neutrinoceros:
Linters are still unhappy. Please see https://results.pre-commit.ci/run/github/88776493/1642620560.PN5I6bCmQRCkJvn6uKC7WA
For the report and solution @zingale

zingale:
okay, I think I fixed the pre-commit.ci issues.

neutrinoceros:
Yes you did, well played ! I will now step in to review this in Madicken's absence.

yt/visualization/volume_rendering/scene.py
```diff
@@ -367,6 +367,7 @@ def save_annotated(
         dpi: int = 100,
         sigma_clip: Optional[float] = None,
         render: bool = True,
+        tf_rect: Optional[list] = None,
```
neutrinoceros, line 370: requires `from typing import List` at the top level
```suggestion
        tf_rect: Optional[List[float]] = None,
```

yt/visualization/volume_rendering/scene.py
```diff
@@ -437,15 +443,55 @@ def save_annotated(
...
+        else:
+            # set the origin and width and height of the colorbar region
+            if tf_rect is not None:
+                try:
+                    cbx0, cby0, cbw, cbh = tf_rect
+                except ValueError:
+                    print(
+                        "tf_rect does not have correct number of values.  Using default"
+                    )
+                    tf_rect = None
```
neutrinoceros, line 475: I don't think it's necessary or helpful to catch this error.
```suggestion
                cbx0, cby0, cbw, cbh = tf_rect
```

yt/visualization/volume_rendering/scene.py
```diff
@@ -437,15 +443,55 @@ def save_annotated(
...
+
+        # number of transfer functions?
+        num_trans_func = 0
+        for rs in self._get_render_sources():
+            try:
+                tf = rs.transfer_function
+                num_trans_func += 1
+            except AttributeError:
+                print("error")
+                pass
```
neutrinoceros, line 458: Why should this error be caught ? I wonder wether there's a legit reason (in which case the `except` block should contain more information), or if we should simply drop the `try/except` struct.

yt/visualization/volume_rendering/scene.py
```diff
@@ -409,6 +410,11 @@ def save_annotated(
             If True, will render the scene before saving.
             If False, will use results of previous render if it exists.
             Default: True
+        tf_rect : seqeunce, optional
+           A rectangle that defines the location of the transfer
+           function legend for the case where there are multiple
+           volume sources with associated transfer functions.  tf_rect
```
neutrinoceros, line 416: so you're saying this argument has no effect if there's only one volume source w/ a transfer function ?
I'm thinking we could log a warning if the argument is passed but not used. Additionally I think the doctoring should make it crystal clear wether the argument is consumed or not.

yt/visualization/volume_rendering/scene.py
```diff
@@ -437,15 +443,55 @@ def save_annotated(
...
+            for i, rs in enumerate(self._get_render_sources()):
+                ax = self._render_figure.add_axes(
+                    [cbx0, cby0 + i * cbh_each, 0.8 * cbw, 0.8 * cbh_each]
+                )
+                try:
+                    tf = rs.transfer_function
+                    label = rs.data_source.ds._get_field_info(rs.field).get_label()
+                    self._annotate_multi(ax, tf, rs, label=label, label_fmt=label_fmt)
+                except AttributeError:
+                    pass
```
neutrinoceros, line 494: again it's not clear to me why we'd want to catch an error here. Please either drop the try/except construct or document a reason for doing in in the except block.

yt/visualization/volume_rendering/transfer_functions.py
```diff
@@ -605,7 +605,7 @@ def y_format(y, pos):
         ax.set_ylabel("Opacity")
         ax.set_xlabel("Value")
 
-    def vert_cbar(self, resolution, log_scale, ax, label=None, label_fmt=None):
+    def vert_cbar(self, resolution, log_scale, ax, label=None, label_fmt=None, size=10):
```
neutrinoceros, line 608: I'd suggest making new parameters to public functions keyword-only while it's time (it's much easier to do it when parameters are introduced than after sharing the code, when changing the signature is a breaking change)
```suggestion
    def vert_cbar(self, resolution, log_scale, ax, label=None, label_fmt=None, *, size=10):
```

yt/visualization/volume_rendering/scene.py
```diff
@@ -505,6 +551,18 @@ def _annotate(self, ax, tf, source, label="", label_fmt=None):
             log_scale=source.log_field,
         )
 
+    def _annotate_multi(self, ax, tf, source, label="", label_fmt=None):
```
neutrinoceros, line 554: Right now this private method is only called in one place where all parameters are passed explicitly, so I suggest we drop unused default values
```suggestion
    def _annotate_multi(self, ax, tf, source, label, label_fmt):
```

neutrinoceros (commented):
In general this is good and I've checked that it works as intended.
Here are a couple suggestions. My main concern is about the use of try/except blocks in places where I don't see the gain they provide. In addition to my detailed comments, note that running code through `try/except` structures has a performance cost (though that is not the reason why I'm advocating to revise them here, as they are used in places that are far from performance-critical).

Another concern, or rather question that I have, is do we want to support a recipe that takes 20min to run (as timed on my laptop) ? I don't know what our current standards for volume rendering recipes are, or how often they are actually run (if at all), so this is a very genuine and naive question. It may very well be that the computational cost is actually negligible.

yt/visualization/volume_rendering/scene.py
zingale: it's been so long since I initially wrote this -- but I think some sources might not have transfer functions (like a point source), so we want to skip them.

zingale: okay, indeed, I looked at the original vis I did with this, and I had 2 volume sources and 1 point source, so we need to skip the point source, since that has no transfer function.

yt/visualization/volume_rendering/transfer_functions.py
zingale: fine with me.

yt/visualization/volume_rendering/scene.py
zingale: okay, that's fair.

yt/visualization/volume_rendering/scene.py
```diff
@@ -367,7 +367,7 @@ def save_annotated(
         dpi: int = 100,
         sigma_clip: Optional[float] = None,
         render: bool = True,
-        tf_rect: Optional[list] = None,
+        tf_rect: Optional[List[float]] = None,
```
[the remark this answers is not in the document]
zingale: sorry, this is my first experience with typing.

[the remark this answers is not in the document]
neutrinoceros: no worries. Admittedly typing is a hassle. In Python 3.9, the builtin name `list` can be used instead of `typing.List`, which makes things easier, but we still support Python 3.7 so...

zingale:
I don't know about the runtime question -- it takes a few minutes on my workstation.  But this is not the only VR of IsolatedGalaxy, so it can't be that much more expensive than existing ones in the cookbook.

neutrinoceros:
> I don't know about the runtime question -- it takes a few minutes on my workstation. But this is not the only VR of IsolatedGalaxy, so it can't be that much more expensive than existing ones in the cookbook.

Fair enough

yt/visualization/volume_rendering/scene.py
neutrinoceros: Alright thanks for the explanation. In that case here's a refactor suggestion
```suggestion
            if hasattr(rs, "transfer_function"):
                num_trans_func += 1
```

yt/visualization/volume_rendering/scene.py
neutrinoceros: question answered in the previous thread

yt/visualization/volume_rendering/scene.py
zingale: I like that.  Thanks.

yt/visualization/volume_rendering/scene.py
```diff
@@ -437,15 +443,49 @@ def save_annotated(
...
+            for i, rs in enumerate(self._get_render_sources()):
+                ax = self._render_figure.add_axes(
+                    [cbx0, cby0 + i * cbh_each, 0.8 * cbw, 0.8 * cbh_each]
+                )
+                try:
+                    tf = rs.transfer_function
+                    label = rs.data_source.ds._get_field_info(rs.field).get_label()
+                    self._annotate_multi(ax, tf, rs, label=label, label_fmt=label_fmt)
+                except AttributeError:
+                    pass
```
neutrinoceros, line 484: I suggest making the error catching narrower so we don't unintentionally catch anything else than we need
```suggestion
                try:
                    tf = rs.transfer_function
                except AttributeError:
                    pass
                else:
                    label = rs.data_source.ds._get_field_info(rs.field).get_label()
                    self._annotate_multi(ax, tf, rs, label=label, label_fmt=label_fmt)
```

yt/visualization/volume_rendering/scene.py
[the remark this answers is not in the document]
zingale: this should be fixed now

yt/visualization/volume_rendering/scene.py
zingale: I've clarified this

neutrinoceros (approved):
Indeed. So I think we'd need a second approval to merge this. I'll advertise it on slack. Thank you @zingale !

yt/visualization/volume_rendering/scene.py
```diff
@@ -437,15 +444,46 @@ def save_annotated(
...
+        else:
+            # set the origin and width and height of the colorbar region
+            if tf_rect is not None:
+                cbx0, cby0, cbw, cbh = tf_rect
+
+            if tf_rect is None:
+                cbx0 = 0.80
+                cby0 = 0.12
+                cbw = 0.12
+                cbh = 0.9
```
neutrinoceros, line 472: Oh, sorry I didn't see this earlier. What do you think about this simplification ?
```suggestion
            if tf_rect is None:
                tf_rect = 0.80, 0.12, 0.12, 0.9
            cbx0, cby0, cbw, cbh = tf_rect
```

neutrinoceros: Alternatively it could even be done in one line
```suggestion
            cbx0, cby0, cbw, cbh = tf_rect or (0.80, 0.12, 0.12, 0.9)
```

zingale: I like the 3 line version better.  It's more readable.

neutrinoceros (dismissed):

yt/visualization/volume_rendering/scene.py
```diff
@@ -437,15 +444,41 @@ def save_annotated(
...
+        # which transfer function?
+        if num_trans_func == 1:
+            rs = self._get_render_sources()[0]
+            tf = rs.transfer_function
+            label = rs.data_source.ds._get_field_info(rs.field).get_label()
+            self._annotate(ax.axes, tf, rs, label=label, label_fmt=label_fmt)
+        else:
+            # set the origin and width and height of the colorbar region
+            if tf_rect is None:
+                tf_rect = 0.80, 0.12, 0.12, 0.9
```
neutrinoceros, line 466: This is the easiest way to fix the newly introduced typing error. Alternatively you could just drop the type annotation for this variable as it's actually a little too restrictive (passing a tuple or np.ndarray would be equally fine, but it's currently hard to express this in a simple type annotation). Sorry about the churn I should have anticipated this...
```suggestion
                tf_rect = [0.80, 0.12, 0.12, 0.9]
```

neutrinoceros: Not a big deal indeed but this actually makes it harder for downstream code to type check, FTR

neutrinoceros (approved):

neutrinoceros:
Looks like the docs build timed out. Let's hope that's not reproducible
@yt-fido test this please

zingale:
I just noticed that I made the render big, so I'll reduce it to speed it up for the cookbook.  I'll update in a sec.

zingale:
okay, I made it reuse the existing render and made it smaller.  At some point in the future, it would be good to separate the size of the rendering from the save_annotated image size, but I think that is a separate PR.  I also fixed a bug from my merging that lost the ability to set the size of the transfer function labels.  Now I can make them more appropriately sized.

neutrinoceros (approved):

matthewturk (approved):

zingale:
thanks @matthewturk and @neutrinoceros !!

PR #2568, opened 2020-04-28
Change the default answer test unit system to code

matthewturk:
This is a test PR to see if we get any failures that are not *answer value* failures if we change the base unit system for answer tests to `code` universally.

If this works, I will update this with a bit more of the rationale, and I will post to the yt mailing list.  It is not mergable as-is.

### DIFF
--- a/yt/utilities/answer_testing/framework.py
+++ b/yt/utilities/answer_testing/framework.py
@@ -442,14 +442,12 @@ class FieldValuesTest(AnswerTestingTest):
     _attrs = ("field", )
 
     def __init__(self, ds_fn, field, obj_type = None,
-                 particle_type=False, decimals = 10,
-                 unit_system = "cgs"):
+                 particle_type=False, decimals = 10):
         super(FieldValuesTest, self).__init__(ds_fn)
         self.obj_type = obj_type
         self.field = field
         self.particle_type = particle_type
         self.decimals = decimals
-        self.unit_system = unit_system
 
     def run(self):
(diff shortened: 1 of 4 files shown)


matthewturk:
It looks like *most* of the answers are wrong in ways we expect them to be wrong.  The Ramses answers are slightly confusing, as I *think* they might be showing different array lengths.

matthewturk:
It's also worth mentioning that this only applies to when a string is fed into things like `small_patch_amr`, so more work will need to be done, and this PR will require iteration as I turn some of these on.

matthewturk:
I verified that the RAMSES tests are in fact doing the right thing.  So in the next update, I will be upping the answer test IDs for the affected tests.

Xarthisius:
@yt-fido test this please

matthewturk:
I have verified that the enzo tests are all still correct -- just changed the unit system -- and that the RAMSES tests are a result of the `weight_field` being unitless.  This commit updates the RAMSES answer test and enzo answer test items.

yt/utilities/answer_testing/framework.py
```diff
@@ -1047,8 +1061,10 @@ def sph_answer(ds, ds_str_repr, ds_nparticles, fields):
                     yield PixelizedProjectionValuesTest(
                         ds, axis, field, weight_field,
                         dobj_name)
+            # See the other note about unit_system
```
munkm: maybe add a line number so it's clear what note you're referring to.

Issue #2564, opened 2020-04-27
annotate_clear is unclear 


munkm:
This is an issue I'm raising from yt-project/yt#2498 and yt-project/yt#2562 . In that PR a new function is added that lists the existing annotations added to a plot object. Following the convention of `annotate_clear()`, which clears the annotations on the plot, the author created a method called `annotate_list()`. I think in the case of these functions, we should modify them so that the action is the first word in the method, e.g. `clear_annotations()` and `list_annotations()`, since these are methods that operate on the plot annotations, rather than annotating a plot. 

`annotate_list()` has been changed to `list_annotations()` by the author of that PR. `annotate_clear()` should be updated to match it. In this case, I think we'll want to maintain backwards compatibility with the old method name with a deprecation warning on `annotate_clear`, we'll want to update all uses of annotate_clear() in the docs to be `clear_annotations()`, and we'll want to add a link in the api docs to the updated function. 

Does anybody have any issues with changing this?

zingale:
I like this name change -- it seems more natural.

jcoughlin11:
I like this change! I issued a small PR to address this (https://github.com/yt-project/yt/pull/2569)

matthewturk:
What if: we also added an `annotate_clearly` that took as an argument another annotation, and then what it *did* was just do that other annotation, but with a much thicker linestyle?  Like, "You can annotate velocity vectors, or you can annotate them *very clearly* like you'd used a sharpie"?

munkm:
I think your suggestion would look amazing on a figure using the `doom` colormap, @matthewturk.

neutrinoceros:
I agree that the new propropsed names would be clearer, but I do have a minor concern with it.
The current names allow the user to discover and learn about `annotate_list` and `annotate_clear` while playing with annotation callback through autocompletion, which I find is valuable. I think that's how I learned about them myself, I don't know if I would have liked the yt's annotation features as much is I missed those functions at first.

munkm:
Hm, that's a good point @neutrinoceros. I wonder if there's a way we could make them discoverable without having ambiguous names? In the original PR I pointed out that with `annotate_list` the function really sounds like the user would be annotating a list to their plot, not listing the annotations of the plot. I can easily see a user make this mistake with the list function (though clear is a little more nuanced).

matthewturk:
I think namespacing them would help, but is perhaps a bit too far.  For
instance, `p.annotate.vectors()` or something.  But, I agree, this is a
[...]

PR #2562, opened 2020-04-26
add a list_annotations() method

zingale:
## PR Summary

The `annotate_clear()` method takes an optional index of the annotation to remove, but there
is not high-level method to see the indices of the currently applied annotations. This adds a list_annotations() method to allow us to see what annotations are currently applied.

This reissues #2498. 

## PR Checklist

- [ ] Code passes flake8 checker
- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #2569, opened 2020-04-29
annotate_clear -> clear_annotations

jcoughlin11:
## PR Summary
Makes the functionality of the `annotate_clear` method, well, clearer, by changing it to `clear_annotations`.

Addresses https://github.com/yt-project/yt/issues/2564

## PR Checklist

### DIFF
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -44,7 +44,8 @@ from yt.frontends.ytdata.data_structures import \
     YTSpatialPlotDataset
 from yt.funcs import \
     mylog, iterable, ensure_list, \
-    fix_axis, fix_unitary, obj_length
+    fix_axis, fix_unitary, obj_length, \
+    issue_deprecation_warning
 from yt.units.unit_object import \
     Unit
 from yt.units.unit_registry import \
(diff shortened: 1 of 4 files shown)


yt/visualization/plot_window.py
```diff
@@ -1103,12 +1104,30 @@ def setup_callbacks(self):
...
     def annotate_clear(self, index=None):
         """
         Clear callbacks from the plot.  If index is not set, clear all
         callbacks.  If index is set, clear that index (ie 0 is the first one
         created, 1 is the 2nd one created, -1 is the last one created, etc.)
+
+        .. note::
+
+            Deprecated in favor of `clear_annotations`.
+
```
munkm, line 1116: I think the deprecation notice should be at the start of this docstring (it doesn't need to be in a note). If a user looks at this function, the very first thing they read should be that it has been deprecated in favor of clear_annotations.

chummels:
Just a note about why I chose to make this originally called `annotate_clear()` when I implemented it.  It was mostly because all of the existing annotations were `annotate_x()` or `annotate_y()`, so I wanted it to be visible and memorable for clearing them.  But I'm fine on it being changed to `clear_annotations()`.

munkm:
I just noticed that yt-project/yt#2562 is pointing to the 4.0 branch and this PR is pointing to master. This PR should also go to 4.0 or wait to be merged until after the 4.0 merge so it is accessible alongside the list_annotations feature.

PR #2570, opened 2020-04-30
Store unitful field values and don't compare weight fields

matthewturk:
This changes our testing to *not* compare `weight_field` in our projection values tests, and it also changes our tests to return unitful arrays from the `FieldValues` test.

I believe that this *should* pass as-is, and then I will update the `tests.yaml` to store new versions of the tests so we retain units in them.

Following this, we can unblock #2568 , where we will identify the items that need updates there, as well.

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -1,45 +1,45 @@
 answer_tests:
 
-  local_amrvac_003:
+  local_amrvac_004:
     - yt/frontends/amrvac/tests/test_outputs.py
     
-  local_artio_001:
+  local_artio_002:
     - yt/frontends/artio/tests/test_outputs.py
 
-  local_athena_005:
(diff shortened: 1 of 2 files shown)


matthewturk:
All [934 errors](https://tests.yt-project.org/job/yt_py38_git/194/#showFailuresLink) are `YTUnitConversionError`, and so I am fairly certain we can bump the test versions and move forward.

@munkm @Xarthisius @jzuhone @brittonsmith what do you think?

Xarthisius:
I'd just regenerate. You're messing with the code related only to storing answers. They changed. No big deal. If you want to be extra careful you could verify that all the failures are due to `YTUnitConversionError`

matthewturk:
I did verify!  OK, I will just bump them all.

PR #2575, opened 2020-05-04
[BUGFIX] Fix particle selection for corner cases of morton index values

### LINKED ISSUE #2574: Odd difference in sphere selection
I've been working on #2437 and adding redundant checks to, among other things, the sphere selection.  What I'm finding is something I don't know how to interpret.  This script demonstrates it:

```python
import numpy as np
import yt
from yt.testing import ParticleSelectionComparison

ds = yt.load_sample("snapshot_033")

dd = ds.all_data()
sp = ds.sphere('c', (0.1, 'unitary'))

r = (((dd["PartType1", "particle_position"] - sp.center)**2).sum(axis=1))**0.5
good_r = r < ds.quan(0.1, 'unitary')

ind_good = dd["PartType1", "particle_index"][good_r]
ind_bad = sp["PartType1", "particle_index"]

v_in_good_not_bad = np.lib.arraysetops.setdiff1d(ind_good, ind_bad)
v_in_bad_not_good = np.lib.arraysetops.setdiff1d(ind_bad, ind_good)
```

Note that here I am using `PartType1`, which has a smoothing length of 0.0 (i.e., it's not in `ds._sph_ptypes`).  What I see is that `v_in_good_not_bad` has a size of 1259, which is the difference between the sizes of `ind_good` and `ind_bad` (i.e., no duplicate entries).

I don't know why these produce different results.  A few things I thought might cause it -- smoothing length and periodicity -- are not issues in this particular setup, from what I can tell.

One odd thing that I found is that for this specific set of spheres, the smoothed particles give the right results, but that may be coincidence.

I'll update this ticket as I am able to track this down better.

matthewturk:
Thanks to @langmm for debugging this with me and for identifying the precise problem.  We discovered there was a corner case for filling subregions of mi1 and mi2, which showed up in #2574.  This corrects that by choosing the region correctly.

This is a bug identified by some work I'm doing that will add comprehensive testing to particle 
selection.

This fixes #2574.


matthewturk:
I would like to get reviews on this (the tests should pass as I had originally pushed backwards logic for `is_refined`) and hld off on merging until I demonstrate that it fixes functionality; I have a PR incoming on that.  Then once that is shown via failing (new) tests, we had hit the button on this, then hit the button on the other one.

yt/geometry/particle_oct_container.pyx
```diff
@@ -1752,11 +1751,14 @@ cdef class ParticleBitmapSelector:
...
-            ind1[0]+indexgap-1, ind1[1]+indexgap-1, ind1[2]+indexgap-1)
-        for mi in range(imi, fmi+1):
+        cdef np.uint64_t start_ind[3], end_ind[3]
+        cdef np.uint64_t shift_by = (self.bitmap.index_order1 - nlevel)
+        for i in range(3):
+            start_ind[i] = ind1[i] << shift_by
+            end_ind[i] = start_ind[i] + (1 << shift_by) - 1
+        imi = encode_morton_64bit(start_ind[0], start_ind[1], start_ind[2])
+        fmi = encode_morton_64bit(end_ind[0], end_ind[1], end_ind[2])
+        for mi in range(imi, fmi):
```
langmm, line 1761: If the morton indices are calculated in this way, I think the loop should be inclusive of `fmi`, which is the last morton index in the division identified by `ind1`. It can be left as non-inclusive if the version from my comment w/ the suggestion is used because then `fmi` will be the first index in the next morton ordered division.

yt/geometry/particle_oct_container.pyx
```diff
@@ -1532,11 +1532,11 @@ cdef class ParticleBitmapSelector:
     @cython.cdivision(True)
     @cython.initializedcheck(False)
     cdef void add_coarse(self, np.uint64_t mi1, int bbox = 2):
-        cdef bint flag_ref = self.is_refined(mi1)
+        cdef bint flag_ref
```
langmm, line 1535: This variable is unused now.
```suggestion
```

yt/geometry/particle_oct_container.pyx
langmm, line 1761: Is there a reason not to calculate `imi` with single bitshift and `fmi` from `imi` to eliminate second call to `encode_morton_64bit`? e.g.

```suggestion
        cdef np.uint64_t shift_by = 3 * (self.bitmap.index_order1 - nlevel)
        imi = encode_morton_64bit(ind1[0], ind1[1], ind1[2]) << shift_by
        fmi = imi + (1 << shift_by)
        for mi in range(imi, fmi):
```

yt/geometry/particle_oct_container.pyx
```diff
@@ -1767,12 +1769,15 @@ cdef class ParticleBitmapSelector:
...
-        for mi2 in range(imi, fmi+1):
+        cdef np.uint64_t shift_by = (self.bitmap.index_order2 +
+                                     self.bitmap.index_order1) - nlevel
+        cdef np.uint64_t start_ind[3], end_ind[3]
+        for i in range(3):
+            start_ind[i] = ind2[i] << shift_by
+            end_ind[i] = start_ind[i] + (1 << shift_by) - 1
+        imi = encode_morton_64bit(start_ind[0], start_ind[1], start_ind[2])
+        fmi = encode_morton_64bit(end_ind[0], end_ind[1], end_ind[2])
+        for mi2 in range(imi, fmi + 1):
```
langmm, line 1780: Similarly to above, why not use

```suggestion
        cdef np.uint64_t shift_by = 3 * ((self.bitmap.index_order2 +
                                          self.bitmap.index_order1) - nlevel)
        imi = encode_morton_64bit(ind2[0], ind2[1], ind2[2]) << shift_by
        fmi = imi + (1 << shift_by)
        for mi2 in range(imi, fmi):
```

yt/geometry/particle_oct_container.pyx
matthewturk: Thank you for the suggestions!  I went ahead and committed all three of them.

Xarthisius:
After merging this to 2577 tests still fail for me:

```
FAIL: This checks that we correctly subselect from a dataset, first by making
  File "/home/xarth/.local/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/xarth/codes/yt-project/yt/yt/frontends/gadget/tests/test_outputs.py", line 129, in test_particle_subselection
    assert_equal(psc.compare_dobj_selection(sp3) , True)
  File "/home/xarth/.local/lib/python3.6/site-packages/numpy/testing/_private/utils.py", line 428, in assert_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
 ACTUAL: False
 DESIRED: True
```

matthewturk:
The failing test is `sp3` which wraps around the domain along all three axes.  I'm adding some more test cases to narrow this down.  It does not seem to be related to the smoothing length.

matthewturk:
OK, I've changed my mind, it *is* related to smoothing length, indirectly.  Overriding the smoothing length *in the io-handler* causes this to work.

I believe what is happening is that the *periodic* smoothing lengths are not being selected inside `identify_file_masks`, which is what chooses which files are read.  I'm looking into that now, as I need to determine if it's not choosing an *unrefined* cell or a *refined* cell.

(This is precisely the kind of bug this test was supposed to ferret out, too!)

matthewturk:
The current commit reworks the smoothing length bitmap indexing at the *coarse* level for periodic boundary conditions.  It will not be ready to go in until I've both addressed this in the refined level *and* tested this.

At present for my test case, this fixes the problem, but it *also* throws errors for computing the bitmap indices, as it is now setting a few refined cells that are not on the coarse level.  Until at a bare minimum that is addressed this cannot go in.

matthewturk:
At present, I'm working to make similar changes to the refined index that I made to the coarse index.  It has been slow going, but hopefully today I'll be able to wrap it up.  Where it stands is that when I mark all the refined indices within a smoothing length of a particle as being collisions, I run out of allocated array values.

matthewturk:
The current state includes some C++ code and whatnot that may not make it; it's still not working but I wanted to push to get some thoughts from @langmm .

matthewturk:
OK, I have worked through this a fair bit, and I believe this is now correct.  All of the formerly failing tests for particle selection are now working.

What is not obvious to me, however, is if I am somehow making worse the overall speed.  This is something to be investigated.  I've done my *best* to make sure that the refined index generation does not become unbearably slow; it turns out that the particles that would get missed happen to be the least important ones, so that's ... fun.

The main change here from previous iterations is that now I am using two criteria for determining refinement that we weren't before -- the number of collisions (by default, any collisions at all will do it) and the number of particles (by default, 256).  Right now these are not user-facing parameters.  (They look like they are, but they are hard to make that be the case.)

I'm interested in folks taking a look.  Not all of the code is commented -- and suggestions for which parts would be *improved* by comments, as opposed to just *lengthened* by them (😁) would be appreciated.

I'm working on a full write-up of what caused this and how it's now fixed, but it will take a couple days.  In that write-up I'll also note places I'm unsure of.

yt/utilities/lib/ewah_bool_wrap.pyx
```diff
@@ -901,6 +901,8 @@ cdef class BoolArrayCollection:
             if it_map1 != ewah_coll1[0].end():
                 mi1_ewah1 = dereference(it_map1).second
                 mi1_ewah1.logicalor(mi1_ewah2, ewah_coll3[0][mi1])
+            else:
```
matthewturk, line 904: @langmm, when you get a chance, could you check this specific change to see if it matches what you think it should?  This was how I got some of the tests to start passing again.

matthewturk:
OK, now that I've explicitly cast to `uword` in the vendored ewah code, I am out of ideas for why it's failing to compile on appveyor.  :rage:

matthewturk:
One thing I would note is that I've snuck in an update to using bionic, as per @Xarthisius 's suggestion, on Travis.

matthewturk:
The travis answer test failure is itty bitty, and likely a result of a minor compiler change.  I'd like to propose that review commence, and *then* I update the answer tests.

matthewturk:
PR for answers is here: https://github.com/yt-project/answer-store/pull/4

yt/geometry/tests/test_particle_octree.py
```diff
@@ -175,11 +176,12 @@ def test_bitmap_no_collisions():
...
     posgen = yield_fake_decomp('sliced', npart, nfiles,
                                left_edge, right_edge)
+    coll = None
     for i, (pos, hsml) in enumerate(posgen):
-        nsub_mi = reg._refined_index_data_file(
-            pos, hsml, mask, sub_mi1, sub_mi2, i, 0)
-        reg._set_refined_index_data_file(
-            sub_mi1, sub_mi2, i, nsub_mi)
+        nsub_mi, coll = reg._refined_index_data_file(
+            coll, pos, hsml, mask, sub_mi1, sub_mi2, i,
```
munkm, line 182: Could you use `None` on l182 here like you did at 218? Or am I missing something subtle that coll needs to be defined here on l179 but not there?

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: python
-dist: xenial
+dist: bionic
```
munkm, line 2: I must have missed the reason for switching dists in travis. What was that for?

yt/utilities/lib/ewahboolarray/README
```diff
@@ -3,6 +3,6 @@ available at:
 
 https://github.com/lemire/EWAHBoolArray
 
-Currently this is at revision 80881379f8a582f45dda1be9edfc84d244846427.
+Currently this is at revision 88b25a3345b82353ccd97a7de6064e6c179a7cc2
```
munkm, line 6: Ahhh I see this got updated

.travis.yml
ax3l: Uff, this slipped through and caused a regression: older GCC (<6) need explicit C++11 flags in the build system, otherwise they will fail to build: #2892

PR #2582, opened 2020-05-06
Execute setup only on a parent thread to prevent a failure with spawn

Xarthisius:
## PR Summary

In python3.8 on OSX multiprocessing switched from *fork*ing to *spawn*ing. In order to prevent parallel cythonize messing up the concurrency `setup()` has to be executed only on the main process.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -387,63 +387,65 @@ class sdist(_sdist):
         )
         _sdist.run(self)
 
-setup(
-    name="yt",
-    version=VERSION,
-    description="An analysis and visualization toolkit for volumetric data",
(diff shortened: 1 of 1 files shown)


Xarthisius:
> What happens with `pip install .` ?

`pip install` invokes `python setup.py`

jzuhone:
@matthewturk I had the same problem either way

munkm:
@neutrinoceros this PR might fix the issue you had installing 4.0 on Python 3.8.

neutrinoceros:
@munkm It does ! 🎉 
Thanks a lot @Xarthisius !

PR #2584, opened 2020-05-06
Coerce mx, my, mz to code-length

matthewturk:
This coerces `mx`, `my`, and `mz` to `code_length` prior to creating an array.  It only does this for *dimensionless* quantities, as this is where we're *specifically* hitting the issue with the AMRVAC frontend.

This eliminates a couple hundred test failures in #2437.  In the absence of a hetereogeneously-united array, this is the best I could figure out.


jzuhone:
yeah, this is the only way we can do this now. hadn't realized this was an issue before.

PR #2587, opened 2020-05-07
make set_zlim units aware (sanitize zmin and zmax)

neutrinoceros:
## PR Summary

Because `set_zlim` only accepts floats, combining `pw.set_zlim` with `pw.set_unit` can create some unexpected results. For instance, this script is valid though hardly reliable

```python
import yt
ds = yt.testing.fake_amr_ds(fields=["density"])

my_min, _ = ds.find_min("density")    # returns a YTQuantity in cgs 
p.set_unit("density", "kg/mm**3")


zmin = my_min.value  # naive conversion to float 
p.set_zlim("density", zmin=zmin, zmax=10*zmin)   # <- now, it is incorrectly assumed that zmin is expressed in  "kg/mm**3"
```

With this change, set_zlim correctly handles zmin and zmax if they are YTQuantity objects, or tuples that convert to it.

### DIFF
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -895,12 +895,20 @@ class ImagePlotContainer(PlotContainer):
             zmin = zmax / dynamic_range.
 
         """
-        plot_units = self.frb[field].units
+
         def _sanitize_units(z):
+            # convert dimensionful inputs to float
             if isinstance(z, tuple):
                 z = YTQuantity(*z)
             if isinstance(z, YTQuantity):
-                z = z.to(plot_units).value
+                try:
+                    plot_units = self.frb[field].units
+                    z = z.to(plot_units).value
+                except AttributeError:
+                    # only certain subclasses have a frb attribute they can rely on for inspecting units
+                    from yt.funcs import mylog
+                    mylog.warning("%s class doesn't support zmin/zmax set as tuples or YTQuantity" %self.__class__.__name__)
+                    z = z.value
             return z
 
         myzmin = _sanitize_units(zmin)
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -900,7 +900,7 @@ class ImagePlotContainer(PlotContainer):
         def _sanitize_units(z):
             # convert dimensionful inputs to float
             if isinstance(z, tuple):
-                z = YTQuantity(*z)
+                z = self.ds.quan(*z)
             if isinstance(z, YTQuantity):
                 try:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Turns out, this doesn't work for PhasePlots, since they don't have a `frb` attribute to rely on for units. It seems much harder to even find where units are stored fo PhasePlots, so for now, I'll go with a warning log.

yt/visualization/plot_container.py
```diff
@@ -12,6 +12,7 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 #-----------------------------------------------------------------------------
+from yt import YTQuantity
```
Xarthisius, line 15: Could you change that to 
```suggestion
from yt.units import YTQuantity
```
?

neutrinoceros: sure, thanks !

yt/visualization/plot_container.py
```diff
@@ -894,8 +895,24 @@ def set_zlim(self, field, zmin, zmax, dynamic_range=None):
...
+            # convert dimensionful inputs to float
+            if isinstance(z, tuple):
+                z = YTQuantity(*z)
+            if isinstance(z, YTQuantity):
+                try:
+                    plot_units = self.frb[field].units
+                    z = z.to(plot_units).value
+                except AttributeError:
+                    # only certain subclasses have a frb attribute they can rely on for inspecting units
+                    from yt.funcs import mylog
```
brittonsmith, line 909: I think it's safe to put this in the main imports. I usually only do in-function imports if it's an optionally-installed module or something like that.

neutrinoceros: Thanks for this rule of thumb, I'll try to keep this in mind :)

yt/visualization/plot_container.py
```diff
@@ -894,8 +896,23 @@ def set_zlim(self, field, zmin, zmax, dynamic_range=None):
             zmin = zmax / dynamic_range.
 
         """
-        myzmin = zmin
-        myzmax = zmax
+
+        def _sanitize_units(z):
+            # convert dimensionful inputs to float
+            if isinstance(z, tuple):
+                z = YTQuantity(*z)
```
matthewturk, line 903: Can we change this to be `ds.quan` instead?  That way we can get units that are specific to the dataset, like comoving, hubble constant, redshift, etc.

neutrinoceros: excellent suggestion, done !

PR #2589, opened 2020-05-08
Avoid importing 'yt' deep in the code base

Xarthisius:
## PR Summary

It would be nice if we could avoid doing `import yt` within the yt submodules themselves. I noticed it was happening solely due to `yt._called_from_pytest`, which I think can be achieved using existing config idiom used with nose.


yt/testing.py
```diff
@@ -1190,7 +1189,7 @@ def skip(*args, **kwargs):
             msg = "`{}` backend not found, skipping: `{}`".format(backend, func.__name__)
             print(msg)
             pytest.skip(msg)
-        if yt._called_from_pytest:
+        if ytcfg.get("yt", "__withintesting"):
```
matthewturk, line 1192: I think the error comes from this, which will return `True` for *both* nose (which doesn't gloss over `pytest.skip()` calls) and pytest (which will *not* fail with `pytest.skip()` calls).

matthewturk: In the triage we thought one option might be to add another keyword to the `ytcfg` for pytest-status.

matthewturk:
@Xarthisius can you update the conflicts?

yt/testing.py
Xarthisius: Changed to `__withinpytest`

munkm:
🎉

PR #2592, opened 2020-05-15
Style: enforce import ordering with isort (YTEP0037 1/6)

neutrinoceros:
## PR Summary

[isort](https://pypi.org/project/isort/) is a tool that automatically sorts import statements.
It can be configured as a precommit hook too, so it should not be necessary to have this kind of PR, updating the whole code base more than once.

Imo, the 4.0 release would be a nice time to include this kind of nitpicky change, though I can easily imagine how the present PR would clog the yt-4.0 -> master merge, so I wouldn't suggest considering it for a merge ahead of this major task.

CURRENT STATE: This PR should include everything needed to smoothly adopt isort, but doesn't run it on the code base as it used to.

### Validation steps
- [x] ensure compatibility with flake8 (noqa lines)
- [x] add minimal black-compatible isort configuration to setup.cfg
- [x] check commutativity with black
- [x] identify and declare files to be ignored by isort
- [x] polish isort configuration : declare unyt as second or third party depending on the outcome of YTEP0037's vote 

### Documentation and workflow
- [x] add "pass isort" to PR template's checklist
- [x] add isort to the Dev guide (https://yt-project.org/docs/dev/developing/developing.html#automatically-checking-code-style)
- ~OPTIONAL: add a precommit hook ?~ (Done in #2600)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,7 +20,8 @@ detail.  Why is this change required?  What problem does it solve?-->
 
 <!-- Note that some of these check boxes may not apply to all pull requests -->
 
-- [ ] Code passes flake8 checker
+- [ ] pass `flake8 yt/`
+- [ ] pass `isort -rc . --check-only`
 - [ ] New features are documented, with docstrings and narrative docs
 - [ ] Adds a test for any bugs fixed. Adds tests for new features.
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,6 +82,10 @@ jobs:
       python: 3.6
       script: flake8 yt/
 
+    - stage: Lint
+      python: 3.6
+      script: isort --check-only -rc yt/
+
     - stage: tests
       name: "Python: 3.6 Minimal Dependency Unit Tests"
       python: 3.6
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -747,6 +747,27 @@ error and warning list
 blacklist a large number of the full list of rules that are checked by
 ``flake8`` by default.
 
+Import Formatting
+-----------------
+
+We use ``isort`` to enforce PEP-8 guidelines for import ordering.
+By decreasing priority order:
+FUTURE > STDLIB > THIRD PARTY > FIRST PARTY > EXPLICITLY LOCAL
+
+``isort`` can be installed via ``pip``
+
+.. code-block:: bash
+
+    $ pip install isort
+
+To validate import order, run ``isort`` recursively at the top level
+
+.. code-block:: bash
+
+    $ isort -rc . --check-only
+
+If any error is detected, rerun this without the ``--check-only`` flag to fix them.
+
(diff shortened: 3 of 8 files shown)


matthewturk:
I think we should do this, and possibly run black as well.  It might be
fine if we did this now and just made all incoming PRs also run it, but I
don't know -- #2437 makes me anxious about this.

neutrinoceros:
Actually I was reading through pandas' issue tracker to try and discover how they managed their transition to black on such a large project. Couldn't find the answer yet but doing so I just learned about isort, which seemed like a first step, much more lightweighted rewiew-wise.

So yeah, I would also like to see black adopted, but I can certainly understand the anxiety ! In any case I don't want to get in the way of #2437

matthewturk:
In all fairness if we ran black on the codebase now, and ran it on the PRs,
it should be OK.  :)

neutrinoceros:
Just a quick note: I just tried to combo isort + black on a personal project and they seem to have minor disagreements over formatting of comma-separated multiple-lines imports.
The conclusion is that it's definitely a good idea to discuss them together, but there's some tweaking needed to keep them playing nicely with each other.

neutrinoceros:
Here's the pandas thread https://github.com/pandas-dev/pandas/issues/26926
including a solution to the isort + black issue :)

neutrinoceros:
Apparently doing the isort pass on the whole code base at once breaks the installation (pip install -e . still runs but then the lib is unusable). Going to try a more iterative approach to identify what's going wrong.

neutrinoceros:
After a few iterations and rebasing, I think I've covered every little problem caused by isort-ing the code base. I also made sure to leave it in a state such that (black + isort) would be a commutative combo (only one change was needed for this).

There remain 2 open questions:
- should I bundle in a precommit hook ? This would guarantee future contributions keep the code isort-ed, but it forces developers to install isort (not an issue imo, as long as it's properly documented)
- more generally, and to anticipate a later adoption of black, what line-length should we settle for ? isort also has such an argument; I currently set it to 100, which is larger than black's default 79.

neutrinoceros:
Interestingly, isort gives slightly different results in python 3.6 on travis VS 3.8 on my machine.
I think that fixing this just requires a bit of tweaking in setup.cfg

neutrinoceros:
@yt-fido test this please

neutrinoceros:
It seems that the windows build started failing on the most recent commit, which only affects `.travis.yaml`, so I'm pretty confident this can be regarded as collateral damage from other failures. If so, this is one more reason appveyor should be reevaluated (#2599)

neutrinoceros:
This should not go in before #2640 and #2641

neutrinoceros:
Conflicts will show up every time someone changes/adds an import statement in master. I'll wait for this PR to be prioritised to solve this in one clean rebase.

neutrinoceros:
quick sanitising of this branch:
- rebased onto master
- put the "isort run" commit at the end of the history so it's easier to remove and rerun for future rebasing
- document line-length 88 and "unyt is third party"

PR #2593, opened 2020-05-15
add support for arbitrary polyhedra in the ExodusII format

tukss:
## PR Summary
In addition to a collection of equal-size polyhedra the ExodusII format also allows us to use collections of polyhedra with an unequal number of nodes. This is used in the output of the [FleCSI](https://github.com/laristra/flecsi) framework, for example. Software like ParaView loads these files without problems, but yt throws an error because the `connect` variable is only one-dimensional in this case.

I attach an example of such an output file, created with the [FleCSALE](https://github.com/laristra/flecsale) code.
[exo-example.zip](https://github.com/yt-project/yt/files/4636833/exo-example.zip)

This PR should give partial support for loading these files. At the moment it only supports 2D (the `nsided` case), but not 3D (`nfaced`). It also only works if the number of nodes per element is equal within one block. I realize that this restriction doesn't give any benefit beyond the previous implementation, but even this simple change would allow outputs in the more generic format to be loaded, like it is always used in FleCSI codes.

I am not sure if an unstructured mesh of elements with an unequal number of sides/faces is supported in yt at the moment. If so, I'd be happy to extend my code to support it. Parsing the connectivity array would be fairly straightforward.

## PR Checklist

This is my first contribution to yt. I could add my example output file as an additional test if pointed in the right direction.


matthewturk:
Wow, this is awesome!  Thank you.  I look forward to reviewing this!

matthewturk:
Hi @tukss , do you mind if I push to your fork with conflict resolution?

tukss:
No, I don't mind. Please go ahead.

matthewturk:
OK, looks like our magic format command didn't quite work this time, so I manually ran black and pushed.  This should be good to go, I think.

tukss:
Looks fine to me. Thanks for fixing the formatting.

matthewturk (approved):
LGTM!

yt/frontends/exodus_ii/data_structures.py
```diff
@@ -336,7 +336,30 @@ def _read_connectivity(self):
...
+                    elif elem_type == "nfaced":
+                        raise NotImplementedError(
+                            "3D arbitrary polyhedra are not implemented yet"
+                        )
+                    else:
+                        arbitrary_polyhedron = False
+                except AttributeError:
+                    arbitrary_polyhedron = False
+
+                conn = var[:].astype("i8")
```
neutrinoceros, line 353: Seems redundant since you're already imposing the dtype to `i8` earlier
```suggestion
                conn = var[:]
```

yt/frontends/exodus_ii/data_structures.py
```diff
@@ -336,7 +336,30 @@ def _read_connectivity(self):
...
+                if arbitrary_polyhedron:
+                    nodes_per_element = ds.variables["ebepecnt%d" % (i + 1)]
```
neutrinoceros, line 355: We normally use fstrings for string formatting now (though this is more recent than the opening of this PR !)
```suggestion
                    nodes_per_element = ds.variables[f"ebepecnt{i + 1}"]
```

yt/frontends/exodus_ii/data_structures.py
```diff
@@ -336,7 +336,30 @@ def _read_connectivity(self):
...
+                    if np.any(nodes_per_element != nodes_per_element[0]):
+                        raise NotImplementedError("only equal-size polyhedra supported")
+                    conn.shape = (
+                        conn.shape[0] // nodes_per_element[0],
+                        nodes_per_element[0],
+                    )
```
neutrinoceros, line 361: Reusing the first element of `nodes_per_element` more than twice grants a variable name imo.
Additionally, and this is probably very nitpicky, but I am not 100% confident in the way the `conn` array is being reshaped, so I'm suggesting refactoring it with a check.
```suggestion
                    npe = nodes_per_element[0]
                    if np.any(nodes_per_element != npe):
                        raise NotImplementedError("only equal-size polyhedra supported")
                    q, r = np.divmod(len(conn), npe)
                    assert r == 0
                    conn.shape = (q, npe)
```

neutrinoceros (approved):
I left a few, admittedly nitpicky suggestions for improvement, but overall I'm very happy with the patch, thank you !

Note that at the very least the `fstring` suggestion should be used, otherwise we'll get style check errors down the road.

yt/frontends/exodus_ii/data_structures.py
tukss: Yes, fstrings are nice. I normally avoid them for compatibility with Python < 3.6 (and even Python 2 in the past), but if that's the yt style, I'm perfectly OK with that.

yt/frontends/exodus_ii/data_structures.py
tukss: Agreed.

yt/frontends/exodus_ii/data_structures.py
```diff
@@ -336,7 +336,30 @@ def _read_connectivity(self):
...
+                try:
+                    elem_type = var.elem_type.lower()
+                    if elem_type == "nsided":
+                        arbitrary_polyhedron = True
+                    elif elem_type == "nfaced":
+                        raise NotImplementedError(
+                            "3D arbitrary polyhedra are not implemented yet"
+                        )
+                    else:
+                        arbitrary_polyhedron = False
```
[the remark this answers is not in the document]
tukss: I think this won't result in the correct behavior for files that use another `elem_type` (like `quad` etc.) because they would all be set to `nsided`.
As far as I know `nsided` and `nfaced` are the only two types that don't have the same number of nodes for each element, so they need special treatment.

yt/frontends/exodus_ii/data_structures.py
tukss: I'm fine with this. Is it yt style to prefer `len(conn)` to `conn.shape[0]`. I find the latter a bit clearer.

yt/frontends/exodus_ii/data_structures.py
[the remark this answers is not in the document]
tukss: I found a comprehensive list of possible values for `elem_type` in the source of the VTK Exodus writer.
https://gitlab.kitware.com/vtk/vtk/-/blob/master/IO/Exodus/vtkExodusIIWriter.cxx#L1200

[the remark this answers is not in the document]
neutrinoceros: But this doesn't assign `elem_type` !
The last lign assigns the Boolean result of `elem_type == "nsided"` to `arbitrary_polyhedron`, so it's just the compact form of what you wrote. :)

yt/frontends/exodus_ii/data_structures.py
neutrinoceros: So this suggestion reflects my now understanding here. I am under the impression that this array can only be 1D before we reshape it, in which case I think using `len()` is ok.
Imo using .shape[0] gives the false impression that the array is multidimensional, in which case the reshaping line would be broken (I think).

tukss: OK, agreed.

yt/frontends/exodus_ii/data_structures.py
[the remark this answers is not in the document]
tukss: Ah, sorry, I missed the double equals. I agree then.

neutrinoceros:
Thanks @tukss for your patience in going over each of my review points ! Now everything is resolved and I'll merge this once tests pass :)

PR #2594, opened 2020-05-16
Add mode argument to h5py.File

olebole:
## PR Summary

h5py is changing to require the mode (`'r'`, `'w'`, etc) specified when `h5py.File` is used.  The Debian version of h5py has already applied the upstream commits making this mandatory, which require this patch.

For reference: This was [Debian#960573](https://bugs.debian.org/960573), and we [fixed this locally in release 3.6.0-3](https://salsa.debian.org/debian-astro-team/yt/-/blob/6cf4a4d9ee190913b3648cd2291738f4cb601cf7/debian/patches/Add-mode-argument-to-h5py.File.patch).

## PR Checklist

### DIFF
--- a/doc/source/analyzing/analysis_modules/absorption_spectrum.rst
+++ b/doc/source/analyzing/analysis_modules/absorption_spectrum.rst
@@ -195,7 +195,7 @@ errors produced in saving to a ascii file will negatively impact fit quality.
 
 .. code-block:: python
 
-    f = h5py.File('spectrum.h5')
+    f = h5py.File('spectrum.h5', 'r')
     wavelength = f["wavelength"][:]
     flux = f['flux'][:]
     f.close()
--- a/doc/source/developing/external_analysis.rst
+++ b/doc/source/developing/external_analysis.rst
@@ -395,7 +395,7 @@ import, and then save things out into a file.
 .. code-block:: python
 
    import h5py
-   f = h5py.File("some_file.h5")
+   f = h5py.File("some_file.h5", "w")
    f.create_dataset("/data", data=some_data)
 
 This will create ``some_file.h5`` if necessary and add a new dataset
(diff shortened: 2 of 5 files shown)


doc/source/analyzing/analysis_modules/absorption_spectrum.rst
```diff
@@ -195,7 +195,7 @@ errors produced in saving to a ascii file will negatively impact fit quality.
 
 .. code-block:: python
 
-    f = h5py.File('spectrum.h5')
+    f = h5py.File('spectrum.h5', 'r')
```
neutrinoceros, line 198: Since this API is prone to changes, which is the very motivation for your PR, I would recommend using the explicit form
```suggestion
    f = h5py.File('spectrum.h5', mode='r')
```

olebole: That would be a quite larger change than that: a quick count that `h5py.File`  is used >150 times in yt (and everywhere with the mode as positional argument).

neutrinoceros: Here's a script that can do it for you !

```python
import re
from pathlib import Path

exp = re.compile("h5py\.File\([^,]+,\s*")
yt_dir = Path(__file__).parent
for file in yt_dir.glob("**/*.py"):
    matches = None
    with open(file, mode="r") as fileobj:
        text = fileobj.read()
        matches = re.findall(exp, text)
        for match in matches:
            print(type(match))
            text = text.replace(match, match + " mode=")
    if matches:
        with open(file, mode="w") as fileobj:
            fileobj.write(text)
```

save it as `yt/magic.py` and run `python magic.py`, then you can validate that it did the correct thing with `git diff` (and `git stash` to revert immediately in case something went wrong). I tested it on my machine and it worked :)

neutrinoceros: In case you are using email-notifications: sorry, my first attempt was actually broken, the script should work now !

munkm: How about if just the changes you've made in this PR use `mode`, @olebole ? We could issue a separate PR to update the use in the rest of the codebase (we're not consistent with the usage of `mode` already, so I think it's fine to not update the other lines). That script looks great though @neutrinoceros!

neutrinoceros: (side-note: the script is idempotent as it is, it doesn't require that all occurrences be changed at the same time !)

PR #2595, opened 2020-05-16
ci: bump flake8 linter version, fix newly detected errors (YTEP0037 2/6)

neutrinoceros:
We currently run flake8 3.6 on travis. A more recent release (3.8) catches precedently undetected errors.

Important note: there are _more_ errors than the ones I patch here.
This PR is specifically targeted at errors that can't be solved automatically with black.
Namely, these are almost exclusively string formatting errors, with either too many or not enough substitutions declared, as compared with the number of placeholders.

### DIFF
--- a/tests/lint_requirements.txt
+++ b/tests/lint_requirements.txt
@@ -1,6 +1,6 @@
-flake8==3.6.0
+flake8==3.8.1
 mccabe==0.6.1
-pycodestyle==2.4.0
-pyflakes==2.0.0
+pycodestyle==2.6.0
+pyflakes==2.2.0
 isort==4.3
 black==19.10b0
--- a/yt/data_objects/tests/test_fluxes.py
+++ b/yt/data_objects/tests/test_fluxes.py
@@ -69,15 +69,16 @@ class ExporterTests(TestCase):
         rhos = [0.5, 0.25]
         trans = [0.5, 1.0]
         for i, r in enumerate(rhos):
+            basename = "my_galaxy_color_%d" % i
             surf = ds.surface(sp,'density',r)
-            surf.export_obj("my_galaxy_color".format(i),
+            surf.export_obj(basename,
                             transparency=trans[i],
                             color_field='temperature', dist_fac=1.0,
                             plot_index=i, color_field_max=ma,
                             color_field_min=mi)
 
-        assert os.path.exists('my_galaxy_color.obj')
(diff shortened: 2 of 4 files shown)


neutrinoceros:
Apparently my attempted fix for the buggy test in `yt/data_objects/tests/test_fluxes.py` wasn't successful. I have a hard time understanding why files are seemingly not written properly (which is what the test is checking). @ngoldbaum, any thoughts ? :)

neutrinoceros:
I think I got it this time. The problem with the buggy test (namely `yt/data_objects/tests/test_fluxes.py::ExporterTests::test_export_obj`) is that there were two ways this line could be wrong, with 2 different fixes.
```python
"my_galaxy_emis".format(i)
```

My initial assumption was that there was a placeholder missing
```python
"my_galaxy_emis_{}".format(i)
```

But it now seems to me that no formatting is actually expected here
```python
"my_galaxy_emis"
```

yt/visualization/eps_writer.py
```diff
@@ -1280,14 +1280,14 @@ def multiplot_yt(ncol, nrow, plots, fields=None, **kwargs):
         if fields is None:
             fields = plots.fields
         if len(fields) < nrow*ncol:
-            raise RuntimeError("Number of plots is less "\
+            raise RuntimeError("Number of plots(%d) is less "\
```
cphyc, line 1283:
```suggestion
            raise RuntimeError("Number of plots (%d) is less "\
```

yt/visualization/eps_writer.py
```diff
@@ -1280,14 +1280,14 @@ def multiplot_yt(ncol, nrow, plots, fields=None, **kwargs):
...
                                "than nrow(%d) x ncol(%d)." % \
                                (len(fields), nrow, ncol))
             return
         figure = multiplot(ncol, nrow, yt_plots=plots, fields=fields, **kwargs)
     elif isinstance(plots, list) and isinstance(plots[0], (PlotWindow, PhasePlot)):
         if len(plots) < nrow*ncol:
-            raise RuntimeError("Number of plots is less "\
+            raise RuntimeError("Number of plots(%d) is less "\
```
cphyc, line 1290:
```suggestion
            raise RuntimeError("Number of plots (%d) is less "\
```

neutrinoceros:
promoted this to YTEP0037 2/6 since the slot was free and this can in principle be applied at any time.

PR #2596, opened 2020-05-16
Black adoption (YTEP0037 4/6)

neutrinoceros:
## PR Summary

this PR is based off, and dependent on https://github.com/yt-project/yt/pull/2598

CURRENT STATE: this PR should contain everything need to smoothly adopt black, but it doesn't actually run it on the code base (as it used to)

**nb**
#2595 introduces E117 to flake8's ignores, which isn't necessary after black is applied.

## PR Checklist

- [x] pass `flake8 yt/`
- [x] add black checker to Travis CI lint stage
- [x] update PR template (include `black --check` passing condition)
- [ ] add documentation
- [x] REBASE onto master or #2595 depending on timing


neutrinoceros:
note : Adding a `pyproject.toml` file messes up with `pip install -e .`. I was able to fix it (kinda) by adding bluid-system specifications but it still breaks the minimal testing env, and I currently do not understand how to fix this specific case.
I'm thinking that this should propably be a done in a separate prep PR.

munkm:
So it's documented here, [here's the last discussion](https://mail.python.org/archives/list/yt-dev@python.org/thread/VPXJWLZNTJV557NS32AREEAMTKOHCFEO/#VPXJWLZNTJV557NS32AREEAMTKOHCFEO) we had in the mailing list related to black.

neutrinoceros:
Thanks ! I wasn't aware this discussion already happened. I must add that, even though recent versions of git can indeed be configured so that `git blame` ignores specific commits, there is currently no integration for this in github's UI for `git blame` (according to black's own README).

tests/lint_requirements.txt
```diff
@@ -2,4 +2,8 @@ flake8==3.6.0
 mccabe==0.6.1
 pycodestyle==2.4.0
 pyflakes==2.0.0
+<<<<<<< HEAD
```
cphyc, line 5: There's something bad here!

PR #2597, opened 2020-05-17
Internally import from unyt

neutrinoceros:
## PR Summary

Quoting `CONTRIBUTING.rst`
> * Internally, only import from source files directly -- instead of:
     ``from yt.visualization.api import ProjectionPlot``
   do:
     ``from yt.visualization.plot_window import ProjectionPlot``

I think this principle could and should be applied to `unyt`, which we currently wrap under `yt.units`.
Directly importing from `unyt` instead of the wrapper would reduce the overhead of this wrapper: it's easier to learn the origin of a function/class if we avoid layered import. This should reduce the maintenance burden of the wrapper as well, though the primary goal is to make definitions easier to discover from everywhere else in `yt/`

**highlight points**
- This should not remove anything from namespaces `yt.units` (and its children) in any way, so as not to break any downstream code.
- Documentation (website + docstrings) remains unchanged, this is not about telling users not to use the wrapper, but only a matter of clarifying the code base.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
     hooks:
     -   id: flake8
 -   repo: https://github.com/timothycrosley/isort
-    rev: ''
+    rev: '5.1.4'
     hooks:
     -   id: isort
 -   repo: https://github.com/ambv/black
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,6 @@ combine_as_imports=True
 line_length=88
 # isort can't be applied to yt/__init__.py because it creates circular imports
 skip = venv, doc, benchmarks, yt/__init__.py, yt/extern
-known_third_party = IPython, nose, numpy, sympy, matplotlib, unyt, git, yaml, dateutil, requests, coverage, pytest, pyx
+known_third_party = IPython, nose, numpy, sympy, matplotlib, unyt, git, yaml, dateutil, requests, coverage, pytest, pyx, OpenGL
 known_first_party = yt
 sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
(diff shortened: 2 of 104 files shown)


neutrinoceros:
A possible extension of this change would be to stop using YTArray and YTQuantity aliases internally and only leave them exist as part of the user facing api defined by yt.units as a wrapper for unyt.

neutrinoceros:
This is the one proposal originally part of YTEP0037 that didn't reach consensus. It is not considered part of that YTEP any more.

neutrinoceros:
@yt-fido test this please

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@
     hooks:
     -   id: flake8
 -   repo: https://github.com/timothycrosley/isort
-    rev: ''
+    rev: '5.1.4'
```
neutrinoceros, line 6: this is in direct conflict with https://github.com/yt-project/yt/pull/2777, I'll remove this after #2777 is merged

neutrinoceros:
you mean this ? https://github.com/yt-project/yt/blob/9138a0a7330e642e040f8dc0a459f7dba347fa05/yt/units/__init__.py#L74

Yup, still there. This PR removes nothing, its one and only purpose is to replace indirect imports by direct ones.
There is indeed a handful of functionalities that are still defined in `yt.units`, but with this branch it is much easier to spot them.

munkm:
Here's a discussion we had about internally importing from unyt  in the code styling ytep (links to the first comment in a thread): 

https://github.com/yt-project/ytep/pull/14#discussion_r443853394

munkm:
I have some thoughts about doing this; I will weigh in when I get back early Sept.

neutrinoceros:
Note that this project

https://github.com/sqlalchemyorg/zimports

Can be used to perform this change in a semi-automated way by
1) replacing all imports from yt.units with star imports from unyt 
2) running zimports to expand explicit required imports

neutrinoceros:
If I'm ever reattempting this again, I'd rather start fresh. It's dishearthing to see how many conflicts this has been accumulating in less than a year.

PR #2598, opened 2020-05-18
Add pyproject.toml (YTEP0037 3/6)

neutrinoceros:
## PR Summary

In my work on #2596, I had to tweak the newly added `pyproject.toml` file (required by black for configuration) to make installation of yt with `pip install -e` work properly across python versions (>=3.6).

In order to validate it, I'm opening this separate PR to introduce the file to the repo without all black-related changes.

This file may introduce redundancy with `.travis.yml`, `setup.py`. Sadly, it won't be mergeable with `setup.cfg`, which we use to configure flake8 (currently not supporting pyproject.toml).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,12 @@
+[build-system]
+# See https://github.com/scipy/scipy/pull/10431 for the AIX issue.
+requires = [
+  "setuptools>=19.6",
+  "wheel",
+
+  # keep in sync with travis.yml "minimal" specs (Cython and numpy for py36)
+  "Cython>=0.26.1",
+  "numpy==1.13.3; python_version=='3.6' and platform_system!='AIX'",
+  "numpy==1.18.1; python_version=='3.7' and platform_system!='AIX'",
+  "numpy==1.18.4; python_version=='3.8' and platform_system!='AIX'",
+]


neutrinoceros:
Welp, it failed on windows + python 3.8. I do not own a windows machine so I won’t be able to test for this locally, but I’ll see if I can find out what’s up with that.

neutrinoceros:
Still failing on windows (py38 AND py37), despite my attempt to replicate the build specs that happened to work last. There's something I'm missing, I could use a hand if anyone has an idea.

matthewturk:
So I'm not 100% sure, but it *looks* to me like there's something funny going on with the compiler on windows, specifically related to whether or not it's getting the symbol `M_LOG2E`, which may just be a canary for something else.  I think this could either be because `M_LOG2E` is not defined anymore (is that likely?) or that somehow `pyproject.toml` ticks over into using a different compiler stack somehow.

@jzuhone do you have an idea why this might be the case?

jzuhone:
@matthewturk I don't have an idea about this currently, but I can try testing this on my own Windows machine sometime this week.

neutrinoceros:
Note: black can actually be configured in a dedicated file (e.g. black.toml), though it doesn’t support auto detection for anything different than my project.toml, so running it requires the additional (minor) overhead  of specifying the path to the custom config file.
ie 
```bash
black —config black.toml
```

We can first switch to black using this and solve the pyproject.toml issues later to simplify the linting process.

neutrinoceros:
fingers crossed but I think I fixed it.

PR #2600, opened 2020-05-19
Add precommit hooks for flake8, isort and black (YTEP0037 5/6)

neutrinoceros:
## PR Summary

As support to #2592 and #2596, add a precommit hook configuration file to allow devs to install  isort, black and flake8 as hooks locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,13 @@
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: ''
+    hooks:
+    -   id: flake8
+-   repo: https://github.com/timothycrosley/isort
+    rev: ''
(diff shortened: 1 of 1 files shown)


matthewturk:
I've done precommit hooks before but only for my personal repos.  What is the process for installing these locally?  After you clone, do you run `pre-commit install` in the main repo?

neutrinoceros:
Yes that’s it. I don’t think it needs to be mandatory though, just a nice option to offer to people if we start asking them to use black

neutrinoceros:
note that we already merged the corresponding documentation before this !

matthewturk:
I'm merging this because it's covered in YTEP-0037 and it makes things a lot easier for folks to get stuff to pass.

PR #2606, opened 2020-05-23
add progress bar display for yt.load_sample() (requires tqdm)

neutrinoceros:
# PR Summary
building on top of #2417, use progress bar already offered by pooch.
I chose to set `pbar=True` by default. Also this capability requires tqdm to be installed.

### DIFF
--- a/yt/utilities/load_sample.py
+++ b/yt/utilities/load_sample.py
@@ -11,7 +11,7 @@ from yt.funcs import mylog
 from yt.convenience import load
 from yt.utilities.on_demand_imports import _pooch as pch
 
-def load_sample(name=None, specific_file=None):
+def load_sample(name=None, specific_file=None, pbar=True):
     """
     Load sample data with yt. Simple wrapper around yt.load to include fetching
     data with pooch.
@@ -32,6 +32,9 @@ def load_sample(name=None, specific_file=None):
         `enzo_cosmology_plus/DD0003/DD0003`, and the argument passed to this
         variable would be `DD0003/DD0003`
 
+    pbar: bool
+        display a progress bar
+
     """
     fido = sd.Fido()
     if name is None:
@@ -45,12 +48,18 @@ def load_sample(name=None, specific_file=None):
     base_path = fido.fido.path
     fileext, name, extension = _validate_sampledata_name(name)
 
+    if pbar:
+        from pooch import HTTPDownloader
+        downloader = HTTPDownloader(progressbar=True)
+    else:
+        downloader = None
+
     if extension == "h5":
-        fname = fetch_noncompressed_file(fileext, fido)
+        fname = fetch_noncompressed_file(fileext, fido, downloader=downloader)
     else:
         # we are going to assume most files that exist on the hub are
         # compressed in .tar folders. Some may not.
-        fname = fetch_compressed_file(fileext, fido)
+        fname = fetch_compressed_file(fileext, fido, downloader=downloader)
 
     # The `folder_path` variable is used here to notify the user where the
     # files have been unpacked to. However, we can't assume this is reliable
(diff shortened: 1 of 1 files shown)


brittonsmith:
The internal progress bar relies on a copy of `tqdm` living in `yt/extern`. Would that work here instead?

neutrinoceros:
Indead, it should be possible. Atm I don't know how to make this work with pooch, as an optional dep.

brittonsmith:
Oh right, yes, that would make this tough. Ok, then I think how you have it is fine.

matthewturk:
Maybe we should un-vedor tqdm.

brittonsmith:
> Maybe we should un-vedor tqdm.

Also fine with me!

yt/utilities/load_sample.py
```diff
@@ -45,12 +48,21 @@ def load_sample(name=None, specific_file=None):
...
     fileext, name, extension = _validate_sampledata_name(name)
 
+    downloader = None
+    if pbar:
+        try:
+            import tqdm  # noqa: F401
+            from pooch import HTTPDownloader
+            downloader = HTTPDownloader(progressbar=True)
+        except ImportError:
+            mylog.warning("tqdm is not installed, progress bar can not be displayed.")
```
munkm, line 58: I like this message!

yt/utilities/load_sample.py
```diff
@@ -45,12 +48,21 @@ def load_sample(name=None, specific_file=None):
     base_path = fido.fido.path
     fileext, name, extension = _validate_sampledata_name(name)
 
+    downloader = None
+    if pbar:
+        try:
+            import tqdm  # noqa: F401
+            from pooch import HTTPDownloader
```
munkm, line 55: I tried your PR and it works! However, if pooch isn't importing properly the error message will still default to the tqdm one. Is there a way that this can be made clearer? 


Also maybe you want to use the on demand import of pooch defined at line 12?

neutrinoceros: I don't think this can happen since if you attempt to use `yt.load_sample` while pooch isn't installed, it will raise :
```python
ImportError: This functionality requires the pooch package to be installed.
```
whether or not tqdm is installed. (I just checked this :))

neutrinoceros: And I just switched to the on-demand-import for pooch, thanks for catching this !

yt/utilities/load_sample.py
```diff
@@ -45,12 +48,21 @@ def load_sample(name=None, specific_file=None):
     base_path = fido.fido.path
     fileext, name, extension = _validate_sampledata_name(name)
 
+    downloader = None
+    if pbar:
+        try:
+            import tqdm  # noqa: F401
+            from pch import HTTPDownloader
+            downloader = HTTPDownloader(progressbar=True)
```
munkm, line 56:
```suggestion
            downloader = pch.pooch.HTTPDownloader(progressbar=True)
```

yt/utilities/load_sample.py
munkm: I made a suggestion to avoid any pooch-related imports, just in case! I think you'll need to use pch.pooch to access the HTTPDownloader.

neutrinoceros: approved, it's clearer this way and I like that it avoids imports outside of top-level (except for tqdm which I don't see a solution to)

munkm: Yeah I think the tqdm import here is a good solution. 🙂  I can't think of an alternative solution either.

neutrinoceros:
> Maybe we should un-vedor tqdm.
Actually now I'm thinking if we want to do this, maybe the 4.0 release is a good time for it ?

PR #2610, opened 2020-05-27
Volume rendering for octrees

cphyc:
This PR adds support for doing volume rendering for octree-based datasets (e.g. RAMSES). It is built on top of #2425 to provide vertex-centred data (not yet wired though).

It follows the same idea as #2442, but relying instead with an independent octree ray tracer. This has the advantage of making the code much simpler, but generates a bit of code repetition and requires rebuilding an octree.

Currently, it supports the yt API for volume rendering.

## Features
- [x] support vertex centered data for trilinear interpolation
- [x] wire with yt volume rendering API
- [x] can handle any geometry, takes care of CPU domains nicely
- [x] handle different AMR levels
- [x] handle "holes" in the dataset (this means you can use an *arbitrary* data source for the volume renderer)
- [x] Deallocate properly all the stuff (i.e. find the origin of the memory leak)
- [x] Possible to improve "aliasing" effect? See image of galaxy in next post.
- [x] Merge #2425 before merging this one.

## Demo
```python
import yt
ds = yt.load_sample('output_00080', 'info_00080.txt')
sp = ds.all_data()
sc = yt.create_scene(sp, lens_type='perspective')
source = sc[0]

source.tfh.set_bounds((3e-31, 5e-27))
source.tfh.grey_opacity = True
source.tfh.plot('transfer_function.png', profile_field='density')

sc.save('rendering.png', sigma_clip=4)
```

rendering

## Notes on performance

The entire script above takes about 20s to run from end to end. Casting the rays and integrating the transfer function actually only takes 4s on 8 cores (with hyperthreading), 5s on 4 cores and 16s on a single core. The dataset consists of 1,749,455 cells.


cphyc:
After quite a long time working on this, I think this should be ready for review! There's probably ways to make all this cleaner, so don't hesitate!

It is also probably still missing some tests, @matthewturk any suggestion regarding the way to go? Duplicate all 3d rendering tests already happening with grid-based datasets?

cphyc:
There seems to be a compilation issue caused with an older version of gcc used on jenkins. Any guess how to solve this?

yt/fields/geometric_fields.py
```diff
@@ -94,9 +94,10 @@ def _zeros(field, data):
     def _ones(field, data):
         """Returns one for all cells"""
         arr = np.ones(data.ires.shape, dtype="float64")
+        tmp = data.apply_units(arr, field.units)
```
neutrinoceros, line 97: I find it weird that `tmp` is the return variable while `arr` never leaves the function. Wouldn't it be more natural the other way around ?

yt/frontends/ramses/data_structures.py
```diff
@@ -197,13 +214,95 @@ def fill(self, fd, fields, selector, file_handler):
...
+                   file_inds, ndim, all_fields, fields, tr,
+                   oct_handler,
+                   domains=domains)
+        return tr
+
+    @property
+    def fwidth(self):
+        fwidth = super(RAMSESDomainSubset, self).fwidth
+        if self._num_ghost_zones > 0:
+            fwidth = super(RAMSESDomainSubset, self).fwidth.reshape(-1, 8, 3)
```
neutrinoceros, line 261:
```suggestion
            fwidth = fwidth.reshape(-1, 8, 3)
```

yt/frontends/ramses/data_structures.py
```diff
@@ -254,13 +353,14 @@ def _detect_output_fields(self):
...
     def _identify_base_chunk(self, dobj):
+        ngz = dobj._num_ghost_zones
         if getattr(dobj, "_chunk_info", None) is None:
             domains = [dom for dom in self.domains if
                        dom.included(dobj.selector)]
             base_region = getattr(dobj, "base_region", dobj)
             if len(domains) > 1:
                 mylog.debug("Identified %s intersecting domains", len(domains))
-            subsets = [RAMSESDomainSubset(base_region, domain, self.dataset)
+            subsets = [RAMSESDomainSubset(base_region, domain, self.dataset, num_ghost_zones=ngz)
```
neutrinoceros, line 363:
```suggestion
            subsets = [RAMSESDomainSubset(base_region, domain, self.dataset, num_ghost_zones=dobj._num_ghost_zones)
```

yt/frontends/ramses/data_structures.py
```diff
@@ -254,13 +353,14 @@ def _detect_output_fields(self):
         self.field_list = self.particle_field_list + self.fluid_field_list
 
     def _identify_base_chunk(self, dobj):
+        ngz = dobj._num_ghost_zones
```
neutrinoceros, line 356:
```suggestion
```

yt/frontends/ramses/data_structures.py
cphyc: I'm addressing this one on #2425 where it first appeared.

yt/frontends/ramses/data_structures.py
cphyc: I'm addressing this one on #2425 where it first appeared.

yt/frontends/ramses/data_structures.py
cphyc: See above

yt/frontends/ramses/io_utils.pyx
```diff
@@ -74,6 +74,8 @@ def read_amr(FortranFile f, dict headers,
                                     count_boundary = 1)
                 if n > 0:
                     max_level = max(ilevel - min_level, max_level)
+                if n != ng:
+                    raise Exception('Expected %s octs, got %s' % (ng, n))
```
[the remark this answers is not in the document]
cphyc: I guess so. I'm changing it in #2425 as well.

cphyc:
I made some high-res images of a high-resolution galaxy (dx ~ 30pc) and I am bit dubious about the "boxy" rendering.

In particular, I don't know if this is due to the dataset or from a mistake in one of 2,000 lines I wrote… I am now taking bets as where the error (if any) is!

rendering_sigma_2

cphyc:
Converted this to draft because we first need to merge #2425!

cphyc:
There's a massive performance hit that will be resolved by 24a7343 (from #2735). Otherwise I guess it's good to go!

yt/visualization/volume_rendering/render_source.py
```diff
@@ -114,6 +119,18 @@ def set_zbuffer(self, zbuffer):
...
+    if isinstance(data_source, Dataset):
+        ds = data_source
+    else:
+        ds = data_source.ds
+    index_class = ds.index.__class__
+    if issubclass(index_class, GridIndex):
+        return KDTreeVolumeSource(data_source, field)
+    elif issubclass(index_class, OctreeIndex):
+        return OctreeVolumeSource(data_source, field)
+
```
neutrinoceros, line 132:
```suggestion
    else:
        raise NotImplementedError
```

yt/visualization/volume_rendering/render_source.py
```diff
@@ -465,28 +540,96 @@ def render(self, camera, zbuffer=None):
...
+        if self._volume is None:
+            mylog.info("Creating volume")
+            volume = OctreeRayTracing(self.data_source)
+
+            data = self.data_source
+            ds = data.ds
+
+            xyz = np.stack(
+                [data[_].to("unitary").value for _ in "x y z".split()], axis=-1
```
neutrinoceros, line 571:
```suggestion
                [data[key].to("unitary").value for key in "xyz"], axis=-1
```

yt/visualization/volume_rendering/render_source.py
```diff
@@ -465,28 +540,96 @@ def render(self, camera, zbuffer=None):
...
-            image = self.volume.reduce_tree_images(image, camera.lens.viewpoint)
-        image.shape = camera.resolution[0], camera.resolution[1], 4
-        # If the call is from VR, the image is rotated by 180 to get correct
-        # up direction
-        if not self.transfer_function.grey_opacity:
-            image[:, :, 3] = 1
-        return image
+        self.zbuffer = zbuffer
+        self.set_sampler(camera)
+        assert self.sampler is not None
```
neutrinoceros, line 604:
```suggestion
        if self.sampler is None:
            raise RuntimeError(message...)
```

yt/utilities/lib/pyoctree_raytracing.py
```diff
@@ -0,0 +1,101 @@
...
+    _cell_index = None
+    _tvalues = None
+
+    def __init__(self, data_source):
+        self.data_source = data_source
+        LE = np.array([0, 0, 0], dtype=np.float64)
+        RE = np.array([1, 1, 1], dtype=np.float64)
+        depth = data_source.ds.parameters["levelmax"]
+
+        self.octree = CythonOctreeRayTracing(LE, RE, depth)
```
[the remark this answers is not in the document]
cphyc: It _is_ very odd. We should discuss better options. @matthewturk @munkm what's your opinion? The issue is that my implementation is multi-layered, there's a C++ implementation that's wrapped in a cython file that's wrapped in a python file... So we could probably get rid of most of these layers, but which ones?

yt/visualization/volume_rendering/render_source.py
```diff
@@ -418,6 +431,68 @@ def set_sampler(self, camera, interpolated=True):
...
+            you are manually calling render on multiple sources. Scene.render
+            uses this internally.
+
+        Returns
+        -------
+        A :class:`yt.data_objects.image_array.ImageArray` instance containing
+        the rendered image.
+
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
cphyc: No, this is now meant to be an abstract class that's implemented independently for patch-based and oct-based datasets.

yt/visualization/volume_rendering/render_source.py
```diff
@@ -465,28 +540,96 @@ def render(self, camera, zbuffer=None):
...
-        return disp
+        data = self.data_source
+
+        dx = data["dx"].to("unitary").value[:, None]
+        xyz = np.stack([data[_].to("unitary").value for _ in "x y z".split()], axis=-1)
+        LE = xyz - dx / 2
+        RE = xyz + dx / 2
+
+        mylog.debug("Gathering data")
+        dt = np.stack([_ for _ in self.volume.data] + [*LE.T, *RE.T], axis=-1).reshape(
```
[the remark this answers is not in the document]
cphyc: I guess so!

yt/frontends/ramses/data_structures.py
```diff
@@ -347,20 +347,20 @@ def fill(self, fd, fields, selector, file_handler):
             )
 
     def retrieve_ghost_zones(self, ngz, fields, smoothed=False):
-        new_subset = getattr(self, "_subset_with_gz", None)
-        if not new_subset:
+        try:
```
neutrinoceros, line 350: Love the EAPF pattern I'm seeing here !

yt/visualization/volume_rendering/render_source.py
```diff
@@ -465,28 +540,96 @@ def render(self, camera, zbuffer=None):
...
+            1, len(dx), 14, 1
+        )
+        mask = np.full_like(dt[0, ...], 1, dtype=np.uint8)
```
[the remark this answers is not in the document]
cphyc: It would. It would actually be better like
```suggestion
        mask = np.full(dt.shape[1:], 1, dtype=np.uint8)
```
wouldn't it?

yt/visualization/volume_rendering/render_source.py
```diff
@@ -465,28 +540,96 @@ def render(self, camera, zbuffer=None):
...
+        dims = np.array([1, 1, 1])
```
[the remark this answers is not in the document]
cphyc: I'm going for 
```suggestion
        dims = np.array([1, 1, 1], dtype=int)
```

yt/visualization/volume_rendering/render_source.py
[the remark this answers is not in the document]
neutrinoceros: then you could make it a subclass of `abc.ABC` and mark this method as `@abc.abstractmethod` for clarity :)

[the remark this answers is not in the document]
cphyc: This should be done now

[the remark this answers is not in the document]
neutrinoceros: thanks !

cphyc:
The failing test are real; I'll have to dig into it.

cphyc:
I converted it to draft as the failing tests are real and I won't have time to work on it until Sept.

yt/utilities/lib/pyoctree_raytracing.py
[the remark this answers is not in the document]
cphyc: I changed the naming so that the python file is named `octree_raytracing.py`, the cython wrapper is `_octree_raytracing.p{xd,yx}` and the C++ file is `_octree_raytracing.hpp`.

setup.py
```diff
@@ -45,6 +46,11 @@
 else:
     std_libs = ["m"]
 
+if get_default_compiler == "msvc":
```
neutrinoceros:
```suggestion
if get_default_compiler() == "msvc":
```

cphyc:
This is now ready to go!

matthewturk (approved):
I think we should do it.  One slight reservation is that we're changing the appveyor image,b ut that's not necessarily a big deal.

cphyc:
> I think we should do it. One slight reservation is that we're changing the appveyor image, but that's not necessarily a big deal.

I am actually unsure this is still necessary. Let me try using the old one.

cphyc:
> I think we should do it. One slight reservation is that we're changing the appveyor image,b ut that's not necessarily a big deal.

I reverted the change of the appveyor image, as it was not necessary anymore!

neutrinoceros:
@yt-fido test this please

PR #2548, opened 2020-04-17
Saving Raw Answer Test Arrays in Pytest Framework

jcoughlin11:
## PR Summary

This PR adds the ability to save the actual raw arrays produced by answer tests in addition to the hashes.

TODO:

- [x] Check projection_values, grid_values, and pixelized_projection_values, as they are pre-hashed due to python-specific anchors.

## PR Checklist

- [ x] Code passes flake8 checker
- [ x] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #2611, opened 2020-05-28
Port grid frontends

jcoughlin11:
## PR Summary
Implements the answer tests for the Enzo frontend in the pytest framework. This includes saving both the hashed answers as well as the raw data products produced by the tests.

This is a proof-of-concept before adding in the rest of the grid-based frontends in order to facilitate easier review.

This change exists alongside the nose tests, so there should be no disruption on that front.

This PR is part of the larger effort to port yt's testing framework from nosetests to pytest. 

## PR Checklist

This is a continuation of PR #2548


Xarthisius:
> @Xarthisius The python 3.8 build is failing on Jenkins because nose is trying to access some test-related code that is meant for pytest only. For travis and appveyor I was able to modify the config files, but I don't have access to this for Jenkins. All I did was add the `--ignore-files test_outputs_pytest` option to nosetests. If it's not too much trouble, would you mind adding that option when you get a chance? I think that should be all that's needed to get this PR to pass all of the checks. Thank you!

You actually have access and you can make that change a part of this PR. See:

https://github.com/yt-project/yt/blob/master/tests/nose_runner.py#L83

These are the default options passed to all nose processes. Add whatever you need there.

tests/nose_runner.py
```diff
@@ -80,7 +80,8 @@ def generate_tasks_input():
                       if DROP_TAG not in line])
     tests = yaml.load(data, Loader=yaml.FullLoader)
 
-    base_argv = ['-s', '--nologcapture', '--with-xunit']
+    base_argv = ['-s', '--nologcapture', '--with-xunit',
+        '--ignore-files test_outputs_pytest']
```
Xarthisius, line 84:
```suggestion
        '--ignore-files', 'test_outputs_pytest']
```

neutrinoceros:
Just a quick remark before I start a more in-depth review: I'm seeing this adds py35 answers, but no py38. The latter seems necessary while I don't think the former is.

jcoughlin11:
Hi @neutrinoceros, sorry for the late reply! I'm not sure what you mean about adding py35 but not py38. Sorry if this is a stupid question! Thanks!

neutrinoceros:
I meant python 3.5 VS 3.8

munkm:
Since @jcoughlin11 is reissuing the pytest PRs for yt-4 i'm going to close this out knowing we'll replace it with other PRs.

Issue #2608, opened 2020-05-25
YTEP-0003 conventions for ARTIO 


cavestruz:
### Bug report

**Bug summary**

ARTIO still has variable names that need to be modified to [YTEP-0003 conventions](https://ytep.readthedocs.io/en/master/YTEPs/YTEP-0003.html).  

- [ ]  modify fields such as `HeIII density` -> `He_p2_density` in line 53 of yt/frontends/artio/fields.py
- [ ]  add old field names to yt/fields/field_aliases.py to allow for backwards compatibility using `ds._setup_deprecated_fields()`

Note: Self-assigning with @provider10

**Expected outcome**

To be able to access ARTIO fields with either field name.

cavestruz:
Hi @neutrinoceros,

@provider10 has almost finished all of the changes needed according to the checklist above [in a branch off of our fork](https://github.com/cavestruz/yt/tree/issue/2608/artio-with-ytep-0003-conv).  But, we do want to make sure that folks using the other field names can easily use (and know to use) the back-compatibility functionality.  Might there be any thoughts or suggestions for putting in a specific error message for ARTIO users to let them know that they are missing the if the `ds._setup_deprecated_fields()` if the user has _not_ added it?

Thank you!  
- Camille and Raziq

neutrinoceros:
Hi ! Can you open a PR from this branch ? it would make the review easier since the diff would immediately available.
In any case, I do have an idea how this could be implemented but I would like the opinion of more experienced yt devs on this. I'll add my idea on your PR as a comment for it to be discussed :)

Provider10:
This issue is addressed with [PR 2613](https://github.com/yt-project/yt/pull/2613) and I think should be closed.

neutrinoceros:
Agreed, thanks for reminding me !

PR #2604, opened 2020-05-20
added H2 density in artio frontend

Provider10:
General Summary: Adds one line in yt/frontends/artio/fields.py to include H2 Density for analysis on ARTIO data. 

## PR Summary

H2 was a missing field that previously had no assigned units. 

## PR Checklist

- [x] Code passes flake8 checker
- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #2613, opened 2020-05-29
Modified artio frontend fields to ytep 0003 conv: Issue/2608/artio

cavestruz:
## PR Summary

Modified ARTIO variable names to YTEP-0003 conventions for Issue #2608 

- [x] modify fields such as HeIII density -> He_p2_density in line 53 of yt/frontends/artio/fields.py
- [x] add old field names to yt/fields/field_aliases.py to allow for backwards compatibility using ds._setup_deprecated_fields()

Question:  @provider10 and I would like to make sure that folks using the other field names can easily use (and know to use) the back-compatibility functionality. We are open to any thoughts or suggestions for putting in a specific error message for ARTIO users to let them know that they are missing the if the ds._setup_deprecated_fields() if the user has not added it, which we can add before this PR is merged.

Addresses: Issue #2608

## PR Checklist

### DIFF
--- a/yt/fields/field_aliases.py
+++ b/yt/fields/field_aliases.py
@@ -137,7 +137,6 @@ _field_name_aliases = [
     ("HeI density",                      "He_density"),
     ("HeII density",                     "He_p1_density"),
     ("HeIII density",                    "He_p2_density"),
-#    ("H2 density",                       "H2_density"),
 ]
 
 _field_units_aliases = [
--- a/yt/fields/field_aliases.py
+++ b/yt/fields/field_aliases.py
@@ -132,7 +132,12 @@ _field_name_aliases = [
     ("VorticityRPGrowthTimescale",       "vorticity_radiation_pressure_growth_timescale"),
     ("DiskAngle",                        "theta"),
     ("Height",                           "height"),
+    ("HI density",                       "H_density"),
+    ("HII density",                      "H_p1_density"),
+    ("HeI density",                      "He_density"),
+    ("HeII density",                     "He_p1_density"),
     ("HeIII density",                    "He_p2_density"),
+#    ("H2 density",                       "H2_density"),
 ]
 
 _field_units_aliases = [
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Thanks you for opening this !

>Question: @Provider10 and I would like to make sure that folks using the other field names can easily use (and know to use) the back-compatibility functionality. We are open to any thoughts or suggestions for putting in a specific error message for ARTIO users to let them know that they are missing the if the ds._setup_deprecated_fields() if the user has not added it, which we can add before this PR is merged.

Disclaimer: I am unfamiliar with the way we maintain support to unconventional (violating YTEP-0003) field names.
It appears that they are indeed collected in `yt/fields/field_aliases.py`, in which case I don't see a way to add a warning specifically for ARTIO, but I would at least suggest to add comments to the new lines to clarify they were added in support to ARTIO.

After careful examination my take on this is that the code in `Dataset.setup_deprecated_fields` should be revised to include the warning, but this would affect all frontends equally so I believe it goes beyond the scope of this PR.

I would like to request a second opinion from someone more experienced. @brittonsmith, since you were the one to mention the violation of YTEP-0003 originally, could you give any feedback ?

Provider10:
@neutrinoceros -@cavestruz and I are considering opening a new issue for the deprecated fields warning, unless someone else wanted to do that.

Also @brittonsmith and @matthewturk: can both this PR and PR [#2604](https://github.com/yt-project/yt/pull/2604) be a squashed and merged to master so I can do analysis from the master branch?

brittonsmith:
Sorry for taking so long to weigh in on this. The fields in `yt/fields/field_aliases.py` are yt-2.x field names that we probably should have removed altogether a long time ago. That said, I think putting them there is the quickest way to getting this PR merged as soon as possible. Adding machinery to allow frontend-specific deprecated fields should be relatively straightforward, but should not hold this up.

I'm in favor of merging this and creating an issue to remind ourselves that the new fields added to `_field_name_aliases` should be moved to the artio frontend if and when frontend-specific machinery for deprecated fields is implemented.

@neutrinoceros, I will approve this and if you agree, you can be the third approval and merge this.

brittonsmith:
Let's also merge PR #2604 as well and maybe someone can follow up with a quick PR to add the correct field alias for `H2 density`. I'm pretty much out of time for today. @neutrinoceros, can you take care of this?

neutrinoceros:
> maybe someone can follow up with a quick PR to add the correct field alias for H2 density
not sure I understand what remains to be done actually. I'm okay to merge this however !

PR #2615, opened 2020-06-04
Support vertex centred data

cphyc:
Adds support for vertex centred data. This has the effect of making isocontours possible on octree datasets using the same API as for the rest of yt.

This should be merged after #2425 has been merged. I marked it as draft to prevent people from reviewing it, since many files have been modified in the aforementioned PRs.

This allows for example to export a dataset to [sketchfab](https://skfb.ly/6SXyJ).

Note that in its current form, the isosurface algorithm does work. However, it is very slow as it performs the marching cube algorithm on each _individual oct_. It also returns a non-connected set of points (that represent the isosurface), since the output of the marching cube algorithms aren't stitched together. Note that a quick hack would be to merge vertices of the isosurface if they are very close one from another. A better solution however would be to feed sets of connected octs that live at the same level to the marching cube algorithms, but this is well beyond the scope of the current PR.

### DIFF
--- a/yt/data_objects/octree_subset.py
+++ b/yt/data_objects/octree_subset.py
@@ -104,7 +104,7 @@ class OctreeSubset(YTSelectionContainer):
 
     def select_blocks(self, selector):
         mask = self.oct_handler.mask(selector, domain_id=self.domain_id)
-        slicer = OctreeSubsetBlockSlice(self)
+        slicer = OctreeSubsetBlockSlice(self, self.ds)
         for i, sl in slicer:
             yield sl, np.atleast_3d(mask[i, ...])
 
@@ -528,6 +528,7 @@ class OctreeSubsetBlockSlicePosition:
         self.block_slice = block_slice
         nz = self.block_slice.octree_subset.nz
         self.ActiveDimensions = np.array([nz, nz, nz], dtype="int64")
+        self.ds = block_slice.ds
 
     def __getitem__(self, key):
         bs = self.block_slice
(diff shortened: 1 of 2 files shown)


cphyc:
@yt-fido test this please.

cphyc:
@yt-fido test this please

cphyc:
This one's ready to go!

yt/data_objects/octree_subset.py
```diff
@@ -572,22 +573,67 @@ def dds(self):
...
+                nocts = orig_field.shape[-1]
+                new_field = np.zeros((3, 3, 3, nocts), order="F")
+                new_field += orig_field[1:, 1:, 1:]
+                new_field += orig_field[:-1, 1:, 1:]
+                new_field += orig_field[1:, :-1, 1:]
+                new_field += orig_field[1:, 1:, :-1]
+                new_field += orig_field[:-1, 1:, :-1]
+                new_field += orig_field[1:, :-1, :-1]
+                new_field += orig_field[:-1, :-1, 1:]
+                new_field += orig_field[:-1, :-1, :-1]
```
neutrinoceros, line 628: This suggestion wasn't trivial for me to write, and is maybe not that much of an improvement in terms of readability, but I do think it is slightly more worthy of confidence since it either gets everything wrong or everything right.
```suggestion
                new_field = np.zeros((3, 3, 3, nocts), order="F")
                from itertools import product, repeat
                bounds = (slice(1, None), slice(None, -1))
                for s1, s2, s3 in product(*repeat(bounds, 3)):
                    new_field += orig_field[s1, s2, s3]
```

yt/frontends/ramses/data_structures.py
```diff
@@ -349,9 +349,31 @@ def fill(self, fd, fields, selector, file_handler):
...
+                "ignoring it."
+            )
+            smoothed = False
+
+        try:
+            new_subset = self._subset_with_gz
+            mylog.debug(
+                "Reusing previous subset with ghost zone for domain %s", self.domain_id
+            )
+        except AttributeError:
```
neutrinoceros, line 365: nice pattern !

cphyc:
> I left a refactoring suggestion.
> Unfortunately it's not really clear to me how this goes down to the user. Is this new code already hit by existing tests ? If so, which ones ?

This add support for computing e.g. isocontours for RAMSES datasets. This could in principle be extended to any oct-based dataset, but that'd require extra work.

Issue #2616, opened 2020-06-05
ParticleSubselectionTests for yt4-merge


matthewturk:
We need to add independent checks of the particle machinery in addition to answer tests:

 - [x] eagle
 - [x] gadget_fof
 - [x] flash (particle)
 - [x] owls
 - [x] rockstar
 - [x] swift
 - [x] tipsy
 - [x] ytdata (?)
 - [x] sph
 - [x] open_pmd
 - [x] art
 - [x] ahf
 - [x] arepo

Here is an example of doing this: https://github.com/yt-project/yt/blob/b20c3edbbc5b057077d12dc456974f797ce726f6/yt/frontends/gadget/tests/test_outputs.py#L111

jzuhone:
I can take some of these on.

matthewturk:
I will take gadget_fof, rockstar, tipsy and open_pmd

matthewturk:
I went ahead and implemented all of the remaining ones.

munkm:
I think this is fixed

PR #2617, opened 2020-06-05
Abstract some particle tests

matthewturk:
This starts to pull some code out of the particle tests specifically and putting them in the base class, as per #2616 .

Please feel free to push to this branch.

### DIFF
--- a/yt/frontends/owls/tests/test_outputs.py
+++ b/yt/frontends/owls/tests/test_outputs.py
@@ -1,7 +1,8 @@
 from collections import OrderedDict
 
 from yt.testing import \
-    requires_file
+    requires_file, \
+    ParticleSelectionComparison
 from yt.utilities.answer_testing.framework import \
     requires_ds, \
     data_dir_load, \
@@ -26,6 +27,8 @@ _fields = OrderedDict(
 @requires_ds(os33, big_data=True)
 def test_snapshot_033():
     ds = data_dir_load(os33)
+    psc = ParticleSelectionComparison(ds)
+    psc.run_defaults()
     for test in sph_answer(ds, 'snap_033', 2*128**3, _fields):
         test_snapshot_033.__name__ = test.description
         yield test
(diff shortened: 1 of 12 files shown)


matthewturk:
In this PR I've abstracted the basic tests that I used in gadget, so now it should be sufficient to get decent coverage by just calling `run_defaults`.

yt/testing.py
```diff
@@ -1269,3 +1269,47 @@ def compare_dobj_selection(self, dobj):
                 obj_results.append(chunk[ptype, "particle_position"])
             obj_results = np.concatenate(obj_results, axis = 0)
             assert_equal(sel_pos, obj_results)
+
+    def run_defaults(self):
```
munkm, line 1273: I like the addition of creating this default function.

yt/testing.py
```diff
@@ -1269,3 +1269,47 @@ def compare_dobj_selection(self, dobj):
...
+        """
+        This runs lots of samples that touch different types of wraparounds.
```
munkm, line 1275: Can you either add inline comments or to this docstring the specific comparisons that are being made. That way a future frontend writer will know what selectors they need to have coverage for if they want to use non-default comparisons in their testing.

yt/frontends/owls/tests/test_outputs.py
```diff
@@ -26,6 +27,8 @@
 @requires_ds(os33, big_data=True)
 def test_snapshot_033():
     ds = data_dir_load(os33)
+    psc = ParticleSelectionComparison(ds)
```
munkm, line 30: So @jzuhone doesn't need to do the owls, flash, eagle, or arepo tests anymore?

matthewturk: right!

yt/testing.py
matthewturk: Sure.

matthewturk:
They aren't necessary after all.

PR #2621, opened 2020-06-08
Update bitmask_version for #2575

matthewturk:
I thought I had incremented the bitmask version in #2575, but I had not.  This needs to go in pretty quickly as the bitmasks all need to be re-generated.


munkm:
Once the tests pass I'm happy to merge this.

PR #2618, opened 2020-06-06
[yt-4.0] added domain_overrride to fix bbox

AshKelly:
## PR Summary

Added a `ds.domain_override` when a `bbox` is applied to a dataset. This prevents the particle index from caching which ensures the particle selection is correct. Fixes #2612 

## PR Checklist

- [x] Code passes flake8 checker
- *NA* New features are documented, with docstrings and narrative docs
- *NA* Adds a test for any bugs fixed. Adds tests for new features.

PR #2624, opened 2020-06-09
Update TNG tests and add another dont_cache check for loading

matthewturk:
This is the next step after #2618, with one remaining.  This checks whether or not an `index_filename` has been provided; if it has, it overrides the `dont_cache` for whether we should load from the files.

This covers the case that we aren't caching, but an index file exists, and it also covers for if an index file exists and is specified.  It's not yet possible to create an index file since `dont_cache` doesn't check `index_filename`.

Tests for this will be forthcoming, once I've stabilized the yt4-merge in #2437 and the Arepo tests there.

### DIFF
--- a/yt/frontends/arepo/tests/test_outputs.py
+++ b/yt/frontends/arepo/tests/test_outputs.py
@@ -1,7 +1,7 @@
 import os
 import tempfile
 from collections import OrderedDict
-from yt.testing import requires_file
+from yt.testing import requires_file, ParticleSelectionComparison
 from yt.utilities.answer_testing.framework import \
     data_dir_load, \
     requires_ds, \
@@ -17,6 +17,12 @@ _tng59_bbox = [[45135.0, 51343.0], [51844.0, 56184.0], [60555.0, 63451.0]]
 def test_arepo_hdf5():
     assert isinstance(data_dir_load(bullet_h5),
                       ArepoHDF5Dataset)
+
+@requires_file(bullet_h5)
+def test_arepo_hdf5_selection():
+    ds = data_dir_load(bullet_h5)
+    psc = ParticleSelectionComparison(ds)
+    psc.run_defaults()
 
 
 bullet_fields = OrderedDict(
@@ -76,3 +82,9 @@ def test_index_override():
                                            'bounding_box': _tng59_bbox})
     ds.index
     assert len(open(tmpname, "r").read()) == 0
+
+@requires_file(tng59_h5)
+def test_arepo_tng59_selection():
+    ds = data_dir_load(tng59_h5)
+    psc = ParticleSelectionComparison(ds)
+    psc.run_defaults()
--- a/yt/frontends/arepo/tests/test_outputs.py
+++ b/yt/frontends/arepo/tests/test_outputs.py
@@ -1,3 +1,5 @@
+import os
+import tempfile
 from collections import OrderedDict
 from yt.testing import requires_file
 from yt.utilities.answer_testing.framework import \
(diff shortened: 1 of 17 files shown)


matthewturk:
So this *also* updates the TNG Halo test to use a bounding box that includes particles.  I'll change the name.

munkm:
@AshKelly I added you to review this since you've been thinking about this too.

yt/frontends/arepo/tests/test_outputs.py
```diff
@@ -55,12 +58,21 @@ def test_tng_hdf5():
     ]
 )
 
-
 @requires_ds(tng59_h5)
 def test_arepo_tng59():
-    ds = data_dir_load(tng59_h5)
+    ds = data_dir_load(tng59_h5, kwargs = {'bounding_box': _tng59_bbox})
     for test in sph_answer(ds, 'halo_59', 10107142,
                            tng59_fields):
```
AshKelly, line 65: How come adding the bounding box didn't change the number of particles, `10107142`?

yt/geometry/particle_geometry_handler.py
```diff
@@ -128,13 +129,14 @@ def _initialize_index(self):
...
             fname = ds.index_filename
 
+        dont_load = dont_cache and not hasattr(ds, 'index_filename')
         try:
             rflag = self.regions.load_bitmasks(fname)
             rflag = self.regions.check_bitmasks()
             self._initialize_frontend_specific()
-            if rflag == 0:
+            if dont_load or rflag == 0:
                 raise OSError
```
AshKelly, line 138: If I follow this: in the case that we don't want to load, `dont_load = True` we still try and load the bitmasks, but then we raise an `OSError` which triggers the exception and thus, we then initialize from scratch? 

If so, I guess this is ok, because loading is fast.

yt/geometry/particle_geometry_handler.py
```diff
@@ -128,13 +129,14 @@ def _initialize_index(self):
...
-        except OSError:
+        except (OSError, struct.error):
```
AshKelly, line 139: What can cause a `struct.error` here? I presume something inside `load_bitmasks`?

yt/frontends/arepo/tests/test_outputs.py
matthewturk: Great question.  I set up the bounding box to include all of the particles.  The TNG Halo we have has a *domain* set to be the full domain of the TNG simulation, but it's just one halo.  So things like `"c"` were going to the center of the domain, the `unitary` unit was a bit strange, etc.  Now using this, it's constrained to just the location of the halo.

yt/geometry/particle_geometry_handler.py
matthewturk: That was the idea, yeah, but I think maybe it's not the best idea anymore.  It might need to be worked through a bit more carefully.

yt/geometry/particle_geometry_handler.py
matthewturk: Yup, if the file is mal-formed (i.e., under 8 bytes).  That shows up in the tests, for instance.

yt/geometry/particle_geometry_handler.py
AshKelly: I guess you could put the `if dont_load: raise OSError` straight after the `try` and then at the end of the `try` have the `if flag == 0: raise OSError` - otherwise `yt` outputs:
`Loading particle index...`
`Generating particle index...`

matthewturk: good point, will do this

matthewturk:
Updated as per @AshKelly 's suggestion

PR #2627, opened 2020-06-09
Adding nbody answer tests using particle plots

matthewturk:
This adds answer tests for the nbody-only dataset in FLASH, and will allow other nbody-only answer tests to be added as well.

This uses the `ParticleProjectionPlot` to generate the data.  This is a blocker for #2437 but needs to be reviewed independently.

### DIFF
--- a/yt/utilities/answer_testing/framework.py
+++ b/yt/utilities/answer_testing/framework.py
@@ -983,7 +983,7 @@ def big_patch_amr(ds_fn, fields, input_center="max", input_weight="density"):
                         dobj_name)
 
 
-def _particle_answer(ds, ds_str_repr, ds_nparticles, fields, proj_test_class):
+def _particle_answers(ds, ds_str_repr, ds_nparticles, fields, proj_test_class):
     if not can_run_ds(ds):
         return
     assert_equal(str(ds), ds_str_repr)
--- a/yt/utilities/answer_testing/framework.py
+++ b/yt/utilities/answer_testing/framework.py
@@ -556,6 +556,13 @@ class PixelizedProjectionValuesTest(AnswerTestingTest):
         self.weight_field = weight_field
         self.obj_type = obj_type
 
+    def _get_frb(self, obj):
+        proj = self.ds.proj(self.field, self.axis,
+                              weight_field=self.weight_field,
+                              data_source = obj)
+        frb = proj.to_frb((1.0, 'unitary'), 256)
+        return proj, frb
+
     def run(self):
         if self.obj_type is not None:
             obj = create_obj(self.ds, self.obj_type)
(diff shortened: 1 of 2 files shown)


yt/utilities/answer_testing/framework.py
```diff
@@ -969,6 +983,31 @@ def big_patch_amr(ds_fn, fields, input_center="max", input_weight="density"):
...
+    assert_equal(dd["particle_position"].shape, (ds_nparticles, 3))
+    tot = sum(dd[ptype, "particle_position"].shape[0]
+              for ptype in ds.particle_types_raw)
+    assert_equal(tot, ds_nparticles)
+    for dobj_name in dso:
+        for field, weight_field in fields.items():
+            if field[0] in ds.particle_types:
+                particle_type = True
+            else:
+                particle_type = False
```
AshKelly, line 1001:
```suggestion
            particle_type = field[0] in ds.particle_types
```

yt/utilities/answer_testing/framework.py
```diff
@@ -969,6 +983,31 @@ def big_patch_amr(ds_fn, fields, input_center="max", input_weight="density"):
...
+            for axis in [0, 1, 2]:
+                if not particle_type:
+                    yield PixelizedParticleProjectionValuesTest(
+                        ds, axis, field, weight_field,
+                        dobj_name)
+            yield FieldValuesTest(ds, field, dobj_name,
+                                  particle_type=particle_type)
+
+
```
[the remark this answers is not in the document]
matthewturk: Ah, field values test just checks that the field values are all the same.  It's not super sophisticated.

yt/utilities/answer_testing/framework.py
```diff
@@ -969,6 +983,28 @@ def big_patch_amr(ds_fn, fields, input_center="max", input_weight="density"):
                         dobj_name)
 
 
+def nbody_answer(ds, ds_str_repr, ds_nparticles, fields):
```
neutrinoceros, line 986: so if I'm reading this right, this is almost strictly equivalent to the old `sph_answer` routine, with the only difference being the class of the projection test.
Wouldn't it be clearer to rewrite these functions as follow ?

```python
def nbody_answer(ds, ds_str_repr, ds_nparticles, fields,
    projection_value_class=PixelizedParticleProjectionValuesTest):
    ...

def sph_answer(ds, ds_str_repr, ds_nparticles, fields):
    return nbody_answer(ds, ds_str_repr, ds_nparticles, fields,
        projection_value_class=PixelizedProjectionValuesTest)
```

matthewturk: yup it would, incoming...

matthewturk:
@AshKelly @neutrinoceros updated with the suggested refactoring

AshKelly:
LGTM!

Issue #1467, opened 2017-06-23
Make ambiguous field accesses an error


ngoldbaum:
Right now yt lets you use a string field name instead of a full field tuple to access a field. To translate from the field name to a field tuple it uses a set of heuristics to guess the field type corresponding to the field name.

Unfortunately, sometimes it guesses wrong, or worse, the guess changes based on IO patterns. For example:

```
In [8]: ds._get_field_info('particle_mass')
Out[8]: Alias Field for "('PartType0', 'Masses')" (PartType0, particle_mass): (units: code_mass, particle field)

In [9]: ad['all', 'particle_mass']
Out[9]:
YTArray([  8.93692344e+39,   1.07352366e+40,   9.48756720e+39, ...,
           8.89038231e+39,   8.87535665e+39,   8.91641743e+39]) g

In [10]: ds._get_field_info('particle_mass')
Out[10]: Alias Field for "('all', 'Masses')" (all, particle_mass): (units: g, particle field)
```

(`ds._get_field_info` is a private function used to facilitate this field guessing system)

I'd like to propose a backward incompatible change. Rather than doing this guessing business and sometimes being wrong or sometimes switching which field we return based on the history of field accesses, we should instead unconditionally raise an error if a field access is ambiguous.

Concurrently, we should also update our documentation to use either field tuples or access fields using the `ds.fields` object,  (like `ds.fields.gas.density` instead of `('gas', 'density')`).

I'm not sure how disruptive this is but I bet there are a bunch of places inside of yt where we are doing ambiguous field accesses.

qobilidop:
Could we also add a feature to temporarily set a default field type to work with? Because it's very common to do analysis on a single field type. And it would be tedious and error-prone having to type it explicitly every time.

I'm proposing a context manager API:

```
with ad.default_field_type('all'):
    print(ad['particle_mass'])
```

And except for within the context, field access using the short-hand string form would err.

chummels:
That's an interesting idea. So inside the context manager, you could use strings as fields or also tuples of fields with other ftypes, but outside the context manager, you could only use tuples as fields?

That seems pretty useful for shorthanding.

chummels:
This topic was just discussed at a yt-4.0 release meeting. It was agreed that in yt-4.0, we will accept both string and tuple field names. If someone specifies a string field name and there is any ambiguity if there are two potential tuple fields that could be accessed, the code will error out with an error message requiring the user to be more explicit.

[...]

PR #2637, opened 2020-06-12
Fields autocompletion for data objects

### LINKED ISSUE #2397: Add tab-completion for fields
Data objects all define a `__getitem__` method, which accepts fields from the dataset `field_list` and `derived_field_list`.  This list can be enormously long, however.  IPython (5.0+) defines a very straightforward way of constructing tab completions which we could utilize:

https://ipython.readthedocs.io/en/stable/config/integrating.html#tab-completion

Defining this on the base object would make it much simpler to navigate large lists of fields.

neutrinoceros:
## PR Summary

fix #2397
This is a simple implementation of the `_ipython_key_completion_` method for data objects.

**example**

Screenshot 2020-06-12 at 20 22 09

**limitation**
I couldn't find how to make it work with the double key syntax yet, e.g

```python
ad = ds.all_data()
ad["gas", "<TAB>
```

### DIFF
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -267,7 +267,11 @@ class YTDataContainer(metaclass = RegisteredDataContainer):
             rv = self.ds.arr(self.field_data[key], fi.units)
         return rv
 
-
+    def _ipython_key_completions_(self):
+        keys = self.ds.field_list + self.ds.derived_field_list
+        types = list(set(k[0] for k in keys))
+        labels = list(set(k[1] for k in keys))
+        return types + labels
 
     def __setitem__(self, key, val):
         """
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -257,14 +257,17 @@ class YTDataContainer(metaclass = RegisteredDataContainer):
         # hanging off the dataset to define this unit object.
         # Note that this is less succinct so that we can account for the case
         # when there are, for example, no elements in the object.
-        rv = self.field_data.get(f, None)
-        if rv is None:
+        try:
+            rv = self.field_data[f]
+        except KeyError:
             if isinstance(f, tuple):
                 fi = self.ds._get_field_info(*f)
             elif isinstance(f, bytes):
                 fi = self.ds._get_field_info("unknown", f)
             rv = self.ds.arr(self.field_data[key], fi.units)
         return rv
+
+
 
     def __setitem__(self, key, val):
         """
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -1,5 +1,6 @@
 import itertools
 import uuid
+from distutils.version import LooseVersion
 
 import numpy as np
 import weakref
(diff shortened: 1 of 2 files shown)


neutrinoceros:
This is loosely related to #1467 and should be kept in mind if we take action there.

cphyc:
That rocks! If I understand correctly, here you're completing with both the field types and the field names, aren't you? I am wondering if we could find a way around this…

neutrinoceros:
Yes, as of now it _will_ suggest completion for the `ftypes` as well as `fnames`, although it feels a bit half baked if you're using it to complete the `ftype` and then ... you're on your own !
The problem is I'm not sure that ipython supports that kind of syntax at all, I've tried adding the tuples `(ftype, fname)` to autocompletion keys to no avail...

neutrinoceros:
So after discussion with @cphyc , we think adding support for 2-keys syntax requires a change upstream in IPython. Corentin came very close to the appropriate change already (some regex magic is required), so I'm hopeful this can happen soon. In any case the present PR is now complete, except if there are base data objects class unrelated to either `YTRegion` and `RegionExpression` that I'm missing.

yt/data_objects/data_containers.py
```diff
@@ -257,15 +257,23 @@ def __getitem__(self, key):
...
                 fi = self.ds._get_field_info("unknown", f)
             rv = self.ds.arr(self.field_data[key], fi.units)
         return rv
 
+    def _ipython_key_completions_(self):
+        # to keep in sync with yt.data_objects.region_expression.RegionExpression
+        keys = self.ds.field_list + self.ds.derived_field_list
+        ftypes = list(set(k[0] for k in keys))
+        fnames = list(set(k[1] for k in keys))
+        return ftypes + fnames
```
cphyc, line 275: If https://github.com/ipython/ipython/pull/12395 is merged, then this could be rewritten simply as 
```suggestion
        return self.ds.field_list + self.ds.derived_field_list
```
and same below for `region_expression`.

neutrinoceros:
awesome ! I'll wait on IPython's devs' answer so we can choose the most appropriate path to add this feature to yt with potentially different levels of support depending on IPython's version.

neutrinoceros:
Converted to draft in the mean time

neutrinoceros:
So it seems that @cphyc ’s PR was accepted for IPython’s next major release 8.0
I’ll patch this one accordingly

neutrinoceros:
> For reference, the worst case scenario for IPythons that do not support tuple completion is that the auto completion won't work (the suggestions as tuples are silently ignored).

Indeed, I've simplified the implementation since it's not actually needed to check for IPython's version to get consistent results !

cphyc:
> > For reference, the worst case scenario for IPythons that do not support tuple completion is that the auto completion won't work (the suggestions as tuples are silently ignored).
> 
> Indeed, I've simplified the implementation since it's not actually needed to check for IPython's version to get consistent results !

Mmmh, I am actually unsure about your last commit, because with IPython 8+, it will provide autocompletion for both the tuple-aware keys and the others (which will result in erroneous suggestions where we could have only good ones). I can't think of a better approach than the one of 43583f4.

neutrinoceros:
>  will result in erroneous suggestions where we could have only good ones

I'm not sure what you mean here. Are you worried ambiguous fnames might show up in the suggestions (already the case without tuple-completion), or do you mean that mixing tuple_keys + fnames will produce meaningless combinations in the suggestions ?

cphyc:
> > will result in erroneous suggestions where we could have only good ones
> 
> I'm not sure what you mean here. Are you worried ambiguous fnames might show up in the suggestions (already the case without tuple-completion), or do you mean that mixing tuple_keys + fnames will produce meaningless combinations in the suggestions ?

Oh I see! I was thinking assuming things like `ad['density']` are "wrong" suggestions, but that's totally valid. Forget about my last comment then.

yt/data_objects/data_containers.py
```diff
@@ -257,15 +265,19 @@ def __getitem__(self, key):
         # hanging off the dataset to define this unit object.
         # Note that this is less succinct so that we can account for the case
         # when there are, for example, no elements in the object.
-        rv = self.field_data.get(f, None)
-        if rv is None:
+        try:
```
matthewturk, line 268: I honestly can't remember why it was this way.  I think this is better, though, and more clear.

PR #2638, opened 2020-06-12
Meaningful errors for add_field methods

### LINKED ISSUE #2700: Confusing error message in ds.add_field()
### Bug report

**Bug summary**

The current implementation for `ds.add_field` is probably somewhat messy (I intentionally didn't look it up before reporting), so it throws confusing and unhelpful errors on bad input.

**Code for reproduction**

```python
import yt

def _test(field, data):
    return data["x"]
ds = yt.testing.fake_amr_ds(fields=["density"])

# the error is that I'm passing "func" instead of "function"
ds.add_field(name="test", func=_test, sampling_type="cell", units="m")
```

**Actual outcome**

```python
TypeError                                 Traceback (most recent call last)
<ipython-input-29-01d8197cb9a9> in <module>
      6     return data["x"].base * unyt.msun
      7 ds = yt.testing.fake_amr_ds(fields=["density"])
----> 8 ds.add_field(name="test", func=_test, sampling_type="cell", units="kg")

~/dev/python/yt-project/yt/yt/data_objects/static_output.py in add_field(self, name, function, sampling_type, **kwargs)
   1271         self.field_info.add_field(name, sampling_type, function=function, **kwargs)
   1272         self.field_info._show_field_errors.append(name)
-> 1273         deps, _ = self.field_info.check_derived_fields([name])
   1274         self.field_dependencies.update(deps)
   1275 

~/dev/python/yt-project/yt/yt/fields/field_info_container.py in check_derived_fields(self, fields_to_check)
    395         for field in fields_to_check:
    396             if field not in self:
--> 397                 raise YTFieldNotFound(str(field))
    398             fi = self[field]
    399             try:

TypeError: __init__() missing 1 required positional argument: 'ds'
```
Not only does this not help me, as a user, to discover what I'm doing wrong, it also looks like a plain bug, which I _thought_ it was since this was my first time using yt since we merged the 4.0 branch.

**Expected outcome**

A more helpful error message such as
```python
TypeError: add_field() missing 1 required positional argument: 'function'

# or
TypeError: add_field() missing 1 required keyword argument: 'function'
```

depending on what is more idiomatic to the current implementation.

**Version Information**
  * yt version: master

neutrinoceros:
## PR Summary

My original motivation for this was that when using the `add_field` dataset method, I always got a warning (`UserWarning: Because 'sampling_type' not specified, yt will assume a cell 'sampling_type'`), which I do not care about since this default behaviour happens to be the one I always want.

I then realised the docstrings were not up to date so I fixed them accordingly.

There are two ways to solve the "issue" of displaying a warning on default behaviour:
- remove the warning and let the users figure out if they are using the method correctly by reading the docstrings (may not be ideal)
- remove the default value (cleaner _but_ breaking change)

Important note: I went with the first option here BUT `changelog.rst` contains directions for this (notes from yt 3.4 release)
> For now a ``'cell'`` ``sampling_type`` is assumed if ``sampling_type`` is not specified but in the future ``sampling_type`` will always need to be specified.

so it seems that the second route was already agreed upon 3yrs ago. I'm happy to implement the breaking version of this change if there is no argument against it.

**changes**

Here I illustrate how this change makes the api more user friendly based on an actual example that I wrote myself as a user
(the fundamental mistake I'm doing here is misremembering the name of the `function`argument and writing `func` instead)
```python
import yt
import numpy as np
ds = yt.testing.fake_amr_ds(fields=["density"])
def _nada(field, data):
    return np.ones_like(data["density"])

# 1 - respecting the argument order 
ds.add_field(name="nada", func=_nada, sampling_type="cell") 
old >> runs, but then ds.r["nada"] fails with YTFieldNotFound
new >> TypeError: add_field() missing 1 required positional argument: 'function'

# 2- reverse position of arguments (func is still broken)
ds.add_field(func=_nada, name="nada2", sampling_type="cell")
old >> YTFieldNotFound : Could not find field 'nada2' in AMRGridData.
new >> TypeError: add_field() missing 1 required positional argument: 'function'
```

fix #2700 (by the number of the issue VS this PR, you can tell I actually reported this _after_ fixing it and forgetting about it)


yt/fields/local_fields.py
```diff
@@ -29,10 +27,6 @@ def add_field(self, name, function=None, sampling_type=None, **kwargs):
             else:
                 sampling_type = "particle"
         if sampling_type is None:
```
cphyc, line 29: We should instead set `sampling_type = "cell"` in the function definition.

yt/data_objects/static_output.py
```diff
@@ -1277,6 +1277,8 @@ def add_field(self, name, function=None, sampling_type=None, **kwargs):
         function : callable
            A function handle that defines the field.  Should accept
            arguments (field, data)
+        sampling_type: str
+           "cell" (default) or "particle"
```
cphyc, line 1281: We should set `sampling_type = "cell"` in the function defintion.

yt/fields/local_fields.py
neutrinoceros: Indeed, I always forget which builtin iterable types are mutable or not, but str isn't so it's okay to use default str values.

neutrinoceros: Actually I think the parameter should have no default value at all in order to embrace the vision that was described at the time of yt 3.4
Right now I'm testing against using "cell" as a default, since the reimplementation should slightly affect the function's behaviour. If all goes well I'll also try to remove the default value all together.

neutrinoceros:
the failing test looks like it's actually _silently_ broken on master too... I really don't understand what's wrong with it but I think I'm more revealing something is broken than breaking new stuff?? I'll get back to this later

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

yt/data_objects/static_output.py
```diff
@@ -1060,7 +1059,7 @@ def _assign_unit_system(self, unit_system):
         self.unit_registry.unit_system = self.unit_system
 
     def _create_unit_registry(self, unit_system):
-        import yt.units.dimensions as dimensions
+        from yt.units import dimensions as dimensions
```
cphyc, line 1062: Is there a specific reason for this import to happen here instead of the beginning of the file?

yt/data_objects/static_output.py
```diff
@@ -1359,27 +1358,12 @@ def add_field(self, name, function=None, sampling_type=None, **kwargs):
             )
         # Handle the case where the field has already been added.
         if not override and name in self.field_info:
-            mylog.error(
+            mylog.warning(
                 "Field %s already exists. To override use " + "force_override=True.",
```
cphyc, line 1362:
```suggestion
                "Field %s already exists. To override use `force_override=True`.",
```

yt/fields/field_info_container.py
```diff
@@ -247,7 +246,50 @@ def setup_fluid_aliases(self, ftype="gas"):
...
+        RuntimeError
+            If conflicting parameters are passed.
+        """
+        try:
+            sampling_type = sampling_type.lower()
+        except AttributeError as e:
+            raise TypeError("sampling_type should be a string.") from e
+
+        if sampling_type not in ("cell", "particle", "local"):
+            raise ValueError
```
cphyc, line 277:
```suggestion
        acceptable_samplings =("cell", "particle", "local") 
        if sampling_type not in acceptable_samplings:
            raise ValueError(
                "Invalid sampling type %s. Valid sampling types are %s",
                sampling_type,
                ", ".join(acceptable_samplings)
            )
```

yt/fields/local_fields.py
```diff
@@ -1,50 +1,47 @@
...
+            name = (ftype, name)
+
         override = kwargs.get("force_override", False)
         # Handle the case where the field has already been added.
         if not override and name in self:
-            mylog.error(
-                "Field %s already exists. To override use " + "force_override=True.",
-                name,
+            mylog.warning(
+                "Field %s already exists. To override use force_override=True.", name,
```
cphyc, line 26: Just to be consistent with the above-mentioned suggestion
```suggestion
               "Field %s already exists. To override use `force_override=True`.",
```

yt/data_objects/static_output.py
neutrinoceros: I don't believe so. Fixing.

PR #2640, opened 2020-06-13
Cleanup some old python 2 idioms

neutrinoceros:
- Inheritance from `object` class is implicit in python 3 so there's no need to keep enforcing it after we dropped support for python 2.

The following were done semi-automatically through `2to3` + manual validation, since in many cases this tool will produce incorrect results (e.g. with sorting numpy arrays)
- cleanup metaclass declarations
- remove future imports
- cleanup rare reliquary occurrences of `long` (-> `int`) and `unicode` (-> `str`)
- ⚠️ two classes were implementing a `next` method instead of `__next__`, this change could break some downstream code if the deprecated syntax `self.next()` was used (should be a niche case and is easily fixed)
- use builtin `a = sorted(data)` instead of `a=list(data); a.sort()` where relevant (i.e. where it doesn't cause performance hits by adding a copy step)
- consistently use `while True` instead of `while 1`
- cleanup type comparisons: use builtin `isinstance(a, mytype)` instead of `type(a) is mytype`.  Within tests I conservatively assumed it's preferable to leave the more strict type comparison but replaced some `==` comparisons with `is` (I don't know if it's even possible to build case where they are not equivalent but the later is definetely correct).

### DIFF
--- a/yt/data_objects/tests/test_octree.py
+++ b/yt/data_objects/tests/test_octree.py
@@ -12,7 +12,7 @@ def test_building_tree():
     '''
     ds = fake_sph_grid_ds()
     octree = ds.octree(n_ref=1)
-    assert(type(octree) == YTOctree)
+    assert(type(octree) is YTOctree)
     assert(octree[('index', 'x')].shape[0] == 456)
 
 def test_saving_loading():
--- a/yt/frontends/swift/tests/test_outputs.py
+++ b/yt/frontends/swift/tests/test_outputs.py
@@ -15,7 +15,7 @@ EAGLE_6 = "EAGLE_6/eagle_0005.hdf5"
 @requires_file(keplerian_ring)
 def test_non_cosmo_dataset():
     ds = load(keplerian_ring)
-    assert(type(ds) == SwiftDataset)
+    assert(type(ds) is SwiftDataset)
 
     field = ('gas', 'density')
     ad = ds.all_data()
--- a/yt/geometry/coordinates/spherical_coordinates.py
+++ b/yt/geometry/coordinates/spherical_coordinates.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+
 import numpy as np
 from .coordinate_handler import \
     CoordinateHandler, \
--- a/yt/geometry/coordinates/spherical_coordinates.py
+++ b/yt/geometry/coordinates/spherical_coordinates.py
@@ -1,4 +1,3 @@
-
 import numpy as np
 from .coordinate_handler import \
     CoordinateHandler, \
--- a/yt/geometry/coordinates/tests/test_cartesian_coordinates.py
+++ b/yt/geometry/coordinates/tests/test_cartesian_coordinates.py
@@ -13,8 +13,7 @@ def test_cartesian_coordinates():
     # We're going to load up a simple AMR grid and check its volume
     # calculations and path length calculations.
     ds = fake_amr_ds()
-    axes = list(set(ds.coordinates.axis_name.values()))
-    axes.sort()
+    axes = sorted(set(ds.coordinates.axis_name.values()))
     for i, axis in enumerate(axes):
         dd = ds.all_data()
         fi = ("index", axis)
--- a/yt/geometry/tests/fake_octree.py
+++ b/yt/geometry/tests/fake_octree.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+
 from yt.geometry.fake_octree import create_fake_octree
 from yt.geometry.oct_container import RAMSESOctreeContainer, ParticleOctreeContainer
 import numpy as np
--- a/yt/geometry/tests/fake_octree.py
+++ b/yt/geometry/tests/fake_octree.py
@@ -1,4 +1,3 @@
-
 from yt.geometry.fake_octree import create_fake_octree
 from yt.geometry.oct_container import RAMSESOctreeContainer, ParticleOctreeContainer
 import numpy as np
--- a/yt/mods.py
+++ b/yt/mods.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+
 
 #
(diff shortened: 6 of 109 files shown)


yt/frontends/swift/tests/test_outputs.py
```diff
@@ -15,7 +15,7 @@
 @requires_file(keplerian_ring)
 def test_non_cosmo_dataset():
     ds = load(keplerian_ring)
-    assert(type(ds) == SwiftDataset)
+    assert(type(ds) is SwiftDataset)
```
cphyc, line 18:
```suggestion
    assert(isinstance(ds, SwiftDataset)
```

yt/geometry/tests/fake_octree.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import print_function
+
 from yt.geometry.fake_octree import create_fake_octree
```
cphyc, line 2:
```suggestion
from yt.geometry.fake_octree import create_fake_octree
```

yt/mods.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+
 
 #
```
cphyc, line 3:
```suggestion
#
```

yt/utilities/decompose.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import division
+
 
 import numpy as np
```
cphyc, line 3:
```suggestion
import numpy as np
```

yt/utilities/grid_data_format/__init__.py
```diff
@@ -1,3 +1,3 @@
-from __future__ import absolute_import
+
 from .conversion import *
```
cphyc, line 2:
```suggestion
from .conversion import *
```

yt/utilities/grid_data_format/conversion/__init__.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+
 from .conversion_abc import Converter
```
cphyc, line 2:
```suggestion
from .conversion_abc import Converter
```

yt/utilities/grid_data_format/conversion/conversion_athena.py
```diff
@@ -1,5 +1,5 @@
-from __future__ import print_function
-from __future__ import absolute_import
+
+
 import numpy as np
```
cphyc, line 3:
```suggestion
import numpy as np
```

yt/utilities/lib/cykdtree/tests/test_utils.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import division
+
 
 import numpy as np
```
cphyc, line 3:
```suggestion
import numpy as np
```

yt/utilities/sdf.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import print_function
+
 from io import StringIO
```
cphyc, line 2:
```suggestion
from io import StringIO
```

yt/visualization/volume_rendering/lens.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import division
+
 from yt.funcs import mylog
```
cphyc, line 2:
```suggestion
from yt.funcs import mylog
```

yt/data_objects/tests/test_octree.py
```diff
@@ -12,7 +12,7 @@ def test_building_tree():
     '''
     ds = fake_sph_grid_ds()
     octree = ds.octree(n_ref=1)
-    assert(type(octree) == YTOctree)
+    assert(type(octree) is YTOctree)
```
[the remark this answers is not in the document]
neutrinoceros: In the context of a test that I didn't write I assume it's best to check for exact typing. Using `isinstance()` here would make the test more flexible and I see no reason to allow this as long as it's not motivated by the data we actually test.

yt/frontends/swift/tests/test_outputs.py
neutrinoceros: same as above, not willing to make tests more flexible here.

neutrinoceros:
Oops. I used `[ci-skip]` instead of `[ci skip]` and now the CI pipeline will be running for ever 🤦

yt/geometry/coordinates/tests/test_cartesian_coordinates.py
```diff
@@ -13,8 +13,7 @@ def test_cartesian_coordinates():
     # We're going to load up a simple AMR grid and check its volume
     # calculations and path length calculations.
     ds = fake_amr_ds()
-    axes = list(set(ds.coordinates.axis_name.values()))
-    axes.sort()
+    axes = sorted(set(ds.coordinates.axis_name.values()))
```
munkm, line 16: 👍 to this change

yt/utilities/math_utils.py
```diff
@@ -819,31 +819,30 @@ def quartiles(a, axis=None, out=None, overwrite_input=False):
     """
     if overwrite_input:
         if axis is None:
-            sorted = a.ravel()
-            sorted.sort()
+            a_sorted = sorted(a.ravel())
         else:
             a.sort(axis=axis)
-            sorted = a
+            a_sorted = a
```
munkm, line 825: I like this clearer variable name

yt/geometry/coordinates/spherical_coordinates.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import print_function
+
```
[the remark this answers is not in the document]
neutrinoceros: good catch, fixing it right now

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

PR #2641, opened 2020-06-13
Cleanup legacy python checks

neutrinoceros:
There are many places in the code base where we check the python version. 99% of the time it's done to handle special cases where python 2 and python 3 have different behaviours. Since `yt/__init__.py` now raises an exception when python 2 is used, I think there is no question that those other checks should be removed.
However there are very rare (two) occurrences where we check for versions more recent than 3.0 (namely 3.3 and 3.4). Since we do not test against any version older than 3.6, and a vast majority of the python3 community is using 3.6 or later (3% are using 3.5 and another 3% is using  older versions) ([source](https://www.jetbrains.com/lp/python-developers-survey-2019/)), I assume it's not an issue to remove those checks too.
In particular, this removes the need for `yt.utilities.lru_cache` which is a backport of std lib meant for python 3.2 and older.

I think it'd be good to explicitly state somewhere what minor version of python 3.x we're supporting (is it 3.5 or 3.6 ?). It could also be checked in __init__.py for safety.


yt/frontends/tipsy/data_structures.py
```diff
@@ -20,9 +20,7 @@
 from .fields import \
     TipsyFieldInfo
 
-import sys
-if sys.version_info > (3,):
-    long = int
+long = int
```
cphyc, line 23: We should get rid of any occurrence of `long` in the following of the file.

neutrinoceros: This is done in #2640. I think these PRs should be merged together, but I forgot to mention it !

munkm: It's ok, you don't need to merge them together! Noting they're complimentary in the comments here or the PR works too (and smaller ones are easier to review anyways).

neutrinoceros:
> Since this PR unvendors lru_cache I'd delete references to it (like in setup.cfg, .coveragerc, and .pep8speaks ) too.
Thanks ! Indeed I didn't think it was mentioned anywhere explicitly.

yt/frontends/tipsy/data_structures.py
cphyc: Indeed, my bad I went through this PR before #2640!

PR #2623, opened 2020-06-09
Add mode kwarg to h5py.File calls

neutrinoceros:
## PR Summary

This is built on top of #2594 as per suggested there.

for reference this change was performed automatically with the following script:
```python
import re
from pathlib import Path

yt_dir = Path(__file__).parent.parent

exp = re.compile("""h5py\.File\([^,]+,\s*['"]""")

def repl_exp(matchobj):
    core = matchobj.group(0)[:-1].rstrip()
    sep = matchobj.group(0)[-1]
    return f"{core} mode={sep}"

for file in yt_dir.glob("**/*.py"):
    with open(file, mode="r") as fileobj:
        text = fileobj.read()

    if (new_text := exp.sub(repl_exp, text)) == text:
        continue

    with open(file, mode="w") as fileobj:
        fileobj.write(new_text)
```

PR #2643, opened 2020-06-15
Use 'a' mode for saving arrays to hdf. Fixes #2642

### LINKED ISSUE #2642: Saving multiple YTArrays to a single hdf clobbers the file
Writing more than one YTArray to a single hdf5 file clobbers the file during each write.

**Code for reproduction**
Taken straight from http://yt-project.org/docs/dev/analyzing/units/symbolic_units.html with minor modifications:

```python
import numpy as np
from yt.units.yt_array import YTArray

my_dens = YTArray(np.random.random(10), 'Msun/kpc**3')
my_temp = YTArray(np.random.random(10), 'K')
my_dens.write_hdf5("my_data.h5", dataset_name="density")
my_temp.write_hdf5("my_data.h5", dataset_name="temperature")
read_dens = YTArray.from_hdf5("my_data.h5", dataset_name="density")
print(read_dens)
print(my_dens)
```

**Actual outcome**
```
  File "/tmp/ala.py", line 8, in <module>
    read_dens = YTArray.from_hdf5("my_data.h5", dataset_name="density")
  File "/home/xarth/codes/xarthisius/yt/yt/units/yt_array.py", line 1019, in from_hdf5
    dataset = g[dataset_name]
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/home/xarth/.local/lib/python3.6/site-packages/h5py/_hl/group.py", line 264, in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5o.pyx", line 190, in h5py.h5o.open
KeyError: "Unable to open object (object 'density' doesn't exist)"
```

**Expected outcome**

```
[0.15325644 0.06909042 0.13625627 0.79578573 0.97937465 0.53984921
 0.63471635 0.35383393 0.23706361 0.42651924] Msun/kpc**3
[0.15325644 0.06909042 0.13625627 0.79578573 0.97937465 0.53984921
 0.63471635 0.35383393 0.23706361 0.42651924] Msun/kpc**3
```

**Version Information**
  * Operating System: Ubuntu 18.04.4 LTS
  * Python Version: 3.6.9
  * yt version: yt-3.5.1-1372-g1f4051157

Xarthisius:
## PR Summary

When saving an array to hdf5 file open the file read/write if it exists, create it otherwise.

#### Note
Without this PR documentation build is failing.

### DIFF
--- a/yt/units/yt_array.py
+++ b/yt/units/yt_array.py
@@ -962,7 +962,7 @@ class YTArray(np.ndarray):
         if dataset_name is None:
             dataset_name = 'array_data'
 
-        f = h5py.File(filename, "w")
+        f = h5py.File(filename, mode="a")
         if group_name is not None:
             if group_name in f:
                 g = f[group_name]


neutrinoceros:
note:  this collides with #2623, could you please use the keyword syntax to specify the mode ?

Also, evidently this change fixes the issue in the specific case you reported it, but I'm worried that it would create a bug in an other situation where the user specifically wants to overwrite the file.
One way around this would be to expose the "mode" argument in the `write_hdf5` function, and to settle for either `"w"` or `"a"` as a default. I personally think defaulting to `"w"` is cleaner but as long as it's documented in the docstring I don't mind if you prefer going the other way :)

Xarthisius:
> note: this collides with #2623, could you please use the keyword syntax to specify the mode ?

added.

> Also, evidently this change fixes the issue in the specific case you reported it, but I'm worried that it would create a bug in an other situation where the user specifically wants to overwrite the file.
> One way around this would be to expose the "mode" argument in the `write_hdf5` function, and to settle for either `"w"` or `"a"` as a default. I personally think defaulting to `"w"` is cleaner but as long as it's documented in the docstring I don't mind if you prefer going the other way :)

Clobbering a file via "w" mode appeared here due to https://github.com/yt-project/yt/pull/2594 . It changes the behavior of `.write_hdf5` without any notice or change to the docstring. I'm merely restoring the API that was accidentally altered. Additionally, it wasn't possible to change the mode before #2594 either.

I'm not opposing changing the default behavior, nor exposing `mode` as kwargs, but that's not the point of this PR.

PR #2644, opened 2020-06-16
Don't allow edges to extend beyond domain for SPH region sources.

matthewturk:
This showed up when generating the gallery using the `TipsyGalaxy` dataset.  What would happen is that if you set the width to `(1.0, 'unitary')` it would extend the *plot* beyond the edge of the domain, which would then create a *region* that went beyond the edge of the domain, and then break because the dataset wasn't periodic.  This prevents that from happening.

### DIFF
--- a/yt/geometry/coordinates/cartesian_coordinates.py
+++ b/yt/geometry/coordinates/cartesian_coordinates.py
@@ -286,10 +286,18 @@ class CartesianCoordinateHandler(CoordinateHandler):
                 le, re = data_source.data_source.get_bbox()
                 xa = self.x_axis[dim]
                 ya = self.y_axis[dim]
-                le[xa] = max(bounds[0], self.ds.domain_left_edge[xa])
-                le[ya] = max(bounds[2], self.ds.domain_left_edge[ya])
-                re[xa] = min(bounds[1], self.ds.domain_right_edge[xa])
-                re[ya] = min(bounds[3], self.ds.domain_right_edge[ya])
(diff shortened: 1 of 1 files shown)


matthewturk:
I had to update this (I am sorry!) because I realized that when the data *is* periodic, we want the region to go as it was, over the boundaries.

@cphyc @Xarthisius @munkm another quick look?

yt/geometry/coordinates/cartesian_coordinates.py
```diff
@@ -284,10 +284,21 @@ def _ortho_pixelize(self, data_source, field, bounds, size, antialias,
...
+                else:
+                    le[xa] = bounds[0]
+                    re[xa] = bounds[1]
+                if not self.ds.periodicity[ya]:
+                    le[ya] = max(bounds[2], self.ds.domain_left_edge[ya])
+                    re[ya] = min(bounds[3], self.ds.domain_right_edge[ya])
+                else:
+                    le[ya] = bounds[2]
+                    re[ya] = bounds[3]
+                # We actually need to clip these
```
munkm, line 301: super minor but it might be worth modifying the inline comment to say why the clip needs to happen here?

neutrinoceros:
Sidenote, I can't wait to test this against my own (amrvac) datasets. I've been using workarounds to avoid this issue, and never took the time to report or inspect it. Thanks a lot !

matthewturk:
Those images were generated with this change.

PR #2645, opened 2020-06-17
Support for Gadget HDF5-like datasets without smoothing lengths included

jzuhone:
# PR Summary

A number of frontends are derived from the `GadgetHDF5Dataset` class (and its associated field info and index classes). Most of the time, the `SmoothingLength` field is included in the dataset, but if not, it needs to be generated. This PR provides the code needed to generate the smoothing length in the event it does not exist. 

If a dataset does not have the `SmoothingLength` field, this PR generates a set of smoothing lengths for the SPH particles which are stored to auxiliary file(s), the number of such files depending on the number of files in the dataset itself. Then the smoothing lengths are read from these files.

I had to touch a lot more files than I would have liked in order to make sure all the parts worked together. This also has side-effects on Arepo datasets (which also do not have smoothing lengths but take care of them in other ways) which I had to fix.

## PR Checklist

This is not ready for merge yet. It needs:

- [x] Logic to prevent the `hsml` files from being overwritten if they already exist and are consistent with the dataset
- [ ] Tests, especially since the logic to go from `data_files` (based on chunking) to the `hsml` filenames is a bit odd and I'm not 100% sure I did it right.

### DIFF
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -7,24 +7,22 @@ metal_elements = ["He", "C", "N", "O", "Ne", "Mg", "Si", "Fe"]
 
 
 class ArepoFieldInfo(GadgetFieldInfo):
-    known_particle_fields = GadgetFieldInfo.known_particle_fields + (
-        ("smoothing_length", ("code_length", [], None)),
-        ("MagneticField", ("code_magnetic", ["particle_magnetic_field"], None)),
-        (
-            "MagneticFieldDivergence",
-            ("code_magnetic/code_length", ["magnetic_field_divergence"], None),
-        ),
-        ("GFM_Metallicity", ("", ["metallicity"], None)),
-        ("GFM_Metals_00", ("", ["H_fraction"], None)),
-        ("GFM_Metals_01", ("", ["He_fraction"], None)),
-        ("GFM_Metals_02", ("", ["C_fraction"], None)),
-        ("GFM_Metals_03", ("", ["N_fraction"], None)),
-        ("GFM_Metals_04", ("", ["O_fraction"], None)),
-        ("GFM_Metals_05", ("", ["Ne_fraction"], None)),
-        ("GFM_Metals_06", ("", ["Mg_fraction"], None)),
-        ("GFM_Metals_07", ("", ["Si_fraction"], None)),
-        ("GFM_Metals_08", ("", ["Fe_fraction"], None)),
-    )
+    known_particle_fields = GadgetFieldInfo.known_particle_fields + \
+                            (("MagneticField",
+                              ("code_magnetic", ["particle_magnetic_field"], None)),
+                             ("MagneticFieldDivergence",
+                              ("code_magnetic/code_length", ["magnetic_field_divergence"], None)),
+                             ("GFM_Metallicity", ("", ["metallicity"], None)),
+                             ("GFM_Metals_00", ("", ["H_fraction"], None)),
+                             ("GFM_Metals_01", ("", ["He_fraction"], None)),
+                             ("GFM_Metals_02", ("", ["C_fraction"], None)),
+                             ("GFM_Metals_03", ("", ["N_fraction"], None)),
+                             ("GFM_Metals_04", ("", ["O_fraction"], None)),
+                             ("GFM_Metals_05", ("", ["Ne_fraction"], None)),
+                             ("GFM_Metals_06", ("", ["Mg_fraction"], None)),
+                             ("GFM_Metals_07", ("", ["Si_fraction"], None)),
+                             ("GFM_Metals_08", ("", ["Fe_fraction"], None)),
+                             )
 
     def __init__(self, ds, field_list, slice_info=None):
         if ds.cosmological_simulation:
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -7,7 +7,6 @@ metal_elements = ["He", "C", "N", "O", "Ne", "Mg", "Si", "Fe"]
 
 
 class ArepoFieldInfo(GadgetFieldInfo):
-
     def __init__(self, ds, field_list, slice_info=None):
         if ds.cosmological_simulation:
             GFM_SFT_units = "dimensionless"
@@ -16,8 +15,10 @@ class ArepoFieldInfo(GadgetFieldInfo):
         self.known_particle_fields += (
             ("GFM_StellarFormationTime", (GFM_SFT_units, ["stellar_age"], None)),
             ("MagneticField", ("code_magnetic", ["particle_magnetic_field"], None)),
-            ("MagneticFieldDivergence",
-                ("code_magnetic/code_length", ["magnetic_field_divergence"], None)),
+            (
+                "MagneticFieldDivergence",
+                ("code_magnetic/code_length", ["magnetic_field_divergence"], None),
+            ),
             ("GFM_Metallicity", ("", ["metallicity"], None)),
             ("GFM_Metals_00", ("", ["H_fraction"], None)),
             ("GFM_Metals_01", ("", ["He_fraction"], None)),
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -15,10 +15,23 @@ class ArepoFieldInfo(GadgetFieldInfo):
             GFM_SFT_units = "code_length/code_velocity"
         self.known_particle_fields += (
             ("GFM_StellarFormationTime", (GFM_SFT_units, ["stellar_age"], None)),
+            ("MagneticField", ("code_magnetic", ["particle_magnetic_field"], None)),
+            ("MagneticFieldDivergence",
+                ("code_magnetic/code_length", ["magnetic_field_divergence"], None)),
+            ("GFM_Metallicity", ("", ["metallicity"], None)),
+            ("GFM_Metals_00", ("", ["H_fraction"], None)),
+            ("GFM_Metals_01", ("", ["He_fraction"], None)),
+            ("GFM_Metals_02", ("", ["C_fraction"], None)),
+            ("GFM_Metals_03", ("", ["N_fraction"], None)),
+            ("GFM_Metals_04", ("", ["O_fraction"], None)),
+            ("GFM_Metals_05", ("", ["Ne_fraction"], None)),
+            ("GFM_Metals_06", ("", ["Mg_fraction"], None)),
+            ("GFM_Metals_07", ("", ["Si_fraction"], None)),
+            ("GFM_Metals_08", ("", ["Fe_fraction"], None)),
         )
         super(ArepoFieldInfo, self).__init__(ds, field_list, slice_info=slice_info)
 
-    def setup_particle_fields(self, ptype):
+    def setup_particle_fields(self, ptype, *args, **kwargs):
         FieldInfoContainer.setup_particle_fields(self, ptype)
         if ptype == "PartType0":
             self.setup_gas_particle_fields(ptype)
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -7,22 +7,6 @@ metal_elements = ["He", "C", "N", "O", "Ne", "Mg", "Si", "Fe"]
 
 
 class ArepoFieldInfo(GadgetFieldInfo):
-    known_particle_fields = GadgetFieldInfo.known_particle_fields + \
-                            (("MagneticField",
-                              ("code_magnetic", ["particle_magnetic_field"], None)),
-                             ("MagneticFieldDivergence",
-                              ("code_magnetic/code_length", ["magnetic_field_divergence"], None)),
-                             ("GFM_Metallicity", ("", ["metallicity"], None)),
-                             ("GFM_Metals_00", ("", ["H_fraction"], None)),
-                             ("GFM_Metals_01", ("", ["He_fraction"], None)),
-                             ("GFM_Metals_02", ("", ["C_fraction"], None)),
-                             ("GFM_Metals_03", ("", ["N_fraction"], None)),
-                             ("GFM_Metals_04", ("", ["O_fraction"], None)),
-                             ("GFM_Metals_05", ("", ["Ne_fraction"], None)),
-                             ("GFM_Metals_06", ("", ["Mg_fraction"], None)),
-                             ("GFM_Metals_07", ("", ["Si_fraction"], None)),
-                             ("GFM_Metals_08", ("", ["Fe_fraction"], None)),
-                             )
 
     def __init__(self, ds, field_list, slice_info=None):
         if ds.cosmological_simulation:
--- a/yt/frontends/arepo/io.py
+++ b/yt/frontends/arepo/io.py
@@ -1,6 +1,8 @@
-from yt.frontends.gadget.api import IOHandlerGadgetHDF5
 import numpy as np
+
+from yt.frontends.gadget.api import IOHandlerGadgetHDF5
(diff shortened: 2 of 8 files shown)


yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
             yield key, pos
         f.close()
 
+    def _generate_smoothing_length(self, data_files, kdtree):
```
jzuhone, line 83: The logic in this routine is a bit squirrely and I would like @matthewturk and others to check it.

yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
...
+        positions = []
+        counts = defaultdict(int)
+        for data_file in data_files:
+            for _, ppos in self._yield_coordinates(
+                    data_file, needed_ptype=self.ds._sph_ptypes[0]):
+                counts[data_file.filename] += ppos.shape[0]
+                positions.append(ppos)
+        if len(positions) == 0:
+            return
+        offsets = {}
```
jzuhone, line 105: I probably should add some comments in here

yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
...
+            return
+        # TODO: We need to do something like this here but I'm not sure how
+        """
+        if os.path.exists(self.hsml_filename):
+            with open(self.hsml_filename, 'rb') as f:
+                file_hash = struct.unpack('q', f.read(struct.calcsize('q')))[0]
+            if file_hash != self.ds._file_hash:
+                os.remove(self.hsml_filename)
+            else:
+                return
```
neutrinoceros, line 94: better ask for forgiveness than permission in Python ([EAFP](https://docs.python.org/3/glossary.html?highlight=ask%20forgiveness))
```python
try:
     with open(self.hsml_filename, 'rb') as f:
         file_hash = struct.unpack('q', f.read(struct.calcsize('q')))[0]

     if file_hash == self.ds._file_hash:
         return

     os.remove(self.hsml_filename)
except FileNotFoundError:
    pass
```

yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
...
+    def _generate_smoothing_length(self, data_files, kdtree):
+        if not self.ds.gen_hsmls:
+            return
+        # TODO: We need to do something like this here but I'm not sure how
+        """
+        if os.path.exists(self.hsml_filename):
+            with open(self.hsml_filename, 'rb') as f:
+                file_hash = struct.unpack('q', f.read(struct.calcsize('q')))[0]
+            if file_hash != self.ds._file_hash:
+                os.remove(self.hsml_filename)
```
neutrinoceros, line 92: If you're going to remove a file conditionally, I believe it should be logged as a warning (mentioning why it's done too). Also I think the attribute should be dealt with (either set to None or deleted ?)

yt/frontends/gadget/data_structures.py
```diff
@@ -322,6 +322,7 @@ def __repr__(self):
         return os.path.basename(self.parameter_filename).split(".")[0]
 
     def _get_hvals(self):
+        self.gen_hsmls = False
```
neutrinoceros, line 325: maybe log this as debug message

yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
...
+                return
+        """
+        positions = []
+        counts = defaultdict(int)
+        for data_file in data_files:
+            for _, ppos in self._yield_coordinates(
+                    data_file, needed_ptype=self.ds._sph_ptypes[0]):
+                counts[data_file.filename] += ppos.shape[0]
+                positions.append(ppos)
+        if len(positions) == 0:
```
neutrinoceros, line 103:
```suggestion
        if not positions:
```

yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
...
+            return
+        offsets = {}
+        offset = 0
+        for fn in counts:
+            offsets[fn] = offset
+            offset += counts[fn]
```
neutrinoceros, line 109:
```suggestion
        for fn, count in counts.items():
            offsets[fn] = offset
            offset += count
```

yt/frontends/arepo/fields.py
```diff
@@ -35,7 +34,22 @@ def __init__(self, ds, field_list, slice_info=None):
...
+                               ("GFM_Metals_01", ("", ["He_fraction"], None)),
+                               ("GFM_Metals_02", ("", ["C_fraction"], None)),
+                               ("GFM_Metals_03", ("", ["N_fraction"], None)),
+                               ("GFM_Metals_04", ("", ["O_fraction"], None)),
+                               ("GFM_Metals_05", ("", ["Ne_fraction"], None)),
+                               ("GFM_Metals_06", ("", ["Mg_fraction"], None)),
+                               ("GFM_Metals_07", ("", ["Si_fraction"], None)),
+                               ("GFM_Metals_08", ("", ["Fe_fraction"], None)),
+                               )
+        self.known_particle_fields += new_particle_fields
```
neutrinoceros, line 52: Why not adding these known fields directly in the initialisation line for `know_particle_fields` ?

yt/frontends/arepo/io.py
```diff
@@ -5,6 +5,10 @@
 class IOHandlerArepoHDF5(IOHandlerGadgetHDF5):
     _dataset_type = "arepo_hdf5"
 
+    def _generate_smoothing_length(self, data_files, kdtree):
```
neutrinoceros, line 8: if this empty function necessary ? I can't find where it's called

yt/frontends/gadget/io.py
```diff
@@ -223,6 +271,9 @@ def _identify_fields(self, data_file):
                         self._vector_fields[kk] = g[kk].shape[1]
                     fields.append((ptype, str(kk)))
 
+        if self.ds.gen_hsmls:
```
neutrinoceros, line 274: This should be _after_ the `f.close()` line (better closing a file as soon as we're done with io operations)

yt/frontends/arepo/fields.py
```diff
@@ -13,8 +13,7 @@
 
 class ArepoFieldInfo(GadgetFieldInfo):
     known_particle_fields = GadgetFieldInfo.known_particle_fields + \
-                            (("smoothing_length", ("code_length", [], None)),
```
neutrinoceros, line 16: Is this backward compatibility breaking change ?

yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
             yield key, pos
         f.close()
 
+    def _generate_smoothing_length(self, data_files, kdtree):
+        if not self.ds.gen_hsmls:
+            return
+        # TODO: We need to do something like this here but I'm not sure how
```
neutrinoceros, line 86: Could you elaborate what you are uncertain about here ?

yt/frontends/gadget/data_structures.py
jzuhone: I'm not sure why we would want to do that. This is just supposed to be the default value of this attribute in the case that we don't want to generate smoothing lengths (which we would not want to do normally).

yt/frontends/arepo/fields.py
jzuhone: This is a mistake, I'll fix it.

jzuhone: Actually, no, it's not a mistake, because it's taken care of by the addition of this field to the `GadgetFieldInfo` class if it needs to exist.

yt/frontends/gadget/io.py
jzuhone: This was copied over directly from the Tipsy frontend. It is supposed to prevent the smoothing length files from being regenerated if they are already there and are consistent with the dataset. But in this case 1) the files are HDF5 and not unformatted like Tipsy and 2) in this case there are also going to be multiple HDF5 files for the smoothing length, being the same number as the number of files in the dataset. So we have to adjust the logic here but I am not sure how.

yt/frontends/gadget/io.py
jzuhone: Agreed.

yt/frontends/arepo/fields.py
jzuhone: One of these fields has information that is determined at load time, so rather than messing with `known_particle_fields` twice in here I've decided to do it only once in `__init__`.

yt/frontends/gadget/io.py
jzuhone: This is better, yes. Will include it.

yt/frontends/gadget/io.py
jzuhone: Agreed.

yt/frontends/gadget/io.py
jzuhone: See my comment below, all of this needs to be changed. This is the way it is handled in the Tipsy frontend.

yt/frontends/gadget/io.py
jzuhone: See comment below, this needs to be changed entirely.

yt/frontends/arepo/io.py
jzuhone: This is called from `SPHParticleIndex._initialize_index`. If we can't find smoothing lengths, we need to generate them. However, in the Arepo frontend we want to generate them a different way than in the case being handled in this PR. However, there is probably a better way to do this so I will see what I can do.

neutrinoceros:
btw this looks like a fix to the currently failing Eagle test (`yt/frontends/eagle/tests/test_outputs.py::test_Snipshot`)
error message: `KeyError: "Unable to open object (object 'SmoothingLength' doesn't exist)"`

(see for instance
https://tests.yt-project.org/job/yt_py38_git/346/testReport/junit/yt.frontends.eagle.tests/test_outputs/test_Snipshot/
)

matthewturk:
@jzuhone Do you want another set of hands on this?

jzuhone:
@matthewturk Kid is home this week because of school vacation, so maybe. We need to figure out how to check the hash of already-existing `hsml` files that this creates and decide whether or not to rewrite them. There is commented code from the Tipsy frontend to do this, so that's a start, but I am not sure where to go from there.

yt/frontends/arepo/io.py
matthewturk: Just so I am sure I understand -- does this mean that when setting up the index, we don't have the smoothing lengths?  So we won't set the ghost zones needed in the index precisely right?

yt/frontends/eagle/fields.py
```diff
@@ -26,10 +26,9 @@ class EagleNetworkFieldInfo(OWLSFieldInfo):
          'Fe17', 'Fe18', 'Fe19', 'Fe20', 'Fe21', 'Fe22', 'Fe23',\
          'Fe24', 'Fe25', 'Fe25', 'Fe27',)
 
-    def __init__(self, *args, **kwargs):
-        
-        super(EagleNetworkFieldInfo,self).__init__( *args, **kwargs )
-        
+    def __init__(self, ds, field_list, slice_info=None):
```
matthewturk, line 29: We probably don't need this if it's empty, I think.  But not necessary to remove it.

yt/frontends/gadget/io.py
matthewturk: So I *think* that you might be able to replace the logic with:

1. Allowing an HSML file name to be generated on a per-file basis
2. Checking if that filename exists when thinking about reading it
3. Replacing the `file_hash` with an h5py file open.

yt/frontends/gadget/io.py
matthewturk: no sense breaking with tradition imo

yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
...
+                counts[data_file.filename] += ppos.shape[0]
+                positions.append(ppos)
+        if not positions:
+            return
+        offsets = {}
+        offset = 0
+        for fn, count in counts.items():
+            offsets[fn] = offset
+            offset += count
+        positions = np.concatenate(positions)[kdtree.idx]
```
matthewturk, line 110: Is there *any* chance these will be unitful?  If so we should use `uconcatenate`

yt/frontends/gadget/io.py
```diff
@@ -77,13 +80,58 @@ def _yield_coordinates(self, data_file, needed_ptype=None):
...
+        hsml = generate_smoothing_length(
+            positions, kdtree, self.ds._num_neighbors)
+        dtype = positions.dtype
+        hsml = hsml[np.argsort(kdtree.idx)].astype(dtype)
+        for data_file in data_files:
+            si, ei = data_file.start, data_file.end
+            fn = data_file.filename
+            hsml_fn = data_file.filename+".hsml"
```
matthewturk, line 118: if they're hdf5 we should probably append h5 or something too

matthewturk:
@jzuhone What do you think about having this PR include a re-enabling of the snipshot tests, and turning them off in master?  That would help us reduce the "we know it's failing" red checks on other PRs.

jzuhone:
@matthewturk this is fine for me.

yt/frontends/gadget/io.py
jzuhone: What are we comparing the h5py file open to? Do we need to write something in the hsml file to check? There's some logic in the Tipsy frontend I don't fully understand having to do with packing a struct with "q"

matthewturk: So `q` is a `long long` and I would guess that tipsy is storing a hash.  It generates the hash in its own frontend, then stores it, and if it's different, it regenerates.

I think we don't need to do the file hash thing here, unless we already have a file hash, and if we do, then go ahead and make it an `attr` on the root node of the hdf5 file.

jzuhone: I think we'd have to do that if we don't want to regenerate the file, tbh. So we could use the UUID for the dataset or something

jzuhone: Or, oops, the `_file_hash`

jzuhone:
Argh, I screwed up the rebase onto master somehow. Will fix.

yt/frontends/arepo/io.py
jzuhone: @matthewturk I thought about this and it doesn't mean that. it just means that in the Arepo case we don't have to generate the lengths since we're just computing them from the volume.

neutrinoceros:
Hey @jzuhone , sorry the linting failed on travis. I don't know how much you've been following the recent evolutions on this side but you should be able to solve this by running this on your branch

```
black yt
isort . # ! need isort~5.1
```

let me know if I can help with this !

neutrinoceros:
Note that the following line should be removed in this PR
https://github.com/yt-project/yt/blob/53b2d0aa2842bedede76d7ced2a1450f3402206d/tests/tests.yaml#L165

jzuhone:
Do we know what's going on with the one Jenkins failure?

neutrinoceros:
@jzuhone Yes, @Xarthisius fixed it here https://github.com/yt-project/yt/pull/2779,
Once it passes on Appveyor, we can merge it !

munkm:
@yt-fido test this please

PR #2650, opened 2020-06-19
fix an unintuitive behaviour in ProfilePlot

neutrinoceros:
## PR Summary

Because of the way it's currently implemented, `ProfilePlot` produces unintuitive results when passed `x_log=None`, which is effectively, equivalent has passing `x_log=True` despite the fact that `bool(None)==False`.

with this change, the following produce the same result
```python
import yt
ds = yt.testing.fake_random_ds(128)
yt.ProfilePlot(ds, 'y', ['density', "velocity_magnitude"], n_bins=500, x_log=None)
yt.ProfilePlot(ds, 'y', ['density', "velocity_magnitude"], n_bins=500, x_log=0)
yt.ProfilePlot(ds, 'y', ['density', "velocity_magnitude"], n_bins=500, x_log=False)
```

This issue was reported on slack by Jack Jenkins.
I'm opening this as a draft since there may be other appropriate changes to ProfilePlot.__init__() that could go with it.

### DIFF
--- a/yt/visualization/profile_plotter.py
+++ b/yt/visualization/profile_plotter.py
@@ -181,9 +181,9 @@ class ProfilePlot(object):
         arguments.  For example, dict(color="red", linestyle=":").
         Default: None.
     x_log : bool
-        If not None, whether the x_axis should be plotted with a logarithmic
-        scaling.
-        Default: None
+        Whether the x_axis should be plotted with a logarithmic
+        scaling (True), or linear scaling (False).
+        Default: True.
     y_log : dict
         A dictionary containing field:boolean pairs, setting the logarithmic
         property for that field. May be overridden after instantiation using 
@@ -237,14 +237,11 @@ class ProfilePlot(object):
                  weight_field="cell_mass", n_bins=64,
                  accumulation=False, fractional=False,
                  label=None, plot_spec=None,
-                 x_log=None, y_log=None):
+                 x_log=True, y_log=None):
 
         data_source = data_object_or_all_data(data_source)
 
-        if x_log is None:
-            logs = None
-        else:
-            logs = {x_field:x_log}
+        logs = {x_field: bool(x_log)}
 
         if isinstance(data_source.ds, YTProfileDataset):
             profiles = [data_source.ds.profile]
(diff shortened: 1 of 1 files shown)


yt/visualization/profile_plotter.py
```diff
@@ -704,14 +701,8 @@ def _set_font_properties(self):
...
         yfi = profile.field_info[field_y]
-        if self.x_log is None:
-            x_log = profile.x_log
-        else:
-            x_log = self.x_log
-        if field_y in self.y_log:
-            y_log = self.y_log[field_y]
-        else:
-            y_log = yfi.take_log
+        x_log = self.x_log if self.x_log is not None else profile.x_log
```
munkm, line 704: I think this isn't exactly readable. Would you mind splitting the logic for x_log across a couple of lines here?

yt/visualization/profile_plotter.py
```diff
@@ -416,11 +417,7 @@ def _initialize_instance(cls, obj, profiles, labels, plot_specs, y_log):
         obj._font_color = None
         obj.profiles = ensure_list(profiles)
         obj.x_log = None
-        obj.y_log = {}
-        if y_log is not None:
-            for field, log in y_log.items():
-                field, = obj.profiles[0].data_source._determine_fields([field])
-                obj.y_log[field] = log
+        obj.y_log = sanitize_field_tuple_keys(y_log, obj.profiles[0].data_source) or {}
```
munkm, line 420: since you moved the logic for fields = y_fields above this block, maybe adding this logic for the y_log immediately after would be good too. it would make this profile creation cleaner.

yt/visualization/profile_plotter.py
neutrinoceros: you got it

yt/visualization/profile_plotter.py
neutrinoceros: I am not sure what you mean here. Aren't those very distant blocks from 2 different method def ? (even if the name `_initialize_instance` puzzles me...)

munkm: Ha, yes sorry. Ignore this comment. I clearly hadn't had enough ☕ yet.

Issue #2646, opened 2020-06-19
Newest Version of Numpy does not work [have to degrade to 1.16.0]


claytonstrawn:
### Bug report

**Bug summary**

If you upgrade to the most recent numpy version, yt.load does not function correctly

**Code for reproduction**

pip install --upgrade numpy
python
>>> import yt
>>> yt.load('/Users/claytonstrawn/VELA_v2_27/10MpcBox_csf512_a0.500.d')

**Actual outcome**

```yt : [INFO     ] 2020-06-18 16:56:12,637 discovered particle_header:/Users/claytonstrawn/VELA_v2_27/PMcrda0.500.DAT
yt : [INFO     ] 2020-06-18 16:56:12,637 discovered particle_data:/Users/claytonstrawn/VELA_v2_27/PMcrs0a0.500.DAT
yt : [INFO     ] 2020-06-18 16:56:12,638 discovered particle_stars:/Users/claytonstrawn/VELA_v2_27/stars_a0.500.dat
yt : [INFO     ] 2020-06-18 16:56:12,665 Using root level of 13
yt : [INFO     ] 2020-06-18 16:56:12,667 Discovered 6 species of particles
yt : [INFO     ] 2020-06-18 16:56:12,667 Particle populations:   5287936    634880    106752     17168   2090807   1939360 
Traceback (most recent call last):
  File "/Users/claytonstrawn/anaconda2/envs/myenv_testforinterns/lib/python3.6/site-packages/numpy/core/function_base.py", line 117, in linspace
    num = operator.index(num)
TypeError: 'float' object cannot be interpreted as an integer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/claytonstrawn/yt_trident_temp/yt/yt/convenience.py", line 70, in load
    return candidates[0](*args, **kwargs)
  File "/Users/claytonstrawn/yt_trident_temp/yt/yt/frontends/art/data_structures.py", line 175, in __init__
    unit_system=unit_system)
  File "/Users/claytonstrawn/yt_trident_temp/yt/yt/data_objects/static_output.py", line 307, in __init__
    self._parse_parameter_file()
  File "/Users/claytonstrawn/yt_trident_temp/yt/yt/frontends/art/data_structures.py", line 356, in _parse_parameter_file
    self.current_time = self.quan(b2t(self.parameters['t']), 'Gyr')
[...]

PR #2654, opened 2020-06-19
[bugfix] update art frontend to be compatible with numpy 1.18

munkm:
## PR Summary

This is a bugfix to resolve yt-project/yt#2646 . 

A similar fix was implemented in yt-project/yt#2448. Here I've made the last argument in logspace explicitly labeled with the `num` kwarg, so it's consistent with updated numpy docs. I also call `int` on the variable so we don't get the error reported in the issue. 

I also updated an inline comment to be consistent with this change in the code. 

Note: this is a quick bugfix for the art frontend. I'll probably submit a later PR to address similar ambiguities in the codebase. 

@claytonstrawn I can load art frontend files with this fix applied. 

## PR Checklist

### DIFF
--- a/yt/frontends/art/io.py
+++ b/yt/frontends/art/io.py
@@ -582,7 +582,7 @@ def find_root(f, a, b, tol=1e-6):
 
 
 def quad(fintegrand, xmin, xmax, n=1e4):
-    spacings = np.logspace(np.log10(xmin), np.log10(xmax), n)
+    spacings = np.logspace(np.log10(xmin), np.log10(xmax), num=int(n))
     integrand_arr = fintegrand(spacings)
     val = np.trapz(integrand_arr, dx=np.diff(spacings))
     return val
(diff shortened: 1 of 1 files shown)


munkm:
Maybe I'm missing something, but I really don't understand why we don't see these test failures in our CI. Travis runs the art frontend tests and the versions of numpy used on the non-minimal versions are 1.18, and there aren't test failures coming from the error that we see in 2646. Maybe I'm missing something simple?

neutrinoceros:
> Maybe I'm missing something, but I really don't understand why we don't see test failures related to the original issue in our CI.

So I had a look at [numpy 1.18.1 release notes](https://numpy.org/doc/1.18/release/1.18.1-notes.html) and it's mentioned that

>  Downstream developers should use Cython >= 0.29.14 for Python 3.8 support

While on travis we're currently testing against Python 3.8 with numpy 1.18.1 and Cython 0.29.6
Not 100% confident this is the reason why we're not catching those errors but might be worth bumping the Cython version to make sure.

PR #2661, opened 2020-06-22
Offset particle selection test for unitary coords

matthewturk:
This offsets the particle selection centers, as in unitary coordinates
that cross over the origin the left edge is not at 0, 'unitary'.  This
makes checks for periodicity in these cases (i.e., PKDGrav) more useful.

### DIFF
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -1358,7 +1358,7 @@ class ParticleSelectionComparison:
                    [0.96, 0.96, 0.96]]
         r = self.ds.quan(0.1, "unitary")
         for center in centers:
-            c = self.ds.arr(center, "unitary")
+            c = self.ds.arr(center, "unitary") + self.ds.domain_left_edge.in_units("unitary")
             if not all(self.ds.periodicity):
                 # filter out the periodic bits for non-periodic datasets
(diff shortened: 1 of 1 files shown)


matthewturk:
A slight modification to this should fix the FLASH dataset tests in #2677, as this is the error thrown:

```
RuntimeError: Error: yt attempted to read outside the boundaries of a non-periodic domain along dimension 0.
Region left edge = 4.4000000000000006e+24 code_length, Region right edge = 3.96e+25 code_length
Dataset left edge = -2.2e+25 code_length, Dataset right edge = 2.2e+25 code_length

This commonly happens when trying to compute ghost cells up to the domain boundary. Two possible solutions are to load a smaller region that does not border the edge or override the periodicity for this dataset.
```

This comes from a region selector, though, so I need to do a bit of work to work that out.

matthewturk:
OK, current one should fix the FLASH tests, I think.

cphyc:
@yt-fido test this please.

For some reason, the Jenkins test was giving a 404, so I am rerunning the tests.

cphyc:
@yt-fido test this please

Issue #2666, opened 2020-06-23
Flake-8 bugbear errors (YTEP00037)


neutrinoceros:
### Bug report

**Bug summary**

I proposed adding the flake8 plugging bugbear to our CI-linting stage in YTEP0037. Here's a report of edgy behaviours these additional linting rules catch:

**Code for reproduction**

```bash
pip install flake8-bugbear
flake8 yt/ --select B
```

**Actual outcome**

```
# excerpt 
yt/testing.py:38:15: B009 Do not call getattr with a constant attribute value, it is not any safer than normal property access.
yt/testing.py:39:21: B009 Do not call getattr with a constant attribute value, it is not any safer than normal property access.
yt/testing.py:519:32: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
yt/testing.py:531:9: B007 Loop control variable 'subcell' not used within the loop body. If this is intended, start the name with an underscore.
yt/testing.py:544:25: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
yt/pmods.py:306:17: B007 Loop control variable 'l' not used within the loop body. If this is intended, start the name with an underscore.
yt/visualization/eps_writer.py:350:21: B302 Python 3 does not include `.view*` methods on dictionaries. Remove the `view` prefix from the method name. For Python 2 compatibility, prefer the Python 3 equivalent unless you expect the size of the container to be large or unbounded. Then use `six.view*` or `future.utils.view*`.
yt/visualization/eps_writer.py:353:32: B302 Python 3 does not include `.view*` methods on dictionaries. Remove the `view` prefix from the method name. For Python 2 compatibility, prefer the Python 3 equivalent unless you expect the size of the container to be large or unbounded. Then use `six.view*` or `future.utils.view*`.
yt/visualization/eps_writer.py:555:43: B006 Do not use mutable data structures for argument defaults.  They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them.
yt/visualization/eps_writer.py:906:29: B006 Do not use mutable data structures for argument defaults.  They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them.
[...]

PR #2668, opened 2020-06-23
fix flake8-bugbear B009 (YTEP0037 6/6, part 2)

neutrinoceros:
## PR Summary

This addresses part of #2666, namely B009 "Do not call getattr with a constant attribute value".

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -532,9 +532,7 @@ class Dataset(metaclass=RegisteredDataset):
                 continue
             v = getattr(self, a)
             mylog.info("Parameters: %-25s = %s", a, v)
-        if hasattr(self, "cosmological_simulation") and getattr(
-            self, "cosmological_simulation"
-        ):
+        if hasattr(self, "cosmological_simulation") and self.cosmological_simulation:
             for a in [
                 "current_redshift",
                 "omega_lambda",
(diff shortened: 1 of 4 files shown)


yt/frontends/ytdata/data_structures.py
```diff
@@ -752,9 +752,7 @@ def print_key_parameters(self):
             v = getattr(self, a)
             if v is not None:
                 mylog.info("Parameters: %-25s = %s", a, v)
-        if hasattr(self, "cosmological_simulation") and getattr(
-            self, "cosmological_simulation"
-        ):
+        if hasattr(self, "cosmological_simulation") and self.cosmological_simulation:
```
matthewturk, line 755:
```suggestion
        if getattr(self, "cosmological_simulation", False):
```

yt/data_objects/static_output.py
```diff
@@ -532,9 +532,7 @@ def print_key_parameters(self):
                 continue
             v = getattr(self, a)
             mylog.info("Parameters: %-25s = %s", a, v)
-        if hasattr(self, "cosmological_simulation") and getattr(
-            self, "cosmological_simulation"
-        ):
+        if hasattr(self, "cosmological_simulation") and self.cosmological_simulation:
```
matthewturk, line 535:
```suggestion
        if getattr(self, "cosmological_simulation", False):
```

matthewturk:
Oh, I see.  Everything I suggested was what you were correcting!  I am pretty ambivalent about this.  But, whatever.  :)

PR #2669, opened 2020-06-23
ignore flake8-bugbear B008 errors (YTEP0037 6/6 part 3)

neutrinoceros:
## PR Summary

This addresses part of #2666, namely B008 "Do not perform function calls in argument defaults". The solution here is simply to ignore this error locally since it's only triggered in lines where calling a function in default attrs is the intended behaviour (in test functions).

### DIFF
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -574,7 +574,7 @@ def fake_sph_grid_ds(hsml_factor=1.0):
     return load_particles(data=data, length_unit=1.0, bbox=bbox)
 
 
-def construct_octree_mask(prng=RandomState(0x1D3D3D3), refined=None):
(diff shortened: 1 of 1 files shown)


yt/testing.py
```diff
@@ -600,7 +600,7 @@ def construct_octree_mask(prng=RandomState(0x1D3D3D3), refined=None):
 
 
 def fake_octree_ds(
-    prng=RandomState(0x1D3D3D3),
+    prng=RandomState(0x1D3D3D3),  # noqa B008
```
matthewturk, line 603:
```suggestion
    prng=RandomState(0x4D3D3D3),  # noqa B008
```

neutrinoceros: oh my, I have no clue what this changes does

matthewturk: probably best to just hit the accept button on the suggestion then :P

PR #2670, opened 2020-06-23
fix flake8-bugbear B014 errors (redundancy in exception catching) (YTEP0037 6/6, part 4)

neutrinoceros:
## PR Summary

This addresses part of #2666, namely B014 "Redundant exception types"

### DIFF
--- a/yt/frontends/flash/data_structures.py
+++ b/yt/frontends/flash/data_structures.py
@@ -459,7 +459,7 @@ class FLASHDataset(Dataset):
             fileh = HDF5FileHandler(args[0])
             if "bounding box" in fileh["/"].keys():
                 return True
-        except (IOError, OSError, ImportError):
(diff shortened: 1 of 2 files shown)


yt/frontends/flash/data_structures.py
```diff
@@ -459,7 +459,7 @@ def _is_valid(self, *args, **kwargs):
             fileh = HDF5FileHandler(args[0])
             if "bounding box" in fileh["/"].keys():
                 return True
-        except (IOError, OSError, ImportError):
+        except (OSError, ImportError):
```
matthewturk, line 462: I genuinely did not know this was redundant.

neutrinoceros: It wasn't always ! Only since python 3.3

PR #2671, opened 2020-06-23
fix flake8-bugbear B001 (bare excepts) (YTEP0037 6/6 part 5)

neutrinoceros:
## PR Summary

This partly addresses #2666, namely B001 "bare excepts".
This is also a follow up to #2474, and it shows that these kind of code degradation will continue to pop up until we have a linting tool to catch them (flake8-bugbear does the job here !)

### DIFF
--- a/yt/frontends/arepo/data_structures.py
+++ b/yt/frontends/arepo/data_structures.py
@@ -56,9 +56,8 @@ class ArepoHDF5Dataset(GadgetHDF5Dataset):
                 )
             )
             fh.close()
-        except:
+        except Exception:
             valid = False
-            pass
         return valid
 
     def _get_uvals(self):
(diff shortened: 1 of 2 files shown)


yt/frontends/arepo/data_structures.py
```diff
@@ -56,9 +56,8 @@ def _is_valid(self, *args, **kwargs):
                 )
             )
             fh.close()
-        except:
+        except Exception:
```
matthewturk, line 59: Oof, I'm glad we're adding `Exception` but we should probably improve this.  (I know you made this easier to improve in your load refactor work.)

neutrinoceros: Yes, it's also one of the goals on this PR: https://github.com/yt-project/yt/pull/2760 !

PR #2672, opened 2020-06-23
fix flake8-bugbear B010 errors (YTEP0037 6/6 part 6)

neutrinoceros:
## PR Summary
This partly addresses #2666, namely B010 "Do not call setattr with a constant attribute value"

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -116,11 +116,7 @@ class MutableAttribute:
             pass
         if self.display_array:
             try:
-                setattr(
-                    ret,
-                    "_ipython_display_",
-                    functools.partial(_wrap_display_ytarray, ret),
-                )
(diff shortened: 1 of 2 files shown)


yt/data_objects/static_output.py
```diff
@@ -116,11 +116,7 @@ def __get__(self, instance, owner):
             pass
         if self.display_array:
             try:
-                setattr(
-                    ret,
-                    "_ipython_display_",
-                    functools.partial(_wrap_display_ytarray, ret),
-                )
+                ret._ipython_display_ = functools.partial(_wrap_display_ytarray, ret)
```
matthewturk, line 119: I think I wrote this code, and I think I had a reason for using `setattr` but for the life of me I can't remember, so 🤷

neutrinoceros: I'll take the blame it this ever breaks :)

PR #2677, opened 2020-06-23
[BUGFIX] Call function wrapped with requires_file

cphyc:
WIthout this, *any* test wrapped with `requires_file` isn't run and is marked as ok by the testing suite.

A quick grep through the code indicates that this affects 200 tests :(.

### DIFF
--- a/yt/utilities/tests/test_minimal_representation.py
+++ b/yt/utilities/tests/test_minimal_representation.py
@@ -1,6 +1,5 @@
 import numpy as np
 import os.path
-from unyt.exceptions import UnitOperationError
 import yt
 from yt.config import ytcfg
 from yt.testing import \
--- a/yt/utilities/tests/test_minimal_representation.py
+++ b/yt/utilities/tests/test_minimal_representation.py
@@ -1,5 +1,6 @@
 import numpy as np
 import os.path
+from unyt.exceptions import UnitOperationError
 import yt
 from yt.config import ytcfg
 from yt.testing import \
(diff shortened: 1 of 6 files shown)


cphyc:
There are errors related to the tests:
- [x] AdaptaHOP: 3/3 fixed in #2678 
- [ ] Art (1)
- [ ] Athena (4)
- [ ] Eagle (1)
- [ ] Enzo (2)
- [ ] Flash (2)
- [ ] Gadget_fof (1)
- [ ] Gizmo (2)
- [ ] Open_pmd (4)
- [x] RAMSES: 6/6 fixed in #2679
- [ ] rockstar (1)
- [ ] ytdata (2)
- [ ] yt.geometry.tests (4)
- [ ] yt.utilities.tests (1)

cphyc:
@matt @ngoldbaum the test yt.utilities.tests.test_minimal_representation.test_store is failing because it is assumed that the `==` operation between two arrays that have different units raises an Exception. However, since we moved to unyt, this does not fail but instead return an array of False (is it the expected behavior?).
Anyway, I guess the fix, if any, should happen on unyt-side.

neutrinoceros:
for reference here's the page with all the failing tests this revealed
https://tests.yt-project.org/job/yt_py38_git/316/#showFailuresLink

ngoldbaum:
Why should a change happen on the unyt side for this? Can’t the test just raise an exception if the units are different?

ngoldbaum:
Just to clarify in case it’s not clear: this was an intentional behavior change with respect to yt.units, see https://github.com/yt-project/unyt/pull/109

cphyc:
> Why should a change happen on the unyt side for this? Can’t the test just raise an exception if the units are different?

Thanks for the answer. Maybe my phrasing was not very precise, sorry! What I meant is that either this is an intentional behavior (it is), in which case we should alter the test to reflect this. If it were not intentional, then it should have been fixed on unyt side.

yt/utilities/tests/test_minimal_representation.py
```diff
@@ -37,11 +35,13 @@ def test_store():
 
     def fail_for_different_method():
         proj2_c = ds.proj(field, "z", data_source=sp, method="mip")
-        return np.equal(proj2[field], proj2_c[field]).all()
-    assert_raises(YTUnitOperationError, fail_for_different_method)
+        assert_equal(proj2[field], proj2_c[field])
+    # A note here: a unyt.exceptions.UnitOperationError is raised
```
munkm, line 39: Thanks for adding the inline comment here. I think that's helpful.

PR #2679, opened 2020-06-24
[BUGFIX] Prevent RAMSES fail with recent versions of `f90nml`

cphyc:
Recent versions of f90nml (>=1.2.0), which is used to read the parameter file RAMSES uses, fail with a `StopIteration` instead of `ValueError`. This PR adapts to capture this new exception.

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -573,7 +573,7 @@ class RAMSESDataset(Dataset):
                     nml = f90nml.read(f)
             except ImportError as e:
                 nml = "An error occurred when reading the namelist: %s" % str(e)
-            except ValueError as e:
+            except (ValueError, StopIteration) as e:
                 mylog.warn("Could not parse `namelist.txt` file as it was malformed: %s" % str(e))
                 return
 


cphyc:
I filled an issue on the github repo of f90ml, so let me mention it in case _another_ exception may raised once this is fixed (if there is such thing as a fix): https://github.com/marshallward/f90nml/issues/118.

yt/frontends/ramses/data_structures.py
```diff
@@ -573,7 +573,7 @@ def read_namelist(self):
                     nml = f90nml.read(f)
             except ImportError as e:
                 nml = "An error occurred when reading the namelist: %s" % str(e)
-            except ValueError as e:
+            except (ValueError, StopIteration) as e:
```
munkm, line 576: StopIteration will only happen if the end of the file is malformed, right? Or is there another case where that will happen? Would it be helpful to split out these excepts and to explicitly mention this in the logging message for StopIteration?

cphyc: This piece of code is a bit unfortunate. Essentially, f90nml crashes when reading the file. It used to raise a ValueError because it was not parsing one of the token correctly (for version `<1.2`). With recent changes (version `>=1.2`), it is crashing differently, this time with a StopIteration (i.e. Python decides it has had enough of f90nml trying to parse the file).

The issue is that f90nml doesn't know the file is malformed, so there's no fixed API as how it should crash. I think we're kind of stuck with exhausting all the ways it can crash for now.

The only other option would be to catch *all* exceptions, which would be future-proof but is quite a bad practice.

cphyc: FYI, there is a discussion on https://github.com/marshallward/f90nml/issues/118 regarding what exception is raised by f90nml when it fails to parse a namelist.

munkm: Ahhh I see! I read that discussion but your explanation here really helped things click.

PR #2681, opened 2020-06-24
Allow throttling of number of cores for build.

brittonsmith:
This allows the user to set a maximum number of cores to use for the build. This is helpful on shared systems where jobs using all available cores get terminated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,12 @@ import pkg_resources
 def _get_cpu_count():
     if platform.system() != "Windows":
         cpu_count = os.cpu_count()
-        max_cores = min(cpu_count, int(os.getenv('MAX_BUILD_CORES', cpu_count)))
+        try:
+            user_max_cores = int(os.getenv('MAX_BUILD_CORES'), cpu_count)
+        except ValueError as e:
+            # print useful error message
+            raise e
+        max_cores = min(cpu_count, user_max_cores)
         return max_cores
     return 0
 
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ def _get_cpu_count():
         try:
             user_max_cores = int(os.getenv('MAX_BUILD_CORES', cpu_count))
         except ValueError as e:
-            # print useful error message
+            print ("MAX_BUILD_CORES needs to be set to an integer.")
             raise e
         max_cores = min(cpu_count, user_max_cores)
         return max_cores
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ def _get_cpu_count():
     if platform.system() != "Windows":
         cpu_count = os.cpu_count()
         try:
-            user_max_cores = int(os.getenv('MAX_BUILD_CORES'), cpu_count)
+            user_max_cores = int(os.getenv('MAX_BUILD_CORES', cpu_count))
         except ValueError as e:
             # print useful error message
             raise e
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,9 @@ import pkg_resources
 
 def _get_cpu_count():
     if platform.system() != "Windows":
-        return os.cpu_count()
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -20,7 +20,9 @@
 
 def _get_cpu_count():
     if platform.system() != "Windows":
-        return os.cpu_count()
+        cpu_count = os.cpu_count()
+        max_cores = min(cpu_count, int(os.getenv('MAX_BUILD_CORES', cpu_count)))
```
cphyc, line 24: I would wrap the cast to integer in a try [...] catch statement in case the user set it to something fishy.
```suggestion
        try:
            user_max_cores = int(os.getenv('MAX_BUILD_CORES'), cpu_count)
        except ValueError as e:
            # print useful error message
            raise e
        max_cores = min(cpu_count, user_max_cores)
```

setup.py
```diff
@@ -20,7 +20,14 @@
 
 def _get_cpu_count():
     if platform.system() != "Windows":
-        return os.cpu_count()
+        cpu_count = os.cpu_count()
+        try:
+            user_max_cores = int(os.getenv('MAX_BUILD_CORES'), cpu_count)
```
Xarthisius, line 25:
```suggestion
            user_max_cores = int(os.getenv('MAX_BUILD_CORES', cpu_count))
```

brittonsmith: oops, yes.

setup.py
```diff
@@ -20,7 +20,14 @@
 
 def _get_cpu_count():
     if platform.system() != "Windows":
-        return os.cpu_count()
+        cpu_count = os.cpu_count()
+        try:
+            user_max_cores = int(os.getenv('MAX_BUILD_CORES', cpu_count))
+        except ValueError as e:
+            print ("MAX_BUILD_CORES needs to be set to an integer.")
+            raise e
```
neutrinoceros, line 28: unless you mean to explicitly use variable `e`, you can simply write
```suggestion
        except ValueError:
            print ("MAX_BUILD_CORES needs to be set to an integer.")
            raise
```

setup.py
```diff
@@ -20,7 +20,14 @@
 
 def _get_cpu_count():
     if platform.system() != "Windows":
```
neutrinoceros, line 22: while you're at it, since the function is becoming bigger, I'd suggest getting the windows-only lines closer, like so
```python
if platform.system() == "Windows":
    return 0
... # the rest of the function goes here
```

setup.py
munkm: Many of our exceptions use e, so I don't think it's an issue to use it here. Maybe it would be a more helpful error message to include whatever e is in the log message though.

brittonsmith:
I refactored this per @neutrinoceros's and @munkm's suggestions. Now we raise another `ValueError` with the useful message. The original `ValueError` will also be shown above it.

setup.py
```diff
@@ -19,10 +19,18 @@
...
-        return os.cpu_count()
-    return 0
+    if platform.system() == "Windows":
+        return 0
 
+    cpu_count = os.cpu_count()
+    try:
+        user_max_cores = int(os.getenv('MAX_BUILD_CORES', cpu_count))
+    except ValueError:
+        raise ValueError(
```
neutrinoceros, line 29: So, while this looks fine, I think it’s not the best way to reraise an exception, see https://stackoverflow.com/questions/9157210/how-do-i-raise-the-same-exception-with-a-custom-message-in-python

munkm:
All of the test failures look to match those in yt-project/yt#2677.

setup.py
brittonsmith: I read through that originally when I was looking for a better way, but I missed the solution presented lower down. The change to the resulting error message is pretty minor, but better. The chances of the user ending up here seem pretty low.

PR #2683, opened 2020-06-24
Parallelize pixelize_sph_kernel_projection. Fixes #2682

### LINKED ISSUE #2682: SPH pixelization routine is slow
### Bug report

**Bug summary**

One of the cookbook recipes takes a lot of time on testing box:

```
$ time python doc/source/cookbook/image_resolution.py 

real	16m6.577s
user	16m5.294s
sys	0m4.467s
```

I think that the `pixelize_sph_kernel_projection` could be easily parallelized to circumvent this.

Xarthisius:
### PR Summary

This PR parallelizes sph kernel projection using OpenMP. On my laptop the speed up is modarate (<0.5 for 4 cores), but it yields a significant improvement on testing infra:

```
$ time python doc/source/cookbook/image_resolution.py   # without this patch

real	16m6.577s
user	16m5.294s
sys	0m4.467s

$ time python doc/source/cookbook/image_resolution.py # with this patch, all cores (40)
real	3m11.529s
user	23m2.800s
sys	0m14.970s

$ export OMP_NUM_THREADS=8  # the default setting we use during a test suite build
$ time python doc/source/cookbook/image_resolution.py 

real	4m9.637s
user	14m26.263s
sys	0m2.226s
```

### DIFF
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -1039,7 +1039,7 @@ def pixelize_sph_kernel_projection(
         #   1) use multiple cores to process individual particles (the outer loop)
         #   2) use multiple cores to process individual pixels for a given particle
         #      (the inner loops)
-        # Depending on the ratio of particles' "sphere of influnce" (a.k.a. the smoothing
+        # Depending on the ratio of particles' "sphere of influence" (a.k.a. the smoothing
         # length) to the physical width of the pixels, different parallelization
         # strategies may yield different speed-ups. Strategy #1 works better in the case
         # of lots of itty bitty particles. Strategy #2 works well when we have a
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -33,7 +33,7 @@ from yt.utilities.lib.fp_utils cimport (
 from yt.utilities.exceptions import YTElementTypeNotRecognized, YTPixelizeError
 
 from cpython.exc cimport PyErr_CheckSignals
-from cython.parallel cimport prange
+from cython.parallel cimport parallel, prange
 from libc.stdlib cimport free, malloc
 from vec3_ops cimport cross, dot, subtract
 
@@ -993,22 +993,21 @@ def pixelize_sph_kernel_projection(
 
     cdef np.intp_t xsize, ysize
     cdef np.float64_t x_min, x_max, y_min, y_max, prefactor_j
-    cdef np.int64_t xi, yi, x0, x1, y0, y1
+    cdef np.int64_t xi, yi, x0, x1, y0, y1, xxi, yyi
     cdef np.float64_t q_ij2, posx_diff, posy_diff, ih_j2
     cdef np.float64_t x, y, dx, dy, idx, idy, h_j2, px, py
     cdef np.float64_t period_x, period_y
     cdef int index, i, j, ii, jj
     cdef np.float64_t[:] _weight_field
-    cdef int xiter[2]
-    cdef int yiter[2]
-    cdef np.float64_t xiterv[2]
-    cdef np.float64_t yiterv[2]
+    cdef int * xiter
+    cdef int * yiter
+    cdef np.float64_t * xiterv
+    cdef np.float64_t * yiterv
+    cdef np.float64_t * local_buf
 
     if weight_field is not None:
         _weight_field = weight_field
 
-    xiter[0] = yiter[0] = 0
-    xiterv[0] = yiterv[0] = 0.0
     if period is not None:
         period_x = period[0]
         period_y = period[1]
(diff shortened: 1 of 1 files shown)


chummels:
This is awesome!

Xarthisius:
After switching to inner loop parallelization:

```
(blah) fido@c1af1189021c /tmp/yt-4 $ export OMP_NUM_THREADS=8
(blah) fido@c1af1189021c /tmp/yt-4 $ time python doc/source/cookbook/image_resolution.py 

real	2m38.511s
user	17m34.109s
sys	0m2.471s
```

matthewturk:
I think it's worth leaving a comment in the code mentioning that there are two different regimes where we could apply parallelization.  This was optimized for the regime where we have a not-very-large-number of reasonably large-compared-to-pixels particles.  So in these, we have a good number of `yi` values that each one can take on during an `xi` iteration, so there's "work to be done" in that loop.  In the case of lots of itty bitty particles, this wouldn't be quite as efficient.

Xarthisius:
> You mentioned in the PR that one of these changes is the inner loop parallelization? Or did that get overwritten in a push?

All three methods are now using parallelization on the inner loop.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1003,9 +1003,27 @@ def pixelize_sph_kernel_projection(
...
+        # Depending on the ratio of particles' "sphere of influnce" (a.k.a. the smoothing
+        # length) to the physical width of the pixels, different parallelization
+        # strategies may yield different speed-ups. Strategy #1 works better in the case
+        # of lots of itty bitty particles. Strategy #2 works well when we have a
+        # not-very-large-number of reasonably large-compared-to-pixels particles. We
+        # currently employ #1 as its workload is more even and consistent, even though it
+        # comes with a price of an additional, per thread memory for storing the
+        # intermediate results.
+
+        local_buff = <np.float64_t *> malloc(sizeof(np.float64_t) * xsize * ysize)
```
matthewturk, line 1022: I'm surprised this doesn't need to be `cdef`'d.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1301,7 +1330,13 @@ def pixelize_sph_kernel_slice(
                         continue
 
                     # see equations 6, 9, and 11 of the SPLASH paper
-                    buff[xi, yi] += prefactor_j * kernel_func(q_ij)
+                    local_buff[xi + yi*xsize] += prefactor_j * kernel_func(q_ij)
+
+        with gil:
```
matthewturk, line 1335: as long as this is the recommended way to do the reduce, i think it's okay, but i had to check indentation levels to make sure I understood what was happening and which scope this was in.

yt/utilities/lib/pixelization_routines.pyx
Xarthisius: I doesn't change the performance, but I fixed this anyway in 2909cef

yt/utilities/lib/pixelization_routines.pyx
Xarthisius: I don't know if it's **the** recommended, but that was the only one that worked :)

Xarthisius:
@yt-fido test this please

matthewturk:
@Xarthisius I'd like this to go in, but the conflicts I see are not immediately obvious to me.  Can you take another shot?

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1031,9 +1032,33 @@ def pixelize_sph_kernel_projection(
...
-    with nogil:
+    with nogil, parallel():
         # loop through every particle
-        for j in range(0, posx.shape[0]):
+        # NOTE: this loop can be quite time consuming. However it is easily
+        # parallelizable in multiple ways, such as:
+        #   1) use multiple cores to process individual particles (the outer loop)
+        #   2) use multiple cores to process individual pixels for a given particle
+        #      (the inner loops)
+        # Depending on the ratio of particles' "sphere of influnce" (a.k.a. the smoothing
```
neutrinoceros, line 1042:
```suggestion
        # Depending on the ratio of particles' "sphere of influence" (a.k.a. the smoothing
```

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1447,6 +1500,7 @@ def pixelize_sph_kernel_arbitrary_grid(np.float64_t[:, :, :] buff,
     kernel_func = get_kernel_func(kernel_name)
 
     with nogil:
+        # TODO make this parallel without using too much memory
```
neutrinoceros, line 1503: Just to be clear, is this a leftover that you forgot to address or are you planning to address it in the future ?

neutrinoceros (approved):
This looks good to me. My expertise in Cython is still extremely shallow but I think I can approve of the changes presented here.

yt/utilities/lib/pixelization_routines.pyx
Xarthisius: I plan to address it in the future, provided that I'll figure out how to do that...

neutrinoceros: Alright, just checking !

PR #2684, opened 2020-06-24
[BUGFIX] Fix athena and gizmo tests

Xarthisius:
## PR Summary
This should fix the Athena and Gizmo errors appearing in #2677.

### DIFF
--- a/yt/frontends/gizmo/tests/test_outputs.py
+++ b/yt/frontends/gizmo/tests/test_outputs.py
@@ -67,7 +67,7 @@ def test_gas_particle_fields():
     ptype = "PartType0"
     derived_fields = []
     # Add species fields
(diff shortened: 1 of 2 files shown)


yt/frontends/gizmo/tests/test_outputs.py
```diff
@@ -80,7 +80,7 @@ def test_gas_particle_fields():
     
     ptype = "gas"
     derived_fields = []
-    for species in ["H", "H_p0", "H_p1"]:
+    for species in ["H_p0", "H_p1"]:
```
cphyc, line 83: What is the rationale behind not testing "H" anymore?

Xarthisius: It doesn't exist anymore https://github.com/yt-project/yt/pull/2336

PR #2686, opened 2020-06-25
Update OpenPMD test to include 'nbody' ptype

Xarthisius:
## PR Summary

This should fix the OpenPMD errors appearing in #2677.

Update: I sneaked in a minor cosmetic fix for Arepo, that I botched in #2677. It's only avoiding doing the same thing twice.

### DIFF
--- a/yt/frontends/open_pmd/tests/test_outputs.py
+++ b/yt/frontends/open_pmd/tests/test_outputs.py
@@ -1,3 +1,4 @@
+from itertools import product
 from yt.frontends.open_pmd.data_structures import \
     OpenPMDDataset
 from yt.testing import \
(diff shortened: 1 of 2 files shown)


yt/frontends/open_pmd/tests/test_outputs.py
```diff
@@ -198,39 +133,20 @@ def test_no_particles_out():
...
+        ('openPMD', 'J_y'),
+        ('openPMD', 'J_z'),
+        ('openPMD', 'e-chargeDensity')
+    ]
     domain_dimensions = [32, 64, 64] * np.ones_like(dss[0].domain_dimensions)
     domain_width = [0.0002752, 0.0005504, 0.0005504] * np.ones_like(dss[0].domain_left_edge)
 
     assert_equal(len(dss), 101)
-    for ds in dss:
+    for i in range(0, len(dss), 20):  # Test only every 20th ds out of the series
```
munkm, line 148: Is the reason for one out of every 20th to speed up testing? or something else?

Xarthisius: Yes, speed. Since that test doesn't do anything different for an individual ds right now, there's really no point in doing it for each and everyone.

munkm: Makes sense!

Issue #2685, opened 2020-06-24
"unyt.exceptions" 


Provider10:
### Bug report

**Bug summary**

On nightly build, when importing yt, the following error comes up: "No module named 'unyt'" I assume this is just a typo? 

**Code for reproduction**

`import yt`

**Actual outcome**

![Screen Shot 2020-06-24 at 4 28 38 PM](https://user-images.githubusercontent.com/7698003/85624785-6f4d9280-b638-11ea-973b-23a7eccdeacc.png)

```
# If applicable, paste the console output here
#
#
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-2d2292a375dc> in <module>
----> 1 import yt

~/repos/yt-git/yt/__init__.py in <module>
     18 import numpy # In case anyone wishes to use it by name
     19 
---> 20 from yt.funcs import \
     21     iterable, \
     22     get_memory_usage, \

~/repos/yt-git/yt/funcs.py in <module>
     27 from yt.utilities.logger import ytLogger as mylog
     28 from yt.utilities.lru_cache import lru_cache
---> 29 from yt.utilities.exceptions import YTInvalidWidthError
     30 from yt.extern.tqdm import tqdm
     31 from yt.units import YTArray, YTQuantity

~/repos/yt-git/yt/utilities/exceptions.py in <module>
      3 import os.path
      4 
----> 5 from unyt.exceptions import UnitOperationError
      6 
      7 class YTException(Exception):

```

**Expected outcome**

**Version Information**

  * Operating System: 
  * Python Version: 3.7.7
  * yt version: Latest nightly build 
  * Other Libraries (if applicable): 

Conda installed, updated to version 3.4.1 . Channel py37h33eb3a4_2
`conda install -c yt-project/label/dev `

Xarthisius:
Hi @Provider10 and thanks for the report! I'll fix the build asap. In the meantime you can do `conda install unyt` to fix this.

Xarthisius:
I just checked and conda nightlies haven't been built for past 9 months. I'm not sure how did you end up with that state.
I'd suggest removing conda package and installing yt straight via pip:

```
pip install git+https://github.com/yt-project/yt@master
```

Provider10:
Hi @Xarthisius. Thanks! I got that from the README that's included in this branch. Perhaps that is something that should be included? Or is this build still useful somehow? Thanks again!

Provider10:
[...]

PR #2689, opened 2020-06-25
Remove the reference to conda nighthlies since they're not being buil…

Xarthisius:
…t anymore

As was mentioned in #2685 we're encouraging people to install conda nightlies in our README.md, even though the last build happened 9 months ago...

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,6 @@ or with pip:
 pip install yt
 ```
 
-If you want the latest nightly build, you can manually install from our
-repository:
-
-```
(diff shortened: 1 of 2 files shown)


munkm:
@Xarthisius there's also a reference to the nightly builds in `doc/source/installing.rst`. Would you mind removing that reference too?

Xarthisius:
> @Xarthisius there's also a reference to the nightly builds in `doc/source/installing.rst`. Would you mind removing that reference too?

Done

PR #2691, opened 2020-06-25
deprecate simulation answer testing framework

munkm:
## PR Summary

Deprecates `can_run_sim` and `requires_sim` in our answer testing framework.

These functions aren't used anywhere by our answer tests, so they should probably be deprecated. 

If anybody feels strongly about it, we could delete them too. I don't see why these would be used downstream. But maybe I'm missing something. But deprecation feels like the safer option if we don't know for sure. 

@brittonsmith you added these in 2014. Do we need them anymore? 

## PR Checklist


brittonsmith:
`requires_sim` is used in the `yt_astro_analysis` tests. If we're going to remove it, could it be moved there?

matthewturk:
@brittonsmith moving that one to astro analysis seems like a good idea to me

matthewturk:
Once tests pass I'll merge this

neutrinoceros:
@munkm could you update the branch (rebasing is probably the cleanest option) ?

neutrinoceros:
@yt-fido test this please

PR #2692, opened 2020-06-25
make YTOutputNotIdentified subclass FileNotFoundError

neutrinoceros:
## PR Summary

This is part of a larger effort to harmonise and simplify the data loading api.
This change makes catching `YTOutputNotIdentified` and the builtin `FileNotFoundError` redundant. The idea is that it makes it easier to handle loading errors on the user side since it doesn't require the additional import from `yt.utilities.exceptions`.

It makes this user script relevant
```python
import yt
try:
    yt.load("nothing")
except FileNotFoundError:
    print("gotcha")
```
While not breaking downstream code that would specifically catch `YTOutputNotIdentified`.

### DIFF
--- a/yt/utilities/exceptions.py
+++ b/yt/utilities/exceptions.py
@@ -11,7 +11,7 @@ class YTException(Exception):
 
 # Data access exceptions:
 
-class YTOutputNotIdentified(YTException):
+class YTOutputNotIdentified(YTException, FileNotFoundError):
     def __init__(self, args, kwargs):
         self.args = args
         self.kwargs = kwargs


neutrinoceros:
On second thoughts I think this is the wrong approach: if invalid but existing data files are passed, it doesn’t make much sense to cast this as FileNotFoundError

Another take :
Currently, yt.load() only fails with YTOutputUnidentifed, which makes less sense than FileNotFoundError in case the argument represents a file that’s simply not there. It could be interesting to either diversify it’s raises, or alternatively better convey the error through the error messages.

PR #2694, opened 2020-06-25
bugfix: Account for rounding errors in grid reconstruction check (stream frontend)

### LINKED ISSUE #2665: Generic AMR loader fail to find cell edge & grid edge alignment
### Bug report

**Bug summary**
Generic AMR loader fail to find cell edge & grid edge alignment
Giving error: 
`yt.utilities.exceptions.YTIllDefinedAMR: Grids on the level 2 are not properly aligned with cell edges on the parent level (x axis)`

At line 905 in data_structures.py, it trying to find cell edge & grid edge alignment by set subtraction. But somehow array generated by Numpy linspace is not considered as the same as manually inputted array. 
**Code for reproduction**

```python
>>> list1 = np.linspace(0,1.,10,endpoint=False)
>>> list1
array([0. , 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9])
>>> list2 = np.array([0. , 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9])
>>> set(list1)-set(list2)
{0.30000000000000004, 0.6000000000000001, 0.7000000000000001}
```
Force data type to be the same seems solve the problem.
```python
>>> list1 = np.linspace(0,1.,10,endpoint=False,dtype=np.float32)
>>> list2 = np.array([0. , 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]).astype(np.float32)
>>> set(list1) - set(list2)
set()
```
**Version Information**

  * Operating System: OSX Catalina
  * Python Version:      Python 3.7.3
  * yt version:               3.6.dev0

neutrinoceros:
## PR Summary

fix #2665 as per @matthewturk's suggestion

## PR Checklist

### DIFF
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -901,7 +901,7 @@ def load_amr_grids(grid_data, domain_dimensions,
                 # check for uniqueness
                 found[i] = len(np.nonzero(matches)[0]) == 1
 
-            if any(~found):
+            if not found.all():
                 raise YTIllDefinedAMR(lvl, ax)
 
     if length_unit is None:
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -890,7 +890,12 @@ def load_amr_grids(grid_data, domain_dimensions,
             cell_edges = np.linspace(domain_left_edge[iax],
                                      domain_right_edge[iax],
                                      dims[iax], endpoint=False)
-            if set(grid_left_edges[idx, iax]) - set(cell_edges):
+
+            # check that every value in cell_edges closely approximates
+            # a member of grid_left_edges[idx, iax]
+            # (account for rounding errors up to machine precision)
+            diffs = [min(np.abs(cell_edges-ss)) for ss in grid_left_edges[idx, iax]]
+            if max(diffs) > 1e-17:
                 raise YTIllDefinedAMR(lvl, ax)
 
     if length_unit is None:
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -894,8 +894,14 @@ def load_amr_grids(grid_data, domain_dimensions,
             # check that every value in cell_edges closely approximates
             # a member of grid_left_edges[idx, iax]
             # (account for rounding errors up to machine precision)
-            diffs = [min(np.abs(cell_edges-ss)) for ss in grid_left_edges[idx, iax]]
-            if max(diffs) > 1e-17:
+            atol = grid_left_edges[idx, iax].min() / 1e3
+            found = np.zeros_like(grid_left_edges[idx, iax], dtype=bool)
+            for i, gle in enumerate(grid_left_edges[idx, iax]):
+                matches = np.isclose(cell_edges, gle, atol=atol, rtol=1e-16)
+                # check for uniqueness
+                found[i] = len(np.nonzero(matches)[0]) == 1
+
+            if any(~found):
                 raise YTIllDefinedAMR(lvl, ax)
 
     if length_unit is None:
--- a/yt/frontends/stream/tests/test_stream_amrgrids.py
+++ b/yt/frontends/stream/tests/test_stream_amrgrids.py
@@ -89,4 +89,4 @@ def test_load_despite_rounding_errors_during_grid_construction():
 
     bbox = np.array([[-0.18,0.18],[-0.18,0.18],[-0.18,0.18]])
     dimensions = [128,128,128]
-    ds = load_amr_grids(grid_data, dimensions, bbox)
+    load_amr_grids(grid_data, dimensions, bbox)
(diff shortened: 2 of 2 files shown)


yt/frontends/stream/tests/test_stream_amrgrids.py
```diff
@@ -69,3 +69,23 @@ def load_grids():
         load_amr_grids(grid_data, dims, bbox=bbox, periodicity=(0, 0, 0),
                        length_unit=1.0, refine_by=2)
     assert_raises(YTIllDefinedAMR, load_grids)
+
+
+def test_2665():
```
munkm, line 74: I think it would be helpful if this test name were something other than the issue number. It's ok if the name is long, it's better if it's descriptive!

yt/frontends/stream/data_structures.py
```diff
@@ -890,7 +890,12 @@ def load_amr_grids(grid_data, domain_dimensions,
             cell_edges = np.linspace(domain_left_edge[iax],
                                      domain_right_edge[iax],
                                      dims[iax], endpoint=False)
-            if set(grid_left_edges[idx, iax]) - set(cell_edges):
+
+            # check that every value in cell_edges closely approximates
```
munkm, line 894: Ah, this is a good inline comment. Thank you!

yt/frontends/stream/tests/test_stream_amrgrids.py
neutrinoceros: oh yes of course, I did mean to change it before opening the PR !

yt/frontends/stream/data_structures.py
```diff
@@ -890,7 +890,18 @@ def load_amr_grids(grid_data, domain_dimensions,
...
+            # a member of grid_left_edges[idx, iax]
+            # (account for rounding errors up to machine precision)
+            atol = grid_left_edges[idx, iax].min() / 1e3
+            found = np.zeros_like(grid_left_edges[idx, iax], dtype=bool)
+            for i, gle in enumerate(grid_left_edges[idx, iax]):
+                matches = np.isclose(cell_edges, gle, atol=atol, rtol=1e-16)
+                # check for uniqueness
+                found[i] = len(np.nonzero(matches)[0]) == 1
```
matthewturk, line 902: Wouldn't it be fun if in a few years it turns out there was some weird edge case where this could be 2 or more?!  :rofl:

yt/frontends/stream/data_structures.py
```diff
@@ -890,7 +890,18 @@ def load_amr_grids(grid_data, domain_dimensions,
...
+
+            if any(~found):
```
matthewturk, line 904: This is not a necessary change, but since `found` is defined as an array, this could be: `if not found.any()`

yt/frontends/stream/tests/test_stream_amrgrids.py
```diff
@@ -69,3 +69,24 @@ def load_grids():
...
+         level=1,
+         dimensions=[16,16,16])
+    ]
+
+    for g in grid_data:
+        g["density"] = np.random.random(g["dimensions"]) * 2 ** g["level"]
+
+    bbox = np.array([[-0.18,0.18],[-0.18,0.18],[-0.18,0.18]])
+    dimensions = [128,128,128]
+    load_amr_grids(grid_data, dimensions, bbox)
```
matthewturk, line 92: is there an "assert not raises" or something we could put here, to check that it is working?

and, could we add a second line, where we load it a second time, but with the left edge of one of the grids offset by some not-quite-negligible value, just to break it, and do an assert raises?

yt/frontends/stream/data_structures.py
neutrinoceros: Let's remove this line and find out ! 😇

yt/frontends/stream/data_structures.py
neutrinoceros: Elegance may not be necessary, but I'll do it anyway !

neutrinoceros: actually no ! Those are not equivalent conditions !
Screenshot 2020-06-26 at 20 47 43

yt/frontends/stream/tests/test_stream_amrgrids.py
neutrinoceros: > is there an "assert not raises" or something we could put here, to check that it is working?

Well I think this is the very definition of a passing test, and that it doesn't need a dedicated function.
It could be made more clear with comments though, I'll update this with your suggestion :)

neutrinoceros: actually adding this test made it clear that there's a problem in my fix, since the new test fails.

neutrinoceros:
I'm switching to draft since this is actually not passing its own tests and I'm still scratching my head over it.

neutrinoceros:
I really can't make sense of this, and the branch has festered. I don't think it's worth keeping, so I'll close.

PR #2695, opened 2020-06-26
Simplify loader functions

neutrinoceros:
## PR Summary

A radical refactor/simplification for `yt.load()`. Given the crucial role of this functionality, I try to give here an extensive justification to this change.

**tl;dr**
Improves both user and dev experience with docs, more useful error messages, and better maintainability.
A quick visual review of this implementation of `yt.load()` behaves with wrong inputs, compared with the current ("old")
```python
yt.load("not_a_file")
old >> YTOutputNotIdentified: provided ("not_a_file"), {} but could not load !
new >> FileNotFoundError: no such file or directory: not_a_file

yt.load("real_file_with_no_valid_frontend")
old >> YTOutputNotIdentified: provided ("real_file_with_no_valid_frontend"), {} but could not load !
new >> YTOutputNotIdentified: Could not determine input format from ('real_file_with_no_valid_frontend',) {}

# wrong input type
yt.load(yt)
old >> YTOutputNotIdentified: Supplied (<module 'yt' from 'yt/yt/__init__.py'>,) {}, but could not load!
new >> TypeError: expected str, bytes or os.PathLike object, not module

# ambiguous data type (this is a fake example using real classes)
yt.load("file_with_multiple_acceptable_formats")
old >> 
yt : [ERROR    ] 2020-08-05 20:00:55,152 Multiple output type candidates for"file_with_multiple_acceptable_formats":
yt : [ERROR    ] 2020-08-05 20:00:55,164     Possible: <class 'yt.frontends.art.data_structures.ARTDataset'>
yt : [ERROR    ] 2020-08-05 20:00:55,185     Possible: <class 'yt.frontends.boxlib.data_structures.BoxlibDataset'>
YTOutputNotIdentified: provided ("file_with_multiple_acceptable_formats",) {}, but could not load!

new >> YTAmbiguousDataType: Multiple data type candidates for file_with_multiple_acceptable_formats
  - yt.frontends.art.data_structures.ARTDataset
  - yt.frontends.boxlib.data_structures.BoxlibDataset
A workaround is to directly instanciate one of the above.
Please report this to https://github.com/yt-project/yt/issues/new
```

**rationale**
As a user, I find it confusing that yt claims it couldn't identify an output when really, the file I'm pointing to isn't even there.
So I wanted to refactor `yt.load()` to throw a more appropriate error message on when data is missing.
Because of the important complexity of the existing implementation, this snowballed a bit and resulted in a complete makeover of this core function of the package. I'm quite happy with the result, and believe it is more user-friendly as well as easier to expand or maintain.

**user facing differences**
- `yt.load()` and `yt.simulation()` now throw `FileNotFoundError` instead of `YTOuptutNotIdentified` (resp `YTSimulationNotIdentified`) if datafiles/dirs are not present. This is further supported for time series by #2690 (merged)
- Much more detailed docstrings, letting out implementation detail and giving a more complete description of usage.
- `yt.load()` now seamlessly supports data loading from `str` `os.Pathlike` or `byte` types with `~` user token expansion, or anything that's supported by `os.path.expanduser` now and in the future.
- `YTOutputNotIdentified` error message is slightly rephrased. (reason: current version says it "can not load" but raising this error could make sense outside of loading functions)
- removed `EnzoRunDatabase` since it's was never usable in the first place according to @matthewturk 

**dev side differences**
1) The current implementation for `yt.load()` is more or less agnostic with respect to its positional arguments and looks for a valid path/url in all of them before passing them down to the obj (Dataset subclass) constructor.
In this proposed refactored version, `yt.load()` takes one positional argument that holds the data file/dir path/pattern/url.
I haven't found a single case in the documentation or tests where it is used differently so I think it is reasonable to simplify the functionality.
2) I removed some special-case handling that looked completely dead (`_stream` types ?)
3) Hopefully the refactored function is simpler to debug and maintain in the future. It should also be much more new-contributor friendly, which seems important given that this is the one function everyone knows and uses. 

## ⚠️Discussion of potentially controversial changes

Naming the first positional argument of `yt.load()` is delicate because of the following implications:
- it becomes a reserved name that can not be used in `__init__` methods of `Dataset` subclasses
- it allows users to write scripts like so:
```python
yt.load(fn="my/path/to/datafile.hdf5")
```
which means that we can't change the name afterwards without breaking backward compatibility.
Alternative:
```python
# 1) this proposal
def load(fn, *args, **kwargs):
    fn = os.path.expanduser(fn)
    ...

# 2) a more conservative alternative
def load(*args, **kwargs):
    fn = args[0] = os.path.expanduser(args[0])
    ...
```
A drawback of this alternative implementation is that less informative error messages are produced:
```python
yt.load()
1 >> load() missing 1 required positional argument: 'fn'
2 >> IndexError: tuple index out of range
```
note: I chose to name the param `fn` for "filename" because it fits some existing internal api, but mostly because it's a short and obscure enough name that it doesn't encourage the use of explicit keyword syntax as above.
 
I also changed `yt.simulation()`'s first argument name to `fn` to harmonise with `yt.load()`. In the documentation and tests, this argument was never written as a keyword, so I expect this change shouldn't affect too many users.
A possible way to handle this more smoothly would be to re-add `parameter_filename` as a deprecated keyword alias. In practice however this deprecation process requires that all arguments be given default values (undesirable), so imo it's not worth it.

## PR Checklist

**Note to reviewers**
I aggressively rebased this branch throughout its development to try and provide a narrative-guided history. There shouldn't be back-and forth changes in the presented history so it's easier to read one commit at a time.


Xarthisius:
This allowed me to spot #2696 so good job! However, it also suggests that `yt/utilities/answer_testing/utils.py:can_run_ds` and possibly others need to be updated ;)

neutrinoceros:
> However, it also suggests that yt/utilities/answer_testing/utils.py:can_run_ds and possibly others need to be updated ;)

done ! thank you for this suggestion :)
Since tests are passing I'm opening this for review

yt/convenience.py
```diff
@@ -12,112 +12,114 @@
...
-           and isinstance(args[0], str):
+           and len(args) == 0 \
+           and isinstance(fn, str):
             erdb = EnzoRunDatabase()
-            fn = erdb.find_uuid(args[0])
+            fn = erdb.find_uuid(fn)
             n = "EnzoDataset"
             if n in output_type_registry \
                and output_type_registry[n]._is_valid(fn):
                 return output_type_registry[n](fn)
```
neutrinoceros, line 81: This special case looks dead to me: it is neither hit by the tests or documented anywhere.
I'd love to get feedback on this ! Could it also be cut out into its own function ?

neutrinoceros: I'm cutting this in a clearly identified commit. I'll resurrect it if needed.

neutrinoceros:
Adding the 4.0 label because I'd like to get this ready and approved in time for the release, but if anyone feels strongly about this it can be postponed !

yt/convenience.py
neutrinoceros: Since it seems impossible to know in advance if this is still in use, I reintroduced it as a separate function `load_enzo_db()`. If accepted in this state, this will need to be clearly documented in the release notes.

yt/convenience.py
```diff
@@ -5,119 +5,124 @@
...
-    if not any(valid_file):
-        try:
-            from yt.data_objects.time_series import DatasetSeries
-            ts = DatasetSeries.from_filenames(*args, **kwargs)
-            return ts
-        except (TypeError, YTOutputNotIdentified):
-            pass
-        # We check if either the first argument is a dict or list, in which
-        # case we try identifying candidates.
-        if len(args) > 0 and isinstance(args[0], (list, dict)):
```
neutrinoceros, line 61: This is the only functionality I have completely removed. The reason for this was that I could not find any registered output type whose key starts with `"_stream"`, so I concluded this block is dead.

neutrinoceros:
> In yt/data_objects/time_series.py, I noticed time series will only work with str-like objects.

Note that because the first thing we do in `yt.load()` is passing the filename through `os.path.expanduser()`, it's _always_ a string by the time it is fed to DatasetSeries, so the current implementation _already_ supports loading time series from `os.PathLike` objects 😄

yt/convenience.py
```diff
@@ -5,119 +5,124 @@
...
-        if c._is_valid(*args, **kwargs): candidates.append(n)
-
-    # convert to classes
-    candidates = [output_type_registry[c] for c in candidates]
+            raise OSError("No such file or directory: %s" % fn)
+
+    candidates = []
+    for cls in output_type_registry.values():
+        if cls._is_valid(fn, *args, **kwargs):
+            candidates.append(cls)
```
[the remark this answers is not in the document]
neutrinoceros: I originally wrote it this way. The reason I switched back is that we'll soon apply black to the code base and it would re-expand the list-comprehension into 4 lines, nullifying the compacity gain and not more readable imo. :)

yt/convenience.py
```diff
@@ -5,119 +5,124 @@
...
-            pass
-        # We check if either the first argument is a dict or list, in which
-        # case we try identifying candidates.
-        if len(args) > 0 and isinstance(args[0], (list, dict)):
-            # This fixes issues where it is assumed the first argument is a
-            # file
-            types_to_check = dict((n, v) for n, v in
-                    output_type_registry.items() if n.startswith("stream_"))
-            # Better way to do this is to override the output_type_registry
+    fn = os.path.expanduser(fn)
```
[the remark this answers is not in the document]
neutrinoceros: true. I relied on the tests and documentation for this (no test was broken by the change), but I haven't yet gone through a complete review of frontends. I'm fairly confident the feature wasn't used at all though :)

yt/tests/test_load_errors.py
```diff
@@ -0,0 +1,26 @@
...
+from yt.convenience import load, simulation
+from yt.testing import assert_raises
+from yt.utilities.exceptions import YTOutputNotIdentified, YTSimulationNotIdentified
+
+def test_load_unexisting_data():
+    with tempfile.TemporaryDirectory() as tmpdir:
+        assert_raises(OSError, load, os.path.join(tmpdir, "not_a_file"))
+        assert_raises(OSError, simulation, os.path.join(tmpdir, "not_a_file"), "Enzo")
+
+        # this one is an design choice: it is preferable to report the most imoportant
```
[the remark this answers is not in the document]
neutrinoceros: thanks ! nothing like a fresh pair of eyes to find those !

yt/utilities/answer_testing/framework.py
```diff
@@ -272,7 +272,7 @@ def can_run_ds(ds_fn, file_check = False):
             result_storage is not None
     try:
         load(ds_fn)
-    except YTOutputNotIdentified:
+    except (OSError, YTOutputNotIdentified):
```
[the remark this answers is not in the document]
neutrinoceros: Excellent point. @Xarthisius, do you agree ?

yt/frontends/exodus_ii/simulation_handling.py
```diff
@@ -95,15 +94,14 @@ def _check_for_outputs(self, potential_outputs):
...
-                except YTOutputNotIdentified:
-                    mylog.error("Failed to load %s", output)
+            try:
+                ds = load(output)
+                my_storage.result = {"filename": output,
+                                     "num_steps": ds.num_steps}
+            except OSError:
+                pass
+            except YTOutputNotIdentified:
+                mylog.error("Failed to load %s", output)
```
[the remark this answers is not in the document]
neutrinoceros: Definitely. I consider the factorisation to be beyond the point of this PR though, but it wouldn't be hard to deduplicate these and it's certainly a good idea to.

yt/convenience.py
```diff
@@ -5,119 +5,124 @@
...
-    This function attempts to determine the base data type of a filename or
-    other set of arguments by calling
-    :meth:`yt.data_objects.static_output.Dataset._is_valid` until it finds a
-    match, at which point it returns an instance of the appropriate
-    :class:`yt.data_objects.static_output.Dataset` subclass.
+    Additional arguments, if any, are passed down to the return class.
+
+    Returns
+    -------
+    Dataset
```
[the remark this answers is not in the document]
neutrinoceros: I think that one link in the header of the docstring is enough for the website to be user-friendly, while they make the raw visualisation harder from a python session. I'm okay with changing it if it's considered best practice for yt though, but I don't know of any guideline on this point.

yt/convenience.py
```diff
@@ -5,119 +5,124 @@
...
-            mylog.error("None of the arguments provided to load() is a valid file")
-            mylog.error("Please check that you have used a correct path")
-            raise YTOutputNotIdentified(args, kwargs)
-    for n, c in types_to_check.items():
-        if n is None: continue
-        if c._is_valid(*args, **kwargs): candidates.append(n)
-
-    # convert to classes
-    candidates = [output_type_registry[c] for c in candidates]
+            raise OSError("No such file or directory: %s" % fn)
```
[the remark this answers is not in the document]
neutrinoceros: Thanks for this suggestion. I'm not a fan of exclamative points in error messages but I can not think of anything better myself for now. I'll come back to this if I can think of something !

yt/convenience.py
```diff
@@ -5,119 +5,124 @@
...
-        # case we try identifying candidates.
-        if len(args) > 0 and isinstance(args[0], (list, dict)):
-            # This fixes issues where it is assumed the first argument is a
-            # file
-            types_to_check = dict((n, v) for n, v in
-                    output_type_registry.items() if n.startswith("stream_"))
-            # Better way to do this is to override the output_type_registry
+    fn = os.path.expanduser(fn)
+
+    if any([wildcard in fn for wildcard in "[]?!*"]):
```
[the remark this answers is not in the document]
neutrinoceros: Elegant !

yt/convenience.py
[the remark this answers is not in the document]
neutrinoceros: done !

yt/convenience.py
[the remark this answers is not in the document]
neutrinoceros: Also this would require to check that the data dir was defined. Otherwise this would print
```
 Also tried /does/not/exist/<fn> with no better luck, sorry!
```
which isn't exactly more clear than leaving it out 😅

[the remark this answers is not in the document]
neutrinoceros: done !

neutrinoceros:
@yt-fido test this please

yt/convenience.py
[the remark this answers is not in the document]
neutrinoceros: note: #2739 is even more proof that this change is okay, and updates all frontends to reflect the de facto general behaviour

matthewturk:
As I mentioned on the PR triage, there are precisely zero users of `EnzoRunDatabase`, if I am remembering its history correctly.  I believe I halfway implemented something in enzo that created sqlite database and it was never finalized nor included in enzo.

neutrinoceros:
> As I mentioned on the PR triage, there are precisely zero users of EnzoRunDatabase, if I am remembering its history correctly. I believe I halfway implemented something in enzo that created sqlite database and it was never finalized nor included in enzo.

Thanks a lot @matthewturk for clarifying this, I removed the functionality completely and edited the PR's text !

yt/convenience.py
```diff
@@ -6,132 +6,114 @@
...
-    raise YTOutputNotIdentified(args, kwargs)
-
-
-def simulation(parameter_filename, simulation_type, find_outputs=False):
+        return candidates[0](fn, *args, **kwargs)
+
+    if len(candidates) > 1:
+        mylog.error("Multiple output type candidates for %s:", fn)
+        for c in candidates:
+            mylog.error("    Possible: %s", c)
```
matthewturk, line 78: Instead, could we subclass `YTOutputNotIdentified` with something like `YTOutputMultipleIdentified` or something so that even if they have logging off they'll see the issue?

yt/convenience.py
```diff
@@ -6,132 +6,114 @@
...
-    simulation type.
+    Load a simulation time series object of the specified simulation type.
+
+    Parameters
+    ----------
+    fn : str, os.Pathlike, or byte (types supported by os.path.expandusers)
+        Name of the data file or directory.
+
+    simulation_type : str
+        E.g. 'Enzo'
```
matthewturk, line 93: is `Enzo` caps in the simulation type?

yt/frontends/enzo/simulation_handling.py
```diff
@@ -656,18 +656,18 @@ def _check_for_outputs(self, potential_outputs):
...
-                    mylog.error("Failed to load %s", filename)
+            try:
+                ds = load(filename)
+                my_storage.result = {
+                    "filename": filename,
+                    "time": ds.current_time.in_units("s"),
+                }
+                if ds.cosmological_simulation:
+                    my_storage.result["redshift"] = ds.current_redshift
+            except OSError:
```
matthewturk, line 667: can we move this up so that we do a minimal catching?

yt/frontends/exodus_ii/simulation_handling.py
```diff
@@ -90,14 +89,13 @@ def _check_for_outputs(self, potential_outputs):
...
-                    ds = load(output)
-                    if ds is not None:
-                        num_steps = ds.num_steps
-                        my_storage.result = {"filename": output, "num_steps": num_steps}
-                except YTOutputNotIdentified:
-                    mylog.error("Failed to load %s", output)
+            try:
+                ds = load(output)
+                my_storage.result = {"filename": output, "num_steps": ds.num_steps}
+            except OSError:
```
matthewturk, line 95: can we move these two catches up?

yt/frontends/gadget/simulation_handling.py
```diff
@@ -521,18 +521,18 @@ def _check_for_outputs(self, potential_outputs):
...
-                    mylog.error("Failed to load %s", output)
+            try:
+                ds = load(output)
+                my_storage.result = {
+                    "filename": output,
+                    "time": ds.current_time.in_units("s"),
+                }
+                if ds.cosmological_simulation:
+                    my_storage.result["redshift"] = ds.current_redshift
+            except OSError:
```
matthewturk, line 532: same thing about the catching

yt/tests/test_load_errors.py
```diff
@@ -0,0 +1,43 @@
+import os
+import tempfile
+from pathlib import Path
+
+from yt.convenience import load, simulation
+from yt.testing import assert_raises
+from yt.utilities.exceptions import YTOutputNotIdentified, YTSimulationNotIdentified
+
+
+def test_load_unexisting_data():
```
matthewturk, line 10:
```suggestion
def test_load_nonexistent_data():
```

yt/convenience.py
neutrinoceros: Sounds great, I'll address this !

yt/convenience.py
neutrinoceros: according to this test, yes
https://github.com/yt-project/yt/blob/fdc9210a35e41b45c114f271798ea46c2179d2eb/doc/source/cookbook/time_series_profiles.py#L4

yt/frontends/enzo/simulation_handling.py
neutrinoceros: good idea. This and similar suggestions should be addressed now !
Note that I also changed it so that the logging message happens with both OSError and YTOutputNotIdentifiedError (I don't really see why I treated them separately in retrospect)

yt/convenience.py
neutrinoceros: How about `YTAmbiguousDataType` ? I'm thinking the term "output" is slightly inappropriate here since it's actually _input_ from yt's point of view. I could also make `YTOutputNotIdentified` an alias for `YTUnidentifiedDataType` for consistency. WDYT ?

neutrinoceros: (the alias stuff can be done in a follow up PR)

neutrinoceros: @matthewturk updated the PR description with this. It looks SO much better, thanks a lot !

neutrinoceros:
Ah, with the new `YTAmbiguousDataType` error, a lot of problems are now showing up on Jenkins. I don't get how this is possible since the errors should have shown up as `YTOutputNotIdentified` in previous version, and because the new error is subclassing it, it should still be caught in try blocks.
Even weirder is the fact that Jenkins started failing only after a minor change in the error message, not at the introduction of the error itself.

neutrinoceros:
Nevermind the failure was really caused by a change I used to test it locally and accidentally pushed. Should be good now !

neutrinoceros:
update: I've changed my mind about the error that should be raised in cases where datafiles simply aren't there. I changed it from `OSError` to `FileNotFoundError`. Originally, I wanted to avoid `FileNotFoundError` since sometimes it's not a file that's missing, but a directory. I now realise that, unless it _does_ exist, there's no way to tell in advance if a path represents a file or a directory if it doesn't end with `/` (or `\` on windows).
Indeed, the standard library itself seems to encourage this change. For instance

```python
>>> with open("not_a_file", mode="rb") as fh:
>>>    print("really ?")

# Python < 3.3:
IOError: [Errno 2] No such file or directory: 'not_a_file'

# Python >= 3.3
FileNotFoundError: [Errno 2] No such file or directory: 'not_a_file'
```
note that `IOError` is now an alias of `OSError`

Finally, quoting the [std lib doc](https://docs.python.org/3/library/exceptions.html?highlight=filenotfounderror#FileNotFoundError)
> exception FileNotFoundError
Raised when a file or directory is requested but doesn’t exist.

yt/convenience.py
neutrinoceros: here's the follow up #2818

yt/utilities/answer_testing/framework.py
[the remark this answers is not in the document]
neutrinoceros: I'm going ahead with this (only catch FileNotFoundError)

yt/convenience.py
```diff
@@ -2,136 +2,122 @@
...
-            )
-            # Better way to do this is to override the output_type_registry
+    fn = os.path.expanduser(fn)
+
+    if any(wildcard in fn for wildcard in "[]?!*"):
+        from yt.data_objects.time_series import DatasetSeries
+
+        return DatasetSeries(fn, *args, **kwargs)
+
+    if not (os.path.exists(fn) or fn.startswith("http")):
```
cphyc, line 59:
```suggestion
    # Unless the dataset starts with http, look for it using the path or relative to the data dir (in this order).
    if not (os.path.exists(fn) or fn.startswith("http")):
```

yt/convenience.py
```diff
@@ -2,136 +2,122 @@
...
-            continue
-        if c._is_valid(*args, **kwargs):
-            candidates.append(n)
-
-    # convert to classes
-    candidates = [output_type_registry[c] for c in candidates]
+            msg = f"No such file or directory: '{fn}'"
+            if os.path.exists(data_dir):
+                msg += f"\n(Also tried '{alt_fn}')"
+            raise FileNotFoundError(msg)
```
cphyc, line 68:
```suggestion
            msg = f"No such file or directory: '{fn}'."
            if os.path.exists(data_dir):
                msg += f"\n(Also tried '{alt_fn}')."
            raise FileNotFoundError(msg)
```

yt/frontends/gadget/simulation_handling.py
```diff
@@ -521,18 +521,18 @@ def _check_for_outputs(self, potential_outputs):
...
-                    if ds is not None:
-                        my_storage.result = {
-                            "filename": output,
-                            "time": ds.current_time.in_units("s"),
-                        }
-                        if ds.cosmological_simulation:
-                            my_storage.result["redshift"] = ds.current_redshift
-                except YTOutputNotIdentified:
-                    mylog.error("Failed to load %s", output)
+            try:
```
cphyc, line 524: This whole method should be moved into the parent class, as it is used here and in `exodus_ii`.

yt/tests/test_load_errors.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        assert_raises(
+            YTSimulationNotIdentified,
+            simulation,
+            empty_file_path,
+            "unregistered_simulation_type",
+        )
+
+
+def test_load_ambiguous_data():
```
cphyc, line 53: Note: this may break the tests at some point, as two new datasets will be registered for yt. You should add a fixture to save and restore `yt.utilities.parameter_file_storage.output_file_registry` (see https://github.com/yt-project/yt/blob/master/yt/data_objects/static_output.py#L80) to prevent this from making other tests fail.

yt/utilities/answer_testing/utils.py
```diff
@@ -12,16 +12,17 @@
...
 from yt.convenience import load, simulation
 from yt.data_objects.selection_data_containers import YTRegion
 from yt.data_objects.static_output import Dataset
 from yt.frontends.ytdata.api import save_as_dataset
 from yt.units.yt_array import YTArray, YTQuantity
-from yt.utilities.exceptions import YTOutputNotIdentified
+from yt.visualization import (
+    particle_plots as particle_plots,
+    plot_window as pw,
+    profile_plotter as profile_plotter,
```
cphyc, line 24:
```suggestion
    particle_plots,
    plot_window as pw,
    profile_plotter,
```

yt/utilities/exceptions.py
```diff
@@ -14,12 +14,33 @@ def __init__(self, message=None, ds=None):
...
         self.kwargs = kwargs
 
     def __str__(self):
-        return "Supplied %s %s, but could not load!" % (self.args, self.kwargs)
+        msg = "Could not determine input format from %s" % self.filename
+        if self.args is not None:
+            msg += ", %s" % self.args
+        if self.kwargs is not None:
+            msg += ", %s" % self.kwargs
+        return msg
```
cphyc, line 28:
```suggestion
        msg += '.'        
        return msg
```

yt/utilities/exceptions.py
```diff
@@ -14,12 +14,33 @@ def __init__(self, message=None, ds=None):
...
+        self.filename = filename
+        self.candidates = candidates
+
+    def __str__(self):
+        msg = f"Multiple data type candidates for {self.filename}\n"
+        msg += "The following independent classes were detected as valid :\n"
+        for c in self.candidates:
+            msg += f"{c}\n"
+        msg += "A possible workaround is to directly instantiate one of the above.\n"
+        msg += "Please report this to https://github.com/yt-project/yt/issues/new"
```
cphyc, line 42:
```suggestion
        msg = f"Multiple data type candidates for {self.filename}.\n"
        msg += "The following independent classes were detected as valid :\n"
        for c in self.candidates:
            msg += f"{c}\n"
        msg += "A possible workaround is to directly instantiate one of the above.\n"
        msg += "Please report this to https://github.com/yt-project/yt/issues/new."
```

yt/utilities/exceptions.py
```diff
@@ -14,12 +14,33 @@ def __init__(self, message=None, ds=None):
...
         self.args = args
         self.kwargs = kwargs
 
     def __str__(self):
-        return "Supplied %s %s, but could not load!" % (self.args, self.kwargs)
+        msg = "Could not determine input format from %s" % self.filename
+        if self.args is not None:
+            msg += ", %s" % self.args
+        if self.kwargs is not None:
+            msg += ", %s" % self.kwargs
```
cphyc, line 27:
```suggestion
        msg =f "Could not determine input format from {self.filename}"
        if self.args is not None:
            msg += f", {self.args}"
        if self.kwargs is not None:
            msg += f", {self.kwargs}"
```
f-strings frenzy!

yt/tests/test_load_errors.py
cphyc: I'm actually surprised it does not break anything...

yt/convenience.py
```diff
@@ -2,136 +2,122 @@
...
-            types_to_check = dict(
-                (n, v)
-                for n, v in output_type_registry.items()
-                if n.startswith("stream_")
-            )
-            # Better way to do this is to override the output_type_registry
+    fn = os.path.expanduser(fn)
+
+    if any(wildcard in fn for wildcard in "[]?!*"):
+        from yt.data_objects.time_series import DatasetSeries
```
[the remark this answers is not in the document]
neutrinoceros: can't be done, this creates circular imports :/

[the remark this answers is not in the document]
cphyc: Oh. That's unfortunate.

yt/utilities/exceptions.py
neutrinoceros: I'd rather not add a "." at the end of the url so as to make it less error prone when it's copy-pasted

yt/utilities/exceptions.py
neutrinoceros: how about running this ?
```
flynt yt.utilities.exceptions
```
(in another PR)

yt/frontends/gadget/simulation_handling.py
neutrinoceros: Sounds good to me, but would rather do this in a follow up PR if that's okay to you :-)

yt/tests/test_load_errors.py
neutrinoceros: good thinking ! thanks for pointing this out. Should be fixed now.

yt/utilities/exceptions.py
cphyc: Fair enough

yt/frontends/gadget/simulation_handling.py
cphyc: That's fine for me.

neutrinoceros:
Many thanks to all participants, I'm so glad this is now part of yt !
For reference, #2739 and #2818 are direct follow ups to this PR.

PR #2696, opened 2020-06-26
Fix khi_cartesian_2D filename

Xarthisius:
## PR Summary

This fixes the filename used in tests to match what we have on disk.
Note: I also noticed that `Richtmyer_Meshkov_dust_2D` was missing on the testing box, so this PR should yield a lot of NoOldAnswer. I'll bump answer store, once I see that's the case.

### DIFF
--- a/yt/frontends/amrvac/tests/test_outputs.py
+++ b/yt/frontends/amrvac/tests/test_outputs.py
@@ -107,7 +107,7 @@ def test_khi_cartesian_3D():
 @requires_ds(jet_cylindrical_25D)
 def test_jet_cylindrical_25D():
    ds = data_dir_load(jet_cylindrical_25D)
-   for test in small_patch_amr(ds, ds.field_list):
+   for test in small_patch_amr(ds, _get_fields_to_check(ds)):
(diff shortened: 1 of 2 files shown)


yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -11,7 +11,7 @@
 from yt.units import YTQuantity
 
 blastwave_spherical_2D = "amrvac/bw_2d0000.dat"
-khi_cartesian_2D = "amrvac/kh_2D0000.dat"
+khi_cartesian_2D = "amrvac/kh_2d0000.dat"
```
munkm, line 14: the fact that 2d and 3D use different conventions is confusing.

neutrinoceros: indeed. The AMRVAC sample files are straight from AMRVAC's testsuite which _doesn't_ have naming conventions other than the suffix's shape `[0-9]{4}.dat`

munkm:
> so this PR should yield a lot of NoOldAnswer

🔮 🔮 🔮 🔮 🔮 🔮 🔮 🔮 🔮 🔮

PR #2698, opened 2020-06-26
Fix Rockstar PSC tests with a nulp check

matthewturk:
As noted in #2677, we're getting errors with Rockstar.  The arrays differ by about 1e-14, which is likely due to roundtripping unit conversions in different precisions.  This allows for wiggle room in the comparison of 5 units in the last place of the floating point values.


matthewturk:
Rockstar no longer fails in this, so should be good to go.

PR #2701, opened 2020-06-29
promote log message from warning to error in add_field methods

neutrinoceros:
## PR Summary

Very quick PR to promote a warning message to error since it "warns" on code doing nothing.
(as a user I personally set logger to error level and this feels like something even I should be informed of)

I'm noting that there's a lot of code duplication in those functions, I'll try to address that in a separate PR.


neutrinoceros:
Here's how I got here : I was resurrecting an old (2 months) notebook that was relying on a plugin file. The plugin was defining some fields but I forgot and tried to reuse the same name for another field in the notebook. I was clueless for a few minutes before I realised what was going on.

munkm:
Ah, ok! That's a case I hadn't considered. I see no issue with an error level logger then.

PR #2703, opened 2020-06-29
fix: Improved typing flexibility and type checking for Dataset units_override

### LINKED ISSUE #2702: undesirable behaviour units_override + json
### Bug report

**Bug summary**
`Dataset.__init__` supports units overriding through kwarg `units_override` following schema `Dict[str, Tuple]`, but will crash on `Dict[str, List]`.
This is a problem for two reasons:
- the error message is not helpful
- if the user wish to store and retrieve this dictionary in a json file, a workaround is required on their part since JSON doesn't know tuples.

**Code for reproduction**

```python
import yt
import json

my_units = {"mass_unit": (1.0, "kg")}
print(my_units)
ds = yt.load("amrvac/bw_2d0000.dat", units_override=my_units) # -> all good

storage = json.dumps(my_units)
my_units = json.loads(storage)
print(my_units)
ds = yt.load("amrvac/bw_2d0000.dat", units_override=my_units) # -> crash
```

**Actual outcome**
```python
{'mass_unit': (1.0, 'kg')}
{'mass_unit': [1.0, 'kg']}
 RuntimeError: unyt_quantity values must be numeric
```

**Expected outcome**
The alternative schema should be valid as long as the list has size 2.

neutrinoceros:
## PR Summary

This improves error handling for `Dataset(units_override=data)` when ill formatted data is passed by checking and crashing early.
This adds flexibility (fixes #2702) and improves the sanitizing step by checking that overrides are dimensionally consistent, and checks for unrecognised keys as well (previously those where silently passed and had no effect).

**old VS new behaviour** against spurious parameter values

```python
import yt
fn = "IsolatedGalaxy/galaxy0030/galaxy0030"

# inconsistent unit/dim
yt.load(fn, units_override={"length_unit": (1, "K")})
old >> IllDefinedUnitSystem: Cannot create unit system with inconsistent mapping from dimensions to units. Received:
OrderedDict([((length), code_length), ((mass), code_mass), ((time), code_time), ((temperature), code_temperature), ((angle), rad), ((current_mks), None), ((luminous_intensity), cd), ((logarithmic), Np)])
# the above error is raised from unyt and it may be hard to understand what is wrong from the user perspective
new >> ValueError: Inconsistent dimensionality in units_override. Received length_unit = 1 K

# typos or unknown "unit" keys:
yt.load(fn, units_override={"len_unit": (1, "cm")})
old >> # loads silently
new >> ValueError: units_override contains invalid keys: {'len_unit'}

# flexible typing input (this was reported in #2702 and fixed here)
yt.load(fn, units_override={"length_unit": [1, "cm"]}) 
old >> RuntimeError: unyt_quantity values must be numeric
new >> # loads correctly

# invalid "0" value
yt.load(fn, units_override={"length_unit": 0})
old >>
.../unyt/unyt/array.py:1753: RuntimeWarning: divide by zero encountered in true_divide
  out_arr = func(
.../unyt/unyt/array.py:1753: RuntimeWarning: invalid value encountered in multiply
  out_arr = func(
# the above still loads, only RuntimeWarnings are thrown
new >> ValueError: Invalid 0 normalisation factor in units_override for length_unit.
```

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -10,6 +10,7 @@ from collections import defaultdict
 from stat import ST_CTIME
 
 import numpy as np
+from unyt.exceptions import UnitConversionError, UnitParseError
 
 from yt.config import ytcfg
 from yt.data_objects.data_containers import data_object_registry
@@ -222,9 +223,7 @@ class Dataset(metaclass=RegisteredDataset):
         self.known_filters = self.known_filters or {}
         self.particle_unions = self.particle_unions or {}
         self.field_units = self.field_units or {}
-        if units_override is None:
-            units_override = {}
-        self.units_override = units_override
+        self.units_override = Dataset._sanitize_units_override(units_override)
 
         # path stuff
         self.parameter_filename = str(filename)
@@ -1081,7 +1080,7 @@ class Dataset(metaclass=RegisteredDataset):
         self.unit_registry.unit_system = self.unit_system
 
     def _create_unit_registry(self, unit_system):
-        import yt.units.dimensions as dimensions
+        from yt.units import dimensions as dimensions
 
         # yt assumes a CGS unit system by default (for back compat reasons).
         # Since unyt is MKS by default we specify the MKS values of the base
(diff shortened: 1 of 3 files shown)


neutrinoceros:
switching to draft because I thought of a better solution but don't have time to implement it right now

neutrinoceros:
Last Jenkins build went really bad, I must be doing something wrong with the sanitising. I'm guessing this line is the culprit

```python
                    uo[key] = (float(val), cgs)
```
maybe I'm not anticipating what could go wrong with the float conversion. I'm going to try to avoid the conversion, see what happens.

yt/data_objects/static_output.py
```diff
@@ -1089,23 +1090,72 @@ def set_code_units(self):
...
+        invalid_keys_found = set(units_override.keys()) - valid_keys
+        if invalid_keys_found:
+            raise ValueError("units_override contains invalid keys: {}".format(invalid_keys_found))
+        for key, cgs in default_cgs.items():
+            try:
+                val = units_override[key]
+            except KeyError:
+                continue
+            try:
+                uo[key] = YTQuantity(val) #(val.v, str(val.units))
```
neutrinoceros, line 1117: I'll clean this up if it proves useless after all tests pass

neutrinoceros:
Switching to draft again after I found some extremely puzzling bugs with the current state

neutrinoceros:
sigh... switching to draft yet again after it became clear that this exposes a bug in AMRVAC's frontend. The errors where correctly exposed by tests run on Jenkins but I ignored those failures since we know there are currently unsolved errors there.

neutrinoceros:
This became a branching mess so I'm closing here and will reopen with a new branch that fixes the failures seen here.

PR #2705, opened 2020-06-29
Bump unyt to 2.7.2

matthewturk:
We had been installing unyt from git, but this bumps the requirement to
the newly-released 2.7.2.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,7 +32,7 @@ install:
     # Install specified version of numpy and dependencies
     - "conda install --yes -c conda-forge numpy scipy nose pytest setuptools ipython git
     Cython sympy fastcache h5py matplotlib=3.1.3 mock pandas cartopy conda-build pooch pyyaml"
-    - "pip install git+https://github.com/yt-project/unyt@de443dff7671f1e68557306d77582cd117cc94f8#egg=unyt"
+    - "pip install unyt"
     # install yt
     - "pip install -e ."
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,9 +30,8 @@ install:
     - "python --version"
 
     # Install specified version of numpy and dependencies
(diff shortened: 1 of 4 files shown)


appveyor.yml
```diff
@@ -32,7 +32,7 @@ install:
     # Install specified version of numpy and dependencies
     - "conda install --yes -c conda-forge numpy scipy nose pytest setuptools ipython git
     Cython sympy fastcache h5py matplotlib=3.1.3 mock pandas cartopy conda-build pooch pyyaml"
-    - "pip install git+https://github.com/yt-project/unyt@de443dff7671f1e68557306d77582cd117cc94f8#egg=unyt"
+    - "pip install unyt"
```
munkm, line 35: is there a reason to leave unyt as pip installed and not with the other conda installs in the lines above? I assume we used pip to get the specific hash when installing from source.

setup.py
```diff
@@ -437,7 +437,7 @@ def run(self):
             'sympy>=1.2',
             'numpy>=1.10.4',
             'IPython>=1.0',
-            'unyt>=2.2.2',
+            'unyt>=2.7.2',
```
munkm, line 440: Are our unyt fixes not backwards compatible with previous unyt versions? They have to install this version of unyt for yt to work?

matthewturk: I would say that since our answer tests require this version, we should as well -- the main thing I want to make sure of is the `mu` symbol being correct.

appveyor.yml
matthewturk: We probably did.  I'll go back to conda.

setup.py
munkm: Yes, I think it's reasonable that we at least match the versions we test in test_minimal_requirements. Hopefully as time moves on the versions between test_requirements and test_minimal_requiements start to differ.

PR #2706, opened 2020-06-30
[AMRVAC] units override check (follow up #2703)

neutrinoceros:
## PR Summary
This is a follow up to #2703 (only the last commit is unique to this PR)
Given that the AMRVAC frontend already had a mechanism to check `units_override`, this unifies the existing mechanism with the sanitizing function introduced by the previous PR.

Edit:  #2703 helped revealing a bug in AMRVAC's frontend where the special key "number_density" in units_override is not handled anywhere. Passing it now raises :
```python
ValueError: units_override contains invalid keys: {'numberdensity_unit'}
```
Thanks to @n-claes , we already had some tests for sanity checks in units_override, but now they fail (which is expected).

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -10,6 +10,7 @@ from collections import defaultdict
 from stat import ST_CTIME
 
 import numpy as np
+from unyt.exceptions import UnitConversionError, UnitParseError
 
 from yt.config import ytcfg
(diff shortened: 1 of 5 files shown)


neutrinoceros:
I'm closing this because the branch name doesn't reflect correctly what's been done, I'll reopen with a clearer one.

PR #2688, opened 2020-06-25
[BUGFIX] Art: fix dm_only frontend

cphyc:
## PR Summary

Removes tests that are intended for particle datasets. They were incorrectly run on ART, an octree dataset.

PR #2712, opened 2020-07-01
Make sure we include IDs as a particle type in Gadget FOF

matthewturk:
This adds the IDs particle type to the list of particle types to check
in the Gadget FOF frontend.  It should eliminate the error seen in the
test suite on Fido.

### DIFF
--- a/yt/frontends/gadget_fof/data_structures.py
+++ b/yt/frontends/gadget_fof/data_structures.py
@@ -90,6 +90,20 @@ class GadgetFOFParticleIndex(ParticleIndex):
         ds.field_units.update(units)
         ds.particle_types_raw = ds.particle_types
 
+    def _setup_filenames(self):
+        if not hasattr(self, "data_files"):
+            template = self.ds.filename_template
+            ndoms = self.ds.file_count
+            cls = self.ds._file_class
+            self.data_files = [
+                cls(self.ds, self.io, template % {'num':i}, i, frange=None)
+                for i in range(ndoms)
+            ]
+        if not hasattr(self, "total_particles"):
+            self.total_particles = sum(
+                sum(d.total_particles.values()) for d in self.data_files
+            )
+
     def _setup_data_io(self):
         super(GadgetFOFParticleIndex, self)._setup_data_io()
         self._setup_filenames()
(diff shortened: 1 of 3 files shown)


brittonsmith:
The particle ids are not meant to be queries from standard data containers as they don't usually come with positions. I don't think the `IDs` field should be detected in the field list for the main dataset. They are accessed through a separate field list. What is the error that you're seeing?

matthewturk:
@brittonsmith What we were seeing was:

 * `member_ids` never got added because ...
 * `data_files` only had one member because ...
 * `total_particles` was uniformly zero even in files that had IDs

So we were never able to access the `nbody, member_ids` field that we check for in the tests, because yt didn't know about the other data files.

Xarthisius:
> What is the error that you're seeing?

```python
import yt

ds = yt.load("gadget_halos/data/groups_056/fof_subhalo_tab_056.0.hdf5")
halo = ds.halo("Group", 0)
print(halo["member_ids"])
```

yields

```bash
  File "pr2712.py", line 5, in <module>
    print(halo["member_ids"])
  File "yt/data_objects/data_containers.py", line 249, in __getitem__
    f = self._determine_fields([key])[0]
  File "yt/data_objects/data_containers.py", line 1390, in _determine_fields
    finfo = self.ds._get_field_info("unknown", fname)
  File "yt/data_objects/static_output.py", line 772, in _get_field_info
yt.utilities.exceptions.YTFieldNotFound: Could not find field '('nbody', 'member_ids')' in fof_subhalo_tab_056.
```

Xarthisius:
@matthewturk I should note that for the test case above, this PR yields and empty array:
```
[] dimensionless
```

Xarthisius:
I think the way to go is to override `ParticleIndex._setup_filenames` and restore it to a state that was before yt-4 merge:

```diff
diff --git a/yt/frontends/gadget_fof/data_structures.py b/yt/frontends/gadget_fof/data_structures.py
index 26c3ff4bb..171a605a1 100644
--- a/yt/frontends/gadget_fof/data_structures.py
+++ b/yt/frontends/gadget_fof/data_structures.py
@@ -90,6 +90,21 @@ class GadgetFOFParticleIndex(ParticleIndex):
         ds.field_units.update(units)
         ds.particle_types_raw = ds.particle_types
 
+    def _setup_filenames(self):
+        if self.real_ds._instantiated_index is None:
+            template = self.real_ds.filename_template
+            ndoms = self.real_ds.file_count
+            cls = self.real_ds._file_class
+            self.data_files = [
+                cls(self.dataset, self.io, template % {'num':i}, i, frange=None)
+                for i in range(ndoms)
+            ]
+        else:
+            self.data_files = self.real_ds.index.data_files
+        self.total_particles = sum(
+            sum(d.total_particles.values()) for d in self.data_files
+        )
+
     def _setup_data_io(self):
         super(GadgetFOFParticleIndex, self)._setup_data_io()
         self._setup_filenames()
```

Xarthisius:
Ultimately it turned out that GadgetFOF frontend was doing the right thing all along, it just wasn't updated to reflect changes that happened in parent class and was overriding wrong methods.

cphyc:
A similar issue happened with Art (#2688) where we had to override the default implementation of `_setup_filenames`. Is it possible the latter is not defined where it should?

matthewturk:
This is mostly @Xarthisius 's PR now, and I think with @brittonsmith 's approval it's good to go in.  Merging.

PR #2714, opened 2020-07-02
Ramses max level

cphyc:
## PR Summary

Add the possibility to only load a RAMSES dataset down to a specified level.

## PR Checklist

- [x] Code passes isort checker
- [x] Code passes black checker

## Demo
info_00080_Slice_x_density max_level=6
info_00080_Slice_x_density max_level=7
info_00080_Slice_x_density max_level=8
info_00080_Slice_x_density max_level=9
info_00080_Slice_x_density max_level=10


yt/frontends/ramses/data_structures.py
```diff
@@ -723,7 +731,9 @@ def read_rhs(f, cast):
             self.omega_lambda = rheader["omega_l"]
             self.omega_matter = rheader["omega_m"]
             self.hubble_constant = rheader["H0"] / 100.0  # This is H100
-        self.max_level = rheader["levelmax"] - self.min_level - 1
+        self.max_level = (
+            min(self._force_max_level, rheader["levelmax"]) - self.min_level - 1
```
neutrinoceros, line 735: If I'm reading this right, `self._force_max_level` is meant to use the save number convention as RAMSES, not yt ?
I'm inclined to think that yt's convention should be followed uniformly on the user side and conversions (+/-1) should only happen under the surface as implementation details. Whichever convention you decide to settle for, it would be good to explicitly document it, at _least_ as an inline comment.

yt/frontends/ramses/particle_handlers.py
```diff
@@ -351,7 +351,7 @@ def read_header(self):
             fields = list(self.known_fields)
 
         for i in range(self.ds.dimensionality * 2 + 1):
-            for j in range(self.ds.max_level, self.ds.min_level):
+            for j in range(self.ds.min_level, self.ds.max_level):
```
neutrinoceros, line 354: weird. This is actually an unrelated bugfix, right ?

cphyc: It is, but it seems it was never tested...

yt/frontends/ramses/data_structures.py
cphyc: You're right. The way it is implemented here is that `self._force_max_level` uses RAMSES convention, which means that level=1 is the base grid (1x1x1) and you can go up to `levelmax`.
The yt convention starts numbering at the finest level where the grid is "full".

As a RAMSES user I'd rather use RAMSES convention because that's what everybody would expect, but I would be happy with other suggestions as well.

Maybe one last solution would be to provide the ability to feed `_force_max_level` as a tuple, like `(force_max_level, "yt")` `(force_max_level, "ramses")`. Either we make the tuple notation explicit, or we add a warning when using the non-tuple format saying whichever convention was picked. Wdyt?

neutrinoceros: That would work. I've been trying to think of a better way but I couldn't.

yt/frontends/ramses/data_structures.py
```diff
@@ -1,6 +1,7 @@
 import os
 import weakref
 from collections import defaultdict
+from collections.abc import Iterable
```
cphyc, line 4:
```suggestion
```

yt/frontends/ramses/data_structures.py
```diff
@@ -590,6 +608,39 @@ def __init__(
 
         self.storage_filename = storage_filename
 
+    def _sanitize_max_level(self, max_level, max_level_convention):
+        # NOTE: the initialisation of the dataset class requires
+        #       sets self.min_level _and_ requires force_max_level
```
neutrinoceros:
```suggestion
        #       self.min_level _and_ requires force_max_level
```

cphyc:
This is good to go!

neutrinoceros:
I'd happily push the button but I think we need another review since this isn't a bugfix.

yt/frontends/ramses/data_structures.py
```diff
@@ -590,6 +608,40 @@ def __init__(
...
         self.storage_filename = storage_filename
 
+    @staticmethod
+    def _sanitize_max_level(max_level, max_level_convention):
+        # NOTE: the initialisation of the dataset class sets
+        #       self.min_level _and_ requires force_max_level
+        #       to be set, so we cannot convert from to yt/ramses
+        #       conventions
+        if max_level is None and max_level_convention is None:
+            return (999, "yt")
```
matthewturk, line 618: `2**999` is a nice even number.

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -221,9 +222,9 @@ def test_ramses_sink():
         "particle_prop_0_1",
         "particle_prop_0_2",
         "particle_prop_0_3",
-        "particle_prop_1_0",
-        "particle_prop_1_1",
-        "particle_prop_1_2",
+        "particle_prop_0_4",
```
matthewturk, line 225: This seems like a bigger change?

cphyc: Yes ; the previous behavior was inconsistent with RAMSES internals:
```fortran
     ! Write sink_stat
     do idim=1,ndim*2+1
        do ilevel=levelmin,nlevelmax
           do i=1,nsink
              xdp(i)=sink_stat(i,ilevel,idim)
           end do
           write(ilun) xdp
        enddo
     enddo
```
so the `sink_stats` which are named "`particle_prop_ilvl_idim`" in yt should have `idim=0..2*ndim` and `ilvl=0..levelmax` (assuming arrays are indexed from 0 instead of 1).

If you're unhappy with this bugfix being included in this PR, I can open another (bugfix) one.

matthewturk: Nope, not unhappy.

matthewturk:
LGTM; needs one more reviewer I think.

matthewturk:
Oops, we already have two reviews.  Merging.

PR #2715, opened 2020-07-02
fix: log error instead of crashing in smoothed grid computation if ghost-zones can't be reconstructed

### LINKED ISSUE #2710: Issue plotting vorticity from amrvac data
### Bug report

**Bug summary**

Attempting to plot_2d the vorticity_magnitude derived field within yt will return an error when applied to non-periodic amrvac data. The traditional hack of manually setting ds.periodicity=[True]*3 is insufficient for successfully passing the command to plot as it gets hung up, then, on the unavailability of ghost cells within the dataset.

**Code for reproduction**

curldrop file--> http://use.yt/upload/5e14fb46

```python
file_index=500
overrides = dict(length_unit=(1e8, 'cm'), temperature_unit=(1e6, 'K'), numberdensity_unit=(1e8, 'cm**-3'))
ds=yt.load('jk2020{0:04d}'.format(file_index)+'.dat', units_override=overrides)
plot=yt.plot_2d(ds, "vorticity_magnitude")
plot.show()
```

**Actual outcome**

```
NeedsGridType                             Traceback (most recent call last)
~/codes/yt/yt/data_objects/data_containers.py in _generate_fluid_field(self, field)
    308         try:
--> 309             finfo.check_available(gen_obj)
    310         except NeedsGridType as ngt_exception:

~/codes/yt/yt/fields/derived_field.py in check_available(self, data)
    210         for validator in self.validators:
--> 211             validator(data)
    212         # If we don't get an exception, we're good to go

~/codes/yt/yt/fields/derived_field.py in __call__(self, data)
    487         if not getattr(data, '_spatial', False):
--> 488             raise NeedsGridType(self.ghost_zones,self.fields)
    489         if self.ghost_zones <= data._num_ghost_zones:

NeedsGridType: (1, [('gas', 'velocity_x'), ('gas', 'velocity_y'), ('gas', 'velocity_z')])

RuntimeError                              Traceback (most recent call last)
<ipython-input-26-c39a34fa7c89> in <module>
      3 ds=yt.load('/Volumes/TRANSCEND/Projects/amrvac/2020_prom_form/spex_dm_4_24_528_4224_3G/jk2020{0:04d}'.format(file_index)+'.dat', units_override=overrides)
      4 #sub=ds.region(left_edge=[-11.9,0.1,0], right_edge=[11.9,24.9,0],center=)
----> 5 plot=yt.plot_2d(ds, "vorticity_magnitude")
      6 plot.show()

~/codes/yt/yt/visualization/plot_window.py in plot_2d(ds, fields, center, width, axes_unit, origin, fontsize, field_parameters, window_size, aspect, data_source)
   2174                                 field_parameters=field_parameters,
   2175                                 window_size=window_size, aspect=aspect,
-> 2176                                 data_source=data_source)

~/codes/yt/yt/visualization/plot_window.py in __init__(self, ds, axis, fields, center, width, axes_unit, origin, right_handed, fontsize, field_parameters, window_size, aspect, data_source, buff_size)
   1331             slc = ds.slice(axis, center[axis], field_parameters=field_parameters,
   1332                            center=center, data_source=data_source)
-> 1333             slc.get_data(fields)
   1334         validate_mesh_fields(slc, fields)
   1335         PWViewerMPL.__init__(self, slc, bounds, origin=origin,

~/codes/yt/yt/data_objects/data_containers.py in get_data(self, fields)
   1617 
   1618         fields_to_generate += gen_fluids + gen_particles
-> 1619         self._generate_fields(fields_to_generate)
   1620         for field in list(self.field_data.keys()):
   1621             if field not in ofields:

~/codes/yt/yt/data_objects/data_containers.py in _generate_fields(self, fields_to_generate)
   1637                 fi = self.ds._get_field_info(*field)
   1638                 try:
-> 1639                     fd = self._generate_field(field)
   1640                     if hasattr(fd, 'units'):
   1641                         fd.units.registry = self.ds.unit_registry

~/codes/yt/yt/data_objects/data_containers.py in _generate_field(self, field)
    291                 tr = self._generate_particle_field(field)
    292             else:
--> 293                 tr = self._generate_fluid_field(field)
    294             if tr is None:
    295                 raise YTCouldNotGenerateField(field, self.ds)

~/codes/yt/yt/data_objects/data_containers.py in _generate_fluid_field(self, field)
    309             finfo.check_available(gen_obj)
    310         except NeedsGridType as ngt_exception:
--> 311             rv = self._generate_spatial_fluid(field, ngt_exception.ghost_zones)
    312         else:
    313             rv = finfo(gen_obj)

~/codes/yt/yt/data_objects/data_containers.py in _generate_spatial_fluid(self, field, ngz)
    358                         ind += wogz.select(
    359                             self.selector,
--> 360                             gz[field][ngz:-ngz, ngz:-ngz, ngz:-ngz],
    361                             rv, ind)
    362         if accumulate:

~/codes/yt/yt/data_objects/data_containers.py in __getitem__(self, key)
    254                 return self.field_data[f]
    255             else:
--> 256                 self.get_data(f)
    257         # fi.units is the unit expression string. We depend on the registry
    258         # hanging off the dataset to define this unit object.

~/codes/yt/yt/data_objects/construction_data_containers.py in get_data(self, fields)
    727                 self._fill_particles(part)
    728 
--> 729         if len(fill) > 0: self._fill_fields(fill)
    730         for a, f in sorted(alias.items()):
    731             if f.sampling_type == 'particle' and not is_sph_field:

~/codes/yt/yt/data_objects/construction_data_containers.py in _fill_fields(self, fields)
   1186         if len(fields) == 0: return
   1187         ls = self._initialize_level_state(fields)
-> 1188         min_level = self._compute_minimum_level()
   1189         # NOTE: This usage of "refine_by" is actually *okay*, because it's
   1190         # being used with respect to iref, which is *already* scaled!

~/codes/yt/yt/data_objects/construction_data_containers.py in _compute_minimum_level(self)
   1172             ils.data_source.loose_selection = False
   1173             min_level = self.level
-> 1174             for chunk in ils.data_source.chunks([], "io"):
   1175                 # With our odd selection methods, we can sometimes get no-sized ires.
   1176                 ir = chunk.ires

~/codes/yt/yt/data_objects/data_containers.py in chunks(self, fields, chunking_style, **kwargs)
   1497     def chunks(self, fields, chunking_style, **kwargs):
   1498         # This is an iterator that will yield the necessary chunks.
-> 1499         self.get_data() # Ensure we have built ourselves
   1500         if fields is None: fields = []
   1501         # chunk_ind can be supplied in the keyword arguments.  If it's a

~/codes/yt/yt/data_objects/data_containers.py in get_data(self, fields)
   1540     def get_data(self, fields=None):
   1541         if self._current_chunk is None:
-> 1542             self.index._identify_base_chunk(self)
   1543         if fields is None: return
   1544         nfields = []

~/codes/yt/yt/geometry/grid_geometry_handler.py in _identify_base_chunk(self, dobj)
    272             dobj._chunk_info[0] = weakref.proxy(dobj)
    273         elif getattr(dobj, "_grids", None) is None:
--> 274             gi = dobj.selector.select_grids(self.grid_left_edge,
    275                                             self.grid_right_edge,
    276                                             self.grid_levels)

~/codes/yt/yt/data_objects/data_containers.py in selector(self)
   1492                 self, self._data_source.selector, sclass(self))
   1493         else:
-> 1494             self._selector = sclass(self)
   1495         return self._selector
   1496 

~/codes/yt/yt/geometry/selection_routines.pyx in yt.geometry.selection_routines.RegionSelector.__init__()

RuntimeError: Error: yt attempted to read outside the boundaries of a non-periodic domain along dimension 0.
Region left edge = -12.017045454545455 code_length, Region right edge = -11.84659090909091 code_length
Dataset left edge = -12.0 code_length, Dataset right edge = 12.0 code_length

This commonly happens when trying to compute ghost cells up to the domain boundary. Two possible solutions are to load a smaller region that does not border the edge or override the periodicity for this dataset.
```

**Expected outcome**

A plot of the vorticity across the whole field-of-view

**Version Information**
  * Operating System: MacOS 10.14.6
  * Python Version: 3.7.6
  * yt version: 4.0dev (edited by @neutrinoceros)

python was installed using conda
yt was installed using 'git clone [...]', cd <yt folder location>, 'pip install -e .'

neutrinoceros:
## PR Summary

A semi-fix to #2710 
The core issue in #2710 is that we don't have ghost-zone reconstruction in the AMRVAC frontend.
With this change, instead of crashing, yt logs an error message pointing to the defective class for lacking ghost-on support.
Results are obviously buggy (see below) but it's still better than nothing and at least the user gets a better idea what's wrong.

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -1,6 +1,7 @@
 import fileinput
 import io
 import os
+import warnings
 import zipfile
 from functools import wraps
 from re import finditer
@@ -1348,6 +1349,8 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
         if not iterable(self.ds.refine_by):
             refine_by = [refine_by, refine_by, refine_by]
         refine_by = np.array(refine_by, dtype="i8")
+
+        runtime_errors_count = 0
         for level in range(self.level + 1):
             if level < min_level:
                 self._update_level_state(ls)
@@ -1373,8 +1376,17 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
                     refine_by,
                 )
             if level == 0 and tot != 0:
-                raise RuntimeError
+                runtime_errors_count += 1
             self._update_level_state(ls)
+        if runtime_errors_count:
+            warnings.warn(
+                "Something went wrong during field computation. "
+                "This is likely due to missing ghost-zones support "
+                "in class %s",
+                self.ds.__class__,
+                category=RuntimeWarning,
+            )
+            mylog.debug(f"Caught {runtime_errors_count} runtime errors.")
         for name, v in zip(fields, ls.fields):
             if self.level > 0:
                 v = v[1:-1, 1:-1, 1:-1]
--- a/yt/fields/field_exceptions.py
+++ b/yt/fields/field_exceptions.py
@@ -8,7 +8,8 @@ class NeedsGridType(ValidationException):
         self.fields = fields
 
     def __str__(self):
-        return f"({self.ghost_zones}, {self.fields})"
+        s = "s" if self.ghost_zones != 1 else ""
+        return f"fields {self.fields} require {self.ghost_zones} ghost zone{s}."
 
 
 class NeedsOriginalGrid(NeedsGridType):
--- a/yt/geometry/selection_routines.pyx
+++ b/yt/geometry/selection_routines.pyx
@@ -999,14 +999,16 @@ cdef class RegionSelector(SelectorObject):
             else:
                 if LE[i] < DLE[i] or RE[i] > DRE[i]:
                     raise RuntimeError(
-                        "Error: yt attempted to read outside the boundaries of "
+                        "yt attempted to read outside the boundaries of "
                         "a non-periodic domain along dimension %s.\n"
                         "Region left edge = %s, Region right edge = %s\n"
                         "Dataset left edge = %s, Dataset right edge = %s\n\n"
                         "This commonly happens when trying to compute ghost cells "
                         "up to the domain boundary. Two possible solutions are to "
-                        "load a smaller region that does not border the edge or "
-                        "override the periodicity for this dataset." % \
+                        "select a smaller region that does not border domain edge "
+                        "(see https://yt-project.org/docs/analyzing/objects.html?highlight=region)\n"
+                        "or override the periodicity with e.g\n"
+                        "ds.periodicity = 3*[True]" % \
                         (i, dobj.left_edge[i], dobj.right_edge[i],
                          dobj.ds.domain_left_edge[i], dobj.ds.domain_right_edge[i])
                     )


neutrinoceros:
Note that some plot can be completed by simply removing the lines that raise the RuntimeError

Screenshot 2020-07-02 at 14 23 03

obviously it's completely broken. I'm now trying to flesh out the error message so at least some useful insight reaches the user.

neutrinoceros:
So for now I would be inclined to think that the machinery is there, but the frontend lacks support for these fields. I'm now trying to catch the problem earlier and raise a single, informative error, see if tests pass.

neutrinoceros:
Alright, next try: I'll just replace the runtime error with a runtime warning since it can be argued than getting a partly broken result is still better than no result + cryptic error.

neutrinoceros:
Alright this seems to not break any existing test. I think this is the best fix I can provide for now without spending much more time implementing ghost zones support in the frontend.

@matthewturk , would you be able to improve on my runtime warning error message ? It's not really accurate as it just says that "something went wrong", but I don't really know what. In particular, I do not know if this error is systematically related to missing ghost zones, maybe that should be mentionned.
Ideally the warning should state that the error is frontend specific, if that's the case.

neutrinoceros:
I have updated the logged error message and changed the PR's summary and title.

neutrinoceros:
note that I've tried to use a uniform grid to plot vorticity directly with matplotlib and it seems to be broken as well:
```python
import yt
ds = yt.load("amrvac/jk20200500.dat")
ds.periodicity = 3*[True]

level = ds.min_level
dims = np.ones(3)
dims[: ds.dimensionality] = ds.parameters["domain_nx"] * 2 ** level

grid = ds.covering_grid(level=level, left_edge=ds.domain_left_edge, dims=dims, num_ghost_zones=1)
plt.contourf(grid["vorticity_magnitude"].squeeze())
```

Screenshot 2020-07-14 at 15 11 53

neutrinoceros:
Since I realised that `ds.covering_grid` works correctly with AMRVAC data (for fields that don't require spatial derivatives) and doesn't complain about missing ghost zones I'm now wondering why vorticity is computed through a `YTSmoothedCoveringGrid` object... I'll have another look.

neutrinoceros:
Good news ! turns out forcing the internal machinery to use a simple covering_grid (not smoothed) gives better results 

jake2_Slice_z_vorticity_z

So now it's mostly a matter of finding how to make that part of the code smart enough that it chooses wisely when to use smoothed grids, depending on level of support for ghost zones.
jake2_grid_Slice_z_vorticity_z

neutrinoceros:
@yt-fido test this please

matthewturk:
I'm kind of torn about this since it seems likely to result in people getting bad results and not seeing the warnings.  But I will sign off.

neutrinoceros:
Well I don't know how one could deactivate `warnings.warning`, I'm using it intentionally here to make it harder to skip it ! (also this has the nice benefit that the warning only gets printed once if multiple cpus hit it)

yt/data_objects/construction_data_containers.py
```diff
@@ -1376,8 +1379,16 @@ def _fill_fields(self, fields):
...
                 )
             if level == 0 and tot != 0:
-                raise RuntimeError
+                runtime_errors_count += 1
             self._update_level_state(ls)
+        if runtime_errors_count:
+            warnings.warn(
+                "Something went wrong during field computation. "
+                "This is likely due to missing ghost-zones support "
+                "in class %s" % self.ds.__class__,
```
munkm, line 1388: Should the formatting here be similar to what was done in yt-project/yt#2820?

neutrinoceros: and justice for all 🤘 . That being said, since we're actively decluttering the queue on Appveyor right now, I'll wait until flynt PRs are merged to fix this. Thanks for catching it !

PR #2719, opened 2020-07-03
Hotfix broken raise statement

neutrinoceros:
## PR Summary

fix a broken raise statement (missing positional argument in `YTFieldNotFound` init).
While I'm at it, I'm changing the pattern to EAFP.

edit:
Wait. This error is actually encapsulating another one that's pretty much identical

```python
KeyError: 'No field named blblblb'

YTFieldNotFound 
YTFieldNotFound: Could not find field 'blblblb' in MYFILE.
```

So I think the best course of action is simply to get rid of this redundant error-chaining.

But there's worse: I'm getting this while I'm trying to *add* a field to a dataset, so none of these errors makes much sense to me.
-> this is a different problem and is actually fixed with #2638

### DIFF
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -428,8 +428,6 @@ class FieldInfoContainer(dict):
         unavailable = []
         fields_to_check = fields_to_check or list(self.keys())
         for field in fields_to_check:
-            if field not in self:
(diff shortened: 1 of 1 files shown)


matthewturk:
So I am trying to recall if there are places we catch `YTFieldNotFound` as a signal, but I can't find them.  So this looks good to me.

PR #2722, opened 2020-07-03
fix: add support to init from pathlike objects for time series

### LINKED ISSUE #2633: support pathlike arguments for user-specificed files/directories
### Bug report

So in https://github.com/yt-project/yt/pull/2405 I've added support for loading datafiles specified as `pathlib.Path` objects. I'm realizing the global api would feel more consistent if that was also supported by other functions accepting files/directories as arguments.

I can currently think of 2 places where this change would apply, `yt.utilities_load_sample()` and `yt.funcs.enable_plugins()`.

Please report any other similar api parts that would benefit from this change.

(In the spirit of the dev meeting we had yesterday and the recent email distributed to users, I took the liberty of creating an "api-consistency" label)

neutrinoceros:
#  PR Summary

fix #2633 

This is a very light change, but the branch is based on top of #2695 (merged ! 🎉 )
Reason: in order to avoid type checking with `isinstance(output, str)` and allow allow pathlike types to be treated equally, I use this pattern to determine if an output represents loaded or loadable data
```python
try:
    return load(output) # loadable data ?
except TypeError:
    return output # assumed loaded data
```
... and this pattern is only relevant if `load` is able to throw `TypeError`. To a certain extent, this proves the point I made in #2695 : it makes for a more maintainable code base.
Based on master, I would need to catch `YTOutputNotIdentified`, which would not be specific enough and could cause very weird behaviour downstream.

## PR Checklist

### DIFF
--- a/yt/data_objects/tests/test_time_series.py
+++ b/yt/data_objects/tests/test_time_series.py
@@ -2,8 +2,9 @@ import os
 import tempfile
 from pathlib import Path
 
-from yt.data_objects.time_series import get_filenames_from_glob_pattern
+from yt.data_objects.time_series import DatasetSeries, get_filenames_from_glob_pattern
 from yt.testing import assert_raises
+from yt.utilities.exceptions import YTOutputNotIdentified
 
 
 def test_pattern_expansion():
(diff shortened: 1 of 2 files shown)


neutrinoceros:
switching to draft to clear the PR backlog a bit

neutrinoceros:
No idea if the docs failure is real...
@yt-fido test this please

cphyc (approved):
LGTM!

yt/data_objects/tests/test_time_series.py
```diff
@@ -14,14 +15,50 @@ def test_pattern_expansion():
...
+            FileNotFoundError, DatasetSeries._get_filenames_from_glob_pattern, pattern
+        )
+
+
+def test_init_fake_dataseries():
+
+    file_list = [f"fake_data_file_{str(i).zfill(4)}" for i in range(10)]
+    with tempfile.TemporaryDirectory() as tmpdir:
+        pfile_list = [Path(tmpdir) / file for file in file_list]
+        sfile_list = [os.path.join(tmpdir, f) for f in file_list]
```
cphyc, line 38: For the sake of consistency
```suggestion
        tmpdir_path = Path(tmpdir)
        pfile_list = [tmpdir_path / file for file in file_list]
        sfile_list = [os.path.join(tmpdir, file) for file in file_list]
```

yt/data_objects/time_series.py
```diff
@@ -113,10 +97,11 @@ class DatasetSeries:
 
     Parameters
     ----------
-    outputs : list or pattern
+    outputs : list of filenames, or pattern
         A list of filenames, for instance ["DD0001/DD0001", "DD0002/DD0002"],
         or a glob pattern (i.e. containing wildcards '[]?!*') such as "DD*/DD*.index".
         In the latter case, results are sorted automatically.
+        Filenames and patterns can be of type str, os.Patlike or bytes.
```
cphyc, line 104:
```suggestion
        Filenames and patterns can be of type str, os.Pathlike or bytes.
```

yt/data_objects/tests/test_time_series.py
neutrinoceros: Looks good. might as well skip `os.path` and use only pathlib while we're at it

PR #2724, opened 2020-07-04
add support for region selection with numpy-like ellipsis syntax

neutrinoceros:
## PR Summary

Region selection mostly feels like array slicing with magic, though there's one way to write array slicing that is currently not supported in yt : ellipsis `...` !
This PR adds support for this by expanding an ellipsis into the required number of `slice` instances.

For reference here's yt's current behaviour:
```python
ds = yt.testing.fake_amr_ds()

reg1 = ds.r[...]  # is equivalent to ds.r[:,:,:]
old >> TypeError: object of type 'ellipsis' has no len()

reg2 = ds.r[(1, "cm"), ...]  # is equivalent to ds.r[(1, "cm"),:,:]
old >> YTDimensionalityError: Dimensionality specified was 2 but we need 3

# though this might be considered equivalent to ds.r[:, (1, "cm"), :]
# it's regarded as ambiguous in the general case in numpy so we won't support this
# but throw a relevant error instead, in numpy's fashion
reg3 = ds.r[..., (1, "cm"), ...]  
old >> RuntimeError: unyt_quantity values must be numeric
new >> IndexError: an index can only have a single ellipsis ('...')
```

Note: I'm not currently aware of any other place in the code base where it would be useful, though this pattern should be trivial to isolate as a helper function if needed.

## PR Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,6 @@ combine_as_imports=True
 line_length=88
 # isort can't be applied to yt/__init__.py because it creates circular imports
 skip = venv, doc, benchmarks, yt/__init__.py, yt/extern
-known_third_party = IPython, nose, numpy, sympy, matplotlib, unyt, git, yaml, dateutil, requests, coverage, pytest, pyx
+known_third_party = IPython, nose, numpy, sympy, matplotlib, unyt, git, yaml, dateutil, requests, coverage, pytest, pyx, glue
 known_first_party = yt
 sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -12,7 +12,6 @@ from unyt.exceptions import UnitConversionError, UnitParseError
 import yt.geometry.selection_routines
 from yt.data_objects.field_data import YTFieldData
 from yt.data_objects.profiles import create_profile
-from yt.fields.derived_field import DerivedField
 from yt.fields.field_exceptions import NeedsGridType
 from yt.frontends.ytdata.utilities import save_as_dataset
 from yt.funcs import (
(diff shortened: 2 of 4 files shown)


matthewturk:
I like this, but it looks like some tests are failing.

neutrinoceros:
Right, I wasn't paying attention. I'll have a look !

neutrinoceros:
should be fixed now

neutrinoceros:
I should add that I had to refactor `FieldInfoContainer._determine_fields` to avoid having to catch `YTFieldNotFound` during the region object creation

neutrinoceros:
@yt-fido test this please

jzuhone:
@neutrinoceros could you elaborate a bit more on what was going on with the changes to `FieldInfoContainer`?

neutrinoceros:
> @neutrinoceros could you elaborate a bit more on what was going on with the changes to FieldInfoContainer?

Sure, I'd agree that it is not at all obvious why this was needed !
So the thing is that, in the first few lines of `RegionExpression.__getitem__`, there's this:
```python
        try:
            return self.all_data[item]
        except (TypeError, YTFieldNotParseable):
            pass
```
see
https://github.com/yt-project/yt/blob/9138a0a7330e642e040f8dc0a459f7dba347fa05/yt/data_objects/region_expression.py#L28

and the problem is that without modifications, calling `self.all_data[...]` would lead to the underlying `FieldInfoContainer` instance to raise a `YTFieldNotFound` error. A simple fix would have been to simply catch this error as well but it wouldn't make much sense in the long run, and I wanted to keep a clear separation between `YTFieldNotFound`, which I think should only ever get raised when the input "key" actually makes sense as a field name, and `YTFieldNotParseable`, which seems relevant in this case.

yt/data_objects/data_containers.py
```diff
@@ -1428,52 +1427,72 @@ def _field_type_state(self, ftype, finfo, obj=None):
...
+            # in some naming convention, before returning it as a possible field type
+            if (
+                (ftype, fname) not in self.ds.field_info
+                and (ftype, fname) not in self.ds.field_list
+                and fname not in self.ds.field_list
+                and (ftype, fname) not in self.ds.derived_field_list
+                and fname not in self.ds.derived_field_list
+                and (ftype, fname) not in self._container_fields
+            ):
+                raise YTFieldNotFound((ftype, fname), self.ds)
```
neutrinoceros, line 1495: So here I didn't really try to simplify the logic but it seems weird to me that this check is even a thing. I think that this line
```python
finfo = self.ds._get_field_info(ftype, fname)
```
would fail already if the field isn't present. @matthewturk, I think you authored this check, do you believe it is still relevant after the refactor ?

setup.cfg
```diff
@@ -52,6 +52,6 @@ combine_as_imports=True
 line_length=88
 # isort can't be applied to yt/__init__.py because it creates circular imports
 skip = venv, doc, benchmarks, yt/__init__.py, yt/extern
-known_third_party = IPython, nose, numpy, sympy, matplotlib, unyt, git, yaml, dateutil, requests, coverage, pytest, pyx
+known_third_party = IPython, nose, numpy, sympy, matplotlib, unyt, git, yaml, dateutil, requests, coverage, pytest, pyx, glue
```
munkm, line 55: is this glue addition relevant to this PR?

neutrinoceros: Isort side effect. Maybe with the updated version it's not necessary anymore but I had to add it otherwise the isort hook was unhappy about a line that's actually fine.

munkm: I guess I'm confused why isort would want another package added to this line? these other third party packages are hard dependencies or used for development. glue should be one of our optional dependencies.

neutrinoceros: There's no différenciation between hard and optional deps as far as isort is concerned. It just means "if you see THIS being imported, treat it as third party" and indeed the one line where that's important is not a top level import but buried somewhere in a function. Isort only recently gained the ability to sort even those so it's not too surprising that we're seeing évolutions there :)

PR #2725, opened 2020-07-04
fix: automatic slice->region expansion support for 1D and 2D datasets

neutrinoceros:
## PR Summary

Very simple refactor for `RegionExpression.__getitem__` :
- simplify existing code with try block
- add flexibility for single-slice expansion (previously only expanded to 3D)

```python
ds = yt.load("amrvac/bw_2d0000.dat")
ds.r[1:2]
old >> YTDimensionalityError: Dimensionality specified was 3 but we need 2
new >> YTSlice (bw_2d0000.dat): , axis=2, coord=0.5 cm
```

### DIFF
--- a/yt/data_objects/region_expression.py
+++ b/yt/data_objects/region_expression.py
@@ -41,8 +41,8 @@ class RegionExpression:
                 # possible use case of this would be where we supply something
                 # like ds.r[::256j] .  This would be expanded, implicitly into
                 # ds.r[::256j, ::256j, ::256j].  Other cases would be if we do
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Some image tests seem to be broken by this on Jenkins (wind tunnel), but I don't recall how to reproduce them locally.
I would be inclined to think that those tests probably rely on the current behaviour for slice expansion in 2D, though I don't understand how that wouldn't cause an error.

matthewturk:
The wind tunnel tests may be unrelated.

Issue #2699, opened 2020-06-26
yt.load_sample() has issues loading datasets with null "load_name" in the sample_data_registry


munkm:
### Bug report

**Bug summary**

When going using the load_sample function with yt I am able to download a sample dataest, but if the "load_name" val is null load_sample doesn't work. We should make the behavior more intutitive or supply a default file for affected datasets in the sample_data_registry.json . I've tried this with a few datasets that have null for this keyword and all have the same issue. 

**Code for reproduction**

```python
import yt
ds = yt.load_sample("Orbit")
```

**Actual outcome**
```
yt : [WARNING  ] 2020-06-26 13:20:52,112 tqdm is not installed, progress bar can not be displayed.
yt : [INFO     ] 2020-06-26 13:20:52,140 Files located at /Users/madicken/yt_data/Orbit.tar.gz.untar/Orbit/orbit_hdf5_chk_00
yt : [INFO     ] 2020-06-26 13:20:52,141 Default to loading None for Orbit dataset
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-344ecbc38ac4> in <module>
----> 1 ds = yt.load_sample("Orbit")

~/repos/yt/yt/utilities/load_sample.py in load_sample(name, specific_file, pbar)
     80         mylog.info("Default to loading %s for %s dataset", file_lookup, name)
     81         loaded_file = os.path.join(base_path, "%s.untar" %fileext,
---> 82                                    name, file_lookup)
     83     else:
     84         mylog.info("Loading %s for %s dataset", specific_file, name)

~/python/anaconda/envs/ytdev37/lib/python3.7/posixpath.py in join(a, *p)
     92                 path += sep + b
     93     except (TypeError, AttributeError, BytesWarning):
---> 94         genericpath._check_arg_types('join', a, *p)
     95         raise
     96     return path

~/python/anaconda/envs/ytdev37/lib/python3.7/genericpath.py in _check_arg_types(funcname, *args)
    151         else:
    152             raise TypeError('%s() argument must be str or bytes, not %r' %
--> 153                             (funcname, s.__class__.__name__)) from None
    154     if hasstr and hasbytes:
    155         raise TypeError("Can't mix strings and bytes in path components") from None

TypeError: join() argument must be str or bytes, not 'NoneType'


```

**Expected outcome**

The dataset should load

**Version Information**
  * Python Version: 3.7.3
[...]

PR #2726, opened 2020-07-05
hotfix: support time series loading from load_sample

neutrinoceros:
## PR Summary

This is a quick (and dirty ?) fix for #2699 
It works for the one sample @munkm reported on, but I'm not sure that it generalises well to other data samples with "null" `load_name`.

edit: it already doesn't work with armvac samples (not time series), because they untar into a nested folder, whose outer layer isn't a valid argument for yt.load, so this is definitely not sufficient.

Now I'm not sure that there exists a universal fix for this other than filling the currently undefined `load_name` parameters

### DIFF
--- a/yt/utilities/load_sample.py
+++ b/yt/utilities/load_sample.py
@@ -72,7 +72,7 @@ def load_sample(name=None, specific_file=None, pbar=True):
 
     # Location of the file to load automatically, registered in the Fido class
     info = fido[fileext]
-    file_lookup = info['load_name']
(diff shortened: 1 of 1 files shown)


matthewturk:
This has bitten me in the past and I think this solution looks great.

neutrinoceros:
Arf sorry Matt I should have switched to draft earlier. It’s now clear to me that this fix isn’t general enough so I don’t think it should be merged as is.

neutrinoceros:
This needs a lot more work, I'm closing for now.

PR #2728, opened 2020-07-05
[bugfix] ensure correct axis units when unit_system=code

n-claes:
This (again) fixes the issue that axis units are not displaying `code_length` for datasets loaded with `unit_system='code'` (issue: #2353, PR: #2354). This also extends to `annotate_timestamp` not displaying the units correctly (issue: #2432, PR: #2435). I assume that merging yt 4.0 undid these.

As far as I could see the unit systems no longer have 'us' in their name, but are simply 'cgs' or 'mks'. The code unit system however has the name 'e0ff37ce2df3cfe508167372302fd83d' (at least in my case) and I'm not sure where this comes from...

The fix is easy, but I don't know if this is going to break other stuff. If it's fine I'll add a test so this doesn't happen again.

```python
    ds = yt.load('amrvac/bw_3d0000.dat', unit_system='code')
    p = yt.SlicePlot(ds, 'x', 'density')
    p.annotate_timestamp()
```

## Before
code_units_before

## After
code_units_after

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1108,8 +1108,6 @@ class Dataset(abc.ABC):
         us = create_code_unit_system(
             self.unit_registry, current_mks_unit=current_mks_unit
         )
-        keyname = str(unit_system).lower()
-
         if unit_system != "code":
             us = unit_system_registry[str(unit_system).lower()]
 
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1112,8 +1112,9 @@ class Dataset(abc.ABC):
 
         if unit_system == "code":
             us.name = keyname  # replace an otherwise random label
+
         else:
-            us = unit_system_registry[keyname]
+            us = unit_system_registry[str(unit_system).lower()]
 
         self.unit_system = us
         self.unit_registry.unit_system = self.unit_system
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1108,8 +1108,10 @@ class Dataset(abc.ABC):
         us = create_code_unit_system(
             self.unit_registry, current_mks_unit=current_mks_unit
         )
+        keyname = str(unit_system).lower()
         if unit_system != "code":
-            us = unit_system_registry[str(unit_system).lower()]
+            us = unit_system_registry[keyname]
+        us.name = keyname
         self.unit_system = us
         self.unit_registry.unit_system = self.unit_system
 
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1110,11 +1110,10 @@ class Dataset(abc.ABC):
         )
         keyname = str(unit_system).lower()
 
-        if unit_system == "code":
-            us.name = keyname  # replace an otherwise random label
-
-        else:
+        if unit_system != "code":
             us = unit_system_registry[str(unit_system).lower()]
+
+        us._code_flag = unit_system == "code"
 
         self.unit_system = us
         self.unit_registry.unit_system = self.unit_system
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1109,9 +1109,12 @@ class Dataset(abc.ABC):
             self.unit_registry, current_mks_unit=current_mks_unit
         )
         keyname = str(unit_system).lower()
-        if unit_system != "code":
+
+        if unit_system == "code":
+            us.name = keyname  # replace an otherwise random label
+        else:
             us = unit_system_registry[keyname]
-        us.name = keyname
+
         self.unit_system = us
         self.unit_registry.unit_system = self.unit_system
 
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -2471,7 +2471,7 @@ class TimestampCallback(PlotCallback):
         if self.time:
             # If no time_units are set, then identify a best fit time unit
             if self.time_unit is None:
-                if plot.ds.unit_system.name == "code":
+                if plot.ds.unit_system.name.startswith("code"):
                     # if the unit system is in code units
                     # we should not convert to seconds for the plot.
                     self.time_unit = plot.ds.unit_system.base_units[dimensions.time]
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -2471,7 +2471,7 @@ class TimestampCallback(PlotCallback):
         if self.time:
             # If no time_units are set, then identify a best fit time unit
             if self.time_unit is None:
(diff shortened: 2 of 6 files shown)


neutrinoceros:
So we clearly need a test for this. Also, according to our discussion on slack, this fix is a workaround the fact that `ds.unit_system.name` looks like a random string in case we use "code", which in itself could be regarded as a deeper bug, and has a lot of other repercussions.

For instance, using the same amrvac dataset:
```python
fn = 'amrvac/bw_3d0000.dat'
ds_cgs = yt.load(fn)

ds_cgs.r["thermal_pressure"]
>> unyt_array([....]

ds_code = yt.load(fn, unit_system='code')
ds_code.r["thermal_pressure"]
>> YTFieldNotFound
```

So while this quick fix adresses the graphic issue well, it doesn't deal with the root of the problem.

neutrinoceros:
Here's the part where `"code"` is differentiated from other unit systems
https://github.com/yt-project/yt/blob/ddfb25236594da6747ca201952abbaaf8f340f8c/yt/data_objects/static_output.py#L975

yt/data_objects/static_output.py
```diff
@@ -974,8 +974,12 @@ def _assign_unit_system(self, unit_system):
 
         us = create_code_unit_system(
```
neutrinoceros, line 975: apparently this line needs to be executed even when `unit_system != "code"`, or else some tests 
(e.g. `yt/data_objects/tests/test_sph_data_objects.py::test_compare_arbitrary_grid_slice`) will fail. I'm not convinced the current behaviour isn't actually buggy...

n-claes:
Hi @matthewturk, I kinda lost track of this PR to be honest, wasn't aware there were merge conflicts. I'll update my local branches and rebase this one onto master, that ought to do it.. One sec

n-claes:
Rebasing was a huge mess, went over it carefully and I hope nothing was screwed up. Tested it locally and everything seems to work

neutrinoceros:
only failures are stylistic.
Solution:
run 
```bash
black yt/
flynt yt/data_objects/data_containers.py
```
and rebase again (the flake8 error is on the master side but was fixed since)

matthewturk:
I think we can run that here with slash commands?

n-claes:
Style checks are all passing now, should be good to go!

yt/visualization/plot_modifications.py
```diff
@@ -2492,15 +2490,14 @@ def __call__(self, plot):
...
                 un = self.time_unit.latex_representation()
                 time_unit = r"$\ \ (" + un + r")$"
-            else:
-                # the 'smallest_appropriate_unit' function will return a
-                # string, so we shouldn't need to format it further.
-                time_unit = self.time_unit
+            except AttributeError:
+                # this should only happen if self.time_unit is "code_time"
+                # in which case we drop the "_" in the time stamp
+                time_unit = str(self.time_unit).replace("_", " ")
```
neutrinoceros, line 2500: Just to make sure we're not accidentally hitting this in any unforeseen way.
```suggestion
                # in which case we drop the "_" in the time stamp
                assert plot.ds.unit_system._code_flag == "code"
                time_unit = str(self.time_unit).replace("_", " ")
```

neutrinoceros:
I hate to be the bringer of bad news but I just tested this again with the following snippet
```python
import yt

for unit_system in ("cgs", "code"):
    ds = yt.load("amrvac/bw_2d0000.dat", unit_system=unit_system)
    p = yt.plot_2d(ds, "density")
    p.annotate_timestamp()
    p.save(f"/tmp/us_{unit_system}")
```
and I got this:
(cgs)
us_cgs_Slice_phi_density
(code)
us_code_Slice_phi_density

So the zlabel and timestamps behave correctly, but the x and y labels do not and are always presented in "code length" even when supplying `unit_system="cgs"` (I note that this is also the case on master currently).

n-claes:
Opened an issue for this (#2913) since the behaviour just mentioned is apparently not amrvac-frontend-specific. I'll try to figure out what goes wrong and include the fix here since it's probably related to pieces of code modified in this PR.

neutrinoceros:
> I'll try to figure out what goes wrong and include the fix here since it's probably related to pieces of code modified in this PR.

good call. Plus, this isn't labelled as frontend specific either so no problem at all !

neutrinoceros:
@n-claes can you please rebase this please ?

n-claes:
> @n-claes can you please rebase this please ?

Done!

yt/visualization/plot_modifications.py
```diff
@@ -2471,10 +2470,9 @@ def __call__(self, plot):
         if self.time:
             # If no time_units are set, then identify a best fit time unit
             if self.time_unit is None:
-                if plot.ds.unit_system.name.startswith("us"):
-                    # if the unit system name startswith "us", that means it is
-                    # in code units and we should not convert to seconds for
-                    # the plot.
+                if plot.ds.unit_system._code_flag:
+                    # if the unit system is in code units
```
neutrinoceros, line 2474: maybe add another comment that this attribute is added from the Dataset class. It feels like I would expect it to be a normal attribute (defined in unyt) if I wasn't specifically reviewing this bit.

yt/visualization/plot_modifications.py
neutrinoceros: This is still relevant but I know have a better idea

neutrinoceros, line 2500:
```suggestion
            except AttributeError as err:
                if plot.ds.unit_system._code_flag == "code":
                    raise RuntimeError(
                        "The time unit str repr didn't match expectations, something is wrong."
                    ) from err
                time_unit = str(self.time_unit).replace("_", " ")
```

neutrinoceros (commented):
Ok I agree with past-me about the one suggestion I made in September, I even made it a little better. Otherwise this is good to go I think, thank you @n-claes !

munkm:
@n-claes do you want to implement the suggestion by @neutrinoceros on this? Or do you have other thoughts about it?

n-claes:
Oh for sure, I've just committed @neutrinoceros's suggested change. Been quite busy with other projects recently and haven't spent as much time on yt as I would have liked.. Thanks for following up on this, should be good to go now!

neutrinoceros:
Excellent, thank you Niels !

neutrinoceros:
Hey I finally managed to write a proper test out of your example, mind if I push it to this branch @n-claes ?

n-claes:
> Hey I finally managed to write a proper test out of your example, mind if I push it to this branch @n-claes ?

Of course, feel free!

neutrinoceros:
There you go. I checked that the test fails on the main branch and passes on yours 🎉 
Haven't looked for a way to test the timestamp part though, but it's still better than no tests :)

neutrinoceros (approved):

neutrinoceros:
🎉

PR #2731, opened 2020-07-07
[bugfix] Ensure SPH visualizations respect periodicity

jzuhone:
## PR Summary

This PR makes changes to SPH pixelization for slices, projections, and arbitrary grids which use scatter smoothing to ensure they respect periodicity. This was not true previous to this PR, so (for example) if you had a periodic box simulation and you made a projection near a domain edge the regions outside the box would be blank.

I also discovered a bug in region selection for SPH particles when the region crosses a periodic boundary which is fixed here. 

The pixelization methods which use "gather" smoothing appear to not need any changes. 

I don't think this works for off-axis projections because in that case the particle positions are rotated to the new frame first and then projected. So we need to figure out how to handle that.

Also need to decide if answer testing is sufficient or if we need to test in some other way as well. 

## PR Checklist

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -871,6 +871,12 @@ class YTCoveringGrid(YTSelectionContainer3D):
 
         bounds, size = self._get_grid_bounds_size()
 
+        period = self.ds.coordinates.period.copy()
+        if hasattr(period, 'in_units'):
+            period = period.in_units("code_length").d
+        # TODO maybe there is a better way of handling this
+        periodic = any(self.ds.periodicity)
+
         if smoothing_style == "scatter":
             for field in fields:
                 fi = self.ds._get_field_info(field)
@@ -902,6 +908,8 @@ class YTCoveringGrid(YTSelectionContainer3D):
                         field_quantity,
                         bounds,
                         pbar=pbar,
+                        check_period=int(periodic),
+                        period=period
                     )
                     if normalize:
                         pixelize_sph_kernel_arbitrary_grid(
@@ -915,6 +923,8 @@ class YTCoveringGrid(YTSelectionContainer3D):
                             np.ones(dens.shape[0]),
                             bounds,
                             pbar=pbar,
+                            check_period=int(periodic),
+                            period=period
                         )
 
                 if normalize:
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -875,7 +875,7 @@ class YTCoveringGrid(YTSelectionContainer3D):
         if hasattr(period, "in_units"):
             period = period.in_units("code_length").d
         # TODO maybe there is a better way of handling this
-        periodic = any(self.ds.periodicity)
+        is_periodic = int(any(self.ds.periodicity))
 
         if smoothing_style == "scatter":
             for field in fields:
@@ -908,7 +908,7 @@ class YTCoveringGrid(YTSelectionContainer3D):
                         field_quantity,
                         bounds,
                         pbar=pbar,
-                        check_period=int(periodic),
+                        check_period=is_periodic,
                         period=period,
                     )
                     if normalize:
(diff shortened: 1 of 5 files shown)


matthewturk:
I believe this is right -- I've read the `xiterv` etc stuff carefully and it looks like the logic is all good.  Thank you for catching this.

matthewturk:
It is entirely likely that for some outputs we will need to regenerate answer tests.

yt/geometry/selection_routines.pyx
```diff
@@ -1076,11 +1076,7 @@ cdef class RegionSelector(SelectorObject):
         cdef np.float64_t r2 = radius**2
         cdef np.float64_t dmin = 0
         for i in range(3):
-            if self.periodicity[i] and self.check_period[i]:
```
jzuhone, line 1079: I now see that what I did here is wrong but I am not sure how to fix it. Anyone have any ideas? ping @matthewturk @AshKelly

matthewturk: This stuff always confuses me.  Would it be OK to go back to the original version, and punt on this for a later commit?  Because I *think* it should mostly still work ... right?

jzuhone: @matthewturk no, unfortunately the original way doesn't work correctly either. I think we have to resolve it now.

jzuhone: Never mind, I figured this out!!

jzuhone:
I am pretty sure the answer test failures have to do with the fact that we weren't handling periodicity appropriately.

jzuhone:
@matthewturk did you still have that website set up where we could look at image answer differences?

matthewturk:
Oh my gosh, it feels like that was ten years ago.  But, yes!

https://github.com/matthewturk/yt4-gallery
https://matthewturk.github.io/yt4-gallery/

If you want to go ahead and issue the PR to the repo that'd be fine, or I can generate some.

jzuhone:
Aside from the fact that this probably doesn't fix off-axis projections, this is good to go.

jzuhone:
Anyone else want to have a look at this? It's ready to go.

neutrinoceros:
I'm in.

yt/data_objects/construction_data_containers.py
```diff
@@ -871,6 +871,12 @@ def _fill_sph_particles(self, fields):
 
         bounds, size = self._get_grid_bounds_size()
 
+        period = self.ds.coordinates.period.copy()
+        if hasattr(period, "in_units"):
```
neutrinoceros, line 875: wait, why would this condition ever _not_ be checked ? can this array ever be a vanilla ndarray ?

yt/data_objects/construction_data_containers.py
```diff
@@ -871,6 +871,12 @@ def _fill_sph_particles(self, fields):
...
+            period = period.in_units("code_length").d
+        # TODO maybe there is a better way of handling this
+        periodic = any(self.ds.periodicity)
```
neutrinoceros, line 878: since boolean value is never used directly... I'd also suggest a minor variable name change
```suggestion
        is_periodic = int(any(self.ds.periodicity))
```

yt/data_objects/selection_data_containers.py
```diff
@@ -773,7 +773,7 @@ def _get_bbox(self):
         """
         Return the minimum bounding box for the region.
         """
-        return self.left_edge, self.right_edge
+        return self.left_edge.copy(), self.right_edge.copy()
```
neutrinoceros, line 776: good catch !

yt/data_objects/construction_data_containers.py
jzuhone: I'm not sure--this logic was copied over from the different coordinate handlers, which seem to think this is a possibility.

yt/data_objects/construction_data_containers.py
jzuhone: Done.

yt/data_objects/construction_data_containers.py
matthewturk: Yes, it can be.

neutrinoceros: Ok, this could probably be bullet-proofed but it's not the point here. Thanks for your prompt reply !

matthewturk: At some point we also want to get to the point where it does not ever require conversion, IMO -- which is not the same as requiring it is vanilla ndarray.

PR #2735, opened 2020-07-10
Update test dependencies

Xarthisius:
I'm bumping a few dependencies to minimal versions that have binary wheels for py38 in order to avoid building them each time. In particular `pandas` takes some time to build.

### DIFF
--- a/tests/test_requirements.txt
+++ b/tests/test_requirements.txt
@@ -5,7 +5,7 @@ fastcache==1.0.2
 glueviz==0.13.3
 h5py==2.10.0
 ipython==7.1.1
-matplotlib==3.1.3
+matplotlib==3.3.0
 mock==2.0.0; python_version < '3.0'
 nose-timer==1.0.0
 nose==1.3.7
--- a/tests/test_requirements.txt
+++ b/tests/test_requirements.txt
@@ -1,15 +1,15 @@
-astropy==3.0.5
+astropy==3.2.3
 codecov==2.0.15
-coverage==4.5.1
+coverage==4.5.4
 fastcache==1.0.2
 glueviz==0.13.3
 h5py==2.10.0
 ipython==7.1.1
 matplotlib==3.1.3
 mock==2.0.0; python_version < '3.0'
-nose-timer==0.7.3
+nose-timer==1.0.0
 nose==1.3.7
-pandas==0.23.4
+pandas==0.25.3
 pytest~=5.2
 requests==2.20.0
 scipy==1.3.3
--- a/yt/utilities/command_line.py
+++ b/yt/utilities/command_line.py
@@ -1,3 +1,4 @@
+# isort: skip-file
 import argparse
 import base64
 import getpass
@@ -28,7 +29,7 @@ from yt.funcs import (
     mylog,
     update_hg_or_git,
 )
-from yt.startup_tasks import parser, subparsers
+
 from yt.utilities.configure import set_config
 from yt.utilities.exceptions import (
     YTCommandRequiresModule,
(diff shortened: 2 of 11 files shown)


Xarthisius:
@yt-fido test this please

matthewturk:
this looks good, modulo fido's approval

Xarthisius:
@yt-fido test this please

tests/test_requirements.txt
```diff
@@ -1,15 +1,15 @@
...
+astropy==3.2.3
 codecov==2.0.15
-coverage==4.5.1
+coverage==4.5.4
 fastcache==1.0.2
 glueviz==0.13.3
 h5py==2.10.0
 ipython==7.1.1
-matplotlib==3.1.3
+matplotlib==3.3.0
```
neutrinoceros, line 8: For now it's alright but maybe we should future proof this and use their bug fixes releases as soon as they come out ?
```suggestion
matplotlib~=3.3
```

tests/test_requirements.txt
```diff
@@ -1,15 +1,15 @@
...
 mock==2.0.0; python_version < '3.0'
-nose-timer==0.7.3
+nose-timer==1.0.0
 nose==1.3.7
-pandas==0.23.4
```
neutrinoceros, line 12: I'm guessing you have a good reason for this downgrade but I couldn't find it in the commit messages, just curious :)

Xarthisius: 0.23.4 -> 0.25.3 is not a downgrade ;)

neutrinoceros: oops, I only saw the third digit and then my brain shut down, apparently 😅

neutrinoceros: Although that reminds me that pandas 1.0 came out almost a year ago... I would naively expect the latest version to be more efficient, do you know if it is ?

Xarthisius: I think the idea behind `test_requirements.txt` is to have the earliest possible version that works. I don't entirely agree with that, but it is what it is. We should probably have a separate test build where all those things are completely unpinned.

neutrinoceros: Yes, that would help catch breaking updates from our deps early. May be worth keeping this in mind next time we discuss the testing infrastructure as a whole :)
thank you for clarifying this !

matthewturk:
This ready to go?

Xarthisius:
> This ready to go?

Yeah, as soon as appveyor finishes. Docs are failing for some reason, but I've seen it on other PRs too.

matthewturk:
Once this passes tests, I think we should merge!  Then we'll be back to green!

Xarthisius:
@neutrinoceros PTAL at https://github.com/yt-project/yt/pull/2735/commits/aefd6c25c7845c4d44e28de8852115518acf9809 That config option needs to be set before importing `startup_tasks`. Otherwise our subcommands for `yt` are broken, which is what's causing our docs to hang currently on master. I tried to use *isort*'s "action comments", but every time I run `isort` the line setting the `__command_line` is getting moved after imports. Can you figure out how to do that properly?

neutrinoceros:
@Xarthisius I think you might just need to update isort: version 5.1.4 seems to work correctly with the action comments (though the file isn't skipped, at least if I target it specifically).

yt/utilities/command_line.py
```diff
@@ -1,3 +1,4 @@
+# isort: skip-file
```
neutrinoceros, line 1: note that it'd be easier to manage if instead of this comment, the file was added to isort's ignore list in setup.cfg, but it shouldn't hold this PR back.

@matthewturk , can you do it in https://github.com/yt-project/yt/pull/2756 ?

PR #2736, opened 2020-07-10
Remove all occurences of VR.render() called right before VR.save()

Xarthisius:
This PR removes unnecessary and very costly `.render()` call, since `Scene.save()` explicitly calls `Scene.render()`.


Xarthisius:
@yt-fido test this please

chrishavlin:
related: pull request #2743 

the two PRs are compatible, but note that after 2743 is merged, doc/source/cookbook/sigma_clip.py could be user `render=False` in the second and third calls to `sc.save()` since the `sigma_clip` is applied after rendering.

matthewturk:
@Xarthisius can you rebase or merge with master so we can run the tests?

PR #2518, opened 2020-04-06
[WIP] support halo member id loading for FOF and HOP

brittonsmith:
## PR Summary

In conjunction with [PR #52](https://github.com/yt-project/yt_astro_analysis/pull/52) of yt_astro_analysis, this adds support to load halo member particles for halo catalogs created with yt HOP and FOF. This allows one to do:

```
import yt
ds = yt.load(...)
halo = ds.halo('halos', <id>)
print (halo['member_ids'])
```

## PR Checklist

- [x] Refactor into abstract classes used by `halo_catalog` and `gadget_fof` frontends
- [ ] Code passes flake8 checker
- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

Issue #83, opened 2011-06-02
Arguments to add_field are undocumented


yt-fido:
Originally reported by: **The yt Project (Bitbucket: [yt_analysis](https://bitbucket.org/yt_analysis), GitHub: Unknown)**

----------------------------------------

== Imported Ticket ==
 * **Summary:** Arguments to add_field are undocumented
 * **Component:** documentation
 * **Milestone:** 1.0
 * **Reporter:** mturk
 * **Owner:** mturk
 * **Resolution: fixed
 * **Status:** closed
 * **Created:** 1209486762000000
 * **Description:** These all need to be documented if we want a working plugin system.

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1212126597000000
 * **Field:** status
 * **Oldvalue:** new
 * **Newvalue:** assigned

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1212126597000000
 * **Field:** component
 * **Oldvalue:** lagos
 * **Newvalue:** documentation

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1212126597000000
 * **Field:** comment
 * **Oldvalue:** 1
 * **Newvalue:** 

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1215378259000000
 * **Field:** status
 * **Oldvalue:** assigned
 * **Newvalue:** closed

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1215378259000000
 * **Field:** resolution
 * **Oldvalue:** 
 * **Newvalue:** fixed

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1215378259000000
 * **Field:** comment
 * **Oldvalue:** 2
 * **Newvalue:** Closed in r650

----------------------------------------
- Bitbucket: https://bitbucket.org/yt_analysis/yt/issue/83

PR #2738, opened 2020-07-11
[new feature] load particles from yt HoP/FoF catalogs (alternative to #2518)

brittonsmith:
## PR Summary

This is an alternative to PR #2518, which I messed up after a few too many merges with master.

This allows one to access halo particles for yt HoP/FoF catalogs using `halo` data containers in the following fashion:
```
import yt
ds = yt.load(...)
halo = ds.halo('halos', <id>)
print (halo['member_ids'])
```

The original version of this PR included a refactor of the analogous infrastructure in the `gadget_fof` frontend, but in the interest of time, I'd like to put that part off and see this merged first. I think it will be easier to review a refactor separate from the feature additions here. This implementation includes a couple base classes that I plan to implement in `gadget_fof` at a later date.

## PR Checklist


brittonsmith:
@Xarthisius, when you have a moment, could you put the `tiny_fof_halos` dataset from [website PR #83](https://github.com/yt-project/website/pull/83) on the testing machine?

yt/frontends/adaptahop/data_structures.py
```diff
@@ -35,9 +37,24 @@
 from .fields import \
     AdaptaHOPFieldInfo
 
+class AdaptaHOPParticleIndex(ParticleIndex):
```
brittonsmith, line 40: @cphyc, this was `HaloCatalogParticleIndex`, but I realized that it wasn't being used anywhere else so I moved it here. Let me know if you'd like me to move it back. It's also possible that this can work simply as `ParticleIndex`, but I didn't test that.

brittonsmith:
@matthewturk @neutrinoceros I just ran `black` on this PR and merged with `master` and now the diff is hundreds of files. Specifically, I did:
```
git checkout hnew
black --line-length 88 yt
git merge --strategy ours master
git commit -am 'Merge with master'
git push <my_fork> hnew
```
Did I do this incorrectly?

neutrinoceros:
@brittonsmith I'm so sorry this is on me, clearly the `git merge --strategy ours master` line is the problem here and in retrospect this seems only viable for very atomic changes (maybe not even all of them).

The _prefered_ way to deal with this is rebasing and solving conflicts as they show up. Here you'd need to do something like this
```bash
git reset --hard HEAD^
git rebase -i  master
# ... solve conflicts
git push --force
```

yt/frontends/adaptahop/data_structures.py
cphyc: I'm fine with this!

brittonsmith:
@neutrinoceros, thanks I will give this a try as soon as I can. I'll need some time to do this carefully. In the meantime, I've force pushed this back to its original state.

neutrinoceros:
note that on your other PR there are still some (important) flake8 errors:
https://travis-ci.org/github/yt-project/yt/jobs/715161254

brittonsmith:
Ah, thanks! I'll deal with those now.

brittonsmith:
Ok, this is updated and ready for review now. If anyone would like to test this out, the required sample data is [here](http://yt-project.org/data/tiny_fof_halos.tar.gz).

yt/frontends/halo_catalog/io.py
```diff
@@ -104,9 +78,143 @@ def _count_particles(self, data_file):
...
+                else:
+                    member_fields[ptype].append(field)
+
+        all_data = self._read_scalar_fields(dobj, scalar_fields)
+        all_data.update(self._read_member_fields(dobj, member_fields))
+
+        for field, field_data in all_data.items():
+            yield field, field_data
+
+    def _read_particle_selection(self, dobj, fields):
```
brittonsmith, line 116: This function is a direct copy of `IOHandlerGadgetFOFHaloHDF5._read_particle_selection`. The original lost version of this PR included a refactor which merged them. I intend to re-do that refactor at some point, but I think it would be easier to do after this PR has been merged. I appeal for clemency until then.

I'm also happy to (for now) replace this function definition explicitly with:
```
_read_particle_selection = IOHandlerGadgetFOFHaloHDF5._read_particle_selection
```
That would at least cut down on code to review and make it clear what work will be done in the future.

yt/geometry/particle_geometry_handler.py
```diff
@@ -46,6 +46,9 @@ def convert(self, unit):
         return self.dataset.conversion_factors[unit]
 
     def _setup_filenames(self):
+        if hasattr(self, "data_files"):
+            pass
```
matthewturk, line 50: I think this should be "return" or something -- a `pass` here will still call the stuff below.

brittonsmith: Oops, you're totally right! For what it's worth, there do seem to be some instances (unrelated to this PR) where this gets called more than once. Maybe we can put a pin in that. We could also convert `data_files` to a property that calls `_setup_filenames`.

neutrinoceros:
> I think it will be easier to review a refactor separate from the feature additions here

for the record I agree this is an acceptable route if it makes reviewing easier. I'll have a go at it !

yt/frontends/halo_catalog/tests/test_outputs.py
```diff
@@ -78,11 +86,46 @@ def test_halo_catalog_boundary_particles(self):
...
+def test_halo_quantities():
+    ds = data_dir_load(t46)
+    ad = ds.all_data()
+    for i in range(ds.index.total_particles):
+        hid = int(ad["halos", "particle_identifier"][i])
+        halo = ds.halo("halos", hid)
+        for field in ["mass", "position", "velocity"]:
+            v1 = ad["halos", "particle_%s" % field][i]
+            v2 = getattr(halo, field)
+            assert_allclose_units(
```
neutrinoceros, line 113: if I'm reading this right, you're the arrays you're comparing shouldn't just be approximately equal but rather be the same exact objects ?
so should this be
```python
assert v1 is v2
```
or
```python
assert_array_equal(
```
?

yt/geometry/particle_geometry_handler.py
```diff
@@ -46,6 +46,9 @@ def convert(self, unit):
         return self.dataset.conversion_factors[unit]
 
     def _setup_filenames(self):
+        if hasattr(self, "data_files"):
```
neutrinoceros, line 49: I'm not sure why this is needed. Seems to me like the only way this attribute is already set is that the method was overriden in some subclass, in which case this is never called. Unless you except this method to be called multiple times downstream ? Since this is a "private" method, all calls should be within the codebase and not in user codes, so if there _are_ multiple calls indeed I think it could be considered a bug and fixed there instead of here.

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -93,10 +90,17 @@ def _read_particle_positions(self, ptype, f=None):
         return pos
 
 
-class HaloCatalogDataset(SavedDataset):
+class YTHaloCatalogDataset(SavedDataset):
```
neutrinoceros, line 93: Not from this PR but this inheritance confuses me. What does `YTHaloCatalogDataset` have in common with `SavedDataset`, conceptually ?

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+                    np.array(
+                        [data_file.index_start[ptype] for data_file in self.data_files]
+                    ),
+                )
+                for ptype in self.ds.particle_types_raw
+            ]
+        )
+
+    def _create_halo_id_table(self):
+        pass
```
neutrinoceros, line 219:
```suggestion
        raise NotImplementedError
```
not sure what the intention is here, might be the wrong approach but if that's the case then an inline comment would be nice for clarity :)

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+            [
+                (
+                    ptype,
+                    np.array(
+                        [data_file.index_start[ptype] for data_file in self.data_files]
+                    ),
+                )
+                for ptype in self.ds.particle_types_raw
+            ]
+        )
```
neutrinoceros, line 216: Here's an attempt at making this look flatter
```suggestion
        self._halo_index_start = {}
        for ptype in self.ds.particle_types_raw:
            d = [data_file.index_start[ptype] for data_file in self.data_files]
            self._halo_index_start.update({ptype: np.array(d)})
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+
+    def _create_halo_id_table(self):
+        pass
+
+    def _detect_output_fields(self):
+        field_list = []
+        scalar_field_list = []
+        units = {}
+        pc = dict(
```
neutrinoceros, line 225: same pattern as above could be applied here

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+        self.field_list = field_list
+        self.scalar_field_list = scalar_field_list
+        ds = self.dataset
+        ds.scalar_field_list = scalar_field_list
+        ds.particle_types = tuple(set(pt for pt, ds in field_list))
+        ds.field_units.update(units)
+        ds.particle_types_raw = ds.particle_types
+
+    def _get_halo_file_indices(self, ptype, identifiers):
+        return np.digitize(identifiers, self._halo_index_start[ptype], right=False) - 1
```
neutrinoceros, line 254: naive question; could you explain the "-1" in an inline comment ?

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+        scalar_index = identifier - self._halo_index_start[ptype][i_scalar]
+        return scalar_index
+
+    def _get_halo_values(self, ptype, identifiers, fields, f=None):
+        """
+        Get field values for halo data containers.
+        """
+
+        # if a file is already open, don't open it again
+        filename = None if f is None else f.filename
```
neutrinoceros, line 267:
```suggestion
        try:
            filename = f.filename
        except AttributeError:
            filename = None
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+
+        data = defaultdict(lambda: np.empty(identifiers.size))
+        i_scalars = self._get_halo_file_indices(ptype, identifiers)
+        for i_scalar in np.unique(i_scalars):
+            target = i_scalars == i_scalar
```
neutrinoceros, line 272: I got this after reading a few times, but could you add a comment to clarify that "target" is a mask here ?

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+                    my_f, ptype, field, scalar_indices[target]
+                )
+
+            if self.data_files[i_scalar].filename != filename:
+                my_f.close()
+
+        return data
+
+    def _identify_base_chunk(self, dobj):
+        pass
```
neutrinoceros, line 293:
```suggestion
        raise NotImplementedError
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+
+    def _read_halo_particle_field(self, fh, ptype, field, indices):
+        return fh[field][indices]
+
+    def _read_particle_fields(self, fields, dobj, chunk=None):
+        if len(fields) == 0:
```
neutrinoceros, line 299:
```suggestion
        if not fields:
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+            return {}, []
+        fields_to_read, fields_to_generate = self._split_fields(fields)
+        if len(fields_to_read) == 0:
```
neutrinoceros, line 302:
```suggestion
        if not fields_to_read:
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+            "particle_types_raw",
+            "particle_types",
+            "periodicity",
+        ]:
+            setattr(self, attr, getattr(self.real_ds, attr))
+
+        super(HaloDataset, self).__init__(self.real_ds.parameter_filename, dataset_type)
+
+    def print_key_parameters(self):
+        pass
```
neutrinoceros, line 339:
```suggestion
        raise NotImplementedError
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+
+    def _set_derived_attrs(self):
+        pass
```
neutrinoceros, line 342:
```suggestion
        raise NotImplementedError
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+            "hubble_constant",
+            "omega_lambda",
+            "omega_matter",
+            "unique_identifier",
+        ]:
+            setattr(self, attr, getattr(self.real_ds, attr))
+
+    def set_code_units(self):
+        for unit in ["length", "time", "mass", "velocity", "magnetic", "temperature"]:
+            my_unit = "%s_unit" % unit
```
neutrinoceros, line 364: For what it's worth I think "unit" isn't a very appropriate name for this iterator, but well, not super important.
```suggestion
            my_unit = f"{unit}_unit"
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+            setattr(self, my_unit, getattr(self.real_ds, my_unit, None))
+        self.unit_registry = self.real_ds.unit_registry
+
+    def __repr__(self):
+        return "%s" % self.real_ds
```
neutrinoceros, line 369:
```suggestion
        return str(self.real_ds)
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+
+    _index_class = YTHaloParticleIndex
+    _file_class = YTHaloCatalogFile
+    _field_info_class = YTHaloCatalogHaloFieldInfo
+
+    def __init__(self, ds, dataset_type="ythalo"):
+        super(YTHaloDataset, self).__init__(ds, dataset_type)
+
+    def _set_code_unit_attributes(self):
+        pass
```
neutrinoceros, line 388:
```suggestion
        raise NotImplementedError
```

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+
+    @classmethod
+    def _is_valid(self, *args, **kwargs):
+        return False
```
neutrinoceros, line 392: same as above. If it's really not meant not to be instanciated with yt.load, please state so (we currently have at least one dataset class that should be loadable but isn't because the validator was never completed)

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+        self.ptype = ptype
+        self._current_particle_type = ptype
+        super(HaloContainer, self).__init__(ds, {})
+
+        self._set_identifiers(particle_identifier)
+
+        # Find the file that has the scalar values for this halo.
+        i_scalar = self.index._get_halo_file_indices(ptype, [self.particle_identifier])[
+            0
+        ]
```
neutrinoceros, line 421: wow black is acting pretty weird here. Maybe the single worst example I've seen yet. Any way this line could be made a tiny bit shorter so we don't have to keep this ? (look at me complaining about the tool I advocated for, saying that it would save us future complaints about code style...)

yt/frontends/halo_catalog/data_structures.py
```diff
@@ -146,3 +177,363 @@ def _is_valid(self, *args, **kwargs):
...
+        halo_fields = self._get_member_fieldnames()
+        my_data = self.index._get_halo_values(
+            self.ptype, np.array([self.particle_identifier]), halo_fields
+        )
+        self._io_data = dict(
+            (field, np.int64(val[0])) for field, val in my_data.items()
+        )
+
+    def __repr__(self):
+        return "%s_%s_%09d" % (self.ds, self.ptype, self.particle_identifier)
```
neutrinoceros, line 481:
```suggestion
        return f"{self.ds}_{self.ptype}_{self.particle_identifier:09d}"
```

yt/frontends/halo_catalog/io.py
```diff
@@ -104,9 +79,93 @@ def _count_particles(self, data_file):
...
+                        )
+                        field_data[field_start:field_end] = my_data
+            field_start = field_end
+        return all_data
+
+    def _read_scalar_fields(self, dobj, scalar_fields):
+        all_data = {}
+        if not scalar_fields:
+            return all_data
+        with h5py.File(dobj.scalar_data_file.filename, "r") as f:
```
neutrinoceros, line 166: The h5py api has become less permissive
```suggestion
        with h5py.File(dobj.scalar_data_file.filename, mode="r") as f:
```

yt/frontends/halo_catalog/io.py
```diff
@@ -104,9 +79,93 @@ def _count_particles(self, data_file):
...
+            return all_data
+        field_start = 0
+        for i, data_file in enumerate(dobj.field_data_files):
+            start_index = dobj.field_data_start[i]
+            end_index = dobj.field_data_end[i]
+            pcount = end_index - start_index
+            if pcount == 0:
+                continue
+            field_end = field_start + end_index - start_index
+            with h5py.File(data_file.filename, "r") as f:
```
neutrinoceros, line 151: Same as bellow
```suggestion
            with h5py.File(data_file.filename, mode="r") as f:
```

yt/frontends/halo_catalog/io.py
```diff
@@ -104,9 +79,93 @@ def _count_particles(self, data_file):
...
+                id_fields = [("halos", field) for field in f["particles"]]
+            else:
+                id_fields = []
+
+        return scalar_fields + id_fields, scalar_fields, id_fields, units
+
+    def _read_member_fields(self, dobj, member_fields):
+        all_data = defaultdict(lambda: np.empty(dobj.particle_number, dtype=np.float64))
+        if not member_fields:
+            return all_data
```
neutrinoceros, line 142:
```suggestion
        if not member_fields:
            return all_data
        all_data = defaultdict(lambda: np.empty(dobj.particle_number, dtype=np.float64))
```

yt/data_objects/static_output.py
```diff
@@ -1115,41 +1116,48 @@ def set_units(self):
...
-                    new_unit,
-                    my_u.base_value / (1 + self.current_redshift),
-                    dimensions.length,
-                    "\\rm{%s}/(1+z)" % my_unit,
-                    prefixable=True,
-                )
-            self.unit_registry.modify("a", 1 / (1 + self.current_redshift))
+            if getattr(self, "current_redshift", None):
+                # Comoving lengths
+                for my_unit in ["m", "pc", "AU", "au"]:
```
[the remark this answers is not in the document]
brittonsmith: Not as far as I know.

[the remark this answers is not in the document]
neutrinoceros: worth a shot :)

doc/source/examining/loading_data.rst
```diff
@@ -2252,11 +2252,30 @@ available here are similar to other catalogs.  Any addition
...
    print(ad["halos", "particle_mass"])
 
+Halo Data Containers
+""""""""""""""""""""
+
+Halo particles can be accessed by creating halo data containers with the
+type of halo ("halos") and the halo id and then querying the "member_ids"
+field. Halo containers have mass, radius, position, and velocity
+attributes. Additional fields for which there will be one value per halo
+can be accessed in the same manner as conventional data containers.
```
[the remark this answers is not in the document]
brittonsmith: Doing `halo["mass"]` will raise a `YTFieldNotFound` error. There is the following mapping between the attributes and the fields they query:
- `mass` -> `particle_mass`
- `position` -> `particle_position`
- `velocity` -> `particle_velocity`
- `radius` -> `virial_radius`

This is consistent with the other frontend that supports halo data containers like this (`gadget_fof`), so I think it makes sense to keep it like this until we decide on a new standard. The other issue is that doing `halo.particle_mass` might lead the users to expect to be getting the masses of all the member particles themselves.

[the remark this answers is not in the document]
neutrinoceros: All hail consistency ! Ok thank you for the detailed answer, I get a better grasp of this now :)

yt/frontends/halo_catalog/io.py
```diff
@@ -38,7 +39,7 @@ def _yield_coordinates(self, data_file):
         with h5py.File(data_file.filename, "r") as f:
             units = parse_h5_attr(f[pn % "x"], "units")
             x, y, z = (
-                self.ds.arr(f[pn % ax].value.astype("float64"), units) for ax in "xyz"
+                self.ds.arr(f[pn % ax][()].astype("float64"), units) for ax in "xyz"
```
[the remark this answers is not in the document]
brittonsmith: The `.value` is deprecated `h5py` syntax. The recommended change is go from `<dataset>.value` to `<dataset>[()]`. Most of these were updated long ago, so I'm not sure how this one slipped through.

yt/frontends/halo_catalog/io.py
brittonsmith: I'm not sure I understand this one. How can we return `all_data` before defining it?

neutrinoceros: we can't. I reviewed this before going to bed. Clearly I was being over-confident 😅

yt/data_objects/static_output.py
```diff
@@ -1115,41 +1116,48 @@ def set_units(self):
...
+        if not getattr(self, "cosmological_simulation", False):
+            return
+
+        # Set dynamical dark energy parameters
+        use_dark_factor = getattr(self, "use_dark_factor", False)
+        w_0 = getattr(self, "w_0", -1.0)
+        w_a = getattr(self, "w_a", 0.0)
+
+        # many frontends do not set this
+        setdefaultattr(self, "omega_radiation", 0.0)
```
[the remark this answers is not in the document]
neutrinoceros: Maybe that's one more thing we would need to factor in to design a new set of Dataset abstract classes... but that's way out of scope here. Thanks for your reply

yt/frontends/adaptahop/data_structures.py
```diff
@@ -15,20 +15,37 @@
 
 from yt.data_objects.data_containers import YTSelectionContainer
 from yt.data_objects.static_output import Dataset
-from yt.frontends.halo_catalog.data_structures import (
-    HaloCatalogFile,
-    HaloCatalogParticleIndex,
-)
+from yt.frontends.halo_catalog.data_structures import HaloCatalogFile
```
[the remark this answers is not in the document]
neutrinoceros: Hum yeah, maybe one day we can clean this up. But not today !

yt/frontends/adaptahop/data_structures.py
```diff
@@ -15,20 +15,37 @@
...
+from yt.geometry.particle_geometry_handler import ParticleIndex
 from yt.units import Mpc
 from yt.utilities.cython_fortran_utils import FortranFile
 
 from .definitions import HEADER_ATTRIBUTES
 from .fields import AdaptaHOPFieldInfo
 
 
+class AdaptaHOPParticleIndex(ParticleIndex):
+    def _setup_filenames(self):
```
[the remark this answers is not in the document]
neutrinoceros: please do, I'll be happy to help !

yt/frontends/halo_catalog/io.py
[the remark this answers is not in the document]
neutrinoceros: that's a _weird_ recommendation afaic, but well, it's not up to you, thanks for the clarification

yt/data_objects/static_output.py
```diff
@@ -1115,41 +1116,48 @@ def set_units(self):
...
+            if getattr(self, "current_redshift", None):
+                # Comoving lengths
+                for my_unit in ["m", "pc", "AU", "au"]:
+                    new_unit = f"{my_unit}cm"
+                    my_u = Unit(my_unit, registry=self.unit_registry)
+                    self.unit_registry.add(
+                        new_unit,
+                        my_u.base_value / (1 + self.current_redshift),
+                        dimensions.length,
+                        "\\rm\{%s\}/(1+z)" % my_unit,
```
brittonsmith, line 1128: I reverted this because the suggested change produced a flake8 error:
```
E999 SyntaxError: f-string: single '}' is not allowed
```
I tried running it through `flynt` and it had no suggestions.

yt/geometry/particle_geometry_handler.py
brittonsmith: This is a good question. When developing this PR, I found a couple instances where this function does get called multiple times. Functionally, it's fine, but it can be a performance hit. This does not apply to this PR anymore (i.e., `_setup_filenames` only ever gets called once).

If I change this to raising an exception, there are ~3-4 tests in the test suite that will fail. So there is a bug that may require some refactoring. I can remove this and we can accept that small performance hit for now or I could file an issue for this to be revisited. What do you think?

matthewturk:
I think we should not do this as a condition of this particular PR, and
instead file a bug.  We have other things that need to get refactored that
are quite close to this in the call-stack.

yt/geometry/particle_geometry_handler.py
neutrinoceros: So if I'm reading this right, this line is actively silencing failing tests ?
Since the problem is older than this PR, keeping track of it as an issue would be a good first step, and I agree the problem doesn't _need_ to be addressed here.

matthewturk:
It's less that it's silencing them, and more that we have some situations
where it's possible to initialize the geometry handler multiple times --
which is not ideal.

yt/data_objects/static_output.py
neutrinoceros: if flynt is happy, I'm happy.

yt/frontends/halo_catalog/data_structures.py
neutrinoceros: I don't mind leaving this as an Easter egg actually: it's pretty funny that it happens, but only because it's so rare !

brittonsmith:
@neutrinoceros, I think that's all of your comments addressed. If I have missed or mistakenly resolved any comments, please do let me know. Thanks for taking the time to review this!

yt/frontends/halo_catalog/tests/test_outputs.py
```diff
@@ -44,7 +51,7 @@ def test_halo_catalog(self):
         fn = fake_halo_catalog(data)
         ds = yt_load(fn)
 
-        assert isinstance(ds, HaloCatalogDataset)
+        assert isinstance(ds, YTHaloCatalogDataset)
```
neutrinoceros, line 54: Oh I missed this but in tests, checking for exact typing might be preferable (same applies a few lines bellow this)
```suggestion
        assert type(ds) is YTHaloCatalogDataset
```

neutrinoceros:
Yeah actually there are still a couple of them but Github hides them by default... I'll ping you on the most important one

yt/frontends/halo_catalog/data_structures.py
neutrinoceros: @brittonsmith  👋

yt/frontends/halo_catalog/tests/test_outputs.py
brittonsmith: Good catch, I'll fix that.

yt/frontends/halo_catalog/data_structures.py
brittonsmith: Done.

yt/frontends/halo_catalog/data_structures.py
brittonsmith: No, this will be called by `YTSelectionContainer.get_data` but is unnecessary here, so I want to pass.

yt/frontends/halo_catalog/data_structures.py
brittonsmith: This is called by the base class and I want to do nothing since this is basically a hidden dataset.

yt/frontends/halo_catalog/data_structures.py
brittonsmith: Same, I want to pass here.

yt/frontends/halo_catalog/data_structures.py
brittonsmith: Same.

brittonsmith:
@neutrinoceros, ok now I think I've finished everything.

yt/frontends/halo_catalog/data_structures.py
neutrinoceros: Ok well it's not crucial here since the function isn't user-facing, but for the record, I find that try-except blocks are an extremely useful pattern for handling user data. Most of the bugs I've been dealing with recently are solved by changing ifs into tries. But then again I didn't realise performance was an issue here, so this is fine :)

neutrinoceros:
When tests pass I'll merge this

brittonsmith:
Thanks!

PR #2739, opened 2020-07-12
cleanup: harmonise _is_valid methods (followup #2695)

neutrinoceros:
## PR Summary

Follow-up PR to #2695 :  update all `_is_valid()` and `__new__()` implementations for `Dataset` children classes.

note: this will conflict with https://github.com/yt-project/yt/pull/2807, and should be rebased after it goes in, leaving this as a draft in the mean time

## todo
- [x] replace `args[0]` with named positional argument `filename` in
  * [x] `<DatasetChild>._is_valid`
  * [x] `<DatasetChild>.__init__` (nothing to be done :tada:)
  * [x] `<DatasetChild>.__new__`
  * [x] update `<DatasetChild>._is_valid`'s first positional argument to `cls` (it's a class method but currently some implementations incorrectly name the first arg `self`) (not really thematically linked to this PR but while I'm at it, might as well fix this)
- [x] update `YTOutputNotIdentified`

Issue #2708, opened 2020-07-01
Decreasing code duplication


cphyc:
### Issue

**Issue summary**

Currently there is a lot of code duplication in `yt/frontends/`. Fixing this would likely require a massive revamping and update of the codebase, but would improve code readability and maintenance.

I am opening the issue to track this issue and provide some scripts to work on it.

**Code for reproduction**

```bash
pylint --disable=all --enable=similarities yt/frontends/
```
The output can be found in [yt.frontends.similarities.log](https://github.com/yt-project/yt/files/4858256/yt.frontends.similarities.log).

**Expected outcome**

No duplication! Or only a few lines where we want things to be explicit, e.g. when defining fields.

neutrinoceros:
I'd like to see a plot with the history of that number (maybe normalised by the total number of lines), maybe sample once every 10 commits or so ?

cphyc:
![duplicate_ratio](https://user-images.githubusercontent.com/5411875/86256412-35d4d400-bbb0-11ea-8244-d15f22d9c299.png)

---

The data to generate the plot are obtained from
```bash
#!/bin/bash

echo "commit,date,nlines,nduplicates" > duplicates.csv
for commit in $(git rev-list master | python -c 'import sys; v=sys.stdin.read(); print("\n".join(v.split("\n")[::500]))'); do
    git checkout $commit -q || exit 1
    commit_date=$(git show -s --format=%ci $commit)
    nduplicates=$(pylint --disable=all --enable=similarities $(git ls-files yt/frontends) | wc -l)
    nlines=$(git ls-files yt/frontends | xargs cat | wc -l)
    echo $commit, $commit_date, $nlines, $nduplicates
done >> duplicates.csv
```

The plot is made with
```python
import pandas as pd
import matplotlib.pyplot as plt
df = pd.read_csv('duplicates.csv')
df['date'] = pd.to_datetime(df['date'])
df['duplicate_ratio'] = df['nduplicates'] / df['nlines'] * 100

plt.plot('date' ,'duplicate_ratio', data=df)
plt.xlabel('Date')
plt.ylabel('Duplicate ratio (%)', color='C0')

plt.twinx()
plt.plot('date', 'nduplicates', data=df, color='C1')
plt.ylabel('# duplicates', color='C1')
plt.savefig('duplicate_ratio.png')
```

matthewturk:
I would be so incredibly interested in pursuing de-duplication.

cphyc:
[...]

PR #2740, opened 2020-07-12
[EXP] refactor _is_valid methods (frontends code deduplication)

neutrinoceros:
## PR Summary
Follow up to #2695  and #2739 :
reduce code duplication in `_is_valid` methods with a decorator that implements a common pattern found in format validators:

```python
def _is_valid(filename):
    try:
        with open(filename) as fileh:
            valid = fileh.readlines()[999] == "myformat"
        return valid
    except (OSError, IndexError):
        return False

# is equivalent to
@invalidate_exceptions(OSError, IndexError)
def _is_valid(filename):
    with open(filename) as fileh:
        valid = fileh.readlines()[999] == "myformat"
    return valid
```
A possible alternative name for this decorator factory could be `known_invalidators`.

Refactoring  a lot of `_is_valid` implementations (address #2708 ):
most of those refactors are logic simplifications that could be performed anytime. Other times I simply removed typing checks for `filename` since it's always going to be a string after #2695 
In rare cases (maybe just one), I also removed a Python version check (special casing Python2, not supported anymore).

There is still a lot of repeated code here. In particular, the way fits and hdf5 file are validated or rejected upon missing dependencies could be harmonised and simplified, but I'm not happy with the ideas I've tried so far in this direction.

todo:
- [x] finish replacing `Exception` with more specific exception types


yt/frontends/art/data_structures.py
```diff
@@ -620,30 +616,24 @@ def create_field_info(self):
...
-            except Exception: pass
+
+        with open(filename, 'rb') as fh:
+            amr_prefix, amr_suffix = filename_pattern['amr']
+            for c in glob.glob(os.path.dirname(os.path.abspath(filename))+"/*"):
+                cname = os.path.basename(c)
+                if cname.endswith(amr_suffix) and cname.startswith(amr_prefix):
+                    return False
+            print("boh")
             try:
```
neutrinoceros, line 637: this try block seems much weaker after deeper inspection : it's very easy to pass this without raising any exception, using fake data.

neutrinoceros:
This has become virtually impossible to rebase. Closing.

Issue #2733, opened 2020-07-10
Clarity of volume rendering workflow examples (Scene.render() vs Scene.save())


chrishavlin:
### Bug report 

Not a bug, but the proper workflow for when to use `Scene.render()` and `Scene.save()` needs clarification. 

There are several docstring examples (e.g., [Scene.render](https://yt-project.org/doc/reference/api/yt.visualization.volume_rendering.scene.html#yt.visualization.volume_rendering.scene.Scene.render) and [Scene.save](https://yt-project.org/doc/reference/api/yt.visualization.volume_rendering.scene.html#yt.visualization.volume_rendering.scene.Scene.save)) for rendering that suggest the following workflow : 

```python
import yt
ds = yt.load('IsolatedGalaxy/galaxy0030/galaxy0030')
sc = yt.create_scene(ds)
im = sc.render() 
sc.save(sigma_clip=4.0) 
```
But `sc.save()` contains a call to `sc.render()`, so this workflow will render twice. 

`sc.save()` was changed (see [this commit](https://github.com/yt-project/yt/commit/cd2184576cd38468a6f59ad4611c06817bdfec0b)) so that `sc.render()` is always called, so presumably this is the desired behavior and the docstrings should just be corrected, but there should also be a comment in the code of `Scene.save()` explaining the need to always call `Scene.render()`. Are there ever cases where one would call `Scene.render()` ? 

The volume rendering examples [here](https://yt-project.org/docs/dev/visualizing/volume_rendering.html) only include the call to `Scene.scene()`, the examples [here](https://yt-project.org/docs/dev/cookbook/simple_plots.html#simple-volume-rendering) sometimes use `yt.volume_rendering()` and also `write_png`: 

```
im, sc = yt.volume_render(ds, 'density')
im.write_png("original.png", sigma_clip=8.0)
```
Note also that the docstring for [Scene.save](https://yt-project.org/doc/reference/api/yt.visualization.volume_rendering.scene.html#yt.visualization.volume_rendering.scene.Scene.save)  also has and alternate example in which `Scene.render()` is not called, but it's not clear why one would use either example. 

Perhaps a diagram of when to use what method would be useful?

chrishavlin:
@ngoldbaum , your input here would probably be insightful here since you modified `Scene.save()` :)

matthewturk:
@chrishavlin I think Nathan is taking a bit of time off, and hopefully we can work through this.  I'll dig in and report what I can.

[...]

PR #2743, opened 2020-07-13
Adds render option to scene.save to avoid re-rendering

chrishavlin:
## PR Summary

Adds an optional `render` boolean argument to `scene.save()`. Default behavior is `render=True` in which case `scene.save()` will call `scene.render()` prior to saving the image array. If `render=False`, will write the most recently rendered image array if it exists. Useful for avoiding unnecessary calls to `scene.render()` when saving at a range of `sigma_clip` values or when saving an image from a notebook after calling `sc.show()`. 

Includes updates to relevant docstrings. 

Addresses #2733 

## PR Checklist

- [x] Added  tests for new features: yt/visualization/volume_rendering/tests/test_save_render.py

### DIFF
--- a/doc/source/visualizing/volume_rendering.rst
+++ b/doc/source/visualizing/volume_rendering.rst
@@ -873,8 +873,7 @@ scene and volumes, you can simply call
 calling :meth:`~yt.visualization.volume_rendering.scene.Scene.render`. If you
 wish to save the most recently rendered image without rendering again, set
 ``render=False`` in the call to
-:meth:`~yt.visualization.volume_rendering.scene.Scene.save` and the most
-recently rendered image array will be used.  Cases where you
+:meth:`~yt.visualization.volume_rendering.scene.Scene.save`.  Cases where you
 may wish to use ``render=False`` include saving images at different
 ``sigma_clip`` values (see :ref:`cookbook-sigma_clip`) or when saving an image
 that has already been rendered in a Jupyter notebook using
--- a/doc/source/visualizing/volume_rendering.rst
+++ b/doc/source/visualizing/volume_rendering.rst
@@ -80,7 +80,7 @@ The scene's most important functions are
 :meth:`~yt.visualization.volume_rendering.scene.Scene.render` for
 casting rays through the scene and
 :meth:`~yt.visualization.volume_rendering.scene.Scene.save` for saving the
-resulting rendered image to disk.
+resulting rendered image to disk (see note on :ref:`when_to_render`).
 
 The easiest way to create a scene with sensible defaults is to use the
 functions:
@@ -329,7 +329,7 @@ To add a single gaussian layer with a color determined by a colormap value, use
    source.tfh.plot('transfer_function.png', profile_field='density')
 
    sc.save('rendering.png', sigma_clip=6)
-   
+
 
 add_gaussian
 """"""""""""
@@ -400,7 +400,7 @@ the volume rendering.
    def linramp(vals, minval, maxval):
        return (vals - vals.min())/(vals.max() - vals.min())
 
-   tf.map_to_colormap(np.log10(3e-31), np.log10(5e-27), colormap='arbre', 
+   tf.map_to_colormap(np.log10(3e-31), np.log10(5e-27), colormap='arbre',
                       scale_func=linramp)
 
    source.tfh.tf = tf
@@ -408,7 +408,7 @@ the volume rendering.
 
    source.tfh.plot('transfer_function.png', profile_field='density')
 
-   sc.save('rendering.png', sigma_clip=6)   
+   sc.save('rendering.png', sigma_clip=6)
 
 Projection Transfer Function
 ++++++++++++++++++++++++++++
@@ -522,10 +522,10 @@ adjusts for an opening view angle, so that the scene will have an
 element of perspective to it.
 :class:`~yt.visualization.volume_rendering.lens.StereoPerspectiveLens`
 is identical to PerspectiveLens, but it produces two images from nearby
-camera positions for use in 3D viewing. How 3D the image appears at viewing 
-will depend upon the value of 
-:attr:`~yt.visualization.volume_rendering.lens.StereoPerspectiveLens.disparity`, 
-which is half the maximum distance between two corresponding points in the left 
+camera positions for use in 3D viewing. How 3D the image appears at viewing
+will depend upon the value of
+:attr:`~yt.visualization.volume_rendering.lens.StereoPerspectiveLens.disparity`,
+which is half the maximum distance between two corresponding points in the left
 and right images. By default, it is set to 3 pixels.
 
 
@@ -556,11 +556,11 @@ see `the YouTube help: Upload virtual reality videos
 <https://support.google.com/youtube/answer/6316263?hl=en>`_).
 `This virtual reality video
 <https://youtu.be/ZYWY53X7UQE>`_ on YouTube is an example produced with
-:class:`~yt.visualization.volume_rendering.lens.StereoSphericalLens`. As in 
-the case of  
-:class:`~yt.visualization.volume_rendering.lens.StereoPerspectiveLens`, the 
-difference between the two images can be controlled by changing the value of 
-:attr:`~yt.visualization.volume_rendering.lens.StereoSphericalLens.disparity` 
+:class:`~yt.visualization.volume_rendering.lens.StereoSphericalLens`. As in
+the case of
+:class:`~yt.visualization.volume_rendering.lens.StereoPerspectiveLens`, the
+difference between the two images can be controlled by changing the value of
+:attr:`~yt.visualization.volume_rendering.lens.StereoSphericalLens.disparity`
 (See above).
 
 .. _annotated-vr-example:
@@ -573,7 +573,7 @@ Annotated Examples
              information can be hard.  We've provided information about best
              practices and tried to make the interface easy to develop nice
              visualizations, but getting them *just right* is often
-             time-consuming.  It's usually best to start out simple and expand 
+             time-consuming.  It's usually best to start out simple and expand
              and tweak as needed.
 
 The scene interface provides a modular interface for creating renderings
@@ -653,7 +653,7 @@ function. Example:
 
     import numpy as np
     import yt
-  
+
 
     ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
     sc = yt.create_scene(ds, 'density')
@@ -856,3 +856,28 @@ dark.  ``sigma_clip = N`` can address this by removing values that are more
 than ``N`` standard deviations brighter than the mean of your image.
 Typically, a choice of 4 to 6 will help dramatically with your resulting image.
 See the cookbook recipe :ref:`cookbook-sigma_clip` for a demonstration.
+
+.. _when_to_render:
+
+When to Render
+^^^^^^^^^^^^^^
+
+The rendering of a scene is the most computationally demanding step in
+creating a final image and there are a number of ways to control at which point
+a scene is actually rendered. The default behavior of the
+:meth:`~yt.visualization.volume_rendering.scene.Scene.save` function includes
+a call to :meth:`~yt.visualization.volume_rendering.scene.Scene.render`. This
+means that in most cases (including the above examples), after you set up your
+scene and volumes, you can simply call
+:meth:`~yt.visualization.volume_rendering.scene.Scene.save` without first
+calling :meth:`~yt.visualization.volume_rendering.scene.Scene.render`. If you
+wish to save the most recently rendered image without rendering again, set
+``render=False`` in the call to
+:meth:`~yt.visualization.volume_rendering.scene.Scene.save` and the most
+recently rendered image array will be used.  Cases where you
+may wish to use ``render=False`` include saving images at different
+``sigma_clip`` values (see :ref:`cookbook-sigma_clip`) or when saving an image
+that has already been rendered in a Jupyter notebook using
+:meth:`~yt.visualization.volume_rendering.scene.Scene.show`. Changes to the
+scene including adding sources, modifying transfer functions or adjusting camera
+settings generally require rendering again.
--- a/yt/visualization/volume_rendering/scene.py
+++ b/yt/visualization/volume_rendering/scene.py
@@ -204,6 +204,16 @@ class Scene:
         >>> sc = yt.create_scene(ds)
         >>> # Modify camera, sources, etc...
         >>> im = sc.render()
+        >>> sc.save(sigma_clip=4.0,render=False)
+
+        Altneratively, if you do not need the image array, you can just call
+        save as follows. 
+
+        >>> import yt
+        >>> ds = yt.load('IsolatedGalaxy/galaxy0030/galaxy0030')
+        >>>
+        >>> sc = yt.create_scene(ds)
+        >>> # Modify camera, sources, etc...
         >>> sc.save(sigma_clip=4.0)
 
         """
@@ -216,13 +226,13 @@ class Scene:
         self._last_render = bmp
         return bmp
 
-    def save(self, fname=None, sigma_clip=None):
+    def save(self, fname=None, sigma_clip=None, render=True):
         r"""Saves the most recently rendered image of the Scene to disk.
 
-        Once you have created a scene and rendered that scene to an image
-        array, this saves that image array to disk with an optional filename.
-        If an image has not yet been rendered for the current scene object,
-        it forces one and writes it out.
+        Once you have created a scene, this saves that image array to disk with
+        an optional filename. This function calls render() to generate an
+        image array, unless the render parameter is set to False, in which case
+        the most recently rendered scene is used if it exists.
 
         Parameters
         ----------
@@ -239,6 +249,10 @@ class Scene:
             Default: None
 
             floor(vals > std_dev*sigma_clip + mean)
+        render: boolean, optional
+            If True, will always render the scene before saving.
+            If False, will use results of previous render if it exists.
+            Default: True
 
         Returns
         -------
@@ -252,10 +266,11 @@ class Scene:
         >>>
         >>> sc = yt.create_scene(ds)
         >>> # Modify camera, sources, etc...
-        >>> sc.render()
         >>> sc.save('test.png', sigma_clip=4)
 
-        Or alternatively:
+        When saving multiple images without modifying the scene (camera,
+        sources,etc.), render=False can be used to avoid re-rendering.
+        This is useful for generating images at a range of sigma_clip values:
 
         >>> import yt
         >>> ds = yt.load('IsolatedGalaxy/galaxy0030/galaxy0030')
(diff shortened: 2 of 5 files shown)


chummels:
Thanks for writing this up and making the change.  I think this is a good solution.  

I think it would be beneficial to make an addition in the narrative docs for the Volume Rendering advertising this functionality.  Perhaps you could add a bullet here (https://yt-project.org/doc/visualizing/volume_rendering.html#volume-rendering-frequently-asked-questions) about how rendering is expensive, and how you can turn off rendering when you save with your new keyword, if the scene hasn't changed since the last render?  Or something?

Also, because the `volume_render()` function also internally calls `render()`, I think you could potentially swap out the `save()` for a `save(render=False)` call in this example in the narrative docs too: https://yt-project.org/doc/visualizing/volume_rendering.html#annotated-examples

chrishavlin:
Ok, I wasn't sure where to note the change in the narrative docs --- the FAQ makes sense! I'll add the bullet point shortly.

yt/visualization/volume_rendering/scene.py
```diff
@@ -204,6 +204,16 @@ def render(self, camera=None):
         >>> sc = yt.create_scene(ds)
         >>> # Modify camera, sources, etc...
         >>> im = sc.render()
+        >>> sc.save(sigma_clip=4.0,render=False)
+
+        Altneratively, if you do not need the image array, you can just call
```
munkm, line 209: This is a nice enhancement to the docs showing how to use .save() both ways. One way that this could be enhanced is by showing an interim operation on `im` in the example above, though I'm not positive of what a good thing to do would be here.

yt/visualization/volume_rendering/scene.py
```diff
@@ -252,19 +266,20 @@ def save(self, fname=None, sigma_clip=None):
         >>>
         >>> sc = yt.create_scene(ds)
         >>> # Modify camera, sources, etc...
-        >>> sc.render()
         >>> sc.save('test.png', sigma_clip=4)
 
-        Or alternatively:
+        When saving multiple images without modifying the scene (camera,
+        sources,etc.), render=False can be used to avoid re-rendering.
```
munkm, line 272:
```suggestion
        sources,etc.), render=False can be used to avoid re-rendering when a scene is saved.
```

yt/visualization/volume_rendering/scene.py
```diff
@@ -252,19 +266,20 @@ def save(self, fname=None, sigma_clip=None):
...
+        This is useful for generating images at a range of sigma_clip values:
 
         >>> import yt
         >>> ds = yt.load('IsolatedGalaxy/galaxy0030/galaxy0030')
         >>>
         >>> sc = yt.create_scene(ds)
         >>> # save with different sigma clipping values
         >>> sc.save('raw.png')
```
munkm, line 280:
```suggestion
        >>> sc.save('raw.png')  # The initial render call happens here
```

yt/visualization/volume_rendering/scene.py
```diff
@@ -287,7 +302,11 @@ def save(self, fname=None, sigma_clip=None):
             suffix = '.png'
             fname = '%s%s' % (fname, suffix)
 
-        self.render()
+        # in most cases we want to render every time, but in some cases pulling
```
munkm, line 305: I love the explanations here. Nice.

yt/visualization/volume_rendering/scene.py
```diff
@@ -352,17 +372,20 @@ def save_annotated(self, fname=None, label_fmt=None,
...
-           keyword/value pairs to pass through to the matplotlib text() 
+           list containing a tuple of coordinates (in normalized figure
+           coordinates), the text to display, and, optionally, a dictionary of
+           keyword/value pairs to pass through to the matplotlib text()
            function.
 
            Each item in the main list is a separate string to write.
-
+        render: boolean, optional
+            If True, will always render the scene before saving.
```
munkm, line 386:
```suggestion
            If True, will render the scene before saving.
```

chrishavlin:
OK, added a note to the FAQ. 

I didn't adjust the example in https://yt-project.org/doc/visualizing/volume_rendering.html#annotated-examples because the camera gets adjusted after the initial call to `volume_render()`, so I think it does need to re-render in this case. But I do know of a few other places where the `render=False` keyword should be added (e.g., https://yt-project.org/doc/cookbook/complex_plots.html#cookbook-sigma-clip) but as #2736 touches that file I was planning to wait until that gets merged before adjusting it. 

@munkm I just noticed your suggestions -- I'll make the changes and push up again :)

munkm:
Yeah sorry I must have been reviewing while you were making changes!!!

yt/visualization/volume_rendering/scene.py
chrishavlin: I'm not sure either. I have ideas, but they seem too limited in scope to be useful. For example, you could plot an image histogram of the RGB channels from `im` (something like this https://en.wikipedia.org/wiki/Image_histogram#/media/File:SunHistacp.jpg ?) or you could apply an image processing function like one of the exposure adjustments in https://scikit-image.org/docs/stable/api/skimage.exposure.html to the image array but both of these feel too involved (and limited in usefulness) for a docstring example?

chrishavlin:
https://yt-project.org/docs/dev/visualizing/volume_rendering_tutorial.html#volume-rendering-tutorial seems to have errors? In any case I *think* that if you call `sc.show()` after `sc.render()` as it does in the notebook, it will show the most recent rendering without rendering again (i.e., I don't think `sc.show()` needs a `render` argument like `sc.save()`). But I could add a cell block demonstrating how to save an image once you like it.

munkm:
> https://yt-project.org/docs/dev/visualizing/volume_rendering_tutorial.html#volume-rendering-tutorial seems to have errors?

Yeah the notebook issue is what's reported in yt-project/yt#2471 I think. The errors look similar. 

> But I could add a cell block demonstrating how to save an image once you like it.

Yes, sorry I should have been explicit in what I was saying above. I think it would be good adding a block or blocks demonstrating the new the save functionality in that notebook, since it already demonstrates the `show()` method but not `save()`. I'd also add a note describing how you can skip using `.render()` if you don't modify the kwarg in `.save()`. Does that make sense?

chrishavlin:
Makes sense! Will push up a modified notebook today.

chrishavlin:
Hey @munkm  I found a few spots peppered throughout the notebook that could use updating related to this PR, ended up just restructuring the whole notebook (adding headers, etc.). You can check out the rendered notebook here https://github.com/chrishavlin/yt_scratch/blob/master/notebooks/Volume_Rendering_Tutorial.ipynb think it's too much of a change for this PR? or should I open a new PR for updating the volume rendering tutorial notebook?

munkm:
I love your new notebook! I think it'll be a really nice intro to volume rendering and the flow you've constructed in it is wonderful. 

I personally think it is fine to add it to this PR -- though others should say so if they feel differently. Make sure to clear the notebook outputs before committing it to yt; the cells should get run in the docs build. I'd also update the title of the PR to include something about updating the volume rendering tutorial. e.g. `Add render option to scene.save; improve volume rendering tutorial`

doc/source/visualizing/volume_rendering.rst
```diff
@@ -856,3 +856,28 @@ dark.  ``sigma_clip = N`` can address this by removing values that are more
...
+:meth:`~yt.visualization.volume_rendering.scene.Scene.save` function includes
+a call to :meth:`~yt.visualization.volume_rendering.scene.Scene.render`. This
+means that in most cases (including the above examples), after you set up your
+scene and volumes, you can simply call
+:meth:`~yt.visualization.volume_rendering.scene.Scene.save` without first
+calling :meth:`~yt.visualization.volume_rendering.scene.Scene.render`. If you
+wish to save the most recently rendered image without rendering again, set
+``render=False`` in the call to
+:meth:`~yt.visualization.volume_rendering.scene.Scene.save` and the most
+recently rendered image array will be used.  Cases where you
```
neutrinoceros, line 877: This sentence is a bit redundant.

yt/visualization/volume_rendering/scene.py
```diff
@@ -204,6 +204,16 @@ def render(self, camera=None):
         >>> sc = yt.create_scene(ds)
         >>> # Modify camera, sources, etc...
         >>> im = sc.render()
+        >>> sc.save(sigma_clip=4.0,render=False)
+
+        Altneratively, if you do not need the image array, you can just call
+        save as follows.
```
neutrinoceros, line 210:
```suggestion
        ``save`` as follows.
```

yt/visualization/volume_rendering/scene.py
```diff
@@ -216,13 +226,13 @@ def render(self, camera=None):
         self._last_render = bmp
         return bmp
 
-    def save(self, fname=None, sigma_clip=None):
+    def save(self, fname=None, sigma_clip=None, render=True):
         r"""Saves the most recently rendered image of the Scene to disk.
```
neutrinoceros, line 230: This line should be updated, I think it has become a bit contradictory with behavior you're aiming for.

yt/visualization/volume_rendering/tests/test_save_render.py
```diff
@@ -0,0 +1,39 @@
...
+class SaveRenderTest(TestCase):
+    # This toggles using a temporary directory. Turn off to examine images.
+    use_tmpdir = True
+
+    def setUp(self):
+        if self.use_tmpdir:
+            self.curdir = os.getcwd()
+            # Perform I/O in safe place instead of yt main dir
+            self.tmpdir = tempfile.mkdtemp()
+            os.chdir(self.tmpdir)
```
neutrinoceros, line 23: I tend to see `cd`s as weak points in a test function. Could you instead save to the tmpdir directly ?

yt/visualization/volume_rendering/tests/test_save_render.py
```diff
@@ -0,0 +1,39 @@
...
+        else:
+            self.curdir, self.tmpdir = None, None
+
+    def tearDown(self):
+        if self.use_tmpdir:
+            os.chdir(self.curdir)
+            shutil.rmtree(self.tmpdir)
+
+    def test_save_render(self):
+        ds = fake_random_ds(32)
```
neutrinoceros, line 33:
```suggestion
        ds = fake_random_ds(ndims=32)
```

yt/visualization/volume_rendering/tests/test_save_render.py
```diff
@@ -0,0 +1,39 @@
...
+        sc = yt.create_scene(ds)
+        sc.save('raw.png') # will use render = True by default
+        sc.save('clip_2.png', sigma_clip=2, render=False) # will pull render
+        sc.save('clip_4.png', sigma_clip=4.0, render=False)
+
```
neutrinoceros, line 38: Shouldn't we compare the images to check they are indeed different ?

yt/visualization/volume_rendering/scene.py
```diff
@@ -287,7 +302,11 @@ def save(self, fname=None, sigma_clip=None):
             suffix = '.png'
             fname = '%s%s' % (fname, suffix)
 
-        self.render()
+        # in most cases we want to render every time, but in some cases pulling
+        # the previous render is desirable (e.g., if only changing sigma_clip or
+        # saving after a call to sc.show()).
+        if render or hasattr(self,'_last_render') is False:
```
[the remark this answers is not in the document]
chrishavlin: For point 1 - Great suggestion! Yes, adding `_last_render` as an attribute at init will make it clearer. 

For point 2 - I like adding the message. Since whether or not to call `render()` before `save()` may be confusing for new users (hopefully the updated docs help!), would it be helpful to add a second message to notify that the most recently rendered image is being used? i.e., 

```
if not render and self._last_render is None:
    mylog.warning("No previous rendered image found, rendering now.")
    render = True
else: 
    mylog.warning("Saving most recently rendered image") 
```

Or is that too much info?

yt/visualization/volume_rendering/tests/test_save_render.py
chrishavlin: I agree -- I wasn't entirely sure what exactly we should test here... so I copied over the test for `sigma_clip` as a starting point. So yes, can save to tmpdir directly.

yt/visualization/volume_rendering/scene.py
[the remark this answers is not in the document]
neutrinoceros: Not too much info imo, I like it, although it might just be an "info" level logging message, instead of a warning :)

yt/visualization/volume_rendering/tests/test_save_render.py
chrishavlin: Sure! That would make sense -- then it becomes an actual test of functionality, not just a test of keyword arguments... 

is there a way you'd recommend to do this? Load the images back in and use `yt.testing.assert_equal` or `assert_almost_equal`? 

In terms of loading the image back in -- is there a "yt" way to do it? I could use `matplotlib.pyplot.imread` but maybe there's a more appropriate method?

chrishavlin:
Thanks for the helpful comments, @neutrinoceros !

yt/visualization/volume_rendering/tests/test_save_render.py
neutrinoceros: Yes there's a "yt way" to run image comparison tests.
Unfortunately, I can't remember off the top of my head... but I know @munkm does !
In the mean time, you may want to read this https://yt-project.org/docs/dev/developing/testing.html#answer-testing

chrishavlin: Ah, yes, I need to re-read that :) `How to Write Image Comparison Tests` has a description of using `GenericImageTest`

chrishavlin: Though in this case I need to check that the 3 saved rendered images are NOT equal. But I'll read through that and figure it out.

neutrinoceros: ah yes, that's the one I meant to link, sorry !
I remember I was quite confused last time I had to do something with it, but @munkm's help was extremely useful

neutrinoceros: > Though in this case I need to check that the 3 saved rendered images are NOT equal. But I'll read through that and figure it out.

Well the easy way out is to check they are different when you first run it, store the 3 outputs and then compare against those in the future.

yt/visualization/volume_rendering/scene.py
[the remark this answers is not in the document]
chrishavlin: Since adding the `if` statement adds a few lines and those lines need to also happen in `save_annotated`, any objection to creating a new function to avoid copy/pasting the 6-7 lines, e.g., `_check_render()`? 

I'm also thinking that it may be better to further split up the logging into: 

```
        if self._last_render is None:
            mylog.warning("No previous rendered image found, rendering now.")
            render = True
        elif render:
            mylog.info("Overwriting previous rendered image with new rendering.")
        else:            
            mylog.info("Saving most recently rendered image.")
```

Maybe the second condition should be a `warning` as well? Since it is a desired behavior, `info` feels appropriate but happy to change it.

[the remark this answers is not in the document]
neutrinoceros: No objection, on the contrary. If you need to reuse code it's _much_ better to make it... reusable. :-)
I think your idea to further split this if statement is great. I may have a small preference for a warning in the `elif render` case but your choice !

yt/visualization/volume_rendering/scene.py
```diff
@@ -226,8 +226,25 @@ def render(self, camera=None):
         self._last_render = bmp
         return bmp
 
+    def _check_render(self, fname, render=True):
```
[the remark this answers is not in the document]
chrishavlin: how about `_check_then_render_missing` ? or `_render_if_missing`?

yt/visualization/volume_rendering/scene.py
```diff
@@ -226,8 +226,25 @@ def render(self, camera=None):
...
+    def _check_render(self, fname, render=True):    
+        # checks for existing render before saving,  in most cases we want to 
+        # render every time, but in some cases pulling the previous render is 
+        # desirable (e.g., if only changing sigma_clip or
+        # saving after a call to sc.show()).
+        if self._last_render is None:
+            mylog.warning("No previous rendered image found, rendering now and saving to %s", fname)
+            render = True
+        elif render:
+            mylog.info("Overwriting previous rendered image with new rendering, saving to %s", fname)
```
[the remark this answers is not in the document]
chrishavlin: OK -- good point, I added it because I thought the "saving to fname" that gets  logged was repetitive, but ya, if the call to render goes wrong it's helpful to have a separate message!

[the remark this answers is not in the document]
chrishavlin: Ok, added back the separate Save message log

yt/visualization/volume_rendering/scene.py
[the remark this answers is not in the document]
chrishavlin: went with `_render_if_missing` :)

chrishavlin:
Just to summarize to this point: the outstanding change is to add an appropriate answer test.

yt/visualization/volume_rendering/scene.py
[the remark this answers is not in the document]
neutrinoceros: > If your function's name has conjunctions, maybe it's time to refactor

Can't remember where I heard that exactly but that's why I can't find a good name for this !
In particular `_render_if_missing` feels wrong since it can also render if passed `render=True`...

I'm thinking the logic could be slightly changed so that the function would be named `_sanitize_render_switch` and called early in `save()` and `save_annotated()`, leaving the actual rendering part to those functions (which would also make them clearer, even if this implies repeating 2 lines). What do you think ?

yt/visualization/volume_rendering/scene.py
```diff
@@ -407,7 +447,8 @@ def save_annotated(
             suffix = ".png"
             fname = "%s%s" % (fname, suffix)
 
-        self.render()
+        self._render_if_missing(fname, render)
```
neutrinoceros, line 450: this line is responsible for the failing tests: `fname` shouldn't be passed to this function anymore

chrishavlin: yup! caught that last night, but forgot to push up the commit to fix it. Will do that now...

chrishavlin: Ok, it's up. Includes the change to using `_sanitize_render`. Though I didn't make it a static function like you mentioned -- wasn't sure since it requires checking for `self._last_render`. If it were static, wouldn't the function be built on initialization when `self._last_render=None`?

neutrinoceros: yeah I was clearly tired, I went back and reverted that suggestion on slack 😅

matthewturk:
Is this one ready to go?  It would be great to get the time down for the tests.  :)

yt/visualization/volume_rendering/scene.py
```diff
@@ -216,13 +226,31 @@ def render(self, camera=None):
         self._last_render = bmp
         return bmp
 
-    def save(self, fname=None, sigma_clip=None):
-        r"""Saves the most recently rendered image of the Scene to disk.
+    def _sanitize_render(self, render=True):
```
neutrinoceros, line 229: one last comment maybe: I don't think it's useful or worth it to have a default value here

chrishavlin:
@matthewturk the one remaining thing to do is add an answer test for this change (I've only added a unit test so far). In terms speeding up the tests, #2736 will actually do that as well (and my changes here won't conflict or break that PR).

chrishavlin:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

PR #2744, opened 2020-07-14
fix: min/max detection consistency

neutrinoceros:
## PR Summary

This fix inconsistencies between `Dataset.find_min` and `Dataset.find_max` methods by making special cases of a new, more general class method `Dataset._find_extremum`.

## Fix

For non-cartesian datasets, `find_min` didn't contain the "hack" from `find_max` to handle dimensionless coords
```python
import yt
ds = yt.load("amrvac/bw_2d0000.dat") # polar dataset
ds.find_max("density")
>> (unyt_quantity(6.9545098e-24, 'g/cm**3'), unyt_array([1.5703125 , 1.71756761, 0.5       ], 'code_length'))

# inconsistency for find_min
ds.find_min("density")
old >> IterableUnitCoercionError: Received a list or tuple of quantities with nonuniform units: [unyt_quantity(0.9921875, 'cm'), unyt_quantity(3.16613635, '(dimensionless)'), unyt_quantity(0.5, 'cm')]
new >> (unyt_quantity(2.82470365e-25, 'g/cm**3'), unyt_array([2.0078125 , 1.55950623, 0.5       ], 'code_length'))
```

## new feature

I expose two new arguments to the user for the following reasons
- maybe the user does *not* want a `unyt_array` typed output, especially with non-cartesian datasets.
- the "source" (object) in which the search is performed could be anything, it doesn't have to be `ds.all_data()`
```python
# don't convert to "code_lenght"
ds.find_max("density", to_array=False)
>> (unyt_quantity(6.9545098e-24, 'g/cm**3'),
 [unyt_quantity(1.5703125, 'cm'),
  unyt_quantity(1.71756761, '(dimensionless)'),
  unyt_quantity(0.5, '(dimensionless)')])

# region selection
region = ds.r[1:1.2, :]
ds.find_max("density", source=region, to_array=True)
>> (unyt_quantity(2.51770532e-24, 'g/cm**3'), unyt_array([1.0234375, 0.159534 , 0.5      ], 'code_length'))
```

## PR Checklist

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -895,6 +895,32 @@ class Dataset(metaclass=RegisteredDataset):
         setattr(self, name, obj)
 
     def _find_extremum(self, field, ext, source=None, to_array=True):
+        """
+        Find the extremum value of a field in a data object (source) and its position.
+
+        Parameters
+        ----------
+        field: str or tuple(str, str)
+
+        ext: str
+            'min' or 'max', select an extremum
+
+        source: a Yt data object
+
+        to_array: bool
+            select the return type.
+
+        Returns
+        -------
+        val, coords
+
+        val: unyt.unyt_quantity
+            extremum value detected
+
+        coords: unyt.unyt_array or list(unyt.unyt_quantity)
+            Conversion to a single unyt_array object is only possible for coordinate
+            systems with homogeneous dimensions across axes (i.e. cartesian).
+        """
         ext = ext.lower()
         if source is None:
             source = self.all_data()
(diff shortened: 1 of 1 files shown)


neutrinoceros:
note that unfortunatly this *doesn't* resolve #2446

yt/data_objects/static_output.py
```diff
@@ -871,47 +871,41 @@ def _add_object_class(self, name, base):
...
+        mylog.info("%s value is %0.5e at %0.16f %0.16f %0.16f", ext, val, *coords)
+        if to_array:
+            # This is a hack to fix the fact that some non-cartesian datasets have
+            # dimensionless quantities, and we can't yet handle that.
+            alt_coords = []
+            for x in coords:
+                alt_coords.append(
+                    self.quan(x.v, "code_length") if x.units.is_dimensionless else x
+                )
+            coords = self.arr(alt_coords, dtype="float64").to("code_length")
```
cphyc, line 893: This looks a bit bizarre to me. If a user calls the method with `to_array` in non-cartesian coordinates, why don't we throw an error or issue a warning and disable `to_array`?

cphyc: Oh, I just read the deleted code and realize this was the already-existing behavior. Anyways, I would argue _against_ this, since it will silently return an array with incorrect units in the non-cartesian directions...

neutrinoceros: I'm not a big fan either. Maybe we should throw an error instead (or maybe just a runtime warning ?) in the non-cartesian cases and warn the user they should set `to_array=False`... wdyt ?

cphyc: I believe throwing an error encouraging to use `to_array` would be better, as most yt user I encountered do not even read the warnings anyway.

neutrinoceros:
@yt-fido test this please

yt/data_objects/static_output.py
neutrinoceros: I added an error and docstrings !

neutrinoceros: Ah, of course this is breaking a lot of tests on Jenkins... which probably means potentially a lot of downstream code... @cphyc, @matthewturk  do you think this is worth the break ?

munkm:
@yt-fido test this please

neutrinoceros:
This is still going to fail !
These errors are "designed" here since I added a `raise` statement that's hit by existing tests. This is discussed above but I'm waiting on @cphyc and/or @matthewturk (or any one really) to answer the fundamental question here: do you guys think this is worth the breakage ? I'd personally prefer we just raise a runtime warning here and deal with this non-cartesian VS cartesian issue on a broader level later. WDYT ?

munkm:
Sorry I couldn't see the test result from the last failure, so I wasn't sure if it was a leftover relic from master failing a while back. I decided to rerun. Ignore that!

yt/data_objects/static_output.py
cphyc: If that's breaking a lot of downstream code, I'm also ok with issuing a warning.

neutrinoceros: adressed !

PR #2745, opened 2020-07-14
[AMRVAC] fix: improve read_amrvac_namelist api 

neutrinoceros:
## PR Summary

automatically expand '~' token… and natively support pathlike objects

## PR Checklist

### DIFF
--- a/yt/frontends/amrvac/io.py
+++ b/yt/frontends/amrvac/io.py
@@ -4,13 +4,51 @@ AMRVAC-specific IO functions
 
 
 """
+import os
 
 import numpy as np
 
+from yt.funcs import ensure_list
 from yt.geometry.selection_routines import GridSelector
 from yt.utilities.io_handler import BaseIOHandler
+from yt.utilities.on_demand_imports import _f90nml as f90nml
 
 from .datfile_utils import get_single_block_field_data
+
+
+def read_amrvac_namelist(parfiles):
+    """Read one or more parfiles, and return a unified f90nml.Namelist object.
+
+    This function replicates the patching logic of MPI-AMRVAC where redundant parameters
+    only retain last-in-line values EXCEPT `&filelist:base_filename`, which is accumulated.
+    When passed a single file, this function acts as a mere wrapper of f90nml.read().
+
+    Parameters
+    ----------
+    parfiles : str or list
+        A file path, or a list of file paths to MPI-AMRVAC configuration parfiles.
+
+    Returns
+    -------
+    unified_namelist : f90nml.Namelist
+        A single namelist object. The class inherits from ordereddict.
+
+    """
+    parfiles = [os.path.expanduser(pf) for pf in ensure_list(parfiles)]
+
+    # first merge the namelists
+    namelists = [f90nml.read(parfile) for parfile in parfiles]
+    unified_namelist = f90nml.Namelist()
+    for nml in namelists:
+        unified_namelist.patch(nml)
+
+    # accumulate `&filelist:base_filename`
+    base_filename = "".join(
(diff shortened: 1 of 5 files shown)


yt/frontends/amrvac/io.py
```diff
@@ -4,15 +4,54 @@
...
+    """Read one or more parfiles, and return a unified f90nml.Namelist object.
+
+    This function replicates the patching logic of MPI-AMRVAC where redundant parameters
+    only retain last-in-line values, with the exception of `&filelist:base_filename`,
+    which is accumulated. When passed a single file, this function acts as a mere
+    wrapper of f90nml.read().
+
+    Parameters
+    ----------
+    parfiles : str, os.Pathlike, byte, or an iterable returning those types
```
munkm, line 29: Is there an iterable that isn't a list that's supported here? If not, I'd be explicit and change this to list.

yt/frontends/amrvac/tests/test_read_amrvac_namelist.py
```diff
@@ -1,13 +1,19 @@
 from copy import deepcopy
 from os import path
```
munkm, line 2: just a note that I missed last time. it's recommended in the documentation of `path` that 

> Instead of importing this module directly, import os and refer to
this module as os.path.

yt/frontends/amrvac/tests/test_read_amrvac_namelist.py
```diff
@@ -1,13 +1,19 @@
...
 test_dir = path.dirname(path.abspath(__file__))
-blast_wave_parfile = path.join(test_dir, "sample_parfiles/bw_3d.par")
-modifier_parfile = path.join(test_dir, "sample_parfiles/tvdlf_scheme.par")
+blast_wave_parfile = path.join(test_dir, "sample_parfiles", "bw_3d.par")
+modifier_parfile = path.join(test_dir, "sample_parfiles", "tvdlf_scheme.par")
+
+
+@requires_module("f90nml")
+def test_read_pathlike():
+    read_amrvac_namelist(Path(blast_wave_parfile))
```
munkm, line 16: Wouldn't this test be closer to user behavior if it wasn't passed through `Path`?

yt/frontends/amrvac/io.py
neutrinoceros: A numpy array or a tuple in particular should work already. Though I can't think of any other useful utterable, I think ALL iterables should be supported already, given the way the first line is written.

yt/frontends/amrvac/tests/test_read_amrvac_namelist.py
neutrinoceros: Do you mean something like this ?
```python
    filepath = Path(blast_wave_parfile)
    read_amrvac_namelist(file path)
```

munkm: I was meaning that I think a user is more likely to do something like `read_amrvac_namelist(blast_wave_parfile)` without using `Path` at all.

neutrinoceros: Of course they are, but this test exist solely to check that they (... I 🙈) can do it.

yt/frontends/amrvac/tests/test_read_amrvac_namelist.py
neutrinoceros: done !

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

PR #2756, opened 2020-07-17
Handful of isort changes and fixes

matthewturk:
This reformats a few of the files that were still failing, and it
changes the setup.cfg section to be isort rather than tool:isort, which
seems to be the right thing for isort 4.3.


yt/visualization/volume_rendering/input_events.py
```diff
@@ -3,11 +3,11 @@
 import random
 from collections import defaultdict, namedtuple
 
+import cyglfw3 as glfw
 import matplotlib.cm as cm
 import numpy as np
-
-import cyglfw3 as glfw
 import OpenGL.GL as GL
```
neutrinoceros, line 9: might be worth adding OpenGL and cyglfw3 to the list of known third party in setup.cfg

.travis.yml
```diff
@@ -84,7 +84,7 @@ jobs:
 
     - stage: Lint
       python: 3.6
-      script: isort --check-only -rc yt/
+      script: isort --check-only -rc
```
neutrinoceros, line 87: while this may be correct (not 100% sure), it think this would be clearer
```suggestion
      script: isort --check-only -rc .
```

neutrinoceros:
So @cphyc  showed that there are yet again inconsistent results with isort on travis (see https://github.com/yt-project/yt/pull/2759)
This is because I ran isort 4.3.21 locally, while travis is running 4.3.0 (I didn't pin the last digit and didn't anticipate pip would read it as 4.3.0). Meanwhile, I've also tried the latest release 5.1.2 (only a few hours old !) and it fixes even more bugs so I think we should simply pin this one.

note that, starting from isort 5.0, the `-rc` (recursive) option is implicit so it can be removed.
@matthewturk , do you agree to make those infrastructure changes in this PR ? We can open a follow up PR with the new isort pass afterwards.
here's what's needed: 
- update `tests/lint_requirements.txt` to pin the latest version
- update `.travis.yml` and `.github/PULL_REQUEST_TEMPLATE.md` (remove the `-rc` flag)

matthewturk:
Went ahead and updated as per your suggestion!

setup.cfg
```diff
@@ -23,6 +23,7 @@ exclude = doc,
 
           # individual files
           yt/visualization/_mpl_imports.py,
+          yt/utilities/command_line.py,
```
neutrinoceros, line 26: you should also remove the inline comment in that file (might need to rebase this branch :/)

matthewturk:
So, unfortunately, I'm struggling to get this to all work.  Can we take this as-is and remove the comment in another PR?

matthewturk:
(As in, specifically, getting the git commit hook to work with the version of isort that also works here etc etc)

neutrinoceros:
> So, unfortunately, I'm struggling to get this to all work. Can we take this as-is and remove the comment in another PR?

sure, works for me

neutrinoceros:
note that linting is expected to fail here, and is fixed by the companion PR #2764

neutrinoceros:
In the end it turns out this branch isn't necessary after tests passed in https://github.com/yt-project/yt/pull/2764, so as we agreed on slack I'm closing this :)

neutrinoceros:
(turns out there's no "close" button on the mobile app yet !)

PR #2760, opened 2020-07-19
cleanup: wrapped exceptions

neutrinoceros:
## PR Summary

Pylint is now able to detect places from `raise Exception from parent_exception` pattern would be meaningful, namely when exception are wrapped into more friendly ones.

Here is the difference between the two patterns:
```python
try:
    assert(False)
except AssertionError:
    raise FriendlyError
# error in handling >> During handling of the above exception, another exception occurred:

try:
    assert(False)
except AssertionError as e:
    raise FriendlyError from e
# The above exception was the direct cause of the following exception:
```
The second indicates more clearly what's actually going on and reflects the developer's intention best in almost every cases.

Generating the report
```bash
pip install git+https://github.com/PyCQA/pylint
pylint yt | grep W707
```

Additionally pylint detects "too broad exceptions" (W703), notably where we simply use `except Exception`, which in most cases is an overkill. Not going to fix all of them here, only the ones that also fall in W707

### DIFF
--- a/yt/fields/xray_emission_fields.py
+++ b/yt/fields/xray_emission_fields.py
@@ -216,11 +216,11 @@ def add_xray_emissivity_field(
     if not isinstance(metallicity, float) and metallicity is not None:
         try:
             metallicity = ds._get_field_info(*metallicity)
-        except YTFieldNotFound:
+        except YTFieldNotFound as e:
             raise RuntimeError(
                 f"Your dataset does not have a {metallicity} field! "
                 + "Perhaps you should specify a constant metallicity instead?"
-            )
+            ) from e
 
     if table_type == "cloudy":
         # Cloudy wants to scale by nH**2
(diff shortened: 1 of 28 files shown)


yt/fields/xray_emission_fields.py
```diff
@@ -327,16 +327,17 @@ def _photon_emissivity_field(field, data):
...
-            else:
+            try:
+                # normal behaviour, if dist is a YTQuantity
                 dist = ds.quan(dist.value, dist.units)
+            except AttributeError as e:
+                try:
+                    dist = ds.quan(*dist)
+                except (RuntimeError, TypeError):
+                    raise TypeError(
+                        "dist should be a YTQuantity " "or a (value, unit) tuple!"
```
cphyc, line 338:
```suggestion
                        "dist should be a YTQuantity or a (value, unit) tuple!"
```

PR #2762, opened 2020-07-19
cleanup: legacy python doc

neutrinoceros:
## PR Summary
complementary PR to https://github.com/yt-project/yt/pull/2641, targeted at docs !
⚠️ also update `setup.py` to explicitly require python 3.6 (note that it currently accepts python 3.5)


benchmarks/asv.conf.json
```diff
@@ -25,7 +25,7 @@
 
     // The Pythons you'd like to test against.  If not provided, defaults
     // to the current version of Python used to run `asv`.
-    // "pythons": ["2.7", "3.3"],
+    // "pythons": ["3.7", "3.8"],
```
cphyc, line 28: I know this isn't you... but I thought comments were illegal in json?

cphyc: Also it may be worth noting that the file doesn't seem to have survived the transition from mercurial to git(hub) (see https://github.com/yt-project/yt/commit/b361c8e03ae4a50bb8868d4b4ff831b1caadbab3). @matthewturk do you think it would be worth reviving asv?

neutrinoceros: oh you're right, no comments in json... well that explains the "weird" syntax highlighting on GH 😅

matthewturk: I have thought it was useful in the past, but I don't know anymore; we also never got it actually running propertly.  Let's leave it for now and then later we can decide in a separate PR.

matthewturk:
Has anybody tried installing with this new script?

PR #2763, opened 2020-07-19
cleanup: remove a class that was only used in python 2

neutrinoceros:
## PR Summary
Unless I'm misreading its docstring, this class isn't required anymore.


matthewturk:
We may also no longer need `poster` either.

neutrinoceros:
I'm fine with removing the whole module. Should I ?

neutrinoceros:
switching to draft, I'll try removing the whole module later and see what happens.

neutrinoceros:
This is superseded by #2783

PR #2749, opened 2020-07-16
YTEP0037 application: transition to black + isort

neutrinoceros:
This contains the actual transition to isort + black after https://github.com/yt-project/yt/pull/2592 and https://github.com/yt-project/yt/pull/2596 were merged with all the required configuration.

PR #2764, opened 2020-07-19
second isort pass with isort 5.1.4

neutrinoceros:
## PR Summary

Companion PR to #2756 and follow up to #2749 : second pass with the latest version of isort, which fixes some bugs on their side and now seamlessly runs on Cython files


neutrinoceros:
right, I should update it here too to prove it works !

matthewturk:
I think the cimport sorting might have broken the build.  I'm going to try taking a crack at this and let you know.

neutrinoceros:
> I think the cimport sorting might have broken the build. I'm going to try taking a crack at this and let you know.

wow, may be worth reporting to isort's author if this is indeed the case.

matthewturk:
@neutrinoceros I suspect it's because of a sneaky transitive cimport issue -- probably not worth reporting.

PR #2767, opened 2020-07-20
cleanup: replace AssertionError raise statements where undue

neutrinoceros:
## PR Summary

quoting [Python's official documentation](https://docs.python.org/3/library/exceptions.html#AssertionError)
> exception AssertionError
Raised when an assert statement fails.

This PR replaces a handful of `raise AssertionError` found in the code base with more appropriate errors (namely `TypeError` and `ValueError`). It also adds a deprecation warning for `SlicePlot`'s argument `axis`.

Note that I purposely left untouched a few `raise AssertionError` found in tests/testing framework because I _think_ it may be ok to use it there.

## PR Checklist

- [x] pass `flake8 yt/`
- [x] pass `isort -rc . --check-only`

### DIFF
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -2352,17 +2352,23 @@ def SlicePlot(ds, normal=None, fields=None, axis=None, *args, **kwargs):
     ...                 north_vector=[0.2,-0.3,0.1])
 
     """
-    # Make sure we are passed a normal
-    # we check the axis keyword for backwards compatibility
-    if normal is None:
+    if axis is not None:
+        issue_deprecation_warning(
+            "SlicePlot's argument 'axis' is a deprecated alias for 'normal', it "
+            "will be removed in a future version of yt."
+        )
+        if normal is not None:
+            raise TypeError(
+                "SlicePlot() received incompatible arguments 'axis' and 'normal'"
+            )
         normal = axis
-    if normal is None:
-        raise AssertionError("Must pass a normal vector to the slice!")
 
-    # to keep positional ordering we had to make fields a keyword; make sure
-    # it is present
+    # to keep positional ordering we had to make 'normal' and 'fields' keywords
+    if normal is None:
+        raise TypeError("Missing argument in SlicePlot(): 'normal'")
+
     if fields is None:
-        raise AssertionError("Must pass field(s) to plot!")
+        raise TypeError("Missing argument in SlicePlot(): 'fields'")
 
     # use an AxisAlignedSlicePlot where possible, e.g.:
     # maybe someone passed normal=[0,0,0.2] when they should have just used "z"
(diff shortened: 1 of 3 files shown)


yt/visualization/plot_window.py
```diff
@@ -2268,17 +2268,23 @@ def SlicePlot(ds, normal=None, fields=None, axis=None, *args, **kwargs):
...
 
     """
-    # Make sure we are passed a normal
-    # we check the axis keyword for backwards compatibility
-    if normal is None:
+    if axis is not None:
+        issue_deprecation_warning(
+            "SlicePlot's argument 'axis' is a deprecated alias for 'normal', it "
+            "will be removed in a future version of yt."
+        )
```
matthewturk, line 2275: Hmmm.  I have mixed feelings about this.

neutrinoceros: What is wrong with it ? I feel like I'm missing something entirely.

matthewturk: Well, I'll say that it's not your fault, I just have mixed feelings with encouraging folks to supply a normal vector.  I know that internally it's the same, but I think we made it that way so that you could swap out `"x"` and `[0.2, 0.3, 0.4]` or something without changing the argument name.

neutrinoceros: So you mean this argument should be positional ? In python 3.8 positional-only arguments are a thing, but in the mean time, a different approach could be to remove the default value for normal, making it mandatory and encouraging the positional-arg syntax. The trade-off here is that it can not be done while also supporting backward compatibility.
While I'm all about removing dead code where I can, it was never signaled to users that the old behaviour was discouraged/deprecated, hence my choice of going for a smooth deprecation path. What do you think ?

cphyc: If I read the code correctly, it still is possible to call `yt.SlicePlot(ds, whatever, 'density')` with `whatever='x'` or `whatever=[0.2,0.3,0.4]`. @neutrinoceros am I wrong?

neutrinoceros: Absolutely ! (no, you're not wrong)
I'm only trying to handle the case where the keyword `axis` is used instead of `normal`, but for scripts that write them positionally, it shouldn't change anything !

neutrinoceros:
@fido test this please

neutrinoceros:
@yt-fido test this please

yt/visualization/plot_window.py
cphyc: So I'm ok with the way it is coded right now. @matthewturk is it good to go for you? I'm ok to merge

matthewturk:
same

neutrinoceros:
So this is good to go then ?

yt/visualization/plot_window.py
matthewturk: lgtm

matthewturk:
I *think* this needs three approvals to merge, right?

PR #2777, opened 2020-07-23
update pre-commit hooks

neutrinoceros:
## PR Summary

Something I did not consider when I added the pre-commit hooks is the order in which they trigger :
sometimes `flake8` errors are automatically fixed by `black` or `isort` so it's cleaner to run the `flake8` check last.

Note to early adopters: once this goes in, you won't need to reinstall the hooks to take advantage of this change :)

I'm also changing the PR template to reflect this order (may save a few minutes of work to contributors if they don't realise most of there flake8 errors can be solved automatically).

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@
     - id: black
       language_version: python3.7
 -   repo: https://github.com/timothycrosley/isort
-    rev: '5.1.4'  # keep in sync with tests/lint_requirements.txt
+    rev: '5.2.0'  # keep in sync with tests/lint_requirements.txt
     hooks:
     -   id: isort
 -   repo: https://gitlab.com/pycqa/flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,13 @@
--   repo: https://gitlab.com/pycqa/flake8
-    rev: ''
-    hooks:
-    -   id: flake8
--   repo: https://github.com/timothycrosley/isort
-    rev: ''
-    hooks:
-    -   id: isort
 -   repo: https://github.com/ambv/black
     rev: 19.10b0
     hooks:
     - id: black
       language_version: python3.7
+-   repo: https://github.com/timothycrosley/isort
+    rev: ''
+    hooks:
+    -   id: isort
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,13 +1,13 @@
--   repo: https://gitlab.com/pycqa/flake8
-    rev: ''
+-   repo: https://github.com/ambv/black
+    rev: 19.10b0
```
munkm, line 2: so when a stable version of this release is out do we plan to bump it? (I know the versioning isn't changing in this PR, but I'm just curious!).

neutrinoceros: I guess we should at some point. Although that could again have large scale effects on the code base, since Black devs state the style _might_ change in the future (not in any major way though).

munkm: Yeah, I think when there's a stable release we should bump to that, since I don't think it's that good if we have multiple beta releases (or bleeding edge packages) needed to do dev work on yt. But we can deal with that (and the cascading effects) later. 🙂

neutrinoceros: I agree !

neutrinoceros: In the mean time I'm watching all repos of tools related to YTEP0037, I get notifications when they do releases :)

munkm:
@yt-fido test this please

neutrinoceros:
Sorry for pushing post-approval but I figured I might as well pin the isort and flake8 versions in the hooks to match those we have in `lint_requirements.txt` since I've witness some cases where isort `5.1.4` would disagree with the hook. It's probably due for another release soon but until then it's best to keep our config self-consistent.

neutrinoceros:
> It's probably due for another release soon but until then it's best to keep our config self-consistent.

Well isort 5.2.0 was just released and contains even more bug fixes, but its output is perfectly consistent with 5.1.4 for yt, so no reason to wait, I upgraded the pinned version here.

neutrinoceros:
Nope, good to go !

neutrinoceros:
(_but_ I'll still be bumping here as long as it's not merged and isort's author keeps releasing bug fixes !)

PR #2782, opened 2020-07-24
Remove some redundancy in VR tests

Xarthisius:
## PR Summary

This PR modifies the tests to avoid situation where code paths are tested more than once, e.g. it's sufficient to rotate vr scene once instead of 5 times to check if it works. Same goes for rendering the same scene over and over again for each possible image suffix...

### DIFF
--- a/yt/visualization/volume_rendering/tests/test_scene.py
+++ b/yt/visualization/volume_rendering/tests/test_scene.py
@@ -69,17 +69,14 @@ class RotationTest(TestCase):
         mi_bound = ((ma - mi) * (0.10)) + mi
         ma_bound = ((ma - mi) * (0.90)) + mi
         tf.map_to_colormap(mi_bound, ma_bound, scale=0.01, colormap="Reds_r")
-        sc.render()
-        for suffix in ["png", "eps", "ps", "pdf"]:
-            fname = "test_scene.{}".format(suffix)
-            sc.save(fname, sigma_clip=6.0)
-            assert_fname(fname)
+        fname = "test_scene.pdf"
+        sc.save(fname, sigma_clip=6.0)
+        assert_fname(fname)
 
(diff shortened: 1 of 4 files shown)


yt/visualization/volume_rendering/tests/test_scene.py
```diff
@@ -69,17 +69,14 @@ def test_rotation(self):
...
-            fname = "test_scene.{}".format(suffix)
-            sc.save(fname, sigma_clip=6.0)
-            assert_fname(fname)
-
-        nrot = 2
-        for i in range(nrot):
-            sc.camera.pitch(2 * np.pi / nrot)
-            sc.render()
-            sc.save("test_rot_%04i.png" % i, sigma_clip=6.0)
+        fname = "test_scene.pdf"
```
matthewturk, line 72: do we even need the PDF one?  I think we should just do png.

yt/visualization/volume_rendering/tests/test_varia.py
```diff
@@ -63,11 +63,8 @@ def test_multiple_fields(self):
     def test_rotation_volume_rendering(self):
         im, sc = yt.volume_render(self.ds)
 
-        angle = 2 * np.pi
-        frames = 4
-        for i in range(frames):
-            sc.camera.yaw(angle / frames)
-            sc.render()
+        sc.camera.yaw(np.pi)
```
matthewturk, line 66: This one I'm sort of -0 on removing.

munkm: Do you think it helps test something that isn't in other tests? Or?

matthewturk: Nah, it's probably fine. I'm just a bit reactionary.

PR #2783, opened 2020-07-25
remove poster module

neutrinoceros:
##  PR Summary

Looks like the poster module is unused and for good reasons: its functionalities are now superseded by `requests`.
Trying to simply remove it and see if anything fails.


neutrinoceros:
So apparently the `MinimalRepresentation.upload` method (indirectly used is several data objects) relies on `poster` functionalities that were not ported to python 3... looks like rotten code to me, so maybe I should also remove this completely.

neutrinoceros:
I also ended up remove a some `MinimalRepresentation` methods because they were using `pickle` to store and load data, which is unsafe to do in general (see https://nedbatchelder.com/blog/202006/pickles_nine_flaws.html)

yt/utilities/minimal_representation.py
```diff
@@ -165,64 +148,13 @@ def restore(self, storage, ds):
...
-                return
-            else:
-                raise ex
-        uploader_info = json.loads(rv)
-        new_url = url + "/handler/%s" % uploader_info["handler_uuid"]
-        for i, (cn, cv) in enumerate(chunks):
-            f = TemporaryFile()
-            np.save(f, cv)
-            f.seek(0)
-            pbar = UploaderBar("%s, % 2i/% 2i" % (self.type, i + 1, len(chunks)))
```
munkm, line 209: This is the only place in the code that the UploaderBar class is used. Maybe that should be removed too?

neutrinoceros: nice catch ! Removed it !

PR #2787, opened 2020-07-26
boxlib/amrex parameter parsing: periodicity

ax3l:
AMReX' `geometry.is_periodic` is an optional input parameter that defaults to `False` per direction.

This fixes a crash that makes this parameter non-optional in post-processing with `yt`:
```
File .local/lib/python3.6/site-packages/yt/frontends/boxlib/data_structures.py", line 1571, in _parse_parameter_file
    is_periodic = self.parameters['geometry.is_periodic'].split()
KeyError: 'geometry.is_periodic'
```

~~This also changes the default padding to `False` as is the meaning in AMReX.~~

CC @atmyers for potential review :)

Refs.:
- https://amrex-codes.github.io/amrex/docs_html/InputsProblemDefinition.html
- https://travis-ci.com/github/ECP-WarpX/WarpX/jobs/364923589

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -1650,13 +1650,10 @@ class WarpXDataset(BoxlibDataset):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        is_periodic = []
+        self.periodicity[:] = False
         if "geometry.is_periodic" in self.parameters:
             is_periodic = self.parameters["geometry.is_periodic"].split()
-        periodicity = [bool(val) for val in is_periodic]
-        for _ in range(self.dimensionality, 3):
-            periodicity += [False]  # pad to 3D
-        self.periodicity = ensure_tuple(periodicity)
+            self.periodicity[: len(is_periodic)] = [val == "1" for val in is_periodic]
 
         particle_types = glob.glob(self.output_dir + "/*/Header")
         particle_types = [cpt.split(os.sep)[-2] for cpt in particle_types]
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -1650,7 +1650,7 @@ class WarpXDataset(BoxlibDataset):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        self.periodicity[:] = False
+        self.periodicity = (False, False, False)
         if "geometry.is_periodic" in self.parameters:
             is_periodic = self.parameters["geometry.is_periodic"].split()
             self.periodicity[: len(is_periodic)] = [val == "1" for val in is_periodic]
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -1650,10 +1650,14 @@ class WarpXDataset(BoxlibDataset):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        self.periodicity = (False, False, False)
-        if "geometry.is_periodic" in self.parameters:
-            is_periodic = self.parameters["geometry.is_periodic"].split()
-            self.periodicity[: len(is_periodic)] = [val == "1" for val in is_periodic]
+        self.periodicity = [True, True, True]
+        #    note: non-periodicity currently cause problems in volume rendering
+        # self.periodicity = [False, False, False]  # AMReX default
+        # if "geometry.is_periodic" in self.parameters:
+        #     is_periodic = self.parameters["geometry.is_periodic"].split()
+        #     self.periodicity[: len(is_periodic)] = [
+        #         val == "1" for val in is_periodic]
+        self.periodicity = ensure_tuple(self.periodicity)
 
         particle_types = glob.glob(self.output_dir + "/*/Header")
         particle_types = [cpt.split(os.sep)[-2] for cpt in particle_types]
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -1650,10 +1650,12 @@ class WarpXDataset(BoxlibDataset):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        is_periodic = self.parameters["geometry.is_periodic"].split()
+        is_periodic = []
+        if 'geometry.is_periodic' is self.parameters:
+            is_periodic = self.parameters["geometry.is_periodic"].split()
         periodicity = [bool(val) for val in is_periodic]
         for _ in range(self.dimensionality, 3):
-            periodicity += [True]  # pad to 3D
+            periodicity += [False]  # pad to 3D
         self.periodicity = ensure_tuple(periodicity)
 
         particle_types = glob.glob(self.output_dir + "/*/Header")
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -1651,7 +1651,7 @@ class WarpXDataset(BoxlibDataset):
 
         # set the periodicity based on the integer BC runtime parameters
         is_periodic = []
-        if 'geometry.is_periodic' is self.parameters:
+        if "geometry.is_periodic" in self.parameters:
             is_periodic = self.parameters["geometry.is_periodic"].split()
         periodicity = [bool(val) for val in is_periodic]
         for _ in range(self.dimensionality, 3):


yt/frontends/boxlib/data_structures.py
```diff
@@ -1650,10 +1650,12 @@ def _parse_parameter_file(self):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        is_periodic = self.parameters["geometry.is_periodic"].split()
+        is_periodic = []
+        if 'geometry.is_periodic' is self.parameters:
```
neutrinoceros, line 1654:
```suggestion
        if "geometry.is_periodic" in self.parameters:
```

neutrinoceros:
Hi @ax3l , thank you for this PR !
While the change is fine (except maybe for the typo I commented above), there's a more "ytic" way to implement this. That's because `self.periodicity` already exists as part of the base class and isn't a simple list, it's what we call a `MutableAttribute`, so you can simply update its values.
Here's a suggestion on how to rewrite this

```python
    self.periodicity[:] = False
    try:
        is_periodic = self.parameters["geometry.is_periodic"].split()
        self.periodicity[:len(is_periodic)] = [bool(val) for val in is_periodic]
    except KeyError:
        pass
```

As a note, and this is not your change, but I find it a bit weird how the values are read... since we're using the `.split()` method I'm guessing `self.parameters["geometry.is_periodic"]` is a string ? but then the way it is mapped to booleans is completely broken since in Python, any non-empty string maps to `True`

```python
In [1]: bool("True")
Out[1]: True

In [2]: bool("False")
Out[2]: True
```
So unless I'm missing something this is _another_ bug in this logic. Can you confirm this hunch ? Thanks again !

yt/frontends/boxlib/data_structures.py
neutrinoceros: the `'` VS `"` stuff is just to satisfy `black`, but the `is -> in` looks like a serious typo (and it's what's causing flake8 errors)

zingale:
one thing to note: yt doesn't implement all the BC types that we use in AMReX, and for a lot of operations, yt crashes unless we force the dataset to be periodic.

neutrinoceros:
> yt crashes unless we force the dataset to be periodic.

Yes, I'm guessing this kind of problem will sundenly become apparent after we fix the bug I caught in my comment !

ax3l:
@neutrinoceros thank you for the review!

Hm, I usually refrain from try-catch constructs for things one can test, but we could still default-init with the `self.periodicity[:] = False`?

You are right, the value parsing is interesting... In practice, people put 0 or 1, but this is still parsed to `True` as strings:
```py
In [1]: bool(0)
Out[1]: False

In [2]: bool(1)
Out[2]: True

In [3]: bool("1")
Out[3]: True

In [4]: bool("0")
Out[4]: True
```

neutrinoceros:
> Hm, I usually refrain from try-catch constructs for things one can test

I agree that it's rather counter-intuitive but I respectfully insist that `try` is worth the shot here since it avoid repeating `self.parameters["geometry.is_periodic"]`. More generally, there are several reasons why "trying" should be preferred to "checking" in Python, see for instance https://www.geeksforgeeks.org/eafp-principle-in-python/ 
This pattern is known as EAFP (Easier to Ask for Forgiveness than Permission) and is encouraged by the Python langage documentation (see EAFP entry here https://docs.python.org/3/glossary.html).
I've personally found that this generally makes for less repetitive and more maintainable code.

Note that your latest update is almost certainly broken since `is_periodic` is used before it's defined in the case where `"geometry.is_periodic"` isn't present !

> Thank you for the note @zingale! I think in practice the boxlib/AMReX reader just mapped (accidentally?) everything to True for now, avoiding those kinds of bugs?

My thought exactly. Thanks for confirming and fixing this bug !

yt/frontends/boxlib/data_structures.py
```diff
@@ -1650,11 +1650,11 @@ def _parse_parameter_file(self):
...
-        is_periodic = self.parameters["geometry.is_periodic"].split()
-        periodicity = [bool(val) for val in is_periodic]
-        for _ in range(self.dimensionality, 3):
-            periodicity += [True]  # pad to 3D
-        self.periodicity = ensure_tuple(periodicity)
+        self.periodicity[:] = False
+        if "geometry.is_periodic" in self.parameters:
+            is_periodic = self.parameters["geometry.is_periodic"].split()
+        self.periodicity[:len(is_periodic)] = [val == "1" for val in is_periodic]
+        self.periodicity = ensure_tuple(self.periodicity)
```
neutrinoceros, line 1657: btw, I know it's not your fault, but this line is not only useless but also harmful since it overwrites the original type for this attribute (`MuttableAttribute`).
Sorry you happened to touch such a buggy bunch of lines !

ax3l:
re @zingale's comment (offline discussed): the problem that non-periodic boundaries break yt analysis routines seems to be mainly related to volume rendering.

I leave this to others to test since I am not very familiar with this part. But it sounds to me like a potential work-around should go in the volume rendering routines.

zingale:
Axel, it looks like your PR only affects WarpX, so I am fine with however you want to go there.  Castro and MAESTROeX infer more stuff from the `job_info` files, and we do the periodic workaround in our VR scripts currently.

yt/frontends/boxlib/data_structures.py
```diff
@@ -1650,11 +1650,10 @@ def _parse_parameter_file(self):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        is_periodic = self.parameters["geometry.is_periodic"].split()
-        periodicity = [bool(val) for val in is_periodic]
-        for _ in range(self.dimensionality, 3):
-            periodicity += [True]  # pad to 3D
-        self.periodicity = ensure_tuple(periodicity)
+        self.periodicity[:] = False
```
ax3l, line 1653:
```
    self.periodicity[:] = False
TypeError: 'tuple' object does not support item assignment
```

neutrinoceros: Yup, apparently in this particular class we're forcing the tuple type, but there's also as note that says it should be fixed...
https://github.com/yt-project/yt/blob/9138a0a7330e642e040f8dc0a459f7dba347fa05/yt/frontends/boxlib/data_structures.py#L624

Anyway, this is a different problem than the one you're fixing, sorry again... I'll fix this in a separate branch and if all goes well we'll be able to merge this one right after.

neutrinoceros: So it's much worse than I anticipated, and there is a lot of inconsistency right now in how we handle this particular attribute across frontends. You may simply disregard this and backpedal to using tuples. I'm sorry for the back and forth, thanks for your reactivity.

ax3l: I think I will set it to `True` as other readers in the frontend and ignore it altogether. This will cause the least breakage.

yt/frontends/boxlib/data_structures.py
```diff
@@ -1650,7 +1650,7 @@ def _parse_parameter_file(self):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        self.periodicity[:] = False
+        self.periodicity = (False, False, False)
         if "geometry.is_periodic" in self.parameters:
             is_periodic = self.parameters["geometry.is_periodic"].split()
             self.periodicity[: len(is_periodic)] = [val == "1" for val in is_periodic]
```
neutrinoceros: This won't work either because tuples are immutable, here's a fix (backpedaling again...)
```suggestion
        self.periodicity = [False, False, False]
        if "geometry.is_periodic" in self.parameters:
            is_periodic = self.parameters["geometry.is_periodic"].split()
            self.periodicity[: len(is_periodic)] = [val == "1" for val in is_periodic]
        self.periodicity = ensure_tuple(self.periodicity)
```

ax3l: applied as code comment.

ax3l:
This PR only does two things:
- it avoids breakage when the input parameter `geometry.is_periodic`, which is optional, is not set
- it makes the parsing "bug" that we treat everything as periodic, which masks other bugs, obivous

neutrinoceros:
About the volume rendering stuff, I think I should clarify:
while it's true that a number of core functionalities in yt require periodicity (or rather, they need knowledge of boundary conditions), note that it is not blocking to have non-periodic datasets since the periodicity attribute is exposed to the user. So the common workaround is to write (on the user side)

```python
ds.periodicity = [True]*3
```

Another way to handle the boundary issue is to select a region that doesn't include the domain edges. Note that those workaround should be clearly documented in the error message you'll get if you encounter such issues.

atmyers:
It's okay with me if we just set this to `True` for now (which it was doing anyway but I think accidently). If there is a volume rendering issue with `False` it can be fixed in the volume rendering code and worked around in user scripts until then.

ax3l:
Agreed, sound good to me! :+1:

neutrinoceros:
I respectfully disagree with the current state of this PR. I don't think we should embrace existing bugs, especially since the corrected version of the code can be worked around in downstream codes if necessary.

> I think I will set it to True as other readers in that frontend and ignore it altogether.

I don't think this is what the other readers are doing, but it looks like the parsing for this particular attribute is actually broken in more than one place. I have issued a broader fix here #2792

neutrinoceros:
@zingale , so for clarification, you mean the bug that I'm fixing here, right ?

Although I'm not a user of the BoxLib frontend, my (amrvac) datasets are _never_ fully periodic so I'm used to the fact that yt's functionalities easily break unless I force periodicity. I don't consider this as a bug as much as partially implemented/supported features.
Usually you hit this error:
https://github.com/yt-project/yt/blob/c19c3c615b996c8a6e418362ffea9041a616d673/yt/geometry/selection_routines.pyx#L1001
Which is imperfect since the user aren't usually trying to retrieve ghost zones directly (I've been working on making the user experience less confusing in some of my open PRs), but the message already indicates the workaround, so I'd say that it's at least easily discovered.

What do you think we should do ? Is the present PR acceptable in this state ?

zingale:
that looks like one way to fix it.  I think better would be to just leave the portion outside of the domain empty in the plot.

neutrinoceros:
@zingale 
I'm so sorry, I got confused between the different branches/PRs, I though this was #2792

neutrinoceros:
> that looks like one way to fix it. I think better would be to just leave the portion outside of the domain empty in the plot.

@zingale Unfortunately this solution isn't directly applicable in some cases, because we still need ghost-zones in order to compute space-derivatives, even inside the domain. This problem is general to yt, and in no way specific to boxlib so I think this warrants for a broader discussion with other devs :)

matthewturk (approved):

matthewturk:
Is this OK to merge?  It seems good to me...

neutrinoceros:
> Is this OK to merge? It seems good to me...

It's not ! Although the original state of the PR was acceptable, it's now more of a draft and I really think it shouldn't be merged as is. Additionally #2792 fixes the issue at hand with a broader scope (code dedup in the frontend).

PR #2789, opened 2020-07-27
Slash commands

matthewturk:
This PR implements three chatops style commands: `example` (which we can later disable), `rebase` and `black`.

It does not yet work, and will need both the repository for the slash commands transferred over to `yt-project` (it's currently at https://github.com/matthewturk/slash-command-processor ) and the personal access token set correctly.

I also think there's a possibility we want to change some of the parameters, maybe even add `isort`, and set @yt-fido to be the user in charge.  I am planning to issue a PR to YTEP-0037 as well.

### DIFF
--- a/.github/workflows/slash-command-dispatch.yml
+++ b/.github/workflows/slash-command-dispatch.yml
@@ -14,4 +14,4 @@ jobs:
             help
             format
             rebase
-          repository: matthewturk/slash-command-processor
+          repository: yt-project/slash-command-processor
--- a/.github/workflows/slash-command-dispatch.yml
+++ b/.github/workflows/slash-command-dispatch.yml
@@ -14,4 +14,6 @@ jobs:
             help
             format
             rebase
+            isort
+            black
           repository: yt-project/slash-command-processor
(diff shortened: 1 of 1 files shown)


neutrinoceros:
> woof!

What the ... ??

This looks awesome, thank you so much for digging into this Matt !

matthewturk:
@neutrinoceros It was fun.  I've added `/isort` and `/help`.  Might be better to do both isort and black in a single command, like `/format` but that can come later.

I need to get the repo transferred and switch the commits to be fido.

neutrinoceros:
>  Might be better to do both isort and black in a single command, like /format but that can come later.

I'd say let's do it sooner rather than later, it's best to keep a single command aliased for a group of related jobs here since we might update the meaning of such a command later.

matthewturk:
OK, now it supports help, format, and rebase.

munkm:
So with the `format` command if it's not automatically fixable in some way does that mean all the formatting steps need to be done manually? Maybe we should have `format` and then one for each formatting (so adding in `black` and `isort` as additional commands) command so that way we can still use slash commands for the things that are automatically fixable.

matthewturk:
@munkm I don't know what happens if it can't automatically do it.  I've not run into that case with either black or isort.  I'm happy to add another command that does each one individually, though!

PR #2792, opened 2020-07-28
[BOXLIB] consistent parsing for periodicity parameters

neutrinoceros:
## PR Summary

This supersedes #2787 and adresses the issue at the frontend level. Turns out there is actually more than one place where periodicity is incorrectly or inconsistently parsed.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -756,7 +756,7 @@ class BoxlibDataset(Dataset):
             elif param == "amr.ref_ratio":
                 vals = self.refine_by = int(vals[0])
             elif param == "Prob.lo_bc":
-                vals = self.periodicity = ensure_tuple([i == 0 for i in vals.split()])
+                vals = self.periodicity = ensure_tuple([p == "1" for p in vals.split()])
             elif param == "castro.use_comoving":
                 vals = self.cosmological_simulation = int(vals)
             else:
@@ -915,9 +915,8 @@ class BoxlibDataset(Dataset):
         tmp = self.domain_dimensions.tolist()
         tmp.extend((1, 1))
         self.domain_dimensions = np.array(tmp)
-        tmp = list(self.periodicity)
-        tmp[1] = False
-        tmp[2] = False
+        self.periodicity = list(self.periodicity)
+        self.periodicity[1:] = False
         self.periodicity = ensure_tuple(tmp)
 
     def _setup2d(self):
@@ -1160,17 +1159,15 @@ class CastroDataset(BoxlibDataset):
         self.parameters["HydroMethod"] = "Castro"
 
         # set the periodicity based on the runtime parameters
-        periodicity = [True, True, True]
-        if not self.parameters["-x"] == "interior":
-            periodicity[0] = False
-        if self.dimensionality >= 2:
-            if not self.parameters["-y"] == "interior":
-                periodicity[1] = False
-        if self.dimensionality == 3:
-            if not self.parameters["-z"] == "interior":
-                periodicity[2] = False
+        self.periodicity = [True, True, True]
+        for i, axis in enumerate("xyz"):
+            try:
+                self.periodicity[i] = self.parameters["-%s" % axis] == "interior"
+            except KeyError:
+                break
 
-        self.periodicity = ensure_tuple(periodicity)
+        self.periodicity = ensure_tuple(self.periodicity)
+
         if os.path.isdir(os.path.join(self.output_dir, "Tracer")):
             # we have particles
             self.parameters["particles"] = 1
@@ -1244,7 +1241,7 @@ class MaestroDataset(BoxlibDataset):
             except KeyError:
                 pass
 
-        self.periodicity = ensure_tuple(periodicity)
+        self.periodicity = tuple(periodicity)
 
 
 class NyxHierarchy(BoxlibHierarchy):
@@ -1597,11 +1594,13 @@ class WarpXDataset(BoxlibDataset):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        is_periodic = self.parameters["geometry.is_periodic"].split()
-        periodicity = [bool(val) for val in is_periodic]
-        for _ in range(self.dimensionality, 3):
-            periodicity += [True]  # pad to 3D
-        self.periodicity = ensure_tuple(periodicity)
+        self.periodicity = [True] * 3
+        try:
+            is_periodic = self.parameters["geometry.is_periodic"].split()
+            self.periodicity[: len(is_periodic)] = [p == "1" for p in is_periodic]
+        except KeyError:
+            pass
+        self.periodicity = ensure_tuple(self.periodicity)
 
         particle_types = glob.glob(self.output_dir + "/*/Header")
         particle_types = [cpt.split(os.sep)[-2] for cpt in particle_types]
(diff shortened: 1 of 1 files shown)


yt/frontends/boxlib/data_structures.py
```diff
@@ -733,7 +733,7 @@ def _parse_cparams(self):
             elif param == "amr.ref_ratio":
                 vals = self.refine_by = int(vals[0])
             elif param == "Prob.lo_bc":
-                vals = self.periodicity = ensure_tuple([i == 0 for i in vals.split()])
+                vals = self.periodicity = ensure_tuple([p == "1" for p in vals.split()])
```
neutrinoceros, line 736: this line is clearly bugged since it relies on comparing a string to an integer. From what I think I understood, this should be the correct version for it.

yt/frontends/boxlib/data_structures.py
```diff
@@ -1177,17 +1176,15 @@ def _parse_parameter_file(self):
...
-        self.periodicity = ensure_tuple(periodicity)
+        self.periodicity = [True, True, True]
+        for i, axis in enumerate("xyz"):
+            try:
+                self.periodicity = self.parameters["-%s" % axis] == "interior"
+            except KeyError:
+                break
+
+        self.periodicity = ensure_tuple(self.periodicity)
+
```
neutrinoceros, line 1187: Here is just a refactor to avoid repeated code.

yt/frontends/boxlib/data_structures.py
```diff
@@ -1263,17 +1260,14 @@ def _parse_parameter_file(self):
...
-            periodicity[1] = False
-
-        if not self.parameters["bcz_lo"] == -1:
-            periodicity[2] = False
+        self.periodicity = [True, True, True]
+        for i, axis in enumerate("xyz"):
+            try:
+                self.periodicity[i] = self.parameters["bc%s_lo" % axis]
+            except KeyError:
+                break
```
neutrinoceros, line 1268: Here is just a refactor to avoid repeated code (bis).

yt/frontends/boxlib/data_structures.py
```diff
@@ -1650,11 +1644,13 @@ def _parse_parameter_file(self):
...
-        for _ in range(self.dimensionality, 3):
-            periodicity += [True]  # pad to 3D
-        self.periodicity = ensure_tuple(periodicity)
+        self.periodicity = [True] * 3
+        try:
+            is_periodic = self.parameters["geometry.is_periodic"].split()
+            self.periodicity[: len(is_periodic)] = [p == "1" for p in is_periodic]
+        except KeyError:
+            pass
+        self.periodicity = ensure_tuple(self.periodicity)
```
neutrinoceros, line 1653: This is the part addressed originally in #2787

yt/frontends/boxlib/data_structures.py
```diff
@@ -1650,11 +1644,13 @@ def _parse_parameter_file(self):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        is_periodic = self.parameters["geometry.is_periodic"].split()
-        periodicity = [bool(val) for val in is_periodic]
-        for _ in range(self.dimensionality, 3):
-            periodicity += [True]  # pad to 3D
-        self.periodicity = ensure_tuple(periodicity)
+        self.periodicity = [True] * 3
```
ax3l, line 1647: The default in AMReX is `False` if not specified: https://amrex-codes.github.io/amrex/docs_html/InputsProblemDefinition.html

yt/frontends/boxlib/data_structures.py
```diff
@@ -1650,11 +1644,13 @@ def _parse_parameter_file(self):
                     self.parameters[l[0].strip()] = l[1].strip()
 
         # set the periodicity based on the integer BC runtime parameters
-        is_periodic = self.parameters["geometry.is_periodic"].split()
```
ax3l, line 1653: This line is the crash that #2787 fixes.

yt/frontends/boxlib/data_structures.py
neutrinoceros: Thank you for checking this ! I assumed the dominant behavior was correct but I guess that's one more bug to take down here. Will fix as soon as I get my hands on my laptop.

neutrinoceros: I went back and applied the same fix to BoxlibDataset itself as well as Castro following the docs.

yt/frontends/boxlib/data_structures.py
```diff
@@ -621,7 +621,7 @@ class BoxlibDataset(Dataset):
     _output_prefix = None
 
     # THIS SHOULD BE FIXED:
```
neutrinoceros, line 623: @matthewturk can I have your review on this comment you wrote ? I don't know what was there to fix (default value, type, anything else ?) and if that comment should be cleaned up or updated.

munkm:
I think because this PR supersedes another one then it is important that @ax3l and @zingale review and approve it. What are both of your thoughts on this solution?

yt/frontends/boxlib/data_structures.py
```diff
@@ -888,10 +888,9 @@ def _setup1d(self):
         tmp = self.domain_dimensions.tolist()
         tmp.extend((1, 1))
         self.domain_dimensions = np.array(tmp)
-        tmp = list(self.periodicity)
-        tmp[1] = False
-        tmp[2] = False
-        self.periodicity = ensure_tuple(tmp)
+        self.periodicity = list(self.periodicity)
+        self.periodicity[1:] = False
```
ax3l, line 892: I think I don't understand this line easily, can you please clarify or add a comment? :)

ax3l (commented):
Looks good to me, just one question.

@zingale and @atmyers do you like to give this a 2nd view? :)

yt/frontends/boxlib/data_structures.py
neutrinoceros: Thanks for reviewing this ! In light of recent developments (especially https://github.com/yt-project/yt/pull/2893, where I remove `ensure_tuple`), I just gave this another go. The specific line you're commenting here is now gone, and now periodicity should be only parsed once in `BoxlibDataset._parse_cparams`, in a dimensionality-agnostic way. Is it clearer now ?

yt/frontends/boxlib/data_structures.py
```diff
@@ -718,8 +717,6 @@ def _parse_cparams(self):
         for line in (line.split("#")[0].strip() for line in open(self.cparam_filename)):
             if "=" not in line:
                 continue
-            if len(line) == 0:
```
neutrinoceros, line 721: this has nothing to do with the PR but I think this check is 100% redundant with the one above so I simply removed it

zingale:
I can confirm that this works with Castro and correctly detects the periodicity (at least for the data set I tried).

zingale (approved):

neutrinoceros:
Thank you @zingale ! 🎉

yt/frontends/boxlib/data_structures.py
neutrinoceros: @ax3l We'll wait for your reply before this is merged :)

neutrinoceros:
hi @ax3l, I'm sorry to ask again but I couldn't find another way to contact you: do you approve of this PR in the current state ? thanks !

ax3l (approved):
LGTM, thank you!

yt/frontends/boxlib/data_structures.py
ax3l: > Is it clearer now ?

yes, thanks a lot.

Issue #2784, opened 2020-07-25
Pickling is unsafe and brittle


neutrinoceros:
### Bug report

**Bug summary**

I noticed there's a handful of places in the code base where we use the `pickle` module to dump and load data.
After reading [this blogpost](https://nedbatchelder.com/blog/202006/pickles_nine_flaws.html) (in particular point 1 and 2), I feel like these should be reimplemented with safer data storing methods when possible, or simply removed if they turn out not used.

brittonsmith:
This sounds familiar, as in I think getting rid of `pickle` code is a conclusion that we've arrived at before, but have not followed through on. There may be exceptions, but I think `yt.save_as_dataset` is probably the most reliable method for saving and reloading data. At least, I believe its the one we endorse the most.

neutrinoceros:
I'm curious what is this function used for ? Building datasets from scratch with the Stream frontend and saving them, or something else ?

brittonsmith:
It's basically a way of saving arbitrary array data to a format that can be yt.loaded. It's become the basis for the `HaloCatalog` frontend and I use it in `ytree` as well. [Here](https://yt-project.org/docs/dev/analyzing/saving_data.html) are the docs in it.

neutrinoceros:
Nice one ! indeed this is much better, and I think we could probably just get rid of of `save_object` methods, going to try this now !

PR #2793, opened 2020-07-29
cleanup: remove (seemingly) unused and deprecated pickling procedures

neutrinoceros:
## PR Summary

Removes some save/load procedures that don't seem used anywhere in the code base and are redundant with, as well as less safe than `obj.save_as_dataset`

Partly addresses #2784 

## PR Checklist

- [x] pass `flake8 yt/`


munkm:
Your PR title mentions that they're deprecated, but it seems like they don't have deprecation warnings on them at all? So is this assumed because `save_as_dataset` exists?

neutrinoceros:
@munkm yes, exactly. I assume that it would be smoother to add deprecation warnings instead of removing them altogether, but I'm genuinely surprised this didn't break anything.

PR #2795, opened 2020-07-29
Misc fixes to issues found during docs building

Xarthisius:
## PR Summary

A few fixes for errors and warnings found by Jenkins

## PR Checklist

- [ ] pass `flake8 yt/`

### DIFF
--- a/doc/source/visualizing/callbacks.rst
+++ b/doc/source/visualizing/callbacks.rst
@@ -94,7 +94,7 @@ dataset from AMRVAC :
     import yt
 
     ds = yt.load("amrvac/bw_polar_2D0000.dat")
-    s = yt.plot2d(ds, 'density')
+    s = yt.plot_2d(ds, "density")
     s.set_background_color("density", "black")
 
     # Plot marker and text in data coords
--- a/yt/visualization/color_maps.py
+++ b/yt/visualization/color_maps.py
@@ -508,7 +508,11 @@ if cmocean is not None:
 # Add colormaps in _colormap_data.py that weren't defined here
 _vs = np.linspace(0, 1, 256)
 for k, v in list(_cm.color_map_luts.items()):
-    if k not in yt_colormaps and k not in mcm.cmap_d:
+    try:
+        colormaps = mcm._cmap_registry
+    except AttributeError:  # mpl < 3.3.0
+        colormaps = mcm.cmap_d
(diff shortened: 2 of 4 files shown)


doc/source/visualizing/callbacks.rst
```diff
@@ -94,7 +94,7 @@ dataset from AMRVAC :
     import yt
 
     ds = yt.load("amrvac/bw_polar_2D0000.dat")
-    s = yt.plot2d(ds, 'density')
+    s = yt.SlicePlot(ds, axis=2, fields=['density'])
```
neutrinoceros, line 97: I think the explicit wart syntax is much better, especially in documentation, but I'm curious: why the change from plot_2d to SlicePlot ?

yt/visualization/color_maps.py
```diff
@@ -508,7 +508,7 @@ def add_colormap(name, cdict):
 # Add colormaps in _colormap_data.py that weren't defined here
 _vs = np.linspace(0, 1, 256)
 for k, v in list(_cm.color_map_luts.items()):
-    if k not in yt_colormaps and k not in mcm.cmap_d:
+    if k not in yt_colormaps and k not in mcm._cmap_registry:
```
neutrinoceros, line 511: From what I've seen this is indeed the new place where this dict lives in mpl as of 3.3, but do you know if it's backward compatible?

doc/source/visualizing/callbacks.rst
Xarthisius: It's simple, I couldn't find plot2d and I didn't even know that `plot_2d` existed :)

neutrinoceros: Oh nice, I'm only seeing the typo now, haha.
That's what code review is for I guess, always can use another pair of eyes !

yt/visualization/color_maps.py
Xarthisius: fixed

doc/source/visualizing/callbacks.rst
Xarthisius: reverted and fixed the typo

yt/visualization/color_maps.py
neutrinoceros: Well done ! I wonder if we should refactor to just NOT use this dict since they are not keeping it exposed... but this is out of scope here.

PR #2798, opened 2020-07-31
feature: support for gradient computation of several fields at a time, and support simple fnames as inputs

neutrinoceros:
## PR Summary

Using the newly refactored version of `YTDataContainer._determine_fields` method, allow for more than one field to be passed to `ds.add_gradient_fields` at a time and allow passing simple `fname`, when `(ftype, fname)` tuples were mandatory previously.
The latter feature is not as robust as I'd wish since we currently lack (I think) a mechanism to error out in case the inferred `ftype` is ambiguous. I believe the required change for this should happen in `Dataset._get_field_info`, though it is beyond the scope of the present PR.

## PR Checklist

- [x] pass `flake8 yt/`


neutrinoceros:
Adding the "api-consistency" label since there are already a lot of user-facing methods that accept single `fname` as a `field` label and determine the inferred `ftype` in a similar fashion internally, and this is actually the primary reason for this PR.

matthewturk (approved):

PR #2799, opened 2020-07-31
bugfix: remove B007 from flake8 ignore list and fix all errors detected

neutrinoceros:
## PR Summary

Bugbear B007 error is
```
Loop control variable 'x' not used within the loop body.
```
When bugbear was added to the linting stage of CI, I originally ignored this error because there were a lot of warnings and most of them looked like valid or at least not-broken code.
Indeed most of the time what happens is that someone writes a loop using `enumerate` but the loop is rewritten down the road and the index becomes unused, but the `enumerate` is kept.
Other times ones needs to iterate over some but not all items from in a tuple iterator, so the unused items are named but unused.
Those cases are not bug, but they actually mask some _actual_ bugs if we ignore the warning altogether.
After a thorough review it appears that this detects a handful of broken loops, some of which are part of tests ! so now I think it's worth fixing the error and catching it early in future contributions by enabling B007.

All of the changes and handwritten and I spent maybe 10s on each one tops, so I advise caution in reviews because it's possible that I'm introducing new bugs with this first try.
In many cases, I didn't take the time to truly understand what the loops were doing. I just assumed the code was correct and added a `_` prefix to the unused variable name (which marks it as "intentionally unused"). Sometimes this is legit to do even with only one var returned by the iterator, but other times this marks a plain broken loop, so I hope I haven't actively hidden existing bugs doing so.

## PR Checklist

- [x] pass `flake8 yt/`


yt/frontends/art/io.py
```diff
@@ -372,7 +372,7 @@ def cfc(root_level, level, le):
     if root_level is None:
         root_level = np.floor(np.log2(le.max() * 1.0 / coarse_grid))
         root_level = root_level.astype("int64")
-        for i in range(10):
+        for level in range(10):
```
neutrinoceros, line 375: this change is intended as a bugfix, because the original code looked broken to me.

yt/visualization/volume_rendering/interactive_loop.py
```diff
@@ -84,9 +84,7 @@ def setup_loop(self, scene, camera):
         return callbacks
 
     def start_loop(self, scene, camera):
-        callbacks = self.setup_loop(scene, camera)
```
neutrinoceros, line 87: removing dead code here exposes that the method is currently just a direct wrapper for another one

yt/visualization/profile_plotter.py
```diff
@@ -355,10 +355,10 @@ def _repr_html_(self):
         ret = ""
         unique = set(self.plots.values())
         if len(unique) < len(self.plots):
-            iters = zip(range(len(unique)), sorted(unique))
+            iters = sorted(unique)
         else:
-            iters = self.plots.items()
-        for uid, plot in iters:
+            iters = self.plots.values()
+        for plot in iters:
```
neutrinoceros, line 361: this change comes with a simplification of lines right above

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -404,7 +404,7 @@ def free_workgroup(self, workgroup):
     def free_all(self):
         for wg in self.workgroups:
             self.free_workgroup(wg)
-        for i in range(len(self.workgroups)):
+        while self.workgroups:
```
neutrinoceros, line 407: this seems like a correct, more pythonic refactor to me

yt/utilities/answer_testing/framework.py
```diff
@@ -214,7 +214,7 @@ def get(self, ds_name, default=None):
         except urllib.error.HTTPError:
             raise YTNoOldAnswer(url)
         else:
-            for this_try in range(3):
+            for _ in range(3):
```
neutrinoceros, line 217: this loop feels weird to me but it may be correct (try up to 3 times the same exact request with a 0.01s delay between each attempt)

yt/testing.py
```diff
@@ -585,7 +585,7 @@ def construct_octree_mask(prng=RandomState(0x1D3D3D3), refined=None):  # noqa B0
         return refined
 
     # Loop over subcells
-    for subcell in range(8):
+    for _ in range(8):
```
neutrinoceros, line 588: this code is potentially broken but the fix isn't obvious to me

yt/frontends/sdf/io.py
```diff
@@ -27,7 +27,7 @@ def _read_particle_coords(self, chunks, ptf):
             for obj in chunk.objs:
                 data_files.update(obj.data_files)
         assert len(data_files) == 1
-        for data_file in sorted(data_files, key=lambda x: (x.filename, x.start)):
```
neutrinoceros, line 30: Loops in this file look broken since we sort the datafiles but don't use them at all during iteration

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -244,7 +244,7 @@ def test_ramses_sink():
     ad = ds.all_data()
 
     for field in expected_fields:
-        assert ("sink", "field") not in ds.field_list
```
neutrinoceros, line 247: I think this test was bug and this is an attempt to fix it

yt/data_objects/construction_data_containers.py
```diff
@@ -1542,7 +1542,7 @@ def get_data(self, fields=None, sample_type="face", no_ghost=False):
             mylog.info("Extracting (sampling: %s)" % (fields,))
         verts = []
         samples = []
-        for io_chunk in parallel_objects(self.data_source.chunks([], "io")):
+        for _io_chunk in parallel_objects(self.data_source.chunks([], "io")):
```
neutrinoceros, line 1545: weird that this is not used. Is this a bug ?

yt/data_objects/data_containers.py
```diff
@@ -413,12 +411,12 @@ def _generate_particle_field(self, field):
         return rv
 
     def _count_particles(self, ftype):
-        for (f1, f2), val in self.field_data.items():
+        for (f1, _f2), val in self.field_data.items():
             if f1 == ftype:
                 return val.size
         size = 0
```
matthewturk, line 417: In retrospect, this may not be correct, but I do not have an immediate fix.  I *think* we probably want to return `val.shape[0]` which in most cases will be identical to `val.size` but in the case of vector or array fields will not.

yt/data_objects/particle_trajectories.py
```diff
@@ -111,15 +111,15 @@ def __init__(
             mylog.setLevel(old_level)
 
         times = []
-        for fn, (time, indices, pfields) in sorted(my_storage.items()):
+        for _fn, (time, _indices, _pfields) in sorted(my_storage.items()):
```
matthewturk, line 114: @jzuhone can you take a look at this?

yt/data_objects/tests/test_slice.py
```diff
@@ -39,40 +39,38 @@ def test_slice(pf):
         uc = [np.unique(c) for c in coords]
         slc_pos = 0.5
         # Some simple slice tests with single grids
-        for ax, an in enumerate("xyz"):
+        for ax in range(3):
             xax = ds.coordinates.x_axis[ax]
             yax = ds.coordinates.y_axis[ax]
-            for wf in ["density", None]:
```
matthewturk, line 45: oh right slices don't have weight fields ... :)

yt/frontends/art/io.py
matthewturk: I'm not completely sure that it is.  The `10` may not refer to a max level.  And it does have `level` passed in from above.  So it's possible that `10` is some constant, like, sufficient number of times to iterate.

yt/frontends/gadget/io.py
```diff
@@ -53,7 +53,7 @@ def _read_particle_coords(self, chunks, ptf):
             si, ei = data_file.start, data_file.end
             f = h5py.File(data_file.filename, mode="r")
             # This double-reads
-            for ptype, field_list in sorted(ptf.items()):
+            for ptype in sorted(ptf):
```
matthewturk, line 56: good catch; this was copy/pasta from the reading of fields

yt/frontends/ramses/tests/test_outputs.py
matthewturk: I think so too

yt/utilities/answer_testing/framework.py
matthewturk: (in a separate PR)

Alternately, we should just clean up this whole file once pytest goes in.

yt/frontends/art/io.py
neutrinoceros: wow. So would it work to actually use `io_chunk` in lieu of `self.data_source` down there, for clarity ?

yt/visualization/volume_rendering/interactive_loop.py
neutrinoceros: cool !

yt/frontends/art/io.py
matthewturk: Yes, but I think maybe this is one of those things that we shouldn't change until we're being deliberate about it, and maybe the time for deliberate decisions about chunking has arrived ...

neutrinoceros: Alright, thanks for the feedback !

yt/data_objects/tests/test_slice.py
neutrinoceros: I guess not ? here I think the loop was just doing twice the same exact stuff.

yt/data_objects/data_containers.py
neutrinoceros: Let me know if I should just implement this.

yt/data_objects/construction_data_containers.py
neutrinoceros: Would it make sense to use `io_chunk` inside the loop in lieu of `self.data_source` then ?

yt/data_objects/data_containers.py
neutrinoceros: Let me know if I should just implement this.

neutrinoceros:
@Xarthisius @matthewturk may I ask your help with the Jenkins failure ? it looks like a single bug to me but I don't know how to reproduce those locally.

Xarthisius:
> @Xarthisius @matthewturk may I ask your help with the Jenkins failure ? it looks like a single bug to me but I don't know how to reproduce those locally.

The following code can reproduce the error:

```python
import yt
d9p = "D9p_500/10MpcBox_HartGal_csf_a0.500.d"
ds = yt.load(d9p)
proj = ds.proj(("gas", "density"), 0, weight_field=None)
```

Error goes away when I revert this bit:
```diff
diff --git a/yt/frontends/art/io.py b/yt/frontends/art/io.py
index 8d8650710..41d25d581 100644
--- a/yt/frontends/art/io.py
+++ b/yt/frontends/art/io.py
@@ -372,7 +372,7 @@ def _read_art_level_info(
     if root_level is None:
         root_level = np.floor(np.log2(le.max() * 1.0 / coarse_grid))
         root_level = root_level.astype("int64")
-        for level in range(10):
+        for i in range(10):
             fc = cfc(root_level, level, le)
             go = np.diff(np.unique(fc)).min() < 1.1
             if go:
```

neutrinoceros:
Thanks @Xarthisius , that's on me for trying to fix so many things at once, I owe you one !

yt/frontends/open_pmd/data_structures.py
```diff
@@ -250,7 +250,7 @@ def _count_grids(self):
         self.vpg = int(self.dataset.gridsize / 4)  # 4Byte per value (f32)
 
         # Meshes of the same size do not need separate chunks
-        for (shape, spacing, offset, unit_si) in set(self.meshshapes.values()):
+        for (shape, _spacing, _offset, _unit_si) in set(self.meshshapes.values()):
```
cphyc, line 253:
```suggestion
        for shape, *_ in set(self.meshshapes.values()):
```

yt/frontends/adaptahop/io.py
```diff
@@ -273,7 +273,7 @@ def _todo_from_attributes(attributes):
 
 def _find_attr_position(key):
     j = 0
-    for attrs, l, k in HALO_ATTRIBUTES:
+    for attrs, _l, _k in HALO_ATTRIBUTES:
```
cphyc, line 276:
```suggestion
    for attrs, *_ in HALO_ATTRIBUTES:
```

yt/frontends/adaptahop/io.py
```diff
@@ -67,7 +67,7 @@ def _read_particle_fields(self, chunks, ptf, selector):
                 data_files.update(obj.data_files)
 
         def iterate_over_attributes(attr_list):
-            for attr, length, dtype in attr_list:
+            for attr, _length, _dtype in attr_list:
```
cphyc, line 70:
```suggestion
            for attr, *_ in attr_list:
```

yt/fields/tests/test_fields.py
```diff
@@ -116,13 +116,13 @@ def __call__(self):
 def get_base_ds(nprocs):
     fields, units = [], []
 
-    for fname, (code_units, aliases, dn) in StreamFieldInfo.known_other_fields:
+    for fname, (code_units, _aliases, _dn) in StreamFieldInfo.known_other_fields:
```
cphyc, line 119:
```suggestion
    for fname, (code_units, *_) in StreamFieldInfo.known_other_fields:
```

yt/data_objects/particle_trajectories.py
```diff
@@ -111,15 +111,15 @@ def __init__(
...
             times.append(time)
         self.times = self.data_series[0].arr([time for time in times], times[0].units)
```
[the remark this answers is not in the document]
neutrinoceros: wow, I didn't know about the syntax you use in the second version. Is there a name for this ? I'm tempted to use it because it's new and shiny but I'm not sure about readability.

yt/data_objects/tests/test_ortho_rays.py
```diff
@@ -8,7 +8,7 @@ def test_ortho_ray():
     dx = (ds.domain_right_edge - ds.domain_left_edge) / ds.domain_dimensions
 
     axes = ["x", "y", "z"]
-    for ax, an in enumerate(axes):
+    for ax, _an in enumerate(axes):
```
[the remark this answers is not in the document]
neutrinoceros: Indeed, pretty weird. I've fixed this pattern the same exact way in other cases but again, went through a lot of them in a short time. Thanks !

yt/frontends/open_pmd/data_structures.py
cphyc: Also the code doesn't seem to be in agreement with the comment, shouldn't this be instead
```suggestion
        shapes = set(shape for (shape, *_) in self.meshshapes.values())
        for shape in shapes:
```

yt/frontends/art/io.py
```diff
@@ -426,7 +426,7 @@ def get_ranges(
     arr_size = np_per_page * real_size
     idxa, idxb = 0, 0
     posa, posb = 0, 0
-    for page in range(num_pages):
+    for _page in range(num_pages):
         idxb += np_per_page
         for i, fname in enumerate(["x", "y", "z", "vx", "vy", "vz"]):
```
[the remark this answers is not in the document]
neutrinoceros: This looks exactly equivalent to me, what am I missing ? @cphyc

yt/frontends/ramses/tests/test_outputs.py
neutrinoceros: Happens to the best of us 🥇

yt/frontends/sdf/io.py
neutrinoceros: Oops I didn't see this reply. Any clue on how this _should_ be fixed ?

yt/data_objects/particle_trajectories.py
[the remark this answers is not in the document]
cphyc: For future reference, this is named "Extended Iterable Unpacking", see https://www.python.org/dev/peps/pep-3132/.

yt/frontends/art/data_structures.py
```diff
@@ -254,7 +254,7 @@ def _parse_parameter_file(self):
         # read the amr header
         with open(self._file_amr, "rb") as f:
             amr_header_vals = fpu.read_attrs(f, amr_header_struct, ">")
-            for to_skip in ["tl", "dtl", "tlold", "dtlold", "iSO"]:
+            for _to_skip in ["tl", "dtl", "tlold", "dtlold", "iSO"]:
                 fpu.skip(f, endian=">")
```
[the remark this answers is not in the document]
neutrinoceros: I like it

yt/data_objects/particle_trajectories.py
```diff
@@ -111,15 +111,15 @@ def __init__(
...
 
         self.particle_fields = []
         output_field = np.empty((self.num_indices, self.num_steps))
         output_field.fill(np.nan)
         for field in ("particle_position_%s" % ax for ax in "xyz"):
-            for i, (fn, (time, indices, pfields)) in enumerate(
```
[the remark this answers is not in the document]
neutrinoceros: Looks like a good idea. Will push with this change in a few minutes

yt/data_objects/tests/test_sph_data_objects.py
```diff
@@ -170,7 +170,7 @@ def test_ray():
 def test_cutting():
     ds = fake_sph_orientation_ds()
     for (normal, center), answer in CUTTING_ANSWERS.items():
-        for i in range(-1, 2):
+        for _ in range(3):
```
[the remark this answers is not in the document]
neutrinoceros: seems like another broken test. @matthewturk, any idea what the intention here was ?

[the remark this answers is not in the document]
matthewturk: So `range(-1,2)` gives `[-1, 0, 1]`.  My guess is that this is supposed to be an offset.  0.1 is within the radius of the smoothing length for all the particles, so what the test is probably trying to do is to offset the center along the axis.  I'd guess that maybe either `c + 0.1 * c` is supposed to be `c + 0.1 * i` or `cen[i]` gets offset or something was the original intention.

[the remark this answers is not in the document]
neutrinoceros: sounds likely. I'll update this, but then I suppose it is excepted that the test will fail since it was apparently validated with wrong data.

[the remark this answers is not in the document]
neutrinoceros: welp. It didn't fail. https://trello-attachments.s3.amazonaws.com/58e3f7c543422d7f3ad84f33/5f0751415f713e68c3cd2653/41a92bd352d8a14b62152d3c79f759df/coverage.png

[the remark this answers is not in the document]
matthewturk: Good news!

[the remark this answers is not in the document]
neutrinoceros: Is it ? I'm now suspecting the test isn't really doing anything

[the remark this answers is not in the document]
matthewturk: I think the simpler answer is that it's passing.  What it's doing is checking that when we select a cutting plane that we are also selecting the particles whose smoothing length puts them within the range of the cutting plane.

If we were to reduce the smoothing lengths to, so, 1e-6, it would potentially fail.  Which would mean it is indeed working.

PR #2802, opened 2020-08-01
Correct child mask for sibling grids

cevans216:
## PR Summary

This pull request corrects how `AMRGridPatch` handles sibling grids (`OverlappingSiblings`) when calculating `child_mask` and `child_index_mask` by passing `dlevel=0` to all relevant calls to `_fill_child_mask`. This parameter specifies the number of refinement levels by which the two grids differ and is used to compare indices between them. The default value of `dlevel=1` is appropriate for true child grids (in the sense that they belong to the next highest refinement level),  but for siblings at the same level this should be called with `dlevel=0`.

This appears to have been addressed a while back in 8ef99073fa8fbd5b4b4af3cb828e801815ee3ede, but it was (unintentionally it seems) reverted with fcaa302a89f4edb8ffb23d9cef60c9a7717a89a5.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/data_objects/grid_patch.py
+++ b/yt/data_objects/grid_patch.py
@@ -214,7 +214,7 @@ class AMRGridPatch(YTSelectionContainer):
         for child in self.Children:
             self._fill_child_mask(child, child_mask, 0)
         for sibling in self.OverlappingSiblings or []:
-            self._fill_child_mask(sibling, child_mask, 0)
+            self._fill_child_mask(sibling, child_mask, 0, dlevel=0)
         return child_mask
 
     @property
@@ -232,7 +232,7 @@ class AMRGridPatch(YTSelectionContainer):
         for child in self.Children:
             self._fill_child_mask(child, child_index_mask, child.id)
         for sibling in self.OverlappingSiblings or []:
-            self._fill_child_mask(sibling, child_index_mask, sibling.id)
+            self._fill_child_mask(sibling, child_index_mask, sibling.id, dlevel=0)
         return child_index_mask
 
     def retrieve_ghost_zones(self, n_zones, fields, all_levels=False, smoothed=False):


neutrinoceros:
Hi @cevans216 , thank you very much for reporting and fixing this !
It's really odd to me that the error is so long-lived even though, as you showed, it seems it was actively fixed before.
I don't know this part of the code base very well, and I think it'd be useful if you could add an example of how this breaks your downstream code, for later reference.

Ideally, such an example could also be used to write a test to make sure this bug doesn't reappear. If you're comfortable writing such a test, please go ahead. Otherwise we can work it out for you.

ping @Xarthisius for review :-)

yt/data_objects/grid_patch.py
```diff
@@ -214,7 +214,7 @@ def child_mask(self):
         for child in self.Children:
             self._fill_child_mask(child, child_mask, 0)
         for sibling in self.OverlappingSiblings or []:
-            self._fill_child_mask(sibling, child_mask, 0)
+            self._fill_child_mask(sibling, child_mask, 0, dlevel=0)
```
neutrinoceros, line 217: while we are at it, it would be clearer to use full explicit keywords here, though it's not required for merging this.
```suggestion
            self._fill_child_mask(child=sibling, mask=child_mask, tofill=0, dlevel=0)
```

yt/data_objects/grid_patch.py
```diff
@@ -232,7 +232,7 @@ def child_index_mask(self):
         for child in self.Children:
             self._fill_child_mask(child, child_index_mask, child.id)
         for sibling in self.OverlappingSiblings or []:
-            self._fill_child_mask(sibling, child_index_mask, sibling.id)
+            self._fill_child_mask(sibling, child_index_mask, sibling.id, dlevel=0)
```
neutrinoceros, line 235: same as above
```suggestion
            self._fill_child_mask(child=sibling, mask=child_index_mask, tofill=sibling.id, dlevel=0)
```

matthewturk:
In principle this is something that we *should* already have a test for in the answers, somewhere.  I will see if I can come up with something though that is short and restricted.

matthewturk:
So I've experimented a little bit and it turns out that my first hunch for how to test it won't work out of the box because we don't handle `OverlappingSiblings` in the stream frontend.  In fact it kind of looks like the only place that it is left is in `gdf/data_structures.py` which does it manually in the `_populate_grid_objects` phase.

Anyway, here's something that I have been experimenting with:

```
import numpy as np
import yt

grids = [
    {'density': np.ones((16, 16, 16)),
     'level': 0,
     'left_edge': [0.0, 0.0, 0.0],
     'right_edge': [1.0, 1.0, 1.0],
     'dimensions': [16, 16, 16]},
    {'density': 2.0 * np.ones((8, 8, 8)),
     'level': 1,
     'left_edge': [0.25, 0.25, 0.25],
     'right_edge': [0.5, 0.5, 0.5],
     'dimensions': [8, 8, 8]},
    {'density': 2.0 * np.ones((8, 8, 8)),
     'level': 1,
     'left_edge': [0.25, 0.25, 0.25],
     'right_edge': [0.5, 0.5, 0.5],
     'dimensions': [8, 8, 8]}
]

#grids.pop(-1)

bbox = np.array([[0.0, 1.0], [0.0, 1.0], [0.0, 1.0]])

ds = yt.load_amr_grids(grids, bbox = bbox, domain_dimensions = [16, 16, 16])

ds.index.grids[1].OverlappingSiblings = [ds.index.grids[2]]
ds.index.grids[2].OverlappingSiblings = [ds.index.grids[1]]

print("level", (ds.all_data()["grid_level"] == 1).sum())
print("ones", ds.all_data()["ones"].size)
```

If you allow `grids.pop(-1)` to run, it should give the same results as if you don't, according to my thinking.  But, it doesn't.  And I'm not sure what's up.  I've instrumented and it's indeed setting things to zero in the masking, as it should be (with your change).  So, hmm.

munkm:
Ok, I'm going to merge this since it seems like we have more work to do on sibling grids anyways. I've added a note to the issue opened by @matthewturk that we need to add more support for tests, and I'll link to this PR in the issue so people can see the discussion about tests.


@cevans216 thank you so much for this PR! We really appreciate your fix and hope to see you around in the future. 🙂

PR #2803, opened 2020-08-03
Fix bugs that prevented correct saving of non-cartesian data with save_as_dataset

neutrinoceros:
## PR Summary

This fixes several bugs:
- grid fields were not passed
- geometry attribute wasn't passed or read
- missing return statement in yt.save_as_dataset

## PR Checklist

- [x] pass `flake8 yt/`


yt/data_objects/tests/test_io_non_cartesian_data.py
```diff
@@ -0,0 +1,21 @@
...
+
+@requires_module("h5py")
+def test_preserve_geometric_properties():
+    for geom in ("cartesian", "cylindrical", "spherical"):
+        ds1 = fake_amr_ds(fields=[("gas", "density")], geometry=geom)
+        ad = ds1.all_data()
+        with NamedTemporaryFile(suffix=".h5", dir=os.getcwd()) as tmpf:
+            fn = ad.save_as_dataset(tmpf.name, fields=["density"])
+            ds2 = YTDataContainerDataset(fn)
+            dfl = ds2.derived_field_list
```
neutrinoceros, line 16: this line single-handedly makes the test much more time-consuming (about 15s on my laptop VS < 1s without this part). @Xarthisius , is it worth investigating for optimisations here or is there a known workaround ?

neutrinoceros: (moved the file to yt/data_objects/tests/test_io_geometry.py)

munkm:
Ok we can let britton off the hook. 🙂

PR #2806, opened 2020-08-04
hotfix: add missing _is_valid method to AMReXDataset class

neutrinoceros:
## PR Summary

fix #2805 
This is now based on top of #2807, do not merge this first

## PR Checklist

- [x] pass `black --check yt/`
- [x] pass `isort . --check --diff`
- [x] pass `flake8 yt/`

PR #2807, opened 2020-08-04
fix: refactor boxlib dataformat validators (code deduplication), fix bugs revealed on the way there

neutrinoceros:
## PR Summary

this is part of #2806, I want to see if it breaks stuff on its own.

`WarpXDataset._is_valid` is probably too permissive as it stands, it makes other classes easily interfere with it.

adresses #2708

Edit: found more bugs along the way (and added tests)

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -2,6 +2,7 @@ import glob
 import inspect
 import os
 import re
+import warnings
 from collections import namedtuple
 from stat import ST_CTIME
 
@@ -125,18 +126,22 @@ class BoxLibParticleHeader:
     def __init__(self, ds, directory_name, is_checkpoint, extra_field_names=None):
 
         self.particle_type = directory_name
-        header_filename = ds.output_dir + "/" + directory_name + "/Header"
+        header_filename = os.path.join(ds.output_dir, directory_name, "Header")
         with open(header_filename, "r") as f:
             self.version_string = f.readline().strip()
 
             particle_real_type = self.version_string.split("_")[-1]
-            particle_real_type = self.version_string.split("_")[-1]
-            if particle_real_type == "double":
-                self.real_type = np.float64
-            elif particle_real_type == "single":
-                self.real_type = np.float32
-            else:
-                raise RuntimeError("yt did not recognize particle real type.")
+            known_real_types = {"double": np.float64, "single": np.float32}
+            try:
+                self.real_type = known_real_types[particle_real_type]
+            except KeyError:
+                warnings.warn(
+                    f"yt did not recognize particle real type {particle_real_type}"
+                    "assuming double",
+                    category=RuntimeWarning,
+                )
+                self.real_type = known_real_types["double"]
+
             self.int_type = np.int32
 
             self.dim = int(f.readline().strip())
@@ -231,14 +236,14 @@ class AMReXParticleHeader:
     def __init__(self, ds, directory_name, is_checkpoint, extra_field_names=None):
 
         self.particle_type = directory_name
-        header_filename = ds.output_dir + "/" + directory_name + "/Header"
+        header_filename = os.path.join(ds.output_dir, directory_name, "Header")
         self.real_component_names = []
         self.int_component_names = []
         with open(header_filename, "r") as f:
             self.version_string = f.readline().strip()
 
             particle_real_type = self.version_string.split("_")[-1]
-            particle_real_type = self.version_string.split("_")[-1]
+
             if particle_real_type == "double":
                 self.real_type = np.float64
             elif particle_real_type == "single":
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -675,26 +675,10 @@ class BoxlibDataset(Dataset):
     def _is_valid(cls, *args, **kwargs):
         # fill our args
         output_dir = args[0]
-        # boxlib datasets are always directories
-        if not os.path.isdir(output_dir):
-            return False
         header_filename = os.path.join(output_dir, "Header")
-        jobinfo_filename = os.path.join(output_dir, "job_info")
-        if not os.path.exists(header_filename):
-            # We *know* it's not boxlib if Header doesn't exist.
-            return False
-        args = inspect.getcallargs(cls.__init__, args, kwargs)
-        # This might need to be localized somehow
-        if args["cparam_filename"] is None:
-            return True  # Treat as generic boxlib data
-        inputs_filename = os.path.join(
-            os.path.dirname(os.path.abspath(output_dir)), args["cparam_filename"]
-        )
-        if not os.path.exists(inputs_filename) and not os.path.exists(jobinfo_filename):
-            return True  # We have no parameters to go off of
-        # If we do have either inputs or jobinfo, we should be deferring to a
-        # different frontend.
-        return False
+        # boxlib datasets are always directories, and
+        # We *know* it's not boxlib if Header doesn't exist.
+        return os.path.exists(header_filename)
 
     def _parse_parameter_file(self):
         """
@@ -1041,24 +1025,16 @@ class OrionDataset(BoxlibDataset):
 
     @classmethod
     def _is_valid(cls, *args, **kwargs):
-        # fill our args
+        if not super(OrionDataset, cls)._is_valid(*args, **kwargs):
+            return False
         output_dir = args[0]
-        # boxlib datasets are always directories
-        if not os.path.isdir(output_dir):
-            return False
-        header_filename = os.path.join(output_dir, "Header")
         jobinfo_filename = os.path.join(output_dir, "job_info")
-        if not os.path.exists(header_filename):
-            # We *know* it's not boxlib if Header doesn't exist.
-            return False
+
         args = inspect.getcallargs(cls.__init__, args, kwargs)
-        # This might need to be localized somehow
         inputs_filename = os.path.join(
             os.path.dirname(os.path.abspath(output_dir)), args["cparam_filename"]
         )
-        if not os.path.exists(inputs_filename):
-            return False
-        if os.path.exists(jobinfo_filename):
+        if not os.path.exists(inputs_filename) or os.path.exists(jobinfo_filename):
             return False
         # Now we check for all the others
         warpx_jobinfo_filename = os.path.join(output_dir, "warpx_job_info")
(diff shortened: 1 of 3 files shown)


neutrinoceros:
Looks stable enough, opening for review 🎉

neutrinoceros:
ping @zingale  and @matthewturk for the Nyx "cosmology info" bug reported in #2808 and partly fixed here already.
Looks to me like this comment may be wrong since we have a sample dataset for Nyx that doesn't contain cosmological info
https://github.com/yt-project/yt/blob/fdc9210a35e41b45c114f271798ea46c2179d2eb/yt/frontends/boxlib/data_structures.py#L1353

could also be that default values for cosmological parameters are used and were not saved here. In which case, we need to reproduce them in the frontend.

neutrinoceros:
>I think this should work, but the one thing I am worried about is that Castro has a mode where it can restart from a MAESTRO/MAESTROeX calculation, so I am wondering if the check on the sub class will get confused there. I think it is clean (atleast git grep seems to think so). But there are dozens of other AMReX codes, and I wonder if they have similar issues.

Thank you so much for this, I wasn't aware (and it shows in how I redesigned the code). Fortunately, the current version shows a weakness with OrionDataset _and_ the original validator for this class has exactly the mechanism we need to introduce the required flexibility. I'll patch this in a few minutes

neutrinoceros:
Well I'd say it would be _much_ easier to introduce the required flexibility for this case after this refactor goes in. Right now I just want to make it more maintainable without breaking existing functionalities. Would it be okay to merge this (once it's _actually_ passing all tests) and keep track of your point as an issue ?

zingale:
yeah, I agree.  Those specific cases would be easier to address.  

How should I test this for you?  Should I just ensure that when I try to open a MAESTROeX or Castro dataset, that it is detected in the right way?

neutrinoceros:
There are tests to do just this in `yt/frontends/boxlib/tests/test_outputs.py` (see for instance `test_CastroDataset`), and I'm adding more with the cases I reported in #2808 . Once I'm happy with this (should take another few minutes), it'd be very helpful if you could bullet proof it a bit further yes :)

neutrinoceros:
Alright this wasn't as easy as I would have hoped but now I'm passing all my local tests _at the same time_, so finger crossed with Jenkins. If you're up for it please attempt to break this !

zingale:
This fails to load a MAESTROeX dataset that `master` loads without issue.  Here's the crash:

```
In [1]: import yt                                                                                            
                    

In [2]: ds = yt.load("reacting_bubble_3d_plt0000000")                                                        
                    
StopIteration                             Traceback (most recent call last)
<ipython-input-2-8fe190cfd475> in <module>
----> 1 ds = yt.load("reacting_bubble_3d_plt0000000")

/raid/zingale/development/yt-mz/yt/yt/convenience.py in load(*args, **kwargs)
     86     candidates = find_lowest_subclasses(candidates)
     87     if len(candidates) == 1:
---> 88         return candidates[0](*args, **kwargs)
     89     if len(candidates) == 0:
     90         if (

/raid/zingale/development/yt-mz/yt/yt/frontends/boxlib/data_structures.py in __init__(self, output_dir, cpara
m_filename, fparam_filename, dataset_type, storage_filename, units_override, unit_system)
   1171     ):
   1172 
-> 1173         super(MaestroDataset, self).__init__(
   1174             output_dir,
   1175             cparam_filename,

/raid/zingale/development/yt-mz/yt/yt/frontends/boxlib/data_structures.py in __init__(self, output_dir, cpara
m_filename, fparam_filename, dataset_type, storage_filename, units_override, unit_system)
    647         self.storage_filename = storage_filename
    648 
--> 649         Dataset.__init__(
    650             self,
    651             output_dir,

/raid/zingale/development/yt-mz/yt/yt/data_objects/static_output.py in __init__(self, filename, dataset_type,
 file_style, units_override, unit_system)
    254         self._create_unit_registry(used_unit_system)
    255 
--> 256         self._parse_parameter_file()
    257         self.set_units()
    258         self._assign_unit_system(unit_system)

/raid/zingale/development/yt-mz/yt/yt/frontends/boxlib/data_structures.py in _parse_parameter_file(self)
   1196                     fields = line.split(":")
   1197                     self.parameters[fields[0]] = fields[1].strip()
-> 1198                 line = next(f)
   1199             # get the runtime parameters
   1200             for line in f:

StopIteration: 

```

zingale:
a link to that file, if you want: http://bender.astro.sunysb.edu/random/reacting_bubble_3d_plt0000000.tgz

zingale:
I can confirm that it seems to work fine on Nyx and Castro

neutrinoceros:
> This fails to load a MAESTROeX dataset that master loads without issue. Here's the crash:

Wait the crash happens in `MaestroDataset._parse_parameter_file` so this looks like _another_ bug (similar to the one I'm trying to fix at the same time for Nyx Data).

I bet that it only loads on master because it _doesn't_ recognise it as a Maestro file (while this branch does !), and returns a simpler BoxlibDataset

zingale:
ah you are right... on `master` it is loaded as a `BoxlibDataset`.  I wonder when that broke.

neutrinoceros:
thanks for reporting back. I'll try to fix this now.

zingale:
note: my Nyx file is loaded as `NyxDataset` and my Castro file is loaded as `CastroDataset`.

neutrinoceros:
Phew, now it should load your Maestro dataset correctly !

zingale:
yep!  I can verify it does.  Awesome job!

yt/frontends/boxlib/data_structures.py
```diff
@@ -1354,24 +1284,30 @@ def _parse_parameter_file(self):
...
+            if has_cosmo_info:
+                for line in f:
+                    if "Omega_m (comoving)" in line:
+                        self.omega_matter = float(line.split(":")[1])
+                    elif "Omega_lambda (comoving)" in line:
+                        self.omega_lambda = float(line.split(":")[1])
+                    elif "h (comoving)" in line:
+                        self.hubble_constant = float(line.split(":")[1])
+            else:
+                raise NotImplementedError
```
neutrinoceros, line 1310: @zingale If nothing else is broken, I would still like to finish the PR with default parameters here. The dataset we have that hits this  is `nyx_small` :)

zingale: sounds good.

I think in the future, I'll do a PR to add some additional MAESTROeX stuff to the frontend, since the format has changed slightly.  But at least with this change detecting it as a MAESTROeX dataset, yt won't get the velocity stuff wrong.

neutrinoceros: so do you know what would be the defaults here ? or if it is in fact possible that this Nyx simulation is not cosmological ?

zingale: I'm really not a Nyx person, but @jmsexton03 is.  Maybe she can answer this?

jmsexton03: I don't know what `nyx_small` is, I do know that the job_info files from Nyx plotfiles I generate will have those cosmology fields in them. I'm not clear on what's broken here.

zingale: looking at that file, it looks like it was produced in `Nyx/Exec/simple_output`, but that directory no longer exists in Nyx.  The file was originally output in 2012, so a lot must have changed.

jmsexton03: Nyx now also includes an omega_radiation cosmology term (almost always 0). If we add that number to job_info in a similar way, can yt store it in the same data structure?
https://github.com/yt-project/yt/blob/fdc9210a35e41b45c114f271798ea46c2179d2eb/yt/utilities/cosmology.py#L41

neutrinoceros:
> I think this should work, but the one thing I am worried about is that Castro has a mode where it can restart from a MAESTRO/MAESTROeX calculation, so I am wondering if the check on the sub class will get confused there. I think it is clean (atleast git grep seems to think so). But there are dozens of other AMReX codes, and I wonder if they have similar issues.


huho, looks like it's happening sooner than I thought: your bubble dataset is now ambiguous in the eye of yt.load if I add the missing bits for AMRexDataset validator since there's both "maestro" and "amrex" in the lowercase version of this : 

```
build dir:     /raid/zingale/development/MAESTROeX/Exec/test_problems/reacting_bubble
AMReX dir:     /home/zingale/development/amrex/
```

Note that you may solve this simply by adding a MaestroexDataset class derived from either (or both) AMRexDataset and MaestroDataset, then yt.load will only select this class as it's more specialized

yt/frontends/boxlib/data_structures.py
neutrinoceros: Sounds like yt is very unlikely to ever come across these outdated format then. I can probably just make this a special case with `cosmological_simulation = 0`

zingale: indeed, I suspect a lot of the sample data is quite outdated.  Maybe we should have a push (separate from this PR) to update some of the datasets to current code output.

jmsexton03: Nyx simulations have had cosmology fields in job_info since 2016 when Nyx was first on github, so it's highly unlikely a Nyx plotfile without cosmology fields in job_info will be used now. It looks like your fix should handle it well.

neutrinoceros:
I fixed a handful of new errors that arose as some of our sample datasets are suddenly being loading with the proper classes. Remaining errors should be failing answer tests and are likely results of this same exact change.

neutrinoceros:
so I think that all remaining failing tests are all yielded by `yt/frontends/boxlib/tests/test_outputs.py:test_beam`
However I note that the dataset in question is correctly read as a `WarpXDataset` instance both on master and this branch, contrary to my original suspicion.
This probably indicates that something is actually broken in the corresponding frontend, or possibly it _was_ broken and I fixed it accidentally so the answers are different ? In any case the change seems real so it warrants for further inspection.

neutrinoceros:
> This probably indicates that something is actually broken in the corresponding frontend, or possibly it was broken and I fixed it accidentally so the answers are different ?

I was able to trace it down and the difference is that, on master, `BoxlibDataset._parse_cparams` doesn't do anything for this dataset (the configuration file isn't correctly passed so everything is left to default values). A side effect of this refactor is that this bug is fixed, so I think those answers actually need an update indeed.

neutrinoceros:
> This probably indicates that something is actually broken in the corresponding frontend, or possibly it was broken and I fixed it accidentally so the answers are different ?

I was able to trace it down and the difference is that, on master, `BoxlibDataset._parse_cparams` doesn't do anything for this dataset (the configuration file isn't correctly passed so everything is left to default values). A side effect of this refactor is that this bug is fixed, so I think those answers actually need an update indeed.

neutrinoceros:
This is now ready for review. May I ask yours, @zingale ? :-)

zingale:
okay, I just confirmed again that this loads Castro and MAESTROeX (and detects the MAESTROeX as a MaestroDataset).

zingale (approved):

PR #2809, opened 2020-08-04
Refactor walk_volume in volume rendering

matthewturk:
## PR Summary

In order to support non-Cartesian volume rendering, we need to be able to switch the volume walk algorithm on the fly.  This sets up the preliminary means to do that, by turning `walk_volume` into a function pointer, and (similar to the smoothing kernel work) using a getter function that looks up the type of geometry and supplies the correct one.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/geometry/selection_routines.pyx
+++ b/yt/geometry/selection_routines.pyx
@@ -28,8 +28,8 @@ from yt.utilities.lib.geometry_utils cimport (
     morton_neighbors_coarse,
     morton_neighbors_refined,
 )
-from yt.utilities.lib.grid_traversal cimport sampler_function, walk_volume
 from yt.utilities.lib.volume_container cimport VolumeContainer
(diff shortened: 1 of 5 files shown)


cphyc:
Note that this PR likely clashes with #2610, so it would be worth joining efforts!

matthewturk:
@cphyc !!! you're right, how did I miss that?!

cphyc:
So after a more thorough inspection, it seems these two PRs aren't clashing that much, but it would be great to disentangle the code so that we have
- a cartesian ray caster on grid-based datasets
- a cylindrical ray caster on grid-based datasets (this PR)
- a cartesian ray caster on oct-based datasets (#2610)

matthewturk:
I'm converting this to draft until after 4.0 goes out so that it won't show up in my filters.  It also needs to be updated with @cphyc 's suggestions and changes.

PR #2815, opened 2020-08-05
fix latex rendering of the transfer function

zingale:
## PR Summary

The exponent in the latex label needs to be in `{}` if it is not a positive single digit, otherwise it does not render entirely as a superscript.

## PR Checklist

- [ ] pass `flake8 yt/`

### DIFF
--- a/doc/source/cookbook/amrkdtree_downsampling.py
+++ b/doc/source/cookbook/amrkdtree_downsampling.py
@@ -71,8 +71,13 @@ tf.add_layers(
     alpha=10.0 * np.ones(4, dtype="float64"),
     colormap="RdBu_r",
 )
-tf.add_layers(4, 0.01, col_bounds=[-27.5, -25.5],
-              alpha=10.0 * np.ones(4, dtype='float64'), colormap='RdBu_r')
+tf.add_layers(
+    4,
+    0.01,
+    col_bounds=[-27.5, -25.5],
+    alpha=10.0 * np.ones(4, dtype="float64"),
+    colormap="RdBu_r",
+)
 sc.save("v4.png", sigma_clip=6.0)
 #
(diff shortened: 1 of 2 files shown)


matthewturk:
after tests, let's merge!

neutrinoceros:
Oops it seems I don't have rights to use this command on someone else's branch.

zingale:
not sure that did anything

doc/source/cookbook/amrkdtree_downsampling.py
```diff
@@ -71,8 +71,13 @@
     alpha=10.0 * np.ones(4, dtype="float64"),
     colormap="RdBu_r",
 )
-tf.add_layers(4, 0.01, col_bounds=[-27.5, -25.5],
-              alpha=10.0 * np.ones(4, dtype='float64'), colormap='RdBu_r')
+tf.add_layers(
```
neutrinoceros, line 74: this showed up in another PR too as an effect of the slash command "format", but it doesn't correspond to a problem seen by black. I checked and this is not even a file we ask black to ignore so I don't really get why it does this. @matthewturk

zingale: yeah, I didn't understand why that showed up

matthewturk: Maybe it's running with a different config.  The invocation is just `black` I think.

https://github.com/yt-project/slash-command-processor/blob/master/.github/workflows/black-command.yml
https://github.com/yt-project/slash-command-processor/blob/master/.github/workflows/format-command.yml

zingale: also, I manually ran `black` as well, since it wasn't clear to me that the command worked the first time.

PR #2817, opened 2020-08-05
[WIP] Pytest for yt4

jcoughlin11:
## PR Summary

This PR is meant to do several things: first, it is meant to update all of the answer tests that have been converted to use pytest for yt4.  Second, it is meant to significantly refactor the frontend answer tests in order to facilitate speedier test runs.

Currently, all of the frontend answer tests have been updated for yt4. The non-frontend answer tests have not (indeed, their pytest implementations have been held back for now in order to try and facilitate an easier review process).

Several of the frontends have also been refactored: `ahf`, `amrvac`, `adaptahop`, `art`, `artio`, and `athena`. The focus for review should be on these frontends until they pass. In particular, it seems that some smaller tweaks that have recently been made to the tests on the nose side of things have slipped through the cracks on the pytest end, which causes several issues. Hopefully with more eyes on this, as well as CI support, these issues will be more readily caught and fixed.

## PR Checklist

- [x] pass `flake8 yt/`


neutrinoceros:
> Several of the frontends have also been refactored: ahf, amrvac, adaptahop, art, artio, and athena. The focus for review should be on these frontends until they pass.

Obviously I'm up to review the AMRVAC refactors :-)
Though I'm not sure I read this right: should we start reviewing now or wait for current or future test jobs to pass ?

yt/frontends/amrvac/tests/test_outputs_pytest.py
```diff
@@ -0,0 +1,252 @@
...
+    assert_allclose_units(ds.mass_unit, YTQuantity(*mass_unit))
+    assert_allclose_units(ds.velocity_unit, YTQuantity(*velocity_unit))
+    assert_allclose_units(ds.pressure_unit, YTQuantity(*pressure_unit))
+    assert_allclose_units(ds.time_unit, YTQuantity(*time_unit))
+    assert_allclose_units(ds.magnetic_unit, YTQuantity(*magnetic_unit))
+
+
+@pytest.mark.answer_test
+@pytest.mark.usefixtures("answer_file", "answer_compare")
+class TestAMRVac:
```
neutrinoceros, line 104: not crucial but no one in this code community would use this case : this name is an acronym, it's not just the "amr" part :)
(if you're curious, VAC means "Versatile Advection Code" and is the name for AMRVAC's ancestor. The full name of the current code is actually MPI-AMRVAC but it doesn't exactly roll on the tongue so everyone uses "AMRVAC")

jcoughlin11:
@neutrinoceros Thanks for the feedback! I added the *_pytest.py files to try and make having the pytest changes coexisting with nose a bit easier. I guess, when it comes time to drop nose, I could just copy and paste pytest content into the current test_outputs.py files? That loses the history of the pytest files, but that history is probably not as important. 

As for whether to review now or after the tests all pass, I'm not really sure which is better. I would guess that the changes that need to be made to get the tests to pass are all (hopefully!) quite small, so a review could probably start now.

yt/frontends/amrvac/tests/test_outputs_pytest.py
jcoughlin11: Gotcha, thanks! I'll change it!

neutrinoceros:
Yeah I guess that would work as a transition strategy, I admit I didn't think about that aspect when I commented. My point still stands that it's pretty hard to review as it stands since GitHub doesn't know it's really a diff and not just new files.
I know it's a burden and I'm sorry to ask but would it be possible to use the old names for your refactors right now and move the contempt of old files to <old_file>_nose.py instead ? That way, reviewing becomes easy again and we keep the flexibility of your transitioning strat (at the cost of adjusting things a bit).

jcoughlin11:
@neutrinoceros I can do that, yeah.

yt/utilities/answer_testing/utils.py
```diff
@@ -17,6 +17,9 @@
 import yaml
 from matplotlib.testing.compare import compare_images
 
+import yt.visualization.particle_plots as particle_plots
+import yt.visualization.plot_window as pw
+import yt.visualization.profile_plotter as profile_plotter
```
neutrinoceros, line 22:
```suggestion
from yt.visualization import particle_plots
import yt.visualization.plot_window as pw
from yt.visualization import profile_plotter
```

munkm:
> That loses the history of the pytest files, but that history is probably not as important.

I mean, since you're changing all of the files in one commit here anyways, it's not like there's a substantial history anyways? Each one should have only one commit

> I could just copy and paste pytest content into the current test_outputs.py files 

Is there a way that you could have two files running lists of which files are run with nose vs. pytest? Then you don't need to have separate copies of each file living in the test directories. As files get updated each list could be modified.

jcoughlin11:
You can definitely tell pytest to ignore certain files and functions. That's sort of what I do in the nose config files and pytest.ini files. Do you mean just have all the tests in one file (test_outputs.py) and ignore certain functions when running pytest and the others when running nose?

munkm:
Yes, I think that's what I'm saying? I'll elaborate below; let me know if this is what you meant. 

 Each frontend would have the same test files that they had pre-pytest (test_outputs.py + whatever else), but with changes within that file (as needed) for pytest functionality. Then the file that says that that file (my_frontend/test_outputs.py) should run with nose would get modified to delete that file, and the file that says it should be run with pytest would get updated. So for a specific frontend ~3 files would change -- the test file, the file that says which files are run with nose, and the file that says which files are run with pytest.

yt/visualization/tests/test_geo_projections.py
```diff
@@ -1,79 +1,58 @@
...
+
+"""
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
```
neutrinoceros, line 12: very quick : this smells like something went wrong during rebase or merging.

jcoughlin11:
@munkm I think I understand what you're saying? Both the nose and pytest tests would live in the same file (my_frontend/test_outputs.py). Then there would be no need for test_outputs_pytest to exist. Then, in the nose.cfg files we would specify the pytest functions to ignore and in the pytest.ini files we would specify the nose tests to ignore, correct? Sorry if I'm misunderstanding!

yt/visualization/tests/test_geo_projections.py
jcoughlin11: @neutrinoceros How do you mean? Is this related to the PR removing these kinds of headers?

neutrinoceros: yes, they were all removed, so I'm worried to see them reappearing here.

neutrinoceros:
@jcoughlin11 since @munkm is on leave atm I'll fill in, but I believe you are missing her point here. My understanding is that she was suggesting to have just one version of each test, and to migrate from nose to pytest at the scale of individual test files. When you're done updating a test file for compatibility with pytest, we can just stop using nose for it and switch to pytest instead. This would make it so each change at the function level would be much easier to track on the diff, and avoid code duplication in the intermediate steps.

yt/visualization/tests/test_geo_projections.py
neutrinoceros: This is one of the files in which you currently have conflicts, so I wouldn't be surprised if that was the reason.

jcoughlin11: Ok, cool, thanks! I'll do another merge with master locally and then push, see where that gets us!

jcoughlin11:
@neutrinoceros Gotcha, thanks!

Xarthisius:
Could you please add `- '--ignore=test_outputs_pytest'` to `tests/tests.yaml` in the `unittests` section at the bottom?

jcoughlin11:
@Xarthisius of course!

yt/frontends/stream/tests/test_outputs_pytest.py
```diff
@@ -53,7 +52,9 @@ class TestStream:
     def test_load_empty_file(self):
         assert_raises(FileNotFoundError, load, "not_a_file")
         assert_raises(
-            YTUnidentifiedDataType, load, tempfile.mkstemp("empty_file", dir=os.getcwd())[1]
+            YTUnidentifiedDataType,
```
neutrinoceros, line 55: Hey Jared, I removed this test a while ago, and reimplemented it in `yt/tests/test_load_errors.py`
I think this is another signal that we should not manage separate files (with the _pytest suffix) in this branch: this particular change is creating an unnecessary conflict with the master branch and I suspect there are more like this one happening within the PR.
I'm sorry, this is obviously a pain but I think that you should consider solving the root problem with how this branch develops before it gets out of control.
If you need help with defining a rebase/merge strategy to sanitize it, don't hesitate to reach out, I'm here to help :)

jcoughlin11: @neutrinoceros I would love some help on coming up with a strategy! I've never actually done any rebasing before, so I'm open to learning!

munkm:
> I'm sorry, this is obviously a pain but I think that you should consider solving the root problem with how this branch develops before it gets out of control. 

I agree with this. I think having single test files rather than duplicates and using a file list to run the tests (as suggested previously) will be a good solution here.

Xarthisius:
@neutrinoceros I'm not sure what conflicts are talking about. Shouldn't they pop up in the GitHub UI?

jcoughlin11:
@Xarthisius pytest has [this](https://github.com/pytest-dev/pytest-xdist) plugin for doing test parallelization. Would you recommend I add that in this PR or just get things working sequentially first and issue a new PR implementing parallelization?

Xarthisius:
> @Xarthisius pytest has [this](https://github.com/pytest-dev/pytest-xdist) plugin for doing test parallelization. Would you recommend I add that in this PR or just get things working sequentially first and issue a new PR implementing parallelization?

If you believe jenkins, the current testsuite ran with nose in serial would take 4h14m. I don't think having a sequential testsuite on master at any time would be pleasant experience for anyone. However, testing things by modifying this PR and seeing what happens on jenkins after 1h is going to be very unpleasant experience for you. What I'd suggest is to test everything locally on your laptop/desktop to see how/if it works before pushing it here.

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

jcoughlin11:
@Xarthisius @munkm Sorry if this was answered elsewhere, but I'm not sure what to do about the answer-store conflict (submodules confuse me), so any help would be greatly appreciated! Thank you!

cphyc (changes requested):
Congrats on getting the tests to pass, that's a big step!

I guess that now's the time for review then, so in order for them to be possible — we really don't want to miss any error there —, would it be possible to address the following points:
- currently some tests are not tested by pytest yet (only unit tests if I understand correctly), and since nosetests has been deactivated on the CI, they are essentially not tested anymore. Would it be possible to leave nosetest activated until the migration to pytest is complete? This means replacing lines like `coverage run $(which pytest) -c pytest_unit.ini` by something like
```bash
coverage run $(which pytest) -c pytest_unit.ini
coverage run $(which nosetests) -c nose_unit.ini
```
- As is, we have a lot of new files (`test_*pytest.py`) that are impossible to compare to their nose counterpart. Could you use the script @neutrinoceros sent you to rename those files and overwrite the old ones made for nose, _while keeping the git history_?
- I see that you applied flynt to `yt/extern`, which leads to lots of changes unrelated to pytest. It'd be great to revert these ones.

jcoughlin11:
@cphyc I'll do that today, thanks! I agree that the history should be kept as much as possible, and I want to make sure I didn't mess anything up and that the tests still give the same answers. I added options, `--raw-answer-store` and `--answer-raw-arrays` to have pytest save the actual arrays produced by the tests to h5 files to try and help make comparing with nose easier, if that helps?

yt/frontends/amrvac/tests/test_outputs_pytest.py
```diff
@@ -0,0 +1,138 @@
...
+            fields.append(("gas", "energy_density"))
+            field_ids.append("energy_density")
+        if "rhod1" in raw_fields_labels:
+            fields.append(("gas", "total_dust_density"))
+            field_ids.append("total_dust_density")
+            # note : not hitting dust velocity fields
+        # return [fields, field_ids]
+        return [ds, fields]
+    except FileNotFoundError:
+        return [None, [None,]]
```
cphyc, line 67: This should rather be moved upward, as close as possible from `ds = utils.data_dir_load(fname)`.

yt/utilities/answer_testing/utils.py
```diff
@@ -206,90 +218,105 @@ def _save_result(data, outputFile):
...
     Parameters
     ----------
-    data : dict
-        Just-generated test results.
+    new_data : dict
+        Just-generated hashed test results.
 
-    outputFile : str
-        Name of file where answers are already saved.
+    old_data: dict
```
cphyc, line 231:
```suggestion
    old_data : dict
```

neutrinoceros:
I'm seeing you're moving the files around as suggested, good !
FYI, and this may be a side effect of my lousy script, some files' renaming seems dubious (for instance `yt/visualization/tests/test_plotwindow_pytest.py` apparently got changed directly to `yt/visualization/tests/test_plotwindow_nose.py` !)

cphyc:
The failing tests on travis are due to the tests timing out. It seems many are run twice.

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
-    for test in small_patch_amr(ds, _get_fields_to_check(ds)):
-        test_rmi_cartesian_dust_2D.__name__ = test.description
-        yield test
+    # return [fields, field_ids]
+    return [ds, fields]
+
+
+def get_pairs():
+    f_list = [_get_fields_to_check(fname) for fname in ds_list]
+    pairs = [(i[0], f) for i in f_list for f in i[1]]
```
neutrinoceros, line 72: This is very difficult to read to me, to the point I'm not even sure I can suggest a better version.

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
+        for lb in ds.domain_left_edge:
+            assert int(lb) == 0
+        for rb in ds.domain_right_edge:
+            assert int(rb) == 2
+        for w in ds.domain_width:
+            assert int(w) == 2
+
+    @requires_file(blastwave_cartesian_3D)
+    def test_grid_attributes(self):
+        # "Check various grid attributes"
```
neutrinoceros, line 98:
```suggestion
        # Check various grid attributes
```

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
+        ds = utils.data_dir_load(blastwave_cartesian_3D)
+        grids = ds.index.grids
+        assert ds.index.max_level == 2
+        for g in grids:
+            assert isinstance(g, AMRVACGrid)
+            assert isinstance(g.LeftEdge, YTArray)
+            assert isinstance(g.RightEdge, YTArray)
+            assert isinstance(g.ActiveDimensions, np.ndarray)
+            assert isinstance(g.Level, (np.int32, np.int64, int))
```
neutrinoceros, line 107: I don't know why such a variability is allowed in typing here, I don't think there's a good reason for it. I'd rather have that test fail so we can fix it. I think the correct type here would be either `int` or `np.int8`.
Note that when we really don't care about the exact type but need to check it's a integer of some sort (for instance when it's user provided data), you can use `isinstance(var, (int, np.integer))`

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
+
+    @pytest.mark.usefixtures("hashing")
+    @pytest.mark.parametrize("ds", ds_list, indirect=True)
+    def test_gh_pr(self, ds):
```
neutrinoceros, line 111:
```suggestion
    def test_grid_hierarchy_parentage_relashionships(self, ds):
```

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
+        self.hashes.update({"grid_hierarchy": grid_hierarchy(ds)})
+        self.hashes.update({"parentage_relationships": parentage_relationships(ds)})
+
+    @pytest.mark.usefixtures("hashing")
+    @pytest.mark.parametrize("ds, f", get_pairs(), indirect=True)
+    def test_gv(self, f, ds):
```
neutrinoceros, line 117:
```suggestion
    def test_grid_values(self, f, ds):
```

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
+        if ds and f is None:
+            pytest.skip("Data not found.")
+        self.hashes.update({"grid_values": grid_values(ds, f)})
+
+    @pytest.mark.usefixtures("hashing")
+    @pytest.mark.parametrize("ds, f", get_pairs(), indirect=True)
+    @pytest.mark.parametrize("d", dso_list, indirect=True)
+    def test_fv(self, d, f, ds):
```
neutrinoceros, line 125:
```suggestion
    def test_field_values(self, d, f, ds):
```

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
+    jet_cylindrical_25D,
+    riemann_cartesian_175D,
+    blastwave_cartesian_3D,
+    blastwave_polar_2D,
+    blastwave_cylindrical_3D,
+    rmi_cartesian_dust_2D,
+]
+a_list = [0, 1, 2]
+w_list = [None, "density"]
+dso_list = [None, ("sphere", ("max", (0.1, "unitary")))]
```
neutrinoceros, line 42: I don't have a clue what those lists represent or are for (even after reading further in). Rather than having their type `list` as part of their names, I would like to have descriptive names here.

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
+    answer_file = None
+    saved_hashes = None
+
+    @requires_file(khi_cartesian_2D)
+    def test_AMRVACDataset(self):
+        assert isinstance(utils.data_dir_load(khi_cartesian_2D), AMRVACDataset)
+
+    @utils.requires_ds(blastwave_cartesian_3D)
+    def test_domain_size(self):
+        # "Check for correct box size, see bw_3d.par"
```
neutrinoceros, line 87:
```suggestion
        # Check for correct box size, see bw_3d.par
```

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -21,118 +26,113 @@
...
+        self.hashes.update({"field_values": field_values(ds, f, d)})
+
+    @pytest.mark.usefixtures("hashing")
+    @pytest.mark.parametrize("ds, f", get_pairs(), indirect=True)
+    @pytest.mark.parametrize("d", dso_list, indirect=True)
+    @pytest.mark.parametrize("a", a_list, indirect=True)
+    @pytest.mark.parametrize("w", w_list, indirect=True)
+    def test_pv(self, a, d, w, f, ds):
+        if ds and f is None:
+            pytest.skip("Data not found.")
```
neutrinoceros, line 137: Once again I'm having a hard time understanding what's happening here (because I can't read get_pairs). I would suggest specifying in the skip message exactly what data is missing but I'm not sure how.

neutrinoceros (commented):
Just reviewing the AMRVAC tests here. So in general I really like how you managed to deduplicate much of the code there, but I have a few comments on readability because I think it ought to be improved to ease maintenance later on.

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -13,589 +14,494 @@
...
+            "particle_prop_0_2",
+            "particle_prop_0_3",
+            "particle_prop_0_4",
+            "particle_prop_0_5",
+            "particle_prop_0_6",
+            "particle_velocity_x",
+            "particle_velocity_y",
+            "particle_velocity_z",
+        ]
+        # Check that sinks are autodetected
```
cphyc, line 192: I know this isn't you, but wouldn't it be possible to parametrize this test (with `@pytest.mark.parametrize("ds", [ramses_sink, output_00080], indirect=True)`) and remove the code duplication below?

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -13,589 +14,494 @@
...
+            assert ("sink", field) in ds.field_list
+            # test that field access works
+            ad["sink", field]
+        # Checking that sinks are autodetected
+        ds = yt.load(output_00080)
+        ad = ds.all_data()
+        for field in expected_fields:
+            assert ("sink", field) not in ds.field_list
+
+    @requires_file(ramses_new_format)
```
cphyc, line 205: Any reason here and below not to use test parametrization instead of `@requires_file`? I am a bit confused by the consistency of these test calls, but I guess there is something I don't quite understand.

Note that if you change this test, you likely want to update some of the followings ones.

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -13,589 +14,494 @@
...
+    def test_output_00080(self, a, d, w, f, ds):
+        ppv = pixelized_projection_values(ds, a, f, w, d)
+        self.hashes.update({"pixelized_projection_values": ppv})
+        fv = field_values(ds, f, d)
+        self.hashes.update({"field_values": fv})
+        dobj = create_obj(ds, d)
+        s1 = dobj["ones"].sum()
+        s2 = sum(mask.sum() for block, mask in dobj.blocks)
+        assert_equal(s1, s2)
+        assert_equal(ds.particle_type_counts, {"io": 1090895, "nbody": 0})
```
cphyc, line 50:
```suggestion
        @pytest.mark.parametrize("ds", [output_00080], indirect=True)
    def test_output_00080(self, ds, axis, field, weight_field, data_object):
        ppv = pixelized_projection_values(ds, axis, field, weight_field, data_object)
        self.hashes.update({"pixelized_projection_values": ppv})
        fv = field_values(ds, field, data_object)
        self.hashes.update({"field_values": fv})
        dobj = create_obj(ds, data_object)
        s1 = dobj["ones"].sum()
        s2 = sum(mask.sum() for block, mask in dobj.blocks)
        assert_equal(s1, s2)
        assert_equal(ds.particle_type_counts, {"io": 1090895, "nbody": 0})
```

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -13,589 +14,494 @@
...
-def test_namelist_reading_should_not_fail():
-
-    for ds_name in (ramses_empty_record, output_00080):
-        # Test that the reading does not fail for malformed namelist.txt files
-        ds = data_dir_load(ds_name)
-        ds.index  # should work
-
-
-ramses_mhd_128 = "ramses_mhd_128/output_00027/info_00027.txt"
+@pytest.mark.answer_test
```
cphyc, line 34: I believe only the first test of this test suite is an answer test. The other are unit tests.

yt/frontends/adaptahop/tests/test_outputs.py
```diff
@@ -1,80 +1,59 @@
...
-
-    # Check sphere contains all the members
-    id_sphere = halo.sphere["io", "particle_identity"].astype(int)
-    assert len(np.lib.arraysetops.setdiff1d(members, id_sphere)) == 0
-
-    # Check region contains *only* halo particles
-    id_reg = np.sort(halo["io", "particle_identity"].astype(int))
-
-    np.testing.assert_equal(members, id_reg)
+@pytest.mark.answer_test
```
cphyc, line 12: Shouldn't it be a unit test?

cphyc (commented):
I skimmed through the adaptahop and RAMSES tests and left a bunch of comment. In general, I am still a bit confused with the usage of `@requires_ds` vs. test parametrization.

I think it would simplify the writing of tests if we were to use `@pytest.mark.parametrize` everywhere, and remove as many `@requires_file` as possible to only keep those tests for which we test the behaviour of `yt.load` (or using directly the dataset class).

yt/frontends/amrvac/tests/test_outputs.py
```diff
@@ -1,15 +1,20 @@
...
 
 import yt  # NOQA
 from yt.frontends.amrvac.api import AMRVACDataset, AMRVACGrid
 from yt.testing import requires_file
 from yt.units import YTArray
-from yt.utilities.answer_testing.framework import (
-    data_dir_load,
-    requires_ds,
-    small_patch_amr,
+from yt.utilities.answer_testing import utils
```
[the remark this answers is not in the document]
jcoughlin11: I agree. I was thinking of renaming it `testing_utilities`. Is that better?

jcoughlin11:
@cphyc They should be run twice, right? Once for nose and once for pytest?

cphyc:
> @cphyc They should be run twice, right? Once for nose and once for pytest?

Mmmh, I actually didn't think at it this way. I don't know whether we should run all tests with both nose and pytest in the transition process. @neutrinoceros @munkm @Xarthisius what's your opinion?

In any case, we ought to run them as independent jobs on travis.

```diff
diff --git a/.travis.yml b/.travis.yml
index cc65867e5..fe2498d67 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,33 +68,49 @@ jobs:
       name: "Python: 3.6 Minimal Dependency Unit Tests"
       python: 3.6
       env: MINIMAL=1
-      script: >-
-        coverage run $(which pytest) -c pytest_unit.ini;
-        coverage run $(which nosetests) -c nose_unit.cfg -c nose_ignore.cfg
+      script:
+        coverage run $(which pytest) -c pytest_unit.ini
 
     - stage: tests
       name: "Python: 3.6 Unit Tests"
       python: 3.6
-      script: >-
-        coverage run $(which pytest) -c pytest_unit.ini;
-        coverage run $(which nosetests) -c nose_unit.cfg -c nose_ignore.cfg
+      script:
+        coverage run $(which pytest) -c pytest_unit.ini
 
     - stage: tests
       name: "Python: 3.8 Unit Tests"
       python: 3.8
-      script: >-
-        coverage run $(which pytest) -c pytest_unit.ini;
-        coverage run $(which nosetests) -c nose_unit.cfg -c nose_ignore.cfg
+      script:
+        coverage run $(which pytest) -c pytest_unit.ini
 
     - stage: tests
       name: "Python: 3.7 Answer Tests"
       python: 3.7
-      script: >-
-        coverage run $(which pytest) -c pytest_answer.ini;
-        coverage run $(which nosetests) -c nose_answer.cfg -c nose_ignore.cfg
+      script:
+        coverage run $(which pytest) -c pytest_answer.ini
       # This script has not yet been converted to use pytest, so skip for now
       # after_failure: python tests/report_failed_answers.py -f -m --xunit-file "answer_pytest.xml"
 
+      # !!! REMOVE THIS ONCE THE TRANSITION TO PYTEST IS OVER
+    - stage: nose tests
+      name: "Python: 3.6 Minimal Dependency Unit Tests (nose)"
+      python: 3.6
+      env: MINIMAL=1
+      script:
+        coverage run $(which nosetests) -c nose_unit.cfg -c nose_ignore.cfg
+
+    - stage: nose tests
+      name: "Python: 3.8 Unit Tests (nose)"
+      python: 3.8
+      script:
+        coverage run $(which nosetests) -c nose_unit.cfg -c nose_ignore.cfg
+
+    - stage: nose tests
+      name: "Python: 3.7 Answer Tests (nose)"
+      python: 3.7
+      script:
+        coverage run $(which nosetests) -c nose_answer.cfg -c nose_ignore.cfg
+
 after_success:
   - codecov
```

yt/frontends/amrvac/tests/test_outputs.py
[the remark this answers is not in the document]
neutrinoceros: how about moving `yt.utilities.answer_testing` to `yt.testing` ? (this should be done in a different PR though, this one has enough going on as is :) )
I also don't understand why we have a `yt.utilities.answer_testing.utils` and  `yt.utilities.answer_testing.framework`. Do you expect those two could be merged in the future ?

yt/frontends/ramses/tests/test_outputs.py
neutrinoceros: In Jared's defense, I also blindly assume that `frontends/<x>/tests/test_outputs.py` is dedicated to answer tests. I don't know if we have an actual rule for this.

neutrinoceros:
> I don't know whether we should run all tests with both nose and pytest in the transition process. @neutrinoceros @munkm @Xarthisius what's your opinion?
In any case, we ought to run them as independent jobs on travis.

I agree with that last part. I think it's okay for this one PR to have twice as many test run for now but I don't see the point of keeping the nose tests around for much longer. I do want to see one run with nose running in isolation and no failures before I'm conformable with removing those though :)

yt/frontends/amrvac/tests/test_outputs.py
jcoughlin11: @neutrinoceros It is hard to read, I agree. Basically, I needed `(ds, field)` pairs and I couldn't find an easy way to "distribute" the dataset object to each field required by that dataset. I looked at itertools, but didn't see anything (though it's possible I missed the solution). Basically, in the original version, each function was calling the exact same code, but for a different dataset (e.g., `test_blastwave_cartesian_3D` and `test_blastwave_spherical_2D` have the exact same code). The goal here was to streamline this into one function that could be parametrized by pytest to work for each dataset. The issue was that, as said above, I needed to pair each dataset with each of the fields corresponding to that, hence the gross list comprehension.

neutrinoceros: You may want to take a look a `more_itertools`, though I can't recall off the top of my head if they have something similar. It's not part of the standard library, but it's already part of a yt-compatible env thanks to transitive dependencies.
Even if there's no better approach we can think of, I find it confusing that `i` iterates over `f_list` while `f` iterates over `i[1]`. Also I don't know exactly what `f` is supposed to represent. I'm guessing a clearer refactor would look something like
```suggestion
    pairs = [(ftype, fname) for (ftype, fnames) in f_list for f in fnames]
```

pytest_unit.ini
```diff
@@ -0,0 +1,9 @@
+[pytest]
+# Ignore those files that have tests that contain nose-specific code
+addopts = 
+    -s
+    -v
+    -rs
+    --ignore-glob='*_nose.py'
+    --ignore='yt/frontends/owls_subfind/tests/test_outputs.py'
+    --ignore='yt/frontends/ramses/tests/test_outputs.py'
```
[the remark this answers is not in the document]
jcoughlin11: @cphyc They are, but last I heard they weren't working, so I disabled them. If they're fixed, then that line should be removed (same for owls_subfind)

tests/test_minimal_requirements.txt
```diff
@@ -4,7 +4,6 @@ sympy==1.2
 nose==1.3.7
 nose-timer==0.7.3
 pytest~=5.2; python_version >= '3.0'
```
neutrinoceros, line 6:
```suggestion
pytest~=5.2
```
Have you tried the latest pytest version (6.1.0 was just released) ?

jcoughlin11: I haven't. I can, though. I'll have to check out what's new and make sure stuff doesn't break

tests/nose_runner.py
```diff
@@ -132,3 +134,15 @@ def generate_tasks_input():
     while num_jobs:
         result = results.get()
         num_jobs -= 1
+
+    pytest_args = [
+        f"--local-dir={os.path.join(ytcfg.get('yt', 'test_data_dir'), 'answers')}",
+        "-c=pytest_answer.ini",
+        "--junitxml=unittests.xml",
+        "--answer-big-data",
+        # f"-n {int(os.environ.get('NUM_WORKERS', 6))}",
```
Xarthisius, line 143: Can you just enable it and see what happens? It will require adding `pytest-xdist` to `tests/test_requirements.txt` if you haven't done it already.

Xarthisius: Also please remove all the code related to nose from this script. When we talked about simultaneously running nose and pytest for some time, I meant running them as independent test suites, not in a single run.

jcoughlin11: @Xarthisius sounds good! Should I instead move all of the pytest code to a file called, say, `pytest_runner.py` and leave the nose stuff as is? Then jenkins could run `python pytest_runner.py` or `python nose_runner.py` as needed? Further, would this entail commenting out the nose-specific test jobs in `.travis.yaml` (so that the two test suites are not run in a single run)?

tests/test_minimal_requirements.txt
jcoughlin11: @neutrinoceros After a cursory investigation, it seems as if pytest 6.1.0 works

cphyc:
I am sorry @jcoughlin11, I pushed (by mistake) onto your branch. This contains a fix for travis (the indentation was wrong), sorry if this causes any issue on your side!

jcoughlin11:
@cphyc No worries! I had indented the file because I thought it needed to be, so this fixes that

Xarthisius:
@yt-fido test this please

pytest_answer.ini
```diff
@@ -0,0 +1,15 @@
...
+# Ignore those files that have tests that contain nose-specific code
+# For the first run, we need to create the gold standard answers
+# NOT running the big data tests for now
+addopts = 
+    -s
+    -v
+    --with-answer-testing
+    -rs
+    --answer-store
+    --force-overwrite
```
Xarthisius, line 12: @jcoughlin11 Correct me if I'm wrong, but does it mean that currently tests are overwriting the results with every run instead of comparing the answers?

jcoughlin11: @Xarthisius Whoops! Yes, you're right, they are overwriting! That's my fault! I needed to generate a set of answers and then I thought I deleted those options, but apparently I did not. I'll delete them and re-push (though, locally, I'm having an issue with the hashes of ImageArrays)

tests/nose_runner.py
```diff
@@ -1,134 +1,16 @@
...
-        "--with-answer-testing",
+if __name__ == "__main__":
+    pytest_args = [
+        f"--local-dir={os.path.join(ytcfg.get('yt', 'test_data_dir'), 'answers')}",
+        "-c=pytest_answer.ini",
+        "--junitxml=unittests.xml",
         "--answer-big-data",
-        "--local",
+        f"-n {int(os.environ.get('NUM_WORKERS', 6))}",
+        "--dist loadscope",
```
Xarthisius, line 14:
```suggestion
        "--dist=loadscope",
```
Otherwise `pytest.main()` treats it as a separate arg and tries to execute test called `loadscope` which doesn't exist and thus no tests are being run on Jenkins...

jcoughlin11: Gotcha, thanks! I'll also regenerate the answers, just to be safe

PR #2818, opened 2020-08-06
Rename YTOutputNotIdentified to YTUnidentifiedDataType

neutrinoceros:
## PR Summary

This is a follow up to #2695.
Rationale:
In discussing #2695 with @matthewturk , I realised the word "output" is a bit weird in the context this error is used, since the data in question is an _input_ to `yt.load()` (but in general is an _output_ from a simulation, hence the term).

This change makes the original error a subclass of the new one, so as not to break any downstream code that relies on catching it.

Opening as a draft until #2695 is accepted.

## PR Checklist

- [x] pass `flake8 yt/`


neutrinoceros:
Note that I'd rather not merge this before the yt.loaders PRs !

Issue #2757, opened 2020-07-17
transitioning to fstrings for string formating ?


neutrinoceros:
### Bug report

**Bug summary**
fstrings were introduced in python 3.6, which is the current oldest python version supported by yt (at least on CI).
They are the most readable, concise _and_ more efficient way to format strings, as opposed to `%` and `str.format` older methods.
[flynt](https://github.com/ikamensh/flynt) is a tool that automates the conversion to fstrings. Unfortunately, their usage is not recommended with `logging`, because it is optimised to evaluate formatted strings late, depending on the logger level, only when the message is actually going to be outputted, and it can't do that with fstrings.

So I ran a experimental PR here https://github.com/yt-project/yt/pull/2605, but it this operation would require manually selecting non-logging related changes, so it turns out to be much less efficient than I originally hoped.

I currently don't have a solution to automate the process so I'm not going to perform the switch.
I think this transition would be the first actual breaking change for python 3.5 if it's not actually broken yet. Just for this reason I think it's best to hold up until we have a global solution before we start using them punctually.

PR #2820, opened 2020-08-06
[f-string 1/2] Remove mylog formatting strings

cphyc:
## PR Summary

Replaces all occurrence of `mylog.info("Some information message about %s %s" % ("foo", "bar"))` by `mylog.info("Some information message about %s %s", "foo", "bar")`

## PR Checklist

- [x] pass `flake8 yt/`
- [x] revert insane loglevel to initial value _once tests pass_

## Extra information

This relies on the following code to detect the occurences:
```python
import argparse
import ast
import os
import re
from glob import glob

parser = argparse.ArgumentParser()
parser.add_argument("folder", type=str, default=".")
parser.add_argument("-v", "--verbose", action="store_true")


class MyLogVisitor(ast.NodeVisitor):
    def __init__(self, filename, source):
        self.filename = filename
        self.source = source

    def visit_Call(self, node):
        if (
            isinstance(node, ast.Call)
            and isinstance(node.func, ast.Attribute)
            and hasattr(node.func.value, "id")
            and node.func.value.id == "mylog"
            and node.func.attr in ("debug", "info", "warn", "error", "warning", "critical")
            and not isinstance(node.args[0], ast.Str)
        ):
            lstart = max(0, node.lineno - 2)
            lend = min(len(self.source), node.lineno + 2)
            print(f"{self.filename}:{node.lineno}:{node.col_offset} error")

        self.generic_visit(node)


def main():
    args = parser.parse_args()

    # Find filenames
    if os.path.isfile(args.folder):
        filenames = [args.folder]
    else:
        filenames = []
        for root, dirs, files in os.walk(args.folder):
            if '__pycache__' in root:
                continue
            for fn in files:
                if fn.endswith('.py'):
                    path = os.path.join(root, fn)
                    filenames.append(path)
        filenames = sorted(filenames)

    # Find filenames
    for fn in filenames:
        if args.verbose:
            print(f"Analysing {fn}")
        with open(fn) as f:
            line_by_line = f.readlines()
            f.seek(0)
            lines = f.read()
        try:
            tree = ast.parse(lines, fn)
        except SyntaxError as e:
            print(e)
        mlv = MyLogVisitor(fn, line_by_line)
        mlv.visit(tree)


if __name__ == "__main__":
    main()
```

If you're using vs code, you can also add a problem detector using
```json
{
    // See https://go.microsoft.com/fwlink/?LinkId=733558
    // for the documentation about the tasks.json format
    "version": "2.0.0",
    "tasks": [
        {
            "label": "find bad mylog calls",
            "type": "shell",
            "command": "python /path/to/find_mylog_errors.py .",
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "isBackground": false,
            "problemMatcher": {
                "owner": "python",
                "fileLocation": ["relative", "${workspaceFolder}"],
                "pattern": {
                  "regexp": "^(.*):(\\d+):(\\d+) \\s+(warning|error)$",
                  "file": 1,
                  "line": 2,
                  "column": 3,
                  "severity": 4
                }
              }
        }
    ]
}
```


neutrinoceros:
Looks like this would enable solving #2757 with a mere call to `flynt yt/` !!! Love it.

neutrinoceros:
Thanks for including the detection script, this allowed me to catch this line
```python
            and node.func.attr in ("debug", "info", "warn", "error")
```
where you forgot to include the fifth and ultimate logging level `"critical"` ! Also `"warning"` exists and I'm pretty sure it's still used in yt (though it's a deprecated alias for `"warn"`, see https://docs.python.org/3/library/logging.html#logging.Logger.warning)

cphyc:
> Thanks for including the detection script, this allowed me to catch this line
> 
> ```python
>             and node.func.attr in ("debug", "info", "warn", "error")
> ```
> 
> where you forgot to include the fifth and ultimate logging level `"critical"` ! Also `"warning"` exists and I'm pretty sure it's still used in yt (though it's a deprecated alias for `"warn"`, see https://docs.python.org/3/library/logging.html#logging.Logger.warning)
> 
> edit: my bad, it's the other way around, `warn` is the deprecated one. It was addressed in #2285, but we still have 4 occurrences of it in the code base (two of which are in the RAMSES frontend stuck_out_tongue )

Good catch, and you're right that I missed a handful of them!

neutrinoceros:
`mylog.critical` is actually never used atm, so it doesn't make a difference actually, but I see you caught a lot more anyway, well done !

.git-blame-ignore-revs
```diff
@@ -2,8 +2,13 @@
 7edfcee093cca277307aabdb180e0ffc69768291
 81418e459f16c48d6b7a75d6ef8035dfe9651b39
 
-# transisiton to black
+# transition to black
```
neutrinoceros, line 5: oops, this is on me, thanks !

yt/funcs.py
```diff
@@ -238,14 +238,6 @@ def check_parallel_rank(*args, **kwargs):
     return check_parallel_rank
 
 
-def rootloginfo(*args):
```
cphyc, line 241: Note: this was not used anywhere.

setupext.py
```diff
@@ -113,13 +113,13 @@ def check_for_openmp():
...
-                    "OpenMP program (output was {0})".format(output)
+                    "OpenMP program (output was %s)",
+                    output,
                 )
                 using_openmp = False
         else:
             log.warn(
-                "Unexpected output from test OpenMP "
-                "program (output was {0})".format(output)
+                "Unexpected output from test OpenMP " "program (output was %s)", output
```
neutrinoceros, line 122:
```suggestion
                "Unexpected output from test OpenMP program (output was %s)", output
```

cphyc:
@matthewturk @neutrinoceros before merging this, do you think we could set the default logging level to 40 (and then revert it) so that _all_ logs are visited and we're sure nothing is going wrong?

neutrinoceros:
sure ! may as well go for 50 ! :-) (disregard if I'm mistaken, I _think_ this is the maximal meaningful level)

matthewturk:
I think it is!  Let's try it.

yt/visualization/volume_rendering/old_camera.py
```diff
@@ -2370,8 +2370,8 @@ def __init__(self, *args, **kwargs):
         if self.disparity <= 0.0:
             self.disparity = self.width[0] / 1000.0
             mylog.info(
-                "Warning: Invalid value of disparity; "
-                "now reset it to %f" % self.disparity
+                "Warning: Invalid value of disparity; " "now reset it to %f",
```
neutrinoceros, line 2373:
```suggestion
                "Warning: Invalid value of disparity; now reset it to %f",
```

cphyc:
I'm converting to draft until the tests are done, after that I'll revert 19f78f1 and that'll be good to go!

cphyc:
Good to go now!

neutrinoceros:
Awesome ! There's a huge queue on CI runner right now, so I'll happily merge this in 10 to 12 hours when it clears.
For immediate reference, is this the travis job that ran with all log entries activated ?
https://tests.yt-project.org/job/yt_py38_git/928/

cphyc:
Yes ; it ran successfully before reverting the commit that was (temporarily) setting the default log value to 50.

PR #2821, opened 2020-08-06
[f-string 2/2] apply flynt

cphyc:
# PR Summary

Follows #2820. This PR should not touch *any* mylog calls.

Note that this cannot be really automated, as we don't really want f-strings in mylog.XXX calls (see #2820), which itself cannot be easily automated.

## PR Checklist

- [x] pass `flake8 yt/`
- [x] #2820 is merged


neutrinoceros:
supersedes #2605 and fixes #2757 🎉

neutrinoceros:
> Note that this cannot be really automated, as we don't really want f-strings in mylog.XXX calls (see #2820), which itself cannot be easily automated.

So I pondered a bit on this and I think that if we add `flynt --fail-on-change yt` to the linting stage with this change, it's actually a pretty solid detection method for the future and it _would_ catch wrongly formatted logging messages such as the ones you corrected in the previous PR.

The only case of bad code that wouldn't be caught by flynt (seems to me) would be logging messages using fstrings, so _maybe_ it's easier to catch just that with a slight change of your custom tool, and add it to CI too ?

cphyc:
> > Note that this cannot be really automated, as we don't really want f-strings in mylog.XXX calls (see #2820), which itself cannot be easily automated.
> 
> So I pondered a bit on this and I think that if we add `flynt --fail-on-change yt` to the linting stage with this change, it's actually a pretty solid detection method for the future and it _would_ catch wrongly formatted logging messages such as the ones you corrected in the previous PR.
> 
> The only case of "bad code" that wouldn't be caught by flynt (seems to me) would be logging messages using fstrings, so _maybe_ it's easier to catch just that with a slight change of your custom tool, and add it to CI too ?
> 
> Note that you'd need to update the contribution guide to state we only use fstrings now.

My "tool" has lots of false-positive and may have some false-negative as well, so I wouldn't be confident to use it on the entire codebase.
Regarding flynt, I'm happy to have it as part of our linting pass. Note that currently flynt doesn't have a "dry-run" mode, which make the whole process much more complicated. If https://github.com/ikamensh/flynt/pull/72 (adds support for "dry-run" mode) ever gets merged, than we could rely on the next release of flynt.

cphyc:
Note to reviewers: this PR is 100% automatic, so once all tests pass, this should be good to go.

yt/fields/xray_emission_fields.py
```diff
@@ -83,7 +83,7 @@ class XrayEmissivityIntegrator:
     def __init__(self, table_type, redshift=0.0, data_dir=None, use_metals=True):
 
         filename = _get_data_file(table_type, data_dir=data_dir)
-        only_on_root(mylog.info, "Loading emissivity data from %s." % filename)
+        only_on_root(mylog.info, f"Loading emissivity data from {filename}.")
```
munkm, line 86: I think this is a mylog statement?

matthewturk:
I tried to push to your branch, but it is giving me permission errors.  All that needs to happen is removing test pickle.

yt/fields/xray_emission_fields.py
neutrinoceros: yup, but in this case because it's wrapped, the string formatting is evaluated early no matter what, so may as well use a fstring. :)

munkm: right. I'm just pointing out a mylog instance since the PR specifically says this shouldn't touch mylog calls.

matthewturk: OK if we really want to get pedantic, the *correct* fix should be

```suggestion
        only_on_root(mylog.info, "Loading emissivity data from %s", filename)
```

since `only_on_root` has this signature:

```
def only_on_root(func, *args, **kwargs):
```

neutrinoceros: Isn't being pedantic the whole point here ? 😄

munkm: Maybe we should just delete this line 😈 😈 😈 😈 😈

matthewturk: image

munkm: Hi @cphyc we're all here to offer you only the *best* PR commentary. 🍾 🥂

neutrinoceros: AFAIC Matt won by a landslide.

matthewturk:
post-tests I think we should merge

neutrinoceros:
About 8 hours to go before Appveyor's run completes. If all goes well, I should be able to merge this in about 10 to 12 hours

neutrinoceros:
@cphyc one failure on Travis :/
Looks like the "haunted" test we saw the other, @matthewturk

neutrinoceros:
Once again I can't reproduce this locally (MacOS, Python 3.7 and matplotlib 3.3... in an env created specifically for this) :/

cphyc:
I can't reproduce the error. Is it real?

neutrinoceros:
I suspect this could be a very niche regression in matplotlib 3.3.0 on macOS with whatever backend we use on Travis, which doesn't have graphic outputs. Though I don't get why it's not showing on every other PRs as well.

munkm:
Agh, sorry it looks like the removed module that we merged an hour ago or so has a conflict now with this PR. Can you resolve it, and then we can merge?

neutrinoceros:
> agh, sorry it looks like the removed module that we merged an hour ago or so has a conflict now with this PR. Can you resolve it, and then we can merge?

I also had to rebase 3 of 4 PRs because of this. For a dead file this guy sure grabs a lot attention 😄

cphyc:
I decided to merge instead of rebasing so that I don't have to deal with conflicts and change the commit in the file listing commits to ignore when git blaming. Note that locally, if I run
```
flynt --fail-on-change -e yt/extern
```
everything runs smoothly!

munkm:
> I decided to merge instead of rebasing so that I don't have to deal with conflicts and change the commit in the file listing commits to ignore when git blaming

Seems totally reasonable to me!

neutrinoceros:
> Note that locally, if I run
> 
> ```
> flynt yt/ --fail-on-change -e yt/extern
> ```

Oh right ! the externs ! have you made sure that we also exclude those in Travis checks, the PR template and the guidelines ? I don't think I've seen since in the next PR !

cphyc:
> > Note that locally, if I run
> > ```
> > flynt yt/ --fail-on-change -e yt/extern
> > ```
> 
> Oh right ! the externs ! have you made sure that we also exclude those in Travis checks, the PR template and the guidelines ? I don't think I've seen since in the next PR !

- PR template b4a2bcd477cbd373f046dab4ed2efff2ef2bab65
- travis linting stage b4a2bcd477cbd373f046dab4ed2efff2ef2bab65
- Contributing b4a2bcd477cbd373f046dab4ed2efff2ef2bab65
- github action https://github.com/yt-project/slash-command-processor/commit/8b573e102c2ab4db554b3a9435d91cd650b6fec9

Am I missing one?

neutrinoceros:
> Am I missing one?

I think you're good ! Well played 🚀

PR #2822, opened 2020-08-06
remove unused module

neutrinoceros:
## PR Summary

Working on #2744, I think I found a module/class (`yt/geometry/object_finding_mixin.py`) that is not used anymore, and doesn't look user-facing either.
I propose to remove it.


matthewturk:
Yeah, it's probably good to go.

PR #2825, opened 2020-08-07
[f-string 2/2 | alternative] apply flynt

cphyc:
Same as #2821 , but also use the `--transform-concats` argument (with a line-length of 88).


cphyc:
Closing this in favour of #2821 .

PR #2826, opened 2020-08-07
add 2 labels to mergeable list

neutrinoceros:
mark ~'proposal'~, 'dead code' and 'refactor' as acceptable labels in the eye of the "mergeable" bot.

### DIFF
--- a/.github/mergeable.yml
+++ b/.github/mergeable.yml
@@ -9,5 +9,5 @@ mergeable:
           message: "WIP pull requests can't be merged."
       - do: label
         must_include:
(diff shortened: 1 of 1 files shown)


munkm:
I don't think proposal should be a mergable label? I think proposal should be used when somebody is proposing a semi-major code change in an issue before opening a PR and it is agreed upon by the community. Won't almost all proposals be an enhancement of some kind?

matthewturk:
I kinda agree, the proposal one shouldn't be mergable...

neutrinoceros:
I actually don't have a strong opinion on this, I reverted it !

PR #2828, opened 2020-08-07
Rearrange data objects

matthewturk:
This is a strict refactoring which moves objects from selection data objects into a subdirectory, `selection_objects`.  It is in support of #2827 .

Issue #2827, opened 2020-08-07
Refactor and rearrange code for clarity


matthewturk:
This is a proposal; I've started a draft PR as well which I will edit this to reference.

We have a couple really big files in yt.  I'd like to propose that we begin to extract them into subfiles, to make it easier to modify them, and to identify where one begins and another ends.  The three files in question I would like to apply this to first are:

Filename | Lines
-|-
`yt/data_objects/data_containers.py` | 3024
`yt/data_objects/construction_data_containers.py` | 2845
`yt/data_objects/selection_data_containers.py` | 1332

Additionally, it seems that the `yt/visualization/` directory could use a rearrangement, which I will detail a bit further down.

Ideally, this would be a *strict* refactoring -- there would be *no* changes in functionality, and all tests should continue to pass.  Furthermore, because this would be largely in areas of the code that don't see much churn, it should avoid conflicts with existing PRs.

neutrinoceros:
update:

Filename | Lines
-|-
`yt/data_objects/data_containers.py` | 1538 (-1486)
`yt/data_objects/construction_data_containers.py` | 2869 (+24)
`yt/data_objects/selection_data_containers.py` | 0 (file deleted)

I think this can be closed now that #2828 and #2829 were merged ?

matthewturk:
I do think this can be, but I wonder if we should expand this to include things like the visualization tools.  I suppose that could be a different PR+Issue combo, though!

PR #2829, opened 2020-08-07
Rearrange index component objects

matthewturk:
In support of #2827, this moves the data objects that are strictly subobjects of index objects (i.e., grid patches, unstructured mesh index objects, etc) into their own package and subdirectory.

This may need rebasing or reworking if #2828 is accepted.


cphyc (approved):

yt/data_objects/index_subobjects/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .grid_patch import AMRGridPatch
```
cphyc, line 1: Is this required? It seems the only place it is imported from is in `yt/data_objects/api.py` above.

cphyc (approved):

matthewturk:
I think I added this to make sure that the object registered correctly.

cphyc:
> I think I added this to make sure that the object registered correctly.

Fair enough!

PR #2831, opened 2020-08-08
add a yt.loaders (api) module

neutrinoceros:
## PR Summary

We have a bunch of user-facing functions named `load_something` but they currently live in 3 different parts of the code. Namely
- yt.convenience
- yt.utilities
- yt.frontends.stream.data_structures

I'd like to propose we gather them together in a unified `yt.loaders` module. This offers the following benefits

- make it easier to maintain or improve the consistency between these functions' respective apis.
- make them easier to discover to a contributor
- provide a single namespace to import from within the codebase

This PR is a first step in that direction and adds a `yt/loaders.py` file which plays a similar role`api.py` files: it's basically just another namespace with no new definitions of its own.

Eventually, I'd like to actually move the `load_*` functions to this module, but this is not required to open the discussion and I believe this first change has value as a standalone.

A side effect of this unification is that `yt.simulation` should be renamed `yt.load_simulation`, which is done in this PR. I also keep the old alias with a deprecation warning.

## PR Checklist

- [x] pass `flake8 yt/`


munkm:
I think in the future this type of code change (where quite a bit of code is going to be shifted around) should be an issue with a "proposal" label to start -- that way the community can decide together a bit in advance how we want to structure some of this.

neutrinoceros:
Oh yeah, totally makes sense. This wasn't much longer to do than writing an issue though, so I don't really mind if it's not accepted. I get your point that issues are more appropriate for discussion starters in general though :)

neutrinoceros:
I renounced changing the internal imports to point toward this module before the functions are actually defined there because it creates circular imports.
And it doesn't look like the Jenkins failures are real... @yt-fido test this please

matthewturk:
I think this is a good path forward, and one that could align with *both* restructuring the `Stream` frontend so that it is less "do everything" and also consolidate lots of these functions where they can more easily be maintained.  If this is ready for review I'm happy to take a look.

neutrinoceros:
> If this is ready for review I'm happy to take a look.

it is !
The follow-up where I actually start moving functions around isn't though (#2831)

neutrinoceros:
Note: this branch didn't need rebasing per se, but the followup one did, so I rebased both of them onto master. Otherwise this one is still good to go (needs one more approval I think)

PR #2834, opened 2020-08-10
remove a test that's been disabled for 7 yrs

neutrinoceros:
## PR Summary

A dead test, probably better to remove it than trying to revive it after so long.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/utilities/tests/test_amr_kdtree.py
+++ b/yt/utilities/tests/test_amr_kdtree.py
@@ -2,49 +2,7 @@ import itertools
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


neutrinoceros:
my reasoning exactly :-)

matthewturk:
so have we checked that it fails?

neutrinoceros:
Nope, but it relies on refine_amr, which is going to be removed :) (and that's how I discovered it)

matthewturk:
oh, got it.  Thanks.

PR #2835, opened 2020-08-10
make enable_plugin() raise an error instead of a login entry

neutrinoceros:
## PR Summary

I authored this logging error and didn't realise at the time it would be systematically seen in test pipelines when yt is called on the command line.
This should fix this issue.

## PR Checklist

- [x] pass `flake8 yt/`


neutrinoceros:
maybe the actual problem is that `ytcfg.getboolean("yt", "loadfieldplugins")` defaults to `True` despite the face that this parameter is only kept for backwards compatibility ?

matthewturk:
One thing we could do is have it default to something else, and if it's unset and a plugin file is found, have it report a deprecation warning.  That would catch the "default" problem.

neutrinoceros:
Good idea, what do you think now ?

yt/__init__.py
```diff
@@ -176,3 +176,16 @@
 from yt.units.unit_systems import UnitSystem, unit_system_registry
 
 _called_from_pytest = False
+
+
+def _check_deprecated_parameters():
+    from yt.config import ytcfg
+    from yt.funcs import issue_deprecation_warning
+
+    if ytcfg.getboolean("yt", "loadfieldplugins"):
```
munkm, line 185: If this load field plugins was added since the last release I don't think we need to issue a deprecation warning.

neutrinoceros: Indeed it seems to be the opposite: this is an old option that apparently fell out of use but we have a few checks still around for backward compatibility
See this comment
https://github.com/yt-project/yt/blob/7f488b615a7cf42bdabb46eeb4eb7a259665b4e4/yt/utilities/command_line.py#L48

neutrinoceros:
@yt-fido test this please

neutrinoceros:
> I was secretly hoping someone would fix this.

Me too actually... until I found out it was my fault 🙈

yt/__init__.py
munkm: ahhh ok I thought it had been more recently added!

PR #2836, opened 2020-08-10
[bugfix] fix particle selection for clump objects

brittonsmith:
## PR Summary

Running the `find_clumps.py` cookbook recipe, doing the following will give an error:
```
leaf_clumps[0].quantities.total_mass()
```
Error: http://paste.yt-project.org/show/265/

This fixes that by returning an empty mask if there are no levels relevant to the clump. I've verified manually that this gives the same answer as the brute force method that is used if `scipy` is not installed. The lines added to cookbook recipe can serve as a test.

## PR Checklist

- [ ] pass `flake8 yt/`


yt/data_objects/selection_data_containers.py
```diff
@@ -1163,12 +1163,16 @@ def _part_ind_KDTree(self, ptype):
             ],
             axis=1,
         ).value
+
+        mask = np.zeros(ppos.shape[0], dtype=bool)
+
         levels = self[("index", "grid_level")].astype("int32").value
+        if levels.size == 0:
```
matthewturk, line 1170: So this should trigger if there are no cells, right?  But `mask` here is the particle position?  So won't it give a differently shaped array in this case?  Or is the particle size the correct array shape?

brittonsmith: @cphyc likely knows better than me, but I think we only get the levels to get the grid cells on that level. We then use the kdtree to check the distances of those cells to the particle and mask out the ones that are too far away. If the levels array is empty, then we don't want to go into that loop at all.

Xarthisius:
@yt-fido test this please

brittonsmith:
I just added a WIP to this because I've noticed another issue. If one loops over all clumps instead of just the leaf clumps, a different error appears. I don't have time to investigate this in detail, so I will create an issue.

brittonsmith:
@matthewturk, changing the chunking style to `'all'` as per our discussion in Issue #2841 seemed like the easiest solution. The clump finder does not run in parallel, so I hope this is an acceptable solution, at least for now.

brittonsmith:
@yt-fido test this please

yt/data_objects/selection_data_containers.py
cphyc: Yup, if there are no levels then we cannot loop over them and the code will likely fail.

cphyc (approved):
This looks good to me!

PR #2840, opened 2020-08-11
add pooch and pyyaml to test_requirements

neutrinoceros:
## PR Summary

In order to build a fully fletched dev env for yt, `pip install -r tests/tests_requirement.txt` is currently insufficient; Namely, I've identified two missing packages that are used in tests: pooch, and pyyaml 
This fixes it.

### DIFF
--- a/tests/test_requirements.txt
+++ b/tests/test_requirements.txt
@@ -29,3 +29,5 @@ f90nml>=1.1.2
 MiniballCpp>=0.2.1
 pooch>=0.7.0
 pykdtree==1.3.1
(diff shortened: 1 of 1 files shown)


neutrinoceros:
wait. Actually both packages are already in there..., I went too fast.

neutrinoceros:
So I think the problem I got was in fact due to the fact that cartopy failed to install and then pip sorta gave up on the rest, so I ended up with an incomplete environment. The real issue here is cartopy, but I think it's a known one.

PR #2842, opened 2020-08-11
Add a check for reduced dimensionality slices

matthewturk:
This adds a check to the `SliceSelector`.  When we are slicing along an axis that is greater than or equal to the dimensionality of the dataset, we want to always say "yes" we are selecting a zone.

Note that I did not modify the grid selection, since it *should* work out of the box with the way we clip to the edges.

There may be other objects that need this, but Slice is the most obvious one from my reading.

This should fix the Phantom Bugs in the other PRs, and I've also added a test (that fails without this changeset) to verify it's working.

### DIFF
--- a/yt/geometry/selection_routines.pyx
+++ b/yt/geometry/selection_routines.pyx
@@ -1473,10 +1473,19 @@ cdef class SliceSelector(SelectorObject):
     cdef int axis
     cdef np.float64_t coord
     cdef int ax, ay
+    cdef int reduced_dimensionality
 
     def __init__(self, dobj):
         self.axis = dobj.axis
         self.coord = _ensure_code(dobj.coord)
+        # If we have a reduced dimensionality dataset, we want to avoid any
+        # checks against it in the axes that are beyond its dimensionality.
+        # This means that if we have a 2D dataset, *all* slices along z will
+        # select all the zones.
+        if self.axis >= dobj.ds.dimensionality:
+            self.reduced_dimensionality = 1
+        else:
+            self.reduced_dimensionality = 0
 
         self.ax = (self.axis+1) % 3
         self.ay = (self.axis+2) % 3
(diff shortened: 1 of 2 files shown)


yt/geometry/selection_routines.pyx
```diff
@@ -1473,10 +1473,19 @@ cdef class SliceSelector(SelectorObject):
...
+    cdef int reduced_dimensionality
 
     def __init__(self, dobj):
         self.axis = dobj.axis
         self.coord = _ensure_code(dobj.coord)
+        # If we have a reduced dimensionality dataset, we want to avoid any
+        # checks against it in the axes that are beyond its dimensionality.
+        # This means that if we have a 2D dataset, *all* slices along z will
+        # select all the zones.
+        if self.axis >= dobj.ds.dimensionality:
```
Xarthisius, line 1485: I know folks that really love `x-z` domains and do a lot of `y`-slices :-)

matthewturk: So I *think* in that case, we'll have set `geometry = ("cartesian", "xzy")` and so when we slice along coord 2 it will still represent `y`, right?

yt/geometry/selection_routines.pyx
```diff
@@ -1473,10 +1473,19 @@ cdef class SliceSelector(SelectorObject):
...
+            self.reduced_dimensionality = 1
```
neutrinoceros, line 1486: So this attribute is essentially a boolean, right ? Any reason not to use True and False as values ? (maybe that's not recommended in Cython)

matthewturk: Ah!  So `bool` doesn't work as a cdef.  I think `bint` might but I was hoping to keep it as close to C and not C++ as possible.

neutrinoceros: alright that works for me !

Issue #2780, opened 2020-07-24
Coverage badge is broken


neutrinoceros:
### Bug report

I'm just reporting here that the coverage badge in `README.md` is currently displaying "0%" which fortunately is not true.

neutrinoceros:
closing after merging https://github.com/yt-project/yt/pull/2846

PR #2846, opened 2020-08-12
Deactivate coverage badge

neutrinoceros:
## PR Summary

The coverage badge is currently essentially useless since we do not upload test reports at the moment.
I think it'd be preferable to deactivate it until we have a way of generating coverage reports that doesn't significantly impede the CI speed. 

Not really a "fix", but related to https://github.com/yt-project/yt/issues/2780


matthewturk:
ping @Xarthisius

PR #2832, opened 2020-08-09
Populate yt.loaders module (followup #2831)

neutrinoceros:
## PR Summary

Follow up to #2831: move loader functions to yt.loaders and redirect internal imports
(note that this could be merged first now that it is stabilised, making #2831 redundant)

As a side benefit, I slightly refactored the load_sample function (gave more sensible names to some inner variables and a helper class).
Another side effect is the removal of the `refine_amr` function, which implementation was very tricky to isolate without creating circular imports, and which apparently fell out of use. This removal and that of the affected test lines is the reason for the significant line count diff (~ -220 lines).

## PR Checklist

- [x] pass `black --check yt/`
- [x] pass `isort . --check --diff`
- [x] pass `flake8 yt/`
- [x] isolate stream.data_stuctures.py content in 3 files
- [x] relocate stream loaders to yt.loaders (refactoring needed to avoid circular imports)
- [x] relocate convenience loaders (trivial)
- [x] relocate load_sample
- [x] ~relocate all load tests to yt/tests/test_loaders.py~ I now think it's okay to leave them where they are actually
- [x] cleanup stream.api
- [x] update documentation references
- [x] reactivate flake8 for yt.loaders

PR #2847, opened 2020-08-12
Fix some issues with stream frontend non-cartesian particles

matthewturk:
## PR Summary

The stream frontend -- specifically for AMR and unigrid datasets -- was not respecting the axis names for particles.  This adds the ability to use particles in r, theta, phi, and also adds some tests.

## PR Checklist

- [x] pass `flake8 yt/`


neutrinoceros:
Note that this conflicts with https://github.com/yt-project/yt/pull/2832
I'm happy to backport the bugfix afterwards if that's the easy way out

yt/frontends/stream/data_structures.py
```diff
@@ -439,9 +443,13 @@ def assign_particle_data(ds, pdata, bbox):
             grid = {"number_of_particles": 0}
             grid_pdata.append(grid)
 
+        particle_index_fields = [
```
neutrinoceros, line 446: looks to me that this repetition isn't necessary. You could define `particle_index_fields` as a property
```python
@property
def particle_index_fields(self):
    return [f"particle_position_{ax}" for ax in ds.coordinates.axis_order] + ["particle_position"]

```
and then reuse this downstream

yt/frontends/stream/data_structures.py
```diff
@@ -439,9 +443,13 @@ def assign_particle_data(ds, pdata, bbox):
...
+            f"particle_position_{ax}" for ax in ds.coordinates.axis_order
+        ]
         for ptype in ds.particle_types_raw:
             if (ptype, "particle_position_x") in pdata:
-                x, y, z = (pdata[ptype, f"particle_position_{ax}"] for ax in "xyz")
+                # we call them x, y, z even though they may be different field names
+                x, y, z = (pdata[ptype, pi_field] for pi_field in particle_index_fields)
             elif (ptype, "particle_position") in pdata:
```
neutrinoceros, line 453: for instance this elif becomes redundant with my suggestion above

yt/frontends/stream/io.py
```diff
@@ -62,7 +62,10 @@ def _read_particle_coords(self, chunks, ptf):
                     if (ptype, "particle_position") in gf:
                         x, y, z = gf[ptype, "particle_position"].T
                     else:
-                        x, y, z = (gf[ptype, f"particle_position_{ax}"] for ax in "xyz")
+                        x, y, z = (
+                            gf[ptype, f"particle_position_{ax}"]
+                            for ax in self.ds.coordinates.axis_order
+                        )
```
neutrinoceros, line 68: this generator would stay but be changed to            
```suggestion
                        x, y, z = (gf[ptype, pfield] for field in self.particle_index_fields)
```

yt/frontends/stream/tests/test_stream_particles.py
```diff
@@ -456,3 +456,73 @@ def test_stream_sph_projection():
...
     assert image.max() > 0
     assert image.shape == (256, 256)
+
+
+def test_stream_non_cartesian_particles_unigrid():
+    eps = 1e-6
+    r, theta, phi = np.mgrid[
+        0.0 : 1.0 - eps : 64j, 0.0 : np.pi - eps : 64j, 0.0 : 2.0 * np.pi - eps : 64j
+    ]
+    np.random.seed(0x4D3D3D3)
```
neutrinoceros, line 466: Now Tain I could get into.

yt/frontends/stream/tests/test_stream_particles.py
```diff
@@ -456,3 +456,73 @@ def test_stream_sph_projection():
     image = frb["gas", "density"]
     assert image.max() > 0
     assert image.shape == (256, 256)
+
+
+def test_stream_non_cartesian_particles_unigrid():
```
neutrinoceros, line 461: note for the future: code duplication in these tests could be dealt with in one pytest fixture
Maybe add this as an inline comment for future-us ?
```suggestion
@pytest.mark.parametrize("loader", (load_uniform_grid, load_amr_grid))
def test_stream_non_cartesian_particles(loader):
```

yt/frontends/stream/data_structures.py
matthewturk: So, yes, I agree... sort of.

What I'm nervous about here is the idea that we're edging closer and closer to refactoring into having index fields listed, which is a *good* thing, but which I would like to do much more deliberately.

neutrinoceros: that's a valid point. Unfortunately I lack a vision of the big picture here so I'm focusing on the scale of this single PR instead. I agree that it'd be much better to start from the ground up if we're going to refactor this at a lower level.

neutrinoceros:
Please have a look at my suggestions _before_ rebasing :)

matthewturk:
I will address these comments, but I still want this to go in.

matthewturk:
(After I address the comments, I mean.)

munkm:
Hey @matthewturk how is this PR going? I'm curious how the tests will handle if it's updated to the current version of `main`

matthewturk:
I've merged, which involved some transplanting of changes, so hopefully that went correctly.

munkm (approved):
This looks good to me!

neutrinoceros:
closing/reopening to refresh CI logs

neutrinoceros:
This seems pretty easy to salvage: there is one minor conflict and some trivial NameErrors to fix. Because this is a bug fix and it has one approval already, I'll apply the minimal amount of changes to get CI green myself and set it for auto-merge.

neutrinoceros:
There. I disabled auto-merged after all because we haven't seen the tests go green for a while here and I don't want to risk merging in an unstable state again.
I'm also not 100% confident that this will survive CI in the current state because it introduces some pytest-only tests to an existing test module which isn't yet blacklisted from nosetest, but I think that may be the only change still needed, and I'll do it if necessary

neutrinoceros (approved):

Issue #2430, opened 2020-01-28
bugs: ds.derived_field_list gets populated with unavailable fields


neutrinoceros:
##  Bug report

**Bug summary**
simply calling
```python
ds = yt.load(mydataset)
ad = ds.all_data()
ad[field]
```

will raise errors (and corrupt the ad object, apparently) for some values of `field`.
For now I just want to report what I know is not working, using two public datasets.

**Code for reproduction**
Here's how I get the list of buggy fields
```python
from pprint import pprint
def get_buggy_fields(ds):
    buggy_fields = []
    ad = ds.all_data()
    for f in ds.derived_field_list:
        try:
            ad[f]
        except Exception as e:
            buggy_fields.append((f, e))
            # reinit the ad object, otherwise all subsequent fields will be seen as buggy with this method
            ad = ds.all_data()
    return buggy_fields
```

two examples (using datasets with different geometries)
```python
ds = yt.load("amrvac/kh_2d0000.dat") # cartesian 2D
pprint(get_buggy_fields(ds))

# output
[(('amrvac', 'vertex_x'),
  AttributeError("'AMRVACHierarchy' object has no attribute 'meshes'")),
 (('amrvac', 'vertex_y'),
  AttributeError("'AMRVACHierarchy' object has no attribute 'meshes'")),
 (('amrvac', 'vertex_z'),
  AttributeError("'AMRVACHierarchy' object has no attribute 'meshes'")),
 (('gas', 'averaged_density'),
  RuntimeError('Error: yt attempted to read outside the boundaries of a non-periodic domain along dimension 1.\nRegion left edge = -0.16666666666666666 code_length, Region right edge = 1.1666666666666665 code_length\nDataset left edge = 0.0 code_length, Dataset right edge = 4.0 code_length\n\nThis commonly happens when trying to compute ghost cells up to the domain boundary. Two possible solutions are to load a smaller region that does not border the edge or override the periodicity for this dataset.')),
 (('gas', 'density_gradient_magnitude'),
  RuntimeError('Error: yt attempted to read outside the boundaries of a non-periodic domain along dimension 1.\nRegion left edge = -0.16666666666666666 code_length, Region right edge = 1.1666666666666665 code_length\nDataset left edge = 0.0 code_length, Dataset right edge = 4.0 code_length\n\nThis commonly happens when trying to compute ghost cells up to the domain boundary. Two possible solutions are to load a smaller region that does not border the edge or override the periodicity for this dataset.')),
[...]

PR #2848, opened 2020-08-12
hotfix: register "vertex" fields only in unstructured meshes

neutrinoceros:
## PR Summary

fix https://github.com/yt-project/yt/issues/2430

This is a hotfix, I'm convinced there are better ways to solve this but they would require much more work (refactoring).

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -499,13 +499,29 @@ class FieldInfoContainer(dict):
         # this violates isolation principles and should be refactored
         dfl = set(self.ds.derived_field_list).union(deps.keys())
         dfl = list(sorted(dfl, key=tupleize))
+
         if not hasattr(self.ds.index, "meshes"):
             # the meshes attribute characterizes a unstructured-mesh data structure
 
             # ideally this fitlering should not be required
             # and this could maybe be handled in fi.get_dependencies
             # but it's a lot easier to do here
-            dfl = [(ft, fn) for ft, fn in dfl if "vertex" not in fn]
+
+            filtered_dfl = []
+            for field in dfl:
+                try:
+                    ftype, fname = field
+                    if "vertex" in fname:
+                        continue
+                except ValueError:
+                    # in very rare cases, there can a field represented by a single
+                    # string, like "emissivity"
+                    # this try block _should_ be removed and the error fixed upstream
+                    # for reference, a test that would break is
+                    # yt/data_objects/tests/test_fluxes.py::ExporterTests
+                    pass
+                filtered_dfl.append(field)
+            dfl = filtered_dfl
 
         self.ds.derived_field_list = dfl
         return deps, unavailable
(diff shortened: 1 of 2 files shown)


yt/fields/field_info_container.py
```diff
@@ -469,8 +471,9 @@ def check_derived_fields(self, fields_to_check=None):
         for field in fields_to_check:
             fi = self[field]
             try:
+                # fd: field detector
                 fd = fi.get_dependencies(ds=self.ds)
-            except Exception as e:
+            except (YTFieldNotFound, NeedsConfiguration, UnitConversionError) as e:
```
neutrinoceros, line 476: this is not strictly needed here, but as a side effect of exploring this function I figured I might as well catch relevant exceptions only. If I've missed any, tests should tell !

neutrinoceros:
added a test that fails on master and passes here

neutrinoceros:
switching to draft since #2850 will make this a tiny bit simpler.

neutrinoceros:
So I think I'm going to split this between the hotfix and the experimental part where I build a long exception list to avoid `except Exception`

neutrinoceros:
moved the exception expansion there #2851

matthewturk:
Hm, I guess I replied over email but it didn't get through.  Anyway, I think there's a possibility we could have vertex-centered fields in some grid calculations.  Does this preclude that?

neutrinoceros:
I don't think so. My understanding is that vertex fields only make sense for data where `ds.index.meshes` is defined. If I'm wrong then this needs to be updated of course.
More broadly, I think since is one more case where having specialised abstract Dataset subclasses could help.

matthewturk:
@neutrinoceros As long as this does not *preclude* vertex centered fields (which might be handled with some weird logic at present) for grid fields, this looks good to me and I think it can go in.

matthewturk (approved):

neutrinoceros:
I don't really know how to test this, or how to demonstrate this is fine

matthewturk:
OK, let's hold off, it may not be the case that there are any grid
frontends to be worried about at the moment with this issue.

yt/fields/field_info_container.py
```diff
@@ -493,6 +494,34 @@ def check_derived_fields(self, fields_to_check=None):
...
+        # now populate the derived field list with results
+        # this violates isolation principles and should be refactored
         dfl = set(self.ds.derived_field_list).union(deps.keys())
-        self.ds.derived_field_list = list(sorted(dfl, key=tupleize))
+        dfl = list(sorted(dfl, key=tupleize))
+
+        if not hasattr(self.ds.index, "meshes"):
+            # the meshes attribute characterizes a unstructured-mesh data structure
+
+            # ideally this fitlering should not be required
```
cphyc, line 506:
```suggestion
            # the meshes attribute characterizes an unstructured-mesh data structure

            # ideally this filtering should not be required
```

cphyc (approved):

yt/fields/field_info_container.py
neutrinoceros: thanks for your careful review !

PR #2850, opened 2020-08-12
fix: unbounded variable in shear calculation

neutrinoceros:
## PR Summary

Working on #2848 I opened Pandora's box and found several bugs in shear computation.
This fixes them.
Namely, the function used dimensionality checks where really, what we want is to know wether velocity component are available. The difference is important for AMRVAC datasets, for instance, where one can run a model in a 2D box, but all velocity components are still present.
In particular, if called with 1D data, the intermediate variable `f` would end up unbounded (which is how I discovered this).

So in summary:
- correctly include all terms in shear computation for datasets with more velocity components than dimensions (so mostly AMRVAC I think)
- raise a reasonable error on purpose (`ValueError`) if called with data that only has 1 velocity component, instead of a defacto `UnboundedLocalError`
- Acknowledge the functionality isn't supported yet for non-cartesian data with an NotImplementedError

The reason these changes shouldn't break anything is that the one time this internal function is called is inside a try block that currently catches everything, so introducing early exception raising here is not a problem.

## PR Checklist 

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/fields/fluid_vector_fields.py
+++ b/yt/fields/fluid_vector_fields.py
@@ -513,7 +513,10 @@ def setup_fluid_vector_fields(registry, ftype="gas", slice_info=None):
             vx = data[ftype, "relative_velocity_x"]
             vy = data[ftype, "relative_velocity_y"]
             dvydx = (
-                vy[sl_right, sl_center, sl_center] - vy[sl_left, sl_center, sl_center]
+                # tmp edit to see if tests catch it
+                vy[
+                    sl_right, sl_center, sl_center
+                ]  # - vy[sl_left, sl_center, sl_center]
             ) / div_fac
             dvxdy = (
                 vx[sl_center, sl_right, sl_center] - vx[sl_center, sl_left, sl_center]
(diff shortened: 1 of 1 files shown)


munkm:
Ok, I'll do my best to review this soon. Would you mind updating the PR description to list (maybe a bulleted list) of all of the bugs that this PR addresses so it's easier to summarize it when we do a release? 🙂

neutrinoceros:
done !

neutrinoceros:
I'll add tests now and you guys can merge when they pass :)

neutrinoceros:
Ah I just noticed that the _shear_mach definition is almost an exact duplicate of this one and has the same issues.
I should probably address this here too. Switching to draft for now, I'd like to see if we have tests that will detect if I intentionally break this definition.

neutrinoceros:
Looks like it's not covered. For the time being, I'll just reapply the same fix there, but I think there's room for a larger code deduplication in this file, that I will attempt to do later.

neutrinoceros:
for reference, this dataset would be perfect to test some of the edge cases here:
https://github.com/yt-project/website/pull/86

neutrinoceros:
The travis failure looks like another random fail like the one @matthewturk fixed this week. Reported https://github.com/yt-project/yt/issues/2861

PR #2855, opened 2020-08-13
Avoid generating kdtree until it's really required. Fixes #2854

### LINKED ISSUE #2854: KDTree is always generated for a particle dataset
Since https://github.com/yt-project/yt/pull/2645 KDTree is always generated for particle datasets, regardless whether it's needed or not. It's caused by:

https://github.com/yt-project/yt/blob/64159003d0fc3cf371402f59738b40635e945727/yt/frontends/sph/data_structures.py#L86-L87

`kdtree` property is used as an argument to `_generate_smoothing_length` and thus KDTree is always generated even if `_generate_smoothing_length` detects that hsml are present and simply returns.

Xarthisius:
## PR Summary

This PR defers the generation of the KDTree in `_generate_smoothing_length` until all checks for existing hsmls fail, by simply passing the index as an argument.

## PR Checklist

- [ ] pass `flake8 yt/`

### DIFF
--- a/yt/frontends/arepo/io.py
+++ b/yt/frontends/arepo/io.py
@@ -7,7 +7,7 @@ from yt.utilities.on_demand_imports import _h5py as h5py
 class IOHandlerArepoHDF5(IOHandlerGadgetHDF5):
     _dataset_type = "arepo_hdf5"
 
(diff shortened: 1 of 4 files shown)


neutrinoceros:
We got three approvals so I think we can let @matthewturk off the hook.

Issue #2859, opened 2020-08-14
Too many status updates printed when processing fields in AMRVAC data


Kiradorn:
### Bug report

**Bug summary**

When processing AMRVAC data that lacks ghostfield information, currently yt prints many many instances of warnings that there are gaps in the data. Please may these status updates be modified to RuntimeWarnings instead?

**Code for reproduction**

File for reproduction has already been requested to be added to the online repository but here's the tarball: http://use.yt/upload/1cbbc47e

```python
overrides = dict(length_unit=(1e8, 'cm'), temperature_unit=(1e6, 'K'), numberdensity_unit=(1e8,'cm**-3'))
ds=yt.load('file.dat', units_override=overrides)
ds.periodicity = 3*[True]
p=yt.plot_2d(ds, 'vorticity_magnitude')
p.show()
```

**Actual outcome**

```
yt : [WARNING  ] 2020-08-14 09:33:26,380 Overriding code units: Use this option only if you know that the dataset doesn't define the units correctly or at all.
yt : [INFO     ] 2020-08-14 09:33:26,381 Overriding length_unit: 1e+08 cm.
yt : [INFO     ] 2020-08-14 09:33:26,382 Overriding temperature_unit: 1e+06 K.
yt : [INFO     ] 2020-08-14 09:33:26,384 Overriding numberdensity_unit: 1e+08 cm**-3.
yt : [INFO     ] 2020-08-14 09:33:26,442 Parameters: current_time              = 925.0
yt : [INFO     ] 2020-08-14 09:33:26,443 Parameters: domain_dimensions         = [528 528   1]
yt : [INFO     ] 2020-08-14 09:33:26,444 Parameters: domain_left_edge          = [-12.   0.   0.]
yt : [INFO     ] 2020-08-14 09:33:26,445 Parameters: domain_right_edge         = [12. 25.  1.]
yt : [INFO     ] 2020-08-14 09:33:26,446 Parameters: cosmological_simulation   = 0
yt : [INFO     ] 2020-08-14 09:33:27,311 Using full MHD energy for thermal pressure.
/Users/kiradorn/codes/yt/yt/frontends/amrvac/data_structures.py:67: RuntimeWarning: ghost-zones interpolation/smoothing is not currently supported for AMRVAC data.
  category=RuntimeWarning,
yt : [INFO     ] 2020-08-14 09:33:28,824 zeros found in density, patching them to compute corresponding velocity field.
[...]

PR #2860, opened 2020-08-14
update amrvac error printing

Kiradorn:
## PR Summary
Relegates the notification of zero-density to RuntimeWarning to prevent the copious printing of the same error. Expect this fix to be temporary in the event a debug option is preferred.

Addresses the error in issue #2859.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/frontends/amrvac/fields.py
+++ b/yt/frontends/amrvac/fields.py
@@ -42,9 +42,8 @@ def _velocity(field, data, idir, prefix=None):
     mask1 = rho == 0
     if mask1.any():
         warnings.warn(
-            "zeros found in %sdensity, "
+            f"zeros found in {prefix}density"
             "patching them to compute corresponding velocity field.",
-            prefix,
             category=RuntimeWarning
         )
         mask2 = moment == 0
--- a/yt/frontends/amrvac/fields.py
+++ b/yt/frontends/amrvac/fields.py
@@ -7,6 +7,7 @@ AMRVAC-specific fields
 import functools
 
 import numpy as np
+import warnings
 
 from yt import mylog
 from yt.fields.field_info_container import FieldInfoContainer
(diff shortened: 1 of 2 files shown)


yt/frontends/amrvac/fields.py
```diff
@@ -40,10 +41,11 @@ def _velocity(field, data, idir, prefix=None):
 
     mask1 = rho == 0
     if mask1.any():
-        mylog.info(
+        warnings.warn(
             "zeros found in %sdensity, "
             "patching them to compute corresponding velocity field.",
             prefix,
```
neutrinoceros, line 47:
```suggestion
            f"zeros found in {prefix}density"
            "patching them to compute corresponding velocity field.",
```

yt/frontends/amrvac/fields.py
```diff
@@ -40,10 +41,10 @@ def _velocity(field, data, idir, prefix=None):
 
     mask1 = rho == 0
     if mask1.any():
-        mylog.info(
-            "zeros found in %sdensity, "
+        warnings.warn(
```
munkm, line 44: So, is there a reason this is being switched over to a warning? My impression is that warnings.warn should be used if the user can do something about it, and that logging.warning should be used if the event should be noted but isn't something that can be changed by the user?  

The table [in the logging docs](https://docs.python.org/3/howto/logging.html#when-to-use-logging) about issuing warnings distinguishes these events.

munkm: I see the point in the PR description about the copious printing, but I wonder if there is a solution consistent with the use of logger/warning

neutrinoceros: That's an excellent point I didn't think of. For reference, here's an issue to discuss it more thoroughly https://github.com/yt-project/yt/issues/2865

neutrinoceros:
Closing this since #2859 was resolved more globally in #2871

Issue #2839, opened 2020-08-11
Use GitHub actions for linting


neutrinoceros:
I propose we separate linting from testing by using GitHub actions instead of Travis for this stage.
This would require adding a `.github/workflows/style-checks.yaml`

Here's the documentation for reference https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow

This would allow Travis' tests to complete even when style checks are failing. Since Appevyor is usually our bottleneck and isn't blocked by style checks anyway, this shouldn't be a blocker.

 More importantly, this would have the benefit of automatically _notifying_ the dev that those checks are failing (GitHub notification + email depending on the account's setup).

Note that I included a flynt stage, this isn't to be used before https://github.com/yt-project/yt/pull/2824/files is merged :)

neutrinoceros:
I've experimented with this in a trash fork (second account) here https://github.com/neutrinoceros2/yt/actions
and it works pretty nicely.
All checks are run in parallel in under 1 min and you get an email notification if anything goes wrong !

PR #2862, opened 2020-08-14
add a style check GH workflow

neutrinoceros:
## PR Summary

fix https://github.com/yt-project/yt/issues/2839

Here's what it look like when triggered on a PR against a master branch that already includes this file
https://github.com/neutrinoceros2/yt/pull/1

## PR Checklist

- [x] remove Travis linting

### DIFF
--- a/.github/workflows/style-checks.yaml
+++ b/.github/workflows/style-checks.yaml
@@ -1,0 +1,78 @@
+name: Style Checks
+on: [push]
+
+jobs:
+  flake8:
+    name: flake8
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Setup Python
+      uses: actions/setup-python@master
+      with:
+        python-version: 3.8
+    - name: install
+      id: install
+      run : pip install -r tests/lint_requirements.txt
+
+    - name: check
+      id: flake8
+      run: |
+        flake8 --version
+        flake8 yt/
+
+  black:
+    name: black
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Welp so the checks are not triggered here, which means only the master version of workflows is used (makes a lot of sense actually).

matthewturk:
Any chance you could see what the time savings is if we consolidate the steps into a single job, so we only checkout once?

neutrinoceros:
In fact I _tried_ to do it as a single job originally, but it's actually longer because then it doesn't run in parallel.
Also, it's not desirable because in that config it just reports the first error and then gives up on other checkers.
What we're missing here is a "stage", not a job, but it seems to me that there's no such thing currently in GitHub workflows.

matthewturk:
Ah, got it!

neutrinoceros:
Too bad this hit the current random failure on travis... It's not affecting the codebase itself but because it _does_ affect the conf file for Travis, maybe the run should be retriggered ? Though I'm not sure if flipping a coin is worth 2.5 h of cpu time tbh.

.github/workflows/style-checks.yaml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+    - uses: actions/checkout@master
+    - name: Setup Python
+      uses: actions/setup-python@master
+      with:
+        python-version: 3.8
+    - name: install
+      id: install
+      run : pip install -r tests/lint_requirements.txt
+    - name: check
```
neutrinoceros, line 74:
```suggestion

    - name: check
```

munkm:
I'll retrigger the travis tests.

PR #2867, opened 2020-08-15
remove dead test file

neutrinoceros:
## PR Summary

Another deactivated test that hasn't been run in at least 6yr according to git blame. I think it's safe to remove it.


munkm:
I do think it's safe to remove this test. However, the inline comment says 

> skip for now, in principle we can reimplement this in the demeshening

Since we've merged 4.0 and the demeshening has happened, do we want to reimplement this test (here or elsewhere in the code)? I think others who worked on that should weigh in here.

neutrinoceros:
> I think others who worked on that should weigh in here.

From git blame, it looks like a job for @matthewturk and/or @jzuhone :)

matthewturk (approved):

munkm (approved):

PR #2868, opened 2020-08-15
A path forward for the long-standing periodicity-hack

neutrinoceros:
## PR Summary

There's a long standing hack in yt user codes, that's been encouraged in the cookbook itself for lack of a better approach where some methods require periodicity in the simulation box to function. For instance

```python
ds = yt.load(filename)
ds.periodicity = (True, True, True)
yt.boundary_requiring_method()
```

Here I propose to separate this hack from the normal setup for the underlying periodicity attribute, by introducing a user-facing `Dataset.force_periodicity()` method and protecting the original data in a hidden attribute `Dataset._periodicity` that is accessed in other classes as a property `Dataset.periodicity`.

This provides a way forward in making this hack unnecessary. For instance:
- in 4.0, keep backwards compatibility but deprecate the old hack in favour of `force_periodicity`
- in 4.1, remove the setter (break backwards compatibility) so that
```python
ds.periodicity = (True, True, True)
```
becomes an error.
We can also start using `force_periodicity` internally in methods that can't work without it, which would provide the following benefits:
- make the hack completely unnecessary in user codes (but a warning would still be due)
- give devs an easy way to track down where the hack is used so it can be fixed.

I'll need to add some documentation for this.
A nice followup to this would be to harmonize type diversity in existing _periodicity attributes: most are tuples, but some are lists, or ndarrays.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -293,9 +293,9 @@ class Dataset(abc.ABC):
             raise TypeError(err_msg)
         if len(val) != 3:
             raise ValueError(err_msg)
-        if any(not isinstance(p, bool) for p in val):
+        if any(not isinstance(p, (bool, np.bool_)) for p in val):
             raise TypeError(err_msg)
-        self._periodicity = tuple(val)
+        self._periodicity = tuple(bool(p) for p in val)
 
     def force_periodicity(self, val=True):
         """
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -288,10 +288,14 @@ class Dataset(abc.ABC):
             "In the future, this will become an error. "
             "Use `Dataset.force_periodicity` instead."
         )
-
+        err_msg = f"Expected a 3-element boolean tuple, received `{val}`."
+        if not is_sequence(val):
+            raise TypeError(err_msg)
         if len(val) != 3:
-            raise ValueError("periodicity setter expected a 3-element boolean tuple.")
-        self._periodicity = ensure_tuple(val)
+            raise ValueError(err_msg)
+        if any(not isinstance(p, bool) for p in val):
+            raise TypeError(err_msg)
+        self._periodicity = tuple(val)
 
     def force_periodicity(self, val=True):
         """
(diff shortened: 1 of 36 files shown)


matthewturk:
So one thing I want to put out, because I think it's relevant, is that I *think* a proper fix would involve setting the boundary conditions, rather than specifying periodicity.  So this would mean having a `boundary_condition` property (or something like it) that had, among potentially other things, a fixed boundary setting (or analytic), periodic, isolated, etc.  I know that @zingale has had a lot of frustrations with the current yt setup and he might have some thoughts too.

zingale:
I think the basics we should support are reflect (odd and even), first-order extrapolation, and periodic, and then the frontends would need to map their code BCs to these, in the best way possible.  Reflect will be a little tricky, since for vector fields, it is odd reflection for the normal component and even for the others, usually.  Fancier things could come later, and build from these.

neutrinoceros:
>  I think a proper fix would involve setting the boundary conditions, rather than specifying periodicity.

That's totally right, sorry for not making it clearer in the PR description !
This is meant to be a first step in the right direction, so that it becomes absolutely clear this is nothing else than a workaround. `force_periodicity" isn't mean to stay alive forever, and we can deprecate it once we're on track to the actual fix !
I'm extremely interested in @zingale 's opinion as well, since I had this idea from discussions he was part of :-)

zingale:
I think that this sounds like a good interim step.

neutrinoceros:
The failures on travis are real, I must have messed up somewhere, but I don't get how to reproduce those locally.

neutrinoceros:
got it ! Unless I hit the random failure on travis, this should now be ready for review :)

matthewturk:
This seems OK to me, since @zingale thinks it's a good interim step.

matthewturk (approved):

yt/data_objects/static_output.py
```diff
@@ -274,6 +275,40 @@ def unique_identifier(self):
...
+        if len(val) != 3:
+            raise ValueError(err_msg)
+        if any(not isinstance(p, (bool, np.bool_)) for p in val):
+            raise TypeError(err_msg)
+        self._periodicity = tuple(bool(p) for p in val)
+
+    def force_periodicity(self, val=True):
+        """
+        Override box periodicity to (True, True, True).
+        Use ds.force_periodicty(False) to use the actual box periodicity.
```
cphyc, line 304:
```suggestion
        Use ds.force_periodicity(False) to use the actual box periodicity.
```

cphyc (approved):

PR #2874, opened 2020-08-17
Solve some h5py deprecation warnings

neutrinoceros:
## PR Summary

`h5py` deprecated their `dataset.value` attribute and now recommend using `dataset[()]` instead.
The change was already performed in most of the code, this PR adresses the remaining occurrences.

## PR Checklist

- [x] pass `flake8 yt/`
- [x] add tests (a new GH workflow)

### DIFF
--- a/.github/workflows/rules-checks.yaml
+++ b/.github/workflows/rules-checks.yaml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master
-    - name: check h5py import
+    - name: check-h5py-import
       # check that we don't alias h5py to h5
       # reason: discoverability is important since this module's api
       # is unstable and we want to be able to check for potential future failures
@@ -24,7 +24,7 @@ jobs:
       # check that a mode argument is always present in calls to h5py.File()
       # reason: the default value is different in older versions 'w' VS newer ones 'r'
       run: |
-        egrep -r -n "h5py\.File\([^,]+\)" yt > h5-mode.log
+        grep -E -r -n "h5py\.File\([^,]+\)" yt | cat > h5-mode.log
         if [ -s h5-mode.log ] ; then
           echo "h5py.File() should never be called without an explicit mode argument."
           echo "Here are the faulty lines."
--- a/.github/workflows/rules-checks.yaml
+++ b/.github/workflows/rules-checks.yaml
@@ -1,0 +1,33 @@
+name: Auto review bad practice
+on: [pull_request]
+
+jobs:
+  h5py-bad-practices:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: check h5py import
+      # check that we don't alias h5py to h5
+      # reason: discoverability is important since this module's api
+      # is unstable and we want to be able to check for potential future failures
+      id: h5-import
+      run: |
+        grep -r -n "import _h5py as h5" yt | grep -v "import _h5py as h5py" | cat > h5-imports.log
+        if [ -s h5-imports.log ] ; then
+          echo "Please do not import h5py as h5. Here are the faulty lines."
+          cat h5-imports.log
+          exit 1
+        fi
+
+    - name: check-h5py-filemode
+      id: h5-file-mode
+      # check that a mode argument is always present in calls to h5py.File()
+      # reason: the default value is different in older versions 'w' VS newer ones 'r'
+      run: |
+        egrep -r -n "h5py\.File\([^,]+\)" yt > h5-mode.log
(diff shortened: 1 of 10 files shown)


neutrinoceros:
making this a draft for now. I'll check in the test logs if I've missed any occurrences.

neutrinoceros:
There is still one deprecation warning
```
H5pyDeprecationWarning: The default file mode will change to 'r' (read-only) in h5py 3.0. To suppress this warning, pass the mode you need to h5py.File(), or set the global default h5.get_config().default_file_mode, or set the environment variable H5PY_DEFAULT_READONLY=1. Available modes are: 'r', 'r+', 'w', 'w-'/'x', 'a'. See the docs for details.
```
But I couldn't locate the problematic `h5py.File()` call, so I reapplied the script from #2623 to make it a bit easier to discover with regex (still failed).

munkm:
> But I couldn't locate the problematic `h5py.File()` call,

Is it the handler in the OpenPMD frontend? It doesn't explicitly seem to use the mode anywhere that I see.

munkm:
Also lines 97 and 146 of `yt/geometry/geometry_handler.py` don't use `mode=`.

neutrinoceros:
> Is it the handler in the OpenPMD frontend? It doesn't explicitly seem to use the mode anywhere that I see.

I still can't find it... but because the data_structures.py file imports h5py as `h5`, I wasn't able to auto-match it. Nice catch !

> Also lines 97 and 146 of yt/geometry/geometry_handler.py don't use mode=.

right ! they also weren't matched by my regex because the mode argument is passed as a variable rather than a plain string, well played !

I just fixed those, but I would really like to know where exactly it the remaining outcast !

munkm:
> I just fixed those, but I would really like to know where exactly it the remaining outcast !

Oh! None of those fixed the warning???

neutrinoceros:
I mean I just fixed them _right now_ because I didn't catch them before you pointed them out !
So... maybe that did it actually. Wait and see ! :)

neutrinoceros:
there's still one in the log, but I really can't find it's origin
```
ProjectionValues_DD0011_all_('enzo', 'Density')_1_density ... H5pyDeprecationWarning: The default file mode will change to 'r' (read-only) in h5py 3.0. To suppress this warning, pass the mode you need to h5py.File(), or set the global default h5.get_config().default_file_mode, or set the environment variable H5PY_DEFAULT_READONLY=1. Available modes are: 'r', 'r+', 'w', 'w-'/'x', 'a'. See the docs for details.
```

> Is it the handler in the OpenPMD frontend? It doesn't explicitly seem to use the mode anywhere that I see.

@munkm I've been trying to locate but if you're referring to this line
```python
        self._handle = HDF5FileHandler(filename)
```
that's not it: this calls `h5py.File(filename, mode="r")` underneath.

munkm:
> I've been trying to locate 

You already fixed the instances I was talking about with h5 in OpenPMD . 🙂 

There are a couple of other instances where hdf5 is imported as h5 -- I bet the culprit is one of those. They all use h5.File() without `mode=`

The files I see that do that import are:
* `yt/utilities/grid_data_format/conversion/conversion_athena.py`
* `yt/utilities/grid_data_format/tests/test_writer.py`
* `yt/utilities/minimal_representation.py`

neutrinoceros:
nice catch, I'm on it

neutrinoceros:
done ! This allowed me to find the one call that didn't have a "mode" argument at all (hence wasn't future proof). Thanks a lot for your assistance with this !

munkm:
Wooo!!! ... With our powers combined! 😉

neutrinoceros:
I've added a workflow to catch those on future pull requests !
this catches `import _h5py as h5` as errors (because as we saw, those hurt discoverability), and calls to `h5py.FIle()` without a second argument. I've checked on my fork that the workflow correctly detect the problems I'm fixing here 🎉

neutrinoceros:
Welp. I had to update a file opening mode (previously implicit) from `"w"` to `"r"` in order to pass tests.
specifically, the test involved here is
`yt/utilities/tests/test_minimal_representation.py::test_store`

However I have serious doubts that this is the intended behaviour here since the affected method is named `store`, it really sounds to me like it's supposed to be writing stuff. The problem is that opening with `mode="w"` fails because the storage file is apparently already open (in "r" mode, I guess ?), but I wasn't able to track down completely when it's done. I lost track of it here
https://github.com/yt-project/yt/blob/2438fd1166d5ef0bf0aaefc73b40104bb074898f/yt/data_objects/static_output.py#L1020

and I don't get where/how `self.region` is being defined

matthewturk (approved):
seems good to me, and I think adding more stuff to the rules will be fine over time

PR #2875, opened 2020-08-17
hotfix a refactoring mistake that broke load_sample

neutrinoceros:
## PR Summary

Quick fix for `yt.load_sample` that I broke in my refactor #2832

### DIFF
--- a/yt/frontends/arepo/io.py
+++ b/yt/frontends/arepo/io.py
@@ -15,7 +15,7 @@ class IOHandlerArepoHDF5(IOHandlerGadgetHDF5):
         ptype = self.ds._sph_ptypes[0]
         ind = int(ptype[-1])
         si, ei = data_file.start, data_file.end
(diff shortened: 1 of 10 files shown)


munkm:
It looks like the Untar() function is also imported incorrectly based on the changes from this PR. I personally think this should have been done in steps where the functions were moved first, and then refactored in a separate PR. I do not think changing argument names and the logic within the load functions was a good idea to do at the same time as moving them around -- it made import changes like these easy to miss in PR review.

neutrinoceros:
Yeah that did not happen by design but rather by accident: After I moved the function it broke the only test that uses this function. As I was trying to understand what went wrong I ended up renaming some stuff to make it clearer until the test wasn't broken anymore, I'm surprised that the function ended up buggy still, and I'm sorry about this inconvenience.
Thank you for merging quickly !

PR #2876, opened 2020-08-17
fix np deprecation warnings for non-tuple sequence in multi-dim indexing

neutrinoceros:
## PR Summary

in recent versions of numpy, this pattern is deprecated:

```python
# let arr be a 3D array
arr[[slice(None) for _ in range(3)]]
```
and one gets this warning:
```
FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different
```

I discovered the places this pattern was used in yt by checking the logs on Travis.

Note that this is not a bug _yet_, but I'm still using the label since it will become an error.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/data_objects/grid_patch.py
+++ b/yt/data_objects/grid_patch.py
@@ -414,7 +414,7 @@ class AMRGridPatch(YTSelectionContainer):
         else:
             slices = get_nodal_slices(source.shape, nodal_flag, dim)
             for i, sl in enumerate(slices):
-                dest[offset : offset + count, i] = source[sl][np.squeeze(mask)]
+                dest[offset : offset + count, i] = source[tuple(sl)][np.squeeze(mask)]
         return count
 
     def count(self, selector):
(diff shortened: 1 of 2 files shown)


yt/data_objects/grid_patch.py
```diff
@@ -414,7 +414,7 @@ def select(self, selector, source, dest, offset):
         else:
             slices = get_nodal_slices(source.shape, nodal_flag, dim)
             for i, sl in enumerate(slices):
-                dest[offset : offset + count, i] = source[sl][np.squeeze(mask)]
+                dest[offset : offset + count, i] = source[sl][tuple(np.squeeze(mask))]
```
Xarthisius, line 417: I think it should be `tuple()` around `sl` instead.

neutrinoceros: oh right, good catch !

neutrinoceros: fixed !

PR #2877, opened 2020-08-17
More fixes for load_sample

munkm:
## PR Summary

I noticed the name of the first arg in load sample changed from name to fn, even though this argument is specific to a file name. I've changed it back because name is clearer.

I've also fixed the import of pooch.Untar() since it was transferred with an error. 

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1337,7 +1337,7 @@ def load_sample(fn=None, specific_file=None, pbar=True):
             mylog.warning("tqdm is not installed, progress bar can not be displayed.")
 
     if extension == "h5":
-        processor = pooch.Untar()
+        processor = pooch.pooch.Untar()
     else:
         # we are going to assume most files that exist on the hub are
         # compressed in .tar folders. Some may not.
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1291,7 +1291,7 @@ def load_sample(fn=None, specific_file=None, pbar=True):
 
     Parameters
     ----------
-    fn : str or None
+    name : str or None
         The name of the sample data to load. This is generally the name of the
         folder of the dataset. For IsolatedGalaxy, the name would be
         `IsolatedGalaxy`.  If `None` is supplied, the return value
@@ -1312,7 +1312,7 @@ def load_sample(fn=None, specific_file=None, pbar=True):
 
     fido = PoochHandle()
 
-    if fn is None:
+    if name is None:
         keys = []
         for key in fido._registry:
             for ext in _extensions_to_strip:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
About the "fn" VS "filename" naming convention, I wrote a lengthy argument about it when I proposed to change yt.load here #2695
In a word, I went "fn" _specifically_ because it's less clear that it stands for "filename", since the same argument sometimes represents a directory. I made it short so as to not encourage using the keyword argument syntax in downstream code so we can later change it to a positional-only argument (python 3.8).

> I've also fixed the import of pooch.Untar() since it was transferred with an error.

So sorry the refactor broke this loader so much :/ I think there's a cookbook recipe that relies on it somewhere but I guess it doesn't hit all the code. Any way, thanks for the fix !

neutrinoceros:
So indeed the original name was "name" and note "filename", I guess I was trying to harmonize the api, which is the end-goal of moving everything into a common place.

munkm:
Hmmm ok. 

> In a word, I went "fn" specifically because it's less clear that it stands for "filename", since the same argument sometimes represents a directory. 

I can see this argument but when reading this as a dev I would assume that this variable was referring to a function, not a file or folder. I think we chose "name" as the arg here because no matter what it represents the name of the simulation (either as a folder or a file).

neutrinoceros:
> I can see this argument but when reading this as a dev I would assume that this variable was referring to a function, not a file or folder.

I agree, I also find this name quite unfortunate tbh, but it is widely used in the code base !

> I think we chose "name" as the arg here because no matter what it represents the name of the simulation (either as a folder or a file).

yeah that's a valid choice also. My main concern is that I'd like to keep the naming consistent between `load`, `load_simulation` and `load_sample`.

munkm:
> My main concern is that I'd like to keep the naming consistent between load, load_simulation and load_sample. 

I agree. I think this is a good thing for us to be mindful about. I'm going to open an issue to start a discussion about it! 🙂 It seems like if you and I are both bothered by fn we should see if it is something that should be addressed more widely.

neutrinoceros:
Ok now that we have a dedicated issue for discussing the naming problem, what's the best route for this PR in the meantime ?
I would personally prefer we avoid changing the name back and forth but the bugfix should be merged asap so... I'll agree to merge this anyway if you think it's best to keep the naming reversal, but I'll wait for your go.

munkm:
I've changed `name` back to `fn` so we can merge this bugfix soon rather than waiting on a resolution. I hope we get a resolution on a different arg name based on the issue discussion. Once that happens we can open another pr to update the arg names.

PR #2878, opened 2020-08-17
code deduplication: image filename validation

neutrinoceros:
## PR Summary

Refactoring around `yt.funcs.get_image_suffix`, which gives noisy warnings on Travis such as

```
WARNING:yt:Unsupported image suffix requested (.h5)
WARNING:yt:Unsupported image suffix requested (.h5_Slice_z_density)
WARNING:yt:Unsupported image suffix requested (.h5)
WARNING:yt:Unsupported image suffix requested (.h5_Projection_z_density)
```
while the images are still saved in the end, as png.
Since calls to this function are almost _always_ followed by some way to default to `".png"` as a suffix, I
replaced this function with `yt.visualization._commons.validate_image_name`
Doing so, I ended up removing duplicated code with the function `get_canvas`, which already existed but wasn't used

Since some of the affected parts of the code use a lot of nested ifs, I'm pretty sure I must have broken stuff on the way there, so I'm opening as a draft for now🤞 

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -1,4 +1,3 @@
-import os
 from distutils.version import LooseVersion
 from io import BytesIO
 
@@ -13,7 +12,7 @@ from yt.funcs import (
     mylog,
 )
 
-from ._commons import validate_image_name
+from ._commons import get_canvas, validate_image_name
 
 backend_dict = {
     "GTK": ["backend_gtk", "FigureCanvasGTK", "FigureManagerGTK"],
@@ -35,8 +34,6 @@ backend_dict = {
     "nbAgg": ["backend_nbagg", "FigureCanvasNbAgg", "FigureManagerNbAgg"],
     "agg": ["backend_agg", "FigureCanvasAgg"],
 }
-
-_AGG_FORMATS = (".png", ".jpg", ".jpeg", ".raw", ".rgba", ".tif", ".tiff")
 
 
 class CallbackWrapper:
@@ -126,12 +123,6 @@ class PlotMPL:
 
     def save(self, name, mpl_kwargs=None, canvas=None):
         """Choose backend and save image to disk"""
-        from ._mpl_imports import (
-            FigureCanvasAgg,
-            FigureCanvasPdf,
-            FigureCanvasPS,
-            FigureCanvasSVG,
-        )
 
         if mpl_kwargs is None:
             mpl_kwargs = {}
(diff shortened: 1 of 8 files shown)


neutrinoceros:
Hum. Once again I broke some answer tests and I'm having a hard time navigating the source code to find out what's going wrong (or specifically, which part of my change is causing this). @Xarthisius, can I ask your help with this ?

neutrinoceros:
Nevermind I found where it happened by encapsulating the error.

yt/visualization/base_plot_types.py
```diff
@@ -6,36 +6,35 @@
...
     get_interactivity,
     iterable,
     matplotlib_style_context,
     mylog,
 )
 
-backend_dict = {
+from ._commons import get_canvas, validate_image_name
+
+BACKEND_SPECS = {
```
neutrinoceros, line 17: nothing is currently achieved by this change. I was trying to look if this dictionary could be useful elsewhere, and refactored a bit around it to make sense out of it. Now I don't think this would actually be useful as part of yt/visualization/_commons.py but the code that uses it is now slightly more readable, so I think I'll just keep it as is.

yt/visualization/profile_plotter.py
```diff
@@ -287,39 +269,30 @@ def save(self, name=None, suffix=None, mpl_kwargs=None):
...
         xfn = self.profiles[0].x_field
         if isinstance(xfn, tuple):
             xfn = xfn[1]
-        fns = []
+
+        names = []
         for uid, plot in iters:
             if isinstance(uid, tuple):
                 uid = uid[1]
-            if fullname:
```
neutrinoceros, line 315: I note that I broke that part of the logic tree. I think this is one of the places where we implement a "smart" pattern to allow a full path name (that is, the exact desired name of the output file, with an extension included) be passed, while this is not the general way output file names are constructed.
Meanwhile, it doesn't make much tore use a unique exact name inside a loop as is done here.
All in all, I think the code would be much more maintainable if we renounced inferring an extension from the `name` argument, _at least_ when more than one image file is to be saved.

yt/visualization/profile_plotter.py
```diff
@@ -1297,21 +1270,22 @@ def save(self, name=None, suffix=None, mpl_kwargs=None):
...
-                else:
-                    suffix = "png"
-            fn = f"{prefix}_{middle}.{suffix}"
-            names.append(fn)
-            self.plots[f].save(fn, mpl_kwargs)
+                name = name + (os.sep if name[-1] != os.sep else "")
+                name += str(self.profile.ds)
+
+            new_name = validate_image_name(name, suffix)
+            if new_name == name:
```
neutrinoceros, line 1277: see comment above

yt/visualization/plot_container.py
```diff
@@ -498,37 +493,42 @@ def save(self, name=None, suffix=None, mpl_kwargs=None):
...
-        if suffix is None:
-            suffix = get_image_suffix(name)
-            if suffix != "":
-                for v in self.plots.values():
-                    names.append(v.save(name, mpl_kwargs))
-                return names
+
+        new_name = validate_image_name(name, suffix)
+        if new_name == name:
+            for v in self.plots.values():
```
neutrinoceros, line 499: see below

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

matthewturk (approved):

matthewturk:
@neutrinoceros You want any support on this?

neutrinoceros:
I think it's done ? Just waiting for a second review afaic :-)

neutrinoceros:
note that this would benefit from #3064 regarding type hints

munkm (approved):

PR #2880, opened 2020-08-18
hotfix error message format

neutrinoceros:
## PR Summary

Fix a formatting mistake from #2818

### DIFF
--- a/yt/utilities/exceptions.py
+++ b/yt/utilities/exceptions.py
@@ -21,9 +21,9 @@ class YTUnidentifiedDataType(YTException):
 
     def __str__(self):
         msg = f"Could not determine input format from {self.filename}"
-        if self.args is not None:
-            msg += ", {self.args}"
-        if self.kwargs is not None:
+        if self.args:
+            msg += f", {self.args}"
+        if self.kwargs:
             msg += f", {self.kwargs}"
         msg += "."
         return msg
--- a/yt/utilities/exceptions.py
+++ b/yt/utilities/exceptions.py
@@ -14,16 +14,16 @@ class YTException(Exception):
 
 
 class YTUnidentifiedDataType(YTException):
-    def __init__(self, filename, args=None, kwargs=None):
+    def __init__(self, filename, *args, **kwargs):
         self.filename = filename
         self.args = args
         self.kwargs = kwargs
 
     def __str__(self):
         msg = [f"Could not determine input format from `'{self.filename}'"]
-        if self.args not in (None, ()):
+        if self.args:
             msg.append(", ".join(str(a) for a in self.args))
-        if self.kwargs not in (None, {}):
+        if self.kwargs:
             msg.append(", ".join(f"{k}={v}" for k, v in self.kwargs.items()))
         return ", ".join(msg) + "`."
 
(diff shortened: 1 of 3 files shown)


neutrinoceros:
You're absolutely right. In fact, this has already bitten me in the past...
The thing we're always passing *args and **kwargs when this error is raised, but they may very well be an empty tuple or an empty dict respectively, so you get a message such as 
```
YTUnidentifiedDataType: Could not determine input format from my/filename, (), {}
```
and I do think that the trailing `, (), {}` is a bit misleading and looks like an error.
I'll try do make it better.

neutrinoceros:
So I refactored it completely in a way that the error message should now reproduce the call signature.

For instance
```python
 yt.load(".travis.yml", 42, 128, params={"length units": (12, "m")}, force=True)
```
outputs
```
YTUnidentifiedDataType: Could not determine input format from `'.travis.yml', 42, 128, params={'length units': (12, 'm')}, force=True`.
```

yt/utilities/exceptions.py
```diff
@@ -20,13 +20,12 @@ def __init__(self, filename, args=None, kwargs=None):
...
-        if self.kwargs is not None:
-            msg += f", {self.kwargs}"
-        msg += "."
-        return msg
+        msg = [f"Could not determine input format from `'{self.filename}'"]
+        if self.args not in (None, ()):
+            msg.append(", ".join(str(a) for a in self.args))
+        if self.kwargs not in (None, {}):
+            msg.append(", ".join(f"{k}={v}" for k, v in self.kwargs.items()))
+        return ", ".join(msg) + "`."
```
munkm, line 28: I would prefer the message formatting to be above and then `return msg` as previously. I think it's a bit easier to read.

yt/utilities/exceptions.py
```diff
@@ -20,13 +20,12 @@ def __init__(self, filename, args=None, kwargs=None):
...
     def __str__(self):
-        msg = f"Could not determine input format from {self.filename}"
-        if self.args is not None:
-            msg += ", {self.args}"
-        if self.kwargs is not None:
-            msg += f", {self.kwargs}"
-        msg += "."
-        return msg
+        msg = [f"Could not determine input format from `'{self.filename}'"]
+        if self.args not in (None, ()):
```
munkm, line 24: I'm not convinced this is the place we should be handling empty tuples and dicts if they're passed.

munkm:
I like the refactoring of the error message! That's really nice.

yt/utilities/exceptions.py
neutrinoceros: Is it a problem of any sort if they are ?
I mean this should only ever happen because `yt.load()` passes down its `args` and `kwargs` to the error, even if they are empty... in fact I'm thinking it doesn't make much sense to default to None for both here, they should simply be defined as
```
__init__(filename, *args, **kwargs)
```
to make it consistent. *Then* we won't need to test for None at all because we know for sure that they are a tuple and a dictionnary respectively, in which case I can revert to `if self.args:` !

yt/utilities/exceptions.py
neutrinoceros: adressed !

yt/utilities/exceptions.py
neutrinoceros: refactored _again_. I like it much more now, I hope you will too :)

neutrinoceros:
@munkm I think this is ready to go in

munkm (approved):

PR #2881, opened 2020-08-18
fix: adapt PhasePlot's to future upstream api change in matplotlib

neutrinoceros:
## PR Summary

Future proof yt for matplotlib 3.5.
rationale:
mpl 3.3 deprecated the current default behaviour in `Axis.pcolormesh(x, y, c)`, see details there
https://matplotlib.org/3.3.1/gallery/images_contours_and_fields/pcolormesh_grids.html#sphx-glr-gallery-images-contours-and-fields-pcolormesh-grids-py

This is only used in `yt.PhasePlot`. In short, the easiest fix is to explicitly pass a `shading` value. The current default `"flat"` will break in mpl 3.5, so the easiest fix is to use `"nearest"` instead, which will output very similar but not quite identical results (we'd need to update answers).

Just in case someone would absolutely need to use the `"flat"` shading again, it's easy enough to expose it in yt. This also allows to use the `"gouraud"` shading mode for phase plots, see the following examples.

```python
import yt
ds = yt.load_sample("HiresIsolatedGalaxy", specific_file="DD0044/DD0044")
ad = ds.all_data()
for shading in ("nearest", "gouraud", "flat"):
    p = yt.PhasePlot(ad, "density", "temperature", "cell_mass", weight_field=None, shading=shading)
    p.show()
    p.save(f"/tmp/phase_{shading}")
```
## flat (current default)
phase_flat_2d-Profile_density_temperature_cell_mass

## nearest (proposed new default for yt)
phase_nearest_2d-Profile_density_temperature_cell_mass

## Gouraud (continuous linear interpolation in between grid nodes)
phase_gouraud_2d-Profile_density_temperature_cell_mass

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -86,7 +86,7 @@ answer_tests:
     - yt/frontends/owls/tests/test_outputs.py:test_snapshot_033
     - yt/frontends/owls/tests/test_outputs.py:test_OWLS_particlefilter
 
-  local_pw_031:  # PR 2902
+  local_pw_032:  # PR 2881
     - yt/visualization/tests/test_plotwindow.py:test_attributes
     - yt/visualization/tests/test_plotwindow.py:test_attributes_wt
     - yt/visualization/tests/test_particle_plot.py:test_particle_projection_answers
--- a/yt/visualization/particle_plots.py
+++ b/yt/visualization/particle_plots.py
@@ -369,6 +369,7 @@ class ParticlePhasePlot(PhasePlot):
         deposition="ngp",
         fontsize=18,
         figure_size=8.0,
+        shading="nearest",
     ):
 
         # if no z_fields are passed in, use a constant color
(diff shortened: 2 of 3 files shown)


neutrinoceros:
Some answer tests are failing, which is expected since the shading method was changed.
So the _correct_ fix would be to pass the quad nodes to pcolormesh and still use `shading=flat`. I don't think I can do this atm.

neutrinoceros:
answers are updated in https://github.com/yt-project/answer-store/pull/7

matthewturk (approved):

matthewturk (approved):

cphyc (approved):
Provided the shading arg does in the doc, that looks good.

neutrinoceros:
> Provided the shading arg is documented that looks good to go!

good catch. I added the kwarg to the 3 impacted docstrings !

PR #2882, opened 2020-08-18
hotfix: only trigger style checks on PR (avoid duplicated runs)

neutrinoceros:
## PR Summary

Another hotfix for the GH style checks workflow. I noticed that currently, one gets duplicate notifications when something is wrong since the workflow is triggered both on the main repo side and the fork.


neutrinoceros:
We see that the style checks ran here, which means the change is correct !

PR #2885, opened 2020-08-18
skip tests whose requirements are not met

neutrinoceros:
## PR Summary

So in the current state, tests using a `requires_something` decorator are marked as PASS when the requirements aren't available.
This is confusing and in practice I can rarely tell if my tests are run correctly or just ignored and broken, be that locally or remotely.
This makes them marked as SKIPPED instead. Note that I'm using `nose.SkipTest`, which is supported by pytest, so it works with both frameworks and we can later change it to the pytest equivalent when we drop nose.

I also note there's room for code deduplication here but my understanding is that most of the custom testing framework is going to be dumped when we switch to pytest, so hopefully this won't be an issue for long :)

## PR Checklist
- [x] pass `flake8 yt/`


neutrinoceros:
ping @jcoughlin11 for potential review

neutrinoceros:
@yt-fido test this please

munkm:
I think this is a great idea. I would prefer a more visible skip if a requirement isn't satisfied, so that way we're aware of the behavior.

neutrinoceros:
Hey @Xarthisius, I just solved a conflict with your recent "strict requires" update, I could use your review here :)

cphyc (approved):
Awesome! Thanks for the fix. Out of curiosity, why aren't you importing `SkipTest` and `wraps` in the global scope in all the files?

neutrinoceros:
Thanks for asking. The reason is that this needs to be updated when we migrate out of nose and I wanted to keep the temporary bits close to each other to make the transition easier :)

Xarthisius:
For some reason it skipped a significant number of tests on Jenkins. What's puzzling is that they're single cases in a set of tests for a given dataset, e.g.:

```
ProjectionValues_bw_2d0000_dat_sphere_('max', (0_1, 'unitary'))_magnetic_energy_density_2_None ... ok
ProjectionValues_bw_2d0000_dat_sphere_('max', (0_1, 'unitary'))_magnetic_energy_density_2_density ... SKIP
```

it's literally the same test with `weight_field` equal `None` and `density` respectively.

neutrinoceros:
> The question I have is whether we really want that behavior? It's true that some tests were skipped by process running unittests, but it doesn't mean those tests didn't run.

I honestly don't get why this wouldn't be desirable. Unless I'm misunderstanding something (which is entirely possible), it sounds to me like a process running without the `--with-answer` flag should indeed simply skip answer tests. Am I missing something ?

Xarthisius:
> I honestly don't get why this wouldn't be desirable. Unless I'm misunderstanding something (which is entirely possible), it sounds to me like a process running without the `--with-answer` flag should indeed simply skip answer tests. Am I missing something ?

Take a look here: https://tests.yt-project.org/job/yt_py38_git/1234/testReport/ The issue is that they reported twice: once as skipped, and then as passed (often with a name change).

neutrinoceros:
Oh I see. Isn't that more of a problem with how reports are being merged then ? (I'm not too hopeful that there's anything to be done about it but might as well ask)

Xarthisius:
> Oh I see. Isn't that more of a problem with how reports are being merged then ? (I'm not too hopeful that there's anything to be done about it but might as well ask)

If it were singular tests we could experiment with nose ids, but in this case answer tests yield a bunch of smaller tests, e.g.:
`test_blastwave_cartesian_3D` is a single skipped test, but when it runs it actually yields 78 tests called `<something>_bw_3d0000_<something>`. 

Ultimately I'm fine with this change and behavior as long as everybody understands the nuances of the report.

Xarthisius (approved):

Xarthisius:
I think there might be a technical solution after all: https://gist.github.com/Xarthisius/c383a69fd8199a4824689aa665282abb
This should exclude all the tests that are run via answer_tests in `tests/tests.yaml`

neutrinoceros:
Awesome ! Feel free to push to this branch to try it out.

Xarthisius:
@yt-fido test this please

Xarthisius:
After this run finishes there should be ~23 skipped tests on jenkins, which are actually tests that are supposed to be running but are not...

Xarthisius:
With 8ced4a6 we have 19 skipped:
* 8 - yt.visualization.volume_rendering.tests  (pyembree is not installed and tests are not running)
* 2 - yt.data_objects.tests.test_particle_trajectories (commented out in tests.yaml)
* 2 - yt.data_objects.tests.test_spheres (`@requires_module("MiniballCpp")` not installed)
* 2 - yt.frontends.owls_subfind.tests.test_outputs (commented out in tests.yaml)
* 1 - yt.frontends.sdf.tests.test_outputs.test_scivis (not called anywhere...)
* 1 - yt.frontends.art.tests.test_outputs.test_d9p (not called anywhere...)
* 1 - yt.frontends.exodus_ii.tests.test_outputs.test_displacement_fields (not called anywhere...)
* 1 - yt.frontends.rockstar.tests.test_outputs.test_fields_r1 (not called anywhere...)

neutrinoceros:
Awesome, so this is ready to go in then ? I don't know if your approval still counts since you've contributed more than I did :)

Xarthisius:
> Awesome, so this is ready to go in then ? I don't know if your approval still counts since you've contributed more than I did :)

It's good to go as far as I'm concerned.

munkm:
Since this is touching some of the pytest infrastructure @jcoughlin11 should also review this PR.

jcoughlin11 (approved):

jcoughlin11:
This looks good to me! The pytest behavior is the same (using `pytest.skip`), so switching should be easy. The only thing I'm not sure about is how pytest handles parallel testing, but that's a different matter.

tests/nose_runner.py
```diff
@@ -69,21 +58,24 @@ def __call__(self):
...
     def __str__(self):
         return "WILL DO self.name = %s" % self.name
 
 
 def generate_tasks_input():
     pyver = "py{}{}".format(sys.version_info.major, sys.version_info.minor)
+    if sys.version_info < (3, 0, 0):
+        DROP_TAG = "py3"
+    else:
+        DROP_TAG = "py2"
```
neutrinoceros, line 72: @Xarthisius sorry I didn't pay attention to this earlier, may I ask why this change was necessary ? I mean we don't support python 2 at all anymore so I'm a bit confused.

Xarthisius: It's not necessary. I will drop it

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

Issue #2861, opened 2020-08-14
novel phantom bug in macOS tests


neutrinoceros:
### Bug report

**Bug summary**

this tests is now failing on Travis (MacOS only)
yt/frontends/stream/tests/test_stream_unstructured.py::test_multi_field

spotted here
https://travis-ci.org/github/yt-project/yt/jobs/717825920

The failure is random. I ran this twice locally and it failed only once.
I suspect this is related to https://github.com/yt-project/yt/pull/2842

neutrinoceros:
1) macOS builds have been working fine lately
2) we migrated from travis to github workflow
3) I think we can safely conclude this is fixed and/or not relevant any more.

PR #2539, opened 2020-04-13
Use proper dimensionality in bbox/point selection for meshes. Fixes #2535

Xarthisius:
## PR Summary

Currently for unstructured meshes that consist of 2d points we use selection routines that assume 3d. As a result we check random garbage in memory to e.g. decide whether a region intersects with a mesh or not.

PR #2886, opened 2020-08-19
Fix uninitialized values in bbox selection

Xarthisius:
This provided sane values for the reduced dimension while selecting region of a 2d mesh. Not a full and proper fix, but hopefully will make the annoying error on travis go away. It's similar to #2539 but I hope to merge this before we figure out a proper fix for 2D data.

### DIFF
--- a/yt/geometry/selection_routines.pxd
+++ b/yt/geometry/selection_routines.pxd
@@ -20,6 +20,7 @@ cdef class SelectorObject:
     cdef public np.int32_t max_level
     cdef int overlap_cells
     cdef np.float64_t domain_width[3]
+    cdef np.float64_t domain_center[3]
     cdef bint periodicity[3]
     cdef bint _hash_initialized
     cdef np.int64_t _hash
--- a/yt/geometry/selection_routines.pyx
+++ b/yt/geometry/selection_routines.pyx
@@ -149,6 +149,7 @@ cdef class SelectorObject:
             DRE = _ensure_code(ds.domain_right_edge)
             for i in range(3):
                 self.domain_width[i] = DRE[i] - DLE[i]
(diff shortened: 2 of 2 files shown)


neutrinoceros:
Could you elaborate what's the error you are fixing here, and what test(s) are hitting it ?

Xarthisius:
> Could you elaborate what's the error you are fixing here, and what test(s) are hitting it ?

It's the error we've been occasionally seeing on travis w/ OSX for past few weeks.

neutrinoceros:
Ah ! I reported it here https://github.com/yt-project/yt/issues/2861
And Matt already attempted a fix there https://github.com/yt-project/yt/pull/2842

So the randomness arise from the fact we check for uninitialised values ? Fascinating ! Do you see a reliable way to check this indeed fixes the problem ?

Xarthisius:
> Ah ! I reported it here #2861
> And Matt already attempted a fix there #2842
> 
> So the randomness arises from the fact we check for uninitialised values ? Fascinating ! Do you see a reliable way to check this indeed fixes the problem ?

The randomness was caused by the fact that we used a random garbage in `select_bbox` to decide whether a region overlaps 2d points or not. I don't have a reliable way of verifying that this fixes #2861. I was never able to reproduce it on linux, but sounds like you can!

neutrinoceros:
Yeah I can try on my Mac.
I'm now running this horrible loop on master to see _if_ I can reproduce with a somewhat relevant probability.
```bash
for i in {1..100} ; do pytest yt/frontends/stream/tests/test_stream_unstructured.py::test_multi_field; done
```

neutrinoceros:
Oh sorry I just found out that my follow up never reached GitHub... anyway none of my 100 tries failed, so it looks like I was extremely lucky the first time I tried.

neutrinoceros (approved):

Issue #2393, opened 2019-12-15
Unexpected errors when passing wrong shape to YTSphere


cphyc:
### Bug report

**Bug summary**

When building a YTSphere object with a centre with a shape (1, 3), yt subsequently fails with cryptic messages.

**Code for reproduction**

```python
ds = yt.load('output_00080/info_00080.txt')
@yt.particle_filter(name='single', requires=['particle_identity'], filtered_type='io') 
def _filter(pfilter, data): 
    ptype = pfilter.filtered_type 
    return (data[ptype, 'particle_identity'] == 2) & (data[ptype, 'particle_mass'] > 3e-7)
ds.add_particle_filter('single')

centre = ds.r['single', 'particle_position']  # has shape (1, 3)
ad = ds.all_data()
sp = ds.sphere(centre, (100, 'kpc'))

sp['density']  # Fails with: TypeError: only size-1 arrays can be converted to Python scalars
ad - sp  # Fails with TypeError: Cannot convert YTSphere to yt.geometry.selection_routines.SelectorObject

```

**Actual outcome**

For `sp['density']`
```
Traceback (most recent call last):
  File "/tmp/fail.py", line 13, in <module>
    sp['density']
  File "/home/ccc/Documents/prog/yt/yt/data_objects/data_containers.py", line 256, in __getitem__
    self.get_data(f)
  File "/home/ccc/Documents/prog/yt/yt/data_objects/data_containers.py", line 1482, in get_data
    self.index._identify_base_chunk(self)
  File "/home/ccc/Documents/prog/yt/yt/frontends/ramses/data_structures.py", line 277, in _identify_base_chunk
    domains = [dom for dom in self.domains if
  File "/home/ccc/Documents/prog/yt/yt/frontends/ramses/data_structures.py", line 278, in <listcomp>
    dom.included(dobj.selector)]
  File "/home/ccc/Documents/prog/yt/yt/data_objects/data_containers.py", line 1434, in selector
    self._selector = sclass(self)
  File "yt/geometry/selection_routines.pyx", line 717, in yt.geometry.selection_routines.SphereSelector.__init__
    self.center[i] = _ensure_code(dobj.center[i])
TypeError: only size-1 arrays can be converted to Python scalars
```
For `ad-sp`:
```
Traceback (most recent call last):
  File "/tmp/fail.py", line 13, in <module>
    ad-sp
  File "/home/ccc/Documents/prog/yt/yt/data_objects/data_containers.py", line 1653, in __sub__
[...]

PR #2888, opened 2020-08-25
Squeeze domain center array in data_containers

chrishavlin:
## PR Summary

Squeezes the center array to remove extra dimensions so that the user can supply center arrays of arbitrary shape (e.g., shape of `(1,3)`, `(3,1)`, `(1,3,1)`, etc. are reduced to `(3,)`). Fixes issue #2393 

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -207,6 +207,11 @@ class YTDataContainer:
                 self.center = self.ds.find_min(center[4:])[1]
         else:
             self.center = self.ds.arr(center, "code_length", dtype="float64")
+
+        if self.center.ndim > 1:
+            mylog.warning("Removing singleton dimensions from 'center'.")
+            self.center = np.squeeze(self.center)
+
         self.set_field_parameter("center", self.center)
 
     def get_field_parameter(self, name, default=None):
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -211,6 +211,12 @@ class YTDataContainer:
         if self.center.ndim > 1:
             mylog.debug("Removing singleton dimensions from 'center'.")
             self.center = np.squeeze(self.center)
+            if self.center.ndim > 1:
+                msg = (
+                    "center array must be 1 dimensional, supplied center has "
+                    f"{self.center.ndim} dimensions with shape {self.center.shape}."
+                )
+                raise YTException(msg)
 
         self.set_field_parameter("center", self.center)
 
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -209,7 +209,7 @@ class YTDataContainer:
             self.center = self.ds.arr(center, "code_length", dtype="float64")
 
         if self.center.ndim > 1:
-            mylog.warning("Removing singleton dimensions from 'center'.")
+            mylog.debug("Removing singleton dimensions from 'center'.")
             self.center = np.squeeze(self.center)
 
         self.set_field_parameter("center", self.center)
--- a/yt/data_objects/tests/test_center_squeeze.py
+++ b/yt/data_objects/tests/test_center_squeeze.py
@@ -6,15 +6,15 @@ def test_center_squeeze():
     # to the data container.
 
     # list of fields to populate fake datasets with
-    fldz = ("density", "velocity_x", "velocity_y", "velocity_z")
+    fldz = ("Density",)
 
     # create and test amr, random and particle data
-    check_single_ds(fake_amr_ds(fields=fldz))
-    check_single_ds(fake_random_ds(16, fields=fldz))
-    check_single_ds(fake_particle_ds(npart=100), check_morton=False)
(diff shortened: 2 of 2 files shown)


yt/data_objects/tests/test_center_squeeze.py
```diff
@@ -0,0 +1,57 @@
...
+        new_center = center.reshape(test_shape)
+        sp = ds.sphere(new_center, 0.25)
+        sl = ds.slice(0, 0.25, center=new_center)
+        reg = ds.region(new_center, [-0.25, -0.25, -0.25], [0.25, 0.25, 0.25])
+
+        # compare each to the reference containers
+        for ob, ob0, i_ob0 in [(sp, sp0, i_sp0), (sl, sl0, i_sl0), (reg, reg0, i_reg0)]:
+
+            # check that selection field values match the reference
+            for fld in ds.field_list:
```
neutrinoceros, line 50: Isn't it an overkill to check more than an a single field here ?

yt/data_objects/tests/test_center_squeeze.py
```diff
@@ -0,0 +1,57 @@
...
+    # list of fields to populate fake datasets with
+    fldz = ("density", "velocity_x", "velocity_y", "velocity_z")
+
+    # create and test amr, random and particle data
+    check_single_ds(fake_amr_ds(fields=fldz))
+    check_single_ds(fake_random_ds(16, fields=fldz))
+    check_single_ds(fake_particle_ds(npart=100), check_morton=False)
+
+
+def check_single_ds(ds, check_morton=True):
```
neutrinoceros, line 17: I'd suggest removing the default value here and explicitly set the parameter to True where required.

yt/data_objects/data_containers.py
```diff
@@ -207,6 +207,11 @@ def _set_center(self, center):
                 self.center = self.ds.find_min(center[4:])[1]
         else:
             self.center = self.ds.arr(center, "code_length", dtype="float64")
+
+        if self.center.ndim > 1:
+            mylog.warning("Removing singleton dimensions from 'center'.")
```
neutrinoceros, line 212: Not that crucial a message that it warrants a warning, imo
```suggestion
            mylog.debug("Removing singleton dimensions from 'center'.")
```

yt/data_objects/data_containers.py
```diff
@@ -207,6 +207,11 @@ def _set_center(self, center):
...
+            self.center = np.squeeze(self.center)
+
```
neutrinoceros, line 214: while you're at it, I'd suggest adding an error here in case the `self.center.ndim > 1` still holds. It doesn't make sense to pass an array with shape say, (3,3), but it could happen accidentally, so better catch the error as early as possible :-)

neutrinoceros (approved):
Hi Chris, thanks for this fix; I've left a few minor suggestions and questions but otherwise lgtm !

yt/data_objects/tests/test_center_squeeze.py
chrishavlin: yes, probably :) do you think it's also overkill to test the 3 types of test data types? i.e., it's testing `fake_amr_ds`, `fake_random_ds` and `fake_particle_ds`. I erred on the side of over testing since `data_containers` is used by virtually everything, but maybe they're not all needed?

yt/data_objects/data_containers.py
chrishavlin: yes, good idea! will add it.

yt/data_objects/tests/test_center_squeeze.py
neutrinoceros: I think it's alright, the fact that those 3 data set types are using the same code here is an implementation detail and may not be true in the future, so it's good to test all three imo, but other reviewers might disagree :)

Xarthisius: I don't understand the point of this test. `_set_center` is called immediately during 3d data object initialization. Way before any field is involved. Checking the `center` attribute right after object creation should be sufficient:

```
center = ds.domain_center
for test_shape in [(1, 3), (1, 1, 3)]:
    new_center = center.reshape(test_shape)
    assert_equal(ds.sphere(new_center, 0.25).center, center)
    assert_equal(ds.slice(0, 0.25, center=new_center).center, center)
    assert_equal(ds.region(new_center, [-0.25, -0.25, -0.25], [0.25, 0.25, 0.25]).center, center)
```
 BTW, as it is now, it takes almost 5s to run on Travis, which is a lot.

matthewturk:
Anything left on this?

munkm (approved):

munkm:
@chrishavlin you should be able to see Kacper's comment in the "resolved" popouts on github -- it looks like you resolved that conversation so it gets minimized if you look at it on the web interface.

Xarthisius (approved):

chrishavlin:
ah, thanks @munkm . must have hit that by accident before actually resolving it... thanks!

PR #2908, opened 2020-09-16
[PROPOSITION] Run windows tests on travis

cphyc:
## PR Summary

This PR brings support for running tests on travis instead of appveyor. This has the advantage of decreasing the number of platforms we use, at the cost of overloading a bit more our travis infrastructure.

## Details

The test setup relies on visual studio for compiling (from what I could gather) and miniconda to handle the Python (3.8) distribution.

## Stuff to discuss

- In order to keep the file cleaner, we could also rely on external scripts to setup the testing environment (see e.g. https://github.com/trichter/conda4travis/blob/master/conda4travis.sh).
- It would in principle be possible to migrate to another testing infrastructure (e.g. Azure Pipelines), but that would require an entirely new setup


neutrinoceros:
One more argument in favour of this (if we needed one), is that it apparently reveals at least one platform specific bug that Appveyor never caught (??). Or is my interpretation off ?

cphyc:
The failure is actually a _test_ failure. I don't know why it didn't appear on Appveyor though...

17 Sep 2020 21:43:47 Clément Robert <notifications@github.com>:

neutrinoceros:
Matters of semantics. I for one still count this as a bug even if it doesn't affect the code base itself :)

tests/test_requirements.txt
```diff
@@ -12,7 +12,7 @@ nose==1.3.7
 pandas==0.25.3
 pytest~=5.2
 requests==2.20.0
-scipy==1.3.3
+scipy==1.3.2
```
cphyc, line 15: This is the closest version that has a build for Windows on conda-forge. If this causes issues, this can either be upgraded or downgraded.

neutrinoceros: maybe specify `~=1.3` then ? I don't think we should ping patch versions any way.

neutrinoceros (approved):
LGTM

cphyc:
For potential reviewers, if we decide this PR is a go, we'll have to deactivate AppVeyor (otherwise it'll constantly fail).

.travis.yml
```diff
@@ -23,18 +23,58 @@ env:
...
+
+        # Install dependencies
+        CYTHON=$(grep cython tests/test_prerequirements.txt)
+        NUMPY=$(grep numpy tests/test_prerequirements.txt)
+
+        CARTOPY=$(grep cartopy tests/test_requirements.txt)
+        H5PY=$(grep h5py tests/test_requirements.txt)
+        MATPLOTLIB=$(grep matplotlib tests/test_requirements.txt)
+        SCIPY=$(grep scipy tests/test_requirements.txt)
+        conda install --yes -c conda-forge $CYTHON $NUMPY $CARTOPY $H5PY $MATPLOTLIB $SCIPY
```
munkm, line 68: is there a reason these are in the pre-install and the install steps?

munkm: maybe we want to have a minimal=1 block in the pre install so the install of these packages happen in the same steps on each job.

cphyc: That's a very good point. The reason it is like this is because I didn't want to change much the `install` part, but this comes at the cost of having the `before_install` section a bit cluttered.

I guess the good way out of it would be to have two bash scripts `tests/ci_before_install.sh` and `tests/ci_install.sh` that would take the logic out of the `travis.yml` file. What do you think?

munkm: I like that a lot! I'm definitely in support.

munkm:
> It would in principle be possible to migrate to another testing infrastructure (e.g. Azure Pipelines), but that would require an entirely new setup 

I support switching all of our jobs to Travis for now, and we can decide on a switchover to Azure or something later.

cphyc:
> > It would in principle be possible to migrate to another testing infrastructure (e.g. Azure Pipelines), but that would require an entirely new setup
> 
> I support switching all of our jobs to Travis for now, and we can decide on a switchover to Azure or something later.

That's actually an argument in favor of moving logic out of `.travis.yml` and into independent `ci_*.sh` files!

tests/ci_install.sh
```diff
@@ -0,0 +1,43 @@
...
+if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
+    # Install some dependencies using conda
+    CYTHON=$(grep cython tests/test_prerequirements.txt)
+    NUMPY=$(grep numpy tests/test_prerequirements.txt)
+
+    CARTOPY=$(grep cartopy tests/test_requirements.txt)
+    H5PY=$(grep h5py tests/test_requirements.txt)
+    MATPLOTLIB=$(grep matplotlib tests/test_requirements.txt)
+    SCIPY=$(grep scipy tests/test_requirements.txt)
+    conda install --yes -c conda-forge $CYTHON $NUMPY $CARTOPY $H5PY $MATPLOTLIB $SCIPY
```
munkm, line 15: so won't these packages be installed with both pip and conda on the windows stage, then? the next if/else the windows stage should not be in the MINIMAL==1 stage, right?

tests/test_requirements.txt
munkm: We are pinning patch versions elsewhere in this file. I think it's all right for the test setup to pin like this.

tests/ci_before_install.sh
```diff
@@ -0,0 +1,48 @@
+set -x   # Show which command is being run
+
+case $TRAVIS_OS_NAME in
```
munkm, line 3: Since these files have variables that are unique to Travis should we call them `travis_before_install.sh` and `travis_install.sh`?

tests/ci_install.sh
cphyc: This is no problem for now because the path `MINIMAL == 1 and TRAVIS_OS_NAME == windows` is not triggered, but you're right that this should be refactored somehow. I'll update.

munkm (approved):

munkm:
So even though appveyor is getting removed on master we probably shouldn't disable it until a bit down the road (maybe a week or two)? there are some PRs open that will run with it and we probably don't want to force people to rebase for a .... 3rd time(?) in the past few months. What do you all think? I'm fine merging PRs with a failed appveyor if we know they're recent and don't have the appveyor config file.

cphyc:
Maybe one solution would be to create a dummy Appveyor config file? So that already-opened PRs do not need any update, newly-opened ones have super fast Appveyor tests.

Once we are certain that most (if not all) opened PRs have moved to the full Travis testing suite, we can delete Appveyor?

neutrinoceros:
@cphyc , I love this solution !

cphyc:
@munkm @neutrinoceros I added the dummy Appveyor config (Appveyor now runs in 23s!).

munkm:
I love that solution @cphyc !

appveyor.yml
```diff
@@ -1,48 +1,15 @@
...
-    - "echo suppressStreamLogging = True >> %userprofile%\\.config\\yt\\ytrc"
-    - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-    - "copy tests\\matplotlibrc ."
-
-    # Install the build and runtime dependencies of the project.
-    # Create a conda environment
-    - "conda update -q --yes conda"
-    - "conda create -q --yes -n test python=%PYTHON_VERSION%"
-    - "activate test"
+    - "echo 'Not installing anything.'"
```
munkm, line 10: These are not the droids you're looking for.

cphyc: Damned, I missed a really good opportunity.

neutrinoceros: Good one,
> I'm sorry Dave, I'm afraid I cannot do that

takes all the love, to the point it's almost annoying sometimes :)

PR #2906, opened 2020-09-16
hotfix/flaking_ramses

cphyc:
## PR Checklist

- [x] pass `black --check yt/`
- [x] pass `isort . --check --diff`
- [x] pass `flake8 yt/`
- [x] pass `flynt yt/ --fail-on-change --dry-run -e yt/extern`
- [x] New features are documented, with docstrings and narrative docs
- [x] Adds a test for any bugs fixed. Adds tests for new features.

This should fix the flaking errors that's poping up in other PRs.

PR #2909, opened 2020-09-16
Fix indexing issues in sph pixelization

Xarthisius:
## PR Summary

Recent changes meant to ensure that SPH pixelization routines respect periodicity (#2731) introduced a couple of minor issues:

1. Indices used for periodic passes overlapped with indices used to iterate over buffer array in some places (see pseudo code below):
    ```
    for xi in range(2):
        ...
        for yi in range(2):
            ...
            for xi in range(x0, x1):
                ...
                for yi in range(y0, y1):
    ```
    As result, periodic passes never actually happen if inner loops set `xi` and `yi` to value >0
1. Particle y-position in 2nd x pass was incremented twice due to the use of `+=`:
    ```
    for xi in range(2):
        px += xiterv[xi]
        for yi in range(2):
             py += yiterv[yi]  # <---- This will happen twice if periodic in x
    ```

This PR updates outer loops in SPH pixelization routines to use unique indices, and fixes particle positions by using assignment, rather than in-place addition.

#### Note

There are places where I used a verbose notation `posx[j]` etc, instead of `px` etc. namely, during `xiterv` initialization. It was done on purpose to make my life easier during a rebase of #2683 where I need to pay attention what's parallel and what's not.

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -12,7 +12,7 @@ answer_tests:
     - yt/frontends/amrvac/tests/test_outputs.py:test_riemann_cartesian_175D
     - yt/frontends/amrvac/tests/test_outputs.py:test_rmi_cartesian_dust_2D
     
-  local_arepo_006:
+  local_arepo_007:  # PR 2909
     - yt/frontends/arepo/tests/test_outputs.py:test_arepo_bullet
     - yt/frontends/arepo/tests/test_outputs.py:test_arepo_tng59
     - yt/frontends/arepo/tests/test_outputs.py:test_index_override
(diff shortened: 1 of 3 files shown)


matthewturk:
Yeah, this is a good fix.

matthewturk:
So the flake8 issue that's showing up -- isn't that fixed in master now, with #2906 in?  I am surprised it's suddenly showing up.

neutrinoceros:
@matt the flake8 job is run on the latest commit in the branch, not on a merger with master. It's fixed by rebasing or merging :)

jzuhone (approved):

matthewturk (approved):

Xarthisius:
@yt-fido test this please (just to make sure answers are all right, cause Jenkins cheats sometimes...)

PR #2910, opened 2020-09-17
ci: only test against latest python version on windows to reduce CI bottleneck

neutrinoceros:
## PR Summary

As proposed by @Xarthisius on Slack, let's cut the testing time on windows by half and only test against Python 3.8

note: this is superseded by https://github.com/yt-project/yt/pull/2908 !

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,6 @@ environment:
       PYTHON: "C:\\Miniconda36-x64"
 
   matrix:
-      - PYTHON_VERSION: "3.6"
       - PYTHON_VERSION: "3.8"
 
 platform:


Xarthisius (approved):

matthewturk (approved):

munkm:
So by this do you mean that the testing only on 3.8 is a temporary thing, and if #2908 goes in we'll go back to more widely testing on windows?

neutrinoceros:
Well it depends what we agree to do there. I just meant that I except this PR to be short lived if we drop Appveyor soon after :)

munkm:
Ok, I think we should continue the testing conversation in that PR then!

matthewturk:
I don't think we should hold off because we anticipate a fix -- sometimes
those fixes take a long time and we can benefit now!

neutrinoceros:
> I don't think we should hold off because we anticipate a fix -- sometimes
those fixes take a long time and we can benefit now!

That's the idea indeed.

munkm:
Right. I wasn't saying this shouldn't be merged. I was meaning that the conversation about what coverage we want for windows should be in that PR.

PR #2915, opened 2020-09-17
Fix: matplotlib deprecation warnings

### LINKED ISSUE #2914: Matplotlib deprecation warnings
### Bug report

**Bug summary**

There are currently two matplotlib deprecation warnings in our CI logs
```
yt.visualization.tests.test_plotwindow.test_symlog_colorbar ... MatplotlibDeprecationWarning: default base will change from np.e to 10 in 3.4.  To suppress this warning specify the base keyword argument.
2803MatplotlibDeprecationWarning: default base will change from np.e to 10 in 3.4.  To suppress this warning specify the base keyword argument.
2804MatplotlibDeprecationWarning: default base will change from np.e to 10 in 3.4.  To suppress this warning specify the base keyword argument.
2805ok (6.1096s)
2806yt.visualization.tests.test_plotwindow.test_nan_data ... MatplotlibDeprecationWarning: default base will change from np.e to 10 in 3.4.  To suppress this warning specify the base keyword argument.
```
```
yt.visualization.tests.test_mesh_slices.test_perfect_element_intersection ... ok (0.5227s)
2761Tests functionality of off_axis_projection and write_projection. ... MatplotlibDeprecationWarning: Passing parameters norm and vmin/vmax simultaneously is deprecated since 3.3 and will become an error two minor releases later. Please pass vmin/vmax directly to the norm when creating it.
```
Those need to be addressed to future-proof Yt against matplotlib 3.4 and 3.5

neutrinoceros:
## PR Summary

fix #2914

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -162,7 +162,7 @@ answer_tests:
   local_axialpix_006:
     - yt/geometry/coordinates/tests/test_axial_pixelization.py:test_axial_pixelization
 
-  local_cylindrical_background_003:  # PR 2902
+  local_cylindrical_background_004:  # PR 2915
     - yt/geometry/coordinates/tests/test_cylindrical_coordinates.py:test_noise_plots
 
   #local_particle_trajectory_001:
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -224,12 +224,18 @@ class ImagePlotMPL(PlotMPL):
         elif cbnorm == "symlog":
             if cblinthresh is None:
                 cblinthresh = float((np.nanmax(data) - np.nanmin(data)) / 10.0)
-            norm = matplotlib.colors.SymLogNorm(
-                cblinthresh,
+
+            symlog_kwargs = dict(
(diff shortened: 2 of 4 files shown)


neutrinoceros:
The one failure on Jenkins is expected
https://tests.yt-project.org/job/yt_py38_git/1437/testReport/yt.geometry.coordinates.tests/test_cylindrical_coordinates/GenericImage_AMRGridData_all_create_image_None_None/
The diff is caused by a change of base in SymlogNorm.
Prior to mpl 3.2, the base arg wasn't exposed but defaulted to `np.e`. I think that what we really want is a base 10 symlog  norm (I think the difference is that the linthresh reads differently against different norms). This is also going to be the new default behaviour of matplotlib.colors.SymlogNorm starting from mpl 3.4

Admittedly this choice creates a discrepancy between the results from yt+mpl<3.2 and yt+mpl>=3.2.
Alternatively we could embrace the current default base, but I don't think we should.

neutrinoceros:
Hum, got some failure after I switched api in tests. switching to draft again while it's unstable.

cphyc (approved):

PR #2916, opened 2020-09-17
fix a FutureWarning from numpy

neutrinoceros:
## PR Summary

CI logs still contain one FutureWarning from numpy
for instance
```
yt/geometry/tests/test_grid_container.py::test_grid_arrays_view
  /Users/clm/.pyenv/versions/3.8.5/envs/yt_dev/lib/python3.8/site-packages/numpy/core/_internal.py:38: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
    format = dtype(obj[0], align=align)
```

This fixes it. However, I expect this to break the minimal setup run so I'll open as a draft for now.

### DIFF
--- a/yt/frontends/gadget/testing.py
+++ b/yt/frontends/gadget/testing.py
@@ -29,7 +29,7 @@ def write_block(fp, data, endian, fmt, block_id):
     assert fmt in [1, 2]
     block_id = "%-4s" % block_id
     if fmt == 2:
-        block_id_dtype = np.dtype([("id", "S", 4), ("offset", endian + "i4", 1)])
+        block_id_dtype = np.dtype([("id", "S", 4), ("offset", endian + "i4")])
         block_id_data = np.zeros(1, dtype=block_id_dtype)
         block_id_data["id"] = block_id
         block_id_data["offset"] = data.nbytes + 8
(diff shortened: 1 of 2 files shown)


matthewturk:
Can you post the deprecation warning here?

neutrinoceros:
Here you go, that's what I found locally so far
```
yt/frontends/gadget/tests/test_outputs.py::test_gadget_binary
  /Users/clm/dev/python/yt-project/yt/yt/frontends/gadget/testing.py:56: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
    header_dtype = np.dtype(
```
```
/Users/clm/dev/python/yt-project/yt/yt/frontends/gadget/testing.py:32: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
    block_id_dtype = np.dtype([("id", "S", 4), ("offset", endian + "i4", 1)])
```
To be clear, I inspect the logs on Travis, then reproduce them locally. I think I got them all now.

matthewturk:
Wow, I'm not sure what the distinction between the two would be.  Is it that one would be a non-flat array if you accessed just the dtype?

I also must confess I find the numpy change to be a tiny bit annoying, as it seems more natural to me the way we've always done it, by explicitly stating the number of items.  Then again, I am sure there are aspects I'm not seeing or realizing!

neutrinoceros:
> Wow, I'm not sure what the distinction between the two would be. Is it that one would be a non-flat array if you accessed just the dtype?

I. Have. No. Clue. 😅 (the warning is not exactly super clear to me)
I just want to clear the logs of preventable warnings so we can start paying attention to them when we get new ones :)


> I also must confess I find the numpy change to be a tiny bit annoying, as it seems more natural to me the way we've always done it, by explicitly stating the number of items. Then again, I am sure there are aspects I'm not seeing or realizing!

I absolutely agree with you, but it seems reasonable to expect that they have good reasons to do this. In any case, numpy is too big a machine for us to fight against, I'd rather just ride the wave !

matthewturk (approved):

PR #2917, opened 2020-09-18
Drop support for python 3.6

neutrinoceros:
## PR Summary

With the upcoming Python 3.9 release, and following the guidelines in https://numpy.org/neps/nep-0029-deprecation_policy.html, we can soon drop support for Python versions older than 3.7.0
This will enable easier introduction of type annotations in yt (see https://docs.python.org/3.7/whatsnew/3.7.html#whatsnew37-pep563)

This PR should contain the minimal changes to enforce this support drop, but is not meant to be merged before we are able to test against Python 3.9 (most likely not before end of October 2020).


yt/__init__.py
```diff
@@ -9,10 +9,9 @@
 """
 import sys
 
-if sys.version_info[0] < 3:
-    raise Exception(
-        "Python 2 no longer supported.  Please install Python 3 for use with yt."
-    )
+py_version = sys.version_info
+if not (py_version.major == 3 and py_version.minor >= 7):
```
matthewturk, line 13: Maybe we should make this non-fatal and just say, caveat emptor

neutrinoceros: I was actually thinking the opposite, since the current `raise Exception` is catchable, so I replaced it with `sys.exit` specifically to make it fatal. I figure that when we _do_ switch to 3.7 as minimal, the code base will quickly become unstable against with 3.6 and `import yt` will fail no matter what, so I do think it's preferable to exit early with a controlled error message.

cphyc: Any reason not to use 
```suggestion
if py_version < (3, 7):
```
I _know_ you're going to mention Python 4, but to quote Guido (https://twitter.com/gvanrossum/status/1306270482384076800?s=20)
>  If there ever is a version 4, the transition from 3 to 4 will be more like that from 1 to 2 rather than 2 to 3.

neutrinoceros: Yup I'm aware Python 4.0 is not proposed yet, but I still think it's better to assume the worst case scenario.

cphyc: I'm fine with your solution though ;)

README.md
```diff
@@ -1,5 +1,6 @@
 # The yt Project
 
+![PyPI - Python Version](https://img.shields.io/pypi/pyversions/yt)
```
neutrinoceros, line 3: note: we might want to hold off this line since it doesn't reflect the currently supported Python versions accurately (but I hope it will for the 4.0 release)

cphyc (approved):

yt/__init__.py
```diff
@@ -9,10 +9,9 @@
...
+    raise sys.exit("yt requires Python 3.7 or later.")
```
munkm, line 14: I support switching our testing infra to not explicitly test against python 3.6, but I am not sure we should have a hard enforcement here. What about issuing a warning for python < 3.7 that we don't explicitly support those versions and users should expect odd behavior or something?

neutrinoceros: Ok so I should have stated it earlier: the point is to make our lives easier when we start adding type annotations since there level of support is a bit limited in 3.6
I think it's preferable to have a clean PR that breaks compatibility on purpose, as part of a cycle, so it's clear to everyone we can start supporting 3.9 AND don't have to support 3.6 any longer.

munkm: Right, I understand that. I just don't know why we can't choose to say "use at your own risk. This Python version is not explicitly supported in this version of yt" and then users can try to install and use yt with 3.6 if they want to, rather than issuing this more aggressive enforcement.

neutrinoceros: I don't think it can be done in a way that the message would actually show up before a syntaxError is raised. Otherwise I would support this.

matthewturk: Huh, I didn't realize that we used py 3.7 syntax.  Where do we use that?

My preference would be that until we actually move forward with the type annotation syntax that *isn't* supported, we not *hard* cancel for py3.6.  I say this because we have an unknown set of folks using py36, which was the install default for a long time, and we don't want to have a huge barrier to entry, since upgrading the python version can be confusing.  Plus, we might have folks using it with supercomputer python installs, and I think we need to err on the side of forgiveness there.

One thing we *should* do -- which we are several years behind on doing! -- is take a survey of the community to find out, among other things, what the python version distribution is.

neutrinoceros: > Huh, I didn't realize that we used py 3.7 syntax. Where do we use that?

To be clear: we don't, atm.
I should add that I didn't write this PR with the intention of getting it merged early, it could very well sit here unmerged right until the moment we decide to use Python 3.7+ stuff, such as type annotations with decent support, but I wanted this to be ready when that day comes.
Personally I believe yt 4.0 should still support Python 3.6 but we should be allowed to break it for the following release.

Such a survey would be incredibly useful Matt !

munkm: I agree that 4.0 should still support 3.6 -- once we up the release cadence I won't be as worried about dropping python versions but I do worry that dropping 3.6 for 4.0 could cause problems for our users. I also agree with matt that erring on the side of forgiveness for users and not hard cancelling a specific version. I think we need to find a solution that will be forgiving to users.

neutrinoceros: Agreed. To sum up: this PR will be necessary at some point and contains (or should contain) every change needed _when_ we're start annotating code (or using other py3.7+ stuff).
At _that_ point, python 3.6's interpreter won't be able to run `import yt`, so I don't think it will be possible to transition with a "use at your own risks" warning.
In the mean time, no rush, we can keep python 3.6 as our oldest supported version.

neutrinoceros:
Note that numpy is in the process of explicitly dropping support for Python 3.6 in there latest release (candidate)
https://github.com/numpy/numpy/releases/tag/v1.20.0rc2

neutrinoceros:
I have reclassified this with a "doc" label because the main gain is being able to start adding type hints to the code base without the inherent limitation they have in prior to python 3.7 (no forward declaration).

neutrinoceros:
Note to myself that this should be rebased and include a change in the new static badge for python version

munkm:
Related discussion in yt-project/yt#3070

neutrinoceros:
Quick note that the numpy update broke this, I don't understand why yet.

neutrinoceros:
another note that
setup.py's `install_requires` section is due for an update too, but that's not the reason why the build is failing now

neutrinoceros:
switching to draft to prevent accidental merging

munkm:
Awesome! Thank you!

neutrinoceros:
I just learned about pypistats, which is a valuable source of feedback to know what versions of Python our users are installing yt against https://pypistats.org/packages/yt

Though it's mildly disappointing seeing how noisy the data is in the particular case of yt, and in particular the significant chunks of "null" data points. Anyway this doesn't tell the whole story since it only works for pypi, exlcluding conda installs and maybe even more importantly, installs from source.

neutrinoceros:
@yt-fido tzst this please

neutrinoceros:
@fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros:
The single failure looks random _but_ seems reproducible. However, I have no clue what test it is from the logs I'm seeing on Jenkins. @Xarthisius, would you have any idea ?

neutrinoceros:
Considering:
- Python 3.6 reaches end of life (EOL) in a few month (https://www.python.org/dev/peps/pep-0494/#lifespan)
- We just released a major version of yt that still supports it but is explicitly documented as the last version to do so.

I think we're almost set to merge this. Of course I'd like to cause as little disruption as possible so I propose we agree to merge on a fixed date (end of September, say ?), and warn everyone on official channels well ahead of time, maybe even repeatedly. Any objections ? @matthewturk @cphyc @munkm @Xarthisius @jzuhone @brittonsmith @chummels ?

matthewturk:
This seems to be relevant still, since we are changing our minimal python version.  Is it still the appropriate fix for the issue?

neutrinoceros:
Yup. Been curating this PR for a year now, so I still think it is necessary. Let's the merge the conflicts now.

matthewturk:
@Xarthisius ptal

neutrinoceros:
@yt-fido test this please

PR #2918, opened 2020-09-21
Simplify isort conf

neutrinoceros:
## PR Summary

Moving the isort conf from setup.cfg to pyproject.toml (recommended by isort's doc), and using `profile="black"` to simplify the configuration, guaranteeing we won't get conflicts between isort and black.
While I'm at it I bumped isort's version. I doesn't affect the stability of the code base right now (almost 0 churn except for 2 additional newlines) but there were a number of bugfix since last time so it seems worth the shot.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,16 +39,3 @@ ignore = E203, # Whitespace before ':' (black compatibility)
          B302, # this is a python 3 compatibility warning, not relevant since don't support python 2 anymore
 
 jobs=8
-
-# To be kept consistent with "Import Formatting" section in CONTRIBUTING.rst
-[tool:isort]
-multi_line_output=3
-include_trailing_comma=True
-force_grid_wrap=0
-combine_as_imports=True
-line_length=88
-# isort can't be applied to yt/__init__.py because it creates circular imports
-skip = venv, doc, benchmarks, yt/__init__.py, yt/extern
-known_third_party = IPython, nose, numpy, sympy, matplotlib, unyt, git, yaml, dateutil, requests, coverage, pytest, pyx, glue
-known_first_party = yt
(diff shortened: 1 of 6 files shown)


cphyc (approved):
This looks good. Out of curiosity, why benefit is there in bumping the isort version?

neutrinoceros:
Almost none atm, but it provides more stability in the future if we need to change the config (many bugfixes solved conflicted configs).

Here's a list of changes I could collect that _do_ provide some benefits right now:

* colored diff https://github.com/PyCQA/isort/issues/1405
* Update precommit to default to filtering files that are defined in skip.

setup.cfg
```diff
@@ -39,16 +39,3 @@ ignore = E203, # Whitespace before ':' (black compatibility)
...
 
 jobs=8
-
-# To be kept consistent with "Import Formatting" section in CONTRIBUTING.rst
-[tool:isort]
-multi_line_output=3
-include_trailing_comma=True
-force_grid_wrap=0
-combine_as_imports=True
-line_length=88
```
munkm, line 49: I'm guessing these config settings are what are covered by the new `profile=black` setting, is that right?

munkm (approved):
One question but I think this looks good (and I like the idea of trying to minimize conflict between the two).

setup.cfg
neutrinoceros: That's the whole idea of `profile="black"` yeah. To be completely transparent I didn't find a fully fleshed description of what's covered by it exactly, so I just removed everything that had no impact on `isort yt/`, and assumed it was simply redundant.

munkm: It looks like it are these settings: https://black.readthedocs.io/en/stable/compatible_configs.html#configuration 

which are these lines plus a few more!

neutrinoceros: cool !

PR #2871, opened 2020-08-16
Add a filter to ytLogger to avoid duplicated successive entries

neutrinoceros:
## PR Summary

If there's any appetite for  https://github.com/yt-project/yt/issues/2865, here's the fix
fix https://github.com/yt-project/yt/issues/2865
fix https://github.com/yt-project/yt/issues/2859

## A dummy test
Inserting this at the end of `yt/__init__.py`
```python
for _ in range(30):
    mylog.critical("All work and no play makes Jack a dull boy.")
```

and running
```bash
python -c "import yt"
```

## without the filter
```
yt : [CRITICAL ] 2020-08-16 19:24:52,893 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,893 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,894 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,895 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,895 All work and no play makes Jack a dull boy.
yt : [CRITICAL ] 2020-08-16 19:24:52,895 All work and no play makes Jack a dull boy.
[...]

PR #2927, opened 2020-09-28
[bugfix] Construct log message using args.

brittonsmith:
## PR Summary

The log filtering added in PR #2871 introduced a small issue when the logger is called as `mylog.info(message, arg1, arg2, ...)`. Within the logger, only the `msg` attribute was being used to check for similarity to the last log. This caused the prints that happen in `yt.load` to be skipped since they all use the same formatting string.

## PR Checklist

- [ ] pass `flake8 yt/`

### DIFF
--- a/yt/utilities/logger.py
+++ b/yt/utilities/logger.py
@@ -74,7 +74,8 @@ class DuplicateFilter(logging.Filter):
     # source
     # https://stackoverflow.com/questions/44691558/suppress-multiple-messages-with-same-content-in-python-logging-module-aka-log-co  # noqa
     def filter(self, record):
-        current_log = (record.module, record.levelno, record.msg)
+        msg = record.msg % record.args
+        current_log = (record.module, record.levelno, msg)
         if current_log != getattr(self, "last_log", None):
             self.last_log = current_log
             return True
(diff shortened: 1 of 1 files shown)


cphyc (approved):

yt/utilities/logger.py
```diff
@@ -74,7 +74,8 @@ class DuplicateFilter(logging.Filter):
     # source
     # https://stackoverflow.com/questions/44691558/suppress-multiple-messages-with-same-content-in-python-logging-module-aka-log-co  # noqa
     def filter(self, record):
-        current_log = (record.module, record.levelno, record.msg)
+        msg = record.msg % record.args
+        current_log = (record.module, record.levelno, msg)
```
neutrinoceros, line 78: There's probably no need to perform the formatting operation, how about this ?
```suggestion
        current_log = (record.module, record.levelno, record.msg, record.args)
```

neutrinoceros (approved):
I confirm this fixes the bug. I left a refactor suggestion but I'm happy to merge this as is if you don't like it :)

munkm (approved):

yt/utilities/logger.py
brittonsmith: Sure, I like it.

PR #2928, opened 2020-09-28
More flexible RAMSES loader

cphyc:
## PR Summary

This adds the possibility to load a RAMSES dataset given the name of its folder, of any of its (supported) output file or the info filename (only the latter was supported before).

This comes very handy when you get tired of always writing
```python
import yt
ds = yt.load('output_00123/info_00123.txt')
# now you can do
ds = yt.load('output_00123')
# or even
ds = yt.load('output_00123/amr_00123.out00001')
```

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -33,6 +33,82 @@ from .io_utils import fill_hydro, read_amr
 from .particle_handlers import get_particle_handlers
 
 
+class RAMSESFileSanitizer:
+    """A class to handle the different files that can be passed and associated
+    safely to a RAMSES output."""
+
+    def __init__(self, filename):
+        # Resolve so that it works with symlinks
+        filename = Path(filename).resolve()
+
+        self.original_filename = filename
+
+        self.output_dir = None
+        self.info_fname = None
+
+        for check_fun in (self.test_with_standard_file, self.test_with_folder_name):
+            ok, output_dir, info_fname = check_fun(filename)
+            if ok:
+                break
+
+        # Early exit if the ok flag is False
+        if not ok:
+            return
+
+        # Special case when organized in groups
+        if output_dir.name == "group_00001":
+            self.root_folder = output_dir.parent
+            self.group_folder = output_dir.name
+        else:
+            self.root_folder = output_dir
+            self.group_folder = None
+        self.info_fname = info_fname
+
+    @property
+    def is_valid(self):
+        return (self.root_folder is not None) and (self.info_fname is not None)
+
+    @staticmethod
+    def check_standard_files(folder, iout):
+        """Return true if the folder contains an amr file and the info file."""
+        # Check that the "amr_" and "info_" files exist
+        ok = (folder / f"amr_{iout}.out00001").is_file()
+        ok &= (folder / f"info_{iout}.txt").is_file()
+        return ok
+
+    @classmethod
+    def test_with_folder_name(cls, output_dir):
+        iout_match = OUTPUT_DIR_RE.match(output_dir.name)
+        ok = output_dir.is_dir() and iout_match is not None
+        if ok:
+            iout = iout_match.group(2)
+            ok &= cls.check_standard_files(output_dir, iout)
+            info_fname = output_dir / f"info_{iout}.txt"
+        else:
+            output_dir, info_fname = None, None
+
+        return ok, output_dir, info_fname
+
+    @classmethod
+    def test_with_standard_file(cls, filename):
+        iout_match = OUTPUT_DIR_RE.match(filename.parent.name)
+        ok = (
+            filename.is_file()
+            and STANDARD_FILE_RE.match(filename.name) is not None
+            and iout_match is not None
+        )
+
+        if ok:
+            iout = iout_match.group(2)
+            ok &= cls.check_standard_files(filename.parent, iout)
+            output_dir = filename.parent
+            info_fname = output_dir / f"info_{iout}.txt"
+        else:
+            output_dir, info_fname = None, None
+
+        return ok, output_dir, info_fname
+
+
 class RAMSESDomainFile:
     _last_mask = None
     _last_selector_id = None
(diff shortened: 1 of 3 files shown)


yt/frontends/ramses/data_structures.py
```diff
@@ -857,9 +867,59 @@ def read_namelist(self):
 
             self.parameters["namelist"] = nml
 
+    @staticmethod
+    def _sanitize_filename(path):
+        # Do not forget to resolve so that it works with symlinks
+        filename = Path(path).resolve()
```
neutrinoceros, line 873: I would suggest using the same variable name. `filename` matches the line where it's used so it's the easy choice.

yt/frontends/ramses/data_structures.py
```diff
@@ -857,9 +867,59 @@ def read_namelist(self):
 
             self.parameters["namelist"] = nml
 
+    @staticmethod
+    def _sanitize_filename(path):
```
neutrinoceros, line 871: I think this function is complex enough to warrant a class definition. It would also allow its members (functions) be be unit-tested

yt/frontends/ramses/data_structures.py
```diff
@@ -857,9 +867,59 @@ def read_namelist(self):
...
+                output_dir, info_fname = None, None
+
+            return ok, output_dir, info_fname
+
+        for check_fun in (test_with_standard_file, test_with_folder_name):
+            ok, output_dir, info_fname = check_fun(filename)
+            if ok:
+                return output_dir, info_fname
+        else:
+            return None, None
```
neutrinoceros, line 916:
```suggestion
        return None, None
```

yt/frontends/ramses/data_structures.py
```diff
@@ -857,9 +867,59 @@ def read_namelist(self):
...
+
     @classmethod
-    def _is_valid(self, *args, **kwargs):
-        if not os.path.basename(args[0]).startswith("info_"):
+    def _is_valid(self, filename, *args, **kwargs):
```
neutrinoceros, line 919: Not related to this PR but there's a mistake here, might as well fix it. (also, this is the focus point of #2739)
```suggestion
    def _is_valid(cls, filename, *args, **kwargs):
```

yt/frontends/ramses/data_structures.py
```diff
@@ -857,9 +867,59 @@ def read_namelist(self):
...
+        output_dir, info_fname = self._sanitize_filename(filename)
+
+        if output_dir is None or info_fname is None:
             return False
-        fn = args[0].replace("info_", "amr_").replace(".txt", ".out00001")
-        return os.path.exists(fn)
+        else:
+            return True
```
neutrinoceros, line 925: possible one liner :
```python
return None not in self._sanitize_filename(filename)
```

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -54,7 +54,11 @@ def test_output_00080():
 
 @requires_file(output_00080)
 def test_RAMSESDataset():
-    assert isinstance(data_dir_load(output_00080), RAMSESDataset)
+    # Test that we can load a RAMSES dataset by giving it the folder name,
+    # the info file name or an amr file name
+    base_dir, info_file_fname = os.path.split(output_00080)
+    for fname in (output_00080, base_dir, os.path.join(base_dir, "amr_00080.out00001")):
+        assert isinstance(data_dir_load(fname), RAMSESDataset)
```
neutrinoceros, line 61: I think this test case should be isolated instead.

neutrinoceros (commented):
Overall this looks good. I left some suggestion for further improvement (in particular to make this more testable)

yt/frontends/ramses/tests/test_outputs.py
cphyc: What do you mean it should be isolated?

neutrinoceros: I mean the original test should be left unchanged and the thing you're testing here should be added as a separate test.

cphyc: Will do!

yt/frontends/ramses/data_structures.py
```diff
@@ -18,14 +19,100 @@
...
+        self.output_dir = None
+        self.info_fname = None
+
+        for check_fun in (self.test_with_standard_file, self.test_with_folder_name):
+            ok, output_dir, info_fname = check_fun(filename)
+            if ok:
+                break
+
+        # Early exit if the ok flag is False
+        if not ok:
```
neutrinoceros, line 59:
```suggestion
        else:
```

yt/frontends/ramses/data_structures.py
```diff
@@ -18,14 +19,100 @@
...
+            self.group_name = None
+        self.info_fname = info_fname
+
+    @property
+    def is_valid(self):
+        return (self.root_folder is not None) and (self.info_fname is not None)
+
+    @staticmethod
+    def check_standard_files(folder, iout):
+        """Return true if the folder contains an amr file and the info file."""
```
neutrinoceros, line 77:
```suggestion
        """Return True if the folder contains an amr file and the info file."""
```

yt/frontends/ramses/data_structures.py
```diff
@@ -591,28 +678,33 @@ def __init__(
             max_level, max_level_convention
         )
 
-        # Infer if the output is organized in groups
-        root_folder, group_folder = os.path.split(os.path.split(filename)[0])
+        file_handler = RAMSESFileSanitizer(filename)
+
+        # This should not happen, but let's check nonetheless.
+        if not file_handler.is_valid:
+            raise RuntimeError(
```
neutrinoceros, line 685: RuntimeError should be reserved to when no other builtin error is relevant
```suggestion
            raise ValueError(
```

neutrinoceros (commented):
A few minor suggestions !

yt/frontends/ramses/data_structures.py
matthewturk: The `else` clause on a for loop is kinda confusing for folks, so I think this is `ok` since it's more explicit.

matthewturk (approved):
LGTM

yt/frontends/ramses/data_structures.py
neutrinoceros: It makes me sad that you're right.

yt/frontends/ramses/data_structures.py
```diff
@@ -591,28 +678,33 @@ def __init__(
...
-            # Count the number of groups
-            # note: we exclude the unlikely event that one of the group is actually a
-            # file instad of a folder
+        if file_handler.group_name is not None:
             self.num_groups = len(
                 [
                     _
-                    for _ in glob(os.path.join(root_folder, "group_?????"))
+                    for _ in glob(str(file_handler.root_folder / "group_?????"))
                     if os.path.isdir(_)
```
neutrinoceros, line 698: I'm down for pathlib anytime, just don't mix it with os.path:
```suggestion
                    for _ in file_handler.root_folder.glob("group_?????")
                    if _.is_dir()
```

neutrinoceros, line 698: I'm down for pathlib anytime, just don't mix it with os.path:
```suggestion
                    str(_)
                    for _ in file_handler.root_folder.glob("group_?????")
                    if _.is_dir()
```

neutrinoceros (approved):
left a final comment, otherwise, LGTM

matthewturk:
The "docs" box isn't checked; does this need documentation still?

cphyc:
> The "docs" box isn't checked; does this need documentation still?

I don't think so, unless someone disagrees :)

PR #2930, opened 2020-09-29
[artio] follow YTEP-0003

Provider10:
## PR Summary
Just saw some stuff (that we edited in [PR #2613](https://github.com/yt-project/yt/pull/2613) ) that needed some minor adjustment to be compliant to [YTEP-0003](https://ytep.readthedocs.io/en/latest/YTEPs/YTEP-0003.html) standards. 

One thing that the YTEP-0003 mentions is that backwards compatibility should be maintained to insure that neutral ionic species written as `MM_p0_` are mirrored to `MM_`, but I don't know of a good way to do that besides adding it to the field aliases. Thoughts?

## PR Checklist
(I'm afraid that I don't know how to manage these checklists, so I could use some guidance as to which of these I should look out for)

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/fields/field_aliases.py
+++ b/yt/fields/field_aliases.py
@@ -116,11 +116,13 @@ _field_name_aliases = [
     ("VorticityRPGrowthTimescale", "vorticity_radiation_pressure_growth_timescale"),
     ("DiskAngle", "theta"),
     ("Height", "height"),
-    ("HI density", "H_density"),
+    ("HI density", "H_p0_density"),
     ("HII density", "H_p1_density"),
-    ("HeI density", "He_density"),
+    ("HeI density", "He_p0_density"),
     ("HeII density", "He_p1_density"),
     ("HeIII density", "He_p2_density"),
+    ("H_density", "H_p0_density"),
+    ("He_density", "He_p0_density"),
 ]
 
 _field_units_aliases = [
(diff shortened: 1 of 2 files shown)


yt/fields/field_aliases.py
```diff
@@ -116,11 +116,13 @@
...
     ("Height", "height"),
-    ("HI density", "H_density"),
+    ("HI density", "H_p0_density"),
     ("HII density", "H_p1_density"),
-    ("HeI density", "He_density"),
+    ("HeI density", "He_p0_density"),
     ("HeII density", "He_p1_density"),
     ("HeIII density", "He_p2_density"),
+    ("H_density", "H_p0_density"),
+    ("He_density", "He_p0_density"),
```
brittonsmith, line 125: I think we need to be careful with these two. I think some frontends have aliases `X_number_density` to `X_nuclei_density`, which would be inconsistent with this. The `X_[number_]density` fields have changed meaning a few times. They were once the primary field for neutral ions, then the alias, and now even other things (like nuclei density) in other places. My preference would be to retire them altogether and update YTEP-0003 to reflect that.

matthewturk: @brittonsmith Do you think that should happen here before this is merged?

brittonsmith: Yes, I would prefer that I think.

neutrinoceros:
It's been 3 weeks. I hope it's okay to just close.

PR #2931, opened 2020-10-01
cleanup amrvac frontend __init__.py

neutrinoceros:
## PR Summary

Remove dead import statements from `yt/frontends/amrvac/__init__.py`
The only important difference is that `read_amrvac_parfiles` should not be imported on the user side from `yt.frontends.amrvac` but from `yt.frontends.amrvac.api`. I expect to be the only user potentially affected by this change since since function is not well known.


Xarthisius (approved):

munkm (approved):
This looks good to me!

PR #2934, opened 2020-10-07
adds pickleability to geometric selection objects

chrishavlin:
## PR Summary

This PR adds support for pickling the geometric `selector` objects in `yt/geometry/selection_routines.pyx` (`PointSelector`,`SphereSelector`,`RegionSelector`, `DiskSelector`, `CuttingPlaneSelector`,`SliceSelector`,  `OrthoRaySelector`,`RaySelector`,`EllipsoidSelector`). A subsequent PR will add pickle support for the remaining `selector` types. 

## PR Checklist

- [ ] pass `flake8 yt/`
- [ ] Adds tests for new features.

### DIFF
--- a/yt/geometry/selection_routines.pyx
+++ b/yt/geometry/selection_routines.pyx
@@ -703,7 +703,7 @@ cdef class SelectorObject:
                 ("domain_width[1]", self.domain_width[1]),
                 ("domain_width[2]", self.domain_width[2]))
 
-    def _getstatelist_(self):
+    def _get_state_list(self):
         # return a list of attrs needed for __setstate__: implement for each subclass
         if isinstance(self,SelectorObject):
           return []
@@ -715,7 +715,7 @@ cdef class SelectorObject:
         # get list of the attributes we need to rebuild the state:
         base_atts = ["min_level","max_level","overlap_cells",
                      "periodicity","domain_width","domain_center"]
-        child_atts = self._getstatelist_()
+        child_atts = self._get_state_list()
         
         # assemble the state_tuple (('a1',a1val),('a2',a2val),...)
         state_tuple = () 
@@ -808,7 +808,7 @@ cdef class PointSelector(SelectorObject):
                 ("p[1]", self.p[1]),
                 ("p[2]", self.p[2]))
 
-    def _getstatelist_(self):
+    def _get_state_list(self):
         return ['p']
 
 point_selector = PointSelector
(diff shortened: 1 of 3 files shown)


yt/geometry/selection_routines.pyx
```diff
@@ -703,9 +703,40 @@ cdef class SelectorObject:
                 ("domain_width[1]", self.domain_width[1]),
                 ("domain_width[2]", self.domain_width[2]))
 
+    def _get_state_list(self):
+        # return a list of attrs needed for __setstate__: implement for each subclass
+        if isinstance(self,SelectorObject):
```
matthewturk, line 708: I am not sure I understand this.  In what instances will this not return True?  Subclasses should return True here.

```
In [1]: class A: pass

In [2]: class B(A): pass

In [3]: c = B()

In [4]: isinstance(c, A)
Out[4]: True
```

yt/geometry/selection_routines.pyx
```diff
@@ -703,9 +703,40 @@ cdef class SelectorObject:
...
+          return []
+        else:
+          raise NotImplementedError 
+        
+    def __getstate__(self):
+        # returns a tuple containing (attribute name, attribute value) tuples needed to
+        # rebuild the state:
+        base_atts = ["min_level","max_level","overlap_cells",
```
matthewturk, line 716: Can this be a tuple instead of a list?

matthewturk (commented):
Great stuff.  Just a couple comments inline, but I would also like to see the style fixed up, making sure we have spaces after commas and the like.

yt/geometry/selection_routines.pyx
chrishavlin: yes, it can be

yt/geometry/selection_routines.pyx
chrishavlin: oh right, ya, that's wrong.

matthewturk (approved):

matthewturk:
@chrishavlin can you take a quick look at the total runtime for the tests to see if we notice any performance decrease?  I believe that the way `public` attributes are implemented we shouldn't.

chrishavlin:
Looking at the individual test durations on jenkins, I don't see any appreciable differences. seems to be some differences up to +/- a few seconds in some cases, but it seems random (by eye at least). So this PR isn't introducing any obvious performance decrease.

chrishavlin:
@yt-fido test this please

Xarthisius (approved):

PR #2937, opened 2020-10-08
Update isort

neutrinoceros:
## PR Summary

isort 5.6.0 just came out. Several bugs were fixed, but the one important change to us is that it now sorts .pxd files as well.

One question for Cython experts out there (@matthewturk, @cphyc ?) 
I notice that in some places we have 
```python
import numpy as np
cimport numpy as np
```
Is the first line useless or am I missing something ?

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,6 +1,7 @@
 # transition to isort
 7edfcee093cca277307aabdb180e0ffc69768291
 81418e459f16c48d6b7a75d6ef8035dfe9651b39
+60f670d75a23a6d094879437a8df455a66acbeaf
 
 # transition to black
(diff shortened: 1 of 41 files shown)


matthewturk:
My understanding is that the `cimport numpy as np` and `import numpy as np` set up the cimports -- so anything usable in cython-only code (and type definitions, etc) as well as the python imports.  They're both functional.

I think in many cases, *now* it's recommended to `cimport numpy as cnp` and that's perhaps something we should aim for in future code to reduce ambiguity, but going back and changing it should probably not be a priority.

neutrinoceros:
ok cool !
note: there's already a hotfix 5.6.1 release, I'm switching to draft while I'm applying it.

Xarthisius (approved):

munkm (approved):

PR #2938, opened 2020-10-08
fix: implement AMReXDataset._is_valid

### LINKED ISSUE #2805: load particles from generic AMReX dataset
I am attempting to read in particle data generated by AMReX and note that if I comment out `AMReXDataset._is_valid()` then I can access my particle data. Essentially I just want the `AMReXDataset._parse_parameter_file()` function, which is effectively permanently disabled by `AMReXDataset._is_valid` always returning `False`. Is it possible to do away with this function and just rely on the `BoxlibDataset` version?

https://github.com/yt-project/yt/blob/4acdd33ac2b9fc646388c24aaf74857aa3bf844c/yt/frontends/boxlib/data_structures.py#L1722

neutrinoceros:
## PR Summary

another take at #2806, fix #2805 
The test pass locally, but I expect it to fail in CI since I'm not sure how the `tar.gz` file is untared there.
This change also required a fix affecting the common validation for all boxlib datasets, so I also expect surprises there.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -688,7 +688,7 @@ class BoxlibDataset(Dataset):
         return None
 
     @classmethod
-    def _is_valid(cls, *args, **kwargs):
+    def _is_valid(cls, cparam_filename=None, *args, **kwargs):
         output_dir = args[0]
         header_filename = os.path.join(output_dir, "Header")
         # boxlib datasets are always directories, and
@@ -701,9 +701,7 @@ class BoxlibDataset(Dataset):
             # Further checks are performed on subclasses.
             return True
 
-        try:
-            cparam_filename = kwargs.get("cparam_filename") or args[1]
-        except IndexError:
+        if cparam_filename is None:
             cparam_filename = (
                 signature(cls.__init__).parameters["cparam_filename"].default
             )
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -689,33 +689,17 @@ class BoxlibDataset(Dataset):
 
     @classmethod
     def _is_valid(cls, *args, **kwargs):
-        # fill our args
         output_dir = args[0]
         header_filename = os.path.join(output_dir, "Header")
         # boxlib datasets are always directories, and
         # We *know* it's not boxlib if Header doesn't exist.
-        return os.path.exists(header_filename)
-
-    @classmethod
-    def _lookup_cparam_filepath(cls, output_dir, cparam_filename):
-        lookup_table = [
-            os.path.abspath(os.path.join(p, cparam_filename))
-            for p in (output_dir, os.path.dirname(output_dir))
-        ]
-        found = [os.path.exists(file) for file in lookup_table]
-
-        if not any(found):
-            return None
-
-        return lookup_table[found.index(True)]
-
-    @classmethod
-    def _is_valid_subtype(cls, *args, **kwargs):
-        # this is used by derived classes
-        output_dir = args[0]
-
-        if not BoxlibDataset._is_valid(output_dir):
+        if not os.path.exists(header_filename):
             return False
+
+        if cls is BoxlibDataset:
+            # Stop checks here for the boxlib base class.
+            # Further checks are performed on subclasses.
+            return True
 
         try:
             cparam_filename = kwargs.get("cparam_filename") or args[1]
@@ -730,6 +714,19 @@ class BoxlibDataset(Dataset):
 
         lines = [line.lower() for line in open(cparam_filepath).readlines()]
         return any(cls._subtype_keyword in line for line in lines)
+
+    @classmethod
+    def _lookup_cparam_filepath(cls, output_dir, cparam_filename):
+        lookup_table = [
+            os.path.abspath(os.path.join(p, cparam_filename))
+            for p in (output_dir, os.path.dirname(output_dir))
+        ]
+        found = [os.path.exists(file) for file in lookup_table]
+
+        if not any(found):
+            return None
+
+        return lookup_table[found.index(True)]
 
     def _parse_parameter_file(self):
         """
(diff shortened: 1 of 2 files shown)


neutrinoceros:
To my surprise is looks like it all went smoothly after all. The new test isn't skipped on Jenkins so all looks well !

neutrinoceros:
Hi @RevathiJambunathan and @darylbond, could you please confirm that this change works with your personal datasets ?

As of the current state of the master branch, one needs to import deep into the api as
```python
from yt.frontends.boxlib.api import AMReXDatafile
ds = AMReXDatafile(myfile)
print(type(ds)
```

while with this change, you should be able to obtain the same result with
```python
import yt
ds = yt.load(myfile)
print(type(ds))
```
note however that because of how similar boxlib datafiles are, we rely on the `cparam_filename` argument to determine the exact type, so in the case of AMReX, unless your parameter file is named `job_info`, you'll need to explicitly pass it as
```python
ds = yt.load(myfile, cparam_filename=my_cparam_filename)
print(type(ds))
```
Note that it should also work if passed as a positional argument as
```python
ds = yt.load(myfile, my_cparam_filename)
```
though I would personally advise against this since it is intrinsically harder to maintain and I don't think we can or should guarantee this will still work in the future.

RevathiJambunathan:
@neutrinoceros Thank you for this PR. I just tested your branch and it recognizes the WarpXDataset.
```
import yt
ds = yt.load(myfile)
print(type(ds))
<class 'yt.frontends.boxlib.data_structures.WarpXDataset'>
```

But we (me, @ax3l and a few others in our group) notice something with both the master branch and this branch when using `ds.load(myfile) `
When I print out the print(ds.field_list), there is a new species that was not there is the simulation called `nbody` added to the field_list.!
```
[('all', 'particle_cpu'),
 ('all', 'particle_id'),
 ('all', 'particle_momentum_x'),
 ('all', 'particle_momentum_y'),
 ('all', 'particle_momentum_z'),
 ('all', 'particle_position_x'),
 ('all', 'particle_position_y'),
 ('all', 'particle_position_z'),
 ('all', 'particle_weight'),
 ('boxlib', 'Bx'),
 ('boxlib', 'By'),
 ('boxlib', 'Bz'),
 ('boxlib', 'Ex'),
 ('boxlib', 'Ey'),
 ('boxlib', 'Ez'),
 ('boxlib', 'divE'),
 ('boxlib', 'jx'),
 ('boxlib', 'jy'),
 ('boxlib', 'jz'),
 ('boxlib', 'part_per_cell'),
 ('boxlib', 'rho'),
 ('nbody', 'particle_cpu'),
 ('nbody', 'particle_id'),
 ('nbody', 'particle_momentum_x'),
 ('nbody', 'particle_momentum_y'),
 ('nbody', 'particle_momentum_z'),
 ('nbody', 'particle_position_x'),
 ('nbody', 'particle_position_y'),
 ('nbody', 'particle_position_z'),
 ('nbody', 'particle_weight'),
 ('plasma_e', 'particle_cpu'),
 ('plasma_e', 'particle_id'),
 ('plasma_e', 'particle_momentum_x'),
 ('plasma_e', 'particle_momentum_y'),
 ('plasma_e', 'particle_momentum_z'),
 ('plasma_e', 'particle_position_x'),
 ('plasma_e', 'particle_position_y'),
 ('plasma_e', 'particle_position_z'),
 ('plasma_e', 'particle_weight'),
 ('plasma_p', 'particle_cpu'),
 ('plasma_p', 'particle_id'),
 ('plasma_p', 'particle_momentum_x'),
 ('plasma_p', 'particle_momentum_y'),
 ('plasma_p', 'particle_momentum_z'),
 ('plasma_p', 'particle_position_x'),
 ('plasma_p', 'particle_position_y'),
 ('plasma_p', 'particle_position_z'),
 ('plasma_p', 'particle_weight')]

```

But when I use 
```
import yt
from yt.frontends.boxlib.api import AMReXDataset
ds = AMReXDataset(myfile)
print(type(ds))
ds.field_list
```
I get the AMReXDataset as expected 
<class 'yt.frontends.boxlib.data_structures.AMReXDataset'>
Also we dont see `nbody` species.

Any thoughts as to why the nbody species may show up for WarpXDataset?

neutrinoceros:
@RevathiJambunathan , have you tried passing the cparam_filename kwarg to `yt.load` ? Currently this is the only way we can guarantee that your dataset loads with the intended type, so the fact that it's being incorrectly recognised as `WarpXDataset` in your test isn't sufficient.

> Any thoughts as to why the nbody species may show up for WarpXDataset?

I don't have any idea why this happens, but in any case its a different bug at worst. I wouldn't worry about it since you're not actually loading WarpXDataset data, but this may indeed be a sign that something else is wrong with that datatype. Not related to this PR though, as you mentioned it's consistent with what happens on master :-)

RevathiJambunathan:
@neutrinoceros I am sorry. I should have mentioned. This is a WarpXDataset. So it recognizes the dataset correctly. But adds a species `nbody` that was not present in the simulation data? May be I can create a separate issue for this? Since it is not specific to this PR?

neutrinoceros:
Thanks for clarifying ! Yes I think this should go in a separate issue, thank you.

yt/frontends/boxlib/data_structures.py
```diff
@@ -713,7 +711,14 @@ def _is_valid_subtype(cls, *args, **kwargs):
         if not BoxlibDataset._is_valid(output_dir):
             return False
 
-        cparam_filepath = cls._lookup_cparam_filepath(*args, **kwargs)
+        try:
```
matthewturk, line 714: OK, I have to admit, this seems extra magical, even for loading files.  Can you at least put some comments here justifying this incantation?

matthewturk (approved):
Other than my minor quibble about needing comments I think this is okay, but I am still anxious about using signature etc etc.

yt/frontends/boxlib/data_structures.py
neutrinoceros: yeah I'm not a fan either, now that I'm not struggling with it anymore.
I'll try to clarify this, thanks for pointing this out

neutrinoceros: I came out with a better solution. I hope you like it better now !

cphyc (approved):
Looks good to me!

PR #2939, opened 2020-10-08
Fix GCC 5 Builds

### LINKED ISSUE #2892: Fresh yt installation not importing
### Bug report

**Bug summary**

yt fails to import on a fresh development installation. The error is that `ModuleNotFoundError: No module named 'yt.utilities.lib.misc_utilities'` Similar to [Issue 2685](https://github.com/yt-project/yt/issues/2685)

**Code for reproduction**
Installed using `pip install git+git://github.com/yt-project/yt.git`

```python
import yt
```

**Actual outcome**

```
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-2d2292a375dc> in <module>
----> 1 import yt

~/.conda/envs/ytgit/lib/python3.8/site-packages/yt/__init__.py in <module>
     63 )
     64 
---> 65 from yt.fields.api import (
     66     field_plugins,
     67     DerivedField,

~/.conda/envs/ytgit/lib/python3.8/site-packages/yt/fields/api.py in <module>
      1 # from . import species_fields
----> 2 from . import (
      3     angular_momentum,
      4     astro_fields,
      5     cosmology_fields,

~/.conda/envs/ytgit/lib/python3.8/site-packages/yt/fields/angular_momentum.py in <module>
      1 import numpy as np
      2 
----> 3 from yt.utilities.lib.misc_utilities import (
      4     obtain_position_vector,
      5     obtain_relative_velocity_vector,

ModuleNotFoundError: No module named 'yt.utilities.lib.misc_utilities'
```

**Expected outcome**

**Version Information**

  * Operating System: CentOS Linux 7 (Core)x86_64
  * Python Version: 3.8.5
  * yt version: 4.0.dev0
  * Other Libraries (if applicable): N/A

Installed using `pip install git+git://github.com/yt-project/yt.git`

Thanks!

ax3l:
## PR Summary

Use GCC 5 to trigger compiler errors with non-default  C++14 compilers.

Follow-up to regressions from #2575 and #2610
Fix #2892

```
export CC=$(which gcc-5)
export CXX=$(which g++-5)
python3 -m pip wheel -v .
```

## PR Checklist

- [ ] pass `flake8 yt/`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,6 +70,7 @@ cythonize_aliases = {
     "FIXED_INTERP": "yt/utilities/lib/fixed_interpolator.cpp",
     "ARTIO_SOURCE": glob.glob("yt/frontends/artio/artio_headers/*.c"),
     "CPP14_FLAG": CPP14_FLAG,
+    "CPP03_FLAG": CPP03_FLAG,
 }
 
 lib_exts = [
--- a/setup.py
+++ b/setup.py
@@ -38,13 +38,13 @@ with open("README.md") as file:
     long_description = file.read()
 
 OMP_CONFIG = defaultdict(
-    lambda: ["--fopenmp"], {"unix": ["--fopenmp"], "msvc": ["/openmp"]}
+    lambda: ["-fopenmp"], {"unix": ["-fopenmp"], "msvc": ["/openmp"]}
 )
 CPP14_CONFIG = defaultdict(
-    lambda: ["--std=c++14"], {"unix": ["--std=c++14"], "msvc": ["/std:c++14"]}
+    lambda: ["-std=c++14"], {"unix": ["-std=c++14"], "msvc": ["/std:c++14"]}
 )
 CPP03_CONFIG = defaultdict(
-    lambda: ["--std=c++03"], {"unix": ["--std=c++03"], "msvc": ["/std:c++03"]}
+    lambda: ["-std=c++03"], {"unix": ["-std=c++03"], "msvc": ["/std:c++03"]}
 )
 
 _COMPILER = get_default_compiler()
--- a/setup.py
+++ b/setup.py
@@ -37,9 +37,6 @@ if os.path.exists("MANIFEST"):
 with open("README.md") as file:
     long_description = file.read()
 
-OMP_CONFIG = defaultdict(
-    lambda: ["-fopenmp"], {"unix": ["-fopenmp"], "msvc": ["/openmp"]}
-)
 CPP14_CONFIG = defaultdict(
     lambda: ["-std=c++14"], {"unix": ["-std=c++14"], "msvc": ["/std:c++14"]}
 )
@@ -49,10 +46,7 @@ CPP03_CONFIG = defaultdict(
 
 _COMPILER = get_default_compiler()
 
-if check_for_openmp():
-    omp_args = OMP_CONFIG[_COMPILER]
-else:
-    omp_args = []
+omp_args, _ = check_for_openmp()
 
 if os.name == "nt":
     std_libs = []
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 import glob
 import os
 import sys
+from collections import defaultdict
 from distutils.ccompiler import get_default_compiler
 from distutils.version import LooseVersion
 
@@ -36,8 +37,20 @@ if os.path.exists("MANIFEST"):
 with open("README.md") as file:
     long_description = file.read()
 
+OMP_CONFIG = defaultdict(
+    lambda: ["--fopenmp"], {"unix": ["--fopenmp"], "msvc": ["/openmp"]}
+)
+CPP14_CONFIG = defaultdict(
+    lambda: ["--std=c++14"], {"unix": ["--std=c++14"], "msvc": ["/std:c++14"]}
+)
+CPP03_CONFIG = defaultdict(
+    lambda: ["--std=c++03"], {"unix": ["--std=c++03"], "msvc": ["/std:c++03"]}
+)
+
+_COMPILER = get_default_compiler()
+
 if check_for_openmp():
-    omp_args = ["-fopenmp"]
+    omp_args = OMP_CONFIG[_COMPILER]
 else:
     omp_args = []
 
@@ -46,10 +59,8 @@ if os.name == "nt":
 else:
     std_libs = ["m"]
 
-if get_default_compiler() == "msvc":
-    CPP14_FLAG = ["/std:c++14"]
-else:
-    CPP14_FLAG = ["--std=c++14"]
+CPP14_FLAG = CPP14_CONFIG[_COMPILER]
+CPP03_FLAG = CPP03_CONFIG[_COMPILER]
 
 cythonize_aliases = {
     "LIB_DIR": "yt/utilities/lib/",
@@ -65,6 +76,7 @@ cythonize_aliases = {
     "FIXED_INTERP": "yt/utilities/lib/fixed_interpolator.cpp",
     "ARTIO_SOURCE": glob.glob("yt/frontends/artio/artio_headers/*.c"),
     "CPP14_FLAG": CPP14_FLAG,
+    "CPP03_FLAG": CPP03_FLAG,
 }
 
 lib_exts = [
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,6 @@ cythonize_aliases = {
     "FIXED_INTERP": "yt/utilities/lib/fixed_interpolator.cpp",
     "ARTIO_SOURCE": glob.glob("yt/frontends/artio/artio_headers/*.c"),
     "CPP14_FLAG": CPP14_FLAG,
-    "CPP03_FLAG": CPP03_FLAG,
 }
 
 lib_exts = [
--- a/yt/utilities/lib/cykdtree/kdtree.pyx
+++ b/yt/utilities/lib/cykdtree/kdtree.pyx
@@ -3,7 +3,7 @@
 # distutils: sources = yt/utilities/lib/cykdtree/c_utils.cpp
 # distutils: depends = yt/utilities/lib/cykdtree/c_kdtree.hpp, yt/utilities/lib/cykdtree/c_utils.hpp
 # distutils: language = c++
-# distutils: extra_compile_args = CPP03_FLAG
+# distutils: extra_compile_args = CPP14_FLAG
 import cython
 import numpy as np
 
--- a/yt/utilities/lib/cykdtree/kdtree.pyx
+++ b/yt/utilities/lib/cykdtree/kdtree.pyx
@@ -3,7 +3,7 @@
 # distutils: sources = yt/utilities/lib/cykdtree/c_utils.cpp
 # distutils: depends = yt/utilities/lib/cykdtree/c_kdtree.hpp, yt/utilities/lib/cykdtree/c_utils.hpp
 # distutils: language = c++
-# distutils: extra_compile_args = CPP14_FLAG
+# distutils: extra_compile_args = CPP03_FLAG
 import cython
 import numpy as np
 
--- a/yt/utilities/lib/cykdtree/kdtree.pyx
+++ b/yt/utilities/lib/cykdtree/kdtree.pyx
@@ -3,7 +3,7 @@
 # distutils: sources = yt/utilities/lib/cykdtree/c_utils.cpp
 # distutils: depends = yt/utilities/lib/cykdtree/c_kdtree.hpp, yt/utilities/lib/cykdtree/c_utils.hpp
 # distutils: language = c++
-# distutils: extra_compile_args = -std=c++03
+# distutils: extra_compile_args = CPP03_FLAG
 import cython
 import numpy as np
 
--- a/yt/utilities/lib/cykdtree/utils.pyx
+++ b/yt/utilities/lib/cykdtree/utils.pyx
@@ -2,7 +2,7 @@
 # distutils: sources = yt/utilities/lib/cykdtree/c_utils.cpp
 # distutils: depends = yt/utilities/lib/cykdtree/c_utils.hpp
 # distutils: language = c++
-# distutils: extra_compile_args = CPP03_FLAG
+# distutils: extra_compile_args = CPP14_FLAG
 import numpy as np
 
 cimport cython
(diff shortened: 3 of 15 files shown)


.travis.yml
```diff
@@ -17,11 +17,14 @@ addons:
       - proj-bin
       - libgeos-dev
       - libopenmpi-dev
+      - gcc-5
 
 env:
   global:
     PIP=pip
     HOMEBREW_NO_AUTO_UPDATE=1
+    CC=gcc-5
```
ax3l, line 26: changes in this file are temporary, unless we want to add coverage in CI for older compilers permanently

ax3l:
Hi @matthewturk @cphyc,

I saw your PRs in #2575 and #2610 and am not an expert in the places I touch here. Help & hints welcome :)

ax3l:
@RevathiJambunathan does this PR work for you, with respect to #2892?
```bash
python3 -m pip uninstall yt
pip install git+git://github.com/ax3l/yt.git@fix-gcc5
```

tests/travis_install.sh
```diff
@@ -42,4 +43,4 @@ fi
 # Step 3: install yt
 $PIP install -e .
```
ax3l, line 44: fyi: there is no editable install with PEP517 projects anymore [[1]](https://www.python.org/dev/peps/pep-0517/#get-requires-for-build-sdist) [[2]](https://github.com/pypa/pip/issues/7953#issuecomment-646588548)

cphyc: Oh, I didn't know that! I believe a simple fix would be to use `$PIP install .` instead, right?

cphyc:
It seems the bug triggered on OSX may be fixed by replacing `-fopenmp` by `-fopenmp=libomp` (https://stackoverflow.com/questions/47822547/non-default-compilers-gcc-or-vanilla-clang-on-macos-high-sierra-linking-issue/50313990#50313990). I don't understand why it is failing now and used to work before?

cphyc:
I am sorry, I pushed onto your branch by mistake. I meant to push on my fork instead…

tests/travis_install.sh
ax3l: Just found out as well.

According to my understanding, PEP517 packages should be installed locally like this:
```bash
python3 -m pip wheel .
python3 -m pip install *whl
```

brave new world, I guess.

setup.py
```diff
@@ -36,8 +37,20 @@
 with open("README.md") as file:
     long_description = file.read()
 
+OMP_CONFIG = defaultdict(
+    lambda: ["--fopenmp"], {"unix": ["--fopenmp"], "msvc": ["/openmp"]}
```
ax3l, line 41:
```suggestion
    lambda: ["-fopenmp"], {"unix": ["-fopenmp"], "msvc": ["/openmp"]}
```

setup.py
```diff
@@ -36,8 +37,20 @@
...
+)
+CPP14_CONFIG = defaultdict(
+    lambda: ["--std=c++14"], {"unix": ["--std=c++14"], "msvc": ["/std:c++14"]}
```
ax3l, line 44:
```suggestion
    lambda: ["-std=c++14"], {"unix": ["-std=c++14"], "msvc": ["/std:c++14"]}
```

setup.py
```diff
@@ -36,8 +37,20 @@
...
+)
+CPP03_CONFIG = defaultdict(
+    lambda: ["--std=c++03"], {"unix": ["--std=c++03"], "msvc": ["/std:c++03"]}
```
ax3l, line 47:
```suggestion
    lambda: ["-std=c++03"], {"unix": ["-std=c++03"], "msvc": ["/std:c++03"]}
```

ax3l (commented):
The compiler flags are wrong.

ax3l (commented):
Fix compiler flags for OpenMP and C++ std.

ax3l:
@cphyc can you please add the `bug` label to this PR? the `mergable` CI test fails due to that missing classification.

setup.py
```diff
@@ -36,8 +37,20 @@
 with open("README.md") as file:
     long_description = file.read()
 
+OMP_CONFIG = defaultdict(
+    lambda: ["-fopenmp"], {"unix": ["-fopenmp"], "msvc": ["/openmp"]}
```
ax3l, line 41: unix: this is a bit optimistic but will work for GCC and Clang at least

ax3l:
@cphyc if compile errors occur in the yt cythonization process then the overall build process does not always abort. I am no cython / distutils expert, but ideally this should be addressed independently of this PR, otherwise users get faulty installs because such errors are not aborting the build process and get lost between many build output messages and compiler warnings in the code base.

setup.py
ax3l: furthermore: AppleClang does not support OpenMP by default.
If you want to unconditionally add an OpenMP prerequisite, you could install `brew install libomp` on CI (and need to document this new dependency on macOS).

Practically, one might want to implement a compiler check and keep OpenMP optional.

RevathiJambunathan:
@ax3l Thank you for this PR!
I tried installing yt from your branch and I am able to successfully import it.
```
(yt_env) rjambunathan@zilla:~/Documents/SOFTWARES/yt$ python
Python 3.7.9 (default, Aug 18 2020, 06:24:24) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import yt
>>> yt.__version__
'4.0.dev0'
```

setup.py
```diff
@@ -36,8 +37,20 @@
...
+CPP14_CONFIG = defaultdict(
+    lambda: ["-std=c++14"], {"unix": ["-std=c++14"], "msvc": ["/std:c++14"]}
+)
+CPP03_CONFIG = defaultdict(
+    lambda: ["-std=c++03"], {"unix": ["-std=c++03"], "msvc": ["/std:c++03"]}
+)
+
+_COMPILER = get_default_compiler()
+
 if check_for_openmp():
```
Xarthisius, line 52: `check_for_openmp()` makes some assumptions about openmp flags. As a result your code will only be called only if `-fopenmp` works, otherwise it will default to `[]`

ax3l: yes, the assumption about flags (tested and applied) should be in one place, e.g. `check_for_openmp`. we currently duplicate the flags (and assumptions).

ax3l:
@cphyc I simplified and fixed the compiler logic changes you pushed on this PR.

If you don't mind, I would love to not add features in this PR. If you like to widen the support for compilers and OpenMP to more platforms, can we do so in a follow-up PR? Let's focus on the bugfix here.

setup.py
```diff
@@ -65,6 +70,7 @@
     "FIXED_INTERP": "yt/utilities/lib/fixed_interpolator.cpp",
     "ARTIO_SOURCE": glob.glob("yt/frontends/artio/artio_headers/*.c"),
     "CPP14_FLAG": CPP14_FLAG,
+    "CPP03_FLAG": CPP03_FLAG,
```
ax3l, line 73: If possible, build all binaries with the same C++ standard. Especially between C++03 and C++11 were ABI breaks in the STL and with MSVC the math libs and STL will also cause problems.

matthewturk:
Oh my goodness, thank you for taking such a close look at this.  C++ ABIs
and whatnot are things I had not even considered when I was looking at the
level of C++ support we needed.

(Although, maybe I'm also a bit oversensitive on this because I remember
the endless discussion of C++ ABI breakages back in the early 2000s... ;-)

ax3l:
Did something change in `pykdtree` recently? It seems to also try to unconditionally build (as a dependency of yt) with `-fopenmp` flags, which fails on macOS.

ax3l:
On windows, it looks like we need to potentially remove some of the older work-arounds in `yt/utilities/lib/platform_dep.h`?

setup.py
cphyc: I believe it is fine to replace the `CPP03_flag` with the `CPP14_flag` in the two files that use it. I added commit suggestions where relevant.
```suggestion
```

yt/utilities/lib/cykdtree/kdtree.pyx
```diff
@@ -3,7 +3,7 @@
 # distutils: sources = yt/utilities/lib/cykdtree/c_utils.cpp
 # distutils: depends = yt/utilities/lib/cykdtree/c_kdtree.hpp, yt/utilities/lib/cykdtree/c_utils.hpp
 # distutils: language = c++
-# distutils: extra_compile_args = -std=c++03
+# distutils: extra_compile_args = CPP03_FLAG
```
cphyc, line 6:
```suggestion
# distutils: extra_compile_args = CPP14_FLAG
```

yt/utilities/lib/cykdtree/utils.pyx
```diff
@@ -2,7 +2,7 @@
 # distutils: sources = yt/utilities/lib/cykdtree/c_utils.cpp
 # distutils: depends = yt/utilities/lib/cykdtree/c_utils.hpp
 # distutils: language = c++
-# distutils: extra_compile_args = -std=c++03
+# distutils: extra_compile_args = CPP03_FLAG
```
cphyc, line 5:
```suggestion
# distutils: extra_compile_args = CPP14_FLAG
```

yt/utilities/lib/geometry_utils.pyx
```diff
@@ -1,6 +1,6 @@
 # distutils: libraries = STD_LIBS
-# distutils: extra_compile_args = OMP_ARGS
-# distutils: extra_link_args = OMP_ARGS
+# distutils: extra_compile_args = CPP14_FLAG OMP_ARGS
```
cphyc, line 2: I believe adding the CPP14_FLAG doesn't do anything unless you also specify the language, doesn't it?

```suggestion
# distutile: language = c++
# distutils: extra_compile_args = CPP14_FLAG OMP_ARGS
```

cphyc:
> @cphyc I simplified and fixed the compiler logic changes you pushed on this PR.
> 
> If you don't mind, I would love to not add features in this bugfix PR. If you like to widen the support for compilers and OpenMP to more platforms, can we do so in a follow-up PR? I would suggest to improve the compile tests performed in `check_for_openmp` by just testing more known flags of popular compilers. Let's focus on the bugfix here.

Sure! I initially did not wish to push to your PR, so feel free to take it in any direction you think is the best!

tests/travis_install.sh
neutrinoceros: wow. Sorry for asking the naive question, but by "PEP517 project" you mean any project with a `pyproject.toml` file, right ?

yt/utilities/lib/geometry_utils.pyx
neutrinoceros: careful there, you misspelled `distutils` in your suggestion :)

tests/travis_install.sh
ax3l: Yep... I know...

neutrinoceros: thanks for bringing this up, I was the person who introduced pyproject.toml to the project but wasn't aware of this :/

ax3l:
Hm, I wonder why macOS still tries to build with OpenMP and what is wrong with the `erf` on Windows?

cphyc:
> Hm, I wonder why macOS still tries to build with OpenMP and what is wrong with the `erf` on Windows?

According to https://github.com/storpipfugl/pykdtree/issues/51, this is an issue on pykdtree's side. One solution is to not use openmp when compiling on MacOSX. This can be done by setting `USE_OMP=0` in the environment variables.

I have tried that in https://github.com/cphyc/yt/commit/2ad8d1b066a07d1fc27fd0fe38fcee019cac2803, and it seems to do the trick!

cphyc:
Good news, with the HEAD of https://github.com/cphyc/yt/tree/fix-gcc5 (https://github.com/cphyc/yt/commit/af550d9b0098acbe4f69854d285759d9638e2866), the OSX job now builds! Note that it still doesn't work with Windows.

cphyc:
I've issues my fixes in another PR (#2943), which hopefully should contribute to make the tests pass on this one.

yt/utilities/lib/field_interpolation_tables.pxd
```diff
@@ -10,7 +13,8 @@ cimport cython
 cimport numpy as np
 from yt.utilities.lib.fp_utils cimport imax, fmax, imin, fmin, iclip, fclip, fabs
 from libc.stdlib cimport malloc
-from libc.math cimport isnormal
+cdef extern from "<cmath>" namespace "std":
+    bint isnormal(double x) nogil
```
cphyc, line 17: Note: for some reason, Visual Studio doesn't happy about this. However, if we replace by
```suggestion
cdef extern from "<math.h>" namespace "std":
    bint isnormal(double x) nogil
```
it compiles.

cphyc (approved):
Provided the tests pass, I would be happy to merge this!

cphyc:
@neutrinoceros @Xarthisius @ax3l  I have pushed some more commits to make the Windows build build, do you think this is ready to go? I have added some code myself so I don't feel like pushing the merge button ;)

setup.py
```diff
@@ -36,20 +37,24 @@
...
+CPP14_CONFIG = defaultdict(
+    lambda: ["-std=c++14"], {"unix": ["-std=c++14"], "msvc": ["/std:c++14"]}
+)
+CPP03_CONFIG = defaultdict(
+    lambda: ["-std=c++03"], {"unix": ["-std=c++03"], "msvc": ["/std:c++03"]}
+)
+
+_COMPILER = get_default_compiler()
+
+omp_args, _ = check_for_openmp()
```
Xarthisius, line 49: If we are distinguishing between compile and link flags inside `check_for_openmp()`, shouldn't we do the same thing during cythonization, i.e. use two separate aliases? Otherwise, why return two variables?

Xarthisius (changes requested):

yt/utilities/lib/cykdtree/utils.pyx
```diff
@@ -2,7 +2,7 @@
 # distutils: sources = yt/utilities/lib/cykdtree/c_utils.cpp
 # distutils: depends = yt/utilities/lib/cykdtree/c_utils.hpp
 # distutils: language = c++
-# distutils: extra_compile_args = -std=c++03
+# distutils: extra_compile_args = CPP14_FLAG
```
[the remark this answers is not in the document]
cphyc: It depends, this was changed assuming that one flag was better than two. In this approach, the default C++ language version is C++14. If we want to make minimal changes, then this should be replaced by `CPP03_FLAG`. What do you think is best?

Xarthisius (approved):

ax3l:
Thank you everyone for getting this PR done and merged.

@RevathiJambunathan just a last check, does the yt `master` branch compile again on your system with GCC 5?

munkm:
Thank *you* so much for the fix!! We really appreciate your contribution.

Issue #2720, opened 2020-07-03
SlicePlot of cut region creates fragmented image [AMRVAC]


jorishermans96:
### Bug report

**Bug summary**

Creating a simple slicePlot of a cutted region based on density using either cut_region or include_above yields a fragmented figure (see below). I'm using a 2D Cartesian dataset. The bug also appears when using the other amrvac datasets.

**Code for reproduction**

import yt

overrides = dict(length_unit=(1e9, 'cm'), temperature_unit=(1e6, 'K'), numberdensity_unit=(1e9, 'cm**-3'))
ds = yt.load('test.dat', units_override=overrides, unit_system='cgs')

ad = ds.all_data()
cool_ad = ad.cut_region(["obj['density'] > 2*10**-15"])

p = yt.SlicePlot(ds, 'z', 'density', origin='native', data_source=cool_ad)
p.set_log('density', True)
p.set_unit('density','g/cm**3')
p.set_cmap('density','algae')
p.set_axes_unit('Mm')
p.set_zlim('density', 1e-15, 6*10**-15)
p.save('test', suffix = 'png')

**Actual outcome**

The figure does not show a continuous high density region but it is fragmented.

![test_Slice_z_density](https://user-images.githubusercontent.com/17383484/86446273-0343ec00-bd14-11ea-9e8a-3b6187b84b8c.png)

The console output: 

yt : [WARNING  ] 2020-07-03 09:44:06,237 'staggered' flag was found, but is currently ignored (unsupported)
yt : [WARNING  ] 2020-07-03 09:44:06,238 Overriding code units: Use this option only if you know that the dataset doesn't define the units correctly or at all.
yt : [INFO     ] 2020-07-03 09:44:06,238 Overriding length_unit: 1e+09 cm.
yt : [INFO     ] 2020-07-03 09:44:06,238 Overriding temperature_unit: 1e+06 K.
yt : [INFO     ] 2020-07-03 09:44:06,240 Overriding numberdensity_unit: 1e+09 cm**-3.
yt : [INFO     ] 2020-07-03 09:44:06,437 Parameters: current_time              = 176.4
yt : [INFO     ] 2020-07-03 09:44:06,437 Parameters: domain_dimensions         = [100 100   1]
yt : [INFO     ] 2020-07-03 09:44:06,437 Parameters: domain_left_edge          = [0. 0. 0.]
[...]

PR #2945, opened 2020-10-14
cut region AMRGrid selector fix

chrishavlin:
## PR Summary

This is a tentative fix for issue #2720, but needs some discussion! The origin of the bug in #2720 is a floating point comparison. The `SelectorObject` traverses grid coordinates by incrementing positions in a nested loop while the `AMRGridPatch.select_fcoords` re-calculates the same coordinates by multiplication of indices, resulting in small differences at the level of floating point precision. Since the `CutRegion` selector identifies masked coordinates by comparison of floating point values, these differences lead to points not being selected that should be.

The current fix in this PR feels a bit hacky -- simply changing how the `SelectorObject` nested loop calculates coordinates are calculated seems to have been sufficient, but there may be a better fix (e.g., have the `SelectorObject` return the coordinates, not just the mask indices?). It also will currently fail the `amrvac` answer tests but given the nature of the bug, those answer stores may simply need updating (this code passed all the other answer tests I ran locally, but I did not run them all...).

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -1,6 +1,6 @@
 answer_tests:
 
-  local_amrvac_006:
+  local_amrvac_007: # PR 2945
     - yt/frontends/amrvac/tests/test_outputs.py:test_domain_size
     - yt/frontends/amrvac/tests/test_outputs.py:test_bw_polar_2d
     - yt/frontends/amrvac/tests/test_outputs.py:test_blastwave_cartesian_3D
@@ -20,7 +20,7 @@ answer_tests:
   local_artio_003:
     - yt/frontends/artio/tests/test_outputs.py:test_sizmbhloz
 
-  local_athena_007:
+  local_athena_008: # PR 2945
     - yt/frontends/athena/tests/test_outputs.py:test_cloud
     - yt/frontends/athena/tests/test_outputs.py:test_blast
     - yt/frontends/athena/tests/test_outputs.py:test_stripping
@@ -44,7 +44,7 @@ answer_tests:
     - yt/frontends/enzo/tests/test_outputs.py:test_ecp
     - yt/frontends/enzo/tests/test_outputs.py:test_nuclei_density_fields
 
-  local_enzo_p_008:  # PR 2735
+  local_enzo_p_009: # PR 2945
     - yt/frontends/enzo_p/tests/test_outputs.py:test_hello_world
     - yt/frontends/enzo_p/tests/test_outputs.py:test_particle_fields
 
--- a/yt/geometry/selection_routines.pyx
+++ b/yt/geometry/selection_routines.pyx
@@ -517,18 +517,15 @@ cdef class SelectorObject:
         if level == self.max_level:
             this_level = 1
         with nogil:
-            pos[0] = left_edge[0] + dds[0] * 0.5
             for i in range(dim[0]):
-                pos[1] = left_edge[1] + dds[1] * 0.5
+                pos[0] = left_edge[0] + (i + 0.5) * dds[0]
                 for j in range(dim[1]):
-                    pos[2] = left_edge[2] + dds[2] * 0.5
+                    pos[1] = left_edge[1] + (j + 0.5) * dds[1]
                     for k in range(dim[2]):
+                        pos[2] = left_edge[2] + (k + 0.5) * dds[2]
                         if child_mask[i, j, k] == 1 or this_level == 1:
                             mask[i, j, k] = self.select_cell(pos, dds)
                             total += mask[i, j, k]
-                        pos[2] += dds[2]
-                    pos[1] += dds[1]
-                pos[0] += dds[0]
         return total
 
     @cython.boundscheck(False)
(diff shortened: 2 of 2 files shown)


neutrinoceros:
Indeed it doesn't affect just AMRVAC answers, also Enzo_p and Athena. Some differences go as high as 6%, which seems surprisingly big considering you're essentially fixing floating point comparisons.

chrishavlin:
interesting! 6% is larger than I'd expect too. When I get a chance I'll step through some of the tests with the largest discrepancies to understand the differences. I suspect there are other approaches that would fare better but will require more careful refactoring -- perhaps modifying how the `CutRegion` `_positions` are calculated and stored initially, or writing a new selection routine for calculating the positions that can be used by `AMRGridPatch.select_fcoords` and the internal cython selection methods.

Xarthisius:
6% of what? The number of cells/values affected or field values? Cause the latter shouldn't be really looked upon, right?

Xarthisius (approved):

neutrinoceros:
Oh you're right it shows up in projections, I didn't notice.

Xarthisius:
Just to clarify: the field values in a single pixel can change a lot if you select additional neighboring cell. Same goes for average values which, contrary to the name, is what `FieldValueTest` returns.

In general I don't understand why this PR would be controversial. From the algorithm point of view it's a noop since it only changes the order of operations. From numerical point of view it changes a lot of additions of small number to a big number (which is "a Bad Thing"^{TM}), into a lot of FMAs that should yield much lower roundoff error (which is "a Good Thing"^{TM}). Not to mention it solves an open issue.

neutrinoceros:
> In general I don't understand why this PR would be controversial.

I tend to align here but I don't want to undermine @matthewturk 's comment if he thinks there's room for improvement for whatever reason.
Otherwise I would personally be happy with merging this as is and have a deeper refactor later on.

matthewturk:
I'm OK with moving forward, I just want to be extra sure we know where the
changes are coming from.

chrishavlin:
Update! So I've confirmed that the differences between the failed tests come entirely from points falling on the boundary of the selection objects. The updated method in this PR tends to pick up more points that are on the boundary. You can see that pretty clearly by looking at the image diffs in the failed tests, but I spent some time confirming why this happens. **Short answer**: incrementing within the loop (e.g., `pos[0] += dds[0]`) introduces a bit more floating point error (or at least differences) than multiplication to get the current position (e.g., `pos[0] = left_edge[0] + (i + 0.5) * dds[0]`) resulting in some points moving very slightly with reference to a geometric selection. 

More details below, but I'll bump the test config and push up shortly. 

Details: 

So I rewrote the `fill_mask_selector` and several other related `SelectorObject` methods in `yt/geometry/selection_routines.pyx` for debugging purposes. 

Here's a code snippet of `fill_mask_selector` that increments the position in the method of this PR (the `pos` array) and the original (`pos2`) and flags positions that result in different selections:

```python
pos2[0] = left_edge[0] + dds[0] * 0.5                
for i in range(dim[0]):                
    pos[0] = left_edge[0] + (i + 0.5) * dds[0]                 
    pos2[1] = left_edge[1] + dds[1] * 0.5                 
    for j in range(dim[1]):
        pos[1] = left_edge[1] + (j + 0.5) * dds[1]
        pos2[2] = left_edge[2] + dds[2] * 0.5
        for k in range(dim[2]):
            pos[2] = left_edge[2] + (k+ 0.5) * dds[2]
            if child_mask[i, j, k] == 1 or this_level == 1:
                in_mask2 = self.select_cell(pos2, dds)
                in_mask = self.select_cell(pos, dds) 
                if in_mask != in_mask2:    
                    <<<<<< DO SOME PRINTING AND STUFF >>>>>>>>>>               
            pos2[2] += dds[2]
        pos2[1] += dds[1]
    pos2[0] += dds[0]
```

The `PRINTING AND STUFF` is a bunch of messy prints and calculations that aren't worth including here... but for example, with the athena test sphere using 
```
blast = "MHDBlast/id0/Blast.0100.vtk"
ds = yt.load(blast)
sp = ds.sphere('max',(0.1,"unitary"))
```
one of the `i,j,k` nodes that differs is `i,j,k =14,50,97`, which `pos` identifies as within the sphere but `pos2` does not. The components of the two arrays here are

pos[0] and pos2[0]: 
```python
-0.3549999999999999822 
-0.3549999999999998712 
```
pos[1] and pos2[1]:
```python
0.5050000000000000044    
0.5050000000000002265
```
pos[2] and pos2[2] =:
```python
0.4749999999999999778
0.4750000000000005884
```
 
which ends up giving the squared-distances for `pos` and `pos2` and the squared radius:
```
0.0225000000000000026   <----- pos 
0.0225000000000000338   <----- pos2
0.0225000000000000061   <----- radius ** 2 for this sphere
```

resulting in the `pos` being inside the sphere. All of the ~40 extra points that the `pos` version of the calculation picks up follow this pattern and having even a couple extra points is enough to result in changes in mean values that cause the `FieldValuesTest`   and `ProjectionValuesTest` to fail (note that the extrema quantities are unchanged, only the weighted means of the  `FieldValuesTest` fail). 

So, I'll go ahead and bump those tests...

neutrinoceros (approved):

PR #2946, opened 2020-10-17
An api to programmatically explore the sample data collection

neutrinoceros:
## PR Summary

The idea is to enable easy exploration of the sample dataset collection with pandas, the end goal being an improved discoverability of relevant datasets when writing new tests.
See the added notebook for a demo.

As a side effect, this enables an overhaul of `load_sample`, exposing its gears with improved modularity making it more testable and reliable (for now I'm not replacing the original function).

## PR Checklist

- [ ] find a better home for the demo notebook
- [ ] find a better home/name for the api file ?
- [ ] fix failing tests (add missing entries to registries)
- [ ] produce a unified data frame with columns `["id", "load name", "url", "type", "geometry", "byte size", "dimensionality"]`


matthewturk:
whoa

neutrinoceros:
The main difficulty here is to provide a unified data frame with all relevant data, combining what we currently store in the main repo and what's in the website repo... Mostly, we have missing entries for "load name", which breaks `load_sample`, _and_ there are inconsistencies in labelling between the two repos.
I'll try to fix as much as possible with scripts but I except I'll need to to most of it by hand.

neutrinoceros:
@jcoughlin11 I found some datasets that are not registered in `yt/sample_data_registry.json` yet. Do you remember what's the workflow to compute a hash for them ?

yt/sample_data/api.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    try:
+        s = s.upper()
+    except AttributeError:
+        # input is not a string (likely a np.nan)
+        return pd.NA
+    val = float(re.search(num_exp, s).group())
+    unit = re.search(byte_unit_exp, s).group()
+    exponent = {"B": 0, "KB": 3, "MB": 6, "GB": 9}[unit]
+    return int(val * 10 ** exponent)
```
cphyc, line 39: Should this be instead
```suggestion
    exponent = {"B": 0, "KB": 10, "MB": 20, "GB": 30}[unit]
    return int(val * 2 ** exponent)
```
? That's actually a fair question, I don't know if the sizes in the registry are in metric or base-2...

neutrinoceros: Wait do you mean Kilo Mega and Giga are ever used outside base ten ??

cphyc: Well...  `ls` reports `94848B` as `93KB`, and not `95KB` :D
image

neutrinoceros: wow. I don't think we can safely assume those values were reliably reported so... Maybe I could just normalise the corresponding column in a companion PR, but that would require downloading everything to my drive, and I can't afford this atm.

cphyc: I think that the two are used interchangeably (they should not !), but the difference for 1GiB vs. 1GB is 7% "only". I am fine with you leaving it as is.

neutrinoceros:
note: This should ideally fix #2699

neutrinoceros:
I now think I should cut out the load_sample refactoring since it's already a significant improvement. Exposing a stable and cleanly design "api" (or really just a unified database) is the hard part, especially because some of the data is simply missing, so I should probably separate the tasks.

matthewturk:
I do really like this idea of a dataset explorer.

neutrinoceros:
superseded by #3089 (merged) 🎉

PR #2951, opened 2020-10-26
fix: autocomplete some missing entries in sample data registry file

neutrinoceros:
## PR Summary

This is an attempt at fixing #2699 by auto-discovering missing load names.
I used the following script to auto-update the json file.

```python
import wget
import json
from pathlib import Path
import tarfile
import requests
import tempfile
import yt
from yt.utilities.exceptions import YTUnidentifiedDataType
import struct
from urllib.error import HTTPError

MAX_SIZE = 1e9  # 1 Gb
blacklist = ["output_00080_halos.tar.gz", "acisf05356N003_evt2.fits.gz", "CompactObjects.tar.gz"]

reg_file = "yt/sample_data_registry.json"
with open(reg_file) as fh:
    reg = json.load(fh)

for name, data in reg.items():
    if data["load_name"] or name in blacklist:
        continue
    response = requests.head(data["url"])
    try:
        if int(response.headers["Content-Length"]) > MAX_SIZE:
            # skipping files larger than a certain size
            continue
    except KeyError:
        # unknown file size
        pass
    print(f"Looking for a load name for {name}")
    print(data["url"])
    with tempfile.TemporaryDirectory() as data_dir:
        dest = Path(data_dir) / data["url"].split("/")[-1]
        try:
            wget.download(data["url"], str(dest))
        except HTTPError:
            print("Http error")
            continue
        print()
        if tarfile.is_tarfile(dest):
            with tarfile.open(dest) as fh:
                fh.extractall(data_dir)

        for file in Path(data_dir).glob("**/*"):
            try:
                ds = yt.load(file)
                break
            except (YTUnidentifiedDataType, struct.error, RuntimeError, OSError):
                continue
        else:
            print("failed")
            continue
        data["load_name"] = str(file.relative_to(data_dir))
    with open(reg_file, "wt") as fh:
        json.dump(reg, fh, indent=2)
```

I the case of amrvac sample files, I can confirm this does what we want. However, I'm not 100% confident the results are desirable for all other samples. Note that there remains some missing values. Most of them correspond to samples that I wasn't able to download or load, and they'll need to be addressed later.

### DIFF
--- a/yt/sample_data_registry.json
+++ b/yt/sample_data_registry.json
@@ -74,7 +74,7 @@
   "DMonly.tar.gz": {
     "hash": "45ad285c0df4d7c9e08f497047ac04da373580cd70a8e6fefbe824726235c5c7",
     "load_kwargs": {},
-    "load_name": null,
+    "load_name": "DMonly/PMcrs0.0100.DAT",
     "url": "https://yt-project.org/data/DMonly.tar.gz"
   },
   "DeeplyNestedZoom.tar.gz": {
@@ -116,7 +116,7 @@
   "ExodusII_tests.tar.gz": {
     "hash": "d21c5ef320f2b4705ec4cacf26488a3a59544fab8cc3a9726e8989ad1bd8d177",
     "load_kwargs": {},
-    "load_name": null,
+    "load_name": "ExodusII/gold.e",
     "url": "https://yt-project.org/data/ExodusII_tests.tar.gz"
   },
   "FIRE_M12i_ref11.tar.gz": {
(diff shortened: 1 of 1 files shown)


munkm (approved):
Thank you for fixing this! 

I am not totally sure about other frontends being loaded here so we should have another set of eyes on it before merging. Maybe @matthewturk can take a look too. 

I originally thought that we should create a function that could automatically / intelligently choose which file in a folder to load, but TBH I don't think it is worth our dev time to come up with a solution for that at this time. I think filling in the `null` load names seems like a good solution for now!

matthewturk:
I agree.

Issue #2021, opened 2018-09-13
Functionality using pickle should be removed


brittonsmith:
A number of issues (e.g., Issue #1836, Issue #1898) have come up related to pickling of yt data objects or saving yt data objects using pickle-related functions. Much of this relies on the `parameter_file_storage` functionality that is also no longer functional. However, the `ytdata` frontend and the `save_as_dataset` machinery have made pickle-related stuff mostly redundant. Given this, we should remove the picke-related stuff in favor of `ytdata`-related functionality.

MatthewTurk:
Seems reasonable.

PR #2954, opened 2020-10-29
switch csv writer file IO mode for python 3

chrishavlin:
This PR fixes a bug in the on-disk dataset parameter file storage. 

To reproduce the bug on master, enable in the on disk parameter file storage in the `ytrc` config with `StoreParameterFiles = True` then on importing yt you'll get this error: 

```python
>>> import yt
  File "/home/chavlin/src/yt/yt/__init__.py", line 80, in <module>
    from yt.data_objects.api import (
  File "/home/chavlin/src/yt/yt/data_objects/api.py", line 1, in <module>
    from . import construction_data_containers as __cdc, selection_objects as __sdc
  File "/home/chavlin/src/yt/yt/data_objects/construction_data_containers.py", line 14, in <module>
    from yt.data_objects.selection_objects.data_selection_objects import (
  File "/home/chavlin/src/yt/yt/data_objects/selection_objects/__init__.py", line 1, in <module>
    from .boolean_operations import (
  File "/home/chavlin/src/yt/yt/data_objects/selection_objects/boolean_operations.py", line 8, in <module>
  File "/home/chavlin/src/yt/yt/data_objects/static_output.py", line 63, in <module>
    _ds_store = ParameterFileStore()
  File "/home/chavlin/src/yt/yt/utilities/parameter_file_storage.py", line 62, in __init__
    self._records = self.read_db()
  File "/home/chavlin/src/yt/yt/utilities/parallel_tools/parallel_analysis_interface.py", line 243, in single_proc_results
    return func(self, *args, **kwargs)
  File "/home/chavlin/src/yt/yt/utilities/parameter_file_storage.py", line 193, in read_db
    for v in vals:
  File "/home/chavlin/miniconda3/envs/yt_dev/lib/python3.7/csv.py", line 112, in __next__
    row = next(self.reader)
_csv.Error: iterator should return strings, not bytes (did you open the file in text mode?)
```

The error comes from a difference in the csv writer behavior in python 2 vs 3: python 2 required the file to be opened in binary mode, python 3 switched to plain text. Obviously no one has used this in a while... but this option simplified some parallel operations with dask that I've been experimenting with.

### DIFF
--- a/yt/utilities/parameter_file_storage.py
+++ b/yt/utilities/parameter_file_storage.py
@@ -196,4 +196,5 @@ class ParameterFileStore:
                 v["last_seen"] = 0.0
             else:
                 v["last_seen"] = float(v["last_seen"])
+        f.close()
         return db
--- a/yt/utilities/parameter_file_storage.py
+++ b/yt/utilities/parameter_file_storage.py
@@ -173,7 +173,7 @@ class ParameterFileStore:
         if self._read_only:
             return
         fn = self._get_db_name()
-        f = open(f"{fn}.tmp", "wb")
+        f = open(f"{fn}.tmp", "w")
         w = csv.DictWriter(f, _field_names)
         maxn = ytcfg.getint("yt", "maximumstoreddatasets")  # number written
         for h, v in islice(
@@ -187,7 +187,7 @@ class ParameterFileStore:
     @parallel_simple_proxy
     def read_db(self):
         """ This will read the storage device from disk. """
-        f = open(self._get_db_name(), "rb")
+        f = open(self._get_db_name(), "r")
         vals = csv.DictReader(f, _field_names)
         db = {}
         for v in vals:


neutrinoceros:
I wasn't able to reproduce the problem on the master branch
```python
>>> import yt
>>> yt.config.ytcfg.getboolean("yt", "StoreParameterFiles")
True
```
Any idea what might be the difference ? (I'm using python 3.8.6 here)

Xarthisius:
> Any idea what might be the difference ? (I'm using python 3.8.6 here)

The parameter file has to exist for the error to trigger, so you had to use it in the past. It's gonna be `$HOME/.yt/parameter_files.csv` by default, which btw should be changed to `$HOME/.config/yt` I guess, but that's a different issue.

Another way to trigger is during write. Once you have `StoreParameterFiles` set to True, just try loading any dataset.

yt/utilities/parameter_file_storage.py
```diff
@@ -187,7 +187,7 @@ def _write_out(self):
     @parallel_simple_proxy
     def read_db(self):
         """ This will read the storage device from disk. """
-        f = open(self._get_db_name(), "rb")
+        f = open(self._get_db_name(), "r")
```
Xarthisius, line 190: Since you're modifying this code already, could you make sure that `f` is closed? Preferably with:

```
with open(self._get_db_name(), "r") as f:
    vals = csv.DictReader(f, _field_names)
```

Xarthisius (approved):

chrishavlin:
oh right, @Xarthisius is correct: you do need to already have an existing parameter store on disk to get the error on import. If you don't have that, then the error occurs when `ParameterFileStore._write_out()`, which from what I can see, only happens when pickling a dataset. So you get the same error on master with:

```python
import yt
import pickle 
ds = pickle.loads(pickle.dumps(yt.load("snapshot_033")))
```

Given the move away from using pickle for long term dataset storage (e.g., https://github.com/yt-project/yt/issues/2021), it's not surprising this went unnoticed. I only stumbled across it in exploring some dask functionality with yt since dask uses pickle to serialize objects. 

I'll add the suggested changes shortly...

matthewturk (approved):

munkm (approved):

PR #2956, opened 2020-10-29
Particle trajectories minor improvements

cphyc:
## PR Summary

This addresses some issues on the particle trajectory machinery.

- prevent datasets from being loaded multiple times
- specify non-dynamic runs, to prevent bug if changes are made to the `piter` API

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/data_objects/particle_trajectories.py
+++ b/yt/data_objects/particle_trajectories.py
@@ -241,8 +241,10 @@ class ParticleTrajectories:
         pbar = get_pbar(
             f"Generating [{fields_str}] fields in trajectories", self.num_steps,
         )
-        my_storage = {}
 
+        # Note: we explicitly pass dynamic=False to prevent any change in piter from
+        # breaking the assumption that the same processors load the same datasets
+        my_storage = {}
         for i, (sto, ds) in enumerate(
             self.data_series.piter(storage=my_storage, dynamic=False)
         ):
--- a/yt/data_objects/particle_trajectories.py
+++ b/yt/data_objects/particle_trajectories.py
@@ -65,7 +65,7 @@ class ParticleTrajectories:
         self.num_steps = len(outputs)
         self.times = []
         self.suppress_logging = suppress_logging
-        self.ptype = ptype
+        self.ptype = ptype if ptype else "all"
 
         if fields is None:
             fields = []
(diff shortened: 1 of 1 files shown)


yt/data_objects/particle_trajectories.py
```diff
@@ -142,9 +148,11 @@ def has_key(self, key):
     def keys(self):
         return self.field_data.keys()
 
-    def _get_full_field_name(self, field):
-        ds_first = self.data_series[0]
-        dd_first = ds_first.all_data()
+    def _get_full_field_name(self, field, ds_first=None, dd_first=None):
```
neutrinoceros, line 151: Since this function is called exactly ONCE in the whole code base _and_ is private, I don't think it's worth giving default values to arguments here (even if they make perfect sense)
What's more is that we only ever use it to retrieve the `ftype` part of the full name, so might as well align the function's return value (and name) with its actual usage.

yt/data_objects/particle_trajectories.py
```diff
@@ -111,8 +115,10 @@ def __init__(
             mylog.setLevel(old_level)
 
         sorted_storage = sorted(my_storage.items())
-        times = [time for _fn, (time, *_) in sorted_storage]
-        self.times = self.data_series[0].arr(times, times[0].units)
+        times = [time.to("Myr") for _fn, (time, *_) in sorted_storage]
+        self.times = self.data_series[0].arr(
+            [time.value for time in times], times[0].units
+        )
```
neutrinoceros, line 121: I'm guessing you force-convert times to `"Myr"` as a way to fix a bug where `self.times` would be incorrectly constructed from a heterogenous array of times. I would however favour harmonising the unit to `times[0].units` instead of hardcoding `Myr` which looks like a severe "yt-ism" to me and wouldn't make much sense outside of astronomy.

yt/data_objects/particle_trajectories.py
```diff
@@ -237,13 +245,15 @@ def _get_data(self, fields):
...
-            f"Generating [{', '.join(missing_fields)}] fields in trajectories",
-            self.num_steps,
+            f"Generating [{fields_str}] fields in trajectories", self.num_steps,
         )
         my_storage = {}
 
-        for i, (sto, ds) in enumerate(self.data_series.piter(storage=my_storage)):
+        for i, (sto, ds) in enumerate(
+            self.data_series.piter(storage=my_storage, dynamic=False)
+        ):
```
neutrinoceros, line 256: I think the comment you wrote above about the use of `dynamic=False` might be useful here too.
One way to avoid duplication would be to add this comment to the method's docstring.

neutrinoceros (approved):
LGTM. I'm down with merging this when my comments are addressed.

yt/data_objects/particle_trajectories.py
cphyc: Should be good now!

PR #2958, opened 2020-11-02
Enable ghost zones > 1 for octrees

cphyc:
## PR Summary

This enables support for more than 1 ghost zones with octree datasets.

Also add a minor speed improvement of the mesh-sampling implementation (always start with the largest domain).

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -370,7 +370,9 @@ class RAMSESDomainSubset(OctreeSubset):
         try:
             new_subset = _subset_with_gz[ngz]
             mylog.debug(
-                "Reusing previous subset with ghost zone for domain %s", self.domain_id
+                "Reusing previous subset with %s ghost zones for domain %s",
+                ngz,
+                self.domain_id,
             )
         except KeyError:
             new_subset = RAMSESDomainSubset(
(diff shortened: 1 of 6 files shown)


cphyc:
Note that using more than one ghost zone is very slow, but it works nonetheless.

yt/frontends/ramses/data_structures.py
```diff
@@ -362,23 +365,28 @@ def retrieve_ghost_zones(self, ngz, fields, smoothed=False):
...
+            new_subset = _subset_with_gz[ngz]
             mylog.debug(
-                "Reusing previous subset with ghost zone for domain %s", self.domain_id
+                "Reusing previous subset with %s ghost zones for domain %s",
+                ngz,
+                self.domain_id,
             )
-        except AttributeError:
+        except KeyError:
             new_subset = RAMSESDomainSubset(
```
matthewturk, line 378: Would this result in, for instance, having both the entire subset *and* the subset + GZ in memory?

matthewturk (approved):

yt/frontends/ramses/data_structures.py
cphyc: Yes; the reason I decided to store them in memory is that computing the ghost zones is (very) slow, as it grows roughly as `N log(N) (Ngz**3-8)`, where `N` is the number of octs in the subset.

matthewturk:
Makes sense, thanks!

PR #2963, opened 2020-11-11
Migrate from Travis to GitHub Actions

Xarthisius:
It's my attempt to migrate our cloud testing infra to GitHub actions. I believe it has a parity with current Travis setup. I skipped one of the test runs (Py3.6 unittests non-minimal), because I think it was redundant. However, it can be trivially brought back. Comments are welcome at this point.


.github/workflows/build_test.yaml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8, 3.9]
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - name: Install Python
```
matthewturk, line 36: There's a conda action available too

Xarthisius: I'm just wrapping my head around it. I'll improve it once I have a semi working thing

.github/workflows/build-test.yaml
```diff
@@ -0,0 +1,68 @@
...
+  HOMEBREW_NO_AUTO_UPDATE: 1
+
+jobs:
+
+  build:
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-18.04]
+        python-version: [3.8]
+        minimal: [false]
```
matthewturk, line 19: It looks to me like the variable value shows up in the GHA workflow name.  Would it be possible to instead of having `minimal: [true, false]` have something like `extent: [minimal, full]`?

.github/workflows/build-test.yaml
```diff
@@ -0,0 +1,68 @@
...
+        update-conda: true
+        conda-channels: conda-forge
+        activate-conda: true
+        python-version: ${{matrix.python-version}}
+    - name: Show env
+      shell: bash
+      run: |
+        echo PATH $PATH
+        echo python `which python`
+        python -c "import sys; print(sys.version)"
```
matthewturk, line 53: still need this?

.github/workflows/build-test.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - name: Install dependencies and yt
+      shell: bash
+      run: |
+        if [ "${{ matrix.minimal }}" == "true" ]; then
+          export MINIMAL=1
+        fi
+        source ./tests/ci_install.sh
+    - name: Run Tests
+      run: |
+        if [ "${{ matrix.answer-tests }}" == "true" ]; then
```
matthewturk, line 63: I believe ti's possible to use an `env:` [block](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#env) where we set the variables to the values here, as described [in this SO post](https://stackoverflow.com/questions/59302035/switch-environment-variable-in-github-actions). That might make this a bit more terse.

tests/ci_install.sh
```diff
@@ -1,9 +1,33 @@
 set -x   # Show which command is being run
 
-echo "Travis build stage: $TRAVIS_BUILD_STAGE_NAME"
+[[ -n ${RUNNER_OS} ]] && export TRAVIS_OS_NAME=${RUNNER_OS}
```
matthewturk, line 3: Would love to see this file be reduced/eliminated, but not sure if that's feasible...

matthewturk (approved):
This is awesome.  A few comments/suggestions in review.

munkm:
I will review this as soon as I finish the release. I'm totally in support of the migration!

tests/ci_install.sh
Xarthisius: I missed that. It was unnecessary and now it's dropped

.github/workflows/build-test.yaml
Xarthisius: Done.

.github/workflows/build-test.yaml
Xarthisius: nope, gone

.github/workflows/build-test.yaml
Xarthisius: done

appveyor.yml
```diff
@@ -1,15 +0,0 @@
...
-platform:
-    - x64
-
-install:
-    - "echo 'Not installing anything.'"
-
-build: false
-
-test_script:
-    - "echo 'Not testing anything. This is done on Travis'"
```
cphyc, line 15: Deleting the `appveyor.yml` file will make this PR and all subsequent one fail until we deactivate appveyor completely. I would suggest delaying this to a future PR (otherwise I think we'd have to update all existing PRs to remove the appveyor.yml file).

.github/workflows/build-test.yaml
```diff
@@ -0,0 +1,56 @@
...
+        python-version: ${{matrix.python-version}}
+    - name: Install dependencies and yt
+      shell: bash
+      env:
+        dependencies: ${{ matrix.dependencies }}
+      run: source ./tests/ci_install.sh
+    - name: Run Tests
+      env:
+        testsuite: ${{ matrix.tests-type }}
+      run: python -m nose -c nose_${testsuite}.cfg --traverse-namespace
```
cphyc, line 56: Aren't we losing the coverage information?

tests/ci_install.sh
```diff
@@ -1,9 +1,31 @@
...
+    sudo apt-get install \
+      libhdf5-serial-dev \
+      libnetcdf-dev \
+      libproj-dev \
+      proj-data \
+      proj-bin \
+      libgeos-dev \
+      libopenmpi-dev
+    ;;
+osx|macOS)
```
munkm, line 14: does there need to be a windows step here too?

.github/workflows/build-test.yaml
munkm: We will also probably need to modify the codecov webhook I think (?)

munkm (approved):
Thank you so much for doing this @Xarthisius!!! I'm generally in approval of this, pending a few updates. 

Do we want to add a Python 3.9 testing stage any time soon?

appveyor.yml
munkm: +1 to this. I think we should have dummy appveyor and travis files like what corentin already did with appveyor. What if we added a comment in both files that they'll be removed officially from master in ~3-6 months (and the date can be recorded in the file maybe) or something?

.github/workflows/build-test.yaml
Xarthisius: Is the coverage working right now?

appveyor.yml
Xarthisius: There's always going to be a point when there are outstanding PRs and removing this file on master will force them to merge/rebase. I don't think setting a date will make a difference. As for appveyor I was planning on disabling it even before this goes in.

tests/ci_install.sh
Xarthisius: nope, it's covered below.

appveyor.yml
Xarthisius: Also I don't think anyone will need to remove appveyor.yml, unless they made some changes locally that were not on master.

tests/ci_install.sh
```diff
@@ -12,34 +34,34 @@ if [[ "$TRAVIS_OS_NAME" == "windows" ]] && [[ $MINIMAL != 1 ]]; then
...
+    CFLAGS="$CFLAGS -DACCEPT_USE_OF_DEPRECATED_PROJ_API_H" python -m pip install -r tests/test_requirements.txt
 fi
 
 # Step 3: install yt
-$PIP install -e .
+python -m pip install -e .
 
-set +x  # Do not show which command is being run (Travis default)
\ No newline at end of file
+set +x  # Do not show which command is being run (Travis default)
```
neutrinoceros, line 67: Is this line still necessary ? Even if it is, the comment should be updated

tests/ci_install.sh
```diff
@@ -1,9 +1,31 @@
...
+    sudo chown -R "${USER}:admin" /usr/local/man
+    HOMEBREW_NO_AUTO_UPDATE=1 brew install hdf5 proj geos open-mpi netcdf ccache
+    ;;
+esac
+
+mkdir -p $HOME/.config/yt
+echo "[yt]" > $HOME/.config/yt/ytrc
+echo "suppressStreamLogging = True" >> $HOME/.config/yt/ytrc
+cat $HOME/.config/yt/ytrc
+cp tests/matplotlibrc .
```
neutrinoceros, line 25: Is this strictly part of the migration process ? I can't find where this came from or why it's a necessary step.

matthewturk:
This seems like a vast improvement.  Is there anything left?  It looks like coverage (maybe) and some comments?

tests/ci_install.sh
Xarthisius: It's not really part of migration process, it's always been there. It's our way to setup a proper matplotlib backend to something that's non-interactive / doesn't open windows.

tests/ci_install.sh
Xarthisius: I don't know what's the default for GHA environment, but I don't think that setting it explicitly is a bad thing either. I'll drop the comment.

tests/ci_install.sh
neutrinoceros: ok ! I do think a comment would be nice though :)

appveyor.yml
munkm: > There's always going to be a point when there are outstanding PRs and removing this file on master will force them to merge/rebase. 

Yes that's true. The reason I proposed this is because we've asked contributors to do this several times this year, which hasn't been the case in the past. I could see that be frustrating or even offputting.

Xarthisius:
Just to sum up the current status:

* coverage is not working with our current infra and hasn't been for a longtime (check out https://codecov.io/gh/yt-project/yt). I basically removed it because it's broken. I hope we can add it as a part of future PR
* Appveyor has been disabled and the removal of `appveyor.yml` just completes the process. It shouldn't affect existing PRs in anyway, except for those PRs that explicitly modify content of that file.

@munkm @neutrinoceros @cphyc is there anything else for me to address?

neutrinoceros (approved):
> @munkm @neutrinoceros @cphyc is there anything else for me to address?

Afaic this is now good to go !

cphyc:
> @munkm @neutrinoceros @cphyc is there anything else for me to address?

It's good to go for me!

munkm:
🎉 thank you so much @Xarthisius !!!!

Issue #832, opened 2014-04-22
Ambiguous field names should error instead of guessing the field type


yt-fido:
Originally reported by: **jmoloney (Bitbucket: [jmoloney](https://bitbucket.org/jmoloney), GitHub: [jmoloney](https://github.com/jmoloney))**

----------------------------------------

For Enzo datasets, using the field name 'dynamical_time' instead of the full field ('all', 'dynamical_time'), ('io', 'dynamical_time'), or ('gas', dynamical_time') produces inconsistent results. As far as I can tell, this is because it is the only field that is defined with the same name for both particles and gas. Depending on which fields have been called previously, the field returned by 'dynamical_time' may be either one or the other. I haven't been able to figure out a clear pattern for when each is returned.

This example illustrates the issue:

```
#!python

import yt

pf = yt.load('IsolatedGalaxy/galaxy0030/galaxy0030')
reg = pf.h.all_data()

key_list = ['dynamical_time','Density','dynamical_time','density',
	'dynamical_time', 'temperature', 'dynamical_time', 'particle_mass',
	'dynamical_time', 'cooling_time', 'dynamical_time', 'cooling_time',
	'dynamical_time', 'cooling_time', 'dynamical_time', 'Density',
	'dynamical_time','density','dynamical_time']

for key in key_list:
    print key, ' ', reg[key].size
```

----------------------------------------
- Bitbucket: https://bitbucket.org/yt_analysis/yt/issue/832

ngoldbaum:
I still think this is a good idea. That said, it's probably too disruptive for any 3.x release, so I'm going to add a 4.0 label.

chummels:
This topic was just discussed at a yt-4.0 release meeting.  It was agreed that in yt-4.0, we will accept both string and tuple field names.  If someone specifies a string field name and there is any ambiguity if there are two potential tuple fields that could be accessed, the code will error out with an error message requiring the user to be more explicit. 

The issue was raised that it might be confusing to some users that certain fields differ only by capitalization (e.g., "Density" and "density" for the Enzo frontend: ("Enzo", "Density") and ("gas", "density")), but the consensus was to remain case-sensitive, and just include a note in the Field documentation that this was the expected behavior.

chummels:
This issue is a repeat of Issue #1467 .  Closing this one so we are only tracking it with one issue.

PR #2967, opened 2020-11-13
Throw an error if field accesses is ambiguous.

matthewturk:
## PR Summary

This addresses both #832 and #1467.

It instruments `__setitem__` in the field info container to check if there is a possible ambiguity, which we then check against in the "guessing" portion of `_get_field_info`.  We need to investigate the potential performance penalty for this instrumentation of `__setitem__` but I hope it will be minimal.

This may also break our tests, but ideally it will not.  I think it will likely break workflows where the particle type is not set, as it will throw an error for particle unions that are identical to particle types (i.e, if you just have 'nbody' particles, then 'nbody' and 'all' will be the same.)

## PR Checklist

- [x] pass `flake8 yt/`


yt/utilities/exceptions.py
```diff
@@ -872,3 +872,16 @@ class GenerationInProgress(Exception):
...
+class YTAmbiguousFieldName(YTException):
+    def __init__(self, fname, possible_ftypes):
+        self.fname = fname
+        self.possible_ftypes = possible_ftypes
+
+    def __str__(self):
+        msg = f"You requested the field '{self.fname}'.\n"
+        msg += "Unfortunately, that could correspond to any one of "
+        msg += f"the field types\n {self.possible_ftypes}."
+        msg += "Please explicitly utilize a tuple to select the ftype you mean."
```
neutrinoceros, line 886:
```suggestion
    msg = (
        f"The requested field name '{self.fname}'"
        "is ambiguous and corresponds to any one of "
        f"the following field types\n {self.possible_ftypes}."
        "Please specify the requested field as an explicit "
        "tuple (ftype, fname)."
    )
```

yt/fields/field_info_container.py
```diff
@@ -402,6 +404,12 @@ def add_output_field(self, name, sampling_type, **kwargs):
         kwargs.setdefault("ds", self.ds)
         self[name] = DerivedField(name, sampling_type, NullFunc, **kwargs)
 
+    def __setitem__(self, key, value):
+        ftype, fname = key
+        if any(fname == _fname for _ftype, _fname in self):
+            self._ambiguous_field_names[fname].append(ftype)
```
neutrinoceros, line 410: I'd be happier with an implementation that couldn't append redundant values, though I could easily be convinced that this is fine.

neutrinoceros (approved):
Thank you for doing this. I left minor comments and suggestions but as a whole I'm happy with the PR as it is :)

yt/fields/field_info_container.py
matthewturk: I've changed it to a set!

neutrinoceros: Perfect. This was my original suggestion but I forgot about `set.add` and thought there wasn't an equivalent to `list.append` :)

yt/fields/field_info_container.py
```diff
@@ -402,6 +404,12 @@ def add_output_field(self, name, sampling_type, **kwargs):
         kwargs.setdefault("ds", self.ds)
         self[name] = DerivedField(name, sampling_type, NullFunc, **kwargs)
 
+    def __setitem__(self, key, value):
+        ftype, fname = key
+        if any(fname == _fname for _ftype, _fname in self):
```
cphyc, line 409: It took me multiple passes to understand this bit. It would help to add a comment here
```suggestion
        # Mark fields with same `fname`s (but difference `ftype`s) as ambiguous
        if any(fname == _fname for _ftype, _fname in self):
```

yt/data_objects/static_output.py
```diff
@@ -817,6 +818,10 @@ def _get_field_info(self, ftype, fname=None):
         # storing this condition before altering it
         guessing_type = ftype == "unknown"
         if guessing_type:
+            if fname in self.field_info._ambiguous_field_names:
```
cphyc, line 821: It may be more robust to create a method for this, to prevent accessing an "hidden" attribute (i.e. starting with a `_`).
```suggestion
            if self.field_info.is_ambiguous_field_name(fname):
```

yt/fields/field_info_container.py
```diff
@@ -402,6 +404,12 @@ def add_output_field(self, name, sampling_type, **kwargs):
...
+            self._ambiguous_field_names[fname].add(ftype)
+        super().__setitem__(key, value)
+
```
cphyc, line 412: I'd suggest adding the method
```suggestion

     def is_ambiguous_field_name(self, fname):
         """Detect ambiguous field names
         
         Parameters
         ------------
         fname : str, the field name to check
         
         Returns
         --------
         True if the field is ambiguous, False otherwise. Ambiguous fields
         are fields with the same field name but different field types.
         """
         return fname in self._ambiguous_field_names
 
```

cphyc (approved):
I am glad this feature is finally there, this is really great! 

I am fine with this being merged with or without my comments taken into account (and assuming tests pass).

yt/utilities/exceptions.py
```diff
@@ -872,3 +872,19 @@ class GenerationInProgress(Exception):
...
+class YTAmbiguousFieldName(YTException):
+    def __init__(self, fname, possible_ftypes):
+        self.fname = fname
+        self.possible_ftypes = possible_ftypes
+
+    def __str__(self):
+        msg = (
+            f"The requested field name '{self.fname}'"
+            "is ambiguous and corresponds to any one of "
+            f"the following field types\n {self.possible_ftypes}."
```
cphyc, line 886:
```suggestion
            f"the following field types\n {self.possible_ftypes}. "
```

yt/utilities/exceptions.py
```diff
@@ -872,3 +872,19 @@ class GenerationInProgress(Exception):
...
+            "Please specify the requested field as an explicit "
+            "tuple (ftype, fname)."
```
cphyc, line 888:
```suggestion
            f"The requested field name '{self.fname}' "
            "is ambiguous and corresponds to any one of "
            f"the following field types\n {self.possible_ftypes}. "
            "Please specify the requested field as an explicit "
            "tuple (ftype, fname)."
```

neutrinoceros: Oops, indeed I got the spacing wrong it seems.

cphyc:
In order to magically get diffs that would fix the failing tests, you can use this script:
```python
import sys
import os
from difflib import unified_diff

from nose.plugins import Plugin
from yt.utilities.exceptions import YTAmbiguousFieldName
from nose.plugins.plugintest import run

ROOT = os.path.abspath(__file__)

class AmbiguousResolvePlugin(Plugin):
    """Plugin that takes no command-line arguments"""
    name = "ambiguous-solver"
    enabled = True

    def configure(self, options, conf):
        pass

    def options(self, parser, env={}):
        pass

    def addError(self, test, err: sys.exc_info):
        try:
            test_path = test.context.__file__
        except:
            test_path = os.path.join(
                ROOT,
                test.context.__module__.replace(".", "/")
            )
        t, v, tb = err
        if t is not YTAmbiguousFieldName:
            print("HERE", t)
            return

        import traceback
        # print('TYPE:', t)
        # print('VALUE:', v)
        # print('TRACEBACK:', tb)

        ambiguous_fname = v.fname
        ok = False
        for ft, _ in traceback.walk_tb(tb):
            line = open(ft.f_code.co_filename, "r").readlines()[ft.f_lineno].replace("\n", "")
            # print(f"{ft.f_code.co_filename}:{ft.f_code.co_firstlineno} {line}")
            if test_path == ft.f_code.co_filename:
                ft_err = ft
                ok = True

        if not ok:
            return

        # Now, ft contains the current frame,
        # need to correct the test!
        # print(f"error from {ft_err.f_code.co_filename}:{ft_err.f_code.co_firstlineno}:{ft_err.f_lineno}")

        with open(test_path, 'r') as f:
            lines = f.readlines()

        lineno = ft_err.f_lineno - 1

        suggested_ftype = v.possible_ftypes[-1]
        corrected = lines.copy()
        corrected[lineno] = corrected[lineno].replace(
            '"%s"' % ambiguous_fname,
            '("%s", "%s")' % (suggested_ftype, ambiguous_fname)
        ).replace(
            "'%s'" % ambiguous_fname,
            "('%s', '%s')" % (suggested_ftype, ambiguous_fname)
        )

        sys.stderr.writelines(
            unified_diff(lines, corrected, fromfile=test_path, tofile=test_path)
        )


run(
    argv=["nosetests", "yt"],
    plugins=[AmbiguousResolvePlugin()]
)
```
---
Example
```bash
$ pwd
/path/to/yt/root
$ python nose_debug.py > /dev/null # Diffs are shown in stderr
--- /path/to/yt/root/yt/yt/data_objects/level_sets/tests/test_clump_finding.py
+++ /path/to/yt/root/yt/yt/data_objects/level_sets/tests/test_clump_finding.py
@@ -42,7 +42,7 @@
     master_clump.add_validator("min_cells", 1)
 
     def _total_volume(clump):
-        total_vol = clump.data.quantities.total_quantity(["cell_volume"]).in_units(
+        total_vol = clump.data.quantities.total_quantity([("stream", "cell_volume")]).in_units(
             "cm**3"
         )
         return "Cell Volume: %6e cm**3.", total_vol
--- /path/to/yt/root/yt/yt/data_objects/level_sets/tests/test_clump_finding.py
+++ /path/to/yt/root/yt/yt/data_objects/level_sets/tests/test_clump_finding.py
@@ -105,7 +105,7 @@
     leaf_clumps = master_clump.leaves
 
     fn = master_clump.save_as_dataset(
-        fields=["density", "x", "y", "z", "particle_mass"]
+        fields=["density", ("enzo", "x"), "y", "z", "particle_mass"]
     )
     ds2 = load(fn)
```

As you can see with the last bit, it is far from perfect!

cphyc:
> 
> As you can see with the last bit, it is far from perfect!

Also, there is some freedom to pick which `ftype` should be used. By default here, it takes the last one but since it is fed from a set, well, that doesn't really matter.

Note that you'll likely need to apply the following patch (if it is indeed a fix for the many fails that popped...)
```diff
--- a/yt/data_objects/derived_quantities.py
+++ b/yt/data_objects/derived_quantities.py
@@ -23,7 +23,7 @@ def get_position_fields(field, data):
             ftype = finfo.name[0]
         position_fields = [(ftype, f"particle_position_{d}") for d in axis_names]
     else:
-        position_fields = axis_names
+        position_fields = [("index", d) for d in axis_names]
 
     return position_fields
```

cphyc:
What's the status of this PR?

matthewturk:
Mostly: it's harder than I thought, and then I got caught up in a bunch of
stuff.  So, help would be greatly appreciated!

cphyc:
FYI I have applied the nose trick in https://github.com/yt-project/yt/pull/3092 to catch as many ambiguous access calls as possible. Apart from removing ambiguous accesses, is there something more you want to add in this PR @matthewturk ?

jzuhone:
@matthewturk do you want help with this? I can look at the errors

cphyc:
> @matthewturk do you want help with this? I can look at the errors

@jzuhone I've investigated a bit the issue(s) and the problem is that most of our tests use ambiguous field accesses. For example there are hundreds of occurences of `ad["x"]`, but `x` is provided by `("gas", "x")`, `("index", "x")`, `("stream", "x")` so each of one will result in an error. I have came up with an automated way of fixing them (using the type inferred by yt to replace the ambiguous line), which is located here #3101 .

munkm (approved):
Right now I think this looks good! 

In the triage meeting today we decided to add an interim step for the ambiguous fields that for 4.0 we will issue a warning about the field being ambiguous to allow users to migrate scripts. In 4.1 the error will start and hopefully by then users will have shifted over scripts.

cphyc:
@yt-fido test this please

cphyc:
I am expected a handful of tests to fail due to the pixelization routines relying on ambiguous fields that I couldn't fix in #3101.

cphyc:
Note for other reviewers: I have updated the PR following @munkm comments about throwing a warning rather than an error until 4.1.0 and added some more tests. I am still marked as "approving" the PR, but at this stage I'd rather be considered a contributor than a reviewer, so please ignore my approval!

neutrinoceros (commented):
Since the tests are using pytest, the test file should be added to the ignore lists in tests/tests.yaml and nose_unit.cfg

yt/fields/tests/test_ambiguous_fields.py
```diff
@@ -0,0 +1,44 @@
...
+    # Test warnings are issued for ambiguous fields
+    for fname in ambiguous_fnames:
+        with pytest.warns(VisibleDeprecationWarning, match=msg.format(fname)):
+            ds.r[fname]
+
+    # Test no warnings are issued for single fname access that aren't ambiguous
+    for fname in unambiguous_fnames:
+        with pytest.warns(None) as record:
+            ds.r[fname]
+        assert len(record) == 0
```
neutrinoceros, line 37: shouldn't the assert statement be part of the with block ?

yt/fields/tests/test_ambiguous_fields.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Test no warning are issued for tuple access
+    for ftype, fname in ds.field_list:
+        with pytest.warns(None) as record:
+            ds.r[(ftype, fname)]
+
+        assert len(record) == 0
```
neutrinoceros, line 44: same as above

yt/fields/tests/test_ambiguous_fields.py
```diff
@@ -0,0 +1,44 @@
...
+    assert len(unambiguous_fnames) > 2
+
+    # Test warnings are issued for ambiguous fields
+    for fname in ambiguous_fnames:
+        with pytest.warns(VisibleDeprecationWarning, match=msg.format(fname)):
+            ds.r[fname]
+
+    # Test no warnings are issued for single fname access that aren't ambiguous
+    for fname in unambiguous_fnames:
+        with pytest.warns(None) as record:
```
neutrinoceros, line 35: I've never seen this used this way before and I can't find it mentioned in the documentation, are you sure this does what you think it does and can you point a reference for that ?

neutrinoceros (approved):
I got a few questions regarding the test but aside from that, I think this is very well done.

yt/data_objects/static_output.py
```diff
@@ -821,6 +821,15 @@ def _get_field_info(self, ftype, fname=None):
...
+                msg = (
+                    f"The requested field name '{fname}' "
+                    "is ambiguous and corresponds to any one of "
+                    f"the following field types\n {self.field_info._ambiguous_field_names[fname]}. "
+                    "Please specify the requested field as an explicit "
+                    "tuple (ftype, fname)."
```
chummels, line 830: I think it would be beneficial to inform the user which field it defaulted to in this pass, by including something like: "Defaulting to ('ftype', 'fname')."

chummels (changes requested):
I tested this out, and it works great.  I just have the requested comment above about informing the user what field was actually used.

yt/fields/tests/test_ambiguous_fields.py
cphyc: Here it is https://docs.pytest.org/en/stable/warnings.html#recwarn!

neutrinoceros: perfect 👌

yt/fields/tests/test_ambiguous_fields.py
cphyc: It doesn't seem to matter, but I am following here the doc.

yt/fields/tests/test_ambiguous_fields.py
cphyc: See above :)

yt/data_objects/static_output.py
cphyc: Should be good now!

yt/data_objects/static_output.py
```diff
@@ -808,6 +808,22 @@ def _setup_particle_types(self, ptypes=None):
...
+
+        if ambiguous:
+            ft, fn = field_info.name
+            msg = (
+                f"The requested field name '{fn}' "
+                "is ambiguous and corresponds to any one of "
+                f"the following field types:\n {self.field_info._ambiguous_field_names[fn]}\n"
+                "Please specify the requested field as an explicit "
+                "tuple (ftype, fname).\n"
+                f"Defaulting to '({ft}, {fn})'."
```
neutrinoceros, line 821: so the suggestion can be copy-pasted back into the culprit code
```suggestion
                f"Defaulting to ('{ft}', '{fn}')."
```

yt/data_objects/static_output.py
```diff
@@ -822,17 +838,20 @@ def _get_field_info(self, ftype, fname=None):
         guessing_type = ftype == "unknown"
         if guessing_type:
             ftype = self._last_freq[0] or ftype
+            ambiguous = fname in self.field_info._ambiguous_field_names
```
neutrinoceros, line 841: So that it's more obvious that it's a bool
```suggestion
            is_ambiguous = fname in self.field_info._ambiguous_field_names
```

yt/data_objects/static_output.py
```diff
@@ -822,17 +838,20 @@ def _get_field_info(self, ftype, fname=None):
...
+        else:
+            ambiguous = False
```
neutrinoceros, line 843:
```suggestion
            is_ambiguous = False
```

yt/data_objects/static_output.py
```diff
@@ -822,17 +838,20 @@ def _get_field_info(self, ftype, fname=None):
...
         field = (ftype, fname)
 
         if (
             field == self._last_freq
             and field not in self.field_info.field_aliases.values()
         ):
-            return self._last_finfo
+            return self._last_finfo, ambiguous
```
neutrinoceros, line 850:
```suggestion
            return self._last_finfo, is_ambiguous
```

yt/data_objects/static_output.py
```diff
@@ -822,17 +838,20 @@ def _get_field_info(self, ftype, fname=None):
...
         if field in self.field_info:
             self._last_freq = field
             self._last_finfo = self.field_info[(ftype, fname)]
-            return self._last_finfo
+            return self._last_finfo, ambiguous
```
neutrinoceros, line 854:
```suggestion
            return self._last_finfo, is_ambiguous
```

yt/data_objects/static_output.py
```diff
@@ -850,7 +869,7 @@ def _get_field_info(self, ftype, fname=None):
             ):
                 field = self.default_fluid_type, field[1]
             self._last_freq = field
-            return self._last_finfo
+            return self._last_finfo, ambiguous
```
neutrinoceros, line 872:
```suggestion
            return self._last_finfo, is_ambiguous
```

yt/data_objects/static_output.py
```diff
@@ -866,7 +885,7 @@ def _get_field_info(self, ftype, fname=None):
                 if (ftype, fname) in self.field_info:
                     self._last_freq = (ftype, fname)
                     self._last_finfo = self.field_info[(ftype, fname)]
-                    return self._last_finfo
+                    return self._last_finfo, ambiguous
```
neutrinoceros, line 888:
```suggestion
                    return self._last_finfo, is_ambiguous
```

yt/data_objects/static_output.py
neutrinoceros: the way you implemented it now works too

munkm (approved):
I think this is looking great now!

chummels (approved):

PR #2968, opened 2020-11-13
Allow .plot() to run outside of the notebook

matthewturk:
## PR Summary

We shouldn't necessarily fail if we call `.plot()` on slices or projections
outside of the notebook, since it's also useful just to get a `PlotWindow`
object back.  This catches the one non-fatal Exception.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -27,6 +27,7 @@ from yt.units.unit_object import Unit
 from yt.units.yt_array import YTArray, uconcatenate
 from yt.utilities.exceptions import (
     YTNoAPIKey,
+    YTNotInsideNotebook,
     YTParticleDepositionNotImplemented,
     YTTooManyVertices,
 )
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Interesting. While reviewing, I wondered if there are any other places in the codebase where `YTNotInsideNotbook` could be raised, and if they should be updated as well. Now I'm thinking: is this error _ever_ critical ? If not, would it make sense to replace it with a warning ?

matthewturk:
Good point, @neutrinoceros .  I looked through the code, and it looks like we raise it inside the `.show()` calls for `ProfilePlot`, `Scene`, `Camera` (in `old_camera.py`), and when the interactive engine isn't set in `PlotContainer` and in `display_in_notebook`.

But that's just *direct* raises.  I think *those* places it might make sense (we may still want to change it), simply because we're explicitly asking for the plot to display, and they don't have any side effects of returning a value.  The calls to `.plot()` have the side effect of both creating a plot *and* showing it.

That being said, yeah, we could probably just switch all of these over to warnings instead of exceptions.

neutrinoceros (approved):
Ok, considering the clarifications above, I'm inclined to consider this a bugfix. The suggested changes are not required to get this in, imo.

munkm (approved):

munkm:
> That being said, yeah, we could probably just switch all of these over to warnings instead of exceptions. 

I think that could be done in a future PR. This PR is a nice separate enhancement and if needed those can be updated separately. 🙂

PR #2972, opened 2020-11-18
Build process testing

cphyc:
## PR Summary

This PR adds support for a regular (which event should trigger this should be discussed) test of the building process of yt on different combinations of compilers & platforms. So far, I have only implemented Ubuntu + gcc-5.

## PR Checklist

- [ ] pass `flake8 yt/`

### DIFF
--- a/.github/workflows/daily-test.yaml
+++ b/.github/workflows/daily-test.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8]
         c_compiler: [gcc, gcc-5]
         cpp_compiler: [g++, g++-5]
         exclude:
--- a/.github/workflows/daily-test.yaml
+++ b/.github/workflows/daily-test.yaml
@@ -5,6 +5,11 @@ on:
     # Run daily at 00:00
     - cron: '0 0 * * *'
   push:
+  pull_request:
+
+defaults:
+  run:
+    shell: bash
 
 jobs:
   build:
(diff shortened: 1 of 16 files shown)


.github/workflows/daily-test.yaml
```diff
@@ -0,0 +1,49 @@
+name: Daily build
+
+on:
+  schedule:
```
Xarthisius, line 4: Why daily instead of regular push/PR? There are days (sometimes weeks) passing by when master doesn't change.

.github/workflows/daily-test.yaml
```diff
@@ -0,0 +1,49 @@
...
+    shell: bash
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+        python-version: [3.8]
+        c_compiler: [gcc, gcc-5]
+        cpp_compiler: [g++, g++-5]
```
Xarthisius, line 21: Default gcc/g++ will be tested in the regular build run. Adding it here makes it a bit redundant. Other OSes and their non-standard compilators would be interesting though, but maybe that's out of scope for this PR.

.github/workflows/daily-test.yaml
cphyc: This was following our discussion in the yt 4 meeting that we didn't want these builds to happen on all PR to reduce the burden from contributors. But I guess we could leave it to push events on master (i.e. when PRs get merged).

.github/workflows/daily-test.yaml
cphyc: True, I'll remove `gcc` from the building matrix. What other combinations would you suggest?

PR #2973, opened 2020-11-22
ci: dead code detection gh workflow

### LINKED ISSUE #2872: Dead code detection
### Bug report

**Bug summary**

Just found out about [vulture](https://pypi.org/project/vulture/#description), a python CLI app that attempts to detect dead code. Maybe this tool could be used to help and detect dead/unreachable code, depending how well it can be adjusted to avoid false positive. Worst case scenario this provides help in detecting untested code very quickly without running the test suite or using coverage.

**Code for reproduction**

A basic report can be generated such as
```bash
vulture yt --sort-by-size > dead_code.txt
```
A lot more finesse can be achieved by defining a whitelist (e.g. for __init__ and api files).

**Actual outcome**

[dead_code.txt](https://github.com/yt-project/yt/files/5081354/dead_code.txt)

**Expected outcome**

Nothing :-)

neutrinoceros:
## PR Summary

fix #2872

todo
- [x] validate a gh workflow to run vulture and catch unreachable code
- [ ] remove detected unreachable code

### DIFF
--- a/.github/workflows/dead_code.yaml
+++ b/.github/workflows/dead_code.yaml
@@ -1,0 +1,22 @@
+name: Dead code detection
+on: [pull_request]
+
+jobs:
+  vulture:
+    name: vulture
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Setup Python
+      uses: actions/setup-python@master
+      with:
+        python-version: 3.8
+    - name: install
+      id: install
+      run : pip install vulture
+
+    - name: check
+      id: vulture
+      run: |
+        vulture --version
+        vulture yt --sort-by-size --min-confidence 100 --exclude "**/api.py" --exclude "yt/extern"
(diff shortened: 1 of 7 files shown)


matthewturk (approved):

munkm (commented):
I think this is a cool workflow! I guess we should discuss whether it should be a blocker for all PRs or not? Can we set up the workflow for a nightly job or something? I worry slightly that adding a ton of checks could be a barrier to entry for a new contributor.

neutrinoceros:
I think this particular one will rarely be blocking, but that it should be.
What I'm more hesitant about right now is that it catches some dead bits that are unused arguments in functions, some of which are "required" in methods that follow a distinct template... the way vulture supports whitelist however is pretty cumbersome and would add too much maintenance cost for the benefit it offers imo.

neutrinoceros:
I changed my mind about making this a workflow. It would be nice to add vulture to our pre-commit hooks though, but it's lacking support upstream. I may reopen this (or a very similar PR) after evaluating if it's worth it, and if the required change is accepted upstream.

PR #2976, opened 2020-11-23
doc: add pypi versions badge

neutrinoceros:
## PR Summary

This should only be merged a short time ahead of the 4.0 release in order to be relevant.
Currently, this looks like this:
Screen Shot 2020-11-23 at 22 37 17

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![Users' Mailing List](https://img.shields.io/badge/Users-List-lightgrey.svg)](https://mail.python.org/archives/list/yt-users@python.org//)
 [![Devel Mailing List](https://img.shields.io/badge/Devel-List-lightgrey.svg)](https://mail.python.org/archives/list/yt-dev@python.org//)
 [![Build Status](https://img.shields.io/travis/yt-project/yt.svg?branch=main)](https://travis-ci.org/yt-project/yt)
+![PyPI - Python Version](https://img.shields.io/pypi/pyversions/yt)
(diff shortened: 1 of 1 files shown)


munkm (approved):
+1 to adding this badge. I agree that it should be merged around the 4.0 release (so that the outdated support info isn't advertised on our gh page) but I don't think it specifically matters if it's included in the release. Based on yt-project/yt#2528 and others that change our setup.py the badge should be updating with whatever is there. We should make sure that setup.py doesn't include python 3.5 in the release though. 

I think maybe the badge should come after the mailing list badges, since the badges afterwards are about the code/build?

neutrinoceros:
>  I don't think it specifically matters if it's included in the release

absolutely, I merely put the label so that we don't forget about it when the release is done !

> I think maybe the badge should come after the mailing list badges, since the badges afterwards are about the code/build?

sure thing

neutrinoceros:
Note: with this many badges, they don't fit on a single line anymore so here's how I'm rearranging them thematically (code properties/stability VS useful links). I'm open to suggestions if that's unsatisfactory for any reason.
Screen Shot 2020-11-24 at 23 34 10

munkm:
I think this looks great!

matthewturk:
We could ditch the data hub one and the two user list ones.

neutrinoceros:
> We could ditch the data hub one and the two mailing list ones.

why ? IMO having 2 lines of badges is better than reducing the discoverability of these. Maybe you'd like to replace them with a slack button or something ?

neutrinoceros:
Resolved in #3058

PR #2980, opened 2020-11-25
[EXP] ci: simplify windows build, stop using conda

neutrinoceros:
## PR Summary
In support of #2978 , I'm trying here to get rid of conda in the windows test build. According to @cphyc , the original motivation to use conda there was to easily select a python version on windows. I think it is likely not necessary anymore since we moved out of travis to GitHub actions. Let's try this out.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -38,13 +38,6 @@ jobs:
       with:
         submodules: recursive
         fetch-depth: 0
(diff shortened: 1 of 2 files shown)


Xarthisius:
I think the original motivation for using conda on windows is that it's the only reasonable way to get C/C++ dependencies. It's also the most common env that users have on windows.

neutrinoceros:
Well in that case I guess I should just drop this altogether. Thanks for your insight !

PR #2893, opened 2020-09-04
Clarify some utility functions (`yt.funcs`) with more_itertools

neutrinoceros:
## PR Summary

Refactor some internals using `more_itertools` (which is already a transitive dependency).

### remove `yt.funcs.ensure_list`
this function has two use cases
- it is used in many places to iterate over one or more objects transparently. While `ensure_list` is bit clunky and requires maintenance on our part, the more powerful `more_itertools.always_iterable` does the job.
- in places it is used as a sanitizer, it can be replaced by `lambda x: list(always_iterable(x))`

When we specifically need to iterate transparently over a arbitrary number of fields, `always_iterable` doesn't work out of the box since a single 2-tuple should be treated as a single field, but it's easy to configure, so I added a simple `yt.funcs.iter_fields` function:
```python
def iter_fields(field_or_fields):
    return always_iterable(field_or_fields, base_type=(tuple, str, bytes))
```

### rename `yt.funcs.iterable`
This function's name is confusing because it doesn't indicate that it's really a checker that returns a boolean (`is_iterable` would be clearer in that sense). On top of this it is also wrong because it actually only checks that `len(obj)` is valid, while some iterables have no length (e.g. infinite generators).
I renamed the existing function to `is_sequence`.

other associated renames
- `yt.funcs.validate_iterable` -> `validate_sequence`
- `yt.geometry.coordinates.coordinate_handler.validate_sequence_width` -> `validate_sequence_width`

### rework `yt.funcs.ensure_tuple` and `yt.funcs.just_one` 

Each can be simplified as combination of two functions from `more_itertools`.

Additionally, `ensure_tuple` is now only ever used to validate `ds.periodicity`, so there is some overlapping with #2868 and the functionality could be moved to the `Dataset` class directly.

This PR refactors the one place that `ensure_tuple` was used differently, namely `yt.frontends.enzo_p.misc.nested_dict_get`.
I also fixed a bug (and added a test for it) in this function while I was at it.

## PR Checklist

- [x] rename `iterable` to `is_sequence`
- [x] add an `iter_fields` function
- [x] search and replace `ensure_list` calls with either `iter_fields` where an iterator is needed, `list(iter_fields())` where we actually need a list of fields, and `list(always_iterable())` elsewhere.
- [x] remove `ensure_list`
- [x] search and replace `ensure_tuple` calls with `tuple(always_iterable())` (or update ensure_tuple to the same effect ?)
- [x] rework `just_one` 
- [x] explicitly add more_itertools to deps
[...]

PR #2981, opened 2020-11-25
Migrate config to toml

cphyc:
## PR Checklist

- [x] pass `flake8 yt/`


neutrinoceros:
looks like a lot of the changes are just flynt running lose on documentation (which was not done already, to my surprise ?). It's nice but I maybe it's out of scope here, or am I missing something ?

yt/config.py
```diff
@@ -66,68 +73,142 @@
...
+        if defaults is None:
+            defaults = {"yt": {}}
+        self.defaults = defaults
+        self.values = {}
+        self.update(defaults)
+
+    def get(self, section, *path, strict=True, **kwargs):
+        config = self.values[section]
+
+        # This works as follow: if we try to access
```
matthewturk, line 116: Does this leave open the opportunity to do frontend-specific config?

yt/config.py
```diff
@@ -66,68 +73,142 @@
...
+        # field > gas > density > lognorm
+        # we try in this order:
+        #   field > gas > density > lognorm
+        #   field > gas > lognorm
+        #   field > lognorm
+        if len(path) == 0:
+            return config
+
+        ok = False
```
matthewturk, line 125: wow this statement is a real mood, I have to say

yt/config.py
```diff
@@ -66,68 +73,142 @@
...
-ytcfg.read([_OLD_CONFIG_FILE, CURRENT_CONFIG_FILE, "yt.cfg"])
-if not ytcfg.has_section("yt"):
-    ytcfg.add_section("yt")
+# Walk the tree up until we find a config file
+ytcfg = YTConfig(ytcfg_defaults)
+if os.path.exists(CURRENT_CONFIG_FILE):
+    ytcfg.read(CURRENT_CONFIG_FILE)
+cwd = Path(".").absolute()
+while cwd.parent != cwd:
+    cfg_file = cwd / "yt.toml"
```
matthewturk, line 207: love pathlib

yt/config.py
```diff
@@ -66,68 +73,142 @@
...
+    if cfg_file.exists():
+        ytcfg.read(cfg_file)
+        break
+    cwd = cwd.parent
```
matthewturk, line 211: I'm looking at this and trying to figure out if it's possible to get to the root, or to get into an infinite loop, and I can't quite tell.  Is it ever possible we'll see an exception for trying to get to the parent of a non-parented path?

matthewturk (approved):
I like this.  One question I had, since we're in a mood to add the occasional dependency, is if there is a config library out there (that perhaps we even have a transitive dependence on) that would be a good idea to migrate to.

yt/config.py
```diff
@@ -66,68 +73,164 @@
...
 
-ytcfg = YTConfigParser(ytcfg_defaults)
-ytcfg.read([_OLD_CONFIG_FILE, CURRENT_CONFIG_FILE, "yt.cfg"])
-if not ytcfg.has_section("yt"):
-    ytcfg.add_section("yt")
+# Walk the tree up until we find a config file
+ytcfg = YTConfig(ytcfg_defaults)
+if os.path.exists(CURRENT_CONFIG_FILE):
+    ytcfg.read(CURRENT_CONFIG_FILE)
+cwd = Path(".").absolute()
```
neutrinoceros, line 227:
```suggestion
cwd = Path.cwd()
```

yt/config.py
```diff
@@ -66,68 +73,164 @@
...
+while cwd.parent != cwd:
```
neutrinoceros, line 228: I don't get how this loop is supposed to stop.

yt/config.py
cphyc: It does! You can actually do something like 
image

yt/config.py
cphyc: Changed it, sorry!

yt/config.py
cphyc: I _think_ the code is correct, but I'll let others weigh in.

yt/config.py
cphyc: It stops when you reach the root, since the parent of the root is the root itself (i.e. `Path("/") == Path("/").parent`). AFAIK this is only true when you reach the root. I'm curious how this is handled on WIndows though.

cphyc:
> looks like a lot of the changes are just flynt running lose on documentation (which was not done already, to my surprise ?). It's nice but I maybe it's out of scope here, or am I missing something ?

I've moved them to #2983

yt/config.py
neutrinoceros: Cool, I didn't know about this !

cphyc:
> I like this. One question I had, since we're in a mood to add the occasional dependency, is if there is a config library out there (that perhaps we even have a transitive dependence on) that would be a good idea to migrate to.

I could not find any library that was relying on toml, but given the size of our configuration, this may be a bit overkill. Currently, the config manager repropduces part of the ConfigParser API.

neutrinoceros:
I have a general question here: is it still relevant to store the global config file in `$HOME/.config/yt/yt.toml` rather than `$HOME/.config/yt.toml` ? what else is expected to be there ?

yt/config.py
cphyc: I updated this bit as it was causing me headache to make sure it will work in *any* case.

yt/config.py
```diff
@@ -66,74 +76,132 @@
...
+            self.config_root.get_child(section)
+            return True
+        except KeyError:
+            return False
+
+    def add_section(self, section):
+        self.config_root.add_child(section)
+
+    def remove_section(self, section):
+        if self.has_section(section):
```
neutrinoceros, line 148: I'm not sure why this check is necessary ? It seems that it could be making some errors silent for no good reason I can think of.

yt/config.py
```diff
@@ -66,74 +76,132 @@
...
+    def set(self, *args, metadata=None):
+        section, *keys, value = args
+        if metadata is None:
+            metadata = {"source": "runtime"}
+        self.config_root.upsert_from_list(
+            [section] + list(keys), value, extraData=metadata
+        )
+
+    def __setitem__(self, keys, value):
+        self.set(*keys, value, metadata=None)
```
neutrinoceros, line 160: wait. isn't the first arg in keys going to be read as "section" by `set` ?
... ok I think it's actually what you mean to do. I don't know how to suggest a better approach here but I'm slightly confused. This is likely acceptable but please double check if you can make this any clearer.

yt/config.py
```diff
@@ -66,74 +76,132 @@
...
-        self.get(key[0], key[1])
+        section, *keys = key
+        return self.get(section, *keys)
+
+    def remove(self, *args):
+        self.config_root.pop_leaf(args)
+
+    def read(self, file_names):
+        if not isinstance(file_names, (tuple, list)):
+            file_names = (file_names,)
```
neutrinoceros, line 171: you could use `more_itertools.always_iterable` here !
https://github.com/yt-project/yt/pull/2893 is likely to be merged shortly, please keep this in mind :)

yt/config.py
```diff
@@ -66,74 +76,132 @@
...
+        for fname in file_names:
+            if not os.path.exists(fname):
+                continue
+            metadata = {"source": f"file: {fname}"}
+            self.update(toml.load(fname), metadata=metadata)
+            file_names_read.append(fname)
+
+        return file_names_read
+
+    def write(self, fd):
```
neutrinoceros, line 183: I guess "fd" stands for "file dump" ? Though I only get this in the context of reviewing the PR. I think "fh" for "file handle" is more broadly used in the code base and still appropriate here ?

yt/config.py
```diff
@@ -66,74 +76,132 @@
...
+if os.path.exists(GLOBAL_CONFIG_FILE):
+    ytcfg.read(GLOBAL_CONFIG_FILE)
 
-# This should be implemented at some point.  The idea would be to have a set of
-# command line options, fed through an option parser, that would override
-# the settings in ytcfg.  *However*, because we want to have the command-line
-# scripts work, we'd probably want to have them only be long options, and also
-# along the lines of --yt-something=somethingelse.  The command line scripts
-# would then not get their options from sys.argv, but instead from this module.
+cwd = Path.cwd().resolve()
```
neutrinoceros, line 201: I don't think calling `.resolve()` has any effect here
```suggestion
cwd = Path.cwd()
```

yt/config.py
```diff
@@ -66,74 +76,132 @@
...
+for folder in chain([cwd], cwd.parents):
+    cfg_file = folder / "yt.toml"
+    if cfg_file.exists():
+        ytcfg.read(cfg_file)
+        LOCAL_CONFIG_FILE = str(cfg_file)
```
neutrinoceros, line 206: I have to say I find rather unsettling that we're using a constant module-wide variable (later to be imported elsewhere) for this value, which is extremely user-dependent. I may change my mind reading what's left.

yt/fields/tests/test_fields_plugins.py
```diff
@@ -18,8 +18,8 @@ def myfunc():
 
 
 def setUpModule():
-    my_plugin_name = ytcfg.get("yt", "pluginfilename")
-    # In the following order if pluginfilename is: an absolute path, located in
+    my_plugin_name = ytcfg.get("yt", "pluginFilename")
```
neutrinoceros, line 21: why was this case-change necessary ?

yt/geometry/geometry_handler.py
```diff
@@ -77,7 +77,7 @@ def _initialize_data_storage(self):
             writeable = os.access(dir_to_check, os.W_OK)
         else:
             writeable = os.access(fn, os.W_OK)
-        writeable = writeable and not ytcfg.getboolean("yt", "onlydeserialize")
+        writeable = writeable and not ytcfg.get("yt", "only_deserialize")
```
neutrinoceros, line 80: hum. I asked why you changed a parameter to camel case earlier, and here you're changing to snake case. I'm not fundamentally opposed to changing the case in parameters that have more than one word, but I would like it much better that the chosen case be consistent among them.

yt/mods.py
```diff
@@ -28,5 +28,5 @@
 # This way, other command-line tools can be used very simply.
 # Unfortunately, for now, I think the easiest and simplest way of doing
 # this is also the most dangerous way.
-if ytcfg.getboolean("yt", "loadfieldplugins"):
+if ytcfg.get("yt", "loadFieldPlugins"):
```
neutrinoceros, line 31: see previous questions

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -106,14 +106,14 @@ def enable_parallelism(suppress_logging=False, communicator=None):
...
+    ytcfg["yt", "internals", "global_parallel_rank"] = communicator.rank
+    ytcfg["yt", "internals", "global_parallel_size"] = communicator.size
+    ytcfg["yt", "internals", "parallel"] = True
     if exe_name == "embed_enzo" or ("_parallel" in dir(sys) and sys._parallel):
-        ytcfg["yt", "inline"] = "True"
+        ytcfg["yt", "inline"] = True
     if communicator.rank > 0:
-        if ytcfg.getboolean("yt", "LogFile"):
-            ytcfg["yt", "LogFile"] = "False"
+        if ytcfg.get("yt", "logFile"):
```
neutrinoceros, line 115: see above

yt/utilities/command_line.py
```diff
@@ -799,9 +799,14 @@ def __call__(self, args):
...
             raise YTCommandRequiresModule("requests") from e
-        if ytcfg.get("yt", "hub_api_key") != "":
-            print("You seem to already have an API key for the hub in")
-            print(f"{CURRENT_CONFIG_FILE} . Delete this if you want to force a")
+        hub_api_key, config_file = ytcfg.get(
+            "yt",
+            "hub_api_key",
+            callback=lambda leaf: (leaf.value, leaf.extraData.get("source", None)),
+        )
+        if hub_api_key != "":
```
neutrinoceros, line 807:
```suggestion
        if hub_api_key:
```

yt/utilities/command_line.py
```diff
@@ -799,9 +799,14 @@ def __call__(self, args):
...
+            print("You seem to already have an API key for the hub in ")
+            print(f"{config_file} . Delete this if you want to force a ")
             print("new user registration.")
```
neutrinoceros, line 810:
```suggestion
            print(
                "You seem to already have an API key for the hub in "
                f"{config_file} . Delete this if you want to force a "
                "new user registration."
            )
```

yt/utilities/command_line.py
```diff
@@ -1617,65 +1622,115 @@ def __call__(self, args):
...
+class YTConfigLocalConfigMixin:
+    def load_config(self, args):
+        from yt.config import GLOBAL_CONFIG_FILE, LOCAL_CONFIG_FILE
+        from yt.utilities.configure import CONFIG
+
+        if "local" in args and args.local:
+            if LOCAL_CONFIG_FILE is None:
+                LOCAL_CONFIG_FILE = os.path.join(
+                    os.path.abspath(os.path.curdir()), "yt.toml"
+                )
```
neutrinoceros, line 1634: I'm having mixed feelings here. Could you explain why this isn't/should be part of `LOCAL_CONFIG_FILE`'s definition in the module it originates from ? I'm trying to disentangle the different use cases but I can't figure this out in a satisfying way.

neutrinoceros (commented):
Still 2 files to go but here are my first comments. I'll finish this tomorrow !

neutrinoceros (commented):
some more minor comments. I'll approve when we converge on the global VS local config file discussion :-)

yt/config.py
```diff
@@ -66,74 +76,132 @@
...
+    def get_most_specific(self, section, *keys, **kwargs):
+        use_fallback = "fallback" in kwargs
+        fallback = kwargs.pop("fallback", None)
+        try:
+            return self.config_root.get_deepest_leaf(section, *keys)
+        except KeyError as e:
+            if use_fallback:
+                return fallback
+            else:
+                raise e
```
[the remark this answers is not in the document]
cphyc: I changed the `e` into `err` but not the `use_fallback`. Otherwise, the function will fail for missing keys with `fallback=None` but not with `fallback="something"`. This behavior is consistent with the behavior of `configparser.ConfigParser.get`.

yt/config.py
cphyc: Once again, this is merely for consistency with `configparser.ConfigParser.remove_section`. I noticed it returns True or False depending on whether the section was removed successfully, so updating accordingly.

yt/config.py
cphyc: Agreed, this could actually be stored in the config handler instead of globally.

yt/fields/tests/test_fields_plugins.py
cphyc: There were some places in yt where `pluginfilename` was camel-cased and others where it was not, so I decided to use the camel-case version. This was no issue with the old config, but has now become one.

yt/geometry/geometry_handler.py
cphyc: This is an unfortunate change indeed, I will revert it. Unfortunately, if we change the spelling of config entries, it will make the migration harder (unless we explicitly map from the old keys to the new one). If we vote for this, I would advocate in favor of using snake case everywhere.

yt/utilities/command_line.py
cphyc: I think this should go once we settle on whether we want to load both global and local config files or not. This could be moved into the YTConfig object.

yt/utilities/configuration_tree.py
```diff
@@ -0,0 +1,180 @@
...
+        key, *next_keys = keys
+        if len(next_keys) == 0:  # reach the end of the upsert
+            leaf = self.get_child(
+                key,
+                lambda: ConfigLeaf(key, parent=self, value=value, extraData=extraData),
+            )
+            leaf.value = value
+            leaf.extraData = extraData
+            if not isinstance(leaf, ConfigLeaf):
+                raise RuntimeError(f"Expected a ConfigLeaf, got {leaf}!")
```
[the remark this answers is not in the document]
cphyc: The `upsert_from_list` function is called when setting an item in the config, which can be called directly by the user using `ytcfg.set("yt", "foo", "bar", "baz")`.
With the config
```
[yt.test.foo]
bar = 42
```
a `RuntimeError` would be triggered by `ytcfg.set("yt", "test", "foo", "some_value")` because foo is not an option (it is a section).

yt/utilities/configuration_tree.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            next_node = self.get_child(key, lambda: ConfigNode(key, parent=self))
+            if not isinstance(next_node, ConfigNode):
+                raise RuntimeError(f"Expected a ConfigNode, got {next_node}!")
```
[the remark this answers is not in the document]
cphyc: This is an equivalent case as above (can be triggered by the user, so we don't want an assertion error to be raised).
This should however be caught downstream.

yt/utilities/configuration_tree.py
```diff
@@ -0,0 +1,180 @@
...
+                node_list.append(node)
+            except KeyError:
+                break
+
+        # For each node, starting from the deepest, try to find the leaf
+        for node in reversed(node_list):
+            try:
+                leaf = node.get_child(leaf_key)
+                if not isinstance(leaf, ConfigLeaf):
+                    raise RuntimeError(f"Expected a ConfigLeaf, got {leaf}!")
```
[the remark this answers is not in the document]
cphyc: Same here, can be triggered by the user so we do not want an AssertionError.

yt/utilities/configuration_tree.py
```diff
@@ -0,0 +1,180 @@
...
+    def update(self, other, extraData=None):
+        def _helper(d, keys):
+            for key, val in d.items():
+                new_keys = keys + [key]
+                if isinstance(val, dict):
+                    _helper(val, new_keys)
+                else:
+                    self.upsert_from_list(new_keys, val, extraData)
+
+        _helper(other, [])
```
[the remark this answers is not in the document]
cphyc: I have updated it, it should be more readable now.

doc/source/reference/command-line.rst
```diff
@@ -341,19 +341,18 @@ This will print the list of available subcommands:
...
+automatic migration of the old config, you are encouraged to run:
 
 .. code-block:: bash
 
    yt config migrate
 
-that will copy your current config file to the new location and store a backup
-copy as ``$HOME/.yt/config.bak``.
+that will convert your old config file to the toml format store a backup
+copy as ``$XDG_HOME_CONFIG/yt/ytrc.bak``.
```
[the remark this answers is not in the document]
cphyc: Updated!

yt/utilities/configuration_tree.py
```diff
@@ -0,0 +1,182 @@
+class ConfigNode:
```
cphyc, line 1: Before we merge this, I would like to add some docstrings.

doc/source/reference/command-line.rst
```diff
@@ -341,19 +341,18 @@ This will print the list of available subcommands:
...
 
-Since the yt version 3.3.2, the previous location of the configuration file
-(``$HOME/.yt/config``) has been deprecated in favor of a location adhering to the
+Since yt version 4, the configuration file is located in (``$XDG_HOME_CONFIG/yt/yt.toml``) adhering to the
 `XDG Base Directory Specification
 <https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html>`_.
-(``$XDG_HOME_CONFIG/yt/ytrc``). In order to perform an automatic migration of
-the old config, you are encouraged to run:
+The previous configuration file (``$XDG_HOME_CONFIG/yt/ytrc``) has been deprecated.. In order to perform an
+automatic migration of the old config, you are encouraged to run:
```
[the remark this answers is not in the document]
cphyc: Done!

doc/source/reference/configuration.rst
```diff
@@ -9,18 +9,20 @@ This customization is done through :ref:`configuration-file` and
...
 
-The Configuration File
-----------------------
+The Configuration
+-----------------
 
-The configuration is a simple text file setting internal yt variables to
-custom default values to be used in future sessions.
+The configuration is stored in simple text files (in the `toml <https://github.com/toml-lang/toml>`_ format).
+They allow to set internal yt variables to custom default values to be used in future sessions.
```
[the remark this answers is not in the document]
cphyc: I rephrased it, should be better now.

yt/utilities/configuration_tree.py
```diff
@@ -0,0 +1,180 @@
...
+    def __init__(self, key, parent=None):
+        self.key = key
+        self.children = {}
+        self.parent = parent
+
+    def add(self, key, child):
+        self.children[key] = child
+        child.parent = self
+
+    def update(self, other, extraData=None):
```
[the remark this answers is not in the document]
cphyc: I changed it here and everywhere else (I hope...)

Xarthisius:
@yt-fido test this please

yt/config.py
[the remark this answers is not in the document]
neutrinoceros: fair enough !

doc/source/reference/configuration.rst
```diff
@@ -43,6 +45,29 @@ options from the configuration file, e.g.:
    $ yt config rm yt maximumstoreddatasets
 
 
+Local Configuration
+^^^^^^^^^^^^^^^^^^^
+
+yt will look for any file name ``yt.toml`` in the current directory or any
+of its parent, up to the root. If present, its options override those in the
+global configuration.
```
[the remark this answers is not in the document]
cphyc: Indeed!

yt/fields/tests/test_fields_plugins.py
neutrinoceros: Ok I get it, thanks !

yt/utilities/configuration_tree.py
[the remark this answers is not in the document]
neutrinoceros: wait, have you ?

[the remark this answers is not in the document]
cphyc: ⇒ https://github.com/yt-project/yt/pull/2981/files#diff-6743a387d29ad447580a150acf1938bce5bee82249fb401fcb3ae361ea2c79beR20

yt/utilities/configuration_tree.py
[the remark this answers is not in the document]
neutrinoceros: Either your explanation has inconsistency or I'm not following :/
Shouldn't "baz" be the value instead of `42`?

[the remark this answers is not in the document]
cphyc: I can't understand my own initial comment, so let's try again. Essentially, these functions are exposed almost directly to the user (through the `YTConfig` object, which just queries the config tree), so I don't want to have it throw assertion errors, and rather something slightly more useful (though not _very_ useful).

yt/config.py
```diff
@@ -196,18 +165,52 @@ def write(self, file_handler):
...
+        raise SystemExit(msg)
+
+
+if not os.path.exists(_global_config_file):
+    cfg = {"yt": {}}
+    try:
+        with open(_global_config_file, mode="w") as fd:
+            toml.dump(cfg, fd)
+    except OSError:
+        warnings.warn("unable to write new config file")
```
neutrinoceros, line 205: Hum, I don't really like this. I assume that this block was initially written to ease development, but I don't see a good reason to keep it. This is not directly related to this PR however.

yt/config.py
```diff
@@ -196,18 +165,52 @@ def write(self, file_handler):
...
-for folder in chain([cwd], cwd.parents):
-    cfg_file = folder / "yt.toml"
-    if cfg_file.exists():
-        ytcfg.read(cfg_file)
-        LOCAL_CONFIG_FILE = str(cfg_file)
-        break
+# Try loading the local config first, otherwise fall back to global config
+if os.path.exists(_local_config_file):
+    ytcfg.read(_local_config_file)
+elif os.path.exists(_global_config_file):
```
neutrinoceros, line 215: note that this currently is equivalent to an `else`, because of the afore-commented block. But an `elif` is better imo since it won't break even if we support _not_ having a config file at all in the future.

yt/utilities/command_line.py
```diff
@@ -71,6 +72,15 @@ def _fix_ds(arg, *args, **kwargs):
 def _add_arg(sc, arg):
     if isinstance(arg, str):
         arg = _common_options[arg].copy()
```
neutrinoceros, line 74: `_common_options` looks potentially undefined. I don't get where it's supposed to come from

yt/utilities/command_line.py
```diff
@@ -1626,37 +1636,78 @@ def __call__(self, args):
...
-            config_file = LOCAL_CONFIG_FILE
+        args = vars(args)
+
+        local_config_file = YTConfig.get_local_config_file()
+        global_config_file = YTConfig.get_global_config_file()
+
+        local_exists = os.path.exists(local_config_file)
+        global_exists = os.path.exists(global_config_file)
+
+        if "local" in args and args["local"]:
```
neutrinoceros, line 1652:
```suggestion
        if args.get("local", False):
```

yt/utilities/command_line.py
```diff
@@ -1626,37 +1636,78 @@ def __call__(self, args):
...
+            config_file = local_config_file
+        elif "global" in args and args["global"]:
```
neutrinoceros, line 1654:
```suggestion
        elif args.get("global", False):
```

yt/config.py
cphyc: The issue is that if `~/.config/yt/yt.toml` is not writeable, you cannot use yt at all...

doc/source/reference/changelog.rst
```diff
@@ -191,9 +191,9 @@ Minor Enhancements and Bugfixes
   fixed by `PR 2200 <https://github.com/yt-project/yt/pull/2200>`__.
 - fix an issue related to creating a ds9 region with 
   FITS `PR 2335 <https://github.com/yt-project/yt/pull/2335>`__.
-- add a check to see if pluginfilename is specified in 
+- add a check to see if plugin_filename is specified in
```
neutrinoceros, line 194:
```suggestion
- add a check to see if ``plugin_filename`` is specified in
```

yt/utilities/command_line.py
cphyc: It is defined weeeeeell below. I am ok to move it, but maybe as part of another PR?

doc/source/reference/command-line.rst
```diff
@@ -341,19 +341,20 @@ This will print the list of available subcommands:
 
 .. config_help:: yt config
 
-Since the yt version 3.3.2, the previous location of the configuration file
-(``$HOME/.yt/config``) has been deprecated in favor of a location adhering to the
+
+Since yt version 4, the configuration file is located in (``$XDG_HOME_CONFIG/yt/yt.toml``) adhering to the
```
neutrinoceros, line 345: wait, did you really change the name of the directory as well ? If so, what's the rationale ? I think the common practice for this is indeed to have `.` prefix for such dirs !

doc/source/reference/command-line.rst
```diff
@@ -341,19 +341,20 @@ This will print the list of available subcommands:
...
 `XDG Base Directory Specification
 <https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html>`_.
-(``$XDG_HOME_CONFIG/yt/ytrc``). In order to perform an automatic migration of
-the old config, you are encouraged to run:
+The old configuration file (``$XDG_HOME_CONFIG/yt/ytrc``) is deprecated.
```
neutrinoceros, line 348: same as above (though this typo was already present, so it seems)

doc/source/reference/command-line.rst
```diff
@@ -341,19 +341,20 @@ This will print the list of available subcommands:
...
+In order to perform an automatic migration of the old config, you are
+encouraged to run:
 
 .. code-block:: bash
 
    yt config migrate
 
-that will copy your current config file to the new location and store a backup
-copy as ``$HOME/.yt/config.bak``.
+that will convert your old config file to the toml format. The original file
```
neutrinoceros, line 356:
```suggestion
this will convert your old config file to the toml format. The original file
```

doc/source/reference/command-line.rst
cphyc: In yt 2, the config used to be in `$HOME/.yt/config`.
Since yt 3, the config is in `$XDG_CONFIG_HOME/yt == $HOME/.config/yt`.

This doc update merely reflects the fact the migration will happen from yt 3 to 4 rather than yt 2 to 3.

doc/source/reference/configuration.rst
```diff
@@ -9,39 +9,63 @@ This customization is done through :ref:`configuration-file` and
...
-   $ yt config set yt loglevel 1
-   $ yt config rm yt maximumstoreddatasets
+   $ yt config set yt log_level 1
+   $ yt config rm yt maximum_stored_datasets
+
+
+Local Configuration
+^^^^^^^^^^^^^^^^^^^
+
+yt will look for a file name ``yt.toml`` in the current directory. If present, its options
```
neutrinoceros, line 51:
```suggestion
yt will look for a file named ``yt.toml`` in the current directory. If present, its options
```

doc/source/reference/configuration.rst
```diff
@@ -9,39 +9,63 @@ This customization is done through :ref:`configuration-file` and
...
+.. code-block:: none
+
+   $ yt config -h
+   $ yt config list --local
+   $ yt config set --local yt log_level 1
+   $ yt config rm --local yt maximum_stored_datasets
+   $ yt config print-path --local
+
+If no local configuration file is present, these commands will create an (empty) one
+in the current working directly.
```
neutrinoceros, line 68:
```suggestion
in the current working directory.
```

yt/__init__.py
```diff
@@ -184,9 +184,9 @@ def _check_deprecated_parameters():
     from yt.config import ytcfg
     from yt.funcs import issue_deprecation_warning
 
-    if ytcfg.getboolean("yt", "loadfieldplugins"):
+    if ytcfg.get("yt", "load_field_plugins"):
         issue_deprecation_warning(
-            "Found deprecated parameter 'loadfieldplugins' parameter in yt rcfile."
+            "Found deprecated parameter 'load_field_plugins' parameter in yt rcfile."
```
neutrinoceros, line 189:
```suggestion
            "Found deprecated parameter 'load_field_plugins' in yt rcfile."
```

yt/funcs.py
```diff
@@ -1088,14 +1088,14 @@ def _func(*args, **kwargs):
...
-def enable_plugins(pluginfilename=None):
+def enable_plugins(plugin_filename=None):
     """Forces a plugin file to be parsed.
 
     A plugin file is a means of creating custom fields, quantities,
     data objects, colormaps, and other code classes and objects to be used
     in yt scripts without modifying the yt source directly.
 
-    If <pluginfilename> is omited, this function will look for a plugin file at
+    If <plugin_filename> is omited, this function will look for a plugin file at
```
neutrinoceros, line 1098:
```suggestion
    If `plugin_filename` is omitted, this function will look for a plugin file at
```

doc/source/reference/changelog.rst
```diff
@@ -191,9 +191,9 @@ Minor Enhancements and Bugfixes
   fixed by `PR 2200 <https://github.com/yt-project/yt/pull/2200>`__.
 - fix an issue related to creating a ds9 region with 
   FITS `PR 2335 <https://github.com/yt-project/yt/pull/2335>`__.
-- add a check to see if pluginfilename is specified in 
+- add a check to see if ``plugin_filename`` is specified in
```
neutrinoceros, line 194: Just realised I'm not confortable with retroactively changing the name of the parameter in release notes
```suggestion
- add a check to see if ``pluginfilename`` is specified in
```

yt/utilities/command_line.py
neutrinoceros: sure

yt/utilities/command_line.py
```diff
@@ -1617,65 +1634,144 @@ def __call__(self, args):
         print(r.text)
 
 
-class YTConfigGetCmd(YTCommand):
+class YTConfigLocalConfigMixin:
```
neutrinoceros, line 1637: I'm slightly disturbed that this class has "Mixin" in its name but doesn't use inheritance. I don't understand the naming.

yt/utilities/command_line.py
```diff
@@ -1617,65 +1634,144 @@ def __call__(self, args):
...
+                s = (
+                    "Yt detected a local and a global configuration file, refusing "
+                    "to proceed.\n"
+                )
+                s += f"Local config file: {local_config_file}\n"
+                s += f"Global config file: {global_config_file}\n"
+                s += (
+                    "Specify which one you want to use using the `--local` or the "
+                    "`--global` flags."
+                )
```
neutrinoceros, line 1665:
```suggestion
                s = (
                    "Yt detected a local and a global configuration file, refusing "
                    "to proceed.\n"
                    f"Local config file: {local_config_file}\n"
                    f"Global config file: {global_config_file}\n"
                    "Specify which one you want to use using the `--local` or the "
                    "`--global` flags."
                )
```

yt/utilities/command_line.py
```diff
@@ -1617,65 +1634,144 @@ def __call__(self, args):
...
-        dict(short="section", help="The section containing the option."),
-        dict(short="option", help="The option to set."),
-        dict(short="value", help="The value to set the option to."),
+    args = tuple(
+        [
+            dict(short="section", help="The section containing the option."),
+            dict(short="option", help="The option to set."),
+            dict(short="value", help="The value to set the option to."),
+        ]
+        + _global_local_args
```
neutrinoceros, line 1729: Maybe a matter of personal tastes
```suggestion
        dict(short="section", help="The section containing the option."),
        dict(short="option", help="The option to set."),
        dict(short="value", help="The value to set the option to."),
        *_global_local_args
```

yt/utilities/command_line.py
```diff
@@ -1617,65 +1634,144 @@ def __call__(self, args):
...
     description = "remove a config option"
-    args = (
-        dict(short="section", help="The section containing the option."),
-        dict(short="option", help="The option to remove."),
+    args = tuple(
+        [
+            dict(short="section", help="The section containing the option."),
+            dict(short="option", help="The option to remove."),
+        ]
+        + _global_local_args
```
neutrinoceros, line 1749:
```suggestion
        dict(short="section", help="The section containing the option."),
        dict(short="option", help="The option to remove."),
        *_global_local_args
```

yt/utilities/command_line.py
cphyc: I changed it to `Handler`.

yt/utilities/configuration_tree.py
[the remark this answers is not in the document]
neutrinoceros: ok, makes a lot of sense. I didn't realise this code was so close to the surface.

yt/config.py
neutrinoceros: I get that this is currently the case. My point is that it shouldn't :)

cphyc: The thing is that yt's configuration is optional, so I don't think we should force the user to have one, should we?

neutrinoceros: I just realised that I misread the code when I first commented. In the end we agree, please disregard.

cphyc:
> I have a general question here: is it still relevant to store the global config file in `$HOME/.config/yt/yt.toml` rather than `$HOME/.config/yt.toml` ? what else is expected to be there ?

I am unsure I replied to this, but the reason we're having it all in a yt folder is so that we can store the plugin file in the same folder.

cphyc:
@yt-fido test this please.

neutrinoceros (approved):

yt/config.py
```diff
@@ -66,74 +75,154 @@
...
+        self.config_root.pop_leaf(args)
 
-    def get(self, section, option, *args, **kwargs):
-        val = super(YTConfigParser, self).get(section, option, *args, **kwargs)
-        return os.path.expanduser(os.path.expandvars(val))
+    def read(self, file_names):
+        file_names = always_iterable(file_names)
 
+        file_names_read = []
+        for fname in file_names:
```
neutrinoceros, line 147:
```suggestion
        file_names_read = []
        for fname in always_iterable(file_names):
```

matthewturk (approved):
I think we're good!

neutrinoceros:
@munkm Do you want to review this ?

munkm:
I think it's good to merge with your and Matt's approvals. I don't want to block this from being merged! 🙂 I'll do a post-merge review and ask questions if I have any.

munkm:
However, if @cphyc would like my review pre-merge I'm happy to do one (today or Monday)!

cphyc:
> However, if @cphyc would like my review pre-merge I'm happy to do one (today or Monday)!

I would sleep tighter if I knew a third pair of eyes reviewed it! There's no rush on my side though, but I would be keen to have feedback from us developers before releasing the new config system to the public.

neutrinoceros:
Note that I'd like to coordinate with @cphyc on this and https://github.com/yt-project/yt/pull/3007 to see what should be done about the deprecation warnings related to config file formats.

cphyc:
@munkm any chance you review this? Otherwise, I'm fine with it being merged in.

cphyc:
The Jenkins failure are due to the fact Jenkins does not install the requirements listed in `setup.py`, which now include `toml` as a dependency. This should probably be fixed on Jenkins' side rather than in this PR (by installing those requirements).

matthewturk:
@Xarthisius Can that be fixed without your intervention?

I think otherwise this may be ready to go, once the tests pass.

neutrinoceros:
Following the merging of #2981 , I need to patch this with the new deprecation utilities. Then I believe it will finally be ready for going in :)

brittonsmith:
I'll merge this when the tests finish.

PR #2983, opened 2020-11-25
Flynt+flake+isort doc

cphyc:
## PR Summary

Apply fint to the doc and fix a couple of isort/flake8 issues there.
On a side note, there are many files that depend on `hg` which we haven't been using for years now. Should we clean them up?


neutrinoceros:
Thanks for doing this ! Could you also add `doc` to the targeted locations in the precommit hooks and gh workflow checks ?

neutrinoceros:
adding the dead code label because of 3171ebc733b98256e95f9f8801a8d100854a1b6a

neutrinoceros (approved):
Thank you for doing this, I carefully went over it and I approve fully.

neutrinoceros:
> On a side note, there are many files that depend on hg which we haven't been using for years now. Should we clean them up?

I think @matthewturk could answer this question !

matthewturk (approved):

matthewturk:
I'm OK with removing the hg-related files and support tools, but maybe in a separate PR.

PR #2985, opened 2020-12-01
Optimize Athena++ i/o

Xarthisius:
* Read coordinates in one op instead of for loop over grids
* Do not override default _chunk_io to properly chunk grid reads

The main issue was a custom `_chunk_io` that was for some reason returning chunks of size 1, which in turn resulted in very costly reads of 16^3 floats for each grid. Let's see if the default works fine and tests pass.


Xarthisius:
Looks like the tests are passing and their execution time decreased from ~20 minutes to 4 min 33 sec.

jzuhone (approved):

matthewturk (approved):

PR #2986, opened 2020-12-02
[bugfix] save_as_dataset with ArbitraryGrid

brittonsmith:
## PR Summary

As reported on slack, if one creates an `ArbitraryGrid` from a particle dataset and then saves it with `save_as_dataset`, the data is not saved with the correct field type. Specifically, since all fields are deposited onto the `ArbitraryGrid`, they are always of grid type there even if they were originally of particle type. An example of this can be found [here](http://paste.yt-project.org/show/302/).

This fix represents the simplest solution to the problem, i.e., marking the `ArbitraryGrid` class as always using grid-type fields. I'm open to suggestions on whether this is the right approach or if I should use a different attribute name.

## PR Checklist

- [ ] pass `flake8 yt/`

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -1174,6 +1174,7 @@ class YTArbitraryGrid(YTCoveringGrid):
     """
 
     _spatial = True
+    _always_grid_fields = True
     _type_name = "arbitrary_grid"
     _con_args = ("left_edge", "right_edge", "ActiveDimensions")
     _container_fields = (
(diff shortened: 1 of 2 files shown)


brittonsmith:
Huh, well, this won't work. The failing test creates an `ArbitraryGrid` from an Enzo dataset and in that case, the particle fields are not deposited onto the grid, but seem to be the particles contained within that grid. So, there are two different particle-type fields that operate differently within `ArbitraryGrids`. Does anyone have an idea of how to distinguish them?

jzuhone:
@brittonsmith where does this stand?

brittonsmith:
@jzuhone I believe this is still stuck because there is no way to distinguish the two different types of behaviors of particle fields within ArbitraryGrids based on the field itself. I think this is stuck on the same issue that PR #2004 is. Any help is very welcome.

PR #2990, opened 2020-12-07
maint: drop mock as a third party dependency 

neutrinoceros:
Rationale:
`mock` was standardized in Python 3.3


munkm (approved):
Thank you! I think this is a good plan.

matthewturk (approved):

PR #2991, opened 2020-12-07
doc: add minimal structure to SkeletonDataset class

neutrinoceros:
## PR Summary

- Init required instance attributes in order to leave a valid Dataset structure.
- minor doc update (hopefully making the instructions a little clearer)

edit:
the changes to `SkeletonDataset` make it so the following code pass without error
```python
yt.load("myfile.from.a.new.format")
```
where the only changes required from the author of a new frontend are
- adding the frontend to `yt/frontends/api.py`
- implementing `NewDataset._is_valid()`

### DIFF
--- a/yt/frontends/_skeleton/data_structures.py
+++ b/yt/frontends/_skeleton/data_structures.py
@@ -148,7 +148,8 @@ class SkeletonDataset(Dataset):
         #   self.geometry  <= a lower case string
         #                     ("cartesian", "polar", "cylindrical"...)
         #                     (defaults to 'cartesian')
-        pass
+        self.current_time = -1  # required, change this !
+        self.cosmological_simulation = 0  # required. Change this if need be.
 
     @classmethod
     def _is_valid(self, *args, **kwargs):
--- a/yt/frontends/_skeleton/data_structures.py
+++ b/yt/frontends/_skeleton/data_structures.py
@@ -148,8 +148,14 @@ class SkeletonDataset(Dataset):
         #   self.geometry  <= a lower case string
         #                     ("cartesian", "polar", "cylindrical"...)
         #                     (defaults to 'cartesian')
-        self.current_time = -1  # required, change this !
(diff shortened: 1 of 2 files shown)


matthewturk (commented):
Nice!!  Looking forward to review, but this looks good so far.

neutrinoceros:
Actually I was expecting to do more with this as I went along implementing my new frontend (PR still not open), but I didn't need to. This is now ready for review.

yt/frontends/_skeleton/data_structures.py
```diff
@@ -143,7 +148,8 @@ def _parse_parameter_file(self):
         #   self.geometry  <= a lower case string
         #                     ("cartesian", "polar", "cylindrical"...)
         #                     (defaults to 'cartesian')
-        pass
+        self.current_time = -1  # required, change this !
```
munkm, line 151: It might be worth suggesting what a good value would be for this in a code that doesn't have time dependence. Since it is required, a helpful suggestion might be good here.

munkm (approved):
One minor comment but everything looks good to me.

yt/frontends/_skeleton/data_structures.py
neutrinoceros: Good point. This didn't occur to me. Actually I don't know why it's required at all. I should look for the reason first.
Do you think ˋ0.0` would make sense here ?

matthewturk:
This seems fine to me.  My suggestion would be a time of zero for default, but setting to -1 makes it so that they explicitly have to set it if they don't want some weird value.

neutrinoceros:
So I'm down with whatever default value you deem most sensible @munkm , though I don't know what would be a helpful suggestion here :/

munkm:
I think leaving it as -1 in the code but having an inline comment that we suggest 0 would work.

neutrinoceros:
done !

PR #2995, opened 2020-12-15
Splitting selection_routines into subfiles

matthewturk:
## PR Summary

This is an experiment to try to reduce the complexity and file size of the `selection_routines.pyx` file.  It doesn't make any changes to it other than to split each selector (and the set of boolean selectors as a group) into their own files.

This should make implementing a new selector a bit easier, as it can be templated, and it should *also* make it much easier if we want to apply templates via `DEF` and multiple `include` calls.

## PR Checklist

- [x] pass `flake8 yt/`


yt/geometry/_selection_routines/disk_selector.pxi
```diff
@@ -0,0 +1,169 @@
...
+        #                 H += (temp * self.norm_vec[n])
+        #                 D += temp*temp
+        #             R2 = (D - H*H)
+        #             if R2 < self.radius2 :
+        #                 any_radius = 1
+        #                 if fabs(H) < self.height: return 1
+        #             if H < 0: all_over = 0
+        #             if H > 0: all_under = 0
+        # if all_over == 0 and all_under == 0 and any_radius == 1: return 1
+        # return 0
```
neutrinoceros, line 101: any reason we'd want to keep this commented code around ?

neutrinoceros (approved):
Left a question but I trust that you did nothing else than cut & paste (+ `#include`), in which case fire away !

yt/geometry/_selection_routines/disk_selector.pxi
matthewturk: I'll try to remove in a future PR!

Issue #1880, opened 2018-07-06
Remove `refine_by` as a parameter on the `Dataset` object


MatthewTurk:
Having looked through the code base, I believe getting rid of `refine_by` as a required attribute on the `Dataset` object is feasible.  This would align with more flexible systems of loading and representing data.

neutrinoceros:
sounds like something that'd be actually required to support rectilinear grids with arbitrary spacing (something I'm still interested to do and use). Matt, are you still on board with this idea ?

matthewturk:
I think I am.

PR #2998, opened 2020-12-15
ENH: Enable variable cell widths for grid datasets

matthewturk:
## PR Summary

I'm starting the process of adding variable `dds` attributes to grid objects, so that we can potentially support arbitrarily spaced grids without requiring that we use the semi structured mesh work.

I'd like this to be in a constant state of workability, so it won't show up as "mergeable" but I do hope the GHA tests run.

## PR Checklist

- [ ] pass `flake8 yt/`


yt/geometry/selection_routines.pxd
```diff
@@ -53,12 +53,13 @@ cdef class SelectorObject:
...
-                                np.ndarray[np.uint8_t, ndim=3, cast=True] child_mask,
-                                np.ndarray[np.uint8_t, ndim=3] mask,
-                                int level)
+    cdef int fill_mask_selector_regular_grid(self, np.float64_t left_edge[3],
+                                             np.float64_t right_edge[3],
+                                             np.float64_t dds[3], int dim[3],
+                                             np.ndarray[np.uint8_t, ndim=3,
+                                                        cast=True] child_mask,
+                                             np.ndarray[np.uint8_t, ndim=3]
+                                             mask, int level)
```
neutrinoceros, line 62:
```suggestion
                                             np.ndarray[np.uint8_t, ndim=3, cast=True] child_mask,
                                             np.ndarray[np.uint8_t, ndim=3] mask,
                                             int level)
```

neutrinoceros (dismissed):
Looks fine to me, but I figure we should document what "regular" means precisely in this context. No need to do it for this PR though.

yt/geometry/selection_routines.pxd
matthewturk: Where did we end up with black-on-Cython?

neutrinoceros:
There's a minor conflict to solve here.

matthewturk:
So with this I have introduced a couple things that need to happen; it is likely not going to be ready for 4.0 (but honestly who knows) and I think what I need to do is collect my thoughts on it.  I promised @neutrinoceros I would do this some time ago, and then I didn't, and I am going to hold myself to my word.

neutrinoceros:
For the record: I care very much about this feature but I certainly don't want it to delay the release. Thanks for getting back at me Matt, I appreciate it :)

neutrinoceros:
I see you started pushing again here, so I'm requesting a new review from myself, which effectively suppresses my previous approval :)

matthewturk:
I'm converting this to draft until after 4.0 goes out so that it won't show up in my filters.

neutrinoceros:
@matt what's the status for this now ?

matthewturk:
It'll need rebasing when I get back to it.  I am eternally optimistic I will be able to do something on it in the near future.  The realism of that is uncertain to me.

neutrinoceros:
Ok here's a different approach : do we _need_ this PR to make further progress on stretched grids ?

yt/geometry/_selection_routines/selector_object.pxi
```diff
@@ -404,6 +405,100 @@ cdef class SelectorObject:
...
+        cdef int i, j, k
+        cdef int total = 0, this_level = 0
+        cdef np.float64_t pos[3]
+        if level < self.min_level or level > self.max_level:
+            return 0
+        if level == self.max_level:
+            this_level = 1
+        cdef np.float64_t *offsets[3]
+        cdef np.float64_t c = 0.0
+        cdef np.float64_t tdds[3]
```
neutrinoceros, line 472: what does `tdds` stands for ? (I'm already not sure I understand what `dds` means, though I think I get what it's used for)

yt/geometry/_selection_routines/selector_object.pxi
```diff
@@ -404,6 +405,100 @@ cdef class SelectorObject:
...
+            c = left_edge[i]
+            for j in range(dim[i]):
+                offsets[i][j] = c
+                c += dds[i][j]
+        with nogil:
+            # We need to keep in mind that it is entirely possible to
+            # accumulate round-off error by doing lots of additions, etc.
+            # That's one of the reasons we construct (ahead of time) the edge
+            # array. I mean, we don't necessarily *have* to do that, but it
+            # seems OK.
```
neutrinoceros, line 484: Not clear what the "edge array" refers to in the context.
Also, I don't understand the chronology: does "we construct the edge array" refer to what follows, or what happens before that ?

neutrinoceros (commented):
So I went as far as convincing myself we absolutely need this patch, and made some progress in understanding what's going on in this (very unfamiliar to me) part of the code.
Here are some questions to help me better understand next time.

I think I said this in an earlier review too, but so far my biggest and probably only concern is with the vague naming scheme (regular VS irregular grid). I really think we ought to do better on this point, and that clear definitions of the terms we end up using.

All this considered, this looks to be in a mergeable state now, so is there anything missing besides a thorough review ?

neutrinoceros:
I'll refresh the CI validation status here with a quick close + reopen

yt/geometry/_selection_routines/selector_object.pxi
```diff
@@ -404,6 +405,100 @@ cdef class SelectorObject:
...
+        free(dds)
+        if total == 0: return None
+        return mask.astype("bool")
+
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    @cython.cdivision(True)
+    cdef int fill_mask_selector(self, np.float64_t left_edge[3],
+                                np.float64_t right_edge[3],
+                                np.float64_t **dds, int dim[3],
```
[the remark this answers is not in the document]
cphyc: This is a pointer to a pointer (i.e. a 2D array)

yt/geometry/_selection_routines/selector_object.pxi
```diff
@@ -404,6 +405,100 @@ cdef class SelectorObject:
...
+                                np.ndarray[np.uint8_t, ndim=3] mask,
+                                int level):
+        cdef int i, j, k
+        cdef int total = 0, this_level = 0
+        cdef np.float64_t pos[3]
+        if level < self.min_level or level > self.max_level:
+            return 0
+        if level == self.max_level:
+            this_level = 1
+        cdef np.float64_t *offsets[3]
```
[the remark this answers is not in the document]
cphyc: offsets here represent 3 pointers to undefinite-sized arrays (see line 474 for their allocation).

yt/geometry/_selection_routines/selector_object.pxi
cphyc: I would guess it is `temporary dds`?

n-claes:
Any news yet on the approximate status of this PR? If this is merged at some point, how far is yt then from fully supporting stretched grids? We're currently facing an increased demand for yt visualisations of MPI-AMRVAC stretched grid datasets (and by extension also datasets from EUHFORIA, which uses the same datastructure), both for Cartesian and cylindrical/spherical coordinate systems.

When it becomes possible to do this in yt I'd like to get started on adding support (at least for the amrvac frontend), so I'm gonna keep an eye on this

yt/geometry/_selection_routines/selector_object.pxi
matthewturk: yup, `temporary dds`!

yt/geometry/_selection_routines/selector_object.pxi
matthewturk: In this case, "edge array" refers to the cell edges -- rather than the cell widths, inside this function, we are most interested in the edges, so that we can take the edge, offset halfway to the next cell, and then utilize the cell width for our `select_cell` call.

matthewturk:
@n-claes Sorry for the delay.  I have taken this up again, and I am more optimistic about a near-term solution than I was.  The latest set of changes get us much closer to having this work in a semi-general case without impacting performance for "classic" non-stretched grids.

What needs to be done:

 * Uniformly utilize cell widths in the stream frontend
 * Finish implementation and testing of `fcoords`, `fwidth`, `icoords` and `ires`.  `get_global_startindex` is also necessary.

I am, however, optimistic this is do-able in the medium term.

matthewturk:
I believe that the next task will be modifying projections to manage the fwidths correctly (which, in retrospect, they may actually not need much adjustment for) and to ensure that our global startindex is correctly set.

matthewturk:
Something that has come up now is the appropriate way of mapping from the tuple `(icoord_x, icoord_y, icoord_z, ires)` to `(fcoord_x, fcoord_y, fcoord_z, fwidth_x, fwidth_y, fwidth_z)`.  For *most* AMR dataset this is reasonably straightforward, in that this mapping is combined through `refine_by` and `domain_width` to get back a set of values.  The key thing to understand in yt is whether or not we can rely on all `(icoord_x, icoord_y, ires)` mapping to the same `(fcoord_x, fcoord_y, fwidth_x, fwidth_y)` or whether some overlapping-along-z grids would have different mappings.

I see two possible cases:

1. There's a fixed function that maps -- this would be the case for, say, logarithmic spacings with AMR included.  This could be managed by having a function defined that maps the coordinates!
2. There are grid values that map to these, and we store them in the root grid or root index.  Then we would iterate until we found a grid that contains our `icoord` values and we grab its `cell_width` values to map.

@neutrinoceros I'm pinging you because you have the nearest-term use case for this.  Any suggestions?

neutrinoceros:
So, first of, I do not know of a data format where stretched grids _and_ overlaps can coexist.

Now for the formats I do know:
Idefix/Pluto data is monogrid and contains the cell edges. It's also not very practical to reconstruct the grid independently because some scheme are actually quite tricky to implement in the simulation codes themselves, so it could easily lead to subtle mistakes, and break backward or forward compatibility.
So in my use case, reading exact coordinates from disk is both the easiest thing to do as well as the most robust option.

AMRVAC is more tricky because IIRC, coordinates aren't written to disk, so the only way to go is functional. 

Can we add a class method (in Grid objects ?) to be overwritten by frontends that would hide this complexity and allow for both implementations flavors ?

matthewturk:
@neutrinoceros I think if we're talking about a *global* `icoords, ires` to floating point stuff (which I think from your comment we *are*) it would make the most  sense to put it on the index or the dataset, much like we do for the refinement factor (which this would utilize).  I am going to take a shot at that.

matthewturk:
Projections now work with `stream` data, although I would like to see better functional forms implemented.  I'm working on the documentation, and then we might be able to call this feature "preliminary."

doc/source/examining/loading_data.rst
```diff
@@ -164,8 +164,9 @@ Appropriate errors are thrown for other combinations.
 * particle data: currently not supported (but might come later)
 * staggered grids (AMRVAC 2.2 and later): yt logs a warning if you load
   staggered datasets, but the flag is currently ignored.
-* "stretched grids" as defined in AMRVAC have no correspondence in yt,
-  hence will never be supported.
+* "stretched grids" are being implemented in yt, but are not yet
+  fully-supported.  (Previous versions of this file suggested they would
+  "never" be supported, which we hope to prove incorrect.)
```
neutrinoceros, line 169: oopsy. Nice catch :)

doc/source/examining/loading_data.rst
```diff
@@ -1961,10 +1962,25 @@ have assumed your data is stored in the three-dimensional array
 Stretched Grid Data
 -------------------
 
+.. warning::
+
+   API consistency for loading stretched grids is not guaranteed until yt 4.2!
```
neutrinoceros, line 1967: I'd suggest not making such version-specific promises. It'd be a shame to hold 4.2 just because of an issue we haven't anticipated yet.

doc/source/examining/loading_data.rst
```diff
@@ -1964,22 +1964,54 @@ Stretched Grid Data
...
    complete and flexible support!
 
 To load a stretched grid, you use the standard (and now rather-poorly named)
-``load_uniform_grid`` function, but supplying a ``cell_widths`` argument.  
+``load_uniform_grid`` function, but supplying a ``cell_widths`` argument.  This
+argument should be a list of three arrays, corresponding to the first, second
+and third index-direction cell widths.  (For instance, in a "standard"
+cartesian dataset, this would be x, y, z.)
+
+This script, a variation of which is used for testing this functionality,
```
neutrinoceros, line 1987: Fun fact, but maybe not super useful info for users ?
```suggestion
This script,
```

doc/source/examining/loading_data.rst
```diff
@@ -1964,22 +1964,54 @@ Stretched Grid Data
...
+
+   N = 8
+
+   data = {"density": np.random.random((N, N, N))}
+
+   cell_widths = []
+   for i in range(3):
+       _ = np.random.random(N)
+       _ /= _.sum()
+       cell_widths.append(_)
```
neutrinoceros, line 2003: Maybe we should avoid using `_` as a variable name in docs.

matthewturk:
OK, I'm getting kind of confused about the pre-commit error.  I change it manually, and it changes it back!  So what do I do?!

yt/geometry/grid_geometry_handler.py
```diff
@@ -450,6 +450,25 @@ def _chunk_io(
                 with self.io.preload(dc, preload_fields, 4.0 * size):
                     yield dc
 
+    def _icoords_to_fcoords(self, icoords, ires, axes=None):
+        """
+        Accepts icoords and ires and returns appropriate fcoords and fwidth.
+        Mostly useful for cases where we have irregularly spaced or structured
+        grids.
+        """
+        dds = self.ds.domain_width[axes,] / (
```
neutrinoceros, line 459: I think this is what flake8 complains about
```suggestion
        dds = self.ds.domain_width[axes] / (
```

matthewturk: Hm, I think at the time I wrote it, I *thought* I was changing things by using the comma, but I suppose not.  Once the tests are done running I'll do this, so that we can at least see if the current run will pass.

neutrinoceros: I convinced myself they are perfectly equivalent:
```python
>>> import numpy as np
>>> a = np.arange(10)
>>> a[0]
0
>>> a[0,]
0
>>> a[0, ...]
array(0)
>>> a[0, :]
...
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

neutrinoceros (commented):
Ok here are a couple questions
More importantly, I see a handful of tests not passing yet. I think it should be prioritised.

yt/data_objects/index_subobjects/stretched_grid.py
```diff
@@ -0,0 +1,64 @@
...
+        ).T
+        return coords
+
+    def select_fcoords(self, dobj):
+        mask = self._get_selector_mask(dobj.selector)
+        if mask is None:
+            return np.empty((0, 3), dtype="float64")
+        cell_centers = [
+            self.LeftEdge[i]
+            + np.add.accumulate(self.cell_widths[i])
```
neutrinoceros, line 52: There's a bug here because `self.LeftEdge` is presumably a `unyt_array` with length dimensionality, while `np.add.accumulate` currently is *not* supported by unyt (it doesn't preserve units). If you mean to get unitless arrays here this could be a fix, although we should also double check that `LeftEdge` and `cell_widths` are stored using the same unit.
```suggestion
            self.LeftEdge[I].d
            + np.add.accumulate(self.cell_widths[i])
```

neutrinoceros: I further report this is the *only* blocking bug for yt_idefix https://github.com/neutrinoceros/yt_idefix/pull/120
🎉

neutrinoceros: After further inspection I *think* it's safe to assume both cell_widths and LeftEdge are provided in `"code_length"` units, so the minimal suggestion _should_ be self-sufficient

yt/data_objects/index_subobjects/stretched_grid.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
+
+from yt.geometry.selection_routines import convert_mask_to_indices
+
+from .grid_patch import AMRGridPatch
+
+
+class StretchedGrid(AMRGridPatch):
```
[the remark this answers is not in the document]
neutrinoceros: I can answer this myself now: yes it is.

doc/source/examining/loading_data.rst
neutrinoceros: Maybe worth specifying _why_ they are not supported yet for AMRVAC (missing compatibility with multi grids and AMR)

yt/frontends/stream/data_structures.py
```diff
@@ -133,6 +157,36 @@ def __init__(self, ds, dataset_type=None):
...
+    def _icoords_to_fcoords(self, icoords, ires, axes=None):
+        """
+        We check here that we have cell_widths, and if we do, we will provide them.
+        """
+        if self.grid_cell_widths is None:
+            return super()._icoords_to_fcoords(icoords, ires, axes)
+        # Ideally this would be ported to something more efficient.  As it
+        # stands, this will use a pure python loop!
+        if axes is None:
+            axes = [0, 1, 2]
```
neutrinoceros, line 169: If we use a tuple instead of a list, it'll be safe to use as a default value, an this would become unneeded

matthewturk: I'm not totally sure I am ready to make this change, since using `None` is what's done in the baseclass.  But I will look into it.  This particular routine is also likely able to be made much more efficient with numpy tricks -- I just couldn't figure it out.

yt/data_objects/index_subobjects/stretched_grid.py
matthewturk: I actually don't know why this works for me -- do you?

matthewturk: (i.e., why didn't I see this bug?)

yt/frontends/artio/data_structures.py
```diff
@@ -333,6 +333,25 @@ def _read_fluid_fields(self, fields, dobj, chunk=None):
         )
         return fields_to_return, fields_to_generate
 
+    def _icoords_to_fcoords(self, icoords, ires, axes=None):
```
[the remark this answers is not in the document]
matthewturk: Base class *doesn't* implement the method.

yt/data_objects/index_subobjects/stretched_grid.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, id, cell_widths, filename=None, index=None):
+        self.cell_widths = [np.array(_) for _ in cell_widths]
+        super().__init__(id, filename, index)
+
+    def _check_consistency(self):
+        computed_right_edge = self.LeftEdge + [_.sum() for _ in self.cell_widths]
+        assert (computed_right_edge == self.RightEdge).all()
```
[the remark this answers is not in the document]
matthewturk: Probably -- can we keep it in for just a little while, while we stress-test it?

yt/data_objects/index_subobjects/stretched_grid.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def _get_selector_mask(self, selector):
+        if self._cache_mask and hash(selector) == self._last_selector_id:
```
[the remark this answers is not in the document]
matthewturk: Maybe?  Came right out of `grid_patch.py`

yt/frontends/artio/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: Oh that's right. Looks like it should be declared as an abstract method in the base class then, since it's not optional (unless I'm missing something again ?)

yt/data_objects/index_subobjects/stretched_grid.py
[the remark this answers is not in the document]
neutrinoceros: sounds good. Note however that it's very likely to have the very same bug I pointed to elsewhere regarding summing unyt arrays and raw numpy arrays

matthewturk:
So it does look like a scaling issue, related to the `icoords_to_fcoords` in `refine_by == 4` datasets.  Here are the reference and new plots:

(reference)
image

(new)
image

Here's another, where I've set `origin="native"` to demonstrate a bit more clearly:

image

Off by a factor of 4.  But, I can't figure out why it's doing this -- and why I have a comment in `icoords_to_fcoords` suggesting it may be off by a factor of `refine_by` (which it is in this case but no others.)

matthewturk:
The most recent change fixes things.  It happened because when we add `ires`, in order to get it to work with the quad tree, we use an `ires_factor`.  This undoes that, which then fixes the cell width.  This should not need to be applied in the stream frontend.

matthewturk:
whoa, all the tests pass!!!

matthewturk:
I'm working on cythonizing the `_icoords_to_fcoords` implementation in the stream frontend, and will use this to make the process more general and to support multiple sets of cell_widths.

matthewturk:
@yt-fido test this please

matthewturk:
This is interesting.  The tests are currently not passing for some of the chombo tests.  For instance, with the test `yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'density')_0_None (from nosetests)` we get:

```
Not equal to tolerance rtol=1e-10, atol=0
px values of ('gas', 'density') (None weighted) projection (axis 0) not equal.
Mismatched elements: 28 / 26800 (0.104%)
Max absolute difference: 3.0859
Max relative difference: 65.
 x: array([-2.421875, ...,  2.421875])
 y: array([-2.421875, ...,  2.421875])
```

```
  File "/usr/lib64/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib64/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/usr/lib64/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/fido/shiningpanda/jobs/a7d58705/virtualenvs/d41d8cd9/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/fido/workspace/yt_py38_git/yt/utilities/answer_testing/framework.py", line 400, in __call__
    self.compare(nv, ov)
  File "/home/fido/workspace/yt_py38_git/yt/utilities/answer_testing/framework.py", line 600, in compare
    assert_allclose_units(
  File "/home/fido/workspace/yt_py38_git/yt/testing.py", line 1292, in assert_allclose_units
    return assert_allclose(act, des, rt, at, **kwargs)
  File "/home/fido/shiningpanda/jobs/a7d58705/virtualenvs/d41d8cd9/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 1527, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/home/fido/shiningpanda/jobs/a7d58705/virtualenvs/d41d8cd9/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 844, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-10, atol=0
px values of ('gas', 'density') (None weighted) projection (axis 0) not equal.
Mismatched elements: 28 / 26800 (0.104%)
Max absolute difference: 3.0859
Max relative difference: 65.
 x: array([-2.421875, ...,  2.421875])
 y: array([-2.421875, ...,  2.421875])
```

matthewturk:
The full list of failing tests (9) is:

```
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'density')_0_None
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'density')_1_None
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'density')_2_None
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'velocity_magnitude')_0_None
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'velocity_magnitude')_1_None
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'velocity_magnitude')_2_None
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'magnetic_field_x')_0_None
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'magnetic_field_x')_1_None
yt.frontends.chombo.tests.test_outputs.ProjectionValues_data_0077_3d_hdf5_sphere_('max', (0_1, 'unitary'))_('gas', 'magnetic_field_x')_2_None
```

matthewturk:
My suspicion, before I dig in, is that a change in which cells are included has hit.  I intend to verify this.

matthewturk:
As an update, I manually compared, and I think the issue is related to something with the projection code, so I'll dig in there -- the fun thing is that it's not impacting the density values, just the projection position!  So it should be easier to dig out.

```
$ python -i compare.py
Field density has 26800 elements with 26800 equal.
Field pdx has 26800 elements with 26772 equal.
Field pdy has 26800 elements with 26772 equal.
Field px has 26800 elements with 26772 equal.
Field py has 26800 elements with 26772 equal.
Field weight_field has 26800 elements with 26800 equal.
```

matthewturk:
I have to sign off in a few, but to close this up -- something seems to be up with the level 0 `pdx` values.  (Something that momentarily tripped me up was that the `pdx` values are *half*-widths.)  I believe that it's showing up because we end up with no level 0 values in the full-domain projections, but I have to verify that.  Anyway, I'm tracking it down, as it seems all the level 0 values are off by a factor of 2.

matthewturk:
I have figured out the root of the problem.

What is happening is that for cells on the edge of a data selector (i.e., the sphere in this case), we can in some cases contribute only *partially* to the cell.  When we are using an `ires_factor`, this results in some cases where we have intermediate levels included -- for instance, we might refine a level 0 quadtree node to level 1, but only *some* of the level 1 nodes get refined to level 2.  This results in having some "hanging" nodes.  These exist in the old code as well, but because we directly track the `pdx` and `pdy` values, they don't cause problems.

Here, though, they do -- because we're in a situation where we get back from the calculation things like "level 1" which can't exist, since we have an `ires_factor` set to 2.  (i.e., all of our results should be multiples of two.)

I'm not completely sure yet that I know what I want the solution to be.  I will continue thinking about it; these are non-operative cells, so in principle our solution could be to ignore them in the tests, or just build a new set of answer tests for this dataset.  But I'm not sure yet.

matthewturk:
I am bumping the test values.  I believe this is a suitably small edge case, with no impact on the visualizations, that we can do this safely.

jzuhone:
What else do we need to get this across?

matthewturk:
I think it's good.

neutrinoceros:
Awesome. I will review this very soon, but more importantly I will setup some new tests in yt_idefix to check specifically the integration !

doc/source/examining/loading_data.rst
```diff
@@ -1951,6 +1963,62 @@ have assumed your data is stored in the three-dimensional array
...
+-------------------
+
+.. warning::
+
+   API consistency for loading stretched grids is not guaranteed until at least
+   yt 4.2!  There may be changes in between then and now, as this is a
+   preliminary feature.
+
+With version 4.1, yt has the ability to specify cell widths for grids.  This
+allows situations where a grid has a functional form for its widths, or where
```
neutrinoceros, line 1978: Or something similar
```suggestion
allows situations where a grid has a functional form for cell widths, or where
```

doc/source/examining/loading_data.rst
```diff
@@ -1951,6 +1963,62 @@ have assumed your data is stored in the three-dimensional array
...
+widths are provided in advance.
+
+.. note::
+
+   At present, support is available for a single grid with varying cell-widths,
+   loaded through the stream handler.  Future versions of yt will have more
+   complete and flexible support!
+
+To load a stretched grid, you use the standard (and now rather-poorly named)
```
neutrinoceros, line 1987: so we don't advertise how to plug this into a frontend for now, right ? I'm okay with this, since we know it's unstable, and I will gladly have yt_idefix be the avant-guard here, just checking we're on the same page

yt/data_objects/index_subobjects/stretched_grid.py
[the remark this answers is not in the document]
neutrinoceros: last comment here is still relevant (re: units)

yt/data_objects/index_subobjects/stretched_grid.py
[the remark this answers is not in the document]
neutrinoceros: actually it's a boolean, and it's currently _never_ set to anything else than `True`, so I propose we remove it instead of cargo cutting it any further (https://github.com/yt-project/yt/pull/4048)

yt/data_objects/index_subobjects/stretched_grid.py
```diff
@@ -0,0 +1,64 @@
...
+            mask = self._last_mask
+        else:
+            mask = selector.fill_mask(self)
+            if self._cache_mask:
+                self._last_mask = mask
```
neutrinoceros, line 23: see https://github.com/yt-project/yt/pull/4048
```suggestion
            self._last_mask = mask
```

yt/frontends/stream/data_structures.py
```diff
@@ -72,6 +76,27 @@ def Children(self):
         return [self.index.grids[cid - self._id_offset] for cid in self._children_ids]
 
 
+class StreamStretchedGrid(StretchedGrid):
+    _id_offset = 0
+
+    def __init__(self, id, index):
+        cell_widths = index.grid_cell_widths[id - self._id_offset]
+        super().__init__(id, cell_widths, None, index)
```
neutrinoceros, line 84: seems clearer to omit an argument with a default value rather than passing a `None` positionally
```suggestion
        super().__init__(id, cell_widths, index=index)
```

yt/geometry/grid_geometry_handler.py
```diff
@@ -450,6 +450,25 @@ def _chunk_io(
                 with self.io.preload(dc, preload_fields, 4.0 * size):
                     yield dc
 
+    def _icoords_to_fcoords(self, icoords, ires, axes=None):
```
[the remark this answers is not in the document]
neutrinoceros: still relevant

neutrinoceros (dismissed):
A couple questions and suggestions. Overall nothing blocking, thank you again for doing this, it means a lot to me !

yt/data_objects/construction_data_containers.py
```diff
@@ -279,18 +279,23 @@ def get_data(self, fields=None):
         # TODO: Add the combine operation
         xax = self.ds.coordinates.x_axis[self.axis]
         yax = self.ds.coordinates.y_axis[self.axis]
-        ox = self.ds.domain_left_edge[xax].v
-        oy = self.ds.domain_left_edge[yax].v
-        px, py, pdx, pdy, nvals, nwvals = tree.get_all(False, merge_style)
+
+        ix, iy, ires, nvals, nwvals = tree.get_all(False, merge_style)
+        px, pdx = self.ds.index._icoords_to_fcoords(
+            ix[:, None], ires // self.ds.ires_factor, axes=(xax,)
```
neutrinoceros, line 285: In testing with yt_idefix, I found a problem with this line:
since `ires_factor` is defined as `log2(ds.refine_by)`, this leads to a `RuntimeWarning: divide by zero encountered in floor_divide` for Dataset subclasses that set `refine_by = 1` (usually meaning "this isn't AMR, leave me be"). We can solve this by
- special casing for `self.ds.ires_factor == 0` here
- modify `ires_factor` somehow
- set a standard for what to do with `refine_by` in formats that are not AMR (set `refine_by = 2` anyway ? override the `ires_factor` property ?)

It's definetly a bug, I'm just not sure what part, this branch or frontends, need a fix

matthewturk: Hmm, good point.  My *first* inclination is to not allow `ires_factor` to go below 1.  The point really is to make a number that we can use here, and not really anywhere else (yet).

doc/source/examining/loading_data.rst
matthewturk: I'm happy to add this once we get AMR support working, I think.  But yeah for now let's not?

doc/source/examining/loading_data.rst
matthewturk: We should, you're right.  Can you or @chrishavlin suggest an improvement?

yt/geometry/grid_geometry_handler.py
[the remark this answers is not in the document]
matthewturk: OK, I'll work on adding them.

yt/geometry/oct_geometry_handler.py
```diff
@@ -116,3 +116,22 @@ def _mesh_sampling_particle_field(field, data):
             units=units,
             take_log=take_log,
         )
+
+    def _icoords_to_fcoords(self, icoords, ires, axes=None):
```
matthewturk, line 120: I'll add here too

matthewturk (commented):
I'll keep working on the remaining suggestions as well.

yt/data_objects/construction_data_containers.py
neutrinoceros: Sounds good. Now that I think about it, `refine_by = 1` doesn't make much sense (the real issue being that this attribute is needed even for formats where it's irrelevant -> #1880)
Maybe it would also be ok to change these rare frontends to use `refine_by = 2` instead ? this would avoid increasing the complexity of the base implementation of `ires_factor`. WDYT ?

doc/source/examining/loading_data.rst
neutrinoceros: still relevant (Sorry GitHub is such a pain with long discussions)

doc/source/examining/loading_data.rst
neutrinoceros: I'd just go with `widths`

yt/data_objects/construction_data_containers.py
matthewturk: It would, but I think the change to ires factor is fine as well.

yt/data_objects/index_subobjects/stretched_grid.py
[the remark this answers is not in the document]
matthewturk: I can't find the comment about unyt arrays -- can you briefly remind me?

yt/data_objects/construction_data_containers.py
neutrinoceros: sure, we can absolutely do both

neutrinoceros: or not, being able to grep "refine_by = 1" has value in a sense.

chrishavlin (commented):
I didn't look too thoroughly at the implementation details here, but did check out the branch and play around and it's working great. Left a few minor comments and questions.

yt/frontends/stream/tests/test_stream_stretched.py
```diff
@@ -0,0 +1,49 @@
+import numpy as np
+
+from yt import load_uniform_grid
+from yt.testing import assert_almost_equal, assert_equal
+
+
+def test_variable_dx():
```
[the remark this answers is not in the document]
neutrinoceros: Stakeholders that know of are almost exclusively using non-cartesian geometries, so I'd say "probably"

yt/frontends/artio/data_structures.py
```diff
@@ -333,6 +333,25 @@ def _read_fluid_fields(self, fields, dobj, chunk=None):
...
+        grids.
+        """
+        dds = self.ds.domain_width[(axes,)] / (
+            self.ds.domain_dimensions[
+                axes,
+            ]
+            # This may need an additional refinement factor conversion for
+            # non-refine-by-2 cases.
+            * self.ds.refine_by ** ires[:, None]
+        )
```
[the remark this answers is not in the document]
neutrinoceros: (I took the liberty to edit Chris' suggestion to remove a space char so black doesn't complain)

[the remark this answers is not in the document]
matthewturk: Oh, and also -- it's no longer true!

yt/frontends/stream/tests/test_stream_stretched.py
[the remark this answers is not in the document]
matthewturk: So I'm not sure.  I looked at it and I think it's very straightforward to add if we use `pytest.parameterize` (can we do that now?) but I also don't know that we actually gain any info.

yt/geometry/geometry_handler.py
```diff
@@ -48,6 +48,19 @@ def __init__(self, ds, dataset_type):
...
+        self,
+        icoords: np.ndarray,
+        ires: np.ndarray,
+        axes: Optional[Tuple[int, ...]] = None,
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        # What's the use of raising NotImplementedError for this, when it's an
+        # abstract base class?  Well, only *some* of the subclasses have it --
+        # and for those that *don't*, we should not be calling it -- and since
+        # it's a semi-private method, it shouldn't be called outside of yt
+        # machinery.  So we shouldn't ever get here!
```
neutrinoceros, line 61: ™️ 
... jokes aside thank you for this comment :)

yt/geometry/_selection_routines/selector_object.pxi
```diff
@@ -404,6 +405,100 @@ cdef class SelectorObject:
                             total += mask[i, j, k]
         return total
 
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    @cython.cdivision(True)
+    def fill_mask(self, gobj):
+        # This is for an irregular grid.  We make no assumptions about
+        # the shape of the dds values, except that we will regularize to the
+        # max dimension.
```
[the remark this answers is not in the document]
matthewturk: I had originally planned a static allocation to `MAX(dim[0], dim[1], dim[2])` but decided against it.  Changing comment!

yt/frontends/stream/tests/test_stream_stretched.py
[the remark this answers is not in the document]
matthewturk: So if we do this, we'll need to change our axis name index field checks -- which I'm actually not sure gains us any information.  One option instead would be to do a check on the total volume.  But again, I'm kinda not sure yet if it's worth it.

matthewturk (commented):
I have to head out for a bit, but I think this is in a good shape for final review -- I *think* I've addressed everything with the one potential outstanding issue of the parameterized tests.

yt/frontends/stream/tests/test_stream_stretched.py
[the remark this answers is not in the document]
neutrinoceros: We can use pytest.mark.parametrize but only at the cost of making the test module pytest-only (which means adding it to *two* blacklists) I'd say it's not necessarily worth it here. yt_idefix will be ground 0 for catching any bugs with spherical, cylindrical, and polar geometries ;-)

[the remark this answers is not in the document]
chrishavlin: > which I'm actually not sure gains us any information.

Ya, that's why I was uncertain -- I couldn't tell if it'd be more of a test of yt's existing coordinate-system handling or a test of the actual dds implementation. 

> yt_idefix will be ground 0 for catching .... 

I think leaving this test alone and testing dds with different geometries more indirectly through frontends is totally reasonable -- I would like to add some actual image tests for the geographic geometries (if I can figure out cartopy testing...), so that would cover this for the cases I've been playing with.

yt/data_objects/index_subobjects/stretched_grid.py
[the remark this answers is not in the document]
neutrinoceros: I was incorrect, `_cache_mask` is a boolean, so, nothing to do here

yt/data_objects/index_subobjects/stretched_grid.py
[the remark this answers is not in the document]
neutrinoceros: resolved

yt/frontends/artio/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: done

neutrinoceros (approved):
very happy to report all my threads are now closed !

matthewturk:
OK so the blacken thing I see, but I'm kinda annoyed about the cm1 tests!

matthewturk:
pre-commit.ci autofix

matthewturk:
I'm trying setting the `cm1` value for `refine_by` to 2 to see if that fixes it; if it does, we'll know that there's been something corrected.

matthewturk:
If this current change works -- with `refine_by` set to 2 -- then I believe the *actual* fix could be to set `refine_by` back to 1 and change the levels in the `cm1` frontend to 0 from 1.

matthewturk:
Torn between laughing and crying:

```
Arrays are not equal

Mismatched elements: 1 / 1 (100%)
Max absolute difference: 1
Max relative difference: 1.
 x: array([[0]], dtype=int32)
 y: array([[1]], dtype=int32)
```

This is because I explicitly changed them.  :sob:

matthewturk:
Now that that is the only one that is changed -- all the actual results are identical -- I will bump the answer tests for cm1.

neutrinoceros:
Can you link any existing issue this resolves ? I can think of a couple candidates within [Support Stretched Grids](https://github.com/orgs/yt-project/projects/2) , but I'm not sure which ones are still relevant once this goes in

matthewturk:
This should fix both #3746 and most of #3694.

neutrinoceros (dismissed):

neutrinoceros:
resolved a couple conflicts from #4047

chrishavlin (approved):
Looks good!

I resolved the one thread about whether or not we should add tests for different geometries. From that discussion, it seems clear that we can rely on the frontends that use the variable cell widths in different geometries rather than adding a bunch of explicit tests here.

neutrinoceros:
I'm ready to merge, unless we want another specific reviewer ?

PR #3000, opened 2020-12-16
refactor: remove yt.funcs.ensure_tuple

neutrinoceros:
## PR Summary

This is a follow up to #2893.
`ensure_tuple` is only used to sanitize `Dataset.periodicity` attributes in a such a fashion that it doesn't make any difference to use builtin tuple conversion. Hence, the function serves no purpose anymore and can be removed.

### DIFF
--- a/yt/frontends/amrvac/data_structures.py
+++ b/yt/frontends/amrvac/data_structures.py
@@ -343,9 +343,9 @@ class AMRVACDataset(Dataset):
             self.geometry = "cartesian"
 
         # parse peridiocity
-        per = self.parameters.get("periodic", np.array([False, False, False]))
-        missing_dim = 3 - len(per)
-        self.periodicity = np.append(per, [False] * missing_dim)
+        periodicity = self.parameters.get("periodic", ())
+        missing_dim = 3 - len(periodicity)
+        self.periodicity = (*periodicity, *(missing_dim * (False,)))
 
         self.gamma = self.parameters.get("gamma", 5.0 / 3.0)
 
--- a/yt/frontends/athena/data_structures.py
+++ b/yt/frontends/athena/data_structures.py
@@ -5,7 +5,7 @@ import numpy as np
 
 from yt.data_objects.index_subobjects.grid_patch import AMRGridPatch
 from yt.data_objects.static_output import Dataset
-from yt.funcs import ensure_tuple, mylog, sglob
+from yt.funcs import mylog, sglob
 from yt.geometry.geometry_handler import YTDataChunk
 from yt.geometry.grid_geometry_handler import GridIndex
 from yt.utilities.chemical_formulas import default_mu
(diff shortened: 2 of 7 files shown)


matthewturk:
Love that this is #3000.

matthewturk:
I need to think about this, because of the `always_iterable` call, to make sure that I agree it can't backfire.

neutrinoceros:
The way I rewrote `ensure_tuple` in the previous PR revealed what the function was meant to do, in a vacuum, but I'm fairly confident that the designed property (being able to loop over a possibly scalar variable) was not used anywhere.

neutrinoceros:
@matthewturk Looks like there's indeed a class where the `always_iterable` part was helping: `EnzoDataset`

original code
```python
self.periodicity = ensure_tuple(self.parameters["LeftFaceBoundaryCondition"] ==3)
```

Now it seems the easy fix would be
```python
self.periodicity = 3 * (self.parameters["LeftFaceBoundaryCondition"] ==3,)
```
with the side benefit that periodicity would always register as a three element tuple, instead of just one.
I hope this fixes the 7 failures seen on Travis that I got with my first attempt:
```python
self.periodicity = tuple(self.parameters["LeftFaceBoundaryCondition"] ==3)
```

neutrinoceros:
wow. Now I'm getting a very surprising error:

```
1906
ERROR: yt.data_objects.tests.test_io_geometry.test_preserve_geometric_properties
1907
1908
1909
  File "C:\Miniconda\lib\site-packages\nose\case.py", line 198, in runTest
1910
    self.test(*self.arg)
1911
  File "D:\a\yt\yt\yt\testing.py", line 803, in true_wrapper
1912
    return func(*args, **kwargs)
1913
  File "D:\a\yt\yt\yt\data_objects\tests\test_io_geometry.py", line 21, in test_preserve_geometric_properties
1914
    ds2 = load(fn)
1915
  File "D:\a\yt\yt\yt\loaders.py", line 99, in load
1916
    raise YTAmbiguousDataType(fn, candidates)
1917
yt.utilities.exceptions.YTAmbiguousDataType: Multiple data type candidates for C:\Users\RUNNER~1\AppData\Local\Temp\tmpvtkaxbc5\savefile.h5
1918
1919
<class 'yt.frontends.ytdata.data_structures.YTDataContainerDataset'>
1920
<class 'yt.frontends.ytdata.data_structures.YTDataset'>
1921
<class 'yt.frontends.ytdata.data_structures.SavedDataset'>
1922
<class 'yt.frontends.athena.data_structures.AthenaDataset'>
1923
A possible workaround is to directly instantiate one of the above.
1924
Please report this to https://github.com/yt-project/yt/issues/new
```
The weirdest part is that it only shows up on windows... after a commit that affects NONE of the incriminated classes.

neutrinoceros:
Got it. Looks like I was incredibly unlucky that the tmp dir was named `tmpvtkaxbc5`, which validates `AthenaDataset._is_valid()`, see https://github.com/yt-project/yt/blob/b86eaea8746f437a72f28d6c632eefd658b4f637/yt/frontends/athena/data_structures.py#L649
I'll just relaunch the CI and mark this as a different issue.

neutrinoceros:
I'm bundling a bugfix for AMRVAC while I'm at it... `AMRVACDataset.periodicity` was improperly parsed as a `numpy.ndarray`

matthewturk:
@neutrinoceros Looks like things are passing now -- this ready to go?

neutrinoceros:
Yes it is !

PR #3003, opened 2020-12-17
Add some minor release tooling

matthewturk:
## PR Summary

This adds support for [`bump2version`](https://pypi.org/project/bump2version/)
to make it easier to update versions.  It also adds a *currently
non-functional* release yaml.  If we want to do this, we'll need to add a PyPI
secret to the repository, which is completely feasible.

I believe, but am not sure, that we could potentially wrap the entirety of the
release process into this workflow.  What we would need to do in addition would
be to kick off the conda builds and whatnot.

## PR Checklist

- [x] pass `flake8 yt/`


munkm:
I like the version bumping infrastructure. 

I don't think the release yaml will necessarily make our workflow that much faster in a release. We still need to wait for multibuild to get the wheels built to finalize a pypi release and if there's a time gap between when the source tarball gets uploaded to pypi and the wheels get uploaded, somebody could feasibly try to pip install and then end up doing an (unwanted) source build. I am -1 on that part because I think uploading the wheels at the tarball at the same time is important.

setup.cfg
```diff
@@ -1,3 +1,24 @@
+[bumpversion]
+current_version = 4.0.dev0
+commit = True
+tag = True
+tag_name=yt-{new_version}
+serialize = {major}.{minor}.dev{patch}
+parse = (?P<major>\d+)\.(?P<minor>\d+)\.dev(?P<patch>\d+)
```
neutrinoceros, line 7: If I'm following everything, you're plan is to manually remove the `dev` part each time we're about to perform a release, but this feels a bit anticlimactic as this work is about automation.
From `bump2version`'s doc, it seems to me that a solution can be elaborated from their `optional_value =` section with the `<release>` variable, though I don't have time to build this myself right now.

neutrinoceros (commented):
I love this, it's awesome. I left a few questions/suggestions here and there. Another global suggestion is to maybe add inline comments with the effect of `#  do not edit this manually` to lines that are targeted by the tool.

setup.cfg
matthewturk: Ok, I *think* I understand it from reading this [blogpost](https://williamhayes.medium.com/versioning-using-bumpversion-4d13c914e9b8) and I'll give a shot at that...

matthewturk:
@munkm I'm going to reply to your comments; right now I've sidestepped that discussion, but will return to it.

@neutrinoceros I've changed the `bump2version` setup to match what you suggest.

Now it versions it as `{major}.{minor}.{patch}-dev{build}`.  What this means is that as of right now, it's set to install as 4.0.0-dev0.  When I do `bump2version build` it'll update to `4.0.0-dev1`, and on and on.  If we do `bump2version release` it'll suddenly become `4.0.0`.  Then when we update `patch` it'll go to `4.0.1-dev0` and so forth.

matthewturk:
@munkm So what I'm thinking about with respect to the releases is a couple things.  The first is that I was really only thinking about source distributions with this, and more specifically, I was thinking about how we could automate the process of getting source builds for development and release candidates up on pypi.  (I need to figure out if I've done something strange with the versioning for that, though.)

But, that aside, I know you have looked at lots of binary building things, and there are things to consider there.  So it should be a longer discussion where I defer to your thoughts.

neutrinoceros:
> @neutrinoceros I've changed the bump2version setup to match what you suggest.
>
> Now it versions it as {major}.{minor}.{patch}-dev{build}. What this means is that as of right now, it's set to install as 4.0.0-dev0. When I do bump2version build it'll update to 4.0.0-dev1, and on and on. If we do bump2version release it'll suddenly become 4.0.0. Then when we update patch it'll go to 4.0.1-dev0 and so forth.

Excellent ! Well done !

matthewturk:
@neutrinoceros Do you happen to know or have opinions on how we should have the `dev` in there, by what string format?

neutrinoceros (approved):

neutrinoceros:
I don't have pre-existing opinions on this, but I'll say that I'm happy with your current proposal. It doesn't overshadow the semantic-version number and also offers rooms for internal increments: it's perfect imo.

neutrinoceros:
@yt-fido test this please

neutrinoceros:
quick note: I'm trying to set this up within a different project and I report that this tool doesn't play nice with the "Trim Trailing Whitespace" pre-commit hook :/

munkm:
Bumping versions isn't super challenging for our current setup in yt releases. I personally think this is worth setting aside for now and reviewing for a later release.

matthewturk:
OK, I understand.  I do think we need to get to a situation where as much as possible is both automated and happening regularly, so I hope we can revisit this.

neutrinoceros:
side note: pandas is using versioneer, which looks like a similar project (maybe simpler ?)
https://pypi.org/project/versioneer/

PR #3006, opened 2020-12-21
boxlib: Fix _read_header for 2D data

dpgrote:
## PR Summary

When reading in raw data from a WarpX plot file for a 2D run, at line 76 of the routine _read_raw_field in boxlib/io.py it does box[0] - nghost, but box[0] is length 2 and nghost is length 3, raising an error. nghost is length 3 when it was set in _read_header in boxlib/data_structures.py since it is nghost = [ng, ng, ng]. This fix checks the number of dimensions and instead sets nghost accordingly.

## PR Checklist

- [x] pass `flake8 yt/`

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -1408,7 +1408,7 @@ def _read_header(raw_file, field):
 
             if nghost is None:
                 try:
-                    nghost = np.array(len(lo_corner)*[ng])
+                    nghost = np.array(len(lo_corner) * [ng])
                 except NameError:
                     # If lo_corner is not defined.
                     # Will this ever happen?
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -1387,14 +1387,10 @@ def _read_header(raw_file, field):
             f.readline()  # version
             f.readline()  # how
             f.readline()  # ncomp
-            nghost_lines = f.readline().strip().split()
-            try:
-                ng = int(nghost_lines[0])
-                # nghost will be set below after the number of dimensions is determined
-                nghost = None
-            except ValueError:
-                nghosts = nghost_lines[0][1:-1].split(",")
-                nghost = np.array([int(ng) for ng in nghosts])
+
+            # nghost_line will be parsed below after the number of dimensions is determined
+            nghost_line = f.readline().strip().split()
+
             f.readline()  # num boxes
 
             # read boxes
@@ -1406,13 +1402,15 @@ def _read_header(raw_file, field):
                 lo_corner, hi_corner, node_type = _line_to_numpy_arrays(clean_line)
                 boxes.append((lo_corner, hi_corner, node_type))
 
-            if nghost is None:
-                try:
-                    nghost = np.array(len(lo_corner) * [ng])
-                except NameError:
-                    # If lo_corner is not defined.
-                    # Will this ever happen?
-                    nghost = np.array([ng, ng, ng])
+            try:
+                # nghost_line is a single number
+                ng = int(nghost_line[0])
+                ndims = len(lo_corner)
+                nghost = np.array(ndims * [ng])
+            except ValueError:
+                # nghost_line is (#,#,#)
+                nghost_list = nghost_line[0][1:-1].split(",")
+                nghost = np.array([int(ng) for ng in nghost_list])
 
             # read the file and offset position for the corresponding box
             file_names = []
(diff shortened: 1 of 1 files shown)


yt/frontends/boxlib/data_structures.py
```diff
@@ -1405,6 +1406,14 @@ def _read_header(raw_file, field):
                 lo_corner, hi_corner, node_type = _line_to_numpy_arrays(clean_line)
                 boxes.append((lo_corner, hi_corner, node_type))
 
+            if nghost is None:
+                try:
+                    nghost = np.array(len(lo_corner) * [ng])
+                except NameError:
+                    # If lo_corner is not defined.
+                    # Will this ever happen?
```
neutrinoceros, line 1414: I don't think it's worth catching an exception if we don't even know why it would happen.

yt/frontends/boxlib/data_structures.py
```diff
@@ -1405,6 +1406,14 @@ def _read_header(raw_file, field):
                 lo_corner, hi_corner, node_type = _line_to_numpy_arrays(clean_line)
                 boxes.append((lo_corner, hi_corner, node_type))
 
+            if nghost is None:
```
neutrinoceros, line 1409: I would suggest to only parse `nghost_lines`* here once `lo_corner` is defined, in order to make the process a bit easier to follow. To be clear, my suggestion is to move the previous try block here and merge it with this `if` statement.
Furthermore, I think it'd be clearer to use a variable here , e.g. `ndim = len(lo_corner)`

* come to think of it the name of this variable is also misleading since it really refers to a _single_ line being parsed into _several_ values... can you also rename it while you're at it ?

neutrinoceros (commented):
Hi @dpgrote, thank you for this patch !
I am favorable to this going in but I think the process should be easier to follow for the sake of future maintainability. I left some suggestions in that regards, let me know if you need help !

yt/frontends/boxlib/data_structures.py
dpgrote: Thanks for the comments - changes made. Did you mean to rename `nghost_lines`?

yt/frontends/boxlib/data_structures.py
```diff
@@ -1405,6 +1403,16 @@ def _read_header(raw_file, field):
...
                 boxes.append((lo_corner, hi_corner, node_type))
 
+            try:
+                # nghost_line is a single number
+                ng = int(nghost_line[0])
+                ndims = len(lo_corner)
+                nghost = np.array(ndims * [ng])
+            except ValueError:
+                # nghost_line is (#,#,#)
+                nghost_list = nghost_line[0][1:-1].split(",")
```
neutrinoceros, line 1413: took me a little while to understand why we'd want to ignore the first and last characters, I would refactor this line as 
```suggestion
                nghost_list = nghost_line[0].strip("()").split(",")
```

yt/frontends/boxlib/data_structures.py
```diff
@@ -1405,6 +1403,16 @@ def _read_header(raw_file, field):
...
+                nghost = np.array([int(ng) for ng in nghost_list])
```
neutrinoceros, line 1414: while we're at it, might as well make this more idiomatic to numpy
```suggestion
                nghost = np.array(nghost_list, dtype="int64")
```

yt/frontends/boxlib/data_structures.py
```diff
@@ -1405,6 +1403,16 @@ def _read_header(raw_file, field):
                 lo_corner, hi_corner, node_type = _line_to_numpy_arrays(clean_line)
                 boxes.append((lo_corner, hi_corner, node_type))
 
+            try:
+                # nghost_line is a single number
+                ng = int(nghost_line[0])
+                ndims = len(lo_corner)
+                nghost = np.array(ndims * [ng])
+            except ValueError:
+                # nghost_line is (#,#,#)
```
neutrinoceros, line 1412: This comment is helpful _but_ I think it's slightly misleading because you are specifying the format of `nghost_line[0]`, not `nghost_line` (and I have no idea what the rest of the list is... is the `.split()` a security to exclude possible comments ?)

neutrinoceros (commented):
Thanks a lot for taking my first review into account ! now the code is much more readable imo, which allows me to give additional minor suggestions/comments. I'll approve + merge once those adressed :-)

yt/frontends/boxlib/data_structures.py
neutrinoceros: exactly, and you nailed it !

neutrinoceros (approved):

dpgrote:
Thanks for merging!

neutrinoceros:
Thanks again for taking the time to patch this, we appreciate it !

Issue #2911, opened 2020-09-17
discussion : a clear deprecation cycle after yt 4.0


neutrinoceros:
Lots of features have been marked as deprecated over the years but there are currently no clear rules to enforce a clean deprecation cycle.
I propose that the 4.0 release be the last without such rules.
A simple way to transition would be to produce a 4.1 release that would be identical to 4.0 _except_ that it would remove everything that was deprecated earlier than 3.6, and clearly mark the rest of deprecations as due for removal in 4.2, in a user-visible way.
After that, all new deprecations would require a targeted "date" (version) for removal, so we can clear them in time for the appropriate release.
A sensible pattern would be remove deprecated stuff after two releases: one that visibly indicates deprecation, and removal in the following one.

I note that currently some deprecated features are invisible to the users and only marked as such in comments. Following the above proposed rules, those should survive in 4.1 but be removed in 4.2

I intend to bring this up in the next team meeting.

neutrinoceros:
this is fixed since #3007

PR #3007, opened 2020-12-25
maint: remove deprecated stuff

neutrinoceros:
## PR Summary

Gather deprecation utilities functions to a new module dedicated to maintenance, and remove stuff that was marked as deprecated as of yt 3.6.0

This is aimed at solving #2911 for yt 4.0 (or 4.1)

Note: I'm not going to touch the config-file related code until we're done with #2981

## PR Checklist

- [x] move every deprecation-related helper functions/classes to an isolated module
- [x] remove every deprecated feature older than yt 3.6.0
- [x] check for comments specifying "backward compatibility" and apply an actual marker where needed
- [x] update surviving deprecation warnings with clear version-of-deprecation and removal version
- [x] remove `issue_demeshening_deprecation_warning` (repeat a string if need be, the important aspect is that this warning should not be harder to discover than the rest of them)
- [x] rename `yt.maintenance` to `yt._maintenance` to clarify that this not a user-facing module
- [x] cleanup deprecation module
- [x] update documentation


yt/data_objects/selection_objects/cut_region.py
```diff
@@ -59,6 +59,7 @@ def __init__(
         validate_object(field_parameters, dict)
         validate_object(base_object, YTSelectionContainer)
         if base_object is not None:
+            # todo: handle this (where's the explicit deprecation ?)
```
neutrinoceros, line 62: @matthewturk , this is a question for you. I believe you wrote the surrounding code.
I would like to know whether it's safe to remove this if-block or if I should add a call to `issue_deprecation_warning`.

matthewturk: I believe I did write this.  But I think that we can safely remove this.

neutrinoceros: thanks !

benchmarks/benchmarks/data_objects.py
```diff
@@ -67,15 +67,5 @@ def time_slice(self):
     else:
 
         def time_slice(self):
-            slc = self.ds.h.slice(0, 0.5, self.field_name)
+            slc = self.ds.slice(0, 0.5, self.field_name)
             slc[self.field_name]
-
```
matthewturk, line 72: I'm not saying we should keep the `benchmarks` directory.  But if we're going to, we should probably only excise the yt 2 stuff in this function.

neutrinoceros: TBH I don't know what this code is supposed to be run and I didn't want to go to deep into this. Here I just made sure that this file wouldn't break horribly.
I'm not confident in updating the benchmark files further: looks to me like most of the if/else are written with only yt 2 VS yt 3 in mind, and the mere fact that we are now at 4.x is going to break them anyway.

matthewturk: How about just remove benchmarks as part of this?

matthewturk: Unless @Xarthisius ever wants to bring it back -- I think he was not convinced we gained much beyond what the test infrastructure brings.

neutrinoceros: I'd be inclined to remove those in a separate PR, because I intend that this one can later serve as a model for how to prep a release on the specific aspect of deprecation :)

Xarthisius: It can go away as far as I'm concerned.

neutrinoceros: Thanks, it's gone now !

neutrinoceros:
@yt-fido test this please

matthewturk (approved):

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

yt/data_objects/static_output.py
```diff
@@ -1672,7 +1635,20 @@ def add_smoothed_particle_field(
...
+
+        issue_deprecation_warning(
+            "This method is deprecated."
+            "Since yt-4.0, it's no longer necessary to add a field specifically for "
+            "smoothing, because the global octree is removed. The old behavior of "
+            "interpolating onto a grid structure can be recovered through data objects "
+            "like ds.arbitrary_grid, ds.covering_grid, and most closely ds.octree. The "
+            "visualization machinery now treats SPH fields properly by smoothing onto "
+            "pixel locations. See this page to learn more: "
+            "https://yt-project.org/doc/yt4differences.html",
```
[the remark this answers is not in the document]
neutrinoceros: Good point, I hadn't noticed. However, I note that the source file for it _does_ exist `doc/source/yt4differences.rst`; but for some reason it's not included in the final build.
It's expected that this exact url doesn't work yet, but `https://yt-project.org/dev/yt4differences.html` *should*. I think this is a bug you just found but it should be fixed in a dedicated PR.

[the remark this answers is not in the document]
matthewturk: Strange, yeah.  Huh!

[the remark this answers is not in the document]
neutrinoceros: reported as https://github.com/yt-project/yt/issues/3029

PR #3010, opened 2020-12-28
maint: add simple pre-commit hooks

neutrinoceros:
## PR Summary

Add three simple hooks from the `pre-commit-hooks` repo 
- trailing-whitespace
- end-of-file-fixer (add `\n` at the end of files that don't have it already)
- no-commit-to-branch: formally forbid accidental commits directly going to master

I also remove a fragile local script.
most of the changes in this PR are automatic, obtained with
```bash
pre-commit run trailing-whitespace --all-files
pre-commit run end-of-file-fixer --all-files
```


neutrinoceros:
note: it seems to me that none of the remaining files in `scripts/` is still useful and could be thrown out too, though that should be done in a different PR

matthewturk:
This all seems OK to me; any things in particular to look at?

neutrinoceros:
Well it should be noted that these changes won't be stable unless we adopt pre-commit.ci immediately after. For now I didn't get feedback outside of yours (positive), so I'd be inclined to wait a few more days before taking action here.

PR #3009, opened 2020-12-28
maint: fix several problems in .pre-commit-config.yaml

neutrinoceros:
## PR Summary

- Add missing deps for isort and flake8 pre-commit hooks.
- reorder black and flynt hooks for better stability
- add a global exclude list (this is needed to later setup pre-commit.ci which run `pre-commit run --all-files`)

PR #3013, opened 2020-12-28
pre-commit ci adoption: docs

neutrinoceros:
## PR Summary

In support of my latest proposal to adopt https://pre-commit.ci (see https://mail.python.org/archives/list/yt-dev@python.org/thread/QHDXCWXFMEPKWS3Y5TJLU4UAS3I3UGHH/).
This PR should not be merged before the tool is deployed, and can be closed is case of rejection.

todo:
- [x] remove comments `# keep in sync with lint_requirement.txt` from `pre-commit-config.yaml` (will likely conflict with #3009)
- [x] possibly remove redundant exclude lists from `setup.cfg` and `pyproject.toml` since flake8, black and isort should not be run directly on the command line any more (note: I'm keeping this for an isolated follow up PR)
- [x] explicitly document that linters should not be run directly on the command line or in text editors but only through pre-commit (be it with it CLI). The reason for this is to avoid conflicts between different versions of each tool.
- [x] rebasing (after #3020 goes in)


tests/test_install_script.py
```diff
@@ -110,24 +110,21 @@ def verify_yt_installation(binary_yt):
...
             if dep == "rockstar":
                 # compiling rockstar is broken on newer MacOS releases
                 if platform.mac_ver()[0].startswith(("10.12", "10.13")):
                     continue
                 cmd = (
-                    "LD_LIBRARY_PATH={} ".format(
-                        os.sep.join([os.curdir, yt_dir, "lib"])
-                    )
-                    + cmd
+                    f"LD_LIBRARY_PATH={os.sep.join([os.curdir, yt_dir, 'lib'])} " + cmd
```
cphyc, line 121: I know this isn't your code, but it could be replaced by
```suggestion
                    f"LD_LIBRARY_PATH={os.path.join(os.curdir, yt_dir, 'lib')} " + cmd
```

cphyc (approved):
I appreciate the reduction of manual intervention this incurs!

tests/test_install_script.py
neutrinoceros: I'd rather keep this PR as pure as possible. I replicated those automated changes in #3020

Xarthisius (approved):

PR #3016, opened 2021-01-03
Fix fragile imports

neutrinoceros:
## PR Summary

For a while there was a note in pyproject.toml that said isort couldn't be applied to yt/__init__.py because it would create circular imports. Turns out there are only a handful of fragilities to fix in that regard, and I'm doing it here.


yt/__init__.py
```diff
@@ -17,77 +17,72 @@
 __version__ = "4.0.dev0"
 
 # First module imports
-import numpy as np  # For modern purposes
 import numpy  # In case anyone wishes to use it by name
+import numpy as np  # In case anyone wishes to use it by name
```
neutrinoceros, line 21: note to reviewers: the comment duplication here is an unfortunate side effect (I dare say bug) in isort, likely because it gets confused that we import the same module twice...

Though this has nothing to do with the original scope of this PR, I would like to propose we remove these imports completely: unless I'm missing the point completely, I do not think it is wise to encourage users to import numpy FROM yt, which is the only thing these lines get u, seems to me.

neutrinoceros: bug confirmed upstream (note that I reported it with a different version of isort, so it's not exactly a reproduction of what we're seeing here): https://github.com/PyCQA/isort/issues/1631#issuecomment-753828006

matthewturk: I think it's OK to not import it here; I briefly thought maybe it might have implications for either ensuring numpy is imported (it does set up a global state) or making sure that it's accessible to the plugins, but I think in both cases it's not actually relevant.

matthewturk (approved):

neutrinoceros:
Indeed the one test that failed was the one defining a sample plugin file: it broke because `np` wasn't defined anywhere else.
I went ahead and fixed it but this _does_ mean that this removal is able to cause breakage downstream...

matthewturk:
I think able, but not likely.

neutrinoceros:
Well merge away if you're up for it, but I'm adding the "backward incompatible" label for the release manager because the natural error message is really not helpful:

```
# from the test
...
yt.utilities.exceptions.YTFieldNotFound: Could not find field ('unknown', 'random') in UniformGridData.
```

users should be warned that they need to import numpy themselves from the plugin file if they intend to use it there.

PR #3021, opened 2021-01-16
Dead code detection: now with pre-commit hooks

neutrinoceros:
## PR Summary

This is a revamp of #2973, using the newly defined pre-commit hook from vulture instead of a harder-to-maintained unique GitHub workflow.

The idea is still to prevent code to rot in place long after it stops being useful.
Important note: vulture doesn't honor noqas too well, but in some situations it reports function arguments as unused that we can't just remove because e.g. they are part of the original signature from the parent class and breaking compatibility is not worth it. The way around this problem is to prefix said argument's names with `_`, which makes vulture happy and is also a common indication for unused arguments. It might still not be satisfactory, but IMHO this is reasonable on the condition that it breaks no existing test.

I also note that vulture has a whitelisting mechanism but I'd rather avoid using it because it requires adding an obscure generated `.py` file to the repo, which is counter productive as the goal here is to de-clutter the project.

~edit: this is currently failing with errors tracing back to lines that are de facto in conflict with  #3007, so I'd rather keep this waiting in line for _that_ PR to be dealt with.~ ✅

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -70,7 +70,7 @@ class MutableAttribute:
         self.data = weakref.WeakKeyDictionary()
         self.display_array = display_array
 
-    def __get__(self, instance, owner):
+    def __get__(self, instance, _owner):
         if not instance:
             return None
         ret = self.data.get(instance, None)
@@ -393,7 +393,7 @@ class Dataset(abc.ABC):
         return False
 
     @classmethod
-    def _guess_candidates(cls, base, directories, files):
+    def _guess_candidates(cls, _base, _files):
         """
         This is a class method that accepts a directory (base), a list of files
         in that directory, and a list of subdirectories.  It should return a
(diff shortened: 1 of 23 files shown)


neutrinoceros:
@yt-fido test this please

neutrinoceros:
fix #2872

yt/data_objects/static_output.py
```diff
@@ -403,10 +403,15 @@ def _guess_candidates(cls, base, directories, files):
         This function doesn't need to catch all possibilities, nor does it need
         to filter possibilities.
         """
+        # arguments named should be prefixed by a '_' if and only if they are not used
+        # in the function's body
+        # note that the original signature included a `directories` args
+        # that was not used in any subclass
```
cphyc, line 409: Should this comment disappear?

yt/fields/particle_fields.py
```diff
@@ -847,33 +847,6 @@ def _pfunc_avg(field, data):
     return fn
 
 
-def add_volume_weighted_smoothed_field(
```
cphyc, line 850: Aren't we supposed to keep this one until `yt 4.0.1`?

neutrinoceros: that's open for discussion. IMO, keeping it around after removing the body of the function doesn't make much sense, so it should either be removed completely or restored to its real form for one more minor release.

yt/data_objects/static_output.py
neutrinoceros: yeah that should have been a GitHub comment. Thanks for pointing it out.

neutrinoceros:
I don't think this is worth keeping around anymore. The fact that satisfying vulture requires maintaining a whitelist *or* mandates that arguments, albeit unused, be renamed, is too high a cost for the expected benefit.

PR #3022, opened 2021-01-18
[bugfix] Prevent v1.3.1 of cftime (upstream dep)

cphyc:
There seems to be something fishy happening with cftime 1.3.1 on windows, trying to fix this.

PR #3024, opened 2021-01-18
Revert cftime pinning

cphyc:


neutrinoceros:
For context this is a direct reversal of #3022, which is not necessary anymore since cftime 1.3.1 now has wheels for windows, see https://github.com/Unidata/cftime/issues/222#issuecomment-762314002

Since the windows build passed, I see no point in waiting for the rest of test to complete.

neutrinoceros (approved):

PR #3025, opened 2021-01-19
ENH: uniformize mpl_kwargs

### LINKED ISSUE #547: Standardize arguments to plot annotations
Originally reported by: **Matt Turk (Bitbucket: [MatthewTurk](https://bitbucket.org/MatthewTurk), GitHub: [MatthewTurk](https://github.com/MatthewTurk))**

----------------------------------------

`plot_args` and `text_args` should only be one or the other.  Having multiple argument names is confusing.

----------------------------------------
- Bitbucket: https://bitbucket.org/yt_analysis/yt/issue/547

neutrinoceros:
## PR Summary

This is targeted at fixing  #547
The issue was left as "needing more discussion" 5 years ago, so I'm moving forward with a patch to revive or close it.

Here's the set of rules I'm enforcing throughout plot callbacks:
- dict-type arguments that is designed to pass arbitrary parameters down to a matplotlib object, should be named `mpl_kwargs`
- ... unless it's a text/box object, in which case, in which case it should be named `text_kwargs` and `inset_box_kwargs`
- another exception is `size_bar_kwargs` in scale annotation callback

The main reason for having "exceptions" here is that these arguments are typically used in the same functions so it's still the easiest way to handle them clearly.
Note that a large portion of the names I changed here originally had a `_args` suffix instead of `_kwargs`. I find that `kwargs` is clearer in that it's the de facto convention (in mpl) for passing arbitrary dicts of arguments down.

All changes performed here come with a deprecation warning so this shouldn't break backward compatibility despite the sheer number of them.

As a last note, I went with `mpl_kwargs` as a uniform varname, but `plot_kwargs` would be closer to the pre-existing convention (`plot_args`), I am fine with either, though I think the latter may turn out a little too restrictive, so my personal preference goes to the former.

Related:
- #2564
- #3007 (deprecation cycle)

fix #547

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -2,10 +2,12 @@ import re
 import warnings
 from functools import wraps
 from numbers import Number
+from typing import Any, Dict
 
 import matplotlib
 import numpy as np
 
+from yt._maintenance.deprecation import DEPRECATED_DEFAULT, issue_deprecation_warning
 from yt.data_objects.data_containers import YTDataContainer
 from yt.data_objects.level_sets.clump_handling import Clump
 from yt.data_objects.selection_objects.cut_region import YTCutRegion
@@ -29,6 +31,21 @@ from yt.utilities.on_demand_imports import NotAModule
 from yt.visualization.image_writer import apply_colormap
 
 callback_registry = {}
+
+
+def _handle_kwarg_deprecation_warning(
+    old: str, new: str, namespace: Dict[str, Any]
+) -> None:
+    # old and new should be variable names
+    # namespace is normally supposed to be `locals()`
+    old_val = namespace[old]
+    new_val = namespace[new]
+    if old_val is not DEPRECATED_DEFAULT:
+        new_val.update(old_val)
+        warning = "keyword `{}` is deprecated, use `{}` instead."
+        issue_deprecation_warning(
+            warning.format(old, new), removal="4.3.0", since="4.1.0"
+        )
 
 
 def _verify_geometry(func):
@@ -295,7 +312,7 @@ class VelocityCallback(PlotCallback):
     """
     Adds a 'quiver' plot of velocity to the plot, skipping all but
     every *factor* datapoint. *scale* is the data units per arrow
-    length unit using *scale_units* and *plot_args* allows you to
+    length unit using *scale_units* and *mpl_kwargs* allows you to
     pass in matplotlib arguments (see matplotlib.axes.Axes.quiver
     for more info). if *normalize* is True, the velocity fields
     will be scaled by their local (in-plane) length, allowing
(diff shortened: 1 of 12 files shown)


neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

munkm (commented):
I wonder if there is a way for us to maintain some sort of backwards compatibility that if an old kwarg dict name is passed that we could automatically fix the name and issue a warning? That might help make sure we have some level of continuity with scripts and also alert users to our new naming conventions without breaking their scripts?

munkm:
I see that @chummels was the original commenter on this issue. Do you think that using a standard mpl_kwargs makes things more confusing because they're pointing to different objects? (I just want to make sure i'm understanding your original comment!)

And thank you @neutrinoceros for resurrecting this old issue. 🙂 👻

neutrinoceros:
> we could automatically fix the name and issue a warning

I'm sorry I'm not following; how would that be different from what this is already doing?

Also, indeed there's a tradeoff to be dealt with and I see no ideal solution to @chummels ' original concern, but I do think it is worth uniformizing the api with the reasonable limitations I listed in the initial description.
Let me know what you think @chummels :)

matthewturk:
This is one of those times that defining an `Interface`-like object would be really handy, eh?

matthewturk (approved):
This looks good to me.

matthewturk:
@chummels can you comment on this?

neutrinoceros:
Ah. So many conflicts, I'll try to fix them now.

neutrinoceros:
Buildtime failures are the best failures... Switching to draft until I can resolve this.

neutrinoceros:
Stabilized !
@chummels all yours :)

neutrinoceros:
@yt-fido test this please

yt/visualization/plot_modifications.py
```diff
@@ -1041,19 +1073,27 @@ class LinePlotCallback(PlotCallback):
...
+            mpl_kwargs = {"color": "white", "linewidth": 2}
+        handle_kwarg_deprecation_warning("plot_args", "mpl_kwargs", locals())
+        self.mpl_kwargs = mpl_kwargs
         if data_coords:
             coord_system = "data"
-            warnings.warn(
-                "The data_coords keyword is deprecated.  Please set "
-                "the keyword coord_system='data' instead."
+            issue_deprecation_warning(
+                "`data_coords` keyword arguemnet is deprecated. "
```
chummels, line 1095:
```suggestion
                "`data_coords` keyword argument is deprecated. "
```

chummels (commented):
Nice work on a tricky subject here, @neutrinoceros .  I'm sort of +0.5 on this going in.  I realize it simplifies things to have slightly fewers `kwargs` options, but it could lead to some confusion by the users.  More below.

- I think that these changes should be clearly described in the yt4diffs doc, because this is going to throw a lot of warnings for people running old scripts.

- I think we need to make sure that the docstrings for each function explicitly say what the MPL object associated with these `mpl_kwargs` are (potentially even linking directly to the MPL docs page for it).  Because MPL kwargs are not universal from object to object.  if I'm a user trying to do a plot_annotation with let's say `ds.annotate_sphere()`, I need to explicitly know that the code is creating an MPL circle, so I know what kwargs are appropriate to use.  Previously, this was a bit easier to divine (but not foolproof!) because of the name of `circle_args` or whatever, but now it's totally hidden.

- This may potentially limit the number of MPL objects we can include in any single function call, since we can only have a single `mpl_kwargs` keyword for any function.  Notably, when we run [a callback like `annotate_scale()`](https://yt-project.org/docs/dev/visualizing/callbacks.html#add-a-physical-scale-bar) that has text as well as an inset box as well as a rectangle, we need to pass three separate kwarg dicts for each of those properties, which I assume is one of the reasons why you left `inset_box_kwargs` and `text_kwargs` in addition to `mpl_kwargs`.  We may encounter a time when we need to pass more `mpl_kwargs` to a new function for multiple MPL objects contained therein, in which case, we'll have to create new `kwargs` dicts, just as things are done today?  But maybe that's OK?  It just feels a little arbitrary to compress down some of these kwarg lists to `mpl_kwargs`, but I guess that's how it has to be?

neutrinoceros:
CI is now failing because warnings are treated as errors, which means I need to clean up tests using stuff that I'm deprecating here. I'll make this a draft in the mean time.

neutrinoceros:
I'm seeing some
```
TypeError: __init__() got an unexpected keyword argument 'mpl_args'
```
in the docs build (which weirdly shows a green nevertheless ?), but I haven't been able to locate where the errors were coming from for now, switching back to draft then...

neutrinoceros:
I'm now convinced I don't want to go through with this attempt, and I also don't want this to cause distraction on the mailing list for now, so I'll probably only bring this back in the future.

PR #3026, opened 2021-01-20
idiomatic python upgrade: Python 3.6

neutrinoceros:
## PR Summary

[pyupgrade](https://github.com/asottile/pyupgrade) is a versatile tool that automatically replaces "old" python idioms with their modern equivalent.
In particular it does what flynt did for us (enforcing strings everywhere it can) and much more (dict comprehension, old-style classes, removes extraneous `# coding: ...` cookies to name a few).
Here's a non exhaustive list of my previous PRs that could have been at least partially automated by this tool:
- #2640
- #2641 (not sure about this one)
- #2990 

Here I'm using it to replace every idiom from Python < 3.6
In the future (#2917), it will be trivial to incrementally modernise the code base by replacing `--py36-plus` with `--py37-plus` then `--py38-plus` and so on...

### DIFF
--- a/yt/utilities/sdf.py
+++ b/yt/utilities/sdf.py
@@ -413,8 +413,8 @@ class SDFRead(dict):
             for v in vnames:
                 str_types.append((v, vtype))
             l = ascfile.readline()
-        num = l.strip("}[]")
-        num = num.strip("\\;\\\n]")  # NOQA B005
+        spec_chars = r"{}[]\;\n\\"
+        num = l.strip(spec_chars)
         if len(num) == 0:
             # We need to compute the number of records.  The DataStruct will
             # handle this.
--- a/yt/utilities/sdf.py
+++ b/yt/utilities/sdf.py
@@ -230,7 +230,7 @@ class HTTPDataStruct(DataStruct):
     """docstring for HTTPDataStruct"""
 
     def __init__(self, *args, **kwargs):
-        super(HTTPDataStruct, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
         HTTPArray, PageCacheURL = get_thingking_deps()
         self.HTTPArray = HTTPArray
         self.pcu = PageCacheURL(self.filename)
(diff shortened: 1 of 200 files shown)


neutrinoceros:
converting to draft for now. I need to carefully review this my self, but I want to build confidence in the tool by running our tests first :)

yt/utilities/sdf.py
```diff
@@ -414,7 +414,7 @@ def parse_struct(self, line, ascfile):
                 str_types.append((v, vtype))
             l = ascfile.readline()
         num = l.strip("}[]")
-        num = num.strip("\;\\\n]")  # NOQA B005
+        num = num.strip("\\;\\\n]")  # NOQA B005
```
neutrinoceros, line 417: so far this is the only change that raises my eyebrow. This line is supposed to remove every occurence of `\`, `;` and `\n` from the `num`.
1) I probably wrote the NOQA comment myself, but I think it ought to be removed
2) I'm not sure why or how pyupgrade decided to edit this
3) This is also removing `]` occurrences, though this is redundant with the previous line so this part is a null op.

neutrinoceros (commented):
I'm halfway there for a full review. So far, only one changes looks fishy to me, and it may be just me. I'll finish this later today.

neutrinoceros (commented):
I cannot formally "approve" this myself since as the author though I merely enabled the changes rather than performed them, but I want to emphasise that I went through it all manually and did not find anything suspicious outside of my first half review.
pyupgrade generally helps with readability, especially in complex dict comprehensions where an expression can be reduced in volume by as much as 5 lines (mostly brackets).

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

matthewturk:
I approve of this.

matthewturk (approved):

matthewturk:
I think that the tests passing should be a sufficient check for large refactorings like this, along with a brief skim of the changes.

matthewturk (approved):

neutrinoceros:
well it wasn't brief, I'll tell you that, but it's *done*.
Also, I agree that all tests passing is more than likely enough bullet-proofing for this.

cphyc (approved):

PR #3028, opened 2021-01-20
[WIP] pyannotate -- create type annotations

cphyc:


matthewturk:
Do you think we should review this?

cphyc:
No I don't think you should. The code is not working, nor are the type annotations correct in many places. You can however have a global look at it to assess the amount of reviewing work adding type annotations to the entire code repository _at once_ would require.

FYI, in this PR I have had the type annotations generated using only RAMSES tests' (and all the functions that are subsequently accessed). In some sense, this PR only shows a fraction of all the places where type annotations should appear if we were to add them globally.

PR #3030, opened 2021-01-21
switch internal instances of collections.OrderedDict to regular dict

neutrinoceros:
## PR Summary

…when the differences do not matter any more: starting from Python 3.6, normal dict objects preserve insertion order.
I was originally aiming at replacing _all_ ordereddict by normal dict but I left every user-facing one untouched because _comparison_ is still different between the two classes, so I don't mean to cause bugs downwards where OrderedDict is possibly still relevant.

### DIFF
--- a/yt/data_objects/particle_trajectories.py
+++ b/yt/data_objects/particle_trajectories.py
@@ -1,5 +1,3 @@
-from collections import OrderedDict
-
 import numpy as np
 
 from yt.config import ytcfg
@@ -69,7 +67,6 @@ class ParticleTrajectories:
 
         if fields is None:
             fields = []
-        fields = list(OrderedDict.fromkeys(fields))
 
         if self.suppress_logging:
             old_level = int(ytcfg.get("yt", "loglevel"))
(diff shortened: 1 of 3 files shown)


yt/data_objects/particle_trajectories.py
```diff
@@ -69,7 +67,6 @@ def __init__(
 
         if fields is None:
             fields = []
-        fields = list(OrderedDict.fromkeys(fields))
```
neutrinoceros, line 72: This may not look obvious but the current implementation here is actually doing _nothing_, so I went with the option of preserving behaviour rather than intent. I think the original intent was to _sort_ the keys, in which case it should be simply reimplemented as
```python
fields = sorted(keys)
```
@jzuhone , do you remember this ? :)

cphyc (approved):

yt/data_objects/particle_trajectories.py
jzuhone: I think this is fine. I also think that someone came in here and did this later, because I don't recognize the code to be honest.

jzuhone (approved):

yt/data_objects/particle_trajectories.py
neutrinoceros: Cool, thanks for your prompt answer

PR #3031, opened 2021-01-22
deprecate confusing alias "mylog"

neutrinoceros:
## PR Summary

`yt.funcs.mylog` is a widely spread alias for `yt.utilities.logger.ytLogger`. I find that the name is a bit confusing for it's not immediately clear who "my" refers to (the library ? the module ? a single function or class ?). Additionally, the code base is currently not harmonised and we have several co-existing ways to import the same object internally

```python
from yt.funcs import mylog
from yt import mylog
```
This last one being mostly recommended from an external user's standpoint, so it needs to be dealt with accordingly with a proper deprecation process.

This PR proposes an internal harmonisation and a clean deprecation to avoid breaking downstream code in the next release.

### DIFF
--- a/yt/__init__.py
+++ b/yt/__init__.py
@@ -55,6 +55,7 @@ from yt.funcs import (
     is_root,
     is_sequence,
     memory_checker,
+    mylog,
     only_on_root,
     parallel_profile,
     print_tb,
@@ -78,7 +79,7 @@ from yt.units import (
     uvstack,
 )
 from yt.units.unit_object import define_unit
-from yt.utilities.logger import set_log_level, ytLogger
+from yt.utilities.logger import set_log_level
 
 frontends = _frontend_container()
 
--- a/yt/__init__.py
+++ b/yt/__init__.py
@@ -78,7 +78,7 @@ from yt.units import (
     uvstack,
 )
 from yt.units.unit_object import define_unit
-from yt.utilities.logger import set_log_level, ytLogger as mylog
+from yt.utilities.logger import set_log_level, ytLogger
 
 frontends = _frontend_container()
 
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -19,7 +19,7 @@ from yt.data_objects.selection_objects.data_selection_objects import (
 )
 from yt.fields.field_exceptions import NeedsGridType, NeedsOriginalGrid
 from yt.frontends.sph.data_structures import ParticleDataset
-from yt.funcs import get_memory_usage, is_sequence, iter_fields, mylog, only_on_root
+from yt.funcs import get_memory_usage, is_sequence, iter_fields, only_on_root, ytLogger
 from yt.geometry import particle_deposit as particle_deposit
 from yt.geometry.coordinates.cartesian_coordinates import all_data
 from yt.loaders import load_uniform_grid
@@ -288,12 +288,12 @@ class YTProj(YTSelectionContainer2D):
         # Now we run the finalizer, which is ignored if we don't need it
         field_data = np.hsplit(data.pop("fields"), len(fields))
         for fi, field in enumerate(fields):
-            mylog.debug("Setting field %s", field)
+            ytLogger.debug("Setting field %s", field)
             input_units = self._projected_units[field]
             self[field] = self.ds.arr(field_data[fi].ravel(), input_units)
         for i in list(data.keys()):
             self[i] = data.pop(i)
-        mylog.info("Projection completed")
+        ytLogger.info("Projection completed")
         self.tree = tree
 
     def to_pw(self, fields=None, center="c", width=None, origin="center-window"):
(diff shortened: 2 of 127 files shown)


brittonsmith:
I don't mind harmonizing the imports, but I am not in favor of renaming the instances of `mylog`. It may just be a question of taste, but I find `mylog` to be very intuitive and use it frequently in my analysis scripts. This is an unnecessary change in my opinion.

neutrinoceros:
I for one feel like the `my` prefix is very much a signal for a "hidden" or "protected" variable (which Python lacks) that shouldn't be used outside of the code base itself, but admittedly this PR is opinionated, and I don't mean to impose my views.

neutrinoceros:
@brittonsmith actually, could you please clarify what you need the logger exposed for ? I was thinking the only reason to import it was to call `mylog.setLevel()ˋ, which can be done now with `yt.set_logger_level()`, maybe I'm just missing something ?

matthewturk:
So!  I am of two minds.  Were we to do it again, yeah, maybe `my` is bad.  But I am reasonably -- although not completely! -- sure that I named it this because I was following the example from the PyMOTW post from 2007 where I first learned about it: https://pymotw.com/2/logging/ .

Anyway, I'm ambivalent, but I think we should maybe make a list of things to change, and then pull the bandaid off all at once, and this should go on that list.

jzuhone:
> @brittonsmith actually, could you please clarify what you need the logger exposed for ? I was thinking the only reason to import it was to call `mylog.setLevel()ˋ, which can be done now with `yt.set_logger_level()`, maybe I'm just missing something ?

I use `mylog` in yt-based scripts when I want to log something. Likely this is the case for a number of users, so there is a potential to break a lot of stuff downstream. At the very least we need to warn people about it.

neutrinoceros:
> At the very least we need to warn people about it.

Agreed @jzuhone ! Note that in its current state, this PR doesn't break anything but adds a depreciation warning on the user side :)

However I admit I didn't think about it being used to log stuff from outside applications, and the message reflects my own misunderstanding.

In any case I was thinking this could affect a large number of users so I proposed to keep the alias for 2 minor releases (4.0 and 4.1) and only remove it in 4.2 . This information is _also_ included in the proposed deprecation message.

> Anyway, I'm ambivalent, but I think we should maybe make a list of things to change, and then pull the bandaid off all at once, and this should go on that list.

It's been on my list for a while for this very reason, but I feel more comfortable about it now that we have a mechanism to warn users an arbitrary number of releases before we actually break anything.

brittonsmith:
Breaking user scripts was a big concern for me as well. I use it like @jzuhone as parallel-friendly log output in scripts and other little packages. This is my own bias, but I've interpreted the my in `mylog` to mean accessible to me (either developer or user) and I think it being importable `from yt`is also suggestive of this.

I could live with an extended deprecation that removes it in a 4.2 release, although I think I'd prefer something shorter than `ytLogger`, like `ytlog`.

Perhaps this next statement is too much discussion creep, but I think at this point we have worn out our license to continue breaking backward compatibility prior to the 4.0 release. We've been suggesting that users adopt this version of the code for more than a year now. Barring necessary exceptions, I think we need to treat it like it's been released or hurry up and release it.

neutrinoceros:
> This is my own bias, but I've interpreted the my in mylog to mean accessible to me (either developer or user) and I think it being importable from yt is also suggestive of this.

That's my starting point exactly, the difference being that I don't think it _should_ be the case.

> I could live with an extended deprecation that removes it in a 4.2 release, although I think I'd prefer something shorter than `ytLogger`, like `ytlog`.

I have no preference between those two, but changing the actual variable name would cause even more breakage.
Besides, no one can stop users from aliasing it at import time
```python
from yt import ytLogger as ytlog
```
Would this work for you ?

> I think at this point we have worn out our license to continue breaking backward compatibility prior to the 4.0 release. (...) Barring necessary exceptions, I think we need to treat it like it's been released or hurry up and release it.

Agreed. If it makes you feel any more confortable I would advise to keep this change for next time (4.1) and remove `mylog` alias only in 4.3

munkm:
I guess I don't see how this deprecation improves a whole lot for us, but it does cause a *lot* of issues for our downstream users. I personally support deprecating features when we're replacing them with things that improve the yt codebase by adding new features or by large refactoring that helps make the code more accessible for new contributors. While the logging function could have aa better name if we were starting a new project, this rename seems like we're going to cause a lot of issues because of preferences that only benefit longtime devs of the project. This is a good way for us to alienate users that will see little benefit from the logging change, even if it benefits developers.

neutrinoceros:
Fine. This has been bugging me for a while but it's clearly causing much more annoyance than resolution. Thank you all for your feedback; I'll close it now :-)

munkm:
I'm sorry, I don't mean to be demotivating. I just think that if we have two experienced devs/users saying that they use mylog in their scripts, that means that we probably have a *lot* of users in that position too. I don't doubt your annoyance or that this could be better, but I do want to make sure we're not burning users unnecessarily.

neutrinoceros:
I understand, no worries !

PR #3032, opened 2021-01-22
docs: apply black format

neutrinoceros:
## PR Summary

... and stabilise it with a pre-commit hook
This would have been included in [YTEP 0037](https://ytep.readthedocs.io/en/master/YTEPs/YTEP-0037.html) had I known about it.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -6,6 +6,7 @@
 
 # transition to black
 ebadee629414aed2c7b6526e22a419205329ec38
+6f0e37c4b926f1411b14bd35da85c7e7e27d0cd9
 
 # automated trailing whitespace removal
 3ee548b04a41dfbc009921c492fba6a0682651ca
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,3 +41,8 @@ repos:
     hooks:
     - id: flake8
       additional_dependencies: [mccabe, flake8-bugbear]
+-   repo: https://github.com/asottile/blacken-docs
+    rev: v1.9.1
+    hooks:
+    -   id: blacken-docs
+        additional_dependencies: [black==20.8b1]
--- a/doc/source/analyzing/domain_analysis/clump_finding.rst
+++ b/doc/source/analyzing/domain_analysis/clump_finding.rst
@@ -29,8 +29,7 @@ acts as the base for clump finding.
 
    ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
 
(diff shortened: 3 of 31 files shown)


neutrinoceros:
Oops, looks like this particular hook is very unstable and fails on its own outputs. Closing.

neutrinoceros:
So actually... the tool isn't broken, but our _doc_ is, because it contains several syntax errors, I'll reopen after I push a fix for those !

neutrinoceros:
I think the mere fact that I was able to discover problems within the docs demonstrates how useful this tool is :)

matthewturk (approved):

munkm (approved):
I think this is great (and thank you for fixing the errors). 

One thing that might enhance our dev docs (which you can do on this PR or separately) is that we should consider listing what tools are being run by the bot and would be installed with the pre-commit install. Our section in the docs https://yt-project.org/docs/dev/developing/developing.html#automatically-checking-and-fixing-code-style that talks about these tools is helpful, but at this point we have a few that are run and it might be helpful for future contributors to know what types of checks are going to be run.

neutrinoceros:
I've been thinking about documenting them but it seems to me that it would generate unnecessary maintenance burden.

>  would be installed with the pre-commit install 

I'm not sure this clarification is needed but I'll still run with it: `pre-commit install` in fact installs the hooks in isolation (I think they are not even installed together but each one gets a virtual env of its own) so they are never in conflict with the user's env, but it also means they are not available outside of the framework.

The cool thing is that `.pre-commit-config.yaml` is naturally self-documenting because it necessarily refers to repositories by url, so I think the easiest way to address your concern would be to document that the exact tool bet is described in this file. WDYT ?

munkm:
So the reason I suggested the change was because I went looking in our docs and I was surprised that even though we are using black for our code (and if this PR is merged, our docs) we have no documentation of that in our development guide. We don't have to describe in detail what is installed or what we're using -- I just think a bulleted list that describes our tooling and what they're used for would help enhance that part of the documentation. Something like this:

* **black** is used for code style and formatting consistency 
* **isort** is used to ensure imports are consistently ordered
etc. etc. 


Also, this way a potential dev doesn't need to get in the weeds in the dev process to know what to expect. Does that make sense? 
I hear you on the maintenance burden, but I (personally) think it's better to have out of date docs than no docs for things and documentation is a really good item for a first time contributor to contribute to yt with!

munkm:
Anyways, I don't see any reason for that to hold up this PR! I just wanted to raise the issue with our docs. 🙂

PR #3035, opened 2021-01-25
Add possibility to silence some warnings

cphyc:
## PR Summary

As devs, we may want to silence some warnings when developing. For example, we may want to ignore the warning that both `~/.config/yt/ytrc` and `~/.config/yt/yt.toml` exist when working on both yt 3 and yt 4.

This PR adds supports for adding a list of warnings to simply ignore for devs convenience. I did *not* write any doc for this, as it is intended for devs only (otherwise users may simply ignore all warnings).

To use it, find the warning that is getting on your nerve, and save it in your config using
```toml
[yt.developers]
ignore_warnings = ["config:both_file_present"]
```

### DIFF
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -304,6 +304,7 @@ class PlotContainer:
             issue_deprecation_warning(
                 "Deprecated api, use bools for *state*.",
                 deprecation_id="plot_container:bools_for_state",
+                since="4.0.0",
                 removal="4.1.0",
             )
             state = {"on": True, "off": False}[state.lower()]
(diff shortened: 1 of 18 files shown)


matthewturk:
Seems good to me!

matthewturk (approved):

yt/visualization/plot_container.py
```diff
@@ -302,7 +302,9 @@ def set_minorticks(self, field, state):
         """
         if isinstance(state, str):
             issue_deprecation_warning(
-                "Deprecated api, use bools for *state*.", removal="4.1.0"
+                "Deprecated api, use bools for *state*.",
+                deprecation_id="plot_container:bools_for_state",
+                removal="4.1.0",
```
neutrinoceros, line 307:
```suggestion
                deprecation_id="plot_container:bools_for_state",
                since="4.0.0",
                removal="4.1.0",
```

neutrinoceros (approved):

PR #3038, opened 2021-01-26
exp: remove DatasetSeries.__iter__

neutrinoceros:
## PR Summary

I am under the impression that this method could be simplified... to the point of not being necessary any more.
Trying to naively remove it and see what breaks.
edit: nothing ! 🎉

### DIFF
--- a/yt/data_objects/time_series.py
+++ b/yt/data_objects/time_series.py
@@ -201,16 +201,6 @@ class DatasetSeries:
             raise FileNotFoundError(f"No match found for pattern : {pattern}")
         return sorted(file_list)
 
-    def __iter__(self):
-        # We can make this fancier, but this works
-        for o in self._pre_outputs:
-            try:
-                ds = self._load(o, **self.kwargs)
-                self._setup_function(ds)
-                yield ds
-            except TypeError:
-                yield o
-
     def __getitem__(self, key):
         if isinstance(key, slice):
             if isinstance(key.start, float):


cphyc:
Wait, won't it prevent users from doing this?
```python
ts = yt.load("/path/to/timeseries")
for ds in ts:
    ...
```

neutrinoceros:
In fact it _doesn't_ (just tested this again), that's the whole point: the mere fact that `DatasetSeries.__getitem__` is implemented does the job. You can convince yourself with this simple snippet:

```python
class A:
    def __init__(self):
        self.data = '123'
    def __getitem__(self, key):
        return self.data[key]

for a in A():
    print(a)
```

with correctly prints
```
1
2
3
```

I don't know for sure why we still have __iter__, maybe it's a legacy from old school python 2 classes ?

matthewturk:
This is mind-blowing.

neutrinoceros:
just for extra safety, I also tested this
```python

from time import sleep

class A:
    def __init__(self):
        self.data = '123'
    def __getitem__(self, key):
        return self.data[key]

for a in A():
    print(a, flush=True)
```
and the result is also correct: we get a 1s delay between each print, not a 3s delay followed by 3 prints. The implication is that this doesn't affect performance (to the best of my knowledge), since items are still correctly yielded one at a time.

brittonsmith:
Wow, learned something!

neutrinoceros:
Thanks for looking it up, I actually have no idea why I _thought_ about trying this because I don't remember reading about it. I feel much better with this source (and wow, it's actually much older than Python 3.0).

munkm:
That `__getitem__` is much older than even `__iter__`?

matthewturk (approved):

munkm:
I didn't get to fully review this, but I guess I was wondering why we were deleting the stuff in `__iter__` and falling back to `__getitem__` when it reads more clearly in the code to have everything in `__iter__`? When I was reading up on the difference between the two it seemed like `__iter__` could be used to do more sophisticated things.

matthewturk:
I guess I got a little merge-happy.  I'm sorry for jumping the gun on this.

neutrinoceros:
> I didn't get to fully review this, but I guess I was wondering why we were deleting the stuff in __iter__ and falling back to __getitem__ when it reads more clearly in the code to have everything in __iter__?

I guess I forgot to mention my original motivation for doing this: __iter__ and __getitem__ contained duplicated code that needed to be changed in #3037; I figured it could be removed to avoid maintaining twice as much buggy code in the future.

PR #3039, opened 2021-01-26
bugfix/import-cycle

cphyc:
## PR Checklist


neutrinoceros (changes requested):
This deprecation helper function is turning into hardly maintainable _spaghetti_ code. I would love to have a better solution

cphyc:
Agreed, I think we should start by reverting #3035 and think whether this is really worth the trouble.

PR #3040, opened 2021-01-26
Revert "Add possibility to silence some warnings"

cphyc:
Reverts yt-project/yt#3035

This created import cycles that are hard to fix without spaghetti code (see #3039).


neutrinoceros:
Not ideal but I like this better than the first attempt. imo this functionality stops being worth it the second it turns into technical debt :/

neutrinoceros (approved):

munkm (approved):

munkm:
> Not ideal but I like this better than the first attempt. 

I agree. This seems like a good interim solution!

PR #3041, opened 2021-01-27
Set _periodicity to avoid deprecated setter.

brittonsmith:
This fixes the `ytdata` frontend to set `self._periodicity` instead of `self.periodicity`, for which the setter function is deprecated.

## PR Checklist


neutrinoceros (approved):
sorry about not catching this earlier, and thanks for fixing it !

matthewturk (approved):

PR #3044, opened 2021-01-27
Migration script: migrate with user input

cphyc:
## PR Summary

At the moment, if a user has `~/.config/yt/ytrc` but no `~/.config/yt/yt.toml`, a hard-exit happens which prevent yt from being used. Though we display an explicit message advising the user on how to fix the issue, we may give the user a chance to interactively perform the migration immediately to lower the friction of this transition.

### DIFF
--- a/yt/config.py
+++ b/yt/config.py
@@ -6,7 +6,6 @@ from more_itertools import always_iterable
 
 from yt._maintenance.deprecation import issue_deprecation_warning
 from yt.utilities.configuration_tree import ConfigNode
-from yt.utilities.configure import interactive_migrate_config
 
 ytcfg_defaults = {}
 
@@ -210,6 +209,8 @@ if os.path.exists(OLD_CONFIG_FILE):
                 since="4.0.0",
                 removal="4.1.0",
             )
+            from yt.utilities.configure import interactive_migrate_config
+
             interactive_migrate_config()
 
 
--- a/yt/config.py
+++ b/yt/config.py
@@ -6,6 +6,7 @@ from more_itertools import always_iterable
 
 from yt._maintenance.deprecation import issue_deprecation_warning
 from yt.utilities.configuration_tree import ConfigNode
+from yt.utilities.configure import interactive_migrate_config
 
 ytcfg_defaults = {}
 
@@ -209,7 +210,7 @@ if os.path.exists(OLD_CONFIG_FILE):
                 since="4.0.0",
                 removal="4.1.0",
             )
-            raise SystemExit
+            interactive_migrate_config()
 
 
 if not os.path.exists(_global_config_file):
--- a/yt/utilities/configure.py
+++ b/yt/utilities/configure.py
@@ -47,6 +47,19 @@ def write_config(config_file):
     CONFIG.write(config_file)
 
 
+def interactive_migrate_config():
+    prompt = "Perform the migration now [yn]? "
+    user_input = input(prompt).lower()
+    while user_input not in ("y", "yes", "n", "no"):
+        print(f"Did not understand your input '{user_input}'. Please enter 'y' or 'n'.")
+        user_input = input(prompt).lower()
+
+    if user_input in ("y", "yes"):
+        migrate_config()
+    else:
+        raise SystemExit("Migration not performed: exiting.")
+
+
 def migrate_config():
     if not os.path.exists(OLD_CONFIG_FILE):
         print("Old config not found.")


neutrinoceros (approved):

Xarthisius:
If I updated yt on a cluster and run a batch job, I'd be very unhappy that my CPUh were spent on waiting on `input()`. I'd be infinitely less unhappy if it died right away.

matthewturk:
Ok, this is an extremely good point. Let's brainstorm solutions.

Xarthisius:
Can we just do it like we do with everything else? Start with deprecation warning over the course of one release, then die if not fixed after next + 1 release?

neutrinoceros:
The point was to protect users from ignoring the warning + wondering what went wrong later. In all honesty, I'm reconsidering if it's even worth it. Thanks @Xarthisius for bringing this up. It actually happened to me once and I confirm it didn't make me a fan of the code in question...

Can we automatically perform the migration BUT keep the original file without appending ".bak" to it ?
Then both confit files would be present and we'd raise a non blocking warning at load time until the old one is (re)moved.

cphyc:
I have made a recap of all the solutions that emerged out of the discussion. Please feel free to edit the pros/cons so that we converge relatively quickly.

| Solution | Pros | Cons
|-----|----|----|
| Stop early with a `raise SystemExit` (current behaviour) | Ensures users perform the migration | We may receive many complaints about yt "not working" |
| Raise warning then error in yt 4.0.1 | Transition is smooth for the users | Some users may ignore the warning and not perform the migration, yet still think `~/.config/yt/ytrc` is their config file |
| Raise warning, migrate automatically | Transition is even smoother for the users | We're modifying on-disk files without user's consent |
| Raise warning, migrate automatically after confirmation | Transition is smooth | Will result in annoying behaviour if launched on a non-interactive environment (ex. on a cluster) as we're using for user input |

neutrinoceros:
I voted 🎉 because the "cons" is actually aligned with yt's old behaviour: we used to create a ytrc file when none was found, BUT I actually don't like this "feature" very much (for the exact reason you're mentioning).
I think it is reasonable to migrate automatically provided two restrictions are followed
1) do not write to yt.toml if the file already exists
2) do not _create_ a config file ex-nihilo, only do it if the old one is present.

Again, if this turns into a monstrosity of maintainability, my second preferred option is "👍🏻"

cphyc:
Regarding

> I voted tada because the "cons" is actually aligned with yt's old behaviour: we used to create a ytrc file when none was found, BUT I actually don't like this "feature" very much (for the exact reason you're mentioning).
> I think it is reasonable to migrate automatically provided two restrictions are followed
> 
>     1. do not write to yt.toml if the file already exists
> 
>     2. do not _create_ a config file ex-nihilo, only do it if the old one is present.
> 
> 
> Again, if this turns into a monstrosity of maintainability, my second preferred option is "rocket"

Regarding your points:

1. currently, the error is only triggered iff `ytcfg` exists but `yt.toml` does not. If both are present, there's only a warning.
2. this would imply a larger change (but why not) as currently yt creates an empty config file if none is found. Note that this behaviour is only here because I left it in #2981, but is not necessary _per se_. Another point: the current issue is only affecting users who aren't new to yt. They should all have a `ytrc` file since it is created automatically. For new users, there's no problem as they don't have a `ytrc` file.

neutrinoceros:
>1. currently, the error is only triggered if ytcfg exists but yt.toml does not. If both are present, there's only a warning.

excellent

>2. the current issue is only affecting users who aren't new to yt. They should all have a ytrc file since it is created automatically. For new users, there's no problem as they don't have a ytrc file.

Fair enough, I didn't connect the dots.

matthewturk:
Are we also putting folks into a weird spot if they have both a 3.x and 4.x installation present?

cphyc:
> Are we also putting folks into a weird spot if they have both a 3.x and 4.x installation present?

Currently, the only problematic case is when `~/.config/yt/ytrc` exists, `~/.config/yt/yt.toml` does not *and* users are on the `main` branch of yt.

If both config files exists, there is a warning and nothing else.

matthewturk:
OK, that actually sounds like an OK solution to having both, to me...

cphyc:
@matthewturk @Xarthisius @neutrinoceros I have created two other PRs (#3047 and #3046) that implement the automatic migration and the simple warning solution respectively. Could you have a look and tell me which one should be kept?

neutrinoceros:
#3047 is my preferred option, given some further changes.

munkm:
It seems like 🚀 is winning the poll?

neutrinoceros:
Yup. As the only advocate for a different option I gracefully resign. I think the poll was open long enough that everyone who wanted to vote already did. Moving forward ? :)

munkm:
I think 🚀 was still your second choice option though! 

> Moving forward ? 

I support it!

neutrinoceros:
It was. That's why I'm _gracefully_ resigning, instead of being this guy.
image

cphyc:
Closing this now in favour of https://github.com/yt-project/yt/pull/3046.

cphyc:
Closing this now in favour of https://github.com/yt-project/yt/pull/3046.

neutrinoceros (commented):
Oops, forgot to push the "submit" button yesterday. Might as well publish this review even if the PR is dead

Issue #3045, opened 2021-01-28
Unnecessary config file creation


neutrinoceros:
### Bug report

As discussed in  #3044, we are still creating a yt config file at import time when none is already available, _but_ this is an implementation detail inherited from before we switch to toml config files and apparently is not necessary any more (according to @cphyc).
I think there is no good reason to keep this behaviour now.

Additionally, this behaviour makes working with a local configuration file pretty clunky. For instance if one doesn't have a global file (yet), but runs
```
echo "[yt]\nserialize = false" > yt.toml
yt config list
```
they get
```
Yt detected a local and a global configuration file, refusing to proceed.
Local config file: /Users/robcleme/dev/yt-project/yt/yt.toml
Global config file: /Users/robcleme/.config/yt/yt.toml
Specify which one you want to use using the `--local` or the `--global` flags.
```

PR #3046, opened 2021-01-28
Migration script: warn and do not migrate

cphyc:
Related to #3044.

Here is the behaviour
- on yt 4, `yt.toml` and `ytcfg` are present: raise a warning
- on yt 4, `yt.toml` does not exist, `ytcfg` exists: raise another warning and **do not migrate**
- in any case, a `yt.toml` will be created (see #3045)

### DIFF
--- a/yt/config.py
+++ b/yt/config.py
@@ -209,7 +209,6 @@ if os.path.exists(OLD_CONFIG_FILE):
                 since="4.0.0",
                 removal="4.1.0",
             )
-            raise SystemExit
 
 
 if not os.path.exists(_global_config_file):
--- a/yt/config.py
+++ b/yt/config.py
@@ -190,25 +190,13 @@ if os.path.exists(OLD_CONFIG_FILE):
             removal="4.1.0",
         )
     else:
-        # We have an issue here: when calling from the command line,
-        # we do not want this to exit, as it would prevent `yt config migrate`
-        # from running. The issue is that yt.config (this file) is imported
-        # from yt.__init__, which is imported *before* yt.utilities.configure
-        # is executed, so the latter cannot set some internal variable before
-        # we arrive here.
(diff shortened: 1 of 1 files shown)


neutrinoceros (approved):
Just a remark: it is not clear from the warning message wether the deprecated config file is being loaded or ignored in the CLI call case. Otherwise, LGTM

yt/config.py
```diff
@@ -190,26 +190,13 @@ def get_local_config_file():
...
-        if len(stack) < 2 or stack[-2].function != "importlib_load_entry_point":
-            issue_deprecation_warning(
-                f"The configuration file {OLD_CONFIG_FILE} is deprecated. "
-                f"Please migrate your config to {_global_config_file} by running: "
-                "'yt config migrate'",
-                since="4.0.0",
-                removal="4.1.0",
-            )
-            raise SystemExit
+        issue_deprecation_warning(
```
cphyc, line 193: The reason we can simplify this is that there is no circular import issue any more AND no errors. The slight caveat is that now `yt config migrate` will advise users ... to use `yt config migrate`!
```bash
$ yt config migrate
[…]/yt/yt/utilities/logger‧py:4: VisibleDeprecationWarning: The configuration file […]/.config/yt/ytrc is deprecated. Please migrate your config to […]/.config/yt/yt‧toml by running: 'yt config migrate'
Deprecated since v4.0.0 . This feature will be removed in v4.1.0
  from yt‧config import ytcfg
INFO: using configuration file: […]/.config/yt/yt‧toml.
Writing a new config file to: […]/.config/yt/yt‧toml
Backing up the old config file: […]/.config/yt/ytrc‧bak
```
Note that this shouldn't be an issue as an subsequent calls to `yt config migrate`  won't show the warning and won't do anything (as the config has already been migrated).

munkm (approved):

PR #3047, opened 2021-01-28
Migration script: migrate automatically

cphyc:
Related to #3044.

Here is the behaviour
- on yt 4, `yt.toml` and `ytcfg` are present: raise a warning
- on yt 4, `yt.toml` does not exist, `ytcfg` exists: raise another warning and **do the migration automatically** without confirmation
- in any case, a `yt.toml` will be created (see #3045)

### DIFF
--- a/yt/config.py
+++ b/yt/config.py
@@ -209,7 +209,6 @@ if os.path.exists(OLD_CONFIG_FILE):
                 since="4.0.0",
                 removal="4.1.0",
             )
-            raise SystemExit
 
 
 if not os.path.exists(_global_config_file):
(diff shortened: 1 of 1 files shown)


neutrinoceros:
As I said, I think this route is preferable, with one important condition:
We should note move the old config file ourselves. Let the user do it. This means the migration function should be updated to conform to this requirement, and the warning messages should acknowledge it (in particular the one triggered in case both the old and the new file exist).

cphyc:
Closing this for now following https://github.com/yt-project/yt/pull/3044#issuecomment-768925446.

PR #3052, opened 2021-01-31
ci: restrain build-test workflow to pull request trigger

neutrinoceros:
## PR Summary

This avoids having failures reported from both personal forks and the main repo.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1,6 +1,6 @@
 name: Build and Test
 
-on: [push, pull_request]
+on: [pull_request]
 
 defaults:
   run:


matthewturk:
Hmm, doesn't this mean that if we do a direct push it won't test it?

neutrinoceros:
If you push against an existing PR tests will still run (just checked with the h5py workflow that relies only on the pull_request trigger)

munkm:
I just checked the workflows documentation and it seems if the PR is "synchronized" the PR workflow will be triggered. 

However, it also seems that the gh workflows won't be triggered by forked repositories by default, and must be enabled manually by users in the actions tab of their personal fork? Were we having issues with lots of personal forks?

neutrinoceros:
>  Were we having issues with lots of personal forks?

@munkm this may be just me but I'm used to opening PRs as soon as I publish a branch so I very frequently get double-notified when the same workflow fails on my fork _and_ here.

munkm:
Ok so the use case that this would effect is if somebody wants to test their local branches with actions (assuming they are enabled on their fork) by pushing to their branch but they haven't opened up a PR yet.

munkm:
I've been thinking about this more and I will say that my workflow is a bit different than yours -- I don't open a PR right away usually. In some cases we do have side projects that are large that exist on a fork until they're in a later stage (I think the dask stuff could be a good example here). I think it could be worth leaving on the push trigger for those cases?

neutrinoceros:
Possibly, but I'd argue that this leaves no possibility for a workflow with low notification noise when a specific task is hard or complex enough that it goes through a large number of iterations where some tests are temporarily broken.

neutrinoceros:
I take that back, I'm just now realising, as you actually hinted earlier, that I can manually turn actions off on my fork. I don't feel like this PR is necessary anymore, closing.

PR #3053, opened 2021-01-31
maint: remove legacy mercurial (hg) supported files and functions

neutrinoceros:
## PR Summary

These files and functions don't seem to be used (or usable) anymore.


matthewturk:
I'd actually prefer if we kept `.hgchurn` (maybe?) and `.hgtags`, for now at least since they include reference info, but the rest looks good.

neutrinoceros:
@matthewturk  you got it

brittonsmith (approved):

yt/funcs.py
```diff
@@ -774,38 +703,6 @@ def bb_apicall(endpoint, data, use_pass=True):
...
-            )
-            print(f"{supp_path}")
-            print()
-            sys.exit(1)
-        rv = hglib.clone("http://bitbucket.org/yt_analysis/yt-supplemental/", supp_path)
-        if rv:
-            print("Something has gone wrong.  Quitting.")
-            sys.exit(1)
-    # Now we think we have our supplemental repository.
-    return supp_path
```
munkm, line 806: I know this repo isn't carried in the github project org, but I'm curious what the legacy is from it @matthewturk or @brittonsmith?

matthewturk: OK, that takes me back.  I had to look it up.  Last commit I see was 2012, but `yt-supplemental` was a set of subrepos that included some things that didn't quite belong in yt proper.

```
yt-doc = http://hg.enzotools.org/yt-doc
hgbb = http://BitBucket.org/MatthewTurk/hgbb
pasteboard_template = http://BitBucket.org/MatthewTurk/pasteboard_template
cedit = http://bitbucket.org/MatthewTurk/cedit
yt-advanced-plotting = http://bitbucket.org/samskillman/plot_modifications
```

So it was my fork of the `hgbb` extension to make it easier to manage bitbucket (kind of like a really lightweight `hub` or `gh` command), a `pasteboard` for sharing scripts (never totally worked but it was fun), the docs, a configuration editing extension for hg, and @samskillman 's advanced plotting code that fiddled with plotting settings.

munkm: Very cool!!! Maybe we should revisit some of these things (like the advanced plotting stuff)? 

So it seems like it's very likely this isn't used by anybody? Or should this be in a deprecation cycle instead? Or do we even want to remove it?

matthewturk:
We should remove it; I don't think it's used.

munkm (approved):

yt/funcs.py
neutrinoceros: the urls are dead atm, which is why I'm proposing to delete this. Of course if the material is still available elsewhere (and still compatible), it should probably be kept and updated instead.

munkm: I totally support deletion! I didn't realize they were dead -- and it seems likely that this is unused (I didn't see any uses outside of forks) so I think it makes sense totally remove. It might be cool later on to revisit adding in "advanced" modules  in a formalized way for the project.

PR #3054, opened 2021-02-01
ci: install wheel as part of our gh build workflows

neutrinoceros:
## PR Summary

Attempt to circumvent breakage from just-released dependencies with no wheels available. Currently `cftime` 1.4.0 is breaking our CI because they don't have windows wheels yet.

### DIFF
--- a/tests/test_prerequirements.txt
+++ b/tests/test_prerequirements.txt
@@ -1,3 +1,7 @@
 # We need this file mostly because of Cartopy..
 numpy>=1.19.4
 cython>=0.29.21
+
+# this is sometimes useful to avoid CI breakage when a dependency release comes out
+# but some wheels (typically windows) are missing for a few hours/days
+wheel


neutrinoceros:
Now _answer_ tests are failing on ubuntu. I believe this to be related to the latest bumping of the answer store, which was updated following a numpy release. Since yet another numpy was released yesterday and the diff in the failing test looks very similar to last time, I interpret this as a bugfix on their side. Now trying to backpedal on the answer store.

matthewturk:
@neutrinoceros Can you clarify the specifics of where the answer test changed?  I'm hesitant to do any reversions until we're sure of the specifics.

neutrinoceros:
@matthewturk here's the diff http://use.yt/upload/46f5164

matthewturk:
Oh no, 404!

neutrinoceros:
Alright here's a screenshot
Screen Shot 2021-02-01 at 12 16 04

anyway the backbedaling didn't solve it. I still think the change comes indirectly from numpy  but there might be something fragile on our side too.

munkm:
So this PR is adding a comment but not installing a wheel now? Or did something get mixed up in pushing?

neutrinoceros:
@munkm it didn't get mixed up. The comment takes most of the space but there _is_ an actual addition: installing the `wheel` package itself so that next time our CI tries to install the latest version of a dep when no some wheels are still unavailable, we'll be able to fall back on building them ourselves instead of failing.

munkm (approved):
I'm trying to think if there would be any negative consequences of adding this. I can't think of any (?), and I think it's a good idea to put this in prerequirements so users don't install it when they're trying to run tests locally.

cphyc (approved):

PR #3056, opened 2021-02-04
doc: replace dead travis badge with corresponding GH workflow build status

neutrinoceros:
This will fail test for the same reason I discovered it was necessary: the main branch is currently broken against our test suite but the build badge reports that it's all good... because it hasn't been updated since we dropped travis.


neutrinoceros:
Weirdly enough THIS build didn't fail. 🙃

PR #3057, opened 2021-02-04
doc: specificly target main branch in gh badge

neutrinoceros:
## PR Summary

a quick follow up to #3056

PR #3058, opened 2021-02-05
doc: add badges for isort, precommit-ci, and minimal supported python

neutrinoceros:
## PR Summary

And rearrange badges over two lines.
This is another attempt than #2976 to acknowledge our minimal supported Python version. This time with a custom static badge since pypi releases may often be out-of sync with the main branch, as is the case right now.

this also includes the change from #3057

Here's how it looks
## current state
Screen Shot 2021-02-05 at 12 46 35
## this branch
Screen Shot 2021-02-05 at 12 45 57

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,20 @@
 # The yt Project
 
+![Supported Python Version](https://img.shields.io/badge/python%20version-≥%203.6-important)
+[![Latest Documentation](https://img.shields.io/badge/docs-latest-brightgreen.svg)](http://yt-project.org/docs/dev/)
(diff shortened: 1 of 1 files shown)


munkm (approved):
I think this looks good! If we end up getting to cluttered we can remove the black and isort badges sometime down the road. 

I still think it's worth adding black, isort, etc. to our dev documentation.

neutrinoceros:
> I still think it's worth adding black, isort, etc. to our dev documentation.

@munkm I'll open a PR for this shortly :)

Issue #2647, opened 2020-06-19
Missing machinery for frontend-specific deprecated field names


neutrinoceros:
### Report

In the discussion around #2613, @brittonsmith proposed that we add a mechanism to attach deprecated field names to specific frontends. Currently all deprecated field names are stored in `yt/fields/field_aliases.py`.

It seems likely to me that _all_ field names collected in this file are in fact meant to support a specific frontend, and maybe it they can be traced back with git blame.

neutrinoceros:
Note that the new yt/_maintenance/deprecation.py module should be home to any internal functions and classes this would require.

neutrinoceros:
I believe this was fixed in #3059

PR #3059, opened 2021-02-05
Make energy and momentum field names more accurate and add field deprecation mechanism

jzuhone:
## PR Summary

The various energy fields in yt have an inconsistent naming scheme. For example, the `"kinetic_energy"` derived field is an energy density (say ergs/cm^3), but `"thermal_energy"`, which usually is aliased from output fields in frontends, is usually a specific energy (erg/g). Also, momentum fields, which they appear in certain frontends, are almost always "momentum density" fields instead. 

This PR addresses this inconsistency by renaming fields wherever they appear using the following scheme:

* Energy density fields are named like this: `"kinetic_energy"` --> `"kinetic_energy_density"`
* Specific energy fields are named like this: `"thermal_energy"` --> `"specific_thermal_energy"`
* Momentum density fields are named like this: `"momentum_x"` --> `"momentum_density_x"`

Eventually, we really ought to have some kind of domain context which handles all this--and this PR should be seen as the prep work for something like that, in addition to a new class which handles different equations of state which I would also like to do. 

~~I have left aliases using the old field names to the new ones in a couple of places, with a comment in the field definition indicating deprecation. This is how we've handled other deprecated field names--however it would be nice for a deprecation to actually show up when it's used, but then show up only once. Not sure how to handle that.~~

This PR also adds mechanisms to mark fields as deprecated. The first is a new method for the `FieldInfoContainer` class, `add_deprecated_field`:

```python
def _cylindrical_radial_absolute(field, data):
    """This field is deprecated and will be removed in a future version"""	           
    return np.abs(data[ftype, f"{basename}_cylindrical_radius"])	         

registry.add_deprecated_field(
    (ftype, f"cylindrical_radial_{basename}_absolute"),
    sampling_type="local",
    function=_cylindrical_radial_absolute,
    units=field_units,	
    validators=[ValidateParameter("normal")],
    since="4.0",
    removal="4.1",
)
```

The second is to update the `alias` method to mark a field as deprecated:

```python
registry.alias(
    (ftype, "kinetic_energy"), (ftype, "kinetic_energy_density"), deprecate=("4.0", "4.1")
)
```

In both cases, a logged warning will appear:

```pycon
yt : [WARNING  ] 2021-02-19 13:30:01,701 The Derived Field ('gas', 'cylindrical_radial_velocity_absolute')
is deprecated since yt v4.0 and will be removed in yt v4.1. 
```

```pycon
yt : [WARNING  ] 2021-02-19 13:28:18,766 The Derived Field ('gas', 'kinetic_energy')
is deprecated since yt v4.0 and will be removed in yt v4.1. Use ('gas', 'kinetic_energy_density') instead.
```

Note that in the `alias` case the field that you should use is reported. Also, this logged warning only appears once.

I've taken the liberty in this PR of changing the implementation of fields which previously only had deprecation comments in their field definitions to use one of the methods.

I've also added new derived field definitions for momentum fields, for frontends which only have velocity fields.
 
## PR Checklist


neutrinoceros:
Thanks John, I support this !
I just want to point out that the issue of how to deprecate field names has come up in the past and corresponds to #2647 if anyone wants to discuss this issue in itself.

zingale:
I'm okay with this change, with the caveat that "thermal energy" for our codes is usually not just "thermal".  But we will probably never find a general term to make all the frontends happy.  But I might want to remove the word thermal from the Castro and MAESTROeX frontends just to avoid confusion.

jzuhone:
@zingale -- yes! we should have "internal" (or whatever you want) energies as well, and we should have a new EOS class which handles all of these distinctions, so you just tell it which fields your frontend has and it defines the rest.

neutrinoceros:
@jzuhone the build failures are not on you but you might want to address the static checker errors reported by pre-commit.ci :)

jzuhone:
So I'd like to figure out a way to raise a deprecation warning the first time one uses a deprecated field, but not thereafter. And this should not happen at field detection, but the first time one actually "uses" it. I've looked at it a bit but I'll need some help thinking through this one.

neutrinoceros:
I think this might be achievable with a custom logging filter. There's already one such filter defined in our logging module, if you're looking for inspiration :)

yt/fields/derived_field.py
```diff
@@ -34,6 +34,16 @@ def NullFunc(field, data):
     raise YTFieldNotFound(field.name)
 
 
+def DeprecatedFunc(new_field):
```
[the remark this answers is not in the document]
jzuhone: @neutrinoceros I'm still settling on the exact code here, I made some changes locally I haven't pushed up yet. But in response to your points:

1. Are we sure about that? This factory function lives in the same space as `NullFunc` and `TranslationFunc`, which are right above `DeprecatedFunc`. It uses machinery that already exists elsewhere, and is only for this purpose of decorating a field function, so I think it belongs here.
2. The camel case (which I agree is not formally correct here) actually matches what is going on with `NullFunc` and `TranslationFunc`, which was why I chose it. I'm fine with switching it to snake case, but we should probably then handle the other two (separate PR).
3. I don't see why a new logging filter is necessary. I have verified that the existing logging filter `DuplicateFilter`, which is already implemented, ensures that the warning is only logged once, so I regard that as sufficient. No need to reinvent the wheel. 
4. This is a way to fix #2647, but it doesn't do that quite yet. I agree that would be a separate contribution.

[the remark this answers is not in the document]
neutrinoceros: @jzuhone 
1. No I'm not so sure actually, you're making a good case :)
2. I didn't realise it was consistent with the surrounding functions, you're right that renaming shouldn't be done until another PR (and even then I'm not sure it's worth it now)
3. I'm confused. The existing filter only prevents _consecutive_ duplicated log entries, are you sure that's enough here ?
4. cool !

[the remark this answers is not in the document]
jzuhone: You're right about 3, I read that incorrectly. Thanks!

matthewturk:
I think this is the right thing to do, but ... it is going to be painful, right?  Won't this have a lot of downstream impact?

jzuhone:
@matthewturk Not sure if it will be or not--it should not break any scripts because the old fields will still work for now, you'll just get a deprecation warning.

jzuhone:
I started working on doc edits the other day, that's essentially what's left

jzuhone:
@matthewturk should we send a note to yt-dev, hoping to catch the developers of the frontends which I fiddled with?

yt/utilities/logger.py
```diff
@@ -84,6 +84,28 @@ def filter(self, record):
...
+    def filter(self, record):
+        if record.msg.startswith("The Derived Field"):
+            field = record.args[0]
+            if field in self.logged_fields:
+                return False
+            else:
+                self.logged_fields.append(field)
+                return True
+        else:
+            return True
```
neutrinoceros, line 103: Here's a flattened version of this logic tree
```suggestion
        if not record.msg.startswith("The Derived Field"):
            return True
            
        field = record.args[0]
        if field in self.logged_fields:
            return False

        self.logged_fields.append(field)
        return True
```

yt/data_objects/tests/test_covering_grid.py
```diff
@@ -109,7 +109,9 @@ def _run_tests(cg):
         assert_equal(xarr.z, cg["z"][0, 0, :])
 
     for level in [0, 1, 2]:
-        ds = fake_random_ds(16, fields=["density", "temperature", "thermal_energy"])
+        ds = fake_random_ds(
```
neutrinoceros, line 112: headsup: this will conflict with #3088 (ping @cphyc)

yt/fields/derived_field.py
```diff
@@ -6,7 +6,7 @@
 from more_itertools import always_iterable
 
 import yt.units.dimensions as ytdims
-from yt.funcs import iter_fields
+from yt.funcs import iter_fields, mylog
```
neutrinoceros, line 9:
```suggestion
from yt.funcs import iter_fields
from yt.utilities.logging import ytLogger as mylog
```

yt/fields/derived_field.py
```diff
@@ -34,6 +34,20 @@ def NullFunc(field, data):
...
+        if data.ds.fields_detected:
+            args = [field.name]
+            msg = "The Derived Field %s is deprecated. "
+            if ret_field != field.name:
+                msg += "Use %s instead."
+                args.append(ret_field)
+            mylog.warning(msg, *args)
+        return func(field, data)
+
+    return _DeprecatedFunc
```
neutrinoceros, line 48: I think this should use `functools.wraps` to support better tracebacks in case deprecated functions raise errors.
(the same change should be applied to TranslationFunc)

Another important point IMO is that a deprecation message should contain data about _when_ the deprecation was introduced and, more importantly, when it's due for removal.
This is already implemented as part of `yt._maintenance.deprecation.issue_deprecation_warning`, and it should probably be used here to avoid code duplication.

matthewturk:
Yeah, let's send a note.

yt/fields/field_info_container.py
```diff
@@ -427,14 +427,25 @@ def alias(self, alias_name, original_name, units=None):
...
         self.add_field(
             alias_name,
-            function=TranslationFunc(original_name),
+            function=function,
             sampling_type=self[original_name].sampling_type,
             display_name=self[original_name].display_name,
             units=units,
         )
 
+    def add_deprecated_field(self, name, function, sampling_type, **kwargs):
```
neutrinoceros, line 441: So my understanding is that we should use `alias` with `deprecate=True` to deprecate an alias, and call `add_deprecated_field` for deprecated fields that are not aliases. Am I right ? (I think you said you were going to add docs, so I'm writing the question I'd like such doc to answer :) )
Why isn't `add_deprecated_field` used in `alias` I don't fully understand though

yt/fields/fluid_fields.py
```diff
@@ -56,6 +57,31 @@ def _cell_mass(field, data):
...
+    def momentum_xyz(v):
+        def _mom(field, data):
+            return data["gas", "mass"] * data["gas", f"velocity_{v}"]
+
+        def _momd(field, data):
+            return data["gas", "density"] * data["gas", f"velocity_{v}"]
+
+        return _mom, _momd
+
+    for v in "xyz":
```
neutrinoceros, line 70: careful, these only make sense in cartesian coordinate systems. I believe the more general implementation involves
```python
for ax in registry.ds.coordinates.axis_order:
    ...
```

neutrinoceros (commented):
I went over the important parts and left comments and suggestions. I'll be happy to review this more thoroughly once it's settled.

yt/data_objects/tests/test_covering_grid.py
```diff
@@ -109,7 +109,9 @@ def _run_tests(cg):
...
+            16, fields=["density", "temperature", "specific_thermal_energy"]
+        )
```
cphyc, line 114: Conflict with #3088 ahead! The fix is something like (check the units of the specific thermal energy)
```suggestion
        fields = ["density", "temperature", "specific_thermal_energy"]
        units = ["g/cm*3", "K", "erg/g"]
        ds = fake_random_ds(16, fields=fields, units=units)
```

yt/fields/derived_field.py
jzuhone: @matthewturk what do you think about using `functools.wraps` here?

neutrinoceros:
@jzuhone I note that the pre-commit.ci bot is more or less "stealing" your authorship on this PR as it consistently reformat your commits, which makes it the last "person" git blame points to. I'm not forcing anyone but I think you should consider using pre-commit locally to avoid this. You can also rewrite the branch history to squash the bot's commits into yours, though that's a more painful workflow in the long run.

munkm:
Maybe we should update this CI to run after the number of approving PRs is met, so that way it doesn't run as somebody is developing (and hopefully we don't bog down our history with a ton of bot commits).

neutrinoceros:
> Maybe we should update this CI to run after the number of approving PRs is met, so that way it doesn't run as somebody is developing (and hopefully we don't bog down our history with a ton of bot commits).

We can't (it's not configurable). That's the whole point of this tool: catch and fix everything it can as soon as possible.

munkm:
> That's the whole point of this tool: catch and fix everything it can as soon as possible. 

I understand that. I just don't think it's particularly good to ask contributors to run pre-commit locally and I would like to find a good solution around it.

neutrinoceros:
Looks like its ready for a final round of reviews @jzuhone ?

jzuhone:
No--still finishing up with your original suggestions. Unfortunately it's my on-call week with Chandra so I am squeezing things in as I can. Hoping to finish today.

yt/fields/field_info_container.py
jzuhone: Fixed both of these points.

jzuhone:
I'm with @munkm -- we should find a way to make pre-commit run only after a certain point which we can specify. I'm sure there must be some way to do it.

neutrinoceros:
> I'm with @munkm -- we should find a way to make pre-commit run only after a certain point which we can specify. I'm sure there must be some way to do it.

I seriously doubt it's possible: not only is it not documented as such, but delaying action is very much not in line with the author's intention.

jzuhone:
But what about what's best for us? Nothing is stopping us from deciding something different, or even forking that project and changing it to adapt to our needs.

neutrinoceros:
TBF I can't imagine how that would represent a smaller effort than installing pre-commit locally.

neutrinoceros:
I also want to stress that delaying formatting to the late stages of writing a PR isn't a solution to the problem I was initially discussing. It's not desirable to pollute the history with formatting-only commits, because they hide the last time that the code was touched by a human (git blame), which is extremely valuable information.
pre-commit's purpose is to solve this. The bot is also much more stable and low maintenance than the tools it replaced (slash commands in GH + manual updates for hooks). I've spent dozen of hours researching the less painful way to automate formatting for everyone but eventually I don't think we can outsmart the people who write these tools in a shorter time than it takes to slightly adjust our workflows.

(Admittedly this is a rant-like comment, but it is meant to be pedagogical rather than passive-aggressive. Please forgive me if I failed conveying that.)

yt/frontends/enzo_p/fields.py
```diff
@@ -40,6 +40,14 @@ class EnzoPFieldInfo(FieldInfoContainer):
     def __init__(self, ds, field_list, slice_info=None):
         super().__init__(ds, field_list, slice_info=slice_info)
 
+    def setup_fluid_fields(self):
+        super().setup_fluid_fields()
+        self.alias(
+            ("gas", "total_energy"),
+            ("gas", "specific_total_energy"),
```
neutrinoceros, line 47: I wonder if `total_specific_energy` would be more intuitive (to me it is, but that's very low level statistics haha)

neutrinoceros (commented):
Overall this is excellent work and I'm glad we're making this step towards a clarified and clean unit system.

I have two general (but minor-ish) concerns:
- I don't really get why we need _two_ ways to deprecate field names. I must be missing something but why isn't alisasing with deprecation info enough ?
- in many places, this will collide with #3101. I'm sorry for both of you. I think we should team up and combine these efforts. The best way, seems to me, would be to leave no room for ambiguity here... but it might be easier to deal with is in #3101, I don't know. What do you guys think ?

doc/source/analyzing/fields.rst
```diff
@@ -350,6 +350,33 @@ second sets the corresponding ``value``. Currently available format properties a
...
+Fields in yt representing energy and momentum quantities follow a specific
+naming convention (as of yt-4.x). In hydrodynamic simulations, the relevant
+quantities are often energy or momentum per unit mass or volume. To distinguish
+clearly between the different types of fields, the following naming convention
+is adhered to:
+
+* Energy per unit mass fields are named as ``specific_*_energy``
+* Energy per unit volume fields are named as ``*_energy_density``
+* Momentum fields should be named ``momentum_density_[xyz]`` for momentum per
+  unit density, or ``momentum_[xyz]`` for momentum.
```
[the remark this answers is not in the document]
jzuhone: This needs fixed, thanks for catching it.

yt/frontends/enzo_p/fields.py
neutrinoceros: makes senses, thank you

doc/source/developing/creating_derived_fields.rst
```diff
@@ -354,3 +354,79 @@ And you wanted to debug it, you could do:
 
 And now, when that derived field is actually used, you will be placed into a
 debugger.
+
+Deprecating Derived Fields
```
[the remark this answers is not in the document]
jzuhone: Sure, sounds good.

doc/source/analyzing/fields.rst
```diff
@@ -350,6 +350,33 @@ second sets the corresponding ``value``. Currently available format properties a
...
+For example, in the case of kinetic energy, the fields should be
+``kinetic_energy_density`` and ``specific_kinetic_energy``.
+
+In versions of yt previous to v4.0, these conventions were not adopted, and so
+energy fields in particular could be ambiguous with respect to units. For
+example, the ``kinetic_energy`` field was actually kinetic energy per unit
+volume, whereas the ``thermal_energy`` field, usually defined by various
+frontends, was typically thermal energy per unit mass. The above scheme
+rectifies these problems, but for the time being the previous field names are
+mapped to the current field naming scheme with a deprecation warning.
```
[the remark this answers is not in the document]
jzuhone: Probably 4.1, I'll add that

doc/source/developing/creating_derived_fields.rst
```diff
@@ -354,3 +354,79 @@ And you wanted to debug it, you could do:
...
+        units=field_units,
+        validators=[ValidateParameter("normal")],
+    )
+
+Note that the signature for
+:meth:`~yt.fields.field_info_container.FieldInfoContainer.add_deprecated_field``
+is the same as :meth:`~yt.fields.field_info_container.FieldInfoContainer.add_field``,
+with the exception of the ``since`` and ``removal`` arguments which indicate in
+what version the field was deprecated and in what version it will be removed.
+The effect is to add a warning to the logger when the field is first used:
```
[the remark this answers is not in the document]
jzuhone: Sure

neutrinoceros:
> For these latter fields, we don't need an alias because we're not changing the name, we're just removing the field in a future release. 

alright this makes a lot of sense, thanks !

yt/fields/field_info_container.py
```diff
@@ -421,7 +421,7 @@ def add_output_field(self, name, sampling_type, **kwargs):
         kwargs.setdefault("ds", self.ds)
         self[name] = DerivedField(name, sampling_type, NullFunc, **kwargs)
 
-    def alias(self, alias_name, original_name, units=None, deprecate=None):
+    def alias(self, alias_name, original_name, units=None, deprecate:Optional[Tuple[str]]=None):
```
neutrinoceros: I'm sorry I really didn't mean to suggest you actually used this !
this will break CI unless you also add
```python
from typing import Optional, Tuple
```
at the top !

jzuhone:
I'm actually not clear on how this conflicts a lot with #3101? Maybe in some places, but I just looked it over and it seems not many.

yt/data_objects/level_sets/clump_validators.py
```diff
@@ -42,7 +42,7 @@ def _gravitationally_bound(
     kinetic = (
         0.5
         * (
-            clump["gas", "cell_mass"]
+            clump["gas", "mass"]
```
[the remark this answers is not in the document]
jzuhone: `"cell_mass"` doesn't make sense for SPH datasets, and in yt-4.0 we have a new field `"mass"` which for grid-based datasets is aliased to `"cell_mass"`. This was just a one-off that I caught while doing the rest.

yt/fields/derived_field.py
```diff
@@ -34,6 +35,23 @@ def NullFunc(field, data):
     raise YTFieldNotFound(field.name)
 
 
+def DeprecatedFunc(ret_field, func, since, removal):
+    def _DeprecatedFunc(field, data):
+        if data.ds.fields_detected:
```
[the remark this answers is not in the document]
jzuhone: We don't want to emit deprecation messages about fields while the initial field detection stage is going on (which calls every field to see if they work)--because then one could get many, many messages for fields that one doesn't care about. So this makes sure we've already done field detection.

doc/source/developing/creating_derived_fields.rst
```diff
@@ -354,3 +354,79 @@ And you wanted to debug it, you could do:
...
+
+.. code-block:: python
+
+    registry.alias(
+        (ftype, "kinetic_energy"),
+        (ftype, "kinetic_energy_density"),
+        deprecate=("4.0", "4.1"),
+    )
+
+In this case, the log message reports to the user what field they should use:
```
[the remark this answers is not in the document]
jzuhone: Sorry, I thought you said that we needed to use the logger to avoid doing things twice?

yt/data_objects/tests/test_profiles.py
```diff
@@ -480,7 +480,10 @@ def test_profile_sph_data():
     ds = fake_sph_orientation_ds()
     # test we create a profile without raising YTIllDefinedProfile
     yt.create_profile(
-        ds.all_data(), ["density", "temperature"], ["kinetic_energy"], weight_field=None
+        ds.all_data(),
+        ["density", "temperature"],
```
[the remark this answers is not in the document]
jzuhone: It is `"gas"`--I'm actually not sure why I had to change this code and I'll have to check.

jzuhone:
It looks like the docs failure is unrelated?

jzuhone:
@yt-fido test this please

jzuhone:
Tests pass, this is ready to go.

yt/fields/fluid_fields.py
```diff
@@ -56,6 +57,31 @@ def _cell_mass(field, data):
     )
     registry.alias((ftype, "mass"), (ftype, "cell_mass"))
 
+    # momentum
+    def momentum_xyz(v):
+        def _mom(field, data):
```
munkm, line 62: Would this be better as `_momm` to match `_momd`?

munkm (commented):
A few opening review comments. I will review more soon.

brittonsmith:
@jzuhone, sorry for taking so long to look at this! I will review this weekend.

brittonsmith (approved):
@jzuhone I think this is great. I caught only a few referencing issues in the docs, but otherwise I'm happy. Nice work here.

doc/source/developing/deprecating_features.rst
```diff
@@ -30,3 +30,79 @@ If a whole function or class is marked as deprecated, it should be removed from
...
+
+The second way to deprecate a derived field is to take an existing field
+definition and change its name. In order to mark the original name as deprecated,
+use the :meth:`~yt.fields.field_info_container.FieldInfoContainer.alias`` method
+and pass the ``since`` and ``removal`` arguments (see above) as a tuple in the
+``deprecate`` keyword argument:
+
+.. code-block:: python
+
+    registry.alias(
```
munkm, line 92: Since the ordering of these tuples matters, it should be explicitly stated here I think.

yt/fields/field_info_container.py
```diff
@@ -421,7 +422,13 @@ def add_output_field(self, name, sampling_type, **kwargs):
         kwargs.setdefault("ds", self.ds)
         self[name] = DerivedField(name, sampling_type, NullFunc, **kwargs)
 
-    def alias(self, alias_name, original_name, units=None):
+    def alias(
+        self,
+        alias_name,
+        original_name,
+        units=None,
+        deprecate: Optional[Tuple[str]] = None,
```
munkm, line 430: I think we've put type annotations on hold for now.

yt/fields/geometric_fields.py
```diff
@@ -289,29 +277,12 @@ def _cylindrical_theta(field, data):
...
-
-    registry.add_field(
-        ("index", "height"),
-        sampling_type="cell",
-        function=_height,
-        validators=[ValidateParameter("center"), ValidateParameter("normal")],
-        units=unit_system["length"],
-        display_field=False,
+    registry.alias(
+        ("index", "height"), ("index", "cylindrical_z"), deprecate=("4.0.0", "4.1.0")
```
munkm, line 287:
```suggestion
    registry.alias(
        ("index", "height"), 
        ("index", "cylindrical_z"), 
        deprecate=("4.0.0", "4.1.0")
```

munkm: just for consistency with your other defs.

yt/fields/particle_fields.py
```diff
@@ -586,26 +568,16 @@ def _particle_velocity_spherical_radius(field, data):
...
-    # particel_velocity_spherical_radius is simply aliased to
-    # "particle_radial_velocity" for convenience
-    registry.add_field(
+    registry.alias(
         (ptype, "particle_radial_velocity"),
-        sampling_type="particle",
-        function=_particle_spherical_velocity_radius,
-        units=unit_system["velocity"],
-        validators=[ValidateParameter("normal"), ValidateParameter("center")],
+        (ptype, "particle_velocity_spherical_radius"),
```
munkm, line 579: Do we want to deprecate here? It seems like that wasn't the intent originally?

yt/frontends/enzo/fields.py
```diff
@@ -213,7 +213,7 @@ def setup_energy_field(self):
             self.add_output_field(
                 ("enzo", te_name), sampling_type="cell", units="code_velocity**2"
             )
-            self.alias(("gas", "thermal_energy"), ("enzo", te_name))
+            self.alias(("gas", "specific_thermal_energy"), ("enzo", te_name))
```
munkm, line 216:
```suggestion
            self.alias(
                ("gas", "specific_thermal_energy"), 
                ("enzo", te_name)
             )
```

munkm (approved):
Ok this is the rest of my review! I'm really happy with this feature and I think it is a nice addition. My approval assumes comments will be resolved. 🙂

yt/fields/field_info_container.py
neutrinoceros: That's not what I understood from the discussion in YTEP-0038 :/

munkm: It is 0038 still hasn't been resolved or merged and a YTEP shouldn't be a blocker for this PR.

neutrinoceros: I'm not saying it's a blocker, I'm specifically referring to this comment https://github.com/yt-project/ytep/pull/17#issuecomment-788188675 from @matthewturk (in response to my own)

in short: refraining ourselves from adding type annotations now in parts of the code we're already spending energy figuring out is a waste of time.

munkm: There is a much better way to phrase this than saying "waste of time"

neutrinoceros: I'm sorry about that, I was in a rush, which is no excuse for being rude.

yt/fields/particle_fields.py
jzuhone: You're right, will fix

yt/fields/fluid_fields.py
jzuhone: Yes!

doc/source/analyzing/fields.rst
```diff
@@ -350,6 +350,35 @@ second sets the corresponding ``value``. Currently available format properties a
...
             options are ``"plus_minus"`` and ``"roman_numeral"``
 
+.. _efields:
+
+Energy and Momemtum Fields
+--------------------------
+
+Fields in yt representing energy and momentum quantities follow a specific
+naming convention (as of yt-4.x). In hydrodynamic simulations, the relevant
+quantities are often energy or momentum per unit mass or volume. To distinguish
```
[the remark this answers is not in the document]
jzuhone: I can't think of anything right now (besides maybe enthalpy, which I can add). Eventually we will have just plain energy fields, once the old deprecations are removed.

doc/source/analyzing/fields.rst
```diff
@@ -350,6 +350,35 @@ second sets the corresponding ``value``. Currently available format properties a
...
+``kinetic_energy_density`` and ``specific_kinetic_energy``.
+
+In versions of yt previous to v4.0.0, these conventions were not adopted, and so
+energy fields in particular could be ambiguous with respect to units. For
+example, the ``kinetic_energy`` field was actually kinetic energy per unit
+volume, whereas the ``thermal_energy`` field, usually defined by various
+frontends, was typically thermal energy per unit mass. The above scheme
+rectifies these problems, but for the time being the previous field names are
+mapped to the current field naming scheme with a deprecation warning. These
+aliases will be removed in yt v4.1.0.
```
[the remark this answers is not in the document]
jzuhone: Yeah, that's a good idea. I think it should still stick at 4.1 though.

yt/fields/field_info_container.py
```diff
@@ -433,12 +440,37 @@ def alias(self, alias_name, original_name, units=None):
...
+                alias_name,
+                function=function,
+                sampling_type=self[original_name].sampling_type,
+                display_name=self[original_name].display_name,
+                units=units,
+            )
+
+    def add_deprecated_field(
+        self, name, function, sampling_type, since, removal, ret_name=None, **kwargs
+    ):
```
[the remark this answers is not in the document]
jzuhone: Yup

jzuhone:
All right--I have addressed all of the comments from @munkm and @brittonsmith, and most of them from @neutrinoceros. I also rebased interactively and squashed all of the pre-commit commits with mine. Let me know if there's anything else.

yt/fields/derived_field.py
jzuhone: @neutrinoceros let's make an issue for using `functools.wraps` here and do a separate PR.

neutrinoceros: Is this snippet still relevant ? I remember seeing a different version where my second comment was taken into account already.

doc/source/developing/creating_derived_fields.rst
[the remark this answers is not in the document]
neutrinoceros: This might be a leftover from a first draft of my review. I went back and corrected some stuff when I understood why you couldn't use issue_deprecation_warning, but I likely forgot about this one. You can resolve !

yt/data_objects/tests/test_profiles.py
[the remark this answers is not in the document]
neutrinoceros: I'm seeing this as "outdated". Is there anything left to discuss here ? If not, you may resolve ! (sorry for taking 4 days to respond)

neutrinoceros (approved):
I went back in search for the very few of my points that are not resolved yet. I believe that they are mostly reminiscing because it's getting hard to keep track of all the threads, and I left answers to all 3 I could find. Please let me know if I missed anything. Meanwhile, I don't see much point in delaying my approval any longer !
Thank you for the care you put into this John !

yt/fields/derived_field.py
jzuhone: The second comment, yes--but not the `functools.wraps` bit. I wanted to think a bit more about how to do the latter part, which may require a second PR. There may be warrant for a more general refactor here.

neutrinoceros: fine by me :)

Issue #3029, opened 2021-01-21
Doc: missing pages on the website


neutrinoceros:
Currently this page (https://yt-project.org/doc/yt4differences.html) does not exist.

_Originally posted by @cphyc in https://github.com/yt-project/yt/pull/3007#discussion_r559586405_

neutrinoceros:
Here's the complete list of doc files that are not included anywhere:

```
doc/source/analyzing/domain_analysis/index.rst
doc/source/analyzing/domain_analysis/xray_data_README.rst
doc/source/developing/deprecating_features.rst
doc/source/reference/api/modules.rst
doc/source/reference/api/yt.convenience.rst
doc/source/reference/api/yt.pmods.rst
doc/source/reference/api/yt.utilities.configuration_tree.rst
doc/source/yt4differences.rst
```

Specifically, the warning those all trigger is
```
 WARNING: document isn't included in any toctree
```

neutrinoceros:
Here's the full list of existing files with no references
```
doc/source/analyzing/domain_analysis/index.rst
doc/source/analyzing/domain_analysis/xray_data_README.rst
doc/source/developing/deprecating_features.rst
doc/source/yt4differences.rst
```

the common warning for these is `WARNING: document isn't included in any toctree`

neutrinoceros:
I think this was solved a while ago

PR #3061, opened 2021-02-06
exp: remove unused module pmods

neutrinoceros:
## PR Summary

I believe this is an unused external module who survived the removal of yt/externs


matthewturk:
Hmm, let's discuss this one. I think it may still be used by user-written
scripts.

neutrinoceros:
sure !

neutrinoceros:
note: this module depends on `imp`, which is a deprecated module from the standard lib (https://docs.python.org/3/library/imp.html), so if we want to keep it, it will be necessary to update it at least for this reason.

neutrinoceros:
Coincidently, it appears that the corresponding documentation is not included in the docs build, see https://github.com/yt-project/yt/issues/3029
@matthewturk , how likely is it that anyone is still using it ?

matthewturk:
I am growing skeptical that it is used.  And, more convincingly (to myself), I'm starting to think including it is the wrong approach.  I think we should remove it for 4.0.  Can you notify yt-dev?  I will chime in with my supporting reasons.

neutrinoceros:
>  Can you notify yt-dev?

done !

chummels:
I thought I used to use this in my parallel computation scripts, but I guess maybe I'm mistaken and I just used `piter()`?  Come to think of it, I don't remember exactly what `pmods` did and looking at the source isn't seeming to help. :)

matthewturk:
You probably did use it -- it *used* to manage the import process to try to
alleviate the strain of many-nodes-importing-at-once.

Issue #2864, opened 2020-08-14
Type annotation


matthewturk:
This is meant to be a starting point for some discussion, although there have been previous discussions, too!  #2352 for instance.

Now that we are Python 3-only, we should explore adding type hints to the codebase.  I think the most straightforward way of doing this piecemeal would be to use the [`pytest-annotate`](https://blog.kensho.com/pytest-annotate-is-now-open-source-5dd6f6d51d0f) plugin for pytest, and then run it through [pyannotate](https://github.com/dropbox/pyannotate).  I've done an example of this [here](https://gist.github.com/matthewturk/7cd0cd153e987b038aef40cc9347cb46/).

There are a numbero f things that are maybe sub-optimal about it, though -- specifically, it results in a lot of absolute imports, and sometimes the import structure resolves differently than what we use in yt so it adds double-imports, etc.  (i.e., we might import `RandomState` from `np.random` but they might import from `np.random.mtrand`.)

What we might consider doing is hitting the most relevant areas of the code, specifically the user-facing APIs in the data containers, the developer-facing APIs in the frontend code, and then see how far that gets us.  It would also be a good opportunity to connect this with using [traitlets](https://github.com/yt-project/ytep/pull/8) since we'd have a blueprint for which types are what.

neutrinoceros:
> What we might consider doing is hitting the most relevant areas of the code, specifically the user-facing APIs in the data containers, the developer-facing APIs in the frontend code

excellent plane, I support this !

neutrinoceros:
Quick note that those two flake8-plugins may be useful in getting there
https://pypi.org/project/flake8-annotations/
https://github.com/asottile/flake8-typing-imports

neutrinoceros:
and another one https://github.com/best-doctor/flake8-annotations-complexity

neutrinoceros:
note that numpy exposes a `testing` module https://numpy.org/doc/stable/reference/typing.html#module-numpy.typing
It is however documented as requiring Python 3.8 so I guess we can start using it next year.

neutrinoceros:
I think it's safe to close this now that #3546 went in

PR #3062, opened 2021-02-06
Consistent return statements

neutrinoceros:
## PR Summary

Our style guide contains the following rule

> Only declare local variables if they will be used later. If you do not use the		
   return value of a function, do not store it in a variable.

Unfortunately this rule is not automatically enforced _yet_, and the code base currently contains 363 violations of it.
The good news is that [flake8-return](https://github.com/afonasev/flake8-return) defines this as rule `R504`. This plugin also defines a couple of other rules for which we have a cumulated number of violations of 55.

Here I fix all of those, and I add a handful of noqas specifically in cases that can't really be refactored straightforwardly.
(mostly places where `tmpdir` is used without context managers, I might fix those later)
 
Happy side effects:
- in most cases, fixing these errors produces more concise and clearer code
- R503 helps discovering branches of code where errors should be raised
- coincidently, fixing R502 and R503 violations will be helpful later to add type hints to the code base (#2864)

## todo

- [x] finish up cleaning existing errors
- [x] review myself (in particular see the resulting consistency of _is_valid methods)
- [x] check that nothing breaks
- [x] rebase and squash commits

### DIFF
--- a/yt/config.py
+++ b/yt/config.py
@@ -108,15 +108,12 @@ class YTConfig:
         return node_or_leaf
 
     def get_most_specific(self, section, *keys, **kwargs):
-        use_fallback = "fallback" in kwargs
-        fallback = kwargs.pop("fallback", None)
         try:
             return self.config_root.get_deepest_leaf(section, *keys)
         except KeyError as err:
-            if use_fallback:
-                return fallback
-            else:
-                raise err
+            if "fallback" in kwargs:
+                return kwargs["fallback"]
+            raise err
 
     def update(self, new_values, metadata=None):
         if metadata is None:
--- a/yt/data_objects/profiles.py
+++ b/yt/data_objects/profiles.py
@@ -40,24 +40,6 @@ def _sanitize_min_max_units(amin, amax, finfo, registry):
     return rmin, rmax
 
 
-def preserve_source_parameters(func):
-    def save_state(*args, **kwargs):
-        # Temporarily replace the 'field_parameters' for a
-        # grid with the 'field_parameters' for the data source
-        prof = args[0]
-        source = args[1]
-        if hasattr(source, "field_parameters"):
-            old_params = source.field_parameters
-            source.field_parameters = prof._data_source.field_parameters
-            tr = func(*args, **kwargs)
-            source.field_parameters = old_params
-        else:
-            tr = func(*args, **kwargs)
-        return tr
-
-    return save_state
-
-
 class ProfileFieldAccumulator:
     def __init__(self, n_fields, size):
         shape = size + (n_fields,)
--- a/yt/fields/field_functions.py
+++ b/yt/fields/field_functions.py
@@ -33,11 +33,9 @@ def get_radius(data, field_prefix, ftype):
         np.add(radius2, r, radius2)
         if data.ds.dimensionality < i + 1:
             break
-    # Now it's cm.
-    np.sqrt(radius2, radius2)
-    # Alias it, just for clarity.
-    radius = radius2
-    return radius
+
+    np.sqrt(radius2, r)
+    return r
 
 
 def get_periodic_rvec(data):
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -357,7 +357,7 @@ class FieldInfoContainer(dict):
                     return f
 
                 return create_function
-            return
+            return None
         # add_field can be used in two different ways: it can be called
         # directly, or used as a decorator (as yt.derived_field). If called directly,
         # the function will be passed in as an argument, and we simply create
@@ -375,7 +375,7 @@ class FieldInfoContainer(dict):
 
         if isinstance(name, tuple):
             self[name] = DerivedField(name, sampling_type, function, **kwargs)
-            return
+            return None
 
         sampling_type = self._sanitize_sampling_type(
             sampling_type, particle_type=kwargs.get("particle_type")
@@ -394,6 +394,7 @@ class FieldInfoContainer(dict):
             self.alias(name, tuple_name)
         else:
             self[name] = DerivedField(name, sampling_type, function, **kwargs)
+        return None
 
     def load_all_plugins(self, ftype="gas"):
         loaded = []
@@ -409,8 +410,7 @@ class FieldInfoContainer(dict):
             f = field_plugins[plugin_name]
         orig = set(self.items())
         f(self, ftype, slice_info=self.slice_info)
-        loaded = [n for n, v in set(self.items()).difference(orig)]
-        return loaded
+        return [n for n, v in set(self.items()).difference(orig)]
 
     def find_dependencies(self, loaded):
         deps, unavailable = self.check_derived_fields(loaded)
--- a/yt/frontends/amrvac/data_structures.py
+++ b/yt/frontends/amrvac/data_structures.py
@@ -229,27 +229,31 @@ class AMRVACDataset(Dataset):
     @classmethod
     def _is_valid(cls, filename, *args, **kwargs):
         """At load time, check whether data is recognized as AMRVAC formatted."""
-        validation = False
-        if filename.endswith(".dat"):
-            try:
-                with open(filename, mode="rb") as istream:
-                    fmt = "=i"
-                    [datfile_version] = struct.unpack(
+        if not filename.endswith(".dat"):
+            return False
+        try:
+            with open(filename, mode="rb") as istream:
+                fmt = "=i"
+                [datfile_version] = struct.unpack(
+                    fmt, istream.read(struct.calcsize(fmt))
+                )
+                if 3 <= datfile_version < 6:
+                    fmt = "=ii"
+                    offset_tree, offset_blocks = struct.unpack(
                         fmt, istream.read(struct.calcsize(fmt))
                     )
-                    if 3 <= datfile_version < 6:
-                        fmt = "=ii"
-                        offset_tree, offset_blocks = struct.unpack(
-                            fmt, istream.read(struct.calcsize(fmt))
-                        )
-                        istream.seek(0, 2)
-                        file_size = istream.tell()
-                        validation = (
-                            offset_tree < file_size and offset_blocks < file_size
-                        )
-            except Exception:
-                pass
-        return validation
+                    istream.seek(0, 2)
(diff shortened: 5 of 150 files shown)


yt/data_objects/profiles.py
```diff
@@ -40,24 +40,6 @@ def _sanitize_min_max_units(amin, amax, finfo, registry):
     return rmin, rmax
 
 
-def preserve_source_parameters(func):
```
neutrinoceros, line 43: this decorator seems dead

neutrinoceros:
A note to reviewers: in some cases, I needed to refactor around file opening/closing with h5py. For instance:
```python
# this violates R504
fh = h5py.File(datafile, mode="rb")
result = fh.some_operation(...)
fh.close()
return result
```
into
```python
with h5py.File(datafile, mode="rb") as fh:
    return fh.some_operation(...)
```
I researched the topic to make extra sure this is okay: from the release notes of h5py 2.0, using `h5py.File` as a context manager was already supported, and this is also the first version to support Python 3 so I think there's no possible backlash here !
see https://docs.h5py.org/en/stable/whatsnew/2.0.html?highlight=context%20manager

yt/config.py
```diff
@@ -92,15 +92,12 @@ def callback(leaf):
...
         try:
             return self.config_root.get_deepest_leaf(section, *keys)
         except KeyError as err:
-            if use_fallback:
-                return fallback
-            else:
-                raise err
+            if "fallback" in kwargs:
+                return kwargs["fallback"]
+            raise err
```
neutrinoceros, line 100: this is not _exactly_ replicating the existing behaviour because the `kwargs` dict isn't modified in place as with `dict.pop`. I'm confident that it's not important, but just for safety I'd like @cphyc 's review specifically here :)

yt/frontends/enzo/simulation_handling.py
```diff
@@ -477,7 +477,7 @@ def _calculate_cycle_outputs(self):
             or "CycleSkipDataDump" not in self.parameters
             or self.parameters["CycleSkipDataDump"] <= 0.0
         ):
-            return []
+            return
```
neutrinoceros, line 480: this case is a bit weird. The function has only one explicitly return statement so it will usually return `None` except here where it return an empty list. I figure it is probably useless now but the function might need to be fixed differently

yt/frontends/chombo/data_structures.py
```diff
@@ -732,13 +728,13 @@ def _is_valid(cls, filename, *args, **kwargs):
...
-                fileh = h5py.File(filename, mode="r")
-                valid = "CeilVA_mass" in fileh.attrs.keys()
-                valid = (
-                    "Chombo_global" in fileh["/"] and "Charm_global" not in fileh["/"]
-                )
-                valid = valid and "CeilVA_mass" in fileh.attrs.keys()
-                fileh.close()
+                with h5py.File(filename, mode="r") as fileh:
+                    valid = "CeilVA_mass" in fileh.attrs.keys()
+                    valid &= (
```
neutrinoceros, line 733: here I'm actually trying to _fix_ a bug because I noticed that the first condition was actually irrelevant to the return value. Maybe the bug is that the line
```python
valid = "CeilVA_mass" in fileh.attrs.keys()
```
is _actually_ irrelevant and should be deleted
@atmyers might know better

yt/frontends/amrvac/data_structures.py
```diff
@@ -227,27 +227,30 @@ def __init__(
...
-                pass
-        return validation
+                    istream.seek(0, 2)
+                    file_size = istream.tell()
+                    return offset_tree < file_size and offset_blocks < file_size
+                elif datfile_version < 15:
+                    mylog.warning(
+                        "This may be a amrvac output, "
+                        f"but this format version ({datfile_version}) is not supported yet."
+                    )
```
neutrinoceros, line 250: This is brand new but as the primary author of the function I'm merely fixing a case I forgot to factor in.

yt/frontends/stream/io.py
```diff
@@ -110,11 +110,10 @@ def _read_particle_coords(self, chunks, ptf):
                 )
 
     def _read_smoothing_length(self, chunks, ptf, ptype):
-        for data_file in sorted(
+        data_file = sorted(
```
neutrinoceros, line 113: I removed the loop here because it looks superfluous

neutrinoceros:
pre-commit check has passed, so I'll open for review

neutrinoceros:
There is clearly a lot of failures, most likely due to the `raise` statements I inserted. I'm switching this back to draft mode for now.

yt/config.py
cphyc: It's much better not to alter `kwargs` at all (as you're doing here) rather than pop'ing arguments (as I was doing).

neutrinoceros: Cool, thanks !

yt/frontends/gadget/data_structures.py
```diff
@@ -666,24 +666,12 @@ def _set_owls_eagle_units(self):
...
+        # need_groups = ["Header"]
+        # veto_groups = ["FOF", "Group", "Subhalo"]
         try:
-            fh = h5py.File(filename, mode="r")
-            valid = all(ng in fh["/"] for ng in need_groups) and not any(
-                vg in fh["/"] for vg in veto_groups
-            )
-            fh.close()
+            with h5py.File(filename, mode="r") as fh:
+                # valid = all(ng in fh["/"] for ng in need_groups)
```
neutrinoceros, line 673: On the main branch, the first try block has no lasting effect on the value of `valid`. In my refactor, I tried making it relevant and it broke a lot of tests, so now I'm commenting out the broken bits and I see no path for them but removal unless someone with more knowledge of the frontend could jump in and tell me how it should be fixed. From the git blame, I _think_ @brittonsmith might be able to play that role but it's not really clear as this function has a lot of contributors.

neutrinoceros:
making progress, I'm left with one error (5 failures, but all with the same issue):
```
raise YTAmbiguousDataType(fn, candidates)
yt.utilities.exceptions.YTAmbiguousDataType: Multiple data type candidates for /mnt/yt/KeplerianRing/keplerian_ring_0020.hdf5
<class 'yt.frontends.gadget.data_structures.GadgetHDF5Dataset'>
<class 'yt.frontends.gadget.data_structures.GadgetDataset'>
<class 'yt.frontends.swift.data_structures.SwiftDataset'>
```
So I must have made at least one of those classes' `_is_valid` a little less accurate. `GadgetDataset` is the last one I touched.
I also note that the error message is slightly misleading (my fault) because it states that `GadgetHDF5Dataset` and `GadgetDataset` are "independent" which is simply not true.

yt/frontends/enzo/simulation_handling.py
matthewturk: Probably need @brittonsmith 's eyes on this

yt/frontends/stream/io.py
matthewturk: Hmmmmmm.  I believe you that it's likely true, but looking at it I don't see how.

yt/frontends/gadget/data_structures.py
matthewturk: If I am remembering correclty, this is trying to discern between a couple different very-mildly-different formats, so it's probably a bit tricky.

neutrinoceros: Well the road was indubitably bumpy, but I think I got the PR to a stable state with not YTAmbiguousDataFormat or YTUnidentifiedDataFormat so I'm a lot more confident that this is correct

munkm: > indubitably 

😂 thank you for using this amazing word in a PR discussion

neutrinoceros: Tbh I think it is slightly more common (and less ridiculous) in French, but I stand by it :)

neutrinoceros:
This is now rebased on top of #3154, in hopes it will fix one of the issues.

yt/fields/field_functions.py
```diff
@@ -33,11 +33,9 @@ def get_radius(data, field_prefix, ftype):
         np.add(radius2, r, radius2)
         if data.ds.dimensionality < i + 1:
             break
-    # Now it's cm.
-    np.sqrt(radius2, radius2)
-    # Alias it, just for clarity.
-    radius = radius2
-    return radius
+
+    np.sqrt(radius2, r)
```
cphyc, line 37:
```suggestion
    np.sqrt(radius2, out=r)
```

yt/fields/field_info_container.py
```diff
@@ -392,6 +392,7 @@ def create_function(f):
             self.alias(name, tuple_name)
         else:
             self[name] = DerivedField(name, sampling_type, function, **kwargs)
+        return None
```
cphyc, line 395: I see `return None` statement all over the place ; isn't it equivalent to just return None? Why is it better (if it is)?

cphyc (approved):
Looks good to me!

yt/fields/field_info_container.py
neutrinoceros: it *is* equivalent to `return`. In the context of functions that may return anything else, it is preferable to make it explicit that some branch in fact return None. It is "better" in the sense that
> explicit is better than implicit
(Zen of Python)

I'm convinced it will come in handy for YTEP 0038 too :)

neutrinoceros:
quoting @matthewturk from the soon-to-be-drop remake PR:
> I don't think we should prioritize this just now. I am personally not that into the changes to the return statements anyway, but I think more broadly this should be discussed.
One reason I am particularly not interested in this is that having it assigned to a local variable does no harm and additionally will make it easier to debug.

To clarify, this work is much more about paving the road for typing annotations than it is about saving a few variable assignments.
Let's take a look at
https://github.com/afonasev/flake8-return#errors

I'm mostly interested in fixing R501, R502 and R503, while you seem more concerned about R504 (though admittedly this is indeed the point that got me started in the first place). Seems to me there is room for discussion !

neutrinoceros:
Taking a step back, it saddens me a bit but I'd rather throw this away than keep it stale or rotting forever, and it's pretty clear it didn't raise half as much enthusiasm as would be needed to actually review this. And off this goes ➡️ 🗑️

PR #3064, opened 2021-02-07
TYP: add a yt._typing module and start annotating yt.loaders

neutrinoceros:
## PR Summary

This is small step in the direction of #2864
- add a yt._typing module (a place to define custom/complex type definitions (imitating pandas)
- start annotating fonctions in yt.loaders (experimental, it's dauntingly simple to create circular imports by accident here)

also add a flake8 plugin (flake8-typing-imports) to enforce safe guarding of typing imports from day one.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,6 +53,7 @@ repos:
         flake8-bugbear>=20.3.2, # GH PR 2851
         flake8-logging-format,
         flake8-2020==1.6.0,
+        flake8-typing-imports==1.10.1,
       ]
 -   repo: https://github.com/asottile/blacken-docs
     rev: v1.11.0
--- a/yt/_typing.py
+++ b/yt/_typing.py
@@ -1,4 +1,42 @@
 import os
-from typing import Union
+from typing import Tuple, TypeVar, Union
 
-PathType = Union[str, bytes, "os.PathLike[str]", int]
+from matplotlib.colors import ListedColormap
+from unyt.unit_object import Unit
+
+# this defines a useful generic type to maintain type information
+# across generic functions and parametrization
+# It is close to the concept of templates in C++
+# see https://docs.python.org/3/library/typing.html#generics
+T = TypeVar("T")
+
+FileLike = Union[str, bytes, "os.PathLike[str]", int]
+
+# Note that Tuple is currently the only Sequence-like type
+# that support specifying the size
+# In most places where a field-like input is expected,
+# List[str] should be acceptable as long as the size is exactly 2.
+#
+# here's a proto adapter function in python>=3.10
+# def unpack_field(field:FieldLike) -> Tuple[str, str]:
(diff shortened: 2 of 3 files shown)


munkm:
Thanks for starting on this! 

Since this is a change that actually involves a potential change to our API and the way users might interact with yt, maybe it is worth opening a YTEP about this before we dive too deep in development? I think this is the type of thing that community decision making (like what we want the API to eventually behave like, and how we want users to interact with this) would be really good for! What do you think?

neutrinoceros:
If we're aiming at (and I think we are) adding type hints everywhere, we'll need automation at some point. I know @cphyc and I are independently experimenting with this but atm it looks like a medium to long term task. If and when we manage to figure it out it will likely warrant a YTEP indeed, but right now I just feel like it may not be worth the time investment. Here my goal is to set the conditions to enable organic growth of our typing coverage but defining a place for common definitions before they start poping everywhere and it becomes harder to collect them.
Can I request @cphyc and/or @matthewturk thoughts ?

neutrinoceros:
Incidentally this is in line with the guidelines here https://mypy.readthedocs.io/en/stable/existing_code.html#using-mypy-with-an-existing-codebase

matthewturk:
So I tend to agree with @munkm that before we go full-scale, we should have a YTEP.  I am not yet sure in my head the extent that it needs to cover, but it's definitely of the scale we should discuss.  There's the traitlets YTEP that could potentially be modified to include type hinting, but I think type hinting might be considerably less invasive and so could be its own thing.

cphyc:
I also agree that we ought to discuss this before going full scale. As per my comment below, I'd however argue that if we want type annotations to happen, we should start as soon as possible if we follow a gradual approach. Now the question is how to achieve this without asking too much from the ones pushing for type annotations (in particular @neutrinoceros) and while not deterring other contributors from contributing.

I had a few attempts to quickly generate type annotations for yt, but I now doubt this is the right way to go. I found the type inferred not to be very accurate (they were _too_ specific) and required a lot of manual interventions to correct them. I guess we should instead do it gradually (either by asking contributors to add type annotations, or by following a one-module-one-pull-request). I also think we should start with the user facing part, as this is the most "visible" one.

That may not be the clearest comment I've ever written but I hope this is clear nonetheless!

matthewturk (commented):
I really like this.  I'll get a YTEP sketch going today.

matthewturk:
So I remain excited about this, but cannot recall how I can help move it along -- is it just the YTEP?

neutrinoceros:
YTEP 0038 has far to large of a scope to be realistically implemented in one quick shot, so I think we'd need to get it approve and then move along with this PR as a first step ?
Note that since we're going to drop Python 3.6 in a month, we could save ourself some churn by doing this one soon after.
In the mean time, I'm seeing conflicts, let's solve these now to get the PR in a healthy state.

PR #3065, opened 2021-02-08
make Scene.save less verbose

neutrinoceros:
## PR Summary
Currently the answer test build's log is polluted with these warnings:
```
VRImageComparison_UniformGridData_vr_yaw_plane-parallel_0001 ... WARNING:yt:Previously rendered image exists, but rendering anyway. Supply 'render=False' to save previously rendered image directly.
INFO:yt:Rendering scene (Can take a while).
INFO:yt:Saving rendered image to /tmp/tmp7ivhf93i.png
ok
VRImageComparison_UniformGridData_vr_pitch_plane-parallel_0002 ... WARNING:yt:Previously rendered image exists, but rendering anyway. Supply 'render=False' to save previously rendered image directly.
INFO:yt:Rendering scene (Can take a while).
INFO:yt:Saving rendered image to /tmp/tmp4tvkj13k.png
ok
VRImageComparison_UniformGridData_vr_roll_plane-parallel_0003 ... WARNING:yt:Previously rendered image exists, but rendering anyway. Supply 'render=False' to save previously rendered image directly.
INFO:yt:Rendering scene (Can take a while).
INFO:yt:Saving rendered image to /tmp/tmpyd_sw8nh.png
ok
VRImageComparison_UniformGridData_vr_perspective_0000 ... WARNING:yt:Previously rendered image exists, but rendering anyway. Supply 'render=False' to save previously rendered image directly.
```
This aims to solve this by properly controlling the `render` flag in `Scene.save...` calls

### DIFF
--- a/yt/visualization/volume_rendering/scene.py
+++ b/yt/visualization/volume_rendering/scene.py
@@ -1,6 +1,7 @@
 import builtins
 import functools
 from collections import OrderedDict
+from typing import Optional
 
 import numpy as np
 
@@ -274,7 +275,12 @@ class Scene:
         mylog.info("Saving rendered image to %s", fname)
         return fname
 
-    def save(self, fname=None, sigma_clip=None, render=True):
+    def save(
+        self,
+        fname: Optional[str] = None,
+        sigma_clip: Optional[float] = None,
+        render: bool = True,
+    ):
         r"""Saves a rendered image of the Scene to disk.
 
         Once you have created a scene, this saves an image array to disk with
@@ -358,12 +364,12 @@ class Scene:
 
     def save_annotated(
         self,
-        fname=None,
-        label_fmt=None,
-        text_annotate=None,
-        dpi=100,
-        sigma_clip=None,
-        render=True,
+        fname: Optional[str] = None,
+        label_fmt: Optional[str] = None,
+        text_annotate: Optional[str] = None,
+        dpi: int = 100,
+        sigma_clip: Optional[float] = None,
+        render: bool = True,
     ):
         r"""Saves the most recently rendered image of the Scene to disk,
         including an image of the transfer function and and user-defined
--- a/yt/visualization/volume_rendering/tests/test_scene.py
+++ b/yt/visualization/volume_rendering/tests/test_scene.py
@@ -75,7 +75,8 @@ class RotationTest(TestCase):
 
         fname = "test_rot.png"
         sc.camera.pitch(np.pi)
-        sc.save(fname, sigma_clip=6.0)
+        sc.render()
+        sc.save(fname, sigma_clip=6.0, render=False)
         assert_fname(fname)
 
 
@@ -97,7 +98,8 @@ def test_annotations():
         assert np.where((im == c).all(axis=-1))[0].shape[0] > 0
     sc[0].tfh.tf.add_layers(10, colormap="cubehelix")
     sc.save_annotated(
-        "test_scene_annotated.png", text_annotate=[[(0.1, 1.05), "test_string"]]
+        "test_scene_annotated.png",
+        text_annotate=[[(0.1, 1.05), "test_string"]],
     )
     image = imread("test_scene_annotated.png")
     assert image.shape == sc.camera.resolution + (4,)


neutrinoceros:
I need another iteration: we're still having two warnings
```
VRImageComparison_UniformGridData_vr_plane-parallel_0000 ... WARNING:yt:No previously rendered image found, rendering now.
ok (4.1270s)
VRImageComparison_UniformGridData_vr_yaw_plane-parallel_0001 ... WARNING:yt:Previously rendered image exists, but rendering anyway. Supply 'render=False' to save previously rendered image directly.
```

neutrinoceros:
while I'm at it I'm adding type hints too

neutrinoceros:
@chrishavlin could you please review this ?

munkm (approved):
LGTM

yt/visualization/volume_rendering/scene.py
```diff
@@ -359,12 +365,12 @@ def save(self, fname=None, sigma_clip=None, render=True):
...
-        label_fmt=None,
-        text_annotate=None,
-        dpi=100,
-        sigma_clip=None,
-        render=True,
+        fname: Optional[str] = None,
+        label_fmt: Optional[str] = None,
+        text_annotate: Optional[str] = None,
+        dpi: int = 100,
+        sigma_clip: Optional[float] = None,
```
munkm, line 372: This looks like it's causing a failure on jenkins.

munkm, line 372: I just realized if we're going to wait on type hinting until we come up with a community plan maybe we should remove the type`Optional` type hinting here?

yt/visualization/volume_rendering/scene.py
```diff
@@ -246,7 +247,12 @@ def _sanitize_render(self, render):
 
         return render
 
-    def save(self, fname=None, sigma_clip=None, render=True):
+    def save(
+        self,
+        fname: Optional[str] = None,
+        sigma_clip: Optional[int, float] = None,
+        render: bool = True,
```
munkm, line 254: and here

munkm:
Sorry for all the comments. I reviewed this specifically looking at the rerender stuff and then after review saw the typing stuff. I think we should remove since (a) we don't have a community plan for implementing type hinting yet and (b) it isn't a direct need for the fix that this PR is solving.

yt/visualization/volume_rendering/scene.py
neutrinoceros: I'd rather keep this PR waiting, it's not fixing anything important

neutrinoceros: Actually those hints are perfectly compliant with python 3.6 so I don't think we need to keep the PR waiting (of course I still need to handle the failures)
Btw there are _already_ one or two similar type hints in the main branch.

munkm:
This should not be merged until a decision is made in yt-project/yt#3064

yt/visualization/volume_rendering/scene.py
munkm: My point is that the discussion in the other PR is pausing our type hinting implementation until we have a community discussion about it. This seems to be moving forward with it. 

> Btw there are already one or two similar type hints in the main branch.

I don't see any "from typing" or "import typing" places anywhere in main?

neutrinoceros: in fact the issue was that I wrote `Optional[int, float]`, which is a syntax error, instead of `Optional[Union[int, float]]`. It is solved now !

neutrinoceros: Right, but I'm pretty sure one or two type hints were sprinkled in #2981 (@cphyc ?)
Note that some hints are simple enough that they don't require any imports, for instance
```python
def spam(bacon: int, eggs: str) -> float:
    return 1.0
```
that being said I'm having a hard time gripping the ones I think got in already.

yt/visualization/volume_rendering/scene.py
neutrinoceros: 253 was the line causing the failure, it is fixed now

chrishavlin (changes requested):
a couple of small comments, but more generally if the goal here is to make `_sanitize_render` quieter, why not just remove the warnings? The only really helpful one IMO is the warning for when a previous rendering exists but `render=True` (though in my comments I note that the changes here no longer catch that case correctly). I'd be happy with having the other `mylog` calls removed. As long as `_sanitize_render` ensures that `render=True` if there is no previous rendering (which the changed version still does), I don't really have any strong feelings on the messages.

yt/visualization/volume_rendering/tests/test_scene.py
```diff
@@ -97,7 +98,9 @@ def test_annotations():
         assert np.where((im == c).all(axis=-1))[0].shape[0] > 0
     sc[0].tfh.tf.add_layers(10, colormap="cubehelix")
     sc.save_annotated(
-        "test_scene_annotated.png", text_annotate=[[(0.1, 1.05), "test_string"]]
+        "test_scene_annotated.png",
+        text_annotate=[[(0.1, 1.05), "test_string"]],
+        render=False,
```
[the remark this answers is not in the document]
neutrinoceros: Indeed the test shouldn't break because of this, but indeed it's not ideal. I don't remember why I set `render=False` here. I was probably just trying to get rid of the warning...

yt/visualization/volume_rendering/scene.py
```diff
@@ -228,12 +229,12 @@ def render(self, camera=None):
...
+    def _sanitize_render(self, render: bool):
         # checks for existing render before saving, in most cases we want to
         # render every time, but in some cases pulling the previous render is
         # desirable (e.g., if only changing sigma_clip or
         # saving after a call to sc.show()).
-        if self._last_render is None:
+        if self._last_render is None and not render:
             mylog.warning("No previously rendered image found, rendering now.")
             render = True
         elif render:
```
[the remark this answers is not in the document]
neutrinoceros: Indeed ! thank you so much for catching this

neutrinoceros:
Thank you for reviewing @chrishavlin !
Indeed removing the noisy warnings might be the most maintainable and cleanest option here (can't believe I'm writing this 😅), you're right that only one of them really matters (and the rest seem broken).

yt/visualization/volume_rendering/scene.py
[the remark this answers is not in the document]
neutrinoceros: is it better now ?

chrishavlin:
hmm, I'm not sure why the rest would be "broken." I'd say they were behaving as expected depending on how `sc.render()` and `sc.save()` were called. But ya, I'm happy removing the warnings  (I originally only added them at your suggestion I believe haha). Is there a reason why your latest changes keep one of them? That one was the most unnecessary of the bunch IMO and if you're removing the others, why not remove that one too?

chrishavlin (approved):
looks good now (obviously pending the typing PR)! I'm happy leaving that final warning if you think it's warranted.

neutrinoceros:
So I gave this more thoughts. Let me explain my reasoning as an answer

> The only really helpful one IMO is the warning for when a previous rendering exists but render=True 

I _would_ agree with you _if_ the default value for `Scene.save`'s `render` argument wasn't `True`.
Unfortunately changing the behaviour of the function would certainly cause downstream (silent !) breakage, so I'm treating this default value as immutable (for now).
Now because it's set to `True`, it means that the following user workflow is "correct" but extremely noisy

```python
for shift_angle in np.linspace(0, 91, 91):
    scene.camera.pitch(shift_angle)
    scene.save(f"{shift_angle}.png")
```

Currently (main branch), the only way to suppress the warnings (and nothing else) from the user standpoint is to rewrite it as

```python
for shift_angle in np.linspace(0, 91, 91):
    scene.camera.pitch(shift_angle)
    scene.render()
    scene.save(f"{shift_angle}.png", render=False)
```
but this feels clunky and we probably don't want to encourage this as the optimal workflow.

>  I'm not sure why the rest would be "broken."

The fundamental issue for us is that we have no way of comparing what was last rendered with the current state of the Scene object, so we never know how relevant those warnings are. This is what I mean by "broken".

The one exception I see is when the Scene object was _never_ rendered _and_ the user calls `Scene.save(render=False)`. This is the only case for which I'm keeping a warning, because we need to inform the user that the keyword argument is being ignored out of necessity.

All things considered, I think we have a choice to make between a backward compatibility breaking change (switch default value to `render=False`) and a somewhat confusing api. I'm going with the latter here because it's easy enough to make the api less noisy, but deep down I think the former option is preferable in the long run, but warrants a larger discussion and we have already many small backward compatibility breakage in store for the 4.0 release, I don't really want to add one at the last minute.

Does it seem reasonable ?

## Follow up ?

If we decide we want to change the default behaviour, this would encourage the following workflows as equally correct (and silent)

```python
for shift_angle in np.linspace(0, 91, 91):
    scene.camera.pitch(shift_angle)
    scene.render()
    scene.save(f"{shift_angle}.png")
```

```python
for shift_angle in np.linspace(0, 91, 91):
    scene.camera.pitch(shift_angle)
    scene.save(f"{shift_angle}.png", render=True)
```

which IMO is where the api _should_ be (so... maybe for 4.1 ?)

chrishavlin:
Ok, thanks for clarifying! I agree that for this PR keeping `render=True` and simply making it less noisy is the way to go. 

Longer term, changing the default option is fine by me but I think it's worth thinking about adding some sort of state monitoring to solve the underlying issue of not knowing if scene settings have changed since the last render. Maybe a hashed attribute that stores the state used when `render` is called that we can compare against (similar to the attribute hashing we use for datasets and selection objects in some places, e.g., data_objects.static_output.ParticleFile or geomtery._selection_routines.*)?

neutrinoceros:
Excellent suggestion. I'll open an issue to outlast this PR since it seems like we're done here.

yt/visualization/volume_rendering/scene.py
Xarthisius: Not sure if that community discussion happened (I can't find it), but there are several places in the codebase that do `import typing`. Since AFAICT that was the only issue raised with this PR I'm gonna merge it.

Xarthisius (approved):

PR #3060, opened 2021-02-06
doc: add short descriptions of main code formatters

neutrinoceros:
## PR Summary

By request of @munkm

PR #3066, opened 2021-02-08
Hotfix docs build

neutrinoceros:
## PR Summary

Looks like the docs build is broken on the main branch. See for instance this page: https://yt-project.org/dev/developing
whereas https://yt-project.org/doc/developing loads fine.

My work hypothesis is that the docs build isn’t triggered when the only paged touched in CONTRIBUTING.rst, so https://github.com/yt-project/yt/pull/3060 was merged without an actual check.

Here I'm fixing the syntax error from that recent PR + a couple of other problems I detected during a local build in order to trigger the docs build.


Xarthisius:
Could describe problem that you're seeing? Cause http://yt-project.org/docs/dev/developing/ looks fine to me.
Also `CONTRIBUTING.rst` is not a part of docs, so nothing there would affect the build. FTR `docs/` is where documentation lives and only changes there trigger docs build.

neutrinoceros:
> Could you describe problem that you're seeing? Cause http://yt-project.org/docs/dev/developing/ looks fine to me.

sorry, the page didn't load at all but now it does... maybe my slow network is to be blame ?

> Also CONTRIBUTING.rst is not a part of docs, so nothing there would affect the build

I beg to differ: it is part of the docs by inclusion, see
https://github.com/yt-project/yt/blob/main/doc/source/developing/developing.rst

munkm:
> Here I'm fixing the syntax error from that recent PR + a couple of other problems I detected during a local build in order to trigger the docs build.

What are the other problems? I see the removals you've made but I'm curious why they needed too be removed?

Xarthisius:
> I beg to differ: it is part of the docs by inclusion, see
> https://github.com/yt-project/yt/blob/main/doc/source/developing/developing.rst

I wasn't aware of that. Oh well, Jenkins isn't AI. It doesn't stand a chance if someone's trying to game the system.

neutrinoceros:
> What are the other problems? I see the removals you've made but I'm curious why they needed too be removed?

@munkm those are references to removed files, they are reported by sphinx as
```
WARNING: toctree contains reference to nonexisting document 'analyzing/analysis_modules/index'
```

> I wasn't aware of that. Oh well, Jenkins isn't AI. It doesn't stand a chance if someone's trying to game the system.

of course ! What do you think is the preferred course of action here ? I see two possibilities:
- moving the content of CONTRIBUTING.rst to yt/doc/developing/developing.rst and replacing it with a url to the dev doc + a reference to the source file
- adding an exception to Jenkins' rules just for this one file.

Xarthisius:
> What do you think is the preferred course of action here @Xarthisius ? I see two possibilities:
> 
> * moving the content of CONTRIBUTING.rst to yt/doc/developing/developing.rst and replacing it with a url to the dev doc + a reference to the source file
> * adding an exception to Jenkins' rules just for this one file.

The latter. I already fixed it.

neutrinoceros:
@yt-fido test this please

doc/source/index.rst
```diff
@@ -182,7 +182,6 @@ Table of Contents
    cookbook/index
    visualizing/index
    analyzing/index
```
munkm, line 184:
```suggestion
   analyzing/index
   analyzing/domain_analysis/index
```

munkm (approved):
Everything looks good! One suggestion for something that was renamed.

doc/source/index.rst
neutrinoceros: done, thanks !

PR #3072, opened 2021-02-10
[BUGFIX] Support old gcc versions

### LINKED ISSUE #3071: Compiling yt with old gcc versions
### Bug report

**Bug summary**

The code base passes `-std=c++14` to gcc when compiling C++ files. This flag is however only supported on GCC 5 and later. For earlier version, it seems we could escape by passing `-std=c++1y` instead, but I don't know what implication that would have (if any).

cphyc:
Some files require some C++14 features which aren't supported on old versions of GCC. This PR tries to find the most suitable compiler flags to supports these features (though it does not capture all cases).

I tested this PR with GCC 4.7.4 (can now compile yt, could not before). With GCC 4.6.4 and older versions, the compilation fails as the support for C++14 is unsufficient to compile yt.

Fixes #3071.

### DIFF
--- a/setupext.py
+++ b/setupext.py
@@ -30,6 +30,12 @@ int main() {
 }
 """
 
+CPPCODE = """
+int main() {
+    auto lambda = [](auto x, auto y) {return x + y;};
+}
+"""
+
 
 @contextlib.contextmanager
 def stdchannel_redirected(stdchannel, dest_filename):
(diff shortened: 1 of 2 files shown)


matthewturk (approved):
Looks great - thank you for doing this. I've run into this issue.

opaco:
Following up our Slack conversation, this fix for gcc version ultimately doesn't solve the error that I get when I try to load yt in Python.  Briefly, the error is
```   File "yt/utilities/lib/misc_utilities.pyx", line 1, in init yt.utilities.lib.misc_utilities # distutils: libraries = STD_LIBS ```
``` ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject```

File: setupext.py [hunk 1]
```diff
@@ -30,6 +30,22 @@
...
+struct node {
+    std::vector<int> vic;
+    bool visited = false;
+};
+
+int main() {
+    return 0;
+}
+"""
+
```
neutrinoceros, line 48: Should this be moved to the one function that uses it ?

setupext.py
```diff
@@ -145,6 +161,47 @@ def check_for_openmp():
...
+    # Create a temporary directory
+    ccompiler = new_compiler()
+    customize_compiler(ccompiler)
+
+    tmp_dir = tempfile.mkdtemp()
+    start_dir = os.path.abspath(".")
+
+    ok = True
+    try:
+        os.chdir(tmp_dir)
```
neutrinoceros, line 174: Two reasons:
- os.chdir should not produce any error that's caught in this try block
- removing the return variable (see following suggestion)
```suggestion
    os.chdir(tmp_dir)
    try:
```

setupext.py
```diff
@@ -145,6 +161,47 @@ def check_for_openmp():
...
+        # Compile, link, and run test program
+        with stdchannel_redirected(sys.stderr, os.devnull):
+            ccompiler.compile(
+                ["test_cpp14.cpp"], output_dir="objects", extra_postargs=compile_flags
+            )
+    except CompileError:
+        ok = False
+    finally:
+        os.chdir(start_dir)
+    return ok
```
neutrinoceros, line 190: reason: #3062
(you may wonder if the `finally` part is still executed after a return statement, and the answer is: yes 🎉 )
```suggestion
            )
           return True
    except CompileError:
        return False
    finally:
        os.chdir(start_dir)
```

setupext.py
```diff
@@ -145,6 +161,47 @@ def check_for_openmp():
...
+
+
+def check_CPP14_flags(possible_compile_flags):
+    for flags in possible_compile_flags:
+        if check_CPP14_flag([flags]):
+            return flags
+
+    log.warn(
+        "Your compiler seems to be too old to support C++14. "
+        "yt may not be able to be compiled. Please use a newer version."
```
neutrinoceros, line 200:
```suggestion
        "yt may not be able to compile. Please use a newer version."
```

neutrinoceros (approved):
left a couple of minor suggestions

setupext.py
cphyc: I moved it together with another chunk of code that's checking for openmp support.

PR #3073, opened 2021-02-10
Update install doc

neutrinoceros:
## PR Summary

- add an `--upgrade` flag to pip install instruction to mitigate the problem with users getting broken installations when compiling with a more recent numpy than the one they have installed.
- fix an out of date indication for minimal python version

### DIFF
--- a/doc/source/installing.rst
+++ b/doc/source/installing.rst
@@ -386,7 +386,7 @@ which can be installed with ``pip``:
 
 .. code-block:: bash
 
-  $ pip install numpy cython
+  $ pip install --upgrade numpy cython
 
 You may also want to install some of yt's optional dependencies, including
 ``jupyter``, ``h5py`` (which in turn depends on the HDF5 library), ``scipy``,
--- a/doc/source/installing.rst
+++ b/doc/source/installing.rst
@@ -379,7 +379,7 @@ source include:
 
 - ``git``
 - A C compiler such as ``gcc`` or ``clang``
-- ``Python >= 3.5``
+- ``Python >= 3.6``
 
 In addition, building yt from source requires ``numpy`` and ``cython``
 which can be installed with ``pip``:


doc/source/installing.rst
```diff
@@ -379,14 +379,14 @@ source include:
...
-- ``Python >= 3.5``
+- ``Python >= 3.6``
 
 In addition, building yt from source requires ``numpy`` and ``cython``
 which can be installed with ``pip``:
 
 .. code-block:: bash
 
-  $ pip install numpy cython
+  $ pip install --upgrade numpy cython
```
munkm, line 389: Maybe we should have two separate lines here, one with the original command and then another that says something like 

> if you want to have the newest numpy and cython available, you can request that the specified packages are upgraded with the `upgrade` flag
>
> ` $ pip install --upgrade numpy cython`

neutrinoceros: it's not just a matter of "wanting" the latest version. With the current state of `pyproject.toml`, running `pip install -e .` for yt _will_ download the latest numpy version and use it to compile yt. At the moment, this _guarantees_ a faulty installation except for people who already upgraded numpy.

munkm: I guess I'm confused here, because this particular part of the docs is about building a fresh install of yt. If I execute `$pip install --upgrade numpy cython` in a fresh environment and then do `$ pip install -e .`, I do not have a faulty installation of yt. 

From what I recall in the dev discussion, the faulty installs were from when conda and pip had differing versions of numpy available and this ended up causing an issue. I would support a callout box here mentioning that there can be issues when packages are being released with different versions on conda/pip. In the box we can suggest a few options, including the --upgrade one.

munkm: We also have a different section in the docs for building from source for conda-based installs where we should make this note too.

neutrinoceros: > If I execute $pip install --upgrade numpy cython in a fresh environment and then do $ pip install -e ., I do not have a faulty installation of yt.

True, but I don't expect most readers to create fresh environments. We had this problem reported in various forms by at least 3 persons over the past week.

> From what I recall in the dev discussion, the faulty installs were from when conda and pip had differing versions of numpy available and this ended up causing an issue

I _think_ that's a different issue from what I'm trying to address here. Here the problem is that pyproject.toml defines the possible numpy versions for compiling yt with `pip install -e .`, but it doesn't matter to pip what version of numpy may already be present in the env: it will download its own (and with the current specs, it will get the latest). This results in faulty install unless the user also happens to have the latest numpy version installed already.

> We also have a different section in the docs for building from source for conda-based installs where we should make this note too.

I don't use conda anymore so I'm not sure I can do a decent job there (the irony being that my first contribution to yt was exactly this section...).

munkm (approved):

PR #3075, opened 2021-02-11
docs: fix missing inclusions

neutrinoceros:
## PR Summary

this fixes part of #3029

I'd like some feedback (@brittonsmith ?) for the following, currently unincluded pages
```
doc/source/analyzing/domain_analysis/index.rst
doc/source/analyzing/domain_analysis/xray_data_README.rst
```
Are they still relevant to the main project ? If so, where should they be referenced ?

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -33,7 +33,7 @@ Table of Contents
       <tr valign="top">
        <td width="25%">
          <p>
-           <a href="yt4differences.html">yt 3.0</a>
+           <a href="yt4differences.html">yt 4.0</a>
          </p>
        </td>
        <td width="75%">
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -33,6 +33,16 @@ Table of Contents
       <tr valign="top">
        <td width="25%">
          <p>
+           <a href="yt4differences.html">yt 3.0</a>
+         </p>
+       </td>
+       <td width="75%">
+         <p class="linkdescr">How yt-4.0 differs from past versions</p>
+       </td>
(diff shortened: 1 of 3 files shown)


doc/source/index.rst
```diff
@@ -31,6 +31,16 @@ Table of Contents
        </td>
      </tr>
       <tr valign="top">
+       <td width="25%">
+         <p>
+           <a href="yt4differences.html">yt 3.0</a>
```
munkm, line 36:
```suggestion
           <a href="yt4differences.html">yt 4.0</a>
```

doc/source/reference/api/api.rst
```diff
@@ -721,7 +721,6 @@ Function List
 
 .. autosummary::
 
-   ~yt.convenience.load
```
munkm, line 724: wasn't this removed in yt-project/yt#3066

munkm (commented):
`/analyzing/domain_analysis/index` will be included after yt-project/yt#3066 is merged. 

I think it was not included by an oversight. This module used to be called `analysis_modules` and I suspect the docs were not updated.

doc/source/index.rst
neutrinoceros: good catch !

doc/source/reference/api/api.rst
neutrinoceros: apparently yeah

neutrinoceros: since the change is identical, it's not causing a conflict, but indeed this is now redundant.

PR #3077, opened 2021-02-15
CI: enforce consistent string formatting syntax in logger calls

neutrinoceros:
## PR Summary

This [flake8 plugging](https://github.com/globality-corp/flake8-logging-format) enforces the change that was performed in #2820 by @cphyc (and more). I'm also fixing a couple of mistakes that survived the previous PR or were introduced since then: Using this plugging will provide stability.

edit 1: here's the report from running this at a point before #2820
[g.log](https://github.com/yt-project/yt/files/5980515/g.log)

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -929,9 +929,10 @@ class Dataset(abc.ABC):
         if to_array:
             if any(x.units.is_dimensionless for x in coords):
                 mylog.warning(
-                    f"dataset {self} has angular coordinates. "
+                    "dataset `%s` has angular coordinates. "
                     "Use 'to_array=False' to preserve "
-                    "dimensionality in each coordinate."
+                    "dimensionality in each coordinate.",
+                    str(self),
                 )
 
             # force conversion to length
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -443,9 +443,10 @@ class RAMSESDomainSubset(OctreeSubset):
     def retrieve_ghost_zones(self, ngz, fields, smoothed=False):
         if smoothed:
             mylog.warning(
-                f"{self}.retrieve_ghost_zones was called with the "
-                f"`smoothed` argument set to True. This is not supported, "
-                "ignoring it."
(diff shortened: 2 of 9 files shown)


yt/data_objects/static_output.py
```diff
@@ -929,9 +929,10 @@ def _find_extremum(self, field, ext, source=None, to_array=True):
         if to_array:
             if any(x.units.is_dimensionless for x in coords):
                 mylog.warning(
-                    f"dataset {self} has angular coordinates. "
+                    "dataset `%s` has angular coordinates. "
                     "Use 'to_array=False' to preserve "
-                    "dimensionality in each coordinate."
+                    "dimensionality in each coordinate.",
+                    str(self),
```
cphyc, line 935: Why `str(self)` and not `self`?

yt/frontends/ramses/data_structures.py
```diff
@@ -950,9 +951,10 @@ def read_namelist(self):
                     nml = f90nml.read(f)
             except ImportError as e:
                 nml = f"An error occurred when reading the namelist: {str(e)}"
-            except (ValueError, StopIteration) as e:
+            except (ValueError, StopIteration) as err:
                 mylog.warning(
-                    "Could not parse `namelist.txt` file as it was malformed: %s", e
+                    "Could not parse `namelist.txt` file as it was malformed:",
+                    exc_info=err,
```
cphyc, line 957: What does this do?

cphyc (approved):
Looks good to me!

yt/data_objects/static_output.py
neutrinoceros: it's equivalent in this context, but IMO it's a bit clearer to perform the conversion explicitly. Could be convinced otherwise though.

yt/frontends/ramses/data_structures.py
neutrinoceros: that's the recommended way to log a captured traceback (I learned about it with this tool).

yt/data_objects/static_output.py
cphyc: Fair enough, I don't think this matters.

Issue #3074, opened 2021-02-11
Technical debt with deps specifications


neutrinoceros:
### Bug report

**Bug summary**

Currently, there are several files specifying constraints for our deps. I'll try to list them all here with short descriptions

- `setup.py`: defines mandatory runtime dependencies (update: migrated to setup.cfg in #3078)
- `pyproject.toml`: defines only Cython and numpy as compile-time deps
- `tests/test_prerequirements.txt`: again, constrains for Cython and numpy, but this time it's needed because Cartopy (an optional dep to us) doesn't define its own compile time deps correctly (?)
- `tests/test_requirements.txt`: contains all optional deps of yt and (mostly) up-to-date versions of mandatory deps
- `tests/ci_install.sh`: contains "minimal" requirements for  Cython and numpy (for Cartopy again) 
- `tests/test_minimal_requirements.txt`: defines minimal mandatory runtime deps

Evidently this is getting pretty hard to maintain, especially in PRs like https://github.com/yt-project/yt/pull/2917/ (ok maybe the plural isn't appropriate)... and quite confusing for everyone.

It seems to me that, for the most part (i.e. ignoring the Cartopy Problem™️), the problem can be addressed by using _only_™️  `setup.cfg` (or `setup.py`) to specify minimal runtime deps
```ini
# this illustrates the syntax, content is irrelevant
[options]
install_requires =
    setuptools>=19.6
    matplotlib>=1.5.3; python_version < 3.0
    sympy>=1.2
    numpy>=1.10.4
    IPython>=1.0
    unyt>=2.7.2
    more_itertools>=8.4
    tqdm>=3.4.0
    toml>=0.10.2
```

and a "full" env could be specified as
```ini
[options.extras_require]
# again, content is irrelevant
full =
    requests>=2.20.0
    netCDF4>=1.5.3
    ...
```
though I admit it is not clear to me if this would still work with the most recent versions of pip.
Note that the identifier `"full"` is purely arbitrary.

### The Cartopy Problem ™️ 

It seems a lot of the technical debt described above is due to the fact Cartopy isn't trivial to install for non-conda environments. However our own tests are run with conda so the setup could still be simplified (I think).

Moreover, we have this helpful comment

>    Getting cartopy installed requires getting cython and numpy installed
>    first; this is potentially going to be fixed with the inclusion of
>    pyproject.toml in cartopy.

[...]

PR #3078, opened 2021-02-15
Migrate metadata from setup.py to setup.cfg

neutrinoceros:
## PR Summary

This is a first response to #3074, collecting metadata from setup.py and storing it in setup.cfg instead.
This migration is performed semi-automatically with https://github.com/asottile/setup-py-upgrade
Also add a [pre-commit hook](https://github.com/asottile/setup-cfg-fmt) to keep setup.cfg tidy.

Why is setup.cfg a better place to store metadata (such as runtime dependencies) ? because it's a lot easier to query programatically
```python
>>> from configparser import ConfigParser
>>> conf = ConfigParser()
>>> conf.read("setup.cfg")
>>> print(conf["options"]["install_requires"].replace("\n", " "))
' IPython>=1.0 matplotlib>=1.5.3 more_itertools>=8.4 numpy>=1.10.4 setuptools>=19.6 sympy>=1.2 toml>=0.10.2 tqdm>=3.4.0 unyt>=2.7.2'
```
so if for any reason setup.cfg + setup.py data isn't enough in certain environments (Jenkins ? @Xarthisius), we don't need an extra specification file like that will inevitably grow out of sync (as is already the case).

note : `pyproject.toml` is getting incrementally more support throughout the years (see for instance [PEP 621](https://www.python.org/dev/peps/pep-0621/)) , and I think `setup.cfg`'s metadata will eventually become a subset of the area covered by thing file. It is not the case at the time of writing.


neutrinoceros:
this is a pure coincidence but I went to look at pandas for guidance and apparently they just merge a similar PR 5h ago... https://github.com/pandas-dev/pandas/pull/38852

setup.cfg
```diff
@@ -1,35 +1,86 @@
...
+    Topic :: Scientific/Engineering :: Visualization
+keywords = astronomy astrophysics visualization amr adaptivemeshrefinement
+project_urls =
+    Homepage=https://yt-project.org/
+    Documentation=https://yt-project.org/doc/
+    Source=https://github.com/yt-project/yt/
+    Tracker=https://github.com/yt-project/yt/issues
 
+[options]
+packages = find:
```
cphyc, line 38: Wouldn't this fit instead in `pyproject.toml`?

cphyc (approved):
Overall this looks great, though I have to confess I am totally confused by the usage (in the Python ecosystem) of `setup.py`, `setup.cfg` and `pyproject.toml`. Which one should hold which information?

matthewturk:
I'm similarly confused as @cphyc !  I can never keep the usage of the three files correct in my head.

setup.cfg
neutrinoceros: I don't think so.

neutrinoceros:
@cphyc @matthewturk 
So my understanding is that `pyproject.toml` is destined to hold everything but doesn't live up to its definition atm.
For now, it _has_ to hold compile-time dependencies, or more exactly, it specifies the build system.

`setup.py` is the file required by our build particular system (setuptools, which happens to be the default option) to compile/install the package.

`setup.cfg` can be used to declare a lot of the data needed by the build system we're using (`setuptools.setup()`), but only the bits that are serialisable. The fancy stuff (custom classes) are still declared in `setup.py`.

The end goal is to have a little build/runtime information running around as possible.
In the current state of Python, I believe this means having these three files _and nothing else_. Requirements files should only be needed in projects that depend on stuff that's not on Pypi (or conda), because they can hold the url of a desired source.

This specific PR isn't _necessary_ but it's improving maintainability and makes it easier to reuse the data currently declared in `setup.py` (most notably runtime deps) if we need to (I think our existing requirement files are mostly needed because of Jenkins ?). This opens the way to declare minimal deps (`options.install_requires`), and full test environments in a single file (`setup.cfg`) later.

Note that `pyproject.toml` isn't required in general atm (we have it because that's where black is configured), but its existence affects pip's behaviour, which is the reason why we need to specify build time deps in ours.
I'm hoping that a few minor Python release from now, `pyproject.toml` will be able to assimilate `setup.cfg`'s data, but I don't expect it to happen before a toml parser is added to the standard library.

Is it any clearer ? :)

Xarthisius:
`tests/test_*requirements.txt` exist because testing assumes all optional dependencies (that are not expressed anywhere else) are mandatory to check the relevant code paths. As long as they're expressed *somewhere* it's fine from testing POV.

neutrinoceros:
Excellent, then I'm game for moving everything to setup.cfg but I expect it will take at least some iteration to get right, so I'd like to do it in a follow up PR and leave this one as automated as can be :)

setup.cfg
neutrinoceros: For reference, these two PEPs are ongoing
https://www.python.org/dev/peps/pep-0621/
https://www.python.org/dev/peps/pep-0631/

neutrinoceros:
removing the infrastructure label because it triggers the currently broken answer tests

cphyc (approved):

brittonsmith (approved):

PR #3081, opened 2021-02-22
Fix mistake in call to partial function

cphyc:
## PR Summary

Steps to reproduce:
```python
import yt
ds = yt.load_sample("D9p_500")
ad = ds.all_data()
ad["particle_velocity_x"]
```

## PR Checklist

### DIFF
--- a/yt/frontends/art/io.py
+++ b/yt/frontends/art/io.py
@@ -121,7 +121,7 @@ class IOHandlerART(BaseIOHandler):
                 off = 1.0 / dd
                 tr[field] = rp(fields=[ax])[0] / dd - off
             if fname.startswith(f"particle_velocity_{ax}"):
-                (tr[field],) = rp(["v" + ax])
+                (tr[field],) = rp(fields=["v" + ax])
         if fname.startswith("particle_mass"):
             a = 0
             data = np.zeros(npa, dtype="f8")
--- a/yt/frontends/art/tests/test_outputs.py
+++ b/yt/frontends/art/tests/test_outputs.py
@@ -19,6 +19,7 @@ _fields = (
     ("gas", "temperature"),
     ("all", "particle_mass"),
     ("all", "particle_position_x"),
+    ("all", "particle_velocity_y"),
 )
 
 d9p = "D9p_500/10MpcBox_HartGal_csf_a0.500.d"


brittonsmith (approved):

matthewturk (approved):

yt/frontends/art/io.py
```diff
@@ -121,7 +121,7 @@ def _get_field(self, field):
                 off = 1.0 / dd
                 tr[field] = rp(fields=[ax])[0] / dd - off
             if fname.startswith(f"particle_velocity_{ax}"):
-                (tr[field],) = rp(["v" + ax])
+                (tr[field],) = rp(fields=["v" + ax])
```
neutrinoceros, line 124: I don't understand why this is a fix; I'm not denying it is, but I'd like to gain more insight. It's clearly an _improvement_ over using a partial with positional arguments, but given 4 out of 5 arguments of `read_particles` are filled in the partial def, I'd assume the fifth is the only one that can be passed to the function. In short I don't get why this causes a change in behaviour.

cphyc: I think there is some edge-case when mixing positional/non-positional arguments. With the old version, I get the following error:
```python
> import yt
… ds = yt.load_sample('D9p_500')
… ad = ds.all_data()
… ad['particle_velocity_x']"

  File "<string>", line 4, in <module>
  File "/home/ccc/Documents/prog/yt/yt/data_objects/data_containers.py", line 248, in __getitem__
    self.get_data(f)
  File "/home/ccc/Documents/prog/yt/yt/data_objects/selection_objects/data_selection_objects.py", line 203, in get_data
    particles, self, self._current_chunk
  File "/home/ccc/Documents/prog/yt/yt/geometry/geometry_handler.py", line 209, in _read_particle_fields
    chunks, selector, fields_to_read
  File "/home/ccc/Documents/prog/yt/yt/utilities/io_handler.py", line 204, in _read_particle_selection
    for field_r, vals in self._read_particle_fields(chunks, ptf, selector):
  File "/home/ccc/Documents/prog/yt/yt/frontends/art/io.py", line 101, in _read_particle_fields
    data = self._get_field((ptype, field))
  File "/home/ccc/Documents/prog/yt/yt/frontends/art/io.py", line 124, in _get_field
    (tr[field],) = rp(["v" + ax])
TypeError: read_particles() got multiple values for argument 'idxa'
```
To be honest, I am unsure why this is happening in the first place but I am sure the new behaviour is different!

cphyc:
At the moment, the test(s) are not triggered at all. I'm working on it but in the meantime I'm switching to draft.

matthewturk (approved):

yt/frontends/art/io.py
neutrinoceros: fair enough

cphyc:
@matthewturk @brittonsmith @neutrinoceros  this is ready to go (I just need to bump the answer tests for Jenkins). However, it seems #3087 would require bumping the answer tests _again_. My question is thus: should we fix #3087 in this PR before merging or submit the answer tests knowing the velocities may be off?

matthewturk:
I think whatever gets this in and closed is best.

neutrinoceros:
We lost the report, I need to rerun tests
@yt-fido  test this please

munkm (approved):

munkm:
I think it's fine to not fix yt-project/yt#3087 in this PR. We can do a follow up PR for that, and now that we have the submodule for answers I don't think it's a huge issue to have lots of answer-store commits. Unless I'm missing something?

neutrinoceros:
> I don't think it's a huge issue to have lots of answer-store commits. Unless I'm missing something?

FWIW, I think it's already hard enough to properly handle a single PR there at a time, I don't think anyone wants to abuse them if it can be avoided. But if we need 2 updates instead of one, that's fine by me :)

PR #3082, opened 2021-02-22
fix a FutureWarning from numpy

neutrinoceros:
## PR Summary

Here's the warning as it pops up on Jenkins 
```
/var/jenkins_home/workspace/yt_docs_new/yt/visualization/volume_rendering/render_source.py:906: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  self.current_image[:, :, 0][locs] = color[0]
```
this is not exactly a bug for now but it will be in the future so I'm using the bug label.

### DIFF
--- a/yt/visualization/volume_rendering/render_source.py
+++ b/yt/visualization/volume_rendering/render_source.py
@@ -901,7 +901,7 @@ class MeshSource(OpaqueSource):
         if color is None:
             color = np.array([0, 0, 0, alpha])
 
-        locs = [self.sampler.amesh_lines == 1]
+        locs = (self.sampler.amesh_lines == 1,)
 
         self.current_image[:, :, 0][locs] = color[0]
         self.current_image[:, :, 1][locs] = color[1]
--- a/doc/source/analyzing/units.rst
+++ b/doc/source/analyzing/units.rst
@@ -210,8 +210,8 @@ Using parsec as an example,
   * ``pc/h``
     Proper parsecs, normalized by the scaled hubble constant, :math:`\rm{pc}/h`.
 
-Overriding Code Unit Defintions
--------------------------------
+Overriding Code Unit Definitions
+--------------------------------
 
 On occasion, you might have a dataset for a supported frontend that does not
 have the conversions to code units accessible or you may want to change them


neutrinoceros:
The warning was indeed resolved, now opening for review

yt/visualization/volume_rendering/render_source.py
```diff
@@ -901,7 +901,7 @@ def annotate_mesh_lines(self, color=None, alpha=1.0):
         if color is None:
             color = np.array([0, 0, 0, alpha])
 
-        locs = [self.sampler.amesh_lines == 1]
+        locs = tuple(self.sampler.amesh_lines == 1)
```
matthewturk, line 904: Are there any times this could result in an expensive array-to-tuple conversion? Would it be better to do `(self.sampler..., )`?

neutrinoceros: good point, indeed the conversion is expensive for large arrays.
I now think the best approach would simply be
```python
     locs = self.sampler.amesh_lines == 1
```

neutrinoceros: Welp I won't investigate why this broke the the docs build. I'm switching to your suggestion

yt/visualization/volume_rendering/render_source.py
```diff
@@ -901,7 +901,7 @@ def annotate_mesh_lines(self, color=None, alpha=1.0):
         if color is None:
             color = np.array([0, 0, 0, alpha])
 
-        locs = [self.sampler.amesh_lines == 1]
+        locs = (self.sampler.amesh_lines == 1,)
```
cphyc, line 904: I am confused as to why
```suggestion
        locs = self.sampler.amesh_lines == 1
```
doesn't work. Are you sure the issue isn't on the doc build's side?

neutrinoceros: I'm confused as well but I'm not confident that those _should_ be equivalent as far as numpy slicing is concerned.
I'm not sure of anything but the current state seems to do the job and at least it's pretty obvious that it stays faithful to the original code.

PR #3083, opened 2021-02-22
fix a NameError in yt.mods

neutrinoceros:
## PR Summary

I'm not actually convinced that the file should be kept (dead code ?) but it's easy enough to fix the NameError one gets with `import yt.mods` atm

### DIFF
--- a/yt/mods.py
+++ b/yt/mods.py
@@ -4,6 +4,8 @@
 
 import os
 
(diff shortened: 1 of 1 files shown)


matthewturk (approved):
I think removing mods is a good first-after-4 change. Thanks for fixing this.

PR #3089, opened 2021-02-23
feature: cleanup (rewrite) load_sample

neutrinoceros:
## PR Summary
This is another take at #2946, focusing on the fundamental bits so that the size of the PR remains manageable for me as the author _and_ for reviewers. My hope is that this can be looked at and accepted before the 4.0 release, because the feature has not be shipped yet and I'd like to seize the opportunity to make backward incompatible changes while the feature is still considered unstable.

### refactor yt.load_sample with the following benefits
- no magic is used in determining if the data needs untaring
- improved  modularity: the core functionalities are separate functions
- download (and untar) data files to more predictable location (either test_data_dir or cwd if unset)
- remove the `specific_file` keyword argument and instead integrate the functionality natively within the first argument.

The last two effects combined mean that `yt.load_sample` in a script can be replaced transparently with `yt.load` after the first successful call (after data is downloaded). Reversely, `yt.load` can be replaced with `yt.load_sample` in scripts and tests that use sample datasets to make them more robust.

### add (experimental) user facing utility for data sample registry exploration
`yt.sample_data.api.get_data_registry_table`

### Note to reviewers
I think I've covered the scenarios I expect the notebook from #2946 (those should eventually be rewritten as tests), but please try this out and try to break it.

### todo
- [x] fix the existing testing issue (logger is deactivated in CI, preventing the test to capture logs) -> this will be much easier after #3102 and #3106 are merged)
- [ ] add more tests (especially deal with at least one case of "specific file")

### possible follow up
one of the functions I added here, namely `yt.sample_data.api.lookup_on_disk_data`, should be used in `yt.loaders.load` as well (code deduplication), and possibly in other places as well.

⚠️ : this is now based on top of  #3169


neutrinoceros:
my test is failing because the config used for running test involved disabling stream logging.
@matthewturk @Xarthisius , how would you feel if I changed `yt.utilities.logger.disable_stream_logging` slightly so that instead of using a `logging.NullHandle`, we'd just set the log level to 100 (effectively suppressing all logs) ?
This would allow me to temporarily activate stream logging for the duration of the test.

neutrinoceros:
Fix #3097

neutrinoceros:
So the plan for this is that @jzuhone and @munkm are going to review the code.
I'll need to add documentation of the new user-facing function `get_data_registry_table` once it's stable review-wise and then maybe Stephanie could come in and review the docs.
I should also refactor the tests to use pytest fixtures, which will make them much cleaner and easier to maintain. In order to validate this, I need
- a mechanism to turn the logger on at least for these tests in CI. @Xarthisius, any advice for this part ?
- a way to tell the nose runner to skip these tests since they won't be compatible with it. @jcoughlin11 , how should I do that ?

yt/sample_data/api.py
```diff
@@ -0,0 +1,174 @@
...
+    # load local data
+    pooch_json = json.load(
+        pkg_resources.resource_stream("yt", "sample_data_registry.json")
+    )
+    pooch_table = pd.DataFrame(pooch_json.values())
+
+    # merge tables
+    unified_table = website_table.merge(pooch_table, on="url", how="outer")
+
+    # unified_table.set_index("filename", inplace=True)
```
chrishavlin, line 108: maybe add a comment here explaining why you're leaving the commented code (I believe it's the same reason as the note in `load_sample`)?

chrishavlin (commented):
I know this PR is still a draft but I'm excited to use this! 

A few minor code suggestions, but my main question is: 

what is the expected/desired behavior when someone is offline? 

If I'm following the code right, if someone calls `yt.load_sample` while offline the attempt to fetch the `datafiles.json` will raise an error. But if they have previously called `load_sample` for a datafile (while online), we could theoretically still load that file. I could see it being tricky to implement, but I do think a new yt user could be confused about why `load_sample` is failing when they ran it before and already downloaded the data. So if we want to require an internet connection for this functionality, I think it would help new yt users if we stress this requirement in the documentation for `load_sample` and the public methods in `sample_data.api`. And maybe the error message could suggest something along the lines of "try using `yt.load()` if you're offline and you've previously downloaded the file"?

yt/sample_data/api.py
```diff
@@ -0,0 +1,174 @@
...
+This is a collection of helper functions to yt.load_sample
+"""
+import json
+import re
+from functools import lru_cache
+from itertools import chain
+from pathlib import Path
+
+import pkg_resources
+import requests
```
neutrinoceros, line 11: note to self: requests is still an optional dep, I need to update this with an on-demand import

yt/sample_data/api.py
```diff
@@ -0,0 +1,174 @@
...
+    api_url = "https://raw.githubusercontent.com/yt-project/website/master/data/datafiles.json"  # noqa
+
+    response = requests.get(api_url)
+
+    if not response.ok:
+        raise RuntimeError(
+            "Could not retrieve registry data. Please check your network setup."
+        )
+
+    website_json = json.loads(response.text)
```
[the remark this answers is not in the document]
neutrinoceros: nice ! You can tell I'm not used to requests :)

yt/sample_data/api.py
```diff
@@ -0,0 +1,174 @@
...
+    # this dict follows this schema: {frontend_name: {flat dataframe-like}}
+
+    columns = ["code", "filename", "size", "url"]
+    website_table = pd.concat(pd.DataFrame(d) for d in website_json.values())[columns]
```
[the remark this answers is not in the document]
neutrinoceros: I don't remember tbh. I may not have had a good reason besides keeping things as tidy as possible to make it work. I'm fine with adding it back.

yt/sample_data/api.py
neutrinoceros: exactly ! thank you so much for figuring this out. I'll update this.

yt/utilities/on_demand_imports.py
```diff
@@ -232,25 +232,6 @@ def __version__(self):
...
-        if self._pooch is None:
-            try:
-                import pooch as pooch
-            except ImportError:
-                pooch = NotAModule(self._name)
-            self._pooch = pooch
-        return self._pooch
-
-
-_pooch = pooch_imports()
```
munkm, line 251: Why was this deleted from on_demand_imports? We still need pooch here.

munkm (commented):
Minor comment to start. Will review more soon.

yt/utilities/on_demand_imports.py
neutrinoceros: indeed. In fact I don't remember doing this, this may be an rebasing accident !

neutrinoceros: fixed !

yt/sample_data/api.py
```diff
@@ -0,0 +1,172 @@
...
+    raise FileNotFoundError(err_msg)
+
+
+# create a pooch.Pooch instance
+# pardon the poor variable naming: it is common practice to give those dog names...
+data_registry = get_data_registry_table()
+cache_storage = _get_test_data_dir_path() / "yt_download_cache"
+pooch_registry = {k: v["hash"] for k, v in sample_data_registry.items()}
+base_url = "https://yt-project.org/data/"
+poochie = pooch.create(path=cache_storage, base_url=base_url, registry=pooch_registry)
```
munkm, line 156:
```suggestion
poochie = pooch.pooch.create(path=cache_storage, base_url=base_url, registry=pooch_registry)
```

yt/loaders.py
```diff
@@ -1268,90 +1268,102 @@ def flatten(l):
...
 
-    if fn is None:
-        keys = []
-        for key in fido._registry:
-            for ext in _extensions_to_strip:
-                if key.endswith(ext):
-                    key = key[: -len(ext)]
-            keys.append(key)
-        return keys
+    pooch_logger = pooch.utilsget_logger()
```
munkm, line 1303:
```suggestion
    pooch_logger = pooch.pooch.utils.get_logger()
```

munkm (commented):
This is round 1! I will come back later! Thanks for working on this!!!!

stonnes (commented):
Nice, I just have a few comments on the explanation bits!  Then we can work on narrative docs to make sure the examples all line up with the finalized version.

yt/sample_data/api.py
```diff
@@ -0,0 +1,172 @@
...
+import re
+from functools import lru_cache
+from itertools import chain
+from pathlib import Path
+
+import pkg_resources
+
+from yt.funcs import mylog
+from yt.utilities.on_demand_imports import (
+    _pandas as pandas,
```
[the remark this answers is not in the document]
neutrinoceros: oh it's not ? Well I'll open a different PR to make it so then, because pandas is _definitely_ not a hard dependency. Not counting this line, it's only imported in 3 places in the code, 2 of which are not top level, and the third is in a try-except block.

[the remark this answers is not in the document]
neutrinoceros: #3148 !

yt/sample_data/api.py
```diff
@@ -0,0 +1,172 @@
...
+    if not response.ok:
+        raise RuntimeError(
+            "Could not retrieve registry data. Please check your network setup."
+        )
+
+    website_json = response.json()
+    # this dict follows this schema: {frontend_name: {flat dataframe-like}}
+
+    columns = ["code", "filename", "size", "url", "description"]
+    website_table = pd.concat(pd.DataFrame(d) for d in website_json.values())[columns]
```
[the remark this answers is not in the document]
neutrinoceros: damn I have no idea how this got into such a broken state, thanks for catching it !

yt/sample_data/api.py
neutrinoceros: I don't think that's correct
Screenshot 2021-03-25 at 21 30 29

yt/loaders.py
neutrinoceros: This is half correct. I'm indeed missing a `.`

neutrinoceros: should be fixed now, I'll let you resolve

yt/sample_data/api.py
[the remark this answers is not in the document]
neutrinoceros: should be fixed now, I'll let you resolve

yt/sample_data/api.py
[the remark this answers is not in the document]
neutrinoceros: This PR is now rebased on top of the aforementioned on.

yt/sample_data/api.py
munkm: It is for the on demand import version 

image

yt/loaders.py
```diff
@@ -1268,90 +1268,102 @@ def flatten(l):
...
+    `yt.sample_data.api.get_data_registry_table`.
+
+    This function requires pandas and pooch to be installed.
+
     Parameters
     ----------
-    fn : str or None
-        The name of the sample data to load. This is generally the name of the
-        folder of the dataset. For IsolatedGalaxy, the name would be
-        `IsolatedGalaxy`.  If `None` is supplied, the return value
```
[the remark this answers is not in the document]
neutrinoceros: The reason I cut it out is that it was all kinds of broken. An important number of names it used to display were actually not directly usable
I really don't thing this feature should be shipped without being more thoroughly tested.
Fixing this means fixing the underlying metadata, which we definitely ought to do but is also very hard without appropriate infrastructures: one needs to systematically test that all datasets can be downloaded, and manually find and fix those that can't.
This version of load_sample and its building blocks give us a better environment to do exactly that, so I'd argue that it should still go in without the feature.

yt/loaders.py
```diff
@@ -1268,90 +1268,102 @@ def flatten(l):
...
+    # however we don't want to do it right now because the "filename" column is
+    # currently incomplete
+    specs = registry_table.query(f"`filename` == '{topdir}'").iloc[0]
 
-    downloader = None
-    if pbar:
-        downloader = pooch.pooch.HTTPDownloader(progressbar=True)
+    if not specs["load_name"]:
+        raise ValueError(
+            "Registry appears to be corrupted: could not find a 'load_name' entry for this dataset."
```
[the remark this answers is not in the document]
neutrinoceros: oh but _is_ currently corrupted: we're missing many metadata points, making this feature inherently broken for a number of sample datasets.
see for instance
https://github.com/yt-project/yt/pull/3118 (still not fixed)
or
https://github.com/yt-project/yt/pull/3121

The reason this error is not "could not find ..." is that the line _before_ that will already error out in cas the requested dataset isn't findable, but we still need this check because recognising the entry doesn't mean we're able to load (currently).

yt/sample_data/api.py
neutrinoceros: Oh right, I forgot _some_ on_demand_import add an unnecessary layer of complexity. I strongly feel like modules obtained from `yt.utilities.on_demand_import` should behave as drop-in replacements, so I'll update _there_.

neutrinoceros: (should be fixed now)

neutrinoceros:
I finally took the time to properly rewrite the test with pytest, I hope it works in any testing env now, and in particular within the specific config we have in CI envs.
I'll still need to polish it so even if everything pass I'm not removing the draft status for now

neutrinoceros:
I added some new tests to check the (partial) sanity of the assembled registry. One of them is failing (one data sample's size is parsed as 0) so I'll fix it later

neutrinoceros:
Got it to a much better place (all tests I wrote pass locally, including the doctest ones)

yt/loaders.py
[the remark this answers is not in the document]
matthewturk: I'm of two minds on this; on the one hand, if I am close to knowing a dataset name (for instance, do I capitalize "Sloshing" in "GasSloshing"?) then having a list to compare against is useful.  But, it didn't work without additional arguments for some of the files.  (Not all.)  Maybe what *would* work is to have a hint in the docstring about how to find the registry as a dict?

yt/sample_data/api.py
```diff
@@ -0,0 +1,182 @@
...
+    <NA>
+    """
+    try:
+        s = s.upper()
+    except AttributeError:
+        # input is not a string (likely a np.nan)
+        return pd.NA
+    val = float(re.search(num_exp, s).group())
+    unit = re.search(byte_unit_exp, s).group()
+    exponent = {"B": 0, "KB": 3, "MB": 6, "GB": 9}[unit]
```
matthewturk, line 54: Genuinely surprised this isn't in the stdlib!  Also I always forget the meaning of when an `i` is in the name like `MiB` but I'll assume this is correct.

matthewturk (approved):
minor comments

yt/sample_data/api.py
neutrinoceros: 1) this may be in the stdlib, I don't remember checking this... I will update this thread after researching it more.
2) I don't think any existing value in this column is using an "i"

matthewturk: I don't think it is, I just was noting that I always get confused.

neutrinoceros: I couldn't find anything to support this in the stdlib... but what I DID find was a SO post https://stackoverflow.com/questions/12523586/python-format-size-application-converting-b-to-kb-mb-gb-tb

In fact this reminds me of an early concern from @cphyc , that I failed to address (until now), regarding the fact that 1KB = 1024B, rather than 1KB=1000B as I previously assumed.
Implementing the correct version of this parser made me realise that file sizes are often given with a limited amount of digits, which make them intrinsically fuzzy because, for instance "4.66GB" doesn't translate to an integer (interpreting `"GB"` as `*1024**3`).
Please, could one or either of you review this bit again ?

yt/sample_data/api.py
```diff
@@ -18,41 +18,54 @@
...
         s = s.upper()
     except AttributeError:
         # input is not a string (likely a np.nan)
         return pd.NA
+
     val = float(re.search(num_exp, s).group())
     unit = re.search(byte_unit_exp, s).group()
-    exponent = {"B": 0, "KB": 3, "MB": 6, "GB": 9}[unit]
-    return int(val * 10 ** exponent)
+    prefixes = ["B", "K", "M", "G", "T"]
```
matthewturk, line 66: LGTM!

yt/loaders.py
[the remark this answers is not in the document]
neutrinoceros: I was about to implement your suggestion @matthewturk, but then plot twist: this is exactly what's already done.

neutrinoceros:
I'm super confused with the test configuration right now.
I'm trying to ignore `yt/tests/test_load_sample.py` in nose runs but instead it seems that I'm cancelling the existing ignore for `yt/tests/test_load_errors.py`.
What bugs me is that is looks like _that_ file is being ignored in two files (`tests.yaml` and `nose_unit.cfg`), but the first one is apparently used by pytest as well. @Xarthisius @jcoughlin11 , can you help me, please ?

cphyc (approved):

jzuhone (approved):

nose_unit.cfg
```diff
@@ -6,5 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
-ignore-files=test_load_errors.py
+ignore-files=(test_load_errors.py|test_load_sample.py)  # tests written for pytest
```
Xarthisius, line 9: If you ran it with -vv to test it locally, you'd see that your comment is treated as a part of the regex:

```
$ nosetests -vv -c nose_unit.cfg --collect-only tests/
...
nose.config: INFO: Ignoring files matching ['(test_load_errors.py|test_load_sample.py)  # tests written for pytest']
...
```

Drop it and it will work.

Xarthisius:
> What bugs me is that is looks like _that_ file is being ignored in two files (`tests.yaml` and `nose_unit.cfg`), but the first one is apparently used by pytest as well. @Xarthisius @jcoughlin11 , can you help me, please ?

* `tests.yaml` is used by Jenkins in `py38` test that runs everything
* `nose_unit.cfg` is used by GHA
* pytest uses neither of them
* py38 will go away once we fully adopt pytest

Right now, we're in a transition period and it is what it is.

nose_unit.cfg
neutrinoceros: sorry I never had the patience to learn nose... Thanks a lot !

neutrinoceros:
> Right now, we're in a transition period and it is what it is.

I'm not complaining about that, I'm just genuinely confused :)
Thanks a lot !

munkm:
Is this still a draft?

neutrinoceros:
I just noticed that this implementation doesn't work with amrvac data samples. It likely due to how they're packaged (my fault), I'll see if the fix needs to be done here or in a separate PR.

For posterity
```python
>>> import yt
>>> yt.load_sample("amrvac_blastwave_polar_2D")
yt : [INFO     ] 2021-04-09 09:47:44,059 'amrvac_blastwave_polar_2D' is not available locally. Looking up online.
yt : [INFO     ] 2021-04-09 09:47:44,060 Downloading from https://yt-project.org/data/amrvac_bw_polar_2d.tar.gz
100%|███████████████████████████████████████| 468k/468k [00:00<00:00, 1.55GB/s]
yt : [INFO     ] 2021-04-09 09:47:46,092 Untaring downloaded file to '<my data dir>'
...
FileNotFoundError: No such file or directory: '<my data dir>/amrvac_blastwave_polar_2D/amrvac/bw_polar_2D0000.dat'.
```

PR #3092, opened 2021-02-24
[WIP] No ambiguous fields

cphyc:
Only used to spot all ambiguous field access.

I am taking advantage of the a nose plugin + our testing infrastructure
1. Catch all ambiguous field calls,
2. Generate a diff that would likely fix them,
3. Upload it using `yt upload`

Repeating the process multiple times should yield all ambiguous field access and provide a way to semi-automatically fix them. Once done, I'll remove the modifications to the  testing infrastructure and the new file I added in the trunk (`yt/auto_update_ambiguous.py` which defines the nose plugin).


cphyc:
Superseeded by #3101.

PR #3098, opened 2021-02-25
make requests a first class dependency

neutrinoceros:
## PR Summary

It seems that we in a bunch of places the code base dances around importing requests because it's not a mandatory dependency. The ways in which we avoid making it mandatory are however pretty inconsistent with each other, creating a small but noticeable amount of technical debt. Additionally, I'm suspecting that the reason it wasn't made a first class dependency originally was that the project was young and unstable. I think it is reasonable now to make it a direct and mandatory dependency, especially taking into account the fact that it is now a dependency of many other projects in the ecosystem, I'm fairly convinced almost everyone installs it already by virtue of transitive dependencies.

alternatively, if we decide not to make it a mandatory dependency, we should at least uniformise the "dancing". Here I'm just going with the easy (and easy to maintain) option.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -148,6 +148,7 @@ if __name__ == "__main__":
             "more_itertools>=8.4",
             "tqdm>=3.4.0",
             "toml>=0.10.2",
+            "requests>=2.20.0",
         ],
         extras_require={"hub": ["girder_client"], "mapserver": ["bottle"]},
         cmdclass={"sdist": sdist, "build_ext": build_ext},
--- a/tests/test_requirements.txt
+++ b/tests/test_requirements.txt
@@ -10,7 +10,6 @@ nose-timer~=1.0.0
 nose~=1.3.7
 pandas~=1.1.2
 pytest~=5.2
-requests~=2.20.0
 scipy~=1.5.0
 sympy~=1.5
 pyqt5~=5.15.2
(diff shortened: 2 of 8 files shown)


munkm:
I'm curious why this wasn't made an issue discussion, since it seems like that's the content of your message here?

neutrinoceros:
That's because writing the message takes slightly more time and doesn't reflect as well the amount of work required than doing the work itself :)

This is indeed meant as a discussion, but participants don't have to grep through the code base to learn as much as I did: it's all there.

munkm:
I worry about us putting a lot of stuff in our required dependencies. We intentionally try to keep the install_requires list minimal to avoid asking users to install unnecessary packages for their use cases. Installing requests in a fresh env when i have all of our other hard deps installed will require another 10 pacakges to be installed. Is this necessary to ask every user to do?

neutrinoceros:
Oh it's certainly not _necessary_. It's just more convenient from a dev perspective. I know I don't have a strong case with this. The less controversial route is to use the NotAModule class within on_demand_imports, I'll open an alternative PR for this soon.

cphyc:
Thanks for doing this! I think #3100 is the best solution to this issue.

neutrinoceros:
Cool, then I can close this one

PR #3100, opened 2021-02-25
make requests a part of yt.utilities.on_demand_imports

neutrinoceros:
## PR Summary
This is an alternative PR to #3098, and hopefully less controversial

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -2559,7 +2559,7 @@ class YTSurface(YTSelectionContainer3D):
 
     @parallel_root_only
     def _upload_to_sketchfab(self, data, files):
-        import requests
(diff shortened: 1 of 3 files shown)


cphyc (approved):

neutrinoceros:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

munkm (commented):
I agree that this is a better solution than a hard dependency!! 

I think you might want to add requests to the testing requirements files/infrastructure here too.

neutrinoceros:
Oh right that's what I'm missing, thanks !
That also reminds me I would very much like to move forward with #3078 so we can start gathering our deps (both required and optional) into one file !

matthewturk (approved):

munkm (approved):

PR #3102, opened 2021-02-26
Updates to pytest infrastructure [01/NN]

Xarthisius:
## PR Summary

This is an attempt to split https://github.com/yt-project/yt/pull/2817 into manageable chunks. Changes in this PR apply to basic pytest infrastructure. They don't provide any new answer tests yet.

## TODO

- [ ] Enable pytest unittests on GHA and jenkins
- [x] (optionally) port/add one of the answer tests to allow reviewers to run it.

## For Reviewers
Two points of interest are: `conftest.py` and `yt/utilities/answer_testing/testing_utilities.py`. The latter is a modified `yt/utilities/answer_testing/framework.py` which I intentionally committed as a separate file to a) make it easier to read b) keep the nose testing alive. 

## How to test?
1. Start with visual inspection.
2. Minimal setup to test answer tests locally:
    1. Download GDF `sedov` data and put it in your `test_data_dir`. 
    1. `mkdir /tmp/answers`
    1. `pytest -s -v --local-dir=/tmp/answers --with-answer-testing  --answer-store yt/frontends/gdf/tests/test_outputs.py`
    2.  run above without `--answer-store`.


yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,576 @@
...
+    """
+    Doc string.
+    """
+    for k, v in d.items():
+        if isinstance(v, dict):
+            h5_sub_grp = h5grp.create_group(k)
+            _parse_raw_answer_dict(v, h5_sub_grp)
+        else:
+            if not isinstance(k, str):
+                k = str(k)
```
neutrinoceros, line 288:
```suggestion
            k = str(k)
```

neutrinoceros (commented):
very quick low-hanging fruits, will review in much greater detail soon !

conftest.py
```diff
@@ -32,39 +27,56 @@ def pytest_addoption(parser):
...
-    ytcfg["yt", "internals", "within_pytest"] = True
-    # Make sure that the answers dir exists. If not, try to make it
-    if not os.path.isdir(answer_dir):
-        os.mkdir(answer_dir)
-    # Read the list of answer test classes and their associated answer
-    # file
-    with open(answer_file_list) as f:
-        # devnote: this is never used, likely a bug
-        answer_files = yaml.safe_load(f)  # noqa F841
+    yt._called_from_pytest = True
```
cphyc, line 76: Is there a specific reason not to use ` ytcfg["yt", "internals", "within_pytest"] = True`?

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
+    """
+    return request.param
 
 
-def _param_list(request):
-    r"""
-    Saves the non-ds, non-fixture function arguments for saving to
-    the answer file.
+@pytest.fixture(scope="class")
+def d(request):
```
cphyc, line 308: It would be much better to have something more explicit than just `d` (or `a`, `w`,`N`).

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
+    return request.param
 
 
-@pytest.fixture(scope="function")
-def hashing(request):
-    r"""
-    Handles initialization, generation, and saving of answer test
-    result hashes.
+@pytest.fixture(scope="class")
+def a(request):
```
cphyc, line 317:
```suggestion
def axis(request):
```

conftest.py
cphyc: I also don't quite understand how this is necessary from the docstring?

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
-        # directly to those methods, like so:
-        >>> class TestNewFrontend:
-        >>>     @pytest.mark.usefixtures('hashing')
-        >>>     def test1(self):
-        >>>         ...
-        >>>     def test2(self):
-        >>>         ...
+
+@pytest.fixture(scope="class")
+def w(request):
```
cphyc, line 326:
```suggestion
def weight_field(request):
```

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
+def ds_repr(request):
+    """
+    Fixture for returning the string representation of a dataset.
+    Needed because indirect=True is used for loading the datasets.
+    """
+    return request.param
+
+
+@pytest.fixture(scope="class")
+def N(request):
```
cphyc, line 344:
```suggestion
def Npart(request):
```
or
```suggestion
def number_of_particles(request):
```
Also if there are different particle types, which one is returned?

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
+
+@pytest.fixture(scope="class")
+def big_data(request):
+    """
+    There isn't a way to access command-line options given to pytest from
+    an actual test, so we need a fixture that retrieves and returns its
+    value. In this case, the --answer-big-data option.
+    """
+    if request.config.getoption("--answer-big-data"):
+        return True
```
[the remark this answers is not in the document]
cphyc: Note that there's a typo in @neutrinoceros comment (there's a trailing `:` that shouldn't be there).

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,576 @@
+import functools
```
cphyc, line 1: This file looks very much similar to `yt/utilities/answer_testing/utils.py`. Any idea how to compare them?

pytest_unit.ini
```diff
@@ -0,0 +1,8 @@
+[pytest]
+addopts =
+    -s
+    -v
+    -rs
+    --ignore-glob='*_nose.py'
+    --ignore='yt/frontends/owls_subfind/tests/test_outputs.py'
+    --ignore='yt/frontends/ramses/tests/test_outputs.py'
```
cphyc, line 8: Why are these  specific tests skipped?

jcoughlin11:
@Xarthisius This is great! Thanks for doing this!

pytest_unit.ini
jcoughlin11: I was told a while ago that there was an issue with ramses. As for owls, it just took absolutely forever to run, so I skipped it in the interest of time. I actually totally forgot they were skipped until seeing this comment, though.

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: It's the same file. I just renamed it due to a comment that `utils.py` wasn't a very good file name.

conftest.py
jcoughlin11: This change (as well as the ones below to `N` and `w`) will require changing the name of the passed fixture in every test that uses it. These fixtures are really only necessary because of the use of `indirect=True` when parameterizing the tests. This was needed in order to get the loading of the datasets correct, if I recall correctly.

conftest.py
jcoughlin11: See the comment I left below on the change to `a`

conftest.py
jcoughlin11: Not that I'm aware of. I just didn't know that was a thing.

conftest.py
cphyc: I can see this will cause a lot of changes, but the issue is that we then end up with tests that look like this
```python
@pytest.mark.usefixtures("hashing")
    @pytest.mark.parametrize("ds", [d9p], indirect=True)
    @pytest.mark.parametrize("f", fields, indirect=True)
    @pytest.mark.parametrize("w", weights, indirect=True)
    @pytest.mark.parametrize("d", objs, indirect=True)
    @pytest.mark.parametrize("a", axes, indirect=True)
    def test_d9p_pixelized_projection_values(self, f, d, a, w, ds):
        [...]
```
which I find extremely hard to read, since the variables aren't self explanatory.

pytest_unit.ini
cphyc: I don't think there is an issue with ramses (as the maintainer of the frontend). Maybe there's an issue with ramses and pytest though. In any case these tests should be revived before this goes in.

conftest.py
cphyc: This may has probably been merged since you forked the repository. I'd suggest using it to prevent using a global variable :)

pytest_unit.ini
Xarthisius: Don't worry about ignored test at the moment. Those ones are answer tests and this PR doesn't enable any answer tests yet.

yt/utilities/answer_testing/testing_utilities.py
Xarthisius: I'd suggest not comparing them. Treat it as what it is in this PR: a completely new file. This will allow to get rid of any cruft that accumulated in there over the years.

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,578 @@
...
+        Test results to be saved.
+
+    outputFile : str
+        Name of file to save results to.
+    """
+    with open(outputFile, "a") as f:
+        yaml.dump(data, f)
+
+
+def _compare_result(new_data, old_data):
```
Xarthisius, line 223: Is that method really necessary? Why not simple `==`?

```
>>> a = dict(a=1,b=2)
>>> b = dict(a=1,b=2)
>>> a == b
True
>>> id(a) == id(b)
False
```

tests/tests_pytest.yaml
```diff
@@ -0,0 +1,183 @@
+TestAdaptahop:
```
Xarthisius, line 1: I don't think this file is necessary at all. For nose that was a way to pseudo parallelize tests, but pytest-xdist handles that without such tricks. Both yaml and h5 file names could be autogenerated based on test class name.

conftest.py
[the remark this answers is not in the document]
neutrinoceros: Oopsy. All the more reason not to commit this suggestion directly

munkm:
That has to be the fastest WIP to non-WIP PR ever.

tests/tests_pytest.yaml
matthewturk: That would be quite nice.

yt/frontends/gdf/tests/test_outputs.py
```diff
@@ -1,30 +1,27 @@
...
-
+@pytest.mark.answer_test
+class TestGDF:
+    answer_file = None
+    saved_hashes = None
 
-@requires_file(sedov)
-def test_GDFDataset():
-    assert isinstance(data_dir_load(sedov), GDFDataset)
+    @pytest.mark.usefixtures("hashing")
```
matthewturk, line 16: Some of the marks we use show up over and over again. I think there's a way to reduce the boilerplate with config or something, isn't there?

conftest.py
neutrinoceros: +1 @cphyc . Additionally, one-letter variable/function names are virtually impossible to grep for, making them all the more difficult to read.

pytest_unit.ini
neutrinoceros: I would still suggest we replace those ignore flags with `@pytest.mark.skip` decorators. In particular because it forces us to write down a `reason` argument, and it's harder to miss when looking a the test file itself.

yt/frontends/gdf/tests/test_outputs.py
neutrinoceros: I didn't know about `pytest.mark.usefixtures`, but as I'm reading the docs I don't get why it's used here, as opposed to writing

```python
def test_sedov_tunnel(self, hashing, ...)
```

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
+    try:
+        if isinstance(request.param, list):
+            # data_dir_load can take the cls, args, and kwargs. These optional
+            # arguments are given in a dictionary as the second element of the
+            # list
+            ds_fn = request.param[0]
+            opts = request.param[1]
+            cls = None if "cls" not in opts.keys() else opts["cls"]
+            args = None if "args" not in opts.keys() else opts["args"]
+            kwargs = None if "kwargs" not in opts.keys() else opts["kwargs"]
```
neutrinoceros, line 289: Here's the pythonic version
```suggestion
            cls = opts.get("cls")
            args = opts.get("args")
            kwargs = opts.get("kwargs")
```

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
+            dataset = data_dir_load(ds_fn, cls=cls, args=args, kwargs=kwargs)
```
neutrinoceros, line 290: Here's the pythonic version
```suggestion
            dataset = data_dir_load(
                ds_fn, 
                cls=opts.get("cls"),
                args=opts.get("args"),
                kwargs=opts.get("kwargs")
            )
```
```

File: conftest.py [hunk 15]
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
-        >>>     def test1(self):
-                    ...
+def ds(request):
+    try:
+        if isinstance(request.param, list):
+            # data_dir_load can take the cls, args, and kwargs. These optional
+            # arguments are given in a dictionary as the second element of the
+            # list
+            ds_fn = request.param[0]
+            opts = request.param[1]
```
neutrinoceros, line 286: This is better because it breaks in case `request.param` is longer than two elements, while the current implementation silently swallows unrecognised args.
```suggestion
            ds_fn, opt = request.param
```

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
-    --------
-        # This fixture should be used whenever a new answer class is
-        # defined
-        >>> @pytest.mark.usefixtures('answer_file')
-        >>> class TestNewFrontend:
-        >>>     def test1(self):
-                    ...
+def ds(request):
+    try:
+        if isinstance(request.param, list):
```
neutrinoceros, line 281: the try block should be as short as possible to improve readability and reduce chances of catching other errors
testing for a list specifically may be problematic because any 2-sized sequence would behave the same. Checking for typing.Sequence instead would allow this to not work also with tuples and other sequences types

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
+            # list
+            ds_fn = request.param[0]
+            opts = request.param[1]
+            cls = None if "cls" not in opts.keys() else opts["cls"]
+            args = None if "args" not in opts.keys() else opts["args"]
+            kwargs = None if "kwargs" not in opts.keys() else opts["kwargs"]
+            dataset = data_dir_load(ds_fn, cls=cls, args=args, kwargs=kwargs)
+        else:
+            dataset = data_dir_load(request.param)
+        return dataset
```
neutrinoceros, line 293: note that storing a variable just for the sake of naming it before a return isn't compliant with one of our style guide rules (see #3062)

conftest.py
```diff
@@ -32,39 +27,56 @@ def pytest_addoption(parser):
...
+        action="store_true",
+    )
+    parser.addoption(
+        "--no-hash",
         action="store_true",
     )
+    parser.addoption("--local-dir", default=None, help="Where answers are saved.")
+    # Tell pytest about the local-dir option in the ini files. This
+    # option is used for creating the answer directory on CI
+    parser.addini("local-dir", default="../answer-store", help="answer directory.")
```
neutrinoceros, line 58: careful not to break windows paths here

conftest.py
```diff
@@ -111,156 +276,87 @@ def temp_dir():
...
-        request.cls.hashes = {}
+    return request.param
+
+
+@pytest.fixture(scope="class")
+def big_data(request):
+    """
+    There isn't a way to access command-line options given to pytest from
+    an actual test, so we need a fixture that retrieves and returns its
+    value. In this case, the --answer-big-data option.
```
neutrinoceros, line 357: This doctoring feels much more like a comment to me. Also I'm torn that this is a fixture instead of a decorator but I realise why a decorator may be harder to maintain here so I'm not sure what to suggest.

pytest_answer.ini
```diff
@@ -0,0 +1,10 @@
+[pytest]
+local-dir=/home/travis/build/yt-project/yt/answer-store
+addopts =
+    -s
+    -v
+    --with-answer-testing
+    -rsfE
```
neutrinoceros, line 7: isn't the `-s` part redundant ?
I'd suggest commenting each flag and/or having a link to the official docs (I'd prefer comments as they don't require to be online)

pytest_answer.ini
```diff
@@ -0,0 +1,10 @@
+[pytest]
+local-dir=/home/travis/build/yt-project/yt/answer-store
```
neutrinoceros, line 2: this obviously needs to be updated (provided the file is even relevant still)

tests/pytest_runner.py
```diff
@@ -0,0 +1,17 @@
...
+from yt.config import ytcfg
+
+if __name__ == "__main__":
+    os.environ["OMP_NUM_THREADS"] = "1"
+    pytest_args = [
+        f"--local-dir={os.path.join(ytcfg.get('yt', 'test_data_dir'), 'answers')}",
+        "-c=pytest_answer.ini",
+        "--junitxml=unittests.xml",
+        "--answer-big-data",
+        f"-n {int(os.environ.get('NUM_WORKERS', 6))}",
```
neutrinoceros, line 14: why 6 workers ?
How does pytest react when we ask for more workers than the machine allows ?

Xarthisius: Cause it's a nice even number that's greater than 1 and less than infinity ;-) if you want to dive even deeper in the root cause of that: I was trained to use 1.5 * n, where n is number of cores available and my laptop happens to have 4 of them. I don't want to make this decision less significant, but this file will only ever be used on Jenkins, which happens to set NUM_WORKERS variable to 12. 
As for the 2nd question I think you're confusing number of processes with number of available processors. I seriously doubt anyone will ever run `pytest -n $(cat /proc/sys/kernel/pid_max)` or `pytest -n $(ulimit -u)`. However, I must shamefully admit I have no idea what will happen if you try to run pytest with 32769 processes. I'd hazard a guess that it would anger the kernel gods and they'd immediately punish the puny user by killing all their processes.

conftest.py
Xarthisius: I think we could just drop big-data altogether. Last time it was actually used, i.e. we run tests without it, was before Jenkins.

pytest_unit.ini
Xarthisius: What was I was trying to say there is that those tests are not ignored, they *don't exist* yet. We're gonna follow this PR with other adding answer test on frontend by frontend basis. When times come for Ramses or owls_subfind, those entries will be removed.

jcoughlin11:
@Xarthisius when I try to push to this PR I get an error about not having permission

Xarthisius:
> @Xarthisius when I try to push to this PR I get an error about not having permission

I added you directly to my repo. You should be able to push there instead of the PR branch on yt-project/yt

tests/pytest_runner.py
neutrinoceros: Excellent answer, thanks ! (indeed I must have been confusing processes and processors, I have shallow knowledge of parallelism, sometimes it shows :))

tests/test_requirements.txt
```diff
@@ -6,10 +6,10 @@ glueviz~=0.13.3
 h5py~=3.1.0
 ipython~=7.6.1
 matplotlib~=3.3
+mock
```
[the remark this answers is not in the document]
jcoughlin11: @neutrinoceros Just want to make sure I got the change right so that this conversation can be closed.

conftest.py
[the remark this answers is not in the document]
jcoughlin11: @neutrinoceros Just want to make sure I got the change right so that this conversation can be closed.

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: @neutrinoceros Just want to make sure I got the change right so that this conversation can be closed.

conftest.py
jcoughlin11: I changed all of the offending fixture names. I'll have to change them in the tests as they are added in future PRs.

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: I got rid of the method and added the `==` directly to the hashing fixture

tests/tests_pytest.yaml
jcoughlin11: I removed the file and now have the test names generated from the test class. The only thing I didn't do was get the test version number (e.g., frontend_000.yaml), so right now the answer files are TestClassName.yaml and TestClassName.h5.

conftest.py
```diff
@@ -243,7 +226,7 @@ def hashing(request):
         _save_result(hashes, request.cls.answer_file)
     # Compare hashes
     elif not no_hash and not store_hash:
-        _compare_result(hashes, request.cls.saved_hashes)
+        assert hashes == request.cls.saved_hashes
```
neutrinoceros, line 230: come to think of it I'm not sure it's good practice to have assert statements in a fixture. Is there any resource available on the topic ?

tests/tests_pytest.yaml
Xarthisius: We still need some sort of tag for new answer generation, but you could use e.g. attribute on a class or something like that.

yt/frontends/gdf/tests/test_outputs.py
jcoughlin11: IIRC, I did that because the hashing fixture doesn't actually return anything.  Additionally, it needs to do some work both before and after the test is run, so `yield` is employed. If there's a better way to do this, I'm open to it!

conftest.py
jcoughlin11: @neutrinoceros Just want to make sure I got the change right so that this conversation can be closed.

conftest.py
jcoughlin11: I reworked the fixture to make the code cleaner. I think I addressed both of your points. Let me know if I didn't.

conftest.py
jcoughlin11: @neutrinoceros Just want to make sure I got the change right so that this conversation can be closed.

conftest.py
jcoughlin11: I removed the fixture and the `big-data` option

pytest_answer.ini
jcoughlin11: @neutrinoceros What's the new path supposed to be?

conftest.py
jcoughlin11: Not sure. I'll have a look.

conftest.py
neutrinoceros: Actually I'm not sure about your new solution. Let me close this thread and start a new one with the current implementation.

conftest.py
```diff
@@ -55,17 +51,15 @@ def pytest_addoption(parser):
     parser.addoption("--local-dir", default=None, help="Where answers are saved.")
     # Tell pytest about the local-dir option in the ini files. This
     # option is used for creating the answer directory on CI
-    parser.addini("local-dir", default="../answer-store", help="answer directory.")
+    parser.addini("local-dir",
+        default=os.path.join("..", "answer-store"),
```
neutrinoceros, line 56: So this was rewritten following one of my comments expressing concerns regarding the Windows compatibility of this bit.
I'm now uncertain whether this implementation is indeed cross platform. There's also the issue that this will break on any platform if `pytest` isn't called from the top level of the repo.

The one way I know works everywhere to refer to parent directories would read
```python
default=str(Path(__file__).parent / "answer-store"),
```

tests/test_requirements.txt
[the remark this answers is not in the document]
neutrinoceros: something went wrong, apparently mock was removed _and added back_.

yt/frontends/gdf/tests/test_outputs.py
neutrinoceros: I think there is, let me get back at you.

pytest_answer.ini
neutrinoceros: I don't know, but we stopped using Travis a couple mouths ago so it's clearly not usable anymore as it stands.

tests/test_requirements.txt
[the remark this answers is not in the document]
jcoughlin11: Yeah, I added it back in since I didn't see it in the file, but I thought it needed to be there based on your comment. Did I mess up and misunderstand?

tests/tests_pytest.yaml
jcoughlin11: Sounds good. Should that go in this PR or a future one?

pytest_answer.ini
jcoughlin11: @Xarthisius Do you know the updated path?

tests/tests_pytest.yaml
munkm: Probably this one since it's an infrastructure PR and later ones I think are the actual tests.

conftest.py
jcoughlin11: I think if pytest isn't called from the top-level directory then it won't find the proper `conftest.py` file, correct? Doesn't that mean that pytest must be invoked from the root yt directory?

jcoughlin11:
@matthewturk @Xarthisius @neutrinoceros @munkm @cphyc Anyone have  any thoughts on the failing profile plot tests?

pytest_answer.ini
Xarthisius: it's not used right now. We'll change the path in a separate PR that will enable pytest with GHA. Leave it for now.

conftest.py
neutrinoceros: I'm 90% sure that pytest walks the file tree until it finds a configuration file it recognizes, so unless _we_ break this mechanism, we're supposed to be able to run pytest from anywhere if it's config file lives at the top level. This is especially useful for yt where tests files are not centralised into one location, so please don't break it :)

Note that since 6.0, pyproject.toml is supported (https://docs.pytest.org/en/stable/customize.html?highlight=configure#pyproject-toml) so we actually don't need a new file.

Xarthisius:
Re: `pytest version of gdf tests had been accidentally included. Now removed.` @jcoughlin11 they were included **on purpose**. See the description of this PR...

conftest.py
jcoughlin11: This is currently only used by arepo when checking the total number of particles in the sph test, so it doesn't deal with the different particle types at all because it currently doesn't need to.

tests/tests_pytest.yaml
jcoughlin11: Ok. Added.

pytest_answer.ini
neutrinoceros: fine, but can we at least change the value to a meaning placeholder like "UNSET" ?

conftest.py
jcoughlin11: So I didn't manage to find anything either way. However, none of the examples I saw in the pytest documentation do it, so I'm guessing it's just implicitly frowned upon. However, I don't see why we can't have the `hashing` fixture call `_compare_result` and, if an `AssertionError` is raised then have `hashing` call `pytest.fail()` in order to mark the test as failing, not let the `AssertionError` pass silently, but also not stop the pytest run. Thoughts?

conftest.py
jcoughlin11: I added the ini options to the `pyproject.toml` file and removed the two pytest ini files. The only option that isn't in the toml file is `--with-answer-testing`. That should be added manually on the command line when invoking pytest because that's how the answer tests are turned on and the unit and answer tests are mutually exclusive.

yt/frontends/gdf/tests/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+# Test parameters. Format:
+# {test1: {param1 : [(val1, val2,...), (id1, id2,...)], param2 : ...}, test2: ...}
+test_params = {
+    "test_sedov_tunnel": {
+        "a": [(0, 1, 2), ("0", "1", "2")],
+        "d": [(None, ("sphere", ("max", (0.1, "unitary")))), ("None", "sphere")],
+        "w": [(None, "density"), ("None", "density")],
+        "f": [("density", "velocity_x"), ("density", "vx")],
```
munkm, line 18: could the parameter names be clearer here?

yt/frontends/gdf/tests/test_outputs.py
```diff
@@ -1,30 +1,28 @@
...
+    saved_hashes = None
+    answer_version = "000"
 
-@requires_file(sedov)
-def test_GDFDataset():
-    assert isinstance(data_dir_load(sedov), GDFDataset)
+    @pytest.mark.usefixtures("hashing")
+    @pytest.mark.parametrize("ds", [sedov], indirect=True)
+    def test_sedov_tunnel(self, a, d, w, f, ds):
+        self.hashes.update(small_patch_amr(ds, f, w, a, d))
```
munkm, line 20: With this test could the variables passed to the test be named more clearly?

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,560 @@
...
+        if isinstance(data, dict):
+            hd = _hash_dict(data)
+        elif data is None:
+            hd = hashlib.md5(bytes(str(-1).encode("utf-8"))).hexdigest()
+        else:
+            raise
+    return hd
+
+
+def _save_result(data, outputFile):
```
munkm, line 207:
```suggestion
def _save_result(data, output_file):
```

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,560 @@
...
+    Parameters
+    ----------
+    data : dict
+        Test results to be saved.
+
+    outputFile : str
+        Name of file to save results to.
+    """
+    with open(outputFile, "a") as f:
+        yaml.dump(data, f)
```
munkm, line 220:
```suggestion
    output_file : str
        Name of file to save results to.
    """
    with open(output_file, "a") as f:
        yaml.dump(data, f)
```

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,560 @@
...
+                # no proxy, so we have to make one ourselves. Using -1
+                if test_data is None:
+                    test_data = -1
+                grp.create_dataset(test_name, data=test_data)
+
+
+def _parse_raw_answer_dict(d, h5grp):
+    """
+    Doc string.
+    """
```
munkm, line 266: Can you rewrite this docstring or remove it?

conftest.py
```diff
@@ -111,156 +261,72 @@ def temp_dir():
...
 
-    Each class that performs answer tests (e.g., TestEnzo) has a
-    corresponding answer file. These answer files are stored in
-    test_data_dir/answers. The names of the files and their
-    corresponding answer class are in ./tests/tests.yaml.
+@pytest.fixture(scope="class")
+def field(request):
+    """
+    Fixture for returning the field. Needed because indirect=True is
+    used for loading the datasets.
```
munkm, line 285: Why would indirect=True make it so fields aren't accessible?

munkm (commented):
I have a few comments. I'll review again once they're addressed! 🙂

tests/test_minimal_requirements.txt
```diff
@@ -3,9 +3,10 @@ matplotlib~=1.5.3
...
 nose~=1.3.7
 nose-timer~=0.7.3
-pytest~=5.2
 pyyaml>=4.2b1
 coverage~=4.5.1
 codecov~=2.0.15
 unyt~=2.8.0
 more_itertools==8.4
+pytest~=6.1
+mock
```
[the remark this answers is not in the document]
neutrinoceros: @jcoughlin11 it seems like this keep coming back. Please be careful what you check in.

pytest_answer.ini
neutrinoceros: this file is now gone right ? In which case, you can resolve this @jcoughlin11

pytest_answer.ini
neutrinoceros: same as above:
this file is now gone right ? In which case, you can resolve this @jcoughlin11

conftest.py
neutrinoceros: Excellent. If this is already documented you can resolve this thread. I'll review more thoroughly tommorow

yt/frontends/gdf/tests/conftest.py
jcoughlin11: Sorry, yup! I changed the fixture names but forgot about these. I'll do that now.

conftest.py
jcoughlin11: I'm not sure, to be honest. All I know is that things did not work correctly unless the dataset parameterizations used `indirect=True` and as soon as I did that, the other fixtures stopped working until I made them use `indirect=True`, as well.

yt/frontends/gdf/tests/test_outputs.py
jcoughlin11: Fixed.

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: Fixed.

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: Fixed.

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: I removed it.

yt/frontends/gdf/tests/test_outputs.py
jcoughlin11: @neutrinoceros Find anything? If not, no worries!

jcoughlin11:
@matthewturk @munkm @Xarthisius @neutrinoceros @cphyc I think I've addressed everything. If I missed something, let me know! Thank you for all of the help!

yt/frontends/gdf/tests/test_outputs.py
neutrinoceros: Sorry for not coming back. I researched this a little and it's still quite confusing to me because the docs only mention this marker as useful to add fixtures to test classes.
(https://docs.pytest.org/en/stable/fixture.html#usefixtures)

This SO thread contains an answer that seems inline with your reasoning
https://stackoverflow.com/questions/50593556/pytest-how-to-use-a-mark-to-inject-a-fixture

TBH the fixture system in pytest is still half black magic to me, so I'm not too worried I don't really get this. Moving on !

conftest.py
neutrinoceros: So I tried running this
```python
import pytest

@pytest.fixture()
def orwell():
    assert 2 + 2 == 5

def test_dystopia(orwell):
    assert 1 + 1 == 2
```

and learned that this doesn't "FAIL", but `ERROR`s out. I don't think it's reasonable to alter pytest's indications; if the authors wanted to make a distinction between failure and error, let's not mingle with it.
Considering this, I'd advise against raising `pytest.fail`, but I can actually get along with the (current ?) implementation I first commented. I see that it's now outdated so I'm not even sure this is relevant anymore, sorry :/

matthewturk (approved):
I approve of this.  I think we still need sign-off from others.

Xarthisius:
I cannot physically push the approve button (github doesn't let me do that since I "authored" this PR) but I approve :+1:

conftest.py
neutrinoceros: @jcoughlin11 ... is it ?

jzuhone (approved):

conftest.py
munkm: @jcoughlin11 is this still relevant to this PR?

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,557 @@
...
+
+    Parameters
+    ----------
+    data : dict
+        Test results to be saved.
+
+    output_file : str
+        Name of file to save results to.
+    """
+    with open(output_file, "a") as f:
```
cphyc, line 219: Should this be `"w"` instead of `"a"`?

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,557 @@
...
+    r"""
+    Convenience function used in plot_window_attribute_test.
+
+    Parameters
+    ----------
+    ds : Dataset
+        The Dataset object from which the plotting data is taken.
+
+    ptype : string
+        Type of plot to make (e.g., SlicePlot).
```
cphyc, line 468: Everywhere else in the code, `ptype` is particle type. Shouldn't we rename that to `plot_type` as in `_create_phase_plot_attribute_plot`?

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,557 @@
...
+
+    pkwargs : dict
+        Any keywords to be passed when creating the plot.
+    """
+    if ptype is None:
+        raise RuntimeError("Must explicitly request a plot type")
+    cls = getattr(pw, ptype, None)
+    if cls is None:
+        cls = getattr(particle_plots, ptype)
+    plot = cls(*(ds, axis, field), **pkwargs)
```
cphyc, line 484:
```suggestion
    plot = cls(ds, axis, field, **pkwargs)
```

yt/utilities/answer_testing/testing_utilities.py
```diff
@@ -0,0 +1,557 @@
...
+        raise RuntimeError("Must explicitly request a plot type")
+    cls = getattr(profile_plotter, plot_type, None)
+    if cls is None:
+        cls = getattr(particle_plots, plot_type)
+    plot = cls(*(data_source, x_field, y_field, z_field), **plot_kwargs)
+    return plot
+
+
+def get_parameterization(fname):
+    try:
```
cphyc, line 524: Could you add a docstring to this function? If this is only used in this file, it should be probably prepended with an underscore.

cphyc (approved):
Overall this looks good to me, I just left some minor comments.

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: The answers are written method-by-method instead of all at once, so if "w" was used then the file would constantly be overwriting itself.

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: Good catch!

yt/utilities/answer_testing/testing_utilities.py
jcoughlin11: I added one. The function is used in, for example, the enzo tests, to help parametrize over every field in a dataset.

yt/frontends/gdf/tests/test_outputs.py
munkm: @neutrinoceros is this resolved?

yt/utilities/answer_testing/testing_utilities.py
neutrinoceros: what do you mean "method by method" ? wouldn't it be more reproducible to update an in-memory dict and dump it in one go at the end ?

jcoughlin11: I mean that each individual test is called, then run, then has its results written to disk instead of updating an in-memory dictionary and dumping all the results at once. I didn't do it that way because it was harder to implement, if I recall correctly, but I'm open to suggestions.

munkm:
@jcoughlin11 the sedov tunnel and GDF tests are failing on this PR right now.

Xarthisius:
> @jcoughlin11 the sedov tunnel and GDF tests are failing on this PR right now.

That was my fault. It should be fixed now.

munkm:
Ok! thanks for fixing @Xarthisius !

yt/frontends/gdf/tests/test_outputs.py
neutrinoceros: let's say it is. If I ever get a better grasp of this and decide this is bad practice, I'll propose a patch then.

munkm: That sounds like a good plan!

neutrinoceros:
🎉

munkm:
🎉 🎈 🎊 🎉  Good job everybody!!!

conftest.py
jcoughlin11: @neutrinoceros Sorry, I missed your response! I looked, and it is still relevant, since I currently do an `assert` in the fixture and, if there's an error, raises `pytest.fail()`. I also just saw that this was merged, so I can change it in the next pytest PR, if needed.

PR #3105, opened 2021-02-27
Make config related paths dynamical. Fixes #3104

### LINKED ISSUE #3104: test_config is not parallel-safe
### Bug report

**Bug summary**

During the execution of `yt/utilities/tests/test_config.py` we modify the actual `<XDG_CONFIG_HOME>/yt/yt.toml`. This is problematic since tests run in parallel as separate processes and can pick up whatever garbage is being dumped into temporary yt configuration. This can manifest as e.g.:

```
THIS IS THE REAL CAUSE OF THE FAILURE:
  File "doc/source/cookbook/custom_transfer_function_volume_rendering.py", line 6, in <module>
  File "/home/fido/workspace/yt_py38_latest_deps/yt/loaders.py", line 86, in load
    raise FileNotFoundError(msg)
FileNotFoundError: No such file or directory: 'Enzo_64/DD0043/data0043'.
```

Entire thing went unnoticed cause until yesterday test suite didn't rely on config file. Instead I was doing the following:

```
 sed -i -e '/test_data_dir/ s:/does/not/exist:/mnt/yt:g' \
    -e '/xray_data_dir/ s:/does/not/exist:/mnt/yt/xray_data:g' \
    -e '/supp_data_dir/ s:/does/not/exist:/mnt/yt/supp_data:g' \
    -e '/__strict_requires/ s/False/True/g' \
    -e '/suppressstreamlogging/ s/False/True/g' yt/config.py
```

prior to test execution to set whatever is necessary, which is now restored back.

**Code for reproduction**

```bash
$ sudo chown root:root -R $HOME/.config/yt 
$ nosetests -s -vv yt/utilities/tests/test_config.py
```

**Actual outcome**

```
ERROR: test suite for <module 'yt.utilities.tests.test_config' from '/home/xarth/codes/xarthisius/yt/yt/utilities/tests/test_config.py'>
  File "/home/xarth/codes/xarthisius/yt/pytest/lib/python3.8/site-packages/nose/suite.py", line 210, in run
    self.setUp()
  File "/home/xarth/codes/xarthisius/yt/pytest/lib/python3.8/site-packages/nose/suite.py", line 293, in setUp
    self.setupContext(ancestor)
  File "/home/xarth/codes/xarthisius/yt/pytest/lib/python3.8/site-packages/nose/suite.py", line 316, in setupContext
    try_run(context, names)
  File "/home/xarth/codes/xarthisius/yt/pytest/lib/python3.8/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/home/xarth/codes/xarthisius/yt/yt/utilities/tests/test_config.py", line 54, in setUpModule
    os.rename(cfgfile, cfgfile + ".bak_test")
PermissionError: [Errno 13] Permission denied: '/home/xarth/.config/yt/yt.toml' -> '/home/xarth/.config/yt/yt.toml.bak_test'
```

**Expected outcome**

Config not modified.

Xarthisius:
## PR Summary

This PR ensures that every time a "config path" is requested it respects the current value of XDG_HOME_DIR.

### DIFF
--- a/yt/config.py
+++ b/yt/config.py
@@ -66,14 +66,31 @@ ytcfg_defaults["yt"] = dict(
 )
 
 
-CONFIG_DIR = os.environ.get(
-    "XDG_CONFIG_HOME", os.path.join(os.path.expanduser("~"), ".config", "yt")
-)
-if not os.path.exists(CONFIG_DIR):
-    try:
-        os.makedirs(CONFIG_DIR)
-    except OSError:
-        warnings.warn("unable to create yt config directory")
+def config_dir():
+    config_root = os.environ.get(
+        "XDG_CONFIG_HOME", os.path.join(os.path.expanduser("~"), ".config")
+    )
+    conf_dir = os.path.join(config_root, "yt")
+
+    if not os.path.exists(conf_dir):
+        try:
+            os.makedirs(conf_dir)
+        except OSError:
+            warnings.warn("unable to create yt config directory")
+    return conf_dir
+
+
+def old_config_file():
+    return os.path.join(config_dir(), "ytrc")
+
+
+def old_config_dir():
+    return os.path.join(os.path.expanduser("~"), ".yt")
+
+
+# For backward compatibility, do not use these vars internally in yt
+CONFIG_DIR = config_dir()
+_OLD_CONFIG_FILE = old_config_file()
 
 
 class YTConfig:
@@ -167,21 +184,20 @@ class YTConfig:
 
     @staticmethod
     def get_global_config_file():
-        return os.path.join(CONFIG_DIR, "yt.toml")
+        return os.path.join(config_dir(), "yt.toml")
 
     @staticmethod
     def get_local_config_file():
         return os.path.join(os.path.abspath(os.curdir), "yt.toml")
 
 
-OLD_CONFIG_FILE = os.path.join(CONFIG_DIR, "ytrc")
 _global_config_file = YTConfig.get_global_config_file()
 _local_config_file = YTConfig.get_local_config_file()
 
-if os.path.exists(OLD_CONFIG_FILE):
+if os.path.exists(old_config_file()):
     if os.path.exists(_global_config_file):
         issue_deprecation_warning(
-            f"The configuration file {OLD_CONFIG_FILE} is deprecated in "
+            f"The configuration file {old_config_file()} is deprecated in "
             f"favor of {_global_config_file}. Currently, both are present. "
             "Please manually remove the deprecated one to silence "
             "this warning.",
@@ -190,7 +206,7 @@ if os.path.exists(OLD_CONFIG_FILE):
         )
     else:
         issue_deprecation_warning(
-            f"The configuration file {OLD_CONFIG_FILE} is deprecated. "
+            f"The configuration file {_OLD_CONFIG_FILE} is deprecated. "
             f"Please migrate your config to {_global_config_file} by running: "
             "'yt config migrate'",
             since="4.0.0",
--- a/yt/fields/tests/test_fields_plugins.py
+++ b/yt/fields/tests/test_fields_plugins.py
@@ -1,9 +1,20 @@
 import os
+import shutil
+import tempfile
 import unittest
 
 import yt
-from yt.config import CONFIG_DIR, ytcfg
+from yt.config import YTConfig, config_dir, ytcfg
 from yt.testing import assert_raises, fake_random_ds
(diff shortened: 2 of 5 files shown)


yt/config.py
```diff
@@ -67,14 +67,21 @@
...
-)
-if not os.path.exists(CONFIG_DIR):
-    try:
-        os.makedirs(CONFIG_DIR)
-    except OSError:
-        warnings.warn("unable to create yt config directory")
+def config_dir():
+    conf_dir = os.environ.get(
+        "XDG_CONFIG_HOME", os.path.join(os.path.expanduser("~"), ".config", "yt")
+    )
```
neutrinoceros, line 73: I have to run I can't review fully now but I think there's a mistake here
```suggestion
    XDG = os.environ.get("XDG_CONFIG_HOME",  os.path.join(os.path.expanduser("~"), ".config"))
    conf_dir = os.path.join(os.path.expanduser(XDG, "yt")
```

matthewturk:
Not sure if it helps, but pooch (an optional dependency for yt) depends on appdirs which abstracts this a bit.

yt/config.py
```diff
@@ -67,14 +67,22 @@
...
-        os.makedirs(CONFIG_DIR)
-    except OSError:
-        warnings.warn("unable to create yt config directory")
+def config_dir():
+    config_root = os.environ.get(
+        "XDG_CONFIG_HOME", os.path.join(os.path.expanduser("~"), ".config")
+    )
+    conf_dir = os.path.join(config_root, "yt")
+
+    if not os.path.exists(conf_dir):
```
neutrinoceros, line 76: I'm really not a fan of this feature (silently create the config file). I like it even less as a function, since it allows for repetitive warnings in case anything goes wrong at creation.
If you think there's a good reason (other than historic) to do this, can you please document it in #3045 ?

yt/fields/tests/test_fields_plugins.py
```diff
@@ -1,10 +1,22 @@
...
+import shutil
+import tempfile
 import unittest
 
 import yt
-from yt.config import CONFIG_DIR, ytcfg
+from yt.config import YTConfig, config_dir, ytcfg
 from yt.testing import assert_raises, fake_random_ds
 
+XDG_CONFIG_HOME = os.environ.get("XDG_CONFIG_HOME")
```
neutrinoceros, line 10: why isn't this part of the test setup ?

yt/funcs.py
```diff
@@ -959,7 +959,7 @@ def enable_plugins(plugin_filename=None):
         # - obsolete config dir.
         my_plugin_name = ytcfg.get("yt", "plugin_filename")
         old_config_dir = os.path.join(os.path.expanduser("~"), ".yt")
```
neutrinoceros, line 961: while we're at it, shouldn't this be changed to a getter function too ?

neutrinoceros (commented):
Thank you for doing this; I left a couple notes but overall LGTM

yt/funcs.py
Xarthisius: getter for what? `~/.yt` was a hardcoded path we used to have.

neutrinoceros: I meant a similar function as config_file() and old_config_dir() that you are introducing in this PR. It's not required, it's just a matter of consistency :)

matthewturk (approved):

yt/config.py
Xarthisius: It's not creating config file, but a config directory, which is pretty much a default behavior for a lot of software:

* https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/__init__.py#L419-L422

Also noting the most important thing: it was already there and changing config related behavior is not the point of this PR.

yt/fields/tests/test_fields_plugins.py
Xarthisius: What's the difference? It's not going to be imported by anything

yt/config.py
neutrinoceros: alright I'll open a different PR to fix this

yt/fields/tests/test_fields_plugins.py
neutrinoceros: You're right, but in this states it looks like it's going to be used in more than one test (which would make it a candidate for fixture in pytest), but it's not the case, so might as well define it in the only scope it used.

munkm (approved):

neutrinoceros (approved):

PR #3106, opened 2021-03-03
ENH: (optional) rich-based logging (YTEP-0039)

neutrinoceros:
## PR Summary

This is the first PR in support to [YTEP-0039](https://github.com/yt-project/ytep/pull/18)
This is also the first PR that's _enabled_ by @cphyc's work with the new config file (#2981)
A side effect of modernising the log-related configuration is that this PR actually fixes some bugs here and there.
In particular, this improve the setter `yt.utilities.logger.set_log_level` to enforce a consistent state between the shared configuration object `ytcfg` and the logger's current log level. Using this setter everywhere instead of `mylog.setLevel` is then more robust.

I've also found some bugs in the command line configuration tool `yt config`, so I'm fixing them as well (tests are only at the draft stage).

## PR Checklist

- [x] complete implementation of the YTEP specification for a yt.logging config file section
- [x] Add default values for this section
- [x] add documentation

Proposed release cycle:
- targeting yt 4.0: offer rich logging as an experimental option (opt-in) and don't change anything with the legacy log style.
- next minor/major release after 4.0: consider the feature stable, keep "legacy" as the default handler but raise visible deprecation warnings (offer an automated migration script at this point).
- next one after that: make rich logging the default.

Hopefully the feature will be well received and tested in its experimental phase. If not, we can keep it experimental for longer or even not make it the default ever.

Notes for documentation and reviewers 
- `use_color=False` won't play nicely with `custom_theme = "my/awesome/rich/theme.ini"` because deactivating colors requires using a theme file already and there's no trivial way to patch them. It may not look like it's worth exposing the `custom_theme` argument but I don't want to undermine colorblind folks out there so I want all users to have the option of setting colors themeselves. This is the easiest and most permissive way to do so.
- I'm keeping changes to the config file migration script (`yt config migrate`) minimal here. It's already complicated enough and I don't think this change is worth taking the risk to break it. It may however become necessary to provide a separate migration script when/if the rich logger is promoted to default in the future.
- converting the legacy logging related parameters in place proved necessary because they were called in many different parts of the code. Doing this means that in case legacy parameters and their replacement collide, the legacy ones win.
- This PR updates the documentation for the CLI configuration tool but it doesn't actually make it work as documented (note that it's actually broken even with the main branch's examples). To keep the present PR focused, I opened a separate followup branch to address this issue #3112
- in a effort to minimise the difficulty of lifting any additional technical debt generated by this change without using deprecation warnings, code bits due for eventual removal receive a comment starting with `# YTEP-0039`
- I don't want to add any new test for now. It will be much easier to write tests for this when we have a pytest-based CI, at least partially, so hopefully soon. Given how hard it was to not break existing tests I figure the basics are already covered anyway.

### DIFF
--- a/yt/config.py
+++ b/yt/config.py
@@ -227,7 +227,10 @@ class YTConfig:
         if "colored_logs" in self["yt"]:
             self["logging", "use_color"] = self["yt", "colored_logs"]
 
-        if "stdout_stream_logging" in self["yt"] and self["yt", "std_stream_logging"]:
+        if (
+            "stdout_stream_logging" in self["yt"]
+            and self["yt", "stdout_stream_logging"]
+        ):
             self["logging", "stream"] = "stdout"
 
         if (
--- a/yt/config.py
+++ b/yt/config.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import warnings
 
@@ -7,16 +8,29 @@ from more_itertools import always_iterable
 from yt._maintenance.deprecation import issue_deprecation_warning
 from yt.utilities.configuration_tree import ConfigLeaf, ConfigNode
 
+# this is a non standard logging level name
+ALL = 1
+
+
+def loglevel_int2str(level: int) -> str:
+    if level == ALL:
+        return "ALL"
+    return logging._levelToName[level]
+
+
+def loglevel_str2int(level: str) -> int:
+    level = level.upper()
+    if level == "ALL":
+        return ALL
+    return logging._nameToLevel[level]
+
+
 ytcfg_defaults = {}
 
 ytcfg_defaults["yt"] = dict(
     serialize=False,
     only_deserialize=False,
     time_functions=False,
-    colored_logs=False,
-    suppress_stream_logging=False,
-    stdout_stream_logging=False,
-    log_level=20,
     inline=False,
     num_threads=-1,
     store_parameter_files=False,
@@ -60,6 +74,17 @@ ytcfg_defaults["yt"] = dict(
         topcomm_parallel_size=1,
         command_line=False,
     ),
+)
+
+ytcfg_defaults["logging"] = dict(
+    level="INFO",
+    use_color=False,
+    stream="stderr",
+    handler="legacy",
+    format=r"%(message)s",
+    width=-1,  # <0 means "auto"
+    date_format="[%m/%d/%Y %H:%M:%S]",
+    custom_theme="",
 )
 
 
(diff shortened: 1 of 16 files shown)


neutrinoceros:
@yt-fido test this please

neutrinoceros:
@cphyc I know I said I wanted to make the `[plot(ting)]` and `[log(ging)]` sections follow a common naming convention but as I was going through the changes (`logging` -> `log`) I realised I was not making anything clearer but instead, I was making stuff harder to grep for (logging VS log function). So in the end I think I still prefer using the long form here (logging) even if it means breaking a uniformity.

neutrinoceros:
This PR has rotten quite a bit and I was not careful enough in rebasing it... The currently failure is however very easy to reproduce locally (phew):

```python
from yt.config import ytcfg
```
raises
```
AttributeError: 'ConfigLeaf' object has no attribute 'get_from_list'
```

yt/config.py
```diff
@@ -65,6 +79,19 @@
...
+        level="INFO",
+        use_color=False,
+        stream="stderr",
+        handler="legacy",
+        format=r"%(message)s",
+        width=-1,  # <0 means "auto"
+        date_format="[%m/%d/%Y %H:%M:%S]",
+        custom_theme="",
+    ),
+)
```
cphyc, line 93: This should be a dict, not a tuple!

```suggestion
ytcfg_defaults["logging"] = dict(
    level="INFO",
    use_color=False,
    stream="stderr",
    handler="legacy",
    format=r"%(message)s",
    width=-1,  # <0 means "auto"
    date_format="[%m/%d/%Y %H:%M:%S]",
    custom_theme="",
)
```

yt/config.py
```diff
@@ -190,6 +217,31 @@ def get_global_config_file():
     def get_local_config_file():
         return os.path.join(os.path.abspath(os.curdir), "yt.toml")
 
+    def _convert_legacy_logging_parameters(self):
+        # YTEP-0039: maintain backward compatibility for a period of time
+        # after which this function should be removed.
+        try:
+            self["logging", "use_color"] = self.get("yt", "colored_logs")
```
cphyc, line 224: For consistency I'd favour
```suggestion
            self["logging", "use_color"] = self["yt", "colored_logs"]
```

yt/utilities/exceptions.py
```diff
@@ -864,6 +867,19 @@ def __str__(self):
         return msg
 
 
+class YTConfigError(YTException):
+    def __init__(self, key: str, choices: Optional[Any] = None) -> None:
```
cphyc, line 871: Isn't `Optional[Any]` equivalent to `Any`?

yt/utilities/logger.py
```diff
@@ -48,24 +94,18 @@ def set_log_level(level):
...
-ufstring = "%(name)-3s: [%(levelname)-9s] %(asctime)s %(message)s"
-cfstring = "%(name)-3s: [%(levelname)-18s] %(asctime)s %(message)s"
-
-if ytcfg.get("yt", "stdout_stream_logging"):
-    stream = sys.stdout
-else:
-    stream = sys.stderr
-
-ytLogger = logging.getLogger("yt")
+    ytLogger.debug("log level to %s", level)
```
cphyc, line 108:
```suggestion
    ytLogger.debug("log level set to %s", level)
```

cphyc (commented):
Haven't reviewed everything but I may have found the origin of your issue

yt/config.py
neutrinoceros: Thank you so much !

yt/utilities/exceptions.py
neutrinoceros: Oh my. Of course it is ! Well played

yt/config.py
neutrinoceros: I wonder: did you keep the `get` method just for backwards compatibility ? In which case I'm going to drop it everywhere this is currently used in lines touched by this PR (and soon, the world)

yt/utilities/exceptions.py
neutrinoceros: I think I got a more useful annotation now.

neutrinoceros:
this is failing after a rebase... I hope I can find a solution later.

neutrinoceros:
Withdrawing the YTEP for reasons detailed here: https://github.com/yt-project/ytep/pull/18#issuecomment-998747132

PR #3107, opened 2021-03-03
parallelize index calculations

brittonsmith:
## PR Summary

This parallelizes the most expensive bits of `_initialize_coarse_index` and `_initialize_refined_index`.

I am testing this now on some large datasets, but I thought I'd open the PR for review just in case anyone spots anything obviously wrong.

Is there a way to directly examine the resulting ewah files for comparison/testing?

## PR Checklist


matthewturk:
@langmm can you take a look?

brittonsmith:
Here is a update on testing this with a 2048^3 dataset. Scaling up to 8 cores is pretty good. I did not go higher than that as it's about the limit of our filesystem. There is some variation in runtime due to heavy filesystem use on the cluster, but running with 8 cores was at least 5x faster.

Unfortunately, the size of the resulting ewah file is not constant with number of cores, so there is a problem.
- 1 core: 504MB
- 2 cores: 497MB
- 4 cores: 494MB
- 8 cores: 492MB

My guess is this is in the refined index calculation since this seems to be a cumulative process where the loop starts with one `BoolArrayCollection` object that is updated each iteration. I'll try a few other things, but if anyone has thoughts on what to do here, they are welcome.

matthewturk:
The refined index might end up being tricky.  I reviewed the code a little bit and I *think* that the hard part is in the `append`.  I think each process will have to call `append`, and probably in the same order.  One issue that might come up is that until an array is `append`-ed, it isn't compressed.  I need to think a little bit about this, but it should be tractable to parallelize it without too invasive of a change.

brittonsmith:
Ok, thanks @matthewturk. We can return to this later.

yt/geometry/particle_geometry_handler.py
```diff
@@ -206,8 +208,10 @@ def _initialize_coarse_index(self):
                 else:
                     hsml = None
                 self.regions._coarse_index_data_file(pos, hsml, data_file.file_id)
-            self.regions._set_coarse_index_data_file(data_file.file_id)
         pb.finish()
+        self.regions.masks = self.comm.mpi_allreduce(self.regions.masks, op="sum")
```
langmm, line 212:
```suggestion
        self.regions.masks = self.comm.mpi_allreduce(self.regions.masks, op="sum")
        self.regions.particle_counts = self.comm.mpi_allreduce(self.regions.particle_counts, op="sum")
```

`_coarse_index_data_file` also modifies the `particle_counts` class variable on the regions object which is used during refinement. The suggested additional reduction resulted in consistent file sizes regardless of number of processes for me. I will keep looking for additional speedups, but if this passes, I think it is good to go forward.

brittonsmith: Thanks a ton for looking into this. I'll test it out and report back.

brittonsmith:
After @langmm's suggested change, I do get ewah files that are exactly the same size for different numbers of processors. Here are some timing numbers for different numbers of cores.
1 1241.4727783203125 seconds
2 960.233216047287 seconds
4 436.43122267723083 seconds
8 251.01400780677795 seconds
16 140.09403777122498 seconds

I would say this is working now.

Xarthisius (approved):

langmm (approved):

chummels:
So I tried running this on the IllustrisTNG-50 dataset, which is 2.7 TB and 2048**3 cells.  I'm running this on Stampede2 with the KnightsLanding nodes 96GB/node.  By default, it was allocating ~10GB chunks to each of the cores for the coarse indexing, so I would fail until I dropped down to using 8 cores (12GB/core), and then it was able to get through the coarse indexing OK.  But then it immediately failed on the refined indexing, for a similar out of memory error, being unable to allocate _another_ 10GB more to each of the cores: ```P001 yt : [ERROR    ] 2021-08-08 18:45:26,271 MemoryError: Unable to allocate 9.59 GiB for an array with shape (262144, 39287) and data type uint8```.  I'm wondering if it is supposed to allocate that second batch of memory for the refined indexing (this may very well be the case), or perhaps there is a memory leak after the first set of arrays from the coarse indexing weren't de-allocated?  Either way, I'll boost this to multiple nodes to give it more memory to use in hopes that I can get things working.  

Thanks for implementing this for those of us using big big datasets.

chummels (approved):

chummels:
My error above I guess was just an out of memory error for the normal processing of these datasets.  I think this is good to go.

PR #3109, opened 2021-03-03
new frontend: ASPECT pvtu

chrishavlin:
This adds a front end for the unstructured mesh output from ASPECT (https://aspect.geodynamics.org/, https://github.com/geodynamics/aspect), based off of `exodus_ii` frontend. It introduces an optional dependency on `xmltodict`. Here's a (slightly out of date) notebook showing the frontend working and slicing through a 3D model of lithosphere extension: [link](https://github.com/chrishavlin/yt_scratch/blob/master/notebooks/ASPECT_yt_demo.ipynb). 

**Sample data (still need to update answer store...)**: `curl -JO  http://use.yt/upload/b908f406` will fetch `aspect.tar.gz`, which contains the output used in the tests: `aspect/output_convection_box_3d/nproc_1/` and  `aspect/output_convection_box_3d/nproc_4/` which are single processor and multiprocessor output from the first couple timesteps of ASPECT's [3D convection in a box cookbook](https://github.com/geodynamics/aspect/blob/master/cookbooks/convection_box_3d.prm). 

**Current limitations (for future PRs?):** 
* pvtu/vtu files only (ASPECT also outputs in HDF)
* ~~handles 3D datasets only~~ (2d works now!)
* no particle support (ASPECT outputs separate files for particles if they are used)
* assumes single cell type (ASPECT  doesn't generally mix cell types, but relaxing this assumption could make this a more general `deal.ii` front end rather than just ASPECT)
* loads single timesteps only, would be nice to load a simulation directory (`exodus_ii` has this, shouldn't be too hard to add)
* could parse aspect's parameter file for extra metadata not in the pvtu/vtu files
* yt's unstructured mesh is not fully parallelized yet, so currently have to load the entire connectivity and mesh arrays into memory along with any field being plotted/sampled.


chrishavlin:
Oh and note that this frontend uses a very rudimentary hidden sidecar-JSON framework to stash dataset info to speed up subsequent loads. I know @matthewturk  is interested in a wider sidecar file framework and this could perhaps be better linked with yt's existing toml file use and expanded into it's own module (I added this to the ASPECTDataset class for now...).

chrishavlin:
ready for review I think! Just added a new test, so will fail initially...

chrishavlin:
@yt-fido test this please

chrishavlin:
I decided to move development over to a plugin to speed up iteration with ASPECT users who are testing this out. New repo is at https://github.com/data-exp-lab/yt_aspect/ Maybe down the line I'll work on getting this into yt proper, but I think it fits best as a plugin for now.

PR #3110, opened 2021-03-04
cleanup: remove dead config option

neutrinoceros:
## PR Summary

I think the `log_file` config option isn't used or usable anymore. It looks like the one time we're checking this option in the code base is to force-disable it.

I discovered this while working on #3106


neutrinoceros:
Welp. I don't understand the failure seen on Jenkins. My work hypothesis is that maybe `log_file` is actually used its config file ? any clue @Xarthisius ?

matthewturk:
I cannot fathom where this error is coming from.

matthewturk:
@yt-fido test this please

Issue #2256, opened 2019-05-01
Can't annotate velocity vectors with SPH data in yt4.0


ibutsky:
### Bug report

**Bug summary**

When I make a projection or slice plot of Tipsy data with the yt-4.0 branch, I can't annotate velocity. I've tried both "annotate_quiver", "annotate_velocity", and "annotate_streamlines", with various aliases for the velocity fields. 

**Code for reproduction**

Dataset: "TipsyGalaxy" from the yt data hub. Download link:http://yt-project.org/data/TipsyGalaxy.tar.gz

```python
import yt
ds = yt.load('galaxy.00300')
v, cen = ds.h.find_max(('gas', 'density'))
p = yt.ProjectionPlot(ds, 'x', ('gas', 'temperature'),\
 center = cen, width = (200, 'kpc'), weight_field = ('gas', 'density'))
p.annotate_velocity()
p.save()

```

**Actual outcome**

```
# If applicable, paste the console output here
yt : [INFO     ] 2019-04-30 15:54:22,479 Parameters: current_time              = 20.100000000000044
yt : [INFO     ] 2019-04-30 15:54:22,479 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2019-04-30 15:54:22,479 Parameters: domain_left_edge          = None
yt : [INFO     ] 2019-04-30 15:54:22,479 Parameters: domain_right_edge         = None
yt : [INFO     ] 2019-04-30 15:54:22,479 Parameters: cosmological_simulation   = 0.0
yt : [INFO     ] 2019-04-30 15:54:22,513 Loading KDTree from galaxy.00300.kdtree
yt : [INFO     ] 2019-04-30 15:54:22,536 Allocating for 3.154e+05 particles
yt : [INFO     ] 2019-04-30 15:54:24,262 Max Value is 1.12117e-20 at 0.1224512308835983 -0.0390233956277370 0.3707219660282135
yt : [INFO     ] 2019-04-30 15:54:24,272 xlim = -100.039023 99.960977
yt : [INFO     ] 2019-04-30 15:54:24,272 ylim = -99.629278 100.370722
yt : [INFO     ] 2019-04-30 15:54:24,273 xlim = -100.039023 99.960977
yt : [INFO     ] 2019-04-30 15:54:24,273 ylim = -99.629278 100.370722
yt : [INFO     ] 2019-04-30 15:54:24,282 Making a fixed resolution buffer of (('gas', 'temperature')) 800 by 800
[...]

Issue #1717, opened 2018-03-07
annotate_velocity fails in slices and projections


chummels:
### Bug report

**Bug summary**

In demesh branch, annotating velocity vectors on slices and projections fails.

**Code for reproduction of `slice` error**

```
import yt
ds = yt.load('~/src/yt-data/snapshot_033/snap_033.0.hdf5')
p = yt.SlicePlot(ds, 'x', ('gas', 'density'))
p.annotate_velocity()
p.save('test.png')
```

**Outcome**

Errors out with traceback:

```
yt : [INFO     ] 2018-03-07 12:06:22,786 Parameters: current_time              = 4.343952725460923e+17 s
yt : [INFO     ] 2018-03-07 12:06:22,786 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2018-03-07 12:06:22,786 Parameters: domain_left_edge          = [ 0.  0.  0.]
yt : [INFO     ] 2018-03-07 12:06:22,786 Parameters: domain_right_edge         = [ 25.  25.  25.]
yt : [INFO     ] 2018-03-07 12:06:22,787 Parameters: cosmological_simulation   = 1
yt : [INFO     ] 2018-03-07 12:06:22,787 Parameters: current_redshift          = -4.8118916649e-05
yt : [INFO     ] 2018-03-07 12:06:22,787 Parameters: omega_lambda              = 0.762
yt : [INFO     ] 2018-03-07 12:06:22,787 Parameters: omega_matter              = 0.238
yt : [INFO     ] 2018-03-07 12:06:22,787 Parameters: hubble_constant           = 0.73
yt : [INFO     ] 2018-03-07 12:06:23,083 Allocating for 4.194e+06 particles
Loading particle index: 100%|████████████████████████████████████████████████████████| 12/12 [00:00<00:00, 175.08it/s]
yt : [INFO     ] 2018-03-07 12:06:27,417 xlim = 0.000000 25.000000
yt : [INFO     ] 2018-03-07 12:06:27,417 ylim = 0.000000 25.000000
yt : [INFO     ] 2018-03-07 12:06:27,418 xlim = 0.000000 25.000000
yt : [INFO     ] 2018-03-07 12:06:27,418 ylim = 0.000000 25.000000
yt : [INFO     ] 2018-03-07 12:06:27,424 Making a fixed resolution buffer of (('gas', 'density')) 800 by 800
Traceback (most recent call last):
[...]

PR #3111, opened 2021-03-06
fix for velocity_annotation with unstructured grids

chrishavlin:
## PR Summary

This fixes several related issues that were causing `velocity_annotation` to fail for unstructured grids and improves some of the Stream frontend's unstructured mesh testing. 

The following reproduces the error on `main`:

```python

import yt

# test dataset, similar to yt.testing.fake_hexahedral_ds but with velocity fields
import numpy as np
from yt.frontends.stream.sample_data import hexahedral_mesh
coords = hexahedral_mesh._coordinates
cons =  hexahedral_mesh._connectivity
node_data = {}
dist = np.sum(coords ** 2, 1) # distance from origin
node_data[("connect1", "test")] = dist[cons - 1]
for dim in ['x','y','z']:
    node_data[("connect1", f"velocity_{dim}")] = dist[cons - 1]
ds =  yt.load_unstructured_mesh(cons - 1, coords, node_data=node_data)

# slice and annotate it
sl = yt.SlicePlot(ds, "z", "test")
sl.annotate_velocity()
sl.save("/tmp/stream_z")
```
resulting in 

```
  File "/home/chavlin/src/yt/yt/visualization/plot_window.py", line 1263, in run_callbacks
    callback(cbw)
  File "/home/chavlin/src/yt/yt/visualization/plot_modifications.py", line 42, in _check_geometry
    return func(self, plot)
  File "/home/chavlin/src/yt/yt/visualization/plot_modifications.py", line 375, in __call__
    return qcb(plot)
  File "/home/chavlin/src/yt/yt/visualization/plot_modifications.py", line 42, in _check_geometry
    return func(self, plot)
  File "/home/chavlin/src/yt/yt/visualization/plot_modifications.py", line 526, in __call__
    pixX = plot.data.ds.coordinates.pixelize(
  File "/home/chavlin/src/yt/yt/geometry/coordinates/cartesian_coordinates.py", line 170, in pixelize
    ftype, fname = field
ValueError: too many values to unpack (expected 2)
```
The issue here is similar to other issues (https://github.com/yt-project/yt/issues/2256 and https://github.com/yt-project/yt/issues/1717), arising because the `VelocityCallback` makes some assumptions about the fields (noted here https://github.com/yt-project/yt/issues/2256#issuecomment-490531376). Those issues seem to be fixed now because `geometry.cartesian_coordinates.pixelize` behaves differently for the mesh/particle types referenced in those issues, but the unstructured mesh still hits that troublesome line 170 in `pixelize`. 

With the fixes here, the above code results in the following image:

stream_z_Slice_z_test

The fix:
1. adds a `determine_field` in `VelocityCallback` immediately after the fields are set (so that we get a full `(fieldtype, fieldname)` tuple
2. sets the `default_fluid_type` to be the same field type as the `default_field` in `load_unstructured_mesh`  (otherwise, the `default_fluid_type` is `gas`, which does not exist...)
3. adds an unstructured mesh test to the existing `visualization.tests.test_callbacks.test_velocity_callback()` function (I could add this as a new test but it seemed appropriate to add there?)

## PR Checklist

### DIFF
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -360,7 +360,7 @@ def fake_tetrahedral_ds():
     return ds
 
 
-def fake_hexahedral_ds():
+def fake_hexahedral_ds(fields=()):
     from yt.frontends.stream.sample_data.hexahedral_mesh import (
         _connectivity,
         _coordinates,
@@ -372,6 +372,9 @@ def fake_hexahedral_ds():
     node_data = {}
     dist = np.sum(_coordinates ** 2, 1)
     node_data[("connect1", "test")] = dist[_connectivity - 1]
+
+    for field in fields:
+        node_data[("connect1", field)] = dist[_connectivity - 1]
 
     # each element gets a random number
     elem_data = {}
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -373,7 +373,7 @@ def fake_hexahedral_ds(fields=None):
     dist = np.sum(_coordinates ** 2, 1)
     node_data[("connect1", "test")] = dist[_connectivity - 1]
 
-    for field in fields:
+    for field in always_iterable(fields):
         node_data[("connect1", field)] = dist[_connectivity - 1]
 
     # each element gets a random number
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -10,6 +10,7 @@ import unittest
 
 import matplotlib
 import numpy as np
+from more_itertools import always_iterable
 from numpy.random import RandomState
 from unyt.exceptions import UnitOperationError
 
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -360,7 +360,7 @@ def fake_tetrahedral_ds():
     return ds
 
 
-def fake_hexahedral_ds(fields=()):
+def fake_hexahedral_ds(fields=None):
     from yt.frontends.stream.sample_data.hexahedral_mesh import (
         _connectivity,
         _coordinates,
--- a/yt/frontends/stream/tests/test_stream_unstructured.py
+++ b/yt/frontends/stream/tests/test_stream_unstructured.py
@@ -66,6 +66,7 @@ def test_multi_field():
     )
 
     ds = load_unstructured_mesh(connect, coords, data)
+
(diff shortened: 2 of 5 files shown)


neutrinoceros:
Without looking at the changes you made, only judging by the picture, I'm a bit suspicious about the uniform direction of the arrows. Maybe that's the correct answer in your example, but I've been bit before when I tried using arrow annotations in non cartesian coordinates.

chrishavlin:
well I was just setting each x,y,z component to the same array (distance from the origin, so always positive values), so in this slice all the arrows are pointing in the direction I'd expect (and their length is determined by how far they are from the origin). It also looked good in the tests I've done with the aspect front end. Here's a fun one: 

aspect_veloc_z

neutrinoceros:
Ok that one looks really cool and useful as a test because I expect it is easy to break.

matthewturk (approved):

yt/testing.py
```diff
@@ -360,7 +360,7 @@ def fake_tetrahedral_ds():
     return ds
 
 
-def fake_hexahedral_ds():
+def fake_hexahedral_ds(fields=()):
```
neutrinoceros, line 363: Empty tuples as default values are bad practice
```suggestion
def fake_hexahedral_ds(fields=None):
```

yt/testing.py
```diff
@@ -373,6 +373,9 @@ def fake_hexahedral_ds():
     dist = np.sum(_coordinates ** 2, 1)
     node_data[("connect1", "test")] = dist[_connectivity - 1]
 
+    for field in fields:
```
neutrinoceros, line 376: you'll need to add `from more_itertools import always_iterable`
```suggestion
    for field in always_iterable(fields):
```

neutrinoceros (commented):
I left a suggestion for improvement, but I'll be happy to approve+merge once it's taken into account

neutrinoceros (approved):

munkm (approved):

neutrinoceros:
Looks like we have a new random bug (not specific to this pr)
@yt-fido test this please

PR #3112, opened 2021-03-06
fix config CLI

neutrinoceros:
## PR Summary

The command line interface is all kinds of broken. In particular, not all documented examples for `yt config set/rm/get` currently work. It's actually to be expected because we don't have proper testing for the CLI.
I am adding some some and am fixing them as I go.

⚠️ 
- This branch is based on #3106 (first usage of toml files nested sectioning)
- requires pytest (YTEP-0035)
- I'm using `benedict` (https://pypi.org/project/python-benedict/) make accessing nested values a little easier. I do think it's worth adding such a nice little dependency in support of the new configuration file format (toml), but I will need to clean up a bit my usage for it after I get all tests to pass.

❗ I did a mistake in drafting this. I assumed every parameter would live it the `[yt]` root section, so this part could be ommited in the api to retrieve any which parameter.

### DIFF
--- a/doc/source/developing/building_the_docs.rst
+++ b/doc/source/developing/building_the_docs.rst
@@ -198,7 +198,7 @@ and build it using Sphinx:
 If all of the dependencies are installed and all of the test data is in the
 testing directory, this should churn away for a while (several hours) and
 eventually generate a docs build.  We suggest setting
-:code:`suppress_stream_logging = True` in your yt configuration (See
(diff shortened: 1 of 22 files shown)


neutrinoceros:
there are too many reasons why this will never make it to the finish line. I now believe that the configuration system is broken enough that it warrant a full refactor. Even the tests I wrote here would need an overhaul so I'd rather start fresh.

Issue #85, opened 2011-06-02
HDF5 set information


yt-fido:
Originally reported by: **The yt Project (Bitbucket: [yt_analysis](https://bitbucket.org/yt_analysis), GitHub: Unknown)**

----------------------------------------

== Imported Ticket ==
 * **Summary:** HDF5 set information
 * **Component:** lagos
 * **Milestone:** 1.0
 * **Reporter:** mturk
 * **Owner:** mturk
 * **Resolution: fixed
 * **Status:** closed
 * **Created:** 1209700128000000
 * **Description:** We need to be able to access the list of HDF5 datasets that are sub-members of a given node via the HDF5LightReader interface.  (Returning a list is perfectly fine.)

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1209747061000000
 * **Field:** status
 * **Oldvalue:** new
 * **Newvalue:** closed

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1209747061000000
 * **Field:** resolution
 * **Oldvalue:** 
 * **Newvalue:** fixed

=== Update to Ticket ===
 * **Author:** mturk
 * **Changetime:** 1209747061000000
 * **Field:** comment
 * **Oldvalue:** 1
 * **Newvalue:** Fixed in r427.

----------------------------------------
- Bitbucket: https://bitbucket.org/yt_analysis/yt/issue/85

PR #3114, opened 2021-03-07
ENH: rich progress bars (YTEP-0039)

neutrinoceros:
## PR Summary

This is the second PR in support to [YTEP-0039](https://github.com/yt-project/ytep/pull/18)
The branch is currently based on #3115

Fortunately we already have a `get_pbar` function that's used everywhere in the code base to create progress bars, so the change is very straightforward.

Rich progress bars are supposed to work well in all sorts of contexts (terminal with various levels of support for colors, as well as notebooks), so I don't expect any complication with CI but life, box, chocolates...
See rich's doc for details on how progress bar should behave in limited environments (or when writing to a file)
https://rich.readthedocs.io/en/stable/console.html?highlight=environment#terminal-detection

I took advantage of rich's api to force the distinction between two ways pbars can be updated using `pbar.update()`:
`yt.func.RichProgressBar.update` now takes keyword arguments only (and should be provided exactly one), which will be either `completed` (state the current completion) or `advance` (increment completion).
In most places where bars are used in yt we implicitly want `completed`, but there are one or two cases where `advance` would be more suited, which allows me to simplify the surrounding code.

Overall I normalised the way pbars are used in the code base and I believe I also fixed some bugs here and there (including the one reported in #3116)

TODO:
- [x] make sure that I'm replacing every usage of `tqdm`
- [x] remove tqdm from required deps
- [x] either deprecate or remove `TqdmProgressBar` and cleanup `RichProgressBar`'s docstring accordingly

Here's a minimal demo script you can use to try this out for yourself

```python
# pbar_demo.py
from yt.funcs import get_pbar
from time import sleep

def main(title, ntasks=100):
    p = get_pbar(title=title, maxval=ntasks)
    for i in range(ntasks):
        sleep(0.05)
        try:
            p.update(i+1) # this syntax only works with the main branch
        except TypeError:
            # this is one correct way to do it with this branch
            p.update(completed=i+1)
            # and within this context, the following line is also equivalent
            # p.update(advance=1)
    p.finish()

if __name__ == "__main__":
    main(title="Loading...")
```
And here's a short video showing how this looks on the main branch (with tqdm), then with this one (with, then without color)

https://user-images.githubusercontent.com/14075922/110307343-16276280-7fff-11eb-8488-2133244dbbd9.mp4

In a few cases we had "progress bars" (no bars really) generated by calling tqdm directly instead of using `yt.funcs.get_pbar`. The reason for that was that the total number of iterations was hard to predict. I added support for such a case, in line with was tqdm was getting us

```python
# tqdm2rich.py
from tqdm import tqdm
from time import sleep

with tqdm(
    desc="pbar without a max (tqdm)"
) as pbar:
    for i in range(100):
        sleep(0.02)
        pbar.update(1)

from yt.funcs import get_pbar

pbar = get_pbar(title="pbar without a max (yt+rich)")
for i in range(100):
    sleep(0.02)
    pbar.update(advance=1)
pbar.finish()
```

https://user-images.githubusercontent.com/14075922/110336401-1f2b2a80-8025-11eb-8936-29ed2774d2a1.mp4

update:
Regarding perfs, this benchmark just came out and shows that rich and tqdm are extremely comparable
https://tqdm.github.io/tqdm/#benchmarks.track_alternatives

## Follow ups

This work will gain support from other PRs, and can be adapted if then get in first, or will know smaller follow up PRs otherwise. Namely

- The present PR has very little actual overlap with #3106 but there's one thing that will need to be addressed as a follow up:  the new proposed parameter `logging.use_color` currently has no effect on progress bars (it should).
- I wrote a patch upstream for pooch in support of this work https://github.com/fatiando/pooch/pull/228.

### DIFF
--- a/yt/funcs.py
+++ b/yt/funcs.py
@@ -20,13 +20,23 @@ import urllib.request
 import warnings
 from functools import lru_cache, wraps
 from numbers import Number as numeric_type
-from typing import Any, Callable, Type
+from typing import Any, Callable, Type, Optional, Sequence, Union
 
 import matplotlib
 import numpy as np
 from more_itertools import always_iterable, collapse, first
 from packaging.version import parse as parse_version
-from tqdm import tqdm
+from rich.progress import (
+    BarColumn,
+    Progress,
+    ProgressColumn,
+    SpinnerColumn,
+    Task,
+    Text,
+    TextColumn,
+    TimeElapsedColumn,
+    TimeRemainingColumn,
+)
 
 from yt.units import YTArray, YTQuantity
 from yt.utilities.exceptions import YTInvalidWidthError
@@ -307,22 +317,72 @@ def insert_ipython(num_up=1):
 #
 
 
-class TqdmProgressBar:
-    # This is a drop in replacement for pbar
-    # called tqdm
-    def __init__(self, title, maxval):
-        self._pbar = tqdm(leave=True, total=maxval, desc=title)
-        self.i = 0
+class CompletionColumn(ProgressColumn):
+    """Renders the current number of iterations completed."""
 
-    def update(self, i=None):
-        if i is None:
-            i = self.i + 1
-        n = i - self.i
-        self.i = i
-        self._pbar.update(n)
+    def render(self, task: Task) -> Text:
+        return Text(f"{task.completed} it", style="progress.percentage")
+
+
+class CompletionSpeedColumn(ProgressColumn):
+    """Renders human readable completion speed."""
+
+    # This is based off rich.progress.TransferSpeedColumn
+
+    def render(self, task: Task) -> Text:
+        """Show data transfer speed."""
+        speed = task.finished_speed or task.speed
+        if speed is None:
+            return Text("?", style="progress.data.speed")
+        data_speed = int(speed)
+        return Text(f"{data_speed} it/s", style="progress.data.speed")
+
+
+class RichProgressBar:
+    """A thin wrapper around rich.progress.Progress"""
+
+    def __init__(
+        self,
+        title: str,
+        maxval: Optional[int] = None,
+        columns: Optional[Sequence[ProgressColumn]] = None,
+    ):
+        if columns is None:
+            columns = [
+                SpinnerColumn(),
+                TextColumn("[progress.description]{task.description}"),
+            ]
+            if maxval is None:
+                columns += [
+                    CompletionColumn(),
+                    TextColumn("[progress.data.speed]@"),
+                    CompletionSpeedColumn(),
+                ]
+                # this is a workaround. As of rich 9.13.0, some of the columns
+                # we're using here require a non-None value in `task.total` to render
+                maxval = np.inf
+            else:
+                columns += [
+                    BarColumn(),
+                    TextColumn("[progress.percentage]{task.percentage:>3.0f}%"),
+                    TimeRemainingColumn(),
+                ]
+            columns += [TimeElapsedColumn()]
+        self._pbar = Progress(*columns)
+
+        self._pbar.start()
+        self.task = self._pbar.add_task(title, total=maxval)
+
+    def update(
+        self, *, completed: Optional[int] = None, advance: Optional[int] = None
+    ) -> None:
+        # We only expose the relevant subset of keyword arguments
+        # supported by `rich.progress.Progress.update`
+        # note that they should NOT be combined (use either `completed` or `advance` but not both)
+        self._pbar.update(self.task, completed=completed, advance=advance)
 
     def finish(self):
-        self._pbar.close()
+        self._pbar.stop()
 
 
 class DummyProgressBar:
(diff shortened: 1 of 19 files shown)


neutrinoceros:
@yt-fido test this please

neutrinoceros:
fix #3209

matthewturk:
So the YTEP is what we're waiting on, right?

neutrinoceros:
> So the YTEP is what we're waiting on, right?

yup

brittonsmith:
As you've seen, this inspired me to implement a rich progress bar for ytree in [PR #85](https://github.com/ytree-project/ytree/pull/85). I like how the new pbars look. The main issue I encountered is that if the progress bar is not implemented as a context manager, there is some strange exception catching that makes it difficult to kill the code with a KeyboardInterrupt. Here is a simple test:

http://paste.yt-project.org/show/332/

If I run this with `python -i` and try to `ctrl-c` out of it, it will stop the execution, but leave me trapped inside. Changing the use in that to a context manager fixes that.

Is there a workaround for this? If not, you might have to implement usage here as a context manager.

Apart from this issue, I am quite in favor of this.

neutrinoceros:
Thank you so much for this feedback. I'm not 100% sure, but this might be considered a bug in rich itself. I'll see what William thinks.

brittonsmith:
One more comment from my recent experience. I've found that you have to be careful not to end up inside multiple progress bars at once, otherwise you will get the error "Only one live display may be active at once". There are a few tricky instances where this can sneak up. For example, we have a progress bar to parse the AMR hierarchy file in Enzo datasets. If a user setup their own progress bar that started just after `yt.load`, this could potentially be active before the hierarchy parse pbar, and you might end up with this error.

I encountered a situation just like this in ytree and one solution I've found is to create a global `current_pbar` that gets returned by `get_pbar` is one has already been created. There may be other ways around this, but this seems to work. I should mention that I only see this happen when the progress bar is used as a context manager, so if you've found a workaround for the other issue I brought up above, this may not be a problem.

brittonsmith:
One last thing I've found is that terminal output is swallowed in a strange way if I have a debugger inside an active progress bar. For example, just the following:
```
from rich.progress import Progress

with Progress() as pbar:
    for i in range(10):
        breakpoint()
```

The debugger is running, but everything I type is swallowed and the cursor is gone. This could make developing a bit difficult if it's the expected/normal behavior.

neutrinoceros:
Ouch. Sounds like something that could be improved upstream, but indeed this is problematic. Thank you very much for experimenting with it, I appreciate it !

brittonsmith:
Sure thing, very glad to. This does seem like it could be a very nice improvement with intriguing possibilities. Hopefully, some of these things can be improved upstream.

yt/funcs.py
```diff
@@ -308,22 +318,72 @@ def insert_ipython(num_up=1):
...
+            ]
+            if maxval is None:
+                columns += [
+                    CompletionColumn(),
+                    TextColumn("[progress.data.speed]@"),
+                    CompletionSpeedColumn(),
+                ]
+                # this is a workaround. As of rich 9.13.0, some of the columns
+                # we're using here require a non-None value in `task.total` to render
+                maxval = np.inf
```
neutrinoceros, line 364: no need to use numpy here actually
```suggestion
                maxval = float("inf")
```

neutrinoceros:
Ach, so sadly your example was lost in the limbs @brittonsmith. I think I may be able to inspect this further in the near future, any chance you'd be able to summon another example to illustrate your problem ?

Here's a template you can use to embed your example in a GH comment, in a collapsible way:

```
<details>
<summary>Example</summary>

    ```python
    from example import code

    do_stuff()
    ```
</details>

```

Which looks like this
<details>
<summary>Example</summary>

```python
 from example import code

 do_stuff()
```
</details>

(note that the template adds a layer of indendation that should be removed in the actual comment)

brittonsmith:
@neutrinoceros, I think my main issue what was debuggers get swallowed and mangled in a way that is tough to describe. Typed text is quickly deleted, you can't end the session with a keyboard interrupt (i.e., ctrl-c). I think if you run this, you'll see what I mean:
<details>
<summary>Example</summary>

    ```python
    from rich.progress import Progress

    with Progress() as progress:
        breakpoint()
    ```
</details>

neutrinoceros:
yeah it looks to me that the experience isn't that much greater with tqdm actually (though text don't get swallowed up, so that's probably something that could be improved upstream. Do you want to report it yourself ?

brittonsmith:
> yeah it looks to me that the experience isn't that much greater with tqdm actually (though text don't get swallowed up, so that's probably something that could be improved upstream. Do you want to report it yourself ?

I'm not sure what you mean about the tqdm experience. In the following example, debugging and keyboard interrupts work exactly as I would expect and hope.
<details>
<summary>this thing</summary>

    import yt

    pbar = yt.get_pbar("Thing", 10)
    for i in range(10):
        pbar.update(i+1)
        breakpoint()
    pbar.finish()
</details>

I couldn't find a link to the repo in the rich docs, so I'm not sure how to report this. If you're in communication with the developers, can you let them know?

neutrinoceros:
It's right here https://github.com/willmcgugan/rich

brittonsmith:
Thanks, I've opened an [issue](https://github.com/willmcgugan/rich/issues/1465).

neutrinoceros:
Thank you, I've subscribed to that issue so I don't forget about It :)

matthewturk:
@neutrinoceros @brittonsmith looks like we're still waiting on that bug, right?  Is that a full blocker for this?

neutrinoceros:
Yeah I think so. I don't think it's worth prettifying our progress bars if it degrade the debugging experience. I know Will wants to fix it upstream but it seems like a nasty one so I guess this part of the YTEP is going to wait. On the other hand I don't think there's such a blocking issue with the [other rich PR](https://github.com/yt-project/yt/pull/3106) (logging), except for the fact that it's currently broken and I need to fix it :/

neutrinoceros:
Withdrawing the YTEP for reasons detailed here: https://github.com/yt-project/ytep/pull/18#issuecomment-998747132

PR #3115, opened 2021-03-08
cleanup: simplify get_pbar by removing a vestigial class that's not used anymore

neutrinoceros:
## PR Summary

related to YTEP-0039 but not part of it 
I noticed the `parallel` flag from `yt.funcs.get_pbar` wasn't used anywhere so I removed the unnecessary code around it.

(This will conflict with #3114. I'd like the present change to go in first)

### DIFF
--- a/yt/funcs.py
+++ b/yt/funcs.py
@@ -337,21 +337,7 @@ class DummyProgressBar:
         return
 
 
-class ParallelProgressBar:
-    # This is just a simple progress bar
-    # that prints on start/stop
-    def __init__(self, title, maxval):
-        self.title = title
-        mylog.info("Starting '%s'", title)
-
-    def update(self, *args, **kwargs):
-        return
-
-    def finish(self):
-        mylog.info("Finishing '%s'", self.title)
-
-
-def get_pbar(title, maxval, parallel=False):
+def get_pbar(title, maxval):
     """
     This returns a progressbar of the most appropriate type, given a *title*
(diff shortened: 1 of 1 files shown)


brittonsmith:
It is used in yt_astro_analysis. I am getting rid of it with some new changes, but maybe we can hold off on this so I don't have to fix yt-astro first.

neutrinoceros:
I think the fix is simply to replace any instance of `ParallelProgressBar` with `DummyProgressBar`
I can either submit a patch to yt_astro_anaylsis or deprecate de class here, what do you prefer ?

neutrinoceros:
Actually @brittonsmith , I went to yt_astro_analysis to try and it looks to mean it's not used there (anymore ?).

matthewturk (approved):

munkm:
@brittonsmith let us know when this is fixed in astro_analysis and we can merge after that.

neutrinoceros:
I've double checked and I confirm that `git grep ParallelProgressBar` returns nothing on the master (and only) branch there.

munkm (approved):

neutrinoceros:
Looks like this is actually mergeable with 2 approvals ? @matthewturk @munkm ?

brittonsmith (approved):

brittonsmith:
For posterity, I finally remembered what package uses this. It's one of mine. I'll move the code there.

PR #3118, opened 2021-03-09
fix a missing loadname in sample dataset registry

neutrinoceros:
## PR Summary

I was able to discover this with my reworked version of `yt.load_sample` (https://github.com/yt-project/yt/pull/3089)

Note that this dataset in particular is a bit tricky to load as it really requires two steps
```python
import yt
ds_parent = yt.load_sample("output_00080")
ds = yt.load_sample("output_00080_halos", parent_ds=ds_parent)
```

Unfortunately, this can not be expressed in `load_kwargs`... and json files don't support comments. I don't know how it should be documented.

### DIFF
--- a/yt/sample_data_registry.json
+++ b/yt/sample_data_registry.json
@@ -729,7 +729,7 @@
   "output_00080_halos.tar.gz": {
     "hash": "fb8f573ea7cb183e1531091b3fce91cf7015cfdf5ccbce9ef2742e3135b52d4c",
     "load_kwargs": {},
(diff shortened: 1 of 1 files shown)


cphyc (approved):
LGTM assuming the tests pass!

neutrinoceros:
@munkm @jcoughlin11 ?

neutrinoceros:
actually I can't reproduce the issue now, this dataset loads fine. Closing this

PR #3119, opened 2021-03-09
blacklist RecursionError to reveal failing tests

neutrinoceros:
## PR Summary
Attempt to solve a recursion errors that are currently caught in tests.
First things first: stop catching them and see how bad the situation is (how many tests fall).

### DIFF
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -489,7 +489,7 @@ class FieldInfoContainer(dict):
         # - a greylist (exceptions that may be covering bugs but should be checked)
         # See https://github.com/yt-project/yt/issues/2853
         # in the long run, the greylist should be removed
-        blacklist = ()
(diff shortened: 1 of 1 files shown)


neutrinoceros:
here's a simple one
```python
import yt

r0 = "output_00080/info_00080.txt"
r1 = "output_00080_halos/tree_bricks080"

ds_parent = yt.load(r0)
ds = yt.load(r1, parent_ds=ds_parent)

ds.derived_field_list

```
Traceback
```
<class 'RecursionError'> raised for field ('deposit', 'halos_cic')
```

matthewturk:
Thanks for identifying these!

neutrinoceros:
They are a pain to work through though. In an hour of work yesterday I didn't manage to solve any. And the later this is fixed the most likely it will get worse.

PR #3121, opened 2021-03-10
fix a broken load_name in data sample metadata

neutrinoceros:
## PR Summary

```python
import yt
yt.load_sample("CfRadialGrid")
```
fails with 
```
FileNotFoundError
```

This fixes it.
Note that this particular dataset still doesn't load (`YTUnidentifiedDataType`). I believe it was added in support to #1990 which is still in the shop. 

I except this will fail the same way #3118 does: the hash probably needs to be updated as well.
edit: nope. Must be a problem with changing the type (from null to str) of the field, which is not a concern here.


chrishavlin (approved):
great, thanks for the quick fix!

munkm (approved):

PR #3124, opened 2021-03-12
making ParticleContainers pickleable

chrishavlin:
## PR Summary

Update: this PR has broadened a bit to include some wider pickle and `ParameterFileStore` fixes. Happy to split into separate PRs if warranted. 

The main update that may be worth some discussion is that I'm modifying the base `__reduce__` used by `YTDataContainer`. Since `save_as_dataset` is now the recommended method for long term dataset storage and the pickling seems to have been broken for some time, I assumed this was OK. But it does mean that if there happened to be any legacy pickled datasets out there for which the reload was NOT broken, this change will definitely break them. 

Original: 

I'm not sure this qualifies as a bug per se as this isn't something that you would do accidentally (feel free to remove the bug tag).... but currently on main, if you try to pickle a chunk of a particle data container, the pickling fails:

```python
import yt
import pickle
ds = yt.load("snapshot_033")
ad = ds.all_data()
ds.index._identify_base_chunk(ad)
chunks = list(ds.index._chunk_io(ad, cache=False))
chu_p = pickle.loads(pickle.dumps(chunks[0]))
```

results in 

```
  File "/home/chavlin/.pyenv/versions/yt_dev/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3427, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-69d3e1f050db>", line 7, in <module>
    chu_p = pickle.loads(pickle.dumps(chunks[0]))
  File "/home/chavlin/src/yt/yt/data_objects/data_containers.py", line 1535, in _reconstruct_object
    cls = getattr(ds, dtype)
AttributeError: 'OWLSDataset' object has no attribute 'particle_container'
```

**Why would you do this?** This came up during some Daskening  (related PR to the Daskening branch to follow soon...), in which I distribute chunks of a particle dataset to dask. 

**The fix** The issue comes from the fact that the `ParticleContainer` objects embedded in the `index` chunks are not pickleable. The `ParticleContainer` inherits a `__reduce__` method from the general `YTDataContainer`, which was designed for unpickling datasets. The fix (1) overrides the `__reduce__` method in `ParticleContainer`  and (2) adds the `base_selector` as an explicit argument to `ParticleContainer.__init__`. The reason for thing (2) is that having the `base_selector` as an argument avoids the need to rebuild the index on unpickling a chunk. 

I'm open to other approaches -- I think this behavior indicates a wider refactoring of the indexing/chunking might be warranted, but not sure where to start? and I think adding a unit test might make sense here?

### DIFF
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -1336,7 +1336,7 @@ class YTDataContainer:
         []
         """
         args = self.__reduce__()
-        return args[0](self.ds, *args[1][1:])[1]
+        return args[0](self.ds, *args[1][1:])
 
     def __repr__(self):
         # We'll do this the slow way to be clear what's going on
@@ -1483,29 +1483,16 @@ class YTDataContainer:
                 o.field_parameters = cache_fp
 
 
-# Many of these items are set up specifically to ensure that
-# we are not breaking old pickle files.  This means we must only call the
-# _reconstruct_object and that we cannot mandate any additional arguments to
-# the reconstruction function.
+# PR3124: Given that save_as_dataset is now the recommended method for saving
+# objects (see Issue 2021 and references therein), the following has been re-written.
+#
+# Original comments (still true):
 #
 # In the future, this would be better off being set up to more directly
 # reference objects or retain state, perhaps with a context manager.
 #
 # One final detail: time series or multiple datasets in a single pickle
 # seems problematic.
-
-
-class ReconstructedObject(tuple):
-    pass
-
-
-def _check_nested_args(arg, ref_ds):
-    if not isinstance(arg, (tuple, list, ReconstructedObject)):
-        return arg
-    elif isinstance(arg, ReconstructedObject) and ref_ds == arg[0]:
-        return arg[1]
-    narg = [_check_nested_args(a, ref_ds) for a in arg]
-    return narg
 
 
 def _get_ds_by_hash(hash):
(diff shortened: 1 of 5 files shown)


matthewturk (approved):
This looks good to me.  I'm not sure that tests beyond "can I pickle and unpickle and get the same results" are necessary.  Those could be done exclusively in-memory, as well, without writing to disk.

chrishavlin:
"I'm not sure that tests beyond "can I pickle and unpickle and get the same results" are necessary." -- ya, that's what I was thinking. Just a test of `pickle.loads(pickle.dumps())`. Will add that...

chrishavlin:
Converting to draft for now: in writing a "simple" unit test I found some wider pickle issues. Almost have a fix...

matthewturk:
Failing unit test is in athena, which is not specific to this PR

chrishavlin:
Switching back to ready for review... as noted in my update to the PR summary, I'm now modifying the base `__reduce__` used by `YTDataContainer`. Seems OK to me since `save_as_dataset` is now the recommended method for long term dataset storage and the pickling seems to have been broken for some time (e.g., https://github.com/yt-project/yt/issues/2021). Any reason this might be an issue? 

Not sure about that failing test -- Matt, are you saying it's unrelated to my changes? I haven't been able to reproduce it locally, not sure how to address it...

matthewturk:
I believe it is unrelated to your changes.

matthewturk (approved):
I've re-read this after the new changes and LGTM!

jzuhone (approved):

PR #3127, opened 2021-03-14
[bugfix] Various SPH fixes

jzuhone:
This PR fixes a few issues related to SPH. 

* Arepo needed to have `"cmcm"` defined in its `_unit_base` but it did not.
* There was a string in the `particle_oct_container` code which didn't have enough format specifiers. I've fixed it and simply changed it to an f-string. 
* In `IOHandlerGadgetBinary`, we were not skipping a chunk/particle type combination in the `_read_particle_coords` method if it had no particles, which is what we should be doing (and what we do in `IOHandlerGadgetHDF5`). I've simply applied the same fix that is used in `IOHandlerGadgetHDF5`.

### DIFF
--- a/yt/frontends/arepo/data_structures.py
+++ b/yt/frontends/arepo/data_structures.py
@@ -77,6 +77,7 @@ class ArepoHDF5Dataset(GadgetHDF5Dataset):
 
     def _set_code_unit_attributes(self):
         self._unit_base = self._get_uvals()
+        self._unit_base["cmcm"] = 1.0 / self._unit_base["UnitLength_in_cm"]
         super()._set_code_unit_attributes()
         munit = np.sqrt(self.mass_unit / (self.time_unit ** 2 * self.length_unit)).to(
             "gauss"
--- a/yt/frontends/arepo/data_structures.py
+++ b/yt/frontends/arepo/data_structures.py
@@ -76,7 +76,7 @@ class ArepoHDF5Dataset(GadgetHDF5Dataset):
         else:
             # We assume this is comoving, because in the absence of comoving
             # integration the redshift will be zero.
-            self._unit_base["cmcm"] = 1.0 / self._unit_base["UnitLength_in_cm"]
+            uvals["cmcm"] = 1.0 / uvals["UnitLength_in_cm"]
         return uvals
 
     def _set_code_unit_attributes(self):
--- a/yt/frontends/arepo/data_structures.py
+++ b/yt/frontends/arepo/data_structures.py
@@ -77,7 +77,10 @@ class ArepoHDF5Dataset(GadgetHDF5Dataset):
 
     def _set_code_unit_attributes(self):
         self._unit_base = self._get_uvals()
-        self._unit_base["cmcm"] = 1.0 / self._unit_base["UnitLength_in_cm"]
+        if "UnitLength_in_cm" in self._unit_base:
+            # We assume this is comoving, because in the absence of comoving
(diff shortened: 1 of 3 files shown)


matthewturk (approved):

yt/frontends/arepo/data_structures.py
```diff
@@ -77,6 +77,10 @@ def _get_uvals(self):
 
     def _set_code_unit_attributes(self):
         self._unit_base = self._get_uvals()
+        if self._unit_base is not None and "UnitLength_in_cm" in self._unit_base:
+            # We assume this is comoving, because in the absence of comoving
+            # integration the redshift will be zero.
+            self._unit_base["cmcm"] = 1.0 / self._unit_base["UnitLength_in_cm"]
```
munkm, line 83: Would it make sense to move this logic inside the `if unit in handle` block in the `_get_uvals` defintion? It seems like some of this logic is replicating that.

yt/frontends/gadget/io.py
```diff
@@ -360,6 +360,8 @@ def _read_particle_coords(self, chunks, ptf):
             tp = data_file.total_particles
             f = open(data_file.filename, "rb")
             for ptype in ptf:
+                if tp[ptype] == 0:
+                    continue
```
munkm, line 364:
```suggestion
                    # skip if there are no particles 
                    continue
```

munkm (approved):
Looks good to me. I have one minor question but I'm happy to merge if you don't feel like changing it.

yt/frontends/arepo/data_structures.py
jzuhone: I moved this inside `_get_uvals`--thanks for the suggestion!

yt/frontends/arepo/data_structures.py
```diff
@@ -73,6 +73,10 @@ def _get_uvals(self):
         handle.close()
         if missing:
             uvals = None
+        else:
+            # We assume this is comoving, because in the absence of comoving
+            # integration the redshift will be zero.
+            self._unit_base["cmcm"] = 1.0 / self._unit_base["UnitLength_in_cm"]
```
munkm, line 79: I think you want `uvals[]` and here ant not `self._unit_base[]`

yt/frontends/arepo/data_structures.py
```diff
@@ -73,6 +73,10 @@ def _get_uvals(self):
         handle.close()
         if missing:
             uvals = None
+        else:
```
munkm, line 76: Is there a chance `"UnitLength_in_cm"` would not exist if `"UnitMass_in_g"` or `"UnitVelocity_in_cm_per_s"` does? Maybe it would be more robust with 

```suggestion
        if uvals["UnitLength_in_cm"]:
```

?

munkm (approved):

yt/frontends/arepo/data_structures.py
jzuhone: Doh, yes! Thanks!

yt/frontends/arepo/data_structures.py
jzuhone: Yes, I agree.

jzuhone: Actually, this doesn't work, because the logic of the code demands that all three exist for the units to be specified at all.

munkm: ok! then leaving it as is is good!

PR #3137, opened 2021-03-23
cleanup: remove turbo colormap (added to matplotlib in version 3.3.0)

neutrinoceros:
## PR Summary
A year ago, @brittonsmith and I added support for `turbo` and `turbo_r` colormaps respectively.
Since then, it's been added in matplotlib (version 3.3.0). I don't think it should be kept in yt because our registration mechanism shadows the colormap that ships with Matplotlib, which may be updated (though unlikely) in the future.

Note that the cmap was shipped as part of yt 3.6.0 and 3.6.1


munkm (approved):
I don't see any issue with this. Can anybody think of a user case where you'd want to use yt 4 without matplotlib 3.3 and the turbo colormap?

matthewturk (approved):

PR #3138, opened 2021-03-23
cleanup: remove yt.funcs.get_requests, use yt.utilities.on_demand_imports

neutrinoceros:
## PR Summary

This is a quick follow up to #3100, cleaning up a unique mechanism that was used to specify requests as an optional dep, which is not necessary any more.


neutrinoceros:
I guess I've created a circular dependency somehow, working on a fix

neutrinoceros:
Switching to daft and decomposing the original commit into 4 steps (one per file) until I find the one that broke the install in CI

jzuhone (approved):

chrishavlin (approved):
looks good!

PR #3135, opened 2021-03-23
Enable CI for pytest [02/NN]

Xarthisius:
## PR Summary

Enable CI for tests that were merged with https://github.com/yt-project/yt/pull/3102. There are three additional jobs being enabled with this PR: unittest on GHA, unit and answer tests on Jenkins called *py38_unittests* and *py38_answertests* respectively. The latter will fail due to missing answers. The proposed way to deal with answer generation is:
1. pytest run @ Jenkins uses https://github.com/yt-project/answer-store to look for answers
2. if answers are not found, test are marked as failed for proper reporting (this is a change: before they were marked as skipped)
3. After the main run is finished, all tests that failed due to lack of answers are re-run with `--answer-store`
4. Resulting artifacts are published as Jenkins job artifacts (see e.g. https://tests.yt-project.org/job/yt_py38_answertests/4/ Build Artifacts > TestGDF_000.yaml)
5. Contributor creating a PR that needs new answers to be generated is expected to manually download required yaml files and issue a separate PR to https://github.com/yt-project/answer-store (for now).

### NOTE
Jenkins jobs will only run on PRs that have label `infrastructure` set for now.

### TODO

- [x] merge https://github.com/yt-project/answer-store/pull/14
- [x] update `answer-store` to point to the merge commit from previous step

PR #3140, opened 2021-03-23
use distutils instead of pkg_resource to parse version in optional deps

neutrinoceros:
## PR Summary

Fixing the one place in the code base where we use anything else than `distutils.version.LooseVersion` to check versions in optional deps.
Making this uniform will help later in greping for those and specifying requirements for optional dependencies properly (see #3078).

### DIFF
--- a/yt/utilities/on_demand_imports.py
+++ b/yt/utilities/on_demand_imports.py
@@ -1,6 +1,5 @@
 import sys
-
-from pkg_resources import parse_version
+from distutils.version import LooseVersion
 
 
 class NotAModule:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
I suspect the failures on Jenkins are related to @Xarthisius' comment in #3135, given that I'm using the "infrastructure" label here

matthewturk (approved):

neutrinoceros:
(force push after removing the infrastructure label to get the "✅"...)

munkm (approved):

PR #3143, opened 2021-03-24
maint: drop compatibility layers for matplotlib 1.5

neutrinoceros:
## PR Summary

Rationale:
matplotlib doesn't document their supported Python versions very well, which makes it a bit hard to guess which version should be considered minimal for us.

Here's what I got out of my research:
Python 3.6 (our current earliest supported Python version) came out on December 23, 2016
[Matplotlib 2.0.0](https://github.com/matplotlib/matplotlib/releases/tag/v2.0.0) was the very first release after that (January 17, 2017), and came out only a few months after [1.5.3](https://github.com/matplotlib/matplotlib/releases/tag/v1.5.3) (September 6, 2016), which is our  _currently_ earliest tested matplotlib version.

Because we have acquired a small but noticeable amount of technical debt in the form of compatibility layers as we attempt to support the broadest range of versions possible, and since most of it only supports a short lived version from over 4 years ago (1.5.3) that didn't explicitly support Python 3.6, I think it's reasonable to bump this.

For closure, this isn't a _necessary_ change at the time of writing, but I'm hoping it's a small enough breakage that it can be considered for inclusion soon, since it's part of an effort I'm doing to clarify dependency specifications.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -139,7 +139,7 @@ if __name__ == "__main__":
         packages=find_packages(),
         include_package_data=True,
         install_requires=[
-            "matplotlib>=1.5.3",
+            "matplotlib>=2.0.2",
             "setuptools>=19.6",
             "sympy>=1.2",
             "numpy>=1.10.4",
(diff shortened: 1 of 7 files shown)


matthewturk:
I strongly support us being more modern with matplotlib support.

matthewturk (approved):

brittonsmith (approved):

munkm (approved):

Issue #3126, opened 2021-03-13
off_axis_projection can't plot fields which need field parameters


jzuhone:
### Bug report

**Bug summary**

When one attempts to off-axis project a field that needs a field parameter (in this case, `"gas", "velocity_los"`), you end up with a crash because when calculating the field bounds you create a new object with `ds.all_data()` that doesn't know about the field parameter: 

**Code for reproduction**

```python
import yt
ds = yt.load("GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150")
prj = yt.OffAxisProjectionPlot(ds, [0, 0, 1], "velocity_los", weight_field="density")
```

**Actual outcome**

```
NeedsParameter: (['axis'])
---------------------------------------------------------------------------
NeedsParameter                            Traceback (most recent call last)
<ipython-input-3-57190e41cb45> in <module>
----> 1 prj = yt.OffAxisProjectionPlot(ds, [0, 0, 1], "velocity_los", weight_field="density")
~/Source/yt/yt/visualization/plot_window.py in __init__(self, ds, normal, fields, center, width, depth, axes_unit, weight_field, max_level, north_vector, right_handed, volume, no_ghost, le, re, interpolated, fontsize, method, data_source, buff_size)
   2112         # Hard-coding the origin keyword since the other two options
   2113         # aren't well-defined for off-axis data objects
-> 2114         PWViewerMPL.__init__(
   2115             self,
   2116             OffAxisProj,
~/Source/yt/yt/visualization/plot_window.py in __init__(self, *args, **kwargs)
    803             self._plot_type = kwargs.pop("plot_type")
    804         self._splat_color = kwargs.pop("splat_color", None)
--> 805         PlotWindow.__init__(self, *args, **kwargs)
    806 
    807     def _setup_origin(self):
~/Source/yt/yt/visualization/plot_window.py in __init__(self, data_source, bounds, buff_size, antialias, periodic, origin, oblique, right_handed, window_size, fields, fontsize, aspect, setup)
    230                 self._field_transform[field] = linear_transform
    231         self.setup_callbacks()
--> 232         self._setup_plots()
    233 
    234     def __iter__(self):
~/Source/yt/yt/visualization/plot_window.py in _setup_plots(self)
    897             return
    898         if not self._data_valid:
--> 899             self._recreate_frb()
    900             self._data_valid = True
    901         self._colorbar_valid = True
~/Source/yt/yt/visualization/plot_window.py in _recreate_frb(self)
[...]

PR #3147, opened 2021-03-25
Fix field parameters not working with off axis projections

jzuhone:
## PR Summary

This is an attempt to fix issue #3126 to ensure that field parameters can be used with off-axis projections. ~~This is a work in progress.~~ 

I *think* that I only need to restore the underscore that was removed in this commit:

https://github.com/yt-project/yt/commit/cde25a88935c5a0a7b6fa17602bb27b7f2e8a864#diff-2d1d24ee7df2e3b516614aad6fa0d7bd09a0268684a1c6b748eee75c1b823002

which is here, in line 57: 

https://github.com/yt-project/yt/blob/bc33f4e188f9b29a55588a4f9f7e88d65a81f2cd/yt/visualization/volume_rendering/render_source.py#L52-L65

but it sounds like with it in there other things were broken, so we're going to see what happens with the tests. 

EDIT: All tests pass with just this one commit. 

## PR Checklist

### DIFF
--- a/yt/fields/tests/test_vector_fields.py
+++ b/yt/fields/tests/test_vector_fields.py
@@ -3,6 +3,7 @@ import numpy as np
 from yt.testing import assert_allclose_units, fake_random_ds, requires_file
 from yt.units import cm, s
 from yt.utilities.answer_testing.framework import data_dir_load
+from yt.visualization.volume_rendering.off_axis_projection import off_axis_projection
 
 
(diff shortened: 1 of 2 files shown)


jzuhone:
This is ready to go. I see no reason why this change isn't the right one.

munkm (approved):

Xarthisius (approved):

PR #3148, opened 2021-03-25
add pandas to yt.utilities.on_demand_imports

neutrinoceros:
## PR Summary

As remarked by @munkm in https://github.com/yt-project/yt/pull/3089#discussion_r601793845
pandas is currently not part of our `on_demand_imports`, but the only places it's imported are clearly documented as fragile because it's _not_ a hard dependency.


jzuhone (approved):

chrishavlin (approved):
looks good to me

PR #3149, opened 2021-03-26
cleanup: remove cubehelix colormap

neutrinoceros:
## PR Summary

followup to #3143 and fix for a failing workflow in #3078 with the following error message
```
ValueError: Trying to re-register the builtin cmap 'cubehelix'.
```

Indeed we don't need to vendor this colormap anymore since it was already present in matplotlib 2.0.2
See https://matplotlib.org/2.0.2/users/colormaps.html?highlight=cubehelix

Note that registering an existing cmap name actually causes an error in matplotlib 3.4

### DIFF
--- a/yt/visualization/color_maps.py
+++ b/yt/visualization/color_maps.py
@@ -173,38 +173,6 @@ cdict = {
 
 add_colormap("purple_mm", cdict)
 
-# This one comes from
-# http://permalink.gmane.org/gmane.comp.python.matplotlib.devel/10518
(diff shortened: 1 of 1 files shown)


neutrinoceros:
While I'm at it I also remove an error catching that becomes useless after we dropped support for matplotlib < 2.0.2

cphyc (approved):

neutrinoceros:
answers are failing. I don't think it has anything to do with the PR :/

munkm (approved):

PR #3150, opened 2021-03-26
maint: pin matplotlib maximal supported version

neutrinoceros:
## PR Summary

This is a hot fix for failing answer-tests following the release of matplotlib 3.4.0
Obviously this boundary should be lifted in the near future.

### DIFF
--- a/tests/test_requirements.txt
+++ b/tests/test_requirements.txt
@@ -5,7 +5,7 @@ fastcache~=1.0.2
 glueviz~=0.13.3
 h5py~=3.1.0
 ipython~=7.6.1
-matplotlib~=3.3
+matplotlib<3.4
 nose-timer~=1.0.0
 nose~=1.3.7
 pandas~=1.1.2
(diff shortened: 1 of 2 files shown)


tests/test_requirements.txt
```diff
@@ -5,7 +5,6 @@ fastcache~=1.0.2
 glueviz~=0.13.3
 h5py~=3.1.0
 ipython~=7.6.1
-matplotlib~=3.3
```
Xarthisius, line 8: Please don't remove that. Change it to `matplotlib<3.4.0` instead.

neutrinoceros: Fine for now, but FYI I'm trying to move yt to a better place in terms of dependency specifications, so I'll be working on this later.

matthewturk: @neutrinoceros Before you dig any deeper on that, can we have a video conference about it so that (at least) I can understand it?  I'm anxious about moving too much since the pytest stuff is startign to get pulled in, and we don't want to have to move stuff on jenkins etc around too much just yet.

neutrinoceros: Sure !

Issue #1976, opened 2018-08-15
Cut Regions fail for particle-based fields


chummels:
### Bug report

**Bug summary**

Cut regions, a method for filtering on fields, fail for particle-based fields.  See yt docs here: http://yt-project.org/docs/dev/analyzing/filtering.html#cut-regions

**Code for reproduction**

```python
import yt
ds = yt.load('FIRE_M12i_ref11/snapshot_600.hdf5')
ad = ds.all_data()
cool = ad.cut_region(["obj['temperature'] < 1e5"])
yt.ProjectionPlot(ds, 'x', ('gas', 'density'), data_source=cool).save()
```

**Actual outcome**

```
python ~/src/yt-data/test_cut_region.py
/Users/chummels/src/yt-conda/lib/python3.6/site-packages/cmocean/tools.py:76: MatplotlibDeprecationWarning: The is_string_like function was deprecated in version 2.1.
  if not mpl.cbook.is_string_like(rgbin[0]):
yt : [INFO     ] 2018-08-15 16:18:08,198 Calculating time from 1.000e+00 to be 4.355e+17 seconds
yt : [INFO     ] 2018-08-15 16:18:08,198 Assuming length units are in kpc/h (comoving)
yt : [INFO     ] 2018-08-15 16:18:08,221 Parameters: current_time              = 4.3545571088051386e+17 s
yt : [INFO     ] 2018-08-15 16:18:08,221 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2018-08-15 16:18:08,222 Parameters: domain_left_edge          = [ 0.  0.  0.]
yt : [INFO     ] 2018-08-15 16:18:08,222 Parameters: domain_right_edge         = [ 60000.  60000.  60000.]
yt : [INFO     ] 2018-08-15 16:18:08,222 Parameters: cosmological_simulation   = 1
yt : [INFO     ] 2018-08-15 16:18:08,223 Parameters: current_redshift          = 0.0
yt : [INFO     ] 2018-08-15 16:18:08,223 Parameters: omega_lambda              = 0.728
yt : [INFO     ] 2018-08-15 16:18:08,223 Parameters: omega_matter              = 0.272
yt : [INFO     ] 2018-08-15 16:18:08,223 Parameters: hubble_constant           = 0.702
yt : [INFO     ] 2018-08-15 16:18:08,426 Allocating for 4.787e+06 particles
[...]

PR #3152, opened 2021-03-26
Update filtering docs to reflect that particles should not be used with cut regions

jzuhone:
## PR Summary

Cut regions fail for particle-based fields, as pointed out in issue #1976. 

Some work probably needs to be done to decide if we want to support them in cut regions at all, but @chummels made a good suggestion in that issue that we should update the docs to reflect the fact that particle filters should be used instead. This PR does that, and fixes an error on the same page. 

EDIT: We now provide a function that checks if the fields provided to filter a `YTCutRegion` object are mesh fields, and throws an error if not.

### DIFF
--- a/doc/source/analyzing/filtering.rst
+++ b/doc/source/analyzing/filtering.rst
@@ -101,7 +101,7 @@ filtering out unwanted regions. Such wrapper functions are methods of
    overpressure_and_fast = overpressure_and_fast.include_above('velocity_magnitude', 1e2, 'km/s')
 
    print('Density of all data: ad["density"] = \n%s' % ad['density'])
-   print('Density of "overpressure and fast" data: overpressure_and_fast['density'] = \n%s' %
+   print('Density of "overpressure and fast" data: overpressure_and_fast["density"] = \n%s' %
           overpressure_and_fast['density'])
 
 The following exclude and include functions are supported:
--- a/doc/source/analyzing/filtering.rst
+++ b/doc/source/analyzing/filtering.rst
@@ -120,9 +120,12 @@ The following exclude and include functions are supported:
 
 .. warning::
 
-    Cut regions can have unexpected results with particle fields, or often not work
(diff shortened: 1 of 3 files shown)


munkm (approved):
I think these changes are good! 

Maybe we should add a warning log message in the `cut_region` function if it's clear it's being used on a particle field. What do you think?

jzuhone:
@munkm or maybe an error?

munkm:
Either works for me! Since the doc you wrote says it is "unstable" I at least thing users should be warned if it doesn't crash.

jzuhone:
@yt-fido test this please

doc/source/analyzing/filtering.rst
```diff
@@ -118,8 +118,11 @@ The following exclude and include functions are supported:
    - :func:`~yt.data_objects.data_containers.YTSelectionContainer3D.exclude_below` - Exclude values below given value
 
 
-Cut regions can also operate on particle fields, but a single cut region object
-cannot operate on both particle fields and mesh fields at the same time.
+.. warning::
+
+    Cut regions can have unexpected results with particle fields, or often not work
+    at all with some operations. If you want to filter particle fields, see the
+    next section :ref:`filtering-particles` instead.
```
munkm, line 125:
```suggestion
    Cut regions are unstable when used onparticle fields, and will currently
    error if you pass a particle field through a cut region. 
    If you want to filter particle fields, see the
    next section :ref:`filtering-particles` instead.
```

munkm (approved):
Since you're using an error now I'd update the doc too! But this is great and thank you for the changes.

doc/source/analyzing/filtering.rst
jzuhone: I've got a slightly modified version of this I'll add

munkm: Totally cool!

matthewturk (approved):

PR #3154, opened 2021-03-27
rewrite load tests with the pytest framework

neutrinoceros:
## PR Summary

I'm experiencing issues with these tests running on pytest in https://github.com/yt-project/yt/pull/3062, specifically:

https://tests.yt-project.org/job/yt_py38_unittests/18/testReport/junit/yt.tests/test_load_errors/test_load_nonexistent_data/

I originally wrote these tests for pytest but had to rewrite them to get them merged, so I'm rewriting them back in hopes that will help fixing the initial problem in #3062 

For reference I could not reproduce the issue locally but I assume pytest + Jenkins is less well behave with tests using a different framework.

### DIFF
--- a/nose_unit.cfg
+++ b/nose_unit.cfg
@@ -6,4 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
+ignore-files=test_load_errors.py
 exclude-test=yt.frontends.gdf.tests.test_outputs.TestGDF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -175,6 +175,7 @@ other_tests:
   unittests:
      - '--exclude=test_mesh_slices'  # disable randomly failing test
      - '--ignore=test_outputs_pytest'
+     - '--ignore-files=test_load_errors.py'
      - '--exclude-test=yt.frontends.gdf.tests.test_outputs.TestGDF'
   cookbook:
      - 'doc/source/cookbook/tests/test_cookbook.py'
--- a/yt/tests/test_load_errors.py
+++ b/yt/tests/test_load_errors.py
@@ -1,10 +1,7 @@
-import os
-import tempfile
-from pathlib import Path
+import pytest
 
 from yt.data_objects.static_output import Dataset
 from yt.loaders import load, load_simulation
-from yt.testing import assert_raises
 from yt.utilities.exceptions import (
     YTAmbiguousDataType,
     YTSimulationNotIdentified,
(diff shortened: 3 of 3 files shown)


neutrinoceros:
Since this change alone doesn't trigger the problematic job on Jenkins yt_py38_unittests, we should see the interesting results in #3062 instead

neutrinoceros:
Ah yes, this test file is actually part of the nose-based GH workflow as well, so no wonder this crashes as well.
Welp @Xarthisius, @jcoughlin11, any idea how to go around  this without adding to our technical debt ? Should I just wait for nose to be fully dropped ?

neutrinoceros:
Note that in #3062's log we can see that it didn't fix the problem: my tests relying on tmp dir are apparently broken on fido no matter how I write them (while they work in any case locally)
https://tests.yt-project.org/job/yt_py38_unittests/23/testReport/junit/yt.tests/test_load_errors/test_load_not_a_file/

munkm:
Would you mind renaming this PR to something clearer since you're only modifying one test? It's helpful when I'm writing the changelog.

Xarthisius:
@yt-fido test this please.

py38_answertests are failing cause it needs #3135

Xarthisius:
py38_unittests passes and new tests [were run](https://tests.yt-project.org/job/yt_py38_unittests/24/testReport/yt.tests/test_load_errors/).

> Ah yes, this test file is actually part of the nose-based GH workflow as well, so no wonder this crashes as well.
> Welp @Xarthisius, @jcoughlin11, any idea how to go around this without adding to our technical debt ? Should I just wait for nose to be fully dropped ?

I'm not sure how to quantify the increase of technical debt introduced by 030b3e7 but I hope it's on the acceptable level. There's no need  to wait for nose to be fully dropped.

neutrinoceros:
It's actually 8 tests but yeah it's only one file. I'll update the title

neutrinoceros:
@yt-fido test this please

neutrinoceros:
> I'm not sure how to quantify the increase of technical debt introduced by 030b3e7 but I hope it's on the acceptable level. There's no need to wait for nose to be fully dropped.

It seems reasonable to me, with the condition that the test file is still run somewhere, and that it will eventually be tested in GH workflows again. To be clear, where is that commit from ? is it part of any active PR, or were you suggesting I apply it here ?

Xarthisius:
> It seems reasonable to me, with the condition that the test file is still run somewhere, and that it will eventually be tested in GH workflows again. To be clear, where is that commit from ? is it part of any active PR, or were you suggesting I apply it here ?

It's already here (see commits). I used the sneaky "feature" that allows me to remotely hack your github repo.

neutrinoceros:
Ok cool, fine by me. Thanks for doing this !

Xarthisius (approved):

jzuhone (approved):

PR #3156, opened 2021-03-28
fix: typing for Dataset.cosmological_simulation

neutrinoceros:
## PR Summary
This attribute is supposed to be an int and is wrongly set to a float in some frontends.
It should really be an bool but that's beyond the scope of this PR

edit: while I'm at it I split separate statements into separate lines to clarify formatting in these regions

### DIFF
--- a/yt/fields/field_detector.py
+++ b/yt/fields/field_detector.py
@@ -52,9 +52,10 @@ class FieldDetector(defaultdict):
             # required attrs
             ds = fake_dataset(lambda: 1)
             ds["Massarr"] = np.ones(6)
-            ds.current_redshift = (
-                ds.omega_lambda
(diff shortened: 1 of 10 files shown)


matthewturk (approved):
Seems OK, but I'm also ambivalent about float/int/bool for this.  ;-)

neutrinoceros:
What do you mean ?
My personal opinion is that this parameter _should_ probably be a bool, but I don't know why it was made an int in the first place so I'm not questioning it here. In any case it's _documented_ as an int.

matthewturk:
I mean, this is a good reason to change it to an int everywhere.  That's
all.  :-)

brittonsmith (approved):

PR #3158, opened 2021-03-29
Bugfix #3155 + improvement in config handler

### LINKED ISSUE #3155: YtConfig __getitem__ is slightly broken
### Bug report

If one attempts to display the main (and currently only) section of the configuration from an interactive session it should be doable like this
```python
from yt.config import ytcfg
print(ytcfg["yt"])
```
Though currently this fails with

```
KeyError: 'Cannot get key y'
```
Because the underlying code was only ever tested with double keys, the string`"yt"` is treated as an iterable.
It's probably an easy fix but I'd rather just report it for now.
ping @cphyc in case you're interested in looking this up

cphyc:
## PR Summary

This fixes #3155 and improves support for inclusion statement (i.e. you can now use `"thing" in ytcfg`).

As a nice addition, this also adds rich display in the notebook (almost) for free
image

### DIFF
--- a/yt/config.py
+++ b/yt/config.py
@@ -149,11 +149,11 @@ class YTConfig:
         )
 
     def __setitem__(self, args, value):
-        section, *keys = args
+        section, *keys = always_iterable(args)
         self.set(section, *keys, value, metadata=None)
 
     def __getitem__(self, key):
-        section, *keys = key
+        section, *keys = always_iterable(key)
         return self.get(section, *keys)
 
     def remove(self, *args):
--- a/yt/config.py
+++ b/yt/config.py
@@ -195,6 +195,8 @@ class YTConfig:
     def __contains__(self, item):
         return item in self.config_root
 
+    # Add support for IPython rich display
+    # see https://ipython.readthedocs.io/en/stable/config/integrating.html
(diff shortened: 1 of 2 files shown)


yt/config.py
```diff
@@ -99,13 +99,15 @@ def __init__(self, defaults=None):
...
-        return self.config_root.get_leaf(section, *keys, callback=callback)
+    def get(self, section, *keys, callback=None):
+        node_or_leaf = self.config_root.get(section, *keys)
+        if isinstance(node_or_leaf, ConfigLeaf):
+            if callback is None:
+                return node_or_leaf.value
+            else:
+                return callback(node_or_leaf)
+        else:
+            return node_or_leaf
```
neutrinoceros, line 110:
```suggestion
        if isinstance(node_or_leaf, ConfigLeaf):
            if callback is not None:
                return callback(node_or_leaf)
            return node_or_leaf.value
        return node_or_leaf
```

matthewturk (approved):

neutrinoceros (approved):
Aside from one suggestion, looks great to me, thanks !

jzuhone (approved):

PR #3159, opened 2021-03-30
add a pytest-compatible version of requires_module

neutrinoceros:
## PR Summary

This function is meant as a drop-in replacement for the existing `required_module` decorator.
The end goal is to provide a pytest-compatible alternative (and eventual replacement when nose is dropped).
While I'm at it, and since it's relevant in the context of #3089 (containing the initial implementation), I added support for
an arbitrary number of arguments as an alternative to decorator stacking.

It is under testing in #3089 at the time of writing, and I confirm that it works as intended locally.

### DIFF
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -889,6 +889,47 @@ def requires_module(module):
         return ftrue
 
 
+def requires_module_pytest(*module_names):
+    """
+    This is a replacement for yt.testing.requires_module that's
+    compatible with pytest, and accepts an arbitrary number of requirements to
+    avoid stacking decorators
+
+    Important: this is meant to decorate test functions only, it won't work as a
+    decorator to fixture functions.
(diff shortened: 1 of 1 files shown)


yt/testing.py
```diff
@@ -889,6 +889,47 @@ def true_wrapper(*args, **kwargs):
...
+        required_modules = {
+            name: getattr(odi, f"_{name}")._module for name in module_names
+        }
+        missing = [
+            name
+            for name, mod in required_modules.items()
+            if isinstance(mod, odi.NotAModule)
+        ]
+
+        # note that order between these two decorators matter
```
neutrinoceros, line 919:
```suggestion
        # note that order between these two decorators matters
```

matthewturk (approved):

munkm (approved):

PR #3160, opened 2021-04-01
[bugfix] Fix units of potential

cphyc:
## PR Summary

Until recently, the potential dumped by RAMSES was assumed to be in energy units, whereas it should be in specific energy units.

## PR Checklist

### DIFF
--- a/yt/frontends/ramses/fields.py
+++ b/yt/frontends/ramses/fields.py
@@ -91,7 +91,7 @@ class RAMSESFieldInfo(FieldInfoContainer):
         ("x-acceleration", (ra_units, ["acceleration_x"], None)),
         ("y-acceleration", (ra_units, ["acceleration_y"], None)),
         ("z-acceleration", (ra_units, ["acceleration_z"], None)),
-        ("Potential", (specific_ener_units, ["potential"], None)),
+        ("Specific-Potential", (specific_ener_units, ["specific_potential"], None)),
         ("B_x_left", (b_units, ["magnetic_field_x_left"], None)),
         ("B_x_right", (b_units, ["magnetic_field_x_right"], None)),
         ("B_y_left", (b_units, ["magnetic_field_y_left"], None)),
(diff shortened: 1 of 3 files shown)


cphyc:
@jzuhone I think I followed the convention introduced in #3059, do you concur?

munkm (approved):

jzuhone:
@cphyc so there are a couple of things here. First, the RAMSES field name in yt should follow the convention that the RAMSES uses. So if RAMSES calls it `"Potential"` (with a capital P), the name should not be changed. In the case of the frontend field type, we always just call a field what it's called by the code. 

On top of this, the physics convention for gravitational potential is to call the specific energy field "potential" and the energy field "potential energy", so I think that we should leave it as `"potential"` and `"potential_energy"` (or even `"gravitational_potential"` and `"gravitational_potential_energy"`, but this gets long).

jzuhone:
@cphyc so there are a couple of things here. First, the RAMSES field name in yt should follow the convention that the RAMSES uses. So if RAMSES calls it `"Potential"` (with a capital P), the name should not be changed. In the case of the frontend field type, we always just call a field what it's called by the code. 

On top of this, the physics convention for gravitational potential is to call the specific energy field "potential" and the energy field "potential energy", so I think that we should leave it as `"potential"` and `"potential_energy"` (or even `"gravitational_potential"` and , but this gets long)

yt/frontends/ramses/fields.py
```diff
@@ -90,7 +91,7 @@ class RAMSESFieldInfo(FieldInfoContainer):
         ("x-acceleration", (ra_units, ["acceleration_x"], None)),
         ("y-acceleration", (ra_units, ["acceleration_y"], None)),
         ("z-acceleration", (ra_units, ["acceleration_z"], None)),
-        ("Potential", (ener_units, ["potential"], None)),
```
jzuhone, line 93: Per my comment, I think these should stay as `"Potential"` and `"potential"`.

yt/frontends/ramses/fields.py
```diff
@@ -181,6 +182,21 @@ def _temperature(field, data):
         if ("gas", "magnetic_field_x_left") in self:
             self.create_magnetic_fields()
 
+        # Potential field
+        if ("gravity", "Specific-Potential") in self:
+            self.create_gravity_fields()
+
+    def create_gravity_fields(self):
+        def potential(field, data):
```
jzuhone, line 190: And here the energy field should be `"potential_energy"`.

cphyc:
> @cphyc so there are a couple of things here. First, the RAMSES field name in yt should follow the convention that the RAMSES uses. So if RAMSES calls it `"Potential"` (with a capital P), the name should not be changed. In the case of the frontend field type, we always just call a field what it's called by the code.

Internally, RAMSES uses `phi` for the potential and `f` for the gravitational acceleration (it is a multidimensional array), together grouped under the name of `poisson`. Would you suggest I perform the following renaming?
- `("gravity", "Potential")` → `("poisson", "phi")`
- `("gravity", "x-acceleration")` → `("poisson", "f_x")`


> On top of this, the physics convention for gravitational potential is to call the specific energy field "potential" and the energy field "potential energy", so I think that we should leave it as `"potential"` and `"potential_energy"` (or even `"gravitational_potential"` and `"gravitational_potential_energy"`, but this gets long).

Fair enough, I have fixed the names!

jzuhone (approved):

Issue #2890, opened 2020-08-28
ProjectionPlot throws a MPL error when attempting to save()


chummels:
### Bug report

**Bug summary**

`ProjectionPlot` throwing an Matplotlib error on yt sample dataset, when it never did in the past.  This is part of a script in the standard tests used in Trident, which has long worked OK.  Note, that this simple example doesn't require trident at all in order to run.  Perhaps this is related to the transition to the most recent version of matplotlib?

**Code for reproduction**

```python
import yt
fn = 'FIRE_M12i_ref11/snapshot_600.hdf5'
ds = yt.load(fn)
p = yt.ProjectionPlot(ds, 'x', ('gas', 'density'))
p.save('projection.png')
```

**Actual outcome**

```
Traceback (most recent call last):
  File "gizmo_script.py", line 5, in <module>
    p.save('projection.png')
  File "/Users/chummels/src/yt/yt/visualization/plot_container.py", line 79, in newfunc
    rv = f(*args, **kwargs)
  File "/Users/chummels/src/yt/yt/visualization/plot_container.py", line 505, in save
    names.append(v.save(name, mpl_kwargs))
  File "/Users/chummels/src/yt/yt/visualization/base_plot_types.py", line 162, in save
    canvas.print_figure(name, **mpl_kwargs)
  File "/Users/chummels/src/miniconda3/lib/python3.7/site-packages/matplotlib/backend_bases.py", line 2217, in print_figure
    **kwargs)
  File "/Users/chummels/src/miniconda3/lib/python3.7/site-packages/matplotlib/backend_bases.py", line 1639, in wrapper
    return func(*args, **kwargs)
  File "/Users/chummels/src/miniconda3/lib/python3.7/site-packages/matplotlib/backends/backend_agg.py", line 509, in print_png
    FigureCanvasAgg.draw(self)
  File "/Users/chummels/src/miniconda3/lib/python3.7/site-packages/matplotlib/backends/backend_agg.py", line 407, in draw
    self.figure.draw(self.renderer)
  File "/Users/chummels/src/miniconda3/lib/python3.7/site-packages/matplotlib/artist.py", line 41, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/Users/chummels/src/miniconda3/lib/python3.7/site-packages/matplotlib/figure.py", line 1864, in draw
    renderer, self, artists, self.suppressComposite)
[...]

PR #3161, opened 2021-04-01
[bugfix] Fix log plots when using matplotilb 3.3.x if there are zeros in the plot

jzuhone:
## PR Summary

This PR fixes issue #2890, following a suggestion by @chrishavlin, which uses symlog scaling if there are zeros in the plot, avoiding a thrown exception by matplotlib if log-scaling is requested. 

I also updated the log messages to be more specific about why symlog scaling is being used.

## PR Checklist

### DIFF
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -970,7 +970,7 @@ class PWViewerMPL(PlotWindow):
                     # a fix is in -- see PR #3161 and linked issue.
                     cutoff_sigdigs = 15
                     if (
-                        np.log10(np.nanmax(image[np.isfinite(image)]]))
+                        np.log10(np.nanmax(image[np.isfinite(image)]))
                         - np.log10(np.nanmin(image[image > 0]))
                         > cutoff_sigdigs
                     ):
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -953,13 +953,32 @@ class PWViewerMPL(PlotWindow):
                     )
                 elif not np.any(np.isfinite(image)):
                     msg = f"Plot image for field {f} is filled with NaNs."
-                elif np.nanmax(image) > 0.0 and np.nanmin(image) <= 0:
+                elif np.nanmax(image) > 0.0 and np.nanmin(image) < 0:
                     msg = (
                         "Plot image for field %s has both positive "
                         "and negative/zero values. Min = %f, Max = %f."
                         % (f, np.nanmin(image), np.nanmax(image))
                     )
                     use_symlog = True
+                elif np.nanmax(image) > 0.0 and np.nanmin(image) == 0:
+                    # normally, a LogNorm scaling would still be OK here because
+                    # LogNorm will mask 0 values when calculating vmin. But
+                    # due to a bug in matplotlib's imshow, if the data range
+                    # spans many orders of magnitude while containing zero points
+                    # vmin can get rescaled to 0, resulting in an error when the image
+                    # gets drawn. So here we switch to symlog to avoid that until
+                    # a fix is in -- see PR #3161 and linked issue.
+                    cutoff_sigdigs = 15
+                    if (
+                        np.log10(np.nanmax(image))
+                        - np.log10(np.nanmin(image[image > 0]))
+                        > cutoff_sigdigs
+                    ):
+                        msg = (
+                            "Plot image for field %s has both positive "
+                            "and zero values with large dynamic range." % (f,)
+                        )
+                        use_symlog = True
                 if msg is not None:
                     mylog.warning(msg)
                     if use_symlog:
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -970,7 +970,7 @@ class PWViewerMPL(PlotWindow):
                     # a fix is in -- see PR #3161 and linked issue.
                     cutoff_sigdigs = 15
                     if (
-                        np.log10(np.nanmax(image))
+                        np.log10(np.nanmax(image[np.isfinite(image)]]))
                         - np.log10(np.nanmin(image[image > 0]))
                         > cutoff_sigdigs
                     ):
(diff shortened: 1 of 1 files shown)


chrishavlin:
interesting! so it IS failing some image comparison tests... here's a simplified code snippet that will show why the image comparison is failing:

```python
WT = "WindTunnel/windtunnel_4lev_hdf5_plt_cnt_0030"
plot_field = "density"
ds = yt.load(WT)
slc = yt.SlicePlot(ds,"z",plot_field)
```

So on main, this code runs and generates:

whatthewhat_Slice_z_density

But it turns out the blank white area of that plot are actually zero values, so with the symlog check change those 0 values get filled (resulting in the failed image tests):

whatthewhat_part2_Slice_z_density

Which makes me suspect that there's something else going on to cause #2890  to fail?

jzuhone:
@chrishavlin what's the version of matplotlib you're using?

matthewturk:
Did we change initialization of the background buffer from 0's to NaNs, or
vice-versa, in any of these recent PRs?

chrishavlin:
Oh good call, I used  matplotlib 3.3.4 there

chrishavlin:
Oh, but the #2890 errors on 3.3.4 while the `WT` example does not error.

jzuhone:
@matthewturk @chrishavlin I note that the failure in #2890 is consistently with SPH datasets whereas this is grid-based here

chrishavlin:
Good point, so it's not a straight comparison necessarily. And slicing the dataset in #2890 works fine (no matplotlib errors from the 0 values).

chrishavlin:
So I actually think we're seeing a rounding error bug in matplotlib here... moved discussion back to #2890

yt/visualization/plot_window.py
```diff
@@ -953,19 +953,20 @@ def _setup_plots(self):
...
                     msg = f"Plot image for field {f} is filled with NaNs."
-                elif np.nanmax(image) > 0.0 and np.nanmin(image) < 0:
+                elif np.nanmax(image) > 0.0 and np.nanmin(image) <= 0:
                     msg = (
                         "Plot image for field %s has both positive "
-                        "and negative values. Min = %f, Max = %f."
+                        "and negative/zero values. Min = %f, Max = %f."
                         % (f, np.nanmin(image), np.nanmax(image))
                     )
                     use_symlog = True
```
chrishavlin, line 962: Now that I understand the underlying problem here...it may be better to keep existing symlog check at line 956 but then add a new check to switch to symlog if there is a zero value and if the data range spans a large order of magnitude. e.g., this seems to work ok: 

```python
                elif np.nanmax(image) > 0.0 and np.nanmin(image) == 0:
                    # normally, a LogNorm scaling would still be OK here because
                    # LogNorm will mask 0 values when calculating vmin. But
                    # due to a bug in matplotlib's imshow, if the data range
                    # spans many orders of magnitude while containing zero points
                    # vmin can get rescaled to 0, resulting in an error when the image
                    # gets drawn. So here we switch to symlog to avoid that until
                    # a fix is in -- see PR #3161 and linked issue.  
                    cutoff_sigdigs = 15   
                    if np.log10(np.nanmax(image)) - np.log10(np.nanmin(image[image>0])) > cutoff_sigdigs:
                        msg = (
                            "Plot image for field %s has both positive "
                            "and zero values with large dynamic range."
                        )
                        use_symlog = True
```

This should pass the image comparisons. The somewhat arbitrary `cutoff_sigdigs` isn't ideal... but maybe there's a smarter way to do this?

jzuhone:
So that latest change did fix everything except that one test, oddly.

chrishavlin:
Interesting. I'll take a look at the failing test.

yt/visualization/plot_window.py
```diff
@@ -956,16 +956,36 @@ def _setup_plots(self):
...
+                    # normally, a LogNorm scaling would still be OK here because
+                    # LogNorm will mask 0 values when calculating vmin. But
+                    # due to a bug in matplotlib's imshow, if the data range
+                    # spans many orders of magnitude while containing zero points
+                    # vmin can get rescaled to 0, resulting in an error when the image
+                    # gets drawn. So here we switch to symlog to avoid that until
+                    # a fix is in -- see PR #3161 and linked issue.
+                    cutoff_sigdigs = 15
+                    if (
+                        np.log10(np.nanmax(image))
```
chrishavlin, line 973: that last failing test is due to `image` having `inf` values, so using the finite max should work here: 

```suggestion
                        np.log10(np.nanmax(image[np.isfinite(image)]]))
```

chrishavlin:
oops -- my suggestion had an extra `]` in there.

jzuhone:
This is ready to go!!

munkm:
I'll review after my next meeting finishes!

chrishavlin:
woo! After reviews I can also open a more general issue than #2890 so this fix isn't forgotten. Once the rounding issue is fixed on the matplotlib side the fix here could be removed (or adjusted to have a matplotlib version check for a while). I also am not super happy with the hard coded parameter I added... but I don't see a simple fix without it.

neutrinoceros:
> Did we change initialization of the background buffer from 0's to NaNs, or
vice-versa, in any of these recent PRs?

@matthewturk this rung a tiny bell, I indeed changed the background buffer filler value here https://github.com/yt-project/yt/pull/2517

neutrinoceros:
@chrishavlin ok I'm done reading this. I could not find a link to a bug report on matplotlib's side, has it been done yet ?

chrishavlin:
https://github.com/matplotlib/matplotlib/issues/19748 is the main bug report, it links to this PR in progress: https://github.com/matplotlib/matplotlib/pull/18782 the PR introduces a new kwarg to use a different method for the rescaling that will be log-safe, so we'd have to add that to our matplotlib calls once it's in (presumably with a matplotlib version check for a while?).

The NaN vs 0 in the background buffer is good to know, but it it's not required for this bug to appear. The matplotlib bug report I submitted before I noticed the open one has some simpler examples showing weird behavior for large data ranges even without the zeros: https://github.com/matplotlib/matplotlib/issues/19851

munkm (approved):
So, I agree that this isn't an ideal solution and we should view it as temporary until a fix is implemented upstream. We should also open an issue on our own repo that links to this PR, the upstream issues/PRs, and will make sure we make changes here as soon as everything upstream is resolved.

PR #3163, opened 2021-04-03
remove mccabe from linting specs (already a direct dependency of flake8)

neutrinoceros:
## PR Summary

See https://github.com/PyCQA/flake8/blob/e0116d8e77fb04acc62095734ec49e779f24ee54/setup.cfg#L43

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,6 @@ repos:
     hooks:
     - id: flake8
       additional_dependencies: [
-        mccabe,
         flake8-bugbear>=20.3.2, # GH PR 2851
         flake8-logging-format
       ]


matthewturk (approved):
Looks OK to me -- unless we're directly calling mccabe, which I think we aren't, this should be fine.  It's probably best to only specify absolute direct deps.

neutrinoceros:
> ooks OK to me -- unless we're directly calling mccabe, which I think we aren't, this should be fine

to be cristal clear:
- this only affects the pre-commit config, in which we don't have a hoo for mccabe (and we wouldn't need one since it's embedded in flake8). Hooks installed via pre-commit don't interfere with tools installed locally. 
- it doesn't actually change anything because mccabe is a direct dep of flake8, so it will still be installed anyway.

This won't break *anyone*'s workflow, I promise. I'm just cleaning up a useless line, which _used_ have a purpose when mccabe was a plugin to flake8.

jzuhone (approved):

PR #3166, opened 2021-04-06
[bugfix] Fix field calculation for octrees

cphyc:
At the moment, the current fields cannot be computed for octree datasets (RAMSES & ART).
- [x] `("gas", "average_density")`
- [ ] `("gas", "shear_mach")`

# Steps to reproduce
```python
import yt

ds.r["gas", "shear_mach"]       # raises RuntimeError
ds.r["gas", "average_density"]  # raises YTFieldNotFound
```

## Expected behaviour
The script should be able to run till the end. Note that it may catch some `yt.fields.field_exceptions.NeedsParameter` exceptions.

```python
import yt
from tqdm import tqdm
import numpy as np

# ds = yt.load_sample("enzo_tiny_cosmology")
ad = ds.all_data()

for ftype, fname in tqdm(np.random.permutation(ds.derived_field_list)):
    if "cutting_plane" in fname:  # need to skip cutting planes as they require a cutting plane!
        continue
    if fname in ("shear_mach", "average_density"):  # these trigger catastrophic failures, skip them
        continue
    yt.mylog.debug("Computing field %s", (ftype, fname))
    try:
        ad[ftype, fname]
    except:
        yt.mylog.error("Got exception for field (%s, %s)", ftype, fname, exc_info=True)
```

### DIFF
--- a/yt/fields/vector_operations.py
+++ b/yt/fields/vector_operations.py
@@ -609,13 +609,19 @@ def create_averaged_field(
     validators += [ValidateSpatial(1, [(ftype, basename)])]
 
     def _averaged_field(field, data):
-        nx, ny, nz = data[(ftype, basename)].shape
+        def atleast_4d(array):
+            if array.ndim == 3:
+                return array[..., None]
+            else:
+                return array
+
+        nx, ny, nz, ngrids = atleast_4d(data[(ftype, basename)]).shape
         new_field = data.ds.arr(
-            np.zeros((nx - 2, ny - 2, nz - 2), dtype=np.float64),
+            np.zeros((nx - 2, ny - 2, nz - 2, ngrids), dtype=np.float64),
             (just_one(data[(ftype, basename)]) * just_one(data[(ftype, weight)])).units,
         )
         weight_field = data.ds.arr(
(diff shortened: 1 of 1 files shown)


cphyc:
The second item turns out to be quite involved. After some digging, it is related to the way the grid level is computed for octrees. I cannot find a good way of fixing it at the moment, so I'd be happy to have this PR merged to fix item 1 and I'll open an issue for solving item 2.

munkm:
That sounds good! Since i'll be writing the changelog soon -- this specifically fixes field calculation for `average_density` on octrees, right?

munkm (approved):
If possible, I think we should add a test for a field that was fixed in this PR.

cphyc:
> That sounds good! Since i'll be writing the changelog soon -- this specifically fixes field calculation for `average_density` on octrees, right? Are there any other fields that this could apply to?

In principle the same approach could be reused for other fields, but I couldn't find any for which it was required. I'll add a test.

munkm:
Do you want me to merge and open an issue for the test, or wait until one is added here to merge?

cphyc:
Opening an issue is probably best!

8 Apr 2021 8:06:10 pm Madicken Munk ***@***.***>:

PR #3169, opened 2021-04-07
fix a broken import statement in requires_module_pytest

neutrinoceros:
## PR Summary

quick hotfix for #3159
I was too bold in cutting this function out of #3089 initially, now it's actually usable.

### DIFF
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -902,7 +902,7 @@ def requires_module_pytest(*module_names):
 
     So that it can be later renamed to `requires_module`.
     """
-    from pytest.mark import skipif
+    import pytest
 
     from yt.utilities import on_demand_imports as odi
 
@@ -917,7 +917,7 @@ def requires_module_pytest(*module_names):
         ]
 
         # note that order between these two decorators matters
-        @skipif(
+        @pytest.mark.skipif(
             missing,
             reason=f"missing requirement(s): {', '.join(missing)}",
         )


matthewturk (approved):

munkm (commented):
Sorry, why do you have to import the full pytest? Was there an error with just importing the function?

neutrinoceros:
Yes, pytest.mark cannot be imported from.

munkm:
Is there a reason we can't use `pytest.importorskip`?

munkm (approved):
I'm fine with this as is, but I'm a bit unhappy about the full import of pytest in a decorator.

neutrinoceros:
me too, that's why I naively refactored it originally.

munkm:
FWIW importorskip can be imported with from and it will skip a test if the module isn't returned.

neutrinoceros:
Interesting, I didn't know it existed. However it works very differently. In particular it will only report one missing module at a time, while my implementation of `requires_module_pytest` reports every missing package, so I don't think it's worth using `pytest.importorskip` instead.

munkm:
Yeah so what I was suggesting was not using pytest.importorskip in place of the requires_module_pytest decorator, but rather using it in the decorator you've written to avoid having to import the full pytest in it.

neutrinoceros:
That's what I understood. I don't know _if_ the refactor you're suggesting is even doable, it seems to me that it wouldn't be able to report more than one missing package at a time.
I'd like this hot fix to go in as is because it can't be worse than the broken version we have so far. It'll always be possible to explore better implementations later.

PR #3171, opened 2021-04-08
[bugfix] Fix issues with FITS images

jzuhone:
## PR Summary

There were bugs in propagating units correctly in `FITSImageData` instances if they were created from other instances using `pop` or `combine_images`. It was also the case that the code would attempt to make the units of a FITS image made from a cosmological simulation in terms of comoving quantities, little h, and the scale factor. For various reasons, this does not work well with AstroPy, so this code also ensures that "proper" units are used in the FITS files that are produced. 

This PR fixes those issues, and also adds the hubble parameter and the current redshift to FITS images made from cosmological simulations so that the cosmological units can be re-created if necessary. 

## PR Checklist


munkm (approved):
Looks good! Thanks for fixing this @jzuhone

PR #3172, opened 2021-04-08
maint: pin numpy maximal version

neutrinoceros:
## PR Summary

Considering that each new feature release from numpy breaks our tests, I think it's worth explicitly pinning the current latest version as the max supported one.
- it will make _our_ next release reliable for much longer
- we won't need to apply this patch _after_ numpy 1.21 is released, breaks our CI and we need a hotfix.

I'll open a companion PR in which I try to deal with some of the _existing_ deprecation warnings from numpy that are found in our logs. Note that numpy's deprecation warnings usually state that something is going to produce an error _in the future_, but very rarely (if ever ?) do they explicitly contain a date/release number. This is the reason I propose to pin the _current_ version as latest supported instead of, say 1.22.


neutrinoceros:
This conflicts with #3236 and doesn't seem to raise much interest. Closing.

PR #3173, opened 2021-04-08
Fix some numpy deprecation warnings

neutrinoceros:
## PR Summary

As I was searching logs for an explicit mention of a future version of numpy that will break the current state of yt, I was able to fix _some_ of the existing warnings (definitely not all, some are much harder to track down than others).

### DIFF
--- a/yt/data_objects/tests/test_points.py
+++ b/yt/data_objects/tests/test_points.py
@@ -62,7 +62,7 @@ def test_fast_find_field_values_at_points():
     # sample 100 particles
     nparticles = 100
     ppos = ad["all", "particle_position"]
-    ppos = ppos[np.random.random_integers(0, len(ppos) - 1, size=nparticles)]
+    ppos = ppos[np.random.randint(low=0, high=len(ppos), size=nparticles)]
 
     ppos_den = ds.find_field_values_at_points("density", ppos)
     ppos_vel = ds.find_field_values_at_points("velocity_x", ppos)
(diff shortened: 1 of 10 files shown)


cphyc (approved):

yt/data_objects/tests/test_points.py
```diff
@@ -62,7 +62,7 @@ def test_fast_find_field_values_at_points():
     # sample 100 particles
     nparticles = 100
     ppos = ad["all", "particle_position"]
-    ppos = ppos[np.random.random_integers(0, len(ppos) - 1, size=nparticles)]
+    ppos = ppos[np.random.randint(low=0, high=len(ppos), size=nparticles)]
```
matthewturk, line 65: Just verifying: `high` is not inclusive, right?

yt/testing.py
```diff
@@ -729,8 +729,7 @@ def add_noise_fields(ds):
 
     def _binary_noise(field, data):
         """random binary data"""
-        res = prng.random_integers(0, 1, data.size).astype("float64")
-        return res
+        return prng.randint(low=0, high=2, size=data.size).astype("float64")
```
matthewturk, line 732: I see from *this* one that it isn't!  Never mind the above.

matthewturk (approved):
lgtm

yt/data_objects/tests/test_points.py
neutrinoceros: yes, that's the important differences between the two apis, and it's advertised in the deprecation warning itself :)

PR #3175, opened 2021-04-08
supporting MPL 3.4 : stop registering colormaps from cmocean

neutrinoceros:
## PR Summary

The problem is discussed in #3165
Right now, using an env with yt's bleeding edge and
```
cmocean
matplotlib==3.4
```
breaks everything because of a naming collision that matplotlib used to tolerate but doesn't any more
```shell
$ python -c "import yt"
...
ValueError: Trying to re-register the builtin cmap 'gray'.
```

The solution is to stop registering cmocean's colormaps unprefixed.
Sadly this is a backward breaking change but it's necessary in order to make yt compatible with matplotlib 3.4, so I believe it's worth the cost.

Note that yt can still be used in combination with cmocean or similar projects (such as cmasher), at the discretion of the user.

This used to work (with cmocean installed)
```python
import matplotlib
import yt

ds = yt.testing.fake_amr_ds()
p = yt.SlicePlot(ds, "z", "Density")
p.set_cmap("all", "balance")
```

This is how it should work instead
```python
import matplotlib
import yt
import cmocean

ds = yt.testing.fake_amr_ds()
p = yt.SlicePlot(ds, "z", "Density")
p.set_cmap("all", "cmo.balance")
```
note that this second version works with any existing version of yt.

### DIFF
--- a/doc/source/visualizing/colormaps/index.rst
+++ b/doc/source/visualizing/colormaps/index.rst
@@ -50,24 +50,6 @@ not be interpolated, and can be useful for creating
 colorblind/printer/grayscale-friendly plots. For more information, visit
 `http://colorbrewer2.org <http://colorbrewer2.org>`_.
 
-.. _cmocean-cmaps:
-
-Colormaps from cmocean
-~~~~~~~~~~~~~~~~~~~~~~
-
-In addition to ``palettable``, yt will also import colormaps defined in the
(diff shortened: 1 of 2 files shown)


matthewturk:
I think this is necessary.

matthewturk (approved):

cphyc (approved):

PR #3176, opened 2021-04-08
Hotfix image filename validation

neutrinoceros:
## PR Summary

The following script has two bugs
```python
from yt import SlicePlot

ds = fake_amr_ds()
p = SlicePlot(ds, "z", "Density")
p.save("1.2.3")
```

1 - the logger contains a stupid entry (not meaning offense to anyone but myself, I authored it)
```
yt : [WARNING  ] ... Ignoring supplied suffix '.png' in favour of '.png'
```
2 - the resulting file is named `1.2.png`, while it should *clearly* be `1.2.3.png` (discovered this while trying to embed matplotlib version numbers in file names...)

This fixes both these problems and provides better testing (yay, pytest !) for the culprit function.


yt/visualization/_commons.py
```diff
@@ -13,28 +14,39 @@
...
-
     name, psuffix = os.path.splitext(filename)
     if psuffix in SUPPORTED_FORMATS:
-        if suffix:
+        if suffix is not None:
+            suffix = normalize_extension_string(suffix)
+        if suffix in SUPPORTED_FORMATS and suffix != psuffix:
             mylog.warning(
-                "Ignoring supplied suffix '%s' in favour of '%s'", suffix, psuffix
+                "Received two valid image format '%s' and '%s'. "
```
cphyc, line 33:
```suggestion
                "Received two valid image formats '%s' and '%s'. "
```

yt/visualization/_commons.py
```diff
@@ -13,28 +14,39 @@
...
+            return f"{name}{suffix}"
         return str(filename)
-    elif psuffix:
-        mylog.warning(
-            "Found unsupported suffix '%s', using '%s' instead.", psuffix, suffix
-        )
 
-    return f"{name}{suffix}"
+    if suffix is None:
+        suffix = ".png"
```
cphyc, line 42: Why not?
```suggestion
        suffix = ytcfg.get("yt", "default_image_format")
```

cphyc (approved):
LGTM!

(sorry for closing/reopening, that was a mistake!)

yt/visualization/_commons.py
neutrinoceros: Good point, I don't think this always happens as close to the surface at it should.

yt/visualization/_commons.py
neutrinoceros: I'm ambivalent. I don't like having a parameter with "default" in its name, makes it a bit hard to know what we're talking about. That being said I like the idea of parametrising a "preferred" output format (though that name isn't perfect either). I'd like to think about this a bit more and maybe do it in a separate PR.

matthewturk (approved):

PR #3181, opened 2021-04-10
Fix and clean strings

cphyc:
This removes all occurrences of `x = "foo"  "bar"` by `x = "foobar"` and applies other cosmetic changes. In particular, some of our messages are missing a space at the joint between string parts, e.g.
```
x = "Houston, we have" "a problem"  # == "Houston, we havea problem"
# should be
x = "Houston, we have a problem"
```
These aren't "bugs" _per se_ as they shouldn't prevent yt from functioning properly on the machine side, but should be corrected nonetheless.

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -2487,7 +2487,7 @@ class TimestampCallback(PlotCallback):
                     toffset = plot.ds.quan(self.time_offset, self.time_unit)
                 elif not isinstance(self.time_offset, YTQuantity):
                     raise RuntimeError(
-                        "'time_offset' must be a float, tuple, or" "YTQuantity!"
+                        "'time_offset' must be a float, tuple, orYTQuantity!"
                     )
                 t -= toffset.in_units(self.time_unit)
             try:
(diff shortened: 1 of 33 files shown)


yt/visualization/plot_modifications.py
```diff
@@ -2487,7 +2487,7 @@ def __call__(self, plot):
                     toffset = plot.ds.quan(self.time_offset, self.time_unit)
                 elif not isinstance(self.time_offset, YTQuantity):
                     raise RuntimeError(
-                        "'time_offset' must be a float, tuple, or" "YTQuantity!"
+                        "'time_offset' must be a float, tuple, orYTQuantity!"
```
cphyc, line 2490:
```suggestion
                        "'time_offset' must be a float, tuple, or YTQuantity!"
```

yt/visualization/plot_window.py
```diff
@@ -932,9 +932,7 @@ def _setup_origin(self):
         y_in_bounds = yc >= yllim and yc <= yrlim
 
         if not x_in_bounds and not y_in_bounds:
-            msg = (
-                "origin inputs not in bounds of specified coordinate sytem" + "domain."
-            )
+            msg = "origin inputs not in bounds of specified coordinate sytemdomain."
```
cphyc, line 935:
```suggestion
            msg = "origin inputs not in bounds of specified coordinate system domain."
```

neutrinoceros (approved):
Zenny. Thanks for doing this ! Note that kn one or two cases those are technically bug fixes ;)

neutrinoceros (approved):
more bugfixes, more approval

PR #3193, opened 2021-04-13
Use io._vector_fields to get a field shape. Fixes #3191

### LINKED ISSUE #3191: Field detection hardcodes vector fields
As discovered by @Xarthisius while working with @jcoughlin11 , the current field detection method hardcodes the names of possible vector fields. This leads to not detecting aliases for common names like `positions`.

It should be possible to check the vector field list, since that ought to be available at runtime through the `io` object.

Xarthisius:
I need to check if that doesn't introduce any unnecessary overhead, but on the bright side it removes a lot of hardcoded information.

### DIFF
--- a/yt/fields/field_detector.py
+++ b/yt/fields/field_detector.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 import numpy as np
 
 from yt.units.yt_array import YTArray
+from yt.utilities.io_handler import io_registry
 
 from .field_exceptions import NeedsGridType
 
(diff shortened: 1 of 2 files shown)


matthewturk (approved):
definite improvement

yt/fields/field_detector.py
```diff
@@ -146,24 +147,17 @@ def __missing__(self, item):
                     self[item] = vv.ravel()
                 return self[item]
         elif finfo is not None and finfo.sampling_type == "particle":
-            if (
-                "particle_position" in (item, item[1])
```
munkm, line 150: Is it ok that we're not checking `particle_position`, `particle_velocity`, or `particle_magnetic_field` anymore with these updates (unless I'm missing something?)?

munkm (approved):
Generally this looks good! I have one question and am happy to merge once it's answered!

yt/fields/field_detector.py
Xarthisius: tbh I don't know. However, no test fails and I couldn't identify the place where it was used (if it was used).

Xarthisius: I digged a little. Normally those fields are aliases to some frontend specifc fields (like `particle_position` -> `Coordinates` for Gadget etc.). They'd only play a role if a frontend had a native field named like that, but then it should be defining it in its specific `IOHandler._vector_fields`.

jcoughlin11:
@Xarthisius @matthewturk @munkm I'm just wondering if this is ready to be merged? I ask because it would help a ton for my work on the qmc frontend. No rush, though!

Xarthisius:
As far as I'm concerned this is ready to go.

PR #3093, opened 2021-02-24
add a test to check that default data dirs are not valid dirnames

neutrinoceros:
## PR Summary

In working on #3089 I need to make assertions on the existence or absence of data files.
I am not convinced that "/does/not/exist" (the current default value for `"test_data_dir"`) is indeed an invalid directory name for all platforms (in particular on windows), so I'm adding a test to check this assertion. If the name turns out valid for windows, I'll change the default value to something else (probably `"\0"`)

PR #3195, opened 2021-04-13
hotfix: yt.load_sample + yt_data_dir unset

neutrinoceros:
## PR Summary

This is an attempt at fixing a bug reported by @matthewturk on Slack.
I think the bug is plateform dependent and I don't have time to properly test it just yet, so I'm opening early in hopes that @matthewturk can test it locally first.

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1338,7 +1338,8 @@ def load_sample(fn, progressbar: bool = True, timeout=None, **kwargs):
 
     try:
         save_dir = _get_test_data_dir_path()
-    except FileNotFoundError:
+        assert save_dir.is_dir()
+    except (OSError, AssertionError):
         mylog.warning(
             "yt test data directory is not properly set up. "
             "Data will be saved to the current work directory instead."


matthewturk (approved):
Seems good, although I'm somewhat partial to the idea of saving it in the `XDG_CACHE` directory.

neutrinoceros:
Where is this done ? I feel like I'm missing something

matthewturk:
I think pooch will do it by default.

matthewturk:
Ah, so what I mean is that pooch uses the `userdirs` package to come up with default locations.  So, if left unspecified, the download location is set by `XDG_CACHE` or the default-for-that-system value for `XDG_CACHE`.  This way it is still a user-specific, systemwide location that will be identical across runs independent of working dir.  (i.e., if I cd to a subdir and re-run `load_sample`, it won't re-download)

neutrinoceros:
I honestly can't tell which behaviour would feel intuitive or "correct" to most people, but I agree the change should be considered. Since you approved this I take it that you're ok with this patch going in as in since it's just fixing the current intended behaviour ?

matthewturk:
I am fine with it going in as-is to fix the current behavior.

neutrinoceros:
A slightly different approach would be to set `yt_data_dir`'s default value as XDG_CACHE, WDYT ?

cphyc (approved):
This looks good to me!

neutrinoceros:
@matthewturk I also have an PR to change the default dir names here #3093
You have a good point that `XDG_CACHE` is likely a solid candidate default for yt_data_dir, so maybe we ought to open the discussion and this how well it applies to the rest of them too ?

PR #3198, opened 2021-04-13
Add a heuristic for resetting index_order2

### LINKED ISSUE #3192: Memory and CPU usage with small Gadget data sets
After the change in #2575, we run into some real problems with some smaller datasets, where essentially the entire dataset gets expanded up to the maximum bitmap level.  This is problematic for datasets like `Gadget3-snap-format2`, which have multiple data files, very few particles, but big smoothing lengths that cover many low-order cells.

Default behavior for this script:

```python
import yt

with yt.memory_checker():
    ds = yt.load_sample("Gadget3-snap-format2")
    ds.index
```

produces a ridiculous amount of memory usage and takes a hubble time.

This can be essentially completely remedied by setting the `index_order` to have a *lower* order for the second component.  I'm tempted to consider a heuristic setting of the index order to be the appropriate fix.

matthewturk:
## PR Summary

Fixes #3192

This PR accomplishes two specific goals.  The first is that it enables, in the case of smoothed particles, an on-the-fly adjustment of the `index_order2` value.

What happens in many cases with the default `index_order2` value is that we have some gas particles that have very large smoothing lengths; this often happens with small, isolated datasets -- particles out at the edges have big `hsml` values.  It's very expensive to manage these at high `index_order2` values, since that involves lots of setting things to 1, lots of map accesses, etc.  It also is not broadly *useful* since even if they collide at coarse levels, they are so big that the degree of collision is pretty minimal.

This allows those cases to dynamically update the `index_order2` value to be representative of the max `hsml`.  Whether this should be max, mean, etc, is not entirely clear to me, and will likely vary based on the dataset.  I'm inclined to think we need some real-world checks, or a much better heuristic in place.  But this solves out problem with the one specific dataset.

The other thing it does is change the default value for our indexing from (7, 5) to (5, 3).  This means that at the coarse level we'll be applying an index of `2**5` (32) instead of `2**7` (128) and at the fine level we'll be applying `2**8` (256) instead of `2**12` (4096).  This change is definitely up for discussion, as I think perhaps I've over-corrected here in favor of small datasets.

## PR Checklist

### DIFF
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -210,7 +210,8 @@ class ParticleIndex(Index):
                     hsml = self.io._get_smoothing_length(
                         data_file, pos.dtype, pos.shape
                     )
-                    max_hsml = max(max_hsml, hsml.max())
+                    if hsml is not None and hsml.size > 0.0:
+                        max_hsml = max(max_hsml, hsml.max())
                 else:
                     hsml = None
                 self.regions._coarse_index_data_file(pos, hsml, data_file.file_id)
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -217,7 +217,7 @@ class ParticleIndex(Index):
             self.regions._set_coarse_index_data_file(data_file.file_id)
         pb.finish()
         self.regions.find_collisions_coarse()
-        if max_hsml > 0.0:
+        if max_hsml > 0.0 and len(self.data_files) > 1:
             new_order2 = self.regions.update_mi2(max_hsml)
             mylog.info(
                 "Updating index_order2 from %s to %s", ds.index_order[1], new_order2
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -163,12 +163,16 @@ class ParticleIndex(Index):
         )
 
         # Load Morton index from file if provided
-        if getattr(ds, "index_filename", None) is None:
-            fname = ds.parameter_filename + ".index{}_{}.ewah".format(
-                self.regions.index_order1, self.regions.index_order2
-            )
-        else:
-            fname = ds.index_filename
+        def _current_fname():
+            if getattr(ds, "index_filename", None) is None:
+                fname = ds.parameter_filename + ".index{}_{}.ewah".format(
+                    self.regions.index_order1, self.regions.index_order2
+                )
+            else:
+                fname = ds.index_filename
+            return fname
+
+        fname = _current_fname()
 
         dont_load = dont_cache and not hasattr(ds, "index_filename")
         try:
@@ -183,6 +187,8 @@ class ParticleIndex(Index):
             self.regions.reset_bitmasks()
             self._initialize_coarse_index()
             self._initialize_refined_index()
+            # We now update fname since index_order2 may have changed
+            fname = _current_fname()
             wdir = os.path.dirname(fname)
             if not dont_cache and os.access(wdir, os.W_OK):
                 # Sometimes os mis-reports whether a directory is writable,
@@ -195,6 +201,7 @@ class ParticleIndex(Index):
 
     def _initialize_coarse_index(self):
         pb = get_pbar("Initializing coarse index ", len(self.data_files))
+        max_hsml = 0.0
         for i, data_file in enumerate(self.data_files):
             pb.update(i)
             for ptype, pos in self.io._yield_coordinates(data_file):
@@ -203,12 +210,19 @@ class ParticleIndex(Index):
                     hsml = self.io._get_smoothing_length(
                         data_file, pos.dtype, pos.shape
                     )
+                    max_hsml = max(max_hsml, hsml.max())
                 else:
                     hsml = None
                 self.regions._coarse_index_data_file(pos, hsml, data_file.file_id)
             self.regions._set_coarse_index_data_file(data_file.file_id)
         pb.finish()
         self.regions.find_collisions_coarse()
+        if max_hsml > 0.0:
+            new_order2 = self.regions.update_mi2(max_hsml)
+            mylog.info(
+                "Updating index_order2 from %s to %s", ds.index_order[1], new_order2
+            )
+            self.ds.index_order = (self.ds.index_order[0], new_order2)
 
     def _initialize_refined_index(self):
         mask = self.regions.masks.sum(axis=1).astype("uint8")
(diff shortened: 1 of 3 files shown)


cphyc:
Seems you got jenkins stuck!

matthewturk:
Yeah, I think I made a mistake. Needs some more work.

matthewturk:
Interestingly, the change that's breaking `test_rays.py` is the change to the default index order, not the change in the refined index, which suggests there may be something going on with the coarse index, not the refined.

chummels (approved):
In terms of reviewing the code, I don't really know how to do so, since I have very little understanding of the chunking mechanisms in yt, and there is very little discussion of it in the documentation (here or elsewhere).  I don't think it's appropriate for this PR to held up requesting a ton more documentation on the chunking infrastructure, but I do think it would be helpful in the long run to have a narrative discussion of it somewhere.  Maybe the new yt paper is that location? 

As far as "testing" this PR, I can do that.  When I run this to see it's impact on loading up low-res particle-based datasets, I see large speedups.  On the `FIRE_M12i_ref11` sample dataset, which is a minimal dataset of 260MB from FIRE, I get a speedup from a 15 second load time to a 6 second load time (2.5x speedup).  When I run this on the low-res 1GB AREPO dataset that I tried last a few weeks ago, I get a speedup from a 640 second load time to a 57 second load time (11x speedup).  Notably, the FIRE dataset is a single hdf5 file, whereas the AREPO is in 8 hdf5 files.  

Just to see if this PR has any impact on subsequent image generation, after deleting the ewah files, I similarly ran the code to load a dataset and then make a gas density projection on it.  For both the FIRE and AREPO datasets, I see no significant change in image generation time and the images look the same.  

So I consider this a big win!  Thanks for working on this, @matthewturk .

matthewturk:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

brittonsmith:
Looks like there are a handful of failures with the gasoline dm-only dataset, but that's it.

matthewturk:
Yeah, I'm not totally sure why they're failing, but I do believe they may
be real.

chummels:
@yt-fido test this please

chummels:
I'm not sure why you're getting 4 test failures, all in: `yt.frontends.tipsy.tests.test_outputs.test_gasoline_dmonly` for particle mass and velocity fields.  But looking around, this is our only DM-only particle-based dataset, so maybe it's related to the fact that this data lacks smoothing lengths compared to other particle-based datasets?

matthewturk:
Could be! I haven't dug in yet on these specific errors.

yt/geometry/particle_geometry_handler.py
```diff
@@ -163,12 +163,16 @@ def _initialize_index(self):
...
-            fname = ds.parameter_filename + ".index{}_{}.ewah".format(
-                self.regions.index_order1, self.regions.index_order2
-            )
-        else:
-            fname = ds.index_filename
+        def _current_fname():
+            if getattr(ds, "index_filename", None) is None:
+                fname = ds.parameter_filename + ".index{}_{}.ewah".format(
+                    self.regions.index_order1, self.regions.index_order2
+                )
```
neutrinoceros, line 170:
```suggestion
                fname = "{}.index{}_{}.ewah".format(
                    ds.parameter_filename,
                    self.regions.index_order1,
                    self.regions.index_order2,
                )
```

yt/geometry/particle_geometry_handler.py
```diff
@@ -163,12 +163,16 @@ def _initialize_index(self):
...
+            else:
+                fname = ds.index_filename
+            return fname
```
neutrinoceros, line 173:
```suggestion
                return "{}.index{}_{}.ewah".format(
                    ds.parameter_filename,
                    self.regions.index_order1,
                    self.regions.index_order2,
                )
            else:
                return ds.index_filename
```

matthewturk:
I am unable to reproduce the errors that I remember seeing locally.  I'll see if they still show up on fido ...

matthewturk:
@yt-fido test this please

chummels:
I believe that all of the errors were confined to the `yt.frontends.tipsy.tests.test_outputs.test_gasoline_dmonly` test.  That's what we've seen the last two times that fido ran the tests.

matthewturk:
@chummels That's what I was unable to reproduce :(

matthewturk:
So, it hung, and died, and that makes me nervous ... and I've been encouraged to re-run!

matthewturk:
@yt-fido test this please

Xarthisius:
It's hanging on this:

```
import yt

BE_Gadget = "BigEndianGadgetBinary/BigEndianGadgetBinary"
ds = yt.load(BE_Gadget)
data = ds.all_data()
data["Halo", "Velocities"]
```

Normally this snippet takes <10s. With this PR it never finishes. The fact it tries to allocate 75.5Gb of RAM might have something to do with it...

Also can someone please explain to me why it's been marked as an answer test for past 3 years and yet we're not storing nor comparing anything in there? FTR:
https://github.com/yt-project/yt/blame/main/yt/frontends/gadget/tests/test_outputs.py#L123-L127

matthewturk:
Thank you, @Xarthisius. I will address that issue. That happening runs counter to my expectations and I don't have an answer to your other questions, unfortunately.

chummels:
Whoa--this is super interesting!

matthewturk:
Well, *here*'s the problem!

```
Updating index_order2 from 2 to 14
```

Wowza.

yt/geometry/particle_oct_container.pyx
```diff
@@ -504,6 +501,32 @@ cdef class ParticleBitmap:
     def _bitmask_intersects(self, ifile, bcoll):
         return self.bitmasks._intersects(ifile, bcoll)
 
+    def update_mi2(self, np.float64_t characteristic_size):
+        """
+        This takes a characteristic size and attempts to compute the mi2 such
+        that the cell is roughly equivalent to the characteristic size.  It
+        will return whether or not it was able to update; if the mi2 has
+        already been used, it does not update.
```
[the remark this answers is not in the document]
matthewturk: I've added a longer comment!

matthewturk:
I've now limited the heuristic, but I'm somewhat slightly concerned that what it now does is check for the *old* `index_order` in the filename, rather than the new, so it's entirely possible that if it does any modifications to the `index_order`, it will always always generate new ewah files.

@Xarthisius I think this is probably something that would show up on CI as a longer runtime for the example files.

One possible way around this would be to have the filename just have `index_order1` in it, and if it's auto-generated, have it call it "auto" or something.

matthewturk:
Oh, and I should mention, this came up in this particular dataset *because* the gas particles had itty bitty smoothing lengths, because they were just an itty bitty fraction of the domain.  So to accommodate this, the heuristic reset index order 2 to be an appropriate size with respect to the full domain -- thus, 14.  So it was allocating arrays of `2**(3*14)` bits for every collision.

I've been talking about some multi index stuff, which @brittonsmith and I have worked on previously, with @chrishavlin lately, and I think that it would be able to address this by having a separate index for each particle type.

matthewturk:
I think this should move forward and be reviewed.

PR #3203, opened 2021-04-17
fix: correct more-itertools package name

Xarthisius:
I don't know if they renamed the package, but there's no such thing as `more_itertools` package, only `more-itertools`. I'm surprised only py38_latest was failing due to this...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -131,7 +131,7 @@ if __name__ == "__main__":
             "numpy>=1.10.4",
             "IPython>=1.0",
             "unyt>=2.7.2",
-            "more_itertools>=8.4",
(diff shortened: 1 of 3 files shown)


neutrinoceros (approved):
Oh good catch !
It's not that they renamed it, it's because Pypi normalizes packages names (`_` -> `-`) so, while this package is _used_ as `more_itertools` (e.g. `import more_itertools`), it's installed with `pip install more-itertools`.
I'm also surprised that it didn't fail earlier because this mistake on mine has been on the main branch for months, but now that you point it out it's pretty obvious to me.

PR #3205, opened 2021-04-17
DOC: Misc numpydoc syntax fixing.

Carreau:
This uses https://github.com/Carreau/velin to try to automatically fix a
couple of issues in docstrings to match the numpydoc syntax.

 - Space before colon in Parameter sections is necessary for numpydoc to
 properly recognize parameter name and type.
 - Naming of sections in numpydoc is plural usually

This also remove a couple of documented parameters that don't exists
(either removed, or copy-past is my guess).

### DIFF
--- a/yt/_maintenance/deprecation.py
+++ b/yt/_maintenance/deprecation.py
@@ -18,7 +18,7 @@ def issue_deprecation_warning(msg, *, removal, since=None, stacklevel=3):
     """
     Parameters
     ----------
-    msg: str
+    msg : str
         A text message explaining that the code surrounding the call to this function is
         deprecated, and what should be changed on the user side to avoid it.
 
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -2112,10 +2112,6 @@ class YTSurface(YTSelectionContainer3D):
 
         Parameters
         ----------
-        filename : string
-            The file this will be exported to.  This cannot be a file-like object.
-            Note - there are no file extentions included - both obj & mtl files
-            are created.
         transparency : float
             This gives the transparency of the output surface plot.  Values
             from 0.0 (invisible) to 1.0 (opaque).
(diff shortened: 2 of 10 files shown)


neutrinoceros:
Very cool, thanks Matthias !
I see that Velin exposes a pre-commit hook as well, which is awesome. Would you mind adding it to our `.pre-commit-config.yaml` ?

munkm (approved):
This is great! Thanks Matthias!!!!

munkm:
Maybe we can open a "good first contribution" issue to add the pre-commit hook! That would be a nice fix for a new contributor. What do you think @neutrinoceros ?

neutrinoceros:
I think it'd take much less time to do it myself TBF. I'm giving priority to Matthias here as courtesy to the author :)

munkm:
> I think it'd take much less time to do it myself TBF. 

That's the nature of all "good first contribution" issues 🙂

neutrinoceros:
True. I also don't think we should advertise "solve our infrastructure" as good first issues. Most potential contributors wouldn't know what to do, and the others could question our ability to do the job.

Carreau:
> I see that Velin exposes a pre-commit hook as well, which is awesome. Would you mind adding it to our `.pre-commit-config.yaml` ?

I think it is not yet stable enough to be added as a an alway on pre-commit; there are still bugs, like for example if you forget the blank line before the Return section, because of how numpydoc works it will just nuke the section. 

There are also case where it get things wrong like:

```
Yields
------

A tuple of :class:`FooBar`
```

It will replace to ``: class:`FooBar` `` which is invalid RST syntax. 

So maybe a bit later, when I'm more confident that it won't just throw away a paragraph you took 10 minutes to write :-)

neutrinoceros:
Totally fine by me. In that case we can merge this as in AFAIC. Thanks again !

neutrinoceros (approved):

PR #3204, opened 2021-04-17
Upading cut_region docs and use tuples everywhere - fixes #3201

cphyc:
## PR Checklist

Updates a bunch of tuple access to always use the `data[("foo", "bar")]` syntax and update the cut_region doc to use tuples throughout.

- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #3206, opened 2021-04-17
Apply `pre-commit`  on doc - fix invalid Python in docs

cphyc:
## PR Summary

While working on #3204, I realized our doc isn't following the same formatting standard as the rest of the code.

I have applied https://github.com/cphyc/pre-commit-all-the-way-down/ on the doc/**/*.rst files. This applies all our pre-commit on chunks of Python code found in the doc, and here is the results!

[OLD CONTENT BELOW]
> To fix this, I hacked a small script (see below) to apply black + isort + flynt on our code snippets so that they're formatted as the main code.
> I am unsure what would be the best solution to automate that (if we wanted to)?

> Unfortunately, I also realised that many field accesses do not use the tuple syntax. This may become a problem once #832 is solved.

```python
from glob import glob
from textwrap import dedent, indent

import black
import flynt
import isort
from tqdm import tqdm

isort_config = isort.Config("pyproject.toml")


def fix_block(block):
    block = dedent(block)
    black_pass = black.format_file_contents(block, fast=False, mode=black.FileMode())
    isort_pass = isort.code(black_pass, config=isort_config)
    flynt_pass, _N = flynt.api.fstringify_code_by_line(isort_pass)

    return indent(
        flynt_pass,
        prefix="   ",
    )


def fix_python_docstrings(lines):
    state = 0
    newLines = []
    for iline, line in enumerate(lines):
        line = line.rstrip()
        # Start of block
        if line == ".. python-script::":
            state = 1
            start = iline + 1
            newLines.append(line)
        elif state == 1 and not (line.startswith(" ") or line == ""):
            block = "".join(lines[start:iline])
            newLines.append("")
            newLines.extend(fix_block(block).split("\n"))
            newLines.append(line)
            state = 0
        elif state == 0:
            newLines.append(line)

    # Happens if we reach EOF
    if state == 1:
        block = "".join(lines[start:iline])
        newLines.append("")
        newLines.extend(fix_block(block).split("\n"))
        newLines.append(line)

    return newLines


for fname in tqdm(glob("doc/**/*.rst", recursive=True)):
    lines = open(fname).readlines()
    newLines = fix_python_docstrings(lines)

    open(fname, "w").write("\n".join(newLines) + "\n")
```


neutrinoceros:
Wait:
- we are already using https://github.com/asottile/blacken-docs via pre-commit
- there are no equivalent for `isort`
- we opt-out of `flynt` (using py-upgrade instead)

I recommend you make this one or two pre-commit hooks if you want to automate the process. It's pretty easy to build local ones, but we could also define them in a dedicated project, just like blacken-docs.

cphyc:
Yes we have `blacken-docs` indeed, but the `.. python-script::` code blocks do not seem to be detected with blacken-docs. For example, we use `'` to form strings in the doc, even though black changes this to `"`.
In my script above, what I do is:
1. Extract all code snippets,
2. Pass them to black (with our line-length), then isort (reading our doc), then flynt (but could possibly use `py-upgrade`),
3. Write back the modified code snippet to the file.

cphyc:
I have started developing a pre-commit hook for this purpose, see `https://github.com/cphyc/pre-commit-all-the-way-down/`. Once ready, it will allow us to call all our `pre-commit` hooks automatically on Python chunks in our doc (and _also_ in docstrings :tada:).
Until the aforementioned project is mature enough, I am leaving this opened.

doc/source/developing/testing.rst
```diff
@@ -278,7 +278,7 @@ You can find examples there of how to write a test.  Here is a trivial example:
        _attrs = ("field",)
 
        def __init__(self, ds_fn, field):
-           super(MaximumValueTest, self).__init__(ds_fn)
+           super().__init__(ds_fn)
```
neutrinoceros, line 281: this never gets old ❤️

neutrinoceros (approved):
Awesome

doc/source/visualizing/sketchfab.rst
```diff
@@ -296,9 +296,10 @@ like the following:
 
 .. code-block:: python
 
-   import bpy
    from math import *
```
neutrinoceros, line 299: Looks to me like this would do the job
```suggestion
   from math import radians
```

neutrinoceros (approved):
LGTM

doc/source/visualizing/colormaps/index.rst
```diff
@@ -72,7 +72,9 @@ to blue.  These will be accessible for the rest of the yt session as
 .. code-block:: python
 
     yt.make_colormap(
-        [("blue", 20), ("white", 20), ("red", 20)], name="french_flag", interpolate=False
+        [("blue", 20), ("white", 20), ("red", 20)],
+        name="french_flag",
```
cphyc, line 76: 🇫🇷

cphyc:
@yt-fido test this please

doc/source/visualizing/colormaps/index.rst
neutrinoceros: 🐓

matthewturk (approved):

PR #3207, opened 2021-04-17
Adding note in docs about cmocean cmap changes.

chummels:
PR #3175 made a backwards incompatible change to yt dev that breaks existing scripts (it broke mine right under my nose).  This sort of thing needs to be documented in our yt4 differences page to minimize problems for users making a transition to yt4.  I've tried to update the docs appropriately.

Update1: Including @neutrinoceros 's changes.  
Update2: Ensuring the codeblocks work.

### DIFF
--- a/doc/source/yt4differences.rst
+++ b/doc/source/yt4differences.rst
@@ -55,19 +55,25 @@ names raise errors. The fix is to explicitly import the ``cmocean`` module and p
 (like ``balance``) with ``cmo.``.  Note that this solution works with any yt-supported version of Matplotlib, but is not backward compatible with earlier versions of yt. Example:
 
 Old Script (no longer works):
+
 .. code-block:: python
-import yt
-ds = yt.testing.fake_amr_ds()
-p = yt.SlicePlot(ds, "z", "Density")
-p.set_cmap("Density", "balance")
+
+    import yt
+
+    ds = yt.testing.fake_amr_ds()
+    p = yt.SlicePlot(ds, "z", ("gas", "Density"))
+    p.set_cmap(("gas", "Density"), "balance")
 
 New Script:
+
 .. code-block:: python
-import yt
-import cmocean
-ds = yt.testing.fake_amr_ds()
-p = yt.SlicePlot(ds, "z", "Density")
-p.set_cmap("Density", "cmo.balance")
+
+    import yt
+    import cmocean
+
+    ds = yt.testing.fake_amr_ds()
+    p = yt.SlicePlot(ds, "z", ("gas", "Density"))
+    p.set_cmap(("gas", "Density"), "cmo.balance")
 
 Cool New Things
 ---------------
--- a/doc/source/yt4differences.rst
+++ b/doc/source/yt4differences.rst
@@ -53,7 +53,7 @@ yt used to automatically register colormaps from ``cmocean``, unprefixed.
 This unfortunately became unsustainable with the 3.4 release of Matplotlib,
 in which colormaps with colliding names raise errors. The fix is to explicitly
 import the ``cmocean`` module and prefix ``cmocean`` colormaps (like
-``balance``) with ``cmo.``.  Note that this solution works with any 
+``balance``) with ``cmo.``.  Note that this solution works with any
 yt-supported version of Matplotlib, but is not backward compatible with earlier
 versions of yt. Example:
 
--- a/doc/source/yt4differences.rst
+++ b/doc/source/yt4differences.rst
@@ -47,6 +47,27 @@ are being removed entirely. When the deprecated field names are used for the
 first time in a session, a warning will be logged, so it is advisable to set
 your logging level to ``WARNING`` at a minimum to catch these.
 
+Colormaps
+^^^^^^^^^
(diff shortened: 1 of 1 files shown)


doc/source/yt4differences.rst
```diff
@@ -47,6 +47,27 @@ are being removed entirely. When the deprecated field names are used for the
 first time in a session, a warning will be logged, so it is advisable to set
 your logging level to ``WARNING`` at a minimum to catch these.
 
+Colormaps
+^^^^^^^^^
+yt previously included ``cmocean`` colormaps.  Due to collisions between some
+of our dependencies, you now have to explicitly import the ``cmocean`` module
+and prefix ``cmocean`` colormaps (like ``balance``) with ``cmo.``.  Example:
```
neutrinoceros, line 54:
```suggestion
yt used to automatically register colormaps from ``cmocean``, unprefixed.
This unfortunately became unsustainable with the 3.4 release of Matplotlib, in which colormaps with colliding
names raise errors. The fix is to explicitly import the ``cmocean`` module and prefix ``cmocean`` colormaps
(like ``balance``) with ``cmo.``.  Note that this solution works with any versions of yt+Matplotlib. Example:
```

doc/source/yt4differences.rst
```diff
@@ -47,6 +47,27 @@ are being removed entirely. When the deprecated field names are used for the
...
+p = yt.SlicePlot(ds, "z", "Density")
+p.set_cmap("Density", "balance")
+
+New Script:
+.. code-block:: python
+import yt
+import cmocean
+ds = yt.testing.fake_amr_ds()
+p = yt.SlicePlot(ds, "z", "Density")
+p.set_cmap("Density", "cmo.balance")
```
neutrinoceros, line 69: While we're at it, let's avoid adding more ambiguous fields :)
```suggestion
p = yt.SlicePlot(ds, "z", ("gas", "Density"))
p.set_cmap(("gas", "Density"), "balance")

New Script:
.. code-block:: python
import yt
import cmocean
ds = yt.testing.fake_amr_ds()
p = yt.SlicePlot(ds, "z", ("gas", "Density"))
p.set_cmap(("gas", "Density"), "cmo.balance")
```

neutrinoceros (approved):
Thanks for doing this Cameron, I should know better by now... anyway here are some suggestions.

doc/source/yt4differences.rst
neutrinoceros: I originally wrote that your example worked with any version of yt+matplotlib. After working on #3212, I realised this was unfortunately not the case, so I updated this accordingly.

doc/source/yt4differences.rst
neutrinoceros, line 69: I think Sphinx requires some more whitespacing to render code blocks, and while we're at it, let's avoid adding more ambiguous fields :)
```suggestion
Old Script (no longer works):

.. code-block:: python

    import yt

    ds = yt.testing.fake_amr_ds()
    p = yt.SlicePlot(ds, "z", ("gas", "Density"))
    p.set_cmap(("gas", ("gas", "Density")), "balance")

New Script:

.. code-block:: python

    import yt
    import cmocean

    ds = yt.testing.fake_amr_ds()
    p = yt.SlicePlot(ds, "z", ("gas", "Density"))
    p.set_cmap(("gas", "Density"), "cmo.balance")
```

munkm:
Is there a way we can build some sort of warning in 4.0 and actually fully switch in 4.1? Like we could register the colormaps our new way, but if somebody uses an old cmocean color name we can issue a warning and call with the new name internally? If it exists in both matplotlib and cmocean we can actually error out.

neutrinoceros:
That's a possibility, though I'm not certain how it should/could be done. In any case #3212 should be merged first. Right now, as correctly pointed out by Cameron, nothing works !

munkm:
Sure. I hadn't reviewed yet, I just think we need to find (and discuss) a good way to make sure this change has minimal impact on users if we can.

neutrinoceros:
actually I think it should be pretty straightforward.
Can someone please review and merge #3212 ? I think I know how to build a minimal adapter code on top of it that would make the transition easier.

munkm:
I see @chummels is a pending reviewer and I think he is a good choice for it!

doc/source/yt4differences.rst
```diff
@@ -47,6 +47,37 @@ are being removed entirely. When the deprecated field names are used for the
...
+^^^^^^^^^
+yt used to automatically register colormaps from ``cmocean``, unprefixed.
+This unfortunately became unsustainable with the 3.4 release of Matplotlib,
+in which colormaps with colliding names raise errors. The fix is to explicitly
+import the ``cmocean`` module and prefix ``cmocean`` colormaps (like
+``balance``) with ``cmo.``.  Note that this solution works with any
+yt-supported version of Matplotlib, but is not backward compatible with earlier
+versions of yt. Example:
+
+Old Script (no longer works):
```
neutrinoceros, line 60: Following #3214
```suggestion
Old Script (will not work in the near future):
```

cphyc (approved):
Thanks for doing this!

cphyc (approved):

PR #3211, opened 2021-04-18
docs: normalize python code blocks + formatting

neutrinoceros:
## PR Summary

This is in reaction to #3206, which made me realise that our code formatter for documentation isn't working properly because some code blocks are specified as

```rst
.. python-script::
```
instead of
```rst
.. code-block:: python
```
This fixes those, and pre-commit + blacken-docs does the rest automatically.

edit: switching to draft after @cphyc brought to my attention that the `.. python-script::` directive isn't just an alias, and has special rules attached to it that allow these block to be run in parallel in the docs build.


neutrinoceros:
closing this since #3206 will be MUCH more powerful and stable soon

PR #3212, opened 2021-04-18
fix a bug where set_background_color wouldn't work with colormaps from third party libs

neutrinoceros:
## PR Summary

fix #3208

edit: looks like my solution doesn't work across all versions of matplotlib.
For reference, I wrote it for ppl 3.3, now trying to generalise it.
edit 2: it also work with matplotlib 3.4
edit 3: should work for all currently supported versions of matplotlib now 🎉  (thanks to @cphyc  for his help in testing locally with "old" versions)

PR #3214, opened 2021-04-20
add a visible deprecation warning for unprefixed cmocean maps

neutrinoceros:
## PR Summary

Follow up to #3175, #3212 and complementary to #3207

This reintroduces the possibly to use unprefixed colormaps from cmocean, using isolated adapter code that can be dropped after the 4.0 release, and adding a visible deprecation warning were due.

The change is documented in #3207 (where I'll suggest changes to account for this adaptation).

### DIFF
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -1,6 +1,7 @@
 import base64
 import builtins
 import os
+import textwrap
 from collections import defaultdict
 from functools import wraps
 
@@ -19,6 +20,36 @@ from yt.utilities.definitions import formatted_length_unit_names
 from yt.utilities.exceptions import YTNotInsideNotebook
 
 from ._commons import validate_image_name
+
+try:
+    import cmocean  # noqa
+except ImportError:
+    cmocean = None
+
+CMOCEAN_DEPR_MSG = textwrap.dedent(
+    """\
+    It looks like you requested a colormap from the cmocean library. In the future, yt won't
+    recognize this bare (unprefixed) name. Please follow the recommended usage from cmocean's doc:
+    add an explicit `import cmocean` to your script, and prefix their colormap names with 'cmo.'
+    See our release notes for why this change was necessary."""
+)
+
+
+def _sanitize_cmap(cmap):
+    # this function should be removed entierly after yt 4.0.0 is released
+    if not isinstance(cmap, str):
+        return cmap
+    if cmocean is not None:
+        try:
+            cmo_cmap = f"cmo.{cmap}"
+            get_cmap(cmo_cmap)
+            issue_deprecation_warning(CMOCEAN_DEPR_MSG, since="4.0.0", removal="4.1.0")
+            return cmo_cmap
+        except ValueError:
+            pass
+    get_cmap(cmap)
+    return cmap
+
 
 latex_prefixes = {
     "u": r"\mu",
@@ -870,7 +901,7 @@ class ImagePlotContainer(PlotContainer):
 
         """
         self._colorbar_valid = False
-        self._colormap_config[field] = cmap
+        self._colormap_config[field] = _sanitize_cmap(cmap)
         return self
 
     @accepts_all_fields


neutrinoceros:
switching to draft until I can write reliable tests

neutrinoceros:
I didn't manage to write proper tests for this with the time I could afford spending here, but here's the manual test that I used
```python
import yt
import cmocean

ds = yt.testing.fake_amr_ds(fields=[("gas", "density")], units=["g * cm**-3"])

print("set a cmocean native cmap with a proper prefix")
p.set_cmap(("gas", "density"), "cmo.thermal")
p.save("/tmp/1.png")

print("now without a prefix")
p.set_cmap(("gas", "density"), "balance")
p.save("/tmp/2.png")
```

neutrinoceros:
Good news, answer tests on Jenkins correctly caught a bug in my first attempt where I was de facto crippling the ability to set _any_ non-cmocean cmap. Should be fixed now.

yt/visualization/plot_container.py
```diff
@@ -20,6 +21,36 @@
...
+    import cmocean  # noqa
+except ImportError:
+    cmocean = None
+
+CMOCEAN_DEPR_MSG = textwrap.dedent(
+    """
+    It looks like you requested a colormap from the cmocean library. In the future, yt won't "
+    recognize this bare (unprefixed) name. Please follow the recommended usage from cmocean's doc:
+    add an explicit `import cmocean` to your script, and prefix their colormap names with 'cmo.'
+    See our release notes for why this change was necessary."""
```
chummels, line 34:
```suggestion
    WARNING: It looks like you requested a colormap from the cmocean library. In the future, yt won't
    recognize this bare (unprefixed) name. Please follow the recommended usage from cmocean's doc:
    add an explicit `import cmocean` to your script, and prefix their colormap names with 'cmo.'
    See our release notes for why this change was necessary."""
```

chummels (approved):
I tested this and it works great.  Just one small suggestion, to make it more obvious that this is a deprecation warning, but if others think this is too loud, we don't to include it.  I just wanted to make sure that it didn't get lost in the system command output log.

yt/visualization/plot_container.py
neutrinoceros: well it's only ever displayed as a `VisibleDeprecationWarning`, I don't think it's necessary to add "WARNING" to the text itself. @chummels

yt/visualization/plot_container.py
```diff
@@ -20,6 +21,36 @@
...
 from ._commons import validate_image_name
 
+try:
+    import cmocean  # noqa
+except ImportError:
+    cmocean = None
+
+CMOCEAN_DEPR_MSG = textwrap.dedent(
+    """
+    It looks like you requested a colormap from the cmocean library. In the future, yt won't "
```
chummels, line 31:
```suggestion
    It looks like you requested a colormap from the cmocean library. In the future, yt won't
```

yt/visualization/plot_container.py
chummels: OK, no worries.

chummels (approved):

yt/visualization/plot_container.py
```diff
@@ -1,12 +1,13 @@
...
 import builtins
 import os
+import textwrap
 from collections import defaultdict
 from functools import wraps
 
 import matplotlib
 import numpy as np
-from matplotlib.cm import get_cmap
+from matplotlib.cm import get_cmap as get_cmap
```
cphyc, line 10: Is this necessary?

yt/visualization/plot_container.py
```diff
@@ -20,6 +21,36 @@
...
+except ImportError:
+    cmocean = None
+
+CMOCEAN_DEPR_MSG = textwrap.dedent(
+    """
+    It looks like you requested a colormap from the cmocean library. In the future, yt won't
+    recognize this bare (unprefixed) name. Please follow the recommended usage from cmocean's doc:
+    add an explicit `import cmocean` to your script, and prefix their colormap names with 'cmo.'
+    See our release notes for why this change was necessary."""
+)
```
cphyc, line 35: This is required to prevent the string from starting with a line break :)

```suggestion
CMOCEAN_DEPR_MSG = textwrap.dedent(
    """\
    It looks like you requested a colormap from the cmocean library. In the future, yt won't
    recognize this bare (unprefixed) name. Please follow the recommended usage from cmocean's doc:
    add an explicit `import cmocean` to your script, and prefix their colormap names with 'cmo.'
    See our release notes for why this change was necessary."""
)
```

cphyc (approved):

yt/visualization/plot_container.py
neutrinoceros: nope, it's clearly a leftover, good catch

Issue #2737, opened 2020-07-10
Dataset summary


matthewturk:
When yt only did AMR data, it had `print_stats()`.  We don't have this kind of functionality that much anymore.

It would be nice if we could have a base class `Dataset` implementation of something like `summary` that gives some good info, and then dispatch to the subclasses to supplement.  For instance, for an AMR dataset this might be the "level stats" we used to do.  For particle, it might be the particle counts, maybe something about the EWAH index filling fraction, etc.

neutrinoceros:
related (and ancient) issue : #823

matthewturk:
You are absolutely right - very relevant and appropriate.

neutrinoceros:
Random thought: pandas.Dataframe has a `.describe` method that I think does a similar job. Using this name would have some benefits over `print_stats`
- very discoverable to users coming from pandas
- reduces the overhead when you're learning both libraries over a short time
- it doesn't infer a print statement is involved, which leaves rooms for... Jupyter widget intégration !! (If that's ever going to be useful)

matthewturk:
+1000 on `.describe`! Also, I've sketched out some overview ideas in the
widgyts repo in a PR.

neutrinoceros:
Note on our private discussion about this earlier today:
A different but related improvement would be to change `Dataset.__repr__` from

```python
def __repr__(self):
    return self.basename
```

to something more useful such as
```python
def __repr__(self):
    return f"{self.class}: {self.basename}\n{self.field_list}"
```
The only problem being that it may break downstream code where instances of Dataset are directly fed into formatted strings. A possible workaround would be to take advantage of the fact that `__repr__`  is used in string formatting if and only if `__str__`  isn’t implemented, so we would only need to change our current `__repr__` to `__str__` and that'd do the trick.

neutrinoceros:
> t would be nice if we could have a base class Dataset implementation of something like summary that gives some good info, and then dispatch to the subclasses to supplement

I'm thinking a nice way to do that would be to have two methods
```python
class Dataset:
    def _create_summary(self) -> str:
        ...
        return summary

    def summary(self) -> None:
        print(self._create_summary())
```


The `summary` method itself, which has the hard-to-extend side effect of printing, wouldn't need to be extended, and children classes would have the freedom to either
- use the base implementation directly (no overload)
- extend it additively
```python
class DatasetA(Dataset):
    def _create_summary(self) -> str:
        base = super()._create_summary()
[...]

Issue #2879, opened 2020-08-17
Use `fn`, `filename`, `name`, or `my_cool_argname` in the codebase? 


munkm:
This was brought up in yt-project/yt#2877 and yt-project/yt#2695 . 

The second PR in that list makes the arguments passed to various loader functions consistent. Following other examples in our codebase, `fn` is used as the name of the variable to which we assign the input file (or folder). `fn` does (intentionally, I think, though somebody can correct me) create an ambiguity (between file name and folder name?) since we're not necessarily assuming a file is going to be passed into these loaders. Because we can have a folder passed, `filename` isn't the best name to use because it isn't always correct (same with foldername). The `load_sample` function addressed this by using `name` as this argument's name. 

Does anybody have any suggestions of what to use? I personally think `fn` is misleading because I read it is "function" if I'm scanning code. Is there something better?  

Maybe this is bikeshedding, but since load is what essentially all of our users use to access yt I think choosing a clearer argument name is important.

neutrinoceros:
`fname` is another option that I think we can all agree on rejecting immediately since it's used everywhere to mean "field name".
I would love for this argument to be positional only so that we don't have to embrace a bad name forever. But that's not possible until Python 3.8 becomes our oldest supported Python version (so somewhere along the release of Python 3.10 in October 2021, maybe ?)

neutrinoceros:
`fname` seems to be the dominant convention in numpy for reference (https://numpy.org/doc/stable/search.html?q=fname)

PR #3217, opened 2021-04-20
implement a more useful __repr__ for the base Dataset class

neutrinoceros:
## PR Summary

Try to implement an idea mentioned in https://github.com/yt-project/yt/issues/2737#issuecomment-765740805

Following the insight from the accepted answer in https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr

I'm proposing this small change that I think will help in inspecting datasets in REPLs, while hopefully avoiding breakage in any code that relies on the current implem e.g. to format strings.

To be clear on what this (doesn't) change(s), lets run this script
```python
import yt

ds = yt.load_sample("ToroShockTube")  # our tiniest data sample <3
print("raw printing")
print(ds)

print("\nconverting to str")
print(str(ds))

print("\nusing string formatting")
print(f"{ds}")

# simulate what happens if one runs the following line in any REPL (python, IPython, Jupiter Notebook, Jupyter Lab...)
# >>> ds
print("\ncalling __repr__")
print(ds.__repr__())
```

output (main branch) (cutting out logging outputs)
```
raw printing
data0001

converting to str
data0001

using string formatting
data0001

calling repr
data0001
```

this branch
```
raw printing
data0001

converting to str
data0001

using string formatting
data0001

calling repr
EnzoDataset: /Users/robcleme/dev/yt-project/testdata/ToroShockTube/DD0001/data0001
```

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -330,6 +330,9 @@ class Dataset(abc.ABC):
         return (_reconstruct_ds, args)
 
     def __repr__(self):
+        return f"{self.__class__.__name__}: {self.parameter_filename}"
+
+    def __str__(self):
         return self.basename
 
     def _hash(self):
--- a/yt/frontends/art/data_structures.py
+++ b/yt/frontends/art/data_structures.py
@@ -201,7 +201,7 @@ class ARTDataset(Dataset):
             else:
                 setattr(self, "_file_" + filetype, None)
 
-    def __repr__(self):
+    def __str__(self):
         return self._file_amr.split("/")[-1]
 
     def _set_code_unit_attributes(self):
@@ -482,7 +482,7 @@ class DarkMatterARTDataset(ARTDataset):
             else:
                 setattr(self, "_file_" + filetype, None)
 
-    def __repr__(self):
+    def __str__(self):
         return self._file_particle.split("/")[-1]
 
     def _set_code_unit_attributes(self):
--- a/yt/frontends/athena/data_structures.py
+++ b/yt/frontends/athena/data_structures.py
@@ -639,5 +639,5 @@ class AthenaDataset(Dataset):
     def _skip_cache(self):
         return True
 
-    def __repr__(self):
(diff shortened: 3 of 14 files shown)


neutrinoceros:
Looks like some tests are failing, which I didn't really expect, though I'm still pretty confident the failures are likely revealing fragile code relying on an implementation detail, which should be relatively easy to fix.

neutrinoceros:
I think the one side effect I overlooked initially is that some child classes override `Dataset.__repr__` so the fix is to change those to use `__str__` to keep the existing behaviour every where that it actually matters (`str(ds)` is used thoroughly in downhill code). I hope it's good now.

brittonsmith (approved):

yt/data_objects/static_output.py
```diff
@@ -330,6 +330,9 @@ def __reduce__(self):
         return (_reconstruct_ds, args)
 
     def __repr__(self):
+        return f"{self.__class__.__name__}: {self.parameter_filename}"
```
cphyc, line 333: Any reason not to do something like
```suggestion
        return f"{self.__class__.__name__}(filename={self.parameter_filename})"
```
This would allow an almost 1-to-1 mapping between the way the class was initialised and how it is `repr`esented.

cphyc (approved):
This looks like an improvement thanks!
I just left a minor comment that may be ignore depending on your preference ;)

yt/data_objects/static_output.py
neutrinoceros: I don't think this is a bad idea on paper, your suggestions hits a problem that was raised by @munkm as https://github.com/yt-project/yt/issues/2879, so I don't think it's worth doing it this way for now.

This __repr__ should be seen as an implementation detail however, and I do think it can be further improved in the future since any downstream code relying on it should be considered broken IMO.

cphyc: Fair enough; I'm fine with it as is

Issue #3190, opened 2021-04-12
Unable to generate gradient fields for non-periodic dataset


mameehan5:
### Bug report

**Bug summary**

yt does not allow the generation of gradient fields for data sets that are not periodic, even if the field to generate is not at the domain boundaries

**Code for reproduction**

```python
import yt

ds = yt.load('/Users/mikemeehan/Research/PeleData/bplume_Ri4_Re200/plt66875')
ds.add_gradient_fields(('boxlib','z_velocity'))
ds.periodicity = (False, False, False)
sx = yt.SlicePlot(ds,'x','z_velocity_gradient_z',origin="native",center=(0.0,0,0.375),width=0.25)
sx.save()
```
Note that 
```python
ds.domain_left_edge = [-0.5 -0.5  0. ]
ds.domain_right_edge = [0.5 0.5  1.0]
```
for what I am testing.

**Actual outcome**

```
RuntimeError: Error: yt attempted to read outside the boundaries of a non-periodic domain along dimension 1.
Region left edge = -0.515625 code_length, Region right edge = -0.359375 code_length
Dataset left edge = -0.5 code_length, Dataset right edge = 0.5 code_length

This commonly happens when trying to compute ghost cells up to the domain boundary. Two possible solutions are to load a smaller region that does not border the edge or override the periodicity for this dataset.
```

**Expected outcome**

No error should occur since the slice is not at the domain boundary. 

**Version Information**
  * Operating System: macOS Catalina
  * Python Version: 3.6.13
  * yt version: 3.6.1

**Additional Notes**
I installed yt from source.

This something that is relevant for my research so I happy to try resolving it and submit a PR to fix the issue! I am also working on putting together some higher order interpolation so I am already into the source code relevant for the issue.

PR #3218, opened 2021-04-21
ENH: Improved interpolation schemes for smoothed_covering_grid

mameehan5:
## PR Summary

This PR is to include better interpolation methods for generating for the cascaded interpolation procedure in `smoothed_covering_grid`. At the moment, only linear interpolation is supported. With this PR, we will be able to support linear, cubic, quintic, Akima spline and natural spline interpolation.

The original motivation for including higher-order interpolation is that the computation gradient fields at the AMR boundaries can be very inaccurate using linear interpolation. For example, if a flow variable is oscillating smoothly in space, and there is maximum in the value, the coarse cells will under resolve the data compared to the finer level, and the linear interpolation is unable to approximate that peak between the cells. 

Here is a minimal script showing the original field and the gradient field using linear interpolation using an AMReX dataset.
```python
ds = yt.load('plt66875')
ds.add_gradient_fields(('boxlib','z_velocity'))

# Streamwise velocity
fld = 'z_velocity'
sx = yt.SlicePlot(ds,'x',fld,origin="native",center=(0.0,0,0.375),width=0.25)
sx.set_cmap(field=fld, cmap='Blues')
sx.annotate_grids()
sx.set_axes_unit('cm')
sx.set_log(fld,False)
sx.save()

# Streamwise velocity gradient
fld = 'z_velocity_gradient_z'
sx = yt.SlicePlot(ds,'x',fld,origin="native",center=(0.0,0,0.375),width=0.25)
sx.set_cmap(field=fld, cmap='RdBu')
sx.set_axes_unit('cm')
sx.set_log(fld,False)
sx.set_zlim(fld,-3.5,3.5)
sx.save()
```
This produces the following images.
<p float="left">

</p>

Clearly, even with the smooth flowfield, cascaded linear interpolation is not good enough to approximate the gradient at the centerline across the AMR boundary.

Although gradients are the original motivation, including high-order interpolation methods would be very important for a number of other reasons where a high-degree of accuracy is needed.

## PR Checklist

- [x] Refactor functions in `ds.smoothed_covering_grid` 
- [x] Linear interpolation
- [x] Cubic interpolation
- [x] Quintic interpolation
- [x] Akima spline interpolation
- [x] Natural spline interpolation
- [ ] Add tests for different interpolations against analytical functions
- [ ] Address extrapolation problem for vertex centered data
- [x] Add ability to choose interpolation for any smoothed covering grid generated using a particular data container


mameehan5:
Note that the buffer region is _not_ necessarily equal on both sides. They are only equal if the number of cells we are interpolating too (aka, whatever `dims` is as an input) is evenly divisible the ratio of dx of the current level to the interpolated level. If it isn't, the right edge needs a little extra.

yt/data_objects/construction_data_containers.py
```diff
@@ -969,13 +969,31 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
...
+        else:
+            raise NotImplementedError
+        # Assess buffer region surrounding grid
+        if nbuf == None:
+            if order > 0: 
+                nbuf = order//2
+            elif order == -1:
+                nbuf = 3
+        else:
+            assert nbuf >= 1
```
neutrinoceros, line 993: reasons:
- `None` is a singleton, using `==` is bad practice
- don't use `assert` in business-logic code (because assert statements can be ignored with the `-O` flag of the Python executable, no matter how unlikely the actual usage)

Other remarks:
- I'm not sure how `order` could end up being negative, seems impossible in this implementation
- shouldn't the `nbuf` variable depend on the dimensionality of the data somehow ?

```suggestion
        if nbuf is None:
            if order > 0: 
                nbuf = order//2
            elif order == -1:
                nbuf = 3
        elif nbuf < 1:
           raise ValueError
```

cphyc:
Hi @mameehan5 !
Thanks so much for this, it looks wonderful. Do you think it would be possible to target your PR to yt-4? This can be done by clicking the edit button (see image below) and then `wants to merge 2 commits into [main]` rather than into `stable`.
image

neutrinoceros (commented):
quick comments

neutrinoceros:
> Thanks so much for this, it looks wonderful. Do you think it would be possible to target your PR to yt-4? This can be done by clicking the edit button (see image below) and then wants to merge 2 commits into [main] rather than into stable.

good call @cphyc , actually I don't think there's any reason not to do it right away, so I just did.

mameehan5:
My mistake, thanks for catching those! Yes, would you mind rebasing for me? I actually was using yt-3 because I keep getting this error using yt4.0.0
```
(yt-git) ~/Research/yt_bug $ python yt_grad_debug.py 
/Users/mikemeehan/Research/yt-git/yt/utilities/logger.py:4: VisibleDeprecationWarning: The configuration file /Users/mikemeehan/.config/yt/ytrc is deprecated. Please migrate your config to /Users/mikemeehan/.config/yt/yt.toml by running: 'yt config migrate'
Deprecated since v4.0.0 . This feature will be removed in v4.1.0
  from yt.config import ytcfg
(yt-git) ~/Research/yt_bug $ yt config migrate
/Users/mikemeehan/Research/yt-git/yt/utilities/logger.py:4: VisibleDeprecationWarning: The configuration file /Users/mikemeehan/.config/yt/ytrc is deprecated. Please migrate your config to /Users/mikemeehan/.config/yt/yt.toml by running: 'yt config migrate'
Deprecated since v4.0.0 . This feature will be removed in v4.1.0
  from yt.config import ytcfg
(yt-git) ~/Research/yt_bug $ 
```
Anything to do to fix this so I can test/run using yt4.0?

cphyc:
@mameehan5 the issue you're seeing is only a warning. It is due to the configuration file `/Users/mikemeehan/.config/yt/ytrc`being deprecated. You can get rid of it by migrating your config to `/Users/mikemeehan/.config/yt/yt.toml` by running: 
```shell
yt config migrate
```

mameehan5:
Thanks so much @cphyc and @neutrinoceros ! Got everything working locally again with yt-4. Going to add a lot of commits soon!

mameehan5:
In order to simplify the code, I included a buffer region _even_ if we are interpolating to a grid at `l=0`. This is technically unnecessary if we are creating a smoothed covering grid at `l=0`, but if someone tried doing that, they should be using `covering_grid` anyways since no interpolation is required. Happy to discuss further though!

mameehan5:
It looks like the other places in the code actually use `ghost_zone_interpolate` for extrapolation. I personally wouldn't expect this behavior, and I think that we should add an explicit function, say `ghost_zone_extrapolate` to do this. This would also be useful to extrapolate data at domain boundaries for non-periodic datasets.

yt/data_objects/grid_patch.py
```diff
@@ -0,0 +1,437 @@
...
+"""
+
+#-----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#-----------------------------------------------------------------------------
+
```
neutrinoceros: We don't do those anymore, you can drop this :)
```suggestion
```

yt/data_objects/grid_patch.py
```diff
@@ -0,0 +1,437 @@
+"""
```
neutrinoceros, line 1: Wait are you reintroducing a whole deleted file here ?

mameehan5: Whoops, yes, looks like I did that by mistake.

yt/data_objects/construction_data_containers.py
```diff
@@ -1290,35 +1290,51 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
...
+        elif interp == "cubic":
+            order = 4
+        elif interp == "quintic":
+            order = 6
+        elif interp == "akima":
+            order = -1
+        elif interp == "natural":
+            order = -2
+        else:
+            raise NotImplementedError
```
neutrinoceros, line 1313:
```suggestion
        order = {
            "natural": -2,
            "akima": -1,
            "linear": 2,
            "cubic": 4,
            "quintic": 6,
           }[interp]
```

yt/data_objects/construction_data_containers.py
```diff
@@ -1290,35 +1290,51 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
     _min_level = None
 
     @wraps(YTCoveringGrid.__init__)
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, interp="linear", nbuf=None, **kwargs):
```
neutrinoceros, line 1293: so as to avoid any possible collision with positional arguments (`*args`), it may be worth making these new arguments keyword only :
```suggestion
    def __init__(self, *args, *, interp="linear", nbuf=None, **kwargs):
```

yt/data_objects/construction_data_containers.py
neutrinoceros: Btw I think `kind` would be a clearer name than "interp" as a user-exposed parameter.

mameehan5: I agree. I will update that!

yt/data_objects/construction_data_containers.py
mameehan5: Yep, that sounds good. I didn't realize that is how you did it.

yt/utilities/lib/interpolators.pyx
```diff
@@ -113,53 +113,99 @@ def TrilinearlyInterpolate(np.ndarray[np.float64_t, ndim=3] table,
 @cython.cdivision(True)
 @cython.wraparound(False)
 @cython.boundscheck(False)
-def ghost_zone_interpolate(int rf,
+def ghost_zone_interpolate(int rf, int order,
                            np.ndarray[np.float64_t, ndim=3] input_field,
                            np.ndarray[np.float64_t, ndim=1] input_left,
                            np.ndarray[np.float64_t, ndim=3] output_field,
                            np.ndarray[np.float64_t, ndim=1] output_left):
```
cphyc, line 120: Just for the sake of having indentation in cython file that ressembles that of python files
```suggestion
def ghost_zone_interpolate(
    int rf,
    int order,
    np.ndarray[np.float64_t, ndim=3] input_field,
    np.ndarray[np.float64_t, ndim=1] input_left,
    np.ndarray[np.float64_t, ndim=3] output_field,
    np.ndarray[np.float64_t, ndim=1] output_left
):
```

yt/utilities/lib/interpolators.pyx
```diff
@@ -113,53 +113,99 @@ def TrilinearlyInterpolate(np.ndarray[np.float64_t, ndim=3] table,
...
-            opos[1] += ods[1]
-        opos[0] += ods[0]
+    # Counting indices
+    cdef int io, jo, ko
+    cdef int ii, ji, ki, li, mi, ni
+    cdef int l0, m0, n0
+    cdef int i, j, k
+    # dx of the input and output fields
+    cdef double dxi = 1.0
+    cdef double dxo = dxi/rf
```
cphyc, line 128:
```suggestion
    cdef double dxo = dxi / rf
```

yt/utilities/lib/interpolators.pyx
```diff
@@ -113,53 +113,99 @@ def TrilinearlyInterpolate(np.ndarray[np.float64_t, ndim=3] table,
...
+                    ki = iclip(ki, 0, nxi[2]-nxs[2])
+                    for k in range(0,nxs[2]): zn[k]=ki+k-(nxs[2]/2-1)
+                    zw = lagrange_weights(zn, kposi)
+                    output_field[io,jo,ko] = 0.0
+                    for l0,li in enumerate(range(ii,ii+nxs[0])):
+                        for m0,mi in enumerate(range(ji,ji+nxs[1])):
+                            for n0,ni in enumerate(range(ki,ki+nxs[2])):
+                                output_field[io,jo,ko] += xw[l0]*yw[m0]*zw[n0]*\
+                                                          input_field[li,mi,ni]
+
```
cphyc, line 183:
```suggestion
                    for l0, li in enumerate(range(ii, ii + nxs[0])):
                        for m0, mi in enumerate(range(ji, ji + nxs[1])):
                            for n0, ni in enumerate(range(ki, ki + nxs[2])):
                                output_field[io, jo, ko] += xw[l0] * yw[m0] * zw[n0] * \
                                                          input_field[li, mi, ni]

```

yt/utilities/lib/interpolators.pyx
```diff
@@ -113,53 +113,99 @@ def TrilinearlyInterpolate(np.ndarray[np.float64_t, ndim=3] table,
...
+            for jo in range(nxo[1]):
+                ji = ifloor(jposi)
+                ji = iclip(ji, 0, nxi[1]-nxs[1])
+                for j in range(0,nxs[1]): yn[j]=ji+j-(nxs[1]/2-1)
+                yw = lagrange_weights(yn, jposi)
+                kposi = kposi0
+                for ko in range(nxo[2]):
+                    ki = ifloor(kposi)
+                    ki = iclip(ki, 0, nxi[2]-nxs[2])
+                    for k in range(0,nxs[2]): zn[k]=ki+k-(nxs[2]/2-1)
```
cphyc, line 175:
```suggestion
                    for k in range(0, nxs[2]):
                        zn[k] = ki + k - (nxs[2] / 2 - 1)
```

yt/utilities/lib/interpolators.pyx
```diff
@@ -113,53 +113,99 @@ def TrilinearlyInterpolate(np.ndarray[np.float64_t, ndim=3] table,
...
+        for io in range(nxo[0]):
+            ii = ifloor(iposi)
+            ii = iclip(ii, 0, nxi[0]-nxs[0])
+            for i in range(0,nxs[0]): xn[i]=ii+i-(nxs[0]/2-1)
+            xw = lagrange_weights(xn, iposi)
+            jposi = jposi0
+            for jo in range(nxo[1]):
+                ji = ifloor(jposi)
+                ji = iclip(ji, 0, nxi[1]-nxs[1])
+                for j in range(0,nxs[1]): yn[j]=ji+j-(nxs[1]/2-1)
```
cphyc, line 169:
```suggestion
                for j in range(0, nxs[1]):
                    yn[j] = ji + j - (nxs[1] / 2 - 1)
```

yt/utilities/lib/interpolators.pyx
```diff
@@ -113,53 +113,99 @@ def TrilinearlyInterpolate(np.ndarray[np.float64_t, ndim=3] table,
...
+    iposi0 = (output_left[0]+0.5/rf) - (input_left[0]+0.5) - (nxs[0]/2-1)
+    jposi0 = (output_left[1]+0.5/rf) - (input_left[1]+0.5) - (nxs[1]/2-1)
+    kposi0 = (output_left[2]+0.5/rf) - (input_left[2]+0.5) - (nxs[2]/2-1)
+
+    if order > 0:
+        iposi = iposi0
+        for io in range(nxo[0]):
+            ii = ifloor(iposi)
+            ii = iclip(ii, 0, nxi[0]-nxs[0])
+            for i in range(0,nxs[0]): xn[i]=ii+i-(nxs[0]/2-1)
```
cphyc, line 163:
```suggestion
            for i in range(0, nxs[0]):
                xn[i] = ii + i - (nxs[0] / 2 - 1)
```

yt/utilities/lib/interpolators.pyx
```diff
@@ -113,53 +113,99 @@ def TrilinearlyInterpolate(np.ndarray[np.float64_t, ndim=3] table,
...
+    for i in range(1,nx):
+        c2 = 1.0
+        c5 = c4
+        c4 = x[i] - xp
+        for j in range(i):
+            c3 = x[i] - x[j]
+            c2 = c2*c3
+            if j == i-1:
+                wgts[i] = -c1*c5*wgts[i-1]/c2
+            wgts[j] = c4*wgts[j]/c3
```
cphyc, line 209:
```suggestion
            c2 = c2 * c3
            if j == i - 1:
                wgts[i] = -c1 * c5 * wgts[i-1] / c2
            wgts[j] = c4 * wgts[j] / c3
```

yt/data_objects/construction_data_containers.py
```diff
@@ -1404,8 +1415,10 @@ def _fill_fields(self, fields):
             )
             mylog.debug("Caught %d runtime errors.", runtime_errors_count)
         for name, v in zip(fields, ls.fields):
-            if self.level > 0:
-                v = v[1:-1, 1:-1, 1:-1]
+            # Remove buffer region
+            v = v[
+                self.nbuf : -self.nbuf, self.nbuf : -self.nbuf, self.nbuf : -self.nbuf
+            ]
```
cphyc, line 1421: If you want to not repeat yourself, may I suggest
```suggestion
            # Remove buffer region
            sl = slice(self.nbuf, -self.nbuf)
            v = v[(sl, sl, sl)]
```

yt/data_objects/construction_data_containers.py
```diff
@@ -1290,35 +1290,46 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
...
+            "quintic": 6,
+        }[kind]
+        # Assess buffer region surrounding grid
+        if nbuf is None:
+            if order > 0:
+                nbuf = order // 2
+            else:
+                nbuf = 3
+        elif nbuf < 1:
+            raise ValueError
```
cphyc, line 1316: I love this chunk!

I think this would fail with a `KeyError`, say if I pass `kind="dinosaur"`. Would it be possible to catch this and raise a more useful exception as this is directly user-facing?

```suggestion
       order_nbuf = {
            "natural": (-2, 3),
            "akima": (-1, 3),
            "linear": (2, 1),
            "cubic": (4, 2),
            "quintic": (6, 3)
        }
        try:
            order, nbuf_from_order = order_nbuf[kind]
        except KeyError:
            msg = f"Interpolation '{kind}' has not been implemented. "
            msg += "Only the following are supported: "
            msg += ", ".join(order_nbuf.keys())
            raise NotImplementedError(msg)
        nbuf = nbuf or nbuf_from_order
        if nbuf < 1:
            raise ValueError
```

cphyc (commented):
This is taking shape! I have left several code suggestions (which you can accept directly from github) regarding code formatting.

yt/data_objects/construction_data_containers.py
neutrinoceros: the raise statement should be a `raise from`. Other than that, great suggestion

yt/data_objects/construction_data_containers.py
mameehan5: Yes, I like this much better!

yt/data_objects/construction_data_containers.py
mameehan5: @cphyc I really like this as well! @neutrinoceros , which line exactly do you mean? Both raise statements?

neutrinoceros: I mean the nested raise.
The difference is that `raise from` clearly indicates that the original error wasn't meant to be handled, but rather wrapped, while if you just raise an error directly in à `except` block, the error message looks like the new error wasn't intended.

mameehan5: Got it. So we should change line 1311 to `except KeyError as k` and line 1315 to `raise NotImplementedError(msg) from k`, right?

mameehan5: Or something like that.

yt/utilities/lib/interpolators.pyx
mameehan5: Yep, that is good!

yt/data_objects/construction_data_containers.py
```diff
@@ -1290,35 +1290,48 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
...
+            "cubic": (4, 2),
+            "quintic": (6, 3),
+        }
+        try:
+            order, nbuf_from_order = order_nbuf[kind]
+        except KeyError as k:
+            msg = f"Interpolation '{kind}' has not been implemented. "
+            msg += "Only the following are supported: "
+            msg += ", ".join(order_nbuf.keys())
+            raise NotImplementedError(msg) from k
```
neutrinoceros, line 1315: Sorry to bother you _again_ with this. Please avoid one-letter variable names, they are much harder to grep for =)
```suggestion
        except KeyError as err:
            msg = f"Interpolation '{kind}' has not been implemented. "
            msg += "Only the following are supported: "
            msg += ", ".join(order_nbuf.keys())
            raise NotImplementedError(msg) from err
```

mameehan5: Definitely no problem! Changing now.

mameehan5:
At this point, extrapolation has been add but it needs to be generalized. The issue in generalizing is the following.

If we are attempted to extract a smoothed covering grid that extends beyond the domain of a non-periodic data set, the RegionSelector throws an error immediately without exception. However, we want there to be two exceptions:
1. if the user sets up an extrapolation method, or
2. if the data chunk actually never gets used (related to #3190)

I will need to give this more thought on how to generalize further to account for this. I tried a couple different things, the most promising of which appeared to assuming data is always periodic, then replacing the periodic cells with extrapolated cells. The issue with this was that I couldn't propagate the information in the mask that this cell needed to replaced by extrapolation.

neutrinoceros:
I'll rerun tests on Jenkins now because the reports are lost

@yt-fido test this please

FYI @mameehan5, we're doing a "PR mentorship week" starting on August 2nd, so it will be a good opportunity to give this the final push, if it's still not merged at that point and it you're interested :)

mameehan5:
@neutrinoceros that would be excellent! Sorry I am making such small recent progress. There is _one last_ issue I am running into for a very edge case. Once I have that figured out, I will put in all the tests I have and update other outstanding issues. It is very close :) It will get finished at some point though, and my dissertation as well ha!

neutrinoceros:
> Sorry I am making such small recent progress.

You don't have anything to apologize for and you don't owe me anything, don't worry.

Good luck with your dissertation !

yt/frontends/boxlib/io.py
```diff
@@ -122,6 +122,7 @@ def _read_chunk_data(self, chunk, fields):
                         local_offset = 0
                     else:
                         local_offset += size
+            f.close()
```
neutrinoceros, line 125: @mameehan5 wasn't this supposed to be it's own PR ?

matthewturk: It's probably OK to include it here if we're aiming to get this PR in soon.

neutrinoceros: sure, but I'd merge this on the spot if it had it's own PR :)

mameehan5: Yes, I knew to add it as a separate PR but I was testing first on this branch. Updating this branch is still on my todo list but I am not sure how soon it will get done. I added a new PR for just the close statement though! #3515

matthewturk:
@mameehan5 I really appreciate these changes and would love to see them in.  Any way we can help you out?

mameehan5:
I would to :) let me set aside some time this week to get this into a (somewhat) mergeable form so we can get the process moving again!

mameehan5:
I wasn't able to make any meaningful progress during that time I set aside before so I didn't put anything here. I plan to complete this PR though after I defend in early May. I just want to make perfectly sure that the changes I put in don't break anything. This will require some additional testing routines, which I am not as comfortable putting in compared to the actual algorithms but I will figure it out!

matthewturk:
Hey Mike, thanks for the update!  Hope you're doing well, and best of
luck with the defense!

mameehan5:
I have been digging into this problem a bit and I think I need to back track to first address a couple other issues I am having specific to computing gradients and other quantities at non-periodic domain boundaries. Rather than try to address these here, and given that this PR is seriously outdated, I will close this PR and create a new PR first addressing domain boundary issues. After that is fixed, then I will come back (with a new PR) and address gradients at AMR boundaries.

PR #3220, opened 2021-04-21
Apply `pre-commit` on docstrings - fix invalid Python in docs

cphyc:
Attempts to use https://github.com/cphyc/pre-commit-all-the-way-down/ on our docstrings.


.pre-commit-config.yaml
```diff
@@ -49,3 +49,8 @@ repos:
     hooks:
     -   id: blacken-docs
         additional_dependencies: [black==20.8b1]
+-   repo: ../pre-commit-all-the-way-down/
```
neutrinoceros, line 52: shouldn't it replace blacken-docs completely ?

neutrinoceros: Anyway I see you used is with a local config, I except adding the hook here is actually temporary since the project may not be stable enough yet?

cphyc:
I think https://github.com/cphyc/pre-commit-all-the-way-down/ is too slow at the moment to be used in prod. There are many ways it can improve in the future (and I intend to do it), but in the meantime I think this PR can go as is!

neutrinoceros (approved):
Thanks a lot for doing this ! Beyond the consistency and arguable "prettiness" improvements here, the mere fact that you were able to run black there means that you validated the doctests as valid Python. It doesn't mean they _pass_, but that's certainly an excellent first step to perform before we can make them part of CI via pytest.

neutrinoceros:
Classifying this as Infra (CI) by anticipation.

yt/data_objects/selection_objects/data_selection_objects.py
```diff
@@ -682,8 +682,8 @@ def cut_region(self, field_cuts, field_parameters=None, locals=None):
 
         >>> ds = yt.load("RedshiftOutput0005")
         >>> ad = ds.all_data()
-        >>> cr = ad.cut_region(["obj['gas', 'temperature'] > 1e6"])
-        >>> print(cr.quantities.total_quantity(("gas", "cell_mass")).in_units("Msun"))
+        >>> cr = ad.cut_region(["obj[('gas', 'temperature')] > 1e6"])
+        >>> print(cr.quantities.total_quantity(("gas", "cell_mass")).in_units('Msun'))
```
neutrinoceros: looks like black will overwrite this in parts

cphyc: Damn, I accepted the wrong side of the merge...

cphyc:
I have tagged it as "bug" because it fixes several invalid code snippets in our docstrings (see e.g. https://github.com/yt-project/yt/pull/3220/files#diff-459eb66cb29de05519a40f30db1bf3653d4c13c115829dd48cc4409fc3149e0bR2344, where the final `)` was missing).

matthewturk:
LGTM.  Got some conflicts, but I think it should go in once those are resolved.

matthewturk (approved):

PR #3221, opened 2021-04-22
fix a deprecation warning where unyt_array instances would be constructed from 'ragged' arrays

### LINKED ISSUE #3096: A visible deprecation warning in CI
### Bug report

**Bug summary**
We have one test that triggers a VisibleDeprecationWarning from numpy via unyt.
I'm not sure atm where the problem should be addressed:
- in the test itself ?
- in the yt code under test ?
- in unyt ?

**Code for reproduction**

```shell
pytest yt/data_objects/tests/test_derived_quantities.py::test_in_memory_sph_derived_quantities
```

**Actual outcome**

```
.../python3.9/site-packages/unyt/array.py:250: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
    return unyt_array(np.array(input_object), ff, registry=registry)
```
This error is visible in our non minimal GH CI workflows.

neutrinoceros:
## PR Summary

Fix #3096

### DIFF
--- a/yt/data_objects/derived_quantities.py
+++ b/yt/data_objects/derived_quantities.py
@@ -201,11 +201,11 @@ class TotalMass(TotalQuantity):
         if ("gas", "mass") in fi:
             gas = super().__call__([("gas", "mass")])
         else:
-            gas = self.data_source.ds.arr([0], "g")
+            gas = self.data_source.ds.quan(0.0, "g")
         if ("nbody", "particle_mass") in fi:
             part = super().__call__([("nbody", "particle_mass")])
         else:
-            part = self.data_source.ds.arr([0], "g")
+            part = self.data_source.ds.quan(0.0, "g")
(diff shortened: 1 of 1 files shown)


matthewturk (commented):
I'm surprised this ever worked if it doesn't work now.  The issue if I read it correctly is that we're setting up a situation where we might be doing the unyt-equivalent of:

`np.array( [ scalar, [ scalar ] ])`

right?

neutrinoceros:
That's exactly it. To be clear, it *does* work for now, but the deprecation warning informs us that it *won't* in a future version of numpy, which is why I was concerned with it :)

matthewturk (approved):

jzuhone (approved):

PR #3224, opened 2021-04-22
cleanup some deprecated field names

neutrinoceros:
## PR Summary

This is follow up to  #3059
I'm trying to cleanup the associated deprecation warnings generated by this change that are visible in our CI logs.


neutrinoceros:
@yt-fido test this please

neutrinoceros:
one of the warnings I'm addressing here is
```
WARNING:yt:The Derived Field ('gas', 'cylindrical_tangential_velocity_absolute') is deprecated as of yt v4.0.0 and will be removed in yt v4.1.0.
```
So it looks like there is no replacement for this one ? I've attempted to use `"velocity_cylindrical_theta_absolute"` instead but indeed this fails.
@jzuhone , is this intended ? what do you think should be done with this cookbook file ($doc/source/cookbook/customized_profile_plot.py`) ?

cphyc (approved):

matthewturk (approved):

jzuhone:
@neutrinoceros the absolute fields have been marked as deprecated for some time in a comment inside the field definition, so I would venture we shouldn't have them in the docs at all. I think you can change the cookbook to profile something else.

PR #3238, opened 2021-04-30
fix a bug with DatasetSeries instances created via __new__

### LINKED ISSUE #3200: missing declaration of used attribute?
### Bug report

while trying to load enzo data into yt (cf. python code below), we run into an issue that the (lazy?) loading in `yt.data_objects.DatasetSeries` is triggered when accessing `__get_item__ `of the resulting EnzoSimulation objects which fails with an AttributeError (cf. stacktrace below) for using `self.kwargs`. Afaik this attribute is only declared in `DatasetSeries.__init__` which itself is not invoked by its subclasses involved here (`yt.data_objects.SimulationTimeSeries` -> `yt.frontends.enzo.simulation_handling.EnzoSimulation`). Hence the loading failed.

**Code for reproduction**

```python
import yt
from yt.extensions.astro_analysis.halo_analysis import HaloCatalog
yt.enable_parallelism()
 
es = yt.load_simulation("/mnt/lustre/carla/Box100/RandomSeedStudy/Wunder-L4/parameter_file.txt", "Enzo", find_outputs=True)
 
hc = HaloCatalog(data_ds=es, finder_method='rockstar', finder_kwargs={"num_readers": 2, "num_writers": 4})
hc.create()
```

**Actual outcome**

```
File "/users/carla/scripts/run_rockstar-galaxies.py", line 9, in <module>
    hc.create()
  File "/opt/apps/pkgs/yt_astro_analysis/313cf31/intel64/gnu_9.1.0/cpython_3.8.9/lib/python3.8/site-packages/yt_astro_analysis/halo_analysis/halo_catalog/halo_catalog.py", line 333, in create
    self._run(save_halos, save_output,
  File "/opt/apps/pkgs/yt/4.0.dev0/intel64/gnu_9.1.0/cpython_3.8.9/lib/python3.8/site-packages/yt/utilities/parallel_tools/parallel_analysis_interface.py", line 310, in barrierize
    retval = func(*args, **kwargs)
  File "/opt/apps/pkgs/yt_astro_analysis/313cf31/intel64/gnu_9.1.0/cpython_3.8.9/lib/python3.8/site-packages/yt_astro_analysis/halo_analysis/halo_catalog/halo_catalog.py", line 224, in _run
    self.finder_method(self)
  File "/opt/apps/pkgs/yt_astro_analysis/313cf31/intel64/gnu_9.1.0/cpython_3.8.9/lib/python3.8/site-packages/yt_astro_analysis/halo_analysis/halo_catalog/halo_finding_methods.py", line 39, in __call__
    return self.function(hc, *self.args, **self.kwargs)
  File "/opt/apps/pkgs/yt_astro_analysis/313cf31/intel64/gnu_9.1.0/cpython_3.8.9/lib/python3.8/site-packages/yt_astro_analysis/halo_analysis/halo_catalog/halo_finding_methods.py", line 87, in _rockstar_method
    rh = RockstarHaloFinder(ds, **finder_kwargs)
  File "/opt/apps/pkgs/yt_astro_analysis/313cf31/intel64/gnu_9.1.0/cpython_3.8.9/lib/python3.8/site-packages/yt_astro_analysis/halo_analysis/halo_finding/rockstar/rockstar.py", line 234, in __init__
    tds = ts[-1] # Cache a reference
  File "/opt/apps/pkgs/yt/4.0.dev0/intel64/gnu_9.1.0/cpython_3.8.9/lib/python3.8/site-packages/yt/data_objects/time_series.py", line 214, in __getitem__
    o = self._load(o, **self.kwargs)
P000 yt : [ERROR    ] 2021-04-14 15:27:09,625 AttributeError: 'EnzoSimulation' object has no attribute 'kwargs'
```

**Expected outcome**

well, I expect no AttributeError ;)

**Version Information**
  * Operating System: CentOS 7
  * Python Version: 3.8.9  (cpython)
  * yt version: 4.0.dev0 
  * Other Libraries (if applicable): yt_astro_analysis (latest commit on master i.e. 313cf31)

python itself as well as yt and yt_astro_analysis are compiled/installed from source. Supporting packages/dependencies (e.g. unyt) installed with pip. 

Thanks again for any help you can provide to clarify this issue.

neutrinoceros:
## PR Summary

I believe this is the appropriate to fix #3200, though I'll admit I haven't been able to test it.

### DIFF
--- a/yt/data_objects/time_series.py
+++ b/yt/data_objects/time_series.py
@@ -155,6 +155,7 @@ class DatasetSeries:
             pass
         ret = super().__new__(cls)
         ret._pre_outputs = outputs[:]
+        ret.kwargs = {}
         return ret
 
     def __init__(


brittonsmith:
I just left a comment on the issue, but the fix is that `get_time_series` needs to be called after `load_simulation`. This line was missing from yt-astro-analysis docs, but I fixed that in a recent PR there. We should probably add better error handling to make the error more informative. I can do that.

neutrinoceros:
Good to known, thank you for clarifying this.
I still think this change should be included, because I can't think of a reason why the DatasetSeries object should be different depending on whether it was created through __init__ or __new__

matthewturk:
@brittonsmith during PR triage @neutrinoceros had a question about why we have a `__new__` method and my recollection is that it was related to needing the glob to be called before anything else.  But, it's also possible that that's leftover from Py2, and also from our previous thinking about usage patterns.  Do you think we could set up some tests that might help elucidate if it's still necessary?

brittonsmith:
@matthewturk looking at it now, I don't see why the `_get_filnames_from_glob_pattern` functionality cannot be moved to `__init__`. I'm in favor of doing that.

neutrinoceros:
ok I'll try this in this PR then

neutrinoceros:
Well actually I'm still not confortable with touching this, I really don't feel like I understand this enough to confidently refactor it.

Currently we have 8 classes that define a __new__ method, some of which look like this

```python
    def __new__(cls, *args, **kwargs):
        obj = object.__new__(cls)
        obj.__init__(*args, **kwargs)
        return obj
```
which to me looks like a pure no-op (except maybe in classes that inherit a more complicated implementation that for some reason is undesirable in children ??)
and from I can tell _all_ the classes in questions also define a __init__ method.
I have _no_ idea what motivated these patterns. I'm sure there were reasons for it at the time, but the lack of comment makes it difficult to reevaluate.
Anyway it looks like you (@matthewturk) authored most of the __new__ methods we have about 12 yrs ago, so I would get it if you tell me you don't remember the reasoning now.

matthewturk:
If I am remembering correctly, we often used `__new__` in instances where we either had classmethods or wanted singletons, but I may be misremembering.

neutrinoceros:
I'm good 🙂

PR #3240, opened 2021-05-01
DEPR: deprecation cycle (4.0 -> 4.1)

neutrinoceros:
## PR Summary

This removes all code that was targeted for removal for yt 4.1, and associated documentation, with the notable exception of ambiguous fields. This part is left to be solved with #3416 (also included in the 4.1 milestone).

### DIFF
--- a/doc/source/cookbook/complex_plots.rst
+++ b/doc/source/cookbook/complex_plots.rst
@@ -46,10 +46,7 @@ or during image creation with the ``buff_size`` argument to any
 of the four preceding classes.
 
 2. ``figure_size``, which can be altered with either
-:meth:`~yt.visualization.plot_container.PlotContainer.set_figure_size`
-or with :meth:`~yt.visualization.plot_container.PlotWindow.set_window_size`
-(the latter simply calls
-:meth:`~yt.visualization.plot_container.PlotContainer.set_figure_size`),
+:meth:`~yt.visualization.plot_container.PlotContainer.set_figure_size`,
 or can be set during image creation with the ``window_size`` argument.
 This sets the size of the final image (including the visualization and,
 if applicable, the axes and colorbar as well) in inches.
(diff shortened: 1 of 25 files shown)


neutrinoceros:
(adding the "refactor" label just to satisfy the mergeable bot)

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@matthewturk do you want to issue a warning on the mailing list ?

chummels:
I can try to look at this in the next week to add these deprecated/removed features to the yt4differences docs page, but honestly, I'm not super comfortable with this being merged until then.  That page is explicitly written for users on how to update their scripts to yt4 to avoid breakages, so it really deserves to be updated with information about these functions that are being removed from the code which will cause script breakages.

neutrinoceros:
Arguably such updates would need to be backported to the 4.0.x branch. Otherwise it won't reach the stable documentation before it's too late (4.1). The present PR however has to stay on the 4.1 branch (current main). If only for this reason, I'd like to suggest that the docs be updated in a separate PR, regardless of the final merging order.

neutrinoceros:
Following our discussion on Slack, I've repurposed #3451 and targeted it for the 4.0.2 release. Now if I open an issue to remind ourself to document these removals in time for the 4.1 release, can we get this forward @chummels ?

chummels:
I know we keep going back and forth here, but I don't understand why we don't just write the documentation now before these features are removed?  If someone is using the dev version of the code, and this PR gets merged, their scripts will fail and they'll want to figure out why, and likely go to the docs to figure it out.  Why wait to document this change until the release of 4.1?  Can you just include the relevant documentation before merging the code?

neutrinoceros:
I'm working to document them _right now_, but it needs to be done in a separate PR because the docs need to be back ported to the stable branch while the present PR needs to stay on the dev branch only

neutrinoceros:
done in #3665

neutrinoceros:
@chummels can you merge this now ?

chummels (approved):
Thanks for documenting this.

PR #3242, opened 2021-05-01
maint: cleanup usage and mentions of deprecated material

neutrinoceros:
## PR Summary

I discovered some deprecated functionalities are still used internally or documented as usable while they really aren't (add_volume_weighted_smoothed_field)

### DIFF
--- a/doc/source/analyzing/fields.rst
+++ b/doc/source/analyzing/fields.rst
@@ -806,10 +806,6 @@ default this is 64, but it can be supplied as the final argument to
 ``add_nearest_neighbor_field``.  For the example above, it would be
 ``nearest_neighbor_64``.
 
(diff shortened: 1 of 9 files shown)


neutrinoceros:
Failures on Jenkins are real. Some of the answer tests' names contains the name of a deprecated function `set_window_size`, so the newly named generated tests don't match any existing answer.
I read the docs in search for a solution and found
https://yt-project.org/doc/developing/testing.html#enabling-answer-tests-on-jenkins
>If the answer doesn’t exist on the server yet, nosetests is run twice and during first pass --answer-store is added to the commandline.

which doesn't seem appropriate.
I'll try bumping the answers and hope it'll trigger regeneration.

matthewturk (approved):

PR #3236, opened 2021-04-29
maint: merge dependency specifications into setup.cfg

neutrinoceros:
## PR Summary

This is a second step forward in trying to solve #3074, following #3078.
Goal: remove as much content as possible from
- `tests/test_minimal_requirements.txt`
- `tests/test_requirements.txt`
- and`tests/test_prerequirements.txt`

and migrate it to setup.cfg
The migration process is error prone and I expect a few iterations will be required.

TODO
- [x] define a `test` target
- [x] cleanup duplicated content
- [x] define a `minimal` target
- [x] cleanup duplicated content
- [x] check that the overlap between minimal req and requirements leads to a correct env (e.g if we mix `matplotlib>=2.0.2` with `matplotlib==2.0.2`, do we actually satisfy that last requirement ?)
- [x] define a `full` target (corresponding to `test_requirements.txt`
- [x] cleanup duplicated content
- [x] check `pyyaml` VS `pyaml` (both are currently "required" from test_requirements.txt but I bet that's a useless redundancy)
- [x] cleanup (self review)
- [x] reactivate the pre-commit hook (can be done at any point but I had problems with it in my first attempt)
- [x] reviews
- [ ] deal with the Jenkins side of things

This last step can be performed with a conditional similar to
```
if [ -f test_requirements.txt ] ; then
    # old strategy
else
   # new strategy (replicate what I'm doing in ci_install.sh
fi
```

### Cleanup yaml deps.
Currently minimal build requires `pyyaml`, but we use `pyaml` (which itself _depends_ on `pyyaml`) in the other test envs. Note that both import as `yaml`... Furtermore, `yaml` is a soft dependency: part of `on_demand_imports`, tests use `@requires_module("yaml")`), _but_ there's actually one place in the code base that we import it without going through on_demand_imports.py : https://github.com/yt-project/yt/blob/16cf684ca50a3ad034e8721b184a7609dd192772/yt/utilities/mesh_code_generation.py#L1
so I'd be inclined to think that it's de-facto already a hard dependency ?

### Cleanup pre-existing env targets ?
see https://github.com/yt-project/yt/blob/16cf684ca50a3ad034e8721b184a7609dd192772/setup.cfg#L61
apparently the `hub` target only adds one requirement (`girder_hub`) and is only used in parts of the command line tool. I'm not sure atm wether this is still used or tested for.

Same goes for the `bottle` target.

update : according to Kacper the first one is still used.

### Remove `tests/windows_conda_requirements.txt`
[...]

PR #3245, opened 2021-05-02
add the pytest-randomly pytest plugin to test env

neutrinoceros:
## PR Summary

This is based on top of #3236 and only add one line (last commit of this branch is unique)

pytest-randomly shuffles our tests' execution order to ensure that they don't rely on each others' possible side effect to pass.


neutrinoceros:
This showed that indeed we have at least one fragility in

yt.utilities.tests.test_config.TestYTConfigCommands testMethod=testConfigCommands

neutrinoceros:
switching to draft because I don't except to be able to work on this in the foreseeable future.

PR #3246, opened 2021-05-03
exp: refactor the on demand import module

### LINKED ISSUE #3237: Out of date requirements (unused optional deps ?)
### Bug report

**Bug summary**

~while working on #3236 I noticed that we still have a couple of optional deps that are not using the common mechanism from `on_demand_imports.py`~ (**this point is resolved in the first comments**) 
- pyaml
- firefly_api
- gluviz (imports as glue)
- mpi4py
- xarray

Also, some of our optional deps are out of date, in the sense that they do not seem to be used anymore:
- fastcache
- pyqt5 (I suspect this is meant to support Matplotlib GUI)
- pykdtree (weird case, it looks like we define this internally but it's still part of CI reqs ?)

neutrinoceros:
### Rationale:
Right now there's a lot a code duplication with slight variations in how the on-demand-import pattern is expressed and run.
Some packages are truly imported on-demand, i.e. the actual import is delayed to the first usage, but not all: other packages like h5py, pandas, requests... are imported at yt's import time if available, even when not needed.

This is an attempt to make _all_ our optional deps *true* on-demand resources, while hugely reducing the amount of copy-pasting in the module, and improving the maintainability of the module.
An extra benefit is that it reduces ever so slightly the import time for `yt` in an env that has many (all) of the optional deps available. 

The idea for this refactor came about during the discussion in #3237, so I consider this a fix because it removes the need for the issue
fix #3237

## PR Checklist
- [ ] pass CI
- [x] cleanup the special cases for Miniball and Cartopy
- [x] reuse new mechanism in some places that currently reimplement the pattern with a simple try block + custom error (yt_astro_analysis, thingking, bottle, f90nml, firefly_api, mpi4py, resource...)
- [ ] remove the `NotAModule` class (or deprecate it in case it's used downstream ?)


neutrinoceros:
This strategy is failing only for scipy, though it works on my machine, and updating scipy apparently doesn't help. More investigation required here.

yt/utilities/on_demand_imports.py
```diff
@@ -22,6 +22,58 @@ def __call__(self, *args, **kwargs):
...
+    """
+
+    def __init__(self, name: str):
+        if name.startswith("astropy"):
+            print(name)
+        self._name = name
+        self._module = None
+        self._submodules = {}
+
+    @property
```
cphyc, line 38: This would be a good usecase for `functools.cached_property`.

yt/utilities/on_demand_imports.py
```diff
@@ -231,359 +136,11 @@ def __version__(self):
...
-            except ImportError:
-                FullLoader = NotAModule(self._name)
-            self._FullLoader = FullLoader
-        return self._FullLoader
-
-
-_yaml = yaml_imports()
-
-
 class NotMiniball(NotAModule):
```
cphyc, line 139: What is special about Miniball?

yt/utilities/on_demand_imports.py
```diff
@@ -22,6 +22,58 @@ def __call__(self, *args, **kwargs):
...
+    "netCDF4",
+    "nose",
+    "pandas",
+    "pooch",
+    "requests",
+    "scipy",
+    "yaml",
+]
+for target in targets:
+    exec(f"{target} = OnDemandImport('{target}')")
```
cphyc, line 74: This should achieve the same thing as above, but imho is clearer
```suggestion
import sys
_module = sys.modules[__name__]
for target in targets:
    setattr(_module, target, OnDemandImport("target"))
```

See https://stackoverflow.com/questions/1429814/how-to-programmatically-set-a-global-module-variable for a discussion about doing this ;)

yt/utilities/on_demand_imports.py
```diff
@@ -37,167 +89,20 @@ def __init__(self, pkg_name):
             # yt-project/yt#1966
             self.error = ImportError(
                 "This functionality requires the %s "
-                "package to be installed." % self.pkg_name
+                "module to be installed." % self.pkg_name
```
cphyc, line 92: While we're at it, let's use f-strings!

yt/utilities/on_demand_imports.py
neutrinoceros: which was my original implementation actually, but it's only available from Python 3.8 :)

yt/utilities/on_demand_imports.py
neutrinoceros: The error message is unique. I left it for later but I'll get to it (and cartopy as well)

yt/utilities/on_demand_imports.py
cphyc: image

yt/utilities/on_demand_imports.py
neutrinoceros: Thanks. Indeed I was not pleased with using `exec` so I agree your solution is better. Right now I want to go across my TODO list but I'll include your suggestion before I remove the draft status !

yt/utilities/on_demand_imports.py
neutrinoceros: I intend to remove the `NotAModule` class in a later stage of this PR.
Normally the noun "package" is still relevant here because the error should trigger only on missing packages (which In consider a special class of modules).

yt/visualization/tests/test_geo_projections.py
```diff
@@ -130,7 +130,16 @@ def test_projection_object(self):
...
     def test_nondefault_transform(self):
-        from yt.utilities.on_demand_imports import _cartopy as cartopy
+        import sys
+
+        if sys.platform.startswith("win"):
+            # this test currently fails on Windows
+            # TODO: report this
+            from nose import SkipTest
+
+            raise SkipTest
```
neutrinoceros, line 140: This test consistently fails one windows, and exclusively on this branch, while the rest of the cartopy tests seem to be ok (to be confirmed), so I would conclude that the test is actually revealing an actual bug, but most likely not something I introduced in this branch. If skipping this only test indeed allows CI to pass, I'll open an issue from this comment.

neutrinoceros: Looks like I got the wrong idea. I'll try to skip _every_ cartopy test on windows just to see if that's really where the problem is.

neutrinoceros:
I think that to get this back on track I'll need to iteratively remove optional deps from the CI env until I find out which one is causing memory errors on windows. However this will be much simpler when #3236 is merged, so I'll wait for that to happen first.

neutrinoceros:
I think this branch is doomed... I'll reopen and push in small increments

PR #3248, opened 2021-05-04
Adds current_cycle attribute to EnzoP Dataset object

stefanarridge:
## PR Summary

Adds 'current_cycle' as an attribute for the EnzoP Dataset Object. It can come in handy when analysing a series of datasets.

## PR Checklist


matthewturk:
@stefanarridge As a quick follow-up, is this already accessible in the `.parameters` attribute of the EnzoP dataset object?

neutrinoceros:
Hi @stefanarridge ! I think that attaching code-specific data to a dataset as attributes should be avoided. We normally use their `parameters` attribute, which is a dictionary, to do just that. Is there a reason this is not satisfying to you ?

stefanarridge:
> @stefanarridge As a quick follow-up, is this already accessible in the `.parameters` attribute of the EnzoP dataset object?

No it isn't, as far as I can tell. `.parameters' just contains the parameters given in the parameter file, which will be the same for every dataset produced by a particular run. I want to be able to access the cycle number, which is different for each dataset.

matthewturk:
Interesting -- for original Enzo, we parse it for each individual dataset, so things like the cycle number will be updated.

I'm not sure about adding this as an attribute, but at the *very* least, the populating of the `.parameters` dict is something that should be looked at.  @brittonsmith have you taken a look at this?

stefanarridge:
> Hi @stefanarridge ! I think that attaching code-specific data to a dataset as attributes should be avoided. We normally use their `parameters` attribute, which is a dictionary, to do just that. Is there a reason this is not satisfying to you ?
> 
> edit: @matthewturk beat me to it :)

I suppose it could be added to the `parameters' attribute, although strictly speaking, the cycle number of a particular dataset is not a parameter of the simulation. I added it as a separate attribute so it is analagous to `current_time'. I understand your point about not wanting code-specific attributes.

neutrinoceros:
> No it isn't, as far as I can tell. .parameters just contains the parameters given in the parameter file, which will be the same for every dataset produced by a particular run. I want to be able to access the cycle number, which is different for each dataset.

This is _odd_. It might have to do with how `DatasetSeries` works but I doubt it, my biggest suspicion is that there is a mistake in the frontend. Wherever the problem lies, I think it should be considered a bug.

> I suppose it could be added to the parameters attribute, although strictly speaking, the cycle number of a particular dataset is not a parameter of the simulation.

Yeah well, many attributes of the Dataset classes have worn out their original name in my opinion (for instance, many of our frontends use `Dataset.parameter_file` to store the actual datafile). I am certainly not blaming you for thinking that wasn't appropriate :)

brittonsmith:
Hi @stefanarridge, thanks for looking at this. I do think this is probably better implemented as an entry in the `parameters` dict. It's true that in the enzo-p frontend that most of the entries in the parameters dict are parameters that pertain to the simulation as a whole, but I would call that a special case of how enzo-p records dataset and simulation parameters. I believe the spirit of the parameters dict is to contain any general metadata related to the dataset, whereas the dataset attributes should be limited to a select few that have general meaning across frontends.

stefanarridge:
> Hi @stefanarridge, thanks for looking at this. I do think this is probably better implemented as an entry in the `parameters` dict. It's true that in the enzo-p frontend that most of the entries in the parameters dict are parameters that pertain to the simulation as a whole, but I would call that a special case of how enzo-p records dataset and simulation parameters. I believe the spirit of the parameters dict is to contain any general metadata related to the dataset, whereas the dataset attributes should be limited to a select few that have general meaning across frontends.

Okay, I can look into implementing this instead but I'm not sure how straightforward it would be.

Also, does anyone know how to add a label to this PR? Google hasn't been much help here.

neutrinoceros:
> Also, does anyone know how to add a label to this PR? Google hasn't been much help here.

That's a privilege that belongs to the maintainers. Not that this is how Github works, it's not specific to this project.

> Okay, I can look into implementing this instead but I'm not sure how straightforward it would be.

If I'm reading @brittonsmith 's comment right, the change is as staighforward as it gets, and it should suffice to write something like

```python
self.parameters["current_cycle"] = ...
```

neutrinoceros:
pre-commit.ci autofix

yt/frontends/enzo_p/data_structures.py
```diff
@@ -405,6 +405,7 @@ def _parse_parameter_file(self):
         # all blocks are the same size
         ablock = fh[list(fh.keys())[0]]
         self.current_time = ablock.attrs["time"][0]
+        self.parameters["current_cycle"] = ablock.attrs["cycle"][0]
```
neutrinoceros, line 408: Can we assume this attribute will _always_ be present ?

neutrinoceros (commented):
I approve of this on the condition that EnzoP dataset are guaranteed to contain the `"cycle"` field, otherwise the implementation should allow for a fallback :)

brittonsmith (approved):

neutrinoceros (approved):

chummels (approved):

chummels:
It seems like the tests that are failing are two answer tests totally unrelated to this PR.  With ticks being flipped on the axis from MPL.  Can we just merge this?

neutrinoceros:
> It seems like the tests that are failing are two answer tests totally unrelated to this PR. With ticks being flipped on the axis from MPL. Can we just merge this?

That would be the issue I fixed in #3257
I'll relaunch the CI jobs and merge this asap !

neutrinoceros:
I don’t know why but it still failed.
@stefanarridge, would you be able to either merge the main branch into yours or rebase ?

stefanarridge:
> I don’t know why but it still failed.
> @stefanarridge, would you be able to either merge the main branch into yours or rebase ?

Just did it :)

brittonsmith:
When the tests pass, I will merge this.

PR #3250, opened 2021-05-08
fix a bug where yt.load_sample would crash against the default value for test_data_dir

### LINKED ISSUE #3227: yt.load_sample() broken on macOS since commit 900667ac
### Bug report

**Bug summary**

yt.load_sample() is broken since commit 900667ac "Merge branch 'main' into no_ambiguous_fields_automatic".

**Code for reproduction**

To reproduce, simply run the second cell of "yt/doc/source/quickstart/2)_Data_Inspection.ipynb".  Or,

```python
import yt
```

The output is

```
yt : [INFO     ] 2021-04-27 13:37:38,888 'IsolatedGalaxy' is not available locally. Looking up online.
yt : [INFO     ] 2021-04-27 13:37:38,889 Downloading from https://yt-project.org/data/IsolatedGalaxy.tar.gz
OSError                                   Traceback (most recent call last)
<ipython-input-2-38f11556a912> in <module>
----> 1 ds = yt.load_sample("IsolatedGalaxy")

~/yt/yt/loaders.py in load_sample(fn, progressbar, timeout, **kwargs)
   1353     filename = urlsplit(specs["url"]).path.split("/")[-1]
   1354 
-> 1355     tmp_file = _download_sample_data_file(
   1356         filename, progressbar=progressbar, timeout=timeout
   1357     )

~/yt/yt/sample_data/api.py in _download_sample_data_file(filename, progressbar, timeout)
    195 
    196     poochie = _get_pooch_instance()
--> 197     poochie.fetch(filename, downloader=downloader)
    198     return Path.joinpath(poochie.path, filename)

/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/pooch/core.py in fetch(self, fname, processor, downloader)
    518 
    519         # Create the local data directory if it doesn't already exist
--> 520         make_local_storage(str(self.abspath))
    521 
    522         url = self.get_url(fname)

/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/pooch/utils.py in make_local_storage(path, env)
    280             # try to create the path at the same time. Use exist_ok to avoid
    281             # raising an error.
--> 282             os.makedirs(path, exist_ok=True)
    283         else:
    284             action = "write to"

/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/os.py in makedirs(name, mode, exist_ok)
    211     if head and tail and not path.exists(head):
    212         try:
--> 213             makedirs(head, exist_ok=exist_ok)
    214         except FileExistsError:
    215             # Defeats race condition when another thread created the path

/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/os.py in makedirs(name, mode, exist_ok)
    211     if head and tail and not path.exists(head):
    212         try:
--> 213             makedirs(head, exist_ok=exist_ok)
    214         except FileExistsError:
    215             # Defeats race condition when another thread created the path

/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/os.py in makedirs(name, mode, exist_ok)
    211     if head and tail and not path.exists(head):
    212         try:
--> 213             makedirs(head, exist_ok=exist_ok)
    214         except FileExistsError:
    215             # Defeats race condition when another thread created the path

/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/os.py in makedirs(name, mode, exist_ok)
    221             return
    222     try:
--> 223         mkdir(name, mode)
    224     except OSError:
    225         # Cannot rely on checking for EEXIST, since the operating system

OSError: [Errno 30] Read-only file system: '/does'
```

**Expected outcome**

`pooch` is not able to create the directory "/does" due to permission of the root file system and raised an OSError.  Most likely this is not a bug on `pooch` itself but incorrect setting of the download file location.

If I check out an earlier version: `git checkout 0c892879`, then the error is not there.

**Version Information**

  * Operating System: macOS 11.3
  * Python Version: 3.8
  * yt version: develop version, this occurs starting at commit 900667ac.  I checked ~ 5 different commits after 900667ac and they show the same error.
  * Other Libraries (if applicable): pooch v1.3.0, although this is likely a problem on the yt side.

This was discovered on macOS 11.3 with most of the yt dependencies installed by conda, and yt installed from source with development mode.

neutrinoceros:
## PR Summary

fix #3227

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1336,15 +1336,7 @@ def load_sample(fn, progressbar: bool = True, timeout=None, **kwargs):
             mylog.info("Ignoring the `timeout` keyword argument received.")
         return load(loadable_path, **kwargs)
 
-    try:
-        save_dir = _get_test_data_dir_path()
-        assert save_dir.is_dir()
-    except (OSError, AssertionError):
-        mylog.warning(
-            "yt test data directory is not properly set up. "
-            "Data will be saved to the current work directory instead."
-        )
-        save_dir = Path.cwd()
+    save_dir = _get_test_data_dir_path()
 
     # effectively silence the pooch's logger and create our own log instead
     pooch_logger.setLevel(100)
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1323,20 +1323,19 @@ def load_sample(fn, progressbar: bool = True, timeout=None, **kwargs):
 
     kwargs = {**specs["load_kwargs"], **kwargs}
 
-    try:
-        data_dir = lookup_on_disk_data(fn)
-    except FileNotFoundError:
-        mylog.info("'%s' is not available locally. Looking up online.", fn)
-    else:
+    save_dir = _get_test_data_dir_path()
+
+    data_path = save_dir.joinpath(fn)
+    if data_path.exists():
         # if the data is already available locally, `load_sample`
         # only acts as a thin wrapper around `load`
-        loadable_path = data_dir.joinpath(specs["load_name"], specific_file)
-        mylog.info("Sample dataset found in '%s'", data_dir)
+        loadable_path = data_path.joinpath(specific_file or specs["load_name"])
+        mylog.info("Sample dataset found in '%s'", data_path)
         if timeout is not None:
             mylog.info("Ignoring the `timeout` keyword argument received.")
         return load(loadable_path, **kwargs)
 
-    save_dir = _get_test_data_dir_path()
+    mylog.info("'%s' is not available locally. Looking up online.", fn)
 
     # effectively silence the pooch's logger and create our own log instead
     pooch_logger.setLevel(100)
--- a/yt/sample_data/api.py
+++ b/yt/sample_data/api.py
@@ -129,10 +129,20 @@ def get_data_registry_table():
     return unified_table
 
 
+@lru_cache(maxsize=128)
 def _get_test_data_dir_path():
+    # the caching is mostly useful to avoid repeating the warning
     from yt.config import ytcfg
 
-    return Path(ytcfg.get("yt", "test_data_dir"))
+    p = Path(ytcfg.get("yt", "test_data_dir"))
+    if p.is_dir():
+        return p
+    mylog.warning(
+        "Storage directory from yt config doesn't exist "
+        f"(currently set to '{p}'). "
+        "Current working directory will be used instead."
+    )
+    return Path.cwd()
 
 
 def lookup_on_disk_data(fn):


neutrinoceros:
Reviving a comment from #3195 by @matthewturk 
> Ah, so what I mean is that pooch uses the userdirs package to come up with default locations. So, if left unspecified, the download location is set by XDG_CACHE or the default-for-that-system value for XDG_CACHE. This way it is still a user-specific, systemwide location that will be identical across runs independent of working dir. (i.e., if I cd to a subdir and re-run load_sample, it won't re-download)

I've actually tried to do this but it seems to me that pooch doesn't allow the target path to be left unspecified.

neutrinoceros:
@Xarthisius I think there may be side effects in our CI (Jenkins) that I did not anticipate. Does any of the following datasets happen to be present in the dir from which Jenkins tests are run ?
- ToroShockTube
- IsolatedGalaxy

Xarthisius:
> @Xarthisius I think there may be side effects in our CI (Jenkins) that I did not anticipate. Does any of the following datasets happen to be present in the dir from which Jenkins tests are run ?
> 
> * ToroShockTube
> * IsolatedGalaxy

All datasets are available in test_data_dir on Jenkins. Also I'm noting that none of the answer test on Jenkins failed so why does it matter?

Xarthisius:
We don't want to download datasets every time the test are run , do we? Can you mock it instead?

neutrinoceros:
> We don't want to download datasets every time the test are run , do we? Can you mock it instead?

actually I'd argue it's preferable to actually download them... to an isolated place. The tests are already pretty fragile and I think that adding a mocking layer would make them both useless and unmaintainable. (I'm limiting this kind of test to very small datasets <150KB).

Anyway it _shouldn't_ be an issue if the datasets happen to be present on the Jenkins server, but I'd like to know if it is the case, because if it's not I have no clue how to fix this PR.
On the other hand, if that's what's happening, it probably means I need to "nerf" yt.load_sample so that it only looks for data in the current dir _if_ test_data_dir is unset (now that I think of it it's probably the desirable behaviour anyway).

neutrinoceros:
> All datasets are available in test_data_dir on Jenkins. Also I'm noting that none of the answer test on Jenkins failed so why does it matter?

Oops sorry I didn't see you answered up there !
That's what I expect, but I wonder if somehow we also have some datasets in the `cwd` where tests are run. I'm already mocking the `test_data_dir` but the content of `cwd` also change the behaviour of the tested functionality.

Xarthisius:
FTR this is what happens in the job if you want to try to reproduce locally:

```
git clean -f -d -x yt

export LC_ALL=en_US.utf8
export LANG=en_US.utf8
export OMP_NUM_THREADS=8
export TMPDIR=$(mktemp -d)
export XDG_CONFIG_HOME=$(mktemp -d)
mkdir -p "${XDG_CONFIG_HOME}/matplotlib"
cp tests/matplotlibrc "${XDG_CONFIG_HOME}/matplotlib"

echo $(date) > jenkins_is_stupid

export PYTHONPATH=$PWD
export YT_BUILD=$PWD
export PATH=$HOME/.local/bin:$PATH

sed -i -e '/test_data_dir/ s:/does/not/exist:/mnt/yt:g' \
    -e '/xray_data_dir/ s:/does/not/exist:/mnt/yt/xray_data:g' \
    -e '/supp_data_dir/ s:/does/not/exist:/mnt/yt/supp_data:g' \
    -e '/strict_requires/ s/False/True/g' \
    -e '/suppress_stream_logging/ s/False/True/g' yt/config.py

sed -i -e '/^where=/d' *.cfg
sed -e '/#html_last_updated_fmt/ s/#//' -i doc/source/conf.py
sed -i -e '/^SPHINXBUILD/ s/sphinx-build/& -j 8/g' doc/Makefile
sed -e '/download_datasets =/ s/True/False/' \
    -i doc/source/quickstart/1\)_Introduction.ipynb
sed -e '/sys.stderr/d' -e 's/mystderr.getvalue()/""/' -i tests/nose_runner.py

echo "/usr/local/src/rockstar" > rockstar.cfg
echo "/usr/" > embree.cfg

[[ -f tests/test_prerequirements.txt ]] && pip install -r tests/test_prerequirements.txt
[[ -f tests/test_requirements.txt ]] && pip install -r tests/test_requirements.txt
[[ -f tests/lint_requirements.txt ]] && pip install -r tests/lint_requirements.txt

export CFLAGS="-Wno-cpp -fno-strict-aliasing -O3 -march=native -pipe -g -ggdb"
export CXXFLAGS=$CFLAGS

[[ -f tests/test_requirements.txt ]] && python setup.py build_ext -i -f
[[ ! -f tests/test_requirements.txt ]] && python -m pip install -e .[test,full]

export NUM_WORKERS=20
echo $VIRTUAL_ENV

pytest --junit-xml=unittests.xml --pyargs yt -n 4 --dist=loadscope

rm -rf $TMPDIR $XDG_CONFIG_HOME
```

It assumes you are executing that in a virtualenv.

neutrinoceros:
Closing in favor of #3262

neutrinoceros:
Closing in favor of #3262

PR #3257, opened 2021-05-13
CI: hotfix failing answer tests

### LINKED ISSUE #3256: Failing answer tests
### Bug report

**Bug summary**

Currently 2 answer tests are failing on the main branch (since mpl 3.4.2 was released last week).
Here are their names:
`GenericImage_UniformGridData_all_line_answers_multi_line_plot_None_None`
`GenericImage_UniformGridData_all_line_answers_line_plot_None_None`

I don't understand where they are being generated so it's pretty much impossible for me to start fixing them atm.

The bug is that axis ticks are inverted somehow (possibly a bug upstream ?)

Screenshot 2021-05-12 at 21 38 58

neutrinoceros:
## PR Summary

fix #3256 (?)
For now I'm just commenting out two tests functions that I strongly suspect are the ones generating the failing answers.
If CI validates my suspicions I'll be able to start digging into it for real.

update for posterity:
since I've rebased and force pushed a different fix, there is no trace left of the tests functions I was referring to, so here they are:
- `yt/visualization/tests/test_line_plots.test_line_plot`
- `yt/visualization/tests/test_line_plots.test_multi_line_plot`
eg the only tests that use the `LinePlot` class (that I'm aware of)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ project_urls =
 packages = find:
 install_requires =
     IPython>=1.0
-    matplotlib>=2.0.2,<3.5
+    matplotlib!=3.4.2,>=2.0.2,<3.5
     more-itertools>=8.4
     numpy>=1.10.4
     setuptools>=19.6
--- a/tests/test_requirements.txt
+++ b/tests/test_requirements.txt
@@ -5,7 +5,7 @@ fastcache~=1.0.2
 glueviz~=0.13.3
 h5py~=3.1.0
 ipython~=7.6.1
-matplotlib<3.5
+matplotlib<3.5,!=3.4.2
 nose-timer~=1.0.0
 nose~=1.3.7
 pandas~=1.1.2


neutrinoceros:
Looks like I was right, now let's bisect it down.

neutrinoceros:
Good news, the issue seem to lie in an annotation method, which narrows it down much better than if it was in yt.LinePlot.__init__

neutrinoceros:
After a couple iterations I established that calling _any_ of the following methods causes the ticks to be inverted:

- `LinePlot.annotate_legend`
- `LinePlot.set_x_unit`
- `LinePlot.set_unit`
- `LinePlot.annotate_title`

The common denominator is that all these use the `@invalidate_plot` decorator, which forces `self._plot_valid = False`.
Then I need to understand when that invalidation is going to affect the result and how, and it seems to quickly get a lot messier, as the dataflow goes back and forth between several classes and `_setup_plot` methods. It would be extremely valuable to narrow it down to the part where we force inward ticks, but atm I can't find it.

neutrinoceros:
I was able to confirm that the regression is indeed upstream (reported as #20219)
So I will patch this by updating our requirements (forbid 3.4.2)

brittonsmith (approved):

PR #3258, opened 2021-05-13
New fields for Arepo and Gadget frontends

jzuhone:
## PR Summary

This PR adds new fields for both the Arepo and Gadget frontends.

* For Arepo: adds the ``"CosmicRaySpecificEnergy"`` field and ``"PassiveScalars"``
* For Gadget: adds specifications to support Magneticum halos (http://www.magneticum.org) and known fields
* For Gadget: If we already have a temperature field in the dataset, there is no need to create a derived field for it. This was technically a bug. 

EDIT: This got more complicated. It turns out that if you naively run `sph_answer` on the Magneticum dataset that you get an error because you project a sphere at the domain center with no data in it (the data is confined to a small region of the domain). The error will need to be fixed in a separate PR. 

It would have been simple to set a bounding box around the data, but when I tried this it took forever to construct the index. So I did a simple (and probably useful in the future thing) and allowed the sphere in `sph_answer` and `nbody_answer` to be set to some alternate center.

## PR Checklist


jzuhone:
pre-commit.ci autofix

jzuhone:
I added more doc edits--it turns out we had a number of things in Gadget and Arepo data that needed explaining in the docs. 

For the tests to pass this needs the dataset which will be part of https://yt-project.org/data once PR https://github.com/yt-project/website/pull/96 is merged (ping @Xarthisius) and I upload new answers.

yt/frontends/gadget/tests/test_outputs.py
```diff
@@ -124,3 +125,25 @@ def test_bigendian_field_access():
     ds = data_dir_load(BE_Gadget)
     data = ds.all_data()
     data["Halo", "Velocities"]
+
+
+mag_fields = OrderedDict(
```
neutrinoceros, line 130: quick note: since Python 3.6, builtin dict objects are also ordered. At the time it was deemed an implementation detail, but it's been embraced as a feature since.
Some minor differences persist (see https://stackoverflow.com/questions/34305003/difference-between-dictionary-and-ordereddict), but it's doubtful we need them here

jzuhone: So if you look at the top of the file:

https://github.com/yt-project/yt/blob/0c0ad20e93750e34b60ac9863098c9e561ac5ec6/yt/frontends/gadget/tests/test_outputs.py#L27-L35

We have another one of these in there, which I copied to make the new one below. For some reason there are two `("gas","temperature")` entries in here--`OrderedDict` simply replaces the first one with the second one silently without complaining. When I tried to change this to a `dict` the pre-commit bot complained. 

So the real question is if `sph_answer` is expecting us to try projections with and without the weight field for `("gas","temperature")`. Does anyone know (I can look myself, but if anyone knows please comment)? @matthewturk @brittonsmith @chummels

chummels: It seems to me taking an unweighted projection of `("gas", "temperature")` would be useless and non-sensical, so I don't think `sph_answer` should do that.

matthewturk: It probably is expecting that.  A non-weighted temperature projection is generally not useful, but it can also serve as a proxy for testing in situations where the temperature itself is a derived field (rather than intrinsic) and we want to test the machinery of unweighted temperature projections.  I don't know that we need to keep it in this instance, but I don't think we should *always* require that answer tests always map 1:1 to useful quantities.

matthewturk (approved):
LGTM; @jzuhone what is pending still?

jzuhone:
@matthewturk need to fix the pre-commit issue, update test answers

jzuhone:
@Xarthisius it looks like I screwed up the answer testing--can we remove `local_gadget_007` on the build box?

Xarthisius:
@yt-fido test this please

jzuhone:
This is finally ready.

yt/sample_data_registry.json
```diff
@@ -259,6 +259,15 @@
     "load_name": "plt00015",
     "url": "https://yt-project.org/data/Laser.tar.gz"
   },
+  "MagneticumCluster.tar.gz": {
```
neutrinoceros, line 262: in order for this to integrate with yt.load_sample, you'll need to update https://github.com/yt-project/website/blob/master/data/datafiles.json

jzuhone: @neutrinoceros This was done already (you looked at the PR): https://github.com/yt-project/website/pull/96

neutrinoceros: I didn't realise it was the same one. In any case I've tested this branch and it currently doesn't work
```python
import yt
yt.load_sample("MagneticumCluster")
```

jzuhone: Do you have a traceback or something for more information?

yt/sample_data_registry.json
```diff
@@ -259,6 +259,15 @@
...
+    "hash": "403c2224c205a44ee92a5925053be8aef6e5d530d9297c18151cdb0f588c32a7",
+    "load_kwargs": {
+      "long_ids": true,
+      "field_spec": "magneticum_box2_hr"
+    },
+    "load_name": "snap_132",
+    "url": "https://yt-project.org/data/Laser.tar.gz"
```
neutrinoceros, line 269: ah, I think that's the real problem here: looks like a copy-pasted url :)

neutrinoceros, line 269: most likely (I'll test this locally now)
```suggestion
    "url": "https://yt-project.org/data/MagneticumCluster.tar.gz"
```

jzuhone: Thanks for catching it--now it works for me

chummels (approved):
In general this looks good; I have just a few comments inline.  Thanks for putting some love into the AREPO/Gadget frontends.  It's good that they're getting a bit more attention with the yt4 refactor.

yt/frontends/gadget/fields.py
```diff
@@ -65,31 +67,107 @@ def _Metal_density(field, data):
...
-                x_H = _primordial_mass_fraction["H"]
-                gamma = 5.0 / 3.0
-                a_e = data[ptype, "ElectronAbundance"]
-                mu = 4.0 / (3.0 * x_H + 1.0 + 4.0 * x_H * a_e)
-                ret = data[ptype, "InternalEnergy"] * (gamma - 1) * mu * mp / kb
-                return ret.in_units(self.ds.unit_system["temperature"])
+        This gets used with the magneticum_box2_hr format
+        as defined in the gadget_field_specs in frontends/gadget/definitions.py
+        """
+        metal_names = ["He", "C", "Ca", "O", "N", "Ne", "Mg", "S", "Si", "Fe", "Ej"]
```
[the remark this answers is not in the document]
jzuhone: Yes, that's correct--that's the name that the magneticum folks use for it.

yt/frontends/gadget/fields.py
```diff
@@ -65,31 +67,107 @@ def _Metal_density(field, data):
...
+
+            self.add_field(
+                (ptype, metal_name + "_density"),
+                sampling_type="particle",
+                function=_Density_wrap(i),
+                units=self.ds.unit_system["density"],
+            )
+
+        # hydrogen fraction and density
+        def _h_fraction(field, data):
```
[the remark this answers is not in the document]
jzuhone: Yes, that's the way they are defined in magneticum.

yt/frontends/gadget/fields.py
```diff
@@ -65,31 +67,107 @@ def _Metal_density(field, data):
...
         )
+
+    def setup_gas_particle_fields(self, ptype):
+        if (ptype, "Temperature") not in self.ds.field_list:
+
+            if (ptype, "ElectronAbundance") in self.ds.field_list:
+
+                def _temperature(field, data):
+                    # Assume cosmic abundances
+                    x_H = _primordial_mass_fraction["H"]
```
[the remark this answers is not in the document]
jzuhone: Hmm. we need to think about this one a bit more carefully. I didn't change the field definition--I just moved it within a new if-then block. But I see what you mean. I'll try changing it, but I suspect it will break answer tests.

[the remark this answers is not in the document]
chummels: @jzuhone pointed me to the [TNG documentation](https://www.tng-project.org/data/docs/faq/#gen5) where they themselves are calculating temperature using this formula, with primordial hydrogen abundances.  So we should mimic that behavior and I retract my point above.

PR #3261, opened 2021-05-14
fix: resolve #3260

### LINKED ISSUE #3260: data sample KeplerianDisk doesn't load with load_sample
### Bug report

**Bug summary**

**Code for reproduction**
Currently this only partially works: the dataset is correctly downloaded but can't load, even on retry (which goes through a different branch)
```python
yt.load_sample("KeplerianDisk/disk.out1.00000.athdf")
```
Luckily this dataset is small enough that we can simply add a test for it.

Note that in an interactive session, the dataset can still be loaded as
```python
>>> yt.load_sample("KeplerianDisk/disk.out1.00000.athdf")
>>> yt.load("KeplerianDisk/disk.out1.00000.athdf")
```
So I think it's easy enough to fix.

edit:
So actually the following version already works fine on the main branch
```python
yt.load_sample("KeplerianDisk")
```
but I'm only realising this now that I have a fix that works for both

neutrinoceros:
## PR Summary
fix a bug in yt.load_sample where a dataset could be downloaded and still failed to load
fix #3260

edit: this will very likely conflict with #3250, so I'll try to rebase on that branch when it's stable. For now #3250 has priority.

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1299,6 +1299,10 @@ def load_sample(fn, progressbar: bool = True, timeout=None, **kwargs):
 
     pooch_logger = pooch.utils.get_logger()
 
+    if isinstance(fn, str):
+        # normalize path for platform portability
+        fn = fn.replace("/", os.path.sep)
+
     topdir, _, specific_file = str(fn).partition(os.path.sep)
 
     registry_table = get_data_registry_table()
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Ah. The new test is also revealing that calling load_sample with "/" separated paths breaks on windows. I will fix this by normalizing those to a platform specific sep, which I believe is the most sensible thing to do as it will make scripts using load_sample more portable

neutrinoceros:
switching to draft until I manage to pass CI on windows

neutrinoceros:
Closing in favor of #3262

PR #3262, opened 2021-05-14
multiple fixes for load_sample

### LINKED ISSUE #3227: yt.load_sample() broken on macOS since commit 900667ac

### LINKED ISSUE #3260: data sample KeplerianDisk doesn't load with load_sample

### LINKED ISSUE #3296: Enzo_64 dataset not loadable by load_sample()
### Bug report

As I was updating the quickstart notebooks, I came upon a failure for `load_sample()` being unable to load the `Enzo_64` dataset.  See [here](https://yt-project.org/docs/dev/quickstart/simple_visualization.html).  When I attempt this command locally, yt automatically downloads and untars the dataset, but it seems like it's trying to look at a filename called `None`.  No problems in loading the dataset manually with `load()`.

**Code for reproduction**

```python
ds = yt.load_sample("Enzo_64", "DD0043/data0043")
s = yt.SlicePlot(ds, "z", ["density", "velocity_magnitude"], center="max")
s.set_cmap("velocity_magnitude", "kamae")
s.zoom(10.0)
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-12-7024538d34af> in <module>
----> 1 ds = yt.load_sample("Enzo_64", "DD0043/data0043")
      2 s = yt.SlicePlot(ds, "z", ["density", "velocity_magnitude"], center="max")
      3 s.set_cmap("velocity_magnitude", "kamae")
      4 s.zoom(10.0)

/var/jenkins_home/workspace/yt_docs/yt/loaders.py in load_sample(fn, progressbar, timeout, **kwargs)
   1363         if timeout is not None:
   1364             mylog.info("Ignoring the `timeout` keyword argument received.")
-> 1365         return load(loadable_path, **kwargs)
   1366 
   1367     try:

/var/jenkins_home/workspace/yt_docs/yt/loaders.py in load(fn, *args, **kwargs)
     78     # either in the current dir or yt.config.ytcfg['data_dir_directory']
     79     if not fn.startswith("http"):
---> 80         fn = str(lookup_on_disk_data(fn))
     81 
     82     candidates = []

/var/jenkins_home/workspace/yt_docs/yt/sample_data/api.py in lookup_on_disk_data(fn)
    162     if not alt_path.exists():
    163         err_msg += f"\n(Also tried '{alt_path}')."
--> 164         raise FileNotFoundError(err_msg)
    165 
    166     return alt_path

FileNotFoundError: No such file or directory: '/mnt/yt/Enzo_64/None'.
(Also tried '/mnt/yt/Enzo_64/None').
```

neutrinoceros:
## PR Summary

Since #3250 and #3261  require tweaking the same lines, this is an attempt to combine them early rather than later when I don't remember them and I can't manage their conflicts.

fix #3227
fix #3260
fix #3296

cool side effect:
now load_sample works in a cross-platform way when requesting specific files as paths (using "/" a sep)

## TODO:
- [x] update docs on the newly formed design
- [x] rebase


neutrinoceros:
All CI jobs passed, so I'm closing the PRs this supersedes. I'll add some documentation and maybe a test or two

neutrinoceros:
pre-commit.ci autofix

yt/loaders.py
```diff
@@ -1263,7 +1264,9 @@ def flatten(l):
 
 
 # --- Loader for yt sample datasets ---
-def load_sample(fn=None, progressbar: bool = True, timeout=None, **kwargs):
+def load_sample(
+    fn: Optional[str] = None, progressbar: bool = True, timeout=None, **kwargs
```
matthewturk, line 1268: Genuine question: why are we marking this optional if it's also a kwarg, and if the others are not explicitly marked optional?

yt/loaders.py
```diff
@@ -1292,9 +1295,15 @@ def load_sample(fn=None, progressbar: bool = True, timeout=None, **kwargs):
...
         requests.get via pooch.HTTPDownloader
 
-    Any additional keyword argument is passed down to `yt.load`.
-    Note that in case of collision with predefined keyword arguments as set in
+    Notes
+    -----
+    - Any additional keyword argument is passed down to `yt.load`.
+    - In case of collision with predefined keyword arguments as set in
     the data registry, the ones passed to this function take priority.
+    - Datasets with slashes '/' in their names can safely be used even on Windows.
```
matthewturk, line 1303: did not know this...

Now what if we named a dataset `CON`?  :)

yt/sample_data/api.py
```diff
@@ -112,9 +113,8 @@ def get_data_registry_table():
     )
 
     # load local data
-    pooch_json = json.load(
-        pkg_resources.resource_stream("yt", "sample_data_registry.json")
-    )
+    with pkg_resources.resource_stream("yt", "sample_data_registry.json") as fh:
```
matthewturk, line 116: Since we are refactoring, is it worth getting rid of the loading completely and using pd to parse it from JSON to begin with?  I believe we could just use `pd.read_json` on the object returned by the resource stream.

(Not necessary)

yt/sample_data/api.py
```diff
@@ -132,7 +132,15 @@ def get_data_registry_table():
 def _get_test_data_dir_path():
     from yt.config import ytcfg
 
-    return Path(ytcfg.get("yt", "test_data_dir"))
+    p = Path(ytcfg.get("yt", "test_data_dir"))
+    if p.is_dir():
+        return p
+    warn(
+        "Storage directory from yt config doesn't exist "
```
matthewturk, line 139: This is an improvement.  I still think that I have a slight preference for allowing pooch to set this, but I am not going to push it, as this may genuinely be better for the use cases commonly found.

matthewturk (approved):
minor comments, none necessary for merging

yt/loaders.py
neutrinoceros: Alas, `typing.Optional[str]` strictly means `str | None` (where "|" reads as "or"). The later syntax is only valid from Python 3.10 (not released) though. Arguably this is a design mistake from early versions of the `typing` module.

yt/loaders.py
neutrinoceros: > did not know this...

Well it is only true _with_ to the present PR :)

> Now what if we named a dataset CON? :)

I don't get it :(

yt/sample_data/api.py
neutrinoceros: I've inspected pooch's source code and as far as I understand, it _requires_ that the storage dir be specified explicitly.
We could propose a patch upstream though, but it's not feasible in this PR, unless I'm mistaken.

yt/sample_data/api.py
neutrinoceros: I haven't found a way to do this. The format used for this json file (mandated by pooch) is a little tricky to get into pandas, hence my 3 lines where one _should_ be enough. I'm happy to integrate a change here if anyone can figure it out though

matthewturk: nah

yt/loaders.py
matthewturk: https://help.interfaceware.com/v6/windows-reserved-file-names

neutrinoceros:
> Huh, I thought for sure it would default to XDG_CACHE.  OK.

Now's the time. I also agree it's preferable. I can implement it now

matthewturk:
@neutrinoceros no wait, don't switch to XDG_CACHE.  Leave it as is.  CWD is better now that I've reflected on it.  Please don't change it!

neutrinoceros:
Ok I won't

yt/loaders.py
```diff
@@ -1298,9 +1301,15 @@ def load_sample(fn=None, progressbar: bool = True, timeout=None, **kwargs):
...
+      On the contrary, paths using backslashes '\' won't work outside of Windows, so
+      it is recommended to favour the UNIX convention ('/') in scripts that are meant
+      to be cross-platform.
```
chummels, line 1312: Would it be possible to include a short description of what `load_sample` does here in the docstring as well as what format the filenames should take.  Yes, they're strings, but for `IsolatedGalaxy` it seems like that is sufficient to load that dataset, whereas `Enzo_64` one needs to specify the specific dataset within that directory fully.  Maybe also have the URL of the actual location where the datasets are?  Or that what the behavior is when you specify no arguments? I could do this as a separate PR if need be.

chummels (approved):
Looks good to me.  One small suggestion on docstrings.

yt/loaders.py
neutrinoceros: The specificity in the Enzo_64 tarball is that there's a whole time series in there. We could choose one specific snapshot that would load if nothing else is supplied. I don't want to add to much detail here because I don't want to let users think that the design is 100% intentional. I'll try to clarify a bit though.

neutrinoceros: I intentionally avoided going too much into the details, but here's an attempt to satisfy your question !

yt/loaders.py
```diff
@@ -1331,9 +1333,18 @@ def load_sample(fn=None, progressbar: bool = True, timeout=None, **kwargs):
...
+    - This function is experimental as of yt 4.0.0, do not rely on its exact behaviour.
+    - Any additional keyword argument is passed down to `yt.load`.
+    - In case of collision with predefined keyword arguments as set in
     the data registry, the ones passed to this function take priority.
+    - Datasets with slashes '/' in their names can safely be used even on Windows.
+      On the contrary, paths using backslashes '\' won't work outside of Windows, so
+      it is recommended to favour the UNIX convention ('/') in scripts that are meant
+      to be cross-platform.
+    - This function is requires pandas and pooch.
+    - Corresponding data lives in https://yt-project.org/data
```
chummels, line 1347:
```suggestion
    - This function requires pandas and pooch.
    - Corresponding sample data live at https://yt-project.org/data
```

yt/loaders.py
```diff
@@ -1302,20 +1303,21 @@ def flatten(l):
...
     data with pooch from remote source.
 
-    yt sample data can be found at:
-    https://yt-project.org/data.
-
     The data registry table can be retrieved and visualized using
     `yt.sample_data.api.get_data_registry_table()`.
-
-    This function requires pandas and pooch to be installed.
+    The `filename` column therin contains usable keys that can be passed
```
chummels, line 1317:
```suggestion
    The `filename` column contains usable keys that can be passed
```

chummels (approved):
Just some simple typo fixes.  Thanks for adding to the docstrings!  Looks great!

neutrinoceros:
@matthewturk requested a review from @Xarthisius, but this is at two approvals already. I'll let you guys decide wether the additional review is needed.

PR #3263, opened 2021-05-14
Load sample err

chummels:
Consistent with the discussion in slack and with the [docs](https://yt-project.org/docs/dev/examining/loading_data.html), I recently tried to run:

```
import yt
yt.load_sample()
```

and encountered an error.  Thus, I'm making it so one *can* run this command and it gives the user some information on how to see all of the available datasets.


yt/loaders.py
```diff
@@ -1291,6 +1296,11 @@ def load_sample(fn, progressbar: bool = True, timeout=None, **kwargs):
     the data registry, the ones passed to this function take priority.
     """
 
+    if fn is None:
+        print("One can see which sample datasets are available at: https://yt-project.org/data")
+        print("or alternatively by running: yt.sample_data.api.get_data_registry_table()")
+        return
```
neutrinoceros, line 1302: The print statements should be merged and print to sys.sidéré
The return statement should be an explicit None

neutrinoceros (commented):
expedite review (on my phone)

yt/loaders.py
chummels: I thought it was generally bad form to have code lines that extended beyond 80 characters.  Having these lines merged would do so.

I don't know what sys.sidéré is.

neutrinoceros: > I thought it was generally bad form to have code lines that extended beyond 80 characters. Having these lines merged would do so.

our limit is 88, and there is a way to merge strings without going over it :)

> I don't know what sys.sidéré is.
that what I get for reviewing on my phone...

neutrinoceros, line 1302: Here's an actual suggestion :)
```suggestion
    if fn is None:
        print(
            "One can see which sample datasets are available at: https://yt-project.org/data\n"
            "or alternatively by running: yt.sample_data.api.get_data_registry_table()",
            file=sys.stderr
        )
        return None
```

neutrinoceros: to avoid dupplicating data between the message and the docstring, we can also do this:
```suggestion
    if fn is None:
        print(load_sample.__doc__, file=sys.stderr)
        return None
```

chummels: Hahahhaha...:)

neutrinoceros:
@chummels , pre-commit-ci is showing a missing import + two other problems that can be auto fixed

chummels:
I don't understand why the pre-commit is flagging this as not mergeable.  I've rebased this with "main" and it checks out.

neutrinoceros:
There is indeed a merge conflict left to solve, the failure in pre-commi.ci is expected.

chummels:
pre-commit.ci autofix

neutrinoceros (approved):

PR #3264, opened 2021-05-14
Small doc fixes

chummels:
I noticed on the https://yt-project.org/docs/dev/examining/loading_data.html docs page that there were a bunch of instances of code text that was in italics instead of in code font.  This happens with rst files if you surround your text with a single set of `s instead of two ``s.

I went ahead and corrected all of this issues in that document.


matthewturk (approved):

munkm (approved):

neutrinoceros:
there's a pre-commit hook for this, I suggest we seize the opportunity to try it out 
https://github.com/pre-commit/pygrep-hooks (rst-backticks therein)

chummels:
Whoa--I guess I'm not really familiar with pre-commit hooks.  So you're saying there is some external program that will search all the docs and replace single ` instances with ``?  There *are* some cases where you want single `'s, like for references and such.  But maybe the precommit hooks account for that?

neutrinoceros:
> Whoa--I guess I'm not really familiar with pre-commit hooks. So you're saying there is some external program that will search all the docs and replace single instances with ``? There *are* some cases where you want single's, like for references and such. But maybe the precommit hooks account for that?

It's not an auto-fixer, it will only report them, and from what I saw, there was no false positive in the report :)

PR #3265, opened 2021-05-14
Accounting for AREPO datasets to be correctly identified.

chummels:
I currently have an AREPO dataset from the AURIGA simulation suite, and it is failing to get loaded in.  It appears that one of the requirements for the AREPO frontend is that all datasets must have a CONFIG group in their HDF5 file.  This one does not.  However, any datasets that have fields beginning with `GFM_` (for Galaxy Formation Model) *are* AREPO datasets, so I'm using that as an alternate hook for ID'ing AREPO datasets.

I have written to Dylan Nelson from TNG and Freeke van de Voort from Auriga to confirm that GFM fields mean AREPO, so this should not be merged until I report back here.

UPDATE1: I received word back from Dylan that GFM fields are only present in Auriga, SURGE, and TNG datasets (all part of AREPO), so this should be accurate and ready for review.

### DIFF
--- a/yt/frontends/arepo/data_structures.py
+++ b/yt/frontends/arepo/data_structures.py
@@ -55,7 +55,7 @@ class ArepoHDF5Dataset(GadgetHDF5Dataset):
                     or "AMR" in fh["/Config"].attrs.keys()
                 )
                 # Datasets with GFM_ fields present are AREPO
-                or any(True for field in fh["/PartType0"] if field.startswith("GFM_"))))
+                or any(field.startswith("GFM_") for field in fh["/PartType0"])))
             fh.close()
         except Exception:
             valid = False
--- a/yt/frontends/arepo/data_structures.py
+++ b/yt/frontends/arepo/data_structures.py
@@ -47,13 +47,14 @@ class ArepoHDF5Dataset(GadgetHDF5Dataset):
         valid = True
         try:
             fh = h5py.File(filename, mode="r")
-            valid = (
+            valid = ((
                 all(ng in fh["/"] for ng in need_groups)
                 and not any(vg in fh["/"] for vg in veto_groups)
                 and (
                     "VORONOI" in fh["/Config"].attrs.keys()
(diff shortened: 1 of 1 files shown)


yt/frontends/arepo/data_structures.py
```diff
@@ -47,14 +47,15 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+            valid = ((
                 all(ng in fh["/"] for ng in need_groups)
                 and not any(vg in fh["/"] for vg in veto_groups)
                 and (
                     "VORONOI" in fh["/Config"].attrs.keys()
                     or "AMR" in fh["/Config"].attrs.keys()
                 )
-            )
+                # Datasets with GFM_ fields present are AREPO
+                or any(True for field in fh["/PartType0"] if field.startswith("GFM_"))))
```
neutrinoceros, line 58: I don't think it does what you think it does. If I'm reading this right it will *always* evaluate to `True`.
I believe _this_ is what you want
```suggestion
                or any(field.startswith("GFM_") for field in fh["/PartType0"])))
```

chummels: I checked and it seemed to work for both AREPO and non-AREPO gadget datasets.  As far as I can tell, our logic is the same.  But I'm fine using your option because it is more succinct.

neutrinoceros: oh actually you're right, I misread your original implementation. Yeah it would also work, but indeed my solution is more concise and "pythonic", probably. Anyway, glad you like it.

neutrinoceros:
@chummels if you're done here, you can run pre-commit either locally or via a comment to fix style issues.

chummels:
pre-commit.ci autofix

jzuhone (approved):

chummels:
I'm done tweaking this PR.  I'll make any further arepo frontend mods in another one, so it doesn't get confusing.

PR #3270, opened 2021-05-16
RFC: Refactor on-demand-import mechanism

### LINKED ISSUE #3237: Out of date requirements (unused optional deps ?)

neutrinoceros:
## PR Summary

This is another attempt at refactoring the on-demand-import mechanism (supersedes #3246).
The original branch has become virtually impossible to fix, so I'm retrying here with incremental updates.

fix #3237 

## TODO
- [x] migrate every special classes to the new OnDemandImport mechanism
- [x] drop `_` prefixes in on demand import names (reason: right now the prefix is effectively ignored in import statements via the `as` keyword, so there's a precedent of importing stuff from the on demand module as drop-in replacements)
- [x]  generalise new mechanism to some optional deps that are currently not using the on-demand-import module (bottle, firefly_api, mpi4py, resource, yt_astro_analysis)
- [ ] cleanup (remove) the `NotAModule` class. should be trivial after the previous step is checked (NOT APPLICABLE, see comment bellow)
- [x] cleanup on_demand_imports.py (make the `_targets` list a dictionary, and maybe other nitpicks)
- [x] add documentation (programming guide).

### DIFF
--- a/doc/source/developing/index.rst
+++ b/doc/source/developing/index.rst
@@ -27,3 +27,4 @@ and contributing code!
    creating_frontend
    external_analysis
    deprecating_features
+   managing_optional_dependencies
--- a/doc/source/developing/managing_optional_dependencies.rst
+++ b/doc/source/developing/managing_optional_dependencies.rst
@@ -1,0 +1,38 @@
+How to manage soft (optional) dependencies
+------------------------------------------
+
+We might sometimes rely on heavy external libraries to support some features
+outside of yt's core. Typically, many frontends require HDF5 support, provided
+by the ``h5py`` library, but many users do not need it and shouldn't have to
+install such a large library to get yt.
+
+A mechanism to support soft-dependencies is implemented in the
+``yt/utilities/on_demand_imports.py`` module. Existing soft dependencies are
+listed in a ``optional_dependencies`` dictionary using package names for keys.
+The dictionary is then processed to generate drop-in replacement for the actual
+packages, which behave as normal instances of the packages they mirror:
+
+.. code-block:: python
+
+    from yt.utilities.on_demand_imports import h5py
+
+
+    def process(file):
+        with h5py.File(file, mode="r") as fh:
+            # perform interesting operations
+            ...
+
+In case the package is missing, an ``ImportError`` will be raised at runtime if
+and only if the code using it is run. In the example above, this means that as
+long as the ``process`` function is not run, no error will be raised. An
+implication is that soft dependencies cannot be used inconditionally in the
+global scope of a module, because it is run at yt's importtime, but should
+rather be restricted to lower level scopes, inside of functions or properly
+guarded conditional blocks.
+
+Note that, mocking an external package using the ``OnDemandImport`` class also
+delays any actual import to the first time it is used (this is how we can afford
+to fail only at the conditions we have to). A direct implication is that objects
+defined as such can safely be imported at the top level of a module, where the
+bulk of import statements live, without affecting import durations in any
+significant way.
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -2595,7 +2595,7 @@ class YTSurface(YTSelectionContainer3D):
 
     @parallel_root_only
     def _upload_to_sketchfab(self, data, files):
-        from yt.utilities.on_demand_imports import _requests as requests
+        from yt.utilities.on_demand_imports import requests
 
         SKETCHFAB_DOMAIN = "sketchfab.com"
         SKETCHFAB_API_URL = f"https://api.{SKETCHFAB_DOMAIN}/v2/models"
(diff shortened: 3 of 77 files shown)


matthewturk:
Hi -- I know this is draft, but I would prefer we not drop the underscore.  I would really like to avoid redefining objects that mirror the external module names.

In general, I guess I also didn't realize this was a priority -- did something change that means we have to address this?

neutrinoceros:
I found that 2 packages are effectively optional dependencies (in the sense that the corresponding import statements are embedded in try/except blocks), but migrating them to use the actual on-demand-import mechanism isn't trivial:
- `resource`: making it an actual on-demand-import lead to a `MemoryError` in windows CI. It is however used in only one line of the yt codebase. There's probably something to be done here but it will have to wait for next time.
- `yt_astro_analysis`: this is a weird case where `yt` is a natural hard dependency to `yt_astro_analysis`, so by importing it _in_ yt we create a circular dep. It is however not obvious how the situation could be mitigated since yt_astro_analysis is used in a yt frontend, which is by currently not something that's easy to migrate. I believe the general solution would be to improve the extensibility of yt by allowing users to define extension frontends without modifying the source code. This is obviously way out of scope here.
pinging @brittonsmith for reference, but if it is ever discussed more thoroughly, we should do it in a specific issue :)

neutrinoceros:
> Hi -- I know this is draft, but I would prefer we not drop the underscore. I would really like to avoid redefining objects that mirror the external module names.

I don't understand how that's a problem. A majority of the import statements we have on the main branch already redefine those with the `as` keyword, such as follows
```python
```

> In general, I guess I also didn't realize this was a priority -- did something change that means we have to address this?

It's not a priority, but it _is_ something that I can't solve in one go and want to be done with eventually. It's part of my broader effort to solve #3074

neutrinoceros:
@matthewturk I could be easily convinced that the leading underscore is a good thing, I'm actually more concerned about consistency. I could rewrite every import statement to use `as`. What's your preference ?

matthewturk:
My preference, which is I confess somewhat weakly held, is that we should
not be redefining the module names in the modules in which we're importing
them.  I'd much prefer defining them with the leading underscore to remove
any ambiguity, and it also helps us to be explicit in our import statements
and to help us catch the appropriate uses.

neutrinoceros:
That sounds fine to me. A concern I do have with the change you're proposing is that it will make this PR even more of a sweeping change, and unless it's merged soon, I will end up maintaining it conflict after conflict for what I can only be assume would be a while. On the other hand, if this goes in as is, there's less of a rush, and I can provide a follow up PR with that change, which will be a breeze to review by itself, so everything can be done in a matter of hours. Does this sound acceptable ?

matthewturk:
Wait, what I'm suggesting is actually to *preserve* the previous behavior
-- import as the real module name, but in the on-demand section, define
them with a leading underscore.  It should result in *fewer* files being
touched.

neutrinoceros:
> Wait, what I'm suggesting is actually to *preserve* the previous behavior
-- import as the real module name, but in the on-demand section, define
them with a leading underscore.  It should result in *fewer* files being
touched.

then I don't understand your argument. If we're not going to _use_ the underscore-prefixed names, then it makes no difference in the code that's using them, except for the fact that import statement get slightly "bloated".

matthewturk:
OK -- I would like to engage on this, but I am trying to find the mental bandwidth to work on the 4.0 stuff.  I'm sorry; I have to check out just for a little while, and try to get the last little bit of push.  :)  Would it be possible to return to this after the 4.0 release, and we can have a high-bandwidth low-latency chat over video or audio?  And in the interim, freeze the PR for the time being?

neutrinoceros:
You got it.

brittonsmith:
I only want to comment for now on the issue of importing yt_astro_analysis. I don't like that we do that. We only do it for the halo plot annotation. I think it might make more sense to move that plot callback into yt_astro_analysis as it's specific for halo analysis anyway. In my view, this is line with moving analysis_modules to yt_astro_analysis in the first place. I probably would have done it, but didn't catch it at the time.

neutrinoceros:
I 100% agree with this, thanks @brittonsmith

neutrinoceros:
close + reopening to retrigger test but I have no particular expectations

neutrinoceros:
I still can't understand why this is failing NetCDF4 tests. I vaguely feel like the real issue is upstream but in any case this PR is both more controversial than I anticipated at the time and not manageable. Overall it doesn't seem worth the effort of keeping it alive.

PR #3272, opened 2021-05-17
Implement CIC deposition for ParticleProjectionPlot, add FITSParticleProjection, fix bugs

jzuhone:
## PR Summary

This PR adds "cloud-in-cell" deposition for `ParticleProjectionPlot`, as well as adding an option to the same class to divide the projected values by the pixel area to get a "surface density". This should allow for better-looking plots of projected particle data in some cases. This also implemented a FITS version, `FITSParticleProjection`.

See https://nbviewer.jupyter.org/gist/jzuhone/447c7c725f59ac4e149e6e415249da85 for some examples.

This PR also fixes the following bugs I ran across in the process:

* The `CICDeposit_2` Cython function was computing the `i1` and `j1` indices incorrectly for deposition into the array. 
* `FITSImageData` instances as a rule do not have fields with code units--they are converted to some other unit base. However, I realized that though the units were changed in the file header, the array was never actually converted to the new units!
* Along similar lines, the logic to check for code units was not finding units mixed of code and non-code units.

I also refactored particle phase plots slightly so that when deposition is done (either CIC or NGP), it uses a "mask" array of integers to determine which cells/pixels were actually used in the deposit. This mirrors what was already done for `ParticleProjectionPlot` and is safer than checking a floating-point array for values > 0.0. 

I need to add tests. 

## PR Checklist


yt/utilities/lib/particle_mesh_operations.pyx
```diff
@@ -87,6 +87,7 @@ def CICDeposit_2(np.float64_t[:] posx,
                  np.float64_t[:] mass,
                  np.int64_t npositions,
                  np.float64_t[:, :] field,
+                 np.int_t[:, :] field_mask,
```
matthewturk, line 90: Can you change this, and the previous declaration, to be an int with an explicit size?  Since you're just setting it to 1, you could use `uint8` (which I think is the backing store for `bool`)

yt/visualization/fits_image.py
```diff
@@ -880,6 +886,16 @@ def construct_image(ds, axis, data_source, center, image_res, width, length_unit
...
             frb = data_source.to_frb(width[0], (nx, ny), center=center, height=width[1])
+    elif isinstance(data_source, ParticleAxisAlignedDummyDataSource):
+        bounds = (
+            center[0] - width[0] / 2,
+            center[0] + width[0] / 2,
+            center[1] - width[1] / 2,
+            center[1] + width[1] / 2,
+        )
+        frb = ParticleImageBuffer(
+            data_source, bounds, (nx, ny), periodic=np.all(ds.periodicity)
```
matthewturk, line 897: Is this the same as `all(ds.periodicit)`?

matthewturk (approved):

yt/visualization/fits_image.py
jzuhone: yes

matthewturk:
Looks like the errors are on purpose, right?  Adding "Projected" to the particle mass legend?

jzuhone:
Yes, I need to check them to be sure.

yt/data_objects/profiles.py
```diff
@@ -937,17 +937,20 @@ def _bin_chunk(self, chunk, fields, storage):
             else:
                 deposit_vals = wdata * fdata[:, fi]
 
+            field_mask = np.zeros(self.size).astype("uint8")
```
neutrinoceros, line 940: save a copy !
```suggestion
            field_mask = np.zeros(self.size, dtype="uint8")
```

yt/visualization/fixed_resolution.py
```diff
@@ -666,21 +670,67 @@ def __getitem__(self, item):
...
         splat_vals = weight_data[mask] * data[mask]
 
+        x_bin_edges = np.linspace(0.0, 1.0, self.buff_size[0] + 1)
+        y_bin_edges = np.linspace(0.0, 1.0, self.buff_size[1] + 1)
+
         # splat particles
         buff = np.zeros(self.buff_size)
-        buff_mask = np.zeros(self.buff_size).astype("int")
-        add_points_to_greyscale_image(buff, buff_mask, px[mask], py[mask], splat_vals)
+        buff_mask = np.zeros(self.buff_size).astype("uint8")
```
neutrinoceros, line 678:
```suggestion
        buff_mask = np.zeros(self.buff_size, dtype="uint8")
```

yt/visualization/fixed_resolution.py
```diff
@@ -666,21 +670,67 @@ def __getitem__(self, item):
...
+        ia = ImageArray(buff, units=units, info=info)
 
         # divide by the weight_field, if needed
         if weight_field is not None:
             weight_buff = np.zeros(self.buff_size)
-            weight_buff_mask = np.zeros(self.buff_size).astype("int")
-            add_points_to_greyscale_image(
-                weight_buff, weight_buff_mask, px[mask], py[mask], weight_data[mask]
-            )
+            weight_buff_mask = np.zeros(self.buff_size).astype("uint8")
```
neutrinoceros, line 718:
```suggestion
            weight_buff_mask = np.zeros(self.buff_size, dtype="uint8")
```

neutrinoceros (commented):
quick suggestions, I'll review more in depth later

chummels (approved):
Looking this over, the code looks good.  I also ran this on the FIRE_M12i_ref11 sample dataset and it behaved as predicted.  Initially, I thought that the "density" stuff should be relabeled "surface density", but upon further reflection I think it's fine as is.  When it's a zlabel, it shows as "Projected FIELD Density", which is accurate, and making the kwarg a "surface_density" instead of "density" is too cumbersome.  So I like it how it is.  A nice addition!  Thanks, @jzuhone .

neutrinoceros:
I will review this now, but won't merge before @britton lets us know wether he wants to go over this or not :)

yt/visualization/particle_plots.py
```diff
@@ -261,8 +275,12 @@ def __init__(
             weight_field,
             field_parameters=field_parameters,
             data_source=data_source,
+            deposition=deposition,
+            density=density,
         )
 
+        self.projected = weight_field is None
```
neutrinoceros, line 282: Unless this attribute already exists on the main branch, I'd prefer that the name better reflect its boolean type
```suggestion
        self.is_projected = weight_field is None
```

neutrinoceros (approved):
This is extremely well done, I'm impressed ! Here are some minor comments

yt/visualization/particle_plots.py
```diff
@@ -182,6 +186,14 @@ class ParticleProjectionPlot(PWViewerMPL):
     data_source : YTSelectionContainer Object
          Object to be used for data selection.  Defaults to a region covering
          the entire simulation.
+    deposition : string, optional
```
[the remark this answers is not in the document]
jzuhone: Our docstrings all over the codebase reflect this convention--if it's a kwarg, it's deemed "optional." If you want that changed that will have to be addressed in a separate PR after some discussion, because it's a long-standing precedent.

yt/visualization/particle_plots.py
jzuhone: This attribute already exists in `PlotWindow`

yt/visualization/fixed_resolution.py
```diff
@@ -666,21 +670,67 @@ def __getitem__(self, item):
...
         buff[buff_mask == 0] = np.nan
-        ia = ImageArray(buff, units=data.units, info=self._get_info(item))
+
+        # Normalize by the surface area of the pixel or volume of pencil if
+        # requested
+        info = self._get_info(item)
+        if density:
+            width = self.data_source.width
+            norm = width[self.xax] * width[self.yax] / np.prod(self.buff_size)
+            norm = norm.in_base()
```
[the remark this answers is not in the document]
jzuhone: The "base" is different depending on the unit system. When we do a "normal" projection we always convert it to the base units, so this is following the same convention

yt/utilities/lib/particle_mesh_operations.pyx
```diff
@@ -113,8 +114,8 @@ def CICDeposit_2(np.float64_t[:] posx,
         if (ypos < -0.5001) or (ypos > edgey):
             continue
 
-        i1  = <int> (xpos)
-        j1  = <int> (ypos)
+        i1  = <int> (xpos + 0.5)
+        j1  = <int> (ypos + 0.5)
```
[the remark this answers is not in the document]
jzuhone: If you look above in the same file at `CICDeposit_3`, the additional 0.5 is also present. This follows that rule (and is consistent with the overall algorithm) and also was the cause of negative mass values I was seeing.

doc/source/visualizing/plots.rst
```diff
@@ -1590,6 +1590,17 @@ along the line of sight. The inner region is dominated by low mass
...
+Both :class:`~yt.visualization.particle_plots.ParticleProjectionPlot` and
+:class:`~yt.visualization.particle_plots.ParticlePhasePlot` objects
+accept a ``deposition`` argument which controls the order of the "splatting"
+of the particles onto the pixels in the plot. The default option, ``"ngp"``,
+corresponds to the "Nearest-Grid-Point" (0th-order) method, which simply
+finds the pixel the particle is located in and deposits all of the particle
+or its plotted quantity into that pixel. The other option, ``"cic"``,
+corresponds to the "Cloud-In-Cell" (1st-order) method, which linearly
+interpolates the particle or its plotted quantity into the four nearest
+pixels in the plot.
```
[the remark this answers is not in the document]
jzuhone: The way the algorithm is specified, it will deposit all of that mass into that cell.

yt/visualization/particle_plots.py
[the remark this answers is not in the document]
jzuhone: Going to implement this suggestion in any cased

doc/source/visualizing/plots.rst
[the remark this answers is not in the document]
neutrinoceros: excellent, thanks !

yt/visualization/fixed_resolution.py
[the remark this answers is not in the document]
neutrinoceros: ok cool

yt/visualization/particle_plots.py
neutrinoceros: indeed, then disregard !

neutrinoceros:
Just noticed I originally pinged the wrong "Britton" above (edited).
@brittonsmith , since John originally requested your review, I'll let a chance to do it if you want

PR #3273, opened 2021-05-17
Converting Enzo-P to Enzo-E

chummels:
## PR Summary

Enzo-P is a hydro code originally meaning Enzo-Petascale.  Because it has been under development for about a decade, it's name has now been updated to Enzo-E (Exascale).  There is some confusion in these two names, so this PR just updates all references to Enzo-P and changes them to Enzo-E.  There should be no functional change to the code.


yt/frontends/enzo_p/api.py
```diff
@@ -1,6 +0,0 @@
-from . import tests
```
neutrinoceros, line 1: ⚠️ It looks like you deleted this file instead of renaming it

neutrinoceros: never mind I now see that you already reintroduced it correctly.

neutrinoceros:
I added the backward incompatible label because this will break any code downstream importing from `yt.frontends.enzo_p.api`. While I believe affected users will be actually happier with the renamed frontend, it still mandates a mention as backward incompatible in our changelog, IMO

neutrinoceros (approved):

chummels:
Oh good call.  Yeah, I can add this to the yt4 diffs page if it is warranted, but I figured there would be some step (perhaps at our next triage meeting?) where we go through the changelog and add relevant notes to the yt4diffs page.

I *think* I got all of the instances of enzo-p correctly converted.  I did a global search for enzo and tried to address all of the cases, but I guess the tests will reveal if I messed it up.

chummels:
@brittonsmith : I included you as a reviewer on this PR, since I believe you're the original author of this frontend.  I think this PR is ready to merge, and directly addresses concerned raised in the Enzo slack (Tumlinson) about there being confusion about searching for enzo-p things.  I had hoped that maybe I could sneak this in fast before there are a bunch of enzo-e PRs that get issued? But maybe that's not best, since there is already #3274 .  Anyway, I leave it to you (and others) if we want to merge this now or hold off until all after the enzo meeting this week.

neutrinoceros:
@chummels , looks like you made a small mistake in migrating the test files ? see the failure on Jenkins https://tests.yt-project.org/job/yt_py38_git/3095/testReport/junit/nose.failure/Failure/runTest/

brittonsmith:
I think `yt/tests/tests.yaml` needs to be updated to account for the path change.

matthewturk:
You may also need to rename the tests in the answer test repository.

chummels:
> You may also need to rename the tests in the answer test repository.

I'm not sure how to do that.  When I download the answer test repo (this [one](https://github.com/yt-project/answer-store), right?), I get no hits when I grep for "enzo" or `find . -name '*enzo*'.  So I'm not sure how to change this appropriately.

chummels:
> I think `yt/tests/tests.yaml` needs to be updated to account for the path change.

Thanks, I must've missed this, but now I've included it.

matthewturk:
@chummels I think on reflection you probably don't have to do anything with that repo.  I was mistaken.

tests/tests.yaml
```diff
@@ -47,9 +47,9 @@ answer_tests:
     - yt/frontends/enzo/tests/test_outputs.py:test_ecp
     - yt/frontends/enzo/tests/test_outputs.py:test_nuclei_density_fields
 
-  local_enzo_p_010: # PR 2945
-    - yt/frontends/enzo_p/tests/test_outputs.py:test_hello_world
-    - yt/frontends/enzo_p/tests/test_outputs.py:test_particle_fields
+  local_enzo_e_010: # PR 2945
```
neutrinoceros, line 50: the number should be downgraded
```suggestion
  local_enzo_e_001: # PR 3273
```

chummels: Oh I see.  So because I'm converting enzo-p to enzo-e, this is effectively the first answer tests associated with this "new" frontend?  OK, that's fine.

neutrinoceros:
Looks ready to ship !

chummels:
Yeah, I'm not sure if @brittonsmith was going to look at this or not, since he was the original author of the frontend?  But I know it's getting late over there.

brittonsmith (approved):

PR #3275, opened 2021-05-18
Set current time off redshift if time attribute missing from data.

brittonsmith:
## PR Summary

It seems there are some variants of the OWLS Subfind frontend that are missing an attribute for the current time in Gyr. However, this is redundant given the current redshift and cosmological parameters, so we can set time off that if necessary.

## PR Checklist

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1216,6 +1216,9 @@ class Dataset(abc.ABC):
             w_a=w_a,
         )
 
+        if not hasattr(self, "current_time"):
+            self.current_time = self.cosmology.t_from_z(self.current_redshift)
+
         if getattr(self, "current_redshift", None) is not None:
             self.critical_density = self.cosmology.critical_density(
                 self.current_redshift
--- a/yt/frontends/owls_subfind/data_structures.py
+++ b/yt/frontends/owls_subfind/data_structures.py
@@ -112,7 +112,8 @@ class OWLSSubfindDataset(ParticleDataset):
         self.refine_by = 2
 
         # Set standard values
-        self.current_time = self.quan(hvals["Time_GYR"], "Gyr")
+        if "Time_GYR" in hvals:
+            self.current_time = self.quan(hvals["Time_GYR"], "Gyr")
         self.domain_left_edge = np.zeros(3, "float64")
         self.domain_right_edge = np.ones(3, "float64") * hvals["BoxSize"]
         self.domain_dimensions = np.ones(3, "int32")


matthewturk (approved):

matthewturk:
LGTM, and shouldn't impact negatively existing datasets.

brittonsmith:
Great, agreed! I think this shouldn't hurt anyone else.

jzuhone (approved):

yt/data_objects/static_output.py
```diff
@@ -1216,6 +1216,9 @@ def setup_cosmology(self):
             w_a=w_a,
         )
 
+        if not hasattr(self, "current_time"):
+            self.current_time = self.cosmology.t_from_z(self.current_redshift)
```
neutrinoceros, line 1220: I would argue that it'd be preferable to do this in the frontend instead of the parent class. `Dataset` has already too much responsability to my taste. Plus it feels weird to have this in the common dataflow, as someone who never used a cosmological dataset. Is there something I'm missing ?

matthewturk:
Ah actually, I missed that this was in the dataset. I tend to agree that
this should be done in the frontend.

jzuhone:
There are plenty of cosmological datasets across different frontends and I can see this being potentially useful for others, however.

neutrinoceros:
> There are plenty of cosmological datasets across different frontends and I can see this being potentially useful for others, however.

Of course. What I'm biting at here is that there should really be more than one monolithic abstract Dataset class, otherwise we're always going to have to choose between code duplication and premature specialisation. This is something I mean to address in the future, but I'd lean on the side of code duplication for now since there is no known need for this feature in other frontend atm.

matthewturk:
I don't want this to get held up on the location of the cosmological time, but I do think we should address it in the very near future.  Not all datasets will have both a time and a redshift, for instance.

neutrinoceros:
@brittonsmith you have two approvals already, and I don't wish to be a nuisance, so you have my blessing to merge this yourself if you don't want to address my remark.

yt/data_objects/static_output.py
brittonsmith: I see what you're saying about not adding further specialism in the Dataset class, but this is inside the `setup_cosmology` method, so I don't think it goes outside of appropriate boundaries. Similar to what @jzuhone pointed out, I was anticipating that this might be useful for other cosmological frontends to avoid having duplicate subclass methods to solve the same problem.
More generally, I am quite in favor of moving some of this to generalized subclasses like `CosmologyDataset`. I think the way I did this here does not make that future work more difficult.

@neutrinoceros, for the sake of discussion, we can leave this open so you can respond if you like. If you're not convinced, I'll change it to being frontend-specific. I'm not that bothered either way.

matthewturk: It being in cosmology setup seems fine, now that I continue to think about it!  :)

neutrinoceros: Oh that's a very good point, I didn't realise it was happening in an already cosmology-specific method (or that the base Dataset class had any). Please disregard my original concern then

PR #3276, opened 2021-05-19
avoid copying arrays with np.ndarray.astype where possible

neutrinoceros:
## PR Summary

Just a couple of very minor simplifications. There are probably many more that can be found but in many cases, the `dtype` argument was added in various numpy functions throughout the years, and it's not easy to automate detection. I may come up with a flake8 plugin to do that in the future.


neutrinoceros:
looks like a giant rabbit hole, won't fix

Issue #3269, opened 2021-05-16
PhasePlots don't correctly render ion fields for z_axis label


chummels:
### Bug report

**Bug summary**

Usage of ion fields such as `H_p0_mass` (for H I mass) are not rendered properly for `PhasePlot` z-axis labels, but they work correctly for `ProjectionPlot`s z-axis labels.  I'm happy to fix this, but I cannot find where in the code `ProjectionPlot` is doing this step.  Suggestions are welcome.

**Code for reproduction**

```python
import yt
ds = yt.load_sample('IsolatedGalaxy')
yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), ('gas', 'H_p0_mass')).save()
yt.ProjectionPlot(ds, 'x', ('gas', 'H_p0_number_density')).save()
```

**Actual outcome**

![galaxy0030_2d-Profile_density_temperature_H_p0_mass](https://user-images.githubusercontent.com/8250999/118381594-321c2700-b5a1-11eb-9e5b-c3efe59b8dee.png)

![galaxy0030_Projection_x_H_p0_number_density](https://user-images.githubusercontent.com/8250999/118381589-27fa2880-b5a1-11eb-9f58-839c5940df84.png)

neutrinoceros:
It seems that this was fixed. Running your example again I get the expected result
![galaxy0030_2d-Profile_density_temperature_H_p0_mass](https://user-images.githubusercontent.com/14075922/186029181-a49634ed-6994-48cd-8db8-fdd0aa209b93.png)

chummels:
I think this got fixed in #3282.

PR #3282, opened 2021-05-20
Set the display name for new fields

chummels:
## PR Summary

This PR addresses Issue #3269.  If a new derived field has not defined a `display_name` (used for rendering in figures) at creation time, this generates it at the end of initialization.  This probably has little impact on most fields being generated, but this will ensure that species fields (e.g., H_p0 = `H I`) are rendering correctly in both `PhasePlots` and `ProfilePlots` instead of just in `ProjectionPlots` and `SlicePlots`.  Example (see `z_label`):

galaxy0030_2d-Profile_density_temperature_H_p0_mass


chummels:
It looks like I broke a bunch of answer tests with this change.  The main failing culprits are `PhasePlot` objects  from grid-based datasets that have a `mass` weighting (now showing `Cell Mass` as `z_label`).  I *think* what's happening is that when a new field is created, like `('Enzo`, 'cell_mass')`, it is now getting a `display_name` created for that field: `Cell Mass`.  And then there is an alias field of `('gas', 'mass')` that aliases back to `('Enzo', 'cell_mass')`, it inherits its `display_name`.  Which means that even when a `PhasePlot` uses `('gas', 'mass')` as its weight field, it's displaying the `display_name` for the original field: `Cell Mass`.  I'll have to rethink how to beat this problem.

neutrinoceros:
>  think what's happening is that when a new field is created, like ('Enzo', 'cell_mass'), it is now getting a display_name created for that field: Cell Mass. And then there is an alias field of ('gas', 'mass') that aliases back to ('Enzo', 'cell_mass'), it inherits its display_name

sounds about right. I'd argue that inheriting the display name is actually what we want, and "Cell Mass" makes more sense than "Mass" as a label, in my opinion. Unless I'm missing an argument here, I think the answers should actually be updated.

jzuhone:
> I'd argue that inheriting the display name is actually what we want, and "Cell Mass" makes more sense than "Mass" as a label, in my opinion.

In general what we want to do is avoid referencing the underlying data structures (cell or particle) and focus on the physical quantities so as to make scripts as portable as possible across different frontends. In most use cases there's no reason why a user shouldn't be able to take a script written for Gadget and use it for Enzo, etc. or vice-versa.

matthewturk:
I thought we did?  In the field tuples there's the option for that.

chummels:
I agree with @jzuhone that we don't want `cell_mass` in the PhasePlot zlabel when it comes from a `('gas', 'mass')` field, even when that is aliased in some cases to `('FRONTEND', 'cell_mass')`.  

Any ideas anyone has on where I could appropriately set the `display_name` to avoid this problem?  The way it works with `ProjectionPlot` and `SlicePlot` is in `FixedResolutionBuffer` right when the plot is being generated.  But I'm not sure where to make that happen for `ProfilePlot`s and `PhasePlot`s--I guess they don't rely on an FRB for those plots.

matthewturk:
Kinda seems like this would be a pretty big change at this point in the
game?

jzuhone:
@matthewturk can you elaborate a bit?

matthewturk:
I guess just thinking about changing "Cell Mass" in the z label of
plots to something different struck me as a big change.

jzuhone:
If you're plotting the `cell_mass` field from the frontend, it should say "Cell Mass", but otherwise it should just say "Mass", IMO.

chummels:
> I guess just thinking about changing "Cell Mass" in the z label of plots to something different struck me as a big change.

I think there's some confusion here, potentially because of the long string of comments, so I'll try to summarize.  The existing behavior in yt for a PhasePlot weighted by the `('gas', 'mass')` field is to (correctly) created a `z_label` of `mass`. 
 This PR is trying to correct a problem associated with species fields showing up as `H p0 mass` instead of `H I mass` as the `z_label` in `PhasePlots`.  The solution that I submitted fixes the species problem, but *breaks* the current paradigm of displaying `z_label`s for a mass field that is aliased to a grid-based frontend: it should display `mass`, but it's displaying `cell_mass` instead because its inheriting the `display_name` when it aliases the field.  I'm looking to change my current PR implementation to preserve the correct behavior of displaying `mass` as instead of `cell_mass`, which is how yt already does things.  This can be seen in the failed answer tests.  
 
So PhasePlot/ProfilePlot doesn't create an FRB.  I guess I can just try to set the label in those functions directly.

chummels:
So I rewrote this to set the display_field in `ProfilePlot` and `PhasePlot` directly, using the same machinery that is in the FRB class that works for `ProjectionPlot` and `SlicePlot`.  As I was doing so, I realized that the parentheses around the label units in `PhasePlot`s are wonky, so I fixed them.  Those led to the failing answer tests.  I would argue that these failures are all desired.  Here is a representative image difference of one of the failures:

Screen Shot 2021-05-25 at 1 13 24 AM

If people are OK with this solution, I will make some tests and update the answer tests.

yt/visualization/profile_plotter.py
```diff
@@ -729,15 +729,16 @@ def _get_field_label(self, field, field_info, field_unit, fractional=False):
...
-            field_name = r"$\rm{" + field.replace("_", r"\ ").title() + r"}$"
+            field_name = field_info.get_latex_display_name()
         elif field_name.find("$") == -1:
             field_name = field_name.replace(" ", r"\ ")
             field_name = r"$\rm{" + field_name + r"}$"
         if fractional:
             label = field_name + r"$\rm{\ Probability\ Density}$"
         elif field_unit is None or field_unit == "":
             label = field_name
+        elif "frac" in field_unit:
```
cphyc, line 740: This gets me worried a little if you have a unit name that contains the four letters `frac`. None of them comes to my mind right now, but since users can define custom units, you could imagine something like `fraction_density` or whatever.
Shouldn't this be replaced by
```suggestion
        elif r"\frac" in field_unit:
```
or

```suggestion
        elif r"\dfrac" in field_unit:
```

yt/visualization/profile_plotter.py
```diff
@@ -1017,15 +1018,16 @@ def _get_field_label(self, field, field_info, field_unit, fractional=False):
...
```
cphyc, line 1029: Same here

cphyc (approved):

yt/visualization/profile_plotter.py
neutrinoceros: I get by your worrying but I don't understand what any of your alternatives does ?

cphyc: The piece of code is checking that the latex representation is e.g.
```latex
\frac{\rm{g}}{\rm{cm}^{3}}
```
which we can catch by explicitly checking that the string contains the `\frac` substring (or even `\frac{`)

neutrinoceros: oh right, that makes sense

chummels:
Good call, @cphyc .  I think the code is now stable against any fields that might have `frac` in them.

So this should be ready to go, I think.  In short, it corrects the problem described in Issue #3269.  It also corrects some issues with parentheses in field units displayed in PhasePlot (the parentheses were too small to cover fractional units like g/cm**3).  See the image above in the comments.  I guess I need to generate new answer tests for this?

neutrinoceros:
@chummels your latest change is causing image comparison tests to fail, but the change is actually desirable !
reference
Screenshot 2021-10-04 at 19 35 55

your patch
Screenshot 2021-10-04 at 19 36 09

so you actually fixed bad latex here, thank you !
Now you're going to need to open a PR on the answer store to bump these answers, but note that your change is impacting the same answers as the one bumped here https://github.com/yt-project/answer-store/pull/23 , so I'd suggest we deal with _that_ PR (and #3520) first and then come back to _this_ PR to include your change. How does this sound ?

chummels:
Sure that sounds fine.  We can get all these label issues fixed one at a time.

neutrinoceros:
#3520 is in. Now we need to bump the answer store for this PR. Let's run the answer tests again to get an up to date result we can submit to the answer store

chummels:
I'm happy either way, whichever is easiest.  I'm happy to make a companion PR to the answer store, but I'm not sure how to do this.  I'd love your assistance with it since you offered.

neutrinoceros:
Alright so here are the steps:
1) go the the relevant job's log (https://github.com/yt-project/yt/runs/3838740844?check_suite_focus=true)
2) at the bottom of the page, unroll the "report failed answers" section
Screenshot 2021-10-09 at 08 24 01
3) at the bottom top *that* section, you'll find two links, the first own downloads a visual html report of your changes for reference, and the second one downloads a zip archive.
Screenshot 2021-10-09 at 08 27 46
4) Get that zip
5) fork the answer store repo https://github.com/yt-project/answer-store
6) clone your fork locally (keeping a separate copy from your main yt clone will make things much easier in case anything breaks later)
7) checkout the `yt-4.0` branch
8) create a new branch from there (for instance `support_3282`)
9) now unzip the archived from step 4 from the top level of the repo
10) commit
11) push your branch
12) open a PR

Admittedly this is hard to guess, but I think you have everything you need here. Don't hesitate to reach out in case something isn't clear.
After the companion PR is merged, we'll also need to update the submodule here.

neutrinoceros:
Ah, answer tests failed again. Apparently we needed to rebase this branch ahead of regenerating images for it to take into account the changes from #3520 . I rebased the branch here and produced a new companion PR for the answer store https://github.com/yt-project/answer-store/pull/25

matthewturk:

neutrinoceros:
Let's try this one last time now that a lot of other PRs went in. If it fails we'll resort to manual backport

PR #3288, opened 2021-05-20
Refactoring "default" species fields

jzuhone:
## PR Summary

Two years ago, PR #2245 refactored (among other things) a number of fields which depended on assumptions about the mean molecular weight and the electron number density, such as `("gas","entropy")`, `("gas","szy")`, etc. The decision was made to have the `("gas","mean_molecular_weight")`, `("gas","El_number_density")`, and H/He nuclei and number density fields always exist for datasets which had a `("gas","density")` field, regardless of whether or not species are defined in the dataset. In the case they were not, the assumption was made of a primordial H/He composition with full ionization. 

This was probably never the ideal situation--though it was documented, it is an assumption that would not necessarily apply to many datasets. More importantly, certain downstream projects (like Trident, see https://github.com/trident-project/trident/issues/165) relied on the assumption that if H and He were not defined that they could supply a new field based on physical assumptions about the medium, but trust that if the field was present in the dataset that it was correct.

Based on these considerations, and after a long and productive conversation with @chummels, this PR partially reverts the effect of PR #2245 by making this feature optional. It adds a new kwarg to the `Dataset` class, `default_species_fields`, which by default is `None`, and can also take the options of `"ionized"` or `"neutral"`. These options are:

* `"ionized"`: sets up nuclei and number density fields for primordial H/He and free electrons assuming full ionization.
* `"neutral"`: sets up nuclei and number density fields for primordial H/He assuming full neutrality.

In theory this scheme could be extended to support other scenarios which are common. 

A number of temperature fields in various frontends rely on the assumption of a mean molecular weight--if one is not defined by the dataset, at load time, or by `default_species_fields`, the default will continue to be that of a fully ionized H/He-dominated plasma. This is usually overridden by the frontend if the dataset has a value for it, and many frontends allow you to change it when `load`ing a dataset. 

Unfortunately this also required adding the `default_species_fields` kwarg to nearly every subclass of `Dataset`--I tried to only do it for those which it would definitely make sense based on domain context. 

This should fix https://github.com/trident-project/trident/issues/165. 

EDIT: This also fixes a bug I found in the RAMSES frontend (ping @cphyc) which wasn't actually registering the species from the dataset and mapping them to the standard yt names, so you weren't getting number and nuclei density fields. This bug was actually being hidden(!) because the default fields for H were being set up by the previous behavior (further showing why this was the wrong thing to do on my part originally). 

### Is there an alternative? 

Unfortunately not one that is less hacky. It was considered that one could create a method such as `ds.setup_species_fields`, but this would need to be run _before_ field detection (basically before you did anything else after loading a dataset), and there are a small number of frontends that have to do field detection at `load` time in any case. 

## PR Checklist

### DIFF
--- a/yt/fields/species_fields.py
+++ b/yt/fields/species_fields.py
@@ -208,30 +208,42 @@ def add_nuclei_density_fields(registry, ftype):
 
     # Here, we add default nuclei and number density fields for H and
     # He if they are not defined above. This assumes full ionization!
-    for element in ["H", "He"]:
-        if element in elements:
-            continue
-        registry.add_field(
-            (ftype, f"{element}_nuclei_density"),
-            sampling_type="local",
-            function=_default_nuclei_density,
-            units=unit_system["number_density"],
-        )
-        if element == "H":
-            registry.alias((ftype, "H_p1_number_density"), (ftype, "H_nuclei_density"))
-
-        if element == "He":
-            registry.alias(
-                (ftype, "He_p2_number_density"), (ftype, "He_nuclei_density")
+    if registry.ds.default_species_fields is not None:
+        dsf = registry.ds.default_species_fields
+        for element in ["H", "He"]:
+            if element in elements:
+                continue
+            registry.add_field(
+                (ftype, f"{element}_nuclei_density"),
+                sampling_type="local",
+                function=_default_nuclei_density,
+                units=unit_system["number_density"],
             )
+            if element == "H":
+                if dsf == "ionized":
+                    state = "p1"
+                elif dsf == "neutral":
+                    state = "p0"
+                registry.alias(
+                    (ftype, f"H_{state}_number_density"), (ftype, "H_nuclei_density")
+                )
 
-    if (ftype, "El_number_density") not in registry:
-        registry.add_field(
-            (ftype, "El_number_density"),
-            sampling_type="local",
-            function=_default_nuclei_density,
-            units=unit_system["number_density"],
-        )
+            if element == "He":
+                if dsf == "ionized":
+                    state = "p2"
+                elif dsf == "neutral":
+                    state = "p0"
+                registry.alias(
+                    (ftype, f"He_{state}_number_density"), (ftype, "He_nuclei_density")
+                )
+
+        if (ftype, "El_number_density") not in registry and dsf == "ionized":
+            registry.add_field(
+                (ftype, "El_number_density"),
+                sampling_type="local",
+                function=_default_nuclei_density,
+                units=unit_system["number_density"],
+            )
 
 
 def _default_nuclei_density(field, data):
@@ -239,7 +251,7 @@ def _default_nuclei_density(field, data):
     element = field.name[1][: field.name[1].find("_")]
     amu_cgs = data.ds.units.physical_constants.amu_cgs
     if element == "El":
-        # This assumes full ionization!
+        # If we got here, this assumes full ionization!
         muinv = 1.0 * _primordial_mass_fraction["H"] / ChemicalFormula("H").weight
         muinv += 2.0 * _primordial_mass_fraction["He"] / ChemicalFormula("He").weight
     else:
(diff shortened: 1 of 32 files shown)


yt/fields/species_fields.py
```diff
@@ -208,38 +208,50 @@ def add_nuclei_density_fields(registry, ftype):
...
+        for element in ["H", "He"]:
+            if element in elements:
+                continue
+            registry.add_field(
+                (ftype, f"{element}_nuclei_density"),
+                sampling_type="local",
+                function=_default_nuclei_density,
+                units=unit_system["number_density"],
+            )
+            if element == "H":
```
matthewturk, line 222: Can you leave comments here explaining?  I mean, *obviously* I totally understand and don't need you to explain this to me, but, you know, in case someone else came across it, like my friend, his username is tatthewmurk I think, if he wanted to have a better understanding, yeah?

jzuhone: Tell tatthewmurk that we've got some comments in there now :)

jzuhone:
Since @brittonsmith had a lot of comments on PR #2245 a couple of years ago I'd love to get some feedback from him on this.

chummels (approved):
This looks great, @jzuhone .  I just included a couple of slight suggestions, but only commit them if you agree.  Thanks for coding this up and being willing to modify things.  I'm sorry I missed the original PR two years ago!

yt/utilities/chemical_formulas.py
```diff
@@ -34,12 +34,13 @@ def __repr__(self):
...
-    # This assumes full ionization!
-    muinv = 2.0 * _primordial_mass_fraction["H"] / ChemicalFormula("H").weight
-    muinv += 3.0 * _primordial_mass_fraction["He"] / ChemicalFormula("He").weight
+def compute_mu(ion_state):
+    if ion_state == "ionized" or ion_state is None:
+        Z_H = 2.0
+        Z_He = 3.0
+    elif ion_state == "neutral":
+        Z_H = 1.0
+        Z_He = 2.0
```
brittonsmith, line 43: Shouldn't this be 1.0? With that, I get neutral mu ~ 1.22.

brittonsmith:
I'm happy with this change in general. Had just a couple minor comments.

yt/utilities/chemical_formulas.py
jzuhone: @brittonsmith sorry--what do you mean by "this"?

yt/fields/species_fields.py
```diff
@@ -207,42 +207,62 @@ def add_nuclei_density_fields(registry, ftype):
...
-        )
-        if element == "H":
-            registry.alias((ftype, "H_p1_number_density"), (ftype, "H_nuclei_density"))
-
-        if element == "He":
-            registry.alias(
-                (ftype, "He_p2_number_density"), (ftype, "He_nuclei_density")
+    # He if they are not defined above, and if it was requested by
+    # setting "default_species_fields"
+    if registry.ds.default_species_fields is not None:
```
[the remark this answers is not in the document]
jzuhone: Yep, sounds good.

yt/fields/species_fields.py
```diff
@@ -207,42 +207,62 @@ def add_nuclei_density_fields(registry, ftype):
...
+                function=_default_nuclei_density,
+                units=unit_system["number_density"],
+            )
+            # Set up number density fields for hydrogen, either fully ionized or neutral.
+            if element == "H":
+                if dsf == "ionized":
+                    state = "p1"
+                elif dsf == "neutral":
+                    state = "p0"
+                registry.alias(
```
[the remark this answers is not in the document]
jzuhone: Yes, I'll put some error handling in.

yt/utilities/chemical_formulas.py
jzuhone: @brittonsmith argh--sorry, you're right. I have no idea where my head is at. will fix.

yt/frontends/ramses/fields.py
```diff
@@ -48,6 +48,21 @@
     ]
 }
```
jzuhone, line 50: @cphyc can you look at the change I made here?

cphyc (approved):
After staring in the void for a few minutes I think I convinced myself this is fine for RAMSES (and others as well).

PR #3289, opened 2021-05-20
Restore auto creation of save dirs in PlotContainer.save

### LINKED ISSUE #3210: Saving a PlotWindow object to a non-existent directory with a file prefix no longer works
### Bug report

In the past, it was possible to save a SlicePlot or ProjectionPlot out to a non-existent directory with a custom prefix. but this no longer works.  Now, you must create the directory first before saving there, or else you get a `FileNotFoundError`.  

What *does* work is this.  

```
import yt
ds = yt.testing.fake_amr_ds()
p = yt.SlicePlot(ds, "z", 'Density')
p.save("images/")
```

It works even when `images` does not yet exist, by first creating the `images` directory and then saving the file `AMRGridData_Slice_z_Density.png` into the `images` directory.

What does not work is this:

```
import yt
ds = yt.testing.fake_amr_ds()
p = yt.SlicePlot(ds, "z", 'Density')
p.save("images/test")
```

If you do not have an `images` directory created, it will fail with a `FileNotFoundError`.  Previously, it would create an `images` directory, then output `test_Slice_z_Density.png` to it, similar to the non-prefixed example above.

The error that I now receive is:

```
  File "test_images.py", line 4, in <module>
    p.save("images/test")
  File "/Users/chummels/src/yt/yt/visualization/plot_container.py", line 76, in newfunc
    rv = f(*args, **kwargs)
  File "/Users/chummels/src/yt/yt/visualization/plot_container.py", line 564, in save
    names.append(v.save(name, mpl_kwargs))
  File "/Users/chummels/src/yt/yt/visualization/base_plot_types.py", line 149, in save
    canvas.print_figure(name, **mpl_kwargs)
  File "/Users/chummels/miniconda3/lib/python3.8/site-packages/matplotlib/backend_bases.py", line 2119, in print_figure
    result = print_method(
  File "/Users/chummels/miniconda3/lib/python3.8/site-packages/matplotlib/backends/backend_agg.py", line 535, in print_png
    with cbook.open_file_cm(filename_or_obj, "wb") as fh:
  File "/Users/chummels/miniconda3/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/Users/chummels/miniconda3/lib/python3.8/site-packages/matplotlib/cbook/__init__.py", line 418, in open_file_cm
    fh, opened = to_filehandle(path_or_file, mode, True, encoding)
  File "/Users/chummels/miniconda3/lib/python3.8/site-packages/matplotlib/cbook/__init__.py", line 403, in to_filehandle
    fh = open(fname, flag, encoding=encoding)
FileNotFoundError: [Errno 2] No such file or directory: 'images/test_Slice_z_Density.png'
```

I guess maybe this is minor since people should be already create the directories to where they're saving files?  But given that this used to work up until recently, and that it still works when you don't specify a prefix, maybe it's an easy fix and worth doing?

At the very least, if this is going to be left as is, I think we need to document it on the yt4 differences page.  I tried to run a script that used to work and it broke and it took a while to track down what was going wrong.

neutrinoceros:
## PR Summary

Intent: fix #3210
The patch is not elegant at all, but it works ™️ (at least the tests I wrote for it pass)
Note that in order to make this work I "sacrificed" a feature that was deemed unintuitive in the original disscussion. Namely,
`PlotContainer.save("img")` used to interpret `"img"` as a file prefix OR an output directory depending on wether such a directory existed already or not. This PR intentionally breaks this undocumented (AFAIK) feature to make room for the one @chummels wanted back in his original report.

Following the discussions in the issue, I think the best course of action in the foreseeable future would be rethink the amount of magic we want to support in saving functions, but for now, we've agreed that it's best to just try and restore the old behaviours in time for the 4.0 release.

### DIFF
--- a/nose_unit.cfg
+++ b/nose_unit.cfg
@@ -6,5 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
-ignore-files=(test_load_errors.py|test_load_sample.py|test_commons.py|test_ambiguous_fields.py|test_field_access_pytest.py)
+ignore-files=(test_load_errors.py|test_load_sample.py|test_commons.py|test_ambiguous_fields.py|test_field_access_pytest.py|test_save.py)
 exclude-test=yt.frontends.gdf.tests.test_outputs.TestGDF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -183,6 +183,7 @@ other_tests:
      - '--ignore-files=test_load_sample.py'
      - '--ignore-files=test_field_access_pytest.py'
      - '--ignore-files=test_ambiguous_fields.py'
+     - "--ignore-files=test_save.py"
      - '--exclude-test=yt.frontends.gdf.tests.test_outputs.TestGDF'
   cookbook:
      - 'doc/source/cookbook/tests/test_cookbook.py'
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -555,15 +555,24 @@ class PlotContainer:
         names = []
         if mpl_kwargs is None:
             mpl_kwargs = {}
-        if isinstance(name, (tuple, list)):
-            name = os.path.join(*name)
+
         if name is None:
             name = str(self.ds)
+
+        # ///// Magic area. Muggles, keep out !
+        if isinstance(name, (tuple, list)):
+            name = os.path.join(*name)
+
         name = os.path.expanduser(name)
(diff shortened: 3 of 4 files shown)


chummels (approved):
Thanks for fixing this, @neutrinoceros ! I looked over the code and the logic works.  I also downloaded it and tested it locally and it works as described.  Much appreciated.

matthewturk (approved):

PR #3293, opened 2021-05-22
Write demeshening description.  Fixes #2629.

### LINKED ISSUE #2629: Add narrative documentation for how the demeshening works

matthewturk:
This adds narrative docs about demeshening and how it works from the perspective of someone using it.  It includes links to the yt 4 paper description, and I've also added cross-references within the docs where appropriate.


matthewturk:
This should fix #2629.

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,146 @@
...
+curve, using a `enhanced word-aligned
+hybrid<https://github.com/lemire/ewahboolarray>` boolean array as their
+backend.
+
+.. note::
+
+   You may see scattered references to "the demeshening" (including in the
+   filename for this document!)  This was a humorous name used in the yt
+   development process to refer to removing a global (octree) mesh for
+   particle codes.
```
neutrinoceros, line 20: Punctuation :)
Also, really cool to have this note, I was going to comment that maybe the word "demeshening" was probably note really to pass the test of time very well, seeing that it's mostly historic, but it makes sense to have it documented as such, at least for a few years !
```suggestion
   You may see scattered references to "the demeshening" (including in the
   filename for this document!). This was a humorous name used in the yt
   development process to refer to removing a global (octree) mesh for
   particle codes.
```

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,146 @@
...
+For example, if you were to use the sample Gadget-3 dataset, you could override
+the default values and use values of 5 and 5 by specifying this argument to the
+``load_sample`` function; this works with ``load`` as well.
+
+.. code-block:: python
+
+   ds = yt.load_sample("Gadget3-snap-format2", index_order=(5, 5))
+
+So this is how you *change* the index order, but it doesn't explain precisely
+what this "index order" actually is.
```
neutrinoceros, line 51: Now you're making me uncomfortable: will you please stop anticipating my remarks ? I'd like to have **something** to complain about !

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,146 @@
...
+particle is assigned a "coarse" index, which is global to the full domain of
+the collection of particles, and *if necessary* an additional "refined" index
+is assigned to the particle, within that coarse index.
+
+The index "order" values refer to the number of entries on a side that each
+index system is allowed.  For instance, if we allow the particles to be
+subdivided into 8 "bins" in each dimension, this would correspond to an index
+order of 3 (as $2^3 = 8$); correspondingly, an index order of 5 would be 32
+bins in each dimension, and an index order of 7 would be 128 bins in each
+dimension.  Each particle is then assigned a set of i, j, k values for the bin
```
neutrinoceros, line 67: Maybe this isn't really clearer, likely just a matter of personal preferences...
```suggestion
The index "order" values refer to the number of entries on a side that each
index system is allowed.  For instance, if we allow the particles to be
subdivided into 8 "bins" in each direction, this would correspond to an index
order of 3 (as $2^3 = 8$); correspondingly, an index order of 5 would be 32
bins in each direction, and an index order of 7 would be 128 bins in each
direction.  Each particle is then assigned a set of i, j, k values for the bin
```

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,146 @@
...
+rather than large datasets.
+
+In these situations, it is typically sufficient to set the "fine" index order
+to be much lower than its default value.  For instance, setting the
+``index_order`` to (5, 3) means that the full domain will be subdivided into 32
+bins in each dimension, and any "collision" zones will be further subdivided
+into 8 bins in each dimension (corresponding to an effective 256 bins across
+the full domain).
+
+If you are experiencing very long index times, this may be a productive
```
neutrinoceros, line 124: How long is "very" ? From reading this section I feel like the first thing I'd want to do before processing a lot of data is to find the optimal values in `index_order` for my application, but I don't know really how much I should expect to gain from it or loose if I can't afford this optimisation step.

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,146 @@
...
+Index Caching
+-------------
+
+The index values are cached between instantiation, in a sidecar file named with
+the name of the dataset file and the suffix ``.indexII_JJ.ewah``, where ``II``
+and ``JJ`` are ``index_order1`` and ``index_order2``.  So for instance, if
+``index_order`` is set to (5, 7), and you are loading a dataset file named
+"snapshot_200.hdf5", after indexing, you will have an index sidecar file named
+``snapshot_200.hdf5.index5_7.ewah``.  On subsequent loads, this index file will
+be reused, rather than re-generated.
```
neutrinoceros, line 136: where is this file written ? If I'm reading this correctly I'd assume it's generated next to the main data file, but I'm not 100% sure.

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,146 @@
...
+
+The filename scheme can be changed by supplying an alternate filename to the
+``load`` command with the argument ``index_filename``.  For instance, if you
+are accessing data in a read-only location, you can specify that the index will
+be cached in a location that is write-accessible to you.
+
+These files contain the *compressed* bitmap index values, along with some
```
neutrinoceros, line 143: Can you give an estimate of the resulting file size ? I suppose it grows as `2^index_order1 + 2^index_order2` ?

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,146 @@
...
+metadata that describes the version of the indexing system they use and so
+forth.  If the version of the index that yt uses has changed, they will be
+regenerated.
```
neutrinoceros, line 146: a question for us devs: is this determined by actually checking the `yt.__version__` constant, or is anything else taken into account. I worry that if only the version is compared, it might be a delicate process to implement changes there in between versions as it would break existing ewah files, likely without a clear error/warning.

neutrinoceros (approved):
Thank you for doing this Matt ! here are a couple of questions I had while reading it.

doc/source/reference/demeshening.rst
matthewturk: It doesn't, unfortunately -- it's compressed, so very dependent on the specific layout of the file.  In general it will be small, but I will expand this with information about this, repeating my comment.

doc/source/reference/demeshening.rst
matthewturk: It's already checked internally; it's not necessarily related to the yt version, but instead to the version of the file format.  Right now an `OSError` will be raised internally in `particle_oct_container.pyx`, which gets cause in `particle_geometry_handler.py` and this message is displayed before it continues:

```        if ver != _bitmask_version:
            raise OSError("The file format of the index has changed since "
                          "this file was created. It will be replaced with an "
                          "updated version.")
```

doc/source/reference/demeshening.rst
matthewturk: I'll add this

doc/source/reference/demeshening.rst
matthewturk: Added some discussion!

doc/source/reference/demeshening.rst
matthewturk: :)

neutrinoceros:
Thanks for your prompt response !
I'm curious wether yt will log a message to inform the user how big the ewah files end up being ? It could be done unconditionally or maybe restricted to particularly gluttonous cases such as you describe.

matthewturk:
It doesn't currently log that, no.  I don't think it's out of the question to do so, but I still think it's going to be unecessary.  A completely uncompressed set of coarse bitarrays would take up:

`(N_chunks * 2**(3*index_order1)) / 8` bytes.  So for example, this means that (assuming I did my math right) a set of 1024 chunks at `index_order1` of 5 (the default) would take up a little over 4 megs.  Now, if we assume a 10% collision rate, this would be an *uncompressed* 8 gigabytes per chunk, but this is *exceedingly* unlikely, and also completely ignores the compression (which would drop it by a few orders of magnitude).

Looking at the data in our sample collection, it seems to me that the largest index we have is about 111M, for `snipshot_399`, which is roughly a sixth of the size of the full dataset.  This is big, but it's not *terrible*, I would argue.  And it's also an `index_order` (7, 5), which is pretty fine-grained.

Anyway, I think this is something that could be discussed, but right now I only have these heuristics to go on -- and I don't know that we want to include them if they're going to potentially be weirdly wrong or something.

neutrinoceros:
If it's _possible_ to inflate disk usage by 17%, however unlikely, I think it warrants a log entry, albeit conditional. I've worked on clusters where exceeding my space quota by even 1MB would block any writing, so I don't think I'd be very happy if I had to go through yt's documentation to understand how that happened, most importantly because I would probably not _think_ of yt as a suspect as I search for the reason why I keep going over my quota.

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,161 @@
...
+     to.  Note that this is almost certainly *reductive* -- there will be fewer
+     bits set than there are particles, which is *by design*.
+  2. Once all chunks or files have been assigned an array of bits that
+     correspond to the places where, according to the coarse indexing scheme,
+     they have particles, identify all those "bits" that have been set by more
+     than one chunk.  All of these bits correspond to locations where more than
+     one file contains particles -- so if you want to select something from
+     this region, you'd need to read more than one file.
+  3. For each "collision" location, apply a *second-order* index, to identify
+     which sub-regions are touched by more than one file.
```
chummels, line 86: You're referencing a second file here, and I'm not sure what you mean?  A second chunk? It seems like any time you have more particles than there are coarse bins available, you're going to trigger this step, regardless of if the dataset is broken into multiple files.  But perhaps I'm misunderstanding.  In re-reading this a few times, I think it may be clearer if you continuing to define things as chunks instead of jumping between files and chunks.

matthewturk:
I think it's out of scope for this particular PR but we may want to add
another blocker issue.

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,161 @@
...
+
+At the end of this process, each file will be associated with a single "coarse"
+index (which covers the entire domain of the data), as well as a set of
+"collision" locations, and in each "collision" location a set of bitarrays that
+correspond to that subregion.
+
+When reading data, yt will identify which "coarse" index regions are necessary
+to read.  If any of those coarse index regions are covered by more than one
+file, it will examine the "fine" index for those regions and see if it is able
+to subset more efficiently.  Because all of these operations can be done with
```
chummels, line 96: It seems to me that the coarse index operates like a root grid and the fine index operates sort of like AMR of 1-level deep.  But I think this means that it is possible to overload an index with too many particles in some circumstances, depending on the indexing orders chosen, right?  Is this ever a concern or would it choose a second fine index for a second level of AMR?  Or would it just error out?

If I choose, (3,2) as my index order for a 3D dataset, I think it means that my coarse array supports 2**3=8 bins on any side, so 8**3 = 512 total index locations.  And then my refined index supports 2**2=4 bins on any side, so 4**3 = 64 index locations within that.  So it would only support 512*64 = 32k particles?  Is that right, or am I totally off base?  

Based on the values you have chosen as default (7,5), it seems like this won't be a problem: 2**7**3 * 2**5**3 = 2**36 ~ 7e10, but with (5,3): 2**5**3 * 2**3**3 = 2**24 ~ 2e7 it might be.

neutrinoceros:
I agree !

chummels (commented):
This is really terrific work.  The writing is clean and easy to read but also extremely informative.  And it includes practical information on how to speed up for your own datasets.  Thanks for writing this!

I included a couple of comments.  The only other additional comment is something that *could* be included in this PR or we could work out in another PR: how does yt now treat smoothed particle datasets for the bulk of its operations?  Previously, we deposited particles into octrees, and then generated things like slices and projections from that gridded data, but now we're flying free without the mesh.  I think we need a base level description of how yt conducts the line integrals that traverse these smoothing kernels is warranted.  But as I said, it doesn't have to be you who writes it, and it doesn't have to be in this PR, and I'm certainly open to discussion on this point.  Thanks for writing this PR, @matthewturk .

doc/source/reference/demeshening.rst
matthewturk: I'll see if I can clear this up with a revision to the text, but no, you will not always trigger this.

Imagine you have two files (or chunks in those files -- it doesn't actually matter, as the sub-chunk inside a file will be the atomic unit) and in the first one, it contains only particles between [0.0, 0.0, 0.0] and [1.0, 1.0, 0.5], and in file two, the particles are all [0.0, 0.0, 0.5], [1.0, 1.0, 1.0] .  There would be no collisions between these two.

Now imagine that they aren't as cleanly load balanced as that, and you have some mixing.  In those bins where both files "touch" (by means of a particle) a bin, the second phase of refined indexing will be triggered.

doc/source/reference/demeshening.rst
matthewturk: This will not happen.  The indexing is not used for anything other than choosing which files/chunks to read particles from, and then these particles are validated using the standard selection methods.

The only danger of having too coarse of an index is that too many particles will be read, found to be outside the region of interest, and discarded.  The particles are still all regarded as discrete points (or smoothed points) individually and validated that way, and the indexing system only attempts to reduce the amount of overzealous IO that precedes that validation.

You could in principle index with (0, 0).  It would just accomplish nothing, because every chunk would match for every selection, and so yt would read every single particle in the simulation, validate it's within the selector, and then either retain or discard it.

chummels: OK, I guess I was thinking of the coarse/refined indexing as an octree/kd-tree like structure, as to where the particle locations were stored and accessed.  But that seems that's not quite right here, because I'd think that when you get to the "leaf" of the tree, it's just a single particle.  I guess I'm thinking of the wrong paradigm.

jzuhone (approved):

doc/source/reference/demeshening.rst
matthewturk: I've added a bit to the text about why we're doing the indexing to begin with, which I hope will clear this up.  Let me know what you think; it's in the most recent commit.

matthewturk:
@chummels once you've taken a look, let me know which areas need mroe explanation, and if you feel it's OK, please go ahead and click "resolve."  And if they need more work, I'll get right on it!  :)

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,196 @@
...
+     which sub-regions are touched by more than one file.
+
+At the end of this process, each file will be associated with a single "coarse"
+index (which covers the entire domain of the data), as well as a set of
+"collision" locations, and in each "collision" location a set of bitarrays that
+correspond to that subregion.
+
+When reading data, yt will identify which "coarse" index regions are necessary
+to read.  If any of those coarse index regions are covered by more than one
+file, it will examine the "fine" index for those regions and see if it is able
```
chummels, line 130:
```suggestion
file, it will examine the "refined" index for those regions and see if it is able
```

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,196 @@
...
+as smoothed particle hydrodynamics, where particles have associated smoothing
+lenghts -- these are taken into account when conducting the indexing system.
+
+Efficiency of Index Orders
+--------------------------
+
+What this can lead to, however, is situations where (particularly at the edges
+of regions populated by SPH particles) the indexing system identifies
+collisions, but the relatively small number of particles and correspondingly
+large "smoothing lengths" result in a large number of "fine" index values that
```
chummels, line 145:
```suggestion
large "smoothing lengths" result in a large number of "refined" index values that
```

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,196 @@
...
+need to be set.
+
+Counterintuitively, this actually means that occasionally the "fine" indexing
```
chummels, line 148:
```suggestion
Counterintuitively, this actually means that occasionally the "refined" indexing
```

doc/source/reference/demeshening.rst
```diff
@@ -0,0 +1,196 @@
...
+process can take an inordinately long amount of time for *small* datasets,
+rather than large datasets.
+
+In these situations, it is typically sufficient to set the "fine" index order
```
chummels, line 152:
```suggestion
In these situations, it is typically sufficient to set the "refined" index order
```

chummels (approved):
Thanks for the additional notes--they helped clear up some of my misunderstanding.  Very thorough job here.  My suggestions should not hold things up, but I just noted them since you switched from "refined index" to "fine index" about halfway through.  Great writeup!

neutrinoceros:
note @matthewturk , in order for #3298 to affect this PR, you'll need to rebase or merge from master

matthewturk:
I think the merge is done and this should be ready to go!  Thanks for the review, folks.

jzuhone:
Hi @matthewturk --

Maybe not part of this PR, but one other thing that may be relevant from the perspective of the user is how slices and projections of SPH data have changed due to the demeshening, and why these are more accurate. 

I am not an expert on this stuff, otherwise I would be happy to write it myself. Maybe a couple of paragraphs would suffice?

matthewturk:
Good idea.  What if I linked to https://matthewturk.github.io/yt4-gallery/
?  I'm somewhat hesitant to check all those images in here.

jzuhone:
How about we just show one (the first one) and link to that page for the rest?

PR #3294, opened 2021-05-22
test: revert PR 3068. Reactivate a randomly failing answer test

### LINKED ISSUE #3069: Randomly failing answer test
### Bug report

**Bug summary**

Since the release of numpy 1.20.0 (and still with 1.20.1), one of our answer test is randomly failing in all PRs.

The name of the test is
`VRImageComparison_UniformGridData_vr_pitch_plane-parallel_0002`

it is generated from
`yt/visualization/volume_rendering/tests/test_vr_orientation.py`

The quick and dirty remedy is to deactivate this test (this is #3068), though it should be inspected in more details in the future so I'm opening this issue to keep track of it.

neutrinoceros:
## PR Summary

Revert #3068
Aim: fix #3069 

I'll try to reproduce the failure in 5 CI runs or less.
If it doesn't pop up, the PR could probably go in as is. If it does, we'll see if it can be fixed by increasing the precision tolerance for this specific test.

update
- run 1: passed ✅
- run 2: failed ❌ 
RMS Value: 0.005627314338711377
Tolerance:  1e-12

For posterity, here's the diff
Screenshot 2021-05-22 at 18 14 05

It seems like a big jump to bump the tolerance from 1e-3 to 1e-2...

### DIFF
--- a/yt/visualization/volume_rendering/tests/test_vr_orientation.py
+++ b/yt/visualization/volume_rendering/tests/test_vr_orientation.py
@@ -34,10 +34,13 @@ def test_orientation():
     orientations = [[-0.3, -0.1, 0.8]]
 
     theta = np.pi / n_frames
-    decimals = 12
     test_name = "vr_orientation"
 
-    for lens_type in ["perspective", "plane-parallel"]:
+    for lens_type, decimals in [("perspective", 12), ("plane-parallel", 2)]:
+        # set a much lower precision for plane-parallel tests, see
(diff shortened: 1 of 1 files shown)


chummels:
Super bizarre heisenbug.  I'm not a huge fan of loosening the tolerances, but I don't see a reasonable alternative right now.

chummels (approved):

jzuhone (approved):

PR #3295, opened 2021-05-22
Make symlog choose a better default linthresh

chummels:
## PR Summary

Following a patch (PR #3161) for an upstream bug in MPL (see Issue #2890), yt now uses symlog scaling to plot some particle-based datasets.  But the default behavior of symlog sometimes guesses poorly at the `linthresh`, the transition value between scaling the data linearly vs log.  This PR attempts to do a slightly better job of this so users get a decent plot out of the box.

I am definitely open to suggestions on how to improve this behavior.  I feel like there is a lot to unpack here, and I'm not sure this is the right way to do it, but it's a start.  In particular, I realize that by choosing a `linthresh` that is the smallest non-zero value in a dataset, it's wasting the colorbar range between 0 and `linthresh`.  I think one could do a better job by modifying the values of `linthresh` and `linscale` that are passed to MPL (see [this page](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.colors.SymLogNorm.html)) depending on the data range being plotted.

Here is an example with the FIRE sample dataset.  

```
import yt
p = yt.ProjectionPlot(ds, 'x', ('gas', 'density')).save()
```

Before this PR:

snapshot_600_Projection_x_density

And now with this PR:

snapshot_600_Projection_x_density

## PR Checklist

UPDATE1: This changes the default behavior.  Currently, if a user specifies that they want `log` scaling, and there is a negative number present, it switches to `linear`.  This PR changes that behavior to roughly remain in `log` scaling but switch to `symlog` to plot things up in negative log.  The failing answer test demonstrates this change in behavior.  I would argue that this is closer to what the user would want than switching to linear.

As noted above, this optionally allows the user to specify `symlog` scaling without specifying a `linthresh` value, which was previously not an option.  Again, I'm open to hearing people's opinions on this before I write docs and tests for this behavior, but I think it's operating as desired.

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -211,16 +211,25 @@ class ImagePlotMPL(PlotMPL):
         elif cbnorm == "linear":
             cbnorm_cls = matplotlib.colors.Normalize
         elif cbnorm == "symlog":
+            # if cblinthresh is not specified, try to come up with a reasonable default
+            vmin = float(np.nanmin(data))
+            vmax = float(np.nanmin(data))
             if cblinthresh is None:
-                cblinthresh = float((np.nanmax(data) - np.nanmin(data)) / 10.0)
+                # negative and positive range: choose val closest to zero
+                if vmin < 0 and vmax > 0:
+                    vmin_pos = np.nanmin(data[data > 0])
+                    vmax_neg = np.nanmax(data[data < 0])
+                    cblinthresh = np.min([vmin_pos, -vmax_neg])
+                # only positive range: choose smallest non-zero number
+                elif vmin >= 0 and vmax > 0:
+                    cblinthresh = np.nanmin(data[data > 0])
+                # only negative range: same as positive but cblinthresh needs to be
+                # positive!
+                else:
+                    cblinthresh = -np.nanmax(data[data < 0])
+                cblinthresh = float(cblinthresh)
 
-            cbnorm_kwargs.update(
-                dict(
-                    linthresh=cblinthresh,
-                    vmin=float(np.nanmin(data)),
-                    vmax=float(np.nanmax(data)),
-                )
-            )
+            cbnorm_kwargs.update(dict(linthresh=cblinthresh, vmin=vmin, vmax=vmax))
             MPL_VERSION = LooseVersion(matplotlib.__version__)
             if MPL_VERSION >= "3.2.0":
                 # note that this creates an inconsistency between mpl versions
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -213,21 +213,16 @@ class ImagePlotMPL(PlotMPL):
         elif cbnorm == "symlog":
             # if cblinthresh is not specified, try to come up with a reasonable default
             vmin = float(np.nanmin(data))
-            vmax = float(np.nanmin(data))
+            vmax = float(np.nanmax(data))
             if cblinthresh is None:
                 # negative and positive range: choose val closest to zero
                 if vmin < 0 and vmax > 0:
                     vmin_pos = np.nanmin(data[data > 0])
                     vmax_neg = np.nanmax(data[data < 0])
                     cblinthresh = np.min([vmin_pos, -vmax_neg])
-                # only positive range: choose smallest non-zero number
-                elif vmin >= 0 and vmax > 0:
-                    cblinthresh = np.nanmin(data[data > 0])
-                # only negative range: same as positive but cblinthresh needs to be
-                # positive!
+                # only positive or negative range: choose smallest non-zero number
                 else:
-                    cblinthresh = -np.nanmax(data[data < 0])
-                cblinthresh = float(cblinthresh)
+                    cblinthresh = np.nanmin(np.absolute(data)[np.absolute(data) > 0])
 
             cbnorm_kwargs.update(dict(linthresh=cblinthresh, vmin=vmin, vmax=vmax))
             MPL_VERSION = LooseVersion(matplotlib.__version__)
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -215,14 +215,7 @@ class ImagePlotMPL(PlotMPL):
             vmin = float(np.nanmin(data))
             vmax = float(np.nanmax(data))
             if cblinthresh is None:
-                # negative and positive range: choose val closest to zero
-                if vmin < 0 and vmax > 0:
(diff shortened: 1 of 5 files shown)


chummels:
It seems like these test failures have nothing to do with this PR.

jzuhone:
@neutrinoceros said that there were some test issues earlier today.

yt/visualization/plot_window.py
```diff
@@ -992,22 +992,19 @@ def _setup_plots(self):
...
-                    msg = (
-                        "Plot image for field %s has no positive "
-                        "values.  Max = %f." % (f, np.nanmax(image))
+                    msg = "Plotting %s: All negative values. Max = %f." % (
+                        f,
+                        np.nanmax(image),
                     )
+                    use_symlog = True
                 elif not np.any(np.isfinite(image)):
-                    msg = f"Plot image for field {f} is filled with NaNs."
```
jzuhone, line 1005: I see you had to rewrite this, but is there a reason that you converted it away from an f-string?

jzuhone:
The Jenkins failures look potentially real though?

neutrinoceros:
Indeed the failure you're getting isn't specific to this PR.

yt/visualization/plot_window.py
chummels: My mistake.  Sorry, @jzuhone .  I was just trying to shrink them all so each fit on a single line.  I can take these changes out if people are against it though.  I'll change it back to an f-string.  I think black is going to force them all to be f-strings.

chummels:
I realize that `set_log()` has now been cannibalized to allow `symlog` explicitly without setting a `linthresh`.  Ideally, it would be best to refactor `set_log` to be something like a `set_scaling()` function with args={`log`, `linear`, or `symlog`}, but I wasn't ready to do something super backwards incompatible here.  But maybe that's what *should* be done since this is the big release? I dunno.  It seems like an awfully big change.  Note that this part isn't required for the auto-selection of the `linthresh` to fix the bug, but it seemed like a useful thing to do so people could select `symlog` without selecting a `linthresh`.

chummels:
> The Jenkins failures look potentially real though?

Yes, you're right.  I mixed up two arguments in a string formatting.  This should now be fixed.

yt/visualization/plot_window.py
jzuhone: No worries!

neutrinoceros:
Just my two cents but I think the most general solution would be to have a set_norm, since Matplotlib is adding new ones in recent release and there are much more than 3 already, it seems like the most flexible option to just allow any norm object to be passed into yt. However it _is_ an important change and since the discussion has started pretty late into our effort to release yt 4.0 I'd argue that it's preferable to not make such a change now, but preferably early in the transition to 4.1 so we have more time to stabilize it.

neutrinoceros:
@chummels I've attempted to relaunch the GH actions jobs here but they still failed even though the patch was merged in #3298. Can you rebase (or merge from master) ?

jzuhone:
I think the answer test failure here looks acceptable based on what you did. You'll just need to update the answer number for this test in `tests.yaml`

jzuhone:
Or maybe not. That failed test image looks better with a linear scale.

chummels:
While I agree that the failed test image *looks better* in linear scale, I would argue that the desire behavior when someone sets `log` scaling and has negative values is to switch to symlog and still plot it in `log` scaling  but in negative numbers.

jzuhone:
True!

neutrinoceros:
> While I agree that the failed test image looks better in linear scale, I would argue that the desired behavior when someone sets log scaling and has negative values is to switch to symlog and still plot it in log scaling but in negative numbers, which is what this PR does.

I agree, but I'm not 100% sure that the `log` scale is indeed explicitly requested in this test. In anycase, choosing a good default for scaling is _hard_ because anything other than log/symlog is irrelevant in cosmology (so I understand), but there are many fields in which log scale is rarely if ever relevant.
Since no default can possibly be equally satisfying to everyone (and let's face it, yt seems to still have a cosmologist-heavy user base), _consistently_ using log/symlog by default everywhere seems like the best course of action for now.

neutrinoceros (approved):

jzuhone:
@chummels if you update the answer tests then I'm happy to approve.

neutrinoceros:
@chummels updating this test only requires bumping a number in tests/tests.yaml, I think. The game is to guess which one :)

neutrinoceros:
@chummels here's the patch you need, for this line 
https://github.com/yt-project/yt/blob/d7b1b994ee5d30882b276e7275a75fbf147a523e/tests/tests.yaml#L170

```patch
-    local_cylindrical_background_006:  # PR 3153
+    local_cylindrical_background_007:  # PR 3295
```

jzuhone:
@chummels are you good with @neutrinoceros's patch?

chummels:
Yeah, I can include that.  I was nominally trying to add some more tests for the symlog_auto kwarg and a few symlog plots, but maybe this will complicate things too much in terms of new tests?

neutrinoceros:
When passing existing tests is high enough of a fence that it doesn't happen by accident, I don't think that adding new tests is required, afaic

Xarthisius:
@yt-fido test this please.

jzuhone:
@chummels can you rebase this on the main branch? Should fix the test failures.

yt/visualization/base_plot_types.py
```diff
@@ -211,16 +211,20 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
...
-                )
-            )
+                # negative and positive range: choose val closest to zero
+                if vmin < 0 and vmax > 0:
+                    vmin_pos = np.nanmin(data[data > 0])
+                    vmax_neg = np.nanmax(data[data < 0])
+                    cblinthresh = np.min([vmin_pos, -vmax_neg])
+                # only positive or negative range: choose smallest non-zero number
+                else:
+                    cblinthresh = np.nanmin(np.absolute(data)[np.absolute(data) > 0])
```
cphyc, line 225: I am unsure this is necessary. Wouldn't the following simplification achieve exactly the same goal?
```suggestion
                cblinthresh = np.nanmin(np.absolute(data)[data != 0])
```

cphyc (approved):

yt/visualization/base_plot_types.py
neutrinoceros: agreed

chummels: This is a great simplification!  Thank you!  I will add this when I figure out how to fix my darned PR.

jzuhone (approved):

PR #3298, opened 2021-05-24
Pin proj to 7.x per SciTools/cartopy#1140

Xarthisius:
Pin proj library on osx until SciTools/cartopy#1140 is fixed.

## PR Summary

Should fix failing OSX build.

PR #3299, opened 2021-05-24
ci: pin proj to 7.x version on OsX while cartopy isn't updated

neutrinoceros:
## PR Summary

this should fix the build failures we're currently getting in OsX CI.
Thanks @Xarthisius for unraveling the issue


neutrinoceros:
🤦🏻  just saw that there's #3298 already. Apologies

PR #3306, opened 2021-05-25
Update quickstart

chummels:
Updates the quickstart notebook docs to work with yt4 and fix failures.  Note, I cannot update the https://yt-project.org/docs/dev/quickstart/data_objects_and_time_series.html notebook to fix the failures.  That should be done by someone with more knowledge of the current state of the `HaloFinder` machinery like @brittonsmith .


brittonsmith:
I don't have time to do this. I think it's better to leave halo analysis documentation to yt-astro-analysis anyway. I would just remove that section entirely.

chummels:
Removed.

matthewturk (approved):

neutrinoceros (approved):

PR #3309, opened 2021-05-28
fix load_sample's signature

### LINKED ISSUE #3297: load_sample: keyword-only arguments
### Bug report

**Bug summary**

The signature in yt.load_sample should be updated from
```python
def load_sample(
    fn: Optional[str] = None, progressbar: bool = True, timeout=None, **kwargs
):
```
to
```python
def load_sample(
    fn: Optional[str] = None, *, progressbar: bool = True, timeout=None, **kwargs
):
```
Rationale: there are reminiscent calls to load_sample in our cookbooks using the original signature, where the second positional argument was `specific_file`, which is now effectively evaluated as a bool (`progressbar`), so it should be considered a silent bug.
I'm not going to include this change in #3262 for now because CI is currently broken, and I _expect_ this is going to create some failures.
This is mostly a reminder for when CI is op again.

neutrinoceros:
## PR Summary

fix #3297

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1304,7 +1304,7 @@ def load_unstructured_mesh(
 
 # --- Loader for yt sample datasets ---
 def load_sample(
-    fn: Optional[str] = None, progressbar: bool = True, timeout=None, **kwargs
+    fn: Optional[str] = None, *, progressbar: bool = True, timeout=None, **kwargs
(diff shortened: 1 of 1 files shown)


cphyc (approved):

matthewturk (approved):
dang, nice use of the asterisk

PR #3310, opened 2021-05-28
Update docstrings to be consistent with function signature

cphyc:
## PR Summary

With time, many docstrings have grown out of sync with their actual function. This PR attemps to fix as many as possible.

This PR fixes as many as possible:
- [x] Non-compliant docstrings (e.g. `* ds : Dataset` should be `ds : Dataset`)
- [x] Arguments mentioned in the docstring that do not exist in the function signature
- [x] Plot callbacks and fixed resolution buffer filters had their signatures incorrectly guessed (see screenshots). Note that the best solution would be to refactor the callbacks and remove the current two hacks to get the proper function signature. See issue #3311 .

The mismatches have been found using https://gist.github.com/cphyc/afc1a143943220ec3a1c5ca16e14cd13 which allows to compare the list of argument parsed from the docstring to the actual signature of the function.

## Currently on main
```python
import yt
p = yt.SlicePlot(ds, "x", "density")
p.annotate_sphere?
```
On main (`center` is missing from the signature):
image

With this PR:
image

## Note on the signature issue
The origin of the signature mismatch is due to the weird structure we are using to register plot annotation callbacks. Each callback is a class which constructor we were binding to the caller PlotWindow instance using `types.MethodType`. This function alters the displayed signature by dropping the first argument which is assumed to be `self`. In our case, because we are binding a constructor method rather than a function, the first argument in the signature is actually the first argument _after_ `self` in the `__init__` method.

Here's a simple example:
```
>>> import types
>>> class Foo:
...     def __init__(self, arg1, arg2):
...         pass
...         
... class Bar:
...     def __init__(self):
...          self.foo = types.MethodType(Foo, self)
... 
... b = Bar()
... b.foo?
Signature: b.foo(arg2)   # ! Should have been b.foo(arg1, arg2)
Docstring: <no docstring>
Type:      method
```


yt/loaders.py
```diff
@@ -1065,9 +1065,9 @@ def load_unstructured_mesh(
     coordinates,
     node_data=None,
     elem_data=None,
-    length_unit=None,
     bbox=None,
     sim_time=0.0,
+    length_unit=None,
```
neutrinoceros, line 1070: If you're worried about order consistency, please sort the docstring, not the signature itself. Specifically positional arguments (keyword-only arguments can be safely reordered without breaking backwards compatibility however)

yt/visualization/color_maps.py
```diff
@@ -312,7 +312,7 @@ def make_colormap(ctuple_list, name=None, interpolate=True):
         If you wish this colormap to be added as a valid colormap to the
         current session, specify a name here.  Default: None
 
-    interpolation: boolean, optional
+    interpolate: boolean, optional
```
neutrinoceros, line 315: let's fix the wording too ! "optional" should map to `typing.Optional`, meaning that "this can be `None`"
(I'm aware that `None` evaluates as False so technically `None` is a usable value here, but it's neither the default value nor a specific check, so there is no reason I can see to mark this as optional)
```suggestion
    interpolate: boolean
```

yt/visualization/fits_image.py
```diff
@@ -616,7 +616,7 @@ def writeto(self, fileobj, fields=None, overwrite=False, **kwargs):
         fields : list of strings, optional
             The fields to write to the file. If not specified
             all of the fields in the buffer will be written.
-        clobber : overwrite, optional
+        overwrite : boolean, optional
```
neutrinoceros, line 619:
```suggestion
        overwrite : boolean
```

yt/visualization/line_plot.py
```diff
@@ -222,7 +222,7 @@ def from_lines(
         figure_size : int or two-element iterable of ints
             Size in inches of the image.
             Default: 5 (5x5)
-        fontsize : int
+        font_size : int
```
neutrinoceros, line 225: I agree with the annotated type, but the default value's type should also reflect it, can you fix it too ?

neutrinoceros (commented):
thank's for doing this, here are a couple suggestions

yt/visualization/line_plot.py
cphyc: Should be good now!

yt/visualization/fixed_resolution.py
```diff
@@ -505,10 +505,31 @@ def setup_filters(self):
             if key in ignored:
                 continue
             filtername = filter_registry[key]._filter_name
-            FilterMaker = filter_registry[key]
-            filt = apply_filter(FilterMaker)
-            filt.__doc__ = FilterMaker.__doc__
-            self.__dict__["apply_" + filtername] = types.MethodType(filt, self)
+
+            # We need to wrap to create a closure so that
```
cphyc, line 509: I have updated the PR description to explain why this is required.

neutrinoceros (commented):
I'm happy to have this go in as is given the fact we have a freshly opened issue to remind us that this code is due for a deeper refactor. Thanks again !

neutrinoceros (approved):

PR #3317, opened 2021-06-01
BUG: fixup amrvac sample data file names in tests

### LINKED ISSUE #3315: CI: update amrvac tests with new datafilee
Following #3280, AMRVAC test using data sample must be updated to use the new files.

neutrinoceros:
## PR Summary

fix #3315
substitutions were done semi-automatically with the [apply-subs](https://pypi.org/project/apply-subs/) CLI, using this input file:
```json
{
    "bw_cartesian_3d/output0001.dat": "amrvac/bw_3d0000.dat",
    "bw_cylindrical_3d/output0001.dat": "amrvac/bw_cylindrical_3D0000.dat",
    "bw_polar_2d/output0001.dat": "amrvac/bw_polar_2D0000.dat",
    "bw_spherical_2d/output0001.dat": "amrvac/bw_2d0000.dat",
    "kh2d/output0001.dat": "amrvac/kh_2d0000.dat",
    "kh3d/output0001.dat": "amrvac/kh_3D0000.dat",
    "riemann1d/output0001.dat": "amrvac/R_1d0005.dat"
}
```
and using the following command
```shell
git ls-files | egrep '.py$' | xargs apply-subs -s subs.json --inplace
```

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -3,7 +3,7 @@ answer_tests:
   local_art_003: # PR 3081, 3101
     - yt/frontends/art/tests/test_outputs.py:test_d9p
 
-  local_amrvac_008: # PR 2945
+  local_amrvac_2_001: # PR 3317
     - yt/frontends/amrvac/tests/test_outputs.py:test_domain_size
     - yt/frontends/amrvac/tests/test_outputs.py:test_bw_polar_2d
     - yt/frontends/amrvac/tests/test_outputs.py:test_blastwave_cartesian_3D
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -3,16 +3,34 @@ answer_tests:
   local_art_003: # PR 3081, 3101
     - yt/frontends/art/tests/test_outputs.py:test_d9p
 
-  local_amrvac_2_001: # PR 3317
+  local_amrvac_domain_size_001: # PR 3317
     - yt/frontends/amrvac/tests/test_outputs.py:test_domain_size
+
+  local_amrvac_bw_polar_2D_001: # PR 3317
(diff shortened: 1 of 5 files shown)


neutrinoceros:
@Xarthisius we may have a problem here. Answer tests are failing on Jenkins. The is expected, since some files changed names, but the issue is that only the filename (not the containing directory) seems to be used to name the storage file, while several of the new files are named `output0001.dat`, so I don't think we can avoid collisions here
```
There is no old answer available.
output0001.dat : GridHierarchy_output0001_dat_all
```
Is there any pre-existing mechanism to help us here ? I'm thinking we'd need to change the naming convention for those storage files, but that would imply changing ALL of them, probably, which seems a little too much.

neutrinoceros:
@yt-fido test this please

Xarthisius:
For now you can split them into separate answer files. There's no rule saying it has to be one file per frontend. Use as many as you need. The issue will go away with pytest migration.

neutrinoceros:
Nice. I think it's safe to drop the old ones completely so I'm going to replace the old file. I hope that's ok !

Xarthisius:
> Nice. I think it's safe to drop the old ones completely so I'm going to replace the old file. I hope that's ok !

That doesn't solve the issue that all datasets have the same ds_name, which will conflict in the answer file.

neutrinoceros:
oh that's right, thanks for your quick reply !

neutrinoceros:
I forgot to review this and now the logs are gone...
@yt-fido test this please

neutrinoceros:
@Xarthisius I guess I must have done something wrong here, possibly corrupted my own answers... any chance you could take a look ?

PR #3318, opened 2021-06-02
remake #3062

neutrinoceros:
## PR Summary

#3062 broke since it's last conflict-solving rebasing, and the one giant commit is hard to navigate in search of the one(two) mistake(s) I made, so here's a second attempt. I'm going to push files in batches of 10 (or 20 if I'm lazy) until I hit the error on Jenkins, then I'll fix it and resume. Pre-commit will validate when I correctly include _all_ changes from #3062

here's a one liner I'll use to perform these gradual changes semi-automatically:
```shell
git diff --name-only flake8-return | head -n 10 | xargs git checkout flake8-return
```
where flake8-return is the name of the branch attached to #3062

update: I know think that the failures on #3062 are not specific to _that_ branch, but are consistently popping on every build, so _this_ PR is likely not necessary at all. I'll close it when the situation resolves. (reported as #3319)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,7 +48,8 @@ repos:
     - id: flake8
       additional_dependencies: [
         flake8-bugbear>=20.3.2, # GH PR 2851
-        flake8-logging-format
+        flake8-logging-format,
+        flake8-return,
       ]
 -   repo: https://github.com/asottile/blacken-docs
     rev: v1.10.0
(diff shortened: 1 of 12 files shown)


matthewturk:
I don't think we should prioritize this just now.  I am personally not that into the changes to the return statements anyway, but I think more broadly this should be discussed.

One reason I am particularly not interested in this is that having it assigned to a local variable does no *harm* and additionally will make it easier to *debug*.

Issue #3319, opened 2021-06-02
CI: failures on the main branch (?)


neutrinoceros:
### Bug report

**Bug summary**

Jenkins CI is apparently broken as some 44 failures are showing up on different PRs (#3234, #3236, #3270, #3295, to name a few I could identify)
see for instance : PR #3295 https://tests.yt-project.org/job/yt_py38_git/3236/#showFailuresLink

The errors seem to be related to incorrect field access, though it's not clear to me what change may have caused this.
@Xarthisius, do you know if this could be a result of changes on the server, rather than on the main branch of the repo ?

PR #3326, opened 2021-06-02
Make uniq_id a little bit more unique

Xarthisius:
## PR Summary

Add some more uniqueness to `unique_identifier`

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -43,6 +43,7 @@ from yt.utilities.exceptions import (
     YTIllDefinedParticleFilter,
     YTObjectNotImplemented,
 )
+from yt.utilities.lib.fnv_hash import fnv_hash
 from yt.utilities.minimal_representation import MinimalDataset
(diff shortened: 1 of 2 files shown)


matthewturk (approved):
This seems like a definite improvement.

neutrinoceros (approved):

neutrinoceros:
This is intended as a fix to https://github.com/yt-project/yt/issues/3319, right ?

Xarthisius:
> This is intended as a fix to #3319, right ?

Yes, but now I see new errors that are all related to cached ewah files

matthewturk:
I can help to address those if need be, but the ewah file caching may need
additional information, as right now it's a sidecar file that only uses
filename, ewah index version, and the order.

Xarthisius:
Questions are: why it's being generated during every run, and why it's messing up the field detection.

Xarthisius:
My latest commit should fix the failures, but it's only a workaround. Actual bug: #3327

jzuhone (approved):

PR #3330, opened 2021-06-03
load_sample: fix a handful of broken load names

neutrinoceros:
## PR Summary

This (with #3328 and #3329) exhausts the list of sample data files that can't currently load with `yt.load_sample` because of a simple "mistake" in the local json registry.
All of them can _download_ properly on the main branch already, but this patch is necessary to make them actually loadable with `yt.load_sample`.

I should add that all these changes were tested on my machine, and are in no way automated.

### DIFF
--- a/yt/sample_data_registry.json
+++ b/yt/sample_data_registry.json
@@ -158,7 +158,7 @@
   "GasSloshing.tar.gz": {
     "hash": "3e16c5975f310bafd1e859080310de7ca01d5aa055b3504c10138c93154202a4",
     "load_kwargs": {},
-    "load_name": "GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150",
+    "load_name": "sloshing_nomag2_hdf5_plt_cnt_0150",
     "url": "https://yt-project.org/data/GasSloshing.tar.gz"
   },
   "GasSloshingLowRes.tar.gz": {
@@ -326,13 +326,13 @@
   "MOOSE_Sample_data.tar.gz": {
     "hash": "0a43e40931e25c4a0e4f205b9902666c2d7a89cd5959e2d5eb7819059cabe01f",
     "load_kwargs": {},
-    "load_name": "MOOSE_sample_data/out.e",
+    "load_name": "out.e",
     "url": "https://yt-project.org/data/MOOSE_Sample_data.tar.gz"
   },
   "MoabTest.tar.gz": {
     "hash": "22609631c0e417964e0247b4ad62199a60c2d7979f38f8587960a9f8b2147fab",
     "load_kwargs": {},
-    "load_name": "MoabTest/fng_usrbin22.h5m",
+    "load_name": "fng_usrbin22.h5m",
     "url": "https://yt-project.org/data/MoabTest.tar.gz"
   },
   "MultiRegion.tar.gz": {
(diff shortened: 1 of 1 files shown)


cphyc (approved):
This looks fine to me - though I'd rest reassured if I could test this somehow.

neutrinoceros:
adding systematic tests for all datasets is virtually impossible. I just ran a script to do exactly this yesterday and it took 3h to complete. Some of the datasets fixed here are several hundreds MB, it's already too much for systematic testing.
You don't have to trust me however, but it _does_ take a few minutes to test this manually.
I intend to produce many such fixes, and rerun my 3h script to produce a before/after report, hopefully just in time for the big release :)

neutrinoceros:
oops I read "if *you* could test this somehow" (as in *me*). You meant that you don't know how to double check this, sorry.

Here's how you do it, taking for instance the `"MoabTest"` dataset
```
$ python -c "import yt; yt.load_sample('MoabTest')"
```
on the main branch, this currently fails with a `FileNotFoundError`, even though the data is in fact properly downloaded and untared (this description is general to all datasets I'm fixing here). The thing is that `yt.load_sample` assumes that `MoabTest` is also the name of the directory that the tarball is going to unpack to, and the `load_name` is then joined on top of _that_ to produce a fully functional path. Clearly this convention wasn't followed for. The datasets I'm fixing here (not blaming anyone, this convention only exists because it's the most common case, but it's not currently documented).

In general, the string you should pass to `yt.load_sample` should match the name of the tarball, `.tar.gz` removed.

yt/sample_data_registry.json
```diff
@@ -326,13 +326,13 @@
   "MOOSE_Sample_data.tar.gz": {
     "hash": "0a43e40931e25c4a0e4f205b9902666c2d7a89cd5959e2d5eb7819059cabe01f",
     "load_kwargs": {},
-    "load_name": "MOOSE_sample_data/out.e",
+    "load_name": "out.e",
```
chummels, line 329: This seems pretty generic to have such a short filename.

chummels (commented):
Is the `load_name` the identifier that you pass to `load_sample()` in order to tell it which dataset you are trying to load?  If so, then I think this PR is somewhat problematic.

Generally when I think about the sample datasets, I don't think of the specific filename, I think of the directory.  So for example, `IsolatedGalaxy` is the rare dataset where I actually *know* the filename associated with it `galaxy0030/galaxy0030`, but most of them I just know the directory as the identifier.  But given that we advertise these datasets by their directory on `https://yt-project.org/data`, doesn't it seem appropriate for the identifier associated with them to *be* the directory and not the filename?  

But perhaps I'm misunderstanding what the `load_name` parameter actually references in `load_sample()`?

neutrinoceros:
> Is the load_name the identifier that you pass to load_sample() in order to tell it which dataset you are trying to load? If so, then I think this PR is somewhat problematic.

@chummels it's not ! It really is just a path, relative to the directory that the tarball gets unfolded to.
And you can also see it as a "default" value for tarballs that contain actual time series, and you're always free to change it when you call load_sample if you want to load a different snap shot.

Right now those are literally broken, as they don't comply to this specification. I believe they were written while the specs were still a little confusing, and I'm rectifying them as it's becoming clearer what the dominant rule is.

neutrinoceros:
> But given that we advertise these datasets by their directory on https://yt-project.org/data, doesn't it seem appropriate for the identifier associated with them to be the directory and not the filename?

This is _exactly_ the rule I'm enforcing here :)

chummels:
OK, awesome!  Thanks for the clarification, @neutrinoceros !

chummels (approved):

PR #3332, opened 2021-06-03
Fix #3314 by allowing user customization

### LINKED ISSUE #3314: Custom RAMSES gravity files
I have customize some RAMSES simulations and now I have extra gravity fields inside my file grav.out.
Trying to read them with YT causes a bug because, the file size is expected to be smaller.
```
ad['acceleration_x']
ValueError: Size obtained (8) does not match with the expected size (4) of record
```
Could you point out to me where/how to customize YT so that I can handle the new fields ?

cphyc:
## PR Checklist


yt/frontends/ramses/field_handlers.py
```diff
@@ -263,6 +264,18 @@ def offset(self):
...
         return self._offset
 
+    @classmethod
+    def load_fields_from_yt_config(cls) -> Tuple[List[str], bool]:
+        if cls.config_field and ytcfg.has_section(cls.config_field):
+            # Or this is given by the config
+            cfg = ytcfg.get(cls.config_field, "fields")
+            fields = []
+            for field in (_.strip() for _ in cfg.split("\n") if _.strip() != ""):
+                fields.append(field.strip())
```
neutrinoceros, line 274: looks like a leftover from an intermediate step
```suggestion
            for field in (_.strip() for _ in cfg.split("\n") if _.strip() != ""):
                fields.append(field)
```

yt/frontends/ramses/field_handlers.py
```diff
@@ -263,6 +264,18 @@ def offset(self):
         self._level_count = level_count
         return self._offset
 
+    @classmethod
+    def load_fields_from_yt_config(cls) -> Tuple[List[str], bool]:
```
neutrinoceros, line 268: The way I read the function body, it looks like the bool should match `bool(fields)` (except in the case a section is found but no fields are specified, which to me would actually look like a bug ?)
I'd suggest simplifying the function so that it matches this signature (or alternatively give an explanation of why this wouldn't work)
```suggestion
    def load_fields_from_yt_config(cls) -> List[str]:
```

neutrinoceros (commented):
quick first pass, I'll review completely later !

doc/source/examining/loading_data.rst
```diff
@@ -2593,7 +2593,7 @@ Any field data or material data on the mesh can then be viewed just like any oth
 RAMSES Data
 -----------
 
-In yt-3.0, RAMSES data is fully supported.  If you are interested in taking a
+In yt-4, RAMSES data is fully supported.  If you are interested in taking a
```
neutrinoceros, line 2596:
```suggestion
In yt-4.x, RAMSES data is fully supported.  If you are interested in taking a
```

PR #3333, opened 2021-06-03
Remove unused RAMSES dataset

cphyc:
This simply remove a dataset which isn't used anywhere in the code and that cannot be loaded from `load_sample`.

This is due to the `.tar.gz` having the structure:
```
ramses_extra_fields
├── amr_00001.out00001
├── amr_00001.out00002
├── amr_00001.out00003
├── amr_00001.out00004
├── amr_00001.out00005
├── amr_00001.out00006
├── amr_00001.out00007
├── amr_00001.out00008
├── compilation.txt
├── cooling_00001.out
├── grav_00001.out00001
├── grav_00001.out00002
├── grav_00001.out00003
├── grav_00001.out00004
├── grav_00001.out00005
├── grav_00001.out00006
├── grav_00001.out00007
├── grav_00001.out00008
├── header_00001.txt
├── hydro_00001.out00001
├── hydro_00001.out00002
├── hydro_00001.out00003
├── hydro_00001.out00004
├── hydro_00001.out00005
├── hydro_00001.out00006
├── hydro_00001.out00007
├── hydro_00001.out00008
├── hydro_file_descriptor.txt
├── info_00001.txt
├── makefile.txt
├── namelist.txt
├── part_00001.out00001
├── part_00001.out00002
├── part_00001.out00003
├── part_00001.out00004
├── part_00001.out00005
├── part_00001.out00006
├── part_00001.out00007
└── part_00001.out00008
```
Valid RAMSES datasets should always be contained in a folder named `output_XXXXX` where `XXXXX` should match the value in the `info_XXXXX.txt` filename. In this case, the structure should be `ramses_extra_fields/output_00001/` for it to be loaded properly.

### DIFF
--- a/yt/sample_data_registry.json
+++ b/yt/sample_data_registry.json
@@ -777,12 +777,6 @@
     "load_name": "output_00003/info_00003.txt",
     "url": "https://yt-project.org/data/ramses_empty_record.tar.gz"
   },
-  "ramses_extra_fields.tar.gz": {
-    "hash": "f3cc7b95bd527c33c095e3fff04c083fcb9681629bdfba359deff2871e765e3f",
-    "load_kwargs": {},
-    "load_name": "info_00001.txt",
-    "url": "https://yt-project.org/data/ramses_extra_fields.tar.gz"
-  },
   "ramses_extra_fields_small.tar.gz": {
     "hash": "8d2dde6e6150f0767ebe3ace84e2e790503154c5d90474008afdfecbac68ffed",
     "load_kwargs": {},


neutrinoceros:
I agree there is no motivation to fix this if it isn't used anywhere, and I'd rather remove it completely.
Beware that you'd need to open a complementary PR on the website repo to remove registration there too !

matthewturk:
So, if we did fix the structure, would it be a good dataset to use to check
extra fields?

cphyc:
@matthewturk no, not really. We already have `ramses_extra_fields_small` which does the same job but is much smaller.

neutrinoceros (approved):

Xarthisius:
Just as comment, not a review of this PR, but I don't like the fact that `load_sample` suddenly dictates how we structure CI or what datasets we support. Both, all previous versions of AMRVAC dataset, and this Ramses dataset are perfectly fine and can be loaded into yt. While `load_sample` is nice for new users, there's **no need** for it to be able to load **ALL** datasets from yt-project.org... Should I known `load_sample` will end up so disruptive I'd oppose adding it in the first place.

neutrinoceros:
I'm sorry this is causing you pain @Xarthisius. I'm sincerely hoping https://github.com/yt-project/website/pull/103 will be the last of its kind, and it's intended as such.

The original version of load_sample wasn't thoroughly tested, but was merged regardless... I'm really trying to cause as little disruption as possible while still making it robust enough for shipping. I'm really sorry so much of it is falling onto you.

cphyc:
@Xarthisius I'm just commenting for the content of this very specific PR, not on the more global comment regarding the relation with `load_sample`.

The dataset I am removing *cannot* be loaded as is by either `yt.load` or `yt.load_sample` due to the reason I detailed in the initial comment. This could be easily fixed (simply move all stuff into a folder named `output_00001` in the tarball), but since we're not using it anywhere I thought I would delete it.

PR #3334, opened 2021-06-03
load_sample: fix sample data registry for 7 datasets

neutrinoceros:
## PR Summary

This is complementary with https://github.com/yt-project/website/pull/103 and should fix load_sample for 6 more datasets that were not following the dominant naming conventions in tarballs

this will also need to be followed by a systematic cleanup akin to #3317

### DIFF
--- a/yt/sample_data_registry.json
+++ b/yt/sample_data_registry.json
@@ -771,6 +771,12 @@
     "load_name": "groups_008/group_008.0.hdf5",
     "url": "https://yt-project.org/data/owls_fof_halos.tar.gz"
   },
+  "ParticleCavity.tar.gz": {
+    "hash": "37b4da16294cf139510dca2b930084a9e5111f8129c96338d03031804cd36803",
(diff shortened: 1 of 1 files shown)


cphyc:
Why are the hashes changing?

neutrinoceros:
Because the original tarballs didn't follow the de facto rule implemented but every other data sample: files need to unfold to a directory with the same name as the tarball (minus the .tar.gz extension), which also matches the "key" that's passed to load_sample

neutrinoceros:
this will very likely be superseded by #3366 
Switching to draft for now, but I except this will end up being closed rather than merged.

neutrinoceros:
Probably, but I'd like to keep it around for now if you don't mind.

neutrinoceros:
closing this to clear the backlog as much as possible.
#3366 is the only change we need, and the one performed here can be done without a PR when #3366 is completed

PR #3336, opened 2021-06-05
load_sample: fix datasets without a loadname + bugfix

neutrinoceros:
## PR Summary

This fixes a few leftovers from  #2699+#2951: datasets that cannot  currently load with `yt.load_sample` because they lack a default "load_name".
I needed to tweak `yt.load_sample` itself, and I want to stress that I've performed a systematic check that the change *doesn't* break any of the previously working cases (the check takes a few *hours* to download the whole database).
Note to reviewers: I encourage you to manually test this patch (I already did), but note that some of the datasets that this patch fixes are pretty big (between 5 and 15GB), so it might take a little while for them do download.

### DIFF
--- a/yt/sample_data_registry.json
+++ b/yt/sample_data_registry.json
@@ -116,7 +116,7 @@
   "Enzo_64.tar.gz": {
     "hash": "2b17fffb6a2e8e471ef85dbdbfa5568f09f7bdb77715c69b40dfe48be8f71bc9",
     "load_kwargs": {},
-    "load_name": null,
+    "load_name": "DD0043/data0043",
     "url": "https://yt-project.org/data/Enzo_64.tar.gz"
   },
   "ExodusII_tests.tar.gz": {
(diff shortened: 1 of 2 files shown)


yt/sample_data_registry.json
```diff
@@ -726,7 +726,7 @@
   "maestro_xrb_lores_23437.tar.gz": {
     "hash": "a17b63609a2c50a829e26420f093ee898237b1ef8256ef38ee8be56a6f824460",
     "load_kwargs": {},
-    "load_name": "",
+    "load_name": null,
```
neutrinoceros, line 729: In case you're wondering what I did here, this doesn't affect the program functionally, but it was the only dataset left with an empty string for a load_name, so I figure we should embrace `null` as a signal that the the directory from the tarball _is_ the loadable path (as opposed to the general case where we need to append a load_name to form a complete loadable path).
There might be other non-functional datasets left in the repo using `"load_name": null` and which fail in different ways. I'll fix those in a followup patch if needed.

matthewturk (approved):

PR #3346, opened 2021-06-08
Allow to set units from config file

### LINKED ISSUE #3345: Allow users to provide list of known fields and their units
### Bug report

**Bug summary**

At the moment, on-disk fields have their default units and string representation stored in `FieldInfoContainer` subclasses. This prevents users from customizing them, and also implies that unknown fields will all default to being dimensionless.
This was brought up in #3332 and on the chat by @dnarayanan .

**Code for reproduction**

For example, in the following
```python
>>> ds = yt.load_sample("output_00080")
>>> ad = ds.all_data()
... ad["nbody", "particle_extra_field_1"]
unyt_array([0., 0., 0., ..., 0., 0., 0.], '(dimensionless)')
```
there is now way to tell yt that `particle_extra_field_1` should have given units (in this case I _think_ it is homogenous to a time).
At the moment, the only possibility is to create a derived field.

**Suggested solution**

Allow the user to list entries in the configuration file following this format:
```toml
[fields.nbody.particle_extra_field_1]
aliases = "particle_other_fancy_name"
# or
aliases = ["particle_other_fancy_name", "particle_alternative_fancy_name"]
units = "code_time"
display_name = "Dinosaurs Density"
```

cphyc:
Fixes #3345.

This allows the user to customize the fields automatically detected by yt from the config file using the following syntax
```toml
[fields.nbody.particle_extra_field_1]
aliases = ["particle_other_fancy_name", "particle_alternative_fancy_name"]
units = "code_time"
display_name = "Dinosaurs Density"
```

TODO
-----
- [ ] Add test
- [x] Add doc


matthewturk:
@cphyc Do you want some help with this?

cphyc:
@matthewturk It is essentially just missing a test and some doc, but I don't have much time to work on that at the moment. I'd be happy if you want to push changes directly

matthewturk:
@cphyc I know I offered, but I realized I'm not totally sure I know the right set of docs to add.

chummels:
It seems like maybe these would be good places to note this functionality in the docs:

https://yt-project.org/docs/dev/analyzing/fields.html
https://yt-project.org/docs/dev/reference/configuration.html

Perhaps a fuller description in one place and then just a sentence explanation in the other and a link to the fuller description?

cphyc:
I would approve if I was allowed to!

matthewturk:
@neutrinoceros @cphyc This look OK to you?

yt/fields/field_info_container.py
```diff
@@ -144,10 +145,16 @@ def setup_particle_fields(self, ptype, ftype="gas", num_neighbors=64):
                 raise RuntimeError
             if field[0] not in self.ds.particle_types:
                 continue
+            units = self.ds.field_units.get(field, None)
+            if units is None:
+                try:
+                    units = ytcfg.get("fields", *field, "units")
+                except KeyError:
+                    units = ""
```
neutrinoceros, line 153: I forget. Is this a valid way to express "dimensionless" ? If so, isn't there another, more explicit way ?

yt/fields/field_info_container.py
```diff
@@ -219,8 +226,19 @@ def setup_fluid_aliases(self, ftype="gas"):
...
+            args = known_other_fields.get(field[1], None)
+            if args is not None:
+                units, aliases, display_name = args
+            else:
+                try:
+                    node = ytcfg.get("fields", *field).as_dict()
+                except KeyError:
+                    node = dict()
+
+                units = node.get("units", "")
```
neutrinoceros, line 238: see above comment

yt/fields/field_info_container.py
jzuhone: It's valid

jzuhone:
I just have one question--have we checked to see if there is any significant performance regression?

matthewturk:
@jzuhone There shouldn't be, because it only touches code paths that run at dataset instantiation.

jzuhone (approved):

PR #3347, opened 2021-06-09
load_sample: fix missing data sample registration

### LINKED ISSUE #3341: load_sample: some datasets are missing registration
### Bug report

**Bug summary**

The following datasets currently cannot be loaded with `load_sample` for they lack proper registration:
- ParticleCavity
- cm1_tornado_lofs
- acisf05356N003_evt2.fits.gz
- cm1_supercell_native
- grs-50-cube.fits.gz
- m33_hi.fits.gz
- SmartStars
- tiny_fof_halos

Attempting to load any of the above will fail with
```
ValueError: The requested dataset seems to be improperly registered.
```

What's required to fix these is to add entries for each one in `yt/sample_data_registry.json`. The website also probably should be updated (in particular, the file extensions shouldn't be part of the keys as is currently the case for some). I'm hoping that it won't be necessary to repackage the tarballs (load_sample requires that the key matches the output directory name), but I'll look into that.

neutrinoceros:
## PR Summary

fix #3341

- fix a bug in load_sample where non-tarballs files couldn't be moved to the test-data-dir
- register sample datasets cm1_tornado, SmartStars, tiny_fof_halos
- update broken urls

note: two of the datasets reported in #3341 cannot be trivially fixed for now, namely:
- ParticleCavity. Reason: the tarball is "malformed" according to the rule followed by the rest of the datasamples and implemented in load_sample (update: this one is fixed in #3334 + https://github.com/yt-project/website/pull/103 )
- cm1_supercell_native.tar.gz: I wasn't able to find a way to load this one once untared. No matter what specific file is passed to `yt.load`, it fails with `YTUnidentifiedDataFormat`.

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1435,7 +1435,7 @@ def load_sample(
             fh.extractall(save_dir)
         os.remove(tmp_file)
     else:
-        os.replace(tmp_file, save_dir)
+        os.replace(tmp_file, os.path.join(save_dir, fn))
 
     loadable_path = Path.joinpath(save_dir, fn)
     if load_name not in str(loadable_path):
--- a/yt/sample_data_registry.json
+++ b/yt/sample_data_registry.json
@@ -500,7 +500,7 @@
     "hash": "6fc406dd056b525592f704b202dadf52c41024660dbbd96cfbeb830d710a70d8",
     "load_kwargs": {},
     "load_name": null,
-    "url": "https://yt-project.org/data/acisf05356N003_evt2.fits.gz.tar.gz"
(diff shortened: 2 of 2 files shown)


cphyc (approved):

neutrinoceros:
closed-reopen to rettriger tests. Long term, this will be superseded by #3366, but I think it still has value so I'll reopen. Actually it also has one approval already so maybe it could be considered for inclusion in yt 4.0.2 ?

neutrinoceros:
Let's try to get some attention to this PR for 4.0.3

PR #3348, opened 2021-06-10
make CoordinateHandler an abstract class

neutrinoceros:
## PR Summary

`CoordinateHandler` defines a base class that is clearly not meant to be instantiated as most of its methods go straight to `NotImplementedError`. While it is an easy way to implement base classes, it is not trivial for new comers to guess where the missing code should be injected in case they run into one of the raise statements, since it should really be added to the specialised child class rather than the base class that the traceback leads to.

To make things clearer, I'm making `CoordinateHandler` an actual abstract class.
I will use CI to discover which methods should be added to child classes, and write `NotImplementedError`s _there_.

### DIFF
--- a/yt/geometry/coordinates/coordinate_handler.py
+++ b/yt/geometry/coordinates/coordinate_handler.py
@@ -1,3 +1,4 @@
+import abc
 import weakref
 from numbers import Number
 
@@ -61,7 +62,7 @@ def validate_sequence_width(width, ds, unit=None):
             )
 
 
-class CoordinateHandler:
+class CoordinateHandler(abc.ABC):
     name = None
 
     def __init__(self, ds, ordering):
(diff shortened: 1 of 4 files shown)


matthewturk:
I like this idea, and this is an area of the code that could use some
attention. Thanks!

yt/geometry/coordinates/coordinate_handler.py
```diff
@@ -70,38 +71,46 @@ def __init__(self, ds, ordering):
 
     def setup_fields(self):
         # This should return field definitions for x, y, z, r, theta, phi
-        raise NotImplementedError
+        pass
```
cphyc, line 74: Why isn't this one not an abstract method?

cphyc (approved):

yt/geometry/coordinates/coordinate_handler.py
neutrinoceros: looks like a mistake on my part, thanks for catching it !

chrishavlin (approved):
this is helpful! changes look good to me.

PR #3349, opened 2021-06-10
cleanup Python version checking 

neutrinoceros:
## PR Summary

This cleans up reminiscent compatibility layers for Python 2, but more importantly, it fixes two bugs:
- in yt/startup_tasks.py where a conditional would break when Python 3.10 is released (later this year).
- in `yt/__init__.py`, where the version check would wrongfully pass under Python >=3.0;<3.5 
By adding a flake8 plugin to the pre-commit tool belt, I also prevent this kind of mishap to happen again in the future.

### DIFF
--- a/yt/__init__.py
+++ b/yt/__init__.py
@@ -7,13 +7,6 @@ yt is a toolkit for analyzing and visualizing volumetric data.
 * Contribute: https://github.com/yt-project/yt
 
 """
-import sys
-
-if sys.version_info[0] < 3:
-    raise Exception(
-        "Python 2 is no longer supported.  Please install Python 3 for use with yt."
-    )
-
 __version__ = "4.0.dev0"
 
 import yt.units as units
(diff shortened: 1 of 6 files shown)


yt/__init__.py
```diff
@@ -7,13 +7,6 @@
 * Contribute: https://github.com/yt-project/yt
 
 """
-import sys
-
-if sys.version_info[0] < 3:
-    raise Exception(
-        "Python 2 is no longer supported.  Please install Python 3 for use with yt."
-    )
-
```
neutrinoceros, line 16: I'm removing this check because it is impossible to trigger: yt _cannot be installed_ with Python 2 (or any version <3.6), as the requirement is already specified in `setup.cfg`, so the failure should and will happen at compile time rather than runtime.

matthewturk (approved):
I see no reason not to merge immediately

Issue #3281, opened 2021-05-19
Update all docs to reflect yt4 changes and fix breakages


chummels:
### Bug report

**Bug summary**

I think we need to scan through all of the docs to see where things are broken in yt4.  I've noticed a number of areas where embedded code is broken in the dev docs (e.g., the [notebook embedded in the particle filtering docs](https://yt-project.org/docs/dev/analyzing/filtering.html#filtering-particle-fields)). These corrections should be addressed before we release yt4.  This is in addition to the other docs issues that need to be addressed before yt4 release: Issues #3244 , #3231, #3230, #2629, and #2218.

- [x] Check Introduction to yt docs
- [x] Check yt4 differences docs (Issue #3231)
- [x] Check Installation docs (Issue #3244 -> PR #3350)
- [x] Check Quickstart docs (PR #3306)
- [x] Check Loading and Examining docs (PR #3284)
- [x] Check Cookbook docs (PR #3284)
- [x] Check Visualizing docs (PR #3355)
- [x] Check General Data Analysis docs (PR #3355)
- [x] Check Domain-Specific Analysis docs (PR #3355)
- [x] Check Developer Guide docs (PR #3367)
- [x] Check Sharing Data docs (PR #3355)
- [x] Check Reference Materials docs
- [x] Check FAQ docs (PR #3354)
- [x] Check Help docs (PR #3355)
- [x] Check About yt docs (PR #3355)

We can all work together to check these docs over.  If you encounter an issue in one of the docs, either create a PR to address it, or document it in another Issue before checking off that section in the above list.  We're almost there, everyone!

neutrinoceros:
this looks like a nice complement to the Yt-4.0 meta issue, I'll pin it as well

neutrinoceros:
Just a note that if we want to consider #3244 as a part of this blocker, then #3236 should be discussed as well.

matthewturk:
#3354 checks a box

matthewturk:
#3355 includes my re-reading of all the sections marked as related above.  Notably, I did not do the Reference docs or the Developer docs.

matthewturk:
I've checked off the boxes that were targeted in now-closed PRs

neutrinoceros:
I think we should collectively decide wether #3369 should be added to this list. With the current heavy updates in the documentation, the chances to introduce mistakes (broken refs, invalid code...) seems high. I think we need to keep an eye on the logs to avoid introducing new mistakes while we check all the boxes here :)

[...]

PR #3355, opened 2021-06-15
Doc review for yt-4.0

matthewturk:
This is a draft PR that goes through the docs, attempting to check the remaining boxes on #3281.

I'll mark it as ready when I'm done, but I'm going through in sequence, so comments are fine at any stage.


matthewturk:
ready for review

doc/source/about/index.rst
```diff
@@ -30,36 +30,30 @@ to the project.  For a list of those members and a description of their
...
-contributions from many people, the hospitality of several institutions, and
-benefiting from numerous grants.  With this community-driven approach
-and contributions from a sizeable population of developers, it has evolved
-into a fully-featured toolkit for analysis and visualization of
-multidimensional data.  It relies on no proprietary software -- although it
-can be and has been extended to interface with proprietary software and
-libraries -- and has been designed from the ground up to enable users to be
-as immersed in the data as they desire.
+yt was originally created to study datasets generated by cosmological
+simulations of galaxy and star formation conducted by the simulation code Enzo.
```
neutrinoceros, line 37: Sounds a little off to me but maybe that's how people actually phrase it ?
```suggestion
simulations of galaxies and star formation conducted by the simulation code Enzo.
```

doc/source/cookbook/opengl_ipython.py
```diff
@@ -1,30 +0,0 @@
-import yt
```
neutrinoceros, line 1: This file is referenced explicitly in doc/source/cookbook/complex_plots.rst
I'm not against its removal, but the reference need be cleaned up too :)

doc/source/cookbook/opengl_vr.py
```diff
@@ -1,28 +0,0 @@
-import yt
```
neutrinoceros, line 1: Same as above

doc/source/visualizing/plots.rst
```diff
@@ -295,7 +296,7 @@ example:
    ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
    prj = yt.ProjectionPlot(
        ds,
-       2,
+       "z",
```
neutrinoceros, line 299: out of curiosity, how did you find those ?

neutrinoceros (changes requested):
Good work, I love reviewing when the diff is mostly red :)

You'll need to cleanup some references to deleted files, but other than this nothing major for me !

doc/source/visualizing/plots.rst
matthewturk: I noticed them while reading

doc/source/about/index.rst
matthewturk: I think I'm going to leave this but I'll wait to hear from @chummels

neutrinoceros: Ok 👍

neutrinoceros (approved):

neutrinoceros:
@yt-fido test this please

doc/source/about/index.rst
```diff
@@ -30,36 +30,30 @@ to the project.  For a list of those members and a description of their
...
-collaborators and friends to contribute and use yt.  As the community grew,
-so did the capabilities of yt.  It is now a community-developed project with
-contributions from many people, the hospitality of several institutions, and
-benefiting from numerous grants.  With this community-driven approach
-and contributions from a sizeable population of developers, it has evolved
-into a fully-featured toolkit for analysis and visualization of
-multidimensional data.  It relies on no proprietary software -- although it
-can be and has been extended to interface with proprietary software and
-libraries -- and has been designed from the ground up to enable users to be
-as immersed in the data as they desire.
```
chummels, line 54: I know you're very humble, Matt, but I think it is useful for people to know the whole history of yt started with you.  I like what you did below, but perhaps continuing to mention that it started in 2007 with your work on enzo analysis.

chummels: Oh, now I see the thing below about you as the original author.  OK, that should be good enough.  But please, you need to get credit for having done an ENORMOUS amount on this code.

doc/source/help/index.rst
```diff
@@ -177,25 +177,17 @@ the problem!
...
 your request.
 
-Alternatively you can go to our IRC channel, which does not require an
-invite. You don't even need to have any special IRC client in order to join the
-IRC channel.  We are the #yt channel on irc.freenode.net, but you can also
-connect using your web browser by going to https://yt-project.org/irc.html .
-There are usually 2-8 members of the user base and development team online, so
-you'll probably get your answers quickly.  Remember to bring the information
-from the :ref:`last step <isolate_and_document>`.
-
```
chummels, line 198: Are we doing away with IRC?  I guess that makes sense, since slack is pretty widespread now.

doc/source/visualizing/callbacks.rst
```diff
@@ -22,7 +22,7 @@ of available callbacks.  For example:
 
 .. code-block:: python
 
-   slc = SlicePlot(ds, 0, "density")
+   slc = SlicePlot(ds, "x", "density")
```
chummels, line 25: don't we want to do `("gas", "density")` here too?

chummels (commented):
Looks good so far--thanks for working on this!  I think we should consider switching over all field references from strings to tuples in this PR or another PR before release.  I'm open to *not* doing this, but given that this is what has happened in the cookbook and quickstart then it probably makes sense for consistency.

chummels (approved):

doc/source/about/index.rst
matthewturk: If we can figure out a way to include it as part of a broader story, maybe in governance, so that it's not just About Matt, then I would be okay with it.

doc/source/visualizing/callbacks.rst
matthewturk: Probably! I'll take a crack at this and the rest.

doc/source/help/index.rst
matthewturk: Well ... So nobody is really on the irc channel anymore. Also, freenode kind of imploded in a rather outrageous and horrifying way, and our channel probably doesn't even exist anymore. So I think we might be okay ditching it? That doesn't mean I'm not sad about it, just that I think we should emphasize where we *do* have a presence, y'know?

doc/source/visualizing/callbacks.rst
chummels: Awesome!

doc/source/help/index.rst
chummels: Makes sense.

doc/source/about/index.rst
chummels: I like stories About Matt, especially when they're true like this one. :)

matthewturk:
I've updated `plots.rst` and `callbacks.rst`, which I think are the main source of offenses.

matthewturk:
I'd appreciate review, and if we're set for merging, let's merge this, and then we can inspect the results and potentially do other updates.

Updating `plots.rst` was quite a slog.  I'm ... not totally convinced that having so many examples is productive, rather than confusing.

doc/source/visualizing/callbacks.rst
matthewturk: Resolving!

doc/source/visualizing/plots.rst
```diff
@@ -1304,15 +1323,17 @@ In the following example we create a plot and add a simple annotation.
 
    ds = yt.load("enzo_tiny_cosmology/DD0046/DD0046")
    ad = ds.all_data()
-   plot = yt.ProfilePlot(ad, "density", ["temperature"], weight_field=None)
+   plot = yt.ProfilePlot(ad, ("gas", "density"), [("gas", "temperature")], weight_field=None)
    plot.annotate_text(1e-30, 1e7, "Annotated Text")
    plot.save()
 
-To add annotations to a particular set of fields we need to pass in the list of fields as follows:
+To add annotations to a particular set of fields we need to pass in the list of fields as follows, where ``ftype1`` and ``ftype2`` are the field types (and may be the same):
```
neutrinoceros, line 1330:
```suggestion
To add annotations to a particular set of fields we need to pass in the list of fields as follows,
where ``"ftype1"`` and ``"ftype2"`` are the field types (and may be the same):
```

doc/source/visualizing/plots.rst
```diff
@@ -1711,10 +1732,10 @@ of ``particle_position_x`` versus ``particle_velocity_z``, with the ``particle_m
    import yt
 
    ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
-   p = yt.ParticlePlot(ds, "particle_position_x", "particle_velocity_z", ["particle_mass"])
-   p.set_unit("particle_position_x", "Mpc")
-   p.set_unit("particle_velocity_z", "km/s")
-   p.set_unit("particle_mass", "Msun")
+   p = yt.ParticlePlot(ds, ("all", "particle_position_x"), ("all", "particle_velocity_z"), [("all", "particle_mass")])
```
neutrinoceros, line 1735: For sake of simplicity + consistency with the other example (line 1748 as of this writing)
```suggestion
   p = yt.ParticlePlot(ds, ("all", "particle_position_x"), ("all", "particle_velocity_z"), ("all", "particle_mass"))
```

matthewturk: OK, so, this changes what was in there before ... I suppose they do the same thing, right?

neutrinoceros: This is what I'm assuming too

jzuhone (approved):

PR #3363, opened 2021-06-16
Drop yt hub command

Xarthisius:
## PR Summary

This removes outdated and unused `yt hub` command. It might be potentially re-added later, but it shouldn't block 4.0 work.


Xarthisius:
pre-commit.ci autofix

neutrinoceros (commented):
Nice ! should we also remove default values for hub-related keys from yt/config.py ?

Xarthisius:
> Nice ! should we also remove default values for hub-related keys from yt/config.py ?

Dropped in https://github.com/yt-project/yt/pull/3363/commits/e01658a465d6b6a69076b2517079ecb277c70e52

neutrinoceros (approved):

jzuhone (approved):

PR #3365, opened 2021-06-18
Fill out more yt 4 differences

### LINKED ISSUE #3231: Filling out the yt4differences docs page
**Summary**

Right now the [yt4 differences doc page](https://yt-project.org/docs/dev/yt4differences.html) is pretty good, but I think it needs to be fleshed out to really hold the hands of users as they make the transition from yt3 to yt4.  There's nothing so frustrating as to move to a new version of a code, find all of your scripts break, and not know how to fix them.

I've already described this in a comment on Issue #3125 , but it seemed worth having this in a standalone trackable issue.

**Details**

In particular, the [yt4 differences page](https://yt-project.org/docs/dev/yt4differences.html) needs 

- a high level description of the changes between yt3 and yt4, including major changes under the hood as well as API changes.

- Documentation on how to convert old scripts to work with yt4.  Maybe we should describe a bullet point list as a sort of algorithm that a user can follow to convert their old scripts, like what was done in the [yt3 differences page](https://yt-project.org/docs/dev/yt3differences.html). So something like: "Search your scripts and replace any field name strings as field name tuples" with an example.

In particular, it might be worthwhile to sit down with a friend or colleague who is used to working with the stable version of yt, and take them through the process of the upgrade. Then have them use their scripts, and see what fails. On the things that fail, were the error messages obvious as to what needed to change and how to change it? Did we miss it in our description of "what changed" in yt4 and do we need to include something else in our list of items that users need to account for in converting yt3 scripts to yt4?

Anyway, these are all just ideas from someone who remembers how frustrating it can be to upgrade to a new version and have all of your old scripts break. We cannot make it automated, but we can help a lot in easing this process for the user.

matthewturk:
I have gone through all the yt4 diffs that are there, and I think this should finish it up.  I think with this we can close #3231.

### DIFF
--- a/doc/source/yt4differences.rst
+++ b/doc/source/yt4differences.rst
@@ -401,7 +401,7 @@ information.
 New External Packages
 ^^^^^^^^^^^^^^^^^^^^^
 
-As noted above (:ref:`yt-units-is-now-unyt`), ``unyt`` has been extracted from
-yt, and we now use it as an external package.  In addition, other parts of yt
+As noted above (:ref:`yt-units-is-now-unyt`), unyt has been extracted from
+yt, and we now use it as an external library.  In addition, other parts of yt
 such as :ref:`interactive_data_visualization` have been extracted, and we are
-working toward a more modular approach for things such as Jupyter widgets and other "value-added" integrations.gq
+working toward a more modular approach for things such as Jupyter widgets and other "value-added" integrations.
--- a/doc/source/yt4differences.rst
+++ b/doc/source/yt4differences.rst
@@ -363,6 +363,16 @@ You can also use alternate unit names in more complex algebraic unit expressions
 In this example the common british spelling ``"kilometre"`` is resolved to
 ``"km"`` and ``"hour"`` is resolved to ``"hr"``.
 
+Field-Specific Configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can now set configuration values on a per-field basis.  For instance, this
+means that if you always want a particular colormap associated with a particular
+field, you can do so!
+
+This is documented under :ref:`per-field-config`, and was added in `PR
+1931<https://github.com/yt-project/yt/pull/1931>`_.
+
 New Method for Accessing Sample Datasets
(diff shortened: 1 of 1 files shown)


doc/source/yt4differences.rst
```diff
@@ -372,5 +384,24 @@ and load them into a yt session, but now this occurs from within a python
...
+displayed interactively as well.
+
+The package `widgyts <https://widgyts.readthedocs.io>`_ provides interactive,
+yt-specific visualization of slices, projections, and additional dataset display
+information.
+
+New External Packages
+^^^^^^^^^^^^^^^^^^^^^
+
+As noted above (:ref:`yt-units-is-now-unyt`), ``unyt`` has been extracted from
```
neutrinoceros, line 404: For consistency (since we don't usually bartick yt itself)
```suggestion
As noted above (:ref:`yt-units-is-now-unyt`), unyt has been extracted from
```

doc/source/yt4differences.rst
```diff
@@ -372,5 +384,24 @@ and load them into a yt session, but now this occurs from within a python
...
+yt, and we now use it as an external package.  In addition, other parts of yt
```
neutrinoceros, line 405: Maybe it's just me, but I dislike the term "package" more and more in the Python ecosystem because I find it's  being used in lieu of a more descriptive term more often than not
```suggestion
yt, and we now use it as an external library.  In addition, other parts of yt
```

doc/source/yt4differences.rst
```diff
@@ -372,5 +384,24 @@ and load them into a yt session, but now this occurs from within a python
...
+such as :ref:`interactive_data_visualization` have been extracted, and we are
+working toward a more modular approach for things such as Jupyter widgets and other "value-added" integrations.gq
```
neutrinoceros, line 407: typo ?
```suggestion
working toward a more modular approach for things such as Jupyter widgets and other "value-added" integrations.
```

neutrinoceros (approved):
looks good !

doc/source/yt4differences.rst
matthewturk: heh `gq` for formatting in vim

jzuhone (approved):

PR #3366, opened 2021-06-18
Use yt.hub for load_sample

Xarthisius:
## PR Summary

Switch from https://yt-project.org/data to https://hub.yt/


neutrinoceros:
I expect this wil likely break yt.tests.test_load_sample.py, but the fix should be trivial.
When we get CI green, I'll run a much more thorough test to systematically test every dataset and see wether this introduces any breakage !

neutrinoceros:
So there's something I'm not getting atm @Xarthisius:
since you've added an endpoint (https://girder.hub.yt/api/v1/ythub/5e178e0868085e00016c2f73/registry) for the pooch registry, I don't think we need to keep a hard copy of in the main repo ?

Another note for future us: ideally we should not have to use pandas to built the finale "unified_table" (ret value from yt.sample_data.api.get_data_registry_table), since it can be the result of another endpoint ! We don't have to do this now, but it's been an end goal for me since I've started working on load_sample and I can see now that we're getting really close thanks to your work !

neutrinoceros:
Ok so I think this is ready. To mitigate conflicts I think it would be best to merge #3347 first (and migrate the corresponding updates in the registry to your new endpoint), and then I'll be able to get the most accurate comparison between state-of the art and this PR :)

neutrinoceros:
@Xarthisius I took the liberty to rebase this branch to resolve two minor conflicts.

neutrinoceros:
@Xarthisius I'd love this to move forward, but I think we need to discuss how to manage the rest of my load_sample PRs. I expect the right thing to do would be to drop them and integrate some of their changes on the server but I think we need to discuss how we could further iterate when new issues appear.
Any chance you'd have some time to discuss this with me over slack or zoom sometime during the next 2 weeks ?

neutrinoceros:
Quick note: this change probably enables dropping pandas as a dependency for load_sample, which should also simply tests

neutrinoceros:
Can you explain why you closed this ?

Xarthisius:
> Can you explain why you closed this ?

Cause I no longer want to work on it. Feel free to grab whatever you need and open your PR.

neutrinoceros:
Fair enough, thanks for the clarification !

PR #3368, opened 2021-06-19
doc: fix some bugs/warnings in documentation build

neutrinoceros:
## PR Summary

This PR fixes about 40 various warnings (errors, really) in the docs build.
I also fixed a couple deprecations warnings from matplotlib while I was at it.
At the time of writing there are 37 warnings left to solve, which fall into 3 or 4 distinct categories, but I can't find a solution for any of these at the moment, so I'll open a separate issue for them and consider this PR complete (#3369)

PR #3370, opened 2021-06-21
Update reference docs for 4.0

jzuhone:
## PR Summary

This PR updates the reference docs for yt 4.0, and fixed a bug with the `yt stats` command that I missed in a previous PR. 

I need some help with the "code support"--there are a few codes that I don't know about the level of support.


doc/source/reference/code_support.rst
```diff
@@ -12,66 +12,78 @@ as inputs.  Below is a list showing what level of support is provided for
...
++-----------------------+------------+-----------+------------+-------+----------+----------+------------+-------------+
+| WarpX                 |     Y      |     Y     |      Y     |   Y   |    Y     |    Y     |     Y      |   Full      |
++-----------------------+------------+-----------+------------+-------+----------+----------+------------+-------------+
 
 .. [#f1] one-dimensional base-state not read in currently.
 .. [#f2] These handle mesh fields using an in-memory octree that has not been parallelized.
          Datasets larger than approximately 1024^3 will not scale well.
 .. [#f3] Newer versions of Castro that use BoxLib's standard particle format are supported.
-	 The older ASCII format is not.
+	     The older ASCII format is not.
```
neutrinoceros, line 84: I think it has to match the indentation level of the text above. Tabs should be avoided
```suggestion
          The older ASCII format is not.
```

doc/source/reference/python_introduction.rst
```diff
@@ -643,7 +637,7 @@ covers mechanisms for combining arrays of different shapes and sizes, which can
 be tricky but also extremely powerful.  We won't discuss the idea of
 broadcasting here, simply because I don't know that I could do it justice!  The
 NumPy Docs have a great `section on broadcasting
-<https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html>`_.
+<https://numpy.org/doc/stable/user/basics.broadcasting.html>`_.
```
neutrinoceros, line 640: good catch ! While you're at it, can you update the remaining link to docs.scipy...numpy ? It's in doc/source/conf.py :)

doc/source/reference/command-line.rst
```diff
@@ -131,17 +131,6 @@ help
...
-
-bugreport
-+++++++++
-
-Encountering a bug in your own code can be a big hassle, but it can be
-exponentially worse to find it in someone else's.  That's why we tried to
-make it as easy as possible for users to report bugs they find in yt.
-After you go through the necessary channels to make sure you're not just
-making a mistake (see :ref:`asking-for-help`), you can submit bug
-reports using this nice utility.
```
neutrinoceros, line 143: For reference, I support making the docs less repetitive and verbose.

doc/source/reference/command-line.rst
```diff
@@ -208,29 +197,50 @@ command-line.
 rpdb
 ++++
 
-Connect to a currently running (on localhost) rpd session.
+Connect to a currently running (on localhost) rpd session. See
```
neutrinoceros, line 200:
```suggestion
Connect to a currently running (on localhost) rpdb session. See
```
Not to be confused with https://pypi.org/project/pdbr/ !

doc/source/reference/command-line.rst
```diff
@@ -208,29 +197,50 @@ command-line.
...
 
 stats
 +++++
 
 This subcommand provides you with some basic statistics on a given dataset.
 It provides you with the number of grids and cells in each level, the time
-of the dataset, the resolution, and the maximum density in a variety of units.
-It is tantamount to performing the ``print_stats()`` inside of yt.
+of the dataset, and the resolution. It is tantamount to performing the
+``print_stats()`` inside of yt.
```
neutrinoceros, line 217:
```suggestion
of the dataset, and the resolution. It is tantamount to calling the
``Dataset.print_stats`` method.
```

doc/source/reference/command-line.rst
```diff
@@ -208,29 +197,50 @@ command-line.
...
+
+.. code-block:: bash
+
+   yt stats GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150 --max --min
+
+or a different field can be specified using the ``-f`` flag:
+
+.. code-block:: bash
+
+   yt stats GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150 --max --min -f temperature
```
neutrinoceros, line 230: just curious, does this comment support ftype, fname tuples ?

doc/source/reference/command-line.rst
```diff
@@ -208,29 +197,50 @@ command-line.
...
+
+The field-related stats output from this command can be directed to a file using
+the ``-o`` flag:
+
+.. code-block:: bash
+
+   yt stats GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150 --max -o out_stats.dat
```
neutrinoceros, line 237: Does it do anything different than this ?
```
yt stats GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150 --max > out_stats.dat
```

doc/source/reference/configuration.rst
```diff
@@ -203,7 +203,7 @@ by defining ``plugin_filename`` in your ``yt.toml`` file, as mentioned above.
 .. note::
 
    You can tell that your system plugin file is being parsed by watching for a logging
-   message when you import yt.  Note that both the ``yt load`` and ``iyt``
+   message when you import yt. Note that both the ``yt load`` and ``iyt``
    command line entry points parse the plugin file, so the ``my_plugins.py``
    file will be parsed if you enter yt that way.
```
neutrinoceros, line 208: This sentence seems repetitive, maybe we can simply remove it ?
```suggestion
   command line entry points parse the plugin file.
```

doc/source/reference/index.rst
```diff
@@ -15,5 +15,6 @@ code.
    api/api
    configuration
    python_introduction
+   demeshening
```
neutrinoceros, line 18: I'm also fixing this in #3368. It's likely that fixing it twice will result in a duplicate entry instead of a single one, can you drop it here ?

doc/source/reference/python_introduction.rst
```diff
@@ -50,22 +50,16 @@ signs.  Let's inaugurate the occasion appropriately -- type this::
 
 As you can see, this printed out the string "Hello, world." just as we
 expected.  Now let's try a more advanced string, one with a number in it.  For
-this we'll use the percent (``%``) operator, which is the manner by which
-values are fed into a formatted string.  We'll print pi, but only with three
-digits of accuracy.::
+this we'll use an "f-string", which is the manner by which values are fed into
+a formatted string.  We'll print pi, but only with three digits of accuracy.::
```
neutrinoceros, line 54: Using "the" when there are at least 4 or 5 different ways to do this in Python seems exaggerated.
```suggestion
this we'll use an "f-string", which is the preferred way to format strings in modern Python.
We'll print pi, but only with three digits of accuracy.::
```

doc/source/reference/python_introduction.rst
```diff
@@ -128,8 +122,8 @@ This works for floating points as well.  Now we can do math on these numbers::
    >>> print(a + 5.1)
    >>> print(a / 2.0)
 
-Because of a historical aversion to floating point division in Python (which is
-now changing) it's always safest to ensure that either the numerator or the
+Because of a historical aversion to floating point division in Python (which has
+now changed) it's always safest to ensure that either the numerator or the
 denominator is a floating point number.
```
neutrinoceros, line 127: I think this remark is completely unnecessary in Python 3, so we can drop it :)
```suggestion
```

yt/utilities/command_line.py
```diff
@@ -1174,9 +1174,9 @@ def __call__(self, args):
             vals["min"] = ds.find_min(field)
             print(f"Minimum {field.name}: {vals['min'][0]:0.5e} at {vals['min'][1]}")
         if args.output is not None:
-            t = ds.current_time * ds["years"]
+            t = ds.current_time.to("yr")
```
neutrinoceros, line 1177: how did we not catch this earlier ??
anyway, nice job !

neutrinoceros (commented):
Good work here. Here are a couple comments and suggestions !

doc/source/reference/code_support.rst
```diff
@@ -12,66 +12,78 @@ as inputs.  Below is a list showing what level of support is provided for
...
+| GAMER                 |     Y      |     Y     |      Y     |   Y   |    Y     |    Y     |     Y      |   Full      |
++-----------------------+------------+-----------+------------+-------+----------+----------+------------+-------------+
+| Gasoline              |     Y      |     Y     |      Y     |   Y   |    Y     |    Y     |     Y      |   Full      |
++-----------------------+------------+-----------+------------+-------+----------+----------+------------+-------------+
+| Gizmo                 |     Y      |     Y     |      Y     |   Y   |    Y     |    Y     |     Y      |   Full      |
++-----------------------+------------+-----------+------------+-------+----------+----------+------------+-------------+
+| Grid Data Format (GDF)|     Y      |    N/A    |      Y     |   Y   |    Y     |    Y     |     Y      |   Full      |
++-----------------------+------------+-----------+------------+-------+----------+----------+------------+-------------+
+| IAMR                  |     ?      |     ?     |      ?     |   ?   |    ?     |    ?     |     ?      | ?           |
```
neutrinoceros, line 55: Just highlighting that this is where help is needed. I don't personally know how to complete this line either.

doc/source/reference/command-line.rst
jzuhone: Yes, if you specify them like this:
```bash
yt stats GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150 --max --min -f gas,temperature
```

doc/source/reference/command-line.rst
jzuhone: Yes--it outputs only the max and min stats information. 

In my opinion this option deserves a refactor, but out of scope for this PR.

doc/source/reference/index.rst
jzuhone: Sure

doc/source/reference/python_introduction.rst
jzuhone: Yep

doc/source/reference/command-line.rst
neutrinoceros: ok cool. In this case I'd argue that it's probably best to use it here, so we don't create a bug if we ever decide to drop support for single key field (not sure we should, but it's becoming a possibly lately).

doc/source/reference/command-line.rst
neutrinoceros: Yeah, that's totally fine.

doc/source/reference/command-line.rst
jzuhone: done.

neutrinoceros (approved):
thanks for quickly taking my suggestions in consideration. I know this is unfinished since it's missing some information in the support table, but I approve of everything else !

matthewturk (approved):
I think this is excellent.  Thank you, @jzuhone .

neutrinoceros:
Should we just merge now to avoid building up conflicts while we work on other docs PRs ? I think it's fine if we keep track of the missing line of the support table in an issue.

matthewturk:
I think we should

neutrinoceros:
Ok then it's your call John !

PR #3373, opened 2021-06-23
CI: bump answers for numpy 1.21

neutrinoceros:
## PR Summary

With the release of numpy 1.21, the following answer tests finale images changed slightly
yt/visualization/tests/test_plotwindow.py:test_attributes_wt

here's the existing reference
tmpokwzf291
and here's the new result
tmp7j8gqehr

Here's a minimal script to reproduce this (thanks @Xarthisius for posting this on Slack)
```python
import yt
from yt.utilities.answer_testing.framework import data_dir_load
WT = "WindTunnel/windtunnel_4lev_hdf5_plt_cnt_0030"
plot_field = ("gas", "density")
ds = data_dir_load(WT)
ax = "z" 
slc = yt.SlicePlot(ds, ax, plot_field)
slc.set_zlim(plot_field, 1e-25, 1e-23)
slc.save()
```

The actual range in the data is [0.05956075, 6.61460495] (g/cm^3), so the resulting image is kinda irrelevant in both cases and the new one is actually consistent with what can be obtained with a very simple, pure mpl + numpy example script:

```python
from matplotlib.colors import LogNorm
import numpy as np
data = np.ones((10, 10))
fig, ax = plt.subplots()
im = ax.imshow(data, norm=LogNorm(vmin=1e-25, vmax=1e-23))
ax.set(title=f"numpy v{np.__version__}")
fig.colorbar(im)
fig.savefig(f"/tmp/v{np.__version__}.png")
```
v1 21 0

My conclusion is that there is nothing to fix on our side so we can simply bump our answers and call it a day.

update: as discussed on Slack with @Xarthisius and @jzuhone , I've updated the PR to simply drop the problematic test.

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -94,7 +94,6 @@ answer_tests:
 
   local_pw_037:  # PR 3373
     - yt/visualization/tests/test_plotwindow.py:test_attributes
-    - yt/visualization/tests/test_plotwindow.py:test_attributes_wt
     - yt/visualization/tests/test_particle_plot.py:test_particle_projection_answers
     - yt/visualization/tests/test_particle_plot.py:test_particle_projection_filter
     - yt/visualization/tests/test_particle_plot.py:test_particle_phase_answers
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -92,7 +92,7 @@ answer_tests:
     - yt/frontends/owls/tests/test_outputs.py:test_snapshot_033
     - yt/frontends/owls/tests/test_outputs.py:test_OWLS_particlefilter
 
-  local_pw_036:  # PR 3272
+  local_pw_037:  # PR 3373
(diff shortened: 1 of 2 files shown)


tests/tests.yaml
```diff
@@ -92,7 +92,7 @@ answer_tests:
     - yt/frontends/owls/tests/test_outputs.py:test_snapshot_033
     - yt/frontends/owls/tests/test_outputs.py:test_OWLS_particlefilter
 
-  local_pw_036:  # PR 3272
+  local_pw_037:  # PR 3272
```
neutrinoceros, line 95:
```suggestion
  local_pw_037:  # PR 3373
```

jzuhone:
LGTM

jzuhone (approved):

jzuhone (approved):

Xarthisius (approved):

PR #3374, opened 2021-06-23
[bugfix] generate_smoothing_length expects 64-bit positions

jzuhone:
## PR Summary

In PR #2645 we added the capability to generate smoothing lengths for a Gadget HDF5 dataset that didn't have one. It turns out the example I used had 64-bit particle positions (which is not typical, Gadget datasets usually have 32-bit positions), and the `generate_smoothing_length` function expects 64-bit positions. 

There may be a better way of handling this (with a fused type or something) but this is a quick fix. Note that we have to cast the `positions` array upon entry into the `generate_smoothing_length` function so that the original array retains its original type.


neutrinoceros (approved):
LGTM.

neutrinoceros:
@yt-fido test this please

PR #3375, opened 2021-06-23
[bugfix] Fix FITSParticleProjection for all axes

jzuhone:
## PR Summary

This PR fixes the recently added `FITSParticleProjection` class. The bounds for the FITS image were incorrectly being calculated--they assumed that the appropriate values for the center of the image were those in the x and y axes, which is only true for a projection along the z-axis. This fix ensures that the correct values of the center coordinate are chosen depending on the axis of the projection.

### DIFF
--- a/yt/visualization/fits_image.py
+++ b/yt/visualization/fits_image.py
@@ -888,11 +888,12 @@ def construct_image(ds, axis, data_source, center, image_res, width, length_unit
         else:
             frb = data_source.to_frb(width[0], (nx, ny), center=center, height=width[1])
     elif isinstance(data_source, ParticleAxisAlignedDummyDataSource):
+        axes = axis_wcs[axis]
         bounds = (
-            center[0] - width[0] / 2,
-            center[0] + width[0] / 2,
-            center[1] - width[1] / 2,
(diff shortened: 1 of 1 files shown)


yt/visualization/fits_image.py
```diff
@@ -888,11 +888,12 @@ def construct_image(ds, axis, data_source, center, image_res, width, length_unit
         else:
             frb = data_source.to_frb(width[0], (nx, ny), center=center, height=width[1])
     elif isinstance(data_source, ParticleAxisAlignedDummyDataSource):
+        axes = axis_wcs[axis]
```
neutrinoceros, line 891: seems like the right fix, although I can imagine it wouldn't be so obvious to me if I read this in the code first without the additional context you provided in the report. Can you add a short description for `axis_wcs` close to its definition while you're at it please ?

jzuhone: done

neutrinoceros: thanks !

neutrinoceros (approved):

neutrinoceros:
@yt-fido test this please

PR #3376, opened 2021-06-23
Future proofing: add a scheduled job with dev versions of core deps

neutrinoceros:
## PR Summary

This adds a workflow to run essential tests against Python 3.10, with dev versions of numpy and matplotlib. The goal is to buy ourselves more time to fix issues (or reporting upstream) before it's too late.

The job is setup to be triggered at least once a day, as well as on every push to the main branch.
GH actions currently do not support a way to signal this properly ([yet](https://github.com/actions/toolkit/issues/399)),
so I won't make it trigger on PRs

This is _heavily_ inspired from a similar workflow in the pandas repo.

While I'm at it I'm updating the triggers for regular CI workflows so we can save a few hours of CPU time by skipping tests on PRs that only touch documentation.

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -1,0 +1,54 @@
+name: CI (bleeding edge)
+# this workflow is heavily inspired from pandas, see
+# https://github.com/pandas-dev/pandas/blob/master/.github/workflows/python-dev.yml
+
+# goals: check stability against
+# - dev version of Python, numpy, and matplotlib
+# - building with future pip default options
+
+on:
+  push:
+    branches:
+      - main
+  schedule:
+    # run this every day at 3 am UTC
+    - cron: '0 3 * * *'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Python3.10-dev
+    timeout-minutes: 60
+
+    concurrency:
+      group: ${{ github.ref }}-dev
+      cancel-in-progress: true
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python Dev Version
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.10-dev'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install --upgrade setuptools wheel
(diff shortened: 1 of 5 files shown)


neutrinoceros:
Ok so I got the workflow to run. It's failing because of a newly revealed api breakage from matplotlib, which is the whole purpose of this first line of defense. To avoid polluting everyone's PRs with `:x:`, I changed the triggers: this workflow will not trigger on PRs but will run on the main branch upon merge, and 4 times a day even when nothing else happens.
The status of the workflow will be displayed in the README for convenience as
https://github.com/yt-project/yt/actions/workflows/bleeding-edge.yaml/badge.svg

.github/workflows/bleeding-edge.yaml
```diff
@@ -0,0 +1,54 @@
...
+# - dev version of Python, numpy, and matplotlib
+# - building with future pip default options
+
+on:
+  push:
+    branches:
+      - main
+  schedule:
+    # run this every 6 hours (time is UTC)
+    - cron: '0 3,9,15,21 * * *'
```
Xarthisius, line 15: Let's think about rainforests and our grandchildren. Once a day is more than plenty...

neutrinoceros: You're right. I got excited, but I'll fix this

neutrinoceros:
For reference I'm working on the breaking change that was checked-in in matplotlib recently. I'll probably end up reporting upstream and hopefully it can be fixed before matplotlib 3.5 is actually released.

neutrinoceros:
marking this as draft for now because it removes the hard pinning to forbid matplotlib 3.5, and this doesn't look like a good idea at the moment.

neutrinoceros:
reported upstream https://github.com/matplotlib/matplotlib/issues/20520

neutrinoceros:
now that the release is out, I believe this is safe to merge as is.

neutrinoceros:
The problem with matplotlib has allegedly been fixed upstream 🎉 
I'm going to rerun this job here to check that, and see how we're doing now. Moving to draft in the mean time.

neutrinoceros:
Great news, the job passed (https://github.com/yt-project/yt/actions/runs/1010839948)

cphyc (approved):
LGTM, that seems like a nice addition :)

matthewturk (approved):
I think this would be a good addition.

PR #3378, opened 2021-06-25
Future proofing: forbid cython 3.0 until we can properly test against it

neutrinoceros:
## PR Summary

Cython 3.0 is coming soon and already has some release-candidates available.
It seems very likely that yt will _not_ be compatible with it out of the box, but I don't think anyone has time to future proof this right now, so I think it's much preferable to forbid it completely for now. For reference, this is inline with the current specs in numpy itself, see
[numpy's pyproject.toml](https://github.com/numpy/numpy/blob/57863f598ddaade14b3d81e6e91ebecca881f996/pyproject.toml#L7)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,10 @@ requires = [
 
   # cython version is imposed by that of numpy, see release notes
   # https://github.com/numpy/numpy/releases/tag/v1.19.2
-  "Cython>=0.26.1; python_version=='3.6'",
-  "Cython>=0.29.21; python_version>='3.7'",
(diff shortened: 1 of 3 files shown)


matthewturk:
I think we need to stay on top of this when it does come out, but for 4.0 it makes sense to me to "forbid" it.

matthewturk (approved):

neutrinoceros:
> I think we need to stay on top of this when it does come out, but for 4.0 it makes sense to me to "forbid" it.

Agreed. I'll build on top of PR #3376 to include this !

PR #3380, opened 2021-06-26
Future proofing: treat most warnings as errors in pytest, and fix existing ones

### LINKED ISSUE #3379: Future proofing yt 4.0
### Bug report

**Bug summary**
Our CI is reporting some deprecation warnings from upstream dependencies (mostly numpy and matplotlib).
I think it's worth fixing them ahead of our next release, so yt 4.0 doesn't instantly breaks next time numpy issues a version.

I'm working on automating the detection of such warnings in #3376, which already helped me in reporting a (future) regression upstream to matplotlib, and I except I'll be able to fix all existing warnings in a short amount of time, but right now I'm struggling with setting up the right pytest configuration.

neutrinoceros:
## PR Summary

fix #3379
Though independent, this is complementary with #3376.

An important note is that there are some internal deprecation warnings with no obvious solution yet, so I'm adding them to the ignore-list in conftest.py but they will need to be addressed in future releases. Specifically, I'm talking about fields `("all", "px")` and `("gas", "px")` (respectively `"py"`, `"dpx"` and `"dpy"`), that are used internally in slice selectors, but never with explicit field/particle types. Getting to a stable solution here requires much more thought and design that what I'm solving here so I'll defer it to future works, and in any case it's not as urgent as other deprecation warnings because it's internal so it won't break haphazardly.

## TODO
- [x] pass CI
- [x] document _why_ each ignored warning gets a pass (comments)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -76,6 +76,7 @@ def pytest_configure(config):
         # treat most warnings as errors
         "error",
         # >>> internal deprecation warnings with no obvious solution
+        # see https://github.com/yt-project/yt/issues/3381
         (
             r"ignore:The requested field name 'pd?[xyz]' is ambiguous and corresponds "
             "to any one of the following field types.*:yt._maintenance.deprecation.VisibleDeprecationWarning"
--- a/conftest.py
+++ b/conftest.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import tempfile
+from importlib.util import find_spec
 from pathlib import Path
 
 import pytest
@@ -71,6 +72,90 @@ def pytest_configure(config):
     config.addinivalue_line(
         "markers", "big_data: Run answer tests that require large data files."
     )
+    for value in (
+        # treat most warnings as errors
+        "error",
+        # >>> internal deprecation warnings with no obvious solution
+        (
+            r"ignore:The requested field name 'pd?[xyz]' is ambiguous and corresponds "
+            "to any one of the following field types.*:yt._maintenance.deprecation.VisibleDeprecationWarning"
+        ),
+        # >>> warnings emitted by testing frameworks, or in testing contexts
+        # we still have some yield-based tests, awaiting for transition into pytest
+        "ignore::pytest.PytestCollectionWarning",
+        # imp is used in nosetest
+        "ignore:the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses:DeprecationWarning",
+        # the deprecation warning message for imp changed in Python 3.10, so we ignore both versions
+        "ignore:the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses:DeprecationWarning",
+        # matplotlib warnings related to the Agg backend which is used in CI, not much we can do about it
+        "ignore:Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.:UserWarning",
+        "ignore:tight_layout . falling back to Agg renderer:UserWarning",
+        #
+        # >>> warnings from wrong values passed to numpy
+        # these should normally be curated out of the test suite but they are too numerous
+        # to deal with in a reasonable time at the moment.
+        "ignore:invalid value encountered in log10:RuntimeWarning",
+        "ignore:divide by zero encountered in log10:RuntimeWarning",
+        "ignore:invalid value encountered in true_divide:RuntimeWarning",
+        #
+        # >>> there are many places in yt (most notably at the frontend level)
+        # where we open files but never explicitly close them
+        # Although this is in general bad practice, it can be intentional and
+        # justified in contexts where reading speeds should be optimized.
+        # It is not clear at the time of writing how to approach this,
+        # so I'm going to ignore this class of warnings altogether for now.
+        "ignore:unclosed file.*:ResourceWarning",
+    ):
+        config.addinivalue_line("filterwarnings", value)
+
+    # at the time of writing, astropy's wheels are behind numpy's latest
+    # version but this doesn't cause actual problems in our test suite, so
+    # we allow this warning to pass.
+    # last checked with astropy 4.2.1
+    config.addinivalue_line(
+        "filterwarnings",
+        (
+            "ignore:numpy.ndarray size changed, may indicate binary incompatibility. "
+            "Expected 80 from C header, got 88 from PyObject:RuntimeWarning"
+        ),
+    )
+    if find_spec("astropy") is not None:
+        # astropy triggers this warning from itself, there's not much we can do on our side
+        # last checked with astropy 4.2.1
+        config.addinivalue_line(
+            "filterwarnings", "ignore::astropy.wcs.wcs.FITSFixedWarning"
+        )
+
+    if find_spec("cartopy") is not None:
+        # cartopy still triggers this numpy warning
+        # last checked wtih cartopy 0.19.0
+        config.addinivalue_line(
+            "filterwarnings",
+            (
+                "ignore:`np.float` is a deprecated alias for the builtin `float`. "
+                "To silence this warning, use `float` by itself. "
+                "Doing this will not modify any behavior and is safe. "
+                "If you specifically wanted the numpy scalar type, use `np.float64` here."
+                ":DeprecationWarning: "
+            ),
+        )
+        # this warning *still* shows up on cartopy 0.19 so we'll ignore it
+        config.addinivalue_line(
+            "filterwarnings",
+            (
+                r"ignore:The default value for the \*approx\* keyword argument to "
+                r"\w+ will change from True to False after 0\.18\.:UserWarning"
+            ),
+        )
+        # this one could be resolved by upgrading PROJ on Jenkins,
+        # but there's isn't much else that can be done about it.
+        config.addinivalue_line(
+            "filterwarnings",
+            (
+                "ignore:The Stereographic projection in Proj older than 5.0.0 incorrectly "
+                "transforms points when central_latitude=0. Use this projection with caution.:UserWarning"
+            ),
+        )
 
 
 def pytest_collection_modifyitems(config, items):
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ install_requires =
     toml>=0.10.2
     tqdm>=3.4.0
     unyt>=2.8.0
-python_requires = >=3.6
+python_requires = >=3.6,<3.12
 include_package_data = True
 scripts = scripts/iyt
 zip_safe = False
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,11 +70,11 @@ doc =
     sphinx-bootstrap-theme
     sphinx-rtd-theme
 full =
-    astropy~=4.0.1
+    astropy>=4.0.1,<5.0.0
     f90nml>=1.1.2
     fastcache~=1.0.2
     glueviz~=0.13.3
-    h5py~=3.1.0
+    h5py>=3.1.0,<4.0.0
     libconf~=1.0.1
     miniballcpp>=0.2.1
     mpi4py~=3.0.3
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ project_urls =
 packages = find:
 install_requires =
     iPython>=1.0
-    matplotlib!=3.4.2,>=2.0.2,<3.5
+    matplotlib!=3.4.2,>=2.0.2,<3.6
     more-itertools>=8.4
     numpy>=1.13.3
     pyyaml>=4.2b1
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -8,6 +8,7 @@ from re import finditer
 from tempfile import NamedTemporaryFile, TemporaryFile
 
 import numpy as np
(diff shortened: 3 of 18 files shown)


conftest.py
```diff
@@ -71,6 +72,74 @@ def pytest_configure(config):
...
+        # where we open files but never explicitly close them
+        # Although this is in general by practice, it can be intentional and
+        # justified in contexts where reading speeds should be optimized.
+        # It is not clear at the time of writing how to approach this,
+        # so I'm going to ignore this class of warnings altogether for now.
+        "ignore:unclosed file.*:ResourceWarning",
+    ):
+        config.addinivalue_line("filterwarnings", value)
+
+    # at the time of writing, astropy's wheel are behind numpy's latest
```
neutrinoceros, line 112:
```suggestion
    # at the time of writing, astropy's wheels are behind numpy's latest
```

conftest.py
```diff
@@ -71,6 +72,74 @@ def pytest_configure(config):
...
+        # >>> warnings from wrong values passed to numpy
+        # these should normally be curated out of the test suite but they are too numerous
+        # to deal with in a reasonable time at the moment.
+        "ignore:invalid value encountered in log10:RuntimeWarning",
+        "ignore:divide by zero encountered in log10:RuntimeWarning",
+        "ignore:invalid value encountered in true_divide:RuntimeWarning",
+        #
+        # >>> there are many places in yt (most notably at the frontend level)
+        # where we open files but never explicitly close them
+        # Although this is in general by practice, it can be intentional and
```
neutrinoceros, line 104:
```suggestion
        # Although this is in general bad practice, it can be intentional and
```

yt/visualization/plot_window.py
```diff
@@ -48,6 +48,21 @@
 )
 from .plot_modifications import callback_registry
 
+import sys  # isort: skip
+
+if sys.version_info < (3, 10):
+    # this function is deprecated in more_itertools
+    # because it is superseeded by the standard library
```
neutrinoceros, line 55:
```suggestion
    # because it is superseded by the standard library
```

neutrinoceros:
There remains only one failure in the py38_unittest job but I'm struggling to reproduce it locally. Any help would be greatly appreciated.
Hopefully this is the last thing that needs solving here, though I'm writing this before I can see the results from the py38 job on Jenkins.

setup.cfg
```diff
@@ -47,7 +47,7 @@ install_requires =
     toml>=0.10.2
     tqdm>=3.4.0
     unyt>=2.8.0
-python_requires = >=3.6
+python_requires = >=3.6,<3.12
```
neutrinoceros, line 50: To anyone wondering: the reasoning here is that the distutils builtin package is slanted for removal in Python 3.12 (which should be released around October 2023), but it is currently used in several places in yt, making it de facto incompatible with Python 3.12
Solving this doesn't seem like a priority but I'll document it as an issue !

conftest.py
```diff
@@ -71,6 +72,91 @@ def pytest_configure(config):
...
+        "ignore:the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses:DeprecationWarning",
+        # the deprecation warning message for imp changed in Python 3.10, so we ignore both versions
+        "ignore:the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses:DeprecationWarning",
+        # matplotlib warnings related to the Agg backend which is used in CI, not much we can do about it
+        "ignore:Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.:UserWarning",
+        "ignore:tight_layout . falling back to Agg renderer:UserWarning",
+        #
+        # >>> warnings from wrong values passed to numpy
+        # these should normally be curated out of the test suite but they are too numerous
+        # to deal with in a reasonable time at the moment.
```
matthewturk, line 97: I think some like this may arise from field detection.

yt/utilities/fortran_utils.py
```diff
@@ -199,10 +189,12 @@ def read_vector(f, d, endian="="):
             vec_size,
         )
     vec_num = int(vec_len / vec_size)
-    if isinstance(f, file):  # Needs to be explicitly a file
+    if isinstance(f, io.TextIOBase):
```
matthewturk, line 192: This one I'm not totally sure about.

> The TextIOBase ABC extends IOBase. It deals with streams whose bytes represent text, and handles encoding and decoding to and from strings. TextIOWrapper, which extends TextIOBase, is a buffered text interface to a buffered raw stream (BufferedIOBase). Finally, StringIO is an in-memory stream for text.

I think the check here is whether it's a filelike object; according to the numpy docs:

> file or str or Path - Open file object or filename.

(also accepts path objects)

I don't think we actually want to change this to `io.TextIOBase`, since we're explicitly really just looking for whether or not it's something that can be supplied to `np.fromfile`.  `np.frombuffer).  `io.BufferedIOBase` is not the opposite of this -- that is triggered when the file / stream object has buffering enabled, but `frombuffer` is really only interested in if something has the buffer interface.

I think the original `if` logic is okay, even if we change over the `fromstring` to `frombuffer`.  Basically, we're attempting to determine whether we need to call `read` on the object or not; I think the check here should be, is this a file, or is it an IOBase that requires `read`.

yt/utilities/tests/test_cosmology.py
```diff
@@ -216,7 +216,8 @@ def test_cosmology_calculator_answers():
     """
 
     fn = os.path.join(local_dir, "cosmology_answers.yml")
-    data = yaml.load(open(fn), Loader=yaml.FullLoader)
+    with open(fn) as fh:
+        data = yaml.load(fh, Loader=yaml.FullLoader)
```
matthewturk, line 220: I missed that we were using `FullLoader`, hmmm.

yt/visualization/volume_rendering/lens.py
```diff
@@ -180,8 +180,8 @@ def _get_sampler_params(self, camera, render_source):
         north_vec = camera.unit_vectors[1]
         normal_vec = camera.unit_vectors[2]
 
-        px = np.mat(np.linspace(-0.5, 0.5, camera.resolution[0]))
-        py = np.mat(np.linspace(-0.5, 0.5, camera.resolution[1]))
+        px = np.linspace(-0.5, 0.5, camera.resolution[0])
```
matthewturk, line 183: I am not terribly comfortable with this, but my limited testing has suggested it should produce identical results.

matthewturk (commented):
I'd like to see one of the changes reworked, but otherwise this looks good to me.

yt/utilities/tests/test_cosmology.py
neutrinoceros: Is this a problem too ? do you want to make it a new issue ?

yt/visualization/volume_rendering/lens.py
neutrinoceros: Same here. np.mat is deprecated and I wasn't able to find a reason for using it here...

yt/utilities/tests/test_cosmology.py
matthewturk: nope

yt/utilities/fortran_utils.py
neutrinoceros: The problem with the original logic is that is was built on the fact that `file` was a builtin type in Python 2. `file` was documented as "the type of the object returned by `open()`, which is `io.TextIOBase` in Python 3.

what's deprecated here is `np.fromfile`'s ability to read binary buffers

> I don't think we actually want to change this to io.TextIOBase, since we're explicitly really just looking for whether or not it's something that can be supplied to np.fromfile

I've inspected the source code upstream, but it's written in C and it's really not obvious to me how it translates...

> I think the check here should be, is this a file, or is it an IOBase that requires read.

do you mean filename here ? as I said, `file` isn't a thing in Python 3 :/

matthewturk: Nope, I mean file.

The question for me isn't whether it's a file or not -- it's whether it's a buffer or something that needs to be `read` first.

The two cases are essentially, "is this an open file handle / IO handle / etc", or is this a blob that's already been read in, that we're parsing through bit-by-bit.

neutrinoceros: OK I must confess I feel really out of my league here. Weirdly I never could get my head around the differences that seem to be important here. Would you be able to make a working suggestion, or to push to my branch ? (if so, please feel free to !)

conftest.py
neutrinoceros: oh, good call ! I'll keep this in mind if I ever feel daring enough to fix these :)

yt/utilities/fortran_utils.py
matthewturk: I think, after reading the `io` docs, that what we want is to check if it's a subclass of `io.IOBase`.  If it is, then we call `read` on it first and feed that to frombuffer.  Otherwise, we pass it to `frombuffer` directly, since it's more expensive to check if it actually exposes the buffer interface.  (And skip the else at the end, because it's OK for numpy to be the source of our error.)

The difference is essentially if we're doing:

```
f = open( ... )
read_vec(f)
```

or

```
buf = f.read(...)
read_vec(buf)
```

neutrinoceros: ok I'll patch this then, thanks a lot !

neutrinoceros: Here it goes !

yt/visualization/volume_rendering/lens.py
cphyc: I _think_ this is equivalent, but if we want to be 100% sure, we can do:
```suggestion
        px = np.linspace(-0.5, 0.5, camera.resolution[0])[np.newaxis, :]
```
here and for `py` below.
which will create an array with shape `(1, camera.resolution[0])`.

yt/visualization/volume_rendering/lens.py
```diff
@@ -180,8 +180,8 @@ def _get_sampler_params(self, camera, render_source):
         north_vec = camera.unit_vectors[1]
         normal_vec = camera.unit_vectors[2]
 
-        px = np.linspace(-0.5, 0.5, camera.resolution[0])
-        py = np.linspace(-0.5, 0.5, camera.resolution[1])
+        px = np.linspace(-0.5, 0.5, camera.resolution[0])[np.newaxis, :]
+        py = np.linspace(-0.5, 0.5, camera.resolution[1])[np.newaxis, :]
```
cphyc, line 184: Very nice syntax!

yt/data_objects/construction_data_containers.py
```diff
@@ -638,7 +639,10 @@ def __init__(
         self.ActiveDimensions = self._sanitize_dims(dims)
 
         rdx = self.ds.domain_dimensions * self.ds.relative_refinement(0, level)
-        rdx[np.where(np.array(dims) - 2 * num_ghost_zones <= 1)] = 1  # issue 602
+
+        # normalize dims as a non-zero dim array
+        dims = np.array(list(always_iterable(dims)))
+        rdx[np.where(dims - 2 * num_ghost_zones <= 1)] = 1  # issue 602
```
cphyc, line 645: What is the purpose of `np.where` here? Can't we just use 
```suggestion
        rdx[dims - 2 * num_ghost_zones <= 1] = 1  # issue 602
```

yt/data_objects/index_subobjects/grid_patch.py
```diff
@@ -268,7 +269,9 @@ def retrieve_ghost_zones(self, n_zones, fields, all_levels=False, smoothed=False
         cube._base_grid = self
         return cube
 
-    def get_vertex_centered_data(self, fields, smoothed=True, no_ghost=False):
+    def get_vertex_centered_data(
+        self, fields: List[Tuple[str, str]], smoothed=True, no_ghost=False
```
cphyc, line 273:
```suggestion
        self, fields: List[Tuple[str, str]], smoothed: bool=True, no_ghost: bool=False
```

cphyc (approved):

yt/data_objects/construction_data_containers.py
neutrinoceros: Looks reasonable to me, though I can't test it right now because I'm afk, and I'm never confident that I understand no.where. Give me a screenshot and I'll commit this :)

neutrinoceros:
pre-commit.ci autofix

yt/data_objects/construction_data_containers.py
cphyc: image

matthewturk: It is faster, but this only gets called once for each grid, and the rdx sizes are quite small.

neutrinoceros: still, it's also more readable, so win+win. I was only afraid it might not be equivalent. Committing !

neutrinoceros: Here's the problem:

```
> >       rdx[dims - 2 * num_ghost_zones <= 1] = 1  # issue 602
910
E       IndexError: boolean index did not match indexed array along dimension 0; dimension is 3 but corresponding boolean dimension is 1
```

this is from the minimal build CI. Looks like on old versions of numpy, your suggestion doesn't work. Either that or we overlooked something. In any case, I'll revert.

Issue #2639, opened 2020-06-12
[yt-4.0] ds bbox selection bug (periodicity)


AshKelly:
### Bug report

I've found a bug that when a `bbox` is applied to a dataset the yt selection routines return too many particles. In particular if you select a region, and ask the yt chunking system to load particles it will return some particles which *are not* in that region.

This bug goes away if you set `ds.periodicity = [False, False, False]`

The bug is caused by the variable `self.right_edge_shift[i] = RE[i] - DW[i]` on `L1009` of `yt/geometry/selection_routines.pyx`.

When a `bbox` is applied and a region is selected (with periodicity on) the `right_edge_shift` gets set to an array of numbers which are less than, or equal to, the `left_edge`. However, due to the `bbox` they are *NOT* less than or equal to 0 (as is true for no `bbox`)

As such on `L1064` - `select_point` the function loops through every particle and decides whether we should select them as follows:

```
    @cython.boundscheck(False)
    @cython.wraparound(False)
    @cython.cdivision(True)
    cdef int select_point(self, np.float64_t pos[3]) nogil:
        cdef int i
        for i in range(3):
            if (self.right_edge_shift[i] <= pos[i] < self.left_edge[i]) or \
               pos[i] >= self.right_edge[i]:
                return 0
        return 1
``` 

the `(self.right_edge_shift[i] <= pos[i] < self.left_edge[i])` means that any particle that resides below lower left edge will be incorrectly selected because the `self.right_edge_shift[i]` is greater than zero. I think the solution might be to set `self.right_edge_shift[i]` to `-np.inf` if a `bbox` is applied and the selected `right_edge` is less than the `domain_right_edge`.

However, I'm not sure this works for all cases. I think we need to think hard about this and get some tests added. It's not a trivial problem!

**Edit: the more I think about this, the more I suspect periodicity will never work in this way with a bbox applied and we should just disable it upon applying the bbox**

**Code for reproduction**

```
import numpy as np
import yt

n_ref = 32
fname = "../test_data/GadgetDiskGalaxy/snapshot_200.hdf5"

ptypes = ['PartType1']
left, right = (
    np.array([28007.81291869, 22181.34092477, 28461.73994773]),
    np.array([33147.21291312, 33320.7409192 , 33601.13994216]))

# Test a GadgetDiskGalaxy snapshot
bbox1 = [
    [left[0], right[0]],
    [left[1], right[1]],
    [left[2], right[2]],
]
bbox2 = None


[...]

Issue #3327, opened 2021-06-02
Bounding box is not verified in cached ParticleBitmap index


Xarthisius:
### Bug report

Existing ewah file is always used, regardless if `bounding_box` used to load dataset matches the previous value.

**Bug summary**

**Code for reproduction**

```python
from yt.utilities.answer_testing.framework import data_dir_load

def access_some_data(ds):
    field = ("gas", "temperature")
    weight_field = ("gas", "density")
    axis = 0
    obj = ds.sphere("c", (0.1, "unitary"))
    proj = ds.proj(field, axis, weight_field=weight_field, data_source=obj)
    frb = proj.to_frb((1.0, "unitary"), 256)
    frb[field]

tng59_h5 = "TNGHalo/halo_59.hdf5"
_tng59_bbox = [[45135.0, 51343.0], [51844.0, 56184.0], [60555.0, 63451.0]]

# Load for the first time with bbox to dump an ewah file
ds = data_dir_load(tng59_h5)
access_some_data(ds)

ds = data_dir_load(tng59_h5, kwargs={"bounding_box": _tng59_bbox})
access_some_data(ds)
```

**Actual outcome**

```
Traceback (most recent call last):
  File "/home/xarth/codes/xarthisius/yt/yt/visualization/fixed_resolution.py", line 159, in __getitem__
    units = self.data_source._projected_units[it]
KeyError: ('gas', 'temperature')

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/data_containers.py", line 264, in __getitem__
    rv = self.field_data[f]
KeyError: ('gas', 'temperature')

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "bug.py", line 17, in <module>
    access_some_data(ds)
  File "bug.py", line 10, in access_some_data
    frb[field]
  File "/home/xarth/codes/xarthisius/yt/yt/visualization/fixed_resolution.py", line 161, in __getitem__
    units = self.data_source[item].units
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/data_containers.py", line 270, in __getitem__
    rv = self.ds.arr(self.field_data[key], fi.units)
KeyError: ('gas', 'temperature')
```

**Expected outcome**

No angry traceback.

**Version Information**
  * Operating System: Ubuntu 18.04
  * Python Version: 3.8
[...]

PR #3386, opened 2021-06-29
BUG: Turn off periodicity for particle datasets when a bounding box is explicitly set

jzuhone:
## PR Summary

In Issue #2639 it was reported that yt selection routines return too may particles when a bounding box is explicitly applied to a particle dataset with periodicity. 

In the comment thread, it was noted that at least for now it makes sense to turn off periodicity when a bounding box is applied, which this PR does. 

Closes (in a sense) Issue #2639.

## PR Checklist

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -2034,6 +2034,17 @@ class ParticleDataset(Dataset):
             default_species_fields=default_species_fields,
         )
 
+    def _disable_periodicity_for_bbox(self):
+        # This is called when a bounding box is explicitly set by the
+        # user--currently in these circumstances periodicity behaves
+        # very oddly (see https://github.com/yt-project/yt/issues/2639),
+        # so to prevent buggy behavior we simply disable it.
+        mylog.warning(
+            "A bounding box was explicitly specified, so we "
+            "are disabling periodicity."
+        )
+        self._periodicity += (False,) * 3
+
 
 def validate_index_order(index_order):
     if index_order is None:
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -323,13 +323,8 @@ class Dataset(abc.ABC):
     def periodicity(self):
         if self._force_periodicity:
             return (True, True, True)
-        elif hasattr(self, "_domain_override"):
+        elif getattr(self, "_domain_override", False):
             # dataset loaded with a bounding box
-            if any(self._periodicity):
-                mylog.warning(
-                    "A bounding box was explicitly specified, so we "
-                    "are disabling periodicity."
-                )
             return (False, False, False)
         return self._periodicity
 
@@ -615,6 +610,12 @@ class Dataset(abc.ABC):
                     continue
                 v = getattr(self, a)
                 mylog.info("Parameters: %-25s = %s", a, v)
+        if getattr(self, "_domain_override", False):
+            if any(self._periodicity):
+                mylog.warning(
+                    "A bounding box was explicitly specified, so we "
+                    "are disabling periodicity."
+                )
 
     @parallel_root_only
     def print_stats(self):
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -2043,7 +2043,7 @@ class ParticleDataset(Dataset):
             "A bounding box was explicitly specified, so we "
             "are disabling periodicity."
         )
-        self._periodicity += (False,) * 3
+        self._periodicity = (False,) * 3
 
 
 def validate_index_order(index_order):
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -2039,11 +2039,12 @@ class ParticleDataset(Dataset):
         # user--currently in these circumstances periodicity behaves
         # very oddly (see https://github.com/yt-project/yt/issues/2639),
         # so to prevent buggy behavior we simply disable it.
-        mylog.warning(
-            "A bounding box was explicitly specified, so we "
-            "are disabling periodicity."
-        )
-        self._periodicity = (False,) * 3
+        if self._periodicity != (False,) * 3:
+            mylog.warning(
+                "A bounding box was explicitly specified, so we "
+                "are disabling periodicity."
+            )
+            self._periodicity = (False,) * 3
 
 
 def validate_index_order(index_order):
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -2042,18 +2042,6 @@ class ParticleDataset(Dataset):
             default_species_fields=default_species_fields,
         )
 
-    def _disable_periodicity_for_bbox(self):
-        # This is called when a bounding box is explicitly set by the
-        # user--currently in these circumstances periodicity behaves
-        # very oddly (see https://github.com/yt-project/yt/issues/2639),
-        # so to prevent buggy behavior we simply disable it.
-        if self._periodicity != (False,) * 3:
-            mylog.warning(
-                "A bounding box was explicitly specified, so we "
-                "are disabling periodicity."
-            )
-            self._periodicity = (False,) * 3
-
 
 def validate_index_order(index_order):
     if index_order is None:
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -323,6 +323,14 @@ class Dataset(abc.ABC):
     def periodicity(self):
         if self._force_periodicity:
             return (True, True, True)
+        elif hasattr(self, "_domain_override"):
+            # dataset loaded with a bounding box
+            if any(self._periodicity):
+                mylog.warning(
+                    "A bounding box was explicitly specified, so we "
+                    "are disabling periodicity."
+                )
+            return (False, False, False)
         return self._periodicity
 
     def force_periodicity(self, val=True):
--- a/yt/frontends/sdf/data_structures.py
+++ b/yt/frontends/sdf/data_structures.py
@@ -116,8 +116,6 @@ class SDFDataset(ParticleDataset):
         self.domain_dimensions = np.ones(3, "int32")
         if self.parameters.get("do_periodic", False):
             self._periodicity = (True, True, True)
-            if getattr(self, "_domain_override", False):
-                self._disable_periodicity_for_bbox()
         else:
             self._periodicity = (False, False, False)
 
--- a/yt/frontends/sdf/data_structures.py
+++ b/yt/frontends/sdf/data_structures.py
@@ -57,6 +57,7 @@ class SDFDataset(ParticleDataset):
                 bbox = bbox.transpose()
             self.domain_left_edge = bbox[:, 0]
             self.domain_right_edge = bbox[:, 1]
+            self._disable_periodicity_for_bbox()
         else:
             self.domain_left_edge = self.domain_right_edge = None
         self.sdf_header = sdf_header
--- a/yt/frontends/sdf/data_structures.py
+++ b/yt/frontends/sdf/data_structures.py
@@ -114,7 +114,7 @@ class SDFDataset(ParticleDataset):
             self.domain_right_edge *= self.parameters.get("a", 1.0)
 
         self.domain_dimensions = np.ones(3, "int32")
-        if "do_periodic" in self.parameters and self.parameters["do_periodic"]:
+        if self.parameters.get("do_periodic", False):
             self._periodicity = (True, True, True)
             if getattr(self, "_domain_override", False):
                 self._disable_periodicity_for_bbox()
--- a/yt/frontends/sdf/data_structures.py
+++ b/yt/frontends/sdf/data_structures.py
@@ -116,10 +116,10 @@ class SDFDataset(ParticleDataset):
         self.domain_dimensions = np.ones(3, "int32")
         if "do_periodic" in self.parameters and self.parameters["do_periodic"]:
             self._periodicity = (True, True, True)
+            if getattr(self, "_domain_override", False):
+                self._disable_periodicity_for_bbox()
         else:
             self._periodicity = (False, False, False)
-        if self._domain_override:
-            self._disable_periodicity_for_bbox()
 
         self.cosmological_simulation = 1
 
--- a/yt/frontends/sdf/data_structures.py
+++ b/yt/frontends/sdf/data_structures.py
@@ -57,7 +57,6 @@ class SDFDataset(ParticleDataset):
                 bbox = bbox.transpose()
             self.domain_left_edge = bbox[:, 0]
             self.domain_right_edge = bbox[:, 1]
-            self._disable_periodicity_for_bbox()
         else:
             self.domain_left_edge = self.domain_right_edge = None
         self.sdf_header = sdf_header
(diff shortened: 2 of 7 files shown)


yt/data_objects/static_output.py
```diff
@@ -1934,6 +1934,17 @@ def __init__(
...
+    def _disable_periodicity_for_bbox(self):
+        # This is called when a bounding box is explicitly set by the
+        # user--currently in these circumstances periodicity behaves
+        # very oddly (see https://github.com/yt-project/yt/issues/2639),
+        # so to prevent buggy behavior we simply disable it.
+        mylog.warning(
+            "A bounding box was explicitly specified, so we "
+            "are disabling periodicity."
+        )
+        self._periodicity += (False,) * 3
```
neutrinoceros, line 1946: I believe this should solve the failures
```suggestion
        self._periodicity = (False,) * 3
```

neutrinoceros (commented):
Thank you job for patching this.

yt/data_objects/static_output.py
jzuhone: Whoops, thanks

matthewturk (approved):
I think this is correct.  I see no reason not to merge immediately.

jzuhone:
@neutrinoceros Wait on merging until I check something real quick

AshKelly (approved):

jzuhone:
Ok, I did this in a hurry last night and I wasn't careful enough. I think that in some cases this new function gets called _before_  `_parse_parameter_file` sets the periodicity itself, so I need to fix that before this can go in.

Sorry about this.

neutrinoceros:
> Ok, I did this in a hurry last night and I wasn't careful enough. I think that in some cases this new function gets called before _parse_parameter_file sets the periodicity itself, so I need to fix that before this can go in.
Sorry about this.

No problem, we share a responsibility in reviewing this in a hurry too. I think we're all a bit shaky with the perspective of the upcoming release. Could you switch to draft mode and back to give a clear signal when this is ready for review again ?

jzuhone:
If the tests pass it's good

yt/frontends/sdf/data_structures.py
```diff
@@ -119,6 +118,8 @@ def _parse_parameter_file(self):
             self._periodicity = (True, True, True)
         else:
             self._periodicity = (False, False, False)
+        if self._domain_override:
+            self._disable_periodicity_for_bbox()
```
neutrinoceros, line 122: I think that this will wrongfully emit a warning even in case the `else` right before this is hit. IMO this conditional should be included in the previous one.

neutrinoceros (changes requested):
thanks for being more cautious ! I think there's one more tiny thing to improve here but otherwise this looks good.

yt/frontends/sdf/data_structures.py
jzuhone: This is fixed in the latest commit--thanks for pointing it out

yt/frontends/sdf/data_structures.py
```diff
@@ -116,6 +116,8 @@ def _parse_parameter_file(self):
         self.domain_dimensions = np.ones(3, "int32")
         if "do_periodic" in self.parameters and self.parameters["do_periodic"]:
```
neutrinoceros, line 117: Could we make idioms consistent here while we're at it ?
```suggestion
        if self.parameters.get("do_periodic", False):
```

neutrinoceros (dismissed):

jzuhone:
So it looks like we'd have to fix this by modifying `sph_answer`--the Arepo test fails because one of the things it checks involves getting a region which goes outside of the bounding box, and of course our typical solution to that is to _enable_ periodicity. Not sure what to do about this now. 

Don't let this hold up 4.0.

neutrinoceros:
ok, just going to use the "do not merge" label right now to prevent accidents since we have THREE approvals (mine included  🙈) on a PR that's not ready yet.
Thanks John for your patience, I hope we can solve this soon :)

jzuhone:
I'm resurrecting this PR and I think I'll need to fix the Arepo answer tests.

neutrinoceros:
requesting a review from myself, I'll look over this after CI completes !

yt/data_objects/static_output.py
```diff
@@ -1971,6 +1971,17 @@ def __init__(
             default_species_fields=default_species_fields,
         )
 
+    def _disable_periodicity_for_bbox(self):
+        # This is called when a bounding box is explicitly set by the
+        # user--currently in these circumstances periodicity behaves
```
neutrinoceros, line 1976:
```suggestion
        # user--previously in these circumstances periodicity behaved
```

yt/data_objects/static_output.py
```diff
@@ -1971,6 +1971,17 @@ def __init__(
...
+        # very oddly (see https://github.com/yt-project/yt/issues/2639),
+        # so to prevent buggy behavior we simply disable it.
+        mylog.warning(
+            "A bounding box was explicitly specified, so we "
+            "are disabling periodicity."
+        )
+        self._periodicity = (False,) * 3
```
neutrinoceros, line 1983: I guess the warning should only be emitted if the method has an effect (i.e. if periodicity isn't already False)

yt/data_objects/static_output.py
jzuhone: It's still true that periodicity behaves oddly if we apply a bounding box--I don't think the change in tense is what we want here.

neutrinoceros: ah I see, nevermind

neutrinoceros:
note: the docs build failed with 
```
ERROR: Could not build wheels for yt, which is required to install pyproject.toml-based projects
```
this is unrelated to the PR and caused by newest versions of setuptools, see https://github.com/pypa/setuptools/issues/3322

jzuhone:
@Xarthisius I think the Jenkins test are failing because of Issue #3327

Xarthisius:
> @Xarthisius I think the Jenkins test are failing because of Issue #3327

I'm trying to come up with something that wouldn't be a fully-fledged solution like redesigning ewah storage format, and on the other hand, wouldn't rely on removing all ewah files for each test run. How about allowing to ignore existing ewah as a kwarg to load?

jzuhone:
> I'm trying to come up with something that wouldn't be a fully-fledged solution like redesigning ewah storage format, and on the other hand, wouldn't rely on removing all ewah files for each test run. 

Do you need to remove all of the files, or just the ones being used in the particular PR? 

> How about allowing to ignore existing ewah as a kwarg to load? Another option: appending unique_id to ewah filename?

I think either of these solutions is acceptable.

Xarthisius:
> Do you need to remove all of the files, or just the ones being used in the particular PR?

From the tests' perspective I don't know which files are affected without doing some advanced parsing of diffs or something similar. Also I'd have to do it always, not once. If I removed ewah e.g. for your PR only, new one would be generated and it would break all the other test runs until your PR is merged.

> > How about allowing to ignore existing ewah as a kwarg to load? Another option: appending unique_id to ewah filename?
> 
> I think either of these solutions is acceptable.

I'll try to issue a PR in a (hopefully) near future.

neutrinoceros:
> How about allowing to ignore existing ewah as a kwarg to load? Another option: appending unique_id to ewah filename?

For the record, I have a preference for the second solution. Adding kwargs to `load` is a last resort IMO.

neutrinoceros:
In #3930 I'm trying to overhaul `unique_identifier`'s implementation (reducing duplication, mostly). Doing so I see that most frontends implement a reproducible id (only depends on the input filename), but some don't, and default to using `time.time()`. Is this an obstacle to your second solution @Xarthisius ? Are there downsides to changing/simplifying the implementations, like additional maintenance ?

Xarthisius:
> In #3930 I'm trying to overhaul `unique_identifier`'s implementation (reducing duplication, mostly). Doing so I see that most frontends implement a reproducible id (only depends on the input filename), but some don't, and default to using `time.time()`. Is this an obstacle to your second solution @Xarthisius ? Are there downsides to changing/simplifying the implementations, like additional maintenance ?

I wasn't planning on using Datasets' `unique_identifier`. As you noted, it's mostly based on the original filename and it doesn't capture the info we need to store in ewah to avoid collisions (like bounding box).

neutrinoceros:
Oh I see, thanks for the clarification !

neutrinoceros:
With #3916 in mind, I wonder if the present patch is acceptable. Doesn't it make it completely impossible to work with a bounding box spanning over multiple corners of a periodic domain ?

matthewturk:
@neutrinoceros I think in those particular cases we should probably allow for a translation argument.  But is it the case that a bounding box that spans multiple corners will already run up against issues?  I think it may very well be.

neutrinoceros:
> But is it the case that a bounding box that spans multiple corners will already run up against issues?

Most likely yeah, since there are already problems without a bounding box, as we can see in https://github.com/yt-project/yt/issues/3916

I'm worried about trying to solve _that_ issue only to get locked out of solution because of the present PR.

jzuhone:
@neutrinoceros Once you've specified a bounding box that is potentially different from the box that the set of particles originally came from, the whole notion of periodicity goes out the window. Especially if you cross only one axis. This solution is needed currently to make datasets like this behave even somewhat normally under these circumstances until we come up with a better strategy.

yt/data_objects/static_output.py
```diff
@@ -1984,6 +1984,18 @@ def __init__(
             default_species_fields=default_species_fields,
         )
 
+    def _disable_periodicity_for_bbox(self):
+        # This is called when a bounding box is explicitly set by the
+        # user--currently in these circumstances periodicity behaves
+        # very oddly (see https://github.com/yt-project/yt/issues/2639),
+        # so to prevent buggy behavior we simply disable it.
+        if self._periodicity != (False,) * 3:
```
neutrinoceros, line 1992:
```suggestion
        if any(self.periodicity):
```

yt/data_objects/static_output.py
```diff
@@ -1984,6 +1984,18 @@ def __init__(
...
+            mylog.warning(
+                "A bounding box was explicitly specified, so we "
+                "are disabling periodicity."
+            )
+            self._periodicity = (False,) * 3
```
neutrinoceros, line 1997: using the encapsulated form will ease refactoring. It also sets the `self._force_periodicity` flag. 
```suggestion
            self.force_periodicity(False)
```

neutrinoceros (commented):
Ok thanks for clarifying. I think the better (long term) solution would be to have Dataset objects store boundary conditions for each direction/side, rather than a single boolean per direction for periodicity.
In that framework, it should be possible to set special boundaries for when a bounding box is set.

One last question: how do you know you've updated all frontends that needed this change ?

yt/data_objects/static_output.py
neutrinoceros, line 1997: actually, this logic could just be encapsulated inside the `periodicity` property, couldn't it ?
Then you wouldn't need to update anything else, as the change would already affect all frontends !

jzuhone: You don't want to change all frontends--only the ones that accept a `bounding_box` argument

neutrinoceros: Aren't those the same that use the optional attribute `_domain_override` ?

jzuhone: Yes, but I'm not sure what you are getting at here.

neutrinoceros: What I mean is that it seems to me that all of this could be accomplished with the `Dataset.periodicity` property definition using a `if getattr(self, "_domain_override", False):` clause. Is this incorrect ?

jzuhone: Ah, ok--yes, that's correct.

neutrinoceros: Here's exactly what I have in mind
```patch
diff --git a/yt/data_objects/static_output.py b/yt/data_objects/static_output.py
index 96d013992..328f66262 100644
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -322,6 +322,14 @@ class Dataset(abc.ABC):
     def periodicity(self):
         if self._force_periodicity:
             return (True, True, True)
+        elif hasattr(self, "_domain_override"):
+            # dataset loaded with a bounding box
+            if any(self._periodicity):
+                    "A bounding box was explicitly specified, so we "
+                    "are disabling periodicity."
+                )
+            return (False, False, False)
         return self._periodicity

     def force_periodicity(self, val=True):
```

jzuhone: Ok I've added it

neutrinoceros:
@jzuhone is this ready ? I'll take one last look again tomorrow but if you're done with it I might just merge it then

jzuhone:
I recognize now that the `_disable_periodicity_for_bbox` method is unnecessary, but I will wait to remove it until after I see the tests pass--a bit worried about the potential answer test failure due to the wrong EWAH file. Will know in a few minutes...

jzuhone:
@Xarthisius so this PR will keep failing the `halo_59` answer tests because of https://github.com/yt-project/yt/issues/3327. Any idea what we can do about it?

Xarthisius:
> @Xarthisius so this PR will keep failing the `halo_59` answer tests because of #3327. Any idea what we can do about it?

Sans fixing the issue you've mentioned, we could remove ewah file every time the test is run.

jzuhone:
So it turns out that this is fixed by simply centering the bounding box of the AREPO dataset on the halo center. The EWAH file is a non-issue here.

neutrinoceros:
@jzuhone how can we check that the test isn't flaky ?

jzuhone:
@neutrinoceros The reason the test was having so much trouble before is that I had originally changed the bounding box in such a way that one of the sphere tests was landing in a part of the domain where there was no data (I did not realize this originally). Now I have placed the halo in the center of the domain and the immediate result was that the test failed because the answers were not identical, which is precisely what we should we should expect if we change the bounding box. I looked at the projection difference images and it showed that the object merely shifted, which is also what I expected. So I am pretty confident in the test now.

neutrinoceros:
Alright that make sense, thank you for this summary ! I'll run the test a couple times more just in case
@yt-fido tes this please

jzuhone:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

jzuhone:
@yt-fido test this please

jzuhone:
Ok, we ran the tests twice more--this is good to go.

neutrinoceros (commented):
I think we should add a couple lines to test that `ds.periodicity`  is correctly outputted with and without the bounding box, to avoid future regressions. Other than that LGTM

yt/data_objects/static_output.py
```diff
@@ -323,6 +323,14 @@ def unique_identifier(self, value):
     def periodicity(self):
         if self._force_periodicity:
             return (True, True, True)
+        elif hasattr(self, "_domain_override"):
+            # dataset loaded with a bounding box
+            if any(self._periodicity):
+                mylog.warning(
```
neutrinoceros, line 329: This warning might trigger at surprising times because the periodicity property is often accessed internally. It would probably be preferable to issue the warning when the bounding box is created instead.

neutrinoceros (approved):

PR #3402, opened 2021-07-01
Change the loop order to avoid executing the same test 3 times

Xarthisius:
Just to be extra sure that we're doing the right thing, we've been running the same `FieldValuesTest` **3 times** every time `small_patch_amr` was used. No big deal! It's better to be safe than sorry! However, as I grow older I discover a daredevil in me. So I say: let's live our lives dangerously and run this test only once! How about that?

## PR Summary

This PR switches the order of loops in `small_patch_amr` to allow `FieldValuesTest` to be run once, since it's independent of `axis` parameter.


neutrinoceros (approved):
This makes sense. Good catch !

neutrinoceros:
got a network error in load_sample tests, I relaunched the culprit job.

matthewturk (approved):
I honestly feel like this is my fault and this is a huge improvement

PR #3403, opened 2021-07-02
Update MANIFEST.in to include *.pxi and conftest.py

Xarthisius:
## PR Summary

`check-manifest` shows a lot of missing files. Especially `yt/geometry/_selection_routines/*.pxi` look critical, i.e. building yt from source will fail without them. I threw in `conftest.py` as a cherry on top. Prunes / excludes that I added are mostly cosmetic. With this PR I still get:

```
missing from sdist:
  yt/utilities/lib/cykdtree/c_kdtree.cpp
  yt/visualization/mapserver/html/Leaflet.Coordinates-0.1.5.css
  yt/visualization/mapserver/html/Leaflet.Coordinates-0.1.5.src.js
```

`yt/utilities/lib/cykdtree/c_kdtree.cpp` is empty and I don't know if it's needed. Maybe we should remove it. What about vendored Leaflet?

PR #3404, opened 2021-07-02
MAINT: add check-manifest to pre-commit toolkit

neutrinoceros:
## PR Summary
In reaction to PR #3403, add check-manifest to .pre-commit-config.yaml
This will *fail* on pre-commit.ci as long as #3403 isn't merged and the remaining issues aren't solved, but this PR shouldn't be necessary for the 4.0 release.
@Xarthisius, feel free to integrate this in #3403 if you wish _and_ if that PR is intended to solve *all* issues currently reported by check-manifest.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,6 +20,10 @@ repos:
     rev: v1.17.0
     hooks:
     -   id: setup-cfg-fmt
+-   repo: https://github.com/mgedmin/check-manifest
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Changed my mind. This hook _cannot_ run fast since it needs to check to the entire repo, so it is completely impractical in local mode. A custom GH workflow might be a better solution here. I'll close this in the mean time.

PR #3405, opened 2021-07-02
MAINT: add a github workflow to run check-manifest

neutrinoceros:
## PR Summary
Complementary to #3403, but not required for the release.
I might need to configure the tool in pyproject.toml


neutrinoceros:
Ok the workflow failed _as expected_. I'll rerun it by rebasing after #3403 is merged, and _then_ I expect it to pass.

neutrinoceros:
As discussed there, the workflow _still_ won't pass without a solution to express that it's okay that we have .c and .cpp files including in the dist that are _not_ part of version control.
For now I won't be working on this to avoid jamming CI queues.

neutrinoceros:
There's a final blocking issue resisting me, where `answer-store` is reported as missing from sdist, despite the fact that it's already explicitly pruned in MANIFEST.in, and I can't reproduce this locally.

neutrinoceros:
This was solved by adding  a step to the workflow to initialise submodules. I believe this is now ready !

cphyc (approved):
LGTM!

.github/workflows/check-manifest.yaml
```diff
@@ -0,0 +1,26 @@
+name: check MANIFEST.in
+on: [pull_request]
```
matthewturk, line 2: Do we want this on every PR?

pyproject.toml
```diff
@@ -72,3 +72,64 @@ addopts = '''
     --ignore='yt/frontends/owls_subfind/tests/test_outputs.py'
     --ignore='yt/frontends/ramses/tests/test_outputs.py'
 '''
+
+
+[tool.check-manifest]
```
matthewturk, line 77: This does add another place that we include all the names of the generated cython files, which I'm a little hesitant about.

.github/workflows/check-manifest.yaml
neutrinoceros: For lack of a smarter approach, I think so ? Would you prefer we ran this at a fixed frequency?

pyproject.toml
neutrinoceros: Yeah, not a fan either. If you think the cost isn't worth the time this supposedly saves on releases, maybe we can just drop it.

.github/workflows/check-manifest.yaml
matthewturk: I think I'm maybe reacting to seeing all the existing github workflows and how they can, at times, be a bit intimidating.

neutrinoceros: FWIW the h5py one has seen so little failures I'm aiming at getting rid of it soon :)

neutrinoceros:
@Xarthisius since this is mostly motivated by your work on the 4.0 release, I'll leave you judge whether this has value. Otherwise feel free to close :)

PR #3406, opened 2021-07-02
Port wheels build to GHA

Xarthisius:
## PR Summary

This adds GHA for building wheels. It covers both the release process and dev (building after every push to master).

Things noteworthy:
* We need to statically link `libstdc++` due to fact we're shipping/building a lot of C++14 code. Otherwise linux wheels are unusable  due to the ABI conflicts. Some info on this can be found [here](https://github.com/pypa/cibuildwheel/blob/184a9feb24ee38d0b0a39fa93b6bb8247cbd636b/docs/cpp_standards.md)
* I switched to `oldest-supported-numpy` for numpy build dependency.
* 32bit linux build fails and I disable it. I don't think anyone uses i686 these days anyway...

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -46,7 +46,7 @@ jobs:
         run: |
           python -m cibuildwheel --output-dir dist
         env:
-          CIBW_BUILD: "cp37-* cp38-* cp39-*"
+          CIBW_BUILD: "cp36-* cp37-* cp38-* cp39-*"
           CIBW_ARCHS_LINUX: "x86_64"
           CIBW_ARCHS_MACOS: "x86_64"
           CIBW_ARCHS_WINDOWS: "auto"
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -1,0 +1,60 @@
+name: Build CI Wheels
+
+on:
+  push:
+    branches:
+      - main
+      - stable
+    tags:
+      - 'yt-*'
+
+jobs:
+  build_wheels:
+    name: Wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macos-latest]
+      fail-fast: false
+
+    steps:
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: 3.9
+
+      - uses: s-weigand/setup-conda@v1
+        if: matrix.os == 'windows-latest'
+        with:
+          update-conda: true
+          conda-channels: conda-forge
+          activate-conda: true
+          python-version: 3.9
+
+      - uses: actions/checkout@v2
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==1.9.0
+
+      - name: Install dependencies and yt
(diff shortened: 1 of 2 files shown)


matthewturk:
For what it's worth, the `c++14` stuff was surprising when I discovered it became a dep in the EWAH code after upgrading to a more recent version.  I am hopeful that in the future we'll be able to get around it somewhat by removing the EWAH code directly and exposing only Cython wrappers from external libraries.

.github/workflows/wheels.yaml
```diff
@@ -0,0 +1,60 @@
+name: Build CI Wheels
+
+on:
+  push:
+    branches:
+      - main
+      - stable
+    tags:
+      - yt-*
```
matthewturk, line 9: am I reading right that this means anytime we push, *and* anytime we push a tag with a version, it builds?

matthewturk (approved):
This is absolutely awesome.  Thank you.

.github/workflows/wheels.yaml
matthewturk: ah nm, I see:

> It covers both the release process and dev (building after every push to master).

neutrinoceros:
This is so cool, thank you!

pyproject.toml
```diff
@@ -11,8 +11,7 @@ requires = [
   # we forbid it until we can properly test against it
   "Cython>=0.26.1,<3.0; python_version=='3.6'",
   "Cython>=0.29.21,<3.0; python_version>='3.7'",
-  "numpy==1.13.3; python_version=='3.6' and platform_system!='AIX'",
-  "numpy>=1.19.2; python_version>='3.7' and platform_system!='AIX'",
+  "oldest-supported-numpy",
```
neutrinoceros, line 14: btw the "AIX issue" comment above should be dropped at the same time

.github/workflows/wheels.yaml
```diff
@@ -0,0 +1,60 @@
...
+        env:
+          dependencies: "full"
+          LDFLAGS: "-static-libstdc++"
+        run: source ./tests/ci_install.sh
+
+      - name: Build wheels for CPython
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: "cp37-* cp38-* cp39-*"
```
neutrinoceros, line 49: Is Python 3.6 left out intentionally ?

Xarthisius: I thought it's EOL already, but apparently that happens only in 4 months...

pyproject.toml
Xarthisius: done (forced push with commit adding `oldest-supported-numpy`)

.github/workflows/wheels.yaml
neutrinoceros: FYI, we'll drop support for Python 3.6 on _next_ release, see #2917

pyproject.toml
neutrinoceros: thank you !

neutrinoceros:
I don't believe the crash in the minimal build is reproducible, so I relaunched the job.

neutrinoceros (approved):

neutrinoceros:
I'll leave you guys decide when to merge this !

PR #3407, opened 2021-07-02
MAINT: bump version to 4.0.0

neutrinoceros:
## PR Summary
:tada:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = yt
-version = 4.0.dev0
+version = 4.0.0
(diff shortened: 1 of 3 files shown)


Xarthisius:
I think you meant `4.1.dev0` for main :)

munkm:
This shouldn't happen via PR. What we'll do is bump to 4.0.0 on stable and then issue a PR on main to bump to 4.1.dev0. But we should do it after the release is finished so we can also get the tag from stable into main.

neutrinoceros:
Sorry guys, this workflow wasn't clear to me, indeed there's no problem, disregard ! :)

PR #3408, opened 2021-07-03
MAINT: remove a unreachable conditional

neutrinoceros:
## PR Summary

Found this while working on something else.
This guard condition is unreachable since PR #2556 made the base Dataset class abstract, which means it _cannot_ be instantiated.


matthewturk (approved):
Is this really

PR #3410, opened 2021-07-04
refactor: determine compatible matplotlib backend semi-dynamically. Incidentally add support for .svgz image format.

neutrinoceros:
## PR Summary

Drop a hard-coded list of supported file formats and discover them at runtime via Matplotlib instead.
Incidentally, this adds support for saving to .svgz, which is compatible with `FigureCanvasSVG` but was left out
on our side, likely because it wasn't a thing at the time this code was written. The list of supported file format will then naturally expand if more formats are added upstream, and reciprocally, this avoids creating a situation where we list a format as supported even though it requires a more recent version of Matplotlib than is installed.

This refactor passes yt/visualization/tests/test_commons.py


yt/visualization/_commons.py
```diff
@@ -44,9 +66,9 @@ def validate_image_name(filename, suffix: Optional[str] = None) -> str:
     suffix = normalize_extension_string(suffix)
 
     if suffix not in SUPPORTED_FORMATS:
-        raise ValueError("Unsupported file format '{suffix}'.")
+        raise ValueError(f"Unsupported file format '{suffix}'.")
```
cphyc, line 69: Nice catch!

cphyc (approved):

yt/visualization/_commons.py
neutrinoceros: always nice to fix my own mistakes before I get caught :)

Xarthisius:
@yt-fido test this please

neutrinoceros:
I don't remember why you triggered a new job on Jenkins, but I imagine you'd like to revisit this, @Xarthisius ?

matthewturk:
I'm good with this post resolution

neutrinoceros:
Alright I'll resolve the conflict now and come back later to rettriger Jenkins !

neutrinoceros:
@matthewturk I assume you'd want to use the auto-merge button here :)

neutrinoceros:

PR #3413, opened 2021-07-06
Updating EWAH from upstream lemire/EWAHBoolArray

matthewturk:
This updates to the most recent EWAHBoolArray ( https://github.com/lemire/EWAHBoolArray/ ) to attempt to fix Windows build issues around ambiguous uses of `size_t` and various `uint` types that are breaking our conda-forge builds in https://github.com/conda-forge/yt-feedstock/pull/50 .


jzuhone:
I don't understand the rockstar failure.

munkm:
Very non-rockstar, amirite? 🚀 ⭐ 😎

matthewturk:
So I have dug in a little bit.  What I've found is that the `bad_alloc` happens whenever the gadget halos are loaded from an on-disk index.  If I remove it, it works fine; if I do not, it does not.  So I'm thinking there's something about the way we're caching the data that causes it to read in, maybe big/little endian or something, incorrectly.

matthewturk:
So the problem seems to be something about how we're using the read and write functions, which were previously defined in `ewah.h` and are now in `ewah-inl.h`.  Previously:

```c++
template <class uword>
size_t EWAHBoolArray<uword>::read(std::istream &in, const bool savesizeinbits) {
  size_t read = 0;
  if (savesizeinbits) {
    in.read(reinterpret_cast<char *>(&sizeinbits), sizeof(sizeinbits));
    read += sizeof(sizeinbits);
  } else {
    sizeinbits = 0;
  }
  size_t buffersize(0);
  in.read(reinterpret_cast<char *>(&buffersize), sizeof(buffersize));
  read += sizeof(buffersize);
  buffer.resize(buffersize);
  if (buffersize > 0) {
    in.read(reinterpret_cast<char *>(&buffer[0]),
            static_cast<std::streamsize>(sizeof(uword) * buffersize));
    read += sizeof(uword) * buffersize;
  }
  return read;
}
```

and now:

```c++
template <class uword>
size_t EWAHBoolArray<uword>::read(std::istream &in, const bool savesizeinbits) {
  size_t read = 0;
  if (savesizeinbits) {
    uint64_t tmp;
    in.read(reinterpret_cast<char *>(&tmp), sizeof(tmp));
    read += sizeof(tmp);
    sizeinbits = static_cast<size_t>(tmp);
  } else {
    sizeinbits = 0;
  }
  size_t buffersize(0);
  uint64_t tmp;
  in.read(reinterpret_cast<char *>(&tmp), sizeof(tmp));
  read += sizeof(tmp);
  buffersize = static_cast<size_t>(tmp);
  buffer.resize(buffersize);
  if (buffersize > 0) {
    in.read(reinterpret_cast<char *>(&buffer[0]),
            static_cast<std::streamsize>(sizeof(uword) * buffersize));
    read += sizeof(uword) * buffersize;
  }
  return read;
}
```

The segfault happens in the `buffer.resize` call.  When I do a `std:cerr` of `buffersize` in the new version it produces `93825034668176`, which is definitely too much memory.  I have to stop working on this for a little bit right now, but I want to finish by noting that this happens when we load index files created by the new version -- not just when we load index files created by the old version.  So, it's self-contained, and likely due to a mismatch somewhere in *our* code, that previously was OK and no longer is.

Xarthisius:
@yt-fido test this please

matthewturk:
I think I have figured out what the issue is, and I am working to fix it.  I believe we are feeding finalized `sstream` objects in, and because it's using temporary variables instead of instance variables, they are not being read, and they are retaining whatever their initial value was (uninitialized).  Working on a fix.

matthewturk:
There were two ways I could see to fix it -- initialize the `tmp` variable (which would modify upstream code) or just do a check for `EOF`, and I went with the latter.  (I added a super long comment about this too.)

Xarthisius (approved):

neutrinoceros (approved):
LGTM

PR #3418, opened 2021-07-07
DEPR: deprecate iyt

neutrinoceros:
## PR Summary

This script can essentially be emulated with
```shell
$ ipython -i "import yt"
```
It hasn't received any significant update/maintenance for 13yrs, and doesn't have a clear added value now with a decade of improvements in IPython's autocompletion (and it's getting env better in IPython 8.0, thanks to our own @cphyc : https://github.com/ipython/ipython/pull/12395).
I think it should be removed.

### DIFF
--- a/doc/cheatsheet.tex
+++ b/doc/cheatsheet.tex
@@ -116,7 +116,6 @@ Commands can be followed by
 {\bf {-}{-}help} (e.g. {\bf yt render {-}{-}help}) for detailed help for that command
 including a list of the available flags.
 
-\texttt{iyt}\textemdash\ Load yt and IPython. \\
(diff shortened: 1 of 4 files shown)


matthewturk:
I'm OK with this!

matthewturk (approved):

cphyc (approved):
Seems good to slowly fade this out.

chummels:
I am a little concerned with this in a few regards:

1) I regularly use `yt load` from the command-line to start up ipython sessions with yt pre-loaded.  I would've liked to have had some say in this before it got merged.
2) When did we only need one approve before merging code that deprecates and removes a long-standing feature?  Particularly when it was proposed and merged in less than 24h and with a single approve.

jzuhone:
@chummels this doesn't get rid of `yt load`.

PR #3419, opened 2021-07-07
MAINT: readd firefly_api to full install target

### LINKED ISSUE #3415: BUG: firefly-dependent code is currently broken
### Bug report

**Bug summary**

In #3362 I dropped firefly from our optional deps following its removal from GitHub + PyPI
However the parts of yt that depend on it are still here and weren't updated to work with the replacement.

I don't think that fixing this is urgent, but I'm noting it down as a reminder that it should be addressed for yt 4.1

ping @agurvich :)

neutrinoceros:
## PR Summary
fix #3415
This is a temporary yet stable solution to #3415 while firefly_api's replacement isn't as easy to install.
Thanks to @agurvich for reuploading firefly_api to PyPI, making this patch possible.

An issue I've run into while testing this locally is that the one test this enable
(namely yt/data_objects/tests/test_firefly.py) is now failing because firefly raises a warning
(treated as an error since #3380).

If the warning can be easily avoided by refactoring the test, this would be the prefered solution.
Otherwise, if we _know_ that the future replacement for firefly_api isn't going to pose a similar "complication", we could just go back to ignoring this warning as was implicitly done last time we ran this test.

Here's the exact warning I'm getting:

```
firefly_api.errors.FireflyWarning: JSONdir: /var/folders/xh/yfmp7l9n3x50ln2yvrndt7c801pbkm/T/tmpwm92eyqz/data/test -- is not a sub-directory of Firefly/static/data.
This may produce confusing or inoperable results. As such, we will create a symlink for you. You're welcome.
```
@agurvich, can you weight in on this please ?

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -158,6 +158,12 @@ def pytest_configure(config):
             ),
         )
 
+    if find_spec("firefly_api"):
+        # see https://github.com/yt-project/yt/pull/3419
+        config.addinivalue_line(
+            "filterwarnings", r"ignore::firefly_api.errors.FireflyWarning"
+        )
+
 
 def pytest_collection_modifyitems(config, items):
     r"""
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,6 +74,7 @@ full =
     astropy>=4.0.1,<5.0.0
     f90nml>=1.1.2
     fastcache~=1.0.2
+    firefly_api==0.0.3
     glueviz~=0.13.3
     h5py>=3.1.0,<4.0.0
     libconf~=1.0.1


agurvich:
I've since changed that warning to a print statement, it's an important warning since the browser does not have permission to read data outside the "Firefly/data" sub-directory but the way I handle this is different nowadays, I allow the user to save the data anywhere and then will create a symlink (and will just notify the user that I have done so). 

So 

> Otherwise, if we know that the future replacement for firefly_api isn't going to pose a similar "complication", we could just go back to ignoring this warning as was implicitly done last time we ran this test.

We can be certain that it won't pose a similar "complication" in the future. If we can just ignore the warning that is appropriate for now. I am working on uploading the latest version of "firefly_api" (not really just the api anymore, but the whole package so I'm still mulling over the exact branding since `firefly` is taken) but once that's done I will submit a PR to address all this posthaste, it's at the top of my priority list I promise!

agurvich:
Thank you for spending so much of your time thinking about this, I really do apologize for the trouble :[

neutrinoceros:
No worries, this is part of the maintenance job, no need for apologies :)
I appreciate your prompt answer. I will ignore the warning for now an you can revisit the whole integration with yt whenever best suits you !

neutrinoceros:
Surprisingly to me, the test still fails, but now in a bad way:

```
E                       FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/xh/yfmp7l9n3x50ln2yvrndt7c801pbkm/T/tmpr5lma5eo/data/test' -> '/Users/robcleme/.pyenv/versions/3.9.5/envs/yt-dev/lib/python3.9/static/data/test'
```
Maybe it wasn't run for a while on our side and has been broken for much longer than I thought. In any case, any help here is welcome :)

agurvich:
Hopefully I git'd correctly, I put in a [PR for your firefly_support branch](https://github.com/neutrinoceros/yt/pull/7). I updated for the new Firefly api and updated the documentation accordingly.

neutrinoceros:
closing since the material here was incorporated to #3435

PR #3421, opened 2021-07-08
ENH: Allow stream frontends to accept callables

matthewturk:
## PR Summary

At present, all the stream frontends can only accept actual, instantiated arrays.  So for instance, this means that any calls to `load_uniform_grid` require that the entire array already be present in memory.

This isn't a hard-and-fast requirement of the Stream frontend -- and in fact, is opposite the original purpose of the Stream frontend! -- but it has evolved to be the only way to load data into it.  This PR starts the process of reversing this, and enabling functions to be supplied, instead.

Ideally, while this is somewhat useful at present, in the long run this will allow us to expose a much simpler interface for prototyping frontends, and supplying items to the Stream frontend without actually writing a fully-fledged frontend *and* without requiring that the data be loaded in advance.

It is still at a draft state, and is somewhat brittle.

## PR Checklist


matthewturk:
@chummels I wanted to ping you on this one, as I think it would make it much easier to prototype some of the frontends you've been looking at, specifically as it might allow you to dynamically repartition grids.

yt/frontends/stream/definitions.py
```diff
@@ -177,7 +180,9 @@ def process_data(data, grid_dims=None):
     # At this point, we have arrays for all our fields
     new_data = {}
     for field in data:
-        n_shape = len(data[field].shape)
+        n_shape = 3
```
matthewturk, line 183: I wasn't incredibly happy with this.  Right now, if it's callable, it assumes it's 3D.  I'm wondering if maybe we could have a decorator that supplied some metadata about the field that would either be returned or would be assigned to the function object.

yt/frontends/stream/io.py
```diff
@@ -18,6 +18,8 @@ def __init__(self, ds):
     def _read_data_set(self, grid, field):
         # This is where we implement processor-locking
         tr = self.fields[grid.id][field]
+        if callable(tr):
```
matthewturk, line 21: This, and the change below, is not the *most* elegant.  I'm not sure I have a better solution.

cphyc (approved):
Awesome! Provided we add tests & doc, this would be a really good addition

matthewturk:
I've rebased this for fixing the tests.

matthewturk:
@chrishavlin and I have been talking and there are a handful of changes we want to make, in addition to fixing any broken tests and adding docs and examples.

 1. [ ] Make the callables accept a set of kwargs and args that are well-defined.
 2. [ ] Set up "virtual" grids for them, so that you can specify directly the base size and have it decompose using the existing machinery.

Probably more, but that was the main thing at the moment.

matthewturk:
Ok, with a little more experimentation, I've come up with this as a workable thing:

```python
import yt
from yt.utilities.decompose import get_psize, decompose_array
import numpy as np
import h5py

with h5py.File("example.h5", "w") as f:
    f1, f2, f3 = np.mgrid[0.0:1.0:256j, 0.0:1.0:256j, 0.0:1.0:256j]
    f.create_dataset("/density", data = f1)
    f.create_dataset("/temperature", data = f2)
    f.create_dataset("/dinosaurs", data = f3)

class WeirdFunGenerator:
    def __init__(self, filename):
        self.filename = filename
        self._handle = h5py.File(filename, "r")
        
    def read_data(self, grid, field_name):
        ftype, fname = field_name
        si = grid.get_global_startindex()
        ei = si + grid.ActiveDimensions
        return self._handle[fname][si[0]:ei[0], si[1]:ei[1], si[2]:ei[2]]

wfg = WeirdFunGenerator("example.h5")

grid_data = []
psize = get_psize(np.array((256, 256, 256)), 16)
left_edges, right_edges, shapes, slices = decompose_array((256, 256, 256), psize, np.array([[0.0, 1.0], [0.0, 1.0], [0.0, 1.0]]))
for le, re, s in zip(left_edges, right_edges, shapes):
    grid_data.append({'left_edge': le, 'right_edge': re, 'dimensions': s, 'density': wfg.read_data, 'temperature': wfg.read_data, 'dinosaurs': wfg.read_data, 'level': 1})
ds = yt.load_amr_grids(grid_data, [256, 256, 256])

ds.r[:,0.5,:].plot([("gas", "density")])
```

which does what we want!

image

I think if we get docs and a wrapper function that handles all the decomp, we should be good to go.  The test-case example we could use is a very simple mmap-ing of an on-disk gigantic array.

matthewturk:
With a little bit more work, I made this:

https://gist.github.com/84911dce25e6aac8d6ece9a3d6dbe668

which I think might be getting close to what I'd be happy with.

matthewturk:
I think that I'd like to get two loaders into `loaders.py` before this is ready -- one that does the hdf5 thing and one that accepts a dict of slice-receiving set of objects.

For the latter, the example that I was thinking of was going to be just a set of shaped grids that return their xyz coordinates, to demonstrate it.  This would also work for the tests.

doc/source/examining/loading_data.rst
```diff
@@ -1788,6 +1788,12 @@ Generic AMR Data
 See :ref:`loading-numpy-array` and
 :func:`~yt.frontends.stream.data_structures.load_amr_grids` for more detail.
 
+.. note::
+
+   It is now possible to load data using only *callable* functions, rather
```
neutrinoceros, line 1793: "callable function" is a pleonasm. Maybe just say "function" here (even if "callable" by itself is the more technically correct term, it's also not as clear to most users)

yt/frontends/stream/definitions.py
neutrinoceros: we could provide a `@shape(...)` decorator that would add a `.shape` attribute to callable field functions so this line would work transparently without change. It would work something like this

```python
@shape(1, 2, 3)
def my_callable_field(...):
    ....

my_callable_field.shape  == (1, 2, 3)
```

that said, if we only really car about dimensionality, then `len(data[field].shape)`, then we should just call `data[field].ndim` and have a corresponding, simpler decorator instead.

```python
@ndim(3)
def my_callable_field(...):
    ...
```

yt/frontends/stream/io.py
neutrinoceros: it's typing madness, but I'd say that however inelegant, it's at least very compact and doesn't leak out of scope, so I think it's a reasonable approach overall

neutrinoceros (commented):
here's some feedback

doc/source/examining/loading_data.rst
matthewturk: uh, just in case anybody besides me needs more information:

> ple·o·nasm
> /ˈplēəˌnazəm/
> [Learn to pronounce](https://www.google.com/search?sxsrf=ALiCzsZNFQ8TA2ZODgoV5SFrg0muHxaIrw:1659623425663&q=how+to+pronounce+pleonasm&stick=H4sIAAAAAAAAAOMIfcRoxS3w8sc9YSnDSWtOXmPU5uINKMrPK81LzkwsyczPExLhYglJLcoV4pHi4uIoyEnNz0sszrViUWJKzeNZxCqZkV-uUJKvUADUkw_UlKoAUwIA94fwUlsAAAA&pron_lang=en&pron_country=us&sa=X&ved=2ahUKEwjDj_2Fs635AhW8IDQIHWvjBKgQ3eEDegQIBRAK)
> noun
> the use of more words than are necessary to convey meaning (e.g. see with one's eyes ), either as a fault of style or for emphasis.

neutrinoceros: oops, sorry, again a case of words-that-are-much-more-common-in-french 🙈

yt/frontends/stream/definitions.py
matthewturk: Great idea.  I personally think an `ndim` would be minimum ... I think a more expansive one could come later.

yt/frontends/stream/io.py
matthewturk: OK

matthewturk:
I have two more things to do:

 - add an even simpler slicer loader, that we can feed `mmap` data into even if it's not hdf5
 - really shore up some documentation and examples
 - add some tests for this, but likely not hdf5 loader tests (yet)
 - put in stubs for ndim, but probably not implement in this situation yet

I'd appreciate some feedback on the existing function I added for HDF5 files.

matthewturk:
I suspect the docs build is related to my stubbing of non-existent files etc.

matthewturk:
I made a notebook that I think I might start with for showing off the way you can load data from functional forms.  I also intend to add on an example that just loads from individual files, too -- maybe replicating the enzo data loading.  I got a little confused about the best way to show overlapping wavemodes without doing any whole-domain convolution and I'm not sure I like this all that much ... any suggestions for a better way to demonstrate?

https://gist.github.com/82e13dfab8a2f0d956cce791cddb9611

(n.b., this has no narration)

doc/source/examining/loading_data.rst
```diff
@@ -62,6 +62,70 @@ will be much faster.
...
+
+.. code-block:: python
+
+   import yt
+   ds = yt.load_hdf5_file("UnigridData/turb_vels.h5")
+
+At this point, we now have a dataset that we can do all of our normal
+operations on, and all of the known yt derived fields will be available.
+
+.. code-block:: python
```
neutrinoceros, line 127: @matthewturk I think that's where blacken-docs crashes: you added an empty code block here

matthewturk:
I've spent some time looking at the slicing stuff and I think I want to hold off for the time being.  I'll check that one off and we can consider doing it at a later time.  I'll get the tests added for the HDF5 file.

neutrinoceros:
Taken your last message I'm going to assume it's safe to remove this from the milestone

matthewturk:
Sorry, no, what I meant was: the slices don't need to go in.  I think I am ready for this to go in.  I want to add tests for the existing functionality, but that's all.

yt/loaders.py
```diff
@@ -1561,3 +1561,107 @@ def del_callback(self):
...
+        # We apply a pretty simple heuristic here.  We don't want more than
+        # about 64^3 zones per chunk.  So ...
+        full_size = np.prod(shape)
+        nchunks = full_size // (64**3)
+        mylog.info("Auto-guessing %s chunks from a size of %s", nchunks, full_size)
+    grid_data = []
+    psize = get_psize(np.array(shape), nchunks)
+    left_edges, right_edges, shapes, _ = decompose_array(shape, psize, bbox)
+    for le, re, s in zip(left_edges, right_edges, shapes):
+        data = {_: reader for _ in fields}
```
chrishavlin, line 1664: this is dependent on on my question/comment in `process_data` on including units... but if `process_data` were modified to allow `(callable, units)` tuples and if there was a way of getting units info passed in to this function, you could modify this line to pass that unit info along. For passing in units info -- we could allow the `fields` argument here to be a list of `(fieldname string, units string)` tuples?

yt/loaders.py
```diff
@@ -1561,3 +1561,107 @@ def del_callback(self):
...
 
     return ds
+
+
+def load_hdf5_file(
+    fn: Union[str, "os.PathLike[str]"],
+    root_node: Optional[str] = "/",
+    fields: Optional[List[str]] = None,
+    bbox: np.ndarray = None,
+    nchunks: int = 0,
```
chrishavlin, line 1571: does nchunks=0 have meaning beyond being a flag to trigger the auto-chunking? I'm inclined to change this to 

```suggestion
    nchunks: Optional[int] = None,
```
and change the `nchunks` logic below to match. But I don't feel strongly about it.

chrishavlin (commented):
Took a quick look at the current state since the hdf5 loader was added. In addition to the in-line questions/comments, a couple more:

- the `load_amr_grids` docstring entry for `grid_data` should be updated to reflect that you can supply callables
- do you intend this to work fully with  `load_uniform_grid` as well? It should work as it is now when `nproc=1`, but it'll break when `nproc>1`. I think the domain decomposition when `nproc > 1` could be rewritten to use the `domain_dimensions` argument along with how you handled the `nchunks` in `load_hdf5_file`? but I'm not sure that's needed for this PR.

doc/source/examining/Loading_Data_via_Functions.ipynb
```diff
@@ -0,0 +1,279 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "35cb9c22-04eb-476c-a7b6-afd27690be4d",
+   "metadata": {},
+   "source": [
+    "# Loading Data via Functions\n",
+    "\n",
+    "One of the things that yt provides is the notion of \"derived fields.\"  These are fields that exist only in memory, and are defined by a functional form, typically transforming other fields.  Starting with version 4.1 of yt, however, we have made it much, much easier to define so-called \"on-disk\" fields through their functional forms, akin to how derived fields are generated.\n",
```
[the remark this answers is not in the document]
matthewturk: I think we should hold off -- I'll add a mention.

matthewturk:
I've updated *most* of the suggestions, but I am running out of gas on this for the moment.  I think the biggest item of what is left from the suggestions is to throw an error if `nproc > 1` and there are any callables.
 
@chrishavlin @neutrinoceros take a look?

chrishavlin (dismissed):
new tests look good to me!

In addition to adding the `nproc` error check, I think it'd be good to update the load_hdf5 docstring to mention that it's for loading grids (for now at least? unless I'm confused...).

neutrinoceros:
No further comments for me at this point, thank you for adding tests !

matthewturk:
I think I've addressed them all!  I also added unit stuff like @chrishavlin suggested.

neutrinoceros:
I am very confused by the type checking error. It hardly seems related to anything we're doing here. What seems to be happening is that we're running mypy, targeting Python 3.7, but using numpy 1.23, which uses some Python 3.8+ syntax, which raises an error and prevent type checking to be triggered. This seems like something that should happen every time the job is run, or not at all, but I've never seen it happening before.

neutrinoceros:
Before we do anything else, I suggest upgrading mypy, hoping that the issue (that I'm assuming is internal to mypy) was resolved already. I'm working on this in #4139

neutrinoceros (approved):

jzuhone (approved):

PR #3436, opened 2021-07-14
FIX: fix a bug where region slicing with all-none-empty slices would break for 2D datasets

### LINKED ISSUE #3429: Grid bug
### Bug report

**Bug summary**

When slicing and specifying the amount of grid points to interpolate onto, I get a dimensionality error. The data sets are 2d, giving it a 2D slice, YT says it expects a 3D slice. Giving it a 3d slice, YT says it expects a 2D slice.

I used the sample Castro data set (Sedov problem, 2-d dataset, 4 levels) to reproduce this problem. It might be related to my previous issue but is different enough I thought it'd be best to make 2 issues.
 
**Code for reproduction**

```python
ds = yt.load('castro_sedov_2d_cyl_in_cart_plt00150')
flame_slice = ds.r[.1:.2:8j, :] #can slice in x (this works)
flame_slice = ds.r[:,1:2:8j] #can slice in y (this works)
flame_slice = ds.r[.1:.2:8j, 1:2:8j] # this doesn't work 
```
output: RuntimeError: Length of dims must match the dimensionality of the dataset
It's checking `if len(dims) != len(self.ds.domain_left_edge)` where 

```python
ds = yt.load('castro_sedov_2d_cyl_in_cart_plt00150')
print(len(ds.domain_left_edge)) # which is 3
print(ds.dimensionality) #But this is 2
```

Trying to fix the above error by giving it 3 dimensions instead of 2:
```python
ds = yt.load('castro_sedov_2d_cyl_in_cart_plt00150')
flame_slice = ds.r[.1:.2:3j, 1:2:3j, :]
```
It now has: 
`YTDimensionalityError: Dimensionality specified was 3 but we need 2`

**Version Information**

  * Operating System: Mac os
  * Python Version: 3.8.5
  * yt version: 4.0

Installed using pip

neutrinoceros:
## PR Summary

Fix #3429. Hopefully doesn't break anything else.


neutrinoceros:
switching to draft until I manage to fix the bug without causing other failures

yt/data_objects/region_expression.py
```diff
@@ -152,14 +152,17 @@ def _slice_to_edges(self, ax, val):
     def _create_region(self, bounds_tuple):
         left_edge = self.ds.domain_left_edge.copy()
         right_edge = self.ds.domain_right_edge.copy()
-        dims = []
+        dims = [1, 1, 1]
```
neutrinoceros, line 155: note to reviewers: everything I do here, besides forcing dims to have 3 elements even in reduced dimensionality, isn't _necessary_, it's just small improvements I wrote while trying to fix the issue and I figured I might as well keep them.

neutrinoceros:
Note to reviewers:
An alternative to this patch we be to make the `.r` accessor consistently return YTRegion objects, and never YTArbitraryGrid as is currently special cased when slices are completely specified. I think this would make the api more predictable and more robust, however I expect it would be a pretty disruptive change so it could only be implemented through a clear deprecation cycle.

matthewturk:
@neutrinoceros I actually don't like the change you're proposing to `.r` but it should indeed be discussed.

PR #3438, opened 2021-07-15
UX: add a clear error message in case a DerivedField is passed a lambda instead of a function

neutrinoceros:
## PR Summary

closes #3434
Since I can't find a way to add support for lambdas in derived fields, I propose to explicitly forbid them for now.
This leaves room for a future contribution to overcome this limitation, and it makes the situation a lot clearer for anyone attempting this in the mean time.

Issue #3434, opened 2021-07-13
Creating derived fields with lambda functions


forrestglines:
### Bug report

**Bug summary**

Adding fields using a lambda function fails to add the field to a dataset. Trying to access created fields that use a lambda function results in a `YTFieldNotFound` exception.

**Code for reproduction**

```python
import yt

ds = yt.testing.fake_amr_ds(fields=["density"],units=["g/cm**3"])

for i in range(3):
    ds.add_field(
            name=("gas", f"field_{i}"),
            function=lambda field,data : i*data["density"],
            sampling_type="local",
            units="g/cm**3"
            )

for i in range(3):
    name = ("gas",f"field_{i}")
    print(f"{name} in ds.field_info:", name in ds.field_info) #All false
```

**Actual outcome**
```
yt : [INFO     ] 2021-07-13 16:50:44,250 Parameters: current_time              = 0.0
yt : [INFO     ] 2021-07-13 16:50:44,250 Parameters: domain_dimensions         = [32 32 32]
yt : [INFO     ] 2021-07-13 16:50:44,250 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2021-07-13 16:50:44,250 Parameters: domain_right_edge         = [1. 1. 1.]
yt : [INFO     ] 2021-07-13 16:50:44,250 Parameters: cosmological_simulation   = 0.0
('gas', 'field_0') in ds.field_info: False
('gas', 'field_1') in ds.field_info: False
('gas', 'field_2') in ds.field_info: False
```

None of the fields defined with lambda functions get added to `field_info`. No errors or exceptions are reported by `add_field`. But when I try to access the supposedly created fields, I get a `YTFieldNotFound` exception.

**Expected outcome**

Lambda functions should work with `add_field` like other function objects. Or it should be documented that they don't work.

I remember running into this issue once before several years ago when I was just starting out with `yt`, it might have existed for a while.

**Version Information**

  * Operating System: Linux
  * Python Version: Python 3.9.5
  * yt version: Fork of `4.0.dev0`, installed from source
  * Other Libraries (if applicable):

matthewturk:
Hi @forrestglines -- can you verify that if the field is defined as a function (potentially in its own closure), it does get added?

forrestglines:
Yes - I tried out a few ways of adding the fields
```python
#Add the fields with functions defined in loop - works
for i in range(3,6):
    def _field_i(field,data):
[...]

PR #3440, opened 2021-07-15
FIX: add support for setting up derived fields from lambdas

neutrinoceros:
## PR Summary
This is an alternative fix to #3434, and supersedes #3438.
The solution tried here was suggested by @forrestglines. It's not clear to me why the problematic check was ever necessary but it's
been here for 8yrs so my hypothesis is that it may have become inappropriate in Python 3 ?
I'm doubting this would break anything since there doesn't seem to be a single use case of lambdas in derived fields in the code base/docs/tests. On the other hand, the test I'm adding here fails on the main branch (as reported in #3434).

### DIFF
--- a/yt/fields/derived_field.py
+++ b/yt/fields/derived_field.py
@@ -250,10 +250,7 @@ class DerivedField:
         This returns a list of names of fields that this field depends on.
         """
         e = FieldDetector(*args, **kwargs)
-        if self._function.__name__ == "<lambda>":
-            e.requested.append(self.name)
-        else:
-            e[self.name]
+        e[self.name]
         return e
 
     def _get_needed_parameters(self, fd):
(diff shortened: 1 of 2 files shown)


yt/fields/derived_field.py
```diff
@@ -250,10 +250,7 @@ def get_dependencies(self, *args, **kwargs):
         This returns a list of names of fields that this field depends on.
         """
         e = FieldDetector(*args, **kwargs)
-        if self._function.__name__ == "<lambda>":
```
matthewturk, line 253: So I've done some thinking, and I *think* this comes from waaaay back in the day when we may have used lambdas to represent on-disk fields.

matthewturk (approved):
I think this is good.

neutrinoceros:
Thanks Matt, now I'm much more confident in calling this a plain bug fix :)

jzuhone (approved):

neutrinoceros:
@jzuhone , @matthewturk , anyone in particular that you would like to take a look a this too or is it good for merging ?

PR #3441, opened 2021-07-16
ENH: migrate native colormaps out of yt and use cmyt as a dependency (YTEP-0040)

neutrinoceros:
## PR Summary

Following the acceptance + merge of [YTEP-0040](https://github.com/yt-project/ytep/pull/20), I just published [cmyt 0.1.1](https://pypi.org/project/cmyt/).

This PR makes cmyt a dependency to yt, cleans up the duplicated code, and add a backward compatibility layer to it so this should not create _any_ disruption downstream.

I'm still updating documentation and tests to use the verbose and _stable_ api (using the `"cmyt."` prefix) to use colormaps from cmyt, because this is the only method that we can guarantee will not stop working when, e.g., matplotlib decides to ship a new colormap under a name we happen to use as well.

I'll consider this a draft for now, having CI passing here is a necessary condition to bump cmyt's version to 1.0.0 (this is https://github.com/yt-project/cmyt/issues/12)

TODO:
- [x] pass CI
- [x] figure out what happened to the "spectral" colormap
- [x] double check that documentation was properly updated
- [x] update doc/source/visualizing/colormaps/index.rst
- [x] rebase


neutrinoceros:
@yt-fido test this please

neutrinoceros:
about the "spectral" colormap: I believe it was always a MPL native colormap, but was renamed in MPL 2.0.0
https://matplotlib.org/devdocs/api/prev_api_changes/api_changes_2.0.0.html?highlight=nipy_spectral#the-spectral-colormap-is-now-nipy-spectral

yt/visualization/_colormap_data.py
```diff
@@ -8440,3 +7470,16 @@
 for k,v in color_map_luts.items():
     temp[k+"_r"] = (v[0][::-1], v[1][::-1], v[2][::-1], v[3][::-1])
 color_map_luts.update(temp)
+
+# YTEP-0040 backward compatibility layer
```
neutrinoceros, line 7474: Note to reviewers: I'm highlighting this since I seems that the web viewer is making it possibly harder to discover.

neutrinoceros:
update: this migration is turning out a little more subtle than I anticipated. I think the most problematic aspect is the fact that we have a shared dict of "luts" in `yt.visualization._colormap_data.color_map_luts` that contains a lot of hard-coded data (mostly from matplotlib actually) and I can't understand why atm. This dict is imported in exactly 2 places, and I'm _pretty_ sure we could refactor them to just save the need for it, but I'd like to get some more insight before I attempt this. @matthewturk , if you have any memory of this, it would help a lot :)

matthewturk:
What hard-coded data does it contain that is duplicated?

neutrinoceros:
@matthewturk for instance, see https://github.com/yt-project/yt/blob/c5f6641662659ff7c937803d0c5f7a4c1914e225/yt/visualization/_colormap_data.py#L10
This is the lut for the matplotlib-native colormap RdBu, but it can be obtained at runtime as
```python
RdBu = plt.get_cmap("RdBu")
RdBu._init()
lut = RdBu._lut
```
I'm now experimenting with this in #3442

matthewturk:
I genuinely don't remember why RdBu was included; I know at one point we grabbed a couple so we'd have access to them without importing matplotlib (i.e., in cases where the import was fragile -- for instance, on HPC systems back when X11 not being present could sometimes cause a hard abort) but RdBu wasn't one of those.

neutrinoceros:
@matthewturk I'm sure this was done for historically good reason but I'm willing to bet they don't apply any more. Btw cleaning up RdBu and other matplotlib-native cmap is actually part of @3442, which is ready for review, as opposed to this PR.

a-gilbert:
Dropping in to say that building from the repo Linux Mint 19.3 (Ubuntu derivative) with mpl 3.4.2 still produces this issue. Just wanted to check what was still left to be done.

neutrinoceros:
@a-gilbert I'm sorry, could you elaborate ? What issue are you talking about ? This pull request isn't supposed to fix an ongoing situation.

a-gilbert:
@neutrinoceros Sure! When trying to import yt I currently get `ValueError: Trying to re-register the builtin cmap 'cubehelix'.` On the hunt for a fix I came across your summary in issue #3165 and was under the impression that this proposal was the plan to fix the problem.

neutrinoceros:
It's related, for sure, but the case of cubehelix was dealt with earlier. Note that yt is explicitly declared as incompatible with matplotlib 3.4.2 (because of an undocumented change that was later reversed in 3.4.3), so I would expect building yt against this specific version to be impossible actually. In any case, if your problem persists, could you please open a dedicated issue for it ? Thanks

neutrinoceros:
I'd like #3442 to go in first, this is the reason why this PR is still in draft mode

cphyc (approved):

matthewturk:
I have no problem with this -- do we need to socialize the change or do anything else before it can go in?  Is cmyt released?

matthewturk (approved):
with the back-compat layer you've added, this looks good to me

neutrinoceros:
Just rebased to fix a conflict in`tests/windows_conda_requirements.txt`, apart from that the diff is identical (although it now fits in a single commit).

> do we need to socialize the change or do anything else before it can go in?

we should announce this on the dev mailing list at least because local git installs will break after this goes in (the fix being to install cmyt locally)

> Is cmyt released?

Yes. I consider it beta for now but plan to make it officially stable (1.0) after the present PR goes in.
Currently cmyt is only available on PyPI and GitHub but I also want to publish the upcoming 1.0 release to condo-forge

matthewturk:
I've merged the answer store request.

neutrinoceros:
@matthewturk wait I think you're mistaking this for #3637

matthewturk:
you're right

neutrinoceros:
Announcement done. I don’t think there’s any reason to wait on anything to get this in now, is there?

matthewturk (approved):

PR #3442, opened 2021-07-18
YTEP-0040: prep work to decouple from hardcoded duplicated data"

neutrinoceros:
## PR Summary
This is an experiment. My end goal is to decouple 
yt.visualization._colormap_data.py from the rest of the code base as much as possible,
which I believe will make #3441 both cleaner and easier to manage.

If the first commit pass CI, I'll remove a bunch of luts that are hardcoded in this module,
which are likely unnecessary (now) since they are just duplicated from matplotlib.
For the record I believe they _used_ to be necessary to make upcoming colormaps from matplotlib available
to yt users running older versions, but from what I can grasp it has fell out of necessity now that our earliest support MPL version is 2.0.2


yt/visualization/color_maps.py
```diff
@@ -186,8 +191,16 @@ def add_colormap(name, cdict):
         add_colormap(k, cdict)
 
 
-def _extract_lookup_table(cmap_name):
-    cmap = mcm.get_cmap(cmap_name)
+def get_colormap_lut(cmap_id: Union[Tuple[str, str], str]):
```
cphyc, line 194: I'd also suggest adding a docstring to this function, esp. because `lut` is a bit cryptic.

```suggestion
def get_colormap_lut(cmap_id: Union[Tuple[str, str], str]) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
```

cphyc (commented):
I _think_ I am in favour of this PR, but I have to admit I am a bit lost in this part of the code.

yt/visualization/color_maps.py
neutrinoceros: I'm not 100% sure adding such type hints really helps atm, in particular because there is no way to express what shapes to expect, or the relationships between these arrays' respective shapes (but I expect it will be supported by numpy in the future).

neutrinoceros: as for the docstring, I need to admit I don't know what to write.
"lut" in the context means "lookup table", but the only reason this function exists is historical (I think). Do you think adding a similar comment would be valuable ?

cphyc: At least so that one knows what `lut` stands for. At the moment, it is still unclear to me what the function actually achieves.

neutrinoceros: added a comment

matthewturk:
@neutrinoceros can you resolve the conflicts?

neutrinoceros:
done

matthewturk:
@neutrinoceros pre-commit seems red -- did I do that?

PR #3443, opened 2021-07-19
ENH: Directly open compressed files

cphyc:
## PR Summary

This PR allows the user to directly load a dataset stored in a .tar.gz file (or any foramt supported by ratarmount).
It works by first mounting the .tar file as a read-only filesystem, which can then be used directly in yt.

## PR Checklist

## Example
```bash
# Create .tar file
tar czf output_00080.tar.gz output_00080
```
Now, one can load the dataset in the archive directly using
```python
import yt
ds = yt.load_archive("output_00080.tar.gz", "output_00080")
```


yt/loaders.py
```diff
@@ -1457,3 +1461,46 @@ def load_sample(
         loadable_path = loadable_path.joinpath(load_name, specific_file)
 
     return load(loadable_path, **kwargs)
+
+
+# --- Loader for tar-based datasets ---
+def load_archive(
+    fn: str, path: str, ratarmount_kwa: dict = None, *args, **kwargs
```
neutrinoceros, line 1468: trying to have this type check
```suggestion
    fn: str, path: str, ratarmount_kwa: Optional[Dict] = None, *args, **kwargs
```

yt/loaders.py
```diff
@@ -1457,3 +1461,46 @@ def load_sample(
...
+) -> Dataset:
+    if ratarmount_kwa is None:
+        ratarmount_kwa = {}
+
+    from yt.utilities.on_demand_imports import _ratarmount
```
neutrinoceros, line 1473: the way on_demand_imports work, it doesn't cost anything to put this at the top level

yt/loaders.py
```diff
@@ -1457,3 +1463,108 @@ def load_sample(
...
+
+    - Any additional keyword argument is passed down to :func:`~yt.loaders.load`.
+    - This function requires ratarmount.
+    - This function does not work on Windows system.
+    """
+    if platform.system() not in ("Linux", "Darwin"):
+        raise RuntimeError(
+            "Reading from an archive is not supported on your operating system"
+            f" ({platform.system()})."
+        )
```
neutrinoceros, line 1505: I assume this limitation is really coming from ratarmount, in which case we shouldn't add another barrier on top of it, in case a future version has support for windows. I would suggest to wrap the first call to ratarmount that may break in a try statement and possibly raise a custom error message (as you're doing here), but be careful about using a conditional (if it fails, it's _probably_ because of platform support, but we don't know it for a fact).

yt/loaders.py
```diff
@@ -1457,3 +1463,108 @@ def load_sample(
...
+
+    if isinstance(fn, Path):
+        fn = str(fn.resolve().absolute())
```
neutrinoceros, line 1508: this is the canonical way to support `os.PathLike` objects as transparently as possible (i.e. without using isinstance)
```suggestion
    fn = os.path.realpath(fn)
```

yt/loaders.py
```diff
@@ -1457,3 +1463,108 @@ def load_sample(
...
+        )
+
+    proc = Process(target=mount, args=(fn, tempdir, ratarmount_kwa))
+    proc.start()
+    # Note: the mounting needs to happen in another process which
+    # needs be run in the foreground (otherwise it may
+    # unmount). To prevent a race-condition here, we wait
+    # for the folder to be mounted within a reasonable time.
+    retry = 0
+    while retry < 100:
```
neutrinoceros, line 1548: up to 10s of looping in vain seems a bit excessive to me. 10 attempt should be more than enough, shouldn't it ?

neutrinoceros (commented):
here is a first batch of comments

yt/tests/test_load_archive.py
```diff
@@ -0,0 +1,77 @@
...
+    time.sleep(2)
+
+    ## Mount path should not exist anymore *and* have been deleted
+    assert not mount_path.is_mount()
+    assert not mount_path.exists()
+
+
+@pytest.mark.skipif(
+    not sys.platform.startswith("win"),
+    reason="Should not work on Windows",
```
neutrinoceros, line 73: please document the version of ratarmount you last checked this with

cphyc:
For reference, here is a simplistic benchmark:

| Access | Time (s)   |
|-------|--------------|
| Direct | 1.75 |
| Through .tar.gz | 5.57 |
| Through .tar | 2.03 |

This uses the following code:
```ipython
>>>  %%timeit
... ds = yt.load("output_00080")
... ad = ds.all_data()
... ad["density"]
>>> %%timeit
... ds = yt.load_archive("output_00080.tar.gz", "output_00080")
... ad = ds.all_data()
... ad["density"]
>>> %%timeit
... ds = yt.load_archive("output_00080.tar", "output_00080")
... ad = ds.all_data()
... ad["density"]
```
Note that the first access is much slower as the entire tar file has to be read from disk.

.github/workflows/build-test-pytest.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
 
   build:
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
neutrinoceros: I don't think it is useful enough to justify the carbon footprint it would generate. IMO we shouldn't keep it, but I'm fine with using it punctually if it helps you debugging your PR :)

yt/tests/test_load_archive.py
```diff
@@ -0,0 +1,99 @@
...
+    ytcfg.set("yt", "test_data_dir", str(tmp_path))
+
+    yield tmp_path
+
+    ytcfg.set("yt", "test_data_dir", pre_test_data_dir)
+
+
+# Note: ratarmount cannot currently be installed on Windows as of v0.8.1
+@requires_module_pytest("pooch")
+@requires_module_pytest("ratarmount")
```
neutrinoceros, line 30: This decorator adds a supplement of magic over `requires_module`, it is implemented to provide more complete errors :)
```suggestion
@requires_module_pytest("pooch", "ratarmount")
```

yt/tests/test_load_archive.py
```diff
@@ -0,0 +1,99 @@
...
+def tmp_data_dir(tmp_path):
+    pre_test_data_dir = ytcfg["yt", "test_data_dir"]
+    ytcfg.set("yt", "test_data_dir", str(tmp_path))
+
+    yield tmp_path
+
+    ytcfg.set("yt", "test_data_dir", pre_test_data_dir)
+
+
+# Note: ratarmount cannot currently be installed on Windows as of v0.8.1
```
neutrinoceros, line 28: why isn't this a `@pytest.mark.skipif` ?

cphyc: That's probably a better option, since ratarmount _can_ be installed on Windows but won't work.

cphyc:
pre-commit.ci autofix

cphyc:
I don't really understand why it fails on Jenkins but it's really hard to find out without having direct access to it (it just doesn't seem to mount properly). I'd be happy to somehow mark to feature as experimental for the time being and see what happens with users on real cases. Wdyt @munkm @matthewturk @neutrinoceros?

matthewturk:
That plan sounds good to me.

neutrinoceros (dismissed):

neutrinoceros:
I see a minor flake8 issue so disabling auto-merge while I fix that.

neutrinoceros (dismissed):

neutrinoceros:
now mypy is unhappy (I think this branch is older than type-checking so it's not too surprising). @cphyc, do you want to tackle these nits yourself ? Otherwise I'm happy to step in.

neutrinoceros (approved):

PR #3445, opened 2021-07-20
Fix for 2D AMReX datasets in which 3 values are in the amr.n_cell entry in the job_info file.

atmyers:
The WarpX team has noticed that since the yt-4.0 release, some of our 2D CI jobs have been failing. To reproduce this, see the attached script and dataset. `ds.domain_dimensions` should be [64, 64, 1], but it is now shown as [64, 64, 64].  

I think this was introduced in [this](https://github.com/yt-project/yt/commit/8bcf84dbd938fc32f87f8ae476e583509866f74f#diff-d4ca70f009bc419366a8305237dfe5143869b6c6a781c363f4c4a0082f239a99) commit, which made changes to how 1d and 2d amrex datasets are set up. The issue is that we sometimes use inputs files that work for both 2D and 3D simulations. Hence, the `amr.n_cell` entry to `warpx_job_info` may have three values in it, even if the dataset is 2D. 

This PR seems to solve the issue for us. The fix was just to use `ds.dimensionality` to cap the number of entries used from the `job_info` file. 

```
import yt
ds = yt.load("diag100001")
print(ds.domain_dimensions)  # should be [64, 64, 1]
```

[diag100001.tar.gz](https://github.com/yt-project/yt/files/6850781/diag100001.tar.gz)

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -138,7 +138,7 @@ answer_tests:
     - yt/frontends/boxlib/tests/test_outputs.py:test_units_override
     - yt/frontends/boxlib/tests/test_outputs.py:test_raw_fields
 
-  local_boxlib_particles_008:
+  local_boxlib_particles_009:
     - yt/frontends/boxlib/tests/test_outputs.py:test_LyA
     - yt/frontends/boxlib/tests/test_outputs.py:test_nyx_particle_io
     - yt/frontends/boxlib/tests/test_outputs.py:test_castro_particle_io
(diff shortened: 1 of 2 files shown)


atmyers:
The test that is failing is another 2D dataset where on `main` the `domain_dimensions` are incorrectly set to 3D - is it possible that the answers to the test were bumped by mistake?

neutrinoceros:
Hi Andrews ! Sorry for authoring the culprit commit, and thank you for catching it !
I wasn't aware of this subtlety in AMReX, but the fix makes sense to me.

> is it possible that the answers to the test were bumped by mistake?

I don't think so, or at least I cannot understand how that would have happened atm. I think that what we're seeing is in fact that these specific answers were _never_ correct in the first place.

neutrinoceros (approved):
I'll formally approve of this patch, and I bumping the answers as well

atmyers:
Thank you for looking at this so quickly! I have bumped the answers.

PR #3446, opened 2021-07-20
BUG: fix a broken error message for yt.load in case of FileNotFoundError

neutrinoceros:
## PR Summary

reproduction: assuming `ytcfg["yt", "test_data_dir"]` is correctly setup to any existing folder:
```python
import yt
yt.load("~/Downloads/nope")
```

on main:
```
FileNotFoundError: No such file or directory: '/Users/robcleme/Downloads/nope'.
(Also tried '/Users/robcleme/Downloads/diag100001').
```
here:
```
FileNotFoundError: No such file or directory: '/Users/robcleme/Downloads/nope'`.
```

The bug was that a `pathlib.Path` instance cannot be joined with a an absolute path (the result is the absolute path itself).
I added a conditional to avoid this edge case.


neutrinoceros:
@jzuhone I assume you removed the draft status by mistake, but this is now _really_ ready for review :)

yt/sample_data/api.py
```diff
@@ -143,7 +143,7 @@ def _get_test_data_dir_path():
     return Path.cwd()
 
 
-def lookup_on_disk_data(fn):
+def lookup_on_disk_data(fn) -> Path:
```
cphyc, line 146:
```suggestion
from typing import Union
def lookup_on_disk_data(fn: Union[str, Path]) -> Path:
```

cphyc (approved):

yt/sample_data/api.py
neutrinoceros: I did not add a type here on purpose, become "filelike" is a little more complicated than this and I'd prefer adding a consistent `FileLike` type alias later at the level of the code base (see #3064)

chrishavlin (approved):
looks good!

yt/sample_data/api.py
neutrinoceros: @cphyc , if you're contempt with my response here I think it's ready to go, otherwise let's discuss it further ?

cphyc: That's fine! Mergin'

PR #3455, opened 2021-07-27
Make sure we add the Parameters group to hvals if it's present

jzuhone:
## PR Summary

Gadget 4 datasets have a `Parameters` group in the HDF5 file which store runtime parameters. Crucially, the parameters needed to determine the cosmology are now stored in this group and are not in the `Header` group. This PR ensures that these parameters can be found regardless of the Gadget version used. 

## PR Checklist


neutrinoceros (approved):
I confirm that this fixes the issue addressed, thanks John !

PR #3456, opened 2021-07-27
fix combining FITS images

jzuhone:
…ed as the PrimaryHDU

## PR Summary

The `combine_images` method of the `FITSImageData` class allows one to combine multiple images into one, but a FITS file is only valid if the first image is a `PrimaryHDU`. The previous implementation, when stitching multiple `FITSImageData` instances together, was attempting to create multiple `PrimaryHDU` instances (the first image from each instance). This PR fixes it and adds a single line to the `FITSImageData` tests which ensure that this functionality works correctly.

## PR Checklist

### DIFF
--- a/yt/visualization/fits_image.py
+++ b/yt/visualization/fits_image.py
@@ -735,12 +735,12 @@ class FITSImageData:
         w = first_image.wcs
         img_shape = first_image.shape
         data = []
-        for is_first, _is_last, fid in mark_ends(image_list):
+        for fid in image_list:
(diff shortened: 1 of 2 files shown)


neutrinoceros (changes requested):
I don't think the test correctly validates the change: I've tried running the modified test function _without_ the patch in `fits_image.py` and it still passed.

neutrinoceros (approved):
Much better ! the new test correctly fails on the main branch and pass here, confirming the bug and the fix. Thanks John !

Issue #3449, opened 2021-07-21
Bug: ProjectionPlot line annotation modifies array


claytonstrawn:
### Bug report

When I use annotate_line, it modifies the original array of startpoints and endpoints, and doesn't keep a consistent size. It has something to do with "unitary" from the UniformGridData unit_registry

**Code for reproduction**

```python
import yt
yt.funcs.mylog.setLevel(50)

data_array = np.random.random((10,10,10))
bbox = np.array([[-100,100],[-100,100],[-100,100]])
data = {('gas','density'):(data_array, 'g*cm**(-3)')}
ds = yt.load_uniform_grid(data, data_array.shape, length_unit="kpc", bbox=bbox)

start_source = np.array((0,0,-0.5))
end_source =  np.array((0,0,0.5))
start = ds.arr(start_source,'unitary')
end = ds.arr(end_source,'unitary')

p = yt.ProjectionPlot(ds,0,'number_density')
p.annotate_line(start,end)
print('\n before')
print("start",start,type(start))
print("end",end,type(end))
print("start_source",start_source,type(start_source))
print("end_source",end_source,type(end_source))
p.show()
print('\n after')
print("start",start,type(start))
print("end",end,type(end))
print("start_source",start_source,type(start_source))
print("end_source",end_source,type(end_source))

start = ds.arr(start,'unitary')
end = ds.arr(end,'unitary')
print('\n trying to coerce')
print("start",start,type(start))
print("end",end,type(end))
print("start_source",start_source,type(start_source))
print("end_source",end_source,type(end_source))
```

**Actual outcome**

```
 before
start [ 0.   0.  -0.5] unitary <class 'unyt.array.unyt_array'>
end [0.  0.  0.5] unitary <class 'unyt.array.unyt_array'>
start_source [ 0.   0.  -0.5] <class 'numpy.ndarray'>
end_source [0.  0.  0.5] <class 'numpy.ndarray'>
Image

 after
start [   0.    0. -100.] code_length <class 'unyt.array.unyt_array'>
end [  0.   0. 100.] code_length <class 'unyt.array.unyt_array'>
start_source [   0.    0. -100.] <class 'numpy.ndarray'>
end_source [  0.   0. 100.] <class 'numpy.ndarray'>

 trying to coerce
start [   0.    0. -100.] unitary <class 'unyt.array.unyt_array'>
end [  0.   0. 100.] unitary <class 'unyt.array.unyt_array'>
[...]

PR #3463, opened 2021-08-06
Changed _project_coords to create new array instead of modifying in-place

claytonstrawn:
## PR Summary

When running plot.annotate_line() in a SlicePlot or ProjectionPlot, it modifies the existing array in-place, without correctly using the units in certain cases. When the same line is used again for another plot (say, at a different angle or something) it and its original source array are unexpectedly modified. Fixed by creating a new array converting with `.to` instead of `.convert_to_units`. Since this is used for 2 arrays which are only 3 elements long, the memory added is negligible.

This is the fix to issue #3449 suggested by @chrishavlin 
## PR Checklist

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -78,9 +78,9 @@ class PlotCallback:
         if len(coord) == 3:
             if not isinstance(coord, YTArray):
                 coord_copy = plot.data.ds.arr(coord, "code_length")
-            # coord is being copied so that if the user has a unyt_array already
-            # we don't change the user's version
             else:
+                # coord is being copied so that if the user has a unyt_array already
+                # we don't change the user's version
                 coord_copy = coord.to("code_length")
             ax = plot.data.axis
             # if this is an on-axis projection or slice, then
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -78,7 +78,7 @@ class PlotCallback:
         if len(coord) == 3:
             if not isinstance(coord, YTArray):
                 coord = plot.data.ds.arr(coord, "code_length")
-            coord.convert_to_units("code_length")
+            coord.to("code_length")
             ax = plot.data.axis
             # if this is an on-axis projection or slice, then
             # just grab the appropriate 2 coords for the on-axis view
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -80,7 +80,8 @@ class PlotCallback:
                 coord_copy = plot.data.ds.arr(coord, "code_length")
             #coord is being copied so that if the user has a unyt_array already
             #we don't change the user's version
-            coord_copy = coord.to("code_length")
+            else:
+                coord_copy = coord.to("code_length")
             ax = plot.data.axis
             # if this is an on-axis projection or slice, then
             # just grab the appropriate 2 coords for the on-axis view
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -78,7 +78,7 @@ class PlotCallback:
         if len(coord) == 3:
             if not isinstance(coord, YTArray):
                 coord = plot.data.ds.arr(coord, "code_length")
-            coord.to("code_length")
+            coord = coord.to("code_length")
             ax = plot.data.axis
             # if this is an on-axis projection or slice, then
             # just grab the appropriate 2 coords for the on-axis view
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -78,8 +78,8 @@ class PlotCallback:
         if len(coord) == 3:
             if not isinstance(coord, YTArray):
                 coord_copy = plot.data.ds.arr(coord, "code_length")
-            #coord is being copied so that if the user has a unyt_array already
-            #we don't change the user's version
+            # coord is being copied so that if the user has a unyt_array already
+            # we don't change the user's version
             else:
                 coord_copy = coord.to("code_length")
             ax = plot.data.axis
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -77,8 +77,10 @@ class PlotCallback:
         """
         if len(coord) == 3:
             if not isinstance(coord, YTArray):
-                coord = plot.data.ds.arr(coord, "code_length")
-            coord = coord.to("code_length")
+                coord_copy = plot.data.ds.arr(coord, "code_length")
+            #coord is being copied so that if the user has a unyt_array already
+            #we don't change the user's version
+            coord_copy = coord.to("code_length")
             ax = plot.data.axis
             # if this is an on-axis projection or slice, then
             # just grab the appropriate 2 coords for the on-axis view
@@ -87,7 +89,7 @@ class PlotCallback:
                     plot.data.ds.coordinates.x_axis[ax],
                     plot.data.ds.coordinates.y_axis[ax],
                 )
-                coord = (coord[xi], coord[yi])
+                ret_coord = (coord_copy[xi], coord_copy[yi])
 
             # if this is an off-axis project or slice (ie cutting plane)
             # we have to calculate where the data coords fall in the projected
@@ -95,20 +97,20 @@ class PlotCallback:
             elif ax == 4:
                 # transpose is just to get [[x1,x2,...],[y1,y2,...],[z1,z2,...]]
                 # in the same order as plot.data.center for array arithmetic
-                coord_vectors = coord.transpose() - plot.data.center
+                coord_vectors = coord_copy.transpose() - plot.data.center
                 x = np.dot(coord_vectors, plot.data.orienter.unit_vectors[1])
                 y = np.dot(coord_vectors, plot.data.orienter.unit_vectors[0])
                 # Transpose into image coords. Due to VR being not a
                 # right-handed coord system
-                coord = (y, x)
+                ret_coord = (y, x)
             else:
-                raise ValueError("Object being plot must have a `data.axis` defined")
+                raise ValueError("Object being plotted must have a `data.axis` defined")
 
         # if the position is already two-coords, it is expected to be
         # in the proper projected orientation
         else:
             raise ValueError("'data' coordinates must be 3 dimensions")
-        return coord
+        return ret_coord
 
     def _convert_to_plot(self, plot, coord, offset=True):
         """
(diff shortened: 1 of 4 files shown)


matthewturk (approved):
I'll be honest and note that I did not know that `.to` was different from `convert_to_units`.  I guess I usually use `in_units`.

neutrinoceros (changes requested):
I'm not convinced this change is acceptable as it stands.
```python
coord.convert_to_units("code_length")
```
indeed converts the array in place, while 
```python
coord.to("code_length")
```
creates a new array and returns it, *but* the return value isn't captured, so this line is now a no-op with extra steps, seems to me.

I think what needs to happen for it to work would be to make a clear distinction between the input variable and the output value as two coexisting variables.
Furthermore I note that not a single test failed with this change. Could you please provide either an actual test or at least a usable script to reproduce the situation you're describing in the PR ?

neutrinoceros:
For clarity, `unyt.unyt_quantity.to` is a strict alias to `unyt.unyt_quantity.in_units`, but `unyt.unyt_quantity.convert_to_units` is indeed an entirely different function, though they appear to share a lot of duplicated code (but not enough that refactoring would be trivial).

neutrinoceros:
classifying as bug since it seems extremely unlikely that anyone is relying on the existing behaviour

claytonstrawn:
@neutrinoceros oops! Yes, the line I meant to add was `coord = coord.to("code_length")`, as it was written before it indeed did nothing. That was the one that fixed the problem.

> I think what needs to happen for it to work would be to make a clear distinction between the input variable and the output value as two coexisting variables.
> Furthermore I note that not a single test failed with this change. Could you please provide either an actual test or at least a usable script to reproduce the situation you're describing in the PR ?

I'm not sure I follow. There is a usable script in #3449 which reproduces the problem, which is that when the user is holding on to a unyt array (of a start and end of a ray) and tries to use it multiple times to plot different views of the sightline, the array gets its units mangled and does not appear in the right place. To be honest, I do not understand why `convert_to_units` would cause this problem, however.

neutrinoceros:
> I'm not sure I follow.

essentially what I mean is that I'd like to avoid changing the `coord` input variable and use a different variable for the return value. 

>  There is a usable script in #3449 which reproduces the problem

sorry I missed that, here it is as a test function (unpolished, there's room to improve it but I have to run):
```python
import yt
import numpy as np


def test_ds_arr_invariance_under_projection_plot():
    data_array = np.random.random((10, 10, 10))
    bbox = np.array([[-100, 100], [-100, 100], [-100, 100]])
    data = {("gas", "density"): (data_array, "g*cm**(-3)")}
    ds = yt.load_uniform_grid(data, data_array.shape, length_unit="kpc", bbox=bbox)

    start_source = np.array((0, 0, -0.5))
    end_source = np.array((0, 0, 0.5))
    start = ds.arr(start_source, "unitary")
    end = ds.arr(end_source, "unitary")

    start_i = start.copy()
    end_i = end.copy()

    p = yt.ProjectionPlot(ds, 0, "number_density")
    p.annotate_line(start, end)
    p.show()

    # for lack of a unyt.testing.assert_unit_array_equal function
    np.testing.assert_array_equal(start_i, start)
    assert start_i.units == start
    np.testing.assert_array_equal(end_i, end)
    assert end_i.units == end

```
can you please add it to your PR ?

claytonstrawn:
@neutrinoceros Thanks, I changed the file to not override the input `coord` variable, changing it to `coord_copy` and the return to `ret_coord` (which is a `tuple` instead of a `unyt_array` anyway and maybe never should've shared a name with it anyways). 

I added a test file/function which works correctly in a Jupyter notebook, but running it from terminal gives 
```Traceback (most recent call last):
  File "/Users/claytonstrawn/yt/yt/visualization/tests/test_line_annotation_unit.py", line 21, in test_ds_arr_invariance_under_projection_plot
    p.show()
  File "/Users/claytonstrawn/yt/yt/visualization/plot_container.py", line 107, in newfunc
    rv = f(*args, **kwargs)
  File "/Users/claytonstrawn/yt/yt/visualization/plot_container.py", line 665, in show
    raise YTNotInsideNotebook
yt.utilities.exceptions.YTNotInsideNotebook: This function only works from within an IPython Notebook.
>>> ```

Unfortunately the bug itself seems to be generated only when `show()` is called, so I'm not sure how I should change that. I've never actually written a test for `yt` before, so apologies if I missed something.

**Comment by neutrinoceros** [comment]
Indeed the `.show()` method isn't suited for tests. What's important for your test is that the plot object gets generated. The simplest way to replace the call to `p.show` is to use `p.save()` instead. Now to avoid polluting the disk space for anyone, you'll need to run this test in a temporary directory. There's more than one way to do it, the easiest of which is to use a pytest fixture. Here's how it would go

```python
def test_annotation_unit(tmp_path):
    ... # setup
    p = ProjectionPlot(...)
    p.save(tmp_path) # replaces p.show()
    assert ... # your test
```

We'll also need to explicitly exclude the test from our nosetest-based CI jobs since it's pytest-based. This is done by adding a line at the end of `tests/tests.yaml` (`--ignore-files=...`) as well as `nose_unit.cfg`. (Sorry about how complicated this process currently is. We're still actively migrating from nose to pytest so it's currently a little tricky to add pure-pytest tests).

neutrinoceros:
note that there's at least one bug that was introduced, your new test isn't the only one failing :
```
>           coord_copy = coord.to("code_length")
1627
E           AttributeError: 'numpy.ndarray' object has no attribute 'to'
```

yt/visualization/plot_modifications.py
```diff
@@ -77,8 +77,10 @@ def _project_coords(self, plot, coord):
         """
         if len(coord) == 3:
             if not isinstance(coord, YTArray):
-                coord = plot.data.ds.arr(coord, "code_length")
-            coord.convert_to_units("code_length")
+                coord_copy = plot.data.ds.arr(coord, "code_length")
+            #coord is being copied so that if the user has a unyt_array already
+            #we don't change the user's version
+            coord_copy = coord.to("code_length")
```
neutrinoceros, line 83: I think this should fix the bug that was introduced by changing the variable names
```suggestion
                coord_copy = plot.data.ds.arr(coord, "code_length")
            else:
                #coord is being copied so that if the user has a unyt_array already
                #we don't change the user's version
                coord_copy = coord.to("code_length")
```

neutrinoceros (changes requested):

yt/visualization/tests/test_line_annotation_unit.py
```diff
@@ -0,0 +1,27 @@
...
+    end = ds.arr(end_source, "unitary")
+
+    start_i = start.copy()
+    end_i = end.copy()
+
+    p = yt.ProjectionPlot(ds, 0, "number_density")
+    p.annotate_line(start, end)
+    p.show()
+
+    # for lack of a unyt.testing.assert_unit_array_equal function
```
neutrinoceros, line 23: https://github.com/yt-project/unyt/pull/196

claytonstrawn:
> There's more than one way to [save in a temp directory], the easiest of which is to use a pytest fixture. Here's how it would go
> 
> ```python
> def test_annotation_unit(tmp_path):
>     ... # setup
>     p = ProjectionPlot(...)
>     p.save(tmp_path) # replaces p.show()
>     assert ... # your test
> ```

Does this just mean add "tmp_path" as an argument to the test? I don't really understand how pytest will know what kind of variable this is, but it makes sense why that change would be needed.

> We'll also need to explicitly exclude the test from our nosetest-based CI jobs since it's pytest-based. This is done by adding a line at the end of `tests/tests.yaml` (`--ignore-files=...`) as well as `nose_unit.cfg`. (Sorry about how complicated this process currently is. We're still actively migrating from nose to pytest so it's currently a little tricky to add pure-pytest tests).
>
Unfortunately I do not understand this. I added a line to the bottom of the `.yaml` and in the `.cfg` file to ignore this test following the format for others, but I will admit I don't really know how pytest works. Does all this seem right?

neutrinoceros:
> Does this just mean add "tmp_path" as an argument to the test? I don't really understand how pytest will know what kind of variable this is, 

this is what pytest calls a fixture. I'll agree with you that this feels a bit magical, and it doesn't ressemble any other framework I know, but _yes_, pytest will understand exactly what it means because it's a builtin fixture (see https://docs.pytest.org/en/6.2.x/tmpdir.html)

> Unfortunately I do not understand this. I added a line to the bottom of the .yaml and in the .cfg file to ignore this test following the format for others, but I will admit I don't really know how pytest works. Does all this seem right?

Perfect, from what I can tell, well done !

Now linters are complaining, please read the bottom of this report to see how you can fix this with a comment here.
https://results.pre-commit.ci/run/github/88776493/1629226629.jinP8T3zSY-JxJ-UEpiK6w

you might also want to learn about pre-commit, so you can fix these errors in real time as you develop locally. See https://yt-project.org/doc/developing/developing.html#coding-style-guide

claytonstrawn:
pre-commit.ci autofix

yt/visualization/plot_modifications.py
```diff
@@ -77,8 +77,11 @@ def _project_coords(self, plot, coord):
         """
         if len(coord) == 3:
             if not isinstance(coord, YTArray):
-                coord = plot.data.ds.arr(coord, "code_length")
-            coord.convert_to_units("code_length")
+                coord_copy = plot.data.ds.arr(coord, "code_length")
+            # coord is being copied so that if the user has a unyt_array already
+            # we don't change the user's version
+            else:
```
neutrinoceros, line 83: Can you move the comment so that it's clearer what branch is discussed in it please ? After that I think I'll be able to approve + merge (unless CI fails)
```suggestion
            else:
                # coord is being copied so that if the user has a unyt_array already
                # we don't change the user's version
```

yt/visualization/tests/test_line_annotation_unit.py
```diff
@@ -1,6 +1,7 @@
-import yt
 import numpy as np
 
+import yt
```
neutrinoceros, line 3: ah, so sorry I'm only catching this now... We should avoid importing the whole library in a test file. Can you please replace this with relevant imports as
```python
from yt.loaders import load_uniform_grid
from yt.visualization.plot_window import ProjectionPlot
```
and remove the `yt.` namespace from lines 10 and 20 respectively.
```

claytonstrawn:
pre-commit.ci autofix

neutrinoceros:
We're getting there. Thank you so much for your patience !

neutrinoceros (approved):

matthewturk:

PR #3475, opened 2021-08-16
ENH: ensure consistent usage of shebangs

neutrinoceros:
## PR Summary
Cleanup unnecessary shebangs and fix executable permissions for genuine scripts.


chrishavlin (approved):
This is useful and I'm marking as approved but a couple questions/comments:

1. None of the `doc/source/cookbook/` scripts are executable -- I assume this is by design? 
2. Given 1 I got curious about what executables we **do** have. If you do a project search for `if __name__ == "__main__"`, most of the hits that come up don't have executable permissions or shebangs, so they aren't caught by the new pre-commit checks here. Looking through what comes up, though, I suspect that most of these files would need some work to clean up and actually make executable. So beyond the scope of this PR but worth noting.

neutrinoceros:
1. I didn't design this but that was my assumption too
2. I may be wrong but I think files using the `__name__ == "__main__"` construct should be considered mainly as modules, so IMO they should have neither shebangs *nor* executable bits. I don't know if that's the best practice but it seems to be close to what we're doing up to now.

Xarthisius (approved):

PR #3476, opened 2021-08-18
[WIP] SnapshotSeries Object

jcoughlin11:
## PR Summary

Currently, yt has the `DatasetSeries` object for dealing with time series data. However, this object was designed for the case when the data for each time step is contained in its own file (e.g., the data at t=0 is in file0, t=1 is in file1, etc). For certain frontends (the qmc frontend, in this case) the data for multiple time steps can be contained in the same file.

Rather than muck up the existing `DatasetSeries` object to be able to handle this functionality, this PR introduces a new object, the `SnapshotSeries` object. This object is purposely similar to the `DatasetSeries` object, but is geared towards handling multiple time steps existing in the same data file.

A check to use and return such an object is done in `load()` right after the check to determine if a `DatasetSeries` is needed.

## PR Checklist

### DIFF
--- a/yt/data_objects/time_series.py
+++ b/yt/data_objects/time_series.py
@@ -13,6 +13,7 @@ from yt.config import ytcfg
 from yt.data_objects.analyzer_objects import AnalysisTask, create_quantity_proxy
 from yt.data_objects.particle_trajectories import ParticleTrajectories
 from yt.funcs import is_sequence, mylog
+from yt.loaders import load
 from yt.units.yt_array import YTArray, YTQuantity
 from yt.utilities.exceptions import YTException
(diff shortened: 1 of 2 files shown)


neutrinoceros:
I see this is WIP so I won't look too much into details for now but would it make sense to have _some_ inheritance between this class and the DatasetSeries ? Even if none is a subclass of the other, could they derive from a common (abstract ?) ancestor ?

jcoughlin11:
I agree that that would make sense conceptually. I'm not sure how, in practice, though. I'll look into it.

matthewturk:
@jcoughlin11 Do you see yourself returning to this?

Issue #3479, opened 2021-08-24
matplotlib 3.4.3 returns get_data_path as PosixPath


zingale:
### Bug report

When using the latest yt (on `main` in git) and MPL 3.4.3, I get a traceback:

```
Traceback (most recent call last):
  File "/raid/zingale/development/Castro/Exec/science/massive_star/massive_star_slice.py", line 48, in <module>
    sp = yt.SlicePlot(ds, "theta", f,
  File "/raid/zingale/development/yt-mz/yt/yt/visualization/plot_window.py", line 2484, in SlicePlot
    return AxisAlignedSlicePlot(ds, normal, fields, *args, **kwargs)
  File "/raid/zingale/development/yt-mz/yt/yt/visualization/plot_window.py", line 1614, in __init__
    PWViewerMPL.__init__(
  File "/raid/zingale/development/yt-mz/yt/yt/visualization/plot_window.py", line 872, in __init__
    PlotWindow.__init__(self, *args, **kwargs)
  File "/raid/zingale/development/yt-mz/yt/yt/visualization/plot_window.py", line 221, in __init__
    super().__init__(data_source, window_size, fontsize)
  File "/raid/zingale/development/yt-mz/yt/yt/visualization/plot_container.py", line 912, in __init__
    super().__init__(data_source, figure_size, fontsize)
  File "/raid/zingale/development/yt-mz/yt/yt/visualization/plot_container.py", line 260, in __init__
    font_path = matplotlib.get_data_path() + "/fonts/ttf/STIXGeneral.ttf"
TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'
```

The issue is that MPL returns a PosixPath object now:

```
In [4]: matplotlib.get_data_path()                                                                                                                                               
Out[4]: PosixPath('/usr/share/matplotlib/mpl-data')
```

Note that MPL 3.4.1 seems to return a string, so this is a very recent change.

**Bug summary**

**Code for reproduction**

```python
# Paste your code here
#
#
```

**Actual outcome**

```
# If applicable, paste the console output here
#
#
```

**Expected outcome**

**Version Information**

  * Operating System:
  * Python Version:
  * yt version:
  * Other Libraries (if applicable):

zingale:
we can probably just do a `try` / `except` block here.  I'll try to PR a fix shortly.

neutrinoceros:
I can't reproduce this on my system. The function still returns a string on matplotlib 3.4.3 and even on their master branch.

[...]

PR #3480, opened 2021-08-24
BUG: fix mpl data path handling for Windows

zingale:
## PR Summary

This addresses #3479 

## PR Checklist

### DIFF
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -257,7 +257,9 @@ class PlotContainer:
             self.figure_size = float(figure_size[0]), float(figure_size[1])
         else:
             self.figure_size = float(figure_size)
-        font_path = matplotlib.get_data_path() + "/fonts/ttf/STIXGeneral.ttf"
+        font_path = os.path.join(
+            matplotlib.get_data_path(), "fonts", "ttf", "STIXGeneral.ttf"
+        )
         self._font_properties = FontProperties(size=fontsize, fname=font_path)
         self._font_color = None
         self._xlabel = None


yt/visualization/plot_container.py
```diff
@@ -257,7 +257,13 @@ def __init__(self, data_source, figure_size, fontsize):
             self.figure_size = float(figure_size[0]), float(figure_size[1])
         else:
             self.figure_size = float(figure_size)
-        font_path = matplotlib.get_data_path() + "/fonts/ttf/STIXGeneral.ttf"
+
+        try:
+            font_path = matplotlib.get_data_path() + "/fonts/ttf/STIXGeneral.ttf"
+        except TypeError:
+            # MPL 3.4.3 returns a PosixPath
+            font_path = str(matplotlib.get_data_path()) + "/fonts/ttf/STIXGeneral.ttf"
```
neutrinoceros, line 265:
```suggestion
        # MPL >= 3.4.3 returns a Path
        font_path = os.path.join(matplotlib.get_data_path(), "fonts", "ttf", "STIXGeneral.ttf")
```
- using `os.path.join` will not only convert to string, it will also make this line windows-compatible for free !
- the try block is actually not needed since the fallback option is always going to work, so let's go with that one in any case
- the comment needs slight correction: the exact return type of `matplotlib.get_data_path` is not going to be `PosixPath` on windows, (rather a `WindowsPath`, see pathlib's doc https://docs.python.org/3/library/pathlib.html)

neutrinoceros (changes requested):

neutrinoceros:
Thank you for catching this. I must say I'm slightly annoyed that MPL keeps making backwards incompatible changes in patch releases, but well, not your fault.
Unless we plan to have a 4.1 release soon (which seems unlikely), I believe this patch should be back ported into a bug fix release (4.0.2). Another option would be to report to MPL and tell them that this change is breaking for us, see what they say and wether they would agree to revert it. Or we could do both.
What do you think @munkm ?

yt/visualization/plot_container.py
zingale: okay, great, thanks for the ideas.  I'll update.

neutrinoceros:
failures are absolutely unrelated and caused by a backward compatibility breakage upstream (firefly-vis), reported https://github.com/ageller/Firefly/issues/62

neutrinoceros:
Here's the quick-and-dirty patch to solve this issue quickly: #3482

neutrinoceros:
Ah, there's something very fishy here. Here's the definition of `matplotlib.get_data_path` as of MPL 3.4.3

```python
@_logged_cached('matplotlib data path: %s')
def get_data_path():
    """Return the path to Matplotlib data."""
    return str(Path(__file__).with_name("mpl-data"))
```
It also has changed for 7 months, and the return statement wasn't update in the last couple years, so I don't know _how_ this ended up breaking for you but it would be very interesting to have a deeper look, I think.
@zingale could you produce an script that shows the error you wanted to fix here ? Your patch is important in any case, but I'd like to know if we need to report/fix something upstream.

yt/visualization/plot_container.py
```diff
@@ -257,7 +257,10 @@ def __init__(self, data_source, figure_size, fontsize):
             self.figure_size = float(figure_size[0]), float(figure_size[1])
         else:
             self.figure_size = float(figure_size)
-        font_path = matplotlib.get_data_path() + "/fonts/ttf/STIXGeneral.ttf"
+
+        # MPL >= 3.4.3 returns a Path
+        font_path = os.path.join(matplotlib.get_data_path(), "fonts", "ttf", "STIXGeneral.ttf")
+
```
neutrinoceros, line 263: Actually the comment turns out to be ill-informed, so I suggest cleaning it up.
```suggestion
        font_path = os.path.join(matplotlib.get_data_path(), "fonts", "ttf", "STIXGeneral.ttf")
```

neutrinoceros:
Note: failures are not related, seen on multiple PRs, and reported as #3491

neutrinoceros:
@yt-fido test this please

Xarthisius (approved):

matthewturk:

PR #3482, opened 2021-08-25
MNT: set upper limit to firefly-vis version in [full] target as a workaround a backward incompatibility breakage

neutrinoceros:
## PR Summary
As reported upstream https://github.com/yt-project/yt/runs/3416844787?check_suite_focus=true), the newest version of firefly-vis breaks our firefly tests
See for instance https://github.com/yt-project/yt/runs/3416844787?check_suite_focus=true

As a temporary fix, I'm pinning the version we test against.


neutrinoceros:

Issue #3473, opened 2021-08-16
BUG: `ds.quan` VS integers on windows


neutrinoceros:
### Bug report

**Bug summary**

Following a discussion on Slack with @ianww, I'd like to report a seemingly platform dependent problem he encountered on windows, and that I can't easily reproduce for lack of a windows machine.

**Code for reproduction**

Going through this notebook https://yt-project.org/doc/quickstart/volume_rendering.html
Ian reported that his Jupyter kernel died on `sc.show()`. He was able to work around this crash by passing an explicit float literal instead of an integer to `ds.quan` as 
```python
sc.camera.set_width(ds.quan(20., 'cm'))
```
This is as close to a minimal example I can get without a windows machine.

**Version Information**
  * Operating System: Windows 10
  * Python Version: 3.7
  * yt version: 4.0.1
  * Other Libraries (if applicable): unyt 2.8.0

@ianww, do not hesitate to add further information here if I forgot anything important, and thanks again for your initial report and patience.

neutrinoceros:
So this seems hardly actionable (maybe not even reproducible). I'm going to close this, but I'll still ping @jzuhone in case he wants to have a look.

PR #3483, opened 2021-08-25
MNT: add button to manually run bleeding edge CI job

neutrinoceros:
## PR Summary
This will help closing #3473 faster by triggering a run of the bleeding edge CI job as this PR is merged.
It adds a button so the job can be triggered manually next time a similar situation occurs instead of having to wait the next natural trigger (scheduled or merge) to validate a patch upstream.


munkm:
And generally we shouldn't be self-merging either.

PR #3494, opened 2021-08-31
MNT: migrate away from packaging.version.parse_versio

neutrinoceros:
## PR Summary

Currently, `packaging.version.parse_version` is implemented as
```python
def parse(version):
    # type: (str) -> Union[LegacyVersion, Version]
    """
    Parse the given version string and return either a :class:`Version` object
    or a :class:`LegacyVersion` object depending on if the given version is
    a valid PEP 440 version or a legacy version.
    """
    try:
        return Version(version)
    except InvalidVersion:
        return LegacyVersion(version)
```

but `LegacyVersion` is deprecated since https://github.com/pypa/packaging/pull/342
```python
class LegacyVersion(_BaseVersion):
    def __init__(self, version):
        # type: (str) -> None
        self._version = str(version)
        self._key = _legacy_cmpkey(self._version)

        warnings.warn(
            "Creating a LegacyVersion has been deprecated and will be "
            "removed in the next major release",
            DeprecationWarning,
        )
```
and the plan for their next release is to have `packaging.version.parse` basically be an alias for `packaging.version.Version.__init__`, see https://github.com/pypa/packaging/pull/407

So I'm proposing we move forward with is and stop using the `parse` function as it may cause confusion in the future and is more complicated to use.
Note that since we're now treating warnings as errors in CI (#3380) means that we can switch with confidence because we _know_ that we don't currently have any such warnings to deal with.


neutrinoceros:
A possible alternative is to build our own `parse_version` function with a fall back mechanism that would return an object with the same important properties as `packaging.version.LegacyVersion`, i.e. that would always compare as lower than any given valid version.

Xarthisius (approved):

neutrinoceros:

PR #3495, opened 2021-08-31
ENH: warn user in case a Dataset subclass is redefined

neutrinoceros:
## PR Summary

Following a recent discussion on the Slack space in which we discussed how yt extensions
provided support to create third party frontends, which allows users to extend over native one as

```python
import yt
import yt.extensions.my_frontend
```
The only issue I was able to see with this workflow is that, in case such a frontend gets integrated
into yt itself, then users may not realize that the extension import became useless, or worse. Shadowing a "first citizen" frontend may actually be useful as temporary measure, which is why I'm not making this an error, but it seems fair to emit a warning as a forgiving gesture to transiting users.

## PR Checklist

[N/A] New features are documented, with docstrings and narrative docs

note : I want to add documentation for the whole extension frontend workflow later, but I think this small patch counts as its own documentation.

### DIFF
--- a/nose_unit.cfg
+++ b/nose_unit.cfg
@@ -6,5 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
-ignore-files=(test_load_errors.py|test_load_sample.py|test_commons.py|test_ambiguous_fields.py|test_field_access_pytest.py|test_save.py|test_line_annotation_unit.py|test_eps_writer.py)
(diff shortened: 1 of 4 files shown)


brittonsmith:
This seems like a reasonable solution to me. I wouldn't want to prevent an extension over-writing an internal frontend, but a warning to make it clear what's happening is good.

brittonsmith (approved):

neutrinoceros:
@Xarthisius @brittonsmith, any reason not to merge this ?

PR #3496, opened 2021-09-01
BUG: fix two bugs in yt.visualization.eps_writer

neutrinoceros:
## PR Summary
Noticed that eps_writer didn't have tests so I ran an example from the documentation and was able to find a trivial but blocking bug.
While testing this I also fixed another bug where save_fig didn't accept Path objects.

## PR Checklist

[N/A] New features are documented, with docstrings and narrative docs

### DIFF
--- a/nose_unit.cfg
+++ b/nose_unit.cfg
@@ -6,5 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
-ignore-files=(test_load_errors.py|test_load_sample.py|test_commons.py|test_ambiguous_fields.py|test_field_access_pytest.py|test_save.py|test_line_annotation_unit.py)
+ignore-files=(test_load_errors.py|test_load_sample.py|test_commons.py|test_ambiguous_fields.py|test_field_access_pytest.py|test_save.py|test_line_annotation_unit.py|test_eps_writer.py)
 exclude-test=yt.frontends.gdf.tests.test_outputs.TestGDF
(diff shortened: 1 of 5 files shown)


neutrinoceros:
Turns out testing this in CI is _hard_. I'll get back to it.

jwise77:
Thanks for creating this test and fixing the bug. All of the changes you've made so far look good. I can review any other future changes in this PR.

neutrinoceros:
So I think the reason why CI was failing is that the "tex" executable isn't available on the virtual machines we test with. Unfortunately this means that my test is doomed to never run automatically, which limits greatly its value. On the other hand it seems clear that this module isn't heavily used currently since no body seemed to have noticed it was broken, and it's still better to have a reference test than no tests.

On a broader level, I'm thinking this module likely has applicability far beyond yt and it would probably make sense to migrate it to its own project ? it would also help making sure it's properly tested.
And even before that, I understand that you wrote this functionality out of frustration with matplotlib's own rasterising capabilities, any idea if it got any better since then ?
This discussion goes well beyond the scope of this PR though, maybe we should open a dedicated issue.

neutrinoceros:
What's important to review this PR is to make sure my test works at least locally.

jwise77 (approved):
This all looks good. Thanks again. I agree that it'll be easier to test this module if it's an extension. I'll make an issue and work on it soon.

jzuhone (approved):

PR #3499, opened 2021-09-05
ENH: inline yt.funcs.is_sequence

neutrinoceros:
## PR Summary

This is in support to YTEP 0038, and more broadly in support to better typing in yt.
It is a follow up to #2893, in which I renamed `has_len` into `is_sequence` based on a misconception
I was having about the word ["sequence" in Python](https://docs.python.org/dev/glossary.html#term-sequence).

I encourage reviewers to read this short page https://docs.python.org/3/library/collections.abc.html

So, the main reason this function needs to be discouraged is that its name is misleading. It is _not_ equivalent to the following implementation
```python
from collections.abc import Sequence
def is_sequence(obj):
    # check that `obj` has __getitem__, __len__, __reversed__ ...
    return isinstance(obj, Sequence)
```
On the other hand, it _is_ equivalent to
```python
from collections.abc import Sized
def is_sequence(obj):
    # check that obj has __len__
    return isinstance(obj, Sized)
```
and because this implementation is try-free as well as simpler, I replaced the current implem with it. 

Now, since it is apparent that it can now be inlined at the same cost (one import), it seems reasonable to drop its usage internally, but we can still keep it for now, to avoid breaking downstream applications.

In many places where it is used, checking that a given object knows the `__len__` protocol isn't exactly the appropriate thing to do. Sometimes we really want to check that objets are instances of `collections.abc.Sequence` or `collections.abc.Collection`. However, to avoid breaking anything by changing a lot of code at once, I stopped at the inlining step, except in one or two obvious cases, which I will highlight here with comments.

It is worth noting that `np.ndarray` objects are _not_ considered as `collections.abc.Sequence`, because they don't implement the `__reversed__` protocol. See https://docs.python.org/dev/glossary.html#term-sequence for more detail

## PR Checklist

### DIFF
--- a/yt/data_objects/selection_objects/data_selection_objects.py
+++ b/yt/data_objects/selection_objects/data_selection_objects.py
@@ -1,6 +1,7 @@
 import itertools
 import uuid
 from collections import defaultdict
+from collections.abc import Iterable, Sized
 from contextlib import contextmanager
 
 import numpy as np
@@ -12,7 +13,7 @@ from yt.data_objects.data_containers import YTDataContainer
 from yt.data_objects.derived_quantities import DerivedQuantityCollection
 from yt.data_objects.field_data import YTFieldData
 from yt.fields.field_exceptions import NeedsGridType
-from yt.funcs import fix_axis, is_sequence, iter_fields, validate_width_tuple
+from yt.funcs import fix_axis, iter_fields, validate_width_tuple
 from yt.geometry.selection_routines import compose_selector
 from yt.units import YTArray, dimensions as ytdims
 from yt.utilities.exceptions import (
@@ -594,7 +595,7 @@ class YTSelectionContainer2D(YTSelectionContainer):
             )
 
             validate_width_tuple(width)
-            if is_sequence(resolution):
+            if isinstance(resolution, Iterable):
                 resolution = max(resolution)
             frb = CylindricalFixedResolutionBuffer(self, width, resolution)
             return frb
@@ -603,9 +604,9 @@ class YTSelectionContainer2D(YTSelectionContainer):
             center = self.center
             if center is None:
                 center = (self.ds.domain_right_edge + self.ds.domain_left_edge) / 2.0
-        elif is_sequence(center) and not isinstance(center, YTArray):
+        elif isinstance(center, Sized) and not isinstance(center, YTArray):
             center = self.ds.arr(center, "code_length")
-        if is_sequence(width):
+        if isinstance(width, Sized):
             w, u = width
             if isinstance(w, tuple) and isinstance(u, tuple):
                 height = u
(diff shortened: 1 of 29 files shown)


yt/data_objects/selection_objects/data_selection_objects.py
```diff
@@ -594,7 +595,7 @@ def to_frb(self, width, resolution, center=None, height=None, periodic=False):
             )
 
             validate_width_tuple(width)
-            if is_sequence(resolution):
+            if isinstance(resolution, Iterable):
```
neutrinoceros, line 598: This is one of the few cases that I corrected the logic

yt/data_objects/time_series.py
```diff
@@ -169,7 +170,7 @@ def __init__(
     ):
         # This is needed to properly set _pre_outputs for Simulation subclasses.
         self._mixed_dataset_types = mixed_dataset_types
-        if is_sequence(outputs) and not isinstance(outputs, str):
+        if isinstance(outputs, Sequence) and not isinstance(outputs, str):
```
neutrinoceros, line 173: Here too

yt/visualization/plot_container.py
```diff
@@ -253,7 +254,11 @@ def __init__(self, data_source, figure_size, fontsize):
         self.data_source = data_source
         self.ds = data_source.ds
         self.ts = self._initialize_dataset(self.ds)
-        if is_sequence(figure_size):
+        if isinstance(figure_size, Sized):
+            if len(figure_size) != 2:
```
neutrinoceros, line 258: I added a sensible check here

matthewturk:
Can you remind me if this has or had any impact on our usage of `ensure_list` to make sure that field lists are in fact lists of fields?

neutrinoceros:
@matthewturk `ensure_list` was removed in #2893
The present PR is designed to have _no_ impact on anything, besides the two (extremely minor) corrections highlighted above.
It is only about avoiding to entertain an incorrect notion of what "sequences" mean in Python, and make room to later use `is instance(..., Sequence)` where relevant.

neutrinoceros:
The GH workflow failures seem to be due to a change upstream on MacOs + home-brew + proj + gets + Python 3.9.7... So fixing this is out of scope here. However, the failures on Jenkins are real, so I'll make this a draft for now.

neutrinoceros:
This allowed me to discover a bug in unyt https://github.com/yt-project/unyt/issues/201
though this problem is blocking for the current PR so I'll _keep_ it as a draft until it is resolved upstream.

neutrinoceros:
Turns out the bug isn't in unyt, but in numpy. I think this PR is doomed because numpy arrays have weird behaviours and don't seem to play well with `collections.abc`. Down the line these problems completely break what I was trying to do here

See https://github.com/numpy/numpy/issues/19833, and a related, long standing issue https://github.com/numpy/numpy/issues/2776

I'll open another PR to try and reflect this knowledge better in our code base.

PR #3504, opened 2021-09-07
MNT: hotfix CI

neutrinoceros:
## PR Summary

Right now our Github Actions-based CI is completely broken as it fails the install stage.
I'm suspecting the most recent updates in setuptools to be responsible. Specifically https://github.com/pypa/setuptools/issues/2775

This is meant as a temporary patch to allow our CI to run.

### DIFF
--- a/.github/workflows/build-test-pytest.yaml
+++ b/.github/workflows/build-test-pytest.yaml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [
-          # macos-latest, # TEMP see https://github.com/yt-project/yt/pull/3504#issuecomment-914365378
+          macos-latest,
           windows-latest,
           ubuntu-18.04,
         ]
--- a/.github/workflows/build-test-pytest.yaml
+++ b/.github/workflows/build-test-pytest.yaml
@@ -21,7 +21,11 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


neutrinoceros:
So this is a success because it fixes the issue addressed (install stage is now go). The job is _still_ failing because of yet another, yet unidentified upstream change, but seemingly this is specific to macOS, so let's try deactivating as a sanity check.

neutrinoceros:
setuptools 58.0.3 was released a couple hours ago and should resolve the problem, so I'm reverting the initial hotfix. Also this a draft since currently the only net effect of this PR isn't what's described in the OP

neutrinoceros:
Alright so the newest setuptools version still breaks our CI. I believe this time it can be considered intentional as they are actively removing support for packages using the `use_2to3` flag. This seems like an opportunity to bumps some of our "minimal" requirements. `IPython` 1.0.0 was released 8 years ago, I think it's now okay to require a more recent version, so let's try that.

neutrinoceros:
It may be okay to reactivate macOS now that https://github.com/actions/virtual-environments/issues/4020 was closed.

neutrinoceros:
At last, everything is now working, but this PR has 5 commits (which I'm going to keep as "documentation" of what happened), 4 of which are mutually cancelling, so I'm going to open a different PR with only the last relevant commit. I'll close this one once CI goes full green.

PR #3511, opened 2021-09-14
[WIP] Dask particle read

chrishavlin:
Draft PR for a YTEP in progress. Will add more details here soon...

### DIFF
--- a/yt/frontends/gadget/io.py
+++ b/yt/frontends/gadget/io.py
@@ -67,6 +67,19 @@ class IOHandlerGadgetHDF5(IOHandlerSPH):
                 yield ptype, (x, y, z), hsml
             f.close()
 
+    def _datafile_has_ptf(self, data_file, ptf):
+        """
+        checks a data_file object to see if any particle types in ptf dict
+        have a non-zero particle count in this file. use to avoid opening
+        files if we know the particle counts a priori (the index construction
+        does not differentiate by particle type).
+        """
+        for ptype, _ in sorted(ptf.items()):
+            if data_file.total_particles[ptype] > 0:
+                # at least one ptype has particle types, need to read this file
+                return True
+        return False
+
     def _yield_coordinates(self, data_file, needed_ptype=None):
         si, ei = data_file.start, data_file.end
         f = h5py.File(data_file.filename, mode="r")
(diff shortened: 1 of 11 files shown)


yt/utilities/parallel_tools/dask_helper.py
```diff
@@ -0,0 +1,145 @@
...
+from dask.distributed import Client
+
+from yt.utilities.logger import ytLogger as mylog
+
+# a couple of warning messages used in several places
+here_be_dragons = (
+    "You have chosen to attempt multi-threading. Many frontends are not"
+    " thread-safe and many yt functions are built for single-threading so "
+    "you will likely encounter bugs (including silent failures and hangs)."
+    "Good luck!"
```
matthewturk, line 11: Seems like a good place to include a few emojis, since we're now Py3 only!

yt/utilities/parallel_tools/dask_helper.py
```diff
@@ -0,0 +1,145 @@
...
+    a container to help with managing dask clients for yt. Enforces
+    single-threading for new clients and allows us to easily track if a client
+    is open.
+
+    Intentionally does NOT spin up a dask client on instantiation so that
+    a user is responsible for starting their own client. With no client,
+    the wrapper functions in dask_helper assume one single-threaded processor.
+    """
+
+    def __init__(self):
```
matthewturk, line 32: We may need to do something to do our best to make sure this is a singleton.  I'm not 100% sure that this will be an issue, but in some cases I believe it is conceivable that the contents of a module may be executed multiple times.  This information may be outdated; however, in the case that it *can* still happen, it's not completely impossible that this could be run twice and the top-level `dask_client` instance may be different depending on which reference is utitilized.

yt/frontends/gadget/io.py
```diff
@@ -67,6 +67,19 @@ def _read_particle_coords(self, chunks, ptf):
...
             f.close()
 
+    def _datafile_has_ptf(self, data_file, ptf):
+        """
+        checks a data_file object to see if any particle types in ptf dict
+        have a non-zero particle count in this file. use to avoid opening
+        files if we know the particle counts a priori (the index construction
+        does not differentiate by particle type).
+        """
+        for ptype, _ in sorted(ptf.items()):
```
matthewturk, line 77: I know you didn't originate this, but we can change this to just be `for ptype in sorted(ptf)`

neutrinoceros:
Just merged #3526
I assume this needs a rebase ?

chrishavlin:
Yup, a rebase and some rewriting!

Issue #3510, opened 2021-09-14
Multiple data type candidates


mameehan5:
### Bug report

**Bug summary**

Multiple data types were found to be eligible for loading the dataset due to shear coincidence that the string in `_subtype_keyword` matches an expression for the name of the computer.

**Code for reproduction**
```python
ts = yt.DatasetSeries("plt?????", parallel=True)
```
The dataset is an `AMReXDataset`, which is how I would prefer to load it, but in `job_info`, there is a line 
```
build machine: Linux onyx07 4.4.178-94.91-default #1 SMP Thu May 2 20:14:31 UTC 2019 (d683ca1) x86_64 x86_64 x86_64 GNU/Linux
```
which has the `_subtype_keyword` `nyx`, which means a `NyxDataset` is also valid.

**Actual outcome**

```
yt.utilities.exceptions.YTAmbiguousDataType: Multiple data type candidates for plt66875
The following independent classes were detected as valid :
<class 'yt.frontends.boxlib.data_structures.NyxDataset'>
<class 'yt.frontends.boxlib.data_structures.AMReXDataset'>
A possible workaround is to directly instantiate one of the above.
Please report this to https://github.com/yt-project/yt/issues/new
```

**Expected outcome**

Ideally, this would load as normal. The possible workaround to instantiate using `from yt.frontends.boxlib.data_structures import AMReXDataset` to load the datset does not work for data series. 

My thinking is that it would be nice to avoid this all together by passing a keyword, such as
```
yt.DatasetSeries("plt?????", parallel=True, dataset='AMReXDataset')
```
which would force yt to load the files as that dataset. This could be useful for others who ever have a similar issue.

**Version Information**

  * Operating System: macOS Catalina
  * Python Version: 3.6.13
  * yt version: 4.1.dev0

Using conda and an install from source.

neutrinoceros:
> The possible workaround to instantiate using from yt.frontends.boxlib.data_structures import AMReXDataset to load the datset does not work for data series.

just a quick note that the normal way to do this from the user standpoint is to import from `yt.frontends.boxlib.api` instead
```
from yt.frontends.boxlib.api import AMReXDataset
```

but indeed this is the workaround for this unfortunate situation.

> My thinking is that it would be nice to avoid this all together by passing a keyword

This would go against the design patterns in place, but yeah I guess we could have such a keyword as supported workaround, but I'd argue that the argument should be a class, not a str representing a class name, for it to be sane.

[...]

Issue #3005, opened 2020-12-21
amrex dataset detection appears broken


zingale:
### Bug report

**Bug summary**

If I try to load a Castro dataset, I get:

```
  File "/raid/zingale/development/yt-mz/yt/yt/loaders.py", line 98, in load
    raise YTAmbiguousDataType(fn, candidates)
yt.utilities.exceptions.YTAmbiguousDataType: Multiple data type candidates for det_strang_cfl0.5_dtnuce1e+200_nzones256/det_x_plt00210
The following independent classes were detected as valid :
<class 'yt.frontends.boxlib.data_structures.CastroDataset'>
<class 'yt.frontends.boxlib.data_structures.AMReXDataset'>
A possible workaround is to directly instantiate one of the above.
Please report this to https://github.com/yt-project/yt/issues/new
```

**Code for reproduction**

Using this dataset:
http://bender.astro.sunysb.edu/random/det_x_plt00000.tgz

Do
```
yt.load("det_x_plt00000")
```

A workaround is to use the hash `9071f0dc3` for yt.

```python
# Paste your code here
#
#
```

**Actual outcome**

```
# If applicable, paste the console output here
#
#
```

**Expected outcome**

should be able to load the dataset

**Version Information**

  * Operating System: Fedora 33
  * Python Version: 3.9.0
  * yt version: 4.0.dev0
  * Other Libraries (if applicable): 

installed yt from source

neutrinoceros:
Hi @zingale, sorry about this, this must be an undesirable side effect of https://github.com/yt-project/yt/pull/2938
I worked a fair bit on simplifying the validators for Boxlib child classes, but it's still not very robust. In fact this is a predictable situation where both strings `"amrex"` and `"castro"` are present in the paramfile (`job_info`).
Could you explain why both are showing up and how should yt guess that it's indeed a Castro-typed dataset ?

zingale:
yes, they will both always show up in all MAESTROeX, NyX, and Castro datasets.  The reason is that they are all different git repos, so we need to keep track of the individual hashes of each.

So Castro will always report the Castro, AMReX, and Microphysics hashes (the last doesn't matter to yt really)
MAESTROeX will always report the MAESTROeX, AMReX, and Microphysics
Nyx will always report the Nyx, AMReX hashes

probably true for WarpX too

neutrinoceros:
I see. I don't know what we should do about it then because I don't know of a reliable way yt could guess the appropriate type.
[...]

PR #3512, opened 2021-09-14
ENH: Allow user to set data set of the time for ambiguous data

mameehan5:
## PR Summary

This pull request is meant to fix an issue when a time series object is created but there is ambiguity in the data type. This PR should enable the user to set the data type in the argument, for example
```python
ts = yt.DatasetSeries("plt?????", parallel=True, dataset='AMReXDataset')
```
to ensure that data loads as an AMReX data set. This is meant to address #3510 .

## PR Checklist

- [x] New features are add

Would love some feed back to see if this modification makes sense or what needs to be changed before adding the documentation. Thanks!

### DIFF
--- a/yt/data_objects/time_series.py
+++ b/yt/data_objects/time_series.py
@@ -418,14 +418,14 @@ class DatasetSeries:
 
     _dataset_cls = None
 
-    def _load(self, output_fn, **kwargs):
+    def _load(self, output_fn, dataset=None, **kwargs):
         from yt.loaders import load
 
         if self._dataset_cls is not None:
             return self._dataset_cls(output_fn, **kwargs)
         elif self._mixed_dataset_types:
             return load(output_fn, **kwargs)
-        ds = load(output_fn, **kwargs)
+        ds = load(output_fn, dataset=dataset, **kwargs)
         self._dataset_cls = ds.__class__
         return ds
 
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -34,7 +34,7 @@ from yt.utilities.on_demand_imports import _pooch as pooch
 # --- Loaders for known data formats ---
 
 
-def load(fn, *args, **kwargs):
+def load(fn, *args, dataset=None, **kwargs):
     """
     Load a Dataset or DatasetSeries object.
     The data format is automatically discovered, and the exact return type is the
@@ -84,7 +84,8 @@ def load(fn, *args, **kwargs):
     candidates = []
     for cls in output_type_registry.values():
         if cls._is_valid(fn, *args, **kwargs):
-            candidates.append(cls)
+            if dataset is None or dataset in cls.__name__:
+                candidates.append(cls)
 
     # Find only the lowest subclasses, i.e. most specialised front ends
     candidates = find_lowest_subclasses(candidates)


mameehan5:
I am happy to terminate this PR since I think we have resolved #3510 . Or if we think this is worth pursuing, I can keep it open then make changes at a later time.

neutrinoceros:
apparently @Xarthisius is ok with adding a simple `dataset_cls` keyword, and I'd approve it too because if you went through this process, it means that the workaround isn't easily discovered or intuitive enough. My hope is that a simple keyword argument would be easier to discover (and document), so I'm still favourable to it.

neutrinoceros:
@mameehan5 Do you still wish to work on this, and if so, do you need a hand ?
Potentially this could close two issues (#3005 and #3510) at once, so I believe it is worth pursuing. I can also take over this PR if you wish.

mameehan5:
Hi @neutrinoceros , I didn't realize this and the original issue were still open! I am happy to pursue this a bit further, but if I can't get anything by the end of the week, I will hand the keys over to you.

I will say, following the thread of #3005 , it sounds like adding a frontend/code/etc. argument was the popular vote for the fix. This is something I am happy to do! However, both #3005 and #3510 are a result of the same issue with AMReX datasets: my PeleLM data, which builds off AMReX, has the word `nyx` in the `job_info` file because that is the HPC we run on. So my question would be, for addressing these two issues, would it be better to actually address the robustness of finding the correct frontend for AMReX data, that way users won't have to do this in the future?

To @matthewturk 's question/suggestion in #3005 , I do have a good handle of the hierarchy from the PeleC/PeleLM/IAMR/AMReX but not as much for Castro/Maestro/etc. and probably the many others. The idea of letting AMReX be a default with the option to be a more specific code could be useful for the many users that want to do something simple (slice, basic stats, etc.). Then, if yt uses the default AMReX, and the user is not okay with that for a specific reason, then the keyword can be used to override. 

Thoughts?

neutrinoceros:
> So my question would be, for addressing these two issues, would it be better to actually address the robustness of finding the correct frontend for AMReX data, that way users won't have to do this in the future?

Oh yeah, if you know how to do that, then by all means, feel free to !

> To @matthewturk 's question/suggestion in #3005 , I do have a good handle of the hierarchy from the PeleC/PeleLM/IAMR/AMReX but not as much for Castro/Maestro/etc. and probably the many others. The idea of letting AMReX be a default with the option to be a more specific code could be useful for the many users that want to do something simple (slice, basic stats, etc.). Then, if yt uses the default AMReX, and the user is not okay with that for a specific reason, then the keyword can be used to override.
>
>Thoughts?

Sounds good to me !

mameehan5:
@neutrinoceros I have probably sunk too much time into this problem and not getting really anywhere so I will let you take over this PR! Let me know if you need me to do anything specific on my end. Really curious what solution you propose!

neutrinoceros:
okay thanks for your initial effort, and thank you for getting back to me. I should already be able to push directly to your branch. I'll ping you in case I'm not, and in any case I'd like to request your review when I'm done, if that's alright :)

neutrinoceros:
Ah. I see that you wrote this PR from the main branch of your fork, so I'm not going to push there after all. Instead, I'll close this PR and open a new one, but I'll still keep your contribution as the basis of my own branch !

PR #3519, opened 2021-09-16
ENH: improve error messages from PLViewerMPL._setup_origin (user input validation).

neutrinoceros:
## PR Summary
Started with removing an assert statement that was apparently used to validate user input; this is bad practice because the Python runtime can be run with a `-O`flag that suppresses assert statements.
Then I found an out of date (and ~ unreachable) error message, so I updated it.
Then I saw that all the error raised were `RuntimeError`s, when really `ValueError` is appropriate, so I changed that too.
Finally I changed some existing log entrues from `warning` to `debug` because well... if we're going to raise an error right after that, a warning doesn't mean much, especially when the same bits of information are also contained in the error message itself.

I can (and should) add some tests for this later, but please feel free to comment now if for any reason you think this is a bad idea.

## PR Checklist


neutrinoceros:
I've started adding tests and refactoring some more, I need to come back to it to get test to green

PR #3520, opened 2021-09-17
BUG: restore yt's classic math font even with matplotlib >= 3.4

### LINKED ISSUE #3514: Change of style in matplotlib
### Bug report

**Bug summary**

Matplotlib 3.4 came with a change in default math text font that affects how yt plots look by default. IMO the "classic" yt look is much nicer so I'd love that it stayed the default style as far as yt is concerned, with no regards to what matplotlib version is installed.

**Code for reproduction**

Here's a simple demo using yt 4.0 and matplotlib 3.4
```python
import yt

def myplot():
    sp = yt.SlicePlot(ds, "z", ("gas", "density"))
    return sp

# default options with matplotlib 3.4
myplot().save("/tmp/defaults")
```

**Actual outcome**

defaults_Slice_z_density

**Expected outcome**
This is the "classic yt" look that required no tweaking before Matplotlib 3.4
classic_yt_Slice_z_density

now it can be reproduced with any version of matplotlib with some additional effort on the user side.
```python
plt.rcParams["mathtext.fontset"] = "cm"
myplot().save("/tmp/classic_yt")
```
this is the simplest way to do it, but unfortunately it has side effects, as it won't just change the settings for that one yt plot, but globally change how matplotlib looks through out the session.

Side effects can be avoided using [a context manager](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.rc_context.html)
```python
# with temporary settings (no side effects)
with plt.rc_context({"mathtext.fontset": "cm"}):
    myplot().save("/tmp/classic_yt")
```
but this technique is IMO too disruptive.

I suggest we could use such context manager internally to restore yt's default look with no tweaking on the user side. In order to preserve flexibility (users should still be able to set a different font), some design decisions would be needed.
What do other users and developers think ? 

**Version Information**
  * Operating System: OsX
  * Python Version: 3.9.5
  * yt version: 4.0.1
  * Other Libraries (if applicable): Matplotlib 3.4.2

neutrinoceros:
## PR Summary
Endgoal: resolve #3514
Basic testing seems to indicate this works, however I need to further check a few things. Is it still possible to change the math font using the matplotlib API ?
Submitting as a draft for now so I can at least check if image comparison tests pass as is.

## PR Checklist


neutrinoceros:
test cases to add:

```python
import matplotlib as mpl
from yt.visualization.api import SlicePlot

ds = yt.testing.fake_amr_ds(fields=[("gas", "density")], units=["g/cm**-3"])
sp = yt.SlicePlot(ds, "z", ("gas", "density"))
```
Expected result

3 4 3_Slice_z_density

```
# this case requires matplotlib 3.4 or greater
sp = yt.SlicePlot(ds, "z", ("gas", "density"))
sp.set_font({"math_fontfamily": "dejavusans"})

# same result is expected here 
with mpl.rc_context({"mathtext.fontset": "dejavusans"}):
    sp = yt.SlicePlot(ds, "z", ("gas", "density"))
```
Expected result
3 4 3_mode_Slice_z_density

Note that this last case currently only works with matplotlib 3.4 + yt's main branch. It doesn't work with matplotlib 3.3 or earlier or with the current state of this PR, so I'm not 100% sure we _need_ to support it, but it would be nice if we could avoid closing a door that matplotlib just opened for us.

neutrinoceros:
So the image tests revealed two kinds of failures:
- line plots seem to not be properly affect by this patch (acutal bug)
- exponents in minor tick labels in colorbars for phase plots with logscale (weirdly specific...) are affected, but IMO they actually got better, so I think I'll propose to adopt this change

For illustration:
### main branch
Screenshot 2021-09-17 at 18 46 00


### this patch
Screenshot 2021-09-17 at 18 45 53

neutrinoceros:
I'm now down to 9 errors (previously 41) for GH answer tests. All remaining errors are the kind I described in the previous comment, so I believe it' is mature enough to warrant bumping the answer store.
On the other hand, some of the failures on Jenkins at least are real: tests that check wether font parameters can be changed from the outside are showing that this feature is broken, so I still need to work on that. I'll open a PR to the answer store for starters, and focus on solving real issues on Jenkins before I can ask to bump this baseline too if need.

neutrinoceros:
Here's the PR to the answer store https://github.com/yt-project/answer-store/pull/23
requesting early reviews from @matthewturk, @chummels, @Xarthisius regarding this point. Would it be okay to merge that PR before the present one is 100% finished or should I focus on getting things working on Jenkins first ?

neutrinoceros:
Alright I believe that all failures, GH and Jenkins alike, are the kind I propose to adopt. Opening this proposing to approvals before I can bump references.

neutrinoceros:
Note to reviewers: please consider reviewing this soon, the reports are not going to stay available for more than a couple days, after which we’d need to manually retrigger tests.

matthewturk (approved):
lgtm

neutrinoceros:
@yt-fido test this please

neutrinoceros:
I got "no old answer" twice on Jenkins after bumping the answer number. Either I don't understand how it's supposed to be done anymore or something isn't working. @Xarthisius, can you have a look ?

Xarthisius:
@yt-fido test this please

yt/visualization/profile_plotter.py
```diff
@@ -1336,77 +1337,6 @@ def save(self, name=None, suffix=".png", mpl_kwargs=None):
             self.plots[f].save(name, mpl_kwargs)
         return names
 
-    @invalidate_plot
-    def set_font(self, font_dict=None):
```
chummels, line 1340: Wait, so can we no longer manually set the font in our plots? I don't think i ever have, but it might be something people have used in the past.

neutrinoceros: No, it will still be possible, thanks to inheritance. Because I'm making `ProfilePlot` a subclass of `PlotContainer`, we don't need to have this method here, which was duplicated code from `PlotContainer` anyway :-)

chummels:
OK, I'm a bit confused based on the description of this PR and the comments what this PR actually does (e.g., two "expected results" plots)  Presumably this continues to use the old math font instead of the new MPL fonts for our plots.  can you confirm?

neutrinoceros:
> Presumably this continues to use the old math font instead of the new MPL fonts for our plots. can you confirm?

Yes, this PR forces matplotlib to use the "cm" math font family, which was its default before matplotlib 3.4.0

A side effect of doing so, using matplotlib 3.4's api, is that negative exponents in ticklables are also forced to use this font, something we apparently were not doing in yt already, but arguably should have. This is the reason that I have a "bump answer store" PR attached :)

chummels (approved):
OK, looks good to me.  Thanks!

yt/visualization/plot_container.py
```diff
@@ -248,19 +249,20 @@ class PlotContainer:
...
             self.figure_size = float(figure_size)
-        font_path = os.path.join(
-            matplotlib.get_data_path(), "fonts", "ttf", "STIXGeneral.ttf"
-        )
-        self._font_properties = FontProperties(size=fontsize, fname=font_path)
+
+        if sys.version_info >= (3, 9):
+            font_dict = DEFAULT_FONT_PROPERTIES | {"size": fontsize}
+        else:
+            font_dict = {**DEFAULT_FONT_PROPERTIES, "size": fontsize}
```
cphyc, line 263: Why aren't you using simply
```suggestion
        font_dict = {**DEFAULT_FONT_PROPERTIES, "size": fontsize}
```
I realize the syntax with `|` is only supported on Python 3.9 and above, but I am not sure the check about the python version is worth the trouble just to use a new syntax (and it increases the programmatic complexity).

cphyc (approved):

yt/visualization/plot_container.py
neutrinoceros: I got enthusiastic after I managed to patch pyupgrade so that it can cleanup this kind of structure when the time comes (https://github.com/asottile/pyupgrade/pull/530), and it so happens that `|` isn't just simpler than `**..., **...`, it's also faster.
Anyway I see this as basically "free technical debt" because we have the technology to automatically get rid of it later. Now that we're on the same page about that, let me know if you changed your mind. If not I'll simplify the code as you suggested :)

cphyc:
Thanks for the explanation, that's fine by me!

6 Oct 2021 17:40:49 Clément Robert ***@***.***>:

neutrinoceros:
Great, it's now down to a single error, same kind as the ones already bumps. Can I bump that one too ?

neutrinoceros:
alright, going straight to 009 then, thanks for the heads !

neutrinoceros:

Issue #3523, opened 2021-09-21
Heap buffer overflow in octree_build_node


saethlin:
This bug was discovered by @smlower

### Bug report

**Bug summary**

Heap buffer overflow when building an octree, which is eventually visible as generic memory allocator corruption.

It's unclear to me if a similar crash can be triggered by other files, or if the octree bounds need to be carefully chosen based on the properties of the snapshot.

http://use.yt/upload/79268ffc

```python
import yt

ds = yt.load('galaxy_600_constSFH.hdf5')

left = [23237.53979567, 21050.7424277, 11669.21951731]
right = [23917.53979567, 21730.7424277, 12349.21951731]

ds.octree(left, right, n_ref=32)
```

**Actual outcome**

Sidney mostly observed vague allocator errors like:
```
*** Error in `python': double free or corruption (!prev): 0x00005614b49fd5a0 ***
```
So I recompiled all of yt with asan+ubsan and turned on debug symbols to get a slightly more helpful error. Since this is a linear heap buffer overflow I feel like it should be possible to turn on bounds checks in Cython and get a more direct error message, but I tried that and only got another vague allocator error, instead of a clean exception. I suspect a more experienced operator can coax out a cleaner error message.

```
$ LD_PRELOAD=/usr/lib/libasan.so:/usr/lib/libubsan.so python crash_yt.py
yt : [INFO     ] 2021-09-21 16:10:54,441 Calculating time from 1.000e+00 to be 4.375e+17 seconds
yt : [INFO     ] 2021-09-21 16:10:54,442 Assuming length units are in kpc/h (comoving)
yt : [INFO     ] 2021-09-21 16:10:54,490 Parameters: current_time              = 4.3748565942961094e+17 s
yt : [INFO     ] 2021-09-21 16:10:54,490 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2021-09-21 16:10:54,490 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2021-09-21 16:10:54,490 Parameters: domain_right_edge         = [25000. 25000. 25000.]
yt : [INFO     ] 2021-09-21 16:10:54,490 Parameters: cosmological_simulation   = 1
yt : [INFO     ] 2021-09-21 16:10:54,490 Parameters: current_redshift          = 4.440892098500626e-16
yt : [INFO     ] 2021-09-21 16:10:54,490 Parameters: omega_lambda              = 0.7
[...]

PR #3524, opened 2021-09-22
BUG: Fix re-allocation of octree

AshKelly:
Fix failed re-allocation of octree reported in #3523 .

## PR Summary

There are strange edge cases where the creation of more nodes, some which are empty and some which need to be refined, can cause the total number nodes to overflow the maximum. By moving the check for re-allocation further up we ensure that re-allocation will always be performed.

## PR Checklist

### DIFF
--- a/yt/utilities/lib/cyoctree.pyx
+++ b/yt/utilities/lib/cyoctree.pyx
@@ -80,17 +80,17 @@ cdef int octree_build_node(Octree * tree, long int node_idx):
     cdef np.int64_t splits[9]
     cdef np.int64_t i, j, k, n, start, end
     cdef np.float64_t lx, ly, lz, sz, inv_size
+        
+    # If we are running out of space in our tree, then we *try* to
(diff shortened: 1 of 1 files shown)


matthewturk (approved):
Ah, this makes sense!

AshKelly:
pre-commit.ci autofix

neutrinoceros (approved):
I confirm that
1) I was able to reproduce the bug on my laptop
2) this patch fixes it

thanks @AshKelly

saethlin:
Could we have a patch release of yt available on PyPI with this fix in it? Someone else has run into this _because_ they upgraded to yt 4 via `pip install yt`.

matthewturk:
That sounds quite reasonable, and I believe with the changes @munkm made during the release of 4.0 we have a streamlined patch release setup now.  Thank you for the suggestion, @saethlin , and I'll make sure to ping this issue when we have it.

PR #3526, opened 2021-09-24
ENH: Particle reader cleanup

matthewturk:
This is supposed to be a no-op set of changes to the particle readers.  It is in support of #3511 , so that we can keep moving forward with it with minimal changes in the medium term and to lighten the review.

Ideally, if the tests pass, this should be able to go in -- we've been testing here locally but there may be unforeseen consequences.  The type annotations should be compatible with currently supported versions of python.

Future work will happen in another, manageable pull request, and will include moving some of the IO handling for particles into the `ParticleFile` subclasses, but largely keeping the API for the io handler the same, while eliminating duplicated code.


yt/data_objects/static_output.py
```diff
@@ -1879,7 +1880,13 @@ def _reconstruct_ds(*args, **kwargs):
 
 
 @functools.total_ordering
-class ParticleFile:
+class ParticleFile(abc.ABC):
+    filename: str
+    file_id: int
+    start: int = None
+    end: int = None
+    total_particles: ParticleTypeSizes = None
```
neutrinoceros, line 1888: These won't type check, you need to `from typing import Optional`
```suggestion
    start: Optional[int] = None
    end: Optional[int] = None
    total_particles: Optional[ParticleTypeSizes] = None
```
furthermore, it looks to me that the underlying logic can be simplified a lot in `__init__`. `start` doesn't need a `None` default value, the real default is `0`, right now it's equivalent but with extra steps.

yt/frontends/adaptahop/io.py
```diff
@@ -50,7 +50,7 @@ def _read_particle_coords(self, chunks, ptf):
             if pcount == 0:
                 continue
             pos = self._get_particle_positions()
-            yield ptype, [pos[:, i] for i in range(3)]
+            yield ptype, [pos[:, i] for i in range(3)], 0.0
```
neutrinoceros, line 53: This will break any third party particle frontend, I think. I don't know that there are any, and in any case it is trivial to mitigate but it needs to be mentioned in our next release notes, I think.

yt/frontends/art/io.py
```diff
@@ -85,7 +85,7 @@ def _read_particle_coords(self, chunks, ptf):
                 x = self._get_field((ptype, "particle_position_x"))
                 y = self._get_field((ptype, "particle_position_y"))
                 z = self._get_field((ptype, "particle_position_z"))
-                yield ptype, (x, y, z)
+                yield ptype, (x, y, z), 0.0
```
neutrinoceros, line 88: Looks like you haven't change the Base class here, but still the behaviour changes ?
Does this mean that _every_ third party frontend that implements `_read_particle_fields` is broken by this change ?

yt/frontends/gadget/data_structures.py
```diff
@@ -559,8 +559,12 @@ def _is_valid(cls, filename, *args, **kwargs):
         return header.validate()
 
 
+class GadgetHDF5File(ParticleFile):
+    pass
+
+
 class GadgetHDF5Dataset(GadgetDataset):
-    _file_class = ParticleFile
+    _file_class = GadgetHDF5File
```
neutrinoceros, line 567: I don't immediately get why this new hollow class is necessary

yt/frontends/swift/data_structures.py
```diff
@@ -10,10 +10,14 @@
...
 
+class SwiftParticleFile(ParticleFile):
+    pass
+
+
 class SwiftDataset(SPHDataset):
     _index_class = SPHParticleIndex
     _field_info_class = SPHFieldInfo
-    _file_class = ParticleFile
+    _file_class = SwiftParticleFile
```
neutrinoceros, line 20: same as above

yt/utilities/io_handler.py
```diff
@@ -12,6 +13,26 @@
 
 use_caching = 0
 
+FluidFieldType = str
+FluidFieldName = str
+FluidFieldTuple = typing.Tuple[FluidFieldType, FluidFieldName]
```
neutrinoceros, line 18: This feels pretty more complicated than necessary
```suggestion
FluidFieldTuple = Tuple[str, str]
```

yt/utilities/io_handler.py
```diff
@@ -12,6 +13,26 @@
...
+
+ParticleFieldType = str
+ParticleFieldName = str
+ParticleFieldTuple = typing.Tuple[ParticleFieldType, ParticleFieldName]
```
neutrinoceros, line 22:
```suggestion
ParticleFieldTuple = Tuple[str, str]
```
or actually, since there is no semantic difference between `ParticleFieldTuple` and plain `FieldTuple`, I think we should keep only the latter

yt/utilities/io_handler.py
```diff
@@ -12,6 +13,26 @@
...
+
+ParticleTypeFields = typing.Dict[ParticleFieldType, typing.List[ParticleFieldName]]
+ParticleTypeSizes = typing.Dict[ParticleFieldType, int]
+ParticleFieldSize = typing.Dict[ParticleFieldTuple, int]
+ParticleCoordinateSet = typing.Tuple[
+    str,
+    typing.Tuple[np.ndarray, np.ndarray, np.ndarray],
+    typing.Union[float, np.ndarray],
```
neutrinoceros, line 30: this line is probably not as general as necessary. I bet any real number (`from numbers import Real`) is valid, which includes numpy floating and integer scalar types. A List should probably be valid too where a np.ndarray works, so this single line should probably have a type of its own

yt/utilities/io_handler.py
```diff
@@ -12,6 +13,26 @@
...
+FluidFieldName = str
+FluidFieldTuple = typing.Tuple[FluidFieldType, FluidFieldName]
+
+ParticleFieldType = str
+ParticleFieldName = str
+ParticleFieldTuple = typing.Tuple[ParticleFieldType, ParticleFieldName]
+
+ParticleTypeFields = typing.Dict[ParticleFieldType, typing.List[ParticleFieldName]]
+ParticleTypeSizes = typing.Dict[ParticleFieldType, int]
+ParticleFieldSize = typing.Dict[ParticleFieldTuple, int]
```
neutrinoceros, line 26: some simplifications here too

neutrinoceros (commented):
Here's a first batch of comments and questions.
Mostly my two points are:
- I'm not confortable with calling this PR a "no-op" change in that it seems to be breaking any existing third party frontend that supports particles. The change is probably acceptable nonetheless, but we should keep this in mind. I could also be convinced that the change isn't breaking, but the fact that particle generators' signature changes is noteworthy.
- most custom types here feel uncalled for, but the essence of them is relevant. We can simplify this. Also please use `from typing import ...` rather than `import typing`. I makes for longer import statements but cleaner code that can be automatically upgraded to more modern annotation style later with `pyupgrade`

matthewturk:
> Here's a first batch of comments and questions.
> Mostly my two points are:
> 
> * I'm not confortable with calling this PR a "no-op" change in that it seems to be breaking any existing third party frontend that supports particles. The change is probably acceptable nonetheless, but we should keep this in mind. I could also be convinced that the change isn't breaking, but the fact that particle generators' signature changes is noteworthy.

OK, it's noteworthy.  Would it be sufficient to insert an unpacking for conditional creation of the default argument?

> * most custom types here feel uncalled for, but the essence of them is relevant. We can simplify this. Also please use `from typing import ...` rather than `import typing`. I makes for longer import statements but cleaner code that can be automatically upgraded to more modern annotation style later with `pyupgrade`

Ugh, I really don't like that notation, though.  Because it's all `typing` and `typing`-related stuff I much prefer having it explicitly namespaced -- *especially* since some will conflict with names from `traitlets`.  But, that's still a theoretical concern, so I suppose I can update it.

yt/data_objects/static_output.py
matthewturk: Hmm, they aren't optional though.  Is your suggestion still applicable?

matthewturk: Or should I leave them without initialization?

yt/frontends/adaptahop/io.py
matthewturk: Agreed!

yt/frontends/art/io.py
matthewturk: Yes, *however*, I did a quick check and we have extremely few if *any* that implement this, unless they are hidden somewhere.  Because this was introduced in yt-4.0, it should also be mitigated.  Would a wrapper around the functions via metaclass work for you?

yt/frontends/gadget/data_structures.py
matthewturk: That's because this was supposed to be largely no-op!  But upcoming changes will connect this with Swift.

Additionally, we want `ParticleFile` to be an ABC.

yt/utilities/io_handler.py
matthewturk: So I kind of disagree, there *is* a semantic difference -- there's no *technical* difference since they both resolve the same, but the key item is the difference between the meaning.  The thought process that led to [`NewType`](https://docs.python.org/3/library/typing.html#typing.NewType) is relevant here, but we can accomplish this without runtime checks.

neutrinoceros:
> Ugh, I really don't like that notation, though.

Here are two arguments to back up my concern:
- https://docs.python.org/3/library/typing.html has 0 occurrences of "import typing" and 12 occurrences of "from typing import" so it's pretty one-sided as far as the standard library is concerned
- unfortunately there isn't a single namespace that contains every relevant types we might want to import. `typing` is an important  player but there are also `types` and `collections.abc`. I don't think it would make type hints any clearer to use (and mix) namespaces everywhere.

yt/data_objects/static_output.py
neutrinoceros: `typing.Optional` strictly means that `None` is an acceptable value despite the explicit type being incompatible.
Omitting it is discouraged by PEP 484
>A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:
>```python
>def handle_employee(e: Employee = None): ...
>```
>This would have been treated as equivalent to:
>```python
>def handle_employee(e: Optional[Employee] = None) -> None: ...
>```
>This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

source:
https://www.python.org/dev/peps/pep-0484/#union-types

neutrinoceros: So please keep the default None value, but use `Optional` too, otherwise we'll need to change add it later when we perform static type checking.

yt/frontends/art/io.py
neutrinoceros: Since we haven't been promoting (yet) the recipe for third party frontends, it's indeed unlikely that this actually breaks anyone, so I don't think we need to change the way you're doing it, and a quick public announcement (an email to yt-dev) should be more than enough.

yt/frontends/gadget/data_structures.py
neutrinoceros: fair enough

yt/data_objects/static_output.py
neutrinoceros: you might want to try: https://pypi.org/project/auto-optional/

yt/utilities/io_handler.py
neutrinoceros: I didn't know about NewType, but I note that it actually doesn't make any sort of check or transform at runtime
```python
from typing import NewType

MyStr = NewType("MyStr", str)

def foo(a:MyStr):
    b = MyStr(a)
    print(type(a), type(b))
    print(f"{b!r}")

foo("example")
foo(1)
foo(None)
```
prints
```
<class 'str'> <class 'str'>
'example'
<class 'int'> <class 'int'>
1
<class 'NoneType'> <class 'NoneType'>
None
```
It doesn't seem relevant to use here either, however, since it would make type checking much much harder as we'd need to use these new types explicitly everywhere.

neutrinoceros: Anyway you're right that there's a semantic difference, I completely misused this word, sorry. Still, I am not convinced yet that introducing these type aliases for str is the right approach here and I'd rather do that later when we have more experience with typing and we can make an informed decision on what `ParticleFieldType` should be (a type alias ? a `Literal` type ? an actual subclass ?). I am trying to be particularly careful here since we'll be using typing completely in the dark as long as we don't turn on some kind of static type checker on PRs, but I don't think this can realistically happen overnight, so I'd rather keep type hints as simple as possible for now. I hope this sounds acceptable.

munkm:
> 
> Ugh, I really don't like that notation, though. Because it's all `typing` and `typing`-related stuff I much prefer having it explicitly namespaced -- _especially_ since some will conflict with names from `traitlets`. But, that's still a theoretical concern, so I suppose I can update it.

Do we have any explicit documentation in our dev guide requiring one or the other? If two seasoned developers are disagreeing about this, I think it could be a major issue for newcomers.

neutrinoceros:
No docs on our side atm. I also don't think we need any, `from typing import ...` already largely dominates usage in the Python ecosystem.

Taking the example of numpy :
- `import typing`: a single match, which happens to be an extremely special case in `numpy/typing/mypy_plugin.py`
- `from typing import`: 968 matches

or pandas:
- `import typing`: 5 matches
- `from typing import`: 790 matches

munkm:
Ok, but I think @matthewturk 's point about stuff conflicting with traitlets is a good one. Do you have a proposal for an alternative that would make sure it doesn't happen?

neutrinoceros:
Good point indeed.
From what I can find on traitlets, it is indeed problematic since they apparently recommend using `from` import statement as well, and have a couple conflicting names (`Tuple`, `List`, `Union`, `Dict` and `Any`, from what I can tell).
The good news is that we'll be able to reserve most of these names for traitlets as soon as we drop Python 3.6, since that will unlock `from __future__ import annotations`, making `typing.Tuple`, `typing.List`, `typing.Union` and `typing.Dict` not needed, as we'll be able to use builtin types instead. `typing.Union` also becomes unnecessary because we can use the "or" (`|`) operator instead. So in the end, the only remaining name collision is `Any` but that's probably a non-issue considering:
- the only use I know for `typing.Any` is to turn off type checking locally after requiring type hints everywhere, something that we are eons away from considering.
- we currently almost never use traitlets, so I don't expect we'd need `traitlets.Any` often enough that it could become a problem (or ever ?)

yt/data_objects/static_output.py
matthewturk: OK, I will think on this and work through the process of coming around to realizing you are correct... :)

yt/frontends/art/io.py
matthewturk: I was in a seminar today where @gvwilson talked about the "contracts" we have with software from a design perspective, and how we should only ever *relax* our specifications on what is input to a function and *strengthen* what is allowed to be output by a function.  This violates that principle, and I am thinking about it in a new way.  So let me think a bit.

yt/utilities/io_handler.py
matthewturk: > Anyway you're right that there's a semantic difference, I completely misused this word, sorry.

I apologize for the strength with which I corrected the usage.

> Still, I am not convinced yet that introducing these type aliases for str is the right approach here and I'd rather do that later when we have more experience with typing and we can make an informed decision on what `ParticleFieldType` should be (a type alias ? a `Literal` type ? an actual subclass ?). I am trying to be particularly careful here since we'll be using typing completely in the dark as long as we don't turn on some kind of static type checker on PRs, but I don't think this can realistically happen overnight, so I'd rather keep type hints as simple as possible for now. I hope this sounds acceptable.

Alright, I am coming around a tiny bit, but I still do want to make them as explicit as possible with as wordy of names for situations where it's unclear.  I do like the idea of communicating more information than just `str`.  But, it might not be the right thing to do, or it might be solving the problem in the wrong way.  So let's indeed make this part of a bigger discussion.

Do you think that we should allow this in, while marking an issue to revisit once the decision has been made, or revert to the simpler ones before it goes in?  My inclination is the former, not just because it aligns with my already-done-work, but because it would be easier to remove the information in the future than to re-add it.

matthewturk:
I think that `traitlets.Any` is indeed not something we should worry about.

yt/utilities/io_handler.py
neutrinoceros: > I apologize for the strength with which I corrected the usage.

no apologies needed, I genuinely appreciate being corrected on human language :-)

> Do you think that we should allow this in, while marking an issue to revisit once the decision has been made, or revert to the simpler ones before it goes in? My inclination is the former, not just because it aligns with my already-done-work, but because it would be easier to remove the information in the future than to re-add it.

I'd be far less worried about this adding technical debt if the custom types were defined in a "private" module, and on that condition I would get behind your preferred approach !

jzuhone (dismissed):

yt/frontends/art/io.py
chrishavlin: @matthewturk any further thoughts on this? 

I think if we didn't want to add the 0 `hsml` to all the `_read_particle_coords` methods, we could instead check the length of the returned tuple where needed? For example in `BaseParticleIOHanlder._count_particles_chunks` we have:

```python
            for ptype, (x, y, z), hsml in self._read_particle_coords(chunks, ptf):
                psize[ptype] += selector.count_points(x, y, z, hsml) 
```
but could do something along the lines of 

```python
            for coord_tuple in self._read_particle_coords(chunks, ptf):
                ptype = coord_tuple[0]
                x, y, z = coord_tuple[1]
                if len(coord_tuple) == 3:
                    hsml = coord_tuple[2]
                else:
                    hsml = 0.
                psize[ptype] += selector.count_points(x, y, z, hsml)
```

matthewturk:
We might be able to do star unpacking to accomplish this, but I'm not 100%
sure.  It might be just what you're looking at here, anyway.

chrishavlin:
> We might be able to do star unpacking to accomplish this, but I'm not 100% sure. It might be just what you're looking at here, anyway.

ya, I've been playing with a couple different ways, will see if star unpacking works. But I was curious on if you came down on either side of the philosophical fence in terms of do we make all `_read_particle_coords` return the same tuple or do we handle different outputs?

matthewturk:
I think if we put the function to sanitize the output somewhere else (can be anywhere, just not in the body of the loop) it should be fine to manage the different types of output.  I think.

yt/utilities/io_handler.py
chrishavlin: I've been working on outstanding suggestions on this PR. @neutrinoceros, in terms of possibly defining these custom types in a private module, now that `main` now has a `_typing` module, would you suggest we move these custom type definitions over there (after a merge)? 

In terms of how to actually resolve the custom type discussion: I don't feel strongly one way or the other... but one possibility would be to put in the simplifications suggested by @neutrinoceros but add some comments that add the context that @matthewturk and I had to work through to create these custom types. Meaning in this particular case:

```
ParticleFieldTuple = Tuple[str, str]  # (field type, field name)
```
or a less trivial case:

```
ParticleTypeFields = typing.Dict[str, typing.List[str]]  # maps a particle type to fields of that type
```

So that the typing remains minimal but we still explain the intent of these types?

neutrinoceros: > now that main now has a _typing module, would you suggest we move these custom type definitions over there (after a merge)?

yes


> one possibility would be to put in the simplifications suggested by @neutrinoceros but add some comments that add the context that @matthewturk and I had to work through to create these custom types. 

that would work for me. I don't want you guys to throw away the information you gathered while working on this.
Keep in mind that, while defining `FluidFieldType = str` is harmless to static type-checking, it's arguably more obscure from the user standpoint.
See for instance how, in the following example, my IDE represents the signature of a function I'm trying to use

Screenshot 2022-01-07 at 10 13 26

One can see the alias type is not resolved (note that it's exactly the same situation with alternative approaches using `NewType`, `Literal`,  or `TypeAlias`). It can be argued that `FluidFieldType` better communicates the meaning of the variable, but at the same time it obscures the actual expected type, which I think is counter productive for a type annotation. Another drawback to this approach is that it doesn't encourage using clear variable names.

TBH I don't fully understand the use cases for `NewType` or `TypeAlias` (and I'm almost convinced `Literal` is inappropriate here), so I don't think we should commit to any complexity greater than minimal at first.

chrishavlin: > It can be argued that FluidFieldType better communicates the meaning of the variable, but at the same time it obscures the actual expected type, which I think is counter productive for a type annotation.

Yes! I think we were using the typing to add info that really should be in the comments (or if refactoring were easier we might rename some variables to be more descriptive). I'm now even more on the side of simplifying the typing we added while adding comments to better explain the function arguments... will give it a shot, hopefully will update soon.

yt/frontends/art/io.py
chrishavlin: Note: we decided to keep the hsml in all the `_read_particle_coords` (the base class now reflects this change). Having the different behavior with variable outputs proved difficult with the type checking and overall more complicated, so I think making this change while there are few if any third party plugins to break is the way to go.

yt/frontends/enzo/io.py
```diff
@@ -61,7 +61,17 @@ def _read_exception(self):
         return (KeyError,)
 
     def _read_particle_coords(self, chunks, ptf):
-        yield from self._read_particle_fields(chunks, ptf, None)
+        yield from (
```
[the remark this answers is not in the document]
chrishavlin: thanks for that syntax suggestion, it is indeed MUCH more readable now!

yt/frontends/gadget/io.py
```diff
@@ -383,51 +378,45 @@ def _read_particle_coords(self, chunks, ptf):
...
-            for ptype, field_list in sorted(ptf.items()):
-                if tp[ptype] == 0:
-                    continue
-                if getattr(selector, "is_all_data", False):
-                    mask = slice(None, None, None)
+    def _read_particle_data_file(self, data_file, ptf, selector=None):
+        return_data = {}
+        poff = data_file.field_offsets
+        tp = data_file.total_particles
+        f = open(data_file.filename, "rb")
```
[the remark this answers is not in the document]
chrishavlin: Yes, definitely safer... but I left it as is for now because we have a follow-up PR coming that adds some better file context management and takes care of this in the process.

yt/utilities/io_handler.py
```diff
@@ -2,10 +2,11 @@
 from collections import defaultdict
 from contextlib import contextmanager
 from functools import _make_key, lru_cache
-from typing import Tuple
+from typing import DefaultDict, Iterator, List, Mapping, Tuple
```
neutrinoceros, line 5: I have no experience with using `Iterator` as a return type, and I can't find an existing example. Since the only place you're using this annotation is in a `NotImplemented` method, I'm not 100% sure I can trust mypy to reveal if you're using it wrong. Can you provide some additional insight ?

Also I note that importing it from typing is deprecated as of Python 3.9 (https://docs.python.org/3/library/typing.html#typing.Iterator) (this is also true for `Mapping`) so here's how to work around deprecation warnings:

```suggestion
import sys
from typing import DefaultDict, List, Tuple

if sys.version_info >= (3, 9):
    from collections.abc import Iterator, Mapping
else:
    from typing import Iterator, Mapping
```
further note that this -admittedly- bloated structure will be automatically cleaned up in the future when we require Python 3.9 thanks to pyupgrade, into:
```python
from collections.abc import Iterator, Mapping
```
 (see https://github.com/asottile/pyupgrade#python2-and-old-python3x-blocks)

chrishavlin: Well,  `_read_particle_coords` is a generator, so I was (somewhat blindly) just following the description in [`typing.Generator'](https://docs.python.org/3.6/library/typing.html#typing.Generator) for a `Generator` that only yields: you can either annotate with `Generator` using `SendType` and `ReturnType` set to `None` or just annotate with an `Iterator`, which is the route I went because it seemed clearer.

The main point is that we would want the object being yielded to resolve correctly. e.g., in a toy case using `Generator`:

```python
from typing import Generator, Tuple

def genthing() -> Generator[Tuple[int, int], None, None]:
    for thing in range(10):
        yield (thing, thing)

g = genthing()
print(type(g))  # <class 'generator'>
for gval in g:
    print(gval)
```
the hints correctly evaluate `g` as a `Generator` object and gval as `Tuple[int,int]`. e.g., in vscode:

Screenshot from 2022-01-11 12-15-26

Happy to use a `Generator` instead of `Iterator` if that's any clearer. 

As to whether mypy is actually testing this here... I'm not sure either. The child classes that implement this function inherit the typing characteristics though, don't they? They do at least evaluate in vscode and pycharm correctly, but I don't know enough about mypy to know if that actually helps. I can say that during my attempts to type hint when the SPH frontends were overriding `_read_particle_coords` to additionally yield the `hsml` variable I got a lot of confusing mypy errors related to "number of values to unpack" errors where `_read_particle_coords` is called, and those errors were fixed by normalizing all the `_read_particle_coords` methods to what they are now. 

I also did wonder if making `BaseIOHandler` a true abstract base class would help? But I wasn't sure about how much of a refactor that would lead to so held off on the attempt.

neutrinoceros: > Well, _read_particle_coords is a generator, so I was (somewhat blindly) just following the description in https://docs.python.org/3.6/library/typing.html#typing.Generator for a Generator that only yields

Thank you very much, I hadn't thought of looking there, and indeed this suits your use case.

> Happy to use a Generator instead of Iterator if that's any clearer.

no need, I was just uneducated. I agree now that `Iterator` is the appropriate annotation

> As to whether mypy is actually testing this here... I'm not sure either. The child classes that implement this function inherit the typing characteristics though, don't they?

Really not sure about that. Mypy doesn't seem to detect type inconsistencies that well through inheritance, see for instance

```python
from collections.abc import Iterator


class A:
    def foo(self) -> Iterator[int]:
        raise NotImplementedError

class B(A):
    def foo(self):
        for i in range(2):
            yield None # mypy doesn't detect the inconsistency here
```

that said, if I annotate the child method, mypy correctly errors in all cases that break the substitution principle. So annotating the base class is apparently not sufficient to maximise correctness, but it's a crucial first step (and we don't need to take this any further for the present PR)

> I also did wonder if making BaseIOHandler a true abstract base class would help? But I wasn't sure about how much of a refactor that would lead to so held off on the attempt.

I tried this approach on a toy example and it doesn't seem to make any difference as far as mypy is concerned. That said it would still make sense to do it because the class is _effectively_ abstract, so I think it's worth marking it as such, but that should happen in a different PR.

So in conclusion, nothing to do here, except maybe taking my suggestion above regarding imports

chrishavlin: Great! Will definitely put in the import suggestions.

neutrinoceros:
pre-commit.ci autofix

yt/frontends/adaptahop/io.py
neutrinoceros: going to add it to a draft release note for yt 4.1

neutrinoceros (approved):
I think all my concerns were addressed. Please consider my approval as sticky even if further change is needed.

jzuhone (approved):

neutrinoceros:
I went through all remaining threads and ended up closing them all since they were actually addressed already. Since we have 2 approvals I'll merge this.

PR #3530, opened 2021-09-27
ENH: rewrite backports from Python std lib in a fashion that can be automatically cleaned up with pyupgrade

neutrinoceros:
## PR Summary
Title.
With the latest version of pyupgrade (that we already use as part of our pre-commit hooks),
"versioned code branches" attached to specific versions of Python 3 will now be cleaned up automatically, making their technical debt effectively 0, but they have to include an `else` statement.

see https://github.com/asottile/pyupgrade/pull/530

### DIFF
--- a/yt/geometry/geometry_handler.py
+++ b/yt/geometry/geometry_handler.py
@@ -16,6 +16,24 @@ from yt.utilities.parallel_tools.parallel_analysis_interface import (
 )
 
 
+def cached_chunk_property(func):
+    n = f"_{func.__name__}"
+
+    def cached_func(self):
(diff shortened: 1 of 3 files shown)


neutrinoceros:
didn't expect this to break. Will fix later

neutrinoceros:
This won't work, the implementation we have for `cached_property` is *not* a back port of functools.cached_property and is not even compatible with it. Fixing this is not worth the required time for it atm, imo.

matthewturk:
If I am reading it correctly, I added cached property myself.  Should we just change its name to avoid ambiguity with stdlib?

neutrinoceros:
That would be a good idea IMO

Issue #3529, opened 2021-09-27
BUG: incorrect default plot limits in spherical coordinates


neutrinoceros:
### Bug report

**Bug summary**
Spherical data is incorrectly framed by default in SlicePlot and ProjectionPlot

**Code for reproduction**

```python
import yt

ds = yt.load_sample("bw_spherical_2d", unit_system="code")
p = yt.SlicePlot(ds, "phi", "density")
p.save("/tmp/raw.png")

# zooming out to reveal the actual extension of the domain
p.zoom(0.5)
p.save("/tmp/zoom_out.png")
```

**Actual outcome**

![raw](https://user-images.githubusercontent.com/14075922/134873845-9a75402d-2b7a-48f7-a1d1-cc5f661bc3ed.png)
![zoom_out](https://user-images.githubusercontent.com/14075922/134873852-c39b06ae-0d09-496b-a34c-cefb5e4f5e41.png)

**Expected outcome**
Vertical limits of the plot should include the whole domain. Horizontal limits also seem buggy here as the frame is *larger* than needed.

neutrinoceros:
Actually it's not just phi-normal slices that are broken, in fact spherical slicing is broken in all three directions.
For demonstration, here's a generalised version of my example above
```python
import yt

ds = yt.load_sample("bw_spherical_2d", unit_system="code")

for normal in ("theta", "phi", "r"):
    p = yt.SlicePlot(ds, normal, "density")
    p.save(f"/tmp/{normal}.png")
```
![theta](https://user-images.githubusercontent.com/14075922/135416056-8eb87be8-b0f0-424c-bf50-2fd505d1acb7.png)
expected: correct framing
![phi](https://user-images.githubusercontent.com/14075922/135416074-bbb3b29c-57ec-4194-88c2-08a8ae9c48dc.png)
expected: correct framing
![r](https://user-images.githubusercontent.com/14075922/135416156-84a9c3bd-38a0-47e7-9952-a6d55ab4b295.png)
expected: image should be rotated. Axes and their respective limits seem correct but every pixel is misplaced.

PR #3533, opened 2021-09-27
BUG: fix incorrect default limits for slice plots along curvilinear directions (phi, theta)

neutrinoceros:
## PR Summary
addresses #3529 for the phi and theta cases
TODO:
- [x] fix "phi" slices
- [x] fix "theta" slices
- [x] cleanup

Here's the result from #3529's script with this branch:
fix_2

### DIFF
--- a/yt/geometry/coordinates/spherical_coordinates.py
+++ b/yt/geometry/coordinates/spherical_coordinates.py
@@ -1,8 +1,15 @@
+import sys
+
 import numpy as np
 
 from yt.utilities.lib.pixelization_routines import pixelize_aitoff, pixelize_cylinder
 
 from .coordinate_handler import CoordinateHandler, _get_coord_fields, _unknown_coord
+
+if sys.version_info >= (3, 8):
+    from functools import cached_property
+else:
+    from yt._maintenance.backports import cached_property
 
 
 class SphericalCoordinateHandler(CoordinateHandler):
@@ -285,7 +292,8 @@ class SphericalCoordinateHandler(CoordinateHandler):
     def period(self):
         return self.ds.domain_width
 
-    def _get_poloidal_bounds(self):
+    @cached_property
+    def _poloidal_bounds(self):
         ri = self.axis_id["r"]
         ti = self.axis_id["theta"]
         rmin = self.ds.domain_left_edge[ri]
@@ -323,7 +331,8 @@ class SphericalCoordinateHandler(CoordinateHandler):
 
         return Rmin, Rmax, zmin, zmax
 
-    def _get_conic_bounds(self):
+    @cached_property
+    def _conic_bounds(self):
         ri = self.axis_id["r"]
         pi = self.axis_id["phi"]
         rmin = self.ds.domain_left_edge[ri]
@@ -375,12 +384,12 @@ class SphericalCoordinateHandler(CoordinateHandler):
         if name == "r":
             display_center = center
         elif name == "theta":
-            xxmin, xxmax, yymin, yymax = self._get_conic_bounds()
+            xxmin, xxmax, yymin, yymax = self._conic_bounds
             xc = (xxmin + xxmax) / 2
             yc = (yymin + yymax) / 2
             display_center = (xc, 0 * xc, yc)
         elif name == "phi":
-            Rmin, Rmax, zmin, zmax = self._get_poloidal_bounds()
+            Rmin, Rmax, zmin, zmax = self._poloidal_bounds
             xc = (Rmin + Rmax) / 2
             yc = (zmin + zmax) / 2
             display_center = (xc, yc)
@@ -398,12 +407,12 @@ class SphericalCoordinateHandler(CoordinateHandler):
         elif name == "theta":
             # Remember, in spherical coordinates when we cut in theta,
             # we create a conic section
-            xxmin, xxmax, yymin, yymax = self._get_conic_bounds()
+            xxmin, xxmax, yymin, yymax = self._conic_bounds
             xw = xxmax - xxmin
             yw = yymax - yymin
             width = [xw, yw]
         elif name == "phi":
-            Rmin, Rmax, zmin, zmax = self._get_poloidal_bounds()
+            Rmin, Rmax, zmin, zmax = self._poloidal_bounds
             xw = Rmax - Rmin
             yw = zmax - zmin
             width = [xw, yw]


neutrinoceros:
My current solution works for the exact dataset that I wrote the test for, however it is correct in the general case. What's missing is a way to calculate the min and max for elevation (z) and algebraic cylindrical radius R, composing only with left and right edges of the domain.

neutrinoceros:
Got a much more general solution for the phi-oriented case:

lol

Now there's the issue of duplicated code, but I'll deal with that when I get every case to work (including theta-oriented plots)

neutrinoceros:
I think I got the the right approach now

neutrinoceros:
While I was at it, I also fixed boxing in the "theta"-oriented case
```python
import yt

ds = yt.load_sample("bw_spherical_2d", unit_system="code")
p = yt.SlicePlot(ds, "theta", "density")
p.save("/tmp/raw.png")
```
main branch:

raw


this branch:
raw


Now I need to come back one more time to clean up this messy code before I can open this to review

neutrinoceros:
I'm having a rough time figuring out how to fix the "theta-normal" case. In my previous comment, I noted that axis ticks were x/y inverted, despite the width and centre being correctly computed and the overall result being correct (maybe not the aspect, but well, that would be another sign of a single problem).
I think it way have something to do with :
- the fact that `SphercialCoordinateHandler._sanitize_center` returns a tuple (`display_center`) with 3 elements instead of 2, the second one being completely unimportant except for the fact that something else downstream looks for a third element
- the definitions of `SphericalCoordinateHandler._x_pairs` (`.y_pairs` respectively) which don't look easy to retro-engineer. I think they are supposed to be "right-handed" and "left-handed" direction pairs but that assumption doesn't hold, and I really don't get how they are currently written.
@matthewturk , you seem to be the most recent contributor here, any chance you would spot something terribly wrong here ? (or maybe provide some insight on how those pairs are defined ?)
https://github.com/yt-project/yt/blob/c531b73cac1fef014f936ce6e898d4c61c80dc82/yt/geometry/coordinates/spherical_coordinates.py#L273

neutrinoceros:
I found a pretty ugly hack that fixes the issue:
theta

I'm really not found of the current solution as a maintainer, but it seems valuable (maybe not *viable*) if it works. On the user side I don't have any complaints left so I'll open this to review and hope for the best, maybe we can sort this out.

neutrinoceros:
Alright, figured out the last (fatal) issue I had with my Idefix dataset came from some floating point truncation in a case where the domain right edge in the azimuthal direction was supposed to be 2 PI, so I added a small tolerance margin, large enough for single precision

Here's the plot for this dataset with yt's main branch
good_boy

and this branch
good_boy

neutrinoceros:
So I've simplified this as much as I could but I wasn't able to fix the r-case. Still this PR fixes 2/3 of the problematic cases reported in #3529 so I'm going to open it for review.

yt/geometry/coordinates/spherical_coordinates.py
```diff
@@ -158,6 +158,14 @@ def pixelize(
                 data_source, field, bounds, size, antialias, dimension, periodic
             )
         elif name in ("theta", "phi"):
+            if name == "theta":
+                # This is admittetly a very hacky way to resolve a bug
```
matthewturk, line 162:
```suggestion
                # This is admittedly a very hacky way to resolve a bug
```

yt/geometry/coordinates/spherical_coordinates.py
```diff
@@ -158,6 +158,14 @@ def pixelize(
...
+                # it's very likely that the *right* fix would have to
+                # be applied upstream of this function, *but* this case
+                # has never worked properly so maybe it's still preferable to
+                # not having a solution ?
+                # see https://github.com/yt-project/yt/pull/3533
+                bounds = (*bounds[2:4], *bounds[:2])
```
matthewturk, line 168: Your comment makes me think that we need to have a discussion about moving some of these functions around a bit.  It seems that you've likely already come up to speed on what everything does and would be a good person to speculate about potential refactorings that touch the coordinate handlers.  Could we set up a time to chat?

yt/geometry/coordinates/spherical_coordinates.py
```diff
@@ -277,27 +285,105 @@ def image_axis_name(self):
...
+        thetamin = self.ds.domain_left_edge[ti]
+        thetamax = self.ds.domain_right_edge[ti]
+        corners = [
+            (rmin, thetamin),
+            (rmin, thetamax),
+            (rmax, thetamin),
+            (rmax, thetamax),
+        ]
+
+        def to_poloidal_plane(r, theta):
```
matthewturk, line 302: should this be defined externally?  or as a method?

yt/geometry/coordinates/spherical_coordinates.py
```diff
@@ -277,27 +285,105 @@ def image_axis_name(self):
...
+        thetamax = thetamax.d
+
+        Rmin = min(R for R, z in cylindrical_corner_coords)
+
+        if thetamin <= np.pi / 2 <= thetamax:
+            Rmax = rmax
+        else:
+            Rmax = max(R for R, z in cylindrical_corner_coords)
+
+        zmin = min(z for R, z in cylindrical_corner_coords)
```
matthewturk, line 321: Would it be possible to consolidate this and the `Rmin` calculation into a set of by-axis min/max numpy call?

yt/geometry/coordinates/spherical_coordinates.py
```diff
@@ -277,27 +285,105 @@ def image_axis_name(self):
...
         name = self.axis_name[axis]
         if name == "r":
             display_center = center
         elif name == "theta":
-            display_center = (
-                0.0 * display_center[0],
-                0.0 * display_center[1],
-                0.0 * display_center[2],
-            )
+            xxmin, xxmax, yymin, yymax = self._get_conic_bounds()
```
matthewturk, line 378: Since the results of this function never change, couldn't we pre-compute them at initialization and make them properties?  That would work well with a baseclass-level attribute.

matthewturk (commented):
Thank you for doing this.  Some comments...

yt/geometry/coordinates/spherical_coordinates.py
neutrinoceros: Actually I went with a very dumb approach of touching everything in a sensible way until something worked. Now I wouldn't say I was able to understand most of it, and I don't think I would be in a position to champion such a refactor, but _yes_, let's set up a time to discuss that over zoom :)

yt/geometry/coordinates/spherical_coordinates.py
neutrinoceros: I wrote it as a method initially but it felt a little awkward since it's really only meant as a small helper function. Because its use is so narrow I figured I might as well keep it in the only place it's ever going to be used so as to keep the code tidier. What would be the benefits of keeping it separated ?

yt/geometry/coordinates/spherical_coordinates.py
neutrinoceros: do we need to have them exposed ? I feel like bare hidden attributes would do the job just as well.

neutrinoceros: Turns out it's not that trivial to run the computation once and for all at init because at that moment, `self.ds.domain_left_edge` is a pure numpy.ndarray instead of a unyt_array, which then causes issue downstream breaking the plot. I don't think I would be able to clean this up without a much deeper refactor.

neutrinoceros: and to be clear I'm not proposing that we do such a refactor, or at least not in this PR.

yt/geometry/coordinates/spherical_coordinates.py
neutrinoceros: hum. Actually, and this may be fatigue speaking but it seems to me that turning these into numpy operations would make the code a lot harder to read and maintain for a very negligible performance gain.

yt/geometry/coordinates/spherical_coordinates.py
matthewturk: Oh, yeesh, yes, OK.  Good point.

yt/geometry/coordinates/spherical_coordinates.py
matthewturk: Sounds good!

neutrinoceros:
@matthewturk I've backported `functools.cached_property` from Python 3.8 to help mitigate your concerns about running the exact same computations multiple times.
I also added a test but I'm not 100% convinced it will check what I want.

neutrinoceros:
@yt-fido test this please

matthewturk:
Just to be clear, my concerns were not articulated as well as I'd like. I am not really concerned about the performance impact as much as just the conceptual nature of it being a singly-calculated item that could be done just once.

neutrinoceros:
yup, agreed that it didn't make much sense. Now, as a bonus, we'll be able to start using cached_property before we even drop python 3.7, yay.

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

Issue #3535, opened 2021-09-30
BUG: symlog colormap is not drawn properly


neutrinoceros:
Taking this away from #3229 where it was initially reported in the discussion, because it's a separate issue that I think I can fix independently from #3229

```python
import numpy as np
import yt

d = {'my_fake_density': np.random.random(size=(64,64,64)) - 0.5}

ds = yt.load_uniform_grid(d, [64,64,64])
p = yt.SlicePlot(ds, "z", ('stream', 'my_fake_density'))
p.set_log(('stream', 'my_fake_density'), True, linthresh=1.e-3)
p.set_zlim(('stream', 'my_fake_density'), -1, 1)
p.save()
```

![UniformGridData_Slice_z_my_fake_density](https://user-images.githubusercontent.com/89019/117355884-b9b6b700-ae78-11eb-964d-02950a1453ff.png)

_Originally posted by @matthewturk in https://github.com/yt-project/yt/issues/3229#issuecomment-833807567_

neutrinoceros:
So I found that an easy patch for this issue is to simply turn off ticks drawning for symlog colorbars, for instance commenting this line
https://github.com/yt-project/yt/blob/c531b73cac1fef014f936ce6e898d4c61c80dc82/yt/visualization/plot_window.py#L1202
I obtain 
![symlog](https://user-images.githubusercontent.com/14075922/135499864-77fa538c-bd0f-4c48-9b4f-bfdb014efa81.png)
Now this looks like a matplotlib bug, so I'll try to make sure it can be reproduced without yt, report it if necessary, and maybe open a PR with the suggested patch here to mitigate the issue depending on what I find.

neutrinoceros:
The breaking change in MPL bisects down to this commit
https://github.com/matplotlib/matplotlib/commit/dc51a68352fd5f1a209c55f2ffb7bdc6f7728159

neutrinoceros:
I was finally able to reproduce in pure MPL. Will report when I have a true minimal example, and I'll link the report in a comment in yt

neutrinoceros:
reported upstream https://github.com/matplotlib/matplotlib/issues/21258
🎉

PR #3536, opened 2021-09-30
BUG: fix colorbar minor ticks drawing for symlog norms

neutrinoceros:
## PR Summary

This is the patch I suggested in #3535
As far as I can tell the bug is very likely to be in Matplotlib itself, though I wasn't able to replicate it in isolation for now.
I'm opening this a a draft while I attempt to show that this buggy behaviour can be reproduced without yt, in which case I'll report upstream and link the report in a comment.

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -167,7 +167,7 @@ answer_tests:
   local_axialpix_007:
     - yt/geometry/coordinates/tests/test_axial_pixelization.py:test_axial_pixelization
 
-  local_cylindrical_background_007:  # PR 3295
+  local_cylindrical_background_008:  # PR 3536
     - yt/geometry/coordinates/tests/test_cylindrical_coordinates.py:test_noise_plots
 
   #local_particle_trajectory_001:
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -280,7 +280,7 @@ class ImagePlotMPL(PlotMPL):
                     10
                     ** np.arange(
                         np.rint(np.log10(cblinthresh)),
-                        np.ceil(np.log10(np.nanmax(data))) + 1,
+                        np.ceil(np.log10(np.nanmax(data))),
                     )
                 )
             elif np.nanmax(data) <= 0.0:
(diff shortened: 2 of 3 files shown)


matthewturk (approved):
Good catch.  Assuming you move this to non-draft, LGTM.

neutrinoceros:
So I'm actually going to change this to a slightly less aggressive form. I still don't have a clear matplotlib-isolated example but I'm now certain that the bug is on their side and was introduced in matplotlib 3.2.0
Here's the results of your script from #3535 using two different combinations of versions:
 
### yt 4.0.1 + matplotlib 3.2.0
symlog

### yt 4.0.1 + matplotlib 3.1.3 (latest version before 3.2.0)
symlog

neutrinoceros:
Making a matplolitb-only example of this bug turns out to be much more challenging than I would have guessed, so I won't be able to report it properly for now, but I can at least avoid disable minor ticks drawing with Matplotlib < 3.2, so I made the skipping conditional. Hopefully the comment is clear enough for the next person to inquire this in the future.
There's exactly one failing test with my change, and I think we can agree that the new version is preferable to the current baseline.
See
https://tests.yt-project.org/job/yt_py38_git/3861/testReport/junit/yt.geometry.coordinates.tests/test_cylindrical_coordinates/GenericImage_AMRGridData_all_create_image_None_None/

@matthewturk, do I have your approval to bump this baseline ?

neutrinoceros:
@chummels you might want to review this too

neutrinoceros:
thanks to a discussion with matplotlib maintainers https://github.com/matplotlib/matplotlib/issues/21258#issuecomment-932699643, I got an alternative solution for the upcoming matplotlib 3.5 that almost works.
symlog
So now there's some useless whitespace at the top of the colorbar, but I believe this is a bug in yt, so now let's try to fix this as well.

neutrinoceros:
fixed that last issue in https://github.com/yt-project/yt/pull/3536/commits/e317c03e54b67137d9aa01b9acf7aa565078777d
The commit has no effect with current versions of matplotlib but fixes the issue with the upcoming matplotlib 3.5 noted in my previous comment

chummels (approved):
Looks good.  Thanks for digging into this and finding a workable solution!

PR #3538, opened 2021-10-04
BUG: fix set_zlim method with symlog norm

### LINKED ISSUE #3229: set_zlim does not work for symlog scaling
### Bug report

**Bug summary**

`set_zlim` seems to not work when applying limits to symlog scalings on plots.  This is using tip of dev.

**Code for reproduction**

```python
import yt
ds = yt.testing.fake_random_ds(3)
p = yt.SlicePlot(ds, "z", ('gas', 'density'))
p.set_log(('gas', 'density'), True, linthresh=1.e-3)
p.set_zlim(('gas', 'density'), -1, 1)
p.save()
```

**Actual outcome**

UniformGridData_Slice_z_velocity_x

neutrinoceros:
## PR Summary
fix #3229

```python
import numpy as np
import yt

d = {'my_fake_density': np.random.random(size=(64,64,64)) - 0.5}

ds = yt.load_uniform_grid(d, [64,64,64])
p = yt.SlicePlot(ds, "z", ('stream', 'my_fake_density'))
p.set_log(('stream', 'my_fake_density'), True, linthresh=1.e-3)
p.set_zlim(('stream', 'my_fake_density'), -1e-3, 1e5)
p.save()
```
### on main

zlim_symlog_main
This plot is identical before/after calling the `set_zlim` method.

### this branch
zlim_symlog

## PR Checklist


neutrinoceros:
requesting @chummels for review on this one :)

yt/visualization/base_plot_types.py
```diff
@@ -275,35 +273,40 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
...
+                yticks = [zmin] + list(
                     10
                     ** np.arange(
                         np.rint(np.log10(cblinthresh)),
-                        np.ceil(np.log10(np.nanmax(data))),
+                        np.ceil(np.log10(zmax)),
                     )
                 )
-            elif np.nanmax(data) <= 0.0:
+                print("branch 1")
```
chummels, line 287:
```suggestion
```

yt/visualization/base_plot_types.py
```diff
@@ -275,35 +273,40 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
...
+                                np.floor(np.log10(-zmin)),
                                 np.rint(np.log10(cblinthresh)) - 1,
                                 -1,
                             )
                         )
                     )
-                    + [np.nanmax(data).v]
+                    + [zmax]
                 )
+                print("branch 2")
```
chummels, line 302:
```suggestion

```

yt/visualization/base_plot_types.py
```diff
@@ -314,10 +317,11 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
                         10
                         ** np.arange(
                             np.rint(np.log10(cblinthresh)),
-                            np.ceil(np.log10(np.nanmax(data))),
+                            np.ceil(np.log10(zmax)),
                         )
                     )
                 )
+                print("branch 3")
```
chummels, line 324:
```suggestion
```

chummels (approved):
Thank you!  I tested this out both with your example as well as my example from Issue #3229 , and they both behave correctly now.  I just included some suggestions to remove extra debug print statements.  

The only thing that I'll note is that the zmax doesn't appear to show up in as a tickmark/ticklabel in either of the examples.  I imagine it could be fixed by slightly bumping the zmax by some epsilon value upward (zmax * 1.01).  But other than that this is a great fix.

neutrinoceros:
> The only thing that I'll note is that the zmax doesn't appear to show up in as a tickmark/ticklabel in either of the examples.

that sounds like a reasonable request, however, I anticipate that it's be very hard to implement this in a way that would work consistently with every matplotlib versions we want to support, as they have been changing the way colorbars work a couple times between 3.0 and 3.5

chummels:
But won't my option of just slightly increasing the zmax work regardless of the version? It seems like we just want to have the tick label, so making zmax exceed the tick marker value should work for any MPL version.

chummels:
Maybe bumping down zmin by a similar amount would be wise too? Just to ensure it always has a tick label as well.

neutrinoceros:
I've seen similar tricks turn into maintenance puzzles with every recent matplotlib release. Matplotlib itself may be the place where this needs fixing. In any case I feel very uncomfortable to introduce colorbar tricks in Yt when matplotlib 3.5 is coming up with a complete colorbar overhaul.

matthewturk:
I'm a little uncertain about removing the nanmax's in favor of straight max's, but I think it looks good.

matthewturk:
I've marked this for merging once the conflicts are done.

chummels:
I agree with @neutrinoceros that there are no removed nanmaxes, as it's just defaulting to either nanmax or the value provided by the user (which will not be a nan).

chummels:
I don't see what conflicts exist or need to be addressed before this gets merged.

matthewturk:
> I don't see what conflicts exist or need to be addressed before this gets merged.

`yt/visualization/base_plot_types.py`

neutrinoceros:

neutrinoceros:
Let's try this one last time now that a lot of other PRs went in. If it fails we'll resort to manual backport

PR #3540, opened 2021-10-04
TST: add answer testing to bleeding edge CI

neutrinoceros:
## PR Summary
matplotlib 3.5 is about to be released, and answer tests are clearly much better at catching integration bugs, so it seems like agood opportunity to add our answer test suite to the "bleeding edge" CI workflow. Hopefully we don't catch anything weird. Otherwise, matplotlib's maintainers are on the lookout so we still have a chance to report stuff before the actual release is out.

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -17,8 +17,16 @@ on:
 
 jobs:
   build:
+    strategy:
+      matrix:
+        include:
+          - tests-type: unit
+            python-version: '3.10-dev'
+          - tests-type: answer
+            python-version: 3.7
+
     runs-on: ubuntu-latest
-    name: Python3.10-dev
+    name: Future Env
     timeout-minutes: 60
 
     concurrency:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
This work is now finished. I only got two non-null diffs in image comparison tests, which both look like this
Screenshot 2021-10-05 at 07 14 04

So we should be able to bump these answers and then merge this PR.
However #3520 is set to bump the very same images for different reasons, so the present works should wait for _that_ PR to resolve first. I'll keep it as a draft in the mean time.

neutrinoceros:
We also have good reasons to think that we'll need to bump images using symlog colorbars after matplotlib 3.5 drops, since it should re-enable minor ticks drawing. The present PR should not be considered for merging before that.

neutrinoceros:
Changed my mind, I don't think this is a good idea anymore: catching subtle changes like the one above well ahead of any matplotlib release will not help much, and I fear that subtle, unimportant changes are far more common than the ones we'd want to see early. In general doing this will increase maintenance cost but the expected gain isn't clear. Closing.

Issue #3541, opened 2021-10-04
Sample notebook error


zhafen:
### Bug report

**Bug summary**

As of this moment, the sample notebook hosted at https://yt-project.org/doc/examining/generic_particle_data.html appears to be failing with `RuntimeError: length_unit (pc) is invalid.`
It's the first google result for "yt particle data", so it's a relatively visible failure.

No code for reproduction necessary. I'm referring to the html page generated and hosted at the above link.

neutrinoceros:
I'm going to look into this, sorry it's taken so long.
So, we already know that the first error is easily fixed thanks to #3542, but there is at least one other problem that seems harder to track down.
This failure seems to go so far back in our history that I can't even git bisect it easily as I'm stumbling on pip/cython errors using present-day versions.
At the moment the only point in time that I was able to get a stable run for is 42c037dc

neutrinoceros:
So I tried again to bisect this problem using `git bisect skip` for commits that I cannot compile on my laptop but the process is daunting. Here's what I got so far
<details> <summary> git bisect log </summary>

```
git bisect start
# bad: [564450aa2b0efe5195811b11c533561f4926adca] Merge pull request #3831 from neutrinoceros/migrate2tomli
git bisect bad 564450aa2b0efe5195811b11c533561f4926adca
# bad: [2c1c46b686155a2c337b1ac97240e8e869c5d28f] Merge pull request #3551 from yt-project/pre-commit-ci-update-config
git bisect bad 2c1c46b686155a2c337b1ac97240e8e869c5d28f
# good: [42c037dce30e37010792511c039e3d9c5c50e68e] Merge branch 'master' of github.com:yt-project/yt
git bisect good 42c037dce30e37010792511c039e3d9c5c50e68e
# bad: [60fd5f31a4f36399e35a7260162afe554a97d878] Merge pull request #2696 from Xarthisius/amrvac_test_ds
git bisect bad 60fd5f31a4f36399e35a7260162afe554a97d878
# bad: [7dae7a31b2675501596218d73924517de5975d4e] Fix the species list
git bisect bad 7dae7a31b2675501596218d73924517de5975d4e
# skip: [43237a90787f9a6ff122f8efcaef6d7f61eb0082] reimplemeneted richcmp operator overload
git bisect skip 43237a90787f9a6ff122f8efcaef6d7f61eb0082
# skip: [43237a90787f9a6ff122f8efcaef6d7f61eb0082] reimplemeneted richcmp operator overload
[...]

PR #3542, opened 2021-10-04
BUG: Fix issue 3541

cphyc:
This fixes issue #3541 by correcting the error in the notebook.

### DIFF
--- a/doc/source/examining/Loading_Generic_Particle_Data.ipynb
+++ b/doc/source/examining/Loading_Generic_Particle_Data.ipynb
@@ -2,20 +2,16 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {},
    "source": [
     "This example creates a fake in-memory particle dataset and then loads it as a yt dataset using the `load_particles` function.\n",
     "\n",
     "Our \"fake\" dataset will be numpy arrays filled with normally distributed randoml particle positions and uniform particle masses.  Since real data is often scaled, I arbitrarily multiply by 1e6 to show how to deal with scaled data."
-   ]
+   ],
+   "metadata": {}
   },
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": false
-   },
-   "outputs": [],
    "source": [
     "import numpy as np\n",
     "\n",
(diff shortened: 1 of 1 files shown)


doc/source/examining/Loading_Generic_Particle_Data.ipynb
```diff
@@ -24,84 +20,84 @@
...
-   "outputs": [],
    "source": [
     "import yt\n",
     "from yt.units import parsec, Msun\n",
     "\n",
     "bbox = 1.1*np.array([[min(ppx), max(ppx)], [min(ppy), max(ppy)], [min(ppz), max(ppz)]])\n",
     "\n",
-    "ds = yt.load_particles(data, length_unit=parsec, mass_unit=1e8*Msun, bbox=bbox)"
-   ]
+    "ds = yt.load_particles(data, length_unit=1*parsec, mass_unit=1e8*Msun, bbox=bbox)"
```
matthewturk, line 73: I honestly find it surprising that this fixed it!  I did not think the old way would be an error.  Does it also work to use `length_unit = "pc"`?

matthewturk (approved):

doc/source/examining/Loading_Generic_Particle_Data.ipynb
cphyc: Yup it works with "pc". The nicer fix may actually be for `load_particles` to accept units instead of quantities, by implicitly converting `parsec` to `1*parsed`. Wdyt?

cphyc: The bad news is that now the next cell crashes

cphyc: @AshKelly or any SPH-dataset guru to the rescue!

chummels:
It appears to me that this is an old notebook that was generated pre-yt4.0.  So while it's creating a particle-based dataset, it's then trying to access deposited fields, which I believe are no longer done by default in yt4, so it crashes.  It's also using `n_ref` and such, which are deprecated as of yt4.  Probably the best thing to do here is to clean this up to use the new (simpler) way of plotting/accessing particle-based fields in yt4 (i.e., not deposited fields).

cphyc:
@chummels that's also my understanding. If you want to have a stab at updating it in a way that makes more sense, feel free to override this PR. It have little to no experience with particles and yt, so I am not sure I'll promote good practices!

neutrinoceros:
closing + reopening to rettriger CI and enable merging if we want to do that later

neutrinoceros:
Closing this on behalf of @cphyc

PR #3550, opened 2021-10-11
Restrict sympy to 1.8

Xarthisius:
See https://github.com/sympy/sympy/issues/22241


neutrinoceros:
There's one failure remaining, which I believe is only due to an unfortunate chronology with merging #3520 and #3533 
the actual result is correct, @Xarthisius, can you bump ?

neutrinoceros:
We need this to resolve issues on the 4.0.x branch

Issue #3537, opened 2021-10-01
viz: Should a linear scaling be used automatically when the value range is narrow ?


neutrinoceros:
Trying to fix #3229 makes me think that it could be an opportunity to make norm selection a little more clear, and in particular it seems like a good time to propose a new rule.
So here's an example to illustrate how our current scale selection rules can be damaging for data readability
```python
import numpy as np
import yt

d = {'my_fake_density': np.random.random_integers(1, 3, size=(64,64,64))}

ds = yt.load_uniform_grid(d, [64,64,64])
p = yt.SlicePlot(ds, "z", ('stream', 'my_fake_density'))
p.set_log(('stream', 'my_fake_density'), True, linthresh=1.e-3)
p.set_zlim(('stream', 'my_fake_density'), -1, 1)
p.save()
```
output:
![UniformGridData_Slice_z_my_fake_density](https://user-images.githubusercontent.com/14075922/135618002-adef4a03-72a2-4e20-afcb-277353e2afdb.png)
The default colorscale is actually pretty obscure because a log norm isn't appropriate here.
(note: I've used my branch #3536 to produce this picture to avoid distraction from an unrelated colorbar bug)
<details> <summary>Using yt's main branch</summary>

![UniformGridData_Slice_z_my_fake_density](https://user-images.githubusercontent.com/14075922/135618140-3cf9bf0f-bc12-439d-9584-7413ad158149.png)

</details>
)

As an easy fix to the issue above, I'm proposing that we use linear scaling by default instead of log *on the condition that the value range is below 2 orders of magnitude*. I believe this can be achieved at the same time as a fix to #3229.

Of course this criterion isn't sufficient and it would also make sense to compare the variance of the image VS the variance of its log, but I don't really like the idea of piling up layers of magic here, as the existing norm-selection mechanism is already proving difficult to maintain against a large range of matplotlib versions.
In an ideal world, we should expose the color norm completely so all of matplotlib's flexibility is available to our users.
I would like to do that in the future, but for now I want to focus on having sensible defaults.

neutrinoceros:
arguably the example in the OP is terribly malformed, I don't think it's illustrating the point correctly now so I'd rather close it and maybe reopen when things stabilise with matplotlib 3.5

PR #3565, opened 2021-10-13
BUG: fix for matplotlib 3.5 + symlog + linthresh value outside data range

### LINKED ISSUE #3564: BUG: symlog colorbar with linear threshold < vmin + matplotlib 3.5
### Bug report

**Bug summary**

**Code for reproduction**

```python
import matplotlib
import numpy as np
import yt

prng = np.random.RandomState(0)
d = {'my_fake_density': prng.random_integers(1, 3, size=(64,64,64))}

ds = yt.load_uniform_grid(d, [64,64,64])
p = yt.SlicePlot(ds, "z", ('stream', 'my_fake_density'))
p.set_log(('stream', 'my_fake_density'), True, linthresh=1.e-3)
p.save(f"/tmp/mpl_{matplotlib.__version__}_")
```

**Actual outcome**
mpl_3 5 0rc1__Slice_z_my_fake_density

**Expected outcome**
This is the result of the same script, using matplotlib 3.4.3
mpl_3 4 3__Slice_z_my_fake_density

Arguably, the behaviour with MPL 3.4 can be considered canonical, but maybe we need to add a warning when the value of `linthresh` is under the actual minimum value.

I note that this reference behaviour,  setting the lower limit of the colobar to the data's minimum rather than the user required linthresh, is in line with what can be obtained in both matplotlib 3.4 and 3.5 without yt:
```python
from matplotlib.colors import SymLogNorm
import matplotlib
import numpy as np

prng = np.random.RandomState(0)
data = prng.random_integers(1, 3, size=(64,64))

fig, ax = plt.subplots()
im = ax.imshow(data, norm=SymLogNorm(linthresh=1e-2))
cb = fig.colorbar(im)
file = f"/tmp/threshold_mpl_{matplotlib.__version__}.png"
fig.savefig(file)
```
gives
threshold_mpl_3 4 3

neutrinoceros:
## PR Summary
fix #3564

Here's the results from #3564's script with this branch with matplotlib 3.4 or 3.5
mpl_3 5 0rc1__Slice_z_my_fake_density

So we even gained minor ticks for matplotlib 3.4, which alleviate most of the problem I reported in #3537 🎉 
opening as a draft for now, will open for review if CI stays green

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -1,3 +1,4 @@
+import warnings
 from io import BytesIO
 
 import matplotlib
@@ -12,7 +13,7 @@ from yt.funcs import (
     mylog,
 )
 
-from ._commons import get_canvas, validate_image_name
+from ._commons import MPL_VERSION, get_canvas, validate_image_name
 
 BACKEND_SPECS = {
     "GTK": ["backend_gtk", "FigureCanvasGTK", "FigureManagerGTK"],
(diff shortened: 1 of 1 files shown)


neutrinoceros:
switching to draft while we're stabilising the auto backport system

neutrinoceros:

PR #3567, opened 2021-10-14
DEPR: derpecate HaloCatalogCallback, let it migrate to yt_astro_analysis

### LINKED ISSUE #3566: DEPR: deprecate annotate_halos (HaloCatalogCallback)
As soon as [PR #98](https://github.com/yt-project/yt_astro_analysis/pull/98) in yt_astro_analysis is accepted, the HaloCatalogCallback should be deprecated or removed. This will get rid of the need to import yt_astro_analysis inside yt itself.

neutrinoceros:
## PR Summary

complementary part of https://github.com/yt-project/yt_astro_analysis/pull/98
addressing https://github.com/yt-project/yt_astro_analysis/issues/93
and fix #3566
This should probably not go in before yt_astro_analysis 1.1 is released, see https://github.com/yt-project/yt_astro_analysis/milestone/2


matthewturk:
@brittonsmith I think this needs your stamp of approval

neutrinoceros:
Checking for this helped me find a bug in yt_astro https://github.com/yt-project/yt_astro_analysis/pull/131

With the associated bugfix + this branch, I've checked that halo plotting still works as documented

brittonsmith:
I've thought about this a bit and I've come back to the opinion that it would be good to have some way to inform the user that this callback has been removed and where to find it now. It's non-trivial enough for the user to figure out how to get this functionality back, I'd rather not remove it with no message. Would it be possible to make a `HaloCatalogCallback` class with an `__init__` function that accepts `*args, **kwargs` and a `__call__ ` method that just raises an exception with a message about this having moved to yt-astro?

neutrinoceros:
I experimented with this approach and in summary yes, it is possible.
I've tested 4 different ways to import yt + the astro extensions from the user side and checked that the Halo callback that persists in the registry is always from the extension when available (1.1+) and always from yt otherwise (duh).

So in fact we don't have to make the code here disappear, and we can *just* add a deprecation warning, which is much better. I'll update.

brittonsmith (approved):
Ok, that's great. Thanks!

PR #3595, opened 2021-10-15
CI: Only allow merges that are yes or no backport labeled

matthewturk:
This changes our "mergeability" configuration so that we need to label all pull requests with whether or not they should be backported.

### DIFF
--- a/.github/mergeable.yml
+++ b/.github/mergeable.yml
@@ -11,3 +11,7 @@ mergeable:
         must_include:
           regex: 'bug|enhancement|new feature|docs|infrastructure|dead code|refactor'
           message: "Please label this pull request with one of: bug, enhancement, new feature, docs or infrastructure."
+      - do: label
+        must_include:
+          regex: 'no-backport|backport-.*'
+          message: "Please label this pull request with either no-backport or a backport label for the appropriate branch."


munkm:
Does it matter when the backport label is applied? We could decide at release which bugfixes we'd like backported, rather than doing it at merge time.

neutrinoceros:
As far as I understand yes it matters. Labels only work if applied before merging. After the original PR was merged, we can decide to backport it with a comment.
Now I'm thinking, here are our concerns:

- because merging order may matter to avoid conflicts, it seems best to have a system that allows backports to happen as close as possible to the original merg. I assume this is Matt's motivation here. It also would reduce the risk of forgetting a bugfix that's too deep in the backlog.
- it's preferable if we don't *have* to use it, for the reasons Madicken and I gave.

So I'm thinking, why not use a milestone-based system for bug fixes that are issued at a time we know we want to do a bugfix release (such as now), and use comments for backports realize we need later ?

munkm:
Something that was mentioned in the workspace slack that I'm adding here for transparency is that it might be a waste of CI time to separately test each backport PR rather than doing them in one single PR. If we have a single PR for backports with *at least* a commit for each backport (depending on if we use our already existing backport script or not), the merging order won't matter. 

Generally to the conversation of this PR, I am generally not sold on requiring backport directions in every single PR that is submitted. Backport PRs that are user-facing should be a small fraction of the PRs submitted to the codebase and it shouldn't be necessary to label those for the sake of bugfix releases.

neutrinoceros:
note: please don't setup the milestone based workflow for now. I've been triaging backport PRs there to keep track of them, I realize now that if we were to tell Meeseeks Box to backport everything that gets merge into the milestone we would end up with back ports of back ports (possibly ?).

neutrinoceros:
Right. Should we close this then ?

matthewturk:
No, not yet.

matthewturk:
I think this has validity.  But, I'm going to close it as we need to get comfortable with the current processes.

PR #3600, opened 2021-10-17
ENH: only parse matplotlib's version once at runtime

neutrinoceros:
## PR Summary
A recent PR added a `MPL_VERSION` constant to the `visualization._commons` module.
This means we can now import from there instead of repeating the parsing recipe everywhere it's needed.


neutrinoceros:
Indeed. This is why I opened https://github.com/pypa/packaging/issues/465 upstream. Maybe one day we won't need to import `packaging.version.Version` more than once :)

neutrinoceros:
dropping this in favour of #3604

PR #3602, opened 2021-10-17
BUG: use tight layout by default when saving figures (PlotMPL) to avoid cropping labels out

### LINKED ISSUE #3601: BUG: saving images crops out colorbar labels
### Bug report

**Bug summary**
This is a bug that affects some of our image references, it happens when the dynamic range of data is small (less than an order of magnitude) and a log norm is used  for the colorbar (which happens by default).

**Code for reproduction**

```python
import yt

def background(fields, data):
    return data["gas", "density"] + data["gas", "density"].max()*10000

ds = yt.testing.fake_amr_ds([("gas", "density")], units=["g/cm**3"])
ds.add_field(("gas", "my_density"), background, sampling_type="cell", units="g/cm**3")
p = yt.SlicePlot(ds, "z", "my_density")
p.save(f"/tmp/box.png")
```

**Actual outcome**

box

**Expected outcome**
In fact the desirable output can be obtained as
```python
p.save(f"/tmp/box.png", mpl_kwargs={"bbox_inches": "tight"})
```
box

But I'd argue that this should be the default behaviour. This seems extremely easy to fix but I expect it will take some time to merge a patch here because changing the default behaviour may affect a lot of image references.

neutrinoceros:
fix #3601

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -137,6 +137,7 @@ class PlotMPL:
         ):
             mpl_kwargs["papertype"] = "auto"
 
+        mpl_kwargs.setdefault("bbox_inches", "tight")
         name = validate_image_name(name)
 
         try:


neutrinoceros:
So apparently this completely breaks answer testing as the produced images have different pixel sizes than the references. As a result, reports cannot even uploaded. This suddenly looks like one or two orders of magnitude more work than I anticipated, so I'd rather close it now.

PR #3615, opened 2021-10-20
ENH: Remove duplicate code in boxlib

cphyc:
Removes duplication in boxlib frontend.

This is one step aimed at minimizing code duplication.

The duplicates are spotted using
```bash
pylint --disable=all --enable=duplicate-code yt/
```
Note that I have **not** attempted to fix all duplicate here.

I am not fully satisfied with the way I have de-duplicated code with chombo/boxlib. Any input is welcome!
I'd also be happy to split the PR in multiple chunks.


matthewturk:
@zingale Can you peek at this real quick?  It looks OK to me.

zingale (approved):

zingale:
it looks good to me.  I haven't explicitly tested it (and don't really work with particles).

neutrinoceros:
Seems like we haven't been testing Orion2 data with sink files, and this patch broke that (#4078)
I get lost very quickly trying to isolate the part that needs fixing so I'm going to propose a simple revert.

PR #3618, opened 2021-10-21
BUG: fix a bug where 2D plots in spherical geometries would always use code-units in plot axes

### LINKED ISSUE #2913: Axes labels always in code for spherical datasets, independent of "unit_system"
### Bug report

**Bug summary**

@neutrinoceros discovered this when we went over #2728, what is happening is that the axes units are always, independent of `unit_system="cgs"`. It turns out that this is only the case for spherical datasets, and that it is not frontend-related. I managed to reproduce this issue with the [KeplerianDisk](https://girder.hub.yt/#item/5e17944f68085e00016c308b) dataset on the yt Hub as well, which is the athena_pp frontend I believe. 

**Code for reproduction**
```python
blast = "amrvac/bw_2d0000.dat"
disk = "KeplerianDisk/disk.out1.00000.athdf"
for file, name in zip((blast, disk), ("amrvac", "athenapp")):
    ds = yt.load(file, unit_system="cgs")
    p = yt.SlicePlot(ds, "phi", "density")
    p.save(f"slice_cgs_{name}.png")
```

**Actual outcome**
AMRVAC dataset:
slice_cgs_amrvac
AthenaPP dataset:
slice_cgs_athenapp

**Expected outcome**
R and z axis labels in `"cm"`.
For reference, this is a 3D Cartesian dataset (`"amrvac/bw_3d0000.dat"`)  sliced along `y`, same code as above:
slice_cgs_amrvac_cart
which behaves as it should.

neutrinoceros:
## PR Summary
fix #2913

```python
import yt

file = "KeplerianDisk/disk.out1.00000.athdf"
for us in ("cgs", "code"):
    ds = yt.load(file, unit_system=us)
    for normal in ("phi", "theta"):
        p = yt.SlicePlot(ds, normal, "density")
        p.plots["density"].figure.suptitle(f"unit system = {us}")
        p.save(f"/tmp/disk_slice_{us}.png")
```

gives
disK_slice_cgs
disK_slice_code

disk_slice_theta_cgs
disk_slice_theta_code

Marking this as draft because I ended up removing initialization code that was problematic in this specific case but I'd expect the tests to reveal why it was necessary.
In any cases, the code in question is duplicated in a lot of plot classes, so I'd need to apply consistent changes everywhere even if it turns out that it doesn't break anything.

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -170,7 +170,7 @@ answer_tests:
   local_cylindrical_background_010:  # PR 3520
     - yt/geometry/coordinates/tests/test_cylindrical_coordinates.py:test_noise_plots
 
-  local_spherical_background_002:  # PR 3533 + 3520
+  local_spherical_background_003:  # PR 3618
     - yt/geometry/coordinates/tests/test_spherical_coordinates.py:test_noise_plots
 
   #local_particle_trajectory_001:


neutrinoceros:

PR #3621, opened 2021-10-23
ENH: improve verbosity of yt.enable_parallelism in case of error

neutrinoceros:
## PR Summary
A simple improvement to yt.enable_parallelism:
- add type hints and document the return value
- add a log entry in case the function fails because of communicator size
  (previously an error was logged only if mpi4py wasn't installed)
- promote existing log message from `info` to `error` to give it better visibility

### DIFF
--- a/yt/utilities/parallel_tools/parallel_analysis_interface.py
+++ b/yt/utilities/parallel_tools/parallel_analysis_interface.py
@@ -5,6 +5,7 @@ import sys
 import traceback
 from functools import wraps
 from io import StringIO
+from typing import Optional
 
 import numpy as np
 from more_itertools import always_iterable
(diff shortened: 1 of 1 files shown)


yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -83,14 +86,18 @@ def enable_parallelism(suppress_logging=False, communicator=None):
...
+    -------
+    parallel_capable: bool
+        True if the call was successful. False otherwise.
     """
     global parallel_capable, MPI
     try:
         from mpi4py import MPI as _MPI
     except ImportError:
-        mylog.info("mpi4py was not found. Disabling parallel computation")
-        parallel_capable = False
```
matthewturk, line 92: does it default to `parallel_capable = False` in initialization?

matthewturk (approved):

PR #3624, opened 2021-10-28
ENH: Factorize code in halo containers

cphyc:
Removes duplication in halo containers.

This is one step aimed at minimizing code duplication.

The duplicates are spotted using
```bash
pylint --disable=all --enable=duplicate-code yt/
```
Note that I have **not** attempted to fix all duplicate here.

I am not fully satisfied with the way I have de-duplicated code with chombo/boxlib. Any input is welcome!
I'd also be happy to split the PR in multiple chunks.


matthewturk:
@brittonsmith do you have time to review this?

PR #3626, opened 2021-10-28
RFC: decouple yt.config from initialization

### LINKED ISSUE #3045: Unnecessary config file creation

### LINKED ISSUE #3625: RFC: yt.config module is too tightly coupled to yt.__init__
A bunch of code is run at import time that requires the global `yt.config.ytcfg` object to be set, forcing us to parse configuration mid initialisation.

This is an obstacle to #3045 and I figure that solving this alone is a manageable task so I'm opening a separate issue.

Essentially it should suffice to move every `from yt.config import ytcfg` statements into relevant functions (something one can see is done a lot in `yt/funcs.py` for instance) instead of having it at the top of many modules.

In practice there are a couple places where this isn't sufficient but it seems solvable.

_Originally posted by @neutrinoceros in https://github.com/yt-project/yt/issues/3045#issuecomment-954080091_

neutrinoceros:
## PR Summary

This patch postpones the "live" part of the configuration to the very end of `yt.__init__` and disentangle the configuration code a bit. This is achieved by
- never importing yt.config at the top of any module. Import statements are relocated to inside functions that need to access the global `yt.config.ytcfg` object. This is in line with the existing practice in the `yt.funcs` module
- making yt.config depend on yt.utilities.configure (instead of the other way around). This means a sizeable part of the diff here is simply due to relocating the `YTConfig` class
- adding a small callback system in `yt.utilities.configure`, allowing any module to perform configuration at runtime, but delaying it until `yt.config` is imported from for the first time (in `yt.__init__`)

The first commit fixes #3625 while the second fixes #3045

It also has obvious conflicts with #3106, but I think it'd be easier to merge the present PR first.

### DIFF
--- a/yt/__init__.py
+++ b/yt/__init__.py
@@ -100,6 +100,7 @@ def run_nose(*args, **kwargs):
     return run_nose(*args, **kwargs)
 
 
+from yt.config import _setup_postinit_configuration
 from yt.units.unit_systems import UnitSystem, unit_system_registry  # type: ignore
 
 # Import some helpful math utilities
@@ -153,3 +154,8 @@ from yt.visualization.volume_rendering.api import (
 
 #    TransferFunctionHelper, MultiVariateTransferFunction
 #    off_axis_projection
+
+
+# run configuration callbacks
+_setup_postinit_configuration()
+del _setup_postinit_configuration
--- a/yt/config.py
+++ b/yt/config.py
@@ -1,12 +1,6 @@
 import os
-import warnings
 
-# TODO: import tomllib from the standard library instead in Python >= 3.11
-import tomli as tomllib
-import tomli_w
-from more_itertools import always_iterable
-
-from yt.utilities.configuration_tree import ConfigLeaf, ConfigNode
+from yt.utilities.configure import YTConfig, config_dir, configuration_callbacks
 
 ytcfg_defaults = {}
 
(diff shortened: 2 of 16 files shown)


neutrinoceros:
pre-commit.ci autofix

matthewturk:
This seems like a good refactor; your third bullet point is intriguing to me though and I'd like to hear if you have anything in mind.  If you fix the conflicts I think it's good to go.

neutrinoceros:
>  your third bullet point is intriguing to me though and I'd like to hear if you have anything in mind.

I'm not promoting this callback system as a feature-enabling change, it's mostly there to allow fixing the issues I was looking into without breaking backwards compatibility.

>  If you fix the conflicts I think it's good to go.

solved !

neutrinoceros:
Forced pushed. Marking as draft until CI passes.

matthewturk:
@neutrinoceros Do you think it is worthwhile to explore a wrapper function (for our wrapper function!) that manages the imports appropriately?  Something like how we do on-demand imports, or something like that.  I am slightly anxious about the increased verbosity of adding `from yt.config import ytcfg` inside all the functions; it's not *bad* but I think it might discourage using the config system at all.

neutrinoceros:
I don't have a clear idea how to do that, and I'm convinced that I can come up with a different solution that will not add technical debt :/

neutrinoceros:
Actually here's an idea to avoid having to nest imports:
the callbacks may be run explicitly by calling a master function living inside `yt.config`, so it'd be easy to control when that function is called (at the end of `yt.__init__`). I've tested this locally and can easily update this PR. Would that work for you ?

matthewturk:
@neutrinoceros I think that would be great!

neutrinoceros:
Switching to draft to give priority to https://github.com/yt-project/yt/pull/3831, since these are conflicting but only #3831 needs backporting

neutrinoceros:
@matthewturk conflict solved

matthewturk (approved):

Issue #1796, opened 2018-05-23
Spherical coordinates viz


matthewturk:
After discussion with @ngoldbaum , we determined that the convention that is self-consistently applied throughout yt's coordinate handling is OK for spherical coordinates.  The variable that runs from 0..pi is called theta and the variable that runs from 0..2pi is called phi.  This is not what we expect in physics, but is coordinated with the mathematical convention.

Using thing as-is, volume is computed correctly, and the FLASH frontend is correctly set up.  `geom_transforms` in `yt/testing.py` is OK, too.  We can verify this by computing `.r[:].sum("cell_volume")` on a dataset.

With #1791 , we correctly do things in geographic coordinate handler, but this exposed that the spherical coordinate handler was (I believe) backwards *in the visualization*, but not the rest.

Unless we take on a process of changing all variable names from theta to phi, what needs to happen next:

- [x] Comments throughout the code and docs explaining the pi and 2pi variables
- [ ] Possibly a second class of Spherical handler that switches these, or maybe changes them to "azimuthal" and "axial" or something.
- [x] Carefully adjust the `sanitize_center` and `sanitize_width` functions in the spherical coordinate handler to ensure they are correctly set for the appropriate axes.
- [ ] Add a section to the documentation about "coordinate system reference" -- a table that describes each of the extant coordinate systems/geometries, their expected bounds, and their visualization methods.

Initial inspection suggests that the third item is complete, but I will be constructing a notebook that demonstrates this.

ngoldbaum:
This should also fix the errors like this that recently cropped up in the unit tests:

```
Exception ZeroDivisionError: 'float division' in 'yt.utilities.lib.pixelization_routines.aitoff_thetaphi_to_xy' ignored
ZeroDivisionError: float division
Exception ZeroDivisionError: 'float division' in 'yt.utilities.lib.pixelization_routines.aitoff_thetaphi_to_xy' ignored
```

These errors were introduced by merging #1791. If this take a lot of time to fix, a quick fix specifically for the error messages generated by the unit tests would be to revert the changes to the spherical coordinate handler in that PR.

neutrinoceros:
I think we can check the third box with #3533

neutrinoceros:
@matthewturk so I think items 1 and 3 are done, AFAICT, item 2 isn't necessary, and item 4 is covered with https://yt-project.org/doc/visualizing/plots.html#slice-plots-and-projection-plots-in-spherical-geometry
Can we close this ?

[...]

Issue #3610, opened 2021-10-19
BUG: wrong pixelization for "r-normal" projections in spherical coordinates ?


neutrinoceros:
### Bug report

**Bug summary**
originally reported as #3529

**Code for reproduction**

```python
import yt

ds = yt.load_sample("bw_spherical_2d", unit_system="code")
p = yt.SlicePlot(ds, "r", "density")
p.save(f"/tmp/test.png")
```

**Actual outcome**
![test](https://user-images.githubusercontent.com/14075922/137946301-39d9de61-03f4-4446-b101-cd5511fb211c.png)

**Expected outcome**
Seemingly, working examples of this usage were documented in https://github.com/yt-project/yt/pull/1791
but it also looks like the important code here has really not changed much since then (except for formatting), and the PR included tests, so I'm surprised to see it failing now. I'll try to inspect this deeper later.

neutrinoceros:
(I might be _unable_ to actually, cartopy is notoriously hard to install outside of conda land)

PR #3628, opened 2021-10-29
BUG: fixes for Hammer-Aitoff projection (r-normal) 2D viz in spherical geometries

neutrinoceros:
## PR Summary
This is a jab at the first item in #1796
Spherical geometries need more work but this is a first step to improve maintainability.
Right now, the defintions of "theta" and "phi" are misaligned between the aitoff pixelizer and the rest of the code base.
Fortunately, the latter is correct (aligned with most common notations AFAIC), while the former isn't user facing so it's an easy fix.

edit: this now completes the 3rd item of #1796

### DIFF
--- a/doc/source/visualizing/plots.rst
+++ b/doc/source/visualizing/plots.rst
@@ -444,7 +444,7 @@ keyword arguments, as described in
 Slice Plots and Projection Plots in Spherical Geometry
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-What to expect when plotting data in spherical geometry ? Here we explain
+What to expect when plotting data in spherical geometry? Here we explain
 the notations and projections system yt uses for to render 2D images of
 spherical data.
 
--- a/doc/source/visualizing/plots.rst
+++ b/doc/source/visualizing/plots.rst
@@ -438,6 +438,65 @@ to project along, and a field to project.  For example:
 keyword arguments, as described in
 :class:`~yt.visualization.plot_window.OffAxisProjectionPlot`
 
+
+.. _slices-and-projections-in-spherical-geometry:
+
+Slice Plots and Projection Plots in Spherical Geometry
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+What to expect when plotting data in spherical geometry ? Here we explain
+the notations and projections system yt uses for to render 2D images of
+spherical data.
+
+The native spherical coordinates are
+- the spherical radius :math:`r`
+- the colatitude :math:`\theta`, defined between :math:`0` and :math:`\pi`
+- the azimuth :math:`\varphi`, defined between :math:`0` and :math:`2\pi`
+
+:math:`\varphi`-normal slices are represented in the poloidal plane, with axes :math:`R, z`, where
+- :math:`R = r \sin \theta` is the cylindrical radius
+- :math:`z = r \cos \theta` is the elevation
+
+.. python-script::
+
+   import yt
+
+   ds = yt.load_sample("KeplerianDisk", unit_system="cgs")
+   slc = yt.SlicePlot(ds, "phi", ("gas", "density"))
+   slc.save()
+
+:math:`\theta`-normal slices are represented in a
+:math:`x/\sin(\theta)` VS :math:` y/(\theta)` plane, where
+- :math:`x = R \cos \varphi`
+- :math:`y = R \sin \varphi`
+
+are the cartesian plane coordinates
+
+.. python-script::
+
+   import yt
+
+   ds = yt.load_sample("KeplerianDisk", unit_system="cgs")
+   slc = yt.SlicePlot(ds, "theta", ("gas", "density"))
+   slc.save()
+
+
+Finally, :math:`r`-normal slices are represented following a
+`Aitoff-Hammer projection <http://paulbourke.net/geometry/transformationprojection/>`_
+
+We denote
+- the latitude :math:`\bar\theta = \frac{\pi}{2} - \theta`
+- the longitude :math:`\lambda = \varphi - \pi`
+
+.. python-script::
+
+   import yt
+
+   ds = yt.load_sample("KeplerianDisk", unit_system="cgs")
+   slc = yt.SlicePlot(ds, "r", ("gas", "density"))
+   slc.save()
+
+
 .. _unstructured-mesh-slices:
 
 Unstructured Mesh Slices
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -589,36 +589,38 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
                 r_i += 0.5*dx
             theta_i += dthetamin
 
-cdef int aitoff_thetaphi_to_xy(np.float64_t theta, np.float64_t phi,
+cdef int aitoff_Lambda_btheta_to_xy(np.float64_t Lambda, np.float64_t btheta,
                                np.float64_t *x, np.float64_t *y) except -1:
-    cdef np.float64_t z = math.sqrt(1 + math.cos(phi) * math.cos(theta / 2.0))
-    x[0] = math.cos(phi) * math.sin(theta / 2.0) / z
-    y[0] = math.sin(phi) / z
+    cdef np.float64_t z = math.sqrt(1 + math.cos(btheta) * math.cos(Lambda / 2.0))
+    x[0] = 2.0 * math.cos(btheta) * math.sin(Lambda / 2.0) / z
+    y[0] = math.sin(btheta) / z
     return 0
 
 @cython.cdivision(True)
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def pixelize_aitoff(np.float64_t[:] theta,
-                    np.float64_t[:] dtheta,
-                    np.float64_t[:] phi,
-                    np.float64_t[:] dphi,
+def pixelize_aitoff(np.float64_t[:] azimuth,
+                    np.float64_t[:] dazimuth,
+                    np.float64_t[:] colatitude,
+                    np.float64_t[:] dcolatitude,
                     buff_size,
                     np.float64_t[:] field,
-                    extents, input_img = None,
-                    np.float64_t theta_offset = 0.0,
-                    np.float64_t phi_offset = 0.0):
+                    bounds, # this is a 4-tuple
+                    input_img = None,
+                    np.float64_t azimuth_offset = 0.0,
+                    np.float64_t colatitude_offset = 0.0):
     # http://paulbourke.net/geometry/transformationprojection/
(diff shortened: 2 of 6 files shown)


neutrinoceros:
So while I'm at it I figured I might as well give a shot at fixing #3610 and made some progress there
rerunning the demo script I wrote in a comment (https://github.com/yt-project/yt/issues/3610#issuecomment-948758088)
```python
import yt

ds = yt.load_sample("KeplerianDisk")
p = yt.SlicePlot(ds, "r", "density")
p.save(f"/tmp/disk_slice_r.png")
```
I'm now getting this:
disk_slice_r

There are still a couple issues, but I was able to
- fix axes labels
- ... including units (none ! that's a new special case)
- compute the edges and center of the image correctly

remaining issues:
1) whitespace on the top and bottom of the image should be much thiner. This is a bug in the pixelizer itself that I think was coded with only full spherical domains in mind.
2) the colormap label incorrectly uses code-units (sigh... this bug is never going away, is it ?)

So I still want to fix 1) here, but I may reserve 2) for a different PR

neutrinoceros:
Update:
disk_slice_r

- fixed labels again (introduce a different notation to distinguish latitude (theta') from colatitude (theta)), same with longitude and azimuth. This will need to be better documented (for now I'm commenting the code)
- fixed the figure space's occupancy, though obviously the pixelizer needs more work still

I think I'm getting there, but I need to carefully read the pixelizer code to finish my generalisation

matthewturk:
@neutrinoceros could you expand on the "obviously"?  I'm not sure I understand specifically what you mean.

neutrinoceros:
Let me rephrase this more carefully: 
I *think* I understand everything that happens outside the pixelizer routine and that I'm *currently assuming* that it's the only place left where a bug could lurk.

I've been simplifying the math locally (not pushed yet) but so far I'm still stuck with the visual offset

neutrinoceros:
Currently the routine is hitting the right *amount* of pixels and the end product it very close to what I expect, but something about coordinate transformations seems to be off. Any help / insight would be greatly appreciated, I haven't been able to make any significant progress for a couple hours now.

neutrinoceros:
Now we're getting somewhere
disk_slice_r

I think I'm done with the bugs I wanted to deal with for now. I'll come back to see if I need to fix anything after CI completes. Otherwise I'll just clean up the branch and rebase before I open this to review.

neutrinoceros:
So while this fixes the viz for the Athena++ `"KeplerianDisk"` dataset, it also ruins it in the case of the AMRVAC `"bw_spherical_2d"` dataset (gives out a blank image). As noted in #3610, a potentially key difference between the two formats is in the axis order, see https://github.com/yt-project/yt/issues/3610#issuecomment-948758088

matthewturk:
I think this is definitely the more correct approach, but I am vaguely left
wishing we had a better way to display so we could have something more
terse.

neutrinoceros:
Me too, but there's too much going on, so I'm focusing on making it work for now.
The bounds are not computed correctly for the AMRVAC dataset so I'll try to fix this later today.

neutrinoceros:
There was a confusion in my definition of the latitude I wrote that
latitude = colatitude - PI/2

but it's really
latitude = PI/2 - colatitude

As a result the pixelizer is likely inverting North and South. In fact I know it does: here's a dataset where velocity_theta is positive in the North half and negative in the South half:
data 0009 vtk_Slice_r_VX2
(this is an Idefix dataset, I can't easily share the reproducer for now)

neutrinoceros:
fixed
data 0009 vtk_Slice_r_VX2

neutrinoceros:
Update:
- I changed the notation in the labels again (I really need to document them but I think the ones I got now are the best choice I can find)
- the reason why this is still a draft is because I know a case that completely breaks with this patch/

```python
import yt

ds = yt.load_sample("bw_spherical_2d", unit_system="code")
# using the aspect argument only to limit the height of the second image
# improvement is underway in https://github.com/yt-project/yt/pull/3633
p = yt.SlicePlot(ds, "r", "density", aspect=0.5)
p.save(f"/tmp/test.png")
```

main branch:

test4


this work (current):
test4


so I think what's happening is an unfortunate side effect of the default "origin" argument in SlicePlot which doesn't plays well here because we _can't_ simply translate the frame as we do in cartesian. I haven't looked into how to test hunch this properly, but I know that the bounds are correctly computed in `SphericalCoordinateHandler`.
@matthewturk , any inputs here ?

neutrinoceros:
Fixed it !
test4
Now I'll wait one more round of CI to check I have not broken anything and then I'll rebase + open for review

neutrinoceros:
The one failing test is checking reproducibility of the FRB for basic pixelization calls. It can be visually reproduced with
```python
import yt

p = yt.SlicePlot(ds, "r", "Density")
p.save("/tmp/")
```
### main branch
main_Slice_r_Density
### this branch
branch_Slice_r_Density

So the result is indeed different, but I believe that it's more correct with this branch: Aitoff coordinates range from -1 to +1 in both directions, so with the default aspect ratio (1), projecting a complete sphere should produce a disk, not an ellipse.
I'll request authorisation to bump this answer test.

neutrinoceros:
I found some more inconsistencies, this is getting back to the shop.

neutrinoceros:
I guess that the expected result should be an ellipse after all... I thought that this was a difference between Aitoff and Aitoff-Hammer projections but I can't find a source for that and from what I can find now it looks like they're indeed the same projection. So it's pretty easy actually, I just need to rework this so that the coordinate along the horizontal axis is doubled. Back to the shop again.

https://en.wikipedia.org/wiki/Aitoff_projection

neutrinoceros:
Easy enough
test5_r

neutrinoceros:
@matthewturk, I'll let you be the judge. If you think this is enough to close #1796, you can link the issue. Otherwise, please tell me what more you'd expect/need to see done to fix the whole issue.

neutrinoceros:
@matthewturk can I bump answers here ? it's fine if you want to have a deeper look at the code first, just please let me know

matthewturk:
do it!

neutrinoceros:
Ah, the report is gone, so I'll rebase first and bump tomorrow. Switching to draft again.

doc/source/visualizing/plots.rst
```diff
@@ -438,6 +438,65 @@ to project along, and a field to project.  For example:
 keyword arguments, as described in
 :class:`~yt.visualization.plot_window.OffAxisProjectionPlot`
 
+
+.. _slices-and-projections-in-spherical-geometry:
+
+Slice Plots and Projection Plots in Spherical Geometry
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+What to expect when plotting data in spherical geometry ? Here we explain
```
matthewturk, line 447:
```suggestion
What to expect when plotting data in spherical geometry? Here we explain
```

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -589,36 +589,38 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
...
     return 0
 
 @cython.cdivision(True)
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def pixelize_aitoff(np.float64_t[:] theta,
-                    np.float64_t[:] dtheta,
-                    np.float64_t[:] phi,
-                    np.float64_t[:] dphi,
+def pixelize_aitoff(np.float64_t[:] azimuth,
```
matthewturk, line 602: I'm pretty ambivalent about the idea of changing all the notation in here, but I suppose it *does* make things clearer.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -631,65 +633,74 @@ def pixelize_aitoff(np.float64_t[:] theta,
...
+    dy = yw / (img.shape[1] - 1)
     x = y = 0
     for fi in range(nf):
-        theta_p = (theta[fi] + theta_offset) - PI
-        dtheta_p = dtheta[fi]
-        phi_p = (phi[fi] + phi_offset) - PI/2.0
-        dphi_p = dphi[fi]
+        Lambda_p = (azimuth[fi] + azimuth_offset) - PI
+        dLambda_p = dazimuth[fi]
+        btheta_p = PI/2.0 - (colatitude[fi] + colatitude_offset)
```
matthewturk, line 647: Can you spare a moment to explain why this is different now?  It's the reverse of before, yes?

matthewturk (dismissed):
The changes inside the pixelizer (specifically reversing the order, etc) are a bit confusing to me, but I see that the resultant images are OK, so I don't object.

yt/utilities/lib/pixelization_routines.pyx
neutrinoceros: yes. AFAICT, everywhere in the code base, we denote the _colatitude_ "theta". As the name suggests, _colatitude_ is the complementary angle to _latitude_ (so lat + coat = PI/2). On the main branch, the angle that's noted `phi` here effectively is the opposite of latitude. So I'm changing the name to `btheta` to reflect that it's neither the azimuth (phi) nor the colatitude (theta), and I'm changing its sign because... that's just a bug (I discovered it in testing when I saw a sphere being displayed upside down, with the north emisphere on the bottom half).

neutrinoceros:
Something went terribly wrong in the docs build but I don't believe it's reproducible
@yt-fido test this please

neutrinoceros:
Read too fast, this is perfectly reproducible indeed, but I'm still surprised that we're only seeing this now
```
AttributeError: 'TitleCallback' object has no attribute '_supported_geometries'
```

neutrinoceros:
This error is absolutely mysterious to me. `TittleCallback` instances should always inherit a `_supported_geometries` from the base `PlotCallback` class. Also worth mentioning CI was green on c7304b4 but went red with 110185c with a clearly unrelated error.
I'm lost. @Xarthisius, any idea what might be going on ?

Xarthisius:
> I'm lost. @Xarthisius, any idea what might be going on ?

Have you tried to reproduce it locally with current `main`? Usually that happens when a PR that didn't trigger docs build gets merged and breakage goes unnoticed.

neutrinoceros:
I did try
```
cd doc
python helper_scripts/run_recipes.py
```
but this seems even more broken on my machine (I get stuck in a loop of error messages where processes fail to spawn because of some race condition ? and it didn't get better after I required a single process instead of 6 or replaced `Pool.async_map` with `Pool.map`)

Xarthisius:
> I did try
> 
> ```
> cd doc
> python helper_scripts/run_recipes.py
> ```

Next time, don't try to run everything. Run the thing that fails. E.g. in the traceback I posted (https://github.com/yt-project/yt/issues/3690#issuecomment-975775259) it sufficient to do:

```
python doc/source/visualizing/colormaps/cmap_images.py
```
to easily hit that error.

neutrinoceros:
Thank you, dully noted.
In the mean time I was able to find the bug (which is indeed on the main branch). Posting a PR in a couple minutes

matthewturk (approved):

neutrinoceros:
@matthewturk should this go in ?

Issue #3629, opened 2021-10-30
[GDF] Grids cell width is wrongly set to 1.0 for inappropriate dimension for 2D datasets.


Xarthisius:
### Bug report

**Bug summary**

I'm trying to create a simple plot for a dataset with a specific orientation: 2d, where x-axis is reduced, y-axis is horizontal and z-axis is vertical. It turns out to be pretty impossible (unless I'm missing something obvious).

Data: http://use.yt/upload/d1b9ac33

Note, you can use different data. 3D even. Just try non-standard axis orientation.

**Code for reproduction**

What I want to achieve (roughly):

```
import matplotlib.pyplot as plt
import h5py as h5
import numpy as np

with h5.File("crwind_tst_0001.h5", "r") as h5f:
    data = h5f["/data/grid_0000000000/density"][:].swapaxes(0, 2)
    data = np.squeeze(data)

fig = plt.figure(num=1, figsize=(4, 6.5))
ax = fig.add_subplot(111)
ax.set_ylim(-1000, 1000)
ax.set_xlim(-500, 500)
c = ax.imshow(
    np.log10(data.T),
    aspect="equal",
    extent=[-500, 500, -1000, 1000],
    origin="lower",
)
plt.colorbar(c)
plt.tight_layout()
fig.savefig("manual.png")
```
which yields:

![manual](https://user-images.githubusercontent.com/352673/139555101-601e83a0-f7e7-446c-ba9f-5c293f904319.png)

**Actual outcome aka Code I've tried**

1. Default SlicePlot
```
import yt
from yt.visualization.image_writer import write_image

ds = yt.load("crwind_tst_0001.h5")
slc = yt.SlicePlot(ds, "x", "density")
slc.save()  ### so wrong
```
![crwind_tst_0001_Slice_x_density](https://user-images.githubusercontent.com/352673/139555136-6191deb1-684a-4a1c-86ab-6692366d5020.png)

1. Manual slicing:

```
import numpy as np
import yt
from yt.visualization.image_writer import write_image

ds = yt.load("crwind_tst_0001.h5")
slc = ds.slice(0, 0.0)
ds.coordinates.x_axis["x"] = 2
ds.coordinates.x_axis[0] = 2
ds.coordinates.y_axis["x"] = 1
ds.coordinates.y_axis[0] = 1
frb = slc.to_frb(ds.domain_width[1], (480, 960), center=ds.domain_center, height=ds.domain_width[2])
[...]

PR #3630, opened 2021-10-31
[gdf] dimensionality doesn't imply which dimension is reduced

Xarthisius:
Root cause that I was hitting this is that `b"piernik" != "piernik"`... However, I came to a conclusion that this clause is wrong for any GDF output, not just that one particular flavor. GDF already implies that RE/LE is a 3-element array, so there's no point in making assumptions about 1) which dimension is reduced or 2) what the value of dds in that dimension should be.

Using script from #3629:

```python
import yt

ds = yt.load("crwind_tst_0001.h5")
slc = yt.SlicePlot(ds, "x", "density")
slc.save("zy.png")

ds.coordinates.x_axis["x"] = 2
ds.coordinates.x_axis[0] = 2
ds.coordinates.y_axis["x"] = 1
ds.coordinates.y_axis[0] = 1
slc = yt.SlicePlot(ds, "x", "density")
slc.save("yz.png")
```

now I get:

yz
zy

### DIFF
--- a/yt/frontends/gdf/data_structures.py
+++ b/yt/frontends/gdf/data_structures.py
@@ -46,11 +46,6 @@ class GDFGrid(AMRGridPatch):
         else:
             LE, RE = self.index.grid_left_edge[id, :], self.index.grid_right_edge[id, :]
             self.dds = np.array((RE - LE) / self.ActiveDimensions)
-        if self.ds.data_software != "piernik":
-            if self.ds.dimensionality < 2:
-                self.dds[1] = 1.0
-            if self.ds.dimensionality < 3:
-                self.dds[2] = 1.0
         self.field_data["dx"], self.field_data["dy"], self.field_data["dz"] = self.dds
         self.dds = self.ds.arr(self.dds, "code_length")
 


neutrinoceros:
Well played ! Be

Notes on the output graph:
- the axis labels are not changed
- the aspect of the figure varies widely, I suspect this is because the figure width is hardcoded (and the aspect ratio is always 1 by default)

These are points I'd like addressed along with the original issue you were pointing out: there's a demand for supporting this operation but the current solution is a little hacky as well as undocumented.

I'm writing this note on my phone mostly so I can open an issue from it later. I'm not asking for changes to the present PR.
I'll make an actual review in a couple hours.

neutrinoceros (approved):
So I did a little research on the data format and all I could find so far was a repo of yours
https://github.com/Xarthisius/fgdfio
which gives out a link to the format spec, which is unfortunately broken
https://bitbucket.org/yt_analysis/grid_data_format/

I confirm I can reproduce the error and that the patch fixes it, so I could just merge this, but out of curiosity, do you have any resource on this data format that I could refer to ?

Xarthisius:
> I confirm I can reproduce the error and that the patch fixes it, so I could just merge this, but out of curiosity, do you have any resource on this data format that I could refer to ?

https://github.com/yt-project/grid_data_format

neutrinoceros:

PR #3633, opened 2021-10-31
BUG: fix figure layout in case the long axis is vertical

### LINKED ISSUE #3632: BUG: figures with height > width are too tall
### Bug report

**Bug summary**

related to #3631
Currently when plotting a region that happens to be broader in "height" (vertical axis) than in "width" (horizontal axis), the output figure may be very tall and not as readable as the standard square figure produced by yt.

**Code for reproduction**

```python
ds = yt.load_sample("KelvinHelmholtz")

slc = yt.SlicePlot(ds, "z", "density")
slc.save("/tmp/rect_1.png")
```

**Actual outcome**
zy

**Expected outcome**

I'd except a smaller figure, that would fit my screen while staying readable, on par with what I obtain after flipping axes:
```python
vals = ds.coordinates.y_axis["z"], ds.coordinates.x_axis["z"]
ds.coordinates.x_axis["z"], ds.coordinates.y_axis["z"] = vals
ds.coordinates.x_axis[2], ds.coordinates.y_axis[2] = vals
slc = yt.SlicePlot(ds, "z", "density")
slc.save("/tmp/rect_2.png")
```
yz

I think the root cause is trivially that the figure width is hardcoded while the aspect ratio is constant.
I would make more sense to hardcode the width of the largest dimension, wether it happens to be the width or the height of the figure, and it would be backwards compatible in cases where it doesn't matter.

neutrinoceros:
## PR Summary

fix #3632
Running the script therein I get:
zy

I expect this to fail CI because we surely have a couple images whose size is going to change, and unfortunately I don't think our testing infra is currently able to manage this properly (see #3602)

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -170,7 +170,7 @@ answer_tests:
   local_cylindrical_background_011:  # PR 3670
     - yt/geometry/coordinates/tests/test_cylindrical_coordinates.py:test_noise_plots
 
-  local_spherical_background_004:  # PR 3670
+  local_spherical_background_005:  # PR 3633
     - yt/geometry/coordinates/tests/test_spherical_coordinates.py:test_noise_plots
 
   #local_particle_trajectory_001:
--- a/yt/utilities/answer_testing/framework.py
+++ b/yt/utilities/answer_testing/framework.py
@@ -13,6 +13,7 @@ import sys
 import tempfile
 import time
 import urllib
+import warnings
 import zlib
 from collections import defaultdict
 
@@ -777,6 +778,32 @@ def dump_images(new_result, old_result, decimals=10):
         sys.stderr.write("\n")
 
 
+def ensure_image_comparability(a, b):
+    # pad nans to the right and the bottom of two images to make them comparable
+    # via matplotlib if they do not have the same shape
+    if a.shape == b.shape:
+        return a, b
+
+    assert a.shape[2:] == b.shape[2:]
+
+    warnings.warn(
+        f"Images have different shapes {a.shape} and {b.shape}. "
+        "Padding nans to make them comparable."
+    )
+    smallest_containing_shape = (
+        max(a.shape[0], b.shape[0]),
+        max(a.shape[1], b.shape[1]),
+        *a.shape[2:],
+    )
+    pa = np.full(smallest_containing_shape, np.nan)
+    pa[: a.shape[0], : a.shape[1], ...] = a
+
+    pb = np.full(smallest_containing_shape, np.nan)
+    pb[: b.shape[0], : b.shape[1], ...] = b
+
+    return pa, pb
+
+
 def compare_image_lists(new_result, old_result, decimals):
     fns = []
     for _ in range(2):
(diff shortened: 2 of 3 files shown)


neutrinoceros:
I modified the testing framework to make image comparison possible even when shapes diverge

neutrinoceros:
Adding this to the 4.0.2 meta issue as a candidate for the bug fix release, but it's not 100% clear that it should be considered a bug fix so I'd like this to be discussed in the review.

neutrinoceros:
Got a handful of expected failures, exclusively for images that are targeted by the patch, so I'm opening for review.

neutrinoceros:
I'm not convinced this plays correctly with user-specified aspect. Switching to draft for now

neutrinoceros:
Fixed the issue with user-defined aspect, I believe the result is now what users would expect: the size of the long axis stays consistent
```python
import yt

ds = yt.load_sample("KelvinHelmholtz")
slc = yt.SlicePlot(ds, "z", "density")
slc.save("/tmp/rect_1.png")

slc = yt.SlicePlot(ds, "z", "density", aspect=0.5)
slc.save("/tmp/rect_2.png")

slc = yt.SlicePlot(ds, "z", "density", aspect=0.2)
slc.save("/tmp/rect_3.png")

slc = yt.SlicePlot(ds, "z", "density", aspect=1.5)
slc.save("/tmp/rect_4.png")
```
rect_1
rect_2
rect_3
rect_4

neutrinoceros:
I think I got this to a stable state. Now it seems rather obvious that the code could be simplified/clarified, so I'm keeping the draft status, and I'll make a final push on this when I have time to focus on it again.

neutrinoceros:
I realized that my modified testing framework produces nice reports but will upload polluted artifacts that are not suited to update the answer store.
It needs more work on this point too

neutrinoceros:
I'm going to try to validate this by updating the answer store early. I'll do this in the least disruptive way possible by pushing a branch directly to the mother repo and check it out here so I don't need to iterate with reviewers to see wether my modified framework works as intended.

neutrinoceros:
It's alive ! now let's try to simplify the logic

neutrinoceros:
I think more and more that this should be considered a bug fix since I'm fulfilling the original intent of this code.

neutrinoceros:
@Xarthisius, if you get a chance, could you review the changes I did to the testing framework here please ?

chrishavlin (commented):
So after a lot of thought, I don't think this is quite the right approach. Consider the final two images you plotted:

```
slc = yt.SlicePlot(ds, "z", "density", aspect=0.2)
slc.save("/tmp/rect_3.png")

slc = yt.SlicePlot(ds, "z", "density", aspect=1.5)
slc.save("/tmp/rect_4.png")
```

The standard definition of an aspect ratio is width / height. But those final two images break that definition: in the first, you're setting an aspect ratio < 1 and getting an image where height > width and in the second the opposite. I think we should keep the standard definition.

The base problem you're trying to address is that small aspect ratio plots end up too tall, right? So why not adjust the default figure width to a smaller value if aspect ratio < 1 so that the height ends up more reasonable?

neutrinoceros:
> The standard definition of an aspect ratio is width / height. But those final two images break that definition: in the first, you're setting an aspect ratio < 1 and getting an image where width > height and in the second the opposite. I think we should keep the standard definition.

So I hear what you're saying, unfortunately the `aspect` argument, as it's actually used internally, is rather the ratio of `L0 / vertical space` over `L0 / horizontal space` where `L0` is any reference length in physical units, and vertical/horizontal space are measured in pixels (or any screen/paper relevant unit). I believe my current implementation is faithful to this pre-existing design.

Also note that the baseline example (not using the `aspect` argument actually has an `aspect` of `1`, but it's not a square, because the simulation domain isn't a square either. This is a bit tricky and perhaps exposing the `_unit_aspect` attribute as simply `aspect` was a mistake, but that shipped has sailed and I'm trying to work with the existing API.

chrishavlin:
Ya, I've looked at this a few times the past couple days and I don't feel great about it, but I guess this PR isn't really the place to address the confusing way we handle aspect ratios. I think my main issue is that `aspect` ends up being passed to `imshow` so we should be conforming to the standard definition of the aspect ratio expected by `imshow` regardless of how it's used internally. But ya, I suppose this PR isn't the spot to fix that (if it is even something that should be addressed).

chrishavlin (approved):

neutrinoceros:
ok then, thanks for approving. I'll update this with an actual companion PR to the answer store

neutrinoceros:
Here's the companion PR https://github.com/yt-project/answer-store/pull/29

neutrinoceros:
Update: I've validated the companion PR to the answer store by checking it in on this side, so I believe it's good to go now.
Then I'll update the present PR again to point to the perennial branch of the answer store, and bump into answers for Jenkins as well so we should get CI back to green.

neutrinoceros:
thanks @cphyc for merging the companion PR. I bumped everywhere needed and I now expect CI to turn green

chrishavlin (approved):

PR #3635, opened 2021-10-31
BUG: use tight layout by default when saving matplotlib figures

### LINKED ISSUE #3601: BUG: saving images crops out colorbar labels

neutrinoceros:
## PR Summary

This is a second a second attempt to fix #3601, reusing the patch from #3602
combined with the test framework modifications from #3633 to enable comparisons
for images of different sizes.
Opening as a draft to indicate #3633 takes priority over this.

### DIFF
--- a/yt/utilities/answer_testing/framework.py
+++ b/yt/utilities/answer_testing/framework.py
@@ -13,6 +13,7 @@ import sys
 import tempfile
 import time
 import urllib
+import warnings
 import zlib
 from collections import defaultdict
 
@@ -806,6 +807,32 @@ def dump_images(new_result, old_result, decimals=10):
         sys.stderr.write("\n")
 
 
+def ensure_image_comparability(a, b):
+    # pad nans to the right and the bottom of two images to make them comparable
+    # via matplotlib if they do not have the same shape
+    if a.shape == b.shape:
+        return a, b
+
+    assert a.shape[2:] == b.shape[2:]
+
+    warnings.warn(
+        f"Images have different shapes {a.shape} and {b.shape}. "
+        "Padding nans to make them comparable."
+    )
+    smallest_containing_shape = (
+        max(a.shape[0], b.shape[0]),
+        max(a.shape[1], b.shape[1]),
+        *a.shape[2:],
+    )
+    pa = np.full(smallest_containing_shape, np.nan)
+    pa[: a.shape[0], : a.shape[1], ...] = a
+
+    pb = np.full(smallest_containing_shape, np.nan)
+    pb[: b.shape[0], : b.shape[1], ...] = b
+
+    return pa, pb
+
+
 def compare_image_lists(new_result, old_result, decimals):
     fns = []
     for _ in range(2):
@@ -815,8 +842,12 @@ def compare_image_lists(new_result, old_result, decimals):
     num_images = len(old_result)
     assert num_images > 0
     for i in range(num_images):
-        mpimg.imsave(fns[0], np.loads(zlib.decompress(old_result[i])))
-        mpimg.imsave(fns[1], np.loads(zlib.decompress(new_result[i])))
+        expected = np.loads(zlib.decompress(old_result[i]))
+        actual = np.loads(zlib.decompress(new_result[i]))
+        expected, actual = ensure_image_comparability(expected, actual)
+
+        mpimg.imsave(fns[0], expected)
+        mpimg.imsave(fns[1], actual)
         results = compare_images(fns[0], fns[1], 10 ** (-decimals))
         if results is not None:
             if os.environ.get("JENKINS_HOME") is not None:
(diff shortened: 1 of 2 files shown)


cphyc:
Thinking on top of my head on my phone, but have you tried using `constrained_layout=True` when creating the matplotlib figure?

31 Oct 2021 17:58:52 Clément Robert ***@***.***>:

> 
> *PR Summary*
> 
> This is a second a second attempt to fix #3601[https://github.com/yt-project/yt/issues/3601], reusing the patch from #3602[https://github.com/yt-project/yt/pull/3602]
> combined with the test framework modifications from #3633[https://github.com/yt-project/yt/pull/3633] to enable comparisons
> for images of different sizes.
> Opening as a draft to indicate #3633[https://github.com/yt-project/yt/pull/3633] takes priority over this.
> 
> ----------------------------------------
> 
> *You can view, comment on, or merge this pull request online at:*
> 
>   https://github.com/yt-project/yt/pull/3635
> 
> 
> *Commit Summary*
> 
* > ENH: Improve 2D figure layout in case the long axis is vertical[https://github.com/yt-project/yt/pull/3635/commits/4ffc8896354d1756fe250c6ea6626633859cf8b1]
* > TST: modify image comparison method to enable reports when actual and expected don't have the same shape[https://github.com/yt-project/yt/pull/3635/commits/af7093c1799cdf59fbc93808ffcba73becb72a94]
* > BUG: use tight layout by default when saving figures (PlotMPL) to avoid cropping labels out[https://github.com/yt-project/yt/pull/3635/commits/af2bf3e9efb883a36947804b3b790f5e2a4941d3]
> 
> *File Changes*
> 
> (2 files[https://github.com/yt-project/yt/pull/3635/files])
> 
* > *M* yt/utilities/answer_testing/framework.py[https://github.com/yt-project/yt/pull/3635/files#diff-0588597c99a8f042699cce2585f6e885e3dcbc0e2cc13c4618ca58eadf875a6f] (35)
* > *M* yt/visualization/base_plot_types.py[https://github.com/yt-project/yt/pull/3635/files#diff-813fdd3d212be372fea21282904849bdc1a899e20e90523a82a271a2d2074963] (13)
> 
> *Patch Links:*
> 
* > https://github.com/yt-project/yt/pull/3635.patch
* > https://github.com/yt-project/yt/pull/3635.diff

neutrinoceros:
I have no experience with this paradigm, so no. What benefits do you expect from this approach ?

cphyc:
> I have no experience with this paradigm, so no. What benefits do you expect from this approach ?

According to few sources on matplotlib's issue tracker (see e.g. https://github.com/matplotlib/cheatsheets/issues/30#issuecomment-655594205), `constrained_layout` is supposed to be a better version of `tight_layout` that doesn't screw your layout. To use it, create the figure using
```python
fig = plt.figure(constrained_layout=True)
```
rather than call `fig.tight_layout()` _a posteriori`.

neutrinoceros:
Excellent. I was afraid this would be marginally better and a lot more expensive to implement/maintain, but the cost seems actually equivalent and MPL core devs seem to be unanimous on their preference, so let's try this

neutrinoceros:
@cphyc turns out this is more than a single line patch, lets see what CI has to say.

neutrinoceros:
oh and one thing that should be elucidated is how well this two options play together when combined. Users may already be passing `box_inches="tight"`, so if it turns out it's not compatible with `constrained_layout=True`, it'll be more work to catch that option and warn consumers not to use it.

neutrinoceros:
So it seems to me that `constrained_layout` produces smaller diffs... but also doesn't fix the problem I was going at as `bbox_inches="tight"` does, e.g. here
Screenshot 2021-11-01 at 15 42 24
(the ylabel is fixed by `box_inches="tight"`, while this figure is unchanged using `constrained_layout=True`)

@cphyc, let me know if you want to have a looks, but otherwise I'll just remove the most recent commit (constrained_layout)

neutrinoceros:
I realized that my modified testing framework produce nice reports but will upload polluted artifacts that are not suited to update the answer store. This needs more work, but I don't know when I'll be able to manage this, so I'm closing this PR for now.

PR #3637, opened 2021-11-01
BUG: fix LaTeX typesetting for fractional ylabels in LinePlot

### LINKED ISSUE #3636: BUG: incorrect LaTeX typesetting in ylabel for fractions
### Bug report

**Bug summary**

Just noticed this in some image reference used for testing:
Screenshot 2021-11-01 at 09 29 15

The bug is that parenthesis should be bigger around a fraction. This is very similar to a bug that was fixed in #3282, but apparently we missed some cases.

neutrinoceros:
## PR Summary

Goal: fix #3636
I'm cargo culting a bit here, so I'll open as a draft.

### DIFF
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -781,7 +781,10 @@ class PlotContainer:
                         symbol_wo_prefix = un[1:]
                         if symbol_wo_prefix in self.ds.unit_registry.prefixable_units:
                             un = un.replace(pp, "{" + latex_prefixes[pp] + "}", 1)
-                axes_unit_labels[i] = r"\ \ (" + un + ")"
+                if r"\frac" in un:
+                    axes_unit_labels[i] = r"\ \ \left(" + un + r"\right)"
+                else:
+                    axes_unit_labels[i] = r"\ \ (" + un + r")"
         return axes_unit_labels
 
     def hide_colorbar(self, field=None):


yt/visualization/profile_plotter.py
```diff
@@ -115,7 +115,32 @@ def data_object_or_all_data(data_source):
...
+            field_name = field_info.get_latex_display_name()
+        elif field_name.find("$") == -1:
+            field_name = field_name.replace(" ", r"\ ")
+            field_name = r"$\rm{" + field_name + r"}$"
+        if fractional:
+            label = field_name + r"$\rm{\ Probability\ Density}$"
+        elif field_unit is None or field_unit == "":
+            label = field_name
+        elif r"\frac" in field_unit:
+            label = field_name + r"$\ \ \left(" + field_unit + r"\right)$"
```
cphyc, line 137: Why this logic bloc here? I thought the result of `(x = 1)` and `\left( x = 1 \right)` were the same?

neutrinoceros: you're right. I just cut-pasted the function without even reading it since it had two identical copies but indeed it could be simplified.

neutrinoceros:
The fact that this passed the answer job shows that it's not yet fixing the targeted problem.

neutrinoceros:
Got the right fix, rebased.

yt/visualization/profile_plotter.py
neutrinoceros: I removed this from the PR because the real fix I was going for is actually a single line diff

matthewturk (dismissed):

cphyc (dismissed):
Happy about the changes if the answers are bumped!

neutrinoceros:
done in https://github.com/yt-project/answer-store/pull/26

neutrinoceros:
@matthewturk, @cphyc, any chance you could merge the answer store PR ?

matthewturk:
merged

matthewturk (approved):

neutrinoceros:

neutrinoceros:
the conflict for this one is in the answer-store submodule. Most likely we need to merge #3644 first and then we'll be able to backport this one.

PR #3638, opened 2021-11-01
ENH: Chimera Frontend 

Ronan-Hix:
## PR Summary

This pull request includes a new frontend for use with the CHIMERA code's data format. This code is a grid-based, non-AMR MHD code with spherical geometry and a logarithmic radial coordinate and is utilized for simulations of Type-II core-collapse supernovae in 2 and 3 dimensions. The data required for the answer tests included in the frontend was previously submitted via curl. Development was performed from the basis of the _skeleton frontend, along with advice from John ZuHone and Matt Turk. If this request should be submitted to a different branch, or if there are any other issues, please let me know.

## PR Checklist


matthewturk:
@Ronan-Hix Thank you!  We'll review this ASAP.

neutrinoceros:
Hi @Ronan-Hix, congratulations on opening this PR, this is an impressive amount of work, especially for a first contribution.

I want to make a couple quick observations:

> logarithmic radial coordinate

warning: This isn't supported at the moment in yt (but it's something I really really want to enable too) !

I see your work is already _almost_ passing listing (pre-commit.ci) (impressive!), but there are a couple issues with file permissions you need to fix.

yt/frontends/chimera/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+
+"""
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
```
neutrinoceros, line 14: We don't use this anymore
```suggestion
```

yt/frontends/chimera/api.py
```diff
@@ -0,0 +1,19 @@
...
+"""
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
+
```
neutrinoceros, line 15:
```suggestion
```

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+"""
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
+
```
neutrinoceros, line 15:
```suggestion
```

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+    @classmethod
+    def _is_valid(self, *args, **kwargs):
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+        try:
+            fileh = HDF5FileHandler(args[0])
+            if "fluid" in fileh:
+                if "agr_c" in fileh["fluid"].keys():
+                    print("Chimera file identified")  # Numpy bless
```
neutrinoceros, line 328: I'm guessing this must be a leftover, but _is_valid methods should not print anything
```suggestion
```

yt/frontends/chimera/fields.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
```
neutrinoceros, line 15:
```suggestion
```

yt/frontends/chimera/fields.py
```diff
@@ -0,0 +1,77 @@
+"""
+Skeleton-specific fields
```
neutrinoceros, line 2: ^^
```suggestion
Chimera-specific fields
```

yt/frontends/chimera/io.py
```diff
@@ -0,0 +1,306 @@
...
+
+    """
+
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# -----------------------------------------------------------------------------
```
neutrinoceros, line 14:
```suggestion
"""
```

yt/frontends/chimera/io.py
```diff
@@ -0,0 +1,306 @@
...
+        for n, chunk in enumerate(chunks):
+            file = _find_files(self.filename)
+            """
+            if any(
+                "grid_2" in f for f in file
+            ):  # Checks for Yin-Yang, not currently important
+                yy = True
+            else:
+                yy = False  # Numpy Bless
+            """
```
neutrinoceros, line 78: Please be careful to not include commented code in your contribution
```suggestion
```

yt/frontends/chimera/io.py
```diff
@@ -0,0 +1,306 @@
...
+                                    (int(dims[2] / n_hyperslabs), dims[1], 1),
+                                )
+                                * (cvel * ecoef * 1e-51)
+                            )
+                        dat_1 = lumin[:, :, :].transpose()
+
+                    elif fname == "shock":
+                        sys.exit(
+                            "Error: Currently disfuctional, please select another option"
+                        )  # This overlay does not currently work, calculation code is left here for future refinement and implementation.
```
neutrinoceros, line 230: I believe this is a left over from an older version of the Skeleton frontend, and in any case this isn't the right way to raise an error, please check this.

yt/frontends/chimera/io.py
```diff
@@ -0,0 +1,306 @@
...
+                        """
```
neutrinoceros, line 231: a bunch of commented code here too. Don't be afraid to clean it up, it will always be accessible in your git history :)

neutrinoceros (changes requested):
a very quick and shallow review
note that you've mistakenly included `yt/frontends/chimera/tests/.DS_Store`. It's light enough that you do not need to rewrite your history but it should be removed.

you can edit your `.git/info/exclude` file to ignore `*.DS_Store` files in the future without having to touch the shared `.gitignore` file

yt/frontends/chimera/api.py
```diff
@@ -0,0 +1,19 @@
+"""
+API for yt.frontends._skeleton
```
neutrinoceros, line 2:
```suggestion
API for yt.frontends.chimera
```
```suggestion
API for yt.frontends._skeleton
```

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+def _find_files(
+    filename_c,
+):  # Returns a list of all files that share a frame number with the input
+    dirname, file = os.path.split(filename_c)
+    dir_contents = os.listdir(dirname)
+    frames = [f for f in dir_contents if file[:-7] in f]
+    directory = os.path.dirname(filename_c)
+    index_filenames = [directory + "/" + f for f in frames]
```
neutrinoceros, line 52: windows compatibility !
```suggestion
    index_filenames = [directory + os.path.sep + f for f in frames]
```

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+        self.meshes = []
+        index_filenames = _find_files(
+            self.dataset.filename
+        )  # Retrieves list of all datafiles with the same frame number
+        if any(
+            "grid_2" in file for file in index_filenames
+        ):  # Detects Yin-Yang data format
+            yy = True
+        else:
+            yy = False
```
neutrinoceros, line 74: This is equivalent to
```suggestion
        # Detects Yin-Yang data format
        yy = any("grid_2" in file for file in index_filenames)
```

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+                    nu = np.cos(old_theta)
+                    de = (-1.0*np.sin(old_theta))*np.cos(old_phi)
+                    coords[:,2] = np.arctan2(nu,de)
+                    """
+
+                """
+                Connectivity is an array of rows, each of which corresponds to a grid cell. The 8 elements of each row are
+                integers representing the cell verticies. These integers refrence the numerical index of the element of the
+                "coords" array whcih corresponds to the spatial coordinate.
+                """
```
neutrinoceros, line 143: Please use actual comments. We do not use multiple lines string literals as comments in yt

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def __init__(
+        self,
+        filename,
+        dataset_type="chimera",
+        storage_filename=None,
+        units_override=None,
+    ):
+        # refinement factor between a grid and its subgrid
+        self.refine_by = 2  # Somewhat superfluous for Chimera, but left to avoid errors
```
neutrinoceros, line 239: the current convention to signal that this value is inert is to set it to 1
```suggestion
        self.refine_by = 1  # Somewhat superfluous for Chimera, but left to avoid errors
```

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+        # should be set, along with examples of how to set them to standard
+        # values.
+        #
+        self.length_unit = self.quan(1.0, "cm")
+        self.mass_unit = self.quan(1.0, "g")
+        self.time_unit = self.quan(1.0, "s")
+        self.time_unit = self.quan(1.0, "s")
+        self.velocity_unit = self.quan(1.0, "cm/s")
+        self.magnetic_unit = self.quan(1.0, "gauss")
+        pass
```
neutrinoceros, line 259:
```suggestion
```

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+                        f["mesh"]["x_ef"][(-1)],
+                        f["mesh"]["y_ef"][(-1)],
+                        f["mesh"]["z_ef"][(-1)],
+                    ]
+            # Sets left and right bounds based on earlier definitions
+            self.domain_right_edge = np.array(dre)
+            self.domain_left_edge = np.array(dle)
+
+        self.cosmological_simulation = 0.0  # Chimera is not a cosmological simulation
+        pass
```
neutrinoceros, line 318:
```suggestion
```

yt/frontends/chimera/data_structures.py
Ronan-Hix: It was indeed. Print statement removed in commit 10df6b4.

yt/frontends/chimera/data_structures.py
Ronan-Hix: Removed in 10df6b4

yt/frontends/chimera/data_structures.py
Ronan-Hix: Removed in 10df6b4

yt/frontends/chimera/data_structures.py
Ronan-Hix: Replaced in 10df6b4

yt/frontends/chimera/data_structures.py
Ronan-Hix: Spherical transforms removed and Connectivity text changed to comments in 10df6b4

yt/frontends/chimera/data_structures.py
Ronan-Hix: This was originally set to 2 early in development because using 1 prompted an error, but it seems this error no longer occurs. Change implemented in 10df6b4.

neutrinoceros:
@Ronan-Hix for next time: there's a "add to batch" button you can use to apply more than one suggestion in a single commit. Not a big deal, but it's useful :)

Ronan-Hix:
> @Ronan-Hix for next time: there's a "add to batch" button you can use to apply more than one suggestion in a single commit. Not a big deal, but it's useful :)

I was just thinking that something like that would be nice. My apologies, I'm a somewhat novice git user.

neutrinoceros:
No apology needed here, and everyone is a novice at some point !
I'll try to make a more in depth review in a few days.

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,309 @@
...
+    def _is_valid(self, *args, **kwargs):
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+        try:
+            fileh = HDF5FileHandler(args[0])
+            if "fluid" in fileh:
+                if "agr_c" in fileh["fluid"].keys():
+                    return True  # Numpy bless
+            else:
+                return False
```
neutrinoceros, line 306: I believe this can be further reduced as follow. Note that in Python, conditions chained with `and` are lazy evaluated, i.e., the second condition will not be evaluated if the first one is False, so we don't need an extra indentation level
```suggestion
            if "fluid" in fileh and "agr_c" in fileh["fluid"].keys():
                return True  # Numpy bless
```

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,309 @@
...
+        self.magnetic_unit = self.quan(1.0, "gauss")
+
+    def _parse_parameter_file(self):
+        with h5py.File(self.parameter_filename, "r") as f:
+            if "metadata" in f:
+                self.unique_identifier = f["metadata"]["date_string"][
+                    ()
+                ]  # Uses date as unique identifier if available, otherwise uses a default value
+            else:
+                self.unique_identifier = "Chimera"
```
neutrinoceros, line 245: this is not unique enough. The idea of this attribute is that two dataset shouldn't be able to use a common value. Note that the parent class `Dataset` has a default implementation that should be unique enough already, you don't need to set it in the frontend

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,309 @@
...
+from yt.geometry.geometry_handler import YTDataChunk
+from yt.geometry.unstructured_mesh_handler import UnstructuredIndex
+from yt.utilities.file_handler import HDF5FileHandler
+from yt.utilities.io_handler import io_registry
+from yt.utilities.on_demand_imports import _h5py as h5py
+
+try:
+    from yt.data_objects.unstructured_mesh import SemiStructuredMesh
+except ModuleNotFoundError:
+    from yt.data_objects.index_subobjects.unstructured_mesh import SemiStructuredMesh
```
neutrinoceros, line 21: I guess this isn't needed anymore ?
```suggestion
from yt.data_objects.index_subobjects.unstructured_mesh import SemiStructuredMesh
```

yt/frontends/chimera/api.py
```diff
@@ -0,0 +1,11 @@
+"""
+API for yt.frontends.chimera
+
+
+
+"""
+
+from . import tests
```
neutrinoceros, line 8: this module is meant to represent the parts of the frontend we want to expose to the user. We can't expose test modules because they are not distributed when yt is released
```suggestion
```

matthewturk:
Hi @Ronan-Hix -- apologies for not getting back to this.  I'm hoping to go through this, but I wanted to ask if you needed help pushing it forward?

Ronan-Hix:
> Hi @Ronan-Hix -- apologies for not getting back to this. I'm hoping to go through this, but I wanted to ask if you needed help pushing it forward?

Hello, no worries. Apologies for my own delay, it's finals season at my institution, which has been impinging on my time; I'm aiming to implement the existing reviews in a week or two, once the semester is out. In the interim, please feel free to make any additional reviews or suggestions

neutrinoceros:
Ah. I allowed CI to run before I remembered that it's currently unstable, so don't panic when you see failures everywhere, they are unrelated !

neutrinoceros:
There are a couple nits to fix (see pre-commit.ci logs as well as the type checking job). I can review this again after that's fixed :)

Ronan-Hix:
> There are a couple nits to fix (see pre-commit.ci logs as well as the type checking job). I can review this again after that's fixed :)

Absolutely. The type checking log error should be easy to correct, but I had been wondering about the pre-commit error. Looking at the log, it only shows the files in which the hook found issues, not the issue itself. Is there a way to see a more detailed log or run the pre-commit locally, to make the corrections it desires?

neutrinoceros:
Sure. There are two ways you can auto fix with pre-commit, either by with a specific comment here (it can be found at the bottom of the pre-commit.ci results page) or by running pre-commit locally, see our developer guide :
https://yt-project.org/docs/dev/developing/developing.html#automatically-checking-and-fixing-code-style

yt/frontends/chimera/tests/test_outputs.py
```diff
@@ -0,0 +1,195 @@
...
+        ("chimera", "ylep"),
+        ("chimera", "z_nuc_rep_c"),
+        ("chimera", "zn60"),
+    ]
+    assert_equal(str(ds), "chimera_002715000_grid_1_01.h5")
+    assert_equal(ds.geometry, "spherical")  # Geometry
+    print(
+        ds.domain_right_edge
+        - ds.arr([1.06500257e09, 3.14159265e00, 6.28318531e00], "code_length")
+    )
```
neutrinoceros, line 161: failed to catch this earlier but please make sure there are no print statements in your tests

yt/frontends/chimera/tests/test_outputs.py
```diff
@@ -0,0 +1,195 @@
...
+        ("chimera", " ca40"),
+        ("chimera", " cr48"),
+        ("chimera", " fe52"),
+        ("chimera", " fe56"),
+        ("chimera", " mg24"),
+        ("chimera", " ne20"),
+        ("chimera", " ni56"),
+        ("chimera", " si28"),
+        ("chimera", " ti44"),
+        ("chimera", " zn60"),
```
neutrinoceros, line 38: Why do these field names contain leading whitespace ? I think these ought to be trimmed off (both in this test and wherever the names are read in the frontend)

yt/frontends/chimera/fields.py
```diff
@@ -0,0 +1,68 @@
...
+
+        def _test(field, data):
+
+            return data["chimera", "rho_c"]
+
+        self.add_field(
+            ("chimera", "test"), sampling_type="cell", function=_test, units="g/cm**3"
+        )
+
+        pass
```
neutrinoceros, line 64:
```suggestion
```

yt/frontends/chimera/tests/test_outputs.py
matthewturk: If these are from the file (which they likely are, since they're `chimera`-type) they likely have that whitespace padding in the file.  This often shows up when reading from fixed-size string arrays, like `|S4` or something, in hdf5.

neutrinoceros: I get that, but is there a reason for keeping that whitespace as part of the API ?

matthewturk: We aren't leaving them as part of the API; this is reflecting the exact field names on disk.  When they get aliased (if they get aliased) they won't be `chimera` field-type anymore.

neutrinoceros: Ah, I see that we're actually having a divergence of opinions. I consider even "raw" fields to be part of the public API because
1) they are de-facto accessible from the user side (and without any additional hurdle, as compared to aliased/derived fields)
2) in the absence of an alias they may be the _only_ way to access some fields

If that's not convincing to you I'll withdraw my suggestion.

matthewturk: I suppose my thought process is that we shouldn't necessarily break the idea of iterating over a list that comes from somewhere else, that might have those specific field names from the file.  I am not unconvinced yet; @Ronan-Hix what do you think?

Ronan-Hix: These do indeed come from disk. They're fixed-length name arrays that come from Fortan and get stored that way in the hdf5 file. I could, I suppose, have it trim the whitespace off the names it reads in when generating the fields list, then add a couple of lines in the io routine that would add them back to match the data path, although a hardcoded solution like that might open up compatibility issues if something changes on the data storage side.

neutrinoceros: Sorry I may be a little slow here but I don't get why you'd want to add it back ?

Ronan-Hix: I'm not necessarily against making that change, my only aversion to adding the whitespace back in the io system is that it might break if the chimera file system ever changes (For instance if the whitespace gets removed from the names on disk), whereas the purely procedural system won't.

neutrinoceros: I think what I'm actually proposing is to map on-disk names to their trimmed equivalent at runtime _and keep a reference to that mapping in the indexing process_ so we can consistently access fields without having to readd any amount of whitespace at any point. I think this is actually pretty resilient to upstream changes. I did something similar in another frontend where on-disk names are sometimes upper case and some other times lower case, but I normalize them to lower case.

Ronan-Hix: Which frontend is that implementation in? I can use it as a reference to add whitespace trimming.

neutrinoceros: that's in an externally maintained frontend for idefix/Pluto

here's the relevant part

https://github.com/neutrinoceros/yt_idefix/blob/3fcf1a01d0a73eb3539eb242e96f64226bca49ba/yt_idefix/_io/vtk_io.py#L247-L250

Ronan-Hix:
pre-commit.ci autofix

neutrinoceros:
@yt-fido test this please

neutrinoceros:
What's the status of this ? It looks like like we just need to make new tests passing, which means uploading your test datasets to the yt hub.

Ronan-Hix:
> What's the status of this ? It looks like like we just need to make new tests passing, which means uploading your test datasets to the yt hub.

I agree, I uploaded the test data via curl at the beginning of this process, but the testing infrastructure doesn't seem to be able to find them. Per a chat with Matt, I did `curl -T [filename].tar.gz http://use.yt/upload/`, is there another step?

neutrinoceros:
There are two, actually.
1) create a PR on the website repo. The most recent example of this is https://github.com/yt-project/website/pull/111
2) once it's merged, come back here and edit the data registry as done here https://github.com/yt-project/yt/pull/3815/commits/9c395aab530cb29284315f4a8f58c35aefb0b1ec

Ronan-Hix:
I merged the PR to the website, so the test data is now visible on the website. As for the second step, where can I get the "hash" entry from?

neutrinoceros:
@Ronan-Hix see pooch's docs https://www.fatiando.org/pooch/latest/hashes.html?highlight=hash#calculating-hashes

Ronan-Hix:
It looks like the test server finds the data now, but has an issue with the permissions. Is this something I can resolve from my side, or is it purely server side?

neutrinoceros:
I _think_ it's server side. Maybe @Xarthisius knows what to do.

Xarthisius:
@yt-fido test this please.

Ronan-Hix:
Alright, it looks like the permissions issue is fixed, thanks. I'll take a look at why the test still fails on the server but doesn't locally.

yt/sample_data_registry.json
```diff
@@ -35,6 +35,12 @@
     "load_name": "BigEndianGadgetBinary",
     "url": "https://yt-project.org/data/BigEndianGadgetBinary.tar.gz"
   },
+  "C15-3D-3deg.tar.gz": {
+    "hash": "b59a4ea80b3e5bf0c8fc1ab960241b23e8cba6d7346cab25d071d9a63c3be2b7",
+    "load_kwargs": {},
+    "load_name": "C15-3D-3deg",
```
neutrinoceros, line 41:
```suggestion
    "load_name": "chimera_002715000_grid_1_01.h5",
```

yt/sample_data_registry.json
```diff
@@ -125,6 +131,12 @@
     "load_name": "ExodusII/gold.e",
     "url": "https://yt-project.org/data/ExodusII_tests.tar.gz"
   },
+  "F37_80.tar.gz": {
+    "hash": "ff0eeb2690ec53f25def74a346fdef07878e46d16424ce4f5d31d8c841408a46",
+    "load_kwargs": {},
+    "load_name": "F37_80",
```
neutrinoceros, line 137:
```suggestion
    "load_name": ""chimera_00001_grid_1_01.h5",
```

neutrinoceros (commented):
I tested `yt.load_sample("F37_80")` and `yt.load_sample("C15-3D-3deg")`, here's how to fix the registry for these to work

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,296 @@
...
+        dataset_type="chimera",
+        storage_filename=None,
+        units_override=None,
+    ):
+        # refinement factor between a grid and its subgrid
+        self.refine_by = 1  # Somewhat superfluous for Chimera, but left to avoid errors
+
+        self.fluid_types += ("chimera",)
+        super().__init__(filename, dataset_type, units_override=units_override)
+        self.filename = filename
```
neutrinoceros, line 217: So, since #3772 was merged, this line is both broken and unneeded. It's why tests are still failing, and I believe they should pass if you just remove it :)
```suggestion
```

Xarthisius:
Just as general note: answer tests from this PR are not being run, since they were not added to `tests/tests.yaml`

yt/frontends/chimera/tests/test_outputs.py
```diff
@@ -0,0 +1,191 @@
...
+            yield GenericArrayTest(ds, field_func, args=[field])
+
+
+Three_D = "C15-3D-3deg/chimera_002715000_grid_1_01.h5"
+
+
+@requires_ds(Three_D)
+def test_3D():
+    ds = data_dir_load(Three_D)
+    _fields = [
```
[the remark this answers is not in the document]
Ronan-Hix: That seems reasonable and less resource intensive; I wasn’t sure exactly how meticulous the tests should be. Just to be clear, you’re suggesting an alteration to the field_func passed to the GenericArrayTest so that it returns an array of the min, max, avg, and size of each field array, rather than the entire data array?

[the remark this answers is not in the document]
Xarthisius: Yes, something like that. Alternatively, you could use a selection (e.g. `ds.sphere`) and only test subset of the data.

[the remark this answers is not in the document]
Ronan-Hix: I tried both options, neither of which seemed to make a large difference in the time required (All three versions ran in ~65 seconds). I've opted for the min/max/avg/size option since it should minimize data size.

Ronan-Hix:
It appears the tests are now running, but the answer tests fail due to a lack of previous results to compare to. I'm not sure how to prompt the automatic system to generate these.

neutrinoceros:
IIRC we just need to invoke the runner again, like so:

@yt-fido test this please

Xarthisius:
@yt-fido test this please.

Xarthisius:
> IIRC we just need to invoke the runner again, like so:

Close, but that's not enough. You need to either remove old answers (which I just did) if you have access to the testing box and then call yt-fido, or simply bump the number in `tests.yaml` to create new answers (that can be done by anyone).

neutrinoceros (approved):
Thanks @Ronan-Hix for your continued effort through this long process. I think it is now reasonable to include this frontend as is, and I have no further requests.

neutrinoceros:
I just remembered that we'll need to update documentation too. This can be done in a follow up PR if you'd like, but here are the files that'll need updates:
doc/source/examining/loading_data.rst
doc/source/reference/api/api.rst
doc/source/reference/code_support.rst

Ronan-Hix:
Thanks for all your assistance during the process. I think it'd be best to include the documentation in this PR; I'll add the relevant new entries as soon as possible and commit them here.

neutrinoceros:
Just to let you know, I am actively looking for other reviewers. Thank you for your patience !

chrishavlin (commented):
Looks great!!

Just a few questions and some suggestions, most of which I don't feel strongly about, and you can take or leave them. The only exception is a stray `sys.exit` in the io routines that should raise an error instead -- after that is fixed I'll hit approve!

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,295 @@
...
+
+        self.cosmological_simulation = 0.0  # Chimera is not a cosmological simulation
+
+    @classmethod
+    def _is_valid(self, *args, **kwargs):
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+        try:
+            fileh = HDF5FileHandler(args[0])
+            if "fluid" in fileh and "agr_c" in fileh["fluid"].keys():
```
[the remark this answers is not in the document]
chrishavlin: thought more about this -- the intersection of "fluid" and the "arg_c" key is probably sufficient. But if the output happens to have metadata that is clearly Chimera-specific, I do think it'd be worth adding a narrower check. If not, this is OK!

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,295 @@
...
+        super().__init__(*args, **kwargs)
+
+
+def _find_files(
+    filename_c,
+):  # Returns a list of all files that share a frame number with the input
+    dirname, file = os.path.split(filename_c)
+    dir_contents = os.listdir(dirname)
+    frames = [f for f in dir_contents if file[:-7] in f]
+    directory = os.path.dirname(filename_c)
```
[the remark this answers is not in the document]
neutrinoceros: good call, I'll add that `os.path.splitext` is probably what we should use instead of selecting chars manually as `file[:-7]`

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,295 @@
...
+                # Generates and fills coordinate array
+                coords = np.zeros((nxd, nyd, nzd, 3), dtype="float64", order="C")
+                coords[:, :, :, 0] = r[:, None, None]
+                coords[:, :, :, 1] = theta[None, :, None]
+                coords[:, :, :, 2] = phi[None, None, :]
+                coords.shape = (nxd * nyd * nzd, 3)
+
+                if yy:
+                    # Converts coordinates to cartesian in the case of Yin-Yang data
+                    x = coords[:, 0] * np.sin(coords[:, 1]) * np.cos(coords[:, 2])
```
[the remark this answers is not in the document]
neutrinoceros: > I can only assume that numpy interprets coords[:, 0] as coords[:, 0, :, :]

I can confirm that this is how it works.

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,295 @@
...
+                dle = [
+                    f["mesh"]["x_ef"][(0)],
+                    f["mesh"]["y_ef"][(0)],
+                    f["mesh"]["z_ef"][(0)],
+                ]
+                if self.domain_dimensions[2] > 2:
+                    dre = [
+                        f["mesh"]["x_ef"][(-3)],
+                        f["mesh"]["y_ef"][(-1)],
+                        f["mesh"]["z_ef"][(13)],
```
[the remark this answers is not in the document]
neutrinoceros: > What's the difference between `[(13)]` and just `[13]`?

As far as I can tell: none. `(13) is 13` evaluates to `True`.

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,295 @@
...
+                    dre = [
+                        f["mesh"]["x_ef"][(-1)],
+                        f["mesh"]["y_ef"][(-1)],
+                        f["mesh"]["z_ef"][(-1)],
+                    ]
+            # Sets left and right bounds based on earlier definitions
+            self.domain_right_edge = np.array(dre)
+            self.domain_left_edge = np.array(dle)
+
+        self.cosmological_simulation = 0.0  # Chimera is not a cosmological simulation
```
neutrinoceros, line 283: btw this is really a boolean in disguise, it should be an `int`, not a `float`

yt/frontends/chimera/io.py
```diff
@@ -0,0 +1,231 @@
...
+                        "num_lumin_1",
+                        "num_lumin_2",
+                        "num_lumin_3",
+                        "num_lumin_4",
+                        "shock",
+                        "nse_c",
+                    )
+
+                    a_name_2 = [i.decode("utf-8") for i in f["abundance"]["a_name"]]
+                    a_name_dict = {name.strip(): name for name in a_name_2}
```
neutrinoceros, line 96: these definitions do not depend on the iterated variable and could be taken out of the loop

yt/frontends/chimera/io.py
```diff
@@ -0,0 +1,231 @@
...
+                    if fname not in specials:
+                        if fname in f["fluid"]:
+                            ds = f["fluid"][f"{fname}"]
+                        elif fname in f["abundance"]:
+                            ds = f["abundance"][f"{fname}"]
+                        elif fname in a_name_dict:
+                            ind_xn = a_name_2.index(a_name_dict[fname])
+                            ds = f["abundance"]["xn_c"][:, :, :, ind_xn]
+                        else:
+                            sys.exit("Error: Invalid field name")
```
neutrinoceros, line 106: Actually I'm not even sure: do we want to this to be unrecoverable, or would a warning suffice ? And in any case, it'd be better to print out the problematic `fname` in the message
```suggestion
                            warning.warn(f"Invalid field name {fname}")
```

yt/frontends/chimera/io.py
```diff
@@ -0,0 +1,231 @@
...
+                        "num_lumin_2",
+                        "num_lumin_3",
+                        "num_lumin_4",
+                    ):
+                        dims = f["mesh"]["array_dimensions"]
+                        n_groups = f["radiation"]["raddim"][0]
+                        n_hyperslabs = f["mesh"]["nz_hyperslabs"][()]
+                        ergmev = 1.602177e-6
+                        cvel = 2.99792458e10
+                        h = 4.13567e-21
```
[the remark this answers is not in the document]
neutrinoceros: I wonder if there is a reason not to use unyt here.
```
import unyt as un

ergmev = float((1*un.MeV)/(1*un.erg))
cvel = float(un.c.to("cm/s"))
h = float(un.h.to("MeV * s"))
```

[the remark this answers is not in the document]
chrishavlin: Ya, I was going to suggest that too! I don't see a reason not to use it... but it could also be an easy follow-up PR to go through this section and unyt-fy it.

yt/frontends/chimera/io.py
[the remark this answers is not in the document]
chrishavlin: note, @neutrinoceros also commented on this, suggest a warning instead of an error, which is fine by me! I'll resolve this convo so we're not commenting in multiple places on this.

chrishavlin: Note -- I also commented on this, but resolved my conservation so we're not discussing it in different comments :) I had suggested raising an error instead of the `sys.exit`. A warning is fine by me as well.

neutrinoceros: Sorry I intended this to be a reply to your comment, I meant to say "I'm not even sure *that this should be an error*".
I think it's sane to have redundant/extra validation, since the model is quite complicated.

Xarthisius:
> I think this should go in. Can I get some folks that have participated in the thread to weigh in here at the top level as to whether or not anything is remaining that needs to be done?

I'd consider my comments as "blocking". Specifically "ying-yang" flavor will just fail or work not as intended. I'd suggest dropping this branch, or fix it + add tests.

Xarthisius (changes requested):
I'd consider my comments as "blocking". Specifically "ying-yang" flavor will just fail or work not as intended. I'd suggest dropping this logical branch, or fix it + add tests.

chrishavlin:
Most of my comments where just style suggestions, and I'm OK with this PR going in without addressing those. Only exception is the `sys.exit()`, I really think that should be changed to either a normal error or a warning.

yt/frontends/chimera/data_structures.py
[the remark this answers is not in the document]
Ronan-Hix: While newer iterations of the code do output metadata, it's not very Chimera specific nor reliably invariant from run to run, not to mention older datasets do not have it. I looked into finding better checks, but this one seemed like the best intersection of reliable presence in Chimera files and somewhat unlikelihood to be found elsewhere. What would be possible is to change the check to look for multiple Chimera filed names, rather than just "agr_c", to decrease the chance of accidentally misidentifying a non-chimera file.

yt/frontends/chimera/fields.py
```diff
@@ -0,0 +1,66 @@
...
+
+# We need to specify which fields we might have in our dataset.  The field info
+# container subclass here will define which fields it knows about.  There are
+# optionally methods on it that get called which can be subclassed.
+
+
+class ChimeraFieldInfo(FieldInfoContainer):
+    known_other_fields: KnownFieldsT = (
+        ("e_int", ("erg", ["Internal Energy"], "Internal Energy")),
+        ("entropy", ("", ["Entropy"], None)),
```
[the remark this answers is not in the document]
Ronan-Hix: The entropy on disk is stored in units of Boltzmann's constant; I couldn't find a good way to feed that into yt. If there is a way to pass this, I would be glad to incorporate it.

yt/frontends/chimera/data_structures.py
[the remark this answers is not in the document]
Ronan-Hix: Removing `directory` and using `dirname` instead sounds sensible. As for using `splitext`, it's my understanding that it only separates the base name and file extension; what `file[:-7]` is doing is also removing the section of the file name that specifies the grid number. That way, the code will match all grid files that correspond to the same time-step (the timestep is specified earlier in the file name). If there's a more elegant way to do this with `splitext` or otherwise, I'm open to it.

yt/frontends/chimera/data_structures.py
```diff
@@ -0,0 +1,295 @@
...
+def _find_files(
+    filename_c,
+):  # Returns a list of all files that share a frame number with the input
+    dirname, file = os.path.split(filename_c)
+    dir_contents = os.listdir(dirname)
+    frames = [f for f in dir_contents if file[:-7] in f]
+    directory = os.path.dirname(filename_c)
+    index_filenames = [directory + os.path.sep + f for f in frames]
+    index_filenames.sort()
+    return index_filenames
```
neutrinoceros, line 40: Here's a direct suggestion to make this function more compact (elegant ?). It doesn't resolve the issue with `file[:-7]` yet. Since you clarified in a previous comment that the intention is *not* to split the extension, can you show what the last 7 chars are supposed to look like so I can suggest how to express it more explicitly in the code ?
```suggestion
def _find_files(filename_c):  
    # Returns a list of all files that share a frame number with the input
    dirname, file = os.path.split(filename_c)
    frames = [f for f in os.listdir(dirname) if file[:-7] in f]
    index_filenames = [os.path.join(dirname, f) for f in sorted(frames)]
    return index_filenames
```

yt/frontends/chimera/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: I added a suggestion separate from this thread

yt/frontends/chimera/data_structures.py
Ronan-Hix: Sure, the standard form of the file names is along the lines of `chimera_02050_grid_1_01.h5` the `grid_1` exists for that Yin-Yang data (with 2 grids). Each 3D grid is stored as several files, with the `01` indexing these portions of the grid. files[:-7] matches all files with the same common first section but for all grid numbers and file index numbers.

neutrinoceros: Cool, I'll edit my suggestion above then

neutrinoceros: done. Note that `import re` should be also added at the top level of the module

yt/frontends/chimera/data_structures.py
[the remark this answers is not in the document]
chrishavlin: Ya, adding another field to check for is a good idea I think! But only if you're already changing other things -- it'd be an easy thing to fix in the unlikely event that there are conflicts.

yt/frontends/chimera/fields.py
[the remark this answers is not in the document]
chrishavlin: Just FYI, I've been looking into this -- it's certainly possible to define a new unit that you could use here. But this shouldn't hold up the PR. adding the entropy units would make a  straightforward follow up PR (that I'm happy to work on, can ping you on it after I workout the details of how to do it).

yt/frontends/chimera/data_structures.py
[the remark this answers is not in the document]
Ronan-Hix: The YinYang implementation has always been problematic. In this case, the coords array got reshaped into 2D before being passed through the conversions, but I changed the code such that the reshape occurs after the conversions, but the implementation is still unreliable. It is probably best to remove the YY support from this merge request and include it in a subsequent merge at some point in the future, once it is working correctly.

yt/frontends/chimera/data_structures.py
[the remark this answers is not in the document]
Ronan-Hix: As for the parentheses, I don't know of any reason they were there. The 13 was an attempt to read the domain of a single chunk of 3D data, but for reliability reasons, I've changed this to incorporate the entire data domain. As for -3, CHIMERA outputs "ghost" radial cells at the end of `z_ef` with no data values, the -3 removes these, but I have altered this implementation to be more stable if there are or are not ghost zones, as they may be removed in the future.

yt/frontends/chimera/data_structures.py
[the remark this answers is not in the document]
Ronan-Hix: I've added a second field check, let me know if the field combination looks unique enough.

yt/frontends/chimera/data_structures.py
Ronan-Hix: I incorporated this method, with the slight alteration that the frame number are not of uniform length, so `\d+` is more appropriate than `\d{5}`

neutrinoceros: makes sense, thank you!

Ronan-Hix:
I think that the failures are actually a product of the slightly changed method of pulling the maximum r-axis limits, but regardless, simply updating the test should still resolve the issue. I'm intending to push another commit soon that will fix this and remove the YinYang implementation from this pull request, as it still needs further development before it is working properly (There are some issues with the rotations in the spherical basis and rasterization problems when doing the rotations in cartesian coordinates).

I was waiting to do so while looking into an issue a collaborator had while using the newest iteration of this frontend in the last couple days; I want to see if this is a package version incompatibility or a true frontend bug I should patch before pushing.

matthewturk:
@Ronan-Hix Hey -- I just want to say that I'm really grateful for you continuing to plug away on this.  Thank you.

matthewturk:
@yt-fido test this please

Ronan-Hix:
> @Ronan-Hix Hey -- I just want to say that I'm really grateful for you continuing to plug away on this. Thank you.

Absolutely! Sorry for the somewhat intermittent progress, I'm working on this in between a couple other primary research projects and working on grad school applications.

Right now the failing tests come from changing the code behavior to omit ghost zones more reliably (which changes the array size and slightly changes the average). These changes are intentional, so the stored answers should be regenerated, but I'm not quite sure how to do that.

chrishavlin:
> These changes are intentional, so the stored answers should be regenerated, but I'm not quite sure how to do that.

In the `tests/tests.yaml` you can change `local_chimera_001: #PR 3638` to `local_chimera_002: #PR 3638` and when you push up, it'll store the current state of the tests as the new answer.

Ronan-Hix:
I wanted to ping this to see if someone with write access can resolve this conflict. As far as I can see, the versions don’t actually conflict, so I’m not sure how to resolve it on my end.

neutrinoceros:
done

neutrinoceros:
We probably need @Xarthisius to sign off on the change he requested

neutrinoceros (approved):
I'll defer to @chrishavlin and @Xarthisius to finalise their reviews and possibly merge, but AFAICT their respective blocking remarks seem to have been addressed.

chrishavlin (approved):
Yup, my suggestions are all taken care of. But let's not merge until @Xarthisius looks it over. 

One minor thing is that there are a couple of empty files that could be deleted: `misc.py`, `definitions.py`.

neutrinoceros:
> One minor thing is that there are a couple of empty files that could be deleted: misc.py, definitions.py.

Right, but that's true for a lot of existing frontends. These modules are actually not required so we could at some point decide to clean empty ones.

neutrinoceros:
I'm going to close/reopen to refresh CI as requirements for merging have been updated

Xarthisius (approved):

neutrinoceros:
Congratulations @Ronan-Hix, and thanks again for your patience. We will include this frontend in the upcoming yt 4.1 release !

Ronan-Hix:
Very glad to hear it, many thanks to everyone involved for all the assistance!

Issue #3631, opened 2021-10-31
ENH: encapsulate axis flipping method


neutrinoceros:
### Bug report

**Bug summary**

`PlotWindow` is missing a method to flip horizontal and vertical axes. There is somewhat hacky workaround on the user side but it isn't perfect, as showcased bellow.

**Code for reproduction**

```python
import yt

ds = yt.load_sample("IsolatedGalaxy")
slc = yt.SlicePlot(ds, "x", "density")
slc.zoom(10)
slc.save("/tmp/zy.png")

ds.coordinates.x_axis["x"] = 2
ds.coordinates.x_axis[0] = 2
ds.coordinates.y_axis["x"] = 1
ds.coordinates.y_axis[0] = 1
slc = yt.SlicePlot(ds, "x", "density")
slc.zoom(10)
slc.save("/tmp/yz.png")
```

**Actual outcome**
native
![zy](https://user-images.githubusercontent.com/14075922/139573696-1ba15bcc-6b89-44c4-8914-fd3e564f83a4.png)
flipped
![yz](https://user-images.githubusercontent.com/14075922/139573698-68819e7a-eb05-41e8-9c73-1ff67b55e7ef.png)

**Expected outcome**

axis labels should be flipped too. Currently this is achievable but it necessitates more work on the user side.
The reason that they are not flipped is because we're changing the coordinate system on the fly, rather than the figure itself, so the result makes sense, but it's not what "flipping axis" means in the plotting context.

Another undesirable effect can be seen when the region plotted isn't squared: the figure size can drastically vary after the axes were flipped, as can be seen in #3630 
This is arguably a different but related issue, so I'll open a separate report.

_Originally posted by @neutrinoceros in https://github.com/yt-project/yt/issues/3630#issuecomment-955628272_

matthewturk:
Should this also include the direction of the normal vector, to control the positive/negative direction of the display axes?

neutrinoceros:
Yes that's also something I think we should support, and I agree that the two actions are related enough that it makes sense to work on both at the same time.

chrishavlin:
A few notes since I spent some time experimenting this yesterday:

The hacky workaround only works for cartesian datasets since `PlotWindow`... Not surprising, but it illustrates how the image display axes are tied in directly with the frb generation and pixelation routines. 

[...]

PR #3640, opened 2021-11-03
ENH: swap_axis kwarg for PlotWindow

chrishavlin:
This PR adds callbacks to adjust axes orientations (in support of #3631), inroducing the `swap_axes`, `flip_horizontal` and `flip_vertical` callbacks.

Example usage:

```python
import yt

slc = yt.SlicePlot(ds, "z", ("enzo", "Density"), width=(0.03, 0.01))
slc.save("/tmp/swap_axes_normal.png")

slc.swap_axes()
slc.save("/tmp/swap_axes_swapped.png")

slc = yt.SlicePlot(ds, "z", ("enzo", "Density"), width=(0.03, 0.01), swap_axes=True)
slc.save("/tmp/swap_axes_swapped_2.png")
```

Results in:

The unswapped reference (`swap_axes_normal.png`):

swap_axes_normal

The two images with swapped axes produced by the above snippet are identical:

swap_axes_swapped

Works with projections as well, works with axis aligned and off-axis plots (though it's probably not super useful or intuitive to use it with off-axis plots). 

The `flip_horizontal` and `flip_vertical` are essentially aliases to matplotlib's `invert_xaxis` and `invert_yaxis`. 

Additionally, the callbacks have been adjusted to work with the new callbacks.

Note that there is potential for some user confusion since the operations apply to image axes and not dataset coordinates and the swap always has to happen first, a related issue #3890 was started to track this so it is resolved before next release.


chrishavlin:
Also, note that I'm not tied to the `swap_axes` terminology. Maybe `flip_axes` is more intuitive?

neutrinoceros:
Looks great ! One quick remark is that it'd be pretty useful to deal with #3632 before we can add image tests here.
I'll try to give more thoughts to this before I can answer your questions here

neutrinoceros:
One more thing since you asked about terminology: it'd be cool if we can come up with method names that leave room and feel in line with a sibling feature that Matt already suggested in the issue. Namely, users might want to change the orientation of the normal vector. Yt's convention is to make it point towards the reader, but it could be flipped easily... though it'd mean that either the vertical or the horizontal axis would be reversed in terms of signs... and we'd need to choose which one... sounds more complicated the more I try to describe it.

yt/visualization/plot_window.py
```diff
@@ -1118,6 +1142,14 @@ def _setup_plots(self):
                 ia = ImageArray(ia)
             else:
                 ia = image
+
+            swap_axes = self._swap_axes
+            aspect = self.aspect
+            if swap_axes:
+                extent = [extent[i] for i in (2, 3, 0, 1)]
+                ia = ia.transpose()
+                aspect = 1.0 / aspect
```
neutrinoceros, line 1151: this is almost definetely going to collide with #3633, I think we need to discuss what `self.aspect` exactly means or should mean.

yt/visualization/plot_window.py
```diff
@@ -1118,6 +1142,14 @@ def _setup_plots(self):
                 ia = ImageArray(ia)
             else:
                 ia = image
+
+            swap_axes = self._swap_axes
+            aspect = self.aspect
+            if swap_axes:
+                extent = [extent[i] for i in (2, 3, 0, 1)]
```
neutrinoceros, line 1149: took me a minute to read but I think this is a place where you'd want to use a `collections.deque` just for the sake of the `deque.rotate` method (provided that's actually what's happening here)
https://docs.python.org/3/library/collections.html#collections.deque.rotate

yt/visualization/plot_window.py
```diff
@@ -1175,7 +1207,10 @@ def _setup_plots(self):
                         )
                     else:
                         ymin, ymax = (float(y) for y in extenty)
-                    self.plots[f].image.set_extent((xmin, xmax, ymin, ymax))
+                    new_extent = [xmin, xmax, ymin, ymax]
+                    if swap_axes:
+                        new_extent = [ymin, ymax, xmin, xmax]
```
neutrinoceros, line 1212: no reason not to preserve the existing type
```suggestion
                    new_extent = (xmin, xmax, ymin, ymax)
                    if swap_axes:
                        new_extent = (ymin, ymax, xmin, xmax)
```

yt/visualization/plot_window.py
```diff
@@ -2131,6 +2177,7 @@ def __init__(
         window_size=8.0,
         buff_size=(800, 800),
         aspect=None,
+        swap_axes=False,
```
neutrinoceros, line 2180: Same as above
```suggestion
        *,
        swap_axes=False,
```

yt/visualization/plot_window.py
```diff
@@ -2510,6 +2564,7 @@ def __init__(
         method="integrate",
         data_source=None,
         buff_size=(800, 800),
+        swap_axes=False,
```
neutrinoceros, line 2567:
```suggestion
        *,
        swap_axes=False,
```

yt/visualization/plot_window.py
```diff
@@ -372,6 +375,26 @@ def _recreate_frb(self):
         for key in self.override_fields:
             self._frb[key]
 
+    @property
+    def _swap_axes(self):
```
neutrinoceros, line 379: I think the following method could be simply named `swap_axes` instead of `toggle_swap_axes`.
How about naming this property `_has_swapped_axes` to avoid ambiguity and clarify it's a boolean ? this name doesn't need to be concise since it's not public API.
```suggestion
    def _has_swapped_axes(self) -> bool:
```
Of course this suggestion isn't self contained, you'd have to adapt the rest of the code too

neutrinoceros (commented):
Just reviewing for API clarity here, haven't yet tested this but it looks promising. Thank you for working on this !

matthewturk:
To build on what @neutrinoceros said, do you think there's a way we could have this also supply a way of having it say, I was the x direction positive, y direction negative, etc.

neutrinoceros:
I think the easiest way would be to implement `.flip_horizontal_axis()` and `.flip_vertical_axis()` methods. in the spirit of what's already done here

yt/visualization/plot_window.py
```diff
@@ -203,6 +205,7 @@ def __init__(
         fontsize=18,
         aspect=None,
         setup=False,
+        swap_axes=False,
```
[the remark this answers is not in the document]
chrishavlin: yes, definitely. will do.

yt/visualization/plot_window.py
chrishavlin: ya, I wasn't even entirely sure what `aspect` does here -- assumed it was an aspect ratio of some kind and so it should also be inverted. So yes, this needs more investigation and discussion.

yt/visualization/plot_window.py
chrishavlin: cool, I'll check out deque. The simpler to read but longer code I had initially was just `extent = (extent[2], extent[3], extent[0], extent[1])`.

yt/visualization/plot_window.py
chrishavlin: yup, good suggestion. I went with `toggle_` because we had already had a `toggle_right_handed` method. But grammatically I like just having `swap_axes`. I'll change this.

neutrinoceros: Oh. I had no idea we had a `toggle_right_handed` method. It looks exactly identical to what I'd expect from the `flip_horizontal_axis` method I suggested already, and it shows the way for how to go about a `flip_vertical_axis` method ! So I'm going to go with a bold proposal here and suggest we alias `toggle_right_handed` to `flip_horizontal_axis` and deprecate the original method because it represents only _one_ of the two possible implementations for its name.

yt/visualization/plot_window.py
chrishavlin: oh, this `aspect` is the `imshow(aspect=` keyword arg... it's a string, either `equal` or `auto` . The actual aspect ratio which we call `_unit_aspect` is calculated from `extent` within `WindowPlotMPL.__init__`. So I shouldn't be touching `aspect` here at all.

chrishavlin: oops, maybe I was wrong about that.... but ya, this needs to be sorted out...

chrishavlin: ok, missed that aspect could be "{'equal', 'auto'} or float" to `imshow`, and we set `aspect=1.` earlier on if it is `None`. So in isolation what I did is right, but ya, need to review it in light of #3633

neutrinoceros: in #3633 I've assumed that the `_unit_aspect` ratio should always be read as vertical/horizontal. I'm open to debate if you feel there's a more intuitive interpretation on the user side.

chrishavlin:
Looking through my changes in light of @matthewturk's suggestions: one possibility is that instead of the checks that I have for the new `swap_axes` attribute, I could have a more general function that applies modifications for the display axes, using @neutrinoceros's `flip_horizontal`/`flip_vertical` grammar:

```python
def _apply_display_axes_modifications(self, extent, ia, aspect):
    if self._swap_axes:
        extent = (extent[2], extent[3], extent[0], extent[1])
        ia = ia.transpose()
        aspect = 1./aspect
    if self._flip_horizontal_axes:
        extent = (extent[1], extent[0], extent[2], extent[3])
        ia = np.fliplr(ia)
    if self._flip_vertical_axes:
        extent = (extent[0], extent[1], extent[3], extent[2])
        ia = np.flipud(ia)
    return extent, ia, aspect
```

Though we *do* already have the `_right_handed` attribute that will call `ax.invert_xaxis()` if set. And calling `invert_xaxis` or `invert_yaxis` has the advantage that subsequent plot call backs end up plotting in the correct orientation.

chrishavlin:
I guess nothing other than unifying the flow and making all the modifications in one place. But having the `flip_`s happen after plot instantiation is fine but we already have the `self._right_handed`, so why add a `flip_horizontal` method? Or were you imagining `flip_horizontal` would replace `toggle_right_handed`?

neutrinoceros:
> Or were you imagining flip_horizontal would replace toggle_right_handed?

yes, that's what I had in mind. This would make room to implement `flip_vertical_axis` too, which is another way to "toggle right handed" so to speak.

chrishavlin:
Ok, that makes sense. And in that case I see the `swap_axes` as fairly separate from`flip_h/v_axis` since `swap_axes` has to happen before the initial call to `imshow` and manually adjust the extent. But including all these changes in one PR does make sense. I'll take a stab at adding the `flip_`s (and I'll get in the remaining suggestions).

chrishavlin:
Latest batch of commits:

* adds flip_horizontal, flip_vertical methods
* depreceates right_handed
* does not allow swap_axes when callbacks are set: I think getting this to work properly with callbacks is a bigger task... so went with just not allowing it for now.
* adds some initial tests:
	* I left the right_handed answer test for now,
        * added the flip_ methods to the test_plotwindow.py ATTR framework
        * added some separate tests for swap_axes since they can't be added to the test_plotwindow.py ATTR framework because they won't work with the callbacks. They also aren't answer tests (yet?)

I'll switch this out of draft mode now. Still need docs, but it's at an OK stage for general review I think.

yt/visualization/plot_window.py
chrishavlin: So in the latest batch of commits, I just created a `_swap_axes_extents` function that just does this very explicitly. I didn't use deque cause I thought it was overkill for such a simple thing. But I don't feel strongly about this.

yt/visualization/plot_window.py
chrishavlin: | `_unit_aspect` ratio should always be read as vertical/horizontal.

So I'm still not positive on this, but I think my feeling is that if the aspect ratio is set and then I  call `slc.swap_axes()` then I would want the aspect ratio to flip as well? i.e., I would find it more intuitive if the following resulted in two images with different aspect ratios:

```
slc = yt.SlicePlot(ds, "z", "density", aspect=0.2)
slc.show()
slc.swap_axes()
slc.show()
```

Though maybe it gets more confusing in the case of setting both in the initial plot call. Because I think I would expect `yt.SlicePlot(ds, "z", "density", aspect=0.2, swap_axes=True)` to NOT invert the aspect ratio...

yt/visualization/plot_window.py
chrishavlin: latest batch should resolve this!

yt/visualization/plot_window.py
```diff
@@ -182,9 +182,19 @@ class PlotWindow(ImagePlotContainer):
         The size of the window on the longest axis (in units of inches),
         including the margins but not the colorbar.
     right_handed : boolean
+        Depreceated, please use flip_horizontal.
```
neutrinoceros, line 185: I see that you're already issueing a deprecation warning when it's used (great !), but I think we should also raise an error if it's combined with its replacements.

yt/visualization/plot_window.py
```diff
@@ -182,9 +182,19 @@ class PlotWindow(ImagePlotContainer):
...
         Whether the implicit east vector for the image generated is set to make a right
         handed coordinate system with a north vector and the normal vector, the
         direction of the 'window' into the data.
+    swap_axes: boolean
+        If True, will transpose the final image and swap axes labels. default False.
+        Note that if the swap_axes attribute is set, the behavior of all image
+        modifications (e.g., pan or set_origin) are unchanged and remain relative
+        to the underlying coordinate system and not the image display axes.
```
neutrinoceros, line 193: For the record, this seems fine to me. However I think there is currently a slight inconsistency in that `aspect` is still going to parametrise height/width regardless of this flag's state (that is, unless I'm reading the code wrong)

yt/visualization/plot_window.py
```diff
@@ -182,9 +182,19 @@ class PlotWindow(ImagePlotContainer):
...
+    flip_horizontal: boolean
+        If True, will flip the horizontal axis. default False.
```
neutrinoceros, line 195: Maybe use a different verb in the explanation in case it is not clear. You could mention that it's equivalent to `Axes.invert_xaxis` in matplotlib

yt/visualization/plot_window.py
```diff
@@ -833,7 +875,26 @@ def set_axes_unit(self, unit_name):
...
+            "toggle_right_handed has been deprecated, use '.flip_horizontal()' instead.",
+            since="4.1.0",
+            removal="4.2.0",
+        )
+        self.flip_horizontal()
+
+    @invalidate_plot
+    def flip_horizontal(self):
+        """
+        flips the horizontal axis (the image's abscissa)
```
neutrinoceros, line 888: Same as above, maybe something like
```suggestion
        Revert the direction of the horizontal axis (the image's abscissa)
```

yt/visualization/plot_window.py
```diff
@@ -833,7 +875,26 @@ def set_axes_unit(self, unit_name):
...
+        """
+        self._flip_horizontal = not self._flip_horizontal
+
+    @invalidate_plot
+    def flip_vertical(self):
+        """
+        flips the vertical axis (the image's ordinate)
```
neutrinoceros, line 895: same as above

yt/visualization/plot_window.py
```diff
@@ -372,6 +388,31 @@ def _recreate_frb(self):
         for key in self.override_fields:
             self._frb[key]
 
+    @property
+    def _has_swapped_axes(self):
+        # note: we always run the validations here always in case the states of
```
neutrinoceros, line 393:
```suggestion
        # note: we always run the validations here in case the states of
```

yt/visualization/plot_window.py
```diff
@@ -833,7 +875,26 @@ def set_axes_unit(self, unit_name):
 
     @invalidate_plot
     def toggle_right_handed(self):
-        self._right_handed = not self._right_handed
+        issue_deprecation_warning(
+            "toggle_right_handed has been deprecated, use '.flip_horizontal()' instead.",
```
neutrinoceros, line 879:
```suggestion
            "the `toggle_right_handed` method is deprecated, use `flip_horizontal` instead.",
```

yt/visualization/plot_window.py
```diff
@@ -2780,3 +2912,38 @@ def plot_2d(
         aspect=aspect,
         data_source=data_source,
     )
+
+
+def _swap_axes_extents(extent):
+    """
+    swaps the x and y extent values, preservering type of extent
```
neutrinoceros, line 2919:
```suggestion
    swaps the x and y extent values, preserving type of extent
```

yt/visualization/plot_window.py
```diff
@@ -2780,3 +2912,38 @@ def plot_2d(
...
+
+    Parameters
+    ----------
+    extent : tuple or list
```
neutrinoceros, line 2923:
```suggestion
    extent : sequence of four unyt quantities
```

yt/visualization/plot_window.py
```diff
@@ -2780,3 +2912,38 @@ def plot_2d(
         aspect=aspect,
         data_source=data_source,
     )
+
+
+def _swap_axes_extents(extent):
```
neutrinoceros, line 2917: This looks like the appropriate signature here, might as well use it (you'll need to import `TypeVar` from the `typing` module)
```suggestion
T = TypeVar("T", tuple, list)

def _swap_axes_extents(extent: T) -> T:
```

yt/visualization/plot_window.py
```diff
@@ -2780,3 +2912,38 @@ def plot_2d(
...
+    -------
+    tuple or list
+        the extent axes swapped, now with (ymin, ymax, xmin, xmax).
+
+    """
+    input_type = type(extent)
+    extent = [extent[2], extent[3], extent[0], extent[1]]
+    if input_type is tuple:
+        return tuple(extent)
+    return extent
```
neutrinoceros, line 2937:
```suggestion
    extent_swapped = [extent[2], extent[3], extent[0], extent[1]]
    return type(extent)(extent_swapped)
```

yt/visualization/plot_window.py
```diff
@@ -2780,3 +2912,38 @@ def plot_2d(
...
+
+
+def _check_right_handed(right_handed: bool) -> bool:
+    # temporary function to check if right_handed kwarg has been set. can
+    # remove this after full depreciation.
+    if right_handed is False:
```
neutrinoceros, line 2943:
```suggestion
    if not right_handed:
```

yt/visualization/plot_window.py
```diff
@@ -2780,3 +2912,38 @@ def plot_2d(
...
+        issue_deprecation_warning(
+            "'right_handed' has been deprecated, use 'flip_horizontal' instead.",
```
neutrinoceros, line 2945:
```suggestion
            "The 'right_handed' argument is deprecated. Use 'flip_horizontal' instead.",
```

yt/visualization/tests/test_plotwindow.py
```diff
@@ -804,3 +808,61 @@ def test_dispatch_plot_classes():
     assert isinstance(p2, OffAxisProjectionPlot)
     assert isinstance(s1, AxisAlignedSlicePlot)
     assert isinstance(s2, OffAxisSlicePlot)
+
+
+def test_swap_axes():
+
+    # note: this tests the swap_axes functionality separately from the
+    # ATTR_ARGS framework because swap_axes currently is not valid with
+    # callbacks.
```
neutrinoceros, line 817: I don't know what's the ATTR_ARGS framework. Also, I thought this _was_ compatible with callbacks. It'd be a shame to introduce this functionality without that compatibility, so what's your approach ? how much work do you think it would take and are you positive that it could be done before the 4.1 release ?

yt/visualization/tests/test_plotwindow.py
```diff
@@ -804,3 +808,61 @@ def test_dispatch_plot_classes():
...
+
+    ds = fake_random_ds(16)
+    field = ds.field_list[0]
+    wid = (0.4, 0.2)
+    normal = "x"
+
+    # get the reference plot and store some values
+    slc = SlicePlot(ds, normal, field, width=wid)
```
neutrinoceros, line 825:
```suggestion
    width = (0.4, 0.2)
    normal = "x"

    # get the reference plot and store some values
    slc = SlicePlot(ds, normal, field, width=width)
```

yt/visualization/tests/test_plotwindow.py
```diff
@@ -804,3 +808,61 @@ def test_dispatch_plot_classes():
...
+    assert slc._has_swapped_axes is False
+
+    slc._setup_plots()
+    ref_x_label = slc[field].axes.get_xlabel()
+    ref_y_label = slc[field].axes.get_ylabel()
+    ref_aspect = slc[field]._aspect
+
+    # check that swap_axes swaps the axes
+    slc = SlicePlot(ds, normal, field, swap_axes=True, width=wid)
```
neutrinoceros, line 834:
```suggestion
    slc = SlicePlot(ds, normal, field, swap_axes=True, width=width)
```

yt/visualization/tests/test_plotwindow.py
```diff
@@ -804,3 +808,61 @@ def test_dispatch_plot_classes():
...
+    slc = SlicePlot(ds, "x", ds.field_list[0], swap_axes=True)
+    slc.annotate_cell_edges()
+    assert slc._has_swapped_axes is False
+
+
+def test_extent_swap():
+    input_extent = [1, 2, 3, 4]
+    expected = [3, 4, 1, 2]
+    assert _swap_axes_extents(input_extent) == expected
+    assert _swap_axes_extents(tuple(input_extent)) == tuple([3, 4, 1, 2])
```
neutrinoceros, line 868:
```suggestion
    assert _swap_axes_extents(tuple(input_extent)) == tuple(expected)
```

neutrinoceros (commented):
Overall this looks great. Here are a couple suggestions. I want to review this again later when I have time to test it manually, in particular I'd like to check that it behaves well with spherical and polar geometries.

One more thing I'd like to discuss on a higher level is wether this should ship without the callback compatibility (currently missing if I'm reading this right).

neutrinoceros:
I forgot to mention that the current implementation apparently breaks some answer tests but I don't think it should. Deprecating the `right_handed` argument is the right thing to do, but we can't remove the corresponding instance attribute earlier than yt 4.2

Another remark is that I think this feature would be a nice opportunity to stop making the base plot classes even larger than they already are: I'm thinking maybe the axis-swapping and flipping methods could justify the addition of a small `PlotAxisHandler` class that would be home to essentially every piece of functionality you are adding here.

chrishavlin:
Thanks @neutrinoceros ! I'll take a look at the suggestions, but some general things:

The broken answer tests are probably because I added the `flip_horizontal` and `flip_vertical` tests to the list of plot attritbutes that get tested by `test_plotwindow.py:test_attributes`, so they won't have any answers in the store. 

| One more thing I'd like to discuss on a higher level is wether this should ship without the callback compatibility (currently missing if I'm reading this right).

So the `flip_horizontal` and `flip_vertical` work fine with the callbacks as they modify the underlying matplotlib axes, so any callbacks that plot on top use those updated axes. It's just the `swap_axes` bit that is harder: because it transposes the image data, every callback would have to be changed to check for `swap_axes` and then reverse the arguments used in the callback `plot` call, so my initial inclination was to do that in a separate PR. But I'll think through the idea of a more genreal `PlotAxisHandler` class, maybe there's a way to simplify the callback modifications that would be required with such a class.

neutrinoceros:
> The broken answer tests are probably because I added the flip_horizontal and flip_vertical tests to the list of plot attritbutes that get tested by test_plotwindow.py:test_attributes, so they won't have any answers in the store.

Sorry if I wasn't clear, but this isn't actually why they broke, see the logs:
```
AttributeError: 'AxisAlignedSlicePlot' object has no attribute 'right_handed'
```

> and then reverse the arguments used in the callback plot call, so my initial inclination was to do that in a separate PR. But I'll think through the idea of a more genreal PlotAxisHandler class, maybe there's a way to simplify the callback modifications that would be required with such a class.

Yeah I think the class I suggested to hold the state of the flags and implement the swap and flip methods as well as validation methods. It's probably a good idea to make sure the design can be extended to support callbacks right now instead of later when the dust has settled and we're not sure if we can move code around without breaking stuff.

yt/visualization/plot_window.py
```diff
@@ -372,6 +388,31 @@ def _recreate_frb(self):
...
+        return self._validate_swap_axes(self._swap_axes_input)
+
+    @invalidate_data
+    def swap_axes(self):
+        # toggles the swap_axes behavior
+        new_swap_value = not self._swap_axes_input
+        # note: we also validate here to catch invalid states immediately, even
+        # though we validate on accessing the attribute in `_has_swapped_axes`.
+        self._swap_axes_input = self._validate_swap_axes(new_swap_value)
+
```
neutrinoceros, line 404:
```suggestion
        self._swap_axes_input = self._validate_swap_axes(new_swap_value)
        return self

```

yt/visualization/plot_window.py
```diff
@@ -833,7 +875,26 @@ def set_axes_unit(self, unit_name):
...
+        )
+        self.flip_horizontal()
+
+    @invalidate_plot
+    def flip_horizontal(self):
+        """
+        flips the horizontal axis (the image's abscissa)
+        """
+        self._flip_horizontal = not self._flip_horizontal
+
```
neutrinoceros, line 891:
```suggestion
        self._flip_horizontal = not self._flip_horizontal
        return self

```

yt/visualization/plot_window.py
```diff
@@ -833,7 +875,26 @@ def set_axes_unit(self, unit_name):
...
+    @invalidate_plot
+    def flip_vertical(self):
+        """
+        flips the vertical axis (the image's ordinate)
+        """
+        self._flip_vertical = not self._flip_vertical
```
neutrinoceros, line 898:
```suggestion
        self._flip_vertical = not self._flip_vertical
        return self

```

neutrinoceros (commented):
I tested the methods locally and found that they do not currently return the plot object (as do most if not all plot modifying methods). This set of changes will make the UX much better (and consistent) in a REPL such as Jupyter notebooks !

neutrinoceros:
Note that the grammar you're implementing here enables arbitrary 90° rotations of the plot window. This seems easy to document :)

## raw (baseline)
raw

## vertical flip + swap = rotate 90° (clockwise)
flip_vertical-swap_axes

## double flip = rotate 180°
flip_horizontal-flip_vertical

## horizontal flip + swap = rotate 90° (anti-clockwise)
flip_horizontal-swap_axes

### script for reproduction
```python
from itertools import combinations
import yt

slc = yt.SlicePlot(ds, "x", "density")
slc.zoom(50)
slc.save(f"/tmp/raw.png")
for ops in combinations(["flip_horizontal", "flip_vertical", "swap_axes"], 2):
    slc = yt.SlicePlot(ds, "x", "density")
    slc.zoom(50)
    for op in ops:
        getattr(slc, op)()
    slc.save(f"/tmp/{ops[0]}-{ops[1]}.png")
```

neutrinoceros:
log track of what was failing here
@yt-fido test this please

chrishavlin:
in the latest batch of commits:

1. swap_axes with callbacks! most callbacks seem to be working!
2. latest code suggestions from @neutrinoceros were added
3. documentation update (with description of using the new methods in combination to rotate views)

some notes for me to return to:

1. think further about `aspect`
2. fix some callbacks: 
- `LinePlotCallback`: behaving oddly...
- `TriangleFacetsCallbacks`: needs a more careful look (for now, I have it raising a NotImplementedError if swap_axes is enabled). 
- `Quiver` callbacks: wont work with the `flip_` commands without modification (https://github.com/matplotlib/matplotlib/issues/11857), so currently ignoring flip requests if there is a Quiver related callback. 
3. I need to think more carefully about the the `coord_system` specification (e.g., [docs here](https://yt-project.org/doc/visualizing/callbacks.html#coordinate-systems-in-callbacks) ) in context of axis swaps and flips, make sure it behaves logically.
4. Consider additional tests for callbacks with swap and flip
5. think more about a general axes modification container

I'll link to a test script that runs all of the callbacks from https://yt-project.org/doc/visualizing/callbacks.html#overplot-halo-annotations with and without `swap_axes` and some `flip_`, saving images for each.

yt/visualization/plot_window.py
chrishavlin: went with "inverts"

yt/visualization/plot_window.py
```diff
@@ -1127,17 +1223,21 @@ def _setup_plots(self):
...
-            if not self._right_handed:
-                ax = self.plots[f].axes
-                ax.invert_xaxis()
+            # if self._has_flipped_horizontal:
+            #     ax = self.plots[f].axes
+            #     ax.invert_xaxis()
+            #
+            # if self._has_flipped_vertical:
+            #     ax = self.plots[f].axes
+            #     ax.invert_yaxis()
```
chrishavlin, line 1240: oops, forgot to delete. will remove these in the next commit batch.

yt/visualization/plot_window.py
chrishavlin: oops, also missed this one, next batch will have it.

yt/visualization/plot_window.py
chrishavlin: this one also slipped by, will add to the next batch.

chrishavlin:
here's the aforementioned link to a script that runs through most of the callbacks and saves reference and swapped images https://gist.github.com/chrishavlin/2fd39c054b481b47a3a62e216a02eb70 

Here's the output for the streamlines:
## reference
streamlines
## swapped
streamlines_swapped
## swapped and flipped
streamlines_swapped_flip_h

yt/visualization/plot_window.py
chrishavlin: thanks for catching that! these all return `self` now.

yt/visualization/tests/test_plotwindow.py
chrishavlin: latest commit gets this working with most callbacks.

yt/visualization/plot_window.py
neutrinoceros: did you ? I don't see a change here, maybe you forgot to push one last time ?

doc/source/visualizing/plots.rst
```diff
@@ -706,29 +706,98 @@ two element tuples.
...
    slc.save()
 
-Flipping the plot view axes
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Adjusting the plot view axes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 By default, all :class:`~yt.visualization.plot_window.PlotWindow` objects plot
 with the assumption that the eastern direction on the plot forms a right handed
 coordinate system with the ``normal`` and ``north_vector`` for the system, whether
```
neutrinoceros, line 714: I think that this sentence is not in fact complete, because we don't mention that the normal direction is by default pointing towards the reader. Also we're mixing two paradigms (north/east VS normal direction). Overall, seems like a good opportunity to fix this.

yt/visualization/_commons.py
```diff
@@ -96,3 +96,55 @@ def get_canvas(figure, filename):
...
+    tuple
+        args  with order of pairs switched, i.e,:
+
+        _swap_arg_pair_order(x, y, px, py) returns:
+            y, x, py, px
+
+    """
+
+    assert len(args) % 2 == 0
+    n_pairs = int(len(args) / 2)
```
neutrinoceros, line 144: use euclidian division directly
```suggestion
    n_pairs = len(args) // 2
```

yt/visualization/base_plot_types.py
```diff
@@ -44,7 +44,10 @@ def __init__(self, viewer, window_plot, frb, field, font_properties, font_color)
         self._axes = window_plot.axes
         self._figure = window_plot.figure
         if len(self._axes.images) > 0:
-            self.image = self._axes.images[0]
+            self.raw_image_shape = self._axes.images[0]._A.shape
+            if viewer._has_swapped_axes:
+                # store the original un-transposed shape
+                self.raw_image_shape = self.raw_image_shape[1], self.raw_image_shape[0]
```
neutrinoceros, line 50: haven't tried this yet but *thank you* for working on this !

yt/visualization/plot_modifications.py
```diff
@@ -65,6 +66,7 @@ def __init__(self, *args, **kwargs):
         pass
 
     def __call__(self, plot):
+        # plot is a CallbackWrapper instance
```
neutrinoceros, line 69: This comment reads like a very much missing type annotation (indeed the type of this argument is far from clear from the name). I think there are a couple more places where you could use such a type hint in this module but I won't add individual suggestions everywhere, your IDE is probably a much more confortable place to get them all.
```suggestion
    def __call__(self, plot: CallbackWrapper):
```

yt/visualization/plot_modifications.py
```diff
@@ -26,6 +26,7 @@
 )
 from yt.utilities.math_utils import periodic_ray
 from yt.utilities.on_demand_imports import NotAModule
+from yt.visualization._commons import _swap_arg_pair_order, _swap_axes_extents
 from yt.visualization.image_writer import apply_colormap
```
neutrinoceros, line 30:
```suggestion
from yt.visualization._commons import _swap_arg_pair_order, _swap_axes_extents
from yt.visualization.base_plot_types import CallbackWrapper
from yt.visualization.image_writer import apply_colormap
```

yt/visualization/plot_modifications.py
```diff
@@ -290,6 +295,52 @@ def _set_font_properties(self, plot, labels, **kwargs):
                 label.set_color(plot.font_color)
             label.set_fontproperties(local_font_properties)
 
+    def _set_plot_limits(self, plot, extent=None):
```
neutrinoceros, line 298:
```suggestion
    def _set_plot_limits(self, plot, extent=None) -> None:
```

yt/visualization/_commons.py
```diff
@@ -96,3 +96,55 @@ def get_canvas(figure, filename):
...
+    -------
+    tuple
+        args  with order of pairs switched, i.e,:
+
+        _swap_arg_pair_order(x, y, px, py) returns:
+            y, x, py, px
+
+    """
+
+    assert len(args) % 2 == 0
```
neutrinoceros, line 143: `assert` statements shouldn't be used as part of the business logic because they can be ignored at runtime with `python -O`. That said, I think it may be okay in the context of a private function, unless it's supposed to resurface problems with user inputs, which I suspect is the case here ?
I'll let you decide, just wanted to make you this was an informed choice.
```suggestion
   if len(args) % 2 != 0:
       raise TypeError(...)
```

yt/visualization/plot_modifications.py
```diff
@@ -290,6 +295,52 @@ def _set_font_properties(self, plot, labels, **kwargs):
...
+        ----------
+        plot : CallbackWrapper
+            a CallbackWrapper instance
+        *args
+            x, y plot arguments, must have an even number of *args
+
+        Returns
+        -------
+        tuple
+            either the original args or new args  with x, y pairs switched. i.e.,
```
neutrinoceros, line 333:
```suggestion
            either the original args or new args, with (x, y) pairs switched. i.e.,
```

yt/visualization/plot_modifications.py
```diff
@@ -2322,6 +2377,11 @@ def __call__(self, plot):
         l_cy[1] = self._convert_to_plot(plot, l_cy[1])
         # convert back to shape (nlines, 2, 2)
         l_cy = np.rollaxis(l_cy, 2, 0)
+        if plot._swap_axes:
+            # this one needs a little more thought
+            raise NotImplementedError(
+                "swap_axes is not valid with TriangleFacetsCallback"
+            )
```
neutrinoceros, line 2384: Wow, looks like you did extensive verifications here, that's awesome !

yt/visualization/plot_modifications.py
```diff
@@ -3065,6 +3125,10 @@ def __call__(self, plot):
         lic_data = lic_data / lic_data.max()
         lic_data_clip = np.clip(lic_data, self.lim[0], self.lim[1])
 
+        if plot._swap_axes:
+            lic_data_clip = lic_data_clip.transpose()
```
neutrinoceros, line 3129: this line doesn't look right. I can't convince myself that transposition is equivalent to axes swapping (though it may _look_ equivalent with highly symmetric data)

yt/visualization/plot_modifications.py
```diff
@@ -3177,12 +3240,16 @@ def __call__(self, plot):
         im_buffer = np.zeros((ny, nx, 4), dtype="uint8")
         im_buffer[im > 0, 3] = 255
         im_buffer[im > 0, :3] = self.color
+
+        extent = self._plot_bounds(plot)
+        if plot._swap_axes:
+            im_buffer = im_buffer.transpose((1, 0, 2))
```
neutrinoceros, line 3246: same as above

yt/visualization/plot_window.py
```diff
@@ -876,25 +873,54 @@ def set_axes_unit(self, unit_name):
...
     @invalidate_plot
     def flip_vertical(self):
         """
         flips the vertical axis (the image's ordinate)
         """
         self._flip_vertical = not self._flip_vertical
+        return self
+
+    @property
+    def _has_flipped_vertical(self):
```
neutrinoceros, line 900:
```suggestion
    def _has_flipped_vertical(self) -> bool:
```

yt/visualization/plot_window.py
```diff
@@ -876,25 +873,54 @@ def set_axes_unit(self, unit_name):
...
+        return self._check_flip_quiver_conflict(self._flip_vertical)
+
+    @property
+    def _has_flipped_horizontal(self):
```
neutrinoceros, line 904:
```suggestion
    def _has_flipped_horizontal(self) -> bool:
```

yt/visualization/plot_window.py
```diff
@@ -876,25 +873,54 @@ def set_axes_unit(self, unit_name):
...
+        return self._check_flip_quiver_conflict(self._flip_horizontal)
+
+    @property
+    def _has_quiver_callback(self):
```
neutrinoceros, line 908:
```suggestion
    def _has_quiver_callback(self) -> bool:
```

yt/visualization/plot_window.py
```diff
@@ -876,25 +873,54 @@ def set_axes_unit(self, unit_name):
...
+        q_names = ["Quiver", "CuttingQuiver", "MagField", "Velocity"]
+        callback_list = [c[0] for c in self._callbacks]
+        has_q = any([f"{q}Callback" in callback_list for q in q_names])
+        return has_q
+
+    def _check_flip_quiver_conflict(self, flip_axis):
```
neutrinoceros, line 914: not sure about this one (which is good motivation to add annotations, actually)
```suggestion
    def _check_flip_quiver_conflict(self, flip_axis:bool) -> bool:
```

yt/visualization/tests/test_commons.py
```diff
@@ -36,3 +41,17 @@ def test_custom_valid_ext(name, suffix, expected):
...
+    expected = [3, 4, 1, 2]
+    assert _swap_axes_extents(input_extent) == expected
+    assert _swap_axes_extents(tuple(input_extent)) == tuple(expected)
+
+
+def test_swap_arg_pair_order():
+    assert _swap_arg_pair_order(1, 2) == (2, 1)
+    assert _swap_arg_pair_order(1, 2, 3, 4, 5, 6) == (2, 1, 4, 3, 6, 5)
+    assert_raises(AssertionError, _swap_arg_pair_order, 1)
+    assert_raises(AssertionError, _swap_arg_pair_order, 1, 2, 3)
```
neutrinoceros, line 57: If you're actually _testing_ this then you just convinced me it _should_ be changed to a `TypeError` internally

neutrinoceros (commented):
Look like you've added compatibility with most callbacks, this is awesome ! Here are another bunch of suggestions, mostly about type hints that I think would have helped me reading this work. I'll make a (final ?) round of review when these are dealt with and I'll try to make it crash to see if there's anything missing. Again, great work, and a fantastic addition IMO

neutrinoceros:
Note that flake8 is catching undefined names (missing imports, really)

doc/source/visualizing/plots.rst
```diff
@@ -706,29 +706,98 @@ two element tuples.
...
+   slc.flip_vertical()
+   slc.annotate_title("180 Degree Rotation")
+   slc.save("FlipAxes180.png")
+
+   # horizontal flip + swap = rotate 90 degree rotation (counter clockwise)
+   slc = yt.SlicePlot(ds, "x", ("gas", "velocity_x"))
+   slc.flip_horizontal()
+   slc.swap_axes()
+   slc.annotate_title("90 Degree Counter Clockwise Rotation")
+   slc.save("SwappedAxes90CCW.png")
```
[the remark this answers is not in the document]
chrishavlin: |  These examples worked because I reset the figure's state every time, but here you're applying all operations in a single script without regenerating the initial figure, so I don't think it'll work. 

Is this a docs-specific thing? Seemed to work fine while testing the code itself (but I didn't re-build the docs).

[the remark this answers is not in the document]
neutrinoceros: Oh sorry, reading this again I see that you've actually already done what I suggested... sorry about that, carry on!

yt/visualization/_commons.py
```diff
@@ -1,6 +1,6 @@
 import os
 import sys
-from typing import Optional, Type
+from typing import Optional, Type, TypeVar
```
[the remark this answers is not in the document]
chrishavlin: You added `Tuple` here but not anywhere else -- were you going to suggest that `T = TypeVar("T", tuple, list)` should use `T = TypeVar("T", Tuple, list)`?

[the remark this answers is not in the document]
neutrinoceros: I don't know, maybe that's what I had in mind, but I just checked that the simplest version (using builtin types) correctly type checks even with older versions of Python so it is indeed preferable as is ! Feel free to resolve this thread

chrishavlin:
Latest batch just incorporates the latest suggestions, fixes the merge conflict and linting issues, still to do :

* double check the transpose used in some of the callbacks when using `swap_axes`
* improve the intro sentences in the docs
* consider `aspect` ratio question further
* see if the remaining callbacks that do not work can be addressed (Quiver, TriangleFacets)

One question for @neutrinoceros : what's the proper version number for the depreceation warnings here (for the `right_handed` toggle)? Right now it uses depreceated since 4.1.0, removal 4.2.0. Should those numbers be different?

neutrinoceros:
> One question for @neutrinoceros : what's the proper version number for the depreceation warnings here (for the right_handed toggle)? Right now it uses depreceated since 4.1.0, removal 4.2.0. Should those numbers be different?

`since=4.1` is definitely correct, and scheduling removal for yt 4.2 seems fine to me.

yt/visualization/plot_window.py
neutrinoceros: I forgot to reply here: your solution is great, I agree that using a duque would be an overkill

neutrinoceros:
I resolved a bunch of outdated threads. I left only couple open that I'm not sure were actually addressed, but feel free to close them if they were.

About the aspect issue: I think I now agree with how this works, and anyway I believe any behaviour is fine as long as it's documented as intentional here.

There is still one error that's causing a couple dozen tests to fail on Jenkins, please make sure to fix this.

chrishavlin:
converting back to draft for now -- going to be pushing up a bunch of commits that will probably still need some work.

neutrinoceros:
Alright. Feel free to request my review again whenever :)

doc/source/visualizing/plots.rst
```diff
@@ -769,29 +769,98 @@ two element tuples.
...
-   # switching to right-handed coordinates
-   slc.toggle_right_handed()
-   slc.annotate_title("Right Handed")
-   slc.save("Standard.png")
+   # flip the vertical axis
+   slc = yt.SlicePlot(ds, "x", ("gas", "velocity_x"), flip_vertical=True)
+   slc.annotate_title("Flipped vertical")
+   slc.save("FlippedVertical.png")
+
+Note that all three operations effectively toggle the plot's right-handedness.
```
neutrinoceros, line 808: Ah, I'm sorry, I know that *I* suggested this sentence but I don't think it holds actually. What _does_ change is. he direction in which the normal vector is pointing. By default, it's pointing toward the reader, and applying an odd number of these operations will make it point away from the reader.

Right-handedness isn't actually affected by any of these operations. But more importantly, I think I just convinced myself that right-handedness was never a relevant concept for 2D plots, because in the absence of opacity (effectively what `ProjectionPlot` does), it is irrelevant whether you want to look at a transparent object from the front or the from the back.

matthewturk:
Can you provide a suggestion?  If so, I think we can accept it, right?

neutrinoceros:
I'd suggest simply dropping mentions of right-handedness in the docs. Otherwise I'm bound to accept the PR eventually, just waiting for @chrishavlin to tell us when it's ready !

chrishavlin:
Good points! Happy to remove the right-handedness references. Also hopefully will be ready for review soon -- I've been working on some simpler test cases to convince myself that the swapping and flipping work as intended for all the callbacks and found a few issues to fix. More soon!

yt/visualization/plot_modifications.py
neutrinoceros: closing this thread as not actionnable. Results are correct; it doesn't matter that the code is slightly confusing to me

neutrinoceros:
There are now a couple trivial conflicts, but I'll refrain from fixing them for you as I don't want to mingle with your branch's history.

neutrinoceros:
I report that I (finally) tested this with non-cartesian datasets and it seems to be working as expected. Possibly some callbacks would not be compatible with swapping/flipping out of the box, but it's no big deal since 1) it wouldn't count as a regression 2) callbacks have very limited support for spherical geometry anyway

chrishavlin:
awesome! thanks for testing this more thoroughly! I haven't gotten a chance to do some final checks on this... hopefully soon!

neutrinoceros:
Awesome ! I went to check what image tests were failing and it looks like it's all "toggle_right_handed + callbacks" tests, and the new results are actually fixes, not regressions.

For instance, see this image produced on the main branch:


reference

and with your branch:
new

I'll take more time later today to check that *all* differences fall into that category, and if it's indeed the case I'll approve for bumping images !

neutrinoceros:
I checked everything. All changes fall into the same category as the example above, and most of the failures are just "no old answer", so please feel free to bump existing answers and create the missing ones :)

chrishavlin:
oh, there are a couple lingering comments related to the docs that I never addressed, will do that now...

doc/source/visualizing/plots.rst
chrishavlin: Working on a quick docs change --- I agree, should fix this now! But what's the correct description?

I'm inclined to just delete this sentence and  start with "There are a number of ways the orientation of the initial `PlotWindow` view can be adjusted." and not get into how the default is constructed at all, but maybe that's actually import context to keep?

doc/source/visualizing/plots.rst
chrishavlin: Will delete this line in the next commit!

doc/source/visualizing/plots.rst
neutrinoceros: Hum... I think it's important that we have clearly documented default behaviours, so I am not in favour of deleting the existing sentence, but I would like that it also mentions that the "normal" direction points to the user _by default_. Since it's one of the things that can be adjusted, your existing transition to the next paragraph still works with my proposed addition :)

neutrinoceros (commented):
Last minute thoughts:
I think maybe we shouldn't expose boolean kwargs to invoke axis flipping and swapping at plot initialisation, because there's no way to capture an "order of operation" in this interface so it may turn out to be blocking for some of the solutions proposed in #3890.

Another question that I'm almost sure we answered already (sorry I can't find it back): by deprecating "toggle_right_handed", we're making room for this new API, but are we not also taking away a piece of functionality ? I think I may have initially misunderstood what this method was supposed to do and I'm not convinced any more that it can be replaced by this new API. In short: aren't we just building an API in which plots can only be right-handed ?
(I may answer my own question later, I need to rethink about it)

Edit: indeed, I came (again ?) to the conclusion that right-handedness isn't a relevant concept and that we're not actually taking any functionality away from users with this deprecation.

doc/source/visualizing/plots.rst
neutrinoceros: I'll take that back. After rethinking about it I think the whole notion of "right handed" 2D plots doesn't make sense; you'd get the exact same result by assuming either right-handedness with the normal pointing towards the reader or left-handedness + normal pointing away.
Seems to me we're better off leaving "right-handedness" completely out of documention. Sorry about these back and forth, I get really confused by this at times

chrishavlin:
I'm not sure about the boolean kwargs question. It does seem simpler to not have the axis manipulations as kwargs so that they require explicit callbacks -- it would allow us to easily add arguments down the line. But `right_handed` is a current kwarg that this PR is depreciating, so I'm inclined to keep the kwargs to have a more one-to-one replacement.

doc/source/visualizing/plots.rst
chrishavlin: > Sorry about these back and forth, I get really confused by this at times

Ya, me too :) 

So to capture the default behavior, maybe: 

"By default, all Plot Window objects plot with the normal of the plotting plane pointing to the user. There are a number of ways to adjust the initial orientation ... "

neutrinoceros:
Thing is, it's much easier to add the arguments back later than it is to remove them after they go in. I don't think the `right_handed` kwarg is used often enough that a one-to-one replacement is necessary. But that's just a wild guess.

chrishavlin:
Ok! I'll take out the kwargs and adjust the `right_handed` depreciation message to suggest switching to using the `flip_horizontal` callback (instead of the boolean kwarg).

neutrinoceros:
I think we could merge this sometimes this week provided that you make the last requested changes and we give a couple days windows for other devs to object if need be. After that, we should make a small survey to collect user opinions on https://github.com/yt-project/yt/issues/3890

chrishavlin:
Great! And I'm about to push up some changes that remove the kwargs for the flips and swap. In doing so I realized that I also removed any testing of the new features so I had to write a new answer test just for the axis manipulations. I also went through and added some axis manipulations to most of the other callback tests. Ideally those would all be answer tests as well, but that'd require a big rewrite of that test file. I figure at least having those checks that they run is better than nothing.

Ok, here, we go!

chrishavlin:
(note, expect this to fail since the new test won't exist yet but I'll let everything run in case there are other failures).

neutrinoceros:
We'll want to add answer tests for exact behavior after we've settled on what it should be, so not a big deal that you're not doing it now. I agree that simpler tests may suffice for now.

chrishavlin:
@yt-fido test this please

tests/tests.yaml
```diff
@@ -95,12 +95,13 @@ answer_tests:
     - yt/frontends/owls/tests/test_outputs.py:test_snapshot_033
     - yt/frontends/owls/tests/test_outputs.py:test_OWLS_particlefilter
 
-  local_pw_042:  # PR 2504
+  local_pw_043:  # PR 3640
```
neutrinoceros, line 98: @chrishavlin I think you forgot to bump the answer number again in your most recent commit

chrishavlin: oh! yes I did. thanks for catching!

neutrinoceros (approved):
I'm going to advertise this PR and the follow up issue via the mailing list so hopefully we can get it in soon !

yt/visualization/plot_window.py
```diff
@@ -2128,6 +2201,15 @@ class AxisAlignedProjectionPlot(ProjectionPlot, PWViewerMPL):
          Size of the buffer to use for the image, i.e. the number of resolution elements
          used.  Effectively sets a resolution limit to the image if buff_size is
          smaller than the finest gridding.
+    swap_axes: boolean
```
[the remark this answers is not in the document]
chrishavlin: sure, can clean up now. oh and looks like I also didn't add back the `right_handed` kwarg description in some of the docstrings. I can do that too (with a depreciation note in the docstring).

neutrinoceros (approved):

neutrinoceros:
@chrishavlin seems like we got some conflicts, most likely since #3756 was merged.

Could you resolve them ?

I'm set to merge this PR in a few days since no one seems to object yet.

neutrinoceros (approved):

PR #3657, opened 2021-11-08
BLD: drop wheel building for Python 3.6 and start building for 3.10

neutrinoceros:
## PR Summary

I believe everything upstream is now ready, so let's start building wheels for Python's latest version.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -7,6 +7,7 @@ on:
       - stable
     tags:
       - 'yt-*'
+  workflow_dispatch:
 
 jobs:
   build_wheels:
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -37,7 +37,7 @@ jobs:
 
       - uses: actions/checkout@v2
       - name: Install cibuildwheel
-        run: python -m pip install cibuildwheel==1.9.0
+        run: python -m pip install cibuildwheel==2.2.2
 
       - name: Install dependencies and yt
         shell: bash
@@ -50,7 +50,8 @@ jobs:
         run: |
           python -m cibuildwheel --output-dir dist
         env:
-          CIBW_BUILD: "cp36-* cp37-* cp38-* cp39-*"
+          CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-*"
+          CIBW_SKIP: "*-musllinux_*"  # these fail due to side effects from previous builds
           CIBW_ARCHS_LINUX: "x86_64"
           CIBW_ARCHS_MACOS: "x86_64"
           CIBW_ARCHS_WINDOWS: "auto"


neutrinoceros:
now testing this on my fork, here: https://github.com/neutrinoceros/yt/actions/runs/1434770363

neutrinoceros:
First experiment failed because h5py is not yet ready for Python 3.10
However, I don't think we actually need it to build our own wheels, so I'm now experimenting on minimising build time dependencies. See https://github.com/neutrinoceros/yt/actions/workflows/wheels.yaml

.github/workflows/wheels.yaml
```diff
@@ -25,32 +26,24 @@ jobs:
...
-          python-version: 3.9
-
-      - uses: s-weigand/setup-conda@v1
-        if: matrix.os == 'windows-latest'
-        with:
-          update-conda: true
-          conda-channels: conda-forge
-          activate-conda: true
-          python-version: 3.9
+          python-version: '3.10'
```
Xarthisius, line 29: Why do you change the environment that's used to run `cibuildwheel`? That's completely unnecessary.

.github/workflows/wheels.yaml
```diff
@@ -25,32 +26,24 @@ jobs:
...
+          dependencies: "bare"
           LDFLAGS: "-static-libstdc++"
         run: source ./tests/ci_install.sh
 
       - name: Build wheels for CPython
         run: |
           python -m cibuildwheel --output-dir dist
         env:
-          CIBW_BUILD: "cp36-* cp37-* cp38-* cp39-*"
+          CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-*"
```
Xarthisius, line 46: This should be the only change necessary to add 3.10 support for wheels.

.github/workflows/wheels.yaml
neutrinoceros: thanks for clarifying, I wasn't sure what was needed here since it's _currently_ using the latest supported Python version

.github/workflows/wheels.yaml
neutrinoceros: That's what I was going for originally. However I found that
1) updating cibuildwheel to 2.1.3 was needed to use the latest actual Python 3.10.0 release (instead of a beta)
2) we can't update all the way to 2.2.2 for now because it will build wheels for Musllinux and apparently this is currently broken ? see my failed attempt here https://github.com/neutrinoceros/yt/runs/4139228854?check_suite_focus=true
3) we cannot yet install the "full" yt build (including optional dependencies) with Python 3.10 because of h5py. This is why I changed the dependencies target.

Xarthisius: 1. OK, that's fine, but it's should still work with 3.9 as main env
2. `CIBW_SKIP: "*-musllinux-*"` ?
3. The "full" is installed inside the env that runs cibuildwheel, not within the images that build wheels.

neutrinoceros: thanks a lot, I'll try that

neutrinoceros:
testing with @Xarthisius' suggested changes here : https://github.com/neutrinoceros/yt/actions/runs/1436028691

Xarthisius (approved):

PR #3663, opened 2021-11-11
ENH: Adding Cholla Frontend

chummels:
This adds a frontend to yt for working with datasets from the CHOLLA code (https://github.com/cholla-hydro/cholla). It replaces the previous attempt at this from a few years ago (PR #1944 )

CHOLLA is a super straightforward code format: it's a fixed-resolution grid code saved to HDF5 files with attributes storing simulation parameters.  For simplicity, I've also made it so nprocs=1, so it reads a single grid for the whole dataset, since it's unigrid. After I get this working, I'll make yt chunk it into smaller-sized chunks for for efficiency.  But for now, I just want *something* that works.

The frontend works so far for basic things like reporting fields and parameters and such.  But when one attempts to actually make a slice or something, it fails with a `Arrays are not of correct shape` error.  The main remaining hurdles are getting the `_read_fluid_selection()` and `_read_chunk_data()` functions working in `io.py`.

Any assistance in finishing up this frontend would be greatly appreciated.  The CHOLLA code author, @evaneschneider, and her students are interested in contributing to active development in yt when they can actually get their code to be read in successfully.

For testing purposes, a 1-zone CHOLLA test dataset can be found here: http://chummels.org/cholla/1.h5 .


neutrinoceros:
There are a couple undefined names that caught by flake8 via pre-commit, I guess that's the most urgent thing to fix to get this going

neutrinoceros:
Also, this isn't widely known yet (because we do not advertise it very much), but you could consider building this frontend as a yt extension so that you could iterate much more rapidly with it as it would be outside the main code base.
I don't know if you'd want to do it but just in case here is some insight:

Here are some example: 
- https://github.com/brittonsmith/yt_legacy (two frontends for the price of one!)
- https://github.com/ruithnadsteud/yt_georaster
- https://github.com/Xarthisius/yt_gdf/ (note: this is a fake extension that Kacper created by extracting the yt-native GDF frontend)
- https://github.com/neutrinoceros/yt_idefix

The process to transform what you have into an extension is actually very straightforward. I'll publish my notes in a GitHub gist and post it here soon

chummels:
Thanks for the offer, but I'm not sure that I need this to be an extension.  Once the frontend actually *works* and gets merged, I don't anticipate modifying it much.  It's just a matter of getting it working first, which has been the hangup for literally years.

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,141 @@
...
+        try:
+            fileh = h5py.File(filename, mode="r")
+            attrs = fileh.attrs
+            if "bounds" in attrs and "domain" in attrs:
+                fileh.close()
+                return True
+            fileh.close()
+        except Exception:
+            pass
+        return False
```
neutrinoceros, line 141:
```suggestion
        try:
            fileh = h5py.File(filename, mode="r")
        except OSError:
            return False
        
        try:
            attrs = fileh.attrs
        except AttributeError:
            return False
        else:
            return "bounds" in attrs and "domain" in attrs
        finally:
            fileh.close()
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,141 @@
...
+
+        h5f = h5py.File(self.parameter_filename, mode="r")
+        attrs = h5f.attrs
+        self.parameters = {k: v for (k, v) in attrs.items()}
+        self.domain_left_edge = attrs["bounds"][:].astype("=f8")
+        self.domain_right_edge = attrs["domain"][:].astype("=f8")
+        self.dimensionality = len(attrs["dims"][:])
+        self.domain_dimensions = attrs["dims"][:].astype("=f8")
+        self.current_time = attrs["t"][:]
+        self.periodicity = [False, False, False]
```
neutrinoceros, line 110:
```suggestion
        self._periodicity = (False, False, False)
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class ChollaGrid(AMRGridPatch):
+    _id_offset = 0
+
+    def __init__(self, id, index, level):
+        super().__init__(id, filename=index.index_filename, index=index)
+        self.Parent = None
+        self.Children = []
+        self.Level = level
```
neutrinoceros, line 22:
```suggestion
    def __init__(self, id, index, level, dims):
        super().__init__(id, filename=index.index_filename, index=index)
        self.Parent = None
        self.Children = []
        self.Level = level
        self.ActiveDimensions = dims
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,141 @@
...
+        super().__init__(ds, dataset_type)
+
+    def _detect_output_fields(self):
+        h5f = h5py.File(self.index_filename, mode="r")
+        self.field_list = [("cholla", k) for k in h5f.keys()]
+        h5f.close()
+
+    def _count_grids(self):
+        # This needs to set self.num_grids (int)
+        # Probably need to change this.
```
neutrinoceros, line 48:
```suggestion
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,141 @@
...
+        self.grid_right_edge[0][:] = self.ds.domain_right_edge[:]
+        self.grid_dimensions[0][:] = self.ds.domain_dimensions[:]
+        self.grid_particle_count[0][0] = 0
+        self.grid_levels[0][0] = 1
+        self.max_level = 1
+
+    def _populate_grid_objects(self):
+        self.grids = np.empty(self.num_grids, dtype="object")
+        for i in range(self.num_grids):
+            g = self.grid(i, self, self.grid_levels.flat[i])
```
neutrinoceros, line 62:
```suggestion
            g = self.grid(i, self, self.grid_levels.flat[I], self.grid_dimensions[I])
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,141 @@
...
+            g._prepare_grid()
+            g._setup_dx()
+            self.grids[i] = g
+
+
+class ChollaDataset(Dataset):
+    _index_class = ChollaHierarchy
+    _field_info_class = ChollaFieldInfo
+
+    def __init__(
```
neutrinoceros, line 72: please check that this method's signature is compatible with existing sibling classes.

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,141 @@
...
+        # values.
+        #
+        self.length_unit = self.quan(1.0, "pc")
+        self.mass_unit = self.quan(1.0, "Msun")
+        self.time_unit = self.quan(1000, "yr")
+        self.velocity_unit = self.quan(1.0, "cm/s")
+        self.magnetic_unit = self.quan(1.0, "gauss")
+
+        # this minimalistic implementation fills the requirements for
+        # this frontend to run, change it to make it run _correctly_ !
```
neutrinoceros, line 96: I'm guessing this is a left over from code duplication from the skeleton frontend ?
```suggestion
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,141 @@
...
+        self.current_time = attrs["t"][:]
+        self.periodicity = [False, False, False]
+        h5f.close()
+
+        # CHOLLA cannot yet be run as a cosmological simulation
+        self.cosmological_simulation = 0
+        self.current_redshift = 0
+        self.omega_lambda = 0
+        self.omega_matter = 0
+        self.hubble_constant = 0
```
neutrinoceros, line 118: Purely for the sake of typing consistency
```suggestion
        self.cosmological_simulation = 0
        self.current_redshift = 0.0
        self.omega_lambda = 0.0
        self.omega_matter = 0.0
        self.hubble_constant = 0.0
```

yt/frontends/cholla/fields.py
```diff
@@ -0,0 +1,134 @@
...
+        ("momentum_x", (mom_units, ["momentum_x"], None)),
+        ("momentum_y", (mom_units, ["momentum_y"], None)),
+        ("momentum_z", (mom_units, ["momentum_z"], None)),
+        ("Energy", ("code_pressure", ["total_energy_density"], None)),
+    )
+
+    known_particle_fields = (
+        # Identical form to above
+        # ( "name", ("units", ["fields", "to", "alias"], # "display_name")),
+    )
```
neutrinoceros, line 35:
```suggestion
    known_particle_fields = ()
```

yt/frontends/cholla/fields.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __init__(self, ds, field_list):
+        super().__init__(ds, field_list)
+        # If you want, you can check self.field_list
```
neutrinoceros, line 39: Unless you mean to add more instructions here, this method does nothing
```suggestion
```

yt/frontends/cholla/fields.py
```diff
@@ -0,0 +1,134 @@
...
+               units=unit_system["pressure"],
+           )
+
+           def _pressure(field, data):
+               return (
+                  (data.ds.gamma - 1.0)
+                  * data["cholla", "GasEnergy"]
+               )
+        else:
+          def _pressure(field, data):
```
neutrinoceros, line 88: inconsistent indentation here. This should be auto-fixed with pre-commit

yt/frontends/cholla/fields.py
```diff
@@ -0,0 +1,134 @@
...
+            ("gas", "temperature"),
+            sampling_type="cell",
+            function=_temperature,
+            units=unit_system["temperature"],
+        )
+        
+
+    def setup_particle_fields(self, ptype):
+        super().setup_particle_fields(ptype)
+        # This will get called for every particle type.
```
neutrinoceros, line 134: Same as above. No need to override a method if we don't do anything more than the parent class
```suggestion
```

neutrinoceros (commented):
I wanted to have a look at the code to see if I could spot problems and I think I found one in the `ChollaGrid` class. The rest of my suggestions are minor.
I'd like to help you more here. Could you give a minimal script that shows what's currently not working so I can use it as a starting point ?

matthewturk:
@chummels any chance you have a dataset that's bigger than 1 zone?  One of the things you mentioned in a previous PR was that you wanted to be able to do grid subsetting.  If we had a 32^3 dataset we'd be able to experiment with that.  (Even 16^3, really.)

matthewturk:
Datasets can be uploaded with `curl -T datafile http://use.yt/upload/` and reporting the returned URL.

yt/frontends/cholla/data_structures.py
chummels: I don't understand why you're introducing `I` when it is not referenced anywhere in the previous code.  Or perhaps this is just a typo?

matthewturk: looks like a typo

yt/frontends/cholla/data_structures.py
chummels: Yes, this frontend is still a work in progress.  I was leaving a bunch of the comments in from the original skeleton template until I'd made sure things were working appropriately.

chummels:
@matthewturk and @neutrinoceros Thanks for the suggestions.

I've uploaded a 32**3 CHOLLA dataset here: http://use.yt/upload/42ce9392

A sample script which demonstrates the failure mode for this frontend is:

```
import yt
ds = yt.load('cholla_32.h5')
yt.SlicePlot(ds, 'x', ('gas', 'density')).save()
```

neutrinoceros:
For the record I'm waiting for the undefined named caught by flake8 to be fixed before I can review this.

yt/frontends/cholla/data_structures.py
chummels: I'll worry about removing comments after I've got things working.  For now, they're useful to have around.

chummels:
This missing argument should now be resolved.

yt/frontends/cholla/io.py
```diff
@@ -0,0 +1,55 @@
+import h5py
+
+from yt.utilities.io_handler import BaseIOHandler
```
neutrinoceros, line 3: This is breaking some tests
```suggestion
from yt.utilities.io_handler import BaseIOHandler
from yt.utilities.on_demand_imports import _h5py as h5py
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,145 @@
...
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+        #
+        # The functionality in this method should be unique enough that it can
+        # differentiate the frontend from others. Sometimes this means looking
+        # for specific fields or attributes in the dataset in addition to
+        # looking at the file name or extension.
+        try:
+            fileh = h5py.File(filename, mode="r")
+        except OSError:
```
neutrinoceros, line 135: Not catching ImportError breaks any install where h5py isn't installed
```suggestion
        except (ImportError, OSError):
```

neutrinoceros (commented):
Here's a couple more suggestions to fix CI
Note that we currently have other failures happening on every branch following the release of matplotlib 3.5, which will be resolved in #3670

Note that pre-commit is still unhappy because of some trailing white space. You can fix it with

```
pre-commit run trailing-whitespace --all-files
```
and in general use this command to run all hooks locally
```
pre-commit run --all-files
```
This is sometimes necessary if you adopt pre-commit _while_ you're developing a branch because hooks are only run on staged files when you commit, but pre-commit.ci checks that *every* file is stable.

Equivalently to using the second command locally, you can comment here to invoke pre-commit.ci, which will commit to your branch. See https://results.pre-commit.ci/run/github/88776493/1637212772.pE5ouDciRFi4WdwUgdp6ZA for report and instructions (bottom of the page)

chummels:
Thanks for the suggestions, which I've now included. At this point, the main thing I'm trying to address is getting the frontend to actually work.  I'm happy to clean up the code, fix the CI, remove the comments, document things, write tests, all that, after I've got the code actually working.

neutrinoceros:
I get your frustration but I don't consider that fixing bugs that can be caught by static checking (e.g. flake8) counts as "cleaning". Having `NameErrors` lurking in the code is a good reason for something to not work, which is why I insisted on fixing this first. Thank you for addressing this. I'll take a look this weekend and see if I can spot any bigger bugs.

chummels:
That's fair.  I appreciate the help!  I wasn't trying to be dismissive, just that I think my implementation of `_read_fluid_selection()` is all wrong but I'm not really sure how to fix it.

neutrinoceros:
I'm going to sign off for today, if you want to upload more data that I can use for testing and debugging I figured I may have as well ask now before it's weekend in California.

yt/frontends/cholla/io.py
```diff
@@ -0,0 +1,54 @@
...
+        # This method is not abstract, and has a default implementation
+        # in the base class.However, the default implementation requires that the method
+        # io_iter be defined
+        data = {}
+        h5f = h5py.File(self.ds.parameter_filename, "r")
+        for field in fields:
+            ftype, fname = field
+            data[field] = h5f[fname][:].flatten().astype("=f8")
+        h5f.close()
+        return data
```
neutrinoceros, line 47: I am 99% sure this is the implementation you need. Now I've tested it locally and it's _NOT_ enough to fix the frontend, the reason being that the chunk's `size` that get passed here is 1024 instead of 32768 (32^3, in case of a unigrid with size (32*32*32) as the example dataset you provided). At the moment I don't understand where this 1024 is being generated, and it may be a bit convoluted to track the bug back to the frontend.
```suggestion
        def _read_fluid_selection(self, chunks, selector, fields, size):
        data = {}
        offset = 0

        with h5py.File(self.ds.parameter_filename, "r") as fh:
            for field in fields:
                ftype, fname = field
                data[ftype, fname] = np.empty(size, dtype="float64")
                for chunk in chunks:
                    for grid in chunk.objs:
                        values = fh[fname][:].flatten().astype("=f8")
                        offset += grid.select(selector, values, data[field], offset)
        return data
```

neutrinoceros (commented):
I got a patch that I think fixes the `_read_fluid_selection` method (essentially plagiarized from [my own mono grid frontend](https://github.com/neutrinoceros/yt_idefix/blob/0d614e224c7b91cc4c4e2e3630a1b7b8c8b2cd47/yt_idefix/io.py#L20)), but it's not sufficient to make the frontend functional. However I think that with this, you're only one bug away from making it work.

yt/frontends/cholla/io.py
neutrinoceros, line 47: This implementation worked for me
```suggestion
    def _read_fluid_selection(self, chunks, selector, fields, size):
        data = {}
        offset = 0

        with h5py.File(self.ds.parameter_filename, "r") as fh:
            for field in fields:
                ftype, fname = field
                data[ftype, fname] = np.empty(size, dtype="float64")
                for chunk in chunks:
                    for grid in chunk.objs:
                        values = fh[fname][:].astype("=f8")
                        offset += grid.select(selector, values, data[field], offset)
        return data
```

neutrinoceros (commented):
This time I think I got it. With this script I get

```python
import yt

ds = yt.load("cholla.h5")
for normal in "xyz":
    yt.SlicePlot(ds, normal, ('gas', 'density')).save(f"/tmp/{normal}.png")
```
x
y
z

(I'm a bit sad that there are bugs in the x-normal plot but the good news is that I'm sure these are not from the frontend)

chummels:
Awesome!  Thanks for your help, @neutrinoceros !  I'll test this out locally and give it the stress test.  THANKS!

chummels:
This passes my basic tests, so thank you for your contribution!  @evaneschneider and I are going to use this on a few other datasets and for other functionality to really ensure it's ready to go.  Then I'll clean up the code, write some docs and tests, and then take WIP off of this.  Really looking forward to having this in the codebase soon after so long!

neutrinoceros:
Awesome. Don't hesitate to showcase your successes and failures here as you go !

neutrinoceros:
I suspect that io could be a lot more efficient on large datasets but I haven't tried to optimize it yet on my own unigrid frontend, please report any idea or problems you encounter with that. The _read_fluid_selection method is crucial for performance.

matthewturk (commented):
I think this is definitely on the right track -- I've left a handful of comments that might help, or might just be ignorable.

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,146 @@
...
+        h5f = h5py.File(self.parameter_filename, mode="r")
+        attrs = h5f.attrs
+        self.parameters = {k: v for (k, v) in attrs.items()}
+        self.domain_left_edge = attrs["bounds"][:].astype("=f8")
+        self.domain_right_edge = attrs["domain"][:].astype("=f8")
+        self.dimensionality = len(attrs["dims"][:])
+        self.domain_dimensions = attrs["dims"][:].astype("=f8")
+        self.current_time = attrs["t"][:]
+        self._periodicity = (False, False, False)
+        self.refine_by = 1
```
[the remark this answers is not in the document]
neutrinoceros: this attribute is also required for now so I think this value is appropriate for a unigrid datatype (https://github.com/yt-project/yt/issues/1880)

neutrinoceros:
So the `read_fluid_selection` implementation that I contributed here was directly adapted from the Idefix/Pluto frontend I maintain externally. 
@xshaokun discovered a pretty nasty bug there _and_ found a more robust implementation (many thanks to him !).
@chummels you'll probably need to do something similar here, see the patch
https://github.com/neutrinoceros/yt_idefix/pull/68/files

chummels:
Thanks, @neutrinoceros !  That might explain the behavior we've found that some of the fields seem to be set to the wrong value at initialization, which I'm currently digging into a bit more.  I'll look into this.

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,146 @@
...
+
+        h5f = h5py.File(self.parameter_filename, mode="r")
+        attrs = h5f.attrs
+        self.parameters = {k: v for (k, v) in attrs.items()}
+        self.domain_left_edge = attrs["bounds"][:].astype("=f8")
+        self.domain_right_edge = attrs["domain"][:].astype("=f8")
+        self.dimensionality = len(attrs["dims"][:])
+        self.domain_dimensions = attrs["dims"][:].astype("=f8")
+        self.current_time = attrs["t"][:]
+        self._periodicity = (False, False, False)
```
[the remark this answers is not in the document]
chummels: No, but there is no information in the HDF5 header that allows you to determine periodicity.  @evaneschneider thinks the default here should be to assume non-periodic boundary conditions and then if they *are* periodic, then after loading them, perhaps we can just set this to be true?  Is this just a matter of:

```
import yt
ds = yt.load(fn)
ds._periodicity = (True, True, True)
```

yt/frontends/cholla/fields.py
```diff
@@ -0,0 +1,131 @@
...
+        def kinetic_energy_density(data):
+            return (
+                0.5
+                * data["cholla", "density"]
+                * (
+                    data["gas", "velocity_x"] * data["gas", "velocity_x"]
+                    + data["gas", "velocity_y"] * data["gas", "velocity_y"]
+                    + data["gas", "velocity_z"] * data["gas", "velocity_z"]
+                )
+            )
```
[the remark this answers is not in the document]
chummels: Good call.  Thank you!

chummels:
Thank you, @neutrinoceros !  That was indeed causing some problems that were perplexing us in terms of fields being set incorrectly.

neutrinoceros:
refreshing CI

yt/frontends/cholla/fields.py
```diff
@@ -0,0 +1,155 @@
...
+
+def velocity_field(comp):
+    def _velocity(field, data):
+        return data["cholla", f"momentum_{comp}"] / data["cholla", "density"]
+
+    return _velocity
+
+
+class ChollaFieldInfo(FieldInfoContainer):
+    known_other_fields = (
```
neutrinoceros, line 22: This line is reported by mypy
```
yt/frontends/cholla/fields.py:22: error: Need type annotation for "known_other_fields"  [var-annotated]
```

The fix is this:
```suggestion
    known_other_fields: KnowFieldsT = (
```

(in combination with adding `from yt._typing import KnownFieldsT` at the top level of the module

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __init__(self, id, index, level, dims):
+        super().__init__(id, filename=index.index_filename, index=index)
+        self.Parent = None
+        self.Children = []
+        self.Level = level
+        self.ActiveDimensions = dims
+
+    def __repr__(self):
+        return "ChollaGrid_%04i (%s)" % (self.id, self.ActiveDimensions)
```
neutrinoceros, line 26: this can be dropped entirely now (see #3936)
```suggestion
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,165 @@
...
+        super().__init__(filename, dataset_type, units_override=units_override)
+        self.storage_filename = storage_filename
+
+    def _set_code_unit_attributes(self):
+        # This is where quantities are created that represent the various
+        # on-disk units.  These are the defaults, but if they are listed
+        # in the HDF5 attributes for a file, which is loaded first, then those are
+        # used instead.
+        #
+        if not self.length_unit:
```
neutrinoceros, line 88: is this the intention here ? (if so, please apply to other similar conditionals too)
```suggestion
        if self.length_unit is None:
```

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,165 @@
...
+        if not self.time_unit:
+            self.time_unit = self.quan(1000, "yr")
+        if not self.velocity_unit:
+            self.velocity_unit = self.quan(1.0, "cm/s")
+        if not self.magnetic_unit:
+            self.magnetic_unit = self.quan(1.0, "gauss")
+
+        for key, unit in self.__class__.default_units.items():
+            setdefaultattr(self, key, self.quan(1, unit))
+
```
neutrinoceros, line 101: Looks like all of this could be accomplished by just defining this class' `default_units`

yt/frontends/cholla/data_structures.py
```diff
@@ -0,0 +1,165 @@
...
+                self.time_unit = self.quan(time_unit[0], "s")
+            if velocity_unit:
+                self.velocity_unit = self.quan(velocity_unit[0], "cm/s")
+            if magnetic_unit:
+                self.magnetic_unit = self.quan(magnetic_unit[0], "gauss")
+
+            # this minimalistic implementation fills the requirements for
+            # this frontend to run, change it to make it run _correctly_ !
+            for key, unit in self.__class__.default_units.items():
+                setdefaultattr(self, key, self.quan(1, unit))
```
neutrinoceros, line 137: this is redundant with `_set_code_unit_attributes`

yt/frontends/cholla/fields.py
```diff
@@ -0,0 +1,155 @@
...
+        ("momentum_z", (mom_units, ["momentum_z"], None)),
+        ("Energy", ("code_pressure", ["total_energy_density"], None)),
+        ("scalar0", (rho_units, [], None)),
+        ("metal_density", (rho_units, ["metal_density"], None)),
+    )
+
+    known_particle_fields = ()
+
+    def __init__(self, ds, field_list):
+        super().__init__(ds, field_list)
```
neutrinoceros, line 37: I don't see the point of adding a frame to the call stack here
```suggestion
```

yt/frontends/cholla/fields.py
neutrinoceros: @chummels looks like you forgot the import

yt/frontends/cholla/data_structures.py
chummels: If we didn't allow the user to change the units, I see your that solution works, but I'm not sure that your suggestion will work while still checking each unit in the dataset attributes to see if the user overrides them.

neutrinoceros: The `default_units` mechanism was primarily implemented for the AMRVAC frontend, where we very much allow units override. The key difference here is that AMRVAC data files never contain units, so I need to check if we need to generalize the mechanism to support this case. Maybe that's something better left done as a follow up refactor. I will think about it some more.

yt/frontends/cholla/fields.py
chummels: I found it in the skeleton, and it seems to make sense?  Aren't we just inheriting any sort of field definitions from the superclass?

neutrinoceros: yes we are, but here we override the base `__init__` method just to call it again one stack level deeper, so it accomplishes exactly nothing more than if we just inherited the base method.
I don't actually see a good reason for having this in the skeleton frontend.

yt/sample_data_registry.json
```diff
@@ -47,6 +47,12 @@
     "load_name": "grid1.nc",
     "url": "https://yt-project.org/data/CfRadialGrid.tar.gz"
   },
+  "ChollaSimple.tar.gz": {
+    "hash": "b9b73eda41a7a4e7d2e5b8e263f4230334d316f003da616648002f913c655c43",
```
neutrinoceros, line 51: tip: you can test if the hash is correct by trying to run
```python
import yt
yt.load_sample("ChollaSimple")
```
when the hash isn't correct, pooch raises an error showing both the hash read from this file and the actual one.

I was able to generate the same hash pooch reported with 
```
openssl sha256 ChollaSimple.tar.gz
```
```suggestion
    "hash": "aa277471f694f9de2e46c190efd8cd631c0694044aa29cb8431f127319774d73",
```

yt/frontends/cholla/tests/test_outputs.py
```diff
@@ -0,0 +1,64 @@
...
+from yt.frontends.cholla.api import ChollaDataset
+from yt.testing import assert_equal, requires_file
+from yt.utilities.answer_testing.framework import data_dir_load
+
+_fields = (
+    ("gas", "temperature"),
+    ("gas", "density"),
+)
+
+cholla_0 = "cholla_0.h5"
```
neutrinoceros, line 11: Tests are failing because the file isn't found on server. I think this is the fix
```suggestion
cholla_0 = "ChollaSimple/0.h5"
```

chummels:
I added some basic tests which just ensure that the test file is recognized as a CHOLLA dataset, that it can load it, and that it has the fields that are expected and can access them.  Are there any other canonical tests that are used to test out frontends or is this enough for now?

yt/frontends/cholla/tests/test_outputs.py
```diff
@@ -0,0 +1,64 @@
...
+    ]
+
+    ds = yt.load(ChollaSimple)
+    assert_equal(str(ds), "0.h5")
+    ad = ds.all_data()
+
+    # Check all the expected fields exist and can be accessed
+    for field in expected_fields:
+        assert ("cholla", field) in ds.field_list
+        # assert ("gas", field) in ds.derived_field_list
```
neutrinoceros, line 39:
```suggestion
```

yt/frontends/cholla/tests/test_outputs.py
```diff
@@ -0,0 +1,64 @@
...
+
+ChollaSimple = "ChollaSimple/0.h5"
+
+
+@requires_file(ChollaSimple)
+def test_ChollaDataset():
+    assert isinstance(data_dir_load(ChollaSimple), ChollaDataset)
+
+
+@requires_file(ChollaSimple)
```
neutrinoceros, line 19: `@requires_file` mark answer tests, but these are simple tests that don't really fit in answer testing.

It would actually be the kind of test we traditionally craft for frontends. See `yt/frontends/amrvac/tests/test_outputs.py` for a collection of simple examples.

Also note that you can't use `yt.load` in an answer test, instead you'll need `yt.utilities.answer_testing.framework.data_dir_load`

chummels: Thanks for the tips.  I tried using `@requires_ds` instead of `@requires_file` but the testing framework just skipped tests prepended with that, since it wasn't recognizing my dataset as a valid dataset.

neutrinoceros: oh. So sorry I had it the other way around, indeed `requires_ds` is the one that means "answer test", as the very examples I was pointing to in the AMRVAC frontend illustrate 🤦🏻‍♂️ 
Running those locally isn't straightforward though: https://yt-project.org/docs/dev/developing/testing.html#how-to-run-the-answer-tests

matthewturk:
Looks like this is nearly done -- are we waiting for answer tests?  Is that something someone else could help out with?

matthewturk:
pre-commit.ci autofix

neutrinoceros:
If we're not able to finish this in the next couple weeks I'd suggest leaving it to yt 4.2 which - the way I envision it - should happen much closer in time than 4.1 is to 4.0

matthewturk:
@neutrinoceros I think we need to hold 4.1 until this is in, and I will work on this concurrently with finishing up #3421 .  @chummels can you enumerate what is required still, and I can take it on?

jzuhone:
I agree, this should go in 4.1

chummels:
@matthewturk I think all that remains to be done for this PR is the addition of an answer test or two using the dataset that is already being used in the tests.  It's a super simple dataset with a density gradient across the grids, so I guess just make a ProjectionPlot and SlicePlot and PhasePlot answer test and then call it a day?  I am not sure, but that seems like it should be sufficient?

matthewturk:
I've gone ahead and added some simple tests.  These work locally, but it's hard for me know if I did everything properly, so I guess we'll see what fido says.

neutrinoceros:
close/open to refresh CI now that unrelated failures are resolved on main

neutrinoceros:
FTR I just merged from main, but I really meant to do this for https://github.com/yt-project/yt/pull/3421. This should not cause any trouble other than wasting CI time. Sorry folks

yt/frontends/cholla/data_structures.py
neutrinoceros: Btw this is non blocking

neutrinoceros (approved):
I don't have any blocking comments left. I believe Matt should be able to merge this even as a co-author.

Issue #3513, opened 2021-09-14
BUG: piling up memory bloat while plotting in a loop


neutrinoceros:
### Bug report

**Bug summary**

Generating a simple plot in a loop creates significant memory waste

**Code for reproduction**
Using [`memory_profiler`](https://pypi.org/project/memory-profiler/) as
```shell
wget https://yt-project.org/data/solar_prom2d.tar.gz
tar -xvzf solar_prom2d.tar.gz
mprof run benchmark.py
mprof plot
```
with
```python
# benchmark.py
import sys
import time

import yt

def main(n) -> int:
    yt.set_log_level("error")
    for _ in range(n):
        ds = yt.load("solar_prom2d/output0001.dat")
        pp = yt.ProjectionPlot(ds, "z", ("gas", "density"))
        del ds, pp
    return 0

if __name__ == "__main__":
    sys.exit(main(100))
```

**Actual outcome**

![mprof_res](https://user-images.githubusercontent.com/14075922/133330260-5e52ef72-2854-4474-b4ab-3a2afa08ca49.png)

The job is ending up eating 4GiB (!) or RAM just because it ran 100 times the same 2D render of a 600MiB dataset.

notes
1) I see no difference with our without the `del` statement. Seems like the garbage collector is doing a decent job, or at least I can't really top it with del
2) I tried patching the plot classes to force closure/deletion of embedded  matplotlib figures, to no avail.

I don't currently have any idea what might be causing this bloat and how to address it.

**Expected outcome**

Memory usage shouldn't drift.

**Version Information**
  * Operating System: MacOS
  * Python Version: 3.9.5
  * yt version: main branh (Version = 4.1.dev0, Changeset = 4d81bc35b08f)
  * Other Libraries (if applicable): `memory_profiler` 0.58.0

cphyc:
On top of my head, did you try replacing the line with the projection plot with, say `ad = ds.all_data(); ad["density"]`?
I am wondering if the issue isn't about loading the dataset and keeping some reference of it somewhere, thus preventing the memory from being freed.

Another random thought is that you could try adding  `import gc; gc.collect()` right after `del`ing. This'll force the garbage collection to happen now. That should tell us if the issue is on Python's side who is too last to free the memory or if we're doing something wrong on yt's side.

14 Sep 2021 21:49:22 Clément Robert ***@***.***>:

> 
> *Bug report*
> 
> *Bug summary*
> 
> Generating a simple plot in a loop creates significant memory waste
> 
[...]

PR #3667, opened 2021-11-14
ENH: enable yt + interactive matplotlib interface (plot.show())

### LINKED ISSUE #2126: show() doesn't open an interactive Matplotlib window
### Bug report

**Bug summary**

Calling show on a slice object isn't showing an interactive mpl window. Just to double check it isn't a backend issue, I've tried to explicitly set an interactive backend in the snippet below. 

```python
import matplotlib
import numpy as np
import yt

matplotlib.use('qt5agg')
yt.toggle_interactivity()


@yt.derived_field(name = "funfield", units="cm", take_log=False)
def funfield(field, data):
    return (np.sin(data["phi"])**2 + np.cos(data["theta"])**2) * (1.0*data["r"].uq+data["r"])


ds = yt.load_uniform_grid({}, [128, 128, 128],
                          bbox=np.array([[0.0, 1.0], [0.0, np.pi], [0.0, 2*np.pi]]),
                          geometry="spherical")


s = ds.slice(2, np.pi/2)
p = s.to_pw("funfield", origin="native")
p.set_zlim("all", 0.0, 4.0)
p.show()
```

**Actual outcome**
No window is shown, and the program exits normally.

**Expected outcome**
An interactive Matplotlib window is shown.

**Version Information**

  * Operating System: MacOS
  * Python Version: 3.7.1
  * yt version: 4.5.0 (conda-forge)
  * Other Libraries (if applicable): Matplotlib 3.0.2

neutrinoceros:
## PR Summary

Goals:
- cleanup legacy code / technical debt accumulated in yt
- fix #2126

As noted in #2126, yt hooks with matplotlib at the level of figure canvas and managers.
Historically this was needed to workaround an important limitation from matplotlib where
apropriate backends (GUI/headless) were not determined at runtime, but this was
[solved in matplotlib 3.0.0](https://matplotlib.org/3.0.0/users/whats_new.html#improved-default-backend-selection) (it's even the first item on the changelog !)

Currently the only versions lower than 3.0.0 that we support are matplotlib 2.2.3 to 2.2.5, because 2.2.3 is the first release for which Python 3.7 wheels are available (and we don't support Python 3.6 anymore).
Note that 3.0.0 (Sept 19 2018) actually came out only 38 days after 2.2.3 (Aug 11 2018), and we're bound to bump soon anyway because 2.2.3 doesn't have wheels for Python 3.8 (next in line to become our oldest supported Python version). All things considered I think it is now reasonable to drop support for Matplotlib 2.x

Todo:
- [ ] finish up draft docstring changes
- [ ] cleanup the volume rendering side of yt.visualization
- [ ] cleanup yt.visualization._commons module (get_canvas_class and get_canvas, SUPPORTED_FORMATS and SUPPORTED_CANVAS_CLASSES are probably not needed anymore)

Everything should work as on the main branch with the difference that #2126 should now work as expected.
I've done some amount of testing manually within various Python REPLs (the builtin CPython interpreter, IPython, JupyterLab), but it'd be good to have more feedback than just mine.

I'll open as a draft to see wether I'm ahead in the right direction or if this patch is already breaking a lot of tests.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -113,12 +113,11 @@ def pytest_configure(config):
     ):
         config.addinivalue_line("filterwarnings", value)
 
-    if MPL_VERSION < Version("3.0.0"):
+    if MPL_VERSION < Version("3.1.0"):
         config.addinivalue_line(
             "filterwarnings",
             (
-                "ignore:Using or importing the ABCs from 'collections' instead of from 'collections.abc' "
-                "is deprecated since Python 3.3,and in 3.9 it will stop working:DeprecationWarning"
+                r"ignore:\nexamples.directory is deprecated; in the future, examples will be found relative to the 'datapath' directory.:matplotlib.cbook.deprecation.MatplotlibDeprecationWarning"
             ),
         )
     # at the time of writing, astropy's wheels are behind numpy's latest
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ packages = find:
 install_requires =
     cmyt>=0.2.2
     ipython>=2.0.0
-    matplotlib!=3.4.2,>=2.2.3  # keep in sync with tests/windows_conda_requirements.txt
+    matplotlib!=3.4.2,>=3.0.0  # keep in sync with tests/windows_conda_requirements.txt
     more-itertools>=8.4
     numpy>=1.14.5
     packaging>=20.9
@@ -96,7 +96,7 @@ mapserver =
 minimal =
     cmyt==0.2.2
     ipython==2.0.0
-    matplotlib==2.2.3
+    matplotlib==3.0.0
     more-itertools==8.4
     numpy==1.14.5
     unyt==2.8.0
--- a/tests/matplotlibrc
+++ b/tests/matplotlibrc
@@ -1,3 +1,4 @@
 #### MATPLOTLIBRC FORMAT
 
 backend : Agg
+figure.max_open_warning : 32
--- a/tests/matplotlibrc
+++ b/tests/matplotlibrc
@@ -1,4 +1,3 @@
 #### MATPLOTLIBRC FORMAT
 
 backend : Agg
-figure.max_open_warning : 32
--- a/tests/windows_conda_requirements.txt
+++ b/tests/windows_conda_requirements.txt
@@ -2,5 +2,5 @@ numpy>=1.19.4
 cython>=0.29.21,<3.0
 cartopy~=0.18.0
 h5py~=3.1.0
-matplotlib!=3.4.2,>=2.2.3  # keep in sync with setup.cfg
+matplotlib!=3.4.2,>=3.0.0  # keep in sync with setup.cfg
 scipy~=1.5.0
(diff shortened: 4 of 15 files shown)


neutrinoceros:
Actually I think I'm going to renounce fixing #2126 for the time being because it turns out that my patch, aside from making it work, has the extremely unpleasing side effect that figures cannot be automatically closed or garbage collected anymore (this is a known "feature" in matplotlib).
Maybe we'll be able to emulate the nice parts of `plt.show()` another time but for now I'll focus on cleaning up technical debt so that we can get closer to a solution.

neutrinoceros:
Currently this is failing because somehow, figures are still not garbage-collected and we get a warning from matplotlib that too many figures are open at once (>20).
However I note that my solution (implementing a `__del__` method for `PlotMPL` works as intended in this minimal example:
```python

class Plot:
    def __init__(self):
        self.figure = plt.figure()

    def __del__(self):
        plt.close(self.figure)

for i in range(30):
    p = Plot() 
```
Running this script, I don't get a warning from matplotlib, but I _do_ get it if I remove the custom `__del__` method.
My conclusion is that something is holding references to our figures, preventing them to be garbage collected (possibly related to #3513 ?).
This difficulty aside, I believe that the current state of this patch does fix #2126 after all.

neutrinoceros:
I've tried to locally increase the maximal number of coexisting figures in tests that produce a lot of them by using context managers but it doesn't seem to be reliable:
- some image tests are affected by this change
- figures are still cached for too long on windows + pytest, where we still reach the locally increased threshold

I'm tempted to conclude that #2126 cannot be done by _just_ simplifying the internals, but will probably require actually emulating `plt.show()`

neutrinoceros:
Turns out the image test failures are also happening in daily jobs (first time today), so it’s more likely due to a change upstream. I’ll investigate on that in a different issue/PR

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

neutrinoceros:
I see Kacper fixed the Jenkins instability (as far as I can tell), so trying this again : @yt-fido test this please

matthewturk:
Because this change moves us from the object-oriented interface to pyplot I think it needs a bit more discussion before we continue.

I recognize that the OO interface may be completely replicable from pyplot, and that the advantages are no longer present, but I want to make sure that we have accounted for that before making this change from using the low-level, fine-grained selection of canvases and figure objects.

neutrinoceros:
Moving away from OO is not the goal here, and it's probably not the right way to do it. The current state is barely a proof of concept that enables interactive plotting, but it does so at the cost of fine memory management, and pyplot is known to be terrible at that so it's likely that the actual _right_ way to do things would be to replicate the interesting bits of pyplot, but I can't phantom the technical debt this would create.

neutrinoceros:
I don't think this is worth keeping open. It's too blunt an attempt at solving a difficult problem, and I don't know when/if I'll be able to come back to it.

PR #3670, opened 2021-11-16
MNT: matplotlib 3.5 compat

### LINKED ISSUE #3558: BUG: symlog colorbar image tests need updating for matplotlib 3.5
### Bug report

**Bug summary**

In #3536 I deactivated minor ticks drawing for symlog colorbars with matplotlib between 3.2 and 3.4, but I was able to implement a solution for 3.5rc1.
This means that when 3.5.0 is released, I expect to see some failures in the affected image tests, because minor ticks will suddenly be drawn again. We'll need to bump image baselines accordingly.

neutrinoceros:
## PR Summary

I'm fixing a couple error messages here as an excuse to trigger CI so that I can update the answer store.
This PR will be a blocker unless it turns out that no update is needed.

goal ? fix #3558
why now ? matplotlib 3.5.0 just came out. I'm hoping that my work with the release candidate was enough to make this update pretty transparent
but I still expect we're going to see some small changes in our test images, at a level that will bother CI but not humans 🤞🏻

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -92,11 +92,9 @@ class PlotMPL:
         if figure_manager is not None:
             self.manager = figure_manager(self.canvas, 1)
 
-        for which in ["major", "minor"]:
-            for axis in "xy":
-                self.axes.tick_params(
-                    which=which, axis=axis, direction="in", top=True, right=True
-                )
+        self.axes.tick_params(
+            which="both", axis="both", direction="in", top=True, right=True
+        )
 
     def _create_axes(self, axrect):
         self.axes = self.figure.add_axes(axrect)
@@ -337,8 +335,7 @@ class ImagePlotMPL(PlotMPL):
             self.cb.set_ticks(yticks)
         else:
             self.cb = self.figure.colorbar(self.image, self.cax)
-        for which in ["major", "minor"]:
-            self.cax.tick_params(which=which, axis="y", direction="in")
+        self.cax.tick_params(which="both", axis="y", direction="in")
 
     def _get_best_layout(self):
 
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -92,7 +92,7 @@ answer_tests:
     - yt/frontends/owls/tests/test_outputs.py:test_snapshot_033
     - yt/frontends/owls/tests/test_outputs.py:test_OWLS_particlefilter
 
-  local_pw_040:  # PR 3441
+  local_pw_041:  # PR 3670
     - yt/visualization/tests/test_plotwindow.py:test_attributes
     - yt/visualization/tests/test_particle_plot.py:test_particle_projection_answers
     - yt/visualization/tests/test_particle_plot.py:test_particle_projection_filter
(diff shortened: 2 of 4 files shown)


neutrinoceros:
As expected we're only getting insignificant differences, and only 2 of them I might add
Screenshot 2021-11-16 at 10 38 25

neutrinoceros:
Updates for the answer store are here:
https://github.com/yt-project/answer-store/pull/27

regarding Jenkins, I'll wait for allowance to bump images everywhere needed. From what I've seen, every difference we're getting is equally insignificant so it shouldn't be controversial.

neutrinoceros:
adding "bug" label to satisfy the Mergeable bot.

I note that the prediction I gave in #3558 came true: symlog colorbars suddenly have minor ticks again, hurray !
Screenshot 2021-11-17 at 08 40 36

neutrinoceros:
Added a minimal patch to fix #3673 while I'm at it

yt/visualization/base_plot_types.py
```diff
@@ -337,8 +336,7 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
             self.cb.set_ticks(yticks)
         else:
             self.cb = self.figure.colorbar(self.image, self.cax)
-        for which in ["major", "minor"]:
-            self.cax.tick_params(which=which, axis="y", direction="in")
+        self.cax.tick_params(which="both", axis="y", direction="in")
```
neutrinoceros, line 339: I guess the explicit loop was necessary with old versions of matplotlib but it's not needed for the current oldest supported version (2.2.0)
see https://matplotlib.org/2.2.0/api/_as_gen/matplotlib.axes.Axes.tick_params.html?highlight=tick_params#matplotlib.axes.Axes.tick_params

matthewturk: OK, makes sense.

matthewturk (approved):
I don't see any reason this can't go in other than the minor comments about CI.

neutrinoceros:
> So if I'm reading it right, now our answer tests mandate matplotlib 3.5, right?

not explicitly, but answer testing is always run against the latest available version of matplotlib, so it's ran against 3.5 since it came out yesterday

matthewturk:
Once we get the new answers in and respawn testing this should go in.

neutrinoceros:
Cool, I'm bumping baselines for Jenkins then.

neutrinoceros:
We need to merge https://github.com/yt-project/answer-store/pull/27 as well so I can update here

neutrinoceros:
I forgot to bump one of the references, forced pushed. Now I expect images tests on Jenkins to pass

neutrinoceros:
removing the "backport" label since its purpose is now covered by having the PR triaged in the 4.0.2 milestone already.

cphyc (approved):

neutrinoceros:
Ah, turns out this can't be back ported (not automatically OR manually) unless we backport #3441 as well. This is because `  local_pw` tests were modified when switching to cmyt as a dependency. Now it seems to me that there are two ways we can keep CI green on the backport branch:
- backport #3441 as well as this
- exclude matplotlib 3.5 for the backport branch, which would be a pity since several bug fixes are intented to improve compatibility with it...

After writing this I don't think there's a way around it so I'll just manually backport #3441 + this, hopefully this is acceptable.

Issue #3384, opened 2021-06-28
Future proofing: Python 3.12 compatibility


neutrinoceros:
The distutils builtin package is slanted for removal in Python 3.12, which should be released around October 2023.
It is currently used in several places in yt, making it de facto incompatible with Python 3.12

In #3380, among other things, I'm setting an explicit upper limit to compatible Python versions, as a temporary bandaid that can be shipped with yt 4.0, but it'd be preferable to solve this for 4.1, at the latest.

_Originally posted by @neutrinoceros in https://github.com/yt-project/yt/pull/3380#r659900248_

neutrinoceros:
upstream issue worth keeping an eye on
https://github.com/pypa/setuptools/issues/2372

neutrinoceros:
Discussions with setuptools maintainers helped clarified that we can still rely on setuptools' compatibility shim for distutils for now.
https://github.com/pypa/setuptools/issues/2806#issuecomment-1193336591

This means that yt will still build on Python 3.12 as long as the shim is maintained, but it *will* also be removed at some point. An implication is that we should start pinning setuptools on release branches to make sure that releases can still be built from source in the future.
I've renamed this issue to better convey the actual problem.

PR #3674, opened 2021-11-17
BLD: require setuptools>=59 (partial migration out of distutils)

neutrinoceros:
## PR Summary
Another incremental step away from distutils, which is targetted for removal from the standard library in Python 3.12 (#3384)
This change requires setuptools 59, see https://setuptools.pypa.io/en/latest/history.html#v59-0-0 and in particular https://github.com/pypa/setuptools/issues/2698

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-  "setuptools>=19.6",
+  "setuptools>=59",
   # see https://github.com/numpy/numpy/pull/18389
   "wheel>=0.36.2",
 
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Faillures are expected due to matplotlib 3.5 being released yesterday, and solved in #3670

neutrinoceros:
numpy and SciPy are currently pinning setuptools to <0.49 in their releases, so it doesn't look reasonable to require >= 0.59 for now. Let's close this, it wasn't going to resolve the issue completely anyway.

PR #3678, opened 2021-11-19
Backports mpl compat to yt-4.0.x

neutrinoceros:
## PR Summary

This contains the following manual backports
Contents:
- #3442
- #3441
- #3670

I wasn't initially planning to backport cmyt PRs (3441 and 3442) but because of how image tests are implemented on Jenkins, I think this is the sanest way to keep CI green on the backport branch.
The only alternative approaches that I see would be to simply stop running image tests on the yt-4.0.x branch (or forbid matplotlib 3.5), or accept that they won't ever be green again.


neutrinoceros:
Important note: I also needed to bump the minimal required version of matplotlib to 2.1.0 (previously 2.0.1) on the 4.0 branch. I hope this is acceptable. I personally don't like this but it's still small enough a disruption that I'm considering it. If it is not acceptable I can also work on a new version of cmyt that'll be compatible with matplotlib 2.0.

neutrinoceros:
marking this as a blocker because no other PR to the yt-4.0.x branch has a chance at passing CI without _some_ resolution here

matthewturk (approved):

PR #3686, opened 2021-11-20
ENH: make plot callbacks with 'factor' argument accept 2-tuples to distinguish x/y resolutions

### LINKED ISSUE #3685: ENH: missing flexibility plot callbacks with a factor argument
### Bug report

**Bug summary**

This came out in a discussion with Samuel Whitman on Slack.
Currently, plot callbacks that have a "factor" arguement only accept a single float, which is often sufficient, but not always.
For instance

**Code for reproduction**

```python
import yt

plot = yt.SlicePlot(ds, "z", "density", width=(0.04, 0.01))

plot.annotate_quiver(
    "velocity_x",
    "velocity_y",
    factor=16,
    normalize=True,
    plot_args={"color": "white"},
)

plot.save("/tmp/factor.png")
```

**Actual outcome**
factor

As should be clear here, the resulting quiver plot isn't very useful because it's much denser in one direction than the one.
Internally, this is caused by having the default square resolution of the buffer (800, 800) used even for not physically square domains, and because "factor" is effectively counted in pixel space.

**Expected outcome**
It should not be necessary but we can approach something decent by changing the resolution of the frb itself
```python
plt.set_buff_size(800, 200)
```
factor

This is the solution I gave to Samuel, and it works, but the discoverability feels terrible from the user side.

1) I think that a better solution could be implemented by making `factor` arguments accept pairs of numbers instead of a single one, allowing the users to decouple x/y resolutions in a more discoverable fashion.
2) The _ideal_ solution would probably be to adapt the default size of image buffers for non-square projection domains, though it is probably best to fix #3632 first.

The first solution can be implemented immediately and doesn't collide with the second, nor does it loose all value if we ever implement the second one after that, so I'm just going to implement the easy one now.

neutrinoceros:
## PR Summary

Fix #3685

```python
import yt

plot = yt.SlicePlot(ds, "z", "density", width=(0.04, 0.01))

plot.annotate_quiver(
    "velocity_x",
    "velocity_y",
    factor=(16, 64), # new !
    normalize=True,
    plot_args={"color": "white"},
)

plot.save("/tmp/factor.png")
```
factor

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -2,7 +2,7 @@ import re
 import warnings
 from functools import wraps
 from numbers import Integral, Number
-from typing import Optional, Tuple, Union
+from typing import Any, Dict, Optional, Tuple, Union
 
 import matplotlib
 import numpy as np
@@ -47,9 +47,7 @@ def _verify_geometry(func):
     return _check_geometry
 
 
-def _validate_factor_tuple(
-    factor: Union[Tuple[Integral, Integral], Integral]
-) -> Tuple[int, int]:
+def _validate_factor_tuple(factor) -> Tuple[int, int]:
     if (
         is_sequence(factor)
         and len(factor) == 2
@@ -58,7 +56,7 @@ def _validate_factor_tuple(
         # - checking for "is_sequence" allows lists, numpy arrays and other containers
         # - checking for Integral type allows numpy integer types
         # in any case we return a with strict typing
-        return tuple(int(_) for _ in factor)
+        return (int(factor[0]), int(factor[1]))
     elif isinstance(factor, Integral):
         return (int(factor), int(factor))
     else:
@@ -945,7 +943,7 @@ class StreamlineCallback(PlotCallback):
         plot_args=None,
     ):
         PlotCallback.__init__(self)
-        def_plot_args = {}
+        def_plot_args: Optional[Dict[str, Any]] = {}
         self.field_x = field_x
         self.field_y = field_y
         self.field_color = field_color
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -1,8 +1,8 @@
 import re
 import warnings
 from functools import wraps
-from numbers import Number
-from typing import Optional, Tuple
+from numbers import Integral, Number
+from typing import Optional, Tuple, Union
 
 import matplotlib
 import numpy as np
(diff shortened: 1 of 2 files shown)


yt/visualization/plot_modifications.py
```diff
@@ -46,6 +47,26 @@ def _check_geometry(self, plot):
     return _check_geometry
 
 
+def _validate_factor_tuple(
+    factor: Union[Tuple[Integral, Integral], Integral]
+) -> Tuple[int, int]:
+    if (
+        is_sequence(factor)
+        and len(factor) == 2
+        and all(isinstance(_, Integral) for _ in factor)
```
matthewturk, line 56: Why does it have to be an integer?

matthewturk (commented):
I think we're mostly OK here but I want to verify from @neutrinoceros that it's in the right x,y order, and ask why it has to be an integer.

yt/visualization/plot_modifications.py
```diff
@@ -931,8 +962,8 @@ def __call__(self, plot):
 
         # We are feeding this size into the pixelizer, where it will properly
         # set it in reverse order
-        nx = plot.image._A.shape[1] // self.factor
-        ny = plot.image._A.shape[0] // self.factor
+        nx = plot.image._A.shape[1] // self.factor[0]
+        ny = plot.image._A.shape[0] // self.factor[1]
```
neutrinoceros, line 966: @matthewturk this is why I'm checking for integers. While floats are mathematically compatible here, ultimately we're counting pixels, so I think that passing a float signals either a misconception on the user side or a bug

neutrinoceros:
> I want to verify from @neutrinoceros that it's in the right x,y order

Sorry I don't understand what I need to justify/explain here

yt/visualization/plot_modifications.py
matthewturk: I'm not sure I agree with that.  We're specifically talking about the factor for changing the resolution; there is a substantial use case possible (even if not necessarily typically desirable) for values betweeen 0 and 1.

neutrinoceros: would that work, really ? Running this on the main branch
```python
import yt

plot = yt.SlicePlot(ds, "z", "density", width=(0.04, 0.01))

plot.annotate_quiver(
    "velocity_x",
    "velocity_y",
    factor=0.5;
    normalize=True,
    plot_args={"color": "white"},
)

plot.save("/tmp/factor.png")
```

I'm getting
```
YTPlotCallbackError: annotate_quiver callback failed
```

matthewturk:
@neutrinoceros I mean, just say you checked that the order of x and y factors versus the image make sense, since we often do some strange transposition etc.

neutrinoceros:
It works in cases that I've tried, but I'm not sure how to test is extensively.

yt/visualization/plot_modifications.py
matthewturk: Alright!  Let's consider that addressed then.

matthewturk (dismissed):

cphyc (approved):

neutrinoceros:
Alright guys thank you for reviewing this early but I need to solve a conflict so I expect your approvals to go away after that. I'll remove the draft status when everything is green again :)

neutrinoceros:
well actually we could also use auto-merge so no-one has to wait... removing the draft status now

cphyc (approved):

neutrinoceros:
@yt-fido test this please

neutrinoceros:
I pushed again to resolve a conflict. Auto merge is still enabled but I lost aprovals here. Could anyone approve (again) ?

matthewturk (approved):

PR #3687, opened 2021-11-20
BUG: fix a bug where PlotContainer.save would wrongfully ignore specified field extension

### LINKED ISSUE #3677: BUG: slice plot can not be saved in "pdf" format
### Bug report

**Bug summary**

Slice plot can not be saved in "pdf" format or any other format other than the default "png".

**Code for reproduction**

```python
import numpy as np
import yt

dataRandom = np.random.random([32,32,32])
data       = dict(density=(dataRandom,"Msun/Mpc**3"))

bbox       = [[0,1], [0,1], [0,1]]

ds         = yt.load_uniform_grid(data=data, domain_dimensions=dataRandom.shape, length_unit="Mpc", \
                                  bbox=np.array(bbox))

slc        = yt.SlicePlot(ds, 2, [('gas', 'density')])

slc.save("random.pdf")
```

**Actual outcome**

```
yt : [WARNING  ] 2021-11-18 22:23:25,995 Received two valid image formats 'pdf' (from `filename`) and 'png' (from `suffix`). The former is ignored.
yt : [INFO     ] 2021-11-18 22:23:25,996 Saving plot random_Slice_z_density.png
```

**Expected outcome**

I was able to generate "pdf" plots with yt 3.x. But when I update yt to 4.x recently, the issue appears. I did some tests, it seems to be caused by the following function in the file "yt/visualization/_commons.py"

```python
def validate_image_name(filename, suffix: Optional[str] = None) -> str:
    """
    Build a valid image filename with a specified extension (default to png).
    The suffix parameter is ignored if the input filename has a valid extension already.
    Otherwise, suffix is appended to the filename, replacing any existing extension.
    """
    name, psuffix = os.path.splitext(filename)
    psuffix = normalize_extension_string(psuffix)

    if suffix is not None:
        suffix = normalize_extension_string(suffix)

    if psuffix in SUPPORTED_FORMATS:
        if suffix in SUPPORTED_FORMATS and suffix != psuffix:
            mylog.warning(
                "Received two valid image formats '%s' (from `filename`) "
                "and '%s' (from `suffix`). The former is ignored.",
                psuffix,
                suffix,
            )
            return f"{name}.{suffix}"
        return str(filename)

    if suffix is None:
        suffix = "png"

    if suffix not in SUPPORTED_FORMATS:
        raise ValueError(f"Unsupported file format '{suffix}'.")

    return f"{filename}.{suffix}"
```
As can be seen, even if "psuffix" is a supported format, it is always replaced by "suffix". This is inconsistent with the comment at the beginning of the function that the replacement is only done when the specified format is not supported. After changing the condition sentences in the middle of the function to

```python
    if psuffix in SUPPORTED_FORMATS:
        return str(filename)
    else:
        if suffix in SUPPORTED_FORMATS and suffix != psuffix:
            mylog.warning(
                "Received two valid image formats '%s' (from `filename`) "
                "and '%s' (from `suffix`). The former is ignored.",
                psuffix,
                suffix,
            )
            return f"{name}.{suffix}"
```
everything  goes back to normal and I can save the plot in the format I specified.

**Version Information**

  * Operating System: Linux 2.6.32-504.30.3.el6.x86_64 x86_64
  * Python Version: 4.1.dev0
  * yt version: 3.8.8

neutrinoceros:
## PR Summary
fix #3677, see discussion therein

### DIFF
--- a/yt/visualization/_commons.py
+++ b/yt/visualization/_commons.py
@@ -1,11 +1,10 @@
 import os
 import sys
+import warnings
 from typing import Optional, Type
 
 import matplotlib
 from packaging.version import Version
-
-from yt.utilities.logger import ytLogger as mylog
 
 from ._mpl_imports import (
     FigureCanvasAgg,
@@ -68,11 +67,9 @@ def validate_image_name(filename, suffix: Optional[str] = None) -> str:
 
     if psuffix in SUPPORTED_FORMATS:
         if suffix in SUPPORTED_FORMATS and suffix != psuffix:
-            mylog.warning(
-                "Received two valid image formats '%s' (from `filename`) "
-                "and '%s' (from `suffix`). The former is ignored.",
-                psuffix,
-                suffix,
+            warnings.warn(
+                f"Received two valid image formats {psuffix!r} (from filename) "
+                f"and {suffix!r} (from suffix). The former is ignored."
             )
             return f"{name}.{suffix}"
         return str(filename)
@@ -81,7 +78,7 @@ def validate_image_name(filename, suffix: Optional[str] = None) -> str:
         suffix = "png"
 
     if suffix not in SUPPORTED_FORMATS:
-        raise ValueError(f"Unsupported file format '{suffix}'.")
+        raise ValueError(f"Unsupported file format {suffix!r}")
 
     return f"{filename}.{suffix}"
 
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -530,7 +530,7 @@ class PlotContainer(abc.ABC):
 
         Parameters
         ----------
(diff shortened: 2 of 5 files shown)


neutrinoceros:
switching to draft as I add/fix new tests

matthewturk (approved):

neutrinoceros:
Something terrible happened in image tests and I got a conflict + typing errors (most of them are not easily fixable, it seems, so I just went with `type: ignore`...)

neutrinoceros:
@yt-fido test this please

cphyc (approved):

neutrinoceros:
The bug (#3677) is still present on the yt-4.0.x branch so I'll open a manual backport

Issue #3690, opened 2021-11-22
BUG: image tests on Jenkins are unstable


neutrinoceros:
### Bug report

**Bug summary**

We're getting 237 errors in image tests on Jenkins, consistently on several unrelated PRs:
see queue https://tests.yt-project.org/job/yt_py38_git/

I do not understand what's happening or what might have caused this, internally or externally.

neutrinoceros:
maybe unrelated but I see some weird issues in the doc build too https://tests.yt-project.org/job/yt_docs_new/

Xarthisius:
> maybe unrelated but I see some weird issues in the doc build too https://tests.yt-project.org/job/yt_docs_new/

Why
```
Traceback (most recent call last):
  File "/var/jenkins_home/workspace/yt_docs_new/yt/visualization/plot_window.py", line 1391, in run_callbacks
    callback(cbw)
  File "/var/jenkins_home/workspace/yt_docs_new/yt/visualization/plot_modifications.py", line 39, in _check_geometry
    supp = self._supported_geometries
AttributeError: 'TitleCallback' object has no attribute '_supported_geometries'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/jenkins_home/workspace/yt_docs_new/doc/source/visualizing/colormaps/cmap_images.py", line 16, in <module>
    p.save(f"Projection_{cmap.replace(' ', '_')}.png")
  File "/var/jenkins_home/workspace/yt_docs_new/yt/visualization/plot_container.py", line 78, in newfunc
    args[0]._setup_plots()
  File "/var/jenkins_home/workspace/yt_docs_new/yt/visualization/plot_window.py", line 1279, in _setup_plots
    self.run_callbacks()
  File "/var/jenkins_home/workspace/yt_docs_new/yt/visualization/plot_window.py", line 1395, in run_callbacks
    raise YTPlotCallbackError(callback._type_name) from e
yt.utilities.exceptions.YTPlotCallbackError: annotate_title callback failed
```
AttributeError is a weird issue?

neutrinoceros:
This particular one is, because `TitleCallback` inherits from `PlotCallback`, which has a class-level default for `_supported_geometries`, so I have no idea how it's possible that we end up with 
```
AttributeError: 'TitleCallback' object has no attribute '_supported_geometries'
```

neutrinoceros:
in fact this error is easy to reproduce locally (annotate_title is broken on main), so I'll patch this

[...]

PR #3546, opened 2021-10-07
TYP: setup basic type checking and fix existing type errors

neutrinoceros:
## PR Summary
This is a work in progress in support to the ongoing effort to add more type hints to the code base.
Goals:
- [x] enable continuous type checking with mypy
- [x] fix any existing errors (type inconsistencies) with minimal changes
- [ ] add optional yt dependencies to Mypy workflow env and repeat (this may be left to a follow up PR)

PR #3692, opened 2021-11-22
BUG: hotfix for a functional breakage introduced in #3546

neutrinoceros:
## PR Summary
This should resolve the docs part of #3690

For reproduction:
```python
import yt

p.annotate_title("test")
p.save("/tmp/")
```

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -2,7 +2,7 @@ import re
 import warnings
 from functools import wraps
 from numbers import Number
-from typing import Tuple
+from typing import Optional, Tuple
 
 import matplotlib
 import numpy as np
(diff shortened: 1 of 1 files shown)


matthewturk:
OK, so *why* is this now a bug, when it wasn't before, and when we explicitly said all the typing stuff would be optional?  I'm confused, and this needs to be figured out before we do anything else.

neutrinoceros:
This is because #3546 never triggered the docs build, and apparently the only tests we have that exercise what's broken here are in there

matthewturk:
@neutrinoceros Why is this an error if it's just typing?  Because of mypy?

yt/visualization/plot_modifications.py
```diff
@@ -55,7 +55,7 @@ class PlotCallback:
     # "figure" this is disregarded.  If "force" is included in the tuple, it
     # will *not* check whether or not the coord_system is in axis or figure,
     # and will only look at the geometries.
-    _supported_geometries: Tuple[str, ...]
+    _supported_geometries: Optional[Tuple[str, ...]] = None
```
matthewturk, line 58: So the `Optional` is so that it's not required to be defined in subclasses, right?  Not that it isn't required to be defined as an attribute?  This is rather frustrating if this is the case, because if that is correct, it seems like it may be pointless to have *anything* in a base class that is typed as anything other than `Optional`.  Is that the case?

neutrinoceros:
It's not just typing, apparently I removed the default value for `PlotCallback._supported_geometries` in my typing PR. I was expecting CI to call me out when I broke things (and indeed, it happened a couple times), but it seems like this slipped in still

matthewturk:
I see, ok.

matthewturk (approved):

Issue #3705, opened 2021-12-03
BUG: incompatibility with most recent Cartopy versions


neutrinoceros:
### Bug report

**Bug summary**

We have been pinning cartopy to version 0.18 for a couple month to mitigate upstream compatibility issues.
In order to enable testing yt on Windows + conda + Python 3.10, I tried to loosen that rule to allow more recent versions of cartopy (that are compatible with Python 3.10) to install.
Doing so I discovered that we have one failing test under cartopy 0.19 and 0.20
I am reporting it to keep track of the problem, but short term I consider just skipping it would be preferable to not testing Windows + Python 3.10

**Code for reproduction**

with cartopy 0.19 or newer installed:

```
pytest yt/visualization/tests/test_geo_projections.py::TestGeoProjections::test_nondefault_transform
```

**Actual outcome**

```
E           ValueError: Invalid vmin or vmax
```

This can be reproduced at least with Python 3.9 + cartopy 0.19 and Python 3.10 + cartopy 0.20, so I really think this is a cartopy issue, not a Python incompatibility.

neutrinoceros:
I think I managed to reduce the failure to a "minimal" case that doesn't use yt
```python
import numpy as np
import matplotlib.pyplot as plt
from matplotlib.colors import LogNorm
import cartopy

data = np.random.random_sample((10, 10))
transform = cartopy.crs.Miller()
projection = cartopy.crs.Mollweide()

fig = plt.figure()
ax = fig.add_axes((0, 0, 1, 1), projection=projection)
cax = fig.add_axes((1, 1, 1, 1))

im = ax.imshow(
    data,
    extent=[-180.0, 180.0, -90.0, 90.0],
    norm=LogNorm(),
    transform=transform,
)
cb = fig.colorbar(im, cax=cax)
fig.savefig("/tmp/demo.png")
```

neutrinoceros:
reported upstream as https://github.com/SciTools/cartopy/issues/1955

neutrinoceros:
The current conclusion from upstream discussion is that cartopy 0.20 contains some intended breaking changes that need to be accounted for on our side.
This actually looks like a blocker for testing on Windows + Python 3.10 because cartopy is only tested on Windows so we can't skip it, but cartopy 0.18 doesn't seem compatible with Python 3.10 so we have to upgrade at some point.

At this stage I'm relatively confident that a solution to this can be found rather soon and be shipped with #3706

PR #3706, opened 2021-12-03
MNT: Add explicit support for Python 3.10 on windows

neutrinoceros:
## PR Summary

Complements #3543.
I've been testing this on my fork and I believe it is now ready.

I was also able to remove the pinning added in #3298 since [upstream issue](https://github.com/SciTools/cartopy/issues/1140) was resolved.
As a temporary measure I'm conditionally skipping the one test that fails with cartopy 0.20 (reported as https://github.com/yt-project/yt/issues/3705) because I don't think it is bad enough to prevent us from testing on windows + Python 3.10, the issue is linked in the test code. I assume the regression is actually upstream and should be reported there.

### DIFF
--- a/.github/workflows/build-test-pytest.yaml
+++ b/.github/workflows/build-test-pytest.yaml
@@ -34,12 +34,6 @@ jobs:
             python-version: 3.7
             dependencies: minimal
             tests-type: unit
-          # temporary: Python 3.10 is not available on conda, so we pin this job to Python 3.9
-          - os: windows-latest
-            python-version: '3.9'
-        exclude:
-          - os: windows-latest
-            python-version: '3.10'
 
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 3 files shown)


neutrinoceros:
I don't yet understand why this the test isn't skipped on Jenkins, so I'll switch this to draft for now and remove it from the 4.0.2 milestone

neutrinoceros:
Adding the backport label instead of the milestone to signal that it'd be cool to backport this on merge but it's not a blocker for the release

neutrinoceros:
Moving this to draft for now because it seems possible that a solution could be found quickly to avoid skipping the failing test.

neutrinoceros:
rebased on top of #3708

matthewturk:
This is back in draft -- should it be marked ready?

neutrinoceros:
the only reason this is in draft mode is because it's based on #3708
Now if you'd like, feel free to merge this directly, in which case I'll close and delete the other branch

neutrinoceros:
Failures are unrelated. I'll mark it as ready for review when CI is stabilised

matthewturk (approved):

PR #3708, opened 2021-12-06
BUG: restore compatibility with latest version of cartopy (0.20)

### LINKED ISSUE #3705: BUG: incompatibility with most recent Cartopy versions

neutrinoceros:
## PR Summary
Fix #3705 using the relevant bits of #3706
I'm cleaning a bunch of temporary measures that were recently taken to mitigate incompatibility issues, but the main patch, which consist in setting `extent=None`, was suggested by @greglucas, who I kindly invite to review this PR, particularly yt/visualization/base_plot_types.py

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -140,19 +140,8 @@ def pytest_configure(config):
         )
 
     if find_spec("cartopy") is not None:
-        # cartopy still triggers this numpy warning
-        # last checked with cartopy 0.19.0
-        config.addinivalue_line(
-            "filterwarnings",
-            (
-                "ignore:`np.float` is a deprecated alias for the builtin `float`. "
-                "To silence this warning, use `float` by itself. "
-                "Doing this will not modify any behavior and is safe. "
-                "If you specifically wanted the numpy scalar type, use `np.float64` here."
-                ":DeprecationWarning: "
-            ),
-        )
-        # this warning *still* shows up on cartopy 0.19 so we'll ignore it
+        # This can be removed when cartopy 0.21 is released
+        # see https://github.com/SciTools/cartopy/pull/1957
         config.addinivalue_line(
             "filterwarnings",
             (
--- a/tests/cartopy_requirements.txt
+++ b/tests/cartopy_requirements.txt
@@ -1,5 +1,0 @@
-# cartopy and its dependencies are non trivial to install
-# so we can't easily put these specs into setup.cfg
-shapely
---no-binary=shapely
-cartopy~=0.18.0
--- a/tests/ci_install.sh
+++ b/tests/ci_install.sh
@@ -16,13 +16,7 @@ osx|macOS)
     sudo mkdir -p /usr/local/man
     sudo chown -R "${USER}:admin" /usr/local/man
     brew update
-    # proj can be unpinned when upstream incompatibility issue is resolved. See
-    # https://github.com/SciTools/cartopy/issues/1140
-    export LDFLAGS="$LDFLAGS -L/usr/local/opt/proj@7/lib"
-    export CPPFLAGS="$CPPFLAGS -I/usr/local/opt/proj@7/include"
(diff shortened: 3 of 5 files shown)


greglucas:
This looks fine to me, I don't quite understand what the expected use-case is though. I think if a user wanted to specify a smaller extent for their image to yt, this would overwrite that and force it to be the global domain which may be undesirable? I figured you wanted to only update the extent within the tests and not the main code path.

neutrinoceros:
Well the `extent` param being passed to `imshow` isn't user-specified but rather computed dynamically, depending on the coordinate system, and it's rather obfuscated. I'm going to give as many details as I can here to document my decision process, mostly for posterity.

The test that was failing 

https://github.com/yt-project/yt/blob/491d3ec257609546f7ba5f6097b01b93fe120c84/yt/visualization/tests/test_geo_projections.py#L132

is using the `GeographicCoordinateHandler` class, as specified in the test's setup

https://github.com/yt-project/yt/blob/491d3ec257609546f7ba5f6097b01b93fe120c84/yt/visualization/tests/test_geo_projections.py#L64

This class is defined here 
https://github.com/yt-project/yt/blob/491d3ec257609546f7ba5f6097b01b93fe120c84/yt/geometry/coordinates/geographic_coordinates.py#L12

Now, the methods that are used to compute the plot extent are
https://github.com/yt-project/yt/blob/491d3ec257609546f7ba5f6097b01b93fe120c84/yt/geometry/coordinates/geographic_coordinates.py#L400

and
https://github.com/yt-project/yt/blob/491d3ec257609546f7ba5f6097b01b93fe120c84/yt/geometry/coordinates/geographic_coordinates.py#L422

But it seems to me that they were mostly duplicated from their equivalent methods in the `SphericalCoordinateHandler`, which I recently generalised to make able to plot partial domains, see
- #3529
- #3533
- #3610
- #3628

so my conclusion is that currently, "geographic" plots using cartopy in yt are de facto only supporting cases where the the whole latitudinal and azimuthal domain are present, so it seems ok to let `extent=None` if it's overridden by the cartopy transform afterwards.

That said, it would need to be changed when we decide to spend time adding support for plotting partial angular domains, in which case, I don't know how to handle it in cartopy.

matthewturk (approved):

neutrinoceros:
@matthewturk I was hoping to get your review here because you're the author of the extremely valuable comment I'm continuing here. Who else do you think could/should review it ?

matthewturk:
@neutrinoceros I am OK with this going in.  That comment was me trying to work out what was going on with cartopy - or, rather, how cartopy's totally reasonable actions (which I think ended in a kdTree search?) could get tricky and cause (effective) hangs when we set the extent in the wrong order.  I remember trying on a couple occasions to really narrow it down, and where I got to was that in some cases, setting the extent triggered a call to a kdTree search that took forever to finish and I was never sure why that was the case since I didn't think we were actually drawing on the canvas.

But honestly, I also think at one point I convinced myself of the opposite?  And I think if this works, that is a better test than what I had come up with.

PR #3433, opened 2021-07-13
UX: improve error message when DerivedField is passed invalid field keys (in particular, single strings)

neutrinoceros:
## PR Summary
This patch is a response to a question asked on Slack:
https://yt-project.slack.com/archives/C1YCPSX08/p1626106217214000

It improves the clarity of the error message obtained by running the following example code
```python
import yt

def _gasmetals(field, data):
    return data[('gas', 'density')]

ds = yt.load_sample("IsolatedGalaxy")

ds.add_field(
    'metals', # this is the error (which used to be valid previous to yt 4.0)
    function=_gasmetals, sampling_type='cell',
    units="code_density"
)
```

results:
```
# main branch
Traceback (most recent call last):
  File "/Users/robcleme/dev/yt-project/yt/t.py", line 8, in <module>
    ds.add_field(
  File "/Users/robcleme/dev/yt-project/yt/yt/data_objects/static_output.py", line 1543, in add_field
    self.field_info.add_field(name, function, sampling_type, **kwargs)
  File "/Users/robcleme/dev/yt-project/yt/yt/fields/field_info_container.py", line 394, in add_field
    self.alias(name, tuple_name)
  File "/Users/robcleme/dev/yt-project/yt/yt/fields/field_info_container.py", line 484, in alias
    self.add_field(
  File "/Users/robcleme/dev/yt-project/yt/yt/fields/field_info_container.py", line 396, in add_field
    self[name] = DerivedField(name, sampling_type, function, **kwargs)
  File "/Users/robcleme/dev/yt-project/yt/yt/fields/field_info_container.py", line 428, in __setitem__
    ftype, fname = key
ValueError: too many values to unpack (expected 2)
```
```
# this PR
Traceback (most recent call last):
  File "/Users/robcleme/dev/yt-project/yt/t.py", line 8, in <module>
    ds.add_field(
  File "/Users/robcleme/dev/yt-project/yt/yt/data_objects/static_output.py", line 1543, in add_field
    self.field_info.add_field(name, function, sampling_type, **kwargs)
  File "/Users/robcleme/dev/yt-project/yt/yt/fields/field_info_container.py", line 394, in add_field
    self.alias(name, tuple_name)
  File "/Users/robcleme/dev/yt-project/yt/yt/fields/field_info_container.py", line 484, in alias
    self.add_field(
  File "/Users/robcleme/dev/yt-project/yt/yt/fields/field_info_container.py", line 396, in add_field
[...]

PR #3710, opened 2021-12-07
Do not load plugins if ftype is None.

brittonsmith:
## PR Summary

This PR skips loading plugins if the `ftype` kwarg is None. This came up recently in an external frontend I maintain where the default field type (`default_fluid_type`) is None. Prior to PR #3433, we were silently loading plugins with a `None` field type. PR #3433 causes this to fail, which is probably the right thing, but I argue that we do not (and should not) explicitly disallow setting `default_fluid_type=None` and so should avoid loading the default fluid type plugins in this case. I've also moved the debug print into the function where it seemed to belong more. 

## PR Checklist

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -610,7 +610,6 @@ class Dataset(abc.ABC):
                 if nfields == 0:
                     mylog.debug("zero common fields, skipping particle union 'nbody'")
         self.field_info.setup_extra_union_fields()
-        mylog.debug("Loading field plugins.")
         self.field_info.load_all_plugins(self.default_fluid_type)
         deps, unloaded = self.field_info.check_derived_fields()
         self.field_dependencies.update(deps)
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -378,7 +378,10 @@ class FieldInfoContainer(dict):
         else:
             self[name] = DerivedField(name, sampling_type, function, **kwargs)
 
-    def load_all_plugins(self, ftype="gas"):
+    def load_all_plugins(self, ftype: Optional[str] = "gas"):
+        if ftype is None:
+            return
+        mylog.debug("Loading field plugins for field type: %s.", ftype)
         loaded = []
         for n in sorted(field_plugins):
             loaded += self.load_plugin(n, ftype)
--- a/yt/fields/tests/test_fields.py
+++ b/yt/fields/tests/test_fields.py
@@ -551,3 +551,13 @@ def test_ion_field_labels():
     for f in fields:
         label = getattr(fobj, f).get_latex_display_name()
         assert_equal(label, pm_labels[f])
+
+
+def test_default_fluid_type_None():
+    """
+    Check for bad behavior when default_fluid_type is None.
+    See PR #3710.
+    """
+    ds = fake_amr_ds()
+    ds.default_fluid_type = None
+    ds.field_list


munkm:
More broadly, is there a way we should be capturing these types of incompatibilities in external frontends? It sort of seems like we're putting a new maintenance burden on them when we should be capturing these when they're merged.

neutrinoceros (commented):
Not opposing this change, but I do not understand what's the golden standard for this. What's the purpose of have a `None` default fluid type, and what is the behaviour you're implementing here in fact equivalent ? If it is then I'd say it's much better in your proposed, explicit form.

And to answer Madicken, I don't think we could realistically catch this kind of implicit api change on our side as soon as they get merge: when they do, they've already passed CI so it's clear we're not exercising the case that's breaking, which means we probably have no idea it even exists.

In fact we don't even require as much from our own hard dependencies, that is what our "bleeding-edge" daily CI is for: help us catch upstream breaking changes while they're hot and we have a chance to report/fix them before they get released. For reference, at the moment we're doing this for numpy and matplotlib https://github.com/yt-project/yt/blob/main/.github/workflows/bleeding-edge.yaml

neutrinoceros:
other note: this doesn't need backporting because the breaking change only exists on the main branch. Good job catching it !

neutrinoceros:
@brittonsmith would it be possible to add a test to show what's fixed and help guarantee we won't break it again ?

brittonsmith:
Thanks both for your comments on this!

@munkm, I'm not sure what we can do to prevent this from happening. It was caught by automated testing that we do for the frontend against the development version of yt. I think that's a sufficient mechanism.

@neutrinoceros, for this frontend, the relevant fluid types are all defined at load and they represent satellite images taken at different times and/or detectors. A default fluid type does not really have any meaning in this case.

I think the main point of this PR is that the current implementation of load_all_plugins allows for undefined behavior to happen when ftype is None. I've added a small test for this.

neutrinoceros:
> @munkm, I'm not sure what we can do to prevent this from happening. It was caught by automated testing that we do for the frontend against the development version of yt. I think that's a sufficient mechanism.

agreed

> @neutrinoceros, for this frontend, the relevant fluid types are all defined at load and they represent satellite images taken at different times and/or detectors. A default fluid type does not really have any meaning in this case.

I see, thanks !

> I think the main point of this PR is that the current implementation of load_all_plugins allows for undefined behavior to happen when ftype is None. I've added a small test for this.

Excellent ! I confirm that your test correctly
- pass on yt 4.0.1 (the current golden standard)
- fails on main 
- pass again with this patch

only pre-commit.ci is currently red so I'll use pre-commit GitHub magic to fix it and then set auto-merge

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):

neutrinoceros:
@brittonsmith apparently there are conflicts that prevent merging, could you rebase the branch or merge from main ?

brittonsmith:
Ok, I hope that did what it was supposed to. I just force pushed the rebase to my fork.

neutrinoceros:
Looks like you actually deleted the branch and closed the PR, I doubt this was your intention.
I can't restore the branch for you but you should see a button to do it yourself here. Then I can resolve the conflict for you if you want

brittonsmith:
I've restored it, but the changes are now in main. I'm not sure exactly what happened. All I did was rebase my plugin branch on main and force push to my fork.

neutrinoceros:
For posterity, it seems that the changes were merged after all, despite this PR showing as "closed" rather than "merged".
We figured this was likely a result of an edge case of the brand new "auto-merge" GitHub feature, and there's nothing to be done here after all.

PR #3715, opened 2021-12-13
REL: update cibuildwheel, simplify wheel building workflow

neutrinoceros:
## PR Summary
main difference: not building the repo in place before running cibuildwheels fixes the issue we had with musllinux wheels and reenable them.
The cost is that wheels take a bit longer to build on Linux (about an hour VS 20min), but this seems reasonable, especially given that an hour is also the typical time needed to complete the job on Windows, so it doesn't really affect the duration of a given run.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -23,44 +23,21 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/setup-python@v2
-        name: Install Python
-        with:
-          python-version: 3.9
-
-      - uses: s-weigand/setup-conda@v1
-        if: matrix.os == 'windows-latest'
-        with:
-          update-conda: true
-          conda-channels: conda-forge
-          activate-conda: true
-          python-version: 3.9
-
       - uses: actions/checkout@v2
-      - name: Install cibuildwheel
-        run: python -m pip install cibuildwheel==2.2.2
-
-      - name: Install dependencies and yt
-        shell: bash
-        env:
-          dependencies: "full"
-          LDFLAGS: "-static-libstdc++"
-        run: source ./tests/ci_install.sh
 
       - name: Build wheels for CPython
-        run: |
-          python -m cibuildwheel --output-dir dist
+        uses: pypa/cibuildwheel@v2.3.0
         env:
           CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-*"
-          CIBW_SKIP: "*-musllinux_*"  # these fail due to side effects from previous builds
+          CIBW_SKIP: "*-musllinux_*"  #  numpy doesn't have wheels for musllinux so we can't build some quickly and without bloating
           CIBW_ARCHS_LINUX: "x86_64"
           CIBW_ARCHS_MACOS: "x86_64"
           CIBW_ARCHS_WINDOWS: "auto"
-          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
-          CIBW_MANYLINUX_I686_IMAGE: manylinux2014
           CIBW_ENVIRONMENT: "LDFLAGS='-static-libstdc++'"
+          CIBW_BUILD_VERBOSITY: 1
+          CIBW_BEFORE_BUILD: "rm -rf build/"  # working around .so files accumulation between builds leading up to bloated wheels
 
       - uses: actions/upload-artifact@v2
         with:
           name: wheels
-          path: ./dist/*.whl
+          path: ./wheelhouse/*.whl


neutrinoceros:
I've validated that the workflow runs as intended on my fork: https://github.com/neutrinoceros/yt/actions/runs/1572551486

Xarthisius:
I noticed that your `wheels.zip` is enormous. I checked the content of the whl files itself and it seems that the build is doing something stupid:

```
$ unzip -l yt-4.1.dev0-cp310-cp310-musllinux_1_1_x86_64.whl | grep kdtree.*so
  9147800  2021-12-13 12:15   yt/utilities/lib/particle_kdtree_tools.cpython-38-x86_64-linux-gnu.so
  9016784  2021-12-13 12:15   yt/utilities/lib/particle_kdtree_tools.cpython-310-x86_64-linux-gnu.so
  8889736  2021-12-13 12:15   yt/utilities/lib/particle_kdtree_tools.cpython-37m-x86_64-linux-gnu.so
  9016784  2021-12-13 12:15   yt/utilities/lib/particle_kdtree_tools.cpython-39-x86_64-linux-gnu.so
 10258896  2021-12-13 12:15   yt/utilities/lib/cykdtree/kdtree.cpython-39-x86_64-linux-gnu.so
 10267088  2021-12-13 12:15   yt/utilities/lib/cykdtree/kdtree.cpython-310-x86_64-linux-gnu.so
 10041760  2021-12-13 12:15   yt/utilities/lib/cykdtree/kdtree.cpython-37m-x86_64-linux-gnu.so
  9000008  2021-12-13 12:15   yt/utilities/lib/cykdtree/utils.cpython-39-x86_64-linux-gnu.so
  8860680  2021-12-13 12:15   yt/utilities/lib/cykdtree/utils.cpython-37m-x86_64-linux-gnu.so
 10869168  2021-12-13 12:15   yt/utilities/lib/cykdtree/kdtree.cpython-38-x86_64-linux-gnu.so
  9004104  2021-12-13 12:15   yt/utilities/lib/cykdtree/utils.cpython-310-x86_64-linux-gnu.so
  9135120  2021-12-13 12:15   yt/utilities/lib/cykdtree/utils.cpython-38-x86_64-linux-gnu.so
```

looks like each whl contains the artifacts from previous builds. That causes the bloating... 
Current version on `main` was also doing it wrong cause it contained the .so created by `pip install . ` and whatever the whl was building. It needs to be fixed

Xarthisius:
Also extended build time for musl suggest it's doing something silly like e.g. building numpy. I'd suggest to make the build more verbose for now to see what's going on in there.

neutrinoceros:
> It needs to be fixed, by e.g. invoking git clean -f -d -x yt

at what point should this be done ? In an additional step inserted right before `python -m cibuildwheel` I'd assume ?

> I'd suggest to make the build more verbose for now to see what's going on in there.

done !

neutrinoceros:
new job running on my fork with increased verbosity: https://github.com/neutrinoceros/yt/actions/runs/1574330580

Xarthisius:
> > It needs to be fixed, by e.g. invoking git clean -f -d -x yt
> 
> at what point should this be done ? In an additional step inserted right before `python -m cibuildwheel` I'd assume ?

Yup!

neutrinoceros:
I slightly modified the command. I kept the `-d` flag but removed the path (`yt`) because

> If any paths are specified, -d is irrelevant

see https://git-scm.com/docs/git-clean#Documentation/git-clean.txt--d

neutrinoceros:
new job : https://github.com/neutrinoceros/yt/actions/runs/1574606774

neutrinoceros:
So I'm not sure I'm reading the logs right but I think you got the right idea that we are accidentally building numpy too.
I've simplified the workflow further by using cibuilwheel's dedicated GitHub action instead of setting up Python, pip installing `cibuildwheel` and running it manually, let's see how it behaves now https://github.com/neutrinoceros/yt/actions/runs/1574628503

neutrinoceros:
last attempt ran noticeably quicker but still produced bloated wheels, so I'm trying again with the corrected git clean invoke

neutrinoceros:
hoping that I got it right this time, here's the latest job: https://github.com/neutrinoceros/yt/actions/runs/1575178202

neutrinoceros:
Still bloated. I'll try to investigate this another day. Suggestions still welcome

Xarthisius:
> Still bloated. I'll try to investigate this another day. Suggestions still welcome

That's my bad. Running `git clean` in `yt` only made sense if you were building yt in place, which you just changed. Removing `build/` should do the trick. I pushed to your branch hoping it would trigger the run, but now I see you were running those manually. Anyways, squash my commit and see if that worked.

.github/workflows/wheels.yaml
```diff
@@ -23,44 +23,20 @@ jobs:
...
-          LDFLAGS: "-static-libstdc++"
-        run: source ./tests/ci_install.sh
 
       - name: Build wheels for CPython
-        run: |
-          python -m cibuildwheel --output-dir dist
+        uses: pypa/cibuildwheel@v2.3.0
         env:
           CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-*"
-          CIBW_SKIP: "*-musllinux_*"  # these fail due to side effects from previous builds
```
Xarthisius, line 55: Just as I suspected, prolonged build times are caused by numpy build. Since there are no musl wheels available for numpy, I believe we should keep this `CIBW_SKIP`. Let's not be overeager and don't build all the things...

Xarthisius:
just FTR I tested this on my fork and it fixed the issue of multiple .so files in the wheel.

.github/workflows/wheels.yaml
neutrinoceros: agreed ! thanks for catching this

neutrinoceros:
Alright, disabled musllinux wheels again.
I see you pushed to my fork but I assume this was an accident, and I also forced-pushed into it so the commit is not on my branch anymore. Here it is:
https://github.com/neutrinoceros/yt/commit/9cf3edf22d397c770a0dee543ec1680cff65980f

I'm testing again without it here https://github.com/neutrinoceros/yt/actions/runs/1576637198
I'll restore it if needed

neutrinoceros:
Seems like we effectively minimised both the runtime (44min) and the final size of wheels (307MB total), thanks a lot for your help @Xarthisius !

cphyc:
Aren't numpy wheels built on manylinux now that https://github.com/numpy/numpy/pull/20102 has been merged in?

neutrinoceros:
They are. I don't know what's the Venn diagram for ManyLinux VS MuslLinux distros. I assume that they are disjoint or at least that none is a superset of the other ? Anyway let's keep building on manylinux for now, and we can add musllinux in the future if numpy goes that way

Xarthisius (approved):

PR #3720, opened 2021-12-14
ENH: add unpin_requirements.py script to version control

### LINKED ISSUE #3704: TST: "latest_deps" job is broken
### Bug report

**Bug summary**

#3543 broke the "yt_py38_latest_deps" daily job on Jenkins https://tests.yt-project.org/job/yt_py38_latest_deps/440/console

I believe the issue is that our Python script to edit requirements on the fly isn't smart enough to deal with `python_version < '3.10'` specifications, so it creates invalid specs and breaks installation.
I don't think this script is public, but @Xarthisius linked a version of it in #3392
I'll try to come up with a patch here.

neutrinoceros:
## PR Summary
Closes #3704

### DIFF
--- a/tests/unpin_requirements.py
+++ b/tests/unpin_requirements.py
@@ -1,0 +1,30 @@
+import configparser
+import io
+import re
+from collections.abc import MutableMapping
+
+PINNED_VERSION_REGEXP = re.compile(r",?(<|<=|==)([0-9a-z]+\.?)+")
+
+
+def unpin(s: str) -> str:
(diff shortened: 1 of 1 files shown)


matthewturk:
this is a lot of code here; is this really maintainable and necessary?

neutrinoceros:
This script is already in use on Jenkins, and indeed it required some maintenance recently because it was broken, but not having it in the repo made it harder for me to fix it.

matthewturk (approved):

PR #3729, opened 2021-12-23
TYP: upgrade mypy to 0.930 (enable TypeAlias)

neutrinoceros:
## PR Summary
Upgrading mypy to 0.930 enables using `typing.TypeAlias`, which I think come in handy in #3526
I'm opening as a draft because the new version has a couple regressions affecting us. I reported them upstream as
https://github.com/python/mypy/issues/11836 and https://github.com/python/mypy/issues/11836
In case they are resolved quickly we can avoid adding temporary `type: ignore` comments here

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,6 +49,7 @@ install_requires =
     toml>=0.10.2
     tqdm>=3.4.0
     unyt>=2.8.0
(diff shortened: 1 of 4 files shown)


neutrinoceros:
Probably not worth keeping this alive.

PR #3739, opened 2021-12-29
BUG: fix a bug in masking invalid values in line integral convolution callback

### LINKED ISSUE #3738: BUG: improper masking in line integral convolution callback
### Bug report

**Bug summary**

LIC callback doesn't currently apply any masking, which creates very visible artefacts when plotting non cartesian data, where typically some regions of the plot window are outside the actual simulation box

**Code for reproduction**

For instance
```python
import yt

ds = yt.testing.fake_amr_ds(
    fields=[("gas", "density"), ("gas", "velocity_x"), ("gas", "velocity_y")],
    units=["g/cm**3", "cm/s", "cm/s"],
    geometry="polar",
)
p.annotate_line_integral_convolution("velocity_x", "velocity_y", lim=(0.5, 0.65))
```

**Actual outcome**

before

**Expected outcome**

Regions where there is no data should be masked (as is the background)

neutrinoceros:
## PR Summary
fix #3738

after

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -3090,7 +3090,8 @@ class LineIntegralConvolutionCallback(PlotCallback):
         vectors = np.concatenate((pixX[..., np.newaxis], pixY[..., np.newaxis]), axis=2)
 
         if self.texture is None:
-            self.texture = np.random.rand(nx, ny).astype(np.double)
+            prng = np.random.RandomState(0x4D3D3D3)
+            self.texture = prng.random_sample((nx, ny))
         elif self.texture.shape != (nx, ny):
             raise ValueError(
                 "'texture' must have the same shape "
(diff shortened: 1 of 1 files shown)


neutrinoceros:
making this a draft for two reasons:
- I'm going to check wether other space-filling callback need a similar patch
- CI is currently unstable for different reasons so I'll be able to signal this is ready when things stabilise and I remove the draft status

neutrinoceros:
while I'm at it I also fixed reproducibility of these LIC plots

yt/visualization/plot_modifications.py
```diff
@@ -3104,6 +3105,10 @@ def __call__(self, plot):
         lic_data = lic_data / lic_data.max()
         lic_data_clip = np.clip(lic_data, self.lim[0], self.lim[1])
 
+        mask = ~(np.isfinite(pixX) & np.isfinite(pixY))
+        lic_data[mask] = np.nan
```
matthewturk, line 3109: Optional for future changes:

```
np.putmask(lic_data, mask, np.nan)
np.putmask(lic_data_clip, mask, np.nan)
```

matthewturk (approved):

Issue #3742, opened 2022-01-07
Test failures with numpy==1.22.0


Xarthisius:
### Bug report

**Bug summary**

New version of numpy was released on 12/31/21 and it's causing havoc in test suite, e.g.:

```
Traceback (most recent call last):
  File "/home/fido/shiningpanda/jobs/a7d58705/virtualenvs/d41d8cd9/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/fido/workspace/yt_py38_git/yt/utilities/answer_testing/framework.py", line 398, in __call__
    self.compare(nv, ov)
  File "/home/fido/workspace/yt_py38_git/yt/utilities/answer_testing/framework.py", line 1067, in compare
    compare_image_lists(new_result, old_result, self.decimals)
  File "/home/fido/workspace/yt_py38_git/yt/utilities/answer_testing/framework.py", line 816, in compare_image_lists
    expected = np.loads(zlib.decompress(old_result[i]))
  File "/home/fido/shiningpanda/jobs/a7d58705/virtualenvs/d41d8cd9/lib/python3.8/site-packages/numpy/__init__.py", line 315, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'loads'
```

There might be more.

neutrinoceros:
getting some 
```
RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 96 from PyObject
```
in jobs launched in the last hour. Plausibly related.

Xarthisius:
> getting some
> 
> ```
> RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 96 from PyObject
> ```
> 
> in jobs launched in the last hour. Plausibly related.

Yeah, note it's not even us. I'm seeing that in astropy.

neutrinoceros:
> Yeah, note it's not even us. I'm seeing that in astropy.

yup, I second that.
Weird how often astropy keeps growing incompatibilities with NumPy even though they started `oldest-supported-numpy`

neutrinoceros:
`np.loads()` was effectively deprecated a while back and removed in numpy 1.22, I don't understand how we've missed this.

https://numpy.org/doc/stable/release/1.22.0-notes.html#expired-deprecations-for-loads-ndfromtxt-and-mafromtxt-in-npyio

PR #3743, opened 2022-01-07
TST: fix compatibility with numpy 1.22

neutrinoceros:
## PR Summary
Attempt at fixing #3742
For now I'm just disabling astropy altogether, but a better solution would be prefered, I'm just not sure there's anything we can do about it on our side (other than pinning `numpy < 1.22`)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -121,22 +121,18 @@ def pytest_configure(config):
                 "is deprecated since Python 3.3,and in 3.9 it will stop working:DeprecationWarning"
             ),
         )
-    # at the time of writing, astropy's wheels are behind numpy's latest
-    # version but this doesn't cause actual problems in our test suite, so
-    # we allow this warning to pass.
-    # last checked with astropy 4.2.1
-    config.addinivalue_line(
-        "filterwarnings",
-        (
-            "ignore:numpy.ndarray size changed, may indicate binary incompatibility. "
-            "Expected 80 from C header, got 88 from PyObject:RuntimeWarning"
-        ),
-    )
+
     if find_spec("astropy") is not None:
-        # astropy triggers this warning from itself, there's not much we can do on our side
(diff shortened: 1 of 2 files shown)


neutrinoceros:
somehow test envs still end up including astropy. I'll sign off for today, but if anyone wants to inspect this (and possibly push to this branch), feel free to.

neutrinoceros:
got it. `glueviz` depends on `astropy`, which is why is still installing. Let's see if maybe the problem is _just_ glueviz (and possibly only on Python 3.10, since nosetests jobs are not failing)

neutrinoceros:
So I think I gathered enough evidence to state that the incompatibility in between Astropy's latest wheels and numpy 1.22.0; the problem isn't specific to a platform (seen on Ubuntu and Windows) or to a Python version (seen on Python 3.9 and 3.10).
However it seems to only show up in pytest sessions (not in nosetest) and is reproducible locally with
```
pytest yt/data_objects/tests/test_data_containers.py::TestDataContainers::test_to_astropy_table
```

I could get it down to a minimal reproducer:
```
pytest -c "import astropy"
```

Interestingly this issue seems to be specific to pytest, as the following command succeeds on the same install.
```
python-c "import astropy"
```
this seems simple enough to report upstream.

neutrinoceros:
figured out that the error was specifically caused by our live pytest configuration (`conftest.py`). It probably qualifies as a pytest bug but I don't understand it well enough to report that upstream. Anyway the PR now actually fixes #3742, and I've cleaned up my messy commit history. I'll remove the draft status when CI goes green again.

neutrinoceros:
back to draft because I think I can make it marginally better

cphyc (approved):
This looks good to me, even if catching a warning like the one below doesn't make me particularly happy.
But meh, it was already there before ;)

PR #3748, opened 2022-01-11
Revert "ENH: Directly open compressed files"

neutrinoceros:
Something seems to have gotten wrong with https://github.com/yt-project/yt/pull/3443, which was auto-merged before the Python 3.10 + Macos job completed. Apparently the job now takes almost 6hrs and actually fails.

see : https://github.com/yt-project/yt/actions/runs/1683855478

Reverts yt-project/yt#3443

@cphyc , it'd be much preferable to have an actual fix instead of going back and forth, please let me know if you think one can be found quickly.


neutrinoceros:
making this a draft, #3749 is more likely to be a reasonable solution.

PR #3749, opened 2022-01-11
BUG: ratarmount is broken on MacOS + Python 3.10, making this optional dep linux-only in test envs

neutrinoceros:
## PR Summary

Alternative PR to #3748
If CI passes and doesn't take forever here I'll simply close the revert PR

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -91,7 +91,7 @@ full =
     requests>=2.20.0
     scipy>=1.5.0
     xarray>=0.16.1
-    ratarmount~=0.8.1;platform_system!="Windows"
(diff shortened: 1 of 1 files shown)


cphyc (approved):
Assuming tests now pass. Note that I had to pin ratarmount's version as it seems the API changed from version 0.8 to 0.9.

neutrinoceros:
Thanks ! So I guess we'll advertise this feature as provisional and will need to tune the code a little more in the future ?

cphyc:
I guess so - note that adapting

> 

I think that's the wisest - especially since the feature is most useful on clusters to load archived data, most of which are running Linux.

PR #3751, opened 2022-01-18
TST: don't update homebrew on MacOS CI (revert #3187)

### LINKED ISSUE #3189: Revert changes introduced in #3187
### Bug report

**Bug summary**

#3187 forces brew to update on our CI infrastructure. This significantly increases the build time on OSX. To speed things up, we should remove the `brew update` statement in `test/ci_install.sh`. This will however be possible only if the default version of `brew` is recent enough.
As of 12 April 2021, the default recipes are too old causing `brew` to attempt to download binaries from bintray which is discontinued, thus failing.

This PR is to keep track of this information.

neutrinoceros:
## PR Summary
This reverts a temporary patch that was introduced in #3187, and closes #3189
I've checked on my fork that updating homebrew isn't necessary anymore and skipping it shaves off ~5 min off MacOS tests jobs.

### DIFF
--- a/tests/ci_install.sh
+++ b/tests/ci_install.sh
@@ -16,7 +16,6 @@ linux|Linux)
 osx|macOS)
     sudo mkdir -p /usr/local/man
     sudo chown -R "${USER}:admin" /usr/local/man
-    brew update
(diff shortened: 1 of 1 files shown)


cphyc (approved):
I remember this and I'm glad we don't need it anymore :)

PR #3752, opened 2022-01-18
WIP: particle file transactions and refactoring

chrishavlin:
This is an experimental refactoring of the particle frontends following on #3526. The goal is to simplify and normalize particle reading and selection across frontends in a way that facilitates daskification but also allows creation of a new frontend with less knowledge of yt's chunking and indexing framework. The main structural change is to move the responsibility of opening and reading from a single file to the frontend's `ParticleFile` class. The `io` routines then handle the logic for iterating over the chunks and applying selectors.

If this PR pans out, it might be a sufficient enough change to warrant a YTEP (perhaps on its own or as part of a wider dask-related pr)? But at present, I'd still classify this as experimental. So feel free to peruse and comment with that in mind!

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1872,6 +1872,21 @@ class ParticleFile(abc.ABC):
     def __hash__(self):
         return hash((self.filename, self.file_id, self.start, self.end))
 
+    def _nonzero_ptf(self, ptf):
+        for ptype, field_list in sorted(ptf.items()):
+            if self.total_particles[ptype] == 0:
+                continue
+            yield ptype, field_list, self.total_particles[ptype]
+
+    def _nonzero_ptypes(self, ptypes=None):
+        if ptypes is None:
+            ptypes = self.total_particles.keys()
+
+        for ptype in sorted(ptypes):
+            if self.total_particles[ptype] == 0:
+                continue
+            yield ptype, self.total_particles[ptype]
+
 
 class ParticleDataset(Dataset):
     _unit_base = None
--- a/yt/frontends/adaptahop/io.py
+++ b/yt/frontends/adaptahop/io.py
@@ -36,15 +36,13 @@ class IOHandlerAdaptaHOPBinary(BaseParticleIOHandler):
 
     def _read_particle_coords(self, chunks, ptf):
         # This will read chunks and yield the results.
-        chunks = list(chunks)
-        data_files = set()
+
         # Only support halo reading for now.
         assert len(ptf) == 1
         assert list(ptf.keys())[0] == "halos"
         ptype = "halos"
-        for chunk in chunks:
-            for obj in chunk.objs:
-                data_files.update(obj.data_files)
+
+        data_files = self._data_files_set(chunks)
         for data_file in sorted(data_files, key=attrgetter("filename")):
             pcount = (
                 data_file.ds.parameters["nhalos"] + data_file.ds.parameters["nsubs"]
@@ -56,14 +54,9 @@ class IOHandlerAdaptaHOPBinary(BaseParticleIOHandler):
 
     def _read_particle_fields(self, chunks, ptf, selector):
         # Now we have all the sizes, and we can allocate
-        chunks = list(chunks)
-        data_files = set()
         # Only support halo reading for now.
         assert len(ptf) == 1
         assert list(ptf.keys())[0] == "halos"
-        for chunk in chunks:
-            for obj in chunk.objs:
-                data_files.update(obj.data_files)
 
         def iterate_over_attributes(attr_list):
             for attr, *_ in attr_list:
@@ -73,9 +66,8 @@ class IOHandlerAdaptaHOPBinary(BaseParticleIOHandler):
                     yield attr
 
         halo_attributes = self.ds._halo_attributes
-
         attr_pos = partial(_find_attr_position, halo_attributes=halo_attributes)
-
+        data_files = self._data_files_set(chunks)
         for data_file in sorted(data_files, key=attrgetter("filename")):
             pcount = (
                 data_file.ds.parameters["nhalos"] + data_file.ds.parameters["nsubs"]
--- a/yt/frontends/gadget/data_structures.py
+++ b/yt/frontends/gadget/data_structures.py
@@ -565,14 +565,14 @@ class GadgetDataset(SPHDataset):
 
 class GadgetHDF5File(ParticleFile):
 
-    _fields_with_cols = ["Metallicity_", "PassiveScalars_", "Chemistry_", "GFM_Metals_"]
+    _fields_with_cols = ("Metallicity_", "PassiveScalars_", "Chemistry_", "GFM_Metals_")
 
     def _read_field(self, ptype, field_name, handle=None):
 
         field_name, col = self._sanitize_field_col(field_name)
 
         if self.total_particles[ptype] == 0:
-            # consider returning an empty array here instead of None
+            # TODO: consider returning an empty array here instead of None
             return
 
         if not handle:
--- a/yt/frontends/gadget/data_structures.py
+++ b/yt/frontends/gadget/data_structures.py
@@ -590,9 +590,9 @@ class GadgetHDF5File(ParticleFile):
     def _sanitize_field_col(self, field_name: str):
 
         if any(field_name.startswith(c) for c in self.fields_with_cols):
-            rc = field_name.split("_")
+            rc = field_name.rsplit("_", 1)
             col = int(rc[-1])
-            rfield = rc[0]
+            rfield = rc[:-1]
 
             if rfield == "Chemistry":
                 rfield = rfield + "Abundances"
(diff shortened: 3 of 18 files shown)


matthewturk:
@chrishavlin Do you think this is ready for review?  What would be necessary to get it into review shape?

chrishavlin:
Good question! I'll need to refresh the tests but I think my main hang up was whether there are more frontends that should be switched to the `transaction` handler that we introduced here?

And related -- think the the `transaction` handling should be added to the base `ParticleFile` class? The way we're using it across frontends makes me think yes...

I've also made some efforts to widen the refactoring to simplify `_read_particle_coords`, but my feeling now is that I shouldn't add any more complexity to this PR.

chrishavlin:
Ok, I'll work on adding the `transaction` to the base class and I'll get the tests passing then open it up for review!

yt/data_objects/static_output.py
```diff
@@ -1872,6 +1879,77 @@ def __eq__(self, other):
...
+
+        Returns
+        -------
+        np.ndarray or None : the data read from disk
+        """
+        with self.transaction(handle) as f:
+            data = self._read_from_handle(f, ptype, field)
+        return data
+
+    # @abc.abstractmethod
```
neutrinoceros, line 1933: should this be reactivated or cleaned up ?

yt/data_objects/static_output.py
```diff
@@ -1,4 +1,5 @@
 import abc
+import contextlib
```
neutrinoceros, line 2: Following the stdlib's dominant import style for any given module is preferred (for instance we use `import os; os.path.....` and never `from os.path import`).
In this case, following the docs, it's seems more conventional to write `from contextlib import contextmanager`. I also see that you're not using anything else from the namespace.

```suggestion
from contextlib import contextmanager
```

To avoid creating noise, I won't make the 15-ish related suggestions to change this throughout the PR.

also: I'm aware that the way we use `abc` currently contradicts this rule of thumb.

yt/data_objects/static_output.py
```diff
@@ -1872,6 +1879,77 @@ def __eq__(self, other):
...
+        ----------
+        handle : an open file handle
+        ptype : particle type
+        field : field name
+
+        Returns
+        -------
+        np.ndarray or None (should be np array_like more generally?)
+        """
+        # TO DO: switch to an abstract method when existing frontends implement it
```
neutrinoceros, line 1950: doesn't seem like a lot but this is the dominant convention in the code base so it'd be a shame to make this one escape grepping :)
```suggestion
        # TODO: switch to an abstract method when existing frontends implement it
```

though what would really lower the chance that this goes forgotten would be to open an issue and link it in the comment

yt/data_objects/static_output.py
```diff
@@ -1872,6 +1879,77 @@ def __eq__(self, other):
...
+            if self.total_particles[ptype] == 0:
+                continue
+            yield ptype, self.total_particles[ptype]
+
+    @contextlib.contextmanager
+    def transaction(self, handle=None):
+        # yields an open file handle, without double-reading.
+        if handle is None:
+            with self.open_handle() as handle:
+                yield handle
```
neutrinoceros, line 1902: ok so I'm experimenting on adding type hints to these methods and I don't have a definitive solution yet, but in any case it seems that rebinding the variable with an `as` isn't making the task any easier so I suggest :
```suggestion
            with self.open_handle() as new_handle:
                yield new_handle
```

yt/frontends/adaptahop/io.py
```diff
@@ -36,15 +36,13 @@ def _yield_coordinates(self, data_file):
 
     def _read_particle_coords(self, chunks, ptf):
         # This will read chunks and yield the results.
```
neutrinoceros, line 38: is this comment still relevant ?

neutrinoceros (commented):
Here's a first round of comments. I hope to finish this up later today.
Also on a general level: I'm surprised to see context managers implemented with embedded `with` statement. This is not like anything I could find in the stdlib's docs and I'm thinking maybe there's a more idiomatic way to achieve what they are doing ? I'll think a bit more about it.

yt/frontends/gadget/data_structures.py
```diff
@@ -178,8 +180,18 @@ def __init__(self, ds, io, filename, file_id, range=None):
...
+        self._field_dtype = defaultdict(lambda: default_field_type)
+        self._field_dtype["ParticleIDs"] = np.dtype(io._endian + ds._id_dtype)
+
         super().__init__(ds, io, filename, file_id, range)
 
+    @contextlib.contextmanager
+    def open_handle(self):
+        handle = open(self.filename)
+        yield handle
+        handle.close()
```
neutrinoceros, line 193: This seems like the correct way to write a @contextmanager
```suggestion
        handle = open(self.filename)
        try:
            yield handle
        finally:
            handle.close()
```

yt/frontends/gadget/tests/test_outputs.py
```diff
@@ -114,6 +117,55 @@ def test_particle_subselection():
...
+    ptype, field = "PartType0", "Density"
+    # with no existing handle (will open and close internally)
+    den0 = particle_file._read_field(ptype, field)
+    # with the transaction context manager
+    with particle_file.transaction() as handle:
+        den1 = particle_file._read_field(ptype, field, handle=handle)
+        assert array_equiv(den0, den1)
+        # a nested transaction will avoid re-opening
+        with particle_file.transaction(handle) as f:
+            assert handle == f
```
neutrinoceros, line 164:
```suggestion
            assert handle is f
```

yt/frontends/stream/data_structures.py
```diff
@@ -431,7 +431,9 @@ def update_data(self, data):
 
 
 class StreamParticleFile(ParticleFile):
-    pass
+    def open_handle(self):
+        # stream is in-memory, so do not need to implement open_handle
+        pass
```
neutrinoceros, line 436: IMO `pass` looks more like missing code in this context, when the method actually _is_ implemented by `return None`,so might as well make it explicit rather than implicit ? This isn't the only instance where I wanted to make this comment, but I'll avoid duplicating it.
```suggestion
        return None
```

yt/utilities/io_handler.py
```diff
@@ -292,30 +292,33 @@ def _read_particle_selection(
...
+    def _sorted_chunk_iterator(self, chunks):
+        # yields the data files contained in a chunks iterator sorted by
+        # filename and starting position with each file
+        data_files = self._data_files_set(chunks)
         yield from sorted(data_files, key=lambda x: (x.filename, x.start))
 
+
+# As a note: we don't *actually* want this to be how it is forever.  There's no
+# reason we need to have the fluid and particle IO handlers separated.  But,
+# for keeping track of which frontend is which, this is a useful abstraction.
```
neutrinoceros, line 320: since moving the comment is going to make git blame less useful, can you link to the original PR this was added ?
```suggestion
# for keeping track of which frontend is which, this is a useful abstraction.
# see https://github.com/yt-project/yt/pull/3526
```

neutrinoceros (commented):
another small batch

yt/frontends/gadget/data_structures.py
chrishavlin: yup, good call, I just copied over this bit and didn't think about it enough. What about just changing this block to follow the others:

```python
        with open(self.filename) as new_handle:
               yield new_handle
```

That internal `open` context should handle the closing, ya?

yt/data_objects/static_output.py
chrishavlin: Ya, I wasn't sure how to approach typing here... but will change the variable name!

yt/frontends/gadget/data_structures.py
neutrinoceros: yes this should be equivalent

chrishavlin: updated in the latest commit

yt/frontends/gadget/tests/test_outputs.py
chrishavlin: updated in the latest commit

yt/frontends/stream/data_structures.py
chrishavlin: updated in the latest commit

yt/utilities/io_handler.py
chrishavlin: updated in the latest commit

chrishavlin:
Latest commit takes care of the remaining suggestions, I also caught some missing abstract method implementations that I initially missed because we don't have tests for the http and sdf frontends (see #3908).

neutrinoceros:
@chrishavlin sorry this slipped by. Do we need to get this merged for yt 4.1 ?

chrishavlin:
@neutrinoceros no worries! it's also not critical... and actually might benefit from holding off for a bit. I've been refactoring the dask work off of main and I think it'll be more straightforward if the changes here are not merged until after I finish that initial refactoring (hoping to push some dask updates later this week!)

chrishavlin:
Didn't actually answer your question -- we do **not** need this for 4.1

neutrinoceros:
@chrishavlin what's the current status here ?

chrishavlin:
that's a good question.... I'll have to look :) will switch to draft for now though.

PR #3754, opened 2022-01-20
BUG: add missing MPL version check to a temporary hack introduced in #3161

### LINKED ISSUE #3170: [bug] errors in using log normalization with large data ranges
### Bug report

This is an overview of an upstream matplotlib bug and the temporary fix going into yt to address it so that we do not forget to update yt down the line.

In the recent versions of matplotlib (not sure exactly when this appeared in the 3.3, 3.4 series) if you use a log color scaling with `imshow` and if your data range is very large you can get some unexpected behavior including crashes. This arises from some internal re-scaling that `imshow` does, resulting in rounding error when your max/min data range span many orders of magnitude. If your data also includes zeros, this can lead to the the color scale min being set to 0, which errors for log scaling, which is the source of the error in yt's issue #2890. 

The current yt fix introduced by PR #3161 is to switch to symlog scaling if you have 0s and a data range greater than 15 orders of magnitude. As the fix in #3161 is not ideal, it should be eventually removed as the fixes get merged upstream. Note that the upstream fix introduces a new boolean keyword argument, `interp_postrgba`, that yt would need to pass to its calls to `imshow`, likely with an initial matplotlib version check for some period of time (I believe these changes would be implemented in `yt.visualization.base_plot_types.ImagePlotMPL._init_image`). 

matplotlib links for tracking:
* https://github.com/matplotlib/matplotlib/issues/19748 the issue of record
* https://github.com/matplotlib/matplotlib/issues/19851 an issue I submitted before seeing 19748. Closed as a duplicate but it's still useful because it has some simpler minimal examples reproducing the bug
* https://github.com/matplotlib/matplotlib/pull/18782 the pull request (in progress as of now) that will address this issue.

neutrinoceros:
## PR Summary
This is a proposal to close #3170 by adding a version check for MPL so that the "hack" introduced in #3161 is only applied when strictly necessary (with matplotlib from 3.3 to 3.4)

Taking the example from the original issue #2890 for reference
```python
import yt

ds = yt.load_sample("FIRE_M12i_ref11/snapshot_600.hdf5")
yt.ProjectionPlot(ds, "x", ("gas", "density")).save()
```

Here are the results:

#### w/ Matplotlib 3.2 to 3.4
3 4 3_main

#### w/ Matplotlib 3.5
3 5 1_main
(the one difference is that minor colorbar ticks are drawn, due to unrelated changes upstream)

#### w/ Matplotlib 3.2
3 2 2_branch

#### w/ Matplotlib 3.4
3 4 3_branch

#### w/ Matplotlib 3.5
3 5 1_branch

So my main motivation here is to make it clear when (if) we can drop the hack introduced in #3161 in the future, but
I also think the behaviour observed with this patch + Matplotlib 3.5 is more desirable from the user's standpoint.
I acknowledged that a counter argument would be that the current behaviour (main branch) is preferred because it's more consistent across matplotlib versions.
So I'd like to ask @chummels @jzuhone and @chrishavlin to weigh in and decide if this patch is desirable or if we should just close #3170 without a code change.
Of course everyone else is also welcome to provide feedback

### DIFF
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -1062,7 +1062,11 @@ class PWViewerMPL(PlotWindow):
                         f"Min = {np.nanmin(image)}, Max = {np.nanmax(image)}."
                     )
                     use_symlog = True
-                elif np.nanmax(image) > 0.0 and np.nanmin(image) == 0:
+                elif (
+                    (Version("3.3") <= MPL_VERSION < Version("3.5"))
+                    and np.nanmax(image) > 0.0
+                    and np.nanmin(image) == 0
+                ):
                     # normally, a LogNorm scaling would still be OK here because
                     # LogNorm will mask 0 values when calculating vmin. But
                     # due to a bug in matplotlib's imshow, if the data range


chrishavlin:
Great! I'm not opposed to the version-dependent behavior here. 

I am curious if you've experimented with the new `imshow` keyword parameter (`interpolation_stage`)? I'm actually a little surprised the **This Branch** with mpl 3.5 case didn't error as I thought it would be using the standard `imshow` internal scaling. But maybe there were upstream changes there too? 

I'm happy to approve this but will wait for others to have a chance to weigh in as well.

neutrinoceros:
No I haven't tried the new parameter yet. I think I'm secretly hoping we do not need it anywhere so we have less version-sensible code to maintain.

And yes, let's wait for at least one more approval from either John or Cameron before we consider this for merge

chrishavlin:
>  think I'm secretly hoping we do not need it anywhere so we have less version-sensible code to maintain. 

I totally agree there.

chummels:
I'm very happy to see that MPL updated to fix that problem with the zeros erroring out!  Thanks for staying on top of this, @neutrinoceros .  I certainly don't love my fix in #3295 that forces the symlog version of this plot to have all the orders of magnitude listed on the colorbar, as it's pretty cumbersome.  The default that we had before and can now again work with MPL 3.5 is much cleaner.  I'm a little bit hesitant to have such version-dependent behavior, but I think it probably makes sense in this case.  I'd like to get feedback from @jzuhone or @matthewturk or others on this before we merge it though.

neutrinoceros:
Thanks for your feedback !
So that everyone is on the same page I'd like to give my idea of the bigger picture here, as I've been maintaining many other such hacks in yt that are MPL-version dependent: MPL 3.3 and 3.4 series have been particularly disruptive for yt plots, so I tend to see them as "cursed" releases and I prefer having yt plots look slightly different in these versions rather than letting them dictate how yt plots should look against _any_ MPL version. Again this is just my opinion.

neutrinoceros:
Moving this somewhat aggressively to the 4.0.3 milestone to make sure it's at least discussed. I'd be okay with not included it in the release if anyone wants to oppose.

jzuhone (approved):

chrishavlin (approved):

PR #3757, opened 2022-01-22
ENH: cleanup frbfilter registration

neutrinoceros:
## PR Summary
Found a couple lines that are not needed since #2823 went in.
While I'm at it I'm also making `FixedResolutionBufferFilter` an abstract base class

### DIFF
--- a/yt/visualization/fixed_resolution.py
+++ b/yt/visualization/fixed_resolution.py
@@ -503,10 +503,7 @@ class FixedResolutionBuffer:
         return rv
 
     def setup_filters(self):
-        ignored = ["FixedResolutionBufferFilter"]
         for key in filter_registry:
-            if key in ignored:
-                continue
             filtername = filter_registry[key]._filter_name
 
             # We need to wrap to create a closure so that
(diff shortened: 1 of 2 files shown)


neutrinoceros:
here's an explanation for why the special casing to "ignore" the base class _used to_ be necessary, but isn't anymore:

## How registration used to work
```python
filter_registry = {}
class RegisteredFilter(type):
     def __init__(cls, name, b, d):
         type.__init__(cls, name, b, d)
         filter_registry[name] = cls

class FixedResolutionBufferFilter(metaclass=RegisteredFilter):
     pass
     
print(filter_registry)
```
output
```
{'FixedResolutionBufferFilter': <class '__main__.FixedResolutionBufferFilter'>}
```

## How it works now (since #2823)

```python
fitler_registry = {}

class FixedResolutionBufferFilter:
    def __init_subclass__(cls, *args, **kwargs):
        super().__init_subclass__(*args, **kwargs)
        filter_registry[cls.__name__] = cls

print(filter_registry)
```
output
```
{}
```

So the base class itself isn't part of the registry (which is desired)

matthewturk (approved):

neutrinoceros:
maybe @brittonsmith could give a second review here ?

PR #3758, opened 2022-01-27
REL: automate wheels publication to PyPI

neutrinoceros:
## PR Summary
We've been polishing this workflow on yt-astro-analysis and it's now working correctly so here's the port back to yt itself.
This workflow requires that a secret named "PYPI_TOKEN" be added to the repo (the name is arbitrary, that's just the name we've been using for yt-astro), this can be done by a PyPI maintainer.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -27,6 +27,8 @@ jobs:
 
       - name: Build wheels for CPython
         uses: pypa/cibuildwheel@v2.3.0
+        with:
+          output-dir: dist
         env:
           CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-*"
           CIBW_SKIP: "*-musllinux_*"  #  numpy doesn't have wheels for musllinux so we can't build some quickly and without bloating
@@ -40,4 +42,39 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: wheels
-          path: ./wheelhouse/*.whl
+          path: ./dist/*.whl
+
+  build_sdist:
+    name: Build source distribution
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build sdist
+        run: pipx run build --sdist
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: tarball
+          path: dist/*.tar.gz
+
+  upload_pypi:
+    needs: [build_wheels, build_sdist]
+    runs-on: ubuntu-latest
+    # upload to PyPI on every tag starting with 'yt-'
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/yt-')
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: tarball
+          path: dist
+
+      - uses: actions/download-artifact@v2
+        with:
+          name: wheels
+          path: dist
+
+      - uses: pypa/gh-action-pypi-publish@v1.4.2
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_token }}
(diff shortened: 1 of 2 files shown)


neutrinoceros:
CI is unstable for unrelated reasons, that are adressed in #3760

neutrinoceros:
(rebased on top of #3760)

matthewturk:
This looks good to me, but I can't be the only one to look at it.

neutrinoceros:
@brittonsmith 's review is an absolute requirement as far as I'm concerned

neutrinoceros:
We decided to keep uploading wheels manually for 4.0.2, but I'll add this to the next milestone :)

.github/workflows/wheels.yaml
```diff
@@ -40,4 +42,33 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build sdist
+        run: pipx run build --sdist
+
+      - uses: actions/upload-artifact@v2
+        with:
+          path: dist/*.tar.gz
```
brittonsmith, line 58: Not sure if this is necessary, but in yt_astro_analysis we gave this an explicit name that we used for the download action.
```suggestion
          name: tarball
          path: dist/*.tar.gz
```

.github/workflows/wheels.yaml
```diff
@@ -40,4 +42,33 @@ jobs:
...
+
+  upload_pypi:
+    needs: [build_wheels, build_sdist]
+    runs-on: ubuntu-latest
+    # upload to PyPI on every tag starting with 'yt-'
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/yt-')
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: artifact
```
brittonsmith, line 68: To get the source tarball
```suggestion
          name: tarball
```

.github/workflows/wheels.yaml
```diff
@@ -40,4 +42,33 @@ jobs:
...
+          path: dist
+
```
brittonsmith, line 70: We also needed a second download for the wheels.
```suggestion
    steps:
      - uses: actions/download-artifact@v2
        with:
          name: wheels
          path: dist
```

brittonsmith (commented):
I mostly just compared this to what we now have for yt_astro_analysis.

neutrinoceros:
Thanks a lot @brittonsmith, indeed I made a couple mistakes. I went back and reworked my way there and found the exact them changes you suggested were needed. It should be good now.

neutrinoceros:
taking that back, seems like we just broke the yaml file with invalid syntax, I'll push again (and maybe add a pre-commit hook to validate yaml files so it's easier to detect such mistakes in the future)

brittonsmith (approved):
This looks ok to me now.

neutrinoceros:
@matthewturk how about we move this to the 4.0.3 milestone ?

matthewturk:
Sounds good.

matthewturk:
I think we still need the `PYPI_SECRET`, right?

neutrinoceros:
correct

matthewturk:
I've added `PYPI_TOKEN` to this repository's secrets.

neutrinoceros:
I reaction to Github bumping `"windows-latest"` 3yrs forward sometime this week, I pushed another commit to pin exact versions of the images we use for Windows and MacOS. Building on old OSes maximise forward compatibility (this is the trick used to make Linux wheels as universal as possible with the ManyLinux docker image). We also do this already in yt_astro_analysis

brittonsmith (approved):

neutrinoceros:
back ported manually into #3789

PR #3762, opened 2022-01-30
TYP: add py.typed marker file to improve downstream type-checking

neutrinoceros:
## PR Summary

Add an empty marker file that's required by type-checkers declare our inline-type annotations
This is necessary to make downstream type-checking able to rely on existing annotations in yt.

See
https://mypy.readthedocs.io/en/latest/installed_packages.html#creating-pep-561-compatible-packages

as an example here's how adding this file changes Mypy's report for https://github.com/neutrinoceros/yt_idefix

with main
```
yt_idefix/fields.py:1: error: Skipping analyzing "yt.fields.field_info_container": module is installed, but missing library stubs or py.typed marker
yt_idefix/fields.py:2: error: Skipping analyzing "yt.fields.magnetic_field": module is installed, but missing library stubs or py.typed marker
yt_idefix/io.py:6: error: Skipping analyzing "yt.utilities.io_handler": module is installed, but missing library stubs or py.typed marker
yt_idefix/data_structures.py:13: error: Skipping analyzing "yt.data_objects.index_subobjects.grid_patch": module is installed, but missing library stubs or py.typed marker
yt_idefix/data_structures.py:13: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
yt_idefix/data_structures.py:14: error: Skipping analyzing "yt.data_objects.static_output": module is installed, but missing library stubs or py.typed marker
yt_idefix/data_structures.py:15: error: Skipping analyzing "yt.funcs": module is installed, but missing library stubs or py.typed marker
yt_idefix/data_structures.py:16: error: Skipping analyzing "yt.geometry.grid_geometry_handler": module is installed, but missing library stubs or py.typed marker
yt_idefix/loaders.py:7: error: Skipping analyzing "yt": module is installed, but missing library stubs or py.typed marker
Found 8 errors in 4 files (checked 13 source files)
```

with this marker
```
yt_idefix/data_structures.py: note: In member "__init__" of class "IdefixDataset":
yt_idefix/data_structures.py:124: error: "Type[IdefixDataset]" has no attribute "_dataset_type"
yt_idefix/data_structures.py:127: error: Incompatible types in assignment (expression has type "Optional[str]", variable has type "str")
yt_idefix/data_structures.py: note: In class "IdefixVtkDataset":
yt_idefix/data_structures.py:229: error: Signature of "_is_valid" incompatible with supertype "Dataset"
yt_idefix/data_structures.py:229: note:      Superclass:
yt_idefix/data_structures.py:229: note:          @classmethod
yt_idefix/data_structures.py:229: note:          def _is_valid(cls, filename: Any, *args: Any, **kwargs: Any) -> Any
yt_idefix/data_structures.py:229: note:      Subclass:
yt_idefix/data_structures.py:229: note:          @classmethod
yt_idefix/data_structures.py:229: note:          def _is_valid(cls, fn: Any, *args: Any, **kwargs: Any) -> bool
Found 3 errors in 1 file (checked 13 source files)
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -22,6 +22,7 @@ prune .tours
 recursive-include yt/visualization/volume_rendering/shaders *.fragmentshader *.vertexshader
 include yt/sample_data_registry.json
 include conftest.py
+include yt/py.typed
 prune yt/frontends/_skeleton
 recursive-include yt/frontends/amrvac *.par
 exclude .codecov.yml .coveragerc .git-blame-ignore-revs .gitmodules .hgchurn .mailmap
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,6 +133,7 @@ ignore = [
 
 [tool.mypy]
 python_version = 3.7
+show_error_codes = true
 ignore_missing_imports = true
 warn_unused_configs = true
 warn_unused_ignores = true


setup.cfg
```diff
@@ -117,6 +117,10 @@ typecheck =
     types-setuptools==57.4.0
     types-toml==0.10.0
 
+[options.package_data]
+yt =
+    py.typed
```
neutrinoceros, line 122: I'm surprised this section isn't already present. Maybe I should add this file to MANIFEST.in instead ?

neutrinoceros: Answered my own question thanks to https://stackoverflow.com/questions/45147837/including-data-files-with-setup-py

We may be able to migrate away from MANIFEST.in and put all that stuff in setup.cfg in the future, but that's out of scope here, and I think it's much preferable to keep everything in one place as long as possible. I'll use the MANIFEST instead.

neutrinoceros:
While I'm at it I'm activating an opt-in feature of mypy with f6996b93bc3206d9dbe29594ce768eb6a4502615

Error messages will now explicitly state what rule is violated, so we can choose to ignore just the ones we see with `# type: ignore[rule]` instead of deactivating the checker completely with `# type: ignore` as we do now. This makes it less likely that we'll hide future bugs with these comments, and will also help clarifying why they were added.

matthewturk (approved):

PR #3763, opened 2022-01-30
ENH: add a yt.version_info named tuple to ease version comparison downstream

### LINKED ISSUE #3761: ENH: add a first class citizen version-tuple
With the undergoing deprecation of distutils, the ecosystem is more and more relying on `packaging.version` to compare versions of dependencies at runtime.
However, using `packaging.version` internally makes for somewhat messy/busy code. As an example here's the minimal amount of code we need to write to check if e.g. matplotlib's version is greater than 3.5 at runtime
```python
import matplotlib

if Version(matplotlib.__version__) >= Version("3.5"):
    ...
```

But there is an alternative solution, and it has been popping out in various projects in our ecosystem recently:

- IPython 8.0.0 exposes a `version_info` attribute (taking inspiration from `sys.version_info` for Python itself), which is a simple `tuple[int, int, int]` that can be used for comparison.
- Matplotlib 3.5.0 exposes a `__version_info__` attribute for the same reason,
- numpy 1.23 will ship a `version_tuple` https://github.com/numpy/numpy/pull/20803

Long term, specifically, when we're done supporting matplotlib versions older than 3.5, we'll be able to simplify the example above as
```python
import matplotlib

if matplotlib.__version_info__ >= (3, 5):
    ...
```

So... the benefit of having such a "version tuple" exposed is not immediate as downstream need to require the first version that exposes it to be able to rely on it, however the cost is also very marginal so I think it is worth it. It'll actually lower the maintenance code downstream eventually.

(note: I know that downstream code written in a yt-version-sensitive fashion exists... because I wrote some myself.)

The reason I'm not proposing a patch straight away is that there is currently a naming discrepancy in the ecosystem (not 2 of the examples I gave above chose the same name for the attribute), but a uniformisation effort is being discussed (https://github.com/numpy/numpy/pull/20803#issuecomment-1024470449) so it seems worth waiting on this conversation to resolve first.
I anticipate this would be a reasonable task to add to the 4.1 milestone.

neutrinoceros:
## PR Summary
fix #3761
As mentionned in the issue, I'm waiting/hoping for the discussion upstream to settle on a shared convention for the name of this attribute, which is why I'm opening this as a draft.
I'll also need to add a paragraph of documentation somewhere.

## PR Checklist

### DIFF
--- a/yt/_version.py
+++ b/yt/_version.py
@@ -1,0 +1,55 @@
+from typing import NamedTuple
+
+from packaging.version import Version
+
+__all__ = [
+    "__version__",
+    "version_info",
+]
+
+__version__ = "4.1.dev0"  # keep in sync with setup.cfg
+
+
+class VersionTuple(NamedTuple):
+    """
+    A minimal representation of the current version number
+    that can be used downstream to check the runtime version
+    simply by comparing with builtin tuples, as can be done with
+    the runtime Python version using sys.version_info
+
+    https://docs.python.org/3/library/sys.html#sys.version_info
+    """
+
+    major: int
+    minor: int
+    micro: int
+    releaselevel: str
+    serial: int
+
+
+def _parse_to_version_info(version_str: str) -> VersionTuple:
+    # adapted from matplotlib 3.5
+    """
+    Parse a version string to a namedtuple analogous to sys.version_info.
+    See:
+    https://packaging.pypa.io/en/latest/version.html#packaging.version.parse
+    https://docs.python.org/3/library/sys.html#sys.version_info
+    """
+    v = Version(version_str)
+    if v.pre is None and v.post is None and v.dev is None:
+        return VersionTuple(v.major, v.minor, v.micro, "final", 0)
+    elif v.dev is not None:
+        return VersionTuple(v.major, v.minor, v.micro, "alpha", v.dev)
+    elif v.pre is not None:
+        releaselevel = {"a": "alpha", "b": "beta", "rc": "candidate"}.get(
+            v.pre[0], "alpha"
+        )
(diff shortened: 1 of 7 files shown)


yt/_version.py
```diff
@@ -0,0 +1,55 @@
...
+        return VersionTuple(v.major, v.minor, v.micro, "final", 0)
+    elif v.dev is not None:
+        return VersionTuple(v.major, v.minor, v.micro, "alpha", v.dev)
+    elif v.pre is not None:
+        releaselevel = {"a": "alpha", "b": "beta", "rc": "candidate"}.get(
+            v.pre[0], "alpha"
+        )
+        return VersionTuple(v.major, v.minor, v.micro, releaselevel, v.pre[1])
+    elif v.post is not None:
+        # fallback for v.post: guess-next-dev scheme from setuptools_scm
```
neutrinoceros, line 49: this comment is directly inherited from the function I copy-pasted (and lightly adapted) in matplotlib.
It's not actually relevant in yt where we don't use setuptools_scm, but I'm thinking we should keep the conditional branches anyway rather than opening the possibility for a crash.

matthewturk (dismissed):
Happy to weigh in on the name, but I guess I don't understand the issue well enough.

neutrinoceros:
Looks like the discussion is not headed to prompt resolution on the side of numpy so I'm going to un draft this. Consistency with the rest of the ecosystem is only a secondary goal and it's already inapplicable since IPython and Matplotlib went with different names. As long as we avoid introducing yet a new one I'm happy with this.

matthewturk:
I think we can/should move forward with this.

neutrinoceros:
@matthewturk just resolved conflicts.

matthewturk (approved):

PR #3772, opened 2022-02-03
ENH: avoid storing relative paths in Dataset attributes

### LINKED ISSUE #3774: BUG: loading from urls is (very likely) broken
### Bug report

**Bug summary**
This came up when I opened #3772, I'll just edit the relevant bits from there:

So, supposedly, `yt.load` supports loading from a url, as the following line suggests
https://github.com/yt-project/yt/blob/636057b11db35215432bcb1132fd9d44381ad07d/yt/loaders.py#L93

However, I don't know of a frontend where it's actually exercised.
Moreover I think the patch #3525 actually breaks this functionality (which didn't pop up in tests), but hasn't landed in a release yet.

Is there anywhere in code base that we actually implement loading anything from urls ? if so, then there's a bug to patch for yt 4.1, otherwise we could just simplify `yt.load`.

neutrinoceros:
## PR Summary

Initial motivation:
I found that storing relative paths attributes can lead to bugs, see downstream issues in an externally maintained frontend
https://github.com/neutrinoceros/yt_idefix/issues/88 and https://github.com/neutrinoceros/yt_idefix/issues/91, with the associated fixes https://github.com/neutrinoceros/yt_idefix/pull/89 then https://github.com/neutrinoceros/yt_idefix/pull/92

Indeed the way path attributes are set in the `Dataset` base class (as of current main branch) is unreliable because path data can be destroyed.

Let's reproduce the reference implementation in a self-contained example to illustrate the problem
```python
import os

def get_paths_attrs(filename):
    parameter_filename = os.path.expanduser(filename)

    basename = os.path.basename(filename)
    directory = os.path.dirname(filename)
    fullpath = os.path.abspath(directory)
    return parameter_filename, basename, directory, fullpath

print(get_paths_attrs(".."))
```
Running this in my `/tmp/` dir for illustration
```
('..', '..', '', '/private/tmp')
```

So I'm proposing here we resolve absolute paths instead of using the raw user-value as input here. I don't expect that this change would break anyone, so hopefully it's okay.

A consequence of this proposed change is that `Dataset.fullpath` becomes redundant with `Dataset.directory`, so I'm deprecating the former as I think the name is a bit less clear.

In order to implement this, I found it was useful to make these attribute read-only properties, which allowed me to harmonise their setup in a couple frontends with some custom logic, and I was able to discover and fix a handful of small bugs there.

I am also slightly changing filenames for the `backup_filename` attributes, where on main, we just append a suffix to the original filename, and here I'm stripping out the extension first, generalizing (and fixing) a custom behaviour that was implemented in the Athena frontend. Maybe this part is controversial since it's not backwards compatible with existing backup files for any dataset that's not Athena. I honestly do not know wether this is acceptable or not, but I can easily revert that if it isn't.

Incidentally, this fixes #3774 as well as #3777.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -223,18 +223,7 @@ class Dataset(abc.ABC):
         self.units_override = self.__class__._sanitize_units_override(units_override)
         self.default_species_fields = default_species_fields
 
-        # path stuff
-        filename = os.path.expanduser(filename)
-        self.parameter_filename = filename
-        self.basename = os.path.basename(filename)
-        self.directory = os.path.dirname(filename)
-        self.fullpath = os.path.abspath(self.directory)
-        self.backup_filename = self.parameter_filename + "_backup.gdf"
-        self.read_from_backup = False
-        if os.path.exists(self.backup_filename):
-            self.read_from_backup = True
-        if len(self.directory) == 0:
-            self.directory = "."
+        self._input_filename: str = os.fspath(filename)
 
         # to get the timing right, do this before the heavy lifting
         self._instantiated = time.time()
@@ -267,6 +256,41 @@ class Dataset(abc.ABC):
         except NoParameterShelf:
             pass
         self._setup_classes()
+
+    @property
+    def filename(self):
+        if self._input_filename.startswith("http"):
+            return self._input_filename
+        else:
+            return os.path.abspath(os.path.expanduser(self._input_filename))
+
+    @property
+    def parameter_filename(self):
+        # historic alias
+        return self.filename
+
+    @property
+    def basename(self):
+        return os.path.basename(self.filename)
+
+    @property
+    def directory(self):
+        return os.path.dirname(self.filename)
+
+    @property
+    def fullpath(self):
+        issue_deprecation_warning(
+            "the Dataset.fullpath attribute is now aliased to Dataset.directory, "
+            "and all path attributes are now absolute. "
+            "Please use the directory attribute instead",
+            since="4.1.0",
+        )
+        return self.directory
+
+    @property
+    def backup_filename(self):
+        name, _ext = os.path.splitext(self.filename)
+        return name + "_backup.gdf"
 
     @property
     def unique_identifier(self):
--- a/yt/frontends/athena/data_structures.py
+++ b/yt/frontends/athena/data_structures.py
@@ -494,11 +494,9 @@ class AthenaDataset(Dataset):
             unit_system=unit_system,
             default_species_fields=default_species_fields,
         )
-        self.filename = filename
         if storage_filename is None:
-            storage_filename = f"{filename.split('/')[-1]}.yt"
+            storage_filename = self.basename + ".yt"
         self.storage_filename = storage_filename
-        self.backup_filename = self.filename[:-4] + "_backup.gdf"
         # Unfortunately we now have to mandate that the index gets
         # instantiated so that we can make sure we have the correct left
         # and right domain edges.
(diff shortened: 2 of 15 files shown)


neutrinoceros:
okay with the additional commit I believe this now closes https://github.com/yt-project/yt/issues/3774

neutrinoceros:
rebranding as an enhancement because I'm doing more than a minimal patch here in the end

yt/frontends/athena/data_structures.py
```diff
@@ -494,16 +495,21 @@ def __init__(
...
         # instantiated so that we can make sure we have the correct left
         # and right domain edges.
         self.index
 
+    @property
+    def backup_filename(self):
+        if sys.version_info >= (3, 9):
+            return self.filename.removesuffix(".vtk") + "_backup.gdf"
+        else:
+            return self.filename[:-4] + "_backup.gdf"
```
neutrinoceros, line 511: need to make this more robust

yt/frontends/athena_pp/data_structures.py
```diff
@@ -266,11 +266,13 @@ def __init__(
...
-        self.filename = filename
         if storage_filename is None:
-            storage_filename = f"{filename.split('/')[-1]}.yt"
+            storage_filename = self.basename + ".yt"
         self.storage_filename = storage_filename
-        self.backup_filename = self.filename[:-4] + "_backup.gdf"
+
+    @property
+    def backup_filename(self):
+        return self.filename[:-4] + "_backup.gdf"
```
neutrinoceros, line 275: this looks like a bad copy-pasta from the Athena frontend, I should avoid propagating the bug and align with my refactor in Athena (here we don't want to strip 4 chars but 5 or 6, as the extension is not ".vtk" but ends with athdf, see _is_valid method def here)

yt/data_objects/static_output.py
```diff
@@ -223,18 +223,7 @@ def __init__(
...
-        self.basename = os.path.basename(filename)
-        self.directory = os.path.dirname(filename)
-        self.fullpath = os.path.abspath(self.directory)
-        self.backup_filename = self.parameter_filename + "_backup.gdf"
-        self.read_from_backup = False
-        if os.path.exists(self.backup_filename):
-            self.read_from_backup = True
-        if len(self.directory) == 0:
-            self.directory = "."
+        self._input_filename = os.fspath(filename)
```
neutrinoceros, line 226: os.fspath is great here because it invalidates a `None` and many objects that can't represent paths.
It also normalises `pathlib.Path` objects to `str`, which is desirable. However it's not sufficient to invalidate `bytes` -typed inputs (considered valid representations of paths, but incompatible with this attribute in many places where we implicitly assume a `str`).

yt/frontends/athena_pp/data_structures.py
neutrinoceros, line 275: this looks like a bad copy-pasta from the Athena frontend, I should avoid propagating the bug and align with my refactor in Athena (here we don't want to strip 4 chars but 5 or 6, as the extension is not ".vtk" but ends with athdf, see _is_valid method def here)

neutrinoceros:
(rebased on top of #3776)

yt/data_objects/static_output.py
```diff
@@ -223,18 +223,7 @@ def __init__(
...
         self.default_species_fields = default_species_fields
 
-        # path stuff
-        filename = os.path.expanduser(filename)
-        self.parameter_filename = filename
-        self.basename = os.path.basename(filename)
-        self.directory = os.path.dirname(filename)
-        self.fullpath = os.path.abspath(self.directory)
-        self.backup_filename = self.parameter_filename + "_backup.gdf"
-        self.read_from_backup = False
```
matthewturk, line 233: Hmm, this seems like it's now missing, yeah?

matthewturk (approved):

yt/data_objects/static_output.py
matthewturk: Never mind!

PR #3773, opened 2022-02-03
DEPR: deprecate yt.testing.run_nose and wrap it in the yt namespace

### LINKED ISSUE #3771: pytest is now a hidden hard dependency
### Bug report

We now import `pytest` directly in `yt/testing.py`, but `pytest` is not yet listed as a hard dependency. This will cause a `ModuleNotFoundError` to be raised after installing yt until you install pytest manually.

I'm opening an issue rather than a PR because I don't know whether people would prefer to put the `pytest` import in a try block or make it a dependency.

neutrinoceros:
## PR Summary
fix #3771
Opening as a draft because this will surely break CI until I refactor a couple imports

### DIFF
--- a/yt/__init__.py
+++ b/yt/__init__.py
@@ -89,7 +89,17 @@ from yt.loaders import (
     load_uniform_grid,
     load_unstructured_mesh,
 )
-from yt.testing import run_nose
+
+
+def run_nose(*args, **kwargs):
+    # we hide this function behind a closure so we
+    # don't make pytest a hard dependency for end users
+    # see https://github.com/yt-project/yt/issues/3771
+    from yt.testing import run_nose
+
+    return run_nose(*args, **kwargs)
+
+
 from yt.units.unit_systems import UnitSystem, unit_system_registry  # type: ignore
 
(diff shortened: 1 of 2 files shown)


matthewturk:
If we really want to have it not import at the top level, why not make `run_nose` be a thing here that just imports and runs inside the body of the function?

neutrinoceros:
Depending how big of a patch my current approach leads to this could be a simpler alternative indeed. I may be naive in that I don't expect anyone downstream to rely on this function being at the top level. Am I ?

matthewturk:
downstream maybe not, but people running tests, maybe yes.

neutrinoceros:
Ok I'll just deprecate it then

neutrinoceros:
I assume a deprecation without an alternative is fine ? Usually it isn't, but here I don't think there _is_ an alternative to redirect to.

neutrinoceros:
The brutal approach didn't break any test apparently, this will make the task a lot easier.

brittonsmith (approved):

PR #3776, opened 2022-02-04
TST: update tests using deprecated pytest api

### LINKED ISSUE #3775: TST: CI is broken with new pytest version (7.0)
### Bug report

**Bug summary**
Pytest 7.0 was released about an hour ago, and comes with a deprecation affecting this line
https://github.com/yt-project/yt/blob/d207f806e19930636b00dbba863b6382486b4932/yt/fields/tests/test_ambiguous_fields.py#L35

as a result CI is failing because of the deprecation warning. This seems easy enough to fix.

To reproduce
```
pip install -U pytest
pytest yt/fields/tests/test_ambiguous_fields.py
```

neutrinoceros:
Fix #3775, based on pytest's documentation https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -4,6 +4,7 @@ name: CI (bleeding edge)
 
 # goals: check stability against
 # - dev version of Python, numpy, and matplotlib
+# - pytest pre-releases
 # - building with future pip default options
 
 on:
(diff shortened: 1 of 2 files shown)


neutrinoceros:
I'm also making the bleeding-edge workflow use pytest pre-releases so that we get a headsup next time somthing like this happens, instead of fixing it after the fact.

neutrinoceros:
also checked that the bleeding-edge workflow is still in a good shape here https://github.com/neutrinoceros/yt/actions/runs/1794947483

cphyc (approved):

PR #3780, opened 2022-02-05
Format error string

Yash-10:
## PR Summary

The string to denote the error message seems to not be formatted here:

https://github.com/yt-project/yt/blob/23e65e19700b3892bdabea0d918b8274fb8aa1ff/yt/fields/field_info_container.py#L295

which gives the output:

```py
ValueError: ('Invalid sampling type %s. Valid sampling types are %s', 'discrete', 'cell, particle, local')
```

if an incorrect `sampling_type` is input to the `@yt.derived_field` decorator.

This small change formats the string to get a, hopefully, better output.

Thanks!

### DIFF
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -293,7 +293,8 @@ class FieldInfoContainer(dict):
         acceptable_samplings = ("cell", "particle", "local")
         if sampling_type not in acceptable_samplings:
             raise ValueError(
-                f'Invalid sampling type {sampling_type}. Valid sampling types are {", ".join(acceptable_samplings)}'
+                f"Received invalid sampling type {sampling_type!r}. "
+                f"Expected any of {acceptable_samplings}"
             )
         return sampling_type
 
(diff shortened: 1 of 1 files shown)


yt/fields/field_info_container.py
```diff
@@ -293,9 +293,7 @@ def _sanitize_sampling_type(sampling_type):
         acceptable_samplings = ("cell", "particle", "local")
         if sampling_type not in acceptable_samplings:
             raise ValueError(
-                "Invalid sampling type %s. Valid sampling types are %s",
-                sampling_type,
-                ", ".join(acceptable_samplings),
+                f'Invalid sampling type {sampling_type}. Valid sampling types are {", ".join(acceptable_samplings)}'
```
neutrinoceros, line 296:
```suggestion
                f"Received invalid sampling type {sampling_type!r}. "
                f"Expected any of {acceptable_samplings}"
```

neutrinoceros (commented):
Good catch ! Indeed this error message is buggy. While your fix lives up to the intended formatting, we can seize the occasion to make the message even clearer, so here's a suggestion (the key difference is that values are quoted, visually separating them from the rest of the message).

Thank you for taking the time to point it out and fix it !

yt/fields/field_info_container.py
neutrinoceros: actually I'll just commit this and merge after tests pass

neutrinoceros (approved):

PR #3782, opened 2022-02-06
BUG: fix pixelize_cylinder

### LINKED ISSUE #3781: BUG: Moirés in cylindrical pixelization
### Bug report

**Bug summary**

For datasets in spherical coordinates, plotting a poloidal slice (i.e in the "R, z", "phi"-normal plane) can produce very visible moirés patterns. This effect seems to be strongest when the latitudinal domain is narrow, and not noticeable for sufficiently large domains.

**Code for reproduction**

```python
import numpy as np
import yt

shape = (128, 128, 2)
data = {("gas", "density"): np.random.random_sample(shape)}
for theta_min in (0, 1.3, np.pi / 2 - 0.1):
    ds = yt.load_uniform_grid(
        data,
        shape,
        bbox=np.array([[0.5, 5.0], [theta_min, np.pi / 2 + 0.1], [-0.1, +0.1]]),
        geometry="spherical",
    )

    p = yt.SlicePlot(ds, "phi", "density")
    p.save(f"/tmp/phi_slice_{theta_min=}.png")
```
Going from widest to narrowest theta domains

**Actual outcome**
phi_slice_theta_min=0
phi_slice_theta_min=1 3
phi_slice_theta_min=1 5

This problem appears on yt 4.0.2 and on the dev branch, but it seems to be absent with 4.0.1 version, which was buggy in a different way (plot limits are not being calculated correctly)

phi_slice_yt-4 0 1_theta_min=0
phi_slice_yt-4 0 1_theta_min=1 3
phi_slice_yt-4 0 1_theta_min=1 4707963267948965

Precisely, the problem bisects to  ea10a379ac9802a04162455e163bf9f481487954 (#3533).

So I'm hesitant to call this a regression because the pixelizer routine wasn't changed in this commit, only the plot limits calculation was affected. I believe the patch only made this problem apparent but that it was already present in yt 4.0.1

A possibility I want to explore is that the pixelizer may be doing something wrong when pixels are far from being squares. Indeed the default buffer (800x800) size is most appropriate for square pictures. Right now it's a little hard to test because the `SlicePlot.set_buff_size` method seems to be broken too (it affects the plot limits but it shouldn't). I note however than the effect can be mitigated on the user side by reducing the buffer resolution, which still works for square buffers :
```python
    p.set_buff_size((200, 200))
    p.save(f"/tmp/phi_slice_{theta_min=}_rect_buffer.png")
```
phi_slice_theta_min=1 4707963267948965_rect_buffer

this problem was first discovered downstream in https://github.com/neutrinoceros/yt_idefix/issues/86

neutrinoceros:
## PR Summary
Fix #3781
I found that the actual problem was that the jumps in r and theta were constrained only by `dx` (horizontal pixel size) when they needed to be constained by `dy` too.

Using a variation of the script from the issue to show the change

```python
import numpy as np
import yt

shape = (32, 64, 1)
a = np.linspace(0, 1, 64)
b = np.ones((32, 64))
c = np.reshape(a*b, shape)
data = {("gas", "density"): c}

theta_min = np.pi / 2 - 0.1

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.5, 5.0], [theta_min, theta_min+0.1], [-0.1, +0.1]]),
    geometry="spherical",
)

yt.SlicePlot(ds, "phi", "density").save("/tmp/")
```

on the main branch

main

this branch
patch

### DIFF
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -525,7 +525,8 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
 
     cdef np.float64_t x, y, dx, dy, r0, theta0
     cdef np.float64_t rmax, x0, y0, x1, y1
-    cdef np.float64_t r_i, theta_i, dr_i, dtheta_i, dthetamin
+    cdef np.float64_t r_i, theta_i, dr_i, dtheta_i
+    cdef np.float64_t r_inc, theta_inc
     cdef np.float64_t costheta, sintheta
     cdef int i, pi, pj
 
@@ -558,9 +559,9 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
         rbounds[0] = 0.0
     if y0 < 0 and y1 > 0:
         rbounds[0] = 0.0
-    dthetamin = dx / rmax
-    for i in range(radius.shape[0]):
+    r_inc = 0.5 * fmin(dx, dy)
 
+    for i in range(radius.shape[0]):
         r0 = radius[i]
         theta0 = theta[i]
         dr_i = dradius[i]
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Switching to draft until I have time to go dig into failing tests.

neutrinoceros:
Turns out the error is unrelated, and caused by a simple deprecation warning from SciPy 1.8
I'm pushing the fix here to make CI go green but I'll also make a separate PR to fix the problem for everyone

neutrinoceros:
In fact I'll leave this as a draft until I can test that my optimisations do not break other use cases

neutrinoceros:
- failing tests on Jenkins may just need to be bumped (to be checked manually)
- I think my implementation can be optimised still: radial and angular increments in the loop are currently not dependent on theta, so they are de facto smaller than a pixel in most cases. This could be accelerated with a bit of geometry.

matthewturk:
I think this is the correct fix.  Thank you!

neutrinoceros:
Got it back to a safe state with no overly eager optimisation after I found that I was reintroducing similar bugs. This should now be mergeable

matthewturk:
It wasn't obvious to me that any optimizations were necessary.

neutrinoceros:
Oh I just felt bad about making this routine a little slower since it's now doing more jumps. But I guess I shouldn't use a broken implementation as a baseline !

Issue #3785, opened 2022-02-07
RAMSES loader does not recognize "groups" file structure


mtrebitsch:
### Bug report

**Bug summary**

The RAMSES loader does not recognize the "groups" output structure (`output_XXXXX/group_YYYYY/info_XXXXX.txt`) with the version 4.0.1 of the code. This is due to the way the output structure is recognized by the `RAMSESFileSanitizer` class (see below), and cannot be bypassed by (sym)linking all the output files a new "clean" structure (`output_XXXXX/info_XXXXX.txt`) due to the file sanitizer resolving the symlinks.

**Code for reproduction**

```python
ds = yt.load("/path/to/ramses/data/output_00075/group_00001/info_00075.txt")
```
or
```python
ds = yt.load("/path/to/symlinked/ramses/data/output_00075/info_00075.txt")
```

**Actual outcome**

![Screenshot of the issue](https://user-images.githubusercontent.com/9862893/152833343-05653c2d-f431-41d9-828c-289b0aa05508.png)

```
In [5]: ds = yt.load("/data84/obelisk/RHD/OUTPUT_DIR/output_00075/group_00001/info_00075.txt")
---------------------------------------------------------------------------
YTUnidentifiedDataType                    Traceback (most recent call last)
Input In [5], in <module>
----> 1 ds = yt.load("/data84/obelisk/RHD/OUTPUT_DIR/output_00075/group_00001/info_00075.txt")

File ~/.conda/envs/astro/lib/python3.10/site-packages/yt/loaders.py:98, in load(fn, *args, **kwargs)
     95 if len(candidates) > 1:
     96     raise YTAmbiguousDataType(fn, candidates)
---> 98 raise YTUnidentifiedDataType(fn, *args, **kwargs)

YTUnidentifiedDataType: Could not determine input format from `'/data84/obelisk/RHD/OUTPUT_DIR/output_00075/group_00001/info_00075.txt'`.
```

**Expected outcome**
`ds` should actually be the loaded dataset. Testing on a smaller simulation without the "groups" structure works.

**Version Information**
  * Operating System: GNU/Linux CentOS 8 stream 
  * Python Version: Python 3.10.2
  * yt version: 4.0.1, through `conda install --channel conda-forge yt`

**More details**
[...]

PR #3786, opened 2022-02-07
New "preserve_symlinks" keyword for the RAMSES frontend.

mtrebitsch:
The extra keyword in the loader for the RAMSES frontend makes sure that the symlinks are not resolved by the `RAMSESFileSanitizer`. This is particularly helpful if the file structure is actually a set of links to the files, for instance to bypass the fact that `RAMSESFileSanitizer` doesn't deal well with outputs including "groups".

## PR Summary

With yt 4, the RAMSES loader does not recognize the "groups" output structure (`/path/to/output/files/output_XXXXX/group_YYYYY/info_XXXXX.txt`). One relatively easy way to bypass that, in principle, is to do link all the output files in a separate folder without the nested "group" structure (`/path/to/linked/files/output_XXXXX/info_XXXXX.txt`). Beyond yt, this can be needed/useful for other analysis tools that are less flexible.

However, doing so does not work if the symlinks are resolved to their original location. This PR fixes this by adding an extra keyword `preserve_symlinks` to ensure that said links are preserved by the `RAMSESFileSainitizer`.
This is `False` by default, so it should not break any code.

This partially fixes issue #3785.

## PR Checklist

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -40,9 +40,12 @@ class RAMSESFileSanitizer:
     info_fname = None  # Path | None: path to the info file
     group_name = None  # str | None: name of the first group folder (if any)
 
-    def __init__(self, filename):
+    def __init__(self, filename, preserve_symlinks=False):
+
+        filename = Path(filename)
         # Resolve so that it works with symlinks
-        filename = Path(filename).resolve()
+        if not preserve_symlinks:
+            filename = Path(filename).resolve()
 
         self.original_filename = filename
 
@@ -656,6 +659,7 @@ class RAMSESDataset(Dataset):
         max_level=None,
         max_level_convention=None,
         default_species_fields=None,
+        preserve_symlinks=False,
     ):
         # Here we want to initiate a traceback, if the reader is not built.
         if isinstance(fields, str):
@@ -672,6 +676,9 @@ class RAMSESDataset(Dataset):
         cosmological:
         If set to None, automatically detect cosmological simulation.
         If a boolean, force its value.
+
+        preserve_symlinks:
+        If set to True, does not resolve the symlinks in the filenames.
         """
 
         self._fields_in_file = fields
@@ -685,7 +692,7 @@ class RAMSESDataset(Dataset):
             max_level, max_level_convention
         )
 
-        file_handler = RAMSESFileSanitizer(filename)
+        file_handler = RAMSESFileSanitizer(filename, preserve_symlinks=preserve_symlinks)
 
         # This should not happen, but let's check nonetheless.
         if not file_handler.is_valid:
@@ -983,4 +990,5 @@ class RAMSESDataset(Dataset):
 
     @classmethod
     def _is_valid(cls, filename, *args, **kwargs):
-        return RAMSESFileSanitizer(filename).is_valid
+        preserve_symlinks = kwargs.get("preserve_symlinks", False)
+        return RAMSESFileSanitizer(filename, preserve_symlinks).is_valid


yt/frontends/ramses/data_structures.py
```diff
@@ -40,9 +40,12 @@ class RAMSESFileSanitizer:
...
     group_name = None  # str | None: name of the first group folder (if any)
 
-    def __init__(self, filename):
+    def __init__(self, filename, preserve_symlinks=False):
+
+        filename = Path(filename)
         # Resolve so that it works with symlinks
-        filename = Path(filename).resolve()
+        if not preserve_symlinks:
+            filename = Path(filename).resolve()
```
Xarthisius, line 48: Is there any situation when using `.resolve()` would be a bad thing? If no, how about we skip all the hassle of adding additional kwargs here?

mtrebitsch: The default was to use `.resolve()` so I assumed that there was a good reason for it to be there. I found that _removing_ it fixed my issue, but I didn't want to lead to backward-incompatible code without a good reason.

Xarthisius: Oops, sorry I misread it.

neutrinoceros:
That was quick ! 
So actually after reading the report I now understand that there should be another, cleaner way to address #3785, and I would prefer to go this route if we can avoid adding flags, which should be a last resort (sorry for suggesting it right away !).
I think you can use this patch in production but I'd rather not merge it if we can implement the correct fix. Do you want to have a go at it ?

neutrinoceros:
That being said I stand by my statement that not having the option to skip link resolving is a bug since it breaks your workflow, which seems absolutely sound to me (sorry for going back and forth on this). I think we could merge this on the condition that it _doesn't_ close #3785, which should only be closed by the _correct_ patch. What do you think ?

Note that if you go forward with this patch, you need to take care of linting (pre-commit.ci), see our docs on the matter https://yt-project.org/doc/developing/developing.html#automatically-checking-and-fixing-code-style

yt/frontends/ramses/data_structures.py
```diff
@@ -983,4 +990,5 @@ def read_namelist(self):
 
     @classmethod
     def _is_valid(cls, filename, *args, **kwargs):
```
neutrinoceros, line 992:
```suggestion
    def _is_valid(cls, filename, *args,  preserve_symlinks=False, **kwargs):
```

yt/frontends/ramses/data_structures.py
```diff
@@ -983,4 +990,5 @@ def read_namelist(self):
...
-        return RAMSESFileSanitizer(filename).is_valid
+        preserve_symlinks = kwargs.get("preserve_symlinks", False)
+        return RAMSESFileSanitizer(filename, preserve_symlinks).is_valid
```
neutrinoceros, line 994:
```suggestion
        return RAMSESFileSanitizer(filename, preserve_symlinks).is_valid
```

yt/frontends/ramses/data_structures.py
```diff
@@ -656,6 +659,7 @@ def __init__(
         max_level=None,
         max_level_convention=None,
         default_species_fields=None,
+        preserve_symlinks=False,
```
neutrinoceros, line 662: If we're going to add more flags here, let's make them keyword only
```suggestion
        *,
        preserve_symlinks=False,
```

neutrinoceros (commented):
adding a couple suggestions on function signatures

neutrinoceros:
One last thought for today: how about naming the flag `resolve_symlinks` instead so that there’s a direct correspondence between the verb and the conditional action ?

mtrebitsch:
>  I think you can use this patch in production but I'd rather not merge it if we can implement the correct fix. Do you want to have a go at it ?

Happy to think about it, but I'm not sure I understand enough of the code to not break anything. In any case, it would take a bit more time, but I can try to have a look (I'll maybe bug @cphyc for help with the Ramses frontend too…)

> That being said I stand by my statement that not having the option to skip link resolving is a bug since it breaks your workflow, which seems absolutely sound to me (sorry for going back and forth on this). I think we could merge this on the condition that it _doesn't_ close #3785, which should only be closed by the _correct_ patch. What do you think ?

That would work for me :-)

> Note that if you go forward with this patch, you need to take care of linting (pre-commit.ci), see our docs on the matter https://yt-project.org/doc/developing/developing.html#automatically-checking-and-fixing-code-style

Thanks for this, and for all the suggestions. Implementing and updating ASAP.

yt/frontends/ramses/data_structures.py
```diff
@@ -677,8 +678,8 @@ def __init__(
         If set to None, automatically detect cosmological simulation.
         If a boolean, force its value.
 
-        preserve_symlinks:
-        If set to True, does not resolve the symlinks in the filenames.
+        resolve_symlinks:
+        If set to True, resolves the symlinks in the filenames.
```
neutrinoceros, line 682: This is misleading if we don't also state that True is the default value

neutrinoceros:
I only have one comment remaining, after that I'd be inclined to merge this as a bug fix but I'd require a second reviewer because expanding the api could be regarded as a feature and not a bugfix. In my opinion this is a necessary addition to a half-baked feature (auto-resolving everything) which it's too late to take back, which is why I consider it a bugfix.

If a second reviewer agrees with this view, we could even ship this with the next bugfix release

cphyc:
I'm on holiday till next week so if we want this merged ASAP, dont wait for me!

8 Feb 2022 16:55:35 Clément Robert ***@***.***>:

neutrinoceros:
Found myself working on a related problem here #3801, and now I'm thinking we could maybe avoid adding a flag to the API if we instead try to validate the user-input path with both the resolved and the unresolved version. This could work for all known workflows for RAMSES data, and would improve the user experience. @mtrebitsch, what do you think about this ?

mtrebitsch:
Thanks @neutrinoceros, I'll try that over the week-end and I'll get back to you. This seems much cleaner than adding new keywords all the time indeed :-)

mtrebitsch:
One question though (sorry, it hit me when I had just sent the previous comment): is `original_filename` used anywhere? And is it expected to be the one passed as an argument or actually used by the rest of the function?

neutrinoceros:
As far as I can tell, `original_filename` *should* reflect user-provided data (right now it's not, because symlinks are resolved). Because it can be later used as part of an error message (see #3801), it could be really confusing that it doesn't always match what's been typed on the user side. I don't think it's used at the moment but I very much intend to change that with #3801 :)

I think it's okay rework this class so that this attribute better matches its name, and I'm confident that `RAMSESFileSanitizer` isn't meant to be public api so we're free to change it as much as needed.

neutrinoceros:
Excellent ! I edited the title again. Given my involvement I don't want to merge this myself, but hopefully @cphyc will be happy to review it now :)

neutrinoceros (approved):

cphyc (approved):
I am not particularly happy about how clunky this turns out to be but it'd require a refactor for it to be nicer.

neutrinoceros:
Manually back ported in #3789

Issue #3471, opened 2021-08-12
Lorentz-Heaviside Magnetic Units vs CGS


forrestglines:
I ran into some confusing behavior with magnetic fields while creating a yt frontend for Parthenon/AthenaPK using the Athena++ frontend as a base. I realize that the current Athena++ frontend treats the magnetic fields in CGS but within the code magnetic fields are treated with Lorentz-Heaviside units (i.e. there are no factors of 4 pi in Maxwell's equations, magnetic pressure is "B^2/2"). This leads to some awkward differences between expressions in the code and expressions in an analysis with yt.

For example for the magnetic energy density/pressure
```python
#Won't match yt
ad_cons["athena_pp","Etot"]

magnetic_energy_density = 0.5*( ad_prim["athena_pp","Bcc1"]**2
                               +ad_prim["athena_pp","Bcc2"]**2
                               +ad_prim["athena_pp","Bcc3"]**2)

#Won't match
ad_cons["gas","magnetic_energy_density"]
```
and when calculating the total energy density:
```python
#As implemented in the code
Etot = 0.5*ad_prim["athena_pp","rho"]*( ad_prim["athena_pp","vel1"]**2 
                                       + ad_prim["athena_pp","vel2"]**2
                                       + ad_prim["athena_pp","vel3"]**2) \
      + ad_prim["athena_pp","press"]/(gamma-1) \
      + 0.5*( ad_prim["athena_pp","Bcc1"]**2
             +ad_prim["athena_pp","Bcc2"]**2
             +ad_prim["athena_pp","Bcc3"]**2)

#Won't match yt
ad_cons["athena_pp","Etot"]
```
Both these expressions need factors of 1/4pi in the analysis that don't appear in the code.

The community has seemed to have gotten by fine with the Athena++ frontend by just writing their analysis in Gaussian units while the code lives in Lorentz-Heaviside units. However, this has become awkward when we're trying to use yt to verify AthenaPK code since it's not always obvious where extra factors of 1/4pi need to be included in our analysis.

Is it possible to change how `yt` computes magnetic pressure? Or would it be feasible to add a "lorentz_heaviside" unit system that fits within the CGS unit system but uses the "B^2/2" convection for magnetic pressure?

jzuhone:
[...]

Issue #3619, opened 2021-10-22
BUG: Uninitialized memory in Athena++ selection


neutrinoceros:
### Bug report

**Bug summary**

So there currently seems to be an issue with a single test that I've seen failing repeatedly these days.
Namely, it `yt/frontends/athenapp/tests/test_outputs.py::test_disk` and the failing assertion is:
https://github.com/yt-project/yt/blob/9a6def95a14c1883e74325113055ffa14fb9e48a/yt/frontends/athena_pp/tests/test_outputs.py#L31

So we can see from the recent history that the call to `total_quantity` returns random results, including

- a random float https://tests.yt-project.org/job/yt_py38_git/4106/
- another random float https://tests.yt-project.org/job/yt_py38_git/4107/
- a nan https://tests.yt-project.org/job/yt_py38_git/4109/

Note that all the examples above are from running #3489, which is clearly unrelated to the failing code, but I've seen it for other PRs as well, it's hard to find another example within the recent history because most jobs we still have logs for were failed for different reasons or cancelled.

I assume this is a parellelism bug given that `total_quantity` is `TotalQuantity` object, derived from `DerivedQuantity`, and  workers synchronisation issues seem like a plausible explanation to getting random results to an otherwise deterministic function

see
https://github.com/yt-project/yt/blob/69ada921cd0f15677611303934c140fcd93c1089/yt/data_objects/derived_quantities.py#L55

I think the breaking test isn't worth blocking PRs and consuming many hours of CI from resubmissions, so I'm going to open a with the problematic line deactivated, but it won't close this issue.

jzuhone:
This has likely been resolved by https://github.com/yt-project/yt/pull/4562.

PR #3787, opened 2022-02-07
Fix electromagnetic field unit conversions 

jzuhone:
## PR Summary

Electromagnetic unit conversions in yt are complex. The reasons for this are:

* `unyt` uses MKSA as the base set of units, but `yt` assumes Gaussian CGS units by default
* MKSA has a fundamental unit for current, Ampere, which becomes a component unit of most electromagnetic units (e.g. Coulomb, Tesla, etc.)
* Gaussian CGS units do not have a fundamental unit of current, defining all electromagnetic quantities in terms of mass, length, and time units

This is described in more detail [here](https://ytep.readthedocs.io/en/master/YTEPs/YTEP-0028.html#special-handling-for-magnetic-fields). 

Some frontends assume Gaussian CGS EM units as the code units, and a few assume MKSA EM units for the code units. However, regardless of what one assumes, it ought to be straightforward and transparent to do something like this:

```python
ds = yt.load("some_dataset", unit_system="code")
```

and expect magnetic fields to have SI dimensions instead of Gaussian CGS dimensions if the underlying code units do. It turns out that this doesn't always work correctly, as seen in Issue #3471, which was originally posted about another problem but in fact led to the discovery of this issue. 

The problem is that we need something that comes even before unit detection from the dataset, to let yt know whether or not it should expect something with dimensions of SI current in it when creating the code unit system.  There is a chicken-and-egg problem where we have to set up a code unit system before we can detect units, but only when we detect units can we know whether we need an SI current unit, which is info you need to create the code unit system in the first place. It's quite awful.

This PR is an attempt to fix the logic that handles this situation, as well as the case where we have an SI code system but request a CGS system (and vice-versa), which is currently very brittle. The easiest way to do this (though clearly not the most elegant) is for the `Dataset` object to have a new attribute (called `_use_mks_em_units` in this case) which tells it to expect that the code units will be using MKSA for EM units.

I've updated the documentation and the skeleton frontend to reflect how to ensure that SI/MKSA units are used for code units, and I've also updated the stream dataset documentation to show how to make sure this happens for in-memory datasets.

I also added some new tests and fixed some other ones. Finally, this does not fix Issue #3471--that will have to be addressed in a separate PR.

## PR Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ png.cfg
 rockstar.cfg
 yt_updater.log
 yt/frontends/artio/_artio_caller.c
+yt/frontends/gamer/cfields.c
 yt/analysis_modules/halo_finding/rockstar/rockstar_groupies.c
 yt/analysis_modules/halo_finding/rockstar/rockstar_interface.c
 yt/analysis_modules/ppv_cube/ppv_utils.c
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -29,7 +29,7 @@ answer_tests:
     - yt/frontends/athena/tests/test_outputs.py:test_stripping
 
   local_athena_pp_006:
-    - yt/frontends/athena_pp/tests/test_outputs.py:test_disk
+    #- yt/frontends/athena_pp/tests/test_outputs.py:test_disk
     - yt/frontends/athena_pp/tests/test_outputs.py:test_AM06
 
   local_chombo_005:
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -28,7 +28,7 @@ answer_tests:
     - yt/frontends/athena/tests/test_outputs.py:test_blast
     - yt/frontends/athena/tests/test_outputs.py:test_stripping
 
-  local_athena_pp_005:
+  local_athena_pp_006:
     - yt/frontends/athena_pp/tests/test_outputs.py:test_disk
     - yt/frontends/athena_pp/tests/test_outputs.py:test_AM06
 
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -28,7 +28,7 @@ answer_tests:
     - yt/frontends/athena/tests/test_outputs.py:test_blast
     - yt/frontends/athena/tests/test_outputs.py:test_stripping
 
-  local_athena_pp_004:
+  local_athena_pp_005:
     - yt/frontends/athena_pp/tests/test_outputs.py:test_disk
     - yt/frontends/athena_pp/tests/test_outputs.py:test_AM06
 
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1153,7 +1153,7 @@ class Dataset(abc.ABC):
                         "code_magnetic", self.magnetic_unit.value * 0.1 ** 0.5
                     )
         # _use_mks_em_units tells us if the code unit system needs an MKS current
-        current_mks_unit = "code_current" if self._use_mks_em_units else None
+        current_mks_unit = "code_current" if mks_system else None
         us = create_code_unit_system(
             self.unit_registry, current_mks_unit=current_mks_unit
         )
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1187,6 +1187,8 @@ class Dataset(abc.ABC):
         if unit_system == "mks":
             # 1 T = 1 kg/(A*s**2)
             self.unit_registry.add("code_magnetic", 1.0, dimensions.magnetic_field)
+            # 1 A
+            self.unit_registry.add("code_current", 1.0, dimensions.current_mks)
         else:
             # 1 gauss = 1 sqrt(g)/(sqrt(cm)*s) = 0.1**0.5 sqrt(kg)/(sqrt(m)*s)
             self.unit_registry.add(
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1325,8 +1325,6 @@ class Dataset(abc.ABC):
         self.unit_registry.modify("code_pressure", pressure_unit)
         self.unit_registry.modify("code_density", density_unit)
         self.unit_registry.modify("code_specific_energy", specific_energy_unit)
-        if hasattr(self, "current_unit"):
-            self.unit_registry.modify("code_current", self.current_unit)
         if hasattr(self, "magnetic_unit"):
             if self.magnetic_unit.units.dimensions == dimensions.magnetic_field_cgs:
                 # We have to cast this explicitly to MKS base units, otherwise
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1170,7 +1170,7 @@ class Dataset(abc.ABC):
         self.unit_registry.add("code_mass", 0.001, dimensions.mass)
         self.unit_registry.add("code_density", 1000.0, dimensions.density)
         self.unit_registry.add(
-            "code_specific_energy", 1.0, dimensions.energy / dimensions.mass
+            "code_specific_energy", 1.0e-4, dimensions.energy / dimensions.mass
         )
         self.unit_registry.add("code_time", 1.0, dimensions.time)
         if unit_system == "mks":
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1148,16 +1148,15 @@ class Dataset(abc.ABC):
                     self.unit_registry.modify(
                         "code_magnetic", self.magnetic_unit.value * 1.0e3 * 0.1**-0.5
                     )
-            else:
-                if current_mks in mag_dims:
-                    self.magnetic_unit = self.quan(
-                        self.magnetic_unit.to_value("T") * 1.0e4, "gauss"
-                    )
-                    # The following modification ensures that we get the conversion to
-                    # cgs correct
-                    self.unit_registry.modify(
-                        "code_magnetic", self.magnetic_unit.value * 1.0e-4
-                    )
+            elif current_mks in mag_dims:
+                self.magnetic_unit = self.quan(
+                    self.magnetic_unit.to_value("T") * 1.0e4, "gauss"
+                )
+                # The following modification ensures that we get the conversion to
+                # cgs correct
+                self.unit_registry.modify(
+                    "code_magnetic", self.magnetic_unit.value * 1.0e-4
+                )
         # _use_mks_em_units tells us if the code unit system needs an MKS current
         current_mks_unit = "A" if mks_system else None
         us = create_code_unit_system(
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1140,17 +1140,23 @@ class Dataset(abc.ABC):
             mag_dims = self.magnetic_unit.units.dimensions.free_symbols
             if mks_system:
                 if current_mks not in mag_dims:
-                    self.magnetic_unit = self.magnetic_unit.to("gauss").to("T")
+                    self.magnetic_unit = self.quan(
+                        self.magnetic_unit.to_value("gauss") * 1.0e-4, "T"
+                    )
                     # The following modification ensures that we get the conversion to
                     # mks correct
-                    self.unit_registry.modify("code_magnetic", self.magnetic_unit.value)
+                    self.unit_registry.modify(
+                        "code_magnetic", self.magnetic_unit.value * 1.0e3 * 0.1**-0.5
+                    )
             else:
                 if current_mks in mag_dims:
-                    self.magnetic_unit = self.magnetic_unit.to("T").to("gauss")
+                    self.magnetic_unit = self.quan(
+                        self.magnetic_unit.to_value("T") * 1.0e4, "gauss"
+                    )
                     # The following modification ensures that we get the conversion to
                     # cgs correct
                     self.unit_registry.modify(
-                        "code_magnetic", self.magnetic_unit.value * 0.1 ** 0.5
+                        "code_magnetic", self.magnetic_unit.value * 1.0e-4
                     )
         # _use_mks_em_units tells us if the code unit system needs an MKS current
         current_mks_unit = "code_current" if mks_system else None
@@ -1319,6 +1325,17 @@ class Dataset(abc.ABC):
         self.unit_registry.modify("code_pressure", pressure_unit)
         self.unit_registry.modify("code_density", density_unit)
         self.unit_registry.modify("code_specific_energy", specific_energy_unit)
+        if hasattr(self, "current_unit"):
+            self.unit_registry.modify("code_current", self.current_unit)
+        if hasattr(self, "magnetic_unit"):
+            if self.magnetic_unit.units.dimensions == dimensions.magnetic_field_cgs:
+                # We have to cast this explicitly to MKS base units, otherwise
+                # unyt will convert it automatically to Tesla
+                self.unit_registry.modify(
+                    "code_magnetic", self.magnetic_unit.to("sqrt(kg)/(sqrt(m)*s)")
+                )
+            else:
+                self.unit_registry.modify("code_magnetic", self.magnetic_unit)
         # domain_width does not yet exist
         if self.domain_left_edge is not None and self.domain_right_edge is not None:
             DW = self.arr(self.domain_right_edge - self.domain_left_edge, "code_length")
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1159,7 +1159,7 @@ class Dataset(abc.ABC):
                         "code_magnetic", self.magnetic_unit.value * 1.0e-4
                     )
         # _use_mks_em_units tells us if the code unit system needs an MKS current
-        current_mks_unit = "code_current" if mks_system else None
+        current_mks_unit = "A" if mks_system else None
         us = create_code_unit_system(
             self.unit_registry, current_mks_unit=current_mks_unit
         )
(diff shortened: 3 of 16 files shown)


yt/loaders.py
```diff
@@ -735,6 +740,7 @@ def parse_unit(unit, dimension):
         geometry=geometry,
         unit_system=unit_system,
         default_species_fields=default_species_fields,
+        use_mks_em_units=use_mks_em_units,
```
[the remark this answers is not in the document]
jzuhone: Sorry--confused. It isn't keyword-only?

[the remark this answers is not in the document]
neutrinoceros: Ah, this comment was meant for the definition of `load_hexahedral_mesh`

yt/loaders.py
```diff
@@ -754,6 +760,7 @@ def load_hexahedral_mesh(
     periodicity=(True, True, True),
     geometry="cartesian",
     unit_system="cgs",
+    use_mks_em_units=False,
```
neutrinoceros, line 763: So this is what I meant to suggest early (re: keyword-only arguments)
```suggestion
    *,
    use_mks_em_units=False,
```

yt/loaders.py
```diff
@@ -579,6 +583,7 @@ def load_particles(
     unit_system="cgs",
     data_source=None,
     default_species_fields=None,
+    use_mks_em_units=False,
```
neutrinoceros, line 586: While I'm at it
```suggestion
    *,
    use_mks_em_units=False,
```

yt/loaders.py
```diff
@@ -361,6 +363,7 @@ def load_amr_grids(
     refine_by=2,
     unit_system="cgs",
     default_species_fields=None,
+    use_mks_em_units=False,
```
neutrinoceros, line 366:
```suggestion
    *,
    use_mks_em_units=False,
```

yt/loaders.py
```diff
@@ -162,6 +162,7 @@ def load_uniform_grid(
     geometry="cartesian",
     unit_system="cgs",
     default_species_fields=None,
+    use_mks_em_units=False,
```
neutrinoceros, line 165:
```suggestion
    *,
    use_mks_em_units=False,
```

yt/frontends/stream/data_structures.py
```diff
@@ -450,13 +454,15 @@ def __init__(
         geometry="cartesian",
         unit_system="cgs",
         default_species_fields=None,
```
neutrinoceros, line 456:
```suggestion
        *,
        default_species_fields=None,
```

yt/frontends/stream/data_structures.py
```diff
@@ -274,14 +274,18 @@ def __init__(
...
+        use_mks_em_units=False,
```
neutrinoceros, line 277:
```suggestion
        *,
        use_mks_em_units=False,
```

neutrinoceros:
Aren't we supposed to wait a first approval to bump answer tests ?
Anyway I see that the one you bumped is failing again, but it's also the same flaky test that #3619 is about. If it was migrated to pytest I could propose a solution to mark it as flaky so pytest could run it more than once before degiding it failed but since it's still stuck on nose I'm really tempted to suggest we just deactivate it. That is, unless you or @matthewturk think there is a better way, of course :)

yt/frontends/flash/tests/test_outputs.py
```diff
@@ -54,6 +57,23 @@ def test_units_override():
     units_override_check(sloshing)
 
 
+@disable_dataset_cache
```
neutrinoceros, line 60: I didn't know about this decorator but after reading it, it seems to me that it's currently broken and that the previous state is *never* restored (see `disable_dataset_cache` therein). I've opened #3796 to fix this.

yt/frontends/stream/data_structures.py
```diff
@@ -291,6 +294,25 @@ def __init__(
             default_species_fields=default_species_fields,
         )
 
+    def _check_for_mks_em_units(self):
```
neutrinoceros, line 297: Doesn't hurt
```suggestion
    def _check_for_mks_em_units(self) -> bool:
```

yt/frontends/stream/data_structures.py
```diff
@@ -291,6 +294,25 @@ def __init__(
...
+        from yt.units.dimensions import current_mks
+
+        # We assume CGS EM units by default, unless
+        # magnetic_unit is explicitly defined within MKSA
+        magnetic_unit = self.stream_handler.code_units[-1]
```
neutrinoceros, line 302: I'm not too optimistic here but is there any way to make this less fragile ? If we ever append another unit to `self.stream_handler.code_units`, or change their order, this would break, possibly silently.

yt/frontends/_skeleton/data_structures.py
```diff
@@ -114,6 +115,10 @@ def _set_code_unit_attributes(self):
         # These can also be set:
         # self.velocity_unit = self.quan(1.0, "cm/s")
         # self.magnetic_unit = self.quan(1.0, "gauss")
+        #
+        # If your frontend uses SI EM units, set magnetic units like this
+        # instead:
+        # self.magnetic_unit = self.quan(1.0, "T")
```
neutrinoceros, line 121: This sounds very much like a job for the `default_units` class attribute ! :)

yt/data_objects/static_output.py
```diff
@@ -1165,25 +1185,43 @@ def _create_unit_registry(self, unit_system):
         # yt assumes a CGS unit system by default (for back compat reasons).
         # Since unyt is MKS by default we specify the MKS values of the base
         # units in the CGS system. So, for length, 1 cm = .01 m. And so on.
+        # Note that the values associated with the code units here will be
+        # modified once we actually determine what the code units are from
+        # the dataset
         self.unit_registry = UnitRegistry(unit_system=unit_system)
+        # 1 cm = 0.01 m
         self.unit_registry.add("code_length", 0.01, dimensions.length)
+        # 1 g = 0.001 kg
```
neutrinoceros, line 1194: This is actually helpful info, thank you for adding this series of comments

.gitignore
```diff
@@ -7,6 +7,7 @@ png.cfg
 rockstar.cfg
 yt_updater.log
 yt/frontends/artio/_artio_caller.c
+yt/frontends/gamer/cfields.c
```
neutrinoceros, line 10: good catch !

neutrinoceros (commented):
Okay here's a first batch of comments and questions. I have intentionally not reviewed the logic in `Dataset._assign_unit_system` yet, but I will later !

So I've checked-out the tests without the patches and I saw that only the ones from `yt/fields/tests/test_magnetic_fields.py` fail on main. Just wanted to make sure this is intentional and that the other tests your adding are intended as regression safe-guards (as opposed to being broken, as in failing to check what you wanted to check).

Also while reading this I was reminded of the existence of the (relatively new and rarely used) `Dataset.default_units` class attribute, which is intended to be subclassed. I presume the reason you didn't use it here is because you didn't know about it (can't blame you, I think only one or two frontends currently subclass it) ?

yt/frontends/stream/data_structures.py
jzuhone: @neutrinoceros it's not a bad idea (I thought this myself), but that's a much bigger job than this PR is trying to solve, as it would touch quite a bit of code in the stream frontend.

yt/units/tests/test_magnetic_code_units.py
```diff
@@ -49,3 +50,25 @@ def test_magnetic_code_units():
     mucu = ds4.magnetic_unit.to("code_magnetic")
     assert_allclose(mucu.value, 1.0)
     assert str(mucu.units) == "code_magnetic"
+
+    ds5 = load_uniform_grid(
```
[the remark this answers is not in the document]
jzuhone: What this test is designed to do is to cover a whole bunch of corner cases of the same phenomenon: making sure that you can load magnetic fields with any units and then convert to any other kind of magnetic field units. If one of these sub-tests is broke it's likely that something is wrong that may touch the whole machinery which they are all related to, so I would advise keeping them together.

yt/frontends/stream/data_structures.py
neutrinoceros: okay so I've looked into this some more and found that the code breaks (badly) in every scenario I could think of
- appending a new unit
- changing their order
- dropping the magnetic unit

So it's still not ideal, but at least I'm now confident that this will not break silently if we ever change something around it.

jzuhone: We should definitely change it--I think it deserves a new PR on its own.

jzuhone:
@neutrinoceros:

> Aren't we supposed to wait a first approval to bump answer tests ?

I wasn't aware of this, but it's been a while since I've done it.

> Anyway I see that the one you bumped is failing again, but it's also the same flaky test that https://github.com/yt-project/yt/issues/3619 is about. If it was migrated to pytest I could propose a solution to mark it as flaky so pytest could run it more than once before deciding it failed but since it's still stuck on nose I'm really tempted to suggest we just deactivate it. That is, unless you or @matthewturk think there is a better way, of course :)

This dataset and the test are related to a very botched attempt at trying to handle logarithmic grids with a variable `dds` using the semi-structured mesh machinery. It works for this dataset (but not always apparently) but often did not work for others. I'd be in favor of deactivating this test because frankly we should support datasets like this using the variable `dds` work that seems to be moving along.

yt/units/tests/test_magnetic_code_units.py
```diff
@@ -21,16 +21,17 @@ def test_magnetic_code_units():
     assert_allclose(mucu.value, 1.0)
     assert str(mucu.units) == "code_magnetic"
 
-    ds2 = load_uniform_grid(data, ddims, magnetic_unit=(1.0, "T"), unit_system="cgs")
-
+    ds2 = load_uniform_grid(data, ddims, magnetic_unit=(1.0, "T"),
+                            unit_system="cgs")
```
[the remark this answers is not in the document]
jzuhone: I edited this file in a later commit, and it's proving quite messy to back this particular change out, so I'll just fix the lines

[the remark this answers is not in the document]
neutrinoceros: I was suggesting you could just squash the whole history of this branch into a single commit and leave no trace of the back-and-forth edits :)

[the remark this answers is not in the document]
jzuhone: @neutrinoceros I admit I am not a big fan of that approach--I squashed a few commits but I would prefer to leave things as-is.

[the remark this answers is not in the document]
neutrinoceros: that's fine, it's not a big deal here

yt/data_objects/static_output.py
```diff
@@ -242,10 +243,10 @@ def __init__(
         self.no_cgs_equiv_length = False
 
         if unit_system == "code":
-            # create a fake MKS unit system which we will override later to
+            # create a fake unit system which we will override later to
             # avoid chicken/egg issue of the unit registry needing a unit system
             # but code units need a unit registry to define the code units on
-            used_unit_system = "mks"
+            used_unit_system = "mks" if self._use_mks_em_units else "cgs"
```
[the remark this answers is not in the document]
jzuhone: So in theory it would be nice to use `default_units` for this instead, but there is the issue that the stream frontend would still have to change `default_units` if we used it to check for MKS EM units. I'm supposing that wasn't the intent--that it should not be edited on a per-object basis?

yt/frontends/athena_pp/tests/test_outputs.py
```diff
@@ -1,20 +1,13 @@
...
-)
+from yt.testing import assert_equal, requires_file, units_override_check
 from yt.utilities.answer_testing.framework import (
-    GenericArrayTest,
     data_dir_load,
     requires_ds,
     small_patch_amr,
 )
 
+"""
```
neutrinoceros, line 10: please add a comment with the issue's url so the reason why it's commented is apparent

tests/tests.yaml
```diff
@@ -28,7 +28,7 @@ answer_tests:
     - yt/frontends/athena/tests/test_outputs.py:test_blast
     - yt/frontends/athena/tests/test_outputs.py:test_stripping
 
-  local_athena_pp_004:
+  local_athena_pp_006:
     - yt/frontends/athena_pp/tests/test_outputs.py:test_disk
```
neutrinoceros, line 32: I think we actually need to comment this line out, see
https://tests.yt-project.org/job/yt_py38_git/4630/testReport/junit/nose.failure/Failure/runTest/

yt/data_objects/static_output.py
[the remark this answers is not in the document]
jzuhone: @neutrinoceros any thoughts here?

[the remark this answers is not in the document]
neutrinoceros: Sorry I missed that.

`default_units` is indeed intended as a constant class attribute, but its job is just to provide sane defaults, instances are free to use `units_override` to replace them partially or totally.
Couldn't we migrate this specialised behaviour in the stream frontend ?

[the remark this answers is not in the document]
jzuhone: @neutrinoceros,

So `default_units` is for `units_override` only, correct? `units_override` was originally designed to provide units for datasets that weren't self-describing and didn't have any unit information in them--Athena datasets are notorious for this, for example. Because yt doesn't really work without assuming _some_ units, `units_override` provided sane defaults, so that you could at least do something with the dataset. We enabled `units_override` for all frontends, but it was always considered a solution of last resort, and if your dataset was self-describing you shouldn't even need it unless you were doing something weird.

In this use case for this PR, we need something that comes even before unit detection from the dataset, to let yt know it should expect something with dimensions of SI current in it when creating the code unit system. This has been a very tricky thing to manage  in yt while still allowing the end-user to switch to a different unit system, because there is a chicken-and-egg problem where we have to set up a code unit system before we can detect units, but only when we detect units can we know whether we need an SI current unit, which is info you need to create the code unit system in the first place. It's quite awful.

`default_units` seems like something quite different.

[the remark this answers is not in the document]
neutrinoceros: > So default_units is for units_override only, correct?

yes that was the original intention.

Ok I think I understand this problem better now. It's unfortunate that complexity is going up but I reckon it may be the only way right now. Thanks a lot for going into the details, I appreciate it !

neutrinoceros:
I will finish my review tomorrow. I anticipate that I'll probably just approve with no further comments, but given the size of this PR, I would like a second review even if it's a bugfix, so I won't merge until then.

yt/data_objects/static_output.py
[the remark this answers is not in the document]
jzuhone: NP--I'm going to copy the explanation I gave into the main text so the context is there.

yt/data_objects/static_output.py
```diff
@@ -1119,27 +1120,46 @@ def relative_refinement(self, l0, l1):
...
-                self.unit_registry.modify("code_magnetic", self.magnetic_unit.value)
+        # we need to determine if the requested unit system
+        # is mks-like: i.e., it has a current with the same
+        # dimensions as amperes.
+        mks_system = False
+        if unit_system != "code":
+            # if the unit system is known, we can check if it
+            # has a "current_mks" unit
+            us = unit_system_registry[str(unit_system).lower()]
+            mks_system = us.base_units[current_mks] is not None
```
neutrinoceros, line 1131: isn't `current_mks` undefined at this point ?

yt/data_objects/static_output.py
```diff
@@ -1119,27 +1120,46 @@ def relative_refinement(self, l0, l1):
...
-                # system as an MKS-like system
-                if current_mks in self.magnetic_unit.units.dimensions.free_symbols:
-                    self.magnetic_unit = self.magnetic_unit.to("T").to("gauss")
-                # The following modification ensures that we get the conversion to
-                # cgs correct
-                self.unit_registry.modify(
-                    "code_magnetic", self.magnetic_unit.value * 0.1**0.5
-                )
-
+                if current_mks in mag_dims:
```
neutrinoceros, line 1152:
```python
else:
    if ...:
        ...
```
Looks like this could be simplified into
```python
elif:
    ...
```

neutrinoceros (commented):
two more comments.

yt/data_objects/static_output.py
jzuhone: Agreed.

yt/data_objects/static_output.py
neutrinoceros: Oh right, I think I was confusing it with `current_mks_units` while reading the diff.
I was also surprised that flake8 wouldn't catch an undef var.

stonnes:
I took a look, and should not be an official reviewer, but I have a couple questions:
Does every frontend that uses MKS EM units now need to have the:
_use_mks_em_units = True
line hard-coded in the initial frontend definition?  And if so, will this break all the frontends that need it or did you put it in already?  Can the user set this in the yt.load call?  Not *at all* sure that this would actually be a good idea!

neutrinoceros:
Sounds like a good opportunity to test my own understanding: I think it's always the responsibility of frontends to define this class attribute. I assume there may be some existing frontends for which it _should_ be set to True but neither John or I are familiar with them so there would be a need for follow up fixes there ?

jzuhone:
@stonnes @neutrinoceros I'm going to revisit this--my understanding of what the code was doing got deeper the longer I looked at this, and I might have a way to fix the chicken-and-egg problem that will make for fewer changes. I'll report back later. I'll switch this to draft for now.

jzuhone:
Closed in favor of PR #3805.

PR #3792, opened 2022-02-08
DOC: don't pin a specific yt version in a badge that appears on PyPI

neutrinoceros:
## PR Summary
We forgot to update this version string for yt 4.0.2, now the badge displays support for Python 3.6 to 3.9 (it should be 3.6 to 3.10). On second thought I don't think it's worth pinning there because ultimately it's just one more mistake we can make in the release process, so I'm going for the easy fix here that can be safely backported for 4.0.3 as is

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # The yt Project
 
 [![PyPI](https://img.shields.io/pypi/v/yt)](https://pypi.org/project/yt)
-[![Supported Python Versions](https://img.shields.io/pypi/pyversions/yt/4.0.0)](https://pypi.org/project/yt/)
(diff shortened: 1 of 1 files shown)


matthewturk (approved):
One thing we could do is if we used a bump2version type system is have it replace them when it bumps.  But I'm fine with just not pinning.

neutrinoceros:
I'm not convinced yet that `bump2version` is worth the technical debt that comes along with it. My attempts to use it in much smaller projects failled miserably.

Matt, how can we merge this ? CI is still showing as orange only because of the auto-merge setup, and it looks to me that it won't let us merge until the "answer" job passes, but it's not triggered because no code was touched here

Issue #3554, opened 2021-10-12
BUG: matplotlib 3.5 breaks colorbars


neutrinoceros:
### Bug report

**Bug summary**

Matplotlib has seen a lot of internal changes for colorbars between 3.4 and 3.5, some of which have minor yet noticeable consequences for yt

This is seen in daily jobs on Jenkins (latest_deps) since we allowed pre-releases to install
for instance the test `yt.geometry.coordinates.tests.test_cylindrical_coordinates.GenericImage_AMRGridData_all_create_image_None_None`
(see here https://tests.yt-project.org/job/yt_py38_latest_deps/389/testReport/junit/yt.geometry.coordinates.tests/test_cylindrical_coordinates/GenericImage_AMRGridData_all_create_image_None_None/)

**Code for reproduction**

**Actual outcome**

![tmprwnn5eiu](https://user-images.githubusercontent.com/14075922/136946868-7a42daa4-6328-4ea4-b3b2-9f42d8818010.png)

**Expected outcome**
![tmphtg06g_w](https://user-images.githubusercontent.com/14075922/136946793-d353ca53-d19e-4a60-a9a4-dbc4eac71bce.png)

**Version Information**
  * Operating System:
  * Python Version:
  * yt version: main
  * Other Libraries (if applicable): matplotlib 3.5rc1

This bug is akin to the one I solved for symlog norms already https://github.com/yt-project/yt/pull/3536#issuecomment-932701124

PR #3793, opened 2022-02-08
BUG: fix heuristic to switch from log scale to symlog scale

### LINKED ISSUE #3791: BUG: a "hole" in cylindrical pixelization
### Bug report

**Bug summary**

Making a slice plot in the poloidal plane for a dataset with a simulation box going (in the colatitudinal direction) from 0 (north pole) to PI/2 (equator), one can see that left and right edges are not being treated the same; pixels are coloured up to the equator, but there's a white region around the polar axis, where I think there shouldn't be.
Let's make it very obvious with an extremely coarse grid.

**Code for reproduction**

```python
import numpy as np
import yt

shape = (32, 16, 1)
a = np.linspace(0, 1, 16)
b = np.ones((32, 16))
c = np.reshape(a*b, shape)
data = {("gas", "density"): c}

theta_min = 0
theta_max = np.pi/2

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.5, 1.0], [theta_min, theta_max], [-0.1, +0.1]]),
    geometry="spherical",
)

yt.SlicePlot(ds, "phi", "density").save("/tmp/branch.png")
```

**Actual outcome**

branch

I had a look at the pixelizer routine earlier today and couldn't find the bug.
Maybe the actual bug is in the `load_uniform_grid` grid, but that's what I've been using for lack of an actual dataset with the desired properties.

neutrinoceros:
## PR Summary
Fix #3791

Here's a simpler demo
```python
import numpy as np
import yt

shape = (32, 16, 1)
a = np.linspace(0, 1, 16)
b = np.ones((32, 16))
c = np.reshape(a*b, shape)
data = {("gas", "density"): c}

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.0, 5.0], [0, 1], [-0.1, +0.1]]),
)

p.save("/tmp/branch.png")
```

on main
main

this branch
branch

While the original issue was only about 2D plots, I'm also fixing a similar bug in line plots that I found on the way

### DIFF
--- a/yt/visualization/line_plot.py
+++ b/yt/visualization/line_plot.py
@@ -335,7 +335,7 @@ class LinePlot(PlotContainer):
 
                 # apply log transforms if requested
                 if self._field_transform[field] != linear_transform:
-                    if (y < 0).any():
+                    if (y <= 0).any():
                         plot.axes.set_yscale("symlog")
                     else:
                         plot.axes.set_yscale("log")
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -1056,7 +1056,7 @@ class PWViewerMPL(PlotWindow):
                     use_symlog = True
                 elif not np.any(np.isfinite(image)):
                     msg = f"Plotting {f}: All values = NaN."
-                elif np.nanmax(image) > 0.0 and np.nanmin(image) < 0:
+                elif np.nanmax(image) > 0.0 and np.nanmin(image) <= 0:
                     msg = (
                         f"Plotting {f}: Both positive and negative values. "
                         f"Min = {np.nanmin(image)}, Max = {np.nanmax(image)}."


neutrinoceros:
Not sure what's happening in the two tests that failed. If I can't figure it out quickly I'll just remove this from the milestone

neutrinoceros:
Found that the failing tests were a consequence of https://github.com/yt-project/yt/pull/2517, and I discovered that the hack introduced back then isn't necessary after MPL 3.2 so I'm making it conditional. Now I'm expecting everything to stay green

neutrinoceros:
I've revised my approach to intentionally change some reference images. We have a couple tests (the one that have been failing) for which we plot some "binary" noise (i.e. all values are 0 or 1). The current gold standard looks like this
tmpkkshi4qk
which is not actually what we want: we see that all values plotted are equal (1), but we're masking zeros completely.
With the latest commit I'm getting this locally instead
3793_noise0_log_Slice_z_noise0
Now I'm waiting to see if my changes accidentally affect any other cases, and wether it makes them better or worse.

neutrinoceros:
This is now in the intented state. The only failing image tests are the ones I intend to update. I'll wait for an approval to bump them.

yt/visualization/base_plot_types.py
```diff
@@ -298,7 +307,7 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
                     10
                     ** np.arange(
                         np.rint(np.log10(cblinthresh)),
-                        np.ceil(np.log10(zmax)),
+                        np.ceil(np.log10(1.1 * zmax)),
```
neutrinoceros, line 310: The idea here is to intentionally overshoot zmax with the (excluded) upper boundary so that zmax can be used as a major tick location if it is itself an exact power of 10.

yt/visualization/base_plot_types.py
```diff
@@ -334,10 +343,12 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
                         10
                         ** np.arange(
                             np.rint(np.log10(cblinthresh)),
-                            np.ceil(np.log10(zmax)),
+                            np.ceil(np.log10(1.1 * zmax)),
                         )
                     )
                 )
+            if yticks[-1] > zmax:
+                yticks.pop()
```
neutrinoceros, line 351: necessary cleanup with matplotlib 3.5 where having a ticks that's above the actual maximal value creates a white region at the top end of the colorbar (see https://github.com/yt-project/yt/issues/3554)

yt/visualization/base_plot_types.py
matthewturk: Makes sense.

yt/visualization/base_plot_types.py
```diff
@@ -334,10 +343,12 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
                         10
```
matthewturk, line 343: Just so I understand, would it be possible to do this with `np.logspace`?  I do not think we should in this PR, for continuity sake, however.

matthewturk (approved):

yt/visualization/base_plot_types.py
neutrinoceros: yes it should be possible to refactor this with np.logspace. I also agree that it's preferable to keep diffs minimal for a bugfix PR

neutrinoceros:
Seems like the Jenkins job got killed because of network issues ?
```
hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@37a79109:JNLP4-connect connection from dxl1.ncsa.illinois.edu/141.142.193.70:39688": Remote call on JNLP4-connect connection from dxl1.ncsa.illinois.edu/141.142.193.70:39688 failed. The channel is closing down or has closed down
```
Let's try this again: @yt-fido test this please

matthewturk (approved):

jzuhone (approved):

PR #3801, opened 2022-02-10
BUG: fix error messages in RAMSESDataset validation

### LINKED ISSUE #3800: BUG: RAMSESDataset fails to load
### Bug report

**Bug summary**

This was reported on the mailing list https://mail.python.org/archives/list/yt-users@python.org/thread/5CMIDJGXF452PX6IGBGAGMAE3RKEQYD3/

So I think there are at least two different problems there:
- the error message is malformed 
- it is not informative enough (there are a couple reasons why `RAMSESDataset._is_valid` can be False, but here we're swallowing everything)
- I guess the data should actually load, but there is not enough information in the original report to assess what's actually going on

I'll link this issue to the person who reported, so we can work together on the last point. Meanwhile, I can fix the first two on my own

neutrinoceros:
## PR Summary
This adresses the surface problem reported in #3800, i.e., I'm demystifying errors raised by `RAMSESDataset` at validation time

### Example 1

```python
from yt.frontends.ramses.api import RAMSESDataset

RAMSESDataset("not_a_file")
```

On main

```python-traceback
ValueError: ('Invalid filename found when building a RAMSESDataset object: %s', 'not_a_file')
```

This branch
```python-traceback
FileNotFoundError: No such file or directory '/Users/robcleme/dev/yt-project/yt/not_a_file'
```

### Example 2

```python
import os
from yt.frontends.ramses.api import RAMSESDataset

RAMSESDataset(os.path.expanduser("~"))
```

On main

```python-traceback
ValueError: ('Invalid filename found when building a RAMSESDataset object: %s', '/Users/robcleme')
```

This branch
```python-traceback
ValueError: Could not determine output directory from '/Users/robcleme'
Expected a directory name of form '(output|group)_(\\d{5})'
```

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -19,6 +19,7 @@ from yt.utilities.on_demand_imports import _f90nml as f90nml
 from yt.utilities.physical_constants import kb, mp
 
 from .definitions import (
+    OUTPUT_DIR_EXP,
     OUTPUT_DIR_RE,
     STANDARD_FILE_RE,
     field_aliases,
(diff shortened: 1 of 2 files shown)


matthewturk (approved):
Good improvement.

neutrinoceros:
@matthewturk sorry I just forced pushed again because I realised I made a mistake that could break `yt.load(None)`

yt/frontends/ramses/data_structures.py
```diff
@@ -67,9 +67,32 @@ def __init__(self, filename):
             self.group_name = None
         self.info_fname = info_fname
 
+    def validate(self) -> None:
+        # raise a TypeError if self.original_filename is not a valid path
+        filename: str = os.path.expanduser(self.original_filename)
```
neutrinoceros, line 72: note: using `os.path.expanduser` is important because `os.path.exists("~")` is always `False`

neutrinoceros:
This took a surprising amount of iterations for me to understand enough of what the frontend actually expected so I can provide more useful error messages. So I think the problem Andrew actually had was that his data isn't stored in a dir that's named `output_xxxxx` or `group_xxxxxx`. I've now made this expectation much more apparent in the error message :

```python-traceback
ValueError: Could not determine output directory from '/Users/robcleme'
Expected a directory name of form '(output|group)_(\\d{5})'
```

I'll wait on his reply to see if he's indeed able to load his data after rearranging the files, but I expect that this is now sufficient to close the issue.

cphyc (approved):
LGTM

neutrinoceros:
I don't think so, why ?

neutrinoceros:
(If it turns out that the issue Andrew had was more complex I can always reopen the ticket)

PR #3803, opened 2022-02-10
BUG: fixup POSIX-only path handling that can easily be made cross-platform

### LINKED ISSUE #3802: BUG: many path-handling functions are POSIX-only
### Bug report

**Bug summary**

just grep
```
git grep '+ "/"'
```
easily detects a handful of places where file system paths are constructed in a POSIX only fashion
```
tests/report_failed_answers.py:    upload_url = base_url + "/" + os.path.basename(filename)
yt/frontends/boxlib/data_structures.py:        header_filename = self.ds.output_dir + "/" + directory_name + "/Header"
yt/frontends/boxlib/data_structures.py:        header_file = level_file + "/" + field + "_H"
yt/frontends/open_pmd/data_structures.py:                axis = list(f[bp + pp + "/" + spec + "/position"].keys())[0]
yt/frontends/open_pmd/data_structures.py:                pos = f[bp + pp + "/" + spec + "/position/" + axis]
yt/frontends/open_pmd/data_structures.py:                patch = f[bp + pp + "/" + spec[0] + "/particlePatches"]
yt/frontends/open_pmd/io.py:            pds = self._handle[self.base_path + self.particles_path + "/" + ptype]
yt/frontends/owls/fields.py:            fname = data_dir + "/" + data_file
yt/utilities/command_line.py:        r = requests.put(upload_url + "/" + os.path.basename(args.file), data=fs)
yt/utilities/grid_data_format/conversion/conversion_athena.py:        self.source_dir = source_dir + "/"
yt/visualization/tests/test_save.py:    save_path = os.path.join(tmp_path / "out") + "/"
```
Though hard to detect systematically (I don't know of a tool to do that), it helps that `black` normalises string delimiters.
These should all be trivial to fix using `os.path` functions

neutrinoceros:
## PR Summary
fix #3802

In many cases, using pathlib instead of os.path would help readability but it would also create a bigger diff and would take more time to implement and review, so I'm keeping a conservative approach here.

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -576,7 +576,7 @@ class BoxlibHierarchy(GridIndex):
         self.io = io_registry[self.dataset_type](self.dataset)
 
     def _determine_particle_output_type(self, directory_name):
-        header_filename = self.ds.output_dir + "/" + directory_name + "/Header"
+        header_filename = os.path.join(self.ds.output_dir, directory_name, "Header")
         with open(header_filename) as f:
             version_string = f.readline().strip()
             if version_string.startswith("Version_Two"):
@@ -1305,7 +1305,7 @@ def _guess_pcast(vals):
 
 
 def _read_raw_field_names(raw_file):
-    header_files = glob.glob(raw_file + "*_H")
+    header_files = glob.glob(os.path.join(raw_file, "*_H"))
     return [hf.split(os.sep)[-1][:-2] for hf in header_files]
 
 
@@ -1325,7 +1325,7 @@ def _get_active_dimensions(box):
 
 
 def _read_header(raw_file, field):
-    level_files = glob.glob(raw_file + "Level_*")
+    level_files = glob.glob(os.path.join(raw_file, "Level_*"))
     level_files.sort()
 
     all_boxes = []
@@ -1333,7 +1333,7 @@ def _read_header(raw_file, field):
     all_offsets = []
 
     for level_file in level_files:
-        header_file = level_file + "/" + field + "_H"
+        header_file = os.path.join(level_file, field + "_H")
         with open(header_file) as f:
 
             f.readline()  # version
@@ -1437,7 +1437,7 @@ class WarpXHierarchy(BoxlibHierarchy):
             self._read_particles(ptype, is_checkpoint)
 
         # Additional WarpX particle information (used to set up species)
-        self.warpx_header = WarpXHeader(self.ds.output_dir + "/WarpXHeader")
+        self.warpx_header = WarpXHeader(os.path.join(self.ds.output_dir, "WarpXHeader"))
 
         for key, val in self.warpx_header.data.items():
             if key.startswith("species_"):
@@ -1451,8 +1451,8 @@ class WarpXHierarchy(BoxlibHierarchy):
         super()._detect_output_fields()
 
         # now detect the optional, non-cell-centered fields
-        self.raw_file = self.ds.output_dir + "/raw_fields/"
-        self.raw_fields = _read_raw_field_names(self.raw_file + "Level_0/")
+        self.raw_file = os.path.join(self.ds.output_dir, "raw_fields")
+        self.raw_fields = _read_raw_field_names(os.path.join(self.raw_file, "Level_0"))
         self.field_list += [("raw", f) for f in self.raw_fields]
         self.raw_field_map = {}
         self.ds.nodal_flags = {}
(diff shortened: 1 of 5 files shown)


neutrinoceros:
The good news is that a lot of the code I'm changing (mostly in frontends) is exercised in tests so I can easily spot them and reiterate !

yt/frontends/open_pmd/data_structures.py
```diff
@@ -96,8 +96,8 @@ def _get_particle_type_counts(self):
                     spec = list(f[bp + pp].keys())[0]
                 else:
                     spec = ptype
-                axis = list(f[bp + pp + "/" + spec + "/position"].keys())[0]
-                pos = f[bp + pp + "/" + spec + "/position/" + axis]
+                axis = list(f[os.path.join(bp + pp, spec, "position")].keys())[0]
+                pos = f[os.path.join(bp + pp, spec, "position", axis)]
```
cphyc, line 100: Shouldn't it be the following instead?
```suggestion
                pos = f[os.path.join(bp, pp, spec, "position", axis)]
```

yt/frontends/open_pmd/io.py
```diff
@@ -32,7 +33,9 @@ def _fill_cache(self, ptype, index=0, offset=None):
         """
         if str((ptype, index, offset)) not in self._cached_ptype:
             self._cached_ptype = str((ptype, index, offset))
-            pds = self._handle[self.base_path + self.particles_path + "/" + ptype]
+            pds = self._handle[
+                os.path.join(self.base_path + self.particles_path, ptype)
```
cphyc, line 37: Same as above with bp/pp.

yt/frontends/open_pmd/data_structures.py
neutrinoceros: Good catch. Indeed `self.base_path` is also currently hardcoded in a POSIX-only fashion, I'll fix that too

neutrinoceros: actually I'm not sure what to do here. These path components are also used as keys for `h5py.File` objects, and I can't find a clear reference for how these behave on windows. I think `"/"` should still work there so I don't actually need to edit this particular frontend.

yt/frontends/boxlib/data_structures.py
```diff
@@ -1305,7 +1305,7 @@ def _guess_pcast(vals):
 
 
 def _read_raw_field_names(raw_file):
-    header_files = glob.glob(raw_file + "*_H")
+    header_files = glob.glob(os.path.join(raw_file, "*_H"))
```
matthewturk, line 1308: I don't think this is necessarily correct, is it?  It *may* be correct, but I think we need to verify first that `*_H` is not a suffix applied to files, since we don't use a path separator in the old version.

matthewturk (approved):
I think other than ensuring and verifying that `raw_file` in boxlib always ends with a path separator, this is good to go.

yt/frontends/boxlib/data_structures.py
```diff
@@ -1325,15 +1325,15 @@ def _get_active_dimensions(box):
 
 
 def _read_header(raw_file, field):
-    level_files = glob.glob(raw_file + "Level_*")
+    level_files = glob.glob(os.path.join(raw_file, "Level_*"))
```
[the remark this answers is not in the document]
neutrinoceros: What kind of verification do you need ?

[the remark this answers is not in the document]
matthewturk: I think you just telling me where we make sure it has a `/` in the `raw_file` definition is good.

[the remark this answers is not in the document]
neutrinoceros: well that's actually the other way around.
On main it's defined with a trailing `/`

https://github.com/yt-project/yt/blob/f719813a86f424b14f478fa9c6f7c9ab362c3d15/yt/frontends/boxlib/data_structures.py#L1454

So in this branch I've tracked down every places where this attribute was used to make sure it wasn't needed.
This attribute isn't "hidden" but I doubt that it's meant to be used from the user side. If you want I can add a trailing `os.sep` to make sure it stays compatible for anyone relying on the old def.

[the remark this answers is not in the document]
matthewturk: @neutrinoceros no this is exactly right, this is where we ensured it *before*, which makes your transformations here definitely valid.

neutrinoceros:
I'll take care of backporting this manually

neutrinoceros:
well actually the conflict is big enough that I'm not confident it's worth backporting at all, so let's keep that for the 4.1 release, since nobody asked for it anyway.

matthewturk:
Sounds good to me.

PR #3789, opened 2022-02-07
Manual backports to yt-4.0.x (blockers and bugfixes)

neutrinoceros:
## PR Summary
In support to the newly created 4.0.3 milestone https://github.com/yt-project/yt/milestone/19
This bundles manual backports for recent CI blockers
- #3770
- #3776 
- #3783

as well as a couple bugfix PRs that were merged since 4.0.2 was frozen
- #3780
- #3782 
- #3739 
- #3755
- #3786
- #3805

Also including #3758 which is supposed to be used to ease the release process itself.

PR #3805, opened 2022-02-12
Fix electromagnetic unit conversions (second try)

jzuhone:
## PR Summary

(this is an attempt to do better than PR #3787)

Electromagnetic unit conversions in yt are complex. The reasons for this are:

* `unyt` uses MKSA as the base set of units, but `yt` assumes Gaussian CGS units by default
* MKSA has a fundamental unit for current, Ampere, which becomes a component unit of most electromagnetic units (e.g. Coulomb, Tesla, etc.)
* Gaussian CGS units do not have a fundamental unit of current, defining all electromagnetic quantities in terms of mass, length, and time units

This is described in more detail [here](https://ytep.readthedocs.io/en/master/YTEPs/YTEP-0028.html#special-handling-for-magnetic-fields). 

Some frontends assume Gaussian CGS EM units as the code units, and a few assume MKSA EM units for the code units. However, regardless of what one assumes, it ought to be straightforward and transparent to do something like this:

```python
ds = yt.load("some_dataset", unit_system="code")
```

and expect magnetic fields to have SI dimensions instead of Gaussian CGS dimensions if the underlying code units do. It turns out that this doesn't always work correctly, as seen in Issue #3471, which was originally posted about another problem but in fact led to the discovery of this issue. 

The problem is that we need something that comes even before unit detection from the dataset, to let yt know whether or not it should expect something with dimensions of SI current in it when creating the code unit system.  There is a chicken-and-egg problem where we have to set up a code unit system before we can detect units, but only when we detect units can we know whether we need an SI current unit, which is info you need to create the code unit system in the first place. It's quite awful.

This PR is an attempt to fix the logic that handles this situation, as well as the case where we have an SI code system but request a CGS system (and vice-versa), which is currently very brittle. This requires waiting on creating the `code_magnetic` unit until we can be sure about what kind of magnetic dimensions we will have in the dataset. Unlike the previous attempt (PR #3787), this does not require any new class attributes.

I've updated the documentation and the skeleton frontend to reflect how to ensure that SI/MKSA units are used for code units, and I've also updated the stream dataset documentation to show how to make sure this happens for in-memory datasets.

I also added some new tests and fixed some other ones. Finally, this does not fix Issue #3471--that will have to be addressed in a separate PR.

## PR Checklist

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -29,6 +29,8 @@ answer_tests:
     - yt/frontends/athena/tests/test_outputs.py:test_stripping
 
   local_athena_pp_006:
+    # disabling disk test for now until we have better support
+    # for this dataset
     #- yt/frontends/athena_pp/tests/test_outputs.py:test_disk
     - yt/frontends/athena_pp/tests/test_outputs.py:test_AM06
 
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -28,8 +28,8 @@ answer_tests:
     - yt/frontends/athena/tests/test_outputs.py:test_blast
     - yt/frontends/athena/tests/test_outputs.py:test_stripping
 
-  local_athena_pp_004:
-    - yt/frontends/athena_pp/tests/test_outputs.py:test_disk
+  local_athena_pp_006:
+    #- yt/frontends/athena_pp/tests/test_outputs.py:test_disk
     - yt/frontends/athena_pp/tests/test_outputs.py:test_AM06
 
   local_chombo_005:
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1119,27 +1119,48 @@ class Dataset(abc.ABC):
         return self.refine_by ** (l1 - l0)
 
     def _assign_unit_system(self, unit_system):
-        if unit_system == "cgs":
-            current_mks_unit = None
+        # we need to determine if the requested unit system
+        # is mks-like: i.e., it has a current with the same
+        # dimensions as amperes.
+        mks_system = False
+        if getattr(self, "magnetic_unit", None):
+            mag_dims = self.magnetic_unit.units.dimensions.free_symbols
         else:
-            current_mks_unit = "A"
-        magnetic_unit = getattr(self, "magnetic_unit", None)
-        if magnetic_unit is not None:
-            if unit_system == "mks":
-                if current_mks not in self.magnetic_unit.units.dimensions.free_symbols:
-                    self.magnetic_unit = self.magnetic_unit.to("gauss").to("T")
-                self.unit_registry.modify("code_magnetic", self.magnetic_unit.value)
-            else:
-                # if the magnetic unit is in T, we need to create the code unit
-                # system as an MKS-like system
-                if current_mks in self.magnetic_unit.units.dimensions.free_symbols:
-                    self.magnetic_unit = self.magnetic_unit.to("T").to("gauss")
+            mag_dims = None
+        if unit_system != "code":
+            # if the unit system is known, we can check if it
+            # has a "current_mks" unit
+            us = unit_system_registry[str(unit_system).lower()]
+            mks_system = us.base_units[current_mks] is not None
+        elif mag_dims and current_mks in mag_dims:
+            # if we're using the not-yet defined code unit system,
+            # then we check if the magnetic field unit has a SI
+            # current dimension in it
+            mks_system = True
+        # Now we get to the tricky part. If we have an MKS-like system but
+        # we asked for a conversion to something CGS-like, or vice-versa,
+        # we have to convert the magnetic field
+        if mag_dims is not None:
+            if mks_system:
+                if current_mks not in mag_dims:
+                    self.magnetic_unit = self.quan(
+                        self.magnetic_unit.to_value("gauss") * 1.0e-4, "T"
+                    )
+                    # The following modification ensures that we get the conversion to
+                    # mks correct
+                    self.unit_registry.modify(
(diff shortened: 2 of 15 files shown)


jzuhone:
@matthewturk @neutrinoceros @stonnes this is an attempt to do better than PR #3787, without defining a new class attribute.

yt/frontends/boxlib/tests/test_outputs.py
```diff
@@ -315,11 +322,28 @@ def test_CastroDataset_2():
     assert isinstance(data_dir_load("castro_sod_x_plt00036"), CastroDataset)
 
 
-@requires_file(LyA)
+@requires_file(plasma)
```
neutrinoceros, line 325: why ?

neutrinoceros (dismissed):
This seems like a more elegant solution than the previous attempt, well done !
Left a couple questions, but I'd be fine with merging this as is. However I would still like that someone else had a look too.

yt/frontends/boxlib/tests/test_outputs.py
jzuhone: If you look inside the test, it's using the `plasma` file and not `LyA`.

neutrinoceros: oh wow how could I miss that. Good catch.

yt/fields/tests/test_magnetic_fields.py
```diff
@@ -23,18 +23,22 @@ def test_magnetic_fields():
     for field in data1:
         data2[field] = (data1[field][0] * 1.0e4, "gauss")
 
-    ds1 = load_uniform_grid(data1, ddims, unit_system="cgs")
+    ds1 = load_uniform_grid(data1, ddims, magnetic_unit=(1.0, "T"), unit_system="cgs")
```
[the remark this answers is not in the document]
jzuhone: Done

yt/fields/tests/test_magnetic_fields.py
```diff
@@ -64,16 +71,26 @@ def test_magnetic_fields():
...
-    ) / (2.0 * mu_0)
+    ) / (8.0 * np.pi)
     emag3.convert_to_units("code_pressure")
 
+    emag4 = (
+        dd4[("gas", "magnetic_field_x")] ** 2
+        + dd4[("gas", "magnetic_field_y")] ** 2
+        + dd4[("gas", "magnetic_field_z")] ** 2
+    ) / (2.0 * mu_0)
+    emag4.convert_to_units("code_pressure")
```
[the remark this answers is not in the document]
jzuhone: Done

neutrinoceros (dismissed):

tests/tests.yaml
```diff
@@ -28,8 +28,8 @@ answer_tests:
     - yt/frontends/athena/tests/test_outputs.py:test_blast
     - yt/frontends/athena/tests/test_outputs.py:test_stripping
 
-  local_athena_pp_004:
-    - yt/frontends/athena_pp/tests/test_outputs.py:test_disk
+  local_athena_pp_006:
```
matthewturk, line 31: Just so I know -- why are we skipping 5, and can you put a comment in this file as to why we are deactivating `test_disk`?

yt/data_objects/static_output.py
```diff
@@ -1119,27 +1119,48 @@ def relative_refinement(self, l0, l1):
...
+            # if we're using the not-yet defined code unit system,
+            # then we check if the magnetic field unit has a SI
+            # current dimension in it
+            mks_system = True
+        # Now we get to the tricky part. If we have an MKS-like system but
+        # we asked for a conversion to something CGS-like, or vice-versa,
+        # we have to convert the magnetic field
+        if mag_dims is not None:
+            if mks_system:
+                if current_mks not in mag_dims:
```
matthewturk, line 1145: Since there's no `else` inside `if mks_system` can we consolidate this `if` into the `mks_system` if?

yt/data_objects/static_output.py
```diff
@@ -1165,25 +1186,34 @@ def _create_unit_registry(self, unit_system):
...
+        # 1 cm = 0.01 m
         self.unit_registry.add("code_length", 0.01, dimensions.length)
+        # 1 g = 0.001 kg
         self.unit_registry.add("code_mass", 0.001, dimensions.mass)
+        # 1 g/cm**3 = 1000 kg/m**3
         self.unit_registry.add("code_density", 1000.0, dimensions.density)
+        # 1 erg/g = 1.0e-4 J/kg
         self.unit_registry.add(
-            "code_specific_energy", 1.0, dimensions.energy / dimensions.mass
+            "code_specific_energy", 1.0e-4, dimensions.energy / dimensions.mass
```
matthewturk, line 1202: This seems like a big change for `code_specific_energy`, isn't it?

matthewturk (dismissed):

tests/tests.yaml
jzuhone: Skipping 5 because I made a mistake in incrementing the number--but I can make a comment here

yt/data_objects/static_output.py
jzuhone: Yes--I just checked again and it was wrong before.

yt/data_objects/static_output.py
jzuhone: The consolidation was a bit more complicated than that but I was able to simplify it a bit.

jzuhone:
@neutrinoceros i think we can merge now

neutrinoceros (approved):

neutrinoceros:
manually backported into #3789. @jzuhone, I think you're up to review that one now :)

PR #3811, opened 2022-02-15
BUG: bring back support to group folder structure in RAMSES

### LINKED ISSUE #3785: RAMSES loader does not recognize "groups" file structure

cphyc:
## PR Summary

This brings back support to RAMSES folder structures like `output_XXXXX/group_YYYYY`. Fixes #3785.
In particular, the output number should be read off the `XXXXX` part, not the `YYYYY` part.

## PR Checklist

Note: I have tested the bugfix on a (massive) simulation with the following script. Unfortunately, we do not have any simulation dataset with the group structure on yt-data so testing would imply more work that I can reasonably put
```python
import yt

yt.mylog.setLevel(40)
eps = 0.0001
bbox = [[0.5-eps]*3, [0.5+eps]*3]

def test_me(path):
    print(f"Testing {path=}")
    ds = yt.load(path, bbox=bbox)
    ds.index

    assert len(ds.index.domains) > 0

    ds.r["density"]
    ds.r["DM", "particle_identity"]

test_me("/data84/obelisk/RHD/OUTPUT_DIR/output_00075")
test_me("/data84/obelisk/RHD/OUTPUT_DIR/output_00075/group_00001/")
test_me("/data84/obelisk/RHD/OUTPUT_DIR/output_00075/group_00001/info_00075.txt")
test_me("/data84/obelisk/RHD/OUTPUTS/output_00075/info_00075.txt")
test_me("/data84/obelisk/RHD/OUTPUTS/output_00075")
```

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -113,7 +113,7 @@ class RAMSESFileSanitizer:
         # Three cases for path
         # 1. path == output_XXXXX and group_00001 exists
         # 2. path == output_XXXXX and group_00001 does not exist
-        # 2. path == output_XXXXX/group_YYYYYY
+        # 3. path == output_XXXXX/group_YYYYYY
         iout_match = OUTPUT_DIR_RE.match(path.name)
         if iout_match:
             group_dir = path / "group_00001"
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -1,9 +1,9 @@
 import os
-from typing import Optional, Tuple
 import weakref
 from collections import defaultdict
 from itertools import product
 from pathlib import Path
+from typing import Optional, Tuple
 
 import numpy as np
 
@@ -106,7 +106,9 @@ class RAMSESFileSanitizer:
         return ok
 
     @staticmethod
-    def _match_output_and_group(path: Path) -> Tuple[Path, Optional[Path], Optional[int]]:
+    def _match_output_and_group(
+        path: Path,
+    ) -> Tuple[Path, Optional[Path], Optional[str]]:
         iout_match = OUTPUT_DIR_RE.match(path.name)
 
         if not iout_match:
@@ -115,6 +117,7 @@ class RAMSESFileSanitizer:
         # Two cases:
         # 1. regular structure (all files in the output_XXXXX folder)
         # 2. group structure (files spread in group_YYYYY folders)
+        group_dir: Optional[Path]
 
         if iout_match.group(1) == "output":
             # Case 1, unless the current folder contains a group folder itself
@@ -133,22 +136,27 @@ class RAMSESFileSanitizer:
 
         return output_dir, group_dir, iout_match.group(2) if iout_match else None
 
-
     @classmethod
-    def test_with_folder_name(cls, output_dir) -> Tuple[bool, Optional[Path], Optional[Path], Optional[Path]]:
+    def test_with_folder_name(
+        cls, output_dir: Path
+    ) -> Tuple[bool, Optional[Path], Optional[Path], Optional[Path]]:
         output_dir, group_dir, iout = cls._match_output_and_group(output_dir)
-
         ok = output_dir.is_dir() and iout is not None
+
+        info_fname: Optional[Path]
+
         if ok:
             ok &= cls.check_standard_files(group_dir or output_dir, iout)
             info_fname = (group_dir or output_dir) / f"info_{iout}.txt"
         else:
-            output_dir, group_dir, info_fname = None, None, None
+            info_fname = None
 
         return ok, output_dir, group_dir, info_fname
 
     @classmethod
-    def test_with_standard_file(cls, filename) -> Tuple[bool, Optional[Path], Optional[Path], Optional[Path]]:
+    def test_with_standard_file(
+        cls, filename: Path
+    ) -> Tuple[bool, Optional[Path], Optional[Path], Optional[Path]]:
         output_dir, group_dir, iout = cls._match_output_and_group(filename.parent)
         ok = (
             filename.is_file()
@@ -156,11 +164,13 @@ class RAMSESFileSanitizer:
             and iout is not None
         )
 
+        info_fname: Optional[Path]
+
         if ok:
             ok &= cls.check_standard_files(group_dir or output_dir, iout)
             info_fname = (group_dir or output_dir) / f"info_{iout}.txt"
         else:
-            output_dir, group_dir, info_fname = None, None, None
+            info_fname = None
 
         return ok, output_dir, group_dir, info_fname
 
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -110,19 +110,25 @@ class RAMSESFileSanitizer:
         path: Path,
     ) -> Tuple[Path, Optional[Path], Optional[str]]:
 
-        # Three cases for path
-        # 1. path == output_XXXXX and group_00001 exists
-        # 2. path == output_XXXXX and group_00001 does not exist
-        # 3. path == output_XXXXX/group_YYYYYY
-        iout_match = OUTPUT_DIR_RE.match(path.name)
-        if iout_match:
-            group_dir = path / "group_00001"
+        # Make sure we work with a directory of the form `output_XXXXX`
+        for p in (path, path.parent):
+            match = OUTPUT_DIR_RE.match(p.name)
 
-            return path, group_dir if group_dir.is_dir() else None, iout_match.group(1)
+            if match:
+                path = p
+                break
+
+        if match is None:
+            return path, None, None
+
+        iout = match.group(1)
+
+        # See whether a folder named `group_YYYYY` exists
+        group_dir = path / "group_00001"
+        if group_dir.is_dir():
+            return path, group_dir, iout
         else:
-            iout_match = OUTPUT_DIR_RE.match(path.parent.name)
-
-            return path.parent, path, iout_match.group(1) if iout_match else None
+            return path, None, iout
 
     @classmethod
     def test_with_folder_name(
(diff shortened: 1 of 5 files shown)


yt/frontends/ramses/data_structures.py
```diff
@@ -109,37 +105,74 @@ def check_standard_files(folder, iout):
...
+                group_dir = path / "group_00001"
+            else:
+                group_dir = None
+            output_dir = path
+        else:
+            # Case 2, we need to extract output number from parent output folder
+            group_dir = path
+            output_dir = path.parent
+
+        # Match again to find output number
```
neutrinoceros, line 134: It seems to me that we're pilling more and more unnecessary complexity here because `OUTPUT_DIR_RE` is _too_ flexible. Wouldn't it be a little simpler to have two separate expressions `OUTPUT_DIR_RE` and `GROUP_DIR_RE` ?

neutrinoceros (commented):
I think it'd be worth trying to simplify this, especially if we can't afford to test it thoroughly.

yt/frontends/ramses/data_structures.py
```diff
@@ -109,37 +105,62 @@ def check_standard_files(folder, iout):
...
 
+    @staticmethod
+    def _match_output_and_group(
+        path: Path,
+    ) -> Tuple[Path, Optional[Path], Optional[str]]:
+
+        # Three cases for path
+        # 1. path == output_XXXXX and group_00001 exists
+        # 2. path == output_XXXXX and group_00001 does not exist
+        # 2. path == output_XXXXX/group_YYYYYY
```
neutrinoceros, line 116:
```suggestion
        # 3. path == output_XXXXX/group_YYYYYY
```

yt/frontends/ramses/data_structures.py
```diff
@@ -109,37 +105,62 @@ def check_standard_files(folder, iout):
...
+        # 3. path == output_XXXXX/group_YYYYYY
+        iout_match = OUTPUT_DIR_RE.match(path.name)
+        if iout_match:
+            group_dir = path / "group_00001"
+
+            return path, group_dir if group_dir.is_dir() else None, iout_match.group(1)
+        else:
+            iout_match = OUTPUT_DIR_RE.match(path.parent.name)
+
+            return path.parent, path, iout_match.group(1) if iout_match else None
```
neutrinoceros, line 125: honestly this is still hardly readable to me; to the point it's not easy to find a better alternative.*
Can we having ternaries inside tuples maybe ?

yt/frontends/ramses/data_structures.py
```diff
@@ -109,37 +105,62 @@ def check_standard_files(folder, iout):
...
+        ok = output_dir.is_dir() and iout is not None
+
+        info_fname: Optional[Path]
+
         if ok:
-            iout = iout_match.group(2)
-            ok &= cls.check_standard_files(output_dir, iout)
-            info_fname = output_dir / f"info_{iout}.txt"
+            ok &= cls.check_standard_files(group_dir or output_dir, iout)
+            info_fname = (group_dir or output_dir) / f"info_{iout}.txt"
```
neutrinoceros, line 138: `parent_dir` is more of a place-holder here, I'd suggest finding a better name
```suggestion
            parent_dir = group_dir or output_dir
            ok &= cls.check_standard_files(parent_dir, iout)
            info_fname = parent_dir / f"info_{iout}.txt"
```

yt/frontends/ramses/data_structures.py
neutrinoceros: I'm also struggling to convince myself we actually need this double match now. Can't we just get everything we need by trying to match the regex to the parent dir in all cases ?

yt/frontends/ramses/data_structures.py
```diff
@@ -109,37 +105,70 @@ def check_standard_files(folder, iout):
...
+        # Make sure we work with a directory of the form `output_XXXXX`
+        for p in (path, path.parent):
+            match = OUTPUT_DIR_RE.match(p.name)
+
+            if match:
+                path = p
+                break
+
+        if match is None:
+            return path, None, None
```
neutrinoceros, line 122: ?
```suggestion
        else:
            return path, None, None
```

neutrinoceros:
I feel much more confortable with the simplified implementation, thanks ! would it be doable to add a small unit test for the `RAMSESFileSanitizer` class that wouldn't require pushing another dataset to the database ?

yt/frontends/ramses/tests/test_file_sanitizer.py
```diff
@@ -0,0 +1,105 @@
...
+    for path in chain(*(pt.paths_to_try for pt in valid_path_tuples)):
+        expected_error_message = (
+            f"No such file or directory '{str(path)}/does_not_exist.txt'"
+        )
+        sanitizer = RAMSESFileSanitizer(path / "does_not_exist.txt")
+        with pytest.raises(FileNotFoundError, match=expected_error_message):
+            sanitizer.validate()
+
+    expected_error_message = "No such file or directory '/this/does/not/exist'"
+    sanitizer = RAMSESFileSanitizer("/this/does/not/exist")
```
neutrinoceros, line 103: I anticipate this should fail on windows

yt/frontends/ramses/tests/test_file_sanitizer.py
```diff
@@ -0,0 +1,105 @@
...
+            sanitizer = RAMSESFileSanitizer(path)
+            sanitizer.validate()
+
+            assert sanitizer.root_folder == answer.output_dir
+            assert sanitizer.group_name == answer.group_dir_name
+            assert sanitizer.info_fname == answer.info_file
+
+
+def test_invalid_sanitizing(valid_path_tuples, invalid_path_tuples):
+    for path in chain(*(pt.paths_to_try for pt in invalid_path_tuples)):
```
neutrinoceros, line 77: There's a more "pytest-ic" way to write such tests with parametrised fixtures. Here's an example: https://github.com/neutrinoceros/yt_idefix/blob/14052de1a2d1a6404fcc7c4164a4aa4d1ba283a0/tests/conftest.py#L147

I can help refactoring this particular case if you want

neutrinoceros: @cphyc I'll take that back. I'm convinced these tests could be written in a simpler form but I haven't been able to find how. I'd be prone to accept this PR as is (provided tests actually pass on windows)

neutrinoceros:
@cphyc I see that your tests are also failing on Jenkins because you forgot to ignore them on nose :)

neutrinoceros:
pre-commit fails because of an incompatibility between black and click that happened this week (https://github.com/psf/black/issues/2964)
it'll be fixed on yt's main branch tomorrow with the auto-update bot. In the mean time I'm patching it locally.

neutrinoceros:
Also got an instability from matplotlib + pillow. CI should be stabilised globally by https://github.com/yt-project/yt/pull/3882

neutrinoceros (approved):

neutrinoceros (approved):

neutrinoceros:
manually backported as #3886

Issue #2554, opened 2020-04-21
ensure release and development directions are consistent


munkm:
In preparing the yt-3.6 release and afterwards I noticed that we have differing documentation on releasing and development. I've included a few notes about it here. I will submit a PR with these fixes later, but please weigh in if you want to suggest anything. 

Example1: ytep-0008 and the release documentation
see [ytep-0008](https://ytep.readthedocs.io/en/latest/YTEPs/YTEP-0008.html#long-term-support-branch) and [release documentation](https://yt-project.org/docs/dev/developing/releasing.html)
* ytep 0008 mentions the existence of a long-term support branch (with the naming convention `yt-3.x` and the release notes don't at all. 
* the ytep is very behind on the release managers and the dates of recent yt releases. This should be updated
* the ytep is behind on the release information. this should be updated

Example 2: 'what constitutes a release' in ytep 0008 doesn't include info on conda-forge
* this should be updated 

Example 3: "requirements for code submission" section differs from governance
https://yt-project.org/doc/developing/developing.html#requirements-for-code-submission

* this is mild but we don't mention the requirements for creating an issue in our governance at all. We also don't use closed issues to generate our release notes, so I think this requirement should probably be removed and we should add notes on the required number of approving reviewers. Alternatively, we could remove this section and link directly to the governance so we don't have to maintain these directions in two different places. But maybe the governance isn't the right place for it?

neutrinoceros:
So after a discussion you and I had on #2690, I realised that the dev guide currently doesn't have a "How to deprecate and remove features in yt" section. Such a section would be really helpful to ensure consistency in the procedure and maybe provide insight on how we keep track of stuff that's targeted for deprecation.
Do you think this belongs here ?

munkm:
I think that is definitely relevant to the broader issues of the developer guide. If we do want the release manager to be the dedicated person to make sure all deprecations are removed on a timetable then it's absolutely something that should be in this issue.

neutrinoceros:
I've added instructions for closing a deprecation cycle in #3806

neutrinoceros:
Example 1 is handled in https://github.com/yt-project/ytep/pull/22

PR #3813, opened 2022-02-16
DOC: update contribution guidelines

neutrinoceros:
## PR Summary
This is aimed at fixing #2554
what's missing still is a link to the document where we define the number of reviews needed for all kinds of PRs, because I couldn't find it back.
Keeping this as a draft until then


neutrinoceros:
found the document I had in mind: https://github.com/yt-project/governance/blob/7e8793735f067f2681966ea0e848d542a813b1aa/source/voting.rst
However, I didn't find where this was being deployed outside Github. On second thought I'm not convinced that the number of reviews required for PRs is really something the target audience (new contributors) needs to care about, so I think I'll just avoid bloating the docs with it.

matthewturk (approved):

PR #3815, opened 2022-02-17
Add three new derived fields to the AREPO frontend

jzuhone:
## PR Checklist

* Alias the `("PartType0","pressure")` field to `("gas","pressure")` (not sure why this wasn't already one
* Create a `"cell_volume"` field
* Create a `"cosmic_ray_pressure"` field
* Rename `"cr_energy_density"` and `"specific_cr_energy"` to `"cosmic_ray_energy_density"` and `"specific_cosmic_ray_energy"`

### DIFF
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -150,7 +150,7 @@ class ArepoFieldInfo(GadgetFieldInfo):
             )
 
             def _cr_pressure(field, data):
-                return data["gas", "cr_energy_density"] / 3.0
+                return (data.ds.gamma_cr - 1.0) * data["gas", "cr_energy_density"]
 
             self.add_field(
                 ("gas", "cr_pressure"),
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -148,3 +148,13 @@ class ArepoFieldInfo(GadgetFieldInfo):
                 sampling_type="local",
                 units=self.ds.unit_system["pressure"],
             )
+
+            def _cr_pressure(field, data):
+                return data["gas", "cr_energy_density"] / 3.0
+
+            self.add_field(
+                ("gas", "cr_pressure"),
+                _cr_pressure,
+                sampling_type="local",
+                units=self.ds.unit_system["pressure"],
+            )
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -45,6 +45,18 @@ class ArepoFieldInfo(GadgetFieldInfo):
     def setup_gas_particle_fields(self, ptype):
         super().setup_gas_particle_fields(ptype)
 
+        # Since the AREPO gas "particles" are Voronoi cells, we can
+        # define a volume here
+        def _volume(field, data):
+            return data[ptype, "mass"] / data[ptype, "density"]
+
+        self.add_field(
+            (ptype, "cell_volume"),
+            function=_volume,
+            sampling_type="particle",
+            units=self.ds.unit_system["volume"],
+        )
+
         if (ptype, "InternalEnergy") in self.field_list:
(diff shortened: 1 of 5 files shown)


yt/frontends/arepo/fields.py
```diff
@@ -45,6 +45,18 @@ def setup_particle_fields(self, ptype, *args, **kwargs):
     def setup_gas_particle_fields(self, ptype):
         super().setup_gas_particle_fields(ptype)
 
+        # Since the AREPO gas "particles" are Voronoi cells, we can
+        # define a volume here
+        def _volume(field, data):
+            return data[ptype, "mass"] / data[ptype, "density"]
```
matthewturk, line 51: Both mass and density are in the file, right?

yt/frontends/arepo/fields.py
```diff
@@ -134,3 +148,13 @@ def _cr_energy_density(field, data):
                 sampling_type="local",
                 units=self.ds.unit_system["pressure"],
             )
+
+            def _cr_pressure(field, data):
+                return data["gas", "cr_energy_density"] / 3.0
```
matthewturk, line 153: Why `/ 3.0`?

jzuhone: The cosmic rays (CRs) are an ideal relativistic gas with gamma = 4/3, so P = (gamma-1)*internal_energy_density = 1/3*internal_energy_density.

However, I just remembered that this is a parameter that you can set in the parameter file (even though it is nearly always 1/3), so I should probably make this more general.

jzuhone: Which means it probably needs a test dataset

neutrinoceros (commented):
I'm not opposing that the new field be named `"cr_pressure"` but could it be aliased to a more explicit name `"cosmic_ray_pressure"` ?

jzuhone:
@neutrinoceros if we're going to do that, we should probably also change `specific_cr_energy` and `cr_energy_density` also for consistency.

neutrinoceros:
Oh yes, I didn't notice these existed already. I would be in favour of adding aliases for them too indeed.

yt/sample_data_registry.json
```diff
@@ -29,6 +29,12 @@
     "load_name": "snapshot_150.hdf5",
     "url": "https://yt-project.org/data/ArepoBullet.tar.gz"
   },
+  "ArepoCosmicRays.tar.gz": {
+    "hash": "e664ea7f634fff778511e9ae7b2e375ba9c5099e427f7c037d48bba9944388f4",
+    "load_kwargs": {},
+    "load_name": "snapshot_039.hdf5",
+    "url": "https://yt-project.org/data/ArepoCosmicRays.tar.gz"
+  },
```
neutrinoceros, line 37: I've checked that this new dataset loads perfectly with `yt.load_sample`

neutrinoceros (approved):

yt/frontends/arepo/fields.py
```diff
@@ -125,12 +139,42 @@ def _el_number_density(field, data):
 
         if (ptype, "CosmicRaySpecificEnergy") in self.field_list:
 
+            self.alias(
+                (ptype, "specific_cosmic_ray_energy"),
```
cphyc, line 143: I am a bit confused reading the lines above - I assumed `ptype` was equal to `gas`. Did I miss something?

TBH the entire function looks very obscure to me and I am not used to dealing with particle-based datasets, so this may be me :)

jzuhone: @cphyc in this case, `ptype` is `"PartType0"`, which is the gas type.

chummels (approved):
I've looked this over, and everything looks good.  Great to have CR support for AREPO as well as cell_volume and gas pressure fields for AREPO datasets.

PR #3817, opened 2022-02-18
ENH: add support for VelocityCallback and MagFieldCallBack in spherical coordinates

neutrinoceros:
## PR Summary

minimal demo
```python
import numpy as np
import yt
from unyt import cm, s

shape = (32, 128, 1)
a = np.sin(np.linspace(-np.pi/2, np.pi/2, 128))
b = np.ones((32, 128))
c = np.zeros((32, 128))

data = {
    ("gas", "velocity_r"): np.reshape(a * b, shape) * cm/s,
    ("gas", "velocity_theta"): np.reshape(a * c, shape) * cm/s,
}

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.0, 5.0], [np.pi / 2 - 0.5, np.pi / 2 + 0.5], [-0.1, +0.1]]),
    geometry="spherical",
)

p = yt.SlicePlot(ds, "phi", "velocity_r")
p.annotate_velocity()
p.save("/tmp/")
```

UniformGridData_Slice_phi_velocity_r

image actualized on 2022-08-01
<details> <summary>older version</summary>

UniformGridData_Slice_phi_velocity_r

note that the quiver plot is drawn well outside of the actual data domain. I think this effect is due to the current implementation of the pixelizer routine (`pixelize_cylinder`), combined with the "low resolution" buffer associated with the quiver plot for readability. I consider this a bug, but it's out of scope here.
Edit: this is solved in #3818

</details>

### DIFF
--- a/yt/fields/vector_operations.py
+++ b/yt/fields/vector_operations.py
@@ -195,6 +195,8 @@ def create_vector_fields(registry, basename, field_units, ftype="gas", slice_inf
         validators=[ValidateParameter(f"bulk_{basename}")],
     )
 
+    axis_names = registry.ds.coordinates.axis_order
+
     if not is_curvilinear(registry.ds.geometry):
 
         # The following fields are invalid for curvilinear geometries
@@ -553,11 +555,71 @@ def create_vector_fields(registry, basename, field_units, ftype="gas", slice_inf
                     data[(ftype, "theta")]
                 ) - data[(ftype, f"{basename}_theta")] * np.sin(data[(ftype, "theta")])
 
-        if registry.ds.dimensionality == 3:
+        if registry.ds.dimensionality >= 2:
             registry.add_field(
                 (ftype, f"{basename}_cartesian_z"),
                 sampling_type="local",
                 function=_cartesian_z,
+                units=field_units,
+                display_field=True,
+            )
+
+    if registry.ds.geometry == "spherical":
+
+        def _cylindrical_radius_component(field, data):
+            return (
+                np.sin(data[(ftype, "theta")]) * data[(ftype, f"{basename}_r")]
+                + np.cos(data[(ftype, "theta")]) * data[(ftype, f"{basename}_theta")]
+            )
+
+        registry.add_field(
+            (ftype, f"{basename}_cylindrical_radius"),
+            sampling_type="local",
+            function=_cylindrical_radius_component,
+            units=field_units,
+            display_field=True,
+        )
+
+        registry.alias(
+            (ftype, f"{basename}_cylindrical_z"),
+            (ftype, f"{basename}_cartesian_z"),
+        )
+
+        # define vector components appropriate for 'theta'-normal plots.
+        # The projection plane is called 'conic plane' in the code base as well as docs.
+        # Contrary to 'poloidal' and 'toroidal', this isn't a widely spread
+        # naming convention, but here it is exposed to users as part of dedicated
+        # field names, so it needs to be stable.
+        def _conic_x(field, data):
+            rax = axis_names.index("r")
+            pax = axis_names.index("phi")
+            bc = data.get_field_parameter(f"bulk_{basename}")
+            return np.cos(data[ftype, "phi"]) * (
+                data[ftype, f"{basename}_r"] - bc[rax]
+            ) - np.sin(data[ftype, "phi"]) * (data[ftype, f"{basename}_phi"] - bc[pax])
+
+        def _conic_y(field, data):
+            rax = axis_names.index("r")
+            pax = axis_names.index("phi")
+            bc = data.get_field_parameter(f"bulk_{basename}")
+            return np.sin(data[(ftype, "phi")]) * (
+                data[(ftype, f"{basename}_r")] - bc[rax]
+            ) + np.cos(data[(ftype, "phi")]) * (
+                data[(ftype, f"{basename}_phi")] - bc[pax]
+            )
+
+        if registry.ds.dimensionality == 3:
+            registry.add_field(
+                (ftype, f"{basename}_conic_x"),
+                sampling_type="local",
+                function=_conic_x,
+                units=field_units,
+                display_field=True,
+            )
+            registry.add_field(
+                (ftype, f"{basename}_conic_y"),
+                sampling_type="local",
+                function=_conic_y,
                 units=field_units,
                 display_field=True,
             )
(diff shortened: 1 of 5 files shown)


neutrinoceros:
While I'm at it, I'll add support for "conic" projections too (theta-normal), illustrated here with a purely azimuthal magnetic field
```python
import numpy as np
import yt
from yt.units import G

shape = (32, 128, 128)
a = np.sin(np.linspace(-np.pi / 2, np.pi / 2, 128))
b = np.ones((32, 128, 128))
c = np.zeros((32, 128, 128))

data = {
    ("gas", "magnetic_field_r"): np.reshape(a * c, shape) * G,
    ("gas", "magnetic_field_theta"): np.reshape(a * c, shape) * G,
    ("gas", "magnetic_field_phi"): np.reshape(a * b, shape) * G,
}

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.0, 5.0], [np.pi / 2 - 0.5, np.pi / 2 + 0.5], [-np.pi, +np.pi]]),
    geometry="spherical",
)

p = yt.SlicePlot(ds, "theta", "magnetic_field_phi")
p.annotate_magnetic_field(
    normalize=True
)
p.save("/tmp/")
```
UniformGridData_Slice_theta_magnetic_field_phi

neutrinoceros:
Tests exploding everywhere, switching to draft for now.

yt/fields/vector_operations.py
```diff
@@ -599,6 +599,66 @@ def _cartesian_z(field, data):
...
+                + data[(ftype, f"{basename}_theta")] ** 2
+            )
+
+        registry.add_field(
+            (ftype, f"{basename}_poloidal_magnitude"),
+            sampling_type="local",
+            function=_poloidal_magnitude,
+            units=field_units,
+            display_field=True,
+        )
```
neutrinoceros, line 632: this is not necessary for the rest of the PR, I'm just throwing it in while I'm in there

yt/fields/vector_operations.py
```diff
@@ -599,6 +599,71 @@ def _cartesian_z(field, data):
...
+
+        registry.alias(
+            (ftype, f"{basename}_toroidal_magnitude"),
+            (ftype, f"{basename}_phi"),
+        )
```
neutrinoceros, line 637: Two notes here:
1) these are not necessary for the rest of the PR, I'm just throwing them in while I'm here
2) I'm deliberately not following the existing naming convention for "magnetic_field_poloidal" and "magnetic_field_toroidal" fields that are defined in a field plugin, because their definition is not robust ("bulk magnetic field" is a requirement) and they live in the plugin system, which is fragile, so I prefer setting solid defaults here, but I'll avoid name collisions.

matthewturk:
I think your diagnosis -- the buffer being low-resolution -- is the root cause, but perhaps it can be fixed by more carefully adjusting the arguments to the pixelization routine to query at the specific pixel centers.

yt/fields/vector_operations.py
matthewturk: The plugin system is fragile?

Also, I thought that bulk magnetic field had a default value of 0?

neutrinoceros: I'm saying it's fragile because, AFICT, the field plugins rely on yt.mods, which has has comments saying it's dangerous, and is also deprecated.

I wasn't aware that there was a default bulk magnetic field however.

matthewturk:
I did not realize it depended on `yt.mods`.  Oof!  Also, I could be wrong about the mag field.

neutrinoceros:
> I think your diagnosis -- the buffer being low-resolution -- is the root cause, but perhaps it can be fixed by more carefully adjusting the arguments to the pixelization routine to query at the specific pixel centers.

How about the proposed fix in #3818 ?

yt/fields/vector_operations.py
```diff
@@ -599,6 +599,71 @@ def _cartesian_z(field, data):
                 display_field=True,
             )
 
+    if registry.ds.geometry == "spherical":
```
matthewturk, line 602: I really like the usage of the cylindrical components inside our spherical data.

yt/fields/vector_operations.py
```diff
@@ -599,6 +599,71 @@ def _cartesian_z(field, data):
...
+
+        registry.alias(
+            (ftype, f"{basename}_cylindrical_z"),
+            (ftype, f"{basename}_cartesian_z"),
+        )
+
+        def _poloidal_magnitude(field, data):
+            return np.sqrt(
+                data[(ftype, f"{basename}_r")] ** 2
+                + data[(ftype, f"{basename}_theta")] ** 2
```
matthewturk, line 623: Can you add a comment here?  Just one that describes very specifically what the fields are.

matthewturk (approved):
LGTM

neutrinoceros:
I need to revisit this. Switching to draft in the mean time. For now I'll just resolve a conflict and add a comment as requested by Matt.

yt/fields/vector_operations.py
neutrinoceros: I've added 2 detailed comments.

neutrinoceros:
refreshing CI

neutrinoceros:
rebase to resolve conflicts and refresh CI

yt/fields/vector_operations.py
neutrinoceros: I'll take that back actually:
- `"bulk_magnetic_field"` field parameter is indeed not "required" and correctly set to 0 by default.
- it doesn't matter how fragile the field plugin system is, as it's still part of things we want to continue supporting. So either we'll come up with a smart way to maintain backward compatibility via a refactor, or we can just _not_ follow through yt.mods' deprecation, if that turn out impossible.

I'll lighten this PR and removing these concurrent definitions.

neutrinoceros: Turns out it's harder than I thought, see #4020
I may just remove this PR from the 4.1 milestone if I don't make fast enough progress

neutrinoceros: actually solving this much quicker than I thought 20min ago: #4021

neutrinoceros:
this is now rebased on top of #4021. I cleaned up a couple redundant definitions as previously discussed, and fixed a couple outstanding bugs.

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@matthewturk can we merge this ?

matthewturk (approved):

PR #3818, opened 2022-02-19
BUG: don't colorize pixels that are not completely within the data domain in pixelize_cylinder routine

neutrinoceros:
## PR Summary
This fixes an issue with the cylindrical pixelizer routine discovered in #3817, where pixels that are not completely within the data domain.
Although a bug fix, I don't think this change is really important before #3817 is merged, so I think this can go to the next feature relase and not be backported.

I'll demonstrate the change using the script from #3817
UniformGridData_Slice_phi_velocity_r

opening as a draft because this branch is based on #3817 for testing purposes

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -167,13 +167,13 @@ answer_tests:
     - yt/frontends/ytdata/tests/test_old_outputs.py:test_old_profile_data
     - yt/frontends/ytdata/tests/test_old_outputs.py:test_old_nonspatial_data
 
-  local_axialpix_008: # PR 3628
+  local_axialpix_009: # PR 3818
     - yt/geometry/coordinates/tests/test_axial_pixelization.py:test_axial_pixelization
 
-  local_cylindrical_background_013:  # PR 3859
+  local_cylindrical_background_014:  # PR 3818
     - yt/geometry/coordinates/tests/test_cylindrical_coordinates.py:test_noise_plots
 
-  local_spherical_background_007:  # PR 3859
+  local_spherical_background_008:  # PR 3818
     - yt/geometry/coordinates/tests/test_spherical_coordinates.py:test_noise_plots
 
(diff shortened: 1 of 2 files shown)


neutrinoceros:
The failures are expected. This changes the way we approximate a disk with square pixels: on main, it is approached from the outer side, while here it's approached from the inner side. As a result, on a 800x800 image, we're shaving of a thin band from the reference pixelizations.

neutrinoceros:
Actually it makes more sense to backport since we may otherwise end up with avoidable conflicts with the backport branch when answer tests are bumped. I'm rebasing this on top of the main branch.

matthewturk:
So, I have to think about this a bit more, but what I'm wondering is if we should instead attempt to set up a "dual grid" where the evaluation points are set at the *centers* of the pixels.  Because we fill from left-right, that would mean basically setting up a grid that overlaid it (and so it would be something like size/factor + 1 I think?) rather than the buffer line setting that is done in here.

neutrinoceros:
> what I'm wondering is if we should instead attempt to set up a "dual grid" where the evaluation points are set at the _centers_ of the pixels.

I'm confused. Aren't evaluation points _already_ at the centers ?

neutrinoceros:
This PR doesn't *need* resolution for 4.0.3 , I'd rather take it out of the milestone rather than having it blocking the release. That said, maybe we can resolve it anyway. I'll take another look at the existing code when I have time and see if this can be revised.

neutrinoceros:
yes there's a top/bottom asymmetry in the cylindrical pixelizer, I never could figure out where it came from but it's not new, and not critical either in production, as far as I can tell.

matthewturk:
yup

neutrinoceros:
@matthewturk right on, this is now ready !

matthewturk (approved):

neutrinoceros:

PR #3821, opened 2022-02-20
MNT: remove an unused and broken helper function

neutrinoceros:
## PR Summary

This private method is not used anywhere and it's not general enough to be relevant, so I propose we just ditch it.


jzuhone (approved):

matthewturk (approved):
It's broken?

neutrinoceros:
Yes. There's an implicit assumption in there that the domain is a complete sphere, possibly hollow, but it doesn't work if the azimuth doesn't go all the way to 2PI, or latitude from 0 to PI.

PR #3822, opened 2022-02-21
ENH: add support for varying colors in quiver annotations

neutrinoceros:
## PR Summary

Improve wrapping around `matplotlib.axes.Axes.quiver`'s complicated signature to enable the optional 5th positional argument `C`.

I'm also refactoring `QuiverCallback` and `CuttingQuiverCallback` to avoid duplicating code as much as possible.

A 'bonus' change here is that I'm allowing arbitary keyword arguments to be passed down to the wrapped method, because it's much more intuitive to pass `cmap="inferno"` than `plot_args=dict(cmap="inferno")`. If requested I can remove this from the PR and possibly repurpose it as a new PR, making similar changes to all callbacks that allow it (i.e. the ones that have only one such dictionary argument)

Another bonus is that I fixed a small bug with the `normalize` bool flag where users would get a divide-by-zero warning when using normalisation for a field that has 0-len elements.

usage
```python
import yt

p = yt.ProjectionPlot(ds, "z", "density")

p.annotate_quiver(
    "velocity_x",
    "velocity_y",
    "vorticity_z",
    normalize=True,
    cmap="inferno_r",
)
p.zoom(40)
p.save('/tmp/')
```
galaxy0030_Projection_z_density

note that this works with matplotlib 2.2 (oldest supported) as well as 3.5 (newest).
see their old doc here https://matplotlib.org/2.2.2/api/_as_gen/matplotlib.axes.Axes.quiver.html?highlight=quiver#matplotlib.axes.Axes.quiver

for completness, introducing this new "field_c" parameter is technically a breaking change for any user code relying on "factor" being usable as a third positional argument. I personally think it's okay, as long as it's explicitly mentioned in the release notes, because it's impossible to mutate the signature continuously without a breaking change. I'm making all arguments keyword-only to remedy this source of problems in the future.

### DIFF
--- a/doc/source/visualizing/callbacks.rst
+++ b/doc/source/visualizing/callbacks.rst
@@ -273,20 +273,23 @@ Overplot Quivers
 Axis-Aligned Data Sources
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. function:: annotate_quiver(self, field_x, field_y, factor=16, scale=None, \
+.. function:: annotate_quiver(self, field_x, field_y, field_c=None, factor=16, scale=None, \
                               scale_units=None, normalize=False, plot_args=None)
 
    (This is a proxy for
    :class:`~yt.visualization.plot_modifications.QuiverCallback`.)
 
    Adds a 'quiver' plot to any plot, using the ``field_x`` and ``field_y`` from
-   the associated data, skipping every ``factor`` datapoints in the
-   discretization. ``scale`` is the data units per arrow length unit using
+   the associated data, skipping every ``factor`` pixels in the
+   discretization. A third field, ``field_c``, can be used as color; which is the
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Got CI green, but more work needed to allow usage of arbitrary color normalizations.

neutrinoceros:
I also would like expose a simple API to associate a smaller colorbar, essentially looking like this:
galaxy0030_Projection_z_density

but doing this correctly necessitates (and gives me a pretext for) a deeper refactor of how colorbars and axis labels work (essentially we need to centralise some bits of code to make them reusable).
This is definitely possible and I don't think that I'm closing any door now if I say it's out of scope for the present PR.

matthewturk (approved):

PR #3824, opened 2022-02-22
BUG: fix a bug in aitoff-hammer pixelizer where some pixels were skipped on the central meridian and the equator

### LINKED ISSUE #3823: BUG: skipped pixels in Aitoff-Hammer pixelization
### Bug report

**Bug summary**

The pixelisation routine used for r-normal plots in spherical coordinates is slightly broken, and it seems that some pixels are sometimes skipped in the vicinity of the equator.

**Code for reproduction**

```python
import numpy as np
import yt
from yt.units import g, cm

shape = (32, 33, 200)

data = {
    "density": np.ones(shape) * g / cm**3,
}

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.1, 1], [0.7, np.pi - 0.7], [0, 2 * np.pi]]),
    geometry="spherical",
)

p = yt.SlicePlot(ds, "r", "density")

p.save("/tmp/res.png")
```

**Actual outcome**
res

In this example we get a series of "holes" in the image, all seemingly symmetrical with respect to the equator.
I note that this phenomenon is apparently very sensitive to the grid's resolution, I just selected the case for with I found the highest number of such artefacts, but I saw this happening at a smaller scale in production.

neutrinoceros:
## PR Summary
fix #3823

taking a better example to illustrate the change:
```python
import numpy as np
import yt
from yt.units import g, cm

shape = (1, 3, 3)

data = {
    "density": np.arange(np.product(shape)).reshape(shape) * g / cm**3,
}

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.1, 1], [0.7, np.pi - 0.7], [0, 2 * np.pi]]),
    geometry="spherical",
)

p = yt.SlicePlot(ds, "r", ("gas", "density"))
p.set_log("all", False)
p.save("/tmp/res.png")
```

### on main
res

### this branch
res


yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -670,6 +670,21 @@ def pixelize_aitoff(np.float64_t[:] azimuth,
         ymax = fmax(ymax, y)
         # Now we have the (projected rectangular) bounds.
 
+        # special cases where the projection of the cell isn't
+        # bounded by the rectangle (in image space) that bounds its corners.
+        # We take a generous margin of error by making the rectangle 3 times as
+        # large (or tall), which is enough to accomodate even extremely coarse grids
+        if xmin * xmax < 0.0:
```
matthewturk, line 677: I think I agree with this change, but a question: would we have this issue if we did everything in coords from 0..180 and 0..360?

matthewturk (approved):

yt/utilities/lib/pixelization_routines.pyx
neutrinoceros: Yes. The algorithm is the problem, not the coordinate system. In fact the coordinates we use here are actually more convenient to workaround these special cases.

neutrinoceros:
switching to draft because this is still brittle. With `shape = (1, 500, 1)`
res

neutrinoceros:
I found a more robust solution that can handle the case above, albeit at the cost of a significant performance hit (the script takes 21s instead of 6s with a finer grid).
res

neutrinoceros:
Here's a condensed script to check many cases in one go.
```python
import numpy as np
import yt
from yt.units import g, cm

for shape in (
    (1, 1, 3),
    (1, 3, 1),
    (1, 3, 3),
    (1, 1, 4),
    (1, 4, 1),
    (1, 4, 4),
    (1, 200, 1),
    (1, 1, 200),
):
    data = {
        "density": np.arange(np.product(shape)).reshape(shape) * g / cm**3,
    }

    ds = yt.load_uniform_grid(
        data,
        shape,
        bbox=np.array([[0.1, 1], [0, np.pi], [0, 2 * np.pi]]),
        geometry="spherical",
    )

    p = yt.SlicePlot(ds, "r", ("gas", "density"))
    p.set_log("all", False)
    p.save(f"/tmp/{'_'.join(map(str, shape))}.png")
```

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -668,6 +668,26 @@ def pixelize_aitoff(np.float64_t[:] azimuth,
         xmax = fmax(xmax, x)
         ymin = fmin(ymin, y)
         ymax = fmax(ymax, y)
+        # special cases where the projection of the cell isn't
+        # bounded by the rectangle (in image space) that bounds its corners.
+        # Note that performance may take a serious hit here. The overarching algorithm
```
matthewturk, line 673: Just to follow up on this statement, do you think that in the typical, previous optimization case, this will be a problem?

matthewturk (approved):
This looks good to me, and I appreciate the fix!

yt/utilities/lib/pixelization_routines.pyx
neutrinoceros: No, I couldn't measure any difference in performance for the more typical use case.

neutrinoceros:
@matthewturk anything left to resolve before we can merge ?

PR #3829, opened 2022-02-28
BUG: fix a bug in SliceSelector where domain right edge was excluded from any selection

### LINKED ISSUE #3828: BUG: Selecting the domain's right edge doesn't work
### Bug report

**Bug summary**
This simple script demonstrates that making a SlicePlot exactly on the left edge works, but not on the right edge.
I assume this is reminiscent of a bias somewhere in selection routines, but right now I don't have a clear idea where that would be.
I'm hopeful that fixing this could be a simple matter of changing a `>` into a `>=` or something similar, the challenge lies in finding where.

**Code for reproduction**

```python
import yt

ds = fake_amr_ds(fields=[("gas", "density")], units=["g/cm**3"])

axis = 2
left = ds.domain_center.copy()
left[axis] = ds.domain_left_edge[axis]
right = ds.domain_center.copy()
right[axis] = ds.domain_right_edge[axis]

for name, center in (("left", left), ("right", right)):
    p = yt.SlicePlot(ds, axis, "density", center=center)
    p.save(f"/tmp/center_{name}.png")
```

**Actual outcome**
left edge ✅ 
center_left

right edge ❌
center_right

For other classes of datasets, this can fail in a much worse fashion, for instance using `ds = yt.load_sample("IsolatedGalaxy")`, we get a crash

<details> <summary> Traceback </summary>

```python-traceback
  File "t_center.py", line 16, in <module>
    p = yt.SlicePlot(ds, axis, "density", center=center)
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/plot_window.py", line 1972, in __init__
    slc.get_data(fields)
  File "/Users/robcleme/dev/yt-project/yt/yt/data_objects/selection_objects/data_selection_objects.py", line 196, in get_data
    read_fluids, gen_fluids = self.index._read_fluid_fields(
  File "/Users/robcleme/dev/yt-project/yt/yt/geometry/geometry_handler.py", line 226, in _read_fluid_fields
    fields_to_return = self.io._read_fluid_selection(
  File "/Users/robcleme/dev/yt-project/yt/yt/utilities/io_handler.py", line 128, in _read_fluid_selection
    for field, obj, data in self.io_iter(chunks, fields):
  File "/Users/robcleme/dev/yt-project/yt/yt/frontends/enzo/io.py", line 149, in io_iter
    if fid is not None:
UnboundLocalError: local variable 'fid' referenced before assignment
```
</details>

neutrinoceros:
## PR Summary
Fixes #3828
Both examples I gave in the issue now work as expected (fake data and `"IsolatedGalaxy"`)
Note: I'm note sure which of these two lines _needs_ a change to fix the issue I'm addressing here, but I'm assuming they both need to be kept in sync.

### DIFF
--- a/yt/geometry/_selection_routines/slice_selector.pxi
+++ b/yt/geometry/_selection_routines/slice_selector.pxi
@@ -97,7 +97,7 @@ cdef class SliceSelector(SelectorObject):
                                np.float64_t right_edge[3]) nogil:
         if self.reduced_dimensionality == 1:
             return 1
-        if left_edge[self.axis] - grid_eps <= self.coord < right_edge[self.axis]:
+        if left_edge[self.axis] - grid_eps <= self.coord <= right_edge[self.axis]:
             return 1
         return 0
 
@@ -108,7 +108,7 @@ cdef class SliceSelector(SelectorObject):
                                np.float64_t right_edge[3]) nogil:
         if self.reduced_dimensionality == 1:
             return 2
-        if left_edge[self.axis] - grid_eps <= self.coord < right_edge[self.axis]:
+        if left_edge[self.axis] - grid_eps <= self.coord <= right_edge[self.axis]:
             return 2 # a box with non-zero volume can't be inside a plane
         return 0
 
--- a/yt/geometry/_selection_routines/slice_selector.pxi
+++ b/yt/geometry/_selection_routines/slice_selector.pxi
@@ -108,7 +108,7 @@ cdef class SliceSelector(SelectorObject):
                                np.float64_t right_edge[3]) nogil:
         if self.reduced_dimensionality == 1:
             return 2
-        if left_edge[self.axis] - grid_eps <= self.coord <= right_edge[self.axis]:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
I think that failures are showing that this change breaks projections because some cells are suddenly contributing twice.
I don't know yet how and if this can be addressed. I'll keep this PR in the bugfix milestone for now but if it's still halted when we choose to do the release, it's perfectly okay to remove it.

matthewturk:
Some cells living in between blocks?  Can you describe what you mean by that?

neutrinoceros:
Yup, poor choice of words. What I actually meant was cells that are on the edge of a block, but not on the edge of the domain.

matthewturk:
Does it change the actual icoords calculation?

neutrinoceros:
I don't think there's an easy and scalable way out of this situation; I expect that making left and right edges actually symmetric to selectors would take a disproportionate amount of effort in contrast to my initial one line diff attempt.
I also found a workaround to the issue from the user standpoint so I'll just close this.

PR #3831, opened 2022-03-02
MNT: migrate from toml to tomli + tomli_w

### LINKED ISSUE #3830: Convert to using tomli
The library currently depends on the dead "toml" library, instead of the "tomli" library (which was just accepted into the stdlib as "tomllib" for Python 3.11. It should be a really easy swap; it's almost the same API as toml, it just expects binary files instead of unicode ones if using `toml{i,lib}.load`. `loads` is the same. Writing is a separate library.

Working on pyodide at https://github.com/pyodide/pyodide/pull/2234 and don't want to have to port a dead library.

neutrinoceros:
## PR Summary
Fix #3830. Possibly not exactly the correct approach right now (I'm adding a conditional branch for Python 3.11 before it even has the feature implemented), but I'd like to see if this naive approach survives CI.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,8 @@ install_requires =
     pyyaml>=4.2b1
     setuptools>=19.6
     sympy!=1.9,>=1.2  # see https://github.com/sympy/sympy/issues/22241
-    toml>=0.10.2
+    tomli>=1.2.3
+    tomli-w>=0.4.0
     tqdm>=3.4.0
     unyt>=2.8.0
 python_requires = >=3.7,<3.12
@@ -103,6 +104,8 @@ minimal =
     matplotlib==2.2.3
     more-itertools==8.4
     numpy==1.14.5
+    tomli==1.2.3
+    tomli-w==0.4.0
     unyt==2.8.0
 test =
     codecov~=2.0.15
--- a/yt/config.py
+++ b/yt/config.py
@@ -1,7 +1,9 @@
 import os
 import warnings
 
-import toml
+# TODO: import tomllib from the standard library instead in Python >= 3.11
+import tomli as tomllib
+import tomli_w
 from more_itertools import always_iterable
 
 from yt.utilities.configuration_tree import ConfigLeaf, ConfigNode
@@ -144,14 +146,16 @@ class YTConfig:
             if not os.path.exists(fname):
                 continue
             metadata = {"source": f"file: {fname}"}
-            self.update(toml.load(fname), metadata=metadata)
+            with open(fname, "rb") as fh:
+                data = tomllib.load(fh)
+            self.update(data, metadata=metadata)
             file_names_read.append(fname)
 
         return file_names_read
 
     def write(self, file_handler):
         value = self.config_root.as_dict()
-        config_as_str = toml.dumps(value)
+        config_as_str = tomli_w.dumps(value)
 
         try:
             # Assuming file_handler has a write attribute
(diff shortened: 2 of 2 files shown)


cphyc (dismissed):
Tested locally and this works like a charm!

On a side note - I realised that if the config is malformed yt cannot load properly. Note that is not new with this PR (as of ff5e808ec4dbebad812cab82c75297d2cfcd0519 a `toml.decoder.TomlDecodeError` is raised instead). Is this something we want? For example the following
```bash
echo "! this is invalid" > yt.toml 
python -c "import yt"
```
fails with the backtrace
```backtrace
  File "<string>", line 1, in <module>
  File "/home/XXX/Documents/prog/yt/yt/__init__.py", line 13, in <module>
    import yt.utilities.physical_constants as physical_constants
  File "/home/XXX/Documents/prog/yt/yt/utilities/physical_constants.py", line 3, in <module>
    from yt.units.yt_array import YTQuantity
  File "/home/XXX/Documents/prog/yt/yt/units/yt_array.py", line 3, in <module>
    from yt.funcs import array_like_field  # NOQA: F401
  File "/home/XXX/Documents/prog/yt/yt/funcs.py", line 33, in <module>
  File "/home/XXX/Documents/prog/yt/yt/utilities/logger.py", line 4, in <module>
    from yt.config import ytcfg
  File "/home/XXX/Documents/prog/yt/yt/config.py", line 215, in <module>
    ytcfg.read(_local_config_file)
  File "/home/XXX/Documents/prog/yt/yt/config.py", line 154, in read
    data = tomllib.load(fh)
  File "/home/XXX/anaconda3/envs/py39/lib/python3.9/site-packages/tomli/_parser.py", line 66, in load
    return loads(s, parse_float=parse_float)
  File "/home/XXX/anaconda3/envs/py39/lib/python3.9/site-packages/tomli/_parser.py", line 116, in loads
    raise suffixed_err(src, pos, "Invalid statement")
tomli.TOMLDecodeError: Invalid statement (at line 1, column 1)
```

setup.cfg
```diff
@@ -47,9 +47,10 @@ install_requires =
     pyyaml>=4.2b1
     setuptools>=19.6
     sympy!=1.9,>=1.2  # see https://github.com/sympy/sympy/issues/22241
-    toml>=0.10.2
+    tomli-w>=0.4.0
     tqdm>=3.4.0
     unyt>=2.8.0
+    tomli>=1.2.3;python_version < '3.11'
```
neutrinoceros, line 53: for the record, I picked the most recent versions of each package that had support for Python 3.6 as minimal, to make backporting to yt 4.0.x easier

neutrinoceros:
Sorry for pushing again, I changed my mind about the 3.11 branch. I think it's preferable to keep one robust and testable solution for now, and make tomli a dependency for Python <= 3.10 only when Python 3.11 becomes testable to us (probably sometimes this summer).

cphyc (approved):

yt/config.py
```diff
@@ -1,7 +1,9 @@
 import os
 import warnings
 
-import toml
+# TODO: import tomllib from the standard library instead in Python >= 3.11
```
matthewturk, line 4: This is likely going to break some developer installations where folks do an in-place without dependency checks, so we should send a note to the mailing list.

matthewturk (approved):

yt/config.py
neutrinoceros: You got it.

PR #3832, opened 2022-03-03
MNT: remove pyyaml and IPython from install_requires as they are both de facto optional dependencies

neutrinoceros:
## PR Summary
Follow up to the discussion in #3830, let's experiment with making IPython actually optional as a dependency.
I expect CI could fail at first because we haven't been using on_demand_import.py for IPython.

I note that there's also something weird with yaml support, where we currently require `pyyaml` in `install_deps` but `pyyaml` in the optional `full` environment.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,13 +38,11 @@ project_urls =
 packages = find:
 install_requires =
     cmyt>=0.2.2
-    ipython>=2.0.0
     matplotlib!=3.4.2,>=2.2.3  # keep in sync with tests/windows_conda_requirements.txt
     more-itertools>=8.4
     numpy>=1.14.5
     packaging>=20.9
     pyparsing>0.0  # hard dependency to MPL. We require it (unconstrained) in case MPL drops it in the future
-    pyyaml>=4.2b1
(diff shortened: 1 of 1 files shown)


matthewturk:
We get some stuff as transitive dependencies from ipython, don't we?

neutrinoceros:
I hope not, but let's see what CI has to say about it :-)

matthewturk:
Before we merge, can we verify -- pyyaml *was* used for tracking the shaders in what is now `yt_idv` and also for generation of the FEM functions, right?  But neither is necessary for standard installations?

neutrinoceros:
I went as far back as yt 2.0 and the only use case (outside testing) I found is `yt/utilities/mesh_code_generation.py`, which survived to this day but doesn't actually looks to be used anywhere. Actually it seems to be a script rather than a proper yt module. I'll revert my change to that file.

matthewturk:
That file we should keep -- it's exceedingly useful!  We just don't have occasion to use it that often.

neutrinoceros:
Dully noted.

matthewturk (approved):

PR #3834, opened 2022-03-06
BUG: tentative fix for #2538

### LINKED ISSUE #2538: BUG: combining set_zlim and set_unit methods doesn't work as expected
### Bug report

**Bug summary**

zticks are improperly set when combining `plot.set_zlim` and `plot.set_unit`, resulting in a useless plot.

**Code for reproduction**
```python
import yt
ds = fake_random_ds(128)
zmax = ds.find_max("density")[0]

p.set_zlim("density", zmax/100, zmax)
p.save("/tmp/s1")

p.set_unit("density", "g/AU**3")
p.save("/tmp/s2")
```
**Actual outcome**

s11_Slice_z_density
s12_Slice_z_density

note that the final outcome is identical with the reversed order of operations.

neutrinoceros:
## PR Summary
Fix #2538
So the hard part in fixing this (to me) is to reason about the many interacting classes here.
It seems to me that the easiest way to reconciliate `.set_zlim` and `.set_unit` plot methods in to store `zmin` and `zmax` as unyt_quantity objects instead of floats.
Until now I haven't been able to figure out a way to do that without changing the public API slightly, though my approach should actually make it more flexible as it allows `zmin` or `zmax` to be left unset in `set_zlim`, while on main, both are required.
I'm also not sure how this is going to behave with `PhasePlot` (which descend from `ImagePlotContainer` but don't have a frb attribute), but this needs to be adressed and tested.
For now I'll open as a draft to see if I'm breaking anything that's already exercised in CI.

### DIFF
--- a/yt/visualization/plot_container.py
+++ b/yt/visualization/plot_container.py
@@ -964,7 +964,7 @@ class ImagePlotContainer(PlotContainer):
 
     @accepts_all_fields
     @invalidate_plot
-    def set_zlim(self, field, zmin, zmax, dynamic_range=None):
+    def set_zlim(self, field, zmin=None, zmax=None, dynamic_range=None):
         """set the scale of the colormap
 
         Parameters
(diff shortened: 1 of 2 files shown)


neutrinoceros:
For reference here's the new result from #2538 's demo script
```python
import yt
ds = fake_random_ds(128)
zmax = ds.find_max("density")[0]

p.set_zlim("density", zmax/100, zmax)
p.save("/tmp/s1")

p.set_unit("density", "g/AU**3")
p.save("/tmp/s2")
```
s1_Slice_z_density
s2_Slice_z_density
That said it only works in that order. If I call `set_unit` first it's still broken...

neutrinoceros:
This won't go anywhere, I think a broader change is needed. I'm thinking of adding a small "bookkeeper" object to hold all coupled colorbar parameters such as vmin, vmax, display units and norm. That'll be a different PR though, as I'm not sure when I'll have time to try it out.

PR #3836, opened 2022-03-07
DEPR: add missing since/removal data to deprecation warnings in plot callbacks

neutrinoceros:
## PR Summary
Based on to of #3835 for stability


neutrinoceros:
This should now fix https://github.com/yt-project/yt/issues/3853

chrishavlin (approved):
excellent!

yt/data_objects/tests/test_image_array.py
```diff
@@ -117,8 +117,9 @@ def test_image_array_rgba_png(self):
             warnings.simplefilter("always")
             # Trigger a warning.
             im_arr.write_png("clipped.png", clip_ratio=0.5)
-            assert str(w[0].message) == (
-                "'clip_ratio' keyword is deprecated. Use 'sigma_clip' instead"
+            assert str(w[0].message).startswith(
+                "The 'clip_ratio' keyword argument is a deprecated alias for 'sigma_clip'. "
+                "Please use 'sigma_clip' directly."
```
stonnes, line 122: Should you add the two lines 
since = "3.3"
removal = "4.2" 
here?

stonnes (commented):
Looks good, maybe just missed one since/removal pair that I noted!

yt/data_objects/tests/test_image_array.py
neutrinoceros: Well here I'm matching only the beginning of the deprecation message because I'd like to avoid matching the rest of it, which is controlled by the common function `issue_deprecation_warning` ("since" and "removal" in particular are formatted by that function).

stonnes (approved):

neutrinoceros:
Thank you both for reviewing this ! @chrishavlin, should we merge it ?

chrishavlin:
Yes we should! I'll hit the button...

PR #3841, opened 2022-03-10
Add missing ionized species

cphyc:
## PR Summary

With recent version of RAMSES, which field are present on-disk is documented by a side-car file named `hydro_file_descriptor.txt` which lists which field is which.

Confusingly, yt expects variables to use the old yt convention (e.g. `HII` is ionized hydrogen), while RAMSES moved to follow yt's convention in `hydro_file_descriptor.txt`! The long-term solution would be to migrate everything to yt's convention (`H_p1_fraction` for example) but it would require refactoring a lot of code in the RAMSES frontend.

Instead, I map here the variables read from the file descriptor to the "old" convention used throughout the RAMSES handler, which then maps back to the new yt convention... In other words, yt reads that the n-th variable is `H_p1_fraction`, it reads it into `("ramses", "HII")` which is then aliased to `("gas", "H_p1_fraction")`.

### DIFF
--- a/yt/frontends/ramses/io.py
+++ b/yt/frontends/ramses/io.py
@@ -1,4 +1,6 @@
 from collections import defaultdict
+from pathlib import Path
+from typing import Union
 
 import numpy as np
 
@@ -209,13 +211,13 @@ class IOHandlerRAMSES(BaseIOHandler):
         return tr
 
 
-def _read_part_file_descriptor(fname):
+def _read_part_file_descriptor(fname: Union[str, Path]):
     """
     Read a file descriptor and returns the array of the fields found.
     """
 
     # Mapping
-    mapping = [
+    mapping_list = [
         ("position_x", "particle_position_x"),
         ("position_y", "particle_position_y"),
         ("position_z", "particle_position_z"),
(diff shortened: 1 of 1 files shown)


yt/frontends/ramses/io.py
```diff
@@ -209,13 +211,13 @@ def _read_particle_subset(self, subset, fields):
         return tr
 
 
-def _read_part_file_descriptor(fname):
+def _read_part_file_descriptor(fname: Union[str, Path]):
```
neutrinoceros, line 214:
```suggestion
def _read_part_file_descriptor(fname: Union[str, "os.PathLike[str]"]):
```

yt/frontends/ramses/io.py
```diff
@@ -265,31 +267,35 @@ def _read_part_file_descriptor(fname):
     return fields
 
 
-def _read_fluid_file_descriptor(fname):
+def _read_fluid_file_descriptor(fname: Union[str, Path]):
```
neutrinoceros, line 270:
```suggestion
def _read_fluid_file_descriptor(fname: Union[str, "os.PathLike[str]"]):
```

yt/frontends/ramses/io.py
```diff
@@ -1,4 +1,6 @@
 from collections import defaultdict
+from pathlib import Path
```
neutrinoceros, line 2: See following comments
```suggestion
import os
```

neutrinoceros (commented):
Other than a couple type hint nits, I would suggest that you add a comment somewhere that explains the history of this code as you did in the PR description.

neutrinoceros:
Don't hesitate to add this to the 4.0.3 milestone if appropriate.

neutrinoceros (approved):

PR #3847, opened 2022-03-14
remove unneeded logs using private API

### LINKED ISSUE #3845: DOC: private methods used in cookbook recipe
### Bug report

**Bug summary**

in #2567, `doc/source/cookbook/render_two_fields_tf.py` was added to the cookbook
On the two following lines, the private method `_get_field_info` is used
https://github.com/yt-project/yt/blob/ef08989d8222e7172751d17009959102db05dd20/doc/source/cookbook/render_two_fields_tf.py#L15
https://github.com/yt-project/yt/blob/ef08989d8222e7172751d17009959102db05dd20/doc/source/cookbook/render_two_fields_tf.py#L30

This should be refactored to avoid showing private elements
@zingale, do you know how we could do this ?

zingale:
## PR Summary

this closes #3845 

## PR Checklist

### DIFF
--- a/doc/source/cookbook/render_two_fields_tf.py
+++ b/doc/source/cookbook/render_two_fields_tf.py
@@ -12,7 +12,6 @@ sc = Scene()
 # Add density
 
 field = "density"
-ds._get_field_info(field).take_log = True
 
 vol = create_volume_source(ds, field=field)
 vol.use_ghost_zones = True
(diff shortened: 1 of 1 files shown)


neutrinoceros (approved):
Thanks ! FTR I agree with you that we should show how to do this using only the public API, or actually implement a way to do it if there isn't.

neutrinoceros:
Ah, this is stuck because auto-merge is waiting for a job that will not be triggered for a pure docs patch. I think only an owner can merge this. Ping @matthewturk

PR #3856, opened 2022-03-21
BUG: fix buffer masking in cartesian geometry

### LINKED ISSUE #3854: BUG: inconsistency in cartesian buffer masking
### Bug report

**Bug summary**

There is a discrepancy in what fill value is used in buffers depending on the pixelisation routine. This is apparent e.g. in CylindricalCoordinateHandler:

when cartesian pixelisation is used:
https://github.com/yt-project/yt/blob/0d846025f2bd17d1f8e6c57d21973c384dee0457/yt/geometry/coordinates/cylindrical_coordinates.py#L128

and when cylindrical pixelisation is used:
https://github.com/yt-project/yt/blob/0d846025f2bd17d1f8e6c57d21973c384dee0457/yt/geometry/coordinates/cylindrical_coordinates.py#L144

See https://github.com/yt-project/yt/issues/2480 and https://github.com/yt-project/yt/pull/2517 for historical context.
In short, we switched to NaN as a fill value to resolve issues with setting background colors.

This was not yet generalised to cartesian pixelisation because usually, every single pixel is filled with data values, leaving no room for a background color.
This is however not true in the case of the very example used in [our documentation to demonstrate how to setup a background color](https://yt-project.org/doc/visualizing/plots.html#colormaps), where we deliberately zoom out of the domain bounding box.

This example _happened_ to work until yt 4.0.2 only because the norm used there is log, but it breaks if one switches to linear scaling

```python
import yt

normal="z"
slc = yt.SlicePlot(ds, normal, ("gas", "density"), width=(1.5, "Mpc"))
slc.set_background_color(("gas", "density"), color="C0")
slc.save("/tmp/background_log")
slc.set_log(("gas", "density"), False)
slc.save("/tmp/background_lin")
```
background_log_Slice_z_density
background_lin_Slice_z_density

On the main branch, as well as on the 4.0.x branch, having zeros in the buffer implies that yt will switch to a symlog norm, which breaks even the default case. This behaviour is intentional and happened in https://github.com/yt-project/yt/pull/3793, but it should be apparent by now that this is incompatible with using 0 as a fill value

So ! I made a simple attempt using the following patch
```patch
diff --git a/yt/geometry/coordinates/cartesian_coordinates.py b/yt/geometry/coordinates/cartesian_coordinates.py
index 5930e44a2..c83bcb9b2 100644
--- a/yt/geometry/coordinates/cartesian_coordinates.py
+++ b/yt/geometry/coordinates/cartesian_coordinates.py
@@ -306,7 +306,7 @@ class CartesianCoordinateHandler(CoordinateHandler):
         if hasattr(period, "in_units"):
             period = period.in_units("code_length").d

-        buff = np.zeros((size[1], size[0]), dtype="f8")
+        buff = np.full((size[1], size[0]), np.nan, dtype="float64")
         particle_datasets = (ParticleDataset, StreamParticlesDataset)
         is_sph_field = finfo.is_sph_field

@@ -555,7 +555,7 @@ class CartesianCoordinateHandler(CoordinateHandler):
         from yt.frontends.ytdata.data_structures import YTSpatialPlotDataset

         indices = np.argsort(data_source["pdx"])[::-1].astype(np.int_)
-        buff = np.zeros((size[1], size[0]), dtype="f8")
+        buff = np.full((size[1], size[0]), np.nan, dtype="float64")
         ftype = "index"
         if isinstance(data_source.ds, YTSpatialPlotDataset):
             ftype = "gas"
diff --git a/yt/geometry/coordinates/cylindrical_coordinates.py b/yt/geometry/coordinates/cylindrical_coordinates.py
index 4f34d62d2..a865ff1e0 100644
--- a/yt/geometry/coordinates/cylindrical_coordinates.py
+++ b/yt/geometry/coordinates/cylindrical_coordinates.py
@@ -125,7 +125,7 @@ class CylindricalCoordinateHandler(CoordinateHandler):
         period[1] = self.period[self.y_axis[dim]]
         if hasattr(period, "in_units"):
             period = period.in_units("code_length").d
-        buff = np.zeros(size, dtype="f8")
+        buff = np.full(size, np.nan, dtype="float64")
         pixelize_cartesian(
             buff,
             data_source["px"],
diff --git a/yt/geometry/coordinates/geographic_coordinates.py b/yt/geometry/coordinates/geographic_coordinates.py
index 38c948835..ea5371393 100644
--- a/yt/geometry/coordinates/geographic_coordinates.py
+++ b/yt/geometry/coordinates/geographic_coordinates.py
@@ -246,7 +246,7 @@ class GeographicCoordinateHandler(CoordinateHandler):
         pdx = data_source["pdx"]
         py = data_source["py"]
         pdy = data_source["pdy"]
-        buff = np.zeros((size[1], size[0]), dtype="f8")
+        buff = np.full((size[1], size[0]), np.nan, dtype="float64")
         pixelize_cartesian(
             buff,
             px,
index c52d22ab3..3a0309aab 100644
@@ -256,6 +256,8 @@ def pixelize_cartesian(np.float64_t[:,:] buff,
                                 # This will reduce artifacts if we ever move to
                                 # compositing instead of replacing bitmaps.
                                 if overlap1 * overlap2 < 1.e-6: continue
+                                # make sure pixel value is not a NaN before incrementing it
+                                if buff[i,j] != buff[i,j]: buff[i,j] = 0.0
                                 buff[i,j] += (dsp * overlap1) * overlap2
                             else:
                                 buff[i,j] = dsp
@@ -506,6 +508,8 @@ def pixelize_off_axis_cartesian(
                        fabs(zsp - cz) * 0.99 > dzsp:
                         continue
                     mask[i, j] += 1
+                    # make sure pixel value is not a NaN before incrementing it
+                    if buff[i,j] != buff[i,j]: buff[i,j] = 0.0
                     buff[i, j] += dsp
     for i in range(buff.shape[0]):
         for j in range(buff.shape[1]):
diff --git a/yt/visualization/tests/test_offaxisprojection.py b/yt/visualization/tests/test_offaxisprojection.py
index a7af54e77..cb64c2db2 100644
--- a/yt/visualization/tests/test_offaxisprojection.py
+++ b/yt/visualization/tests/test_offaxisprojection.py
@@ -3,6 +3,8 @@ import shutil
 import tempfile
 import unittest

+import numpy as np
+
 from yt.testing import (
     assert_equal,
     assert_fname,
@@ -98,5 +100,4 @@ def test_field_cut_off_axis_octree():
         (p3.frb[("gas", "density")] == p4.frb[("gas", "density")]).all(), False
     )
     p4rho = p4.frb[("gas", "density")]
-    assert_equal(p4rho.min() == 0.0, True)  # Lots of zeros
-    assert_equal(p4rho[p4rho > 0.0].min() >= 0.5, True)
+    assert_equal(np.nanmin(p4rho[p4rho > 0.0]) >= 0.5, True)
```
which indeed fixed the original problem with background colors, however this also has unexpected and undesirable side effects in other image comparison tests, so I want to have a dedicated PR for this problem, instead of jamming it in #3849.

My conclusion for now is this needs to be adressed for yt 4.0.3

neutrinoceros:
## PR Summary
Goal: fix #3854
Applying the patch from the issue. Let's see what image tests fail.

### DIFF
--- a/yt/visualization/tests/test_offaxisprojection.py
+++ b/yt/visualization/tests/test_offaxisprojection.py
@@ -3,6 +3,8 @@ import shutil
 import tempfile
 import unittest
 
+import numpy as np
+
 from yt.testing import (
     assert_equal,
     assert_fname,
@@ -98,5 +100,4 @@ def test_field_cut_off_axis_octree():
         (p3.frb[("gas", "density")] == p4.frb[("gas", "density")]).all(), False
     )
     p4rho = p4.frb[("gas", "density")]
-    assert_equal(p4rho.min() == 0.0, True)  # Lots of zeros
-    assert_equal(p4rho[p4rho > 0.0].min() >= 0.5, True)
+    assert_equal(np.nanmin(p4rho[p4rho > 0.0]) >= 0.5, True)
--- a/conftest.py
+++ b/conftest.py
@@ -5,6 +5,7 @@ from importlib.util import find_spec
 from pathlib import Path
 
 import matplotlib
+import numpy
 import pytest
 import yaml
 from packaging.version import Version
@@ -20,6 +21,7 @@ from yt.utilities.answer_testing.testing_utilities import (
 )
 
 MPL_VERSION = Version(matplotlib.__version__)
+NUMPY_VERSION = Version(numpy.__version__)
 
 
 def pytest_addoption(parser):
@@ -120,6 +122,15 @@ def pytest_configure(config):
                 "ignore:Using or importing the ABCs from 'collections' instead of from 'collections.abc' "
                 "is deprecated since Python 3.3,and in 3.9 it will stop working:DeprecationWarning"
             ),
+        )
+
+    if NUMPY_VERSION < Version("1.19") and MPL_VERSION < Version("3.3"):
+        # This warning is triggered from matplotlib in exactly one test at the time of writing
+        # and exclusively on the minimal test env. Upgrading numpy or matplotlib resolves
+        # the issue, so we can afford to ignore it.
+        config.addinivalue_line(
+            "filterwarnings",
+            "ignore:invalid value encountered in less_equal:RuntimeWarning",
         )
 
     if find_spec("astropy") is not None:
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -42,7 +42,7 @@ answer_tests:
     - yt/frontends/chombo/tests/test_outputs.py:test_zp
     - yt/frontends/chombo/tests/test_outputs.py:test_kho
 
-  local_enzo_008:  # PR 2926
+  local_enzo_009:  # PR 3856
     - yt/frontends/enzo/tests/test_outputs.py:test_moving7
     - yt/frontends/enzo/tests/test_outputs.py:test_galaxy0030
     - yt/frontends/enzo/tests/test_outputs.py:test_toro1d
@@ -50,7 +50,7 @@ answer_tests:
     - yt/frontends/enzo/tests/test_outputs.py:test_ecp
     - yt/frontends/enzo/tests/test_outputs.py:test_nuclei_density_fields
 
-  local_enzo_e_001: # PR 3273
+  local_enzo_e_002: # PR 3856
     - yt/frontends/enzo_e/tests/test_outputs.py:test_hello_world
     - yt/frontends/enzo_e/tests/test_outputs.py:test_particle_fields
 
@@ -152,7 +152,7 @@ answer_tests:
     - yt/frontends/boxlib/tests/test_outputs.py:test_NyxDataset
     - yt/frontends/boxlib/tests/test_outputs.py:test_WarpXDataset
 
-  local_ramses_004:
+  local_ramses_005:  # PR 3856
     - yt/frontends/ramses/tests/test_outputs.py:test_output_00080
 
   local_ytdata_008:
--- a/yt/geometry/coordinates/cartesian_coordinates.py
+++ b/yt/geometry/coordinates/cartesian_coordinates.py
@@ -306,7 +306,7 @@ class CartesianCoordinateHandler(CoordinateHandler):
         if hasattr(period, "in_units"):
             period = period.in_units("code_length").d
 
-        buff = np.zeros((size[1], size[0]), dtype="f8")
+        buff = np.full((size[1], size[0]), np.nan, dtype="float64")
         particle_datasets = (ParticleDataset, StreamParticlesDataset)
         is_sph_field = finfo.is_sph_field
 
@@ -555,7 +555,7 @@ class CartesianCoordinateHandler(CoordinateHandler):
         from yt.frontends.ytdata.data_structures import YTSpatialPlotDataset
 
         indices = np.argsort(data_source["pdx"])[::-1].astype(np.int_)
-        buff = np.zeros((size[1], size[0]), dtype="f8")
+        buff = np.full((size[1], size[0]), np.nan, dtype="float64")
         ftype = "index"
         if isinstance(data_source.ds, YTSpatialPlotDataset):
             ftype = "gas"
--- a/yt/geometry/coordinates/cylindrical_coordinates.py
+++ b/yt/geometry/coordinates/cylindrical_coordinates.py
@@ -125,7 +125,7 @@ class CylindricalCoordinateHandler(CoordinateHandler):
         period[1] = self.period[self.y_axis[dim]]
         if hasattr(period, "in_units"):
             period = period.in_units("code_length").d
-        buff = np.zeros(size, dtype="f8")
+        buff = np.full(size, np.nan, dtype="float64")
         pixelize_cartesian(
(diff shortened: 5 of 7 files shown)


neutrinoceros:
So far I haven't been able to localise the source of differences in the affected `PixelizedProjectionValuesTest`s
I'm trying to reproduce something similar without the testing framework.

@matthewturk, any idea what might be causing this ?

reference
tmponei_20c
this branch
tmpw2pl575s

matthewturk:
Can you do a floating point comparison?  I'm not entirely sure why the masking would cause these things, *except* that it's possible that some things that were previously 0.0 are now NaN right at the boundary, or vice versa.

neutrinoceros:
So I was able to reproduce this locally by running the following on the tip of main
```python
import yt
import numpy as np
from matplotlib.colors import LogNorm

field = ("gas", "temperature")
ds = yt.load("output_00080/info_00080.txt")
obj = ds.sphere("max", (0.1, "unitary"))
axis = 0
weight_field  = ("gas", "density")
proj = ds.proj(field, axis, weight_field=weight_field, data_source=obj)
frb = proj.to_frb((1.0, "unitary"), 256)
np.save(f"main_weightfield_{weight_field}.npy", frb[field].d)
```
followed by
```
git checkout cartesian_buffer_masking
pip install -e .
```
Then repeating the script but saving to a different file
```python
np.save(f"branch_weightfield_{weight_field}.npy", frb[field].d)
```

and finally running
```python
fig, axs = plt.subplots(nrows=1, ncols=3, figsize=(15, 4))
d1 = np.load(f"main_weightfield_('gas', 'density').npy")
d2 = np.load(f"branch_weightfield_('gas', 'density').npy")

n1 = LogNorm(vmin=np.nanmin(d1), vmax=np.nanmax(d1))
n2 = LogNorm(vmin=np.nanmin(d2), vmax=np.nanmax(d2))
diff = np.nansum((d2, -d1), axis=0)
for data, norm, ax in [
    (d1, n1, axs[0]),
    (d2, n2, axs[1]),
    (diff, LogNorm(), axs[2]),
]:
    im = ax.imshow(data, norm=norm, cmap="cmyt.arbre")
    fig.colorbar(im, ax=ax)
fig.savefig("/tmp/bench.png")
``` 
bench

So it seems that your intuition was right Matt, the only place where there are actual differences is at the boundary of the selection, though I still do not understand why that is.
In any case it seems like more pixels get -correctly ?- colorised with this branch. So as far as I understand this looks more like a fix than a regression. I'm just not satisfied with my limited understanding of this difference.
@matthewturk should we move forward with it ? There are other places where buffers get initialised with zero as a filler, I probably should update them too.

matthewturk:
I agree with your assertion that this is a fix.

My suspicion is that it involves the anti-aliasing argument, and that if it were turned off we would see less difference.  Regardless, this looks like an improvement to me, and I say move forward.

neutrinoceros:
So, my attempt to generalise using nan as a fill value to sph-like particle pixelization routines has resulted in multiple test failures _that I can't seem to reproduce locally_.
As a result, I anticipate that progressing any further in that direction is going to take much more time than I can afford at the moment. @matthewturk, do you think we could get this merged with _just_ the grid/cartesian changes that we already discussed ?

matthewturk:
@neutrinoceros Yes, I think that's reasonable.

neutrinoceros:
@matthewturk this is ready

yt/visualization/tests/test_offaxisprojection.py
```diff
@@ -98,5 +100,4 @@ def test_field_cut_off_axis_octree():
         (p3.frb[("gas", "density")] == p4.frb[("gas", "density")]).all(), False
     )
     p4rho = p4.frb[("gas", "density")]
-    assert_equal(p4rho.min() == 0.0, True)  # Lots of zeros
-    assert_equal(p4rho[p4rho > 0.0].min() >= 0.5, True)
+    assert_equal(np.nanmin(p4rho[p4rho > 0.0]) >= 0.5, True)
```
cphyc, line 103: Why not use the following?
```suggestion
    assert np.nanmin(p4rho[p4rho > 0.0]) >= 0.5
```
Note that I am absolutely happy about the PR as is and don't want to delay it any further, so feel free to push the merge button!

cphyc (approved):

yt/visualization/tests/test_offaxisprojection.py
neutrinoceros: I think I’m going to keep it as is to match the local dominant convention, even though this is actually not a unit test-based test, so your suggestion should indeed work.

matthewturk:
@neutrinoceros did you just merge your own PR, before @cphyc responded to your reply?

cphyc:
In my comment I wrote

> Note that I am absolutely happy about the PR as is and don't want to delay it any further, so feel free to push the merge button!

So that's absolutely fine by me :)

31 Mar 2022 12:00:51 Matthew Turk ***@***.***>:

matthewturk:
@cphyc I deleted my comment right after I posted, because I realized

PR #3863, opened 2022-03-28
BUG: fix a bug where PlotWindow origin parameter wasn't handled correctly in non-cartesian geometries

### LINKED ISSUE #3862: BUG: to_pw ignores geometry specific rules ?
### Bug report

**Bug summary**
In spherical geometry (and probably other non-cartesian geometries too), creating a plot via `ds.slice(...).to_pw` produces inconsistent results as compared by the supposedly equivalent plot done with `yt.SlicePlot`

**Code for reproduction**

```python
import numpy as np
import yt

ds = fake_amr_ds(geometry="spherical")
field = ("stream", "Density")

yt.SlicePlot(ds, "phi", field).save("/tmp/method1.png")

slc = ds.slice("phi", np.pi)
plot = slc.to_pw(field)
plot.save("/tmp/method2.png")
```

**Actual outcome**

method1
method2

Note how the horizontal axis labels are incorrect in the second case
This was originally discovered and reported by @vdbma

neutrinoceros:
## PR Summary
fix #3862 by moving the responsibility of switching to `origin="native"` from `AxisAlignedSlicePlot` an `AxisAlignedProjectionPlot` to `PlotWindow`. Note that I'm not making any change to `OffAxis...Plot` classes, because they don't have support for non-cartesian geometries anyway, and they also don't expose the origin keyword argument.

### DIFF
--- a/yt/data_objects/selection_objects/data_selection_objects.py
+++ b/yt/data_objects/selection_objects/data_selection_objects.py
@@ -536,6 +536,7 @@ class YTSelectionContainer2D(YTSelectionContainer):
             origin=origin,
             frb_generator=frb,
             plot_type=plot_type,
+            geometry=self.ds.geometry,
         )
         pw._setup_plots()
         return pw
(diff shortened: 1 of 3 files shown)


matthewturk (approved):

neutrinoceros:
@matthewturk sorry for pushing again. This should now be ready.

cphyc (approved):

neutrinoceros:
manually backported as #3878

Issue #3867, opened 2022-03-29
RAMSES inaccurate Mpc to cm conversion value


lconaboy:
### Bug report

**Bug summary**

The RAMSES frontend uses the value 3.08e24 to convert from Mpc to cm when determining the current (proper) simulation time. This is probably a hangover from when the value 3.08e24 was widely used in RAMSES itself, but modern versions of the code use the more accurate 3.0856776d+24, which agrees with `cm_per_mpc` defined in yt/utilities/physical_ratios.py.

Originally mentioned in #3864 and might affect #2098.

**Version Information**
  * Operating System: Ubuntu 18.04
  * Python Version: 3.8.5 (Anaconda)
  * yt version: 4.1.dev0 (from source)

Issue #2098, opened 2018-11-23
RAMSES current_time discrepancy


cphyc:
### Bug report

**Bug summary**

When dealing with RAMSES data, the simulation time given by `ds.current_time` is not the same as the one you get from `ds.cosmology.t_from_z(ds.current_redshift)`.

**Code for reproduction**

```python
ds = yt.load('output_00080/info_00080.txt')
tA = ds.current_time.to('Myr')
tB = ds.cosmology.t_from_z(ds.current_redshift).to('Myr')

print(tA-tB)   # Returns -22.47323433951169 Myr
```
**Expected outcome**

I would expect to get exactly the same time(s). This is probably due to https://github.com/yt-project/yt/blob/c3aae45670d38281670fcfcd1196eb741cdf86aa/yt/frontends/ramses/io.py#L28.

ngoldbaum:
Fair enough, it’s probably not necessary to have two implementations of the same thing. I think a PR making it use `t_from_z` would be accepted. Ping @markrichardson who added the function you’re pointing at.

cphyc:
It may not be _as easy_ as this, because `convert_ramses_ages` takes as input a conformal time, while `t_from_z` expects a redshift as input. And we can't just throw `convert_ramses_ages` away since it is required to convert e.g. star ages from conformal to physical time. Any thoughts on this?

ngoldbaum:
Does the cosmology calculator not have a way to go from conformal time to physical time? If it doesn't could we add it?

cphyc:
This is actually a bit tricky: the conformal-to-physical time conversion requires to integrate the expression `dtau = H0 dt / a(t)^2`, where `H0` is the Hubble parameter and `a(t)` is the expansion factor. Most of the uncertainty seem to come from the integration part.

PR #3868, opened 2022-03-29
BUG: Use cm_per_mpc instead of 3.08e24 #3867

lconaboy:
Fix for #3867.

## PR Summary

Use `cm_per_mpc` instead of 3.08e24 for conversion from Mpc to cm. Also partly addresses #2098:

```
tA = ds.current_time.to('Myr')
tB = ds.cosmology.t_from_z(ds.current_redshift).to('Myr')

print(tA-tB)
```
where `tA-tB` is now `-0.5628994599301222 Myr` where it was `-22.47323433951169 Myr`.

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -18,6 +18,7 @@ from yt.utilities.cython_fortran_utils import FortranFile as fpu
 from yt.utilities.lib.cosmology_time import friedman
 from yt.utilities.on_demand_imports import _f90nml as f90nml
 from yt.utilities.physical_constants import kb, mp
+from yt.utilities.physical_ratios import cm_per_mpc
 
 from .definitions import (
     OUTPUT_DIR_EXP,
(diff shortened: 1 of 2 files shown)


neutrinoceros:
I'll leave triage to @cphyc: don't hesitate to put this in the 4.0.3 milestone if it makes sense to you

cphyc (dismissed):
Thanks for the contribution!

lconaboy:
Looks like a test uses the old value of 3.08e24 to calculate the compared simulation time

```
 Age simu (yt old const)=   1.1192112175127416     
 Age simu (yt new const)=   1.1212743045965676    
```

and hence this commit fails. One option is to update the test values, but since this test is on an old output (probably made  before the introduction of the more accurate `Mpc2cm` constant into RAMSES via amr/constants.f90) maybe that isn't appropriate (at least not without warning people).

neutrinoceros:
Good catch. Is it possible to parse (or deduce) which version of RAMSES produced a given output file ? If not, I'd err on the side of embracing the newest version rather than the oldest, but I guess we'd need to expose some parameter to make it possible to reproduce the old behaviour.

lconaboy:
I agree with about there being no way to tell which version. Do you think this needs some kind of warning?

neutrinoceros:
If we're going to support only one convention I agree that it's best to go with the newest. Do you guys have an idea when it was switched in RAMSES ?

lconaboy:
As far as I can tell, it was introduced to `master` around November 2018, and is in the `stable_19_09` and `stable_19_10` branches.

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):
LGTM, but I'll let @cphyc push the button :)

PR #3882, opened 2022-04-03
TST: workaround upstream instabilities

neutrinoceros:
## PR Summary
CI is currently broken by a deprecation warning emmitted from Pillow via Matplotlib, that we have no control over, so let's just filter it.
See https://github.com/matplotlib/matplotlib/pull/22766 for context


neutrinoceros:
So pre-commit.ci is _also_ broken for a completely unrelated reason (see https://github.com/psf/black/issues/2964), but the problem is already fixed upstream so we just need to update black.

chrishavlin (approved):

PR #3899, opened 2022-04-19
BUG: fix compatibility with gadget 2 files using '*.gad.*' template

neutrinoceros:
## PR Summary
This fixes an issue where some Gadget2 files can't be loaded because their names aren't compatible with the filename template we have.
This problem was reported on Slack by @Gustav-Madsen. Here's an example dataset to test this on http://skun.iaa.es/SUsimulations/UchuuDR1/ShinUchuu/SamplingParticle/snapdir_070/
I was able to make a simple plot of this dataset, only downloading the first 3 files

```python
import yt

ds = yt.load("snapdir_070")
yt.ParticleProjectionPlot(ds, "z", ("all", "Mass")).save("/tmp/")
```
SU140_070_samp0p015625_Particle_z_Mass

### DIFF
--- a/yt/frontends/gadget/data_structures.py
+++ b/yt/frontends/gadget/data_structures.py
@@ -427,7 +427,15 @@ class GadgetDataset(SPHDataset):
         )
 
         if hvals["NumFiles"] > 1:
-            self.filename_template = f"{prefix}.%(num)s{self._suffix}"
+            for t in (
+                f"{prefix}.%(num)s{self._suffix}",
+                f"{prefix}.gad.%(num)s{self._suffix}",
+            ):
+                if os.path.isfile(t % {"num": 0}):
+                    self.filename_template = t
+                    break
+            else:
+                raise RuntimeError("Could not determine correct data file template.")
         else:
             self.filename_template = self.parameter_filename
 


neutrinoceros:
> GadgetHDF5Dataset._set_owls_eagle does something similar with filename_template -- is it worth moving your change into its own method at the GadgetDataset level (e.g., GadgetDataset.set_filename_template()) to use in both spots?

I wasn't aware ! Yes, I think this is worth including in the present PR, thank you !

> do you think there's any benefit to adding an explicit way to set the filename_template (maybe a kwarg)?

Clearly it would have a benefit that users that need such patches wouldn't need to either wait for a new release or install from source. On the other hand such changes should hopefully be extremely rare so I don't think it's worth the additional maintenance weight.

neutrinoceros:
I spoke too soon. I don't feel confortable to change the `_set_owls_eagle` method after all
1) I'm not sure such a change is needed or makes sense
2) I don't think the `_set_owls_eagle` method belongs in the Gadget frontend in the first place (should probably be in `owls`, or `eagle` !).

Any way I don't understand these formats well enough to make further changes in that direction.

chrishavlin:
Ya, totally fair. I obviously don't fully understand these formats either :) These are also simple enough to fix down the line if there is need. I'll go ahead and merge since this is a bug fix.

PR #3900, opened 2022-04-20
RFC: a more robust plot norm/colorbar API (companion #3849)

neutrinoceros:
## PR Summary

This is a companion to #3849, it's not meant to be merged by itself.
It contains _only_ new image tests, without any of the code changes needed to make them pass.
I want to use this mostly as a sanity check so I can visualize the differences with the main branch are what I expect, it might also be helpful to reviewers.


neutrinoceros:
I'm seeing only a couple of different errors (there should be much more), reported 6 times each, so clearly I've messed something up.

PR #3902, opened 2022-04-22
BUG: deactivate antialising in pixelize_cartesian when data is uniform

### LINKED ISSUE #3895: BUG: antialiasing produces artefacts with uniform data
### Bug report

**Bug summary**

By default, plotting a field with a single uniform value often produce noisy patterns. I assume is because the buffer image isn't _exactly_ uniform unless it aligns with the data grid, so some pixels are interpolated and subject to floating point errors.
Furthermore, since we use the buffer's extrema (rather than the data's) as the colorbar min and max, this floating point noise, however insignificant, ends up dominating the visualisation.

Buffer are always 800x800 pixels by default, so it's easy to get misaligned data.
This often happens when plotting initial conditions where a field is still uniform.

**Code for reproduction**

Here's a minimal example with both an aligned buffer and a misaligned one.
```python
import yt
import numpy as np

shape = (16, 16, 1)
data = {"density": np.ones(shape)}
ds = yt.load_uniform_grid(data, shape, bbox=np.array([[0, 1], [0, 1], [0, 1]]))
for s in (32, 30):
    p.set_buff_size(s)
    p.annotate_title(f"buffer size = {s}")
    p.save(f"/tmp/test_buff_{s}.png")
```

**Actual outcome**

test_buff_32
test_buff_30

**Expected outcome**

So while there _is_ a workaround in this simple case, it's not very easy to discover without a good understanding of yt's internals, and it should probably not be something that happens by default.

I think this could be solved several ways, but whatever the correct solution is, it's worth keeping in mind that we don't want to take away the ability to visualise data that's _actually_ noisy.
An approach I'm considering would be to use the data's extrema rather than the buffer's to control the color range.

neutrinoceros:
## PR Summary
Fix #3895
I'm not 100% sure that calling `np.nanmin` and `np.nanmax` isn't calling the Python API, which would make this patch potentially hurtful to performance.

### DIFF
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -128,6 +128,8 @@ def pixelize_cartesian(np.float64_t[:,:] buff,
        px.shape[0] != pdy.shape[0] or \
        px.shape[0] != data.shape[0]:
         raise YTPixelizeError("Arrays are not of correct shape.")
+    if np.nanmax(data) == np.nanmin(data):
+        antialias = 0
     xiter[0] = yiter[0] = 0
     xiterv[0] = yiterv[0] = 0.0
     # Here's a basic outline of what we're going to do here.  The xiter and


neutrinoceros:
So a very quick and unscientific benchmark showed that this indeed hurts performance in a non negligible way, adding about 10% overhead to compute a 800x800 image of a 10000x10000 data selection. This does not seem worth it.

PR #3907, opened 2022-04-27
RFC: simplify test skipper decorators

neutrinoceros:
## PR Summary
This is a small step in trying to make the testing framework smaller in size and easier to maintain with the migration from nose to pytest in mind.
The central idea is to systematically use the `unittest.SkipTest` exception to (conditionally) skip tests, because it's supported by all frameworks.

### DIFF
--- a/yt/data_objects/tests/test_glue.py
+++ b/yt/data_objects/tests/test_glue.py
@@ -8,7 +8,7 @@ def setup_func():
     ytcfg["yt", "internals", "within_testing"] = True
 
 
-@requires_module("glue.core")
+@requires_module("glue")
 @nose.with_setup(setup_func)
 def test_glue_data_object():
     ds = fake_random_ds(16)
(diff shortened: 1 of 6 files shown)


neutrinoceros:
So, there are a couple new failures on Jenkins, but I see that _all_ of the tests in questions rely on the `@requires_file` decorator _and_ have been skipped for as long as I can see from the history (most likely they have been skipped for years).

Here are the tests
yt/frontends/ytdata/tests/test_old_outputs.py
yt/fields/tests/test_sph_fields.py

I think my PR is just revealing that they are actually broken, so either the tests fell out of sync with the code base and need adjusting, or they are showing actual bugs that need fixing.
However it's out of scope of the present PR, so I'm going to open an issue and link it in actual skip decorators.

matthewturk:
I like this

matthewturk (approved):

chrishavlin (approved):

PR #3912, opened 2022-05-06
BUG: fix an upcoming incompatibility with matplotlib 3.6

### LINKED ISSUE #3911: Upcoming incompatibility with matplotlib 3.6
### Bug report

**Bug summary**

This PR was just merged upstream https://github.com/matplotlib/matplotlib/pull/22298 and breaks yt by removing a bit of the API we were using (`matplotlib.cm.cmap_d`)

This is a critical breakage, as even `import yt` will raise an `AttributeError`.

The changelog indicates that it's a deprecation cycle end but it seems like we never caught the deprecation warning in time.
It seems likely that the _right_ fix would be to move whatever code is using `cmap_d` from yt to cmyt, however I'll try to make a quick, back portable hotfix first.

neutrinoceros:
## PR Summary
Fix #3911
Since matplotlib 3.6 is likely to be available before yt 4.1.0, we should consider backporting this to 4.0.x

### DIFF
--- a/yt/visualization/color_maps.py
+++ b/yt/visualization/color_maps.py
@@ -34,8 +34,6 @@ def add_colormap(name, cdict):
     """
     # Note: this function modifies the global variable 'yt_colormaps'
     yt_colormaps[name] = cc.LinearSegmentedColormap(name, cdict, 256)
-    mcm.datad[name] = cdict
-    mcm.__dict__[name] = cdict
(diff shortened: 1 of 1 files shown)


neutrinoceros:
@matthewturk could you merge this ? I'm stil getting notified twice a day that the bleeding edge CI job failed because of this problem 🙈

PR #3914, opened 2022-05-09
ENH: Update Enzo-E particle mass/density distinction

brittonsmith:
## PR Summary

Enzo-E has been updated to include a new way of distinguishing whether particles are saved as masses or densities. This is done using the "mass" or "density" parameter in the parameter file. Previously, particle masses were to be considered as densities if the simulation was cosmological. To preserve backward compatibility, Enzo-E now uses an "is_gravitating" group name that I use to tell if the data was produced before or after this change.

## PR Checklist

### DIFF
--- a/yt/frontends/enzo_e/data_structures.py
+++ b/yt/frontends/enzo_e/data_structures.py
@@ -354,6 +354,14 @@ class EnzoEDataset(Dataset):
         if os.path.exists(lcfn):
             with open(lcfn) as lf:
                 self.parameters = libconf.load(lf)
+
+            # Determine if particle masses are actually densities by the
+            # existence of the "mass_is_mass" particles parameter.
+            mass_flag = nested_dict_get(
+                self.parameters, ("Particle", "mass_is_mass"),
+                default=None)
+            self._particle_mass_is_mass = mass_flag is not None
+
             cosmo = nested_dict_get(self.parameters, ("Physics", "cosmology"))
             if cosmo is not None:
                 self.cosmological_simulation = 1
--- a/yt/frontends/enzo_e/data_structures.py
+++ b/yt/frontends/enzo_e/data_structures.py
@@ -355,13 +355,6 @@ class EnzoEDataset(Dataset):
             with open(lcfn) as lf:
                 self.parameters = libconf.load(lf)
 
-            # Determine if particle masses are actually densities by the
-            # existence of the "mass_is_mass" particles parameter.
-            mass_flag = nested_dict_get(
-                self.parameters, ("Particle", "mass_is_mass"), default=None
-            )
-            self._particle_mass_is_mass = mass_flag is not None
-
             cosmo = nested_dict_get(self.parameters, ("Physics", "cosmology"))
             if cosmo is not None:
                 self.cosmological_simulation = 1
--- a/yt/frontends/enzo_e/fields.py
+++ b/yt/frontends/enzo_e/fields.py
@@ -128,13 +128,13 @@ class EnzoEFieldInfo(FieldInfoContainer):
         self.setup_particle_mass_field(ptype)
 
     def setup_particle_mass_field(self, ptype):
-        name = "particle_mass"
+        fname = "particle_mass"
         if ptype in self.ds.particle_unions:
-            add_union_field(self, ptype, name, "code_mass")
+            add_union_field(self, ptype, fname, "code_mass")
             return
 
         constants = nested_dict_get(
-            self.ds.parameters, ("Particle", ptype, "constants"), default=[]
+            self.ds.parameters, ("Particle", ptype, "constants"), default=()
         )
         if not constants:
             names = []
@@ -143,17 +143,33 @@ class EnzoEFieldInfo(FieldInfoContainer):
                 constants = (constants,)
             names = [c[0] for c in constants]
 
+        group_list  = nested_dict_get(
+            self.ds.parameters, ("Particle", ptype, "group_list"), default=()
+        )
+
         if "mass" in names:
-            val = constants[names.index("mass")][2]
-            val = self.ds.quan(val, self.ds.mass_unit)
-            if self.ds.cosmological_simulation:
+            # Prior to the addition of the is_gravitating group list,
+            # particle masses were stored as densities if the simulation
+            # was cosmological.
+            pfield = "mass"
+            is_density = "is_gravitating" not in group_list and \
+              self.ds.cosmological_simulation
+        elif "density" in names:
+            pfield = "density"
+            is_density = True
+        else:
+            pfield = None
+
+        if pfield is not None:
+            val = constants[names.index(pfield)][2] * self.ds.mass_unit
+            if is_density:
                 val = val / self.ds.domain_dimensions.prod()
 
             def _pmass(field, data):
                 return val * data[ptype, "particle_ones"]
 
             self.add_field(
-                (ptype, name),
+                (ptype, fname),
                 function=_pmass,
                 units="code_mass",
                 sampling_type="particle",
--- a/yt/frontends/enzo_e/fields.py
+++ b/yt/frontends/enzo_e/fields.py
@@ -133,45 +133,36 @@ class EnzoEFieldInfo(FieldInfoContainer):
             add_union_field(self, ptype, fname, "code_mass")
             return
 
+        pdict = nested_dict_get(
+            self.ds.parameters, ("Particle"), default=None
+        )
+        if pdict is None:
+            return
+
         constants = nested_dict_get(
-            self.ds.parameters, ("Particle", ptype, "constants"), default=()
+            pdict, (ptype, "constants"), default=()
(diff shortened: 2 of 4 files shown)


brittonsmith:
pre-commit.ci autofix

yt/frontends/enzo_e/fields.py
```diff
@@ -128,13 +128,13 @@ def setup_particle_fields(self, ptype, ftype="gas", num_neighbors=64):
         self.setup_particle_mass_field(ptype)
 
     def setup_particle_mass_field(self, ptype):
-        name = "particle_mass"
+        fname = "particle_mass"
         if ptype in self.ds.particle_unions:
-            add_union_field(self, ptype, name, "code_mass")
+            add_union_field(self, ptype, fname, "code_mass")
             return
```
stefanarridge, line 134: Can you clarify what is going on here? i.e., what particle types are in `self.ds.particle_unions`? And what does `add_union_fields` do here?

yt/frontends/enzo_e/fields.py
```diff
@@ -143,17 +143,34 @@ def setup_particle_mass_field(self, ptype):
...
+            self.ds.parameters, ("Particle", ptype, "group_list"), default=()
+        )
+
         if "mass" in names:
-            val = constants[names.index("mass")][2]
-            val = self.ds.quan(val, self.ds.mass_unit)
-            if self.ds.cosmological_simulation:
+            # Prior to the addition of the is_gravitating group list,
+            # particle masses were stored as densities if the simulation
+            # was cosmological.
```
stefanarridge, line 153: This is incorrect. Particle "masses" were always treated as densities, independent of whether the simulation was cosmological or not.

yt/frontends/enzo_e/fields.py
```diff
@@ -143,17 +143,34 @@ def setup_particle_mass_field(self, ptype):
...
+            pfield = "mass"
+            is_density = (
+                "is_gravitating" not in group_list and self.ds.cosmological_simulation
+            )
```
stefanarridge, line 157: The "cosmological simulation" condition should be removed here.

yt/frontends/enzo_e/fields.py
```diff
@@ -143,17 +143,34 @@ def setup_particle_mass_field(self, ptype):
                 constants = (constants,)
             names = [c[0] for c in constants]
 
+        group_list = nested_dict_get(
+            self.ds.parameters, ("Particle", ptype, "group_list"), default=()
+        )
+
         if "mass" in names:
```
stefanarridge, line 150: As far as I can tell, `names` just contains the names of the particle constants. So what happens when a particle type does not have a "mass" or "density" constant, but rather, has an "attribute array" for these quantities?

yt/frontends/enzo_e/fields.py
```diff
@@ -143,17 +143,34 @@ def setup_particle_mass_field(self, ptype):
...
+        elif "density" in names:
+            pfield = "density"
+            is_density = True
+        else:
+            pfield = None
+
+        if pfield is not None:
+            val = constants[names.index(pfield)][2] * self.ds.mass_unit
+            if is_density:
                 val = val / self.ds.domain_dimensions.prod()
```
stefanarridge, line 167: This only works if the domain width is 1 in all dimensions in code units (which is the case for cosmological simulations). In general, `val` should be multiplied by the root level cell volume.

yt/frontends/enzo_e/fields.py
brittonsmith: Sure, `setup_particle_mass_field` is going to get called for particle types that represent the combination of multiple on-disk particle types, namely the "all" particle type which will represent all particles in the dataset. `add_union_field` is responsible for adding fields represents by these combined types. The `name` to `fname` change was only because I didn't like that I was using both `name` and `names` is rather different contexts.

yt/frontends/enzo_e/fields.py
brittonsmith: Is this a possibility? If so, could you provide me with a sample dataset that does this?

yt/frontends/enzo_e/fields.py
brittonsmith: Thanks for clarifying this. This misconception dates back to the creation of this frontend, possibly when things were different in Enzo-E, or maybe just what I was able to glean from the data I had on hand. I will correct this.

yt/frontends/enzo_e/fields.py
brittonsmith: Will fix. Thanks!

yt/frontends/enzo_e/fields.py
stefanarridge: To be honest, I'm still a little confused. For a concrete example, if we have simulation data with both "dark" and "star" particles, will yt also define an "all" particle type, which will include both "dark" and "star" particles? In this case, will `self.ds.particle_unions` just contain the "all" particle type? If so, does that mean that if `setup_particle_mass_field` is called for the "all" particle type, the `return` statement on line 134 is executed and no conversion from "density" to "mass" happens?

matthewturk: `all` will include all particle fields, yes.  The fields associated with particle unions are all of the fields that can be generated for all of the particle fields in that union.  (I suppose it's moderately humorous to note that the fields accessible for a "particle union" are in fact the intersection of the fields for the components.)

yt/frontends/enzo_e/fields.py
stefanarridge: Now that I've thought about this, I'm not sure that checking for the `"is_gravitating"` particle group is the most robust solution. The reason is that it requires the user to define the `"is_gravitating"` group in the input parameter file whenever they have particles with mass, even if they are not running with gravity, which is a bit counter-intuitive. I think a better solution would be to have a hard-coded parameter called something like "mass_is_mass", which would be guaranteed to exist if the data was produced with a version of Enzo-E built after the change in definition of "mass". The question I have is: how exactly does yt read in Enzo-E parameters?

brittonsmith: Parameters are mainly read in from the .libconfig file. We could, in principle, also look for HDF5 attributes in the data files.

yt/frontends/enzo_e/fields.py
brittonsmith: Ah, the return is there because we don't need to define the field functions for the union types. We only need add the field representing the union and it will call the field functions for all the encompassing types.

yt/frontends/enzo_e/fields.py
```diff
@@ -128,33 +128,40 @@ def setup_particle_fields(self, ptype, ftype="gas", num_neighbors=64):
...
-            self.add_field(
-                (ptype, name),
-                function=_pmass,
-                units="code_mass",
-                sampling_type="particle",
+            return
+
+        if not isinstance(constants[0], tuple):
+            constants = (constants,)
+        names = [c[0] for c in constants]
```
neutrinoceros, line 146: This looks broken to me, I only assume it's not because it's not new code.
Isn't it something that could be expressed with `more_itertools.always_iterable` ?

stefanarridge (approved):

yt/frontends/enzo_e/fields.py
brittonsmith: It's not broken. The format of `constants` is meant to be a tuple consisting of multiple instances of (`name`, `type`, `value`) tuples. Unfortunately, when there is only one entry, the enclosing tuple is dropped and we end up with just the first entry. I don't think `always_iterable` will simplify that.

neutrinoceros: I see. You're right that always_iterable will not help here. However, maybe some of your response could be useful as a comment here !

yt/frontends/enzo_e/data_structures.py
```diff
@@ -354,6 +354,14 @@ def _parse_parameter_file(self):
         if os.path.exists(lcfn):
             with open(lcfn) as lf:
                 self.parameters = libconf.load(lf)
+
+            # Determine if particle masses are actually densities by the
+            # existence of the "mass_is_mass" particles parameter.
+            mass_flag = nested_dict_get(
+                self.parameters, ("Particle", "mass_is_mass"), default=None
+            )
+            self._particle_mass_is_mass = mass_flag is not None
```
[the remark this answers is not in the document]
brittonsmith: That's a fair position with respect to attributes on the dataset class, but I'd rather not have to evaluate this over and over. Enzo-E datasets can tons of blocks. I'll hang this off the io handler, which is not meant to be user facing.

yt/frontends/enzo_e/fields.py
brittonsmith: Fair enough. Will do.

brittonsmith:
pre-commit.ci autofix

matthewturk (approved):

PR #3918, opened 2022-05-10
MNT: cleanup unused workarounds for unsupported version of MPL

neutrinoceros:
## PR Summary
Simple cleanup for dead code bits
For reference, we currently support MPL>=2.2.3

There are other similar workarounds to cleanup but AFAICT they are all dealt with in #3849 so I'll avoid doing it here too.

### DIFF
--- a/yt/visualization/volume_rendering/transfer_function_helper.py
+++ b/yt/visualization/volume_rendering/transfer_function_helper.py
@@ -156,9 +156,7 @@ class TransferFunctionHelper:
             xmi, xma = np.log10(self.bounds[0]), np.log10(self.bounds[1])
         else:
             xfunc = np.linspace
-            # Need to strip units off of the bounds to avoid a recursion error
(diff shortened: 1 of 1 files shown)


yt/visualization/volume_rendering/transfer_function_helper.py
```diff
@@ -190,9 +188,8 @@ def plot(self, fn=None, profile_field=None, profile_weight=None):
                 prof[profile_field]
             except KeyError:
                 prof.add_fields([profile_field])
-            # Strip units, if any, for matplotlib 1.3.1
-            xplot = np.array(prof.x)
-            yplot = np.array(
+            xplot = prof.x
```
matthewturk, line 191: We could probably just dump `prof.x` into the call to the plotting function, but it's probably clearer to leave like this.

matthewturk (approved):
no changes needed; lgtm

PR #3919, opened 2022-05-11
BUG: fixup FieldInfoContainer.add_field

neutrinoceros:
## PR Summary
A couple fixes for this method. Most importantly fixes some broken leftovers from https://github.com/yt-project/yt/pull/3240

- BUG: fix a broken internal function call
- ENH: clarify FieldInfoContainer.add_field's signature (add missing force_override kwarg)
- ENH: avoid confusing exception chaining in sanitizer method

The bug this fixes isn't present on the backport branch so this gets triaged to 4.1

### DIFF
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -296,7 +296,9 @@ class FieldInfoContainer(dict):
             )
         return sampling_type
 
-    def add_field(self, name, function, sampling_type, **kwargs):
+    def add_field(
+        self, name, function, sampling_type, *, force_override=False, **kwargs
+    ):
         """
         Add a new field, along with supplemental metadata, to the list of
         available fields.  This respects a number of arguments, all of which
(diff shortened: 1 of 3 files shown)


matthewturk (dismissed):
Seems fine, although I'm not 100% sure of the usage of the bare asterisk in add field.

neutrinoceros:
> although I'm not 100% sure of the usage of the bare asterisk in add field.

This makes any following argument "keyword only", as in, they cannot be passed positionally. This is in line with the existing behaviour, since the argument is currently only captured as a keyword argument.

see https://peps.python.org/pep-3102/

neutrinoceros:
Just noticed they are other places that could use a very similar patch, switching to draft for a minute

matthewturk (approved):
lgtm

yt/fields/field_info_container.py
```diff
@@ -267,7 +267,7 @@ def setup_fluid_aliases(self, ftype="gas"):
                 self.alias((ftype, alias), field)
 
     @staticmethod
-    def _sanitize_sampling_type(sampling_type):
+    def _sanitize_sampling_type(sampling_type: str) -> str:
```
matthewturk, line 270: Maybe -- not in this PR, or this release cycle -- this could even be an enum!

matthewturk (approved):
LGTM, no changes needed

PR #3921, opened 2022-05-11
ENH: improve error messages when add_field receives a malformed function argument

neutrinoceros:
## PR Summary
This makes life easier for clumsy users like myself who can never remember what argument goes first in a derived field function, so you get a clear error message instead of whatever happens when the malformed function gets called with reversed arguments. Incidentally adds support and tests for passing arbitrary callables as the `function` argument.

edit: I also reworked the way `units="auto"` works to make the `dimensions` argument optional even in that case. This is an opinionated change but I think it's reasonable as long as a warning is displayed when needed.

### DIFF
--- a/nose_unit.cfg
+++ b/nose_unit.cfg
@@ -6,5 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
-ignore-files=(test_load_errors.py|test_load_sample.py|test_commons.py|test_ambiguous_fields.py|test_field_access_pytest.py|test_save.py|test_line_annotation_unit.py|test_eps_writer.py|test_registration.py|test_invalid_origin.py|test_outputs_pytest\.py|test_normal_plot_api\.py|test_load_archive\.py|test_stream_particles\.py|test_file_sanitizer\.py|test_version\.py|\test_on_demand_imports\.py)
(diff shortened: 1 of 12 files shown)


neutrinoceros:
Failures are unrelated and likely to be caused by https://github.com/pypa/pip/issues/11116

matthewturk:
> edit: I also reworked the way `units="auto"` works to make the `dimensions` argument optional even in that case. This is an opinionated change but I think it's reasonable as long as a warning is displayed when needed.

I have found the `units="auto"` functionality to be confusing; my reading of this was always that it would identify whatever units came out, and allow those.  It wouldn't precisely allow for silent failures, but it would stop validating the units, and instead be descriptive rather than prescriptive.  Is that the new behavior?

neutrinoceros:
Kinda. I'm not sure what you mean by "descriptive rather than prescriptive", and I'm not sure that I can say that it doesn't allow for silent failures: with the new behaviour, validation can be omitted completely (but you *do* get a warning)

matthewturk:
I guess what I mean is, *validation* will fail, but you still get the units that come out of the field.  i.e., it still tracks, but does not do validation against a priori values

neutrinoceros:
Yes I think that's correct.

matthewturk (approved):
I'm generally OK with this.  I have a question though: without passing judgment on them, can you estimate how many people might have their derived fields broken by this?  0, a few, or lots?

matthewturk:
I don't think that is necessary.

PR #3922, opened 2022-05-13
DEPR: make some old deprecation warnings user-visible

neutrinoceros:
## PR Summary
Found a couple deprecation warnings that use the builtin `DeprecationWarning` class (not visible by default), let's make them visible so we can actually remove the deprecated code at some point.


neutrinoceros:
Oops. Apparently I already did that in a previous PR...

PR #3923, opened 2022-05-15
MNT: improvements for GH workflows

neutrinoceros:
## PR Summary

The main goal here is to reunify the contents of our CI workflows (nose/pytest), and reduce the overall number of jobs run:
We've been using pytest for a couple years now, it seems clear that it is reliable, so I propose we only run nose for the tests that are not compatible with pytest (namely, `yield` based tests).
While I'm at it I made a couple more improvements:

- MNT: link normal CI badge to GHA page
- TST: merge pytest and nose workflows
- REL: add check-manifest to wheels publish job


.github/workflows/bleeding-edge.yaml
```diff
@@ -19,33 +22,39 @@ on:
...
+        # and Cython are known to be compatible
+        # for next update (to '3.11-dev', 2022), see https://github.com/numpy/numpy/pull/21308
         python-version: '3.10-dev'
 
     - name: Install dependencies
+      # regarding why pip 22.1 is forbidden, see https://github.com/pypa/pip/issues/11116
+      # this can be simplified when a newer version is released
       run: |
-        python -m pip install --upgrade pip
+        python -m pip install --upgrade "pip != 22.1"
```
matthewturk, line 50: Should we create an issue to track this?

.github/workflows/build-test.yaml
```diff
@@ -19,37 +21,58 @@ env:
...
         ]
-        python-version: [3.9]
+        python-version: ['3.10']
         dependencies: [full]
         tests-type: [unit]
+        test-runner: [pytest]
         include:
           - os: ubuntu-latest
-            python-version: 3.7
+            python-version: '3.7'
```
matthewturk, line 38: Will we ever need to update this?

.github/workflows/type-checking.yaml
```diff
@@ -5,31 +5,33 @@ on:
     branches:
       - main
   pull_request:
-    paths-ignore:
-      - "doc/**"
-      - README.md
+    paths:
+      - yt/**/*.py
```
matthewturk, line 9: Do we only want py, or should we also include pyx, pxi, etc?  Does mypy do anything there?

matthewturk (dismissed):
This looks like a nice set of QOL improvements.

.github/workflows/bleeding-edge.yaml
neutrinoceros: I don't think it's needed. It should work forever ™️  and can be cleaned at any point when the condition is met.

.github/workflows/build-test.yaml
neutrinoceros: yup, when we drop support for Python 3.7

.github/workflows/bleeding-edge.yaml
neutrinoceros: pip 22.1.1 is out. This patch is now longer needed, I removed it

jzuhone (approved):

neutrinoceros:
@matthewturk thank you for merging, but it appears that an update is still necessary on the repo settings side.

neutrinoceros:
to be clear, the issue is that the answer test job on GH is still marked as required by auto-merge _under it's previous name_, so all PRs will be marked as "🟡" and won't be mergeable by anyone but owners until this is dealt with 
Screenshot 2022-05-23 at 22 42 58

PR #3925, opened 2022-05-15
BLD: fix a build error on Windows

### LINKED ISSUE #3924: BLD: build fails with
### Bug report

**Bug summary**

We've had a couple instabilities in CI lately, so I'm only catching this now after disantengling the rest of them, but it seems that builds have been failing with the following error for a couple days
```
yt/utilities/lib/platform_dep.h(28): error C2169: 'log2': intrinsic function, cannot be defined
```
I don't know what changed upstream, but it doesn't seem to be numpy or Cython.

neutrinoceros:
## PR Summary
Close #3924
I am not 100% sure I know what I'm doing here.

I also added a small workaround to avoid pip 22.1, which also breaks CI. See https://github.com/pypa/pip/issues/11116, resolved upstream but not released yet.

### DIFF
--- a/tests/ci_install.sh
+++ b/tests/ci_install.sh
@@ -34,7 +34,7 @@ if [[ "${RUNNER_OS}" == "Windows" ]] && [[ ${dependencies} != "minimal" ]]; then
     # keep in sync: setup.cfg
     while read requirement; do conda install --yes $requirement; done < tests/windows_conda_requirements.txt
 else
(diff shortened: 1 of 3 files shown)


jzuhone:
If tests pass I say it's fine

jzuhone (approved):

neutrinoceros:
thanks ! Do we need another review ?

Issue #1053, opened 2015-07-29
add_field fails in Python 3


yt-fido:
Originally reported by: **astrofrog (Bitbucket: [astrofrog](https://bitbucket.org/astrofrog), GitHub: [astrofrog](https://github.com/astrofrog))**

----------------------------------------

The following example:

```python
import os
from yt import load

ds = load(os.path.join('DD0010', 'moving7_0010'))

def _dust_density(field, data):
    return data["density"].in_units('g/cm**3') * 0.01

ds.add_field("dust_density", function=_dust_density, units='g/cm**3')
```

fails in Python 3 with the following error:

```python
Traceback (most recent call last):
  File "derived.py", line 9, in <module>
    ds.add_field("dust_density", function=_dust_density, units='g/cm**3')
  File "/Users/tom/miniconda3/envs/production/lib/python3.4/site-packages/yt/data_objects/static_output.py", line 910, in add_field
    deps, _ = self.field_info.check_derived_fields([name])
  File "/Users/tom/miniconda3/envs/production/lib/python3.4/site-packages/yt/fields/field_info_container.py", line 343, in check_derived_fields
    self.ds.derived_field_list = list(sorted(dfl))
TypeError: unorderable types: str() < tuple()
```

This is with yt 3.2 in Python 3.4. The issue is that in Python 3, sorting of mixed types no longer works.

I think the short-term solution will be to add a custom key to the sorting to help it out.

----------------------------------------
- Bitbucket: https://bitbucket.org/yt_analysis/yt/issue/1053

PR #3928, opened 2022-05-17
MNT: revert a patch that was written in support to a now unsupported use case

neutrinoceros:
## PR Summary
revert 2ffd0a246c83dd846f988eefbeb38e37d18c0f24
for context, this patch was intended to answer https://github.com/yt-project/yt/issues/1053
but the example given in the issue is currently invalid and fails with
```python-traceback
TypeError: Expected a 2-tuple of strings formatted as
(field or particle type, field name)
Received invalid field key: dust_density, with type <class 'str'>
```
This failure is intentional and aligns with the overarching goal of dropping support for single-string field ids in yt 4.1

I hope reverting the patch doesn't break anything else, but the intended use case has clearly not been tested continuously for a while.
edit: looks like not !

### DIFF
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -29,17 +29,6 @@ from .particle_fields import (
 )
 
 
-def tupleize(inp):
-    if isinstance(inp, tuple):
-        return inp
-    # prepending with a '?' ensures that the sort order is the same in py2 and
-    # py3, since names of field types shouldn't begin with punctuation
-    return (
-        "?",
-        inp,
-    )
-
-
 class FieldInfoContainer(dict):
     """
     This is a generic field container.  It contains a list of potential derived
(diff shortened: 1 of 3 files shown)


neutrinoceros:
While I'm at it let's also simplify all 4 occurrences of `list(sorted())` found in the code base.
`sorted` in fact already returns a list regardless of the input type.
```python
>>> sorted((1, 2, 0))
[0, 1, 2]
```

neutrinoceros:
this is actually low priority, no need to have it in the milestone

neutrinoceros:
closing/reopening t refresh CI and test if #3923 has any unforseen side effect on a new PR.

neutrinoceros:
That's a fair question. Let me give a little more insight on this

Only user provided field keys can be ambiguous. yt used to accept single-str field keys (potentially ambiguous) in a field definition, but yt 4.0 doesn't: user-defined field keys _have_ to be `tuple[str, str]`s.

`FieldInfoContainer.find_dependencies` is called internally at definition/setup time, but only after type validation happens, so it's now safe to assume that it only sees tuple field keys.

Seeing that this patch hasn't caused any failure in tests, I'm rather confident that nothing in yt was actually relying on the patch I'm reverting.

brittonsmith (approved):

neutrinoceros:
@matthewturk can this be merged ?

PR #3930, opened 2022-05-17
ENH: upgrade cached_properties, uniformize Dataset.unique_identifier

neutrinoceros:
## PR Summary
Some systematic (but manual) upgrade of all places where `functools.cached_property` is relevant but currently not used.
`functools.cached_property` is only available since Python 3.8 but we already vendor a backport for Python 3.7 that can be used transparently. The boilerplate code will eventually be auto-cleaned with pre-commit + pyupgrade when we drop support for Python 3.7

Content:
- ENH: use functools' @cached_property decorator where appropriate to simplify code
- ENH: cleanup all implementations of the Dataset.unique_identifier property (guarantee str type, drop unneeded frontend-specific, historic implementations)

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1,5 +1,6 @@
 import abc
 import functools
+import hashlib
 import itertools
 import os
 import pickle
@@ -57,6 +58,11 @@ from yt.utilities.minimal_representation import MinimalDataset
 from yt.utilities.object_registries import data_object_registry, output_type_registry
 from yt.utilities.parallel_tools.parallel_analysis_interface import parallel_root_only
 from yt.utilities.parameter_file_storage import NoParameterShelf, ParameterFileStore
+
+if sys.version_info >= (3, 8):
+    from functools import cached_property
+else:
+    from yt._maintenance.backports import cached_property
 
 if sys.version_info >= (3, 9):
     from collections.abc import MutableMapping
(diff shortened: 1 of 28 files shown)


yt/data_objects/static_output.py
```diff
@@ -306,17 +311,12 @@ def backup_filename(self):
...
-
-    @unique_identifier.setter
-    def unique_identifier(self, value):
-        self._unique_identifier = value
+    @cached_property
+    def unique_identifier(self) -> str:
+        retv = int(os.stat(self.parameter_filename)[ST_CTIME])
+        name_as_bytes = bytearray(map(ord, self.parameter_filename))
+        retv += fnv_hash(name_as_bytes)
+        return str(retv)
```
neutrinoceros, line 319: An important note here is that I'm changing the `unique_identifier` from a mutable property to an immutable one (cached properties can't be mutable). The previous implementation was from https://github.com/yt-project/yt/pull/2559, where my main concern was to reduce duplication, but leaving this property as a mutable was just an implementation detail, not a thoughtful design choice. Now I think it _should_ be immutable.

neutrinoceros:
pre-commit.ci auto-fix

yt/data_objects/static_output.py
matthewturk: I agree with that.

matthewturk (approved):

neutrinoceros:
@matthewturk should this go in ?

PR #3931, opened 2022-05-17
TST: fix a broken pytest fixture

neutrinoceros:
## PR Summary
fix a pytest fixture that was broken in #3773
I don't know how this hasn't been caught up to now but the problem can easily be reproduced by invoking

```shell
pytest yt/visualization/tests/test_save.py
```

### DIFF
--- a/yt/visualization/tests/test_save.py
+++ b/yt/visualization/tests/test_save.py
@@ -4,6 +4,7 @@ import re
 import pytest
 
 import yt
+from yt.testing import fake_amr_ds
 
 # avoid testing every supported format as some backends may be buggy on testing platforms
 FORMATS_TO_TEST = [".eps", ".pdf", ".svg"]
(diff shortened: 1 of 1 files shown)


chrishavlin (approved):
That's fun! 

For what it's worth, I think it passes in the full test suite because the other tests have `from yt.testing import ...` statements, which adds `testing` to the `yt` namespace and `import yt; yt.testing.fake_amr_ds`  subsequently works? Even running just `pytest yt/visualization/` passes!

neutrinoceros:
Oh, I'd never thought of this, good that you found an explanation !

chrishavlin:
Ya! Might have caught it this with something like https://pypi.org/project/pytest-random-order/ -- but there are so many tests that it probably would have taken a while before `test_save` ended up being run first. But I do wonder if we have any other test isolation issues... in any case, I'll go ahead and merge this one.

PR #3935, opened 2022-05-20
RFC: encapsulate try blocks from on demand imports

neutrinoceros:
## PR Summary
My latest attempt at making on_demand_imports.py less repetitive and error prone
I expect I'll need to iterate a couple times on this.

Gains:
- developer experience: far less boilerplate required to add optional dependencies, and an overall lighter module
- user experience: no more necessary startup overhead for some optional deps (like pandas and requests) who were imported even when not used.

TODO:
- [x] cleanup history
- [x] add a couple tests to check that using a missing import raises the expected error (currently I think this is not tested because we skip any test that requires deps that aren't available)

### DIFF
--- a/nose_unit.cfg
+++ b/nose_unit.cfg
@@ -6,5 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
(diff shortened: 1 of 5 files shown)


matthewturk:
Looks good to me

matthewturk (approved):

PR #3936, opened 2022-05-21
RFC: deduplicate __repr__ method for AMRGridPatch subclasses

neutrinoceros:
## PR Summary
title

### DIFF
--- a/yt/data_objects/index_subobjects/grid_patch.py
+++ b/yt/data_objects/index_subobjects/grid_patch.py
@@ -142,7 +142,8 @@ class AMRGridPatch(YTSelectionContainer):
         self.dds.units = self.index.grid_left_edge.units
 
     def __repr__(self):
-        return "AMRGridPatch_%04i" % (self.id)
+        cls_name = self.__class__.__name__
+        return f"{cls_name}_{self.id:04i} ({self.ActiveDimensions})"
 
     def __int__(self):
         return self.id
--- a/yt/frontends/ytdata/data_structures.py
+++ b/yt/frontends/ytdata/data_structures.py
@@ -576,9 +576,6 @@ class YTNonspatialGrid(AMRGridPatch):
         self.LeftEdge = self.index.ds.domain_left_edge
         self.RightEdge = self.index.ds.domain_right_edge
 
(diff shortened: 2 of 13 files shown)


yt/data_objects/index_subobjects/grid_patch.py
```diff
@@ -142,7 +142,8 @@ def _setup_dx(self):
         self.dds.units = self.index.grid_left_edge.units
 
     def __repr__(self):
-        return "AMRGridPatch_%04i" % (self.id)
+        cls_name = self.__class__.__name__
+        return f"{cls_name}_{self.id:04i} ({self.ActiveDimensions})"
```
matthewturk, line 146: I think this is fine, and don't want to change it, but I want to note that `self.ActiveDimensions` is the not set in the base class constructor.

yt/frontends/ytdata/data_structures.py
```diff
@@ -576,9 +576,6 @@ def __init__(self, gid, index, filename=None):
         self.LeftEdge = self.index.ds.domain_left_edge
         self.RightEdge = self.index.ds.domain_right_edge
 
-    def __repr__(self):
```
matthewturk, line 579: I don't think that this has `ActiveDimensions`, right?  So removing this override may introduce an error.

yt/data_objects/index_subobjects/grid_patch.py
neutrinoceros: Ah, my mistake. I can make it conditional then

matthewturk:
Or just set it to None

matthewturk:
Almost certainly, yes.

yt/frontends/ytdata/data_structures.py
neutrinoceros: This line implies (maybe erroneously) that all instances of `AMRGridPatch` have an `ActiveDimensions` attribute https://github.com/yt-project/yt/blob/0fbc997f77bf18c0b45419d4cca06a294fb09516/yt/data_objects/index_subobjects/grid_patch.py#L84

matthewturk (approved):

PR #3937, opened 2022-05-24
TYP: improve type checking around calls to is_sized (alias is_sequence)

neutrinoceros:
## PR Summary
`typing.TypeGuard` (new in Pyhton 3.10, backported to older versions via `typing-extensions`) makes is possible to mark a function such as `yt.funcs.is_sequence` such that type checkers (mypy) can narrow down types where the function is called.
Using this type annotation helped me discover a bug, fixed in the second commit.

Because `collections.abc.Sequence` is a actually more contraining than what we mean in `is_squence`, I'm aliasing it to a `is_sized`, which better fits Python's typing vocabulary.

⚠️ the final scope of this PR is not fully defined yet. Right now it's based on top of #3827 to avoid future conflicts.

- deprecated previous behaviour for `is_sequence` because it doesn't actually check for `collections.abc.Sequence` (the standard lib's definition of a "sequence" in Python)
- previous behaviour is now aliased to a new `is_sized` functions. This one is marked as able to narrow down it's input type with `typing.TypeGuard` (see https://docs.python.org/3/library/typing.html#typing.TypeGuard)
- added a `is_scalar` function, which is a shortcut for `not is_sized`, which I think improves readability
- edited a lot of places where `if is_sized(obj) and len(obj) == n` was used when we have a perfectly good encapsulation for this already in `yt.funcs`: `if obj_length(obj) == n`. Also edited a lot of places where `is_sized` was used but we actually needed `is_sequence`.

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -138,7 +138,7 @@ class PlotCallback(ABC):
             ax = plot.data.axis
             # if this is an on-axis projection or slice, then
             # just grab the appropriate 2 coords for the on-axis view
-            if ax >= 0 and ax <= 2:
+            if ax is not None:
                 (xi, yi) = (
                     plot.data.ds.coordinates.x_axis[ax],
                     plot.data.ds.coordinates.y_axis[ax],
@@ -148,7 +148,7 @@ class PlotCallback(ABC):
             # if this is an off-axis project or slice (ie cutting plane)
             # we have to calculate where the data coords fall in the projected
             # plane
-            elif ax == 4:
+            elif ax is None:
                 # transpose is just to get [[x1,x2,...],[y1,y2,...],[z1,z2,...]]
                 # in the same order as plot.data.center for array arithmetic
                 coord_vectors = coord_copy.transpose() - plot.data.center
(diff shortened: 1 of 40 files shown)


yt/visualization/plot_modifications.py
```diff
@@ -54,11 +55,11 @@ def _check_geometry(self, plot):
 
 def _validate_factor_tuple(factor) -> Tuple[int, int]:
     if (
-        is_sequence(factor)
+        isinstance(factor, Sequence)
         and len(factor) == 2
         and all(isinstance(_, Integral) for _ in factor)
```
neutrinoceros, line 60: for clarity, this is an actual bug fix: previously we were checking that `factor` supported `__len__`, but not `__getitem__`, as a result, `all(isinstance(_, Integral) for _ in factor)` could fail to evaluate, raising an impromptu `TypeError` in the middle of a sanitising function

neutrinoceros:
Switched to draft again because I'm considering adding a `is_scalar` function, essentially encapsulating `not is_sized`, in a more legible fashion (?).
Currently `is_sized` (`is_sequence` on the main branch) is used ~80 times, half of which we actually evaluate `not is_sized`

matthewturk:
@neutrinoceros I'm not sure this is worth the effort; if you can't quite make it work right out of the box, we could wait for more upstream fixes to come around before taking it on.

neutrinoceros:
I'm not sure what kind of upstream change you'd expect here.

neutrinoceros:
I'll leave this as a draft until #3827 is resolved.

neutrinoceros:
I'm going to drop this off for now. In many ways this got out of hands and I'm not even sure I see the value of it any more. In any case I don't fancy spending more time keeping it alive.

PR #3939, opened 2022-05-25
RFC: lift ambiguity on a custom implementation of cached_property

neutrinoceros:
## PR Summary
In response to one of my own, misguided, TODO comments from last year.

### DIFF
--- a/yt/visualization/volume_rendering/old_camera.py
+++ b/yt/visualization/volume_rendering/old_camera.py
@@ -7,7 +7,6 @@ import numpy as np
 from yt.config import ytcfg
 from yt.data_objects.api import ImageArray
 from yt.funcs import ensure_numpy_array, get_num_threads, get_pbar, is_sequence, mylog
-from yt.geometry.geometry_handler import cached_property
 from yt.units.yt_array import YTArray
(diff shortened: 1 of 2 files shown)


matthewturk (approved):
seems fine -- pretty no-op, right?

neutrinoceros:
Hopefully

yt/visualization/volume_rendering/old_camera.py
```diff
@@ -726,7 +725,6 @@ def _render(self, double_check, num_threads, image, sampler):
         image = self.finalize_image(image)
         return image
 
-    @cached_property
```
neutrinoceros, line 729: Note : this should never need to be cached. A module, once imported, is part of the global state and behaves like a singleton.
That said, I have _no_ idea why we'd want to hide pyplot behind a property, I suspect this was done with good intentions but a flawed understanding of how imports work.
I may refactor this in a future PR.

neutrinoceros: update: apparently *I* added this decorator around the same time I added the flawed comment I'm fixing here. I can now confidently assert that the original author didn't know what he was doing :)

PR #3940, opened 2022-05-25
ENH: Adds "Min" Method to ProjectionPlot

rjfarber:
Example/which one???

## PR Summary

This change is not required. But I found that I wanted to take a minimum projection plot and maximum is already implemented, so why not?

## PR Checklist

- [x ] New features are documented, with docstrings and narrative docs

Okay :)

### DIFF
--- a/doc/source/cookbook/simple_projection_methods.py
+++ b/doc/source/cookbook/simple_projection_methods.py
@@ -1,7 +1,7 @@
 import yt
 
 # Load the dataset.
-ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
+ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
 
 # Create projections of temperature (with different methods)
 
--- a/doc/source/cookbook/simple_projection_methods.py
+++ b/doc/source/cookbook/simple_projection_methods.py
@@ -1,0 +1,13 @@
+import yt
+
+# Load the dataset.
+fname = "Shattering_V10_hdf5_plt_cnt_2101"
+ds = yt.load("/Users/ryanjsfx/SCP/ReceivedFiles/Shattering/sR1e5/" + fname)
+
+# Create projections of temperature (with different methods)
+
+
+for method in ["integrate", "min", "max"]:
+    yt.ProjectionPlot(
+        ds, "x", ("gas", "temperature"), method=method
+    ).save(fname + "_Projection_x_temperature_" + method + ".png")
--- a/doc/source/cookbook/simple_projection_methods.py
+++ b/doc/source/cookbook/simple_projection_methods.py
@@ -7,8 +7,8 @@ ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
 
 
 for method in ["integrate", "min", "max"]:
-    yt.ProjectionPlot(
-        ds, "x", ("gas", "temperature"), method=method
-    ).save(method + ".png")
+    yt.ProjectionPlot(ds, "x", ("gas", "temperature"), method=method).save(
+        method + ".png"
+    )
 ## note, the default save name doesn't look at the method
 ## so otherwise it'd just overwrite each loop through
--- a/doc/source/cookbook/simple_projection_methods.py
+++ b/doc/source/cookbook/simple_projection_methods.py
@@ -7,8 +7,7 @@ ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
 
 
 for method in ["integrate", "min", "max"]:
-    yt.ProjectionPlot(ds, "x", ("gas", "temperature"), method=method).save(
-        method + ".png"
-    )
+    proj = yt.ProjectionPlot(ds, "x", ("gas", "temperature"), method=method)
+    proj.save()
 ## note, the default save name doesn't look at the method
 ## so otherwise it'd just overwrite each loop through
--- a/doc/source/cookbook/simple_projection_methods.py
+++ b/doc/source/cookbook/simple_projection_methods.py
@@ -1,8 +1,7 @@
 import yt
 
 # Load the dataset.
-fname = "Shattering_V10_hdf5_plt_cnt_2101"
-ds = yt.load("/Users/ryanjsfx/SCP/ReceivedFiles/Shattering/sR1e5/" + fname)
+ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
 
 # Create projections of temperature (with different methods)
 
@@ -10,4 +9,6 @@ ds = yt.load("/Users/ryanjsfx/SCP/ReceivedFiles/Shattering/sR1e5/" + fname)
 for method in ["integrate", "min", "max"]:
     yt.ProjectionPlot(
         ds, "x", ("gas", "temperature"), method=method
-    ).save(fname + "_Projection_x_temperature_" + method + ".png")
+    ).save(method + ".png")
+## note, the default save name doesn't look at the method
+## so otherwise it'd just overwrite each loop through
--- a/doc/source/cookbook/simple_projection_methods.py
+++ b/doc/source/cookbook/simple_projection_methods.py
@@ -8,6 +8,4 @@ ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
 
 for method in ["integrate", "min", "max"]:
     proj = yt.ProjectionPlot(ds, "x", ("gas", "temperature"), method=method)
-    proj.save()
-## note, the default save name doesn't look at the method
-## so otherwise it'd just overwrite each loop through
+    proj.save(f"projection_method_{method}.png")
--- a/doc/source/cookbook/simple_projection_methods.py
+++ b/doc/source/cookbook/simple_projection_methods.py
@@ -1,7 +1,7 @@
 import yt
 
 # Load the dataset.
-ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
+ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
 
 # Create projections of temperature (with different methods)
 
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -185,13 +185,20 @@ class YTProj(YTSelectionContainer2D):
                 removal="4.2",
             )
             method = style
+            issue_deprecation_warning(
+                "The 'mip' method value is a deprecated alias for 'max'. "
+                "Please use max directly.",
+                since="4.2",
+                removal="5.2",
+            )
+            method = "max"
         if method == "sum":
             self.method = "integrate"
             self._sum_only = True
         else:
             self.method = method
             self._sum_only = False
-        if self.method in ["mip", "max"]:
+        if self.method == "max":
             self.func = np.max
         elif self.method == "min":
             self.func = np.min
@@ -256,7 +263,7 @@ class YTProj(YTSelectionContainer2D):
         projected_units = self.comm.mpi_bcast(self._projected_units)
         self._projected_units = projected_units
         # Note that this will briefly double RAM usage
-        if self.method in ["mip", "max"]:  # TODO: suggest mip as deprecated
+        if self.method == "max":
             merge_style = -1
             op = "max"
         if self.method == "min":
@@ -353,7 +360,7 @@ class YTProj(YTSelectionContainer2D):
                 # for future field accesses.
                 finfo.units = str(chunk[field].units)
             field_unit = Unit(finfo.output_units, registry=self.ds.unit_registry)
-            if self.method in ["mip", "min", "max"] or self._sum_only:
+            if self.method in ["min", "max"] or self._sum_only:
                 path_length_unit = Unit(registry=self.ds.unit_registry)
             else:
                 ax_name = self.ds.coordinates.axis_name[self.axis]
@@ -564,7 +571,7 @@ class YTQuadTreeProj(YTProj):
             chunk.ires.size,
             get_memory_usage() / 1024.0,
         )
-        if self.method in ["mip", "min", "max"] or self._sum_only:
+        if self.method in ["min", "max"] or self._sum_only:
             dl = self.ds.quan(1.0, "")
         else:
             ax_name = self.ds.coordinates.axis_name[self.axis]
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -185,11 +185,11 @@ class YTProj(YTSelectionContainer2D):
                 removal="4.2",
             )
             method = style
+        if method == "mip":         
             issue_deprecation_warning(
                 "The 'mip' method value is a deprecated alias for 'max'. "
                 "Please use max directly.",
-                since="4.2",
-                removal="5.2",
+                since="4.1",
             )
             method = "max"
         if method == "sum":
@@ -198,7 +198,7 @@ class YTProj(YTSelectionContainer2D):
         else:
             self.method = method
             self._sum_only = False
-        if self.method == "max":
+        if self.method in ["max","mip"]:
             self.func = np.max
         elif self.method == "min":
             self.func = np.min
@@ -263,7 +263,7 @@ class YTProj(YTSelectionContainer2D):
         projected_units = self.comm.mpi_bcast(self._projected_units)
         self._projected_units = projected_units
         # Note that this will briefly double RAM usage
-        if self.method == "max":
+        if self.method in ["max","mip"]:
             merge_style = -1
             op = "max"
         if self.method == "min":
@@ -360,7 +360,7 @@ class YTProj(YTSelectionContainer2D):
                 # for future field accesses.
                 finfo.units = str(chunk[field].units)
             field_unit = Unit(finfo.output_units, registry=self.ds.unit_registry)
-            if self.method in ["min", "max"] or self._sum_only:
+            if self.method in ["min", "max","mip"] or self._sum_only:
                 path_length_unit = Unit(registry=self.ds.unit_registry)
             else:
                 ax_name = self.ds.coordinates.axis_name[self.axis]
@@ -571,7 +571,7 @@ class YTQuadTreeProj(YTProj):
             chunk.ires.size,
             get_memory_usage() / 1024.0,
         )
-        if self.method in ["min", "max"] or self._sum_only:
+        if self.method in ["min", "max", "mip"] or self._sum_only:
             dl = self.ds.quan(1.0, "")
         else:
             ax_name = self.ds.coordinates.axis_name[self.axis]
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -191,7 +191,7 @@ class YTProj(YTSelectionContainer2D):
         else:
             self.method = method
             self._sum_only = False
-        if   self.method in ["mip","max"]:
+        if self.method in ["mip", "max"]:
             self.func = np.max
         elif self.method in ["min"]:
             self.func = np.min
@@ -256,7 +256,7 @@ class YTProj(YTSelectionContainer2D):
         projected_units = self.comm.mpi_bcast(self._projected_units)
         self._projected_units = projected_units
         # Note that this will briefly double RAM usage
-        if self.method in ["mip", "max"]: # TODO: suggest mip as deprecated
+        if self.method in ["mip", "max"]:  # TODO: suggest mip as deprecated
             merge_style = -1
             op = "max"
         if self.method == "min":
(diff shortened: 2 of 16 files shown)


matthewturk:
I love it! Thanks, @rjfarber !

neutrinoceros (commented):
Quick comment. Note that you also need to somehow conform to code style standards. Our docs (dev guide, code styling) stipulates how.

doc/source/cookbook/simple_projection_methods.py
```diff
@@ -0,0 +1,13 @@
+import yt
+
+# Load the dataset.
+fname = "Shattering_V10_hdf5_plt_cnt_2101"
+ds = yt.load("/Users/ryanjsfx/SCP/ReceivedFiles/Shattering/sR1e5/" + fname)
```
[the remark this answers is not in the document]
rjfarber: The path to the sample doesn't seem to exist in my repo and I didn't see it in git_ignore...probably I'm missing something simple. I guess I'll look around for it tomorrow.

Regarding conforming to code style standards, sorry probably I'm confused/wrong but I thought pre-commit is supposed to run automatically after I open a pull request (?)

I changed the filename back to what was used in the other cookbook script so hopefully that helps.

matthewturk:
pre-commit.ci autofix

matthewturk:
@rjfarber I think it needs to be turned on first, but I've asked the bot to do it for you.  I think you'll need to pull first before you make any more changes.

yt/utilities/lib/quad_tree.pyx
```diff
@@ -108,8 +121,10 @@ cdef class QuadTree:
                   int nvals, bounds, method = "integrate"):
         if method == "integrate":
             self.combine = QTN_add_value
-        elif method == "mip":
+        elif method in ["mip","max"]:
```
neutrinoceros, line 124: I think this may cause a perf regression as it uses a Python idiom that doesn't translate directly to C. I suggest keeping `==` comparisons here

yt/utilities/lib/quad_tree.pyx
```diff
@@ -202,8 +217,10 @@ cdef class QuadTree:
                          np.ndarray[np.float64_t, ndim=2] values,
                          np.ndarray[np.float64_t, ndim=1] wval,
                          method):
-        if method == "mip" or method == -1:
+        if method in ["mip","max"] or method == -1:
```
neutrinoceros, line 220: Same as above

yt/utilities/lib/quad_tree.pyx
```diff
@@ -424,7 +444,7 @@ cdef class QuadTree:
                 vorig[i] = vtoadd[i]
                 vtoadd[i] += node.val[i]
             wtoadd += node.weight_val
-        elif self.merged == -1:
+        elif self.merged in [-1,-2]:
```
neutrinoceros, line 447: Same as above

yt/utilities/lib/quad_tree.pyx
```diff
@@ -108,8 +121,10 @@ cdef class QuadTree:
                   int nvals, bounds, method = "integrate"):
         if method == "integrate":
             self.combine = QTN_add_value
-        elif method == "mip":
+        elif method in ["mip","max"]:
             self.combine = QTN_max_value
+        elif method in ["min"]:
```
neutrinoceros, line 126: Same as above

doc/source/cookbook/simple_projection_methods.py
```diff
@@ -0,0 +1,14 @@
+import yt
+
+# Load the dataset.
+ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
+
+# Create projections of temperature (with different methods)
+
+
+for method in ["integrate", "min", "max"]:
+    yt.ProjectionPlot(ds, "x", ("gas", "temperature"), method=method).save(
```
neutrinoceros, line 10: I suggest separating save() from instantiation

yt/data_objects/construction_data_containers.py
```diff
@@ -191,8 +191,10 @@ def __init__(
         else:
             self.method = method
             self._sum_only = False
-        if self.method == "mip":
+        if self.method in ["mip", "max"]:
             self.func = np.max
+        elif self.method in ["min"]:
```
neutrinoceros, line 196: Should be a simple `==` here

neutrinoceros (commented):
More remarks. If you wish to add a deprecation please see the dev guide for how to do it.

doc/source/cookbook/simple_projection_methods.py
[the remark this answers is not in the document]
rjfarber: Found the existing dataset here: https://yt-project.org/data/index.html
It's downloading but I'll test the new method with it later :)

yt/utilities/lib/quad_tree.pyx
rjfarber: Sounds good! Should be better now :)

yt/data_objects/construction_data_containers.py
```diff
@@ -185,13 +185,20 @@ def __init__(
                 removal="4.2",
             )
             method = style
+            issue_deprecation_warning(
+                "The 'mip' method value is a deprecated alias for 'max'. "
+                "Please use max directly.",
+                since="4.2",
+                removal="5.2",
```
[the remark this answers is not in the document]
neutrinoceros: another important thing is that you should make sure that `stack_level` is correctly set so that users will see the actual line *they* typed in the warning rather than some random line internal to yt that they know nothing about.

yt/data_objects/construction_data_containers.py
```diff
@@ -256,7 +263,7 @@ def get_data(self, fields=None):
         projected_units = self.comm.mpi_bcast(self._projected_units)
         self._projected_units = projected_units
         # Note that this will briefly double RAM usage
-        if self.method in ["mip", "max"]:  # TODO: suggest mip as deprecated
+        if self.method == "max":
```
neutrinoceros, line 266: it's important to maintain backward compatibility as long as the deprecated feature isn't actually removed, so we have to keep checking for `"mip"` here for now

doc/source/cookbook/simple_projection_methods.py
```diff
@@ -0,0 +1,13 @@
...
+ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
+
+# Create projections of temperature (with different methods)
+
+
+for method in ["integrate", "min", "max"]:
+    proj = yt.ProjectionPlot(ds, "x", ("gas", "temperature"), method=method)
+    proj.save()
+## note, the default save name doesn't look at the method
+## so otherwise it'd just overwrite each loop through
```
neutrinoceros, line 13: I think you forgot to keep your custom file name in refactoring. Here's a suggestion to fix it (I'm also removing the comment because this doesn't need to be justified in the recipe)

```suggestion
    proj.save(f"projection_method_{method}.png")
```

doc/source/cookbook/simple_projection_methods.py
```diff
@@ -0,0 +1,13 @@
+import yt
+
+# Load the dataset.
+ds = yt.load("GalaxyClusterMerger/fiducial_1to3_b0.273d_hdf5_plt_cnt_0175")
```
neutrinoceros, line 4: So the script is currently broken (we can see that it breaks the docs build), so I suggest you iterate a bit more locally and check that it's working as you'd expect before pushing again. Additionally, this specific dataset is huge, running the recipe on my machine takes about a minute (and one of the three methods doesn't even work). I would suggest using a smaller dataset for this demo.
```suggestion
ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
```

neutrinoceros (changes requested):

yt/data_objects/construction_data_containers.py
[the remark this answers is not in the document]
rjfarber: Okay, I updated since to 4.1 and removed removal.

Regarding `stack_level` I tried grep -rin "stack_level" from the top level but it didn't return anything. So then I looked at a few other deprecation warnings and it doesn't look like they include it either. So I guess this is a larger issue (?) For example, the deprecation warning for style instead of method in ProjectionPlot is what I copied from...

yt/data_objects/construction_data_containers.py
rjfarber: Okay yeah I wasn't sure if that'd be confusing so I put "mip" back in everywhere :)

yt/data_objects/construction_data_containers.py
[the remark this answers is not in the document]
neutrinoceros: My mistake, the argument is called `stacklevel`. I just tested it and here I think it should be set to `stacklevel=4` so we see the line where the YTProj object is created.
It would also be helpful to have a separate deprecation warning closer to users, directly in `AxisAlignedProjectionPlot.__init__`, and make sure we don't trigger both warnings at the same time (meaning, `AxisAlignedProjectionPlot` should also be able to replace `"mip"` with `"max"` so the `YTProj` object created internally doesn't need to emit a second warning)

yt/data_objects/construction_data_containers.py
neutrinoceros: maintaining deprecated code is confusing indeed, but it's necessary to avoid breaking everyone all the time :)

yt/data_objects/construction_data_containers.py
```diff
@@ -254,9 +263,12 @@ def get_data(self, fields=None):
         projected_units = self.comm.mpi_bcast(self._projected_units)
         self._projected_units = projected_units
         # Note that this will briefly double RAM usage
-        if self.method == "mip":
+        if self.method in ["max","mip"]:
             merge_style = -1
             op = "max"
+        if self.method == "min":
```
neutrinoceros, line 269: This is the line that's breaking your docs recipe:
```suggestion
        elif self.method == "min":
```

yt/data_objects/construction_data_containers.py
rjfarber: Wellll, note that I set method to "max" if method is "mip" (similar to how the value of style gets set to method) so it shouldn't break removing "mip" elsewhere (theoretically...)

doc/source/cookbook/simple_projection_methods.py
rjfarber: Yup, this is a problem. Working on it now.

yt/data_objects/construction_data_containers.py
neutrinoceros: btw it'd be helpful to add an actual error message in the `NotImplementedError` raised a couple lines below this, like
```python
raise NotImlementedError(f"Unknown projection method {self.method!r}")
```

yt/data_objects/construction_data_containers.py
neutrinoceros: yeah I saw that, thank you ! I was just making a case.

doc/source/cookbook/simple_projection_methods.py
neutrinoceros: got the solution, see my latest comment !

yt/data_objects/construction_data_containers.py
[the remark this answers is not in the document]
rjfarber: It doesn't look like stacklevel is being used anywhere so far:

```
(yt-fork-env) ryanjsfx@MB-145 yt-fork % grep -rin stacklevel .
./yt/_maintenance/deprecation.py:19:    msg: str, *, since: str, removal: Optional[str] = None, stacklevel: int = 3
./yt/_maintenance/deprecation.py:50:    warnings.warn(msg, VisibleDeprecationWarning, stacklevel=stacklevel)
```
but okay I set stacklevel=4 (will commit after I get the cookbook working).

Theoretically, should that also be true of the deprecation warning for the use of 'style' instead of 'method' keyword arguments for ProjectionPlot?

====
Regarding AxisAlignedProjectPlot.__init__ triggering the deprecation warning, should that also be true for the style vs method usage then? If so, maybe beyond the scope of this PR...

doc/source/cookbook/simple_projection_methods.py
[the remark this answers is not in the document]
neutrinoceros: > Regarding conforming to code style standards, sorry probably I'm confused/wrong but I thought pre-commit is supposed to run automatically after I open a pull request (?)

Sorry I missed that. pre-commit can be installed locally (recommended) so you can validate it _as you commit_, or even after that with `pre-commit run --all-files` (this takes a couple minutes to run). This is the easiest way to work collectively with pre-commit. Otherwise, it's run in CI in a check-only way, and we need to invoke the CI bot manually via comments to run it again and fix stuff automatically. This workflow isn't ideal because not everything can be auto-fixed.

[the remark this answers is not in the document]
rjfarber: Yup, got pre-commit installed locally on my laptop and it looks like it's running when do git commits now. It's not working on my work desktop though :( so I'll try to avoid using that machine.

yt/data_objects/construction_data_containers.py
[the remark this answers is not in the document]
neutrinoceros: > It doesn't look like stacklevel is being used anywhere so far:

this is indeed a problem we'll want to fix globally. I wasn't aware of this, thank you for calling it out !

> Theoretically, should that also be true of the deprecation warning for the use of 'style' instead of 'method' keyword arguments for ProjectionPlot?

yes !

> Regarding AxisAlignedProjectPlot.init triggering the deprecation warning, should that also be true for the style vs method usage then? If so, maybe beyond the scope of this PR.

Agreed. You don't need to fix these in your PR, but thanks a lot for raising this concern, I'll dedicate a PR to fixing this globally ! However my point about adding a depr to `AxisAlignedProjectionPlot` still stands :)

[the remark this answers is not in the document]
rjfarber: Okay I moved the deprecation warning to AxisAlignedProjectionPlot (which will propagate here once I get pre-commit happy...)

Is `class YTProj` guaranteed to inherit from the axis-aligned version of ProjectionPlot though? If not, the deprecation won't trigger.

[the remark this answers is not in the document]
neutrinoceros: > Okay I moved the deprecation warning to AxisAlignedProjectionPlot (which will propagate here once I get pre-commit happy...)

There's a slight misunderstanding. I meant that we want two deprecation warnings to exist, not move the one you already had.

> Is class YTProj guaranteed to inherit from the axis-aligned version of ProjectionPlot though? If not, the deprecation won't trigger.

So `YTProj` doesn't actually inherit anything from these classes, but I think I know what you mean, and indeed the deprecation should also be seen for `OffAxisProjectionPlot` so it's best to put it in the base `ProjectionPlot` class.

neutrinoceros (commented):
note that there a many cookbook example using `method="mip"`, you'll want to update them as well so they don't trigger deprecation warnings

yt/data_objects/construction_data_containers.py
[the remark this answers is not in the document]
rjfarber: Sooooo `class ProjectionPlot(NormalPlot)` only has a `__new__` method (no `__init__`) and it's not very clear to me if it takes in `method` (maybe as part of *args (?))

I put in the deprecation warning but doesn't look like it fits aesthetically at least 🤔 

Nerp it yields an UnboundLocalError method referenced before assignment

yt/data_objects/construction_data_containers.py
rjfarber: Sooo I did a `grep -rin "NotImplementedError" .` from the root level of my yt-fork directory and `yt/visualization/volume_rendering/old_camera.py doesn't print anything, but it does log...do you think I should add a log message of the error message here too? Or an error message in the NotImplementedError over there?

Still going through the other instances :)

rjfarber: Ran out of steam but will return to this later.

rjfarber:
I only noticed one 🤔

Anyway I did a `grep -rin "mip" .` from the yt-fork root directory and replaced mips with maxs and think that looks pretty good now. The IsolatedGalaxy example does not look great for the maximum method. It also looks rather artefact-y for the minimum method. So I'll test on the GalaxyClusterMerge dataset later (need to re-download it 😬) that'll also provide a closer comparison to the existing simple projection cookbooks.

yt/data_objects/construction_data_containers.py
neutrinoceros: Ok maybe I shouldn't have made this suggestion, but please try and keep this PR's scope tight. Another PR adding actual error messages to bare `raise NotImplementedError` calls would be welcome but it's hardly related to what you're doing here and we don't want to mix up everything, as the bigger the PR, the harder it is to review

rjfarber: Sure, that makes sense. I'll try to hold you to that then :P

neutrinoceros:
ok we got the docs build back up. I see that some tests are failing already but let's see what else is broken so we can iterate further.

> The IsolatedGalaxy example does not look great for the maximum method. It also looks rather artefact-y for the minimum method.

I can't be the judge of that, but I do want to stress that with IsolatedGalaxy, the recipes takes about 10s, whereas with your preferred dataset it takes about a minute. This is about a 2% overhead on our docs build time, it's far from negligible, which is why I would like if we could find a better compromise.

neutrinoceros:
That's fine, I can get behind that, and we can always optimise this later if we actually need to.

neutrinoceros:
Got a  couple failures on CI. Please run the failing tests locally and fix them. Do not hesitate if you have questions or need further help !

yt/data_objects/tests/test_numpy_ops.py
```diff
@@ -123,11 +123,19 @@ def test_min_max():
         )
 
         p = ad.max(("gas", "density"), axis=1)
-        p1 = ds.proj(("gas", "density"), 1, data_source=ad, method="mip")
+        p1 = ds.proj(("gas", "density"), 1, data_source=ad, method="max")
+        assert_equal(p[("gas", "density")], p1[("gas", "density")])
+
+        p = ad.min(("gas", "density"), axis=1)
+        p1 = ds.proj(("gas", "density"), 1, data_source=ad, method="min")
```
neutrinoceros, line 130: Here you're hiding a `NotImplementedError` from https://github.com/yt-project/yt/blob/0f9d2f4eb9625d6aea54f1d16fde7bb54ad9fb70/yt/data_objects/data_containers.py#L1043
which I'm guessing needs to be replaced with actual code. Taking a look at the similar `max` method it should be pretty trivial.

neutrinoceros (commented):
I took a look at failures, here are some pointers for how to solve them

yt/data_objects/construction_data_containers.py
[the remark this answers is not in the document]
neutrinoceros: Yeah this is delicate. Maybe just don't worry about it, I can do it myself once the rest of the PR is ready :)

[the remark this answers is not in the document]
neutrinoceros: Yeah this is delicate. Maybe just don't worry about it, I can do it myself once the rest of the PR is ready :)

yt/data_objects/data_containers.py
```diff
@@ -1006,7 +1006,7 @@ def max(self, field, axis=None):
                 return rv[0]
             return rv
         elif axis in self.ds.coordinates.axis_name:
-            r = self.ds.proj(field, axis, data_source=self, method="mip")
+            r = self.ds.proj(field, axis, data_source=self, method="max")
             return r
```
neutrinoceros, line 1010: While you're at it, let's simplify this :)
```suggestion
            return self.ds.proj(field, axis, data_source=self, method="max")
```

yt/data_objects/tests/test_projection.py
```diff
@@ -130,8 +130,11 @@ def test_projection(pf):
...
-    proj = ds.proj(("gas", "density"), 2, method="mip", max_level=2)
+    proj = ds.proj(("gas", "density"), 2, method="max", max_level=2)
     assert proj[("index", "grid_level")].max() == 2
 
-    proj = ds.proj(("gas", "density"), 2, method="mip")
+    proj = ds.proj(("gas", "density"), 2, method="max")
     assert proj[("index", "grid_level")].max() == ds.index.max_level
+
+    proj = ds.proj(("gas", "density"), 2, method="min")
+    assert proj[("index", "grid_level")].min() == ds.index.max_level
```
neutrinoceros, line 140:
```suggestion

def test_min_level():
    ds = fake_amr_ds(fields=[("gas", "density")], units=["mp/cm**3"])
    proj = ds.proj(("gas", "density"), 2, method="min")
    assert proj[("index", "grid_level")].min() == ds.index.min_level
```

yt/data_objects/tests/test_numpy_ops.py
rjfarber: Done!

rjfarber:
Okay I looked at the failures and think maybe they'll be good now. It looks like the checks are running above so will see if there's a new error to look at.

neutrinoceros:
Awesome, thank you !
When everything passes I will add a final commit to make sure deprecation warnings are correctly implemented, then I'll merge.

neutrinoceros:
@rjfarber one last failure, and it's coming from a test I "helped" with. I'm sorry if I broke it.

neutrinoceros (approved):

neutrinoceros (approved):

PR #3942, opened 2022-05-26
DEPR: fix a broken deprecation warning (missing stacklevel)

### LINKED ISSUE #3941: DEPR: many deprecation warnings are missing proper stacklevel
### Bug report

**Bug summary**

@rjfarber pointed out that the `stacklevel` argument to `issue_deprecation_warning` was currently unused despite the 27 existing calls. It seems unlikely that the default value is actually correct in every single one of these so it should be inspected more carefully.

neutrinoceros:
## PR Summary
Fix #3941
There are a dozen deprecation warnings in `plot_modifications.py` that are currently impossible to fix because of how plot callbacks' init is delayed until the plot is actually setup.
I'll see if I can refactor this slightly so we at least validate arguments when they are passed.

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -183,6 +183,7 @@ class YTProj(YTSelectionContainer2D):
                 "Please use method directly.",
                 since="3.2",
                 removal="4.2",
+                stacklevel=4,
             )
             method = style
         if method == "sum":


neutrinoceros:
I don't see a simple way to address the problem with plot callbacks, so I suppose it's best to leave it as is: deprecation warnings are only raised when the plot is generated. This should be fine since these particular warnings have been here for a while.

matthewturk (approved):

PR #3943, opened 2022-05-26
ENH: make PlotCallback an abstract class

neutrinoceros:
## PR Summary
Seems to me that the base `PlotCallback` class is abstract in all but name, let's make it explicit.

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -70,7 +70,7 @@ def _validate_factor_tuple(factor) -> Tuple[int, int]:
         )
 
 
-class PlotCallback:
+class PlotCallback(ABC):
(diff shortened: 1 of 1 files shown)


matthewturk (approved):
I'm not sure that making init an abstract method is worth it, since it then means that if we want to add default behavior, we cannot.  I'd ever-so-slightly prefer we add in a proper `super` call to the subclasses and just had `__call__` be abstract.

neutrinoceros:
We can always reintroduce `super().__init__` calls if it becomes useful in the future, but it looks like it never was. My point is that having these calls hanging out in child classes is misleading to someone reading the code, as they accomplish nothing.

neutrinoceros:
@matthewturk feel free to close this if you're unconvinced :)

PR #3946, opened 2022-05-27
DOC: Fix typo in Gadget notebook

matthewturk:
This fixes a typo in the Gadget notebook (and thus in the Gadget cookbook) where we create a new object, `ad2`, and then say we will use it, but we *don't*.  Thanks to @EwanBJones98 for spotting it and letting me know!


neutrinoceros (approved):
Nice catch. You'll need to merge yourself on this one

PR #3947, opened 2022-05-27
DEPR: deprecate unit_base argument in SPHDataset subclasses

### LINKED ISSUE #2979: units_override should replace unit_base for SPH datasets
### Bug report

Most of our datasets have the ability to use the `units_override` functionality to change units. Most of our SPH datasets don't allow this, and instead have something called `unit_base` which does the same thing. This is especially needed when your dataset is not self-descripting. 

It turns out that setting `unit_base` by hand does not even work for Arepo and EAGLE/OWLS datasets--it's simply ignored. 

For yt-4.0 we should just rip off the band-aid and make `units_override` the default user-facing way to do this, while touching as little internal code as possible.

neutrinoceros:
## PR Summary
Fix #2979
I seriously doubt I got this right on first try, so opening as a draft

### DIFF
--- a/yt/frontends/arepo/data_structures.py
+++ b/yt/frontends/arepo/data_structures.py
@@ -1,5 +1,6 @@
 import numpy as np
 
+from yt._maintenance.deprecation import issue_deprecation_warning
 from yt.frontends.gadget.api import GadgetHDF5Dataset
 from yt.funcs import mylog
 from yt.utilities.on_demand_imports import _h5py as h5py
@@ -14,20 +15,27 @@ class ArepoHDF5Dataset(GadgetHDF5Dataset):
         self,
         filename,
         dataset_type="arepo_hdf5",
-        unit_base=None,
+        units_override=None,
         smoothing_factor=2.0,
         index_order=None,
         index_filename=None,
         kernel_name=None,
         bounding_box=None,
-        units_override=None,
         unit_system="cgs",
         default_species_fields=None,
+        *,
+        unit_base=None,
     ):
+        if unit_base is not None:
+            issue_deprecation_warning(
+                "The unit_base argument is a deprecated alias to units_override. "
+                "Please use the units_override argument directly.",
+                since="4.1.0",
+            )
+            units_override = unit_base
         super().__init__(
             filename,
             dataset_type=dataset_type,
-            unit_base=unit_base,
             index_order=index_order,
             index_filename=index_filename,
             kernel_name=kernel_name,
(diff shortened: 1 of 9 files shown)


yt/frontends/arepo/data_structures.py
```diff
@@ -14,20 +15,27 @@ def __init__(
...
+        *,
+        unit_base=None,
     ):
+        if unit_base is not None:
+            issue_deprecation_warning(
+                "The unit_base argument is a deprecated alias to units_override"
+                "To define custom units, please use the units_override argument.",
+                since="4.1.0",
+            )
+            units_override = unit_base
```
neutrinoceros, line 35: Initially I didn't want to change the behaviour: this is essentially a fix for a feature that I'm immediately deprecating.
The reason I changed my mind is that it seemed like the simplest way to avoid repeating when instantiating a class with 3 levels of inheritance, all emitting the very same warning.

jzuhone:
I'm not sure why you're calling it "non-functional", because it definitely does work for Gadget datasets.

As I recognized when I tried to get rid of it the first time, `unit_base` has some subtle differences from `units_override`, which made me reluctant to touch it until I understood it better.

I'll have to look at this PR in more detail before I can comment further.

neutrinoceros:
It's entirely possible that I misunderstood the current state of this parameter. I thought you said it didn't work at all in some frontends. Anyway, if there is more to it than what I already did, I'd happily hand the PR over to you, or we can work out what needs to be done together.

neutrinoceros:
I've identified two problems so far:
- keys need to be translated (`"length"` -> `length_unit`)
- the base `Dataset._sanitize_units_override` methods fails to parse units like `Mpccm/h`. Presumably this unit is registered in the frontends we're touching, but the sanitization happens before units are properly defined (which is the whole point) and so we can't use `self.quan` yet (I think).

So yeah, it's not going to be trivial. Any further insight ?

neutrinoceros:
Okay let's remove this from the milestone. It's not release critical and odds are we won't be able to finish this soon.

jzuhone:
I want to have a look at this, just haven't been able to. Will try this week.

neutrinoceros:
That's okay. It'd still make a lot of sense to ship with 4.1

neutrinoceros:
I think I bit more than I can chew here. I don't want to own this PR and be the leading dev on this specific task where my level of expertise is very poor.
@jzuhone, feel free to reuse my commits at any point if they are of any use.

Issue #3381, opened 2021-06-28
Ambiguous field/particle type: special pixel-fields 


neutrinoceros:
### Bug report

**Bug summary**

Right now when running for instance

```shell
pytest yt/frontends/ytdata/tests/test_unit.py
```
one gets deprecation warnings matching the following regexp
```
r"The requested field name 'pd?[xyz]' is ambiguous and corresponds to any one of the following field types.*"
```
This is because internally, slicing and coordinate handler classes are using some special, pixel-related "fields" `"px"`, `"py"`, `"pz"`, `"pdx"`, `"pdy"` and `"pdz"`, but leave the field (particle) type unspecified.
There doesn't seem to be an obvious path to follow here to resolve the warnings, that would work with every kind of dataset (grid/particles). Ideally we should be able to determine the appropriate type at runtime but this is currently not trivial according to discussions with @cphyc.
This will need to be resolved somehow in order to fully remove compatibility with ambiguous fields, as is targeted for yt 4.1.0

Right now I'm going to ignore this warning in #3380, but I'll reference this issue in the code (specifically in `conftest.py`) to keep track of it.

neutrinoceros:
I think this specific problem is another aspect of https://github.com/yt-project/yt/issues/3839

PR #3948, opened 2022-05-28
ENH: allow name-only field keys when all candidates are aliases

### LINKED ISSUE #3839: Aliased fields shouldn't be considered ambiguous
### Bug report

**Bug summary**

Using ambiguous fields (specified using a field name but leaving the field type as implicit) triggers deprecation warnings even when said ambiguity is only between aliased fields.

**Code for reproduction**

```python
import yt

yt.set_log_level("warning")
ds = fake_amr_ds(fields=["density"], units=["g/cm**3"])
yt.SlicePlot(ds, "z", "density")
```

**Actual outcome**

```
/Users/robcleme/dev/yt-project/yt/yt/data_objects/data_containers.py:1404: VisibleDeprecationWarning: The requested field name 'density' is ambiguous and corresponds to any one of the following field types:
 {'gas', 'stream'}
Please specify the requested field as an explicit tuple (ftype, fname).
Defaulting to '("gas", "density")'.
Deprecated since yt 4.0.0
This feature is planned for removal in yt 4.1.0
  finfo = self.ds._get_field_info(field)
```

**Expected outcome**

In this scenario I would expect no deprecation warning. This needs to be solved for yt 4.1
If a simple solution is found before the last bugfix release in the 4.0.x series, it should be back ported.

neutrinoceros:
## PR Summary
A tentative to close #3839, this depends on #3921

I added a test inspired from https://github.com/yt-project/yt/issues/3839
This cuts deep into the field detection machinery, so I'm counting on existing tests to catch any mistake I made.

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -880,7 +880,7 @@ class YTCoveringGrid(YTSelectionContainer3D):
         alias = {}
         for field in gen:
             finfo = self.ds._get_field_info(*field)
-            if finfo._function.__name__ == "_TranslationFunc":
+            if finfo.is_alias:
                 alias[field] = finfo
                 continue
             try:
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -10,9 +10,10 @@ import weakref
 from collections import defaultdict
 from importlib.util import find_spec
 from stat import ST_CTIME
-from typing import DefaultDict, List, Optional, Tuple, Type, Union
+from typing import DefaultDict, List, Optional, Set, Tuple, Type, Union
 
 import numpy as np
+from more_itertools import unzip
 from unyt.exceptions import UnitConversionError, UnitParseError
 
 from yt._maintenance.deprecation import issue_deprecation_warning
@@ -868,8 +869,38 @@ class Dataset(abc.ABC):
             # https://github.com/yt-project/yt/issues/3381
             return field_info
 
-        fi = self.field_info
-        if not all(fi[c].is_alias_to(fi[candidates[0]]) for c in candidates):
+        def _are_ambiguous(candidates: List[Tuple[str, str]]) -> bool:
+            if len(candidates) < 2:
+                return False
+
+            ftypes, fnames = (list(_) for _ in unzip(candidates))
+            assert all(name == fnames[0] for name in fnames)
+
+            fi = self.field_info
+
+            all_aliases: bool = all(
+                fi[c].is_alias_to(fi[candidates[0]]) for c in candidates
+            )
+
+            all_equivalent_particle_fields: bool
+            if all(ft in self.particle_types for ft in ftypes):
+                ptypes = ftypes
+
+                sub_types_list: List[Set[str]] = []
+                for pt in ptypes:
+                    if pt in self.particle_types_raw:
+                        sub_types_list.append({pt})
+                    elif pt in self.particle_unions:
+                        sub_types_list.append(set(self.particle_unions[pt].sub_types))
+                all_equivalent_particle_fields = all(
+                    st == sub_types_list[0] for st in sub_types_list
+                )
+            else:
+                all_equivalent_particle_fields = False
+
+            return not (all_aliases or all_equivalent_particle_fields)
+
+        if _are_ambiguous(candidates):
             ft, fn = field_info.name
             possible_ftypes = [c[0] for c in candidates]
             msg = (
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -10,7 +10,7 @@ import weakref
 from collections import defaultdict
 from importlib.util import find_spec
 from stat import ST_CTIME
-from typing import DefaultDict, List, Optional, Set, Tuple, Type, Union
+from typing import DefaultDict, Dict, List, Optional, Set, Tuple, Type, Union
 
 import numpy as np
 from more_itertools import unzip
@@ -147,7 +147,7 @@ class Dataset(abc.ABC):
     geometry = "cartesian"
     coordinates = None
     storage_filename = None
-    particle_unions = None
+    particle_unions: Optional[Dict[str, ParticleUnion]] = None
     known_filters = None
     _index_class: Type[Index]
     field_units = None
@@ -883,7 +883,13 @@ class Dataset(abc.ABC):
             )
 
             all_equivalent_particle_fields: bool
-            if all(ft in self.particle_types for ft in ftypes):
+            if (
+                self.particle_types is None
+                or self.particle_unions is None
+                or self.particle_types_raw is None
+            ):
+                all_equivalent_particle_fields = False
+            elif all(ft in self.particle_types for ft in ftypes):
                 ptypes = ftypes
 
                 sub_types_list: List[Set[str]] = []
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -10,7 +10,7 @@ import weakref
 from collections import defaultdict
 from importlib.util import find_spec
 from stat import ST_CTIME
-from typing import DefaultDict, Optional, Tuple, Type, Union
+from typing import DefaultDict, List, Optional, Tuple, Type, Union
 
 import numpy as np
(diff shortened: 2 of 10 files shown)


matthewturk:
Thank you for starting this. I really really like this idea and want to
help.

neutrinoceros:
Thanks ! it'd be helpful to handle https://github.com/yt-project/yt/pull/3921 first

neutrinoceros:
In fact this turned out easier than I expected. I'll mark this PR as ready for review when its parent PR is merged or closed.

neutrinoceros:
added 76751ea24465b598e84ced1e8806ac1c620985c2 as a cheap (tentative) fix for #3381

yt/data_objects/construction_data_containers.py
```diff
@@ -865,7 +865,7 @@ def _split_fields(self, fields_to_get):
         alias = {}
         for field in gen:
             finfo = self.ds._get_field_info(*field)
-            if finfo._function.__name__ == "_TranslationFunc":
+            if finfo.is_alias:
```
matthewturk, line 868: Thanks for adding this attribute.  `is_alias` is long overdue.

matthewturk (commented):
Should this wait for the add fiel stuff to be merged?

neutrinoceros:
Preferably ! Most of the changes you can see here atm actually belong to that other PR

matthewturk:
Just so we see some examples, would you mind editing this with a handful of "this wouldn't work before but it will now" from, say, Enzo or FLASH datasets?

matthewturk:
(I mean, edit the PR)

neutrinoceros:
Seems like a reasonable request, so far my one example and test (taken from #3839) involves a fake dataset.
So I tried to find a real example and came up with this:

```python
import yt

ad = ds.all_data()
p = yt.ParticlePlot(ds, "particle_mass", "dynamical_time")
```
Both fields invoked here are ambiguous (as per yt 4.0), and match 3 and 4 field types respectively). I _expected_ them to be aliases families but the warning still happens on this branch, so either this example is doomed because they're not _actually_ aliases (in which case I may need help to find a better example), or something is wrong with my patch.
@matthewturk do _you_ have examples of alias families in yt, or do you know whether these are ?

matthewturk:
The particle fields are the ones I was thinking of, to be honest!  Stuff
like `particle_mass` and `particle_position_x`.

neutrinoceros:
Surely there's only one particle type in IsolatedGalaxy, right ?
I'm not yet excluding the possibility that some of the particle aliases are not explicitly declared as such. In particular, if they use `ParticleFilter.wrap_func`, I can already conclude that they won't be recognised as aliases.

matthewturk:
In general, Enzo always has only one basic particle type (others have
to be filtered) but I think ti's getting trapped in the particle union
/ particle filter, which sets up the "enzo" aliased into "all" and
maybe "io"?

neutrinoceros:
Here are the full warnings
```
/Users/robcleme/dev/yt-project/yt/yt/data_objects/data_containers.py:1402: VisibleDeprecationWarning: The requested field name 'particle_mass' is ambiguous and corresponds to any one of the following field types:
 {'nbody', 'all'}
Please specify the requested field as an explicit tuple (ftype, fname).
Defaulting to '("all", "particle_mass")'.
Deprecated since yt 4.0.0
This feature is planned for removal in yt 4.1.0
  finfo = self.ds._get_field_info(field)


/Users/robcleme/dev/yt-project/yt/yt/data_objects/data_containers.py:1402: VisibleDeprecationWarning: The requested field name 'dynamical_time' is ambiguous and corresponds to any one of the following field types:
 {'gas', 'nbody', 'all'}
Please specify the requested field as an explicit tuple (ftype, fname).
Defaulting to '("all", "dynamical_time")'.
Deprecated since yt 4.0.0
This feature is planned for removal in yt 4.1.0
  finfo = self.ds._get_field_info(field)
```

TBH I don't know what _any_ of these categories are meant to encompass (except may `"all"`, which means "all particles", right ?), but it may be worth noting that `"enzo"` isn't a candidate type here (isn't it reserved to grid based fields ?)

neutrinoceros:
@matthewturk I found a method to detect (and accept) cases where all fields are equivalent particle fields (such as (`"io", "all", "nbody"` in the case of `IsolatedGalaxy`), which removes the warning for the `"particle_mass"` field in my example above. The `"dynamical_time"` field is _still_ considered ambiguous because `("gas", "dynamical_time")` exists, so I think in this case there's nothing to be done ?

yt/data_objects/static_output.py
```diff
@@ -861,22 +862,60 @@ def _setup_particle_types(self, ptypes=None):
...
+
+        if _are_ambiguous(candidates):
             ft, fn = field_info.name
+            possible_ftypes = [c[0] for c in candidates]
             msg = (
-                f"The requested field name '{fn}' "
+                f"The requested field name {fn!r} "
                 "is ambiguous and corresponds to any one of "
-                f"the following field types:\n {self.field_info._ambiguous_field_names[fn]}\n"
+                f"the following field types:\n {possible_ftypes}\n"
```
cphyc, line 915: With this code, isn't it possible that `possible_ftypes` has multiple time the same value, e.g. if `candidates = [("foo", "bar"), ("foo", "baz")]` ?

yt/fields/field_info_container.py
```diff
@@ -448,16 +449,15 @@ def find_dependencies(self, loaded):
         return loaded, unavailable
 
     def add_output_field(self, name, sampling_type, **kwargs):
+        if name[1] == "density":
+            if name in self:
+                # this should not happen, but it does
```
cphyc, line 454: I am still very much confused after reading this comment. Why is this happening?

yt/fields/tests/test_ambiguous_fields.py
```diff
@@ -11,34 +8,29 @@ def test_ambiguous_fails():
     ds = fake_amr_ds(particles=10)
     msg = "The requested field name '{}' is ambiguous"
 
-    fnames_to_ftype = defaultdict(list)
```
cphyc, line 14: Why are all of these tests disappearing? From reading your code below, it seems you are testing that accessing ` mock_field` raises a warning, but you are not verifying accessing `("io", "mock_field")` and `("gas", "mock_field")` are fine.

yt/fields/field_info_container.py
neutrinoceros: TBH I don't know. I assume some frontends must be incorrectly setting the `("gas", "density")` field more than once

neutrinoceros: Should I add a comment to clarify that I really mean `# TODO: remove this check and fix frontends that try to register a field twice` ?

cphyc: That's fine.

yt/fields/tests/test_ambiguous_fields.py
cphyc: I'd sleep tighter if you also checked that `ds.r["io", "mock_field"]` and `ds.r["gas", "mock_field"]` worked without raising any warning.

cphyc (approved):
LGTM - appart from one or two nits.

yt/fields/tests/test_ambiguous_fields.py
neutrinoceros: Done !

yt/fields/tests/test_ambiguous_fields.py
```diff
@@ -12,13 +12,17 @@ def _mock_field(field, data):
         return data["ones"]
 
     # create a pair of ambiguous field
-    ds.add_field(("io", "mock_field"), _mock_field, sampling_type="cell")
+    ds.add_field(("io", "mock_field"), _mock_field, sampling_type="particle")
```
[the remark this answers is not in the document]
neutrinoceros: actually yt got it for me, I just learned that declaring a field with type `"io"` and `sampling_type="cell"` raises an error, but only when the field is called. That could be improved in the future !

neutrinoceros:
@matthewturk Anything else you need from me here ?

matthewturk:
I'll dig in and report back

PR #3951, opened 2022-05-30
MNT: housekeeping

neutrinoceros:
## PR Summary

- remove unused optional `toml-types` dep (not needed since https://github.com/yt-project/yt/pull/3831)
- remove unused temporary warning filter in `conftest.py` (the simple fact that test pass validate that we don't need it anymore)
- partial migration away from distutils (see #3384): 
  - `logging` is the official recommended replacement for `disutils.log`[source](https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html?highlight=distutils#prefer-setuptools)
  - `distutils.errors` is re-exported within `setuptools.errors` since `setuptools 59.0.0` (see https://github.com/pypa/setuptools/pull/2858)
- remove `setuptools` from runtime dependency (this is a left over from an old misunderstanding: setuptools was in fact never a runtime dependency)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-  "setuptools>=19.6",
+  "setuptools>=59.0.1",
   # see https://github.com/numpy/numpy/pull/18389
   "wheel>=0.36.2",
 
--- a/setupext.py
+++ b/setupext.py
@@ -1,5 +1,6 @@
 import contextlib
 import glob
+import logging
 import os
 import platform
 import shutil
@@ -8,9 +9,7 @@ import sys
 import tempfile
 from textwrap import dedent
 from concurrent.futures import ThreadPoolExecutor
-from distutils import log
 from distutils.ccompiler import CCompiler, new_compiler
-from distutils.errors import CompileError, LinkError
 from distutils.sysconfig import customize_compiler
 from subprocess import PIPE, Popen
 from sys import platform as _platform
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-  "setuptools>=19.6",
+  "setuptools>=59.0.0",
```
matthewturk, line 3: I guess it's been a while since we updated this requirement, eh?

setupext.py
```diff
@@ -8,17 +9,17 @@
...
 from distutils.sysconfig import customize_compiler
 from subprocess import PIPE, Popen
 from sys import platform as _platform
 
 from pkg_resources import resource_filename
 from setuptools.command.build_ext import build_ext as _build_ext
 from setuptools.command.sdist import sdist as _sdist
+from setuptools.errors import CompileError, LinkError
 
+log = logging.getLogger("setupext")
```
matthewturk, line 22: should we explicitly note that this is yt's setupext?

matthewturk (approved):

setupext.py
neutrinoceros: I don't think there's any risk of collisions so I guess not ?

pyproject.toml
neutrinoceros: Yep, 0.19 probably doesn't actually work anymore actually

neutrinoceros: for posterity: setuptools 19.6 only supported Python <= 3.5, so it's clearly not viable anymore
https://pypi.org/project/setuptools/19.6/

matthewturk (approved):

PR #3956, opened 2022-06-03
TST: start testing on Python 3.11

### LINKED ISSUE #3954: ENH: use stdlib tomllib when available (Python 3.11)
### Bug report

**Bug summary**

`tomli` is currently a hard dependency third party, but it's been integrated to the standard library as `tomllib` and will be available in Python 3.11
Because it's the same underlying implementation, this means we should be able to use `tomli` as a "backport" of `tomllib` in Python 3.10 and older as

```python
import sys

if sys.version_info >= (3, 11):
    import tomllib
else:
    import tomli as tomllib
```

However I'd rather not implement this just now, because Python 3.11 isn't testable yet: we depend on numpy for this. Here's the upstream PR to keep track of https://github.com/numpy/numpy/pull/21308

neutrinoceros:
## PR Summary
Fix #3954
This is expected to fail until upstream is sufficiently stable. Currently the limiting factor is kiwisolver, but it looks like it'll be fixed in the coming days.


neutrinoceros:
A new release of kiwi solver is expected to drop next week (see https://github.com/nucleic/kiwi/pull/142)
I'll rebase and fix the conflict then.

neutrinoceros:
Looks like now the bottleneck is Cython. This will probably take much more time to resolve.

matthewturk:
So could we bump to c++11 in our build, or remove the line entirely when on python 3.11?

neutrinoceros:
I *think* the reason why certain extensions require C++03 is that we wanted to support old compilers like GCC 5 (see https://github.com/yt-project/yt/pull/2939)
Ideally we should make this conditional for Python <= 3.10, but I'm don't know how.
Otherwise we may just embrace C++11 as the minimal requirement. It's probably more reasonable now than two years ago anyway, especially since our more rapid release cadence makes it less relevant to install from source.

neutrinoceros:
After a long delay, Python 3.11.0b4 is available, let's try this again

neutrinoceros:
It compiles !
Now we're seeing 4 occurrences of the same problem, for instance:
```
DeprecationWarning: It is deprecated to return a value!=None from a test case (<bound method SimpleVRTest.test_simple_vr of <yt.visualization.volume_rendering.tests.test_simple_vr.SimpleVRTest testMethod=test_simple_vr>>)
```

This seems to be a pytest deprecation. I don't understand why it's only showing up now but a quick inspection seems to indicate that in all 4 cases, the return value of these tests are actually not used, so it shouldn't hurt to just remove them.

neutrinoceros:
This is now ready !

@matthewturk can you take a look ? The sooner this gets in the sooner we'll be able to report any problem with 3.11

yt/visualization/volume_rendering/tests/test_points.py
```diff
@@ -66,4 +66,3 @@ def test_points_vr(self):
         sc.add_source(points_source)
         im = sc.render()
         im.write_png("points.png")
-        return im
```
matthewturk, line 69: Hmm, I thought for a minute that removing the return would make this harder to debug, but now I'm not sure that it matters.  So I'm OK with this.

matthewturk (approved):

PR #3957, opened 2022-06-04
RFC: refactor plot callbacks registration

### LINKED ISSUE #3945: Plot callbacks do not validate input parameters on call
### Bug report

**Bug summary**
Plot classes have `annotate_<x>` methods, which are internally referred to as "plot callbacks", and are implemented as classes, inheriting the bas class `PlotCallback`.
Nominally, these classes have an `__init__` and a `__call__` method. By design, `__call__` isn't actually used before the plot is actually rendered (this is the very essence of a "callback"). However, I noticed that it's also the case for `__init__` methods. This makes is possible to call `annotate_<x>` methods with completely invalid arguments without erroring; any error will only show up at render time, which can be quite confusing

**Code for reproduction**

Here's an example session to demonstrate the problem
```python
>>> import yt
>>> from yt.testing import fake_amr_ds

>>> ds = fake_amr_ds(fields=["density"], units=["g/cm**3"])
yt : [INFO     ] 2022-05-27 11:41:21,888 Parameters: current_time              = 0.0
yt : [INFO     ] 2022-05-27 11:41:21,888 Parameters: domain_dimensions         = [32 32 32]
yt : [INFO     ] 2022-05-27 11:41:21,888 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2022-05-27 11:41:21,889 Parameters: domain_right_edge         = [1. 1. 1.]
yt : [INFO     ] 2022-05-27 11:41:21,889 Parameters: cosmological_simulation   = 0
>>> p = yt.SlicePlot(ds, "z", ("gas", "density"))
yt : [INFO     ] 2022-05-27 11:41:35,183 xlim = 0.000000 1.000000
yt : [INFO     ] 2022-05-27 11:41:35,183 ylim = 0.000000 1.000000
yt : [INFO     ] 2022-05-27 11:41:35,183 xlim = 0.000000 1.000000
yt : [INFO     ] 2022-05-27 11:41:35,183 ylim = 0.000000 1.000000
yt : [INFO     ] 2022-05-27 11:41:35,184 Making a fixed resolution buffer of (('gas', 'density')) 800 by 800
>>> p.annotate_contour(("gas", "density"), invalid_kwarg=True)
<yt.visualization.plot_window.AxisAlignedSlicePlot object at 0x1276e7f70>
>>> p.save("/tmp/")
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/_commons.py", line 146, in newfunc
    self._setup_plots()
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/plot_window.py", line 1374, in _setup_plots
    self.run_callbacks()
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/plot_window.py", line 1498, in run_callbacks
    callback = CallbackMaker(*args[1:], **kwargs)
TypeError: ContourCallback.__init__() got an unexpected keyword argument 'invalid_kwarg'
```

**Expected outcome**

The `TypeError` should show up as soon as we call the `annotate_contour` method.

This came up while I was trying to solve #3941: we have a bunch of deprecation warnings in callback methods, but they only pop at render time, so it's impossible to display the problematic line users actually typed in the warning itself.

I don't think this is a trivial fix because the current implementation for `annotate_<x>` methods is quite complicated and involves a lot of function wrapping. Any proposed solution will need to retain gains from #3310
Unless someone can find an elegant and compact solution soon, this shouldn't be required for yt 4.1

neutrinoceros:
## PR Summary
Fix #3945
Right now this is a proof of concept, there's still work to do, but I think it's promising so I'm opening to see how many tests are currently broken.
It should backwards compatible, and retain compatibility with externally defined plot callbacks (such as `HaloCatalogCallback` from `yt_astro_analysis`).

If I can manage to get it 100% complete while keeping a reasonable diff size, I'll propose this for yt 4.1

Note: I ended up cleaning up a lot of signatures in `PlotCallback` subclasses `__init__` methods (making most arguments keyword only). This change is technically not backward compatible. It is motivated by future compatibility concerns: some arguments are already deprecated, but sometimes occupy early positions, so there is *no* way to actually remove them in the future without breaking backward compatibility. Making them keyword only now is the least painful route I can see.
This wasn't the original goal of this PR, however it turned out that to be a requirement to keep this as tidy as possible, I'm happy to discuss it in more detail upon request.

In short, here's what works on the main branch but not here:

```python
import yt

p.annotate_contour(("gas", "density"), 5, 4, (1e-28, 1e-25))
p.save("/tmp/test_contour.png")
```

on this branch it will raise a TypeError (this is 100% by design)
```python-traceback
  File "/Users/robcleme/dev/yt-project/yt/t_callback_validation.py", line 5, in <module>
    p.annotate_contour(("gas", "density"),5,4, (1e-28, 1e-25))
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/plot_modifications.py", line 103, in closure
    self._callbacks.append(cls(*args, **kwargs))
TypeError: ContourCallback.__init__() takes from 2 to 3 positional arguments but 5 were given
```
The fix is simply to use keyword arguments instead of positional:
```python
p.annotate_contour(("gas", "density"), levels=5, factor=4, clim=(1e-28, 1e-25))
```
(in this particular example, the second argument, `levels`, main still be passed positionally)
Hopefully this is not too concerning of a change, since most arguments were _desinged_ to be used as keyword rather than positional, and that's how they've always been demonstrated in docs. The main reason why they were not already implemented as keyword only is that this wasn't a thing in Python 2, but it's been around since Python 3.0

TODO:

- [x] proof of concept for `PlotCallback` subclasses
- [x] generalize to FRB filters
- [x] validate CI
- [x] move geometry checking responsability from `__call__` to `__init__`
- [x] reimplement mechanism to detect unsupported callbacks according to plot type
- [x] check that `annotate_halos` from yt_astro_analysis still works as expected (checked with https://yt-astro-analysis.readthedocs.io/en/latest/halo_plot_callbacks.html?highlight=annotate_halo#annotate_halos)

## PR Checklist

### DIFF
--- a/yt/visualization/fixed_resolution.py
+++ b/yt/visualization/fixed_resolution.py
@@ -533,7 +533,7 @@ class FixedResolutionBuffer:
                 @wraps(FilterMaker)
                 def method(*args, **kwargs):
                     # We need to also do it here as "invalidate_plot"
-                    # and "apply_callback" require the functions'
+                    # requires the functions'
                     # __name__ in order to work properly
                     @wraps(FilterMaker)
                     def cb(self, *a, **kwa):
--- a/yt/visualization/fixed_resolution.py
+++ b/yt/visualization/fixed_resolution.py
@@ -1,9 +1,9 @@
 import weakref
-from functools import wraps
-from typing import Dict, List, Optional
+from typing import TYPE_CHECKING, Dict, List, Optional
 
 import numpy as np
 
+from yt._maintenance.deprecation import issue_deprecation_warning
 from yt.data_objects.image_array import ImageArray
 from yt.frontends.ytdata.utilities import save_as_dataset
 from yt.funcs import get_output_filename, iter_fields, mylog
@@ -15,12 +15,10 @@ from yt.utilities.lib.api import (  # type: ignore
 from yt.utilities.lib.pixelization_routines import pixelize_cylinder
 from yt.utilities.on_demand_imports import _h5py as h5py
 
-from .fixed_resolution_filters import (
-    FixedResolutionBufferFilter,
-    apply_filter,
-    filter_registry,
-)
 from .volume_rendering.api import off_axis_projection
+
+if TYPE_CHECKING:
+    from yt.visualization.fixed_resolution_filters import FixedResolutionBufferFilter
 
 
 class FixedResolutionBuffer:
@@ -105,7 +103,7 @@ class FixedResolutionBuffer:
         antialias=True,
         periodic=False,
         *,
-        filters: Optional[List[FixedResolutionBufferFilter]] = None,
+        filters: Optional[List["FixedResolutionBufferFilter"]] = None,
     ):
         self.data_source = data_source
         self.ds = data_source.ds
@@ -113,11 +111,18 @@ class FixedResolutionBuffer:
         self.buff_size = (int(buff_size[0]), int(buff_size[1]))
         self.antialias = antialias
         self.data: Dict[str, np.ndarray] = {}
-        self._filters = []
         self.axis = data_source.axis
         self.periodic = periodic
         self._data_valid = False
-        self._filters = filters if filters is not None else []
+
+        # import type here to avoid import cycles
+        from yt.visualization.fixed_resolution_filters import (
+            FixedResolutionBufferFilter,
+        )
+
+        self._filters: List[FixedResolutionBufferFilter] = (
+            filters if filters is not None else []
+        )
 
         ds = getattr(data_source, "ds", None)
         if ds is not None:
@@ -133,8 +138,6 @@ class FixedResolutionBuffer:
             yax = self.ds.coordinates.y_axis[axis]
             self._period = (DD[xax], DD[yax])
             self._edges = ((DLE[xax], DRE[xax]), (DLE[yax], DRE[yax]))
-
-        self.setup_filters()
 
     def keys(self):
         return self.data.keys()
@@ -166,8 +169,7 @@ class FixedResolutionBuffer:
             int(self.antialias),
         )
 
-        for name, (args, kwargs) in self._filters:
-            buff = filter_registry[name](*args, **kwargs).apply(buff)
+        buff = self._apply_filters(buff)
 
         # FIXME FIXME FIXME we shouldn't need to do this for projections
         # but that will require fixing data object access for particle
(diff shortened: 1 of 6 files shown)


yt/visualization/plot_modifications.py
```diff
@@ -33,8 +33,9 @@
 from yt.visualization._commons import _swap_arg_pair_order, _swap_axes_extents
 from yt.visualization.base_plot_types import CallbackWrapper
 from yt.visualization.image_writer import apply_colormap
+from yt.visualization.plot_window import PWViewerMPL
 
-callback_registry = {}
+callback_registry: Dict[str, Type["PlotCallback"]] = {}
```
neutrinoceros, line 38: note that this registry becomes less relevant with this patch, since we don't rely on it to setup annotation methods.
However it's still used in some places (most notably in a docs helper script), so I'm keeping it.

yt/visualization/fixed_resolution.py
```diff
@@ -15,13 +15,11 @@
...
 
-from .fixed_resolution_filters import (
-    FixedResolutionBufferFilter,
-    apply_filter,
-    filter_registry,
-)
 from .volume_rendering.api import off_axis_projection
 
+if TYPE_CHECKING:
+    from yt.visualization.fixed_resolution_filters import FixedResolutionBufferFilter
```
matthewturk, line 21: I am confused about why we need to do this if we're including the class name in quotes in the constructor definition.

matthewturk (approved):
seems OK, but also lots more typing fixes than I'd have expected!

neutrinoceros:
Not really fixes actually, mypy wasn't complaining about their absence. I just figured I might as well encode what I learned while I was at it :)

yt/visualization/fixed_resolution.py
neutrinoceros: Ah, good question. Let me clarify this: putting the name between brackets prevents a NameError to be raised at runtime, but it doesn't tell mypy where to look for the actual definition. Since I can't do a normal import at the top level without creating a cycle, I need to use this type-check-only import. This happens very rarely, and I think this is the first time I need it in yt.

matthewturk: Thanks!

brittonsmith (approved):
I'm happy with this as is. Very nice refactor!

neutrinoceros:
Thanks ! anyone wants to merge ?

PR #3963, opened 2022-06-08
BUG: cleanup swap/flip docs

chrishavlin:
The swap/flip keyword arguments were removed in favor of callbacks-only during development of #3640, looks like the corresponding docs were not updated to reflect that.

### DIFF
--- a/doc/source/visualizing/plots.rst
+++ b/doc/source/visualizing/plots.rst
@@ -798,7 +798,8 @@ will invert the plot's y-axis
    slc.save("NotRightHanded.png")
 
    # flip the vertical axis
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Thank you for the quick fix ! I think the docs build might still hang until #3962 is merged

neutrinoceros:
Also, it may be worth adding a zoom in these demo, this dataset looks much more symmetric at large scales

neutrinoceros:
Update: I isolated the blocking docs fixes in #3965 so it can be merged faster than #3962 (still under discussion for different reasons)

neutrinoceros (approved):
I confirm this patch fixes all such problems

neutrinoceros:
@yt-fido test this please

matthewturk (approved):

PR #3964, opened 2022-06-08
BUG: explicitly install setuptools in check-manifest test env

neutrinoceros:
## PR Summary

In #3951 I cleaned up setup.cfg where setuptools was incorrectly declared as a runtime dependency, but this broke the wheel build CI job which was (also incorrectly) relying on this.
This should fix it

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -81,10 +81,16 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install check-manifest
-    - name: Install yt
+    - name: Install build time dependencies
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Switching to draft since it currently doesn't fix the issue after all

neutrinoceros:
I think I got it now. Sorry for the false start @matthewturk, but please avoid auto-merge here since the crucial part of validation here isn't required by it

neutrinoceros:
here's the job (supposedly) validating what I'm doing here: https://github.com/yt-project/yt/runs/6802547620?check_suite_focus=true

matthewturk (approved):

PR #3971, opened 2022-06-12
DEPR: end deprecation period for deprecated fields

neutrinoceros:
## PR Summary
Concludes the 4.0 -> 4.1 deprecation cycle
Followup to #3059

I propose also removing the field deprecation machinery to (hopefully) reduce the maintenance load.
This needs @jzuhone's bleesing (otherwise I'll revert that part)


neutrinoceros:
This is now ready for review. The remaining failures are all answer tests awaiting a bump because of an updated field name

neutrinoceros:
I see no conflicts with other PRs regarding which tests need to be bumped and it should be uncontroversial, so I'm going to bump them now.

matthewturk (approved):
This is definitely the right thing to do, but I will confess that I am always a little afraid of actually pulling the plug on deprecations.

neutrinoceros:
Me too, especially when they're so short-lived. In the future we might prefer not setting exact expiration dates (or versions) so we can keep backward compatibility for longer when it's not costly

PR #3981, opened 2022-06-21
Dask minimal particle reader

chrishavlin:
This is the first refactor and splitting of #3511 

It includes: 
* a minimal particle reader that is dask-optional that immediately reads into memory (no dask-unyt arrays here)
* a global `dask_enabled` config option (default False, for now?) and a by-frontend `_dask_enabled` attribute in the io handler that turns dask on/off for a frontend.
* some minimal dask helper functions to facilitate the dask-optional behavior
* testing infrastructure updates

Dask here is an optional dependency. 

Here's a minimal example that will utilize dask:

```python
import yt
from yt.config import ytcfg
ytcfg.set("yt", "internals", "dask_enabled", True)
ds = yt.load("MagneticumCluster/snap_132")
ad = ds.all_data()
ad[("gas", "density")]
```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -19,6 +19,7 @@ from yt.utilities.answer_testing.testing_utilities import (
     _streamline_for_io,
     data_dir_load,
 )
+from yt.utilities.on_demand_imports import _dask as dask
 
 MPL_VERSION = Version(matplotlib.__version__)
 NUMPY_VERSION = Version(numpy.__version__)
@@ -64,6 +65,16 @@ def pytest_addoption(parser):
         "test_data_dir",
         default=ytcfg.get("yt", "test_data_dir"),
         help="Directory where data for tests is stored.",
+    )
+    # Dask options. Defaults are chosen for github actions.
+    parser.addoption(
+        "--dask_workers", action="store", default=2, help="number of dask workers"
+    )
+    parser.addoption(
+        "--dask_threads_pw",
+        action="store",
+        default=1,
+        help="number of threads per worker",
     )
 
 
@@ -458,3 +469,24 @@ def Npart(request):
     Needed because indirect=True is used for loading the datasets.
     """
     return request.param
+
+
+@pytest.fixture(scope="session")
+def dask_client_fixture(request):
+    """
+    Fixture for spinning up a dask LocalCluster and Client
+    """
+
+    # there are a number of options, including using some of the internal dask
+    # functions for testing, but this seems the simplest way to ensure only a
+    # single cluster is spun up during testing. Some useful discussion here:
+    # https://github.com/dask/distributed/issues/4479#issuecomment-772119435
+    # also note that if running pytest in parallel, this fixture will need
+    # modification (e.g., pytest-xdist would execute this fixture multiple times)
+
+    n_workers = int(request.config.getoption("--dask_workers"))
+    tpw = int(request.config.getoption("--dask_threads_pw"))
+    lc = dask.distributed.LocalCluster
+    with lc(n_workers=n_workers, threads_per_worker=tpw) as cluster:
+        with dask.distributed.Client(cluster) as client:
+            yield client
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,6 +74,8 @@ doc =
     sphinx-rtd-theme
 full =
     astropy>=4.0.1,<6.0.0
+    dask[array]
+    dask[distributed]
     f90nml>=1.1.2
     fastcache>=1.0.2
     firefly-vis>=2.0.4,<3.0.0
@@ -108,6 +110,8 @@ minimal =
 test =
     codecov~=2.0.15
     coverage~=4.5.1
+    dask[array]
+    dask[distributed]
     nose~=1.3.7
     nose-exclude
     nose-timer~=1.0.0
--- a/yt/utilities/io_handler.py
+++ b/yt/utilities/io_handler.py
@@ -15,7 +15,7 @@ import numpy as np
 from yt._typing import ParticleCoordinateTuple
 from yt.geometry.selection_routines import GridSelector
 from yt.utilities.on_demand_imports import _h5py as h5py
-from yt.utilities.parallel_tools.dask_helper import compute, dask_delay_wrapper
+from yt.utilities.parallel_tools.dask_helper import _get_dask_compute, _get_dask_delayed
 
 io_registry = {}
 
@@ -35,6 +35,7 @@ class BaseIOHandler:
     _cache_on = False
     _misses = 0
     _hits = 0
+    _dask_enabled = False  # global dask_enabled config option may override
 
     def __init_subclass__(cls, *args, **kwargs):
         super().__init_subclass__(*args, **kwargs)
@@ -174,9 +175,13 @@ class BaseIOHandler:
         self, chunks, selector, fields: List[Tuple[str, str]]
     ) -> Dict[Tuple[str, str], np.ndarray]:
 
-        # this particle reader constructs a dask-delayed graph and immediately
-        # computes. If dask is not enabled, all dask operations are passthrough
+        # this particle reader is dask-optional. If dask is enabled, it
+        # constructs a dask-delayed graph and immediately computes.
+        # If dask is not enabled, all dask operations are passthrough
         # functions that have no effect.
+
+        dask_delay = _get_dask_delayed(self._dask_enabled)
+        dask_compute = _get_dask_compute(self._dask_enabled)
 
         # re-organize the fields to read
         ptf, field_maps = self._get_particle_field_containers(fields)
@@ -189,12 +194,12 @@ class BaseIOHandler:
             # note: if dask is not enabled, dask_delay_wrapper is an empty
             # decorator and results are immediately returned
             read_f = self._read_particle_data_file
-            delayed_chunk = dask_delay_wrapper(read_f)(data_file, ptf, selector)
+            delayed_chunk = dask_delay(read_f)(data_file, ptf, selector)
             delayed_chunks.append(delayed_chunk)
 
         # since we are returning in-mem arrays with this function, we immediately compute
         # if dask is not enabled, this will pass through the chunks without modification
-        data_by_chunk = compute(*delayed_chunks)
+        data_by_chunk = dask_compute(*delayed_chunks)
 
         # convert from a list of dicts to dict of lists in order to concatenate
         for chunk in data_by_chunk:
--- a/yt/utilities/io_handler.py
+++ b/yt/utilities/io_handler.py
@@ -13,6 +13,7 @@ else:
 import numpy as np
 
 from yt._typing import ParticleCoordinateTuple
+from yt.config import ytcfg
 from yt.geometry.selection_routines import GridSelector
 from yt.utilities.on_demand_imports import _h5py as h5py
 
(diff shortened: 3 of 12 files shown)


setup.cfg
```diff
@@ -74,6 +74,8 @@ doc =
     sphinx-rtd-theme
 full =
     astropy>=4.0.1,<6.0.0
+    dask[array]
+    dask[distributed]
```
chrishavlin, line 78: note to self: add versions here (and below).

setup.cfg
```diff
@@ -108,6 +110,8 @@ minimal =
 test =
     codecov~=2.0.15
     coverage~=4.5.1
+    dask[array]>=2021.5.0
+    dask[distributed]>=2021.5.0
```
neutrinoceros, line 114: I'm 99% sure this is a mistake:
- the `[test]` target is only meant for testing tools.
- most test jobs already include `[full]` dependencies

so either dask is a mandatory dependency, and it should be included in `install_requires`, or it's optional and it should be only under `[full]`

yt/utilities/parallel_tools/tests/test_dask_helper.py
```diff
@@ -0,0 +1,93 @@
...
+import pytest
+
+from yt.config import ytcfg
+from yt.testing import assert_array_equal, assert_equal
+from yt.utilities.on_demand_imports import _dask as dask
+from yt.utilities.parallel_tools import dask_helper
+
+
+@pytest.fixture(autouse=True)
+def run_before_and_after_tests():
```
neutrinoceros, line 11:
```suggestion
@pytest.fixture(autouse=True, scope="module")
def with_dask_enabled():
```

conftest.py
```diff
@@ -214,6 +214,20 @@ def pytest_configure(config):
             ),
         )
 
+    if find_spec("dask") is not None:
+        # this is fixed upstream, it is only necessary here due to a numpy
+        # version conflict in dask and yt. dask version 2.30.0 is the newest
+        # version that will work with numpy 1.14.2 (the minimum for yt).
+        # once the numpy min can be bumped to >= 1.15.1, the min dask version
+        # can be bumped to 2020.12.0 and this filter can be removed
```
neutrinoceros, line 222: thank you for documenting this extensively. If it make sense, we could also make the filter conditional, and skip it for sufficiently recent versions of numpy (or dask ?)

setup.cfg
```diff
@@ -74,8 +74,8 @@ doc =
     sphinx-rtd-theme
 full =
     astropy>=4.0.1,<6.0.0
-    dask[array]>=2021.5.0
-    dask[distributed]>=2021.5.0
+    dask[array]>=2.30.0
+    dask[distributed]>=2.30.0
```
neutrinoceros, line 78:
```suggestion
    dask[array,distributed]>=2.30.0
```

chrishavlin: I tried that initially and the pre-commit `setup-cfg-fmt` mangled it to: `dask[array,>=2.30.0` so I just put them on separate lines. Would definitely be better on a single line though in terms of version management... any suggestion?

conftest.py
chrishavlin: ya, that's a good idea. I'll try that out.

setup.cfg
neutrinoceros: I wasn't aware of this. Clearly looks like a bug in setup-cfg-fmt so, other than reporting upstream, I don't have a suggestion. Let's keep these lines separated for now since we have to.

conftest.py
```diff
@@ -186,6 +197,29 @@ def pytest_configure(config):
...
+                ),
+            )
+
+        # in python >= 3.10, tornado <=6.1 (dask dependency) throws warnings
+        # Fixed and merged upstream, can remove with tornado 6.2 release:
+        # https://github.com/tornadoweb/tornado/issues/3033
+        config.addinivalue_line(
+            "filterwarnings",
+            "ignore:There is no current event loop:DeprecationWarning",
+        )
```
chrishavlin, line 221: Note to self: add a python version check here.

chrishavlin:
Ya, that'd be a good plan!

And I just looked through everything again -- getting other frontends to work will take probably more work than is feasible for 4.1. The changes in #3752 will actually make it much easier for other frontends to use this framework -- I just rebased and I'm fixing a few things, but it's too big of a PR to review before 4.1 I think. 

So ya, getting this PR merged and noting it as experimental for gadget users would be the way to go for 4.1.

conftest.py
```diff
@@ -3,6 +3,7 @@
 import tempfile
 from importlib.util import find_spec
 from pathlib import Path
+from sys import version_info as py_version_info
```
neutrinoceros, line 6: To respect uniformity across the code base
```suggestion
import sys
```

conftest.py
```diff
@@ -186,6 +198,30 @@ def pytest_configure(config):
...
+            # can be bumped to 2020.12.0 and this filter can be removed
+            config.addinivalue_line(
+                "filterwarnings",
+                (
+                    "ignore:distutils Version classes are deprecated. "
+                    "Use packaging.version instead.:DeprecationWarning"
+                ),
+            )
+
+        if py_version_info >= (3, 10, 0):
```
neutrinoceros, line 216:
```suggestion
        if sys.version_info >= (3, 10):
```

conftest.py
```diff
@@ -441,3 +477,24 @@ def Npart(request):
...
+def dask_client_fixture(request):
+    """
+    Fixture for spinning up a dask LocalCluster and Client
+    """
+
+    # there are a number of options, including using some of the internal dask
+    # functions for testing, but this seems the simplest way to ensure only a
+    # single cluster is spun up during testing. Some useful discussion here:
+    # https://github.com/dask/distributed/issues/4479#issuecomment-772119435
+    # also note that if running pytest in parallel, this fixture will need
```
neutrinoceros, line 492: just curious, *do* we ever run pytest in parallel ?

yt/utilities/parallel_tools/dask_helper.py
```diff
@@ -0,0 +1,32 @@
...
+from yt.utilities.on_demand_imports import _dask as dask
+
+
+def is_dask_array(obj):
+    """checks if obj is a dask array (or a unyt-dask array)"""
+    return isinstance(obj, dask.array.Array)
+
+
+def _empty_decorator(func):
+    def call_func(*args, **kwargs):
```
neutrinoceros, line 11: this should probably still use `functools.wraps` to help debugging/inspection

yt/utilities/parallel_tools/dask_helper.py
```diff
@@ -0,0 +1,32 @@
+from yt.config import ytcfg
+from yt.utilities.on_demand_imports import _dask as dask
+
+
+def is_dask_array(obj):
```
neutrinoceros, line 5: might as well
```suggestion
def is_dask_array(obj) -> bool:
```

yt/utilities/parallel_tools/dask_helper.py
```diff
@@ -0,0 +1,32 @@
...
+
+def _empty_decorator(func):
+    def call_func(*args, **kwargs):
+        return func(*args, **kwargs)
+
+    return call_func
+
+
+def _passthrough_function(*args, **kwargs):
+    return args
```
neutrinoceros, line 18: what happens to `kwargs` here ? Is it intended ?

yt/utilities/parallel_tools/tests/test_dask_helper.py
```diff
@@ -0,0 +1,95 @@
...
+
+    original_config = ytcfg.get("yt", "internals", "dask_enabled")
+    ytcfg.set("yt", "internals", "dask_enabled", False)
+
+    yield  # test will run here
+
+    ytcfg.set("yt", "internals", "dask_enabled", original_config)
+
+
+@pytest.mark.skipif(not dask.__is_available__, reason="requires dask")
```
neutrinoceros, line 25: with `from yt.testing import requires_module_pytest` :)
```suggestion
@requires_module_pytest("dask")
```

yt/utilities/parallel_tools/tests/test_dask_helper.py
```diff
@@ -0,0 +1,95 @@
...
+from yt.utilities.on_demand_imports import _dask as dask
+from yt.utilities.parallel_tools import dask_helper
+
+
+@pytest.fixture(autouse=True, scope="module")
+def with_dask_enabled():
+    # Fixture to set dask_enabled config to standard value then reset.
+    # ensures that these tests will work even if default value for dask_enabled
+    # changes while ensuring the config value gets reset to the default value
+    # after a test runs.
```
neutrinoceros, line 15: The function name seems at odd with the comment (and the implementation), shouldn't it be `with_dask_disabled`/`without_dask_enabled`/`without_dask` ?

yt/utilities/parallel_tools/tests/test_dask_helper.py
```diff
@@ -0,0 +1,95 @@
...
+    ytcfg.set("yt", "internals", "dask_enabled", False)
+
+    yield  # test will run here
+
+    ytcfg.set("yt", "internals", "dask_enabled", original_config)
+
+
+@pytest.mark.skipif(not dask.__is_available__, reason="requires dask")
+def test_compute_helper(dask_client_fixture):
+    ytcfg.set("yt", "internals", "dask_enabled", True)  # make sure its enabled
```
neutrinoceros, line 27: isn't this in direct contradiction with the autoused fixture defined above ?
Maybe we need a pair of fixture (corresponding to True / False) or a single, flexible one, and they shouldn't be `autouse` ?

yt/utilities/io_handler.py
```diff
@@ -172,20 +174,52 @@ def _read_particle_data_file(self, data_file, ptf, selector=None):
     def _read_particle_selection(
         self, chunks, selector, fields: List[Tuple[str, str]]
     ) -> Dict[Tuple[str, str], np.ndarray]:
-        data: Dict[Tuple[str, str], List[np.ndarray]] = {}
 
-        # Initialize containers for tracking particle, field information
+        # re-organize the fields to read
+        ptf, field_maps = self._get_particle_field_containers(fields)
+        data: dict = {field: [] for field in fields}
```
neutrinoceros, line 180: Actually having a type hint here is a very good idea, would you mind fleshing it out a little ?

neutrinoceros (commented):
Here's a batch of superficial comments/questions/suggestions
On the global level:
- how does the global flag `dask_enabled` interact with the `_dask_enabled` class attribute ?
- is the feature actually optional for gadget ? looks like it's on by default, so how can one turn it off from the user side ?

These questions feel like they're probably very well answered by the parts I haven't read yet, but at the same time I think it hints at a lack of user-targeted docs.

Otherwise looks like excellent work thus far, can't wait to do another round !

conftest.py
chrishavlin: not that I know of! but `setup.cfg` does have `pytest-xdist~=2.1.0` in the test requirements. So I suspect someone was experimenting with it at some point...

yt/utilities/parallel_tools/tests/test_dask_helper.py
```diff
@@ -0,0 +1,95 @@
...
+    # ensure that results returned by dask_helper.compute match those
+    # returned by standard dask compute methods.
+    f = dask.array.random.random((1000, 5), chunks=500)
```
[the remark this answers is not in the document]
chrishavlin: shouldn't matter here since it's comparing the result of an operation on the array between a standard method (`dask.compute`) and the wrapped method -- `f` is the same for both methods.

conftest.py
neutrinoceros: oh yeah, good catch. Well I guess we'd better _remove_ this from the requirements then. We have enough on our plate with the pytest migration, making sure it also works in parallel seems to be very low priority. Doesn't have to be in this PR, though I wouldn't be against it

yt/utilities/parallel_tools/tests/test_dask_helper.py
chrishavlin: oh, ya, I'll need to double check... In the tests I'm actually changing the `dask_enabled` value as needed and just using this decorator to make sure the config value is returned to its original state. So let me look again at the tests. but good catch!

chrishavlin:
Thanks! responded to some of the questions, will look through the rest soon. 

To answer the question of the config options vs frontend attribute: the class attribute `_dask_enabled` is True if the frontend works with dask. the global config `dask_enabled` is intended as a user option to use dask where possible. So in the context of this PR, having the default config `dask_enabled=False` with the Gadget `_dask_enabled=True` will mean that even though Gadget **could** use dask, it will not by default.

neutrinoceros:
@chrishavlin it's been a while, do you think this can (should ?) be concluded any time soon ? Otherwise we should probably just remove it from the 4.2 milestone

PR #3983, opened 2022-06-22
improve swap/flip docs

chrishavlin:
Improvement to the swap/flip callback figs in the documentation -- switched up the sliceplot axis and zoomed in. Some of the new figs look like:

StandardAxes

SwappedAxes


matthewturk (approved):
These all look fine to me

PR #3984, opened 2022-06-22
[BUG] volume rendering: sigma_clip normalization 

### LINKED ISSUE #3979: sigma_clip doesn't seem to work with save_annotated()
### Bug report

**Bug summary**

Do a VR and use `.save_annotated()` with a `sigma_clip` and it does not seem to work well compared to the unannotated version.
It is almost like it is adding the transfer function annotation and then evaluating the clipping, which skews the results.

Here's a comparison using the cookbook `vol-annotated.py`:

`.save_annotated()` version:

vol_annotated

`.save()` version:

vol_normal

both of these used the `sigma_clip=6.0` from that example.

**Code for reproduction**

```python
# Paste your code here
#
#
```

**Actual outcome**

```
# If applicable, paste the console output here
#
#
```

**Expected outcome**

**Version Information**

  * Operating System:
  * Python Version:
  * yt version:

chrishavlin:
This closes #3979 and fixes an unreported bug in which not providing a `sigma_clip` value while trying to save a non-png results in a crash. 

There were several spots where `sigma_clip` was used to calculated a clipping value, the version in `save_annotated` incorrectly included the alpha channel in the calculation. I added a `ImageArray._clipping_value` method to cut down on some of the duplicate code.

Here's a script based on the cookbook example mentioned in #3979 

```python
import yt
sc = yt.create_scene(ds, lens_type="perspective")
source = sc[0]
source.set_field(("gas", "density"))
source.set_log(True)
# Set up the camera parameters: focus, width, resolution, and image orientation
sc.camera.focus = ds.domain_center
sc.camera.resolution = 800
sc.camera.north_vector = [0, 0, 1]
sc.camera.position = [1.7, 1.7, 1.7]
sc.annotate_axes(alpha=0.02)
sc.annotate_domain(ds, color=[1, 1, 1, 0.01])
text_string = f"T = {float(ds.current_time.to('Gyr'))} Gyr"

sc.save_annotated("vol_annotated_6.png", sigma_clip=6, text_annotate=[[(0.1, 0.95), text_string]])
sc.save("vol_save_6.png", sigma_clip=6, render=False)
sc.save("vol_save_6.jpg", sigma_clip=6, render=False)
sc.save("vol_save.jpg", render=False)  # this one errors on main!
```

The `save_annotated` and `save` versions now match:

vol_annotated_6
vol_save_6

### DIFF
--- a/yt/visualization/volume_rendering/scene.py
+++ b/yt/visualization/volume_rendering/scene.py
@@ -349,10 +349,10 @@ class Scene:
             ax = fig.add_axes([0, 0, 1, 1])
             ax.set_axis_off()
             out = self._last_render
-            nz = out[:, :, :3][out[:, :, :3].nonzero()]
-            max_val = nz.mean() + sigma_clip * nz.std()
+            if sigma_clip:
+                max_val = out._clipping_value(sigma_clip)
+                out = np.clip(out[:, :, :3] / max_val, 0.0, 1.0) * 255
             alpha = 255 * out[:, :, 3].astype("uint8")
-            out = np.clip(out[:, :, :3] / max_val, 0.0, 1.0) * 255
             out = np.concatenate([out.astype("uint8"), alpha[..., None]], axis=-1)
             # not sure why we need rot90, but this makes the orientation
             # match the png writer
@@ -511,11 +511,9 @@ class Scene:
         ax.set_position([0, 0, 1, 1])
 
         if sigma_clip is not None:
-            nz = im[im > 0.0]
-            nim = im / (nz.mean() + sigma_clip * np.std(nz))
+            nim = im / im._clipping_value(sigma_clip)
             nim[nim > 1.0] = 1.0
             nim[nim < 0.0] = 0.0
-            del nz
         else:
             nim = im
         axim = ax.imshow(nim[:, :, :3] / nim[:, :, :3].max(), interpolation="bilinear")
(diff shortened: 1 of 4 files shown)


yt/visualization/volume_rendering/scene.py
```diff
@@ -511,11 +513,9 @@ def _show_mpl(self, im, sigma_clip=None, dpi=100):
         ax.set_position([0, 0, 1, 1])
 
         if sigma_clip is not None:
-            nz = im[im > 0.0]
-            nim = im / (nz.mean() + sigma_clip * np.std(nz))
```
chrishavlin, line 515: just a note to reviewers: this was the source of the bug. It should only normalize based on the RGB channels (i.e., `im[:,:,:3]`)

zingale: awesome.  Nice catch.  I never got around to digging deeper, so I'll happily review your PR ;)

zingale (dismissed):

zingale:
I confirmed the output that this works as expected.  Nice work @chrishavlin !!

yt/visualization/volume_rendering/scene.py
```diff
@@ -349,8 +349,10 @@ def save(
             ax = fig.add_axes([0, 0, 1, 1])
             ax.set_axis_off()
             out = self._last_render
-            nz = out[:, :, :3][out[:, :, :3].nonzero()]
-            max_val = nz.mean() + sigma_clip * nz.std()
+            if sigma_clip:
```
neutrinoceros, line 352: better avoid treating 0 as None (sigma_clip=0 means « clip to the mean value », right ?)
```suggestion
            if sigma_clip is not None:
```

neutrinoceros (changes requested):
Nice job ! Found one caveat hat can be fixed easily. I’ll merge when it’s dealt with

Issue #3944, opened 2022-05-26
symlog and log default behaviors


chrishavlin:
This issue is a dedicated discussion to resolving some open questions on handling default behavior for normalization in the symlog but also IMO log scaling behavior. On the face, the question is just: when using symlog scaling, how should yt select a default value for the linthresh parameter? But answering that is actually tied pretty closely to an upstream matplotlib bug and our attempts at handling it... 

### background

With the new plot norm and colorbar handling in #3849 , we have returned to the discussion of a choosing good linthresh value. Since 3849 did not originate those changes, it's worth taking a broad look at related development threads to think through how to handle these defaults. So first, here's a quick summary of related development history:

- #2890 : a projection plot containing large data range and 0-values errors in yt's call to matplotlib's `imshow` when using the default log scale. This was addressed by switching to symlog when certain conditions are met (0-values in the data, data range above a cutoff value, 3.3 >= matplotlib version < 3.5) ( fixes in #3161 , #3754 ). 
- PR #3295 then adjusted the symlog parameters so that the default lintresh would be the minimum, nonzero absolute value of the data, which results in a nicer image for FIRE datasets. 
- #3858 found that in some situations the new linthresh default could result in matplotlib crashes, this was fixed by #3859 by adding another default linthresh, set to `max(abs(data)) / 1000)` (though it is in fact, only partly fixed, see later on...)
- The latest, #3849, includes only the `max(abs(data)) / 1000)` default.

The main conceptual issue with the current linthresh default is that for most cosmological simulations will have a **much** larger range than 3 orders of magnitude -- below I'll give some ideas for improving that. But underlying all of this, and the reason why I pulled out the development history above, is the same matplotlib bug that gave rise to the need for #3161, #3754 **and** #3858 in the first place, I'll highlight this in below as well. 

### improving the linthresh default

So the current way of calculating the linthresh default, `max(abs(data)) / 1000`, is IMO a nice approach because it will definitely not error for any range of `data`. But it's not great because many (most?) of the datasets that yt handles have much more than 3 orders of magnitude in range. So if we want to keep this as absolutely as simple as possible, I think we can just increase that scale to something like  `max(abs(data)) / 1e10` and end up with a better image for a wider range of datasets. If we're going to have a fixed default setting, I think we should default to one that is likely to work well for the astro datasets that yt mostly handles. 

But we could generalize this a little better by actually checking what the dynamic range of our image array is and scaling by a value appropriate to the data range with some cutoff:

```
[...]

Issue #3858, opened 2022-03-24
BUG: invalid auto symlog parameters


neutrinoceros:
### Bug report

**Bug summary**

In some instances, trying to plot an image containing both negative values and extremely small values results in a crash in matplotlib.

This bug was first discovered and reported on slack by @BenWibking.
He provided a reproducer script and an example dataset where this bug happens in production
https://cloudstor.aarnet.edu.au/plus/s/TBFG6m48BXVKdkE

**Code for reproduction**
Here's a pure yt reproducer
```python
import numpy as np
import yt

shape = (64,64,1)
arr = np.full(shape, 5.e-324)
arr[0, 0] = -1e12
arr[1, 1] = 200
d = {'scalar': arr}

ds = yt.load_uniform_grid(d, shape)
p = yt.SlicePlot(ds, "z", ('stream', 'scalar'))
p.save("/tmp/output.png")
```

**Actual outcome**

```
yt : [INFO     ] 2022-03-24 10:58:41,550 Parameters: current_time              = 0.0
yt : [INFO     ] 2022-03-24 10:58:41,550 Parameters: domain_dimensions         = [64 64  1]
yt : [INFO     ] 2022-03-24 10:58:41,550 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2022-03-24 10:58:41,551 Parameters: domain_right_edge         = [1. 1. 1.]
yt : [INFO     ] 2022-03-24 10:58:41,551 Parameters: cosmological_simulation   = 0
yt : [INFO     ] 2022-03-24 10:58:41,795 xlim = 0.000000 1.000000
yt : [INFO     ] 2022-03-24 10:58:41,795 ylim = 0.000000 1.000000
yt : [INFO     ] 2022-03-24 10:58:41,795 xlim = 0.000000 1.000000
yt : [INFO     ] 2022-03-24 10:58:41,795 ylim = 0.000000 1.000000
yt : [INFO     ] 2022-03-24 10:58:41,796 Making a fixed resolution buffer of (('stream', 'scalar')) 800 by 800
yt : [WARNING  ] 2022-03-24 10:58:41,885 Plotting ('stream', 'scalar'): Both positive and negative values. Min = -1000000000000.0017 dimensionless, Max = 200.00000000000034 dimensionless.
yt : [WARNING  ] 2022-03-24 10:58:41,885 Switching to symlog colorbar scaling.
[...]

PR #3986, opened 2022-06-24
Updating the linthresh guessing

chrishavlin:
This is an attempt at updating how the linthresh value is calculated for SymLogNorms when not provided. Note that I based it off of #3849 , so I'm submitting in draft mode until that one is cleared. The changes here are simple enough that it should be easy to update this PR after #3849 is fully reviewed. 

IMO the linthresh calculation here nicely handles all of the concerns of #3944 without too much complexity. I need to update/add some tests and the docs would need a little updating, but it'd be great to get some feedback on the approach. 

Conceptually, the calculation here uses the minimum absolute nonzero value as a starting guess at linthresh (as in  #3295) but then limits the dynamic range of the symlog scale when necessary to avoid the floating point precision issues that caused the errors in #3858. The limiting is sensitive to the data so it does a decent job for a range of data distributions (and should not error for any).  

Some images:

```python
fn = 'FIRE_M12i_ref11/snapshot_600.hdf5'
ds = yt.load(fn)
p = yt.ProjectionPlot(ds, 'x', ('gas', 'density'), width=(30, 'Mpc'))
p.show()
```
image

And using the sample data from #3858 
```python
import yt

ds = yt.load('symlogsets/yt_bug_reproducer/plt04000')
slc = yt.SlicePlot(ds, "z", ("boxlib", "scalar"))
slc.show()
```

image

### DIFF
--- a/yt/visualization/_handlers.py
+++ b/yt/visualization/_handlers.py
@@ -329,6 +329,11 @@ class NormHandler:
             if kw["vmin"] == kw["vmax"] or not np.any(finite_values_mask):
                 norm_type = Normalize
             elif kw["vmin"] <= 0:
+                # note: matplotlib >= 3.5.2 could use LogNorm here if vmin==0
+                # if we do the following: set vmin to the min nonzero value and
+                # supply the interpolation_stage='rgba' keyword arg to imshow.
+                # this would, however, change default behavior and break answer
+                # tests, so needs a wider discussion.
                 norm_type = SymLogNorm
             else:
                 norm_type = LogNorm
@@ -337,7 +342,7 @@ class NormHandler:
             if self.linthresh is not None:
                 linthresh = self.to_float(self.linthresh)
             else:
-                linthresh = self.to_float(self._guess_linthresh(data))
+                linthresh = self._guess_linthresh(data)
 
             kw.setdefault("linthresh", linthresh)
             if MPL_VERSION >= Version("3.2"):
@@ -348,32 +353,49 @@ class NormHandler:
 
         return norm_type(*args, **kw)
 
-    def _guess_linthresh(self, data):
-        # data is the ImageArray to plot
+    def _guess_linthresh(self, plot_data):
+        # data is an ImageArray or ColorbarHandler data
 
-        # get the data range as unyt quantities
-        nz_data = data[data != 0]
-        min_abs_val, max_abs_val = np.sort(
-            (np.abs(np.nanmin(nz_data)), np.abs(np.nanmax(nz_data)))
-        )
-        min_abs_val = self.ds.quan(min_abs_val, data.units)
-        max_abs_val = self.ds.quan(max_abs_val, data.units)
-        min_abs_nonzero = self.ds.quan(np.nanmin(np.abs(nz_data)), data.units)
+        # get the extrema for the negative and positive values separately
+        # neg_min -> neg_max -> 0 -> pos_min -> pos_max
+        def get_minmax(data):
+            if len(data) > 0:
+                return np.nanmin(data), np.nanmax(data)
+            return None, None
 
-        linthresh = min_abs_nonzero
-        # the above works well for most data. But need to control for a few
-        # edge cases related to floating point precision errors.
+        pos_min, pos_max = get_minmax(plot_data[plot_data > 0])
+        neg_min, neg_max = get_minmax(plot_data[plot_data < 0])
 
-        # first, set the scaling factor
-        cutoff_sigdigs = 30.0  # max allowable range in significant digits
-        actual_log_range = np.log10(max_abs_val) - np.log10(min_abs_val)
-        if actual_log_range < cutoff_sigdigs:
-            cutoff_sigdigs = np.ceil(actual_log_range)
+        has_pos = pos_min is not None
+        has_neg = neg_min is not None
 
-        if np.log10(max_abs_val) - np.log10(linthresh) > cutoff_sigdigs:
-            # this is required to avoid rounding error in the matplotlib normalization
-            # when the range is too large.
-            linthresh = max_abs_val / (10**cutoff_sigdigs)
+        # the starting guess is the absolute value of the point closest to 0:
+        if has_pos and has_neg:
+            linthresh = np.min((np.abs(neg_max), pos_min))
+        elif has_pos:
+            linthresh = pos_min
+        elif has_neg:
+            linthresh = np.abs(neg_max)
+        else:
+            # this condition should handled before here in get_norm
+            raise RuntimeError("No negative or positive values")
+
+        log10_linthresh = np.log10(linthresh)
+
+        # if either the pos or neg ranges exceed cutoff_sigdigs, then
+        # linthresh is shifted to decrease the range to avoid floating point
+        # precision errors in the normalization.
+        cutoff_sigdigs = 15.0  # max allowable range in significant digits
+        if has_pos and np.log10(pos_max) - log10_linthresh > cutoff_sigdigs:
+            linthresh = pos_max / (10.0**cutoff_sigdigs)
+            log10_linthresh = np.log10(linthresh)
+
+        if has_neg and np.log10(np.abs(neg_min)) - log10_linthresh > cutoff_sigdigs:
+            linthresh = np.abs(neg_min) / (10.0**cutoff_sigdigs)
+
+        if hasattr(linthresh, "units"):
+            # if the original plot_data has units, linthresh will have units here
+            return self.to_float(linthresh)
 
         return linthresh
 
--- a/yt/visualization/_handlers.py
+++ b/yt/visualization/_handlers.py
@@ -373,13 +373,13 @@ class NormHandler:
         # the starting guess is the absolute value of the point closest to 0
         # (remember: neg_max is closer to 0 than neg_min)
         if has_pos and has_neg:
-            linthresh = np.min((np.abs(neg_max), pos_min))
+            linthresh = np.min((-neg_max, pos_min))
         elif has_pos:
             linthresh = pos_min
         elif has_neg:
-            linthresh = np.abs(neg_max)
+            linthresh = -neg_max
         else:
-            # this condition should handled before here in get_norm
+            # this condition should be handled before here in get_norm
             raise RuntimeError("No negative or positive values")
 
         log10_linthresh = np.log10(linthresh)
@@ -387,15 +387,15 @@ class NormHandler:
         # if either the pos or neg ranges exceed cutoff_sigdigs, then
         # linthresh is shifted to decrease the range to avoid floating point
         # precision errors in the normalization.
-        cutoff_sigdigs = 15.0  # max allowable range in significant digits
+        cutoff_sigdigs = 15  # max allowable range in significant digits
         if has_pos and np.log10(pos_max) - log10_linthresh > cutoff_sigdigs:
             linthresh = pos_max / (10.0**cutoff_sigdigs)
             log10_linthresh = np.log10(linthresh)
 
-        if has_neg and np.log10(np.abs(neg_min)) - log10_linthresh > cutoff_sigdigs:
+        if has_neg and np.log10(-neg_min) - log10_linthresh > cutoff_sigdigs:
             linthresh = np.abs(neg_min) / (10.0**cutoff_sigdigs)
 
-        if hasattr(linthresh, "units"):
+        if isinstance(linthresh, unyt_quantity):
             # if the original plot_data has units, linthresh will have units here
             return self.to_float(linthresh)
 
--- a/yt/visualization/_handlers.py
+++ b/yt/visualization/_handlers.py
@@ -326,45 +326,19 @@ class NormHandler:
             # allowing to toggle between lin and log scaling without detailed user input
             norm_type = self.norm_type
         else:
-            if kw["vmin"] == kw["vmax"] or not np.any(np.isfinite(data)):
+            if kw["vmin"] == kw["vmax"] or not np.any(finite_values_mask):
                 norm_type = Normalize
             elif kw["vmin"] <= 0:
                 norm_type = SymLogNorm
-            elif (
-                Version("3.3") <= MPL_VERSION < Version("3.5")
-                and kw["vmin"] == 0
-                and kw["vmax"] > 0
-            ):
-                # normally, a LogNorm scaling would still be OK here because
-                # LogNorm will mask 0 values when calculating vmin. But
-                # due to a bug in matplotlib's imshow, if the data range
-                # spans many orders of magnitude while containing zero points
-                # vmin can get rescaled to 0, resulting in an error when the image
-                # gets drawn. So here we switch to symlog to avoid that until
-                # a fix is in -- see PR #3161 and linked issue.
-                cutoff_sigdigs = 15
-                if (
-                    np.log10(np.nanmax(data[np.isfinite(data)]))
-                    - np.log10(np.nanmin(data[data > 0]))
-                    > cutoff_sigdigs
-                ):
-                    norm_type = SymLogNorm
-                else:
-                    norm_type = LogNorm
             else:
                 norm_type = LogNorm
 
         if norm_type is SymLogNorm:
-            # if cblinthresh is not specified, try to come up with a reasonable default
-            min_abs_val, max_abs_val = np.sort(
-                np.abs((self.to_float(np.nanmin(data)), self.to_float(np.nanmax(data))))
-            )
             if self.linthresh is not None:
                 linthresh = self.to_float(self.linthresh)
-            elif min_abs_val > 0:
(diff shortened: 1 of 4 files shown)


chrishavlin:
Note that I experimented with the new imshow kwarg (`interpolation_stage="rgba"`) but still got some rounding error crashes related to the normalization objects (`SymLogNorm`, `LogNorm`) and so gave up on that route...

yt/visualization/_handlers.py
```diff
@@ -0,0 +1,458 @@
...
+        actual_log_range = np.log10(max_abs_val) - np.log10(min_abs_val)
+        if actual_log_range < cutoff_sigdigs:
+            cutoff_sigdigs = np.ceil(actual_log_range)
+
+        if np.log10(max_abs_val) - np.log10(linthresh) > cutoff_sigdigs:
+            # this is required to avoid rounding error in the matplotlib normalization
+            # when the range is too large.
+            linthresh = max_abs_val / (10**cutoff_sigdigs)
+
+        return linthresh
```
chrishavlin, line 379: So this section is the relevant change...

yt/visualization/_handlers.py
```diff
@@ -0,0 +1,458 @@
...
+    def _guess_linthresh(self, data):
+        # data is the ImageArray to plot
+
+        # get the data range as unyt quantities
+        nz_data = data[data != 0]
+        min_abs_val, max_abs_val = np.sort(
+            (np.abs(np.nanmin(nz_data)), np.abs(np.nanmax(nz_data)))
+        )
+        min_abs_val = self.ds.quan(min_abs_val, data.units)
+        max_abs_val = self.ds.quan(max_abs_val, data.units)
```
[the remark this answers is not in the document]
neutrinoceros: ⚠️ Actually `data` cannot be assumed to even *have* units... (see line 317 for instance)

neutrinoceros (commented):
Thank you so much for continuing this gianormous task !
I think I spotted a couple issues with the current approach, but nothing that we can't address.
A related question: did you decide that using `interpolation_stage` (as discussed in #3944) wasn't a reliable solution in the end, and why ?

yt/visualization/_handlers.py
```diff
@@ -327,45 +327,19 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
             # allowing to toggle between lin and log scaling without detailed user input
             norm_type = self.norm_type
         else:
-            if kw["vmin"] == kw["vmax"] or not np.any(np.isfinite(data)):
+            if kw["vmin"] == kw["vmax"] or not np.any(finite_values_mask):
                 norm_type = Normalize
             elif kw["vmin"] <= 0:
                 norm_type = SymLogNorm
-            elif (
-                Version("3.3") <= MPL_VERSION < Version("3.5")
```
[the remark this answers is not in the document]
chrishavlin: Two reasons: first is that the preceeding `kw["vmin"] <= 0"` condition would always catch before this one (if `kw['vmin']==0` then the preceeding `kw['vmin'] <= 0` will always be True) . I considered flagging it in your PR but since it's closely related to symlog/linthresh settings it seemed better to address it here. 

The second reason I went for removal instead of changing the condition is that controlling for the issue when the range is too large happens down in the linthresh calculation so this seemed redundant. Thinking about it now though, I do think it would actually make sense to allow `LogScale`  when possible... but I think this condition probably needs some adjusting rather than straight removal.

I also mean to test this more thoroughly will different matplotlib versions before switching from draft -- so far just been using the latest release (3.5.2).

yt/visualization/_handlers.py
```diff
@@ -0,0 +1,458 @@
...
+        max_abs_val = self.ds.quan(max_abs_val, data.units)
+        min_abs_nonzero = self.ds.quan(np.nanmin(np.abs(nz_data)), data.units)
+
+        linthresh = min_abs_nonzero
+        # the above works well for most data. But need to control for a few
+        # edge cases related to floating point precision errors.
+
+        # first, set the scaling factor
+        cutoff_sigdigs = 30.0  # max allowable range in significant digits
+        actual_log_range = np.log10(max_abs_val) - np.log10(min_abs_val)
```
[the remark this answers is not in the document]
chrishavlin: oh, good call. didn't think about mixed signs possibly having the same abs value. If there's not one yet, I'll add a test for that special case too. 

Ya, renaming variables for clarity throughout is a good call.

yt/visualization/_handlers.py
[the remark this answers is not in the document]
chrishavlin: yikes, good catch. will add a similar attribute check for units.

yt/visualization/_handlers.py
```diff
@@ -327,45 +327,19 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
-            min_abs_val, max_abs_val = np.sort(
-                np.abs((self.to_float(np.nanmin(data)), self.to_float(np.nanmax(data))))
-            )
             if self.linthresh is not None:
                 linthresh = self.to_float(self.linthresh)
-            elif min_abs_val > 0:
-                linthresh = min_abs_val
             else:
-                linthresh = max_abs_val / 1000
+                linthresh = self.to_float(self._guess_linthresh(data))
```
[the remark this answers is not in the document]
chrishavlin: ya, could definitely return the float value from `_guess_linthresh`. I initially had `_guess_linthresh` as a standalone function so put the to_float here and forgot to move it to `_guess_linthresh` after I made it a class method :) will do that!

chrishavlin:
So the issue with using `interpolation_stage` is that it is only an `imshow` kwarg. When using the `Norm` objects directly, it's pretty easy to get it to error with the edge cases we're trying to handle here. e.g., using a linthresh close to the underflow limit will error:

```
from matplotlib.colors import SymLogNorm
v = SymLogNorm(1e-323, vmin=-1e-1, vmax=1e-1)
v(1e-2)  # errors with the familiar `ValueError: Invalid vmin or vmax`. 
```
There's some discussion of adding an interpolation stage argument to the `Norm` objects (e.g, the end of this [issue](https://github.com/matplotlib/matplotlib/issues/21167)), but even if that happens soon we'd then be handling multiple dependency conditions for the same underlying problem. And so I feel like it's a lot simpler to just try to avoid the conditions that give rise to the precision issues.

I mentioned this in one of the replies, in addition to adding to the existing yt tests, I'm planning to do a more thorough matplotlib-dependent suite of tests to check all these edge cases before switching from draft.

neutrinoceros:
Seems reasonable, thank you for your research on this !

yt/visualization/_handlers.py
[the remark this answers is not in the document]
neutrinoceros: okay sounds good !

chrishavlin:
Ok, so I rewrote all of `_guess_linthresh`. It felt clearer to avoid the whole abs(min,max) vs min(abs) confusion and treat the positive and negative ranges separately. It's maybe a little more verbose but I think it's easier to understand now.

yt/visualization/_handlers.py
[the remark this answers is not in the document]
chrishavlin: So I'm back to thinking I'd rather remove these lines... they're not needed in that sense that removing them does not result in errors. But it is true we'd be switching to SymLogNorm in situations that could technically use a LogNorm -- If we wanted to, we could allow a LogNorm when the data is >= 0 if matplotlib >= 3.5.2 **and** if we supply that new interpolation_stage keyword to imshow. So happy to implement that here if you think it's a good idea, but i'd rather wait til your #3849 is merged before doing so.

yt/visualization/_handlers.py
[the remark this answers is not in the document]
chrishavlin: _guess_linthresh will now always return a plain float

chrishavlin:
interesting... that failed test passed locally. Will have to dig in further...

neutrinoceros:
#3849 is in, now this should be rebased so we can see only the relevant bits in the diff

neutrinoceros:
I went for it since the rebase turned out a no op.

yt/visualization/_handlers.py
chummels: This is a smart algorithm.  It seems to match my solution from #3295 but also incorporate the fix for the sigdig error that my solution broke.  Nice way to find a solution!

chummels:
I like this solution overall.  The only thing I see missing from the PR is a brief mention of this in the colorbar linthresh narrative docs, like a sentence describing the linthresh guess (e.g., "linthresh uses the minimum absolute nonzero value but limits the dynamic range of the symlog scale when necessary to avoid floating point precision issues (10**15)").  Thanks for coding this up, Chris!

neutrinoceros (commented):
+1 for documentation. Specifically, the following paragraph should be updated
https://github.com/yt-project/yt/blob/dc7da9e8d56f583656361603275d3f9b55061f1d/doc/source/visualizing/plots.rst#L1000-L1005

I also note that the only 2 test failures we see now are 100% expected and desirable consequences of the change, so we can bump baselines here.

yt/visualization/_handlers.py
```diff
@@ -326,45 +326,24 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
         else:
-            if kw["vmin"] == kw["vmax"] or not np.any(np.isfinite(data)):
+            if kw["vmin"] == kw["vmax"] or not np.any(finite_values_mask):
                 norm_type = Normalize
             elif kw["vmin"] <= 0:
+                # note: matplotlib >= 3.5.2 could use LogNorm here if vmin==0
+                # if we do the following: set vmin to the min nonzero value and
+                # supply the interpolation_stage='rgba' keyword arg to imshow.
+                # this would, however, change default behavior and break answer
+                # tests, so needs a wider discussion.
```
neutrinoceros, line 336: The expected gain is not clear to me from just this comment.
If you think a discussion is still needed after this PR is merged, we could keep #3944 open for discussion,
but I'm not sure I agree of this being a temporary comment with no clear expiration condition.

yt/visualization/_handlers.py
```diff
@@ -374,6 +353,54 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
+        pos_min, pos_max = get_minmax(finite_plot_data[finite_plot_data > 0])
+        neg_min, neg_max = get_minmax(finite_plot_data[finite_plot_data < 0])
+
+        has_pos = pos_min is not None
+        has_neg = neg_min is not None
+
+        # the starting guess is the absolute value of the point closest to 0
+        # (remember: neg_max is closer to 0 than neg_min)
+        if has_pos and has_neg:
+            linthresh = np.min((np.abs(neg_max), pos_min))
```
neutrinoceros, line 376: Unless you think it's less clear (builtin `abs` would also work btw)
```suggestion
            linthresh = np.min((-neg_max, pos_min))
```

yt/visualization/_handlers.py
```diff
@@ -374,6 +353,54 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
+        elif has_pos:
+            linthresh = pos_min
+        elif has_neg:
+            linthresh = np.abs(neg_max)
```
neutrinoceros, line 380: Same as above
```suggestion
            linthresh = -neg_max
```

yt/visualization/_handlers.py
```diff
@@ -374,6 +353,54 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
+        else:
+            # this condition should handled before here in get_norm
```
neutrinoceros, line 382: I would probably not work out of the box here, but FTR this looks like a place where `typing.assert_never` would be useful. (Python 3.11+ only)
```suggestion
            # this condition should be handled before here in get_norm
```

yt/visualization/_handlers.py
```diff
@@ -374,6 +353,54 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
+            raise RuntimeError("No negative or positive values")
```
neutrinoceros, line 383: Isn't the only way to end up here that there are no finite data points ?

yt/visualization/_handlers.py
```diff
@@ -374,6 +353,54 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
+
+        log10_linthresh = np.log10(linthresh)
+
+        # if either the pos or neg ranges exceed cutoff_sigdigs, then
+        # linthresh is shifted to decrease the range to avoid floating point
+        # precision errors in the normalization.
+        cutoff_sigdigs = 15.0  # max allowable range in significant digits
```
neutrinoceros, line 390: An integer feels more appropriate
```suggestion
        cutoff_sigdigs = 15  # max allowable range in significant digits
```

yt/visualization/_handlers.py
```diff
@@ -374,6 +353,54 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
+        if has_pos and np.log10(pos_max) - log10_linthresh > cutoff_sigdigs:
+            linthresh = pos_max / (10.0**cutoff_sigdigs)
+            log10_linthresh = np.log10(linthresh)
+
+        if has_neg and np.log10(np.abs(neg_min)) - log10_linthresh > cutoff_sigdigs:
```
neutrinoceros, line 395: Same as above
```suggestion
        if has_neg and np.log10(-neg_min) - log10_linthresh > cutoff_sigdigs:
```

yt/visualization/_handlers.py
```diff
@@ -374,6 +353,54 @@ def get_norm(self, data: np.ndarray, *args, **kw) -> Normalize:
...
+            linthresh = np.abs(neg_min) / (10.0**cutoff_sigdigs)
+
+        if hasattr(linthresh, "units"):
```
neutrinoceros, line 398: Unlikely that as astropy units array could ever make it to this line, and if it does, we don't want to let it pass, do we ?
```suggestion
        if isinstance(linthresh, unyt_quantity):
```

yt/visualization/tests/test_plotwindow.py
```diff
@@ -768,15 +768,57 @@ def test_symlog_min_zero():
...
+    arr2 = np.full(shape, 5.0e-324)
+    arr2[0, 0] = -1e12
+
+    arr3 = arr.copy()
+    arr3[4, 4] = 0.0
+
+    d = {"scalar_spans_0": arr, "tiny_vmax": arr2, "scalar_tiny_with_0": arr3}
+
+    ds = load_uniform_grid(d, shape)
+    for field in d.keys():
```
neutrinoceros, line 786:
```suggestion
    for field in d:
```

yt/visualization/tests/test_plotwindow.py
```diff
@@ -768,15 +768,57 @@ def test_symlog_min_zero():
...
+    arr = np.full(shape, -1e30)
+    arr[1, 1] = -1e27
+    arr[2, 2] = 1e-12
+    arr[3, 3] = 1e-10
+
+    arr2 = -1 * arr.copy()  # also check the reverse
+    d = {"linthresh_gt_vmax": arr, "linthresh_lt_vmin": arr2}
+
+    ds = load_uniform_grid(d, shape)
+    for field in d.keys():
```
neutrinoceros, line 805:
```suggestion
    for field in d:
```

neutrinoceros (commented):
this is excellent. I got a couple suggestions, mostly nits.

yt/visualization/_handlers.py
chrishavlin: No need to keep 3944 open -- I'll rephrase this comment to be more of a historical note rather than a "to do" (which is how it currently sounds).

yt/visualization/_handlers.py
chrishavlin: I actually think it's a little **more** clear with the `-`! Will change it.

yt/visualization/_handlers.py
chrishavlin: oh, yes. I can switch this to read `no finite data points`.

neutrinoceros:
@chrishavlin there's a missing import reported by flake8

chrishavlin:
Latest batch updates the docs, bumps the answers and incorporates the remaining suggestions. I'll go through and close out obviously outdated threads...

yt/visualization/_handlers.py
chrishavlin: The code snippet rendered above is out of date. The re-written version is the same end result (matching 3295 except in the edge cases), but I re-wrote it to handle the negative and positive ranges sequentially cause it made handling edge cases easier. Just wanted to double check that you saw that!

neutrinoceros (approved):
LGTM. I'll give @chummels a couple days in case he wants to review this.

chummels (approved):
Looks good!  Thanks!

PR #3989, opened 2022-06-29
ENH: Projections of a field's standard deviation

jzuhone:
## PR Summary

Weighted projections are essentially averages along the line of sight of a field. For some fields (like velocity), projecting the weighted standard deviation is also a useful quantity. This PR adds that functionality for both on-axis and off-axis projections to yt, including `ProjectionPlot`s and projections to FITS images. The functionality is demonstrated in this notebook: 

https://gist.github.com/388e2b53d1b848fc19a125669d422cec

Still need to add docs ~~and tests~~. When viewing the diff, you may want to turn off changes in whitespace since I fixed the indentation of a bunch of docstrings. 

I'm not entirely crazy about what I had to do in order to take care of cancellations in the computation of the standard deviation in the final map--basically if we end up with a negative difference between the projection of the quantity squared and the square of the projected quantity, and it can be reasonably attributed to roundoff error, we simply set the difference to zero. There are ways to compute the standard deviation which avoids the cancellation, but doing it here would actually be pretty invasive as far as the projection machinery goes. 

I also fixed a number of issues I found with documentation regarding projections. 

## PR Checklist

### DIFF
--- a/doc/source/analyzing/objects.rst
+++ b/doc/source/analyzing/objects.rst
@@ -609,7 +609,7 @@ This is equivalent to:
    proj = ds.proj(("gas", "density"), "x", data_source=reg, method="min")
    proj.plot()
 
-You can also compute the ``mean`` value, which accepts a field, axis and weight
+You can also compute the ``mean`` value, which accepts a field, axis, and weight
 function.  If the axis is not specified, it will return the average value of
 the specified field, weighted by the weight argument.  The weight argument
 defaults to ``ones``, which performs an arithmetic average.  For instance:
@@ -636,6 +636,34 @@ If an axis is provided, it will project along that axis and return it to you:
 
    rho_proj = reg.mean(("gas", "temperature"), axis="y", weight=("gas", "density"))
    rho_proj.plot()
+
+You can also compute the ``std`` (standard deviation), which accepts a field,
+axis, and weight function. If the axis is not specified, it will
+return the standard deviation of the specified field, weighted by the weight
+argument.  The weight argument defaults to ``ones``. For instance:
+
+.. code-block:: python
+
+   std_rho = reg.std(("gas", "density"))
+   std_rho_by_vol = reg.std(("gas", "density"), weight=("gas", "cell_volume"))
+
+This is equivalent to:
+
+.. code-block:: python
+
+   std_rho = reg.quantities.weighted_standard_deviation(
+       ("gas", "density"), weight_field=("index", "ones")
+   )
+   std_rho_by_vol = reg.quantities.weighted_standard_deviation(
+       ("gas", "density"), weight_field=("gas", "cell_volume")
+   )
+
+If an axis is provided, it will project along that axis and return it to you:
+
+.. code-block:: python
+
+   vy_std = reg.std(("gas", "velocity_y"), axis="y", weight=("gas", "density"))
+   vy_std.plot()
 
 The ``sum`` function will add all the values in the data object.  It accepts a
 field and, optionally, an axis.  If the axis is left unspecified, it will sum
--- a/doc/source/analyzing/objects.rst
+++ b/doc/source/analyzing/objects.rst
@@ -591,11 +591,23 @@ This is equivalent to:
 
 .. code-block:: python
 
-   proj = ds.proj(("gas", "density"), "x", data_source=reg, method="mip")
+   proj = ds.proj(("gas", "density"), "x", data_source=reg, method="max")
    proj.plot()
 
-The ``min`` operator does not do this, however, as a minimum intensity
-projection is not currently implemented.
+The same can be done with the ``min`` operation, computing a minimum
+intensity projection:
+
+.. code-block:: python
+
+   proj = reg.min(("gas", "density"), axis="x")
+   proj.plot()
+
+This is equivalent to:
+
+.. code-block:: python
+
+   proj = ds.proj(("gas", "density"), "x", data_source=reg, method="min")
+   proj.plot()
 
 You can also compute the ``mean`` value, which accepts a field, axis and weight
 function.  If the axis is not specified, it will return the average value of
--- a/doc/source/visualizing/plots.rst
+++ b/doc/source/visualizing/plots.rst
@@ -394,10 +394,16 @@ field is mathematically given by:
 in order to make a weighted projection of the standard deviation of a field
 along a line of sight, the ``moment`` keyword argument should be set to 2.
 
-``mip``:
+``max``:
 ++++++++
 
 This method picks out the maximum value of a field along the line of
+sight given by the axis parameter.
+
+``min``:
+++++++++
+
+This method picks out the minimum value of a field along the line of
 sight given by the axis parameter.
 
 ``sum``:
--- a/doc/source/visualizing/plots.rst
+++ b/doc/source/visualizing/plots.rst
@@ -343,46 +343,73 @@ Types of Projections
 """"""""""""""""""""
 
 There are several different methods of projections that can be made either
-when creating a projection with ds.proj() or when making a ProjectionPlot.
+when creating a projection with :meth:`~yt.static_output.Dataset.proj` or
+when making a :class:`~yt.visualization.plot_window.ProjectionPlot`.
 In either construction method, set the ``method`` keyword to be one of the
 following:
 
-``integrate`` (unweighted)
-    This is the default projection method. It simply integrates the
-    requested field  :math:`f(x)` along a line of sight  :math:`\hat{n}` ,
-    given by the axis parameter (e.g. :math:`\hat{i},\hat{j},` or
-    :math:`\hat{k}`).  The units of the projected field
-    :math:`g(X)` will be the units of the unprojected field  :math:`f(x)`
-    multiplied by the appropriate length unit, e.g., density in
-    :math:`\mathrm{g\ cm^{-3}}` will be projected to  :math:`\mathrm{g\ cm^{-2}}`.
+``integrate`` (unweighted):
++++++++++++++++++++++++++++
+
+This is the default projection method. It simply integrates the
+requested field :math:`f({\bf x})` along a line of sight :math:`\hat{\bf n}` ,
+given by the axis parameter (e.g. :math:`\hat{\bf i},\hat{\bf j},` or
+:math:`\hat{\bf k}`). The units of the projected field
+:math:`g({\bf X})` will be the units of the unprojected field :math:`f({\bf x})`
+multiplied by the appropriate length unit, e.g., density in
+:math:`\mathrm{g\ cm^{-3}}` will be projected to  :math:`\mathrm{g\ cm^{-2}}`.
 
 .. math::
 
-    g(X) = {\int\ {f(x)\hat{n}\cdot{dx}}}
+    g({\bf X}) = {\int\ {f({\bf x})\hat{\bf n}\cdot{d{\bf x}}}}
 
-``integrate`` (weighted)
-    When using the ``integrate``  method, a ``weight_field`` argument may also
-    be specified, which will produce a weighted projection.  :math:`w(x)`
-    is the field used as a weight. One common example would
-    be to weight the "temperature" field by the "density" field. In this case,
-    the units of the projected field are the same as the unprojected field.
+``integrate`` (weighted):
++++++++++++++++++++++++++
+
+When using the ``integrate``  method, a ``weight_field`` argument may also
+be specified, which will produce a weighted projection. :math:`w({\bf x})`
+is the field used as a weight. One common example would
+be to weight the "temperature" field by the "density" field. In this case,
+the units of the projected field are the same as the unprojected field.
 
 .. math::
 
-    g(X) = \frac{\int\ {f(x)w(x)\hat{n}\cdot{dx}}}{\int\ {w(x)\hat{n}\cdot{dx}}}
+    g({\bf X}) = \int\ {f({\bf x})\tilde{w}({\bf x})\hat{\bf n}\cdot{d{\bf x}}}
 
-``mip``
-    This method picks out the maximum value of a field along the line of
-    sight given by the axis parameter.
+where the "~" over :math:`w({\bf x})` reflects the fact that it has been normalized
+like so:
 
-``sum``
-    This method is the same as ``integrate``, except that it does not
-    multiply by a path length when performing the integration, and is just a
-    straight summation of the field along the given axis. The units of the
-    projected field will be the same as those of the unprojected field. This
-    method is typically only useful for datasets such as 3D FITS cubes where
-    the third axis of the dataset is something like velocity or frequency, and
-    should _only_ be used with fixed-resolution grid-based datasets.
+.. math::
+
+    \tilde{w}({\bf x}) = \frac{w({\bf x})}{\int\ {w({\bf x})\hat{\bf n}\cdot{d{\bf x}}}}
+
+For weighted projections using the ``integrate`` method, it is also possible
+to project the standard deviation of a field. In this case, the projected
+field is mathematically given by:
+
+.. math::
+
+    g({\bf X}) = \left[\int\ {f({\bf x})^2\tilde{w}({\bf x})\hat{\bf n}\cdot{d{\bf x}}} - \left(\int\ {f({\bf x})\tilde{w}({\bf x})\hat{\bf n}\cdot{d{\bf x}}}\right)^2\right]^{1/2}
+
+in order to make a weighted projection of the standard deviation of a field
+along a line of sight, the ``moment`` keyword argument should be set to 2.
+
+``mip``:
+++++++++
+
+This method picks out the maximum value of a field along the line of
+sight given by the axis parameter.
+
+``sum``:
+++++++++
+
+This method is the same as ``integrate``, except that it does not
+multiply by a path length when performing the integration, and is just a
+straight summation of the field along the given axis. The units of the
+projected field will be the same as those of the unprojected field. This
+method is typically only useful for datasets such as 3D FITS cubes where
+the third axis of the dataset is something like velocity or frequency, and
+should *only* be used with fixed-resolution grid-based datasets.
 
 .. _off-axis-projections:
 
@@ -444,8 +471,11 @@ to project along, and a field to project.  For example:
 
 ``OffAxisProjectionPlot`` objects can also be created with a number of
 keyword arguments, as described in
-:class:`~yt.visualization.plot_window.OffAxisProjectionPlot`
+:class:`~yt.visualization.plot_window.OffAxisProjectionPlot`.
 
+Like on-axis projections, the projection of the standard deviation
+of a weighted field can be created by setting ``moment=2`` in the call
+to :class:`~yt.visualization.plot_window.ProjectionPlot`.
 
 .. _slices-and-projections-in-spherical-geometry:
 
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -173,6 +173,7 @@ class YTProj(YTSelectionContainer2D):
         method="integrate",
         field_parameters=None,
         max_level=None,
+        moment=1,
         *,
         style=None,
     ):
@@ -208,6 +209,12 @@ class YTProj(YTSelectionContainer2D):
             self.func = np.sum  # for the future
         else:
             raise NotImplementedError(self.method)
+        if moment == 2 and weight_field is None:
+            raise RuntimeError(
+                "Cannot compute the second moment of a projection "
+                "if weight_field=None!"
+            )
+        self.moment = moment
         self._set_center(center)
         self._projected_units = {}
         if data_source is None:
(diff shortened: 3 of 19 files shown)


yt/data_objects/construction_data_containers.py
```diff
@@ -495,6 +534,7 @@ def __init__(
         method="integrate",
         field_parameters=None,
         max_level=None,
+        moment=1,
         *,
```
neutrinoceros, line 538:
```suggestion
        *,
        moment=1,
```

yt/data_objects/construction_data_containers.py
```diff
@@ -401,6 +434,7 @@ def __init__(
...
```
neutrinoceros, line 438:
```suggestion
        *,
        moment=1,
```

yt/visualization/plot_window.py
```diff
@@ -2044,124 +2046,128 @@ class AxisAlignedProjectionPlot(ProjectionPlot, PWViewerMPL):
...
 
     right_handed : boolean
         Depreceated, please use flip_horizontal callback.
         Whether the implicit east vector for the image generated is set to make a right
         handed coordinate system with a north vector and the normal vector, the
         direction of the 'window' into the data.
     data_source : YTSelectionContainer Object
-         Object to be used for data selection.  Defaults to a region covering
-         the entire simulation.
+        Object to be used for data selection.  Defaults to a region covering
```
neutrinoceros, line 2128: is this broken ?
I personally don't fancy obscuring information from git blame with this kind of changes.

neutrinoceros (commented):
super quick comments. I'm game to review this more in depth when it's ready, this looks really cool !
also FYI, CI is currently unstable (see discussion in #3988), but it should be resolved upstream in a few hours/days.

yt/data_objects/construction_data_containers.py
jzuhone: We need a quick explanation somewhere of why the use of the asterisk is so important and the rationale for putting kwargs before and after it is. Most developers are not fully familiar with it yet (I am myself not) and we need clarification on whether this is a style choice or a choice that the project needs to be implementing across the board.

jzuhone: I realize this is not exactly a "new" python feature either--it's just not one that we've really required yt developers to use.

neutrinoceros: It may not be new, but it *does* require Python 3, so it wasn't something we *could* use in yt until early 2020 when we dropped Python 2, which is why it's currently very sporadically used.
Anyway I'd happily open a discussion or explain my general observations/advice about this, where would be a good place for this ?

yt/visualization/plot_window.py
neutrinoceros: ok then I suggest adding these commits to .git-blame-ignore-revs

yt/data_objects/construction_data_containers.py
jzuhone: I'd just open a github issue

neutrinoceros: here you go #3992

yt/funcs.py
```diff
@@ -1324,3 +1324,15 @@ def levenshtein_distance(seq1, seq2, max_dist=None):
         if matrix[x].min() > max_dist:
             return max_dist + 1
     return matrix[size_x - 1, size_y - 1]
+
+
+def validate_moment(moment, weight_field):
+    if moment == 2 and weight_field is None:
+        raise RuntimeError(
+            "Cannot compute the second moment of a projection " "if weight_field=None!"
```
neutrinoceros, line 1332:
```suggestion
            "Cannot compute the second moment of a projection if weight_field=None!"
```

yt/visualization/fits_image.py
```diff
@@ -1371,8 +1382,9 @@ def __init__(
         depth=(1.0, "unitary"),
         method="integrate",
         length_unit=None,
+        moment=1,
```
neutrinoceros, line 1385:
```suggestion
        *,
        moment=1,
```

yt/visualization/tests/test_offaxisprojection.py
```diff
@@ -101,3 +102,38 @@ def test_field_cut_off_axis_octree():
...
+        weight_field=("gas", "density"),
+        moment=1,
+    )
+    p2 = OffAxisProjectionPlot(
+        ds,
+        [1, 1, 1],
+        ("gas", "velocity_los"),
+        weight_field=("gas", "density"),
+        moment=2,
+    )
```
neutrinoceros, line 132: maybe we could use a smaller buffer than the default (800x800) to mimize exec time ?

jzuhone: Done

yt/data_objects/construction_data_containers.py
```diff
@@ -240,12 +251,32 @@ def field(self):
...
+                def _sq_field(field, data):
+                    return data[fname] ** 2
+
+                return _sq_field
+
+            for field in fields:
+                fd = self.ds._get_field_info(*field)
+                self.ds.add_field(
+                    (field[0], f"tmp_{field[1]}_squared"),
+                    make_sq_field(field),
```
neutrinoceros, line 267: This is fragile, using `functools.partial` for avoid late-binding issues, see https://github.com/yt-project/yt/pull/3996

yt/data_objects/construction_data_containers.py
```diff
@@ -308,14 +339,20 @@ def get_data(self, fields=None):
...
+            fvals = field_data[fi].ravel()
+            if self.moment == 2:
+                fvals = compute_stddev_image(field_data[fi + nfields].ravel(), fvals)
+            self[field] = self.ds.arr(fvals, input_units)
         for i in list(data.keys()):
             self[i] = data.pop(i)
         mylog.info("Projection completed")
+        if self.moment == 2:
+            for field in sfields:
+                self.ds.field_info.pop(field)
```
neutrinoceros, line 355: I'm worried that this isn't necessarily sufficient to completely restore the initial state (in particular in the context of #3948). Unless there's a strong argument to do so, I'd suggest just leaving the field be.

yt/data_objects/data_containers.py
```diff
@@ -1257,7 +1257,7 @@ def sum(self, field, axis=None):
             raise NotImplementedError(f"Unknown axis {axis}")
         return r
 
-    def integrate(self, field, weight=None, axis=None):
+    def integrate(self, field, weight=None, axis=None, moment=1):
```
neutrinoceros, line 1260:
```suggestion
    def integrate(self, field, weight=None, axis=None, *, moment=1):
```

yt/data_objects/construction_data_containers.py
jzuhone: We do this elsewhere, see here for the temporary field creation:

https://github.com/yt-project/yt/blob/89073a9b6d91381aff424d977491ef132427456d/yt/visualization/volume_rendering/off_axis_projection.py#L316-L337
and here for its removal:

https://github.com/yt-project/yt/blob/89073a9b6d91381aff424d977491ef132427456d/yt/visualization/volume_rendering/off_axis_projection.py#L402-L403

I'd much prefer a better way of removing fields, but that needs a separate PR. I frankly think it is better to remove it--the `FieldInfoContainer` is pretty jam-packed as is in most situations, and reducing clutter would be ideal.

neutrinoceros: I agree it shouldn't be necessary for this PR, and I wasn't aware we already had examples of this, so thank you for pointing it out. Can you open an issue to remind ourselves that we should better define/encapsulate a method to remove fields ?

yt/visualization/plot_window.py
jzuhone: These changes are unfortunately interwined with other docstring edits that are part of the PR's changes itself in the same commits, so probably best to leave them alone.

neutrinoceros: unfortunate. I'd offer to rewrite this branch' history but it doesn't seem worth the risk.

yt/data_objects/construction_data_containers.py
neutrinoceros: done here: #4082

neutrinoceros (dismissed):
I guess we'd need someone else to sign off before we can merge

jzuhone:
this needs docs--I just need time to write them up

neutrinoceros:
oh sorry I forgot about that. I'll review again when it's done

jzuhone:
This PR is ready for review.

doc/source/analyzing/objects.rst
```diff
@@ -591,13 +591,25 @@ This is equivalent to:
 
 .. code-block:: python
 
-   proj = ds.proj(("gas", "density"), "x", data_source=reg, method="mip")
+   proj = ds.proj(("gas", "density"), "x", data_source=reg, method="max")
```
neutrinoceros, line 594: should we deprecate `"mip"` ?

doc/source/analyzing/objects.rst
```diff
@@ -625,6 +637,34 @@ If an axis is provided, it will project along that axis and return it to you:
...
+   std_rho_by_vol = reg.quantities.weighted_standard_deviation(
+       ("gas", "density"), weight_field=("gas", "cell_volume")
+   )
+
+If an axis is provided, it will project along that axis and return it to you:
+
+.. code-block:: python
+
+   vy_std = reg.std(("gas", "velocity_y"), axis="y", weight=("gas", "density"))
+   vy_std.plot()
```
neutrinoceros, line 666: this paragraph is so similar to the previous one, I think we should merge them

doc/source/visualizing/plots.rst
```diff
@@ -343,46 +343,79 @@ Types of Projections
...
 
-``integrate`` (unweighted)
-    This is the default projection method. It simply integrates the
-    requested field  :math:`f(x)` along a line of sight  :math:`\hat{n}` ,
-    given by the axis parameter (e.g. :math:`\hat{i},\hat{j},` or
-    :math:`\hat{k}`).  The units of the projected field
-    :math:`g(X)` will be the units of the unprojected field  :math:`f(x)`
-    multiplied by the appropriate length unit, e.g., density in
-    :math:`\mathrm{g\ cm^{-3}}` will be projected to  :math:`\mathrm{g\ cm^{-2}}`.
+``integrate`` (unweighted):
```
neutrinoceros, line 351: Not sure why but the `:math:` directive didn't render correctly in the build

Screenshot 2022-09-05 at 09 20 27

yt/funcs.py
```diff
@@ -1332,3 +1332,15 @@ def levenshtein_distance(seq1, seq2, max_dist=None):
         if matrix[x].min() > max_dist:
             return max_dist + 1
     return matrix[size_x - 1, size_y - 1]
+
+
+def validate_moment(moment, weight_field):
```
neutrinoceros, line 1337: not sure `yt.funcs` is a good home for this function. Do you expect it to be adopted downstream ? if not, I'd suggest putting it in a more private space

yt/utilities/math_utils.py
```diff
@@ -1541,3 +1541,43 @@ def get_sph_theta_component(vectors, theta, phi, normal):
...
+    It defines the standard deviation as sigma^2 = <v^2>-<v>^2, where the
+    brackets indicate averages (with the weight) and v is the field in
+    question.
+
+    There is the possibility that if the projection at a particular location
+    is of a constant or a single cell/particle, then <v^2> = <v>^2 and instead
+    of getting zero one gets roundoff error that results in sigma^2 < 0,
+    which is unphysical.
+
+    We handle this here by performing the subtraction and checking that any
```
neutrinoceros, line 1558: thanks a lot for this explanation, that's really helpful

yt/visualization/fits_image.py
```diff
@@ -1067,6 +1070,10 @@ class FITSProjection(FITSImageData):
     length_unit : string, optional
         the length units that the coordinates are written in. The default
         is to use the default length unit of the dataset.
+    moment : integer, optional
```
neutrinoceros, line 1073: I suggest we avoid mixing two different meaning of "optional" in the context of Python function signatures.
In terms of typing, `Optional` means "possibly `None`", but here it is not the case. (FTR this rather confusing name can be avoided in Python 3.9, but for the time being we _do_ use it in yt)
```suggestion
    moment : integer
```

yt/visualization/fits_image.py
```diff
@@ -1355,6 +1365,10 @@ class FITSOffAxisProjection(FITSImageData):
...
```
neutrinoceros, line 1368: same as above
```suggestion
    moment : integer
```

yt/visualization/fits_image.py
```diff
@@ -1079,12 +1086,15 @@ def __init__(
         width=None,
         weight_field=None,
         length_unit=None,
+        moment=1,
```
neutrinoceros, line 1089: For api consistency
```suggestion
        *,
        moment=1,
```

yt/visualization/fits_image.py
```diff
@@ -1381,21 +1397,55 @@ def __init__(
             image_res = (image_res, image_res)
         res = (image_res[0], image_res[1])
         if data_source is None:
-            source = ds
+            source = ds.all_data()
```
neutrinoceros, line 1400: seems like a good idea, but is there a reason for doing this change specifically in this PR that I am not seeing ?

yt/visualization/fits_image.py
```diff
@@ -1381,21 +1397,55 @@ def __init__(
...
         else:
             source = data_source
-        for field in fields:
-            buf[field] = off_axis_projection(
+        fields = source._determine_fields(list(iter_fields(fields)))
+        for fname in fields:
+            buf[fname] = off_axis_projection(
```
neutrinoceros, line 1405: I don't get this change. Doesn't `_determine_fields` yield field tuples ?

yt/visualization/fixed_resolution.py
```diff
@@ -614,6 +616,39 @@ def __getitem__(self, item):
             north_vector=dd.north_vector,
             method=dd.method,
         )
+        if self.data_source.moment == 2:
+
+            def _sq_field(field, data, fname: Tuple[str, str]):
+                return data[fname] ** 2
```
neutrinoceros, line 622: I'm confused; I thought in most places, `fname` is meant to designate a single `str`, not the complete field tuple

yt/visualization/plot_window.py
```diff
@@ -1198,6 +1206,8 @@ def _setup_plots(self):
 
             if colorbar_label is None:
                 colorbar_label = image.info["label"]
+                if getattr(self, "moment", 1) == 2:
+                    colorbar_label = "%s Standard Deviation" % colorbar_label
```
neutrinoceros, line 1210:
```suggestion
                    colorbar_label = "%s \\rm{Standard Deviation}" % colorbar_label
```

yt/visualization/plot_window.py
```diff
@@ -2289,7 +2306,9 @@ def __init__(
         north_vector=None,
         method="integrate",
         data_source=None,
+        moment=1,
```
neutrinoceros, line 2309:
```suggestion
        *,
        moment=1,
```

yt/visualization/plot_window.py
```diff
@@ -2049,6 +2063,7 @@ def __init__(
         window_size=8.0,
         buff_size=(800, 800),
         aspect=None,
+        moment=1,
         *,
```
neutrinoceros, line 2067:
```suggestion
        *,
        moment=1,
```

yt/visualization/volume_rendering/off_axis_projection.py
```diff
@@ -325,7 +325,10 @@ def temp_weightfield(field, data):
             return temp_weightfield
 
         data_source.ds.field_info.add_field(
-            weightfield, sampling_type="cell", function=_make_wf(item, weight)
+            weightfield,
+            sampling_type="cell",
+            function=_make_wf(item, weight),
+            units="",
```
neutrinoceros, line 331: why this change ?

neutrinoceros (commented):
In general I really like this a lot. However, on a general level, I think the docs should acknowledge that this functionality is only supported for cartesian geometries at the moment. I would also recommend adding a test to check that the error message one gets when calling `moment=2` for e.g. a cylindrical geometry is sound and clear.

cphyc (commented):
I haven't reviewed all files in detail but it seems to be working nicely :)
However, when saving the plot, I noticed the filename is
`{ds_name}_Projection_{axis}_{field}_{weight_field}.{extension}`.
Shouln't it be something like 
`{ds_name}_Projection_{axis}_{field}_{weight_field}_standard_deviation.{extension}`
?

doc/source/analyzing/objects.rst
jzuhone: I saw deprecation messages for "mip" already in certain places, not sure if there isn't one here already or not.

neutrinoceros: Oh that's right, I forgot about it. Nothing to do here !

doc/source/visualizing/plots.rst
jzuhone: Funny, this resolved OK for me on my laptop.

yt/visualization/fits_image.py
jzuhone: FWIW this is part of the `numpydoc` docstring standard: 

https://numpydoc.readthedocs.io/en/latest/format.html

and we've used it in plenty of other places in yt. I wasn't aware that we started discouraging this at any point.

yt/visualization/fits_image.py
jzuhone: I can't remember why I did this anymore exactly--I somewhat recall an error of some sort

yt/visualization/volume_rendering/off_axis_projection.py
jzuhone: For some reason `pre-commit` asked for it

yt/funcs.py
jzuhone: Do we have a standard for what should go in here? I'm honestly not certain of where else it should go. I'd obviously not want to create a single file for it.

yt/visualization/fixed_resolution.py
jzuhone: Yes, but the problem here is that there's a name collision--`field` is already in use in the function signature. If you've got any other suggestions I'm all ears

doc/source/analyzing/objects.rst
jzuhone: Ehh. I agree that they are very similar (in fact I copied, pasted, and edited to get this paragraph). But I feel like there are competing interests here. The first and second moments are of course different, and `std` was not documented separately prior to this (an oversight I assume), so I think we should leave it as-is.

yt/visualization/fits_image.py
neutrinoceros: then carry on. I'm more familiar with `typing` than I am with numpy docstring style

yt/data_objects/construction_data_containers.py
```diff
@@ -173,6 +182,7 @@ def __init__(
         method="integrate",
         field_parameters=None,
         max_level=None,
+        moment=1,
         *,
```
neutrinoceros, line 186: Sorry I missed that one
```suggestion
        *,
        moment=1,
```

yt/data_objects/construction_data_containers.py
```diff
@@ -419,6 +455,7 @@ def __init__(
             method,
             field_parameters,
             max_level,
+            moment,
```
neutrinoceros, line 458:
```suggestion
            moment=moment,
```

yt/data_objects/construction_data_containers.py
```diff
@@ -513,6 +555,7 @@ def __init__(
...
```
neutrinoceros, line 558:
```suggestion
            moment=moment,
```

yt/funcs.py
neutrinoceros: never mind, I remembered that we had a couple other `validate_*` methods but turns out they also live in this module. I've never been satisfied with how this module hosts both internal and user-facing functions but I guess it's a bit late to change it and it's certainly out of scope here

yt/visualization/fixed_resolution.py
neutrinoceros: I think `item` would make more sense in this context; Also, just curious, I don't think I've ever seen an example of a derived field where the first argument `field` was used, do you happen to know when it's useful ?

yt/visualization/volume_rendering/off_axis_projection.py
neutrinoceros: I mean, the formatting is just `black` being triggered by the line getting long, but you're also adding a `units=""` argument, and that's what confuses me.

doc/source/visualizing/plots.rst
neutrinoceros: probably nbd then, maybe the html artefacts are just broken

doc/source/analyzing/objects.rst
neutrinoceros: ok seems fair

yt/visualization/volume_rendering/off_axis_projection.py
jzuhone: The `units` here is dimensionless, so I'm just making it explicit here.

yt/visualization/fits_image.py
jzuhone: ah, now I remember--I needed this to be a data object so I could call `_determine_fields`, which we weren't doing before.

neutrinoceros:
@jzuhone The failure is real

neutrinoceros (approved):

neutrinoceros:
@cphyc do you wish to review this more in depth latter or should we advertise this some more to get a review from someone else ?

cphyc:
I'm not sure I can review this in the coming two weeks - so if you can find an extra pair of eyes that'd be great :)

jzuhone:
@matthewturk ?

matthewturk (approved):
I've reviewed these changes and they're great.  I don't see any issues that weren't addressed elsewhere.  Thank you, @jzuhone .

PR #3990, opened 2022-06-29
[TST] pyart deps

chrishavlin:
Experimenting with ways to avoid failing CI. 

First push just avoids the latest pyart (1.12.5).


chrishavlin:
For extra context here, see the discussion in #3988 , particularly  https://github.com/yt-project/yt/pull/3988#issuecomment-1169992330

neutrinoceros:
Thanks for looking into this, this seems reasonable and we don't even need to revert it when a new version is released.

PR #3996, opened 2022-07-05
BUG: fix late-binding errors in closures

neutrinoceros:
## PR Summary

Follow up to #3994
Upgrade flake8-bugbear and fix newly detected gotchas: late-binding errors in closures

If, like me, you didn't know about this problem, here's a helpful blog post to explain what the problem is and its solution(s)
https://docs.python-guide.org/writing/gotchas/#late-binding-closures

Content:

- MNT: manually upgrade flake8-bugbear
- BUG: fix late-binding errors in closures, detected with flake8-bugbear (B023)
- MNT: add noqas to flase positives (B023) in generator based tests

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
     hooks:
     - id: flake8
       additional_dependencies: [
-        flake8-bugbear==21.9.1,
+        flake8-bugbear==22.7.1,
         flake8-logging-format==0.6.0,
         flake8-2020==1.6.1,
       ]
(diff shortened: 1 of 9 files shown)


neutrinoceros:
Turns out most errors are in frontend-specific code

matthewturk:
Do we think these actually resulted in errors?

neutrinoceros:
I am honestly not sure, but it seems likely, and fragile.

matthewturk:
OK, I think the Gadget ones likely did not result in problems (although they should be fixed, as you have) but the RAMSES ones do seem like they may have caused issues *in the specific case* that there were multiple field types being read simultaneously.

neutrinoceros:
Yes, with just one field type the bug should be invisible

matthewturk (approved):

PR #3998, opened 2022-07-05
ENH: Update to fields used in Enzo answer testing

jwise77:
## PR Summary

This PR updates the field names to be tuples in the answer testing of the Enzo frontend. The `hubble_constant` parameter is also not necessarily always written as a float (i.e. 1), so it is read as an int, which causes problems in unyt downstream. Forcing it to be float at initialization.

### DIFF
--- a/yt/frontends/enzo/answer_testing_support.py
+++ b/yt/frontends/enzo/answer_testing_support.py
@@ -96,7 +96,6 @@ class ShockTubeTest:
     def __call__(self):
         # Read in the ds
         ds = load(self.data_file)
-        ds.setup_deprecated_fields()
         exact = self.get_analytical_solution()
 
         ad = ds.all_data()
--- a/yt/frontends/enzo/answer_testing_support.py
+++ b/yt/frontends/enzo/answer_testing_support.py
@@ -100,7 +100,7 @@ class ShockTubeTest:
         exact = self.get_analytical_solution()
 
         ad = ds.all_data()
-        position = ad["x"]
+        position = ad[("index", "x")]
         for k in self.fields:
             field = ad[k].d
             for xmin, xmax in zip(self.left_edges, self.right_edges):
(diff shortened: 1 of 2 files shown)


yt/frontends/enzo/answer_testing_support.py
```diff
@@ -122,8 +121,8 @@ def get_analytical_solution(self):
         pos, dens, vel, pres, inte = np.loadtxt(self.solution_file, unpack=True)
         exact = {}
         exact["pos"] = pos
-        exact["Density"] = dens
```
matthewturk, line 125: I'm not sure about these -- here we are explicitly using the on-disk fields, I think to prevent any unit conversions being made implicitly or explicitly.  The fields would become `"enzo", "Density"` etc.  Do you want them to be on-disk fields, or are you OK with the unit conversions being applied?

jwise77: Thanks for the comment! I was also worried about unit conversions and was trying to keep with the `"Enzo"` fields. I had to use `"gas"` in the end because the `Pressure` and `ThermalEnergy` fields aren't `"enzo"` fields in the shock tube tests.

To give context, I have just updated (enzo-project/enzo-dev#190) all of the tests in Enzo to be compatible with yt-4.1. The unit conversions don't cause a problem since everything is in cgs for these tests.

matthewturk: If you're happy, I'm happy.

matthewturk (approved):

yt/frontends/enzo/answer_testing_support.py
```diff
@@ -122,8 +121,8 @@ def get_analytical_solution(self):
         pos, dens, vel, pres, inte = np.loadtxt(self.solution_file, unpack=True)
         exact = {}
         exact["pos"] = pos
```
neutrinoceros, line 123: @jwise77 should this key be updated too or is there a reason not to ?

jwise77: It shouldn't be updated. This key is only used for interpolation purposes in the answer testing and not directly compared to the field data.

neutrinoceros: fair enough !

neutrinoceros (approved):

PR #4002, opened 2022-07-07
BUG: fix toroidal/poloidal magnetic fields registration for geometries where they are not implemented

### LINKED ISSUE #4000: ENH: missing def for poloidal/toroidal magnetic fields in polar geometry
### Bug report

**Bug summary**

Poloidal and toroidal magnetic fields definitions in polar geometry are missing, see https://github.com/yt-project/yt/blob/e16331088faf7296672ce9d1987a09d036f458ea/yt/fields/magnetic_field.py#L87-L149

neutrinoceros:
## PR Summary
Closes #4000
Turns out the existing implementation for cylindrical geometry is 100% reusable as is for polar geometry, so the fix is pretty trivial.

I'm also avoiding passing `function=None` to `add_derived_field` as this use case is intended for cases where `derived_field` is used as a decorator. See https://github.com/yt-project/yt/blob/e16331088faf7296672ce9d1987a09d036f458ea/yt/fields/field_info_container.py#L363-L369

### DIFF
--- a/yt/fields/magnetic_field.py
+++ b/yt/fields/magnetic_field.py
@@ -84,6 +84,9 @@ def setup_magnetic_field_fields(registry, ftype="gas", slice_info=None):
         units=unit_system["pressure"],
     )
 
+    _magnetic_field_toroidal = None
+    _magnetic_field_poloidal = None
+
     if registry.ds.geometry == "cartesian":
 
         def _magnetic_field_poloidal(field, data):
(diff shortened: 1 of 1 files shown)


neutrinoceros:
@jzuhone it's been a long time but it seems you were the last person to edit this part, maybe you'd like to take a look ?

jzuhone (approved):

PR #4004, opened 2022-07-08
BUG: Use format string 'd' not 'i' in grids

matthewturk:
The current behavior results in an error:

```
File "/home/mturk/yt/yt/yt/data_objects/index_subobjects/grid_patch.py", line 146, in __repr__                                                                                                                        return f"{cls_name}_{self.id:04i} ({self.ActiveDimensions})"                                                                                                                                                    ValueError: Unknown format code 'i' for object of type 'int'
```

This fixes that error.


cphyc (approved):

neutrinoceros:
@cphyc this doesn’t need backporting as the bug was only on the 4.1 branch

PR #4005, opened 2022-07-08
BUG: Make sure Boxlib DLE/DRE are correct index

### LINKED ISSUE #3993: BUG: yt with WarpX: field data can become 0 if particle data has already been read
### Bug report

**Bug summary**

In some cases, the field output of a WarpX simulation read from yt is exactly 0 if the particle data has been read before. I've only observed this issue in 2D simulations for which the lower boundary of the simulation domain in the second direction is <= -1 and the higher boundary is >= 1. This issue has already been posted in the WarpX github repo: https://github.com/ECP-WarpX/WarpX/issues/3136 

**Code for reproduction**

For completeness, I attach here a full reproducer, which includes a bash script containing all the commands needed to reproduce the issue (including the compilation of WarpX). [yt_bug_reproducer.zip](https://github.com/yt-project/yt/files/9021455/yt_bug_reproducer.zip)

I show below a typical failing python script

```python
import sys
import numpy as np
import yt

fn = sys.argv[1]
ds = yt.load(fn)

## If I comment the two lines below then the issue disappears
ad_particles = ds.all_data()
px = ad_particles['electrons', 'particle_momentum_x'].to_ndarray()

ad_field = ds.covering_grid(level=0, left_edge=ds.domain_left_edge, dims=ds.domain_dimensions)
Ex = ad_field[('boxlib','Ex')].squeeze().v
print("Max of Ex field after reading particle data")
print(np.amax(np.abs(Ex)))
```

**Actual outcome**

When the issue is present, the printed output is exactly 0.

**Expected outcome**

The printed output should be strictly greater than 0 (and is strictly greater than 0 if the 2 middle lines in the code snippet are commented out).

**Version Information**

  * Operating System: Linux (Debian)
  * Python Version: 3.8.8
  * yt version: 3.6.1 (but I've also observed the issue with version 4.0.1)
  * Other Libraries (if applicable): I'm using the current development version of WarpX (commit 542ef3c09b467974b5bd9d1e3b25a208e3296562)

yt and python have been installed from conda with the conda-forge channel.

matthewturk:
Fixes #3993.

This change uses the correct index for the z dimension when setting less-than-three-dimension boundaries.

The cascading set of things that happen as a result of this include the dds[2] set to 2.0, rather than 1.0, which in the particular cases identified by @NeilZaim set our point/grid selection to exclude the underlying grid.  With this change, we correctly set the grid *cell* size to not exceed the *domain* size, and now the selection process works OK.

The reason this was caught by the specific instance identified in the reproducer script was that we called the grid selector for setting the base chunk for `ad_particles` in that script, but then used `particle_select` rather than grid select, so the cached mask was incorrect.  When using the covering grid selection, we select based on integer indexes, which normally work, *unless* the mask is set to "none included" which was the case here.

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -142,7 +142,7 @@ answer_tests:
     - yt/frontends/boxlib/tests/test_outputs.py:test_units_override
     - yt/frontends/boxlib/tests/test_outputs.py:test_raw_fields
 
-  local_boxlib_particles_009:
+  local_boxlib_particles_010:
     - yt/frontends/boxlib/tests/test_outputs.py:test_LyA
     - yt/frontends/boxlib/tests/test_outputs.py:test_nyx_particle_io
     - yt/frontends/boxlib/tests/test_outputs.py:test_castro_particle_io
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -372,7 +372,7 @@ class BoxlibHierarchy(GridIndex):
             if self.dimensionality < 2:
                 dx[i].append(DRE[1] - DLE[1])
             if self.dimensionality < 3:
-                dx[i].append(DRE[2] - DLE[1])
+                dx[i].append(DRE[2] - DLE[2])
         self.level_dds = np.array(dx, dtype="float64")
         next(header_file)
         if self.ds.geometry == "cartesian":


neutrinoceros:
Regardless the complex (and interesting) history of how it was discovered, the fix looks very obviously correct to me. We just need to bump some box lib answers.

neutrinoceros (dismissed):

neutrinoceros (approved):

PR #4011, opened 2022-07-12
BUG: fix for composite volume rendering with octree sources

### LINKED ISSUE #4010: BUG: RAMSES/octree rendering issues
### Bug report

**Bug summary**

As reported on slack by Michael Schulreich, `PointSource` objects do not show up in RAMSES volume rendering scenes. After trying various things I've convinced myself it's a bug, but maybe there's a fix I'm not seeing. Additionally, while attempting to debug, I found that using the plane-parallel lens with a RAMSES volume rendering results in a blank image.

**Code for reproduction**

Running the docs example for adding points to a scene  (https://yt-project.org/doc/cookbook/complex_plots.html#volume-rendering-with-points), modified to use a RAMSES sample and to add all white points and increase the point size:

```python
import numpy as np

import yt
from yt.units import kpc
from yt.visualization.volume_rendering.api import PointSource

ds=yt.load_sample("output_00080")  # or "ramses_star_formation"  

sc = yt.create_scene(ds, field=("gas","density"), lens_type="perspective")

# Random particle positions
np.random.seed(1234567)
npoints = 1000
vertices = np.random.random([npoints, 3])
colors = np.ones([npoints, 4])
points = PointSource(vertices, colors=colors, radii=np.full((npoints,), 2))
sc.add_source(points)
sc.save(sigma_clip=2)
```
results in an image with no points 

image

for reference, using `IsolatedGalaxy`  you (correctly) get an image where only the points are visible (because the point opacity is not balanced with the underlying image):

image

**Expected outcome**

An image with points.

**Version Information**
  * yt version: latest dev, also tested latest release

**related issues?**

In trying to debug this, I found that using a plane-parallel lens with or without points results in an empty image:

```python
import yt
ds=yt.load_sample("output_00080") 
sc = yt.create_scene(ds, field=("gas","density"), lens_type="plane-parallel")
sc.show(sigma_clip=2)
```

image

This might point to a common issue with the underlying octree rendering? 

**Finally**, while rendering there are quite a few repeated warnings along the lines of:

```
yt : [WARNING  ] 2022-07-12 10:16:29,125 RAMSESDomainSubset (info_00013): , base_region=YTRegion (info_00013): , center=[3.30568972e+22 3.30568972e+22 3.30568972e+22] cm, left_edge=[0. 0. 0.] cm, right_edge=[6.61137944e+22 6.61137944e+22 6.61137944e+22] cm, domain=RAMSESDomainFile: 1, ds=info_00013.retrieve_ghost_zones was called with the `smoothed` argument set to True. This is not supported, ignoring it.
```
I don't **think** these warnings are related, but it could be...

chrishavlin:
## PR Summary

Changes the octree sampler to initialize image pixel values to the initial image rather than 0 so that any already rendered sources (like points) will show up. 

The following works as expected now:

```python
import numpy as np

import yt
from yt.units import kpc
from yt.visualization.volume_rendering.api import PointSource

ds=yt.load_sample("output_00080")  # output_00080or "ramses_star_formation" or 

sc = yt.create_scene(ds, field=("gas","density"), lens_type="perspective")

# Random particle positions
np.random.seed(1234567)
npoints = 1000
vertices = np.random.random([npoints, 3])
colors = np.ones([npoints, 4])
colors[:,3]=0.01
points = PointSource(vertices, colors=colors, radii=np.full((npoints,), 2))
sc.add_source(points)

sc.show(sigma_clip=2)
```

image

for reference, here is the output of the above code before this change:

image

Closes #4010

### DIFF
--- a/yt/utilities/lib/image_samplers.pyx
+++ b/yt/utilities/lib/image_samplers.pyx
@@ -267,7 +267,7 @@ cdef class ImageSampler:
                     continue
 
                 for i in range(Nch):
-                    idata.rgba[i] = 0
+                    idata.rgba[i] = self.image[vi, vj, i]
                 for i in range(8):
                     vc.mask[i] = 1
 


chrishavlin:
Assuming there aren't wider repercussions to this fix (seems **too** easy), it's probably a good idea to add an image test or two with the octree volume rendering as I don't see any existing tests that actually use the octree sampler (could have missed it though!).

matthewturk (approved):
This looks right!

matthewturk:
Great catch.  I'm not sure if adding tests needs to be in this PR.

PR #4018, opened 2022-07-14
ENH: monkeypatch unyt_array.__format__ protocol

neutrinoceros:
## PR Summary

Workaround a known issue with unyt_array.__format__ in unyt 2.8 by monkeypatching the upstream solution

For context, see
https://github.com/yt-project/unyt/pull/188

The solution is about to be released upstream as part of unyt 2.9, however we can't just require it as a minimal supported version in yt yet because it requires Python 3.8 while yt still supports Python 3.7


chrishavlin (dismissed):
looks good!

One wider question though -- does this mean everywhere within yt should be importing from `yt.units` rather than `unyt` directly? I know we've touched on that question a few times without a clear answer, wondering if this change would affect that discussion...

neutrinoceros:
> One wider question though -- does this mean everywhere within yt should be importing from yt.units rather than unyt directly? 

Nope ! that's the magic of monkey patching ! once *anything* from `yt.units` has been imported, the `unyt.unyt_array` class is modified permanently, and even _retroactively_
For example
```python
import unyt

a = unyt.unyt_array((1, 2, 3), "cm")

# see what happens if this line is uncommented
# from yt.units import display_ytarray 

print(f"here's the formatted array : {a}")
```

neutrinoceros:
eventually @jzuhone should probably sign off for this

jzuhone (approved):

PR #4019, opened 2022-07-14
MNT: cleanup dead branch in doc helper script

neutrinoceros:
## PR Summary

This is necessarilly dead code because it invokes Python 2, which is long gone.


neutrinoceros:
@matthewturk I think only you can merge this

PR #4021, opened 2022-07-14
BUG: fix multiple issues with magnetic field derived fields definitions 

### LINKED ISSUE #4020: BUG: some magnetic field derived fields are unavailable in spherical geometry
### Bug report

**Bug summary**

As an example `"magnetic_field_poloidal"` isn't available. There are no clear reasons why not.

**Code for reproduction**

```python
import numpy as np
import yt
from unyt import gauss

shape = (32, 128, 1)
a = np.sin(np.linspace(-np.pi/2, np.pi/2, 128))
b = np.ones((32, 128))
c = np.zeros((32, 128))

data = {
    ("gas", "magnetic_field_r"): np.reshape(a * b, shape) * gauss,
    ("gas", "magnetic_field_theta"): np.reshape(a * c, shape) * gauss,
}

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.0, 5.0], [np.pi / 2 - 0.5, np.pi / 2 + 0.5], [-0.1, +0.1]]),
    geometry="spherical",
)
ds.r["gas", "magnetic_field_poloidal"]
```

**Actual outcome**

```
YTFieldNotFound: Could not find field ('gas', 'magnetic_field_poloidal') in 
UniformGridData.
```

neutrinoceros:
## PR Summary

Fix #4020

### DIFF
--- a/yt/fields/magnetic_field.py
+++ b/yt/fields/magnetic_field.py
@@ -84,12 +84,13 @@ def setup_magnetic_field_fields(registry, ftype="gas", slice_info=None):
         units=unit_system["pressure"],
     )
 
-    _magnetic_field_toroidal = None
-    _magnetic_field_poloidal = None
+    _magnetic_field_poloidal_magnitude = None
+    _magnetic_field_toroidal_magnitude = None
+    _magnetic_field_spherical_theta = None
 
     if registry.ds.geometry == "cartesian":
 
-        def _magnetic_field_poloidal(field, data):
+        def _magnetic_field_spherical_theta(field, data):
             normal = data.get_field_parameter("normal")
 
             Bfields = ustack(
@@ -105,7 +106,7 @@ def setup_magnetic_field_fields(registry, ftype="gas", slice_info=None):
 
             return get_sph_theta_component(Bfields, theta, phi, normal)
 
-        def _magnetic_field_toroidal(field, data):
+        def _magnetic_field_spherical_phi(field, data):
             normal = data.get_field_parameter("normal")
 
             Bfields = ustack(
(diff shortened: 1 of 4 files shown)


neutrinoceros:
now the definitions should be correct in all curvilinear geometries.
I'll do another pass to check wether it needs finer tuning to support reduced-dimensionality cases.

neutrinoceros:
failures appear to be unrelated. See #4023

neutrinoceros:
@yt-fido test this please

neutrinoceros:
Sigh... now the definitions match what non-cartesian geometry users expect, but are inconsistent with the cartesian defs, and those are actually in usable condition, so I must treat them as user-exposed. I think the right thing to do is to deprecate unsufficently clear field names and reintroduce everything we need with unambiguous names.
I'll revisit this soon.

neutrinoceros:
I think this is now where it should be, so let's undraft

matthewturk:
@jzuhone can you take this one?

jzuhone:
I'll have a look at this today

yt/fields/magnetic_field.py
```diff
@@ -119,59 +120,121 @@ def _magnetic_field_toroidal(field, data):
             phi = data["index", "spherical_phi"]
             return get_sph_phi_component(Bfields, phi, normal)
 
+        def _magnetic_field_poloidal_magnitude(field, data):
```
jzuhone, line 123: Why do we need to put "magnitude" on this field component? We don't do that for any other vector components. Is there a reason we can't give this component a direction? (I see that the way it's computed here doesn't allow you to).

yt/fields/magnetic_field.py
```diff
@@ -119,59 +120,121 @@ def _magnetic_field_toroidal(field, data):
...
-            function=_magnetic_field_poloidal,
+            function=_magnetic_field_toroidal_magnitude,
             units=u,
             validators=[
                 ValidateParameter("normal"),
                 ValidateParameter("bulk_magnetic_field"),
             ],
         )
 
+    if registry.ds.geometry == "cartesian":
```
jzuhone, line 196: I guess I am a little confused as to why in the `"cartesian"` geometry the fields corresponding to the coordinate directions (r, theta, phi) need to be specified explicitly. If we set up `"velocity"` as a vector field in cartesian, we get the cartesian, cylindrical, and spherical coordinate component fields all at once. Does that not happen for the magnetic field for some reason?

yt/fields/magnetic_field.py
neutrinoceros: Well, because it's not a component, it's really the amplitude of the poloidal projection of the total field. That's what myself, and my team, as users, expect to get when querying `B_poloidal`.

jzuhone: Ok sounds good

yt/fields/magnetic_field.py
neutrinoceros: You're right, we already get `"magnetic_field_spherical_phi"` and `"magnetic_field_spherical_theta"` for free. I'll simplify this, thank you very much for pointing it out !

On a general note, most of what I'm doing here could be done in `vector_operations.py` and be usable for both magnetic and velocity fields, and I will very likely make some deduplications in the future. Here I've been trying to get straight to a usable state.

jzuhone (approved):

jzuhone:
LGTM

Issue #4023, opened 2022-07-14
BUG: 29 answer tests failures (changed from unyt 2.9 ?)


neutrinoceros:
### Bug report

**Bug summary**

29 answer tests just started consistently failing a few hours ago, see for instance https://tests.yt-project.org/job/yt_py38_git/5430/

It seems likely that the new version of unyt (2.9.0) is the cause.
I don't know yet if these are actual regressions or if the changes are expected.
for information ping @jzuhone

neutrinoceros:
Solved upstream with https://github.com/yt-project/unyt/pull/249/files

neutrinoceros:
@jzuhone 
It seems like it keeps failing, see https://tests.yt-project.org/job/yt_py38_git/5452/
I'll wait for another job to complete with unyt 2.9.2, but it doesn't look as good as I hoped :/

neutrinoceros:
Nonetheless I can still confirm that unyt 2.9.2 fixes the problematic tests for me (locally)

neutrinoceros:
Phew, it's indeed back to stable 
https://tests.yt-project.org/job/yt_py38_git/5453/
Thank you @jzuhone for your work on the unyt release !

PR #4024, opened 2022-07-15
TMP: pin unyt to 2.8.0

neutrinoceros:
## PR Summary
Going to use this to try and identify the root cause of #4023 (unyt 2.9.0 is the main suspect, though I suspect it's not an actual regression), and potential fixes.
Opening the branch with an empty commit to show that tests fail even on the main branch.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,7 +45,7 @@ install_requires =
     pyparsing>0.0  # hard dependency to MPL. We require it (unconstrained) in case MPL drops it in the future
     tomli-w>=0.4.0
     tqdm>=3.4.0
-    unyt>=2.8.0
+    unyt==2.8.0
     tomli>=1.2.3;python_version < '3.11'
 python_requires = >=3.7,<3.12
 include_package_data = True


neutrinoceros:
It should be clear now that the difference is indeed due to updating unyt 2.8 -> 2.9
Now I'll try to bisect the change locally.

neutrinoceros:
The "regression" bisects to the following commit
https://github.com/yt-project/unyt/commit/9815a485bd3d298b120358bd13438adcaf96ea13

neutrinoceros:
So the changing commit is indeed a bugfix, and I'm pretty sure we should trust the new version over the old one. My conclusion is that the affected answers should be bumped.
If @matthewturk and/or @jzuhone agree with this conclusion, I'll open a new PR to do just that.

jzuhone:
@neutrinoceros can we verify that the reason for the change is that a field is being taken to a zero exponent?

neutrinoceros:
Not for a couple hours at least

neutrinoceros:
actually I was able to establish that the issue does *not* come from an array being raised to power 0 (I hacked the `__pow__` function to raise an error in this case, from a "known good" version of unyt).

After a lot of exploration I think I found a patch... for unyt:
```patch
diff --git a/unyt/array.py b/unyt/array.py
index a96a961..1fbe1c2 100644
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1698,10 +1698,9 @@ class unyt_array(np.ndarray):
         if p == 0.0:
             ret = self.ua
             ret.units = Unit("dimensionless")
+            return ret
         else:
-            ret = self.__array_ufunc__(power, "__call__", self, p)
-
-        return ret
+            return super().__pow__(p)
 
     #
     # Start operation methods
```
It fixes the problem locally, but TBH I don't really know why it makes a difference.

@jzuhone I'll open a PR on unyt if you think we can do a bugfix release soon, and in the mean time we should probably (sadly) forbid unyt 2.9.0 on yt's CI.

neutrinoceros:
I'll try to get Josh involved to see if I'm missing the reason why he implemented this the way he did.

neutrinoceros:
The present PR could actually be merged as a temporary measure to stabilise yt's CI again while we think of all the options, so I'm going to mark it as mergeable.

neutrinoceros:
ok then I guess there's no rush. I'll keep this PR open in case you change your schedule, but if all goes according to plan we'll be able to close it. Thank you !

PR #4025, opened 2022-07-16
ENH: reduce import time by avoiding top-level expensive imports

neutrinoceros:
## PR Summary
Reduce yt's import time by 35-60%

This estimation was performed using other optimizations for core dependencies (cmyt and unyt).
To be clear, I'm comparing yt's import time with and without the present patch, but I'm including cmyt and unyt optimizations in both cases.

Here's the command I've used to measure and target main offenders:

```
python -X importtime -c "import yt" 2> tuna.log && tuna tuna.log
```

### DIFF
--- a/yt/utilities/png_writer.py
+++ b/yt/utilities/png_writer.py
@@ -1,18 +1,10 @@
 from io import BytesIO
 
+from PIL import Image
+
 
 def call_png_write_png(buffer, fileobj, dpi):
-    try:
-        # matplotlib switched from an internal submodule _png to using pillow (PIL)
-        # between v3.1.0 and v3.3.0
-        # So PIL should be available on any system where matplotlib._png doesn't exist
-        import matplotlib._png as _png
-    except ImportError:
-        from PIL import Image
-    try:
-        _png.write_png(buffer, fileobj, dpi)
-    except NameError:
-        Image.fromarray(buffer).save(fileobj, dpi=(dpi, dpi), format="png")
+    Image.fromarray(buffer).save(fileobj, dpi=(dpi, dpi), format="png")
 
 
 def write_png(buffer, filename, dpi=100):
--- a/yt/utilities/png_writer.py
+++ b/yt/utilities/png_writer.py
@@ -1,15 +1,14 @@
 from io import BytesIO
 
-try:
-    # matplotlib switched from an internal submodule _png to using pillow (PIL)
-    # between v3.1.0 and v3.3.0
-    # So PIL should be available on any system where matplotlib._png doesn't exist
-    import matplotlib._png as _png
-except ImportError:
-    from PIL import Image
-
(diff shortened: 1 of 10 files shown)


neutrinoceros:
my optimisations for unyt are public:
https://github.com/yt-project/unyt/pull/250
https://github.com/yt-project/unyt/pull/251

and I already released optimisations for cmyt

matthewturk:
I mostly like this, but it's probably worth changing the name to say we're avoiding top-level *matplotlib* imports, right?

neutrinoceros:
oh yes, absolutely, somehow this survived my commit squash abuse

neutrinoceros:
Actually it's even more general than that because `pkg_resources` has nothing to do with matplotlib, but is _still_ responsible for 5% (!!!) of yt's import time.

matthewturk:
I think we may want to explore a more elegant way of doing the imports, but everything I have come up with involves globals and whatnot, which ... yeah, not great.

neutrinoceros:
There are people thinking about this very seriously at the ecosystem level https://scientific-python.org/specs/spec-0001/

Though it's not clear if/when we'll be able to leverage anything that comes out of it.

matthewturk:
I think that we should move forward, and move back to top level if the ecosystem wide change works out for us. Thank you for doing this.

matthewturk (dismissed):
Looks pretty good to me.

neutrinoceros:
I guess you could merge it now if you want, since failures are unrelated (#4023).

neutrinoceros:
@yt-fido test this please

neutrinoceros:
messed up somewhere, let's switch back to draft for now

neutrinoceros:
There's also an ongoing PEP to support lazy imports support directly in importlib, currently targeted at Python 3.12
I'll keep an eye on this.

yt/utilities/png_writer.py
```diff
@@ -1,19 +1,10 @@
 from io import BytesIO
 
-try:
-    # matplotlib switched from an internal submodule _png to using pillow (PIL)
-    # between v3.1.0 and v3.3.0
-    # So PIL should be available on any system where matplotlib._png doesn't exist
-    import matplotlib._png as _png
-except ImportError:
-    from PIL import Image
+from PIL import Image
```
matthewturk, line 3: Just so we're on the same page -- if I'm reading it right, `PIL` should already be a transitive dependency, right?

matthewturk (approved):

yt/utilities/png_writer.py
neutrinoceros: almost: it is a transitive dependency *if MPL 3.3 or newer is installed*. In the future, as we progressively fade out support to older versions, it will always be a transitive dependency, yes.

PR #4028, opened 2022-07-20
ENH: only validate user-defined field functions (save inspection overhead)

neutrinoceros:
## PR Summary

`inspect.signature` is an expensive function

Using this simple profiling routine
```python
# t.py
import yt
ds.r["gas", "density"]
```
```shell
python -m cProfile -o log.pstats t.py
gprof2dot --colour-nodes-by-selftime log.pstats | dot -Tsvg -o out.svg
```
I was able to measure that ~3.5% of the total runtime is spent validating field functions that are defined internally, so let's restrict its use to user-defined fields to minimise the overhead.

Follow up to #3921

### DIFF
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -1,4 +1,3 @@
-import inspect
 from collections.abc import Callable
 from numbers import Number as numeric_type
 from typing import Optional, Tuple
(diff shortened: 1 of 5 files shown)


neutrinoceros:
actually I got a better idea, I'll repurpose this

yt/fields/field_info_container.py
```diff
@@ -363,31 +362,6 @@ def add_field(
 
         kwargs.setdefault("ds", self.ds)
 
-        if not isinstance(function, Callable):  # type: ignore [arg-type]
-            # type-checking is disabled because of https://github.com/python/mypy/issues/11071
-            # this is compatible with lambdas and functools.partial objects
-            raise TypeError(
-                f"Expected a callable object, got {function} with type {type(function)}"
-            )
```
neutrinoceros, line 371: note that I've dropped this part completely in the process, for two reasons:
- `callable(function)` is much more idiomatic and also faster
- ... and it's also the first thing that's checked when calling `inspect.signature`, so it's redundant

matthewturk (approved):

PR #4029, opened 2022-07-20
ENH: only compute field suggestions if YTFieldNotFound isn't caught

neutrinoceros:
## PR Summary
Another simple performance improvement

The effect of this patch can be measured with
```python
# t.py
import yt
ds.r["gas", "density"]
```

```shell
python -m cProfile t.py | grep levenshtein
```

Without this patch, I see 0.15s (a few percent) is spent on evaluating suggestions for typos I didn't make.
With the patch, it's reduced to 0.

### DIFF
--- a/yt/utilities/exceptions.py
+++ b/yt/utilities/exceptions.py
@@ -1,5 +1,6 @@
 # We don't need to import 'exceptions'
 import os.path
+from typing import List, Tuple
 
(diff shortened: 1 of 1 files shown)


neutrinoceros:
@yt-fido test this please

cphyc (approved):
Looks good to me!

PR #4030, opened 2022-07-21
ENH: cache results from yt.fields.field_functions.get_periodic_rvec

neutrinoceros:
## PR Summary
Another small performance optimization.
This reduces indexing time by about 10% to 25% for the classic IsolatedGalaxy dataset

Measured very non-scientifically with
```python
from time import monotonic_ns
import yt

tstart = monotonic_ns()
ds.index
tstop = monotonic_ns()

print(f"Indexing took {(tstop-tstart) / 1e6:.2e} ms")
```
I still get consistent speedups, from 1.2s (average baseline) to 1.0s with this patch (best run)

### DIFF
--- a/yt/fields/field_functions.py
+++ b/yt/fields/field_functions.py
@@ -1,3 +1,5 @@
+from functools import lru_cache
+
 import numpy as np
 
 from yt.utilities.lib.misc_utilities import obtain_position_vector
@@ -40,6 +42,7 @@ def get_radius(data, field_prefix, ftype):
     return radius
 
 
+@lru_cache(maxsize=128)
 def get_periodic_rvec(data):
     coords = obtain_position_vector(data).d
     if sum(data.ds.periodicity) == 0:


neutrinoceros:
I also get a consistent ~15% gain for the larger `AM06` Athena++ dataset. Seems worth the additional 2 lines.

neutrinoceros:
Doesn't survive CI. I'll try to fix this later, hoping this can be resolved without too much bloat.

neutrinoceros:
So for posterity, the issue with this patch is hashability:
This function's input type isn't too precisely defined, and in practice it works also for `FieldDetector` instances, which are dicts, hence are mutable, hence not hashable. Hashability of input is however a requirement to enable `functools.lru_cache`, so it's not applicable straightforwardly here.
Maybe I'll come back to this problem in the future but right now I have enough on my plate so I'd rather just close this before it grows out of control.

PR #4033, opened 2022-07-25
Add more field options to the Gadget frontend

jzuhone:
## PR Summary

The Gadget frontend has a couple of different ways to detect individual abundance fields of elements for "vanilla" Gadget (i.e., not one of its variants such as GIZMO or AREPO). These depend on the existence of field names such as `"FourMetalFractions"` or `"ElevenMetalMasses"`. Some datasets, however, may have the actual individual names specified, but there is currently no way to check for these. 

This PR adds the functionality to check for the existence of individual abundance fields by searching the SPH field whitelist (it also adds other commonly used elements to the whitelist). 

It also adds aliases for the commonly used fields `"ColdFraction"` and `"HotTemperature"` fields. 

## PR Checklist


yt/frontends/gadget/fields.py
```diff
@@ -122,6 +128,22 @@ def _h_density(field, data):
...
+        species_nums = []
+        for field in self.ds.field_list:
+            if field[0] == ptype:
+                if field[1].endswith("_fraction") or field[1].endswith("_density"):
+                    if field[1] in sph_whitelist_fields:
+                        symbol = field[1].split("_")[0]
+                        species_names.append(symbol)
+                        species_nums.append(periodic_table[symbol].num)
+        if len(species_names) > 0:
+            ret = list(sort_together([species_nums, species_names])[1])
```
neutrinoceros, line 142: Can you please explain why we need to select just the `[1]` element ?

yt/frontends/gadget/fields.py
```diff
@@ -122,6 +128,22 @@ def _h_density(field, data):
 
         return ["H"] + metal_names[:-1]
 
+    def _check_whitelist_species_fields(self, ptype):
+        species_names = []
+        species_nums = []
+        for field in self.ds.field_list:
+            if field[0] == ptype:
+                if field[1].endswith("_fraction") or field[1].endswith("_density"):
```
neutrinoceros, line 136: little known api convenience: `str.endswith` accepts a tuple, which does what you need here
```suggestion
                if field[1].endswith(("_fraction", "_density")):
```

yt/frontends/gadget/fields.py
```diff
@@ -122,6 +128,22 @@ def _h_density(field, data):
...
+                    if field[1] in sph_whitelist_fields:
+                        symbol = field[1].split("_")[0]
```
neutrinoceros, line 138: Ok I see a potential bug here.
For now, because items in `sph_whitelist_fields` that end in `"_fraction"` or `"_density"` all contain exactly 1 underscore, there is no difference between the following
```python
symbol = field[1].split("_")[0]          # your implementation
symbol, _, _ = field[1].partition("_")   # refactor 1
symbol, _, _ = field[1].rpartition("_")  # refactor 2
```
However, to avoid creating bugs in the future, we should anticipate now and make sure this still behaves as expected if we suddenly have a (for instance) `"dark_energy_density"` field in that list, and make the implementation as unambiguous as possible.
If you mean to take "everything before the first underscore", then the first proposed refactor would work. Otherwise, if you mean "everything before the *last* underscore" (and I suspect that's the real intention here), then the second refactor would be the more robust solution.

yt/frontends/gadget/fields.py
jzuhone: @neutrinoceros what we are trying to do here is get a list of elements in the dataset--but we need the list of elements to be sorted by atomic number, so I `sort_together` the two lists of atomic numbers and names but then just discard the former.

jzuhone:
Does anyone know why the `black-jupyter` test is failing? I didn't even touch any notebook files.

yt/frontends/gadget/fields.py
jzuhone: Handled

yt/frontends/gadget/fields.py
jzuhone: handled

jzuhone:
pre-commit.ci autofix

neutrinoceros:
> Does anyone know why the black-jupyter test is failing? I didn't even touch any notebook files.

I think I get the confusion here. `black-jupyter` is a pre-commit hook that runs black all *all* python files, including, but not restricted to notebooks. This name was introduced to be distinct from the first hook, simply called `black`, which only checked `.py` files.

jzuhone:
Ok, thanks. That really needs a less confusing name. In any case, fixed now.

yt/frontends/gadget/fields.py
```diff
@@ -122,6 +128,22 @@ def _h_density(field, data):
...
+                if field[1].endswith(("_fraction", "_density")):
+                    if field[1] in sph_whitelist_fields:
+                        symbol, _, _ = field[1].partition("_")
+                        species_names.append(symbol)
+                        species_nums.append(periodic_table[symbol].num)
+        if len(species_names) > 0:
+            ret = list(sort_together([species_nums, species_names])[1])
+        else:
+            ret = []
+        return ret
```
neutrinoceros, line 145: Now that I get what's done, here's a refactor suggestion. It's flatter, and doesn't require the exotic (but quite nice !) `sort_together` function
```suggestion
        species_names = []
        for field in self.ds.field_list:
            if (
                field[0] == ptype
                and field[1].endswith(("_fraction", "_density"))
                and field[1] in sph_whitelist_fields
            ):
                symbol, _, _ = field[1].partition("_")
                species_names.append(symbol)
        return sorted(specie_names, key=lambda symbol: periodic_table[symbol].num)
```

jzuhone:
pre-commit.ci autofix

neutrinoceros (approved):
LGTM, but I would also like someone more expert than me to sign off here.

matthewturk (approved):

jzuhone:
@chummels @brittonsmith care to have a look at this one?

brittonsmith (approved):
I don't see any problem with this. All good from me.

yt/frontends/sph/fields.py
```diff
@@ -28,6 +28,8 @@ class SPHFieldInfo(FieldInfoContainer):
         ("InitialMass", ("code_mass", [], None)),
         ("TrueMass", ("code_mass", [], None)),
         ("ElevenMetalMasses", ("code_mass", [], None)),
+        ("ColdFraction", ("", ["cold_fraction"], None)),
```
chummels, line 31: What are `ColdFraction` and `HotTemperature` fields?  I've never seen these referenced before.

chummels (commented):
Sorry I'm late to the party.  This looks good to me.  I just didn't realize there would be datasets that had ElectronAbundance set but didn't have the Hydrogen number density set.  I guess this happens or you wouldn't have added it, but it's just surprising that people wouldn't have an H number density in their sim.  Similarly, I don't really know what HotTemperature or ColdFraction fields are.  But the rest of it looks good.

PR #4037, opened 2022-07-27
Refactors oref to equal the number of zones along each dimension

### LINKED ISSUE #4027: ENH: Change block size to be less specified by powers of two
At present, we use the terms `over_refine_factor` and `oref` and `oref_factor` to refer to the "over refinement factor" in our octrees.  Specifically, what this means is, the *size* of an oct leaf node.

For "standard" octrees with 2x2x2 cells in a leaf node, it's set to 1.  The number of zones along each dimension is computed as `1 << oref`.

A long time ago, this was used when we constructed octrees; we still do in fact do this.  But, it's limiting in that it requires that the cell and blocks are completely identical, which is not necessarily the case in block-structured data, while the *remainder* of the way the block-structuring works can repurpose the octree containers.

The solution that I see is to either break the correlation between cells and refinement in how yt manages the structures, or to allow for general specification of cell/leaf/block sizes without using the "powers-of-two" requirement.

BolunThompson:
Refactors over_refine_factor and oref to equal the number of zones along each dimension, instead of nz equaling oref to the power of two.

To maintain backwards compatibility (although this seems to be primarily used internally) `loaders.load_octree` maintains the old behavior.

Closes #4027

### DIFF
--- a/yt/data_objects/index_subobjects/octree_subset.py
+++ b/yt/data_objects/index_subobjects/octree_subset.py
@@ -44,10 +44,10 @@ class OctreeSubset(YTSelectionContainer):
     _block_order = "C"
 
     def __init__(
-        self, base_region, domain, ds, over_refine_factor=1, num_ghost_zones=0
+        self, base_region, domain, ds, over_refine_factor=2, num_ghost_zones=0
     ):
         super().__init__(ds, None)
-        self._num_zones = 1 << (over_refine_factor)
+        self._num_zones = over_refine_factor
         self._num_ghost_zones = num_ghost_zones
         self._oref = over_refine_factor
         self.domain = domain
@@ -443,7 +443,7 @@ class OctreeSubset(YTSelectionContainer):
             [1, 1, 1],
             self.ds.domain_left_edge,
             self.ds.domain_right_edge,
-            over_refine=1,
+            over_refine=2,
         )
         particle_octree.n_ref = nneighbors * 2
         particle_octree.add(morton)
@@ -671,7 +671,7 @@ class YTPositionArray(unyt_array):
         morton = compute_morton(self[:, 0], self[:, 1], self[:, 2], LE, RE)
         return morton
 
-    def to_octree(self, over_refine_factor=1, dims=(1, 1, 1), n_ref=64):
+    def to_octree(self, over_refine_factor=2, dims=(1, 1, 1), n_ref=64):
         mi = self.morton
         mi.sort()
         eps = np.finfo(self.dtype).eps
--- a/yt/geometry/_selection_routines/selector_object.pxi
+++ b/yt/geometry/_selection_routines/selector_object.pxi
@@ -156,7 +156,7 @@ cdef class SelectorObject:
                             visitor.pos[1] = (visitor.pos[1] >> 1)
                             visitor.pos[2] = (visitor.pos[2] >> 1)
                             visitor.level -= 1
-                        elif this_level == 1 and visitor.oref > 0:
+                        elif this_level == 1 and visitor.oref > 1:
                             visitor.global_index += increment
                             increment = 0
                             self.visit_oct_cells(root, ch, spos, sdds,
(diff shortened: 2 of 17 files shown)


yt/geometry/_selection_routines/selector_object.pxi
```diff
@@ -196,7 +196,7 @@ cdef class SelectorObject:
         cdef np.float64_t dds[3]
         cdef np.float64_t pos[3]
         cdef int ci, cj, ck
-        cdef int nr = (1 << (visitor.oref - 1))
+        cdef int nr = (self.oref * 0.5)
```
cphyc, line 199: Are you sure the following code is correct if `oref` if odd, e.g. `self.oref = 3`?

yt/geometry/oct_visitors.pyx
```diff
@@ -373,12 +373,12 @@ cdef class BaseNeighbourVisitor(OctVisitor):
         cdef Oct *neighbour
         cdef bint local_oct
         cdef bint other_oct
-        dx = 1.0 / ((1 << self.oref) << self.level)
+        dx = 1.0 / (self.oref << self.level)
```
cphyc, line 376: Now that I read this piece of code... won't it only work if the size ratio between consecutive level is exactly 2?

matthewturk: Yes -- *but*, this is an assumption that is already made for the code that is being changed.

matthewturk: If we were to fix this, we'd need to do a bit more work on getting the `refinement_factor` code threaded through into Cython.  I am not sure that is appropriate at this point; I believe that we should not expect a strict octree code to ever have something that isn't a factor of 2.  It may be the case that we need to offset this with a multiplicative factor (i.e., if we have one zone refined into four along an axis, we'd want `(2 * self.level)`) but I'm not sure that is necessary at this point.

matthewturk:
@BolunThompson Thank you very much for doing this!

Now that we're actually passing `num_zones`, do you think we could change it to be called something similar?

matthewturk:
I've read the changes and I think they are all pretty good.  I would like to see the attributes be renamed (and a back-compat shim put in place for now to allow specifying `over_refine_factor`, although I have a suspicion that our highest-profile user is @dnarayanan ) and otherwise I think this is good to go.

yt/geometry/_selection_routines/selector_object.pxi
BolunThompson: I suspect it wouldn't work, but before this PR the number of zones (which `oref` is specifying here) was always even, so it won't break anything (and I can't test any changes to make it compatible).

Separately, it should be visitor.oref, not self.oref, so I will change that.

BolunThompson:
pre-commit.ci autofix

matthewturk:
@BolunThompson Let us know when you think this is ready for review?

BolunThompson:
It should be ready for review now.

yt/data_objects/index_subobjects/octree_subset.py
```diff
@@ -43,13 +43,10 @@ class OctreeSubset(YTSelectionContainer):
     _cell_count = -1
     _block_order = "C"
 
-    def __init__(
-        self, base_region, domain, ds, over_refine_factor=1, num_ghost_zones=0
-    ):
+    def __init__(self, base_region, domain, ds, num_zones=2, num_ghost_zones=0):
```
matthewturk, line 46: @BolunThompson Could you add a shim here, so that if `over_refine_factor` is specified it sets `num_zones` appropriately (and emits a deprecation warning)?  We can disable the behavior in 4.2.

matthewturk (commented):
I *think* other than the one thing, this looks good to go.  I still would like @dnarayanan to take a quick look, though.

Xarthisius (approved):
Hi! Thanks for your contribution! It looks good to me!

Issue #4039, opened 2022-07-28
Some geographic mapping issues


chrishavlin:
Starting this issue to document some issues with yt's geographic plotting functionality.

### Test data and environment

Links to the data used below, both are seismic models (internal_geographic in yt jargon) the first is a global model (longitude from -180 to 180, latitude -90 to 90), the second covers only the western US. 
- https://ds.iris.edu/files/products/emc/emc-files/GYPSUM_percent.nc 
- https://ds.iris.edu/files/products/emc/emc-files/wUS-SH-2010_percent.nc 

And here's a function for loading that data in as a internal_geographic uniform grid datasets: https://gist.github.com/chrishavlin/aaf887a5f36eb5f6a7f7ae45f936c2fa

The following used a conda environment with python 3.9.10, cartopy 0.20.3 , geos 3.11.0 , proj 9.0.1 yt 4.1.dev0 on unbuntu 18.04.

### Problem 1: extents

I know there's been some issues here previously (#3705, #3708), but the current state does not do so well... 

The default behavior for a global dataset is good: 

```python
ds = get_internal_IRIS("internal_geographic_global")
c = ds.domain_center
c[0] = 100
p = yt.SlicePlot(ds,'depth','dvs', center=c)
p.set_mpl_projection('Robinson')
p.set_log('dvs',False)
p.set_zlim('dvs',-8, 8)
p.set_cmap('dvs','magma_r')

p._setup_plots()
carto_ax = p.plots['dvs'].axes
carto_ax.add_feature(cfeature.NaturalEarthFeature(
        'cultural', 'admin_0_countries', '10m',
        edgecolor='black', facecolor='none'))
p.show()
```
![Screenshot from 2022-07-28 15-26-40](https://user-images.githubusercontent.com/22038879/181631023-6e495299-7d94-4d10-a508-776e3f975d3e.png)

But any subsets are **VERY** wrong: 

```python
c = ds.domain_center
c[0] = 100
p = yt.SlicePlot(ds,'depth','dvs', center=c, width=[100, 60], origin="native")
p.set_log('dvs',False)
p.set_zlim('dvs',-8, 8)
p.set_cmap('dvs','magma_r')
p._setup_plots()
carto_ax = p.plots['dvs'].axes
carto_ax.add_feature(cfeature.NaturalEarthFeature(
        'cultural', 'admin_0_countries', '10m',
        edgecolor='black', facecolor='none'))
p.show()
```
[...]

PR #4040, opened 2022-07-29
BUG: setting GeoAxes extent

chrishavlin:
This changes how the `extent` for GeoAxes is handled -- this version calls `set_extent` only if the data is not global and always passes the `extent` kwarg to imshow. Seemed to work well locally, but we'll see how CI does...

~~I may broaden the scope of this PR to include changing the default projection to set the `central_longitude` kwarg from the plot arguments when available, but if that proves to be too complex I'll just keep this PR to the `extent` modification.~~ EDIT: Will look into this for a follow up PR.

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -349,7 +349,13 @@ class ImagePlotMPL(PlotMPL):
         # if the axes are cartopy GeoAxes, this checks that the axes extent
         # is properly set.
 
-        if hasattr(self.axes, "set_extent"):
+        if not "cartopy" in sys.modules:
+            # cartopy isn't already loaded, nothing to do here
+            return
+    
+        from cartopy.mpl.geoaxes import GeoAxes
+ 
+        if isinstance(self.axes, GeoAxes):
             # some projections have trouble when passing extents at or near the
             # limits. So we only set_extent when the plot is a subset of the
             # globe, within the tolerance of the transform.
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -276,33 +276,26 @@ class ImagePlotMPL(PlotMPL):
             transform = self.axes.transData
         else:
             transform = self._transform
+
         if hasattr(self.axes, "set_extent"):
-            # CartoPy hangs if we do not set_extent before imshow if we are
-            # displaying a small subset of the globe.  What I believe happens is
-            # that the transform for the points on the outside results in
-            # infinities, and then the scipy.spatial cKDTree hangs trying to
-            # identify nearest points.
-            #
-            # Also, set_extent is defined by cartopy, so not all axes will have
-            # it as a method.
-            #
-            # A potential downside is that other images may change, but I believe
-            # the result of imshow is to set_extent *regardless*.  This just
-            # changes the order in which it happens.
-            #
-            # NOTE: This is currently commented out because it breaks in some
-            # instances.  It is left as a historical note because we will
-            # eventually need some form of it.
-            # self.axes.set_extent(extent)
+            # some projections have trouble when passing extents at or near the
+            # limits. So we only set_extent when the plot is a subset of the
+            # globe, within the tolerance of the transform.
 
-            # possibly related issue (operation order dependency)
-            # https://github.com/SciTools/cartopy/issues/1468
+            # add validation to make sure transform IS a cartopy crs? pretty sure
+            # it should always be if we end up here....
 
-            # in cartopy 0.19 (or 0.20), some intented behaviour changes produced an
-            # incompatibility here where default values for extent would lead to a crash.
-            # A solution is to let the transform object set the image extents internally
-            # see https://github.com/SciTools/cartopy/issues/1955
-            extent = None
+            # note that `set_extent` here is setting the extent of the axes.
+            # still need to pass the extent arg to imshow below in order to
+            # ensure that it is properly scaled. also note that set_extent
+            # expects values in the coordinates of the transform: it will
+            # calculate the coordinates in the projection.
+            global_extent = transform.x_limits + transform.y_limits
+            thresh = transform.threshold
+            if all([extent[ie] < (global_extent[ie] - thresh) for ie in range(4)]):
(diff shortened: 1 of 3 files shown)


yt/visualization/base_plot_types.py
```diff
@@ -276,33 +276,26 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
...
+            # still need to pass the extent arg to imshow below in order to
+            # ensure that it is properly scaled. also note that set_extent
+            # expects values in the coordinates of the transform: it will
+            # calculate the coordinates in the projection.
+            global_extent = transform.x_limits + transform.y_limits
+            thresh = transform.threshold
+            if all([extent[ie] < (global_extent[ie] - thresh) for ie in range(4)]):
+                self.axes.set_extent(extent, crs=transform)
+            else:
+                self.axes.set_global()
```
chrishavlin, line 298: this is probably not needed as the plot will default to global. also, I'll probably move this validation to its own method...

yt/visualization/base_plot_types.py
```diff
@@ -276,33 +276,26 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
...
+            # it should always be if we end up here....
+
+            # note that `set_extent` here is setting the extent of the axes.
+            # still need to pass the extent arg to imshow below in order to
+            # ensure that it is properly scaled. also note that set_extent
+            # expects values in the coordinates of the transform: it will
+            # calculate the coordinates in the projection.
+            global_extent = transform.x_limits + transform.y_limits
+            thresh = transform.threshold
+            if all([extent[ie] < (global_extent[ie] - thresh) for ie in range(4)]):
```
neutrinoceros, line 295: Quick optimisation tip here: `all` accepts a generator, so exhaustive consumption of the iterable via conversion into a list  isn't needed
```suggestion
            if all(extent[ie] < (global_extent[ie] - thresh) for ie in range(4)):
```

yt/visualization/base_plot_types.py
matthewturk: :+1:

chrishavlin:
So in the latest batch of commit, among other things, I adjusted the `test_geo_projections` tests to use a linear instead of log scale to avoid the matplotlib issue, which should separately be fixed by #3849 and #3986 (at which point we could switch the projection tests back to using log, but I don't see a reason to not leave them with the linear scale).

yt/visualization/tests/test_geo_projections.py
```diff
@@ -151,6 +161,21 @@ def test_nondefault_transform(self):
                 self.slc.plots[("stream", "Density")].axes.projection, proj_type
             )
 
+    @requires_module("cartopy")
+    def test_extent(self):
+        # checks that the axis extent is narrowed when doing a subselection
+        axis = "altitude"
+        slc = yt.SlicePlot(self.ds, axis, ("stream", "Density"), origin="native")
```
neutrinoceros, line 168: just a thought here, as I see `origin="native"` being used for every single slice plot in this test module. Does it make sense to use any other value ? if not, then it's something I missed in https://github.com/yt-project/yt/pull/3863, and could easily be corrected (not necessarily in this PR though)

neutrinoceros (approved):
Two quick comments, otherwise LGTM

yt/visualization/base_plot_types.py
```diff
@@ -370,6 +345,27 @@ def _init_image(self, data, cbnorm, cblinthresh, cmap, extent, aspect):
             self.cb = self.figure.colorbar(self.image, self.cax)
         self.cax.tick_params(which="both", axis="y", direction="in")
 
+    def _validate_axes_extent(self, extent, transform):
+        # if the axes are cartopy GeoAxes, this checks that the axes extent
+        # is properly set.
+
+        if hasattr(self.axes, "set_extent"):
```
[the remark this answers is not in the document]
chrishavlin: yup, that would be clearer. Do you have a preference for doing the check like you suggested vs. using `on_demand_imports`? e.g., using

```python
from yt.utilities.on_demand_imports import _cartopy as cartopy  
cartopy.__is_available__ 
```
and I could add the `GeoAxes` to the `on_demand_imports` as well. 

I'm totally happy to commit your suggestion as is, just curious!

yt/visualization/tests/test_geo_projections.py
chrishavlin: Don't think you missed anything -- maybe at one point we didn't have any of the validation to set `origin="native"`? I haven't noticed any problems when omitting the `origin` kwarg. I did remember this morning that I wanted to add a test for a dataset that is not global (I only added a test for plotting a subset of a global dataset). Maybe I'll leave out the kwarg from that one...

yt/visualization/base_plot_types.py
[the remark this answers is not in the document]
neutrinoceros: using `__is_available__` would be a different condition. `"cartopy" in sys.modules` is intentionally more restrictive: even if cartopy is installed, `axes` cannot be a `GeoAxes` instance if cartopy wasn't already imported earlier.
As to importing from on_demand_imports here, I don't think it makes a difference, but I also don't have a true preference, I just went for a self-contained suggestion.

chrishavlin:
oops, just pushed to this branch accidentally... ignore the latest changes, will fix....

chrishavlin:
ok. fixed the accidental push, updated the geoaxes check. and I want to add one more test then this will be good I think!

neutrinoceros:
I forgot to check that `sys` was already imported in my last suggestion. It wasn't.

neutrinoceros:
I fixed the 2 issues (one purely stylistic, reported by flake8) with my latest suggestion.
I'll review again after lunch when CI is complete

neutrinoceros (approved):
I think it's good to go. Before I merge, does this completely fix #4039 (which isn't currently linked) ?

chrishavlin:
Yes, this does fix the bug there. The other part of the discussion in #4039 (setting the `central_longitude` of the default projection when possible) is more of an enhancement that will take some extra work, I'll open up a new issue to track it.

PR #4041, opened 2022-07-30
Preliminary support for tensor fields

jzuhone:
## PR Summary

yt has support for vector fields--but of course in some domains tensor fields are also useful. This is especially true in applications involving special relativity. 

This PR takes a first stab at adding tensor fields to yt by adding the [stress-energy tensor for a perfect fluid](https://en.wikipedia.org/wiki/Perfect_fluid), which is used by GAMER (implemented here) and Athena++ (to be implemented later). 

So for example, accessing the "T^00" component of a tensor field works like this: `dd["T00"]`, and the "T^10" would work like this: `dd["T10"]`, and so on. Have to think about how to handle raising/lowering indices (if at all) and if one wants to support other coordinates than just cartesian (harder, but doable I guess).

I expect there may be some conversation about the field naming convention. Docs forthcoming.

## PR Checklist

### DIFF
--- a/yt/fields/astro_fields.py
+++ b/yt/fields/astro_fields.py
@@ -145,7 +145,7 @@ def setup_astro_fields(registry, ftype="gas", slice_info=None):
 
         return _four_velocity
 
-    for u in "xyz":
+    for u in registry.ds.coordinates.axis_order:
         registry.add_field(
             ("gas", f"four_velocity_{u}"),
             sampling_type="local",
--- a/yt/fields/tensor_fields.py
+++ b/yt/fields/tensor_fields.py
@@ -1,6 +1,5 @@
 from functools import partial
 
-ax = "txyz"
 
 # This is the metric for Minkowski spacetime in SR
 def metric(mu: int, nu: int):
@@ -14,6 +13,8 @@ def metric(mu: int, nu: int):
 
 
 def setup_stress_energy_ideal(registry, ftype="gas"):
+
+    ax = "t" + "".join(registry.ds.coordinates.axis_order)
 
     pc = registry.ds.units.physical_constants
     inv_c2 = 1.0 / (pc.clight * pc.clight)
--- a/yt/fields/tensor_fields.py
+++ b/yt/fields/tensor_fields.py
@@ -2,8 +2,9 @@ from functools import partial
 
 ax = "txyz"
 
-
+# This is the metric for Minkowski spacetime in SR
 def metric(mu: int, nu: int):
+    # This assumes the -+++ signature
     if (mu, nu) == (0, 0):
         return -1
     elif mu == nu:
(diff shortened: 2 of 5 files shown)


yt/frontends/gamer/tests/test_outputs.py
```diff
@@ -97,3 +102,32 @@ def test_jiw():
...
+        for nu in range(4):
+            # matrix is symmetric so only do the upper-right part
+            if nu > mu:
+                Umu = sp[f"four_velocity_{axes[mu]}"]
+                Unu = sp[f"four_velocity_{axes[nu]}"]
+                Tmunu = h * Umu * Unu
+                if mu != nu:
+                    assert_array_almost_equal(sp[f"T{mu}{nu}"], sp[f"T{nu}{mu}"])
+                else:
+                    Tmunu += p
```
neutrinoceros, line 132: this is unreachable given `nu > mu` is checked on this branch.

yt/fields/tensor_fields.py
```diff
@@ -0,0 +1,36 @@
+from functools import partial
+
+ax = "txyz"
```
neutrinoceros, line 3: could this be moved closer to usage ?

jzuhone: Sorry, could you elaborate? Not sure what you mean

jzuhone: Ah, see what you mean. It's there because I anticipated filling this file with other stuff eventually. But now I have a different idea.

yt/frontends/gamer/tests/test_outputs.py
jzuhone: Thanks!

yt/fields/astro_fields.py
```diff
@@ -145,7 +145,7 @@ def _four_velocity(field, data):
 
         return _four_velocity
 
-    for u in "xyz":
+    for u in registry.ds.coordinates.axis_order:
```
matthewturk, line 148: great catch

matthewturk (approved):
this looks pretty good to me.  I don't know that I personally have strnog naming convention feelings other than "easily tokenizable"

yt/fields/astro_fields.py
neutrinoceros: +1

neutrinoceros: although... I'd like to be sure what to expect. Right now the `T_mu_nu` fields you're adding are only compatible with cartesian geometries, right ? If so, this change still has value because it's one less obstacle to overcome when/if we want to extend support to other geometries in the future, but it doesn't do it just by itself, right ?

jzuhone: This change would add the `four_velocity` fields to non-Cartesian datasets now, but not the tensor, no

neutrinoceros: then I guess I'm alright with it

neutrinoceros (approved):

neutrinoceros:
@jzuhone since you marked this as "preliminary", I'd rather ask: do you want this PR mentioned in the changelog for yt 4.1 ? (I'm not including every single PR since there are a lot of non-user-facing ones)

jzuhone:
I'll follow up in a separate PR with docs, and maybe support for Athena++ in the same PR

neutrinoceros:
ok, thank you !

PR #2711, opened 2020-07-01
BLD: Split EWAH bool array code

themousepotato:
## PR Summary

This PR removes the code of EWAH bool array from the yt repository. As suggested by @matthewturk, it is being split out as a new python package.

## PR Checklist

- [ ] Code passes flake8 checker
- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #4043, opened 2022-07-31
BLD: test against Cython 3 pre-releases

neutrinoceros:
## PR Summary

Experimental ground for using Cython 3.0 (currently in alpha) to compile yt

### DIFF
--- a/yt/utilities/lib/ewah_bool_array.pxd
+++ b/yt/utilities/lib/ewah_bool_array.pxd
@@ -91,10 +91,7 @@ cdef extern from "boolarray.h" namespace "ewah":
 cimport cython
 cimport numpy as np
 
-IF UNAME_SYSNAME == "Windows":
-    ctypedef uint32_t ewah_word_type
-ELSE:
-    ctypedef np.uint32_t ewah_word_type
+ctypedef uint32_t ewah_word_type
 ctypedef EWAHBoolArray[ewah_word_type] ewah_bool_array
 ctypedef EWAHBoolArraySetBitForwardIterator[ewah_word_type] ewah_bool_iterator
 ctypedef vector[size_t] bitset_array
--- a/yt/utilities/lib/geometry_utils.pxd
+++ b/yt/utilities/lib/geometry_utils.pxd
@@ -11,12 +11,21 @@ cimport numpy as np
 from libc.float cimport DBL_MANT_DIG
 from libc.math cimport frexp, ldexp, sqrt
 
-DEF ORDER_MAX=20
-DEF INDEX_MAX_64=2097151
+
+cdef extern from *:
+    """
+    #define ORDER_MAX     20
+    #define INDEX_MAX_64  2097151
+    #define XSHIFT 2
+    #define YSHIFT 1
+    #define ZSHIFT 0
+    """
+    int ORDER_MAX
+    int INDEX_MAX_64
+    int XSHIFT
+    int YSHIFT
+    int ZSHIFT
 # TODO: Handle error for indices past max
-DEF XSHIFT=2
-DEF YSHIFT=1
-DEF ZSHIFT=0
 
 @cython.cdivision(True)
 @cython.boundscheck(False)
--- a/yt/utilities/lib/image_samplers.pxd
+++ b/yt/utilities/lib/image_samplers.pxd
@@ -15,7 +15,6 @@ cimport numpy as np
 from .partitioned_grid cimport PartitionedGrid
 from .volume_container cimport VolumeContainer
 
-DEF Nch = 4
 
 # NOTE: We don't want to import the field_interpolator_tables here, as it
 # breaks a bunch of C++ interop.  Maybe some day it won't.  So, we just forward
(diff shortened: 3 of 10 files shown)


neutrinoceros:
See upstream for the report of the current blocker https://github.com/cython/cython/issues/4940

neutrinoceros:
Turns out the only issue was actually with PyYAML, so I added a small workaround and now this should be ready (still going to wait until tests pass to updraft)

neutrinoceros:
actually I'm going to try and fix #4044 while I'm at it, so back to draft

yt/utilities/lib/geometry_utils.pxd
```diff
@@ -11,12 +11,21 @@ cimport numpy as np
 from libc.float cimport DBL_MANT_DIG
 from libc.math cimport frexp, ldexp, sqrt
 
-DEF ORDER_MAX=20
-DEF INDEX_MAX_64=2097151
+
```
matthewturk, line 14: Thanks for fixing this -- but to be honest I don't see this example of removing a DEF as a net improvement for readability.

yt/utilities/lib/image_samplers.pxd
```diff
@@ -31,7 +30,7 @@ ctypedef void generate_vector_info_function(ImageSampler im,
             np.float64_t v_dir[3], np.float64_t v_pos[3]) nogil
 
 cdef struct ImageAccumulator:
-    np.float64_t rgba[Nch]
+    np.float64_t rgba[4]  # Nch = 4
```
matthewturk, line 33: I mean, if we can't use `Nch` here then why have `Nch` above?  If we change `Nch` in the earlier code, this would then break, right?

matthewturk (dismissed):
Thanks for doing this, even if I'm annoyed we have to.

neutrinoceros:
Yeah. I'm still going to hold of for now, as I would like to finalize fixing #4044

yt/utilities/lib/image_samplers.pxd
neutrinoceros: above where ? that's the only line that's using it in this file, and yeah unfortunately the refactor pattern that I used everywhere else doesn't work here

matthewturk: Ah, I meant above this in the file view, which included stuff like octree raytracing, where it is defined with your trick.

neutrinoceros: Looks like it's actually only used here and in `image_samplers.pyx`, I missed this. Let's clean this up then.

neutrinoceros:
for the sake of experimentation, I pushed a revert commit for bd1fb35fe29555979aec6fff1701b8792df0f827; I hope that our tests are able to catch the issue it was fixing (assuming it still exists)

neutrinoceros:
so, either the issue doesn't exist anymore, or our test do not catch it.
FYI @jzuhone

neutrinoceros:
opening for review but as it stands, it absolutely requires @jzuhone 's review.

neutrinoceros:
I'll switch to draft because I don't think we need or want to think about this before yt 4.1 is out. I'll reopen after the release is out the door.

neutrinoceros:
Going to close/reopen to refresh CI
Also I don't think this should be a blocker for 4.2, and it's currently not a priority either, so I'll make the milestone one PR lighter

yt/utilities/lib/ewah_bool_array.pxd
```diff
@@ -91,10 +91,7 @@ cdef extern from "boolarray.h" namespace "ewah":
 cimport cython
 cimport numpy as np
 
-IF UNAME_SYSNAME == "Windows":
-    ctypedef uint32_t ewah_word_type
-ELSE:
-    ctypedef np.uint32_t ewah_word_type
+ctypedef uint32_t ewah_word_type
```
neutrinoceros, line 94: I think this should be reverted to avoid conflicts with #2711

yt/utilities/lib/geometry_utils.pxd
neutrinoceros: Looks like there may be a better solution in the next pre-release, see https://github.com/cython/cython/pull/5242

neutrinoceros:
Switching to draft for now. Will wait until https://github.com/cython/cython/pull/5242 is either merged/released or otherwise closed.

neutrinoceros:
Let's refresh CI to see how we're doing with the first beta.

neutrinoceros:
build fails with Python 3.0b1. We get one error
```
'cython.floating' is not a valid cython.* module
```
and a warning
```
The keyword 'nogil' should appear at the end of the function signature line. Placing it before 'except' or 'noexcept' will be disallowed in a future version of Cython.
```

neutrinoceros:
Everything uncontroversial from this branch as been cut into smaller PRs. The only thing still unique here is the (attempted) removal of the `DEF` macro. Since it is not clear at the moment how best to replace it, I'll close this PR.
See #4355 for an overview of the migration path.

PR #4048, opened 2022-08-02
MNT: cleanup an unused and hidden class attribute

neutrinoceros:
## PR Summary
This boolean attribute is not user exposed and _never_ set to any other than `True`.
It probably had value as a debugging switch but it's not doing anything now and it's also very simple to reimplement if the need arises in a debugging session, so I suggest we just remove it.


matthewturk:
No, we use this in debugging.  I think it is worth keeping as well so that we can manually override it in some very specific cases, which will be relevant after we get some of the callable stream stuff going.

neutrinoceros:
alright, fine by me

PR #4049, opened 2022-08-02
add parameters to StreamHandler and loaders

### LINKED ISSUE #4047: REQ: Supply parameters through stream handlers
An optional parameter to the stream loaders should include the parameters to update the metadata dictionary.

chrishavlin:
This adds the `parameters` dictionary to the `StreamHandler` so that it can be supplied as a keyword argument to all the stream loaders so that there's a convenient way to pass in metadata that is useful to have around. 

closes #4047

### DIFF
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -4,7 +4,7 @@ import uuid
 import weakref
 from itertools import chain, product, repeat
 from numbers import Number as numeric_type
-from typing import Optional, Type
+from typing import Type
 
 import numpy as np
 from more_itertools import always_iterable
@@ -88,7 +88,8 @@ class StreamHandler:
         io=None,
         particle_types=None,
         periodicity=(True, True, True),
-        parameters: Optional[dict] = None,
+        *,
+        parameters=None,
     ):
         if particle_types is None:
             particle_types = {}
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -4,7 +4,7 @@ import uuid
 import weakref
 from itertools import chain, product, repeat
 from numbers import Number as numeric_type
-from typing import Type
+from typing import Optional, Type
 
 import numpy as np
 from more_itertools import always_iterable
@@ -88,6 +88,7 @@ class StreamHandler:
         io=None,
         particle_types=None,
         periodicity=(True, True, True),
+        parameters: Optional[dict] = None,
     ):
         if particle_types is None:
             particle_types = {}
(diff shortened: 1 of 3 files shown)


chrishavlin:
Could use a test or two...

matthewturk (approved):
This looks pretty good to me.

chrishavlin:
So those type checking failures are coming from doing things like the following in the `loaders`:

```python
    handler = StreamHandler(
        grid_left_edges,
        grid_right_edges,
        grid_dimensions,
        grid_levels,
        -np.ones(nprocs, dtype="int64"),
        np.zeros(nprocs, dtype="int64").reshape(nprocs, 1),  # Temporary
        np.zeros(nprocs).reshape((nprocs, 1)),
        sfh,
        field_units,
        (length_unit, mass_unit, time_unit, velocity_unit, magnetic_unit),
        particle_types=particle_types,
        periodicity=periodicity,
        parameters=parameters,
    )

    handler.name = "ParticleData"
    handler.domain_left_edge = domain_left_edge
    handler.domain_right_edge = domain_right_edge
    handler.refine_by = 2
    handler.dimensionality = 3
    handler.domain_dimensions = domain_dimensions
    handler.simulation_time = sim_time
    handler.cosmology_simulation = 0
```

all those attributes (`name`, `domain_left_edge`, etc) aren't defined in `StreamHandler`. It looks like we do this everywhere in the loaders... so I could add these all as optional kwargs in `StreamHandler`? Or define all these as properties so the syntax in the loaders doesn't have to change?  I'll think more about this, but I think there should be a simple fix.

neutrinoceros:
re: type checking
Our current mypy config is that only functions that have _some_ type hints are actually checked. It's helpful to support a smooth transition from no type hints, but it also means that every time you add the first hint for a given function, every existing problems in that function will surface.

Now our frontend code is notoriously hard to type-check because there are many violations of the Liskov substitution principle (which mypy tries to enforce).
Here the cost of adding a single hint is very likely to snowball into fixing a _lot_ of these, and it would most probably outweigh the patch you're actually trying to make.

All that for saying: I wouldn't bother.

neutrinoceros:
That said, if you're up for it and it turns out there *is* a simple fix, by all means, go ahead :)

chrishavlin:
Well, turns out there is a somewhat simple fix for the failures related to the `StreamHandler`, but I'd have to add a bunch of ignores to the ones related to all the `Stream` dataset subclasses... So I'm inclined to just remove the type hint since otherwise this PR is very simple.

neutrinoceros:
Fine by me. I only got one comment overall: can we please make these new arguments keyword-only ?

chrishavlin:
yes, definitely can make it keyword only! I did in one spot but forgot to in all the loader functions... will do that!

neutrinoceros (approved):

Issue #3916, opened 2022-05-09
BUG: Projection of periodic boxes with specified width smaller than boxsize.  


benopp99:
I've seemingly found an issue with periodic box wrapping in ProjectionPlot.  Below is a 25 Mpc/h IllustrisTNG periodic volume, and I kept getting maps zoomed into halos cut off at the periodic boundary.  This was seemingly confounding because when I run a whole box, the wrapper works perfectly.  After a lot of tinkering, I think it has to do with the width argument being not equal to the domain.  I made images of 23000, 24000, 25000, and 26000 kpc/h centered at (22000, 22000, 22000), which I think serves as a good diagnostic of the problem.  A loss gap opens up at the periodic boundary (i.e. 25000) and grows as the width becomes smaller (23000 is upper left, 24000 upper right, and so on).  Therefore for a ProjectionPlot zoomed on a specific halo near the periodic boundary, you often incorrectly miss the halo.  Here's the code snippet (I used FITSProjection which is a wrapper for ProjectionPlot, which has same issue):

```python
    my_slice = ds.box(left_edge, right_edge) 

    Pos = [22000,22000,22000]
    Width = 23000

    proj_fits_slice = yt.FITSProjection(ds, "z", ("gas","O_p6_number_density"), data_source=my_slice, image_res=npix,center=Pos,width=Width)
```

Note this is in all types of periodic simulations- EAGLE, Simba too.  Maybe I'm doing something wrong, and if there is a quick fix, let me know.

I've added a ProjectionPlot of oxygen (instead of O^6+) to more clearly show the 23000 case.

**Version Information**
python 3.9.5
yt 4.1.dev0

Note that this issue has been seen in earlier version of yt going back a couple years.  i.e. before yt 4.0

![gaps_1](https://user-images.githubusercontent.com/50880489/167432823-f2798cb5-24e1-44e7-93c1-2c404c311a11.png)
![gaps_2](https://user-images.githubusercontent.com/50880489/167432814-826eb239-9631-44db-b9af-bc6dce80e9ef.png)

neutrinoceros:
Triaging as "viz: 2D" because this bug manifests itself in a ProjectionPlot, though it seems likely that this is really a particle selection bug.

neutrinoceros:
@benopp99 Could you please share more of the script you're using to produce this buggy image ? Also, can you confirm that you're dealing with a particle field ?

benopp99:
[...]

PR #4050, opened 2022-08-04
BUG: Fix intersections between SPH particles and regions when the boundary is periodic

jzuhone:
## PR Summary

This PR fixes the issue raised by #3916 and seen in other contexts where sometimes particles which properly belong to a rectangular region straddling a periodic boundary would get missed. 

I think something was not quite right in the overall logic, but I have to be honest I am not certain. What I do know for sure is that instead of checking the particle position itself for particles that lay outside the boundary of a region, we should be checking the "left_edge" and "right_edge" of the particle in question (defined by the +/- the radius away from the position), and then computing the radius of the particle to the edge. That was inspired by considering the region/cell intersection implementation: 

https://github.com/yt-project/yt/blob/6c0780f0e00a6ef6ed4f40510d4570d04f7cf59a/yt/geometry/_selection_routines/region_selector.pxi#L81-L92 

This PR implements that logic, and makes the code logic clearer. It also uses the already defined `periodic_difference` in the superclass to compute the distance between the particle position and the region edge. However, I noticed during this exercise that we have this same logic implemented in two different places (both of which are used):

Here: 

https://github.com/yt-project/yt/blob/6c0780f0e00a6ef6ed4f40510d4570d04f7cf59a/yt/geometry/_selection_routines/selector_object.pxi#L266-L278

and here:

https://github.com/yt-project/yt/blob/6c0780f0e00a6ef6ed4f40510d4570d04f7cf59a/yt/geometry/selection_routines.pxd#L81-L89

Should I presume the latter would be faster? 

Also, it's possible this change may break some answer tests.

## PR Checklist

### DIFF
--- a/yt/data_objects/tests/test_sph_data_objects.py
+++ b/yt/data_objects/tests/test_sph_data_objects.py
@@ -82,6 +82,21 @@ def test_region():
                 assert_equal(reg["gas", "density"].shape[0], answer)
 
 
+def test_periodic_region():
+    ds = fake_sph_grid_ds(10.0)
+    coords = [0.7, 1.4, 2.8]
+
+    for x in coords:
+        for y in coords:
+            for z in coords:
+                center = np.array([x, y, z])
+                for n, w in zip((8, 27), (1.0, 2.0)):
+                    le = center - 0.5 * w
+                    re = center + 0.5 * w
+                    box = ds.box(le, re)
+                    assert box["io", "particle_ones"].size == n
+
+
 SPHERE_ANSWERS = {
     ((0, 0, 0), 4): 7,
     ((0, 0, 0), 3): 7,
--- a/yt/geometry/_selection_routines/region_selector.pxi
+++ b/yt/geometry/_selection_routines/region_selector.pxi
@@ -147,9 +147,10 @@ cdef class RegionSelector(SelectorObject):
         cdef np.float64_t dmin = 0
         cdef np.float64_t d = 0
         for i in range(3):
-            if self.right_edge_shift[i] <= pos[i] < self.left_edge[i]:
+            if (pos[i]+radius < self.left_edge[i] and \
+                pos[i]-radius >= self.right_edge_shift[i]):
                 d = self.periodic_difference(pos[i], self.left_edge[i], i)
-            elif pos[i] > self.right_edge[i]:
+            elif pos[i]-radius > self.right_edge[i]:
                 d = self.periodic_difference(pos[i], self.right_edge[i], i)
             dmin += d*d
         return int(dmin <= r2)
--- a/yt/geometry/_selection_routines/region_selector.pxi
+++ b/yt/geometry/_selection_routines/region_selector.pxi
@@ -145,15 +145,13 @@ cdef class RegionSelector(SelectorObject):
         cdef np.float64_t p
         cdef np.float64_t r2 = radius**2
         cdef np.float64_t dmin = 0
+        cdef np.float64_t d = 0
         for i in range(3):
-            if self.periodicity[i] and self.check_period[i]:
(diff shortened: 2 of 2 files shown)


matthewturk:
THANK YOU, John.

My guess is that the latter probably is faster, but there may be cases where we can't call it, or where type conversion may end up being more expensive than we'd expect.

jzuhone:
By the way, this is my modified script from @neutrinoceros:

```python
import numpy as np
import yt


def fake_ds(hsml_factor):
    npart = 3**3
    x = np.empty(npart)
    y = np.empty(npart)
    z = np.empty(npart)

    tot = 0
    for i in range(0, 3):
        for j in range(0, 3):
            for k in range(0, 3):
                x[tot] = i + 0.5
                y[tot] = j + 0.5
                z[tot] = k + 0.5
                tot += 1

    data = {
        "particle_position_x": (x, "cm"),
        "particle_position_y": (y, "cm"),
        "particle_position_z": (z, "cm"),
        "particle_mass": (np.ones(npart), "g"),
        "particle_velocity_x": (np.zeros(npart), "cm/s"),
        "particle_velocity_y": (np.zeros(npart), "cm/s"),
        "particle_velocity_z": (np.zeros(npart), "cm/s"),
        "smoothing_length": (0.05 * np.ones(npart) * hsml_factor, "cm"),
        "density": (np.ones(npart), "g/cm**3"),
        "temperature": (np.ones(npart), "K"),
    }

    bbox = np.array([[0, 3], [0, 3], [0, 3]])

    return yt.load_particles(data=data, length_unit=1.0, bbox=bbox)


ds = fake_ds(10)

for w in (3, 2, 1.5, 1.0, 0.5):
    center = np.array([2.8]*3)
    proj = yt.ProjectionPlot(
        ds, "z", ("gas", "density"), center=center, width=w,
    )
    proj.save(f"/tmp/test_{w=}.png")

    le = center-0.5*w
    re = center+0.5*w
    box = ds.box(le, re)
    print(box["io","smoothing_length"].size)
    proj = yt.ProjectionPlot(
        ds, "z", ("gas", "density"), center=center, data_source=box)
    proj.save(f"/tmp/test_box_{w=}.png")
```

which adds more widths and it also adds a test where we keep the image fixed at the domain width but make the region itself smaller. As expected, the number of particles plotted decreases.

And the results:
test_box_w=0 5
test_box_w=1 0
test_box_w=1 5
test_box_w=2
test_box_w=3
test_w=0 5
test_w=1 0
test_w=1 5
test_w=2
test_w=3

neutrinoceros:
Very nice, thanks for solving this. I will not be able to review this properly until tomorrow, but if Matt wants to push the button, I don't want to make you guys wait for me when it looks "obviously right".

jzuhone:
Ping @benopp let's test on that dataset of yours

matthewturk (dismissed):
I think this looks right.

benopp99:
I'm going to have to refigure out how to access CAMELS data now that I
don't have my account.

-Ben

benopp99:
Okay, this link now should work:

https://wormhole.app/MjqOK#npxkeFwTqrXr9MTxT3yTFg

jzuhone:
@benopp99 I can confirm this code fixes your plots.

benopp99:
Outstanding!  Glad I could find a bug (an important one) stress testing YT
on simulations.  This will be quite useful.  Let me know when the bug fix
is in an updated version.

Best,
Ben

jzuhone:
@benopp99 you can always check out this branch until then :) but it will be merged into `main` no later than tomorrow I assume.

jzuhone:
Ok, I added a test for this. Provided it passes, this is ready to go.

neutrinoceros (approved):
LGTM, nice craft !

Issue #3967, opened 2022-06-09
DEPR: DerivedField.output_units (?)


neutrinoceros:
### Bug report

**Bug summary**

In #3962, it was brought up that the `output_units` attribute of the `DerivedField` class doesn't have a clear meaning and has apparently caused a lot of confusion (and bugs) in the past.
@brittonsmith and I think it may be best to just remove this attribute (and replace it with `DerivedField.units` where it's used internally). We could probably hide it being a (deprecated) property to maintain backward compatibility if anyone is relying on it downstream.

This is a non-critical issue for now but I may give it a go soon, so any insight about why this attribute exists is welcome.

brittonsmith:
I'm strongly in favor of getting rid of `output_units` straight away. At worst, I would be happy with generating a warning if someone tries to use it, but I don't see any reason we need to allow a grace period of use with this functionality. There are already existing methods to change displayed units in visualizations.

neutrinoceros:
It's just a safety principle. I don't think anyone is using it downstream, but you never know. If we can afford a deprecation cycle without creating a heavy maintenance load for ourselves it seems preferable.

neutrinoceros:
This is where this option is currently documented
https://yt-project.org/doc/developing/creating_derived_fields.html#field-options

> output_units
(Advanced) For fields that exist on disk, which we may want to convert to other fields or that get aliased to themselves, we can specify a different desired output unit than the unit found on disk.

neutrinoceros:
After experimenting with "just removing it" in #4061, I'm not convinced this is an easy or even feasible task, as it's deeply baked into particle analysis code.
What's frustrating here is that I still don't really get the motivation for these designs. I'm not saying they are bad, but it's harder to work with or around them when you don't understand them. Maybe we just need better documentation for them.

Issue #3876, opened 2022-03-31
Custom units added to the unit registry do not appear as attributes of `yt.units`


gregsalvesen-work:
Custom units added to the global unit registry using `yt.units.define_unit()` do not get placed in `yt.units`:

```python
import yt
RgSun = 1 * yt.units.G * yt.units.Msun / yt.units.c**2  # Gravitational radius of the Sun
yt.units.define_unit("RgSun", RgSun)
print(yt.units.RgSun)
>> AttributeError: module 'yt.units' has no attribute 'RgSun'
```

I expected my newly defined unit `RgSun` to be accessible via `yt.units.RgSun`. At least this is how the standalone `unyt` package behaves:
```python
import unyt as u
RgSun = 1 * u.G * u.Msun / u.c**2
u.define_unit("RgSun", RgSun)
print(u.RgSun)
>> RgSun
```

I also observe this expected behavior when using `define_unit()` for a data set in `yt`:
```python
import yt
ds = yt.load_uniform_grid({}, [1,1,1])
RgSun = 1 * yt.units.G * yt.units.Msun / yt.units.c**2
ds.define_unit("RgSun", RgSun)
print(ds.units.RgSun)
>> RgSun
```

**Version Information**
  * Operating System: mac OS Big Sur
  * Python Version: 3.9.10
  * yt version: 4.0.2
  * unyt version: 2.8.0

matthewturk:
@gregsalvesen-work Thanks for this.  @jzuhone , do you know how to address unit registry issues?  This seems potentially up your alley.

neutrinoceros:
Hi, thanks for reporting.
My understanding is that yt.units is meant to be a backward compatibility layer from when unyt wasn’t isolated as a stand-alone package. With this in mind, I don’t think we should have to mirror modern unyt behavior. That said, if this used to be how yt.units worked, then I agree it should be fixed.

chrishavlin:
Was just poking around this since I was recently looking at that section of code in yt... and the current behavior is certainly confusing. turns out `yt.units.define_unit()` will actually add to `unyt`'s default registry: 

```python
import yt
import unyt as u
RgSun = 1 * yt.units.G * yt.units.Msun / yt.units.c**2  # Gravitational radius of the Sun
yt.units.define_unit("RgSun", RgSun)
print(u.RgSun)
>> RgSun
```

[...]

PR #4058, opened 2022-08-06
RFC: change DerivedField.units attr type (str -> unyt.Unit)

neutrinoceros:
## PR Summary

Rationale: it's easier to convert a Unit objet to string than vice versa.

Expected gain: overall simplifications.

If I find a reason why this change isn't feasible along the way, at least I'll have learned why, and I may even find some simplifications that can still be done.

### DIFF
--- a/yt/data_objects/selection_objects/data_selection_objects.py
+++ b/yt/data_objects/selection_objects/data_selection_objects.py
@@ -273,7 +273,7 @@ class YTSelectionContainer(YTDataContainer, ParallelAnalysisInterface, abc.ABC):
                         # dimensionless YTArray and verify that field is
                         # supposed to be unitless
                         fd = self.ds.arr(fd, "")
-                        if fi.units != Unit():
+                        if Unit(fi.units) != Unit():
                             raise YTFieldUnitError(fi, fd.units) from None
                     except UnitConversionError as e:
                         raise YTFieldUnitError(fi, fd.units) from e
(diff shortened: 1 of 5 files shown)


neutrinoceros:
So an important difference is that by *not* delaying conversion from str to Unit, we loose the ability to register field globally with units that don't exist in the common registry, e.g.

```python
import yt

@yt.derived_field(name="xxx", sampling_type="cell", units="code_length")
def xxx(field, data):
    return data["index", "x"]
```
However we should probably require that global fields use only units that are globally registered.
This would imply that in my example above, users should register `"code_length"` at a global level first.
We know however that units registration at the global level doesn't work at all (#3876), so probably we should fix that first.

neutrinoceros:
Many tests are still broken because a couple builtin fields are defined using `"code_*"` units, so it falls in the same problem I pointed to above. Most notably particle fields. I don't know why it was implemented that way, but it doesn't look trivial to factor that out, so I expect It could be hard to even experiment on a refactor.
Overall #3876 and #3967 appear to be related, and much higher priorities. I'll keep this as a draft for now (makes it easier to explore failing tests if I want to come back to it) but I may close it soon if I can't think of a way forward.

matthewturk:
I think we shouldn't pursue this at this time.

neutrinoceros:
agreed. I'll focus on related issues first. I may resurrect this in the future.

PR #4060, opened 2022-08-07
ENH: separate yt.units.define_unit from unyt.define_unit

### LINKED ISSUE #3876: Custom units added to the unit registry do not appear as attributes of `yt.units`

neutrinoceros:
## PR Summary
This should fix #3876
I'll add a test later, for now I'd like to see if any existing test fails because of the change.

### DIFF
--- a/doc/source/analyzing/units.rst
+++ b/doc/source/analyzing/units.rst
@@ -335,3 +335,23 @@ second example returns data in low-redshift comoving megaparsecs.
 Wherever possible it's best to do calculations in physical units when working
 with more than one dataset. If you need to use comoving units or code units then
 extra care must be taken in your code to avoid ambiguity.
+
+
+Defining custom units globally
+------------------------------
+
+Units can be added to yt at runtime using ``yt.units.define_unit``.
+Such units can then be imported from the ``yt.units`` namespace.
+
+For example
+
+.. python-script::
+
+   from yt.units import mm, s
+
+   val = 5*mm/s
+   yt.units.define_unit("snail_speed", val)
+
+   from yt.units import snail_speed
+
+   snail_array = [1, 2, 3] * snail_speed
--- a/yt/units/unit_object.py
+++ b/yt/units/unit_object.py
@@ -1,1 +1,34 @@
-from unyt.unit_object import *
+# re-export all of unyt.unit_object's public interface, except for `define_unit`
+from unyt.unit_object import sympy_one, Unit, em_conversions, em_conversion_dims, NULL_UNIT
+
+
+def define_unit(
+    symbol, value, tex_repr=None, offset=None, prefixable=False, registry=None
+):
+    """
+    yt.units.define_unit is a thin wrapper around unyt.define_unit
+    The key difference is that yt.units.define_unit will register new units to
+    yt's default unit registry instead of unyt's
+    Units defined with this function are then available to import from the yt.units
+    namespace.
+
+    See unyt.define_unit for more details.
+    """
+    from unyt import define_unit as unyt_define_unit
+    import yt.units
+
+    if registry is None:
+        registry = yt.units.default_unit_registry
+
+    unyt_define_unit(
+        symbol,
+        value,
+        tex_repr=tex_repr,
+        offset=offset,
+        prefixable=prefixable,
+        registry=registry,
+    )
+
+    if registry is yt.units.default_unit_registry:
+        u = Unit(symbol, registry=registry)
+        setattr(yt.units, symbol, u)
--- a/yt/__init__.py
+++ b/yt/__init__.py
@@ -68,7 +68,7 @@ from yt.units import (
     uunion1d,
     uvstack,
 )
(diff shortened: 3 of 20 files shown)


neutrinoceros:
No failures ! I added a `python-script` to documentation, which will serve as a test (and avoid having to deal with side effects in a test module).

yt/units/unit_object.py
```diff
@@ -1 +1,34 @@
-from unyt.unit_object import *
+# re-export all of unyt.unit_object's public interface, except for `define_unit`
+from unyt.unit_object import sympy_one, Unit, em_conversions, em_conversion_dims, NULL_UNIT
```
neutrinoceros, line 2: Open question to reviewers:
This is the 100% backward compatible refactor, but do we actually want to re-export everything public from this external module ? I think we just really need `Unit`, and the rest can be considered implantation details ?

doc/source/analyzing/units.rst
```diff
@@ -335,3 +335,23 @@ second example returns data in low-redshift comoving megaparsecs.
 Wherever possible it's best to do calculations in physical units when working
 with more than one dataset. If you need to use comoving units or code units then
 extra care must be taken in your code to avoid ambiguity.
+
+
+Defining custom units globally
+------------------------------
+
+Units can be added to yt at runtime using ``yt.units.define_unit``.
+Such units can then be imported from the ``yt.units`` namespace.
```
chrishavlin, line 344: but they won't be available in a `ds`..... which is maybe actually how it should be (see main comment), but is probably worth mentioning here.

chrishavlin (approved):
Thanks for working on this! Looks good to me with the caveat below that is maybe worth some discussion?

I initially had a different expectation of what the scope of this PR was -- I was expecting that adding units to the default registry would then make them available to datasets. i.e., I thought I'd be able to do


```python
import yt
import numpy as np

val = 5*yt.units.mm/yt.units.s
yt.units.define_unit("snail_speed", val)
yt.units.snail_speed

shp = (10, 10, 10)
data = {'speedy_snails': (100000000. * np.random.random(shp), "snail_speed")}
ds = yt.load_uniform_grid(data, shp)
ds.units.snail_speed # attribute error
```

but this doesn't work. because `Dataset._create_unit_registry` does not actually use the `default_unit_registry` (because it needs to allow for using a specified `unit_system`)... so I actually still need to define this unit at the dataset level as well. And I'm left wondering why there's a default registry? what is it the default for? 

But yeah, I think just having the basic functionality actually working (the **actual** scope of this PR) is worth while.

neutrinoceros:
Good call, I actually did not go as far as checking that...
I think the existing default unit registry is merely a limb from the yt/unyt separation. I'll look into it some more and try to make it _useful_.

yt/units/unit_object.py
neutrinoceros: I'm not sure it's technically possible to deprecate importing specific objects.
Since this PR is also not going in the direction of deprecating `yt.units` as a public namespace, I think we could try re-exposing only relevant subsets of unyt (here it would be just `Unit`) at the top level (in `yt.units`), and adapt/discuss if any complaints ensue. I will probably propose something like this early in the 4.2 dev cycle

neutrinoceros:
> but this doesn't work. because Dataset._create_unit_registry does not actually use the default_unit_registry (because it needs to allow for using a specified unit_system)... so I actually still need to define this unit at the dataset level as well. And I'm left wondering why there's a default registry? what is it the default for?

I digged a little further and came to the conclusion that we _cannot_ in general connect the default unit registry with the ones that are attached to datasets, for exactly the reasons Chris exposed already.
So now I don't really think that I took the correct approach. Now I believe `yt.define_unit` should be deprecated altogether. There's no urgency in removing it, but the deprecation message should make it clear that `unyt.define_unit` and `ds.define_unit` are the replacement depending on the intended use case.

PR #4061, opened 2022-08-07
EXP: retire output_units attr

neutrinoceros:
## PR Summary
A simple experiment around #3967 to see what tests break if we suddendly retire output_units


neutrinoceros:
Clearly not as trivial as I had hoped. `output_units` seems to be deeply rooted in particle plots and loading, where a lot of operations involve converting to and from "code units". Right now I don't understand why it was designed this way so I can't guarantee there's a way out of it.

neutrinoceros:
I think this can easily get out of control and I don't think I want to invest the time needed to "fix" or decipher this. I'll close it for now.

PR #4063, opened 2022-08-08
Adds `data_source` argument to covering grid

### LINKED ISSUE #4052: ENH: Allow YTCoveringGrid to accept a data source
Presently, we set up a data source for covering grids in `_setup_data_source`.  We should be able to provide this, thus enabling spheres to be regularized for instance.

The output would still be a box, but this would be a way to produce regularized data from within specific regions.

Alternately, if we are OK with over-generating the data, we could simply apply a `selector` mask to the covering grid.

chrishavlin:
This adds a `data_source` keyword (only) argument to the `YTCoveringGrid` and then intersects  that `data_source` with the region used for the covering grid in order to create a regularized sampling of, e.g., a sphere. 

Here's an example:

```python
import yt
import numpy as np

cg = ds.covering_grid(2, left_edge=[0.0, 0.0, 0.0], dims=[128, 128, 128])
gridded_dens = cg[("gas", "density")]

sp = ds.sphere(ds.domain_center, 0.2)
cg_sp_intx = ds.covering_grid(2, left_edge=[0.0, 0.0, 0.0], dims=[128, 128, 128], data_source=sp)
gridded_dens_in_sphere = cg_sp_intx[("gas", "density")]

f, axs = plt.subplots(nrows=1, ncols=2)
axs[0].imshow(np.log10(gridded_dens[:,:,64]))
axs[1].imshow(np.log10(gridded_dens_in_sphere[:,:,64]))
```

image

Pushing a  nominally working draft now... but when ready this would close #4052

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -666,7 +666,9 @@ class YTCoveringGrid(YTSelectionContainer3D):
             center = None
         else:
             center = field_parameters.get("center", None)
-        YTSelectionContainer3D.__init__(self, center, ds, field_parameters)
+        YTSelectionContainer3D.__init__(
+            self, center, ds, field_parameters, data_source=data_source
+        )
 
         self.level = level
         self.left_edge = self._sanitize_edge(left_edge)
@@ -688,7 +690,6 @@ class YTCoveringGrid(YTSelectionContainer3D):
             )
             + self.ds.domain_offset
         )
-        self._base_data_source = data_source
         self._setup_data_source()
         self.get_data(fields)
 
@@ -821,12 +822,14 @@ class YTCoveringGrid(YTSelectionContainer3D):
         return tuple(self.ActiveDimensions.tolist())
 
     def _setup_data_source(self):
-        self._data_source = self.ds.region(self.center, self.left_edge, self.right_edge)
-        if self._base_data_source is not None:
-            # intersect the extra with the region
-            self._data_source = self.ds.intersection(
-                [self._base_data_source, self._data_source]
-            )
+
+        reg = self.ds.region(self.center, self.left_edge, self.right_edge)
+        if self._data_source is None:
+            # alwasy True for YTArbitraryGrid (does not accept a data_source)
+            self._data_source = reg
+        else:
+            self._data_source = self.ds.intersection([self._data_source, reg])
+
         self._data_source.min_level = 0
         self._data_source.max_level = self.level
         # This triggers "special" behavior in the RegionSelector to ensure we
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -656,6 +656,8 @@ class YTCoveringGrid(YTSelectionContainer3D):
         num_ghost_zones=0,
         use_pbar=True,
         field_parameters=None,
+        *,
+        data_source=None,
     ):
         if field_parameters is None:
             center = None
@@ -683,6 +685,7 @@ class YTCoveringGrid(YTSelectionContainer3D):
             )
             + self.ds.domain_offset
         )
+        self._extra_data_source = data_source  # copy instead of reference???
         self._setup_data_source()
         self.get_data(fields)
 
@@ -816,6 +819,11 @@ class YTCoveringGrid(YTSelectionContainer3D):
 
     def _setup_data_source(self):
         self._data_source = self.ds.region(self.center, self.left_edge, self.right_edge)
+        if self._extra_data_source is not None:
+            # intersect the extra with the region
+            self._data_source = self.ds.intersection(
+                [self._extra_data_source, self._data_source]
+            )
         self._data_source.min_level = 0
         self._data_source.max_level = self.level
         # This triggers "special" behavior in the RegionSelector to ensure we
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -627,6 +627,9 @@ class YTCoveringGrid(YTSelectionContainer3D):
         A list of fields that you'd like pre-generated for your object
     num_ghost_zones : integer, optional
         The number of padding ghost zones used when accessing fields.
+    data_source :
+        An existing data_source to intersect with the covering grid. Grid points
+        outside the data_source will exist as empty values.
 
     Examples
     --------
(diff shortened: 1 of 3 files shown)


yt/data_objects/construction_data_containers.py
```diff
@@ -683,6 +685,7 @@ def __init__(
             )
             + self.ds.domain_offset
         )
+        self._extra_data_source = data_source  # copy instead of reference???
```
matthewturk, line 688: reference is OK here -- if we change anything, I'd suggest changing the field parameters, but even that's not necessary.  I think I made a stink about this in-person but that was misguided.

Can I namesmith it to `_base_data_source` or something?

yt/data_objects/construction_data_containers.py
```diff
@@ -816,6 +819,11 @@ def shape(self):
 
     def _setup_data_source(self):
         self._data_source = self.ds.region(self.center, self.left_edge, self.right_edge)
+        if self._extra_data_source is not None:
```
matthewturk, line 822: OK, I hadn't thought about it, but great catch.

matthewturk (approved):

yt/data_objects/construction_data_containers.py
chrishavlin: ya! name change is good

chrishavlin:
note on current state: no new tests, docstring would need updating, a new example in the documentation would be good.

yt/data_objects/construction_data_containers.py
matthewturk: The real problem is with the internal usage of `data_source`, not anything you did -- but I don't want this to get too far out of scope.  I think what you have is good.

chrishavlin: name changed to `_base_data_source`

yt/data_objects/construction_data_containers.py
```diff
@@ -816,6 +822,11 @@ def shape(self):
 
     def _setup_data_source(self):
         self._data_source = self.ds.region(self.center, self.left_edge, self.right_edge)
+        if self._base_data_source is not None:
+            # intersect the extra with the region
+            self._data_source = self.ds.intersection(
+                [self._base_data_source, self._data_source]
+            )
```
chrishavlin, line 829: So the current test failure is because YTArbitraryGrid does not call `super().__init__`. I'm hesitant to add `_base_data_source` to their common parent (YTSelectionContainer3D) because I'm worried it will get confusing...  

So I could either store the `data_source` kwarg  in `_data_source` and rewrite this bit to be:

```python
        reg = self.ds.region(self.center, self.left_edge, self.right_edge)
        if self._data_source is not None:
            # intersect the extra with the region
            self._data_source = self.ds.intersection([self._data_source, reg])
        else:
            self._data_source = reg
```

OR I could add a `_base_data_source` to YTArbitraryGrid (and set it to `None` and not add it as a kwarg there) 

OR I could have a type or attribute check here.

 I'm inclined to go with the first option -- @matthewturk any feeling here?

matthewturk: First option is definitely my preference.  This change should currently be restricted.

chrishavlin: >  This change should currently be restricted. 

Ok, just pushed up the first option. Not sure what you meant here though -- was it just that it should only work for the covering (and not arbitrary) grid?

matthewturk: I think it could work for AG but I want this PR not to expand too an uncontrollable size.  :)

yt/data_objects/construction_data_containers.py
chrishavlin: out of date now that it's being stored in _data_source... will resolve.

yt/data_objects/construction_data_containers.py
chrishavlin: first option is in

yt/data_objects/construction_data_containers.py
chrishavlin: out of date as well -- current logic is similar but checks `self._data_source` instead

chrishavlin:
edit: added tests and updated docs. Going to switch to ready-to-review. I won't be around for a little while but feel free to review!

neutrinoceros:
refreshing CI. I will review this over the weekend

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

yt/data_objects/construction_data_containers.py
```diff
@@ -815,7 +822,14 @@ def shape(self):
         return tuple(self.ActiveDimensions.tolist())
 
     def _setup_data_source(self):
-        self._data_source = self.ds.region(self.center, self.left_edge, self.right_edge)
+
+        reg = self.ds.region(self.center, self.left_edge, self.right_edge)
+        if self._data_source is None:
+            # always True for YTArbitraryGrid (does not accept a data_source)
```
neutrinoceros, line 828: I worry this comment might turn into confusing technical debt if `YTArbitraryGrid` ever receives the missing update.
Maybe add a link to this PR to the comment to say that it holds *as of* this PR, so future readers have an incentive to *check* wether it's still true ?

yt/data_objects/tests/test_covering_grid.py
```diff
@@ -82,6 +82,46 @@ def test_covering_grid():
                     ]
                     assert_equal(f, g[("gas", "density")])
 
+            # test the data_source kwarg (new with PR 4063)
```
neutrinoceros, line 85: this test is already so dense, I would prefer having a separate test, albeit at the cost of loading twice as much data (which I think we can afford in this case)

neutrinoceros (commented):
Nothing major, but I do have a couple requests. Overall, I really like this.

yt/data_objects/construction_data_containers.py
chrishavlin: good suggestion, will do.

yt/data_objects/tests/test_covering_grid.py
chrishavlin: sure! I went back and forth on it, easy enough to add these tests as a standalone test. will do that.

yt/data_objects/construction_data_containers.py
chrishavlin: updated!

yt/data_objects/tests/test_covering_grid.py
chrishavlin: separate test now

neutrinoceros (approved):
I see @matthewturk previously approved this but I'll hold off merging until he signs of again.

matthewturk (approved):

PR #3820, opened 2022-02-20
BUG: fix plot bounds for slices in cylindrical coordinates

neutrinoceros:
## PR Summary
Reutilizing some bits from #3533, fix #3819
Opening as a draft because there's a secondary problem to solve. Namely, axes ticks are incorrect (0, 0) should always be the origin, not the display center

```python
import numpy as np
import yt
from yt.testing import fake_amr_ds, add_noise_fields

shape = (32, 64, 128)
ds = yt.load_uniform_grid(
    {},
    shape,
    bbox=np.array([[1, 5], [0.1, np.pi / 2 - 0.1], [0, 1]]),
    geometry="polar",
)
add_noise_fields(ds)
p = yt.SlicePlot(ds, "z", "noise1")
p.save("/tmp/")
```
UniformGridData_Slice_z_noise1

PR #4068, opened 2022-08-09
EXP: allow a rotation matrix for cutting planes

matthewturk:
This experiments with allowing the individual to specify a cutting plane through a rotation matrix, rather than through a set of normal and north vectors.

Why do this, you ask?  Well!  It turns out that we currently store this information, but we don't actually allow it to be specified; you can specify normal and north, but not east, and there's a fair bit of normalization that goes on.  I'd like to allow a full set of affine transformations to be applied, which this is a step toward.

One potential use case of this is with NIFTI data, where we have a provided affine transformation that changes from reference to subject space.  It's a bit awkward to set this up, however, unless we allow for a proper rotation matrix to be provided.

This pull request may not pass tests, and may instead become a place where we allow specification of `Orientation` objects or the like instead of the set of vectors currently allowed.  But, just as it is, we get the ability to do something like this:

```
import yt
import numpy as np

ds = yt.load("data/brains/sub-01_ses-01_run-1_T1map.nii")
rot_mat = np.array([ds.parameters[f'srow_{ax}'] for ax in 'zyx'], dtype='f8').T[:3,:3]
cp = ds.cutting(normal = None, center=ds.domain_center, rotation_matrix = rot_mat)
cp.to_pw("intensity")
```

and have it return to us a nice image in the subject space -- i.e., the pre-transformed "actual" "head."  That being said, at present we do not allow non-unitary transformations, so it looks like it has been grotesquely cut off at the margins, but this could be a first step toward correcting that.  Properly applied, we could also allow this to bleed over into particle plots, which would be awfully nice.

### DIFF
--- a/yt/data_objects/selection_objects/slices.py
+++ b/yt/data_objects/selection_objects/slices.py
@@ -204,25 +204,31 @@ class YTCuttingPlane(YTSelectionContainer2D):
         ds=None,
         field_parameters=None,
         data_source=None,
+        rotation_matrix=None,
     ):
-        validate_3d_array(normal)
-        validate_center(center)
-        if north_vector is not None:
-            validate_3d_array(north_vector)
-        validate_object(ds, Dataset)
-        validate_object(field_parameters, dict)
-        validate_object(data_source, YTSelectionContainer)
         YTSelectionContainer2D.__init__(self, 4, ds, field_parameters, data_source)
+        validate_center(center)
         self._set_center(center)
         self.set_field_parameter("center", center)
-        # Let's set up our plane equation
-        # ax + by + cz + d = 0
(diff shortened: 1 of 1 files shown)


matthewturk:
I think this is making progress -- I will reply with some specific ideas on the coordinate handler ASAP.

neutrinoceros:
Interesting ! I think the example plot you show suffers from the same defect we had in polar projections (polar, cylindrical and spherical geometries) where the limits of the figure are incorrectly computed. As a reminder, here's where I fixed similar problems: https://github.com/yt-project/yt/pull/3820

matthewturk:
It might -- *but*, I think it might also be because I supplied a matrix where the norms of individual rotation vectors were not 1, and there may be collapsing sets of assumptions about the scaling.  I will check though!

PR #4066, opened 2022-08-09
Update to grid_patch.py to avoid calling units

yipihey:
This is to speed up calculations of the grid coordinates x, y, z which also benefits radius calculations and other geometric operations in derived fields.

## PR Checklist

- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #4071, opened 2022-08-11
ENH: perf improvements by skipping unnecessary symbolic unit computations

yipihey:
## PR Summary

Profiled script
```python
# t_prof.py

import yt
import numpy as np
from yt import derived_field
import unyt as u

for fn in range(5):
    ds = yt.load_sample("HiresIsolatedGalaxy")

    ds._periodicity = (False, False, False)
    v, c = ds.find_max(("gas", "density"))
    sp = ds.sphere(c, 10 * u.kpc)
    profiles = yt.create_profile(
        sp,
        "radius",
        [
            ("gas", "density"),
            ("gas", "radial_velocity"),
            ("gas", "tangential_velocity"),
            ("gas", "sound_speed"),
            ("gas", "mach_number"),
            ("gas", "temperature"),
        ],
        weight_field=("gas", "cell_mass"),
        units={"radius": "pc", "density": "amu/cm**3"},
        logs={"radius": True},
    )
    profiles.save_as_dataset("/tmp/test.h5")
```

Profiling and viz
```
python -m cProfile -o test.profile t_prof.py && snakeviz test.profile
```

on main:
Screenshot 2022-08-13 at 19 08 45

This branch:
Screenshot 2022-08-13 at 19 07 49

## PR Checklist

_edited by @neutrinoceros with details_

### DIFF
--- a/yt/data_objects/index_subobjects/grid_patch.py
+++ b/yt/data_objects/index_subobjects/grid_patch.py
@@ -338,8 +338,8 @@ class AMRGridPatch(YTSelectionContainer):
             return np.empty((0, 3), dtype="float64")
         coords = convert_mask_to_indices(mask, self._last_count).astype("float64")
         coords += 0.5
-        coords *= self.dds[None, :]
-        coords += self.LeftEdge[None, :]
+        coords *= self.dds.d[None, :]
+        coords += self.LeftEdge.d[None, :]
         return coords
 
     def select_fwidth(self, dobj):
--- a/yt/data_objects/index_subobjects/grid_patch.py
+++ b/yt/data_objects/index_subobjects/grid_patch.py
@@ -348,7 +348,7 @@ class AMRGridPatch(YTSelectionContainer):
             return np.empty((0, 3), dtype="float64")
         coords = np.empty((count, 3), dtype="float64")
         for axis in range(3):
-            coords[:, axis] = self.dds[axis]
+            coords[:, axis] = self.dds.d[axis]
         return coords
 
     def select_ires(self, dobj):
--- a/yt/geometry/geometry_handler.py
+++ b/yt/geometry/geometry_handler.py
@@ -327,7 +327,7 @@ class YTDataChunk:
             c = obj.select_fcoords(self.dobj)
             if c.shape[0] == 0:
                 continue
-            ci[ind : ind + c.shape[0], :] = c
+            ci.d[ind : ind + c.shape[0], :] = c 
             ind += c.shape[0]
         return ci
 
@@ -363,7 +363,8 @@ class YTDataChunk:
             c = obj.select_fwidth(self.dobj)
             if c.shape[0] == 0:
                 continue
-            ci[ind : ind + c.shape[0], :] = c
+            ci.d[ind : ind + c.shape[0], :] = c  
+
             ind += c.shape[0]
         return ci
 
@@ -419,7 +420,7 @@ class YTDataChunk:
             c = obj.select_fcoords_vertex(self.dobj)
             if c.shape[0] == 0:
                 continue
-            ci[ind : ind + c.shape[0], :, :] = c
+            ci.d[ind : ind + c.shape[0], :, :] = c
             ind += c.shape[0]
         return ci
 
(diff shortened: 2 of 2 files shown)


neutrinoceros:
Hi @yipihey,
it looks like this branch includes the patch from #4066 and is a logical continuation of it, so I'd advise to keep this PR and close the first one if that's okay with you.

Also, I have a good idea what's going on because we discussed it on Slack, but it'd be good to add some context here so it's clearer to others. In particular, having your benchmark script and show what observable you measured (and how) would be very useful.

Thanks a lot for working on this with us !

yipihey:
Ah yes. Indeed. This adds one more place with to add a .d to avoid triggering unit calculations when the outcome is known. 
I'm just learning how to keep different pull requests separate from different adds and commits and much appreciate your patience!

matthewturk (approved):
This looks to me like it could be an enormous improvement in performance -- with big impacts on generating ghost zones, as well!  Thanks, @yipihey .  I think if we can guarantee that these will be in the same units (which I think we can) it should be good to go.

yipihey:
Super @matthewturk! 
Yes I imagine as a rule we should always be able to do all computation in code_units. It should be safe to assume the users chose those units as the ones best suited for any calculation on their data fields. Then things like .in_base and methods like it should just show up when plotting labels or coloring etc. to done with the preferred “output” units. 
T

neutrinoceros:
Are you done or do you intend to push again here ?
In any case I will try it out this weekend with the benchmark you posted on slack !

jzuhone:
Hi @yipihey,

> It should be safe to assume the users chose those units as the ones best suited for any calculation on their data fields. 

I would be a bit wary of assuming this in all cases--in this particular I think it's justified, however.

yipihey:
> Hi @yipihey,
> 
> > It should be safe to assume the users chose those units as the ones best suited for any calculation on their data fields. 
> 
> I would be a bit wary of assuming this in all cases--in this particular I think it's justified, however.

Yes. Definitely want to be careful. Perhaps it makes sense to think of it as a style guide? I.e. recommend that code units are used when doing internal calculations. Then over time we can see whether we find a case in which this would not be the best or at least equivalent choice.

neutrinoceros:
@yipihey I completed the original message, reran your test script, and uploaded some basic screenshots from snakeviz.
From the runs I did on my machine I see no actual perf gain so I guess I am missing something. Can you check (and correct !) what I did ? Maybe I'm not up to date on the script you ran, or I am not looking at the right level in the profile result ?

yipihey:
> @yipihey I completed the original message, reran your test script, and uploaded some basic screenshots from snakeviz.
> From the runs I did on my machine I see no actual perf gain so I guess I am missing something. Maybe I'm not up to date on the script you ran, or I am not looking at the right level in the profile result ? Can you check (and correct !) what I did ?

Oh fantastic. Thanks for checking. 
Yes what you tested was the original script I started with and is pointing us into places to fix up. 
This first tiny pull request should only affect creating of fcoords, widths,  and volumes but is dominated by very large memory allocations from other routines. 

I think in the script below we should notice fewer calls to array.py in the snakeviz outputs. 


```python
# Aprof.py: Use yt to make 1-D profiles and save them to a file.
import yt

for i in range(5):
    ds = yt.load_sample("HiresIsolatedGalaxy")

    ds._periodicity = (False, False, False) # Checking periodicity is slow in yt
    readit = ds.all_data()  # already defines xyz

    x = readit.fcoords[:,0]
    y = readit.fcoords[:,1]
    z = readit.fcoords[:,2]

    x2   = readit["index", "x"]
    y2   = readit["index", "y"]
    z2   = readit["index", "z"]
    vol  = readit["index", "volume"]
```

neutrinoceros:
yes with that script I see about 4% gain. It's more than 1%/changed line ! If it meets your goal here, I guess we can just merge as is.

yipihey:
Perfect. Yes. That is all for now. 
Thanks,
Tom

Typed on an IPhone. Please excuse typos and surprising autocorrects.

yipihey:
I found two more spots that are in exactly the same spirit as the previous changes. I think we should keep that in this PR.

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):

yipihey:
Yes let’s call it done for now. 
I think there are more similar tweaks that will be possible but it may touch more files and will take more time. 
Thanks! 
Tom

PR #4073, opened 2022-08-11
DEPR: deprecate yt.define_units

neutrinoceros:
## PR Summary
Since its actual behaviour doesn't meet expectations (#3876), and it appears unfeasible to make this function relevant (#4060), I propose we simply deprecate it.

### DIFF
--- a/yt/__init__.py
+++ b/yt/__init__.py
@@ -8,6 +8,8 @@ yt is a toolkit for analyzing and visualizing volumetric data.
 
 """
 from ._version import __version__, version_info  # isort: skip
+from unyt import define_unit
+
 import yt.units as units
 import yt.utilities.physical_constants as physical_constants
 from yt.data_objects.api import (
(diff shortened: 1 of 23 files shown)


jzuhone:
What this is supposed to do (and does do) is add units to the default unit registry. It was never designed to add units to `yt.units` (having units importable from this namespace is something which came later). So I think we shouldn't rush to judgment on deprecation. 

It is regrettable that it's not easy to make this also add units to the `yt.units` namespace, but I don't see that as a sufficient reason to deprecate it.

jzuhone:
I guess what I'm saying is that this merits a deeper look about how to make it better instead of rushing to judgment about uprooting it.

neutrinoceros:
I'm open to better options. I just can't find any.

> What this is supposed to do (and does do) is add units to the default unit registry

.. too `unyt`'s default unit registry, which is never used in yt (AFAICT). I can't see how that's desirable.

neutrinoceros:
Note that I triaged this for 4.1 but it's not a blocker; we could very well postpone making any decision to resolve #3876

neutrinoceros:
switching to draft for now as this turns out to be unstable after rebasing

neutrinoceros:
closing, as I don't expect to have time to dedicate to this for now.

PR #4074, opened 2022-08-12
BUG: fix a couple broken lines in lodgeit.py

neutrinoceros:
## PR Summary
Fixes the errors reported by mypy

```
yt/utilities/lodgeit.py: note: In function "print_languages":
yt/utilities/lodgeit.py:193: error: Name "cmp" is not defined  [name-defined]
yt/utilities/lodgeit.py: note: At top level:
yt/utilities/lodgeit.py:332: error: Missing positional argument "filename" in call to "main"  [call-arg]
Found 2 errors in 1 file (checked 483 source files)
Error: Process completed with exit code 1.
```

PR #4075, opened 2022-08-12
BUG: Off-axis projections of filtered gas particles are not scaled correctly

jzuhone:
## PR Summary

I discovered that off-axis projections of filtered gas particles have incorrect scaling. This can be seen clearly in the following notebook:

https://gist.github.com/f97283ce13c4850e31f4b0715f31202f

where the colorbar ranges for the on-axis and off-axis projection should be the same (same axis, different methods), but they are not. The difference arises because the off-axis projection code for SPH particles scales by the units of the `smoothing_length` field to compute the path length, and this is done incorrectly for filtered particles. The reason for this is because the projection calculation is done in code units, and for non-filtered gas particles the path length scaling is done in the units of the `smoothing_length`, which is always `code_length`. However, for filtered gas particles, the smoothing length is in default units of cm (or whatever the unit system happens to be).

The solution is to simply assume that the unit for the path length scaling will always be `code_length`, since the calculation above it of the projection is explicitly carried out in code units. 

## PR Checklist

### DIFF
--- a/yt/visualization/volume_rendering/off_axis_projection.py
+++ b/yt/visualization/volume_rendering/off_axis_projection.py
@@ -228,12 +228,10 @@ def off_axis_projection(
                 )
 
             # Assure that the path length unit is in the default length units
-            # for the dataset by scaling the units of the smoothing length
-            path_length_unit = data_source.ds._get_field_info(
-                (ptype, "smoothing_length")
-            ).units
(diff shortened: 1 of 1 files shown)


neutrinoceros:
I don't know why mypy suddenly started reporting these issues but they are definitely not caused by your PR, and solved in #4074

neutrinoceros (approved):

PR #4076, opened 2022-08-12
BLD: build wheels for CPython 3.11

neutrinoceros:
## PR Summary
Upgrade CIBuildWheel [now with support for CPython 3.11](https://github.com/pypa/cibuildwheel/releases/tag/v2.9.0)

In short, CPython 3.11 has reached ABI stability with their first release candidate (3.11.0-rc1), so it's now safe to build wheels for it, even if the final release isn't out yet (expected early October).

Note that numpy has already successfully published some wheels for CPyhton 3.11 (look for "cp311" at https://pypi.org/simple/numpy/) with similar infrastructure.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -18,13 +18,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-      # pinning OS versions to oldest versions available
-      # because forward compatibility is much easier to
-      # guarantee than backward compatibility
         os: [
-          ubuntu-18.04,
+          ubuntu-20.04,
           windows-2019,
-          macos-10.15,
+          macos-11,
         ]
       fail-fast: false
 
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -44,7 +44,6 @@ jobs:
           CIBW_ARCHS_WINDOWS: "auto"
           CIBW_ENVIRONMENT: "LDFLAGS='-static-libstdc++'"
           CIBW_BUILD_VERBOSITY: 1
-          CIBW_BEFORE_BUILD: "rm -rf build/"  # working around .so files accumulation between builds leading up to bloated wheels
 
       - uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -33,11 +33,11 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Build wheels for CPython
-        uses: pypa/cibuildwheel@v2.4.0
+        uses: pypa/cibuildwheel@v2.9.0
         with:
           output-dir: dist
         env:
-          CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-*"
+          CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-* cp311-*"
           CIBW_SKIP: "*-musllinux_*"  #  numpy doesn't have wheels for musllinux so we can't build some quickly and without bloating
           CIBW_ARCHS_LINUX: "x86_64"
           CIBW_ARCHS_MACOS: x86_64 arm64
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -38,6 +38,7 @@ jobs:
           CIBW_SKIP: "*-musllinux_*"  #  numpy doesn't have wheels for musllinux so we can't build some quickly and without bloating
           CIBW_ARCHS_LINUX: "x86_64"
           CIBW_ARCHS_MACOS: x86_64 arm64
+          MACOSX_DEPLOYMENT_TARGET: "10.9" # as of CIBW 2.9, this is the default value, pin it so it can't be bumped silently
           CIBW_ARCHS_WINDOWS: "auto"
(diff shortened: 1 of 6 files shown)


neutrinoceros:
it's not working out of the box for win32 + CPython 3.11
I'll follow numpy's similar patch for now https://github.com/numpy/numpy/pull/22102

neutrinoceros:
So it's actually failing for ALL CPython 3.11 + windows builds.
These are the error messages:
```
    D:\a\yt\yt\yt\utilities\lib\platform_dep_math.hpp(11): error C3861: '_fpclass': identifier not found
    D:\a\yt\yt\yt\utilities\lib\platform_dep_math.hpp(11): error C2065: '_FPCLASS_NN': undeclared identifier
    D:\a\yt\yt\yt\utilities\lib\platform_dep_math.hpp(11): error C2065: '_FPCLASS_PN': undeclared identifier
```
For reference, these identifiers are documented as Windows-only:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fpclass-fpclassf?view=msvc-170

and so is the code block where they appear in `platform_dep_math.hpp`
So it's not clear to me why it's failing _specifically_ with CPython 3.11. There are many places where things can break in the supply chain, but most likely candidates could be  CIBW 2.9 and CPython 3.11.0-rc1

yt/utilities/lib/platform_dep.h
```diff
@@ -1,6 +1,7 @@
 #include <math.h>
 #ifdef MS_WIN32
 #include "malloc.h"
+#include <float.h>
```
neutrinoceros, line 4: `_fpclass`, `_FPCLASS_NN` and `_FPCLASS_PN` are Windows only symbols defined in `<float.h>`. It seems that CPython 3.10 and older leak `<float.h>`, but CPython 3.11 doesn't, hence the fix.

.github/workflows/wheels.yaml
```diff
@@ -33,18 +33,17 @@ jobs:
...
         env:
-          CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-*"
+          CIBW_BUILD: "cp37-* cp38-* cp39-* cp310-* cp311-*"
           CIBW_SKIP: "*-musllinux_*"  #  numpy doesn't have wheels for musllinux so we can't build some quickly and without bloating
           CIBW_ARCHS_LINUX: "x86_64"
           CIBW_ARCHS_MACOS: x86_64 arm64
           CIBW_ARCHS_WINDOWS: "auto"
           CIBW_ENVIRONMENT: "LDFLAGS='-static-libstdc++'"
           CIBW_BUILD_VERBOSITY: 1
-          CIBW_BEFORE_BUILD: "rm -rf build/"  # working around .so files accumulation between builds leading up to bloated wheels
```
neutrinoceros, line 47: I checked that the total wheel size doesn't change by removing this line

neutrinoceros:
this now also clears #4077

.github/workflows/wheels.yaml
```diff
@@ -18,13 +18,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-      # pinning OS versions to oldest versions available
-      # because forward compatibility is much easier to
-      # guarantee than backward compatibility
         os: [
-          ubuntu-18.04,
+          ubuntu-20.04,
```
neutrinoceros, line 22: see #4077 here

neutrinoceros:
@Xarthisius can you have a look ?

tests/ci_install.sh
```diff
@@ -55,11 +57,10 @@ else
     # include a pyproject.toml file or use any pip-comptatible solution to remedy this.
     python -m pip install numpy>=1.19.4 cython~=0.29.21
 
-    # this is required for cartopy. It should normally be specified in our setup.cfg as
-    # cartopy[plotting]
-    # However it doesn't work on Ubuntu 18.04 (used in CI at the time of writing)
-    python -m pip install shapely --no-binary=shapely
```
Xarthisius, line 61: I remember that causing problems in envs different than 18.04 too. Is really worth dropping? This way we ensure that shapely builds against the same libraries as cartopy. Unless of course they've fixed all that mess upstream.

Xarthisius (dismissed):
Except that one comment LGTM!

tests/ci_install.sh
neutrinoceros: Ah, good call I guess the comment isn't accurate enough then. I can revert this change.

I also note that we actually install cartopy only on Windows (via conda), so do you think this block should be moved to the windows-only part of the script ?

Xarthisius: I thought `[full]` brings it in?

Xarthisius: OK, but that script is also used in testing right? Cause otherwise why would we add it in the first place if nothing uses it?

neutrinoceros: I suppose it may be a leftover from a time we _did_ try to have `cartopy` in `[full]`, but honestly not sure. I'm happy to do a pure, conservative revert if you think it's best.

for reference the problem seems to still be real, see
- https://github.com/SciTools/cartopy/issues/805
- https://github.com/SciTools/cartopy/issues/2006

neutrinoceros: (I dropped this commit)

neutrinoceros:
@Xarthisius should someone else take a look too, or is this good to go ?

Xarthisius (approved):

PR #4084, opened 2022-08-16
Compute gradients for datasets with non-periodic boundaries

### LINKED ISSUE #4083: Gradient computation at non-periodic boundaries
### Bug report

**Bug summary**

Currently, in order to visualize a gradient field of the entire domain (e.g., SlicePlot) with non-periodic boundaries, we need to use the function `ds.force_periodicity()` to gather ghost zones assuming periodicity. If the specific flow property isn't too different across the boundary, this is fine. Unfortunately, I have been dealing with a lot of data sets that this is not the case, and by forcing periodicity, this incorrect gradient field ends up changing the color bar substantially and the resulting image is difficult to interpret. 

**Code for reproduction**

An example of this is given below for a 2D buoyant diffusion flame. The curldrop link to access the data is here: http://use.yt/upload/c2383a4c

```python
import yt

ds = yt.load('/Users/mikemeehan/Research/yt_bndry_bug/data/plt144000', unit_system="mks")

sz = yt.SlicePlot(ds,'z',('boxlib','density'), origin="native")
sz.set_cmap(field=('boxlib','density'), cmap='dusk')
sz.set_log(('boxlib','density'), False)
sz.save(name='./images/density.png')

sz = yt.SlicePlot(ds,'z',('boxlib','x_velocity'), origin="native")
sz.set_cmap(field=('boxlib','x_velocity'), cmap='RdBu')
sz.set_log(('boxlib','x_velocity'), False)
sz.save(name='./images/x_velocity.png')

sz = yt.SlicePlot(ds,'z',('boxlib','y_velocity'), origin="native")
sz.set_cmap(field=('boxlib','y_velocity'), cmap='Blues')
sz.set_log(('boxlib','y_velocity'), False)
sz.save(name='./images/y_velocity.png')
```
produces the following fields
<section>
    
    
    
</section>

If we now want to compute dux/dx, we need to do

```python
ds.add_gradient_fields(('boxlib','x_velocity'))
ds.force_periodicity()

sz = yt.SlicePlot(ds,'z',('boxlib','x_velocity_gradient_x'), origin="native")
sz.set_cmap(field=('boxlib','x_velocity_gradient_x'), cmap='RdBu')
sz.set_log(('boxlib','x_velocity_gradient_x'), False)
sz.save(name='./images/x_velocity_gradient_x.png')
```
which then produces
x_velocity_gradient_x

You can see from the right edge of the image, the colorbar is adapting to the incorrect gradients, making understanding the flow development difficult. If periodicity is not forced, an error is produced. As one might guess, this issue is present for a variety of other flow quantities.

**Expected outcome**

The expected outcome is the following image. 
x_velocity_gradient_x
To fix the issue, I replaced the cells that improperly used periodicity using an extrapolation method. I am working on a PR to show how to do this, hopefully in a general enough way to be included in the code base :)

**Version Information**
  * Operating System: macOS catalina
  * Python Version: 3.7.7
  * yt version: 4.0.4

Everything was installed via conda.

mameehan5:
## PR Summary

This PR is an attempt to allow the user to visualize the gradient field of an entire domain even if the boundaries are non-periodic without any errors or setting the dataset to be periodic. This will close #4083 . The reason for this PR is described in #4083 .

The basic idea for the fix is to _always_ assume periodicity, then replace the incorrect cells with extrapolation.

## PR Checklist

- [ ] Add functionality for gradient fields
- [ ] Add functionality for covering grids (i.e., extrapolate data)
- [ ] Ensure checks and warnings are in place
- [ ] Document new features with docstrings and narrative docs
- [ ] Add tests to ensure function work properly

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -661,6 +661,9 @@ class YTCoveringGrid(YTSelectionContainer3D):
         self.level = level
         self.left_edge = self._sanitize_edge(left_edge)
         self.ActiveDimensions = self._sanitize_dims(dims)
+        self.fix_nonperiodic = [False, False, False]
+        self.lo_offset = np.array([0,0,0], dtype=int)
+        self.hi_offset = np.array([0,0,0], dtype=int)
 
         rdx = self.ds.domain_dimensions * self.ds.relative_refinement(0, level)
 
@@ -679,6 +682,19 @@ class YTCoveringGrid(YTSelectionContainer3D):
             + self.ds.domain_offset
         )
         self._setup_data_source()
+        
+        # determine if we need to extrapolate data for non-periodic domains
+        for i in range(3):
+            if not self.ds.periodicity[i] and \
+                (self.left_edge[i] < self.ds.domain_left_edge[i] or \
+                 self.right_edge[i] > self.ds.domain_right_edge[i]):
+                self.fix_nonperiodic[i] = True
+            if self.fix_nonperiodic[i]:
+                if self.left_edge[i] < self.ds.domain_left_edge[i]:
+                    self.lo_offset[i] = np.rint((self.ds.domain_left_edge[i] - self.left_edge[i])/self.dds[i])
+                if self.right_edge[i] > self.ds.domain_right_edge[i]:
+                    self.hi_offset[i] = np.rint((self.right_edge[i] - self.ds.domain_right_edge[i])/self.dds[i])
+
         self.get_data(fields)
 
     def get_global_startindex(self):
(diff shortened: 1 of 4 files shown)


mameehan5:
I would really love any feedback as I work through this PR! Anything from naming, syntax, or other fundamental issues I am missing.

jzuhone:
@mameehan5 thanks for looking into this long-standing issue!

I think one of the reasons it's never been addressed is that simulations have different boundary conditions of course (reflecting, outflow, inflow, etc) and to really compute the gradient at the boundaries you'd want to know what the appropriate BC is. 

But this is probably a good solution until we can properly process BCs from simulations.

mameehan5:
@jzuhone thanks for the support! Yep, I definitely agree, that would be the absolute best way to do it. For the moment, I hope to at least give a decent approximation by using an extrapolation method. Using extrapolation then computing the gradient is likely directly related using a single-sided finite difference stencil, likely in a rigorous way, but just getting something functioning would make me happy!

mameehan5:
I realized I incorrectly tried merging from main. I will close this PR and work from the other one which now updated  as well.

PR #4085, opened 2022-08-17
Fix nonperiodic

mameehan5:
## PR Checklist


neutrinoceros:
Hi Mike, thank you for taking this on !

I'll just add a reference to your previous iteration of this PR as there is already a discussion there: https://github.com/yt-project/yt/pull/4084

While I'm at it, here's a link at our style guide and how to work with linters/pre-commit: https://yt-project.org/docs/dev/developing/developing.html#automatically-checking-and-fixing-code-style

yt/data_objects/construction_data_containers.py
```diff
@@ -666,6 +669,9 @@ def __init__(
         self.level = level
         self.left_edge = self._sanitize_edge(left_edge)
         self.ActiveDimensions = self._sanitize_dims(dims)
+        self.fix_nonperiodic = [False, False, False]
+        self.lo_offset = np.array([0,0,0], dtype=int)
+        self.hi_offset = np.array([0,0,0], dtype=int)
```
neutrinoceros, line 674: it's best practice to specify numpy dtypes more explicitly: here `"int64"` instead of builtin `int`

yt/data_objects/static_output.py
```diff
@@ -339,6 +342,18 @@ def force_periodicity(self, val=True):
...
 
+    def set_periodicity(self, val):
+        """
+        Set box periodicity to user defined periodicity.
+        """
+        if isinstance(val, bool):
+            val = (val, val, val)
+        if not isinstance(val, tuple):
+            raise TypeError("set_periodicity must be a boolean or tuple of booleans.")
+        if len(val) != 3 or not all([isinstance(v, bool) for v in val]):
```
neutrinoceros, line 353: `all` accepts generator, no need for a list comprehension (albeit really cheap here)
```suggestion
        if len(val) != 3 or not all(isinstance(v, bool) for v in val):
```

yt/utilities/lib/interpolators.pyx
```diff
@@ -163,3 +163,47 @@ def ghost_zone_interpolate(int rf,
...
+    nz = field.shape[2]
+    lx = lo_offset[0]
+    ly = lo_offset[1]
+    lz = lo_offset[2]
+    hx = hi_offset[0]
+    hy = hi_offset[1]
+    hz = hi_offset[2]
+
+    # Fix boundaries in the x direction
+    if lx > 0 or hx > 0:
```
neutrinoceros, line 192: nit: can we save an indentation level by inverting this condition and return early  when it's _not_ checked ?
e.g.
```python
if lx <= 0 or hx <= 0:
    return
for ...:
    for ...:
```

yt/data_objects/static_output.py
```diff
@@ -175,6 +175,7 @@ class Dataset(abc.ABC):
     domain_offset = np.zeros(3, dtype="int64")
     _periodicity = MutableAttribute()
     _force_periodicity = False
+    _set_periodicity = None
```
neutrinoceros, line 178:
```suggestion
    _periodicity_override: Optional[Tuple[bool, bool, bool]] = None
```

yt/data_objects/static_output.py
```diff
@@ -326,6 +327,8 @@ def periodicity(self):
         elif getattr(self, "_domain_override", False):
             # dataset loaded with a bounding box
             return (False, False, False)
+        if self._set_periodicity is not None:
+            return self._set_periodicity
         return self._periodicity
```
neutrinoceros, line 332: nit; personally I find this better but don't commit it if you find it confusing
```suggestion
        return self._periodicity_override or self._periodicity
```

yt/data_objects/static_output.py
```diff
@@ -339,6 +342,18 @@ def force_periodicity(self, val=True):
             raise TypeError("force_periodicity expected a boolean.")
         self._force_periodicity = val
 
+    def set_periodicity(self, val):
```
neutrinoceros, line 345:
```suggestion
    def set_periodicity(self, val: Union[bool, Tuple[bool, bool, bool]]) -> None:
```

yt/data_objects/static_output.py
```diff
@@ -339,6 +342,18 @@ def force_periodicity(self, val=True):
...
+        """
+        Set box periodicity to user defined periodicity.
+        """
+        if isinstance(val, bool):
+            val = (val, val, val)
+        if not isinstance(val, tuple):
+            raise TypeError("set_periodicity must be a boolean or tuple of booleans.")
+        if len(val) != 3 or not all([isinstance(v, bool) for v in val]):
+            raise TypeError("user supplied periodicities must be a tuple of 3 booleans.")
+        self._set_periodicity = val
```
neutrinoceros, line 355:
```suggestion
        self._periodicity_override = val
```

yt/data_objects/static_output.py
neutrinoceros: I assume it could be useful to accept `None` too so we're allowed to reset the initial state using only public API

yt/data_objects/static_output.py
mameehan5: I do think this is a little confusing so I would personally leave it as is for the moment.

mameehan5: ... at least for myself

yt/data_objects/static_output.py
mameehan5: Yep, agreed!

yt/data_objects/static_output.py
mameehan5: Whoops, good catch.

yt/utilities/lib/interpolators.pyx
mameehan5: Unfortunately, no. At least not this way because I still need to add y and z extrapolation. Let me sit on this for a bit though to see if it is possible to remove the indent with something of this nature!

yt/data_objects/static_output.py
neutrinoceros: That's fine. In that case I would suggest this instead:
```suggestion
        if self._set_periodicity is not None:
            return self._set_periodicity
        else:
            return self._periodicity
```
It's exactly equivalent but I have a small preference for aligned return statements

yt/utilities/lib/interpolators.pyx
neutrinoceros: best if you can, but not critical if you can't :)

cphyc:
@mameehan5 don't worry about the ‘pre-commit.ci’ test failing, this can be fixed automagically once the PR is ready :)

yt/utilities/lib/interpolators.pyx
```diff
@@ -163,3 +163,47 @@ def ghost_zone_interpolate(int rf,
...
+
+    nx = field.shape[0]
+    ny = field.shape[1]
+    nz = field.shape[2]
+    lx = lo_offset[0]
+    ly = lo_offset[1]
+    lz = lo_offset[2]
+    hx = hi_offset[0]
+    hy = hi_offset[1]
+    hz = hi_offset[2]
```
cphyc, line 189: Assuming this is valid Cython code, I'd suggest for readability

```suggestion
    nx, ny, nz = field.shape
    lx, ly, lz = lo_offset
    hx, hy, hz = hi_offset
```

neutrinoceros: looks like tuple unpacking is indeed supported in Cython, for instance we use it here: https://github.com/yt-project/yt/blob/8c8dd9bf2eb767c9bcfaef72a4d069a2067e6777/yt/utilities/lib/pixelization_routines.pyx#L1562

yt/utilities/lib/interpolators.pyx
```diff
@@ -163,3 +163,47 @@ def ghost_zone_interpolate(int rf,
...
+    if lx > 0 or hx > 0:
+        for j in range(lo_offset[1], ny-hi_offset[1]):
+            for k in range(lo_offset[2], field.shape[2]-hi_offset[2]):
+                # Fix xlo
+                qm = field[lx,   j, k]
+                qp = field[lx+1, j, k]
+                for i in range(lx):
+                    xp = float(i - lx)
+                    xm = 1.0 - xp
+                    field[i, j, k] = xm*qm + xp*qp
```
cphyc, line 201: I have to say I don't understand how this code works. For example, in this chunk, at the first iteration, I have `i=0`, which leads to `xp=-lx` and `xm = 1+lx`. This doesn't seem to yield a sensible result unless `lx=-1` or `lx=0`?

Maybe one thing that'd help would be a docstring detailing what the function is doing and what assumptions are made about the values of `lo_offset` and `hi_offset`.

yt/utilities/lib/interpolators.pyx
mameehan5: Yep! Was thinking of changing this out as it seemed excessively large.

yt/utilities/lib/interpolators.pyx
mameehan5: Sorry for the confusion! I will add a docstring detailing this. The function is essentially doing linear extrapolation using the last two points of "valid" data. I mostly copied the ghost zone interpolation function but limited the functionality to the outer regions that need extrapolated data.

yt/data_objects/static_output.py
mameehan5: Yep, I prefer that style too!

mameehan5:
I think this is just about ready for review. I do have two concerns that I want to point out that need to be thought about with this PR:

1. Is there anything that I am missing for the region selector object? I basically just bypassed the error but there are some details later on that include stuff about periodicity that I am not exactly sure how to handle. 
2. Are there other objects besides the covering grids that need to be updated with this feature? I only really worked with these so that is what I fixed, but I am happy to take a look at other objects as well.

I would appreciate any input! Thanks!!

mameehan5:
pre-commit.ci autofix

yt/data_objects/construction_data_containers.py
```diff
@@ -666,6 +666,9 @@ def __init__(
         self.level = level
         self.left_edge = self._sanitize_edge(left_edge)
         self.ActiveDimensions = self._sanitize_dims(dims)
+        self.fix_nonperiodic = [False, False, False]
+        self.lo_buffer = np.array([0, 0, 0], dtype="int64")
+        self.hi_buffer = np.array([0, 0, 0], dtype="int64")
```
neutrinoceros, line 671: From the names only, it's really not clear to me what these attribute are used for, especially `fix_periodicity` because it shares a name with a function you're importing.

yt/data_objects/construction_data_containers.py
```diff
@@ -684,6 +687,46 @@ def __init__(
...
         self._setup_data_source()
+
+        # determine if we need to extrapolate data for non-periodic domains
+        for i in range(3):
+            if not self.ds.periodicity[i] and (
+                self.left_edge[i] < self.ds.domain_left_edge[i]
+                or self.right_edge[i] > self.ds.domain_right_edge[i]
+            ):
+                self.fix_nonperiodic[i] = True
+            if self.fix_nonperiodic[i]:
```
neutrinoceros, line 698: Seems redundant, or am I missing something ?
```suggestion
```
Also, I find this block hard to read, it would help to invert some early condition checks to save indentation levels if we can

yt/data_objects/construction_data_containers.py
```diff
@@ -684,6 +687,46 @@ def __init__(
...
+                    )
+                if self.right_edge[i] > self.ds.domain_right_edge[i]:
+                    self.hi_buffer[i] = np.rint(
+                        (self.right_edge[i] - self.ds.domain_right_edge[i])
+                        / self.dds[i]
+                    )
+        if any(self.fix_nonperiodic):
+            if np.count_nonzero(self.fix_nonperiodic) == 1:
+                extrap_dims = (
+                    "xyz"[np.where(self.fix_nonperiodic)[0][0]] + " " + "dimension"
```
neutrinoceros, line 711: be careful here that "xyz" only works for cartesian geometries. `self.ds.coordinates.axis_order` is the general iterable for looping over axes names

yt/data_objects/static_output.py
```diff
@@ -339,6 +343,20 @@ def force_periodicity(self, val=True):
             raise TypeError("force_periodicity expected a boolean.")
         self._force_periodicity = val
 
+    def set_periodicity(self, val: Union[bool, Tuple[bool, bool, bool]]) -> None:
```
neutrinoceros, line 346: round 2: this is in direct competition with `force_periodicity`. Since it's much more general, I suggest refactoring around this new method:
- get rid of the `_force_periodicity` attr
- make the `force_periodicity` method call `set_periodicity` internally.

neutrinoceros (changes requested):
I want to review this again, but here are a couple important remarks for now

yt/data_objects/construction_data_containers.py
mameehan5: I agree, "xyz" isn't general, but doesn't a covering grid assume Cartesian geometry? I was under this impression, for a number of reasons, such as the container fields being Cartesian: 
```   
     _container_fields = (
        ("index", "dx"),
        ("index", "dy"),
        ("index", "dz"),
        ("index", "x"),
        ("index", "y"),
        ("index", "z"),
    )
```
or the coarse grid size being
```
self.base_dds = self.ds.domain_width / self.ds.domain_dimensions
```
Happy to change this iterable though, just want to make sure I am not missing anything!

mameehan5:
@neutrinoceros thank you for the detailed comments! Let me address these.

yt/data_objects/static_output.py
mameehan5: Definitely, let me fix this.

yt/data_objects/construction_data_containers.py
mameehan5: Agreed, I will fix this!

yt/data_objects/construction_data_containers.py
mameehan5: I agree, let me change this around and come up with something more clear

yt/data_objects/construction_data_containers.py
neutrinoceros: Let me get back to you on this

yt/data_objects/static_output.py
```diff
@@ -341,12 +338,19 @@ def force_periodicity(self, val=True):
         # workaround in yt user codes.
         if not isinstance(val, bool):
             raise TypeError("force_periodicity expected a boolean.")
-        self._force_periodicity = val
+        if val is True:
+            self.set_periodicity(True)
+        else:
+            self.set_periodicity(None)
```
neutrinoceros, line 344: `force_periodicity` is meant to lock the observable state of `self.periodicity`, reseting defaults with `None` isn't among the things it can do. For full backward compatibility I think it should simply be this
```suggestion
        self.set_periodicity(val)
```

mameehan5: Yes, that makes sense! I should have realized that.

neutrinoceros: In insight I understand it is not obvious, `force_periodicity(False)` is ambiguous from the outside.
I think your `set_periodicity` is much clearer in what it does. Btw I encourage you to grep for `force_periodicity` throughout the code base.
In most places I think you could just replace it with `set_periocity`, especially in documentation. In one case (`doc/source/cookbook/hse_field.py`) it looks like you could just get rid of it.

neutrinoceros:
@mameehan5 careful, it looks like something went wrong a couple commits ago and tests are failing at collection with
```
E   ImportError: cannot import name 'fix_nonperiodic' from 'yt.utilities.lib.interpolators' 
```

neutrinoceros (approved):

neutrinoceros:
Sorry I just wanted to mark my "changes requested" as resolve. I plan to give this an actual second review tonight

yt/data_objects/construction_data_containers.py
```diff
@@ -684,6 +690,44 @@ def __init__(
             + self.ds.domain_offset
         )
         self._setup_data_source()
+
+        # determine if we need to extrapolate data for non-periodic domains
```
neutrinoceros, line 694:
```suggestion
        # extrapolate on non-periodic data if the grid reaches domain edges
```

yt/data_objects/construction_data_containers.py
```diff
@@ -684,6 +690,44 @@ def __init__(
...
+                    + " and "
+                    + "xyz"[np.where(self.extrap_nonperiodic)[0][1]]
+                    + " dimensions"
+                )
+            if np.count_nonzero(self.extrap_nonperiodic) == 3:
+                extrap_dims = "x, y, and z dimensions"
+            warnings.warn(
+                "The region specified is outside of the "
+                "computational domain with non-periodic boundaries. "
+                "Therefore, we will extrapolate data for the "
```
neutrinoceros, line 726: it'd be good to say what kind (order) of interpolation is performed in the warning

yt/data_objects/construction_data_containers.py
```diff
@@ -684,6 +690,44 @@ def __init__(
...
+            if np.count_nonzero(self.extrap_nonperiodic) == 2:
+                extrap_dims = (
+                    "xyz"[np.where(self.extrap_nonperiodic)[0][0]]
+                    + " and "
+                    + "xyz"[np.where(self.extrap_nonperiodic)[0][1]]
+                    + " dimensions"
+                )
+            if np.count_nonzero(self.extrap_nonperiodic) == 3:
+                extrap_dims = "x, y, and z dimensions"
+            warnings.warn(
```
neutrinoceros, line 723: this probably should be a logger warning instead. Actual warnings such as this one should be reserved to cases where users _can_ (or _should_) avoid them.

yt/utilities/tests/test_interpolators.py
```diff
@@ -143,3 +146,33 @@ def test_get_vertex_centered_data():
...
+            lo[1] : hi[1] : nx[1] * 1j,
+            lo[2] : hi[2] : nx[2] * 1j,
+        ]
+        func = x + y + z
+        field = np.zeros(nx, dtype="float64")
+        field[nbuf : nx[0] - nbuf, nbuf : nx[1] - nbuf, nbuf : nx[2] - nbuf] = func[
+            nbuf : nx[0] - nbuf, nbuf : nx[1] - nbuf, nbuf : nx[2] - nbuf
+        ]
+
+        # Use fix_nonperiodic function to extrapolate data to buffer region
```
neutrinoceros, line 172: seems like you missed a renamed method in this comment

yt/utilities/tests/test_interpolators.py
```diff
@@ -143,3 +146,33 @@ def test_get_vertex_centered_data():
...
+        replace_nonperiodic_with_extrap(
+            field, np.array([nbuf, nbuf, nbuf]), np.array([nbuf, nbuf, nbuf])
+        )
+
+        # Since this function is linear, the extrapolation should be exact
+        assert_array_equal(field.ravel(), func.ravel())
```
neutrinoceros, line 178: even then, `assert_array_equal` is discouraged because we can often actually tolerate machine precision rounding errors. I suspect that it could happen here (maybe only on some architecture), so I'd advise caution, and recommend using `assert_allclose` instead (with a very low, but non-zero, margin of error)

yt/utilities/lib/interpolators.pyx
```diff
@@ -163,3 +163,115 @@ def ghost_zone_interpolate(int rf,
...
+
+@cython.cdivision(True)
+@cython.wraparound(False)
+@cython.boundscheck(False)
+def replace_nonperiodic_with_extrap(np.ndarray[np.float64_t, ndim=3] field,
+                                    np.ndarray[np.int_t,     ndim=1] ncell_lo,
+                                    np.ndarray[np.int_t,     ndim=1] ncell_hi):
+    """
+    The purpose of this function is to extrapolate data from the interior of
+    the buffer region to replace the given data in the buffer region using a
```
neutrinoceros, line 175: this sentence isn't clear to me. In particular, I don't understand what you mean by 

> the given data in the buffer region

yt/utilities/lib/interpolators.pyx
```diff
@@ -163,3 +163,115 @@ def ghost_zone_interpolate(int rf,
...
+    field : 3D array
+        Three-dimensional data of the flowfield
+    ncell_lo : 1D array of length 3
+        Number of buffer cells to replace on the low side along each axis
+    ncell_hi : 1D array of length 3
+        Number of buffer cells to replace on the high side along each axis
+
+    This is best illustrated through a one-dimensional example. Consider a
+    linear function: f(x) = x , and the `field` variable was filled using
+    periodicity, so we have the following data:
```
neutrinoceros, line 189: better be precise on what you mean by "filled" here (I'm assuming that you mean "values on each end of an array", or something of the essence)

yt/utilities/lib/interpolators.pyx
```diff
@@ -163,3 +163,115 @@ def ghost_zone_interpolate(int rf,
...
+
+    where field.shape ---> (6,1,1), ncell_lo ---> (1,0,0), and
+    ncell_hi ---> (1,0,0) . For this example, we replace the value of
+    field[0,0,0] using linear extrapolation from field[1,0,0] and field[2,0,0],
+    similarly for upper boundary. Evaluating this function then yields the
+    following output of `field`:
+
+    -------------------------------------------
+    field:   -0.5 | 0.5 | 1.5 | 2.5 | 3.5 | 4.5
+    -------------------------------------------
```
neutrinoceros, line 207: pet peeve, but this would look better if it was as wide as the first table

neutrinoceros (commented):
Ok, round 2. I still need to review the actual extrapolation routine, but so far I must say it's looking great.

mameehan5:
Thanks @neutrinoceros for taking another look! I will work on those changes and let you know once they are done.

yt/utilities/lib/interpolators.pyx
```diff
@@ -163,3 +163,115 @@ def ghost_zone_interpolate(int rf,
...
+    """
+    cdef int i, j, k
+    cdef int ii, ij, ik
+    cdef int nx, ny, nz
+    cdef int lx, ly, lz
+    cdef int hx, hy, hz
+    cdef np.float64_t xp, xm, yp, ym, zp, zm
+    cdef np.float64_t qp, qm
+
+    nx, ny, nz = (field.shape[0], field.shape[1], field.shape[2])
```
neutrinoceros, line 221: assuming this is valid
```suggestion
    nx, ny, nz = field.shape
```

yt/utilities/lib/interpolators.pyx
```diff
@@ -163,3 +163,115 @@ def ghost_zone_interpolate(int rf,
...
+    lx, ly, lz = ncell_lo
+    hx, hy, hz = ncell_hi
+
+    # Fix boundaries in the x direction
```
neutrinoceros, line 225: and similar comments for y and z
```suggestion
    # Extrapolate in the x direction
```

neutrinoceros (commented):
I read the extrapolation method too, I think it looks very good and is likely to be very close to optimal already, so I don't have any major comment here.

yt/data_objects/construction_data_containers.py
```diff
@@ -666,6 +669,9 @@ def __init__(
         self.level = level
         self.left_edge = self._sanitize_edge(left_edge)
         self.ActiveDimensions = self._sanitize_dims(dims)
+        self.extrap_nonperiodic = [False, False, False]
+        self.ncell_lo_extrap = np.array([0, 0, 0], dtype="int64")
+        self.ncell_hi_extrap = np.array([0, 0, 0], dtype="int64")
```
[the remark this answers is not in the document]
mameehan5: I agree!

yt/utilities/tests/test_interpolators.py
mameehan5: good catch!

mameehan5:
Will finish up looking at this in a few hours.

yt/utilities/lib/interpolators.pyx
mameehan5: Unfortunately this is not valid in cython.

yt/utilities/lib/interpolators.pyx
mameehan5: Yep, got it!

yt/utilities/lib/interpolators.pyx
mameehan5: agreed

mameehan5:
This is ready to look at now! Not exactly sure why this test is failing though.

neutrinoceros:
The failure is unrelated, this particular test is just flaky. Let's run it again:
@yt-fido test this please

neutrinoceros (dismissed):
Hey, sorry for the little delay, I wanted to make extra sure this did not have bad side effects in non-cartesian geometries and I report that ... it actually paves the way to fix a very similar problem we're having there by offering a work around that we currently don't have.
Here's for example of a density snapshot in spherical geometry
raw
One can see the radial gradient is clearly negative on the outer rim. Naturally, the domain is *not* periodic in this direction.
This is what the gradient would normally look like
forced_nope

However, on the main branch, obtaining this image isn't possible, and the field seems to be always treated as periodic in the radial direction, no matter if/how one uses `ds.force_periodicity`, and you're stuck with this erroneous viz
forced_nope

I note that the *correct* image requires an explicit call to `force_peridocity(False)` even on this branch, so there will still be something to correct, but it seems clear to me that this PR sets us on the right track, thank you very much for working on this !

neutrinoceros:
I would really like to see this going before the next (upcoming) release, to I'll give it a chance to be discussed with this in mind

mameehan5:
@neutrinoceros no problem for the delay and thanks for checking the non-cartesian data. I am glad this also helps fix those issues you were seeing!

The one last concern that I have is the modifications that I made to yt/geometry/_selection_routines/region_selector.pxi . I am not sure how closely you looked at this, but I want to double check to make sure my changes are appropriate and that I didn't miss anything near the bottom of the code since this seems like a core function.

neutrinoceros:
I don't have a particular expertise there so I trust the tests on that point. I also want a second opinion. I will advertise this PR in development channels if needed.

yt/geometry/_selection_routines/region_selector.pxi
```diff
@@ -41,42 +41,23 @@ cdef class RegionSelector(SelectorObject):
                         i, region_width[i]))
 
         for i in range(3):
+            # First, we check if any criteria requires a period check,
```
matthewturk, line 44: @jzuhone can you look at this in light of your recent changes?

yt/utilities/tests/test_interpolators.py
```diff
@@ -143,3 +151,34 @@ def test_get_vertex_centered_data():
...
+        hi = lo + (nx - 1) * dx
+
+        # Fill field with analytical linear function: f(x,y,z) = x + y + z
+        # excluding the buffer region
+        x, y, z = np.mgrid[
+            lo[0] : hi[0] : nx[0] * 1j,
+            lo[1] : hi[1] : nx[1] * 1j,
+            lo[2] : hi[2] : nx[2] * 1j,
+        ]
+        func = x + y + z
```
cphyc, line 171: Just to be sure we're testing x, y, and z and that there isn't any confusion between different axes, I'd suggest this minor replacement:
```suggestion
        func = x + 2 * y + 3 * z
```

cphyc (dismissed):
Looks good to go! I've added a small suggestion but that should not be a blocker.

jzuhone:
I plan to review this and (almost certainly) approve tomorrow.

yt/utilities/tests/test_interpolators.py
mameehan5: @cphyc thank you! Yes, I agree, this is a good change to be as thorough as possible with the testing.

mameehan5:
@jzuhone sounds good, thank you!

yt/geometry/_selection_routines/region_selector.pxi
jzuhone: @mameehan5 could you explain a bit why these changes were needed? we just dealt with some long-standing issues in this particular class and I want to make sure we're not accidentally breaking anything

yt/geometry/_selection_routines/region_selector.pxi
```diff
@@ -41,42 +41,23 @@ cdef class RegionSelector(SelectorObject):
...
-                        "or override the periodicity with\n"
-                        "ds.force_periodicity()" % \
-                        (i, dobj.left_edge[i], dobj.right_edge[i],
-                         dobj.ds.domain_left_edge[i], dobj.ds.domain_right_edge[i])
-                    )
             # Already ensured in code
             self.left_edge[i] = LE[i]
             self.right_edge[i] = RE[i]
             self.right_edge_shift[i] = RE[i] - DW[i]
-            if not self.periodicity[i]:
```
jzuhone, line 78: I'm wondering particularly about these lines--`right_edge_shift` gets used below in a few places and I want to make sure that it still does the right thing in the case of non-periodic boundaries. It's particles in particular that I am worried about

yt/geometry/_selection_routines/region_selector.pxi
mameehan5: @jzuhone these changes were all mostly to bypass the `RuntimeError` below. Do you remember what this issues was? Or, do you possibly have an example particle dataset you are worried about? I don't see any issues on my end with my dataset, but I do know it is only a small subset of the various types of datasets yt needs to handle.

yt/geometry/_selection_routines/region_selector.pxi
mameehan5: I agree (and note my comments above). This was a section of the code that I was most worried about. Let me take a closer look to make sure what I am doing makes logical sense. Again, if you have a dataset that would be a good test case, I would appreciate it!

jzuhone: @mameehan5 I apologize again for the delay. Hoping to get this settled this week. It won't take long, just trying to find the time.

mameehan5: No problem, just let me know when you get the chance!

jzuhone: Ok--I've looked at this a bit more, and I think I'm concerned that we don't fully understand the ramifications of taking away the line that sets `self.right_edge_shift[i] = -np.inf` for a non-periodic boundary. The rest of it looks ok to me. If you had a selection box that was fully within the domain, then it looks like you would never have to worry about `right_edge_shift`, because whether it was `right_edge-domain_width` or `-np.inf` it would still be outside of the domain. 

But what if you didn't? What if you had code that defined a selection box that extended way off the domain in some direction, say over a domain width? Obviously for a non-periodic boundary it should only keep the data that's within the domain, and this is guaranteed if `right_edge_shift = -np.inf`. But if you had `right_edge_shift = right_edge - domain_width` then it seems like the behavior could be different. 

Is there any way to restore this line without breaking everything you're trying to accomplish? That would be preferable, if it's possible.

jzuhone: My example is obviously a bit silly, but I can see code that iterates over box regions for whatever reason happening upon a box that straddles a boundary, and we'd want to make sure it was always "doing the right thing" from a selection perspective even if the region extending far off the domain makes little physical sense.

mameehan5: That definitely makes sense! I think I should be able to restore that. Let me test it out and get back to you.

mameehan5:
@jzuhone Still working through this. It did in fact pose more issues than anticipated so once I have more time to take a deep dive, I should be able to sort it out. I will let you know by the end of the week.

neutrinoceros:
For later reference, I previously triaged this for the 4.1 milestone, but I'm going to take it back, because that release is already far behind schedule. I would still be happy to include the patch in the release if we can get a second review in time, which seems likely, I just want to signal that the patch shouldn't be considered blocking.

mameehan5:
Got it. Still planning to look at this later in the week.

neutrinoceros:
FYI I merged from main to resolved a couple conflicts

neutrinoceros:
resolved conflicts again and cleaned up to satisfy newly introduced linting rules (unused declarations)

neutrinoceros:
@mameehan5 hey, I hope you are well. It'd still be awesome to have this in yt, but I'll remove this from the yt 4.2 milestone so you can choose to come back to this whenever you like.

mameehan5:
No problem, that sounds good. Thanks for reaching out!

From: Clément Robert ***@***.***>
Date: Tuesday, May 2, 2023 at 3:28 PM
To: yt-project/yt ***@***.***>
Cc: Meehan, Michael Alexander ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [yt-project/yt] Compute gradients for datasets with non-periodic boundaries (PR #4085)
You don't often get email from ***@***.*** Learn why this is important<https://aka.ms/LearnAboutSenderIdentification>

@mameehan5<https://github.com/mameehan5> hey, I hope you are well. It'd still be awesome to have this in yt, but I'll remove this from the yt 4.2 milestone so you can choose to come back to this whenever you like.

PR #4089, opened 2022-08-20
BUG: avoid using deprecated matplotlib.cm API

### LINKED ISSUE #4088: BUG: pending deprecation warnings from MPL 3.6
### Bug report

**Bug summary**

Matplotlib's colormap API is still evolving and new pending deprecation warnings are now emitted, which breaks yt's bleeding edge CI

example failed job: https://github.com/yt-project/yt/runs/7929987453?check_suite_focus=true
origin of the change: https://github.com/matplotlib/matplotlib/pull/23668

I note that `cmyt` is unaffected by these new warnings.

neutrinoceros:
## PR Summary
Fix #4088

running the bleeding edge CI workflow on my fork at https://github.com/neutrinoceros/yt/actions/runs/2894300482

### DIFF
--- a/doc/source/visualizing/colormaps/cmap_images.py
+++ b/doc/source/visualizing/colormaps/cmap_images.py
@@ -1,4 +1,4 @@
-import matplotlib.cm as cm
+import matplotlib as mpl
 
 import yt
 
@@ -8,7 +8,7 @@ ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")
 # Create projections using each colormap available.
 p = yt.ProjectionPlot(ds, "z", "density", weight_field="density", width=0.4)
 
-for cmap in cm.datad:
+for cmap in mpl.colormaps:
     if cmap.startswith("idl"):
         continue
     p.set_cmap(field="density", cmap=cmap)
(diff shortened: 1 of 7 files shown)


neutrinoceros:
I think this and #4076 call for a new bugfix release to make downstream applications more easily testable

doc/source/visualizing/colormaps/cmap_images.py
```diff
@@ -8,7 +8,7 @@
 # Create projections using each colormap available.
 p = yt.ProjectionPlot(ds, "z", "density", weight_field="density", width=0.4)
 
-for cmap in cm.datad:
```
neutrinoceros, line 11: `matplotlib.cm.datad` isn't deprecated but it's also not clearly intended to be exposed, as it just happens to be there as a side effect of a re-export, see https://github.com/matplotlib/matplotlib/blob/9b1fcf67c4228c4a2788af5bcaf0c6fde09a55bf/lib/matplotlib/cm.py#L26

so I've replaced it with current API, which would break for old versions of Matplotlib, but I think it's okay to do that in a docs script ?

neutrinoceros:
@yt-fido test this please

brittonsmith (approved):

PR #4090, opened 2022-08-21
Re-adding scipy optimize to on-demand imports

chummels:
Recently, when the on-demand imports were being cleaned up in [#c501b8c](https://github.com/yt-project/yt/commit/c501b8cc746b4504a9cdcb6d48b85138d6bda390), scipy.optimize was removed as an alias.  It turns out that this broke Trident, so I'd like to add it back in.

### DIFF
--- a/yt/utilities/on_demand_imports.py
+++ b/yt/utilities/on_demand_imports.py
@@ -252,6 +252,12 @@ class scipy_imports(OnDemand):
 
         return ndimage
 
+    @safe_import
+    def optimize(self):
+        from scipy import optimize
+
+        return optimize
+
 
 _scipy = scipy_imports()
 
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Seems reasonable. I suggest adding a comment there so we know why it's still here, though in the long run I'd suggest trident could use it's own on-demand module.

chummels:
Well, I considered adding it to Trident, which used to be part of yt several years back, which may have been the reason that this code was present.  But it's a lot more code to add all the on-demand stuff to Trident rather than just adding these 4 lines to yt.  I'll add a small comment.

neutrinoceros:
I understand that. If you're interested I would gladly volunteer to do the work on Trident too.

neutrinoceros (approved):

PR #4092, opened 2022-08-22
BUG: avoid using deprecated matplotlib.cm API in volume rendering (round 2)

### LINKED ISSUE #4088: BUG: pending deprecation warnings from MPL 3.6

neutrinoceros:
## PR Summary
Fix #4088 again

### DIFF
--- a/yt/visualization/_handlers.py
+++ b/yt/visualization/_handlers.py
@@ -5,7 +5,6 @@ from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import numpy as np
 import unyt as un
-from matplotlib.cm import get_cmap
 from matplotlib.colors import Colormap, LogNorm, Normalize, SymLogNorm
 from packaging.version import Version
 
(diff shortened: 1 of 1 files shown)


brittonsmith (approved):
Looks ok to me. Does the import need to happen inside the property?

neutrinoceros:
It doesn't. The reason I put it there is to make extra sure no one ever relies on being able to import it as
```python
from yt.visualization._handlers import _get_cmap
```
but there are already 2 `_*` prefixes objects along the way so it should be pretty obvious already that it's not a good idea.

I can put the import at the top level if you have a preference.

brittonsmith:
I suppose I'm in favor of moving the import up top, but only mildly. The underscores should be enough to deter importing, and it cuts down on a duplicate line of code.

neutrinoceros:
Done !

brittonsmith (approved):

PR #4095, opened 2022-08-23
TST: bump Python from 3.7 to 3.9 in answer testing

### LINKED ISSUE #3932: TST: incoming need for a Python upgrade in image tests
### Bug report

**Bug summary**

Matplotlib's next feature release, 3.6, will drop support for Python 3.7, see
https://github.com/matplotlib/matplotlib/blob/9cf4611377acbcf4d7191672944df3ae4b30d918/setup.py#L11-L12

Our image comparison answer tests, on the other hand, are pinned to a specific Python version, which is currently 3.7, but their intent is to be run against the newest available matplotlib version at any point in time, since that's what any user will get when installing yt in a new environment.

This means we'll soon need to bump the Python version used for answer tests.
Note that we also cannot bump beyond Python 3.9 since answer tests require nose, which is incompatible with Python 3.10 and beyond. Medium term, we need to migrate these tests to pytest (#3735)

Fixing this will become necessary when MPL 3.6 hits, and require a massive PR to the answer store repo as well as a companion PR on the main repo (this one)

neutrinoceros:
## PR Summary

It's time to fix #3932, as matplotlib is starting the release process for version 3.6 [without wheels for Python 3.7](https://pypi.org/project/matplotlib/3.6.0rc1/#files), which means our reference answers will soon be out of sync with their latest release.

This will require a supporting PR to the answer store submodule to create reference answers for Python 3.9
Edit: here https://github.com/yt-project/answer-store/pull/32

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -50,7 +50,7 @@ jobs:
             # answer tests use 'yield', so they require nose
             # they are also attached to a specific, occasionally updated, Python version
             # but it does *not* have to match the current minimal supported version
-            python-version: '3.7'
+            python-version: '3.9'
             dependencies: full
             tests-type: answer
             test-runner: nose


neutrinoceros:
Merging this will require some updates to the auto-merge requirements. The simplest strategy is to stop requiring the "py37 answers" job, merge, and at some later point maybe start requiring the replacement "py39 answers".

neutrinoceros:
Matplotlib 3.6.0 is moving faster than I expected, [the milestone is currently at 30 items left](https://github.com/matplotlib/matplotlib/milestone/65). It was at ~100 a week ago.
At this rate it would be preferable to move forward with this PR in the coming days.

neutrinoceros:
@matthewturk I will need your help with this. It shouldn't be complicated, but it requires owner privileges.

matthewturk (approved):

PR #4097, opened 2022-08-25
ENH: Store the Enzo-E version number in parameters attribute of EnzoEDataset

mabruzzo:
More recent versions of `Enzo-E` record the version number in the outputs. This PR now stores that in the `parameters` attribute of `EnzoEDataset` instances (the value is associated with 'version_string').

The primary reason that `Enzo-E` now saves this version number is to make it easier to interpret the data, when the interpretation changes in newer versions.

At present, the version string only affects one thing: the distinction between particle masses and densities.
- Several months back `Enzo-E` started to save particle masses such that they should be interpreted as masses (previously they were usually treated as densities).
- At the time, to differentiate the interpretation of this data from outputs written by older code versions, we started to have `Enzo-E` append a parameter called `"mass_is_mass"` to the output copy of the parameter file (PR #3914 started querying of this parameter).
- Unfortunately, that approach introduced problems on certain platforms. Thus, we decided to record the version number as an attribute of output hdf5 files that could be used to make this distinction going forward (this will hopefully make any future changes in interpretations easier to handle as well)

Let me know if you need me to add any new tests for this PR.

### DIFF
--- a/yt/frontends/enzo_e/data_structures.py
+++ b/yt/frontends/enzo_e/data_structures.py
@@ -399,10 +399,10 @@ class EnzoEDataset(Dataset):
         fh = h5py.File(os.path.join(self.directory, fn0), "r")
         self.domain_left_edge = fh.attrs["lower"]
         self.domain_right_edge = fh.attrs["upper"]
-        if 'version' in fh.attrs:
-            version = fh.attrs.get('version').tobytes().decode('ascii')
+        if "version" in fh.attrs:
+            version = fh.attrs.get("version").tobytes().decode("ascii")
         else:
-            version = None # earliest recorded version is '0.9.0'
+            version = None  # earliest recorded version is '0.9.0'
         self.parameters["version"] = version
 
         # all blocks are the same size
--- a/yt/frontends/enzo_e/data_structures.py
+++ b/yt/frontends/enzo_e/data_structures.py
@@ -399,6 +399,11 @@ class EnzoEDataset(Dataset):
         fh = h5py.File(os.path.join(self.directory, fn0), "r")
         self.domain_left_edge = fh.attrs["lower"]
         self.domain_right_edge = fh.attrs["upper"]
+        if 'version' in fh.attrs:
+            version = fh.attrs.get('version').tobytes().decode('ascii')
+        else:
+            version = None # earliest recorded version is '0.9.0'
+        self.parameters["version_string"] = version
 
         # all blocks are the same size
         ablock = fh[list(fh.keys())[0]]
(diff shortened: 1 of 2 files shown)


yt/frontends/enzo_e/data_structures.py
```diff
@@ -399,6 +399,11 @@ def _parse_parameter_file(self):
         fh = h5py.File(os.path.join(self.directory, fn0), "r")
         self.domain_left_edge = fh.attrs["lower"]
         self.domain_right_edge = fh.attrs["upper"]
+        if 'version' in fh.attrs:
+            version = fh.attrs.get('version').tobytes().decode('ascii')
+        else:
+            version = None # earliest recorded version is '0.9.0'
+        self.parameters["version_string"] = version
```
neutrinoceros, line 406: mapping a key ending with `"_string"` to anything other than `str` (e.g. `None`) seems a poor choice. I'd suggest either changing the fall back value to something like `"unknown"`, or renaming the key to `"version"` to avoid suggesting a fixed type

yt/frontends/enzo_e/io.py
```diff
@@ -19,11 +19,16 @@ def __init__(self, *args, **kwargs):
...
+        if self.ds.parameters['version_string'] is not None:
+            # they're masses for enzo-e versions that record a version string
+            mass_flag = True
+        else:
+            # in earlier versions: query the existence of the "mass_is_mass"
+            # particle parameter
+            mass_flag = nested_dict_get(
+                self.ds.parameters, ("Particle", "mass_is_mass"), default=None
+            )
         self._particle_mass_is_mass = mass_flag is not None
```
neutrinoceros, line 32: I know you haven't touched this line here but would it make more sense as follows ?
```suggestion
        self._particle_mass_is_mass = bool(mass_flag)
```

neutrinoceros (commented):
Thanks for your contribution ! I have one comment that I would like to see addressed, and a question that may not be relevant. Otherwise, this looks good to me !

yt/frontends/enzo_e/data_structures.py
mabruzzo: Good point. I fixed that

yt/frontends/enzo_e/io.py
mabruzzo: The way it currently reads just checks for the existence of the parameter. For better or worse, the way that parameter historically got set was unsound and could set the parameter to a random value. This includes a value of zero. So I think we need to keep the current logic.

neutrinoceros: Alright that makes sense to me. If you have a minute to add this information as a comment I think that'd be very valuable :)

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):

neutrinoceros:
Ran into a flaky test, let's run it again
@yt-fido test this please

neutrinoceros:
Same error. Third time's the charm ? @yt-fido test this please

PR #4098, opened 2022-08-26
Correcting deceptive typo in error message.

chummels:
There is a warning that has a typo in it, which is misleading.  I searched the docs and the codebase but couldn't find any reference to `draw_axis` kwarg, until I stumbled upon `draw_axes` instead.  But even after having identified the correct kwarg, I still don't even know what function to pass this kwarg to.  `set_background_color()`? `save()`?

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -489,7 +489,7 @@ class ImagePlotMPL(PlotMPL, ABC):
             # last checked with Matplotlib 3.5
             warnings.warn(
                 f"Previously set background color {self.colorbar_handler.background_color} "
-                "has no effect. Pass `draw_axes=True` if you wish to preserve background color.",
+                "has no effect. Pass `draw_frame=True` if you wish to preserve background color.",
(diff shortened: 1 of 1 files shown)


yt/visualization/base_plot_types.py
```diff
@@ -489,7 +489,7 @@ def _toggle_axes(self, choice, draw_frame=None):
             # last checked with Matplotlib 3.5
             warnings.warn(
                 f"Previously set background color {self.colorbar_handler.background_color} "
-                "has no effect. Pass `draw_axis=True` if you wish to preserve background color.",
+                "has no effect. Pass `draw_axes=True` if you wish to preserve background color.",
```
neutrinoceros, line 492: given the context I think the intention was to adversative `draw_frame`
```suggestion
                "has no effect. Pass `draw_fram=True` if you wish to preserve background color.",
```

neutrinoceros (approved):
thanks for catching this @chummels

chummels:
Glad I could catch it!  And you were right, it was draw_frame.

PR #4105, opened 2022-09-03
DEPR: expire yt.mods deprecation

neutrinoceros:
## PR Summary
yt.mods is deprecated in yt 4.1 and marked for removal in yt 4.2
I want to check wether removing it is actually seamless, so I'm submitting the naive and simple removal patch to CI.
If anything goes wrong, I'll retract the removal promise before yt 4.1 is released.


neutrinoceros:
Looks like nothing breaks, so I'll redirect this to the 4.2 milestone, but keep it as a draft until 4.1 is branched out

neutrinoceros:
since there's no particular difficulty here I'll just add the commit in #4106

PR #4112, opened 2022-09-07
For Boxlib frontend, allow reading of 1D particle data

dpgrote:
## PR Summary

Currently, yt cannot read in 1D particle data using the Boxlib frontend. The target code, WarpX, allows 1D simulations so this capability is needed. The issue traces down to the code in `boxlib/io.py` where the particle data arrays are read in. The code explicitly handles 2D and 3D data, but not 1D (which falls back to the 3D case). This puts erroneous data in the y and z positions and when the selector is used to chose only particles within the domain, the erroneous data is out of the bounds and so all of the particles are rejected.

The fix is to include a test for 1D data. The bit of code was rewritten to use a loop over dimensionality to avoid messy if checks of the dimensionality.

## PR Checklist

No changes are needed in the documentation. The updated code is commented.

### DIFF
--- a/yt/frontends/boxlib/io.py
+++ b/yt/frontends/boxlib/io.py
@@ -146,6 +146,8 @@ class IOHandlerBoxlib(BaseIOHandler):
                             f, pheader.real_type, pheader.num_real * npart
                         )
 
+                        # Allow reading particles in 1, 2, and 3 dimensions,
+                        # setting the appropriate default for unused dimensions.
                         pos = []
                         for idim in [1,2,3]:
                             if g.ds.dimensionality >= idim:
--- a/yt/frontends/boxlib/io.py
+++ b/yt/frontends/boxlib/io.py
@@ -149,11 +149,18 @@ class IOHandlerBoxlib(BaseIOHandler):
                         # Allow reading particles in 1, 2, and 3 dimensions,
                         # setting the appropriate default for unused dimensions.
                         pos = []
-                        for idim in [1,2,3]:
+                        for idim in [1, 2, 3]:
                             if g.ds.dimensionality >= idim:
-                                pos.append(np.asarray(rdata[idim-1 :: pheader.num_real], dtype=np.float64))
+                                pos.append(
+                                    np.asarray(
+                                        rdata[idim - 1 :: pheader.num_real],
+                                        dtype=np.float64,
+                                    )
+                                )
                             else:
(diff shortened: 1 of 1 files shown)


atmyers (dismissed):

neutrinoceros:
pre-commit.ci autofix

yt/frontends/boxlib/io.py
```diff
@@ -145,15 +145,24 @@ def _read_particle_fields(self, chunks, ptf, selector):
...
-                            z = np.ones_like(y)
-                            z *= 0.5 * (g.LeftEdge[2] + g.RightEdge[2])
-                        else:
-                            z = np.asarray(
-                                rdata[2 :: pheader.num_real], dtype=np.float64
-                            )
+
+                        # Allow reading particles in 1, 2, and 3 dimensions,
+                        # setting the appropriate default for unused dimensions.
+                        pos = []
```
neutrinoceros, line 151: Since `pos[0]` is used as a default for missing dimensions, I suggest adding an assertion to check that dimensionality is sane, in case someday it's not and the bug isn't as obvious as possible.
```python
assert self.dimensionality >= 1
```

Another solution would be to use default values that do not depend on the results of the first iteration (my prefered approach)

neutrinoceros (commented):
One comment I would like addressed, otherwise looks good to me. Thank you for your contribution!

yt/frontends/boxlib/io.py
dpgrote: Thanks for your comment. Does my change address your concern?

neutrinoceros: It does, thank you !
Now I just noticed that you're looping on 1, 2, 3 but since you're really using 0-based indices I think the code could be simplified.

dpgrote: Yes, that was intentional, looping over 1, 2, and 3 dimensions. It seemed more clear me what the loop was doing at the expense of slightly less simplified code.

neutrinoceros: Not critical, clarity-in-the-eyes-of-the-author wins here 😄

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):

yt/frontends/boxlib/io.py
dpgrote: Thanks for merging!

PR #4116, opened 2022-09-10
ENH: avoid subclassing builtin dict

neutrinoceros:
### problem
Subclassing builtin types like `dict` is error prone because overriden methods are mostly ignored.
This can lead to very subtle bugs where custom classes don't behave as one would expect

```python
class DoubleDict(dict):
    def __setitem__(self, key, value):
        return super().__setitem__(key, [value] * 2)

d = DoubleDict(a=1)
d["b"] = 2
print(d)
```

outputs
```
{'a': 1, 'b': [2, 2]}
```
We see that the user-defined `__setitem__` method is correctly used when setting a new value, but is ignored at initialization.

### solution
subclass `collections.UserDict` instead, which is intended to avoid these subtle issues

### DIFF
--- a/yt/fields/field_info_container.py
+++ b/yt/fields/field_info_container.py
@@ -1,3 +1,4 @@
+from collections import UserDict
 from collections.abc import Callable
 from numbers import Number as numeric_type
 from typing import Optional, Tuple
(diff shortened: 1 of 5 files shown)


matthewturk (approved):
seems fine

Issue #4102, opened 2022-09-02
Nose test dependency


olebole:
### Bug report

For the Debian package, I received a bug report [Debian#1018675](https://bugs.debian.org/1018675), that yt depends on the nose test framework. Nose is [dead](https://github.com/nose-devs/nose/commit/0f40fa995384afad) and [unmaintained](https://pypi.org/project/nose/#history) since several years, and in Debian we would like to remove nose at some point.

Would it be possible to migrate to [pytest](https://docs.pytest.org) (what I would prefer), or to [nose2](https://docs.nose2.io/) or so?

**Version Information**
  * Operating System: Debian bookworm (upcoming)
  * Python Version: 3.10
  * yt version: 4.0.5

neutrinoceros:
Hi @olebole, this has been on our radar for quite some time and we are well aware that nose is unmaintained.

Migration to pytest is absolutely our plan in the long run, but it’s bound to be a large effort for us since many of our tests are yield-based, and cannot be run with pytest as they exist now.

However I wasn’t personally aware of the existence of nose2. I will look it up to see if it can be switched to in a transparent manner.

neutrinoceros:
It looks like nose2 isn't a drop-in replacement for nose as I hoped, so if even switching to that isn't trivial, I don't think it's worth doing, and our energy is better spent on migrating to pytest directly.
It's a substantial effort and we're currently relying only on volunteer work to perform this migration, so I don't know when we can expect completion, but probably not in the immediate future, and it seems definitely out of reach for existing and upcoming yt releases, I'm sorry about that

neutrinoceros:
I created a repo level project to keep track of the existing effort and things that are left to be done here https://github.com/yt-project/yt/projects/4

PR #4118, opened 2022-09-10
TST: migrate image comparison tests from nose to pytest-mpl (⏰ wait on #4122)

neutrinoceros:
## PR Summary
Partially addresses #4102
 
This is far from done, but I need early iterations with CI to see what problems I encounter.
Right now I'm only aiming at image tests being skipped on nose

approach:
the main difficulty is that our `yield` base image tests are built via loops, while `pytest-mpl` relies on return values of test functions. When `pytest.mark.parametrize` is appropriate, this can be a one-to-one translation, but otherwise I rely on generated code to reproduce existing test (by which I mean that I write many alike plain test functions, just not purely by hand).

TODO:
- [ ] stop using image answer tests (replace them with pytest-mpl)
  - [x] `GenericImageTest`
  - [x] `VRImageComparisonTest`
  - [x] `PhasePlotAttributeTest`
  - [x] `PixelizedParticleProjectionValuesTest`
  - [ ] `PixelizedProjectionValuesTest`
  - [ ] `PlotWindowAttributeTest`
- [ ] create image baseline folder (should probably be a submodule, not sure if a new one is warranted, or if we should just reuse answer-store)
- [ ] add image image comparison-only test job
- [ ] cleanup `tests/report_failed_answers.py`
- [ ] cleanup uses of `filename=...` in mpl_image_compar markers
- [ ] consider refactoring again with `pytest.mark.parametrize`

opened questions:
- How can we handle "big data" image tests, and more generally image tests that use real (large) data ? Currently these tests are run on Jenkins + nose


neutrinoceros:
I was not originally planning to fully replace the "answer tests" job on GHA but I see that after removing all image tests from it, it now runs only a very small set of tests and takes about 20s to run (and 10min to build), so I think I should just clear that out completely and replace the job with a new one 1000% focused on image-comparison tests.

neutrinoceros:
I think I should split this in two parts and separate tests that require real data (running on Jenkins) from the rest (running on GH)

neutrinoceros:
This is probably beyond saving at this point. In any case, #4122 has most of this work in a mergeable state, so I think I should just close this one.

matthewturk:
Can you describe the difference with #4122?

PR #4121, opened 2022-09-13
ENH: public API `render` methods to `PlotContainer` and `FixedResolutionBuffer`

neutrinoceros:
## PR Summary
A simple addition to the public interface would be useful in #4118 to avoid using private methods directly.
Content:
- ENH: add a PlotContainer.render method (public interface for _setup_plots)
- DOC: avoid using private `_setup_plots` method in user docs

For backward compatibility, I avoided simply renaming `_setup_plots`.
Hopefully the intented usage is clear from the docstring and the docs update.

I'm not completely certain that my use of the `@validate_plot` is warranted or justified, and I would particularly appreciate feedback specifically on this.

### DIFF
--- a/doc/source/visualizing/geographic_projections_and_transforms.rst
+++ b/doc/source/visualizing/geographic_projections_and_transforms.rst
@@ -107,12 +107,12 @@ available with matplotlib should be available for customization. Here a
 .. code-block:: python
 
     p.set_mpl_projection("Robinson")
-    p._setup_plots()
+    p.render()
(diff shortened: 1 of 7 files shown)


neutrinoceros:
just noticed that `FixedResolutionBuffer` could also use a `render` method to clarify usage of `fib[item]`, I'll include a patch for this too

neutrinoceros:
yup I actually don't think it's needed.

matthewturk (approved):

Issue #3735, opened 2021-12-28
TST: Migrate to pytest-answer and pytest-mpl


matthewturk:
This is a meta-issue to track a migration to both [pytest-answers](https://github.com/data-exp-lab/pytest-answers) and [pytest-mpl](https://github.com/matplotlib/pytest-mpl) for the appropriate tests.

A first pass will be to convert the `AnswerTestingTest` to use `pytest-answer`, then at a later time extract the image-specific test bases into a set of `pytest-mpl`-utilizing tests.

matthewturk:
In looking over the file `yt/utilities/answer_testing/framework.py` we have a couple specific things we need to update to utilize `pytest-answer`.  There are these top-level classes in `framework.py` we need to think about:

 * `AnswerTestStorage` with two subclasses, `AnswerTestCloudStorage` (likely not used anymore anywhere) and `AnswerTestLocalStorage`.
 * `AnswerTestingTest` with the subclasses:
     * `FieldValuesTest`
     * `AllFieldValuesTest`
     * `ProjectionValuesTest`
     * `PixelizedProjectionValuesTest` and `PixelizedParticleProjectionValuesTest`
     * `GridValuesTest`
     * `VerifySimulationSameTest`
     * `GridHierarchyTest`
     * `ParentageRelationshipsTest`
     * `VRImageComparisonTest`
     * `PlotWindowAttributeTest`
     * `PhasePlotAttributeTest`
     * `GenericArrayTest`
     * `GenericImageTest`
     * `AxialPixelizationTest`

There are also a few functions that we use to string these tests together -- specifically, `small_patch_amr`, `big_patch_amr`, `nbody_answer`, and `sph_answer`.

Looking at these, there are a number of them that are reasonably straightforward to port.  In fact, much of the work done in them is regarding parameters and attributes and creation of objects.

neutrinoceros:
Quick note: while I'm 100% behind this project, I don't think it's reasonable to triage this in a release milestone (4.1). This migration is not user-facing so I don't think it's relevant to make it a blocker to any specific release.

neutrinoceros:
#4122 adresses this for all image tests that don't require actual datasets, and that are run on GHA

chrishavlin:
Thing to be aware of (that I didn't know): enzo uses a bit of yt code in its own answer testing that is not actually tested within yt's test suite, see https://github.com/yt-project/yt/pull/4764 and linked issue on enzo-dev.

PR #4122, opened 2022-09-13
TST: migrate image comparison tests from nose to pytest-mpl (1/2)

neutrinoceros:
## PR Summary
This is a self-contained subset of #4118, targetting only tests that can be run with fake datasets and don't require big data infrastructure. These correspond exactly to the GHA answer-tests job.
partially addressed #3735

TODO:
- [x] check that the answer-tests GHA job is reduced to 0 tests
- [x] polish tests (reduce repetition using `pytest.mark.parametrize` ?)
- [x] make sure all tests have a new baseline (edit: not planning to follow through on this for mesh rendering tests requiring pyembree, because it's really not trivial to install either locally or on CI for me atm)
- [ ] document development workflows (how to update baseline; how to create new image tests)
- [x] build any missing supporting architecture for the newly translated tests (namely, create a separate org-level repo to contain reference image files)


neutrinoceros:
Making progress, merging #4125 and #4127 would help a lot with a couple issues that are tiresome to work around locally.

neutrinoceros:
failures on windows are unrelated (#4128), but I learned something important from the image tests failure:
I was attempting to use pytest-mpl in "hybrid mode" (prioritise hash comparison, then go to image comparison in case of failure), but my baselines (both hashes and images) were generated on my laptop, which gives slightly different results from what is generated in CI; the differences in images are low enough to pass tests but (for good reasons) not differences in *hashes*. So I guess we can't use the hybrid mode (or hash comparison) as it completely negates having a non-zero RMS tolerance.

neutrinoceros:
@yt-fido test this please

neutrinoceros:
Currently experiencing hangs on Jenkins which are specific to this PR. This is the last outputs that are logged before the job hangs.

```
/home/fido/workspace/yt_py38_git/yt/visualization/volume_rendering/render_source.py:85: UserWarning: Failed to switch to embree raytracing engine. The following error was raised:
cannot import name 'mesh_construction' from 'yt.utilities.lib.embree_mesh' (/home/fido/workspace/yt_py38_git/yt/utilities/lib/embree_mesh/__init__.py)
  warnings.warn(
  File "tests/nose_runner.py", line 121, in <module>
    for job in generate_tasks_input():
  File "tests/nose_runner.py", line 83, in generate_tasks_input
    for test in list(tests["other_tests"].keys()):
KeyError: 'other_tests'
```

Assuming I didn't just break the global state of the server somehow, I'm going to try and fix the problems I'm seeing here

neutrinoceros:
one test is currently unstable against matplotlib 3.2, that was only ever tested with mpl 3.5 and 3.6
I'll dig in to see if this can be mitigated or if the test should simply be skipped on older versions

neutrinoceros:
The one thing I originally planned to do but didn't is documentation. I still plan to do it but currently there's nothing to replace (the existing workflow, which requires updating the answer-store, is extremely similar to the one needed after this PR), so it seems fine. In all cases this PR will require follow up to:
- remove answer-tests GHA job (I'm leaving it here, even if it runs 0 tests, because it's _required_ by auto-merge)
- require the replacement job instead
- drop the answer-store submodule

In the mean time, I now consider this stable and ready for review

neutrinoceros:
force-pushed to resolve conflicts

matthewturk:
OK, so I have read this, and I think it has a large number of very-difficult-to-review components.  I see a few options:

- Split the PR into one that removes *no* code and just adds the new image testing *base* functions and one that does the switch
- Check and confirm that we have a) the same number of tests b) the same results if we intentionally break the tests.  (this is my preferred)  For that, you'd have to show that the image comparison tests break if we change the unit representations, or if we add an offset to the field values so that they produce different image buffers, for instance.

neutrinoceros:
> - Split the PR into one that removes no code and just adds the new image testing base functions and one that does the switch

I think that would be my preferred track, but can you clarify what you mean here by "base" functions ?

> Check and confirm that we have a) the same number of tests b) the same results if we intentionally break the tests. (this is my preferred) For that, you'd have to show that the image comparison tests break if we change the unit representations, or if we add an offset to the field values so that they produce different image buffers, for instance.

- I honestly don't know how to do a)
- Introducing errors on purpose to see tests fail can be done from the reviewer standpoint, so what is there for me to do on this step exactly ? Should I provide an explicit "how-to test this branch" here ?

neutrinoceros:
rebased on main with no change to refresh CI. Unless anything bad happens I'll also bump the new test job to Python 3.11

.github/workflows/build-test.yaml
```diff
@@ -92,7 +92,63 @@ jobs:
...
+        python -m pip install cartopy scipy
+
+    - name: Run Image Tests
+      run: |
+        pytest --color=yes --mpl -m mpl_image_compare \
+               --mpl-generate-summary=html \
+               --mpl-results-path=pytest_mpl_results \
+               --mpl-baseline-path=tests/pytest_mpl_baseline
+
+    - name: Generate new image baseline
```
matthewturk, line 132: Just so I understand, this will only run in case of failure, and in that particular case it will upload the artifacts?

conftest.py
```diff
@@ -205,6 +206,16 @@ def pytest_collection_modifyitems(config, items):
             item.add_marker(skip_unit)
 
 
+def pytest_itemcollected(item):
+    # Customize pytest-mpl decorator to add sensible defaults
+
+    mpl_marker = item.get_closest_marker("mpl_image_compare")
```
matthewturk, line 212: Can you explain this?  Why is tolerance 0.5?  What does this do?

nose_unit.cfg
```diff
@@ -6,5 +6,5 @@ nologcapture=1
...
+ignore-files=(test_load_errors.py|test_load_sample.py|test_commons.py|test_ambiguous_fields.py|test_field_access_pytest.py|test_save.py|test_line_annotation_unit.py|test_eps_writer.py|test_registration.py|test_invalid_origin.py|test_outputs_pytest\.py|test_normal_plot_api\.py|test_load_archive\.py|test_stream_particles\.py|test_file_sanitizer\.py|test_version\.py|\test_on_demand_imports\.py|test_set_zlim\.py|test_add_field\.py|test_glue\.py|test_geometries\.py|test_firefly\.py|test_callable_grids\.py|test_external_frontends\.py|test_stream_stretched\.py|test_sanitize_center\.py|test_data_reload\.py|test_image_comp_*\.py|test_line_plots\.py|test_mesh_slices\.py|test_raw_field_slices\.py|test_mesh_render\.py|test_vr_orientation\.py|test_profile_plots\.py|test_image_comp_geo\.py|test_image_comp_2D_plots\.py)
```
matthewturk, line 9: Do you think we'll ever be able to break this into multiple lines? :)

yt/fields/tests/test_field_access.py
```diff
@@ -28,6 +28,6 @@ def test_field_access():
         oap = ProjectionPlot(ds, [1, 1, 1], field)
 
         for plot_object in [s, oas, p, oap, prof, phase]:
-            plot_object._setup_plots()
+            plot_object.render()
```
matthewturk, line 31: Just so I am clear, this PR changes it to use the current API, but the old API is still currently in place, right?  This is just keeping up to date with a deprecation?

yt/utilities/answer_testing/framework.py
```diff
@@ -1040,43 +1022,6 @@ def compare(self, new_result, old_result):
                 )
 
 
-class GenericImageTest(AnswerTestingTest):
```
matthewturk, line 1043: @brittonsmith just wanted to check that this does *not* get used in enzo or grackle, right?

yt/utilities/answer_testing/framework.py
```diff
@@ -1137,7 +1082,7 @@ def requires_answer_testing():
 
 def requires_ds(ds_fn, big_data=False, file_check=False):
     condition = (big_data and not run_big_data) or not can_run_ds(ds_fn, file_check)
-    return skipif(condition, reason="cannot load dataset")
+    return skipif(condition, reason=f"cannot load dataset {ds_fn}")
```
matthewturk, line 1085: big QOL improvement here IMO

matthewturk (approved):
So I'm marking approve, because I think we need to eventually pull the ripcord and go on this.  I think what we need to do is basically convince ourselves we're not missing anything major, and then just make the switch.

Do you have a sense of if this is any faster?

nose_unit.cfg
neutrinoceros: I wish. This one line generates so many conflicts that I have a nickname for it (I call it "the blockchain") ... I think I once tried splitting the argument and found that it wasn't supported. Supposedly it's possible to pass the equivalent command line argument more than once to combine regexes, but AFAIU this doesn't work with static configuration files. So if we replace this line by command line arguments, we may be able to escape conflict hell, but that'd be an entire mini project in itself.

.github/workflows/build-test.yaml
neutrinoceros: Not exactly. Note that steps that do the uploads are conditioned with `always()`, so we try to upload as many artefacts as we produced. The "new baseline" artifact may not exist because it's only generated when needed, but the html report should always be present.

conftest.py
neutrinoceros: Here I'm providing arguments to be passed by default to `@pytest.marker.mpl_image_compare` so they don't have to be repeated.
I went with 0.5, which is [stricter than the default](https://github.com/matplotlib/pytest-mpl#tolerance), because I figured we might want to be extra careful at first since this is a big leap of a transition.

yt/fields/tests/test_field_access.py
neutrinoceros: the `_setup_plots` method isn't actually deprecated, I just think it's best to avoid calling private methods directly from outside a class (which is the whole reason I added a public equivalent in this case).
But yes, it's just a no-op change toward stable API

neutrinoceros:
>  I think what we need to do is basically convince ourselves we're not missing anything major

The only major thing that's missing IMO is dev docs. Because the current process for updating image baselines is _already_ not documented and also extremely similar to the one I settled for here, I figured this could probably wait for a follow up PR, but the more this stays waiting, he harder it'll be for me to write it after the fact. I should be able to prioritise this task this week. What would be easier to review ? another 200 lines within this PR, or a brand new one after this one is merged ?

> Do you have a sense of if this is any faster?

I don't think it is. In fact it currently takes *longer* on GHA

pytest-mpl:
```
==== 143 passed, 30 skipped, 915 deselected, 1 warning in 229.02s (0:03:49) ====
```

nose
```
[146](https://github.com/yt-project/yt/actions/runs/4231076474/jobs/7349149421#step:8:147)
Ran 105 tests in 153.973s
```

I assume the reason is that more tests cases are actually run (on nose, we didn't actually cover cartopy integration and mesh rendering !)

nose_unit.cfg
neutrinoceros: I'm addressing this in #4346

neutrinoceros:
updated following #4346

yt/utilities/answer_testing/framework.py
matthewturk: I don't *think* we do.  But that's why I'm asking!

brittonsmith: Grackle doesn't use either of these classes. Enzo uses `AnswerTestingTest`, but not `GenericImagetest`.

neutrinoceros: `AnswerTestingTest` isn't removed in this PR, so it should be good !

neutrinoceros: @matthewturk any chance we might move forward with this soon ?

matthewturk:
Yes, I actually don't think we need to wait any longer.

matthewturk (approved):
Do we think this is done?  @Xarthisius have you had a chance to look it over?

neutrinoceros:
trying our luck again: ping @Xarthisius

neutrinoceros:
I'd really like to move forward with this one, I consider it blocking for many image-related changes I would like to get going.

Xarthisius:
I'm not sure why I was pinged on this PR. If you're waiting for my review just go ahead.

PR #4129, opened 2022-09-15
TST: unpin h5py and scipy in windows_conda_requirements.txt

### LINKED ISSUE #4128: TST: CI instability (Windows + h5py)
### Bug report

**Bug summary**

Last merge on main failed CI on Windows: https://github.com/yt-project/yt/actions/runs/3056211087/jobs/4935758662

This is the warning that cause a couple dozen tests to fail
```
E       UserWarning: h5py is running against HDF5 1.12.1 when it was built against 1.12.2, this may cause problems
```

H5Py was not updated recently, but we *do* pin it specifically on windows, which may explain the sudden fail.

update: that wasn't the problem (in fact, the way ci is setup, h5py ends up being updated via pip anyway), so we need to understand what changed in the environment.

neutrinoceros:
## PR Summary
should fix #4128

### DIFF
--- a/tests/windows_conda_requirements.txt
+++ b/tests/windows_conda_requirements.txt
@@ -1,6 +1,6 @@
 numpy>=1.19.4
 cython>=0.29.21,<3.0
 cartopy>=0.21.0
-h5py~=3.1.0
+h5py>=3.1.0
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Turns out it doesn’t solve the problem, I’ll inquire this further

PR #4134, opened 2022-09-17
ENH: Enable exporting GAMER and FLASH datasets to octrees

matthewturk:
## PR Summary

Initial value: This allows passing in `domain_dimensions` to the load_octree function, so that you can load octrees that don't have just the one root cell.

New and revised: This PR will enable FLASH and GAMER datasets to be exposed as Octree datasets, which will be a method for testing improvements in performance.

## PR Checklist


yt/loaders.py
```diff
@@ -925,6 +925,7 @@ def load_octree(
     periodicity=(True, True, True),
     over_refine_factor=None,
     num_zones=2,
+    domain_dimensions=None,
```
neutrinoceros, line 928: adding new arguments in between two existing ones for which positional values are accepted is a breaking change, please move it to after the `*,`

jzuhone (approved):

neutrinoceros:
close/open to refresh CI now that unrelated failures are resolved on main

neutrinoceros:
Again refreshing CI

neutrinoceros:
@matthewturk Is this still "WIP" or is it actually ready for merging ?

matthewturk:
So I think it needs another pass from me. But I will check today!

It's currently -8F here with 24MPH winds! Once I get the place under control I'll take a look over this one. Should be a fun task for a funny day.

cphyc:
Best time for a feature freeze!

23 Dec 2022 14:06:10 Matthew Turk ***@***.***>:

matthewturk:
Oh, my. That is a hall of fame review comment.

matthewturk:
So it's ready for *review* but I don't want it to be merged *quite* yet -- a few more minor things I want to address.

yt/loaders.py
```diff
@@ -1015,7 +1019,11 @@ def load_octree(
     # for compatibility
     if over_refine_factor is not None:
         nz = 1 << over_refine_factor
-    domain_dimensions = np.array([nz, nz, nz])
+    if domain_dimensions is None:
+        # We assume that if it isn't specified, it defaults to the number of
+        # zones (i.e., a single root oct.)
+        domain_dimensions = [nz, nz, nz]
+    domain_dimensions = np.array(domain_dimensions)
```
neutrinoceros, line 1026: To reduce the type variability of `domain_dimensions`
```suggestion
    if domain_dimensions is None:
        # We assume that if it isn't specified, it defaults to the number of
        # zones (i.e., a single root oct.)
        domain_dimensions = np.array([nz, nz, nz])
    else:
        domain_dimensions = np.array(domain_dimensions)
```

yt/loaders.py
```diff
@@ -980,6 +981,9 @@ def load_octree(
     parameters: dictionary, optional
         Optional dictionary used to populate the dataset parameters, useful
         for storing dataset metadata.
+    domain_dimensions : array_like, optional
```
neutrinoceros, line 984:
```suggestion
    domain_dimensions : 3 elements array-like, optional
```

yt/geometry/oct_container.pyx
```diff
@@ -185,6 +185,16 @@ cdef class OctreeContainer:
...
 
+    def check_root_mesh(self):
+        cdef count = 0
+        for i in range(self.nn[0]):
+            for j in range(self.nn[1]):
+                for k in range(self.nn[2]):
+                    if self.root_mesh[i][j][k] == NULL:
+                        print("Missing ", i, j, k)
+                        count += 1
+        print("Missing total of %s out of %s" % (count, self.nn[0] * self.nn[1] * self.nn[2]))
```
neutrinoceros, line 196: doesn't look like you're using this method anywhere, is it debugging code ? I'd suggesting dropping it before we can merge this

neutrinoceros (commented):
Here are a couple remarks

neutrinoceros:
looks like this PR grew a couple minor conflicts. I'm also waiting for my comments to be adressed, but it seems like a low-hanging fruit on the journey to yt 4.2

yt/geometry/oct_container.pyx
matthewturk: I don't particularly want to drop this, as it would just need to be reimplemented the next time we need to debug something.  I am OK with making it semi-private.

yt/loaders.py
matthewturk: I don't like this change, to be honest.  Both seem clear, and I think my version is fine, as the pass-through is *right* there -- the distance is the same.

matthewturk: That's pretty irritating.

yt/geometry/oct_container.pyx
```diff
@@ -581,6 +595,12 @@ cdef class OctreeContainer:
         cdef int in_boundary = 0
         # How do we bootstrap ourselves?
         for p in range(no):
+            # We allow specifying curlevel = -1 to query from the levels array
+            # instead.
+            if curlevel == -1:
+                this_level = levels[p]
+            else:
+                this_level = curlevel
```
neutrinoceros, line 603: this doesn't depend on the iteration variable (`p`), so it would be clearer (and more efficient) to put it outside the loop

yt/geometry/oct_container.pyx
```diff
@@ -170,7 +170,7 @@ cdef class OctreeContainer:
                 pos[2] = self.DLE[2] + dds[2]/2.0
                 for k in range(self.nn[2]):
                     if self.root_mesh[i][j][k] == NULL:
-                        raise RuntimeError
+                        raise KeyError(i,j,k)
```
neutrinoceros, line 173: I guess an `IndexError` would be slightly more appropriate

matthewturk: No, I think `KeyError` is fine.  The issue is that we have constructed what amounts to a mapping, and that mapping has nothing in it for a given key.

yt/geometry/oct_container.pyx
matthewturk: `levels[p]`

yt/geometry/oct_container.pyx
neutrinoceros: ok, makes sense

yt/geometry/oct_container.pyx
neutrinoceros: 🤦🏻‍♂️

neutrinoceros (approved):

PR #4138, opened 2022-09-21
BLD: remove unused extra dependencies

neutrinoceros:
## PR Summary

This contains updates in optional requirements mainly targetted at saving build time on CI


neutrinoceros:
Actually we don't currently use coverage at all (in CI), so maybe it'd be simpler to just remove it from the `[test]` extra dep list

cphyc (approved):

PR #4139, opened 2022-09-22
TYP: upgrade mypy, handle newly detected problems

neutrinoceros:
## PR Summary

Upgrading mypy to hopefully resolve issues seen in #3421

- TYP: upgrade mypy (0.910 -> 0.971)
- TYP: ignore mypy warnings in ProjectionPlot and SlicePlot __new__ methods
- TYP: fix typing errors in gadget_fof/fields.py


neutrinoceros:
A nice side effect of this PR is that it diminishes the run time of the type check job because there are wheels for mypy 0.971 + Python 3.10, so we don't have to compile mypy on each job

matthewturk (approved):

PR #4140, opened 2022-09-22
TST: set an upper bound for numpy version in type check env

neutrinoceros:
## PR Summary

Currently type checking is unstable on main, see https://github.com/yt-project/yt/actions/runs/3107473459/jobs/5035621544
The error seems to be triggered because we request mypy to validate Python 3.7 compatibility but the latest version of numpy doesn't have that constraint. It's not clear to me why this hasn't happened before, but this patch should fix it.


matthewturk:
@neutrinoceros how long do you anticipate this being in?  And, are you sure we should ship the release with this?

neutrinoceros:
Maybe it's not okay to ship it actually... I don't know that anyone downstream relies on installing yt with the `[typecheck]` extras, but this would probably break them.

A less invasive solution would be to switch the type-checking env to use Python 3.7 instead. Your call :)

matthewturk:
let's try the less invasive solution

neutrinoceros:
@matthewturk done !

matthewturk (approved):

PR #4142, opened 2022-09-24
ENH: add API to switch raytracing engine at runtime

neutrinoceros:
## PR Summary

This allows to properly define pytest fixtures for switching between "yt" and "embree" ray tracing engine options at runtime (useful in #4122)

Also, a follow up to #3626


matthewturk (dismissed):
Nice!

neutrinoceros:
@matthewturk I just realised I don't need the boolean flag, give me a minute to remove it and then you can set this to auto-merge if you'd like

matthewturk (approved):

neutrinoceros:
Sorry about the additional iterations, I think it's good now.

neutrinoceros:
sigh... actually I do need it, otherwise my fixtures just crash pytest

neutrinoceros:
Nevermind. This is usable as is, and validated via #4122

neutrinoceros:
A problem I failed to see originally here is that `"embree"` is currently the default ray tracing engine. In the current state, this means yt will raise a warning at startup time unless one of the following conditions is met:
- embree is correctly installed
- the parameter is override from the config file

This isn't acceptable.
So here are the options I can think of:

1) Change the default engine to `"yt"`. This has the downside of breaking users relying on being able to use embree with no additional configuration (other than installing it).
2) reintroduce a very boolean flag so that we can filter the warning, and use it specifically at initialisation (cost in code complexity)
3) _never_ warn the `set_raytracing_engine("embree")` fails

None of these are ideal. I think 1) is the correct thing to do, provided it the breaking change clearly mentioned in release notes. Hopefully this feels acceptable not just to me, I'll let reviewers discuss this point.

neutrinoceros:
I've now implemented my preferred solution (1)

matthewturk (approved):

PR #4143, opened 2022-09-24
Updating Firefly-dependent areas to match API changes and PyPI packaging

mtryan83:
## PR Summary

This PR aims to reconcile the changes to Firefly introduced since version 4.0.0 of yt was released. In summary, Firefly tweaked the `Reader` and `ParticleGroup` APIs and the installation of Firefly from PyPI installs as `firefly` (instead of `Firefly` or `Firefly-vis`, the devs were able to get control of the package name).  I've updated the documentation, including that (I believe) Firefly also now prefers their binary `.ffly` format.

I'm very new to Firefly and yt, so I would greatly appreciate it if e.g. @agurvich would double check.

I believe this counts as a fix/update for #3415 

## PR Checklist


mtryan83:
pre-commit.ci autofix

setup.cfg
```diff
@@ -81,7 +81,7 @@ full =
     astropy>=4.0.1,<6.0.0
     f90nml>=1.1.2
     fastcache>=1.0.2
-    firefly-vis>=2.0.4,<3.0.0
+    firefly>=3.0.0
```
neutrinoceros, line 84: I would like to keep this as bullet-proof as possible, so that yt+firefly keeps working in CI even when API changes upstream. AFAIC I'm very happy with the current workflow where this kind of PR helps us migrate to newer API, but we do not depend.
```suggestion
    firefly>=3.2.0,<4.0.0
```

yt/data_objects/data_containers.py
```diff
@@ -703,7 +703,7 @@ def to_glue(self, fields, label="yt", data_collection=None):
 
     def create_firefly_object(
         self,
-        JSONdir,
+        datadir,
```
neutrinoceros, line 706: this particular change is not 100% backward compatible for yt users, we would also need to capture `JSONdir` being passed as a keyword argument and raise a deprecation warning when it's caugh (use `yt._maintainance.deprecation.issue_deprecation_warning`).
This also implies adding a default value to `datadir` (`None`)... well it gets messy pretty quickly. Let me know if you need more detailed instructions. I can also jump in and do this bit myself if you'd like.

yt/utilities/on_demand_imports.py
```diff
@@ -455,13 +455,13 @@ def concat(self):
 class Firefly_imports(OnDemand):
```
neutrinoceros, line 455: better to keep the name of the class in sync too
```suggestion
class firefly_imports(OnDemand):
```

neutrinoceros (changes requested):
Hi @mtryan83, thanks a lot for doing this !
Indeed, it'd be great if @agurvich could sign off here.
In the mean time, here are a couple comments, one of which is crucial (API change)

yt/utilities/on_demand_imports.py
neutrinoceros: (note that you'll need to update line 469 as well)

yt/data_objects/data_containers.py
mtryan83: I think 12c54cc covers this? I followed the approach from yt/data_objects/image_array.py for the notation. I put in 4.2 as the removal version, but that was just a guess.

neutrinoceros: Almost there. I'll add further comments.

yt/data_objects/data_containers.py
```diff
@@ -703,13 +704,14 @@ def to_glue(self, fields, label="yt", data_collection=None):
...
         self,
-        JSONdir,
+        datadir=None,
         fields_to_include=None,
         fields_units=None,
         default_decimation_factor=100,
         velocity_units="km/s",
         coordinate_units="kpc",
         show_unused_fields=0,
+        JSONdir=None,
```
neutrinoceros, line 714: we don't want this deprecated parameter to be valid if passed as _positional_, so let's mark it as keyword only
```suggestion
        *,
        JSONdir=None,
```

yt/data_objects/data_containers.py
```diff
@@ -793,9 +795,18 @@ def create_firefly_object(
...
 
+        if JSONdir is not None:
+            issue_deprecation_warning(
+                "The 'JSONdir' keyword argument is a deprecated alias for 'datadir'."
+                "Please use 'datadir' directly.",
+                since="4.1",
+                removal="4.2",
+            )
+            datadir = JSONdir
+
```
neutrinoceros, line 806: an error should be raised (`TypeError`) if `datadir` is still `None` at this point.

yt/data_objects/data_containers.py
```diff
@@ -793,9 +795,18 @@ def create_firefly_object(
         ## for safety, in case someone passes a float just cast it
         default_decimation_factor = int(default_decimation_factor)
 
+        if JSONdir is not None:
+            issue_deprecation_warning(
+                "The 'JSONdir' keyword argument is a deprecated alias for 'datadir'."
+                "Please use 'datadir' directly.",
+                since="4.1",
+                removal="4.2",
```
neutrinoceros, line 803: no need to specify a removal version, we'll try to keep this for a couple minor versions at least, and possibly longer if it's not a burden to maintain
```suggestion
```

neutrinoceros (commented):
follow up comments on the deprecated API part

yt/data_objects/data_containers.py
mtryan83: `None` is an allowed (and is the default) value for `datadir` though. From the Firefly doc/code, a value of `none` defaults to `$HOME/<file_prefix>` (and `file_prefix` defaults to `Data`).

yt/data_objects/data_containers.py
```diff
@@ -719,7 +721,7 @@ def create_firefly_object(
         Parameters
         ----------
 
-        JSONdir : string
+        datadir : string
             Path to where any `.json` files should be saved. If a relative
             path will assume relative to `${HOME}`
```
mtryan83, line 727:
```suggestion
            path will assume relative to `${HOME}`. A value of `None` will default to `${HOME}/Data`.
```

yt/data_objects/data_containers.py
neutrinoceros: Yes that would make sense indeed ! While we're at it, we should probably also document where additional `**kwargs` are being passed to, since that's also supported.

yt/data_objects/data_containers.py
```diff
@@ -753,6 +753,10 @@ def create_firefly_object(
             A flag to optionally print the fields that are available, in the
             dataset but were not explicitly requested to be tracked.
 
+        kwargs : dict of items
+            Any additional values are passed as parameters in the
+            firefly.data_reader.Reader initialization.
```
neutrinoceros, line 758:
```suggestion
        Any additional keyword arguments are passed to
        firefly.data_reader.Reader.__init__
```

neutrinoceros (dismissed):

agurvich:
So, I am only just now tuning in. I am traveling until next Monday 10/3 and won't have a chance to formally review this until then. @mtryan83, first of all, thank you for putting all of this together! from your original PR summary this seems like exactly the sort of changes that I've had on a to-do list to do since we updated the API and submitted the Firefly paper so I am in your debt for handling it. 

It's been hard to follow the updates to the PR since then, has it all mostly been yt documentation stuff? Is there something in particular I should pay close attention to (it sounds like you covered all the bookkeeping stuff like switching JSONDIR and `.ffly` format, etc.). I can have a look at the file diffs but it may be worth coordinating and changing things upstream if there's something worth cleaning up there (e.g. if it means removing a warning on yt's side because I just do better type checking on firefly's side).

neutrinoceros:
>So, I am only just now tuning in. I am traveling until next Monday 10/3 and won't have a chance to formally review this until then.

No rush on our side !

> it may be worth coordinating and changing things upstream if there's something worth cleaning up there (e.g. if it means removing a warning on yt's side because I just do better type checking on firefly's side).

AFAICT there's nothing we'd want to clean up upstream.

mtryan83:
I think the main thing is just making sure I updated the API correctly.  From what I could tell, it was just `JSONdir` --> `datadir`, `tracked_*` --> `field_*`, `Firefly` --> `firefly`, and `dumpToJSON` --> `writeToDisk` and `options` --> `settings` (though those two only appear in the documentation).

I don't see anything to clean up upstream either.

Looks like I missed a JSONdir though, @neutrinoceros, sorry!

neutrinoceros:
For later reference, I previously triaged this for the 4.1 milestone, but I'm going to take it back, because that release is already far behind schedule. I would still be happy to include the patch in the release if we can get a second review in time, which seems likely, I just want to signal that the patch shouldn't be considered blocking.

mtryan83:
(fast-cache and firefly dependencies were _touching_! Couldn't automerge anymore)

agurvich (approved):
These look great to me, thank you @mtryan83 for cleaning all this up with the new API. I'm hoping that I won't be compelled to change it again any time soon now that we have the .ffly format (which is what inspired most of the changes).

neutrinoceros (approved):

PR #4145, opened 2022-09-30
BUG: Use any() on numpy array in flash blockless check

### LINKED ISSUE #4109: BUG: incorrect conversion from array to boolean
### Bug report

**Bug summary**
This line
https://github.com/yt-project/yt/blob/2f4766957a2e88461cf1473bb5873335e80b529d/yt/frontends/flash/io.py#L164

triggers a bunch of `DeprecationWarning`s in testing
```
/home/fido/workspace/yt_py38_git/yt/frontends/flash/io.py:164: DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
```

It is not directly obvious to me how it should be fixed, since I am not familiar with this region. I'll assign @matthewturk who last edited this line.

matthewturk:
## PR Summary

Closes #4109

### DIFF
--- a/yt/frontends/flash/io.py
+++ b/yt/frontends/flash/io.py
@@ -145,7 +145,7 @@ class IOHandlerFLASH(BaseIOHandler):
                 if _xyz[0].size > 0:
                     bxyz.append(_xyz)
                 mask = selector.select_points(x, y, z, 0.0)
-                blockless = _xyz[0] > 0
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Thanks ! I'll wait for Jenkins CI to complete and check if the deprecation warnings are all gone

PR #4146, opened 2022-10-02
RFC: upgrade Cython 'language level' to Python 3

neutrinoceros:
## PR Summary

This should be a noop, and gets us more in line with Cython 3.0 upcoming standards. See https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html?highlight=language_level#python-3-syntax-semantics

### DIFF
--- a/yt/geometry/_selection_routines/selector_object.pxi
+++ b/yt/geometry/_selection_routines/selector_object.pxi
@@ -1,3 +1,6 @@
+from .oct_visitors cimport CountTotalCells, CountTotalOcts
+
+
 cdef class SelectorObject:
 
     def __cinit__(self, dobj, *args):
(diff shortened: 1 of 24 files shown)


neutrinoceros:
Ouch. Clearly this is *much* more fragile than I thought. I may come back and try to fix it in the coming days, or I'll close it if I don't have time.

neutrinoceros:
Thanks to @cphyc's help, this is now ready for review.
The one bit I was worried about is that using `language_level=3` changes the meaning of the `/` operator for integers (from integer division to float division), and I was concerned that we might miss some updates and create bugs, but the good news is that _we can't_, because `x / y` fails at compile time if `x` is an integer.

cphyc:
So to be precise, `z = x / y` fails at compilation is `z` is of integer type.

matthewturk:
That's interesting.

neutrinoceros:
I note that using `//` in Cython may still have a performance hit because it calls Python API. We should check that it's not used in any hot loop

cphyc:
I don't think this is in any hot loop at the moment, but if this worries you we can always wrap stuff with `@cython.cdivision(True)`.

neutrinoceros:
wow, so many conflicts from the Cython-lint PRs. I'll make this a draft until I can make sure tests still pass

neutrinoceros:
@matthewturk do you want to review this ?

yt/geometry/_selection_routines/selector_object.pxi
```diff
@@ -1,3 +1,6 @@
+from .oct_visitors cimport CountTotalCells, CountTotalOcts
```
matthewturk, line 1: Is this because we don't implicitly import from pxd files now?

matthewturk (commented):
lgtm; lots of rearranging which took a moment to read, but the int division changes all seem reasonable and this is an improvement.

matthewturk (approved):

yt/geometry/_selection_routines/selector_object.pxi
neutrinoceros: TBH I'm not 100% sure I understand how this file used to compile, but this broke only after I resolved the conflicts from #4195 , so I _assume_ that the dependency graph worked because of misplaced import statements that happened to compensate their absence here.

Issue #4054, opened 2022-08-04
Rich display is broken on recent version of ipython


cphyc:
### Bug report

**Bug summary**

On IPython ≥8, yt fields are displayed as if we were in a rich display interface (but we're not!). See screenshot.
![image](https://user-images.githubusercontent.com/5411875/182955615-8b16bf54-f051-4086-82f1-0f33217bb1f8.png)

**Code for reproduction**

In an IPython shell:
```python
>>> from yt.testing import fake_amr_ds
...
... ds = fake_amr_ds()
... ds.fields
```

**Actual outcome**

```
yt : [INFO     ] 2022-08-04 22:27:58,875 Parameters: current_time              = 0.0
yt : [INFO     ] 2022-08-04 22:27:58,875 Parameters: domain_dimensions         = [32 32 32]
yt : [INFO     ] 2022-08-04 22:27:58,875 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2022-08-04 22:27:58,875 Parameters: domain_right_edge         = [1. 1. 1.]
yt : [INFO     ] 2022-08-04 22:27:58,875 Parameters: cosmological_simulation   = 0
Out[6]: HBox(children=(Select(layout=Layout(height='95%'), options=('cell_volume', 'dx', 'dy', 'dz', 'path_element_x',…
HBox(children=(Select(layout=Layout(height='95%'), options=('cell_volume', 'cylindrical_radius', 'cylindrical_…
HBox(children=(Select(layout=Layout(height='95%'), options=('Density', 'cell_volume', 'dx', 'dy', 'dz', 'path_…
Tab(children=(Output(), Output(), Output()), _titles={'0': 'gas', '1': 'index', '2': 'stream'})

```

**Expected outcome**

```
yt : [INFO     ] 2022-08-04 22:28:29,360 Parameters: current_time              = 0.0
yt : [INFO     ] 2022-08-04 22:28:29,361 Parameters: domain_dimensions         = [32 32 32]
yt : [INFO     ] 2022-08-04 22:28:29,361 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2022-08-04 22:28:29,361 Parameters: domain_right_edge         = [1. 1. 1.]
yt : [INFO     ] 2022-08-04 22:28:29,362 Parameters: cosmological_simulation   = 0
Out[1]: <yt.fields.field_type_container.FieldTypeContainer object at 0x7fc037a9ef40>
```

[...]

PR #4155, opened 2022-10-08
BLD: promote ipywidget to a hard dependency

### LINKED ISSUE #4154: BUG: __repr__ crashes in IPython if ipywidget isn't installed
```shell
ipython -c "import yt; ds=yt.load_sample('ChollaSimple'); ds.fields"
```
crashes with
```
ModuleNotFoundError: No module named 'ipywidgets'
```

Using the default `python` interpreter doesn't cause a crash

Note that I'm using the `ChollaSimple` sample dataset here only because it's small, but this bug isn't specific to the cholla frontend.

_Originally posted by @neutrinoceros in https://github.com/yt-project/yt/issues/4054#issuecomment-1206513964_

neutrinoceros:
## PR Summary
Closes #4154

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,6 +39,7 @@ project_urls =
 packages = find:
 install_requires =
     cmyt>=0.2.2
+    ipywidgets>=8.0.0
     matplotlib!=3.4.2,>=3.1  # keep in sync with tests/windows_conda_requirements.txt
     more-itertools>=8.4
     numpy>=1.14.5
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Note that this leaves #4054 unresolved. The core issue is that we do not distinguish between running within IPython and Jupyter Notebook/Lab
ATM I'm not sure there's a supported way to make this distinction at run time.

neutrinoceros:
Turns out @cphyc 's previous inquiry already contains the solution to the core problem https://github.com/yt-project/yt/issues/4054#issuecomment-1206232539

I think this PR may be a correct first step towards resolution of #4054 too, as long as we require ipywidget>=8.0.0

neutrinoceros:
Actually, it may not even be necessary to require ipywidget since `_repr_mimebundle_` actually *returns* a value, that we can customise at runtime depending on availability.

cphyc (approved):

PR #4156, opened 2022-10-10
RAMSES: catch assertion errors with ill-written namelist files

cphyc:
## PR Summary

RAMSES stores along with its output the namelist containing the parameters of the simulation. Unfortunately, with old version of RAMSES, this file may be in binary format (this was a bug that has now been fixed). This then may lead to `f90nml`, the package we rely on to parse the namelist, to raise an `AssertionError`.

This PR catches it.

See also: https://github.com/marshallward/f90nml/issues/147


neutrinoceros:
Cool, I think we can backport this to 4.1.x

neutrinoceros (approved):

PR #4164, opened 2022-10-12
[BUGFIX] Memory leak: free C++ memory before deallocing Python object

cphyc:
## PR Summary

This fixes a memory leak issue reported in https://mail.python.org/archives/list/yt-users@python.org/thread/H3WCK5PA2WNGEDBVHOCWV5YCPV2ZYIXY/.

The root cause of the issue is that, for octree datasets (RAMSES, ART), the base grid is stored using a tree-like structure that is allocated in C. Unfortunately, when associated Python gets deleted, we were not freeing the memory in the process.

This PR fixes this by explicitely removing all entries in the query tree updon deallocation of the Python object.

## Notes

For _very_ large datasets, the operation may be quite slow `O(N log(N))`, where `N` is the number of base grid elements. For a RAMSES dataset with `levelmin=8`, that would be `~3 × 8 × (2^8)^3` operations.
It should in principle be possible to deallocate everything in `3 × N` operations by beeing clever. Alternatively, a longer term solution would be to move away from the manually-implemented `tsearch` function and use e.g. a `C++`-backed regular hashmap.

### DIFF
--- a/yt/geometry/oct_container.pyx
+++ b/yt/geometry/oct_container.pyx
@@ -1073,6 +1073,11 @@ cdef class SparseOctreeContainer(OctreeContainer):
     def __dealloc__(self):
         # This gets called BEFORE the superclass deallocation.  But, both get
         # called.
+        cdef OctKey *ikey
+        for i in range(self.num_root):
+            ikey = &self.root_nodes[i]
+            tdelete(<void *>ikey, &self.tree_root, root_node_compare)
+
         if self.root_nodes != NULL: free(self.root_nodes)
 
 cdef class ARTOctreeContainer(OctreeContainer):


matthewturk:
The only reason, if I remember correctly, that I used `tsearch` instead of something (better) like C++ hashmap was to try to avoid any additional dependencies and remain "pure C."  I am not sure those are concerns anymore!

yt/geometry/oct_container.pyx
```diff
@@ -1073,6 +1073,11 @@ cdef class SparseOctreeContainer(OctreeContainer):
     def __dealloc__(self):
         # This gets called BEFORE the superclass deallocation.  But, both get
         # called.
+        cdef OctKey *ikey
+        for i in range(self.num_root):
+            ikey = &self.root_nodes[i]
+            tdelete(<void *>ikey, &self.tree_root, root_node_compare)
```
neutrinoceros, line 1079: I'm struggling to find documentation about this. Can you show me where `tdelete` comes from ?

matthewturk: It shows up in `man tdelete` and is part of `search.h` https://pubs.opengroup.org/onlinepubs/009695399/functions/tdelete.html

neutrinoceros: thanks ! I don't feel competent to review this from the technical angle, but I *can* check that this resolves the issue. Will report back as soon as I'm able to.

neutrinoceros (approved):
I confirm that this fixes the memory leak, thank you !
I'll refrain from merging right away in case @matthewturk wants to make a more technical review

yt/geometry/oct_container.pyx
cphyc: Soooo, actually, it seems to be shadowed by our own version of [tdelete](https://github.com/cphyc/yt/blob/37ad7b5544b5a4a62ffd29ee57c6c81c86ccee2d/yt/utilities/lib/tsearch.c#L75).

cphyc:
I guess the only question on the table is whether we want to roll with this fix or remove `tsearch/tdelete` altogether?

neutrinoceros:
AFAIC I'd favour the current, minimal fix, since we should backport it and it's preferable to keep backports's footprints as small as possible.

yt/geometry/oct_container.pyx
matthewturk: Oh good!  I forgot we had this.

Issue #4162, opened 2022-10-11
DEP: upcoming incompatibilities with unyt 3.0


neutrinoceros:
### Context

There's a long lived project to implement NEP 18 in unyt, and implementation has just begun (first merge was performed yesterday).
In short, the idea is that builtin numpy functions should be able to preserve and correctly treat units given `unyt_array` inputs.

We reckon that this _cannot_ be performed in a strictly backward-compatible fashion, because of existing code _relying_ on numpy functions to drop units. This is the reason why the scheduled bleeding-edge job on yt failed today:

https://github.com/yt-project/yt/actions/runs/3223984952/jobs/5274634351

The discussion started upstream at https://github.com/yt-project/unyt/issues/291
So, going forward, this (breaking) change is the reason why next release of unyt will be 3.0 (currently it's 2.9.2).

On yt's side, short term, I think our best option is to set an upper limit on what version of unyt we accept (`<3.0`) for the time being. This does not prevent us from fixing issues as we go, but we should probably not aim at full compatibility with unyt 3.0 until NEP 18 is fully implemented (see https://github.com/yt-project/unyt/issues/289)

I'll add this issue to the 4.1.1 milestone until the upper boundary is set, but I expect it should only be closed when full compatibility is achieved (I'm assuming, in time for yt 4.2), and we can relieve this temporary limit.

update: here's the list of PRs that implement part of migration process to unyt 3.0
- #4163 
- #4166

Issue #4136, opened 2022-09-20
ENH: Use GAMER "sibling" information for ghost zones


matthewturk:
GAMER includes a `/Tree/Sibling` dataset which can be used to identify all the neighbors.  We should use this for ghost zone generation.

PR #4166, opened 2022-10-13
BUG: fix future incompatiblities with unyt 3.0 (2/n)

neutrinoceros:
## PR Summary

Follow up to #4136 and https://github.com/yt-project/unyt/pull/293, this adapts yt to stay compatible with latest development in unyt (#4162).

Here's my reasoning:
- some of unyt's API is now marked as deprecated (on the dev branch). Specifically, all unit-preserving numpy wrapper functions (e.g. `unyt.udot` and `unyt.uconcatenate`)
- we cannot yet migrate to the vanilla numpy API (e.g., `np.dot` and `np.concatenate`) that _will_ be the replacement because that will not work _before_ unyt 3.0 is available.
- both the `yt` and `yt.units` namespace currently re-expose this unyt API directly

The simpler option I can see to get around the deprecation warnings in CI *without* simply filtering them is to
vendor specifically this part of unyt's API, but do so in an isolated and hidden module (`yt.units._numpy_wrapper_functions`), and use that internally. This module is meant to be deprecated itself at some point in the future, but we'll be able to control that on yt's own development cycle, instead of being tightly coupled to unyt's.

I ran the bleeding-edge workflow on my fork to show that this branch fixes the current instabilities: https://github.com/neutrinoceros/yt/actions/runs/3241090204

I'm labelling this as 2/n because there will be more changes upstream soon ( we're making numpy API unit-aware), but this is by far the largest PR of this series. Upcoming PRs should be much smaller in size and look more like the second commit of this branch.

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -33,8 +33,9 @@ from yt.funcs import (
 from yt.geometry import particle_deposit as particle_deposit
 from yt.geometry.coordinates.cartesian_coordinates import all_data
 from yt.loaders import load_uniform_grid
+from yt.units._numpy_wrapper_functions import uconcatenate
 from yt.units.unit_object import Unit  # type: ignore
-from yt.units.yt_array import YTArray, uconcatenate  # type: ignore
+from yt.units.yt_array import YTArray
 from yt.utilities.exceptions import (
     YTNoAPIKey,
     YTNotInsideNotebook,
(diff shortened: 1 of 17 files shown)


neutrinoceros:
@jzuhone can you have a look please ?

jzuhone:
@neutrinoceros Can you just leave a bit more explanation in the description as to what this is doing and why? I think I get it but for future reference we should probably have that information.

neutrinoceros:
@jzuhone better ?

jzuhone (approved):

PR #4170, opened 2022-10-14
basic datetime functionality for stream frontend, annotate_timestamp

chrishavlin:
This PR adds some very basic (but I think still useful) datetime functionality that let's you:

* set the `sim_time` parameter in the stream frontend with a `np.datetime64` value (which gets cast to float when setting `ds.current_time`)
* display the `current_time` as a datetime if that's what it was originally set as 
* format the time as a datetime for the `annotate_timestamp` plot callback (this one is not restricted to stream frontends)

Given the potential complexity of adding actual datetime handling internally (which I think would require some unyt work as well), this PR simply adds an ingestion step on dataset initialization to convert any datetime to a float and then adds some simple formatting when displaying the `ds.current_time`. I added it just to the stream frontend because that seemed the most likely use-case (though there may be a benefit to the nc4_cm1 and cf_radial frontends, so I'd be open to adding some of this logic at the base dataset level instead). 

No new tests yet cause I wanted to get a sense for whether or not this is a good idea before putting more time into it.... 

Functionality demo:

```python
import yt
import numpy as np
shape=(7,8,9)

sim_time = np.datetime64('2021-09-09')
data = np.ones(shape)
data[:,5:, :] = 10
ds = yt.load_uniform_grid({'density': data}, shape, sim_time=sim_time)
```
The parameters info will display as:
```
yt : [INFO     ] 2022-10-14 17:06:57,416 Parameters: current_time              = 2021-09-09T00:00:00.000000000
```
And when plotting:
```python
for tunit in [None, "D"]:
    slc = yt.SlicePlot(ds, 'x', ('stream', 'density'))
    slc.annotate_timestamp(time_format='datetime',time_unit=tunit)
    slc.save(f"{tunit}.png")
```
None
D

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -567,6 +567,9 @@ class Dataset(abc.ABC):
             np.seterr(**oldsettings)
         return self._instantiated_index
 
+    def _format_display_time(self, time):
+        return time
+
     @parallel_root_only
     def print_key_parameters(self):
         for a in [
@@ -580,6 +583,8 @@ class Dataset(abc.ABC):
                 mylog.error("Missing %s in parameter file definition!", a)
                 continue
             v = getattr(self, a)
+            if a == "current_time":
+                v = self._format_display_time(v)
             mylog.info("Parameters: %-25s = %s", a, v)
         if hasattr(self, "cosmological_simulation") and self.cosmological_simulation:
             for a in [
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -339,6 +339,7 @@ class StreamDataset(Dataset):
         default_species_fields=None,
     ):
         self.fluid_types += ("stream",)
+        self._display_datetime = False
         self.geometry = geometry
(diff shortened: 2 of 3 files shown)


neutrinoceros:
Initially I thought this was a bad idea because I expected time zone issues would make this a much larger effort to do it *correctly*, but reading numpy's `datetime64` documentation I can see that this class of problems is out of scope

>  This is a “naive” time, with no explicit notion of timezones or specific time scales

Given that, it's probably a very good idea to limit yt's interface to limit support to exactly this class and keep stdlib `datetime` out of scope.

I'm game to review this more in depth (right now I have not even opened the diff tab) when you think it's ready

chrishavlin:
> Given that, it's probably a very good idea to limit yt's interface to limit support to exactly this class and keep stdlib datetime out of scope.

Exactly what I had in mind! I can add some more strict checks (I think the current implementation would error with a stdlib datetime, but it'd be good to have a more useful error message). 

And I'll switch this to WIP for now and switch it back when it's ready.

PR #4175, opened 2022-10-19
Issue 4174: Improvements to field_to_include specifications in create_firefly_object

### LINKED ISSUE #4174: BUG/ENH: create_firefly_object only allow fields that are common to all particle types
### Bug report

**Bug summary**

The documentation of `fields_to_include` in the `data_containers.create_firefly_object()` method does not specify that only fields that are common to all particle types are allowed. This should either be mentioned in the documentation (Option 1) or the behavior changed to allow fields that are not common to all particle types (Option 2, preferred). (Open to other possibilities though)

**Code for reproduction**

```python
# This uses the Gizmo gizmo_mhd_mwdisk sample dataset instead of the Ramses one in the
# "Exporting an Example Dataset to Firefly", but is otherwise mostly equivalent to that example.
# I've changed the names (ramses -> gizmo) and included the bounding box/unit base to 
# ensure the data is loaded correctly. This dataset includes six particle types: PartType0, which
# has the field "Temperature" and PartType1-5, which do not.
bbox = [[-600.0, 600.0], [-600.0, 600.0], [-600.0, 600.0]]
unit_base = {
    "length": (1.0, "kpc"),
    "velocity": (1.0, "km/s"),
    "mass": (1.0, "Msun"),
}
ge = yt.load("sample_data/gizmo_mhd_mwdisk/gizmo_mhd_mwdisk.hdf5",bounding_box=bbox,unit_base=unit_base)
region = ge.sphere(ge.domain_center, (1000, "kpc"))
reader = region.create_firefly_object(
       "GizmoExample",
       fields_to_include=["Temperature"],  # Temperature is only defined for PartType0
       fields_units=["code_temperature"],  
)

## adjust some of the options
reader.settings["color"]["io"] = [1, 1, 0, 1]  ## set default color
reader.particleGroups[0].decimation_factor = 100  ## increase the decimation factor

reader.writeToDisk()
```

**Actual outcome**

This produces the error:

```
YTFieldNotFound: Could not find field ('PartType1', 'Temperature') in gizmo_mhd_mwdisk.
Did you mean:
	('PartType0', 'Temperature')
	('PartType0', 'temperature')
	('gas', 'temperature')
```
And nothing is written to the disk.

**Expected outcome**

The expected behavior depends a bit on what option is chosen. 
- For the documentation change (Option 1), I guess I would expect a more useful error, perhaps: 
```
YTFieldNotFound: Could not find field ('PartType1', 'Temperature') in gizmo_mhd_mwdisk.
Please include only fields that are common to all particle types.
Did you mean:
	('PartType0', 'Temperature')
	('PartType0', 'temperature')
	('gas', 'temperature')
```
with nothing written to the disk.
 - For Option 2, skipping the field if not present for that particle type, I would expect something like:
```
datadir: /XXXX/GizmoExample -- is not a sub-directory of firefly/static/data. 
This may produce confusing or inoperable results. As such, we will create a symlink for you when you writeToDisk.
Make sure each field_array (1) has a field_radius_flag (0), assuming False.
Make sure each field_array (1) has a field_radius_flag (0), assuming False.
Make sure each field_array (1) has a field_radius_flag (0), assuming False.
Make sure each field_array (1) has a field_radius_flag (0), assuming False.
Make sure each field_array (1) has a field_radius_flag (0), assuming False.
Make sure each field_array (1) has a field_radius_flag (0), assuming False.
Warning: Not all particle types contained field(s): Temperature                            <--- New
PartType0 - 4091/409013 particles - 3 tracked fields
PartType1 - 4375/437500 particles - 2 tracked fields
PartType2 - 2500/250000 particles - 2 tracked fields
PartType3 - 1250/125000 particles - 2 tracked fields
PartType4 - 766/76546 particles - 2 tracked fields
PartType5 - 1/1 particles - 2 tracked fields
```
And in Firefly, the ability to set Temperature as a selectable field.

**Version Information**

  * yt version: 4.1.1
  * Other Libraries (if applicable): firefly v 3.2.0

yt installed from source 
python from anaconda
firefly from PyPI

mtryan83:
## PR Summary

Added code to allow fields that aren't in all particle types and/or are specified in (field_type,field_name) form.
Previously, fields_to_include could only consist of fields that were universal to all raw particle types. This was
not specified in the documentation and didn't appear to be "natural" behavior.
Lines 892-901 in `yt/data_objects/data_containers.py` incidentally fixes a bug when no fields are specified.

This is intended to fix #4174.

## PR Checklist


mtryan83:
pre-commit.ci autofix

neutrinoceros (commented):
In general it's hard to know for sure this is working as expected without tests, could you add some ?

mtryan83:
> In general it's hard to know for sure this is working as expected without tests, could you add some ?

I've added some basic tests (d5be708), but I'm not super familiar with this testing system. Is this what you're looking for?

neutrinoceros:
Thank you, and nice catch ! I don't have enough time to do a detailed review right now, but here are some general key advice for writing tests:
- never call test functions explicitly
- if you need to repeat a test with a different parameter set, use `@pytest.mark.parametrize`
- if there are no firefly tests already, I think it'd be best to add a dedicated test file for these new ones
- if you need to reuse a dataset for different tests (useful if loading is expensive), use a `@pytest.fixture`. This can be a bit tricky so don't hesitate with any questions if you need, and this step is not necessary

mtryan83:
Sorry, I just got back to working on this.  
> * never call test functions explicitly

This was a combination of a bad function name and function creep, my bad.
> * if you need to repeat a test with a different parameter set, use `@pytest.mark.parametrize`

Done. This also helped with the previous point.
> * if there are no firefly tests already, I think it'd be best to add a dedicated test file for these new ones

I'm just blind apparently. I moved the addition to the already existing test_firefly.py
> * if you need to reuse a dataset for different tests (useful if loading is expensive), use a `@pytest.fixture`. This can be a bit tricky so don't hesitate with any questions if you need, and this step is not necessary

Done to simplify the @pytest.mark.parametrize stuff. I think I did this correctly, but I'd appreciate a double-check. I've never used some of these capabilities before.

Two general questions: am I trying to do too much with one test function? And is there an equivalent to pytest.warns for the mylog.warning function?

yt/data_objects/tests/test_firefly.py
```diff
@@ -36,3 +42,127 @@ def test_firefly_write_to_disk():
...
+            ["Temperature"],
+            ["code_temperature"],
+            "gas",
+            dnr(),
+        ),  # Test unique field (Temperature)
+        (
+            ["dinos"],
+            ["code_length"],
+            None,
+            pytest.raises(YTFieldNotFound),
```
neutrinoceros, line 111: So I don't think I've ever seen this done this way, and I am uncertain that this is supported, because `pytest.raises` and `pytest.warns` are context managers, and as such they *need* to be initialised through a `with` statement, so I doubt that you can parametrize your expectations this way.
Alternatively, you may parametrize the exception type.

yt/data_objects/tests/test_firefly.py
```diff
@@ -1,6 +1,12 @@
 import tempfile
+from contextlib import nullcontext as dnr
```
neutrinoceros, line 2: why this alias ? `nullcontext` may be longer to type but it's much clearer IMO

yt/data_objects/tests/test_firefly.py
```diff
@@ -36,3 +42,127 @@ def test_firefly_write_to_disk():
...
+    (ds, dd) = firefly_test_dataset
+    with expectation:
+        reader = dd.create_firefly_object(
+            fields_to_include=fields_to_include,
+            fields_units=fields_units,
+            coordinate_units="code_length",
+        )
+    if not isinstance(expectation, dnr):
+        pass
+        return
```
neutrinoceros, line 142: `pass` is redundant if it's not the whole content of a block
```suggestion
        return
```

neutrinoceros (commented):
Thank you for keeping poking at this !

There are some functional issues with how you refactored your tests, as can be seen in
https://tests.yt-project.org/job/yt_py38_git/6076/testReport/junit/yt.data_objects.tests/test_firefly/firefly_test_dataset/
(though I'm not sure I actually understand went wrong here TBH; but let's work through it)

I want to add that, unfortunately we're still slowly migrating from nose to pytest, so some of our infrastructure using nose, and doesn't understand pytest. This means that any test file using pytest should be explicitly ignored in nose.
This requires updates to `tests/tests.yaml` (ass a `--ignore-file` entry), and `nose_unit.cfg` (update the ignore-files regex). 


Now to answer your questions
> am I trying to do too much with one test function?

probably yeah. We _have_ a bunch of very busy tests functions already so I feel like I need to be tolerant about those, but it's best to avoid them if the tests can also be expressed as smaller, more numerous and decoupled functions.
AFAIC, the DRY (Don't Repeat Yourself) principle doesn't apply to tests, and trying to never repeat test code actually hurts readability as well as maintainability IMO.

> And is there an equivalent to pytest.warns for the mylog.warning function?

Sort of. There's [caplog](https://docs.pytest.org/en/6.2.x/reference.html?highlight=caplog#std-fixture-caplog), but it's pretty hard to use within yt tests because logging is typically disabled in CI: https://github.com/yt-project/yt/blob/eda90ffe1f7e889810ad44849cbd5696f2a89289/tests/ci_install.sh#L25

There's a precedent for tests that still use it in yt/tests/test_load_sample.py but it's not as straightforward as I'd like. Don't feel like you have to go through this, we have very little logging tests, but of course feel free to if you'd like !

yt/data_objects/tests/test_firefly.py
mtryan83: I pulled this type of construction directly from the [pytest documentation](https://docs.pytest.org/en/6.2.x/example/parametrize.html#parametrizing-conditional-raising), so I think it's right? The other option I thought of was using 
`pytest.param(["dinos"], ["code_length"], None, mark=pytest.mark.xfail)`
and likewise for the nonexistent tuple, but it seemed like xfails were more for non-implemented features/unfixed bugs?

yt/data_objects/tests/test_firefly.py
mtryan83: Mostly for line length and to match the [pytest doc](https://docs.pytest.org/en/6.2.x/example/parametrize.html#parametrizing-conditional-raising).
I've changed `dnr` to `does_not_raise` to be clearer. Also, I think using `does_not_raise()` is a little clearer than `nullcontext()` in the `@pytest.mark.parametrize` portion.

mtryan83:
> There are some functional issues with how you refactored your tests, as can be seen in https://tests.yt-project.org/job/yt_py38_git/6076/testReport/junit/yt.data_objects.tests/test_firefly/firefly_test_dataset/ (though I'm not sure I actually understand went wrong here TBH; but let's work through it)

I think I know what's wrong actually, I was passing the dataset and data region out of the fixture as a tuple, and then unpacking the tuple in the test function:
```
def test_fields_specification( firefly_test_dataset, ... ):
    (ds, dd) = firefly_test_dataset
```
because I only need `dd`, but if I don't keep `ds` it gets garbage collected and the whole thing breaks. But I guess pytest thinks I'm calling the fixture twice, instead of using the input parameter.  The solution was to change the fixture to only return the dataset and just get the data region in the test function:
`dd = firefly_test_dataset.all_data()`
(I didn't see the fixture message when I ran on command line or I would have fixed it already)
 
> I want to add that, unfortunately we're still slowly migrating from nose to pytest, so some of our infrastructure using nose, and doesn't understand pytest. This means that any test file using pytest should be explicitly ignored in nose. This requires updates to `tests/tests.yaml` (ass a `--ignore-file` entry), and `nose_unit.cfg` (update the ignore-files regex).

Updated.

> probably yeah. We _have_ a bunch of very busy tests functions already so I feel like I need to be tolerant about those, but it's best to avoid them if the tests can also be expressed as smaller, more numerous and decoupled functions. AFAIC, the DRY (Don't Repeat Yourself) principle doesn't apply to tests, and trying to never repeat test code actually hurts readability as well as maintainability IMO.

Hmm. I think I can split it into testing field names and field tuples, so I'll try that.  That should get rid of some of the conditional logic and make it more obvious what's being tested. 

> > And is there an equivalent to pytest.warns for the mylog.warning function?
> 
> Sort of. There's [caplog](https://docs.pytest.org/en/6.2.x/reference.html?highlight=caplog#std-fixture-caplog), but it's pretty hard to use within yt tests because logging is typically disabled in CI:
> 
> https://github.com/yt-project/yt/blob/eda90ffe1f7e889810ad44849cbd5696f2a89289/tests/ci_install.sh#L25
> 
> There's a precedent for tests that still use it in yt/tests/test_load_sample.py but it's not as straightforward as I'd like. Don't feel like you have to go through this, we have very little logging tests, but of course feel free to if you'd like !

I was mainly thinking of using it to test the `"detected but did not request"`/`"requested but not available"` logic, but if logging is disabled typically anyway, there's not much use.

mtryan83:
I split up the test function into 3: field strings, field tuples, and mixed strings and tuples. 

I also noticed that the old firefly code tests (`test_firefly_JSON_string` and `test_firefly_write_to_disk`) were being skipped because they were referencing the old module name. I fixed that and updated the tests slightly to match the new firefly api (ae362e9), but they don't really look that useful. I marked the the `test_firefly_write_to_disk` test as skipped because I don't think it actually tests that anything is written correctly. Anything more would probably be better as a new PR or something. Feel free to revert the commit if you want.

yt/data_objects/tests/test_firefly.py
neutrinoceros: seems reasonable, thanks

neutrinoceros:
>  but if logging is disabled typically anyway

... that's *in the context of testing*. I don't know what most users do with yt's logger, but by default it's always on and set to `INFO` level

mtryan83:
> ... that's _in the context of testing_. I don't know what most users do with yt's logger, but by default it's always on and set to `INFO` level
Sorry, I just meant in the testing context.  I'll definitely leave those messages in!

yt/data_objects/data_containers.py
```diff
@@ -813,6 +813,19 @@ def create_firefly_object(
             datadir=datadir, clean_datadir=True, **kwargs
         )
 
+        ## Ensure at least one field type contains every field requested
+        kys = self.ds.derived_field_list
+        all_fields = list(zip(*kys))[1]
+        for field in fields_to_include:
+            field = field if not isinstance(field, tuple) else field[1]
+            if field not in all_fields:
+                raise YTFieldNotFound(field=field, ds=self.ds)
```
neutrinoceros, line 822: This doesn't seem right; I think we need to check that fields passed as single keys are not ambiguous to match the rest of yt's API
This would be done like so
```suggestion
        # error if any requested field is unknown or ambiguous
        dd = self.ds.all_data()
        fields_to_include = dd._determine_fields(fields_to_include)
```

yt/data_objects/data_containers.py
```diff
@@ -813,6 +813,19 @@ def create_firefly_object(
...
+        ## Also generate equivalent of particle_fields_by_type including
+        ## derived fields
+        kysd = defaultdict(list)
+        for k, v in kys:
```
neutrinoceros, line 826: in order to apply my previous suggestion
```suggestion
        for k, v in self.ds.derived_field_list:
```

yt/data_objects/data_containers.py
```diff
@@ -836,6 +849,19 @@ def create_firefly_object(
...
+                if isinstance(field, tuple):
+                    ftype, fname = field
+                    if not (ftype == ptype or ftype == "all"):
+                        continue
+                    if fname not in kysd[ptype]:
+                        raise YTFieldNotFound(field=(ftype, fname), ds=self.ds)
+                    field = fname
+                elif field not in kysd[ptype]:
+                    mylog.warning("requested (but not available) %s %s", ptype, field)
+                    continue
```
neutrinoceros, line 863: with my main suggestion to use `_determine_fields` uphill, this would simplify as
```suggestion
                ftype, fname = field
                if ftype not in (ptype, "all"):
                    continue
                field = fname
```
while we're here, could you maybe also avoid changing the type of the `field` variable from `tuple[str, str]` to `str`, and just use `fname` where appropriate instead ?

neutrinoceros (commented):
Here are important suggestions. I think if we agree on this I might go for one last, smaller review after this is settled

yt/data_objects/tests/test_firefly.py
```diff
@@ -35,4 +41,195 @@ def test_firefly_write_to_disk():
...
+    assert_array_equal(
+        dd[("gas", "relative_particle_position")].d,
+        reader.particleGroups[1].coordinates,
+    )
+    if fields_to_include is None:
+        return
+    all_pgs = ["PartType1", "gas"]
+    for field in fields_to_include:
+        for ptype in all_pgs:
+            pg = reader.particleGroups[0 if ptype == "PartType1" else 1]
```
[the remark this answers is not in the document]
mtryan83: Python is clearly not my first language :laughing:

yt/data_objects/data_containers.py
mtryan83: I don't disagree with this, but I think it would be a change in behavior of the `create_firefly_object` method.  The doc for `fields_to_include` just specifies
> fields_to_include : array_like of strings
>          A list of fields that you want to include in your Firefly visualization for on-the-fly filtering and colormapping.

which to me implies a potentially ambiguous name like `"Masses"` from the test dataset would be equivalent to `("*", "Masses")` (which I don't think exists in any form), rather than a specific field.  The [4.1 doc](https://yt-project.org/doc/analyzing/fields.html#accessing-fields-without-a-field-type) implies we should just change the `fields_to_include` doc to warn about ambiguous names, but I can also see the value for allowing something like a `"Masses" ==> ("*", "Masses")` constructor for this particular method.  What is the preferred path?

Also, this particular suggestion doesn't seem to work for cases where the particle type hasn't been specified yet. For example, it's breaking the `"Temperature"` test case because while `("gas", "Temperature")` is valid, `("all", "Temperature")` isn't. But the `"gas"` type hasn't been specified yet, so it falls back to the `"all"` type here. Am I missing something from the test definition?

yt/data_objects/data_containers.py
mtryan83: Fixed the `field` conversion ea4eab5. I'm not sure why I didn't do it this way earlier.

yt/data_objects/tests/test_firefly.py
```diff
@@ -35,4 +41,195 @@ def test_firefly_write_to_disk():
...
+    dd = firefly_test_dataset.all_data()
+    with expectation:
+        reader = dd.create_firefly_object(
+            fields_to_include=fields_to_include,
+            fields_units=fields_units,
+            coordinate_units="code_length",
+        )
+    if not isinstance(expectation, does_not_raise):
+        # constructed like this so we don't need to import RaiseError from pytest
+        return
```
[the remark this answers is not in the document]
mtryan83: Separating them really does make all of them much simpler. Thanks!

yt/data_objects/data_containers.py
neutrinoceros: > The [4.1 doc](https://yt-project.org/doc/analyzing/fields.html#accessing-fields-without-a-field-type) implies we should just change the fields_to_include doc to warn about ambiguous names

That is not correct. yt 4.0 used to raise warnings in case of ambiguity, and selected (arbitrarily) a single candidate field, but 4.1 *just errors out* in case of ambiguity, and this is the behaviour we want to enforce everywhere going forward.

I am not sure what you mean by `("*", "Masses")`, but if you mean "all fields with this name should be included, regardless of their field type", then I think that it's just equivalent to `("all", "Masses")`, and that option should already be listed in the error message that you'd get from calling `_determine_fields` as I suggested.

You are right that this is a breaking change, but I think it's required nonetheless, because it's an oversight on our part.
I don't think it'd be trivial to only deprecate the old behaviour, but if you can see a way to, that would be preferable, of course.

mtryan83: > That is not correct. yt 4.0 used to raise warnings in case of ambiguity, and selected (arbitrarily) a single candidate field, but 4.1 _just errors out_ in case of ambiguity, and this is the behaviour we want to enforce everywhere going forward.

Sorry, I wasn't clear. I just meant explicitly mention in the doc that ambiguous fields weren't allowed. 
> I am not sure what you mean by `("*", "Masses")`, but if you mean "all fields with this name should be included, regardless of their field type", then I think that it's just equivalent to `("all", "Masses")`

I was thinking more along the lines of "any fields with this name should be included", rather than "all fields with this name". So e.g. if there were three particle types, and two of them had `"field1"` (like `"ptype1"` and `"ptype2"`, but not `"ptype3"`), specifying `"field1"` would bring in `("ptype1","field1")` and `("ptype2","field1")`. Since `"ptype3"` does not have `"field1"`, `("all","field1")` doesn't exist and it's not equivalent.  But really this is just a convenience, so it's not that important.

> You are right that this is a breaking change, but I think it's required nonetheless, because it's an oversight on our part. I don't think it'd be trivial to only deprecate the old behaviour, but if you can see a way to, that would be preferable, of course.

Sounds good. I also don't see an easy way to just deprecate the old behavior. I'll see if I can clarify the doc too.

mtryan83:
I'm having trouble with the test cases. I can't seem to get e.g. `dd["Temperature"]` to work, when I know it's a unique field (and works outside of running pytest) because it always falls back to searching for `("all", "Temperature")`.  Is there something that needs to be set in the test function to help with that?

yt/data_objects/data_containers.py
neutrinoceros: > Since "ptype3" does not have "field1", ("all","field1") doesn't exist and it's not equivalent.

IIRC, yes it is. I'm curious where could be the source of misunderstanding here. I think I checked the definition of "all" rather recently but I'll check again. (Commenting ahead of time to avoid forgetting about this)

neutrinoceros: This comment made me realise I was mislead.
https://github.com/yt-project/yt/blob/2efdbf3beff5f82604d2689182e425f8ec29bbc4/yt/data_objects/static_output.py#L736

So indeed you are right, we don't have an idiomatic way to express "include *any* matching field".

neutrinoceros:
> (and works outside of running pytest)

Can you produce a minimal example for this ? I do not understand why it would fail only within pytest, so I'm tempted to think you're actually running slightly different code.

mtryan83:
> Can you produce a minimal example for this ? I do not understand why it would fail only within pytest, so I'm tempted to think you're actually running slightly different code.

You're absolutely right. I was using different datasets outside of pytest (e.g. the gizmo_mhd_mwdisk) but hadn't actually tried creating the `fake_particle_ds` outside of pytest. So the problem is with the `fake_particle_ds` and specifically when there are 2+ particle types (removing pt1 causes the `ad["field1"]` call to succeed both times).  Here's an MWE:
```python
import yt
from yt.utilities.exceptions import YTFieldNotFound
from yt.testing import fake_particle_ds
# create dataset
def test_fake_dataset():
    ds_fields = [  
        ("pt1", "particle_position_x"),
        ("pt1", "particle_position_y"),
        ("pt1", "particle_position_z"),
        ("pt2", "particle_position_x"),
        ("pt2", "particle_position_y"),
        ("pt2", "particle_position_z"),
        ("pt2", "field1"), 
    ]
    ds_field_units = ["code_length"]*7
    ds_negative = [0]*7
    ds = fake_particle_ds(
        fields=ds_fields,
        units=ds_field_units,
        negative=ds_negative,
    )
    return ds

ds = test_fake_dataset()
ad = ds.all_data()
try: 
    print(ad["field1"])
except(YTFieldNotFound):
    print("Expected error, looking for (all, field1) which DNE")
print(ad["pt2","field1"]) # Setting gas1 as most recently requested -> works elsewhere
try:
     print(ad["field1"])
except(YTFieldNotFound):
    print("ERROR: This should have produced the same result as (pt2, field1)")
```
Is there something I'm supposed to call when I'm creating the `fake_particle_ds` that will fix this?

neutrinoceros:
I see that your expectations don't match the observed behaviour, but that behaviour is intended: calling `ad["field1"]` should yield the same error (if any) regardless of the query history. The intended rule is that an error should be raised _whenever the user requests a field with an implicit field type that is not uniquely matched_, and it seems to me that it's exactly what we're seeing in your example.

mtryan83:
>  _whenever the user requests a field with an implicit field type that is not uniquely matched_, 

I must be misunderstanding something then, isn't this the error case for ambiguous fields?  I wouldn't have thought that would apply.  
I was assuming that the implicit field type here is `pt2`, as that is the only field type with the field `field1` (confirmed(?) using e.g. `ds.fields`), and so there should only be one unique match: `(pt2, field1)`.  Is that wrong? Is the implicit field type something else?

neutrinoceros:
> isn't this the error case for ambiguous fields? 

yes, but "ambiguous" is not affected by "most recently accessed".
Implicit field types are only ever accepted when there's exactly one match in the whole field collection. This is done precisely to avoid confusion and misconceptions; yt < 4.1 wasn't doing anything as smart as you seemed to assume, it simply took the first matching field and rolled with it, and ordering wasn't actually clearly defined, so it was all just undefined behaviour !

Btw, some of the test failures that you can see in you most recent run are completely unrelated to your PR. I'm assuming we can expect them to be resolved upstream in a matter of hours to days. See https://github.com/yt-project/yt/issues/4224 for context.

mtryan83:
I think there's still a disconnect between what we're talking about, so I'll try to ask in a different way. 
I have two assumptions based on the doc:
1. There is a certain kind of field that can be specified using solely a string name, not a `(ftype, fname)` tuple, such that e.g. `ds["field1"]` is valid.
2. The kind of specification in the first assumption should be valid for the `fields_to_include` parameter of the `create_firefly_object` method.  For example, `create_firefly_object( ..., fields_to_include=["field1"], ...)`.

My question is then _how should I construct a test dataset to test assumption 2?_ Of course, either one of those assumptions could be wrong, in which case I think I can figure out the path forward.  I just can't seem to create a test dataset that works assuming they're both correct.

I've updated the test file; it should be more obvious what is being tested and how.

neutrinoceros:
Sorry for the delay, had a busy week...
I tried to read the existing code with a fresh look, and found that the _existing_ logic in `create_firefly_object` is very much at odds with the rest of yt's API, which is probably why we're having a hard time aligning our points of view.

It also seems to me that yt's code is a little bit contradictory, which of course doesn't help making things clear.
Specifically, the special particle type `"all"` *could* represent a set *union* or an *intersection*, the name itself isn't without ambiguity, but I guess that would be fine if we had one consistent convention. However, it is defined as a `ParticleUnion` object, but implemented as a set *intersection*. This contradiction is already apparent within a 5 lines interval here:
https://github.com/yt-project/yt/blob/45dc49bedea0ca551895109f9dcdaae71f198df6/yt/data_objects/static_output.py#L728-733

I am not sure what would be the best approach to fix this, but it seems that it could easily go way out of scope for the present PR.

Instead, I propose (contrary to what I previously said ! very sorry about that) that we keep the existing logic for `create_firefly_object`, and create actual *unions* of particle types. I think this is in line with the original design as well as what you had in mind initially; again I apologise for the confusion, that I think now was mainly on my side.
We could use a chunky comment explaining this in set theory terms within the function to reduce the risk that this confusion arises again in the future.

mtryan83:
> Instead, I propose (contrary to what I previously said ! very sorry about that) that we keep the existing logic for `create_firefly_object`, and create actual _unions_ of particle types. I think this is in line with the original design as well as what you had in mind initially; again I apologise for the confusion, that I think now was mainly on my side. We could use a chunky comment explaining this in set theory terms within the function to reduce the risk that this confusion arises again in the future.

To make very sure I'm following correctly :sweat_smile: . 
1.  _existing_ here means what's currently in e.g. yt 4.1, **not** the current state of this PR (and/or prior to 21beaf4)? 
2. I'm not sure what exactly you mean by 
   > create actual _unions_ of particle types
   
   Could you clarify, possibly with some psuedocode? Is this along the lines of the `("any", "field")` construction I mentioned?
3. Would it be helpful to deprecate single string field specifications (and update the `create_firefly_object` doc to reflect this)? We'd still have to figure out something for now but we wouldn't need to worry about ambiguities/particle unions/etc in the future. 
4. Following on 2. and 3., what about internally treating a single string spec (`"field"`) as an `("any", "field")` with a deprecation warning?

neutrinoceros:
1. yes I mean yt 4.1
2. by "actual unions" I meant to describe yt 4.1's behaviour for this specific function: aggregate any particle types that match a given particle field, even if not all of them have it, so yes, that would be what we referred to as `("any", "field")`.
3. and 4. Since the current behaviour cannot be replicated with a different, existing API, maybe single string should still work, just not implicitly ? It could be kept an opt-in associated with a flag argument like `match_any_particle_types`, implemented like so

```python
def create_firefly_object(
    ...,
    *,
    match_any_particle_types=None,
):
   if match_any_particle_types is None:
       # not specified, switching to (temporary) default
       issue_deprecation_warning(
           "match_any_particle_types wasn't specified. Its current default value (True) will be changed to False in the future. "
           "Pass an actual value (True or False) to silence this warning. ",
           ...
       )
       match_any_particle_types = True
 ```
WDYT ?

File: yt/data_objects/data_containers.py [hunk 9]
```diff
@@ -808,13 +815,56 @@ def create_firefly_object(
...
+                    # equivalent particle fields
+                    for c in candidates:
+                        unambiguous_fields_to_include.append(c)
+                        unambiguous_fields_units.append(field_unit)
+            fields_to_include = unambiguous_fields_to_include
+            fields_units = unambiguous_fields_units
+        # error if any requested field is unknown or (still) ambiguous
+        # This is also sufficient if match_any_particle_types=False
         dd = self.ds.all_data()
         fields_to_include = dd._determine_fields(fields_to_include)
```
mtryan83, line 869: Finally got back to this.
I think this should fix it. Basically, I kept the old `fields_to_include = dd._determine_fields(fields_to_include)` behavior but added a switch based on your suggested `match_any_particle_types` flag prior to that. If the flag is True, we try to disambiguate any ambiguous single strings by just replacing them with all possible matching candidates and sending the new list to `_determine_fields`. If the flag is False, we just rely on `_determine_fields`.

mtryan83:
I also updated the tests using the new `match_any_particle_types` flag and updated the function doc.
Let me know what you think.

yt/data_objects/data_containers.py
```diff
@@ -808,11 +815,64 @@ def create_firefly_object(
             )
             datadir = JSONdir
 
+        if match_any_particle_types is None:
+            # not specified, switching to (temporary) default
+            issue_deprecation_warning(
+                "match_any_particle_types wasn't specified. Its current default "
+                "value (True) will be changed to False in the future. "
```
neutrinoceros, line 822: In hindsight, I don't think we need to change the default in the future.
```suggestion
                "match_any_particle_types wasn't specified. "
```

yt/data_objects/data_containers.py
```diff
@@ -808,11 +815,64 @@ def create_firefly_object(
...
+                    # skip tuples, they'll be checked with _determine_fields
+                    unambiguous_fields_to_include.append(field)
+                    unambiguous_fields_units.append(field_unit)
+                    continue
+                _, candidates = self.ds._get_field_info_helper(field)
+                if len(candidates) < 2:
+                    # Field is unambiguous, add in tuple form
+                    # This should be equivalent to _tupleize_field
+                    unambiguous_fields_to_include.append(candidates[0])
+                    unambiguous_fields_units.append(field_unit)
```
neutrinoceros, line 856: 0 candidates shouldn't be a valid case, and since we're using `candidates[0]` it's clear we're already assuming we have exactly one here
```suggestion
                if len(candidates) == 1:
                    # Field is unambiguous, add in tuple form
                    # This should be equivalent to _tupleize_field
                    unambiguous_fields_to_include.append(candidates[0])
                    unambiguous_fields_units.append(field_unit)
```

neutrinoceros (approved):
Just one very minor, final remark. Otherwise I believe this is ready, good job and thank you !

yt/data_objects/data_containers.py
```diff
@@ -754,6 +755,12 @@ def create_firefly_object(
             A flag to optionally print the fields that are available, in the
             dataset but were not explicitly requested to be tracked.
 
+        match_any_particle_types : boolean
+            A flag to specify that ambiguous fields in fields_to_include
+            should match all relevant particle groups instead of raising an
+            error. Note that the default value (True, to match previous
+            behavior) will change to False in the future.
+
```
mtryan83, line 763:
```suggestion
            error. 
```

yt/data_objects/data_containers.py
```diff
@@ -714,6 +714,7 @@ def create_firefly_object(
         show_unused_fields=0,
         *,
         JSONdir=None,
+        match_any_particle_types=None,
```
mtryan83, line 717:
```suggestion
        match_any_particle_types=True,
```

yt/data_objects/data_containers.py
```diff
@@ -808,11 +815,64 @@ def create_firefly_object(
...
+                "match_any_particle_types wasn't specified. Its current default "
+                "value (True) will be changed to False in the future. "
+                "Pass an actual value (True or False) to silence this warning. ",
+                since="4.2",
+            )
+            # Note that all of the tests in test_firefly.py have this explicitly
+            # specified (and labeled) to avoid this warning. They will need to be
+            # changed when this behavior changes in the future
+            match_any_particle_types = True
+
```
mtryan83, line 830:
```suggestion
```
If we're not going to change the default in the future, we should remove the deprecation warning, right?

yt/data_objects/tests/test_firefly.py
```diff
@@ -12,9 +17,10 @@ def test_firefly_JSON_string():
         None,
         velocity_units="cm/s",
         coordinate_units="cm",
+        match_any_particle_types=True,  # Explicitly specifying to avoid deprecation warning
```
mtryan83, line 20:
```suggestion
```
No longer need to explicitly specify if deprecation warning is removed.

yt/data_objects/tests/test_firefly.py
```diff
@@ -33,6 +39,249 @@ def test_firefly_write_to_disk():
...
+)
+def test_field_empty_specification(
+    firefly_test_dataset, fields_to_include, fields_units
+):
+    dd = firefly_test_dataset.all_data()
+    reader = dd.create_firefly_object(
+        fields_to_include=fields_to_include,
+        fields_units=fields_units,
+        coordinate_units="code_length",
+        match_any_particle_types=True,  # Explicitly specifying to avoid deprecation warning
```
mtryan83, line 90:
```suggestion
```

yt/data_objects/tests/test_firefly.py
```diff
@@ -33,6 +39,249 @@ def test_firefly_write_to_disk():
...
+    firefly_test_dataset,
+    fields_to_include,
+    fields_units,
+):
+    dd = firefly_test_dataset.all_data()
+    reader = dd.create_firefly_object(
+        fields_to_include=fields_to_include,
+        fields_units=fields_units,
+        coordinate_units="code_length",
+        match_any_particle_types=True,  # Explicitly specifying to avoid deprecation warning
```
mtryan83, line 191:
```suggestion
```

yt/data_objects/tests/test_firefly.py
```diff
@@ -33,6 +39,249 @@ def test_firefly_write_to_disk():
...
+
+
+@requires_module("firefly")
+def test_field_mixed_specification(firefly_test_dataset):
+    dd = firefly_test_dataset.all_data()
+
+    reader = dd.create_firefly_object(
+        fields_to_include=["pt2only_field", ("pt1", "common_field")],
+        fields_units=["code_length", "code_length"],
+        match_any_particle_types=True,  # Explicitly specifying to avoid deprecation warning
```
mtryan83, line 274:
```suggestion
```

mtryan83 (commented):
(Can't find your comment about changing the default, sorry!)
These should be the changes needed if we don't want to change the default any time soon. I mostly just removed the deprecation warning (didn't make sense to keep it if the default's not changing, since we don't need anything from the user at this point) and the setting in the test file.

neutrinoceros:
I changed my mind and removed the comment, that's why you can't find it. Sorry about that ! Either way it's fine, I mostly didn't want to ask for more work but now that you've done it, no need to undo it.

neutrinoceros (approved):

neutrinoceros:
We just had an instability in CI due to the latest version of numpy. You'll need to either merge from main or rebase this branch in order to fix this here (required for merging)

yt/data_objects/data_containers.py
```diff
@@ -754,6 +755,10 @@ def create_firefly_object(
             A flag to optionally print the fields that are available, in the
             dataset but were not explicitly requested to be tracked.
 
+        match_any_particle_types : boolean
+            A flag to specify that ambiguous fields in fields_to_include
+            should match all relevant particle groups instead of raising an
+            error.
```
chrishavlin, line 761: I think the language could be a little clearer here. How about:
```suggestion
        match_any_particle_types : boolean
            If True, when any of the fields_to_include match multiple particle groups then 
            the field will be added for all matching particle groups. If False, an error is raised 
            when encountering an ambiguous field. Default is True.
```

neutrinoceros:
Now this has conflicts... I would solve them myself but you seem to be confortable with rebasing so ... can you do it (again) please ? Sorry this is turning so demanding, I will try to find another reviewer so hopefully we can put it at rest soon.

matthewturk:
I'm happy to serve as a reviewer!

matthewturk:
I went ahead and solved the conflicts

chrishavlin (commented):
This looks good to me, only some minor suggestions and questions.

yt/data_objects/data_containers.py
```diff
@@ -813,6 +818,46 @@ def create_firefly_object(
...
+                    # equivalent particle fields
+                    for c in candidates:
+                        unambiguous_fields_to_include.append(c)
+                        unambiguous_fields_units.append(field_unit)
+            fields_to_include = unambiguous_fields_to_include
+            fields_units = unambiguous_fields_units
+        # error if any requested field is unknown or (still) ambiguous
+        # This is also sufficient if match_any_particle_types=False
+        dd = self.ds.all_data()
+        fields_to_include = dd._determine_fields(fields_to_include)
```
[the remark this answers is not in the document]
mtryan83: Some **brief** testing seems to confirm that it works, but I'm totally fine with waiting until @neutrinoceros weighs in.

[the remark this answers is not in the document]
neutrinoceros: To me the idea was to use _any_ data container's `_determine_fields` method. So indeed, I agree with Chris' suggestion to use the `self` object instead

neutrinoceros:
@chrishavlin, feel free to merge if you feel it's ready. I'll be off for today

chrishavlin (approved):

chrishavlin:
@matthewturk did you still wanna take a look?

PR #4176, opened 2022-10-21
BUG: make a MPL failure non-fatal for set_norm with old version of MPL

neutrinoceros:
## PR Summary

This example script doesn't work with matplotlib <= 3.5 and crashes with an `AttributeError`
```python
import yt
from matplotlib.colors import TwoSlopeNorm

ds = fake_random_ds(16)
field = ("gas", "density")
p = SlicePlot(ds, "z", field)
p.set_norm(field, norm=TwoSlopeNorm(vcenter=0, vmin=-0.5, vmax=1))
```
However this error is not critical so this patch makes it non-fatal, with a clear warning message.
This is extracted from #4122 to enable backporting.
Backporting the tests that come with it in #4122 would be much more complicated, so I hope it's okay to backport just the patch.

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -333,7 +333,7 @@ class ImagePlotMPL(PlotMPL, ABC):
                         "Failed to format colorbar ticks. "
                         "This is expected when using the set_norm method "
                         "with some matplotlib classes (e.g. TwoSlopeNorm) "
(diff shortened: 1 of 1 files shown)


chrishavlin (approved):

neutrinoceros:
@chrishavlin is this ready to go now ?

chrishavlin (approved):

PR #4177, opened 2022-10-21
DEPR: deprecate calling yt.load with first argument passed as keyword

### LINKED ISSUE #2879: Use `fn`, `filename`, `name`, or `my_cool_argname` in the codebase?

neutrinoceros:
## PR Summary
This is a proposed solution to close #2879:
IMO, the name of the first argument to yt.load is bound to be meaningless.
It's current name `fn` is short for "filename", but we support loading from directories and in some cases, urls !
I think that making it a positional-only argument lifts out the weigh of carying a meaningless name forever.
Now, the difficult part is in continuously evolving the API to avoid sudden breaking changes, which is why 99% of this patch consists in adding a custom decorator to do just that, so existing code that might rely on loading data as `yt.load(fn=..., ...)` will not immediately break but will still emit some warnings, and we can keep backward compatibility for a while.

TODO:
- [x] add a test for the warning part

### DIFF
--- a/yt/_maintenance/deprecation.py
+++ b/yt/_maintenance/deprecation.py
@@ -1,5 +1,7 @@
 import warnings
-from typing import Optional
+from functools import wraps
+from types import FunctionType
+from typing import Dict, Optional
 
 
 class VisibleDeprecationWarning(UserWarning):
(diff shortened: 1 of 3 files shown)


neutrinoceros:
I'll note that the fact that this passed all tests without a warning indicates that calling `yt.load(fn=...)` is indeed something we never encouraged

matthewturk:
OK, so, I don't *mind* this, but I'm surprised it's so much code, and that it's that high of a priority.  Do we really need this?

neutrinoceros:
Not a high priority, just something that was on my mind for 2 years, but required dropping Python 3.7
I don't see another way to close #2879 (other than marking it as 'wontfix').

matthewturk (approved):
Let's do it

PR #4179, opened 2022-10-22
ENH: enable plotting curvilinear geometries in native coordinates

### LINKED ISSUE #3324: curvilinear geometries and pseudo-cartesian projections
### Bug report

**Bug summary**

Projecting data points to a rectangular grid with (some) angular coordinates is a very common technique for visualising data that's natively parametrised with a curvilinear geometry (polar, spherical, cylindrical).
For instance it is very useful to represent a polar dataset (spatially covering a "disk"; or a portion of a disk) in `r VS theta` axes.

Currently, I don't know that there's a method for doing this in yt.
The AMRVAC frontend offers a cheap hack to get something similar, by exposing a `geometry_override` kwarg at load time (this was implemented to improve support for old dataset which didn't contain information on their own geometry), so it is possible to get something like that (but not quite what we want). See example

**Code for reproduction**

```python
import yt
ds = yt.load_sample("bw_polar_2d", geometry_override="cartesian")
yt.plot_2d(ds, ("gas", "density")).save("/tmp/cart")
```
cart_Slice_z_density
(plot edited with yt 4.1 because it takes less space)
<details> <summary> Old plot done with yt < 4.1 </summary>

cart_Slice_z_density

</details>

It's better than nothing, and largely a proof that this is feasible in yt, but some improvements can be made:
- the "geometry" of the dataset itself is in fact irrelevant, and shouldn't have to be hacked at load time (this is especially costly for large datasets with a noticeable load time, unlike my toy example above).
- plot axis labels are incorrect (this is expected with my hack)
- there is no user-facing option to tweak the aspect ratio of the axes. This toy example only covers a radial range which happens to stay close to unity, so the end product looks decent, but it would be completely unusable with a radial interval of, say [1000, 10000], as the resulting frame would be extremely elongated.

The general solution is a clear separation between the native coordinate system, and the projection axes (both are currently referred to as the "geometry"; which IMO is fine is the former case); Additionally some way to set the aspect ratio of the final frame is required (when angular coordinates are combined with length coordinates).

I believe all of this can be implemented either at the level of pixelizers with user-facing options in plotting fonctions.

neutrinoceros:
## PR Summary

This adds a new keyword argument to `yt.SlicePlot`, `yt.ProjectionPlot` and `yt.plot_2d` that's primarily targetted at enabling plotting in geometry-native coordinates for curvilinear geometries.
For instance, say I ran a simulation with a cylindrical coordinate system. Its natural axes are (r, theta, z). Now, I want to plot slice using `z` as the normal, the easiest way to do so is to use `r` and `theta` as my plot axes, but that is not what yt does:
```python
import yt

ds = fake_amr_ds(fields=[("gas", "density")], units=["g/cm**3"], geometry="cylindrical")
p.save("/tmp/default.png")
```
default

but this kind of plot is not only easy to do outside yt, it's also extremely useful to vizualise some structures that are common in simulations where cylindrical coordinates are most appropriate, such as spirals.
Just to give some examples (but not trying to be exhaustive !), this is used in studies of Cataclysmic Variables (CVs) (e.g. fig 3 from https://iopscience.iop.org/article/10.3847/0004-637X/823/2/81/pdf) and planet-disk interactions in protoplanetary disk studies (e.g. fig 3 from https://arxiv.org/pdf/2002.02280.pdf)

Under the hood, this PR allows users to choose a different pixelizer than the default one, and most importantly it makes the cartesian pixelizer usable in all geometries for any normal axis.
Now you can do this
```python
p = yt.SlicePlot(ds, "z", ("gas", "density"), axes_transform="geometry_native")
p.save("/tmp/native.png")
```
native
And thanks to yt 4.1's `flip_vertical_axis`, `flip_horizontal_axis` and `swap_axes`, no-one needs to commit to the defaults and we can actually meet all users of this feature where they are at.

A consequence of this change is that plot callbacks, which are all primarily implemented for cartesian geometries, can be unlocked for all geometries within this special case. I have yet to determine if this specific enhancement should be done within the same PR or left to do as a follow up to avoid bloating the first PR.

Another consequence is that it makes plotting with the x and y axis not sharing a dimension (if one coordinate is angular) much more common. Previously the only case where this happened by default was for plotting `r`-normal slices in cylindrical coordinates (where the x axis was `theta` and the y axis was `z`).
I figured, the existing rule that aspect ratios should be 1 by default doesn't make much sense in that case, so I changed that rule, but it should not affect anything else by default.

At the time of writing, this is still a proof of concept. I haven't finished refactoring the SpectralCubeCoordinateHandler class, but I'm opening early to see if I'm breaking any tests unexpectedly.

Closes #3324

## PR Checklist

- [ ] pass CI
- [ ] handle collisions with cartopy API
- [ ] make sure all added `#TODO(clm)` comments are adressed
    - [ ] with docstrings
    - [ ] ... and narrative docs


matthewturk:
Seems cool!  Thanks for doing this.

yt/frontends/nc4_cm1/data_structures.py
```diff
@@ -87,13 +87,6 @@ def __init__(
         )
         self.storage_filename = storage_filename
 
-    def _setup_coordinate_handler(self):
```
neutrinoceros, line 90: The refactor I'm doing in this PR makes `_x_pairs` and `_y_pairs` unused, so this method doesn't do anything anymore. I also cannot see a simple way to emulate this behaviour within my refactor, mostly because this method was violating the responsibility separation principle.
As a result, this is a breaking change for nc4_cm1 users, *but* there is a workaround since plot window objects recently gained axis-manipulation API (`flip_horizontal_axis`, `flip_vertical_axis` and `swap_axes`) which should allow user to reproduce plots, albeit at the cost of an additional explicit function call.

ping @keltonhalbert for information

neutrinoceros:
force pushed as a unique commit to fix many conflicts (and avoid having to do it again)

neutrinoceros:
This currently fixes #4350, however I'll give priority to #4417 as a simpler fix to the same issue.

neutrinoceros:
As much as I'd love this to happen in yt 4.2, I can see it's nowhere near ready now so let's take it out of the milestone

neutrinoceros:
pre-commit.ci run

neutrinoceros:
This has grown a small, but possibly deadly, conflict with #4440. I'm not sure how easily I can recover from it.

PR #4181, opened 2022-10-22
DOC: cleanup calls to private methods _setup_plots in doctests

neutrinoceros:
## PR Summary
Follow up to #4121


chrishavlin (changes requested):
Up higher in the docstring of `set_mpl_projection` it still says

```
        To annotate the plot with coastlines or other annotations,
        `_setup_plots()` will need to be called after this function
        to make the axes available for annotation.
```
should change that to `render()` as well.

neutrinoceros:
@chrishavlin nice catch. I also found a couple tests where that change was missing too

chrishavlin (approved):

Issue #4188, opened 2022-10-25
TST: CI instability caused by pytest 7.2


neutrinoceros:
### Bug report
**Bug summary**

Pytest 7.2, which was just released a couple hours ago, [deprecates support for tests written for nose](https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose)

this causes 14 tests to fail on every pytest job
(see https://github.com/yt-project/yt/actions/runs/3319655194)

I guess the way to go, short term, is to configure pytest to ignore these specific tests. We may need to also isolate the tests from their respective modules.
I don't have time to deal with this right away, but anyone is welcome to poke at it !

PR #4189, opened 2022-10-25
TST: resolve compatibility with pytest 7.2

neutrinoceros:
## PR Summary
Quick fix for #4188
- TST: configure pytest to ignore test modules using deprecated nose support
- TST: migrate a test from nose to pytest API

### DIFF
--- a/nose_unit.cfg
+++ b/nose_unit.cfg
@@ -6,5 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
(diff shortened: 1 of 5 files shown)


cphyc (approved):
This looks good to me, but I'd be happy to have a second pair of eyes to be sure we're not sending tests to `/dev/null`.

neutrinoceros:
I've checked everywhere `requires_module_pytest` is used and found that all affected tests are still run (`yt/tests/test_load_sample.py` and `yt/tests/test_load_archive.py`).

Issue #3487, opened 2021-08-28
Particle indices have to be re-generated every time for some datasets


chummels:
### Bug report

**Bug summary**

As discussed in the yt slack, some datasets must re-generate particle indices every time they are loaded, instead of using the existing .ewah files to skip this step.  This negates the whole purpose of generating these indices (ewah files) and can take hours to load in a dataset depending on the nature of the dataset.  

The issue arises for datasets where yt updates the refined index order to be more efficient, it generates an ewah file specifically named based on the new coarse and refined indices (e.g., ` halo_59.hdf5.index6_4.ewah`).  But when yt loads this dataset the next time, it searches for an ewah file based on the default coarse and refined indices (e.g., `halo_59.hdf5.index7_5.ewah`), so it fails to see the ewah file and thinks it needs to generate it again.

**Code for reproduction**

```python
import yt
ds = yt.load_sample('TNGHalo')
ds.index
ds = yt.load_sample('TNGHalo')
ds.index
```

**Actual outcome**

This code needs to generate the coarse index and refined index both times it loads the dataset.  If you re-run this script with a different dataset, like `FIRE_M12i_ref11`, it only needs to generate the coarse index and refined index once, and the second time, it just loads the particle index data `Loading particle index`.

chummels:
This issue was discussed in #3198 , and a solution was proposed: 

> I've now limited the heuristic, but I'm somewhat slightly concerned that what it now does is check for the old index_order in the filename, rather than the new, so it's entirely possible that if it does any modifications to the index_order, it will always always generate new ewah files.

> One possible way around this would be to have the filename just have index_order1 in it, and if it's auto-generated, have it call it "auto" or something.

chummels:
Alternatively, why do we need to list the indices in the ewah filename at all?  Once an ewah file has been generated, does it matter at all what the coarse and refined indices are?  Perhaps I'm being naive, but it seems like once it's generated, it'll just work for loading in the data.  But perhaps one will see changes in efficiency depending on the future functions applied to that dataset, so maybe it does matter?

One solution would be to just use regular expressions to see if there is any ewah file (with the same filename stem) in the same directory as the dataset, and if so, try to use that.  This would resolve the issue, I think, and be backwards compatible as well.

matthewturk:
That might be a fine solution. I'd be open to an implementation that does
that. I think we need to double check that the orders are stored in the
file itself, but I believe they are.

neutrinoceros:
closed via #4198

PR #4198, opened 2022-11-07
Automatically find EWAH files with increased index_order2

jzuhone:
## PR Summary

When loading particle datasets, a morton index is either created and stored to an EWAH file or is loaded from an already existing file. 

PR #3198 allowed for the automatic updating of the refined index to be more efficient. However, since both the coarse and the fine index are in the filename, an existing EWAH file with the updated index order is not found and thus the index must be re-generated again. 

This PR changes the logic slightly so that EWAH filenames with either the original refined index or the updated one can be found and used. 

Closes Issue #3487. 

## PR Checklist

### DIFF
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -200,7 +200,6 @@ class ParticleIndex(Index):
             if rflag == 0:
                 raise OSError
         except (OSError, struct.error):
-            print("here", dont_cache)
             self.regions.reset_bitmasks()
             self._initialize_coarse_index()
             self._initialize_refined_index()
@@ -213,7 +212,6 @@ class ParticleIndex(Index):
                 try:
                     self.regions.save_bitmasks(fname)
                 except OSError:
-                    print("bad")
                     pass
             rflag = self.regions.check_bitmasks()
 
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -177,7 +177,7 @@ class ParticleIndex(Index):
             if getattr(ds, "index_filename", None) is None:
                 irange = f"[{self.regions.index_order2}-{self.regions.index_order2+2}]"
                 fn_prefix = f"{ds.parameter_filename}.index{self.regions.index_order1}"
-                fn_glob = f"{fn_prefix}_{range}.ewah"
+                fn_glob = f"{fn_prefix}_{irange}.ewah"
                 fns = glob(fn_glob)
                 fname = (
                     f"{fn_prefix}_{self.regions.index_order2}.ewah"
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -3,6 +3,7 @@ import errno
 import os
 import struct
 import weakref
+from glob import glob
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


neutrinoceros:
about the CodeTour watch failure: it might be worth trying bumping this action (current latest is 1.6.3 https://github.com/pozil/codetour-watch/releases)

matthewturk:
@jzuhone this is *awesome* and I think will make a lot of lives easier.  (cc @chummels )

yt/geometry/particle_geometry_handler.py
```diff
@@ -174,8 +175,14 @@ def _initialize_index(self):
         # Load Morton index from file if provided
         def _current_fname():
             if getattr(ds, "index_filename", None) is None:
-                fname = ds.parameter_filename + ".index{}_{}.ewah".format(
-                    self.regions.index_order1, self.regions.index_order2
+                range = f"[{self.regions.index_order2}-{self.regions.index_order2+2}]"
```
neutrinoceros, line 178: I'd rather avoid shadowing builtins (suggesting `irange` for "index range" here, but basically any name works)
```suggestion
                irange = f"[{self.regions.index_order2}-{self.regions.index_order2+2}]"
```

neutrinoceros (commented):
I like the idea, and I assume this is hard to test automatically so I guess it's fine as is, provided it can at least be manually validated.

jzuhone:
@neutrinoceros thanks for the catch on `range`. I did do some manual testing as well.

neutrinoceros:
Note that the fix to CodeTour failures is to update https://github.com/yt-project/yt/blob/main/.tours/particle-indexing.tour to match the new line numbers

jzuhone:
Welp, I updated the line numbers in codetour and updated the hash for the commit it should use and we're still failing

matthewturk:
@jzuhone I'll take a look

matthewturk:
I'll open it up in codetour and take a look asap

jzuhone:
Some googling of the error message it prints out ("Resource not accessible by integration") suggests some kind of weird permissions problem

jzuhone:
The test failure here has something to do with a GitHub secret that apparently has a permissions issue. Not sure if I can say much more than that.

jzuhone:
Is it because I don't have the secret in my fork? I'm not sure why it would fail only on this PR then

neutrinoceros:
I don't think the problem is with your fork, but with the main repo's configuration (maybe the secret's name is incorrect ?)
We absolutely need a repo owner to look at this, but it seems to me that this workflow _never_ worked properly (apparently it's supposed to write automated comments in the discussion thread).
@matthewturk , if you're not able to fix this soon, I think we should just turn off this workflow for now rather than blocking the next bugfix release over this

Xarthisius:
Looking at what it want to paste as a comment:

**CodeTour Watch**


Changed files with possible CodeTour impact:

- `yt/geometry/particle_geometry_handler.py`

Impacted CodeTour files:

- `.tours/particle-indexing.tour`

Make sure to review CodeTour files and update line numbers accordingly.

Maybe if we do that it will go away... :)

Xarthisius:
I checked that GITHUB_TOKEN is enabled, and has write permission to issues (everything really). I don't know why it's failing...

jzuhone:
I filed an issue with the action maintainer

neutrinoceros:
issue: #4213 
and PR: https://github.com/yt-project/yt/pull/4214

neutrinoceros:
Just so this remains 100% relevant to backport, I'll remove my vain commit

neutrinoceros (approved):

PR #4208, opened 2022-11-12
TST: simplify build envs

neutrinoceros:
Cleanup some complexity that is now unneeded


matthewturk:
Why don't we need it?

neutrinoceros:
because all our optional dependencies (specifically PyQt5, cartopy and netcdf4) now ship a `pyproject.toml` file, which correctly specifies their build-time dependencies as
```toml
# pytproject.toml

[build-system]
requires = [
   # ...
]
```

so we don't need to install cython and numpy in the build env anymore

neutrinoceros:
@matthewturk are we good ?

matthewturk (approved):

PR #4211, opened 2022-11-16
[bugfix] H nuclei are being double-counted in the Arepo frontend

jzuhone:
## PR Summary

yt computes "nuclei density" fields by summing up ion number density fields for the same nuclei with different ionization states. The Arepo frontend currently has `"H"`, `"H_p0"`, and `"H_p1"` all added to the `species_names` attribute in `ArepoFieldInfo`, which is then used to compute the set up the species fields and compute the fields for specific abundances. 

The `"H_p0"` and `"H_p1"` do not belong in this list for two reasons:

1. Their respective associated fields are handled later on in the method `ArepoFieldInfo.setup_gas_particle_fields`, and do not need to be handled by `setup_species_fields`.
2. Putting them in `species_names` results in the number of hydrogen nuclei being double-counted

This PR removes those ionization species from the `species_names` attribute and also removes an alias between `H_nuclei_density` and `H_number_density` which was supposed to work (and maybe did at one point) but is no longer working. It is also unnecessary with the above fix. 
 

## PR Checklist

### DIFF
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -127,8 +127,6 @@ class ArepoFieldInfo(GadgetFieldInfo):
                     field = f"{species}{suf}"
                     self.alias(("gas", field), (ptype, field))
 
-            self.alias(("gas", "H_nuclei_density"), ("gas", "H_number_density"))
-
         if (ptype, "ElectronAbundance") in self.field_list:
 
             # If we have ElectronAbundance but not NeutralHydrogenAbundance, assume the
--- a/yt/frontends/arepo/fields.py
+++ b/yt/frontends/arepo/fields.py
@@ -87,10 +87,7 @@ class ArepoFieldInfo(GadgetFieldInfo):
 
         if (ptype, "GFM_Metals_00") in self.field_list:
             self.nuclei_names = metal_elements
-            self.species_names = ["H"]
-            if (ptype, "NeutralHydrogenAbundance") in self.field_list:
-                self.species_names += ["H_p0", "H_p1"]
-            self.species_names += metal_elements
+            self.species_names = ["H"] + metal_elements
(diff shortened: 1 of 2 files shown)


jzuhone:
@matthewturk or @neutrinoceros this is a fix which should be merged pretty quickly if possible

neutrinoceros:
This seems sound. I'd like to get a deeper look at this tomorrow to make sure I understand what's going on. In particular I want to check that removing the alias does indeed not count as a breaking change, as you suggested.

neutrinoceros:
So, I find that there's only one sample dataset that this can be tested with, namely `"TNGHalo"`
Running this script
```python
import yt

ds = yt.load_sample("TNGHalo")

ad = ds.all_data()

print(ad["gas", "H_number_density"])
print(ad["gas", "H_nuclei_density"])
```
I get
```
...
[0.06318717 0.09067829 0.1182974  ... 0.02359772 0.00351616 0.06363773] cm**(-3)
[0.12637434 0.18135657 0.2365948  ... 0.04719544 0.00703232 0.12727546] cm**(-3)
```

I see that indeed, the "alias" field `("gas", "H_nuclei_density")` is incorrectly aliased.
I can also confirm, with this minimal reproducer, that your patch fixes the bug. To my surprise, it also does _break_ my script (I was expecting  `("gas", "H_nuclei_density")` to simply become unavailable), so this patch is indeed eligible for back porting to yt 4.1.x (yay!).
Could you please add a small test case ? it doesn't have to be much more complicated that my script, changing the `print` for some `assert` statements.

neutrinoceros (approved):

jzuhone:
Grrr stupid python 3.9 test

neutrinoceros:
(Soon we'll get rid of these with #4122)

PR #4227, opened 2022-11-24
TYP: typing improvements using latest mypy (0.991)

neutrinoceros:
## PR Summary

- TYP: upgrade mypy (0.971 -> 0.991) and additional typecheck dependencies
- TYP: cleanup obsolete 'type: ignore' comments
- TYP: fix PEP 484 errors (implicit None) in function signatures

All newly detected errors are fixed here, but it's worth noting that `mypy yt` now produces a short list of warnings hinting at unannotated functions that have small bits of typing information in their bodies:
```
yt/data_objects/static_output.py: note: In member "__init__" of class "Dataset":
yt/data_objects/static_output.py:240: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
yt/data_objects/static_output.py: note: In member "_get_field_info_helper" of class "Dataset":
yt/data_objects/static_output.py:921: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
yt/data_objects/static_output.py: note: In member "_assign_unit_system" of class "Dataset":
yt/data_objects/static_output.py:1241: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
yt/visualization/plot_window.py: note: In member "__init__" of class "PWViewerMPL":
yt/visualization/plot_window.py:859: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
yt/visualization/plot_window.py:869: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
yt/visualization/plot_modifications.py: note: In member "__call__" of class "ContourCallback":
yt/visualization/plot_modifications.py:948: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
yt/visualization/plot_modifications.py:958: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
yt/visualization/plot_modifications.py:964: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
yt/utilities/command_line.py: note: In member "load_config" of class "YTConfigLocalConfigHandler":
yt/utilities/command_line.py:1332: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
Success: no issues found in 500 source files
```

[...]

PR #4229, opened 2022-11-27
TYP: add static typing to Dataset._get_field_info_helper

neutrinoceros:
## PR Summary

While working on #4227 I stumbled again on this method, which turns out to be extremely delicate to correctly type.

Because this method is private, it doesn't *need* the amount of type flexibility it currently has: being strict on input type would make it much easier to maintain, and wouldn't take anything away from users.
I'll use this PR as an experimental playground to see how much of the codebase actually falls apart if I take away type flexibility and how much work is needed to fix it. I'm hoping the answer to both questions is "not that much".

### DIFF
--- a/doc/helper_scripts/show_fields.py
+++ b/doc/helper_scripts/show_fields.py
@@ -37,7 +37,7 @@ def _strip_ftype(field):
 
 
 np.random.seed(int(0x4D3D3D3))
-units = [base_ds._get_field_info(*f).units for f in fields]
+units = [base_ds._get_field_info(f).units for f in fields]
 fields = [_strip_ftype(f) for f in fields]
 ds = fake_random_ds(16, fields=fields, units=units, particles=1)
 ds.parameters["HydroMethod"] = "streaming"
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -271,9 +271,10 @@ class YTDataContainer(abc.ABC):
             rv = self.field_data[f]
         except KeyError:
             if isinstance(f, tuple):
-                fi = self.ds._get_field_info(*f)
+                fi = self.ds._get_field_info(f)
             elif isinstance(f, bytes):
-                fi = self.ds._get_field_info("unknown", f)
+                # TODO(4229): see if this branch could be dropped
+                fi = self.ds._get_field_info(("unknown", f))
             rv = self.ds.arr(self.field_data[key], fi.units)
         return rv
 
@@ -296,7 +297,7 @@ class YTDataContainer(abc.ABC):
 
     def _generate_field(self, field):
         ftype, fname = field
-        finfo = self.ds._get_field_info(*field)
+        finfo = self.ds._get_field_info(field)
         with self._field_type_state(ftype, finfo):
             if fname in self._container_fields:
                 tr = self._generate_container_field(field)
@@ -310,8 +311,7 @@ class YTDataContainer(abc.ABC):
 
     def _generate_fluid_field(self, field):
         # First we check the validator
-        ftype, fname = field
-        finfo = self.ds._get_field_info(ftype, fname)
+        finfo = self.ds._get_field_info(field)
         if self._current_chunk is None or self._current_chunk.chunk_type != "spatial":
             gen_obj = self
         else:
@@ -326,7 +326,7 @@ class YTDataContainer(abc.ABC):
         return rv
 
     def _generate_spatial_fluid(self, field, ngz):
-        finfo = self.ds._get_field_info(*field)
+        finfo = self.ds._get_field_info(field)
         if finfo.units is None:
             raise YTSpatialFieldUnitError(field)
         units = finfo.units
@@ -383,7 +383,7 @@ class YTDataContainer(abc.ABC):
         else:
             gen_obj = self._current_chunk.objs[0]
         try:
-            finfo = self.ds._get_field_info(*field)
+            finfo = self.ds._get_field_info(field)
             finfo.check_available(gen_obj)
         except NeedsGridType as ngt_exception:
             if ngt_exception.ghost_zones != 0:
@@ -407,7 +407,7 @@ class YTDataContainer(abc.ABC):
                     ind += data.size
         else:
             with self._field_type_state(ftype, finfo, gen_obj):
-                rv = self.ds._get_field_info(*field)(gen_obj)
+                rv = self.ds._get_field_info(field)(gen_obj)
         return rv
 
     def _count_particles(self, ftype):
@@ -1540,7 +1540,7 @@ class YTDataContainer(abc.ABC):
                 continue
 
             ftype, fname = self._tupleize_field(field)
-            finfo = self.ds._get_field_info(ftype, fname)
+            finfo = self.ds._get_field_info((ftype, fname))
 
             # really ugly check to ensure that this field really does exist somewhere,
             # in some naming convention, before returning it as a possible field type
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -11,9 +11,10 @@ from yt._typing import AnyFieldKey
 from yt.config import ytcfg
 from yt.data_objects.field_data import YTFieldData
 from yt.data_objects.profiles import create_profile
+from yt.fields.derived_field import DerivedField
 from yt.fields.field_exceptions import NeedsGridType
 from yt.frontends.ytdata.utilities import save_as_dataset
-from yt.funcs import get_output_filename, is_sequence, iter_fields, mylog
+from yt.funcs import get_output_filename, iter_fields, mylog
 from yt.units._numpy_wrapper_functions import uconcatenate
 from yt.units.yt_array import YTArray, YTQuantity
 from yt.utilities.amr_kdtree.api import AMRKDTree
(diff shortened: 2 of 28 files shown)


neutrinoceros:
Turns out the situation is much worse than I thought, as this method also gets called with very different data types (I think this happens with region objects), and we might receive something like
```
ftype = (slice(None, None, None), slice(None, None, None), 0.25)
```

I don't know yet how to improve type consistency without breaking anything. Hopefully I can get back to this soon.

matthewturk:
So, it's also entirely possible that we can get rid of the `_last_freq` check now that we don't allow aliases.  It was originally put in place so that if you made two calls in rapid succession:

```
obj['some_ftype', 'some_fname']
obj['some_fname']
```

the second would implicitly use the `some_ftype`.  For instance, this was commonly an issue when just accessing things like particle fields.  But now that we don't allow aliases to be accessed without being explicit, then perhaps we don't need the complexity.

neutrinoceros:
Thank you @matthewturk for pointing this out. This is slightly orthogonal to what I initially wanted to accomplish here, but indeed I agree that we can get rid of this complexity, and this may be an opportunity to do so. I'll explore that in this PR too if it doesn't cause too much scope inflation !

neutrinoceros:
@matthewturk do you think there's any reason to keep the `_last_freq` and `_last_finfo` attributes at all (other than historical) ? I've started working on getting them out completely but I'm not sure this is what you meant to suggest.

matthewturk:
@neutrinoceros as of right now, I cannot think of any reason to retain them.  *That being said* I think if we can verify that people's workflows won't change any more than they already had to with the deprecation of aliased fields, we should be okay, but if we can't, I dunno.

neutrinoceros:
Ok. I think if I can make the transition internally without having to change tests, it should be safe to assume downstream workflows will no be affected.

It's starting to look like we can remove quite a bit of complexity without changing the observable behavior, so I'm hopeful this can be done.

neutrinoceros:
This should be pretty stable now. For reference, I've checked that this doesn't break yt_astro_analysis (the only call to `_get_field_info` there is stable).
I'll wait for #4227 to go in so I can use more specific type hints defined therein, after what this PR should be ready !

doc/helper_scripts/show_fields.py
```diff
@@ -37,7 +37,7 @@ def _strip_ftype(field):
 
 
 np.random.seed(int(0x4D3D3D3))
-units = [base_ds._get_field_info(*f).units for f in fields]
+units = [base_ds._get_field_info(f).units for f in fields]
```
matthewturk, line 40: hm, this is because it's always a tuple now right?

neutrinoceros: More exactly, it's because we don't have a second positional argument anymore. Passing a full tuple as the first argument already worked on main, I just made it the only way to do it.

matthewturk (approved):
This seems fine -- can you tell me what is holding us off from merging?

neutrinoceros:
@matthewturk nothing, it's ready to go

yt/data_objects/data_containers.py
```diff
@@ -269,10 +268,7 @@ def __getitem__(self, key):
         try:
             rv = self.field_data[f]
         except KeyError:
-            if isinstance(f, tuple):
-                fi = self.ds._get_field_info(*f)
-            elif isinstance(f, bytes):
-                fi = self.ds._get_field_info("unknown", f)
```
chrishavlin, line 275: this is interesting. any idea what the history of using a byte string here is? @matthewturk?

yt/data_objects/selection_objects/data_selection_objects.py
```diff
@@ -184,7 +184,7 @@ def get_data(self, fields=None):
         fluids, particles = [], []
         finfos = {}
         for ftype, fname in fields_to_get:
-            finfo = self.ds._get_field_info(ftype, fname)
+            finfo = self.ds._get_field_info((ftype, fname))
```
chrishavlin, line 187: could consider changing this loop so that it does not split the field tuple in the first place (i.e., `for field in fields_to_get`).

yt/frontends/ytdata/data_structures.py
```diff
@@ -629,7 +629,7 @@ def get_data(self, fields=None):
         fluids, particles = [], []
         finfos = {}
         for ftype, fname in fields_to_get:
-            finfo = self.ds._get_field_info(ftype, fname)
+            finfo = self.ds._get_field_info((ftype, fname))
```
chrishavlin, line 632: another spot where it splits the tuple only to re-assemble.

chrishavlin (commented):
Generally looks good!

My only hesitation is removing the `_last_freq` and `_last_finfo` attributes of `Dataset` -- but only because I'm having a hard time working out how these attributes were being used in the first place. The test that this PR removes (`test_fields::test_field_inference`) makes it seem like tracking the last requested field was important... any insight to this? It's the only thing in the PR that seems like it could lead to a change in behavior so I'm trying to understand it more.

neutrinoceros:
> The test that this PR removes (test_fields::test_field_inference) makes it seem like tracking the last requested field was important... any insight to this? 

My understanding of this is that it was an essential part of the "field type guessing" mechanism, which is now disabled since yt 4.1, when we started _refusing_ to guess field types if they are ambiguous. Up to yt 4.0, these attributes would serve as a poor-man's cache to avoid redoing all the guessing work when looping over ambiguous fields. But this is actually me _guessing_ (o, the irony), and I never had a hand in implementing or maintaining this specific part of the mechanism. I think it's completely useless now, but I defer to @matthewturk for confirmation.

doc/helper_scripts/show_fields.py
neutrinoceros: I have no idea, probably no good reason, but this feels just slightly out of scope here. Do you mind if I leave the line as is ?

yt/data_objects/selection_objects/data_selection_objects.py
neutrinoceros: yup, definitely. Looks like something I rewrote with a global find & replace, but you're right that the split doesn't make much sense at all now.

neutrinoceros: done

yt/frontends/ytdata/data_structures.py
neutrinoceros: done

chrishavlin:
> I think it's completely useless now, but I defer to @matthewturk for confirmation.

OK.That makes some more sense then. I'll add an approval now and let Matt have final word after you get the tests passing again.

chrishavlin (dismissed):

neutrinoceros:
Funny how 1b1b2420855de520066a26ed51bc9158d776ca84 (the cleanup commit) breaks tests that were just added in #4175 ...
I don't think I can afford just _dropping_ these tests (which of course crossed my mind for a sec), because the errors are pretty serious:
```
Failed: DID NOT RAISE <class 'yt.utilities.exceptions.YTFieldNotFound'>
```
For now I will only rebase the branch on main with no modifications (just so I have the new tests available), and I'll try to fix the situation tomorrow

mtryan83:
> Funny how 1b1b2420855de520066a26ed51bc9158d776ca84 (the cleanup commit) breaks tests that were just added in #4175 ... I don't think I can afford just _dropping_ these tests (which of course crossed my mind for a sec), because the errors are pretty serious:
> 
> ```
> Failed: DID NOT RAISE <class 'yt.utilities.exceptions.YTFieldNotFound'>
> ```
This seems to come from a change in how `_determine_fields` or `_tupleize_field` works. I'm not really sure what the intended behavior is. Prior to 1b1b242,  you'd get
```python
dd = firefly_test_dataset.all_data()
dd._determine_fields("pt2only_field")
YTFieldNotFound                           Traceback (most recent call last)
Cell In [6], line 2
      1 dd = firefly_test_dataset.all_data()
----> 2 dd._determine_fields("pt2only_field")

File ~/workspace/yt/yt/data_objects/data_containers.py:1542, in YTDataContainer._determine_fields(self, fields)
   1539     continue
   1541 ftype, fname = self._tupleize_field(field)
-> 1542 finfo = self.ds._get_field_info(ftype, fname)
   1544 # really ugly check to ensure that this field really does exist somewhere,
   1545 # in some naming convention, before returning it as a possible field type
   1546 if (
   1547     (ftype, fname) not in self.ds.field_info
   1548     and (ftype, fname) not in self.ds.field_list
   (...)
   1552     and (ftype, fname) not in self._container_fields
   1553 ):

File ~/workspace/yt/yt/data_objects/static_output.py:867, in Dataset._get_field_info(self, ftype, fname)
    866 def _get_field_info(self, ftype, fname=None):
--> 867     field_info, candidates = self._get_field_info_helper(ftype, fname)
    869     if field_info.name[1] in ("px", "py", "pz", "pdx", "pdy", "pdz"):
    870         # escape early as a bandaid solution to
    871         # https://github.com/yt-project/yt/issues/3381
    872         return field_info

File ~/workspace/yt/yt/data_objects/static_output.py:987, in Dataset._get_field_info_helper(self, ftype, fname)
    985             self._last_finfo = self.field_info[(ftype, fname)]
    986             return self._last_finfo, candidates
--> 987 raise YTFieldNotFound(field=INPUT, ds=self)

YTFieldNotFound: Could not find field ('all', 'pt2only_field') in ParticleData.
Did you mean:
	('pt2', 'pt2only_field')
```
Now you get
```python
dd = firefly_test_dataset.all_data()
dd._determine_fields("pt2only_field")

[('pt2', 'pt2only_field')]
```
Here firefly_test_dataset has two particle types (`pt1` and `pt2`) and the field `pt2only_field` is only in pt2.

neutrinoceros:
> Now you get
>```python
>dd = firefly_test_dataset.all_data()
>dd._determine_fields("pt2only_field")
>
>[('pt2', 'pt2only_field')]
>```
>Here firefly_test_dataset has two particle types (pt1 and pt2) and the field pt2only_field is only in pt2.
>
>If the new behavior is intended, the two failing tests should be removed/ignored

Indeed it looks to me that the result on this branch is exactly what I expect, so I will adjust the tests accordingly.
Thank you so much @mtryan83 for spontaneously helping me out, you just saved me a couple hours of vain head scratching, I really appreciate it.

neutrinoceros:
@mtryan83 so for now I've just dropped the broken tests entirely, but it made me realise that `match_any_particle_types` is a very confusing flag name (I know it because _I_ suggested it and I still don't really get what it's supposed to mean 🙉), so maybe we should re-iterate over that, but I'm thinking it'd be easier to do _after_ the present PR goes in, if that sounds acceptable to you.

matthewturk:
So what do we think, should this go ahead?

neutrinoceros:
Looks to me like we have green light from every body involved.

matthewturk (approved):

PR #4231, opened 2022-11-30
Fix pixel size calculation for density option in particle projection plots and FITS images

jzuhone:
## PR Summary

`ParticleProjectionPlot` and `FITSParticleProjection` have a `density` option that divides the projected map by the area of the pixel of the fixed resolution buffer to compute a projected particle density. The computation of the pixel area was being handled incorrectly--it was using the width of the underlying `data_source` for the projection instead of the `bounds` of the FRB to compute the pixel sizes. This PR fixes that. 

## PR Checklist

### DIFF
--- a/yt/visualization/fixed_resolution.py
+++ b/yt/visualization/fixed_resolution.py
@@ -771,8 +771,8 @@ class ParticleImageBuffer(FixedResolutionBuffer):
         # requested
         info = self._get_info(item)
         if density:
-            dpx = (self.bounds[1] - self.bounds[0]) / self.buff_size[0]
-            dpy = (self.bounds[3] - self.bounds[2]) / self.buff_size[1]
+            dpx = (bounds[1] - bounds[0]) / self.buff_size[0]
+            dpy = (bounds[3] - bounds[2]) / self.buff_size[1]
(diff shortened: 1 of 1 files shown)


yt/visualization/fixed_resolution.py
```diff
@@ -771,9 +771,9 @@ def __getitem__(self, item):
         # requested
         info = self._get_info(item)
         if density:
-            width = self.data_source.width
-            norm = width[self.xax] * width[self.yax] / np.prod(self.buff_size)
-            norm = norm.in_base()
+            dpx = (self.bounds[1] - self.bounds[0]) / self.buff_size[0]
```
matthewturk, line 774: Are these guaranteed to be in code length?  Or will they carry units?

matthewturk:
*Great* catch.

jzuhone:
pre-commit.ci autofix

yt/visualization/fixed_resolution.py
jzuhone: @matthewturk now you have the good catch--see https://github.com/yt-project/yt/pull/4231/commits/867db30737dc04a40ceb0cfeed66ddf222969904

matthewturk (approved):

PR #4242, opened 2022-12-08
STY: upgrade flake8-bugbear and fix newly detected errors

neutrinoceros:
## PR Summary
Opening as a draft because it's not obvious that adding `@abstractmethod` decorators here and there will not crash tests

TODO:
- [x] survive CI (use noqa comments when making empty methods abstract breaks things)
- [x] extend checks to B902 + fix
- [x] extend checks to B904 + fix


matthewturk (approved):
seems good and is no-op, right?

PR #4245, opened 2022-12-11
BLD: fix missing specs from minimal requirements

neutrinoceros:
## PR Summary

Just spotted these two out-of-sync requirements from install_requires.


matthewturk (approved):

neutrinoceros:
I'm not convinced this is worth the effort of manually backporting. I'll just leave it at that unless anyone complains about it.

PR #4249, opened 2022-12-15
A bugfix for the new expanded EWAH file reading capability

jzuhone:
## PR Summary

In PR #4198 I implemented a feature to automatically find EWAH files with increased `index_order2` for SPH/Lagrangian datasets. 

There was an oversight in these changes, which involves the fact that although the file with the updated `index_order2` is correctly found and loaded, `index_order2` is not increased by 2 when creating the `ParticleBitmap` object. The result is that for some analyses that try to look on small scales the code either crashes or it gives odd-looking results. 

This PR fixes this issue, but it required a bit of moving things around so that we detect the new value of `index_order2` from the filename before we create the `ParticleBitmap` object.

## PR Checklist

### DIFF
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -162,6 +162,24 @@ class ParticleIndex(Index):
         if not hasattr(self.ds, "_file_hash"):
             self.ds._file_hash = self._generate_hash()
 
+        # Load Morton index from file if provided
+        def _current_fname(order1, order2):
+            if getattr(ds, "index_filename", None) is None:
+                irange = f"[{order2}-{order2+2}]"
+                fn_prefix = f"{ds.parameter_filename}.index{order1}"
+                fn_glob = f"{fn_prefix}_{irange}.ewah"
+                fns = glob(fn_glob)
+                fname = f"{fn_prefix}_{order2}.ewah" if len(fns) == 0 else fns[-1]
+            else:
+                fname = ds.index_filename
+            new_order2 = int(fname.split("_")[-1].split(".")[0])
+            return fname, new_order2
+
+        # We set order2 in case we loaded an index filename
+        # where the order2 is +2 the one which is assumed at
+        # first
+        fname, order2 = _current_fname(order1, order2)
+
         self.regions = ParticleBitmap(
             ds.domain_left_edge,
             ds.domain_right_edge,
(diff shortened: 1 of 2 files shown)


neutrinoceros:
I probably made a mistake in including the previous PR in a bugfix release. What’s the preferred approach for the backport branch 4.1.x now ? Should we include the fix and hope it’s stable or just revert the original patch there and rerelease everything in 4.2?

jzuhone:
> What’s the preferred approach for the backport branch 4.1.x now ? Should we include the fix and hope it’s stable or just revert the original patch there and rerelease everything in 4.2?

This is a good question for @matthewturk. Either way, this is a critical bugfix so we should fix it somehow in 4.1.3.

neutrinoceros:
> Either way, this is a critical bugfix so we should fix it somehow in 4.1.3.

Absolutely

jzuhone:
@neutrinoceros actually it's probably easier to just revert the previous change for now so we can think about this more carefully

neutrinoceros:
>  it's probably easier to just revert the previous change for now so we can think about this more carefully

done in https://github.com/yt-project/yt/pull/4250

yt/geometry/particle_geometry_handler.py
```diff
@@ -162,6 +162,25 @@ def _initialize_index(self):
         if not hasattr(self.ds, "_file_hash"):
             self.ds._file_hash = self._generate_hash()
 
+        # Load Morton index from file if provided
+        def _current_fname(order1, order2):
+            if getattr(ds, "index_filename", None) is None:
+                irange = f"[{order2}-{order2+2}]"
```
matthewturk, line 168: It took me a moment to figure out we were creating an fstring for a glob, but this makes sense.

matthewturk (approved):

neutrinoceros:
@matthewturk what do you think about reverting the feature from the 4.1 branch ?

matthewturk:
We did kinda make a big deal of it in the release notes.  So if we can *fix* it, we should, don't you agree?

neutrinoceros:
Sure. I honestly don't hold strong opinions here. This was my mistake, but I'll settle for any solution you guys can agree on

matthewturk:
@jzuhone I'm OK with this going in and getting backported.  What're your thoughts?

neutrinoceros:
@matthewturk seems like you merged the revert a couple minutes after your last comment. It looks like a mistake on your side this time, but maybe we should actually leave the backport branch alone now, unless anyone feels strongly that we shouldn't ?

matthewturk:
I will not touch anything.

neutrinoceros:
Then if @jzuhone agrees I vote for just merging this with no backport, and let future me deal with explaining the revert in the 4.1.3 release notes :-)

jzuhone:
@neutrinoceros sounds good to me

PR #4251, opened 2022-12-17
TYP: add missing type hints

neutrinoceros:
## PR Summary

Complements #4229 in fixing the remaining warnings from mypy (`[annotation-unchecked]`)

```
yt/data_objects/static_output.py: note: In member "_assign_unit_system" of class "Dataset":
yt/data_objects/static_output.py:1254: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
... 
```

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -12,11 +12,23 @@ from collections import defaultdict
 from functools import cached_property
 from importlib.util import find_spec
 from stat import ST_CTIME
-from typing import Any, DefaultDict, Dict, List, Optional, Set, Tuple, Type, Union
+from typing import (
+    Any,
+    DefaultDict,
+    Dict,
+    List,
+    Literal,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    Union,
+)
 
 import numpy as np
 from more_itertools import unzip
-from unyt import Unit
+from sympy import Symbol
+from unyt import Unit, unyt_quantity
 from unyt.exceptions import UnitConversionError, UnitParseError
 
 from yt._maintenance.deprecation import issue_deprecation_warning
@@ -221,7 +233,7 @@ class Dataset(abc.ABC):
         filename: str,
         dataset_type: Optional[str] = None,
         units_override: Optional[Dict[str, str]] = None,
-        unit_system: str = "cgs",
+        unit_system: Literal["cgs", "mks", "code"] = "cgs",
         default_species_fields: Optional[
             "Any"
         ] = None,  # Any used as a placeholder here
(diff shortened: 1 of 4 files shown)


matthewturk (approved):

yt/data_objects/static_output.py
```diff
@@ -221,7 +233,7 @@ def __init__(
         filename: str,
         dataset_type: Optional[str] = None,
         units_override: Optional[Dict[str, str]] = None,
-        unit_system: str = "cgs",
+        unit_system: Literal["cgs", "mks", "code"] = "cgs",
```
matthewturk, line 236: as a note: are these really the only ones we allow?  I'm not sure that's the case.

matthewturk (commented):
I want some clarification on the unit system stuff

yt/data_objects/static_output.py
neutrinoceros: from what I can tell we actually rely on `unyt` to define `unyt.unit_systems.unit_system_registry`, and the value used here needs to be a (case insensitive) key of this dictionary (unless it's `"code"`). At runtime, this dict has the following keys:
`dict_keys(['cgs', 'mks', 'imperial', 'galactic', 'solar', 'geometrized', 'planck'])`

so you're right that there are more than 3 valid options. Unfortunately we can't define a `Literal` type programmatically which means our options are as follow:
- expand this type by hardcoding every current options, at the risk of it getting out of sync whenever unyt adds or removes one.
- get back to a more generic type (`str`), which is more future proof but doesn't convey much information.

I'd go with the first option because I don't think it's very likely that we change those in unyt any time soon, but I can settle for the second one too !

matthewturk: Let's do option 1

neutrinoceros: done !

yt/data_objects/static_output.py
```diff
@@ -221,7 +233,17 @@ def __init__(
         filename: str,
         dataset_type: Optional[str] = None,
         units_override: Optional[Dict[str, str]] = None,
-        unit_system: str = "cgs",
+        # valid unit_system values include all keys from unyt.unit_systems.unit_systems_registry + "code"
+        unit_system: Literal[
```
matthewturk, line 237: I don't want to balloon this PR.  But: would it be worthwhile to eventually have this somewhere else, so that we can reference it in other places that we might use it?  For instance, in the various `load`s?

matthewturk (approved):
Good to go in as-is!

yt/data_objects/static_output.py
neutrinoceros: yes, that would make sense

PR #4253, opened 2022-12-17
TST: speedup test_callbacks (⏰ wait for #4122)

neutrinoceros:
## PR Summary
This test module takes about 10% of the total test time on Jenkins by my estimate (about 5min on my laptop, and usually the whole suite runs in ~45 on Jnkins), and there are a couple of ways we can save time:
- avoid saving image files that we don't even look at
- reduce buffer size
- prefer SlicePlot over ProjectionPlot
- avoid reloading the same dataset many many times
- avoid using real data, which would also greatly improve portability (currently running all these tests locally requires having the data on disk and h5Py installed)

I'll attempt to do that more or less systematically here.

### DIFF
--- a/yt/visualization/tests/test_callbacks.py
+++ b/yt/visualization/tests/test_callbacks.py
@@ -710,7 +710,8 @@ def test_contour_callback():
         text_args={"fmt": "%1.1f"},
     )
     slc.render()
-    check_axis_manipulation(slc)
+    # deactivated because it makes the test case 7 times as long (10s to 70s)
+    # check_axis_manipulation(slc)
 
     ds = fake_amr_ds(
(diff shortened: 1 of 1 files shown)


neutrinoceros:
I've reduced the runtime of this module from 5min to 2min on my laptop but it doesn't seem to save anything on Jenkins. Eventually these test could be refactored into the actual image tests which would improve load balance between the test workers and make them either to debug, but this will wait until the transition to pytest-mpl actually gets started.

neutrinoceros:
I can forsee this drifting into limbo and becoming hard to pick up again when it's finally possible to convert tests to pytest_mpl. I'd rather close it now and make room in the backlog.

PR #4255, opened 2022-12-19
add some missing test for stream grids with callables

chrishavlin:
This PR adds a test for calling `load_uniform_grid` with data specified via a function. It also moves the whole `test_callable_grids.py` suite to pytest-only.

### DIFF
--- a/nose_unit.cfg
+++ b/nose_unit.cfg
@@ -6,5 +6,5 @@ nologcapture=1
 verbosity=2
 where=yt
 with-timer=1
-ignore-files=(test_load_errors.py|test_load_sample.py|test_commons.py|test_ambiguous_fields.py|test_field_access_pytest.py|test_save.py|test_line_annotation_unit.py|test_eps_writer.py|test_registration.py|test_invalid_origin.py|test_outputs_pytest\.py|test_normal_plot_api\.py|test_load_archive\.py|test_stream_particles\.py|test_file_sanitizer\.py|test_version\.py|\test_on_demand_imports\.py|test_set_zlim\.py|test_add_field\.py|test_glue\.py|test_geometries\.py)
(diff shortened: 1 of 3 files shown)


matthewturk (approved):

neutrinoceros:
It seems that the errors we're seeing on the image tests job are not related to the known-flakiness (usually it's 7 failures or 0, here we got 4). They are probably due to the recent release of numpy 1.24.0, and they also appear to be so subtle the diff tool doesn't produce anything useful. All that to say: it should be safe to bump. @chrishavlin if you're up for it, can you do that ? Otherwise I can take care of it tomorrow.

neutrinoceros:
Actually I went ahead and opened the PR on the answer-store repo: https://github.com/yt-project/answer-store/pull/34

neutrinoceros:
for reference: blocked by https://github.com/yt-project/yt/issues/4257

neutrinoceros:
Since you opened this after commenting https://github.com/yt-project/yt/pull/4244#discussion_r1052564685, I suppose it'd be useful to define a type for the expected function signature while we're at it
Something like
```python
# yt._typing.py
from typing import Callable
import numpy as np

FieldName = str
FieldFunction = Callable[[Grid, FieldName], np.ndarray]
```

(I'm just not sure what's the best way to define a `Grid` type here, maybe a `typing.Protocol` ? or simply just use `YTSelectionContainer` ?)

neutrinoceros:
@chrishavlin seems like re-running tests doesn't help even after the issue was resolved on the main branch. You'll need to either rebase or merge from main

chrishavlin:
>  You'll need to either rebase or merge from main

Ok! Will do!

>  it'd be useful to define a type for the expected function signature while we're at it .... 

Sure! Also, if it's easier in any way, it's fine if this PR waits until after 4244 or if this gets merged into 4244 instead of main. It seemed like adding the test was a quick standalone thing, but it's also not a critical change. 

> or simply just use YTSelectionContainer ?

I think this one... if we wanted to go a little more specific, maybe `AMRGridPatch` ?

neutrinoceros:
> Sure! Also, if it's easier in any way, it's fine if this PR waits until after 4244 or if this gets merged into 4244 instead of main. It seemed like adding the test was a quick standalone thing, but it's also not a critical change.

I think it's actually easier to just merge this PR as its own thing. The type hint can be added either here or in my PR, whatever you like best.

> or simply just use YTSelectionContainer ?
I think this one... if we wanted to go a little more specific, maybe AMRGridPatch ?

Sure, I was mostly throwing ideas there, but I haven't studied this deep enough to have an actual opinion, so whatever sounds best to you is fine by me !

chrishavlin:
Ok, let's keep this PR to just adding the tests then. I'll rebase and push in a moment. And then I'll look into the type hinting -- if it ends up simple then I think it makes sense to add to 4244, if it is more complicated than expected it might be better as a follow up after 4244. In either case I'll continue this thread over if 4244...

PR #4262, opened 2022-12-23
EXP: don't call _setup_plots() method inside invalidate_figure decorator

neutrinoceros:
## PR Summary

This feels out of place. Assuming there's a "good" reason for this function call (maybe _not_ calling it will break something), it should at least bear a comment.
Let's run all tests without it and see what it gets us.

### DIFF
--- a/yt/visualization/_commons.py
+++ b/yt/visualization/_commons.py
@@ -144,7 +144,6 @@ def invalidate_figure(f):
             self.plots[field].figure = None
             self.plots[field].axes = None
             self.plots[field].cax = None
-        # self._setup_plots()
         return retv
 
     return newfunc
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Looks like the only thing broken by deactivating the line is cartopy integration, which to me sounds like improper separation of responsibility and overly tight coupling. I'll try to refactor this to a more sane state.

neutrinoceros:
This is not the easy patch I envisioned. I think the `@validate_*` and `@invalidate_*` decorators are fragile and could use some rethinking; they reduce code duplication but make the data flow hard to follow, and I'm not convinced the trade off is worth it in hindsight, but I'm also not convinced that piling-up small patches is going to help long term, so I'll close this PR.

PR #4265, opened 2022-12-24
ENH: delay expensive imports

neutrinoceros:
## PR Summary
Reduces yt's startup time by ~30% (from 1.57s to 1.07s on my machine, averaged over 100 runs)

Measured with this simple script
```python
from time import monotonic_ns

tstart = monotonic_ns()
import yt
tstop = monotonic_ns()
print(f"{(tstop-tstart)/1e9:.3f}")                                                                      
```
diff

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -2,13 +2,11 @@ import sys
 import warnings
 from abc import ABC
 from io import BytesIO
-from typing import Optional, Tuple, Union
+from typing import TYPE_CHECKING, Optional, Tuple, Union
 
 import matplotlib
 import numpy as np
-from matplotlib.axis import Axis
 from matplotlib.colors import LogNorm, Normalize, SymLogNorm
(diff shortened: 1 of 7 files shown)


neutrinoceros:
Getting some errors at collection time in minimal env tests. I don't immediately get what's happening so I'll switch to draft for now.

neutrinoceros:
I found that urllib.request, which is probably never actually used within yt, costs about 1% (and 2.5% with incoming improvements in our other dependencies) of startup overhead. It's definitely in the realm of fine tuning but it's basically free to implement, so I went for it too.

yt/visualization/base_plot_types.py
```diff
@@ -23,6 +21,10 @@
     validate_image_name,
 )
 
+if TYPE_CHECKING:
```
matthewturk, line 24: this is a good thing to know about!

matthewturk (approved):

PR #4266, opened 2022-12-24
TST: add missing @requires_module("h5py") decorators

neutrinoceros:
## PR Summary
facilitate running tests with tests files installed but not h5py


matthewturk:
Reading this makes me think we should instead make `@requires_file` either know the format (i.e., chosen from a list of existing datasets) or have it accept an argument that includes the necessary modules.

neutrinoceros:
This is a good idea, but I would rather avoid adding complexity to any function in the `yt.testing` namespace as long as we haven't migrated out of nose completely.

matthewturk (approved):

PR #4267, opened 2022-12-25
ENH: uniformize YTException class hierarchy

neutrinoceros:
## PR Summary

Fix a collection of minor, related design issues:

- all `YTXxxxxxxError` class derive from `YTException`
- fix `super()` usage
- simplify base class design so that it contains *no* init code

### DIFF
--- a/yt/utilities/exceptions.py
+++ b/yt/utilities/exceptions.py
@@ -6,9 +6,7 @@ from unyt.exceptions import UnitOperationError
 
 
 class YTException(Exception):
-    def __init__(self, message=None, ds=None):
-        Exception.__init__(self, message)
-        self.ds = ds
+    pass
 
 
 # Data access exceptions:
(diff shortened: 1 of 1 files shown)


matthewturk (commented):
So in doing this, we're never going to call `super()` of `__init__`?  What happens if we end up wanting to do this in the future?  I think as long as this is being done, we should include potentially-non-operative `super` calls in all the inits.

neutrinoceros:
It depends what design we want to push, and *why* we're using inheritance here.
The current state of affairs is that most our custom exceptions derive from `YTException` presumably so that user code may have a way to catch any error coming from yt as `except YTException`. That is one reason to use inheritance. But using `super()` calls implies that all involved classes are designed to work together, which isn't the case at the moment, so in my opinion it's clearer to avoid `super()` if it doesn't fit the *current* design. I also do not see any reason why we might change it in the future but even if we do, we can introduce `super` back as the need arises.

matthewturk:
Alright, I'm OK with this.

matthewturk (approved):

Issue #4270, opened 2022-12-27
ENH: support per-frontend extra requirements with pip


neutrinoceros:
Most users only need a single frontend, but many frontends have optional dependencies (`h5py`, `netcdf4`, `f90nml` ...), which are not exactly documented. The lazy-import mechanism helps discovering missing bits, but only at runtime.

It'd be cool to support installing yt with all optional requirements approriate for a specific frontend, for instance
```
python -m pip install "yt[enzo]"
```
would install `h5py` directly.

### Caveats (and solutions)
A challenge with doing this is to avoid duplicating constraints as much as possible.
Leveraging pip's support for recursive dependencies (unfortunately poorly documented, but known via https://hynek.me/articles/python-recursive-optional-dependencies/), repetitions can be avoided. Here's a example of how it might work

```toml
[project.optional-dependencies]
# generic requirements
HDF5 = ["h5py>=3.0"]
fortran = ["f90nml>=1.1.2"]

# frontend-specific requirements
enzo = ["yt[HDF5]"]
owls = ["yt[HDF5]"]
ramses = ["yt[fortran]"]

# refactor existing "full" target
full = ["yt[enzo, owls, ramses]"]
```

Note that I'm using `pyproject.toml`'s format, while currently our metadata lives in `setup.cfg`. Recursive dependencies are also available in `setup.cfg` but it's destined to be deprecated (see https://github.com/pypa/setuptools/issues/3214), so might as well migrate to `pyproject.toml` while we are at it.

Also note that recursive dependencies require `pip>=21.2`, so it doesn't work out of the box for Python versions with a bundled pip version older than this. For reference, here's a manually collected sample dataset

| Python version | bundled pip version |
| :---------------: | :-------------------: |  
| 3.8.0 | 19.2.3 |
| 3.8.13 | 22.0.4 |
| 3.9.0 | 20.2.3 |
| 3.9.7 | 21.2.3 |
| 3.10.0 | 21.2.3 |

So in order to support this for any patch version of Python 3.8 and 3.9, we'd need to specifically instruct upgrading `pip` itself as part of the installation process, and this extra step can be dropped when we stop supporting versions older than `3.9.7` (so, basically, when we drop Python 3.9, most likely in 2024 at the current pace)

matthewturk:
Would this affect conda-based installs?

neutrinoceros:
No that wouldn't change anything for conda. Also, and to my knowledge, there is nothing similar in the conda world, which is why I left it out.

[...]

PR #4271, opened 2022-12-27
BLD: migrate static metadata to pyproject.toml

neutrinoceros:
## PR Summary

- `setup.cfg` is targetted for deprecation https://github.com/pypa/setuptools/issues/3214
- `setuptools`' support for `pyproject.toml` is out of beta

Follow up to #3078, and prep for #4270

The heavy lifting is done by `ini2toml`. The only effective change is a bump in the minimal requirement for setuptools' version at build time.


pyproject.toml
```diff
@@ -10,6 +10,151 @@ requires = [
...
+text = "BSD 3-Clause"
+
+[project.urls]
+Homepage = "https://yt-project.org/"
+Documentation = "https://yt-project.org/doc/"
+Source = "https://github.com/yt-project/yt/"
+Tracker = "https://github.com/yt-project/yt/issues"
+
+[project.entry-points."nose.plugins.0.10"]
+answer-testing = "yt.utilities.answer_testing.framework:AnswerTesting"
```
neutrinoceros, line 73: it's not clear yet that this auto-migration will work. I'm hoping CI doesn't rely on it because it needs to be dropped or migrated to pytest eventually. If nothing break in the first run I'll try again with this section completely removed to see how things go.

neutrinoceros: Looks like it's used only on GHA for answer tests, so I'll keep it in this PR but it can be removed when #4122 goes in

matthewturk (approved):
LGTM.

PR #4272, opened 2022-12-28
ENH: add support for installing yt with cherry-picked, frontend specific optional dependencies with pip

### LINKED ISSUE #4270: ENH: support per-frontend extra requirements with pip

neutrinoceros:
## PR Summary
Fix #4270
I still need to add documentation, but I'm opening as a draft to see how it goes with CI

## PR Checklist

### DIFF
--- a/doc/source/installing.rst
+++ b/doc/source/installing.rst
@@ -85,7 +85,7 @@ Additional requirements (pip only)
 
 yt knows about several data file formats. In many cases (e.g. HDF5), additional
 dependencies are needed to enable parsing, that are not installed with yt by default.
-In order to install every requirements for a specific format alongside yt,
+In order to install all required packages for a specific format alongside yt,
 one can specify them as, for instance
 
 .. code-block:: bash
--- a/doc/source/installing.rst
+++ b/doc/source/installing.rst
@@ -65,7 +65,7 @@ will work best for you depends on your precise situation:
 Installing a stable release
 +++++++++++++++++++++++++++
 
-The latest stable release can be obtained from Pypi with pip
+The latest stable release can be obtained from PyPI with pip
 
 .. code-block:: bash
 
@@ -78,6 +78,23 @@ Or using the Anaconda/Miniconda Python distributions
 .. code-block:: bash
 
   $ conda install --channel conda-forge yt
+
+
+Additional requirements (pip only)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+yt knows about several data file formats. In many cases (e.g. HDF5), additional
+dependencies are needed to enable parsing, that are not installed with yt by default.
+In order to install every requirements for a specific format alongside yt,
+one can specify them as, for instance
+
+.. code-block:: bash
+
+  $ python -m pip install --upgrade pip
+  $ python -m pip install --user "yt[ramses]"
+
+Extra requirements can be combined, separated by commas (say ``yt[ramses,enzo_e]``).
+Note that all format names are normalized to lower case.
 
 
 .. _install-from-source:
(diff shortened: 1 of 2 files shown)


matthewturk (commented):
I think this looks really well done.  I think we should also have the jupyter widgets and `widgyts` stuff in there as an option some where, and eventually napari.

neutrinoceros:
sure ! what are the requirements for this, and would you like to suggest a target name (maybe `yt[jupyter]` fits ?) ?

doc/source/installing.rst
```diff
@@ -80,6 +80,23 @@ Or using the Anaconda/Miniconda Python distributions
   $ conda install --channel conda-forge yt
 
 
+Additional requirements (pip only)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+yt knows about several data file formats. In many cases (e.g. HDF5), additional
+dependencies are needed to enable parsing, that are not installed with yt by default.
+In order to install every requirements for a specific format alongside yt,
```
matthewturk, line 88:
```suggestion
In order to install all required packages for a specific format alongside yt,
```

doc/source/installing.rst
```diff
@@ -80,6 +80,23 @@ Or using the Anaconda/Miniconda Python distributions
...
+one can specify them as, for instance
+
+.. code-block:: bash
+
+  $ python -m pip install --upgrade pip
+  $ python -m pip install --user "yt[ramses]"
+
+Extra requirements can be combined, separated by commas (say ``yt[ramses,enzo_e]``).
+Note that all format names are normalized to lower case.
```
matthewturk, line 97: So format names are normalized, but `Fortran` and `HDF5` aren't, right?

neutrinoceros: Right. I'm using the word "format" here to avoid too specific nomenclature with frontends and dataset classes. I don't really intend `HDF5` and `Fortan` target as user exposed (they are, of course, but I'm intentionally not advertising them in docs)

matthewturk (approved):

PR #4273, opened 2022-12-28
MNT: re-add lost comments from setup.cfg

neutrinoceros:
## PR Summary
In #4271 I forgot to manually migrate comments contained in `setup.cfg`. This fixes that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,15 +44,15 @@ requires-python = ">=3.8"
 dependencies = [
     "cmyt>=1.1.2",
     "ipywidgets>=8.0.0",
-    "matplotlib!=3.4.2,>=3.2",
+    "matplotlib!=3.4.2,>=3.2", # keep in sync with tests/windows_conda_requirements.txt
     "more-itertools>=8.4",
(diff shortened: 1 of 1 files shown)


matthewturk (commented):
I seem to remember @chrishavlin having had some issue with this.  Or am I misremembering?

neutrinoceros:
issues with what ?

matthewturk:
My memory is vague, but something about comments needing to be on different lines.

neutrinoceros:
Ah, maybe I see what you mean. Before migrating to `pyproject.toml` we actually had a minor issue with formatting because we were using `setup-cfg-fmt` (a formatter dedicated to `setup.cfg`), but it doesn't know how to preserve comments _unless_ they are on the same line as actual data, which is why we adopted this format previously. It's true that the constraint is gone now so we can indeed move comments to separate lines if anyone cares. Let me know !

matthewturk (dismissed):

neutrinoceros:
rebased to resolve conflicts

matthewturk (approved):

PR #4277, opened 2022-12-30
TYP: fix mypy warnings about unchecked hints

neutrinoceros:
## PR Summary

fix current warning from mypy:
```
yt/data_objects/data_containers.py: note: In member "__init__" of class "YTDataContainer":
8
yt/data_objects/data_containers.py:91: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
```

It'd be useful to turn these notes into proper errors to avoid leaking them constantly, I'll look into it.

### DIFF
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -2,7 +2,7 @@ import abc
 import weakref
 from collections import defaultdict
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


neutrinoceros:
I found no way to configure mypy to treat those as errors. I may report that as an issue later but in the mean time this is ready for review

matthewturk (approved):

PR #4281, opened 2023-01-02
BUG: fix a regression in yt.load where '/Users/robcleme' and '~' path weren't expanded

### LINKED ISSUE #4280: BUG: yt.load("~...") doesn't work anymore
### Bug report

**Bug summary**

up until yt 4.0.5, this used to work

```python
import yt

yt.load("~/path/to/my/data")
```
in yt 4.1.x it doesn't (because the `~` is not expanded).

Without running a bisection it seems pretty clear that the breaking change was in #3525, and I can say it wasn't intentional. Luckily the fix is trivial.

neutrinoceros:
fix #4280

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -85,7 +85,7 @@ def load(
     yt.utilities.exceptions.YTAmbiguousDataType
         If the data format matches more than one class of similar specilization levels.
     """
-    fn = os.fspath(fn)
+    fn = os.path.expanduser(fn)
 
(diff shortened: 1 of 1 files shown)


matthewturk (approved):

neutrinoceros:
@matthewturk Sorry I dismissed your approval when I pushed again after realising that `os.path.expanduser(os.fspath(...))` was redundant (`os.fspath(...)` is already called in `os.path.expanduser`)

matthewturk (approved):

PR #4284, opened 2023-01-08
DOC: (devguide) improvements to "How to build a frontend" docs

neutrinoceros:
Follow up to #4272

### DIFF
--- a/doc/source/developing/creating_frontend.rst
+++ b/doc/source/developing/creating_frontend.rst
@@ -170,7 +170,7 @@ example of how this is implemented in the FLASH frontend:
 
     class FLASHFieldInfo(FieldInfoContainer):
         known_other_fields = (
-            ...("magx", (b_units, [], "B_x")),  # Note there is no alias here
+            ("magx", (b_units, [], "B_x")),  # Note there is no alias here
             ("magy", (b_units, [], "B_y")),
             ("magz", (b_units, [], "B_z")),
             ...,
--- a/doc/source/developing/creating_frontend.rst
+++ b/doc/source/developing/creating_frontend.rst
@@ -3,10 +3,6 @@
 Creating A New Code Frontend
 ============================
 
-.. warning: This section is not yet updated to work with yt 3.0.  If you
-            have a question about making a custom derived quantity, please
-            contact the mailing list.
-
 yt is designed to support analysis and visualization of data from
 multiple different simulation codes. For a list of codes and the level
 of support they enjoy, see :ref:`code-support`.
--- a/doc/source/developing/creating_frontend.rst
+++ b/doc/source/developing/creating_frontend.rst
@@ -335,3 +335,6 @@ needs to be delayed until actually needed. This is achieved by importing a wrapp
 
 Such import statements can live at the top of a module without generating overhead or errors
 in case the actual package isn't installed.
+
+If the extra third party dependency is new, a new import wrapper must also be added. To do so, 
+follow the example of the existing wrappers in ``yt.utilities.on_demand_imports.py``.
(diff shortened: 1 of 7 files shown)


neutrinoceros:
switching to draft for now because I'd like to revisit this after #4285 goes in

chrishavlin:
sure! happy to take a look.

doc/source/developing/creating_frontend.rst
```diff
@@ -317,3 +317,21 @@ And that just about covers it. Please feel free to email
...
+At runtime, extra third party dependencies should be loaded lazily, meaning their import
+needs to be delayed until actually needed. This is achieved by importing a wrapper from
+``yt.utitilies.on_demand_imports.py``, instead of the actual package like so
+
+.. code-block:: python
+
+    from yt.utilities.on_demand_imports import _mypackage as mypackage
+
+Such import statements can live at the top of a module without generating overhead or errors
+in case the actual package isn't installed.
```
chrishavlin, line 337:
```suggestion
in case the actual package isn't installed.

If the extra third party dependency is new, a new import wrapper must also be added. To do so, 
follow the example of the existing wrappers in ``yt.utilities.on_demand_imports.py``.
```

Maybe this sentence should go earlier, but I thought it was worth explicitly mentioning (even though it **should** be obvious that you can't import `_mypackage` if it doesn't exist yet)... 

I think the existing wrappers should be clear enough for someone to use as examples, but if you felt like it, it wouldn't hurt to explicitly document how to write a new on demand wrapper (inherit from `OnDemand`, add `safe_import` functions for any submodules that you will need, etc.).

yt/frontends/amrvac/fields.py
```diff
@@ -162,6 +161,7 @@ def dust_to_gas_ratio(field, data):
             )
 
     def setup_fluid_fields(self):
+        from yt.fields.magnetic_field import setup_magnetic_field_aliases
```
chrishavlin, line 164: what's the reason for moving the import statements (here and in the other frontend fields.py files)? Doesn't seem related to the focus of the PR. That said, it's such a small change that I don't mind tacking it on here, I'd just like to understand the change a bit more...

doc/source/developing/creating_frontend.rst
```diff
@@ -28,6 +24,10 @@ be classified into a couple categories:
    either all data in a region or a subset of that data.
 
 
+Note that a frontend can be built as an external package. This is useful to
+develop and maintain a maturing frontend at your own pace. For technical details, see
```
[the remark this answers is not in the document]
neutrinoceros: oh absolutely. In fact I wanted to do exactly that and ... forgot, so it seemed ?

yt/frontends/amrvac/fields.py
neutrinoceros: I took the opportunity to go thought [the whole page about frontend development](https://yt-project.org/docs/dev/developing/creating_frontend.html) and found [this paragraph](https://yt-project.org/docs/dev/developing/creating_frontend.html#creating-aliases-for-magnetic-fields), and most importantly this sentence:
> This function should always be imported and called from within the setup_fluid_fields method of the FieldInfoContainer.

So TBH I don't know _why_ this is the recommendation, but I noticed a misalignment with the reality of existing frontends and picked the easiest way to make them aligned again.

chrishavlin: Oh interesting. I wonder why. in any case, makes sense now, thanks!

doc/source/developing/creating_frontend.rst
neutrinoceros: I think your suggestion covers it satisfyingly, and has the great advantage of not easily getting out-of-sync with the code !

doc/source/developing/creating_frontend.rst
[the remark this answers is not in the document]
neutrinoceros: done !

chrishavlin (approved):

PR #4285, opened 2023-01-08
ENH: automatically load external frontends in load()

Xarthisius:
## PR Summary

This PR modifies the behavior of `yt.load` to auto-include subclasses of `Dataset` from external, installed packages. External packages must define an entry point in `yt.frontends` linked to the Dataset subclass (see docs for an example).

## PR Checklist

### DIFF
--- a/doc/source/developing/extensions.rst
+++ b/doc/source/developing/extensions.rst
@@ -40,7 +40,10 @@ Frontend as an extension
 
 Starting with version 4.2 of yt, any externally installed package that exports
 :class:`~yt.data_objects.static_output.Dataset` subclass as an entrypoint in
-``yt.frontends`` namespace in ``setup.py``:
+``yt.frontends`` namespace in ``setup.py`` or ``pyproject.toml`` will be
+automatically loaded and immediately available in :func:`~yt.loaders.load`.
+
+To add an entrypoint in an external project's ``setup.py``:
 
 .. code-block:: python
 
@@ -62,8 +65,6 @@ or ``pyproject.toml``:
    myFrontend = "my_frontend.api:MyFrontendDataset"
    myOtherFrontend = "my_frontend.api:MyOtherFrontendDataset"
 
-will be automatically loaded and immediately available in
-:func:`~yt.loaders.load`.
 
 Extension Template
 ------------------
--- a/doc/source/developing/extensions.rst
+++ b/doc/source/developing/extensions.rst
@@ -35,6 +35,36 @@ In subsequent versions, we plan to include in yt a catalog of known extensions
 and where to find them; this will put discoverability directly into the code
 base.
 
+Frontend as an extension
+------------------------
+
+Starting with version 4.2 of yt, any externally installed package that exports
+:class:`~yt.data_objects.static_output.Dataset` subclass as an entrypoint in
+``yt.frontends`` namespace in ``setup.py``:
+
+.. code-block:: python
+
+   setup(
+       # ...,
+       entry_points={
+           "yt.frontends": [
+               "myFrontend = my_frontend.api.MyFrontendDataset",
+               "myOtherFrontend = my_frontend.api.MyOtherFrontendDataset",
+           ]
+       }
+   )
+
+or ``pyproject.toml``:
+
+.. code-block:: toml
+
+   [project.entry-points."yt.frontends"]
+   myFrontend = "my_frontend.api:MyFrontendDataset"
+   myOtherFrontend = "my_frontend.api:MyOtherFrontendDataset"
+
+will be automatically loaded and immediately available in
+:func:`~yt.loaders.load`.
+
 Extension Template
 ------------------
 
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -9,6 +9,7 @@ import tarfile
 import time
 import types
 import warnings
+from importlib.metadata import entry_points
 from multiprocessing import Pipe, Process
 from multiprocessing.connection import Connection
 from pathlib import Path
@@ -96,6 +97,15 @@ def load(
     # either in the current dir or yt.config.ytcfg['data_dir_directory']
     if not fn.startswith("http"):
         fn = str(lookup_on_disk_data(fn))
+
+    if sys.version_info >= (3, 10):
+        external_frontends = entry_points(group="yt.frontends")
+    else:
+        external_frontends = entry_points().get("yt.frontends", [])
+
+    # Ensure that external frontends are loaded
+    for entrypoint in external_frontends:
+        entrypoint.load()
 
     candidates = []
     for cls in output_type_registry.values():
--- a/yt/tests/test_external_frontends.py
+++ b/yt/tests/test_external_frontends.py
@@ -19,6 +19,7 @@ class MockEntryPoint:
 
             def _parse_parameter_file(self):
                 self.current_time = 1.0
+                self.cosmological_simulation = 0
 
             def _set_code_unit_attributes(self):
                 self.length_unit = self.quan(1.0, "code_length")
--- a/yt/tests/test_external_frontends.py
+++ b/yt/tests/test_external_frontends.py
@@ -1,10 +1,10 @@
 import sys
-import tempfile
-from unittest import mock
 
+import pytest
+
+import yt
 from yt.data_objects.static_output import Dataset
 from yt.geometry.grid_geometry_handler import GridIndex
-from yt.loaders import load
 from yt.utilities.object_registries import output_type_registry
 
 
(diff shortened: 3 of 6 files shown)


doc/source/developing/extensions.rst
```diff
@@ -35,6 +35,24 @@ In subsequent versions, we plan to include in yt a catalog of known extensions
...
+:class:`~yt.data_objects.static_output.Dataset` subclass as an entrypoint in
+``yt.frontends`` namespace:
+
+.. code-block:: python
+
+   setup(
+       entry_points={
+           "yt.frontends": ["myFrontend = my_frontend.api.MyFrontendDataset"]
+       }
+   )
```
neutrinoceros, line 51: it should be made explicit that this is an example using `setup.py`. I suggest also showing the syntax for `pyproject.toml` (example: https://github.com/neutrinoceros/yt_idefix).

`setup.cfg` is now legacy so it's not worth adding.

yt/loaders.py
```diff
@@ -12,6 +12,7 @@
 from multiprocessing import Pipe, Process
 from multiprocessing.connection import Connection
 from pathlib import Path
+from pkg_resources import iter_entry_points
```
neutrinoceros, line 15: this import is very expensive (about twice as much as `import numpy`). 
It inflicts a ~10% penalty in the current state of things on `import yt` (which I've been trying very hard to accelerate).
Unless there a cheaper alternative, I suggest moving this statement to `def load`.

neutrinoceros (commented):
I had no idea this was possible, but I love it, thank you !

doc/source/developing/extensions.rst
neutrinoceros: Also, I assume from the type (list) that it's possible to define more than one Dataset subclass, but can you confirm ?

Xarthisius: That I don't know, but I believe so.

yt/loaders.py
Xarthisius: moved

yt/loaders.py
```diff
@@ -97,6 +97,13 @@ def load(
     if not fn.startswith("http"):
         fn = str(lookup_on_disk_data(fn))
 
+    # import it here cause it's costly
+    from pkg_resources import iter_entry_points
```
neutrinoceros, line 101: It should be possible to switch to `importlib.metadata`, which would not only help with the perf overhead, but is better standardised.
It is what `flake8` currently uses so I'm hoping it well covers the need here

refs
https://docs.python.org/3.11/library/importlib.metadata.html?highlight=importlib%20metadata#entry-points

https://github.com/PyCQA/flake8/blob/b87034d22490583f99149f5d241750cc8f8e3a90/src/flake8/plugins/finder.py#L35

doc/source/developing/extensions.rst
Xarthisius: It works with multiple entries.

yt/loaders.py
Xarthisius: changes to `importlib.metadata`

yt/tests/test_external_frontends.py
```diff
@@ -0,0 +1,41 @@
...
+    assert "MockDataset" not in output_type_registry
+
+    with tempfile.NamedTemporaryFile() as test_file:
+        with mock.patch("importlib.metadata.entry_points") as ep:
+            ep.return_value = {"yt.frontends": [MockEntryPoint]}
+            ds = load(test_file.name)
+            assert "MockDataset" in str(ds.__class__)
+
+    assert "MockDataset" in output_type_registry
+    output_type_registry.pop("MockDataset")
```
neutrinoceros, line 41: this should be done with a pytest fixture in order to guarantee that cleanup happens regardless of the test's success. `yt/tests/test_load_errors.py` has such a fixture (`catchall_dataset_class`). You could merge the two files into `test_load.py` and call it a day. Well you'd need to find and replace all instances of `test_load_errors` so that nose keeps ignoring it but that should be it.

yt/loaders.py
```diff
@@ -97,6 +97,21 @@ def load(
     if not fn.startswith("http"):
         fn = str(lookup_on_disk_data(fn))
 
+    # import it here cause it's costly
+    from importlib.metadata import entry_points
```
neutrinoceros, line 101: well this one is cheap actually, so we can move it back to top level

yt/loaders.py
```diff
@@ -97,6 +97,21 @@ def load(
...
+    # https://github.com/python/importlib_metadata/pull/278
+    entrypoints = entry_points()
+    if hasattr(entrypoints, "select"):
+        external_frontends = entrypoints.select(group="yt.frontends")
+    else:
+        external_frontends = entrypoints.get("yt.frontends", [])
+
+    # Ensure that external frontends are loaded
+    for entrypoint in external_frontends:
+        entrypoint.load()
```
neutrinoceros, line 113: could this be done at the top level of the module too ? In case ones has a broken plugin installed I assume it's preferable to error sooner.

yt/loaders.py
```diff
@@ -97,6 +97,21 @@ def load(
     if not fn.startswith("http"):
         fn = str(lookup_on_disk_data(fn))
 
+    # import it here cause it's costly
+    from importlib.metadata import entry_points
+
+    # overly complicated dance to suppress spurious warning
+    # https://github.com/python/importlib_metadata/pull/278
+    entrypoints = entry_points()
+    if hasattr(entrypoints, "select"):
```
neutrinoceros, line 106: By making this conditional explicitly about the runtime Python version number we enable auto-cleaning with pyupgrade in the future
```suggestion
    if sys.version_info >= (3, 10):
```

neutrinoceros (commented):
additional pointers. I haven't studied why the test is failing on Windows yet

yt/loaders.py
neutrinoceros: I discovered the minimal required version manually, but it's also documented in https://docs.python.org/3.11/library/importlib.metadata.html?highlight=importlib%20metadata#entry-points
(see Compatibility Note)

yt/loaders.py
neutrinoceros: (on second thought it might create import cycles, so maybe it's not a good idea)

neutrinoceros: ((It should _not_ create import cycles, and currently no one is relying on this so we might as well put the responsibility of avoiding cycles on the external frontend if such an anti-pattern is ever needed))

doc/source/developing/extensions.rst
```diff
@@ -35,6 +35,32 @@ In subsequent versions, we plan to include in yt a catalog of known extensions
...
+
+or ``pyproject.toml``:
+
+.. code-block:: toml
+
+   [project.entry-points."yt.frontends"]
+   IdefixVtkDataset = "yt_idefix.data_structures:IdefixVtkDataset"
+   IdefixDmpDataset = "yt_idefix.data_structures:IdefixDmpDataset"
+   PlutoVtkDataset = "yt_idefix.data_structures:PlutoVtkDataset"
```
neutrinoceros, line 59: love the easter egg, but I think the setup.py and pyproject.toml example should be equivalent (best to include multiple entry points in both)

yt/loaders.py
neutrinoceros: the paragraph further reads 
> In the future, calling entry_points with no parameters will return an EntryPoints object. Users should rely on the selection interface to retrieve entry points by group.

I guess this means that to maximise stability, we should retrieve `external_frontends = entry_points(group="yt.frontends")` directly in Python 3.10 and newer.

Xarthisius: yeah, but now I have to figure out mocking for two versions of python along with a fix for windows...

yt/tests/test_external_frontends.py
Xarthisius: I cannot define it in a similar way to `test_load_errors.py` cause it will instantiate Dataset before entering load, which defeats the whole purpose of the test.

yt/loaders.py
neutrinoceros: Good luck. I may be able to give you a hand a couple hours from now if you need.

Xarthisius:
How to fix windows is beyond me... Why answer tests fail (with blank differences) is also beyond me. I think I addressed everything else.

neutrinoceros:
> How to fix windows is beyond me...

Am I right to assume fixing #4286 will not resolve the problem here ?
I will try to find a solution here.

>  Why answer tests fail (with blank differences) is also beyond me.

This is just flakiness, do not worry about it: it's been like this for months now, and we have a solution in #4122

Xarthisius:
> My two unresolved threads still stand, is it alright if I push one commit to add test cleanup, or would you prefer that I write the fixture as a comment ?

You can push directly.

neutrinoceros:
cool. I'll do this in the next couple hours

yt/loaders.py
neutrinoceros: moving this to top level makes it even more difficult to test, and the overhead I can measure (~10ms) doesn't seem worth the effort.

yt/tests/test_external_frontends.py
neutrinoceros: done !

neutrinoceros:
I will double check that it works as is for yt_idefix

neutrinoceros (dismissed):
It works perfectly with yt_idefix, thanks again for starting this !

chrishavlin (approved):
this is great! 

I do think it will marginally increase the chance of having plugin loading conflicts. Since previously you had to explicitly import the plugin, you'd never end up with two external frontends identifying a file as valid. But now if a user happens to have installed two external entrypoint-enabled frontend plugins that both identify a file as valid they'll get an `YTAmbiguousDataType` error. But the benefit here **far** outweighs the slim chance of encountering that IMO so I don't think it's worth worrying about, but I thought it was worth mentioning.

neutrinoceros:
> But now if a user happens to have installed two external entrypoint-enabled frontend plugins that both identify a file as valid they'll get an YTAmbiguousDataType error. But the benefit here far outweighs the slim chance of encountering that IMO so I don't think it's worth worrying about, but I thought it was worth mentioning.

That's correct, although it's nice to think that this will _not_ be a deadend situation even then thanks to `yt.load`'s `hint` keyword argument !

neutrinoceros (approved):

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):

chrishavlin (approved):

neutrinoceros:
marking this PR for backport but I'll only cherry pick b2762924275b937999d42a12d3fe180724201ccc

PR #4292, opened 2023-01-10
ENH: migrate from pkg_resources to importlib.resources on Python 3.9+

### LINKED ISSUE #4287: ENH: drop pkg_resources ?
There a a couple places where yt depends on `pkg_resources`, which is notorious for having a large import overhead.
We've been mitigating this by moving such imports statements where they are actually used instead of at module level so it doesn't currently impact yt's own import time, but there are alternatives (`importlib.metadata`, `importlib.resources`) in the standard library that may be worth exploring.

neutrinoceros:
## PR Summary

fix #4287
Turns out it was super easy, barely an inconvenience.

### DIFF
--- a/setupext.py
+++ b/setupext.py
@@ -1,5 +1,6 @@
 import contextlib
 import glob
+import importlib.resources
 import logging
 import os
 import platform
@@ -13,8 +14,6 @@ from distutils.ccompiler import CCompiler, new_compiler
 from distutils.sysconfig import customize_compiler
 from subprocess import PIPE, Popen
 from sys import platform as _platform
-
-from pkg_resources import resource_filename
 from setuptools.command.build_ext import build_ext as _build_ext
 from setuptools.command.sdist import sdist as _sdist
 from setuptools.errors import CompileError, LinkError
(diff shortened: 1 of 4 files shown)


setupext.py
```diff
@@ -205,7 +204,7 @@ def check_for_pyembree(std_libs):
     embree_libs = []
     embree_aliases = {}
     try:
-        _ = resource_filename("pyembree", "rtcore.pxd")
+        _ = importlib.resources.files("pyembree") / "rtcore.pxd"
```
matthewturk, line 207: this *doesn't* need changes on the pyembree side, right?

neutrinoceros:
*Of course* the functions I needs were not in the standard library before Python 3.9... adding compatibility for Python 3.8

setupext.py
neutrinoceros: It doesn't. It works exactly like `pkg_resources` except that it returns a path.
In hindsight it's completely useless (in both implementations) to look for an exact file like we do here, because both implementations are happy to produce a path to a file even if it's not there (though they _do_ error with `ModuleNotFoundError` if `pyembree` isn't installed).

setupext.py
```diff
@@ -204,8 +203,16 @@ def check_CPP14_flags(possible_compile_flags):
...
+        _path_finder = importlib.resources.files
+    else:
+        from pkg_resources import resource_filename
+        from functools import partial
+        _path_finder = partial(resource_filename, resource_name="rtcore.pxd")
+
     try:
-        _ = resource_filename("pyembree", "rtcore.pxd")
+        _path_finder("pyembree")
     except ImportError:
```
Xarthisius, line 216: I believe that `importlib.resources.files` raises `ModuleNotFoundError` instead of `ImportError` when the package in question is not installed. We need to expand that `except`.

Xarthisius (approved):

Issue #4291, opened 2023-01-10
BUG: ParticlePhasePlot does not redraw when _switch_ds is called


mtryan83:
### Bug report

**Bug summary**

The cookbook recipe ([Time Series Movie](https://yt-project.org/doc/cookbook/complex_plots.html#time-series-movie)) for animating plots suggests using `plot._switch_ds(new_ds)` to update the plot as part of the animation function. This does not work for `ParticlePhasePlot`s (and likely `PhasePlot`s in general).  Relevant to #4288.

**Code for reproduction**

```python
# This is copied directly from the cookbook with the modifications to create
# a semi-comparable ParticlePhasePlot animation
from matplotlib import rc_context
from matplotlib.animation import FuncAnimation

import yt

ts = yt.load("GasSloshingLowRes/sloshing_low_res_hdf5_plt_cnt_*")

#plot = yt.SlicePlot(ts[20], "z", ("gas", "density"),window_size=2)
plot = yt.ParticlePhasePlot(ts[0], ("gas","x"), ("gas", "y"), ("gas","density"), figure_size=2, x_bins=100, y_bins=100)
plot.set_unit(("gas","x"),"Mpc") # To match the SlicePlot units
plot.set_unit(("gas","y"),"Mpc")

plot.set_zlim(("gas", "density"), 8e-29, 3e-26)

fig = plot.plots[("gas", "density")].figure

# animate must accept an integer frame number. We use the frame number
# to identify which dataset in the time series we want to load
def animate(i):
    ds = ts[i]
    plot._switch_ds(ds)

animation = FuncAnimation(fig, animate, frames=len(ts))

# Override matplotlib's defaults to get a nicer looking font
with rc_context({"mathtext.fontset": "stix"}):
    animation.save("animation.mp4")
```

**Actual outcome**

https://user-images.githubusercontent.com/1833991/211651361-c05f464a-f339-4597-9880-576b085c3724.mp4
Note that the figure is updating, you can add a timestamp for example that will update correctly in each frame, but the underlying plot is not being replotted or something. 

**Expected outcome**

https://user-images.githubusercontent.com/1833991/211652283-2db846ee-76ec-4e91-a6cf-4b8c4ea3c8bd.mp4

This was generated by generating each plot individually, saving them to separate files, and then using `imageio` to combine them into an mp4:
```python
import os
import tempfile
import imageio
import yt

ts = yt.load("GasSloshingLowRes/sloshing_low_res_hdf5_plt_cnt_*")

[...]

PR #4294, opened 2023-01-11
minor refactor phaseplots

neutrinoceros:
## PR Summary

A very innocent noop refactor that I produced as a side effect of trying to debug #4291


matthewturk (commented):
Seems okay.  I wonder if we ever modified `xlim` and `ylim` before.

matthewturk (approved):

Issue #4288, opened 2023-01-09
DOC: _switch_ds being exposed


neutrinoceros:
### Bug report

The `PlotContainer._switch_ds` method's name indicates that it is meant to be private.
It is however used in [this docs section](https://yt-project.org/docs/dev/cookbook/complex_plots.html?highlight=funcanimation#time-series-movie).
I conclude that we should either promote it to public API (remove the trailing underscore) or update the documentation to avoid using it.

chrishavlin:
I vote promotion! Seems like useful functionality. And since we have the docs and at least one person was using it, why not promote it?

neutrinoceros:
I agree

neutrinoceros:
So a couple notes before I forget
- name the public function `switch_dataset`
- keep `_switch_ds` as a deprecated alias
- update all docs examples
- add a docstring

neutrinoceros:
I noticed there's a fair bit of needless coupling in the current basic implementation of the method, because it's trying (and failing) to support every quirk of child classes. I think exposing it will be an opportunity to get rid of that.

neutrinoceros:
Another note: the `data_source` argument seems to be weirdly behaved and it's neither tested nor documented, so I think we should not expose it in the public API. We can always add it back later, if and when anyone misses it.

PR #4295, opened 2023-01-11
BUG: implement _switch_ds for PhasePlot

### LINKED ISSUE #4291: BUG: ParticlePhasePlot does not redraw when _switch_ds is called

neutrinoceros:
## PR Summary

fixes #4291
In short, the basic implementation of `_switch_ds` doesn't work correctly for `PhasePlot` because it doesn't regenerate the underlying profile data.

### DIFF
--- a/yt/visualization/profile_plotter.py
+++ b/yt/visualization/profile_plotter.py
@@ -1,6 +1,7 @@
 import base64
 import builtins
 import os
+import weakref
 from functools import wraps
 from typing import Any, Dict, Iterable, Optional, Tuple, Union
 
@@ -22,6 +23,7 @@ from ._commons import validate_image_name
 from .plot_container import (
     BaseLinePlot,
     ImagePlotContainer,
+    invalidate_data,
     invalidate_plot,
     validate_plot,
 )
@@ -981,6 +983,35 @@ class PhasePlot(ImagePlotContainer):
         obj._setup_plots()
         obj._initfinished = True
         return obj
+
+    @invalidate_data
+    def _switch_ds(self, new_ds, data_source=None):
+        if data_source is not None:
+            raise RuntimeError(
+                "The data_source keyword argument "
+                "is not supported for PhasePlot._switch_ds"
+            )
+
+        old_object = self.data_source
+        type_name = old_object._type_name
+        kwargs = {n: getattr(old_object, n) for n in old_object._con_args}
+        self.data_source = getattr(new_ds, type_name)(**kwargs)
+        self.ds = weakref.proxy(new_ds)
+
+        op = self.profile  # old profile
+        self._profile = create_profile(
+            self.data_source,
+            bin_fields=[op.x_field, op.y_field],
+            fields=op.field_data.keys(),
+            n_bins=(len(op.x_bins), len(op.y_bins)),
+            weight_field=op.weight_field,
+            accumulation=op.accumulation,
+            fractional=op.fractional,
+        )
+        self._profile_valid = True
+        self.plots.data_source = self.data_source
+        self._colorbar_label.data_source = self.data_source
+        self._setup_plots()
 
     def _get_field_title(self, field_z, profile):
         field_x = profile.x_field
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -308,7 +308,13 @@ _fake_amr_ds_default_units = ("g/cm**3",)
 
 
 def fake_amr_ds(
-    fields=None, units=None, geometry="cartesian", particles=0, length_unit=None
+    fields=None,
+    units=None,
+    geometry="cartesian",
+    particles=0,
+    length_unit=None,
+    *,
+    seed=0x4D3D3D3,
 ):
     from yt.loaders import load_amr_grids
 
@@ -323,7 +329,7 @@ def fake_amr_ds(
         },
     )
 
-    prng = RandomState(0x4D3D3D3)
+    prng = RandomState(seed)
     LE, RE = _geom_transforms[geometry]
     LE = np.array(LE)
     RE = np.array(RE)
--- a/yt/visualization/tests/test_profile_plots.py
+++ b/yt/visualization/tests/test_profile_plots.py
@@ -11,6 +11,7 @@ from yt.testing import (
     ANSWER_TEST_TAG,
     assert_allclose_units,
     assert_array_almost_equal,
+    fake_amr_ds,
     fake_random_ds,
 )
 from yt.utilities.answer_testing.framework import (
@@ -350,3 +351,22 @@ def test_phaseplot_showhide_colorbar_axes():
     plot.show_axes()
     with tempfile.NamedTemporaryFile(suffix="png") as f4:
         plot.save(f4.name)
+
+
+def test_phase_plot_switch_dataset():
+    # see bug https://github.com/yt-project/yt/issues/4291
+    fields = [("gas", "field1"), ("gas", "field2"), ("gas", "mass")]
+    units = ["code_length", "code_length", "code_mass"]
+
+    ds1 = fake_amr_ds(fields=fields, units=units, seed=1)
+    pp = PhasePlot(ds1, *fields[:2], z_fields=("gas", "field1"))
+    assert pp.ds == ds1
+    assert pp.profile.ds == ds1
+
+    # the seed isn't used in dataset comparison
+    # but it's useful for visual testing
+    ds2 = fake_amr_ds(fields=fields, units=units, seed=2)
+    assert ds2 != ds1
+    pp._switch_ds(ds2)
+    assert pp.ds == ds2
+    assert pp.profile.ds == ds2


yt/visualization/profile_plotter.py
```diff
@@ -982,6 +983,45 @@ def _initialize_instance(
...
+        name = old_object._type_name
+        kwargs = {n: getattr(old_object, n) for n in old_object._con_args}
+        kwargs["center"] = getattr(old_object, "center", None)
+        if data_source is not None:
+            if name != "proj":
+                raise RuntimeError(
+                    "The data_source keyword argument "
+                    "is only defined for projections."
+                )
+            kwargs["data_source"] = data_source
```
neutrinoceros, line 998: this can probably be dropped completely, but I'll need a hand to confirm that I'm reading this right.

neutrinoceros:
I added a test that fails on main and pass here. I'll open for review when CI is green again.
Note that I'm deliberately leaving #4288 aside for now, and focused on making this patch backportable.

mtryan83:
So this does seem to fix the "nothing is happening aspect", but it doesn't seem to generate the images I would expect. It's a little tricky to show, but if I compare the static image of the last dataset and the last frame of the animation, both generated with the same plot command, I see differences.
```Python
ts = yt.load("GasSloshingLowRes/sloshing_low_res_hdf5_plt_cnt_*")

def makeplot(ds):
    col_field = ("gas","density")
    plot = yt.ParticlePhasePlot(ds, ("gas","x"), ("gas", "y"), col_field, figure_size=2, x_bins=100, y_bins=100)
    plot.set_unit(("gas","x"),"Mpc")
    plot.set_unit(("gas","y"),"Mpc")
    plot.set_zlim(col_field, 8e-29, 3e-26)
    return plot, col_field # returning the color field for id purposes

def makeplotSlice(ds):
    col_field = ("gas","density")
    plot = yt.SlicePlot(ds, "z", col_field,window_size=2)
    plot.set_zlim(col_field, 8e-29, 3e-26)
    return plot, col_field

def make_animation_directly(dss,plotcommand):
    plot,col_field = plotcommand(dss[0])
    fig = plot.plots[col_field].figure
    # animate must accept an integer frame number. We use the frame number
    # to identify which dataset in the time series we want to load
    def animate(i):
        ds = dss[i]
        print(f"Drawing frame {i}", end = "\r")
        plot._switch_ds(ds)
        
    animation = FuncAnimation(fig, animate, frames=len(dss))
    if save_filename is not None:
        animation.save(save_filename)
    return animation

# Generate plots
plot, _ = makeplot(ts[-1])
plot.show()
ani = make_animation_directly(ts, makeplot)
ani
```
plot: 
image

last frame:
image

Note the differing pixel resolution for example. Differences show up earlier too, the final frame/dataset is just the easiest to reproduce.  (And it's much more obvious with the simulation data I'm working on, though I'm not sure why). I don't see a difference if I use the makeplotSlice command instead.  

This might relate to a missing `_recreate_frb` attribute. If I try 
```Python
plot,_,_ = makeplot(ts[0])
plot._switch_ds(ts[-1])
plot.show()
plot,_,_ = makeplot(ts[-1])
plot.show()
```
I get an `AttributeError: 'ParticlePhasePlot' object has no attribute '_recreate_frb'` in line 159 of `yt/yt/visualization/_commons.py` whereas it works fine (and produces identical images) if I use
```Python
plot,_,_ = makeplotSlice(ts[0])
plot._switch_ds(ts[-1])
plot.show()
plot,_,_ = makeplotSlice(ts[-1])
plot.show()
```
image

image

neutrinoceros:
Thanks a lot for the detailed review, it's really helpful. I can see however that the complete patch is certainly going to grow larger than a reasonable backport, so I won't aim for including it in the next bugfix release just yet.

neutrinoceros:
Okay le me dive a little bit deeper


> This might relate to a missing _recreate_frb attribute. If I try
...
I get an AttributeError: 'ParticlePhasePlot' object has no attribute '_recreate_frb' in line 159 of yt/yt/visualization/_commons.py

I think that this is symptomatic of a design weakness we've had in the code base for a while: many class methods in `PlotContainer` and its hierarchy rely on `@(in)validate_(plot|figure|data)` decorators, which supposedly avoids duplicating code, so in principle every class using them should conform to common protocols, but currently they refuse to`. As a result, these decorators end up knowing way too much about their user classes and try desperately to adapt to each of them, but this is evidence that this model doesn't work well (at least it doesn't scale).
There is evidence that this problem was noticed before, as this old comment shows:
https://github.com/yt-project/yt/blob/e43ec1da01938410374f31b3a77441c641438f5a/yt/visualization/_commons.py#L156-L157

So I think what's needed here is a clean refactor of these decorators and their user classes. I may experiment with that within this PR, but I expect this may be a rabbit hole, so I'll switch to draft for now.

matthewturk:
In lieu of that, I would suggest we consider switching to using traitlets.

neutrinoceros:
I can't in good faith say I want to give it a try. At least I need to be convinced that it will make my job easier rather than harder, given that I currently don't know the framework at all.

matthewturk:
That's fair.  Let's talk in lower-latency about it, and if I can help with it, because I have a considerable amount of experience doing it, and I am keen to revisit my attempts at it from a few years ago.

neutrinoceros:
Vitals are not good on this PR. I'll take it out of the 4.2 milestone so it doesn't block the release

mtryan83:
I just finished my defense and can work on this again, is there anything I can do to help here or with #4301?

neutrinoceros:
Hi @mtryan83, first of: congrats !

I still think an important refactor is needed to address this properly, and I already have too many plates spinning, so unfortunately I do not think that I can reasonably get back to it any time soon.
If anyone would like to use my work as a starting point for the refactor I've hinted at, I'll be happy to serve as a reviewer and/or provide some more guidelines. I'd also be happy to see this superseded by #4301 if we can get it going.

neutrinoceros:
I'll close this to clear the backlog but I reaffirm that anyone wanting to push this has my support and I'd be happy to help.

PR #4301, opened 2023-01-13
[WIP] Revive traitlets attempt

matthewturk:
This is a revival of an attempt at adding traitlets support.

I've so far just rebased the initial code I wrote a while back, with the intent of exploring a proof of concept related to #4295.

### DIFF
--- a/yt/visualization/fixed_resolution.py
+++ b/yt/visualization/fixed_resolution.py
@@ -1,9 +1,6 @@
 import weakref
 from functools import partial
 from typing import TYPE_CHECKING, Dict, List, Optional
-
-import numpy as np
-
 from yt._maintenance.deprecation import issue_deprecation_warning
 from yt._typing import FieldKey
 from yt.data_objects.image_array import ImageArray
@@ -23,8 +20,11 @@ from .volume_rendering.api import off_axis_projection
 if TYPE_CHECKING:
     from yt.visualization.fixed_resolution_filters import FixedResolutionBufferFilter
 
+import traitlets
+from yt.utilities.traitlets_support import \
+    YTDimensionfulTrait, YTPositionTrait
 
-class FixedResolutionBuffer:
+class FixedResolutionBuffer(traitlets.HasTraits):
     r"""
     FixedResolutionBuffer(data_source, bounds, buff_size, antialias = True)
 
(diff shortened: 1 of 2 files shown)


yt/visualization/fixed_resolution.py
```diff
@@ -97,6 +97,15 @@ class FixedResolutionBuffer:
         ("index", "theta"),
         ("index", "dtheta"),
     )
+    antialias = traitlets.Bool(True)
+    buff_size = traitlets.Tuple(traitlets.Int(), traitlets.Int())
+    period = traitlets.Tuple(YTDimensionfulTrait(dimensions = dims.length),
+                             YTDimensionfulTrait(dimensions = dims.length))
+    periodic = traitlets.Bool(True)
+    bounds = traitlets.List(YTDimensionfulTrait(dimensions = dims.length))
+    axis = traitlets.Enum([0,1,2,3,4])
```
neutrinoceros, line 106: headsup: I see a collision here with https://github.com/yt-project/yt/pull/3827/commits/b02df62bc936f15a3b66bf3adcede95265b075d9 from #3827

matthewturk: thanks, I'm sure this will get lots of changes in the meantime

matthewturk:
At this stage, while I do think I'll be able to get the tests passing, this does not yet demonstrate what I'd like to see for using traitlets.  It shows some of the type coercion getting moved to a single place, and how it interacts with type hinting, but it doesn't yet show how trait modifications can lead to data-regeneration.  I'm going to try to do that a level up from this.

matthewturk:
The final failing test is:

```
FAILED yt/data_objects/tests/test_data_containers.py::TestDataContainers::test_to_frb - traitlets.traitlets.TraitError: The 'buff_size' trait of a CylindricalFixedResolutionBuffer instance expected a tuple, not the int 64.
```

which I will fix by fixing the test, *but*, it raises an interesting point which is that in some cases, we have a linear inheritance hierarchy *but* we have different arguments to the initializer.  For this particular case, it's the `CylindricalFixedResolutionBuffer` which inherits from `FixedResolutionBuffer`.  The former accepts `radius` and not `bounds` and the latter doesn't accept `radius`.  Because we don't allow kwargs to be passed through, we have to jump-the-line to get them up to the initializer of `HasTraits`, which kind of stinks.  I may handle this, or I might look for another indirect solution.

neutrinoceros:
FWIW I think it's okay to change the signatures of misaligned `__init__` methods and get it closer to alignment. In other words I think (or I hope) we can treat most Plot classes in yt as internal, because they are so hard to work with I don't expect anyone downstream to be brave enough to use them directly or subclass them.

matthewturk:
OK!  That sounds fine to me.

PR #4305, opened 2023-01-17
BUG: fix calling ds.proj.to_frb on curvilinear geometries

### LINKED ISSUE #4304: Projection on some axis fail in curvilinear geometries
### Bug report

**Bug summary**

Certain type of projections ("y" in cylindrical and "z" in polar) fail in curvilinear geometries

**Code for reproduction**

```python
import yt
ds = yt.load("amrvac/bw_cylindrical_3D0000.dat")
proj = ds.proj("density", 1)
frb = proj.to_frb((1.0, "unitary"), 256)
frb.render("density")
```

**Actual outcome**

```
  File "yt/visualization/fixed_resolution.py", line 201, in render
    return self[item]
  File "yt/visualization/fixed_resolution.py", line 574, in __getitem__
    pixelize_cylinder(
  File "yt/utilities/lib/pixelization_routines.pyx", line 522, in yt.utilities.lib.pixelization_routines.pixelize_cylinder
ValueError: Buffer has wrong number of dimensions (expected 2, got 1)
```

**Expected outcome**

No error

**Version Information**
  * Operating System: Ubuntu 20.04
  * Python Version: 3.8.13
  * yt version: https://github.com/yt-project/yt/commit/97307f4ebc78294717b718b77ecab2c1516d601a

neutrinoceros:
## PR Summary
fix #4304

### DIFF
--- a/yt/data_objects/selection_objects/data_selection_objects.py
+++ b/yt/data_objects/selection_objects/data_selection_objects.py
@@ -16,7 +16,7 @@ from yt.data_objects.data_containers import YTDataContainer
 from yt.data_objects.derived_quantities import DerivedQuantityCollection
 from yt.data_objects.field_data import YTFieldData
 from yt.fields.field_exceptions import NeedsGridType
-from yt.funcs import fix_axis, is_sequence, iter_fields, validate_width_tuple
+from yt.funcs import fix_axis, is_sequence, iter_fields
 from yt.geometry.api import Geometry
 from yt.geometry.selection_routines import compose_selector
 from yt.units import YTArray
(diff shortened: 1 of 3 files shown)


neutrinoceros:
A previously existing test is currently failing
https://github.com/yt-project/yt/blob/4eda8b275e0e979f982b32bb700736315caf9f43/yt/data_objects/tests/test_data_containers.py#L115

But more importantly, it's surprisingly similar to the one I'm adding, and apparently it never failed before. I need to understand the difference.

neutrinoceros:
in indsight I'm not sure aiming this to be backported is the right call:
- it's "fixing" code that has been broken for at least a couple years so it's pretty clear no-one is relying on the behaviour it's enabling
- it's barely working at all (hacking the `width/height` interface that clearly was never meant for curvilinear geometries), and still leaves obvious gaps (spherical geometries are not acknowledged, but they should)

So if you're okay with it @Xarthisius , I'm thinking of putting much more thoughts and care into a broader patch, and aim it at yt 4.2

matthewturk:
@neutrinoceros I think it would be worthwhile -- perhaps for 4.2 -- to rethink how we approach the specification of bounds for curvilinear coordinates, and consider shifting it to utilize full-specification using cartesian edges, which would also be related to #4179 and #3324 .  I think it might be worth a brief not-quite-YTEP discussing this, or at least a high-bandwidth conversation.

neutrinoceros:
>  to rethink how we approach the specification of bounds for curvilinear coordinates, and consider shifting it to utilize full-specification using cartesian edges,

Sorry this has taken so long for me to respond. To be completely honest I don't think this is the way to go. From what I can tell people doing simulations with curvinlinear coordinate systems are perfectly happy to stay away from cartesian coordinates as much as possible when post-processing and vizualizing their data. In many ways, yt is already deeply rooted into cartesian coordinates, and there are a number of places where supporting other geometries takes the form of hacks. Eventually™️ I would like to get to a state where cartesian coordinates are just one particular flavour (albeit special in that implementation is simpler and performances are better) instead of an inescapable default, but I don't think that imposing cartesian coordinate even more onto users will help in the long run.

I anticipate that rethinking the code at a high level will be a substantial effort, especially if we worry about backward compatibility (of course we should). In anycase it's much much broader than the original scope of this PR, and I certainly don't have the energy to entertain that thought. Right now I can't tell wether I'll want to come back to this in the short term. If not, I'll probably just close it.

matthewturk:
I'd like to think about my response to this, because I think there's value in the differing type of visualization, and I know @jzuhone has thought about this as well.  But I think it's OK for you not to be on the hook for it, of course!

neutrinoceros:
I'll close this and make room for the rest of the backlog.

PR #4306, opened 2023-01-18
ENH: use tesselation for better N-body rendering

cphyc:


yt/utilities/lib/image_utilities.pyx
```diff
@@ -7,28 +7,132 @@ Utilities for images
...
+                        c[idim] = p3d[i + off[2][0], j + off[2][1], k + off[2][2], idim] - orig[idim]
+
+                    for s in prange(split, nogil=True):
+                        for idim in range(2):  # stop at 2, because z is not used
+                            newp[idim] = orig[idim] + (
+                                ro[s, 0] * a[idim] +
+                                ro[s, 1] * b[idim] +
+                                ro[s, 2] * c[idim]
+                            )
+                            while newp[idim] < 0.0:
```
neutrinoceros, line 127: Can't this be achieved with a single modulus operation ?

cphyc: It can and I've now fixed this!

neutrinoceros:
Too soon for a review I guess but just wanted to say this looks amazing. 
> Cherry on the cake: this PR is actually much faster than SPH interpolation because there's no need to build the KD-tree necessary for the SPH interpolation.

Can you clarify what you compared ? (since the quality can be adjusted I assume increasing it has a performance cost)

cphyc:
> Can you clarify what you compared? (since the quality can be adjusted I assume increasing it has a performance cost)

I've updated the PR description, so hopefully, that should be better. The most crucial bit is that the SPH-interpolation scales as `N log(N)` while this implementation scales as `N` (times the number of samples).

The reason there needs to be some sampling is that we need to compute the volume of the intersection of each tetrahedron with each pixel. In principle, this could be computed exactly but the lazy way is just to (randomly) sample the tetrahedron to estimate the volume.

neutrinoceros:
> The reason there needs to be some sampling is that we need to compute the volume of the intersection of each tetrahedron with each pixel. In principle, this could be computed exactly but the lazy way is just to (randomly) sample the tetrahedron to estimate the volume.

Interesting. I'm guessing there should be a point (critical number of samples) at which the lazy evaluation becomes as expensive as the exact evaluation. Is there an estimate for it ? It's not critical, but if it's known it should probably be documented.

cphyc:
TBH I'd eventually want to use https://github.com/devonmpowell/r3d/ to compute the intersection between voxels and tetrahedron, but that may be too optimistic.

cphyc:
After comments from @yipihey, I went ahead and implemented an "exact" integrator. It works by noticing that the projection of a uniformly-sampled tetrahedron onto the plane is a simple geometrical figure (see below).


There are only two cases to treat: either the projection of the tetrahedron is a triangle or a quadrilateral. 
| Triangular case | Quadrilateral case |
| --- | --- |
| image | image |
My implementation works as follows:
1. Project the four vertices defining each tetrahedron onto a plane,
2. Find the convex hull,
3. If the hull is made of three points, the maximum density is reached at the location of the fourth point.
If the hull is a quadrilateral, split it into two triangles with the density maximum at the intersection of the two diagonals.
5. For each triangle, iterate over all pixels within the bounding box of the triangle.
6. If the pixel is within the triangle, linearly interpolate the projected density at this location.

yipihey:
Very nice!There is a fast way we developed for projections that is particularly efficient on GPUs that we implemented here (see section 4.4)A Novel Approach to Visualizing Dark Matter Simulationsarxiv.orgYou first find the back and the front faces of tetrahedra by dotting the normal of each face with the unit vector defining the direction of your view. Then you add the density times the depth of the face (for each image pixel ) of the front facing faces into one image and for the back facing one’s in another image. When you are done with all tetrahedra you subtract the two images and you get back the integrated density along the line of sight. Voila exact projection of the linear elements with few if statements. Hope this approach may help to speed also your current implementation. TomTyped on an IPhone. Please excuse typos and surprising autocorrects. On Jan 21, 2023, at 19:23, Corentin Cadiou ***@***.***> wrote:﻿
After comments from @yipihey, I went ahead and implemented an "exact" integrator. It works by noticing that the projection of a uniformly-sampled tetrahedron onto the plane is a simple geometrical figure (see below).
There are only two cases to treat: either the projection of the tetrahedron is a triangle or a quadrilateral.



Triangular case
Quadrilateral case








My implementation works as follows:





Project the four vertices defining each tetrahedron onto a plane,
Find the convex hull,
If the hull is made of three points, the maximum density is reached at the location of the fourth point.
If the hull is a quadrilateral, split it into two triangles with the density maximum at the intersection of the two diagonals.
For each triangle, iterate over all pixels within the bounding box of the triangle.
If the pixel is within the triangle, linearly interpolate the projected density at this location.

chummels:
This looks awesome!  Excited to see this functionality.  Thanks for the work on this, @cphyc !

PR #4307, opened 2023-01-18
ENH: Allow to store multiple bitmap indices in the ewah-sidecar

Xarthisius:
## PR Summary

My attempt at fixing https://github.com/yt-project/yt/issues/3327. The main "upgrade" is that we can now store indices of the same order for different domain sizes, which we currently don't allow. Each cached index is identified by: `left_edge`, `right_edge`, hash of data file, `periodicity`, `index_order1`, `index_order2` and `nfiles`.
Drowback is that `h5py` is a strict requirement now.

## PR Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,6 @@ keywords = [
 requires-python = ">=3.8"
 dependencies = [
     "cmyt>=1.1.2",
-    "h5py>=3.1.0",
     "ipywidgets>=8.0.0",
     "matplotlib!=3.4.2,>=3.2", # keep in sync with tests/windows_conda_requirements.txt
     "more-itertools>=8.4",
@@ -76,6 +75,7 @@ answer-testing = "yt.utilities.answer_testing.framework:AnswerTesting"
 
 [project.optional-dependencies]
 # some generic, reusable constraints on optional-deps
+HDF5 = ["h5py>=3.1.0,<4.0.0"]
 netCDF4 = ["netCDF4!=1.6.1,>=1.5.3"]  # see https://github.com/Unidata/netcdf4-python/issues/1192
 Fortran = ["f90nml>=1.1"]
 
@@ -89,40 +89,40 @@ adaptahop = []
 ahf = []
 amrvac = ["yt[Fortran]"]
 art = []
-arepo = []
+arepo = ["yt[HDF5]"]
 artio = []
 athena = []
 athena-pp = []
 boxlib = []
 cf-radial = ["xarray>=0.16.1", "arm-pyart!=1.12.5,>=1.11.4"]
-chimera = []
-chombo = []
-cholla = []
-eagle = []
-enzo-e = ["libconf>=1.0.1"]
-enzo = ["libconf>=1.0.1"]
+chimera = ["yt[HDF5]"]
+chombo = ["yt[HDF5]"]
+cholla = ["yt[HDF5]"]
+eagle = ["yt[HDF5]"]
+enzo-e = ["yt[HDF5]", "libconf>=1.0.1"]
+enzo = ["yt[HDF5]", "libconf>=1.0.1"]
 exodus-ii = ["yt[netCDF4]"]
 fits = ["astropy>=4.0.1", "regions>=0.7"]
-flash = []
-gadget = []
-gadget-fof = []
-gamer = []
-gdf = []
-gizmo = []
-halo-catalog = []
+flash = ["yt[HDF5]"]
+gadget = ["yt[HDF5]"]
+gadget-fof = ["yt[HDF5]"]
+gamer = ["yt[HDF5]"]
+gdf = ["yt[HDF5]"]
+gizmo = ["yt[HDF5]"]
+halo-catalog = ["yt[HDF5]"]
 http-stream = ["requests>=2.20.0"]
-moab = []
+moab = ["yt[HDF5]"]
 nc4-cm1 = ["yt[netCDF4]"]
-open-pmd = []
-owls = []
-owls-subfind = []
+open-pmd = ["yt[HDF5]"]
+owls = ["yt[HDF5]"]
+owls-subfind = ["yt[HDF5]"]
 ramses = ["yt[Fortran]"]
 rockstar = []
 sdf = ["requests>=2.20.0"]
 stream = []
-swift = []
+swift = ["yt[HDF5]"]
 tipsy = []
-ytdata = []
+ytdata = ["yt[HDF5]"]
 
 # "full" should contain all optional dependencies intended for users (not devs)
 # in particular it should enable support for all frontends
@@ -199,7 +199,6 @@ mapserver = [
 ]
 minimal = [
     "cmyt==1.1.2",
-    "h5py==3.1.0",
     "ipywidgets==8.0.0",
     "matplotlib==3.2",
     "more-itertools==8.4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ keywords = [
 requires-python = ">=3.8"
 dependencies = [
     "cmyt>=1.1.2",
+    "h5py>=3.1.0",
     "ipywidgets>=8.0.0",
     "matplotlib!=3.4.2,>=3.2", # keep in sync with tests/windows_conda_requirements.txt
     "more-itertools>=8.4",
@@ -75,7 +76,6 @@ answer-testing = "yt.utilities.answer_testing.framework:AnswerTesting"
 
 [project.optional-dependencies]
 # some generic, reusable constraints on optional-deps
-HDF5 = ["h5py>=3.1.0,<4.0.0"]
 netCDF4 = ["netCDF4!=1.6.1,>=1.5.3"]  # see https://github.com/Unidata/netcdf4-python/issues/1192
 Fortran = ["f90nml>=1.1"]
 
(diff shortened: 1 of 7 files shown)


neutrinoceros:
We need to coordinate this and #2711
@matthewturk

Xarthisius:
> We need to coordinate this and #2711 @matthewturk

I don't think it's necessary. I'm not touching any ewah proper.

neutrinoceros:
I have very little expertise here but I don't fancy adding h5py as a hard requirement. However if it comes to that, this will require a couple edits to reflect it in `pyproject.toml`
Note that the Python 3.11 tests can't possibly run at the moment without building h5py from source, which isn't trivial.

pyproject.toml
```diff
@@ -43,6 +43,7 @@ keywords = [
 requires-python = ">=3.8"
 dependencies = [
     "cmyt>=1.1.2",
+    "h5py>=3.1.0,<4.0.0",
```
neutrinoceros, line 46: I think the `<4.0.0` part was probably me trying to sooth my anger for some reason I don't even remember. In hindsight I view this as an anti pattern and I'd rather we remove it if we're going to promote h5py as a hard requirement. You'll also want to add a the corresponding constraint to the `minimal` target.

neutrinoceros:
> I don't think it's necessary. I'm not touching any ewah proper.

My bad. Viewing on my phone, I got ahead of myself.

jzuhone:
FWIW I'm in favor of making h5py a hard requirement--long ago it used to be, and over ~90% of our functionality uses it in one way or another. 

Surely h5py wheels have been built by now for Python 3.11?

neutrinoceros (commented):
Genuine, possibly naive question: what's the benefit of using h5py here ? I think I understand the maintenance costs much better, so right now it's a little hard for me to judge if it is worth it.

matthewturk:
I can speak to a bit of the HDF5 advantage.  I see three -- the first is that it lets us consolidate into a single file what would otherwise have been spread across multiple.  The second is that we can have, in a forward-compatible way, metadata for the datasets.  And finally, it lets us ensure that we aren't doing anything platform-specific with the EWAH arrays, which we have guarded against but which *is* a possibility, given that the word-alignment may change on different machines.

yt/geometry/particle_geometry_handler.py
```diff
@@ -205,9 +183,7 @@ def _current_fname(order1, order2):
             self._initialize_coarse_index()
             self._initialize_refined_index()
             # We now update fname since index_order2 may have changed
-            fname, _ = _current_fname(
-                self.regions.index_order1, self.regions.index_order2
-            )
+            fname = getattr(ds, "index_filename", None) or f"{ds.parameter_filename}.ewah"
```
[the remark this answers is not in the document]
Xarthisius: Yeah, I don't think that `or` is pretty either, but I exactly wanted to avoid someone/something passing `None` or `''`.

jzuhone:
plus HDF5 makes it trivial for a user to inspect via other means such as `h5dump` etc

neutrinoceros:
Maybe I need to get into the details more, but all of that sounds achievable without it (to me). I get that it makes the code easier to write (?), but so far I'm not convinced it makes it easier to maintain, mostly because it puts a lot of maintenance cost the whole code base.

Just to make my case clear too, here are the costs I anticipate
- added complexity for testing dev versions of our dependencies (again, building h5py from source isn't as easy as numpy or matplotlib)
- new opportunities for blocking bugs when, e.g., numpy breaks h5py (this sort of conflict already happens quite often with our dependencies)
- we'll have to wait for months after the fact to even _test_ a new Python version properly, which to me would be a regression. Last year we were able to test Python 3.11 as early as beta 4 and I discovered a critical compilation error in yt much before it actually reached any user. This is stuff I care about.
- finally, and this is a more personal point, but as a user I tend to avoid `h5py` as much as I can, and I'm glad I don't need it for any of the stuff I do with yt.

I don't mean to block this, I just want to avoid that the decision be made lightly.

jzuhone:
@neutrinoceros,

> Maybe I need to get into the details more, but all of that sounds achievable without it (to me). 

In theory yes, but it's nowhere near as easy, and if the file was an unformatted binary instead, the increased complexity will make it more likely that one will make mistakes in trying to read either the data itself or the metadata. The whole point of having a format like HDF5 is that it makes it much easier to read without making mistakes. 

> I get that it makes the code easier to write (?)

it definitely makes it not only easier to write, but also to interpret. 

> but so far I'm not convinced it makes it easier to maintain, mostly because it puts a lot of maintenance cost on the whole code base.
> Just to make my case clear too, here are the costs I anticipate
> added complexity for testing dev versions of our dependencies (again, building h5py from source isn't as easy as numpy or matplotlib)
> new opportunities for blocking bugs when, e.g., numpy breaks h5py (this sort of conflict already happens quite often with our dependencies)
> we'll have to wait for months after the fact to even test a new Python version properly, which to me would be a regression. Last year we were able to test Python 3.11 as early as beta 4 and I discovered a critical compilation error in yt much before it actually reached any user. This is stuff I care about.

HDF5 is used almost everywhere in yt--not quite everywhere, of course, but most of the frontends use it in some way. Based on a quick scan, these are the ones that use it (at least in part):

* Enzo
* FLASH
* Gadget
* Arepo
* GDF
* Athena++
* Chimera
* Cholla
* Chombo
* Eagle
* Enzo-E
* GAMER
* GIZMO
* Moab
* OWLS
* Swift
* ytdata 

And we use it for file exporting in several places (yt arrays, fixed resolution buffers, data objects, etc). 

> finally, and this is a more personal point, but as a user I tend to avoid h5py as much as I can, and I'm glad I don't need it for any of the stuff I do with yt.

I think it's safe to say that as far as users go, those who can avoid HDF5 are very much in the minority. I don't think your experience is typical at all. Admittedly anecdotal, but several years back the developer of a major astrophysics hydrodynamics code who had avoided HDF5 for years finally gave in and re-wrote their entire I/O using it because it was just too convenient not to use, and doing so saved so much hassle for their users. 

As I noted previously, we had it as a hard dependency in the beginning. I don't remember exactly when we dropped it as such (@matthewturk or @Xarthisius, do you remember?). I recall thinking at the time that dropping it as a hard dependency was unnecessary.

I do not recall since I've started using yt (which has been since 2010) that there have been this many problems with `h5py`--it makes me suspect that this is a recent phenomenon with Python 3.11 and probably will not be a long-term issue. I admit being baffled why they don't have a wheel for it--Anaconda has a binary that works just fine. 

I take your points that there are some challenges, but I definitely think that they are fixable (either by us or `h5py`) and the benefits significantly outweigh the costs. Most of the frontends that need EWAH files use HDF5 anyway already.

jzuhone:
latest status of `h5py` and Python 3.11: https://github.com/h5py/h5py/issues/2146#issuecomment-1376496248

jzuhone:
we could even wait on this until h5py has their act together on wheels

neutrinoceros:
Thanks John for your detailed response.

> I think it's safe to say that as far as users go, those who can avoid HDF5 are very much in the minority. I don't think your experience is typical at all. 

granted !

> I do not recall since I've started using yt (which has been since 2010) that there have been this many problems with h5py--it makes me suspect that this is a recent phenomenon with Python 3.11 and probably will not be a long-term issue. 

I went back in their release history and it seems that this year is indeed an outlier. In recent years (excluding the current one)  wheels for new Python versions were published within a month after the final release. Assuming this better represents what can be typically expected, the situation wouldn't be as bad as I previously described, but we'd still loose the ability to test pre-releases (probably).


>  I definitely think that they are fixable (either by us or h5py) and the benefits significantly outweigh the costs.

well I honestly don't want to spend time studying the benefits too much so I'll trust you on that one.


Two more remarks I thought of in between comments:
- I wondered how much startup overhead this would add to `import yt`, and found that it's about 3% currently, so it's definitely acceptable
- This means we probably don't need to keep `yt.utilities.on_demand_imports._h5py_imports`, but we need to be careful about importing order between `h5py` and `netCDF4` see https://github.com/yt-project/yt/blob/4eda8b275e0e979f982b32bb700736315caf9f43/yt/utilities/on_demand_imports.py#L85-L94

matthewturk:
I'll chip in a bit more but I would suggest that we not change the h5py on-demand-import just yet

neutrinoceros:
why not ?

matthewturk:
Well, because hdf5 wouldn't strictly be a hard dependency outside of particle codes.  For instance, *idefix* wouldn't need it!

neutrinoceros:
No no, it _is_ a strict requirement here just as much as matplotlib: not having it installed will just break `import yt`. Besides, as I previously mentioned, the startup overhead it contributes is small enough that there's no real benefit to trying to delay it.  Even if we did, I don't think that it's feasible in a `.pyx` module.

matthewturk:
I mean, if we use the on-demand, that shouldn't be the case.  Right?

neutrinoceros:
Worth a shot. If it works, would that make every body happy ?

matthewturk:
So I actually think I want to step back and say, we should probably just have h5py be a hard dependency.  In particular, I think it would be extremely helpful to the majority of cases if `conda install yt` always got h5py.

neutrinoceros:
I understand that extra dependencies are not as nice in conda land, but we could have h5py be declared as a hard dependency _in conda builds_ and still keep it as optional for PyPI wheels. I think this would get us the best of both worlds, how about that ?

Xarthisius:
After 20+ comments about h5py, I wouldn't be offended if someone chimed in on whether what I did with ewah sidecar is the way to go, does it work for use cases I haven't tried and/or are we generally happy with that approach regardless of the technology used to store it...

yt/geometry/particle_oct_container.pyx
```diff
@@ -985,47 +994,38 @@ cdef class ParticleBitmap:
...
+            grp.attrs["file_hash"] = self.file_hash
+            grp.attrs["left_edge"] = self.left_edge
+            grp.attrs["right_edge"] = self.right_edge
+            grp.attrs["periodicity"] = self.periodicity
+            grp.attrs["index_order1"] = self.index_order1
+            grp.attrs["index_order2"] = self.index_order2
+
+            for ifile in range(self.nfiles):
+                serial_BAC = self.bitmasks._dumps(ifile)
+                grp.create_dataset(f"nfile_{ifile}", data=np.void(serial_BAC))
```
matthewturk, line 1026: Would you be opposed to making this 0-padded out to like 5 or something?  Maybe that's unnecessary, but it was the first thing I thought of when looking at `h5ls`.

matthewturk (dismissed):
I really like this change.

jzuhone:
I'll test this either tonight or tomorrow

yt/geometry/particle_oct_container.pyx
Xarthisius: fixed in https://github.com/yt-project/yt/pull/4307/commits/89d635a5282c6ddd4f7bac4d580379cf44103405

jzuhone:
@Xarthisius I checked this against a couple of recent use cases (including adding a bounding box) and it works very well!

jzuhone (approved):

neutrinoceros:
@Xarthisius is this still "in progress" ? Looks like we are good on reviews but I'm hesitant to push the button while "WIP" is in the title.

PR #4310, opened 2023-01-23
TST: bump CI from Python 3.10 to 3.11

neutrinoceros:
## PR Summary
Now that h5py 3.8 was released with Python 3.11 wheels, we can bump the bulk of CI to that version.

Note that this currently adds a few minutes of overhead as compared with Python 3.10 because a couple of our optional (some, indirect) dependencies still need to be built from source, but I'll give a hand upstream to remedy that on the medium term.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -27,7 +27,7 @@ jobs:
           windows-latest,
           ubuntu-latest,
         ]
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Most of the build overhead was due to pyart, but it now has wheels for Python 3.11 🎉

Xarthisius (approved):

PR #4311, opened 2023-01-26
TST: a couple improvements for ci_install.sh

neutrinoceros:
## PR Summary

Trying out a couple simplifications that I _think_ should be possible now.

### DIFF
--- a/tests/ci_install.sh
+++ b/tests/ci_install.sh
@@ -34,8 +34,6 @@ if [[ "${RUNNER_OS}" == "Windows" ]] && [[ ${dependencies} != "minimal" ]]; then
     while read requirement; do conda install --yes $requirement; done < tests/windows_conda_requirements.txt
 else
     python -m pip install --upgrade pip
-    python -m pip install --upgrade wheel
-    python -m pip install --upgrade setuptools
 fi
 
 # Step 2: install deps and yt
--- a/tests/ci_install.sh
+++ b/tests/ci_install.sh
@@ -32,6 +32,9 @@ if [[ "${RUNNER_OS}" == "Windows" ]] && [[ ${dependencies} != "minimal" ]]; then
     # keep in sync: setup.cfg
     while read requirement; do conda install --yes $requirement; done < tests/windows_conda_requirements.txt
 else
+    # upgrading pip to guarantee installing extra dependencies with [full] is supported
+    # this is only necessary for some versions of Python 3.8 and 3.9
+    # see https://github.com/yt-project/yt/issues/4270
     python -m pip install --upgrade pip
 fi
 
--- a/tests/ci_install.sh
+++ b/tests/ci_install.sh
@@ -47,7 +47,7 @@ elif [[ ${dependencies} == "full" ]]; then
     # this is required for cartopy. see
     # https://scitools.org.uk/cartopy/docs/latest/installing.html?highlight=install#building-from-source
     python -m pip install shapely --no-binary=shapely
-    CFLAGS="$CFLAGS -DACCEPT_USE_OF_DEPRECATED_PROJ_API_H" python -m pip install -e ".[test,full]"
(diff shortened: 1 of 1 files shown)


tests/ci_install.sh
```diff
@@ -34,26 +33,24 @@ if [[ "${RUNNER_OS}" == "Windows" ]] && [[ ${dependencies} != "minimal" ]]; then
...
-    python -m pip install -e .[test,minimal]
+    python -m pip install -e ".[test,minimal]"
 elif [[ ${dependencies} == "full" ]]; then
     # test with all optional runtime dependencies
 
     # this is required for cartopy. see
     # https://scitools.org.uk/cartopy/docs/latest/installing.html?highlight=install#building-from-source
     python -m pip install shapely --no-binary=shapely
-    CFLAGS="$CFLAGS -DACCEPT_USE_OF_DEPRECATED_PROJ_API_H" python -m pip install -e .[test,full]
+    CFLAGS="$CFLAGS -DACCEPT_USE_OF_DEPRECATED_PROJ_API_H" python -m pip install -e ".[test,full]"
```
neutrinoceros, line 50: It's probably worth trying this simplification too by now
```suggestion
    python -m pip install -e ".[test,full]"
```

tests/ci_install.sh
```diff
@@ -34,26 +33,24 @@ if [[ "${RUNNER_OS}" == "Windows" ]] && [[ ${dependencies} != "minimal" ]]; then
     while read requirement; do conda install --yes $requirement; done < tests/windows_conda_requirements.txt
 else
     python -m pip install --upgrade pip
```
neutrinoceros, line 35: I should also add a comment here why we upgrade pip (it's needed in most Python versions to support `pip install .[full]`, see #4270)

tests/ci_install.sh
```diff
@@ -20,10 +20,9 @@ osx|macOS)
 esac
 
 # Disable excessive output
-mkdir -p $HOME/.config/yt
-echo "[yt]" > $HOME/.config/yt/yt.toml
-echo "suppress_stream_logging = true" >> $HOME/.config/yt/yt.toml
-cat $HOME/.config/yt/yt.toml
+echo "[yt]\nlog_level = 50" > yt.toml
```
matthewturk, line 23: This is a change from before, yes?  Previously we turned off the stream, and now we're setting the level to 50.  This *may* be a distinction without a difference.

tests/ci_install.sh
```diff
@@ -33,27 +32,28 @@ if [[ "${RUNNER_OS}" == "Windows" ]] && [[ ${dependencies} != "minimal" ]]; then
...
 # but the primary intention is to embed this script in CI jobs
 if [[ ${dependencies} == "minimal" ]]; then
     # test with minimal versions of runtime dependencies
-    python -m pip install -e .[test,minimal]
+    python -m pip install -e ".[test,minimal]"
 elif [[ ${dependencies} == "full" ]]; then
     # test with all optional runtime dependencies
 
     # this is required for cartopy. see
     # https://scitools.org.uk/cartopy/docs/latest/installing.html?highlight=install#building-from-source
```
matthewturk, line 51: does this comment need updating?

matthewturk (approved):

tests/ci_install.sh
neutrinoceros: yes it's a intentional change of behaviour. I don't think we have any errors logged in tests but if we do, I figure we might give ourselves a chance to see them

tests/ci_install.sh
neutrinoceros: It doesn't. Actually it's the linked documentation that's out of date (see https://github.com/yt-project/yt/pull/4311#pullrequestreview-1271134788) ;-)

neutrinoceros:
@matthewturk are you satisfied with my responses ?

PR #4316, opened 2023-01-28
check for `_unit_system_name` in `save_as_dataset`

### LINKED ISSUE #4315: reloading after `save_as_dataset` fails
### Bug report

Originally reported on slack: if you `save_as_dataset()` then reload in a new python session, it will error for `athena_pp` datasets (and maybe others?). 

**Code for reproduction**

```python
import yt
fi = "AM06/AM06.out1.00300.athdf"
ds = yt.load(fi)
prj = yt.ProjectionPlot(ds, 'z', ('athena_pp', 'rho'), center='c')
prj.data_source.save_as_dataset()
```
will save `./AM06.out1.00300_quad_proj.h5`. 

If you restart your python kernel and try

```python
import yt
ds = yt.load('AM06.out1.00300_quad_proj.h5')
```

Then you get an error

**Actual outcome**

```
KeyError                                  Traceback (most recent call last)
Input In [1], in <cell line: 2>()
      1 import yt
----> 2 ds_2 = yt.load('AM06.out1.00300_quad_proj.h5')

File ~/src/yt_general/yt/yt/_maintenance/deprecation.py:73, in future_positional_only.<locals>.outer.<locals>.inner(*args, **kwargs)
     65     value = kwargs[name]
     66     issue_deprecation_warning(
     67         f"Using the {name!r} argument as keyword (on position {no}) "
     68         "is deprecated. "
   (...)
     71         **depr_kwargs,
     72     )
---> 73 return func(*args, **kwargs)

File ~/src/yt_general/yt/yt/loaders.py:123, in load(fn, hint, *args, **kwargs)
    120 candidates = find_lowest_subclasses(candidates, hint=hint)
    122 if len(candidates) == 1:
<!--A description of the expected outcome from the code snippet-->
<!--If this used to work in an earlier version of yt, please note the
version it used to work on-->

**Version Information**
<!--Please specify your platform and versions of the relevant libraries you are
using:-->
  * Operating System:
  * Python Version:
  * yt version:

<!--Please tell us how you installed yt and python e.g., from source,
pip, conda. If you installed from conda, please specify which channel you used
if not the default-->

--> 123     return candidates[0](fn, *args, **kwargs)
    125 if len(candidates) > 1:
    126     raise YTAmbiguousDataType(fn, candidates)

File ~/src/yt_general/yt/yt/frontends/ytdata/data_structures.py:366, in YTSpatialPlotDataset.__init__(self, *args, **kwargs)
    365 def __init__(self, *args, **kwargs):
--> 366     super().__init__(*args, dataset_type="ytspatialplot_hdf5", **kwargs)

File ~/src/yt_general/yt/yt/frontends/ytdata/data_structures.py:257, in YTDataContainerDataset.__init__(self, filename, dataset_type, index_order, index_filename, units_override, unit_system)
    255 self.index_order = validate_index_order(index_order)
    256 self.index_filename = index_filename
--> 257 super().__init__(
    258     filename,
    259     dataset_type,
    260     units_override=units_override,
    261     unit_system=unit_system,
    262 )

File ~/src/yt_general/yt/yt/data_objects/static_output.py:302, in Dataset.__init__(self, filename, dataset_type, units_override, unit_system, default_species_fields, axis_order)
    298     used_unit_system = unit_system
    300 self._create_unit_registry(used_unit_system)
--> 302 self._parse_parameter_file()
    303 self.set_units()
    304 self.setup_cosmology()

File ~/src/yt_general/yt/yt/frontends/ytdata/data_structures.py:369, in YTSpatialPlotDataset._parse_parameter_file(self)
    368 def _parse_parameter_file(self):
--> 369     super()._parse_parameter_file()
    370     if self.parameters["container_type"] == "proj":
    371         if (
    372             isinstance(self.parameters["weight_field"], str)
    373             and self.parameters["weight_field"] == "None"
    374         ):

File ~/src/yt_general/yt/yt/frontends/ytdata/data_structures.py:265, in YTDataContainerDataset._parse_parameter_file(self)
    264 def _parse_parameter_file(self):
--> 265     super()._parse_parameter_file()
    266     self.particle_types_raw = tuple(self.num_particles.keys())
    267     self.particle_types = self.particle_types_raw

File ~/src/yt_general/yt/yt/frontends/ytdata/data_structures.py:99, in SavedDataset._parse_parameter_file(self)
     97     unit_system = "cgs"
     98 # reset unit system since we may have a new unit registry
---> 99 self._assign_unit_system(unit_system)
    101 # assign units to parameters that have associated unit string
    102 del_pars = []

File ~/src/yt_general/yt/yt/data_objects/static_output.py:1297, in Dataset._assign_unit_system(self, unit_system)
   1293     mag_dims = None
   1294 if unit_system != "code":
   1295     # if the unit system is known, we can check if it
   1296     # has a "current_mks" unit
-> 1297     us = unit_system_registry[str(unit_system).lower()]
   1298     mks_system = us.base_units[current_mks] is not None
   1299 elif mag_dims and current_mks in mag_dims:
   1300     # if we're using the not-yet defined code unit system,
   1301     # then we check if the magnetic field unit has a SI
   1302     # current dimension in it

KeyError: '2493b7e1126463ce3bae781a48ef579b'
```

**Expected outcome**

reload the dataset

chrishavlin:
Closes #4315 

There might be a better way to fix this, I'm not super familiar with the unit system attributes and how they're handled on re-load. But I wanted to get this in before the weekend.


yt/frontends/ytdata/utilities.py
```diff
@@ -103,7 +103,12 @@ def save_as_dataset(ds, filename, data, field_types=None, extra_attrs=None):
         _yt_array_hdf5_attr(fh, "unit_registry_json", ds.unit_registry.to_json())
 
     if hasattr(ds, "unit_system"):
-        _yt_array_hdf5_attr(fh, "unit_system_name", ds.unit_system.name.split("_")[0])
+        if hasattr(ds, "_unit_system_name"):
+            _yt_array_hdf5_attr(fh, "unit_system_name", ds._unit_system_name)
+        else:
```
chrishavlin, line 108: I suspect this could be tidied up, I can look more closely on monday :)

yt/frontends/ytdata/utilities.py
```diff
@@ -103,7 +103,12 @@ def save_as_dataset(ds, filename, data, field_types=None, extra_attrs=None):
...
+            _yt_array_hdf5_attr(
+                fh, "unit_system_name", ds.unit_system.name.split("_")[0]
+            )
```
brittonsmith, line 111: Maybe something like this? Note, I haven't tested if it solves the problem, just a code suggestion.




```suggestion
        name = getattr(ds, "_unit_system_name", ds.unit_system.name.split("_")[0])
        _yt_array_hdf5_attr(fh, "unit_system_name", name)
```

brittonsmith (commented):
Is this happening because of something special about the base dataset? Why does it not have a unit system name? It might be worth understanding this before this goes in.

chrishavlin:
> Is this happening because of something special about the base dataset? Why does it not have a unit system name? It might be worth understanding this before this goes in. 

I agree! Should have marked it as a WIP on Friday when I pushed it up :)

But I think I figured it out after a bit more digging...

The following

```python
import yt
ds = yt.load("AM06/AM06.out1.00300.athdf")
print(ds.unit_system.name)
```
prints out `'2493b7e1126463ce3bae781a48ef579b'`

That string comes from how the code `UnitSystem` is created:  `Dataset._assign_unit_system`) calls `yt.units.unit_systems.create_code_unit_system` which looks like:

```python
def create_code_unit_system(unit_registry, current_mks_unit=None):
    code_unit_system = UnitSystem(
        name=unit_registry.unit_system_id,
        length_unit="code_length",
        mass_unit="code_mass",
        time_unit="code_time",
        temperature_unit="code_temperature",
        current_mks_unit=current_mks_unit,
        registry=unit_registry,
    )
    <----------- TRIMMED ---------------> 
```

That `unit_registry.unit_system_id` attribute being used to set the name is a hashed property set by `unyt.unit_regsitry.UnitRegistry` (the UnitRegsitry docstring reads: "This is a unique identifier for the unit registry created from a FNV hash."). 

The history around that `name=` line is fairly confusing -- seems to me it was changed temporarily in #2728 (in  [commit `df5bc0336f6f07816290ed3754d236dddeb7ab01`](https://github.com/yt-project/yt/commit/df5bc0336f6f07816290ed3754d236dddeb7ab01) ) to 

```python
    name="code_{}".format(unit_registry.unit_system_id),
```
which actually would fix the bug here since when you call `save_as_dataset,  it would split it and save as `code`: 

```python
        _yt_array_hdf5_attr(
            fh, "unit_system_name", ds.unit_system.name.split("_")[0]
        )
```
but then the `name` was changed back to just the UnitSystem hash while that PR was iterated on. 

The reason why you need to restart the session to get the error is because `yt.unit_system_registry` will get re-initialized only when you initially import yt. So when you load a dataset, you get a unique hash for that dataset's code length `UnitSystem`:

```python
import yt
print(yt.unit_system_registry.keys())
ds = yt.load("AM06/AM06.out1.00300.athdf")
print(yt.unit_system_registry.keys())
```
prints
```
dict_keys(['cgs', 'mks', 'imperial', 'galactic', 'solar', 'geometrized', 'planck'])
dict_keys(['cgs', 'mks', 'imperial', 'galactic', 'solar', 'geometrized', 'planck', '2493b7e1126463ce3bae781a48ef579b'])
```
but when you restart your kernel, `yt.unit_system_registry` is reset, leading to a key error when trying to re-load the file that was saved with `prj.data_source.save_as_dataset()` (since that unique `UnitSystem` hash no longer exists). 


So in the end, I think my fix here might actually be a good approach. I'm also still thinking about how to write a test for this -- it's tricky because it's session-scope dependent. There might be a nice way to do it with pytest...

yt/frontends/ytdata/utilities.py
chrishavlin: After the digging this morning, I thinkg it might also be totally sufficient only write the `_unit_system_name` attribute

```python
_yt_array_hdf5_attr(fh, "unit_system_name", ds._unit_system_name)
```
I initially wasn't 100% sure that having `unit_system` ensures always having `_unit_system_name`, but I suspect it might be the case. I'll check on that more thoroughly...

chrishavlin:
changed to WIP for now while I test things out more carefully -- feel free to continue commenting though! I just didn't want it to get merged inadvertently...

neutrinoceros:
> The history around that name= line is fairly confusing -- seems to me it was changed temporarily in https://github.com/yt-project/yt/pull/2728 (in [commit df5bc0336f6f07816290ed3754d236dddeb7ab01](https://github.com/yt-project/yt/commit/df5bc0336f6f07816290ed3754d236dddeb7ab01) ) to
>
>```python
>    name="code_{}".format(unit_registry.unit_system_id),
>```
>which actually would fix the bug here since when you call save_as_dataset, it would split it and save as code:
>
>```python
>        _yt_array_hdf5_attr(
>            fh, "unit_system_name", ds.unit_system.name.split("_")[0]
>        )
>but then the name was changed back to just the UnitSystem hash while that PR was iterated on.

I think we completely missed this in #2728, great job finding your way back there !

If just re-adding the `code_` prefix fixes the bug I think this is the way to go here. I think adding a couple comments on the `.split` call as well as the line where the prefix is added, linking to #4315 would be sufficient of a guard to prevent this bug from reappearing, if writing a test for it is as complex as it sounds.

chrishavlin:
> If just re-adding the code_ prefix fixes the bug I think this is the way to go here.

I agree! It is simpler, and it does fix the bug here. I did add a minimal unit test -- it's not a full test of the bug, but it at least checks that loading back in a saved dataset has the proper `unit_system.name` (which fails on main).

yt/frontends/ytdata/utilities.py
chrishavlin: I ended up changing how the unit `unit_system.name` is initially set

chrishavlin:
Ok, so turns out switching the name to include the `code` prefix led to some tricky test failures internal to `unyt` (in `unit_registry._sanitize_unit_system`. So I went back to change which attribute to write from within `save_as_dataset`.

neutrinoceros:
If tests pass this time, I suggest rebasing the branch to a single commit so the history will hopefully be easier to follow in the future

neutrinoceros (dismissed):
LGTM, thanks @chrishavlin !
I'll give @brittonsmith some time to respond before I merge

chrishavlin:
pre-commit.ci autofix

neutrinoceros (dismissed):

neutrinoceros (approved):

PR #4319, opened 2023-01-30
MNT: drop useless workflow

neutrinoceros:
## PR Summary

This workflow, while pretty harmless, is also very much useless: it hasn't caught a single problem in ~4.5k runs (over the course ~3 yrs).


matthewturk:
If I remember correctly, this was part of a couple where we checked for code smells, right?  I mean, I'm OK with dropping it, but is it a big deal time-wise?

neutrinoceros:
It takes about a couple seconds on each PR, so no, it's not costly, but IMO it still takes too much space for what little value (0) it actually provides.

matthewturk:
I'm OK with dropping it, to be clear.

matthewturk (approved):

matthewturk:
can't wait for my forthcoming PR that imports all our import-on-demand modules sneakily at the top level

PR #4320, opened 2023-01-31
STY: migrate linting to ruff

### LINKED ISSUE #4203: STY: migrate to ruff ?
I recently learned about [ruff](https://github.com/charliermarsh/ruff), which is an exciting linter basically reimplementing a subset of `flake8`+`flake8-bugbear`+`pyupgrade` (and more) in Rust, which allows it to be x100 faster than the tools it replaces. On top of that, it implements some auto-fixing for certain flake8 errors. It's also intentionally leaving code style out of linting and delegates that part to black, which matches our use case.

Currently, running `pre-commit run --all-files` takes about a minute and half of that time is taken by flake8 alone, so it would help making linting+fixing noticeably more efficient.

Currently the cost of migrating would mainly be the loss of certain features from replaced tools that we care about, but the tool is very new and there's a good chance that the gap narrows down in the future.

The actual migration would be (almost) as trivial as adding a new pre-commit hook and remove the ones it replaces.

### Hooks that ruff could replace
- [ ] flake8
  - [x] pyflake (as far as I can tell it looks complete)
  - [x] pycodestyle (a lot of rules are not implemented, but it seems to be intentional: those would be fixed by black anyway)
  - [x] flake8-2020 (https://github.com/charliermarsh/ruff#flake8-2020-ytt)
  - [ ] flake8-bugbear (https://github.com/charliermarsh/ruff#flake8-bugbear-b)
     - [ ] B001
     - [ ] B902
  - [x] flake8-logging-format (https://github.com/charliermarsh/ruff/issues/1850)
- [x] pyupgrade (https://github.com/charliermarsh/ruff/issues/827)
- [x] isort (not completely clear how to follow progress, but see https://github.com/charliermarsh/ruff#isort-i)

nbqa hooks for flake8, isort on notebooks can supposedly be replaced by `nbqa-ruff`

neutrinoceros:
## PR Summary

Fix #4203

This brings `pre-commit run --all-files` down to 5s on my machine (as opposed to 22s on the main branch), and pre-commit.ci runs complete in 30s instead of 60s

ruff is not completely stable yet but many big projects are already using it (pandas, scipy, jupyter, to name a few), and I wanted to see how big of an effort migration would actually be.
The patch I end up with is small enough to my taste that I feel confident to submit it for others to consider.

I discovered that ruff is slightly more agressive than pyupgrade on a couple rules, especially when it comes to replacing old string formatting (`%`) with more modern equivalents, but even at the scale of our code base adoption appears straightforward !


yt/funcs.py
```diff
@@ -1191,14 +1191,13 @@ def validate_object(obj, data_type):
 
 def validate_axis(ds, axis):
     if ds is not None:
-        valid_axis = ds.coordinates.axis_name.keys()
+        valid_axis = sorted(
+            list(ds.coordinates.axis_name.keys()), key=lambda k: str(k).swapcase()
```
matthewturk, line 1195: What is this doing?

yt/tests/test_funcs.py
```diff
@@ -16,11 +16,10 @@ def test_validate_axis():
...
         ds.slice("r", 0.25)
-    desired = (
-        "Expected axis of int or char type (can be "
-        "[0, 'x', 'X', 1, 'y', 'Y', 2, 'z', 'Z']), received 'r'."
-    )
-    assert_equal(str(ex.exception)[:40], desired[:40])
+    desired = "Expected axis to be any of [0, 1, 2, 'x', 'y', 'z', 'X', 'Y', 'Z'], received 'r'"
+
+    actual = str(ex.exception)
+    assert actual == desired
```
matthewturk, line 22: did you manually change this?  why?

matthewturk (approved):
I'm still pretty ambivalent about switching right now, even if there are other players doing it..

matthewturk (commented):
Sorry, didn't mean to approve, just meant to leave a deflating comment.

yt/funcs.py
neutrinoceros: sorting keys. When sorting strings, uppercase comes before lower case, so I'm reverting that order with a custom `lambda` key to match how the default list (`[0, 1, 2, "x", "y", "z", "X", "Y", "Z"]`) is ordered.

yt/tests/test_funcs.py
neutrinoceros: `ruff` automatically changed a couple strings originally formatted with the `%` operator to use the `str.format` method instead. I did a second, manual, pass to use fstrings instead and took the opportunity to slightly improve some affected error messages. Here I'm changing the test to match the -intentionally- updated message.

yt/funcs.py
neutrinoceros: Yes. Currently we rely on `pyupgrade` for this stuff, and it's documented as "timid" about string formatting. `ruff` is a little less so.

neutrinoceros:
> I'm still pretty ambivalent about switching right now, even if there are other players doing it..

This is fine, but can you be explicit about what conditions should be met for you to be confortable with the migration ?
AFAIC I just don't want to end up with yet another rotting PR.

matthewturk:
I think when you will self-describe it as "Stable" rather than just "Used."

I will confess also to having some linting/CI fatigue.  I am trying to keep on top of things, but I also really want to see enhancements along other directions as *well* as linting, and it's hard to keep my mental energy pushing outwards along too many different vectors.  So I guess that's what's coloring my ambivalence, but at the same time I don't want to just say no to a (mostly) no-op speedup.

neutrinoceros:
Perfectly understandable.
For reference, the vision for a stable version of ruff is laid out in the following comment: 
https://github.com/charliermarsh/ruff/issues/1992#issuecomment-1405497643

I should clarify that what isn't stable is the *API*, which only affects writing additional code for ruff itself, but the user interface has been extremely stable for a project that hasn't reached version 0.1 yet.

> I will confess also to having some linting/CI fatigue. I am trying to keep on top of things, but I also really want to see enhancements along other directions as well as linting, and it's hard to keep my mental energy pushing outwards along too many different vectors.

Somewhat funnily, that's one of my reasons to want to migrate to ruff: it would drastically reduce the number of linting dependencies. I was also thinking, unrelated to ruff, that the time seemed ripe for us to slowdown the pace of pre-commit auto-updates (which I'm happy to do in a separate PR !)

neutrinoceros:
@matthewturk I'll dismiss your review so this doesn't get merged inadvertently.

neutrinoceros:
I forgot to adapt CONTRIBUTING.rst, switching to draft for now

neutrinoceros:
I would like @jzuhone to sign off since he also expressed reserves.
All I can add to my case now is that I've switched almost all of my projects to ruff recently and so far it's been nothing but absolute joy.

jzuhone:
I'm ok with this.

jzuhone (approved):

PR #4323, opened 2023-02-03
Add `Parthenon` frontend

pgrete:
## PR Summary

Add `yt` frontend for codes based on the `Parthenon` AMR framework (https://github.com/parthenon-hpc-lab/parthenon) including AthenaPK (https://github.com/parthenon-hpc-lab/athenapk), Phoebus (https://github.com/lanl/phoebus) or KHARMA (https://github.com/AFD-Illinois/kharma)
Note that the mesh structure in `Parthenon` is heavily derived from `Athena++`, so that's why we used the `Athena++` frontend as baseline.

We've been using this frontend for quite some time now (without any recent major changes) so we consider it stable enough to open it for review and merge to `yt` proper.

At this point I'm looking for
- general feedback
- feedback on how to generally handle/differentiate between Parthenon (as framework) and downstream code related functions/fields/... (the current version of this frontend is geared towards AthenaPK and I'd like to abstract this)
- (assuming a test dataset is desired), how to upload/handle/provide the dataset

## PR Checklist


matthewturk:
This is awesome!  Thank you for contributing it.  I'll provide a review ASAP.

neutrinoceros:
Congratz @pgrete ! I'll make a first pass for QA and will try to focus on stuff that isn't already reported by linters (see pre-commit.ci's report).

pyproject.toml
```diff
@@ -117,6 +117,7 @@ nc4-cm1 = ["yt[netCDF4]"]
 open-pmd = ["yt[HDF5]"]
 owls = ["yt[HDF5]"]
 owls-subfind = ["yt[HDF5]"]
+parthenon = ["yt[HDF5]"]
```
neutrinoceros, line 120: I can't find where h5py is actually used in the frontend, could you clarify the intent here ?

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+        )
+
+    @classmethod
+    def _is_valid(cls, filename, *args, **kwargs):
+        try:
+            if filename.endswith(".phdf") or filename.endswith(".rhdf"):
+                return True
+        except Exception:
+            pass
+        return False
```
neutrinoceros, line 327: No need for complexity here
```suggestion
    def _is_valid(cls, filename:str, *args, **kwargs) -> bool:
        return filename.endswith((".phdf", ".rhdf"))
```

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+            )
+
+        self.parameters["He_mass_fraction"] = He_mass_fraction
+        self.parameters["H_mass_fraction"]  = H_mass_fraction
+
+        self.current_redshift = (
+            self.omega_lambda
+        ) = (
+            self.omega_matter
+        ) = self.hubble_constant = self.cosmological_simulation = 0.0
```
neutrinoceros, line 308: readability ++
(and the `cosmological_simulation` attribute should be an `int`)
```suggestion
        self.current_redshift = 0.0
        self.omega_lambda = 0.0
        self.omega_matter = 0.0
        self.hubble_constant = 0.0
        self.cosmological_simulation = 0
```

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+            #        raise Exception("Inconsistent periodicity in boundary conditions")
+
+            self._periodicity = tuple((bc == "periodic" for bc in inner_bcs))
+
+
+        gamma_attrs = {key:val for key, val in self._handle["Params"].attrs.items() if key.endswith("AdiabaticIndex")} 
+        if "gamma" in self.specified_parameters:
+            self.gamma = float(self.specified_parameters["gamma"])
+        elif len(gamma_attrs) >= 1 :
+            self.gamma = list(gamma_attrs.values())[0]
```
neutrinoceros, line 284: avoid creating a list just to get the first element
```suggestion
            self.gamma = next(iter(gamma_attrs.values()))
```

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,334 @@
...
+        else:
+            self.gamma = 5./3.
+
+
+        He_mass_fraction_attrs = {key:val for key, val in self._handle["Params"].attrs.items() if key.endswith("He_mass_fraction")} 
+        if len(He_mass_fraction_attrs) >= 1 :
+            He_mass_fraction = list(He_mass_fraction_attrs.values())[0]
```
neutrinoceros, line 291:
```suggestion
            He_mass_fraction = next(iter(He_mass_fraction_attrs.values()))
```

yt/frontends/parthenon/tests/test_outputs.py
```diff
@@ -0,0 +1,76 @@
+import numpy as np
```
neutrinoceros, line 1: looks like this module was copied straight from the Athena++ frontend with no modification. It's a good idea to use it as a basis but we'll need to iterate on this

neutrinoceros:
> - feedback on how to generally handle/differentiate between Parthenon (as framework) and downstream code related functions/fields/... (the current version of this frontend is geared towards AthenaPK and I'd like to abstract this)

I think we have one example of a framework-as-a-dataformat already: the `boxlib` frontend. In my experience, t's indeed the one for which automatic determination of exact format is hardest to implement and maintain, and the reason why we added a `hint` keyword argument to `yt.load`. I hope we'll be able to have a cleaner distinction between formats with this frontend but if the binary data format is to be identical I think one practical approach is to reserve some header in binary files to encode some unique key (like the name of the code) that maps to a known format. Otherwise it may be impossible to guess.

At this point, and I'm not saying that to discourage this pull request, you may want to consider [distributing the frontend as an extension package](https://yt-project.org/docs/dev/developing/extensions.html#frontends-as-extensions), which can help if frequent contributions are expected and you want to be free from our relatively slow release cycle.

> - (assuming a test dataset is desired), how to upload/handle/provide the dataset

yes that would be a necessary step for us to merge this PR.
The process requires a companion PR to our website repo. The most recent example of such a PR would be https://github.com/yt-project/website/pull/115

pyproject.toml
pgrete: We directly use  `self._handle = HDF5FileHandler(filename)` rather than explicit calls to `h5py`.
Does this implicit use mean, `hdf5` should not be set as a dependency?
Parthenon currently only supports output in HDF5 so we definitely rely on this lib/module.

neutrinoceros: `HDF5FileHandler` indeed hard-requires `h5py` so you are absolutely right to list it as a dependency, I just missed that. Thank you !

yt/frontends/parthenon/data_structures.py
pgrete: Done.

yt/frontends/parthenon/data_structures.py
pgrete: done.

yt/frontends/parthenon/data_structures.py
pgrete: done

matthewturk:
@pgrete I think as long as it doesn't require `hdf5plugin` it should be OK to use deflate.

pgrete:
> I think that the reason CI is now failing is that #4344 was merged in between runs. Add `test_outputs.py` to the list of files that pytest should ignore to resolve this.

I'm currently figuring out some useful tests to add (with the appropriate data file).

pgrete:
Looks like I'm stuck.
How can I actually run tests?

I first thought that the test are answer tests so that I should call them using a class (like `TestTipsy` described here https://yt-project.org/doc/developing/testing.html#how-to-run-the-answer-tests) but then I realized that the current `test_outputs.py` does not define a class (or I don't see where it does) -- similar to other `test_outputs.py` of other frontends.

Then I tried `pytest` (after reinstalling the project to fix dependencies) and only got 
```
$ pytest yt/frontends/parthenon/tests
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.10.9, pytest-7.2.2, pluggy-1.0.0 -- /usr/bin/python
cachedir: .pytest_cache
rootdir: /home/pgrete/src/yt, configfile: pyproject.toml
plugins: typeguard-2.13.3, anyio-3.6.2
collected 0 items                                                                                                                                                                            

=================================================================================== no tests ran in 0.01s ====================================================================================
```
so what am I missing?

pgrete:
I'm a little stuck here.
I realized that the failing test was run with `py38` so I thought that my python environment may be too new (also wrt the nose PR for python >3.10).

So I installed a clean env with python 3.8 and the "full" version of yt `pip install -e .[full]`

Trying the test now results in:

```bash
$ pytest yt/frontends/parthenon/tests/test_outputs.py 
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.8.16, pytest-7.2.2, pluggy-1.0.0 -- /home/pgrete/miniconda3/envs/py3.8/bin/python
cachedir: .pytest_cache
rootdir: /home/pgrete/src/yt, configfile: pyproject.toml
collected 0 items / 1 error                                                                                                                                                                  

=========================================================================================== ERRORS ===========================================================================================
_______________________________________________________________ ERROR collecting yt/frontends/parthenon/tests/test_outputs.py ________________________________________________________________
yt/frontends/parthenon/tests/test_outputs.py:11: in <module>
    from yt.utilities.answer_testing.framework import (
yt/utilities/answer_testing/framework.py:24: in <module>
    from nose.plugins import Plugin
../../miniconda3/envs/py3.8/lib/python3.8/site-packages/nose/__init__.py:1: in <module>
    from nose.core import collector, main, run, run_exit, runmodule
../../miniconda3/envs/py3.8/lib/python3.8/site-packages/nose/core.py:12: in <module>
    from nose.loader import defaultTestLoader
../../miniconda3/envs/py3.8/lib/python3.8/site-packages/nose/loader.py:21: in <module>
    from nose.importer import Importer, add_path, remove_path
../../miniconda3/envs/py3.8/lib/python3.8/site-packages/nose/importer.py:12: in <module>
    from imp import find_module, load_module, acquire_lock, release_lock
../../miniconda3/envs/py3.8/lib/python3.8/imp.py:31: in <module>
    warnings.warn("the imp module is deprecated in favour of importlib; "
E   DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
================================================================================== short test summary info ===================================================================================
ERROR yt/frontends/parthenon/tests/test_outputs.py - DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
====================================================================================== 1 error in 0.29s ======================================================================================

```

What am I missing?

neutrinoceros:
Sorry for the late reply. `GenericArrayTest` is a subclass of `AnswerTestingTest` and as such, it _cannot_ be run with pytest yet (we are still in the middle of the migration process).
As a result, your tests can only be run with Python 3.8 or 3.9 (since nose isn't compatible with Python 3.10 and later).

pgrete:
So I'd call `python tests/nose_runner.py` (as that's the command being run by the `py38` GitHub check).
And, if yes, can I just run the Parthenon frontend or do I need to run all tests?

pgrete:
I figured it out (for reference `nosetests -v --with-answer-testing --local --local-dir $HOME/src/yt-test-data --answer-store --answer-name=local-parthenon yt.frontends.parthenon`). Going back to the old doc helped.

I just added a plain Parthenon 2D test (which of course will fail given that the data is not available online).

What do the next steps look like in general (in terms of when to upload/open a PR for the reference files) and the combined review?

With respect to this PR, I plan to make the following two additions:
- a 3D example (of the 2D type above) but with dx != dy != dz
- a 3D AthenaPK example (incl changes to work with a code hint) with support for units

Do you recommend any other additional tests?
Is sth else required to complete this PR?

neutrinoceros:
> What do the next steps look like in general (in terms of when to upload/open a PR for the reference files) and the combined review?

AFAICT it seems mature enough that submitting datasets for upload now wouldn't be inappropriate.

> Do you recommend any other additional tests?

I think that frontend tests should be as focused as possible on things that might go wrong _specifically_ with the data format at hand. Tests exposing problems encountered during development are also very welcome, but don't feel like you need to be 100% exhaustive.

> Is sth else required to complete this PR?

Eventually we'll want to have 2 approvals from maintainers, but you're close to getting mine already, since my initial comments were all addressed ! I'll need to review again, and I'm ready to do so as soon as needed. Feel free to ping me if you'd rather get my review before dataset are uploaded, otherwise I'll just wait for tests to go green !

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,342 @@
...
+            key: val
+            for key, val in self._handle["Params"].attrs.items()
+            if key.endswith("AdiabaticIndex")
+        }
+        if "gamma" in self.specified_parameters:
+            self.gamma = float(self.specified_parameters["gamma"])
+        elif len(gamma_attrs) >= 1:
+            self.gamma = next(iter(gamma_attrs.values()))
+        else:
+            self.gamma = 5.0 / 3.0
```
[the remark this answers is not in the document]
forrestglines: It's due to the actual parameter being `Hydro/AdiabaticIndex` in AthenaPK which will be different for the other Parthenon codes. 

If this logic is too opaque, we should instead look for `Hydro/AdiabaticIndex` for AthenaPK and implement the adiabatic index as needed for the other codes.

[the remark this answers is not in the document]
forrestglines: As it's supposed to work right now, an adiabatic index passed in via `gamma=` will take priority over an adiabatic index defined in the parameters of the Parthenon simulation, then `5./3.` if nothing else is found.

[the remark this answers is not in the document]
forrestglines: The `He_mass_fraction` piece isn't working since at some point we stopped outputing `He_mass_fraction` and started including `mbar_over_kb` instead. We could back out `He_mass_fraction` from `mbar_over_kb` instead.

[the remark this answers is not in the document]
forrestglines: I'm trying to fix reading in `He_mass_fraction` from `mbar_over_kb` but I'm getting nonsensical results. I'll have to look at it later with fresh eyes. I'll push what I'm working with in case somone can see what I'm doing wrong.

yt/frontends/parthenon/data_structures.py
```diff
@@ -9,15 +9,15 @@
 from yt.data_objects.static_output import Dataset
 from yt.fields.magnetic_field import get_magnetic_normalization
 from yt.funcs import mylog
+from yt.geometry.geometry_enum import Geometry
```
neutrinoceros, line 12: Following how it's done in the rest of the code base
```suggestion
from yt.geometry.api import Geometry
```

yt/frontends/parthenon/data_structures.py
```diff
@@ -213,6 +199,9 @@ def _set_code_unit_attributes(self):
         self.velocity_unit = self.length_unit / self.time_unit
 
     def _parse_parameter_file(self):
+        # Read Parthenon Params into Params object
+        self.params = dict(self._handle["Params"].attrs)
```
neutrinoceros, line 203: Why not use the already defined `self.parameters` attribute instead ? `self.parameter` can be a flat dict:
```suggestion
        self.parameter.update(dict(self._handle["Params"].attrs))
```
But it doesn't have to be ! You can also use a nested layout if you like it better:
```suggestion
        self.parameters["Params"] = dict(self._handle["Params"].attrs)
```

In any case, additional attributes should be avoided. That's what `self.parameters` is for

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,321 @@
...
+        if self.ds.dimensionality < 3:
+            self.dds[2] = 1.0
+        self.field_data["dx"], self.field_data["dy"], self.field_data["dz"] = self.dds
+
+    def retrieve_ghost_zones(self, n_zones, fields, all_levels=False, smoothed=False):
+        if smoothed:
+            warnings.warn(
+                "ghost-zones interpolation/smoothing is not "
+                "currently supported for Parthenon data.",
+                category=RuntimeWarning,
```
neutrinoceros, line 55: `ruff` was recently updated and now points to this warning not having an explicit `stack_level` arg. In general is should be
```suggestion
                category=RuntimeWarning,
                stack_level=2,
```

pgrete:
In the process of adding data files of two other codes I'm currently wondering
a) if (and how) yt might handle data from general relativistic simulations, i.e., data that might carry a metric around
b) if (and how) multi material simulation data is supported (e.g., different equations of state for different fields)

yt/frontends/parthenon/fields.py
```diff
@@ -0,0 +1,225 @@
...
+    def setup_fluid_fields(self):
+        from yt.fields.magnetic_field import setup_magnetic_field_aliases
+
+        unit_system = self.ds.unit_system
+        # Add velocity fields (if only momemtum densities are given)
+        for i, comp in enumerate(self.ds.coordinates.axis_order):
+            # Support both current and legacy scheme
+            for mom_field_name in ["MomentumDensity", "cons_momentum_density_"]:
+                mom_field = ("parthenon", f"{mom_field_name}{i+1}")
+                if mom_field in self.field_list:
```
pgrete, line 87: I updated quite a bit of the field logic (moving things to `known_other_fields` above) and am now wondering if the registered/resolved fields are potentially already available here.
This would simplify logic here (and at several places below) wrt to just checking if `("gas", "momentum_density_x")` is known rather than lookups for our own names in `self.field_list`.

forrestglines:
Just pushed some quick changes so that coordinate ordering for cylindrical and spherical coordinates follows Athena++'s conventions, which AthenaPK will inherit.

(This is the last functional addition we had planned for this PR)

yt/frontends/parthenon/data_structures.py
```diff
@@ -17,7 +17,9 @@
 from .fields import ParthenonFieldInfo
 
 geom_map = {
-    "UniformCartesian": Geometry.CARTESIAN,
+    "UniformCartesian": (Geometry.CARTESIAN,("x","y","z")),
+    "UniformCylindrical": (Geometry.CYLINDRICAL,("r","theta","z")),
```
neutrinoceros, line 21: @forrestglines I suspect this line is the only reason for your latest commit, so I should point out that you should just use `Geometry.POLAR` here, so you don't need to specify axis_order at all.

forrestglines: Thanks @neutrinoceros that would certainly allow a bit of code cleanup. Unfortunately changing to `Geometry.POLAR` is breaking my scripts which use fixed resolution buffers. For example, this code to make a slice plot works with both `Geometry.POLAR` and the current solution with `Geometry.CYLINDRICAL`:
```python
ds = yt.load(filename)

slc = yt.SlicePlot(ds, "z", ("gas", "density"))
```
but this code below only works with my current solution with `Geometry.CYLINDRICAL`:
```python
ds = yt.load(filename)

slc = yt.SlicePlot(ds, "z", ("gas", "density"))
frb = slc.data_source.to_frb(width=(4,"cm"),resolution=512,center=(0,0))
frb["r"] 
```
with an error being thrown by `pixelize_cylinder`
```
File ~/installs-chicoma/miniconda3/envs/py39/lib/python3.9/site-packages/yt/visualization/fixed_resolution.py:600, in CylindricalFixedResolutionBuffer._generate_image_and_mask(self, item)
    597 @override
    598 def _generate_image_and_mask(self, item) -> None:
    599     buff = np.zeros(self.buff_size, dtype="f8")
--> 600     mask = pixelize_cylinder(
    601         buff,
    602         self.data_source["r"],
    603         self.data_source["dr"],
    604         self.data_source["theta"],
    605         self.data_source["dtheta"],
    606         self.data_source[item].astype("float64"),
    607         self.radius,
    608         return_mask=True,
    609     )
    610     self.data[item] = ImageArray(
    611         buff, units=self.data_source[item].units, info=self._get_info(item)
    612     )
    613     self.mask[item] = mask

File ~/installs-chicoma/miniconda3/envs/py39/lib/python3.9/site-packages/yt/utilities/lib/pixelization_routines.pyx:552, in yt.utilities.lib.pixelization_routines.pixelize_cylinder()

ValueError: Buffer has wrong number of dimensions (expected 2, got 1)
```

Suggestions? Is this something to be fixed in the frontend or in yt? I can provide the data file for this cylindrical dataset. The dataset will be included with our tests

neutrinoceros: Hard to tell for sure without the data file, but I can have a look if you can share it.

chrishavlin:
just FYI, i hit the button for the tests to run, but you'll likely get unrelated failures after the merge with main (e.g., #4785 ).

yt/frontends/parthenon/data_structures.py
forrestglines: Thanks! I've pushed the small changes to switch from `Geometry.CYLINDRICAL` to `Geometry.POLAR`.

Here's the file I was testing on (with an .txt extension due to github's limitations)

[disk_cyl.prim.00000.phdf.txt](https://github.com/yt-project/yt/files/14058885/disk_cyl.prim.00000.phdf.txt)

neutrinoceros: thank you ! I can reproduce locally, which is a great start. I'll try to debug this today.

neutrinoceros: Actually it's a yt bug, I can also reproduce it with AMRVAC data
```python
import yt

ds = yt.load_sample("bw_polar_2d")
slc = yt.SlicePlot(ds, "z", ("gas", "density"))
frb = slc.data_source.to_frb(width=(4, "cm"), resolution=512, center=(0, 0))
frb["r"]
```
I'll open a dedicated issue.

forrestglines: I saw issue #4790, the bug is not related to 2D datasets. I also explored this with a 3D dataset and got the same error.

[disk_cyl.3d.prim.00000.phdf.txt](https://github.com/yt-project/yt/files/14073840/disk_cyl.3d.prim.00000.phdf.txt)

forrestglines: Fantastic that we have a bug fix started with this so soon. While this did eliminate the error with FRB, the polar coordinates are still not behaving like the cylindrical coordinates. In some comparison plots between Athena++ and AthenaPK using FRB's of a slice generated with the same code, I get plots looking like this:

Screen Shot 2024-01-27 at 3 13 50 PM

My Parthenon frontend (using `Geometry.POLAR` with your new bugfix) is only getting one quadrant while the Athena++ frontend (using `Geometry.CYLINDRICAL`) is getting the full disk as expected.

Let me work on a better reproducer and perhaps one using an existing frontend in YT. I'm not convinced either way that this is now the Parthenon frontend's bug or Polar's bug but it's still a problem that changing from cylindrical to polar changes my plots.

forrestglines:
I've pushed @neutrinoceros 's bug fix to this PR although I'm still working on Polar. You can advise on how you'd like me to fix the git history later on if the merges get messy.

forrestglines:
One aspect of `Geometry.SPHERICAL` I'd like to verify -- the ordering is $(r,\theta,\phi)$, right? Where $\theta$ is the polar angle $[0,\pi]$ and $\phi$ is the azimuthal angle $[0,2 \pi]$?

This would be consistent with Parthenon's ordering.

neutrinoceros:
> I've pushed @neutrinoceros 's bug fix to this PR although I'm still working on Polar. You can advise on how you'd like me to fix the git history later on if the merges get messy.

I think we'll need to rebase this branch at least partially to mitigate merge confusion in the future. I would advise not to merge from main until then, or it might become very hard to just manage this branch.

 > One aspect of Geometry.SPHERICAL I'd like to verify -- the ordering is , right? Where  is the polar angle  and  is the azimuthal angle ?
 
Yes. You can check that by grepping for `_default_axis_order` and see that `SphericalCoordinateHandler` has `_default_axis_order = ("r", "theta", "phi")`. Also note that these represent *default* orders, but it's possible to specify different ones where needed, and still use the builtin `CoordinateHandler` classes !

pgrete:
I'm back a little bit late to the show so I'm wondering why there may be merge conflicts?
I regularly updated this branch with main and things have been working fine so far.

matthewturk:
Hm, I don't see any merge conflicts

forrestglines:
> I'm back a little bit late to the show so I'm wondering why there may be merge conflicts? I regularly updated this branch with main and things have been working fine so far.

I merged in #4790 to try to fix `Geometry.POLAR` but that PR is still not working. I'm considering undoing that merge and just using `Geometry.CYLINDRICAL` like the Athena++ frontend does.

Otherwise, the git history is messy. I was planning on nuking the git history if things got too messy since not too many of us are working on this.

matthewturk:
> Otherwise, the git history is messy. I was planning on nuking the git history if things got too messy since not too many of us are working on this.

I'm personally not a huge stickler for neat/tidy git history, but I know it is something others prioritize.  I think when this is merged, we can also squash it, or if you'd like you can edit the history into manageable and discrete chunks before that.

forrestglines:
We have these two files for code verification so far

`advection_2d.out0.final.phdf`

http://use.yt/upload/8f07e626

`athenapk.prim.disk.phdf`

http://use.yt/upload/c534769e

forrestglines:
> I'm personally not a huge stickler for neat/tidy git history, but I know it is something others prioritize. I think when this is merged, we can also squash it, or if you'd like you can edit the history into manageable and discrete chunks before that.

I'll clean up the git history or at least remove the commits from #4790 that are not used. 



> With respect to this PR, I plan to make the following two additions:
> 
>     * a 3D example (of the 2D type above) but with dx != dy != dz
> 
>     * a 3D AthenaPK example (incl changes to work with a code hint) with support for units
> 

@pgrete do we want to add these tests? The tests I just added I think would address everything that is specific to downstream code datasets:

1. Loading in an AthenaPK dataset with the Parthenon frontend
2. Loading units from the `phdf` file
3. Creating derived fields (like temperature) if the right fields are in the `phdf` file
4. Loading in a cylindrical dataset

I could see added more testing for magnetic fields and datasets with dx != dy != dz. Beyond that, I think this is ready for review again

forrestglines:
I added more tests, this PR is finished except for review and requested changes.

BenWibking:
I am getting an error when using the latest version of this branch to load an AthenaPK dataset:
```

File ~/yt_parthenon/yt/loaders.py:99, in load(fn, hint, *args, **kwargs)
     58 """
     59 Load a Dataset or DatasetSeries object.
     60 The data format is automatically discovered, and the exact return type is the
   (...)
     95     If the data format matches more than one class of similar specialization levels.
     96 """
     97 from importlib.metadata import entry_points
---> 99 from yt.frontends import _all  # type: ignore [attr-defined] # noqa
    101 _input_fn = fn
    102 fn = os.path.expanduser(fn)

File ~/yt_parthenon/yt/frontends/__init__.py:57, in __getattr__(value)
     54 if value == "_all":
     55     # recursively import all frontends
     56     for _ in __all__:
---> 57         __getattr__(_)
     58     return
     60 if value not in __all__:

File ~/yt_parthenon/yt/frontends/__init__.py:63, in __getattr__(value)
     60 if value not in __all__:
     61     raise AttributeError(f"yt.frontends has no attribute {value!r}")
---> 63 return importlib.import_module(f"yt.frontends.{value}.api")

File ~/.conda/envs/yt-analysis/lib/python3.11/importlib/__init__.py:126, in import_module(name, package)
    124             break
    125         level += 1
--> 126 return _bootstrap._gcd_import(name[level:], package, level)

File ~/yt_parthenon/yt/frontends/ramses/api.py:2
      1 from . import tests
----> 2 from .data_structures import RAMSESDataset
      3 from .definitions import field_aliases
      4 from .fields import RAMSESFieldInfo

File ~/yt_parthenon/yt/frontends/ramses/data_structures.py:35
     33 from .field_handlers import get_field_handlers
     34 from .fields import _X, RAMSESFieldInfo
---> 35 from .hilbert import get_intersecting_cpus
     36 from .io_utils import fill_hydro, read_amr
     37 from .particle_handlers import get_particle_handlers

File ~/yt_parthenon/yt/frontends/ramses/hilbert.py:6
      3 import numpy as np
      5 from yt.data_objects.selection_objects.region import YTRegion
----> 6 from yt.geometry.selection_routines import (
      7     bbox_intersects,
      8     fully_contains,
      9 )
     11 # State diagram to compute the hilbert curve
     12 _STATE_DIAGRAM = np.array(
     13     [
     14         [
   (...)
     46     ]
     47 )

ImportError: cannot import name 'bbox_intersects' from 'yt.geometry.selection_routines' (/u/bwibking/yt_parthenon/yt/geometry/selection_routines.cpython-311-x86_64-linux-gnu.so)
```

matthewturk:
@BenWibking That's usually what shows up if we add a new function to the Cython code and the library isn't rebuilt.  I'm not sure how you install, but for me it's usually a matter of `python setup.py build_ext -i` in the yt source directory.

BenWibking:
> @BenWibking That's usually what shows up if we add a new function to the Cython code and the library isn't rebuilt. I'm not sure how you install, but for me it's usually a matter of `python setup.py build_ext -i` in the yt source directory.

I tried `pip install -e .` and got a new error:
```
File ~/.conda/envs/yt-analysis/lib/python3.11/importlib/__init__.py:126, in import_module(name, package)
    124             break
    125         level += 1
--> 126 return _bootstrap._gcd_import(name[level:], package, level)

File ~/yt_parthenon/yt/frontends/ramses/api.py:2
      1 from . import tests
----> 2 from .data_structures import RAMSESDataset
      3 from .definitions import field_aliases
      4 from .fields import RAMSESFieldInfo

File ~/yt_parthenon/yt/frontends/ramses/data_structures.py:33
     23 from yt.utilities.physical_ratios import cm_per_mpc
     25 from .definitions import (
     26     OUTPUT_DIR_EXP,
     27     OUTPUT_DIR_RE,
   (...)
     31     ramses_header,
     32 )
---> 33 from .field_handlers import get_field_handlers
     34 from .fields import _X, RAMSESFieldInfo
     35 from .hilbert import get_intersecting_cpus

File ~/yt_parthenon/yt/frontends/ramses/field_handlers.py:12
      9 from yt.utilities.cython_fortran_utils import FortranFile
     11 from .io import _read_fluid_file_descriptor
---> 12 from .io_utils import read_offset
     14 FIELD_HANDLERS: set[type["FieldFileHandler"]] = set()
     17 def get_field_handlers():

ModuleNotFoundError: No module named 'yt.frontends.ramses.io_utils'
```

matthewturk:
That's a weird one!  Is this during the pip install, or after?

BenWibking:
> That's a weird one! Is this during the pip install, or after?

This is after, when doing `yt.load(file)`.

matthewturk:
I'm afraid I'm not sure where this would come from unless the io utils aren't being built ... one way to get around this to test this specific frontend would be just ot comment out that line, and then not load RAMSES data.

pgrete:
What's the status here?

@forrestglines did you remove the fix for non-Cart coordinate systems that didn't quite work as expected (as I see some code left that may go in that direction)

@matthewturk @neutrinoceros what are the next steps with respect to re-review?

neutrinoceros:
I'm assuming I just need to wait until the branch history is cleaned up before I should review again ? I haven't been able to follow through on core fixes for cylindrical and polar geometries lately, is that blocking this PR in any way ?

BenWibking:
Could we squash the PR on merge instead of manually cleaning up the history?

matthewturk:
I'm OK with squashing on merge, but I also am not that opposed to merging with messy history since it's all largely in the same directory.

neutrinoceros:
merging with a messy history isn't what I'm concerned about; I would like the history cleaned up because I'd like to keep reviewing as simple as possible, and conflicts as unlikely as possible.

matthewturk:
@pgrete thank you for jumping through these hoops.  I think this is a fine set of code to review!  I'll take a look at it as well, but I wanted to say thank you for your work.

neutrinoceros:
> the only simplification for a review I can see right now is to force-push a squashing rebase.

That would be what I was expecting. If this branch has users actively pulling in (other than its authors), I suggest backing it up before you force push.

About data samples: sorry this got lost in the conversation. What needs to happen is:
1) a PR should be open to https://github.com/yt-project/website
2) they also need to be registered in the main repo in `yt/sample_data_registry.json`
3) finally, we need to make sure that `yt.load_sample` works correctly for them

neutrinoceros:
> Where do the tarballs actually go?

They'll be installed on the server by someone with admin rights (not me).

> What's the hash I need to add to the registry? Some flavor of checksum?

See [pooch docs](https://www.fatiando.org/pooch/latest/hashes.html#calculating-hashes) for how to produce hashes.

> I assume this won't work until steps 1 and 2 are done?

Correct !

neutrinoceros:
I'll try to review thoroughly again soon (probably not before the weekend).

forrestglines:
I've pushed the sha256 hashes of the tarballs. `yt.load_sample` is working as I would expect it to
```
ds = yt.load_sample("athenapk_cluster")
slc = yt.SlicePlot(ds,"z",fields=["density",])
slc.show()

ds = yt.load_sample("athenapk_disk")
slc = yt.SlicePlot(ds,"z",fields=["density",])
slc.show()

ds = yt.load_sample("parthenon_advection")
slc = yt.SlicePlot(ds,"z",fields=["advected_0_0",])
slc.show()
```

forrestglines:
Needed a small fix to the name of a dataset that I changed as I created the tarball. My nose test passes locally now
```bash
nosetests -v --with-answer-testing --local --local-dir $HOME/trash/yt-test-data --answer-store --answer-name=local-parthenon yt.frontends.parthenon
# Runs through 63 tests quickly
```

Xarthisius:
Please add the following change to trigger answer testing on yt infra:
```diff
diff --git a/tests/tests.yaml b/tests/tests.yaml
index 25e310d87..a87af7b7b 100644
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -164,6 +164,11 @@ answer_tests:
   local_nc4_cm1_002: # PR  2176, 2998
     - yt/frontends/nc4_cm1/tests/test_outputs.py:test_cm1_mesh_fields
 
+  local_parthenon_001: # PR 4323
+    - yt/frontends/parthenon/tests/test_outputs.py:test_loading_data
+    - yt/frontends/parthenon/tests/test_outputs.py:test_cluster
+    - yt/frontends/parthenon/tests/test_outputs.py:test_derived_fields
+
 other_tests:
   unittests:
      # keep in sync with nose_ignores.txt
```

Please note that only those 3 tests are answer test, i.e. they generate data that can be stored and compared. All the other tests in `test_outputs.py` are unit tests and should be using `@requires_file` instead of `@requires_ds` decorator.

forrestglines:
> Please note that only those 3 tests are answer tests, i.e. they generate data that can be stored and compared. All the other tests in `test_outputs.py` are unit tests and should be using `@requires_file` instead of `@requires_ds` decorator.

Done. Does this mean only these tests are run during the CI? Are the unit tests covered during normal CI for yt?

Xarthisius:
> Done. Does this mean only these tests are run during the CI? Are the unit tests covered during normal CI for yt?

Unit tests are detected automatically. Answer tests has to be specified in `tests.yaml`.

BenWibking:
"Full testsuite (nose)" fails because it says "there is no old answer available." (https://tests.yt-project.org/job/yt_py310_git/7946/).

Is there a problem here, or does it just need to be re-run?

BenWibking:
Is there a way to auto-format to satisfy the checks here: https://results.pre-commit.ci/run/github/88776493/1709230377.pe6ltUfYTG-No8lHXu3CpA?

Xarthisius:
@yt-fido test this please.

neutrinoceros:
@BenWibking yes, the magic comment is found at the bottom of the page you linked.
Though I don't expect you have clearance to trigger the bot as you're not the PR's author. I'll let @forrestglines take care of it.

There's exactly *one* lint that cannot be auto-fixed though:
```
yt/frontends/parthenon/io.py:13:15: C400 Unnecessary generator (rewrite as a `list` comprehension)
```
please fix this one manually

forrestglines:
pre-commit.ci autofix

BenWibking:
Did the bot... react with a thumbs-down emoji?

neutrinoceros:
That's new. Let me see if I can convince it with maintainer authority.

neutrinoceros:
pre-commit.ci autofix

forrestglines:
This broken doc build doesn't look like our doing, should we just merge main? (I'm more asking reviewers if a simple merge from main would be better than another rebase)

```
/var/jenkins_home/workspace/yt_docs_new/doc/source/reference/field_list.rst:26: ERROR: Unknown directive type "notebook-cell".

.. notebook-cell::

  import yt
  for i in sorted(ds.field_list):
    print(i)

Sphinx parallel build error:
nbsphinx.NotebookError: RuntimeError in cookbook/fits_xray_images.ipynb:
Kernel died before replying to kernel_info
make: *** [Makefile:51: html] Error 2
Process leaked file descriptors. See https://www.jenkins.io/redirect/troubleshooting/process-leaked-file-descriptors for more information
Build step 'Execute shell' marked build as failure
Archiving artifacts
Recording fingerprints
Recording fingerprints
Setting status of 4bc3027b7b862d76a69d45a3099962f97c1d134c to FAILURE with url https://tests.yt-project.org/job/yt_docs_new/2762/ and message: 'Build finished. '
Using context: docs
Finished: FAILURE
```

Xarthisius:
@yt-fido test this please

Xarthisius:
> This broken doc build doesn't look like our doing, should we just merge main? (I'm more asking reviewers if a simple merge from main would be better than another rebase)

Don't worry about it. It's not blocking this PR. I re-triggered docs build and it will likely pass this time.

neutrinoceros:
Not sure that's a deterministic error. Let me try again:
@yt-fido test this please

neutrinoceros:
Docs built alright but this time regular tests failed (also flaky)
@yt-fido  test this please

neutrinoceros:
All green now ! I hope I'll find time to review this week, but with the imminent release of a major numpy release candidate I may not. Thank you for your patience !

yt/frontends/parthenon/io.py
```diff
@@ -0,0 +1,107 @@
...
+            rv[field] = np.empty(size, dtype="=f8")
+        ng = sum(len(c.objs) for c in chunks)
+        mylog.debug(
+            "Reading %s cells of %s fields in %s blocks",
+            size,
+            [f2 for f1, f2 in fields],
+            ng,
+        )
+        last_dname = None
+        for field in fields:
```
chrishavlin, line 52: So most (all other?) frontends loop over chunks then fields, e.g., 

``` python
for chunk in chunks:
     # do stuff
     for field in fields:
            # get this fields data for this chunk
            # apply selector
```
The main reason being that yt's parallel chunking works by sending chunks to different processors where field data is selected. So if you eventually want the Parthenon frontend to work with yt's parallel operations, you'll want to invert the loop order here. I don't think this needs to be done for this PR, but a refactor in the future might be warranted...

chrishavlin (commented):
Generally looks good to me! Left a number of questions and comments, but nothing too critical.

yt/frontends/parthenon/io.py
BenWibking: I have tried to use yt's parallel functionality before, and have not been able to get it to work, so this might explain why. It would be ideal if this could work. Is inverting the loop order the only change that's needed to get it working?

chrishavlin: So I think I should have said "to work more efficiently with yt's parallel operations" -- generally it's better to process all the fields within a chunk rather than process every chunk for each field. But after thinking more, it might actually not be too important here since it's reading from a single hdf file -- when chunks are spread across files if you put the chunk loop on the inside you'd be incurring extra cost from opening/closing the same file multiple times. 

In terms of getting parallel operations to work -- I don't **think** there should be more changes. When you run a parallel-enabled function with MPI, I believe the `chunks` argument should come in to this function with only the chunks associated with the MPI process running the function. I can test that out though, I've been meaning to get back into some parallel indexing related work....

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,326 @@
...
+            # We set these to cgs for now, but they may have been overridden
+            if getattr(self, unit + "_unit", None) is not None:
+                continue
+            elif unit_param in self.parameters:
+                setattr(
+                    self, f"{unit}_unit", self.quan(self.parameters[unit_param], cgs)
+                )
+            else:
+                mylog.warning(f"Assuming 1.0 = 1.0 {cgs}")
+                setattr(self, f"{unit}_unit", self.quan(1.0, cgs))
```
neutrinoceros, line 205: This looks a lot like it could be simplified with `yt.funcs.setdefaultattr` (we'd loose the log warning, but I don't think that's essential)

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,326 @@
...
+        )
+        self.magnetic_unit.convert_to_units("gauss")
+        self.velocity_unit = self.length_unit / self.time_unit
+
+    def _parse_parameter_file(self):
+        self.parameters.update(self.specified_parameters)
+        for key, val in self._handle["Params"].attrs.items():
+            if key in self.parameters.keys():
+                mylog.warning(
+                    f"Overriding existing 'f{key}' key in ds.parameters from data 'Params'"
```
neutrinoceros, line 220: I think the `f` is a typo ? 
```suggestion
                    f"Overriding existing key {key!r} in ds.parameters from data 'Params'"
```

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,326 @@
...
+
+        if "OutputFormatVersion" in self._handle["Info"].attrs.keys():
+            self.output_format_version = self._handle["Info"].attrs[
+                "OutputFormatVersion"
+            ]
+        else:
+            self.output_format_version = -1
+
+        # Use duck typing to check if num_components is a single int or a list
+        # h5py will return different types if there is a single variable or if there are more
```
neutrinoceros, line 248: sounds like we could just use `isinstance(num_components, list)` and drop the try block as well as the comment. It would also improve type safety internally. Or maybe I'm missing something ?

yt/frontends/parthenon/fields.py
```diff
@@ -0,0 +1,165 @@
...
+                    self.add_field(
+                        ("gas", f"velocity_{comp}"),
+                        sampling_type="cell",
+                        function=velocity_field(mom_field),
+                        units=unit_system["velocity"],
+                    )
+        # Figure out thermal energy field
+        if ("parthenon", "Pressure") in self.field_list or (
+            "parthenon",
+            "prim_pressure",
```
neutrinoceros, line 97: I think formatting would end up less terrible here if we removed this magic comma
```suggestion
            "prim_pressure"
```

yt/frontends/parthenon/fields.py
```diff
@@ -0,0 +1,165 @@
...
+        ) in self.field_list:
+
+            def _specific_thermal_energy(field, data):
+                # TODO This only accounts for ideal gases with adiabatic indices
```
neutrinoceros, line 101: how would we detect if this is appropriate in practice ? It's okay to not support the general case from the get go, but we should at least be able to not add physically inconsistent fields.

yt/frontends/parthenon/fields.py
```diff
@@ -0,0 +1,165 @@
...
+
+            self.add_field(
+                ("gas", "specific_thermal_energy"),
+                sampling_type="cell",
+                function=_specific_thermal_energy,
+                units=unit_system["specific_energy"],
+            )
+        elif ("parthenon", "TotalEnergyDensity") in self.field_list or (
+            "parthenon",
+            "cons_total_energy_density",
```
neutrinoceros, line 116: same as above (formatting)
```suggestion
            "cons_total_energy_density"
```

yt/frontends/parthenon/io.py
```diff
@@ -0,0 +1,107 @@
...
+    _dataset_type = "parthenon"
+
+    def __init__(self, ds):
+        super().__init__(ds)
+        self._handle = ds._handle
+
+    def _read_particles(
+        self, fields_to_read, type, args, grid_list, count_list, conv_factors
+    ):
+        pass
```
neutrinoceros, line 33: this method is not required, so I'd rather not commit noop code
```suggestion
```

yt/frontends/parthenon/io.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def _read_fluid_selection(self, chunks, selector, fields, size):
+        chunks = list(chunks)
+        if any((ftype != "parthenon" for ftype, fname in fields)):
+            raise NotImplementedError
```
neutrinoceros, line 38: I'm not sure this is reachable, and I find it a bit confusing
```suggestion
```

yt/frontends/parthenon/io.py
```diff
@@ -0,0 +1,107 @@
...
+                                ii[n] * nx : (ii[n] + 1) * nx,
+                                jj[n] * ny : (jj[n] + 1) * ny,
+                                kk[n] * nz : (kk[n] + 1) * nz,
+                            ] = ds[id, fdi, :, :, :].transpose()
+                        ind += mesh.select(selector, data, rv[field], ind)  # caches
+                else:
+                    for gs in grid_sequences(chunk.objs):
+                        start = gs[0].id - gs[0]._id_offset
+                        end = gs[-1].id - gs[-1]._id_offset + 1
+                        # Deprecate this fallback on next major Parthenon release
```
neutrinoceros, line 74: Can you clarify what "next" means in this comment ? What's the reference version ?
Or maybe we don't need to deprecate the fallback at all actually, it seems like a reasonable amount of legacy code to carry around, but it'd still be nice if the comment indicated which versions of Parthenon it is for.

yt/frontends/parthenon/io.py
```diff
@@ -0,0 +1,107 @@
...
+        if len(fields) == 0:
+            return rv
+        for field in fields:
+            ftype, fname = field
+            dname, fdi = self.ds._field_map[fname]
+            ds = f[f"/{dname}"]
+            for gs in grid_sequences(chunk.objs):
+                start = gs[0].id - gs[0]._id_offset
+                end = gs[-1].id - gs[-1]._id_offset + 1
+                # Deprecate this fallback on next major Parthenon release
```
neutrinoceros, line 100: same as above.

neutrinoceros (commented):
Sorry for taking so long to get my review in. I note that there's also another, mostly unaddressed (from what I can see) review from @chrishavlin. Both need to be addressed before we can merge this.
Thank you for your patience !

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,326 @@
...
+            self.logarithmic = True
+            raise ValueError(
+                "Logarithmic grids not yet supported in Parthenon frontend."
+            )
+        else:
+            self._index_class = ParthenonHierarchy
+            self.logarithmic = False
+        self._magnetic_factor = get_magnetic_normalization(magnetic_normalization)
+
+        self.geometry = geom_map[self._handle["Info"].attrs["Coordinates"]]
```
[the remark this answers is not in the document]
pgrete: For AthenaPK spherical and cylindrical should also work (though they're not upstreamed yet).

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,326 @@
...
+        for key, val in self._handle["Params"].attrs.items():
+            if key in self.parameters.keys():
+                mylog.warning(
+                    f"Overriding existing 'f{key}' key in ds.parameters from data 'Params'"
+                )
+            self.parameters[key] = val
+
+        xmin, xmax = self._handle["Info"].attrs["RootGridDomain"][0:2]
+        ymin, ymax = self._handle["Info"].attrs["RootGridDomain"][3:5]
+        zmin, zmax = self._handle["Info"].attrs["RootGridDomain"][6:8]
```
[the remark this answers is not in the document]
pgrete: Correct, y/z min/max are always defined even for 1D sims. Dimensionality is determined by the root grid cell count in second and third dim (set to `1`, which is why we check again `== 1` below.

yt/frontends/parthenon/data_structures.py
```diff
@@ -0,0 +1,326 @@
...
+        if self.domain_dimensions[1] == 1:
+            dimensionality = 1
+        self.dimensionality = dimensionality
+        self.current_time = self._handle["Info"].attrs["Time"]
+        self.num_ghost_zones = 0
+        self.field_ordering = "fortran"
+        self.boundary_conditions = [1] * 6
+        self.cosmological_simulation = False
+
+        if "periodicity" in self.parameters:
```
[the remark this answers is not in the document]
pgrete: I think it's more appropriate to say that only periodic boundary conditions are properly handled. Data with non-periodic boundary conditions can be read but results at the boundaries might be off.
I'll update the doc accordingly.

yt/frontends/parthenon/io.py
pgrete: In that case, I'm currently leaning towards leaving as it is and keep this in mind in case we run into issues (which we kind of are, but processing 3TB dumps is a story for another day).

yt/frontends/parthenon/io.py
```diff
@@ -0,0 +1,107 @@
...
+                        else:
+                            data = ds[start:end, fdi, :, :, :].transpose()
+                        for i, g in enumerate(gs):
+                            ind += g.select(selector, data[..., i], rv[field], ind)
+            last_dname = dname
+        return rv
+
+    def _read_chunk_data(self, chunk, fields):
+        if self.ds.logarithmic:
+            pass
```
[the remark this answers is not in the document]
pgrete: I removed all the associated log logic as no downstream code currently uses sth like it (and it's not implemented in Parthenon either)

yt/frontends/parthenon/data_structures.py
pgrete: If I understand the logic correctly, we can get rid of the first `if` but still need to differentiate between setting the explicit unit from the output file and implicit cgs (and are still able to show a warning), don't we? This is also what I did now.
Otherwise, I'm missing sth.

yt/frontends/parthenon/data_structures.py
pgrete: yes, that should do the trick.

yt/frontends/parthenon/fields.py
pgrete: My formatter makes this magic comma come back again and again.

yt/frontends/parthenon/fields.py
pgrete: We're now checking against for the `Hydro/AdiabaticIndex` key (which would make this appropriate for the AthenaPK code), and otherwise show a warning that the field is defined under the assumption of an ideal gamma-type law.
Does that work?

yt/frontends/parthenon/io.py
pgrete: This was copied over from the Athena++ frontend.
Could it have been added if people forcibly overwrite the `dataset_type`?
I don't see any reason why this would be the case and now have removed those two lines.

yt/frontends/parthenon/io.py
pgrete: This does not apply any more as the old version has been deprecated for well over a year (probably two) now.
I removed the conditional and raising an error in `data_structure.py` if the version could not be determined.

pgrete:
Alright, I think I've addressed all review comments (@neutrinoceros @chrishavlin ) pending potential replies by you to the unresolved ones.
I also updated the branch to current `main`.

From my point of view this should be ready to go (assuming all tests pass).

What are the chances this PR will make it to the 4.4 release?

jzuhone:
@pgrete I will also take a look at this tomorrow and I think that we should make this a priority for the 4.4 release.

yt/frontends/parthenon/fields.py
neutrinoceros: Ok I won't go into full bikeshedding mode here (the point of using formatters is to save us from having to discussing it), but just in case it's relevant, I'd like to point out that you should be using the formatting toolbelt that is configured by `pre-commit-config.yml`, otherwise conflicts are inevitable. Are you ?

yt/frontends/parthenon/fields.py
neutrinoceros: I think so. Now the comment should be removed

neutrinoceros:
I would normally oppose stacking more PRs (especially large ones) to a release's milestone at the last minute, but this one has been waiting for so long that it makes sense to try and get it in.

pgrete:
> I would normally oppose stacking more PRs (especially large ones) to a release's milestone at the last minute, but this one has been waiting for so long that it makes sense to try and get it in.

That'd be awesome. We've been using this branch for quite some time now without issues and given that it doesn't change anything in the yt core, I think it should not pose any potential harm to the release code base as a whole.

yt/frontends/parthenon/fields.py
neutrinoceros: alright, feel free to reslove these threads then

neutrinoceros:
Indeed. Though, if for any reason we cannot make it, you guys could also consider [packaging this frontend as an extension](https://yt-project.org/doc/developing/extensions.html#frontends-as-extensions)

chrishavlin (dismissed):
Looks great from my perspective, thanks for addressing my comments @forrestglines and @BenWibking

chrishavlin:
also +1 for making this a priority for 4.4.

pgrete:
"@pgrete dismissed @chrishavlin 's stale review via dd42468" <- I don't know where/how that came from. I just pushed a one line change removing a "Todo comment" I already addressed yesterday.

BenWibking:
> "@pgrete dismissed @chrishavlin 's stale review via [dd42468](https://github.com/yt-project/yt/commit/dd424680a5b35a01a9174baa4a0f829dced582e9)" <- I don't know where/how that came from. I just pushed a one line change removing a "Todo comment" I already addressed yesterday.

This can be caused by the branch protection settings: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-pull-request-reviews-before-merging.

chrishavlin:
@pgrete ya, no worries, that's cause of yt's repo config. I'll hit approve again, and you can consider this my blanket approval if it gets dismissed again :)

chrishavlin (approved):

jzuhone (approved):

neutrinoceros:
Alright, I *think* all my comments have been addressed, and with 2 other approvals I feel like there's no reason to hold this back any longer, so let's get it in ! Thank you @pgrete and @forrestglines for your dedication and *patience* with the process. We'll be trying to get yt 4.4 out the door very shortly, so stay tuned !

pgrete:
Thanks for pulling this in on short notice! I'm very happy to update our downstream notes to pull the main (and future yt 4.4) version rather than a custom branch.

PR #4331, opened 2023-02-09
BUG: fix the error when loading a non-cosmological `Gizmo` dataset

jisuoqing:
Fix the error `'GizmoFieldInfo' object has no attribute 'nuclei_names'` complained by Line 122
https://github.com/yt-project/yt/blob/bd6357eea0121c4071a6580bedcf543122824d35/yt/frontends/gizmo/fields.py#L122-L124
when loading a non-cosmological `Gizmo` dataset.


neutrinoceros (commented):
Nice catch ! Shouldn't `self.species_names` be assigned in the `else` branch too ?

neutrinoceros:
>  Shouldn't self.species_names be assigned in the else branch too ?

I can answer that myself: it's not needed, because this one is already set by `FieldInfoContainer.__init__`. `nuclei_names` on the other hand is a frontend-specific attribute.

neutrinoceros (approved):

PR #4332, opened 2023-02-13
DOC: load unstructured mesh SlicePlot

chrishavlin:
The SlicePlot call in the first load_unstructured_mesh generic example is missing a field type for the field argument.


neutrinoceros:
Good catch !
just to clarify, the current version of the docs is *broken*, and I confirm that this patch fixes it. Thank you !

PR #4335, opened 2023-02-13
[WIP] BUG: avoid division by 0 in unstructured mesh pixelization

### LINKED ISSUE #4334: BUG: blank pixels in unstructured mesh slices
### Bug report

For some unstructured meshes, you can get blank pixels from the pixelization routines.

**Code for reproduction**

I've seen this for a couple different `pvtu` datasets output by [ASPECT](https://aspect.geodynamics.org/) and [TerraFERMA](https://github.com/TerraFERMA/TerraFERMA) that I've been working with in [yt_aspect](https://github.com/data-exp-lab/yt_aspect/). For this report, I stripped down a 2D dataset that I've been working with to simplify this report to use vanilla yt, but I've also seen this with 3D data. The sample data is here:

```bash
curl -JO http://use.yt/upload/c1b9b92c
```
This data includes just the mesh and connectivity arrays required to use `load_unstructured_mesh`, so first load those in, add a field defined at the element nodes and then call `load_unstructured_mesh`:

```
import h5py
import yt
import numpy as np

# load the mesh
with h5py.File("unstructured_mesh_det0_example.h5", 'r') as ds_hdf:
    
    crds = ds_hdf['mesh']['coords'][:]
    cid = ds_hdf['mesh']['connectivity'][:]

# add some random data fields
node_data = {}
node_data["connect1", "T"] = np.ones(cid.shape)    

# load it in 
ds = yt.load_unstructured_mesh(cid, crds, node_data=node_data)
```

When you slice this data, you get a slice that contains some blank pixels: 

image

Decreasing the resolution, the blank pixels stand out much more:

image

**Expected outcome**

No blank pixels

chrishavlin:
Closes #4334 

The unstructured mesh pixelization (`pixelize_element_mesh`) relies on finite element samplers to convert from the coordinates of a given pixel to the local coordinates relative to each element in the mesh. The code used for that conversion is defined in the `map_real_to_unit` function for the various element types defined in `yt/utilities/lib/element_mappings.pyx`. For some element types, a newton iteration is used to calculate that conversion, and in those cases it is possible to divide by 0, which prevents the calculation from converging.

~~I haven't worked out **why** exactly the determinant ends up 0 during the iterations, but simply padding the determinant by a small nonzero epsilon prevents propagation of the `inf` and allows the calculation to continue and converge, resulting in filled rather than blank pixels.~~

See this comment for the underlying cause https://github.com/yt-project/yt/pull/4335#issuecomment-1435053171 

Here's the plot output when running the code snippet in #4334 with this branch:

image


chrishavlin:
WIP cause I'd like to look more closely at the other element samplers to see if there are 0-division concerns with them too...

neutrinoceros:
The number of failed image tests on GH suggests that it's not the usual flakiness we've been seeing (we have 9 failures here instead of 7 with flaky runs), so unfortunately I think we should consider them as real, even though the diff is completely indiscernible by human eyes. You will need to update the answer store.

chrishavlin:
Yup! That was my thought too. They look like real but acceptable failures... but I'll look a bit more closely before bumping the tests.

Xarthisius:
@yt-fido test this please

Xarthisius:
@yt-fido test this please

chrishavlin:
Ok, so I'm pretty confident that it's OK to simply bump the answers. I spent a while digging in.... 

[This repository](https://github.com/chrishavlin/unstructured_mesh_manual_compare) contains some manual answer testing in which I generated figures using code modified from the actual answer tests. I did this so I could save off the underlying image array values and compare the raw values of the fixed resolution buffer (in the case of the mesh slice tests) and the binned values from the profile plots (in the case of the failing profile tests). 

Conclusions:

1. I could not reproduce the PhasePlot failures locally running the standard image comparison with nose. I also could not find any difference in the underlying binned field values. So hard to say why those tests failed.... 

2. For the slices through the unstructured mesh dataset, I also couldn't get the nose comparisons to fail locally. The FRB values from the mesh slices do show some very small (order ~5e-14 or smaller) differences between main and this branch, which is not surprising. Some of the patterns are kinda pretty, go look at [this notebook image gallery](https://nbviewer.org/github/chrishavlin/unstructured_mesh_manual_compare/blob/main/loading_the_mesh_slices.ipynb). In the single-element case (`small_fake_ds_*`), it's clear that the differences are coming from close to the cell nodes (e.g., [this case](https://raw.githubusercontent.com/chrishavlin/unstructured_mesh_manual_compare/main/comparison_figs/small_fake_hexahedral_ds_connect1_test_0_0.png)), which makes sense: as the determinant value in the newton iteration approaches 0, the small buffer that I added to avoid an identical value of 0 will start to influence the result. **All that said,** these differences are tiny, and I don't actually think are what is causing the standard answer tests to fail (since once you convert to RGB space and save the full yt image, you don't see any of these tiny differences). 

What next:

* I'll hold off bumping answers until after some initial reviews
* I'll remove the WIP after I add a new test for the fix here

neutrinoceros:
> I could not reproduce the PhasePlot failures locally running the standard image comparison with nose. I also could not find any difference in the underlying binned field values. So hard to say why those tests failed....

I think those are just our usual flaky tests. I don't think it's worth investigating why they break and how to fix them [and focus our energy on migrating to pytest-mpl instead](https://github.com/yt-project/yt/pull/4122)

I also have another (maybe important) comment to make here that I overlooked earlier: doesn't your patch make plotting a perfectly uniform 0 field impossible ?
I know it's somewhat of an edge case, but I find it's actually quite common in practice when one visualises initial conditions or creates an animation with initial conditions as the first frame. There's already an issue with plotting uniform fields in that yt's antialiasing (which helpfully, can be switched off !) creates noise that will inevitably dominate (define) the data range.
Your approach is still good in the vast majority of cases, I think, but I also believe there should be a use-exposed API to disable it.

chrishavlin:
> doesn't your patch make plotting a perfectly uniform 0 field impossible ? 

Hmm, I'll definitely double check (and add a simple test if there's not one), but I wouldn't think so. The patch here should only affect the calculation of coordinates, not the evaluation at the coordinates. The element sampling looks like:

```python
    cdef double sample_at_real_point(self,
                                     double* vertices,
                                     double* field_values,
                                     double* physical_x) nogil:
        cdef double val
        cdef double mapped_coord[4]

        self.map_real_to_unit(mapped_coord, vertices, physical_x)
        val = self.sample_at_unit_point(mapped_coord, field_values)
        return val
```

I'm only modifying `map_real_to_unit`, which takes you from physical coordinate space to the local element coordinate space. The subsequent `sample_at_unit_point` is unchanged, so uniform fields should still be uniform, even if 0.

yt/utilities/lib/element_mappings.pyx
```diff
@@ -256,7 +256,7 @@ cdef class P1Sampler3D(ElementSampler):
             col1[i] = vertices[3 + i]     - vertices[9 + i]
             col2[i] = vertices[6 + i]     - vertices[9 + i]
 
-        d = determinant_3x3(col0, col1, col2)
+        d = determinant_3x3(col0, col1, col2) + 1e-16
```
matthewturk, line 259: In general I'm not opposed to having a specific offset specified, *but*, it may help if we either used a constant provided by numpy or math.h *or* if we were to use a `DEF` (if those are still allowed) to put it at the top-level.

matthewturk (approved):

yt/utilities/lib/element_mappings.pyx
chrishavlin: that makes sense to me!

chrishavlin: `misc_utilities.pyx` uses

```python
     cdef np.float64_t eps = np.finfo(np.float64).eps
```

in a couple spots, so I can try that here too. If I follow that example, I'd add it to the individual functions though -- do you think it's worth putting it at a higher level than function level? Might be able to add it as a attribute of the base `ElementSampler` class. Or maybe as a standalone variable in the `element_mappings.pxd` ?

chrishavlin: oh actually, I can add a 

```python
from libc.float cimport DBL_MIN
``` 
at the top and use that `DBL_MIN` throughout.

neutrinoceros: I think we can consider ˋDEF` still allowed. I'll handle the migration to whatever ends up replacing it in Cython 3 !

neutrinoceros:
Seems like #4337 is now crashing normal tests too.

neutrinoceros:
is this still WIP ?

chrishavlin:
ah hah!! In trying to write a reduced test, I figured out the underlying reason why the determinate in some cases ends up 0 in the first place.

The data where I'm seeing the blank pixels actually includes some invalid elements where the connectivity all point to the same coordinates! Here's a toy example that reproduces the behavior on main:

```python
import yt 
import numpy as np
_coordinates = np.array([[0.        , 1.        ],
                         [0.22746729, 0.        ],
                         [1.        , 0.87744062],
                         [0.11373325, 0.5       ],
                         [0.61373325, 0.43872031],
                         [0.5       , 0.93872031],
                         [0.4       , 0.6],
                        ])

_connectivity = np.array([[1, 2, 3, 4, 5, 6],
                          [7, 7, 7, 7, 7, 7]])  # add a bad element pointing to 

# the distance from the origin
node_data = {}
dist = np.sum(_coordinates**2, 1)
node_data[("connect1", "test")] = dist[_connectivity - 1]

ds = yt.load_unstructured_mesh(_connectivity - 1, _coordinates, node_data=node_data)
yt.SlicePlot(ds, "z", ("connect1", "test"), buff_size=(20,20), origin='native')
```

Screenshot from 2023-02-17 11-52-36

without the null element that central patch of nan's is not there. 

So ultimately the root cause is an upstream bug in the code where the pvtu files that I'm working with are being written. I can reach out upstream, but it'd be nice to be able to load and plot results from existing simulation data without the blank pixels showing up. 

So what to do:
* I could add some steps to check for and remove these null elements before loading into yt (so close this PR and the related issue as outside the yt scope)
* just keep the fix from this PR

I have a slight preference to keep the fix from this PR as I think for larger datasets, having the yt cython functions guard against this case will be more performant that having to check every element beforehand. But I also don't feel strongly about it, so happy to go with the first option.

@matthewturk  and @neutrinoceros , any thoughts?

chrishavlin:
pushed a test so that the PR is ready depending on which way the discussion related to my previous comment goes.

neutrinoceros:
Ok so disclaimer: I don't understand how unstructured meshed work _at all_, so I'll answer purely from a UX standpoint.

I'm okay with yt patching the data such that even slightly corrupted data may be visualised, but I'm having mixed feelings about it being done unconditionally and (more importantly) silently.
I assume it'd be good to know that your data is slightly malformed, even if yt is able to recover from it, but I'm very uncertain of the complexity cost of refining this.

That said, no strong opinions, so I'd be okay with this patch going in as is.

chrishavlin:
Good points! 

One way to make this fix more focused would be for me to remove the constant offset and instead add a `is_valid()` method to the base `ElementSampler`. Then I could modify the `pixelize_element_mesh` method to skip invalid elements -- I could also track the number of skipped elements then print that after the pixelization completes. The validation would then be applied unconditionally rather than unconditionally adding an eps. 

But maybe it's better to focus on preventing the invalid elements from being added in the first place? In that case, I should just address the problem in a pre-processing step outside of yt: this all came up with data that I'm loading with the [yt_aspect frontend](https://github.com/data-exp-lab/yt_aspect/) (which now also includes a general .pvtu frontend). I could instead add a validation step when it reads in the .vtk chunks and cull invalid elements before they make it to the pixelizer...

neutrinoceros:
If practical, data (in)validation and early sanitising seems like an excellent option indeed.

chrishavlin:
> If practical, 

I agree... lemme take another look. I'll switch this back to WIP til I have a moment to look next week.

chrishavlin:
Ok, so I looked at addressing this on the data ingestion side (from within yt_aspect), and it's definitely do-able. So I'm leaning towards closing this PR out (and the related issue) and moving work over to yt_aspect. Sound good?

Issue #4342, opened 2023-02-18
BUG: yt.funcs.matplotlib_style_context isn't tested


neutrinoceros:
### Bug report

**Bug summary**

This function hasn't received any care for a while, and it's currently not tested.
Currently, it doesn't produce the expected results

**Code for reproduction**

```python
import matplotlib.pyplot as plt
import yt

with yt.funcs.matplotlib_style_context():
    fig, ax = plt.subplots()
    ax.plot([0, 1], [0, 1])
    ax.set(xlabel=r"$\Sigma$", ylabel=r"$\beta$", title="My fake yt plot")
    fig.savefig("/tmp/yt_style_funcs.png")
```

**Actual outcome**
![yt_style_funcs](https://user-images.githubusercontent.com/14075922/219874734-c9dd9266-ce08-40b8-b89a-cfb0f27d835f.png)

**Expected outcome**

Something like
![yt_style_mpl_context](https://user-images.githubusercontent.com/14075922/219874758-e8588d1d-ac64-462e-b91f-5bb74d1e420c.png)

neutrinoceros:
Actually `yt.funcs.matplotlib_style_context` is used internally in a couple of places, so it's not completely untested, but there are no unit tests for it.

PR #4343, opened 2023-02-18
ENH: add reusable matplotlib stylesheet

neutrinoceros:
## PR Summary
Adresses #4342, but doesn't close it because I'm holding off on automated testing until we get pytest_mpl started.

Here are my manual tests for it

```python
import matplotlib as mpl
import yt

from yt.visualization._commons import MPL_VERSION

print(f'{MPL_VERSION=}')


with yt.funcs.matplotlib_style_context():
    fig, ax = plt.subplots()
    ax.plot([0, 1], [0, 1])
    ax.set(xlabel=r"$\Sigma$", ylabel=r"$\beta$", title="My fake yt plot")
    fig.savefig("/tmp/yt_style_funcs.png")

if MPL_VERSION >= Version("3.7"):
    with plt.style.context("yt.classic"):
        fig, ax = plt.subplots()
        ax.plot([0, 1], [0, 1])
        ax.set(xlabel=r"$\Sigma$", ylabel=r"$\beta$", title="My fake yt plot")
        fig.savefig("/tmp/yt_style_plt_context.png")


    with mpl.style.context("yt.classic"):
        fig, ax = plt.subplots()
        ax.plot([0, 1], [0, 1])
        ax.set(xlabel=r"$\Sigma$", ylabel=r"$\beta$", title="My fake yt plot")
        fig.savefig("/tmp/yt_style_mpl_context.png")

    mpl.style.use("yt.classic")
    fig, ax = plt.subplots()
    ax.plot([0, 1], [0, 1])
    ax.set(xlabel=r"$\Sigma$", ylabel=r"$\beta$", title="My fake yt plot")
    fig.savefig("/tmp/yt_style_mpl_use.png")
```

I still plan on adding a minimal test with no actual image comparison, just to make sure this works with minimal requirements too (which i have checked manually).
I will also need to add documentation.

## PR Checklist


neutrinoceros:
I may have missed the point of `matplotlib_style_context`'s default behaviour. I assumed it was meant to reproduce yt's style, when it may be a simple backport of `matplotlib.style.context` for older versions of matplotlib (in which case it's not needed now).
I will focus on getting CI green for now, and I'll come back to this part.

neutrinoceros:
Actually, turns out this is already minimally tested since `matplotlib_style_context` is used internally.

yt/default.mplstyle
```diff
@@ -0,0 +1,15 @@
...
+ytick.right: True
+xtick.minor.visible: True
+ytick.minor.visible: True
+xtick.direction: in
+ytick.direction: in
+
+font.family: stixgeneral
+
+mathtext.fontset: cm
+mathtext.fallback: cm
```
neutrinoceros, line 15: hint to reviewers: this option is the reason why I'm dropping support for matplotlib 3.2 in the same PR (it's invalid before matplotlib 3.3). Previously this was handled dynamically but obviously this behaviour cannot be replicated with a static style sheet.

neutrinoceros:
For reference here's the current output of the example I'm giving in the docs:
YTSTYLE

Does it feels like yt enough ? Am I missing some of the parameters that make up for yt's style ?

matthewturk:
I do like this idea!  Will go over in detail.

matthewturk (dismissed):
Sorry for letting this languish!  I think it's good to go in, don't you?

neutrinoceros:
Turns out the nice API I was hoping for doesn't exist (yet ?) so I'll just push one more time to fix type-checking, and *now* I believe it's ready to go @matthewturk

neutrinoceros:
Well looking back, I actually ended up removing the one parameter that justified dropping support for MPL 3.2, so I'll revert this now.

PR #4344, opened 2023-02-19
TST: don't install nose on python >= 3.10

neutrinoceros:
## PR Summary

Trying to get rid of nose as a test dependency in environments where it's already unusable. This is a necessary first step for testing with Python 3.12, which should become possible later this year.
As a side effect, expose many test modules that depends on it still.


neutrinoceros:
The one failure is real, but could be filtered (deprecation warning from setuptools via glue)

neutrinoceros:
This now fully resolves #4366, including future proofing for when Python 3.8 and 3.9 patch releases ship with `setuptools>=67.5 `

Xarthisius (approved):

PR #4345, opened 2023-02-19
TST: fixup logger configuration in CI

neutrinoceros:
## PR Summary

fix yt configuration file used in CI

The solution is given by a warning
```
/home/runner/work/yt/yt/yt/utilities/configure.py:96: UserWarning: Could not load configuration file /home/runner/work/yt/yt/yt.toml (invalid TOML: Expected newline or end of document after a statement (at line 1, column 5))
```
fix my own mistake from #4311

### DIFF
--- a/tests/ci_install.sh
+++ b/tests/ci_install.sh
@@ -20,7 +20,7 @@ osx|macOS)
 esac
 
 # Disable excessive output
-echo "[yt]\nlog_level = 50" > yt.toml
+echo "[yt]\nlog_level = 50\n" > yt.toml
 cat yt.toml
 
 # Sets default backend to Agg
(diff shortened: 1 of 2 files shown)


neutrinoceros:
It doesn't seem to be working yet, still a lot of `INFO` level log entries in CI, and the warning isn't gone. Maybe test pollution is causing this, I'll dig deeper as soon as I can.

neutrinoceros:
So I don't quite get why writing the configuration file directly wasn't working, but using yt's CLI instead fixes the problem.
Can anyone have a quick look here ?  maybe @matthewturk , or @Xarthisius ?

Xarthisius (approved):

PR #4346, opened 2023-02-21
TST: split nose --ignore-file args over multiple lines to mitigate conflict frequency

neutrinoceros:
## PR Summary
In response to @matthewturk's suggestion in https://github.com/yt-project/yt/pull/4122#discussion_r1112279143


tests/tests.yaml
```diff
@@ -205,34 +205,36 @@ answer_tests:
 
 other_tests:
   unittests:
+     # keep in sync with nose_ignores.txt
```
neutrinoceros, line 208: this is the only real new line, otherwise I just sorted the arguments in alphabetical order and made the style more consistent

neutrinoceros:
@matthewturk can you have a look ?
I'll need to update a couple PRs after that for a (hopefully) last round of conflict resolution

matthewturk:
OK, I need to ask -- do you think this is an improvement?  I'm actually not sure, since it requires a separate file.

neutrinoceros:
I think it's increasingly worth it. Having to append to a single line was fine as when we started using pytest a couple years back but most of the tests we add these days are pytest only, so conflicts now pop up very regularly.

matthewturk (approved):

matthewturk:
let's go ahead and delete the branch `split_blockchain` shall we :stuck_out_tongue_closed_eyes:

PR #4347, opened 2023-02-21
MAESTROeX uses namespaces for paramters

zingale:
this fixes the detection of how temperature is defined

## PR Checklist

### DIFF
--- a/yt/frontends/boxlib/fields.py
+++ b/yt/frontends/boxlib/fields.py
@@ -448,8 +448,10 @@ class MaestroFieldInfo(FieldInfoContainer):
         # pick the correct temperature field
         tfromp = False
         if "use_tfromp" in self.ds.parameters:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Seems pretty straightforward. Does the exact name of this parameter depend on MAESTROeX's version ?

zingale:
MAESTRO is the old code, written in F90.  It is really not used much any more.  MAESTROeX is the rewrite in C++, and it uses namespaces for the parameters, hence the change in the parameter name.  I don't expect any more changes.

neutrinoceros:
Ok, can you add inline comments to indicate what branch is destined to what code ?

zingale:
> Ok, can you add inline comments to indicate what branch is destined to what code ?

done

neutrinoceros (approved):

PR #4362, opened 2023-03-03
enable zooming into spherical center

jisuoqing:
Before this change, the plot center along `theta` axis of a spherical dataset is fixed at `(r_min + r_max) / 2`, and ignores the parameter `center` input from `PlotWindow`. In this case, zooming into the spherical center at `r = 0` is impossible.

This change is similar to the treatment under cylindrical geometries: whenever viewed from the axisymmetric axis (`theta` in cylindrical and `phi` in spherical), we always use existing center values to enable zooming into the axisymmetric center:

https://github.com/yt-project/yt/blob/7e90926ba3b687b520b29aca5a2145075e5c7664/yt/geometry/coordinates/cylindrical_coordinates.py#L212-L224

### DIFF
--- a/yt/geometry/coordinates/spherical_coordinates.py
+++ b/yt/geometry/coordinates/spherical_coordinates.py
@@ -343,6 +343,11 @@ class SphericalCoordinateHandler(CoordinateHandler):
 
     def sanitize_center(self, center, axis):
         center, display_center = super().sanitize_center(center, axis)
+        display_center = [
+            0.0 * display_center[0],
+            0.0 * display_center[1],
+            0.0 * display_center[2],
+        ]
         name = self.axis_name[axis]
         if name == "r":
             xxmin, xxmax, yymin, yymax = self._aitoff_bounds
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Thanks ! I'll review this later today. A minimal reproducer or a test would be useful here

jisuoqing:
Thank you for taking care of this. It seems that for curvilinear datasets, the `center` coordinates along some axis are fixed by the sanitizer, ignoring whatever passed in via `PlotWindow`. Thus, we cannot select an arbitrary center for plotting or zoom-in. But at least this PR makes it possible to zoom in to the spherical center which is more intuitive (same behavior with the cylindrical case).

matthewturk:
@neutrinoceros I know you have expressed reservations about major changes to the zooming, but do you think that there is a clean way to have @jisuoqing 's work bypass the sanitizer in some cases to enable functionality?

neutrinoceros:
I need to think about it. In any case the patch should probably a little bit more general and also consider what happens in the y direction too, as well as the toroidal (r-normal and theta-normal) cases.

neutrinoceros:
Alright, so I think the issue is real (not being able to specify plot bounds in spherical geometries), but this solution is intrusive (in the sense that it breaks the local symmetry in the code) and more importantly, incomplete. I would rather we discuss what a viable general solution would look like in an issue and close this PR for now, if that's okay with you @jisuoqing ?

matthewturk:
Can we leave the PR open and mark it as draft while we work toward a solution?

jisuoqing:
I'm ok with both! Please go ahead with whatever you think is more convenient :)

neutrinoceros:
> Can we leave the PR open and mark it as draft while we work toward a solution?

I think the discussion should be centralised (so I'd prefer to have an issue OR a PR but not both), and I have a preference for using an issue, but, your call.

matthewturk:
I guess my main concern is that if we end up coming back to this, I want to have an easy way to re-use this work rather than re-implement.  But perhaps if we close, but leave a pointer to this PR in an issue, we can re-open it.

neutrinoceros:
Currently it's just +1/-1 so it wouldn't be a problem to reimplement it, but indeed I was intending to point to this PR as the origin of the discussion.

PR #4363, opened 2023-03-03
ENH: add basename for loaders

jisuoqing:
The `handler.name` are hard-coded with a string in dataset loaders, and assigning values to `ds.basename` will pop out an `AttributeError: can't set attribute 'basename'` (which is allowed by previous yt versions, though). 

Here an additional parameter `basename` for dataset loaders is proposed, so users can load multiple datasets with different base names and analyze/visualize them in a workflow (e.g., saving plots with different filenames without worrying about overwriting each other.)


yt/loaders.py
```diff
@@ -202,6 +202,7 @@ def load_uniform_grid(
     axis_order: Optional[AxisOrder] = None,
     cell_widths=None,
     parameters=None,
+    basename=None,
```
neutrinoceros, line 205:
```suggestion
    basename: str = "UniformGridData",
```

yt/loaders.py
```diff
@@ -385,7 +388,10 @@ def load_uniform_grid(
         parameters=parameters,
     )
 
-    handler.name = "UniformGridData"  # type: ignore [attr-defined]
+    if basename is None:
+        handler.name = "UniformGridData"  # type: ignore [attr-defined]
+    else:
+        handler.name = basename  # type: ignore [attr-defined]
```
neutrinoceros, line 394:
```suggestion
    handler.name = basename  # type: ignore [attr-defined]
```

neutrinoceros (changes requested):
I think we can use actual default values instead of `None`. I've made explicit suggestions in one case but it should be generalised

jisuoqing:
Thanks! Learnt a nicer way.

yt/loaders.py
```diff
@@ -1144,7 +1159,7 @@ def load_octree(
         parameters=parameters,
     )
 
-    handler.name = "OctreeData"
+    handler.name = basename
```
chrishavlin, line 1162: You're getting those type checking failures cause you added the `str` annotation to the new argument. I think fixing the type checking here is well beyond the scope of your PR (you're only exposing it here because this is the first type-annotated argument to this function), so I think it's fine to simply add more ignore directives similar to the other loaders. For this line, you'll want:

```suggestion
    handler.name = basename  # type: ignore [attr-defined]
```

But then you'll also need to add that `# type: ignore [attr-defined]` comment to all of the other lines mentioned in the test failure too. Here are the lines for reference (copied from [here](https://github.com/yt-project/yt/actions/runs/4331206959/jobs/7562899186)):

```
yt/loaders.py: note: In function "load_octree":
yt/loaders.py:1162: error: "StreamHandler" has no attribute "name"  [attr-defined]
yt/loaders.py:1163: error: "StreamHandler" has no attribute "domain_left_edge"  [attr-defined]
yt/loaders.py:1164: error: "StreamHandler" has no attribute "domain_right_edge"  [attr-defined]
yt/loaders.py:1165: error: "StreamHandler" has no attribute "refine_by"  [attr-defined]
yt/loaders.py:1166: error: "StreamHandler" has no attribute "dimensionality"  [attr-defined]
yt/loaders.py:1167: error: "StreamHandler" has no attribute "domain_dimensions"  [attr-defined]
yt/loaders.py:1168: error: "StreamHandler" has no attribute "simulation_time"  [attr-defined]
yt/loaders.py:1169: error: "StreamHandler" has no attribute "cosmology_simulation"  [attr-defined]
yt/loaders.py:1174: error: "StreamOctreeDataset" has no attribute "octree_mask"  [attr-defined]
yt/loaders.py:1175: error: "StreamOctreeDataset" has no attribute "partial_coverage"  [attr-defined]
yt/loaders.py:1176: error: "StreamOctreeDataset" has no attribute "num_zones"  [attr-defined]
```

chrishavlin: alternatively, you could change `basename: str = "OctreeData",` to `basename = "OctreeData"`, but I'm in slight favor of adding all the typing ignores. @neutrinoceros do you agree with that?

chrishavlin (commented):
I like this! Left a comment on fixing the type check failures.

yt/loaders.py
neutrinoceros: I do !

jisuoqing: Thanks @chrishavlin ! I do see this error ignoring in the original code for all other types of datasets except `OctreeData`, so I was wondering whether there's a particular reason for that. But now as you suggested, I add those lines for `OctreeData` too.

neutrinoceros:
Had a couple minutes to think about this patch, and while I don't want to oppose it, I want to raise a couple important issues with the design.

> assigning values to ds.basename will pop out an AttributeError: can't set attribute 'basename' (which is allowed by previous yt versions, though).

Setting `ds.basename` is disallowed in yt 4.1 because it breaks the consistency that `ds.basename`, `ds.directory`, `ds.fullpath` and `ds.backup_filename` are supposed to have (they all relate to `ds.filename`). This is why they are now implemented as properties.
The fundamental (private) attribute from which all these properties are derived is `ds._input_filename`, so shouldn't *this* is the attribute that loaders should be setting in order to control the output image prefixes ? (this question isn't rhetorical)

Furthermore, wouldn't `image_prefix` make more sense than `basename` as a parameter ? I think it conveys the intention much better. If not, it means I didn't get what it was meant for.

The image filename generating code is already difficult to maintain, and I'll fully admit that I didn't grasp the full picture in 20min of exploration. I think it's important that we try our best not to make the code even harder to reason about.
In short, we need to have a very clear understanding of how setting `handler.name` results in setting an image prefix, and simplify the code if we can rather than making it more complex.

jisuoqing:
Thanks for your comments! Sorry for my late response since I'm traveling these days. I agree to make the logic simpler rather than more complex, while it might be better to make the name customizable than hard-coded? I raise this question when I have to read a series of dataset (say, with filename `snapshot_000`, `snapshot_001`, etc.), load the content with some loader (say, `load_uniform_grid`) and perform analysis with parallelism, but it sometime causes trouble when all loaded data comes with the same base name (say, `UniformGrid`), including the image prefix issue mentioned before. Therefore, it might be more natural if the base name from the loaders can be customized into `snapshot_000`, `snapshot_001` etc. Not sure whether there's a way to achieve this for now.
I think this PR might be helpful, but would not insist. I total understand that `ds.basename` should not be changed.

neutrinoceros:
> Not sure whether there's a way to achieve this for now.

It is certainly achievable if you construct the filename outside yt and pass it to the `save` method instead of relying on the default (which I agree are not ideal).

jisuoqing:
> It is certainly achievable if you construct the filename outside yt and pass it to the `save` method instead of relying on the default (which I agree are not ideal).

Let me double-confirm: I used to wrap up a data loader into a custom function `ds = load_dataset(fn)` in which the function `load_uniform_grid` is called. Is it possible to let the returned `ds` have the basename of `fn` (so any yt functionality can deal with this `ds` without any issue)?

neutrinoceros:
Thanks @chrishavlin for the detailed walkthrough. I missed that `StreamHandler` overrides the `filename` property, it's much clearer now.
So it seems to me there is plenty room to simplify this internal mess of implementation details, but I'm confident now that this PR can be implemented without worrying about it. The one thing that should be properly discussed for the present patch, since it exposes a new keyword argument, is what we should name it (because we won't be able to change it afterward).

As I already pointed, I'm personally -1 on `basename` and I would rather go for something like `image_prefix`. What do you guys think ?

chrishavlin:
Ya, I agree -1 for `basename`, for all the reasons you point out. I'm also not so sure about `image_prefix` since it's a little more general than that (e.g., `print(ds)` is not obviously related to an image prefix). How about just `dataset_name` with an explanation in the docstring that this is used in place of a `filename` for in-memory stream datasets?

matthewturk:
Is it just images we prefix with this?  Do we use it for sidecar files,
ever?

neutrinoceros:
> How about just dataset_name with an explanation in the docstring that this is used in place of a filename for in-memory stream datasets?

Sounds good to me

chrishavlin:
>Is it just images we prefix with this?  Do we use it for sidecar files, ever?

So from what I can tell, in general, the `ds.filename` attribute ends up being used in different ways (sometimes differing by frontend) and in some cases, yes, it is ultimately used to write to or read from sidecar files: 

**`ds.parameter_file`** is a direct alias to `ds.filename` and that gets used all over the frontends and is often passed down to index objects where it's stored as an `index_filename`. And that's used, e.g., by the `ParticleIndex._initialize_index` to check for existing  `.ewah` files:

```python
        # Load Morton index from file if provided
        fname = getattr(ds, "index_filename", None) or f"{ds.parameter_filename}.ewah"
```    
or to store sidecar `kdtree` files for `SPHDataset`. 


**`storage_filename`** will be derived from `ds.basename` if it's not set and needed for `Index._initialize_data_storage()`. 

I'm not sure how much of that is relevant to the Stream frontend in particular, but I think that yes, the `stream_handler.name` would be used for any sidecar files (if they're used at all). 

All that to say, yes, the `stream_handler.name` will be used for more than just image prefixing so that's  a good reason to use a more general variable name like `dataset_name` as the new keyword argument here.

jisuoqing:
Thanks! Happy to make the change after hearing back from others.

matthewturk:
Let's do it!

neutrinoceros (approved):
thank you ! LGTM

chrishavlin (approved):
Great, thanks @jisuoqing !

Issue #4355, opened 2023-03-02
BLD: compatibility with Cython 3.0


neutrinoceros:
Cython 3.0 is now in beta (after several years of alpha).

#4043 contains an initial effort to discover and patch incompatibilities, but at the moment it seems to me that this migration is better done incrementally, as only some of the patches will be seamless and uncontroversial.
I'll be using this issue to track progress on this effort.

The most urgent issues are listed in the following. A useful resource for this task is the [migration guide](http://docs.cython.org/en/latest/src/userguide/migrating_to_cy30.html).

### blockers
- [x] #4146
- [x] #4357
- [x] #4365 
- [x] #4373
- [x] #4374

### restoring performance
- [x] #4386
- [x] #4390
- [x] #4392 

performance issues are discussed in the present thread.

### deprecation warnings
- [x] #4375
- [ ] `The 'DEF' statement is deprecated and will be removed in a future Cython version`. 
  - [x] see #4043 for a first attempt at removing these
  - [x] #4376
  - [ ] https://github.com/cython/cython/pull/5242

Since the migration path for some items is a moving target (see https://github.com/cython/cython/pull/5242), I think the minimal effort approach would be to set our requirement to `Cython>=3.0,<3.1` at first. According to the current plan for Cython 3.1 and 3.2, we should be able to drop the upper limit completely once we reach compatibility with version 3.2

### low priority (expected gains from dropping Cython 0.29.x)
- [ ] forbid deprecation warnings from numpy C-API (see http://docs.cython.org/en/latest/src/userguide/migrating_to_cy30.html#numpy-c-api)
This one is easy, as it just requires adding some compiler flags to all `*.pyx` files, but it requires dropping Cython 0.29.x first.
For future reference, here's a simple way to apply the change systematically
<details><summary> Details </summary>

with `header.txt`
```shell
# distutils: define_macros=NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION
```
and `t.sh`
```shell
for file in "$@"
do
    cat header.txt "$file" > /tmp/xx.$$
    mv /tmp/xx.$$ "$file"
done
```
apply the change as
```
git ls-files | grep .pyx | xargs bash t.sh
```

</details>

- [ ] (?) Start releasing more portable wheels with `PY_LIMITED_API` https://cython.readthedocs.io/en/latest/src/changes.html?highlight=CYTHON_LIMITED_API#id27

neutrinoceros:
[...]

PR #4365, opened 2023-03-05
BLD: fix compilation errors with Cython 3.0.0b1 (pow return type)

neutrinoceros:
## PR Summary

Fix most of the remaining compilation errors with Cython 3.0.0b1

Address the current top blocking problem in #4355
For context see https://github.com/cython/cython/issues/5299

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -84,7 +84,7 @@ jobs:
     - name: Install build time dependencies
       shell: bash
       run: |
-        python -m pip install "Cython>=0.29.21,<3.0"
+        python -m pip install "Cython>=0.29.33,<3.0"
         python -m pip install oldest-supported-numpy
         python -m pip install --upgrade wheel
         python -m pip install --upgrade setuptools
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ requires = [
   # Cython 3.0 is the next version after 0.29, and a major change,
   # we forbid it until we can properly test against it
   # https://github.com/yt-project/yt/issues/4355
-  "Cython>=0.29.21,<3.0",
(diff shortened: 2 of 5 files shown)


chrishavlin (dismissed):
Generally looks good -- one naive question: if `cython.cpow` is introduced in 3.0, does 2.9x just ignore it?

neutrinoceros:
Great question, I actually hadn't noticed that this decorator was new in Cython 3.0 and now I'm surprised that Cython 0.29 seems just happy to swallow it. I upstreamed your question (see https://github.com/cython/cython/issues/5299#issuecomment-1464560241)

neutrinoceros:
So, this decorator was added to Cython 0.29.x as a no-op for backward compatibility, see [Cython's changelog](https://cython.readthedocs.io/en/latest/src/changes.html#id34)
but it *does* require Cython>=0.29.33, so I'll need to bump the requirement explicitly.

chrishavlin (approved):

.github/workflows/wheels.yaml
```diff
@@ -84,7 +84,7 @@ jobs:
     - name: Install build time dependencies
       shell: bash
       run: |
-        python -m pip install "Cython>=0.29.21,<3.0"
```
neutrinoceros, line 87: forced-pushed again because I forgot to update this line too

neutrinoceros:
Can we merge this to ease up work on the last compilation error ?
@chrishavlin @matthewturk ?

chrishavlin (approved):

chrishavlin:
> Can we merge this to ease up work on the last compilation error ?

OK by me -- I assume bumping the minimum version of cython from 0.29.21 to 0.29.33 shouldn't be an issue, right? seems totally reasonable to me... 

@matthewturk ?

Issue #4366, opened 2023-03-06
BUG: pkg_resources is now fully deprecated


neutrinoceros:
### Bug report

**Bug summary**

`pkg_resources` is still used in `nose`. Nightly tests are failing as a result
https://github.com/yt-project/yt/actions/runs/4339556725

This is already solved in #4344

see https://setuptools.pypa.io/en/latest/history.html#v67-5-0

neutrinoceros:
looks like this is also breaking normal CI because the Windows job gets upgrade `setuptools` for some reason. Upgrading this to "blocker" status.

PR #4368, opened 2023-03-09
[WIP] ENH: Specify negative imaginary components to region expression

matthewturk:
## PR Summary

Currently, we can specify things like:

```python
g = ds.r[::128j,::128j,::128j]
s = ds.r[::128j,::128j,0.5]
```

to get arbitrary grids and fixed resolution buffers with specified resolutions.
At present, getting this in the native resolution of a dataset (for instance in
unigrid) requires looking up the domain dimensions etc etc, and that gets quite
clumsy.

So instead, I added the ability to specify *negative* complex components, which
then act as the multiplicative factor to the base resolution of the domain.

For instance, in a `[128, 128, 128]` dataset, this would provide a `[256, 128,
512]` size grid:

```python
g = ds.r[::-2j,::-1j,::-4j]
```

Same for slices (but the axes there are based on the x/y axes, so it's not in
the same order.)

One specific thing this will make easier is supplying things to `matplotlib`'s
`imshow` and using different interpolation techniques.

## PR Checklist

### DIFF
--- a/yt/data_objects/region_expression.py
+++ b/yt/data_objects/region_expression.py
@@ -136,8 +136,23 @@ class RegionExpression:
             # We'll do this by getting the x/y axes, and then using that.
             width = source.right_edge[xax] - source.left_edge[xax]
             height = source.right_edge[yax] - source.left_edge[yax]
-            # Make a resolution tuple with
+            # We have two cases here.  The first is that the imaginary component
+            # is *positive*.  In that case, the component is the number of
+            # values to sample.
+            # In the case that it is *negative*, the absolute value of it
+            # corresponds to the factor of the base resolution of the dataset in
+            # that dimension.
             resolution = (int(new_slice[xax].step.imag), int(new_slice[yax].step.imag))
+            if resolution[0] < 0:
+                resolution = (
+                    self.ds.domain_dimensions[xax] * -resolution[0],
+                    resolution[1],
+                )
+            if resolution[1] < 0:
+                resolution = (
+                    resolution[0],
+                    self.ds.domain_dimensions[yax] * -resolution[1],
+                )
             sl = sl.to_frb(width=width, resolution=resolution, height=height)
         return sl
 
@@ -164,7 +179,12 @@ class RegionExpression:
             right_edge[ax] = r
             d = getattr(b.step, "imag", None)
             if d is not None:
+                # This allows us (like with slices) to specify negative
+                # imaginary components to get a multiple of the domain size.
+                if d < 0:
+                    d = self.ds.domain_dimensions[ax] * -d
                 d = int(d)
+
             dims[ax] = d
         center = [(cl + cr) / 2.0 for cl, cr in zip(left_edge, right_edge)]
         if None not in dims:


neutrinoceros:
FTR, failures on windows are unrelated (see #4366)

jzuhone:
Ooo fancy

doc/source/analyzing/objects.rst
```diff
@@ -243,6 +263,20 @@ domain but centered at 0.5 in code units, you can do:
 This ``frb`` object then can be queried like a normal fixed resolution buffer,
 and it will return arrays of shape (1024, 1024).
 
+As above, you can also specify with *negative* complex step sizes to indicate
```
chrishavlin, line 266:
```suggestion
As noted above, you can also specify with *negative* complex step sizes to indicate
```

chrishavlin (commented):
Small docs suggestion. 

And a question -- what should happen when a user also specifies the bounds? e.g., I just tried

```
ds = fake_random_ds([9, 17, 23])
region = ds.r[(1, "mm"):(3, "mm"):-1j, (3, "mm"):(4, "mm"):-1j, (7, "mm"):(10, "mm"):-2j]
```
and it returns a grid between the bounds using a resolution that is a factor of the whole domain. It makes total sense to me given that I read through the implementation... but maybe you should add a note to the docs? Or would it make sense to find the resolution of the selected domain and then multiply by a factor (not sure if this actually makes any sense)?

yt/data_objects/region_expression.py
```diff
@@ -136,8 +136,23 @@ def _create_slice(self, slice_tuple):
...
+                resolution = (
+                    self.ds.domain_dimensions[xax] * -resolution[0],
+                    resolution[1],
+                )
+            if resolution[1] < 0:
+                resolution = (
+                    resolution[0],
+                    self.ds.domain_dimensions[yax] * -resolution[1],
+                )
             sl = sl.to_frb(width=width, resolution=resolution, height=height)
```
neutrinoceros, line 156: Modifying a tuple's content is a little awkward. I think it would flow much more naturally if we avoided making a tuple too early.
```suggestion
            if xres := int(new_slice[xax].step.imag) < 0:
                xres *= - self.ds.domain_dimensions[xax]
            if yres := int(new_slice[yax].step.imag) < 0:
                yres *= - self.ds.domain_dimensions[yax]
            sl = sl.to_frb(width=width, resolution=(xres, yres), height=height)
```

doc/source/analyzing/objects.rst
```diff
@@ -200,6 +200,26 @@ could supply:
 This can select both particles and mesh fields.  Mesh fields will be 3D arrays,
 and generated through volume-weighted overlap calculations.
 
+.. note::
```
neutrinoceros, line 203: isn't this note repetitive with the following paragraph ?

doc/source/analyzing/objects.rst
```diff
@@ -200,6 +200,26 @@ could supply:
...
+
+   You can now select easily specify that the resolution should be a multiple
+   of the domain resolution!
+
+By using negative complex integers you can specify that the size be a multiple
+of the underlying domain dimensions.  For instance, if you don't know the size
+of the dataset's base dimensions, you can specify ``-1j`` as the step in a slice
+and get back the native coordinates.
```
neutrinoceros, line 211: This is not clear to me. What if my grid is AMR in that direction ? Do I get to keep the native (varying) resolution, or does everything get interpolated to the lowest (base) resolution ?

Issue #4371, opened 2023-03-11
Continued problems with stored EWAH files


jzuhone:
### Bug report

After https://github.com/yt-project/yt/pull/4307/, there are still some problems with loading the particle bitmap index from the EWAH sidecar file. 

**Code for reproduction**

```python
import yt
ds = yt.load("cutout_58.hdf5")
sp = ds.sphere([8142.44, 14341.0, 25125.9], (20.0, "kpc"))
print(sp.sum(("PartType0", "particle_mass")))
```

**Actual outcome**

Before dumping the EWAH file:

```pycon
yt : [INFO     ] 2023-03-11 12:32:41,144 Calculating time from 1.000e+00 to be 4.356e+17 seconds
yt : [INFO     ] 2023-03-11 12:32:41,182 Parameters: current_time              = 4.355810528213311e+17 s
yt : [INFO     ] 2023-03-11 12:32:41,182 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2023-03-11 12:32:41,183 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2023-03-11 12:32:41,183 Parameters: domain_right_edge         = [35000. 35000. 35000.]
yt : [INFO     ] 2023-03-11 12:32:41,183 Parameters: cosmological_simulation   = 1
yt : [INFO     ] 2023-03-11 12:32:41,183 Parameters: current_redshift          = 2.220446049250313e-16
yt : [INFO     ] 2023-03-11 12:32:41,183 Parameters: omega_lambda              = 0.6911
yt : [INFO     ] 2023-03-11 12:32:41,183 Parameters: omega_matter              = 0.3089
yt : [INFO     ] 2023-03-11 12:32:41,183 Parameters: omega_radiation           = 0.0
yt : [INFO     ] 2023-03-11 12:32:41,183 Parameters: hubble_constant           = 0.6774
yt : [INFO     ] 2023-03-11 12:32:41,211 Allocating for 1.504e+07 particles
Initializing coarse index : 100%|██████████████████████████████████████████████████████████| 37/37 [00:00<00:00, 96.12it/s]
yt : [INFO     ] 2023-03-11 12:32:41,616 Updating index_order2 from 2 to 4
Initializing refined index: 100%|██████████████████████████████████████████████████████████| 37/37 [00:01<00:00, 27.47it/s]
0.598708856382018 code_mass
```

After it is read in on the second time running the same script:

```pycon
[...]

PR #4372, opened 2023-03-11
BUG: Fix index order if loading from EWAH file

jzuhone:
## PR Summary

PR #4307 allowed for the possibility of storing multiple particle indices in the EWAH sidecar file. It stores also the `index_order` in the file, _after_ there is the possibility of the second value in the tuple being updated. 

When this file is reloaded, the updated `index_order` is not taken into account, and the bitmasks that are loaded from the file, which are consistent with the `index_order` in the file, are not the ones consistent with the `index_order` yt thinks it is using (because the update in yt only occurs if the indices are being generated from scratch).

This PR fixes the issue by ensuring that we load the correct `index_order` from the file, and tell the `ParticleBitmap` and the `Dataset`'s `index' accordingly. 

Closes issue #4371. 

## PR Checklist

### DIFF
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -173,6 +173,9 @@ class ParticleIndex(Index):
         try:
             if dont_load:
                 raise OSError
+            # We must make sure the index order is correct, here and
+            # in the particle bitmap
+            self.ds.index_order = self.regions.correct_index_order(fname)
             rflag = self.regions.load_bitmasks(fname)
             rflag = self.regions.check_bitmasks()
             self._initialize_frontend_specific()
--- a/yt/geometry/particle_oct_container.pyx
+++ b/yt/geometry/particle_oct_container.pyx
@@ -1033,6 +1033,21 @@ cdef class ParticleBitmap:
     def reset_bitmasks(self):
         self.bitmasks._reset()
 
+    def correct_index_order(self, fname):
+        import h5py
+        # Verify that file is correct version
+        if not os.path.isfile(fname):
+            raise OSError
+        with h5py.File(fname, mode="r") as fp:
+            try:
+                grp = fp[str(self.hash_value)]
+            except KeyError:
+                raise OSError(f"Index not found in the {fname}")
+            index_order = grp.attrs["index_order1"], grp.attrs["index_order2"]
+        self.index_order1 = index_order[0]
+        self.index_order2 = index_order[1]
+        return index_order
+
     def load_bitmasks(self, fname):
         import h5py
         cdef bint read_flag = 1


yt/geometry/particle_oct_container.pyx
```diff
@@ -1033,6 +1033,21 @@ cdef class ParticleBitmap:
     def reset_bitmasks(self):
         self.bitmasks._reset()
 
+    def correct_index_order(self, fname):
```
Xarthisius, line 1036: I don't understand what's happening here. Since `self.hash_value` is a function of `self.index_order1` and `self.index_order2` there's no way that `grp.attrs["index_order1"], grp.attrs["index_order2"]` read from cache will be different from `self.index_order1` and `self.index_order2`

yt/geometry/particle_geometry_handler.py
```diff
@@ -173,6 +173,9 @@ def _initialize_index(self):
         try:
             if dont_load:
                 raise OSError
+            # We must make sure the index order is correct, here and
+            # in the particle bitmap
+            self.ds.index_order = self.regions.correct_index_order(fname)
```
Xarthisius, line 178: Taking into account my comment below, I think this should be rewritten as:
```suggestion
            self.ds.index_order = (self.regions.index_order1, self.regions.index_order2)
```

yt/geometry/particle_oct_container.pyx
jzuhone: @Xarthisius here's what happens in order (in `main` right now).

These two steps are the same regardless of whether or not the EWAH file already exists:

1. yt sets `index_order1`, `index_order2`
2. yt creates `ParticleBitmap` object as `regions` attribute of `index` (line 162 of `particle_geometry_handler.py`), passes it _initial_ values of `index_order1` and `index_order2` to create `hash_value` in the `ParticleBitmap` object

Now the "a" case is what happens when there is no EWAH file:

3a. If there is no EWAH file yet, yt creates coarse and fine indices, and (sometimes) decides that `index_order2` needs to be updated. 
4a. yt writes EWAH file--with the _updated_ `index_order2` as the underlying, but `hash_value` is not affected (and hence not the HDF5 group name) since it was already set and it is not updated.

The "b" case is what happens if there is an EWAH file:

3b. yt reads the bitmap masks from the HDF5 group, has no problem doing so because yt is using the `hash_value` from initial `index_order1` and `index_order2`.
4b. Because we read the masks from the EWAH file, yt never drops into `_initialize_coarse_index` where the index order is updated, so whenever we do a calculation, the initial `index_order2` is used but the bitmap masks are using the updated one

Very sneaky...

Xarthisius: > 4a. yt writes EWAH file--with the _updated_ `index_order2` as the underlying, but `hash_value` is not affected (and hence not the HDF5 group name) since it was already set and it is not updated.

That's the bug! It means I messed it up. `hash_value` should always match current `index1` and `index2`, not just whatever was passed to the constructor.

Xarthisius: Also the notion that we store meaningful value, that's not just some initial guess, in `ds.index_order` is scary. What happens if I select two spheres and want them to use different indices? Is that disallowed?

jzuhone: So that means you have to update it I guess before it gets written--but now the problem is that yt when it's searching for the masks in the EWAH file never gets to the place where it realizes that index_order2 should be updated, and thus it has the wrong hash the second time around. So that's still an issue.

jzuhone: So this means we have to open the EWAH file before even creating the `regions` object so we can determine the correct `index_order2`, or something like that. Previous to the HDF5 change, this was handled by reading `index_order2` off from the filename, which the current code is not doing at all

PR #4378, opened 2023-03-20
BUG: Fix typo in vector operations and enable plot modifications for spherical geometry

jisuoqing:
- Fixed typos in vector operations in spherical geometry
- Enable `Streamline` and `LineIntegralConvolution` callbacks for spherical geometry
- Callback tests added

Some further thoughts -- when using `Quiver`, `Streamline` and `LineIntegralConvolution` callbacks for curvilinear geometries, we assume users know what vector fields to plot (`_cartesian_x`, `_conic_x`, `_cylindrical_radius` etc.), while this is obscure and hidden in the source code. Maybe improvement could be made in future to generate a warning message when users try to use `inappropriate` vector fields in those callbacks.

### DIFF
--- a/yt/fields/vector_operations.py
+++ b/yt/fields/vector_operations.py
@@ -529,7 +529,7 @@ def create_vector_fields(
                         * np.cos(data[(ftype, "phi")])
                         + data[(ftype, f"{basename}_theta")]
                         * np.cos(data[(ftype, "theta")])
-                        * np.cos([(ftype, "phi")])
+                        * np.cos(data[(ftype, "phi")])
                         - data[(ftype, f"{basename}_phi")]
                         * np.sin(data[(ftype, "phi")])
                     )
@@ -546,7 +546,7 @@ def create_vector_fields(
                         * np.sin(data[(ftype, "phi")])
                         + data[(ftype, f"{basename}_theta")]
                         * np.cos(data[(ftype, "theta")])
-                        * np.sin([(ftype, "phi")])
+                        * np.sin(data[(ftype, "phi")])
                         + data[(ftype, f"{basename}_phi")]
                         * np.cos(data[(ftype, "phi")])
                     )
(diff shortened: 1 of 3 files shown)


yt/fields/vector_operations.py
```diff
@@ -529,7 +529,7 @@ def _cartesian_x(field, data):
                         * np.cos(data[(ftype, "phi")])
                         + data[(ftype, f"{basename}_theta")]
                         * np.cos(data[(ftype, "theta")])
-                        * np.cos([(ftype, "phi")])
+                        * np.cos(data[(ftype, "phi")])
```
neutrinoceros, line 532: oof, thank you for catching these !

neutrinoceros (commented):
Thanks ! could you provide a simple demo script so we can verify that Streamlines and LIC callbacks just work ?

neutrinoceros:
> Some further thoughts -- when using Quiver, Streamline and LineIntegralConvolution callbacks for curvilinear geometries, we assume users know what vector fields to plot (_cartesian_x, _conic_x, _cylindrical_radius etc.), while this is obscure and hidden in the source code.
Maybe improvement could be made in future to generate a warning message when users try to use inappropriate vector fields in those callbacks.

+1/2 on this. Available fields could certainly be better documented, but I think `Quiver`, `Streamline` and `LineIntegralConvolution` callbacks *should* in principle accept any pair of scalar fields (i.e., not flag any usage as "inappropriate").

neutrinoceros:
> Should a warning message appear to give a hint for the users?

IMHO it shouldn't. While plotting velocity vectors and magnetic fields are by far the most common use cases for these callbacks, I don't think that we should invalidate creative usage in any way (no no errors, no warnings).

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):

PR #4387, opened 2023-03-28
ENH: Add millimeter to smallest units

matthewturk:
This is a very small fix.  For datasets that are in units of centimeters, any
zooming in on plots will switch from 1cm to 1e3 um in the base units of the
image axes.

This minor change inserts `mm` as an option for switching, which changes it to
allow for a middle-ground.  Seconds already allows `ms` so I think this should
also be fine.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -532,6 +532,7 @@ class Dataset(abc.ABC):
                 "rsun",
                 "km",
                 "cm",
+                "mm",
                 "um",
                 "nm",
                 "pm",
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -531,6 +531,7 @@ class Dataset(abc.ABC):
                 "au",
                 "rsun",
                 "km",
+                "m",
                 "cm",
                 "mm",
                 "um",
--- a/yt/visualization/tests/test_fits_image.py
+++ b/yt/visualization/tests/test_fits_image.py
@@ -100,8 +100,8 @@ def test_fits_image():
 
     assert_equal(fits_slc2["density"].data, fits_slc["density"].data)
     assert_equal(fits_slc2["temperature"].data, fits_slc["temperature"].data)
-    assert fits_slc.wcs.wcs.crval[0] == ds.domain_center[0].to_value("cm")
-    assert fits_slc.wcs.wcs.crval[1] == ds.domain_center[1].to_value("cm")
+    assert fits_slc.wcs.wcs.crval[0] != 0.0
+    assert fits_slc.wcs.wcs.crval[1] != 0.0
     assert fits_slc2.wcs.wcs.crval[0] == 0.0
     assert fits_slc2.wcs.wcs.crval[1] == 0.0
 


neutrinoceros:
I'm tempted to triage this as a bugfix since the previous behaviour was likely never intended. I've seen this problem before, but do you have a reprod that we can come back for posterity ?

yt/data_objects/static_output.py
```diff
@@ -532,6 +532,7 @@ def get_smallest_appropriate_unit(
                 "rsun",
                 "km",
                 "cm",
```
neutrinoceros, line 534: I guess "m" should also fill in the gap between "km" and "cm" ?

matthewturk:
Oops, forgot to attach, but here's current output.

image

neutrinoceros (dismissed):

neutrinoceros:
woops, one real failure, let's turn auto-merge off

neutrinoceros:
I think the failing test should be modified. It seems to be too tightly coupled with implementation details
This is how I'd fix it
```patch
diff --git a/yt/visualization/tests/test_fits_image.py b/yt/visualization/tests/test_fits_image.py
index 147b8abbf..26cc26a48 100644
--- a/yt/visualization/tests/test_fits_image.py
+++ b/yt/visualization/tests/test_fits_image.py
@@ -100,8 +100,8 @@ def test_fits_image():
 
     assert_equal(fits_slc2["density"].data, fits_slc["density"].data)
     assert_equal(fits_slc2["temperature"].data, fits_slc["temperature"].data)
-    assert fits_slc.wcs.wcs.crval[0] == ds.domain_center[0].to_value("cm")
-    assert fits_slc.wcs.wcs.crval[1] == ds.domain_center[1].to_value("cm")
+    assert fits_slc.wcs.wcs.crval[0] != 0.0
+    assert fits_slc.wcs.wcs.crval[1] != 0.0
     assert fits_slc2.wcs.wcs.crval[0] == 0.0
     assert fits_slc2.wcs.wcs.crval[1] == 0.0
```

Though if you're worried about loosened comparison, here's the patch that would preserve the current spirit of the test
```patch
diff --git a/yt/visualization/tests/test_fits_image.py b/yt/visualization/tests/test_fits_image.py
index 147b8abbf..26cc26a48 100644
--- a/yt/visualization/tests/test_fits_image.py
+++ b/yt/visualization/tests/test_fits_image.py
@@ -100,8 +100,8 @@ def test_fits_image():
 
     assert_equal(fits_slc2["density"].data, fits_slc["density"].data)
     assert_equal(fits_slc2["temperature"].data, fits_slc["temperature"].data)
-    assert fits_slc.wcs.wcs.crval[0] == ds.domain_center[0].to_value("cm")
-    assert fits_slc.wcs.wcs.crval[1] == ds.domain_center[1].to_value("cm")
+    assert fits_slc.wcs.wcs.crval[0] == ds.domain_center[0].to_value("m")
+    assert fits_slc.wcs.wcs.crval[1] == ds.domain_center[1].to_value("m")
     assert fits_slc2.wcs.wcs.crval[0] == 0.0
     assert fits_slc2.wcs.wcs.crval[1] == 0.0
```

neutrinoceros (approved):

PR #4390, opened 2023-03-29
PERF: add explicit noexcept qualifiers to resolve perf regressions (Cython 3)

neutrinoceros:
## PR Summary

As tested on my fork, this resolves _most_ of the performance regressions seen in smoke tests with bleeding-edge CI (getting testing airtime from ~17min to ~8 min).

follow up to #4386

### DIFF
--- a/yt/frontends/artio/_artio_caller.pyx
+++ b/yt/frontends/artio/_artio_caller.pyx
@@ -1626,19 +1626,19 @@ cdef class SFCRangeSelector(SelectorObject):
     @cython.boundscheck(False)
     @cython.wraparound(False)
     @cython.cdivision(True)
-    cdef int select_sphere(self, np.float64_t pos[3], np.float64_t radius) nogil:
+    cdef int select_sphere(self, np.float64_t pos[3], np.float64_t radius) noexcept nogil:
         return 1
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
     @cython.cdivision(True)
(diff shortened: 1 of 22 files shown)


matthewturk:
Am I understanding that our global settings don't propagate to PXD files?

neutrinoceros:
which global settings ?

matthewturk:
Oh, I thought you tested it with setting the `noexcept` globally -- I thought there was a compiloer option to revert to previous settings.

neutrinoceros:
Disabling exception handling globally is only possible Cython 3, so we can't do that until we drop Cython 0.29. `noexcept` is backported as a noop to 0.29.x so it is the only option to transition smoothly. I don't want to add it blindly everywhere so instead I'm targeting functions where regressions are seen :)

matthewturk:
Ah, I understand.  I think for some reason I thought we gatekept the opt behind a version check.  But this is fine.

neutrinoceros:
pushed a bunch more `noexcept` for `SelectorObject.select_*` routines, which I found were responsible for regressions too.
I'm testing the current state of the branch on my fork at https://github.com/neutrinoceros/yt/actions/runs/4566963443

matthewturk (approved):

matthewturk:
I think one thing that's been slightly unexpected that we're learning here is that some of the `inline` functions also need the declarations, even though we want them inlined.

PR #4397, opened 2023-04-05
ENH: allow hint keyword for yt.load to select superclasses

yut23:
## PR Summary

This PR allows users to explicitly select a specific class in a `Dataset` hierarchy, even if one or more of its subclasses are valid. I'm currently trying to improve the boxlib detection code by adjusting the class hierarchy (as brought up in https://github.com/yt-project/yt/issues/3005#issuecomment-809182817), but making e.g. `NyxDataset` inherit from `AMReXDataset` would break cases like #3510.

## PR Checklist

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -116,8 +116,12 @@ def load(
         if cls._is_valid(fn, *args, **kwargs):
             candidates.append(cls)
 
+    # Filter the candidates if a hint was given
+    if hint is not None:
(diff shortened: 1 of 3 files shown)


neutrinoceros (approved):
Thank you for this patch ! It's absolutely in line with what `hint` is supposed to do, there are tests, and even background motivation echoing a comment I wrote a mere 2 years ago...
I'll give other maintainers a chance to have a look, but this looks perfect to me.

PR #4398, opened 2023-04-06
ENH: Enable querying FEM values from Python

matthewturk:
This is a very minor function wrapper to enable querying FEM values from within Python.


neutrinoceros:
I don't think I have necessary expertise to review this, but maybe it should be advertised to someone who does ?

matthewturk:
I think it should be sufficient to verify that my wrapper is a simple
wrapper, no? It's not actually adding any queries, just exposing them.

neutrinoceros:
Could you show an example of how it's used ? I don't know this module at all so it's hard for me to guess

matthewturk:
Hi!  Just as a note, I do want to return to this in the next 24 hours.  I tried to find my big example script but was unsuccessful.  Trying again.  :)

chrishavlin:
I can take a look at this!

chrishavlin (commented):
Generally looks great, found one line that might need adjusting. 

And in testing things out in the notebook [here](https://github.com/chrishavlin/yt_scratch/blob/master/notebooks/dev/mesh_stuff.ipynb), I ended up generally needing both  `map_reals_to_unit`  and `sample_at_real_points`. Since the sampling itself does not enforce element bounds, I needed `map_reals_to_unit` to build a mask so I could fill in empty values to the array I got back from `sample_at_real_points`.  And since `check_inside` isn't available at the python level, I had to go check the code to remind me what the proper element coordinate ranges were for the various elements... so I think it could be useful to have a way to call `check_inside` at the python level as well (that works with arrays the functions you've already added), or maybe an option for `sample_at_real_points` that would call `check_inside` before sampling and fill (with nan?) otherwise? what do you think?

matthewturk:
> otherwise? what do you think?

This sounds like a good idea.  I'll implement it!

yt/utilities/lib/element_mappings.pyx
```diff
@@ -115,6 +115,34 @@ cdef class ElementSampler:
...
+                          np.float64_t[:,::1] positions):
+        cdef double mapped_x[3]
+        cdef int i, n
+        # We have N vertices, which each have three components.
+        cdef np.ndarray[np.float64_t, ndim=2] output_coords
+        output_coords = np.zeros((positions.shape[0], positions.shape[1]), dtype="float64")
+        # Now for each position, we map
+        for n in range(positions.shape[0]):
+            self.map_real_to_unit(mapped_x, &vertices[0,0], &positions[n, 0])
+            for i in range(3):
```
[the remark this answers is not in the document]
matthewturk: I will check, but I think it's because the positions are expanded in dimensions.  But I'll check.

matthewturk:
@chrishavlin What do you think of this change?

yt/utilities/lib/element_mappings.pyx
```diff
@@ -115,6 +133,34 @@ cdef class ElementSampler:
...
+                          np.float64_t[:,::1] positions):
+        cdef double mapped_x[3]
+        cdef int i, n
+        # We have N vertices, which each have three components.
+        cdef np.ndarray[np.float64_t, ndim=2] output_coords
+        output_coords = np.zeros((positions.shape[0], positions.shape[1]), dtype="float64")
+        # Now for each position, we map
+        for n in range(positions.shape[0]):
+            self.map_real_to_unit(mapped_x, &vertices[0,0], &positions[n, 0])
+            for i in positions.shape[1]:
```
chrishavlin, line 150: Still need the `range` here: 
```suggestion
            for i in range(positions.shape[1]):
```

yt/utilities/lib/element_mappings.pyx
```diff
@@ -95,6 +95,24 @@ cdef class ElementSampler:
...
 
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    @cython.cdivision(True)
+    def check_contains(self, np.float64_t[:,::1] vertices,
+                             np.float64_t[:,::1] positions):
+        cdef np.ndarray[np.float64_t, ndim=2] mapped_coords
+        cdef np.ndarray[np.uint8_t, ndim=1] mask
+        mapped_coords = self.map_reals_to_unit(vertices, positions)
+        mask = np.zeros(mapped_coords.shape[0])
```
chrishavlin, line 106: I was getting a buffer type mismatch since np.zeros defaults to float but mask is declared as int, had to add:
```suggestion
        mask = np.zeros(mapped_coords.shape[0], dtype=np.uint8) 
```

chrishavlin (commented):
nice! the new function is great (after a small fix)!

chrishavlin (dismissed):

neutrinoceros:
@matthewturk is this still WIP ? I think it can go in as soon as the WIP check and limiting are green

matthewturk:
pre-commit.ci autofix

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros (approved):

PR #4399, opened 2023-04-06
BUG: Fix some formatting in docs

matthewturk:
In slack @chummels noticed this.  I think the blocks just leaked, and this pretty much reads the same way, so I changed it to be less prone to error.

### DIFF
--- a/doc/source/developing/creating_derived_fields.rst
+++ b/doc/source/developing/creating_derived_fields.rst
@@ -42,15 +42,15 @@ CGS: if a different unit system is selected, the result will be in the same
 dimensions of pressure but different units. See :ref:`units` for more
 information.
 
-Once we've defined our function, we need to notify yt that the field is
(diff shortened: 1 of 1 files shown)


neutrinoceros:
So for context, this is supposed to fix  #4403, but it currently doesn't:
- it's not just one page that's affected
- this patch doesn't actually change how the page is rendered, when compared with the previous artefacts from other recent PRs

matthewturk:
Hey I never said it would fix #4403!  You filed that bug *after* I filed this PR!

neutrinoceros:
Yes, but you're chasing the exact bug I reported, aren't you ?

matthewturk:
Well, yes, but ... :-)

neutrinoceros:
#4403 is now resolved, so this can be closed too, right ?

matthewturk:
Sounds good!

PR #4402, opened 2023-04-06
ENH: improve boxlib data type detection

yut23:
## PR Summary

Make `CastroDataset` and `MaestroDataset` inherit from `AMReXDataset` so they supercede it when calling yt.load. This allows loading output files from those projects without needing to pass `hint`.

I don't personally use Nyx, WarpX, or Orion, so any input from those projects would be appreciated.

## PR Checklist


yut23:
Some of the Castro (and Maestro) sample datasets are from before we migrated from BoxLib to AMReX, so the job_info files don't contain the string "amrex". That means `CastroDataset` and `BoxlibDataset` are valid candidates for yt.load, but `AMReXDataset` is not. This exposed a fun edge case in find_lowest_subclasses, where it would add `AMReXDataset` to the list of candidates.

yt/utilities/hierarchy_inspection.py
```diff
@@ -39,4 +39,4 @@ def find_lowest_subclasses(candidates: List[Type]) -> List[Type]:
 
     count = reduce(lambda x, y: x + y, counters)
 
-    return [x for x in count.keys() if count[x] == 1]
+    return [x for x in candidates if count[x] == 1]
```
neutrinoceros, line 42: while seemingly harmless, I'd rather keep this out of the PR, because it _might_ complicate backporting operations. Unless I am not understanding this correctly and it's not a pure noop ?

neutrinoceros (commented):
Thank you for tackling this long standing problem with the boxlib frontend !

This seems fine to me as a non-expert of this frontend (and the codes it relates too), but I'd very much like @zingale's opinion on this (or possibly recommendations for any other boxlib experts).

I think we should also clearly state how the inheritance hierarchy is designed: is it reflecting how the supported codes relate to each other ? is it a pure abstraction aiming for maximal code reuse ? is it purely historical ? anything else ?

zingale:
I will test this today (and I can test Nyx).  @ax3l can you test WarpX on this PR?

yt/utilities/hierarchy_inspection.py
yut23: It is not a noop; I'll pull it into a separate PR with more details today. I mostly wanted to make sure all the tests passed with both changes, since I couldn't get the answer tests to work locally.

zingale:
I've confirmed that this works with Nyx, Castro, and MAESTROeX without needing to use the hint keyword

yt/utilities/hierarchy_inspection.py
neutrinoceros: for posterity this is #4404 
Thanks @yut23 !

zingale (dismissed):

neutrinoceros:
@yut23 can you cleanup the branch history ? It looks like the first commit is only made of changes that were incorporated already

neutrinoceros (dismissed):

yt/frontends/boxlib/data_structures.py
```diff
@@ -922,6 +922,53 @@ def relative_refinement(self, l0, l1):
...
+            for ptype in self.ds.particle_types:
+                self._read_particles(ptype, is_checkpoint)
+
+
+class AMReXDataset(BoxlibDataset):
+    _index_class: Type[BoxlibHierarchy] = AMReXHierarchy
+    _subtype_keyword = "amrex"
+    _default_cparam_filename = "job_info"
+
+    def __init__(
```
neutrinoceros, line 940: I have doubts that this override is any use, as it does nothing besides calling the overridden function. Am I missing something ?

yut23: I'm pretty sure that's correct, as all the default parameters are the same. I'll remove it.

neutrinoceros (approved):

neutrinoceros:
I actually think the change is small enough that we don't need too much extra manual validation. I'll give @ax3l an other week or so to tune in if he wants but otherwise I think I'll just merge this.

PR #4404, opened 2023-04-07
BUG: make sure find_lowest_subclasses doesn't add extra classes

yut23:
## PR Summary

If a child class and its grandparent (or another ancestor) are the only classes in `candidates`, the previous implementation of `find_lowest_subclasses` would return both the child class and the parent class. This PR changes the logic so that it will only return classes passed in as candidates.

Needed by #4402.

```python
>>> import inspect
>>> from collections import Counter
>>> from functools import reduce

>>> class Grandparent:
...     pass
... class Parent(Grandparent):
...     pass
... class Child(Parent):
...     pass

>>> candidates = [Grandparent, Child]
>>> mros = [inspect.getmro(c) for c in candidates]; mros
[(__main__.Grandparent, object),
 (__main__.Child, __main__.Parent, __main__.Grandparent, object)]

>>> counters = [Counter(mro) for mro in mros]; counters
[Counter({__main__.Grandparent: 1,
          object: 1}),
 Counter({__main__.Child: 1,
          __main__.Parent: 1,
          __main__.Grandparent: 1,
          object: 1})]

# Parent only appeared once, so the original code will include it
>>> count = reduce(lambda x, y: x + y, counters); count
Counter({__main__.Grandparent: 2,
         object: 2,
         __main__.Child: 1,
         __main__.Parent: 1})

>>> [x for x in count.keys() if count[x] == 1]
[__main__.Child, __main__.Parent]

>>> [x for x in candidates if count[x] == 1]
[__main__.Child]
```

## PR Checklist

### DIFF
--- a/yt/utilities/hierarchy_inspection.py
+++ b/yt/utilities/hierarchy_inspection.py
@@ -39,4 +39,4 @@ def find_lowest_subclasses(candidates: List[Type]) -> List[Type]:
 
     count = reduce(lambda x, y: x + y, counters)
 
-    return [x for x in count.keys() if count[x] == 1]
+    return [x for x in candidates if count[x] == 1]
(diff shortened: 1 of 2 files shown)


neutrinoceros:
thank you for providing context. LGTM !

neutrinoceros (approved):

PR #4405, opened 2023-04-08
RFC: reduce code complexity in find_lowest_subclasses helper function

neutrinoceros:
## PR Summary
Reviewing #4404 made me read this function again with a fresh eye, and realized that it was more complex than needed (none of the special cases seem actually necessary).
Eventually it boils down to a two-liner, which to me feels easier to understand than the original implementation, but maybe I'm biased and I just wrote an unreadable and unmaintainable line ?

I'd like to have @yut23's opinion on this :-)

### DIFF
--- a/yt/utilities/hierarchy_inspection.py
+++ b/yt/utilities/hierarchy_inspection.py
@@ -1,7 +1,8 @@
 import inspect
 from collections import Counter
-from functools import reduce
 from typing import List, Type
+
+from more_itertools import flatten
 
 
 def find_lowest_subclasses(candidates: List[Type]) -> List[Type]:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
found a simpler approach using `more_itertools.flatten` instead of `itertools.chain`. Now it should definitely be clearer than the original implementation

neutrinoceros:
Turns out `more_itertools.flatten` is literally a wrapper/alias for `itertools.chain.from_iterable`. I'm choosing to use the former because I think it's easier to understand what it does, which is valuable in the context of a busy one-liner

yut23:
I came up with my own version of this last week (#4411). It wound up being pretty similar to this one, but it uses sets instead of Counters.

yut23 (approved):

neutrinoceros:
closing in favour of #4411

PR #4410, opened 2023-04-10
Export standard deviation to pandas DataFrames and AstroPy QTables

jzuhone:
## PR Summary

Profiles in yt can already export data to [pandas DataFrames](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html) and [AstroPy QTables](https://docs.astropy.org/en/stable/api/astropy.table.QTable.html).  This PR adds the capability to optionally export the standard deviation of the profile to the data frame or table. Docs and tests are added. 

## PR Checklist

### DIFF
--- a/yt/data_objects/profiles.py
+++ b/yt/data_objects/profiles.py
@@ -562,7 +562,7 @@ class Profile1D(ProfileND):
             fields = self.data_source._determine_fields(fields)
         return idxs, masked, fields
 
-    def to_dataframe(self, fields=None, only_used=False):
+    def to_dataframe(self, fields=None, only_used=False, include_std=False):
         r"""Export a profile object to a pandas DataFrame.
 
         This function will take a data object and construct from it and
@@ -577,8 +577,13 @@ class Profile1D(ProfileND):
             profile, along with the bin field, will be exported.
         only_used : boolean, default False
             If True, only the bins which have data will be exported. If False,
-            all of the bins will be exported, but the elements for those bins
+            all the bins will be exported, but the elements for those bins
             in the data arrays will be filled with NaNs.
+        include_std : boolean, optional
+            If True, include the standard deviotion of the profile
+            in the pandas DataFrame. It will appear in the table as the
+            field name with "_stddev" appended, e.g. "velocity_x_stddev".
+            Default: False
 
         Returns
         -------
@@ -600,6 +605,8 @@ class Profile1D(ProfileND):
         pdata = {self.x_field[-1]: self.x[idxs]}
         for field in fields:
             pdata[field[-1]] = self[field][idxs]
+            if include_std:
+                pdata[f"{field[-1]}_stddev"] = self.standard_deviation[field][idxs]
         df = pd.DataFrame(pdata)
         if masked:
             mask = np.zeros(df.shape, dtype="bool")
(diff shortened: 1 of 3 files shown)


yt/data_objects/tests/test_profiles.py
```diff
@@ -706,6 +706,15 @@ def test_export_astropy():
...
     assert_equal(prof.x.d[prof.used], at2["radius"].value)
     assert_equal(prof[("gas", "density")].d[prof.used], at2["density"].value)
+    at3 = prof.to_astropy_table(fields=("gas", "density"), include_std=True)
+    assert_equal(prof[("gas", "density")].d, at3["density"].value)
+    assert_equal(
+        prof.standard_deviation[("gas", "density")].d, at3["density_stddev"].value
+    )
+    assert (
+        prof.standard_deviation[("gas", "density")].units
+        == YTArray.from_astropy(at3["density_stddev"]).units
```
neutrinoceros, line 716: shouldn't the field type also be included in the key (`"gas_density_stddev"` instead of `"density_stddev"`) ?

jzuhone: @neutrinoceros so the field type is not included already even when we do not include the standard deviation (I don't remember why). So you would access the mean of the profile using `at3["density"]`

neutrinoceros: ok, at least it's consistent.

jzuhone: We could add the field type, but I'm a bit leery of using `_` to connect them since `_` could appear in field type strings (and does appear in field name strings), making parsing a bit difficult in some cases. 

Is there another sensible character to use to indicate the breaking of the field type and name? Maybe even `,` is ok? I don't know (cc @matthewturk)

neutrinoceros: I guess that would be a breaking change, so I'm actually -1 on this. Let's keep it as is for now.

yt/data_objects/profiles.py
```diff
@@ -577,8 +577,13 @@ def to_dataframe(self, fields=None, only_used=False):
             profile, along with the bin field, will be exported.
         only_used : boolean, default False
             If True, only the bins which have data will be exported. If False,
-            all of the bins will be exported, but the elements for those bins
+            all the bins will be exported, but the elements for those bins
             in the data arrays will be filled with NaNs.
+        include_std : boolean, optional
+            If True, include the standard deviotion of the profile
```
chrishavlin, line 583:
```suggestion
            If True, include the standard deviation of the profile
```

yt/data_objects/profiles.py
```diff
@@ -627,10 +634,15 @@ def to_astropy_table(self, fields=None, only_used=False):
             to the QTable as rows. If False, all bins are
             copied, but the bins which are not used are masked.
             Default: False
+        include_std : boolean, optional
+            If True, include the standard deviotion of the profile
```
chrishavlin, line 638:
```suggestion
            If True, include the standard deviation of the profile
```

chrishavlin (approved):
Looks good, caught a small typo in the new docstring entries

jzuhone:
Thanks @chrishavlin!

neutrinoceros (approved):

PR #4411, opened 2023-04-10
RFC: rewrite find_lowest_subclasses to better match its intent

yut23:
## PR Summary

This implementation is more straightforward than the original one with fewer edge cases that need special handling.
It ended up being quite similar to the final implementation in #4405, except it can handle having the same class show up in candidates more than once.
It simply makes a set of all the superclasses of the candidates, and filters out any candidates that appear in that set.

### DIFF
--- a/yt/utilities/hierarchy_inspection.py
+++ b/yt/utilities/hierarchy_inspection.py
@@ -1,7 +1,8 @@
 import inspect
 from collections import Counter
-from functools import reduce
 from typing import List, Type
+
+from more_itertools import flatten
 
 
 def find_lowest_subclasses(candidates: List[Type]) -> List[Type]:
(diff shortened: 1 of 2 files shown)


neutrinoceros:
> It ended up being quite similar to the final implementation in https://github.com/yt-project/yt/pull/4405, except it can handle having the same class show up in candidates more than once.

I'm not against dropping my PR in favour of this one in principle, but could explain why this extra bit of behaviour is needed or desired ?

yut23:
It's not needed for any existing code, but I think it's a bit more future-proof. I'm fine with either implementation (the way they work is identical except for that edge case), but the extra tests should be added either way.

neutrinoceros:
Actually I cannot find an implementation that _doesn't_ survive your new tests (both #4405 and the main branch seem to pass them). Since your original post suggests you're adding new behaviour, this may indicate that you're not actually testing what you aimed for.
These are super quick tests so I don't mind adding them to the code base even if they turn out to be redundant, but maybe you'd like to revisit them first ?

yut23:
`test_empty` was an error I ran into while writing this: the full unit tests caught it, but I figured it's good to explicitly test it here; `test_without_grandparents` is just another possible edge case that was resolved by #4397.

I could add a test for a class being in `candidates` multiple times, but I didn't even consider it as a feature when I was writing the code. I only noticed that when I compared it to your implementation.

neutrinoceros:
I think it's safe that this function assumes unicity of candidates, though maybe it should be mentioned in the docstring.

neutrinoceros:
@yut23 I think we could merge your tests and my implementation. Would that work for you ?

yut23:
Sounds good to me!

neutrinoceros:
Then could you include my implementation here ? This seems like the easiest way to avoid conflicts.
What else is keeping this from leaving draft mode ?

neutrinoceros (approved):

PR #4414, opened 2023-04-12
gadget hdf: reduce size 1 header arrays to scalar

chrishavlin:
This PR is one way to fix an issue reported on slack by Ben Oppenheimer: the CAMELS Astrid (Gadget-3, HDF) simulations have some hdf5 attributes saved as single element arrays but yt expects scalars, which leads to errors on dataset initialization. 

The fix here simply reduces length 1 arrays to scalar values in the base gadget class. 

Note that this particular flavor of HDF5 Gadget output does not have a dedicated frontend in yt. Assuming all the yt tests pass  (Ben confirmed it does indeed allow loading of data and outputs from analysis seem correct), the fix here is simple enough that it seemed worth submitting even though a new frontend is a better long term solution. I also don't have a good datafile for testing, so just mocked up a header with nested arrays in the new test that I added.


chrishavlin:
oops, look like I didn't properly exclude the new test from nose...

nose_ignores.txt
```diff
@@ -32,4 +32,5 @@
 --ignore-file=test_stream_particles\.py
 --ignore-file=test_stream_stretched\.py
 --ignore-file=test_version\.py
+--ignore-file=test_gadget_pytests\.py
```
neutrinoceros, line 35: I guess the file should be renamed (`pytests` -> `pytest`)

yt/frontends/gadget/data_structures.py
```diff
@@ -629,6 +629,15 @@ def _get_hvals(self):
         if "Parameters" in handle:
             hvals.update((str(k), v) for k, v in handle["/Parameters"].attrs.items())
         handle.close()
+
+        # ensure that 1-element arrays are reduced to scalars
+        updated_hvals = {}
+        for hvalname, value in hvals.items():
+            if isinstance(value, np.ndarray) and value.size == 1:
+                mylog.info(f"Reducing singleton array {hvalname} to scalar.")
```
neutrinoceros, line 637:
```suggestion
                mylog.info("Reducing single-element array %s to scalar.", hvalname)
```

neutrinoceros: This is a violation of rule `G004`, `ruff` is supposed to flag it (I guess that's a bug that it doesn't)

nose_ignores.txt
chrishavlin: because `test_gadget_pytests.py` only includes a single test? Or did you have another reason?

neutrinoceros: that's just because `pytest` is spelled without an trailing `s`. There's also a convention that some "pytest-only" test modules are named `*_pytest.py` (usually the suffix is only used to avoid name collisions with other modules)

chrishavlin: ah, got it. happy to change it!

neutrinoceros (approved):
other than a stylistic nit I should have caught earlier, LGTM !

jzuhone (dismissed):

neutrinoceros (approved):

neutrinoceros:
let's try backporting this

Issue #4350, opened 2023-02-22
BUG: axes labelling in spectral cube geometry


neutrinoceros:
### Bug report

**Bug summary**

While working on #4179, I noticed that when plotting a SlicePlot of a spectral cube, with `z` as one of the plot axes, no unit is shown for this axis (I think that's expected ?), but you still get an empty pair of parentheses, and the the label ends up as `"z ()"`

**Code for reproduction**

```python
import yt
from yt.testing import fake_amr_ds

ds = fake_amr_ds(fields=[("gas", "density")], units=["g/cm**3"], geometry="spectral_cube")
yt.SlicePlot(ds, "x", ("gas", "density"), buff_size=(8, 8)).save("/tmp/")
```

**Actual outcome**

![AMRGridData_Slice_x_density](https://user-images.githubusercontent.com/14075922/220617352-0b0a3752-a9b0-41cf-9fbd-e82c3d5e7558.png)

This should be relatively easy to fix in the context of #4179, but I'd like to know what's the expected behaviour ? Should there be a unit there (if so, which one ?), or should we simply get rid of the parens, as we normally do for dimensionless axes ?

I assume @jzuhone would know what's best here.

neutrinoceros:
Assuming no unit should be shown, #4179 now resolves this.

neutrinoceros:
@jzuhone can you weigh in ?

jzuhone:
So this is a contrived example--if you're plotting with the spectral axis as one of the plot axes, then the units should be _something_ if you are working with "real" spectral cubes. But `fake_amr_ds` is faking a `FITSDataset` here: 

https://github.com/yt-project/yt/blob/d73939ec7f13d9028cad8603f6d8302a5a61a8d8/yt/frontends/stream/data_structures.py#L345-L364

and not giving the `z` axis a unit. Normally it's going to be velocity, energy, wavelength, or frequency. 

But I guess that for this case the blank label is correct and we should not have any units (and no parentheses) or it should say (dimensionless). I have no preference.

neutrinoceros:
Thank you for clarifying ! Then it's not so much an issue with the spectral cube geometry, but it's still a real issue with how empty strings are transcribed as labels. #4417 fixes it without touching geometry handlers

PR #4417, opened 2023-04-16
BUG: fix inconsistent units labelling between plot axis and colorbar

### LINKED ISSUE #4416: BUG: axes and colorbar units labelling is inconsistent
### Bug report

**Bug summary**

The way units are parenthesised is inconsistent throughout the code base, mostly because code handling this operation is copy-pasted with small mutations here and there.
Sometimes it can be seen within a single plot, where the colorbar uses the same units as axes, but parentheses are visibly different.

**Code for reproduction**

```python
import yt

ds = fake_amr_ds(
    [
        ("gas", "velocity_x"),
        ("gas", "velocity_y"),
        ("gas", "velocity_z"),
        ("gas", "sound_speed"),
    ],
    units=["m/s"] * 4,
)
p = yt.ProjectionPlot(ds, "z", "mach_number")
p.save("/tmp/test.png")
```

**Actual outcome**
test

axes labels use `r"(\rm{cm})"` while the colorbar label uses  `r"\left(\rm{cm}\right)"`, which is generally better (though not in this case)

**Expected outcome**

I think units should look the same, with no regards for where they are in the plot, or what the plot class is. This is an opportunity to implement the correct behaviour once and reuse it throughout the code base.

**Version Information**
yt `4.1.3` + matplotlib `3.7.1`

neutrinoceros:
## PR Summary

closes #4416
follow up to #3282 and #3637

### DIFF
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -102,7 +102,7 @@ answer_tests:
     - yt/frontends/owls/tests/test_outputs.py:test_snapshot_033
     - yt/frontends/owls/tests/test_outputs.py:test_OWLS_particlefilter
 
-  local_pw_047:  #  PR 4132
+  local_pw_048:  # PR 4417
     - yt/visualization/tests/test_plotwindow.py:test_attributes
     - yt/visualization/tests/test_particle_plot.py:test_particle_projection_answers
     - yt/visualization/tests/test_particle_plot.py:test_particle_projection_filter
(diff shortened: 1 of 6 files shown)


neutrinoceros:
Failed image tests on Jenkins are all expected and just reflect the intended change.

neutrinoceros:
companion PR: https://github.com/yt-project/yt_pytest_mpl_baseline/pull/2

neutrinoceros:
I could probably get rid of #4350 while I'm in there, switching to draft while I try it out !

neutrinoceros:
I think @chummels or @jzuhone would be good reviewers here

chummels:
Sorry, @neutrinoceros .  I just saw that you tagged me on this as I was randomly going through github messages from the last couple of weeks.  Thanks for doing this, and hopefully everything looks good now with units!

PR #4420, opened 2023-04-20
Added more variable metadata for CM1 / NetCDF4 file format

leighorf:
## PR Summary

Added updated prognostic and derived fields for cm1 NetCDF front end.

## PR Checklist


matthewturk:
It's so funny to me that PR 4420 is our first pull request "for" April 20.

matthewturk:
@leighorf Thank you!  I think there are some quick/easy structural changes to make here, with the order of the not-terribly-intuitive tuple components, which I'll suggest as changes inline.

yt/frontends/nc4_cm1/fields.py
```diff
@@ -9,42 +9,69 @@ class CM1FieldInfo(FieldInfoContainer):
...
-        ("hvort", ("1/s", ["horizontal_vorticity_magnitude"], None)),
-        ("vortmag", ("1/s", ["vorticity_magnitude"], None)),
-        ("streamvort", ("1/s", ["streamwise_vorticity"], None)),
-        ("khh", ("m**2/s", ["khh"], None)),
-        ("khv", ("m**2/s", ["khv"], None)),
-        ("kmh", ("m**2/s", ["kmh"], None)),
-        ("kmv", ("m**2/s", ["kmv"], None)),
+        ("uinterp", ("m/s", ["x velocity"], "uinterp")),
+        ("vinterp", ("m/s", ["y velocity"], "vinterp")),
+        ("winterp", ("m/s", ["z velocity"], "winterp")),
```
matthewturk, line 14:
```suggestion
        ("uinterp", ("m/s", ["velocity_x"], "x velocity")),
        ("vinterp", ("m/s", ["velocity_y"], "y velocity")),
        ("winterp", ("m/s", [velocity_z"], "z velocity")),
```

The order of the tuples is a bit funny, so I'm suggesting the changes here.  What we need:

`("field_name_on_disk"), ("units", ["field_name_to_alias_to", "another_field_to_alias_to"], "display name")`

The tricky bit is that the field name to alias to has to match the internal yt names (and we try to make them tokenizable) -- so in this case we need to stick with the `velocity_x` etc.  But for the display name, that's what we can see in the data labels.  So I've put the aliased names there, but I'm not sure that's what you want.

yt/frontends/nc4_cm1/fields.py
```diff
@@ -9,42 +9,69 @@ class CM1FieldInfo(FieldInfoContainer):
...
+        ("u", ("m/s", ["x velocity"], "u")),
+        ("v", ("m/s", ["y velocity"], "v")),
+        ("w", ("m/s", ["z velocity"], "w")),
```
matthewturk, line 17: I'll make a suggestion here too for changing it, but before doing so, do we want `u` and `uinterp` to be displayed differently?

matthewturk (commented):
I left a couple questions, then we can keep moving forward with it

yt/frontends/nc4_cm1/fields.py
matthewturk: Also, I messed up the `velocity_z` here and so my suggestion won't quite work

PR #4424, opened 2023-04-28
[ENH] support for sinks in csv format for RAMSES frontend

Lenoble-lab:
Hi everyone,
this pull request is to implement a reader to read the output from ramses sink particle algorithm. 
For now, the sink ouput from the standard ramses (the one from the bitbucket) cannot be read. It's only an other format.

In each output_*****, the sink algo write a single sink_****.csv file with the properties of the sinks. I have added an example in attached file. 
We need to read a single file per output folder in the csv format. 

I think that the ideal would be to have a single ptype corresponding to the sinks (do not create a 'sink_csv' type). The read_header would have to be modified and specified as appropriate. For now, I have impleneted a new name 'sink_csv' to differentiate from the first class. 

The main issue I had is the ` IOHandlerRAMSES._read_particle_fields ` which reads in every chunks concerned, whereas here there is only one single file to read. I hacked this with the if ptype == 'sink_csv' line 217 in io.py but this is really not ideal. 

Thinking about it, I see several solutions: 
- add a flag in the sink class to specify that it should only be read once -> add an if in IOHandlerRAMSES._read_particle_fields to say that if this is the case, only one read is enough. And we avoid entering the for chunk in chunks loop. 

- we create a new IOHandlerRAMSES._read_sink_fields which reads only once the csv file, directly in IOHandlerRAMSES. It is possible to add a field type that would not be separated into cpu (I don't know if other Ramses features also write a single file per output). But then you lose the inheritance of the particle class (or re-create another one?).
[sink_00010.csv](https://github.com/yt-project/yt/files/11353508/sink_00010.csv)

Thanks in advance !
Romain Lenoble

### DIFF
--- a/yt/frontends/ramses/fields.py
+++ b/yt/frontends/ramses/fields.py
@@ -131,6 +131,23 @@ class RAMSESFieldInfo(FieldInfoContainer):
         ("particle_metallicity", ("", [], None)),
         ("particle_family", ("", [], None)),
         ("particle_tag", ("", [], None)),
+        # sink field parameters
+        ("particle_mass", ("code_mass", [], None)),
+        ("particle_angular_momentum_x", (ang_mom_units, [], None)),
+        ("particle_angular_momentum_y", (ang_mom_units, [], None)),
+        ("particle_angular_momentum_z", (ang_mom_units, [], None)),
+        ("particle_formation_time", ("code_time", [], None)),
+        ("particle_accretion_Rate", ("code_mass/code_time", [], None)),
+        ("particle_delta_mass", ("code_mass", [], None)),
+        ("particle_rho_gas", (rho_units, [], None)),
+        ("particle_cs**2", (vel_units, [], None)),
+        ("particle_etherm", (ener_units, [], None)),
+        ("particle_velocity_x_gas", (vel_units, [], None)),
+        ("particle_velocity_y_gas", (vel_units, [], None)),
+        ("particle_velocity_z_gas", (vel_units, [], None)),
+        ("particle_mass_bh", ("code_mass", [], None)),
+        ("particle_level", ("", [], None)),
+        ("particle_radius_star", ("code_length", [], None)),
     )
 
     known_sink_fields: KnownFieldsT = (
--- a/yt/frontends/ramses/io.py
+++ b/yt/frontends/ramses/io.py
@@ -362,34 +362,31 @@ def _read_part_csv_file_descriptor(fname: Union[str, "os.PathLike[str]"]):
     from yt.utilities.on_demand_imports import _pandas as pd
 
     # Fields name from the default csv RAMSES sink algorithm in the yt default convention
-    mapping_list = [
-        (" # id", "particle_identifier"),
-        ("msink", "particle_mass"),
-        ("x", "particle_position_x"),
-        ("y", "particle_position_y"),
-        ("z", "particle_position_z"),
-        ("vx", "particle_velocity_x"),
-        ("vy", "particle_velocity_y"),
-        ("vz", "particle_velocity_z"),
-        ("lx", "particle_angular_momentum_x"),
-        ("ly", "particle_angular_momentum_y"),
-        ("lz", "particle_angular_momentum_z"),
-        ("tform", "particle_formation_time"),
-        ("acc_Rate", "particle_accretion_Rate"),
-        ("del_mass", "particle_delta_mass"),
-        ("rho_gas", "particle_rho_gas"),
-        ("cs**2", "particle_sound_speed"),
-        ("etherm", "particle_etherm"),
-        ("vx_gas", "particle_velocity_x_gas"),
-        ("vy_gas", "particle_velocity_y_gas"),
-        ("vz_gas", "particle_velocity_z_gas"),
-        ("mbh", "particle_mass_bh"),
-        ("level", "particle_level"),
-        ("rsink_star", "particle_radius_star"),
-    ]
-
-    # Convert to dictionary
-    mapping = dict(mapping_list)
+    mapping = {
+        " # id": "particle_identifier",
+        "msink": "particle_mass",
+        "x": "particle_position_x",
+        "y": "particle_position_y",
+        "z": "particle_position_z",
+        "vx": "particle_velocity_x",
+        "vy": "particle_velocity_y",
+        "vz": "particle_velocity_z",
+        "lx": "particle_angular_momentum_x",
+        "ly": "particle_angular_momentum_y",
+        "lz": "particle_angular_momentum_z",
+        "tform": "particle_formation_time",
+        "acc_Rate": "particle_accretion_Rate",
+        "del_mass": "particle_delta_mass",
+        "rho_gas": "particle_rho_gas",
+        "cs**2": "particle_sound_speed",
+        "etherm": "particle_etherm",
+        "vx_gas": "particle_velocity_x_gas",
+        "vy_gas": "particle_velocity_y_gas",
+        "vz_gas": "particle_velocity_z_gas",
+        "mbh": "particle_mass_bh",
+        "level": "particle_level",
+        "rsink_star": "particle_radius_star",
+    }
 
     dat = pd.read_csv(
         fname, delimiter=","
--- a/yt/frontends/ramses/io.py
+++ b/yt/frontends/ramses/io.py
@@ -122,7 +122,7 @@ def _ramses_particle_csv_file_handler(particle, subset, fields, count):
 
     Parameters
     ----------
-    particle : ``ParticleFileHandler``
+    particle: ``ParticleFileHandler``
         the particle class we want to read
     subset: ``RAMSESDomainSubset``
         A RAMSES domain subset object
--- a/yt/frontends/ramses/io.py
+++ b/yt/frontends/ramses/io.py
@@ -238,7 +238,19 @@ class IOHandlerRAMSES(BaseIOHandler):
                 if pn % ax not in field_list:
                     fields.append((ptype, pn % ax))
 
-            if ptype != "sink_csv":
+            if ptype == "sink_csv":
+                subset = chunks[0].objs[0]
+                rv = self._read_particle_subset(subset, fields)
+                for ptype, field_list in sorted(ptf.items()):
+                    x, y, z = (np.asarray(rv[ptype, pn % ax], "=f8") for ax in "xyz")
+                    mask = selector.select_points(x, y, z, 0.0)
+                    if mask is None:
+                        mask = []
+                    for field in field_list:
+                        data = np.asarray(rv.pop((ptype, field))[mask], "=f8")
+                        yield (ptype, field), data
+
+            else:
                 for chunk in chunks:
                     for subset in chunk.objs:
                         rv = self._read_particle_subset(subset, fields)
@@ -252,18 +264,6 @@ class IOHandlerRAMSES(BaseIOHandler):
                             for field in field_list:
                                 data = np.asarray(rv.pop((ptype, field))[mask], "=f8")
                                 yield (ptype, field), data
-
-            else:
-                subset = chunks[0].objs[0]
-                rv = self._read_particle_subset(subset, fields)
-                for ptype, field_list in sorted(ptf.items()):
-                    x, y, z = (np.asarray(rv[ptype, pn % ax], "=f8") for ax in "xyz")
-                    mask = selector.select_points(x, y, z, 0.0)
-                    if mask is None:
-                        mask = []
-                    for field in field_list:
-                        data = np.asarray(rv.pop((ptype, field))[mask], "=f8")
-                        yield (ptype, field), data
 
     def _read_particle_subset(self, subset, fields):
         """Read the particle files."""
--- a/yt/frontends/ramses/io.py
+++ b/yt/frontends/ramses/io.py
@@ -147,7 +147,7 @@ def _ramses_particle_csv_file_handler(particle, subset, fields, count):
     dat = pd.read_csv(
         fname,
         delimiter=",",
-        usecols=[ind[1] for ind in list_field_ind],
+        usecols=[ind for _field, ind in list_field_ind],
         skiprows=2,
         header=None,
     )
--- a/yt/frontends/ramses/io.py
+++ b/yt/frontends/ramses/io.py
@@ -72,17 +72,13 @@ def convert_ramses_conformal_time_to_physical_age(
     return (tsim - t) * t_scale
 
 
-def _ramses_particle_file_handler(fname, foffsets, data_types, subset, fields, count):
-    """General file handler, called by _read_particle_subset
+def _ramses_particle_binary_file_handler(particle, subset, fields, count):
+    """General file handler for binary file, called by _read_particle_subset
 
     Parameters
     ----------
-    fname : string
-        filename to read from
-    foffsets: dict
-        Offsets in file of the fields
-    data_types: dict
-        Data type of the fields
+    particle : ``ParticleFileHandler``
+        the particle class we want to read
     subset: ``RAMSESDomainSubset``
         A RAMSES domain subset object
     fields: list of tuple
@@ -93,6 +89,9 @@ def _ramses_particle_file_handler(fname, foffsets, data_types, subset, fields, c
     tr = {}
     ds = subset.domain.ds
     current_time = ds.current_time.in_units("code_time").v
+    foffsets = particle.field_offsets
+    fname = particle.fname
+    data_types = particle.field_types
     with FortranFile(fname) as fd:
         # We do *all* conversion into boxlen here.
         # This means that no other conversions need to be applied to convert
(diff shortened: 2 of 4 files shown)


yt/frontends/ramses/fields.py
```diff
@@ -131,6 +131,22 @@ class RAMSESFieldInfo(FieldInfoContainer):
         ("particle_metallicity", ("", [], None)),
         ("particle_family", ("", [], None)),
         ("particle_tag", ("", [], None)),
+        # sink field parameters
+        ("particle_msink",     ("code_mass", [], None)),
```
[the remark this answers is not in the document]
Lenoble-lab: Ok I see. Here, I use the format : `"particle_" + name of the header column in the csv file.`

The column names are : `msink`, `x`, `y`, `z`, `vx`,..., `lx`,...
Maybe using the default convention of yt is better

yt/frontends/ramses/io.py
```diff
@@ -92,6 +92,27 @@ def _ramses_particle_file_handler(fname, foffsets, data_types, subset, fields, c
     tr = {}
     ds = subset.domain.ds
     current_time = ds.current_time.in_units("code_time").v
+    #test if sink csv file -> not same reader as the fortran classic files
```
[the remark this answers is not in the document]
Lenoble-lab: Ok that is probably more clear

yt/frontends/ramses/io.py
```diff
@@ -192,6 +213,9 @@ def _read_particle_fields(self, chunks, ptf, selector):
             for ax in "xyz":
                 if pn % ax not in field_list:
                     fields.append((ptype, pn % ax))
+            # attention : read only one chunk as its a global file, don't know wether this is the good way to do it
```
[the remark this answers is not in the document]
Lenoble-lab: ah yes, that would definitely solve our problem and it can be much more easy to later implement reader for any sort of file

neutrinoceros:
@Lenoble-lab when you are done pushing new changes, please take a minute to address linting error reported by pre-commit.ci. All of them seem to be auto-fixable (see the note at the bottom of this page https://results.pre-commit.ci/run/github/88776493/1683098596.TWKhnqatSLSZQeSp2akaZw)

yt/frontends/ramses/io.py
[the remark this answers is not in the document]
Lenoble-lab: I can try to implement this if you want

yt/frontends/ramses/fields.py
```diff
@@ -95,82 +95,82 @@
...
-        ("y-acceleration", (ra_units, ["acceleration_y"], None)),
-        ("z-acceleration", (ra_units, ["acceleration_z"], None)),
-        ("Potential", (specific_ener_units, ["potential"], None)),
-        ("B_x_left", (b_units, ["magnetic_field_x_left"], None)),
-        ("B_x_right", (b_units, ["magnetic_field_x_right"], None)),
-        ("B_y_left", (b_units, ["magnetic_field_y_left"], None)),
-        ("B_y_right", (b_units, ["magnetic_field_y_right"], None)),
-        ("B_z_left", (b_units, ["magnetic_field_z_left"], None)),
-        ("B_z_right", (b_units, ["magnetic_field_z_right"], None)),
+        ("Density",                       (rho_units, ["density"], None)),
```
neutrinoceros, line 98: FYI we have automatic-formatting tools enable that will actively undo this kind of manual formatting.

yt/frontends/ramses/io.py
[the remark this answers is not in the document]
Lenoble-lab: But actually, if you keep something like this : 

```
           tr.update(
                ph.reader(
                    fname, foffsets, data_types, subset, subs_fields, count=count
                )
            )
```

You are still updated the value of tr by appending the variable at the end, iterating over the chunks concerned. 

To keep using this method,  you need to check that the particle you are adding are well in the subset when you read it in the csv file. Thus checking with a few `if` the position of the particle. I don't know if it is fast (but here the files are less than 1e3 sinks so not too fast).

A 2nd method would be to read once for all the data of the csv file, in this case, you need to avoid to go in the iteration line 240 in `_read_particle_fields` : 
```
 for chunk in chunks:
            for subset in chunk.objs:
```

In this case, the class would store all the sink produced, independently of their position. 
It requires to either write some condition in  `_read_particle_fields` before going onto the iteration. 
Maybe, we can implement this with `_read_particle_data_file` method ?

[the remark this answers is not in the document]
cphyc: It might be difficult to check whether a particle is in the current subset though.

I think we have two options here that avoid having code specific to `sinks` here:
- we update `_read_particle_fields` so that it returns a boolean that informs whether all the data has been read (`sink` case) or not (all other cases).
- we update the particle handlers so that they carry a flag that says "all data is in one file, please only read one chunk".

At the end of the day, the reason for this (slight) mess is that the index assumes that the data is spread over `Ncpu` different domains, each of which forms one chunk. The best solution would require having a second index I guess, but that would be a massive change, especially for a simple case like this one.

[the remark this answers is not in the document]
Lenoble-lab: - we update `_read_particle_fields` so that it returns a boolean that informs whether all the data has been read (sink case) or not (all other cases).
    - in that case, we need to modify deeply the class `BaseIOHandler`, as we modify the output of `_read_particle_fields` no ?
    
- we update the particle handlers so that they carry a flag that says "all data is in one file, please only read one chunk".
    - yes, that could be the simplest idea no ?

what I wanted to say is that in the `BaseIOHandler` there is a `_read_particle_data_file`. I don't get it but there is a `raise NotImplementedError`, whereas it is not implemented in the `IOHandlerRAMSES` class. Maybe we can implement a method for data_file like this. So when reading the csv data, the called function is this one and not the former `_read_particle_fields`. The ideal, would be a function setted as an attribute of the particle class, but that might require to modify the class `BaseIOHandler` I think.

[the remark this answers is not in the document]
cphyc: You may be right. I don't remember the inheritance of `BaseIOHandler`. In that case, updating the particle handlers (they're specific to RAMSES and already encode all the specifics of each particle output file anyway) seems the best route.

neutrinoceros:
Let's try to get this across the finish line. The priority would be to fix failing checks.
- [ ] runtime test failure: https://tests.yt-project.org/job/yt_py38_git/7104/testReport/junit/yt.frontends.ramses.tests/test_outputs/test_ramses_part_count/
- [ ] style checks https://results.pre-commit.ci/run/github/88776493/1689259288.1fuat6kqReilI3lnYx0SMw

yt/frontends/ramses/io.py
```diff
@@ -395,9 +395,10 @@ def _read_part_csv_file_descriptor(fname: Union[str, "os.PathLike[str]"]):
     # Convert to dictionary
     mapping = {k: v for k, v in mapping_list}
 
-    dat = pd.read_csv(fname, delimiter=',', nrows=1) #read only the header
+    dat = pd.read_csv(fname, delimiter=',') #read only the header
```
neutrinoceros, line 396: feels like the comment isn't relevant any more

Lenoble-lab:
Hi, I don't get the errors from https://results.pre-commit.ci/run/github/88776493/1690275703.l2NvkRYcQf62QbImBS2O4g

Do you know what are the problems ?

yt/frontends/ramses/io.py
```diff
@@ -393,7 +393,7 @@ def _read_part_csv_file_descriptor(fname: Union[str, "os.PathLike[str]"]):
     # Convert to dictionary
     mapping = {k: v for k, v in mapping_list}
 
-    dat = pd.read_csv(fname, delimiter=',') #read only the header
+    dat = pd.read_csv(fname, delimiter=',') #read only the all file to get the number of particle
```
neutrinoceros, line 396:
```suggestion
    dat = pd.read_csv(fname, delimiter=',') #read the all file to get the number of particle
```

.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ yt/extern\
 )"
 
 ci:
-    autofix_prs: false
+    autofix_prs: true
```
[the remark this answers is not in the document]
Lenoble-lab: Yes sorry, I change and I run pre-commit locally

neutrinoceros:
There's a merge conflict to resolve now. Would you allow me to rebase your branch or would you prefer to resolve it yourself ?

Lenoble-lab:
hum, as you want I was eating so I didn't see it. Now, I don't have time but later in the afternoon I can have a deeper look. I run 'pre-commit run --all-files` but it changes deeply some files

neutrinoceros:
fixed. I condensed the branch's history to eliminate back-and-forth edits which make stuff like `git bisect` and `git blame` less useful for future maintenance.

Here are the commands you need to run to update your local clone of the repo

```shell
git checkout main
git branch -D sink_csv
git fetch
git checkout sink_csv
```

yt/frontends/ramses/io.py
```diff
@@ -189,12 +232,29 @@ def _read_particle_fields(self, chunks, ptf, selector):
...
         ]
+
         for ptype, field_list in sorted(ptf.items()):
             for ax in "xyz":
                 if pn % ax not in field_list:
                     fields.append((ptype, pn % ax))
-        for chunk in chunks:
-            for subset in chunk.objs:
+
+            if ptype != "sink_csv":
```
neutrinoceros, line 241: I'm confused. Shouldn't the guard clause be `if ptype == "sink_csv":` here ?

yt/frontends/ramses/io.py
```diff
@@ -300,6 +355,59 @@ def _read_part_file_descriptor(fname: Union[str, "os.PathLike[str]"]):
...
+        ("vx_gas", "particle_velocity_x_gas"),
+        ("vy_gas", "particle_velocity_y_gas"),
+        ("vz_gas", "particle_velocity_z_gas"),
+        ("mbh", "particle_mass_bh"),
+        ("level", "particle_level"),
+        ("rsink_star", "particle_radius_star"),
+    ]
+
+    # Convert to dictionary
+    mapping = dict(mapping_list)
```
neutrinoceros, line 392: if we're going to immediately convert to dict and never use the `list` form, I suggest doing it all over "one" line (i.e., `mapping = dict([ ... ])`)

yt/frontends/ramses/io.py
```diff
@@ -300,6 +355,59 @@ def _read_part_file_descriptor(fname: Union[str, "os.PathLike[str]"]):
...
+
+    dat = pd.read_csv(
+        fname, delimiter=","
+    )  # read the all file to get the number of particle
```
neutrinoceros, line 396:
```suggestion
    # read the all file to get the number of particle
    dat = pd.read_csv(fname, delimiter=",")
```

neutrinoceros (commented):
here are a couple of light weighted questions and suggestions to improve clarity and readability

yt/frontends/ramses/io.py
Lenoble-lab: There is an if and then an else, so it's a bit equivalent, isn't it  ?

I was to force to add this condition as the reader is really not the same between the fortran file and the csv file. I didn't manage to find a way without.

yt/frontends/ramses/io.py
Lenoble-lab: yes you are right !

yt/frontends/ramses/io.py
```diff
@@ -118,6 +117,50 @@ def _ramses_particle_file_handler(fname, foffsets, data_types, subset, fields, c
...
+    ds = subset.domain.ds
+    current_time = ds.current_time.in_units("code_time").v
+    foffsets = particle.field_offsets
+    fname = particle.fname
+
+    for field in sorted(fields, key=lambda a: foffsets[a]):
+        ind = foffsets[field]
+        dat = pd.read_csv(
+            fname, delimiter=",", usecols=[ind], skiprows=2, header=None
+        )  # read only selected fields
```
neutrinoceros, line 146: I'm wondering if we could call `pd.read_csv` just once ahead of the loop instead of doing multiple (costly) reading operations ? If there was a chance of memory errors, I would expect the current implementation to hit those any way.

yt/frontends/ramses/io.py
neutrinoceros: Indeed there's an `else` in there, and having this `if/else` branching isn't an issue, but I would expect the *old* code to correspond to previously supported cases (i.e. `if ptype != "sink_csv"`). Here, it seems to me that the *old* code is run specifically in the `else` clause, which matches the case you are adding. Am I missing anything ?

yt/frontends/ramses/particle_handlers.py
```diff
@@ -282,3 +289,36 @@ def read_header(self):
...
+
+        for ind, field in enumerate(fields):
+            field_offsets[self.ptype, field] = ind
+            _pfields[self.ptype, field] = "d"
+
+        self.field_offsets = field_offsets
+        self.field_types = _pfields
+
+    def sec_read(self):
+        pass
```
neutrinoceros, line 324: I think this got simplified into oblivion by formatters, and it didn't break anything because this method isn't actually used, so might as well drop it ?
```suggestion
```

yt/frontends/ramses/particle_handlers.py
```diff
@@ -282,3 +289,36 @@ def read_header(self):
...
+class SinkParticleFileHandlerCsv(ParticleFileHandler):
+    """Handle sink files from a csv file, the format from the sink particle in ramses"""
+
+    ptype = "sink_csv"
+    fname = "sink_{iout:05d}.csv"
+    file_descriptor = None
+    config_field = "ramses-sink-particles"
+    reader = _ramses_particle_csv_file_handler
+    attrs = (("nsink", 1, "i"), ("nindsink", 1, "i"))
+    second_read = False
```
neutrinoceros, line 303: this doesn't seem to be used ?
```suggestion
```

Lenoble-lab: yes, I'm not using it as we can read directly the all file

yt/frontends/ramses/io.py
Lenoble-lab: ok I see. I will change with this idea

yt/frontends/ramses/io.py
Lenoble-lab: ok I change now. 
But, it's far from being optimized : I nead to read the entire file in `_read_part_csv_file_descriptor` to get the number of lines (to then initialize `self.local_particle_count`). And then I read it again (but only the columns i'm interested in)

Lenoble-lab: I think reading is a csv file is to different from readinf the fortan binay file, so thying to use the same structure is not ideal

yt/frontends/ramses/io.py
```diff
@@ -118,6 +117,57 @@ def _ramses_particle_file_handler(fname, foffsets, data_types, subset, fields, c
...
+    from yt.utilities.on_demand_imports import _pandas as pd
+
+    tr = {}
+    ds = subset.domain.ds
+    current_time = ds.current_time.in_units("code_time").v
+    foffsets = particle.field_offsets
+    fname = particle.fname
+
+    list_field_ind = [
+        [field, foffsets[field]] for field in sorted(fields, key=lambda a: foffsets[a])
```
neutrinoceros, line 143: functionally equivalent, but slightly more idiomatic
```suggestion
        (field, foffsets[field]) for field in sorted(fields, key=lambda a: foffsets[a])
```

yt/frontends/ramses/io.py
```diff
@@ -118,6 +117,57 @@ def _ramses_particle_file_handler(fname, foffsets, data_types, subset, fields, c
...
+    ]
+
+    # read only selected fields
+    dat = pd.read_csv(
+        fname,
+        delimiter=",",
+        usecols=[ind[1] for ind in list_field_ind],
```
neutrinoceros, line 150: Let's keep the meaning of the name `ind` consistent
```suggestion
        usecols=[ind for _field, ind in list_field_ind],
```

yt/frontends/ramses/io.py
```diff
@@ -118,6 +117,57 @@ def _ramses_particle_file_handler(fname, foffsets, data_types, subset, fields, c
     return tr
 
 
+def _ramses_particle_csv_file_handler(particle, subset, fields, count):
+    """General file handler for csv file, called by _read_particle_subset
+
+    Parameters
+    ----------
+    particle : ``ParticleFileHandler``
```
neutrinoceros, line 125:
```suggestion
    particle: ``ParticleFileHandler``
```

neutrinoceros (commented):
Almost there ! just a couple more small nits

Lenoble-lab:
Sorry for the delay, I was away and I completely forget...

neutrinoceros:
No need to apologise, and thank you for your contribution !

PR #4426, opened 2023-04-28
BUG: More robust detection and checking of EWAH files

jzuhone:
## PR Summary

Issue #4371 highlighted some problems with EWAH files for SPH data and @matthewturk outlined a possible route to fix them. 

The following script fails on `main`, assuming no EWAH file has been written:

```python
import yt
from yt.testing import assert_almost_equal

ds = yt.load("TNGHalo/halo_59.hdf5")

_, c = ds.find_max(("gas","density"))
c += ds.quan(0.5, "Mpc")

sp = ds.sphere(c, (20.0, "kpc"))

mass = sp.sum(("gas","mass"))

#print(ds.index.pii.filename)

del ds

ds2 = yt.load("TNGHalo/halo_59.hdf5")

_, c2 = ds2.find_max(("gas","density"))
c2 += ds2.quan(0.5, "Mpc")

sp2 = ds2.sphere(c, (20.0, "kpc"))

mass2 = sp2.sum(("gas","mass"))

assert_almost_equal(mass, mass2)
```

In the first load of the dataset, there is no EWAH file, and it creates the bitmasks, caches them in the new EWAH file, and then computes the mass within the sphere. In the second load of the dataset, it reads the bitmasks from the EWAH file, which don't match up with the `index_order2` that was used to create them, different particles are selected for the sphere, and the mass calculation fails.

What this PR does is follow @matthewturk's suggestions to resolve the issue. 

* Create a new `ParticleIndexInfo` class to store information about the index order (and if one has been specified), if there is a filename, what the original `index_order2` was before it may or may not have been updated later, and if we are allowed to change the index order. 
* Store the maximum smoothing length to the file and use it to potentially update the index order whether we are loading from an EWAH file or not. 

With these changes, the above script does not fail anymore. I would like to change the above script into a test, but we need to find some way to guarantee that the test starts cold--i.e. there is no EWAH file that exists at the beginning of the test for this dataset, which won't work if we have already run this test on a file. Any clever suggestions here would be appreciated.

Closes Issue #4371. 

## PR Checklist

### DIFF
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -250,7 +250,7 @@ class ParticleIndex(Index):
         for data_file in self.data_files:
             self.regions._set_coarse_index_data_file(data_file.file_id)
         self.regions.find_collisions_coarse()
-        if max_hsml > 0.0 and len(self.data_files) > 1:
+        if max_hsml > 0.0 and self.pii.mutable_index:
             # By passing this in, we only allow index_order2 to be increased by
             # two at most, never increased.  One place this becomes particularly
             # useful is in the case of an extremely small section of gas
@@ -259,12 +259,14 @@ class ParticleIndex(Index):
             # domain, it will correspond to a very very high index order, which
             # is a large amount of memory!  Having multiple indexes, one for
             # each particle type, would fix this.
-            new_order2 = self.regions.update_mi2(max_hsml, ds.index_order[1] + 2)
+            new_order2 = self.regions.update_mi2(max_hsml, self.pii.order2 + 2)
             mylog.info(
-                "Updating index_order2 from %s to %s", ds.index_order[1], new_order2
+                "Updating index_order2 from %s to %s", self.pii.order2, new_order2
             )
             self.pii.order2 = new_order2
-            self.ds.index_order = (self.ds.index_order[0], new_order2)
+            return max_hsml
+        else:
+            return 0.0
 
     def _initialize_refined_index(self):
         mask = self.regions.masks.sum(axis=1).astype("uint8")
--- a/yt/geometry/particle_geometry_handler.py
+++ b/yt/geometry/particle_geometry_handler.py
@@ -202,27 +202,42 @@ class ParticleIndex(Index):
         try:
             if dont_load:
                 raise OSError
-            rflag = self.regions.load_bitmasks(fname)
+            rflag, max_hsml = self.regions.load_bitmasks(fname)
+            if max_hsml > 0.0 and self.pii.mutable_index:
+                self._order2_update(max_hsml)
             rflag = self.regions.check_bitmasks()
             self._initialize_frontend_specific()
             if rflag == 0:
                 raise OSError
         except (OSError, struct.error):
             self.regions.reset_bitmasks()
-            self._initialize_coarse_index()
+            max_hsml = self._initialize_coarse_index()
             self._initialize_refined_index()
             wdir = os.path.dirname(fname)
             if not dont_cache and os.access(wdir, os.W_OK):
                 # Sometimes os mis-reports whether a directory is writable,
                 # So pass if writing the bitmask file fails.
                 try:
-                    self.regions.save_bitmasks(fname)
+                    self.regions.save_bitmasks(fname, max_hsml)
                 except OSError:
                     pass
             rflag = self.regions.check_bitmasks()
 
         self.ds.index_order = (self.pii.order1, self.pii.order2)
         del self.pii
+
+    def _order2_update(self, max_hsml):
+        # By passing this in, we only allow index_order2 to be increased by
+        # two at most, never increased.  One place this becomes particularly
+        # useful is in the case of an extremely small section of gas
+        # particles embedded in a much much larger domain.  The max
+        # smoothing length will be quite small, so based on the larger
+        # domain, it will correspond to a very very high index order, which
+        # is a large amount of memory!  Having multiple indexes, one for
+        # each particle type, would fix this.
+        new_order2 = self.regions.update_mi2(max_hsml, self.pii.order2 + 2)
+        mylog.info("Updating index_order2 from %s to %s", self.pii.order2, new_order2)
+        self.pii.order2 = new_order2
 
     def _initialize_coarse_index(self):
         max_hsml = 0.0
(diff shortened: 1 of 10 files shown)


yt/geometry/particle_geometry_handler.py
```diff
@@ -159,43 +207,57 @@ def _initialize_index(self):
...
                 # Sometimes os mis-reports whether a directory is writable,
                 # So pass if writing the bitmask file fails.
                 try:
-                    self.regions.save_bitmasks(fname)
+                    self.regions.save_bitmasks(fname, max_hsml)
                 except OSError:
                     pass
             rflag = self.regions.check_bitmasks()
 
+        self.ds.index_order = (self.pii.order1, self.pii.order2)
```
matthewturk, line 247: I think I agree with all of this -- but here's where I start to wonder if having a separate class instead of properties would be better.  I wonder if proxying `index_order` as a property that defers to `self.pii` might reduce the level of indirect?

matthewturk (dismissed):
This looks like a good solution.

yt/geometry/particle_geometry_handler.py
jzuhone: I had a look at this. The issue is that we pass this in to `yt.load` and immediately make it a property of `Dataset` (or more usually `ParticleDataset`). That's how it's accessed in the first place when we construct the `ParticleIndex` instance. So instead we'd have to pass it through `Dataset` to the index construction. I'm fine with this in principle, but it does mean adding a bunch of arguments to other functions and methods so we can pass the value down.

matthewturk: Nah, let's skip it then.

neutrinoceros:
> With these changes, the above script does not fail anymore. I would like to change the above script into a test, but we need to find some way to guarantee that the test starts cold--i.e. there is no EWAH file that exists at the beginning of the test for this dataset, which won't work if we have already run this test on a file. Any clever suggestions here would be appreciated.


Here's a solution:
```python
import yt
from numpy.testing import assert_array_equal

from yt.sample_data.api import _get_test_data_dir_path

@requires_file("TNGHalo/halo_59.hdf5")
def test_ewah(tmp_path):
    mock_file = tmp_path / "halo_59.hdf5"
    mock_file.symlink_to(_get_test_data_dir_path() / "TNGHalo" / "halo_59.hdf5")

    ds = yt.load(mock_file)
    _, c = ds.find_max(("gas","density"))
    c += ds.quan(0.5, "Mpc")
    sp = ds.sphere(c, (20.0, "kpc"))
    mass = sp.sum(("gas","mass"))

    ds2 = yt.load(mock_file)
    _, c2 = ds2.find_max(("gas","density"))
    c2 += ds2.quan(0.5, "Mpc")
    sp2 = ds2.sphere(c, (20.0, "kpc"))
    mass2 = sp2.sum(("gas","mass"))

    assert_array_equal(mass, mass2)
```

jzuhone:
@Xarthisius is there some clever way we can delete any existing EWAH file when we run a test?

jzuhone:
@neutrinoceros didn't see your script when I posted, thanks I'll check it out

jzuhone:
I added a test (thanks @neutrinoceros), and another attribute that stores whether or not we loaded from a file. 

I tested four different cases:

1) loading a file, no bitmasks stored, no `index_order` specified
2) loading a file, loading bitmasks, no `index_order` specified
3) loading a file, no bitmasks stored, `index_order` specified
4) loading a file, loading bitmasks, `index_order` specified

I think this pretty much covers everything.

yt/geometry/particle_geometry_handler.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, order1, order2, filename, mutable_index):
         self._order2_orig = order2
         self.filename = filename
         self.mutable_index = mutable_index
+        self.loaded = False
```
neutrinoceros, line 42: This doesn't seem like something we'd need to expose
```suggestion
        self._is_loaded = False
```

yt/geometry/particle_geometry_handler.py
```diff
@@ -230,6 +231,7 @@ def _initialize_index(self):
             self._initialize_frontend_specific()
             if rflag == 0:
                 raise OSError
+            self.pii.loaded = True
```
neutrinoceros, line 234:
```suggestion
            self.pii._is_loaded = True
```

yt/geometry/tests/test_particle_octree.py
```diff
@@ -398,6 +406,70 @@ def test_bitmap_select():
...
+    assert ds.index.pii.mutable_index
+    assert ds.index.pii.order1 == 6
+    assert ds.index.pii.order2_orig == 2
+    assert ds.index.pii.order2 == 4
+    assert not ds.index.pii.loaded
+    c += ds.quan(0.5, "Mpc")
+    sp = ds.sphere(c, (20.0, "kpc"))
+    mass = sp.sum(("gas", "mass"))
+
+    del ds
```
neutrinoceros, line 425: I don't see what these calls bring. Garbage collection is *not* triggered by `del`, nor is it necessary for the test to function

yt/geometry/tests/test_particle_octree.py
```diff
@@ -398,6 +406,70 @@ def test_bitmap_select():
...
+    assert not ds3.index.pii.loaded
+    c3 += ds3.quan(0.5, "Mpc")
+    sp3 = ds3.sphere(c, (20.0, "kpc"))
+    mass3 = sp3.sum(("gas", "mass"))
+
+    assert_array_equal(mass, mass3)
+
+    del ds3
+
+    ds4 = load(mock_file, index_order=(5, 3))
```
neutrinoceros, line 457: if we're going for more than 2 repeats, I think 3 would suffice, and we might as well make it in an actual loop to ensure that various repeats are perfectly equivalent

jzuhone:
@neutrinoceros ok I think I've addressed your comments. I would like to keep all four iterations of the test in (I did put them in a loop), since they are testing 4 distinct modes of operation under the hood and we want to make sure that everything does happen correctly

neutrinoceros:
thanks, I actually missed that all 4 blocks weren't identical, it's much clearer now !

neutrinoceros:
if @matthewturk is still happy with the way things go here, and tests pass, I believe this is good to go

matthewturk:
lgtm

jzuhone:
Ok, I had to restart a test because it flaked out but we should be good

neutrinoceros:
The new test is failing on Jenkins because it's pytest-only. The solution is to move it to a separate file and add this file to ignore lists in `tests/tests.yaml` and `nose_ignores.txt`

yt/geometry/tests/test_particle_octree.py
```diff
@@ -407,7 +409,8 @@ def test_bitmap_select():
 
 
 @requires_file("TNGHalo/halo_59.hdf5")
-def test_ewah_write_load(tmp_path):
+def test_ewah_write_load():
+    tmp_path = Path(tempfile.mkdtemp())
```
neutrinoceros, line 413: from `tempfile`'s doc
> The user of [mkdtemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) is responsible for deleting the temporary directory and its contents when done with it.

I'd rather we keep the `pytest` form, but if you don't want to, I'd advise using a context manager (`tempfile.TemporaryDirectory`) to avoid side effects

neutrinoceros (approved):

neutrinoceros:
FYI I think this PR would justify a bugfix release, however given the current gap between the main branch and the backport branch, I expect it would actually be easier to go for a feature release. I will raise the question on Slack later.

PR #4431, opened 2023-05-02
[BUG] Allow smoothing of derived SPH fields onto covering grids

jzuhone:
## PR Summary

The following script fails on `main`:

```python
from yt import load
bbox = [[40669.34, 56669.34], [45984.04, 61984.04], [54114.9, 70114.9]]
ds = load("TNGHalo/halo_59.hdf5", bounding_box=bbox)
w = ds.quan(0.2, "Mpc")
le = ds.domain_center-0.5*w
re = ds.domain_center+0.5*w
g = ds.r[le[0]:re[0]:128j,le[1]:re[1]:128j,le[2]:re[2]:128j]
print(g["gas","temperature"].min())
print(g["gas","kT"].min())
```

with this error:

```pycon
  File "/Users/jzuhone/fix_grid.py", line 9, in <module>
    print(g["gas","kT"].min())
          ~^^^^^^^^^^^^
  File "/Users/jzuhone/Source/yt/yt/data_objects/data_containers.py", line 235, in __getitem__
    self.get_data(f)
  File "/Users/jzuhone/Source/yt/yt/data_objects/construction_data_containers.py", line 899, in get_data
    self._fill_sph_particles(fields)
  File "/Users/jzuhone/Source/yt/yt/data_objects/construction_data_containers.py", line 977, in _fill_sph_particles
    raise KeyError(f"{ptype} is not a SPH particle type!")
KeyError: 'gas is not a SPH particle type!'
```

It appears that derived fields were not supported for this operation. This PR addresses this issue by making extra checks to determine if a field is either an SPH on-disk field, derived field, or a particle filter type of an SPH field. 

However, in order to avoid many more onerous code changes, I had to provide for the creation of fields such as `("gas", "particle_position")`, `("gas", "particle_mass")`, and `("gas","particle_position_[xyz]")`, which currently do not exist because the SPH position fields are aliased to `("gas", "[xyz]")` and the mass field is aliased to `("gas", "mass")`. Happy to discuss if this doesn't feel right to anyone.

## PR Checklist

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -969,21 +969,41 @@ class YTCoveringGrid(YTSelectionContainer3D):
         # TODO maybe there is a better way of handling this
         is_periodic = int(any(self.ds.periodicity))
 
+        sph_ptypes = self.ds._sph_ptypes
+
+        raise_error = False
+
         if smoothing_style == "scatter":
             for field in fields:
                 fi = self.ds._get_field_info(field)
-                ptype = fi.name[0]
-                if ptype not in self.ds._sph_ptypes:
-                    raise KeyError(f"{ptype} is not a SPH particle type!")
+                ptype = sph_ptypes[0]
+                ppos = [f"particle_position_{ax}" for ax in "xyz"]
+                if field[0] in self.ds.known_filters:
+                    if field[0] not in sph_ptypes:
+                        raise_error = True
+                    else:
+                        ptype = field[0]
+                        ppos = ["x", "y", "z"]
+                elif fi.is_alias:
+                    if fi.alias_name[0] not in sph_ptypes:
+                        raise_error = True
+                    elif field[0] != "gas":
+                        ptype = field[0]
+                else:
+                    if fi.name[0] not in sph_ptypes and fi.name[0] != "gas":
+                        raise_error = True
+                if raise_error:
+                    raise KeyError(f"{field[0]} is not a SPH particle type!")
+
                 buff = np.zeros(size, dtype="float64")
                 if normalize:
                     buff_den = np.zeros(size, dtype="float64")
 
                 pbar = tqdm(desc=f"Interpolating SPH field {field}")
                 for chunk in self._data_source.chunks([field], "io"):
-                    px = chunk[(ptype, "particle_position_x")].in_base("code").d
-                    py = chunk[(ptype, "particle_position_y")].in_base("code").d
-                    pz = chunk[(ptype, "particle_position_z")].in_base("code").d
+                    px = chunk[(ptype, ppos[0])].in_base("code").d
+                    py = chunk[(ptype, ppos[1])].in_base("code").d
+                    pz = chunk[(ptype, ppos[2])].in_base("code").d
                     hsml = chunk[(ptype, "smoothing_length")].in_base("code").d
                     mass = chunk[(ptype, "particle_mass")].in_base("code").d
                     dens = chunk[(ptype, "density")].in_base("code").d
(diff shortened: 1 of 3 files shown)


yt/data_objects/construction_data_containers.py
```diff
@@ -950,6 +950,28 @@ def _fill_particles(self, part):
         for p in part:
             self[p] = self._data_source[p]
 
+    def _check_sph_type(self, fi):
```
matthewturk, line 953: Can we change `fi` to `finfo` for this function?

yt/data_objects/construction_data_containers.py
```diff
@@ -950,6 +950,28 @@ def _fill_particles(self, part):
...
+            else:
+                ptype = field[0]
+        elif fi.is_alias:
+            if fi.alias_name[0] not in sph_ptypes:
+                raise_error = True
+            elif field[0] != "gas":
+                ptype = field[0]
+        else:
+            if fi.name[0] not in sph_ptypes and fi.name[0] != "gas":
+                raise_error = True
```
matthewturk, line 970: I know I should be able to figure this out, but the nested logic is not super easy to follow.  Can we put in a comment a 'truth table' or something to show the cases we're evaluating here?

matthewturk (dismissed):
So I am OK with this, and I think it's OK to go in as-is, but I did have two requests that might help future maintainers.

yt/data_objects/construction_data_containers.py
```diff
@@ -950,6 +950,28 @@ def _fill_particles(self, part):
         for p in part:
             self[p] = self._data_source[p]
 
+    def _check_sph_type(self, fi):
+        field = fi.name
```
neutrinoceros, line 954: since `field[0]` ends up being used all the time in the following I'd favour giving explicit names to the tuple elements to make it clearer and idiomatic
```suggestion
        ftype, fname = fi.name
```

yt/data_objects/construction_data_containers.py
```diff
@@ -950,6 +950,28 @@ def _fill_particles(self, part):
...
+        elif fi.is_alias:
+            if fi.alias_name[0] not in sph_ptypes:
+                raise_error = True
+            elif field[0] != "gas":
+                ptype = field[0]
+        else:
+            if fi.name[0] not in sph_ptypes and fi.name[0] != "gas":
+                raise_error = True
+        if raise_error:
+            raise KeyError(f"{field[0]} is not a SPH particle type!")
```
neutrinoceros, line 972: It might be easier to follow if we instead instantiated the error early as 
```python
err = KeyError(f"{field[0]} is not a SPH particle type!")
```

and replaced every `raise_error = True` with `raise err`

jzuhone: Done

yt/data_objects/construction_data_containers.py
jzuhone: Done

yt/data_objects/construction_data_containers.py
jzuhone: Done

yt/data_objects/construction_data_containers.py
jzuhone: Done

jzuhone:
pre-commit.ci autofix

yt/data_objects/construction_data_containers.py
```diff
@@ -950,26 +950,38 @@ def _fill_particles(self, part):
         for p in part:
             self[p] = self._data_source[p]
 
-    def _check_sph_type(self, fi):
-        field = fi.name
+    def _check_sph_type(self, finfo):
+        """
+        This function checks if a particle field has an SPH
+        type. There are several ways that this can happen,
```
neutrinoceros, line 956:
```suggestion
        Check if a particle field has an SPH type.
        There are several ways that this can happen,
```

jzuhone:
I think this should be ready to go now.

matthewturk (approved):

neutrinoceros (approved):

PR #4434, opened 2023-05-03
DEPR: deprecate numpy.testing reexports in yt.testing

### LINKED ISSUE #4433: DEPR: deprecate re-exports of numpy.testing functions in yt.testing
### Report

Currently 7 functions from the `numpy.testing` namespace are re-exported in `yt.testing`. The code currently looks like this:

```python
# we import this in a weird way from numpy.testing to avoid triggering
# flake8 errors from the unused imports. These test functions are imported
# elsewhere in yt from here so we want them to be imported here.
from numpy.testing import assert_array_equal, assert_almost_equal  # NOQA isort:skip
from numpy.testing import assert_equal, assert_array_less  # NOQA isort:skip
from numpy.testing import assert_string_equal  # NOQA isort:skip
from numpy.testing import assert_array_almost_equal_nulp  # isort:skip
from numpy.testing import assert_allclose, assert_raises  # NOQA isort:skip
from numpy.testing import assert_approx_equal  # NOQA isort:skip
from numpy.testing import assert_array_almost_equal  # NOQA isort:skip
```

Last time this was updated (~3yrs ago, when isort was first enabled), we didn't attempt to get rid of that, but there is no actual gain to doing that excepts for backward compatibility, but we can actually go through a deprecation cycle if we wrap these functions instead of directly re-exporting them.

This is probably a reasonable goal to set for the next feature release so I'll triage this into the 4.2 milestone for now

neutrinoceros:
closes #4433

### DIFF
--- a/yt/data_objects/level_sets/tests/test_clump_finding.py
+++ b/yt/data_objects/level_sets/tests/test_clump_finding.py
@@ -3,12 +3,13 @@ import shutil
 import tempfile
 
 import numpy as np
+from numpy.testing import assert_array_equal, assert_equal
 
 from yt.data_objects.level_sets.api import Clump, add_clump_info, find_clumps
 from yt.data_objects.level_sets.clump_info_items import clump_info_registry
 from yt.fields.derived_field import ValidateParameter
 from yt.loaders import load, load_uniform_grid
-from yt.testing import assert_array_equal, assert_equal, requires_file, requires_module
+from yt.testing import requires_file, requires_module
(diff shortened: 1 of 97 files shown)


matthewturk (dismissed):

neutrinoceros:
had to rebase to resolve a conflict. @matthewturk can we merge this now ?

jzuhone:
We should let the last test pass first

jzuhone (approved):

PR #4435, opened 2023-05-03
RFC: simplify base CoordinateHandler implementation (cached properties)

neutrinoceros:
## PR Summary

Some low-hanging simplifications I found while working on #4179 
`functools.cached_property` wasn't available before Python 3.8 which explains why these properties didn't use it already.

As a side bonus, I also discovered and replaced two occurrences of `abc.abstractproperty`, which is deprecated since Python 3.3

### DIFF
--- a/yt/geometry/coordinates/coordinate_handler.py
+++ b/yt/geometry/coordinates/coordinate_handler.py
@@ -235,7 +235,7 @@ class CoordinateHandler(abc.ABC):
         return ya
 
     @property
-    @abc.abstractproperty
+    @abc.abstractmethod
     def period(self):
         pass
 
--- a/yt/geometry/coordinates/coordinate_handler.py
+++ b/yt/geometry/coordinates/coordinate_handler.py
@@ -1,5 +1,6 @@
 import abc
 import weakref
+from functools import cached_property
 from numbers import Number
 from typing import Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


matthewturk:
> functools.cached_property wasn't available before Python 3.8 which explains why these properties didn't use it already.

well also even if I wrote them today I didn't know it was in `functools` so ...

yt/geometry/coordinates/coordinate_handler.py
```diff
@@ -184,98 +185,57 @@ def convert_to_spherical(self, coord):
...
+    @cached_property
     def data_projection(self):
-        if self._data_projection is not None:
-            return self._data_projection
-        dpj = {}
-        for ax in self.axis_order:
-            dpj[ax] = None
-        self._data_projection = dpj
-        return dpj
+        return {ax: None for ax in self.axis_order}
```
matthewturk, line 190: So this is definitely not a user-facing issue, but from *my* perspective, I am wondering -- is it possible to invalidate the cached property at runtime?

matthewturk (approved):
lgtm

yt/geometry/coordinates/coordinate_handler.py
neutrinoceros: The result of a cached_property can be mutated transparently. Does this answer your question ?

neutrinoceros: Oh I guess you were wondering if it was possible to reset the property to its initial state.
In which case the answer is: not to my knowledge

neutrinoceros: ... and a better answer yet, from the docs:
> The cached value can be cleared by deleting the attribute. This allows the cached_property method to run again.

PR #4437, opened 2023-05-04
[BUG] Raise exception when a file does not exist

### LINKED ISSUE #4436: Fortran reader crashes Python if the file does not exist
### Bug report

**Bug summary**

The Fortran reader crashes with a segfault if the file trying to be read does not exist.

**Code for reproduction**

```python
from yt.utilities.cython_fortran_utils import FortranFile

f = FortranFile("/this/file/does/not/exist")
# no error here, despite the fact the file does not exist
f.read_vector("d")  # will trigger a segfault
```

**Actual outcome**

The code should return an error instead of a segfault.

cphyc:
Closes #4436.

### DIFF
--- a/yt/utilities/cython_fortran_utils.pyx
+++ b/yt/utilities/cython_fortran_utils.pyx
@@ -4,6 +4,7 @@ import numpy as np
 
 from libc.stdio cimport *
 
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Good ol' "add a test and break nose" eh ?

PR #4438, opened 2023-05-04
 [BUG] Raise exception when a file does not exist

cphyc:
Closes https://github.com/yt-project/yt/issues/4436.

### DIFF
--- a/yt/utilities/cython_fortran_utils.pyx
+++ b/yt/utilities/cython_fortran_utils.pyx
@@ -4,6 +4,7 @@ import numpy as np
 
 from libc.stdio cimport *
 
+from cpython.exc cimport PyErr_SetFromErrnoWithFilename
 
 cdef INT32_SIZE = sizeof(np.int32_t)
 cdef DOUBLE_SIZE = sizeof(np.float64_t)
@@ -28,6 +29,10 @@ cdef class FortranFile:
     def __cinit__(self, str fname):
         self.cfile = fopen(fname.encode('utf-8'), 'rb')
         self._closed = False
+
+        if self.cfile is NULL:
+            self._closed = True
+            PyErr_SetFromErrnoWithFilename(FileNotFoundError, fname.encode('utf-8'))
 
     def __enter__(self):
         return self
--- a/yt/utilities/cython_fortran_utils.pyx
+++ b/yt/utilities/cython_fortran_utils.pyx
@@ -4,8 +4,6 @@ import numpy as np
 
 from libc.stdio cimport *
 
-from cpython.exc cimport PyErr_SetFromErrnoWithFilename
-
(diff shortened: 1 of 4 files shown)


yt/utilities/cython_fortran_utils.pyx
```diff
@@ -29,6 +30,10 @@ cdef class FortranFile:
         self.cfile = fopen(fname.encode('utf-8'), 'rb')
         self._closed = False
 
+        if self.cfile is NULL:
+            self._closed = True
+            PyErr_SetFromErrnoWithFilename(FileNotFoundError, fname.encode('utf-8'))
```
neutrinoceros, line 35: I can see it works as expected, but this is the first time I review code using Python's C-API directly. Can you explain why you're using `PyErr_SetFromErrnoWithFilename` here and not `PyExc_FileNotFoundError` ?

cphyc: Actually I can just replace this with a regular Python exception raising (see latest commit)

yt/utilities/tests/test_cython_fortran_utils.py
```diff
@@ -0,0 +1,26 @@
...
+    with open(dummy_file, "bw") as f:
+        f.write(
+            b" \x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x08@\x00\x00\x00\x00\x00\x00\x10@ \x00\x00\x00"
+        )
+
+    with FortranFile(dummy_file) as f:
+        np.testing.assert_equal(
+            f.read_vector("d"),
+            [1.0, 2.0, 3.0, 4.0],
+        )
```
neutrinoceros, line 26: from [pytest's docs](https://docs.pytest.org/en/7.3.x/reference/reference.html#tmpdir):
> These days, it is preferred to use tmp_path.
So let's !

```suggestion
def test_read(tmp_path):
    dummy_file = tmp_path / "test.bin"
    # Write a Fortran-formatted file containing one record with 4 doubles
    dummy_file.write_bytes(
        b" \x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x08@\x00\x00\x00\x00\x00\x00\x10@ \x00\x00\x00"
    )
    with FortranFile(str(dummy_file)) as f:
        np.testing.assert_equal(
            f.read_vector("d"),
            [1.0, 2.0, 3.0, 4.0],
        )
```

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):

PR #4439, opened 2023-05-06
DEPR: deprecate unnecessary functions in the yt.testing namespace, update docs

neutrinoceros:
## PR Summary

follow up to #4434, let's trim out the remaining fat in `yt.testing`, and clean up some import statements that I missed in the first PR

half of the "added" lines are from copying `yt.testing.expand_keywords` into the only test module that uses it

### DIFF
--- a/yt/frontends/enzo/tests/test_outputs.py
+++ b/yt/frontends/enzo/tests/test_outputs.py
@@ -1,12 +1,10 @@
 import numpy as np
+from numpy.testing import assert_almost_equal, assert_array_equal, assert_equal
 
 from yt.frontends.enzo.api import EnzoDataset
 from yt.frontends.enzo.fields import NODAL_FLAGS
 from yt.testing import (
     assert_allclose_units,
-    assert_almost_equal,
-    assert_array_equal,
-    assert_equal,
     requires_file,
     requires_module,
     units_override_check,
--- a/doc/source/analyzing/filtering.rst
+++ b/doc/source/analyzing/filtering.rst
@@ -265,7 +265,7 @@ union into a single ``darkmatter`` field.  The ``all`` particle type is a
 special case of this.
 
 To create a particle union, you need to import the ``ParticleUnion`` class from
-``yt.data_objects.particle_unions``, which you then create and pass into
+``yt.data_objects.unions``, which you then create and pass into
 ``add_particle_union`` on a dataset object.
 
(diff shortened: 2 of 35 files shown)


neutrinoceros:
switching to draft until I get rid of _all_ remaining self-triggered deprecation warnings in answer tests (which requires manual checking since nose cannot treat them as errors)

neutrinoceros:
@yt-fido test this please

yt/frontends/enzo/tests/test_outputs.py
```diff
@@ -121,18 +119,18 @@ def test_moving7():
 def test_galaxy0030():
     ds = data_dir_load(g30)
     yield check_color_conservation(ds)
-    assert_equal(str(ds), "galaxy0030")
+    assert str(ds) == "galaxy0030"
     for test in big_patch_amr(ds, _fields):
         test_galaxy0030.__name__ = test.description
         yield test
-    assert_equal(ds.particle_type_counts, {"io": 1124453})
+    assert ds.particle_type_counts == {"io": 1124453}
```
matthewturk, line 126: One thing I like about `assert_equal` is that it shows the actual and expected values in case of a failure.  Does running this in modern fixtures *also* provide that, or no?

matthewturk (commented):
before reviewing more, asking a question

yt/frontends/enzo/tests/test_outputs.py
neutrinoceros: `pytest` monkey patches the `assert` statement to bridge the gap and make stuff like `assert_equal` unnecessary. *but* this particular test uses `yield` so it cannot be run with pytest until further effort is spent on migrating it. I am admittedly getting a little bit ahead of myself here. Should I revert this part ?

matthewturk: I think it would be easier to review, and since we can't use the new monkeypatched version yet, let's keep it.

jzuhone (approved):

yt/frontends/enzo/tests/test_outputs.py
neutrinoceros: done !

jzuhone (approved):

neutrinoceros:
@matthewturk do you still want to have a look ?

matthewturk:
LGTM

PR #4440, opened 2023-05-10
Off-axis (non-SPH) particle projections

jzuhone:
## PR Summary

This PR adds support for off-axis projections of particles which are not SPH particles, by adding this capability to the `ParticleProjectionPlot` class for plots, and adding a new class `FITSParticleOffAxisProjection` for FITS files. 

Example:

```python
 import yt

 L = [1, 1, 1] # normal or "line of sight" vector
 N = [0, 1, 0] # north or "up" vector

 p = yt.ParticleProjectionPlot(ds, L, [("all", "particle_mass")], width=(0.05, 0.05), depth=0.3,
                               north_vector=N)
 p.set_unit(("all", "particle_mass"), "Msun")
 p.save()
```

Still need to add tests, will do that soon. 

galaxy0030_Particle_particle_mass

## PR Checklist


yt/visualization/fixed_resolution.py
```diff
@@ -708,21 +712,37 @@ def __getitem__(self, item):
...
-            if hasattr(b, "in_units"):
-                b = float(b.in_units("code_length"))
-            bounds.append(b)
+        dd = self.data_source.dd
 
         ftype = item[0]
-        x_data = self.data_source.dd[ftype, self.x_field]
-        y_data = self.data_source.dd[ftype, self.y_field]
-        data = self.data_source.dd[item]
+        if self.axis is None:
```
matthewturk, line 718: Is `code_length` the right thing to do here?  I know we did it above, before, but I need a check on that -- should it be that, or `unitary`?

yt/visualization/fixed_resolution.py
```diff
@@ -708,21 +712,37 @@ def __getitem__(self, item):
...
+                self.data_source.normal_vector,
+                self.data_source.north_vector,
+            )
+            bounds = [rbx0, rbx1, rby0, rby1]
+            x_data = np.array(x_data)
+            y_data = np.array(y_data)
+        else:
+            bounds = []
+            for b in self.bounds:
+                if hasattr(b, "in_units"):
```
matthewturk, line 735: We have an `_ensure_code` in cython, but not in Python.  (We have one that's specific to volume rendering.)

yt/visualization/fixed_resolution.py
```diff
@@ -708,21 +712,37 @@ def __getitem__(self, item):
...
+            y_data = dd[ftype, self.y_field].to_value("code_length")
+        data = dd[item]
 
         # handle periodicity
-        dx = x_data.in_units("code_length").d - bounds[0]
-        dy = y_data.in_units("code_length").d - bounds[2]
-        if self.periodic:
+        dx = x_data - bounds[0]
+        dy = y_data - bounds[2]
+        if self.axis is not None and self.periodic:
```
matthewturk, line 745: `self.periodic` is a tuple, right?  I know this code predates the PR but shouldn't this be an `all` or `self.periodic[self.axis]` (well probably not that) or something?

matthewturk (approved):
Looks alright!  Minor comments.

matthewturk:
While I'm thinking of it, I do also think we should consider a mode of rotation for particle-based datasets.  Specifically, it seems like we could interpose anywhere the IO happens to have the particles rotated.  I don't think this would work with a context manager (necessarily) but maybe could.

yt/visualization/fixed_resolution.py
jzuhone: Not sure why it matters either way--could you clarify?

jzuhone: Not sure why it matters either way--could you clarify?

matthewturk: I think I convinced myself it's OK, since everything will be in *the same* bounds.  I'm sorry for the noise.

yt/visualization/fixed_resolution.py
jzuhone: Actually here it's just a scalar, I checked (you have to kind of dive down but it's in the signature for the `PlotWindow` class.

yt/visualization/fixed_resolution.py
jzuhone: No, no--happy to have the thought.

jzuhone:
> While I'm thinking of it, I do also think we should consider a mode of rotation for particle-based datasets. Specifically, it seems like we could interpose anywhere the IO happens to have the particles rotated. I don't think this would work with a context manager (necessarily) but maybe could.



This is a good thought--I've recently had occasion to make copies of particle datasets where I have rotated and shifted the coordinates and velocities.

jzuhone:
> While I'm thinking of it, I do also think we should consider a mode of rotation for particle-based datasets. Specifically, it seems like we could interpose anywhere the IO happens to have the particles rotated. I don't think this would work with a context manager (necessarily) but maybe could.



This is a good thought--I've recently had occasion to make copies of particle datasets where I have rotated and shifted the coordinates and velocities.

jzuhone:
I am completely baffled by the Jenkins failures.

jzuhone:
@yt-fido test this please

jzuhone:
@Xarthisius any idea what happened with the seemingly unrelated Jenkins failures?

neutrinoceros:
@jzuhone it's not impossible that those failures are due to a new incompatibility with old unyt pickled data and free ones produced with sympy 1.12 (released 2 days ago).

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1839,6 +1828,34 @@ def off_axis_projection_SPH(np.float64_t[:] px,
...
+                            bounds,
+                            center,
+                            width,
+                            np.float64_t[:] quantity_to_smooth,
+                            np.float64_t[:, :] projection_array,
+                            normal_vector,
+                            north_vector,
+                            weight_field=None):
+    # Do nothing in event of a 0 normal vector
+    if np.allclose(normal_vector, np.array([0., 0., 0.]), rtol=1e-09):
```
neutrinoceros, line 1851: I guess this could be simplified ? (relative tolerance has no effect when comparing to 0)
```suggestion
    if np.allclose(normal_vector, 0.):
```

yt/visualization/fits_image.py
```diff
@@ -906,17 +910,20 @@ def construct_image(
...
-            center[axes[1]] - width[1] / 2,
-            center[axes[1]] + width[1] / 2,
-        )
-        frb = ParticleImageBuffer(
-            data_source, bounds, (nx, ny), periodic=all(ds.periodicity)
-        )
+    elif isinstance(data_source, ParticleDummyDataSource):
+        if hasattr(data_source, "normal_vector"):
+            bounds = (-width[0] / 2, width[0] / 2, -width[1] / 2, width[1] / 2)
+            periodic = False
```
neutrinoceros, line 916: I don't think I get this. Wouldn't an off-axis projection be periodic if axis-boundaries are *all* periodic ?

neutrinoceros (commented):
I'll try to make time for a thorough review soon, for now here are a couple comments

yt/utilities/lib/pixelization_routines.pyx
jzuhone: W_iiii_

jzuhone: I refactored a function and this bit of code came over from that, so we've had it for a while. I can change it, though.

yt/visualization/fits_image.py
jzuhone: if I recall correctly we do not support periodicity for any off-axis projection plots. Isn't that correct, @matthewturk?

matthewturk: This is my recollection.

yt/visualization/particle_plots.py
```diff
@@ -243,15 +345,13 @@ def __init__(
         data_source=None,
         deposition="ngp",
         density=False,
+        north_vector=None,
```
neutrinoceros, line 348:
```suggestion
        *,
        north_vector=None,
```

yt/visualization/plot_window.py
```diff
@@ -1618,7 +1618,7 @@ class ProjectionPlot(NormalPlot):
         This is the dataset object corresponding to the
         simulation output to be plotted.
     normal : int, str, or 3-element sequence of floats
-        This specifies the normal vector to the slice.
+        This specifies the normal vector to the projection.
```
neutrinoceros, line 1621: nice catch !

neutrinoceros (commented):
Overall looks great. I got a couple nits and questions, but more importantly, a comment on how to change the name of the argument (`axis` -> `normal`) without immediately breaking backward compat.

yt/visualization/fixed_resolution.py
```diff
@@ -708,21 +712,41 @@ def __getitem__(self, item):
...
+            wd = []
+            for w in self.data_source.width:
+                if hasattr(w, "to_value"):
+                    w = w.to_value("code_length")
+                wd.append(w)
+            x_data, y_data, rbx0, rbx1, rby0, rby1 = rotate_particle_coord(
+                dd[ftype, "particle_position_x"].to_value("code_length"),
+                dd[ftype, "particle_position_y"].to_value("code_length"),
+                dd[ftype, "particle_position_z"].to_value("code_length"),
+                self.data_source.center.to("code_length").d,
```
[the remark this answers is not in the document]
jzuhone: It's not the same because the first two you mention return a NumPy array while `in_units` returns an `unyt_array`

yt/visualization/fixed_resolution.py
```diff
@@ -708,21 +712,41 @@ def __getitem__(self, item):
...
+            x_data = np.array(x_data)
+            y_data = np.array(y_data)
+        else:
+            bounds = []
+            for b in self.bounds:
+                if hasattr(b, "to_value"):
+                    b = b.to_value("code_length")
+                bounds.append(b)
+            x_data = dd[ftype, self.x_field].to_value("code_length")
+            y_data = dd[ftype, self.y_field].to_value("code_length")
```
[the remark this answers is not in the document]
jzuhone: `to_value` and `in_units` are not equivalent since the former drops units while `in_units` keeps them.

yt/visualization/fixed_resolution.py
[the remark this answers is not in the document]
neutrinoceros: right, forget about that secondary suggestion then :)

yt/visualization/particle_plots.py
```diff
@@ -45,14 +55,47 @@ def __init__(
         else:
             self.field_parameters = field_parameters
 
+        fields = self.dd._determine_fields(fields)
+        self.fields = fields
+
+    def _determine_fields(self, *args):
+        return self.dd._determine_fields(*args)
```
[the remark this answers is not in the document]
jzuhone: We have to do a similar thing for off-axis projection data sources for grid or SPH-based data, so it is used elsewhere. But we can change it on line 47.

yt/visualization/fixed_resolution.py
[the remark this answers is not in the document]
neutrinoceros: yup, sorry, carry on !

yt/visualization/fixed_resolution.py
```diff
@@ -708,21 +712,41 @@ def __getitem__(self, item):
             deposition,
         )
 
-        bounds = []
-        for b in self.bounds:
-            if hasattr(b, "in_units"):
-                b = float(b.in_units("code_length"))
-            bounds.append(b)
+        dd = self.data_source.dd
```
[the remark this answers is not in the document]
jzuhone: This is essentially copying what is done with the `OffAxisProjectionDummyDataSource` in `yt/visualization/plot_window.py`. It represents either "all data" or a data source object. In this case, it's set in the subclasses of `ParticleDummyDataSource`: `ParticleAxisAlignedDummyDataSource` and `ParticleOffAxisDummyDataSource`.

yt/visualization/particle_plots.py
```diff
@@ -23,20 +30,23 @@ def __init__(
...
         field_parameters=None,
-        data_source=None,
         deposition="ngp",
         density=False,
     ):
         self.center = center
         self.ds = ds
-        self.axis = axis
         self.width = width
+        self.dd = dd
```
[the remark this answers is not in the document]
jzuhone: See above.

yt/visualization/particle_plots.py
```diff
@@ -227,7 +329,7 @@ class ParticleProjectionPlot(PWViewerMPL):
     def __init__(
         self,
         ds,
-        axis,
+        normal,
```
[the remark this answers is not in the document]
jzuhone: Done

yt/visualization/particle_plots.py
```diff
@@ -61,27 +104,78 @@ def __init__(
...
+        super().__init__(
+            center,
+            ds,
+            width,
+            fields,
+            dd,
+            weight_field=weight_field,
+            field_parameters=field_parameters,
+            deposition=deposition,
+            density=density,
```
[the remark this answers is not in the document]
jzuhone: Done

yt/visualization/particle_plots.py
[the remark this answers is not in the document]
jzuhone: By "elsewhere" I mean within `PlotWindow` methods

yt/visualization/particle_plots.py
```diff
@@ -243,15 +349,26 @@ def __init__(
...
+            issue_deprecation_warning(
+                "The 'axis' argument is a deprecated alias for the 'normal' argument. ",
+                "4.2.0",
+                "4.3.0",
+            )
+            normal = axis
+        if normal is None:
+            raise ValueError(
+                "You must specify an axis name or normal vector for this plot."
+            )
```
neutrinoceros, line 366: Matching the error message produced with, for instance `def foo(bar): pass; foo()`
```suggestion
            raise TypeError(
                "ParticleProjectionPlot() missing 1 required positional argument: 'normal'"
            )
```

yt/visualization/particle_plots.py
```diff
@@ -243,15 +349,26 @@ def __init__(
...
         density=False,
+        *,
+        north_vector=None,
+        axis=None,
     ):
+        if axis is not None:
+            issue_deprecation_warning(
+                "The 'axis' argument is a deprecated alias for the 'normal' argument. ",
+                "4.2.0",
+                "4.3.0",
```
neutrinoceros, line 360: We don't need to try and predict when we'll be able to end deprecation cycles any more :)
```suggestion
                since="4.2.0",
```

neutrinoceros (commented):
finishing with details of the deprecation cycle, I'll try to circle back for an other full review tonight

yt/visualization/particle_plots.py
neutrinoceros: (originally wrote this as a nit, but note that `since` and `removal` arguments are actually keyword-only)

yt/visualization/fits_image.py
```diff
@@ -906,17 +910,20 @@ def construct_image(
...
-            center[axes[0]] - width[0] / 2,
-            center[axes[0]] + width[0] / 2,
-            center[axes[1]] - width[1] / 2,
-            center[axes[1]] + width[1] / 2,
-        )
-        frb = ParticleImageBuffer(
-            data_source, bounds, (nx, ny), periodic=all(ds.periodicity)
-        )
+    elif isinstance(data_source, ParticleDummyDataSource):
+        if hasattr(data_source, "normal_vector"):
```
neutrinoceros, line 914: `data_source.normal_vector` is not actually used in the following so it is a bit obscure why the branching is done depending on wether or not this attribute exis. If we're going to treat `OffAxis...DummyDataSource` classes as a protocol, I'd rather we do it more explicitly (yes it's uglier, but also much easier to understand and refactor if we ever manage to)

```suggestion
        if re.fullmatch(r"OffAxis\w+DummyDataSource", type(data_source).__name__):
```

yt/visualization/fixed_resolution.py
```diff
@@ -708,21 +712,40 @@ def __getitem__(self, item):
...
+            y_data = dd[ftype, self.y_field].to_value("code_length")
+        data = dd[item]
 
         # handle periodicity
-        dx = x_data.in_units("code_length").d - bounds[0]
-        dy = y_data.in_units("code_length").d - bounds[2]
-        if self.periodic:
+        dx = x_data - bounds[0]
+        dy = y_data - bounds[2]
+        if self.axis is not None and self.periodic:
```
neutrinoceros, line 748: haven't we established that `self.periodic is False` implies `self.axis is None` ? I may have missed something but the extra condition seems superfluous to me

yt/visualization/particle_plots.py
```diff
@@ -270,28 +386,58 @@ def __init__(
             self._use_cbar = False
             splat_color = color
 
-        depth = ds.coordinates.sanitize_depth(depth)
+        if isinstance(normal, (int, str)):
```
neutrinoceros, line 389: at this point, `normal` was already sanitized and can only be of type `str` or `np.ndarray` (unless `NormalPlot.sanitize_normal_vector`'s annotations are wrong ?). Something is inconsistent here.

I think the class hierarchy for `ParticleProjectionPlot` should follow the same design pattern as `SlicePlot` and `ProjectionPlot`.

neutrinoceros (changes requested):
Sorry I wasn't able to review as early as I wanted. I have some important concerns with the current design of the `ParticleProjectionPlot` class, and I'd like it be made closer to `ProjectionPlot` while we have the chance.

yt/visualization/particle_plots.py
jzuhone: I am not sure why offhand, but the code didn't work correctly without this last week. I'll have to look but I am not sure how much time I am going to have in the rest of this week.

yt/visualization/fixed_resolution.py
jzuhone: Probably superfluous, will have to check.

yt/visualization/fits_image.py
jzuhone: I'd prefer we find a less ugly way to do this.

jzuhone: It's frankly better to add a comment here why we check this.

neutrinoceros: There's way to implement this check using `isinstance` instead of `hasattr` that
* can be checked continuously
* doesn't require a (possibly outdated) comment 
* doesn't require inheritance

Namely: [`typing.Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) (which we do have access to since we don't support Python < 3.8 any more)

For illustration, here's a minimal implementation
```python
from typing import Protocol, runtime_checkable
import numpy as np

@runtime_checkable
class OffAxisDummyDataSource:
    normal_vector: np.ndarray

class Dumdumdatasource:
    normal_vector = np.zeros(3)
    
assert isinstance(Dumdumdatasource(), OffAxisDummyDataSource)
```

I understand that this goes beyond the scope of this PR, which is why I instead suggested an (ugly) one-line refactor to make the code more expressive, but long term I'd like to start using such protocol classes to replace calls to `hasattr`; IMO they would fit nicely in the `yt._typing` namespace. Is this something we can all get behind ?

jzuhone:
@yt-fido test this please

jzuhone:
I attempted to implement the new functionality in a way that mimicks what we do for `ProjectionPlot`, `AxisAlignedProjectionPlot`, and `OffAxisProjectionPlot`, but it turns out that this was trickier than it first appeared since there's a lot of shared functionality that's required for these particle plots between the two different cases that is not true for the non-particle plots. There may be a better way to do this but it's way out of scope here.  

No idea why the doc build is failing, looks like a separate issue.

jzuhone:
@yt-fido test this please

yt/visualization/particle_plots.py
```diff
@@ -438,20 +440,28 @@ def __init__(
             )
 
             oblique = True
+            periodic = False
+            if origin != "center-window":
+                mylog.warning(
+                    "The 'origin' keyword is ignored for off-axis "
+                    "particle projections, it is always 'center-window'"
+                )
```
neutrinoceros, line 448: shouldn't this be a actual warning (`warnings.warn`) ? My rule of thumb is that `mylog.warning` should only be used for stuff users have no control over, and should be pretty rare.

matthewturk: It's probably fine as is.

jzuhone: I did it this way because we used it similarly in `PlotWindow` for something similar.

jzuhone:
this is ready to go

yt/visualization/fits_image.py
neutrinoceros: I'd like an answer to this before I hit approve. I'm happy to put in the work for yt 4.3.0

jzuhone: I'm obviously happy to discuss anything, and it does look like a better long-term solution, but needless to say we'll have to hash that out in a different forum. 

That ok?

neutrinoceros: Of course, I just didn't want to leave unresolved threads before merging this important PR.

neutrinoceros (approved):

PR #4441, opened 2023-05-12
tmp jenkins trigger

neutrinoceros:
## PR Summary
I suspect sympy 1.12 (released 2 days ago) is causing some subtle incompatibilities with old test pickled objects (as seen in https://tests.yt-project.org/job/yt_py38_git/6760/#showFailuresLink:
```python-traceback
unyt.exceptions.UnitConversionError: Cannot convert between 'K' (dim '(temperature)') and 'K' (dim '(temperature)').
```

I'll use this PR to trigger tests twice (with sympy 1.11.x and 1.12.0) to put my hypothesis to the test

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -221,7 +221,7 @@ test = [
     "pyaml>=17.10.0",
     "pytest>=6.1",
     "pytest-mpl>=0.16.1",
-    "sympy!=1.10,!=1.9", # see https://github.com/sympy/sympy/issues/22241
+    "sympy!=1.10,!=1.9,!=1.12", # see https://github.com/sympy/sympy/issues/22241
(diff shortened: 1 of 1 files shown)


neutrinoceros:
that's definitely the problem. I'm working on a minimal reproducer, I'll close this when/if I can confidently upstream the issue.

neutrinoceros:
reported upstream as https://github.com/sympy/sympy/issues/25134
We'll probably need to address the problem somehow on our side depending on the response we get:
- sympy dev may not want to support our use case of old pickles for ever, in which case we'll need to generate newer ones
- unless this is patched upstream and gets into a bugfix release very soon, we'll need to forbid sympy 1.12 in testing env.

This PR served it's purpose, so I'm closing it, but the discussion may be continued in an issue.

PR #4443, opened 2023-05-13
BUG: fix a bug in Dataset.print_stats() where length units would be printed twice

neutrinoceros:
## PR Summary

reprod: `python -c "import yt; yt.load_sample('IsolatedGalaxy').print_stats()"`

output (main):

```
...
Smallest Cell:
	Width: 1.221e-04 Mpc Mpc
	Width: 1.221e+02 pc pc
	Width: 2.518e+07 AU AU
	Width: 3.767e+20 cm cm
```
output (this branch):
```
...
Smallest Cell:
	Width: 1.221e-04 Mpc
	Width: 1.221e+02 pc
	Width: 2.518e+07 AU
	Width: 3.767e+20 cm
```

`unyt_array` already implements str formatting in a way that includes units, so including it a second time is redundant

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -689,7 +689,7 @@ class RAMSESIndex(OctreeIndex):
         )
         print("\nSmallest Cell:")
         for item in ("Mpc", "pc", "AU", "cm"):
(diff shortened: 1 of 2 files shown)


matthewturk (approved):
Wooo! This has bugged me for a while! Thank you!

PR #4447, opened 2023-05-20
RFC: cleanup unreachable and unnecessary special cases (colorbar minor ticks)

neutrinoceros:


yt/visualization/base_plot_types.py
```diff
@@ -347,22 +347,15 @@ def _set_axes(self, norm: Normalize) -> None:
...
                         flinthresh /= 10
                     mticks = get_symlog_minorticks(flinthresh, norm.vmin, norm.vmax)
-                    if MPL_VERSION < Version("3.5.0b"):
-                        # https://github.com/matplotlib/matplotlib/issues/21258
-                        mticks = self.image.norm(mticks)
                     self.cax.yaxis.set_ticks(mticks, minor=True)
-
-            elif isinstance(norm, LogNorm):
-                self.cax.minorticks_on()
-                self.cax.xaxis.set_visible(False)
```
neutrinoceros, line 364: this line used to be needed to workaround a bug in matplotlib, but it is fixed in all versions we currently support.

neutrinoceros:
While working on this I discovered #4448.
I'll close this in favour of a more engaged patch (I already have a proof of concept, but I may need a couple days before I have stable solution).

PR #4455, opened 2023-05-26
ENH: add support for passing field keys for colors and linewidth in streamline plot annotations

### LINKED ISSUE #4454: ENH: support using a yt field for streamline line width and color
[matplotlib's `streamplot` function](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.streamplot.html#matplotlib.axes.Axes.streamplot) takes a optional `linewidth`  and `color` argument which expect a 2D array. In theory it is already possible to use them in `PWViewerMPL.annotate_streamlines`, but one has to manually retrieve the relevant field data. It would be much more convenient to just pass `plot.annotate_streamlines(..., linewidth=("gas", "velocity_magnitude"))` and let yt do the heavy lifting.

I also (naively ?) except this to be a relatively easy patch to the `StreamlineCallback` class.

neutrinoceros:
## PR Summary
Closes #4454, incidentally fix a couple small issues with `StreamlineCallback`'s implementation.

demo:

```python
import yt

p = yt.ProjectionPlot(ds, "z", ("gas", "density"))
p.annotate_streamlines(
    ("gas", "velocity_x"),
    ("gas", "velocity_y"),
    linewidth=("gas", "velocity_z"),
    linewidth_upscaling=3,
    color=("gas", "velocity_magnitude"),
    cmap="turbo",
)
p.zoom(20)
p.save("/tmp/")
```

galaxy0030_Projection_z_density

## PR Checklist

- [x] write error and deprecation messages
- [x] extra validation of input types (in particular, check consistency with MPL)

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -9,9 +9,10 @@ from typing import Any, Dict, Optional, Tuple, Type, Union
 
 import matplotlib
 import numpy as np
+from unyt import unyt_quantity
 
 from yt._maintenance.deprecation import issue_deprecation_warning
-from yt._typing import AnyFieldKey
+from yt._typing import AnyFieldKey, FieldKey
 from yt.data_objects.data_containers import YTDataContainer
 from yt.data_objects.level_sets.clump_handling import Clump
 from yt.data_objects.selection_objects.cut_region import YTCutRegion
@@ -21,7 +22,12 @@ from yt.geometry.api import Geometry
 from yt.geometry.unstructured_mesh_handler import UnstructuredIndex
 from yt.units import dimensions
 from yt.units.yt_array import YTArray, YTQuantity, uhstack  # type: ignore
-from yt.utilities.exceptions import YTDataTypeUnsupported, YTUnsupportedPlotCallback
+from yt.utilities.exceptions import (
+    YTDataTypeUnsupported,
+    YTFieldNotFound,
+    YTFieldTypeNotFound,
+    YTUnsupportedPlotCallback,
+)
 from yt.utilities.lib.geometry_utils import triangle_plane_intersect
 from yt.utilities.lib.line_integral_convolution import line_integral_convolution_2d
 from yt.utilities.lib.mesh_triangulation import triangulate_indices
@@ -38,6 +44,11 @@ from yt.visualization._commons import (
 from yt.visualization.base_plot_types import CallbackWrapper
 from yt.visualization.image_writer import apply_colormap
 from yt.visualization.plot_window import PWViewerMPL
+
+if sys.version_info >= (3, 10):
+    from typing import TypeGuard
+else:
+    from typing_extensions import TypeGuard
 
 if sys.version_info >= (3, 11):
     from typing import assert_never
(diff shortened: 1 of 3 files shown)


yt/visualization/plot_modifications.py
```diff
@@ -1254,44 +1273,41 @@ def __init__(
...
-                    self.plot_args["linewidth"] = linewidth
-                except ValueError as e:
-                    err_msg = (
-                        "Error applying display threshold: linewidth"
-                        + "must have shape ({}, {}) or be scalar"
-                    )
-                    err_msg = err_msg.format(nx, ny)
-                    raise ValueError(err_msg) from e
+        if linewidth is not None:
+            linewidth *= self._linewidth_upscaling / np.abs(linewidth).max()
```
neutrinoceros, line 1303: I should probably expose the normalisation factor directly and document that leaving it unspecified will result in normalising by the field's maximum.

yt/visualization/plot_modifications.py
```diff
@@ -1254,44 +1273,41 @@ def __init__(
...
         # We are feeding this size into the pixelizer, where it will properly
         # set it in reverse order
         nx = plot.raw_image_shape[1] // self.factor[0]
         ny = plot.raw_image_shape[0] // self.factor[1]
-        pixX = plot.data.ds.coordinates.pixelize(
-            plot.data.axis, plot.data, self.field_x, bounds, (nx, ny)
-        )
-        pixY = plot.data.ds.coordinates.pixelize(
-            plot.data.axis, plot.data, self.field_y, bounds, (nx, ny)
+        pixelize = partial(
```
matthewturk, line 1282: neat use of partial

yt/visualization/plot_modifications.py
matthewturk: Not sure.  Does it know how to normalize on log-scaled fields?

matthewturk (commented):
like it!

yt/visualization/plot_modifications.py
neutrinoceros: It doesn't, but there's actually already a way to use *arbitrary* scaling here by defining custom fields that can then be used as `linewidth`, so in hindsight I think it's already general enough. I could explicitly mention this idea in docs.

neutrinoceros: Maybe it should indeed take logs automatically when required by field_info. TBH I don't think much of this mechanism because I feel like it locks users into arbitrary defaults instead of empowering them, but consistency should prevail, especially since, as I pointed out, there's already a way to get out of defaults completely.
That said, and while we're at it, is there a way to (re)set "take_log" parameters from the user side that I just don't know about ?

matthewturk: I think it's OK to go in as is, and I think improving field definitions to more flexibly support logs is a good medium-term goal.

matthewturk (approved):

neutrinoceros:
had an insignificant conflict, fixed it

neutrinoceros:
@matthewturk if you're still up for it, this is now ready to go

matthewturk (approved):
I actually thought I'd already approved!

neutrinoceros:
you did, I just 'dismissed' your approval when I fixed a conflict !

PR #4460, opened 2023-06-02
ENH: add support for passing arbitrary kwargs for axes.contours in annotate_contour

### LINKED ISSUE #4459: BUG: yt may leak matplotlib configuration globally at runtime
### Bug report

**Bug summary**

Matplotlib contour styles for monochrome contour sets can be changed globally by yt at runtime (this is triggered by importing from `yt.visualization._mpl_imports`, which isn't implied by `import yt`).
The problematic line is very obviously:
https://github.com/yt-project/yt/blob/e525381f745a0b1378ecac55c36c16721516071d/yt/visualization/_mpl_imports.py#L10

(in fact I spotted the problem in the code first, and *then* worked on a script to show its undesired effect)

**Code for reproduction**

```python
# adapted from https://matplotlib.org/stable/gallery/images_contours_and_fields/contour_label_demo.html#sphx-glr-gallery-images-contours-and-fields-contour-label-demo-py
import numpy as np
import matplotlib as mpl
import yt

delta = 0.025
x = np.arange(-3.0, 3.0, delta)
y = np.arange(-2.0, 2.0, delta)
X, Y = np.meshgrid(x, y)
Z1 = np.exp(-X**2 - Y**2)
Z2 = np.exp(-(X - 1)**2 - (Y - 1)**2)
Z = (Z1 - Z2) * 2
zm = Z.mean()
Zp = Z - zm

def make_figure(filename):
    # Basic contour plot
    fig, ax = plt.subplots()
    CS = ax.contour(X, Y, Zp, linestyles=None, colors="black")
    #assert CS.negative_linestyles == "dashed"

    print(f"saving to {filename}")
    fig.savefig(filename)

# draw the figure
make_figure("/tmp/pre.png")

# do stuff with yt
yt.SlicePlot(ds, "z", "density").save("/tmp/")

# try the pure-mpl figure again
make_figure("/tmp/post.png")
```

**Actual outcome**
pre
post

**Expected outcome**

pure matplotlib graphs should not be affected by yt at runtime.

neutrinoceros:
## PR Summary

incidentally fix #4459

TODO: improve (rewrite) docstring, most importantly to clarify where arbitrary kwargs are being sent to


yt/visualization/plot_modifications.py
```diff
@@ -914,13 +917,23 @@ def __init__(
...
+            _plot_args.update(plot_args)
+        _plot_args.update({"negative_linestyles": negative_linestyles, **kwargs})
+
+        self.plot_args = _plot_args
         self.label = label
         if text_args is None:
-            text_args = def_text_args
-        self.text_args = text_args
+            text_args = {}
+        self.text_args = {**def_text_args, **text_args}
```
neutrinoceros, line 936: this fixes an unrelated bug where `def_text_args` is completely ignored if any text_arg is supplied.

yt/visualization/plot_modifications.py
```diff
@@ -914,13 +950,23 @@ def __init__(
...
+            "linestyles": "solid",
+            **(plot_args or {}),
+            **kwargs,
+        }
+
         self.label = label
-        if text_args is None:
-            text_args = def_text_args
-        self.text_args = text_args
+        self.text_args = {"colors": "white", **(text_args or {})}
```
neutrinoceros, line 969: this fixes an unrelated bug where def_text_args is completely ignored if any text_arg is supplied.

neutrinoceros:
I'm going to prioritise #4475 over this.

neutrinoceros:
rebased to resolve merge conflicts.
To be explicit, this is ready for review :)

yt/visualization/plot_modifications.py
```diff
@@ -886,12 +886,49 @@ def _transformed_field(field, data):
...
+        An alternative data source. By default the plot's data_source is queried.
+
+    label: bool (default: False)
+        Toggle contour labels.
+
+    text_args: dict[str, Any] (optional)
+        Used only if `label=True`. Arbitrary keyword arguments to be passed to
+        matplotlib.axes.Axes.clabel
+        In contrast to matplotlib.axes.Axes.clabel, the following
+        parameters are used by default: colors='white'
```
brittonsmith, line 925: Maybe I'm misunderstanding, but it looks like the default color is black. I'm looking at what's in `self.plot_args` below.

brittonsmith (commented):
Is there an advantage to replacing the `plot_args` dict with `**kwargs`? My impression is that we had moved away from `**kwargs` toward dicts as a means of keeping keywords grouped according to destination. For example, in this same function, we have a `text_args` dict. Unless there is a specific benefit, I'm not sure I support this change.

yt/visualization/plot_modifications.py
neutrinoceros: These are the default values for *text*, not contours, and I confirm that they are in line with the code.

neutrinoceros:
> My impression is that we had moved away from **kwargs toward dicts as a means of keeping keywords grouped according to destination. For example, in this same function, we have a text_args dict. Unless there is a specific benefit, I'm not sure I support this change.

This strategy still has value *but* it's not *necessary* to have *all* arguments be passed through dictionaries: (n-1) dictionaries arguments (with n, the number of destinations) is sufficient ! My point (and proposed change) is that we can still leverage `**kwargs` (which is very idiomatic in matplotlib) for the most common destination without limiting design space.

yt/visualization/plot_modifications.py
brittonsmith: Ok, thanks.

brittonsmith:
The minor benefit I see from keeping `text_args` and `plot_args` is that it is clear where the keywords are being shipped off to. Other than this, this change feels like mostly a lateral move with the downside of breaking backward compatibility. I think this is the point where it is probably worth getting other opinions. I'd rather not see API changes made or blocked based on individual opinions, but I am happy to defer to a majority. @matthewturk and I have not discussed this (although maybe you have with him), so he should have a fresh opinion.

neutrinoceros:
The way I see it there are three options to choose from:
1. keep the API from `main` (no support for `**kwargs`)
2. support `**kwargs` and deprecate `plot_args` (this is the current state of the branch)
3. support *both* APIs (same as this branch, but without the deprecation)

I don't think any option is ideal: there are precedents to support options 1. and 2., and I have a personal preference to not keep support for more than one API for ever, but maybe option 3. is probably an agreeable middle ground ?

PR #4461, opened 2023-06-02
Don't force import of netCDF4 bindings along yt

neutrinoceros:
## PR Summary

This is the boldest approach to closing #4451

PR #4464, opened 2023-06-02
Move netcdf4 forced import from netcdf4_imports to h5py_imports

### LINKED ISSUE #4451: DEP: reliably solving import order (h5py VS netcdf4)
`netcdf4` and `h5py` are both (currently) optional dependencies. We use proxy objects defined in `yt.utilities.on_demand_imports` to access their functionalities, which are instances of the `OnDemand` class, whose whole purpose is to delay imports until they are actually needed, and serve a informative error message if packages requested are not installed.

`netcdf4` is however a special case for which we intentionally trigger the import mechanism at yt's startup time:

https://github.com/yt-project/yt/blob/4a3c266cb2216a103069754b828e40971087a988/yt/utilities/on_demand_imports.py#L85-L94

An existing (minor) issue with this workaround is that it doesn't work if `h5py` was imported before `yt`.
A more worrisome one is that, were `h5py` be promoted back to a hard requirement to `yt` (which has been proposed and discussed multiple times recently), this solution would simply not work, because we'd have to give up control over the import order.

Now, a couple naive observations:
- the ticket on `xarray` was closed almost 3 years ago
- travis-ci isn't something we use
- fundamentally, the bug isn't ours to solve

I don't have a deep understanding of the lock issue that motivated this patch, so I may be wrong in my conclusions, but from where I'm standing it looks like we could just stop making `netcdf4` a special case among our optional dependencies.

I believe this issue needs to be resolved if we ever want to promote `h5py` to a hard requirement, but otherwise there's no rush.

ping @chrishavlin @jzuhone

neutrinoceros:
## PR Summary
Alternative to #4461, closes #4451

### DIFF
--- a/yt/utilities/on_demand_imports.py
+++ b/yt/utilities/on_demand_imports.py
@@ -83,16 +83,6 @@ def safe_import(func):
 
 
 class netCDF4_imports(OnDemand):
-    def __init__(self):
-        # this ensures the import ordering between netcdf4 and h5py. If h5py is
-        # imported first, can get file lock errors on some systems (including travis-ci)
-        # so we need to do this before initializing h5py_imports()!
-        # similar to this issue https://github.com/pydata/xarray/issues/2560
-        try:
-            import netCDF4  # noqa F401
-        except ImportError:
-            pass
-
     @safe_import
     def Dataset(self):
         from netCDF4 import Dataset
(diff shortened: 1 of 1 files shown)


yt/utilities/on_demand_imports.py
```diff
@@ -299,6 +289,16 @@ def optimize(self):
 
 
 class h5py_imports(OnDemand):
```
matthewturk, line 291: So I believe that this is correct behavior, but one question I have is if there are any known side-effects of the import of netCDF4.  Does it register anything with h5py (like plugins etc)?

matthewturk (approved):

yt/utilities/on_demand_imports.py
neutrinoceros: To the best of my knowledge, it doesn't

PR #4465, opened 2023-06-02
BUG: Turn range into a list for keyframes

matthewturk:
This fixes a minor issue with the `Keyframe` class.


neutrinoceros:
good catch ! I think there's a similar bug on line 140 too

neutrinoceros (approved):

PR #4469, opened 2023-06-03
DEPR: remove default value for stacklevel in issue_deprecation_warning, fix missing arguments in call sites

### LINKED ISSUE #4468: DEPR: some deprecation warnings don't have `stacklevel` properly set
Everyone (me included) seems to forget about setting/checkin the `stacklevel` argument in `issue_deprecation_warning`. While the default value (2) makes sense in many simple situations, it seems too easy to forget. I think the number of call sites is low enough that I can remove the default value and just grep my way into fixing all of them.

neutrinoceros:
## PR Summary
closes #4468


neutrinoceros:
@yt-fido test this please

matthewturk (approved):

PR #4470, opened 2023-06-04
BUG: Update GIZMO frontend to handle newer GIZMO versions

mtryan83:
I'd like to suggest two changes to the GIZMO frontend to fix what I believe is a potential bug and to handle updates to the hdf5 interface (this was mentioned on the Slack #help channel in December(?) but I no longer have access to that post and can't reference it). I'm starting with a draft because a) technically these are two separate issues and could go into two separate PR's, but the first is so minor it seemed unnecessary and b) I'm not sure I've resolved the 2nd issue in the proper yt fashion.

1. dmetal check - GIZMO runs with METALS flag enabled but without COOL_METAL_LINES_BY_SPECIES, GALSF_FB_FIRE_RPROCESS, GALSF_FB_FIRE_AGE_TRACERS, or STARFORGE_FEEDBACK_TRACERS enabled will have a per-particle metallicity field (loaded as dmetal) with shape `(NUM_PARTICLES, )`.  Thus, the `fh[dmetal].shape[1]` call will fail with an IndexError.  I've added a check to ensure `fh[dmetal]` is 2d first (`fh[dmetal].ndim>1`).

2. Overriding Gadget _parse_parameter_file - newer versions of GIZMO have changed some of the header fields (described uner Reading Snapshots in the GIZMO doc) used by yt for determining the cosmology of the simulation. I've copied this function from the Gadget/data_structures.py with modifications. Now it checks if the ComovingIntegrationOn flag exists - if so, use that to determine if this is a cosmological run and pull cosmological parameters using updated names. Otherwise, fall back to checking for OmegaLambda and setting cosmological flag from that. Copying the whole function seems overkill though, is there a better way to do this?

Creating tests for both bugs should be straightforward (I think they'd be similar to the `test_gizmo_64` test), but it doesn't look like the sample datasets are made with the newer GIZMO versions and I'm not sure how to provide any. I'll figure out the tests if that can get resolved. 

## PR Checklist

### DIFF
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -16,7 +16,7 @@ class GizmoDataset(GadgetHDF5Dataset):
     @classmethod
     def _is_valid(cls, filename, *args, **kwargs):
         need_groups = ["Header"]
-        veto_groups = ["FOF", "Group", "Subhalo"]
+        veto_groups = ["Config", "Constants", "FOF", "Group", "Subhalo"]
         valid = True
         valid_fname = filename
         # If passed arg is a directory, look for the .0 file in that dir
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -73,13 +73,9 @@ class GizmoDataset(GadgetHDF5Dataset):
 
         self.cosmological_simulation = 1
 
-        try:
-            self.current_redshift = hvals["Redshift"]
-        except KeyError:
-            # Probably not a cosmological dataset, we should just set
-            # z = 0 and let the user know
-            self.current_redshift = 0.0
-            only_on_root(mylog.info, "Redshift is not set in Header. Assuming z=0.")
+        self.current_redshift = hvals.get("Redshift", 0.0)
+        if "Redshift" not in hvals;
+            mylog.info("Redshift is not set in Header. Assuming z=0.")
 
         try:
             # In 1d8479, Nov 2020, public GIZMO updated the names of the Omegas
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -81,9 +81,7 @@ class GizmoDataset(GadgetHDF5Dataset):
             # to include an _, added baryons and radiation and added the
             # ComovingIntegrationOn field. ComovingIntegrationOn is always set,
             # but the Omega's are only included if ComovingIntegrationOn is true
-            mylog.debug(
-                "Reading cosmological parameters using post-1d8479 format"
-            )
+            mylog.debug("Reading cosmological parameters using post-1d8479 format")
             self.cosmological_simulation = hvals["ComovingIntegrationOn"]
             if self.cosmological_simulation:
                 self.omega_lambda = hvals["Omega_Lambda"]
@@ -102,9 +100,7 @@ class GizmoDataset(GadgetHDF5Dataset):
             self.cosmological_simulation = self.omega_lambda != 0.0
         else:
             # If these are not set it is definitely not a cosmological dataset.
-            mylog.debug(
-                "No cosmological information found, assuming defaults"
-            )
+            mylog.debug("No cosmological information found, assuming defaults")
             self.omega_lambda = 0.0
             self.omega_matter = 0.0  # Just in case somebody asks for it.
             self.cosmological_simulation = 0
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -76,29 +76,35 @@ class GizmoDataset(GadgetHDF5Dataset):
         if "Redshift" not in hvals:
             mylog.info("Redshift is not set in Header. Assuming z=0.")
 
-        try:
+        if "ComovingIntegrationOn" in hvals:
             # In 1d8479, Nov 2020, public GIZMO updated the names of the Omegas
             # to include an _, added baryons and radiation and added the
             # ComovingIntegrationOn field. ComovingIntegrationOn is always set,
             # but the Omega's are only included if ComovingIntegrationOn is true
-            if "ComovingIntegrationOn" in hvals:
-                self.cosmological_simulation = hvals["ComovingIntegrationOn"]
+            mylog.debug(
+                "Reading cosmological parameters using post-1d8479 format"
+            )
+            self.cosmological_simulation = hvals["ComovingIntegrationOn"]
+            if self.cosmological_simulation:
                 self.omega_lambda = hvals["Omega_Lambda"]
                 self.omega_matter = hvals["Omega_Matter"]
                 self.omega_baryon = hvals["Omega_Baryon"]
                 self.omega_radiation = hvals["Omega_Radiation"]
-                self.hubble_constant = hvals["HubbleParam"]
-            else:
-                # Should still support GIZMO versions prior to 1d8479 too
-                mylog.info(
-                    "ComovingIntegrationOn does not exist, falling back to OmegaLambda",
-                )
-                self.omega_lambda = hvals["OmegaLambda"]
-                self.omega_matter = hvals["Omega0"]
-                self.hubble_constant = hvals["HubbleParam"]
-                self.cosmological_simulation = self.omega_lambda != 0.0
-        except KeyError:
+            self.hubble_constant = hvals["HubbleParam"]
+        elif "OmegaLambda" in hvals:
+            # Should still support GIZMO versions prior to 1d8479 too
+            mylog.info(
+                "ComovingIntegrationOn does not exist, falling back to OmegaLambda",
+            )
+            self.omega_lambda = hvals["OmegaLambda"]
+            self.omega_matter = hvals["Omega0"]
+            self.hubble_constant = hvals["HubbleParam"]
+            self.cosmological_simulation = self.omega_lambda != 0.0
+        else:
             # If these are not set it is definitely not a cosmological dataset.
+            mylog.debug(
+                "No cosmological information found, assuming defaults"
+            )
             self.omega_lambda = 0.0
             self.omega_matter = 0.0  # Just in case somebody asks for it.
             self.cosmological_simulation = 0
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -95,9 +95,14 @@ class GizmoDataset(GadgetHDF5Dataset):
                 self.hubble_constant = hvals["HubbleParam"]
             else:
                 # Should still support old GIZMO versions too
+                only_on_root(
+                    mylog.info, 
+                    "ComovingIntegrationOn does not exist, falling back to OmegaLambda"
+                )
                 self.omega_lambda = hvals["OmegaLambda"]
                 self.omega_matter = hvals["Omega0"]
-                self.hubble_constant = self.omega_lambda == 0.0
+                self.hubble_constant = hvals["HubbleParam"]
+                self.cosmological_simulation = self.omega_lambda == 0.0
         except KeyError:
             # If these are not set it is definitely not a cosmological dataset.
             self.omega_lambda = 0.0
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -42,9 +42,14 @@ class GizmoDataset(GadgetHDF5Dataset):
             valid = all(ng in fh["/"] for ng in need_groups) and not any(
                 vg in fh["/"] for vg in veto_groups
             )
-            dmetal = "/PartType0/Metallicity"
-            if dmetal not in fh or (fh[dmetal].ndim > 1 and fh[dmetal].shape[1] < 11):
-                valid = False
+            # From Apr 2021, 7f1f06f, public gizmo includes a header variable
+            # GIZMO_version, which is set to the year of the most recent commit
+            # We should prefer this to checking the metallicity, which might 
+            # not exist
+            if "GIZMO_version" not in fh["/Header"].attrs:
+                dmetal = "/PartType0/Metallicity"
+                if dmetal not in fh or (fh[dmetal].ndim > 1 and fh[dmetal].shape[1] < 11):
+                    valid = False
             fh.close()
         except Exception:
             valid = False
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -132,7 +132,7 @@ class GizmoDataset(GadgetHDF5Dataset):
             )
         self.parameters = hvals
 
-        prefix = os.path.join(self.directory, self.basename.split('.', 1)[0])
+        prefix = os.path.join(self.directory, self.basename.split(".", 1)[0])
 
         if hvals["NumFiles"] > 1:
             self.filename_template = f"{prefix}.%(num)s{self._suffix}"
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -96,7 +96,7 @@ class GizmoDataset(GadgetHDF5Dataset):
                 self.omega_lambda = hvals["OmegaLambda"]
                 self.omega_matter = hvals["Omega0"]
                 self.hubble_constant = hvals["HubbleParam"]
-                self.cosmological_simulation = self.omega_lambda == 0.0
+                self.cosmological_simulation = self.omega_lambda != 0.0
         except KeyError:
             # If these are not set it is definitely not a cosmological dataset.
             self.omega_lambda = 0.0
@@ -110,7 +110,6 @@ class GizmoDataset(GadgetHDF5Dataset):
                 "and OmegaLambda is 0.0), so we are turning off Cosmology.",
             )
             self.hubble_constant = 1.0  # So that scaling comes out correct
-            self.cosmological_simulation = 0
             self.current_redshift = 0.0
             # This may not be correct.
             self.current_time = hvals["Time"]
--- a/yt/frontends/gizmo/data_structures.py
+++ b/yt/frontends/gizmo/data_structures.py
@@ -135,15 +135,7 @@ class GizmoDataset(GadgetHDF5Dataset):
         prefix = os.path.join(self.directory, self.basename.split('.', 1)[0])
 
         if hvals["NumFiles"] > 1:
-            for t in (
-                f"{prefix}.%(num)s{self._suffix}",
-                f"{prefix}.gad.%(num)s{self._suffix}",
-            ):
-                if os.path.isfile(t % {"num": 0}):
-                    self.filename_template = t
-                    break
-            else:
-                raise RuntimeError("Could not determine correct data file template.")
+            self.filename_template = f"{prefix}.%(num)s{self._suffix}"
         else:
             self.filename_template = self.parameter_filename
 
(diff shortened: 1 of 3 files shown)


jzuhone:
Hi @mtryan83, this is most welcome! I'll try to have a look  at this in the next few days.

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+            # but the Omega's are only included if ComovingIntegrationOn is true
+            if "ComovingIntegrationOn" in hvals:
+                self.cosmological_simulation = hvals["ComovingIntegrationOn"]
+                self.omega_lambda = hvals["Omega_Lambda"]
+                self.omega_matter = hvals["Omega_Matter"]
+                self.omega_baryon = hvals["Omega_Baryon"]
+                self.omega_radiation = hvals["Omega_Radiation"]
+                self.hubble_constant = hvals["HubbleParam"]
+            else:
+                # Should still support old GIZMO versions too
```
neutrinoceros, line 97: Same here: "old" will not be meaningful in the future.

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+                only_on_root(
+                    mylog.info,
+                    "ComovingIntegrationOn does not exist, falling back to OmegaLambda",
+                )
```
neutrinoceros, line 101: see my comment about logger calls. (I see this is done throughout the method so I'll avoid repeating the comment again)

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+                self.omega_lambda = hvals["OmegaLambda"]
+                self.omega_matter = hvals["Omega0"]
+                self.hubble_constant = hvals["HubbleParam"]
+                self.cosmological_simulation = self.omega_lambda == 0.0
+        except KeyError:
+            # If these are not set it is definitely not a cosmological dataset.
+            self.omega_lambda = 0.0
+            self.omega_matter = 1.0  # Just in case somebody asks for it.
```
neutrinoceros, line 109: non-cosmological datasets normally have `omega_matter` set to `0.0`. Is there a reason not to follow the convention here ?

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+            self.cosmological_simulation = 0
+            # Hubble is set below for Omega Lambda = 0.
+
+        # Since new versions of GIZMO output the ComovingIntegrationOn flag
```
neutrinoceros, line 113: again, please be specific

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+            else:
+                # Should still support old GIZMO versions too
+                only_on_root(
+                    mylog.info,
+                    "ComovingIntegrationOn does not exist, falling back to OmegaLambda",
+                )
+                self.omega_lambda = hvals["OmegaLambda"]
+                self.omega_matter = hvals["Omega0"]
+                self.hubble_constant = hvals["HubbleParam"]
+                self.cosmological_simulation = self.omega_lambda == 0.0
```
neutrinoceros, line 105: `self.omega_lambda` is being set on line 102 and overwritten here, I assume one of the two assignments is superfluous ?

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+            # If these are not set it is definitely not a cosmological dataset.
+            self.omega_lambda = 0.0
+            self.omega_matter = 1.0  # Just in case somebody asks for it.
+            self.cosmological_simulation = 0
+            # Hubble is set below for Omega Lambda = 0.
+
+        # Since new versions of GIZMO output the ComovingIntegrationOn flag
+        # we will rely on its presence to determine if this is a cosmological
+        # dataset.
+        if not self.cosmological_simulation:
```
neutrinoceros, line 116: there seem to be some redundancy with previous lines here.

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+                self.current_time,
+            )
+        self.parameters = hvals
+
+        prefix = os.path.abspath(
+            os.path.join(
+                os.path.dirname(self.parameter_filename),
+                os.path.basename(self.parameter_filename).split(".", 1)[0],
+            )
+        )
```
neutrinoceros, line 151: The base `Dataset` class has a bunch of useful properties that allow to write this kind of code in a much more condensed way.

This should be equivalent
```suggestion
        prefix, _ = os.path.splitext(ds.filename)
```
*unless* the filename has more than 1 `'.'`, in which case *this* should be the replacement
```suggestion
        prefix = os.path.join(ds.directory, ds.basename.split('.', 1)[0])
```

(note that both my suggestions have a minor difference in behaviour with respect to your original code, in that they will correctly resolve `~` as the user's `$HOME` dir)

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+
+        if hvals["NumFiles"] > 1:
+            for t in (
+                f"{prefix}.%(num)s{self._suffix}",
+                f"{prefix}.gad.%(num)s{self._suffix}",
+            ):
+                if os.path.isfile(t % {"num": 0}):
+                    self.filename_template = t
+                    break
```
neutrinoceros, line 160: the `for/else` construct might be obscure to some readers. I'm not sure it is justified here, given that this is equivalent to a simple `if/elif/else`

mtryan83:
> Thank you for working on this ! I agree that this work qualifies as a fix and could in principle be included in a bugfix release, but it needs to retain backward compatibility with previous versions of GIZMO. I see you attempted to satisfy this requirement already, but failing tests seem to indicate that it's not working yet.
> 
> Anyway, I know it's still in draft mode, but here are some preliminary remarks and questions.

So looking at the some of the failing tests (the non gizmo tests), it looks like the problem is files that were previously only being considered an OWLS dataset or an EAGLE dataset are now also being considered GIZMO datasets. I don't know enough about OWLS and EAGLE formats to know if that's correct or not. Is that a test problem or I'm not distinguishing GIZMO files well enough?

For the gizmo tests,  like `yt.frontends.gizmo.tests.test_outputs.PixelizedProjectionValues_snap_N64L16_135_all_('gas', 'density')_0_None` where/how is that even run?

Lastly, regarding most of your comments, the _parse_parameter_file function was copied directly from `gadget/data_structures.py`. I then made a few minimal changes (basically the stuff in the `try` block, lines 85-105, and the `if not self.cosmological_simulation` condition+message, line 116-121 ) so that it could use the updated stuff.  So most of these other changes should back-ported (side-ported?) to the Gadget reader too, but I would assume that's out of scope for this PR.

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+        try:
+            self.current_redshift = hvals["Redshift"]
+        except KeyError:
+            # Probably not a cosmological dataset, we should just set
+            # z = 0 and let the user know
+            self.current_redshift = 0.0
+            only_on_root(mylog.info, "Redshift is not set in Header. Assuming z=0.")
+
+        try:
+            # The current version of GIZMO has updated the names of the Omegas
```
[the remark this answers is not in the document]
mtryan83: Makes sense, I'll see if I can find exactly when things changed. I'm not sure there are any actual version numbers though, so it might just be a date/commit hash.

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +54,113 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+
+        self.cosmological_simulation = 1
+
+        try:
+            self.current_redshift = hvals["Redshift"]
+        except KeyError:
+            # Probably not a cosmological dataset, we should just set
+            # z = 0 and let the user know
+            self.current_redshift = 0.0
+            only_on_root(mylog.info, "Redshift is not set in Header. Assuming z=0.")
```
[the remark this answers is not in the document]
mtryan83: No idea about the parallel-ness. This is just how the gadget reader does it :shrug: .  Also, the default value should match the message. Can you update the suggestion?

yt/frontends/gizmo/data_structures.py
mtryan83: Not sure what you mean? Line 105 just sets `self.cosmological_simulation` based on the value of `self.omega_lambda`. Unless I got the python wrong...

yt/frontends/gizmo/data_structures.py
mtryan83: I was updating the gadget comment. I'll just remove it instead, it should be self-explanatory now

yt/frontends/gizmo/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: Better than nothing. A date could also work.

yt/frontends/gizmo/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: Updated. I'm almost certain that this pattern is unnecessary now, maybe @jzuhone has opinions there.

yt/frontends/gizmo/data_structures.py
neutrinoceros: Oh I must have read that wrong earlier, I read it as
```python
self.cosmological_simulation = self.omega_lambda = 0.0
```
when it's really
```python
self.cosmological_simulation = self.omega_lambda == 0.0
```
carry on !

neutrinoceros:
> So most of these other changes should be back-ported (side-ported?) to the Gadget reader too, but I would assume that's out of scope for this PR.

agreed to consider it out of scope, but good to know !

neutrinoceros:
note that image test failures are unrelated (#4540), and should pass next time this branch is updated

yt/frontends/gizmo/data_structures.py
mtryan83: Just fyi, you're the one who set up that structure :laughing: (#3899). Since the point of this was to add Gadget 2 file template support, I'm inclined to just revert the original change.

yt/frontends/gizmo/data_structures.py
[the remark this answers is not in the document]
mtryan83: I've implemented the change. Note that there was a typo (`;` instead of `:`) this is fixed in 835b6b2

yt/frontends/gizmo/data_structures.py
mtryan83: I've implemented the change. Note I believe that there was a typo (`ds` instead of `self`) this is changed in 835b6b2

mtryan83:
The macos test failure doesn't look like it's real.  

Also, since I have once again been reminded why tests are important, how should I go about writing one for this case?  I think it can just be a few unit tests, but obviously none of the sample datasets would trigger this.

neutrinoceros:
> The macos test failure doesn't look like it's real.

Indeed, it's just flaky. Rerunning fixed it.

neutrinoceros:
As for testing, I would advise adding a new sample dataset if you can manage.

yt/frontends/gizmo/data_structures.py
```diff
@@ -49,3 +53,89 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+            else:
+                # Should still support GIZMO versions prior to 1d8479 too
+                mylog.info(
+                    "ComovingIntegrationOn does not exist, falling back to OmegaLambda",
+                )
+                self.omega_lambda = hvals["OmegaLambda"]
+                self.omega_matter = hvals["Omega0"]
+                self.hubble_constant = hvals["HubbleParam"]
+                self.cosmological_simulation = self.omega_lambda != 0.0
+        except KeyError:
```
neutrinoceros, line 100: Large try/except statements are more prone to be the source of subtle bugs. Would it make sense to just avoid it ?
```suggestion
        if "ComovingIntegrationOn" in hvals:
            # In 1d8479, Nov 2020, public GIZMO updated the names of the Omegas
            # to include an _, added baryons and radiation and added the
            # ComovingIntegrationOn field. ComovingIntegrationOn is always set,
            # but the Omega's are only included if ComovingIntegrationOn is true
            self.cosmological_simulation = hvals["ComovingIntegrationOn"]
            self.omega_lambda = hvals["Omega_Lambda"]
            self.omega_matter = hvals["Omega_Matter"]
            self.omega_baryon = hvals["Omega_Baryon"]
            self.omega_radiation = hvals["Omega_Radiation"]
            self.hubble_constant = hvals["HubbleParam"]
        elif "OmegaLambda" in hvals:
            # Should still support GIZMO versions prior to 1d8479 too
            mylog.info(
                "ComovingIntegrationOn does not exist, falling back to OmegaLambda",
            )
            self.omega_lambda = hvals["OmegaLambda"]
            self.omega_matter = hvals["Omega0"]
            self.hubble_constant = hvals["HubbleParam"]
            self.cosmological_simulation = self.omega_lambda != 0.0
        else:
```

mtryan83: Sure. Though since the omega's aren't guaranteed, I'll need to account for that.

yt/frontends/gizmo/data_structures.py
```diff
@@ -38,9 +42,16 @@ def _is_valid(cls, filename, *args, **kwargs):
...
+            # From Apr 2021, 7f1f06f, public gizmo includes a header variable
+            # GIZMO_version, which is set to the year of the most recent commit
+            # We should prefer this to checking the metallicity, which might
+            # not exist
+            if "GIZMO_version" not in fh["/Header"].attrs:
+                dmetal = "/PartType0/Metallicity"
+                if dmetal not in fh or (
+                    fh[dmetal].ndim > 1 and fh[dmetal].shape[1] < 11
+                ):
+                    valid = False
```
mtryan83, line 54: Can I get a second opinion on this @neutrinoceros @jzuhone ? I ran into this while generating a very simple sample dataset which doesn't for example have cooling/metallicity turned on (the basic zeldovich pancake test) but wasn't sure this is actually something to worry about.

neutrinoceros: I'm not sure I understand the question here. Are you worried you might accidentally invalidate gizmo files ?

mtryan83: I suppose, yes, but more that I wasn't sure that checking for GIZMO_version was sufficient. Basically, if we have GIZMO_version, is that all we need to say this is a valid gizmo file? Or should it check for something else (like dmetal) as well?

neutrinoceros: > Basically, if we have GIZMO_version, is that all we need to say this is a valid gizmo file?

I suppose it should be a sufficient (but not necessary) validator, yes.
What's important here is that this function returns `True` for any known gizmo dataset (which if I'm reading your first comment correctly isn't the case at the moment ?) and `False` for every known non-Gizmo dataset, where by "known" I mean "datasets that are used in CI + any additional dataset you can get your hands on". Basically if this ever returns `False` for a datafile you know is from GIZMO, it's a bug :)

jzuhone:
> So most of these other changes should be back-ported (side-ported?) to the Gadget reader too, but I would assume that's out of scope for this PR.

I agree that it's out of scope, but given the significant amount of code duplication here in `_parse_parameter_file` it would be nice if we could do something about it soon. The best way to make sure we do it correctly whenever that happens is to add a new test dataset as @neutrinoceros suggested which uses the new functionality here and we can help you through getting it up on https://yt-project.org/data and getting a test written, which is probably just going to be a small answer test like the existing ones. A small-ish dataset, if possible, would be best.

mtryan83:
I'm working on writing a test now, using a single snapshot from the gizmo Zeldovich pancake test problem (~17 MB).  

For the `_is_valid` portion, should I use 2 files, one with `GIZMO_version` and one without? They would be identical otherwise (so having two seems like a bit of a waste), and would also be testing the metallicity code and the cosmology code. If I only use one file, I'm not sure how to test both the `GIZMO_version` code and the metallicity code without modifying the hdf5 file inside the test function, which seems like a bad idea.

neutrinoceros:
20Mb is small enough that we can afford more than one file, especially for such a critical part of the logic. I say go for it.

mtryan83:
[snapshot_076_no_gizver.tar.gz](https://github.com/yt-project/yt/files/12074487/snapshot_076_no_gizver.tar.gz)
[snapshot_076_wi_gizver.tar.gz](https://github.com/yt-project/yt/files/12074489/snapshot_076_wi_gizver.tar.gz)
Ok, I've added a new test function: `test_gizmo_zeldovich` which requires test files gizmo_zeldovich/snapshot_076_XX_gizver.hdf5 where XX is either wi or no.  I've attached them separately here (I couldn't fit them in one file, combined they were 3MB over the file limit).  

For reference, the files can be generated using the gizmo zeldovich pancake test problem with the additional config flags COOLING and METAL. The _no_gizver then has the GIZMO_version header field manually removed. Snapshot 076 was picked randomly.

mtryan83:
Just FYI, as of right now the only "failing" test is the new test I've developed, which is failing due to the missing files. What do I need to do to get them added as test datasets @neutrinoceros ?

neutrinoceros:
Sorry I meant to reply earlier and forgot. There's a whole process to get the files permanently stored on our server, and it starts with a PR to the yt-project/website repository. I advise you take a look at previous PRs there. I'll have more time to provide more details tomorrow if need be !

mtryan83:
I just put in a [PR](https://github.com/yt-project/website/pull/121) for the website. Is this the right process?

neutrinoceros:
So far, so good 😊 
Note: I'm on vacation for a couple days. I'll pick it up when I come back unless someone else does it while I'm gone !

neutrinoceros:
alright, I see John took care of your companion PR. Now what we need is to do in *this* PR is to update `yt/sample_data_registry.json` with the new entries and check that we're able to load them through `yt.load_sample`
After that it should be all good.

yt/frontends/gizmo/tests/test_outputs.py
```diff
@@ -32,6 +35,32 @@ def test_gizmo_64():
...
+@requires_file(zeld_wg)
+@requires_file(zeld_ng)
+def test_gizmo_zeldovich():
+    """
+    Test loading a recent gizmo snapshot that doesn't have cooling/metallicity
+
+    The gizmo_zeldovich file has no metallicity field on the gas particles
+    but is a cosmological dataset run using GIZMO_version=2022. There are two
+    versions of the file, with GIZMO_version (_wg) and without GIZMO_version
+    (_ng). Check that both load as gizmo datasets and correctly pull the
```
neutrinoceros, line 48:
```suggestion
    versions of the file, with GIZMO_version (_wi_gizver) and without GIZMO_version
    (_no_gizver). Check that both load as gizmo datasets and correctly pull the
```

neutrinoceros:
I do not know why the test is still failing with `FileNotFoundError` despite the file being correctly hosted on the website server. Maybe further action is required to install it on the machine running Jenkins CI ? @Xarthisius can you have a look please ?

Xarthisius:
@yt-fido test this please.

neutrinoceros (approved):
LGTM !
@jzuhone, take your time if you still want to review this, otherwise feel free to merge !

jzuhone (approved):

PR #4475, opened 2023-06-06
ENH: avoid implicitly dropping default parameters in plot annotation methods

neutrinoceros:
## PR Summary

Many plot annotation methods have 'hidden' internal defaults parameters that may be overriden by users, but passing a single argument (or even an empty dict `plot_args={}`) results in dropping *all* default parameters.

For example
```python
import yt
p.annotate_timestamp()
p.save("/tmp/1.png")
```

```python
p.clear_annotations()
p.annotate_timestamp(text_args={"horizontalalignment": "left"})
p.save("/tmp/2.png")
```

(user wants to override `horizontalalignement`, but the colour is also changed !)

In methods that have more than a couple default parameters, this means that the slightest mutation from default behaviour may require a large effort from the user just to replicate the internal defaults.
This doesn't have to be so: we can instead compose parameter sets from internal defaults and user-provided ones.

This is an attempt to implement this composition interface across all plot annotation methods.
Special care is needed to test this because matplotlib may emit warnings about unused parameters with certain combinations. I'll leave this as a draft until I'm confident this is handled properly.

### DIFF
--- a/yt/visualization/_mpl_imports.py
+++ b/yt/visualization/_mpl_imports.py
@@ -1,10 +1,6 @@
 # these imports are very expensive so we delay them to until they are requested
-
-import matplotlib
 from matplotlib.backend_bases import FigureCanvasBase
 from matplotlib.backends.backend_agg import FigureCanvasAgg
 from matplotlib.backends.backend_pdf import FigureCanvasPdf
 from matplotlib.backends.backend_ps import FigureCanvasPS
 from matplotlib.backends.backend_svg import FigureCanvasSVG
-
-matplotlib.rc("contour", negative_linestyle="solid")


neutrinoceros:
Now fixes #4459.
Explanation: to my knowledge, `ContourCallback` is the only chunk of code in yt that cares about contour parameters.
Now, adding `negative_linestyles="solid"` to the internal default parameters of this method is tricky, because it's only used if `linestyle=None`, and matplotlib will emit warnings about unused parameters, and we want to avoid that. I think this was the original motivation for mutating matplotlib's config at runtime instead.
With this PR, `linestyles="solid"` is *always* passed down to matplotlib's `contour` method (unless overridden explicitly by the user), so nothing else is required to make negative contours use solid line too.

neutrinoceros:
I'm not sure this PR can be considered to be a bugfix as a whole, because it isn't meant to be 100% backward compatible
and in any case it's not urgent to backport, so for now I'm aiming it at the next feature release.

matthewturk:
I think I've changed my mind.  This is a bugfix.  And, it fixes a pretty important set of behavior.

matthewturk (approved):
I do really like `**(plot_args or {})` for some reason.

neutrinoceros:
alright let's backport this bad boy then

@meeseeksdev backport to yt-4.2.x

PR #4477, opened 2023-06-06
allow user to set the thickness of the coordinate lines in VR

zingale:
## PR Checklist

### DIFF
--- a/yt/visualization/volume_rendering/scene.py
+++ b/yt/visualization/volume_rendering/scene.py
@@ -4,7 +4,6 @@ from collections import OrderedDict
 from typing import List, Optional
 
 import numpy as np
-
 from yt.config import ytcfg
 from yt.funcs import mylog
 from yt.units.dimensions import length  # type: ignore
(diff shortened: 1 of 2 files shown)


zingale:
grrr... my editor perturbed some source...

zingale:
pre-commit.ci autofix

zingale:
Here's the old (now default) rendering:

galaxy0030_vr_coords

and here's an alternate, using `sc.annotate_axes(alpha=0.01, thick=3)`

galaxy0030_vr_coords_new

neutrinoceros (commented):
A couple minor comments:
- `thickness` might be more meaningful as a parameter name than `thick`
- I'd rather we don't allow instantiation of `CoordinateVectorSource` with all-positional arguments (i.e. `CoordinateVectorSource(None, 1.0, 1)`). In other words I suggest making `thick`/`thickness` a *keyword-only* argument.
- `CoordinateVectorSource`'s docstring needs to be updated

Otherwise that's a great accessibility feature, thank you !

zingale:
thanks for the comments!  I've updated it now.

yt/visualization/volume_rendering/scene.py
```diff
@@ -845,7 +844,7 @@ def annotate_axes(self, colors=None, alpha=1.0, thick=1):
             ``alpha`` is ignored.
         alpha : float, optional
             The opacity of the vectors.
-        thick : int, optional
+        thickness : int, optional
```
neutrinoceros, line 847: does it have to be an int, or would a float work too ?

neutrinoceros (dismissed):

neutrinoceros:
pre-commit.ci autofix

PR #4481, opened 2023-06-07
add the ability to set the transfer function label fontsize

zingale:
## PR Checklist


matthewturk (approved):
I like it!

zingale:
Here's the script for the above:
```
import yt
from yt.visualization.volume_rendering.api import Scene, create_volume_source

# Load the dataset.
vol = create_volume_source(ds.all_data(), field=("gas", "density"))
sc = Scene()
sc.add_source(vol)
cam = sc.add_camera(ds, lens_type="perspective")
cam.resolution = (1920, 1280)

# You may need to adjust the alpha values to get a rendering with good contrast
# For annotate_domain, the fourth color value is alpha.


# Draw a coordinate axes triad
sc.annotate_axes(alpha=0.01, thickness=1)
sc.save_annotated(f"{ds}_vr_coords.png", sigma_clip=4)
```

zingale:
pre-commit.ci autofix

yt/visualization/volume_rendering/scene.py
```diff
@@ -364,6 +364,7 @@ def save_annotated(
         self,
         fname: Optional[str] = None,
         label_fmt: Optional[str] = None,
+        label_fontsize: int = 10,
```
neutrinoceros, line 367: We can't add allowed-positional arguments to public functions in arbitrary places in the signature because it might break existing user code. Since everything here is currently allowed-positional, any new argument must be appended at the end. And because allowed-positional are a nightmare to deprecate if they ever need to be, let's make it keyword-only.

neutrinoceros (changes requested):

yt/visualization/volume_rendering/scene.py
zingale: I don't really use type hints, so I messed this up.  Does this mean I should add `Optional[]`?

neutrinoceros: No, your type hint is perfect (unless floats are also valid, in which case the annotation should reflect it). For reference, `Optional[T]` means "can be `T` or `None`". It's not needed here because the default value already has the proper type.

What I'm advocating for here is making new arguments keyword-only.

zingale: updated

yt/visualization/volume_rendering/transfer_functions.py
```diff
@@ -603,7 +603,15 @@ def y_format(y, pos):
...
-        self, resolution, log_scale, ax, label=None, label_fmt=None, *, size=10
+        self,
+        resolution,
+        log_scale,
+        ax,
+        *,
+        label=None,
+        label_fmt=None,
+        label_fontsize=10,
+        size=10,
```
neutrinoceros, line 614: `label` and `label_fmt` are currently allowed-positional, so making them suddenly keyword-only is a breaking change
```suggestion
        self,
        resolution,
        log_scale,
        ax,
        label=None,
        label_fmt=None,
        *,
        label_fontsize=10,
        size=10,
```

yt/visualization/volume_rendering/scene.py
```diff
@@ -369,6 +369,8 @@ def save_annotated(
         sigma_clip: Optional[float] = None,
         render: bool = True,
         tf_rect: Optional[List[float]] = None,
+        *,
+        label_fontsize: int = 10,
```
neutrinoceros, line 373: given that it ends up being passed to `matplotlib.axes.Axes.tick_params`, the proper type is actually `Union[float, str]` (`'large'` also works) (see https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.tick_params.html)

yt/visualization/volume_rendering/scene.py
```diff
@@ -399,6 +401,9 @@ def save_annotated(
         label_fmt : str, optional
            A format specifier (e.g., label_fmt="%.2g") to use in formatting
            the data values that label the transfer function colorbar.
+        label_fontsize : int, optional
```
neutrinoceros, line 404: the docstring should mention what the default value is.

zingale: updated

yt/visualization/volume_rendering/scene.py
zingale: updated

yt/visualization/volume_rendering/transfer_functions.py
zingale: updated

zingale:
pre-commit.ci autofix

yt/visualization/volume_rendering/scene.py
neutrinoceros: ... and now it should be synced with the type annotation.
It'd be worth linking to https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.tick_params.html#matplotlib.axes.Axes.tick_params
too, so users get first hand documentation for what strings mean in this context.
But maybe we could just type this as `float` and hide this implementation detail.

(sorry that it's taking so many iterations, I promise we'll converge after this one)

neutrinoceros: I know a couple docstring where urls to matplotlib doc would be very useful, but here it's not crucial. Thanks !

yt/visualization/volume_rendering/transfer_functions.py
```diff
@@ -686,7 +694,7 @@ def y_format(y, pos):
         ax.set_xlim(0.0, max_alpha)
         ax.get_xaxis().set_ticks([])
         ax.set_ylim(visible[0].item(), visible[-1].item())
-        ax.tick_params(axis="y", colors="white", size=10)
+        ax.tick_params(axis="y", colors="white", labelsize=label_fontsize)
```
neutrinoceros, line 697: Took me long enough (sorry again) but I just realised that you're exposing the *tick* label size, not the size of the colorbar label itself (in fact, it's already exposed as the `size` parameter). I think the parameter would be better named `tick_label_size`.

I also see now that the actual colorbar label size is automatically scaled according to the resolution, so wouldn't it make sense to do the same thing with tick labels so their size, relative to the colorbar label, would be resolution-independent ?
There's still value in exposing both parameters independently as you're doing here, so we can keep the bulk of your patch as is, but we'd need remove references to special strings from matplotlib.

matthewturk:
I think we can probably accept the patch as-is and supply additional patches if truly necessary.  This is a useful piece of functionality and I don't want us to get too strict.

neutrinoceros:
I'd be happy to make the change I'm proposing as a follow up PR *if it wasn't going to break what's introduced here*. I can however propose to make the patch myself within this PR

yt/visualization/volume_rendering/transfer_functions.py
zingale: all of the label parameters (this would make 3) are for the tick labels.  So if you change this one, all 3 should be changed.

neutrinoceros: ah, I see. Let's keep the name then. What about auto-scaling ?

matthewturk:
I think we should try to keep the number of iterations *as low as possible*, which I know you also agree with -- and to that end, compromise on some things that will have minimal impact on the potential or actual end users.  (It's hard because I too don't always know exactly the right end point or where things will go when I start reviewing!)  I don't think we need to get into it, because I think we're probably in agreement, and I just wanted to note that I think that balance is something that in this particular case we should continue to pay attention to.

zingale:
I'd prefer to deal with the autoscaling in a different PR.  I'm indifferent to it.

neutrinoceros:
Ok, I'm willing to do the follow up PR but I might not be able to guarantee 100% stability with respect to this one doing so (I expect you won't *need* to use your new, still valuable, keyword argument after that). Does this sound acceptable to you ?

zingale:
oh, I think I misunderstood you.  It is important to be able to set the size of the variable name and the size of the labels (what you call ticks) independently.

neutrinoceros:
Agreed, that's why I still think your patch has value in itself. But getting legible label size by default would also help, wouldn't it ?

matthewturk:
This looks pretty good to me as is, and I'm not sure I understand the disagreement.

neutrinoceros:
I apologies for how long it took me to circle back here. It's become hard for me to return to the mindset in which I wrote my last review, and I don't want to block this any longer.
Resolving the merge conflict should be sufficient to get this in.

neutrinoceros (approved):

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

PR #4496, opened 2023-06-10
ENH: avoid unnecessary array copies through np.ndarray.astype where possible

neutrinoceros:
## PR Summary

`np.ndarray.astype` returns a *copy* of an array by default. In many cases, copies can be avoided by using the target datatype from the start.
This is a thorough, yet conservative refactor: I've stayed away from cases where precision was truncated *after* a calculation, so as to not introduce new floating point errors.

Additionally, I've changed a couple places where we were casting to `"int"` or `int` to use `"int64"` instead, which is identical on POSIX platform but improve consistency on Windows, where `astype("int")` might be equivalent to `astype("int32")` (I couldn't find a reference to back this up, so I may be misremebering).

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -153,7 +153,7 @@ class YTStreamline(YTSelectionContainer1D):
         for mi, (i, pos) in enumerate(zip(pids, self.positions[points_in_grid])):
             if not points_in_grid[i]:
                 continue
-            ci = ((pos - grid.LeftEdge) / grid.dds).astype("int")
+            ci = ((pos - grid.LeftEdge) / grid.dds).astype("int64")
             if grid.child_mask[ci[0], ci[1], ci[2]] == 0:
                 continue
             for j in range(3):
@@ -1581,7 +1581,7 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
             # How many root cells do we occupy?
             end_index = np.rint(cell_end).astype("int64")
             dims = end_index - start_index + 1
-        return start_index, end_index.astype("int64"), dims.astype("int32")
+        return start_index, end_index, dims.astype("int32")
 
     def _update_level_state(self, level_state):
         ls = level_state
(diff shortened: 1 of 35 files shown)


yt/data_objects/construction_data_containers.py
```diff
@@ -153,7 +153,7 @@ def _get_cut_mask(self, grid):
         for mi, (i, pos) in enumerate(zip(pids, self.positions[points_in_grid])):
             if not points_in_grid[i]:
                 continue
-            ci = ((pos - grid.LeftEdge) / grid.dds).astype("int")
+            ci = ((pos - grid.LeftEdge) / grid.dds).astype("int64")
```
matthewturk, line 156: This particular one doesn't necessarily need promotion to int64 (although I recognize numpy may do it internally) because it's a relative offset.

matthewturk (commented):
This requires some careful review which I intend to give it.  Thank you!

neutrinoceros:
No need to review this before I'm done stabilising tests, but thank you for showing interest in it !

yt/data_objects/construction_data_containers.py
neutrinoceros: this doens't change behaviour, I'm just making the size explicit rather than implicit.

neutrinoceros:
force-pushed to resolve merge conflicts

matthewturk:
I think this looks good, and I've read through it and think it's fine.  I would like another pass at it before it gets merged, but I think it's on the right track.

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@matthewturk any chance you could have a look soon ?

yt/utilities/amr_kdtree/amr_kdtree.py
```diff
@@ -128,7 +128,7 @@ def check_tree(self):
             nre = self.ds.arr(node.get_right_edge(), units="code_length")
             li = np.rint((nle - gle) / dds).astype("int32")
```
matthewturk, line 129: outside of scope for this PR but my eyebrows raised a little seeing this casting to int32 instead of int64.

matthewturk (approved):
LGTM, and I don't see any behavior changes to worry about.

yt/utilities/amr_kdtree/amr_kdtree.py
neutrinoceros: yeah, fortunately these variables don't seem to be actually used (except for that one `assert np.all(dims > 0)` which is runtime-skippable)

PR #4497, opened 2023-06-10
[WIP] ENH: Cartesian fields in spherical geometry

matthewturk:
## PR Summary

This adds some cartesian fields (`cartesian_x` etc) to the spherical coordinate handler.

Here is a script that plots what these look like:

```python
import yt
import yt.testing

c = ds.domain_center

ds.r[c[0],:,:].plot("cartesian_x").set_log("all", False).save()
ds.r[c[0],:,:].plot("cartesian_y").set_log("all", False).save()
ds.r[c[0],:,:].plot("cartesian_z").set_log("all", False).save()

ds.r[:,c[1],:].plot("cartesian_x").set_log("all", False).save()
ds.r[:,c[1],:].plot("cartesian_y").set_log("all", False).save()
ds.r[:,c[1],:].plot("cartesian_z").set_log("all", False).save()

ds.r[:,:,c[2]].plot("cartesian_x").set_log("all", False).save()
ds.r[:,:,c[2]].plot("cartesian_y").set_log("all", False).save()
ds.r[:,:,c[2]].plot("cartesian_z").set_log("all", False).save()
```

The images that result (which I will attach momentarily to this PR) look correct, I *think*, except where the AMR regions interact.  In those areas, I'm much less certain, as it's *possible* that they are correct because we are doing cell-centering and the slicing cuts oddly, but I'm genuinely not sure because they do also look out of place.  It would be very helpful to have some other thoughts on that.

### DIFF
--- a/yt/geometry/coordinates/spherical_coordinates.py
+++ b/yt/geometry/coordinates/spherical_coordinates.py
@@ -81,6 +81,53 @@ class SphericalCoordinateHandler(CoordinateHandler):
             units="code_length",
         )
 
+        self.setup_cartesian_fields(registry)
+
+    def setup_cartesian_fields(self, registry):
+        def _cartesian_x(field, data):
+            return (
+                data[("index", "r")]
+                * np.sin(data[("index", "theta")])
+                * np.cos(data[("index", "phi")])
+            )
+
+        def _cartesian_y(field, data):
+            return (
+                data[("index", "r")]
+                * np.sin(data[("index", "theta")])
+                * np.sin(data[("index", "phi")])
+            )
+
+        def _cartesian_z(field, data):
+            return data[("index", "r")] * np.cos(data[("index", "theta")])
+
+        registry.add_field(
+            ("index", "cartesian_x"),
+            sampling_type="local",
+            function=_cartesian_x,
+            units="code_length",
+            display_field=True,
+            take_log=False,
+        )
+
+        registry.add_field(
+            ("index", "cartesian_y"),
+            sampling_type="local",
+            function=_cartesian_y,
+            units="code_length",
+            display_field=True,
+            take_log=False,
+        )
+
+        registry.add_field(
+            ("index", "cartesian_z"),
+            sampling_type="local",
+            function=_cartesian_z,
+            units="code_length",
+            display_field=True,
+            take_log=False,
+        )
+
     def pixelize(
         self, dimension, data_source, field, bounds, size, antialias=True, periodic=True
     ):


matthewturk:
Slices along phi (i.e., a half-circle that runs along the z axis) of the three coordinate values:

AMRGridData_Slice_phi_cartesian_x
AMRGridData_Slice_phi_cartesian_y
AMRGridData_Slice_phi_cartesian_z

Slices along theta (i.e., a conical section) of the three coordinate values:

AMRGridData_Slice_theta_cartesian_x
AMRGridData_Slice_theta_cartesian_y
AMRGridData_Slice_theta_cartesian_z

Slices along r (i.e., a 'surface') of the three coordinate values:

AMRGridData_Slice_r_cartesian_x
AMRGridData_Slice_r_cartesian_y
AMRGridData_Slice_r_cartesian_z

I'd really appreciate thoughts from @neutrinoceros and @jzuhone about these.

matthewturk:
(and @chrishavlin if he has the time!)

neutrinoceros:
This looks great, and is absolutely useful, thank you so much !
I note that plot bounds seem to be off in phi-normal and theta-normal slice plots, which has probably nothing to do with this PR but unless you set them manually to be that way, that would be a bug. Can you include the script you wrote to generate these ?

yt/geometry/coordinates/spherical_coordinates.py
```diff
@@ -81,6 +81,53 @@ def _path_phi(field, data):
             units="code_length",
         )
 
+        self.setup_cartesian_fields(registry)
+
+    def setup_cartesian_fields(self, registry):
```
neutrinoceros, line 86: this doesn't seem to be an existing method name in other CoordinateHandler classes and I can't see what we get from splitting this into a separate method, is there any benefit I'm not seeing ?

yt/geometry/coordinates/spherical_coordinates.py
```diff
@@ -81,6 +81,53 @@ def _path_phi(field, data):
...
+                * np.sin(data[("index", "theta")])
+                * np.sin(data[("index", "phi")])
+            )
+
+        def _cartesian_z(field, data):
+            return data[("index", "r")] * np.cos(data[("index", "theta")])
+
+        registry.add_field(
+            ("index", "cartesian_x"),
+            sampling_type="local",
```
neutrinoceros, line 106: why "local", and not "cell" ?

matthewturk:
@neutrinoceros yup, script is at top.

yt/geometry/coordinates/spherical_coordinates.py
matthewturk: I'm okay with it being in the same function. I was using this for debugging, and wasn't sure if we wanted to make it a property of the superclass.

neutrinoceros:
sorry I missed it ! definitely a bug, I will look into this !

chrishavlin:
This is very useful! I was kind of struggling to understand what the plots **should** show exactly... so I started playing with using these fields to interpolate values to a regular cartesian grid: [link to notebook](https://github.com/chrishavlin/yt_scratch/blob/master/notebooks/general_scratch/cartesian_from_sphere.ipynb). Seems to work pretty well!

A question: How would you feel about adding the same fields for geogrpahic datasets?  the `GeographicCoordinateHandler.setup_fields` already adds `theta`, `phi` and `r` fields. Not exactly elegant, but just copy/pasting the fields you've added here over in the geopraphic handler should work. I don't really see a more elegant solution without refactoring the geographic handler significatnly (to e.g., inherit from the spherical handler).

PR #4498, opened 2023-06-10
BUG: fix internal divide-by-zero warnings

neutrinoceros:
## PR Summary

I am (perhaps naively) thinking this may be a low hanging fruit, let's see how many tests actually raise this warning and if they're fixable.


neutrinoceros:
This turned out very light-weighted, so we might as well aim it to the next bugfix release

matthewturk (approved):
Looks good to me.  I always feel like the 'replace this value with this other value' operation in numpy has a purpose-specific function associated with it, but I can't find one that does this, so.

neutrinoceros:
I don't know of one either.

PR #4500, opened 2023-06-11
BUG: fix plot bounds for YTSlice.plot in non-cartesian coordinates

### LINKED ISSUE #4499: BUG: plot bounds are off with `ds.r[...].plot` in non-cartesian geometries
### Bug report

**Bug summary**

This was discovered in #4497, which I reproduce here for self-consistency

**Code for reproduction**

```python
import yt
import yt.testing

c = ds.domain_center
ds.r[:,c[1],:].plot("Density").save("/tmp/test.png")
```

**Actual outcome**

test

**Expected outcome**

The output should look identical to
```python
import yt
import yt.testing

c = ds.domain_center
yt.SlicePlot(ds, "theta", "Density").save("/tmp/test2.png")
```
test2

neutrinoceros:
## PR Summary

Resolves the main issue within #4499.
I started with commenting out most of the existing, cartesian-only logic in this function, but I expect I'll need to refine this approach to pass every test. I'll keep this as a draft until then.
Hopefully this remains a light-weight patch that can be backported.

### DIFF
--- a/yt/data_objects/selection_objects/slices.py
+++ b/yt/data_objects/selection_objects/slices.py
@@ -116,21 +116,46 @@ class YTSlice(YTSelectionContainer2D):
         return pw
 
     def plot(self, fields=None):
-        if hasattr(self._data_source, "left_edge") and hasattr(
-            self._data_source, "right_edge"
+        from yt.geometry.api import Geometry
+        from yt.visualization.plot_window import get_window_parameters
+
+        geometry = self.ds.geometry
+        if (
+            geometry is Geometry.POLAR
+            or geometry is Geometry.CYLINDRICAL
+            or geometry is Geometry.SPHERICAL
         ):
-            left_edge = self._data_source.left_edge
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Tests pass, but upon closer inspection, it appear that the function isn't actually tested (not even in cartesian geometries). Closing #4499 requires thorough testing, which will be especially complicated because of known problems with plotting zoomed-in views in spherical geometries (see https://github.com/yt-project/yt/pull/4362)

In order to keep this patch simple, I'll update the PR with a bug-for-bug compatibility layer to make sure we don't break known-working cases.

neutrinoceros:
I want to think a little bit longer about wether this is actually the same expected behavior. Switching to draft for now.

neutrinoceros:
I haven't made my mind up yet, but I don't want this to hold yt 4.2.1

neutrinoceros:
I don't know what should be the preferred approach yet but I've convinced myself this isn't it

PR #4501, opened 2023-06-12
ENH: cleanup empty cache directory after a successful call to load_sample

neutrinoceros:
## PR Summary

Context: internally, `yt.load_sample` uses `pooch` to retrieve data from the hub server. Because data is obtained most often as compressed archives, we first download it to a temporary directory (`yt_download_cache`). This is an implementation detail that should not be directly visible to users.
This PR does two things:
- clean up the directory after it's served its purpose (this doesn't prevent pooch from re-creating it during the same session if needed)
- ~prepend a `.` to its name so its "hidden" in the conventional POSIX sense~

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1516,6 +1516,7 @@ def load_sample(
         _download_sample_data_file,
         _get_test_data_dir_path,
         get_data_registry_table,
+        get_download_cache_dir,
     )
 
     pooch_logger = pooch.utils.get_logger()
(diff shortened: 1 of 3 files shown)


matthewturk:
I don't think we should hide it -- dumping a couple gigs into a hidden directory is not something I *really* want to push.  That being said, does this mean that if I use `yt.load_sample` multiple times in multiple sessions, each time it will download afresh?

neutrinoceros:
> dumping a couple gigs into a hidden directory is not something I really want to push.

You have a point, I'll revert that part.

> does this mean that if I use yt.load_sample multiple times in multiple sessions, each time it will download afresh?

it doesn't ! when the *untared* data is already on disk, `load_sample` doesn't download anything. This dir is only for tarballs, which we already auto-remove when untaring is successful.

matthewturk:
> it doesn't ! when the _untared_ data is already on disk, `load_sample` doesn't download anything. This dir is only for tarballs, which we already auto-remove when untaring is successful.

Great -- for some reason I thought it ran a hash on the tarball to see if it needed to re-download, rather than looking at contents of the directory.

chrishavlin (approved):
looks good to me.

PR #4509, opened 2023-06-14
improve docs for FieldValidator objects

chrishavlin:
This PR:
* moves the docstrings for the `FieldValidator` subclasses up out of their `__init__` methods so they get picked up by the docs build
* adds some details to the docstrings
* adds some minimal typing 
* auto-populates the base `FieldValidator` docstring to point to the available subclasses. 

Background: I was working using one of the `FieldValidator`and noticed that all the docstrings [here](https://yt-project.org/doc/reference/api/yt.fields.derived_field.html#yt.fields.derived_field.FieldValidator) were empty despite their presence in the `__init__` methods. Thought it'd be useful to bump those up so we catch them during the docs build... and then I made an effort to improve the docstrings. 

I'm not **super** familiar with the different validators, so it's possible I misinterpreted the arguments (both the descriptions and any typing I added). 

Also, this includes one small code refactoring as well as a little function to dynamically update the `FieldValidator` docstring -- I'm happy to remove those if it's better to simplify this PR.

### DIFF
--- a/yt/fields/derived_field.py
+++ b/yt/fields/derived_field.py
@@ -493,8 +493,13 @@ class FieldValidator:
     Base class for FieldValidator objects. Available subclasses include:
     """
 
-    # note: the above docstring is dynamically modified with a list of child
-    # classes by _set_field_validator_base_docstring() below.
+    def __init_subclass__(cls, **kwargs):
+        # add the new subclass to the list of subclasses in the docstring
+        class_str = f" :class:`{cls.__name__}`"
+        if ":class:" in FieldValidator.__doc__:
+            class_str = "," + class_str
+        FieldValidator.__doc__ += class_str
+
     pass
 
 
@@ -504,9 +509,9 @@ class ValidateParameter(FieldValidator):
 
     Parameters
     ----------
-    parameters: Union[str, Iterable[str]
+    parameters: str, iterable[str]
         a single parameter or list of parameters to require
-    parameter_values: Optional[dict]
+    parameter_values: dict
         If *parameter_values* is supplied, this dict should map from field
         parameter to a value or list of values. It will ensure that the field
         is available for all permutations of the field parameter.
@@ -564,7 +569,7 @@ class ValidateProperty(FieldValidator):
 
     Parameters
     ----------
-    prop: Union[str, Iterable[str]]
+    prop: str, iterable[str]
         the required property or properties to require
     """
 
@@ -586,10 +591,10 @@ class ValidateSpatial(FieldValidator):
 
     Parameters
     ----------
-    ghost_zones: Optional[int]
+    ghost_zones: int
         If supplied, will validate that the number of ghost zones required
         for the field is <= the available ghost zones. Default is 0.
-    fields: Optional str, tuple(str, str), or any iterable of the previous types.
+    fields: Optional str, tuple[str, str], or any iterable of the previous types.
         The field or fields to validate.
 
     """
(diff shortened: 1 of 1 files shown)


yt/fields/derived_field.py
```diff
@@ -589,3 +625,14 @@ def __call__(self, data):
         if getattr(data, "_type_name", None) == "grid":
             return True
         raise NeedsOriginalGrid()
+
+
+def _set_field_validator_base_docstring():
```
neutrinoceros, line 630: This would be suited for `FiedlValidator.__init_subclass__`

neutrinoceros (commented):
This is a good idea. Here are a couple comments

yt/fields/derived_field.py
```diff
@@ -515,11 +532,17 @@ def __call__(self, data):
 
 
 class ValidateDataField(FieldValidator):
+    """
+    A :class:`FieldValidator` that ensures the output file has a given data field stored
+    in it.
+
+    Parameters
+    ----------
+    field: str, tuple(str, str), or any iterable of the previous types.
```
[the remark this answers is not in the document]
chrishavlin: I think I copy/pasted this from another function, I'll find that one and update it as well.

yt/fields/derived_field.py
chrishavlin: good idea! i'll move it and refactor a bit as needed.

chrishavlin: now in `__init_subclass__`

yt/fields/derived_field.py
```diff
@@ -489,17 +489,39 @@ def __copy__(self):
...
+    """
+
+    def __init_subclass__(cls, **kwargs):
+        # add the new subclass to the list of subclasses in the docstring
+        class_str = f" :class:`{cls.__name__}`"
+        if ":class:" in FieldValidator.__doc__:
+            class_str = "," + class_str
+        FieldValidator.__doc__ += class_str
+
     pass
```
chrishavlin, line 503: oh I guess I could remove this `pass` now

yt/fields/derived_field.py
```diff
@@ -489,17 +489,34 @@ def __copy__(self):
...
-    def __init__(self, parameters, parameter_values=None):
-        """
-        This validator ensures that the dataset has a given parameter.
+    """
+    A :class:`FieldValidator` that ensures the dataset has a given parameter.
 
-        If *parameter_values* is supplied, this will also ensure that the field
+    Parameters
+    ----------
+    parameters: Union[str, Iterable[str]
```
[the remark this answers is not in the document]
chrishavlin: removed all the Union and Optional occurrences. left the iterable though? Should I remove that too? it is nice to note that it could be a str or any iterable with str's. but I could remove the iterable and rephrase the description

yt/fields/derived_field.py
```diff
@@ -493,8 +493,13 @@ class FieldValidator:
...
-    # note: the above docstring is dynamically modified with a list of child
-    # classes by _set_field_validator_base_docstring() below.
+    def __init_subclass__(cls, **kwargs):
+        # add the new subclass to the list of subclasses in the docstring
+        class_str = f" :class:`{cls.__name__}`"
+        if ":class:" in FieldValidator.__doc__:
+            class_str = "," + class_str
+        FieldValidator.__doc__ += class_str
+
     pass
```
neutrinoceros: this `pass` can now be removed

yt/fields/derived_field.py
```diff
@@ -493,8 +493,13 @@ class FieldValidator:
     Base class for FieldValidator objects. Available subclasses include:
     """
 
-    # note: the above docstring is dynamically modified with a list of child
-    # classes by _set_field_validator_base_docstring() below.
+    def __init_subclass__(cls, **kwargs):
+        # add the new subclass to the list of subclasses in the docstring
+        class_str = f" :class:`{cls.__name__}`"
+        if ":class:" in FieldValidator.__doc__:
+            class_str = "," + class_str
```
neutrinoceros, line 500:
```suggestion
        class_str = f":class:`{cls.__name__}`"
        if ":class:" in FieldValidator.__doc__:
            class_str = ", " + class_str
```

neutrinoceros (approved):

PR #4513, opened 2023-06-16
BUG: Fix cm1 chunk reader

matthewturk:
Fixes 4512.

We need to convert chunks to a list so we don't exhaust it the first time we iterate, and we also need to reset offset to 0 every loop over fields.


chrishavlin (approved):
looks good to me!

neutrinoceros:
@meeseeksdev backport to yt-4.2.x

Issue #4379, opened 2023-03-23
BLD: dropping support for matplotlib < 3.5 ?


neutrinoceros:
I would like to propose dropping support for versions of Matplotlib older than 3.5

Motivations:
- MPL is the one dependency for which we entertain version-dependent code (beside Python itself)
- most current version-dependent blocks make a distinction between MPL pre/post 3.5, and some are actually non-functional with older versions, for instance
https://github.com/yt-project/yt/blob/00c3110921c20be9fc8e0ffba124a32ff62e0d95/yt/visualization/base_plot_types.py#L331-L334
- MPL 3.5-3.7 currently represent about 66% of all downloads (source https://pepy.tech/project/matplotlib?versions=3.2.*&versions=3.3.*&versions=3.4.*&versions=3.5.*&versions=3.6.*&versions=3.7.*&display=monthly)

Of course we should not just support the very latest version, but since version 3.2, 3.3 and 3.4 are not actually 100% compatible with yt, it makes more and more sense to just drop them.
Thoughts ? objections ?

matthewturk:
I think before we aggressively drop older versions we need to understand
our specific install base, particularly on HPC systems. We've been a bit
more aggressive with compile -time stuff, but I am not sure I agree with
dropping this yet.

neutrinoceros:
I get that, but I also genuinely wonder if it makes a difference: in what environment is one able to upgrade yt (or build it from source) *and* unable to upgrade matplotlib ?

matthewturk:
I think the main environment that this would happen is when folks rely on a systemwide installation that doesn't include yt.  Which ... isn't that uncommon?  yt is not a universally installed (or even known-of) package!

neutrinoceros:
@jzuhone any inputs on this ? IIRC you only ever install yt from source on clusters

jzuhone:
@neutrinoceros I install yt from source pretty much everywhere, but I also never use any Python that is installed by the sysadmins--I always put conda in my home directory.

neutrinoceros:
@matthewturk what would be the criterion for this to become acceptable to you ? Should it time-based, loosely constant version span (dropping support for older versions as new one are released), something else ?

matthewturk:
I think you convinced me and we can do this.

Issue #4515, opened 2023-06-18
TST: numpy 1.25 breaks CI (via other dependencies)


neutrinoceros:
### Bug report

**Bug summary**

Some tests are currently failing on deprecation warnings new in numpy 1.25, see: https://github.com/yt-project/yt/actions/runs/5303784716

Importantly, none of the deprecated function calls live in yt itself (we already handled our own), but *other* dependencies (h5py, pandas, pyproj via cartopy) use deprecated API. Let's see what's up with them.

- [ ] [h5py 3.9.0 is about to be released](https://github.com/h5py/h5py/pull/2277) and [has compatibity with numpy 1.25](https://github.com/h5py/h5py/pull/2242), so it may not be worth doing anything about it (unless the release drags for more than a couple days).
- [x] [pandas 2.0.2 is already patched](https://github.com/pandas-dev/pandas/pull/53343), but because `pyart` pinned it to `<2.0`, our CI is effectively stuck on version 1.5.3 . [This is already addressed upstream](https://github.com/ARM-DOE/pyart/pull/1430) -> update: resolved https://github.com/ARM-DOE/pyart/releases/tag/v1.15.2
- [ ] [I've opened a patch on cartopy](https://github.com/SciTools/cartopy/pull/2194)

PR #4517, opened 2023-06-18
PERF: delay rarely used imports (netcdf4, importlib, multiprocessing, tarfile, tomllib, tomli_w)

neutrinoceros:
## PR Summary

- `import netCDF4` is responsible for ~6% startup overhead, let's delay it when we can afford it.

- `import importlib.metadata` is about ~3% (we still need to import it to determine matplotlib's version at runtime, which we'll be able to avoid [when support for mpl<3.5 is dropped](https://github.com/yt-project/yt/issues/4379))

- multiprocessing, tarfile, tomllib, tomli_w are (collectively) about ~1%

nothing life-changing here but every % counts (yt's CLI still takes a couple seconds to do *anything*).

### DIFF
--- a/yt/funcs.py
+++ b/yt/funcs.py
@@ -30,11 +30,6 @@ from yt.utilities.exceptions import YTFieldNotFound, YTInvalidWidthError
 from yt.utilities.logger import ytLogger as mylog
 from yt.utilities.on_demand_imports import _requests as requests
 
-if sys.version_info >= (3, 9):
-    import importlib.resources as importlib_resources
-else:
-    import importlib_resources
-
 # Some functions for handling sequences and other types
 
 
@@ -552,6 +547,10 @@ def get_git_version(path):
 
 
 def get_yt_version():
+    if sys.version_info >= (3, 9):
+        import importlib.resources as importlib_resources
+    else:
+        import importlib_resources
     version = get_git_version(os.path.dirname(importlib_resources.files("yt")))
     if version is None:
         return version
@@ -1003,6 +1002,10 @@ def matplotlib_style_context(style="yt.default", after_reset=False):
     """
     # FUTURE: this function should be deprecated in favour of matplotlib.style.context
     # after support for matplotlib 3.6 and older versions is dropped.
+    if sys.version_info >= (3, 9):
+        import importlib.resources as importlib_resources
+    else:
+        import importlib_resources
     import matplotlib as mpl
     import matplotlib.style
 
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -5,15 +5,11 @@ This module gathers all user-facing functions with a `load_` prefix.
 import atexit
 import os
 import sys
-import tarfile
 import time
 import types
 import warnings
-from importlib.metadata import entry_points
-from multiprocessing import Pipe, Process
-from multiprocessing.connection import Connection
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union, cast
 from urllib.parse import urlsplit
 
 import numpy as np
@@ -44,6 +40,8 @@ from yt.utilities.object_registries import (
 )
 from yt.utilities.on_demand_imports import _pooch as pooch, _ratarmount as ratarmount
 
+if TYPE_CHECKING:
+    from multiprocessing.connection import Connection
 # --- Loaders for known data formats ---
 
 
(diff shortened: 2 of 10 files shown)


neutrinoceros:
blocked by #4515

matthewturk:
Do we have a breakdown of import costs?  Maybe it could be a good way to look at splitting stuff up?

neutrinoceros:
I profile startup with tuna (`python -m pip install tuna`)

This command opens an interactive flame graph showing everything that happens on `import yt`
```shell
python -X importtime -c "import yt" 2> import.log && python -m tuna import.log
```
`unyt` (together with `sympy`) still largely dominates the startup time, but we can separate yt from its most expensive dependencies with a more complex invoke:
```shell
python -X importtime -c "import numpy; import matplotlib; import sympy; import unyt; import yt" 2> import.log && python -m tuna import.log
```

I highly recommend running this profile with Python 3.11, and latest versions of sympy, unyt, and cmyt, all of which contain optimisations that make yt's startup slightly faster.

Also note that when switching branch/making changes, the measured startup time include byte code compilation, so results should not be considered realistic on the very first run.


Splitting up modules is unlikely to bring direct gains. Unyt is by far the biggest offender, but I think I've made every possible "easy" optimisations between v2.8 and v2.9.5 already.

neutrinoceros:
> I think I've made every possible "easy" optimisations between v2.8 and v2.9.5 already.

https://github.com/yt-project/unyt/pull/429 👀

neutrinoceros:
currently blocked by https://github.com/yt-project/yt/issues/4540

chrishavlin:
So I've looked at this a few times and I'm feeling on the fence: when it's just one or two uses within a module, nesting those imports inside function calls is OK by me, particularly if it is an import of an external package. Where I'm uncertain is for our internal imports that are used throughout a module: e.g., I think the readability of `parallel_analysis_tools` suffers quite a bit by nesting all those `from yt.data_objects.image_array import ImageArray` and I'd like to know just how much of a speedup we're getting to justify the conflict with our style guide (which is admittedly vague -- maybe this is enough of a "good reason" to not import up top?).

If there are parts of yt that both take a long time to load and are not used in the majority of sessions, then I think it'd be worth looking at a more formal lazy import framework using something like [`importlib.util.LazyLoader`](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader). If the import is something that **is** used in most sessions then I'd personally stick with keeping the import cost up front in the initial `import yt` (e.g., is `ImageArray` ever **not** used when someone uses yt? this is a genuine question). 

So I think I'd add an approval here but would appreciate a bit more discussion before this goes in cause it feels like a shift in coding style -- it seems hard to maintain these gains going forward without recommending nested imports in new code, because it certainly doesn't seem tenable for folks to simply remember which imports should be nested and which don't need it.

neutrinoceros:
Thank you for having a look. I'm okay to drop parts of this PR, as the gains are absolutely not worth any level of controversy :)

#4539 is a much more efficient, and completely different approach, towards the same goal, so if *that* one is okay I'll be happy to re-evaluate the present patch.

chrishavlin:
Well I don't think this strays into controversial :) It's more a practical question of whether the gains you get here will be easily maintained down the line. I think the external imports are easy enough, but not so sure about internal yt imports. The very specific yt imports (like your changes to `plot_modifications`) probably are pretty easy, but it seems very likely that someone would write some new unrelated code at some point that imports `ImageArray` and inadvertently un-delays your delay.

neutrinoceros:
Yeah that's the biggest problem with this technique, it's very easy to ruin hours of effort with just one `from matplotlib import pyplot as plt`, it is much preferable to gain time by reducing side effects. I find that in practice, these are easier to spot when other effects (costly imports) are neutralised.

neutrinoceros:
@chrishavlin so I did my homework: here's a box plot of import times for the main branch against this branch
import

I ran `import yt` 200 times on each version. The first commit by itself seems to correspond to a statistically significant gain, but that isn't true for commits 2 and 3, which I think are the ones that got you on the fence, so I'm happy to drop them if that's satisfactory to you :)

chrishavlin:
Thanks for breaking it down by commit! Ya, I think with just the first commit I'd hit approve right away :)  

Does look like the monkeypatch` adjustment in `yt/tests/test_external_frontends.py` should probably be kept from the 2nd commit though. And I'm OK with keeping the changes to external imports (`matplotlib` and `csv`) from your 2nd/3rd commits, but if they're not actually getting you a speedup, might not be worth picking out those changes to include here.

neutrinoceros:
revamped ! I kept the first commit, and followed your suggestions on what to keep fro the 2nd and 3rd

chrishavlin (approved):
Great! LGTM now!

PR #4519, opened 2023-06-19
DEPR: use builtin DeprecationWarnings

### LINKED ISSUE #4518: DEPR: `VisibleDeprecationWarning` considered noisy
Context:
yt doesn't use builtin `DeprecationWarning`s but instead uses a custom class `VisibleDeprecationWarning`, whose sole intent is to work around Python's default warning filters which hide `DeprecationWarning`s by default (the idea being that they should only be made visible if users opt-in).

The design currently used in yt made sense a while back, because we want deprecation warnings to reach all users, not just maintainers of dependent code with high quality standards. The trade-off being that our warnings will surface to users no matter how many layers of dependencies there are between the code being written and yt; for instance, if `yt_astro_analysis` or `trident` uses deprecated yt API, *their* users don't have much control over it and it can be argued they shouldn't see these "noisy" warnings by default. 

Since Python 3.7 ([PEP 565](https://peps.python.org/pep-0565/)), `DeprecationWarnings` attributed to the `__main__` module are again showed by default, meaning they will reach direct users (including in REPLs) but will not surface (by default) when deprecated API calls live in an intermediate dependency layer.

It seems to me that our current solution isn't necessary anymore and that we should use simple `DeprecationWarning`s from now on.

neutrinoceros:
closes #4518

### DIFF
--- a/yt/_maintenance/deprecation.py
+++ b/yt/_maintenance/deprecation.py
@@ -4,19 +4,6 @@ from types import FunctionType
 from typing import Dict, Optional
 
 
-class VisibleDeprecationWarning(UserWarning):
-    """Visible deprecation warning, adapted from NumPy
-
-    The nose runner does not show users DeprecationWarning.
-    This ensures that a deprecation warning is visible to users
-    if that is desired.
-    """
-
-    # this class becomes useless after the tests are migrated from nose to pytest
-
-    pass
-
-
 def issue_deprecation_warning(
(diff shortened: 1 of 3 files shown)


matthewturk:
makes sense, thanks to the explanation you provided in #4518.

matthewturk (approved):

jzuhone (approved):

PR #4520, opened 2023-06-19
PERF: reduce startup overhead from cmap registration

neutrinoceros:
## PR Summary

Saves about 1% startup overhead by eliminating unnecessary conversions in cmap registration.


matthewturk (approved):

neutrinoceros:
As I'm trying to push this to its limit, I find that the gain is actually not statistically significant.

PR #4521, opened 2023-06-19
DEP: drop support for matplotlib<3.5

### LINKED ISSUE #4379: BLD: dropping support for matplotlib < 3.5 ?

neutrinoceros:
## PR Summary

closes #4379

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -125,12 +125,6 @@ def pytest_configure(config):
             "ignore:pkg_resources is deprecated as an API:DeprecationWarning",
         )
 
-    if MPL_VERSION < Version("3.5.0"):
-        config.addinivalue_line(
-            "filterwarnings",
-            r"ignore:distutils Version classes are deprecated:DeprecationWarning",
-        )
-
     if MPL_VERSION < Version("3.5.2") and PILLOW_VERSION >= Version("9.1"):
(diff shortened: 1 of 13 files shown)


matthewturk:
I'm good with this but want someone else to approve.

jzuhone:
I'm for this.

jzuhone (approved):

PR #4529, opened 2023-06-21
PERF: Use grid tree for counting grid cell inclusion

matthewturk:
Instead of using the grid objects, by default let's use the grid tree to count the number of cells included in a data object.

This is an independent PR, and may have test failures that can be corrected within this PR.  It's independent of #4525, but related.

PR #4530, opened 2023-06-21
PERF: Use bitarray to count grid cells

matthewturk:
This requires #4525, and when used in combination with #4529 will result in using faster code paths for counting the number of cells included in a selector for patch grid datasets.

The place this could particularly result in selection-based speedups would be in ghost zone identification.


neutrinoceros:
Can we make this a draft so as to softly redirect attention to the other, self-contained PRs ?

PR #4538, opened 2023-06-25
PERF: reduce startup overhead from cmap registration

neutrinoceros:
## PR Summary

another, maybe more aggressive attempt at minimising startup overhead from colormaps registration (2nd take following #4520)


yt/visualization/color_maps.py
```diff
@@ -3,23 +3,27 @@
...
     """
     Adds a colormap to the colormaps available in yt for this session
     """
-    # Note: this function modifies the global variable 'yt_colormaps'
-    yt_colormaps[name] = LinearSegmentedColormap(name, cdict, 256)
-    mpl.colormaps.register(yt_colormaps[name])
+    from matplotlib.colors import LinearSegmentedColormap
+
+    issue_deprecation_warning(
+        "yt.visualization.color_maps.add_colormap is deprecated",
```
matthewturk, line 22: what is it being replaced with?

yt/visualization/color_maps.py
```diff
@@ -48,26 +52,19 @@ def register_yt_colormaps_from_cmyt():
...
-        continue
-    cdict = {
-        "red": np.transpose([_vs, v[0], v[0]]),
-        "green": np.transpose([_vs, v[1], v[1]]),
-        "blue": np.transpose([_vs, v[2], v[2]]),
-    }
+for k, v in _cm.color_map_luts.items():
     try:
-        add_colormap(k, cdict)
+        mpl.colormaps.register(ListedColormap(v, name=k, N=256))
```
matthewturk, line 67: I mean, this is slightly flatter, but I don't think it's a big deal, either.

neutrinoceros: yeah, this PR yields disappointing results, but just in case, I want to clarify that the gain here is not expected to come from the *inlining*, but from instantiating `ListedColormap` instead of `LinearSegmentedColormap`, which saves a conversion from the format we actually define our maps in.

neutrinoceros: any way #4538 and #4517 are both much more efficient at reducing the startup time, I really don't mind if we end up just closing this one.

yt/visualization/color_maps.py
```diff
@@ -48,26 +52,19 @@ def register_yt_colormaps_from_cmyt():
     """
 
     for hist_name, alias in _HISTORICAL_ALIASES.items():
-        cmap = mpl.colormaps[alias].copy()
-        cmap.name = hist_name
-        mpl.colormaps.register(cmap)
-        mpl.colormaps.register(cmap.reversed())
+        cmap = mpl.colormaps[alias]
```
neutrinoceros, line 55: note for later that this even if we close this PR, this line change is worth using if only because `mpl.colormaps.__getitem__` *already* returns copies, so the current implementation of this line is unjustified.

neutrinoceros:
I'll drop this (again) because it's far too big a patch for such a small gain. I'll reopen a much smaller PR with just the simplest avoid-double-copies patch (which impacts 2 lines)

PR #4539, opened 2023-06-26
PERF: delay frontends initialization until yt.load is actually called

neutrinoceros:
## PR Summary

Still pursuing my quest of making yt's CLI decently fast. This saves about 50 to 100ms overhead on startup time on my (relatively quick) laptop.


matthewturk (approved):

yt/frontends/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@lru_cache(maxsize=None)
+def __getattr__(value):
+    import importlib
+
+    if value == "_all":
+        for _ in __all__:
+            __getattr__(_)
+        return
```
neutrinoceros, line 52: note: the reason this special case is needed is that nesting `from yt.frontends import *` is syntactically invalid.

yt/frontends/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+]
+
+from functools import lru_cache
+
+
+@lru_cache(maxsize=None)
+def __getattr__(value):
+    import importlib
+
+    if value == "_all":
```
chrishavlin, line 49: maybe cause i'm sleep deprived, but took me a few minutes to realize this was a recursive call and i was very confused for a while lol. Mind adding the following comment (or something similar)?

```suggestion
    if value == "_all":
        # recursively import all frontends
```

yt/frontends/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+        for _ in __all__:
+            __getattr__(_)
+        return
+
+    if value not in __all__:
+        raise AttributeError
```
chrishavlin, line 55: Is it worth adding a custom message here? Or is the message covered somewhere up higher in the stack?

chrishavlin (approved):
Looks good -- a couple very small comments.

yt/loaders.py
```diff
@@ -90,6 +90,8 @@ def load(
     yt.utilities.exceptions.YTAmbiguousDataType
         If the data format matches more than one class of similar specialization levels.
     """
+    from yt.frontends import _all  # type: ignore [attr-defined] # noqa
```
chrishavlin, line 93: Oh and one extra comment/question -- by nesting this import in the `load` and `load_simulation` calls, does that mean one could `import yt` and then directly instantiate a frontend without ever importing all the other frontends? That would actually be a pretty nice consequence beyond simply delaying the imports... might be worth a mention in the docs even. Could make a big difference for batch processing scripts.

neutrinoceros: indeed, I think with this patch you'd be able to short-circuit all frontends being loaded, but I think at this point it should be seen as an implementation detail rather than a feature (in particular because circumventing `yt.load` would require different code if your frontends isn't builtin, but lives in an extension)

yt/frontends/__init__.py
neutrinoceros: sure !

chrishavlin (approved):

neutrinoceros:
@matthewturk this is ready !

Issue #4540, opened 2023-06-26
BUG: new error in cartopy image tests


neutrinoceros:
### Bug report

**Bug summary**

All cartopy image tests have started failing on all PRs today, and choke on the following error
```
ValueError: 'x' must be finite, check for nan or inf values
```

I don't know yet what in the environment changed overnight, but we can exclude cartopy itself as well as its direct dependencies, none of which received a new release since yesterday.

neutrinoceros:
In #4541 I found that the change is that scipy 1.11 now invalidates nans in a kdtree queries (https://github.com/scipy/scipy/pull/18502)

Here's a "minimal" reproducer:

```python
import yt
from yt.testing import fake_amr_ds

ds = fake_amr_ds(geometry="geographic")
field = ("stream", "Density")
sl = yt.SlicePlot(ds, "altitude", field, origin="native")
sl.set_mpl_projection("PlateCarree")
sl.render()
sl.plots[field].figure.savefig("/tmp/test.png")
```

From where I'm standing now, it looks like the invalid data (nans) is generated from within cartopy, in which case the next step would be to try and eliminate yt from the minimal reproducer and see what happens.

neutrinoceros:
here's a yt-less reproducer
```python
import matplotlib.pyplot as plt
import numpy as np
from cartopy import crs as ccrs

data = [[0, 1], [0, 1]]
fig = plt.figure()
ax = fig.add_axes(
    (0, 0, 1, 1),
    projection=ccrs.Mollweide(),
)
ax.imshow(
    data,
    transform=ccrs.PlateCarree(),
)
```

Now I need to now wether it makes sense to combine `Mollweide` projection with `PlateCarree` transform, my gut says "no way" but I'm not an expert. In any case this is what yt is currently doing.

chrishavlin:
In principle there's not a problem with using a `PlateCarree` transform and `Mollweide` projection. in cartopy: 

> The projection argument is used when creating plots and determines the projection of the resulting plot (i.e. what the plot looks like). The transform argument to plotting functions tells Cartopy what coordinate system your data are defined in.

I suspect those nan's have always been there, and it's the scipy change that's exposing it now. I'll have some time this afternoon to look closely!

neutrinoceros:
Thank you ! If you think this usage is legitimate, we should probably upstream to cartopy. I can do it in half an hour

chrishavlin:
is your yt-less reproducer throwing an error for you?

[...]

PR #4541, opened 2023-06-26
TST: try pinning scipy in image tests

neutrinoceros:
## PR Summary

Context: #4540

Scipy 1.11 was released 16hrs ago, so it's my number one suspect at the moment.
This PR is an attempt to narrow it down.


neutrinoceros:
Seems like this indeed fixes image tests, though it's not clear wether the problem is a pure regression in scipy, or an incompatibility in cartopy. Until we know who to upstream this discussion to, I advise not the merge this.

chrishavlin (approved):
Looks good, but can you open a new issue to track the cartopy issue? Or not close the existing one until we see some discussion upstream? I don't think we want to pin scipy in yt's dependencies, since this bug will only affect those with the latest scipy and using cartopy (so this PR is totally sufficient as is)... but we might want to pin cartopy to the newest version once they release a fix for the issue.

neutrinoceros:
Fair point, let's unlink the issue for now

PR #4544, opened 2023-06-27
BUG: Fix docstring for load octree

### LINKED ISSUE #4542: BUG: Octree loading with 1 zone not functioning at present
This comes from the `load_octree` docstring:

```python
import yt
import numpy as np
oct_mask = np.zeros(25)
oct_mask[[0,  5,  7, 16]] = 8
octree_mask = np.array(oct_mask, dtype=np.uint8)
quantities = {}
quantities["gas", "density"] = np.random.random((22, 1))
bbox = np.array([[-10.0, 10.0], [-10.0, 10.0], [-10.0, 10.0]])

ds_octree = yt.load_octree(
    octree_mask=octree_mask,
    data=quantities,
    bbox=bbox,
    num_zones=1,
    partial_coverage=0,
)

s = yt.SlicePlot(ds_octree, "x", "density")
s.save()
```

At present it isn't working; I am working on a fix.  I believe it's related to the `nzones != 2` PR, but I'm not certain.

matthewturk:
Fixes #4542.

Note that this did not require modifying the source, just the docstring.

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1096,11 +1096,11 @@ def load_octree(
     -------
 
     >>> import numpy as np
-    >>> oct_mask = np.zeros(25)
+    >>> oct_mask = np.zeros(33) # 5 refined values gives 7 * 4 + 5 octs to mask
     ... oct_mask[[0,  5,  7, 16]] = 8
     >>> octree_mask = np.array(oct_mask, dtype=np.uint8)
     >>> quantities = {}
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Is this the expected output ?
OctreeData_Slice_x_density

matthewturk:
Yes.  See related bug #4545 , which addresses why this is the case.

neutrinoceros (approved):

PR #4547, opened 2023-06-27
BUG: Fix initialization of max level in load_octree

### LINKED ISSUE #4545: BUG: min_level and max_level are not reliably set
Some of the handlers are not correctly setting `max_level`, which then propagates to things like `ds.r[:]`.  For instance:

```python
import yt
import numpy as np
oct_mask = np.zeros(33)
oct_mask[[0,  5,  7, 16]] = 8
octree_mask = np.array(oct_mask, dtype=np.uint8)
quantities = {}
quantities["gas", "density"] = np.random.random((29, 1))
bbox = np.array([[-10.0, 10.0], [-10.0, 10.0], [-10.0, 10.0]])

ds_octree = yt.load_octree(
        octree_mask=octree_mask,
        data=quantities,
        bbox=bbox,
        num_zones=1,
        partial_coverage=0,
)
#ds_octree.max_level = 10

print(ds_octree.r[:]["ones"])
```

This does not actually return anything!  The array is empty, because the `_selector` that is created has a `max_level` that it inherits from the `ds_octree`, which is set to 0.  Setting it to any value at or greater than the maximum level will work to get all the values.

This is kind of a big deal I think, and probably worth being a blocker.  I don't know when this was introduced.

matthewturk:
Fixes #4545.

We don't store the level trait for an Oct as an attribute of the Oct itself, so it has to be collected during the initial traversal.  This adds it to the `load_octree` command via the `LoadOctree` visitor, and then it sets an attribute on the dataset.  I looked and this only causes problems for the stream octrees.

This fixes an issue wherein loading an octree then results in just the root oct being selected regardless of the selection method.

### DIFF
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -866,6 +866,8 @@ class StreamOctreeHandler(OctreeIndex):
             "partial_coverage": self.ds.partial_coverage,
         }
         self.oct_handler = OctreeContainer.load_octree(header)
+        # We do now need to get the maximum level set, as well.
+        self.ds.max_level = self.oct_handler.max_level
 
     def _identify_base_chunk(self, dobj):
         if getattr(dobj, "_chunk_info", None) is None:
--- a/yt/frontends/stream/tests/test_stream_octree.py
+++ b/yt/frontends/stream/tests/test_stream_octree.py
@@ -1,4 +1,5 @@
 import numpy as np
+from numpy.testing import assert_equal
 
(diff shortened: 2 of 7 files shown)


neutrinoceros (approved):

neutrinoceros:
I confirm this fixes the issue, but I have ~0 expertise in this area of the code so I'd like to keep this open for a day or two so others can take a look if they want to.

neutrinoceros (approved):

Issue #3992, opened 2022-06-29
API (discussion): using keyword-only arguments


neutrinoceros:
In my reviews,I've been encouraging other devs to use keyword-only arguments. Since the actual rationale for this isn't obvious, and at @jzuhone's request (#3989), here's a conversation starter around using them more often in yt.

### What are keyword-only arguments ?

In short, any argument defined after `*` cannot be passed positionally
```python
def foo(a, b, *, c):
    pass

foo(a=1, b=2, c=3)  # valid
foo(1, 2, 3)        # invalid
foo(1, 2, c=3)      # valid
```

Keyword-only arguments were introduced in Python 3.0, hence, we could not use them before we dropped Python 2 support ~2 years ago. This is why most of the code base doesn't use this syntax at all.

See [PEP 3102](https://peps.python.org/pep-3102/) for more details.

### So what's wrong with allowing arguments to be passed positionally ?

The following observations are only relevant for user-exposed functions.
There are two main issues

1) it allows users to write very obscure code, which may be hard for anyone (include themselves) to debug. To illustrate, here's a perfectly valid application of `yt.SlicePlot`
```python
slc = yt.SlicePlot(ds, "z", ("gas", "density"), [0, 1, 0], (10, 100), (0.23, 0.5, 'domain'), "cm", True, 12)
```
Even to an expert user, it might not be obvious what any of the arguments beyond `("gas", "density")` are, and one could easily mix them up, resulting in potentially very confusing errors.

Of course this example isn't *sane*, most users *would* rather write it this way
```python
src = yt.SlicePlot(ds, "z", ("gas", "density"),
    center=[0, 1, 0],
    width=(10, 100),
    origin=(0.23, 0.5, 'domain'),
    axes_units="cm",
    right_handed=True,
    fontsize=12
)
```
and this is also how we consistently document this kind of complex API, for good reasons.
Still, yt has enough users that a similar situation is very likely to happen, most often to beginner users who don't know any better (yet).

2) Even if discouraged, the first example is _valid_, so maintaining backward compatibility requires that it _stays_ valid in the future. Now what happens if one of the arguments gets deprecated ? Removing it would necessarily be a backward incompatible change, because the order of all following arguments would change too ! The only exception being the very last(s) argument(s).
This situation is _not_ hypothetical. It already happened and _continues_ to happen because many arguments were deprecated during the yt 3.x dev cycle that are not yet removed.

### What does it get us ?

[...]

PR #4556, opened 2023-06-30
option to include min/max limits on the color field for export_sketchfab

evanoconnor:
## PR Summary

For exporting surfaces to 3D objects there are a few options, for example `export_blender` and `export_sketchfab.`  `export_sketchfab` does not include the ability to limit the range of the color field (`export_blender` does).  This PR brings that functionality to `export_sketchfab` through optional arguments.

## PR Checklist

- [ x ] New features are documented, with docstrings and narrative docs

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -2495,10 +2495,12 @@ class YTSurface(YTSelectionContainer3D):
             color_log,
             sample_type,
             color_field_max=color_field_max,
-            color_field_min=color_field_min
+            color_field_min=color_field_min,
         )
 
-    def _color_samples(self, cs, color_log, color_map, color_field_max, color_field_min, arr):
+    def _color_samples(
+        self, cs, color_log, color_map, color_field_max, color_field_min, arr
+    ):
         if color_log:
             cs = np.log10(cs)
         if color_field_min is None:
@@ -2587,7 +2589,9 @@ class YTSurface(YTSelectionContainer3D):
             f.write(b"property uchar blue\n")
             v = np.empty(self.vertices.shape[1], dtype=vs)
             cs = self.vertex_samples[color_field]
-            self._color_samples(cs, color_log, color_map, color_field_max, color_field_min, v)
+            self._color_samples(
+                cs, color_log, color_map, color_field_max, color_field_min, v
+            )
         else:
             v = np.empty(self.vertices.shape[1], dtype=vs[:3])
         line = "element face %i\n" % (nv / 3)
@@ -2600,7 +2604,9 @@ class YTSurface(YTSelectionContainer3D):
             # Now we get our samples
             cs = self[color_field]
             arr = np.empty(cs.shape[0], dtype=np.dtype(fs))
-            self._color_samples(cs, color_log, color_map, color_field_max, color_field_min, arr)
+            self._color_samples(
+                cs, color_log, color_map, color_field_max, color_field_min, arr
+            )
         else:
             arr = np.empty(nv // 3, np.dtype(fs[:-3]))
         for i, ax in enumerate("xyz"):
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -2499,7 +2499,14 @@ class YTSurface(YTSelectionContainer3D):
         )
 
     def _color_samples(
-        self, cs, color_log, color_map, color_field_max, color_field_min, arr
+        self,
+        cs,
+        color_log,
+        color_map,
+        arr,
+        *,
+        color_field_max=None,
+        color_field_min=None,
     ):
         if color_log:
             cs = np.log10(cs)
@@ -2590,7 +2597,12 @@ class YTSurface(YTSelectionContainer3D):
             v = np.empty(self.vertices.shape[1], dtype=vs)
             cs = self.vertex_samples[color_field]
             self._color_samples(
-                cs, color_log, color_map, color_field_max, color_field_min, v
+                cs,
+                color_log,
+                color_map,
+                v,
+                color_field_max=color_field_max,
+                color_field_min=color_field_min,
             )
         else:
             v = np.empty(self.vertices.shape[1], dtype=vs[:3])
@@ -2605,7 +2617,12 @@ class YTSurface(YTSelectionContainer3D):
             cs = self[color_field]
             arr = np.empty(cs.shape[0], dtype=np.dtype(fs))
             self._color_samples(
-                cs, color_log, color_map, color_field_max, color_field_min, arr
+                cs,
+                color_log,
+                color_map,
+                arr,
+                color_field_max=color_field_max,
+                color_field_min=color_field_min,
             )
         else:
             arr = np.empty(nv // 3, np.dtype(fs[:-3]))
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -2437,6 +2437,8 @@ class YTSurface(YTSelectionContainer3D):
         color_field=None,
         color_map=None,
         color_log=True,
+        color_field_max=None,
+        color_field_min=None,
         sample_type="face",
         no_ghost=False,
     ):
@@ -2457,6 +2459,10 @@ class YTSurface(YTSelectionContainer3D):
             Which color map should be applied?
         color_log : bool
             Should the color field be logged before being mapped?
+        color_field_max : float
+            Maximum value of the color field across all surfaces.
+        color_field_min : float
+            Minimum value of the color field across all surfaces.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


yt/data_objects/construction_data_containers.py
```diff
@@ -2437,6 +2437,8 @@ def export_ply(
         color_field=None,
         color_map=None,
         color_log=True,
+        color_field_max=None,
+        color_field_min=None,
```
neutrinoceros, line 2441: new arguments cannot be added in arbitrary places, you need to put them at the end of the signature, and preferably behind a `*,` (to make them keyword only). See https://github.com/yt-project/yt/issues/3992 for a more in-depth explanation.

yt/data_objects/construction_data_containers.py
```diff
@@ -2504,6 +2532,8 @@ def _export_ply(
...
```
neutrinoceros, line 2536: same here

yt/data_objects/construction_data_containers.py
```diff
@@ -2599,6 +2629,8 @@ def export_sketchfab(
...
```
neutrinoceros, line 2633: and here

yt/data_objects/construction_data_containers.py
```diff
@@ -2673,6 +2709,8 @@ def export_sketchfab(
             color_field,
             color_map,
             color_log,
+            color_field_max,
+            color_field_min,
```
neutrinoceros, line 2713: aaand here

neutrinoceros:
Please fix the indentation. At the moment your code is not valid Python (see https://results.pre-commit.ci/run/github/88776493/1688125829.Xtb3gkPGTg6Woa7ABHKa8Q)

matthewturk:
@evanoconnor Thank you for this pull request!  I think this is a good addition, and I appreciate you adding it to the code base.  The changes suggested are pretty minor I think, and I'm happy to help out if you'd like to get it across the finish line.

yt/data_objects/construction_data_containers.py
evanoconnor: ok, I see now.  I'm new to the `*,` so some questions.  Should all keyword arguments be after the `*,`, i.e. should I add this even earlier in the signature (I'm guessing these are calls that have not yet been migrated to this new format)

neutrinoceros: In an ideal world, yes ! Unfortunately we cannot do that without breaking someone's code (if anyone is passing these arguments positionally, making them keyword-only after the fact will break them).

yt/data_objects/construction_data_containers.py
evanoconnor: this one is internal (`_export_ply`).  Do you still suggest to make these optional, since it is only called from `export_ply`

neutrinoceros: I agree that it's a less crucial for an internal function, but many projects depend on internal functionalities so I'd prefer to be over-cautious :)

yt/data_objects/construction_data_containers.py
evanoconnor: ok, I think I understand this now

evanoconnor:
ok, I'm a little bit unsure about the error after the last commit [d736bfc].  It worked on my install, but I think I'm missing something about `*,`

yt/data_objects/construction_data_containers.py
```diff
@@ -2675,6 +2714,8 @@ def export_sketchfab(
             color_log,
             sample_type="vertex",
             no_ghost=no_ghost,
+            color_field_max,
+            color_field_min,
```
neutrinoceros, line 2718: @evanoconnor The error you're getting is indeed related to the use of `*,`

> positional argument follows keyword argument

In Python, some arguments may be passed positionally (for instance in `abs(-2)`), and others may be passed by name (as 'keywords') (for instance in `range(0, 10, step=2)`, `step` is passed as keyword). However what's not allowed is passing positional arguments *after* keyword ones.

Here `sample_type` and `no_ghost` are passed as keywords, so it's invalid to pass `color_field_max/min` positionally as is currently done. To fix it, pass these args as keywords:
```suggestion
            color_field_max=color_field_max,
            color_field_min=color_field_min,
```

Is this explanation clear ?

evanoconnor: oh course! sorry.

So, for `_export_ply`:

```
def _export_ply(
        self,
        filename,
        bounds=None,
        color_field=None,
        color_map=None,
        color_log=True,
        sample_type="face",
        *,
        color_field_max=None,
        color_field_min=None,
    ):
```
 there is a `*,` but since the call to it:

```
       self._export_ply(
            filename,
            bounds,
            color_field,
            color_map,
            color_log,
            sample_type,
            color_field_max,
            color_field_min
        )
```
 didn't specify any keyword arguments it was fine? even with the `*,`? Would you suggest adding 

```
            color_field_max=color_field_max,
            color_field_min=color_field_min
```

here too?

neutrinoceros: Yes, that's actually necessary, since the whole point of using `, *` is to _forbid_ passing arguments positionally :)

neutrinoceros:
pre-commit.ci autofix

yt/data_objects/construction_data_containers.py
```diff
@@ -2481,13 +2488,35 @@ def export_ply(
...
 
-    def _color_samples(self, cs, color_log, color_map, arr):
+    def _color_samples(self, cs, color_log, color_map, color_field_max, color_field_min, arr):
         if color_log:
             cs = np.log10(cs)
-        mi, ma = cs.min(), cs.max()
+        if color_field_min is None:
+            cs = [float(field) for field in cs]
+            cs = np.array(cs)
+            mi = cs.min()
```
neutrinoceros, line 2509: I'm not sure I understand this change. What's the gain here, compared to `mi = cs.min()` ?

yt/data_objects/construction_data_containers.py
```diff
@@ -2481,13 +2488,35 @@ def export_ply(
...
+        else:
+            mi = color_field_min
+            if color_log:
+                mi = np.log10(mi)
```
neutrinoceros, line 2513: shouldn't this be applied also in the case `color_field_min is None` ?

yt/data_objects/construction_data_containers.py
```diff
@@ -2481,13 +2488,35 @@ def export_ply(
...
+            color_field,
+            color_map,
+            color_log,
+            sample_type,
+            color_field_max=color_field_max,
+            color_field_min=color_field_min
         )
 
-    def _color_samples(self, cs, color_log, color_map, arr):
+    def _color_samples(self, cs, color_log, color_map, color_field_max, color_field_min, arr):
```
neutrinoceros: Sorry I missed that one in my first round. I would suggest to put the new arguments at the end of the signature here too (ideally making them keyword only, but that's secondary)

neutrinoceros (commented):
Nice progress, now the code seems functional. I think I see a couple ways this could be simplified, though maybe I'm just missing something. Any way, here are a couple questions !

yt/data_objects/construction_data_containers.py
evanoconnor: no, since the field itself is logged above, then the min is the logged already

yt/data_objects/construction_data_containers.py
evanoconnor: I'm not 100% (this is just a copy of the max/min treatment for the `export_obj` subroutine.  I'm guessing this gets rid of the units though?

neutrinoceros: ah ! if dropping units is the intention here it should probably just be `mi = cs.ndview.min()`, which is not only more compact but also avoid copying data and conversions.

yt/data_objects/construction_data_containers.py
neutrinoceros: fair enough, I missed that !

yt/data_objects/construction_data_containers.py
chrishavlin: I'm fairly baffled by this (and `export_obj`, so this comment isn't directed at you, @evanoconnor :) ). From what i can tell the input `cs` will always be an array? so `cs = [float(field) for field in cs]` a few lines above this will iterate over every element and individually cast to float just to convert back to an array afterwards? that seems not great.... am I missing something?

matthewturk: I believe this is indeed to drop units, and *possibly* it's to do some conversion away from numpy arrays, because of some business with the exporting, but I cannot recall.

evanoconnor: `cs` is used a bit further down to construct the color scale from 0 to 1.  So I guess this then:

```
            cs = np.asarray(cs)
            mi = cs.min()
```

before deciding on the best way here, what is the general practice for units?  should the `color_field_max` and `color_field_max` have units at the onset, that would fail here because `cs` drops the units and `mi` and `ma` would retain them.

neutrinoceros: yt's interface regarding scalar values with units isn't exactly consistent at the moment. In many places, we expect floats where we could in principle support passing quantities.
I think it's fine that `color_field_min` and `color_field_max` only support float values for now. We can always add support for quantities in the future.

Agreed that `cs = np.asarray(cs)` is the most robust and efficient way to drop units.

neutrinoceros (approved):

matthewturk:
@evanoconnor Would you consider adding yourself to the yt paper? https://github.com/yt-project/yt-4.0-paper/#authorship-policy

evanoconnor:
@matthewturk Thanks, but I don't quite think this warrants authorship, and I don't have the time to help with the paper (plus it seems like it is nearing the end!) I am looking forward to have such a resource though!

PR #4560, opened 2023-07-02
TST: switch future-dependencies tests from daily to weekly frequency, add auto-reporting in case of failure

### LINKED ISSUE #3731: ENH: automated issues on failures for bleeding-edge CI ?
A problem with our existing test setup, in particular regarding the bleeding-edge daily workflow, is that, as far as I understand, only the author of the workflow (me) gets notified when a job fails. Other maintainers may not be informed unless I report the issue manually, or they happen to notice that the badge in the README is red.

This has been the situation for a while now, but I just found out that there is a simple workaround this limitation in GitHub Actions.
Indeed it is possible to configure the workflow to open an issue in case of failure. See how xarray does it with their own "upstream dev CI" workflow (which is exactly the same idea as our "bleeding edge" job):

https://github.com/pydata/xarray/blob/main/.github/workflows/upstream-dev-ci.yaml

The one "cons" I can see is that adding this to our workflow increases ever so slightly our maintenance burden if it ever breaks, but this seems worth it. Any objection ?

neutrinoceros:
## PR Summary

Automate reporting of failures in scheduled tests with future dependencies, also reduce their frequency from daily to weekly. This technique is taken from matplotlib. As a test, I deployed it in a personal repo and found that it worked without any further configuration.

closes #3731

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -15,8 +15,8 @@ on:
     paths:
       - .github/workflows/bleeding-edge.yaml
   schedule:
-    # run this every day at 3 am UTC
-    - cron: '0 3 * * *'
+    # run this every Wednesday at 3 am UTC
+    - cron: 0 3 * * 3
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


chrishavlin (approved):
great!!

Issue #4563, opened 2023-07-04
BUG: doc builds are broken


neutrinoceros:
### Bug report

**Bug summary**

Docs builds are currently all failing. For example: https://tests.yt-project.org/job/yt_docs_new/2600

```python-traceback
Traceback (most recent call last):
  File "./helper_scripts/show_fields.py", line 239, in <module>
    fi = getattr(this_f, fi_name)
AttributeError: module 'yt.frontends.gadget' has no attribute 'GadgetFieldInfo'
```

I think the most likely explanation is that it was broken by a PR that didn't directly touch documentation and so failed to trigger a build.

The error bisects to #4539 (which was the obvious suspect). I am able to reproduce the error locally with `python doc/helper_scripts/show_fields.py`.

neutrinoceros:
Renouncing alphabetical order in `yt.frontends.__init__.__all__` to prevent that `arepo` and `eagle` be initialised before `gadget` seems to resolve the issue. I would prefer a more robust solution but that's probably the only one I'll be able to find for now, so I'll open a PR with this patch.

PR #4564, opened 2023-07-05
DOC: hotfix doc builds

neutrinoceros:
## PR Summary

close https://github.com/yt-project/yt/issues/4563

test with:
```shell
$ python doc/helper_scripts/show_fields.py
```


neutrinoceros:
yes; for the record I don't like this solution either, and plan to research it more when I have more time :)

PR #4570, opened 2023-07-09
Fix fake_dataset periodicity attribute in FieldDetector

jzuhone:
## PR Summary

The `FieldDetector` class mocks up a fake dataset from a `defaultdict` if a dataset is not passed. It has to attach a number of basic yt attributes to this `fake_dataset` class, but at the end it tries to use the `force_periodicity` method, which of course exists on the `Dataset` class but not here. I'm just setting the periodicity here by hand, which is the previous behavior and should never have been changed here. 

I ran into this bug on accident--not sure if it affects anything, but this is the right fix for now until we do something better.

## PR Checklist

### DIFF
--- a/yt/fields/field_detector.py
+++ b/yt/fields/field_detector.py
@@ -63,7 +63,7 @@ class FieldDetector(defaultdict):
             ds.domain_left_edge = np.zeros(3, "float64")
             ds.domain_right_edge = np.ones(3, "float64")
             ds.dimensionality = 3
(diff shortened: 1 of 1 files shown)


matthewturk:
I also actually ran into this, but was not certain it was a bug.  Nice fix.

matthewturk (approved):

neutrinoceros (approved):

PR #4580, opened 2023-07-19
TST: run cookbook recipes in a deterministic order

neutrinoceros:
## PR Summary

Extracted from #4575

This is helpful in reproducing tests jobs locally. In particular, when some test hangs the process, it helps decipher which one.
Also improves portability for Windows


neutrinoceros:
also removed some dead code while I'm in there

neutrinoceros:
Let's try treating warnings as errors here so stuff like #4586 and #4585 can be discovered systematically in the future

neutrinoceros:
Looks like treating warnings as errors in recipes is complicated on this machine + Python version (the site module fails to import because of a deprecation warning). Not worth investing more effort here until we've upgraded the environment, so I'll need to revert this part of the PR.

chrishavlin (approved):

PR #4581, opened 2023-07-19
PERF: Add noexcept qualifiers to improve performance on Cython 3

neutrinoceros:
## PR Summary

Extracted from #4575 (patch by @matthewturk)
Partially adresses #4355

### DIFF
--- a/yt/frontends/gamer/cfields.pyx
+++ b/yt/frontends/gamer/cfields.pyx
@@ -7,39 +7,39 @@ cimport numpy as np
 import numpy as np
 
 
-cdef np.float64_t h_eos4(np.float64_t kT, np.float64_t g):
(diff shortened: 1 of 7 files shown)


matthewturk:
Can we instead just get that one in?  I don't want to have my commits subsumed into this PR.

neutrinoceros:
https://github.com/yt-project/yt/pull/4575 is messy because we're both working on it, so rebasing can only happen once. I'd like to avoid destructive history rewriting, which is why I'm proposing to extract known-good bits instead.

matthewturk:
I would prefer we keep the history in this case, even though we are both working on it.  I will also admit I was rather taken aback that you took my commits and compressed them into a new PR that was authored by you without first suggesting or asking, and I think I would like to suggest that step not be taken in the future.

neutrinoceros:
Sorry about that, I meant no offense.

PR #4582, opened 2023-07-19
DOC: fix string interpolation (units double-printing) in particle_filter cookbook recipe

neutrinoceros:
reminiscent of #4443


matthewturk (approved):
good catch

PR #4584, opened 2023-07-19
more SPH field improvements

jzuhone:
## PR Summary

Continuing to find improvements to be made to detection of SPH fields:

* Supporting Gadget fields such as `"ColdFraction"`, `"CloudFraction"`, and `"HotPhaseTemperature"` with yt-like aliases. 
* Support AREPO/TNG fields `"cooling_rate"` and `"cooling_time"`.
* Metal fields can be of `"local"` `sampling_type` or `"particle"` `sampling_type` depending on if they are star or gas particles. 
* The core of the PR is to make detection of metal fields much more flexible for Gadget and its derivatives, accounting for the (somewhat unfortunate) variety of naming schemes for how these fields are stored in files.

Tests, and probably a bit of documentation about how metal fields are detected, are incoming. 

## PR Checklist

### DIFF
--- a/yt/frontends/gadget/fields.py
+++ b/yt/frontends/gadget/fields.py
@@ -1,6 +1,7 @@
 from functools import partial
 
 from yt.fields.particle_fields import sph_whitelist_fields
+from yt.frontends.gadget.definitions import elem_names_opts
 from yt.frontends.sph.fields import SPHFieldInfo
 from yt.utilities.periodic_table import periodic_table
 from yt.utilities.physical_constants import kb, mp
@@ -44,7 +45,7 @@ class GadgetFieldInfo(SPHFieldInfo):
         This gets used with the Gadget group0000 format
         as defined in the gadget_field_specs in frontends/gadget/definitions.py
         """
-        metal_names = ["C", "O", "Si", "Fe"]
+        metal_names = elem_names_opts[4]
 
         def _fraction(field, data, i: int):
             return data[(ptype, "FourMetalFractions")][:, i]
@@ -94,7 +95,7 @@ class GadgetFieldInfo(SPHFieldInfo):
                 )
                 return 1.0 - ret
 
-            elem_names = ["He", "C", "Ca", "O", "N", "Ne", "Mg", "S", "Si", "Fe", "Ej"]
+            elem_names = elem_names_opts[11]
 
         elif fname == "MetalMasses":
             n_elem = len(
@@ -104,28 +105,7 @@ class GadgetFieldInfo(SPHFieldInfo):
                     if fd[0] == ptype and fd[1].startswith("MetalMasses")
                 ]
             )
-            elem_names = {
-                7: ["C", "N", "O", "Mg", "Si", "Fe", "Ej"],
-                8: ["He", "C", "O", "Mg", "S", "Si", "Fe", "Ej"],
-                11: ["He", "C", "Ca", "O", "N", "Ne", "Mg", "S", "Si", "Fe", "Ej"],
-                15: [
-                    "He",
-                    "C",
-                    "Ca",
-                    "O",
-                    "N",
-                    "Ne",
-                    "Mg",
-                    "S",
-                    "Si",
-                    "Fe",
-                    "Na",
-                    "Al",
-                    "Ar",
-                    "Ni",
-                    "Ej",
-                ],
-            }[n_elem]
+            elem_names = elem_names_opts[n_elem]
             no_He = "He" not in elem_names
 
             def _fraction(field, data, i: int):
--- a/yt/frontends/gadget/fields.py
+++ b/yt/frontends/gadget/fields.py
@@ -101,7 +101,6 @@ class GadgetFieldInfo(SPHFieldInfo):
             def _metallicity(field, data):
                 mass = 0.0
                 start_idx = int(not no_He)
-                print(start_idx)
                 for i in range(start_idx, n_elem):
                     mass += data[(ptype, f"MetalMasses_{i:02d}")]
                 ret = mass / data[(ptype, "particle_mass")]
--- a/yt/frontends/gadget/fields.py
+++ b/yt/frontends/gadget/fields.py
@@ -76,6 +76,10 @@ class GadgetFieldInfo(SPHFieldInfo):
             def _fraction(field, data, i: int):
                 return data[(ptype, "ElevenMetalMasses")][:, i] / data[(ptype, "particle_mass")]
 
+            def _metallicity(field, data):
+                ret = data[(ptype, "ElevenMetalMasses")][:,1].sum(axis=1) / data[(ptype, "particle_mass")]
+                return ret
+
             def _h_fraction(field, data):
                 ret = data[(ptype, "ElevenMetalMasses")].sum(axis=1) / data[(ptype, "particle_mass")]
                 return 1.0 - ret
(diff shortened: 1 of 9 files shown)


yt/frontends/gadget/fields.py
```diff
@@ -40,32 +46,116 @@ def _setup_four_metal_fractions(self, ptype):
...
                 sampling_type="particle",
-                function=partial(_Metal_density, i=i),
+                function=partial(_metal_density, i=i),
                 units=self.ds.unit_system["density"],
             )
 
         return metal_names
 
-    def _setup_eleven_metal_masses(self, ptype):
+    def _make_fraction_functions(self, ptype, fname):
```
matthewturk, line 74: Is it possible at all to reuse the code in `periodic_table.py`?

yt/frontends/gadget/fields.py
```diff
@@ -40,32 +46,116 @@ def _setup_four_metal_fractions(self, ptype):
...
+
+            def _metallicity(field, data):
+                mass = 0.0
+                start_idx = int(not no_He)
+                for i in range(start_idx, n_elem):
+                    mass += data[(ptype, f"MetalMasses_{i:02d}")]
+                ret = mass / data[(ptype, "particle_mass")]
+                return ret
+
+            def _h_fraction(field, data):
```
matthewturk, line 145: Just to clarify, this is done this way because it doesn't track H specifically?

matthewturk (approved):

yt/frontends/gadget/fields.py
jzuhone: That's right

matthewturk:
What else are you thinking before moving to ready-to-review?

jzuhone:
@matthewturk tests and docs, sorry, got sidetracked

jzuhone:
This PR now depends on https://github.com/yt-project/website/pull/122 and the corresponding file (already on dickenson) needs to be put on Jenkins (@Xarthisius)

neutrinoceros:
@yt-fido test this please

jzuhone:
@Xarthisius just wanted to ping you again on the above ^^ (sorry for being a nuisance)

Xarthisius:
> @Xarthisius just wanted to ping you again on the above ^^ (sorry for being a nuisance)

Sorry I missed it! It's up now.

jzuhone:
Let's see if the new test passes here (it did on my laptop). If it does, we're ready to go here.

jzuhone:
@yt-fido test this please

jzuhone:
@Xarthisius Jenkins failures appear unrelated?

neutrinoceros:
@jzuhone regular tests are a bit flaky since we switch to Python 3.10, which is what we're seeing here. And from what I understand, doc builds are unstable until #4663 is done.

jzuhone:
@yt-fido test this please

jzuhone:
@yt-fido test this please

jzuhone:
This is ready to go.

yt/frontends/gadget/fields.py
```diff
@@ -40,32 +46,121 @@ def _setup_four_metal_fractions(self, ptype):
...
+                    "Si",
+                    "Fe",
+                    "Na",
+                    "Al",
+                    "Ar",
+                    "Ni",
+                    "Ej",
+                ],
+            }[n_elem]
+            no_He = "He" not in elem_names
```
munkm, line 129: I found the logic of this part a little bit confusing. But I don't have a good alternative suggestion.

yt/frontends/gadget/fields.py
```diff
@@ -40,32 +46,121 @@ def _setup_four_metal_fractions(self, ptype):
...
+                [
+                    fd
+                    for fd in self.ds.field_list
+                    if fd[0] == ptype and fd[1].startswith("MetalMasses")
+                ]
+            )
+            elem_names = {
+                7: ["C", "N", "O", "Mg", "Si", "Fe", "Ej"],
+                8: ["He", "C", "O", "Mg", "S", "Si", "Fe", "Ej"],
+                11: ["He", "C", "Ca", "O", "N", "Ne", "Mg", "S", "Si", "Fe", "Ej"],
```
munkm, line 110: Out of curiosity, is there a possibility that an output will have both `MetalMasses` and `ElevenMetalMasses` fields?

yt/frontends/gadget/fields.py
```diff
@@ -40,32 +46,121 @@ def _setup_four_metal_fractions(self, ptype):
...
+
+        elif fname == "MetalMasses":
+            n_elem = len(
+                [
+                    fd
+                    for fd in self.ds.field_list
+                    if fd[0] == ptype and fd[1].startswith("MetalMasses")
+                ]
+            )
+            elem_names = {
```
munkm, line 107: Maybe moving the `elem_names` dict outside of both field defs and using the same list of elements `elem_names[11]` for both `ElevenMetalMasses` and `MetalMasses` fields would reduce some of the redundancy in your code?

yt/frontends/gadget/tests/test_outputs.py
```diff
@@ -147,3 +154,48 @@ def test_magneticum():
...
+        "N",
+        "Ne",
+        "Mg",
+        "S",
+        "Si",
+        "Fe",
+        "Na",
+        "Al",
+        "Ar",
+        "Ni",
```
munkm, line 185: I noticed your `MetalMasses_` definition didn't include a definition for `14`. Did you want to add that too since that's the use case for your test?

munkm (approved):
This looks good! I have a few minor comments/questions.

yt/frontends/gadget/fields.py
```diff
@@ -40,32 +46,121 @@ def _setup_four_metal_fractions(self, ptype):
...
+                    "Mg",
+                    "S",
+                    "Si",
+                    "Fe",
+                    "Na",
+                    "Al",
+                    "Ar",
+                    "Ni",
+                    "Ej",
+                ],
```
neutrinoceros, line 127: feels like a good place to use `# fmt: skip`

yt/frontends/gadget/fields.py
jzuhone: Not really--both are in fact somewhat arbitrary. `ElevenMetalMasses` is a translation field name for Gadget binary datasets defined [here](https://github.com/yt-project/yt/blob/83c96215858a4899e0c0546438c0afca6da12fae/yt/frontends/gadget/definitions.py#L68C3-L85). `MetalMasses` is a convention I've seen used in Gadget HDF5, and (even worse) `Mass of Metals` is one I saw used in another similar dataset.

neutrinoceros:
I meant to fill in as a reviewer but I'm happy to get behind @munkm's approval.

yt/frontends/gadget/fields.py
jzuhone: I agree with you--I can add some comments to explain what's going on.

yt/frontends/gadget/tests/test_outputs.py
jzuhone: So `14` has to be treated differently. It's not a real element, it's basically just the mass of the rest of the elements not specified in the list, and in the codes that use it it's called `Ej` for "ejecta" (because it comes from supernovae explosions). We define an `Ej_fraction` field, but the reason I can't stick it in the `elem` list is that list the `Ej_mass` field doesn't exist, as it's not a real element and it doesn't get passed through yt's species machinery (though maybe it should be defined on its own).

jzuhone:
@munkm @neutrinoceros see https://github.com/yt-project/yt/pull/4584/commits/c8f460407a86161e61b6a411a5f8c82d64acb3f5 and https://github.com/yt-project/yt/pull/4584/commits/5156e2906906334accc395af9e22b322a656eb88, thanks for the helpful suggestions that influenced these.

jzuhone:
This is ready.

yt/frontends/gadget/tests/test_outputs.py
munkm: Ok so your new commits do define it on its own now, is my understanding of that commit correct?

yt/frontends/gadget/fields.py
munkm: Ok, cool! Thanks for the clarification.

yt/frontends/gadget/fields.py
munkm: The new comments are helpful! Thank you for adding them.

munkm (approved):
I like the improvements you made. Thank you for the changes. 

One thing I thought about overnight -- is there any possibility a file would have more than one grouping of `n_elem`? (that is, that more than one option could be triggered by your logic) I can't think of a situation where that would happen, but if it does it it something we should add handling for?

PR #4586, opened 2023-07-20
BUG: avoid divide-by-zero runtime warning from np.log10 in AMRKDTree.get_brick_data

neutrinoceros:
reprod: `python -Werror doc/source/cookbook/render_two_fields.py`

note that with numpy>=1.25.0, #4585 also gets in the way


matthewturk (approved):
I thought we had a separate NaN filter, but this looks fine.  MAybe that was yt_idv.

PR #4597, opened 2023-07-24
reduce code duplication in IOHandler _read_particle_coords and _read_particle_fields 

chrishavlin:
This PR reduces duplicated code related to iterating over `data_file` chunks in `_read_particle_coords` and `_read_particle_fields` across IOHandlers. Shouldn't affect any behavior...

### DIFF
--- a/yt/frontends/adaptahop/io.py
+++ b/yt/frontends/adaptahop/io.py
@@ -36,16 +36,11 @@ class IOHandlerAdaptaHOPBinary(BaseParticleIOHandler):
 
     def _read_particle_coords(self, chunks, ptf):
         # This will read chunks and yield the results.
-        chunks = list(chunks)
-        data_files = set()
         # Only support halo reading for now.
         assert len(ptf) == 1
         assert list(ptf.keys())[0] == "halos"
         ptype = "halos"
-        for chunk in chunks:
-            for obj in chunk.objs:
-                data_files.update(obj.data_files)
-        for data_file in sorted(data_files, key=attrgetter("filename")):
+        for data_file in self._sorted_chunk_iterator(chunks):
             pcount = (
                 data_file.ds.parameters["nhalos"] + data_file.ds.parameters["nsubs"]
             )
@@ -56,14 +51,10 @@ class IOHandlerAdaptaHOPBinary(BaseParticleIOHandler):
 
     def _read_particle_fields(self, chunks, ptf, selector):
         # Now we have all the sizes, and we can allocate
-        chunks = list(chunks)
-        data_files = set()
+
         # Only support halo reading for now.
         assert len(ptf) == 1
         assert list(ptf.keys())[0] == "halos"
-        for chunk in chunks:
-            for obj in chunk.objs:
-                data_files.update(obj.data_files)
 
         def iterate_over_attributes(attr_list):
             for attr, *_ in attr_list:
@@ -76,7 +67,7 @@ class IOHandlerAdaptaHOPBinary(BaseParticleIOHandler):
 
         attr_pos = partial(_find_attr_position, halo_attributes=halo_attributes)
 
-        for data_file in sorted(data_files, key=attrgetter("filename")):
+        for data_file in self._sorted_chunk_iterator(chunks):
             pcount = (
                 data_file.ds.parameters["nhalos"] + data_file.ds.parameters["nsubs"]
             )
@@ -193,6 +184,10 @@ class IOHandlerAdaptaHOPBinary(BaseParticleIOHandler):
                 fpu.skip(todo.pop(0))
             members = fpu.read_attrs(todo.pop(0))["particle_identities"]
         return members
+
+    def _sorted_chunk_iterator(self, chunks):
+        data_files = self._get_data_files(chunks)
+        yield from sorted(data_files, key=attrgetter("filename"))
 
 
 def _todo_from_attributes(attributes: ATTR_T, halo_attributes: ATTR_T):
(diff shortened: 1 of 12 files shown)


yt/utilities/io_handler.py
```diff
@@ -224,31 +224,34 @@ def _read_particle_selection(
...
+    def _get_data_files(chunks):
         chunks = list(chunks)
         data_files = set()
         for chunk in chunks:
             for obj in chunk.objs:
                 data_files.update(obj.data_files)
+        return data_files
+
+    def _sorted_chunk_iterator(self, chunks):
+        data_files = self._get_data_files(chunks)
```
matthewturk, line 244: I think we should explicitly turn chunks into a list here, unless it has been elsewhere higher up the stack; recently we exhausted an iterator and ran into issues.  (Maybe with `nc_cm1`?)  Also, we sort with `x.start` and `x.filename` but not all frontends did that -- does `x.start` have a default value of 0 or something constant that we can rely on?

matthewturk (dismissed):

yt/frontends/ahf/io.py
```diff
@@ -67,15 +69,13 @@ def _identify_fields(self, data_file):
 
     # Helper methods
 
-    def _get_data_files(self, chunks, ptf):
+    @staticmethod
+    def _validate_particle_ptf(ptf):
```
neutrinoceros, line 73: this function is now 100% `assert` statements, which means it can be treated as noop by linters/type-checkers (because `assert`s can be ignored at runtime). I suggest rewriting it with actual an `if/raise` clause. It's also possible to *leverage* the type checking system so that this function actually helps narrowing down the type of `ptf` where it's used, using `TypeGuard`. IMO this would also help reading the code (I had no idea what `ptf` stood like or what type of data was assumed until I researched it for the sake of this suggestion)

```suggestion
    def _validate_particle_ptf(ptf: Dict[str, Any]) -> TypeGuard[HalosPTF]:
        if list(ptf) != ["halos"]:
            raise ValueError(...)
        return True
```

with
```python
from typing import TypedDict, Any
class HalosPTF(TypedDict):
    halos: Any
```

(I'm using `Any` mostly because I'm not sure what type these values are)

yt/utilities/io_handler.py
chrishavlin: Oh and I left the front ends that do not sort by both filename and start intact. They override this method so that they are unchanged

yt/frontends/ahf/io.py
chrishavlin: a bit of extra context -- I only added this function because this particular frontend had lumped in the validation of this dictionary to a `_get_data_files` function and I wanted to separate out the validation so that the call signature of `_get_data_files` matches the new base class function. So it seemed convenient to add this new function, but there are actually at least 3-4 other frontends that do this **exact** assert validation within the body of `_read_particle_fields` and `_read_particle_coords`, so I could simply do that here rather than adding the new function. I do think raising errors would indeed be better, but it may be simpler to refactor all the halo particle readers at once to raise errors instead of assert (in a separate PR). 

also happy to just use your suggestion :)

yt/frontends/adaptahop/io.py
```diff
@@ -194,6 +185,10 @@ def members(self, ihalo):
             members = fpu.read_attrs(todo.pop(0))["particle_identities"]
         return members
 
+    def _sorted_chunk_iterator(self, chunks):
+        data_files = self._get_data_files(chunks)
+        yield from sorted(data_files, key=attrgetter("filename"))
```
chrishavlin, line 190: @matthewturk just pointing out an example where I added an override of `_sorted_chunk_iterator` so that the frontend continues only sorting by filename. The change in `ahf` is similar.

yt/frontends/ahf/io.py
neutrinoceros: It would probably be easier to refactor them all at once, and I don't want my suggestion to leak way out of scope here so it's probably best to preserve the similarity for now

yt/frontends/sdf/io.py
```diff
@@ -31,13 +27,9 @@ def _read_particle_coords(self, chunks, ptf):
...
-        chunks = list(chunks)
-        data_files = set()
         assert len(ptf) == 1
         assert ptf.keys()[0] == "dark_matter"
-        for chunk in chunks:
-            for obj in chunk.objs:
-                data_files.update(obj.data_files)
+        data_files = self._get_data_files(chunks)
         assert len(data_files) == 1
         for _data_file in sorted(data_files, key=lambda x: (x.filename, x.start)):
```
chrishavlin, line 34: just a note: I did not use the `_sorted_chunk_iterator` here because I wanted to keep the `assert len(data_files) == 1` line above.

yt/frontends/ahf/io.py
chrishavlin: OK, that sounds good to me. I'll go ahead and move these asserts to the body of the functions then.

chrishavlin: done

yt/utilities/io_handler.py
chrishavlin: OK, so `ParticleFile.start` does initialize to 0 if it's not provided (within `ParticleFile.__init__` ([link to relevant code](https://github.com/yt-project/yt/blob/8a113e9337ab526e3c7691ee6b2abd3d543696fc/yt/data_objects/static_output.py#L2074C8-L2087)). 

 But in any case, the changes in this PR do not change any behavior -- only the frontends that were sorting by `filename` and `start` are using this function. Those that were not already doing that either override this method or don't use this method at all.

neutrinoceros (approved):
Glad to see this merged if @matthewturk is still happy with it.

matthewturk (approved):

PR #4600, opened 2023-07-25
TST: cleanup unused test code

neutrinoceros:
## PR Summary

This utility function doesn't seem to be used anywhere (I think `check_recipe` is doing the same job, but is actually used)


matthewturk (approved):
Once upon a time I know this was necessary because errors were being swallowed, and I think it might have been a function that got swapped out for a different one when we needed to.  I'm not sure if that's still the case or not.

neutrinoceros:
@yt-fido test this please

PR #4601, opened 2023-07-25
BUG: fix string interpolation (units double-printing) in GridIndex

neutrinoceros:
follow up #4443 and #4582


matthewturk (approved):
good catch!

PR #4605, opened 2023-07-27
Updates to the _parse_parameter_file function following PR 4470

mtryan83:
## PR Summary

This change updates the `_parse_parameter_file` function in the Gadget frontend to match the changes developed in PR #4470. In summary:
 1. Change try/except to if/else
 2. Remove only_on_root wrapper to mylog.info
 3. Change fall-back omega_matter from 1 to 0
 4. Condense prefix logic

I've left the for/else structure in, since in #4470 I just removed it (didn't need to support Gadget-2 format for Gizmo files) and so haven't fully thought of what the replacement should look like. 

## PR Checklist

I believe this neither adds new features, nor fixes any bugs, and is more of a code cleanup than anything else.

### DIFF
--- a/yt/frontends/gadget/data_structures.py
+++ b/yt/frontends/gadget/data_structures.py
@@ -376,6 +376,8 @@ class GadgetDataset(SPHDataset):
             # If these are not set it is definitely not a cosmological dataset.
             self.omega_lambda = 0.0
             self.omega_matter = 0.0  # Just in case somebody asks for it.
+            # omega_matter has been changed to 0.0 for consistency with
+            # non-cosmological frontends
             self.cosmological_simulation = 0
             # Hubble is set below for Omega Lambda = 0.
 
--- a/yt/frontends/gadget/data_structures.py
+++ b/yt/frontends/gadget/data_structures.py
@@ -363,24 +363,20 @@ class GadgetDataset(SPHDataset):
         self.domain_dimensions = np.ones(3, "int32")
         self._periodicity = (True, True, True)
 
-        self.cosmological_simulation = 1
+        self.current_redshift = hvals.get("Redshift", 0.0)
+        if "Redshift" not in hvals:
+            mylog.info("Redshift is not set in Header. Assuming z=0.")
 
-        try:
-            self.current_redshift = hvals["Redshift"]
-        except KeyError:
-            # Probably not a cosmological dataset, we should just set
-            # z = 0 and let the user know
-            self.current_redshift = 0.0
-            only_on_root(mylog.info, "Redshift is not set in Header. Assuming z=0.")
-
-        try:
+        if "OmegaLambda" in hvals:
             self.omega_lambda = hvals["OmegaLambda"]
             self.omega_matter = hvals["Omega0"]
             self.hubble_constant = hvals["HubbleParam"]
-        except KeyError:
+            self.cosmological_simulation = self.omega_lambda != 0.0
+        else:
             # If these are not set it is definitely not a cosmological dataset.
             self.omega_lambda = 0.0
-            self.omega_matter = 1.0  # Just in case somebody asks for it.
+            self.omega_matter = 0.0  # Just in case somebody asks for it.
+            self.cosmological_simulation = 0
             # Hubble is set below for Omega Lambda = 0.
 
         # According to the Gadget manual, OmegaLambda will be zero for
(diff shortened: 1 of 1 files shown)


yt/frontends/gadget/data_structures.py
```diff
@@ -409,20 +402,14 @@ def _parse_parameter_file(self):
...
             )
         self.parameters = hvals
 
-        prefix = os.path.abspath(
-            os.path.join(
-                os.path.dirname(self.parameter_filename),
-                os.path.basename(self.parameter_filename).split(".", 1)[0],
-            )
-        )
+        prefix = os.path.join(self.directory, self.basename.split(".", 1)[0])
```
neutrinoceros, line 412: A pet-peeve of mine. Not needed if it degrades readability to others.
```suggestion
        basename, _, _ = self.basename.partition(".")
        prefix = os.path.join(self.directory, basename)
```

neutrinoceros (commented):
just a small comment. I'll review properly tomorrow,

neutrinoceros (approved):
actually I don't have other comments, LGTM

brittonsmith (dismissed):

neutrinoceros (dismissed):

jzuhone (dismissed):

jzuhone:
I guess my question is if this is now in line with PR #4470, then it seems like we should have done this first and then tried to find a way for the GIZMO frontend's method to call this one as `super` and then do the things it needed to do. But maybe that's not possible.

chummels:
So it seems like PR #4470 went through because of changes in the Gizmo frontend, but do we know that equivalent changes occurred in the underlying Gadget code as well?  My concern is that this might break how we're loading in Gadget cosmological simulations, since we're no longer assuming `Redshift` to be set in the underlying dataset, and now basing our assessment of cosmology on the presence of the `OmegaLambda` parameter.

mtryan83:
I don't know of any changes to the Gadget frontend, but my understanding is that nothing here should change the logic of how the parameter file is parsed. For example, the `Redshift`  parameter was assumed but not guaranteed and either way did not effect whether yt treated it as a cosmological dataset:
```python
# previous version of the code
try:
    self.current_redshift = hvals["Redshift"]
except KeyError:
    # Probably not a cosmological dataset, we should just set
    # z = 0 and let the user know
    self.current_redshift = 0.0
    only_on_root(mylog.info, "Redshift is not set in Header. Assuming z=0.")
```
and then
```python
# Previous version
if self.omega_lambda == 0.0:
    ### ... Not cosmological ...
    self.current_redshift = 0.0 ### This is set to 0, even if it wasn't before
```

Please feel free to correct me though and, of course, I have no idea if the previous logic is correct.

To @jzuhone 's point, I believe you are correct. In hindsight, I think the Gizmo frontend could call `super` and then just reevaluate only the omega/cosmology section (i.e. lines 370-409). I don't have time to work on this before mid-September though.

yt/frontends/gadget/data_structures.py
```diff
@@ -363,24 +363,20 @@ def _parse_parameter_file(self):
...
             self.omega_lambda = hvals["OmegaLambda"]
             self.omega_matter = hvals["Omega0"]
             self.hubble_constant = hvals["HubbleParam"]
-        except KeyError:
+            self.cosmological_simulation = self.omega_lambda != 0.0
+        else:
             # If these are not set it is definitely not a cosmological dataset.
             self.omega_lambda = 0.0
-            self.omega_matter = 1.0  # Just in case somebody asks for it.
+            self.omega_matter = 0.0  # Just in case somebody asks for it.
```
JBorrow, line 378: Why change this to 0.0, instead if 1.0? This would imply there's nothing in the volume.

mtryan83: From @neutrinoceros 
> non-cosmological datasets normally have omega_matter set to 0.0. Is there a reason not to follow the convention here ?

in PR #4470

JBorrow: Ok, so it was changed for consistency with other frontends? That's ok, but maybe warrants changing the comment from 'Just incase somebody asks for it' to 'Changed to 0.0 for consistency with non-cosmological frontends' in case somebody comes looking?

mtryan83: Fair point. The comment was actually left unchanged from the previous code (and I believe is just referencing that omega_matter is being set, not necessarily it's value), I had only changed the numerical value here. 

I've added your comment, though I left the original as well, since it does describe why omega_matter is being set at all.

jzuhone (approved):

PR #4607, opened 2023-07-28
BLD: forbid deprecated numpy C API

### LINKED ISSUE #4606: BLD: future-compat issue with numpy (numpy 1.7 deprecated C API)
### Bug report

**Bug summary**

We've had build-time warnings `"Using deprecated NumPy API..."` for ever, and until recently there was nothing to be done about it, because the C code using deprecating numpy API was generated by Cython and there was no way to prevent that. However [it's now possible with Cython 3.0](http://docs.cython.org/en/latest/src/userguide/migrating_to_cy30.html?highlight=numpy%20deprecated%20api#numpy-c-api).

It's not clear exactly *when* Numpy 1.7 deprecated API will be dropped but with numpy 2.0 around the corner, I'd expect it sooner rather than later.

neutrinoceros:
## PR Summary

Close #4606

### DIFF
--- a/setupext.py
+++ b/setupext.py
@@ -409,6 +409,12 @@ def create_build_ext(lib_exts, cythonize_aliases):
             self.include_dirs.append(numpy.get_include())
             self.include_dirs.append(ewah_bool_utils.get_include())
 
+            define_macros = [("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")]
+            if self.define is None:
+                self.define = define_macros
+            else:
+                self.define.extend(define_macros)
+
         def build_extensions(self):
             self.check_extensions_list(self.extensions)
 


Xarthisius (approved):

chrishavlin:
read through the cython 3 link in the linked issue and it says:

> One side-effect of the different C-API usage is that your code may now require a call to the [NumPy C-API initialisation function](https://docs.scipy.org/doc/numpy-1.17.0/reference/c-api.array.html#importing-the-api) where it previously got away without doing so.

Would we get test failures with this PR if that were the case for us?

neutrinoceros:
Yes we are ! We switched the requirements to Cython>3.0 in a previous PR, so I expect we'd see errors in CI if this was impacting us.

chrishavlin (approved):

PR #4621, opened 2023-08-05
TST: simplify cartopy integration testing (using cartopy 0.22)

neutrinoceros:
Cartopy 0.22 was just released and is supposed to resolve a couple workarounds we've been using to test it.
Most notably, this version doesn't depend on GEOS, and as such, it can afford to have *wheels*, which should allow us to properly test it on all platforms for cheap !

### DIFF
--- a/tests/ci_install.sh
+++ b/tests/ci_install.sh
@@ -1,37 +1,30 @@
 set -x   # Show which command is being run
 
-# geos is needed for cartopy, see
-# https://scitools.org.uk/cartopy/docs/latest/installing.html?highlight=install#building-from-source
-case ${RUNNER_OS} in
-linux|Linux)
-    sudo apt-get -qqy update
-    sudo apt-get -qqy install \
-      libhdf5-serial-dev \
-      libnetcdf-dev \
-      libgeos-dev \
-      libopenmpi-dev \
-      libfuse2
-    ;;
-osx|macOS)
-    sudo mkdir -p /usr/local/man
-    sudo chown -R "${USER}:admin" /usr/local/man
-    HOMEBREW_NO_AUTO_UPDATE=1 brew install hdf5 geos open-mpi netcdf ccache macfuse
-    ;;
-esac
-
 # Sets default backend to Agg
 cp tests/matplotlibrc .
 
 # Step 1: pre-install required packages
-if [[ "${RUNNER_OS}" == "Windows" ]] && [[ ${dependencies} != "minimal" ]]; then
-    # windows_conda_requirements.txt is a survivance of test_requirements.txt
-    # keep in sync: pyproject.toml
-    conda install --yes --file=tests/windows_conda_requirements.txt
-else
+if [[ ${dependencies} == "full" || ${dependencies} == "cartopy" ]]; then
     # upgrading pip to guarantee installing extra dependencies with [full] is supported
     # this is only necessary for some versions of Python 3.8 and 3.9
     # see https://github.com/yt-project/yt/issues/4270
-    python -m pip install --upgrade pip
+    python -m pip install 'pip>=21.2'
+
+    case ${RUNNER_OS} in
+    linux|Linux)
+        sudo apt-get -qqy update
+        sudo apt-get -qqy install \
+        libhdf5-serial-dev \
+        libnetcdf-dev \
+        libopenmpi-dev \
+        libfuse2
+        ;;
+    osx|macOS)
+        sudo mkdir -p /usr/local/man
+        sudo chown -R "${USER}:admin" /usr/local/man
+        HOMEBREW_NO_AUTO_UPDATE=1 brew install hdf5 open-mpi netcdf ccache macfuse
+        ;;
+    esac
 fi
 
 # Step 2: install deps and yt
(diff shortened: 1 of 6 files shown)


tests/ci_install.sh
```diff
@@ -1,37 +1,39 @@
...
+        conda install shapely scipy
+    else
+        # geos is needed for shapely, which is needed for cartopy, see
+        case ${RUNNER_OS} in
+        linux|Linux)
+            sudo apt-get -qqy update
+            sudo apt-get -qqy install \
+            libhdf5-serial-dev \
+            libnetcdf-dev \
+            libgeos-dev \
```
chrishavlin, line 24:
```suggestion
```

tests/ci_install.sh
```diff
@@ -1,37 +1,39 @@
...
+            libopenmpi-dev \
+            libfuse2
+            ;;
+        osx|macOS)
+            sudo mkdir -p /usr/local/man
+            sudo chown -R "${USER}:admin" /usr/local/man
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install hdf5 geos open-mpi netcdf ccache macfuse
```
chrishavlin, line 31:
```suggestion
            HOMEBREW_NO_AUTO_UPDATE=1 brew install hdf5 open-mpi netcdf ccache macfuse
```

tests/ci_install.sh
```diff
@@ -1,37 +1,39 @@
...
-    conda install --yes --file=tests/windows_conda_requirements.txt
-else
+if [[ ${dependencies} == "full" || ${dependencies} == "cartopy" ]]; then
     # upgrading pip to guarantee installing extra dependencies with [full] is supported
     # this is only necessary for some versions of Python 3.8 and 3.9
     # see https://github.com/yt-project/yt/issues/4270
-    python -m pip install --upgrade pip
+    python -m pip install 'pip>=21.2'
+
+    # see https://scitools.org.uk/cartopy/docs/latest/installing.html?highlight=install#building-from-source
```
chrishavlin, line 13:
```suggestion
```

tests/ci_install.sh
```diff
@@ -1,37 +1,39 @@
...
+            libopenmpi-dev \
+            libfuse2
+            ;;
+        osx|macOS)
+            sudo mkdir -p /usr/local/man
+            sudo chown -R "${USER}:admin" /usr/local/man
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install hdf5 geos open-mpi netcdf ccache macfuse
+            ;;
+        esac
+        python -m pip install shapely --no-binary=shapely
```
chrishavlin, line 34:
```suggestion
```

tests/ci_install.sh
```diff
@@ -1,37 +1,39 @@
...
     # upgrading pip to guarantee installing extra dependencies with [full] is supported
     # this is only necessary for some versions of Python 3.8 and 3.9
     # see https://github.com/yt-project/yt/issues/4270
-    python -m pip install --upgrade pip
+    python -m pip install 'pip>=21.2'
+
+    # see https://scitools.org.uk/cartopy/docs/latest/installing.html?highlight=install#building-from-source
+    if [[ "${RUNNER_OS}" == "Windows" ]]; then
+        conda install shapely scipy
+    else
```
chrishavlin, line 16: I'm not sure this windows case is needed at all anymore? we certainly shouldn't need to install shapely anymore since that installation step was needed to link to geos when building from source.

tests/ci_install.sh
```diff
@@ -40,12 +42,12 @@ fi
 if [[ ${dependencies} == "minimal" ]]; then
     # test with minimal versions of runtime dependencies
     python -m pip install -e ".[test,minimal]"
+elif [[ ${dependencies} == "cartopy" ]]; then
+    python -m pip install 'cartopy>=0.22'
```
chrishavlin, line 46: So I think if we add in a pip install scipy here as well then we can remove the two scipy install lines above (on lines 35 and 15).

chrishavlin: I **do** think we still need an explicit scipy install since it's an optional cartopy dep and an optional yt dep (but we're using the cartopy functionality for which it is required). It is in our full deps though, so we don't need it down below.

chrishavlin (commented):
The `ci_install.sh` script was still installing geos -- which we shouldn't have to do anymore. I left comments pointing out those spots as well as a couple other further simplifications that I think are possible.

Also, adding cartopy>=0.22 as a proper dependency in `yt[full]` would further simplify the setup here. Thoughts on that?

chrishavlin:
Oh huh, that's a shame... I think all my comments can be disregarded in that case.

neutrinoceros:
I checked, and I think the title of [this PR](https://github.com/SciTools/cartopy/pull/2083) says it all

chrishavlin:
Looking more carefully at shapely -- they do package geos. Did you try removing the geos install from the Ubuntu and Mac branch of the install script?

chrishavlin:
> I checked, and I think the title of [this PR](https://github.com/SciTools/cartopy/pull/2083) says it all

But the shapely docs make it pretty clear that you should only need to install geos explicitly if you want a specific version https://shapely.readthedocs.io/en/latest/installation.html maybe that's not the case in practice

neutrinoceros:
ah ! Let's try not to build shapely from source then: it doesn't seem to be recommended any more by cartopy, and I assume it would save us installing geos too

neutrinoceros:
Looks like it worked ! thank you for spotting this Chris, I will iterate a little more and then reopen for review

neutrinoceros:
Note that wheels are only available for Python 3.9 and above. I don't think this is blocking here, but that's a sign that supporting Python 3.8 is becoming harder for us.

neutrinoceros:
Another note: I think this works makes `conda` useless in CI (only used on Windows), but I'd prefer to defer dropping it to a follow-up PR.

chrishavlin (approved):
Nice! And removing conda if possible in a later pr makes sense to me.

PR #4625, opened 2023-08-08
BLD: add wheels for CPython 3.12

neutrinoceros:
celebrating Python 3.12.0rc1 + cibuildwheel 2.15.0

requires numpy wheels (probably won't happen before numpy 1.26)

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -31,11 +31,11 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Build wheels for CPython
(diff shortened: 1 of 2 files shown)


neutrinoceros:
numpy 1.26.0b1 is already ABI stable https://github.com/numpy/numpy/issues/23808#issuecomment-1676838801
so we can use that version to get wheels going and simplify requirements later

neutrinoceros:
turns out we need to get wheels going for ewah-bool-utils first. This is happening at https://github.com/yt-project/ewah_bool_utils/pull/55

Xarthisius (approved):

PR #4634, opened 2023-08-15
MNT: drop unmaintained install script

### LINKED ISSUE #4633: DOC: let's drop install_script.sh ?
### Bug report

**Bug summary**

As [recently discussed on the user mailing list](https://mail.python.org/archives/list/yt-users@python.org/thread/PGFASTGYO7IXTJYKZHPDZNSAZTS6PPUP/), `doc/install_script.sh` hasn't been updated in 2 years, and is currently in a non-functional state.

My personal opinion is that this script isn't bringing much value within the modern ecosystem, where build systems and compilation are completely encapsulated with `pyproject.toml` + `pip`, so I would prefer just dropping the script completely.

neutrinoceros:
Since it doesn't seem controversial, close #4633 

I note that there are still a couple files referencing this script (in `docker/**/Dockerfile`). I assume these should be dropped too, but I'd like @Xarthisius' first hand opinion on it first.


Xarthisius:
Yeah, those Dockerfile reference install_script from bitbucket... Feel free to drop entire `docker/`

Xarthisius (approved):

jzuhone (approved):

Xarthisius (approved):

PR #4640, opened 2023-08-21
ENH: Only call find_spec sometimes

matthewturk:
In some cases, particularly when instantiating lots of child masks, the way the `MutableAttribute` object is set up could lead to `find_spec` being called inside some tight loops.

But, the results of `find_spec` shouldn't change during the course of one interpreter's existence.  So we can move the checks to the outside of the `__get__` call, which reduces costs of grid indexing considerable in some cases.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -131,7 +131,13 @@ class MutableAttribute:
 
     def __init__(self, display_array=False):
         self.data = weakref.WeakKeyDictionary()
-        self.display_array = display_array
+        # We can assume that ipywidgets will not be *added* to the system
+        # during the course of execution, and if it is, we will not wrap the
+        # array.
+        if display_array and find_spec("ipywidgets") is not None:
+            self.display_array = True
+        else:
+            self.display_array = False
 
     def __get__(self, instance, owner):
         if not instance:
@@ -141,7 +147,7 @@ class MutableAttribute:
             ret = ret.copy()
         except AttributeError:
             pass
-        if self.display_array and find_spec("ipywidgets") is not None:
+        if self.display_array:
             try:
                 ret._ipython_display_ = functools.partial(_wrap_display_ytarray, ret)
             # This will error out if the items have yet to be turned into


yt/data_objects/static_output.py
```diff
@@ -131,7 +131,13 @@ class MutableAttribute:
...
     def __init__(self, display_array=False):
         self.data = weakref.WeakKeyDictionary()
-        self.display_array = display_array
+        # We can assume that ipywidgets will not be *added* to the system
+        # during the course of execution, and if it is, we will not wrap the
+        # array.
+        if display_array and find_spec("ipywidgets") is not None:
+            self.display_array = True
+        else:
+            self.display_array = False
```
neutrinoceros, line 140: I have a small preference for the one-liner form, but it's not crucial (I've checked that it is black-compliant)
```suggestion
        self.display_array = display_array and find_spec("ipywidgets") is not None
```

neutrinoceros (approved):
good catch !

neutrinoceros:
> the results of find_spec shouldn't change during the course of one interpreter's existence.

Note that this isn't 100% true: it is possible to install a package at runtime using `import pip; pip.main(["install", "pkg"])` or equivalently (and perhaps more commonly), `IPython`'s magic command `%pip`, but it seems reasonable to me to sacrifice a little flexibility in that exceptional case to save performance in the general case.

matthewturk:
Yeah, I know it's not strictly true, but it's true enough.  We shouldn't bother catching it.

neutrinoceros:
agreed

chrishavlin:
for a bit of extra context: we found that this `find_spec` call was sucking up ~20s when calling `ds.find_max('density')` with some [santa fe light cone output](https://library.ucsd.edu/dc/collection/bb1673671n) -- this PR takes that calculation from 90s down to 70s.

chrishavlin (approved):
Oops, didn't actually hit approve with my last comment :)

neutrinoceros:
Moving this to 4.2.2 upon @matthewturk's request

PR #4641, opened 2023-08-21
ENH: Utilize readonly attributes

matthewturk:
My initial testing suggests this could work.  If we mandate all incoming arrays are in code length and un-writeable, we should be able to avoid using `ensure_code` in some particularly tight loops.

I'll fiddle with the tests if/when they fail.


matthewturk:
So my read of the answer tests is that nearly all of them changed results.  This worries me a bit, as it suggests that there was some reliance on modifications to these arrays.

matthewturk:
Actually, I take that back, and I am going to push a change that undoes something.

yt/data_objects/static_output.py
```diff
@@ -125,23 +125,21 @@ def __init__(self, display_array=False):
...
-                pass
         return ret
 
     def __set__(self, instance, value):
+        if self.display_array and find_spec("ipywidgets") is not None:
+            try:
+                value._ipython_display_ = functools.partial(
+                    _wrap_display_ytarray, value
+                )
+            except AttributeError:
```
neutrinoceros, line 137: I don't think the try/except serves any purpose after your refactor. Am I missing anything ?

matthewturk: Yeah, I think I got confused doing the merge

neutrinoceros: Seeing the failures, I understand how wrong I was to assume that setting a new attribute on any object is always valid; ndarrays and other objects might in fact use protective mechanisms like `__slots__` to prevent it. Anyway, sorry about the confusion!

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros:
Remaining failures are real

matthewturk:
So this is a way I think we can move toward speedups for some particularly difficult and tight loops.

neutrinoceros:
@matthewturk but it's not supposed to provide gains in itself, right ?

yt/data_objects/static_output.py
neutrinoceros: I'd suggest adding a comment to explain this, because my memory apparently cannot retain this reasoning for a whole week 😅

yt/data_objects/static_output.py
```diff
@@ -125,23 +125,21 @@ def __init__(self, display_array=False):
...
+                )
             except AttributeError:
                 pass
-        return ret
-
-    def __set__(self, instance, value):
+        try:
+            value.flags.writeable = False
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
neutrinoceros: I'm also thinking, isn't there a contradiction in making `MutableAttribute`s read-only ?
I've never found this class' name useful or descriptive, so if it needs to go I won't mind.

matthewturk:
With respect to the name, I think it's safe to say that what we're doing is making them immutable, because they *are* mutable.  So I think it's OK?  I dunno, I didn't really do the name initially.

matthewturk:
@neutrinoceros I find it really kind of funny that I committed your suggestion and that somehow that included tabs -- which is throwing an error on ruff!  :)

doc/source/cookbook/thin_slice_projection.py
```diff
@@ -9,12 +9,12 @@
 # as a data_source for the projection.
 
 # Center on the domain center
-center = ds.domain_center
```
neutrinoceros, line 12: it occurs to me that these arrays seem pretty cheap to copy. Would there be a clear downside to make the `MutableAttribute` descriptor always return copies ?

matthewturk: Yeah, that's part of what we were trying to avoid with this.  They might be cheap, but if they get called an enormous number of times, they get expensive -- since we then also utilize a fair bit of unyt code.  That's what the original mutable attribute code did, to ensure that the underlying values were never changed.

neutrinoceros: fair enough

neutrinoceros (approved):

PR #4646, opened 2023-08-23
make load_sample MPI-safe

### LINKED ISSUE #4635: BUG: `load_sample` is not MPI-safe
We should decorate the appropriate function in `load_sample` so that it doesn't get run on all processors after `enable_parallelism` has been called.

chrishavlin:
One option to Close #4635 

This separates the data fetching and path construction from the `yt.load` call that `load_sample` does so that the first part can be easily moved to a blocking root-only call and then the valid path can be broadcast to the mpi processes.

### DIFF
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1535,7 +1535,7 @@ def _get_sample_data(
             "Please report this to https://github.com/yt-project/yt/issues/new"
         )
 
-    kwargs = {**specs["load_kwargs"], **kwargs}
+    load_kwargs = {**specs["load_kwargs"], **kwargs}
     save_dir = _get_test_data_dir_path()
 
     data_path = save_dir.joinpath(fn)
@@ -1547,7 +1547,7 @@ def _get_sample_data(
         mylog.info("Sample dataset found in '%s'", data_path)
         if timeout is not None:
             mylog.info("Ignoring the `timeout` keyword argument received.")
-        return data_path, kwargs
+        return data_path, load_kwargs
 
     mylog.info("'%s' is not available locally. Looking up online.", fn)
 
@@ -1602,7 +1602,7 @@ def _get_sample_data(
         # cache dir isn't empty
         pass
 
-    return loadable_path, kwargs
+    return loadable_path, load_kwargs
 
 
 def load_sample(
@@ -1651,10 +1651,10 @@ def load_sample(
     - Corresponding sample data live at https://yt-project.org/data
 
     """
-    loadable_path, kwargs = _get_sample_data(
+    loadable_path, load_kwargs = _get_sample_data(
         fn, progressbar=progressbar, timeout=timeout, **kwargs
     )
-    return load(loadable_path, **kwargs)
+    return load(loadable_path, **load_kwargs)
 
 
 def _mount_helper(
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -1650,7 +1650,7 @@ def load_sample(
     - Corresponding sample data live at https://yt-project.org/data
 
     """
-    loadable_path = _get_sample_data(fn=fn, progressbar=progressbar, timeout=timeout)
+    loadable_path = _get_sample_data(fn, progressbar=progressbar, timeout=timeout)
     return load(loadable_path, **kwargs)
 
 
--- a/yt/loaders.py
+++ b/yt/loaders.py
@@ -39,6 +39,9 @@ from yt.utilities.object_registries import (
     simulation_time_series_registry,
 )
 from yt.utilities.on_demand_imports import _pooch as pooch, _ratarmount as ratarmount
+from yt.utilities.parallel_tools.parallel_analysis_interface import (
+    parallel_root_only_then_broadcast,
+)
 
 if TYPE_CHECKING:
     from multiprocessing.connection import Connection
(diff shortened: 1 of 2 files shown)


yt/loaders.py
```diff
@@ -1460,52 +1463,14 @@ def flatten(l):
 
 
 # --- Loader for yt sample datasets ---
-def load_sample(
+@parallel_root_only_then_broadcast
+def _get_sample_data(
     fn: Optional[str] = None, *, progressbar: bool = True, timeout=None, **kwargs
```
chrishavlin, line 1468: just realized that `kwargs` here probably needs special treatment... if this functions modifies it (which it **can** do), then that would only happen on the root process.

yt/loaders.py
```diff
@@ -1636,6 +1601,56 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
...
+      the data registry, the ones passed to this function take priority.
+    - Datasets with slashes '/' in their names can safely be used even on Windows.
+      On the contrary, paths using backslashes '\' won't work outside of Windows, so
+      it is recommended to favour the UNIX convention ('/') in scripts that are meant
+      to be cross-platform.
+    - This function requires pandas and pooch.
+    - Corresponding sample data live at https://yt-project.org/data
+
+    """
+    loadable_path = _get_sample_data(fn=fn, progressbar=progressbar, timeout=timeout)
```
chrishavlin, line 1653: ALSO just realied that kwargs isnt' getting passed down, it needs ot

chrishavlin:
@neutrinoceros I switched to draft cause of the kwargs issue I pointed out just now in a few comments, but feel free to review at any point if you want -- totally open to different approaches than what I did here, it just seemed the simplest.

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -326,6 +326,32 @@ def barrierize(*args, **kwargs):
     return barrierize
 
 
+def parallel_root_only_then_broadcast(func):
```
chrishavlin, line 329: one note: I think we could instead make `parallel_root_only` accept a `broadcast_result` boolean so you could use it like:

```python
@parallel_root_only(broadcast_result=True)
def my_root_only_func():
     .... 
```
and have the default behavior continue to broadcast the `all_clear` integer that it does. Might be better than adding this new function?

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -326,6 +326,32 @@ def barrierize(*args, **kwargs):
...
+        rv0 = None
+        if comm.rank == 0:
+            try:
+                rv0 = func(*args, **kwargs)
+            except Exception:
+                traceback.print_last()
+        rv = comm.mpi_bcast(rv0)
+        if not rv:
+            raise RuntimeError
+        rv = comm.mpi_bcast(rv0)
```
neutrinoceros, line 349: this line looks dupplicated ?

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -326,6 +326,32 @@ def barrierize(*args, **kwargs):
...
+    def root_only(*args, **kwargs):
+        if not parallel_capable:
+            return func(*args, **kwargs)
+        comm = _get_comm(args)
+        rv0 = None
+        if comm.rank == 0:
+            try:
+                rv0 = func(*args, **kwargs)
+            except Exception:
+                traceback.print_last()
```
neutrinoceros, line 345: style nit
```suggestion
        if comm.rank == 0:
            try:
                rv0 = func(*args, **kwargs)
            except Exception:
                traceback.print_last()
                rv0 = None
```

yt/loaders.py
neutrinoceros, line 1653: Not important since it's private API, but this very slightly triggers me within the context of the loaders module (`yt.load(fn=...)` being deprecated)
```suggestion
    loadable_path = _get_sample_data(fn, progressbar=progressbar, timeout=timeout)
```

neutrinoceros (commented):
I like the approach (to be honest I can't see any other way, really) ! I got a couple small comments but overall this is looking great, thank you !

yt/loaders.py
chrishavlin: hah, apologies for the trigger :)

yt/utilities/parallel_tools/parallel_analysis_interface.py
chrishavlin: yes, it is! will delete. was left over from a copy/paste from an initial attempt.

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -326,6 +326,32 @@ def barrierize(*args, **kwargs):
...
+    def root_only(*args, **kwargs):
+        if not parallel_capable:
+            return func(*args, **kwargs)
+        comm = _get_comm(args)
+        if comm.rank == 0:
+            try:
+                rv0 = func(*args, **kwargs)
+            except Exception:
+                traceback.print_last()
+                rv0 = None
```
neutrinoceros, line 345: Now that I see it with the style I suggested, I suspect this would cause `UnboundLocalError` when run in parallel (`rv0` is not defined on non-root processes). I wonder how the broadcast function works. Does it know that `None` is only a placeholder value, or does it always broadcast from root ?

chrishavlin: oh, right, that's why it was defined as None to begin with. 

I should be able to easily test if it causes problems. I suspect it does need to be defined -- I think it will always broadcast from root but I think the python object problem needs to be defined in some way on every process first?

chrishavlin: Ya, `rv0` needs to be defined or you get errors on the non-root processors. I'll switch that back in the next push. 

> I wonder how the broadcast function works. Does it know that None is only a placeholder value, or does it always broadcast from root ?

It always broadcasts from a single processor to the others: by default it assumes rank 0 is root, but you can actually specify a different rank to treat as root (i.e., the call signature for `mpi_bcast`  is `def mpi_bcast(self, data, root=0):`). But it does overwrite an existing object, so it needs to be first defined on all ranks (this is more of an MPI/mpi4py thing than it is a yt thing, check out the [examples for collective communication in mpi4py](https://mpi4py.readthedocs.io/en/stable/tutorial.html#collective-communication)).

neutrinoceros: awesome, sorry I introduced this error !

yt/utilities/parallel_tools/parallel_analysis_interface.py
chrishavlin: done

yt/loaders.py
```diff
@@ -1650,7 +1651,9 @@ def load_sample(
     - Corresponding sample data live at https://yt-project.org/data
 
     """
-    loadable_path = _get_sample_data(fn, progressbar=progressbar, timeout=timeout)
+    loadable_path, kwargs = _get_sample_data(
```
neutrinoceros, line 1654: I would slightly prefer we use a different variable name for the output `kwarg` here (to avoid confusion when debugging)

chrishavlin: sure! how about `load_kwargs` ?

neutrinoceros: works for me

yt/loaders.py
```diff
@@ -1569,7 +1536,6 @@ def load_sample(
         )
 
     kwargs = {**specs["load_kwargs"], **kwargs}
```
chrishavlin, line 1538: @neutrinoceros , you wanna rename this `kwargs` as well? e.g., 
```suggestion
    load_kwargs = {**specs["load_kwargs"], **kwargs}
```
(or whatever name we decide on in the other comment thread)

neutrinoceros: It would make sense, sure !

neutrinoceros (approved):

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

PR #4651, opened 2023-08-25
OPT: Optimize GAMER IO

matthewturk:
This PR includes some changes to optimize GAMER IO.  I will be updating it in individual increments with new attempts at optimization, as while my tests here suggest it is indeed still producing correct results, I would like to see both the speed changes and the answers from Fido.

### DIFF
--- a/yt/frontends/gamer/io.py
+++ b/yt/frontends/gamer/io.py
@@ -121,24 +121,20 @@ class IOHandlerGAMER(BaseIOHandler):
                     start = (gs[0].id) * self.pgroup
                     end = (gs[-1].id + 1) * self.pgroup
                     buf = ds[start:end, :, :, :]
-                    ngrid = len(gs)
-                    data = np.empty((ngrid, ps2, ps2, ps2), dtype=self._field_dtype)
-
-                    for g in range(ngrid):
-                        pid0 = g * self.pgroup
-                        data[g, 0:ps1, 0:ps1, 0:ps1] = buf[pid0 + 0, :, :, :]
-                        data[g, 0:ps1, 0:ps1, ps1:ps2] = buf[pid0 + 1, :, :, :]
-                        data[g, 0:ps1, ps1:ps2, 0:ps1] = buf[pid0 + 2, :, :, :]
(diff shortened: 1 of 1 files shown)


yt/frontends/gamer/io.py
```diff
@@ -121,24 +121,20 @@ def _read_fluid_selection(self, chunks, selector, fields, size):
...
+                        pid0 = i * self.pgroup
+                        data[0:ps1, 0:ps1, 0:ps1] = buf[pid0 + 0, :, :, :]
+                        data[0:ps1, 0:ps1, ps1:ps2] = buf[pid0 + 1, :, :, :]
+                        data[0:ps1, ps1:ps2, 0:ps1] = buf[pid0 + 2, :, :, :]
+                        data[ps1:ps2, 0:ps1, 0:ps1] = buf[pid0 + 3, :, :, :]
+                        data[0:ps1, ps1:ps2, ps1:ps2] = buf[pid0 + 4, :, :, :]
+                        data[ps1:ps2, ps1:ps2, 0:ps1] = buf[pid0 + 5, :, :, :]
+                        data[ps1:ps2, 0:ps1, ps1:ps2] = buf[pid0 + 6, :, :, :]
+                        data[ps1:ps2, ps1:ps2, ps1:ps2] = buf[pid0 + 7, :, :, :]
+                        offset += g.select(selector, data, rv[field], offset)
```
matthewturk, line 136:
```suggestion
                        offset += g.select(selector, data.T, rv[field], offset)
```

jzuhone:
LGTM

jzuhone (dismissed):

matthewturk:
Looking at the early results from Fido I'm wondering if perhaps the caching is breaking things.

yt/frontends/gamer/io.py
```diff
@@ -91,13 +92,39 @@ def _read_particle_fields(self, chunks, ptf, selector):
                     data = self._group_particle[field][start:end]
                     yield (ptype, field), data[mask]
 
+    @lru_cache(maxsize=8)  # noqa
```
neutrinoceros, line 95: why the noqa comment ? (making it specific to the warning(s) you want to filter, e.g. `noqa: E555`, would answer my question)

PR #4655, opened 2023-08-27
BLD: re-export ewah-bool-wrap module to avoid having ewah-bool-utils as a runtime dependency

### LINKED ISSUE #4654: DEP: ewah-bool-utils as a build-only dependency (?)
As of yt 4.2.2, ewah-bool-utils is a build-time dependency *and* a runtime dependency to yt. In working on various transitions happening this year (Cython 3, Python 3.12, numpy 2.0), I've come to think that this makes maintenance of *both* packages harder than it needs to be.

There's only *one* line that makes ewah-bool-utils a runtime requirement to yt. Here it is:
https://github.com/yt-project/yt/blob/db6862e2bd476471aa519ab58e1c49dc8a364874/yt/geometry/particle_geometry_handler.py#L8C3-L8C3

I would like to try and find a way to avoid that.

neutrinoceros:
tentatively close #4654 

If this is successful (tests stay green), I'll attempt to trim down the unused bits.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,6 @@ keywords = [
 requires-python = ">=3.8"
 dependencies = [
     "cmyt>=1.1.2",
-    "ewah-bool-utils>=1.0.2",
     "ipywidgets>=8.0.0",
     "matplotlib>=3.5",
     "more-itertools>=8.4",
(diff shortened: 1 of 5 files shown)


neutrinoceros:
I think I understand [Matt's comment](https://github.com/yt-project/yt/pull/2711/files#r1059362637) now; `cimport` statements are not "compiled away", but end up dynamically linking libraries, which means that ewah-bool-utils cannot be reduced to a build-time dependency. Closing this and the corresponding issue.

PR #4658, opened 2023-09-01
allow annotate_grids to force cmap to respect max_levels

zingale:
## PR Summary

At the moment, when you use `annotate_grids()` to show boxes and compare simulations with different
numbers of refinement levels, the colors don't correspond to one another across simulations.  For example,
here are 4 different simulations.  One has base + 1 level, then next 2 levels, then 3, and finally 4 levels of refinement
(all are jumps of 2x).  Notice that the color for level 1 is not the same across all simulations:

subch_Temp_res_compare_orig

Here's the output with this PR if I pass `use_max_level_in_cmap=True` to `annocate_grids()`.

subch_Temp_res_compare

Note: I am unsure why the original behavior was desired, but I left it in place.

## PR Checklist

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -1054,6 +1054,7 @@ class GridBoundaryCallback(PlotCallback):
         cmap="B-W LINEAR_r",
         edgecolors=None,
         linewidth=1.0,
+        use_max_level_in_cmap=False
     ):
         self.alpha = alpha
         self.min_pix = min_pix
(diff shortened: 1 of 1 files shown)


zingale:
I also don't understand the current way it works, but I didn't want to remove it in case someone was relying on "spacebar heating"

neutrinoceros:
Nice reference !
I think we could use a third pair of eyes here. As far as I am concerned, I feel like this kind of "breaking" change could even be acceptable if shipped in a feature release. Let's aim this at the upcoming 4.3 release and see what others think !

matthewturk:
Yeah, we can remove the old behavior, I think.  Thanks for submitting this!

neutrinoceros (approved):
Thanks @matthewturk and @zingale !

PR #4659, opened 2023-09-03
DEP: drop support for CPython 3.8

### LINKED ISSUE #4302: TST: how to drop CPython 3.8 ?
⚠️ **Disclaimer**
_I am not proposing to drop support for CPython 3.8 **now**. I just want to point out why this will be difficult because it'll need to happen eventually, and the sooner we prepare for it, the easier it'll be._

## Timeline: when can we expect to drop it ?
To ensure the total number of versions we support at any given time stays bounded, and [as long as new versions of CPython keep dropping yearly](https://peps.python.org/pep-0602/), we need to drop 1 version per year to keep up.
Following this motivation, the drop typically happens around the same time of year that a new version is released (see #2917 and #4101 for history), so the typical expectation would be that we do this around October 2023.

Were we to closely follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (which so far we have not, so it shouldn't be expected from us), it could be done as soon as April 2023.

## Problem
CPython 3.8 is our oldest supported version at the moment, and is currently used for testing on Jenkins. This means we need to upgrade the Python runtime on CI _before_ we are allowed to bump `requires-python` in `pyproject.toml`. I expect this may be difficult and/or time consuming because all our answer-tests and image tests that rely on actual simulation data are run there, and many (not sure exactly which) of them use `pickle` for reference storage. Now, Python objects saved to disk with `pickle` are _not_ guaranteed to be interoperable with different versions of the `python` runtime, so it is possible that some references cannot be salvaged and must be regenerated from scratch (then validated manually ?).

To make things worse, the most we  can upgrade these jobs is to Python 3.9, because newer versions (3.10+) are just not compatible with the test framework used for these tests (`nose`).

It follows that dropping 3.9 (around October 2024 ?) will not be possible without leaving some of our test suite behind unless we can finish the migration to pytest before then, but that's an other story (sort of).

So I'd like to open the discussion as soon as possible, mostly with @Xarthisius who's still in the delicate position of being our only Jenkins maintainer, and @matthewturk, who's most aware of the issues we're having with transition to pytest.

I *think* (maybe naively ?) that we can afford doing only minimal manual validation for answers on Python 3.9 (the code base has been extensively tested on Python 3.9 for years now), which would make the process easier.
The better solution of course would be to push for the migration to `pytest` but that seems like an unreasonable amount of work for this timeline, given the history of this project.

Thoughts ?

neutrinoceros:
Following [my recent email](https://mail.python.org/archives/list/yt-dev@python.org/thread/F36HH6UX27KT5Z37R3WKSZ2FRP5BMTVG/), closes #4302

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -44,3 +44,6 @@ ef783151bfd7c6777fa25e9e06f95fe47653b3aa
 
 # apply linting to ipynb files
 ec8bb45ea1603f3862041fa9e8ec274afd9bbbfd
+
+# auto upgrade typing idioms from Python 3.8 to 3.9
+4cfd370a8445abd4620e3853c2c047ee3d649fd7
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -33,7 +33,7 @@ jobs:
         test-runner: [pytest]
         include:
           - os: ubuntu-latest
-            python-version: '3.8'
+            python-version: '3.9'
             dependencies: minimal
             tests-type: unit
             test-runner: pytest
--- a/.github/workflows/type-checking.yaml
+++ b/.github/workflows/type-checking.yaml
@@ -29,10 +29,10 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        # run with oldest support python version
+        # run with oldest supported python version
         # so that we always get compatible versions of
         # core dependencies at type-check time
-        python-version: '3.8'
+        python-version: '3.9'
 
     - name: Build
       run: |
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -35,7 +35,7 @@ jobs:
         with:
           output-dir: dist
         env:
-          CIBW_BUILD: "cp38-* cp39-* cp310-* cp311-* cp312-*"
+          CIBW_BUILD: "cp39-* cp310-* cp311-* cp312-*"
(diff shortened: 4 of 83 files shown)


neutrinoceros:
@fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros:
@yt-fido test this please

cphyc (approved):
LGTM, thanks for the hard work! For reference, could you comment on the upgrades required to make `nose` work?

neutrinoceros:
@cphyc As far as I know nose is just patched as follow
```
python3.10 -m pip install nose nose-exclude
find .venv/lib/python*/site-packages/nose -name "*.py" -exec sed -i -e 's/collections.Callable/collections.abc.Callable/g' {} \;
```

see https://github.com/yt-project/yt/issues/4302#issuecomment-1696143424
Maybe @Xarthisius has more insight on this.

Xarthisius:
> @cphyc As far as I know nose is just patched as follow
> 
> ```
> python3.10 -m pip install nose nose-exclude
> find .venv/lib/python*/site-packages/nose -name "*.py" -exec sed -i -e 's/collections.Callable/collections.abc.Callable/g' {} \;
> ```
> 
> see [#4302 (comment)](https://github.com/yt-project/yt/issues/4302#issuecomment-1696143424) Maybe @Xarthisius has more insight on this.

That's everything I did.

Issue #4630, opened 2023-08-11
ENH: Improve performance of or remove `_identify_base_chunk`


matthewturk:
In looking at parallelism, @cindytsai identified a trend for AMR datasets that the graph is roughly flat.  This seems to be because of `_identify_base_chunk`, which is a rather large and all-consuming operation for many of these datasets.

There are two ways to address this:

 1. Remove the need for `_identify_base_chunk`.  This is what we should strive for.
 2. Make it more performant, and enable parallelism in it.

The second option is a much more tractable one.  The main accomplishment of `_identify_base_chunk` is to identify the, well, base chunk from with other chunks will be divided, but this could in principle be done in a distributed manner using intrinsic yt parallelism operations.

cc @chrishavlin @cindytsai

matthewturk:
For reference, here is the current content of `_identify_base_chunk` for `GridIndex`:

```python
    def _identify_base_chunk(self, dobj):
        fast_index = None
        if dobj._type_name == "grid":
            dobj._chunk_info = np.empty(1, dtype="object")
            dobj._chunk_info[0] = weakref.proxy(dobj)
        elif getattr(dobj, "_grids", None) is None:
            gi = dobj.selector.select_grids(
                self.grid_left_edge, self.grid_right_edge, self.grid_levels
            )
            if any(g.filename is not None for g in self.grids[gi]):
                _gsort = _grid_sort_mixed
            else:
                _gsort = _grid_sort_id
            grids = sorted(self.grids[gi], key=_gsort)
            dobj._chunk_info = np.empty(len(grids), dtype="object")
            for i, g in enumerate(grids):
                dobj._chunk_info[i] = g
        # These next two lines, when uncommented, turn "on" the fast index.
        # if dobj._type_name != "grid":
        #    fast_index = self._get_grid_tree()
        if getattr(dobj, "size", None) is None:
            dobj.size = self._count_selection(dobj, fast_index=fast_index)
        if getattr(dobj, "shape", None) is None:
            dobj.shape = (dobj.size,)
        dobj._current_chunk = list(
            self._chunk_all(dobj, cache=False, fast_index=fast_index)
        )[0]
```

I want to note that the 'fast' index should make this somewhat less important, but I am still working to get that in a functional state.  So here note a few things:

 - If it's a grid collection, with no real subselection, it sets `dobj._chunk_info` to be an array of `dtype="object"` where each entry is one of the grids.  This is not the principle use case, however.
 - Otherwise, two things happen that are important:
[...]

PR #4673, opened 2023-09-22
PERF: return the count from fill_mask_regular_grid

chrishavlin:
While working on #4630 , I noticed that we could avoid an np.sum() by having fill_mask_regular_grid return the selection count that it's already tracking. Savings are small, but I thought it was worth the change to avoid recalculating a number we already have.


chrishavlin:
Oops, most likely I didn't catch all the uses, will come back to this on Monday.

chrishavlin:
oh, actually an obvious error, will fix now...

matthewturk (approved):

matthewturk:
Nice!

PR #4677, opened 2023-09-27
TST: restore global state modified during tests

yut23:
## PR Summary

Restore numpy's error handling settings and yt's logging level during module teardown.

I found that `test_fisheye` in `yt/visualization/volume_rendering/tests/test_vr_cameras.py` was failing when I ran just the visualization tests, but not the full test suite.

## PR Checklist

### DIFF
--- a/yt/utilities/lib/tests/test_alt_ray_tracers.py
+++ b/yt/utilities/lib/tests/test_alt_ray_tracers.py
@@ -6,11 +6,13 @@ from yt.testing import amrspace
 from yt.utilities.lib.alt_ray_tracers import _cyl2cart, cylindrical_ray_trace
 
 left_grid = right_grid = amr_levels = center_grid = data = None
+old_settings = None
 
 
 def setup():
     # set up some sample cylindrical grid data, radiating out from center
-    global left_grid, right_grid, amr_levels, center_grid, data
+    global left_grid, right_grid, amr_levels, center_grid, data, old_settings
+    old_settings = np.geterr()
     np.seterr(all="ignore")
     l1, r1, lvl1 = amrspace([0.0, 1.0, 0.0, -1.0, 0.0, 2 * np.pi], levels=(7, 7, 0))
     l2, r2, lvl2 = amrspace([0.0, 1.0, 0.0, 1.0, 0.0, 2 * np.pi], levels=(7, 7, 0))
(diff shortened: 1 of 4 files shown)


yt/utilities/tests/test_set_log_level.py
```diff
@@ -2,6 +2,19 @@
 
 from yt.utilities.logger import set_log_level
 
+old_level = None
```
neutrinoceros, line 5: is this line actually needed ? I feel like it might not be (though maybe a `NameError` is more confusing than an `AttributeError` ?)

yt/utilities/lib/tests/test_alt_ray_tracers.py
```diff
@@ -5,12 +5,13 @@
 from yt.testing import amrspace
 from yt.utilities.lib.alt_ray_tracers import _cyl2cart, cylindrical_ray_trace
 
-left_grid = right_grid = amr_levels = center_grid = data = None
+left_grid = right_grid = amr_levels = center_grid = data = old_settings = None
```
neutrinoceros, line 8: I'm not a fan of these all living on the same line. No harm done in keeping the existing one (though I'd also approve expanding it) but I'd rather have the new binding on a separate line.

neutrinoceros (dismissed):
good catch, thank you !
here are a couple minor comments but I'd merge it as is

neutrinoceros:
> I'd merge it as is

well... when tests pass, I guess 😅

yt/utilities/tests/test_set_log_level.py
yut23: It works fine without, but pylint complains about it

yut23:
Huh, I have no idea why only py3.9 is failing. I wasn't able to get `test_fisheye` to pass locally on 3.10, 3.11, or 3.12.

yt/utilities/tests/test_set_log_level.py
neutrinoceros: [personal opinion] I don't really care for pylint, especially in yt where we don't use it. [/personal opinion]
Anyway I didn't mean to rant, this is fine as is :)

neutrinoceros:
I think I saw this error being flaky a while back in a different context. If I recall correctly this was due to some uninitialised memory being involved in an array division, which *may* trigger this `RuntimeWarning` but only sometimes.

yut23:
I tried running under valgrind, and the test passed with no memory errors. I'm pretty sure valgrind is just swallowing the FPE signals before numpy can handle them, though; I broke into a pdb session and the image array is all zeros with and without valgrind. I also checked against yt 3.6.1 and got the same error.

yut23:
Alright, adding `log_fields=[False]` makes it produce a non-empty image. Should I make a separate PR for that?

neutrinoceros:
> Should I make a separate PR for that?

I don't think that's necessary.

@yt-fido test this please

neutrinoceros:
@yt-fido test this please

neutrinoceros (approved):

PR #4678, opened 2023-09-29
[WIP] Experimentally turn off precounting AMR datasets

matthewturk:
Tests done by me, @cindytsai and @chrishavlin suggest that parallelism *in particular* is greatly limited by the amount of duplicated indexing that goes on across MPI tasks.

This is an experiment to see how we fare with turning off pre-counting, which should greatly reduce the amount of time spent in the counting routines, and potentially speed things up overall.


matthewturk:
Interestingly, the way the grid selections are set up doesn't lend itself to the resizing I'm attempting.  I do think this is still quite a nice area for improvement, though, so I will try to lift the references up a level to perform this surgery more carefully.

matthewturk:
Resize behaves a bit strangely if the original shape is empty and you resize it to a single value.  So this shows up when we have chunks consisting of a single grid with a single value selected, which happens inside `streamlines.py`.  I've fixed this in this particular location, but there are remaining issues with nodal data for enzo.

matthewturk:
@yt-fido test this please

PR #4686, opened 2023-10-05
minor bugfix in cholla frontend

mabruzzo:
## PR Summary

Previously, the ``domain_right_edge`` attribute of ``ChollaDataset`` was initialized by assigning it the values of the ``"domain"`` attribute stored in the hdf5 file. This patch adjusts the logic so that the ``domain_right_edge`` attribute is now the sum of that array and the array stored in the ``domain_left_edge`` attribute.

## PR Checklist

Given the simplicity of this fix, I don't really think there's any need for new documentation. Plus adding a test is probably not worth the effort (we would need to either introduce a new cholla dataset OR duplicate and mutate the existing dataset).

### DIFF
--- a/yt/frontends/cholla/data_structures.py
+++ b/yt/frontends/cholla/data_structures.py
@@ -103,7 +103,9 @@ class ChollaDataset(Dataset):
             attrs = h5f.attrs
             self.parameters = dict(attrs.items())
             self.domain_left_edge = attrs["bounds"][:].astype("=f8")
-            self.domain_right_edge = self.domain_left_edge + attrs["domain"][:].astype("=f8")
+            self.domain_right_edge = self.domain_left_edge + attrs["domain"][:].astype(
+                "=f8"
+            )
             self.dimensionality = len(attrs["dims"][:])
             self.domain_dimensions = attrs["dims"][:].astype("=f8")
             self.current_time = attrs["t"][:]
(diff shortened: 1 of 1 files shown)


matthewturk (approved):

yt/frontends/cholla/data_structures.py
```diff
@@ -103,7 +103,9 @@ def _parse_parameter_file(self):
             attrs = h5f.attrs
             self.parameters = dict(attrs.items())
             self.domain_left_edge = attrs["bounds"][:].astype("=f8")
-            self.domain_right_edge = self.domain_left_edge + attrs["domain"][:].astype("=f8")
+            self.domain_right_edge = self.domain_left_edge + attrs["domain"][:].astype(
+                "=f8"
+            )
```
neutrinoceros, line 108: This is one of the rare cases where black makes things objectively worse looking... here's a possible alternative
```suggestion
            dre = attrs["bounds"][:] + attrs["domain"][:]
            self.domain_right_edge = dre.astype("=f8")
```

chummels (approved):
LGTM.  Thanks for the catch!

neutrinoceros:
Ok since Cameron is happy, I think we can ignore my style nit and merge as in. Thanks everyone !

mabruzzo:
@neutrinoceros - sorry about not addressing that style change (I agree that your version looks better). The last few weeks have been a little hectic. I may try to implement that change in #4702 (since I'm already touching nearby code)

neutrinoceros:
No worries, we use linters and formatters mainly so we can _not_ discuss style nits and focus on what's actually important ;-)

PR #4687, opened 2023-10-05
BLD: require Cython 3.0.3 to fix known issues on CPython 3.12

### LINKED ISSUE #4675: BLD: (yt 4.3) cp312 wheels require Cython 3.0.3
### Context
https://github.com/cython/cython/issues/5724
https://github.com/cython/cython/pull/5725

In short: binaries built with Cython 3.0.2 + Python 3.12 cause a memory leak, so we'd better wait for Cython 3.0.3 to be released so we can bump our requirements, before publishing cp312 wheels.
I'll consider this a release blocker for yt 4.3.0

neutrinoceros:
close #4675

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -47,20 +47,14 @@ jobs:
       # incompatibility with Cython 3
       # see https://github.com/yaml/pyyaml/issues/601
       # and https://github.com/cython/cython/issues/4568
-      # Cython 3.0.2 has a defect on Python 3.12, so we install the -fixed- dev version
-      # from source for now.
-      # see https://github.com/cython/cython/issues/5724
-      # Cython can be installed from PyPI binaries again when the patch is included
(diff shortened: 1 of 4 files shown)


chrishavlin (approved):
Looks good to me.

Since this PR also introduces a minimal unit test with 3.12, it might be useful to add a new issue just to track the progress of moving to full testing with 3.12 (if there's not already one that I've missed). What do you think?

neutrinoceros:
I think this could be useful, though we're almost there already, and it looks like our optional dependencies are moving much faster than in recent years, so I'm actually hopeful that we can just finish it in a couple days.

neutrinoceros:
@chrishavlin I've opened #4689 and linked to it here

cphyc (approved):
Also LGTM! I'd be happy once we fix future incompatibilities with Cython 3.1 though :)

neutrinoceros:
it might actually be a no-op, depending on how https://github.com/cython/cython/issues/4310 turns out

PR #4693, opened 2023-10-09
BUG: fix load_sample with no arguments

### LINKED ISSUE #4691: BUG: calling `yt.load_sample` with no arguments is broken on main
### Bug report

**Bug summary**

This is a regression in yt 4.3 (dev)

I suspect it came from #4646 (cc @chrishavlin). It also shows that this path is currently not tested.

**Code for reproduction**

```shell
$ python -c "import yt; yt.load_sample()`
```

**Actual outcome**

```python-traceback
One can see which sample datasets are available at: https://yt-project.org/data
or alternatively by running: yt.sample_data.api.get_data_registry_table()
  File "<string>", line 1, in <module>
  File "/Users/robcleme/dev/yt-project/yt/yt/loaders.py", line 1666, in load_sample
    loadable_path, load_kwargs = _get_sample_data(
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot unpack non-iterable NoneType object
```

**Expected outcome**
with yt 4.2.2, no error is raised
```
One can see which sample datasets are available at: https://yt-project.org/data
or alternatively by running: yt.sample_data.api.get_data_registry_table()
```

**Version Information**
  * Operating System: any
  * Python Version: any
  * yt version: 4.3.0dev0
  * Other Libraries (if applicable): N/A

chrishavlin:
Closes #4691 

Was easiest to just move the check for `fn is None` out of `_get_sample_data` to the main `load_sample`. Also added a minimal test.

### DIFF
--- a/yt/tests/test_load_sample.py
+++ b/yt/tests/test_load_sample.py
@@ -199,4 +199,4 @@ def test_data_dir_broken():
 
 
 def test_filename_none():
-    assert load_sample(None) is None
+    assert load_sample() is None
--- a/yt/tests/test_load_sample.py
+++ b/yt/tests/test_load_sample.py
@@ -198,5 +198,7 @@ def test_data_dir_broken():
         load_sample("ToroShockTube")
 
 
-def test_filename_none():
+def test_filename_none(capsys):
     assert load_sample() is None
+    captured = capsys.readouterr()
+    assert "yt.sample_data.api.get_data_registry_table" in captured.err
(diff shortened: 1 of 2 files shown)


yt/tests/test_load_sample.py
```diff
@@ -196,3 +196,7 @@ def test_data_dir_broken():
     )
     with pytest.warns(UserWarning, match=msg):
         load_sample("ToroShockTube")
+
+
+def test_filename_none():
+    assert load_sample() is None
```
chrishavlin, line 202: is there a `with pytest. ???` variation that i can use here to check the message that gets printed?

neutrinoceros (approved):
thank you !

PR #4696, opened 2023-10-10
Allow names for the SPH particle type other than "io" in the stream frontend

jzuhone:
## PR Summary

Currently, the stream frontend checks for the presence for a SPH particle type by checking for simultaneous presence of the fields `("io", "smoothing_length")` and `("io","density")`, which mandates that the SPH particle type be `"io"`. This PR makes the specification of this field type more flexible. It also checks that only one SPH type is present (because we currently do not support more than one SPH particle type in general).

I edited the "Loading Generic Particle Data" doc to reflect this, and fixed it because it was no longer working in any case. 

## PR Checklist

### DIFF
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -560,10 +560,12 @@ class StreamParticlesDataset(StreamDataset):
             axis_order=axis_order,
         )
         fields = list(stream_handler.fields["stream_file"].keys())
-        # This is the current method of detecting SPH data.
-        # This should be made more flexible in the future.
-        if ("io", "density") in fields and ("io", "smoothing_length") in fields:
-            self._sph_ptypes = ("io",)
+        sph_ptypes = []
+        for ptype in self.particle_types:
+            if (ptype, "density") in fields and (ptype, "smoothing_length") in fields:
+                sph_ptypes.append(ptype)
+        if len(sph_ptypes) > 0:
+            self._sph_ptypes = tuple(sph_ptypes)
 
     def add_sph_fields(self, n_neighbors=32, kernel="cubic", sph_ptype="io"):
         """Add SPH fields for the specified particle type.
--- a/yt/frontends/stream/data_structures.py
+++ b/yt/frontends/stream/data_structures.py
@@ -564,8 +564,10 @@ class StreamParticlesDataset(StreamDataset):
         for ptype in self.particle_types:
             if (ptype, "density") in fields and (ptype, "smoothing_length") in fields:
                 sph_ptypes.append(ptype)
-        if len(sph_ptypes) > 0:
+        if len(sph_ptypes) == 1:
             self._sph_ptypes = tuple(sph_ptypes)
+        elif len(sph_ptypes) > 1:
+            raise ValueError("Multiple SPH particle types are currently not supported!")
 
     def add_sph_fields(self, n_neighbors=32, kernel="cubic", sph_ptype="io"):
         """Add SPH fields for the specified particle type.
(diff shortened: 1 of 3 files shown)


neutrinoceros:
I'll review this over the weekend !

yt/frontends/stream/data_structures.py
```diff
@@ -560,10 +560,14 @@ def __init__(
...
-        if ("io", "density") in fields and ("io", "smoothing_length") in fields:
-            self._sph_ptypes = ("io",)
+        sph_ptypes = []
+        for ptype in self.particle_types:
+            if (ptype, "density") in fields and (ptype, "smoothing_length") in fields:
+                sph_ptypes.append(ptype)
+        if len(sph_ptypes) == 1:
+            self._sph_ptypes = tuple(sph_ptypes)
+        elif len(sph_ptypes) > 1:
+            raise ValueError("Multiple SPH particle types are currently not supported!")
```
neutrinoceros, line 570: It's probably unlikely, but I'm worried this might break existing applications. A warning + backward compatible behaviour would be more suited.

yt/frontends/stream/tests/test_stream_particles.py
```diff
@@ -218,6 +218,53 @@ def test_load_particles_types():
...
+        ("dm", "particle_position_z"): np.random.random(size=num_particles),
+        ("dm", "particle_velocity_x"): np.random.random(size=num_particles),
+        ("dm", "particle_velocity_y"): np.random.random(size=num_particles),
+        ("dm", "particle_velocity_z"): np.random.random(size=num_particles),
+        ("dm", "particle_mass"): np.ones(num_particles),
+    }
+
+    ds = load_particles(data)
+
+    assert set(ds.particle_types) == {"gas", "dm"}
```
neutrinoceros, line 245: I wondered for a minute why the attribute was converted to `set` here, I assume it's because we don't care about order ? Also, do we actually care about exact set equality ?
In any case, I suggest adding a comment.

neutrinoceros (commented):
Seems alright. I just have a couple questions

yt/frontends/stream/data_structures.py
jzuhone: It's never been possible for anyone to have multiple SPH `ptypes`, so there should not be anything broken. We should raise an error here because a lot of things in yt assume one SPH type.

neutrinoceros: it's never been supported, yes, but the existing logic effectively ignores other sph types if any; so this error would break existing applications that have them. I guess such applications would actually be silently broken and it's good that this makes their bug apparent ?

jzuhone: @neutrinoceros the point is that the code in `main` only recognizes one SPH type, and that is `io`, whereas this supports other type names while still only supporting one. If users are attempting to use multiple SPH ptypes, it is currently failing silently until they try to do something which requires that type to be recognized. So this is in fact an improvement because it's actually raising a specific error telling them what is wrong.

neutrinoceros: ok, so that's what I was suspecting, and in that case I agree that it is an improvement. Thanks for clarifying !

yt/frontends/stream/tests/test_stream_particles.py
matthewturk: We don't care about order.  I think this is OK as it is, especially since we really do want to make sure that the sets are equal, and not just that all of them are included, etc.

neutrinoceros: I wasn't arguing against it, I just wanted to point out that a comment might help make intentions clearer

jzuhone:
@yt-fido test this please

neutrinoceros (approved):

jzuhone:
Anyone else want to take a look at this?

matthewturk (approved):

PR #4699, opened 2023-10-10
BUG: make annotate_sphere and annotate_arrow safe when run after plot invalidation

### LINKED ISSUE #4698: BUG: annotating a sphere then setting font size in a plot results in a crash
### Bug report

**Bug summary**

Discovered by [RevathiJambunathan](https://github.com/RevathiJambunathan). She further reported that this might be a regression in yt 4.2.x

**Code for reproduction**

```python
import yt

sl = yt.SlicePlot(ds, 0, ("gas", "density"))

sl.annotate_sphere(
    [0, 0, 0],
    radius=(0.2, "Mpc"),
    circle_args={"color": "white", "linewidth": 2},
)
sl.set_font_size(24) # everything works fine if this line is dropped *or* moved before sl.annotate_sphere(...)

sl.save("/tmp/")
```

**Actual outcome**

```
/Users/robcleme/dev/yt-project/yt/yt/loaders.py:1670: UserWarning: This dataset appears to be of type EnzoDataset, but the following requirements are currently missing: libconf
Please verify your installation.
  return load(loadable_path, **load_kwargs)
Parsing Hierarchy : 100%|███████████████████████████████████████████████████████████████████████████████████| 173/173 [00:00<00:00, 12522.47it/s]
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/plot_window.py", line 1285, in run_callbacks
    callback(cbw)
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/plot_modifications.py", line 2041, in __call__
    self.radius = self.radius.to(units)
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/robcleme/.pyenv/versions/yt-dev/lib/python3.12/site-packages/unyt/array.py", line 947, in to
    return self.in_units(units, equivalence=equivalence, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/robcleme/.pyenv/versions/yt-dev/lib/python3.12/site-packages/unyt/array.py", line 874, in in_units
    (conversion_factor, offset) = self.units.get_conversion_factor(
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/robcleme/.pyenv/versions/yt-dev/lib/python3.12/site-packages/unyt/unit_object.py", line 694, in get_conversion_factor
    return _get_conversion_factor(self, other_units, dtype)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/robcleme/.pyenv/versions/yt-dev/lib/python3.12/site-packages/unyt/unit_object.py", line 939, in _get_conversion_factor
    raise UnitConversionError(
unyt.exceptions.UnitConversionError: Cannot convert between '1/code_length' (dim '1/(length)') and 'code_length' (dim '(length)').

The above exception was the direct cause of the following exception:

  File "/Users/robcleme/dev/yt-project/yt/t.py", line 14, in <module>
    sl.save("/tmp/")
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/_commons.py", line 162, in newfunc
    self._setup_plots()
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/plot_window.py", line 1226, in _setup_plots
    self.run_callbacks()
  File "/Users/robcleme/dev/yt-project/yt/yt/visualization/plot_window.py", line 1289, in run_callbacks
    raise YTPlotCallbackError(callback._type_name) from e
yt.utilities.exceptions.YTPlotCallbackError: annotate_sphere callback failed
```

**Expected outcome**

No crash

**Version Information**
  * Operating System: macOS
  * Python Version: 3.11.4
  * yt version: 4.3.0

chrishavlin:
Closes #4698

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -1810,7 +1810,7 @@ class ArrowCallback(PlotCallback):
         width = self.width * plot_diag
         head_width = self.head_width * plot_diag
         head_length = None
-        if self.head_length:
+        if self.head_length is not None:
             head_length = self.head_length * plot_diag
 
         if self.starting_pos is not None:
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -2032,7 +2032,8 @@ class SphereCallback(PlotCallback):
 
         if is_sequence(self.radius):
             self.radius = plot.data.ds.quan(self.radius[0], self.radius[1])
-            self.radius = np.float64(self.radius.in_units(plot.xlim[0].units))
+            self.radius = self.radius.in_units(plot.xlim[0].units)
+
         if isinstance(self.radius, YTQuantity):
             if isinstance(self.center, YTArray):
                 units = self.center.units
@@ -2045,16 +2046,18 @@ class SphereCallback(PlotCallback):
         # apply a different transform for a length in the same way
         # you can for a coordinate.
         if self.coord_system == "data" or self.coord_system == "plot":
-            self.radius = self.radius * self._pixel_scale(plot)[0]
+            scaled_radius = self.radius * self._pixel_scale(plot)[0]
         else:
-            self.radius /= (plot.xlim[1] - plot.xlim[0]).v
+            scaled_radius /= plot.xlim[1] - plot.xlim[0]
 
         x, y = self._sanitize_coord_system(
             plot, self.center, coord_system=self.coord_system
         )
 
         x, y = self._sanitize_xy_order(plot, x, y)
-        cir = Circle((x, y), self.radius, transform=self.transform, **self.circle_args)
+        cir = Circle(
+            (x, y), scaled_radius.v, transform=self.transform, **self.circle_args
+        )
 
         plot._axes.add_patch(cir)
         if self.text is not None:
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -1806,11 +1806,12 @@ class ArrowCallback(PlotCallback):
         xx0, xx1, yy0, yy1 = self._plot_bounds(plot)
         # normalize all of the kwarg lengths to the plot size
         plot_diag = ((yy1 - yy0) ** 2 + (xx1 - xx0) ** 2) ** (0.5)
-        self.length *= plot_diag
-        self.width *= plot_diag
-        self.head_width *= plot_diag
-        if self.head_length is not None:
-            self.head_length *= plot_diag
+        length = self.length * plot_diag
+        width = self.width * plot_diag
+        head_width = self.head_width * plot_diag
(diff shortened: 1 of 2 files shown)


chrishavlin:
didn't have a chance to run tests locally (other than verifying the reproducer script now works)... but probably should pass? don't think we have a lot of test coverage for this callback beyond simple unit tests. 

Also -- should I add a new test?

chrishavlin:
oh and note that I played around with storing the `scaled_radius` as a new attribute that it could use on the second pass through but then it got a bit too complicated thinking through cases where it might need to be re-calculated. So went with the simple route of always calculating it.

neutrinoceros:
Thank you for taking this on ! I think a new test would be useful too.

chrishavlin:
added a test and fixed the ArrowCallback as well

yt/visualization/plot_modifications.py
```diff
@@ -1806,11 +1806,12 @@ def __call__(self, plot):
...
-        self.length *= plot_diag
-        self.width *= plot_diag
-        self.head_width *= plot_diag
-        if self.head_length is not None:
-            self.head_length *= plot_diag
+        length = self.length * plot_diag
+        width = self.width * plot_diag
+        head_width = self.head_width * plot_diag
+        head_length = None
+        if self.head_length:
```
neutrinoceros, line 1813: I think using truthiness here is a breaking change because it's a change of behaviour when passing `head_length=0`
```suggestion
        if self.head_length is not None:
```

yt/visualization/tests/test_callbacks.py
```diff
@@ -404,6 +404,23 @@ def test_sphere_callback():
...
+def test_invalidated_annotations():
+    # check that annotate_sphere and annotate_arrow succeed on re-running after
+    # an operation that invalidates the plot (set_font_size), see
+    # https://github.com/yt-project/yt/issues/4698
+    with _cleanup_fname() as prefix:
+        ds = fake_amr_ds(fields=("density",), units=("g/cm**3",))
+        p = SlicePlot(ds, "z", ("gas", "density"))
+        p.annotate_sphere([0.5, 0.5, 0.5], 0.1)
+        p.set_font_size(24)
+        assert_fname(p.save(prefix)[0])
```
neutrinoceros, line 416: I think we could speed up these tests by calling `p.render` instead of `p.save` here (yes, that's also true for a lot of existing tests in this module, but of course it's way out of scope to deal with everything here)

chrishavlin: I'll make this change

neutrinoceros (approved):

PR #4702, opened 2023-10-10
Add support for distributed cholla datasets.

mabruzzo:
## PR Summary

This PR adds support for loading Cholla datasets that are distributed over multiple files. Previously, the frontend could only load Cholla datasets after they were concatenated into a single large dataset.

This functionality is currently a little inefficient right now - we need to read in every hdf5 file to figure out the mapping between spatial locations and locations on disk. This seems like something we can easily improve in the future (possibly by having Cholla write out an extra attribute how 3D locations are mapped into 1D).

## PR Checklist

For this PR, I suspect that we will need to upload a new test dataset.  I just had a few questions:
- It's been a while since I've done this. Could someone remind me of the procedure for doing this? 
- Weirdly enough, I get the following message when I run the unit-tests on the main branch. Do you have any idea why this is happening? (For context, the other 3 tests all run)
    > yt/frontends/cholla/tests/test_outputs.py::test_cholla_data SKIPPED (cannot load dataset ChollaSimple/0.h5)
- Is there any preference for unit tests vs answer-tests when it comes to frontends?


neutrinoceros:
> It's been a while since I've done this. Could someone remind me of the procedure for doing this?

you'll need to
- open a pull request on the website repository (see for example https://github.com/yt-project/website/pull/121)
- add an entry to `yt/sample_data_registry.json` (this repo). This is to support loading the new sample dataset with `yt.load_sample`


> Weirdly enough, I get the following message when I run the unit-tests on the main branch. Do you have any idea why this is happening? (For context, the other 3 tests all run)

Maybe that's a bug with `small_patch_amr`. I suggest trying to work on a simplified version of the test and refine it until it doesn't skip, to discover what's happening.

> Is there any preference for unit tests vs answer-tests when it comes to frontends?

I think unit tests should be preferred whenever they suffice for a couple reasons:
- answer tests are currently deeply rooted in the nose test framework (migration to pytest is still ongoing), so adding more of them makes this long lasting migration ever so slightly harder
- fast tests are easier to scale

That said, if what you *need* is some answer tests, go for it !

yt/frontends/cholla/data_structures.py
```diff
@@ -42,23 +73,92 @@ def _detect_output_fields(self):
...
+                i,
+                index=self,
+                level=level,
+                dims=dims_local,
+                filename=fname,
+            )
+
+            self.grid_left_edge[i] = left_frac
+            self.grid_right_edge[i] = right_frac
+            self.grid_dimensions[i] = dims_local
```
matthewturk, line 144:
```suggestion
            self.grid_left_edge[i,:] = left_frac
            self.grid_right_edge[i,:] = right_frac
            self.grid_dimensions[i,:] = dims_local
```

Just for clarity, could we make it obvious that it's setting a slice to the values?

yt/frontends/cholla/io.py
```diff
@@ -14,22 +13,24 @@ def _read_particle_coords(self, chunks, ptf):
...
-                    for field in fields:
-                        ftype, fname = field
-                        values = fh[fname][:].astype("=f8")
-                        nd = grid.select(selector, values, data[field], ind)
-                    ind += nd
-        return data
+    def io_iter(self, chunks, fields):
+        # this is loosely inspired by the implementation used for Enzo/Enzo-E
+        # - those other options use the lower-level hdf5 interface. Unclear
+        #   whether that affords any advantages...
```
matthewturk, line 19: Good question.  I think in the past it did because we avoided having to re-allocate temporary scratch space, but I am not sure that would hold up to current inquiries.  I think the big advantage those have is tracking the groups within the iteration.

yt/frontends/cholla/io.py
```diff
@@ -14,22 +13,24 @@ def _read_particle_coords(self, chunks, ptf):
...
+        fh, filename = None, None
+        for chunk in chunks:
+            for obj in chunk.objs:
+                if obj.filename is None:  # unclear when this case arises...
```
matthewturk, line 23: likely it will not here, unless you manually construct virtual grids

matthewturk, line 23: likely it will not here, unless you manually construct virtual grids

matthewturk (dismissed):
only minor stuff -- looks good otherwise

yt/frontends/cholla/io.py
mabruzzo: Out of curiosity, what is a virtual grid? 

I realize this may be an involved answer - so if you could just point me to a frontend (or other area of the code) using virtual grids, I can probably investigate that on my own.

mabruzzo:
My apologies for taking a while to follow up on this. I plan to circle back in the next week or so.

yt/frontends/cholla/data_structures.py
```diff
@@ -5,19 +5,50 @@
...
-from yt.funcs import setdefaultattr
+from yt.funcs import get_pbar, setdefaultattr
 from yt.geometry.api import Geometry
 from yt.geometry.grid_geometry_handler import GridIndex
 from yt.utilities.on_demand_imports import _h5py as h5py
 
 from .fields import ChollaFieldInfo
 
 
+def _split_fname_proc_suffix(filename: str):
```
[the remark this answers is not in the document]
mabruzzo: Good point -- I overhauled the docstring to try to make it more clear (and I explicitly addressed how it differs from ``os.path.splitext``)

chrishavlin:
hey @mabruzzo if you merge with main again the skipped test should run.

mabruzzo:
pre-commit.ci autofix

matthewturk:
Hi @mabruzzo -- which would you prefer of those options?

PR #4710, opened 2023-10-20
Linear interpolators updates: refactor and new functionality

chrishavlin:
I started using the linear interpolators yesterday and for my purposes it'd be a tab bit easier to re-use the interpolator objects but swap out different table data, so this PR adds that functionality. To do that though, it was easier to refactor the interpolators to have a common base class (and cut out some repetitive code structures in the process). Everything here should be fully backwards compatible. Also happy to split this PR to pull out the refactor from the new functionality (but the new functionality is pretty simple, most of the PR is the refactor).

### DIFF
--- a/yt/utilities/linear_interpolators.py
+++ b/yt/utilities/linear_interpolators.py
@@ -105,11 +105,7 @@ class UnilinearFieldInterpolator(_LinearInterpolator):
         super().__init__(table, field_names, truncate=truncate, store_table=store_table)
         self.x_name = field_names
         if isinstance(boundaries, np.ndarray):
-            self._validate_bin_boundaries(
-                [
-                    boundaries,
-                ]
-            )
+            self._validate_bin_boundaries((boundaries,))
             self.x_bins = boundaries
         else:
             x0, x1 = boundaries
--- a/yt/utilities/linear_interpolators.py
+++ b/yt/utilities/linear_interpolators.py
@@ -63,6 +63,7 @@ class _LinearInterpolator(abc.ABC):
         return return_arrays
 
     def _validate_bin_boundaries(self, boundaries):
+        # boundaries: tuple of ndarrays
         for idim in range(self._ndim):
             if boundaries[idim].size != self.table_shape[idim]:
                 msg = f"{self._field_names[idim]} bins array not the same length as the data."
@@ -161,8 +162,7 @@ class BilinearFieldInterpolator(_LinearInterpolator):
             self.y_bins = np.linspace(y0, y1, table.shape[1], dtype="float64")
         elif len(boundaries) == 2:
             self._validate_bin_boundaries(boundaries)
-            self.x_bins = boundaries[0]
-            self.y_bins = boundaries[1]
+            self.x_bins, self.y_bins = boundaries
         else:
             mylog.error(
                 "Boundaries must be given as (x0, x1, y0, y1) or as (x_bins, y_bins)"
@@ -225,9 +225,7 @@ class TrilinearFieldInterpolator(_LinearInterpolator):
             self.z_bins = np.linspace(z0, z1, table.shape[2], dtype="float64")
         elif len(boundaries) == 3:
             self._validate_bin_boundaries(boundaries)
-            self.x_bins = boundaries[0]
-            self.y_bins = boundaries[1]
-            self.z_bins = boundaries[2]
+            self.x_bins, self.y_bins, self.z_bins = boundaries
         else:
             mylog.error(
                 "Boundaries must be given as (x0, x1, y0, y1, z0, z1) "
@@ -302,10 +300,7 @@ class QuadrilinearFieldInterpolator(_LinearInterpolator):
             self.w_bins = np.linspace(w0, w1, table.shape[3]).astype("float64")
         elif len(boundaries) == 4:
             self._validate_bin_boundaries(boundaries)
-            self.x_bins = boundaries[0]
-            self.y_bins = boundaries[1]
-            self.z_bins = boundaries[2]
-            self.w_bins = boundaries[3]
+            self.x_bins, self.y_bins, self.z_bins, self.w_bins = boundaries
         else:
             mylog.error(
                 "Boundaries must be given as (x0, x1, y0, y1, z0, z1, w0, w1) "
--- a/yt/utilities/linear_interpolators.py
+++ b/yt/utilities/linear_interpolators.py
@@ -10,13 +10,14 @@ class _LinearInterpolator(abc.ABC):
     _ndim: int
     _dim_i_type = "int32"
 
-    def __init__(self, table, truncate=False, *, store_table=True):
+    def __init__(self, table, field_names, truncate=False, *, store_table=True):
         if store_table:
             self.table = table.astype("float64")
         else:
             self.table = None
         self.table_shape = table.shape
         self.truncate = truncate
+        self._field_names = field_names
 
     def _raise_truncation_error(self, data_object):
         mylog.error(
@@ -61,6 +62,13 @@ class _LinearInterpolator(abc.ABC):
 
         return return_arrays
 
+    def _validate_bin_boundaries(self, boundaries):
+        for idim in range(self._ndim):
+            if boundaries[idim].size != self.table_shape[idim]:
+                msg = f"{self._field_names[idim]} bins array not the same length as the data."
+                mylog.error(msg)
+                raise ValueError(msg)
+
 
 class UnilinearFieldInterpolator(_LinearInterpolator):
     _ndim = 1
@@ -94,12 +102,14 @@ class UnilinearFieldInterpolator(_LinearInterpolator):
         field_data = interp(ad)
 
         """
-        super().__init__(table, truncate=truncate, store_table=store_table)
+        super().__init__(table, field_names, truncate=truncate, store_table=store_table)
         self.x_name = field_names
         if isinstance(boundaries, np.ndarray):
-            if boundaries.size != table.shape[0]:
-                mylog.error("Bins array not the same length as the data.")
-                raise ValueError
+            self._validate_bin_boundaries(
+                [
+                    boundaries,
+                ]
+            )
             self.x_bins = boundaries
         else:
             x0, x1 = boundaries
@@ -147,19 +157,14 @@ class BilinearFieldInterpolator(_LinearInterpolator):
         field_data = interp(ad)
 
         """
-        super().__init__(table, truncate=truncate, store_table=store_table)
+        super().__init__(table, field_names, truncate=truncate, store_table=store_table)
         self.x_name, self.y_name = field_names
         if len(boundaries) == 4:
             x0, x1, y0, y1 = boundaries
             self.x_bins = np.linspace(x0, x1, table.shape[0], dtype="float64")
             self.y_bins = np.linspace(y0, y1, table.shape[1], dtype="float64")
         elif len(boundaries) == 2:
-            if boundaries[0].size != table.shape[0]:
-                mylog.error("X bins array not the same length as the data.")
-                raise ValueError
-            if boundaries[1].size != table.shape[1]:
-                mylog.error("Y bins array not the same length as the data.")
-                raise ValueError
+            self._validate_bin_boundaries(boundaries)
             self.x_bins = boundaries[0]
             self.y_bins = boundaries[1]
         else:
@@ -215,7 +220,7 @@ class TrilinearFieldInterpolator(_LinearInterpolator):
         field_data = interp(ad)
 
         """
-        super().__init__(table, truncate=truncate, store_table=store_table)
+        super().__init__(table, field_names, truncate=truncate, store_table=store_table)
         self.x_name, self.y_name, self.z_name = field_names
         if len(boundaries) == 6:
             x0, x1, y0, y1, z0, z1 = boundaries
@@ -223,15 +228,7 @@ class TrilinearFieldInterpolator(_LinearInterpolator):
             self.y_bins = np.linspace(y0, y1, table.shape[1], dtype="float64")
             self.z_bins = np.linspace(z0, z1, table.shape[2], dtype="float64")
         elif len(boundaries) == 3:
-            if boundaries[0].size != table.shape[0]:
-                mylog.error("X bins array not the same length as the data.")
-                raise ValueError
-            if boundaries[1].size != table.shape[1]:
-                mylog.error("Y bins array not the same length as the data.")
-                raise ValueError
-            if boundaries[2].size != table.shape[2]:
-                mylog.error("Z bins array not the same length as the data.")
-                raise ValueError
+            self._validate_bin_boundaries(boundaries)
             self.x_bins = boundaries[0]
             self.y_bins = boundaries[1]
             self.z_bins = boundaries[2]
@@ -299,7 +296,7 @@ class QuadrilinearFieldInterpolator(_LinearInterpolator):
         field_data = interp(ad)
 
         """
-        super().__init__(table, truncate=truncate, store_table=store_table)
+        super().__init__(table, field_names, truncate=truncate, store_table=store_table)
         self.x_name, self.y_name, self.z_name, self.w_name = field_names
         if len(boundaries) == 8:
             x0, x1, y0, y1, z0, z1, w0, w1 = boundaries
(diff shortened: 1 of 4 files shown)


chrishavlin:
Note that the test for the 4d interpolator is kinda slow cause it uses quite a big array, (`random_data = np.random.random((64, 64, 64, 64))`). I'm be inclined to drop that down in size to speed it up, I can do that after an initial review (and after the tests pass as-is).

yt/utilities/linear_interpolators.py
```diff
@@ -1,11 +1,82 @@
...
 from yt.funcs import mylog
 
 
-class UnilinearFieldInterpolator:
-    def __init__(self, table, boundaries, field_names, truncate=False):
+class _LinearInterpolator(abc.ABC):
+    _ndim: int
+    _dim_i_type = "int32"
+
+    def __init__(self, table, field_names, truncate=False, *, store_table=True):
```
neutrinoceros, line 13: Just a quick remark (I'll do a much more in depth review when I can): I'd put the asterisk before any boolean argument to avoid Boolean traps.

chrishavlin: happy to move it -- but that'd make it not backwards compatible, right? any worry there?

chrishavlin: (and absolutely no rush on the review)

neutrinoceros: I didn't realize this was existing API. Indeed it'd be breaking to move the asterisk without a warning first. It's way out of scope for this PR, so ignore my nit !

matthewturk:
Hmm, how slow is it?  Is it a lot slower than doing a 3D interpolator on (256, 256, 256) data?

chrishavlin:
@matthewturk so the 3D and 4D scale at the same rate with the total array size:
lin_interp_scaling

code here https://gist.github.com/chrishavlin/9370e2a4a1895745a40c46f39d6c44f4 

The full test time (`pytest --durations=10 yt/utilities/tests/test_interpolators.py`):

```
3.20s call     yt/utilities/tests/test_interpolators.py::test_linear_interpolator_4d
0.31s call     yt/utilities/tests/test_interpolators.py::test_ghost_zone_extrapolation
0.25s call     yt/utilities/tests/test_interpolators.py::test_get_vertex_centered_data
0.05s setup    yt/utilities/tests/test_interpolators.py::test_linear_interpolator_1d
0.04s call     yt/utilities/tests/test_interpolators.py::test_linear_interpolator_3d
```
So only 3ish seconds, not as bad as I remember but 2 orders of magnitude longer than the other interpolator tests (and I don't see a particular reason to test with N=64 vs say N=16?).

matthewturk (dismissed):
I think as long as the tests pass, and the API is consistent/unbroken if not unchanged, this is good.

yt/utilities/tests/test_interpolators.py
```diff
@@ -152,3 +153,63 @@ def test_get_vertex_centered_data():
...
+
+_lin_interpolators_by_dim = {
+    1: lin.UnilinearFieldInterpolator,
+    2: lin.BilinearFieldInterpolator,
+    3: lin.TrilinearFieldInterpolator,
+    4: lin.QuadrilinearFieldInterpolator,
+}
+
+
+@pytest.mark.parametrize("ndim", list(range(1, 5)))
```
neutrinoceros, line 166: ?
```suggestion
@pytest.mark.parametrize("ndim", list(_lin_interpolators_by_dim.keys()))
```

yt/utilities/tests/test_interpolators.py
```diff
@@ -152,3 +153,63 @@ def test_get_vertex_centered_data():
...
+    # check that we can do it without storing the initial table
+    interpolator = interp_class(
+        random_data, boundaries, field_names, True, store_table=False
+    )
+    assert_array_almost_equal(interpolator(fv, table=table_2), table_2)
+
+    with pytest.raises(
+        ValueError, match="You must either store the table used when initializing"
+    ):
+        _ = interpolator(fv)
```
neutrinoceros, line 192:
```suggestion
        interpolator(fv)
```

yt/utilities/tests/test_interpolators.py
```diff
@@ -152,3 +153,63 @@ def test_get_vertex_centered_data():
...
+        bounds = bad_bounds
+    else:
+        bounds = [
+            good_bounds,
+        ] * ndim
+        bounds[0] = bad_bounds
+        bounds = tuple(bounds)
+
+    with pytest.raises(ValueError, match=f"{field_names[0]} bins array not"):
+        _ = interp_class(random_data, bounds, field_names)
```
neutrinoceros, line 215:
```suggestion
        interp_class(random_data, bounds, field_names)
```

yt/utilities/tests/test_interpolators.py
```diff
@@ -152,3 +153,63 @@ def test_get_vertex_centered_data():
...
+
+    random_data = np.random.random((sz,) * ndim)
+
+    field_names = "xyzw"[:ndim]
+    slc = slice(0.0, 1.0, complex(0, sz))
+    fv = dict(zip(field_names, np.mgrid[(slc,) * ndim]))
+    boundaries = (0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0)[: ndim * 2]
+
+    interp_class = _lin_interpolators_by_dim[ndim]
+    interpolator = interp_class(random_data, boundaries, field_names, True)
```
neutrinoceros, line 178: For clarity, please avoid passing boolean arguments positionally

yt/utilities/tests/test_interpolators.py
```diff
@@ -152,3 +153,63 @@ def test_get_vertex_centered_data():
...
+    assert_array_almost_equal(interpolator(fv), random_data)
+    table_2 = random_data * 2
+    assert_array_almost_equal(interpolator(fv, table=table_2), table_2)
+
+    # check that we can do it without storing the initial table
+    interpolator = interp_class(
+        random_data, boundaries, field_names, True, store_table=False
```
neutrinoceros, line 185: same as above

yt/utilities/tests/test_interpolators.py
```diff
@@ -152,3 +153,63 @@ def test_get_vertex_centered_data():
...
+    fv = dict(zip(field_names, np.mgrid[(slc,) * ndim]))
+    boundaries = (0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0)[: ndim * 2]
+
+    interp_class = _lin_interpolators_by_dim[ndim]
+    interpolator = interp_class(random_data, boundaries, field_names, True)
+    assert_array_almost_equal(interpolator(fv), random_data)
+    table_2 = random_data * 2
+    assert_array_almost_equal(interpolator(fv, table=table_2), table_2)
+
+    # check that we can do it without storing the initial table
```
neutrinoceros, line 183: This comment would be even more helpful if it also mentioned that `store_table=True` is implicit in the previous example as it's the default value.

yt/utilities/linear_interpolators.py
```diff
@@ -1,11 +1,82 @@
...
+        if store_table:
+            self.table = table.astype("float64")
```
neutrinoceros, line 15: this is an implicit data copy. Is it intended ?

yt/utilities/linear_interpolators.py
```diff
@@ -1,11 +1,82 @@
...
+        self.truncate = truncate
+        self._field_names = field_names
+
+    def _raise_truncation_error(self, data_object):
+        mylog.error(
+            "Sorry, but your values are outside "
+            "the table!  Dunno what to do, so dying."
+        )
+        mylog.error("Error was in: %s", data_object)
+        raise ValueError
```
neutrinoceros, line 28: I'm not sure I see the value of having log errors *and* an exception. In my opinion, the exception is where information should be given, making log errors redundant.

yt/utilities/linear_interpolators.py
```diff
@@ -1,11 +1,82 @@
...
+
+    def _validate_table(self, table_override):
+        if table_override is None:
+            if self.table is None:
+                msg = (
+                    f"You must either store the table used when initializing a new "
+                    f"{type(self).__name__} (set `store_table=True`) or you must provide a `table` when "
+                    f"calling {type(self).__name__}"
```
neutrinoceros, line 36:
```suggestion
                    f"{type(self).__name__} (set `store_table=True`) or you provide a `table` when "
                    f"calling {type(self).__name__}"
```

yt/utilities/linear_interpolators.py
```diff
@@ -21,53 +92,52 @@ def __init__(self, table, boundaries, field_names, truncate=False):
             If False, an exception is raised if the input values are
             outside the bounds of the table.  If True, extrapolation is
             performed.
+        store_table: bool
+            If False, only the shape of the input table is stored and
+            a full table must be provided when calling the interpolator.
```
neutrinoceros, line 97: The default value should be explicit. (This comment applies, of course, to everywhere this is copied)

yt/utilities/linear_interpolators.py
```diff
@@ -1,11 +1,82 @@
...
+                    f"calling {type(self).__name__}"
+                )
+                raise ValueError(msg)
+            return self.table
+
+        if table_override.shape != self.table_shape:
+            msg = f"The table_override shape, {table_override.shape}, must match the base table shape, {self.table_shape}"
+            raise ValueError(msg)
+
+        return table_override.astype("float64")
```
neutrinoceros, line 45: again, is the implicit copy intended ?

yt/utilities/linear_interpolators.py
```diff
@@ -1,11 +1,82 @@
...
+
+    def _get_digitized_arrays(self, data_object):
+        return_arrays = []
+        for dim in "xyzw"[: self._ndim]:
+            dim_name = getattr(self, f"{dim}_name")
+            dim_bins = getattr(self, f"{dim}_bins")
+
+            dim_vals = data_object[dim_name].ravel().astype("float64")
```
neutrinoceros, line 53: `.ravel().astype("float64")` copies data at least once, and up to two times. Inverting order of operations will make the number of copies exactly one in all cases.
```suggestion
            dim_vals = data_object[dim_name].astype("float64").ravel()
```

yt/utilities/linear_interpolators.py
```diff
@@ -1,11 +1,82 @@
...
 
 import yt.utilities.lib.interpolators as lib
 from yt.funcs import mylog
 
 
-class UnilinearFieldInterpolator:
-    def __init__(self, table, boundaries, field_names, truncate=False):
+class _LinearInterpolator(abc.ABC):
+    _ndim: int
+    _dim_i_type = "int32"
```
neutrinoceros, line 11: wow, I'm sure this is just a your refactor making this more obvious, but do you have any idea why we are *not* using the same dtype in all interpolators ?

neutrinoceros (commented):
Overall looks sound. I left a couple minor questions and suggestions.

yt/utilities/linear_interpolators.py
chrishavlin: I do not know :) I meant to look more closely, maybe there's some reason at the cython level? Or...?

yt/utilities/linear_interpolators.py
chrishavlin: it was intended in that I was following what the current implementation on main does (but I'm not sure it's necessary).

yt/utilities/linear_interpolators.py
chrishavlin: yup, I agree, I was again just copying exactly what was here before. In this case though I think it's clear that there's not reason not to change this to simply raise the error.

yt/utilities/linear_interpolators.py
chrishavlin: same reason as before -- I just hadn't dug into the reasons for copying the data in the original code, so decided to just follow what the old code was doing.

yt/utilities/tests/test_interpolators.py
chrishavlin: sure! that's a bit clearer.

yt/utilities/linear_interpolators.py
```diff
@@ -1,11 +1,82 @@
...
+
+        return return_arrays
+
+    def _validate_bin_boundaries(self, boundaries):
+        # boundaries: tuple of ndarrays
+        for idim in range(self._ndim):
+            if boundaries[idim].size != self.table_shape[idim]:
+                msg = f"{self._field_names[idim]} bins array not the same length as the data."
+                mylog.error(msg)
+                raise ValueError(msg)
```
chrishavlin, line 71: another spot that both logs and errors, i'll switch this as well to just error.

yt/utilities/linear_interpolators.py
```diff
@@ -78,75 +146,67 @@ def __init__(self, table, boundaries, field_names, truncate=False):
...
-            self.y_bins = boundaries[1]
+            self._validate_bin_boundaries(boundaries)
+            self.x_bins, self.y_bins = boundaries
         else:
-            mylog.error(
-                "Boundaries must be given as (x0, x1, y0, y1) or as (x_bins, y_bins)"
-            )
-            raise ValueError
+            msg = "Boundaries must be given as (x0, x1, y0, y1) or as (x_bins, y_bins)"
+            raise ValueError(msg)
```
chrishavlin, line 188: another spot where changed from log + error to only error

yt/utilities/tests/test_interpolators.py
chrishavlin: adjusted this and added some more comments

yt/utilities/tests/test_interpolators.py
chrishavlin: done

yt/utilities/tests/test_interpolators.py
chrishavlin: added more comments above

yt/utilities/linear_interpolators.py
chrishavlin: ok, changed this line as well as other spots throughout that were logging and raising errors.

yt/utilities/linear_interpolators.py
chrishavlin: re-wrote this to note the default

yt/utilities/linear_interpolators.py
chrishavlin: Ok, looked back and the superficial reason is that the cython functions for the different interpolators have different int declarations in their call signature: `UnilinearlyInterpolate` and `BilinearlyInterpolate` are declared as `int32_t`, e.g., : 

```
def BilinearlyInterpolate(np.ndarray[np.float64_t, ndim=2] table,
                          np.ndarray[np.float64_t, ndim=1] x_vals,
                          np.ndarray[np.float64_t, ndim=1] y_vals,
                          np.ndarray[np.float64_t, ndim=1] x_bins,
                          np.ndarray[np.float64_t, ndim=1] y_bins,
                          np.ndarray[np.int32_t, ndim=1] x_is,
                          np.ndarray[np.int32_t, ndim=1] y_is,
                          np.ndarray[np.float64_t, ndim=1] output):
```

while `TrilinearlyInterpolate` and `QuadrilinearlyInterpolate` use `int_t`:

```
def TrilinearlyInterpolate(np.ndarray[np.float64_t, ndim=3] table,
                           np.ndarray[np.float64_t, ndim=1] x_vals,
                           np.ndarray[np.float64_t, ndim=1] y_vals,
                           np.ndarray[np.float64_t, ndim=1] z_vals,
                           np.ndarray[np.float64_t, ndim=1] x_bins,
                           np.ndarray[np.float64_t, ndim=1] y_bins,
                           np.ndarray[np.float64_t, ndim=1] z_bins,
                           np.ndarray[np.int_t, ndim=1] x_is,
                           np.ndarray[np.int_t, ndim=1] y_is,
                           np.ndarray[np.int_t, ndim=1] z_is,
                           np.ndarray[np.float64_t, ndim=1] output):
```

As to **why** they differ, I'm really not sure... 

I could try switching `UnilinearlyInterpolate` and `BilinearlyInterpolate` over to use `int_t` as well?

yt/utilities/linear_interpolators.py
chrishavlin: Ok, looking through the code again, I don't see a reason why we'd need a copy.  `table_override.astype("float64", copy=False)` should be totally fine here.

chrishavlin:
oh, one more push coming... i'm going to go through all the other tests and add make the booleans explicit in the interpolator calls.

yt/utilities/linear_interpolators.py
neutrinoceros: > I could try switching UnilinearlyInterpolate and BilinearlyInterpolate over to use int_t as well?

Now might be a good time to try indeed, but it's not critical !

chrishavlin: Just switched all the interpolators to use `int_t`

chrishavlin:
so i was reminded of this PR while answering a question on slack related to `add_interpolated_field`... it'd be nice to get this in. I don't actually care too much about the new functionality I added here (the `store_table` business), but I'd like to save the refactoring work.

chrishavlin:
also, the rebase was kinda tricky, so i gotta double check that it went alright

chrishavlin:
pre-commit.ci autofix

chrishavlin:
ok, just double checked and the current state of this branch is good!

@matthewturk what aspect would have been useful? the new `store_table` functionality or just the refactoring ?

yt/utilities/linear_interpolators.py
chrishavlin: all interpolators on this branch are now using `int64_t`

PR #4712, opened 2023-10-25
[WIP] add depth-limiting to off-axis sph particle projections

### LINKED ISSUE #4711: depth keyword in off-axis projections with sph particles has no effect?
The `depth` keyword in `OffAxisProjectionPlot`s has no effect with sph particles. Reading through the code it looks to me like while the `depth` will get used to set the `bounds`, when projecting with sph particles only the x-y bounds of the rotated particle positions get used, so I'd guess that it has not been implemented fully but it'd be great to get confirmation from others who are more familiar with this part of the code before I attempt a fix.

## reproduction

The following returns identical images:

```python 
import yt 
import numpy as np

ds = yt.load_sample("gizmo_64")  

z_hat_original = ds.arr([0.3, 0.3, 0.3], 'kpc')
y_hat_original = ds.arr([1., 0., 0.], 'kpc')

for d in np.arange(0, 10, 1):
    c = ((center[0], center[1], center[2]), 'Mpc')
    prj = yt.ProjectionPlot(ds, - z_hat_original, 
                                       center = ds.domain_center, 
                                       fields=('gas', 'density'),                                 
                                       depth = (d, 'kpc'), 
                                       north_vector = - y_hat_original, 
                                       )
    prj.save(f"depth_{str(d).zfill(2)}_proj.png")
```
I'd expect the images to vary as the `depth` parameter is changed. 

## what to do

If I'm not missing something, seems to me there are 2 options: 

* short fix: raise a `NotImplemented` error when `depth` is used with particles
* better fix: implement it

## code notes 
By the time we get down to `off_axis_projection`, the bounds are still correct (accounting for the `depth` argument)

https://github.com/yt-project/yt/blob/9bd319d24ec3312ba8670efdc0cd5d5165f9d394/yt/visualization/volume_rendering/off_axis_projection.py#L211-L229

But when calculating the rotated bounds, only the rotated x-y are calculated and used in the kernel projection 

https://github.com/yt-project/yt/blob/9bd319d24ec3312ba8670efdc0cd5d5165f9d394/yt/utilities/lib/pixelization_routines.pyx#L1934-L1950

So I think we'd want `rotate_particle_coord` to also return the rotated z vals and the rotated z bounds and then have `pixelize_sph_kernel_projection` also discard particles outsize those rotated z bounds.. I'm happy to attempt that, but wanted to double check that I'm not going down the wrong path here...

chrishavlin:
Closes #4711 

The initial push does indeed work for sph particle projections, but I intend to refactor... 

that said, 

```python
import yt  
import numpy as np

ds = yt.load_sample("gizmo_64")  

z_hat_original = ds.arr([0.3, 0.3, 0.3], 'kpc')
y_hat_original = ds.arr([1., 0., 0.], 'kpc')

for d in [2, 20]: 
    prj = yt.ProjectionPlot(ds, - z_hat_original, 
                                       center = ds.domain_center, 
                                       fields=('gas', 'density'),                   
                                       depth = (d, 'Mpc'), 
                                       north_vector = - y_hat_original, 
                                       )
    prj.save(f"depth_{str(d).zfill(2)}_proj.png")
```
returns 
depth_02_proj

and 
depth_20_proj

which seems to make sense (first image is 2 Mpc depth, second is 20 Mpc depth).

Still to do:

- [ ] consider default behavior... i'm expecting answer tests to fail because the default `depth` in `OffAxisProjectionPlot` is `(1, '1')` instead of `None`, so it may drop some particles, changing what the plots with default values look like? **UPDATE**: didn't actually break tests, but that may be a function of test coverage and there could be a change in behavior introduced here? 
- [ ] make sure I'm not introducing confusing or diverging behavior between the particle and non-particle projections (which both use the same `OffAxisProjectionDummyDataSource`). I think it's fine, but want to double check... **UPDATE**: I think I'm coming back around on this ... now thinking that it'd be better to simply always apply the depth limitation, following how the off axis projection for gridded data works. Will marinate on this for a bit .... 
- [ ] refactor: in the initial push I created new versions of the pixelation routines based on the existing `pixelize_sph_kernel_projection` and `rotate_particle_coord` and introduced a ton of code duplication in the process. I mainly wanted to avoid creating extra arrays for the rotated z coordinates and running extra checks on z bounds in the case where it's not limiting by z, so started off by simply creating separate functions... but will go back and refactor this now that I have it nominally working.
- [x] add tests  
- [ ] add note in docs

### DIFF
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -1284,6 +1284,191 @@ def pixelize_sph_kernel_projection(
 
     return mask
 
+@cython.initializedcheck(False)
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def pixelize_sph_kernel_projection_with_depth_check(
+        np.float64_t[:, :] buff,
+        np.uint8_t[:, :] mask,
+        any_float[:] posx,
+        any_float[:] posy,
+        any_float[:] posz,
+        any_float[:] hsml,
+        any_float[:] pmass,
+        any_float[:] pdens,
+        any_float[:] quantity_to_smooth,
+        bounds,
+        kernel_name="cubic",
+        weight_field=None,
+        int check_period=1,
+        period=None):
+
+    cdef np.intp_t xsize, ysize
+    cdef np.float64_t x_min, x_max, y_min, y_max, z_min, z_max, prefactor_j
+    cdef np.int64_t xi, yi, x0, x1, y0, y1, xxi, yyi
+    cdef np.float64_t q_ij2, posx_diff, posy_diff, ih_j2
+    cdef np.float64_t x, y, dx, dy, idx, idy, h_j2, px, py, pz
+    cdef np.float64_t period_x = 0, period_y = 0, period_z = 0
+    cdef int i, j, ii, jj, kk
+    cdef np.float64_t[:] _weight_field
+    cdef int * xiter
+    cdef int * yiter
+    cdef int * ziter
+    cdef np.float64_t * xiterv
+    cdef np.float64_t * yiterv
+    cdef np.float64_t * ziterv
+
+    if weight_field is not None:
+        _weight_field = weight_field
+
+    if period is not None:
+        period_x = period[0]
+        period_y = period[1]
+        period_z = period[2]
+
+    # we find the x and y range over which we have pixels and we find how many
+    # pixels we have in each dimension
+    xsize, ysize = buff.shape[0], buff.shape[1]
+    x_min = bounds[0]
+    x_max = bounds[1]
+    y_min = bounds[2]
+    y_max = bounds[3]
+
+    # we also want to skip particles outside the rotated z bounds
+    z_min = bounds[4]
+    z_max = bounds[5]
+
+    dx = (x_max - x_min) / xsize
+    dy = (y_max - y_min) / ysize
+
+    idx = 1.0/dx
+    idy = 1.0/dy
+
+    if kernel_name not in kernel_tables:
+        kernel_tables[kernel_name] = SPHKernelInterpolationTable(kernel_name)
+    cdef SPHKernelInterpolationTable itab = kernel_tables[kernel_name]
+
+    with nogil, parallel():
+        # loop through every particle
+        # NOTE: this loop can be quite time consuming. see description in
+        # pixelize_sph_kernel_projection.
+
+        local_buff = <np.float64_t *> malloc(sizeof(np.float64_t) * xsize * ysize)
+        xiterv = <np.float64_t *> malloc(sizeof(np.float64_t) * 2)
+        yiterv = <np.float64_t *> malloc(sizeof(np.float64_t) * 2)
+        ziterv = <np.float64_t *> malloc(sizeof(np.float64_t) * 2)
+        xiter = <int *> malloc(sizeof(int) * 2)
+        yiter = <int *> malloc(sizeof(int) * 2)
+        ziter = <int *> malloc(sizeof(int) * 2)
+        xiter[0] = yiter[0] = ziter[0] = 0
+        xiterv[0] = yiterv[0] = ziterv[0] = 0.0
+        for i in range(xsize * ysize):
+            local_buff[i] = 0.0
+
+        for j in prange(0, posx.shape[0], schedule="dynamic"):
+            if j % 100000 == 0:
+                with gil:
+                    PyErr_CheckSignals()
+
+            xiter[1] = yiter[1] = ziter[1] = 999
+
+            if check_period == 1:
+                if posx[j] - hsml[j] < x_min:
+                    xiter[1] = +1
+                    xiterv[1] = period_x
+                elif posx[j] + hsml[j] > x_max:
+                    xiter[1] = -1
+                    xiterv[1] = -period_x
+                if posy[j] - hsml[j] < y_min:
+                    yiter[1] = +1
+                    yiterv[1] = period_y
+                elif posy[j] + hsml[j] > y_max:
+                    yiter[1] = -1
+                    yiterv[1] = -period_y
+                if posz[j] - hsml[j] < z_min:
+                    ziter[1] = +1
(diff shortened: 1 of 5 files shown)


yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1284,6 +1284,191 @@ def pixelize_sph_kernel_projection(
...
+    cdef int i, j, ii, jj, kk
+    cdef np.float64_t[:] _weight_field
+    cdef int * xiter
+    cdef int * yiter
+    cdef int * ziter
+    cdef np.float64_t * xiterv
+    cdef np.float64_t * yiterv
+    cdef np.float64_t * ziterv
+
+    if weight_field is not None:
```
matthewturk, line 1322: I think you got this code from elsewhere, but I will say that it definitely confused me.  I at first thought this was a string, but now I see it's either a numpy array or none.  Yowza!

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1284,6 +1284,191 @@ def pixelize_sph_kernel_projection(
...
+        kernel_tables[kernel_name] = SPHKernelInterpolationTable(kernel_name)
+    cdef SPHKernelInterpolationTable itab = kernel_tables[kernel_name]
+
+    with nogil, parallel():
+        # loop through every particle
+        # NOTE: this loop can be quite time consuming. see description in
+        # pixelize_sph_kernel_projection.
+
+        local_buff = <np.float64_t *> malloc(sizeof(np.float64_t) * xsize * ysize)
+        xiterv = <np.float64_t *> malloc(sizeof(np.float64_t) * 2)
```
matthewturk, line 1358: I have a sneaking suspicion it's to avoid a race condition, but I think a long time ago we were supposed to be able to declare variables inside a `parallel` block to get them to be thread-local.  No changes needed.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1284,6 +1284,191 @@ def pixelize_sph_kernel_projection(
...
+                                # now we just use the kernel projection
+                                local_buff[xi + yi*xsize] +=  prefactor_j * itab.interpolate(q_ij2)
+                                mask[xi, yi] = 1
+
+        with gil:
+            for xxi in range(xsize):
+                for yyi in range(ysize):
+                    buff[xxi, yyi] += local_buff[xxi + yyi*xsize]
+        free(local_buff)
+        free(xiterv)
```
matthewturk, line 1463: does this also need to happen inside the parallel context?

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1910,6 +2095,61 @@ def rotate_particle_coord(np.float64_t[:] px,
     return px_rotated, py_rotated, rot_bounds_x0, rot_bounds_x1, rot_bounds_y0, rot_bounds_y1
 
 
+@cython.boundscheck(False)
```
matthewturk, line 2098: So I think this function looks right.  What I'm wondering is if it needs to be in Cython as-is.  I think it could potentially take advantage of being in cython by consolidating the `matmul` stuff with the copying of coordinates -- i.e., iterating over the particles one-by-one and avoiding temporary arrays.  I'm not 100% sure that it's *necessary*, but it could potentially eliminate some allocations.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1926,28 +2166,53 @@ def off_axis_projection_SPH(np.float64_t[:] px,
...
-                                   smoothing_lengths,
-                                   particle_masses,
-                                   particle_densities,
-                                   quantity_to_smooth,
-                                   [rot_bounds_x0, rot_bounds_x1,
-                                    rot_bounds_y0, rot_bounds_y1],
-                                   weight_field=weight_field,
-                                   check_period=0)
+    if depth_set:
+        px_rotated, py_rotated, pz_rotated, \
```
matthewturk, line 2176: We shouldn't fix this here, but is this a case where something like a named tuple or simple data class could simplify our code?

matthewturk (commented):
I like this.  Thank you!  I've left some comments.

yt/utilities/lib/pixelization_routines.pyx
chrishavlin: indeed! copied directly from `pixelize_sph_kernel_projection`. Up in the python level there's a `weight` that is a string, so definitely confusing. Could rename this variable to `weight_array` (and also rename in `pixelize_sph_kernel_projection`).

yt/utilities/lib/pixelization_routines.pyx
chrishavlin: helpful context! I was confused by why this was needed.

yt/utilities/lib/pixelization_routines.pyx
chrishavlin: hard to tell with the github interface, but these lines are all within the 

```
with nogil, parallel():
```

block way up top. 

Screenshot 2023-11-13 at 11 09 45 AM

yt/utilities/lib/pixelization_routines.pyx
chrishavlin: ya, I agree -- it could be written with just numpy matrix operations at the python level or rewritten with a different loop control. 

Also, could actually cut out an entire loop over the particles by calculating the rotated particle positions as needed within `pixelize_sph_kernel_projection_with_depth_check` (and `pixelize_sph_kernel_projection`)?

yt/utilities/lib/pixelization_routines.pyx
chrishavlin: yes, I do think that would help. it might be worth fixing here too :) I think it could cut down on some of the code duplication I'm introducing.

nastasha-w:
I ended up putting in depth limits in SPH projections in #4939 -- too bad it was duplicate work.

chrishavlin:
> I ended up putting in depth limits in SPH projections in #4939 -- too bad it was duplicate work.

great!! I'm just glad the functionality got in :)  I'll close this one out.

PR #4717, opened 2023-10-27
Find output by its attribute

cphyc:
## PR Checklist


neutrinoceros:
Quick input:
- I think this is a great idea
- I'd prefer we avoid making it overly flexible: supporting the ability to search via arbitrary attributes may create unforeseen edge cases which will be impossible to test systematically. I think supporting search by time or redshift is sufficient for now, especially if it can be done in a way that doesn't involve typing the exact attribute names (IMO `current_time` is somewhat badly named, because its meaning excessively context-dependent)

cphyc:
@neutrinoceros could you point me (again) to the doc explaining how to deactivate a test for nose?

matthewturk:
Not sure I completely agree with @neutrinoceros about limiting the flexibility.  The question the other day about flash datasets with particles would suggest there may be times when we don't know the attribute someone will be looking for.

Can you make it use "Price Is Right" rules?  i.e., closest-without-going-over?

cphyc:
I am not sure I understand what you mean by "Price Is Right" rule. Anyways, in its current state, it is already possible to query a dataset by attribute, but the method is hidden (`_get_by_attribute`), because “with great power comes great responsibility”.

Let me know what you think! I can always strip the leading `_` to make it more user-facing, with an obvious comment in the docstring warning against misuses?

matthewturk:
I mean, if I want the closet to z=1 but definitely on one side or the
other, for instance. So having 0.99 take precedence over 1.01. (or even
0.95 over 1.001)

cphyc:
Oh I see, yes, I'll implement a `side` argument. To either get the output on the left or on the right!

neutrinoceros:
I have nothing against that "the price is right rule", but I'd like to point out that left-to-right reading order isn't culturally neutral. I'd prefer that the interface be more inclusive.

cphyc:
@neutrinoceros I've implemented a "smaller", "larger" or "nearest" switch so that there is no ambiguity arising from left-to-right vs. right-to-left reading order. Thanks for pointing it out!

matthewturk:
I think the new language is better. That being said, it's not necessarily true that RTL languages have different number line concepts:

https://math.stackexchange.com/questions/2618537/do-right-to-left-readers-also-reverse-mathematical-concepts-like-number-line

There's also some evidence that filmmaking conventions (such as progression from the left side of the screen to the right) are also commonly used outside of left-to-right languages.

I will say that with redshift the entire situation is odd because the numeric progression is reversed from temporal progression. And I did choose redshift as my example... So I guess this is on me!

cphyc:
@yt-fido test this please

doc/source/analyzing/time_series_analysis.rst
```diff
@@ -77,6 +77,29 @@ see:
...
+
+   import yt
+
+   ts = yt.load("*/*.index")
+   # Get output at 3 Gyr
+   ds = ts.get_by_time((3, "Gyr"))
+   # This will fail if no output is found within 100 Myr
+   ds = ts.get_by_time((3, "Gyr"), tolerance=(100, "Myr"))
+   # Get the output at the time right before and after 3 Gyr
+   ds_before = ts.get_by_time((3, "Gyr"), side="smaller")
```
neutrinoceros, line 93: I still think the name of the argument could be clearer than `side`. What about `prefer="smaller"` ?

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+        if tolerance is not None:
+            if abs(value - getattr(ds_best, attribute)) > tolerance:
+                raise ValueError(
+                    f"No dataset found with {attribute} within {tolerance} of {value}."
+                )
+        return ds_best
+
+    def get_by_time(
+        self,
+        time: Union[unyt_quantity, tuple],
```
neutrinoceros, line 552:
```suggestion
        time: Union[unyt_quantity, tuple[float, Union[Unit, str]]],
```

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+        tolerance: Union[None, unyt_quantity, tuple] = None,
```
neutrinoceros, line 553:
```suggestion
        tolerance: Union[None, unyt_quantity, tuple[float, Union[Unit, str]]] = None,
```

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+        side: Union[
+            Literal["nearest"], Literal["smaller"], Literal["larger"]
+        ] = "nearest",
```
neutrinoceros, line 556: Unions of `Literal` are actually simpler to express as

```suggestion
        side: Literal["nearest", "smaller", "larger"] = "nearest",
```

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+    ):
```
neutrinoceros, line 557:
```suggestion
    ) -> Dataset:
```

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+            "current_time", time, tolerance=tolerance, side=side
+        )
+
+    def get_by_redshift(
+        self,
+        redshift: float,
+        tolerance: Optional[float] = None,
+        side: Union[
+            Literal["nearest"], Literal["smaller"], Literal["larger"]
+        ] = "nearest",
```
neutrinoceros, line 590:
```suggestion
        side: Literal["nearest", "smaller", "larger"] = "nearest",
```

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+    ):
```
neutrinoceros, line 591:
```suggestion
    ) -> Dataset:
```

yt/data_objects/tests/test_time_series.py
```diff
@@ -34,71 +35,115 @@ def test_no_match_pattern():
         )
 
 
-def test_init_fake_dataseries():
-    file_list = [f"fake_data_file_{str(i).zfill(4)}" for i in range(10)]
+@pytest.fixture
+def FakeDataset():
+    class __FakeDataset(Dataset):
```
neutrinoceros, line 40: In Python, it's preferable to avoid naming anything with leading double underscores (and without trailing double underscores), because of [how name mangling works](https://docs.python.org/3/tutorial/classes.html#private-variables)
```suggestion
    class _FakeDataset(Dataset):
```

yt/data_objects/tests/test_time_series.py
```diff
@@ -34,71 +35,115 @@ def test_no_match_pattern():
...
+            return
+
+        def _hash(self):
+            return
+
+        def _setup_classes(self):
+            return
+
+    try:
+        yield __FakeDataset
```
neutrinoceros, line 69:
```suggestion
        yield _FakeDataset
```

yt/data_objects/tests/test_time_series.py
```diff
@@ -34,71 +35,115 @@ def test_no_match_pattern():
...
+    finally:
+        output_type_registry.pop("__FakeDataset")
```
neutrinoceros, line 71:
```suggestion
        output_type_registry.pop("_FakeDataset")
```

yt/data_objects/tests/test_time_series.py
```diff
@@ -34,71 +35,115 @@ def test_no_match_pattern():
...
-        ts = DatasetSeries(sfile_list)
-        assert ts._pre_outputs == sfile_list
+    # rejected input type (str repr of a list) "[file1, file2, ...]"
+    assert_raises(FileNotFoundError, DatasetSeries, str(file_list))
 
-        # init form Path list
-        ts = DatasetSeries(pfile_list)
-        assert ts._pre_outputs == pfile_list
+    # finally, check that ts[0] fails to actually load
+    assert_raises(YTUnidentifiedDataType, ts.__getitem__, 0)
```
neutrinoceros, line 111: now that this file is pytest-only, might as well use `pytest.raises` instead

yt/data_objects/time_series.py
```diff
@@ -146,7 +151,7 @@ class DatasetSeries:
     # this annotation should really be Optional[Type[Dataset]]
     # but we cannot import the yt.data_objects.static_output.Dataset
     # class here without creating a circular import for now
-    _dataset_cls: Optional[type] = None
+    _dataset_cls: Optional[type["Dataset"]] = None
```
neutrinoceros, line 154: good call, now we can also remove the 3-line comment for this annotation

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
             self, indices, fields=fields, suppress_logging=suppress_logging, ptype=ptype
         )
 
+    def _get_by_attribute(
+        self,
+        attribute: str,
+        value: Union[unyt_quantity, tuple[float, str]],
```
neutrinoceros, line 446:
```suggestion
        value: Union[unyt_quantity, tuple[float, Union[Unit, str]]],
```

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+        tolerance: Union[None, unyt_quantity, tuple[float, str]] = None,
```
neutrinoceros, line 447:
```suggestion
        tolerance: Union[None, unyt_quantity, tuple[float, Union[Unit, str]]] = None,
```

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+        side: Union[
+            Literal["nearest"], Literal["smaller"], Literal["larger"]
+        ] = "nearest",
```
neutrinoceros, line 450:
```suggestion
        side: Literal["nearest", "smaller", "larger"] = "nearest",
```

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+    ) -> "Dataset":
+        r"""
+        Get a dataset at or near to a given value.
+
+        Parameters
+        ----------
+        key : str
```
neutrinoceros, line 457: this is out of sync with the argument nam

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+        """
+
+        if side not in ("nearest", "smaller", "larger"):
+            raise ValueError(
+                f"side must be 'nearest', 'smaller' or 'larger', got {side}"
+            )
+
+        # Use a binary search to find the closest value
+        iL = 0
+        iH = len(self._pre_outputs) - 1
```
neutrinoceros, line 480: I don't understand what `H` stands for here. I would expect `iL` and `iR` (left/right) be used instead, can you explain ?

yt/data_objects/time_series.py
```diff
@@ -435,6 +440,179 @@ def particle_trajectories(
...
+        vH = getattr(dsH, attribute)
+
+        if vL < vH:
+            sign = 1
+        elif vL > vH:
+            sign = -1
+        else:
+            raise ValueError(
+                f"{dsL} and {dsH} have both {attribute}={vL}, cannot perform search. "
+                "Try with another key."
```
neutrinoceros, line 506: This seems to me like a poor suggestion, I'd rather have none at all.
```suggestion
```

yt/data_objects/time_series.py
```diff
@@ -2,13 +2,15 @@
 import glob
 import inspect
 import os
+import typing
```
neutrinoceros, line 5:
```suggestion
from typing import TYPE_CHECKING
```

yt/data_objects/time_series.py
```diff
@@ -28,6 +30,9 @@
     parallel_root_only,
 )
 
+if typing.TYPE_CHECKING:
```
neutrinoceros, line 33:
```suggestion
if TYPE_CHECKING:
```

neutrinoceros (commented):
The functionality is useful and the implementation is sound ! I have a couple questions and suggestions (mostly about type annotations)

cphyc:
@yt-fido test this please

cphyc:
@yt-fido test this please

yt/data_objects/time_series.py
```diff
@@ -435,6 +436,172 @@ def particle_trajectories(
...
+                iR = iM
+                dsR = dsM
+            elif sign * value > sign * vM:
+                iL = iM
+                dsL = dsM
+            else:  # Exact match
+                dsL = dsR = dsM
+                break
+
+        if prefer == "smaller":
```
chrishavlin, line 525: So I'm not sure if this needs to happen up in the `while` loop or down here... but I think you need to explicitly handle the case where the chosen value is exactly halfway between two timesteps, particularly when a value for `prefer` has been provided. As this is, the larger value will always be returned when the provided value is halfway, regardless of `prefer` (see my suggested change in `test_time_series.py`). If you prefer the current behavior then I think the behavior should be documented in the docstring and maybe even a log message (in which case you can ignore my suggested change to the test).

chrishavlin (commented):
Looks great! 

I think the case of choosing a value exactly between two datasets needs to either explicitly handled or have its behavior defined in the docstring (see comments), but otherwise this is very useful!

neutrinoceros:
pre-commit.ci autofix

cphyc:
@yt-fido test this please

neutrinoceros:
@yt-fido Test this please

neutrinoceros:
@yt-fido Test this please

neutrinoceros:
@yt-fido Test this please

chrishavlin:
Hey @cphyc , you still hoping for this to go into 4.4? 

I'd still like some clarification on expected behavior when you choose a time exactly between two points while setting the `prefer` keyword (see https://github.com/yt-project/yt/pull/4717#discussion_r1408007888 ), but otherwise I think everything has been addressed?

cphyc:
@chrishavlin should be good now :) I am now testing the behaviour when exactly halfway which is working as expected as far as I could tell.

chrishavlin:
great! thanks, @cphyc !

chrishavlin:
@neutrinoceros I think all your comments have been addressed, mind giving this a quick re-review when you have a moment?

neutrinoceros (approved):
Thanks for the ping ! Indeed my comments were all addressed !

neutrinoceros:
Last thing before we merge this: @cphyc, can you re-title the PR to something a bit more descriptive that I can include directly in release notes ?

cphyc:
Done!

PR #4720, opened 2023-10-31
Ramses-speedup

cphyc:
## PR Summary

RAMSES files require lots of random access in the file. In particular, when you have a lot of fields in the outputs, the performance drastically decreases if you're skipping most of them. In this PR,  I precompute how many "jumps" one needs to do ahead of time, and then do all jumps at once.

The code should be functionally equivalent but results in fewer reads.

### Performances

| slice | Timing on `main` [s] | Timing with this PR [s]|
|-------|----------------------|------------------------|
| all   | 52                   | 23|
| ::2   | 43 | 17 |
| ::4   | 39 | 12 |
| :4    | 35 | 9 |

To estimate the timings, the following code is being run, with the slice as reported in the table above. This is run on [Dial 3](https://dial3-docs.dirac.ac.uk/).

```python
import yt

# This is a (very) large simulation - 500Gib/output, 110 fields on-disk
p = "/lustre/dirac3/home/dc-cadi1/dp265/dc-katz1/MEGATRON/PRODUCTION_CP/output_00048"

# Only load the innermost part of the simulation (sufficient for benchmarking)
bbox = [[0.499]*3, [0.501]*3]

ds = yt.load(p, bbox=bbox)
ad = ds.all_data()

# Retain only the hydro fields + level
fields = [
    ("index", "grid_level"),
    *((ft, fn) for (ft, fn) in ds.field_list if ft == "ramses")
]

from time import time
before = time()
ad.get_data(fields[<slice>])
after = time()
print(f"Took {after-before:.2s}s")
```

### DIFF
--- a/yt/frontends/ramses/io_utils.pyx
+++ b/yt/frontends/ramses/io_utils.pyx
@@ -14,6 +14,10 @@ ctypedef np.int32_t INT32_t
 ctypedef np.int64_t INT64_t
 ctypedef np.float64_t DOUBLE_t
 
+cdef INT32_SIZE = sizeof(np.int32_t)
+cdef INT64_SIZE = sizeof(np.int64_t)
+cdef DOUBLE_SIZE = sizeof(np.float64_t)
+
 @cython.cpow(True)
 @cython.boundscheck(False)
 @cython.wraparound(False)
@@ -134,6 +138,9 @@ cpdef read_offset(FortranFile f, INT64_t min_level, INT64_t domain_id, INT64_t n
 
     return offset, level_count
 
+cdef inline int skip_len(int Nskip, int record_len):
+    return Nskip * (record_len * DOUBLE_SIZE + INT64_SIZE)
+
 @cython.cpow(True)
 @cython.boundscheck(False)
 @cython.wraparound(False)
@@ -154,16 +161,27 @@ def fill_hydro(FortranFile f,
     cdef dict tmp
     cdef str field
     cdef INT64_t twotondim
-    cdef int ilevel, icpu, ifield, nfields, nlevels, nc, ncpu_selected
-    cdef np.ndarray[np.uint8_t, ndim=1] mask
+    cdef int ilevel, icpu, nlevels, nc, ncpu_selected, nfields_selected
 
     twotondim = 2**ndim
-    nfields = len(all_fields)
+    nfields_selected = len(fields)
 
     nlevels = offsets.shape[1]
     ncpu_selected = len(cpu_enumerator)
 
-    mask = np.array([(field in fields) for field in all_fields], dtype=np.uint8)
+    cdef np.int64_t[:] jumps = np.zeros(nfields_selected + 1, dtype=np.int64)
+    cdef int jump_len
+
+    jump_len = 0
+    j = 0
+    for i, field in enumerate(all_fields):
+        if field in fields:
+            jumps[j] = jump_len
+            j += 1
+            jump_len = 0
+        else:
+            jump_len += 1
+    jumps[j] = jump_len
 
     # Loop over levels
     for ilevel in range(nlevels):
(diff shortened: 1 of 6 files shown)


cphyc:
@yt-fido test this please.

yt/frontends/ramses/io_utils.pyx
```diff
@@ -154,41 +162,74 @@ def fill_hydro(FortranFile f,
...
+                        f.seek(skip_len(jump_len, nc), 1)
+                        jump_len = 0
+                    f.read_vector_inplace('d', <void*> &buffer[0, i, j])
+
+                jump_len += jumps[nfields_selected]
+
+            # In principle, we may be left with some fields to skip
+            # but since we're doing an absolute seek above,
+            # we don't need to do anything here
+            ## if jump_len > 0:
```
matthewturk, line 225: do we need to seek to 0?  Or can we assume that gets handled at a later time, or when the file is closed/opened?

yt/frontends/ramses/io_utils.pyx
```diff
@@ -154,41 +162,74 @@ def fill_hydro(FortranFile f,
...
+            ##     f.seek(skip_len(jump_len, nc), 1)
+
+            # Alias buffer into dictionary
+            tmp = {}
+            for i, field in enumerate(fields):
+                tmp[field] = buffer[:, :, i]
```
matthewturk, line 231: This is a copy isn't it?  Because it's unrolling?

matthewturk:
Nice!  I think it might be interesting at a future time to explore the impact of using a memory map, to see if that can farm out the skipping/paging/seeking to the OS.

yt/frontends/ramses/io_utils.pyx
```diff
@@ -83,12 +87,16 @@ def read_amr(FortranFile f, dict headers,
 
     return max_level
 
+
+cdef inline int skip_len(int Nskip, int record_len) noexcept:
```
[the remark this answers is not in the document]
cphyc: Mmmh. The thing is all the code around has the GIL, so I'm afraid adding `nogil` would trigger releasing/retaking the GIL whenever we call the function, and make things slower as a consequence. But I might be misunderstanding the mechanism behind it.

yt/frontends/ramses/io_utils.pyx
cphyc: Maybe this isn't clear, but at the next iteration we do an absolute seek (https://github.com/cphyc/yt/blob/57fef2fd35af61d1738ad2458c87502f4150ee21/yt/frontends/ramses/io_utils.pyx#L203). So we can skip this seek. I have updated the comment to make this clearer (hopefully!).

yt/frontends/ramses/io_utils.pyx
cphyc: It should not be a copy! The whole business is to avoid copying data over and over again.
image

yt/frontends/ramses/io_utils.pyx
```diff
@@ -154,41 +162,74 @@ def fill_hydro(FortranFile f,
...
                 continue
             f.seek(offset)
-            tmp = {}
             # Initialize temporary data container for io
             # note: we use Fortran ordering to reflect the in-file ordering
-            for field in all_fields:
-                tmp[field] = np.empty((nc, twotondim), dtype="float64", order='F')
+            if nc > nc_largest:
+                nc_largest = nc
+                buffer = np.empty((nc, twotondim, nfields_selected), dtype="float64", order='F')
```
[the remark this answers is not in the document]
cphyc: For RAMSES, all fields have rigorously the same shape (and also are written in double precision), so this shouldn't matter. It would show in the answer testing anyway

yt/frontends/ramses/io_utils.pyx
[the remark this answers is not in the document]
cphyc: Just checked, and indeed, the code produced is the same. So I went ahead and added the `nogil`, just to be safe :)

cphyc:
@yt-fido test this please

cphyc:
@yt-fido test this please

yt/frontends/ramses/io_utils.pyx
matthewturk: Oh!  Thanks.

yt/frontends/ramses/io_utils.pyx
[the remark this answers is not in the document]
matthewturk: :+1:

matthewturk (dismissed):

matthewturk:
Perhaps this is a silly question, but have we evaluated the speed costs of computing the (absolute) offsets ahead-of-time for all fields, then doing absolute seeks sorted by the position in the file?  Rather than doing the cumulative seek-by-field-size?

cphyc:
At the moment, the file is already being read in order with a combination of absolute (https://github.com/yt-project/yt/pull/4720/files#diff-74cc1bfe029e80ddd26aba7f83f08aae2cd98986ec28ae0e745be83ce0658f64R203) and relative (https://github.com/yt-project/yt/pull/4720/files#diff-74cc1bfe029e80ddd26aba7f83f08aae2cd98986ec28ae0e745be83ce0658f64R216) seeks, but all of them should follow one another. Is this what you were referring to?

matthewturk:
I know we already sort by that -- I just meant to eliminate *some* of the overhead in computing the field skips etc by precomputing the exact position offsets and seeking directly to them.

cphyc:
Oh, I see. Let me illustrate to know if I understood your point correctly:
```
- iterate over levels
  - iterate over CPUs
	  1.   do an absolute seek
	  2.   do a relative seek to the first field to be read
	  3.   eventually read the ones immediately after
	  4.   do a relative seek to the next batch of fields to be read
	  5.   eventually read the ones immediately after
	  6.   [...]
 	  n.   do a relative seek to the last batch of fields to be read
	  n+1. eventually, read the ones immediately after
```

Following your comment, I think I optimized the hell out of the reader by putting together the first absolute seek + first relative seek (1 + 2).
With 7a953bcbe, I'm now limiting the number of jumps to its minimum. We _could_ compute ahead of time the value of the `skip_len`, but essentially each entry will have a unique value, so we would just be displacing the computation from being inline to being out of the loop (with small extra memory footprint). Does this make sense?

matthewturk:
Not quite, and I also don't want to side-track you too much.  My idea was simply to compute the absolute of each field, but I am now realizing that one big speedup we'd miss would be reading stride 1 data of size greater than a single field, whereas with my question it would require reading a single field at once (even if the seeks were not minimized).  So forget it -- it's pointless!

cphyc:
@yt-fido test this please

matthewturk (approved):

PR #4724, opened 2023-11-01
DEP: declare compatibility with unyt 3.0

### LINKED ISSUE #4162: DEP: upcoming incompatibilities with unyt 3.0

neutrinoceros:
## PR Summary

closes #4162 now that unyt 3.0 is available on PyPI


neutrinoceros:
Looks like we're still having a couple incompatibilities, I'll work out the ones that can be addressed here and solve the rest of them upstream, if any.

neutrinoceros:
@yt-fido Test this please

neutrinoceros:
Okay I got one successful run using a patched version of unyt 3.0.0. Here are all upstream patches
- https://github.com/yt-project/unyt/pull/463
- https://github.com/yt-project/unyt/pull/464
- https://github.com/yt-project/unyt/pull/466

I think I'll yank 3.0.0 from PyPI if @jzuhone agrees, but 3.0.1 will be compatible for sure

neutrinoceros:
@yt-fido test this please

neutrinoceros:
Ok, all set !

jzuhone (approved):

PR #4725, opened 2023-11-02
BUG: handle deprecation warnings from numpy 2.0.0dev0 (np.row_stack -> np.vstack and np.in1d -> np.isin)

neutrinoceros:
## PR Summary

Fix remaining usage of deprecated API. See https://numpy.org/devdocs/release/2.0.0-notes.html#deprecations

### DIFF
--- a/yt/data_objects/particle_trajectories.py
+++ b/yt/data_objects/particle_trajectories.py
@@ -99,9 +99,9 @@ class ParticleTrajectories:
         ):
             dd = ds.all_data()
             newtags = dd[fds["particle_index"]].d.astype("int64")
-            mask = np.in1d(newtags, indices, assume_unique=True)
+            mask = np.isin(newtags, indices, assume_unique=True)
             sort = np.argsort(newtags[mask])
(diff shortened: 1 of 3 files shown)


yt/data_objects/particle_trajectories.py
```diff
@@ -99,9 +99,9 @@ def __init__(
         ):
             dd = ds.all_data()
             newtags = dd[fds["particle_index"]].d.astype("int64")
-            mask = np.in1d(newtags, indices, assume_unique=True)
+            mask = np.isin(newtags, indices, assume_unique=True)
```
matthewturk, line 102: is the returned shape the same?

neutrinoceros: yes. It matches the shape of the first argument in both cases.

matthewturk (approved):
why is this draft?  looks good to go

neutrinoceros:
It is ! I removed the draft status as soon as I saw that tests passed everywhere :)

PR #4728, opened 2023-11-03
MNT: fix auto-reports for bleeding edge CI

### LINKED ISSUE #4721: MNT: automatic issues for nightly failures are broken
### Bug report

**Bug summary**

The change introduced in #4560 is apparently broken, see https://github.com/yt-project/yt/actions/runs/6715141128/job/18249749179

I'm going to prioritise fixing the failing tests over this, but I'll circle back to this infrastructure issue when I'm done.

neutrinoceros:
Goal: fix #4721
I suspect that the `labels` parameter is malformed (this is the only one I'm using that's not used in the matplotlib model workflow, and it's not well documented)

I'm going to test this by removing the conditional trigger.

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -87,7 +87,6 @@ jobs:
         body: |
           The weekly build with future dependencies has failed. Check the logs
           https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}
-        labels: 'tests: running tests,'
         pinned: false
         close-previous: false
       env:


neutrinoceros:
So it turns out testing this properly is a more involved process than I thought because of how GitHub actions and authentification integrate with each other. In short, GitHub actions runners authenticate as their own user, not as the account owning a pull request *and* if the branch is originated from a fork, they only get read permissions to the main repo (which makes sense). This means I needed to test this on my fork instead of here (or to create the branch on the main repo).
Doing so, I was able to confirm that the `labels` argument was the culprit, and that removing it fixed the pipeline.

neutrinoceros:
@yt-fido test this please

chrishavlin (approved):

neutrinoceros:
Let's merge this to give tonight's run a chance

PR #4730, opened 2023-11-03
Allow parallel I/O for hydro and particle reading

cphyc:
## PR Summary

This allows parallel I/O for the RAMSES dataset. Reading a RAMSES dataset has three steps:

- read in the AMR structure
- read the hydro files
- read the particle files

At the moment, only the latter two are parallelized.


yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -1157,9 +1199,9 @@ def alltoallv_array(self, send, total_size, offsets, sizes):
         dtr = send.dtype.itemsize / tmp_send.dtype.itemsize  # > 1
         roff = [off * dtr for off in offsets]
         rsize = [siz * dtr for siz in sizes]
-        tmp_recv = recv.view(self.__tocast)
+        tmp_recv = recv.view(transport_dtype)
```
cphyc, line 1202: The motivation for getting rid of the cast to CHAR is so that we do not hit the limit imposed by MPI of 2**31 elements when communicating arrays as quickly.

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -1114,13 +1150,17 @@ def send_array(self, arr, dest, tag=0):
...
         else:
             unit_metadata = ()
-        self.comm.send((arr.dtype.str, arr.shape) + unit_metadata, dest=dest, tag=tag)
-        self.comm.Send([arr, MPI.CHAR], dest=dest, tag=tag)
+        self.comm.send(
+            (arr.dtype.str, arr.shape, transport_dtype) + unit_metadata,
+            dest=dest,
+            tag=tag,
+        )
+        self.comm.Send([arr, mpi_type], dest=dest, tag=tag)
```
cphyc, line 1158: Isn't there a bug here? I think this should be
```suggestion
        self.comm.send(
            (arr.dtype.str, arr.shape, transport_dtype) + unit_metadata,
            dest=dest,
            tag=tag,
        )
        self.comm.Send([tmp, mpi_type], dest=dest, tag=tag)
```

cphyc:
@yt-fido test this please

yt/utilities/parallel_tools/parallel_analysis_interface.py
matthewturk: I think it should be `tmp`, yeah, but since it has been working I think it's possible that mpi4py was fixing it implicitly.

PR #4732, opened 2023-11-03
ENH: allow cell_widths and nproc > 1 for load_uniform_grid

### LINKED ISSUE #4330: load_uniform_grid with nprocs>1 and cell_widths
### Bug report

Calling `load_uniform_grid` with `nprocs>1` and `cell_widths` does not work. 

**Code for reproduction**

As of now, the following illustrates appears to work:

```python
import numpy as np
import yt

N = 16
data = {"density": np.random.random((N, N, N))}

cell_widths = []
for _ in range(3):
    cw = np.random.random(N)
    cw /= cw.sum()
    cell_widths.append(cw)

ds = yt.load_uniform_grid(
        data,
        data["density"].shape,
        bbox=np.array([[0.0, 1.0], [0.0, 1.0], [0.0, 1.0]]),
        cell_widths=cell_widths,
        nprocs=4,
    )
```
```
yt : [INFO     ] 2023-02-09 11:20:23,001 Parameters: current_time              = 0.0
yt : [INFO     ] 2023-02-09 11:20:23,002 Parameters: domain_dimensions         = [16 16 16]
yt : [INFO     ] 2023-02-09 11:20:23,002 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2023-02-09 11:20:23,003 Parameters: domain_right_edge         = [1. 1. 1.]
yt : [INFO     ] 2023-02-09 11:20:23,003 Parameters: cosmological_simulation   = 0
```
but do anything with that dataset:

```pythoin
dens = ds.all_data()[("stream", "density")]
```

and you get
```
  File "/home/chris/miniconda3/envs/yt_dev_39/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3398, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-e6b5073a1feb>", line 1, in <cell line: 1>
    dens = ds.all_data()[("stream", "density")]
  File "/home/chris/src/yt_/yt_dev/yt/yt/data_objects/static_output.py", line 1194, in all_data
    self.index
  File "/home/chris/src/yt_/yt_dev/yt/yt/data_objects/static_output.py", line 604, in index
    self._instantiated_index = self._index_class(
  File "/home/chris/src/yt_/yt_dev/yt/yt/frontends/stream/data_structures.py", line 167, in __init__
    GridIndex.__init__(self, ds, dataset_type)
  File "/home/chris/src/yt_/yt_dev/yt/yt/geometry/geometry_handler.py", line 39, in __init__
    self._setup_geometry()
  File "/home/chris/src/yt_/yt_dev/yt/yt/geometry/grid_geometry_handler.py", line 41, in _setup_geometry
    self._parse_index()
  File "/home/chris/src/yt_/yt_dev/yt/yt/frontends/stream/data_structures.py", line 209, in _parse_index
    self.grids.append(self.grid(id, self))
  File "/home/chris/src/yt_/yt_dev/yt/yt/frontends/stream/data_structures.py", line 87, in __init__
    cell_widths = index.grid_cell_widths[id - self._id_offset]
IndexError: list index out of range
```

**NOTE:** in #4328 I added a check for when `nprocs>1` being used with `cell_widths` in order to raise an error earlier during the initial call to `load_uniform_grid`. So when that gets merged, 

```python
ds = yt.load_uniform_grid(
        data,
        data["density"].shape,
        bbox=np.array([[0.0, 1.0], [0.0, 1.0], [0.0, 1.0]]),
        cell_widths=cell_widths,
        nprocs=4,
    )
```
will result in the following (or some similar variation):

```
NotImplementedError                       Traceback (most recent call last)
<ipython-input-1-bbf1cc62ea5c> in <module>
     11     cell_widths.append(cw)
     12 
---> 13 ds = yt.load_uniform_grid(
     14         data,
     15         data["density"].shape,

~/src/yt_/yt_dev/yt/yt/loaders.py in load_uniform_grid(data, domain_dimensions, length_unit, bbox, nprocs, sim_time, mass_unit, time_unit, velocity_unit, magnetic_unit, periodicity, geometry, unit_system, default_species_fields, axis_order, cell_widths, parameters)
    354         # cell_widths left as an empty guard value if None
    355         if nprocs != 1:
--> 356             raise NotImplementedError("nprocs must equal 1 if supplying cell_widths.")
    357         cell_widths = _validate_cell_widths(cell_widths, domain_dimensions)
    358 

NotImplementedError: nprocs must equal 1 if supplying cell_widths.
```

**Expected outcome**

A stretched grid decomponsed by nprocs.

Conceptually, I think the issue is just that `cell_widths` needs to be decomposed similar to how the data fields are already decomposed. Splitting a single stretched grid into multiple grid objects should be OK as long as they're all level 0, so I think it comes down to refactoring `load_uniform_grid` to also decompose the `cell_widths`.

chrishavlin:
## PR Summary

Working on getting `load_uniform_grid` to work with nproc > 1 for stretched grids (when `cell_widths` is allowed). Most the changes here are to the grid decomposition. 

Closes #4330 

## PR Checklist

- [x] Adds tests for new features.

Sample from the version just pushed:

```python
import numpy as np
import yt

N = 64
data = {"density": np.random.random((N, N, N))}

cell_widths = []
for _ in range(3):
    cw = 0.1 + np.random.random(N)
    cw /= cw.sum()
    cell_widths.append(cw)

ds = yt.load_uniform_grid(
        data,
        data["density"].shape,
        bbox=np.array([[0.0, 1.0], [0.0, 1.0], [0.0, 1.0]]),
        cell_widths=cell_widths,
        nprocs=32,
    )

slc = yt.SlicePlot(ds, 'x', 'density')
slc.set_log('density', False)
slc.annotate_grids(edgecolors='w', linewidth=4)
```

image

### DIFF
--- a/yt/utilities/decompose.py
+++ b/yt/utilities/decompose.py
@@ -103,7 +103,7 @@ def get_psize(n_d, pieces):
     return p_size
 
 
-def split_array(gle, gre, shape, psize, cell_widths=None):
+def split_array(gle, gre, shape, psize, *, cell_widths=None):
     """Split array into px*py*pz subarrays."""
     n_d = np.array(shape, dtype=np.int64)
     dds = (gre - gle) / shape
@@ -111,12 +111,11 @@ def split_array(gle, gre, shape, psize, cell_widths=None):
     right_edges = []
     shapes = []
     slices = []
+
     if cell_widths is None:
         cell_widths_by_grid = None
     else:
         cell_widths_by_grid = []
-    if isinstance(cell_widths, list) and len(cell_widths) == 0:
-        cell_widths = None
 
     for i in range(psize[0]):
         for j in range(psize[1]):
@@ -130,7 +129,6 @@ def split_array(gle, gre, shape, psize, cell_widths=None):
                     offset_le = []
                     offset_re = []
                     for idim in range(3):
-                        # round off error issues here maybe?
                         cws.append(cell_widths[idim][lei[idim] : rei[idim]])
                         offset_le.append(np.sum(cell_widths[idim][0 : lei[idim]]))
                         offset_re.append(offset_le[idim] + np.sum(cws[idim]))
--- a/yt/utilities/decompose.py
+++ b/yt/utilities/decompose.py
@@ -17,11 +17,11 @@ def decompose_to_primes(max_prime):
         yield max_prime
 
 
-def decompose_array(shape, psize, bbox):
+def decompose_array(shape, psize, bbox, *, cell_widths=None):
     """Calculate list of product(psize) subarrays of arr, along with their
     left and right edges
     """
-    return split_array(bbox[:, 0], bbox[:, 1], shape, psize)
+    return split_array(bbox[:, 0], bbox[:, 1], shape, psize, cell_widths=cell_widths)
 
 
 def evaluate_domain_decomposition(n_d, pieces, ldom):
@@ -103,7 +103,7 @@ def get_psize(n_d, pieces):
     return p_size
 
 
-def split_array(gle, gre, shape, psize):
+def split_array(gle, gre, shape, psize, cell_widths=None):
     """Split array into px*py*pz subarrays."""
     n_d = np.array(shape, dtype=np.int64)
     dds = (gre - gle) / shape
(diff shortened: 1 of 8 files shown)


chrishavlin:
needs tests! but seems to work?

chrishavlin:
> but seems to work?
also seems to have broken many unexpected things :) definitely WIP

chrishavlin:
@yt-fido test this please

chrishavlin:
@yt-fido test this please

yt/utilities/decompose.py
```diff
@@ -103,25 +103,46 @@ def get_psize(n_d, pieces):
...
+                    offset_re = rei * dds
+                lle = gle + offset_le
+                lre = gle + offset_re
                 left_edges.append(lle)
                 right_edges.append(lre)
                 shapes.append(rei - lei)
                 slices.append(np.s_[lei[0] : rei[0], lei[1] : rei[1], lei[2] : rei[2]])
 
-    return left_edges, right_edges, shapes, slices
+    return left_edges, right_edges, shapes, slices, cell_widths_by_grid
```
neutrinoceros, line 148: this looks like a backward incompatible change. Returning `cell_widths_by_grid` should be optional, and the default behaviour should remain stable, unless there's a consensus that `yt.utilities.decompose` isn't considered public API ?

neutrinoceros (changes requested):
As long as the test pass I'm okay with this, but I believe that the signature changes in what looks like public API functions should be discussed, or changed for backward compatibility.

yt/utilities/decompose.py
jzuhone: It's not public API, so this should be ok by me.

chrishavlin: oh! ya, I had assumed without too much thought that this was not public API. I think there is a small chance that there's an external frontend out there that utilizes this function. Happy to change this up to cover that scenario: the simplest change might be to have the return from this function vary and include a deprecation, e.g., 

```python
if cell_widths is not None:
    return left_edges, right_edges, shapes, slices, cell_widths_by_grid
else:
    deprecation warning
    return left_edges, right_edges, shapes, slices
```
Could leave out the deprecation but that might make future typing efforts annoying.

Or I could add a `split_array_with_cell_widths` function and just call that directly from the `load_uniform_grid` when needed. That'd probably result in a bit of code duplication, but it might be clearer to have a separate function. 



Also happy to leave this change as-is.

neutrinoceros: Following John's remark I think the best course of action is to keep it simple and treat it as private. We can always inject additional backward-compatibility fixes later *if* anyone asks for it.

neutrinoceros (approved):

PR #4733, opened 2023-11-04
RFC: refactor usage of numpy.random legacy API

neutrinoceros:
## PR Summary

Enable numpy-specific rules in ruff and manually fix usage of legacy API everywhere except in tests.
The reasoning here is that ruff recently introduced a new rule `NPY201` to check and auto refactor usage of many functions and classes that are marked as deprecated in numpy 2.0 (`NPY201` is currently off by default). Enabling this ruleset early will provide us with an additional safeguard against incompatibilities with numpy 2.0

Interestingly, this rule (`NPY201`) was implemented in ruff by the person who's doing the numpy 2.0 API overhaul.


neutrinoceros:
@yt-fido test this please

yt/frontends/gadget/testing.py
```diff
@@ -96,7 +96,8 @@ def fake_gadget_binary(
             # Generate and write field block
             data = []
             for pt in ptype:
-                data += [np.random.rand(npart[pt], dim)]
+                rng = np.random.default_rng()
+                data += [rng.random((npart[pt], dim))]
```
chrishavlin, line 100: Any reason to not move `rng` outside the loop?

i.e., 
```
            rng = np.random.default_rng()
            for pt in ptype:                
                data += [rng.random((npart[pt], dim))]
```

chrishavlin (approved):
LGTM, left one question

yt/frontends/gadget/testing.py
neutrinoceros: Done !

chrishavlin (approved):

neutrinoceros:
@matthewturk can we merge this ?

neutrinoceros:
I guess this doesn't *need* to be backported, as the legacy API isn't actually deprecated, only discouraged.

PR #4734, opened 2023-11-06
Read AMR domains lazily

cphyc:
## PR Summary

For RAMSES, we already have some domain-based decomposition following a Hilbert curve. This leverages this to quickly find which domain to read when selecting a subregion, allowing to (costly) reading of many AMR domains.

## New things
- When using selection containers, only the domains that intersect the bounding box are now checked for intersection and their AMR structure read So if you're only reading 1% of the simulation, you do not need to read most of the AMR
- When reading particles, we actually do not need to read the AMR structure. The benefit of the improvement above is that it allows us to completely ignore the AMR when reading only the particles. This is a MASSIVE speedup for DM-only simulations, or when focusing only on e.g. stars in a galaxy formation simulation.

This sits nicely on top of https://github.com/yt-project/yt/pull/4720 which allows much improved performances. I'd be happy to add some figures if required.


cphyc:
@yt-fido test this please

yt/frontends/ramses/hilbert.py
```diff
@@ -259,10 +95,50 @@ def hilbert3d(X, bit_length):
...
+    """
+    if LE is None:
+        LE = np.array(
+            [
+                0,
+                0,
+                0,
+            ],
+            dtype="d",
+        )
```
neutrinoceros, line 112: I think the trailing comma after the last "0" is causing this line to be expanded by black, removing it should stabilise styling and readability
```suggestion
        LE = np.array([0, 0, 0], dtype="d")
```

yt/frontends/ramses/data_structures.py
```diff
@@ -197,7 +204,7 @@ def __init__(self, ds, domain_id):
         for t in ["grav", "amr"]:
             setattr(self, f"{t}_fn", basename % t)
         self._part_file_descriptor = part_file_descriptor
-        self._read_amr_header()
+        self.max_level = self.ds.parameters["levelmax"] - self.ds.parameters["levelmin"]
```
matthewturk, line 207: Good catch.  I feel like I've seen this as an issue elsewhere.  Are we correctly setting `min_level` as well?

matthewturk (commented):
This looks good.  Should more of the properties in the domain files be converted to `cached_properties` to reduce the levels of indirection?  Like for instance, `amr_header`?

yt/frontends/ramses/data_structures.py
cphyc: I have to say I'm a bit confused about `max_level`/`min_level`. In RAMSES standard, these are simply related to the number of levels of refinement from the root, i.e. `1/2**level` is the current cell size.
But for yt, if `max_level` is how more refined the deepest level is compared to the base grid, what does `min_level` min? Should it be 0 if there are some cells that are at the resolution of the base grid?

cphyc:
> This looks good. Should more of the properties in the domain files be converted to `cached_properties` to reduce the levels of indirection? Like for instance, `amr_header`?

I am not sure what you exactly mean. Are you suggesting to add _more_ things as `cached_properties` to make the dataset as lazy as possible, or the opposite to reduce indirections?

yt/frontends/ramses/io.py
```diff
@@ -306,6 +310,7 @@ def _read_particle_subset(self, subset, fields):
         return tr
 
 
+@lru_cache(maxsize=32)
```
neutrinoceros, line 313: just curious, why 32 ?

yt/frontends/ramses/particle_handlers.py
```diff
@@ -101,6 +105,15 @@ def read_header(self):
         pass
 
 
+_default_dtypes: dict[int, str] = {
+    1: "c",  # char
+    2: "h",  # short,
+    4: "f",  # float
+    8: "d",  # double
+    # 16: np.float128,  # double double precision
```
neutrinoceros, line 113: should this be uncommented ?

yt/frontends/ramses/particle_handlers.py
```diff
@@ -138,13 +151,13 @@ def read_header(self):
...
-        flen = fd.tell()
-        fd.seek(0)
-        hvals = {}
-        attrs = self.attrs
-        hvals.update(fd.read_attrs(attrs))
+        flen = os.path.getsize(self.fname)
+        with FortranFile(self.fname) as fd:
+            hvals = {}
+            attrs = self.attrs
+            hvals.update(fd.read_attrs(attrs))
```
neutrinoceros, line 158: would this work ?
```suggestion
            hvals = dict(fd.read_attrs(self.attrs))
```

yt/frontends/ramses/particle_handlers.py
cphyc: It should be removed, because the code path will not work (not supported by our reader at the moment.)

yt/frontends/ramses/data_structures.py
```diff
@@ -251,79 +251,139 @@ def amr_file(self):
...
-        self.amr_header = hvals
+
         # update levelmax
         force_max_level, convention = self.ds._force_max_level
         if convention == "yt":
             force_max_level += self.ds.min_level + 1
-        self.amr_header["nlevelmax"] = min(
-            force_max_level, self.amr_header["nlevelmax"]
+        self._amr_header["nlevelmax"] = min(
+            force_max_level, self._amr_header["nlevelmax"]
```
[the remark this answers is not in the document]
cphyc: I merely added `_` to signal that these are private properties. While we may want to create a setter, we're already in a function (`_read_amr_header`) whose main job is to fill this dictionary. What purpose would a property setter fulfil?

yt/frontends/ramses/hilbert.py
```diff
@@ -259,10 +105,50 @@ def hilbert3d(X, bit_length):
...
 
-def get_cpu_list(ds, X):
+def get_cpu_list(
+    ds,
+    region: YTRegion,
+    LE: Optional["np.ndarray[Any, np.dtype[np.float64]]"] = None,
+    dx: float = 1.0,
+    dx_cond: Optional[float] = None,
+    factor: float = 4.0,
+) -> set[int]:
```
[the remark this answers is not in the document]
cphyc: I renamed the function to `get_intersecting_cpus` to make it more explicit and less misleading.

yt/frontends/ramses/io.py
cphyc: I dropped the 32 :)

cphyc:
@yt-fido test this please.

yt/frontends/ramses/io.py
neutrinoceros: so... why 128 then ? 😅

cphyc: This is a question for the CPython mailing list ;)

neutrinoceros: Fine, I don't really mind :)

neutrinoceros (dismissed):
All my points were addressed, thank you !

neutrinoceros (approved):

PR #4418, opened 2023-04-16
DOC: document pytest-mpl update workflow

neutrinoceros:
## PR Summary

Complements #4122, partially adresses #4415

note to reviewers: this workflow may seem very cumbersome (it is), but it's not _new_.
In fact it is extremely close to what we were doing even before #4122, it just happens that this process was never properly documented.

Opening as a draft for now because I need to come back and update the `How to Write Image Comparison Tests` section too.

PR #4740, opened 2023-11-14
ENH: enable quiver callback for geographic geometries

chrishavlin:
This PR enables the quiver callback for geographic geometries. 

### example 1

Here's an example with some real data using `yt_xarray` to load MERRA-2 data into yt

```python
import yt
import yt_xarray
dsx = yt_xarray.open_dataset("sample_nc/MERRA2_100.inst3_3d_asm_Np.19800120.nc4")
ds = dsx.yt.load_grid(fields=['U', 'V'], sel_dict={'time':0}, use_callable=False)

slc = yt.SlicePlot(ds, 'altitude', 'U')
slc.set_log('U', False)
slc.annotate_quiver('U', 'V')
slc.save("geoquiver_merra2_U.png")

slc = yt.SlicePlot(ds, 'altitude', 'V')
slc.set_log('V', False)
slc.annotate_quiver('U', 'V')
slc.save("geoquiver_merra2_V.png")
```

geoquiver_merra2_U

geoquiver_merra2_V

Data is from the MERRA-2 reanalysis dataset ([direct data link](https://disc.gsfc.nasa.gov/datasets/M2I3NPASM_5.12.4/summary)). Quiver arrows vary with the plotted velocity components as expected.

### native yt examples

And a couple examples that just use yt 

#### Random data 

```python
import yt 
ds = fake_amr_ds(geometry='geographic', 
                 fields=['density', 'velocity_ew', 'velocity_ns'],
                 units = ['kg/m**3', 'm/s', 'm/s'])

slc = yt.SlicePlot(ds,'altitude', 'density')
slc.annotate_quiver(('stream','velocity_ew'), ('stream','velocity_ns'))
slc.show()
```
geo_quiver

#### Not random data

```python
import numpy as np
import yt 

def _vel_calculator(grid, ax):
    y_lat = grid.fcoords[:, 1].d
    x_lon = grid.fcoords[:, 2].d
    x_lon[x_lon>180] = x_lon[x_lon>180] - 360.
    dist = np.sqrt((y_lat - yc)**2 + (xc - x_lon)**2)
    if ax == 1:
        sgn = np.sign(y_lat - yc)    
    elif ax == 2:
        sgn = np.sign(x_lon - xc)    
    
    vel = np.exp(-(dist/45)**2) * sgn
    return vel.reshape(grid.shape)

def _calculate_u(grid, field):        
    return _vel_calculator(grid, 2)

def _calculate_v(grid, field):        
    return _vel_calculator(grid, 1)

data = {'u_vel': _calculate_u, 
             'v_vel': _calculate_v}

bbox = [[10., 1000], 
        [-90, 90], 
        [0, 360]]
bbox = np.array(bbox)

ds = yt.load_uniform_grid(data, 
                          (16,128,32), 
                          bbox=bbox, 
                          geometry='geographic', 
                          axis_order=('altitude','latitude','longitude'))

slc = yt.SlicePlot(ds, 'altitude', 'u_vel')
slc.set_log('u_vel', False)
slc.annotate_quiver('u_vel', 'v_vel', scale=50)
slc.save('u_vel_test.png')

slc = yt.SlicePlot(ds, 'altitude', 'v_vel')
slc.set_log('v_vel', False)
slc.annotate_quiver('u_vel', 'v_vel', scale=50)
slc.save('v_vel_test.png')
```
u_vel_test
v_vel_test


neutrinoceros:
I think we need to showcase this with actual data, or at least fake data that has known structures. Using random noise makes it too easy to look over mistakes.

chrishavlin:
sure! was actually planning to replace the random data in the example with a `load_uniform_grid` example.

yt/visualization/tests/test_callbacks_geographic.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+from yt import SlicePlot
+from yt.testing import assert_fname, fake_amr_ds, requires_module
+
+
+@requires_module("cartopy")
+@pytest.mark.parametrize("geometry", ["geographic", "internal_geographic"])
+def test_quiver_callback_geographic(geometry, tmp_path):
```
neutrinoceros, line 9: we could make this an actual image test. The documentation for how to do that is still not merged (I think I need to finish the PR first ...) but it's there: https://github.com/yt-project/yt/pull/4418

chrishavlin: yeah, it'd be a good image test. I'll switch it over after I revise the test data a bit.

chrishavlin:
a couple more variations on the above MERRA-2 plots (at different pressure levels): https://github.com/chrishavlin/yt_scratch/blob/master/notebooks/geodata/geoquiver.ipynb

chrishavlin:
and just added another yt-native example that might make a good image test.

chrishavlin:
@yt-fido test this please

yt/visualization/plot_modifications.py
```diff
@@ -733,10 +734,30 @@ def __call__(self, plot):
...
+            )
+        else:
+            # when we have a cartopy transform, provide the x, y values
+            # in the coordinate reference system of the data and let cartopy
+            # do the transformation. Also check for the exact bounds of the transform
+            # which can cause issues with projections.
+            if _check_geographic_bounds(bounds, plot._transform):
+                # note: cartopy will also raise its own warning, but it is useful to add this
+                # warning as well since the only way to avoid the exact bounds is to change the
+                # extent of the plot.
```
chrishavlin, line 751: note to reviewers: I initially pushed a version that offset the bounds by a small amount to avoid the exact geographic bounds of the transform but in the end decided it was better to just warn the user and let them change the plot width if they want to silence the warning. I could go back to the offset approach, but there might be some annoying edge cases to worry about in deciding what value to use for the offset.

neutrinoceros:
@chrishavlin thanks for adding an image test. You'll need to open a PR at https://github.com/yt-project/yt_pytest_mpl_baseline with the reference image (preferably the one generated in CI) in order for the test to pass here. Other than that, LGTM !

chrishavlin:
I think this is good to go (but could use another review). The accompanying PR for the new image baseline is over at https://github.com/yt-project/yt_pytest_mpl_baseline/pull/5

chrishavlin:
@yt-fido test this please

yt/visualization/geo_plot_utils.py
```diff
@@ -115,3 +115,15 @@ def get_mpl_transform(mpl_proj) -> Optional[FunctionType]:
     if key is None:
         instantiated_func = None
     return instantiated_func
+
+
+def _geographic_bounds(transform) -> tuple:
```
neutrinoceros, line 120: could we type the tuple itself too ?

yt/visualization/geo_plot_utils.py
```diff
@@ -115,3 +115,15 @@ def get_mpl_transform(mpl_proj) -> Optional[FunctionType]:
...
+    xlims: tuple = transform.x_limits
+    ylims: tuple = transform.y_limits
+    return xlims + ylims
+
+
+def _check_geographic_bounds(bounds: tuple, transform) -> bool:
+    # returns True if any of the bounds match the exact transform bounds
+    transform_bounds = _geographic_bounds(transform)
+    return any(bounds[bid].d == transform_bounds[bid] for bid in range(4))
```
neutrinoceros, line 129: This function doesn't need to know how long we expect this iterables to be
```suggestion
    return any(b.d == tb for b, tb in zip(bounds, transform_bounds))
```

neutrinoceros: also, since this is a small and private helper function, inlining it, or defining it closer to where it's actually used, would help readability of the callsite

yt/visualization/plot_modifications.py
```diff
@@ -733,10 +734,30 @@ def __call__(self, plot):
...
+                mylog.warning(
+                    "Using the exact bounds of the transform may cause errors at the bounds."
+                    " To avoid this warning, adjust the width of your plot object to not include "
+                    "the bounds."
+                )
```
neutrinoceros, line 756: To me, the main selling point of using logger warnings is that they can be turned off without affecting how other libraries behave, so I tend to reserve them for stuff that's not easily actionable from an end-user perspective. This feels like it should be a `warnings.warn`, which are also much easier to test for.

yt/visualization/plot_modifications.py
```diff
@@ -769,6 +790,10 @@ def _finalize(self, plot, X, Y, pixX, pixY, pixC):
             "scale_units": self.scale_units,
         }
         kwargs.update(self.plot_args)
+
+        if plot._transform is not None:
+            kwargs["transform"] = plot._transform
```
neutrinoceros, line 795: what if there's already a `transform` keyword argument ? As is, it would simply be silently dropped, which may be the source of subtle bugs down the road.

yt/visualization/tests/test_callbacks_geographic.py
```diff
@@ -0,0 +1,74 @@
...
+    ds = fake_amr_ds(fields=flds, units=units, geometry=geometry)
+    tmpfi = str(tmp_path / "geo_quiver.png")
+
+    for ax in ds.coordinates.axis_order:
+        slc = SlicePlot(ds, ax, "density", buff_size=(50, 50))
+        if ax == ds.coordinates.radial_axis:
+            slc.set_width((359.99, 179.99))
+        slc.annotate_quiver(("stream", "velocity_ew"), ("stream", "velocity_ns"))
+        slc.save(tmpfi)
+        assert_fname(tmpfi)
```
neutrinoceros, line 23: IO is probably making this test longer than it needs to be. Could we just use `slc.render()` instead ? I think it should be sufficient to validate that no exception is raised when the callback actually gets run.

yt/visualization/tests/test_callbacks_geographic.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@pytest.fixture(scope="module")
```
neutrinoceros, line 26: I'm curious, why specify a module scope here ? Does it make a difference if the fixture is only used in one test ?

neutrinoceros (commented):
I'm game to help pushing this forward, here are a couple comments !

yt/visualization/geo_plot_utils.py
chrishavlin: ok, i'll just move it down to where it's actually used. don't remember why i thought a separate function would be clearer...

yt/visualization/plot_modifications.py
chrishavlin: ok! i'll switch it over.

yt/visualization/tests/test_callbacks_geographic.py
chrishavlin: yup, just `render` should be fine, will switch.

yt/visualization/tests/test_callbacks_geographic.py
chrishavlin: I wrote this long enough ago that i dont remember why I provided a scope here... maybe I had planned on writing more tests? but since as you say it's just being used by one test, it wont make a difference, I'll remove it.

yt/visualization/plot_modifications.py
chrishavlin: added a warning.

yt/visualization/plot_modifications.py
chrishavlin: switched

yt/visualization/geo_plot_utils.py
chrishavlin: since these functions were only used once, I just got rid of them and moved the code to where it's needed, so this type hint is gone.

neutrinoceros:
Alright, I just merged the companion PR, so you should be able to bump the submodule here and get CI back to green !

neutrinoceros (approved):

neutrinoceros:
Thanks for you patience on this one @chrishavlin !

PR #4741, opened 2023-11-15
Use stamping approach to deposit off-axis cells

cphyc:
## PR Summary

On main, off-axis projections are very slow due to the need to cast rays through the domain.
This is particularly the case for octree-datasets to to the hot-loop [for grid, mask in data_source.blocks:](https://github.com/cphyc/yt/blob/f8d38de9fc71cd83b40ab74eb2527644bbc1f4be/yt/visualization/volume_rendering/off_axis_projection.py#L446-L464). Indeed, this iterates over all octs in the domain, one-by-one.

While one solution might be to move this chunk to a Cython-accelerated code, it will remain somewhat slow no matter what.

In this PR, I take a completely different approach.
In the first part of the code, I compute the projected depth of a cube as seen from the camera position (i.e. how long is the segment traversing the cube at different pixel locations). One way of seeing this is that we now have a precomputed kernel, which we can apply to each cell at a scale proportional to the cell size.

This yields good performances: ~2s on an i9 for an 800x800 simulation with 10,000,000 cells. Interestingly, this could also be used to offload the rendering to the GPU.

While the code is relatively stable, I still get slight differences between axis-aligned projections and this method, which suggests that there are some minor details that I got wrong (see below).

## Comparisons
Using the code
```python
import yt

p = yt.ProjectionPlot(
    ds,
    "y",
    "density",
    weight_field="density"
)
p.set_zlim(("gas", "density"), 6e-31, 2e-26)
p.save("/tmp/projection.png")
p = yt.OffAxisProjectionPlot(
    ds,
    [0, 1, 0],
    "density",
    north_vector=[1, 0, 0],
    width=(1, "unitary"),
    center=[0.5, 0.5, 0.5],
    buff_size=(800, 800),
    weight_field="density",
)
p.set_zlim(("gas", "density"), 6e-31, 2e-26)
p.set_axes_unit("Mpc")
p.save("/tmp/new.png")
```

I obtain (as of #f8d38de) very similar performances as the on-axis projection (!!!)
|  | Result from `ProjectionPlot` | Result from new off-axis projection |
|--|---|---|
| Image | projection | new |
| Timing (s) | 1.45 | 2.14 (8 cores), 3.13 (1 core) |

### DIFF
--- a/yt/utilities/lib/image_utilities.pyx
+++ b/yt/utilities/lib/image_utilities.pyx
@@ -92,8 +92,8 @@ cdef void _add_cell_to_image_offaxis(
         return
 
     depth = int(ceil(log2(dx * float(max(Nx, Ny)))))
-    if depth <= 0:
-        depth = 0
+    if depth < 1:
+        depth = 1
     if depth >= max_depth:
         depth = max_depth - 1
 
--- a/yt/utilities/lib/image_utilities.pyx
+++ b/yt/utilities/lib/image_utilities.pyx
@@ -41,7 +41,6 @@ def add_points_to_greyscale_image(
         buffer_mask[i, j] = 1
     return
 
-# @cython.boundscheck(False)
 @cython.cdivision(True)
 @cython.wraparound(False)
 @cython.boundscheck(False)
@@ -64,9 +63,9 @@ cdef void _add_cell_to_image_offaxis(
 ) noexcept nogil:
     cdef np.float64_t lx, rx, ly, ry
     cdef int j, k, depth
-    cdef int jmin, jmax, kmin, kmax, jj, kk
+    cdef int jmin, jmax, kmin, kmax, jj1, jj2, kk1, kk2
     cdef np.float64_t cell_max_half_width = cell_max_width / 2
-    cdef np.float64_t xx, yy, dvx, dvy, sw, swq
+    cdef np.float64_t xx1, xx2, yy1, yy2, dvx, dvy, sw, swq, dx_loc, dy_loc
 
     lx = x - cell_max_half_width
     rx = x + cell_max_half_width
@@ -91,47 +90,64 @@ cdef void _add_cell_to_image_offaxis(
         buffer_weight[jmin][kmin] += w
         return
 
-    depth = int(ceil(log2(dx * float(max(Nx, Ny)))))
-    if depth < 1:
-        depth = 1
-    if depth >= max_depth:
-        depth = max_depth - 1
+    depth = int(ceil(log2(2 * sqrt(3) * dx * float(max(Nx, Ny)))))
+    if depth < 1: depth = 1
+    if depth >= max_depth: depth = max_depth - 1
 
     jmax = min(Nsx, 1 << depth)
     kmax = min(Nsy, 1 << depth)
-    for j in range(jmax):
-        xx = ((j + 0.5 - jmax / 2) / jmax * cell_max_width + x) * Nx
-        jj = int(xx)
-        if jj < 0 or jj >= Nx: continue
 
-        dvx = 1 - (xx - jj)
+    dx_loc = cell_max_width / jmax
+    dy_loc = cell_max_width / kmax
+
+    # with gil: print(jmax, kmax)
+
+    for j in range(jmax):
+        xx1 = ((j - jmax / 2.) * dx_loc + x) * Nx
+        xx2 = ((j + 1 - jmax / 2.) * dx_loc + x) * Nx
+
+        jj1 = int(xx1)
+        jj2 = int(xx2)
+        # The subcell is out of the projected area
+        if jj2 < 0 or jj1 >= Nx: continue
+
+        dvx = max(0, min(1, (jj2 - xx1) / (xx2 - xx1)))
+
         for k in range(kmax):
             if stamp_mask[depth, j, k] == 0:
                 continue
 
-            yy = ((k + 0.5 - kmax / 2) / kmax * cell_max_width + y) * Ny
-            kk = int(yy)
-            if kk < 0 or kk >= Ny: continue
+            yy1 = ((k - kmax / 2.) * dy_loc + y) * Ny
+            yy2 = ((k + 1 - kmax / 2.) * dy_loc + y) * Ny
 
-            dvy = 1 - (yy - kk)
+            kk1 = int(yy1)
+            kk2 = int(yy2)
+            # The subcell is out of the projected area
+            if kk2 < 0 or kk1 >= Ny: continue
 
-            swq = stamp[depth, j, k] * w * q
+            # with gil:
+            #     print(f"{j=} {k=} {xx1=} {xx2=} {yy1=} {yy2=} {jj1=} {jj2=} {kk1=} {kk2=}")
+
             sw =  stamp[depth, j, k] * w
+            swq = sw * q
 
-            buffer[jj][kk]        += swq * (dvx * dvy)
-            buffer_weight[jj][kk] +=  sw * (dvx * dvy)
+            dvy = max(0, min(1, (kk2 - yy1) / (yy2 - yy1)))
 
-            if jj < Nx - 1:
-                buffer[jj + 1][kk]        += swq * (1 - dvx) * dvy
-                buffer_weight[jj + 1][kk] +=  sw * (1 - dvx) * dvy
+            if jj1 >= 0 and kk1 >= 0:
+                buffer[jj1][kk1]        += swq * dvx * dvy
+                buffer_weight[jj1][kk1] +=  sw * dvx * dvy
 
-            if kk < Ny - 1:
-                buffer[jj][kk + 1]        += swq * dvx * (1 - dvy)
-                buffer_weight[jj][kk + 1] +=  sw * dvx * (1 - dvy)
-            if jj < Nx - 1 and kk < Ny - 1:
-                buffer[jj + 1][kk + 1]        += swq * (1 - dvx) * (1 - dvy)
-                buffer_weight[jj + 1][kk + 1] +=  sw * (1 - dvx) * (1 - dvy)
+            if jj1 >= 0 and kk2 < Ny:
+                buffer[jj1][kk2]        += swq * dvx * (1 - dvy)
+                buffer_weight[jj1][kk2] +=  sw * dvx * (1 - dvy)
 
+            if jj2 < Nx and kk1 >= 0:
+                buffer[jj2][kk1]        += swq * (1 - dvx) * dvy
+                buffer_weight[jj2][kk1] +=  sw * (1 - dvx) * dvy
+
+            if jj2 < Nx and kk2 < Ny:
+                buffer[jj2][kk2]        += swq * (1 - dvx) * (1 - dvy)
+                buffer_weight[jj2][kk2] +=  sw * (1 - dvx) * (1 - dvy)
 
 @cython.boundscheck(False)
 @cython.cdivision(True)
@@ -170,8 +186,8 @@ def add_cells_to_image_offaxis(
     cdef int Nsx, Nsy
     cdef np.float64_t dx_max = np.max(dXp)
     # The largest cell needs to be resolved by at least this number of pixels
-    Nsx = max(Npix_min, int(ceil(dx_max * float(Nx))))
-    Nsy = max(Npix_min, int(ceil(dx_max * float(Ny))))
+    Nsx = max(Npix_min, int(ceil(2 * dx_max * sqrt(3) * Nx)))
+    Nsy = max(Npix_min, int(ceil(2 * dx_max * sqrt(3) * Ny)))
     cdef int max_depth = int(ceil(log2(max(Nsx, Nsy))))
     cdef int depth
 
--- a/yt/utilities/lib/image_utilities.pyx
+++ b/yt/utilities/lib/image_utilities.pyx
@@ -11,7 +11,7 @@ import numpy as np
 
 cimport numpy as np
 cimport cython
-from libc.math cimport M_PI, M_PI_2, atan2, ceil, floor, log2, sqrt
+from libc.math cimport ceil, floor, log2, sqrt
 from libc.stdlib cimport free, malloc
 
 from yt.utilities.lib.fp_utils cimport iclip
@@ -86,8 +86,8 @@ cdef void _add_cell_to_image_offaxis(
 
     # If the cell is fully within one pixel
     if (jmax == jmin + 1) and (kmax == kmin + 1):
-        buffer[jmin][kmin] += w * q
-        buffer_weight[jmin][kmin] += w
+        buffer[jmin][kmin] += w * q * dx
+        buffer_weight[jmin][kmin] += w * dx
         return
 
     depth = int(ceil(log2(2 * sqrt(3) * dx * float(max(Nx, Ny)))))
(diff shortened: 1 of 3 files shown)


yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+
+    # If the cell is fully within one pixel
+    if (jmax == jmin + 1) and (kmax == kmin + 1):
+        buffer[jmin][kmin] += w * q * dx
+        buffer_weight[jmin][kmin] += w * dx
+        return
+
+    depth = int(ceil(log2(2 * sqrt(3) * dx * float(max(Nx, Ny)))))
+    if depth < 1: depth = 1
+    if depth >= max_depth: depth = max_depth - 1
```
matthewturk, line 95: could we use `iclip`?

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+
+    jmax = min(Nsx, 1 << depth)
+    kmax = min(Nsy, 1 << depth)
```
matthewturk, line 98: `imin`?  I think this might call the python function `min`.

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+    for j in range(jmax):
+        xx1 = ((j - jmax / 2.) * dx_loc + x) * Nx
+        xx2 = ((j + 1 - jmax / 2.) * dx_loc + x) * Nx
+
+        jj1 = int(xx1)
+        jj2 = int(xx2)
+        # The subcell is out of the projected area
+        if jj2 < 0 or jj1 >= Nx: continue
+
+        dvx = max(0, min(1, (jj2 - xx1) / (xx2 - xx1)))
```
matthewturk, line 114: `iclip` again

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+
+    # with gil: print(jmax, kmax)
+
+    for j in range(jmax):
+        xx1 = ((j - jmax / 2.) * dx_loc + x) * Nx
+        xx2 = ((j + 1 - jmax / 2.) * dx_loc + x) * Nx
+
+        jj1 = int(xx1)
+        jj2 = int(xx2)
+        # The subcell is out of the projected area
```
matthewturk, line 111: Can we cast these instead of function calling?

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+            # The subcell is out of the projected area
+            if kk2 < 0 or kk1 >= Ny: continue
+
+            # with gil:
+            #     print(f"{j=} {k=} {xx1=} {xx2=} {yy1=} {yy2=} {jj1=} {jj2=} {kk1=} {kk2=}")
+
+            sw =  stamp[depth, j, k] * w * dx
+            swq = sw * q
+
+            dvy = max(0, min(1, (kk2 - yy1) / (yy2 - yy1)))
```
matthewturk, line 134: `iclip`

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+
+        for k in range(kmax):
+            if stamp_mask[depth, j, k] == 0:
+                continue
+
+            yy1 = ((k - kmax / 2.) * dy_loc + y) * Ny
+            yy2 = ((k + 1 - kmax / 2.) * dy_loc + y) * Ny
+
+            kk1 = int(yy1)
+            kk2 = int(yy2)
```
matthewturk, line 124: cast here too?

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+    sq3 = sqrt(3.)
+
+    # Local buffers
+    cdef np.float64_t **lbuffer
+    cdef np.float64_t **lbuffer_weight
+
+    cdef int num_particles = len(Xp)
+
+    with nogil, parallel():
+        lbuffer = <np.float64_t**> malloc(sizeof(np.float64_t*) * Nx)
```
matthewturk, line 228: Is this a better solution than allocating the full `Nx*Ny` array?  If I am reading it right, this will still allocate the full thing on every processor, won't it?

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+        lbuffer_weight = <np.float64_t**> malloc(sizeof(np.float64_t*) * Nx)
+        for j in range(Nx):
+            lbuffer[j] = <np.float64_t*> malloc(sizeof(np.float64_t) * Ny)
+            lbuffer_weight[j] = <np.float64_t*> malloc(sizeof(np.float64_t) * Ny)
+            for k in range(Ny):
+                lbuffer[j][k] = 0
+                lbuffer_weight[j][k] = 0
```
matthewturk, line 235: might be able to use `calloc` instead of this

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+        # Free memory
+        for j in range(Nx):
+            free(lbuffer[j])
+            free(lbuffer_weight[j])
+        free(lbuffer)
+        free(lbuffer_weight)
+
+
+@cython.boundscheck(False)
+cdef inline np.float64_t det2d(const np.float64_t[::1] a, const np.float64_t[::1] b) noexcept nogil:
```
matthewturk, line 286: this might still do an initialization check, which can be expensive

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+    return a[0] * b[1] - a[1] * b[0]
+
+@cython.cdivision(True)
+cdef bint check_in_parallelogram(
```
matthewturk, line 290: boundscheck and initializedcheck might add speed

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+                zmax = max(z, zmax)
+                Nhit += 1
+
+            within = check_in_parallelogram(ObackA, w2d, u2d, -1, -1, nm)
+            if within:
+                z = Oback[2] + nm[0] * u[2] + nm[1] * w[2]
+                zmin = min(z, zmin)
+                zmax = max(z, zmax)
+                Nhit += 1
+
```
matthewturk, line 399: I think if we use the Cythonized `imin` and `imax` it might go a bit faster

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+            if Nhit == 0:
+                continue
+            elif Nhit == 1:
+                raise RuntimeError("This should not happen")
+            else:
+                buffer[i, j] += zmax - zmin
+                buffer_mask[i, j] |= 1
```
matthewturk, line 406: Doesn't even need to be an or -- could just set it.

chrishavlin:
No insight into why you're getting different results between the on-axis projection and the off-axis when aligned with the axis, but here's an example that results in very different behavior. The standard on axis projection returns what I would expect:

```python
fld = ("ramses", "z")
wt = ("index", "ones")
p = yt.ProjectionPlot(
    ds, 
    "y",
    fld,
    weight_field=wt,
    buff_size=(800,800),
)
p.set_axes_unit("Mpc")
p.set_unit(fld, "Mpc")
p.save('old.png')
```
old

But the new off axis produces a constant value of 1? 

```python
p = yt.OffAxisProjectionPlot(
    ds, 
    [0, 1, 0], 
    fld,
    north_vector=[1, 0, 0], 
    width=(1, "unitary"),
    center=[0.5, 0.5, 0.5],
    buff_size=(800, 800),
    weight_field=wt,
)
p.set_axes_unit("Mpc")
p.set_unit(fld, "Mpc")

mx = p.plots[fld].image.get_array().max()
mn = p.plots[fld].image.get_array().min()
print((mn, mx))
# (1.0, 1.0)
p.save('new.png')
```
new

Been reading and stepping through the code, I kinda think it's related to how the weighting field is handled? maybe units related?

yt/utilities/lib/image_utilities.pyx
```diff
@@ -36,6 +41,370 @@ def add_points_to_greyscale_image(
...
+    const np.float64_t[::1] qty,
+    const np.float64_t[::1] weight,
+    const np.float64_t[:, :] rotation,
+    np.float64_t[:, ::1] buffer,
+    np.float64_t[:, ::1] buffer_weight,
+    const int Nx,
+    const int Ny,
+    const int Npix_min = 4,
+):
+    cdef np.ndarray[np.float64_t, ndim=1] center = np.array([0.5, 0.5, 0.5])
```
chrishavlin, line 168: why is this fixed rather than an argument?

cphyc: In the "stamping" approach, I compute the projection of a cube of size 1 centred on [0.5, 0.5, 0.5]. This is then "stamped" at the location of each cell. The centre here relates to the centre of this reference cube, not to where the centre actually is. This is done from the caller's side of things.

yt/utilities/lib/image_utilities.pyx
cphyc: I just checked, and it does not seem to be checking anything
```
 290: @cython.boundscheck(False)
 291: @cython.wraparound(False)
+292: cdef inline np.float64_t det2d(const np.float64_t[::1] a, const np.float64_t[::1] b) noexcept nogil:
static CYTHON_INLINE __pyx_t_5numpy_float64_t __pyx_f_2yt_9utilities_3lib_15image_utilities_det2d(__Pyx_memviewslice __pyx_v_a, __Pyx_memviewslice __pyx_v_b) {
  __pyx_t_5numpy_float64_t __pyx_r;
/* … */
  /* function exit code */
  __pyx_L0:;
  return __pyx_r;
}
+293:     return a[0] * b[1] - a[1] * b[0]
  __pyx_t_1 = 0;
  __pyx_t_2 = 1;
  __pyx_t_3 = 1;
  __pyx_t_4 = 0;
  __pyx_r = (((*((__pyx_t_5numpy_float64_t const  *) ( /* dim=0 */ ((char *) (((__pyx_t_5numpy_float64_t const  *) __pyx_v_a.data) + __pyx_t_1)) ))) * (*((__pyx_t_5numpy_float64_t const  *) ( /* dim=0 */ ((char *) (((__pyx_t_5numpy_float64_t const  *) __pyx_v_b.data) + __pyx_t_2)) )))) - ((*((__pyx_t_5numpy_float64_t const  *) ( /* dim=0 */ ((char *) (((__pyx_t_5numpy_float64_t const  *) __pyx_v_a.data) + __pyx_t_3)) ))) * (*((__pyx_t_5numpy_float64_t const  *) ( /* dim=0 */ ((char *) (((__pyx_t_5numpy_float64_t const  *) __pyx_v_b.data) + __pyx_t_4)) )))));
  goto __pyx_L0;
```

yt/utilities/lib/image_utilities.pyx
cphyc: done!

cphyc:
I have *finally* updated the code. Here are some examples

# Example 1: One of the galaxy in `output_00080`
```python
import yt

ds = yt.load("output_00080")
ad = ds.all_data()
center = ds.arr(ad.argmax("density"))
width = ds.quan(200, "kpc")
zlim = (6e-30, 2e-24)
data_source = ds.sphere(center, width * 2)

t0 = time()
p = yt.ProjectionPlot(
    ds,
    "x",
    "density",
    buff_size=(800, 800),
    weight_field="density",
    center=center,
    width=width,
    data_source=data_source
)
pref = p
p.set_zlim(("gas", "density"), pref.frb["gas", "density"].min(), pref.frb["gas", "density"].max())
p.save("/tmp/projection.png")

p = yt.OffAxisProjectionPlot(
    ds,
    [1, 0, 0],
    "density",
    north_vector=[0, 0, 1],
    buff_size=(800, 800),
    weight_field="density",
    center=center,
    width=width,
    data_source=data_source
)
p.set_zlim(("gas", "density"), pref.frb["gas", "density"].min(), pref.frb["gas", "density"].max())
p.set_axes_unit("Mpc")
p.save("/tmp/new.png")
```
| On axis projection | Off axis projection | Off axis projection (from [1, 2, 3]) |
|----|----|---|
| on-axis projection | off-axis projection | off-axis projection, different view |


# Example 2: plot of the z component
This is @chrishavlin example from https://github.com/yt-project/yt/pull/4741#issuecomment-1813024283.
| On axis projection | Off axis projection |
|----|----|
| old | new |

yt/visualization/volume_rendering/off_axis_projection.py
```diff
@@ -380,32 +383,90 @@ def temp_weightfield(field, data):
...
+            axis=-1,
+        )
+
+        for idim, periodic in enumerate(data_source.ds.periodicity):
+            if not periodic:
+                continue
+            # Wrap into [-0.5, +0.5]
+            xyz[..., idim] = (xyz[..., idim] + 0.5) % 1 - 0.5
+
+        dx = (data_source["index", "dx"] / wmax).to("1").d
```
chrishavlin, line 412: are octree grids always cubes? I naively expected to see some handling of `dy` and `dz` as well, but this all seems to assume dx=dy=dz.

chrishavlin (commented):
Looks great! Some minor comments that you can take or leave, will hit approve after an answer to my question about grid spacing :)

yt/visualization/volume_rendering/off_axis_projection.py
cphyc: I have added a comment in that sense. The octree grids are always cubic in the astronomical codes I know, but it might not be the general case. I've added a comment in that direction and an explicit check.

neutrinoceros:
I would have liked to review this but unfortunately I'm approaching review request bankruptcy so realistically it's not going to happen. I think it would be fine to merge it with Chris' approval. For release notes though, should this be in new features, perf improvements, minor API change, all three ?

cphyc:
I added labels (improvement and new feature). The API doesn't change, it's a drop-in replacement for the existing off-axis renderer.

neutrinoceros:
Let me rephrase that: is there any observable change besides improved performance ?

cphyc:
Yes; the method does not yield _exactly_ the same images as the previous implementation (although they should be extremely close).

chrishavlin:
I didn't mention it in my review, but while reviewing i did actually run some more comparison tests between old and new methods with a couple datasets and some different cutting planes. the differences were on the order of 1e-40 relative difference in all the cases I looked at. it wasn't exhaustive testing by any means, but I think it's pretty safe to say that it's unlikely for folks to observe any difference by eye but may see differences when comparing actual pixel values. Certainly worth giving a heads up in the release notes though!

neutrinoceros:
thank you both ! Feel free to merge away then !

PR #4742, opened 2023-11-16
ENH: Add threshold config for quiver plots

### LINKED ISSUE #4737: Threshold by a different field in quivers
It would be useful to apply thresholds to vector plots in yt where the threshold is supplied by another field.  For instance, if I am applying a quiver plot to wind velocity, I would like to be able to threshold based on the velocity magnitude.  This can currently be implemented with a combination of derived fields, such as:

```python
@yt.derived_field(name=("gas", "uinterp_thresh"), units="m/s", sampling_type="cell")
def uinterp_thresh(field, data):
    windmag = (data["cm1", "uinterp"]**2.0 + data["cm1", "vinterp"]**2.0)**0.5
    rv = data["cm1", "uinterp"].copy()
    rv[windmag < 20 * rv.uq] = np.nan
    return rv

@yt.derived_field(name=("gas", "vinterp_thresh"), units="m/s", sampling_type="cell")
def uinterp_thresh(field, data):
    windmag = (data["cm1", "uinterp"]**2.0 + data["cm1", "vinterp"]**2.0)**0.5
    rv = data["cm1", "vinterp"].copy()
    rv[windmag < 20 * rv.uq] = np.nan
    return rv
```

But, this is annoying.  And, it's not easy to change the threshold value (20 in this case) on-the-fly.  It would be good to have an argument to the quiver annotation call that included thresholding information, such as field and value (or expression).

matthewturk:
This adds a thresholding option for quiver plots.  Not currently supported in cutting plane.

Closes #4737.

### DIFF
--- a/yt/visualization/plot_modifications.py
+++ b/yt/visualization/plot_modifications.py
@@ -12,7 +12,7 @@ import numpy as np
 from unyt import unyt_quantity
 
 from yt._maintenance.deprecation import issue_deprecation_warning
-from yt._typing import AnyFieldKey, FieldKey
+from yt._typing import AnyFieldKey, FieldKey, Quantity
 from yt.data_objects.data_containers import YTDataContainer
 from yt.data_objects.level_sets.clump_handling import Clump
 from yt.data_objects.selection_objects.cut_region import YTCutRegion
@@ -444,6 +444,9 @@ class VelocityCallback(PlotCallback):
         scale_units=None,
         normalize=False,
         plot_args=None,
+        threshold_field: Optional[FieldKey] = None,
+        lower_threshold: Optional[Quantity] = None,
+        upper_threshold: Optional[Quantity] = None,
         **kwargs,
     ):
         self.factor = _validate_factor_tuple(factor)
@@ -462,6 +465,14 @@ class VelocityCallback(PlotCallback):
             plot_args = kwargs
 
         self.plot_args = plot_args
+
+        if None not in (lower_threshold, upper_threshold):
+            threshold_field = self.field
+        else:
+            threshold_field = None  # Need to specify at least one value
+        self.threshold_field = threshold_field
+        self.lower_threshold = lower_threshold
+        self.upper_threshold = upper_threshold
 
     def __call__(self, plot) -> "BaseQuiverCallback":
         ftype = plot.data._current_fluid_type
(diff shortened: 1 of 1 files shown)


yt/visualization/plot_modifications.py
```diff
@@ -463,6 +466,14 @@ def __init__(
 
         self.plot_args = plot_args
 
+        if None not in (lower_threshold, upper_threshold):
+            threshold_field = self.field
+        else:
+            threshold_field = None  # Need to specify at least one value
+        self.threshold_field = threshold_field
```
chrishavlin, line 473: i think I need more coffee this morning... I'm really struggling with mentally parsing the `if None not in (lower_threshold, upper_threshold)` line. But is the aim of this block is to set the `threshold_field` to `None` if both  `lower_threshold` and `upper_threshold` are `None`?

matthewturk: This line is wrong!  Thank you!

yt/visualization/plot_modifications.py
```diff
@@ -721,6 +738,13 @@ def __init__(
             plot_args.update(kwargs)
 
         self.plot_args = plot_args
+        if upper_threshold is not None or lower_threshold is not None:
+            threshold_field = threshold_field or self.field
+        elif upper_threshold is None and lower_threshold is None:
+            threshold_field = None  # Need to specify at least one value
```
chrishavlin, line 744: for what it's worth, I find this version much more readable :)

matthewturk: Sadly it is also incorrect!  :(

matthewturk:
I just pushed a new version that fixes the logic errors.  Originally I wrote it into the contour callback, which has a `field` value -- but the quiver callback does not.  And the logic should be, "if you specify a threshold field, but don't specify either an upper or lower, then what's the point, skip thresholding."  I think this is now in there correctly.

yt/visualization/plot_modifications.py
```diff
@@ -869,19 +897,46 @@ def _transformed_field(field, data):
...
+            if self.field_c == self.threshold_field:
+                pixC = pixT
+            else:
+                pixC = plot.data.ds.coordinates.pixelize(
+                    plot.data.axis,
+                    plot.data,
+                    self.field_c,
+                    bounds,
+                    (nx, ny),
+                    False,  # antialias
```
neutrinoceros, line 929: why not pass the argument as keyword ?

matthewturk: ?

Issue #4750, opened 2023-11-25
off-axis slice in spherical coordinates


fuwentai:
I think yt is absolutely awesome and has saved me a lot of time. But I've noticed that when working with data in spherical coordinates, the existing slice functionality only offers slicing directions along 'r', 'theta', and 'phi'. However, these options do not fulfill many requirements in a spherical coordinate system. I am wondering if it's possible to implement a feature where I can define slices like off-axis in Cartesian coordinates even though the data is in spherical coordinates?

neutrinoceros:
Hi. I think there are two different questions to be answered here:
1) would such an addition be welcome in the code base ?

absolutely ! The way I see it, neither off-axis slices nor support for spherical geometries is complete while this is missing.

2) would historical contributors be willing to implement it ?

That's more difficult to say. Both features have been in the code base for years at this point and as far as I recall no one has even attempted to implement this. It'd require entirely new code so it's not something we can add incrementally from what's already there. It's also possible that it would require substantial refactors of the existing, cartesian-only, off-axis internals. It's hard to say for sure, but that's my intuition of the problem.

In short: I would absolutely love to see someone championing this feature, but that's a substantial undertaking.

cphyc:
Coincidentally, it might be possible to achieve what you're looking for relatively easily™ thanks to #4741, but this PR would need to be fixed in the first place.

If we approximate each cell by a parallelepiped, the aforementioned PR provides machinery to compute what each one would look like from an arbitrary position. If we then iterate over all parallelepiped, we have the off-axis projection. This could work in the general case for any dataset that is tiled with parallelepiped.

chrishavlin:
I'd also find this hugely helpful.... but don't think I have the time right now to work on it in earnest. 

And for what it's worth, I've spent some time exploring ways to do this with yt as-is and depending on your needs, there are a couple approaches that can work OK:

1. embed a spherical dataset within a cartesian dataset using the Stream with callables. This essentially relies on the outer cartesian dataset to act as a pixelizer. I've been working on using this approach to handle some more general transformations with yt_xarray to avoid pre-interpolation, [here's a sample notebook](https://github.com/chrishavlin/yt_scratch/blob/master/notebooks/geodata/on_the_fly_transformations.ipynb) that defines data in geocentric coordinates and loads it within a cartesian dataset   (and [here's what it looks like in yt_idv](https://youtube.com/shorts/U4FvI_S560Y)). 
[...]

PR #4752, opened 2023-11-28
[WIP] Naive off-axis slice attempt for spherical-ish geometry

matthewturk:
This is an attempt at addressing #4750.  It works alright for offsets in theta, which seem to just rotate, but I'm confusing myself when attempting to test it for having a new normal vector that is set at some declination.

I think I've also confused myself yet again about how the r, theta, phi coordinates work, as my test script is not producing the types of plots I'd expect.  When slicing along the dimension that runs 0 .. pi, I'd expect to get a whole circle, and when slicing along 0 .. 2pi I'd expect half circle.  But that's not what's happening.

Here's an example script:

```python
import numpy as np
import yt, yt.testing

ds = yt.testing.fake_amr_ds(geometry = "spherical")

for i, phi_offset in enumerate(np.mgrid[ds.domain_left_edge[2]:ds.domain_right_edge[2]:100j]):
    s = ds.slice("theta", ds.domain_center[1], offset = [0.0, 0.0, phi_offset])
    s.to_pw("Density").save(f"frame_{i:04d}")
    print(i, phi_offset)
```

### DIFF
--- a/yt/data_objects/selection_objects/slices.py
+++ b/yt/data_objects/selection_objects/slices.py
@@ -15,6 +15,7 @@ from yt.funcs import (
     validate_object,
     validate_width_tuple,
 )
+from yt.geometry.api import Geometry
 from yt.utilities.exceptions import YTNotInsideNotebook
 from yt.utilities.minimal_representation import MinimalSliceData
 from yt.utilities.orientation import Orientation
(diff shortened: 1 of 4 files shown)


matthewturk:
There are lots of problems with this! I am rethinking it but it was absolutely helpful to sketch it out.

For starters, it's incorrect - an offset isn't that same as a new north vector. But I do see a path forward here.

matthewturk:
Moving to draft, and I think I've found the right math now and will revisit.

matthewturk:
This is going to stay WIP, but I've been poking at it a bit and will try to come back to it very soon.

PR #4758, opened 2023-12-12
ENH: allow boxlib frontend to read 1D cylindrical datasets

yut23:
## PR Summary

Removes the check that disallows cylindrical coordinates in 1D. It looks like yt learned how to support these sometime since this code was added in 2014. It also fixes an error in `_parse_header_file()` introduced by #4641.

I tested against an example dataset output by https://github.com/AMReX-Astro/Castro/tree/main/Exec/hydro_tests/Sedov using `inputs.1d.cyl`, and the cell volumes checked out (proportional to 1/r).

## PR Checklist

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -374,10 +374,8 @@ class BoxlibHierarchy(GridIndex):
             default_ybounds = (0.0, 1.0)
             default_zbounds = (0.0, 1.0)
         elif self.ds.geometry == "cylindrical":
-            # Now we check for dimensionality issues
-            if self.dimensionality != 2:
-                raise RuntimeError("yt needs cylindrical to be 2D")
             self.level_dds[:, 2] = 2 * np.pi
+            default_ybounds = (0.0, 1.0)
             default_zbounds = (0.0, 2 * np.pi)
         elif self.ds.geometry == "spherical":
             # BoxLib only supports 1D spherical, so ensure
@@ -892,7 +890,7 @@ class BoxlibDataset(Dataset):
             self.geometry = Geometry(geom_str)
 
         if self.geometry == "cylindrical":
-            dre = self.domain_right_edge
+            dre = self.domain_right_edge.copy()
             dre[2] = 2.0 * np.pi
             self.domain_right_edge = dre
 
--- a/yt/frontends/boxlib/tests/test_outputs.py
+++ b/yt/frontends/boxlib/tests/test_outputs.py
@@ -321,6 +321,11 @@ def test_CastroDataset_2():
     assert isinstance(data_dir_load("castro_sod_x_plt00036"), CastroDataset)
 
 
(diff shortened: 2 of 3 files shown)


yt/frontends/boxlib/data_structures.py
```diff
@@ -892,7 +890,7 @@ def _parse_header_file(self):
             self.geometry = Geometry(geom_str)
 
         if self.geometry == "cylindrical":
-            dre = self.domain_right_edge
+            dre = self.domain_right_edge.copy()
             dre[2] = 2.0 * np.pi
             self.domain_right_edge = dre
```
neutrinoceros, line 895: I don't think I get this change, or why this simple operation is done over 3 lines when a single would suffice ?

neutrinoceros:
I'm triaging this as a bug fix for now because it removes a blocking problem to anyone who would need the functionality, and the change is currently very small, but let's re-assess once we see test go green again.

neutrinoceros:
Actually, failures are clearly unrelated.

neutrinoceros:
For what it's worth I've issued a fix for those failures at https://github.com/yt-project/yt/pull/4759

yt/frontends/boxlib/data_structures.py
yut23: `self.domain_right_edge` is set to be read-only since #4641, so we can't assign to `self.domain_right_edge[2]` directly. `dre` is a mutable copy, so we can modify it.

neutrinoceros: Oh cool, I forgot about this. Thanks !

neutrinoceros (approved):

neutrinoceros:
@yut23 can you rebase to current main so we can get CI green ?

yut23:
It might make more sense to split out the `_parse_header_file()` fix into a separate PR, since that's an actual bug fix, while the rest is an enhancement.

yut23:
We should probably add a sample data set to test with. I'm not sure what the process for that is, though.

neutrinoceros:
You're right. And storage basically free for a 1D dataset so let's do it.
You can use `yt upload` on the command line to share the data sample. Then open a PR to the `yt-project/website` repo (see previously merged ones for an example). You will also need to update the data sample yaml registry in this repo (you may do it within this PR).

cphyc:
@yt-fido test this please.

cphyc (dismissed):
This looks good to me :)

neutrinoceros (changes requested):
there appears to be a problem with the new data sample:

```python
import yt

ds = yt.load_sample("castro_sedov_1d_cyl_plt00150")
ds = yt.load_sample("castro_sedov_1d_cyl_plt00150")
```
fails with `FileNotFoundError` (the exception is raised on the first call if data is already installed, and on the second otherwise). I'm looking into it.

neutrinoceros (approved):

PR #4761, opened 2023-12-14
ENH: Add a TwoSlopeNorm to the norm_handler

matthewturk:
I think that the code works, but I need suggestions on how to more easily expose this behavior.  I'd also like to see how we could potentially describe setting the min/max to be mirrors; for instance, right now the ramp on either side of vcenter changes, but it would be nice to set it to the maximum extent.  (i.e., vmax = -vmin or something).

### DIFF
--- a/yt/visualization/_handlers.py
+++ b/yt/visualization/_handlers.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 import matplotlib as mpl
 import numpy as np
 import unyt as un
-from matplotlib.colors import Colormap, LogNorm, Normalize, SymLogNorm
+from matplotlib.colors import Colormap, LogNorm, Normalize, SymLogNorm, TwoSlopeNorm
 from unyt import unyt_quantity
 
 from yt._typing import Quantity, Unit
@@ -61,6 +61,7 @@ class NormHandler:
         "_dynamic_range",
         "_norm_type",
         "_linthresh",
+        "_vcenter",
         "_norm_type",
         "_norm",
         "prefer_log",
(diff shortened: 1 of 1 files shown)


matthewturk:
I'd also like suggestions on how to expose this in the field plugins

yt/visualization/_handlers.py
```diff
@@ -61,13 +61,15 @@ class NormHandler:
         "_dynamic_range",
         "_norm_type",
         "_linthresh",
+        "_vcenter",
         "_norm_type",
```
chrishavlin, line 65: you didn't introduce this, but `"_norm_type"` is in the `__slots__` declaration twice... could remove one here?

neutrinoceros:
> I need suggestions on how to more easily expose this behavior

I'll quote myself from the Slack conversation

> How about having `set_norm` accept a str as its `norm` argument (currently this must be a `Normalize` object), and arbitrary keyword arguments `**kwargs` passed down to the actual `Normalize` subclass underneath ?
> We could add support for a couple Normalize subclasses that are already in matplotlib (like `"twoslopes" -> TwoSlopeNorm` and `"asinh" -> AsinhNorm` ). It would also be trivial to add more in the future, and the `NormHandler` class wouldn't need to be changed at all (I think).

chrishavlin:
+1 for expanding `set_norm` functionality! I think that's a nice compromise that leaves a clear way forward for future development.

matthewturk:
I can get behind the string key!

Shall we put this in (once y'all are OK with it) and I'll work on the string key in a new PR?

matthewturk:
So I guess the consensus is, no, this needs to all happen at once? :)

yt/visualization/_handlers.py
matthewturk: Good catch!  I removed it.

neutrinoceros:
I think so ?

PR #4762, opened 2023-12-19
ENH: add check for finite values in extrema

chrishavlin:
## PR Summary

This PR adds a `check_finite` kwarg to `quantities.extrema` so that you can find the extrema of fields that include `np.nan` values.  Default value is `False` so that default behavior does not change. 

## PR Checklist

### DIFF
--- a/yt/data_objects/derived_quantities.py
+++ b/yt/data_objects/derived_quantities.py
@@ -578,6 +578,9 @@ class Extrema(DerivedQuantity):
     non_zero : bool
         If True, only positive values are considered in the calculation.
         Default: False
+    check_finite : bool
+        If True, non-finite values will be explicitly exluded.
+        Default: False
 
     Examples
     --------
@@ -588,23 +591,25 @@ class Extrema(DerivedQuantity):
 
     """
 
-    def count_values(self, fields, non_zero):
+    def count_values(self, fields, non_zero, check_finite):
         self.num_vals = len(fields) * 2
 
-    def __call__(self, fields, non_zero=False):
+    def __call__(self, fields, non_zero=False, *, check_finite=False):
         fields = list(iter_fields(fields))
-        rv = super().__call__(fields, non_zero)
+        rv = super().__call__(fields, non_zero, check_finite)
         if len(rv) == 1:
             rv = rv[0]
         return rv
 
-    def process_chunk(self, data, fields, non_zero):
+    def process_chunk(self, data, fields, non_zero, check_finite):
         vals = []
         for field in fields:
             field = data._determine_fields(field)[0]
             fd = data[field]
             if non_zero:
                 fd = fd[fd > 0.0]
+            if check_finite:
+                fd = fd[np.isfinite(fd)]
             if fd.size > 0:
                 vals += [fd.min(), fd.max()]
             else:
--- a/yt/data_objects/derived_quantities.py
+++ b/yt/data_objects/derived_quantities.py
@@ -579,7 +579,7 @@ class Extrema(DerivedQuantity):
         If True, only positive values are considered in the calculation.
         Default: False
     check_finite : bool
-        If True, non-finite values will be explicitly exluded.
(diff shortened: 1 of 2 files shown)


yt/data_objects/derived_quantities.py
```diff
@@ -588,23 +591,25 @@ class Extrema(DerivedQuantity):
 
     """
 
-    def count_values(self, fields, non_zero):
+    def count_values(self, fields, non_zero, check_finite):
```
neutrinoceros, line 594: given the wild diversity that already exist in the various implementations of `DerivedQuantity.count_values` and how common boolean positionally-allowed arguments are, let's dodge this time :)
```suggestion
    def count_values(self, fields, non_zero, *, check_finite):
```

chrishavlin: ya! agreed. changing just this line causes an error because down below in ``__call__`` I currently have it passing along `check_finite` as an arg rather than kwarg: 

```python
rv = super().__call__(fields, non_zero, check_finite)
```
So I'll push up a change with your suggestion and the other line fixed as well (after my meeting that starts in a minute....)

yt/data_objects/derived_quantities.py
```diff
@@ -588,23 +591,25 @@ class Extrema(DerivedQuantity):
...
-    def process_chunk(self, data, fields, non_zero):
+    def process_chunk(self, data, fields, non_zero, *, check_finite=False):
         vals = []
         for field in fields:
             field = data._determine_fields(field)[0]
             fd = data[field]
             if non_zero:
                 fd = fd[fd > 0.0]
+            if check_finite:
+                fd = fd[np.isfinite(fd)]
```
neutrinoceros, line 612: Can't we use `np.nanmin` and `np.nanmax` directly ? Actually I wonder if it's worth changing the interface at all, I would be inclined to consider the current behavior buggy.

matthewturk: If we can distinguish between inf and nan, I might see the need to have this new behavior.  On the other hand, I'm also sort of inclined toward thinking the behavior is buggy.

neutrinoceros: Oh I actually didn't think of `inf`, but I note that `np.nanmin` only excludes `nan`. For instance

```python
>>> np.nanmin([-np.inf, 0, np.nan])
-inf
```
from the title and description, this doesn't seem like what @chrishavlin wants, so we'd still need the proposed opt-in masking to exclude infinities, which means the bug we're discussing should be dealt with in a separate PR (so the fix can be back ported).

neutrinoceros: I opened #4763 to deal with this

chrishavlin: So I actually thought a bit about that question... I should have typed up a summary in my initial PR submission here :) 

TLDR: I like to know when there are NaN's, so I wanted to keep the default behavior and opt-in to filtering NaN's (or infs).

So first, more of a philosophical point: I think there is some value in having the default behavior return NaN's if they are present. I often use `extrema` when I initially load up a new dataset and want to get a sense for various fields and their values and knowing if there are NaN's is a useful indicator of a few things: for observational data it usually (but not always) indicates that there are simply missing points and it's usually appropriate to switch over to using `nanmin/max`. But for simulation output it can be an indicator of a problem with the simulation and always using `nanmin/max` might mask that problem.  

Second -- for how I implemented this PR using `~isfinite()` to mask values, there'd be a small hit on memory usage. Since most folks working with yt are loading simulation data, which ideally would not have NaN's (maybe that's wrong), I didn't want to add that extra mask allocation if it's not needed. Maybe this isn't a really a concern since these methods run on individual chunks... and If we switched it to always using `nanmin/max` then this would no longer be the case.

matthewturk:
I think we might have reached a shared consensus on this?

neutrinoceros:
Please do not hold this off over my ill informed comment :)

matthewturk (approved):

PR #4764, opened 2023-12-20
BUGFIX: Enzo testing framework - ignore units for ShockTube tests

clairekope:
## PR Summary

This bug was found while [fixing the Enzo CI testing](https://github.com/enzo-project/enzo-dev/pull/223). It seems to have been induced by the upgrade to unyt v3. The bug is as follows: 

The `ShockTubeTest` class supports comparing simulation results to an exact solution saved to file; however, the solution in that file doesn't have any units associated with it. When trying to compare solutions, unyt complains. 

This is a very simple change that removes units from the simulation data before comparing. 
The test for it is essentially the Enzo test infrastructure over at https://github.com/enzo-project/enzo-dev? So I'm not sure how I would add a test for this within yt.

### DIFF
--- a/yt/frontends/enzo/answer_testing_support.py
+++ b/yt/frontends/enzo/answer_testing_support.py
@@ -104,7 +104,7 @@ class ShockTubeTest:
             field = ad[k].d
             for xmin, xmax in zip(self.left_edges, self.right_edges):
                 mask = (position >= xmin) * (position <= xmax)
-                exact_field = np.interp(position[mask].value, exact["pos"], exact[k])
+                exact_field = np.interp(position[mask].ndview, exact["pos"], exact[k])
                 myname = f"ShockTubeTest_{k}"
                 # yield test vs analytical solution
                 yield AssertWrapper(
--- a/yt/frontends/enzo/answer_testing_support.py
+++ b/yt/frontends/enzo/answer_testing_support.py
@@ -104,7 +104,7 @@ class ShockTubeTest:
             field = ad[k].d
             for xmin, xmax in zip(self.left_edges, self.right_edges):
                 mask = (position >= xmin) * (position <= xmax)
-                exact_field = np.interp(position[mask], exact["pos"], exact[k])
+                exact_field = np.interp(position[mask].value, exact["pos"], exact[k])
                 myname = f"ShockTubeTest_{k}"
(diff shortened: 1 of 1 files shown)


neutrinoceros:
If this test code isn't used in yt itself we probably ought to remove it at some point, so if Enzo depends on it I think the code should be migrated and maintained there, otherwise we have no means to guarantee its long term stability and it's a risk to anyone that depends on it. That said, the fix is trivial and doesn't affect us so we might as well ship it for now.

@clairekope our current plan is to wait for numpy 2.0.0rc to come out before we cut a new bug fix release. Currently this means that this patch won't be part of a public release for another month (give or take), is that okay for you ?

yt/frontends/enzo/answer_testing_support.py
```diff
@@ -104,7 +104,7 @@ def __call__(self):
             field = ad[k].d
             for xmin, xmax in zip(self.left_edges, self.right_edges):
                 mask = (position >= xmin) * (position <= xmax)
-                exact_field = np.interp(position[mask], exact["pos"], exact[k])
+                exact_field = np.interp(position[mask].value, exact["pos"], exact[k])
```
neutrinoceros, line 107: note that `.value` returns a copy of the underlying data, which is wasteful here. Using a view instead will reduce memory usage
```suggestion
                exact_field = np.interp(position[mask].ndview, exact["pos"], exact[k])
```

neutrinoceros (approved):

clairekope:
> If this test code isn't used in yt itself we probably ought to remove it at some point, so if Enzo depends on it I think the code should be migrated and maintained there, otherwise we have no means to guarantee its long term stability and it's a risk to anyone that depends on it. That said, the fix is trivial and doesn't affect us so we might as well ship it for now.

I don't disagree; it's part of yt's legacy as a primarily Enzo toolchain. It's something I can try to bring up with the Enzo dev community. Do other frontends have such specialized testing tools hosted in yt?

> @clairekope our current plan is to wait for numpy 2.0.0rc to come out before we cut a new bug fix release. Currently this means that this patch won't be part of a public release for another month (give or take), is that okay for you ?

Not a problem; we build from the tip of the git repo for CI testing.

neutrinoceros:
@yt-fido test this please

Issue #4766, opened 2023-12-21
Problems with the particle_trajectories documentation/module (2/2)


mtryan83:
### Bug report

**Bug summary**

Assuming we successfully run the example code for the [particle_trajectories](https://yt-project.org/doc/reference/api/yt.data_objects.particle_trajectories.html#yt.data_objects.particle_trajectories.ParticleTrajectories) module (e.g. using the workaround mentioned in #4765, ie `("all","particle_position_x")` --> `("io","particle_position_x")`), getting a trajectory by index fails due to mixing `str`s and `tuple`s in a call to `sorted(self.field_data.keys())`

**Code for reproduction**

```python
import yt
fields = [ ("all", "particle_position_x"),] # simplified for brevity
ts = yt.load("orbit_hdf5_chk_000?")
ds = ts[0]
init_sphere = ds.sphere(ds.domain_center, (0.5, "unitary"))
indices = init_sphere[("all", "particle_index")].astype("int64")
trajs = ts.particle_trajectories(indices, fields=fields, suppress_logging=True)
trajs.trajectory_from_index(1)
```

**Actual outcome**

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "~/.conda/envs/test_yt/lib/python3.11/site-packages/yt/data_objects/particle_trajectories.py", line 340, in trajectory_from_index
    fields = sorted(self.field_data.keys())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'tuple' and 'str'
```

**Expected outcome**

A dictionary with particle trajectory information. Something like 
```python
{'particle_time': unyt_array([  0., 100., 200., 300., 400., 500., 600., 700., 800., 900.], 'code_time'), 'particle_index': 1, 'particle_position_x': unyt_array([0.25      , 0.25131274, 0.25524942, 0.26175659, 0.27075597,
            0.28215068, 0.2958247 , 0.31164457, 0.32945507, 0.3491025 ], 'code_length'), 'particle_position_y': unyt_array([0.5       , 0.47421621, 0.44873755, 0.42384795, 0.39982133,
            0.37690862, 0.35532256, 0.3353068 , 0.31707255, 0.30078343], 'code_length'), 'particle_position_z': unyt_array([0.5       , 0.5       , 0.5       , 0.5       , 0.50000001,
[...]

Issue #4590, opened 2023-07-21
doc: example Particle Trajectories analysis fails


Xarthisius:
### Bug report

**Bug summary**

See: https://yt-project.org/doc/analyzing/particle_trajectories.html
 
**Code for reproduction**

```python
import glob
from os.path import join

import yt
from yt.config import ytcfg

path = ytcfg.get("yt", "test_data_dir")
my_fns = glob.glob(join(path, "Orbit", "orbit_hdf5_chk_00[0-9][0-9]"))
my_fns.sort()
fields = ["particle_velocity_x", "particle_velocity_y", "particle_velocity_z"]
ds = yt.load(my_fns[0])
dd = ds.all_data()
indices = dd["all", "particle_index"].astype("int")
ts = yt.DatasetSeries(my_fns)
trajs = ts.particle_trajectories(indices, fields=fields, suppress_logging=True)
```

**Actual outcome**

```
Generating [particle_velocity_x, particle_velocity_y, particle_velocity_z] fields in trajectories:   0%| | 0/31 [00:00<Traceback (most recent call last):
  File "/home/xarth/yt/traj/ala.py", line 19, in <module>
    trajs = ts.particle_trajectories(indices, fields=fields, suppress_logging=True)
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/time_series.py", line 434, in particle_trajectories
    return ParticleTrajectories(
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/particle_trajectories.py", line 147, in __init__
    self._get_data(fields)
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/particle_trajectories.py", line 278, in _get_data
    cube = grid.retrieve_ghost_zones(1, grid_fields)
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/index_subobjects/grid_patch.py", line 265, in retrieve_ghost_zones
    cube = self.ds.covering_grid(
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/construction_data_containers.py", line 723, in __init__
    self.get_data(fields)
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/construction_data_containers.py", line 910, in get_data
    self[a] = self._data_source[f]
  File "/home/xarth/codes/xarthisius/yt/yt/data_objects/data_containers.py", line 235, in __getitem__
    self.get_data(f)
[...]

Issue #3734, opened 2021-12-28
TST: Remove yield tests


matthewturk:
To finish migrating to `pytest` we will need to eliminate all `yield`-tests.

Here is a list of places where `yield` shows up in a `test_*.py` file; not all are actual test files.

```
./tests/test_install_script.py
./doc/source/cookbook/tests/test_cookbook.py
./yt/tests/test_load_sample.py
./yt/tests/test_load_errors.py
./yt/visualization/tests/test_mesh_slices.py
./yt/visualization/tests/test_save.py
./yt/visualization/tests/test_callbacks.py
./yt/visualization/tests/test_particle_plot.py
./yt/visualization/tests/test_geo_projections.py
./yt/visualization/tests/test_plotwindow.py
./yt/visualization/tests/test_raw_field_slices.py
./yt/visualization/tests/test_line_plots.py
./yt/visualization/tests/test_profile_plots.py
./yt/visualization/volume_rendering/tests/test_vr_orientation.py
./yt/visualization/volume_rendering/tests/test_mesh_render.py
./yt/fields/tests/test_sph_fields.py
./yt/fields/tests/test_xray_fields.py
./yt/fields/tests/test_fields.py
./yt/frontends/owls_subfind/tests/test_outputs.py
./yt/frontends/ramses/tests/test_outputs_pytest.py
./yt/frontends/ramses/tests/test_outputs.py
./yt/frontends/arepo/tests/test_outputs.py
./yt/frontends/fits/tests/test_outputs.py
./yt/frontends/gdf/tests/test_outputs_nose.py
./yt/frontends/ahf/tests/test_outputs.py
./yt/frontends/gadget_fof/tests/test_outputs.py
./yt/frontends/gizmo/tests/test_outputs.py
./yt/frontends/owls/tests/test_outputs.py
./yt/frontends/moab/tests/test_c5.py
./yt/frontends/tipsy/tests/test_outputs.py
./yt/frontends/rockstar/tests/test_outputs.py
./yt/frontends/chombo/tests/test_outputs.py
./yt/frontends/nc4_cm1/tests/test_outputs.py
./yt/frontends/enzo/tests/test_outputs.py
./yt/frontends/ytdata/tests/test_old_outputs.py
./yt/frontends/ytdata/tests/test_outputs.py
./yt/frontends/gamer/tests/test_outputs.py
./yt/frontends/athena_pp/tests/test_outputs.py
./yt/frontends/exodus_ii/tests/test_outputs.py
[...]

PR #4768, opened 2023-12-22
Switch to field tuples for default fields in particle_trajectories

mtryan83:
## PR Summary

This PR changes two things:

1. Default fields `'particle_position_x/y/z'` now added to field_data through `fds` proper name. Since `fds[field]` is created through `dd_first._determine_fields((self.ptype, field))[0]`, this should ensure the field exists and is a tuple with the field type `ptype`. 
2. Returned trajectory dict now uses `(self.ptype,'particle_index')` and `(self.ptype,'particle_time')` instead of just `'particle_index'` and `'particle_time'`

Change 2 should ensure the documentation for `trajectory_from_index` actually is correct (it plots using `("all","particle_time") which would not match what was actually returned).

Change 1 should fix Issue #4766.

Change 2 should ensure the documentation for `trajectory_from_index` actually is correct (it plots using `("all","particle_time") which would not match what was actually returned).

## PR Checklist


mtryan83:
I'm currently working on adding/updating the test(s). I'll try to update the test file to a pytest at the same time, but it involves converting two `yield` tests (see #3734), which I'm unfamiliar with.

mtryan83:
Just comment: the `test_orbit_traj` and `test_etc_traj` don't seem to be running at all right now. They're excluded from the pytest framework and can't be found under the nose testing.
image
This seems to be true for a number of other tests too, like `test_pickling.test_particle_pickles`, so I don't know if it's intentional or not.

neutrinoceros:
Hi @mtryan83, thanks for investigating this. To answer your question, I don't think that these tests are skipped intentionally, but as we haven't been running any kind of coverage analysis for a while, it's not surprising.
For the record I think your patch is sound as is but I'll wait until you're satisfied with testing to merge.

mtryan83:
The tests probably aren't necessary, but I've already got them, so... 

I've included two tests, one to check that the default fields are set to tuples correctly and one to check that `particle_index` and `particle_time` are tuplized in the `trajectory_from_index` function. 

For each test I've included two versions, one designed for nose and one for pytest (currently commented out) that uses parameterize decorators and fixtures. Technically, the nose ones could also be used in pytest, they just don't use the fancy parts. I think it'd be better just to switch over when the changeover occurs.

I've kept test_particle_trajectories in the pytest ignore list for now, but it doesn't need to be. Since the two answer tests (`test_orbit_traj` and `test_etc_traj`) are still yield tests, pytest skips them. This seems ok to me, since they're currently broken anyway (per comment above and #4590).

neutrinoceros:
Personally I would prefer that we only add the pytest version of these tests. You can do this by making them part of a new module (keep the name, just add a `_pytest` suffix), and adding this module to ignore lists for nose. One of these lists lives in `tests/tests.yaml` and the other one is `nose_ignores.txt`

mtryan83:
I moved the `test_uniqueness` and `test_ptype` functions to `test_particle_trajectories_pytest.py` as well, since they work fine under pytest. This way the only nose tests are the two broken ones.

yt/data_objects/tests/test_particle_trajectories_pytest.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Select all dummy particles
+    print(ts[0].derived_field_list)
+    ids = ts[0].all_data()["dummy", "particle_index"]
+
+    # Build trajectories
+    ts.particle_trajectories(ids, ptype="dummy")
+
+
+@pytest.mark.parametrize("ptype", [None, ("io")])
```
neutrinoceros, line 123: The use of parentheses seems to in indicate you meant to use a tuple here, I think ? The issue is that it's the not parentheses that make a tuple, it's really the comma ! in other words, `("io") == "io"`.
```suggestion
@pytest.mark.parametrize("ptype", [None, "io"])
```

yt/data_objects/tests/test_particle_trajectories_pytest.py
```diff
@@ -0,0 +1,150 @@
...
+    ptype = ptype if ptype else "all"  # ptype defaults to "all"
+    for k in trajs.field_data.keys():
+        assert isinstance(k, tuple), f"Expected key to be tuple, received {type(k)}"
+        assert (
+            k[0] == ptype
+        ), f"Default field type ({k[0]}) does not match expected ({ptype})"
+        assert ("all", k[1]) in pfields, f"Unexpected field: {k[1]}"
+
+
+@pytest.mark.parametrize("ptype", [None, ("io")])
```
neutrinoceros, line 139:
```suggestion
@pytest.mark.parametrize("ptype", [None, "io"])
```

yt/data_objects/tests/test_particle_trajectories_pytest.py
```diff
@@ -0,0 +1,150 @@
...
+        fake_particle_ds(
+            fields=fields, negative=negative, units=units, npart=n_particles, data=data
+        )
+    ]
+    for ds in all_ds:
+        ds.add_particle_filter("dummy")
+    ts = DatasetSeries(all_ds)
+
+    # Select all dummy particles
+    print(ts[0].derived_field_list)
```
neutrinoceros, line 116: Not new, but let's clean this up while we're at it
```suggestion
```

neutrinoceros (approved):
some very minor points, otherwise LGTM, thank you !

jzuhone (approved):

PR #4776, opened 2024-01-01
DOC: fix a broken docs example

### LINKED ISSUE #4774: BUG: modifications through matplotlib engine cannot be properly displayed
### Bug report

**Bug summary**

The Modifications through Matplotlib engine cannot be properly displayed. 

Taking the following code for example, the expected modifications can only be shown by the containing matplotlib figure object like `fig.savefig("sloshing.png")`. 

**Code for reproduction**

adapted from [docs](https://yt-project.org/docs/dev/cookbook/simple_plots.html#accessing-and-modifying-plots-directly) (also broken there)

```python
import numpy as np

import yt

# Load the dataset.
ds = yt.load("GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150")

# Create a slice object
slc = yt.SlicePlot(ds, "x", ("gas", "density"), width=(800.0, "kpc"))

# Get a reference to the matplotlib axes object for the plot
ax = slc.plots[("gas", "density")].axes

# Let's adjust the x axis tick labels
for label in ax.xaxis.get_ticklabels():
    label.set_color("red")
    label.set_fontsize(16)

# Get a reference to the matplotlib figure object for the plot
fig = slc.plots[("gas", "density")].figure

# And create a mini-panel of a gaussian histogram inside the plot
rect = (0.2, 0.2, 0.2, 0.2)
new_ax = fig.add_axes(rect)

n, bins, patches = new_ax.hist(
    np.random.randn(1000) + 20, 50, facecolor="black", edgecolor="black"
)

# Make sure its visible
new_ax.tick_params(colors="white")

# And label it
la = new_ax.set_xlabel("Dinosaurs per furlong")
la.set_color("white")

slc.save()
```

**Actual outcome**

iTerm2 1b3woO sloshing_nomag2_hdf5_plt_cnt_0150_Slice_x_density

**Expected outcome**

The changes of the x-axis tick labels
iTerm2 h1XCIV sloshing

**Version Information**
  * Operating System: MacOS 14.1.1 and Red Hat Enterprise Linux Server release 7.8 (Maipo)
  * Python Version: 3.9
  * yt version: 4.2.1 and 4.3.0

neutrinoceros:
Close #4774

### DIFF
--- a/doc/source/cookbook/simple_slice_matplotlib_example.py
+++ b/doc/source/cookbook/simple_slice_matplotlib_example.py
@@ -8,6 +8,10 @@ ds = yt.load("GasSloshing/sloshing_nomag2_hdf5_plt_cnt_0150")
 # Create a slice object
 slc = yt.SlicePlot(ds, "x", ("gas", "density"), width=(800.0, "kpc"))
 
+# Rendering should be performed explicitly *before* any modification is
+# performed directly with matplotlib.
+slc.render()
+
 # Get a reference to the matplotlib axes object for the plot
 ax = slc.plots[("gas", "density")].axes
 


jzuhone (approved):

xshaokun:
I think it's necessary to emphasize explicitly where to use `.render()` in the docs as you explained in #4774.

neutrinoceros:
You're right. In the context of this example, do you think a comment on the line I added would be sufficient ?

xshaokun:
I'm not sure if it's obvious enough, but it would be better than now :)

neutrinoceros:
I'd like to keep the additional context minimalist to reduce the amount of churn and rotting docs if we can ever find a way to make it unneeded so I'll stick with a simple comment for now.

chrishavlin (approved):

Issue #4785, opened 2024-01-23
TST: `yt/visualization/volume_rendering/tests/test_scene.py::test_annotations` is failing in CI


neutrinoceros:
### Bug report
The following test is currently failing in CI, however I couldn't reproduce the error locally `yt/visualization/volume_rendering/tests/test_scene.py::test_annotations`

see https://github.com/yt-project/yt/actions/runs/7626456177/job/20772882827

neutrinoceros:
Notes:
- the test still passes in minimal env tests
- the fact that we haven't been able to reproduce it locally (by running *just* this test) may be indicative of test pollution, which could be checked by running only this test in CI too.

chrishavlin:
> the fact that we haven't been able to reproduce it locally (by running just this test) may be indicative of test pollution, which could be checked by running only this test in CI too.

I just confirmed this over in #4786:

* running just `pytest yt/visualization/volume_rendering/tests/test_scene.py::test_annotations` passed in CI
* expanding to run all the visualization tests, `pytest yt/visualization/`, passed in CI

neutrinoceros:
> might also be worth doing a run that randomizes test order?

Ideally yes, though I suspect that would be very unstable if we did it now because pollution in our test is possibly high, and it's likely that some of them are only passing because of other tests' side effects (I've put a lid on this for a while ...).

chrishavlin:
ok, well i've narrowed down the subset of tests that you need to run to get the failure: just running the tests in `yt/data_objects/tests` and the failing test:

```
pytest --color=yes yt/data_objects/tests yt/visualization/volume_rendering/tests/test_scene.py::test_annotations
```

no indication of **why** yet. Also, when re-running the failed test, it does pass on second run.

chrishavlin:
very close to an answer! I used the https://github.com/asottile/detect-test-pollution pytest plugin to find that `yt/data_objects/tests/test_firefly.py::test_firefly_JSON_string` is the source of the issue. Running just these two tests you can reproduce the failure:

```
        pytest --color=yes  \
        yt/data_objects/tests/test_firefly.py::test_firefly_JSON_string \
        yt/visualization/volume_rendering/tests/test_scene.py
```

**when reversing the order, tests pass no problem.** Looking at `test_firefly_JSON_string`, there's nothing obvious... but those firefly reader objects seem to do a lot with writing to disk and symlinks and default data directories so maybe there's something there?

neutrinoceros:
[...]

PR #4784, opened 2024-01-23
TST: ignore deprecation warning from pandas (Pyarrow will become a required dependency)

neutrinoceros:
## PR Summary

As noted by @jzuhone in #4782, this warning makes half of CI red.

I don't think we care about it, but for completion, here it is in full:
```
DeprecationWarning: 
Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0),
(to allow more performant data types, such as the Arrow string type, and better interoperability with other libraries)
but was not found to be installed on your system.
If this would cause problems for you,
please provide us feedback at https://github.com/pandas-dev/pandas/issues/54466
```

PR #4786, opened 2024-01-23
[WIP] investigating failing annotation test

chrishavlin:
Looking into #4785 

I wasn't able to reproduce the failure locally, so i'm adding an image test in this PR to be able to check out the image artifact to see if that helps...


chrishavlin:
rebased off #4784

chrishavlin:
the artifact image produced by CI matches what I get locally, so no insight there... latest push adjusts the existing test_annotation unit test to also check the image size before the annotation call.

chrishavlin:
latest push disables the text annotation in the unit test (to isolate the text addition from the colorbar addition) and makes the text in the image comparison test more obvious.

chrishavlin:
well i've got to leave this for a while. no real progress other than ruling out the text overlay. maybe something odd is going on when the colorbar axis is added?

chrishavlin:
latest push adjusts the test workflow to only run the failing test (obviously don't merge this...)

chrishavlin:
summary so far: running just the failed test in CI or the `yt/visualization` subset of tests passes in 3.12

chrishavlin:
latest push strips down the test action to just run 3.12 on ubuntu (just trying to speed up the process of isolating the issue and lessen the load on github actions while I'm iterating here)

chrishavlin:
oh interesting. now that i'm properly catching the failure, `test_annotation` does in fact pass when re-run after failing in the first call (with `pytest --lf`).

chrishavlin:
latest push removes some of the slower to install deps (ones that require wheels to build for 3.12), this will slightly change the test suite, but probably will still fail... after which I'm going to start a bit more trial and error with different test subsets (unless anyone else has a better idea of how to isolate the issue! )

chrishavlin:
oh hey, should try this pytest plugin... https://github.com/asottile/detect-test-pollution I'll keep going with my manual bisecting with the top level yt directories first to cut down on possibilities.

chrishavlin:
Summary of this morning's manual bisecting:

```
pytest -v yt/data_objects yt/fields yt/frontends yt/geometry yt/sample_data yt/tests yt/units yt/utilities yt/visualization

	(full test suite) fails 


pytest -v yt/utilities yt/visualization

	passes

pytest --color=yes yt/sample_data yt/tests yt/units yt/utilities yt/visualization

	passes
	
pytest --color=yes yt/fields yt/frontends yt/geometry yt/sample_data yt/tests yt/units yt/utilities yt/visualization
	
	passes
	
pytest -v yt/data_objects yt/visualization

	fails!
```

chrishavlin:
looking at what has changed recently in data_objects... my extrema-with-nans PR (#4762) **did** just get merged a few days ago. so latest push here reverts those changes just to check (that PR did pass tests prior to merging, but it also touches parallel objects...).

chrishavlin:
unrelated to #4762 , reverted that. latest push further bisects test suite to just data_object/tests

chrishavlin:
just moved a copy of the failing test to the `yt/data_objects/test` directory so i can try to use the detect-test-pollution plugin more easily (without running the entire test suite)

chrishavlin:
oh interesting! detect-test-pollution returns:


```
discovering all tests...
-> discovered 174 tests!
ensuring test passes by itself...
-> OK!
ensuring test fails with test group...
-> OK!
running step 1:
- 173 tests remaining (about 8 steps)
running step 2:
- 86 tests remaining (about 7 steps)
running step 3:
- 43 tests remaining (about 6 steps)
running step 4:
- 22 tests remaining (about 5 steps)
running step 5:
- 11 tests remaining (about 4 steps)
running step 6:
- 6 tests remaining (about 3 steps)
running step 7:
- 3 tests remaining (about 2 steps)
running step 8:
- 2 tests remaining (about 1 steps)
double checking we found it...
-> the polluting test is: yt/data_objects/tests/test_firefly.py::test_firefly_JSON_string
```

not sure exactly how `detect-test-pollution` works, so this could be a false positive, but i'll look more closely...

chrishavlin:
Summary:

```
        pytest --color=yes  \
        yt/visualization/volume_rendering/tests/test_scene.py \
        yt/data_objects/tests/test_firefly.py::test_firefly_JSON_string 
      
```
passes while
```
        pytest --color=yes  \
        yt/data_objects/tests/test_firefly.py::test_firefly_JSON_string \
        yt/visualization/volume_rendering/tests/test_scene.py
```
fails

chrishavlin:
Still fails with just the creation of the firefly obj via `create_firefly_object`, so not write related (it wasn't writing anyway...)

chrishavlin:
to reproduce locally, install Pyarrow!!!!!!!

neutrinoceros:
I confirm I can reproduce locally with
```
pytest --color=yes  \
  yt/data_objects/tests/test_firefly.py::test_firefly_JSON_string \
  yt/visualization/volume_rendering/tests/test_scene.py
```
while the reverse order passes.
However it seems that `pyarrow` isn't needed to reproduce (unless it actually leaves traces after the first failure...) ? I need to check with a brand new env too.

chrishavlin:
> However it seems that pyarrow isn't needed to reproduce

that would certainly make sense since I can't find any evidence of it being installed in the CI action. but something about installing it did let me reproduce the error locally... i'll try again in a fresh environment.

neutrinoceros:
@chrishavlin https://github.com/yt-project/yt/issues/4785#issuecomment-1911680263 👀

chrishavlin:
ya, I saw that -- but this branch is silencing those warnings already? In any case, I've now reproduced this locally without installing pyarrow. Just doing a minimal yt build and installing firefly

```
pip install -e . 
pip install firefly
```
then running the following script:

```python
from yt import create_scene
from matplotlib.image import imread
from firefly import data_reader


def test_annotations(fname):
    ds = fake_random_ds(3)
    sc = create_scene(ds)
    sc.save_annotated(fname)
    image = imread(fname)
    print(f"image shape: {image.shape}")
    print(f"camera res: {sc.camera.resolution}")
    assert image.shape == sc.camera.resolution + (4,)


test_annotations("test_prefire.png") # runs as expected
reader = data_reader.Reader(
            datadir=None, clean_datadir=True
         )  
test_annotations("test_postfire.png") # fails at the assert statement
```
and checking the 2 saved files, the second one does indeed have different pixel dimensions.

matthewturk:
I'll quick take a look but I'd bet it's related to modifying matplotlib parameters.

matthewturk:
So I ran this, and got errors on the first test line.  I put this at the top:

```python
import matplotlib as mpl
mpl.rcParams.update(mpl.rcParamsDefault)
```

and the errors on the first line went away.  (My image size, like Chris's, was 614x614.)  I checked and the only thing different between my rcParams and the other is `figure.dpi` was set to 120, which would explain the exact difference.

So, why is it changing?  My guess or suspicion is that it's related to the current working directory, or some environment variable or something, but I'll check that in a moment.

chrishavlin:
ya, this will show the change in dpi, which changes in the rcParams after the firefly import:

```python
import matplotlib as mpl
rcparams0 = mpl.rcParams.copy()
import firefly.data_reader
rcparams1 = mpl.rcParams.copy()

print(rcparams0['figure.dpi'])
print(rcparams1['figure.dpi'])
```
which prints 100 and 120

chrishavlin:
OK, i think i have a fix on the yt side...

yt/visualization/volume_rendering/scene.py
```diff
@@ -526,7 +526,9 @@ def _show_mpl(self, im, sigma_clip=None, dpi=100):
         from matplotlib.figure import Figure
 
         s = im.shape
-        self._render_figure = Figure(figsize=(s[1] / float(dpi), s[0] / float(dpi)))
+        self._render_figure = Figure(
+            figsize=(s[1] / float(dpi), s[0] / float(dpi)), dpi=dpi
```
chrishavlin, line 530: note, this is the actual fix. the default figure dpi was changing due to the firefly import (a question for firefly folks?) and so this Figure needs to be instantiated with the `dpi` variable value

chrishavlin:
assuming that this fix actually passes I'm going to close this PR and open a new one to avoid all the garbage commits here...

chrishavlin:
Replaced by #4792 . Will close this one after I update the related issue summarizing everything.

chrishavlin:
Ok, closing out now!

Issue #4799, opened 2024-02-01
TST: `yt/tests/test_load_archive.py::test_load_archive` is failing


neutrinoceros:
For instance, see https://github.com/yt-project/yt/actions/runs/7733683213/job/21093334760?pr=4797

The failure is due to a combination of a couple factors:
- Python 3.12 introduced a `DeprecationWarning` about use of `os.fork()` in multi-threaded processes (this is https://github.com/python/cpython/pull/100229)
- Pytest 8.0 was just released a couple days ago. Up until version 7.4.4, unmatched warnings in a `pytest.warns` context would be swallowed silently, which is why we haven't caught this before. (see https://github.com/pytest-dev/pytest/issues/9288)

Note that the failing test requires ratarmount (Linux only) and Python 3.12, so it's still pretty mild, as far as failures go. I suppose it should be escalated upstream to ratarmount, and the warning should be ignored on our side, as it seems out of our control.
What do you think @cphyc ?

cphyc:
I think the way forward is:
1. On the short term, this is fixed by #4819 that ignores the warning on Python 3.12,
2. On the longer term, update the version of ratarmount. At the moment, we're using 0.8.1 but the latest version is now 0.14!

PR #4805, opened 2024-02-04
ENH: add support for Quokka datasets

yut23:
## PR Summary

This allows yt to recognize Quokka output files. They are handled by the AMReX reader, but the reader does not recognize the particle data when loading Quokka outputs using `yt.load`. (Manually creating an `AMReXDataset` object does work, but this should not be necessary for users to invoke directly.)

Unlike other AMReX/Boxlib codes, Quokka doesn't have a `job_info` file, but does have a `metadata.yaml` file (which may be empty). This PR detects this file, and this causes reader to then recognize any particle data that may be present.

Thanks to @BenWibking for noticing this and providing sample outputs!

## PR Checklist

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -1334,6 +1334,12 @@ class NyxDataset(BoxlibDataset):
         setdefaultattr(self, "velocity_unit", self.length_unit / self.time_unit)
 
 
+class QuokkaDataset(AMReXDataset):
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Thank you ! Although not technically a bug fix, I'm tempted to triage this to yt 4.3.1 for practicality. I also note that failures are unrelated (see #4799). @yt-fido test this please

neutrinoceros (approved):

matthewturk (approved):

PR #4811, opened 2024-02-08
Backport PR #4802 on branch yt-4.3.x (Fix bug where race condition results in incorrect fields categorization when computing particle_trajectories)

meeseeksmachine:
Backport PR #4802: Fix bug where race condition results in incorrect fields categorization when computing particle_trajectories

PR #4813, opened 2024-02-10
Revert "Backport PR #4802 on branch yt-4.3.x (Fix bug where race condition results in incorrect fields categorization when computing particle_trajectories)"

neutrinoceros:
Reverts yt-project/yt#4811 (see my comment therein)


neutrinoceros:
Turns out, failures were indeed unrelated to this patch, but they look real still. I'll open a report later today.

PR #4819, opened 2024-02-15
Catch Py312+ warning when using ratarmount

cphyc:
## PR Summary

Since Python 3.12, `os.fork()` throws a deprecation warning when called from multiple threads.
This causes the tests relying on ratarmount (to mount archives) to fail. In this PR, I superficially address the issue by catching the warning on the test side.

The long-term solution would be to update `ratarmount` to a more recent version.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -187,14 +187,16 @@ def pytest_configure(config):
             "filterwarnings",
             r"ignore:datetime\.datetime\.utcfromtimestamp\(\) is deprecated:DeprecationWarning",
         )
-        # On Python 3.12+, there is a deprecation warning when calling os.fork()
-        # in a multi-threaded process
-        config.addinivalue_line(
-            "filterwarnings",
-            r"ignore:This process \(pid=\d+\) is multi-threaded, use of fork\(\) "
-            r"may lead to deadlocks in the child\."
-            ":DeprecationWarning",
-        )
+
+        if find_spec("ratarmount"):
+            # On Python 3.12+, there is a deprecation warning when calling os.fork()
+            # in a multi-threaded process. We use this mechanism to mount archives.
+            config.addinivalue_line(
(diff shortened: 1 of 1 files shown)


cphyc:
@yt-fido test this please

neutrinoceros (changes requested):
We should probably not *except* the deprecation warning to pop up, as this will fail again if/when ratarmount fixes it in the future.
If the fix is just to ignore the warning for now, it should be done in `conftest.py`

neutrinoceros:
I think we also need to report the problem upstream and link the issue in a comment next the the warning filter

conftest.py
```diff
@@ -187,6 +187,13 @@ def pytest_configure(config):
             "filterwarnings",
             r"ignore:datetime\.datetime\.utcfromtimestamp\(\) is deprecated:DeprecationWarning",
         )
+        # On Python 3.12+, there is a deprecation warning when calling os.fork()
+        # in a multi-threaded process
```
neutrinoceros, line 191: this comment should be a condition. Something like
`if find_spec("ratarmount") is not None and sys.version_info >= (3, 12)`:

cphyc:
I have now caught the warning in the config file.

Note however that the warning is *not* a ratarmount issue but a yt one. It is caused by the fact that ratarmount is called in a subprocess (see comment: https://github.com/cphyc/yt/blob/3524c6b305fcd710aa0155dd30e9e4ae39085a5d/yt/loaders.py#L1792-L1795). In practice, this is only a problem is `load_archive` is called from a thread, which happens for pytests, but doesn't surface in most practical cases.

neutrinoceros:
Dear me. Can't wait to see how this behaves in no-gil Python. Anyway, thank you for addressing this !

neutrinoceros (approved):

PR #4828, opened 2024-02-23
BUG: fix parameter parsing for Castro and MAESTRO

yut23:
## PR Summary

Castro (since 18.04), MAESTROeX (since 20.09), and MAESTRO (since 2013) add a prefix `[*]` before runtime parameters that are changed from their defaults. This was being handled by the Dataset subclasses individually, leading to duplicated and inconsistent logic (CastroDataset wasn't stripping the parameter names, leading to duplicates with trailing spaces).

This PR moves the prefix handling into BoxlibDataset, and removes the prefix so analysis scripts don't need to check for both versions in `ds.parameters`.

## PR Checklist

### DIFF
--- a/yt/frontends/boxlib/data_structures.py
+++ b/yt/frontends/boxlib/data_structures.py
@@ -739,6 +739,9 @@ class BoxlibDataset(Dataset):
                 param, vals = (s.strip() for s in line.split("="))
             except ValueError:
                 continue
+            # Castro and Maestro mark overridden defaults with a "[*]" before
+            # the parameter name
+            param = param.removeprefix("[*]").strip()
             if param == "amr.ref_ratio":
                 vals = self.refine_by = int(vals[0])
             elif param == "Prob.lo_bc":
@@ -1128,16 +1131,6 @@ class CastroDataset(AMReXDataset):
                     self.parameters[fields[0]] = fields[1].strip()
                 line = next(f)
 
-            # runtime parameters that we overrode follow "Inputs File
-            # Parameters"
-            # skip the "====..." line
-            line = next(f)
-            for line in f:
-                if line.strip() == "" or "fortin parameters" in line:
-                    continue
-                p, v = line.strip().split("=")
-                self.parameters[p] = v.strip()
-
         # hydro method is set by the base class -- override it here
         self.parameters["HydroMethod"] = "Castro"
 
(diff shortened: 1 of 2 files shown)


yt/frontends/boxlib/data_structures.py
```diff
@@ -739,6 +739,10 @@ def _parse_cparams(self):
                 param, vals = (s.strip() for s in line.split("="))
             except ValueError:
                 continue
+            # Castro and Maestro mark overridden defaults with a "[*]" before
+            # the parameter name
+            if param.startswith("[*]"):
+                param = param.removeprefix("[*]").strip()
```
neutrinoceros, line 745: `str.removeprefix` returns early if it has nothing to do so we shouldn't need to have a conditional. Maybe I'm missing something ?
```suggestion
            param = param.removeprefix("[*]").strip()
```

yt/frontends/boxlib/data_structures.py
```diff
@@ -1128,16 +1132,6 @@ def _parse_parameter_file(self):
                     self.parameters[fields[0]] = fields[1].strip()
                 line = next(f)
 
-            # runtime parameters that we overrode follow "Inputs File
```
neutrinoceros, line 1131: it's not clear to me why this can be removed

yt/frontends/boxlib/data_structures.py
```diff
@@ -1199,21 +1193,8 @@ def _parse_parameter_file(self):
                     fields = line.split(":")
                     self.parameters[fields[0]] = fields[1].strip()
 
-        with open(jobinfo_filename) as f:
```
neutrinoceros, line 1202: nice one

neutrinoceros (commented):
Overall looks great. Although, I don't know that we correctly test for this already, do you have a sample dataset that I could use just for manual checking that this is now doing the right thing ?
Also, if I'm reading the top description right, it sounds like this would qualify as a bug fix and should be back ported. Is that right ?

yt/frontends/boxlib/data_structures.py
yut23: I noticed this shortly after making the PR, but forgot to push.

yt/frontends/boxlib/data_structures.py
yut23: This is doing the same thing as `_parse_cparams()` in BoxlibDataset, but doesn't strip the trailing space from `p` or run `v` through `_guess_pcast()`.

yut23:
There's already a test for the MAESTRO parameters; I'll see if I can add one for Castro shortly.

yt/frontends/boxlib/tests/test_outputs.py
```diff
@@ -443,3 +443,28 @@ def test_maestro_parameters():
...
+
+    # Modified from default (leading [*])
+    assert ds.parameters["castro.do_hydro"] == 1
+    assert ds.parameters["castro.cfl"] == 0.5
+    assert ds.parameters["problem.p_ambient"] == float("1e-06")
+    # Leading [*] should be removed from the parameter name
+    assert "[*] castro.do_hydro" not in ds.parameters
+
+    # Not modified from default
+    assert ds.parameters["castro.pslope_cutoff_density"] == float("-1e+200")
```
neutrinoceros, line 464: I tested this on `main` and this assert fails there with
```
E       AssertionError: assert -1e+20 == -1e+200
E        +  where -1e+200 = float('-1e+200')
```
so I guess this parameter *is* modified from default ? or maybe I'm missing something ?

neutrinoceros (commented):
I tried running the test on main and all 6 first `assert` statements failed there, so I think we can confirm this is indeed a Bugfix :D
I still have one small question about that test, otherwise LGTM

yt/frontends/boxlib/tests/test_outputs.py
yut23: Nope, just a typo on my part. It's fixed now.

neutrinoceros:
I edited the title following the history rewrite so that it matches the commit message. Thanks Eric !

neutrinoceros (approved):

neutrinoceros:
@yt-fido test this please

PR #4806, opened 2024-02-05
RFC: replace vendored version of numpy.trapz with supported API in numpy 2

neutrinoceros:
## PR Summary

Follow up to #4632

xref https://github.com/numpy/numpy/pull/25738

This can be considered a bugfix since the previous solution (shipped in yt 4.3.0) erroneously circumvents NEP 18 implementations in `unyt`.

PR #4829, opened 2024-02-25
RFC: avoid usage of deprecated usage numpy.core (numpy 2 compat)

neutrinoceros:
## PR Summary

Combined with #4806, this replaces usages of `numpy.core` with public API.

Context: `np.core` was always private API, but in numpy 2.0, it was moved to `np._core` to make it clearer. `np.core` continues to be available as an alias for now but importing from it now triggers deprecation warnings. In the few places that we actually import it, we can now easily switch to public API, so let's.

### DIFF
--- a/yt/arraytypes.py
+++ b/yt/arraytypes.py
@@ -1,5 +1,4 @@
 import numpy as np
-import numpy.core.records as rec
 
 # Now define convenience functions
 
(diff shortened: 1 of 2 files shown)


matthewturk (approved):
As long as there're no problems with backcompat, this looks good!

neutrinoceros:
All green, including jobs with oldest numpy we support, so merging !

PR #4830, opened 2024-02-25
BLD: Ditch `np.int_t`, removed in numpy 2's C API (use `np.int64_t` instead)

neutrinoceros:
## PR Summary

This fixes *building* yt against numpy 2.0
Context:
The `np.int_t` alias was removed from numpy's C API

example error:
```
Error compiling Cython file:
------------------------------------------------------------
...
    np.float64_t left_edge_y
    np.float64_t left_edge_z
    np.float64_t right_edge_x
    np.float64_t right_edge_y
    np.float64_t right_edge_z
    np.int_t children_pointers
    ^
------------------------------------------------------------

yt/geometry/grid_visitors.pxd:34:4: 'int_t' is not a type identifier
```

Replacing it with a fixed-size portable `int64` fixes building.
This might not be an exact replacement on Windows where, historically, numpy used int32 by default, but numpy 2.0 will consistently use int64 everywhere so this change feels in line with future behaviour. Furthermore it only affects private API so it shouldn't be classified as breaking for our users.

## PR Checklist

### DIFF
--- a/yt/geometry/coordinates/cartesian_coordinates.py
+++ b/yt/geometry/coordinates/cartesian_coordinates.py
@@ -618,7 +618,7 @@ class CartesianCoordinateHandler(CoordinateHandler):
     def _oblique_pixelize(self, data_source, field, bounds, size, antialias):
         from yt.frontends.ytdata.data_structures import YTSpatialPlotDataset
 
-        indices = np.argsort(data_source["pdx"])[::-1].astype(np.int_)
+        indices = np.argsort(data_source["pdx"])[::-1].astype("int64", copy=False)
         buff = np.full((size[1], size[0]), np.nan, dtype="float64")
         ftype = "index"
         if isinstance(data_source.ds, YTSpatialPlotDataset):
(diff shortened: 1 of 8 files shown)


matthewturk:
I'm pretty sure I'm the one who did that since I don't think anybody else messes with that file, and not specifying size was definitely a mistake!

matthewturk:
(for the grid visitors example I mean)

matthewturk (dismissed):

neutrinoceros:
Thanks for reviewing so promptly! Before I undraft I want to first validate that it doesn't break CI on Windows (I don't expect it to, but better be safe than sorry).

neutrinoceros:
Took me a couple iteration to fix the Python code for Windows but it looks good now.

jzuhone (approved):

neutrinoceros:
in the wake of #4836, I'm going to merge this to reduce the number of failure modes. Thanks @jzuhone for signing this off earlier !

PR #4838, opened 2024-02-28
TST: build ewah_bool_utils from source in bleeding-edge CI

### LINKED ISSUE #4836: TST: Upcoming dependency test failures
The weekly build with future dependencies has failed. Check the logs
https://github.com/yt-project/yt/actions/runs/8074654188

neutrinoceros:
Closes #4836

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -52,7 +52,9 @@ jobs:
         python -m pip install --upgrade setuptools wheel
         python -m pip install --pre --only-binary ":all:" numpy matplotlib \
          --extra-index https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
-        python -m pip install --pre ewah-bool-utils Cython
+        python -m pip install --pre Cython
+        python -m pip install git+https://github.com/yt-project/ewah_bool_utils.git \
+         --no-build-isolation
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Note that we could also close #4836 as won't-fix: the problem will go away "by itself" as soon as `ewah-bool-utils` as a new release out to fix ABI compatibility with numpy 2.0, which I'm ready to make right after numpy 2.0.0rc1 comes out, hopefully later this week.

neutrinoceros:
I think it'd be simpler to just _not_ merge this for the time being, but numpy 2.0 might be delayed again so I'm going to keep this around as a draft for now, and I'll update depending on what happens upstream.

neutrinoceros:
update: numpy 2.0.0rc1 is still not out, and it may be further delayed in favour of a beta release (no ABI stability), see https://github.com/numpy/numpy/issues/25918#issuecomment-1980214917

In this context, I'd rather we merge this so the next 2 or 3 weekly runs are actually relevant.

neutrinoceros:
ping @matthewturk for review ?

neutrinoceros:
change of plans upstream: numpy 2.0 just got a beta release a couple hours ago, which means the release candidate isn't going to be up for a couple weeks.
However, I can just push a beta release for ewah, which will make this PR redundant.

PR #4844, opened 2024-03-05
Remove parenthesis in getitem calls

cphyc:
## PR Summary

In accordance with the style of the code, replace all occurences of `data[("gas", "density")]` with `data["gas", "density"]` (no extra parentheses).

This is merely a stylistic change.


matthewturk:
I feel like once upon a time we did it this way, and then when we moved to mandating field tuples we changed it, and I don't think I particularly cared for the change then but I honestly don't know that I am that invested one way or the other now.  It does feel like a lot of churn though.

cphyc:
To be fair I'm also happy for this _not_ to be merged. However, at present, we're using both styles in the codebase (depending on who wrote it), so for the sake of uniformity, it'd be marginally better to settle on either one of the formats (`data[("gas", "density")]` vs `data["gas", "density"]`).

cphyc:
@yt-fido test this please...

chrishavlin:
+1 from me for removing parenthesis. also agree it'd be nice to merge ASAP if folks agree.

matthewturk:
bandaids are meant to be ripped off, I suppose!

matthewturk:
And actually, how hard would it be to make a rule for ruff to apply this?  Probably easy to write, hard to confine to just yt data objects, now that I'm thinking of it.

neutrinoceros:
I had the same idea but I'm too intimidated to try it out myself :D

> hard to confine to just yt data objects, now that I'm thinking of it.

I believe parentheses are never required anywhere that looks like `[(... ,...)]`, because they are syntactically meaningless there, so I think this *would* actually work as a general purpose listing rule !

matthewturk:
> I believe parentheses are never required anywhere that looks like `[(... ,...)]`, because they are syntactically meaningless there, so I think this _would_ actually work as a general purpose listing rule !

Yeah, you must be right.  I think I got confused because of the way h5py recommends `[()]` which is specifically an empty tuple, so I think I got a bit mixed up.

neutrinoceros:
Would you like to give it a go or is it okay if we just open a feature request on ruff ?

cphyc:
Id say let's open an issue; I won't have time to implement this anytime soon.

neutrinoceros:
My question was meant for Matt, sorry for being unclear !

neutrinoceros:
No conflict after two weeks *and* no objection in sight, so I'll push the button. Thanks all !

neutrinoceros (approved):

neutrinoceros:
Btw I just opened a feature request on ruff https://github.com/astral-sh/ruff/issues/11990

PR #4851, opened 2024-03-13
TST: do not use broken pytest 8.1.1

neutrinoceros:
## PR Summary

A temporary measure to mitigate one of the failures reported in #4850
I believe the bug is upstream and reported as https://github.com/pytest-dev/pytest/issues/12113
I'll undraft this PR when CI is green in the event the problem isn't promptly resolved upstream to unblock other developments.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -217,7 +217,7 @@ minimal = [
 ]
 test = [
     "pyaml>=17.10.0",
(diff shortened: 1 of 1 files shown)


neutrinoceros:
closing this following clarification in https://github.com/pytest-dev/pytest/issues/12113

PR #4849, opened 2024-03-12
BUG: avoid a runtime warning (divide by zero) in yt.visualization.image_writer.apply_colormap

neutrinoceros:
## PR Summary

This fixes a bug that is visible in regular CI since the release of pytest 8.1.1

## PR Checklist

- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #4852, opened 2024-03-13
TST: migrate module-level setup/teardown functions to pytest

neutrinoceros:
## PR Summary

This fixes part of #4850, and effectively migrates part of the test suite away from nosetest.
It's also based off of #4849 

xref https://github.com/pytest-dev/pytest/issues/12113

Opening with a simple find/replace. I'll see which affected modules are still run on nose and wether they can simply be ignored there. If more work is needed to migrate them I'll simply make them nose-only as a quick fix.

### DIFF
--- a/nose_ignores.txt
+++ b/nose_ignores.txt
@@ -37,3 +37,7 @@
 --ignore-file=test_gadget_pytest\.py
 --ignore-file=test_vr_orientation\.py
 --ignore-file=test_particle_trajectories_pytest\.py
+--ignore-file=test_image_array\.py
+--ignore-file=test_alt_ray_tracers\.py
+--ignore-file=test_minimal_representation\.py
+--ignore-file=test_set_log_level\.py
--- a/yt/utilities/tests/test_minimal_representation.py
+++ b/yt/utilities/tests/test_minimal_representation.py
@@ -8,13 +8,17 @@ from yt.testing import requires_file, requires_module
 
 G30 = "IsolatedGalaxy/galaxy0030/galaxy0030"
 
+old_serialize = None
 
-def setup():
+
+def setup_module():
+    global old_serialize
+    old_serialize = ytcfg.get("yt", "serialize")
     ytcfg["yt", "serialize"] = True
 
 
-def teardown():
-    ytcfg["yt", "serialize"] = False
+def teardown_module():
+    ytcfg["yt", "serialize"] = old_serialize
 
 
 @requires_module("h5py")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -191,7 +191,7 @@ doc = [
     "jupyter-client>=8.3.1",
     "nbsphinx>=0.9.3",
     "nose~=1.3.7; python_version < '3.10'",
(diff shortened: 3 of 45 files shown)


neutrinoceros:
Interesting... this fails even on pytest but errors resist introspection (this smells like another case of test pollution). Let me see what I can do.

yut23:
Does `def teardown()` also need to be replaced with `def teardown_module()`?

neutrinoceros:
Oh, good point, I didn't think of that 🤦🏻‍♂️ 
Thanks a lot !

neutrinoceros:
This now closes #4856 . Thanks a bunch @yut23 !

nose_ignores.txt
```diff
@@ -37,3 +37,7 @@
 --ignore-file=test_gadget_pytest\.py
 --ignore-file=test_vr_orientation\.py
 --ignore-file=test_particle_trajectories_pytest\.py
+--ignore-file=test_image_array\.py
```
neutrinoceros, line 40: note to reviewers:
I've added modules that are now pytest-only due to usage of `teardown_module`. Many other modules have a `setup_module` method bu no `teardown_method`, and what they effectively do in setups seems safe (as in, free of side effects), and, to the best of my knowledge, is not actually used in nose.
```python
def setup_module():
    from yt.config import ytcfg

    ytcfg["yt", "internals", "within_testing"] = True
```

neutrinoceros:
@yt-fido test this please

nose_ignores.txt
matthewturk: Thanks, that makes sense.  Do you think that having pytest-only modules will change the relative runtimes of the different test suites?

neutrinoceros: not in any significant way. The four modules I just made pytest-only (combined) only take a couple seconds to run locally.

yt/utilities/tests/test_minimal_representation.py
```diff
@@ -8,13 +8,17 @@
 
 G30 = "IsolatedGalaxy/galaxy0030/galaxy0030"
 
+old_serialize = None
 
-def setup():
+
+def setup_module():
+    global old_serialize
```
chrishavlin, line 15: it'd be nice to have a pytest fixture for patching yt config options...

chrishavlin (approved):

yt/utilities/tests/test_minimal_representation.py
neutrinoceros: I'm not sure that fixtures are flexible enough to do that, but we might be able to achieve this with a context manager wrapped in a decorator ?

chrishavlin: i was thinking that a fixture that used the monkeypatch fixture internally might work  -- I was just playing around with making a copy of a YTConfig instance, modifying a value and then using monkeypatch to set `yt.config.ytcfg` to point to the modified copy and it seemed to work. But it was a little circuitous and a more general context manager would much clearer. 

In any case, a discussion for a different issue/PR :)

PR #4854, opened 2024-03-13
TST: set (very tight) error tolerance margins in `test_sph_data_objects.py`

neutrinoceros:
## PR Summary
While trying to debug #4850, I found these two tests consistently failed on my machine with machine-precision level errors.
At a glance, it seems more reasonable to check that results are very close rather than exactly identical, but I haven't given it much thought. Let me know if anyone think this is hiding a real bug !

~(this is currently based off of #4853)~

### DIFF
--- a/yt/data_objects/tests/test_sph_data_objects.py
+++ b/yt/data_objects/tests/test_sph_data_objects.py
@@ -1,5 +1,5 @@
 import numpy as np
-from numpy.testing import assert_almost_equal, assert_equal
+from numpy.testing import assert_allclose, assert_almost_equal, assert_equal
 
 from yt import SlicePlot, add_particle_filter
 from yt.loaders import load
@@ -303,7 +303,7 @@ def test_gather_slice():
     p.set_buff_size(3)
     buff_gather = p.frb.data[field].d
 
-    assert_equal(buff_scatter, buff_gather)
+    assert_allclose(buff_scatter, buff_gather, rtol=3e-16)
 
 
 def test_gather_grid():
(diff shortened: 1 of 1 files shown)


nastasha-w:
I agree there should be a margin on those tests. The grid and slice calculations have some parallel sections in the cython backend, so getting different rounding errors on different runs seems pretty reasonable.

nastasha-w (approved):
Looks good to me!

nastasha-w (approved):

neutrinoceros:
Thanks Natasha for taking a look ! We'd still need another review from an org member to get this in. @chrishavlin what do you think ?

chrishavlin:
I'm not super familiar with the scatter/gather option here, so was hoping someone else would weigh in :) 

I generally agree that adding a tolerance here makes sense. My only hesitation is that because this test is using a `fake_sph_grid_ds`, there's a chance that it really should be identical? Here's the `fake_sph_grid_ds` for reference:
>  This dataset should have 27 particles with the particles arranged uniformly
    on a 3D grid. The bottom left corner is (0.5,0.5,0.5) and the top right
    corner is (2.5,2.5,2.5). All particles will have non-overlapping smoothing
    regions with a radius of 0.05, masses of 1, and densities of 1, and zero
    velocity.

So let me actually take a little closer look before weighing in more -- I just ran the 2 tests and they failed for me as well (on a M2 macbook).

chrishavlin (approved):
Ok! +1 for adding the error tolerance here.

neutrinoceros:
thank you ! let's backport this bad boy then

PR #4859, opened 2024-03-30
BLD: add support for numpy 2

neutrinoceros:
## PR Summary

numpy 2.0.0rc1 was just released, meaning it's now ABI stable, so we can (and should) start publishing wheels built against.
This is the last blocker for yt 4.3.1

## PR Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ keywords = [
 requires-python = ">=3.9.2"
 dependencies = [
     "cmyt>=1.1.2",
-    "ewah-bool-utils>=1.0.2",
+    "ewah-bool-utils>=1.2.0",
     "ipywidgets>=8.0.0",
     "matplotlib>=3.5",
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Ah, I opened it too soon: I'm also working on the ewah-bool-utils release and wheels are still building. Will refresh when it's actually up.

chrishavlin (dismissed):

neutrinoceros:
Whoops, I forgot to also bump runtime requirements on ewah.

chrishavlin (approved):

chrishavlin:
just in case anyone else looking is initially confused like I was (@neutrinoceros correct me if I'm wrong) -- the tests that run with this PR don't actually end up using np 2.0 because some of yt's dependencies are pinning np<2 still (pandas, scipy, more). but the weekly bleeding edge test does use np 2.0, so we should be in pretty good shape as more dependencies unpin.

PR #4864, opened 2024-04-01
RFC: Replace np.random.random use with np.random.default_rng

chrishavlin:
#4863 caught some uses of the legacy np.random.random api missed by #4733. Since one of the uses is within a test function, I thought it'd be clearer to create a separate PR instead of pushing to #4863 (in case there are failures related to changing the test function)


neutrinoceros (approved):
thank you !

PR #4872, opened 2024-04-10
BUG: Always use the coordinate system normal for computing angular momentum components for particles

jzuhone:
## PR Summary

The current code for computing angular momentum components for particle fields based on their positions and velocities has an incorrect assumption. That is, if there is a `normal` field parameter (such as in the case of disk objects), that this is the normal vector that should be considered in the computation of the angular momentum. If there is no normal parameter, the normal vector is assumed to be [0,0,1]. This normal vector along with the center position is used to compute the components of position and velocity for the computation of the angular momentum vector. 

The problem with using the `normal` vector from the field parameter (if it exists) is that it involves rotating the coordinate system to one other than the default cartesian coordinate system from the dataset, but in fact the latter is the coordinate system that we want the vector components in, so we do not want to rotate. The solution is to assume that normal is always [0,0,1] for Cartesian coordinate systems. 

FWIW, the computation of the gas angular momentum in `yt/fields/angular_momentum.py` does not use the normal vector at all and thus this change brings the calculation of the angular momentum for both particles and gas into agreement with each other. 

## PR Checklist

### DIFF
--- a/yt/fields/particle_fields.py
+++ b/yt/fields/particle_fields.py
@@ -302,12 +302,7 @@ def particle_vector_functions(ptype, coord_names, vel_names, registry):
 
 
 def get_angular_momentum_components(ptype, data, spos, svel):
-    if data.has_field_parameter("normal"):
-        normal = data.get_field_parameter("normal")
-    else:
-        normal = data.ds.arr(
-            [0.0, 0.0, 1.0], "code_length"
-        )  # default to simulation axis
+    normal = data.ds.arr([0.0, 0.0, 1.0], "code_length")  # default to simulation axis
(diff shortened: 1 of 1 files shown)


chrishavlin:
@yt-fido test this please

chummels:
I'm assuming that this doesn't change the actual normal vector for the data object, which should remain what the geometry of the object defines.  So this change just means that the angular momentum vector is correctly calculated for particles in an object in the same way it is for gas?  Sounds good to me.

jzuhone:
@chummels that's correct.

chummels (approved):

Issue #4871, opened 2024-04-05
Volume Rendering Multiple Fields with MPI


biboyd:
### Bug report

**Bug summary**

When creating a volume render with multiple fields, MPI parallelism only plots part of the second field. In this case we are plotting temperature (red) and density (blue). With 2 MPI procs, the render only saves half the density.

**Code for reproduction**

```python
import yt
from yt.visualization.volume_rendering.transfer_function_helper import TransferFunctionHelper
from yt.visualization.volume_rendering.api import Scene, create_volume_source

yt.enable_parallelism()

ds = yt.load("IsolatedGalaxy/galaxy0030/galaxy0030")

# create scene and sources
sc = Scene()

source_temperature = create_volume_source(ds.all_data(), 'temperature')
source_density = create_volume_source(ds.all_data(), 'density')

#create transfer functions
tfh1 = TransferFunctionHelper(ds)
tfh1.set_field('temperature')
tfh1.set_bounds((1e3, 1e7))
tfh1.build_transfer_function()
tfh1.tf.add_layers(6, colormap="Reds")
source_temperature.transfer_function = tfh1.tf

tfh2 = TransferFunctionHelper(ds)
tfh2.set_field('density')
tfh2.set_bounds((1e-30, 1e-26))
tfh2.build_transfer_function()
tfh2.tf.add_layers(6, colormap="Blues")
source_density.transfer_function = tfh2.tf

# add sources to scene
sc.add_source(source_temperature)
sc.add_source(source_density)

# add camera
sc.add_camera(ds)
sc.camera.width = (400, "kpc")

#render/save
sc.save("render.png", sigma_clip=4)
```

**Actual outcome**
When run with 2 procs
```
$ mpirun -np 2 python test.py
```

![image](https://github.com/yt-project/yt/assets/42776109/63def4fc-f19c-49b9-ae3e-5dd0f003723d)

**Expected outcome**
When run without MPI
```
$ python test.py
```

![image](https://github.com/yt-project/yt/assets/42776109/79f7af18-82a7-4341-ab89-73ef333b7546)

**Version Information**

  * Operating System: Ubuntu 22.04.4 LTS
  * Python Version: 3.11.2
  * yt version: 4.3
  * Other Libraries (if applicable): Open MPI 4.1.2, mpi4py 3.1.5

yt installed via pip

chrishavlin:
Thanks for the report! 

Might be a duplicate of #2622 though it's interesting in your case that it happens only when rendering multiple fields.

PR #4877, opened 2024-04-16
Reduce rendered images without considering opacity

biboyd:
### Issue
This addresses issue #4871, where volume renders of multiple fields with MPI would only produce a portion of the second field in the final image. 

I believe I have found where this bug occurs. It mainly has to do with the combining of image arrays in `yt/utilities/amr_kdtree/amr_kdtools.py:receive_and_reduce()` and how the alpha channel is set in `yt/visualization/volume_rendering/render_source.py:VolumeSource.finialize_image()`

After the first source is rendered, the last step of `finalize_image()` sets the alpha channel of the image array to `1`'s 

https://github.com/yt-project/yt/blob/6d121f58543bedb1f587492cd0601a69f06a736a/yt/visualization/volume_rendering/render_source.py#L535-L537

This causes issues when the second source is combining the image arrays in `recieve_and_reduce()` as these two arrays are combined by scaling the `back` image by `ta`, `1` minus the `front` alpha channel. Since the alpha channel was already set by the first source, `ta` is a zeros array. This eliminates any contribution of `back` to the final image.

https://github.com/yt-project/yt/blob/6d121f58543bedb1f587492cd0601a69f06a736a/yt/utilities/amr_kdtree/amr_kdtools.py#L25-L34

### Solution
My proposed solution is to add a pathway to combine image arrays without the consideration of opacity. This would be used when `transfer_function.grey_opacity=False`. In this case, `front` and `back` don't have any relevance and we can simply add the two image arrays.

## PR Checklist

### DIFF
--- a/yt/utilities/amr_kdtree/amr_kdtools.py
+++ b/yt/utilities/amr_kdtree/amr_kdtools.py
@@ -3,34 +3,39 @@ import numpy as np
 from yt.funcs import mylog
 
 
-def receive_and_reduce(comm, incoming_rank, image, add_to_front):
+def receive_and_reduce(comm, incoming_rank, image, add_to_front, use_opacity):
     mylog.debug("Receiving image from %04i", incoming_rank)
     # mylog.debug( '%04i receiving image from %04i'%(self.comm.rank,back.owner))
     arr2 = comm.recv_array(incoming_rank, incoming_rank).reshape(
         (image.shape[0], image.shape[1], image.shape[2])
     )
 
-    if add_to_front:
-        front = arr2
-        back = image
-    else:
-        front = image
-        back = arr2
+    if use_opacity:
+        if add_to_front:
+            front = arr2
+            back = image
+        else:
+            front = image
+            back = arr2
 
-    if image.shape[2] == 3:
-        # Assume Projection Camera, Add
+        if image.shape[2] == 3:
+            # Assume Projection Camera, Add
+            np.add(image, front, image)
+            return image
+
+        ta = 1.0 - front[:, :, 3]
+        np.maximum(ta, 0.0, ta)
+        # This now does the following calculation, but in a memory
+        # conservative fashion
+        # image[:,:,i  ] = front[:,:,i] + ta*back[:,:,i]
+        image = back.copy()
+        for i in range(4):
+            np.multiply(image[:, :, i], ta, image[:, :, i])
         np.add(image, front, image)
-        return image
 
-    ta = 1.0 - front[:, :, 3]
-    np.maximum(ta, 0.0, ta)
-    # This now does the following calculation, but in a memory
-    # conservative fashion
-    # image[:,:,i  ] = front[:,:,i] + ta*back[:,:,i]
-    image = back.copy()
-    for i in range(4):
-        np.multiply(image[:, :, i], ta, image[:, :, i])
-    np.add(image, front, image)
+    else:
+        np.add(image, arr2, image)
+
     return image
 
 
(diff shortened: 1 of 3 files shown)


yt/utilities/amr_kdtree/amr_kdtools.py
```diff
@@ -3,34 +3,39 @@
...
     mylog.debug("Receiving image from %04i", incoming_rank)
     # mylog.debug( '%04i receiving image from %04i'%(self.comm.rank,back.owner))
     arr2 = comm.recv_array(incoming_rank, incoming_rank).reshape(
         (image.shape[0], image.shape[1], image.shape[2])
     )
 
-    if add_to_front:
-        front = arr2
-        back = image
+    if use_opacity:
```
neutrinoceros, line 13: We could reduce indentation and keep the diff size minimal by writing this new logic as an early return instead of moving everything into a branch
```suggestion
    if not use_opacity:
        np.add(image, arr2, image)
        return image
```

yt/utilities/amr_kdtree/amr_kdtools.py
```diff
@@ -3,34 +3,39 @@
 from yt.funcs import mylog
 
 
-def receive_and_reduce(comm, incoming_rank, image, add_to_front):
+def receive_and_reduce(comm, incoming_rank, image, add_to_front, use_opacity):
```
neutrinoceros, line 6: It's not clear that this function is private, so I'd suggest having a default value here too.
Also, I'd rather we avoid introducing boolean arguments in a positionally-allowed fashion
```suggestion
def receive_and_reduce(comm, incoming_rank, image, add_to_front, *, use_opacity=True):
```

yt/utilities/amr_kdtree/amr_kdtree.py
```diff
@@ -294,7 +294,7 @@ def get_reduce_owners(self):
                 owners[temp.node_id] = owners[temp.left.node_id]
         return owners
 
-    def reduce_tree_images(self, image, viewpoint):
+    def reduce_tree_images(self, image, viewpoint, use_opacity=True):
```
neutrinoceros, line 297: Let's make this keyword-only too
```suggestion
    def reduce_tree_images(self, image, viewpoint, *, use_opacity=True):
```

yt/utilities/amr_kdtree/amr_kdtree.py
```diff
@@ -307,7 +307,11 @@ def reduce_tree_images(self, image, viewpoint):
             split_pos = node.parent.get_split_pos()
             add_to_front = viewpoint[split_dim] >= split_pos
             image = receive_and_reduce(
-                self.comm, owners[node.parent.right.node_id], image, add_to_front
+                self.comm,
+                owners[node.parent.right.node_id],
+                image,
+                add_to_front,
+                use_opacity,
```
neutrinoceros, line 314:
```suggestion
                use_opacity=use_opacity,
```

neutrinoceros (changes requested):
This seems sound, but we should probably add a test too. In the mean time here are a couple suggestions

cphyc:
@yt-fido test this please

chrishavlin:
@yt-fido test this please

chrishavlin (approved):
This looks good to me. Apologies for the very delayed review here, @biboyd 

@neutrinoceros I'm not certain what could be added for a test since the changes here only have an effect when running with MPI. any ideas?

neutrinoceros:
Do you mean we don't have any form of continuous testing with MPI yet ? if so, I'd agree to leave it out of this PR, but we should probably open issue for it.

chrishavlin:
> Do you mean we don't have any form of continuous testing with MPI yet ? if so, I'd agree to leave it out of this PR, but we should probably open issue for it.

Ya, that's what I meant (but wasn't positive about it). 

So IMO this PR is OK as it is.

neutrinoceros:
Thanks for clarifying. Feel free to merge then.

PR #4878, opened 2024-04-17
ENH: Make Hilbert intersection faster

cphyc:
- replaces a Python-backed Hilbert implementation with yt's
- pre-extract the Hilbert keys once to improve perfs

On a dataset with 4096 domains, this brings down the time to compute intersection
of a box from ~ 200s to less than a few seconds.

### DIFF
--- a/yt/frontends/ramses/hilbert.py
+++ b/yt/frontends/ramses/hilbert.py
@@ -49,22 +49,22 @@ _STATE_DIAGRAM = np.array(
 
 
 def hilbert3d(
-    X: "np.ndarray[Any, np.dtype[np.float64]]", bit_length: int
+    ijk: "np.ndarray[Any, np.dtype[np.int64]]", bit_length: int
 ) -> "np.ndarray[Any, np.dtype[np.float64]]":
     """Compute the order using Hilbert indexing.
 
     Arguments
     ---------
-    X : (N, ndim) float array
+    ijk : (N, ndim) integer array
       The positions
     bit_length : integer
       The bit_length for the indexing.
     """
-    X = np.atleast_2d(X)
+    ijk = np.atleast_2d(ijk)
     # A note here: there is a freedom in the way hilbert indices are
     # being computed (should it be xyz or yzx or zxy etc.)
     # and the yt convention is not the same as the RAMSES one.
-    return get_hilbert_indices(bit_length, X[:, [1, 2, 0]])
+    return get_hilbert_indices(bit_length, ijk[:, [1, 2, 0]])
 
 
 def get_intersecting_cpus(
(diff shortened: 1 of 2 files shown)


yt/frontends/ramses/hilbert.py
```diff
@@ -221,12 +192,8 @@ def get_cpu_list_cuboid(ds, X: "np.ndarray[Any, np.dtype[np.float64]]") -> set[i
         bounding_min[i] = omin * dkey
         bounding_max[i] = (omin + 1) * dkey
```
cphyc, line 194: The reason I am modifying those lines is because I line-profiled them, and 99.9% of the time was spent extracting those bound_keys! In fact, it is much faster to extract them from the dataset (as I now do above) once and for all.

yt/frontends/ramses/hilbert.py
```diff
@@ -60,51 +61,10 @@ def hilbert3d(
...
-
-        # Compute ordering
-        for i in range(3 * bit_length):
-            order[ip] = order[ip] + i_bit_mask[i] * 2**i
-
-    return order
+    # A note here: there is a freedom in the way hilbert indices are
+    # being computed (should it be xyz or yzx or zxy etc.)
+    # and the yt convention is not the same as the RAMSES one.
+    return get_hilbert_indices(bit_length, X[:, [1, 2, 0]])
```
chrishavlin, line 67: i'm not quite sure why only the windows test failed, but `get_hilbert_indices` is expecting an int array, 

https://github.com/yt-project/yt/blob/65ae2c4cfccaeab3dd8fbf751cf5eabceb8f1c37/yt/utilities/lib/geometry_utils.pyx#L267

chrishavlin: oh, saw that this PR is still a draft, so you probably noticed this already.

cphyc: Thanks for the notification, the type hints were actually erroneous in the version that's currently on main!

neutrinoceros:
@yt-fido test this please

yt/frontends/ramses/hilbert.py
```diff
@@ -48,63 +49,22 @@
...
-            cstate = nstate
-
-        # Compute ordering
-        for i in range(3 * bit_length):
-            order[ip] = order[ip] + i_bit_mask[i] * 2**i
-
-    return order
+    ijk = np.atleast_2d(ijk)
+    # A note here: there is a freedom in the way hilbert indices are
+    # being computed (should it be xyz or yzx or zxy etc.)
```
matthewturk, line 65: Is the yt convention just the reverse?  (i.e., `xyz` to `zyx`) or are we otherwise shuffling?

matthewturk (approved):
The tests pass, this looks reasonable, and I think it's a clear improvement.  Thank you!

yt/frontends/ramses/hilbert.py
cphyc: I _think_ yt's convention is either xyz (or zyx) and RAMSES' seems to be xzy (or yzx). In any ways, it's just a permutation that needs to be done once and for all.

PR #4879, opened 2024-04-17
correct errors in capital letter

Lenoble-lab:
Hello, 
I correct some mistakes in this PR in the reader for Ramses sink files. 
I add this in the code 1yr ago and I just realize there were a mistake with a capital letter. 

In Ramses, the header of the csv file for the sinks has the header `acc_rate` and not `acc_Rate` as was in yt.
Here, I correct this mistake (line 391 of io.py)

Also for consistency I modify the name of the `particle_accretion_rate` field. It use to be called `particle_accretion_Rate` with a capital letter. But none of the others fields has a capital letter in it, so it seems more logical to me to delete this capital letter.

Best regards
Romain Lenoble

### DIFF
--- a/yt/frontends/ramses/fields.py
+++ b/yt/frontends/ramses/fields.py
@@ -137,7 +137,7 @@ class RAMSESFieldInfo(FieldInfoContainer):
         ("particle_angular_momentum_y", (ang_mom_units, [], None)),
         ("particle_angular_momentum_z", (ang_mom_units, [], None)),
         ("particle_formation_time", ("code_time", [], None)),
-        ("particle_accretion_Rate", ("code_mass/code_time", [], None)),
+        ("particle_accretion_rate", ("code_mass/code_time", [], None)),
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Thanks Romain
Just checking, was the misspelled version completely non-functional or is there a chance that someone is relying on it now ?

Lenoble-lab:
Hum, I'm not sure that someone is using it right now. 
The main pb with the current implementation is that the column `acc_rate` is not well named as the mapping is not done because of the capital letter in `io.py/_read_part_csv_file_descriptor`. So instead for the yt's default convention of naming all variables `particle_***` (with *** the name of the field), it is still name `acc_rate`. 

The second point is more cosmetic.

neutrinoceros:
Ok then, sounds like it shouldn't break anyone's workflow then. Thank you !

neutrinoceros (approved):

PR #4882, opened 2024-04-24
BUG: Update Qt backend entries in BACKEND_SPECS

### LINKED ISSUE #4881: Question/Bug: matplotlib Qt backend specification
A question on yt slack reported the following error:

```
File ~/anaconda3/lib/python3.11/site-packages/yt/visualization/base_plot_types.py:137, in PlotMPL.__init__(self, fsize, axrect, norm_handler, figure, axes)
    134     self.axes = axes
    135 self.interactivity = get_interactivity()
--> 137 figure_canvas, figure_manager = self._get_canvas_classes()
    138 self.canvas = figure_canvas(self.figure)
    139 if figure_manager is not None:

TypeError: cannot unpack non-iterable NoneType object
```

The selected matplotlib backend in this instance is `QtAgg` (found with `import matplotlib;
print(str(matplotlib.get_backend())`),  so I think because yt's `BACKEND_SPECS` only contains explicitly versioned Qt entries as keys:

https://github.com/yt-project/yt/blob/d046de2a3a4fc2834b85956e7abc516f7b403376/yt/visualization/base_plot_types.py#L52-L53

`_get_canvas_classes` is catching a KeyError and returning  `None`, resulting in the above `TypeError` : 

https://github.com/yt-project/yt/blob/d046de2a3a4fc2834b85956e7abc516f7b403376/yt/visualization/base_plot_types.py#L152-L161

According to https://matplotlib.org/stable/api/backend_qt_api.html, specifying specific Qt backends is now discouraged, so I'm thinking `BACKEND_SPECS` needs a new `QtAgg` entry. Also looks like Qt4 support was dropped in matlab 3.5.0 (https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.5.0.html#qt-backends) and 3.5.0 is currently the minimum version for yt so I think we can just replace those 2 lines with a single `QtAgg` line?  That all sound sensible? I'm not super familiar with backends, so wanted to check before getting too deep...

chrishavlin:
Closes #4881 

This adds `QtAgg` to `BACKEND_SPECS`, removes `Qt4agg` (which is no longer valid as of matplotlib 3.5.0, yt's current minimum matplotlib version) and raises an error when a backend is not present in `BACKEND_SPECS` (rather than silently returning `None`).

I'm not sure there's a way to add a test for this, but locally I was able to reproduce the error in #4881 and verify that this PR fixes the error by running the following from ipython:

```python
%matplotlib qt

import yt 
yt.toggle_interactivity()

slc = yt.SlicePlot(ds, 'x', 'Density')
slc.show()
```

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -49,8 +49,8 @@ BACKEND_SPECS = {
     "GTKAgg": ["backend_gtkagg", "FigureCanvasGTKAgg", None],
     "GTKCairo": ["backend_gtkcairo", "FigureCanvasGTKCairo", None],
     "MacOSX": ["backend_macosx", "FigureCanvasMac", "FigureManagerMac"],
-    "Qt4Agg": ["backend_qt4agg", "FigureCanvasQTAgg", None],
     "Qt5Agg": ["backend_qt5agg", "FigureCanvasQTAgg", None],
+    "QtAgg": ["backend_qtagg", "FigureCanvasQTAgg", None],
     "TkAgg": ["backend_tkagg", "FigureCanvasTkAgg", None],
     "WX": ["backend_wx", "FigureCanvasWx", None],
     "WXAgg": ["backend_wxagg", "FigureCanvasWxAgg", None],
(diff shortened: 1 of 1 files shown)


yt/visualization/base_plot_types.py
```diff
@@ -157,8 +157,9 @@ def _get_canvas_classes(self):
 
         try:
             module, fig_canvas, fig_manager = BACKEND_SPECS[key]
-        except KeyError:
-            return
+        except KeyError as err:
+            msg = f"{key} is not in list of valid backends: {list(BACKEND_SPECS)}."
```
chrishavlin, line 161: i know i just pushed this up, but i'm now thinking i **should** just remove the whole try-except, the default keyerror message that would print is probably sufficient. but i'll let tests finish running...

neutrinoceros (approved):
Thanks a lot Chris !

PR #4884, opened 2024-04-26
BUG: Fix tipsy file handler opening

### LINKED ISSUE #4883: ENH: Only open auxilliary files for Tipsy as needed
### Bug report

**Bug summary**

In `yt/frontends/tipsy/io.py` around line 160-ish we set up a dict of filehandles for the auxiliary files.  These aren't actually necessary to set up *unless* we want to read from them.  That determination happens later on, around line 172.

It might be possible to simple change the iteration `for afield in self._aux_fields:` to `for afield in afields:`.

matthewturk:
This should not change anything functionally, but can save some time for opening file handles that may be expensive.

Closes #4883.

### DIFF
--- a/yt/frontends/tipsy/io.py
+++ b/yt/frontends/tipsy/io.py
@@ -181,11 +181,11 @@ class IOHandlerTipsyBinary(IOHandlerSPH):
                         )
                     )
                 else:
-                    aux_fh[afield].seek(0)
+                    aux_fh(afield).seek(0)
                     sh = aux_fields_offsets[afield][ptype]
                     if tp[ptype] > 0:
                         aux = np.genfromtxt(
-                            aux_fh[afield], skip_header=sh, max_rows=count
+                            aux_fh(afield), skip_header=sh, max_rows=count
                         )
                         if aux.ndim < 1:
                             aux = np.array([aux])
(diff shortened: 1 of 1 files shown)


yt/frontends/tipsy/io.py
```diff
@@ -209,7 +212,7 @@ def _read_particle_data_file(self, data_file, ptf, selector=None):
 
         # close all file handles
         f.close()
-        for fh in list(aux_fh.values()):
+        for fh in list(_aux_fh.values()):
```
chrishavlin, line 215: since you touched this line, might as well get rid of the unneeded cast to `list`
```suggestion
        for fh in _aux_fh.values():
```

chrishavlin:
@yt-fido test this please

yt/frontends/tipsy/io.py
```diff
@@ -170,11 +173,11 @@ def _read_particle_data_file(self, data_file, ptf, selector=None):
...
-                aux_fh[afield].seek(aux_fields_offsets[afield][ptype])
+                aux_fh(afield).seek(aux_fields_offsets[afield][ptype])
                 if isinstance(self._aux_pdtypes[afield], np.dtype):
                     auxdata.append(
                         np.fromfile(
-                            aux_fh[afield], self._aux_pdtypes[afield], count=count
+                            aux_fh(afield), self._aux_pdtypes[afield], count=count
                         )
                     )
                 else:
```
chrishavlin, line 183: this else block needs to be updated to use the new `aux_fh` function as well (see test failures)

chrishavlin:
just pushed up commits to address my comments and fix the failing test

jzuhone (approved):

PR #4885, opened 2024-04-26
ENH: tipsy use max_rows instead of skip_footer for ascii aux read

chrishavlin:
Using `skip_footer` in `np.genfromtxt` requires scanning the whole file and so slows down the tipsy ascii aux file reads **significantly** for large files. Switching to `max_rows` avoids this.


matthewturk:
This is particularly bad for datasets with lots of subchunks!

jzuhone (approved):

PR #4897, opened 2024-05-07
TST: migrate test_disks.py from nose to pytest

neutrinoceros:
## PR Summary
A side product of my manual fixing of #4895


neutrinoceros:
Looks like I forgot to undraft this. Rebased to fix a conflict.
Can I ask @yut23 , @chrishavlin  or @cphyc for a review ?

nose_ignores.txt
```diff
@@ -44,3 +44,4 @@
 --ignore-file=test_minimal_representation\.py
 --ignore-file=test_set_log_level\.py
 --ignore-file=test_field_parsing\.py
+--ignore-file=test_disks\.py
```
chrishavlin, line 47: Should also this add to yt/tests/test.yaml -- though tests ran just fine, so maybe we don't need to actually be doing that anymore?

neutrinoceros: Indeed it seems that it's not needed here, but I don't know why TBH

chrishavlin: well I"ll hit approve and let you decide if you want to add it down in yt/test/test.yaml as well :)

chrishavlin (approved):

PR #4898, opened 2024-05-07
BLD: fix bleeding-edge build script

### LINKED ISSUE #4896: BLD: non-isolated builds are broken
### Bug report

**Bug summary**

#4894 appears to break non-isolated builds as exercised in bleeding-edge CI:
[example logs](https://github.com/yt-project/yt/actions/runs/8980365465)

@yut23, would you be able to take a look ?

yut23:
## PR Summary

Updates the bleeding-edge CI workflow script to run only pip and adds the numpy API macros to the clib defines.

Fixes #4896.

## PR Checklist

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -61,9 +61,7 @@ jobs:
       # are not installed by pip as specified from pyproject.toml, hence we get
       # to use the dev version of numpy at build time.
       run: |
-        python setup.py build_clib -q
-        python setup.py build_ext -q -j2
-        python -m pip install -e .[test] --no-build-isolation
+        python -m pip -v install -e .[test] --no-build-isolation
 
     - run: python -m pip list
 
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -61,6 +61,7 @@ jobs:
       # are not installed by pip as specified from pyproject.toml, hence we get
       # to use the dev version of numpy at build time.
       run: |
+        python setup.py build_clib -q
         python setup.py build_ext -q -j2
         python -m pip install -e .[test] --no-build-isolation
 
--- a/setup.py
+++ b/setup.py
@@ -7,13 +7,13 @@ from importlib import resources as importlib_resources
 from setuptools import Distribution, setup
 
 from setupext import (
+    NUMPY_MACROS,
     check_CPP14_flags,
     check_for_openmp,
     check_for_pyembree,
     create_build_ext,
     get_python_include_dirs,
     install_ccompiler,
-    NUMPY_MACROS,
 )
 
 install_ccompiler()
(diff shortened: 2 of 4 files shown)


yut23:
I'm not sure if `python setup.py build_ext` is necessary anymore, since the build also works fine if I remove the setup.py calls entirely.

neutrinoceros:
Thanks a lot for following up on this so promptly !

> I'm not sure if python setup.py build_ext is necessary anymore, since the build also works fine if I remove the setup.py calls entirely.

You're right, the script is currently mixing up 2 ways to build yt, and `pip install` already does build + install so setup.py invokes are redundant (and apparently broken) !

There's also a pre-commit lint error (though I don't get why it's poping now).

yut23:
Is `python setup.py develop` still a supported install method? The docs mention it here: https://yt-project.org/docs/dev/visualizing/unstructured_mesh_rendering.html. If it is, then we should probably do something like https://github.com/python-hyper/brotlicffi/pull/92 instead, as `develop` only runs `build_ext`, not `build_clib`.

yut23:
`yt update` also uses `setup.py`, see `update_git()` and `rebuild_modules()` in yt/funcs.py:
https://github.com/yt-project/yt/blob/8d58839bc16881e19f5d96c4f46a40db2b26ef0e/yt/funcs.py#L474-L481
https://github.com/yt-project/yt/blob/8d58839bc16881e19f5d96c4f46a40db2b26ef0e/yt/funcs.py#L515-L522

neutrinoceros:
nice catch. I don't know that we exercise `setup.py develop` anywhere else and I don't know if it still works as intended. `pip install` unfortunately does a bit more than what `rebuild_modules` is supposed to do so I think we should stick to `setup.py` commands in this function and include the `clibs` build step your first proposed there.
FTR I'm not sure anyone still relies on the yt command line to manage a git repo (or that anyone should), so I would advocate for deprecating them, but that's out of scope for the present PR.

.github/workflows/bleeding-edge.yaml
```diff
@@ -61,6 +61,7 @@ jobs:
       # are not installed by pip as specified from pyproject.toml, hence we get
       # to use the dev version of numpy at build time.
       run: |
+        python setup.py build_clib -q
```
neutrinoceros, line 64: I think we should just keep the `pip install` line. As it stands, extensions are needlessly compiled twice (my bad, I think !).

neutrinoceros (approved):

neutrinoceros:
"Mergeable bot" appears to be stuck. As I don't control it, I'm going to try to close/reopen, hoping this wakes it up.

matthewturk:
Ah, this seems to be related to where I was running into issues.  I believe my current development workflow is broken, as I utilize `build_ext -i` constantly and have multiple checkouts I switch between.  I find it annoying that it's not working, but unless I am able to identify a solution I won't belabor the point here and I'll just try to update my usage.

neutrinoceros:
It still works if you run `build_clib` first.

matthewturk:
Oh, awesome.  Thank you :)

PR #4907, opened 2024-05-17
BUG: Make sure fields are read in the order they appear in the file

### LINKED ISSUE #4880: BUG: some RAMSES fields return different values depending on order of operations
With RAMSES datasets, the outcome of some fields depend on previous fields having been accessed or not.

For example, the following script yields two different results depending on whether the line querying `sp["gas", "metallicity"]` is commented out or not.

## Code for reproduction
```python
import yt
import numpy as np
ds = yt.load(PATH_TO_OUTPUT)  # may need adapting
center = ds.arr([0.48722672, 0.50215806, 0.50503517], "code_length")
sp = ds.sphere(center, (100, "kpc"))
# sp["gas", "metallicity"]  # commenting this line generates another result. 
cooling = sp["gas", "cooling_metal"]
print(cooling.min(), cooling.max(), np.isnan(cooling).sum())
```
## Results

```
# When sp["gas", "metallicity"] is commented out
2.0029169125304782e-101 cm**3*erg/s     8.56002136482532e-22 cm**3*erg/s 0
# When it is not
1.3095962168963732e-105 cm**3*erg/s      6.197260629648163e-20 cm**3*erg/s 0
```

## Expected outcome
The two cases should yield the same results.

## Extra information

- yt version: development version
- Python version: 3.12

In fields.py script we can find the function     `def create_cooling_fields(self) -> bool:` that has this line for computing "cooling_metal"  `cool = ( cool * data["gas", "metallicity"] / 0.02 )  # Ramses uses Zsolar=0.02` If the values of metallicity are checked (if they are >1 ) there are two possibilities:  1) you call in your code `data["gas", "metallicity"] `first and then  `data["gas", "cooling_metal"]`, the result is ok. 2) you call directly `data["gas", "cooling_metal"]` the result is wrong; metallicities are >1.

I think the FieldDetector saves values into metallicity and those values do not overwrite with the correct ones. I do not know if more fields have the same issue. 

Cheers!

cphyc:
## PR Summary

This fixes #4880.

The issue was that the reader would assume fields were passed in the order they appeared on file, so that depending on the order in which one would load the data, we'd get different results.

### DIFF
--- a/yt/frontends/ramses/tests/test_outputs.py
+++ b/yt/frontends/ramses/tests/test_outputs.py
@@ -646,21 +646,21 @@ def test_print_stats():
     # FIXME #3197: use `capsys` with pytest to make sure the print_stats function works as intended
 
 
-def _dummy_field(field, data):
-    # Note: this is a dummy field
-    # that doesn't really have any physical meaning
-    # but may trigger some bug in the field
-    # handling.
-    T = data["gas", "temperature"]
-    Z = data["gas", "metallicity"]
-    return T * 1**Z
-
-
 @requires_file(output_00080)
 def test_reading_order():
     # This checks the bug unvovered in #4880
     # This checks that the result of field accession doesn't
     # depend on the order
+
+    def _dummy_field(field, data):
+        # Note: this is a dummy field
+        # that doesn't really have any physical meaning
+        # but may trigger some bug in the field
+        # handling.
+        T = data["gas", "temperature"]
+        Z = data["gas", "metallicity"]
+        return T * 1**Z
+
     fields = [
         "Density",
         "x-velocity",
--- a/yt/frontends/ramses/tests/test_outputs.py
+++ b/yt/frontends/ramses/tests/test_outputs.py
@@ -644,3 +644,42 @@ def test_print_stats():
     ds.print_stats()
 
     # FIXME #3197: use `capsys` with pytest to make sure the print_stats function works as intended
+
+
+def _dummy_field(field, data):
+    # Note: this is a dummy field
+    # that doesn't really have any physical meaning
+    # but may trigger some bug in the field
+    # handling.
+    T = data["gas", "temperature"]
+    Z = data["gas", "metallicity"]
+    return T * 1**Z
+
+
+@requires_file(output_00080)
+def test_reading_order():
+    # This checks the bug unvovered in #4880
+    # This checks that the result of field accession doesn't
+    # depend on the order
+    fields = [
+        "Density",
+        "x-velocity",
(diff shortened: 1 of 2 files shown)


matthewturk:
This looks like it might fix bugs ... were there any?

cphyc:
@matthewturk yes, this fixes the bug reported in #4880 (which IMO is a major one).

Essentially, when yt detects field dependences, it creates a list of on-disk fields in an arbitrary order. This then gets passed to the reader, which (wrongly) was assuming the fields were passed in the order the appear in the file.

In #4880, yt tries to read density, metallicity, pressure but the fields appear on-disk in the order density, pressure, metallicity. As a consequence, pressure and metallicity end up being swapped. This PR fixes this by ordering the fields in the order they appear on disk and voilà.

neutrinoceros:
Hum. This looks simple enough but I'm not comfortable to merge a fix for a major bug without at least an attempt at automated testing. It probably not possible to prove the software is always correct, but a simple regression test seems feasible ?

cphyc:
I'll give it a try, but so far I only was able to trigger it with a somewhat massive simulation (few 10GiB).

matthewturk:
I think the likelihood of a problem is low enough that this should go in.  We do similar logic elsewhere, and this seems absolutely correct.

matthewturk (approved):

neutrinoceros:
Fair enough. @cphyc, suit yourself !

neutrinoceros:
(to be clear, I'm not worried that the patch is incorrect, I'm worried that a similar bug might creep back in at some later point in the future and we have to pay the full price of figuring it out again)

cphyc:
@neutrinoceros and @matthewturk I've checked locally on main, and the test I wrote is failing (but not anymore on #4907!). FYI, I've added the test in the part that's being tested by nose rather than pytest because it appears that no pytest run use actual data…

yt/frontends/ramses/tests/test_outputs.py
```diff
@@ -644,3 +644,42 @@ def test_print_stats():
     ds.print_stats()
 
     # FIXME #3197: use `capsys` with pytest to make sure the print_stats function works as intended
+
+
+def _dummy_field(field, data):
```
neutrinoceros, line 649: I'd suggest to move this inside the test function for clarity

neutrinoceros (approved):
I checked the test fails on main and pass with this branch, thanks a lot ! Feel free to ignore my minor remark and merge if you will, otherwise I'm happy to approve again later today.

neutrinoceros (approved):

PR #4845, opened 2024-03-05
Rename BoxLib frontend as AMReX

henrynjones:
## PR Summary

This PR renames the `boxlib` frontend to `amrex` as BoxLib is deprecated and has been superseded by AMReX. A container `boxlib` frontend which points to the renamed `amrex` frontend has been added for backwards compatibility. 

## PR Checklist

- [x] Adds a test for any bugs fixed. Adds tests for new features.
- [x] New features are documented, with docstrings and narrative docs

PR #4908, opened 2024-05-17
ENH: add support for molecular fields (AMRex)

simonguichandut:
## PR Summary

The boxlib dataset parsing (for CASTRO/MAESTROeX) are currently set up for usual atomic species like "He4" or "C12", or descriptive names like "ash" with no numbers.  The parsing breaks for e.g. molecules like "H2O" or "C6H6".  This fix escapes those cases.

Ideally, we could have some fancy regex to correctly parse the molecules (and create proper a TeX string), but this is a very niche use of those codes.

## PR Checklist

### DIFF
--- a/nose_ignores.txt
+++ b/nose_ignores.txt
@@ -43,3 +43,4 @@
 --ignore-file=test_alt_ray_tracers\.py
 --ignore-file=test_minimal_representation\.py
 --ignore-file=test_set_log_level\.py
+--ignore-file=test_field_parsing\.py
--- a/yt/frontends/amrex/fields.py
+++ b/yt/frontends/amrex/fields.py
@@ -1,4 +1,5 @@
 import re
+import sys
 
 import numpy as np
 
@@ -7,11 +8,14 @@ from yt.fields.field_info_container import FieldInfoContainer
 from yt.units import YTQuantity
 from yt.utilities.physical_constants import amu_cgs, boltzmann_constant_cgs, c
 
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
+
 rho_units = "code_mass / code_length**3"
 mom_units = "code_mass / (code_time * code_length**2)"
 eden_units = "code_mass / (code_time**2 * code_length)"  # erg / cm^3
-
-spec_finder = re.compile(r".*\((\D*)(\d*)\).*")
 
 
 def _thermal_energy_density(field, data):
@@ -357,16 +361,22 @@ class CastroFieldInfo(FieldInfoContainer):
         for _, field in self.ds.field_list:
             if field.startswith("X("):
                 # We have a fraction
-                nice_name, tex_label = _nice_species_name(field)
-                self.alias(
-                    ("gas", f"{nice_name}_fraction"), ("boxlib", field), units=""
+                sub = Substance(field)
+                # Overwrite field to use nicer tex_label display_name
+                self.add_output_field(
+                    ("boxlib", field),
+                    sampling_type="cell",
+                    units="",
+                    display_name=rf"X\left({sub.to_tex()}\right)",
                 )
-                func = _create_density_func(("gas", f"{nice_name}_fraction"))
+                self.alias(("gas", f"{sub}_fraction"), ("boxlib", field), units="")
+                func = _create_density_func(("gas", f"{sub}_fraction"))
                 self.add_field(
-                    name=("gas", f"{nice_name}_density"),
+                    name=("gas", f"{sub}_density"),
                     sampling_type="cell",
                     function=func,
                     units=self.ds.unit_system["density"],
+                    display_name=rf"\rho {sub.to_tex()}",
                 )
 
 
@@ -462,29 +472,27 @@ class MaestroFieldInfo(FieldInfoContainer):
         for _, field in self.ds.field_list:
             if field.startswith("X("):
                 # We have a mass fraction
-                nice_name, tex_label = _nice_species_name(field)
+                sub = Substance(field)
                 # Overwrite field to use nicer tex_label display_name
                 self.add_output_field(
                     ("boxlib", field),
                     sampling_type="cell",
                     units="",
-                    display_name=tex_label,
+                    display_name=rf"X\left({sub.to_tex()}\right)",
                 )
-                self.alias(
-                    ("gas", f"{nice_name}_fraction"), ("boxlib", field), units=""
-                )
-                func = _create_density_func(("gas", f"{nice_name}_fraction"))
+                self.alias(("gas", f"{sub}_fraction"), ("boxlib", field), units="")
+                func = _create_density_func(("gas", f"{sub}_fraction"))
                 self.add_field(
-                    name=("gas", f"{nice_name}_density"),
+                    name=("gas", f"{sub}_density"),
                     sampling_type="cell",
                     function=func,
                     units=unit_system["density"],
-                    display_name=rf"\rho {tex_label}",
+                    display_name=rf"\rho {sub.to_tex()}",
                 )
 
             elif field.startswith("omegadot("):
-                nice_name, tex_label = _nice_species_name(field)
-                display_name = rf"\dot{{\omega}}\left[{tex_label}\right]"
+                sub = Substance(field)
+                display_name = rf"\dot{{\omega}}\left[{sub.to_tex()}\right]"
                 # Overwrite field to use nicer tex_label'ed display_name
                 self.add_output_field(
                     ("boxlib", field),
@@ -493,23 +501,70 @@ class MaestroFieldInfo(FieldInfoContainer):
                     display_name=display_name,
                 )
                 self.alias(
-                    ("gas", f"{nice_name}_creation_rate"),
+                    ("gas", f"{sub}_creation_rate"),
                     ("boxlib", field),
                     units=unit_system["frequency"],
                 )
 
 
-def _nice_species_name(field):
-    spec_match = spec_finder.search(field)
-    nice_name = "".join(spec_match.groups())
-    # if the species field is a descriptive name, then the match
-    # on the integer will be blank
-    # modify the tex string in this case to remove spurious tex spacing
-    lab = r"X\left(^{%s}%s\right)"
-    if spec_match.groups()[-1] == "":
-        lab = r"X\left(%s%s\right)"
-    tex_label = lab % spec_match.groups()[::-1]
-    return nice_name, tex_label
+substance_expr_re = re.compile(r"\(([a-zA-Z][a-zA-Z0-9]*)\)")
+substance_elements_re = re.compile(r"(?P<element>[a-zA-Z]+)(?P<digits>\d*)")
+SubstanceSpec: TypeAlias = list[tuple[str, int]]
+
+
+class Substance:
+    def __init__(self, data: str) -> None:
+        if (m := substance_expr_re.search(data)) is None:
+            raise ValueError(f"{data!r} doesn't match expected regular expression")
+        sub_str = m.group()
+        constituents = substance_elements_re.findall(sub_str)
+
+        # 0 is used as a sentinel value to mark descriptive names
+        default_value = 1 if len(constituents) > 1 else 0
+        self._spec: SubstanceSpec = [
+            (name, int(count or default_value)) for (name, count) in constituents
+        ]
+
+    def get_spec(self) -> SubstanceSpec:
+        return self._spec.copy()
+
+    def is_isotope(self) -> bool:
+        return len(self._spec) == 1 and self._spec[0][1] > 0
+
+    def is_molecule(self) -> bool:
+        return len(self._spec) != 1
+
+    def is_descriptive_name(self) -> bool:
+        return len(self._spec) == 1 and self._spec[0][1] == 0
+
+    def __str__(self) -> str:
+        return "".join(
+            f"{element}{count if count > 1 else ''}" for element, count in self._spec
+        )
+
+    def _to_tex_isotope(self) -> str:
+        element, count = self._spec[0]
+        return rf"^{{{count}}}{element}"
+
+    def _to_tex_molecule(self) -> str:
+        return "".join(
+            rf"{element}_{{{count if count>1 else ''}}}"
+            for element, count in self._spec
+        )
+
+    def _to_tex_descriptive(self) -> str:
+        return str(self)
+
+    def to_tex(self) -> str:
+        if self.is_isotope():
+            return self._to_tex_isotope()
+        elif self.is_molecule():
+            return self._to_tex_molecule()
+        elif self.is_descriptive_name():
+            return self._to_tex_descriptive()
+        else:
+            # should only be reachable in case of a regular expression defect
+            raise RuntimeError
 
 
 def _create_density_func(field_name):
--- a/tests/tests.yaml
+++ b/tests/tests.yaml
@@ -187,6 +187,7 @@ other_tests:
      - "--ignore-file=test_ewah_write_load\\.py"
      - "--ignore-file=test_external_frontends\\.py"
      - "--ignore-file=test_field_access_pytest\\.py"
+     - "--ignore-file=test_field_parsing\\.py"
      - "--ignore-file=test_file_sanitizer\\.py"
      - "--ignore-file=test_firefly\\.py"
      - "--ignore-file=test_geometries\\.py"
--- a/yt/frontends/amrex/tests/test_field_parsing.py
+++ b/yt/frontends/amrex/tests/test_field_parsing.py
@@ -1,0 +1,51 @@
+import pytest
+
+from yt.frontends.amrex.fields import Substance
+
+
+@pytest.mark.parametrize(
+    "data, expected",
+    [
+        pytest.param("X(He5)", [("He", 5)], id="isotope_1"),
+        pytest.param("X(C12)", [("C", 12)], id="isotope_2"),
+        pytest.param("X(A1B2C3)", [("A", 1), ("B", 2), ("C", 3)], id="molecule_1"),
+        pytest.param("X(C12H24)", [("C", 12), ("H", 24)], id="molecule_2"),
+        pytest.param("X(H2O)", [("H", 2), ("O", 1)], id="molecule_3"),
+        pytest.param("X(ash)", [("ash", 0)], id="descriptive_name"),
+    ],
+)
+def test_Substance_spec(data, expected):
+    assert Substance(data)._spec == expected
+
+
+@pytest.mark.parametrize(
+    "data, expected_type",
+    [
+        pytest.param("X(He5)", "isotope", id="isotope_1"),
+        pytest.param("X(C12)", "isotope", id="isotope_2"),
+        pytest.param("X(A1B2C3)", "molecule", id="molecule_1"),
+        pytest.param("X(C12H24)", "molecule", id="molecule_2"),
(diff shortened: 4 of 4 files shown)


yt/frontends/boxlib/fields.py
```diff
@@ -523,15 +525,21 @@ def setup_fluid_fields(self):
...
+        nice_name = "".join(spec_match.groups())
+        # if the species field is a descriptive name, then the match
+        # on the integer will be blank
+        # modify the tex string in this case to remove spurious tex spacing
+        lab = r"X\left(^{%s}%s\right)"
+        if spec_match.groups()[-1] == "":
+            lab = r"X\left(%s%s\right)"
+        tex_label = lab % spec_match.groups()[::-1]
+
+    except:
```
neutrinoceros, line 539: Bare except statements should be avoided, please only catch specific exceptions. The code in the `try` branch should also be limited to the part that's expected to sometimes raise exceptions.

matthewturk:
Thanks for doing this, @simonguichandut !  It's tough for us to anticipate all the different representations, so making sure we're up to date on how codes represent molecules and aligning that with your expectations is really important.  Thank you!

zingale:
I tested this with a Castro and MAESTROeX plotfile and it works.

yt/frontends/boxlib/fields.py
```diff
@@ -523,15 +525,24 @@ def setup_fluid_fields(self):
...
 def _nice_species_name(field):
-    spec_match = spec_finder.search(field)
-    nice_name = "".join(spec_match.groups())
+    try:
+        spec_match = spec_finder.search(field)
+        nice_name = "".join(spec_match.groups())
+
+    except AttributeError as e:
+        # Catch exception cases for e.g. molecules (H2O, C6H6)
+        print("Could not parse species ", field)
```
neutrinoceros, line 534: We do not use `print` to log errors. You could use `ytLogger.error` here instead.

yt/frontends/boxlib/fields.py
```diff
@@ -523,15 +525,24 @@ def setup_fluid_fields(self):
...
+        return nice_name, tex_label
+
     # if the species field is a descriptive name, then the match
     # on the integer will be blank
     # modify the tex string in this case to remove spurious tex spacing
     lab = r"X\left(^{%s}%s\right)"
     if spec_match.groups()[-1] == "":
         lab = r"X\left(%s%s\right)"
     tex_label = lab % spec_match.groups()[::-1]
+
```
neutrinoceros, line 545: Given the existence of #4845, where this file is moved, we're going to get merge conflict anyway so let's make their resolution as easy as possible by keeping the diff minimal.
```suggestion
```

yt/frontends/boxlib/fields.py
```diff
@@ -493,8 +493,10 @@ def setup_fluid_fields(self):
 
                 # Most of the time our species will be of the form
                 # element name + atomic weight (e.g. C12), but
-                # sometimes we make up descriptive names (e.g. ash)
-                if any(char.isdigit() for char in field):
+                # sometimes we make up descriptive names (e.g. ash).
+                # In niche cases, we might have a molecule with a number
+                # in the middle (e.g. H2O). We ignore those.
+                if any(char.isdigit() for char in field) and field[-2].isdigit():
```
neutrinoceros, line 499: I don't think I understand what this branch is supposed to capture. Both examples in the comment (`C12`, `H20`) check `field[-2].isdigit()`, so they both go through, but what I understand from the comment is that `C12` should go, and `H2O` shouldn't. What am I missing ?

yt/frontends/boxlib/fields.py
```diff
@@ -523,15 +525,24 @@ def setup_fluid_fields(self):
...
 
 def _nice_species_name(field):
-    spec_match = spec_finder.search(field)
-    nice_name = "".join(spec_match.groups())
+    try:
+        spec_match = spec_finder.search(field)
+        nice_name = "".join(spec_match.groups())
+
+    except AttributeError as e:
+        # Catch exception cases for e.g. molecules (H2O, C6H6)
```
neutrinoceros, line 533: Upon a closer look, I don't think we don't need a try/except at all here
```suggestion
    if (spec_match := spec_finder.search(field)) is not None:
        nice_name = "".join(spec_match.groups())
    else:
        # molecules (e.g. H2O, C6H6)
```

neutrinoceros (commented):
Thank you for working on this ! Here's a couple questions and suggestions.

yt/frontends/boxlib/fields.py
simonguichandut: Just to make sure I understand, you are only suggesting to remove the blank line here?

zingale: I think #4845 is ready to merge, right?
should we wait on anything else in that PR?

yt/frontends/boxlib/fields.py
simonguichandut: Actually the strings in the plotfiles are X(C12), X(H2O).  Previously the last character before the ")" parenthesis would be a number, but not in the H2O case, which will raise an error in `weight = int(weight)` a few lines down.  Note that this block of code is for the moment superfluous, element and weight are not used.

yt/frontends/boxlib/fields.py
neutrinoceros: > Just to make sure I understand, you are only suggesting to remove the blank line here?

Yes

neutrinoceros: > should we wait on anything else in that PR?

I just want to honor the "one week or so" delay I promised in my last call for objections. Reviews came in faster than I anticipated, but there's no harm in waiting a couple days more, I think.

yt/frontends/boxlib/fields.py
neutrinoceros: The comment should reflect that then. Ideally the code would be expressive enough that a comment wouldn't even be needed.
Are you saying the whole block is superfluous ? If so, I think it should be removed, or at least not touched in this PR.

yt/frontends/boxlib/fields.py
simonguichandut: It's not really an "error" and ytLogger prints a whole traceback, not sure that is the ideal solution

yt/frontends/boxlib/fields.py
simonguichandut: I think it was put in there so eventually we could use the element and weight variables (@zingale ?)
https://github.com/yt-project/yt/blob/d7209f9aab2587bca4264f42c2106d4bf6b5cf88/yt/frontends/boxlib/fields.py#L509-L510

simonguichandut: I only made this change so that the code wouldn't break when there is a species like H2O.

yt/frontends/boxlib/fields.py
neutrinoceros: > It's not really an "error"

please clarify. To me, seeing `return` inside an `except` blocks means we've encountered a recoverable exception (error).


> ytLogger prints a whole traceback

I'm not sure what you mean here. `ytLogger` is meant to print messages to the console and we generally don't include (even partial) tracebacks.

yt/frontends/boxlib/fields.py
neutrinoceros: wouldn't the ideal fix be to extend the regexp so that it doesn't break for these cases ?

yt/frontends/boxlib/fields.py
simonguichandut: Sorry I'm using the wrong terminology. I'm just trying to say that the "error" in this case is not a "big deal". The species is not recognized, so we won't try to produce a nice tex label, that's it.  ytLogger prints out very long error message and the whole call stack, which feels like overkill.

Here's what I get 
```
python $MAESTROEX_DIR/Util/yt/plotsinglevar.py plt_0000100 "X(H2O)"
--- Logging error ---
  File "/Users/simon/anaconda3/lib/python3.11/logging/__init__.py", line 1110, in emit
    msg = self.format(record)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/simon/anaconda3/lib/python3.11/logging/__init__.py", line 953, in format
    return fmt.format(record)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/simon/anaconda3/lib/python3.11/logging/__init__.py", line 687, in format
    record.message = record.getMessage()
                     ^^^^^^^^^^^^^^^^^^^
  File "/Users/simon/anaconda3/lib/python3.11/logging/__init__.py", line 377, in getMessage
    msg = msg % self.args
          ~~~~^~~~~~~~~~~
TypeError: not all arguments converted during string formatting
Call stack:
  File "/Users/simon/Desktop/codes/AMREX-Astro/MAESTROeX/Util/yt/plotsinglevar.py", line 94, in <module>
    plot_single_var(args.plotfile, args.outfile, args.variables, use_log, args.norm, args.minimum, args.maximum)
  File "/Users/simon/Desktop/codes/AMREX-Astro/MAESTROeX/Util/yt/plotsinglevar.py", line 52, in plot_single_var
    plots = yt.SlicePlot(ds, norm_axis, var_names)
  File "/Users/simon/Desktop/codes/yt-git/yt/visualization/plot_window.py", line 1821, in __init__
    (bounds, center, display_center) = get_window_parameters(
  File "/Users/simon/Desktop/codes/yt-git/yt/visualization/plot_window.py", line 70, in get_window_parameters
    width = ds.coordinates.sanitize_width(axis, width, None)
  File "/Users/simon/Desktop/codes/yt-git/yt/geometry/coordinates/coordinate_handler.py", line 300, in sanitize_width
    self.ds.index
  File "/Users/simon/Desktop/codes/yt-git/yt/data_objects/static_output.py", line 613, in index
    self.create_field_info()
  File "/Users/simon/Desktop/codes/yt-git/yt/data_objects/static_output.py", line 665, in create_field_info
    self.field_info.setup_fluid_fields()
  File "/Users/simon/Desktop/codes/yt-git/yt/frontends/boxlib/fields.py", line 475, in setup_fluid_fields
    nice_name, tex_label = _nice_species_name(field)
  File "/Users/simon/Desktop/codes/yt-git/yt/frontends/boxlib/fields.py", line 534, in _nice_species_name
    ytLogger.error("Could not parse species ", field)
Message: 'Could not parse species '
Arguments: ('X(H2O)',)
```

yt/frontends/boxlib/fields.py
simonguichandut: Yes, I mentioned this when I submitted the PR.  I tried for a while but it was too difficult for me.  Ideally you would also make the numbers subscripts instead of superscript for the tex label for molecules.  Again, having molecules in the first place is a very niche use of castro/maestro...

yt/frontends/boxlib/fields.py
neutrinoceros: The traceback isn't a feature of the logger, but shows a mistake in your call to `ytLogger.error`:

Instead of
```python
ytLogger.error("Could not parse species ", field)
```
write
```python
ytLogger.error("Could not parse species %s", field)
```

yt/frontends/boxlib/fields.py
neutrinoceros: Ok, thanks for clarifying. I suggest we slightly rephrase the logging error message to make it clear that it's a missing feature, not an actual bug, so that readers might choose to give it a try too :)

yt/frontends/boxlib/fields.py
simonguichandut: Oh, thank you!

yt/frontends/boxlib/fields.py
simonguichandut: I wrote a brief something, but please feel free to edit

neutrinoceros:
pre-commit.ci autofix

yt/frontends/boxlib/fields.py
```diff
@@ -492,9 +493,12 @@ def setup_fluid_fields(self):
...
                 # Most of the time our species will be of the form
-                # element name + atomic weight (e.g. C12), but
-                # sometimes we make up descriptive names (e.g. ash)
-                if any(char.isdigit() for char in field):
+                # element name + atomic weight (e.g. X(C12)), but
+                # sometimes we make up descriptive names (e.g. X(ash)).
+                # In niche cases, we might have a molecule with a number
+                # in the middle (e.g. X(H2O)). Check if the last character before
+                # the ) is a digit.
+                if field[-2].isdigit():
```
neutrinoceros, line 501: I still find this too fragile, but digging for a way to make it more robust, I think I ended up solving the problem of parsing molecules. Do you mind if I push to your branch ? I could also open a separate PR.

simonguichandut: Sure, whatever you think is best

neutrinoceros: ok I'll set myself a reminder to execute this by next week, thank you for your patience !

neutrinoceros:
I'm going to follow through my comment from last week now. First, I'll rebase the branch to resolve merge conflicts.

neutrinoceros:
Here goes nothing

yt/frontends/amrex/fields.py
```diff
@@ -471,43 +473,27 @@ def setup_fluid_fields(self):
...
-                if any(char.isdigit() for char in field):
-                    # We know this will either have one letter, or two.
-                    if field[3] in string.ascii_letters:
-                        element, weight = field[2:4], field[4:-1]
-                    else:
-                        element, weight = field[2:3], field[3:-1]  # NOQA
-                    weight = int(weight)
-
-                # Here we can, later, add number density using 'element' and
-                # 'weight' inferred above
```
neutrinoceros, line 506: As I was refactoring this branch, I realized it had really been left in a noop "TODO" state for 10 years and I feel like it's not worth keeping it around given that it creates discussion around code that's actually not used.

neutrinoceros:
I ended up radically re-orienting the PR's implementation. Hopefully the behavior I implemented is suitable.

yt/frontends/amrex/fields.py
yut23: We should also remove the same code in `CastroFieldInfo.setup_fluid_fields()`.

nose_ignores.txt
```diff
@@ -43,3 +43,4 @@
 --ignore-file=test_alt_ray_tracers\.py
 --ignore-file=test_minimal_representation\.py
 --ignore-file=test_set_log_level\.py
+--ignore-file=test_field_parsing\.py
```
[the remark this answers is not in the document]
neutrinoceros: indeed I forgot about this. Thank you !

neutrinoceros:
Ok, CI is green now, so let's open back for review. @simonguichandut what do you think ?

yt/frontends/amrex/fields.py
neutrinoceros: Good point. Actually a separable problem so I'll open another PR. Thank you !

yut23:
The TeX formatting is a bit wonky now: molecules and descriptive names are missing the "X" that the isotopes have, and omega dot for isotopes is rendered like `\dot{\omega}[X(He4)]`. I think it would be better to have `Substance.to_tex()` return just the rendered name (e.g. `^{4}He` and `C_{12}H_{24}`) and have `setup_fluid_fields()` add the parentheses and `X`/`\dot{\omega}`.

neutrinoceros:
Good points ! Separation of concerns FTW ! better now ?

yut23:
Yes, much better. Can we also add the tex formatting to CastroFieldInfo?

neutrinoceros:
done !
I'll rebase the branch again if you're happy with it

yt/frontends/amrex/fields.py
```diff
@@ -357,16 +361,15 @@ def setup_fluid_fields(self):
...
+                sub = Substance(field)
+                self.alias(("gas", f"{sub}_fraction"), ("boxlib", field), units="")
+                func = _create_density_func(("gas", f"{sub}_fraction"))
                 self.add_field(
-                    name=("gas", f"{nice_name}_density"),
+                    name=("gas", f"{sub}_density"),
                     sampling_type="cell",
                     function=func,
                     units=self.ds.unit_system["density"],
+                    display_name=rf"X\left({sub.to_tex()}\right)",
```
yut23, line 372: To match Maestro:
```suggestion
                    display_name=rf"\rho {sub.to_tex()}",
```

yt/frontends/amrex/fields.py
```diff
@@ -357,16 +361,15 @@ def setup_fluid_fields(self):
         for _, field in self.ds.field_list:
             if field.startswith("X("):
                 # We have a fraction
-                nice_name, tex_label = _nice_species_name(field)
-                self.alias(
-                    ("gas", f"{nice_name}_fraction"), ("boxlib", field), units=""
-                )
-                func = _create_density_func(("gas", f"{nice_name}_fraction"))
+                sub = Substance(field)
```
yut23, line 364:
```suggestion
                sub = Substance(field)
                # Overwrite field to use nicer tex_label display_name
                self.add_output_field(
                    ("boxlib", field),
                    sampling_type="cell",
                    units="",
                    display_name=rf"X\left({sub.to_tex()}\right)",
                )
```

yut23 (approved):

neutrinoceros:
thanks @yut23 for your input ! Now I think we should give @simonguichandut a couple weeks to comments before we can merge.

yt/frontends/amrex/fields.py
```diff
@@ -493,23 +501,62 @@ def setup_fluid_fields(self):
...
+
+    def __str__(self) -> str:
+        return "".join(f"{element}{count or ''}" for element, count in self._spec)
+
+    def _to_tex_isotope(self) -> str:
+        element, count = self._spec[0]
+        return rf"^{{{count}}}{element}"
+
+    def _to_tex_molecule(self) -> str:
+        return "".join(rf"{element}_{{{count}}}" for element, count in self._spec)
```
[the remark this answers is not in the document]
neutrinoceros: nice catch. I adjusted tests and fixed this bug as you suggested !

simonguichandut:
Looks great to me! I've tested on MAESTROeX plotfiles, with and without molecules, and it works as expected.

neutrinoceros:
Awesome, I squashed all my commits together, let me know if you're okay with doing the same with yours !

simonguichandut:
> Awesome, I squashed all my commits together, let me know if I should do the same with yours !

I'm not sure what this means/does, but either way I will delete my branch and sync with the main repo! :)

neutrinoceros:
I'm proposing to rewrite the branch history so  it contains just one commit. With some tools (mainly `git bisect`) makes it more confortable to debug if we need to revisit this change in the future. The reason I asked is that I don't want to make your life harder in any way, but if it sounds like it wouldn't, so I'll just go ahead.

yut23 (approved):
Looks good to me!

zingale:
How does this handle something like "H2", with is either an isotope (deuterium) or a molecule (molecular H)?
We probably want to default to isotopes.

simonguichandut:
> How does this handle something like "H2", with is either an isotope (deuterium) or a molecule (molecular H)? We probably want to default to isotopes.

H2 does get recognized as the isotope

neutrinoceros:
Yes, that's intentional, because from what I understand molecules are rarely seen in this context, but it's a shame that there's no way to tell for sure from just looking at the string `"H2"`. 
@zingale I could sediment the current behavior by adding a test case for it if you'd like.

zingale:
no, I think it's fine.  Just wanted to double check.

neutrinoceros:
Alright let's just merge then. Thanks all !

PR #4911, opened 2024-05-21
BLD: fix segfault with manylinux2014

yut23:
## PR Summary

Follow-up to https://github.com/yt-project/yt/pull/4909.

We originally had static linking enabled due to ABI conflicts, as discussed in https://github.com/yt-project/yt/pull/3406. However, the cibuildwheel docs currently say that manylinux2014 supports all C++ standards up to C++17 (https://cibuildwheel.pypa.io/en/stable/cpp_standards/).

I think the segfaults may have to do with multiple versions of libstdc++ being loaded at the same time. A couple of compiled modules in matplotlib dynamically link to `libstdc++.so.6`, which seems like it could cause this sort of issue.


neutrinoceros:
Thanks a lot for digging into this ! I merged #4909. Can you rebase this to simplify the diff ?

neutrinoceros (approved):

PR #4624, opened 2023-08-08
TST: stop using conda on Windows CI

neutrinoceros:
## PR Summary

This is made possible by #4621
I tested it on my fork and saw that testing time for windows dropped from ~30min to ~20, with *less* tests being skipped.

PR #4912, opened 2024-05-29
TST: cleanup unnecessary warning filter

neutrinoceros:
## PR Summary

I am now almost sure that the comment I wrote for this filter (circa 2021) was ill-informed and misleading. Astropy's wheels in fact were already being built in a forward-compatible way with `oldest-supported-numpy` at the time and continue to be (now using the more modern mechanism introduced in numpy 1.25). I suspect the real reason why we were seeing this warning was that we installed numpy from conda and then astropy from PyPI, which we haven't been doing since https://github.com/yt-project/yt/pull/4624

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -161,19 +161,6 @@ def pytest_configure(config):
                 ":DeprecationWarning",
             )
 
-    if find_spec("astropy") is not None:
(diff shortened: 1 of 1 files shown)


chrishavlin (approved):
lgtm

PR #4913, opened 2024-05-29
BLD: run pytest on wheels and sdist

yut23:
## PR Summary

The changes to pyproject.toml are needed to allow running the tests on the installed version of the code. The `--ignore` paths are resolved relative to the current directory before being compared to the test files if they're not already absolute. This works fine when running from the repository root, but pytest would have to be run from `site-packages/` to get the same relative file paths and it would try to run the tests of every installed package.

`--rootdir .` makes pytest collect the tests from the installed package instead of the parent directory of the configuration file.

## PR Checklist


neutrinoceros (approved):
Thank you for continuing this thread ! Obviously jobs are longer now: builds complete in about 45min (wall time) instead of 15, but that's a perfectly reasonable amount of time to wait for a release IMO. I'm approving now, but I'll give some time (say a week or two) for others to review.

chrishavlin (approved):
thanks for your work on this, @yut23 !

PR #4915, opened 2024-05-30
add check for Sub_conventions for CFradial

chrishavlin:
This PR adjusts the cf_radial frontend `_is_valid` to look in an additional attribute (`Sub_conventions`). 

NEXRAD (https://cmr.earthdata.nasa.gov/search/concepts/C2020894988-GHRC_DAAC.html) uses CFradial but has slightly different metadata from what we're checking for. In this case it's got the following attributes:

```
Conventions :
    CF-1.7
Sub_conventions :
    CF-Radial instrument_parameters radar_parameters radar_calibration
```

### DIFF
--- a/yt/frontends/cf_radial/data_structures.py
+++ b/yt/frontends/cf_radial/data_structures.py
@@ -310,6 +310,9 @@ class CFRadialDataset(Dataset):
             nc4_file = NetCDF4FileHandler(filename)
             with nc4_file.open_ds(keepweakref=True) as ds:
                 con = "Conventions"  # the attribute to check for file conventions
+                # note that the attributes here are potentially space- or
+                # comma-delimited strings, so we concatenate a single string
+                # to search for a substring.
                 cons = ""  # the value of the Conventions attribute
                 for c in [con, con.lower(), "Sub_" + con.lower()]:
                     if hasattr(ds, c):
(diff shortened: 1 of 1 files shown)


yt/frontends/cf_radial/data_structures.py
```diff
@@ -311,10 +311,10 @@ def _is_valid(cls, filename: str, *args, **kwargs) -> bool:
             with nc4_file.open_ds(keepweakref=True) as ds:
                 con = "Conventions"  # the attribute to check for file conventions
                 cons = ""  # the value of the Conventions attribute
-                for c in [con, con.lower()]:
+                for c in [con, con.lower(), "Sub_" + con.lower()]:
                     if hasattr(ds, c):
                         cons += getattr(ds, c)
-                is_cfrad = "CF/Radial" in cons
+                is_cfrad = "CF/Radial" in cons or "CF-Radial" in cons
```
neutrinoceros, line 317: while your patch seems uncontroversial, I'm having trouble understand the existing implementation: why is `cons` constructed as a concatenated `str` instead of a `list[str]` ?

chrishavlin: I'll double check but I think it's because those attribute values are space-delimted strings. e.g., "CF-Radial instrument_parameters radar_parameters radar_calibration" is actually ```'CF-Radial instrument_parameters radar_parameters radar_calibration'``` so you have to search for a substring. I'll confirm that and add a comment in the code here (or update the code if I'm wrong).

chrishavlin: OK: confirmed just now, and pushed up a comment so that it's clearer in the future

neutrinoceros (approved):

PR #4921, opened 2024-06-04
SWIFT frontend update

smsutherland:
## PR Summary
The SWIFT frontend is very out of date. The last significant commits are many years old, and in that time SWIFT has undergone breaking changes. This PR aims to update some of the breaks. This is by no means an exhaustive fix. It only addresses the issues I have encountered during my own use of yt with SWIFT. This includes loading SWIFT snapshots, and some issues regarding field naming.

## PR Checklist

No new features have been added.

I have not added any new tests, since they would require adding new sample data to test from. Both snapshots on https://yt-project.org/data/ labeled SWIFT are run on an outdated version of SWIFT from 2018.

### DIFF
--- a/yt/frontends/swift/data_structures.py
+++ b/yt/frontends/swift/data_structures.py
@@ -2,10 +2,11 @@ import numpy as np
 
 from yt.data_objects.static_output import ParticleFile
 from yt.frontends.sph.data_structures import SPHDataset, SPHParticleIndex
-from yt.frontends.sph.fields import SPHFieldInfo
 from yt.funcs import only_on_root
 from yt.utilities.logger import ytLogger as mylog
 from yt.utilities.on_demand_imports import _h5py as h5py
+
+from .fields import SwiftFieldInfo
 
 
 class SwiftParticleFile(ParticleFile):
@@ -15,7 +16,7 @@ class SwiftParticleFile(ParticleFile):
 class SwiftDataset(SPHDataset):
     _load_requirements = ["h5py"]
     _index_class = SPHParticleIndex
-    _field_info_class = SPHFieldInfo
+    _field_info_class = SwiftFieldInfo
     _file_class = SwiftParticleFile
 
     _particle_mass_name = "Masses"
--- a/yt/frontends/swift/data_structures.py
+++ b/yt/frontends/swift/data_structures.py
@@ -97,7 +97,6 @@ class SwiftDataset(SPHDataset):
         # Read from the HDF5 file, this gives us all the info we need. The rest
         # of this function is just parsing.
         header = self._get_info_attributes("Header")
-        runtime_parameters = self._get_info_attributes("RuntimePars")
 
         policy = self._get_info_attributes("Policy")
         # These are the parameterfile parameters from *.yml at runtime
@@ -113,7 +112,7 @@ class SwiftDataset(SPHDataset):
         self.dimensionality = int(header["Dimension"])
 
         # SWIFT is either all periodic, or not periodic at all
-        periodic = int(runtime_parameters["PeriodicBoundariesOn"])
+        periodic = int(parameters["InitialConditions:periodic"])
 
         if periodic:
             self._periodicity = [True] * self.dimensionality
@@ -131,7 +130,9 @@ class SwiftDataset(SPHDataset):
                 self.current_redshift = float(header["Redshift"])
                 # These won't be present if self.cosmological_simulation is false
                 self.omega_lambda = float(parameters["Cosmology:Omega_lambda"])
-                self.omega_matter = float(parameters["Cosmology:Omega_m"])
+                self.omega_matter = float(parameters["Cosmology:Omega_b"]) + float(
+                    parameters["Cosmology:Omega_cdm"]
+                )
                 # This is "little h"
                 self.hubble_constant = float(parameters["Cosmology:h"])
             except KeyError:
@@ -155,7 +156,6 @@ class SwiftDataset(SPHDataset):
         # Store the un-parsed information should people want it.
         self.parameters = {
             "header": header,
-            "runtime_parameters": runtime_parameters,
             "policy": policy,
             "parameters": parameters,
             "hydro": hydro,
--- a/yt/frontends/swift/fields.py
+++ b/yt/frontends/swift/fields.py
@@ -1,0 +1,30 @@
+from yt.frontends.sph.fields import SPHFieldInfo
+from yt._typing import KnownFieldsT
+
+
+class SwiftFieldInfo(SPHFieldInfo):
+    known_particle_fields: KnownFieldsT = (
+        ("Coordinates", ("code_length", ["particle_position"], None)),
+        ("Velocities", ("code_velocity", ["particle_velocity"], None)),
+        ("ParticleIDs", ("", ["particle_index"], None)),
+        ("Masses", ("code_mass", ["particle_mass"], None)),
+        ("InternalEnergy", ("code_specific_energy", ["specific_thermal_energy"], None)),
+        ("Density", ("code_mass / code_length**3", ["density"], None)),
+        ("SmoothingLength", ("code_length", ["smoothing_length"], None)),
+    )
+
+    def __init__(self, ds, field_list, slice_info=None):
+        super().__init__(ds, field_list, slice_info)
+
+    def setup_particle_fields(self, ptype, *args, **kwargs):
+        super().setup_particle_fields(ptype, *args, **kwargs)
+
+        if ptype in ("PartType0", "Gas"):
+            self.setup_gas_particle_fields(ptype)
+
+    def setup_gas_particle_fields(self, ptype):
+        self.alias((ptype, "temperature"), (ptype, "Temperatures"))
+        self.alias(("gas", "temperature"), (ptype, "Temperatures"))
+
+        for ax in ("x", "y", "z"):
+            self.alias((ptype, ax), (ptype, "particle_position_" + ax))
--- a/yt/frontends/swift/fields.py
+++ b/yt/frontends/swift/fields.py
@@ -1,5 +1,5 @@
-from yt.frontends.sph.fields import SPHFieldInfo
 from yt._typing import KnownFieldsT
+from yt.frontends.sph.fields import SPHFieldInfo
 
 
 class SwiftFieldInfo(SPHFieldInfo):
@@ -7,10 +7,13 @@ class SwiftFieldInfo(SPHFieldInfo):
         ("Coordinates", ("code_length", ["particle_position"], None)),
         ("Velocities", ("code_velocity", ["particle_velocity"], None)),
         ("ParticleIDs", ("", ["particle_index"], None)),
-        ("Masses", ("code_mass", ["particle_mass"], None)),
-        ("InternalEnergy", ("code_specific_energy", ["specific_thermal_energy"], None)),
-        ("Density", ("code_mass / code_length**3", ["density"], None)),
-        ("SmoothingLength", ("code_length", ["smoothing_length"], None)),
+        ("Mass", ("code_mass", ["particle_mass", "mass"], None)),
+        (
+            "InternalEnergies",
+            ("code_specific_energy", ["specific_thermal_energy"], None),
+        ),
+        ("Densities", ("code_mass / code_length**3", ["density"], None)),
+        ("SmoothingLengths", ("code_length", ["smoothing_length"], None)),
     )
 
     def __init__(self, ds, field_list, slice_info=None):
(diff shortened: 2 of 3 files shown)


neutrinoceros:
Thanks for working on this ! Just a very broad question for now, but are you *sure* this isn't an exhaustive fix, or just strongly suspecting ?

chrishavlin:
nice! this may be answered by the full test suite... but any worry some of these changes would break backwards compatibility with old swift files?

neutrinoceros:
I haven't looked in detail yet but yes, backward compatibility concerns should be a essential concern to reviewers.

smsutherland:
> Thanks for working on this ! Just a very broad question for now, but are you _sure_ this isn't an exhaustive fix, or just strongly suspecting ?

I strongly suspect it is not exhaustive. I am very new to using yt. I am not familiar enough with all its aspects to know for certain one way or another. I only suspect it isn't exhaustive because it seems seems more likely than the alternative.

yt/frontends/swift/io.py
```diff
@@ -66,7 +66,7 @@ def _get_smoothing_length(self, sub_file, pdtype=None, pshape=None):
             pcount = f["/Header"].attrs["NumPart_ThisFile"][ind].astype("int64")
             pcount = np.clip(pcount - si, 0, ei - si)
             # we upscale to float64
-            hsml = f[ptype]["SmoothingLength"][si:ei, ...]
+            hsml = f[ptype]["SmoothingLengths"][si:ei, ...]
```
neutrinoceros, line 69: Here we'd probably need to try both keys for backward compatibility. It's okay to give priority to the newest one but it would also help to add a comment specifying (at least broadly) which version of SWIFT made the change.

yt/frontends/swift/data_structures.py
```diff
@@ -15,7 +16,7 @@ class SwiftParticleFile(ParticleFile):
 class SwiftDataset(SPHDataset):
     _load_requirements = ["h5py"]
     _index_class = SPHParticleIndex
-    _field_info_class = SPHFieldInfo
+    _field_info_class = SwiftFieldInfo
```
neutrinoceros, line 19: not clear to me what are the implications of this change in terms of backward compatibility

yt/frontends/swift/data_structures.py
```diff
@@ -113,7 +113,7 @@ def _parse_parameter_file(self):
         self.dimensionality = int(header["Dimension"])
 
         # SWIFT is either all periodic, or not periodic at all
-        periodic = int(runtime_parameters["PeriodicBoundariesOn"])
+        periodic = int(parameters["InitialConditions:periodic"])
```
neutrinoceros, line 116: Here too, we should really be checking both for new and old names.

yt/frontends/swift/data_structures.py
```diff
@@ -131,7 +131,9 @@ def _parse_parameter_file(self):
                 self.current_redshift = float(header["Redshift"])
                 # These won't be present if self.cosmological_simulation is false
                 self.omega_lambda = float(parameters["Cosmology:Omega_lambda"])
-                self.omega_matter = float(parameters["Cosmology:Omega_m"])
+                self.omega_matter = float(parameters["Cosmology:Omega_b"]) + float(
+                    parameters["Cosmology:Omega_cdm"]
+                )
```
neutrinoceros, line 136: Is this a bug fix or is the `Cosmology:Omega_cdm` key new in some version of SWIFT ? If old versions don't have it, we need to fallback to the old value somehow.

yt/frontends/swift/data_structures.py
```diff
@@ -97,7 +98,6 @@ def _parse_parameter_file(self):
         # Read from the HDF5 file, this gives us all the info we need. The rest
         # of this function is just parsing.
         header = self._get_info_attributes("Header")
-        runtime_parameters = self._get_info_attributes("RuntimePars")
```
neutrinoceros, line 100: This variable should still be defined when `"RuntimePars"` is available

yt/frontends/swift/data_structures.py
```diff
@@ -155,7 +157,6 @@ def _parse_parameter_file(self):
         # Store the un-parsed information should people want it.
         self.parameters = {
             "header": header,
-            "runtime_parameters": runtime_parameters,
```
neutrinoceros, line 158: this key/value pair should still exist when possible

neutrinoceros (changes requested):
It seems to me that we indeed need to refine the patch for backward compatibility, but other than that I have no concerns with it. Thank you !

chrishavlin:
> I have not added any new tests, since they would require adding new sample data to test from. Both snapshots on https://yt-project.org/data/ labeled SWIFT are run on an outdated version of SWIFT from 2018.

Do you have any new sample data we could add? Would be great if possible, I'm happy to help get it where it needs to go so we can run tests against it.

smsutherland:
> > I have not added any new tests, since they would require adding new sample data to test from. Both snapshots on https://yt-project.org/data/ labeled SWIFT are run on an outdated version of SWIFT from 2018.
> 
> Do you have any new sample data we could add? Would be great if possible, I'm happy to help get it where it needs to go so we can run tests against it.

The only SWIFT snapshots I have are many gigabytes. I can contribute one if you'd like (it's better than nothing I suppose), but a file of that size may be impractical for testing purposes or as a sample.
For reference, the EAGLE_6 sample data already present is a (6.25Mpc)^3 volume and 71Mb. The snapshot I was basing my changes on is (37Mpc)^3 and 6.6Gb.

neutrinoceros:
watchout, existing tests are failing. The good news is that the test suite will apparently not let us get away with backward incompatible changes after all

chrishavlin:
> The only SWIFT snapshots I have are many gigabytes.

Ya, that is a common issue with keeping yt's test data up to date. Given that the changes here are pretty minimal, I'd say not worth adding a file that large for the changes here.

yt/frontends/swift/data_structures.py
```diff
@@ -97,7 +98,12 @@ def _parse_parameter_file(self):
         # Read from the HDF5 file, this gives us all the info we need. The rest
         # of this function is just parsing.
         header = self._get_info_attributes("Header")
-        runtime_parameters = self._get_info_attributes("RuntimePars")
+        # RuntimePars were removed from snapshots at SWIFT commit 6271388
+        with h5py.File(self.filename, mode="r") as handle:
+            has_runtime_pars = "RuntimePars" in handle.keys()
+
+        if has_runtime_pars:
+            runtime_parameters = self._get_info_attributes("RuntimePars")
```
neutrinoceros, line 106: I'd rather we avoid possibly unbound names

```suggestion
        if has_runtime_pars:
            runtime_parameters = self._get_info_attributes("RuntimePars")
        else:
            runtime_parameters = {}
```

yt/frontends/swift/io.py
```diff
@@ -65,8 +65,13 @@ def _get_smoothing_length(self, sub_file, pdtype=None, pshape=None):
         with h5py.File(sub_file.filename, mode="r") as f:
             pcount = f["/Header"].attrs["NumPart_ThisFile"][ind].astype("int64")
             pcount = np.clip(pcount - si, 0, ei - si)
+            keys = f[ptype].keys()
+            # SWIFT commit a94cc81 changed from "SmoothingLength" to "SmoothingLengths"
```
neutrinoceros, line 69: does SWIFT have actual version numbers too ? if so, I'd suggest documenting the versions before and after this commit. Otherwise, that's probably the best we can do.

yt/frontends/swift/data_structures.py
neutrinoceros: Anyone want to comment here ? For the record, I don't think it's critical enough to hold back the PR, but I'd like an answer if possible.

neutrinoceros (commented):
It's gotten a lot better, thank you !
Still a couple questions and suggestions, but we're definitely on the right track !

yt/frontends/swift/data_structures.py
smsutherland: Should we then also always add (the potentially empty) runtime_parameters to `self.parameters`?

neutrinoceros: I guess it wouldn't hurt, especially if it simplifies the logic.

neutrinoceros (approved):
That's one approval from me.
If nobody complains about it I'll merge it in a week or two. Thanks again for your work and patience !

smsutherland:
And thank you for your help with this!

yt/frontends/swift/fields.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def setup_particle_fields(self, ptype, *args, **kwargs):
+        super().setup_particle_fields(ptype, *args, **kwargs)
+
+        if ptype in ("PartType0", "Gas"):
+            self.setup_gas_particle_fields(ptype)
+
+    def setup_gas_particle_fields(self, ptype):
+        self.alias((ptype, "temperature"), (ptype, "Temperatures"))
+        self.alias(("gas", "temperature"), (ptype, "Temperatures"))
```
chrishavlin, line 24: one more backward compatibility concern -- as written, wouldn't this would always alias `(ptype, "temperature")` to `(ptype, "Temperatures")` even for the older swift datasets (which will not exist)? I'm not positive this is an issue... but it looks like the existing tests only utilize the `("gas", "density")` field, so the test suite wouldn't catch any errors from this (and I'm on the wrong computer to test it manually right now...). 

I think adding a check for the existence of the field similar to what the base `setup_particle_fields` does would be the way to go: 

https://github.com/yt-project/yt/blob/13dc4fbcba564439c90e9bdd8db013f797dfc9a9/yt/fields/field_info_container.py#L106-L107

The position aliasing below should be safe.

chrishavlin:
@yt-fido test this please

chrishavlin:
re-triggered the tests because the last run skipped 333 tests due to not finding files... if this is due to yt-project.org being down, those same tests will be skipped again and we should re-trigger after yt-project is back up.

yt/frontends/swift/fields.py
chrishavlin: note: I'm back at a computer with test data, I'm not actually sure my comment above is an issue. will update here in a bit...

chrishavlin: oooook, this is **not** an issue, because the first couple lines of the `alias` function check for the existence of the field that you're trying to create the alias for.

yt/frontends/swift/data_structures.py
chrishavlin: Someone with more of an expertise in swift data might want to chime in, but I just spent a while walking through the initialization and aliasing that happens and I don't see any issues. It will still check for all the fields defined by `SPHFieldInfo`, and the additional aliasing that `SwiftFieldInfo` adds will only happen if the new fields exist within the dataset.

chrishavlin (approved):
LGTM! thanks @smsutherland !

chrishavlin:
> If nobody complains about it I'll merge it in a week or two.

Even with the 2 approvals now, I think it'd still be good to wait a bit to see if anyone with more expertise in swift wants to take a look.

yt/frontends/swift/fields.py
smsutherland: Glad to see this is not an issue. Aliasing like this felt less like a proper solution and more like a hack to get it to work.

neutrinoceros:
Fine by me. I'll post-pone my personal reminder to merge this in 2 weeks from *now*.

rennehan:
How was the specific internal energy cosmology handled for this pull request? I see that the Swift frontend still uses the setup_gas_particle_fields() function, which only treats InternalEnergy/InternalEnergies as a physical quantity. In Swift, the internal unit of InternalEnergies is really code_velocity^2 which has a scale factor a^2 inside of it.

smsutherland:
Running a quick test, it looks like yes, internal energies aren't properly handled. The internal energies seem to scale with a^2, rather than a^-2. This is one of a number of issues I've found with the units of swift datasets (for example, temperatures are reported as being dimensionless), so perhaps a follow-up PR is necessary to iron out some of these issues.

rennehan:
Yes, Swift has _consistent_ internal units compared to Gadget/GIZMO which have _inconsistent_ units. The cosmological factors are here: https://github.com/SWIFTSIM/SWIFT/blob/9c17498b4971a43ce57b94629ff0bb1d66e179a8/src/cosmology.c#L224

and the details of why it is consistent compared to Gadget/GIZMO is (derived) here:

https://github.com/SWIFTSIM/SWIFT/blob/9c17498b4971a43ce57b94629ff0bb1d66e179a8/theory/Cosmology/coordinates.tex

PR #4926, opened 2024-06-13
BUG: fix ParamFileStore for on-demand frontend imports

chrishavlin:
Currently, the `Dataset` pickling is broken between python sessions when using `store_parameter_files = true`. For example, given two scripts, `create_a_pickle.py` and `load_a_pickle.py`:

```python
# create_a_pickle.py
import yt
import pickle

if __name__ == '__main__':

    fname = 'IsolatedGalaxy/galaxy0030/galaxy0030'
    ds = yt.load(fname)

    with open('yt_ds.pickle', 'wb') as handle:
        pickle.dump(ds, handle, protocol=pickle.HIGHEST_PROTOCOL)
```

```python
# load_a_pickle.py
import pickle

if __name__ == '__main__':

    with open('yt_ds.pickle', 'rb') as handle:
        ds = pickle.load(handle)
   _ =  ds.index

```
The following fails on the attempt to load:
```
$ python create_a_pickle.py
$ python load_a_pickle.py
  File "/home/chavlin/src/yt_general/performance/yt_perf_testing/miscelleaneous_tests/dask_pickles_etc/pickle_from_file.py", line 5, in <module>
    ds = pickle.load(handle)
  File "/home/chavlin/src/yt_general/yt/yt/data_objects/static_output.py", line 2053, in _reconstruct_ds
    ds = datasets.get_ds_hash(*args)
  File "/home/chavlin/src/yt_general/yt/yt/utilities/parameter_file_storage.py", line 91, in get_ds_hash
    return self._convert_ds(self._records[hash])
  File "/home/chavlin/src/yt_general/yt/yt/utilities/parameter_file_storage.py", line 117, in _convert_ds
    raise UnknownDatasetType(class_name)
yt.utilities.parameter_file_storage.UnknownDatasetType: EnzoDataset
```
(I also ran into this problem when experimenting with some dask functionality...). 

The issue comes how yt now imports frontends on demand (`output_type_registry` is reset between sessions).

### DIFF
--- a/yt/utilities/parameter_file_storage.py
+++ b/yt/utilities/parameter_file_storage.py
@@ -8,7 +8,16 @@ from yt.utilities.parallel_tools.parallel_analysis_interface import (
     parallel_simple_proxy,
 )
 
-_field_names = ("hash", "bn", "fp", "tt", "ctid", "class_name", "last_seen")
+_field_names = (
+    "hash",
+    "bn",
+    "fp",
+    "tt",
+    "ctid",
+    "class_name",
+    "module_name",
+    "last_seen",
+)
 
 
 class NoParameterShelf(Exception):
@@ -104,6 +113,7 @@ class ParameterFileStore:
             "tt": ds.current_time,
             "ctid": ds.unique_identifier,
             "class_name": ds.__class__.__name__,
+            "module_name": ds.__module__,
             "last_seen": ds._instantiated,
         }
 
@@ -114,7 +124,17 @@ class ParameterFileStore:
         fn = os.path.join(fp, bn)
         class_name = ds_dict["class_name"]
         if class_name not in output_type_registry:
-            raise UnknownDatasetType(class_name)
+            # first check that the relevant frontend is imported:
+            # the output_type_registry is refreshed between sessions as
+            # frontends are imported on-demand now.
+            import importlib
+
+            module = ds_dict["module_name"]
+            fe_api = ".".join(module.split(".")[0:3]) + ".api"
+            _ = importlib.import_module(fe_api)
+            # if it is **still** not in the registry, error
+            if class_name not in output_type_registry:
+                raise UnknownDatasetType(class_name)
         mylog.info("Checking %s", fn)
         if os.path.exists(fn):
             ds = output_type_registry[class_name](os.path.join(fp, bn))


chrishavlin:
Setting to draft for now because I want to see if I can write a test with some pytest monkeypatching. 

Also note that the fix here is not backwards compatible -- IMO it's fine because I don't think many folks are using this functionality and it's not meant for long-term dataset storage anyway.

yt/utilities/parameter_file_storage.py
```diff
@@ -8,7 +8,16 @@
...
 
-_field_names = ("hash", "bn", "fp", "tt", "ctid", "class_name", "last_seen")
+_field_names = (
+    "hash",
+    "bn",
+    "fp",
+    "tt",
+    "ctid",
+    "class_name",
+    "module_name",
```
chrishavlin, line 18: this is the potentially backwards incompatible part: when using the on-disk store, this new field will be a new column in the csv. so if an on-disk store already exists, there will be an error due to the different number of columns and the user would need to remove the old on-disk store.

chrishavlin:
oh, those are real failures. will take another go at those pytest fixtures....

yt/utilities/parameter_file_storage.py
```diff
@@ -114,7 +126,17 @@ def _convert_ds(self, ds_dict):
         fn = os.path.join(fp, bn)
         class_name = ds_dict["class_name"]
         if class_name not in output_type_registry:
-            raise UnknownDatasetType(class_name)
+            # first check that the relevant frontend is imported:
+            # the output_type_registry is refreshed between sessions as
+            # frontends are imported on-demand now.
+            import importlib
```
neutrinoceros, line 132: Any reason not to put this import at the top level ? importlib should be "free" to import anyway.

yt/utilities/parameter_file_storage.py
```diff
@@ -114,7 +126,17 @@ def _convert_ds(self, ds_dict):
...
+
+            module = ds_dict["module_name"]
+            fe_api = ".".join(module.split(".")[0:3]) + ".api"
```
neutrinoceros, line 135: Not 100% sure this is equivalent (I don't know how many "." are expected), but I'm willing to bet there's a clearer way to express this change in file extension.

```suggestion
            fe_api = module.rpartition(".")[0] + ".api"
```

yt/utilities/tests/test_parameter_file_storage.py
```diff
@@ -0,0 +1,55 @@
...
+    pfs_init_name = ytcfg.get("yt", "parameter_file_store")
+    pfs_path = tmp_path / "param_file_store.csv"
+    pfs_path.touch()
+    ytcfg.set("yt", "parameter_file_store", str(pfs_path))
+    yield
+    ytcfg.set("yt", "parameter_file_store", pfs_init_name)
+
+
+def test_on_disk_parameter_file_store(
+    monkeypatch, store_parameter_files, set_parameter_file
```
neutrinoceros, line 29: Fixtures that are used as setup/teardown but don't need an accessible name should be declared with `pytest.mark.usefixture`

PR #4927, opened 2024-06-14
Do Not Merge: Checking enzo answer fail with np >= 2

chrishavlin:
This PR is to double check my comment here https://github.com/yt-project/yt/pull/4874#issuecomment-2168530832 for why a couple of enzo answer tests were failing with the latest numpy. This PR should not be merged -- IMO it is more appropriate to bump the answer tests, but this PR will confirm that the change in type promotion with np 2.0 is indeed the cause for the failures. 

This PR branch is off of the branch for https://github.com/yt-project/yt/pull/4874


chrishavlin:
passed! the one failure in the py3.12 windows test is unrelated.

PR #4932, opened 2024-06-22
Add support for Enzo-E simulations with 0 ghost zones

mabruzzo:
## PR Summary

The primary purpose of this PR was to add support for Enzo-E simulations with 0 ghost zones.

While I was here, I added support for Enzo-E simulations where the x, y, and z axes have different numbers of ghost zones (this is a feature that Enzo-E supports, but it's not really used). I also added some explanatory comments.

### DIFF
--- a/yt/frontends/enzo_e/data_structures.py
+++ b/yt/frontends/enzo_e/data_structures.py
@@ -406,8 +406,9 @@ class EnzoEDataset(Dataset):
         ablock = fh[list(fh.keys())[0]]
         self.current_time = ablock.attrs["time"][0]
         self.parameters["current_cycle"] = ablock.attrs["cycle"][0]
-        gsi = ablock.attrs["enzo_GridStartIndex"]  # <- always has 3 elements
+        gsi = ablock.attrs["enzo_GridStartIndex"]
         gei = ablock.attrs["enzo_GridEndIndex"]
+        assert len(gsi) == len(gei) == 3  # sanity check
         # Enzo-E technically allows each axis to have different ghost zone
         # depths (this feature is not really used in practice)
         self.ghost_zones = gsi
--- a/yt/frontends/enzo_e/data_structures.py
+++ b/yt/frontends/enzo_e/data_structures.py
@@ -406,9 +406,12 @@ class EnzoEDataset(Dataset):
         ablock = fh[list(fh.keys())[0]]
         self.current_time = ablock.attrs["time"][0]
         self.parameters["current_cycle"] = ablock.attrs["cycle"][0]
-        gsi = ablock.attrs["enzo_GridStartIndex"]
+        gsi = ablock.attrs["enzo_GridStartIndex"]  # <- always has 3 elements
         gei = ablock.attrs["enzo_GridEndIndex"]
-        self.ghost_zones = gsi[0]
+        # Enzo-E technically allows each axis to have different ghost zone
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Thanks @mabruzzo ! I'll review this tomorrow but I just wanted to point out that you will likely hit a couple unrelated bugs that are solved in #4930 so don't scratch your head too much over failures for now :)

mabruzzo:
> Thanks @mabruzzo ! I'll review this tomorrow but I just wanted to point out that you will likely hit a couple unrelated bugs that are solved in #4930 so don't scratch your head too much over failures for now :)

Gotcha.

FYI this is currently built off of a slightly older point in the main branch. This was mainly for convenience (I was a really hard time setting up yt off of the HEAD of the main-branch. My issues are probably user-error. (I recently switched to a new machine and started using conda for the first time in years). I'm happy to fast-forward this whenever.

neutrinoceros:
> FYI this is currently built off of a slightly older point in the main branch. 

as long as there are no conflicts (which there aren't), this shouldn't matter :)

yt/frontends/enzo_e/data_structures.py
```diff
@@ -406,9 +406,12 @@ def _parse_parameter_file(self):
         ablock = fh[list(fh.keys())[0]]
         self.current_time = ablock.attrs["time"][0]
         self.parameters["current_cycle"] = ablock.attrs["cycle"][0]
-        gsi = ablock.attrs["enzo_GridStartIndex"]
+        gsi = ablock.attrs["enzo_GridStartIndex"]  # <- always has 3 elements
```
neutrinoceros, line 409: Instead of a comment, you could add an `assert` here, which would be less likely to rot silently

yt/frontends/enzo_e/data_structures.py
```diff
@@ -406,9 +406,12 @@ def _parse_parameter_file(self):
...
         gei = ablock.attrs["enzo_GridEndIndex"]
-        self.ghost_zones = gsi[0]
+        # Enzo-E technically allows each axis to have different ghost zone
+        # depths (this feature is not really used in practice)
+        self.ghost_zones = gsi
```
neutrinoceros, line 413: although technically not backward compatible, I don't expect anyone to rely on the old behavior so I'm tempted to just let this change in.

neutrinoceros (commented):
This looks fine to me, just a minor suggestion

yt/frontends/enzo_e/data_structures.py
mabruzzo: Good point! Fixed

yt/frontends/enzo_e/data_structures.py
mabruzzo: I definitely agree, especially because pool of enzo-e users is currently very small. (At most, there are maybe 5 people who regularly use the yt-frontend right now)

But, let me know if you would prefer I change it (and whether there is an obvious preferred way to do it)

neutrinoceros: I think it's fine as is. I'll just give it a couple weeks before merging in case anyone objects.

neutrinoceros (approved):

neutrinoceros:
@yt-fido Test this please

neutrinoceros:
The typing error is unrelated and already fixed on main, so let’s get this in. Thanks again for your patience !

Issue #4787, opened 2024-01-24
yt code_length unit inconsistent with Ramses unit_l


fredt00:
### Bug report

**Bug summary**
I experience the same issue as https://github.com/yt-project/yt/issues/2866: the unit length conversion for a non cosmological simulation is off by a factor of the boxlength.

**Code for reproduction**

```python
print(ds.arr(1,"code_density").in_cgs().value/ds.parameters['unit_d'])
print(ds.arr(1,"code_length").in_cgs().value/ds.parameters['unit_l'])
print(ds.parameters['boxlen'])
```

**Actual outcome**

The length conversion is off by a factor of 150, the value of the box length. However, the density conversion seems to be correct...
```
1.0
150.0
150.0
```

I was wondering if anyone had a simple work around for this? I don't think I can just multiply any results with length units by factors of the box length since the density conversion is correct and includes length**3.

**Version Information**

  * Operating System: macOS Sonoma 14.1.1
  * Python Version: 3.11.6
  * yt version: 4.3.0 (installed via pip)

matthewturk:
Thanks for bringing this up -- I am sorry we didn't address it before!  Perhaps @cphyc can provide some insight?

cphyc:
I'm happy to help but I'm 200% busy until February 9 and won't have time since then. Most likely, these two lines are the right entry points:
https://github.com/yt-project/yt/blob/1e90fd73f165283399dc45d487183a95239196fd/yt/frontends/ramses/data_structures.py#L977 
https://github.com/yt-project/yt/blob/1e90fd73f165283399dc45d487183a95239196fd/yt/frontends/ramses/data_structures.py#L1046

The first sets the length_unit, so `1 * length_unit` equals the box size. The second sets the size of the box.
I wonder if the fix is simply to replace the latter by
```python
        self.domain_right_edge = np.full(3, self.parameters["boxlen"], dtype="float64")
```
Could you try and let us know if this works? If it does, we'd be happy for you to submit a PR, or I can do it in a couple of weeks.

fredt00:
Hi,

Thanks for the quick reply. I tried changing the line you suggested, and this does mean the correct domain_right_edge is returned when loading the data:

```
yt : [INFO     ] 2024-01-24 16:11:33,248 Parameters: domain_right_edge         = [150. 150. 150.]
```

However, the ratio code_length/unit_l is still incorrect by a factor of boxlen and I get the same offset in my plots. 

[...]

Issue #2866, opened 2020-08-15
Ramses data, yt Internal length units are inconsistent with all other units


wadsley:
#2800 # Bug report

**Bug summary**

Ramses has a length unit called unit_l (cgs),   e.g. ds.parameters['unit_l']
This is the correct length unit for ramses that feeds into all derived units (densities, pressures, energies, velocities).
Similarly it has unit_t for time etc...

The yt code_length unit (for converting from code units) does not match this unless the boxlen=1

**Code for reproduction**

Read in any ramses dataset where boxlen !=1, 
e.g. ds=yt.load("output_00001/info_00001.txt")

>>> x=ds.arr(1,"code_length")
>>> x.in_cgs()
 YTArray(9.87424e+21) cm
>>> ds.parameters("unit_l")
 3.0857e+21

These differ by boxlen (in this case 3.2).  
This means that any derived units a user makes from yt's code units are broken and wrong.

e.g.  If you have a new energy variable in code units and want to set it up the following will 
lead to incorrect results.

def _ergperg(field, data):
     return data.ds.arr(data["ergperg_new_raw"],"code_length**2/code_time**2")

ds.add_field(("ramses", "ergperg_new"), sampling_type="cell", function=_ergperg, units="erg/g")

Ramses sometimes uses a unit of boxlen*unit_l for some distance variables.
This is not the code length unit.   yt should recognize that the units of some position data
for ramses datasets is not ramses code length units but fractions of boxlen*unit_l and avoid having
a broken code_length.

Right now there seems to be some kind of internal hack so that units for pressure and so on in yt
and calculated using the Ramses actual length unit ('unit_l') and then the code_length is set to be
inconsistent with that so all those units cannot be reproduced using combinations of code_length,
code_time and code_mass.   This is a nasty gotcha for anyone trying to use yt and ramses.

A general solution is to set the units for position data appropriately, not
to break something so fundamental as the code_length.

cphyc:
Thanks for the report! That should definitely get fixed.

PR #4934, opened 2024-06-26
correct issue ?

Lenoble-lab:
Hello, 
It seems there is an issue with length unit for ramses output. 

The bug has already been identified in [2866](https://github.com/yt-project/yt/issues/2866) and [4787](https://github.com/yt-project/yt/issues/4787). 

Here I propose a modification of two lines in `data_structure`. 

I don't understand why the length_unit is scaled with the boxlen. To me, it should not as then, we loose the units if `boxlen != 1`. The correction induced is to change the yt's boxlen which should be different from 1 if we change the units. 

Thanks,
Romain

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -975,7 +975,7 @@ class RAMSESDataset(Dataset):
         setdefaultattr(self, "temperature_unit", temperature_unit.in_units("K"))
 
         # Only the length unit get scales by a factor of boxlen
-        setdefaultattr(self, "length_unit", self.quan(length_unit * boxlen, "cm"))
+        setdefaultattr(self, "length_unit", self.quan(length_unit, "cm"))
 
     def _parse_parameter_file(self):
         # hardcoded for now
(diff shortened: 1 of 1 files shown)


chrishavlin:
Thanks for the PR @Lenoble-lab ! Looks OK to me but would be good to have eyes on this from someone familiar with Ramses -- @cphyc are you available for a quick look?

The failing 3.12 test is unrelated, but the failing pre-commit check will need to be fixed manually:

```
yt/frontends/ramses/data_structures.py:949:9: F841 Local variable `boxlen` is assigned to but never used
```

cphyc:
I am still not 100% sure this is what we want, but I'm failing to explain why :)
I am also unhappy that this significantly-breaking change didn't trigger any test failure. That means we've never tested that the units were correct for non-cosmological datasets. Especially, I just tested locally, and it seems that your fix doesn't solve 100% of the issue. For example
```python
In [5]: ds = yt.load_sample("DICEGalaxyDisk_nonCosmological")

In [6]: ds.r["x"].to("code_length").max()   # Duh?
Out[6]: unyt_quantity(1.7578125, 'code_length')

In [7]: ds.domain_right_edge.to("kpc")
Out[7]: unyt_array([150.00000002, 150.00000002, 150.00000002], 'kpc')

In [8]: ds.r["x"].to("kpc").max()
Out[8]: unyt_quantity(1.7578125, 'kpc')
```
However, on 13dc4fbcba564439c90e9bdd8db013f797dfc9a (the main branch), I get the following:
```python
In [1]: ds = yt.load_sample("DICEGalaxyDisk_nonCosmological")

In [2]: ds.r["x"].to("kpc").max()
Out[2]: unyt_quantity(149.41406252, 'kpc')

In [3]: ds.domain_right_edge.to("kpc")
Out[3]: unyt_array([150.00000002, 150.00000002, 150.00000002], 'kpc')
```
which seems like a better solution :)

Issue #4788, opened 2024-01-25
bug in Projection Plots for SPH datasets


nastasha-w:
### Bug report

**Bug summary**

To my understanding, the pixel values in projection plots should represent the integral of some field over the line of sight through a pixel center. That means that pixels in different grids, but with the same center, should produce the same integrated values. However, this is not the case for the `FixedResolutionBuffer` projections of SPH datasets. I suspect this is due to the handling of small SPH particles (relative to the grid spacing) in the backend code: their contribution to the line-of-sight integrals seem to be systematically overestimated.

I've proposed a fix, but I'd like someone else to confirm whether this problem is real and seems correctly diagnosed before I open a pull request.

**Bug explanation**

Here, I will explain the issue as I understand it from the backend code. I will demonstrate the problem in the next section.

The backend code at issue here is `pixelize_sph_kernel_projection` in [pixelization_routines.pyx](https://github.com/yt-project/yt/blob/main/yt/utilities/lib/pixelization_routines.pyx), starting on line 1121. This code loops over the input particles, determines the extent of the region each particle might contribute to, and adds the (kernel-integrated path length) * (selected particle field) to each pixel's total.  In #4783, I discuss some background on the reasoning for this method; the pixel values should be the same as the $\sum_{j} dl_j \cdot A_j$, along a `Ray`, where the sum is over SPH particles $j$, $dl_j$ is the path length from e.g., #4783, and $A$ is the particle field to be projected.
For the column density or surface density $N$ along one of those pencil beams, we then get
```math
N \approx \sum_{j=1}^{N'} A_j \cdot (m_j / (\rho_j \cdot h_j^2)) \cdot  \left[ \int dr \, U \left( \sqrt{(b_i/h_i)^2 + r^2}\right) \right],
```
and 
```math
dl_j = (m_j / (\rho_j \cdot h_j^2)) \cdot  \left[ \int dr \, U \left( \sqrt{(b_j/h_j)^2 + r^2} \right) \right].
```
For a particle $j$, $h_j$ is its smoothing length, $m_j$ is its mass, $\rho_j$ is its density, and $b_j$ is its impact parameter: its minimum distance to the line of sight.

[...]

PR #4939, opened 2024-07-04
Sph projections, slices, gridding: backend and off-axis

nastasha-w:
Fixes issues in the SPH projection backend and adds off-axis projection for SPH particles.

## PR Summary

The current version removes the addition of small SPH particles to column densities along lines of sight those particles do not intersect. Additionally, this version supports off-axis projections for SPH datasets.
Similar issues in slice plots and grid value interpolation (scatter method) for SPH datasets are still TODO. Tests are also currently in a separate script [sph_pixelization_tests_2024-07-03.py.zip](https://github.com/user-attachments/files/16092745/sph_pixelization_tests_2024-07-03.py.zip), so I can run the tests on the main and issue branches. 

This addresses some of the points in issue #4788 ; it should eventually address all of them. 

## PR Checklist

- [x] Fix issues in axis-aligned slice plots for SPH datasets
- [x] Add off-axis slice plots for SPH datasets
- [x] Fix issues in 'scatter'-method gridding of SPH data
- [x] Merge up-to-date main branch into this one
- [x] Add tests to the yt codebase.


nastasha-w:
Some nice test images for the off-axis projection: the `gizmo_mhd_mwdisk` projected along the x- and z-axes with axis-aligned projections:
[sph_projection_test_lineofsight_x.pdf](https://github.com/user-attachments/files/16092786/sph_projection_test_lineofsight_x.pdf)
[sph_projection_test_lineofsight_z.pdf](https://github.com/user-attachments/files/16092788/sph_projection_test_lineofsight_z.pdf)
and that same galaxy projected along the (marginally-off) x and z axes, going from x to z in nine steps. Here, unlike for the axis-aligned plots, there is an option to set the image y-direction (`north_vector`), which in this case is always the simulation y-axis: 
[sph_projection_test_lineofsight_step0of9.pdf](https://github.com/user-attachments/files/16092805/sph_projection_test_lineofsight_step0of9.pdf)
[sph_projection_test_lineofsight_step1of9.pdf](https://github.com/user-attachments/files/16092806/sph_projection_test_lineofsight_step1of9.pdf)
[sph_projection_test_lineofsight_step2of9.pdf](https://github.com/user-attachments/files/16092807/sph_projection_test_lineofsight_step2of9.pdf)
[sph_projection_test_lineofsight_step3of9.pdf](https://github.com/user-attachments/files/16092809/sph_projection_test_lineofsight_step3of9.pdf)
[sph_projection_test_lineofsight_step4of9.pdf](https://github.com/user-attachments/files/16092811/sph_projection_test_lineofsight_step4of9.pdf)
[sph_projection_test_lineofsight_step5of9.pdf](https://github.com/user-attachments/files/16092812/sph_projection_test_lineofsight_step5of9.pdf)
[sph_projection_test_lineofsight_step6of9.pdf](https://github.com/user-attachments/files/16092813/sph_projection_test_lineofsight_step6of9.pdf)
[sph_projection_test_lineofsight_step7of9.pdf](https://github.com/user-attachments/files/16092815/sph_projection_test_lineofsight_step7of9.pdf)
[sph_projection_test_lineofsight_step8of9.pdf](https://github.com/user-attachments/files/16092825/sph_projection_test_lineofsight_step8of9.pdf)

I haven't added these as formal answer tests, it's just a visual verification that the outcomes make sense on real data.

neutrinoceros:
@nastasha-w looks like there are a couple defects at the moment. If you need a hand to get CI green let me know !

nastasha-w:
Thanks @neutrinoceros! I think the main issue now is that I branched this off main like 6 months ago, so I'll try merging the lastest version in first. I do also think one or a few answer tests from the previous version *should* fail: these use the old SPH axis-aligned projections.

neutrinoceros:
> I think the main issue now is that I branched this off main like 6 months ago

Nothing a rebase can't fix. It shouldn't be risky actually since you don't have any conflicts.

nastasha-w:
pre-commit.ci autofix

nastasha-w:
pre-commit.ci autofix

nastasha-w:
pre-commit.ci autofix

nastasha-w:
@neutrinoceros , could you take a look at workflow approval for this PR? I've been submitting a lot of commits recently and I suspect it stopped running everything automatically for that reason. Basically, I've been wrestling with the type-checker a lot today and yesterday.
I think I've got the main things working now, but I was also getting some failures on new tests I added due to divide-by-zero warnings. (The zero division occurs in results I'm calculating to compare to the main yt calculations, and I explicitly check for those situations before making the comparison.)

neutrinoceros:
Plot twist: it never ran automatically, it's always me manually approving. 😅
You need to have at least one commit already in the history for CI to run automatically. Maybe you could issue another much smaller PR that we can merge quickly ?

neutrinoceros:
I took the liberty of pushing to your branch to resolve 2 small bugs that caused a few hundreds tests to fail. I'm hoping to get a much clearer picture of what's still broken after that.

neutrinoceros:
On the github actions side we're left with 81 failures that all seem to have one root cause: a divide-by-zero error. Let me know if you need a hand there too, otherwise I'll let you fix it.

nastasha-w:
@neutrinoceros Thanks for the help! I think I've got those divide-by-zero errors now. I'll try to follow the instructions on the website for updating the image tests.

nastasha-w:
pre-commit.ci autofix

jzuhone:
@nastasha-w just a note that I have been keeping track of this and I plan on reviewing it once this has been sorted out.

nastasha-w:
@jzuhone Thanks! It feels like I'm getting close, but then it's felt like that for a few days already

nastasha-w:
pre-commit.ci autofix

nastasha-w:
@neutrinoceros could you help me with the image tests? Under the full nose test suite, the only failing tests now seem to be comparisons of previously generated images of SPH projections to the figures currently produced. Changes are expected there because I changed that backend. I have looked into the instructions in the docs for updating these tests [here](https://yt-project.org/docs/dev/developing/testing.html#creating-and-updating-image-baselines-for-pytest-mpl-tests), but these don't seem to work for me. I have downloaded the image overview from the “Build and Tests” job summary page, but it seems like those don't cover the "full testsuite (nose)" tests where I'm getting the errors. This section [here](https://yt-project.org/docs/dev/developing/testing.html#how-to-run-the-answer-tests) suggests updating the `answer_version` attribute of the relevant test class (in this case `PixelizedParticleProjectionValuesTest`, if I'm not mistaken) should work, but it isn't clear to me what the current test version is or how to increase it, since that attribute is only mentioned in the codebase in a few places: the docs I mentioned, a method in `conftest.py` getting stored and new file names, and one class (`TestGDF`) that seems to define this attribute explicitly. Could you let me know how to update the baseline for these image tests?
Also, I don't know why the `docs` test is failing; it passed the last test suite and I didn't change anything about the documentation, or any docstrings, since then. I am seeing a "Jenkins is stupid" file though; is there some issue in the testing process here, or do I need to do some digging?

neutrinoceros:
> I have looked into the instructions in the docs for updating these tests [here](https://yt-project.org/docs/dev/developing/testing.html#creating-and-updating-image-baselines-for-pytest-mpl-tests), but these don't seem to work for me.

Sorry if this is confusing: these are the instruction for updating *pytest*-based image tests; however, Jenkins is still running nosetest for tests that require real (heavy) datasets.

> This section [here](https://yt-project.org/docs/dev/developing/testing.html#how-to-run-the-answer-tests) suggests updating the answer_version attribute of the relevant test class

that's even worse... I believe this part of our documentation was never actually in sync with our practice (it was introduced as part of a substantial but incomplete effort to migrate our test suite to pytest). Disregard it entirely.


So the actual way we update answers with nose is by bumping version numbers in `tests/tests.yaml`, which triggers a regeneration of baselines on Jenkins.
However, it is customary to not do it until a reviewer explicitly approve of it, so the next logical step is for someone to have a look and check that all changes are both expected and desired. I have only looked at a couple myself and I'll admit that the new images are not clearly better to my eye, and I'm not even sure that I can discard the possibility of a bug.
For instance here's one:
### before (as of yt 4.3)
before

### after (as of this branch)
after


can you explain what's going on ?

neutrinoceros:
And about docs builds:

> Also, I don't know why the docs test is failing; it passed the last test suite and I didn't change anything about the documentation, or any docstrings, since then. I am seeing a "Jenkins is stupid" file though; is there some issue in the testing process here, or do I need to do some digging?

Looking at the [console output](https://tests.yt-project.org/job/yt_docs_new/2822/console) it is also not clear to me that the error is actually related to your PR. Something to be aware of is that we only run docs builds on PRs that actively modify or introduce documentation, so it's possible that sometimes a bug slip into a PR that doesn't do that but still breaks documentation builds. This seems unlikely here given that the only PR that was merged between your last succesful run and now is #4945 and looks even less related.
I suspected that sphinx's new version (7.4.x) might have caused it but looking at logs I see that each time pip resolved to sphinx 7.3.7 so that's not it either (but I'm not excluding some *other* dependency might have changed in the interval).

neutrinoceros:
here's what a systematic diffing shows
```diff
35c35
< dask-2024.7.0
---
> dask-2024.7.1
137c137
< pytest-8.2.2
---
> pytest-8.3.1
166c166
< sphinxcontrib-htmlhelp-2.0.5
---
> sphinxcontrib-htmlhelp-2.0.6
169c169
< sphinxcontrib-qthelp-1.0.7
---
> sphinxcontrib-qthelp-1.0.8
```
maybe we could try pinning 
```
 sphinxcontrib-htmlhelp==2.0.5
 sphinxcontrib-qthelp==1.0.7
```
and see what happens. In fact, I'll try this now (again, taking the liberty to push to your branch)

neutrinoceros:
Looks like it indeed stabilized the builds. I'll be working on a finer and more stable solution in https://github.com/yt-project/yt/pull/4952

neutrinoceros:
Also see #4950 about the (unrelated) failure seen on Windows

neutrinoceros:
To my surprise, I couldn't reproduce the failure in #4950. Let me try to remove my last commit here.

neutrinoceros:
Welp. Just some non-reproducible fluke after all ? 🤷‍♂️
At least it's green now.

nastasha-w:
@neutrinoceros Weird, but thank you very much for checking this out!

As for the images: I did notice some of them don't look as pretty. I do think the approach in this pull request produces more scientifically useful line densities. E.g. in one of the OWLS tests, a projection of a cosmological volume looks decidedly more 'pixelly' with this PR's version. I'm pretty sure this is because the new method is specifically sampling the line densities through the pixel center. That means that with a large volume and a limit number of pixels, each pixel's value depends meaningfully on whether that pencil beam happens to intersect any galaxy's ISM. The old/current main method does 'add in' contributions from other SPH particles that fall within the pixel, but I don't believe that is consistent with the general pencil-beam approach the old/main algorithm follows; it is entirely reasonable for a pencil beam to 'miss' some resolution elements, and adding them in is effectively assuming the path is longer than it should be.
I think that in some cases like that one, a nicer image would be obtained by using a different backend algorithm: instead of integrating densities etc. along a pencil beam through a pixel center, calculate the total mass in the rectangular prism defined by the pixel grid and projection depth, and divide by the pixel area. (This also would be a better way to estimate emission per unit surface density or sky area, whereas the pencil beam method is better suited to absorption.) I suppose it would be possible to add that method in as an option, but that seems like a new issue/PR. (An analogous method could also be used for SPH gridding. Something similar could work for slices, but since those are supposed to be infinitely/infinitesimally thin, I think you'd have to explicitly integrate density over slices -- I'd suggest using a square-voxel, one-layer grid with the mass-distribution method there instead.)
I'll take a closer look at those images tomorrow though. I feel like trying some projections with different number of pixels could give some insight there.

neutrinoceros:
> E.g. in one of the OWLS tests, a projection of a cosmological volume looks decidedly more 'pixelly' with this PR's version. 

For convenience, here's the before/after I think you are talking about
before
after
Though I *thought* the first one was "before" and the second was "after", but it's not actually clear to me from what I can read in Jenkins' output, so, which is it ?

I do want to point out that we should aim to preserve backward compatibility wherever possible, and getting visibly different, possibly less pretty images feels like an obvious regression from a user standpoint, so I'd like to suggest that the new method be added as an option but shouldn't replace the default. Is this practical ?

nastasha-w:
Yeah, it's not really clear to me either what's before and after from those outputs. As for the outputs changing: I understand that that breaks backwards compatibility, but the motivation for all these changes was that the old method produces values that are wrong (explained in issue #4788). I don't think we should be keep backwards compatibility when it comes to erroneous results.

neutrinoceros:
> I don't think we should be keep backwards compatibility when it comes to erroneous results.

I agree with you here. My comment was very high level in the sense that I haven't carefully reviewed the code or spent the time needed to understand the approach, so if it's just irrelevant, please disregard !

nastasha-w:
@neutrinoceros so I've taken another look, and I think the more 'pixelly' images are actually the 'before' versions. I've attached plots showing the OWLS cosmo volume and the Arepo cluster merger projected along 3 axes, comparing the current (maybe yesterday's) main branch to the current sph_proj_backend, on the same color scale.
[imgcomp_arepo_clustermerger_xax_main_20240722_sph_proj_backend_20240722.pdf](https://github.com/user-attachments/files/16338598/imgcomp_arepo_clustermerger_xax_main_20240722_sph_proj_backend_20240722.pdf)
[imgcomp_arepo_clustermerger_yax_main_20240722_sph_proj_backend_20240722.pdf](https://github.com/user-attachments/files/16338599/imgcomp_arepo_clustermerger_yax_main_20240722_sph_proj_backend_20240722.pdf)
[imgcomp_arepo_clustermerger_zax_main_20240722_sph_proj_backend_20240722.pdf](https://github.com/user-attachments/files/16338600/imgcomp_arepo_clustermerger_zax_main_20240722_sph_proj_backend_20240722.pdf)
[imgcomp_owls_cosmovol_xax_main_20240722_sph_proj_backend_20240722.pdf](https://github.com/user-attachments/files/16338601/imgcomp_owls_cosmovol_xax_main_20240722_sph_proj_backend_20240722.pdf)
[imgcomp_owls_cosmovol_yax_main_20240722_sph_proj_backend_20240722.pdf](https://github.com/user-attachments/files/16338602/imgcomp_owls_cosmovol_yax_main_20240722_sph_proj_backend_20240722.pdf)
[imgcomp_owls_cosmovol_zax_main_20240722_sph_proj_backend_20240722.pdf](https://github.com/user-attachments/files/16338603/imgcomp_owls_cosmovol_zax_main_20240722_sph_proj_backend_20240722.pdf)
Aside from the 'main' images using more of the color bar, the OWLS images are definitely more 'pixelly' with the current 'main' branch, and it looks like the 'sph_proj_backend' version gets more converged images as well. 

To check convergence issues, I made these two plots, comparing the pixel value histograms and inferred total masses (sum of pixel line densities * area of each pixel) for both those datasets. Total masses look about right and converged for the 'sph_proj_backend' versions, but are systematically too high for the 'main' versions. (Different curves in the same color and linestyle are for different projection axes.) The value histograms are also better converged in the 'sph_proj_backend' version, with the strongest variations occurring in the frequencies of the rarest column densities. On the other hand, on the 'main' branch, there is a high column density tail that doesn't seem to be poorly sampled (based on variation between sightlines), but does come down systematically as the resolution increases. (This is the behavior I described in #4788, due to SPH particles smaller than the grid scale being included in the total line density when they shouldn't be, or being included with path lengths that are longer than they should be.)
[resolution_stats_arepo_clustermerger.pdf](https://github.com/user-attachments/files/16338663/resolution_stats_arepo_clustermerger.pdf)
[resolution_stats_owls_cosmovol.pdf](https://github.com/user-attachments/files/16338664/resolution_stats_owls_cosmovol.pdf)

nastasha-w:
I suspect the weirdness in the Arepo bullet images you showed is mostly an issue of color bar ranges. Adding in small SPH particles with larger-than-physical columns seems to fill in a higher 'background level' in the top image, meaning we don't get the very low surface density values at the cluster edges that the second image seems to show. On the other hand, the top image shows markedly higher surface densities in the center of the image, which may be erroneous.

nastasha-w:
@jzuhone @neutrinoceros @matthewturk @langmm @chummels I've looped people in who seemed interested in #4783, since this is another SPH backend issue. The only non-passing tests now seem to be image comparison tests, meant to test various frontends. I did not change the frontends themselves, but I did change the back-end projection, slicing and gridding functions for SPH data. @neutrinoceros and I were both not entirely clear which images in the test logs were 'before' and 'after', so I have attached some images here showing the difference: the 'main' images/curves were produced using the current (few days old at most) main branch, while the 'sph_proj_backend' versions use the current version of this pull request branch. I've got more images in the message above, but I've attached pngs here for inline viewability.

The images show the projections made with the two branches (top/bottom rows), at increasing resolution (left to right):

imgcomp_owls_cosmovol_xax_main_20240722_sph_proj_backend_20240722
imgcomp_arepo_clustermerger_yax_main_20240722_sph_proj_backend_20240722

These images compare pixel value statistics (left panels) and estimated total masses (right panels) using different branches (solid/dashed lines) and resolutions (colors in the left-hand panels). Multiple curves in the same color and line style show projections along the three simulation axes for each dataset. 

resolution_stats_owls_cosmovol
resolution_stats_arepo_clustermerger

The pixel value histograms converge as expected for the 'sph_proj_backend' branch: the overall distributions with different numbers of pixels look mostly similar, with a better sampling of the rarer values as the resolution increases. Inferred total masses are calculated as (sum of pixel line densities) x (surface area of each pixel). This compares reasonably with the directly calculated total mass in the simulation (horizontal, gray line in the left-hand plots). For the 'main' branch, the histograms change differently with resolution: lower-resolution images have more high surface density values, and intermediate values seem converged at a given resolution (they vary little with choice of projection axis), but these frequency of these intermediate values, as well as the highest surface densities in the image, decrease(s) with higher resolution. Similarly, the total mass is severely overestimated at low resolution, though this effect decreases systematically with resolution. I explained the reasons for this in #4788; basically, SPH particles smaller than the scale of the pixel grid contribute density to sightlines they do not intersect, and the path length through small particles that do intersect the sightline is still overestimated, on the 'main' branch.

neutrinoceros:
Wow. I'm in awe, the improvement at low resolutions is *drastic*. Thanks for this visual summary, I really appreciate it !

jzuhone:
@nastasha-w I'm still trying to digest all of this, but in the meantime, can you check `yt/visualization/fits_image.py`? I saw you had to make some changes to the `PlotWindow` code and the FITS image code uses similar machinery to dump FRBs from slices and projections to files. Just want to make sure that we don't need to make similar changes there as well. 

Or you can wait for me to catch up after a few days of looking at this.

nastasha-w:
@jzuhone I took a look at that code, and I think we're ok for on-axis stuff: the `to_frb` method should call the functions I modified for SPH particles, as should any `FixedResolutionBuffer` creation. For off-axis projections, I'll probably need to explicitly pass the `depth` kwarg in `__init__` to `off_axis_projection`.
Thanks for flagging this! I hadn't realized these classes were calling image creation functions, especially since the `FixedResolutionBuffer` itself seems to have a few methods for saving image data.

nastasha-w:
@jzuhone I've added a commit to pass the `depth` kwarg in `FITSOffAxisProjection` to the `off_axis_projection` function. From there, it goes through functions I've tested, so that part should work.
Do you happen to have any tests written for these `fits_image.py` classes/methods with SPH data?

nastasha-w:
> Plot twist: it never ran automatically, it's always me manually approving. 😅 You need to have at least one commit already in the history for CI to run automatically. Maybe you could issue another much smaller PR that we can merge quickly ?

@neutrinoceros I've made a pull request for a small docs update, warning people that actually we're currently still using nose for answer testing: PR #4955.

neutrinoceros:
Great ! Will review that tomorrow morning !

nastasha-w:
I'm getting this error ([full version](https://github.com/yt-project/yt/actions/runs/10066586909/job/27828578887?pr=4939)) in one of the unit tests:
```
Fatal Python error: Bus error

Thread 0x00000001705df000 (most recent call first):
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 359 in wait
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 655 in wait
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/tqdm/_monitor.py", line 60 in run
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1073 in _bootstrap_inner
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1030 in _bootstrap

...
UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
yt/frontends/stream/tests/test_stream_stretched.py::test_variable_dx
```

This seems to be a test for `load_uniform_grid` or some index-related `dataset` methods, which is something I have not touched at all. It also seems to fail on one specific python+OS version. Could this be related to #4953?

neutrinoceros:
I think it's the same problem as #4953 indeed.  I'm guessing something changed upstream, but don't worry about it here.

neutrinoceros:
(Also, it manifests as a bus error or a segmentation fault, so we cannot trust which exact test is triggering it. Probably not the last one to run.)

nastasha-w:
I made a small change to the SPH projection backend: I am now only including SPH particles with (center) positions between the minimum and maximum coordinate value along the line of sight in the projection. Formerly, any particle within a smoothing length of this coordinate range was included. 
I went into some more depth on this issue [here](https://github.com/yt-project/yt/issues/4788#issuecomment-2246249603), but basically, the smoothing length margins meant particles near a periodic domain boundary in the line-of-sight direction could be double-counted, and generally meant the 'effective' path length covered by the SPH particles in the projection was (a bit) larger than the input depth.
At least on my laptop, this version still passes the new SPH projection tests, and another look at those OWLS and Arepo datasets [discussed here](https://github.com/yt-project/yt/pull/4939#issuecomment-2243898838) shows the projections are no longer systematically overestimating the mass in the projected volume. (Note that this overestimate was small compared to that produced by the current main branch projections.)
resolution_stats_owls_cosmovol
resolution_stats_arepo_clustermerger
20240722 are the versions from two days ago, 20240724 are the updated images. The images themselves do not look different by eye. Apologies for the overlapping text. The right-hand panels matter here, which show the (sum of pixel line densities) x (pixel surface area) mass estimates as a function of image resolution (x-axis, pixels per side), compared to the directly calculated total mass in each simulation (horizontal gray line).
It's not surprising that the agreement is not exact, since we're sampling a limited number of sightlines, but the estimates should convergence to the directly calculated values with increasing resolution.

neutrinoceros:
I think you're in for a bunch of history rewriting 😅 
https://github.com/yt-project/yt/pull/4955#pullrequestreview-2197370110

nastasha-w:
I think I've got those .DS_Store files out of this history too now.

nastasha-w:
pre-commit.ci autofix

jzuhone:
@nastasha-w @matthewturk @chummels @JBorrow @neutrinoceros let's get this across the finish line next week. I think we still need to have some discussion about it, but we should get it done (as well as the other PR about the rays).

JBorrow:
Yup - will take a detailed look this week. Was at an all-week collaboration meeting last week; should have said that, sorry!

yt/data_objects/construction_data_containers.py
```diff
@@ -995,14 +997,15 @@ def _fill_sph_particles(self, fields):
...
 
         bounds, size = self._get_grid_bounds_size()
 
         period = self.ds.coordinates.period.copy()
         if hasattr(period, "in_units"):
             period = period.in_units("code_length").d
-        # TODO maybe there is a better way of handling this
-        is_periodic = int(any(self.ds.periodicity))
+        # check periodicity per dimension
+        is_periodic = self.ds.periodicity
```
JBorrow, line 1008: Is `self.ds.periodicity` a list? I don't understand the removal of `any()` here. If it's a list, this will always be truthy.

yt/geometry/coordinates/cartesian_coordinates.py
```diff
@@ -323,11 +324,20 @@ def _ortho_pixelize(
...
+        period2[1] = self.period[self.y_axis[dim]]
+        period3 = self.period[:].copy()  # dummy here
+        period3[0] = self.period[self.x_axis[dim]]
+        period3[1] = self.period[self.y_axis[dim]]
+        zax = list({0, 1, 2} - {self.x_axis[dim], self.y_axis[dim]})[0]
+        period3[2] = self.period[zax]
+        if hasattr(period2, "in_units"):
+            period2 = period2.in_units("code_length").d
+        if hasattr(period3, "in_units"):
+            period3 = period3.in_units("code_length").d
```
JBorrow, line 340: What is period3 and period2? What's the reasoning for these variable names? I can see before period was the axis that wasn't being projected..? But 3 is not a valid axis index...

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -1202,75 +1211,98 @@ def pixelize_sph_kernel_projection(
                 with gil:
                     PyErr_CheckSignals()
 
-            xiter[1] = yiter[1] = 999
+            xiter[1] = yiter[1] = ziter[1] = 999
```
JBorrow, line 1214: Very confused as to why this is initialised as 999?

JBorrow (commented):
I'm not extremely familiar with the codebase here, but these seem to be good changes. I just have a few questions about variables, really.

I will say that this 'gather' method of SPH projection is very far from being the industry standard. It was my recollection that both 'scatter' and 'gather' were implemented in yt. 'Scatter' uses every particle in the selection region at all times, and as such doesn't suffer from the same mass-conservation problems. Did that implementation get lost? Maybe someone more familiar with the history can comment on that.

As far as I can tell these changes are good to go. I would have liked to see more clear cut PRs that were smaller for the two different changes (periodicity and the density projection), but I understand that this is difficult to do in a situation like this.

JBorrow:
> > I have looked into the instructions in the docs for updating these tests [here](https://yt-project.org/docs/dev/developing/testing.html#creating-and-updating-image-baselines-for-pytest-mpl-tests), but these don't seem to work for me.
> 
> Sorry if this is confusing: these are the instruction for updating _pytest_-based image tests; however, Jenkins is still running nosetest for tests that require real (heavy) datasets.
> 
> > This section [here](https://yt-project.org/docs/dev/developing/testing.html#how-to-run-the-answer-tests) suggests updating the answer_version attribute of the relevant test class
> 
> that's even worse... I believe this part of our documentation was never actually in sync with our practice (it was introduced as part of a substantial but incomplete effort to migrate our test suite to pytest). Disregard it entirely.
> 
> So the actual way we update answers with nose is by bumping version numbers in `tests/tests.yaml`, which triggers a regeneration of baselines on Jenkins. However, it is customary to not do it until a reviewer explicitly approve of it, so the next logical step is for someone to have a look and check that all changes are both expected and desired. I have only looked at a couple myself and I'll admit that the new images are not clearly better to my eye, and I'm not even sure that I can discard the possibility of a bug. For instance here's one:
> 
> ### before (as of yt 4.3)
> before
> 
> ### after (as of this branch)
> after
> 
> can you explain what's going on ?

This is potentially problematic. It looks like some particles that are outside of the selection region, but have smoothing lengths that contribute to the region, are not being included. Though I can't tell from this; you would need to do a scatter plot of the particles with circles representing their smoothing lengths.

yt/data_objects/construction_data_containers.py
nastasha-w: I changed the backend so it can accept either a single boolean or a list/array/whatever of length 3. That allows consistent handling of simulations that are periodic in some, but not all, coordinate directions. I handle those options in  [these lines](https://github.com/yt-project/yt/blob/0580c692ad36035a97fa3561cd34f891ad02b7c3/yt/utilities/lib/pixelization_routines.pyx#L1699C1-L1712C1) in `pixelize_sph_kernel_arbitrary_grid`, file: `yt/utilities/lib/pixelization_routines.pyx`. Basically, it allows backwards compatibility for single-value bool inputs in the backend, but it isn't using `if check_period` or something like that.

yt/geometry/coordinates/cartesian_coordinates.py
nastasha-w: `period2` is a 2-element iterable, which only contains info about the period in the directions perpendicular to the line of sight, `period3` is a 3-element iterable, which also indicates whether the volume is periodic along the line-of-sight direction. Basically, `period3` is what I use in the SPH backend functions I updated, but `period2` is what backends for other simulation types need. The old SPH versions used `period2` as well, so I basically 'split' this variable into a version for each backend.

yt/utilities/lib/pixelization_routines.pyx
nastasha-w: It's just an arbitrary flag value. If a particle doesn't need to be 'duplicated' in a given dimension to capture its (full) effects in the projection, the second iteration of that duplication loop is skipped. The lines setting `xiter[1]`, `yiter[1]`, and `ziter[1]` to different values if we need the duplication are set in the lines directly following the initialization. Those nested loops for the different dimensions are in [these lines](https://github.com/yt-project/yt/blob/0580c692ad36035a97fa3561cd34f891ad02b7c3/yt/utilities/lib/pixelization_routines.pyx#L1251-L1268).

nastasha-w:
@JBorrow as for the 'scatter' and 'gather' methods: I agree that it seems like 'gather' might miss some stuff. However, the code I changed here all uses 'scatter' methods, and 'scatter' also seems to be the default option. 
The off-axis projections and slices call the 'scatter' on-axis versions in the backend, after rotating the coordinates. On-axis projections only have a 'scatter' version, gridding and on-axis slices have 'scatter' and 'gather' options, where 'scatter' is the default, but this can be overridden by a dataset attribute. 
This is based on my understanding that 'scatter' means: loop over SPH particles, deposit contributions in the relevant output pixels/cells, and 'gather' means: loop over output pixels/cells and get contributions from the closest N particles. That would mean that with 'gather', non-zero contributions from SPH particles that are far from the pixel/cell center, but have large smoothing lengths could be missed.

nastasha-w:
> > > I have looked into the instructions in the docs for updating these tests [here](https://yt-project.org/docs/dev/developing/testing.html#creating-and-updating-image-baselines-for-pytest-mpl-tests), but these don't seem to work for me.
> > 
> > 
> > Sorry if this is confusing: these are the instruction for updating _pytest_-based image tests; however, Jenkins is still running nosetest for tests that require real (heavy) datasets.
> > > This section [here](https://yt-project.org/docs/dev/developing/testing.html#how-to-run-the-answer-tests) suggests updating the answer_version attribute of the relevant test class
> > 
> > 
> > that's even worse... I believe this part of our documentation was never actually in sync with our practice (it was introduced as part of a substantial but incomplete effort to migrate our test suite to pytest). Disregard it entirely.
> > So the actual way we update answers with nose is by bumping version numbers in `tests/tests.yaml`, which triggers a regeneration of baselines on Jenkins. However, it is customary to not do it until a reviewer explicitly approve of it, so the next logical step is for someone to have a look and check that all changes are both expected and desired. I have only looked at a couple myself and I'll admit that the new images are not clearly better to my eye, and I'm not even sure that I can discard the possibility of a bug. For instance here's one:
> > ### before (as of yt 4.3)
> > before
> > ### after (as of this branch)
> > after
> > can you explain what's going on ?
> 
> This is potentially problematic. It looks like some particles that are outside of the selection region, but have smoothing lengths that contribute to the region, are not being included. Though I can't tell from this; you would need to do a scatter plot of the particles with circles representing their smoothing lengths.

I've taken a look at this, and if I'm not mistaken, this is the test yt.frontends.tipsy.tests.test_outputs.PixelizedProjectionValues_galaxy_00300_all_('gas', 'Fe_fraction')_0_None
meaning it's using the `test_tipsy_galaxy` function in `yt/frontends/tipsy/tests/test_outputs.py`, which calls the `sph_answer` function in `yt/utilities/answer_testing/framework.py`. In this case, it seems to calculate the projected `("gas", "Fe_fraction")` field along axis 0, with no weighting. This doesn't seem like a physically useful quantity to integrate over a line of sight; I guess it would get you some sort of effective path length for iron? However, as a frontend test, I'm guessing it's set up more to check if `yt` is reading in the fields and their units correctly, in a way that `yt`'s functions and classes can work with.

As for the differences between the branch in this PR and the main branch: I suspect the high values in the main image are due to the (I would argue unphysical) enhancement of path lengths for particles with smoothing lengths smaller than the grid size there. I suspect that is also causing the higher 'background level' in the main branch image compared to the PR, where 'background' sightlines only occasionally fall within a smoothing length of a small particle. 

I tried to take a deeper look at those differences, but I'm getting some sort of segfault on the main branch and on the sph_proj_backend (this PR) branch on my mac for some reason:
```
Intel MKL WARNING: Support of Intel(R) Streaming SIMD Extensions 4.2 (Intel(R) SSE4.2) enabled only processors has been deprecated. Intel oneAPI Math Kernel Library 2025.0 will require Intel(R) Advanced Vector Extensions (Intel(R) AVX) instructions.
Intel MKL WARNING: Support of Intel(R) Streaming SIMD Extensions 4.2 (Intel(R) SSE4.2) enabled only processors has been deprecated. Intel oneAPI Math Kernel Library 2025.0 will require Intel(R) Advanced Vector Extensions (Intel(R) AVX) instructions.
yt : [INFO     ] 2024-07-31 15:37:44,143 Parameters: current_time              = 20.100000000000044
yt : [INFO     ] 2024-07-31 15:37:44,143 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2024-07-31 15:37:44,143 Parameters: domain_left_edge          = [-2000. -2000. -2000.]
yt : [INFO     ] 2024-07-31 15:37:44,144 Parameters: domain_right_edge         = [2000. 2000. 2000.]
yt : [INFO     ] 2024-07-31 15:37:44,144 Parameters: cosmological_simulation   = 0.0
yt : [INFO     ] 2024-07-31 15:37:44,152 Loading KDTree from galaxy.00300.kdtree
Generate smoothing length:   0%|                                            | 0/76962 [00:00<?, ?it/s]zsh: segmentation fault  python compare_projres.py
(ytdev) nastasha@Nastashas-MacBook-Pro ytdev_misc % /Users/nastasha/opt/anaconda3/envs/ytdev/lib/python3.11/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```
This looks an awful lot like the problems @neutrinoceros found for the tests on macs in #4953, but I'm using numpy 1.26.4. (2021 14-inch macbook pro, M1 chip, OS: Sonoma 14.5). That issue seems to have been resolved upstream, so for now, I'm just waiting for the fix to come through, I guess. However, I do think this is the first time I'm working with a dataset that needs to calculate the smoothing lengths instead of reading them in, so I do worry a bit that something in that part of the code doesn't work well on macs.

neutrinoceros:
> That issue seems to have been resolved upstream, so for now, I'm just waiting for the fix to come through, I guess.

The error message is similar alright, but you might want to check if it's the same underlying function (`np.unique`) crashing. If not, it might be another (very similar) bug in numpy or of its vendored C++ libraries.

nastasha-w:
@neutrinoceros the error persists with the newest numpy and yt. Installing from the yt top directory on my laptop:
```
conda create -n ytdev python=3.12 pip 
python -m pip install git+https://github.com/numpy/numpy.git
python -m pip install -e ".[doc]"
python -m pip install h5py pandas
```
I might open a new issue for this then, but in the mean time, I'll see if I can run a test on the university cluster (linux) instead.

neutrinoceros:
Can you provide a little bit more context for how you hit this new segfault ? Which test is it and how do you run it ?

nastasha-w:
@neutrinoceros I have opened a separate issue for this problem: #4961.

nastasha-w:
@JBorrow I don't think I'm going to be able to test that Tipsy galaxy projection any time soon. I tried to get it to work on the university linux cluster, and I ended up getting the same error ([described in more detail here](https://github.com/yt-project/yt/issues/4961#issuecomment-2264432435)). This was of course after many failed attempts to even get yt installed -- the installation process just seemed to grind to a halt on the login node. I think there might be a broader issue with the smoothing length loading which the Tipsy data seems to require, although I guess it cannot be too broad if the main branch is still passing its tests on a variety of systems.

nastasha-w:
@JBorrow I tested on an OWLS cosmo volume and an Arepo cluster merger here: [linked pdfs of all plots](https://github.com/yt-project/yt/pull/4939#issuecomment-2243624021), [in-line png selection](https://github.com/yt-project/yt/pull/4939#issuecomment-2243898838), [additional in-line results after a new commit](https://github.com/yt-project/yt/pull/4939#issuecomment-2248468277). This is in addition to tests on small, fake datasets, which I added to the test suite. The focus on the Tipsy dataset was because @JBorrow and @neutrinoceros specifically asked about the before-and-after images of that data from the image comparison tests.

I also took a look at the `gizmo_mhd_mwdisk` dataset, from the yt hub like the others here. For that, I made slices and projections along the x- and z-axes, and then off-axis slices/projections with the y-axis as the constant `north_vector`, changing the line-of-sight direction from x to z in small steps. (I used directions slightly off the x- and z-axes for the first and last steps.) I've attached some overviews of this: first projections along the x/z-axes, then projections in direction steps from x to z, and then the same two overviews but for slices.
projections_xz_2024-07-11
projections_xtoz_2024-07-11
slices_xz_2024-07-11
slices_xtoz_2024-07-11
Apologies for the tiny labels; I produced these as individual images and put them together in powerpoint.

nastasha-w:
@JBorrow @neutrinoceros Well, it turns out the Tipsy galaxy `smoothing_length` calculations do work, even on a mac, with a `venv`+`pip` installation of `yt`, just not with a `conda` + `conda`'s `pip` installation. Ugh. I might submit a PR to add a warning about this in the installation docs. 

Anyway, I could reproduce those image test differences on my laptop, but from overplotting the particle smoothing lengths, I think it's the version in this PR that's getting it right.
 
imgcomp_fe_tipsy_galaxy_xax_main_20240805_sph_proj_backend_20240805
Here is a similar image comparison to that for the Arepo and OWLS data: the top row shows the current (2024-08-05) `main` branch image, and the bottom rows shows the same for the `sph_proj_backend` (this PR) branch. The image resolution increases from left to right. In this case, the differences are mostly in the corners of the projected regions: they are empty in the `sph_proj_backend` version.
 
imgcomp_fe_tipsy_galaxy_xax_main_20240805_sph_proj_backend_20240805_hsml
This is the same image, but with the SPH particle locations and smoothing lengths overlayed. All particles are shown, but the image is filled mostly by the projections of a small number of SPH particles. The dotted lines show the plotted domains of the previous region, which are also the bounds of the calculated projection. Here, we see that the `main` version shows non-zero column densities where there are no SPH particles along the line of sight, whereas the images are 'empty' along those sightlines in the `sph_proj_backend` version. The very low surface densities at the edges of the 'filled' region seem to be a straightforward result of the path integral through an SPH kernel being small at impact parameters close to its smoothing length.

neutrinoceros:
> I might submit a PR to add a warning about this in the installation docs.

Please don't (yet). See https://github.com/yt-project/yt/issues/4961#issuecomment-2269748803

JBorrow:
@nastasha-w thank you for this. You are completely right in your assertion that your version is correct and the old version is wrong!

neutrinoceros:
> Here, we see that the main version shows non-zero column densities where there are no SPH particles along the line of sight

I wonder if there's any kind of box selection in your example, and in particular if there would be particles whose centers lie outside the box but still overlap with it and should contribute.

nastasha-w:
@neutrinoceros There is a bounding box here, `[[-2000, 2000], [-2000, 2000], [-2000, 2000]]`, which I copied from the Tipsy galaxy test in `yt/frontends/tipsy/tests/test_outputs.py` (starts on line 84). For some reason, the Tipsy data doesn't have the `("Gas", "smoothing_length")` field available (even though it's in `ds.field_list`) if I load it without a bounding box. If I load it without a bounding box, I get domain edges 
```
In [28]: ds.domain_left_edge.to("kpc")
Out[28]: unyt_array([ -547.51629822,  -557.30091064, -3675.19602173], 'kpc')
In [29]: ds.domain_right_edge.to("kpc")
Out[29]: unyt_array([ 579.153078  ,  472.81018799, 1591.69712036], 'kpc')
```
so loading with a bounding box of `[[-1e5, 1e5]] * 3` should be good enough to get all the particles. However, the results from that are weird:
imgcomp_fe_tipsy_galaxy_xax_main_20240805_sph_proj_backend_20240805_hsml
`bounding_box=[[-2e5, 2e5],] * 3`, `bounding_box=[[-3e4, 3e4],] * 3`,  and `bounding_box=[[-1e4, 1e4],] * 3` give the same (or at least, a similar-looking) weird answer. `bounding_box=[[-3e3, 3e3],] * 3` gives a 'less weird' answer (better overlap with the test box SPH particle size):
imgcomp_fe_tipsy_galaxy_xax_main_20240805_sph_proj_backend_20240805_hsml
and `bounding_box=[[-2.5e3, 2.5e3],] * 3` further narrows the difference:
imgcomp_fe_tipsy_galaxy_xax_main_20240805_sph_proj_backend_20240805_hsml
The color-scale images/projections are the same ones from before, made with the `[[-2000, 2000], [-2000, 2000], [-2000, 2000]]` bounds also used in the test suite image test, but the SPH positions and smoothing lengths were read in using a dataset loaded with other `bounding_box` arguments. It seems like, at least for the largest SPH particle, the bounding box actually affects the smoothing length calculations. I'm really not sure what to make of that, other than that comparing projections with the same bounding box is important here. Note that I did not touch any smoothing length calculation code in this PR.

However, it does not seem like any SPH particles are missed entirely in the test image for the `bounding_box=[[-2000, 2000], [-2000, 2000], [-2000, 2000]]` dataset.

nastasha-w:
Further checks:
* Overlaying the SPH particle positions and smoothing lengths using a dataset read in on the main branch gives the same result.
* The projections and read-in smoothing lengths and positions are consistent in the `sph_proj_backend` (this PR) version: the image below shows a projection and SPH smoothing lengths from the Tipsy dataset read in with `bounding_box=[[-3e3, 3e3],] * 3`: 
  imgcomp_fe_tipsy_galaxy_box3k_xax_main_20240805_sph_proj_backend_20240805_hsml
  I think this also solves the problem of the 'filled-in' corners in the `main` branch version: the volume is assumed to be periodic in all directions in the `main` branch version, while it isn't in the `sph_proj_backend` version. That means that on `main`, the largest SPH particle at the bottom of the image is also filling in the top corners. 

  Note that when loading the dataset with `bounding_box=None` or `bounding_box=[[-2e3, 2e3],] * 3`, in the `main` and `sph_proj_backend` versions, its `periodicity` attribute is `(False, False, False)`. This suggests it is the `sph_proj_backend` version which is correctly using the dataset periodicity.

nastasha-w:
If I remember correctly, this periodicity handling error comes down to the default argument `periodic=True` in the `CartesianCoordinateHandler`'s `pixelize` method on the `main` branch. 
https://github.com/yt-project/yt/blob/0580c692ad36035a97fa3561cd34f891ad02b7c3/yt/geometry/coordinates/cartesian_coordinates.py#L164-L175
This method is called in the `FixedResolutionBuffer`'s `_generate_image_and_mask` method, but this method does not pass on the `FixedResolutionBuffer` `periodic` attribute.
https://github.com/yt-project/yt/blob/39a7563e64e8ec1e8d12ea3af2cc6da5354419d1/yt/visualization/fixed_resolution.py#L179-L187
Since I didn't want to 'mess with' any non-SPH methods, on the `sph_proj_backend` branch, I just ignore the `pixelize` `periodic` argument, and get the periodicity per dimension from the dataset instead if we're dealing with SPH data.
https://github.com/yt-project/yt/blob/0580c692ad36035a97fa3561cd34f891ad02b7c3/yt/geometry/coordinates/cartesian_coordinates.py#L385-L389 
This also allows for consistent handling of situations where the volume is only periodic in some directions. (I believe this is often the case in e.g., cloud-crushing simulations.)

neutrinoceros:
> For some reason, the Tipsy data doesn't have the ("Gas", "smoothing_length") field available (even though it's in ds.field_list) if I load it without a bounding box.

That's almost certainly a bug. It'd be worth investigating but I think it's clearly out of scope here. Thanks for noticing !

Some piece of information I was lacking when reading https://github.com/yt-project/yt/pull/4939#issuecomment-2270101972 is that `kpc` is that code unit for length in this particular dataset. Could you edit this remark in for posterity ?

> It seems like, at least for the largest SPH particle, the bounding box actually affects the smoothing length calculations, or its position along the z-axis. I'm really not sure what to make of that, other than that comparing projections with the same bounding box is important here. Note that I did not touch any smoothing length calculation code in this PR.

To me this sounds like a blocker: something we need to address or at the very least understand before we can merge your PR. Maybe the issue is already on main and just happens to have no observable effect with the existing methods, so I think it's maybe be to inquiry from within your branch, and if any fix is needed that can be separated, submit a new PR for it.

---

Now answering https://github.com/yt-project/yt/pull/4939#issuecomment-2270144600 :
I didn't get the difference between the top row on this picture and that of the previous comment.

> Overlaying the SPH particle positions and smoothing lengths using a dataset read in on the main branch gives the same result.

Can you clarify ? the 'same result' as compared to what ? Sorry I'm lagging behind and I'd like to make sure I'm at least following the right train of thoughts 😅  

> I think this also solves the problem of the 'filled-in' corners in the main branch version: the volume is assumed to be periodic in all directions in the main branch version, while it isn't in the sph_proj_backend version. That means that on main, the largest SPH particle at the bottom of the image is also filling in the top corners.

this makes sense to me.

> Note that when loading the dataset with bounding_box=None or bounding_box=[[-2e3, 2e3],] * 3, in the main and sph_proj_backend versions, its periodicity attribute is (False, False, False). This suggests it is the sph_proj_backend version which is correctly using the dataset periodicity.

or at least it's *aligning* with the expected result (just want to emphasize that it could also happen by "accident") 😅
We'd need to also test with a truly periodic dataset to verify that it's really working as intended.

> Since I didn't want to 'mess with' any non-SPH methods, on the sph_proj_backend branch, I just ignore the pixelize periodic argument, and get the periodicity per dimension from the dataset instead if we're dealing with SPH data.

Hum, I agree that we should restrict the scope of this PR as much as possible, but I also would like to avoid creating API inconsistency on purpose. Can you try the general change, just to see how much of the test suite it breaks ? If it turns out massive, I'd like to ponder a bit more on that. Otherwise (if it's merely breaking any thing), we should open an reminder issue to do it as a follow up to your PR, and revert it here.

nastasha-w:
@neutrinoceros
- `kpc` units: added to the post
- `("Gas", "smoothing_length")` field: I'll open an issue for that
- >>Overlaying the SPH particle positions and smoothing lengths using a dataset read in on the main branch gives the same result.
  
  >Can you clarify ? the 'same result' as compared to what ? Sorry I'm lagging behind and I'd like to make sure I'm at least following the right train of thoughts 😅
  
  I mean that I plotted the positions and smoothing lengths in the plots using the new/PR version of the `yt` code to read in the dataset. However, if I read in the Tipsy dataset using the old/main branch of `yt` instead, I get the same positions and smoothing lengths. (At least insofar as I can visually distinguish in the plots.) I.e., it's not that the differences between the plots are due to different particle positions and smoothing lengths determined by the two code versions.
- >>It seems like, at least for the largest SPH particle, the bounding box actually affects the smoothing length calculations, or its position along the z-axis. I'm really not sure what to make of that, other than that comparing projections with the same bounding box is important here. Note that I did not touch any smoothing length calculation code in this PR. 

  >To me this sounds like a blocker: something we need to address or at the very least understand before we can merge your PR. Maybe the issue is already on main and just happens to have no observable effect with the existing methods, so I think it's maybe be to inquiry from within your branch, and if any fix is needed that can be separated, submit a new PR for it.

  I've also included this in the issue for the smoothing length field: more`bounding_box` weirdness in the Tipsy dataset read-in.
- >>Note that when loading the dataset with bounding_box=None or bounding_box=[[-2e3, 2e3],] * 3, in the main and sph_proj_backend versions, its periodicity attribute is (False, False, False). This suggests it is the sph_proj_backend version which is correctly using the dataset periodicity.

  >or at least it's aligning with the expected result (just want to emphasize that it could also happen by "accident") 😅
We'd need to also test with a truly periodic dataset to verify that it's really working as intended.
  
  I did include all-directions periodic and non-periodic dataset testing in the tests I added in this PR. Here's the file: https://github.com/yt-project/yt/pull/4939/files#diff-fa7c301d743ab68cc8f61cef08e6386dccbf3ee1471d15c3bea50c8d2e39c770
- >>Since I didn't want to 'mess with' any non-SPH methods, on the sph_proj_backend branch, I just ignore the pixelize periodic argument, and get the periodicity per dimension from the dataset instead if we're dealing with SPH data.

  >Hum, I agree that we should restrict the scope of this PR as much as possible, but I also would like to avoid creating API inconsistency on purpose. Can you try the general change, just to see how much of the test suite it breaks ? If it turns out massive, I'd like to ponder a bit more on that. Otherwise (if it's merely breaking any thing), we should open an reminder issue to do it as a follow up to your PR, and revert it here.

  I get your point, but there's another problem with just using the input argument: those are boolean `True`/`False`, while the dataset `periodicity` allows per-axis periodicity checks. E.g., `int(periodic)` is input into a backend function here: https://github.com/yt-project/yt/blob/0580c692ad36035a97fa3561cd34f891ad02b7c3/yt/geometry/coordinates/cartesian_coordinates.py#L347-L365 
  and here: https://github.com/yt-project/yt/blob/0580c692ad36035a97fa3561cd34f891ad02b7c3/yt/geometry/coordinates/cartesian_coordinates.py#L653-L666
  which means there would already be an API inconsistency in what arguments for `periodic` are allowed for SPH and other datasets. This is alongside the fact that the default argument should, in my opinion, probably not be `True`, but 'check the dataset', e.g., through a `None` option. Again, that would be an API change or option difference.

jzuhone:
@nastasha-w does `("gas", "smoothing_length")` exist if you load it without a bounding box? The `"Gas"` field type is specific to Tipsy and in general we want to avoid non-frontend code using non-general field types.

nastasha-w:
@jzuhone the `("Gas", "smoothing_length")` field does not exist if the dataset is not loaded with a specified `bounding_box`. Annoyingly, I do need to load the dataset with `bounding_box=None` to generate the kdtree and smoothing length stored files, because I get a segfault if I delete those files, then try to read in the dataset with a bounding box. Ugh. 

I documented this field issue, and the dependence of particle position on the bounding box, in #4963. I have also confirmed this happens on the current `main` branch as well, so it doesn't seem to be caused by this PR. (That would be odd, since I didn't touch any frontend or smoothing length calculation code for this PR.)

nastasha-w:
@neutrinoceros about those periodicity arguments: I pushed back a bit on that before, but I do agree it's weird to have a periodicity argument in the `CartesianCoordinateHandler`'s `pixelize` method, then to just ignore it when calling the method, and ignoring the argument when setting periodicity for the back-end. I figure maybe the best way to handle this is to either have a default `None` option for periodicity, so that a user can override it if they want to, and if this is the given setting, to use the dataset attributes to the degree the backend supports it (so a 3-tuple for the SPH datasets and `True`/`False` for the others), or to remove the method's periodicity argument altogether and just infer this from the dataset. (I don't see many use cases for deliberately using something different from the dataset values.)

neutrinoceros:
> remove the method's periodicity argument altogether and just infer this from the dataset. (I don't see many use cases for deliberately using something different from the dataset values.)

To me, this sounds like the correct approach. `Dataset` already has a `force_periodicity` method, I don't think we should expose other ways to override it.

nastasha-w:
@neutrinoceros I do generally agree with removing the `pixelize` method's `periodic` argument in the `CartesianCoordinateHandler`. However, I was checking where `pixelize` is used in the code, and other `CoordinateHandler` subclasses (e.g., for spherical, cylindrical, geographic coordinates) also have `pixelize` methods, and projection functions call these without knowing which `CoordinateHandler` subclass they're looking at. In other words, we'd also need to remove the `periodic` argument for these other coordinate systems. Now, in principle, that seems fine to me, assuming the dataset `periodicity` attribute works the same for datasets with these coordinate systems, but I do worry about messing things up there. I'll give things a go, but could you take a look at those things specifically to check that the changes make sense for non-cartesian coordinates?

nastasha-w:
@neutrinoceros given that this might involve some broader changes to the `CoordinateHandler` and its subclasses, would it be worth leaving the API a little janky for now, and opening a separate pull request to just use the simulation/`Dataset` `periodicity` to set the backend periodicity handling? It seemed from comments in some of the other `CoordinateHandler` subclasses that there was some stuff that was still todo there, or that a specific assumption on periodicity was the only implemented case anyway.

jzuhone:
@nastasha-w I've approved PRs #4964 and #4968 which will take care of some of the test failures seen here. 

Is there anything left to do for this? It looks like we may need to bump some answer tests. Otherwise I am satisfied with this.

nastasha-w:
@jzuhone Thanks! I think many of the test failures are reasonable here though: these test if the projections are the same as in the previous code version, and with changes to the backend, I'd expect them to be different. 
There is also the issue of the `periodic` argument in `CartesianCoordinateHandler`'s `pixelize` method the @neutrinoceros brought up: I am currently ignoring that argument for SPH datasets and using the dataset `periodicity` attribute instead. This is because the `FixedResolutionBuffer` doesn't actually supply the `periodic` argument the `pixelize`, meaning it's just always using the default `True` value, and because the new SPH backend allows for setting the periodicity differently along different coordinate directions. The single boolean `periodic` argument does not allow for this. However, just ignoring that argument in one specific case is an API inconsistency.

nastasha-w:
@jzuhone sorry, I hadn't seen your comment on the API/periodicity thing!

jzuhone:
@nastasha-w if the test failures are expected than we need to update the answers so the tests will pass.

neutrinoceros:
> would it be worth leaving the API a little janky for now, and opening a separate pull request to just use the simulation/Dataset periodicity to set the backend periodicity handling?

yes *but* maybe start with an issue instead of a PR so we can start fresh

>  could you take a look at those things specifically to check that the changes make sense for non-cartesian coordinates?

In non-cartesian coordinates periodicity rarely ever makes sense in all directions at once. Radial periodicity is (almost) unheared of, while periodicity in angular directions is extremely common, so the current paradigm of setting none-or-all directions to periodic is pretty much useless in those coordinate systems. Some pixelizers[^1] are also special in that they don't project data in a coordinate native plane but a transformation of it. They should handle periodicity very differently, but at the moment they just ignore it (IIRC).


[^1]: I'm thinking Aitoff-Hammer, maybe that's the only one

jzuhone:
@nastasha-w @neutrinoceros so after we get the non-related test failures fixed from the other PRs, it looks like we can assess what's left and @nastasha-w will need to push the commit that bumps the answers.

I think that @JBorrow approved from his comments, so he can also hit the approve button.

Looking for final comments from @matthewturk, @chummels, @brittonsmith. We should aim to merge this by Tuesday of next week.

jzuhone:
@yt-fido test this please

jzuhone:
Ok, all failing tests now are projection tests for SPH frontends--which is as expected. I looked at quite a few of them, most of which have <10% of pixels failing. 

I'm convinced, then. @nastasha-w let us know if you need help figuring out how to bump the answers--it's a matter of bumping the numbers in `yt/tests/tests.yaml`.

JBorrow (approved):

neutrinoceros:
A quick note that we usually just reference the latest bump in comments, so feel free to replace older comments completely !

nastasha-w:
With the test number bump, many of the previously failing tests now pass. However, for arepo frontend tests specifically, the tests fail with error "There is no old answer available.". Does anyone have any idea what the problem could be, or how to fix it? 
To be clear, this is just arepo; the owls, tipsy, gadget, and gizmo frontend tests pass after increasing the test number.

neutrinoceros:
oh, could this be showing some form of corruption in the database created by not letting the first job finish ?
@Xarthisius ?

jzuhone:
@neutrinoceros that's almost certainly what it is. @Xarthisius needs to fix it

nastasha-w:
The checks passed!!

jzuhone:
@chummels said he was going to have a look at this.

jzuhone (approved):

neutrinoceros:
Let's move it to the 4.4.0 milestone then 🤞🏻

jzuhone:
I'm going to merge this by COB Monday unless someone explicitly jumps in. We have two approvals and it's ready to go.

chummels:
I'm running tests locally and not seeing any difference in behavior for OffAxisProjections and ProjectionPlots between the dev tip and this PR.  I'm using the FIRE m12i z=0 snapshot at a variety of resolutions and for both the ('gas', 'density') and ('gas', 'H_p0_number_density') fields, but getting identical results, so I must be doing something wrong on this side.  I would think as I stepped out in resolution to something like width=(500, 'kpc') for the projection, then some of the smaller particles would get dropped as per the description in the PR, but the results look identical.  I'm just doing a sanity check on all of this, since the PR is so detailed that it is difficult to review thoroughly by simple code review.

neutrinoceros:
@chummels just to get this out of the way: did you make sure to re-compile after switching branches ?

nastasha-w:
@chummels since you mentioned the width, I also wonder whether you're increasing the grid size (number of pixels) in tandem with the image width (i.e., constant width per pixel), or increasing the grid size with constant image width (decreasing width per pixel). The first would not give a resolution effect with either method, and the differences between the methods would depend on which width per pixel you chose (larger -> bigger difference). The second would give the effect I was worried about. (I'm guessing you thought of this, it's just to be sure.)

chummels:
> @chummels just to get this out of the way: did you make sure to re-compile after switching branches ?

Yes, I update between the branches, then `pip install -e .` to recompile the code before re-running my script.

chummels:
> @chummels since you mentioned the width, I also wonder whether you're increasing the grid size (number of pixels) in tandem with the image width (i.e., constant width per pixel), or increasing the grid size with constant image width (decreasing width per pixel). The first would not give a resolution effect with either method, and the differences between the methods would depend on which width per pixel you chose (larger -> bigger difference). The second would give the effect I was worried about. (I'm guessing you thought of this, it's just to be sure.)

I was merely changing the width of the ProjectionPlot and not explicitly changing the number of pixels in the resulting image.  It seems to me that by changing the width of the ProjectionPlot and keeping the number of pixels fixed, it would change the physical pixel size.  It was my understanding from the PR that this would result in different effects when the pixel size became increasingly large relative to the size of the fluid elements, that small fluid elements would be dropped from each pixel's line integral and there would be a change in image.

nastasha-w:
Yes, changing the width and keeping the number of pixels fixed should lead to resolution-dependent effects. I'm just not sure it would be as easy to see those effects by eye. Are you comparing pixel values directly here?

nastasha-w:
I tried a similar test myself:
```
projaxes_cam = ["x", "z", (np.sqrt(0.5), 0., np.sqrt(0.5))]
widths_cam = [62.5, 125., 250., 500.] #kpc
qtys_cam = [('gas', 'density'), ('gas', 'H_p0_number_density')]
```
running on the `sph_proj_backend` (this PR) and `main` (2024-09-11) branches, with the `gizmo_mhd_mwdisk`.

The first difference I found is that the `sph_proj_backend` would not work with the `(np.sqrt(0.5), 0., np.sqrt(0.5))` projection axis; this is expected since off-axis projections were not previously supported for SPH datasets in yt. I therefore limited myself to x- and z-axis projections.
There, I did find the expected differences: as I zoom out to wider views, the difference between the images generated with the two branches increases, especially within the galaxy, where the densities are highest, and the smoothing lengths are therefore smallest.
 
imgcomp_gizmo_mwdisk_xax_gasdenssph_proj_backend_20240911_main_20240911
imgcomp_gizmo_mwdisk_xax_h1denssph_proj_backend_20240911_main_20240911
imgcomp_gizmo_mwdisk_zax_gasdenssph_proj_backend_20240911_main_20240911
imgcomp_gizmo_mwdisk_zax_h1denssph_proj_backend_20240911_main_20240911

nastasha-w:
I also downloaded the FIRE m12i data from the yt hub ([FIRE_M12i_ref11](https://hub.yt/#item/579fb0a67b6f0800011ea166)), and got the following images for x and z projections of gas density and H I number density:
imgcomp_fire_m12i_xax_gasdenssph_proj_backend_20240911_main_20240911
imgcomp_fire_m12i_xax_h1denssph_proj_backend_20240911_main_20240911
8-0a4bcc678699)
imgcomp_fire_m12i_zax_gasdenssph_proj_backend_20240911_main_20240911
imgcomp_fire_m12i_zax_h1denssph_proj_backend_20240911_main_20240911
At first glance, there does not appear to be a difference between the top and middle rows, but at least in the two rightmost images in all four of these comparisons (most zoomed out images), there are in fact small differences between the branches at the images centers. These are not or barely discernible in the small png image panes here, but I could see them using the zoom tool in interactive matplotlib plotting.
@chummels Is it possible that those differences in a small area are also present in your tests?

nastasha-w:
Here are those same tests, but only showing the part of the image from the highest zoom level in each panel. The differences are more apparent here, including that the central 'dot' in in `main` branch image seems to get brighter as the resolution decreases (left to right), while it just gets 'fuzzier' on the `sph_proj_backend` branch.
 
imgcomp_fire_m12i_xax_gasdenssph_proj_backend_20240911_main_20240911_samezoom
imgcomp_fire_m12i_xax_h1denssph_proj_backend_20240911_main_20240911_samezoom
imgcomp_fire_m12i_zax_gasdenssph_proj_backend_20240911_main_20240911_samezoom
imgcomp_fire_m12i_zax_h1denssph_proj_backend_20240911_main_20240911_samezoom

JBorrow:
It's naturally harder to see these kind of artifacts when looking in a logarithmic density space. However, the inaccuracies are always there (as shown by Natasha's plots). I ran into this problem when looking at convergence in this paper https://ui.adsabs.harvard.edu/abs/2021arXiv210605281B/abstract (referenced in the other thread). Specifically this figure:

Screenshot 2024-09-11 at 1 54 53 PM

Looking at the two you don't see differences immediately, but the numerical discrepancies are there. They sensitively depend on the ratio between the typical smoothing length and the pixel size, as Natasha's plots show.

chummels:
So to follow up on this, I guess my old install of yt was quite old indeed.  I removed yt and conda and all dependencies and reinstalled everything from scratch, and *now* I see that this PR gives the desired behavior.  Here are some quick and dirty plots showing that this PR does in fact address the resolution issue with consistent behavior.

FIRE_m12i:

FIRE_comp


AREPO Cluster:

AREPO_comp

Nominally, I think everything is good to go but maybe best to chat in a few hours about all of this before merge?

chummels (approved):

nastasha-w:
@chummels yeah, I think this would be a good discussion to have, on which methods are desirable if nothing else. I'm glad it worked out, but yeah, those installations of different versions are consistently a pain in the heinie.

jzuhone:
Where did we land with this at the meeting?

nastasha-w:
Basically, merge the sph_proj_backend branch with the backend fixes, but with a few small action items for other pull requests:
- add a bit of info on the backend method, and differences between the SPH and AMR/grid backends, to the narrative docs and the docstrings
- clean up some testing functions that are duplicate between the sph_proj_backend PR and the smaller PR for the Ray backend fix (inputting squared vs. not-squared normalized impact parameters into the backend function).

And we decided it would be a good idea to add a different backend for SPH projections as the default, with the pixel center line-of-sight integrations as an alternative option. The new default would average the total contributions from all SPH particles to the rectangular prism defined by each grid pixel over that pixel. E.g., for the a column density, the density of each particle would be multiplied by the integral of the SPH kernel over the pixel rectangular prism, and then the total of those masses would be divided by the pixel surface area. The current method integrates the kernel over a pencil beam through the pixel center and multiples the density by that in stead. Both methods are valid ways of computing a column density, which different methods preferred for different applications. However, the rectangular prism integral is more intuitive, and more similar to the method used for AMR and grid data. This would be another larger PR.

We should probably open issues for these.

Kind regards,
Nastasha

PR #4940, opened 2024-07-07
RFC: explicitly set strict flag in `zip` calls (enforce `B905`), and ignore `UP038`

### LINKED ISSUE #4749: MNT: how should we handle Python 3.10 becoming our oldest supported version ?
Python 3.10 is next in line to become our oldest supported version (currently it's 3.9), sometime about October 2024 (extrapolating from the last couple year's pace).
This version comes with a couple changes that will be picked up but our linters (`pyugrade` and `flake8-bugbear` via `ruff`) and that would affect a large number of lines, so I'd like to start the discussion early to decide what to do about them. 

Here are the rules I know about

## [pyupgrade UP038](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/)

Basically this would change
```python
isinstance(x, (int, float))
```
to
```python
isinstance(x, int | float)
```
It's noted that the more modern ("good") alternative is actually slower, so we might want to just ignore this specific rule. 

## [flake8-bugbear B905](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/)
Python 3.10 added a `strict` argument to the builtin `zip` function which is `False` by default for backward compatibility.
This rule checks that `strict` is explicitly set everywhere, which is always safer or more explicit than leaving it unset.
I don't think it'd be controversial to follow this rule ? It'd a bit tedious to retro-actively set it everywhere but I'd happily volunteer for this boring task. It can actually be addressed right now by using `yt._maintenance.backports.zip` which imitate the Python 3.10 function.

neutrinoceros:
Close #4749

This is a mostly automated style change, but I'm using the test suite to detect places where `strict=False` is either the intended behavior or a bug.

I'll also carefully review this myself to improve coverage.

### DIFF
--- a/yt/data_objects/analyzer_objects.py
+++ b/yt/data_objects/analyzer_objects.py
@@ -1,6 +1,10 @@
 import inspect
+import sys
 
 from yt.utilities.object_registries import analysis_task_registry
+
+if sys.version_info < (3, 10):
+    from yt._maintenance.backports import zip
 
 
 class AnalysisTask:
@@ -14,7 +18,7 @@ class AnalysisTask:
         # does not override
         if len(args) + len(kwargs) != len(self._params):
             raise RuntimeError
-        self.__dict__.update(zip(self._params, args))
+        self.__dict__.update(zip(self._params, args, strict=False))
         self.__dict__.update(kwargs)
 
     def __repr__(self):
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -1,6 +1,7 @@
 import fileinput
 import io
 import os
+import sys
 import warnings
 import zipfile
 from functools import partial, wraps
@@ -62,6 +63,9 @@ from yt.utilities.parallel_tools.parallel_analysis_interface import (
     parallel_root_only,
 )
 from yt.visualization.color_maps import get_colormap_lut
+
+if sys.version_info < (3, 10):
+    from yt._maintenance.backports import zip
 
 
 class YTStreamline(YTSelectionContainer1D):
@@ -150,7 +154,9 @@ class YTStreamline(YTSelectionContainer1D):
         mask = np.zeros(points_in_grid.sum(), dtype="int64")
         dts = np.zeros(points_in_grid.sum(), dtype="float64")
         ts = np.zeros(points_in_grid.sum(), dtype="float64")
-        for mi, (i, pos) in enumerate(zip(pids, self.positions[points_in_grid])):
+        for mi, (i, pos) in enumerate(
+            zip(pids, self.positions[points_in_grid], strict=True)
+        ):
             if not points_in_grid[i]:
                 continue
             ci = ((pos - grid.LeftEdge) / grid.dds).astype("int64")
@@ -784,7 +790,10 @@ class YTCoveringGrid(YTSelectionContainer3D):
     def icoords(self):
         ic = np.indices(self.ActiveDimensions).astype("int64")
         return np.column_stack(
-            [i.ravel() + gi for i, gi in zip(ic, self.get_global_startindex())]
+            [
+                i.ravel() + gi
+                for i, gi in zip(ic, self.get_global_startindex(), strict=True)
+            ]
         )
 
     @property
@@ -1122,7 +1131,7 @@ class YTCoveringGrid(YTSelectionContainer3D):
         if self.comm.size > 1:
             for i in range(len(fields)):
                 output_fields[i] = self.comm.mpi_allreduce(output_fields[i], op="sum")
-        for field, v in zip(fields, output_fields):
+        for field, v in zip(fields, output_fields, strict=True):
             fi = self.ds._get_field_info(field)
             self[field] = self.ds.arr(v, fi.units)
 
@@ -1221,7 +1230,7 @@ class YTCoveringGrid(YTSelectionContainer3D):
             data[field] = (self[field].in_units(units).v, units)
         le = self.left_edge.v
         re = self.right_edge.v
-        bbox = np.array([[l, r] for l, r in zip(le, re)])
+        bbox = np.array([[l, r] for l, r in zip(le, re, strict=True)])
         ds = load_uniform_grid(
             data,
             self.ActiveDimensions,
@@ -1526,7 +1535,7 @@ class YTSmoothedCoveringGrid(YTCoveringGrid):
                 stacklevel=1,
             )
             mylog.debug("Caught %d runtime errors.", runtime_errors_count)
-        for field, v in zip(fields, ls.fields):
+        for field, v in zip(fields, ls.fields, strict=True):
             if self.level > 0:
                 v = v[1:-1, 1:-1, 1:-1]
             fi = self.ds._get_field_info(field)
--- a/yt/data_objects/data_containers.py
+++ b/yt/data_objects/data_containers.py
@@ -1,4 +1,5 @@
 import abc
+import sys
 import weakref
 from collections import defaultdict
 from contextlib import contextmanager
@@ -27,6 +28,9 @@ from yt.utilities.exceptions import (
 from yt.utilities.object_registries import data_object_registry
 from yt.utilities.on_demand_imports import _firefly as firefly
 from yt.utilities.parameter_file_storage import ParameterFileStore
+
+if sys.version_info < (3, 10):
+    from yt._maintenance.backports import zip
 
 if TYPE_CHECKING:
     from yt.data_objects.static_output import Dataset
@@ -797,7 +801,7 @@ class YTDataContainer(abc.ABC):
             # tuples containing some sort of special "any" ParticleGroup
             unambiguous_fields_to_include = []
             unambiguous_fields_units = []
-            for field, field_unit in zip(fields_to_include, fields_units):
+            for field, field_unit in zip(fields_to_include, fields_units, strict=True):
                 if isinstance(field, tuple):
                     # skip tuples, they'll be checked with _determine_fields
                     unambiguous_fields_to_include.append(field)
@@ -849,7 +853,7 @@ class YTDataContainer(abc.ABC):
             field_names = []
 
             ## explicitly go after the fields we want
-            for field, units in zip(fields_to_include, fields_units):
+            for field, units in zip(fields_to_include, fields_units, strict=True):
                 ## Only interested in fields with the current particle type,
                 ## whether that means general fields or field tuples
                 ftype, fname = field
--- a/yt/data_objects/derived_quantities.py
+++ b/yt/data_objects/derived_quantities.py
@@ -1,3 +1,5 @@
+import sys
+
 import numpy as np
 
 from yt.funcs import camelcase_to_underscore, iter_fields
@@ -10,6 +12,9 @@ from yt.utilities.parallel_tools.parallel_analysis_interface import (
 )
 from yt.utilities.physical_constants import gravitational_constant_cgs
 from yt.utilities.physical_ratios import HUGE
+
+if sys.version_info < (3, 10):
+    from yt._maintenance.backports import zip
 
 
 def get_position_fields(field, data):
@@ -415,7 +420,7 @@ class WeightedStandardDeviation(DerivedQuantity):
         fields = list(iter_fields(fields))
         units = [self.data_source.ds._get_field_info(field).units for field in fields]
         rv = super().__call__(fields, weight)
-        rv = [self.data_source.ds.arr(v, u) for v, u in zip(rv, units)]
+        rv = [self.data_source.ds.arr(v, u) for v, u in zip(rv, units, strict=True)]
         if len(rv) == 1:
             rv = rv[0]
         return rv
@@ -431,7 +436,7 @@ class WeightedStandardDeviation(DerivedQuantity):
         my_var2s = [
             (data[weight].d * (data[field].d - my_mean) ** 2).sum(dtype=np.float64)
             / my_weight
-            for field, my_mean in zip(fields, my_means)
+            for field, my_mean in zip(fields, my_means, strict=True)
         ]
         return my_means + my_var2s + [my_weight]
 
@@ -623,7 +628,7 @@ class Extrema(DerivedQuantity):
         # The values get turned into arrays here.
         return [
             self.data_source.ds.arr([mis.min(), mas.max()])
-            for mis, mas in zip(values[::2], values[1::2])
+            for mis, mas in zip(values[::2], values[1::2], strict=True)
         ]
 
 
--- a/yt/data_objects/profiles.py
+++ b/yt/data_objects/profiles.py
@@ -1,3 +1,5 @@
+import sys
+
 import numpy as np
 from more_itertools import collapse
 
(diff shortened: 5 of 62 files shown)


yt/data_objects/profiles.py
```diff
@@ -1456,15 +1465,15 @@ def create_profile(
...
+    for f, n, (mi, ma), l in zip(bin_fields, n_bins, ex, logs, strict=True):
         if mi <= 0 and l:
             raise YTIllDefinedBounds(mi, ma)
         args += [f, n, mi, ma, l]
     kwargs = {"weight_field": weight_field}
     if cls is ParticleProfile:
         kwargs["deposition"] = deposition
     if override_bins is not None:
-        for o_bin, ax in zip(o_bins, ["x", "y", "z"]):
+        for o_bin, ax in zip(o_bins, ["x", "y", "z"], strict=False):
```
neutrinoceros, line 1476: This is one very case where it's apparent that `strict=False` is the intended behavior (found thanks to a test).

yt/data_objects/analyzer_objects.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, *args, **kwargs):
         # does not override
         if len(args) + len(kwargs) != len(self._params):
             raise RuntimeError
-        self.__dict__.update(zip(self._params, args))
+        self.__dict__.update(zip(self._params, args, strict=False))
```
neutrinoceros, line 17: from the context, it seems that `strict=False` is clearly intended here

yt/data_objects/derived_quantities.py
```diff
@@ -623,7 +630,7 @@ def reduce_intermediate(self, values):
         # The values get turned into arrays here.
         return [
             self.data_source.ds.arr([mis.min(), mas.max()])
-            for mis, mas in zip(values[::2], values[1::2])
+            for mis, mas in zip(values[::2], values[1::2], strict=True)
```
neutrinoceros, line 633: This is only safe if `len(values)%2==0` is guaranteed. If not, `strict=False` matches the existing behavior but is possibly buggy (one value is silently dropped)

neutrinoceros:
On second though this isn't a pure stylistic choice, as it *may* (however unlikely) have undesirable effects. Re-triaging as a refactor.

yt/frontends/stream/data_structures.py
```diff
@@ -422,7 +428,7 @@ def _set_code_unit_attributes(self):
             "magnetic_unit",
         )
         cgs_units = ("cm", "g", "s", "cm/s", "gauss")
-        for unit, attr, cgs_unit in zip(base_units, attrs, cgs_units):
+        for unit, attr, cgs_unit in zip(base_units, attrs, cgs_units, strict=True):
```
neutrinoceros, line 431: this is a good example of why enforcing `strict=True` helps finding bugs

yt/data_objects/profiles.py
```diff
@@ -1313,7 +1320,11 @@ def create_profile(
                 data_source.ds.field_info[f].sampling_type == "local"
                 for f in bin_fields + fields
             ]
-            is_local_or_pfield = [pf or lf for (pf, lf) in zip(is_pfield, is_local)]
+            if wf is not None:
+                is_local.append(wf.sampling_type == "local")
```
neutrinoceros, line 1324: highlighting this bit as a bugfix

yt/frontends/athena/io.py
```diff
@@ -18,8 +26,8 @@ class IOHandlerAthena(BaseIOHandler):
 
     def _field_dict(self, fhandle):
         keys = fhandle["field_types"].keys()
-        val = fhandle["field_types"].keys()
-        return dict(zip(keys, val))
+        val = fhandle["field_types"].values()
```
neutrinoceros, line 29: this is a case where the original code just *looks* wrong, but it's been there for 12 years so either it's not really a bug (but it should have a comment), or it's actually not used by anyone...

yt/utilities/fortran_utils.py
```diff
@@ -69,15 +75,7 @@ def read_attrs(f, attrs, endian="="):
...
-            if len(a) != len(v):
-                raise OSError(
-                    "An error occurred while reading a Fortran "
-                    "record. Record length is not equal to expected "
-                    "length: %s %s",
-                    len(a),
-                    len(v),
-                )
-            for k, val in zip(a, v):
+            for k, val in zip(a, v, strict=True):
```
neutrinoceros, line 78: not strictly backward compatible if we assume anyone relies on the exact exception type here, but I'm taking a bet that @cphyc doesn't mind.

neutrinoceros (commented):
Finishing my self-review with a couple notes for reviewers:
First of, if you're wondering why I write
```python
    pass
else:
    from yt._maintenance.backports import zip
```
instead of the simpler
```python
if sys.version_info < (3, 10):
    from yt._maintenance.backports import zip
```
It is because ruff ([rule `UP036`](https://docs.astral.sh/ruff/rules/outdated-version-block/)) can automatically cleanup the former but not the latter when we switch `python-requires` to `>=3.10`. So it's more verbose now but easier to cleanup in the future.

2) in a couple places, I actively changed the code around a `zip` to fix apparent bugs. These should be highlighted by my own comments.

3) Please pay special attention in cases where I set `strict=False`: in most cases, I saw no actual test failing with `strict=True` but the *context* around `zip` made me think `strict=False` was the intended behavior. I may be wrong.

4) Some `zip` calls were zipping some small literal lists, and I refactored them to as lists of tuples to avoid `zip` entirely, which I found was more readable. Maybe I made a mistake in a case or two, so please pay extra attention there too.

neutrinoceros:
I know this isn't exactly an exciting read, but can I ask @matthewturk, @chrishavlin or @cphyc for a review ?

chrishavlin:
Yes! I actually started it a few days ago but didn't finish up. Can definitely get to it in the next few days.

yt/data_objects/data_containers.py
```diff
@@ -797,7 +801,7 @@ def create_firefly_object(
             # tuples containing some sort of special "any" ParticleGroup
             unambiguous_fields_to_include = []
             unambiguous_fields_units = []
-            for field, field_unit in zip(fields_to_include, fields_units):
+            for field, field_unit in zip(fields_to_include, fields_units, strict=True):
```
matthewturk, line 804: If an error were to occur, I think this is a candidate location.

yt/data_objects/data_containers.py
```diff
@@ -849,7 +853,7 @@ def create_firefly_object(
             field_names = []
 
             ## explicitly go after the fields we want
-            for field, units in zip(fields_to_include, fields_units):
+            for field, units in zip(fields_to_include, fields_units, strict=True):
```
matthewturk, line 856: If an error were to occur, I think this is a candidate location.

yt/data_objects/derived_quantities.py
matthewturk: I think using strict here is the right move, as a check on argument inputs.

yt/data_objects/profiles.py
matthewturk: Thank you.  I have an inkling somewhere in my head that there's an extant issue this could be related to but searching does not prove me correct.

yt/data_objects/profiles.py
matthewturk: Great catch.  Yes, I think this is assuming that it'll run out of axes based on the dimensionality.

yt/fields/particle_fields.py
```diff
@@ -197,7 +197,7 @@ def _deposit_field(field, data):
                 validators=[ValidateSpatial(0)],
             )
 
-    for method, name in zip(("cic", "sum"), ("cic", "nn")):
+    for method, name in [("cic", "cic"), ("sum", "nn")]:
```
matthewturk, line 200: this is indeed an improvement

yt/frontends/art/io.py
```diff
@@ -125,7 +129,7 @@ def _get_field(self, field):
         if fname.startswith("particle_mass"):
             a = 0
             data = np.zeros(npa, dtype="f8")
-            for ptb, size, m in zip(pbool, sizes, self.ws):
+            for ptb, size, m in zip(pbool, sizes, self.ws, strict=True):
```
matthewturk, line 132: If an error were to occur, I think this is a candidate location.

yt/frontends/artio/definitions.py
```diff
@@ -27,7 +33,7 @@
     "stars": "STAR",
     "nbody": "N-BODY",
 }
-art_to_yt = dict(zip(yt_to_art.values(), yt_to_art.keys()))
+art_to_yt = dict(zip(yt_to_art.values(), yt_to_art.keys(), strict=True))
```
matthewturk, line 36: This is correct behavior -- but it makes me realize that we are using a dict for a 1:1 mapping, which is not necessarily true.  Fortunately it is here!

yt/frontends/athena/io.py
matthewturk: I don't see any references to it -- this may be vestigial.

yt/frontends/enzo_e/data_structures.py
```diff
@@ -475,7 +475,7 @@ def _set_code_unit_attributes(self):
             setdefaultattr(self, "velocity_unit", self.quan(k["uvel"], "cm/s"))
         else:
             p = self.parameters
-            for d, u in zip(("length", "time"), ("cm", "s")):
+            for d, u in [("length", "cm"), ("time", "s")]:
```
matthewturk, line 478: I can't help but feel like I probably initiated this idiom that you're undoing in several places ... and I think undoing it is the right thing!

yt/frontends/enzo_e/misc.py
```diff
@@ -89,7 +94,7 @@ def get_root_block_id(block, min_dim=3):
 
 def get_child_index(anc_id, desc_id):
     cid = ""
-    for aind, dind in zip(anc_id.split("_"), desc_id.split("_")):
+    for aind, dind in zip(anc_id.split("_"), desc_id.split("_"), strict=True):
```
matthewturk, line 97: if this errors it will be a very valuable error check

yt/frontends/tipsy/data_structures.py
```diff
@@ -122,7 +126,9 @@ def _parse_parameter_file(self):
         hvals = {
             a: c
             for (a, b), c in zip(
-                self._header_spec, struct.unpack(hh, f.read(struct.calcsize(hh)))
+                self._header_spec,
```
matthewturk, line 129: This one I'm less certain of, in that it may be the case that we don't want to force the specification of all fields and just let some go.  But, that's not necessarily what we should build in to doing this check, so :+1: .

yt/utilities/answer_testing/framework.py
```diff
@@ -769,9 +773,17 @@ def run(self):
         return result
 
     def compare(self, new_result, old_result):
-        for newp, oldp in zip(new_result["parents"], old_result["parents"]):
+        for newp, oldp in zip(
```
matthewturk, line 776: this is a good additional check

matthewturk: although, it may result in slightly less informative errors if the lengths are different

yt/utilities/fortran_utils.py
```diff
@@ -136,16 +132,7 @@ def read_cattrs(f, attrs, endian="="):
...
-                raise OSError(
-                    "An error occurred while reading a Fortran "
-                    "record. Record length is not equal to expected "
-                    "length: %s %s",
-                    len(a),
-                    len(v),
-                )
-
-            for k, val in zip(a, v):
+            for k, val in zip(a, v, strict=True):
```
matthewturk, line 135: I do think the other version was somewhat better, as otherwise we're left deducing what caused the iterators to be of different lengths.

matthewturk (approved):
I left some comments where I think there *might* be issues, but I want to note that I think there are no problems with any of this, and any bugs it finds would be an improvement.  Thanks for taking the time and leaving such helpful comments on it.

yt/frontends/athena/io.py
neutrinoceros: you're right. And it's private anyway so let me just remove it now.

yt/utilities/answer_testing/framework.py
neutrinoceros: not sure what you mean here: if lengths are different, there might be no error at all (without strict=True)

matthewturk: Oh, I just mean, that the old way wouldn't pick that up *as-is* but the new way will show a weird error.  It's NBD.

yt/utilities/fortran_utils.py
neutrinoceros: FTR I'm waiting on @chrishavlin to weigh in but happy to change it if he agrees

chrishavlin (approved):
Generally looks good! some minor questions/remarks so i'll hit approve now (assume approval if it gets dismissed!)

I'll also say I had a hard time evaluating the changes in the code for some of the frontends (particularly where the existing zips related to file structures....), but I think we have to trust (hope?) that the tests would catch errors there.

yt/utilities/fortran_utils.py
chrishavlin: i think i prefer keeping the OSError, but don't feel too strongly about it.

yt/data_objects/construction_data_containers.py
```diff
@@ -150,7 +154,9 @@ def _get_cut_mask(self, grid):
         mask = np.zeros(points_in_grid.sum(), dtype="int64")
         dts = np.zeros(points_in_grid.sum(), dtype="float64")
         ts = np.zeros(points_in_grid.sum(), dtype="float64")
-        for mi, (i, pos) in enumerate(zip(pids, self.positions[points_in_grid])):
+        for mi, (i, pos) in enumerate(
+            zip(pids, self.positions[points_in_grid], strict=True)
```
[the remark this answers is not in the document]
neutrinoceros: Good point. One could argue that in the presence of `strict=True` `pids` does play some role in terms of validation but it's convoluted and beyond the point. I imagine `pids` just came from copy-pasta. Let me clean this up now.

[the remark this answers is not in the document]
neutrinoceros: wait no, `i` (element from `pids`) *is* used in this loop. Not touching this here after all. (but I take your point that it is confusing !)

yt/data_objects/region_expression.py
```diff
@@ -166,7 +170,7 @@ def _create_region(self, bounds_tuple):
             if d is not None:
                 d = int(d)
             dims[ax] = d
-        center = [(cl + cr) / 2.0 for cl, cr in zip(left_edge, right_edge)]
+        center = [(cl + cr) / 2.0 for cl, cr in zip(left_edge, right_edge, strict=True)]
```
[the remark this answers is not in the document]
neutrinoceros: nice catch, let me try this

yt/frontends/athena/io.py
```diff
@@ -1,10 +1,16 @@
+import sys
+
 import numpy as np
 
 from yt.funcs import mylog
 from yt.utilities.io_handler import BaseIOHandler
 
 from .data_structures import chk23
 
+if sys.version_info < (3, 10):
```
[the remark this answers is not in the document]
neutrinoceros: thanks for catching this ! clearly vestigial from a auto-fixing

yt/utilities/fortran_utils.py
neutrinoceros: reverted !

neutrinoceros:
Can I get another approval ? Seems that I dismissed the ones I got when I pushed another clean up commit

chrishavlin (approved):

neutrinoceros:
Let's get this in before it starts rotting

PR #4941, opened 2024-07-08
DEP: make ipywidgets/IPython optional dependencies again

### LINKED ISSUE #4054: Rich display is broken on recent version of ipython

neutrinoceros:
## PR Summary

Ipywidgets and IPython are meant to be optional dependencies that we merely hook into if they are installed and actively used, however they were promoted to hard dependencies as a workaround in #4155, which I now think can be avoided.

I think this may close #4054 (but will ask @cphyc for confirmation)

## PR Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,6 @@ requires-python = ">=3.9.2"
 dependencies = [
     "cmyt>=1.1.2",
     "ewah-bool-utils>=1.2.0",
-    "ipywidgets>=8.0.0",
     "matplotlib>=3.5",
     "more-itertools>=8.4",
     "numpy>=1.19.3, <3", # keep minimal requirement in sync with NPY_TARGET_VERSION
@@ -136,7 +135,8 @@ full = [
     "cartopy>=0.22.0",
     "firefly>=3.2.0",
     "glueviz>=0.13.3",
-    "ipython>=2.0.0",
+    "ipython>=7.16.2",
+    "ipywidgets>=8.0.0",
     "miniballcpp>=0.2.1",
     "mpi4py>=3.0.3",
     "pandas>=1.1.2",
@@ -210,7 +210,6 @@ mapserver = [
 minimal = [
     "cmyt==1.1.2",
     "ewah-bool-utils==1.2.0",
-    "ipywidgets==8.0.0",
     "matplotlib==3.5",
     "more-itertools==8.4",
     "numpy==1.19.3",
--- a/yt/_maintenance/ipython_compat.py
+++ b/yt/_maintenance/ipython_compat.py
@@ -1,0 +1,31 @@
+from importlib.metadata import version
+from importlib.util import find_spec
+
+from packaging.version import Version
+
+__all__ = [
+    "IS_IPYTHON",
+    "IPYWIDGETS_ENABLED",
+]
+
+IS_IPYTHON: bool
+HAS_IPYWIDGETS_GE_8: bool
+IPYWIDGETS_ENABLED: bool
+
+
+try:
+    # this name is only defined if running within ipython/jupyter
+    __IPYTHON__  # type: ignore [name-defined] # noqa: B018
+except NameError:
+    IS_IPYTHON = False
+else:
+    IS_IPYTHON = True
+
+
+HAS_IPYWIDGETS_GE_8 = (
+    Version(version("ipywidgets")) >= Version("8.0.0")
+    if find_spec("ipywidgets") is not None
+    else False
+)
+
+IPYWIDGETS_ENABLED = IS_IPYTHON and HAS_IPYWIDGETS_GE_8
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -23,6 +23,7 @@ from unyt import Unit, UnitSystem, unyt_quantity
 from unyt.exceptions import UnitConversionError, UnitParseError
 
 from yt._maintenance.deprecation import issue_deprecation_warning
+from yt._maintenance.ipython_compat import IPYWIDGETS_ENABLED
 from yt._typing import (
     AnyFieldKey,
     AxisOrder,
@@ -119,10 +120,7 @@ class MutableAttribute:
         # We can assume that ipywidgets will not be *added* to the system
         # during the course of execution, and if it is, we will not wrap the
         # array.
-        if display_array and find_spec("ipywidgets") is not None:
-            self.display_array = True
-        else:
-            self.display_array = False
+        self.display_array = display_array and IPYWIDGETS_ENABLED
 
     def __get__(self, instance, owner):
         return self.data.get(instance, None)
--- a/yt/fields/field_type_container.py
+++ b/yt/fields/field_type_container.py
@@ -116,6 +116,7 @@ class FieldNameContainer:
         return False
 
     if IPYWIDGETS_ENABLED:
+        # for discussion of this class-level conditional: https://github.com/yt-project/yt/pull/4941
 
         def _ipython_display_(self):
             import ipywidgets
--- a/yt/fields/field_type_container.py
+++ b/yt/fields/field_type_container.py
@@ -7,6 +7,7 @@ import textwrap
 import weakref
 from functools import cached_property
 
+from yt._maintenance.ipython_compat import IPYWIDGETS_ENABLED
 from yt.fields.derived_field import DerivedField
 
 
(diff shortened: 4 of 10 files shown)


pyproject.toml
```diff
@@ -134,7 +133,8 @@ full = [
     "cartopy>=0.22.0",
     "firefly>=3.2.0",
     "glueviz>=0.13.3",
-    "ipython>=2.0.0",
+    "ipython>=7.16.2",
```
neutrinoceros, line 136: While I was at it, I bumped minimal requirement for IPython here to the oldest version that supported Python 3.9 (i.e. the first one that was released *after* Python 3.9.0)

matthewturk: haha, I love that we had it at 2.0 for a bajillion years. :)

yt/_maintenance/ipython_compat.py
```diff
@@ -0,0 +1,31 @@
...
+IPYWIDGETS_ENABLED: bool
+
+
+try:
+    # this name is only defined if running within ipython/jupyter
+    __IPYTHON__  # type: ignore [name-defined] # noqa: B018
+except NameError:
+    IS_IPYTHON = False
+else:
+    IS_IPYTHON = True
```
matthewturk, line 22: Cool!

yt/data_objects/static_output.py
```diff
@@ -119,10 +120,7 @@ def __init__(self, display_array=False):
         # We can assume that ipywidgets will not be *added* to the system
         # during the course of execution, and if it is, we will not wrap the
         # array.
-        if display_array and find_spec("ipywidgets") is not None:
-            self.display_array = True
-        else:
-            self.display_array = False
+        self.display_array = display_array and IPYWIDGETS_ENABLED
```
matthewturk, line 123: If I am remembering correctly, some of this was added to avoid calling `find_spec` too many times, which led to a pretty bonkers number of import checks at runtime.  Can you assure that this does not re-introduce that performance hit?

yt/fields/field_type_container.py
```diff
@@ -112,46 +115,54 @@ def __contains__(self, obj):
...
-            def _change_field(event):
-                fobj = getattr(_ftype, event["new"])
-                _box.clear_output()
-                with _box:
-                    display(
-                        Markdown(
-                            data="```python\n"
-                            + textwrap.dedent(fobj.get_source())
-                            + "\n```"
+    if IPYWIDGETS_ENABLED:
```
matthewturk, line 118: I really don't like this conditional in the class body.  Would you be open to changing it into a decorator or something?

yt/data_objects/static_output.py
neutrinoceros: Yup. `IPYWIDGETS_ENABLED` is only evaled once.

yt/fields/field_type_container.py
neutrinoceros: I don't mind a decorator, but how should it behave ? We originally promoted ipywidgets as a hard dependency so this method wouldn't raise if it was missing, but I don't see what else a decorator could do (except maybe change the function name at runtime so it would effectively behave the same as my conditional... just more obfuscated)

matthewturk: I'm really feeling anxious about the conditional inside the class body -- in general we don't have places where we evaluate code (other than decorators and traitlet initializers) inside class bodies.  It's not against the rules to do it, per se, but it feels out of place and the additional indent gets in the way; we often have monkey-patched in the past.

But in looking at it, I'm starting to think that we don't need to conditionally define it at all.  If `_ipython_display_` is being called, it's because it's being evaluated inside an ipython environment (or someone is deliberately trying to do something.)  So at that point, I think we could evaluate whether `IPYWIDGETS_ENABLED` is true (inside the function body) and if not, fall back to `_repr_`.

neutrinoceros: > So at that point, I think we could evaluate whether IPYWIDGETS_ENABLED is true (inside the function body) and if not, fall back to _repr_.

This *would* be a sane approach but my understanding is that it's not feasible, since `_ipython_repr_` is supposed to work via side effects, and `__repr__` works by returning a value.

Here's the exact quote from [`_ipython_repr_`'s documentation](https://ipython.readthedocs.io/en/stable/config/integrating.html#MyObject._ipython_display_)
> Displays the object as a side effect; the return value is ignored. If this is defined, all other display methods are ignored.

matthewturk: Good point -- to be honest, I got this idea from another place pointing out that any errors at all defaulted back to calling `_repr_`.  So in principle ... if we just threw an error it would likely swallow that and call `_repr_`...

neutrinoceros: Minimal testing seems to indicate this doesn't work: trying to import `ipywidgets` if it's not installed *will* raise an exception (which is the original motivation for making it a hard dep). Or is there a special exception type we should raise instead ? Can you remember where this is coming from ?

chrishavlin: What about pointing `_ipython_display_` to a callable dynamically during `__init__` or (`__new__`?)? 

i.e., rename the current `_ipython_display_` to something else :

```python
    def _ipython_display_when_needed_(self): # or some other name
        import ipywidgets
        from IPython.display import display 
        ... etc ... 
```

Then in `__init__` or `__new__`:

```python
    def __init__(self):         
        if IPYWIDGETS_ENABLED:
            setattr(self, '_ipython_display_', '_ipython_display_when_needed_')
``` 

this would avoid the conditional at the class body level and would avoid the import error when ipywidget is not installed.

neutrinoceros: This works. In my opinion it's less clear but maybe that's a possible compromise. What do you think @matthewturk ?

matthewturk: I want to take a step back and say, I don't think that my disinclination is necessarily a dealbreaker.  Already we monkey-patch some stuff, which is just the same concept but at a different level.  So, whatever folks are OK with.

neutrinoceros: @chrishavlin can have the final say as far as I'm concerned :)

chrishavlin: hah, ok :) then IMO I think sticking with what you already have is the way to go -- I do agree with Matt that having the conditional logic at the class level is not ideal. it does make it a bit harder to read what's going on there... but I do think it will make it easier to maintain going forward compared to my suggestion of dynamically linking the function, which kinda hides the logic away.

chrishavlin: though i do think adding a comment as to why it's at the class level would be helpful in the future.
 
```suggestion
    if IPYWIDGETS_ENABLED:
        # for discussion of this class-level conditional: https://github.com/yt-project/yt/pull/4941
```

chrishavlin (approved):

chrishavlin (approved):

neutrinoceros:
rebased to resolve a conflict. @chrishavlin, @matthewturk, can we get this in ?

PR #4944, opened 2024-07-09
DEP: stop requiring ewah-bool-utils at runtime

neutrinoceros:
## PR Summary

Scratching a long standing itch of mine: we only ever need a single from ewah-bool-utils at runtime and the import is done in a private module, so I feel like having the whole package as a dependency is a bit wasteful. We should really only need it at compile-time. 

## PR Checklist

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -102,8 +102,8 @@ jobs:
       # keep in sync with pyproject.toml
       run: |
         python -m pip install "Cython>=3.0.3, <3.1"
-        python -m pip install numpy>=1.25
-        python -m pip install ewah-bool-utils>=1.0.2
+        python -m pip install "numpy>=2.0.0"
+        python -m pip install "ewah-bool-utils>=1.2.0"
         python -m pip install --upgrade wheel
         python -m pip install --upgrade setuptools
     - name: build yt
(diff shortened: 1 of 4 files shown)


neutrinoceros:
This simply doesn't work, I must have messed up when I tried it locally.

neutrinoceros:
>  But I also am not sure I see why it wouldn't be a runtime dependency, since it's got runtime code that gets executed -- the bool array collections are actually utilized in our cython, and they aren't all compiletime.

I'm just trying to eliminate dependencies where I can. This one in particular *looks* like it could be eliminated since it boils down to a single (private) `import` statement, so I regularly find myself wondering if and how I could scratch it. But it may not be possible at all.

matthewturk:
Sure, makes sense.  I *think* that it's likely going to be tricky because we make runtime calls to methods defined on `mask` inside `get_ghost_zones` for instance.  Some of these calls are to `cdef` methods and others to `def`...  I think, but I'm not sure, that this means that the Cython import semantics will keep us from getting away with this.

If we were able to only conditionally import the particle support, it might get us further, but I believe that any time we want to have top-level validation of the type we might need to be able to import.  I'm not 100% on this.  But because it's Cython and not Python I think that's the end result.

PR #4945, opened 2024-07-18
Fix memory leakage

cindytsai:
## PR Summary

This is a bug that causes memory leakage in projections.

It forgot to free `vorig` when returning inside the if block.
Since `vorig` won't be used inside that if block anyway, it should be moved under the if block.


chrishavlin (approved):
nice catch!

matthewturk:
I'd like to just add here that @cindytsai was *really* diligent in tracking this down, and really put a lot of effort into finding it.  Thank you, @cindytsai !

PR #4948, opened 2024-07-20
BLD: fix compatibility with Cython 3.1

### LINKED ISSUE #4044: BLD: Future incompatibility with Cython 3.1
### Bug report

**Bug summary**

Building yt's Cython extensions with Cython 3.0.0a10 revealed that some modules are currently relying on deprecated API, as for instance

```
warning: yt/utilities/lib/image_samplers.pxd:18:0: The 'DEF' statement is deprecated and will be removed in a future Cython version. Consider using global variables, constants, and in-place literals instead. See https://github.com/cython/cython/issues/4310
```

Namely, we currently use `IF` and `DEF` here and there, which are going away at some point in a future version of Cython
(see https://github.com/cython/cython/issues/4310)
I estimate that we have 16 occurrences of `DEF` and 2 `IF`, so I'm hopeful that refactoring all of them is feasible

neutrinoceros:
## PR Summary

Close #4044
Following [Cython's migration guide](https://docs.cython.org/en/latest/src/userguide/migrating_to_cy30.html#deprecation-of-def-if)

### DIFF
--- a/yt/geometry/particle_deposit.pxd
+++ b/yt/geometry/particle_deposit.pxd
@@ -14,12 +14,13 @@ import numpy as np
 cimport cython
 from libc.math cimport sqrt
 from libc.stdlib cimport free, malloc
-from numpy.math cimport PI as NPY_PI
 
 from yt.utilities.lib.fp_utils cimport *
 
 from .oct_container cimport Oct, OctreeContainer
 
+cdef extern from "numpy/npy_math.h":
+    double NPY_PI
 
 cdef extern from "platform_dep.h":
     void *alloca(int)
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -50,9 +50,9 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install --upgrade setuptools wheel
-        python -m pip install --pre --only-binary ":all:" numpy matplotlib \
+        python -m pip install --pre --only-binary ":all:" numpy matplotlib Cython \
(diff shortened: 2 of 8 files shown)


chrishavlin (dismissed):

yt/geometry/particle_deposit.pxd
```diff
@@ -14,12 +14,13 @@ import numpy as np
...
 from libc.math cimport sqrt
 from libc.stdlib cimport free, malloc
-from numpy.math cimport PI as NPY_PI
 
 from yt.utilities.lib.fp_utils cimport *
 
 from .oct_container cimport Oct, OctreeContainer
 
+cdef extern from "numpy/npy_math.h":
+    double NPY_PI
```
neutrinoceros, line 23: To see why this change was needed, see https://cython.readthedocs.io/en/latest/src/changes.html#other-changes

> The previously shipped NumPy C-API declarations (cimport numpy) were removed. NumPy has been providing version specific declarations for several versions now. (Github issue [#5842](https://github.com/cython/cython/issues/5842))

chrishavlin:
This looks good to me, but since it's been a while and the sph projections touched a lot of cython, I think it'd be nice to merge with main here to double check that we haven't introduced any additional compatibility issues. What do you think?

neutrinoceros:
rebased on the tip of main !

yt/geometry/particle_deposit.pxd
neutrinoceros: (`NPY_2_PI` is also available)

PR #4949, opened 2024-07-21
MNT: cleanup cibuildwheel configuration and move it to `pyproject.toml`

neutrinoceros:
## PR Summary

This should help testing wheel building locally, and was also an opportunity to clean up our cibuildwheel configuration.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -35,16 +35,6 @@ jobs:
         uses: pypa/cibuildwheel@v2.19.1
         with:
           output-dir: dist
-        env:
-          CIBW_BUILD: "cp39-* cp310-* cp311-* cp312-*"
-          CIBW_SKIP: "*-musllinux_*"  #  numpy doesn't have wheels for musllinux so we can't build some quickly and without bloating
-          CIBW_ARCHS_LINUX: "x86_64"
-          CIBW_ARCHS_MACOS: auto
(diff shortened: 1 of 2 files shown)


.github/workflows/wheels.yaml
```diff
@@ -35,16 +35,6 @@ jobs:
         uses: pypa/cibuildwheel@v2.19.1
         with:
           output-dir: dist
-        env:
-          CIBW_BUILD: "cp39-* cp310-* cp311-* cp312-*"
-          CIBW_SKIP: "*-musllinux_*"  #  numpy doesn't have wheels for musllinux so we can't build some quickly and without bloating
-          CIBW_ARCHS_LINUX: "x86_64"
-          CIBW_ARCHS_MACOS: auto
-          MACOSX_DEPLOYMENT_TARGET: "10.9" # as of CIBW 2.9, this is the default value, pin it so it can't be bumped silently
```
neutrinoceros, line 43: I removed this option entirely because Python 3.13 will only support MacOS 10.13 and above, and cibuildwheel already does the right thing for us so it's not worth constraining it
See https://github.com/pypa/cibuildwheel/blob/6c5cf878718641c3cbf21d69c670ff2e53a8263c/cibuildwheel/macos.py#L297-L308

chrishavlin (approved):

neutrinoceros:
I'll merge this now to help working on #4953 locally

PR #4951, opened 2024-07-21
CLN: cleanup unused internal conf parameter `within_pytest`

neutrinoceros:
## PR Summary
This doesn't seem to be used anywhere and the release notes for pytest 8.3.0 make it clear that it's even redundant so we shouldn't ever actually need it.
See https://github.com/pytest-dev/pytest/pull/12153

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -85,7 +85,6 @@ def pytest_configure(config):
     Reads in the tests/tests.yaml file. This file contains a list of
     each answer test's answer file (including the changeset number).
     """
-    ytcfg["yt", "internals", "within_pytest"] = True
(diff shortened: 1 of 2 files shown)


chrishavlin:
> the release notes for pytest 8.3.0 make it clear that it's even redundant

Should we bump the minimum pytest version for yt testing? It's currently 6.1.

neutrinoceros:
No need to, 8.3.0 just added documentation for a mechanism that already existed, and which we're not using the anyway.

chrishavlin (approved):

jzuhone (approved):

Issue #4931, opened 2024-06-22
CPython 3.13 support (tracker issue)


neutrinoceros:
This issue is to plan and keep track on the effort to add support for CPython 3.13 .
See [PEP 719](https://peps.python.org/pep-0719/) for the release schedule.
At the time of opening, version `3.13.0b2` is available, and NumPy is already publishing nightly wheels for it.

Here's the chronological task list to complete before we can declare support for it.
- [x] wait for our core dependencies to publish nightly builds, or be tested
	- [x] numpy
	- [x] matplotlib https://github.com/matplotlib/matplotlib/pull/28293
      - [x] Pillow
      - [x] contourpy
	- [x] unyt https://github.com/yt-project/unyt/pull/507 (not a real requirement: unyt 3.0.3 appear to be compatible already)
	- [x] cmyt https://github.com/yt-project/cmyt/pull/178
	- [x] ewah-bool-utils https://github.com/yt-project/ewah_bool_utils/pull/108
- [x] ~~https://github.com/yt-project/yt/pull/4958~~ replaced by #5021 
- [x] pass all tests on ubuntu
- [x] test `cp313` wheels builds for all platform in CI (but don't publish them yet)
	- [x] this is helped by #4949 
	- [x] ... and #4971
	- [x] #4973
- [x] enable publication to PyPI 
	- [x] requires ABI stability, i.e., CPython 3.13.0rc1
	- [x] #4973
- [ ] #5021

Supporting the experimental free-threading (a.k.a "nogil") build is a non-goal for this issue.
Relevant PRs are to be linked in this message.

chrishavlin:
updated the checklist above to reflect 5021 replacing 4958.

PR #4958, opened 2024-07-26
TST: weekly tests against Python 3.13

neutrinoceros:
## PR Summary

Partly address https://github.com/yt-project/yt/issues/4931 now that our core dependencies are publishing nightlies.

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -40,7 +40,7 @@ jobs:
         # the '-dev' suffix allows to use pre-releases if no final release is available yet
         # this version should be upgraded as often as possible, typically once a year when
         # Cython, numpy and matplotlib are known to be compatible
-        python-version: '3.12-dev'
+        python-version: '3.13-dev'
 
     - name: Install dependencies
       # PyYAML needs to be installed in isolation for now, because of a known


neutrinoceros:
So it's failing at build time. I don't immediately understand why, but it looks like something is off with numpy's nightlies.
The good news is that I can reproduce locally, albeit on a different arch. Will dig into this.

neutrinoceros:
ah, the important change rather seems to be that I tried using Cython nightlies. It compiles fine (at least locally) with Cython 3.0.10. Let's try this.

neutrinoceros:
I addressed the compatibility issue for Cython 3.1 in #4948

.github/workflows/bleeding-edge.yaml
```diff
@@ -40,16 +40,19 @@ jobs:
         # the '-dev' suffix allows to use pre-releases if no final release is available yet
         # this version should be upgraded as often as possible, typically once a year when
         # Cython, numpy and matplotlib are known to be compatible
-        python-version: '3.12-dev'
+        python-version: '3.13-dev'
 
     - name: Install dependencies
+      # pre-installing kiwisolver (mpl dependency) because it doesn't have wheels
```
chrishavlin, line 46: looks like kiwisolver actually released since you wrote this and now has wheels available (https://pypi.org/project/kiwisolver/1.4.7/).

neutrinoceros: indeed. Let me remove this bit as I rebase the branch

neutrinoceros:
superseded by #5021

Issue #4963, opened 2024-08-12
Tipsy Frontend BUG?: oddities with smoothing lengths/positions and bounding box


nastasha-w:
### Bug report

**Bug summary**

For some reason, the Tipsy dataset ([YT Hub](https://hub.yt/#item/579fb0a77b6f0800011ea19f)) doesn't have the ("Gas", "smoothing_length") field available if I load it without a `bounding_box` argument. This is even though even though it's in the dataset's `ds.field_list`. 
In addition, it seems like the position of at least one SPH particle depends on what `bounding_box` argument the dataset is loaded with.

These issues first came up in the discussion of #4939.

**Code for reproduction: missing `smoothing_length`**

```python
(ytdev_main) nastasha@dhcp-165-124-148-182 code % ipython
Python 3.12.4 (main, Jun  6 2024, 18:26:44) [Clang 15.0.0 (clang-1500.3.9.4)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.26.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: sampledatadir = "/Users/nastasha/code/ytdev_misc/sample_data/"

In [2]: datasetn = sampledatadir + "TipsyGalaxy/galaxy.00300"

In [3]: import yt

In [4]: import numpy as np

In [5]: ds = yt.load(datasetn, bounding_box=None)
yt : [INFO     ] 2024-08-12 13:24:31,222 Parameters: current_time              = 20.100000000000044
yt : [INFO     ] 2024-08-12 13:24:31,222 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2024-08-12 13:24:31,222 Parameters: domain_left_edge          = None
yt : [INFO     ] 2024-08-12 13:24:31,222 Parameters: domain_right_edge         = None
yt : [INFO     ] 2024-08-12 13:24:31,222 Parameters: cosmological_simulation   = 0.0
yt : [INFO     ] 2024-08-12 13:24:31,232 Allocating KDTree for 76962 particles
Generate smoothing length: 100%|███████████████████████████▉| 76961/76962 [00:00<00:00, 102496.20it/s]
yt : [INFO     ] 2024-08-12 13:24:32,044 Allocating for 3.154e+05 particles

In [7]: ds.field_list
Out[7]: 
[('DarkMatter', 'Coordinates'),
 ('DarkMatter', 'Epsilon'),
 ('DarkMatter', 'FeMassFrac'),
[...]

PR #4964, opened 2024-08-15
BUG: tipsy, reading just smoothing length

chrishavlin:
Partial fix for #4963 

This fixes the error when reading the smoothing length for tipsy. Does not address the question of the changing coordinate values when the bounding box changes.


jzuhone (approved):

neutrinoceros:
@yt-fido test this please

PR #4979, opened 2024-09-03
DEP: drop support for Python 3.9

### LINKED ISSUE #4977: DEP: can we drop support for Python 3.9 now ?
It is that time of the year were a new version of Python is (about to be) born, and we can rotate out the oldest-supported one.
This one however is a bit special: Python 3.9 was the last version to work out-of-the-box with the now long-discontinued nosetest.
We already have a manually patched version of nosetest running on Jenkins with Python 3.10 so we know that's doable, but dropping Python 3.9 means it will become even harder to reproduce failures locally.
However, since there hasn't been much progress on the front of migrating our oldest tests to pytest, running the patched version of nose might still be the way to go for now.

@Xarthisius, is your patch publicly visible anywhere ? It would be useful to add it to the repo so it can be used on GHA CI too.

neutrinoceros:
## PR Summary
Tentatively close #4977 (the nose patching bit is experimental at the time of opening)

this is currently based off #4941 to avoid a conflict


matthewturk:
Is anyone else still supporting py39?

neutrinoceros:
Yeah, I mean, Cython still hasn't dropped 2.7 yet (at least not in a stable release)
More seriously, matplotlib still supports 3.9 but NumPy doesn't. SPEC 0 recommends dropping 3.10 next month so I don't feel too guilty about proposing this now.

matthewturk:
That works for me.  And let's get #4941 in first so this is a smaller one.

neutrinoceros:
This is ready for review. As usual with these, it'll require some voodoo from a code owner (@matthewturk, @Xarthisius)

neutrinoceros:
rebased to fix conflicts

jzuhone (approved):

neutrinoceros:
@yt-fido test this please

neutrinoceros:
Thanks @Xarthisius !

PR #4981, opened 2024-09-06
BUG: fix orientation of colorbar in multiplanel plot

### LINKED ISSUE #4382: multiplot_export_to_mpl gets colorbar orientation wrong
### Bug report

**Bug summary**

If I rerun the `multiplot_export_to_mpl.py` cookbook example, but put the colorbars at the bottom, then
the orientation of the colorbar is set incorrectly.

**Code for reproduction**

This is a 1-line change from the cookbook:

```python
import yt

fields = [
    ("gas", "density"),
    ("gas", "velocity_x"),
    ("gas", "velocity_y"),
    ("gas", "velocity_magnitude"),
]

p = yt.SlicePlot(ds, "z", fields)
p.set_log(("gas", "velocity_x"), False)
p.set_log(("gas", "velocity_y"), False)

# this returns a matplotlib figure with an ImageGrid and the slices
# added to the grid of axes (in this case, 2x2)
fig = p.export_to_mpl_figure((2, 2), cbar_location="bottom", cbar_pad="10%")

fig.tight_layout()

fig.savefig("multiplot_export_to_mpl.png")
```

```python
# Paste your code here
#
#
```

**Actual outcome**

Here's the output

multiplot_export_to_mpl

```
# If applicable, paste the console output here
#
#
```

**Expected outcome**

The orientation of the colorbar data appears to be vertical, even though the colorbar itself is presented
horizontally.  I can't seem to find where in yt this is done.

**Version Information**

  * Operating System:
  * Python Version:
  * yt version:

xshaokun:
## PR Summary
fix #4382

## PR Checklist

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -353,7 +353,7 @@ class ImagePlotMPL(PlotMPL, ABC):
                     "Colorbar orientation would be wrong in the current Matplotlib version (< 3.7.0)",
                     stacklevel=6,
                 )
-                self.cb = self.figure.colorbar(self.image, self.cax)
+            self.cb = self.figure.colorbar(self.image, self.cax)
 
         cb_axis: Axis
         if self.cb.orientation == "vertical":
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -343,14 +343,15 @@ class ImagePlotMPL(PlotMPL, ABC):
         self.cax.tick_params(which="both", direction="in")
 
         # For creating a multipanel plot by ImageGrid
-        # we may need the location keyword, which was introduced since Matplotlib 3.7.0
+        # we may need the location keyword, which requires Matplotlib >= 3.7.0
         cb_location = getattr(self.cax, "orientation", None)
         if matplotlib.__version_info__ >= (3, 7):
             self.cb = self.figure.colorbar(self.image, self.cax, location=cb_location)
         else:
             if cb_location in ["top", "bottom"]:
                 warnings.warn(
-                    "Colorbar orientation would be wrong in the current Matplotlib version (< 3.7.0)",
+                    "Cannot properly set the orientation of colorbar. "
+                    "Consider upgrading matplotlib to version 3.7 or newer",
                     stacklevel=6,
                 )
             self.cb = self.figure.colorbar(self.image, self.cax)
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -342,10 +342,19 @@ class ImagePlotMPL(PlotMPL, ABC):
 
         self.cax.tick_params(which="both", direction="in")
 
-        # For creating a multipanel plot by ImageGrid, we need the location keyword
-        self.cb = self.figure.colorbar(
-            self.image, self.cax, location=getattr(self.cax, "orientation", None)
-        )
+        # For creating a multipanel plot by ImageGrid
+        # we may need the location keyword, which was introduced since Matplotlib 3.7.0
+        cb_location = getattr(self.cax, "orientation", None)
+        if matplotlib.__version_info__ >= (3, 7):
+            cb_kwargs = {"location": cb_location}
+        else:
+            cb_kwargs = {}
+            if cb_location in ["top", "bottom"]:
+                warnings.warn(
+                    "Colorbar orientation would be wrong in the current Matplotlib version (< 3.7.0)",
+                    stacklevel=2,
+                )
+        self.cb = self.figure.colorbar(self.image, self.cax, **cb_kwargs)
 
         cb_axis: Axis
         if self.cb.orientation == "vertical":
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -352,7 +352,7 @@ class ImagePlotMPL(PlotMPL, ABC):
             if cb_location in ["top", "bottom"]:
                 warnings.warn(
                     "Colorbar orientation would be wrong in the current Matplotlib version (< 3.7.0)",
-                    stacklevel=2,
+                    stacklevel=6,
                 )
         self.cb = self.figure.colorbar(self.image, self.cax, **cb_kwargs)
 
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -540,7 +540,9 @@ class ImagePlotMPL(PlotMPL, ABC):
         labels = super()._get_labels()
         cbax = self.cb.ax
         labels += cbax.yaxis.get_ticklabels()
+        labels += cbax.xaxis.get_ticklabels()
         labels += [cbax.yaxis.label, cbax.yaxis.get_offset_text()]
+        labels += [cbax.xaxis.label, cbax.xaxis.get_offset_text()]
         return labels
 
     def hide_axes(self, *, draw_frame=None):
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -538,11 +538,12 @@ class ImagePlotMPL(PlotMPL, ABC):
 
     def _get_labels(self):
         labels = super()._get_labels()
-        cbax = self.cb.ax
-        labels += cbax.yaxis.get_ticklabels()
-        labels += cbax.xaxis.get_ticklabels()
-        labels += [cbax.yaxis.label, cbax.yaxis.get_offset_text()]
-        labels += [cbax.xaxis.label, cbax.xaxis.get_offset_text()]
+        if getattr(self.cb, "orientation", "vertical") == "horizontal":
+            cbaxis = self.cb.ax.xaxis
+        else:
+            cbaxis = self.cb.ax.yaxis
+        labels += cbaxis.get_ticklabels()
+        labels += [cbaxis.label, cbaxis.get_offset_text()]
         return labels
 
     def hide_axes(self, *, draw_frame=None):
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -341,7 +341,11 @@ class ImagePlotMPL(PlotMPL, ABC):
         self.image.axes.set_facecolor(self.colorbar_handler.background_color)
 
         self.cax.tick_params(which="both", direction="in")
-        self.cb = self.figure.colorbar(self.image, self.cax)
+
+        # For creating a multipanel plot by ImageGrid, we need the location keyword
+        self.cb = self.figure.colorbar(
+            self.image, self.cax, location=getattr(self.cax, "orientation", None)
+        )
 
         cb_axis: Axis
         if self.cb.orientation == "vertical":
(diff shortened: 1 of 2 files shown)


yt/visualization/base_plot_types.py
```diff
@@ -341,7 +341,15 @@ def _set_axes(self) -> None:
...
-        self.cb = self.figure.colorbar(self.image, self.cax)
+        from mpl_toolkits.axes_grid1.axes_grid import CbarAxesBase
+
+        if isinstance(self.cax, CbarAxesBase):
+            # For creating a multipanel plot by ImageGrid
+            self.cb = self.figure.colorbar(
+                self.image, self.cax, location=self.cax.orientation
+            )
+        else:
+            self.cb = self.figure.colorbar(self.image, self.cax)
```
neutrinoceros, line 352: Let's try minimizing the repetition if we can
```suggestion
            cb_kwargs = {"location": self.cax.orientation}
        else:
            cb_kwargs = {}
        self.cb = self.figure.colorbar(self.image, self.cax, **cb_kwargs)
```

yt/visualization/base_plot_types.py
```diff
@@ -341,7 +341,15 @@ def _set_axes(self) -> None:
         self.image.axes.set_facecolor(self.colorbar_handler.background_color)
 
         self.cax.tick_params(which="both", direction="in")
-        self.cb = self.figure.colorbar(self.image, self.cax)
+        from mpl_toolkits.axes_grid1.axes_grid import CbarAxesBase
```
neutrinoceros, line 344: is this import statement known expensive (long startup time) or could it be moved to the top level of the module ?

neutrinoceros (commented):
Great !
Ideally we need an image test for this, here's the guide https://yt-project.org/docs/dev/developing/testing.html#creating-and-updating-image-baselines-for-pytest-mpl-tests

yt/visualization/base_plot_types.py
xshaokun: The result from `timeit` is about 2 seconds. Would it be considered as expensive?

neutrinoceros: Very much so, yes. I consider anything longer than .1s to me expensive for an import statement.

chrishavlin:
The minimal py3.9.2 test build is failing because it uses yt's minimally supported matplotlib (3.5.0), for which colorbar does not have a location kwarg. Looks like the location kwarg was introduced in matplotlib 3.7.0([release notes here](https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.7.0.html#figure-and-axes-layout)). IMO it's OK for this bug to persist in yt with the minimally supported matplotlib, so the easiest thing to do is just add a check for matplotlib version before passing in a location kwarg.

neutrinoceros:
Maybe we could even emit a warning if we can detect that the colorbar is horizontal and matplotlib is too old.

xshaokun:
I have a question regarding writing a test. Since the bug persists for older versions of MPL and a warning is emitted, the same test will generate different baseline images in new and old versions of MPL, but they will have the same file name. This prevents me from updating images in `pytest_mpl_baseline`. Do you have any better suggestions for handling this?

neutrinoceros:
Yes, image tests are tricky an hardly portable; in fact we only run them on a specific runner image (`ubuntu-latest`) and against the latest version of matplotlib. You shouldn't (and cannot) generate a new baseline image on your own machine, as it might not exactly match what is produced in CI, as previous experiments showed that the only diff tolerance we can support is 0.

Good job on adding a `@pytest.mark.skipif` marker anyway, this is helpful documentation of how much tension we have to bump our requirement on matplotlib !

yt/visualization/base_plot_types.py
```diff
@@ -341,7 +341,20 @@ def _set_axes(self) -> None:
...
+        # we may need the location keyword, which was introduced since Matplotlib 3.7.0
+        cb_location = getattr(self.cax, "orientation", None)
+        if matplotlib.__version_info__ >= (3, 7):
+            cb_kwargs = {"location": cb_location}
+        else:
+            cb_kwargs = {}
+            if cb_location in ["top", "bottom"]:
+                warnings.warn(
+                    "Colorbar orientation would be wrong in the current Matplotlib version (< 3.7.0)",
+                    stacklevel=2,
```
neutrinoceros, line 355: since we are within a private function, I think the correct stacklevel should be at least 3, but we should check that in the case this warning is triggered, it really points to the last line of user code (a manual check is enough for this).
Independently, we should also have a test for this warning, to see that it pops up in the expected conditions.

xshaokun: > but we should check that in the case this warning is triggered, it really points to the last line of user code (a manual check is enough for this).

In this case, I think it should be set to 6, like this.

```
/var/folders/kw/_83g0dzx6v968ryzq81mjm3h0000gn/T/ipykernel_9325/1071457781.py:12: UserWarning: Colorbar orientation would be wrong in the current Matplotlib version (< 3.7.0)
  fig = p.export_to_mpl_figure((2, 2), cbar_location="top", cbar_mode="edge", axes_pad=0)

```

> Independently, we should also have a test for this warning, to see that it pops up in the expected conditions.

So which file do you think would be more appropriate to add this test to? `test_image_compare_2D_plots.py` seems to be used only for image tests.

xshaokun:
> in fact we only run them on a specific runner image (`ubuntu-latest`) and against the latest version of matplotlib.

That makes it easier! In fact, I added `@pytest.mark.skipif` to avoid testing on platforms with a minimal version of matplotlib.

neutrinoceros:
>  In fact, I added @pytest.mark.skipif to avoid testing on platforms with a minimal version of matplotlib.

Doesn't hurt !

yt/visualization/base_plot_types.py
neutrinoceros: I would have recommended putting the test next to other tests exercising `export_to_mpl_figure` ... but then discovered there are currently none ! so, even though this is indeed slightly off-topic for `test_image_compare_2D_plots.py`, it's still the most natural place to put the new test I think.

xshaokun:
@neutrinoceros Passing an empty keyword dict `cb_kwargs = {}` fail the type-checking in [cf4e14d](https://github.com/yt-project/yt/pull/4981/commits/cf4e14ddc272848d5f517a6b83a9f8f7fab1e23a), but passed the test in [c47732d](https://github.com/yt-project/yt/pull/4981/commits/c47732d09918a46bcf9e7dfa2a8e315c660bafd8) before. Would you happen to know why?

Since for now, the workflow for tests requires approval, I abandoned `cb_kwargs ` in the latest commit to make the tests pass locally.

yt/visualization/base_plot_types.py
```diff
@@ -341,7 +341,19 @@ def _set_axes(self) -> None:
...
+        cb_location = getattr(self.cax, "orientation", None)
+        if matplotlib.__version_info__ >= (3, 7):
+            self.cb = self.figure.colorbar(self.image, self.cax, location=cb_location)
+        else:
+            if cb_location in ["top", "bottom"]:
+                warnings.warn(
+                    "Colorbar orientation would be wrong in the current Matplotlib version (< 3.7.0)",
+                    stacklevel=6,
+                )
+                self.cb = self.figure.colorbar(self.image, self.cax)
```
neutrinoceros, line 356:
```suggestion
            self.cb = self.figure.colorbar(self.image, self.cax)
```

neutrinoceros (changes requested):

xshaokun:
I have created the PR there and I thought something should also be done here.

neutrinoceros:
Sorry I missed that. For future reference, this is https://github.com/yt-project/yt_pytest_mpl_baseline/pull/6/
And now that I see the images, I also see a problem: the font family used in colorbar labels doesn't seem to be properly set properly when the colorbar is horizontal. We need to figure this out and update your companion PR accordingly when we have a fix.

xshaokun:
The current output is as follows, there are still some formatting issues, but it might be beyond this PR.

test_top

neutrinoceros:
Great, the output looks good to me. I'll merge your companion PR on the submodule repo when it's up to date ! thank you for your patience !

yt/visualization/base_plot_types.py
```diff
@@ -528,7 +540,9 @@ def _get_labels(self):
         labels = super()._get_labels()
         cbax = self.cb.ax
         labels += cbax.yaxis.get_ticklabels()
+        labels += cbax.xaxis.get_ticklabels()
         labels += [cbax.yaxis.label, cbax.yaxis.get_offset_text()]
+        labels += [cbax.xaxis.label, cbax.xaxis.get_offset_text()]
```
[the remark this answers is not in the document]
xshaokun: That makes sense why only the y-axis was considered here before!

yt/visualization/base_plot_types.py
```diff
@@ -341,7 +341,19 @@ def _set_axes(self) -> None:
         self.image.axes.set_facecolor(self.colorbar_handler.background_color)
 
         self.cax.tick_params(which="both", direction="in")
-        self.cb = self.figure.colorbar(self.image, self.cax)
+
+        # For creating a multipanel plot by ImageGrid
+        # we may need the location keyword, which was introduced since Matplotlib 3.7.0
```
neutrinoceros, line 346:
```suggestion
        # we may need the location keyword, which requires Matplotlib >= 3.7.0
```

yt/visualization/base_plot_types.py
```diff
@@ -341,7 +341,19 @@ def _set_axes(self) -> None:
...
+        cb_location = getattr(self.cax, "orientation", None)
```
neutrinoceros, line 347: I'm not sure I get why the "orientation" attribute is directly used as a "location" value. What are the possible values for `self.cax.orientation` ?

yt/visualization/base_plot_types.py
```diff
@@ -341,7 +341,19 @@ def _set_axes(self) -> None:
...
+        if matplotlib.__version_info__ >= (3, 7):
+            self.cb = self.figure.colorbar(self.image, self.cax, location=cb_location)
+        else:
+            if cb_location in ["top", "bottom"]:
+                warnings.warn(
+                    "Colorbar orientation would be wrong in the current Matplotlib version (< 3.7.0)",
```
neutrinoceros, line 353:
```suggestion
                    "Cannot properly set the orientation of colorbar. "
                    "Consider upgrading matplotlib to version 3.7 or newer",
```

neutrinoceros (commented):
Thanks a lot for your patience, I think this is almost ready for merge but here's a last round of review

yt/visualization/base_plot_types.py
xshaokun: In MPL, `ImageGrid.cbar_axes.orientation` is somehow designed to store the value from `ImageGrid.cbar_location` (see the [code](https://github.com/matplotlib/matplotlib/blob/a254b687df97cda8c6affa37a1dfcf213f8e6c3a/lib/mpl_toolkits/axes_grid1/axes_grid.py#L414C17-L414C28)), so that the value of `self.cax.orientation` here is supposed to be `top`,`right` , etc.

neutrinoceros: I see. This is surprising but kinda out of our reach. Thanks !

neutrinoceros (approved):

PR #4983, opened 2024-09-10
DEP: move dev-only dependencies to requirements files

### LINKED ISSUE #4976: BLD: can we remove dev-only extra targets ?
We have a couple "dev-only" (but still user-visible) extra dependency target that don't really have a use case for end users (or downstream developers, to the best of my knowledge).
Up until now, they haven't really been an issue, but I'm finding that their existence makes working with [`uv`](https://docs.astral.sh/uv/) (my new preferred universal workflow) more cumbersome[^1], so I'm proposing to replace them:

- `yt[doc]` and `yt[typecheck]` can be replaced by `requirements.txt` files (`yt[typecheck]` would actually benefit from this as it could be auto-updated by dependabot)
- `yt[minimal]` is entirely redundant and can be replaced by `uv pip compile pyproject.toml resolution=lowest-direct` in CI

Any objection ?
(putting this in the 4.4.0, hoping it'll get attention by the time we want to cut that release)

[^1]: specifically, it makes `uv sync` kinda-of useless

neutrinoceros:
## PR Summary

Close #4976

TODO:
- [x] adapt doc builds scripts on Jenkins


chrishavlin:
Should update the instructions on building the docs too. This line:

https://github.com/yt-project/yt/blob/c9b12af8d3315de72c575a717ac3b14daae8f6e4/doc/source/developing/building_the_docs.rst?plain=1#L93

neutrinoceros:
Good catch. Done !

chrishavlin:
> Good catch. Done ! 

oh, that did make the docs build fail though...

Xarthisius:
@yt-fido test this please.

Xarthisius:
@yt-fido test this please.

neutrinoceros:
btw thanks @Xarthisius for your work on Jenkins, I'm glad this is working !

chrishavlin (approved):

Xarthisius (approved):

PR #4990, opened 2024-09-13
Fix to load the swiftsim snapshot by removing the unused "RuntimePars"

weiguangcui:
## PR Checklist

### DIFF
--- a/yt/frontends/swift/data_structures.py
+++ b/yt/frontends/swift/data_structures.py
@@ -2,11 +2,10 @@ import numpy as np
 
 from yt.data_objects.static_output import ParticleFile
 from yt.frontends.sph.data_structures import SPHDataset, SPHParticleIndex
+from yt.frontends.sph.fields import SPHFieldInfo
 from yt.funcs import only_on_root
 from yt.utilities.logger import ytLogger as mylog
 from yt.utilities.on_demand_imports import _h5py as h5py
-
-from .fields import SwiftFieldInfo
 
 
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Hi. In #4921 we tried to make sure this bit of code stayed backward compatible and improved forward compatibility. As far as I can tell what you're proposing here is to rip off the backward compatibility layer. Can you explain exactly what is not currently working with the main branch yet ?

neutrinoceros:
(admittedly #4921 is not released yet so I don't expect yt 4.3.1 to work, but the dev branch should)

weiguangcui:
> Hi. In #4921 we tried to make sure this bit of code stayed backward compatible and improved forward compatibility. As far as I can tell what you're proposing here is to rip off the backward compatibility layer. Can you explain exactly what is not currently working with the main branch yet ?

Hi, sorry for forgetting to reply to this question. Yes, the new swift snapshot doesn't include that ["RuntimePars"] in its snapshot anymore. So it is removed, as such the "periodic" parameter is reading from another place: `periodic = int(parameters["InitialConditions:periodic"])`

neutrinoceros:
@weiguangcui I still expect you shouldn't have any issue with the development branch. Have you tried installing it ?
```
python -m pip install git+https://github.com/yt-project/yt.git
```

If I'm right then your proposed patch is unecessary. And if something else still needs fixing we need to take care of backward compatibility too, which is currently missing from your approach.

Issue #4995, opened 2024-09-19
Particle IDs of Gadget HDF5-style data are converted from integers to floats


arkordt:
### Bug report

**Bug summary**

An arepo snapshot whose on-disk ParticleIDs data type is uint32 or uint64 is converted to float32 or float64, respectively.

The conversion can be traced back to two locations. The first conversion is in [yt/utilities/io_handler.py:209](https://github.com/yt-project/yt/blob/66ddd0eb154ce7fa5e1db9bc7f88f66c63fb6a45/yt/utilities/io_handler.py#L209), the second conversion occurs in [unyt/array.py:712](https://github.com/yt-project/unyt/blob/57587bd57cbbf50bb9eafd25acd2db74f4d15d04/unyt/array.py#L712) that is called from [yt/data_objects/selection_objects/data_selection_objects.py:217](https://github.com/yt-project/yt/blob/66ddd0eb154ce7fa5e1db9bc7f88f66c63fb6a45/yt/data_objects/selection_objects/data_selection_objects.py#L217). If both conversions are commented out, the expected output is produced. 

I also submitted an [issue to unyt](https://github.com/yt-project/unyt/issues/517) because I am not sure where it would be better to avoid the type conversion.

**Code for reproduction**

`yt-test-data` should be the path where the YT test data is stored.

```python
>>> import yt
>>> ds = yt.load('yt-test-data/ArepoBullet/snapshot_150.hdf5')
>>> ds.all_data()['PartType0', 'ParticleIDs']
```

**Actual outcome**

```
unyt_array([1710352., 1710346., 1710350., ..., 5721056., 5843308., 5843259.],
      dtype=float32, units='(dimensionless)')
```

**Expected outcome**

```
unyt_array([1710352, 1710346, 1710350, ..., 5721056, 5843308, 5843259],
      dtype=uint32, units='(dimensionless)')
```

**Version Information**
  * Operating System: macOS
  * Python Version: 3.12.6
  * yt version: 4.3.1 or 4.4.dev0 (current main branch)
  * Other Libraries (if applicable): unyt version: 3.0.2

yt was installed via conda (version 4.3.1) or from source (main branch) for testing and bug tracing purposes.

chrishavlin:
I commented over in the unyt issue, but I do think that this is more of a yt issue than a unyt issue. IMO it'd make sense to handle integer, dimensionless quantities (like `particleIDs`) with a special case that avoids the type and unit conversion that is applied to physical fields.

[...]

PR #4996, opened 2024-09-19
ENH: avoid particle_index type cast

chrishavlin:
This is a possible fix for #4995 which would affect all the particle frontends that don't override `_read_particle_selection`.


chrishavlin:
Converting to draft because this needs some more thorough testing: namely, does this introduce any bugs in cython operations that are expecting float arrays? Only tried it out with a projection and it works (because the int field gets converted to float before the projection), but need to check other functionality.

chrishavlin:
Assuming tests pass, I think this is ready for review. 


For reference, the following IO child classes override `_read_particle_selection` and always cast all fields to float64:

* `IOHandlerGadgetFOFHaloHDF5` 
* `IOHandlerChomboHDF5` (actual cast to float occurs in `._read_particles`)
* `IOHandlerOpenPMDHDF5`

The one other IO frontend that overrides `_read_particle_selection` is `IOHandlerOrion` but it does not initialize dtypes of arrays or cast to float64. Every other fronted inherits the base `_read_particle_selection` so would be affected by this PR. Based on my quick read, it'd be easy enough to enable the behavior of this PR in the 3 frontends above and I can do that here after initial review or in a followup PR or I can just open an issue for future reference.

chrishavlin:
oh, I suppose I should update the docs on creating a new frontend too to mention this behavior, but I'll wait for review before doing that in case the functionality changes.

neutrinoceros:
As I said somewhere else I'm currently unable to compile yt so I cannot review this properly at the moment. I hope someone else can step in.

yt/data_objects/selection_objects/data_selection_objects.py
```diff
@@ -214,7 +214,8 @@ def get_data(self, fields=None):
 
         for f, v in read_particles.items():
             self.field_data[f] = self.ds.arr(v, units=finfos[f].units)
-            self.field_data[f].convert_to_units(finfos[f].output_units)
+            if finfos[f].units != finfos[f].output_units:
+                self.field_data[f].convert_to_units(finfos[f].output_units)
```
chrishavlin, line 218: note to reviewers: this change is necessary because unit conversions on unyt arrays always cast the result to float64. e.g., `unyt.unyt_array([1, 2], "1", dtype='int').to("1")` yields an array of type `float64`. So only converting if the units differ allow the int fields to persist through the data read.

matthewturk (dismissed):
This seems good to me as a first pass.

I want to note one other possible place we may want to update things.  I think it might be too much work, but could be simplified with a broader particle type accessing refactor.  (i.e., making it easier to figure out what type to make new fields, etc.)

I can see use cases for making the smoothing and deposition operations utilize integers.  Specifically, if one wanted to deposit the nearest particle's index, that would be a use case we could support.  But, it's not necessary for this.

I would also like to suggest that we avoid (for now) allowing 32bit floats and ints, which you've also avoided here.

matthewturk:
One thing that came to mind, in addition -- using fused types we can make it somewhat easier to address issues in the cython code, but we could also utilize JIT's for fun in the distant future.

yt/utilities/io_handler.py
```diff
@@ -60,6 +60,22 @@ def push(self, grid, field, data):
...
+        # field (or its alias if it exists), which is set as int64. Important
+        # to note that the data type will only be preserved for direct reads
+        # and any operations will either implicity (via unyt) or explicitly
+        # convert to float64.
+        dtypes: defaultdict[FieldKey, type] = defaultdict(lambda: np.float64)
+        for ptype in self.ds.particle_types:
+            p_index = (ptype, "particle_index")
+            if p_index in self.ds.field_info.field_aliases:
+                p_index = self.ds.field_info.field_aliases[p_index]
+            dtypes[p_index] = np.int64
```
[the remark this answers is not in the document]
chrishavlin: a string should be totally fine actually -- anything that numpy can accept for a dtype kwarg should be OK. also totally happy to use a string here, I didn't give much thought to using a string vs a dtype object. You have a preference?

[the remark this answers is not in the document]
matthewturk: I'm good with keeping it.

chrishavlin:
> I can see use cases for making the smoothing and deposition operations utilize integers. Specifically, if one wanted to deposit the nearest particle's index, that would be a use case we could support. But, it's not necessary for this.

I agree! and I think this PR opens up the possibility of doing just that. 

> I would also like to suggest that we avoid (for now) allowing 32bit floats and ints, which you've also avoided here.

agreed! I'll add a comment to the new `_particle_dtypes` attribute to reinforce that in case any frontends in the future end up overwriting the attribute.

yt/utilities/io_handler.py
```diff
@@ -60,6 +60,22 @@ def push(self, grid, field, data):
...
         self.queue[grid][field] = data
 
+    @property
+    def _particle_dtypes(self) -> defaultdict[FieldKey, type]:
+        # returns a defaultdict of data types for particle fields.
+        # defaults for all fields are float64, except for the particle_index
+        # field (or its alias if it exists), which is set as int64. Important
+        # to note that the data type will only be preserved for direct reads
+        # and any operations will either implicity (via unyt) or explicitly
+        # convert to float64.
```
chrishavlin, line 70:
```suggestion
        # convert to float64. Furthermore, float32 and int32 are not supported. 
```

chrishavlin:
pre-commit.ci autofix

jzuhone:
@chrishavlin what needs to get this across the finish line?

chrishavlin:
hey @jzuhone -- it's just in need of some reviews.

chrishavlin:
@yt-fido test this please

yt/utilities/io_handler.py
```diff
@@ -60,6 +60,22 @@ def push(self, grid, field, data):
             raise ValueError
         self.queue[grid][field] = data
 
+    @property
```
jzuhone, line 63: My only comment here is that I'd love it if it was more general (as this relates to the use case that @chunyen-chen was talking about). But if that is not easy to do in this PR we can just submit another one.

chrishavlin: well my thinking was this `_particle_dtypes` dictionary would be a general, front-end dependent dictionary mapping types to fields. Any ideas on how to make it more general/useable? Looking at it now after some months, I do think it could make sense to instead have this field-dtype mapping defined over in the `FieldInfoContainer`, but that would still require each frontend to over-ride or add to it as needed.

chrishavlin:
realized i re-triggered the wrong test... decided to just merge with main since it's been so long.

chrishavlin:
that type-checking failure looks unrelated, will open an issue for it.

jzuhone (approved):

chrishavlin:
I think if we don't hear any objections in the next few days, we should merge this in... @matthewturk do you want to take another look or should we consider your initial comment ([link](https://github.com/yt-project/yt/pull/4996#pullrequestreview-2326119942)) as an approval?

PR #4997, opened 2024-09-20
Autodetect rockstar format

cphyc:
## PR Summary

Rockstar catalogs unfortunately are not self-descriptive. This PR is an attempt to automatically detect the format into which a catalog is written by simply trying all the formats until one is found that looks alright.

### DIFF
--- a/yt/frontends/rockstar/data_structures.py
+++ b/yt/frontends/rockstar/data_structures.py
@@ -1,7 +1,7 @@
 import glob
 import os
 from functools import cached_property
-from typing import List, Optional
+from typing import Any, List, Optional
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


brittonsmith:
@cphyc, is it too late to merge this? Maybe you just need to merge the PR branch with upstream changes.

cphyc:
Hey @mtryan83,

So long as your alternative format has a different number of bytes in the header _or_ in the members of the halos, what you wrote above should work (assuming `m_hires` and `m_lowres` are indeed located at the end of the halo definition).

If this works, would you mind opening a PR (on top of this one)?

As a side note, it would make the code much more readable if the syntax was:
```python
halo_dt: list[HaloDataType] = [
    ("particle_identifier", np.int64),
    ("particle_position_x", np.float32),
    ("particle_position_y", np.float32),
    ("particle_position_z", np.float32),
    ("particle_mposition_x", np.float32, {0}),
    ("particle_mposition_y", np.float32, {0}),
    ("particle_mposition_z", np.float32, {0}),
    [...]
    ("m_pe_d", np.float32, {1, 2}),
    [...]
]
```
At the moment, the third (optional) argument gives a lower and upper boundary for the format revisions that should include the corresponding entry; using a set would be more explicit IMO.

brittonsmith (approved):

brittonsmith (dismissed):

brittonsmith (dismissed):

PR #5008, opened 2024-09-25
TYP: hotfix unsolvable type-checking requirements

neutrinoceros:
## PR Summary
Hotfix a problem I just created by merging #4989 (which apparently didn't run typechecking at all)
[example logs](https://github.com/yt-project/yt/actions/runs/11036617135/job/30655473559)

### DIFF
--- a/.github/workflows/type-checking.yaml
+++ b/.github/workflows/type-checking.yaml
@@ -7,7 +7,8 @@ on:
   pull_request:
     paths:
       - yt/**/*.py
-      - setup.cfg
+      - pyproject.toml
+      - requirements/typecheck.txt
       - .github/workflows/type-checking.yaml
   workflow_dispatch:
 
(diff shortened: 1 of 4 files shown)


chrishavlin:
that windows failure is from the latest h5py release, see https://github.com/h5py/h5py/issues/2505 I'll adjust the test requirements to avoid it (and push up to this PR)

chrishavlin:
oh, err. maybe that wasn't the right move. cause now h5py will be installed in every test environment, including the minimal test which it shouldn't be... I'm going to let tests run then move the h5py exclusion up to the hdf5 install option.

Xarthisius (approved):

neutrinoceros:
thanks Chris and Kacper for finishing this ! FYI h5py 3.12.0 is now yanked from PyPI and I'll be looking into what happened with that release.

PR #3795, opened 2022-02-08
BUG: fix a bug in pixelize_cylinder where buffer shape was read backwards

neutrinoceros:
## PR Summary
Welp, that was easy.
Fix #3794

Just to make it flagrant I'll make an even more drastic version of the script therein
```python
import numpy as np
import yt

shape = (32, 16, 1)
a = np.linspace(1, 100, 16)
b = np.ones((32, 16))
c = np.reshape(a * b, shape)
data = {("gas", "density"): c}

ds = yt.load_uniform_grid(
    data,
    shape,
    bbox=np.array([[0.0, 5.0], [np.pi / 2 - 0.2, np.pi / 2 + 0.2], [-0.1, +0.1]]),
    geometry="spherical",
)

p = yt.SlicePlot(ds, "phi", "density")
p.save("/tmp/buff_default.png")
p.set_buff_size((10, 800))
p.save("/tmp/buff_10_800.png")

p.set_buff_size((10, 10))
p.save("/tmp/buff_10_10.png")
```

buff_default
buff_10_800
buff_10_10

PR #5018, opened 2024-10-04
PERF: release the GIL in cylindrical pixelizer

chrishavlin:
release the gil!

This surprisingly seemed to work with minimal effort... took the pixelization time from 53s to 27s for one example (see [this slack thread](https://yt-project.slack.com/archives/C1YCPSX08/p1727899750601519)).

### DIFF
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -566,6 +566,7 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
     cdef np.float64_t r_inc, theta_inc
     cdef np.float64_t costheta, sintheta
     cdef int i, i1, pi, pj
+    cdef np.float64_t npPI = np.pi
 
     cdef int imin, imax
     imin = np.asarray(radius).argmin()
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Releasing the GIL is likely a good idea here, but I also suspect that the abysmal performance in this routine may be fixed independently (possibly by reverting an earlier patch from mine that I'll dig up for in my future review). Any way, I'll try to review in details in the coming days. Thanks !

chrishavlin:
if you figure out a separate fix I'm happy to go with it and close this one!

chrishavlin:
oh, especially since image tests seem to be failing and it was not as simple as i hoped :)

chrishavlin:
ok, i think i see the issue but won't have time to fix it for a while... so will hold off in case you devise a separate performance fix.

neutrinoceros:
oh wow these diffs are catastrophic. I'll try to tackle the performance issue from another angle and I'll move this to draft if you don't mind.

matthewturk:
These lines:

```
                     ptbounds[0] = ptbounds[0] % (2*np.pi)
                     ptbounds[1] = ptbounds[1] % (2*np.pi)
```

are using Python math operations inside a tight loop.  What if we used a constant for PI instead of `np.pi`, which I believe will do a python-level lookup?

cphyc:
> are using Python math operations inside a tight loop. What if we used a constant for PI instead of `np.pi`, which I believe will do a python-level lookup?

With the latest version, they are actually expanded to
```C
(__pyx_v_ptbounds[0]) = fmod((__pyx_v_ptbounds[0]), (2.0 * __pyx_v_npPI));
```
so there might be a small speedup to be found by precomputing 2π ahead of time, but I'm not sure this is super important.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -611,65 +612,66 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
...
-        theta_i = theta0 - dtheta_i
-        theta_inc = r_inc / (r0 + dr_i)
-
-        while theta_i < theta0 + dtheta_i:
-            r_i = r0 - dr_i
-            costheta = math.cos(theta_i)
-            sintheta = math.sin(theta_i)
-            while r_i < r0 + dr_i:
-                if rmax <= r_i:
+    with nogil:
```
cphyc, line 615: Is this a case where we'd want to use
```python
for i in prange(radius.shape[0], nogil=True):
   pass
```

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -566,6 +566,7 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
     cdef np.float64_t r_inc, theta_inc
     cdef np.float64_t costheta, sintheta
     cdef int i, i1, pi, pj
+    cdef np.float64_t npPI = np.pi
```
cphyc, line 569:
```suggestion
    cdef np.float64_t twoPI = 2 * np.pi
```

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -611,65 +612,66 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
...
+                        corners[2] = math.atan2(xp+dx, yp)
+                        corners[3] = math.atan2(xp+dx, yp+dy)
+                        ptbounds[0] = ptbounds[1] = corners[3]
+                        for i1 in range(3):
+                            ptbounds[0] = fmin(ptbounds[0], corners[i1])
+                            ptbounds[1] = fmax(ptbounds[1], corners[i1])
+
+                        # shift to a [0, PI] interval
+                        ptbounds[0] = ptbounds[0] % (2*npPI)
+                        ptbounds[1] = ptbounds[1] % (2*npPI)
```
cphyc, line 667:
```suggestion
                        ptbounds[0] = ptbounds[0] % twoPI
                        ptbounds[1] = ptbounds[1] % twoPI
```

cphyc (approved):

matthewturk:
> > are using Python math operations inside a tight loop. What if we used a constant for PI instead of `np.pi`, which I believe will do a python-level lookup?
> 
> With the latest version, they are actually expanded to
> 
> ```c
> (__pyx_v_ptbounds[0]) = fmod((__pyx_v_ptbounds[0]), (2.0 * __pyx_v_npPI));
> ```
> 
> so there might be a small speedup to be found by precomputing 2π ahead of time, but I'm not sure this is super important.

Huh.  I guess I wasn't on the latest version!  For me it did a ton of unpacking and python type coercion.  I guess that's one of the advantages of upgrading, eh?

Are our performance tests being run on the latest version?  (i.e., do you see the slowness, or is what I am referring to showing up in our benchmarks)

yt/utilities/lib/pixelization_routines.pyx
chrishavlin: sure! I think prange should work here. but i'll wait until the rest is actually working....

chrishavlin:
still expect the image diffs to fail -- i'll look later today and if it's a quick fix (which i think it is) I'll put it in and do some performance tests.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -566,7 +566,8 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
     cdef np.float64_t r_inc, theta_inc
     cdef np.float64_t costheta, sintheta
     cdef int i, i1, pi, pj
-    cdef np.float64_t twoPI = 2 * np.pi
+    cdef np.float64_t onePI = np.pi
```
neutrinoceros, line 569: here's the direct way to "import" the constant from numpy as the C level
```
cdef extern from "numpy/npy_math.h":
    double NPY_PI
```
(taken from https://github.com/yt-project/yt/pull/4948/files#r1692841029)

neutrinoceros: (NPY_2_PI is also available)

chrishavlin: hah, will import `NPY_2_PI` as well in a minute after tests run this round.... i think this one will pass...

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -611,65 +613,70 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
...
+                            ptbounds[0] = fmin(ptbounds[0], corners[i1])
+                            ptbounds[1] = fmax(ptbounds[1], corners[i1])
+
+                        # shift to a [0, PI] interval
+                        ptbounds[0] = ptbounds[0] % twoPI
+                        if ptbounds[0] < 0:
+                            ptbounds[0] += NPY_PI
+                        ptbounds[1] = ptbounds[1] % twoPI
+                        if ptbounds[1] < 0:
+                            ptbounds[1] += NPY_PI
```
chrishavlin, line 672: so I'm not sure why I needed to add the check for negative numbers here, but  after switching from `ptbounds[1] % (2*np.pi)` to `ptbounds[1] % twoPI`, the result returns numbers in the range (-pi,pi).

yt/utilities/lib/pixelization_routines.pyx
chrishavlin: oh wait, NPY_2_PI is 2/pi, not 2*pi ?

yt/utilities/lib/pixelization_routines.pyx
chrishavlin: ok well using `prange` might be a bit more work than I want to put in right now -- lots of compilation errors along the lines of :

```python-traceback
      yt/utilities/lib/pixelization_routines.pyx:635:20: Cannot read reduction variable in loop body
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
                  while r_i < r0 + dr_i:
                      if rmax <= r_i:
                          r_i += r_inc
                          continue
                      y = r_i * costheta
                      x = r_i * sintheta
                          ^
      ------------------------------------------------------------
      
      yt/utilities/lib/pixelization_routines.pyx:636:20: Cannot read reduction variable in loop body
```

yt/utilities/lib/pixelization_routines.pyx
neutrinoceros: aaaah, missed that, sorry

matthewturk:
I like the comparison.  I immediately thought, "what about in AMR" then realized this is likely a better general purpose diagnostic.

yt/utilities/lib/pixelization_routines.pyx
neutrinoceros: I think we need to understand this bit, and possibly come up with a way to perform the computation that doesn't involve branching; hard-wiring a somewhat expensive calculation generally hurts performance less than ruining branch prediction.
I'm writing this comment early in my review but I'll try to come up with more insight.

neutrinoceros: can you try this patch
```patch
index 194ce7af6..6728375d0 100644
@@ -664,12 +664,8 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
                             ptbounds[1] = fmax(ptbounds[1], corners[i1])
 
                         # shift to a [0, PI] interval
-                        ptbounds[0] = ptbounds[0] % twoPI
-                        if ptbounds[0] < 0:
-                            ptbounds[0] += NPY_PI
-                        ptbounds[1] = ptbounds[1] % twoPI
-                        if ptbounds[1] < 0:
-                            ptbounds[1] += NPY_PI
+                        ptbounds[0] = (ptbounds[0]+twoPI) % twoPI
+                        ptbounds[1] = (ptbounds[1]+twoPI) % twoPI
 
                         if prbounds[0] >= rmin and prbounds[1] <= rmax and \
                            ptbounds[0] >= tmin and ptbounds[1] <= tmax:
```
the result should be the same in all cases, but it avoids branching, so I expect it to be faster.
If this works, we'll want to add a comment for why it's done this way.

neutrinoceros:
I looked for most recent patches to this routine from ~3 years ago when I was focused on correctness and possibly neglecting performance. Here's what I found, in case that's relevant here
- #3782
- #3795
- #3818

#3782 and #3818 may have impacted performance, so it might be worth trying to revert them and see if the performance changes measurably.
If #3818 is costly, it may be argued that it should be reverted, as it really comes down to an arbitrary choice between two equally valid approaches, with converging results in the limit of infinite resolution.
#3782 seems actually more likely to have a negative impact, but it's a lot less trivial to decide what to do if so. In any case I think it's worth investigating now, and if performance reaches an acceptable level with this PR, maybe we don't need to do anything more. I'd just like it to be an informed decision.

neutrinoceros:
given that I've post poned the release to ~10 days for now there should be enough time to squeeze this in, though it's not a blocker.

yt/utilities/lib/pixelization_routines.pyx
matthewturk: I am wondering if we could use `fmod` here.

chrishavlin: looks like in cython, `%` is an alias to `fmod` already when using C functions, which returns a negative number when the first argument is negative (which is why the tests started failing here I think).  but I do think it's clearer to use `fmod` to avoid the confusion in the first place (so `fmod(ptbounds[0]+twoPI, twoPI)` to handle the negative numbers). In case it's helpful, I wrote up some scratch code to double check, here's a plot:

comparison-cython-modulo

neutrinoceros: What is this plot showing ? I'm not seeing labels on those axes

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -611,65 +613,70 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
...
+                        corners[1] = math.atan2(xp, yp+dy)
+                        corners[2] = math.atan2(xp+dx, yp)
+                        corners[3] = math.atan2(xp+dx, yp+dy)
+                        ptbounds[0] = ptbounds[1] = corners[3]
+                        for i1 in range(3):
+                            ptbounds[0] = fmin(ptbounds[0], corners[i1])
+                            ptbounds[1] = fmax(ptbounds[1], corners[i1])
+
+                        # shift to a [0, 2*PI] interval
+                        # note: in cython, % is an alias to fmod and the sign
```
neutrinoceros, line 667: This comment may still be somewhat misleading: it's only true with `@cython.cdivision(True)`
It's also a bit confusing since you're not using `%` anymore.
I say let's keep the comment about `fmod` only to resolve both issues.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -611,65 +613,70 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
...
+                        # of the returned value matches the sign of the first
+                        # argument, so need to offset by 2pi to ensure we
+                        # always get a positive result in [0,2pi]
```
chrishavlin, line 670:
```suggestion
                        # note: with fmod, the sign of the returned value 
                        # matches the sign of the first argument, so need 
                        # to offset by 2pi to ensure a positive result in [0, 2pi]
```

neutrinoceros:
Here's the revert PR https://github.com/yt-project/yt/pull/5034

neutrinoceros:
Alright, to summarize the scattered discussion: I still think there should be a smarter way to restore optimial performances in this routine (which is what I'm trying to do in #5035) *but* so far I haven't been able to find a completely backward compatible solution. This patch *is* 100% backward compatible so I'm choosing to include it in yt 4.4.0, and I'm hoping we can revisit soon after.
Thanks @chrishavlin, and sorry it's taken me so long to settle on your solution !

neutrinoceros (approved):

PR #5028, opened 2024-10-22
Stop using deprecated `urllib.request.FancyURLopener`

### LINKED ISSUE #5027: `urllib.request.URLopener` is deprecated
### Bug report

yt uses [`urllib.request.URLopener`](https://docs.python.org/3/library/urllib.request.html#urllib.request.URLopener), which was deprecated in Python 3.3. Would you accept a PR that uses [`urlretrieve()`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve) instead? Ta!

barneygale:
## PR Summary

Replace use of `urllib.request.FancyURLopener` with `urlretrieve()`. The former has been deprecated since Python 3.3.

Fixes #5027 

## PR Checklist

### DIFF
--- a/yt/funcs.py
+++ b/yt/funcs.py
@@ -618,14 +618,15 @@ def fancy_download_file(url, filename, requests=None):
 
 def simple_download_file(url, filename):
     import urllib.request
+    import urllib.error
 
-    class MyURLopener(urllib.request.FancyURLopener):
-        def http_error_default(self, url, fp, errcode, errmsg, headers):
-            raise RuntimeError(
-                f"Attempt to download file from {url} failed with error {errcode}: {errmsg}."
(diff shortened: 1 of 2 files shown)


barneygale:
Let me know if you'd like tests!

neutrinoceros:
Thanks, indeed it'd be nice t have at least one test for this function: the only way we didn't catch the deprecation warning earlier is that we have none. On the other hand, adding new tests can also prove cumbersome because we're still not finished migrating from nosetests to pytest. We can probably just go with manual testing actually... I'll give it a try tomorrow !

neutrinoceros (approved):
LGTM. Thanks a lot @barneygale !

PR #5033, opened 2024-10-29
fix docs for annotate_contour

zingale:
clim is (lower, upper), not (upper, lower)

## PR Summary

in `plot_modifications.py` `ContourCallback`, we do the contours as:

```
levels = np.linspace(clim[0], clim[1], self.levels) 
```

so `clim[0]` is the lower limit, not upper.

## PR Checklist


neutrinoceros (approved):
nice catch. No idea how we ended up here.

zingale:
I figured it was some cosmology thing ;)

PR #5034, opened 2024-10-29
PERF: restore cylindrical pixelizer performance from yt 4.0.2 (revert gh-3818)

neutrinoceros:
## PR Summary
Reverts #3818 after identifying it as a significant performance regression (see exploration in https://github.com/yt-project/yt/pull/5018#issuecomment-2445037555)

This may complement or replace #5018, but I currently think it should be considered for merging first.

update: diffs are expected, here's the companion PR: https://github.com/yt-project/yt_pytest_mpl_baseline/pull/7

### DIFF
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -561,22 +561,14 @@ def pixelize_cylinder(np.float64_t[:,:] buff,
 ):
 
     cdef np.float64_t x, y, dx, dy, r0, theta0
-    cdef np.float64_t rmin, rmax, tmin, tmax, x0, y0, x1, y1, xp, yp
+    cdef np.float64_t rmax, x0, y0, x1, y1
     cdef np.float64_t r_i, theta_i, dr_i, dtheta_i
     cdef np.float64_t r_inc, theta_inc
     cdef np.float64_t costheta, sintheta
-    cdef int i, i1, pi, pj
+    cdef int i, pi, pj
 
-    cdef int imin, imax
-    imin = np.asarray(radius).argmin()
-    imax = np.asarray(radius).argmax()
-    rmin = radius[imin] - dradius[imin]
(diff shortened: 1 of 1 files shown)


matthewturk:
So I think we should revert this, *but*, I'm trying to convince myself that we can't make this change also happen just for situations where the pixels are at an extrema in one of the dimensions.  Like, do we have to do this for all of them?  It seems like we could split the difference by computing the full "does it go beyond" only in cases where, say, we're toward the end of the `theta` or `r` iteration?  Maybe that's not possible.

It also seems to me that because we're iterating on the inside around theta and on the outside around r, we could potentially use a break instead.  Or move a `continue` somewhere with early termination.

Anyway, we don't really necessarily need to keep this behavior, like you said -- in most if not all cases the pixels will be very small compared to the cell values, so it shouldn't be that big a deal.

Actually, hm.  In theory, this takes a long time because we're doing it for every cell value that gets deposited into a pixel, *for every pixel*.  I'll think about it, because I wonder if it's possible to take all the extrema values of $(r,\theta)$ pairs and compute those, and then apply that to `mask` or `buff` afterward.

neutrinoceros:
>  I wonder if it's possible to take all the extrema values of ( r , θ ) pairs and compute those, and then apply that to mask or buff afterward.

That would be ideal, I suppose.

matthewturk:
> > I wonder if it's possible to take all the extrema values of ( r , θ ) pairs and compute those, and then apply that to mask or buff afterward.
> 
> That would be ideal, I suppose.

I'd propose we include this, but also include a comment pointing a possible way forward -- for those cases where sub-pixel boundaries are critical, in case someone wants to do it.

neutrinoceros:
Actually let me try to do it now.

chrishavlin:
for what it's worth I like the direction this is headed! once it's ready for review again I can try out the gil release again to see how much of an extra boost we can get.

neutrinoceros:
closing in favor of #5035

PR #5035, opened 2024-10-30
PERF: optimize cylindrical pixelizer routine

neutrinoceros:
## PR Summary
Yet another alternative to #5018 and #5034
This separates filtering of pixels that only partially intersect with the domain and aims to do a minimal amount of work there. On my system I'm getting similar performances as compared with #5034, but the goal here is to *not* change the observable behavior. Let's see how this does in CI 🤞🏻


neutrinoceros:
I realized that there's still a defect with my approach: I'm only shaving the pixilization from the outside in, missing most (if not all) pixels on the *inside* of the radial domain. I think I can figure this out too but it'll take more time than I have left today, so it'll wait for another day. I also don't think we have any image test that would reveal this bug as we use `rmin=0` everywhere. I think we should start by adding a test case for this.

I also see a segfault on windows, which is *probably* not actually platform specific, and needs to be addressed in any case.

neutrinoceros:
I fixed the segfault, which was actually trivial to locate by re-enabling bound checking for debugging.

yt/utilities/lib/pixelization_routines.pyx
```diff
@@ -546,6 +546,63 @@ def pixelize_off_axis_cartesian(
...
+    np.float64_t img_y0,
+    np.float64_t img_dx,
+    np.float64_t img_dy,
+    int pixel_i,
+    int pixel_j,
+) -> bool:
+    cdef np.float64_t xp, yp
+    cdef np.float64_t prbounds[2]
+    cdef np.float64_t ptbounds[2]
+    cdef np.float64_t corners[8]
```
neutrinoceros, line 567:
```suggestion
    cdef np.float64_t corners[4]
```

neutrinoceros:
Adding the missing tests in #5036

neutrinoceros:
Hum, I haven't been able to figure out a trick to fix "pixel shaving" on the inner side without compromising performance yet, so I'm starting to think it'd be preferable to go with #5018 first and revisit this after the 4.4.0 release. Does this suite you, @chrishavlin ?

chrishavlin:
Ya, that's fine by me! We can also hold #5018 to after 4.4.0 if that would make the effort in this PR easier.

PR #5036, opened 2024-10-31
TST: add image tests for cylindrical pixelization of a box with rmin>0

neutrinoceros:
## PR Summary

In support of #5035 after identifying such cases as blind spots in the test suite.

update: [companion PR](https://github.com/yt-project/yt_pytest_mpl_baseline/pull/8)

### DIFF
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -431,7 +431,14 @@ _fake_amr_ds_default_units = ("g/cm**3",)
 
 
 def fake_amr_ds(
-    fields=None, units=None, geometry="cartesian", particles=0, length_unit=None
+    fields=None,
+    units=None,
+    geometry="cartesian",
+    particles=0,
+    length_unit=None,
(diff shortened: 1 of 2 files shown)


yt/testing.py
```diff
@@ -431,7 +431,14 @@ def fake_random_ds(
...
 def fake_amr_ds(
-    fields=None, units=None, geometry="cartesian", particles=0, length_unit=None
+    fields=None,
+    units=None,
+    geometry="cartesian",
+    particles=0,
+    length_unit=None,
+    *,
+    domain_left_edge=None,
+    domain_right_edge=None,
```
neutrinoceros, line 441: I needed to expose these parameters because we didn't have a fake_ds function that exposed both domain limits and geometry, which explains why the cases I'm interested in here are not covered yet.

chrishavlin (approved):

Issue #5041, opened 2024-11-05
TST: failing to load `pyart` 


neutrinoceros:
The test pipeline on Jenkins appears to be broken as an `ImportError` is raised when trying to lazy-load pyart

https://tests.yt-project.org/job/yt_py310_git/8571/

I have reported the problem upstream https://github.com/ARM-DOE/pyart/issues/1680
In the meantime we can add `datatree` as an optional requirement on our side as a temporary measure.

kmuehlbauer:
@neutrinoceros Heads-up, datatree was merged into [xarray codebase](https://github.com/pydata/xarray/blob/main/DATATREE_MIGRATION_GUIDE.md) recently.

To use xarray-datatree (as in the test) you would also need to pin xarray to <=2024.9.0 and xradar <=to 0.7.0.

xradar 0.7.0 has a pin for xarray <= 2024.9.0
xradar 0.8.0 has a pin for xarray >= 2024.10.0

HTH

neutrinoceros:
marking this as a blocker since we cannot merge anything until there's at least a workaround for this issue

neutrinoceros:
fixed in #5042

PR #5042, opened 2024-11-05
pin xradar<0.8.0 in tests

chrishavlin:
This is a temporary fix for #5041 on yt's side that should get tests passing but will need to be updated when a fix on the the arm-pyart side is released (at which point we could pin a min version for arm-pyart for the tests).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -200,6 +200,7 @@ test = [
     "pytest-mpl>=0.16.1",
     "sympy!=1.10,!=1.9", # see https://github.com/sympy/sympy/issues/22241
     "imageio!=2.35.0", # see https://github.com/yt-project/yt/issues/4966
+    "xradar<0.8.0",  # see https://github.com/yt-project/yt/issues/5041
 ]
 
 [project.scripts]


chrishavlin:
just to summarize the confusing dependency resolution (and why I just pinnned xradar here):

* as of this writing arm-pyart (v 1.16.1) depends explicitly on xarray, xradar and implicitly on xarray-datatree (which installs as `datatree`)
* xarray-datatree is now deprecated and has been moved to xarray >= 2024.10.0
* xradar > 0.8.0 requires xarray >= 2024.10.0 to account for the xarray-datatree deprecation

So effectively arm-pyart currently has both a hard dependency on xradar < 0.8.0 and xarray-datatree and just pinning xradar < 0.8.0 is sufficient for yt's tests. 

Moving forward, the simplest thing to do will be to replace the change in this PR with a test requirement on `arm-pyart >= 1.16.2` after it is released (and assuming it contains a fix).

neutrinoceros:
Thanks a *lot* for the write up, I was really confused by this situation.

neutrinoceros (approved):

PR #5043, opened 2024-11-05
cf_radial: fix deprecated dims access, add pytest tests with mock ds

chrishavlin:
While working on #5042 I noticed a deprecation warning from cf_radial, e.g., from [this run](https://tests.yt-project.org/job/yt_py310_git/8572/testReport/yt.frontends.cf_radial.tests/test_outputs/test_cf_radial_gridded/):

```
/home/fido/workspace/yt_py310_git/yt/frontends/cf_radial/data_structures.py:281: FutureWarning: The return type of
`Dataset.dims` will be changed to return a set of dimension names in future, in order to be more consistent
with`DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.
 
dims = [xr_ds_handle.dims[d] for d in "xyz"]
```

This PR fixes that warning, adds some pytest-only tests (using a mocked dataset), and fixes a subsequent deprecation warning that was revealed by the new pytest.

### DIFF
--- a/yt/frontends/cf_radial/data_structures.py
+++ b/yt/frontends/cf_radial/data_structures.py
@@ -272,15 +272,19 @@ class CFRadialDataset(Dataset):
         with self._handle() as xr_ds_handle:
             x, y, z = (xr_ds_handle.coords[d] for d in "xyz")
 
-            self.origin_latitude = xr_ds_handle.origin_latitude[0]
-            self.origin_longitude = xr_ds_handle.origin_longitude[0]
-
             self.domain_left_edge = np.array([x.min(), y.min(), z.min()])
             self.domain_right_edge = np.array([x.max(), y.max(), z.max()])
             self.dimensionality = 3
             dims = [xr_ds_handle.sizes[d] for d in "xyz"]
             self.domain_dimensions = np.array(dims, dtype="int64")
             self._periodicity = (False, False, False)
+
+            # note: origin_latitude and origin_longitude arrays will have time
+            # as a dimension and the initial implementation here only handles
+            # the first index. Also, the time array may have a datetime dtype,
+            # so cast to float.
+            self.origin_latitude = xr_ds_handle.origin_latitude[0]
+            self.origin_longitude = xr_ds_handle.origin_longitude[0]
             self.current_time = float(xr_ds_handle.time.values[0])
 
         # Cosmological information set to zero (not in space).
--- a/yt/frontends/cf_radial/data_structures.py
+++ b/yt/frontends/cf_radial/data_structures.py
@@ -278,10 +278,10 @@ class CFRadialDataset(Dataset):
             self.domain_left_edge = np.array([x.min(), y.min(), z.min()])
             self.domain_right_edge = np.array([x.max(), y.max(), z.max()])
             self.dimensionality = 3
-            dims = [xr_ds_handle.dims[d] for d in "xyz"]
+            dims = [xr_ds_handle.sizes[d] for d in "xyz"]
             self.domain_dimensions = np.array(dims, dtype="int64")
             self._periodicity = (False, False, False)
-            self.current_time = float(xr_ds_handle.time.values)
+            self.current_time = float(xr_ds_handle.time.values[0])
 
         # Cosmological information set to zero (not in space).
         self.cosmological_simulation = 0
--- a/yt/frontends/cf_radial/tests/test_cf_radial_pytest.py
+++ b/yt/frontends/cf_radial/tests/test_cf_radial_pytest.py
@@ -1,0 +1,50 @@
+from pathlib import Path
+
+import numpy as np
+
+import yt
+from yt.frontends.cf_radial.data_structures import CFRadialDataset
+from yt.testing import requires_module
+
+
+def create_cf_radial_mock_gridded_ds(savedir: Path) -> Path:
+    # a minimal xarray dataset that should be identified as valid
+    import xarray as xr
+
+    file_to_save = savedir / "mock_gridded_cfradial.nc"
+
+    shp = (16, 16, 16)
+    xyz = {dim: np.linspace(0.0, 1.0, shp[idim]) for idim, dim in enumerate("xyz")}
+    da = xr.DataArray(data=np.ones(shp), coords=xyz, name="reflectivity")
+    ds_xr = da.to_dataset()
+    ds_xr.attrs["conventions"] = "CF/Radial"
+    ds_xr.x.attrs["units"] = "m"
+    ds_xr.y.attrs["units"] = "m"
+    ds_xr.z.attrs["units"] = "m"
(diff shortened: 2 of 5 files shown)


chrishavlin:
note that this was based off of main, so does not include #5042 and will fail initially (so marking as draft for now)

yt/frontends/cf_radial/data_structures.py
```diff
@@ -278,10 +278,10 @@ def _parse_parameter_file(self):
             self.domain_left_edge = np.array([x.min(), y.min(), z.min()])
             self.domain_right_edge = np.array([x.max(), y.max(), z.max()])
             self.dimensionality = 3
-            dims = [xr_ds_handle.dims[d] for d in "xyz"]
+            dims = [xr_ds_handle.sizes[d] for d in "xyz"]
             self.domain_dimensions = np.array(dims, dtype="int64")
             self._periodicity = (False, False, False)
-            self.current_time = float(xr_ds_handle.time.values)
+            self.current_time = float(xr_ds_handle.time.values[0])
```
chrishavlin, line 284: this is a fix for the np deprecation warning:

> DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)

Selecting the first index in the time array is consistent with choosing the first index of origin_latitude and origin_longitude (up around lines 275,276), which are both arrays that have time as a dimension.

yt/frontends/cf_radial/data_structures.py
```diff
@@ -278,10 +278,10 @@ def _parse_parameter_file(self):
             self.domain_left_edge = np.array([x.min(), y.min(), z.min()])
             self.domain_right_edge = np.array([x.max(), y.max(), z.max()])
             self.dimensionality = 3
-            dims = [xr_ds_handle.dims[d] for d in "xyz"]
+            dims = [xr_ds_handle.sizes[d] for d in "xyz"]
```
chrishavlin, line 281: this is a fix for the deprecation warning coming from xarray -- note that the sizes attribute has been around a long time (xarray vs 0.7ish maybe?) so no worry about backwards compatibility in yt.

yt/frontends/cf_radial/data_structures.py
neutrinoceros: If we're going to convert to a Python `float` anyway, I believe this is the idomatic way to do it
```suggestion
            self.current_time = xr_ds_handle.time.values.item()
```

yt/frontends/cf_radial/tests/test_cf_radial_pytest.py
```diff
@@ -0,0 +1,50 @@
...
+    ds_xr.attrs["conventions"] = "CF/Radial"
+    ds_xr.x.attrs["units"] = "m"
+    ds_xr.y.attrs["units"] = "m"
+    ds_xr.z.attrs["units"] = "m"
+    ds_xr.reflectivity.attrs["units"] = ""
+
+    t = np.array([0.0, 1.0])
+    ds_xr = ds_xr.assign_coords({"time": t})
+    ds_xr["origin_latitude"] = xr.DataArray(np.zeros(t.shape), dims=("time"))
+    ds_xr["origin_longitude"] = xr.DataArray(np.zeros(t.shape), dims=("time"))
```
neutrinoceros, line 29:
```suggestion
    ds_xr["origin_latitude"] = xr.DataArray(np.zeros_like(t), dims=("time"))
    ds_xr["origin_longitude"] = xr.DataArray(np.zeros_like(t), dims=("time"))
```
also, I don't know what is the intention in `dims=("time")` but if you meant to write a tuple, note that you need `dims=("time",)`, otherwise you just have a parenthesised string, equivalent to `dims="time"` (because the syntactic element that makes a tuple isn't the parentheses, it's the comma)

neutrinoceros (changes requested):
I just branched yt-4.4.x but we could probably still backport this (I don't expect to do the actual release today).

yt/frontends/cf_radial/tests/test_cf_radial_pytest.py
chrishavlin: good catch, forgot the comma. i did mean a tuple... but looks like xarray doesn't require a tuple if it's a single element. i'll correct it.

chrishavlin:
getting a new error locally after applying the changes, gotta run at the moment though so going to switch back to draft for now

yt/frontends/cf_radial/data_structures.py
```diff
@@ -278,10 +278,10 @@ def _parse_parameter_file(self):
...
             self.domain_dimensions = np.array(dims, dtype="int64")
             self._periodicity = (False, False, False)
-            self.current_time = float(xr_ds_handle.time.values)
+            self.current_time = xr_ds_handle.time.values.item()
```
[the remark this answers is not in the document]
chrishavlin: and i also remembered that the original reason for the cast to float was that the time arrays here are often datetime objects... so i'll add back the cast and add a comment in the next push.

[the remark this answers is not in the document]
chrishavlin: fixed in latest push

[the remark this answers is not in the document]
chrishavlin: (also added a comment about why we need to cast to float and adjusted the mock ds in the tests to use a datetime array for time)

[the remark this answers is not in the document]
chrishavlin: OH, i totally spaced on what `item()` does (it converts an element to a scalar). But looks like it ends up an int for datetime objects:

```
import numpy as np
times = np.array(["2017-05-19T01", "2017-05-19T01:01"], dtype="datetime64[ns]")
print(times.item(0), type(times.item(0)))
1495155600000000000 <class 'int'>
```

but I think having a float is more appropriate... will fix this in the next push when i get to it.

chrishavlin:
@neutrinoceros i'm going to want to rebase this again before it's ready, but i gotta run, not sure if i'll be able to get to it before next monday, so don't worry about backporting before releasing 4.4.0. it can wait to 4.4.1.

yt/frontends/cf_radial/data_structures.py
chrishavlin: So I ended up switching this back because the `time` coordinate here will typically have a datetime  dtype, and `.item()` will cast it to int. But I think it's more appropriate to cast to float here (both cause `current_time` will end up being cast to float when it's upgrade to a unyt quantity and also cause internally, xarray will encode/decode datetime objects with floats, so it's helpful to be consistent with that).

chrishavlin:
Oh! new arm-pyart release just dropped... i'm going to re-run tests here just to make sure we're still green with the new release.

chrishavlin:
@yt-fido test this please

neutrinoceros (approved):

Issue #5047, opened 2024-11-08
failing nosetests (Failed to decode variable 'time')


chrishavlin:
Both #5045  and #5046 failed the nose tests with unrelated failures:

```python-traceback
Traceback (most recent call last):
  File "/home/fido/workspace/yt_py310_git/.venv/lib/python3.10/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/home/fido/workspace/yt_py310_git/.venv/lib/python3.10/site-packages/nose/loader.py", line 417, in loadTestsFromName
    module = self.importer.importFromPath(
  File "/home/fido/workspace/yt_py310_git/.venv/lib/python3.10/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/fido/workspace/yt_py310_git/.venv/lib/python3.10/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/usr/lib/python3.10/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "/usr/lib/python3.10/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 719, in _load
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/fido/workspace/yt_py310_git/yt/frontends/cf_radial/tests/test_outputs.py", line 176, in <module>
    @requires_ds(cf)
  File "/home/fido/workspace/yt_py310_git/yt/utilities/answer_testing/framework.py", line 1094, in requires_ds
    condition = (big_data and not run_big_data) or not can_run_ds(ds_fn, file_check)
  File "/home/fido/workspace/yt_py310_git/yt/utilities/answer_testing/framework.py", line 315, in can_run_ds
    load(ds_fn)
  File "/home/fido/workspace/yt_py310_git/yt/_maintenance/deprecation.py", line 68, in inner
    return func(*args, **kwargs)
  File "/home/fido/workspace/yt_py310_git/yt/loaders.py", line 141, in load
    return cls(fn, *args, **kwargs)
[...]

PR #5048, opened 2024-11-08
switch xradar pin to arm-pyart pin in tests

### LINKED ISSUE #5044: Unpin xradar in tests when possible
This is just a reminder to unpin xradar in tests when feasible.

Relevant issues and PRs:

*  The current patch for yt CI: #5042
* #5041 
* https://github.com/ARM-DOE/pyart/issues/1680
* https://github.com/ARM-DOE/pyart/pull/1681

Once the issue is addressed upstream we should switch from pinning xradar in yt's test deps to pinning arm-pyart (because arm-pyart is the direct dependency used by yt and will be clearer than pinning xradar).

chrishavlin:
new arm-pyart release is out. So close #5044 ... and hopefully #5047 ??

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,7 @@ artio = []
 athena = []
 athena-pp = []
 boxlib = []
-cf-radial = ["xarray>=0.16.1", "arm-pyart!=1.12.5,>=1.11.4"]
+cf-radial = ["xarray>=0.16.1", "arm-pyart>=1.19.2",]
 chimera = ["yt[HDF5]"]
 chombo = ["yt[HDF5]"]
 cholla = ["yt[HDF5]"]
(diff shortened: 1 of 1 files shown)


chrishavlin:
I just realized that  anything that is added to the `pyproject.toml` `test` target will always be installed, even in the minimal test suite (the `ci_install.sh` script will do a `python -m pip install -e ".[test]" -r minimal_requirements.txt` for the minimal test). So having `arm-pyart` (or previously `xradar`) in the test target will mean that it was always be installed (which will include xarray and lots of other dependencies). That doesn't seem ideal... I'm tempted to just pin `arm-pyart` in the frontend requirements instead?

pyproject.toml
```diff
@@ -200,7 +200,7 @@ test = [
     "pytest-mpl>=0.16.1",
     "sympy!=1.10,!=1.9", # see https://github.com/sympy/sympy/issues/22241
     "imageio!=2.35.0", # see https://github.com/yt-project/yt/issues/4966
-    "xradar<0.8.0",  # see https://github.com/yt-project/yt/issues/5041
+    "arm-pyart>=1.19.2", # see https://github.com/yt-project/yt/issues/5044
```
neutrinoceros, line 203: shouldn't we move it to `[project.optional-dependencies]` ?

chrishavlin:
ok, just pushed a new version to just bump the min arm-pyart version. It's a pretty big jump in versions from what we had... but realistically I think it'd be tough to get a working environment at v<1.19.2 with all the shuffling of the arm-pyart dependencies that happened.

neutrinoceros (approved):

PR #5051, opened 2024-11-11
RFC: refactor geometry matching clauses with Python 3.10's pattern matching

neutrinoceros:
## PR Summary

Now that we don't support 3.9 anymore we have access to `match`/`case` statements (pattern matching). Arguably this improves readability in these instances, albeit at the cost of an additional indentation level.

### DIFF
--- a/yt/data_objects/selection_objects/data_selection_objects.py
+++ b/yt/data_objects/selection_objects/data_selection_objects.py
@@ -609,23 +609,22 @@ class YTSelectionContainer2D(YTSelectionContainer):
         >>> write_image(np.log10(frb["gas", "density"]), "density_100kpc.png")
         """
 
-        if (self.ds.geometry is Geometry.CYLINDRICAL and self.axis == 1) or (
-            self.ds.geometry is Geometry.POLAR and self.axis == 2
-        ):
-            if center is not None and center != (0.0, 0.0):
-                raise NotImplementedError(
-                    "Currently we only support images centered at R=0. "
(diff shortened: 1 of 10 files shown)


matthewturk (dismissed):
Oh, I like this a lot more.  Thank you.

Quick question -- is there something about the way the `|` is evaluated in the cases that means it has to be put in the case?  Or, could we create top-level special cases like `HAS_Z_AXIS` for `CARTESIAN | POLAR | SPECTRAL_CUBE` ?

neutrinoceros:
> is there something about the way the | is evaluated in the cases that means it has to be put in the case? 

If you're asking if `|` has a syntactically different meaning in the context of `case ` clauses, the answer is yes.
See https://peps.python.org/pep-0636/#or-patterns

> could we create top-level special cases like HAS_Z_AXIS for CARTESIAN | POLAR | SPECTRAL_CUBE ?

I think this would be achievable using [`IntFlag` enums](https://docs.python.org/3/library/enum.html#enum.IntFlag), however we need `geometry` to be an `StrEnum` for backward compatibility, and I don't fancy trying to implement a common subclass.

neutrinoceros:
I want to push couple times to this branch before I can undraft this.

chrishavlin (approved):
ok, i'm now convinced of match/case's usefulness :) I was kinda ambivalent about it before.

neutrinoceros:
I'll merge now to save myself some conflict solving later.

PR #5059, opened 2024-11-20
[BUG] Fix off-axis rendering when center is not [0.5, 0.5, 0.5], fix periodicity

cphyc:
## PR Summary

The off-axis renderer introduced in #4741 has an issue when the center and the center in rotated frame do not coincide (which happens when it isn't 0.5, 0.5, 0.5 afaik).

In addition, it wraps around the boundaries of the width of the plotting window rather than the simulation domain (if required).

This PR fixes both.

## PR Checklist

## Bug reproduction:

```python
import numpy as np
import yt

ad = ds.all_data()

center = ds.arr(ad.argmax("density"))

sp = ds.sphere(center, (200, "kpc"))

n = np.array([1, 1, 1.])
n /= np.linalg.norm(n)
p = yt.ProjectionPlot(ds, n, "density", center=center, data_source=sp, width=(400, "kpc"), weight_field=("gas", "density"))
p.set_zlim(("gas", "density"), 1e-4, 5)
p.save("frames/")
```

| Before | With this PR |
|---|---|
| info_00080_OffAxisProjection_density_density|  info_00080_OffAxisProjection_density_density |

### DIFF
--- a/yt/visualization/tests/test_offaxisprojection.py
+++ b/yt/visualization/tests/test_offaxisprojection.py
@@ -16,6 +16,7 @@ from yt.testing import (
 from yt.visualization.api import (
     OffAxisProjectionPlot,
     OffAxisSlicePlot,
+    ProjectionPlot,
 )
 from yt.visualization.image_writer import write_projection
 from yt.visualization.volume_rendering.api import off_axis_projection
@@ -208,6 +209,35 @@ def test_field_cut_off_axis_octree():
     assert_equal((p3.frb["gas", "density"] == p4.frb["gas", "density"]).all(), False)
     p4rho = p4.frb["gas", "density"]
     assert_equal(np.nanmin(p4rho[p4rho > 0.0]) >= 0.5, True)
+
+
+def test_off_axis_octree():
+    ds = fake_octree_ds()
+    p1 = ProjectionPlot(
+        ds,
+        "x",
+        ("gas", "density"),
+        center=[0.6] * 3,
+        width=0.8,
+        weight_field=("gas", "density"),
+    )
+    p2 = OffAxisProjectionPlot(
+        ds,
+        [1, 0, 0],
+        ("gas", "density"),
+        center=[0.6] * 3,
+        width=0.8,
+        weight_field=("gas", "density"),
+    )
+
+    # Note: due to our implementation, the off-axis projection will have a
+    # slightly blurred cell edges so we can't do an exact comparison
+    v1, v2 = p1.frb["gas", "density"], p2.frb["gas", "density"]
+    diff = (v1 - v2) / (v1 + v2) * 2
+
+    # Make sure the difference is zero-centered with a small standard deviation
+    assert np.mean(diff).max() < 1e-3  # 0.1%: very little bias
+    assert np.std(diff) < 0.05  # <2% error on average
 
 
 def test_offaxis_moment():
--- a/yt/visualization/volume_rendering/off_axis_projection.py
+++ b/yt/visualization/volume_rendering/off_axis_projection.py
@@ -444,21 +444,23 @@ def off_axis_projection(
         # We need the width of the plot window in projected coordinates,
         # i.e. we ignore the z-component
         wmax = width[:2].max()
-
-        # Normalize the positions & dx so that they are in the range [-0.5, 0.5]
-        xyz = np.stack(
-            [
-                ((data_source["index", k] - center[i]) / wmax).to("1").d
-                for i, k in enumerate("xyz")
-            ],
-            axis=-1,
+        xyz = data_source.ds.arr(
+            np.zeros((len(data_source[vol.field]), 3)), "code_length"
         )
 
         for idim, periodic in enumerate(data_source.ds.periodicity):
+            axis = data_source.ds.coordinates.axis_order[idim]
+            # Recenter positions w.r.t. center of the plot window
+            xyz[..., idim] = data_source["index", axis] - center[idim]
             if not periodic:
                 continue
-            # Wrap into [-0.5, +0.5]
-            xyz[..., idim] = (xyz[..., idim] + 0.5) % 1 - 0.5
+            # If we have periodic boundaries, we need to wrap the corresponding
+            # coordinates into [-w/2, +w/2]
+            w = data_source.ds.domain_width[idim]
+            xyz[..., idim] = (xyz[..., idim] + w / 2) % w - w / 2
+
+        # Rescale to [-0.5, +0.5]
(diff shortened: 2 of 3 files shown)


cphyc:
@yt-fido test this please

yt/visualization/tests/test_offaxisprojection.py
```diff
@@ -210,6 +211,35 @@ def test_field_cut_off_axis_octree():
...
+        ("gas", "density"),
+        center=[0.6] * 3,
+        width=0.8,
+        weight_field=("gas", "density"),
+    )
+
+    # Note: due to our implementation, the off-axis projection will have a
+    # slightly blurred cell edges so we can't do an exact comparison
+    v1, v2 = p1.frb["gas", "density"], p2.frb["gas", "density"]
+    diff = (v1 - v2) / (v1 + v2) * 2
```
chrishavlin, line 236: Would you mind re-writing this to make it a little clearer on a quick glance that you're comparing the difference to the mean of the two values (unless there's a computational reason that I'm missing)? took me longer than it should have to mentally parse it (though that might be due to a caffeine deficit).

i.e.,:

```python
diff = (v1 - v2) / (( v1 + v2) / 2.0)
```

also, regardless of that, it would be good to include an absolute value as a large negative error would pass the assert test on the following line.

yt/visualization/volume_rendering/off_axis_projection.py
```diff
@@ -444,21 +444,23 @@ def temp_weightfield(field, data):
...
-                ((data_source["index", k] - center[i]) / wmax).to("1").d
-                for i, k in enumerate("xyz")
-            ],
-            axis=-1,
+        # Recenter positions w.r.t. center of the plot window
+        xyz = data_source.ds.arr(
+            np.zeros((len(data_source[vol.field]), 3)), "code_length"
         )
 
+        # If we have periodic boundaries, we need to wrap the positions
```
chrishavlin, line 452: can you move this comment to below the 

```
if not periodic:
    continue
```` 

statement? The comment makes it seem like this whole loop is just for periodicity but it's now also calculating `xyz` for all dimensions.

File: yt/visualization/volume_rendering/off_axis_projection.py [hunk 3]
```diff
@@ -444,21 +444,23 @@ def temp_weightfield(field, data):
...
         for idim, periodic in enumerate(data_source.ds.periodicity):
+            axis = "xyz"[idim]
```
chrishavlin, line 454: Do you need to be handling axis-ordering here? I can't remember if octree datasets are required to be ordered xyz.  e.g., 

```suggestion
            axis = data_source.ds.coordinates.axis_order[idim]
```

chrishavlin:
well hmmm....

>  FAILED yt/visualization/tests/test_offaxisprojection.py::test_off_axis_octree - AssertionError: assert unyt_quantity(0.38448323, '(dimensionless)') < 0.2 

Maybe that check I made you add is too restrictive?

chrishavlin:
So part of the trickiness is that the test is highly variable between runs: I just ran the code from `test_off_axis_octree` locally, recording the maximum of the absolute value in the difference and over 10 runs, the max abs fractional difference was `[0.17469636745540013, 0.22734252938752475, 0.5588387477439833, 0.06742315690744759, 0.19014792348598422, 0.32429411325319835, 0.24292832837751724, 0.20697617456665565, 0.3351829388729374, 0.09990611014962943]`. Setting a seed would help make it deterministic... but the large variation might be indicative of some other issue? Or maybe it's expected and you should just revert the comparison of the absolute value in difference?

I did check what the diffs look like, the large differences are always at cell edges. e.g., for one case:


image

for which the standard axis aligned `ProjectionPlot` with annotated cells looks like: 

Screenshot from 2024-11-26 10-39-06

cphyc:
OK - I've actually uncovered a funny one. 
```python
import numpy as np
import yt
ds = yt.load("output_00080")
p1 = yt.ProjectionPlot(
    ds,
    "x",
    ("gas", "density"),
    center=[0.5] * 3,
    width=.8,
    #weight_field=("gas", "density"),
)
p2 = yt.OffAxisProjectionPlot(
    ds,
    [1, 0, 0],
    ("gas", "density"),
    center=[0.5] * 3,
    width=.8,
    #weight_field=("gas", "density"),
)

# Note: due to our implementation, the off-axis projection will have a
# slightly blurred cell edges so we can't do an exact comparison
v1, v2 = p1.frb["gas", "density"], p2.frb["gas", "density"]

fig = plt.figure()
plt.imshow(v1, origin="lower", norm=plt.matplotlib.colors.LogNorm())
plt.colorbar()
fig.savefig("/tmp/1.png")

fig = plt.figure()
plt.imshow(v2, origin="lower", norm=plt.matplotlib.colors.LogNorm())
plt.colorbar()
fig.savefig("/tmp/2.png")

fig = plt.figure()
plt.imshow(v1 / v2, origin="lower")
plt.colorbar()
fig.savefig("/tmp/3.png")
print(np.median(v1 / v2), np.mean(v1 / v2), np.std(v1 / v2))
```

This returns:
| Projection | Off-axis projection | Difference |
|---|---|---|
||||

The funny thing is that the ratio between the on-axis and the off-axis projection is equal (within some uncertainties) to `plot_width * 0.5675` (above, `plot_width = 0.8`). This is the case for a RAMSES dataset (here), but the same can be obtained using `fake_random_ds` instead, yielding the same value.

As much as I see how I could have missed a `plot_width` factor, what the heck is `0.5675`?

chrishavlin:
ya, I'm thoroughly confused now. that is a weird number to be off by... and looking at projections of index fields like ('index', 'ones') or ('index', 'y'), the cell edges pop out very clearly and the projected values are all smaller than they should be, even in cell interiors  ([example notebook](https://github.com/chrishavlin/yt_scratch/blob/master/notebooks/dev/octree_off_axis_proj_issue.ipynb)).

cphyc:
@chrishavlin I think fixing this bug above is *definitely* important, but may be part of another PR. Would you be fine with the following plan?
1. Roll back the test on the absolute value of the difference, since this is hard to get stable,
2. Create an issue with milestone 4.4.1 describing the mismatch when using a projection with no weight,
3. Have this PR merged as is?

Regarding point 1, it think it is possible to correct the cell-edges effects, but I don't have the time to think carefully about it. If you are adamant this should be done, I'm happy to open another issue to keep track of it.

chrishavlin:
@cphyc sounds like a good plan to me!

cphyc:
@chrishavlin, I have opened the issue and rebased this PR.

chrishavlin (approved):

PR #5060, opened 2024-11-20
[ENH] RAMSES: set max_level from parameter file

cphyc:
## PR Summary

RAMSES contains a text file that describes some basic element of the AMR structure, including the maximum level of refinment. However, that level may not be reached. Up to this PR, we would read all AMR domain files to figure out to _effective_ max level of refinment. This is however a costly operation for large simulations; this PR returns a more conservative number for the max level (the highest achievable by the sim) and is MUCH faster.

## PR Checklist

- [x] New features are documented, with docstrings and narrative docs -> no new feature
- [x] Adds a test for any bugs fixed. Adds tests for new features. -> all existing tests should pass


yt/frontends/ramses/data_structures.py
```diff
@@ -312,12 +302,7 @@ def amr_offset(self):
 
     @property
     def max_level(self):
-        self._read_amr_header()
-        return self._max_level
-
-    @max_level.setter
-    def max_level(self, value):
-        self._max_level = value
+        return self.ds.max_level
```
cphyc, line 305: See https://github.com/cphyc/yt/blob/6ed3f95b20c10a33d9c93580a8d89c22e388be36/yt/frontends/ramses/data_structures.py#L1104 for the definition.

matthewturk:
I'm not 100% on this.  In my mind, the idea of the `max_level` parameter should tell the actual-thing-it-is, and the access of the parameter should give what the parameter is set to.

cphyc:
I'm not sure I understand what you mean.

To clarify, before this PR, `max_level` would be the *actual* level reached anywhere in the simulation. This PR replaces it with the maximum level that *can* (but may not) be reached anywhere in the simulation.

The motivation is that the former definition requires reading the entirety of the AMR structure of the simulation, while the latter only requires parsing one small text file.

As far as I could tell, the only place where `max_level` was being used was to make sure that, when selecting a sphere, the radius is at least one `dx` wide, with `dx` being the best resolution. Note that this doesn't prevent a sphere from having 0 cells (for example, if it is centred around a low-resolution region), so I changed the definition of `max_level`. If anything, it makes the test a bit more relaxed.

matthewturk:
Oh, I think I understand -- and I can see the value in speeding it up!  What I'm saying is that the usage of `max_level` in Enzo is the max level in the index, whereas `MaximumRefinementLevel` (also in Enzo) is how you get the maximum allowed.  Just that the property was actual, the parameter was allowed.

But maybe this is OK for RAMSES -- or maybe for the specific case of spheres, we don't do the check?

cphyc:
I see. I guess the issue is that, although `max_level` is defined for all datasets afaik, `MaximumRefinementLevel` is Enzo-specific. There is a similar parameter for RAMSES, but it has a different name (it's `levelmax`).

I can explore the idea of replacing occurrences of `max_level` with `max_potential_level` or some other jargon signalling it is an upper boundary to see what it'd take to have this and will set this PR in draft mode in the meantime.

matthewturk:
Is it possible to do an extremely quick pass to get the max level, without parsing the full file?  I'm not able to convince myself.

chrishavlin:
hey @cphyc over in #5211 you say: "Note: this makes https://github.com/yt-project/yt/pull/5060 more or less useless." So we should close this one, ya? (I just removed it from the 4.5.0 milestone)

PR #5064, opened 2024-11-23
STY: manual fixes for newly flagged violations of UP031

neutrinoceros:
## PR Summary
ruff 0.8.0 is just out and this version flags *all* violations to rule UP031, including the ones it cannot autofix.
In preparation for the auto upgrade, I'm fixing all these by hand.
At the time of opening, I only fixed 62/162 errors, but I'd like to run CI at this point to see if I've already introduced a mistake that the test suite would catch.

useful resources for reviewers:
- https://cplusplus.com/reference/cstdio/printf/
- https://fstring.help/
- https://docs.astral.sh/ruff/rules/printf-string-formatting/
- https://astral.sh/blog/ruff-v0.8.0

## PR Checklist

### DIFF
--- a/yt/data_objects/construction_data_containers.py
+++ b/yt/data_objects/construction_data_containers.py
@@ -2591,7 +2591,7 @@ class YTSurface(YTSelectionContainer3D):
         ]
         f.write(b"ply\n")
         f.write(b"format binary_little_endian 1.0\n")
-        line = "element vertex %i\n" % (nv)
+        line = f"element vertex {nv}\n"
         f.write(line.encode("latin-1"))
         f.write(b"property float x\n")
         f.write(b"property float y\n")
@@ -2612,7 +2612,7 @@ class YTSurface(YTSelectionContainer3D):
             )
         else:
             v = np.empty(self.vertices.shape[1], dtype=vs[:3])
-        line = "element face %i\n" % (nv / 3)
+        line = f"element face {int(nv/3)}\n"
         f.write(line.encode("latin-1"))
         f.write(b"property list uchar int vertex_indices\n")
         if color_field is not None and sample_type == "face":
@@ -2751,9 +2751,9 @@ class YTSurface(YTSelectionContainer3D):
         # to a file.
         if self.vertices.shape[1] > 1e7:
             tfi = 0
-            fn = "temp_model_%03i.ply" % tfi
+            fn = f"temp_model_{tfi:03}.ply"
             while os.path.exists(fn):
-                fn = "temp_model_%03i.ply" % tfi
+                fn = f"temp_model_{tfi:03}.ply"
                 tfi += 1
             open(fn, "wb").write(ply_file.read())
             raise YTTooManyVertices(self.vertices.shape[1], fn)
--- a/yt/data_objects/particle_trajectories.py
+++ b/yt/data_objects/particle_trajectories.py
@@ -335,7 +335,7 @@ class ParticleTrajectories:
         """
         mask = np.isin(self.indices, (index,), assume_unique=True)
         if not np.any(mask):
-            print("The particle index %d is not in the list!" % (index))
+            print(f"The particle index {index} is not in the list!")
             raise IndexError
         fields = sorted(self.field_data.keys())
         traj = {}
@@ -375,7 +375,7 @@ class ParticleTrajectories:
                         [self.times[it]] + [self[field][ix, it] for field in fields]
                     )
                 )
-            fid = open(filename_base + "_%d.dat" % self.indices[ix], "w")
+            fid = open(f"{filename_base}_{self.indices[ix]}.dat", "w")
             fid.writelines(outlines)
             fid.close()
             del fid
--- a/yt/frontends/amrex/data_structures.py
+++ b/yt/frontends/amrex/data_structures.py
@@ -196,9 +196,7 @@ class BoxLibParticleHeader:
         self.known_int_fields = self.known_int_fields[0 : self.num_int_base]
 
         # these are extra integer fields
-        extra_int_fields = [
-            "particle_int_comp%d" % i for i in range(self.num_int_extra)
-        ]
+        extra_int_fields = [f"particle_int_comp{i}" for i in range(self.num_int_extra)]
         self.known_int_fields.extend(
             [(self.particle_type, field) for field in extra_int_fields]
         )
@@ -216,7 +214,7 @@ class BoxLibParticleHeader:
             assert len(extra_field_names) == self.num_real_extra
         else:
             extra_field_names = [
-                "particle_real_comp%d" % i for i in range(self.num_real_extra)
+                f"particle_real_comp{i}" for i in range(self.num_real_extra)
             ]
 
         self.known_real_fields.extend(
@@ -1478,7 +1476,7 @@ class WarpXHeader:
                 if len(line) == 1:
                     line = f.readline()
                     continue
-                self.data["species_%d" % i] = [float(val) for val in line]
+                self.data[f"species_{i}"] = [float(val) for val in line]
                 i = i + 1
                 line = f.readline()
 
@@ -1497,8 +1495,8 @@ class WarpXHierarchy(BoxlibHierarchy):
         for key, val in self.warpx_header.data.items():
             if key.startswith("species_"):
                 i = int(key.split("_")[-1])
-                charge_name = "particle%.1d_charge" % i
-                mass_name = "particle%.1d_mass" % i
+                charge_name = f"particle{i}_charge"
+                mass_name = f"particle{i}_mass"
                 self.parameters[charge_name] = val[0]
                 self.parameters[mass_name] = val[1]
 
--- a/yt/frontends/art/data_structures.py
+++ b/yt/frontends/art/data_structures.py
@@ -320,7 +320,7 @@ class ARTDataset(Dataset):
             self.parameters["wspecies"] = wspecies[:n]
             self.parameters["lspecies"] = lspecies[:n]
             for specie in range(n):
-                self.particle_types.append("specie%i" % specie)
+                self.particle_types.append(f"specie{specie}")
             self.particle_types_raw = tuple(self.particle_types)
             ls_nonzero = np.diff(lspecies)[: n - 1]
             ls_nonzero = np.append(lspecies[0], ls_nonzero)
@@ -611,7 +611,7 @@ class DarkMatterARTDataset(ARTDataset):
             else:
                 particle_header_vals[a1] = arr[:a2]
         for specie in range(n):
-            self.particle_types.append("specie%i" % specie)
+            self.particle_types.append(f"specie{specie}")
         self.particle_types_raw = tuple(self.particle_types)
         ls_nonzero = np.diff(lspecies)[: n - 1]
         ls_nonzero = np.append(lspecies[0], ls_nonzero)
--- a/yt/frontends/enzo/data_structures.py
+++ b/yt/frontends/enzo/data_structures.py
@@ -575,7 +575,7 @@ class EnzoHierarchyInMemory(EnzoHierarchy):
         for i, grid in enumerate(grids):
             if (i % 1e4) == 0:
                 mylog.debug("Prepared % 7i / % 7i grids", i, self.num_grids)
-            grid.filename = "Inline_processor_%07i" % (self.grid_procs[i, 0])
+            grid.filename = f"Inline_processor_{self.grid_procs[i, 0]:07}"
             grid._prepare_grid()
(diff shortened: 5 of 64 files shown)


yt/frontends/amrex/data_structures.py
```diff
@@ -1497,8 +1495,8 @@ def __init__(self, ds, dataset_type="boxlib_native"):
         for key, val in self.warpx_header.data.items():
             if key.startswith("species_"):
                 i = int(key.split("_")[-1])
-                charge_name = "particle%.1d_charge" % i
```
neutrinoceros, line 1500: this is a weird one. As far as I understand `%.1d` is accepted but meaningless, and simply equivalent to `%d`.

yt/frontends/art/data_structures.py
```diff
@@ -320,7 +320,7 @@ def _parse_parameter_file(self):
             self.parameters["wspecies"] = wspecies[:n]
             self.parameters["lspecies"] = lspecies[:n]
             for specie in range(n):
-                self.particle_types.append("specie%i" % specie)
+                self.particle_types.append(f"specie{specie}")
```
neutrinoceros, line 323: I believe in English, 'species' is actually invariable, but I'm keeping bug-for-bug compatibility in this pure refactor

yt/frontends/gadget/testing.py
```diff
@@ -27,7 +27,7 @@ def write_record(fp, data, endian):
 
 def write_block(fp, data, endian, fmt, block_id):
     assert fmt in [1, 2]
-    block_id = "%-4s" % block_id
```
neutrinoceros, line 30: this is the first (and so far only) time I've ever seen this format specifier.
for reference, let me remind reviewers of:
- https://cplusplus.com/reference/cstdio/printf/
- https://fstring.help/

yt/frontends/ramses/data_structures.py
neutrinoceros, line 1: for this file, refactors are quite involved, so I would like to request a review from @cphyc

yt/frontends/ramses/data_structures.py
```diff
@@ -184,21 +184,15 @@ def __init__(self, ds, domain_id):
         self.ds = ds
         self.domain_id = domain_id
 
-        num = os.path.basename(ds.parameter_filename).split(".")[0].split("_")[1]
-        rootdir = ds.root_folder
-        basedir = os.path.abspath(os.path.dirname(ds.parameter_filename))
-        basename = "%s/%%s_%s.out%05i" % (basedir, num, domain_id)
-        part_file_descriptor = f"{basedir}/part_file_descriptor.txt"
+        num = ds.basename.split(".")[0].split("_")[1]
+        basename = os.path.join(ds.directory, f"%s_{num}.out{domain_id:05}")
```
neutrinoceros, line 188: I'm being 100% backward compatible here, but I'm not completely sure if the leftover `%s` is intentional.

yt/utilities/grid_data_format/conversion/conversion_athena.py
```diff
@@ -450,7 +451,7 @@ def write_to_gdf(self, fn, grid):
 
         ## --------- Store Grid Data --------- ##
 
-        g0 = data_g.create_group("grid_%010i" % 0)
+        g0 = data_g.create_group("grid_{0:010}")
```
neutrinoceros, line 454: this one is a little suprising perharps. Effectively this is equivalent to hardcoding `'0000000000'` or, perharps more clearly `'0'*10`. What's disturbing is that the first `0` is sort of meaningless: one would get the same result from `f"{'':010}"`

neutrinoceros:
Alright. This took me... 3 hours (!?) but this time includes a pass of self-review were I was able to catch a few (hopefully most, or even all) of my mistakes. The failure on macOS is unrelated (see #5065), so this should now be ready for review.

yt/frontends/art/data_structures.py
chrishavlin: the [etymology rabbit hole for "specie"](https://en.wiktionary.org/wiki/specie#English) is kinda fun. I could be convinced that it's more appropriate to use specie for particle types in a numerical simulation: maybe having a variety of particle types is more similar to minting different types of coins than it is to variations in taxonomical classification. But most likely species was meant here :)

yt/data_objects/particle_trajectories.py
```diff
@@ -335,7 +335,7 @@ def trajectory_from_index(self, index):
         """
         mask = np.isin(self.indices, (index,), assume_unique=True)
         if not np.any(mask):
-            print("The particle index %d is not in the list!" % (index))
+            print(f"The particle index {index} is not in the list!")
```
chrishavlin, line 338: i know you're going for a 1:1 refactor here... but what about moving this string to the actual `IndexError` below?

chrishavlin (commented):
made it about halfway and need a break :) couple of questions so far.

yt/data_objects/particle_trajectories.py
neutrinoceros: Good idea, but I'd prefer to do it in a follow up PR. I'm worried about making a breaking change for anyone and unintentionally hiding it behind a giant refactor.

yt/frontends/athena/data_structures.py
```diff
@@ -297,7 +297,8 @@ def _parse_index(self):
                 gridread["dimensions"][gridread["dimensions"] == 0] = 1
             if np.prod(gridread["dimensions"]) != gridread["ncells"]:
                 mylog.error(
-                    "product of dimensions %i not equal to number of cells %i",
+                    f"product of dimensions {np.prod(grid['dimensions'])} "
+                    f"not equal to number of cells {grid['ncells']}",
```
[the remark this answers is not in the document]
neutrinoceros: Let me draft this until I have time to dig into this.

yt/frontends/ramses/data_structures.py
cphyc: It is - it is formatted later with a % to match files named e.g. `output_00123/hydro_XXXXX.outYYYYY`, `part_XXXXX.outYYYYY`, etc.

cphyc (commented):
In general, you have replaced all occurences of `%d` and `%i` (which are equivalent) with `f"{whatever}`. This isn't 100% equivalent, see for example this example:
```python
a = 10
print("a=%i" % a, f"a={a}")  # a=10 a=10

v = 10.
print("v=%i" % v, f"v={v}")  # v=10 v=10.
```
I know most of the values being formatted are actual ints, but if not we have diverging implementations.

Note that formatting with `f"{x:d}"` will raise an exception if `x` is not of type int...

neutrinoceros:
> Note that formatting with f"{x:d}" will raise an exception if x is not of type int...

yeah that's why I didn't try to preserve these format specifiers... If you'd like to give a try to re-introducing them, we can probably co-write this PR, but as far as I'm concerned I already spent much longer than I wanted to here so I don't fancy trying it out myself.

yt/data_objects/construction_data_containers.py
```diff
@@ -2612,7 +2612,7 @@ def _export_ply(
             )
         else:
             v = np.empty(self.vertices.shape[1], dtype=vs[:3])
-        line = "element face %i\n" % (nv / 3)
+        line = f"element face {nv/3}\n"
```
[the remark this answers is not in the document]
neutrinoceros: nice catch. I think your second suggestion is more in line with the original. Thanks !

yt/frontends/ramses/data_structures.py
neutrinoceros: thanks !

yt/frontends/athena/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: now that I look at it from an real size computer (as opposed to my phone) I see that this change was incorrect *an* uncalled for (no actual linting violation here). I'll revert it. Thanks a lot for catching it è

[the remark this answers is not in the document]
neutrinoceros: (for context, I think what happened is that I ran a find-and-replace for similar looking lines and *this* one turned bad but went under my radar)

yt/frontends/enzo/data_structures.py
```diff
@@ -574,8 +574,8 @@ def _parse_index(self):
         self.grids = np.empty(len(grids), dtype="object")
         for i, grid in enumerate(grids):
             if (i % 1e4) == 0:
-                mylog.debug("Prepared % 7i / % 7i grids", i, self.num_grids)
-            grid.filename = "Inline_processor_%07i" % (self.grid_procs[i, 0])
+                mylog.debug(f"Prepared {i:>7} / {self.num_grids:>7} grids")
```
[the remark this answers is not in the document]
neutrinoceros: probably just my brain going squishy after 2hrs of doing this. Thanks for catching it !

yt/funcs.py
```diff
@@ -137,7 +137,7 @@ def humanize_time(secs):
     """
     mins, secs = divmod(secs, 60)
     hours, mins = divmod(mins, 60)
-    return "%02d:%02d:%02d" % (hours, mins, secs)
+    return ":".join(f"{t:02}" for t in (hours, mins, secs))
```
chrishavlin, line 140: this is one case where i vastly prefer the older string formatting syntax :) 

but more importantly, if the incoming `secs` is a float (which it probably is? hard to say, this function doesn't seem to be used anywhere in yt...), the formatting differs significantly, should cast to float to keep the behavior identical:

```suggestion
    return ":".join(f"{int(t):02}" for t in (hours, mins, secs))
```

For reference, as written, the new version looks like

```
>>> humanize_time(1000.1)
'0.0:16.0:40.10000000000002'
```
vs the old

```
>>> humanize_time(1000.1)
'00:16:40'
```

yt/geometry/grid_geometry_handler.py
```diff
@@ -208,19 +208,17 @@ def print_stats(self):
...
-                % (
-                    level,
-                    self.level_stats["numgrids"][level],
-                    self.level_stats["numcells"][level],
-                    np.ceil(self.level_stats["numcells"][level] ** (1.0 / 3)),
-                )
+                f"{level:>3}\t"
+                f"{self.level_stats['numgrids'][level]:>6}\t"
+                f"{self.level_stats['numcells'][level]:>14}\t"
+                f"{np.ceil(self.level_stats['numcells'][level] ** (1.0 / 3)):>14}"
```
chrishavlin, line 214: `np.ceil` will return a float here -- would need  to cast to int here to keep it identical 

```suggestion
                f"{int(np.ceil(self.level_stats['numcells'][level] ** (1.0 / 3))):>14}"
```

yt/utilities/grid_data_format/conversion/conversion_athena.py
```diff
@@ -130,8 +130,8 @@ def read_and_write_index(self, basename, ddn, gdf_name):
                 grid["dimensions"][grid["dimensions"] == 0] = 1
             if np.prod(grid["dimensions"]) != grid["ncells"]:
```
chrishavlin, line 131: don't want a walrus here too??? :)

yt/utilities/sdf.py
```diff
@@ -340,7 +340,7 @@ def write(self, filename):
                 k, v = var[0], _rev_types[var[1]]
                 to_write.append(k)
                 f.write(f"\t{v} {k};\n")
-            f.write("}[%i];\n" % s.size)
+            f.write(f"}}[{s.size}];\n")
```
chrishavlin, line 343: my brain is going fuzzy at this point... but pretty sure there's an extra } ? 
```suggestion
            f.write(f"}[{s.size}];\n")
```

neutrinoceros: This is intended ! That how curly braces are escaped in fstrings

chrishavlin: ah! ok, great :)

yt/utilities/grid_data_format/conversion/conversion_athena.py
neutrinoceros: if I were to write this from scratch maybe, but here I like that it's the exact same conditions than 3 lines before.

neutrinoceros:
@chrishavlin apparently I had completly forgotten about this, and ruff 0.9's release reminded me of it. I rebased the branch and took your latest suggestions into account in the most recent commit. Hopefully we can merge this soon, sorry again for falling behind !

chrishavlin (approved):
apologies for taking a while to get back to this! but it all looks good to me now.

neutrinoceros:
thanks ! let's see if this unblocks pre-commit auto upgrades then

PR #5077, opened 2024-12-06
[FIX] Correctly normalize off-axis projections for octree datasets

### LINKED ISSUE #5071: [BUG] Off-axis projections and on-axis projections differ for oct-based datasets
### Bug report

**Bug summary**

*Copied from #5059*
OK - I've actually uncovered a funny one. 
```python
import numpy as np
import yt
ds = yt.load("output_00080")
p1 = yt.ProjectionPlot(
    ds,
    "x",
    ("gas", "density"),
    center=[0.5] * 3,
    width=.8,
    #weight_field=("gas", "density"),
)
p2 = yt.OffAxisProjectionPlot(
    ds,
    [1, 0, 0],
    ("gas", "density"),
    center=[0.5] * 3,
    width=.8,
    #weight_field=("gas", "density"),
)

# Note: due to our implementation, the off-axis projection will have a
# slightly blurred cell edges so we can't do an exact comparison
v1, v2 = p1.frb["gas", "density"], p2.frb["gas", "density"]

fig = plt.figure()
plt.imshow(v1, origin="lower", norm=plt.matplotlib.colors.LogNorm())
plt.colorbar()
fig.savefig("/tmp/1.png")

fig = plt.figure()
plt.imshow(v2, origin="lower", norm=plt.matplotlib.colors.LogNorm())
plt.colorbar()
fig.savefig("/tmp/2.png")

fig = plt.figure()
plt.imshow(v1 / v2, origin="lower")
plt.colorbar()
fig.savefig("/tmp/3.png")
print(np.median(v1 / v2), np.mean(v1 / v2), np.std(v1 / v2))
```

This returns:
| Projection | Off-axis projection | Difference |
|---|---|---|
||||

The funny thing is that the ratio between the on-axis and the off-axis projection is equal (within some uncertainties) to `plot_width * 0.5675` (above, `plot_width = 0.8`). This is the case for a RAMSES dataset (here), but the same can be obtained using `fake_random_ds` instead, yielding the same value.

As much as I see how I could have missed a `plot_width` factor, what the heck is `0.5675`?

**Expected outcome**

Both the projection and off-axis projection should be the same map (up to small differences).
This bug was most likely introduced in #4741.

**Version Information**
  * yt version: 4.4 and above

cphyc:
## PR Summary

Fixes #5071.

## PR Checklist

### DIFF
--- a/yt/visualization/tests/test_offaxisprojection.py
+++ b/yt/visualization/tests/test_offaxisprojection.py
@@ -239,7 +239,7 @@ def test_off_axis_octree():
         v1, v2 = p1.frb["gas", "density"], p2.frb["gas", "density"]
         diff = (v1 - v2) / (v1 + v2) * 2
 
-        # Make sure the difference has a small bis
+        # Make sure the difference has a small bias
         assert np.mean(diff).max() < 1e-3  # 0.1%
 
         # Compute 10-90% percentile
--- a/yt/visualization/tests/test_offaxisprojection.py
+++ b/yt/visualization/tests/test_offaxisprojection.py
@@ -212,32 +212,40 @@ def test_field_cut_off_axis_octree():
 
 
 def test_off_axis_octree():
+    np.random.seed(12345)
     ds = fake_octree_ds()
-    p1 = ProjectionPlot(
-        ds,
-        "x",
-        ("gas", "density"),
-        center=[0.6] * 3,
-        width=0.8,
-        weight_field=("gas", "density"),
-    )
-    p2 = OffAxisProjectionPlot(
-        ds,
-        [1, 0, 0],
-        ("gas", "density"),
-        center=[0.6] * 3,
-        width=0.8,
-        weight_field=("gas", "density"),
-    )
+    center = [0.4, 0.4, 0.4]
 
-    # Note: due to our implementation, the off-axis projection will have a
-    # slightly blurred cell edges so we can't do an exact comparison
-    v1, v2 = p1.frb["gas", "density"], p2.frb["gas", "density"]
-    diff = (v1 - v2) / (v1 + v2) * 2
+    for weight in [("gas", "cell_mass"), None, ("index", "dx")]:
+        p1 = ProjectionPlot(
+            ds,
+            "x",
+            ("gas", "density"),
+            center=center,
+            width=0.8,
+            weight_field=weight,
+        )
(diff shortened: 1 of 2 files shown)


cphyc:
With the code from #5071 (adapted so that the same colorbar is used):
| On-axis | Off-axis | Relative difference |
|---|--|--|
|  |   |  |

The results are now unbiased (within a few percent) and show only minor deviations (at cell edges, as one would expect due to the projection strategy).

neutrinoceros:
@cphyc we haven't been backporting anything to the "stable" branch for years at this point, and backport- labels are not necessary with proper milestoning :)

cphyc:
Oopsy sorry! I keep on forgetting how the labels work :)

yt/visualization/volume_rendering/off_axis_projection.py
```diff
@@ -483,6 +485,10 @@ def temp_weightfield(field, data):
             Nx=resolution[0],
             Ny=resolution[1],
         )
+        # Note: since dx was divided by wmax, we need to rescale by it
+        projected_weighted_qty *= wmax.d / np.sqrt(3)
+        projected_weight *= wmax.d / np.sqrt(3)
```
chrishavlin, line 490: ah! that explains the mostly constant factor that values were off by! nice catch.

yt/visualization/tests/test_offaxisprojection.py
```diff
@@ -212,32 +212,40 @@ def test_field_cut_off_axis_octree():
...
+        # Note: due to our implementation, the off-axis projection will have a
+        # slightly blurred cell edges so we can't do an exact comparison
+        v1, v2 = p1.frb["gas", "density"], p2.frb["gas", "density"]
+        diff = (v1 - v2) / (v1 + v2) * 2
 
-    # Note: due to our implementation, the off-axis projection will have a
-    # slightly blurred cell edges so we can't do an exact comparison
-    v1, v2 = p1.frb["gas", "density"], p2.frb["gas", "density"]
-    diff = (v1 - v2) / (v1 + v2) * 2
+        # Make sure the difference has a small bis
```
chrishavlin, line 242: ?
```suggestion
        # Make sure the difference has a small bias
```

chrishavlin (approved):
nice! glad this turned out to be a relatively small fix!

chrishavlin (approved):

PR #5081, opened 2024-12-12
BUG: calculate default depth for off axis projections

### LINKED ISSUE #5080: BUG: Broken sample outputs of off-axis projection plots in docs
### Bug report

**Bug summary**

The outputs of two samples are broken in the docs
https://yt-project.org/docs/dev/cookbook/simple_plots.html#off-axis-projection
https://yt-project.org/docs/dev/visualizing/plots.html#off-axis-projections

The same code produces the following outputs when run with yt-4.3.1
```python
import yt

L = [1, 1, 0]  # vector normal to cutting plane
north_vector = [-1, 1, 0]
prj = yt.ProjectionPlot(
    ds, L, ("gas", "density"), width=(25, "kpc"), north_vector=north_vector
)
prj.save()
```
galaxy0030_OffAxisProjection_density

```python
import yt

# Load the dataset.

# Create a 15 kpc radius sphere, centered on the center of the sim volume
sp = ds.sphere("center", (15.0, "kpc"))

# Get the angular momentum vector for the sphere.
L = sp.quantities.angular_momentum_vector()

print(f"Angular momentum vector: {L}")

# Create an off-axis ProjectionPlot of density centered on the object with the L
# vector as its normal and a width of 25 kpc on a side
p = yt.ProjectionPlot(
    ds, L, fields=("gas", "density"), center=sp.center, width=(25, "kpc")
)
p.save()

```
galaxy0030_OffAxisProjection_density

chrishavlin:
Close #5080 

Pretty sure this won't break the SPH.... we'll see how the tests go. but would be good to have confirmation from @nastasha-w

### DIFF
--- a/yt/visualization/fixed_resolution.py
+++ b/yt/visualization/fixed_resolution.py
@@ -639,7 +639,6 @@ class OffAxisProjectionFixedResolutionBuffer(FixedResolutionBuffer):
                 self.bounds[5] - self.bounds[4],
             )
         )
-        depth = dd.depth[0] if dd.depth is not None else None
         buff = off_axis_projection(
             dd.dd,
             dd.center,
@@ -652,7 +651,7 @@ class OffAxisProjectionFixedResolutionBuffer(FixedResolutionBuffer):
             no_ghost=dd.no_ghost,
             interpolated=dd.interpolated,
             north_vector=dd.north_vector,
-            depth=depth,
+            depth=dd.depth,
             method=dd.method,
         )
         if self.data_source.moment == 2:
--- a/yt/visualization/plot_window.py
+++ b/yt/visualization/plot_window.py
@@ -83,7 +83,11 @@ def get_window_parameters(axis, center, width, ds):
 
 
 def get_oblique_window_parameters(
-    normal, center, width, ds, depth=None, get3bounds=False
+    normal,
+    center,
+    width,
+    ds,
+    depth=None,
 ):
     center, display_center = ds.coordinates.sanitize_center(center, axis=None)
     width = ds.coordinates.sanitize_width(normal, width, depth)
@@ -101,15 +105,7 @@ def get_oblique_window_parameters(
 
     w = tuple(el.in_units("code_length") for el in width)
     bounds = tuple(((2 * (i % 2)) - 1) * w[i // 2] / 2 for i in range(len(w) * 2))
-    if get3bounds and depth is None:
-        # off-axis projection, depth not specified
-        # -> set 'large enough' depth using half the box diagonal + margin
-        d2 = ds.domain_width[0].in_units("code_length") ** 2
-        d2 += ds.domain_width[1].in_units("code_length") ** 2
-        d2 += ds.domain_width[2].in_units("code_length") ** 2
-        diag = np.sqrt(d2)
-        bounds = bounds + (-0.51 * diag, 0.51 * diag)
-    return (bounds, center)
+    return bounds, center
 
 
 def get_axes_unit(width, ds):
@@ -2389,7 +2385,8 @@ class OffAxisProjectionPlot(ProjectionPlot, PWViewerMPL):
     depth : A tuple or a float
         A tuple containing the depth to project through and the string
         key of the unit: (width, 'unit'). If set to a float, code units
-        are assumed
+        are assumed. In not set, then a depth equal to the diagonal of
+        the domain width plus a small margin will be used.
     weight_field : string
         The name of the weighting field.  Set to None for no weight.
     max_level: int
@@ -2466,6 +2463,13 @@ class OffAxisProjectionPlot(ProjectionPlot, PWViewerMPL):
                 "currently supported geometries:"
                 f" {self._supported_geometries!r}"
             )
+
+        if depth is None:
+            # off-axis projection, depth not specified
+            # -> set 'large enough' depth using half the box diagonal + margin
+            depth = np.sqrt(np.sum(ds.domain_width.in_units("code_length") ** 2)) * 1.02
+        depth = ds.coordinates.sanitize_depth(depth)[0]
+
         # center_rot normalizes the center to (0,0),
         # units match bounds
         # for SPH data, we want to input the original center
@@ -2479,7 +2483,6 @@ class OffAxisProjectionPlot(ProjectionPlot, PWViewerMPL):
             width,
             ds,
             depth=depth,
-            get3bounds=True,
         )
         # will probably fail if you try to project an SPH and non-SPH
         # field in a single call


yt/visualization/plot_window.py
```diff
@@ -101,15 +105,7 @@ def get_oblique_window_parameters(
 
     w = tuple(el.in_units("code_length") for el in width)
     bounds = tuple(((2 * (i % 2)) - 1) * w[i // 2] / 2 for i in range(len(w) * 2))
-    if get3bounds and depth is None:
```
chrishavlin, line 104: if we initially calculate a depth equal to the diagonal + a bit, we don't need this extra logic here as the off-axis projection calls to this function will always have a depth.

yt/visualization/fixed_resolution.py
```diff
@@ -639,7 +639,6 @@ def _generate_image_and_mask(self, item) -> None:
                 self.bounds[5] - self.bounds[4],
             )
         )
-        depth = dd.depth[0] if dd.depth is not None else None
```
chrishavlin, line 642: removing this because the depth sanitization before `dd` creation takes care of this now.

yt/visualization/plot_window.py
```diff
@@ -2466,6 +2463,13 @@ def __init__(
                 "currently supported geometries:"
                 f" {self._supported_geometries!r}"
             )
+
+        if depth is None:
+            # off-axis projection, depth not specified
+            # -> set 'large enough' depth using half the box diagonal + margin
+            depth = np.sqrt(np.sum(ds.domain_width.in_units("code_length") ** 2)) * 1.02
```
chrishavlin, line 2470:
```suggestion
            depth = np.linalg.norm(ds.domain_width.in_units("code_length")) * 1.02
```
going to let tests run before making this change

yt/visualization/plot_window.py
```diff
@@ -2466,6 +2463,13 @@ def __init__(
                 "currently supported geometries:"
                 f" {self._supported_geometries!r}"
             )
+
+        if depth is None:
+            # off-axis projection, depth not specified
+            # -> set 'large enough' depth using half the box diagonal + margin
+            depth = np.linalg.norm(ds.domain_width.in_units("code_length")) * 1.02
```
matthewturk, line 2470: Why 1.02?  Does it need to be an epsilon that large?

nastasha-w: At least for SPH data, periodicities are applied before rotation, and are never taken to be periodic in the rotated coordinates. Therefore, the large margin *shouldn't* cause trouble, but a smaller value ought to be fine as well.

nastasha-w:
This seems to make sense overall; setting a default depth somewhere should be enough, although TBH it's probably not a great idea in most situations to project a whole box off-axis; in practice that means the depth will vary across the image. It's been a while since I looked into the code and checked what's called when here, but this does seem like something that would break the tests for SPH if done wrong. (I do have a few in there that test if at least roughly the right depth is used.)

nastasha-w:
Thanks for looking into this! Should we maybe add an SPH and a grid code off-axis image test to look out for this sort of thing going forward?

chrishavlin:
>  although TBH it's probably not a great idea in most situations to project a whole box off-axis; in practice that means the depth will vary across the image.

i don't disagree necessarily, but that's what was happening implicitly here:

https://github.com/yt-project/yt/blob/e6e5f1ffa90d42eaaf8e60cbfd00cb130ab5bdce/yt/visualization/plot_window.py#L107-L111

So IMO it's clearer to set that default depth explicitly on initialization of the plot window (whether the default depth is the diagonal of the whole domain or something smaller). 

> At least for SPH data, periodicities are applied before rotation, and are never taken to be periodic in the rotated coordinates. Therefore, the large margin shouldn't cause trouble, but a smaller value ought to be fine as well.

I'm not positive on if that's the case for gridded datasets, I'll take a closer look to be sure.

> Should we maybe add an SPH and a grid code off-axis image test to look out for this sort of thing going forward?

ya, definitely -- I'm very surprised there wasn't already an off-axis image test for gridded data to catch this initially...

nastasha-w:
>So IMO it's clearer to set that default depth explicitly on initialization of the plot window (whether the default depth is the diagonal of the whole domain or something smaller).

Makes sense! I mostly used some of those more complicated things because I wanted to *avoid* breaking stuff for e.g., grid code projections. I very much appreciate a more logical/simplified approach.

> ya, definitely -- I'm very surprised there wasn't already an off-axis image test for gridded data to catch this initially...

There aren't for SPH either -- I have some off-axis tests for those, but those are for fake datasets with a small number of particles. I did do a quick visual test though; I'm pasting the code below. It's for off-axis projections and slices through an SPH dataset. I'm guessing something similar would work for other hydro methods.
(I think the standard image tests were meant more as tests of the different frontends than of the different visualizations.)

```python
# looks good!
def visual_check_offaxis_projection():
    '''
    Plot projections along the x, z axes using the axis-aligned
    methods, then check against just-off x, z and intermediate
    lines of sight to see if we get a nice, continuous rotation.
    Note that paths here are system-specific.
    '''
    simpath = '/Users/nastasha/code/ytdev_misc/sample_data/gizmo_mhd_mwdisk/'
    simfile = 'gizmo_mhd_mwdisk.hdf5'
    outdir = '/Users/nastasha/code/ytdev_misc/testimgs/'
    outfilen_temp = 'sph_projection_test_lineofsight_{projdir}.pdf'

    nsteps = 9
    projdirs = ['x', 'z'] + [f'step{i}' for i in range(nsteps)]
    epsilon = 1e-4
    north_vector = [0., 1., 0.] #y-axis
    
    ds = yt.load(simpath + simfile)
    source = ds.all_data()
    center = [0., 0., 0.]

    for projdir in projdirs:
        if projdir in ['x', 'y', 'z']:
            outfilen = outdir + outfilen_temp.format(projdir=projdir)
            projaxis = projdir
            title = f'line of sight along the {projdir}-axis'
            #dle = ds.domain_left_edge.to("kpc")
            center = ds.arr(center, 'kpc')
            le = center - ds.quan(30., 'kpc')
            re = center + ds.quan(30, 'kpc')
            reg = YTRegion(center.to('code_length'), le.to('code_length'),
                           re.to('code_length'), ds=ds)
            # need to specify width *and* region because some giant SPH
            # particle intersects the central region
            prj = yt.ProjectionPlot(ds, projaxis, ("gas", "density"),
                                    weight_field=None,
                                    buff_size=(400, 400),
                                    center=center,
                                    width=(60, "kpc"),
                                    data_source=reg)
        else:
            _projdir = projdir + f'of{nsteps}'
            outfilen = outdir + outfilen_temp.format(projdir=_projdir)
            stepnum = int(projdir[4:])
            angle = stepnum / (nsteps - 1.) * 0.5 * np.pi
            projaxis = [np.cos(angle + epsilon), 0., np.sin(angle + epsilon)]
            title = f'line of sight x- to z-axis: step {stepnum} / {nsteps}'
            prj = yt.ProjectionPlot(ds, projaxis, ("gas", "density"),
                                    width=(60, "kpc"),
                                    weight_field=None,
                                    buff_size=(400, 400),
                                    center=center,
                                    data_source=source,
                                    north_vector=north_vector,
                                    depth=(60, 'kpc'))
        
        prj.annotate_title(title)
        prj.save(outfilen)

# looks ok, seems consisitent with projections
def visual_check_offaxis_slice():
    '''
    Plot projections along the x, z axes using the axis-aligned
    methods, then check against just-off x, z and intermediate
    lines of sight to see if we get a nice, continuous rotation.
    Note that paths here are system-specific.
    '''
    simpath = '/Users/nastasha/code/ytdev_misc/sample_data/gizmo_mhd_mwdisk/'
    simfile = 'gizmo_mhd_mwdisk.hdf5'
    outdir = '/Users/nastasha/code/ytdev_misc/testimgs/'
    outfilen_temp = 'sph_slice_test_lineofsight_{projdir}.pdf'
    # bounding box recommended at loading
    simbounds = [[-335.41967773, -305.92471313, -304.52108765],
                 [283.54318237, 300.88858032, 379.63751221]]

    nsteps = 9
    projdirs = ['x', 'z'] + [f'step{i}' for i in range(nsteps)]
    epsilon = 1e-4
    north_vector = [0., 1., 0.] #y-axis
    
    ds = yt.load(simpath + simfile, bounding_box=simbounds)
    center = ([0., 0., 0.], 'kpc')
    width = (60., "kpc")

    for projdir in projdirs:
        if projdir in ['x', 'y', 'z']:
            outfilen = outdir + outfilen_temp.format(projdir=projdir)
            projaxis = projdir
            title = f'line of sight along the {projdir}-axis'
            slc = yt.SlicePlot(ds, projaxis, ("gas", "density"),
                               center=center, width=width,
                               buff_size=(400, 400),)
        else:
            _projdir = projdir + f'of{nsteps}'
            outfilen = outdir + outfilen_temp.format(projdir=_projdir)
            stepnum = int(projdir[4:])
            angle = stepnum / (nsteps - 1.) * 0.5 * np.pi
            projaxis = [np.cos(angle + epsilon), 0., np.sin(angle + epsilon)]
            title = f'line of sight x- to z-axis: step {stepnum} / {nsteps}'
            slc = yt.SlicePlot(ds, projaxis, ("gas", "density"),
                               center=center, width=width,
                               buff_size=(400, 400), north_vector=north_vector)
        slc.annotate_title(title)
        slc.save(outfilen)


if __name__ == '__main__':
    visual_check_offaxis_projection()
    visual_check_offaxis_slice()
```

chrishavlin:
Thanks for the example tests! I think I'll open an issue for adding off axis projections tests so we can get this bug fix in a bit faster. I think I'd also like to address the magnitude of the margin in a follow up as well, @matthewturk how do you feel about that?

matthewturk:
sounds good to me

chrishavlin:
Ok, so I think that this one is good to go then, but needs an official approval from someone (@matthewturk @nastasha-w or anyone)

yt/visualization/plot_window.py
nastasha-w: I guess I also figured that the half box diagonal would be a very large margin for most line of sight directions anyway

chrishavlin:
Just added a test for gridded datasets at least, and I think this one should be good to go. 

In both the new test and a scratch notebook ([here](https://github.com/chrishavlin/yt_scratch/blob/master/notebooks/dev/offaxisproj_depth_dep.ipynb)), I integrated the (index, ones) field in an off axis projection centered on the left domain edge of the cube and looking at the right edge and the integrated distance correctly maxes out to the expected value (equal to the diagonal of the cube) when the depth keyword is set to a large value. At smaller value, the integrated distance scales as expected.

chrishavlin:
aaaand just for good measure, added a simple sph test. After which i reduced the margin epsilon.

yt/visualization/plot_window.py
chrishavlin: Reduced that epsilon margin based on testing: default depth is now the box diagonal * 1.0001  (epsilon of .01%). Probably could go even smaller? But this seems reasonable to me.

chrishavlin:
oh interesting, i assumed `contourpy` would be available for all the tests since it's a matplotlib dependency. will have to modify that test...

neutrinoceros:
`contourpy` became a dependency to `matplotlib` in [v3.6.0](https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.6.0.html#improved-quad-contour-calculations-via-contourpy), and we currently support 3.5.0

Now, if we were closely following [SPEC 0](https://scientific-python.org/specs/spec-0000/), we would be dropping support for 3.7.0 very soon, so I think we could just bump our minimal requirement to 3.6.0 and avoid additional complexity (soon to be obsolete) on your test.

neutrinoceros:
On the other hand, bumping mpl means we can't backport this branch, so maybe I'll just do it as a clean up.

chrishavlin:
ok, sounds good! i did add a skip here if contourpy is not available, so the test here will get picked up in the minimal environment once yt's minimum matplotlib gets bumped (and IMO it's fine to be skipping this test in the minimal environment anyway...).

neutrinoceros:
on second thought, maybe we should simply add contourpy as a test dependency

chrishavlin:
sure, either way :)

chrishavlin:
> maybe we should simply add contourpy as a test dependency

hey @neutrinoceros , is that your preferred solution? It does make some sense to me since this test does in fact make it a test dependency :) but wanted to double check before making the change.

neutrinoceros:
yes, I like this solution more, let's run with it !

yt/visualization/tests/test_offaxisprojection_pytestonly.py
```diff
@@ -165,3 +169,77 @@ def makemasses(i, j, k):
     # print("expected:\n", expected_out)
     # print("recovered:\n", img.v)
     assert_rel_equal(expected_out, img.v, 4)
+
+
+@pytest.fixture
+def fame_amr_ds_for_proj():
+    return fake_amr_ds()
```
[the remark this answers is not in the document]
chrishavlin: oh! I actually think I was planning to write these tests differently where a fixture would have been more useful and then forgot to remove it. I guess as-is it might save a negligible amount of time on the dataset instantiation? but I think it's clearer if I just move it back into the test itself, thanks for catching it.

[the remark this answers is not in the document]
chrishavlin: removed!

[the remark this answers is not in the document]
neutrinoceros: > it. I guess as-is it might save a negligible amount of time on the dataset instantiation

I don't think it would. Rather, it moves the instantiation to collection time, but overall it saves nothing. Anyway, thanks for removing it !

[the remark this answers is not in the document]
chrishavlin: but wouldn't it instantiate once (at collection time) and then re-use the ds rather than instantiating one for every set of test parameters? Or is that not how fixtures work? Mainly asking to make sure I understand it for other cases where it might actually matter.

nastasha-w (approved):

PR #5084, opened 2024-12-16
use .render instead of ._setup_plots in docs

xshaokun:
## PR Summary
Since we already have the public API `.render`, it’s recommended for users to rely on it rather than using the private method `._setup_plots` in the docs.


neutrinoceros:
Thanks ! for reference, this is a direct follow up to #4121

neutrinoceros (approved):

neutrinoceros:
The one failed test is completely unrelated and fixed in #5083

Issue #5087, opened 2024-12-19
PhasePlot set_title() fails on ambiguous fields even when fully specified


chummels:
### Bug report

**Bug summary**

Plot annotation `set_title()` fails when the associated `PhasePlot` has a field which may be ambiguous, even if you specify the full field name.

**Code for reproduction**

```python
import yt
ds = yt.load_sample('FIRE_M12i_ref11')
p = yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), [('gas', 'z')], weight_field=None)
p.annotate_title('test')
p.save()
```
Notice that the `z` field is fully specified as `('gas', 'z')`, but yt gets unhappy because there are other `z` fields.

**Actual outcome**

```
Traceback (most recent call last):
  File "/Users/chummels/scratch/test_phase.py", line 4, in <module>
    p.annotate_title('test')
  File "/Users/chummels/src/yt/yt/visualization/_commons.py", line 112, in newfunc
    retv = f(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chummels/src/yt/yt/visualization/profile_plotter.py", line 1324, in annotate_title
    self.plot_title[self.data_source._determine_fields(f)[0]] = title
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chummels/src/yt/yt/data_objects/data_containers.py", line 1471, in _determine_fields
    finfo = self.ds._get_field_info(field)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chummels/src/yt/yt/data_objects/static_output.py", line 1001, in _get_field_info
    raise ValueError(
ValueError: The requested field name 'z' is ambiguous and corresponds to any one of the following field types:
 ['index', 'gas', 'PartType0']
Please specify the requested field as an explicit tuple (<ftype>, <fname>).
```

**Expected outcome**

It should just work and generate a PhasePlot.

PR #5088, opened 2024-12-19
BUG: Removing code that removes field type in set_title() 

chummels:
## PR Summary

Removing code that removes field type in set_title().  Not sure why this was ever here.  It is not present for any other plot annotations.

This resolves Issue #5087 .

### DIFF
--- a/yt/visualization/profile_plotter.py
+++ b/yt/visualization/profile_plotter.py
@@ -1319,8 +1319,6 @@ class PhasePlot(ImagePlotContainer):
 
         """
         for f in self._profile.field_data:
-            if isinstance(f, tuple):
-                f = f[1]
             self.plot_title[self.data_source._determine_fields(f)[0]] = title
         return self
 
--- a/yt/visualization/tests/test_profile_plots.py
+++ b/yt/visualization/tests/test_profile_plots.py
@@ -2,11 +2,12 @@ import os
 import shutil
 import tempfile
 import unittest
+import numpy as np
 
 import pytest
 
 import yt
-from yt.testing import assert_allclose_units, fake_random_ds
+from yt.testing import assert_allclose_units, fake_random_ds, fake_random_sph_ds
 from yt.visualization.api import PhasePlot
 
 
@@ -64,6 +65,63 @@ class TestPhasePlotAPI:
         p.set_ylim(1e-2, 1e0)
         p.render()
         return p.plots["gas", "mass"].figure
+
+class TestPhasePlotParticleAPI:
+    @classmethod
+    def setup_class(cls):
+        bbox = np.array([[-1.0, 3.0], [1.0, 5.2], [-1.0, 3.0]])
+        cls.ds = fake_random_sph_ds(
+            50, bbox
+        )
+
+    def get_plot(self):
+        return PhasePlot(
+            self.ds, ("gas", "density"), ("gas", "density"), ("gas", "mass")
+        )
+
+    @pytest.mark.parametrize("kwargs", [{}, {"color": "b"}])
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_annotate_text(self, kwargs):
+        p = self.get_plot()
+        p.annotate_text(1e-4, 1e-2, "Test text annotation", **kwargs)
+        p.render()
+        return p.plots["gas", "mass"].figure
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_title(self):
+        p = self.get_plot()
+        p.annotate_title("Test Title")
+        p.render()
+        return p.plots["gas", "mass"].figure
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_log(self):
+        p = self.get_plot()
+        p.set_log(("gas", "mass"), False)
+        p.render()
+        return p.plots["gas", "mass"].figure
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_unit(self):
+        p = self.get_plot()
+        p.set_unit(("gas", "mass"), "Msun")
+        p.render()
+        return p.plots["gas", "mass"].figure
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_xlim(self):
+        p = self.get_plot()
+        p.set_xlim(1e-3, 1e0)
+        p.render()
+        return p.plots["gas", "mass"].figure
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_ylim(self):
+        p = self.get_plot()
+        p.set_ylim(1e-2, 1e0)
+        p.render()
+        return p.plots["gas", "mass"].figure
+
 
 
 def test_set_units():
--- a/yt/visualization/tests/test_profile_plots.py
+++ b/yt/visualization/tests/test_profile_plots.py
@@ -3,11 +3,9 @@ import shutil
 import tempfile
 import unittest
 
-import numpy as np
 import pytest
 
 import yt
-from yt.testing import assert_allclose_units, fake_random_ds, fake_random_sph_ds
 from yt.visualization.api import PhasePlot
 
 
--- a/yt/visualization/tests/test_profile_plots.py
+++ b/yt/visualization/tests/test_profile_plots.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import tempfile
 import unittest
+import numpy as np
 
 import pytest
 
--- a/yt/visualization/tests/test_profile_plots.py
+++ b/yt/visualization/tests/test_profile_plots.py
@@ -2,8 +2,8 @@ import os
 import shutil
 import tempfile
 import unittest
-import numpy as np
 
+import numpy as np
 import pytest
 
 import yt
@@ -66,13 +66,12 @@ class TestPhasePlotAPI:
         p.render()
         return p.plots["gas", "mass"].figure
 
+
 class TestPhasePlotParticleAPI:
     @classmethod
     def setup_class(cls):
         bbox = np.array([[-1.0, 3.0], [1.0, 5.2], [-1.0, 3.0]])
-        cls.ds = fake_random_sph_ds(
-            50, bbox
-        )
+        cls.ds = fake_random_sph_ds(50, bbox)
 
     def get_plot(self):
         return PhasePlot(
@@ -121,7 +120,6 @@ class TestPhasePlotParticleAPI:
         p.set_ylim(1e-2, 1e0)
         p.render()
         return p.plots["gas", "mass"].figure
-
 
 
 def test_set_units():
--- a/yt/visualization/tests/test_profile_plots.py
+++ b/yt/visualization/tests/test_profile_plots.py
@@ -84,42 +84,36 @@ class TestPhasePlotParticleAPI:
         p = self.get_plot()
         p.annotate_text(1e-4, 1e-2, "Test text annotation", **kwargs)
         p.render()
-        return p.plots["gas", "mass"].figure
 
     @pytest.mark.mpl_image_compare
     def test_phaseplot_set_title(self):
         p = self.get_plot()
         p.annotate_title("Test Title")
         p.render()
-        return p.plots["gas", "mass"].figure
 
     @pytest.mark.mpl_image_compare
     def test_phaseplot_set_log(self):
         p = self.get_plot()
         p.set_log(("gas", "mass"), False)
         p.render()
-        return p.plots["gas", "mass"].figure
 
     @pytest.mark.mpl_image_compare
     def test_phaseplot_set_unit(self):
         p = self.get_plot()
         p.set_unit(("gas", "mass"), "Msun")
         p.render()
-        return p.plots["gas", "mass"].figure
 
     @pytest.mark.mpl_image_compare
     def test_phaseplot_set_xlim(self):
         p = self.get_plot()
         p.set_xlim(1e-3, 1e0)
         p.render()
-        return p.plots["gas", "mass"].figure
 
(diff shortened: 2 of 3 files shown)


neutrinoceros:
This looks like a breaking change to me, the existing logic looks very much intentional (albeit flawed). Other plot annotations don't need to display a field name so I'm not surprised that it's the only one doing this.

chummels:
I added a test for this.  Previously PhasePlot didn't have any SPH-based tests.  Now it does, and it fails for dev tip but succeeds for this PR.  I don't fully understand why this doesn't raise errors for grid-based datasets, but it doesn't seem to.

chummels:
pre-commit.ci autofix

chrishavlin (dismissed):
looks good to me, and i agree with @chummels it's not a breaking change. It's only changing the key that's used to store the desired plot title.

chrishavlin:
I'll give @neutrinoceros a chance to respond, but will plan to merge later today

yt/visualization/profile_plotter.py
```diff
@@ -1319,8 +1319,6 @@ def annotate_title(self, title):
 
         """
         for f in self._profile.field_data:
-            if isinstance(f, tuple):
-                f = f[1]
```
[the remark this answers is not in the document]
chummels: Yeah, I don't really understand why this was here.  But I figured it was in such a niche corner of the code that no one had run into it except me.  I don't think people often use the `set_title()` functionality, let alone for profiles, and specifically with particle-based datasets.

neutrinoceros:
please give me the night ! I don't remember the details but I do want to have another look tomorrow

chrishavlin:
> please give me the night ! I don't remember the details but I do want to have another look tomorrow

Sure thing! I can also wait a bit longer than tomorrow if you need more time.

neutrinoceros:
Ok I'm willing to accept the change as is, *but* we need to make sure that the image tests are run. As long as the baseline submodule isn't updated, CI is expected to *fail*. The fact that it's passing here indicates that they are actually not picked up at all.

neutrinoceros:
Got it: it's because the module they were added to is explicitly ignored for pytest
https://github.com/yt-project/yt/blob/712b724423a308bf69913f5f83e9f567048ee9a9/pyproject.toml#L384

we need to move the test to a different file and make sure the tests are failing before we can move to the next step, as documented here
https://yt-project.org/docs/dev/developing/testing.html#creating-and-updating-image-baselines-for-pytest-mpl-tests

chummels:
Oh.  I didn't realize there were tests that we had which were being explicitly ignored.  Why is that?  It seems like not a great solution to just not be testing the profiler tests for bugs.

To what test file would you like me to move these tests?

neutrinoceros:
> Oh. I didn't realize there were tests that we had which were being explicitly ignored. Why is that? It seems like not a great solution to just not be testing the profiler tests for bugs.

That's because we're stuck in the middle of a migration from two test frameworks (nose to pytest): pytest doesn't support `yield`-based tests and nose is not maintained anymore (and broken on all versions of Python we still support). We have many tests that can only run on one runner, but we haven't collectively invested the time needed to get rid of nosetest altogether.

> To what test file would you like me to move these tests?

`yt/visualization/tests/test_image_comp_2D_plots.py` would be a nice home to these

chummels:
OK, so I moved the tests to the other file.  Hopefully this is ready to go given that it's a 2-line fix for a simple bug.  Answer tests aren't even really needed--just a unit test.  I just copy-pasted the test from a non-SPH Profile test that included `set_title()`.

neutrinoceros:
Indeed, image test are now properly failing. Can you please make sure that the results match your expectations, and if so, open a PR to https://github.com/yt-project/yt_pytest_mpl_baseline, [as documented](https://yt-project.org/docs/dev/developing/testing.html#creating-and-updating-image-baselines-for-pytest-mpl-tests) ?

neutrinoceros:
Actually I just noticed that the failure mode isn't quite right: it should be failing with "missing baseline", but it's showing changes in *exisiting tests* instead. This is because you apparently reused test names, so the image files are colliding with previous ones. Can you make sure you're using unique test names please ?

neutrinoceros:
Looking at the images more closely, it is indeed not needed to add that many tests, in particular because none of them seems to actually check the problem you are fixing here. A single, targeted image test should be enough, but we do need *some* test for this bug, IMO.

chrishavlin:
> in particular because none of them seems to actually check the problem you are fixing here. A single, targeted image test should be enough, but we do need some test for this bug, IMO.

there's a new `test_phaseplot_set_title` that would cover it, no? And I think the motivation for the additional tests here is that there was no coverage at for the phaseplot functionality with particle data so I think it's helpful to add a bit of extra coverage beyond the fix for the actual bug here. Or are you saying you'd prefer those tests be added as unit tests just to make sure the cases run rather than image comparison tests?

chrishavlin:
the bug wasn't with `set_title(tuple)`, it was when the phaseplot included an ambiguous field (https://github.com/yt-project/yt/issues/5087):

```
import yt
p = yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), [('gas', 'z')], weight_field=None)
p.annotate_title('test')
p.save()
```

chrishavlin:
sorry, meant to explain more but got distracted -- 

in 
```
p = yt.PhasePlot(ds, ('gas', 'density'), ('gas', 'temperature'), [('gas', 'z')], weight_field=None)
p.annotate_title('test')
```
that `('gas', 'z')` becomes just `z` in the `annotate_title` on main becuase it was removing the field type within `annotate_title`. this PR doesn't touch the actual label being used (`'test'`).

e.g., here's the result of running the new test on main:

```python
from yt.testing import fake_random_sph_ds
import yt

ds = fake_random_sph_ds(1000, [[0., 1.], [0., 1.], [0., 1]])
p = yt.PhasePlot(
            ds, ("gas", "density"), ("gas", "density"), ("gas", "mass")
        )   
p.annotate_title("Test Title")
p.render()
```

```python-traceback
yt : [INFO     ] 2025-03-04 10:18:28,564 Parameters: current_time              = 0.0
yt : [INFO     ] 2025-03-04 10:18:28,564 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2025-03-04 10:18:28,564 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2025-03-04 10:18:28,565 Parameters: domain_right_edge         = [1. 1. 1.]
yt : [INFO     ] 2025-03-04 10:18:28,565 Parameters: cosmological_simulation   = 0
yt : [INFO     ] 2025-03-04 10:18:28,565 Allocating for 1000 particles
  File "/Users/chavlin/src/temp/tempthing.py", line 8, in <module>
    p.annotate_title("Test Title")
  File "/Users/chavlin/src/yt_/yt_dev/yt/yt/visualization/_commons.py", line 112, in newfunc
    retv = f(self, *args, **kwargs)
  File "/Users/chavlin/src/yt_/yt_dev/yt/yt/visualization/profile_plotter.py", line 1324, in annotate_title
    self.plot_title[self.data_source._determine_fields(f)[0]] = title
  File "/Users/chavlin/src/yt_/yt_dev/yt/yt/data_objects/data_containers.py", line 1471, in _determine_fields
    finfo = self.ds._get_field_info(field)
  File "/Users/chavlin/src/yt_/yt_dev/yt/yt/data_objects/static_output.py", line 1001, in _get_field_info
    raise ValueError(
ValueError: The requested field name 'mass' is ambiguous and corresponds to any one of the following field types:
 ['gas', 'io', 'all']
Please specify the requested field as an explicit tuple (<ftype>, <fname>).
```

chummels:
@chrishavlin is correct.  I added the other tests just to ensure particle-based PhasePlots were getting tested for other annotations, just as grid-based PhasePlots are.  I figured I might as well add them while I was updating things here, to potentially avoid other bugs down the line with more coverage.  I didn't think image tests were necessary per se, since unit tests would catch any breaking failures, as they do with this particular bug.  If I want these to just be unit tests, can I just remove the `@pytest.mark.mpl_image_compare` lines from the tests and then just get this merged?  

I'll be honest that I am not really excited about doing all the steps to generate new image test answers for all of this.  It's a two-line bugfix for a simple bug.  So whatever I can do to expedite this and not use any more of anyone's time is what I'd prefer.

chrishavlin:
> If I want these to just be unit tests, can I just remove the @pytest.mark.mpl_image_compare lines from the tests

Yup, that's all you need to do to make them unit tests, but there'd be a couple of cleanup steps I'd recommend: you'd no longer need the test functions to return anything so you could strip out the `return p.plots["gas", "mass"].figure` lines and then I'd also suggest moving the tests to a different file again (sorry!!) since `test_image_comp_2d_plots.py` implies that all the contained tests are image comparisons. and depending on the results of https://github.com/yt-project/yt/pull/5129 I'd actually recommend moving them back to where you had them in the first place (again, sorry!!).

chummels:
OK, I finally got back to this.  I moved the tests back and changed them back to unit tests.  Let me know if there are any other things to do or if this can be merged.

chrishavlin:
looks like you missed a numpy import in the test file when moving things back

chummels:
Oops!  My bad.  I've now added it.  Hopefully the tests complete OK and we can finally call this one done. :)

yt/visualization/tests/test_profile_plots.py
```diff
@@ -66,6 +67,55 @@ def test_phaseplot_set_ylim(self):
...
+        bbox = np.array([[-1.0, 3.0], [1.0, 5.2], [-1.0, 3.0]])
+        cls.ds = fake_random_sph_ds(50, bbox)
+
+    def get_plot(self):
+        return PhasePlot(
+            self.ds, ("gas", "density"), ("gas", "density"), ("gas", "mass")
+        )
+
+    @pytest.mark.parametrize("kwargs", [{}, {"color": "b"}])
+    @pytest.mark.mpl_image_compare
```
chrishavlin, line 82:
```suggestion
```

yt/visualization/tests/test_profile_plots.py
```diff
@@ -66,6 +67,55 @@ def test_phaseplot_set_ylim(self):
...
+        p.set_unit(("gas", "mass"), "Msun")
+        p.render()
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_xlim(self):
+        p = self.get_plot()
+        p.set_xlim(1e-3, 1e0)
+        p.render()
+
+    @pytest.mark.mpl_image_compare
```
chrishavlin, line 112:
```suggestion
```

yt/visualization/tests/test_profile_plots.py
```diff
@@ -66,6 +67,55 @@ def test_phaseplot_set_ylim(self):
...
+        p.set_log(("gas", "mass"), False)
+        p.render()
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_unit(self):
+        p = self.get_plot()
+        p.set_unit(("gas", "mass"), "Msun")
+        p.render()
+
+    @pytest.mark.mpl_image_compare
```
chrishavlin, line 106:
```suggestion
```

yt/visualization/tests/test_profile_plots.py
```diff
@@ -66,6 +67,55 @@ def test_phaseplot_set_ylim(self):
...
+        p.annotate_title("Test Title")
+        p.render()
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_log(self):
+        p = self.get_plot()
+        p.set_log(("gas", "mass"), False)
+        p.render()
+
+    @pytest.mark.mpl_image_compare
```
chrishavlin, line 100:
```suggestion
```

yt/visualization/tests/test_profile_plots.py
```diff
@@ -66,6 +67,55 @@ def test_phaseplot_set_ylim(self):
...
+        p.annotate_text(1e-4, 1e-2, "Test text annotation", **kwargs)
+        p.render()
+
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_set_title(self):
+        p = self.get_plot()
+        p.annotate_title("Test Title")
+        p.render()
+
+    @pytest.mark.mpl_image_compare
```
chrishavlin, line 94:
```suggestion
```

yt/visualization/tests/test_profile_plots.py
```diff
@@ -66,6 +67,55 @@ def test_phaseplot_set_ylim(self):
...
+        )
+
+    @pytest.mark.parametrize("kwargs", [{}, {"color": "b"}])
+    @pytest.mark.mpl_image_compare
+    def test_phaseplot_annotate_text(self, kwargs):
+        p = self.get_plot()
+        p.annotate_text(1e-4, 1e-2, "Test text annotation", **kwargs)
+        p.render()
+
+    @pytest.mark.mpl_image_compare
```
chrishavlin, line 88:
```suggestion
```

chrishavlin (commented):
Also have to remove all the `@pytest.mark.mpl_image_compare` decorator usages, then I think this should be good to go!

chrishavlin:
@chummels I took the liberty of removing the `@pytest.mark.mpl_image_compare` usages that I noted in the previous comment to get this across the line :) I think tests will pass now, but just a heads up you'll want to pull changes locally if you need to make any more edits.

chrishavlin:
pre-commit.ci autofix

chrishavlin (approved):

chrishavlin:
ok, I think we're good! Going to hit merge since the jenkins test is hanging indefinitely these days (a separate can o' worms).

chummels:
Thanks for getting it across the finish line, Chris!

Issue #5089, opened 2024-12-19
PhasePlot fails when a weight_field is set for a particle-based dataset


chummels:
### Bug report

**Bug summary**

PhasePlot fails when a weight_field is set for a particle-based dataset.  Note, this does not affect grid-based datasets.

**Code for reproduction**

```python
import yt
ds = yt.load_sample("FIRE_M12i_ref11")
my_sphere = ds.sphere("c", (50, "kpc"))
plot = yt.PhasePlot(my_sphere, ("gas", "density"), ("gas", "temperature"), [("gas", "H_p0_fraction")], weight_field=('gas', 'density'))
plot.save()
```

**Actual outcome**

```
Traceback (most recent call last):
  File "/Users/chummels/scratch/test_weight_field.py", line 4, in <module>
    plot = yt.PhasePlot(my_sphere, ("gas", "density"), ("gas", "temperature"), [("gas", "H_p0_fraction")], weight_field=('gas', 'density'))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chummels/src/yt/yt/visualization/profile_plotter.py", line 953, in __init__
    profile = create_profile(
              ^^^^^^^^^^^^^^^
  File "/Users/chummels/src/yt/yt/data_objects/profiles.py", line 1317, in create_profile
    is_local.append(wf.sampling_type == "local")
                    ^^^^^^^^^^^^^^^^
AttributeError: 'tuple' object has no attribute 'sampling_type'```

**Expected outcome**

The PhasePlot completing successfully and making a PhasePlot output file.

**Version Information**
yt dev tip

PR #5090, opened 2024-12-19
BUG: Properly reference weight field in profile code for particle-based datasets

chummels:
## PR Summary

This fixes Issue #5089 .  The weight field was not being properly referenced in the profiling code when a `weight_field` was specified for particle-based datasets.  This code also adds a test to the profiling code to ensure it doesn't break again.


chrishavlin (approved):
looks good to me

PR #5101, opened 2025-01-23
TYP: Fix typing in numpy2_compat.py

### LINKED ISSUE #5100: mypy failure: unused error code in numpy2_compat.py
after merging with main just now, #4996 is failing the type-check test with an unrelated failure ([link](https://github.com/yt-project/yt/actions/runs/12934059343/job/36074148625?pr=4996)):

```
yt/_maintenance/numpy2_compat.py:9: error: Unused "type: ignore" comment  [unused-ignore]
yt/_maintenance/numpy2_compat.py:9: error: Incompatible types in assignment (expression has type "Callable[[_Buffer | _SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | builtins.bool | int | float | complex | str | bytes | _NestedSequence[builtins.bool | int | float | complex | str | bytes], _Buffer | _SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | builtins.bool | int | float | complex | str | bytes | _NestedSequence[builtins.bool | int | float | complex | str | bytes] | None, float, int], generic[Any] | ndarray[tuple[int, ...], dtype[generic[Any]]]]", variable has type overloaded function)  [assignment]
yt/_maintenance/numpy2_compat.py:9: note: Error code "assignment" not covered by "type: ignore" comment
Found 2 errors in 1 file (checked 516 source files)
```

chrishavlin:
Close #5100

### DIFF
--- a/yt/_maintenance/numpy2_compat.py
+++ b/yt/_maintenance/numpy2_compat.py
@@ -6,4 +6,4 @@ if hasattr(np, "trapezoid"):
     # np.trapz is deprecated in numpy 2.0 in favor of np.trapezoid
     trapezoid = np.trapezoid
 else:
-    trapezoid = np.trapz  # type: ignore [attr-defined] # noqa: NPY201
+    trapezoid = np.trapz  # type: ignore[assignment] # noqa: NPY201


jzuhone (approved):

neutrinoceros:
This seems harmless at first sight but I don't have the faintest idea why this suddenly started to fail. Do you ?
mypy is pinned in `requirements/typecheck.txt` so that's not it, but clearly if type checking succeeds here, it means that *something* must have changed in the enviornment. Come to think of it, these comments cannot both co-exist with `warn_unused_ignores` *and* type check against both numpy 1.x and 2.x, which is also worrisome to some degree, but numpy 1 is far less type-checkable anyway so we don't loose much by not trying it.
Anyway, thanks for the ping. Unfortunately I don't understand why this is suddenly necessary, but since it clearly is, I see no reason to block it, so feel free to merge !

chrishavlin:
> This seems harmless at first sight but I don't have the faintest idea why this suddenly started to fail. Do you ?

@neutrinoceros part of the reason for the ping was the hope that you'd have an idea offhand of why this was needed and I confess I didn't look into it too deeply :) 

But looking again now I think I figured it out! np 2.2.2 was released last week and gets picked up in the type checking test here **and** has a number of typing-related changes ([Full release notes](https://github.com/numpy/numpy/releases/tag/v2.2.2)). This change https://github.com/numpy/numpy/pull/28178 would fix the previous attr-defined failure (and give rise to unused error code  failure reported in #5100 ). But just removing the ignore (which is what I first tried locally) results in a new error:

```
yt/_maintenance/numpy2_compat.py:9: error: Incompatible types in assignment (expression has type "Callable[[_Buffer | _SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | builtins.bool | int | float | complex | str | bytes | _NestedSequence[builtins.bool | int | float | complex | str | bytes], _Buffer | _SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | builtins.bool | int | float | complex | str | bytes | _NestedSequence[builtins.bool | int | float | complex | str | bytes] | None, float, int], generic[Any] | ndarray[tuple[int, ...], dtype[generic[Any]]]]", variable has type overloaded function)  [assignment]
```
which... I didn't know what to do with, so added the `ignore[assignment]` just to get tests passing again. But looking again now, maybe setting the expected type for `trapezoid` is a better fix? e.g., the following diff passes mypy locally:

```diff
 import numpy as np
+from typing import Callable
 
+trapezoid: Callable
 if hasattr(np, "trapezoid"):
     # np.trapz is deprecated in numpy 2.0 in favor of np.trapezoid
     trapezoid = np.trapezoid
 else:
-    trapezoid = np.trapz  # type: ignore[assignment] # noqa: NPY201
+    trapezoid = np.trapz  # noqa: NPY201
```

I'm just not sure if `trapezoid: Callable` is sufficient?

neutrinoceros:
Ow, that makes sense ! I knew about numpy 2.2.2 but somehow missed the connection. Welp, now there's not even a residual doubt about it; we should only aim at typechecking against the latest version of numpy at this point. Thanks for the clarification !

neutrinoceros (approved):

neutrinoceros:
@Meeseeks-dev backport to yt-4.4.x

neutrinoceros:
@meeseeksdev backport to yt-4.4.x

PR #5112, opened 2025-02-04
BUG: fix NaNs in off-center FRB slices from ds.r

yut23:
## PR Summary

FRBs of slices made with `ds.r` were selecting the wrong area when the center of the slice didn't match the center of the domain. The existing tests didn't catch this, as they only check regions that have the same center as the domain. I've added a test case that covers this.

Example code and resulting plots:
```python
import yt

region = ds.r[0:0.5:100j, 0.5:1.0:100j, 0.5]

# display NaNs as red
cmap = plt.colormaps["viridis"]
cmap.set_bad(color="red")

plt.imshow(region["index", "x"], origin="lower", vmin=0, vmax=1, cmap=cmap)
plt.colorbar()
```

before
after

## PR Checklist


chrishavlin (approved):
nice catch, the fix looks good to me.

PR #5115, opened 2025-02-11
allow amrex substantance fields to have an "_"

zingale:
this allows for NSE protons in our output ("P_nse")

## PR Checklist

### DIFF
--- a/yt/frontends/amrex/fields.py
+++ b/yt/frontends/amrex/fields.py
@@ -504,7 +504,7 @@ class MaestroFieldInfo(FieldInfoContainer):
                 )
 
 
-substance_expr_re = re.compile(r"\(([a-zA-Z][a-zA-Z0-9_]*)\)")
+substance_expr_re = re.compile(r"\(([a-zA-Z]\w*)\)")
 substance_elements_re = re.compile(r"(?P<element>[a-zA-Z_]+)(?P<digits>\d*)")
 SubstanceSpec: TypeAlias = list[tuple[str, int]]
 


yt/frontends/amrex/fields.py
```diff
@@ -504,7 +504,7 @@ def setup_fluid_fields(self):
                 )
 
 
-substance_expr_re = re.compile(r"\(([a-zA-Z][a-zA-Z0-9]*)\)")
+substance_expr_re = re.compile(r"\(([a-zA-Z][a-zA-Z0-9_]*)\)")
 substance_elements_re = re.compile(r"(?P<element>[a-zA-Z]+)(?P<digits>\d*)")
```
yut23, line 508: This should include `_` too, so it doesn't get mistaken as a molecule

yt/frontends/amrex/fields.py
yut23, line 1: The underscore needs to be escaped in the tex strings

zingale:
I'm happy with the changes

chrishavlin:
nose test failure was from an aborted run after timing out:

> Build timed out (after 240 minutes). Marking the build as aborted.

probably a fluke (since other tests ran recently no problem), will trigger it again.

chrishavlin:
@yt-fido test this please

chrishavlin:
Is this one waiting on a review?

yt/frontends/amrex/fields.py
```diff
@@ -504,8 +504,8 @@ def setup_fluid_fields(self):
                 )
 
 
-substance_expr_re = re.compile(r"\(([a-zA-Z][a-zA-Z0-9]*)\)")
-substance_elements_re = re.compile(r"(?P<element>[a-zA-Z]+)(?P<digits>\d*)")
+substance_expr_re = re.compile(r"\(([a-zA-Z][a-zA-Z0-9_]*)\)")
```
neutrinoceros, line 507: Maybe you guys prefer the verbose style, in which case don't mind this nit, but I couldn't resist pointing out that this expression is now 1-1 equivalent
```suggestion
substance_expr_re = re.compile(r"\(([a-zA-Z]\w*)\)")
```

yt/frontends/amrex/fields.py
```diff
@@ -539,18 +539,22 @@ def __str__(self) -> str:
...
     def _to_tex_molecule(self) -> str:
         return "".join(
-            rf"{element}_{{{count if count > 1 else ''}}}"
+            rf"{self._tex_escape(element)}_{{{count if count > 1 else ''}}}"
             for element, count in self._spec
         )
 
     def _to_tex_descriptive(self) -> str:
-        return str(self)
+        return self._tex_escape(str(self))
```
neutrinoceros, line 557: any reason not to update the corresponding `__str__` implementation instead ?

yt/frontends/amrex/fields.py
```diff
@@ -539,18 +539,22 @@ def __str__(self) -> str:
             f"{element}{count if count > 1 else ''}" for element, count in self._spec
         )
 
+    @staticmethod
+    def _tex_escape(s: str) -> str:
+        return s.replace("_", r"\_")
```
neutrinoceros, line 544: I can see this snowballing into a subtle bug where a single underscore would be escaped multiple times instead of once. I'm sure we make this method idempotent using a regexp:

```suggestion
        # replace '_' with '\_' once, but leave alone already-escaped occurrences
        return re.sub(r"[^\\]_",  r"\\_", s)
```

yt/frontends/amrex/tests/test_field_parsing.py
```diff
@@ -12,6 +12,7 @@
         pytest.param("X(C12H24)", [("C", 12), ("H", 24)], id="molecule_2"),
         pytest.param("X(H2O)", [("H", 2), ("O", 1)], id="molecule_3"),
         pytest.param("X(ash)", [("ash", 0)], id="descriptive_name"),
+        pytest.param("X(P_nse)", [("P_nse", 0)], id="descriptive_underscore"),
```
neutrinoceros, line 15: I did a double take here: the underscore isn't descriptive by itself, or is it ?
```suggestion
        pytest.param("X(P_nse)", [("P_nse", 0)], id="with_an_underscore"),
```

neutrinoceros (changes requested):
Looks fine, and mostly I have questions more than demands, though there's one point I'd really want to see taken into account about making escaping idempotent.

yt/frontends/amrex/tests/test_field_parsing.py
neutrinoceros: (the same applies to the other two test cases)

neutrinoceros:
I'm adding this to the 4.5.0 milestone as it's not currently labeled as a bugfix, but if you guys would like it backported to yt 4.4.x, this seems fine to me, so don't hesitate to changie the milestone accordingly and update the labels (enhancement -> bug).

chrishavlin:
@zingale looks like there are some remaining comments for you to check out here.

yt/frontends/amrex/fields.py
yut23: This regexp deletes the non-backslash character preceding an underscore. It should go in a capturing group, or just use a negative lookbehind instead

yt/frontends/amrex/tests/test_field_parsing.py
zingale: I don't understand the concerrn

yt/frontends/amrex/fields.py
zingale: this is fine with me

PR #5117, opened 2025-02-11
Classes to support AMReX HDF5 Output

kelseasouders:
## PR Summary
The HDF5 plot file output produced by PeleC (and presumably all amrex-based codes) does not work directly with the existing `ChomboDataset` class. This is most easily observable through the dimensionality - the Chombo I/O handler works with `self.dim = self._handle["Chombo_global/"].attrs["SpaceDim"]`, however `self.dim` is not correctly read as an integer when reading the amrex-produced files, so I have fixed it by explicitly setting `self.dim = int(self._handle["Chombo_global/"].attrs["SpaceDim"])`. I've made this change in a few locations in this PR.

I also observed some errors in the spatial positioning when working with the PeleC HDF5 output files, which appeared to come from the `_calc_left_edge()` and `_calc_right_edge()` methods assuming that the domain starts at [0, 0, 0]. The PeleC HDF5 output files contain the edge information directly, so I have elected to read them directly from the file (since amrex includes that in the attributes anyway), which has fixed the spatial misalignment. 

I was torn between adding these classes to the `amrex` frontend or `chombo` frontend, but decided that the `chombo` frontend would be more appropriate since the two share the same underlying functionality. I believe that it would be fairly straightforward to integrate these changes into the `ChomboDataset` class directly, but I refrained since I was unsure if they would lead to issues in other types of chombo data.

## PR Checklist

### DIFF
--- a/yt/frontends/chombo/api.py
+++ b/yt/frontends/chombo/api.py
@@ -9,6 +9,8 @@ from .data_structures import (
     Orion2Hierarchy,
     PlutoDataset,
     PlutoHierarchy,
+    AMReXHDF5Dataset,
+    AMReXHDF5Hierarchy
 )
 from .fields import (
     ChomboFieldInfo,
--- a/yt/frontends/chombo/data_structures.py
+++ b/yt/frontends/chombo/data_structures.py
@@ -916,14 +916,12 @@ class AMReXHDF5Dataset(ChomboDataset):
 
     def _calc_left_edge(self):
         fileh = self._handle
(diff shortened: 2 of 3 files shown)


chrishavlin:
thanks for the PR! 

One quick note -- this may be exposed in tests, but i think you'll need to over-ride the base `ChomboDataset._is_valid` method in your new `AMReXHDF5Dataset`, otherwise the logic used by `yt.load(...)` won't be able to distinguish between the two.

Issue #4993, opened 2024-09-14
ENH: add a 'pixel-averageing' (default) option for SPH projections


nastasha-w:
### Goal

**Summary**

The current method for generating projections from SPH datasets integrates the field to be projected along pencil-beam lines of sight through the center of each pixel in the grid. While this is a physically reasonable approach, and appropriate in some cases, averaging the quantity to be projected across the pixel (or the rectangular prism projected into that pixel) would be more consistent with the approach used for grid and AMR datasets, and would likely better match users' intuition. Therefore, as discussed in a telecon on September 13, 2024 with @matthewturk, @chummels, @neutrinoceros, and @nastasha-w, it would be a good idea to implement a pixel-averaging projection method for SPH datasets. This would be the default setting for SPH projections, and the pixel center line-of-sight method would be an option users can choose, based on what exactly there are trying to do.

**Method**

Our goal is to implement the pixel-averaging projection method described in [this paper](https://ui.adsabs.harvard.edu/abs/2021arXiv210605281B/abstract), by @JBorrow and @AshKelly. One possible improvement is that the current SPH projection back-end already has a fast way to compute the integral of the dimensionless part of the SPH kernel function at a given [impact parameter] / [smoothing length], so we could feasibly use the method in the paper to extend this to full three-dimensional integrals of the SPH kernels over the rectangular prism defining each pixel.

**TODO**
- [ ] first draft of the backend function
- [ ] add the new option to the frontend functions and pass the information through to the back-end call
- [ ] write tests for the new back-end, and add the line-of-sight integral option selection to the tests for the older method
- [ ] test and debug the new implementation
- [ ] add documentation on the two SPH projection method options to the function/method docstrings and to the narrative docs

PR #5121, opened 2025-02-19
WIP: add pixel-averaging option for SPH projections

nastasha-w:
## PR Summary

This PR will add an option (to be set as default) for SPH projections. The current implementation integrates the specified field over a line of sight through the center of each pixel in the grid the simulation is projected onto. This PR will add the option to average the specified field over the pixel area instead. The output units will be the same: input field units per unit pixel area (non-weighted), or input field units (weighted). This should close issue #4993.

The implementation is based on [Borrow & Kelly (2021)](https://ui.adsabs.harvard.edu/abs/2021arXiv210605281B/abstract). 

This PR should add this projection option for axis-aligned and off-axis projections.

## PR Checklist

 
- [ ] Adds tests: unit tests for the same cases as the old/current SPH projections.
- [ ] Adds test: mass conservation for different smoothing length and pixel size combinations.
- [ ] Qualitative test: convergence between SPH projection methods as pixel size approaches 0.

### DIFF
--- a/yt/utilities/lib/pixelization_routines.pyx
+++ b/yt/utilities/lib/pixelization_routines.pyx
@@ -1364,8 +1364,9 @@ def pixelize_sph_kernel_projection_pixelave(
     cdef np.intp_t xsize, ysize
     cdef np.float64_t x_min, x_max, y_min, y_max, z_min, z_max, prefactor_j
     cdef np.int64_t xi, yi, x0, x1, y0, y1, xxi, yyi, nx, ny, nsubx, nsuby
-    cdef np.float64_t q_ij2, posx_diff, posy_diff, ih_j2, insubx, insuby
-    cdef np.float64_t x, y, dx, dy, idx, idy, ipixA, h_j2, px, py, pz
+    cdef np.float64_t q_ij2, posx_diff, posy_diff, ih_j2
+    cdef np.float_64_t insubx, insuby, insample_min, xnorm, ynorm
+    cdef np.float64_t x, y, dx, dy, idx, idy, ipixA, subA, h_j2, px, py, pz
     cdef np.float64_t period_x = 0, period_y = 0, period_z = 0
     cdef int i, j, ii, jj, kk
     cdef np.float64_t[:] _weight_field
@@ -1406,6 +1407,22 @@ def pixelize_sph_kernel_projection_pixelave(
     if kernel_name not in kernel_tables:
         kernel_tables[kernel_name] = SPHKernelInterpolationTable(kernel_name)
     cdef SPHKernelInterpolationTable itab = kernel_tables[kernel_name]
+
+    # pre-calculate kernel 1D integral values to use for small particles
+    kernel1dint = <np.float64_t *> malloc(
+            sizeof(np.float64_t) * nsample_min * nsample_min)
+            # get line-of-sight kernel integral values
+    insample_min = 1.0 / <np.float64_t> nsample_min
+    for i in range(nsample_min):
+        xnorm = -1.0 + 2.0 * (i + 0.5) * insample_min
+        xnorm = xnorm * xnorm
+        for j in range(nsample_min):
+            ynorm = -1.0 + 2.0 * (i + 0.5) * insample_min
+            q_ij2 = xnorm + ynorm * ynorm
+            if q_ij2 >= 1.:
+                kernel1dint[nsample_min * i + j] = 0.0
+            else:
+                kernel1dint[nsample_min * i + j] = itab.interpolate(q_ij2)
     with nogil, parallel():
         # loop through every particle
         # NOTE: this loop can be quite time consuming. However it is easily
@@ -1421,6 +1438,7 @@ def pixelize_sph_kernel_projection_pixelave(
         # currently employ #1 as its workload is more even and consistent, even though it
         # comes with a price of an additional, per thread memory for storing the
         # intermediate results.
+        # !! These comments were written for  
 
         local_buff = <np.float64_t *> malloc(sizeof(np.float64_t) * xsize * ysize)
         xiterv = <np.float64_t *> malloc(sizeof(np.float64_t) * 2)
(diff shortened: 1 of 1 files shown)


nastasha-w:
pre-commit.ci autofix

nastasha-w:
pre-commit.ci autofix

chummels:
I don't yet have time to review this, but I just wanted to comment.  Thank you for this incredibly important work on getting the ProjectionPlot functionality upgraded.  I think this will have a positive impact on the work of hundreds to thousands of people.  Looking forward to discussing further soon!

nastasha-w:
pre-commit.ci autofix

nastasha-w:
pre-commit.ci autofix

nastasha-w:
pre-commit.ci autofix

nastasha-w:
pre-commit.ci autofix

nastasha-w:
I was hitting a bit of a wall with testing: the baseline and `yt` image projections looked similar, but there were differences larger than I'd typically be comfortable with (~25%). Part of the issue was likely that these were comparisons to brute-forces calculations, integrating each kernel over each output grid pixel by subsampling the pixel by a factor of 50 in each direction, and using 50 samples along the line-of-sight direction as well. This is pretty slow though, and in some cases a worse sampling rate than in the `yt` method new to this PR: pixel-averaging. 

I switched to testing against the pencil-beam in a smaller test; this approach gets me better subsampling faster, by simply projecting onto a higher-resolution grid than the testing target, then averaging the values in blocks of those pixels. Based on this testing, I think getting the pixel values right to high precision with this method is maybe just kind of hard...

For this first test, I downsampled pixel values from grids of increasing size, then calculated the total projected mass by adding up the pixel values and multiplying by the area per pixel. The x axis shows the subsampling factor along each grid axis. These projections are the ones shown in the next plot: a simple test dataset with particles of different sizes, and different placements relative to the grid.
resconv_mass_simpledata_2025-03-27
The main conclusion here is that the pixel averaging method does a pretty good job of conserving mass: results are accurate to about 1e-4 fractionally even at the lowest resolution.

However, the values of individual pixels do not seem to converge as rapidly. The image below is messy, as I tend to make them, but convergence does require many images. The top two rows show the pixel averaging method: first, projections of the same dataset, subsampled by the factors listed above, which increase from left to right. Black dashed lines indicate the extent of the seven SPH particles in the test dataset; this dataset was periodic in the x-axis direction, but not in the y-axis direction. The second row shows the difference between each subsampled image and the image at the right, with the highest subsampling factor. The labels at the top of the different images show the maximum absolute value of the relative difference (pixel value) / (pixel value, max. subsampling) - 1. The lowest-resolution images here show ~10%-level differences with the ones subsampled at higher resolution, which is not as good as I had hoped.
resconv_simpledata_2025-03-17

For comparison, the next two rows show the same convergence test, but for the pencil-beam projections. Those definitely don't do as well, with large differences at the lowest subsampling resolution. However, even at a 64x64 subsampling rate, there are still ~1% differences here. The smallest smoothing length in the test data is 0.2, so this subsampling rate corresponds to 0.4 * 64 / 1 = 26 pixels per dimension across the kernel. The minimum in the pixel-averaging method's backend is 32 pixels per dimension. 

The bottom row compares the pencil-beam and pixel-averaging projections at different subsampling resolutions. The images for 128x128 subsampling pixels are identical, because the subsampling pixels here are small enough that the smallest particles in the test dataset are resolved at the minimum level. This also holds for higher resolutions I tried.

tldr:
- I want to switch my baseline image in the tests to one that using the existing pencil-beam projections, since those run faster than my initial tests, and should do the exact same thing.
- Getting converged pixel values is hard. Using a larger subsampling resolution would probably help, but of course, this comes with a speed trade-off. It also makes testing harder, in the sense that setting wider tolerances might also let some genuine problems slip through the cracks.

@JBorrow @neutrinoceros @chummels , any advice on what sort level of convergence we want for these pixel values? As a point of reference, I think the volume integrals are limited to a 1e-4 -- 1e-5 in 1 level accuracy by the kernel line integral approximation used in both projection methods.

nastasha-w:
I still need to write the narrative docs to go with this PR, but I would appreciate if people could start taking a look at the implementation part of this PR. I've run the tests and done the further tests that I planned to for this, but the pixel values don't agree as well with those obtained from very high pixel subsampling rates as I had hoped. @JBorrow and @AshKelly, I would particularly appreciate feedback from you on whether this is a bug that needs fixing, or whether these values just converge slowly.

I'm hoping I can resolve the type checking  rebasing after #5137 is done.

chrishavlin:
One quick comment not related to the details of the implementation:

> This PR will add an option (to be set as default) 

I could be wrong, but this feels like a pretty big change to drop on folks by default in the next feature release (4.5). IMO it'd be a smoother transition to not enable this by default initially: i.e., I'd prefer that next release 4.5 introduce and advertise the functionality and then the subsequent release (4.6 or 5.0?) switch it to default.

nastasha-w:
Good point! It would indeed be better to set the default to "pencilbeam" for now.

JBorrow:
I've taken a brief look at the code and I can't immediately see if there is something wrong.

However, if you are using the subsampling strategy that we demonstrated in Borrow & Kelly (2021), there should be no 'convergence' - you should get the same answer irrespective of your pixel size. I believe there must hence be a bug in the code. You can see in our Fig 9 that for a similar particle setup to you we are able to constrain reconstruction errors to approximately less than 0.01% in all cases.

You could test this yourself against our reference implementation included in swiftsimio. You can find information on the lower level API here: https://swiftsimio.readthedocs.io/en/latest/visualisation/projection.html#lower-level-api

To use that, you would do something like:
```python
# Rescale to 0, 1... then
from swiftsimio.visualisation.projection_backends.subsampled import scatter

parameters = {
  "x": x,
  "y": y,
  "m": masses,
  "h": hsml
}

output = {
  x: scatter(**parameters, res=x) for x in (100, 200, 400, 800)
}
```

Also consider comparing against `subsampled_extreme`.

For what it's worth, I believe we use a different strategy than you to determine how to subsample pixels. We define a minimum number of kernel evaluations (one-sided), this is by default 16. So if our kernel overlaps less than 32x32 pixels, we oversample those pixels. If it overlaps 16x16 < x < 32x32, each pixel is evaluated 4 (2x2) times. For 8x8 < 16x16 we do each pixel 16 (4x4) times, and so forth: https://swiftsimio.readthedocs.io/en/latest/_modules/swiftsimio/visualisation/projection_backends/subsampled.html#scatter

nastasha-w:
@JBorrow Thanks! I was doing 1 sample per pixel if 32x32 pixels were overlapped, I'll see if a higher minimum number of evaluations works better. If not, I'll have to see if there's a deeper problem.

nastasha-w:
Note: I rebased this branch to main after Chris Havelin's type annotation fixes in the SPH projection tests were merged

nastasha-w:
pre-commit.ci autofix

PR #5129, opened 2025-03-06
TST: pytest: un-ignore test_profile_plots.py

chrishavlin:
Maybe I'm missing something... but I don't see why still we have pytest ignoring `test_profile_plots.py`. It seems that https://github.com/yt-project/yt/pull/4122 removed the nose dependence and I'm having a hard time tracking the changes that added it to the `pyproject.toml` list of files for pytest to ignore... But both the pytest unit and mpl image tests run and pass for me locally.


chrishavlin:
the `test_profile_plots.py` image tests did indeed run and pass

Screenshot 2025-03-06 at 10 16 27 AM

chrishavlin:
shortcut to one of the pytest logs showing the `test_profile_plots.py` unit tests passing: https://github.com/yt-project/yt/actions/runs/13703003053/job/38321309846?pr=5129#step:8:880

neutrinoceros (approved):
It looks like I originally added this ignore in #4344, but I honestly can't understand what was my reasoning there (I wish I gave more insight at the time...). Anyway, from where we're standing now, this just looks like a mistake on my behalf: this module actively uses `pytest` and CI demonstrates that there is no problem with running it, so let's run with it. Thank you !

PR #5131, opened 2025-03-07
ENH/DEP: use a more robust, external implementation of the Line Integral Convolution (LIC) algorithm

neutrinoceros:
## PR Summary

I've been spending some time rewriting the Line Integral Convolution (LIC) algorithm in Rust and published it as a new, minimal, Python package: [rLIC](https://github.com/neutrinoceros/rlic)
Because I wanted to see how it compared with yt's builtin Cython extension, in terms of performance, I ran a couple tests, and found that `rLIC` was not only much faster[^1] (more than x2):

bench_res_var_k
**this benchmark shows the execution time of convolving a 512 x 512 image with various kernel sizes**

... it also seems to be more correct. See this very simple test case, where yt's implementation appears to be completely broken:

comp_simple

This is the one example from our docs, as of yt 4.4.0
test_lic_native_Slice_z_density

and here it is with this branch

test_lic_rlic_Slice_z_density

so, instead of fixing yt's implementation, I would like to propose yt adopts `rLIC` as a dependency.

Since yt's implementation doesn't appear to be correct, it is no surprise that `rLIC` isn't a drop-in replacement: new results are not identical. Bug-for-bug backward compatibility being excluded, I also made a couple adjustments, like replacing the legacy prng API that's discouraged in numpy.

Important notes:
- `rLIC` wheels are `abi3`, which means that the package is already future compatible with Python 3.14+ and will remain so until CPython breaks abi3 and creates an abi4. If this happens, I will publish new wheels. The point being that depending on `rLIC` is not expected to cause any disturbance when adding support for newer Python versions in yt
- I support all platforms that numpy 2.2 supports
- ~`rLIC` is not yet distributed through conda-forge, which I'm aware is a requirement for any dependency of another package that wants to be shipped there. I don't mind putting the package on conda-forge too, pending acceptance of this proposal~
update: nevermind I'm going for it anyway https://github.com/conda-forge/staged-recipes/pull/29385
update 2: it's up on conda-forge

## PR Checklist

[^1]: if you're wondering, no, it's not just because my version is in Rust. I did spend time optimizing the implementation, in particular I avoided wasteful speculative executions which are extremely common with the naive implementation of this algorithm.


yt/visualization/plot_modifications.py
```diff
@@ -3238,10 +3238,10 @@ def __init__(
         field_x,
         field_y,
         texture=None,
-        kernellen=50.0,
+        kernellen=50,
         lim=(0.5, 0.6),
         cmap="binary",
-        alpha=0.8,
+        alpha=0.5,
```
neutrinoceros, line 3244: I found that a lower default alpha (opacity) made more sense with the correct implementation where "ink" is more uniformly distributed across the image so it's easier to hide the background completely if you're not careful.

yt/visualization/plot_modifications.py
```diff
@@ -3238,10 +3238,10 @@ def __init__(
         field_x,
         field_y,
         texture=None,
-        kernellen=50.0,
+        kernellen=50,
```
neutrinoceros, line 3241: a float never made sense here: `kernellen` represents the length of a 1D. I'm just fixing this in passing.

neutrinoceros:
(I'm putting this on the 4.5.0 milestone, mostly as to mean I don't want it to be backported, but it shouldn't block any particular release)

matthewturk:
whoa.

yt/visualization/plot_modifications.py
```diff
@@ -3270,29 +3270,18 @@ def __call__(self, plot):
             plot.data.axis, plot.data, self.field_y, bounds, (nx, ny)
         )
 
-        vectors = np.concatenate((pixX[..., np.newaxis], pixY[..., np.newaxis]), axis=2)
-
         if self.texture is None:
-            prng = np.random.RandomState(0x4D3D3D3)
-            self.texture = prng.random_sample((nx, ny))
-        elif self.texture.shape != (nx, ny):
```
neutrinoceros, line 3278: I removed this check because `rlic.convolve` already runs sanity checks over its inputs, including this one.

pyproject.toml
```diff
@@ -56,6 +56,7 @@ dependencies = [
     "unyt>=2.9.2",
     "tomli>=1.2.3;python_version < '3.11'",
     "typing-extensions>=4.4.0;python_version < '3.12'",
+    "rlic>=0.3.2",
```
neutrinoceros, line 59: On my phone at the moment, so note to self:
I should move this above other other unconditional dependencies. I also want to make sure the minimal version is correctly pinned on minimal env jobs.

neutrinoceros:
@olebole, do you have any objection to, or questions about adding this dependency to yt ? Would you expect it could be a burden to package rlic for debian ?

olebole:
@neutrinoceros I will not be able to package it for the upcoming Debian "Trixie" release which will come summer/autumn of this year, but for later this should be possible.
However, this would be my first package that uses Rust, so I am totally unfamiliar with its ecosystem, so expect some trivial questions during the packaging :grin:

neutrinoceros:
Happy to go down this road with you, it is _also_ my first experience mixing rust and Python, so I'm genuinely curious to find how easy or difficult the overall process turns out !

neutrinoceros:
Let's see how it plays out over at conda-forge first https://github.com/conda-forge/staged-recipes/pull/29385

neutrinoceros:
power cycling this to refresh CI

neutrinoceros:
Moving to rLIC should soon unblock the missing feature for LIC: compat with periodic datasets https://github.com/neutrinoceros/rlic/pull/196

neutrinoceros:
@matthewturk Gentle reminder

matthewturk:
Oh!  I am sorry that I was blocking on this.  I'm up for it going in, and appreciate you taking the time to implement it.

PR #5137, opened 2025-03-19
update testing.integrate_kernel type hints

### LINKED ISSUE #5136: mypy failure with np 2.2.4
Running mypy with the latest numpy (v 2.2.4, released 3 days ago) results in a new failure: 

```python
yt/__init__.py:98: note: In module imported here:
yt/testing.py: note: In function "integrate_kernel":
yt/testing.py:148: error: Argument 1 has incompatible type "ndarray[tuple[int, ...], dtype[Any]]"; expected "float"  [arg-type]
yt/utilities/performance_counters.py: note: In member "print_stats" of class "PerformanceCounters":
yt/utilities/performance_counters.py:73: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
Found 1 error in 1 file (checked 516 source files)
```

chrishavlin:
Close #5136

### DIFF
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -17,6 +17,7 @@ import matplotlib
 import numpy as np
 from more_itertools import always_iterable
 from numpy.random import RandomState
+from numpy.typing import NDArray
 from unyt.exceptions import UnitOperationError
 
 from yt._maintenance.deprecation import issue_deprecation_warning
@@ -91,7 +92,7 @@ def assert_rel_equal(a1, a2, decimals, err_msg="", verbose=True):
 
 # tested: volume integral is 1.
 def cubicspline_python(
-    x: float | np.ndarray,
+    x: float | NDArray[np.floating],
 ) -> np.ndarray:
     """
     cubic spline SPH kernel function for testing against more
@@ -118,7 +119,9 @@ def cubicspline_python(
 
 
 def integrate_kernel(
-    kernelfunc: Callable[[float], float], b: float, hsml: float
+    kernelfunc: Callable[[float | NDArray[np.floating]], float | NDArray[np.floating]],
+    b: float,
+    hsml: float,
 ) -> float:
     """
     integrates a kernel function over a line passing entirely
@@ -154,11 +157,11 @@ _zeroperiods = np.array([0.0, 0.0, 0.0])
 
 
 def distancematrix(
-    pos3_i0: np.ndarray,
-    pos3_i1: np.ndarray,
+    pos3_i0: NDArray[np.floating],
+    pos3_i1: NDArray[np.floating],
     periodic: tuple[bool, bool, bool] = (True,) * 3,
-    periods: np.ndarray = _zeroperiods,
-) -> np.ndarray:
+    periods: NDArray[np.floating] = _zeroperiods,
+) -> NDArray[np.floating]:
     """
     Calculates the distances between two arrays of points.
 
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -10,14 +10,14 @@ from collections.abc import Callable, Mapping
 from functools import wraps
 from importlib.util import find_spec
 from shutil import which
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, TypeVar
 from unittest import SkipTest
 
 import matplotlib
 import numpy as np
+import numpy.typing as npt
 from more_itertools import always_iterable
 from numpy.random import RandomState
-from numpy.typing import NDArray
 from unyt.exceptions import UnitOperationError
 
 from yt._maintenance.deprecation import issue_deprecation_warning
@@ -92,8 +92,8 @@ def assert_rel_equal(a1, a2, decimals, err_msg="", verbose=True):
 
 # tested: volume integral is 1.
 def cubicspline_python(
-    x: float | NDArray[np.floating],
-) -> np.ndarray:
+    x: float | npt.NDArray[np.floating],
+) -> npt.NDArray[np.floating]:
     """
     cubic spline SPH kernel function for testing against more
     effiecient cython methods
@@ -119,10 +119,12 @@ def cubicspline_python(
 
 
 def integrate_kernel(
-    kernelfunc: Callable[[float | NDArray[np.floating]], float | NDArray[np.floating]],
-    b: float,
-    hsml: float,
-) -> float:
+    kernelfunc: Callable[
+        [float | npt.NDArray[np.floating]], float | npt.NDArray[np.floating]
+    ],
+    b: float | npt.NDArray[np.floating],
+    hsml: float | npt.NDArray[np.floating],
+) -> float | npt.NDArray[np.floating]:
     """
     integrates a kernel function over a line passing entirely
     through it
@@ -150,18 +152,24 @@ def integrate_kernel(
     dx = np.diff(xe, axis=0)
     spv = kernelfunc(np.sqrt(xc**2 + x**2))
     integral = np.sum(spv * dx, axis=0)
-    return pre * integral
+    result = pre * integral
+    if isinstance(result, np.floating):
+        return result.item()
+    return result
 
 
 _zeroperiods = np.array([0.0, 0.0, 0.0])
 
 
+_FloatingT = TypeVar("_FloatingT", bound=np.floating)
+
+
 def distancematrix(
-    pos3_i0: NDArray[np.floating],
-    pos3_i1: NDArray[np.floating],
+    pos3_i0: npt.NDArray[_FloatingT],
+    pos3_i1: npt.NDArray[_FloatingT],
     periodic: tuple[bool, bool, bool] = (True,) * 3,
-    periods: NDArray[np.floating] = _zeroperiods,
-) -> NDArray[np.floating]:
+    periods: npt.NDArray[_FloatingT] = _zeroperiods,
+) -> npt.NDArray[_FloatingT]:
     """
     Calculates the distances between two arrays of points.
 
--- a/yt/testing.py
+++ b/yt/testing.py
@@ -124,7 +124,7 @@ def integrate_kernel(
     ],
     b: float | npt.NDArray[np.floating],
     hsml: float | npt.NDArray[np.floating],
-) -> float | npt.NDArray[np.floating]:
+) -> npt.NDArray[np.floating]:
     """
     integrates a kernel function over a line passing entirely
     through it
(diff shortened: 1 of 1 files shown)


neutrinoceros (changes requested):
thanks for catching this. However, note that [`np.ndarray` is not a valid type annotation](https://github.com/numpy/numpy/issues/27945#issuecomment-2532076572), we should use `NDArray[np.floating]` instead (`from numpy.typing import NDArray`).

Furthermore, as numpy's type annotations get more and more refined, it is becoming apparent that mypy is not the correct typechecker we should be using. Numpy devs recommend `basedpyright` over it. I might start a migration later, but I just wanted to raise awareness for now. See https://github.com/numpy/numpy/releases/tag/v2.2.1

chrishavlin:
I swear I knew that at one point :) but in this case I actually just copied the type hinting in the previous function without thinking. so i also added a commit here to switch those other occurrences of np.ndarray in type hints in this file.

neutrinoceros:
Looks like you accidentally committed an unrelated (and massive cpp file)

yt/testing.py
```diff
@@ -91,7 +92,7 @@ def assert_rel_equal(a1, a2, decimals, err_msg="", verbose=True):
 
 # tested: volume integral is 1.
 def cubicspline_python(
-    x: float | np.ndarray,
+    x: float | NDArray[np.floating],
```
neutrinoceros, line 95: I guess the return annotation should be changed as well

yt/testing.py
```diff
@@ -118,7 +119,9 @@ def cubicspline_python(
 
 
 def integrate_kernel(
-    kernelfunc: Callable[[float], float], b: float, hsml: float
+    kernelfunc: Callable[[float | NDArray[np.floating]], float | NDArray[np.floating]],
+    b: float,
+    hsml: float,
 ) -> float:
```
neutrinoceros, line 125: the return value is actually `np.float64`, but it'd be safer to resolve this disparity by actually returning a `float`.c Can you update the `return` statement ?

yt/testing.py
```diff
@@ -154,11 +157,11 @@ def integrate_kernel(
 
 
 def distancematrix(
-    pos3_i0: np.ndarray,
-    pos3_i1: np.ndarray,
+    pos3_i0: NDArray[np.floating],
+    pos3_i1: NDArray[np.floating],
     periodic: tuple[bool, bool, bool] = (True,) * 3,
-    periods: np.ndarray = _zeroperiods,
-) -> np.ndarray:
```
neutrinoceros, line 161: actually here the output dtype is bound to be the same as `pos3_i3`, so, instead of `np.floating`, we should use `_FloatingT = TypeVar("_FloatingT", bound=np.floating)` here

yt/testing.py
chrishavlin: oh ya, thought I had put that in. maybe i forgot to commit it.

chrishavlin:
the stray .cpp is gone now.

yt/testing.py
neutrinoceros: Did you forget again ? 😅

chrishavlin: hah, sorry, didn't push any changes yet, only fixed the stray .cpp

yt/testing.py
chrishavlin: hmm. so the typing here doesn't seem to be consistent with how `integrate_kernel` is used elsewhere. here's a test that uses arrays for `b` and `hsml`:

https://github.com/yt-project/yt/blob/d44472136b55a5596041d8a022ca4870e36346ee/yt/data_objects/tests/test_rays.py#L120-L138

not certain if the typing for this function needs to updated so that `b` and `hsml` are `float | NDArray` or if that test there is not quite right. @nastasha-w I believe you wrote that test, any thoughts?

nastasha-w: By the way, I am already editing that function for #5121 , so this will be a fun rebase for me later. Current status:
```python
def integrate_kernel(
    kernelfunc: Callable[[float | np.ndarray], np.ndarray],
    b: float | np.ndarray,
    hsml: float | np.ndarray,
    nsample: int = 500,
) -> float:
```
although I'm now realizing that's wrong, and I actually count on it to return arrays in some cases in a different function.

nastasha-w:
Sorry, I seem to have made a mess of the type annotations in those test functions. In case it helps, I think these are all the files where I use the problem function in `testing.py`:
- yt/visualization/tests/test_offaxisprojection_pytestonly.py
- yt/geometry/coordinates/tests/test_sph_pixelization.py
- yt/geometry/coordinates/tests/test_sph_pixelization_pytestonly.py

yt/testing.py
chrishavlin: > Yeah sorry, the annotations here are a bit of a mess. I basically just edited until the type checker stopped throwing errors;

No worries! I'm forever learning about proper python type checking myself... 

> and I actually count on it to return arrays in some cases in a different function.

So sounds like having `b`, `hsml` and the return value be `float | NDArray` is consistent with how you're using it. I'll go with that here. also, FYI in case you didn't read through all the discussion in this PR , [this comment](https://github.com/yt-project/yt/pull/5137#pullrequestreview-2701898295) on `np.ndarray` vs `np.typing.NDArray` in type hints might be helpful for your changes.

chrishavlin:
@neutrinoceros  just fyi, this is ready for you to look at when you get a chance (no worries if you can't get to it immediately, just wanted to make sure you weren't waiting on me).

nastasha-w:
Welp, I'm learning some new type annotation options here! Would it make sense to also use the _FloatingT type for the kernel integrator? That one also doesn't make assumptions on the exact float precision of the inputs.

yt/testing.py
```diff
@@ -118,8 +119,12 @@ def cubicspline_python(
...
 
 def integrate_kernel(
-    kernelfunc: Callable[[float], float], b: float, hsml: float
-) -> float:
+    kernelfunc: Callable[
+        [float | npt.NDArray[np.floating]], float | npt.NDArray[np.floating]
+    ],
+    b: float | npt.NDArray[np.floating],
+    hsml: float | npt.NDArray[np.floating],
+) -> float | npt.NDArray[np.floating]:
```
neutrinoceros, line 127: It's often much preferable to be strict on the return type. Here I don't see any reason not to be
```suggestion
) -> float:
```

yt/testing.py
```diff
@@ -147,18 +152,24 @@ def integrate_kernel(
     dx = np.diff(xe, axis=0)
     spv = kernelfunc(np.sqrt(xc**2 + x**2))
     integral = np.sum(spv * dx, axis=0)
-    return pre * integral
+    result = pre * integral
+    if isinstance(result, np.floating):
+        return result.item()
+    return result
```
neutrinoceros, line 158:
```suggestion
    return float(pre * integral)
```

neutrinoceros (changes requested):
almost there. Sorry for the delay !

neutrinoceros:
> Would it make sense to also use the _FloatingT type for the kernel integrator? That one also doesn't make assumptions on the exact float precision of the inputs.

depends.
`_FloatingT`, as defined here, doesn't just act as a placeholder for an unknown dtype: it is also a `TypeVar` and as such, it expresses a relation between input types and output types.

Can you point me to the exact function you're refering to ?

yt/testing.py
nastasha-w: I don't think this is a good idea. This function can be called on an arrays of `b` and `hsml` values and `float(array)` will not convert an array to the `np.float` type. Being strict on return type is fine, but then we should either revert to always returning an array, and changing the `dtype`, or do that type cast before picking out the single element for a 0d-array.

nastasha-w:
@neutrinoceros I meant the function `integrate_kernel` in `testing.py` (line 120 in my version of `testing.py`). As-is, the output will depend on the exact float types used for `b` and `hsml`, although it isn't obvious to me which type will be used if they aren't the same... Perhaps your explicit type-casting suggestion is a better idea here, but I would like to retain the ability to return arrays and not just single float values.

yt/testing.py
nastasha-w: ... e.g. `np.float32(pre * integral)` would work for arrays though. We'd need to pick which floating point precision to specify in that case though.

neutrinoceros: I'm confused. Why would `np.float32(...)` work but not `float(...)` ?

neutrinoceros:
Then how is the decision to return an array taken, and would it make sense to unconditionally return an array instead ?

nastasha-w:
Yeah, the version before this PR would always return an array; if the inputs were floats, the result would be a 0d-array. I just called `np.array` on the `hsml` and `b` inputs. 0d-arrays are annoying though, I'd be fine with returning a `float` if neither `b` nor `hsml` was input as an array. e.g.
```python
if not (isinstance(b, np.ndarray) or isinstance(hsml, np.ndarray)):
    return result.item()
return result
```

chrishavlin:
> would it make sense to unconditionally return an array instead ? (@neutrinoceros )

This sounds like a good solution to me. I agree that the conditional return type is not ideal if we can simplify it. 

> 0d-arrays are annoying though (@nastasha-w )

how about always returning an array with `np.atleast_1d`  (i.e., `return np.atleast_1d(pre * integral)`)?

yt/testing.py
nastasha-w: I'm not sure, but that's what worked and didn't work when I tried it out:
```python
Python 3.13.2 (main, Feb  4 2025, 14:51:09) [Clang 16.0.0 (clang-1600.0.26.6)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> a = np.arange(3)
>>> float(a)
  File "<python-input-2>", line 1, in <module>
    float(a)
    ~~~~~^^^
TypeError: only length-1 arrays can be converted to Python scalars
>>> np.float32(a)
array([0., 1., 2.], dtype=float32)
>>> np.float(a)
  File "<python-input-4>", line 1, in <module>
    np.float(a)
    ^^^^^^^^
  File "/Users/nastasha/code/venvs/ytdev_pixav/lib/python3.13/site-packages/numpy/__init__.py", line 397, in __getattr__
    raise AttributeError(__former_attrs__[attr], name=None)
AttributeError: module 'numpy' has no attribute 'float'.
`np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

chrishavlin:
Latest push updates integrate_kernel to always return an array (using `np.atleast_1d` so it's never a 0d array). 

Don't think there was anything left to do?

nastasha-w:
It looks good to me, but then again, so did the initial mess I made here

neutrinoceros (approved):

neutrinoceros:
Ah, now it looks like we can't merge yet because the jenkins server is apparently down ?

PR #5141, opened 2025-03-20
Use matplotlib backend registry when possible

### LINKED ISSUE #5138: replace BACKEND_SPECS with matplotlib.backends.backend_registry
While updating the `BACKEND_SPECS` in #5135 , I noticed that matplotlib 3.9 introduced a `matplotlib.backends.backend_registry` that could be used to replace `BACKEND_SPECS` entirely and offload the backend details to matplotlib.

e.g., could add the following to get the `FigureCanvas` and `FigureManager` classes in [`_get_canvas_classes`](https://github.com/yt-project/yt/blob/5ea70433afd340b2370994ae1f0202ac40091a0b/yt/visualization/base_plot_types.py#L152-L173)

```python
if matplotlib.__version_info__ >= (3, 9):
    from matplotlib.backends import backend_registry

    mod = backend_registry.load_backend_module(key)
    FigureCanvas = getattr(mod, 'FigureCanvas')
    FigureManager = getattr(mod, 'FigureManager')
```

Only thing I'm a little uncertain of is the FigureManager -- from what I understand, the above code pulls out `FigureCanvas` and `FigureManager` classes that are aliased to the specific implementation for each backend and there should always be a `FigureManager` specified (which seems to have been the case for quite a while now... since https://github.com/matplotlib/matplotlib/pull/8773). Any reason to preserve yt's behavior for the backends where `None` is specified for a `FigureManager` in `BACKEND_SPECS`? I think no, but wanted to get a second opinion.

here's the relevant bits of code i'm referring to: 

https://github.com/yt-project/yt/blob/5ea70433afd340b2370994ae1f0202ac40091a0b/yt/visualization/base_plot_types.py#L47-L66

and 

https://github.com/yt-project/yt/blob/5ea70433afd340b2370994ae1f0202ac40091a0b/yt/visualization/base_plot_types.py#L152-L173

Also, I don't see a huge benefit to putting this change in now vs waiting til yt's min matplotlib version is 3.9, but wanted to get these thoughts down while they're fresh.

chrishavlin:
Closes #5138 

I also added a test for the existing `BACKEND_SPECS`, just checking that the backend keys are valid, which resulted in me realizing that gtk, gtkagg and gtkcairo have all been deprecated for a while (https://github.com/ipython/ipython/issues/14368 , https://github.com/matplotlib/matplotlib/pull/10426) in favor of versioned keys (e.g., GTK3Agg). I removed the deprecated keys. 

Also worth noting that with this change, for folks using matplotlib>=3.9, yt plotting should now work with all the interactive backends listed at https://matplotlib.org/stable/users/explain/figure/backends.html#interactive-backends , which includes some that were missing from `BACKEND_SPECS`, e.g.: `GTK4Agg` and `GTK4Cairo`.

### DIFF
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -45,9 +45,6 @@ if TYPE_CHECKING:
 
 
 BACKEND_SPECS = {
-    "gtk": ["backend_gtk", "FigureCanvasGTK", "FigureManagerGTK"],
-    "gtkagg": ["backend_gtkagg", "FigureCanvasGTKAgg", None],
-    "gtkcairo": ["backend_gtkcairo", "FigureCanvasGTKCairo", None],
     "macosx": ["backend_macosx", "FigureCanvasMac", "FigureManagerMac"],
     "qt5agg": ["backend_qt5agg", "FigureCanvasQTAgg", None],
     "qtagg": ["backend_qtagg", "FigureCanvasQTAgg", None],
--- a/yt/visualization/base_plot_types.py
+++ b/yt/visualization/base_plot_types.py
@@ -138,6 +138,8 @@ class PlotMPL:
         figure_canvas, figure_manager = self._get_canvas_classes()
         self.canvas = figure_canvas(self.figure)
         if figure_manager is not None:
+            # with matplotlib >= 3.9, figure_manager should always be not None
+            # see _get_canvas_classes for details.
             self.manager = figure_manager(self.canvas, 1)
 
         self.axes.tick_params(
@@ -151,11 +153,20 @@ class PlotMPL:
 
     def _get_canvas_classes(self):
         if self.interactivity:
-            key = str(matplotlib.get_backend()).lower()
+            key = str(matplotlib.get_backend())
         else:
             key = "agg"
 
-        module, fig_canvas, fig_manager = BACKEND_SPECS[key]
+        if matplotlib.__version_info__ >= (3, 9):
+            # once yt has a minimum matplotlib version of 3.9, this branch
+            # can replace the rest of this function and BACKEND_SPECS can
+            # be removed. See https://github.com/yt-project/yt/issues/5138
+            from matplotlib.backends import backend_registry
+
+            mod = backend_registry.load_backend_module(key)
+            return mod.FigureCanvas, mod.FigureManager
+
+        module, fig_canvas, fig_manager = BACKEND_SPECS[key.lower()]
 
         mod = __import__(
             "matplotlib.backends",
--- a/yt/visualization/tests/test_base_plot_types.py
+++ b/yt/visualization/tests/test_base_plot_types.py
@@ -1,0 +1,19 @@
+import matplotlib
+import pytest
+
+from yt.visualization.base_plot_types import BACKEND_SPECS
+
+
+@pytest.mark.skipif(
+    matplotlib.__version_info__ < (3, 9),
+    reason="This test requires matplotlib 3.9 or higher.",
+)
+@pytest.mark.parametrize("backend_key", BACKEND_SPECS.keys())
+def test_backend_specs(backend_key):
+    # note: while this test itself requires matplotlib 3.9, it is
+    # testing functionality in yt that can be removed once yt has
+    # a minimal matplotlib version of 3.9,
+    # see https://github.com/yt-project/yt/issues/5138
+    from matplotlib.backends import backend_registry
+
+    assert backend_registry.is_valid_backend(backend_key)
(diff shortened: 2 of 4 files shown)


yt/visualization/base_plot_types.py
```diff
@@ -151,11 +150,20 @@ def _create_axes(self, axrect: tuple[float, float, float, float]) -> None:
...
     def _get_canvas_classes(self):
         if self.interactivity:
-            key = str(matplotlib.get_backend()).lower()
+            key = str(matplotlib.get_backend())
         else:
             key = "agg"
 
-        module, fig_canvas, fig_manager = BACKEND_SPECS[key]
+        if matplotlib.__version_info__ >= (3, 9):
+            # once yt has a minimum matplotlib version of 3.9, this branch
```
neutrinoceros, line 158: I would not bother with this comment; to me, it's 100% implied by the code itself. But, NBD if you think it's helpful.

yt/visualization/tests/test_base_plot_types.py
```diff
@@ -0,0 +1,19 @@
...
+
+from yt.visualization.base_plot_types import BACKEND_SPECS
+
+
+@pytest.mark.skipif(
+    matplotlib.__version_info__ < (3, 9),
+    reason="This test requires matplotlib 3.9 or higher.",
+)
+@pytest.mark.parametrize("backend_key", BACKEND_SPECS.keys())
+def test_backend_specs(backend_key):
```
neutrinoceros, line 12: Seems like this test lead to the discovery of invalid entries, so it's great. Do you think it's possible to also test that the dictionary is complete ?

yt/visualization/base_plot_types.py
```diff
@@ -151,11 +150,20 @@ def _create_axes(self, axrect: tuple[float, float, float, float]) -> None:
 
     def _get_canvas_classes(self):
         if self.interactivity:
-            key = str(matplotlib.get_backend()).lower()
+            key = str(matplotlib.get_backend())
```
neutrinoceros, line 153: going down the rabbit hole, I was wondering why we would convert to `str`. I found that the only value that *may* be returned by `matplotlib.get_backend` that is not a string is `None`, and since we don't have a `'none'` key in `BACKEND_SPECS`, I don't see the point of casting.
```suggestion
            key = matplotlib.get_backend()
```
In fact this possibility was only introduced very recently (matplotlib 3.10).


xref #29039

yt/visualization/base_plot_types.py
chrishavlin: mainly added it cause BACKEND_SPECS is not defined within the scope of the function. so i'd rather keep it i think? but also don't feel strongly about it.

yt/visualization/tests/test_base_plot_types.py
chrishavlin: not in a satisfyingly automated way... you can get a list of all backends with:

```
>>> from matplotlib.backends import backend_registry
>>> backend_registry.list_builtin()
['gtk3agg', 'gtk3cairo', 'gtk4agg', 'gtk4cairo', 'macosx', 'nbagg', 'notebook', 'qtagg', 'qtcairo', 'qt5agg', 'qt5cairo', 'tkagg', 'tkcairo', 'webagg', 'wx', 'wxagg', 'wxcairo', 'agg', 'cairo', 'pdf', 'pgf', 'ps', 'svg', 'template']
```

but i'm not sure we want the non-interactive backends in `BACKEND_SPECS` (pdf, svg, etc.). And since the test could only be run with matplotlib 3.9+, it also would not be very future proof. If a new backend shows up in a new matplotlib release, it would cause the completeness test to fail but the new backend would not actually work for the case where `BACKEND_SPEC` actually gets used. So we'd be back to manually maintaining `BACKEND_SPEC` until yt bumps to matplotlib >= 3.9. So I'd prefer to leave `BACKEND_SPEC` alone at this point and encourage folks to upgrade matplotlib instead -- I could add a new try/except to the `BACKEND_SPEC` branch that suggests upgrading matplotlib if you get a key error from `BACKEND_SPEC` ?

neutrinoceros: Thanks for the additional info. Yes, such a try/except could improve UX, but I now think the PR is good as is, so, up to you !

neutrinoceros (approved):

yt/visualization/tests/test_base_plot_types.py
chrishavlin: ok, i'll leave it off then. given the lack of bug reports and questions i don't think it's a huge priority.

PR #5154, opened 2025-04-22
bump ubuntu-20.04 to -22.4 in github actions 

### LINKED ISSUE #5153: TST: github actions need to update ubuntu version
See, e.g., https://github.com/yt-project/yt/actions/runs/14584544248/job/40907635253?pr=4702 , which is still using the deprecated and now unavailable Ubuntu 20.04 runner

chrishavlin:
The minimal test and the wheels action will now use the currently oldest available ubuntu (22.4). 

close #5153


chrishavlin:
oops, like we need to bump the minimal python as well. 3.10.3 is not available on ubuntu 22.04

neutrinoceros:
Should we bump `python-requires` from `pyproject.toml` too ?

chrishavlin:
Oh ya, I suppose we should. Feels a little odd to bump the min python when it's only needed here for CI to run, 3.10.3 will continue to work just fine. But it also wouldn't be ideal cause we wouldn't know if it stopped working... so let's bump it.

neutrinoceros:
Let's backport it too. Thanks !

neutrinoceros (approved):

PR #5155, opened 2025-04-22
BUG: Better behavior when Cholla doesn't write "mu" attribute

mabruzzo:
## Basic Background

Currently, the Cholla frontend always tries to read the ``"mu"`` attribute from the hdf5 files. However, this is not a standard attribute written by Cholla. When this attribute is missing, the frontend falls back to a value of ``1.0``. This is almost always wrong (it is usually ``0.6``,  but in some cases it can be hardcoded to different values).

It's on my todo list to start having Cholla write out the ``"mu"`` attribute (when not using species fields).

## PR Summary

This patch tweaks the frontend to achieve better behavior for all existing Cholla datasets without the ``"mu"`` attribute. Namely, we don't set a default value. The patch also enures that we don't define a temperature field unless the ``"mu"`` attribute is explicitly defined (or the ``default_species_fields`` keyword argument is specified in ``yt.load``)

### DIFF
--- a/yt/frontends/cholla/data_structures.py
+++ b/yt/frontends/cholla/data_structures.py
@@ -1,6 +1,5 @@
 import os
 import weakref
-from contextlib import suppress
 
 import numpy as np
 
@@ -112,7 +111,7 @@ class ChollaDataset(Dataset):
             self.current_time = attrs["t"][:]
             self._periodicity = tuple(attrs.get("periodicity", (False, False, False)))
             self.gamma = attrs.get("gamma", 5.0 / 3.0)
-            with suppress(KeyError):
+            if "mu" in attrs:
                 # other yt-machinery can't handle ds.mu == None, so we simply
                 # avoid defining the mu attribute if we don't know its value
                 self.mu = attrs["mu"]
--- a/yt/frontends/cholla/data_structures.py
+++ b/yt/frontends/cholla/data_structures.py
@@ -8,6 +8,7 @@ from yt.data_objects.static_output import Dataset
 from yt.funcs import setdefaultattr
 from yt.geometry.api import Geometry
 from yt.geometry.grid_geometry_handler import GridIndex
+from yt.utilities.logger import ytLogger as mylog
 from yt.utilities.on_demand_imports import _h5py as h5py
 
 from .fields import ChollaFieldInfo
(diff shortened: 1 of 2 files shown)


yt/frontends/cholla/data_structures.py
```diff
@@ -111,7 +112,10 @@ def _parse_parameter_file(self):
             self.current_time = attrs["t"][:]
             self._periodicity = tuple(attrs.get("periodicity", (False, False, False)))
             self.gamma = attrs.get("gamma", 5.0 / 3.0)
-            self.mu = attrs.get("mu", 1.0)
+            with suppress(KeyError):
```
neutrinoceros, line 115: Why not simply 

```suggestion
            if "mu" in attrs:
```
?

mabruzzo: Obviously that's better! I just fixed it!

Originally, I tried to write a try...except case and then I tried to make it shorter. I clearly wasn't thinking at all

chummels:
I think that I initiated a default `mu` value when I created the frontend to ensure that a temperature field was generated.  If Evan and you are OK with not having a temperature field by default, then I think this PR is OK.  Alternatively, you could just set mu to be 0.6 by default unless otherwise defined in the dataset?

chummels:
I defer to @evaneschneider on this one, since she's the boss on CHOLLA.

mabruzzo:
So I touched base with Evan, and this PR includes the preferred solution.

The basic rationale is that we would prefer that people avoid use a *slightly* incorrect temperature field. We would prefer that people notice that the temperature field is not defined and then go use one of the post-processing scripts from the Cholla repository to manually add the "mu" parameter to the hdf5 files (with the correct value based on the simulation). 

Plus, if people really want a quick and dirty estimate of temperature, they can use ``yt.load`` with the ``default_species_fields`` keyword.

yt/frontends/cholla/data_structures.py
```diff
@@ -111,7 +111,10 @@ def _parse_parameter_file(self):
             self.current_time = attrs["t"][:]
             self._periodicity = tuple(attrs.get("periodicity", (False, False, False)))
             self.gamma = attrs.get("gamma", 5.0 / 3.0)
-            self.mu = attrs.get("mu", 1.0)
+            if "mu" in attrs:
+                # other yt-machinery can't handle ds.mu == None, so we simply
+                # avoid defining the mu attribute if we don't know its value
+                self.mu = attrs["mu"]
```
chrishavlin, line 117: Given your comment here:

> We would prefer that people notice that the temperature field is not defined and then go use one of the post-processing scripts from the Cholla repository to manually add the "mu" parameter to the hdf5 files (with the correct value based on the simulation).

I think it'd be helpful to have a log message when `mu` is not defined, pointing folks in the right direction -- the generic field errors when you try to access a nonexistant field won't be super insightful. Only downside is it might be a tad noisy since you'd see the message every time the dataset is opened, so feel free to ignore this suggestion :) 

If you do want to add it, you can import yt's logger with 

```
```

and check out https://github.com/yt-project/yt/blob/main/yt/loaders.py for some example usage.

chummels: I think this is an excellent idea.  When things break, helpful error messages can save everyone a headache.

mabruzzo: Thanks for the suggestion. I added a message (while I was there, I added an error in case the `default_species_fields` kwarg and `mu` attribute are both specified

chummels (approved):

chrishavlin:
pre-commit.ci autofix

chrishavlin (approved):

PR #5159, opened 2025-04-29
[BUGFIX] RAMSES: Use one registry per dataset

cphyc:
## PR Summary
This prevents issues when trying to load two snapshots/simulations with different variables. It makes all the detection specific to the one dataset at hand

## PR Checklist

- [x] ~New features are documented, with docstrings and narrative docs~

## Code to reproduce bug

```python
import yt

ds0 = yt.load_sample("output_00080")
ds1 = yt.load_sample("ramses_mhd_128")

# When loading sp0 then sp1, the code crashes
# When loading sp1 then sp0, it works :upside_down_face: 
#sp0 = ds0.sphere([0.5]*3, 0.01)
sp1 = ds1.sphere([0.5]*3, 0.01)
sp0 = ds0.sphere([0.5]*3, 0.01)

print(ds0.parameters)

sp0["gas", "velocity_x"].max().to("km/s")
```

Traceback:
```
ValueError
Cell In[28], line 12
      8 # sp0 = ds0.sphere([0.5]*3, 0.01)
     10 print(ds0.parameters)
---> 12 sp0["gas", "velocity_x"].max().to("km/s")  # this raises an OSError when reading the Fortran records.
[...]
File yt/frontends/ramses/field_handlers.py:264, in FieldFileHandler.offset(self)
    250     min_level = self.domain.ds.min_level
    252     # The file is as follows:
    253     # > headers
    254     # loop over levels
   (...)    262     # So there are 8 * nvars records each with length (nocts, )
    263     # at each (level, cpus)
--> 264     offset, level_count = read_offset(
    265         fd,
    266         min_level,
    267         self.domain.domain_id,
    268         self.parameters["nvar"],
    269         self.domain.amr_header,
    270         Nskip=nvars * 8,
    271     )
    273 self._level_count = level_count
    274 return offset

File yt/frontends/ramses/io_utils.pyx:119, in yt.frontends.ramses.io_utils.read_offset()

File yt/frontends/ramses/io_utils.pyx:147, in yt.frontends.ramses.io_utils.read_offset()

File yt/utilities/cython_fortran_utils.pyx:201, in yt.utilities.cython_fortran_utils.FortranFile.read_int()

ValueError: Size obtained (1068636786) does not match with the expected size (4) of record
```

### DIFF
--- a/yt/frontends/ramses/field_handlers.py
+++ b/yt/frontends/ramses/field_handlers.py
@@ -42,6 +42,7 @@ class HandlerMixin:
         self.ds = ds = domain.ds
         self.domain = domain
         self.domain_id = domain.domain_id
+
         basename = os.path.abspath(ds.root_folder)
         iout = int(os.path.basename(ds.parameter_filename).split(".")[0].split("_")[1])
 
@@ -166,6 +167,9 @@ class FieldFileHandler(abc.ABC, HandlerMixin):
             register_field_handler(cls)
 
         cls._unique_registry = {}
+        cls.parameters = {}
+        cls.rt_parameters = {}
+        cls._detected_field_list = {}
         return cls
 
     def __init__(self, domain):
@@ -231,6 +235,10 @@ class FieldFileHandler(abc.ABC, HandlerMixin):
 
         return self._level_count
 
+    @property
+    def field_list(self):
+        return self._detected_field_list[self.ds.unique_identifier]
+
     @cached_property
     def offset(self):
         """
@@ -242,7 +250,7 @@ class FieldFileHandler(abc.ABC, HandlerMixin):
         It should be generic enough for most of the cases, but if the
         *structure* of your fluid file is non-canonical, change this.
         """
-        nvars = len(self.field_list)
+        nvars = len(self._detected_field_list[self.ds.unique_identifier])
         with FortranFile(self.fname) as fd:
             # Skip headers
             nskip = len(self.attrs)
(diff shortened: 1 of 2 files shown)


matthewturk:
This is just RAMSES, right?  Maybe edit the PR title?

cphyc:
A bit of background. There is a ton of tricks in the RAMSES frontend to achieve good performance despite all the data being scattered in hundreds or thousands of file. One such trick is to infer which and how many fields there are on disk, assuming all files from a given snapshot have the same number.

This bug is triggered because the frontend stores the detected list of fields at the class level (I.e. Once per Python instance). If you load two snapshots with different fields and access them in the same order (`load A, load B, access A, access B`), then when assessing A, yt uses the fields detected for B. The fix entails doing that detection per snapshot.

chrishavlin:
This looks good to me, but FYI, the server that the full test suite runs on is nominally dead/dying  (see the slack dev channel discussion) and so the test hangs indefinitely and the new test here won't actually run right now. If there's an easy and quick way to reformulate the test to use pytest I'd suggest doing that. But that's probably hard to do since you really need a dataset here to expose the problematic behavior, in which case I think we should just go ahead and merge assuming you've run the test locally.

cphyc:
I've run the test locally; it fails on main but passes with this PR.
I've now run all other (RAMSES+nose) tests locally to make sure they still pass (they do).

chrishavlin:
Thanks for doing the extra checks, @cphyc !

chrishavlin (approved):

PR #5168, opened 2025-05-15
Add a frontend for Quokka dataset

chongchonghe:
## PR Summary

This PR add a frontend for loading Quokka dataset. This frontend fully support uniform-grid or AMR datasets in cartesian coordinates and particles. QUOKKA is based on the AMReX framework, so we define a new `QuokkaDataset` class that inherits from `AMReXDataset`. 

This PR is fully tested on my computer. We have also added Answer Tests in `frontends/amrex/test_outputs.py` to validate new features with Quokka data that we provided. 

We prepared a Jupyter Notebook with examples and test dataset: https://github.com/Rongjun-ANU/README-of-yt-frontend-for-QUOKKA/blob/main/README.ipynb 

Authors: @Rongjun-ANU @chongchonghe 

[Quokka](https://github.com/quokka-astro/quokka) is a two-moment AMR radiation hydrodynamics (with self-gravity, particles, and chemistry) on CPUs/GPUs for astrophysics. 

## PR Checklist


chongchonghe:
@BenWibking @Rongjun-ANU

chongchonghe:
I couldn't understand the remaining pre-commit.ci failsures. I would appreciate some explanation or help from the reviewers.

chongchonghe:
pre-commit.ci autofix

doc/source/examining/loading_data.rst
```diff
@@ -2133,6 +2133,175 @@ zero.
...
+QUOKKA datasets can also include face-centered variables. These are stored in
+the `fc_vars` directory which is inside the plotfile directory. When loading a
+Quokka dataset, YT automatically detects and loads any face-centered variables
+from the `fc_vars/x*****`, `fc_vars/y*****`, and `fc_vars/z*****`
+subdirectories. These are made available as attributes of the main dataset
+(`ds.ds_fc_x`, `ds.ds_fc_y`, `ds.ds_fc_z`). Each face-centered dataset is a full
+YT dataset object with its own fields and mesh structure, and includes a
+reference back to the parent dataset, `ds.ds_fc_x.parent_ds`. This allows for
+easy access to both cell-centered and face-centered data in the same analysis
+workflow.
```
chrishavlin, line 2286: These lines are the ones pre-commit has flagged: I believe the error is just that your in-line code snippets here need to use two back ticks rather than one. e.g., 

```
    `fc_vars`
```

needs to be 
```
    ``fc_vars``
```

chrishavlin: I think the rest of the pre-commit errors should be handled by the bot, but you may need to manually fix these first?

chongchonghe:
pre-commit.ci autofix

doc/source/examining/loading_data.rst
chongchonghe: Thanks a lot for the help. I have fixed all the pre-commit errors.

chrishavlin:
@chongchonghe I should be able to take a detailed look late next week. 

one quick question: do you have a sample dataset you can share and upload? I think it'd be particularly helpful to have one that has the face-centered data as well. The way you've set up this frontend to have a nested dataset if the face-centered data exists is intriguing. But it may introduce some confusing bugs -- the yt dataset-related classes rely on many class attributes and loading multiple datasets of the same class at once can lead to errors (e.g., like the one that PR #5159 fixed). So it'd be good if we can thoroughly test that case. 

Usually, we try to get sample datasets uploaded to the yt website so that new frontend tests can utilize the data (usually done via a pull request to https://github.com/yt-project/website like [this one](https://github.com/yt-project/website/pull/129) or [this one](https://github.com/yt-project/website/pull/128), but let me check back with you before you do so. The server that the full test suite runs on is having some issues, so I'm not sure we'll follow the usual protocol right now... at the very least it'd be useful if you could share some datasets so that I can download and run the new test(s) locally.

chongchonghe:
@chrishavlin 

This directory contains face-centered Quokka dataset (`sample/HydroWave`) and a script (`test.py`) to test that:
https://github.com/Rongjun-ANU/README-of-yt-frontend-for-QUOKKA . Running `test.py` using the YT from this PR is successful on my computer.

chongchonghe:
Unfortunately, I couldn't get pytest to work on my computer. I always get this error: 

```
E   ModuleNotFoundError: No module named 'imp'
```

It seems like the answer_testing in yt is using nose, which relies on imp, which is deprecated. However, I run the new tests I added in this PR on my computer as a python script and it worked, so I'm sure the unit tests will pass once we upload the dataset to the server.

chrishavlin:
thanks for the dataset! hopefully will have time to try things out later this week.

chrishavlin:
Hi @chongchonghe thanks for the ping, but unfortunately not yet. Had to bump it to this week's to-do list .

pyproject.toml
```diff
@@ -121,6 +121,7 @@ open-pmd = ["yt[HDF5]"]
 owls = ["yt[HDF5]"]
 owls-subfind = ["yt[HDF5]"]
 parthenon = ["yt[HDF5]"]
+quokka = ["PyYAML>=6.0.1"]
```
chrishavlin, line 124: You can chain dependencies here to avoid defining dependencies in multiple places:

```
quokka = ["yt[amrex]"]
```

Note that if we decide to add PyYAML as a quokka-only dependency (see my main review comment), this line would be

```
quokka = ["yt[amrex]", "PyYAML>=6.0.1"]
```

and you'd need to change the `amrex` line back to:

```
amrex = []
```

chrishavlin (commented):
So I'm still working through the code, but wanted to bring up a high-level question. 

Normally, when frontends introduce dependencies, those dependencies should be added only to the frontend that requires them. But right now you're making PyYaml a dependency of all the amrex subclass frontends. It's a pretty small dependency, so I'm inclined to let it go in this instance, especially since it's already a dependency for a full yt install (via pyaml)... @neutrinoceros do you have an opinion on this? 

For reference, here's a diff that would make PyYaml a dependency only for Quokka (adjusting pyproject.toml, moving the imports to the functions where they're used):

<details>

<summary>Diff is here</summary>

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 03300c836..16a5f9b41 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,7 +89,7 @@ Fortran = ["f90nml>=1.1"]
 # We also normalize all target names to lower case for consistency.
 adaptahop = []
 ahf = []
-amrex = ["PyYAML>=6.0.1"]
+amrex = []
 amrvac = ["yt[Fortran]"]
 art = []
 arepo = ["yt[HDF5]"]
@@ -121,7 +121,7 @@ open-pmd = ["yt[HDF5]"]
 owls = ["yt[HDF5]"]
 owls-subfind = ["yt[HDF5]"]
 parthenon = ["yt[HDF5]"]
-quokka = ["PyYAML>=6.0.1"]
+quokka = ["yt[amrex]", "PyYAML>=6.0.1"]
 ramses = ["yt[Fortran]", "scipy"]
 rockstar = []
 sdf = ["requests>=2.20.0"]
diff --git a/yt/frontends/amrex/data_structures.py b/yt/frontends/amrex/data_structures.py
index 7a27f4efb..4e584de6b 100644
--- a/yt/frontends/amrex/data_structures.py
+++ b/yt/frontends/amrex/data_structures.py
@@ -6,7 +6,6 @@ from functools import cached_property
 from stat import ST_CTIME
 
 import numpy as np
-import yaml
 
 from yt.data_objects.index_subobjects.grid_patch import AMRGridPatch
 from yt.data_objects.static_output import Dataset
@@ -1373,6 +1372,7 @@ class QuokkaDataset(AMReXDataset):
     _field_info_class = QuokkaFieldInfo
     _subtype_keyword = ""
     _default_cparam_filename = "metadata.yaml"
+    _load_requirements = ["yaml"]
 
     def __init__(
         self,
@@ -1465,6 +1465,8 @@ class QuokkaDataset(AMReXDataset):
                 mylog.debug(f"No face-centered {direction} dataset found at {fc_dir}")
 
     def _parse_parameter_file(self):
+        import yaml
+
         # Call parent method to initialize core setup by yt
         super()._parse_parameter_file()
 
@@ -1686,6 +1688,8 @@ class QuokkaDataset(AMReXDataset):
         mylog.debug("Parsed header parameters: %s", self.parameters)
 
     def _parse_metadata_file(self):
+        import yaml
+
         # Construct the full path to the metadata file
         metadata_filename = os.path.join(self.output_dir, self.cparam_filename)
         try:
```

</details>

chrishavlin:
> Now I'm in favor of committing your patch :) 

Hah, ok, then let's go with that. @chongchonghe , could you update the code so that `pyyaml` is only needed for quokka (check the diff in my previous comment)?

pyproject.toml
chongchonghe: Thanks. Done.

chongchonghe:
@chrishavlin I've implemented the changes as you suggested. Thanks a lot!

I have also been trying to optimize how we deal with face-centering variables. I came to the conclusion that the current approach is the optimal approach. QUOKKA does not currently support writing FC variables as 'raw' data as WarpX does. So, I did a hack by loading the face-centered quantities as if they are cell-centered. Raw and "nodal" fields will be supported in the future. 

Sorry for my delayed response. I've been overwhelmed with a proposal recently. Please let me know if any revisions are required, or if you need any additional information from me before we can proceed with merging this PR.

neutrinoceros:
I am not currently available to review a large PR as this one, I apologize for the inconvenience.

yt/frontends/amrex/tests/test_outputs.py
```diff
@@ -445,6 +446,96 @@ def test_maestro_parameters():
     assert type(ds.parameters["s0_interp_type"]) is int  # noqa: E721
 
 
+quokka = "quokka_RadiatingParticles_plt026"
```
chrishavlin: Given how you have https://github.com/Rongjun-ANU/README-of-yt-frontend-for-QUOKKA/tree/main set up, it will be easier to just use paths here: 

```suggestion
quokka = "quokka/RadiatingParticles/plt026"
```

Then the `sample` folder from https://github.com/Rongjun-ANU/README-of-yt-frontend-for-QUOKKA/tree/main can be copied to a `quokka` directory within yt's test data directory (rather than renaming, e.g., `sample/RadiatingParticles/plt026` to `sample_RadiatingParticles_plt026`).

yt/frontends/amrex/tests/test_outputs.py
```diff
@@ -445,6 +446,96 @@ def test_maestro_parameters():
...
+    particle_info = ds.parameters["particle_info"]["Rad_particles"]
+    assert particle_info["num_particles"] == 2
+    assert particle_info["num_fields"] == 3
+    assert set(particle_info["fields"]) == {"luminosity", "birth_time", "end_time"}
+    assert particle_info["units"]["luminosity"] == "M^1 L^2 T^-3"
+    assert particle_info["units"]["birth_time"] == "T^1"
+    assert particle_info["units"]["end_time"] == "T^1"
+
+
+quokka_face_centered = "quokka_HydroWave_plt00004"
```
chrishavlin: see previous comment. 

```suggestion
quokka_face_centered = "quokka/HydroWave/plt00004"
```

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1329,414 @@ def _set_code_unit_attributes(self):
...
+                # Read particle data for this type
+                self._read_particles(ptype, is_checkpoint, extra_fields)
+
+
 class QuokkaDataset(AMReXDataset):
-    # match any plotfiles that have a metadata.yaml file in the root
+    # Match any plotfiles that have a metadata.yaml file in the root
+    _index_class = QuokkaHierarchy
+    _field_info_class = QuokkaFieldInfo
     _subtype_keyword = ""
```
chrishavlin, line 1378: Both of the test files you provided contain a `quokka_version` entry in the metadata yaml. If this is expected to be true of all quokka files, you should change this line:

 ```suggestion
    _subtype_keyword = "quokka"
```

I confirmed that your test datasets still load.

File: yt/frontends/amrex/data_structures.py [hunk 6]
```diff
@@ -1327,10 +1329,414 @@ def _set_code_unit_attributes(self):
...
+
+            # expecting mandatory fields: gasDensity, x-GasMomentum, y-GasMomentum, z-GasMomentum, gasEnergy, gasInternalEnergy
+            # Number of fields
+            num_fields = int(f.readline().strip())
+            self.parameters["fields"] = []
+
+            # Metadata flags
+            rad_group_count = 0
+
+            for _i in range(num_fields):
```
chrishavlin: More conventional to just use `_` here: 

```suggestion
            for _ in range(num_fields):
```

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1329,414 @@ def _set_code_unit_attributes(self):
...
+                }
+
+        # Update parameters with particle info
+        self.parameters.update(
+            {
+                "particles": len(detected_particle_types),
+                "particle_types": tuple(detected_particle_types),
+                "particle_info": particle_info,
+            }
+        )
```
chrishavlin: Was there a particular reason for using `update` here? It's a tad roundabout to create a dict just to use update when the following will work just as well: 

```suggestion
        self.parameters["particles"] = len(detected_particle_types)
        self.parameters["particle_types"] = tuple(detected_particle_types),
        self.parameters["particle_info"] = particle_info
```

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1329,414 @@ def _set_code_unit_attributes(self):
...
+                quokka_version = metadata.get("quokka_version", "0.0")
+
+                # Define version comparison function
+                def is_newer_or_equal(ver, compare_to):
+                    return tuple(map(int, str(ver).split("."))) >= tuple(
+                        map(int, str(compare_to).split("."))
+                    )
+
+                # Compare with version 25.03 (year 2025, month 3)
+                if not is_newer_or_equal(quokka_version, "25.03"):
```
chrishavlin: yt has `packaging` as a minimal dependency, you can use it here to simplify this logic: 
 
```suggestion
                from packaging.version import Version
                quokka_version = Version(metadata.get("quokka_version", "0.0"))                
                if quokka_version < Version("25.03"):
```

chrishavlin: (you can move that `from packaging.version import Version` import to the top of this file: I put it in the suggestion here for clarity)

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1329,414 @@ def _set_code_unit_attributes(self):
...
+                            for unit_name, unit_value in value.items():
+                                self.parameters[unit_name] = unit_value
+                        elif key == "constants":
+                            # Ignore constants section
+                            continue
+                        else:
+                            # For everything else, read in directly
+                            self.parameters[key] = value
+
+                    mylog.debug("Metadata loaded successfully (version 25.03 or newer)")
```
chrishavlin: Might as well put the actual version in here
```suggestion
                    mylog.debug(f"Metadata loaded successfully (version 25.03 or newer): {quokka_version}")
```

chrishavlin (commented):
Thanks for your patience @chongchonghe ! I have some minor suggestions here to clean up some of the new code. I'll also see if anyone else is available for a quick review since we like to have at least 2 reviews for new frontends: but I recognize this has been around for quite a while and it'd be nice to get it merged ASAP. 

And I did download the test data and confirmed that the new tests run successfully. I'll make a companion PR to the yt website repo with the data in a convenient form so it ends up where it needs to eventually, but note that the tests won't actually run for now while yt's continuous integration server is down.

yt/frontends/amrex/data_structures.py
chrishavlin: after getting through the rest of the code, I saw that there may be cases where `quokka_version` is not defined... so you can ignore this suggestion. It'd be preferable if there were a way to identify quokka files beyond just the presence of a Header and metadata.yaml file, but if that's not possible it's OK.

chongchonghe:
> hey @chongchonghe , i'm drumming up support for a yt feature release (version 4.5.0) in January, it'd be great to get this merged so it's included in the release! Do you mind fixing the merge conflict and checking out my last batch of comments? they're pretty minor and should be quick. 

Sure. I'll do it today.

chongchonghe:
> hey @chongchonghe , i'm drumming up support for a yt feature release (version 4.5.0) in January, it'd be great to get this merged so it's included in the release! Do you mind fixing the merge conflict and checking out my last batch of comments? they're pretty minor and should be quick.

hi @chrishavlin . Sorry for the late response. I have addressed the comments and resolved the merge conflicts. Can you review and approve this PR?

yt/frontends/amrex/fields.py
```diff
@@ -504,6 +504,258 @@ def setup_fluid_fields(self):
...
+                ["momentum_density_z"],
+                r"\rho w",
+            ),
+        ),
+        # Temperature field is not present in early Quokka datasets
+        ("gasTemperature", ("K", ["temperature"], r"T")),
+        # Scalar fields are not always present
+        ("scalar_0", ("", ["scalar_0"], "Scalar 0")),
+        ("scalar_1", ("", ["scalar_1"], "Scalar 1")),
+        ("scalar_2", ("", ["scalar_2"], "Scalar 2")),
```
BenWibking: The number of scalars may be arbitrary. This should use logic similar to that used to set up the number of radiation groups to support an arbitrary number of scalars.

tests/.DS_Store
BenWibking, line 1: This should be removed.

yt/frontends/amrex/data_structures.py
```diff
@@ -6,13 +6,15 @@
...
 import numpy as np
+from packaging.version import Version
 
 from yt.data_objects.index_subobjects.grid_patch import AMRGridPatch
 from yt.data_objects.static_output import Dataset
 from yt.fields.field_info_container import FieldInfoContainer
 from yt.funcs import mylog, setdefaultattr
 from yt.geometry.api import Geometry
 from yt.geometry.grid_geometry_handler import GridIndex
+from yt.loaders import load
```
matthewturk: This may be an annoying ask, but I was wondering if we might be able to get rid of this import by replacing the call to `yt.load` below (just `load` because of the import) with a direct instantiation of the face-centered dataset base class, which I believe is defined within this file..

I think this might have extremely small speedups, but a big advantage would be avoiding the possibility (real or imagined) of circular imports.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
 
 
+class QuokkaHierarchy(BoxlibHierarchy):
+    def __init__(self, ds, dataset_type="quokka_native"):
+        super().__init__(ds, dataset_type)
+
+        # Dynamically detect particle types by searching for "*_particles" directories
+        particle_dirs = glob.glob(os.path.join(self.ds.output_dir, "*_particles"))
+
+        for pdir in particle_dirs:
```
matthewturk, line 1344:
```suggestion
        for pdir in sorted(particle_dirs):
```

To be safe -- I believe `glob` still is not guaranteed identical results.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+                    "particle_velocity_z",
+                ]
+
+                is_checkpoint = False
+
+                # Pass the fields to _read_particles
+                extra_fields = quokka_extra_real_fields[: ds.dimensionality] + fields
+
+                # Read particle data for this type
+                self._read_particles(ptype, is_checkpoint, extra_fields)
```
matthewturk:
```suggestion

                # Pass the fields to _read_particles
                extra_fields = quokka_extra_real_fields[: ds.dimensionality] + fields

                # Read particle data for this type
                self._read_particles(ptype, is_checkpoint = False, extra_fields = extra_fields)
```

Is there a possibility that `is_checkpoint` will ever be true in this case?  If not, we could wrap it in here and keep the explicit nature of the specification and also reduce assignment.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+
+        try:
+            out_id = int(os.path.basename(self.output_dir)[-5:])
+        except ValueError as err:
+            raise ValueError(
+                f"Output directory {self.output_dir} does not end with at least 5 digits. Cannot infer output index. Cannot load face-centered datasets."
+            ) from err
+
+        # Define the possible face-centered directories
+        fc_dirs = {
```
matthewturk, line 1438: I know it's just a little thing, and I can't put my finger on why, but something a bout the code that assigns to `fc_dirs` just made me smile.  It just looks really appealing to my eye.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+            "x": os.path.join(fc_dir, f"x{out_id:05d}"),
+            "y": os.path.join(fc_dir, f"y{out_id:05d}"),
+            "z": os.path.join(fc_dir, f"z{out_id:05d}"),
+        }
+
+        # For each direction, check if the directory exists and load it
+        for direction, fc_dir in fc_dirs.items():
+            if os.path.isdir(fc_dir):
```
matthewturk: @neutrinoceros may have different style opinions, but my preference for cases like this is usually to do:

```
if not os.path.isdir(fc_dir):
    mylog.debug(f"No face-centered {direction} dataset found at {fc_dir}")
    continue
```

and then continue the rest of the loop at one level reduced in indentation.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+
+                    # Store the dataset as an attribute
+                    setattr(self, f"ds_fc_{direction}", fc_ds)
+
+                    # Add a reference back to the parent dataset
+                    fc_ds.parent_ds = self
+
+                    # Add information about this being a face-centered dataset
+                    fc_ds.fc_direction = direction
+                except Exception as e:
```
matthewturk: Can we have some kind of flag here that re-raises the exception if it's set?  It can be set in the file, just a `DEBUG_FC_LOADING` or something, that would let people dig in if they are running into errors.  It might be defined at the top level or in the `constants.py` file.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+                    mylog.warning(
+                        f"Failed to load face-centered {direction} dataset: {e}"
+                    )
+            else:
+                mylog.debug(f"No face-centered {direction} dataset found at {fc_dir}")
+
+    def _parse_parameter_file(self):
+        import yaml
```
matthewturk: I think we have an `on_demand_imports` for `yaml`, so you can put at the top of the file:

`from yt.utilities.on_demand_imports import _yaml as yaml`

and just use `yaml` here.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+            )
+
+            # Skip empty line
+            f.readline()
+
+            # Grid info
+            self.parameters["grid_info"] = f.readline().strip()
+
+            # Timestamp
+            self.parameters["timestamp"] = list(map(int, f.readline().strip().split()))
```
matthewturk:
```suggestion
            self.parameters["timestamp"] = [int(_) for _ in f.readline().strip().split()]
```

Can you verify that this works correctly?  I'm also not 100% sure if `strip()` is necessary here, since I think split might take into account the specifics of what it's addressing.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+            # Metadata flags
+            rad_group_count = 0
+
+            for _ in range(num_fields):
+                current_field = f.readline().strip()
+                if current_field.startswith("radEnergy-Group"):
+                    rad_group_count += 1
+                self.parameters["fields"].append(current_field)
+
+            # Add metadata for radiation groups, scalars, and field existence flags
```
matthewturk, line 1484: I don't want to propose changes to this section, but if it grows with additional parameters, we may want to set up a parameter-mapping like:

```
parameters = [ ("dimensionality", int), ... ]
for (param_name, type_map) in parameters:
    self.parameters[param_name] = type_map(f.readline())
```

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+            self.parameters["grid_info"] = f.readline().strip()
+
+            # Timestamp
+            self.parameters["timestamp"] = list(map(int, f.readline().strip().split()))
+
+            # Grid sizes for all refinement levels
+            self.parameters["grid_sizes"] = []
+            for _ in range(self.parameters["refinement_level"] + 1):
+                self.parameters["grid_sizes"].append(
+                    list(map(float, f.readline().strip().split()))
```
matthewturk: I think this might benefit from a list comprehension, but also we should avoid the temptation to play too much code golf (difficult for me to do sometimes) by making it a nested list comp. 😁

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+                # Default field names and units
+                field_names = fields[:]
+                field_units = dict.fromkeys(fields, "dimensionless")
+
+                # Initialize variables
+                yaml_field_names = None
+
+                # Check and parse Fields.yaml
+                if os.path.exists(fields_yaml_file):
+                    try:
```
matthewturk: I *think* that the try/except could get away with just wrapping `yaml.safe_load`.  That would decrease the indent a fair bit and would keep us from handling the exception too far from the place the exception gets raised.  I also think this should check for a debug flag like the face-centered datasets -- maybe justh ave a single `DEBUG_QUOKKA` flag.

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+
+        # Set hydro method explicitly
+        self.parameters["HydroMethod"] = "Quokka"
+
+        # Debug output for verification
+        mylog.debug("Header loaded successfully")
+        mylog.debug("Parsed header parameters: %s", self.parameters)
+
+    def _parse_metadata_file(self):
+        import yaml
```
matthewturk: same here with `on_demand_imports`

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1330,404 @@ def _set_code_unit_attributes(self):
...
+
+        # Construct the full path to the metadata file
+        metadata_filename = os.path.join(self.output_dir, self.cparam_filename)
+        try:
+            with open(metadata_filename) as f:
```
matthewturk: same with the raising and with the `try`/`except` block

matthewturk (commented):
This is outstanding work.  Thank you -- and I'm sorry that I didn't review this much sooner.  I have some comments, and while I'd like to see them applied, I don't believe any rise to the level of blockers.

I do think we'll need to get the `uv.lock` situation fixed, but I'm not 100% sure if that's trivial or not.

chongchonghe:
pre-commit.ci autofix

yt/frontends/amrex/data_structures.py
chongchonghe: Done.

yt/frontends/amrex/fields.py
chongchonghe: I have replaced it with `self.setup_scalar_fields()`

yt/frontends/amrex/data_structures.py
```diff
@@ -1373,7 +1378,6 @@ class QuokkaDataset(AMReXDataset):
     _field_info_class = QuokkaFieldInfo
     _subtype_keyword = ""
     _default_cparam_filename = "metadata.yaml"
-    _load_requirements = ["yaml"]
```
neutrinoceros: why is this removed ?

neutrinoceros:
given that at least one file (`.DS_Store`) was accidentally added then removed, and we don't want it to live forever in yt's commit history, please squash your branch history with `git rebase`.

yt/frontends/amrex/data_structures.py
chongchonghe: ok, i use a `ytcfg.get("yt", "quokka_debug")` . i think that is cleaner than a global variable DEBUG_QUOKKA

yt/frontends/amrex/data_structures.py
chongchonghe: ok, i use a ytcfg.get("yt", "quokka_debug") . i think that is cleaner than a global variable DEBUG_QUOKKA

yt/frontends/amrex/data_structures.py
chongchonghe: i have cleaned up the code and now it looks cleaner

yt/frontends/amrex/data_structures.py
chongchonghe: simiarlly resolved by ytcfg.get("yt", "quokka_debug")

yt/frontends/amrex/data_structures.py
chongchonghe: OK, i have replaced it with `QuokkaDataset`, which is a recursive class declaration, but it should be safe because it does a folder structure check.

yt/frontends/amrex/data_structures.py
chongchonghe: that's a mistake. added it back

neutrinoceros:
> I thought that if we delete a branch after Squash and Merge to main branch, it will delete the unused files from git history. Is this not true?

that is correct, however it's very easy for maintainers to forget using Squash-rebase when the time comes so I'd rather we do a rebase now.

yt/frontends/amrex/data_structures.py
matthewturk: Oh geez, I misread and should have realized that `QuokkaDataset` is calling the `load`.  If you make another change, go ahead and put a comment saying that if it ever causes recursion issues to undo that?  I apologize.

yt/frontends/amrex/data_structures.py
matthewturk: Thank you!

chongchonghe:
OK, done. All squashed into a single commit.

matthewturk (dismissed):
I approved!

chongchonghe:
> I think that both the failures on 3.10 and the conflict on uv.lock are symptoms of a single problem, which you _should_ be able to resolve by rebasing on the current state of main. I would advise dropping the commit on uv.lock in the process, as I think it's slightly malformed (`--exclude-newer` wasn't used). I can fix this for you (including the entire rebase), though I feel uncomfortable force pushing to your branch without so much as a warning, so I'll give you a couple days at least to respond and/or do the rebase yourself.

Please feel free to do any kind of uv lock tweaks and push to this PR. I don't see any problem for a maintainer to do this kind of work to make the PR comply with the code requirement. I would also appreciate a contributing guidelines or a PRD template that explains the uv lock trick.

yt/config.py
```diff
@@ -43,6 +43,7 @@
     "supp_data_dir": "/does/not/exist",
     "default_colormap": "cmyt.arbre",
     "ray_tracing_engine": "yt",
+    "quokka_debug": False,
```
neutrinoceros: There's no precedent for this

```suggestion
```

yt/frontends/amrex/data_structures.py
```diff
@@ -1327,10 +1333,385 @@ def _set_code_unit_attributes(self):
...
+        mylog.debug("Updated ds.parameters with particle info: %s", self.parameters)
+
+        # Set hydro method explicitly
+        self.parameters["HydroMethod"] = "Quokka"
+
+        # Debug output for verification
+        mylog.debug("Header loaded successfully")
+        mylog.debug("Parsed header parameters: %s", self.parameters)
+
+    def _parse_metadata_file(self):
```
neutrinoceros, line 1657: the way exception are handled in this function seems highly unmaintainable:
- exception shouldn't be passed to the logger, they should actually cause the program to stop
- we don't use `print`, ever

but more generally, `try/except` makes debugging much harder, so I would suggest just avoiding it and let exceptions actually raise. Hiding them under a (possibly silent) logger isn't doing anyone any favors.

neutrinoceros (dismissed):
I finally pulled off this difficult rebase. For whatever reason, my fork seems to be stuck behind, but I don't think I need to fix this right now.
I ran a 2-min review and found patterns I'd really like to avoid having on stable.

yt/config.py
chongchonghe: OK, i replaced it with `mylog.isEnabledFor(logging.DEBUG)`

yt/frontends/amrex/data_structures.py
chongchonghe: OK, this makes sense. I've remove try/except and use proper error throw

chongchonghe:
@neutrinoceros @matthewturk I addressed the comments, cleaned up some code, and tested it locally. Everything looks fine. Can you review and approve again?

matthewturk:
I think this is good to go!

matthewturk (dismissed):

chongchonghe:
@matthewturk Thank! I made a minor change to adopt to a recent update on particle component names in quokka. Can you approve again? Also, is there anything I need to do before you merge it?

BenWibking:
We are preparing to release Quokka 26.04 in the next few days, and would like to mention the addition of a full-featured frontend in yt to our users. If it's possible to re-review and merge this in that timeframe, please let us know.

neutrinoceros:
In order to harmonize expectations, I'm letting you know I don't have time to review this properly in the forseeable future, so I'm removing myself from review requests here. I trust other maintainers to make sure my existing comments were actually addressed before merging.

chongchonghe:
@matthewturk @chrishavlin Can you approve and merge this PR if there is no further request of change?

chrishavlin:
>I trust other maintainers to make sure my existing comments were actually addressed before merging. 

@neutrinoceros from my quick look at the end of last week, the comments were addressed. I'll double check now...

chrishavlin (approved):
Thanks for all your hard work and patience on this! All the comments are now addressed.

yt/config.py
chrishavlin: resolved! yt.config is no longer being used here.

chrishavlin:
Ok! let's do this... thanks again @chongchonghe @BenWibking @matthewturk @neutrinoceros !

PR #5169, opened 2025-05-20
[BUG] RAMSES: temperature units are incorrect

cphyc:
## PR Summary
For RAMSES, temperatures are returned in T/µ, so it should not be included in the definition of the temperature unit.

To be clear, this does *not* affect the values of the temperature fields read from disk. However, should one read from disk a field akin to a temperature and assume it is expressed in `code_temperature`, there will be a factor 0.76 off.

## PR Checklist

### DIFF
--- a/yt/frontends/ramses/data_structures.py
+++ b/yt/frontends/ramses/data_structures.py
@@ -29,7 +29,7 @@ from .definitions import (
     ramses_header,
 )
 from .field_handlers import get_field_handlers
-from .fields import _X, RAMSESFieldInfo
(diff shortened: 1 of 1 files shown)


cphyc:
I'm not sure how to test this - but the introduction of the factor 0.76 comes from some comment by Sam Geen 11 years ago, and has been implemented by Nathan back then. I couldn't find a trace of the comments tho, so not sure what this pertained to.

chrishavlin:
ya, i'm just gonna merge :)

PR #5170, opened 2025-05-21
Adds support for Cholla's new Concatenation Format

mabruzzo:
## PR Summary

This PR adds support for loading Cholla datasets that are:
- distributed over multiple files
- stored in Cholla's historical concatenation datasets

For context:
- when Cholla writes out snapshots, it writes each block to a separate file.
- The Cholla repository provide python scripts to concatenate data, which has historically concatenate all of the data so that the resulting file would appear like there was only 1 process that evolved a single gigantic block. These scripts were commonly used to simplify data analysis.
- Until this PR, the yt-frontend has **only** known how to load datasets after they were concatenated in this manner. This introduced challenges because:
  1. the yt-frontend must load everything at once (introducing obvious memory issues)
  2. in cases where the number of cells exceeds 2^31 yt would simply break (this came up on large-memory nodes)

This PR introduces support for reading Cholla datasets that are either:
- distributed (i.e. they are not concatenated), OR
- use Cholla's newly added concatenation format.
  - I introduced the new concatenation format with the intent of improving performance when using yt (cholla-hydro/cholla#427). Essentially the new format organizes data in a manner based on the underlying block structure (in a similar manner to what an AMR simulation might do). This let's us reuse a lot of the machinery for reading distributed datasets
  - It's important that the yt-frontend can read in this format because (i) it provides some performance advantages (on parallel filesystems) and (ii) we have scripts to repackage datasets using the old concatenation format so that they now have this new format.

Aside: This PR superseded PR #4702 

## PR Checklist


yt/frontends/cholla/data_structures.py
```diff
@@ -9,16 +13,193 @@
...
 
+# this is a hacky workaround to get _h5py.File to work in annotations. We can probably
+# address this issue more robustly by directly modifying yt.utilities.on_demand_imports
+if typing.TYPE_CHECKING:
+    import h5py as _h5py
+else:
+    from yt.utilities.on_demand_imports import _h5py
+
+
+@dataclass(kw_only=True, slots=True)
```
neutrinoceros: hopefully that makes sense here (mutability makes code much harder to reason about, I'd like to forbid it from the start in new APIs)
```suggestion
@dataclass(kw_only=True, slots=True, frozen=True)
```

yt/frontends/cholla/data_structures.py
```diff
@@ -9,16 +13,193 @@
...
+
+    # ``fname_template.format(blockid=...)`` produces the file containing blockid (this
+    # can properly handle cases where all blocks are stored in a single file)
+    fname_template: str
+    # group containing field data (empty string denotes the root group)
+    field_group: str
+    # maps blockid to an index that select all associated data from a field-dataset
+    field_idx_map: Mapping[int, tuple[int | slice, ...]]
+
+    # in the future, we will add particle_group and particle_idx_map
```
neutrinoceros: I strongly advise against making any binding promises in comments: some become soul-crushing after a while.
```suggestion
    # in the future, we might add particle_group and particle_idx_map
```

yt/frontends/cholla/data_structures.py
```diff
@@ -9,16 +13,193 @@
...
+    if has_explicit_domain_info:
+        # this branch primarily handles concatenated files made with newer logic
+        blockid_location_arr = f["domain/blockid_location_arr"][...]
+        field_idx_map = {
+            int(blockid): (i, slice(None), slice(None), slice(None))
+            for i, blockid in enumerate(f["domain/stored_blockid_list"][...])
+        }
+        consolidated_data = len(field_idx_map) == blockid_location_arr.size
+        if not consolidated_data:
+            # in the near future, we will support one of the 2 cases:
```
neutrinoceros: same here

yt/frontends/cholla/data_structures.py
```diff
@@ -9,16 +13,193 @@
...
+      while this function returns ``("root.h5", "3")``
+    - For "root.h5": os.path.splitext`` returns ``("root", ".h5")``, while this
+      function returns ``("root.h5", "")``
+    """
+
+    # at this time, we expect the suffix to be the minimum number of characters
+    # that are necessary to represent the process id. For flexibility, we will
+    # allow extra zero-padding
+
+    sep_i = filename.rfind(".")
```
neutrinoceros: I think this function could be much more readable if it was written around `str.rsplit` instead of `str.rfind`

neutrinoceros: or better yet: `str.rpartition`

yt/frontends/cholla/data_structures.py
mabruzzo: Good point. I very much agree!

My original source of hesitation was the fact that `field_idx_map` was originally annotated as a `dict`, which is mutable. But, now that I changed the annotation to `Mapping` (since it can hold a `defaultdict` in some cases), I now completely agree

yt/frontends/cholla/io.py
```diff
@@ -14,22 +12,27 @@ def _read_particle_coords(self, chunks, ptf):
...
-                        ftype, fname = field
-                        values = fh[fname][:].astype("=f8")
-                        nd = grid.select(selector, values, data[field], ind)
-                    ind += nd
-        return data
+    def io_iter(self, chunks, fields):
+        # this is loosely inspired by the implementation used for Enzo/Enzo-E
+        # - those other options use the lower-level hdf5 interface. Unclear
+        #   whether that affords any advantages...
+        fh, filename = None, None
```
neutrinoceros:
```suggestion
        fh = None
        filename = None
```

yt/frontends/cholla/io.py
```diff
@@ -14,22 +12,27 @@ def _read_particle_coords(self, chunks, ptf):
...
+        mapper = self.ds.index._block_mapping
+        for chunk in chunks:
+            for obj in chunk.objs:
+                if obj.filename is None:  # unclear when this case arises...
+                    continue
+                elif obj.filename != filename:
```
neutrinoceros: (nit)
```suggestion
                if obj.filename != filename:
```

yt/frontends/cholla/data_structures.py
mabruzzo: good point -- fixed

yt/frontends/cholla/data_structures.py
mabruzzo: fixed

yt/frontends/cholla/io.py
mabruzzo: fixed

yt/frontends/cholla/data_structures.py
mabruzzo: fixed

mabruzzo:
@neutrinoceros, @matthewturk -- I just wanted to ping you to see if there is anything I can do to make this PR easier to review. I would love to see this get merged in time for the 4.5 release (whenever that is)

matthewturk:
Should we merge this?

mabruzzo:
That would be great!

mabruzzo:
@matthewturk and @neutrinoceros, I just wanted to ping you guys in case this fell off the todo list.

neutrinoceros:
Sorry, this is actually not part of my TODO list. I'm way behind on my backlog and I cannot predict how long it'll take me to give this the attention it needs

PR #5182, opened 2025-06-11
WHL: upgrade cibuildwheel to 3.0.0

neutrinoceros:
## PR Summary
This upgrade is necessary to enable building and testing against CPython 3.14 later this year.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -461,12 +461,11 @@ show_error_context = true
 exclude = "(test_*|lodgeit)"
 
 [tool.cibuildwheel]
-build = "cp310-* cp311-* cp312-* cp313-*"
 build-verbosity = 1
 test-skip = "*-musllinux*"
 test-extras = "test"
 test-command = [
-    "pytest -c {project}/pyproject.toml --rootdir . --color=yes --pyargs yt -ra",
+    "python -m pytest -c {project}/pyproject.toml --rootdir . --color=yes --pyargs yt -ra",
 ]
 
 [tool.cibuildwheel.linux]
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Build wheels for CPython
-        uses: pypa/cibuildwheel@v2.23.3
+        uses: pypa/cibuildwheel@v3.0.0
         with:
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -461,12 +461,11 @@ show_error_context = true
 exclude = "(test_*|lodgeit)"
 
 [tool.cibuildwheel]
-build = "cp310-* cp311-* cp312-* cp313-*"
```
neutrinoceros, line 464: this line isn't needed anymore: previously we used it to *un*select pypy builds, but they are now disabled by default, and cibuildwheel is smart enough to guess our supported range automatically.

pyproject.toml
```diff
@@ -461,12 +461,11 @@ show_error_context = true
...
 build-verbosity = 1
 test-skip = "*-musllinux*"
 test-extras = "test"
 test-command = [
-    "pytest -c {project}/pyproject.toml --rootdir . --color=yes --pyargs yt -ra",
+    "python -m pytest -c {project}/pyproject.toml --rootdir . --color=yes --pyargs yt -ra",
```
neutrinoceros, line 468: calling pytest directly still works, but the command is replaced at runtime with `python -m pytest` with a warning, so, might as well make it more explicit and less noisy

neutrinoceros:
note that numpy just similarly upgraded `manylinux`, as will probably the rest of the scientific ecosystem in a near future

chrishavlin:
> So we went from manylinux_2_17_x86_64.manylinux2014_x86_64 to manylinux_2_24_x86_64.manylinux_2_28_x86_64. Is there any objection to just follow along the change in default image, i.e., should we pin manylinux_2_17 for as long as it works ?

I'd like to hear what others think, cause I'm only basing this on all the reading I did just now about manylinux wheels so could be misinterpreting... but IMO it should be OK to follow the default image. This particular jump seems pretty conservative: glibc 2.24 was released in 2016, hard to imagine there is anyone trying to use the latest yt who haven't upgraded their OS to something with glibc>=2.24 since then (e.g., if I'm reading release notes right, we'd be dropping wheels support for Ubuntu <= 16.04.7?).  So ya, +1 for following the default image.

chrishavlin (approved):

neutrinoceros:
No objection was raised, so let's just get this in for 4.4.1

PR #5184, opened 2025-06-13
Enhancement: `create_profile` no longer computes extrema when we provide `override_bins`

mabruzzo:
## PR Summary

The current implementation of `create_profile` computes extrema when `override_bins` is provided. This can significantly increase the cost of an operation when working with large datasets. This PR addresses that scenario.

The PR also refactors argument handling of `create_profile` to be more explicit for `extrema` and `override_bins`. The details of these arguments have a number of undocumented idiosyncracies. While the code is more verbose, I think it makes the requirements a little more explicit (and the error messages are more detailed). In the future, I think we can improve this a lot more, but this seems like a reasonable first-step.

It may be easiest to review this PR commit-by-commit.

## PR Checklist

The PR Checklist isn't applicable since I haven't changed any user-facing behavior


yt/data_objects/profiles.py
```diff
@@ -1300,12 +1338,35 @@ def create_profile(
...
+        # when extrema are all Nones, we treat them as though extrema was set as None.
+        #
+        # In the future, it may be better to replace `not any(...)` with either:
+        #   `all(v is None for v in collapse(extrema.values()))` OR
+        #   `all(v is None or v == (None, None) for v in collapse(extrema.values))`
+        #
+        # Be aware that doing this changes behavior, when extrema has the value:
+        #   `{<bin_field>: (0., None)}` or `{<bin_field>: (None, 0.)}`
+        # I happen to think think that the existing behavior is a bug (i.e. the
+        # behavior changes if we replaced 0. with **ANY** other number)
```
mabruzzo, line 1371: Note to reviewers: This is mostly for posterity. I don't think we should deal with this issue in this PR (since this PR is focused on not changing observable behavior). I plan to make an issue for this after the PR is reviewed

---------

I think we should probably change the if-statement to:

```python
    if extrema is None or all(v is None for v in collapse(extrema.values())):
```
OR
```python
    if extrema is None or all(list(v) == [None, None] for v in extrema.values()):
```

As the comments mention, both choices technically change the behavior when one of the extrema is a `0` and the other is a `None`, but I think the existing behavior is probably a bug. The former option is more backwards compatible while the latter option is more consistent with the rest of our extrema handling.

For more context, the following table indicates whether various choices for the extrema argument are considered valid (✅) or invalid (❌), when bin_fields is `[("gas", "velocity_x"), ("gas", "density")]`


|                       | Current Logic  | Snippet 1 | Snippet 2 |
| ------------- | ------------- | ------------- | ------------- |
| `{("gas","velocity_x"): (0.0, None)}` | ️✅* | ❌ | ❌ |
| `{("gas","velocity_x"): None, ("gas", "density"): None}` | ️✅ | ️✅ | ❌ |
| `{("gas","velocity_x"): (None, None), ("gas", "density"): (None, None)}` | ️✅ | ️✅ | ️✅ |
| `{("gas","velocity_x"): (0.0, None), ("gas", "density"): (None, None)}` | ️✅* | ❌ | ❌ |
| `{("gas","velocity_x"): None, ("gas", "density"): (1e-27, 1e-23)}` | ❌ | ❌ | ❌ |
| `{("gas","velocity_x"): (None,None), ("gas", "density"): (1e-27, 1e-23)}`  | ️✅ | ️✅ | ️✅ |

**Importantly:** The starred cases are scenarios that either of the snippets would "fix." In those cases, the  existing logic totally ignores the fact that the caller specifies that the lower extrema for `("gas","velocity")` as 0.0 and acts like the lower extrema was set to `None`. Furthermore, the existing logic causes the function to behave completely differently for the following cases:
- `{("gas","velocity_x"): (-1.0, None)}` (this is presently considered to be an invalid choice)
- `{("gas","velocity_x"): (1.0, None), ("gas", "density"): (None, None)}`
- `{("gas","velocity_x"): (0.0, None), ("gas", "density"): (None, 1e-20)}`.

matthewturk:
@mabruzzo I just had this bumped when you closed it and looked it over and really like it.  I am sorry I let it drop; do you think that this is not worth pursuing?  I think it might actually be pretty ready to go in.

mabruzzo:
@matthewturk, I definitely think that this would be a worthwhile change. In my view it was definitely an improvement.

I closed this because I have 3 other year-old open PRs that are ready to be reviewed and this one was marked WIP, so I figured I would try to cut down on the backlog. I honestly didn't look into what I changed in this PR (or why it was marked as WIP). I suspect a test may have been failing. I'll happily reopen it and rerun the tests after merging in recent tests. If they all pass (or it's easy for me to identify/resolve any failures) I'm happy to help get this merged in.

mabruzzo:
@matthewturk, this should now be ready for review.

mabruzzo:
FYI, the failing test seems pretty independent of the proposed changes. It looks like some HTTPS messages failed to fetch some data

PR #5185, opened 2025-06-16
WHL: test musllinux wheels

neutrinoceros:
## PR Summary
If I recall correctly the rationale for not testing these wheels was that NumPy didn't support the platform at the time, but they do now, so let's give it a try.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   build_wheels:
-    name: Build ${{ matrix.archs }} wheels on ${{ matrix.os }}
+    name: Build ${{ matrix.select }}-${{ matrix.archs }} wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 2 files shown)


neutrinoceros:
failure on windows+cp313 is unrelated (upstream bug in Python 3.13.4, fixed in 3.14.5)

neutrinoceros:
wow, this is new
```python-traceback
MemoryError: Parser stack overflowed - Python source too complex to parse
```
I can try reporting this upstream as a regression in Python 3.14, hopefully this can still be addressed before the final release.

neutrinoceros:
Alright ! so the good news is that this error still reproduces with a `cp314-cp314` wheel (as opposed to `cp311-abi3`), so we know for sure it has nothing to do with recent changes in how we build wheels.

neutrinoceros:
tentatively reported upstream as https://github.com/python/cpython/issues/137231

neutrinoceros:
ok this works as intended with Python 3.14.0rc2, now all that's needed is a stable cibuildwheel release that includes it.

PR #5189, opened 2025-06-26
[BUG] minor `typing.ClassVar` fix for class variables of `Index`

mabruzzo:
## PR Summary

**Context:** To annotate the types of class variables, you need to wrap the type using the `typing.ClassVar` construct. If the class-variable has type `T`, the type should be annotated as `typing.ClassVar[T]`. [If you don't use this construct, python assumes that it is an instance variable.](https://docs.python.org/3/library/typing.html#typing.ClassVar).

**Overview:** This PR rectified started using the `typing.ClassVar` construct to wrap the type annotations for the `_unsupported_objects` and `_index_properties` class variables of `Index`.

## A quick aside

A quick grep showed that `GridGeometryHandler` is the only class to actually define `_index_properties`, and there isn't any code that examines this attribute. Can we delete it? If we want to keep it as an attribute of `GridGeometryHandler` for posterity, can we delete class attribute from `Index`. If there is interest in doing this, I will happily open another PR

### DIFF
--- a/yt/geometry/geometry_handler.py
+++ b/yt/geometry/geometry_handler.py
@@ -23,7 +23,6 @@ class Index(ParallelAnalysisInterface, abc.ABC):
     """The base index class"""
 
     _unsupported_objects: ClassVar[tuple[str, ...]] = ()
-    _index_properties: ClassVar[tuple[str, ...]] = ()
 
     def __init__(self, ds, dataset_type):
         ParallelAnalysisInterface.__init__(self)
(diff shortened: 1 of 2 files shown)


neutrinoceros:
> there isn't any code that examines this attribute. Can we delete it? 

yes, I don't see why not, this seems like a clear case of dead code. Thanks for finding it !

matthewturk:
I vaguely remember it having a need at one point and possibly being replaced by something else -- maybe even the supported objects.

PR #5193, opened 2025-06-29
RFC/DEPR: deprecate yt.funcs.levenshtein_distance in favor of difflib.get_close_matches

neutrinoceros:
## PR Summary
I just learned about `difflib.get_close_matches`, and it seems like a nice way to remove a custom function of ours.

## PR Checklist


yt/utilities/exceptions.py
```diff
@@ -97,51 +98,36 @@ def __init__(self, field, ds):
...
-        # A threshold of 3 seems like a good middle point.
-        max_distance = 3
+        suggestions: list[FieldKey] = []
+        if ftype is not None:
+            fields_str: dict[str, FieldKey] = {
+                str(df).lower(): df for df in ds.derived_field_list
+            }
+            field_str = str(field).lower()
+            suggestions.extend(
+                fields_str[k] for k in get_close_matches(field_str, fields_str.keys())
```
cphyc: This is not the same behaviour as before, since `get_close_matches` will only return the first 3 close matches by default. It would be great to yield instead *all* matches 
```suggestion
                fields_str[k] for k in get_close_matches(
                    field_str,
                    fields_str.keys(),
                    n=len(fields_str),
                )
```

Note that this won't be backwards compatible anyway, since the distance metric isn't Levenshtein (but I don't think it matters).

cphyc (changes requested):
We are loosing some relevant suggestions (when there are more than 3 possible matches, try e.g. querying `ad["index", "d"]` (it should suggest `x,y,z,dx,dy,dz` at least. With the current PR, it only suggests the last three.

With the changes, I get more irrelevant (but also more relevant ones) suggestions:
```
ad = ds.all_data()

ad["gas", "densty"]
```

## Previously:
```
YTFieldNotFound: Could not find field ('gas', 'densty') in info_00080.
Did you mean:
	('gas', 'density')
```
## This PR:
```
YTFieldNotFound: Could not find field ('gas', 'densty') in info_00080.
Did you mean:
	('gas', 'density')
	('gas', 'dy')
	('ramses', 'Density')
```

neutrinoceros:
rebased to resolve merge conflicts

chrishavlin:
@cphyc does your prior approval on this still stand? I think it was dismissed just from a merge. 

I'll try to give this a second review soon.

PR #5207, opened 2025-07-16
Read single precision fluid field in RAMSES

AnatoleStorck:
## PR Summary

The current back-end for reading RAMSES fluid fields assumes the data is double precision. This PR (in collaboration with @cphyc ) is an initial expansion of the RAMSES fluid reader to read single precision fields. For my simulations, the RT fields are written in single precision so this change is needed to read them in.

## PR Checklist

### DIFF
--- a/yt/frontends/ramses/field_handlers.py
+++ b/yt/frontends/ramses/field_handlers.py
@@ -254,7 +254,7 @@ class FieldFileHandler(abc.ABC, HandlerMixin):
         
         first_field = self.field_list[0][1]
         # Get the size of the data, single precision if RT field, double otherwise
-        field_size = "single" if first_field.startswith("Photon") else "double"
+        is_single = True if first_field.startswith("Photon") else False
         
         with FortranFile(self.fname) as fd:
             # Skip headers
@@ -282,7 +282,7 @@ class FieldFileHandler(abc.ABC, HandlerMixin):
                 self.parameters[self.ds.unique_identifier]["nvar"],
                 self.domain.amr_header,
                 Nskip=nvars * 8,
-                size=field_size,
+                single=is_single,
             )
 
         self._level_count = level_count
--- a/yt/frontends/ramses/field_handlers.py
+++ b/yt/frontends/ramses/field_handlers.py
@@ -549,10 +549,13 @@ class RTFieldFileHandler(FieldFileHandler):
 
         rheader = {}
 
-        def read_rhs(cast):
+        def read_rhs(cast, group=None):
             line = f.readline()
             p, v = line.split("=")
-            rheader[p.strip()] = cast(v)
+            if group is not None:
+                rheader[f"Group {group} {p.strip()}"] = cast(v)
+            else:
+                rheader[p.strip()] = cast(v)
 
         with open(fname) as f:
             # Read nRTvar, nions, ngroups, iions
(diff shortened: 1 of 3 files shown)


doc/source/examining/loading_data.rst
```diff
@@ -3171,12 +3171,15 @@ There are three way to make yt detect all the particle fields. For example, if y
...
 
    .. code-block:: none
 
       [ramses-grav]
-      fields = [ "Potential", "Potential-old", "x-acceleration", "y-acceleration", "z-acceleration" ]
+      fields = [ "Potential,f", "Potential-old,f", "x-acceleration,f", "y-acceleration,f", "z-acceleration,f" ]
+
+   Importantly, the order of the fields should match the order in which they are written in the RAMSES output files.
+   yt will also assume that each entry is formed of a field name followed by its type (``f`` for single precision, ``d`` for double precision), separated by a comma. Field names containing commas are not supported, other precisions (e.g. integers) are not supported either. All fields in a given file type (gravity, hydro, ...) need to have the same precision.
```
chrishavlin: is it worth a mention that though yt can read in the single precision case, they will be converted to double precision for use within yt?

yt/frontends/ramses/field_handlers.py
```diff
@@ -24,6 +24,61 @@ def register_field_handler(ph):
 DETECTED_FIELDS = {}  # type: ignore
 
 
+def check_field_precision(fields: list[tuple[str, str]]) -> bool:
```
chrishavlin: Since this function will only ever return `True` or throw an error (i.e., it will never return `False`), I'd suggest changing it a bit. Rename it to `validate_field_precision` and remove the bool returns.

yt/frontends/ramses/field_handlers.py
```diff
@@ -24,6 +24,61 @@ def register_field_handler(ph):
...
+    fields : list of (str, str)
+        List of tuples containing field names and their precision
+        ('f' for single, 'd' for double).
+
+    Returns
+    -------
+    bool
+        True if all fields have the same precision, False otherwise.
+    """
+    if not fields:
```
chrishavlin: if you're using this to check for a list of length 0, can you change this to check for `len(fields) == 0` explicitly? if `fields` can actually be `None`, then the typing should be updated to `list[tuple[str, str]] | None`.

yt/frontends/ramses/field_handlers.py
```diff
@@ -24,6 +24,61 @@ def register_field_handler(ph):
...
+        ('f' for single, 'd' for double).
+
+    Returns
+    -------
+    tuple of (list of str, bool)
+        A tuple containing a list of field names and a boolean indicating
+        if all fields are single precision, False if double precision.
+    """
+    field_names = [field for field, _ in fields]
+    check_field_precision(fields)
```
chrishavlin: oh hey, you're already ignoring the output `bool`.

yt/frontends/ramses/field_handlers.py
```diff
@@ -597,23 +696,20 @@ def read_rhs(cast):
...
 
-        if ds._fields_in_file is not None:
-            # Case 1: fields are provided by users on construction of dataset
-            fields = list(ds._fields_in_file)
-            ok = True
-        else:
-            # Case 2: fields are provided by users in the config
-            fields = cls.load_fields_from_yt_config()
-            ok = len(fields) > 0
+        # Are fields are provided by users in the config?
```
chrishavlin:
```suggestion
        # Are fields provided by users in the config?
```

AnatoleStorck:
@cphyc, Looks good to me, and is much more generalized than what I had implemented.

PR #5208, opened 2025-07-16
[feature] Simpler way to add derived fields and to define `cut_region` s

cphyc:
## PR Summary

This adds a simple syntax to define a new derived field.
This allows to write:
```python
ds = yt.load(...)

ds.fields.gas.density_square = ds.fields.gas.density * ds.fields.gas.density
ds.add_field(("gas", "double_density"), ds.fields.gas.density * 2)

ds.r["gas", "density_square"]
ds.r["gas", "double_density"]
```

Open for discussion before I write the documentation (or we kill that PR :) )

## PR Checklist

### DIFF
--- a/yt/fields/derived_field.py
+++ b/yt/fields/derived_field.py
@@ -127,7 +127,9 @@ class DerivedFieldBase(abc.ABC):
     def __gt__(self, other) -> "DerivedFieldCombination":
         return DerivedFieldCombination([self, other], op=operator.gt)
 
-    def __eq__(self, other) -> "DerivedFieldCombination":
+    # NOTE: we need to ignore type checking for these methods
+    # since mypy expects the following two to return `bool`
+    def __eq__(self, other) -> "DerivedFieldCombination":  # type: ignore[override]
         return DerivedFieldCombination([self, other], op=operator.eq)
 
     def __ne__(self, other) -> "DerivedFieldCombination":  # type: ignore[override]


matthewturk:
I genuinely love this and I find it to be an incredibly nice, ergonomic way to add fields.

cphyc:
Quick update: we can now do:
```python
import yt
ds = yt.load("output_00080")

# Quickly define new fields
ds.fields.gas.double_density = ds.fields.gas.density * 2

# Quickly filter on the grid
ad = ds.all_data()
left = ad[ds.fields.gas.x < (0.5, "unitary")]

# Quickly filter particles
left_particles = ds.fields.io.particle_position_x < ds.quan(0.5, "unitary")
ds.add_particle_filter(left_particles)
ad[left_particles.name, "particle_position_x"]
```

cphyc:
Note to reviewers: I'm waiting to get approval (if any :D) before writing the documentation. If you'd like to see more examples, I can write the doc before getting approval though.

yt/data_objects/static_output.py
```diff
@@ -1717,7 +1728,7 @@ def quan(self):
         return self._quan
 
     def add_field(
-        self, name, function, sampling_type, *, force_override=False, **kwargs
+        self, name, function, *, sampling_type=None, force_override=False, **kwargs
```
jzuhone, line 1749: So this is a slightly significant change in behavior, isn't it? We've previously mandated that the `sampling_type` be specified. Any concerns about this? I don't have any, but I wondering if anyone else has any ideas about potential downstream effects.

jzuhone: Never mind, I see how you handle this below now.

yt/fields/derived_field.py
```diff
@@ -58,7 +61,129 @@ def _DeprecatedFieldFunc(field, data):
...
+    def __lt__(self, other) -> "DerivedFieldCombination":
+        return DerivedFieldCombination([self, other], op=operator.lt)
+
+    def __geq__(self, other) -> "DerivedFieldCombination":
+        return DerivedFieldCombination([self, other], op=operator.ge)
+
+    def __gt__(self, other) -> "DerivedFieldCombination":
+        return DerivedFieldCombination([self, other], op=operator.gt)
+
+    # def __eq__(self, other) -> "DerivedFieldCombination":
```
jzuhone: Could you leave a comment as to why this is commented out?

cphyc: Ah good point! I went ahead and implemented this feature.

The annoying point is that defining an `__eq__` method forces you to also implement a `__hash__` method (which I did).
But this is worth it, since now we can write something like

```python
ad = ds.all_data()
ds.fields.index.is_level_0 = ds.fields.index.grid_level == 0
ad["index", "is_level_0"]

low_res_only = ad.cut_region(ds.fields.index.grid_level == 0)
```

cphyc:
I found the source of the failures @chrishavlin was talking about. Here's the log of my little adventure:
- Defining the `__eq__` forces us also to define a `__hash__` function for DerivedFields to be hashable (e.g., to make sets of them, or use them as an entry to dictionaries),
- Since we support the syntax `ds.fields.ftype1.fname1 == ds.fields.ftype2.fname2`, which returns a `DerivedFieldCombination`, we need to evaluate the truthness of this expression, i.e. implement a `__bool__` method on `DerivedFieldCombination`,
- That boolean should be True in all cases (since an object in Python is considered as `True`), unless we're comparing a field to another i.e.
```python
# The following two evaluate to the same boolean
bool(ds.fields.ftype1.fname1 == ds.fields.ftype2.fname2)
ds.fields.ftype1.fname1 is ds.fields.ftype2.fname2
```

doc/source/analyzing/filtering.rst
chrishavlin, line 1: did you mean to include the changes in this file in this PR? they seem unrelated

doc/source/analyzing/mesh_filter.ipynb
chrishavlin, line 1: same question on this file

doc/source/analyzing/filtering.rst
cphyc: Yes - I should actually change the PR's name since it allows defining more easily derived fields _and_ cut regions (using the same syntax!).

The changes in this file reflect this change.

doc/source/analyzing/mesh_filter.ipynb
cphyc: Same as above, this describes how one can create `cut_region` using `ad.cut_region(ds.field.gas.temperature < ds.quan(1e6, "K"))`, which is equivalent to the old syntax `ad.cut_region("obj['gas', 'temperature'] < 1e6")`.

The new one is more verbose but let's obscure (what is `obj`?!)

cphyc (commented):
Thanks for your review @chrishavlin - the two modified files you flagged describe another way to define `cut_region` that we got for free with the derived field definition.

jzuhone:
I'll look at this tomorrow!

PR #5209, opened 2025-07-17
BUG: fix missing `refine_by` attribute in `SwiftDataset`

neutrinoceros:
## PR Summary

Fix an `AttributeError` encountered by Zhen Xiang at the 2025 workshop. I don't have the whole story yet, but from what I gather Swift doesn't have grid refinement so this should be safe ?

## PR Checklist


matthewturk (approved):
I think this is right, yup!  Good catch, and sorry we didn't have this before.

PR #5213, opened 2025-07-18
Represent comoving with a `c` prefix

cphyc:
## PR Summary

This changes the default latex representation of comoving axes from $\mathrm{Mpc}/(1+z)$ to being $\mathrm{cMpc}$, the latter being the convention, while the former - albeit correct - is quite confusing and cumbersome.

Happy to *not* merge this if this is deemed controversial.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1394,7 +1394,7 @@ class Dataset(abc.ABC):
                         new_unit,
                         my_u.base_value / (1 + self.current_redshift),
                         dimensions.length,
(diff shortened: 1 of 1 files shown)


neutrinoceros:
well, easy enough to revert of monkeypatch if anyone complains about it later, so I think this is safe to pull in now. Thanks !

neutrinoceros (approved):

neutrinoceros:
(for reference: I'm denying the "refactor" qualification as this has user-visible implications)

PR #5216, opened 2025-07-18
TST/RFC: migrate a test module from nose to pytest with Copilot

matthewturk:
I have a lot of conflicted feelings about this.

But, we talked at the workshop about some things being (potentially) well-suited to rewrites for things like this.  So I decided to have a go at it, and I asked it to rewrite one of our remaining tests.

I'm submitting it *as-is*, to debate whether it worked, how hard it might need to be worked at to fix things, and also so I can get dunked on by every person I've told "using co-pilot is ethically suspect!" and I can eat my words once and for all.

So what do we think?

### DIFF
--- a/yt/data_objects/tests/test_data_containers.py
+++ b/yt/data_objects/tests/test_data_containers.py
@@ -10,7 +10,7 @@ from yt.testing import (
     fake_amr_ds,
     fake_particle_ds,
     fake_random_ds,
-    requires_module,
+    requires_module_pytest as requires_module,
 )
 from yt.utilities.exceptions import YTException, YTFieldNotFound
 


yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
-        for fname in fields_to_test:
-            test_this(fname)
+@pytest.fixture(scope="module")
+def temp_workdir():
+    tmpdir = tempfile.mkdtemp()
+    curdir = os.getcwd()
+    os.chdir(tmpdir)
+    yield
+    os.chdir(curdir)
+    shutil.rmtree(tmpdir)
```
neutrinoceros: Alas, `contextlib.chdir` requires Python 3.11 ...
```suggestion
def temp_workdir(tmp_path):
    from contextlib import chdir
    with chdir(tmp_path):
    	yield tmp_path
```

but we can still make it better on 3.10
```suggestion
def temp_workdir(tmp_path):
    curdir = os.getcwd()
    os.chdir(tmp_path)
    yield tmp_path
    os.chdir(curdir)
```

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
-
-        fields_to_test = [f for f in ds.derived_field_list if f[0] == "massive"]
-
-        def test_this(fname):
-            data = dd[fname]
-            assert_equal(data.shape[0], expected_size)
-
-        for fname in fields_to_test:
-            test_this(fname)
+@pytest.fixture(scope="module")
```
neutrinoceros: I don't know that we need a scope here ?
```suggestion
@pytest.fixture
```

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
+def temp_workdir():
+    tmpdir = tempfile.mkdtemp()
+    curdir = os.getcwd()
+    os.chdir(tmpdir)
+    yield
+    os.chdir(curdir)
+    shutil.rmtree(tmpdir)
+
+
+def test_yt_data_container(temp_workdir):
```
neutrinoceros: same thing, but clearly indicates we won't use the fixture as a variable name
```suggestion
@pytest.mark.usefixture("temp_workdir")
def test_yt_data_container():
```
(the same thing should be done everywhere possible)

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
+    desired = (
+        "Error: ds must be set either through class"
+        " type or parameter to the constructor"
+    )
+    assert_equal(str(err.value), desired)
+
+    # Test if field_data key exists
+    ds = fake_random_ds(5)
+    proj = ds.proj(("gas", "density"), 0, data_source=ds.all_data())
+    assert_equal("px" in proj.keys(), True)
```
neutrinoceros:
```suggestion
    assert "px" in proj
```

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
+    assert_equal("pz" in proj.keys(), False)
```
neutrinoceros:
```suggestion
    assert "pz" not in proj
```

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
+
+    # Delete the key and check if exits
+    del proj["px"]
+    assert_equal("px" in proj.keys(), False)
```
neutrinoceros:
```suggestion
    assert "px" not in proj
```

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
+    del proj["gas", "density"]
+    assert_equal("density" in proj.keys(), False)
```
neutrinoceros:
```suggestion
    assert "density" in proj
```

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
+
+    # Delete a non-existent field
+    with pytest.raises(YTFieldNotFound) as ex:
+        del proj["p_mass"]
+    desired = "Could not find field 'p_mass' in UniformGridData."
+    assert_equal(str(ex.value), desired)
+
```
neutrinoceros:
```suggestion
    with pytest.raises(YTFieldNotFound, match="Could not find field 'p_mass' in UniformGridData."):
        del proj["p_mass"]
```

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +17,179 @@
...
+    dd = ds.all_data()
+    proj = ds.proj(
+        ("gas", "density"),
+        weight_field=("gas", "cell_mass"),
+        axis=1,
+        data_source=dd,
+    )
+    frb = proj.to_frb((1.0, "unitary"), 64)
+    assert_equal(frb.radius, (1.0, "unitary"))
+    assert_equal(frb.buff_size, 64)
```
neutrinoceros:
```suggestion
    assert frb.radius == (1.0, "unitary")
    assert frb.buff_size == 64
```

neutrinoceros (commented):
it still feels easier than doing it *all* manually, but then the work is in reviewing it (which we should *not* leave to an AI any way)

yut23:
We should remove this exclusion from `pyproject.toml` so the updated tests get run in CI: https://github.com/yt-project/yt/blob/760c7a773a8061a34a2d40ee82aa507e8ae61148/pyproject.toml#L309

neutrinoceros:
Alright I think this is now ready.
My two cents about this experiment: I'm not exactly enthusiastic about using AI here (or at all), but this is a monumental task that the tool actually appears to be suited for, that needs to be done, and that we've been postponing literal years. The fact that it makes the job doable at all, albeit with (reduced) human effort in reviewing, seems to be valuable, especially when we have no perspective of getting fundings to do this the right way.

yt/data_objects/tests/test_data_containers.py
```diff
@@ -18,174 +15,182 @@
 from yt.utilities.exceptions import YTException, YTFieldNotFound
```
yut23, line 15: Should `requires_module` be `yt.testing.requires_module_pytest`? (can't comment on the specific line, but it's right above this)

neutrinoceros: I suppose it might as well, thanks !

neutrinoceros (approved):

neutrinoceros:
>  do you think, overall, this is worthwhile?

yes.

> do you think we should be doing the remaining files in one go?

please, no. :)

PR #5217, opened 2025-07-18
WIP: Experimenting with h5py mocking

matthewturk:
This is another WIP experiment.  I have taken to heart what @xarthisius has said for a while about how it would be nice to mock h5py.

This requires a file that can be generated, called `mock_hdf5_structure.json` that we can use.  Here's code for generating a set of them:

https://gist.github.com/matthewturk/5313f6318d89763fd4282f49c654d500

I've also included a fake one in there.  You'll need to search-and-replace the path to get it to match yours, but if you then have a `.hierarchy` and a parameter file next to it, you should be set.  Then, interactively,

```python
from conftest import patch_h5py
patch_h5py(None)

import yt
ds = yt.load("Whatever/galaxy0030/galaxy0030")
```

Right now it'll generate a field list, but I need to properly implement `MockH5pyH5d.open` and `MockH5pyH5d.read` before it'll do anything else.  I'll give that a go.

This would dramatically reduce the amount of data we'd need.  I think with some additional coding we could mock up some other readers, too.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,10 +2,12 @@ import os
 import shutil
 import sys
 import tempfile
+import types
 from importlib.metadata import version
 from importlib.util import find_spec
 from pathlib import Path
 
+import numpy as np
 import pytest
 import yaml
 from packaging.version import Version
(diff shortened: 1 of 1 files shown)


neutrinoceros (changes requested):
quick note: `conftest.py` shouldn't be used as a module to import stuff from, please use a dedicated module for that

matthewturk:
Fair point!  The import of the mock will need to go in there eventually though, right?  So that pytest can pick it up?

Where should I put it to avoid importing `yt.whatever` -- i.e., having the h5py mock exist outside of it?

(Also that's just in my example script, not planning on having that be used anywhere other than testing the test :) )

Issue #5144, opened 2025-03-27
Updating type hints for numpy arrays


chrishavlin:
There's a number of places where the codebase uses `np.ndarray` in type hints, these should be replaced by the more correct usage of `numpy.typing.NDArray`, creating this as a placeholder issue. See https://github.com/yt-project/yt/pull/5137#pullrequestreview-2701898295 for more info on using `NDArray`.

Possible files to update, may contain false positives when `np.ndarray` is used in a docstring (I haven't double checked them all):

- [ ] yt/visualization/plot_window.py
- [ ] yt/visualization/fixed_resolution_filters.py
- [ ] yt/visualization/_handlers
- [ ] yt/visualization/fixed_resolution.py
- [ ] yt/visualization/plot_modifications.py
- [ ] yt/_typing.py
- [ ] yt/frontends/amrvac/io.py
- [ ] yt/frontends/stream/misc.py
- [ ] yt/frontends/ramses/io.py
- [ ] yt/frontends/ramses/hilbert.py
- [ ] yt/frontends/ramses/particle_handlers.py
- [ ] yt/frontends/rockstar/data_structures.py
- [ ] yt/frontends/artio/data_structures.py
- [ ] yt/loaders.py
- [ ] yt/utilities/io_handler.py
- [ ] yt/geometry/grid_geometry_handler.py
- [ ] yt/geometry/coordinates/coordinate_handler.py
- [ ] yt/geometry/geometry_handler.py
- [ ] yt/geometry/oct_geometry_handler.py
- [ ] yt/testing.py -- should be fixed by #5137 
- [x] yt/utilities/lib/cykdtree/plot.py : only used in docstrings

The results of the full grep&sed command to find these is shown below, including line numbers.

<details>

<summary>Detailed occurrences of `np.ndarray` in type hints </summary>

Here's a bunch of specific places where `np.ndarray` is used in type hints, found by running

```
$ grep -rn np.ndarray --include \*.py yt  > possible_ndarray_typehints.txt 
$ sed '/isinstance/d' possible_ndarray_typehints.txt 
```

on the main branch as of 2025-03-27 (commit 66af53cedb7af599faf3c2471bcd380d961e37a6). Note that some of these **are** valid usages, but most are type hint usages that should be updated.

```
yt/visualization/plot_window.py:1396:    def sanitize_normal_vector(ds, normal) -> str | np.ndarray:
[...]

PR #5218, opened 2025-07-18
Allow parallel iteration over data containers

cphyc:
## PR Summary

This allows to iterate in parallel over data containers.

The syntax is

```python
ds = yt.load(...)
ad = ds.all_data()

# Gathering everything in parallel
my_storage = {}
for sto, chunk in ad.piter(storage=my_storage, reduction="cat"):
    sto.result = {}
    sto.result["gas", "density"] = chunk["gas", "density"]

my_storage["gas", "density"]  # exists for all processes and contains everything


# Gathering everything on root
my_storage = {}
for sto, chunk in ad.piter(storage=my_storage, reduction="cat_on_root"):
    sto.result = {}
    sto.result["gas", "density"] = chunk["gas", "density"]

my_storage["gas", "density"]  # None for all processes but root


# Reduce everything
my_storage = {}
for sto, chunk in ad.piter(storage=my_storage, reduction="sum"):
    sto.result = {}
    sto.result["gas", "density"] = chunk["gas", "density"]

my_storage["gas", "density"]  # contains the sum of all the densities
```

This makes #4730 obsolete.

## PR Checklist

### DIFF
--- a/yt/data_objects/selection_objects/data_selection_objects.py
+++ b/yt/data_objects/selection_objects/data_selection_objects.py
@@ -32,6 +32,7 @@ from yt.utilities.lib.marching_cubes import march_cubes_grid, march_cubes_grid_f
 from yt.utilities.logger import ytLogger as mylog
 from yt.utilities.parallel_tools.parallel_analysis_interface import (
     ParallelAnalysisInterface,
+    parallel_objects,
 )
 
 if sys.version_info >= (3, 11):
@@ -501,6 +502,13 @@ class YTSelectionContainer(YTDataContainer, ParallelAnalysisInterface, abc.ABC):
         self.shape = None
         self._current_chunk = None
         self._min_level = value
+
+    def piter(
+        self,
+        *args,
+        **kwargs,
+    ):
+        yield from parallel_objects(self.chunks([], "io"), *args, **kwargs)
 
 
 class YTSelectionContainer0D(YTSelectionContainer):
--- a/yt/utilities/parallel_tools/parallel_analysis_interface.py
+++ b/yt/utilities/parallel_tools/parallel_analysis_interface.py
@@ -5,6 +5,7 @@ import sys
 import traceback
 from functools import wraps
 from io import StringIO
+from typing import Literal
 
 import numpy as np
 from more_itertools import always_iterable
@@ -466,7 +467,14 @@ class ResultsStorage:
     result_id = None
 
 
-def parallel_objects(objects, njobs=0, storage=None, barrier=True, dynamic=False):
+def parallel_objects(
+    objects,
+    njobs: int = 0,
+    storage: dict | None = None,
+    barrier: bool = True,
+    dynamic: bool = False,
+    reduction: Literal[None, "sum", "max", "min", "cat", "cat_on_root"] = None,
+):
     r"""This function dispatches components of an iterable to different
     processors.
 
@@ -503,7 +511,9 @@ def parallel_objects(objects, njobs=0, storage=None, barrier=True, dynamic=False
         This requires one dedicated processor; if this is enabled with a set of
         128 processors available, only 127 will be available to iterate over
         objects as one will be load balancing the rest.
-
+    reduction : Literal[None, "sum", "max", "min", "cat", "cat_on_root"]
+        This specifies the reduction operation to be applied to the results
+        from each processor.  The default is None, which doesn't apply any.
 
     Examples
     --------
@@ -570,10 +580,22 @@ def parallel_objects(objects, njobs=0, storage=None, barrier=True, dynamic=False
     if parallel_capable:
         communication_system.pop()
     if storage is not None:
-        # Now we have to broadcast it
-        new_storage = my_communicator.par_combine_object(
-            to_share, datatype="dict", op="join"
-        )
+        match reduction:
+            case None:
+                new_storage = my_communicator.par_combine_object(
+                    list(to_share.values()), datatype="dict", op="join"
+                )
+            case "cat":
+                new_storage = my_communicator.all_concat(to_share)
+            case "cat_on_root":
+                new_storage = my_communicator.concat(to_share, root=0)
+            case "sum" | "max" | "min":
+                new_storage = my_communicator.reduce(to_share, op=reduction, root=0)
+            case _:
+                raise NotImplementedError(
+                    f"Reduction operation {reduction} not implemented."
+                )
+
         storage.update(new_storage)
     if barrier:
         my_communicator.barrier()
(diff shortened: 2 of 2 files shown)


neutrinoceros:
about type-checking errors: #5195

brittonsmith (dismissed):
This looks really great. Very concise, happy to see this merged. We can use this to refactor the loop over all halos in the HaloCatalog.

neutrinoceros:
Yes, mypy infers reachability in a slightly broken way that may give off false negatives, maybe this lint should actually be turned off entirely.

brittonsmith (dismissed):

matthewturk (dismissed):
lgtm!

cphyc:
aaaaand I just found a bug - let's not merge this just yet. The issue is with RAMSES reader, where we use progressive indexing (a spatial region isn't fully indexed until it's actually read). The problem is then that different MPI tasks end up reading different regions, and so which chunks may overlap a given region depends on what has been read previously.

If you use this PR's parallel reader one after the other, you may end up in a situation where, e.g. MPI task 1 thinks it needs to read CPUs #1, #2, and #3, but MPI task 2 already read #3 and indexed it, so it know it won't overlap with the region, leaving only #1 and #2 to be read. Since the MPI tasks aren't communicating this information with each other, this will lead to undefined behavior.

chrishavlin:
I think with @matthewturk and @brittonsmith 's prior approvals we could go ahead and merge this one (but let's give them the weekend to potentially respond to this ping).

yt/data_objects/selection_objects/data_selection_objects.py
```diff
@@ -535,6 +536,108 @@ def min_level(self, value):
         self._current_chunk = None
         self._min_level = value
 
+    def piter(
+        self,
+        njobs: int = 0,
+        storage: dict | None = None,
+        barrier: bool = True,
+        dynamic: bool = False,
+        reduction: Literal[None, "sum", "max", "min", "cat", "cat_on_root"] = None,
```
chrishavlin, line 545: ya know, with these changes, i don't think it'd be too hard to allow dynamic MPI reduction operations (i.e, https://mpi4py.readthedocs.io/en/stable/reference/mpi4py.MPI.Op.html#mpi4py-mpi-op ) to allow arbitrary reductions in parallel. that'd be neat!  (obviously not something to do in this PR !)

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -460,13 +462,20 @@ def __getitem__(self, key):
         raise KeyError(key)
 
 
+@dataclass
```
chrishavlin, line 466: ya know. I think this is the first use of `@dataclass` in yt outside of a test file.

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -555,25 +574,36 @@ def parallel_objects(objects, njobs=0, storage=None, barrier=True, dynamic=False
             break
     if parallel_capable:
         communication_system.push_with_ids(all_new_comms[my_new_id].tolist())
-    to_share = {}
+    to_share: dict[int, np.ndarray | dict[Any, np.ndarray]] = {}
```
chrishavlin: Mind using `npt.NDArray` here (see https://github.com/yt-project/yt/issues/5144 ) ? you'll have to `import numpy.typing as npt` up top.

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -917,6 +947,111 @@ def mpi_allreduce(self, data, dtype=None, op="sum"):
             # relatively small ( < 1e7 elements )
             return self.comm.allreduce(data, op)
 
+    def all_reduce(self, data, op="sum"):
+        reduction_op = {
+            "sum": np.sum,
+            "max": np.max,
+            "min": np.min,
+        }[op]
+        mpi_reduction_op = {
```
chrishavlin: there's an `op_names` defined up on line 34, 36 you can use :

```
op_names = {"sum": "MPI.SUM", "min": "MPI.MIN", "max": "MPI.MAX"}
```

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -917,6 +947,111 @@ def mpi_allreduce(self, data, dtype=None, op="sum"):
...
+            reduced_values = reduction_op([reduction_op(d) for d in data.values()])
+
+            if self._distributed:
+                reduced_values = self.comm.allreduce(
+                    reduced_values, op=mpi_reduction_op
+                )
+
+            return reduced_values
+
+    def all_concat(self, data: dict[Any, np.ndarray | dict[str, np.ndarray]]):
```
chrishavlin: Another spot to use `numpy.typing.NDArray`

```suggestion
    def all_concat(self, data: dict[Any, npt.NDArray | dict[str, npt.NDArray]]):
```

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -917,6 +947,111 @@ def mpi_allreduce(self, data, dtype=None, op="sum"):
...
+            )
+
+            if self._distributed:
+                # NOTE: mypy thinks this statement is unreachable (it is not)
+                all_values = np.concatenate(self.comm.allgather(all_values))  # type: ignore[unreachable]
+
+            return all_values
+
+    def concat(
+        self, data: dict[Any, np.ndarray | dict[str, np.ndarray]], root: int = 0
```
chrishavlin: here too 
```suggestion
        self, data: dict[Any, npt.NDArray | dict[str, npt.NDArray]], root: int = 0
```

yt/data_objects/selection_objects/data_selection_objects.py
cphyc: This is a good point; I actually thought about doing exactly this and then gave up for the sake of converging!

yt/utilities/parallel_tools/parallel_analysis_interface.py
```diff
@@ -917,6 +948,111 @@ def mpi_allreduce(self, data, dtype=None, op="sum"):
...
+            "min": MPI.MIN,
+        }[op]
+
+        # We assume the dictionary is not empty!
+        first_value = next(iter(data.values()))
+
+        # If we have a dict of dicts, recurse
+        if isinstance(first_value, dict):
+            return {
+                deeper_key: self.reduce(
```
chrishavlin:
```suggestion
                deeper_key: self.all_reduce(
```

? 

I was getting an attribute error when running the example with the sum reduction to get total mass and volume:

```python-traceback
  File "/Users/chavlin/src/yt_/yt_dev/temp/parallel_data_containers/piteryt.py", line 43, in <module>
    for sto, chunk in ad.piter(storage=my_storage, reduction="sum"):
                      ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chavlin/src/yt_/yt_dev/yt/yt/data_objects/selection_objects/data_selection_objects.py", line 632, in piter
    yield from parallel_objects(
    ...<6 lines>...
    )
  File "/Users/chavlin/src/yt_/yt_dev/yt/yt/utilities/parallel_tools/parallel_analysis_interface.py", line 602, in parallel_objects
    new_storage = my_communicator.all_reduce(to_share, op=reduction)
  File "/Users/chavlin/src/yt_/yt_dev/yt/yt/utilities/parallel_tools/parallel_analysis_interface.py", line 970, in all_reduce
    deeper_key: self.reduce(
                ^^^^^^^^^^^
P001 yt : [ERROR    ] 2026-03-20 15:17:08,692 AttributeError: 'Communicator' object has no attribute 'reduce'
```

cphyc:
> just ran the examples successfully!
> 
> one note: in running the sum reduction I noticed that the units get stripped by the reduction operation. Would be nice to fix that, but IMO we should merge this one and open an issue to fix that later, @cphyc what do you think?
> code snippet and output

I'm afraid that doing another round of modifications would delay that PR again, but I 100% agree with you it'd be a good addition!

chrishavlin:
Here's an issue for later reference: https://github.com/yt-project/yt/issues/5428 

Let's merge this!!

PR #5219, opened 2025-07-18
Add register_fields method

brittonsmith:
## PR Summary

This allows users to define or override properties of on-disk fields, namely units, aliases, and display_name. This is particularly useful if a dataset contains extra fields not defined in a frontend for which the user would like to define units. The syntax is:
```
ds = yt.load(...)
ds.register_field(("enzo", "DinosaurDensity"), units="code_mass/code_length**3")
```

Happy to add docs and tests if this looks broadly acceptable.

## PR Checklist

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -180,6 +180,7 @@ class Dataset(abc.ABC):
     _ionization_label_format = "roman_numeral"
     _determined_fields: dict[str, list[FieldKey]] | None = None
     fields_detected = False
+    _registered_fields = None
 
     # these are set in self._parse_parameter_file()
     domain_left_edge = MutableAttribute(True)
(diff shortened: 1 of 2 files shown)


matthewturk (approved):
Like it!

neutrinoceros:
would it be possible to make the api `ds.fields.register` instead ? I'm not sure what type `ds.fields` is, but that's where I'd look first as a user, and `Dataset` is already very busy as far as attributes and methods go.

brittonsmith:
> I think that _might_ require that the index be created first? Which would get in the way. I do like it more, though.

Yes, this is correct. You need to call this before the field_list is populated.

matthewturk:
I think both would be good, but I think for clarity `ds.register_fields` is probably a bit cleaner than adding more to `load`.

brittonsmith:
Unless I am mistaken, adding a keyword argument to load would be fairly invasive as I would have to make sure it is threaded through every frontend.

matthewturk:
I've been mulling this over, and I think I'm increasingly opposed (GASP, HORROR!) to adding more things to `load`.  I don't think it's a *bad* place to do it, but I think we should take the approach that adding new things there that aren't meant for a *single* frontend can be challenging.

I genuinely think that adding it to the `Dataset` object may be the simplest way, and particularly because it mirrors the way we manage other fields.  I also think that it might be possible to do this in a way that -- like adding gradient fields -- we override things appropriate.  That should probably be a follow-up system.

In short, what I'm coming to think is that we are in a time when we're making some compromises about the existing field system that will stay within where it is (such as this) and those that seek to move it forward (such as #5226 ).  I don't think this is such an invasive change that we should forestall it to a day when the field system is improved -- a day which I never thought *before* would come, but now which seems much more likely.  So I think we should do this like it is, even if it's not where we are in agreement about an ideal.

neutrinoceros:
Overall I'm -0.1 on this: I'm leaning against it, but I'm not invested enough that my vote should be blocking.

Issue #5214, opened 2025-07-18
Improving field list tracking and display


chrishavlin:
I've been looking into how `derived_field_list` and `field_list` are populated and exposed for the dataset objects for the sake of improving the representation in notebooks. 

## Current State of display

For example the following:

```python
import yt 
ds = yt.load_sample("Enzo_64")
ds.field_list
```
just prints everything

<details> 
<summary>Expand to see a large list</summary>

```
[('all', 'creation_time'),
 ('all', 'dynamical_time'),
 ('all', 'metallicity_fraction'),
 ('all', 'particle_index'),
 ('all', 'particle_mass'),
 ('all', 'particle_position_x'),
 ('all', 'particle_position_y'),
 ('all', 'particle_position_z'),
 ('all', 'particle_type'),
 ('all', 'particle_velocity_x'),
 ('all', 'particle_velocity_y'),
 ('all', 'particle_velocity_z'),
 ('enzo', 'Average_creation_time'),
 ('enzo', 'Bx'),
 ('enzo', 'By'),
 ('enzo', 'Bz'),
 ('enzo', 'Cooling_Time'),
 ('enzo', 'Dark_Matter_Density'),
 ('enzo', 'Density'),
 ('enzo', 'Electron_Density'),
 ('enzo', 'Forming_Stellar_Mass_Density'),
 ('enzo', 'Galaxy1Colour'),
 ('enzo', 'Galaxy2Colour'),
 ('enzo', 'HII_Density'),
 ('enzo', 'HI_Density'),
 ('enzo', 'HeIII_Density'),
 ('enzo', 'HeII_Density'),
 ('enzo', 'HeI_Density'),
 ('enzo', 'MBHColour'),
 ('enzo', 'Metal_Density'),
 ('enzo', 'PhiField'),
 ('enzo', 'Phi_pField'),
 ('enzo', 'SFR_Density'),
 ('enzo', 'Star_Particle_Density'),
 ('enzo', 'Temperature'),
 ('enzo', 'TotalEnergy'),
 ('enzo', 'gammaHI'),
 ('enzo', 'kphHI'),
 ('enzo', 'kphHeI'),
 ('enzo', 'kphHeII'),
 ('enzo', 'x-velocity'),
 ('enzo', 'y-velocity'),
 ('enzo', 'z-velocity'),
 ('io', 'creation_time'),
 ('io', 'dynamical_time'),
 ('io', 'metallicity_fraction'),
 ('io', 'particle_index'),
[...]

PR #5224, opened 2025-07-21
Refactor: Move `derived_field_list` to `field_info`

chrishavlin:
## PR Summary

Aimed to be a no-op refactor: this PR moves the `derived_field_list` to an attribute within the dataset's `field_info` container to help with https://github.com/yt-project/yt/issues/5214 . I have a follow up with the same change for the `field_list`, but that's much more invasive, so wanted to check how this smaller changeset does in the test suite (but the change here can be merged on its own).

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -169,7 +169,6 @@ class Dataset(abc.ABC):
     known_filters: dict[ParticleType, ParticleFilter] | None = None
     _index_class: type[Index]
     field_units: dict[AnyFieldKey, Unit] | None = None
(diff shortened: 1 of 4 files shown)


yt/data_objects/static_output.py
```diff
@@ -654,9 +653,13 @@ def print_stats(self):
     def field_list(self):
         return self.index.field_list
 
+    @property
+    def derived_field_list(self):
+        self.index
+        return self.field_info.derived_field_list
+
```
chrishavlin, line 660: just occurred to me that I should add a `setter` here for the sake of downstream frontends.

PR #5226, opened 2025-07-23
ENH: allow omitting the field argument in derived field functions

neutrinoceros:
## PR Summary

Rationale:
I find the mandatory signature of derived field functions counter-intuitive and harmful: the `field` argument is rarely used in user-defined functions, but it's still the *first* argument on the *mandatory* signature, so I regularly find myself defining stuff as
```py
def myfield(data, field):
    ...
```
which is currently a footgun, and the motivation for #4028
I would love this to be actually valid, and ideally, the `field` argument, if unused, should be optional. However, for a long time I couldn't see a backward-compatible path forward, so I abstained from proposing this change.
However, I recently came to the realization that it *should* actually be possible to allow this without breaking backward compatibility, on the condition that we stop passing `data` and `field` arguments *positionally* within yt, which is the starting point of the PR. Provided nothing explodes in CI, I'll work on making user field validation more flexible, as promised in the title.

## PR Checklist

### DIFF
--- a/doc/source/developing/creating_derived_fields.rst
+++ b/doc/source/developing/creating_derived_fields.rst
@@ -320,6 +320,37 @@ Other examples for creating derived fields can be found in the cookbook recipe
 
 .. _derived-field-options:
 
+Accessing field metadata
+------------------------
+
+In some cases, it is useful to introspect the field under construction, for instance
+if specialized behavior is needed for different particle types, or to customize return
+units. These use cases are supported through the optional ``field`` keyword argument.
+Let's look at a couple short examples.
+
+.. code-block:: python
+
+    def get_position_fields(data, field):
+        axis_names = [data.ds.coordinates.axis_name[num] for num in [0, 1, 2]]
+        ftype, _fname = data._determine_fields(field)
+        finfo = data.ds.field_info[ftype]
+        if finfo.sampling_type == "particle":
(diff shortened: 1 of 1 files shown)


neutrinoceros:
it's used in some places to retrieve `field.units` and use it in the return value. You can find all remaining use cases by searching for `(field, data)` within this branch !

cphyc:
It's also used when defining fields for particles, where one doesn't want to hardcode which particle type the field is operating on.

brittonsmith:
Maybe I'm restating what @matthewturk said, but it's used frequently for generic field functions that change behavior based on the specific field it's being used for. This is usually done with `field.name`.

brittonsmith:
I would go so far as to say it is usually not needed. From a user perspective, this is a friendly change. As long we retain support for the `field, data` function signature, I think this is a net positive. I would also feel a lot better if accepting `field` wasn't totally eliminated from the documentation. When this is needed, it's very useful and we run the risk of losing this knowledge.

neutrinoceros:
I just removed it everywhere it was *not* used, but I don't think I eliminated it. The diff obviously doesn't show every example I left alone :)

neutrinoceros:
Thanks for checking ! I'll add a couple examples for it then.

jzuhone (dismissed):

neutrinoceros:
Whoops, looks like I forgot about this for a while. I just added the promised examples. Let me know what you think, @brittonsmith !

PR #5227, opened 2025-07-23
AMR: Expand particle deposition onto the grid to vector fields.

AnatoleStorck:
## PR Summary

Currently, only scalar quantites of particles can be deposited to the grid (such as mass), with final shape (Ncells,) I want to be able to deposit a vector quantity (stellar spectra of each star particle) onto the grid with shape (Ncells, spectra_length), with the main goal to build spectra for each cell (nebular continuum + emission lines, and stellar spectra.) I have not been able to make the deposition of the vector fields work yet, so any help is greatly appreciated.

## PR Checklist

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -1194,7 +1194,7 @@ class Dataset(abc.ABC):
             funit = self._get_field_info(field).units
             out.append(self.arr(np.empty((len(coords),)), funit))
             for coord_index, coord in enumerate(coords):
-                out[field_index][coord_index] = self.point(coord)[field]
+                out[field_index][coord_index] = self.point(coord)[field].item()
         if len(fields) == 1:
             return out[0]
         else:


chrishavlin:
@cphyc some octree enhancements/changes here that you might be interested in checking out when you're back from holiday :)

yt/data_objects/index_subobjects/octree_subset.py
```diff
@@ -130,7 +131,7 @@ def select_tcoords(self, dobj):
     def domain_ind(self):
         return self.oct_handler.domain_ind(self.selector)
 
-    def deposit(self, positions, fields=None, method=None, kernel_name="cubic"):
+    def deposit(self, positions, fields=None, method=None, kernel_name="cubic", vector_field=False):
```
cphyc: You're missing the corresponding item in the docstring to explain what `vector_field` means.

yt/data_objects/index_subobjects/octree_subset.py
```diff
@@ -165,12 +166,22 @@ def deposit(self, positions, fields=None, method=None, kernel_name="cubic"):
...
+            cls = getattr(particle_deposit_vector, f"deposit_{method}", None)
+        else:
+            cls = getattr(particle_deposit, f"deposit_{method}", None)
         if cls is None:
             raise YTParticleDepositionNotImplemented(method)
         nz = self.nz
         nvals = (nz, nz, nz, (self.domain_ind >= 0).sum())
-        if np.max(self.domain_ind) >= nvals[-1]:
+        if vector_field:
+            vec_size = fields[0].shape[-1]
```
cphyc: This won't work if `fields = []` (the default is it is left empty, which can happen).

yt/geometry/oct_container.pyx
```diff
@@ -511,6 +511,10 @@ cdef class OctreeContainer:
         # iterate the same way visit_all_octs does, but we need to track the
         # number of octs total visited.
         cdef np.int64_t num_cells = -1
+
+        source = source.reshape(np.shape(source) + (dims,))
+        # If source is a vector field, we need to handle it differently.
+        is_vector_field = len(np.shape(source)) > 5
```
cphyc: Maybe use `source.ndim == 5` instead?

yt/geometry/oct_container.pyx
```diff
@@ -520,15 +524,19 @@ cdef class OctreeContainer:
...
         if dims != 1:
             raise RuntimeError
-        # Just make sure that we're in the right shape.  Ideally this will not
-        # duplicate memory.  Since we're in Cython, we want to avoid modifying
-        # the .shape attributes directly.
-        dest = dest.reshape((num_cells, 1))
-        source = source.reshape((source.shape[0], source.shape[1],
-                    source.shape[2], source.shape[3], dims))
+        if is_vector_field:
+            dest = np.zeros((len(dest), np.shape(source)[-2], dims),
```
cphyc: I am not sure this is correct - if dest is not None, we should modify it in place instead of creating a new array. It is possible that the upstream code relies on `dest` being modified (if passed).

In addition, `np.shape(source)[-2]` is cryptic and requires knowing about https://github.com/yt-project/yt/pull/5227/files#diff-3cef3e330a489f3d2975d39b0f17b55a03f3134b27492604af30bc23f1bf7fa7R179. What about something like:
```python
*_, ncells, ndim = source.shape
# assert ndim == dims?
dest = np.zeros((len(dest), ncells, dims), dtype=source.dtype, order='C')
```

yt/geometry/oct_visitors.pyx
```diff
@@ -71,6 +71,23 @@ cdef class CopyArrayF64(OctVisitor):
...
+cdef class Copy2DArrayF64(OctVisitor):
+    @cython.boundscheck(False)
+    @cython.initializedcheck(False)
+    cdef void visit(self, Oct* o, np.uint8_t selected):
+        # We should always have global_index less than our source.
+        # "last" here tells us the dimensionality of the array.
+        if selected == 0: return
+        # There are this many records between "octs"
+        #print("IN COPY2DARRAYF64")
+        #print(self.dest.shape, self.source.shape)
```
cphyc:
```suggestion
```

yt/geometry/particle_deposit_vector.pxd
```diff
@@ -0,0 +1,43 @@
...
+
+cdef extern from "platform_dep.h":
+    void *alloca(int)
+
+cdef inline int gind(int i, int j, int k, int dims[3]):
+    # The ordering is such that we want i to vary the slowest in this instance,
+    # even though in other instances it varies the fastest.  To see this in
+    # action, try looking at the results of an n_ref=256 particle CIC plot,
+    # which shows it the most clearly.
+    return ((i*dims[1])+j)*dims[2]+k
```
cphyc: Shouldn't this be `(i*dims[0] + j) * dims[1] + k`?

yt/geometry/particle_deposit_vector.pyx
```diff
@@ -0,0 +1,195 @@
...
+
+cdef append_axes(np.ndarray arr, int naxes):
+    if arr.ndim == naxes:
+        return arr
+    # Avoid copies
+    arr2 = arr.view()
+    arr2.shape = arr2.shape + (1,) * (naxes - arr2.ndim)
+    return arr2
+
+cdef class ParticleDepositOperation:
```
cphyc: Could it superclass the one defined in `particle_deposit.pyx` to minimize code duplication?

PR #5233, opened 2025-07-26
MNT: adjust PyPI trove classifiers

neutrinoceros:
## PR Summary

`project.license` is deprecated as a TOML table since `setuptools` `77.0.0`:
https://setuptools.pypa.io/en/latest/history.html#id73
I adjusted for this change using the now prefered standard, [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys), which also mentions the `License :: OSI Approved :: BSD License` classifier being ambiguous (other build backends such as `flit-core` also emmit warnings over it, discouraging its use), so I removed it.
Setuptools is supposed to auto-include files listed in `project.license-files`, so we don't need to explicitly ask for it in `MANIFEST.in`

While I was at it, I noticed we were missing a couple `Programming Language` trove classifiers, so I added them.

## PR Checklist


pyproject.toml
```diff
@@ -15,16 +15,21 @@ description = "An analysis and visualization toolkit for volumetric data"
 authors = [
     { name = "The yt project", email = "yt-dev@python.org" },
 ]
+license = "BSD-3-Clause"
+license-files = [
+    "yt/frontends/artio/artio_headers/LICENSE",
```
chrishavlin, line 22: it's a bit out of date... but `COPYING.txt` has the root level yt license. can we include it here?

neutrinoceros: absolutely ! I knew I was missing something... thanks !

PR #5236, opened 2025-07-28
update COPYING.txt

chrishavlin:
1. update the references to bitbucket to point to github
2. specify which version range are BSD-3


COPYING.txt
```diff
@@ -47,13 +47,14 @@ from all developers and on a public mailing list, yt has been relicensed under
 the BSD 3-clause under a shared copyright model.  For more information, see:
 https://mail.python.org/archives/list/yt-dev@python.org/thread/G4DJDDGB4PSZFJVPWRSHNOSUMTISXC4X/
 All versions of yt prior to this licensing change are available under the
-GPLv3; all subsequent versions are available under the BSD 3-clause license.
+GPLv3; all subsequent versions (yt versions >= 2.0.0) are available under
```
chrishavlin: @matthewturk (or anyone else involved back then... @Xarthisius @jzuhone ) did I guess right? was yt 2.0 the first release under BSD-3?

Xarthisius: According to pypi it was `>= 2.6.0`

Also: https://yt-project.github.io/blog/posts/relicensing/

chrishavlin: great! thanks, will fix it here. And probably link to that blog post, I had missed that in my quick attempt to figure out where the change occurred.

matthewturk: Thanks for making this change!  I think we had a commit at some point that was tagged as last gpl.

matthewturk (approved):

COPYING.txt
Xarthisius: https://github.com/yt-project/yt/releases/tag/last_gplv3

PR #5250, opened 2025-08-04
Convert field_info into a property so it can be created on first access.

brittonsmith:
The field info container is notably brittle in that either a dataset's index or field_list must be accessed/created first before it can come into existence.  This PR makes `Dataset.field_info` into a property allowing it to be created on first access. Additionally, we now poke the index at the start of `create_field_info` to allow this to be run before the `field_list` is created. I argue this should have been considered a bug given the naming of the method without an underscore implying the user was free to call it.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -658,7 +658,7 @@ class Dataset(abc.ABC):
     @property
     def field_info(self):
         if self._field_info is None:
-            self.create_field_info()
+            self.index
         return self._field_info
 
     @field_info.setter
@@ -666,7 +666,13 @@ class Dataset(abc.ABC):
         self._field_info = value
 
     def create_field_info(self):
-        self.index
+        # create_field_info will be called at the end of instantiating
+        # the index object. This will trigger index creation, which will
+        # call this function again.
+        if self._instantiated_index is None:
+            self.index
+            return
+
         self.field_dependencies = {}
         self.derived_field_list = []
         self.filtered_particle_types = []


yt/data_objects/static_output.py
```diff
@@ -654,7 +655,18 @@ def print_stats(self):
     def field_list(self):
         return self.index.field_list
 
+    @property
+    def field_info(self):
+        if self._field_info is None:
+            self.create_field_info()
+        return self._field_info
+
+    @field_info.setter
```
neutrinoceros, line 664: Is this necessary ?
Immutability has a lot of value when dealing with parallelism, so I would advise to avoid introducing mutability where it's not needed.

yt/data_objects/static_output.py
```diff
@@ -654,7 +655,18 @@ def print_stats(self):
...
+    def field_info(self, value):
+        self._field_info = value
+
     def create_field_info(self):
+        self.index
```
brittonsmith: I'm going to admit that this isn't a great thing to do as instantiating the index called `create_field_info`. What I think would be better is renaming this `_create_field_info` to remove the expectation that anyone ever call it. I would then change line 661 above to simply `self.index`.

yt/data_objects/static_output.py
brittonsmith: We are already doing this in far more critical places, e.g., the index. In fact, I would argue that we want some measure of this here.

neutrinoceros: But what for ? Unless there's a test exercising it I don't understand the point.

yt/data_objects/static_output.py
brittonsmith: I've refactored this so that we only ever do one pass through `create_field_info`.

yt/data_objects/static_output.py
brittonsmith: Sorry, I replied without really thinking this through. We don't do this with index, but we probably do want to allow this to be settable.

brittonsmith: Specifically, without this, we can't do what's in the tests in PR #5211.

neutrinoceros: I think I'm even more confused now. I don't see the connection with #5211

brittonsmith: I'm referring to how the tests in that PR manually alter entries in `field_info`. You wouldn't be able to do that without having the setter.

brittonsmith: I've come around on not having a setter for `field_info`. When it wasn't a property we allowed it to be set by anyone, but maybe it's better to be safe since we can.

brittonsmith: Just to tie this off, the setter is necessary if we want to avoid modifying all the places in the frontends that setup the field_info on their own, which I believe we do. I also think we should not worry about immutability since we didn't have that before anyway. There's no reason to impose this restriction now. This change doesn't modify existing behavior and only allows us to refer to `ds.field_info` without having to think about whether it has been created yet. I don't see a reason not to do this.

chrishavlin:
what milestone ya'll want this in? it's got a bug label so i'm assuming we'd want this to ship with 4.4.2?

brittonsmith:
I'll have a look through the tests and see if there are other places where we do that. This is a very good point you raise about side-effects and that dataset objects can be in different states. It's something we should probably try to be more explicit about.

brittonsmith:
> what milestone ya'll want this in? it's got a bug label so i'm assuming we'd want this to ship with 4.4.2?

Yes, I think this was technically a bug, although one we happily lived with for quite some time. I think we can go with 4.4.2. Thanks @chrishavlin!

chrishavlin:
@meeseeksdev backport to yt-4.4.x

PR #5247, opened 2025-08-03
TST: fix image tests reporting-on-error and update failing tests

neutrinoceros:
## PR Summary
Looks like images tests are currently doubly broken: 2 of them fail, but the reporting step is itself broken on `pytest>=8.4`, so I'm starting with a fix for this last part, as a follow up to #5176. I'll see if I can fix the tests themselved then.

## PR Checklist

- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #5253, opened 2025-08-06
[ENH] Define bounding box for any data selection

cphyc:
## PR Summary

This allows bounding boxes to be defined for *any* data selection object, not only in 3D.
By default, the bounding box encompasses the entire simulation. When a data selection inherits from another one, we take the bounding box of the parent bbox, etc.

Before, 1- and 2D data containers would have no bounding box, preventing us from filtering which regions may intersect with the 1- and 2D container. Here, I take a conservative estimate: if a $N\rm D$ container inherits from an $M\rm D$ container, with $N\leq M$, then the former inherits the bounding box of the latter.

This allows speeding up, for example slices (for RAMSES datasets at least):

```python
import yt
yt.set_log_level(10)  # to see the IO footprint
sp = ds.sphere([0.5]*3, 0.1)
p = yt.SlicePlot(ds, "x", ("gas", "density"), data_source=sp)

p.save()
```

In the above example, we go from reading all domains (since we have no bounding box to find the intersection between the on-file domains) to only reading those two that intersect with the sphere. Note that we _could_ be even more efficient and further reduce the size of the bounding box to a thin slab.

## PR Checklist

### DIFF
--- a/yt/data_objects/tests/test_data_containers.py
+++ b/yt/data_objects/tests/test_data_containers.py
@@ -4,8 +4,10 @@ import numpy as np
 import pytest
 from numpy.testing import assert_array_equal, assert_equal
 
-from yt.data_objects.data_containers import YTDataContainer
 from yt.data_objects.particle_filters import particle_filter
+from yt.data_objects.selection_objects.data_selection_objects import (
+    YTSelectionContainer,
+)
 from yt.testing import (
     fake_amr_ds,
     fake_particle_ds,
@@ -33,7 +35,7 @@ def test_yt_data_container():
             " type or parameter to the constructor"
         ),
     ):
-        YTDataContainer(None, None)
+        YTSelectionContainer(None, None)
 
     # Test if field_data key exists
     ds = fake_random_ds(5)


cphyc:
The image test failures are unrelated (see #5247 and https://github.com/yt-project/yt_pytest_mpl_baseline/pull/9).

yt/data_objects/selection_objects/data_selection_objects.py
```diff
@@ -41,15 +42,14 @@
...
 class YTSelectionContainer(YTDataContainer, ParallelAnalysisInterface, abc.ABC):
-    _locked = False
-    _sort_by = None
-    _selector = None
-    _current_chunk = None
-    _data_source = None
+    _locked: bool = False
+    _selector: SelectorObject | None = None
+    _current_chunk: YTDataContainer | None = None
+    _data_source: "YTSelectionContainer | None" = None
```
cphyc: A note here to reviewers: I _think_ `_data_source` cannot be `YTDataContainer`, since it is the base class and is never subclassed except for here. In other words, `YTSelectionContainer` is the most general use-facing class.

We could also implement `get_bbox` as an abstract method to `YTDataContainer` to force it to be defined everywhere.

yt/data_objects/selection_objects/data_selection_objects.py
```diff
@@ -222,6 +222,39 @@ def get_data(self, fields=None):
...
+            case Geometry.CARTESIAN:
+                # mypy does not understand that here, _data_source cannot be None
+                return self._data_source.get_bbox()  # type: ignore [union-attr]
+            case (
+                Geometry.CYLINDRICAL
+                | Geometry.POLAR
+                | Geometry.SPHERICAL
+                | Geometry.GEOGRAPHIC
+                | Geometry.INTERNAL_GEOGRAPHIC
+                | Geometry.SPECTRAL_CUBE
```
matthewturk, line 249: Not necessary here, but we're getting more and more instances of implementing on geometries and I'm wondering if we should have a specific comment we always use or something to indicate all the places we need to exhaustively cover geometries.

jzuhone (approved):

PR #5270, opened 2025-08-25
baroclinic vorticity typo fix in yt/fields/fluid_vector_fields.py

CharlieWillard1:
There was small typo in the code to calculate the x component on the baroclinic vorticity term, just a small switch of z and y component on the density gradient used in the calculation.

## PR Summary

There was small typo in the function to calculate the baroclinic_vorticity_x term. It used the z component of the density gradient rather than the y component of the density gradient. The change I made was simply deleting the z for a y.
Should be a 1 character change.

## PR Checklist

- [] Adds a test for any bugs fixed. Adds tests for new features.

I did not do any testing, just a bug I noticed by eye.


cphyc (approved):
Good catch and thanks for the fix! I'll merge once the tests have run.

PR #5276, opened 2025-08-29
Fix small bugs/edge cases in the parthenon frontend

Yurlungur:
## PR Summary

We recently *finally* refactored curvilinear coordinates in Parthenon and the ordering for cylindrical coordinates was normalized to the standard choice in yt. This MR fixes that in parthenon frontend. I also noticed that the yt frontend was not properly handling variables of tensor rank greater than 1 (i.e., 2d matrices on a mesh). This MR adds a one-line fix that gets that working.

## PR Checklist

### DIFF
--- a/yt/frontends/parthenon/data_structures.py
+++ b/yt/frontends/parthenon/data_structures.py
@@ -164,10 +164,7 @@ class ParthenonDataset(Dataset):
 
         self.geometry = _geom_map[self._handle["Info"].attrs["Coordinates"]]
 
-        if self.geometry is Geometry.CYLINDRICAL:
-            axis_order = ("r", "theta", "z")
-        else:
-            axis_order = None
+        axis_order = None
 
         Dataset.__init__(
             self,
--- a/yt/frontends/parthenon/io.py
+++ b/yt/frontends/parthenon/io.py
@@ -52,7 +52,10 @@ class IOHandlerParthenon(BaseIOHandler):
                 for gs in grid_sequences(chunk.objs):
                     start = gs[0].id - gs[0]._id_offset
                     end = gs[-1].id - gs[-1]._id_offset + 1
-                    data = ds[start:end, fdi, :, :, :].transpose()
+                    if len(ds.shape) == 4:
+                        data = ds[start:end, :, :, :].transpose()
+                    else:
+                        data = ds[start:end, fdi, :, :, :].transpose()
                     for i, g in enumerate(gs):
                         ind += g.select(selector, data[..., i], rv[field], ind)
             last_dname = dname
@@ -72,7 +75,10 @@ class IOHandlerParthenon(BaseIOHandler):
             for gs in grid_sequences(chunk.objs):
                 start = gs[0].id - gs[0]._id_offset
                 end = gs[-1].id - gs[-1]._id_offset + 1
-                data = ds[start:end, fdi, :, :, :].transpose()
+                if len(ds.shape) == 4:
+                    data = ds[start:end, :, :, :].transpose()
+                else:
+                    data = ds[start:end, fdi, :, :, :].transpose()
                 for i, g in enumerate(gs):
                     rv[g.id][field] = np.asarray(data[..., i], "=f8")
         return rv


yt/frontends/parthenon/io.py
```diff
@@ -52,7 +52,10 @@ def _read_fluid_selection(self, chunks, selector, fields, size):
                 for gs in grid_sequences(chunk.objs):
                     start = gs[0].id - gs[0]._id_offset
                     end = gs[-1].id - gs[-1]._id_offset + 1
-                    data = ds[start:end, fdi, :, :, :].transpose()
+                    if len(ds.shape) == 4:
+                        data = ds[start:end, :, :, :].transpose()
+                    else:
+                        data = ds[start:end, fdi, :, :, :].transpose()
```
Yurlungur, line 58: Parthenon's python reader can always split a variable into a flat index space, so I *think* this is now generic.

yt/frontends/parthenon/data_structures.py
```diff
@@ -164,10 +164,7 @@ def __init__(
 
         self.geometry = _geom_map[self._handle["Info"].attrs["Coordinates"]]
 
-        if self.geometry is Geometry.CYLINDRICAL:
-            axis_order = ("r", "theta", "z")
-        else:
-            axis_order = None
+        axis_order = None
```
chrishavlin: this would break things for the old format, ya? anyway to detect the new format so you could make this change backward compatible?

pgrete:
@Yurlungur do you have any data to share with those coordinate that could be added to the regression tests?
I think some choice were specific to the (not merged) changes in AthenaPK, so I'd be happy to follow the new "standard" way if we merge non-Cart coords in AthenaPK.

yt/frontends/parthenon/io.py
pgrete: I'm not sure I understand this change. What's exactly happening here?
a) the Parthenon python reader is not used in yt
b) a flattened index space (from a vector or tensor variable) would always carry the `fdi` index, doesn't it?

Yurlungur: In riot's current format scalar fields are only 4D with the leading index the block. AthenaPK doesn't see this because it has a prim vector, but riot and phoebus do. That's what's this change is. I didn't modify the piece with prim vector, just special case scalars.

yt/frontends/parthenon/data_structures.py
Yurlungur: Thanks for the review! See @pgrete 's comment below. The old version was for a draft of curvilinear coordinates that never made it into the code, so this isn't actually breaking. I think it's safe to merge this without a version check. (That said parthenon does have version checking.)

chrishavlin: > was for a draft of curvilinear coordinates that never made it into the code

ok great,  thanks for clarifying!

pgrete:
btw if you quickly want to run local tests, just 
```
nosetests -v --with-answer-testing --local --local-dir $HOME/src/yt-test-data --answer-store --answer-name=local-parthenon yt.frontends.parthenon
```
where `$HOME/src/yt-test-data` (or whatever folder of you choice) contains the data in the zip file (including the folder) -- and eventually data from https://yt-project.org/data/

Yurlungur:
> Dataset PR is here: [yt-project/website#141](https://github.com/yt-project/website/pull/141)
> 
> Also, I added a skeleton on how to test the data (currently just checking for the field name). Any creative suggestion on how to test the data itself @Yurlungur ?

Sorry for the delay. Been traveling. I added a check that the cell width in theta is 2pi. (This was being incorrectly overwritten, by the way, which I fixed.) I also added a check for the total mass.

pgrete:
@chrishavlin this now looks good to me.
Who can upload the new reference data to the yt website and accept https://github.com/yt-project/website/pull/141 so that the tests can pass?

chrishavlin:
@pgrete I think we need @Xarthisius to move the new dataset from https://github.com/yt-project/website/pull/141 to where it needs to go... but the test server that runs the full test suite is currently on the fritz, so the full nose test suite will not run at present (right now, the only tests that are running are those that do not require sample datasets). Doesn't hurt to have the dataset in place for when we revamp the test server, but not sure when that will be. In the meantime, I'll  double check that the new tests run for me locally right now and we should be good to merge.

yt/frontends/parthenon/data_structures.py
```diff
@@ -164,10 +160,7 @@ def __init__(
...
```
chrishavlin: You can simply remove this entirely (and remove the corresponding line in the Dataset.__init__() call, line 172 below) and rely on the built in default handling. 

```suggestion
```

chrishavlin (commented):
I did run the existing answer tests locally and confirmed that they passed -- was not able to run the new test due to an error in unpacking the new test file (see https://github.com/yt-project/website/pull/141#issuecomment-3377551372 ). 

One final minor code suggestion if you want to fix it, but IMO this is good to merge. I'll hit approve once we get the new test file figured out.

Yurlungur:
🎊 Thanks for your feedback and helping pull it all through @chrishavlin @pgrete !

PR #5287, opened 2025-09-12
BLD: avoid direct dependency on wheel library

neutrinoceros:
## PR Summary

While setting up abi3 wheels for another package (h5py), I discovered this cleaner way to achieve the same result; instead of overriding the `get_tag` method, we just pass `py_limited_api` as an option.


neutrinoceros:
@chrishavlin can you have a look ?

PR #5294, opened 2025-10-05
IPython.core.display.display has been removed

zingale:
this used the new interface
see https://github.com/ipython/ipython/pull/14657

## PR Checklist

### DIFF
--- a/yt/units/__init__.py
+++ b/yt/units/__init__.py
@@ -117,9 +117,6 @@ def display_ytarray(arr):
 
 
 def _wrap_display_ytarray(arr):
-    try:
-        from IPython.display import display
-    except ImportError:
-        from IPython.core.display import display
+    from IPython.display import display
 
     display(display_ytarray(arr))
--- a/yt/units/__init__.py
+++ b/yt/units/__init__.py
@@ -117,6 +117,9 @@ def display_ytarray(arr):
 
 
 def _wrap_display_ytarray(arr):
-    from IPython.core.display import display
+    try:
+        from IPython.core.display import display
+    except ImportError:
+        from IPython.display import display
 
     display(display_ytarray(arr))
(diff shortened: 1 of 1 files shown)


jzuhone (dismissed):

yt/units/__init__.py
```diff
@@ -117,6 +117,9 @@ def _value_updater(change):
 
 
 def _wrap_display_ytarray(arr):
-    from IPython.core.display import display
+    try:
+        from IPython.core.display import display
```
neutrinoceros: This should be the exceptional path, not the first one we try.
Better yet would be to check on IPython's version at runtime instead of using try/except, so the code would self-document as obsolete when we drop support for older versions.

neutrinoceros (changes requested):
Did you make sure this import was also removed everywhere else ?

zingale:
this is the only place that had that import.  There are other imports from `IPython.core.display` that are getting things other than `display`, but I don't know if those are deprecated.  I only had issues with this one.

yt/units/__init__.py
zingale: I've updated the ordering

yt/units/__init__.py
```diff
@@ -117,6 +117,9 @@ def _value_updater(change):
 
 
 def _wrap_display_ytarray(arr):
-    from IPython.core.display import display
+    try:
+        from IPython.display import display
+    except ImportError:
+        from IPython.core.display import display
```
neutrinoceros: actually we don't support versions older than 7.16.2 (see `pyproject.toml`), and this works with all supported versions
```suggestion
    from IPython.display import display
```

neutrinoceros:
FWIW I checked all other `IPython.core` imports; this is the only one that was deprecated (and removed)

yt/units/__init__.py
zingale: works for me

neutrinoceros (approved):

Issue #5096, opened 2025-01-10
Not able to read the Halo Finder output generated using HOP halo finder..


chavan-shende:
What I am trying is to use yt HOP halo finder for identifying halos in a gadget/gizmo snapshot generated for DM cosmological Simulation. I was able to read the file, and then create the HaloCatalog, but when I am trying to load the HaloCatalog I am getting the error...

```python
from yt.extensions.astro_analysis.halo_analysis import HaloCatalog

ds = yt.load("snapshot_012.0.hdf5")

hc = HaloCatalog(data_ds=ds, finder_method="hop")
hc.create()
-------------------------------------------------------------------------------------------------------------
'''  The HaloCatalog was created successfully '''
yt : [INFO     ] 2025-01-09 22:03:43,323 Omega Lambda is 0.0, so we are turning off Cosmology.
yt : [INFO     ] 2025-01-09 22:03:43,324 Assuming length units are in kpc (physical)
yt : [INFO     ] 2025-01-09 22:03:43,354 Parameters: current_time              = 0.39999999999999997
yt : [INFO     ] 2025-01-09 22:03:43,355 Parameters: domain_dimensions         = [1 1 1]
yt : [INFO     ] 2025-01-09 22:03:43,355 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2025-01-09 22:03:43,356 Parameters: domain_right_edge         = [100. 100. 100.]
yt : [INFO     ] 2025-01-09 22:03:43,357 Parameters: cosmological_simulation   = 0
yt : [INFO     ] 2025-01-09 22:03:43,398 Allocating for 2.097e+06 particles
Loading particle index: 100%|█████████████████████████| 9/9 [00:00<00:00, 816.61it/s]
yt : [INFO     ] 2025-01-09 22:03:43,944 Initializing HOP
Calling hop... 2097152 1.600e+02
yt : [INFO     ] 2025-01-09 22:04:08,221 Parsing outputs
Copying arrays for 2097152 particles
Building Tree...
Finding Densities...
Finding Densest Neighbors...
Grouping...
Merging Groups...
Writing Output...
All Done!
nSmooth = 65 kd->nActive = 2097152
Calling regroup...
No minimum group size specified.  Assuming 10 particles.
ngroups = 9545
yt : [INFO     ] 2025-01-09 22:04:24,939 Saving 1239 halos: halo_catalogs/snapshot_012/snapshot_012.0.h5.
````
[...]

PR #5300, opened 2025-10-08
Raise exception when index does not have print_stats method

brittonsmith:
Issue #5096 also noted that `ds.print_stats()` fails uninformatively for index types that don't implement it. This raises a NoteImplementedError that is hopefully a little more useful.

### DIFF
--- a/yt/data_objects/static_output.py
+++ b/yt/data_objects/static_output.py
@@ -649,7 +649,10 @@ class Dataset(abc.ABC):
 
     @parallel_root_only
     def print_stats(self):
-        self.index.print_stats()
+        func = getattr(self.index, "print_stats", None)
+        if func is None:
+            raise NotImplementedError(f"{type(self.index)} has no print_stats method.")
+        func()
 
     @property
     def field_list(self):
--- a/yt/data_objects/tests/test_dataset_access.py
+++ b/yt/data_objects/tests/test_dataset_access.py
@@ -1,3 +1,5 @@
+from unittest import TestCase
+
 import numpy as np
 from nose.tools import assert_raises
 from numpy.testing import assert_almost_equal, assert_equal
@@ -182,6 +184,12 @@ def test_ortho_ray_from_r():
     ray7 = ds.r[0.25:0.75:100j, 0.3, 0.2]
     ray8 = LineBuffer(ds, [0.2525, 0.3, 0.2], [0.7475, 0.3, 0.2], 100)
     assert_equal(ray7["gas", "density"], ray8["gas", "density"])
+
+
+def test_no_print_stats():
+    ds = fake_particle_ds()
+    with TestCase.assertRaises(NotImplementedError):
+        ds.print_stats()
 
 
 def test_particle_counts():


yt/data_objects/tests/test_dataset_access.py
```diff
@@ -184,6 +186,12 @@ def test_ortho_ray_from_r():
     assert_equal(ray7["gas", "density"], ray8["gas", "density"])
 
 
+def test_no_print_stats():
+    ds = fake_particle_ds()
+    with TestCase.assertRaises(NotImplementedError):
```
neutrinoceros: nose.tools.assert_raises is already imported in this module, so maybe we should use it for consistency ?
More importantly, if this test module requires nose, it also means it's not being run any where at the moment so ... maybe add the test to a pytest only test module and use `pytest.raises` instead ?

yt/data_objects/static_output.py
```diff
@@ -649,7 +649,10 @@ def print_key_parameters(self):
 
     @parallel_root_only
     def print_stats(self):
-        self.index.print_stats()
+        func = getattr(self.index, "print_stats", None)
+        if func is None:
+            raise NotImplementedError(f"{type(self.index)} has no print_stats method.")
```
neutrinoceros: this seems very hacky, and makes the code more difficult to reason about. Shouldn't it be the responsibility of the class that's missing the method to raise an exception instead ?

chrishavlin: the base `Index` class (`yt.geometry.geometry_handler.Index`) also doesn't currently define `print_stats` at all -- so would be straightforward to add it there with a `NotImplementedError` default definition.

yt/data_objects/tests/test_dataset_access.py
brittonsmith: I don't see why we would add anything nose specific since we want to ultimately migrate away from nose. TestCase is independent of pytest or nose.

yt/data_objects/static_output.py
brittonsmith: @chrishavlin, thanks, going to the base class was a better solution. I've done that now.

yt/data_objects/tests/test_dataset_access.py
neutrinoceros: yes, we agree on that, but by adding this test case to this module, you're de-facto making it nose-only. This is what I wanted to bring attention to.

yt/data_objects/tests/test_print_stats.py
chrishavlin, line 1: I'm kinda confused why this didn't fail with the github CI nose test :)  

usually have to add the pytest only test scripts to `nose_ignores.txt`...

chrishavlin:
@neutrinoceros I think your concerns have been addressed, mind taking another quick look?

yt/data_objects/tests/test_print_stats.py
neutrinoceros: I'm startled too. Maybe that's because nose is actually compatible with a subset of pytest's capabilities, and this file doesn't use any pytest fixtures, which I suspect are the real reason why nose can't run most pytest modules ?

neutrinoceros (approved):
I'm puzzled to see that nose is apparently able to  run these tests, but as long as it does, LGTM now

PR #5307, opened 2025-10-19
BUG: avoid soon-to-be-deprecated direct mutations of ndarray.shape attributes (1/2)

neutrinoceros:
## PR Summary

xref: https://github.com/numpy/numpy/pull/29536

This is incomplete at the time of opening, I just want to take what I already have for a spin on CI

## PR Checklist

- [ ] New features are documented, with docstrings and narrative docs
- [ ] Adds a test for any bugs fixed. Adds tests for new features.

PR #5308, opened 2025-10-19
ENH: ensure `yt.utils.funcs.ensure_numpy_array` consistently returns a copy

neutrinoceros:
## PR Summary
I noticed this function could be heavily simplified, but also, and perhaps more importantly, that its return type would sometimes be a copy of the input data, but not always. This looks like a footgun so I'd rather make it always return a copy, so as to avoid subtle bugs.

## PR Checklist

### DIFF
--- a/yt/funcs.py
+++ b/yt/funcs.py
@@ -86,15 +86,7 @@ def ensure_numpy_array(obj):
     This function ensures that *obj* is a numpy array. Typically used to
     convert scalar, list or tuple argument passed to functions using Cython.
     """
-    if isinstance(obj, np.ndarray):
-        if obj.shape == ():
-            return np.array([obj])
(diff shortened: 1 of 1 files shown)


matthewturk:
nah it's fine, you pointed out that I was interpreting the change incorrectly.

neutrinoceros:
CI revealed that `np.asarray` didn't have a `copy` argument before numpy 2.0. I could work around this at a local level, but incidentally, I'm adding a general mechanism for handling this problem gracefully across the board in #5307, so I'd prefer to re-work this PR once the #5307 one goes in.

neutrinoceros:
Actually it was easy enough to bring only the backward-compatibility layer from the other PR, so I plan to undraft this after the on-going round of CI.

PR #5309, opened 2025-10-20
BUG: avoid soon-to-be-deprecated direct mutations of ndarray.shape attributes (2/2)

neutrinoceros:
## PR Summary

Complements #5307.
This one is specifically focused on volume rendering stuff, and separated because it's not *strictly* backwards compatible, as it also includes slight changes to (internal ?) APIs in order to reduce the (useless) diversity of this module.

## PR Checklist

### DIFF
--- a/yt/visualization/volume_rendering/lens.py
+++ b/yt/visualization/volume_rendering/lens.py
@@ -341,13 +341,13 @@ class StereoPerspectiveLens(Lens):
 
         uv = np.ones(3, dtype="float64")
 
-        image = self.new_image(camera)
-        vectors_comb = uvstack([vectors_left, vectors_right])
-        positions_comb = uvstack([positions_left, positions_right])
-
-        image.shape = (camera.resolution[0], camera.resolution[1], 4)
-        vectors_comb.shape = (camera.resolution[0], camera.resolution[1], 3)
-        positions_comb.shape = (camera.resolution[0], camera.resolution[1], 3)
+        image = self.new_image(camera).reshape(*camera.resolution, 4)
+        vectors_comb = uvstack([vectors_left, vectors_right]).reshape(
+            *camera.resolution, 3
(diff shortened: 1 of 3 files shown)


yt/visualization/volume_rendering/lens.py
```diff
@@ -652,9 +649,16 @@ def _get_sampler_params(self, camera, render_source):
...
+        positions = np.tile(camera.position, np.prod(camera.resolution)).reshape(
+            *camera.resolution, 3
         )
+
+        R1 = get_rotation_matrix(0.5 * np.pi, [1, 0, 0])
+        R2 = get_rotation_matrix(0.5 * np.pi, [0, 0, 1])
+        uv = np.dot(R1, camera.unit_vectors)
+        uv = np.dot(R2, uv)
+
+        vectors = np.zeros((*camera.resolution, 3), dtype="float64", order="C")
```
neutrinoceros:
```suggestion
        vectors = np.empty((*camera.resolution, 3), dtype="float64", order="C")
```

yt/visualization/volume_rendering/old_camera.py
```diff
@@ -1744,7 +1732,7 @@ def get_sampler_args(self, image):
         return args, {}
 
     def finalize_image(self, image):
-        image.shape = self.resolution, self.resolution, 4
+        return image.reshape(self.resolution, self.resolution, 4)
```
neutrinoceros, line 1735: @chrishavlin this is what I was refering to when I said this wasn't strictly backward compatible. Admittedly I'm leaning heavily on Hyrum's law, but this function changed in two different ways: it's not mutating its argument any more, and it now returns something other than None.

chrishavlin: ok, cool. thanks!!

